PR #19, opened 2014-03-24
Mark the client as focused in client.jumpto first

blueyed:
This avoids awful.autofocus to kick in, focusing some other client in
between (where a focused signal would get emitted for, although the
intermediate client was never meant to have focus).


blueyed:
> We don't magically make it sticky.

#107 does this now.

PR #29, opened 2014-04-07
Remember layout across restarts (PR for client properties)

blueyed:
This a proof of concept to remember layout properties across restarts.

For now, only the floating state of clients is remembered.

This is meant to collect feedback and then also apply it e.g. to the selected layout of tags.
What else should be covered and restored?

Ref: https://awesome.naquadah.org/bugs/index.php?do=details&task_id=759

### DIFF
--- a/lib/awful/client.lua.in
+++ b/lib/awful/client.lua.in
@@ -18,6 +18,7 @@ local capi =
     client = client,
     mouse = mouse,
     screen = screen,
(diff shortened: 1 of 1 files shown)


blueyed:
I've started implementing this for tags in #102.

It works for layouts already (by using their names), and other properties should also work, but are not enabled by default.

Please provide any feedback over there at the new PR.

PR #56, opened 2014-10-22
Layouts can define their own resizing handler (FS#1267)

RobSis:
Rationale: `awful.mouse.resize` checks a hard-coded list of layouts. It's impossible for any third-party layout to easily support mouse resizing.

This patch attempts to provide a way for layouts to define their own mouse resizing function.


lanrat:
What is the next planned Awesome release that will include this patch? I don't see it in the newest 3.5.9.

blueyed:
It will be 3.6.

From the size of the diff I could imagine that there are conflicts when trying to backport it to 3.5.

I suggest to wait (and/or try master already), or provide a PR against the 3.5 branch, so that it would be included in a 3.5 release, which might happen before 3.6.

PR #68, opened 2014-12-29
basic versioning and packaging

akdor1154:
implement basic DEB packaging with CMake. (I don't have an RPM system to test with, but it should be trivial top add that on top of this)
This is very useful for users running from git, as it still allows for easy installation while also having dpkg manage all of awesome's files.

still to do for a full solution:
files intended for /etc doesn't seem to install properly


psychon:
If we start using CMake, I guess our regular releases should be done with CPack, too, right? Right now there is a `make dist` target which runs `build-utils/dist.sh`. This script then creates our tarballs.

akdor1154:
CPack's source packages don't really add anything above the existing dist script, and the existing script will actually yield a cleaner dist package due to git-archive (CPack's documentation recommends a git clone for the same reason). I reckon the existing source package method is worth keeping, but would definitely recommend CPack for all binary package needs.

akdor1154:
arsed this up completely so reopened a new PR: https://github.com/awesomeWM/awesome/pull/79

PR #70, opened 2014-12-31
Add support for notification actions

fbarthelery:
The [desktop notification specification](http://www.galago-project.org/specs/notification/0.9/index.html) says that a notification can have different actions. These actions allow the user to interact with the client application and should be displayed by the notification server.

This pull request : 
  * add DBus function to emit a signal 
  * emit the NotificationClosed signal when a notification is destroyed
  * display notification actions in the notification box
  * emit the ActionInvoked signal when an action is clicked by the user.

### DIFF
--- a/dbus.c
+++ b/dbus.c
@@ -789,6 +789,60 @@ luaA_dbus_disconnect_signal(lua_State *L)
     return 0;
 }
 
+/** Emit a signal on the D-Bus.
+ * \param L The Lua VM state.
+ * \return The number of elements pushed on stack.
+ * \luastack
+ * \lparam A string indicating if we are using system or session bus.
+ * \lparam A string with the dbus path.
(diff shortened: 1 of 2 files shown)


psychon:
Could you base your pull request on the master branch instead of 3.5? "master" is where all the interesting stuff is supposed to happen while 3.5 only exists for stable releases.

psychon:
Thanks a ton for this, it looks great!

If you want to, I can just merge this as-is into awesome (however, into the master branch, not 3.5). If you want to change some of the things that I mentioned above, just say so. I guess otherwise I would just handle these things myself after merging.

fbarthelery:
Thanks for your comments, I will make the changes later today

fbarthelery:
This PR on master should be followed instead :  https://github.com/awesomeWM/awesome/pull/72

PR #72, opened 2015-01-02
Add support for notification actions

fbarthelery:
The [desktop notification specification](http://www.galago-project.org/specs/notification/0.9/index.html) says that a notification can have different actions. These actions allow the user to interact with the client application and should be displayed by the notification server.

This pull request : 
  * add DBus function to emit a signal 
  * emit the NotificationClosed signal when a notification is destroyed
  * display notification actions in the notification box
  * emit the ActionInvoked signal when an action is clicked by the user.

This pull request complete https://github.com/awesomeWM/awesome/pull/70 but on master branch

### DIFF
--- a/dbus.c
+++ b/dbus.c
@@ -812,16 +812,13 @@ luaA_dbus_emit_signal(lua_State *L)
     const char *path = luaL_checkstring(L, 2);
     const char *itface = luaL_checkstring(L, 3);
     const char *name = luaL_checkstring(L, 4);
-    lua_remove(L, 1);
-    lua_remove(L, 1);
-    lua_remove(L, 1);
-    lua_remove(L, 1);
     DBusConnection *dbus_connection = a_dbus_bus_getbyname(bus_name);
     DBusMessage* msg = dbus_message_new_signal(path, itface, name);
 
     DBusMessageIter iter;
     dbus_message_iter_init_append(msg, &iter);
-    int nargs = lua_gettop(L);
+    int top = lua_gettop(L);
+    int nargs = top - 4;
 
     if(nargs % 2 != 0)
     {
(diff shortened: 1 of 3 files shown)


blueyed:
I am using the following hack to disable dbus with naughty, which now fails (in lib/naughty.lua:269: attempt to index field 'dbus' (a nil value):

    local _dbus = dbus; dbus = nil
    local naughty = require("naughty")  -- XXX: if loaded (with dbus available), notifications do not go to dunst/notify-osd anymore
    dbus = _dbus

The reason for this hack is to not use naughty as a listener for desktop notifications, but still for manually displaying them (when called directly).

This probably needs a separate config option then maybe?!

psychon:
Fixed with commit d5088c5f6f7831d19db

blueyed:
@psychon 
Thanks!

psychon:
Sorry that I messed with this after it was merged, but I just split up naughty into naughty.core and naughty.dbus, since things were getting too complicated. However, notify-send doesn't support actions and thus I cannot really test this.

@fbarthelery Could you take a look at current master and figure out if I broke anything in commit 355f10609c4a7902edf5c8c9691bf9d7f32be8ae? (I also did commits 509bbe7230c935aa0fae250fca2685f0b8b7b941 and 239bc02d835c1bdc849d50ac7936702bfc01493b for preparation)

Also, how exactly are you testing this?

And again: Thanks for writing this

fbarthelery:
For testing I use this simple script that i passed to awesome-client.

```
local n=require("naughty")

n.notify({text="hello", title="title", actions={
    "default", "default",
    "first", "premier",
    "sec", "second"}})
```

fbarthelery:
It does not seems to break anything for me. 
My IM client (gajim) send dbus notifications with a default action and it works correctly. althougt I have no soft that sends other actions to test for the other cases.

PR #73, opened 2015-01-06
Displaying only selected tags in taglist

barlik:
Add taglist.filter.selected() for displaying only currently selected tags


barlik:
http://stackoverflow.com/questions/8964778/how-to-retrieve-current-tag-name-in-awesome-wm

This question was asked on stackoverflow. I think that it's more elegant to change filter function.

PR #79, opened 2015-01-19
implement basic DEB/RPM packaging

akdor1154:
continuing on from https://github.com/awesomeWM/awesome/pull/68, which I messed up the history of.

implement basic DEB packaging with CMake. (I don't have an RPM system to test with, but it should be trivial top add that on top of this)
This is very useful for users running from git, as it still allows for easy installation while also having dpkg manage all of awesome's files.

further from that, cpack now knows about awesome's version, so the output files will be named awesome-v3.5.2~git999... instead of awesome-0.1.1.

### DIFF
--- a/Packaging.cmake
+++ b/Packaging.cmake
@@ -1,0 +1,62 @@
+if(NOT VERSION)
+	set(VERSION "unknown")
+endif()
+
+string(REGEX REPLACE "^v?([0-9.]+)-?(.*)$"
+	"\\1;\\2" version_result ${VERSION})
+
+list(LENGTH version_result version_result_list_length)
+
+if(version_result_list_length EQUAL 2)
+	list(GET version_result 0 version_num)
+	list(GET version_result 1 version_gitstamp)
+else(version_result_list_length EQUAL 2)
+	message("Unable to deduce a meaningful version number. \
+Set OVERRIDE_VERSION when you run CMake (cmake .. -DOVERRIDE_VERSION=3.14.159), or \
+just build from a git repository.")
+	set(version_num "0.0.0")
+	set(version_gitstamp "")
+endif(version_result_list_length EQUAL 2)
+
+if(version_gitstamp)
+	set(version_gitsuffix "~git${version_gitstamp}")
+else(version_gitstamp)
+	set(version_gitsuffix "")
+endif(version_gitstamp)
+
+string(REGEX REPLACE "([0-9]+)\\.([0-9]+)\\.([0-9]+)"
+	"\\1;\\2;\\3" version_num_split ${version_num})
+
+list(APPEND version_num_split 0 0 0) #ensure the list(GET )) commands below never fail
+
+list(GET version_num_split 0 CPACK_PACKAGE_VERSION_MAJOR)
+list(GET version_num_split 1 CPACK_PACKAGE_VERSION_MINOR)
+list(GET version_num_split 2 CPACK_PACKAGE_VERSION_PATCH)
+
+set(version_num "${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}")
+
+message(STATUS "Package version will be set to ${version_num}${version_gitsuffix}.")
+
+set(CPACK_PACKAGE_VERSION "${version_num}${version_gitsuffix}")
+
+if(NOT CPACK_GENERATOR)
+	find_program(rpmbuild_path "rpmbuild")
+	if(rpmbuild_path)
+		message(STATUS "rpmbuild found, enabling RPM for the 'package' target")
+		list(APPEND CPACK_GENERATOR RPM)
+	endif(rpmbuild_path)
+
+	find_program(dpkg_path "dpkg")
+	if (dpkg_path)
+		message(STATUS "dpkg found, enabling DEB for the 'package' target")
+		list(APPEND CPACK_GENERATOR DEB)
+	endif(dpkg_path)
+endif(NOT CPACK_GENERATOR)
+
+set(CPACK_PACKAGE_NAME "awesome")
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "devnull@example.com")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "A tiling window manager")
+
+if(CPACK_GENERATOR)
+	include(CPack)
+endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -381,4 +381,6 @@ else()
 endif()
 # }}}
 
+INCLUDE(${CMAKE_SOURCE_DIR}/Packaging.cmake)
+
 # vim: filetype=cmake:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80:foldmethod=marker
--- a/docs/01-readme.md
+++ b/docs/01-readme.md
@@ -12,9 +12,17 @@ After extracting the dist tarball, run:
 
 This will create a build directory, run cmake in it and build awesome.
 
-After building is finished, you can install:
+After building is finished, you can either install via `make install`:
 
     make install  # you might need root permissions
+
+or by auto-generating a .deb or .rpm package, for easy removal later on:
+
+    make package
+    
+    sudo dpkg -i awesome-x.y.z.deb
+    # or
+    sudo rpm -Uvh awesome-x.y.z.rpm
 
 ## Running awesome
 


psychon:
Note that VERSION is only set correctly if git is installed (else it's just "devel") and your regex will also fail for releases (when the tag is just e.g. "v3.6.0"). Right now I don't have any good/easy ideas on how to fix that.

blueyed:
Thanks for your work (and update) on this!
(please note that the build / Travis CI failure is unrelated)

Could this be put in the main CMakeLists.txt file, instead of a separate file?
This would make it more compact IMHO.

Also, can you please add some documentation, about how it should be used - to the instructions/code itself, and then also the install section in README.md / docs/01-readme.md?

How does it compare to the Debian packaging that's used by them?  Could they make use of this somehow?

Elv13:
I think packaging belong to a separate file. CPack code is not that bad, but building and packaging are 2 different concerns

blueyed:
There appears to be a build issue with the latest commit:

    -- Configuring awesome.doxygen

    CMake Error at Packaging.cmake:1 (STRING):

      string sub-command REGEX, mode REPLACE needs at least 6 arguments total to

      command.

      CMakeLists.txt:342 (INCLUDE)

    CMake Error at Packaging.cmake:11 (LIST):

      list index: 1 out of range (-1, 0)

      CMakeLists.txt:342 (INCLUDE)

    CMake Error at Packaging.cmake:12 (LIST):

      list index: 2 out of range (-1, 0)

      CMakeLists.txt:342 (INCLUDE)

    make: *** [cmake] Error 1

    The command "make && sudo make install && awesome --version" exited with 2.

blueyed:
@Elv13 
> I think packaging belong to a separate file. CPack code is not that bad, but building and packaging are 2 different concerns

The `Packaging.cmake` is meant to be called from the main `CMakeLists.txt` only, isn't it?

I consider building a `.deb` / package to be similar to building docs or a `.tgz` etc.

My main concern however is to not unnecessarily "pollute" the main/root directory with `.cmake` files, if it can be just added to the main one.

Elv13:
I think you can do subdirectories and add them to `CMAKE_MODULE_PATH`

akdor1154:
Yeah I put it in a separate file just from seeing how other projects handle
cmake.  A single other file is a little pointless for sure, but as soon as
your cmake gets a bit complex it makes sense to start to split. I have also
seen some projects put all cmakes bar the root in their own directory.
Happy to go with consensus!

psychon:
Any updates on this? This fails to build on travis and the regex that tries to parse the version number cannot ever work.

akdor1154:
It looks like travis is failing because the build script there is forcing the version number to begin with $TRAVIS_BRANCH, which I assume is "master". As far as I can tell Travis is the only situation where this occurs. The regex is working perfectly here and has been for the last six months - as long as $(git describe --dirty) points at a tag with a version number in it, it will be fine.

psychon:
@akdor1154 Please download a tarball from Github and try to build that (= something without a `.git` folder). Alternatively, temporarily deinstall git. In both cases, `git describe` won't be available and the version string is `devel`. So yeah, this needs some "plan B" for when the regex doesn't work.

akdor1154:
Is it OK to just set the package version to 0.0.0 in that case? As far as I can see (you'll know better than me, correct me if I'm wrong!), there is no way to determine a base version in such cases.

psychon:
Yup, fine with me. Anything except just breaking.

akdor1154:
that should be pretty robust now, and also will use OVERRIDE_VERSION if it is supplied to cmake and makes sense (i.e. matches the version regex, which Travis' invocation unfortunately does not). Thus, versions can still be supplied by someone building from a tarball if they desire.

akdor1154:
rpm enabled (on my Ubuntu system with rpm installed, that is -- if anyone is running an rpm-based system it might be worth checking that everything is in order).
I have no further features to add to this, it's grown enough already. :)

Packaging.cmake
```diff
@@ -0,0 +1,64 @@
+if(NOT VERSION)
+	set(VERSION "unknown")
+endif()
```
psychon: Is this part needed? `awesomeConfig.cmake` sets `VERSION` to `devel` if it doesn't find anything better.

Packaging.cmake
```diff
@@ -0,0 +1,64 @@
...
+list(GET version_num_split 0 CPACK_PACKAGE_VERSION_MAJOR)
+list(GET version_num_split 1 CPACK_PACKAGE_VERSION_MINOR)
+list(GET version_num_split 2 CPACK_PACKAGE_VERSION_PATCH)
+
+set(version_num "${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}")
+
+message(STATUS "Package version will be set to ${version_num}${version_gitsuffix}.")
+
+set(CPACK_PACKAGE_VERSION "${version_num}${version_gitsuffix}")
+set(CPACK_DEBIAN_PACKAGE_VERSION "${version_num}${version_gitsuffix}")
```
psychon: Are both of these `set()`s needed? Google (and `cpack --help-variable CPACK_DEBIAN_PACKAGE_VERSION`) suggests that `CPACK_DEBIAN_PACKAGE_VERSION` defaults to `CPACK_PACKAGE_VERSION` if unset.

Packaging.cmake
```diff
@@ -0,0 +1,64 @@
...
+endif(rpmbuild_path)
+
+find_program(dpkg_path "dpkg")
+if (dpkg_path)
+	message(STATUS "dpkg found, enabling DEB for the 'package' target")
+	list(APPEND CPACK_GENERATOR DEB)
+endif(dpkg_path)
+
+set(CPACK_PACKAGE_NAME "awesome")
+set(CPACK_DEBIAN_PACKAGE_ARCHITECTURE "amd64")
```
psychon: Why is this set explicitly?
```
$ cpack --help-variable CPACK_DEBIAN_PACKAGE_ARCHITECTURE
cpack version 2.8.8
  CPACK_DEBIAN_PACKAGE_ARCHITECTURE

            Mandatory : YES
            Default   : Output of dpkg --print-architecture (or i386 if dpkg is not found)
            The debian package architecture
```
(And this suggests that `dpkg` isn't even needed to build `deb`s.)

Packaging.cmake
```diff
@@ -0,0 +1,64 @@
...
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "devnull@example.com")
```
psychon: The following suggests that both (1) CPack has a bug in its documentation and (2) you should set `CPACK_PACKAGE_CONTACT` instead of `CPACK_DEBIAN_PACKAGE_MAINTAINER`.
```
$ cpack --help-variable CPACK_DEBIAN_PACKAGE_MAINTAINER 
cpack version 2.8.8
  CPACK_DEBIAN_PACKAGE_MAINTAINER

            Mandatory : YES
            Default   : CPACK_PACKAGE_CONTACT
            The debian package maintainer

 $ cpack --help-variable CPACK_PACKAGE_CONTACT
cpack version 2.8.8
Argument "CPACK_PACKAGE_CONTACT" to --help-variable is not a defined variable.  Use --help-variable-list to see all defined variables.
```

psychon:
Thanks for the update. This looks mostly fine to me, except for some minor things that I mentioned above.

Could you also squash this into a single commit? Thanks

psychon:
Oh, one more thing: Grepping for `FATAL` in `CPackRPM.cmake` and `CPackDeb.cmake`: While RPM does require `rpmbuild`, DEB apparently doesn't have any dependencies. I'd suggest to drop the check for `dpkg`. If this does indeed have some dependencies, we will learn from bug reports. :-)

Packaging.cmake
akdor1154: You're right, one should suffice.

Packaging.cmake
akdor1154: I'm not sure what's going on in awesomeConfig, but when I removed `git` (via `chmod -x`), `VERSION` was empty and hence the next line was failing with wrong number of arguments. It's possible that my own `chmod -x` leads to different behaviour than actually removing `git` from `$PATH`, though...

psychon: Oh, I think that `cmake` has some kind of variable scoping. Variables are only set locally (in `awesomeConfig.cmake`) and not in the calling "scope" (`CMakeLists.txt`). Never mind then, just leave this as-is.

Packaging.cmake
akdor1154: Wow, you are right. I had in my head that this was explicitly required, sorry; either their behaviour has changed or I have always been wrong... probably the latter. :)

Packaging.cmake
akdor1154: The lack of a global default is intentional -- the maintainer field is required to build a Debian package, but any packages built by this system are not in any way official to the Awesome project and don't really have a maintainer (in the Debian packaging sense) at all. Hence the devnull placeholder address.

akdor1154:
Thanks for the look-over. :) I'll squish and fix tomorrow, heading to bed now, but most of your suggestions make lots of sense. The dpkg check is on purpose, though -- I know that building a deb with CPack doesn't require dpkg, however I wanted to avoid building debs on Joe Fossman's Fedora box by default, as he probably just wants `make package` to build him an rpm. I can change this behvaiour if you like to always build debs; just letting you know there is some logic there. :)

psychon:
Fine with me for now, but perhaps it makes more sense then to have an option that defaults to "is `dpkg` found?". That way I can also build .debs on my Fedora box without installing dpkg. However, feel free to leave it the way it is now.

psychon:
Ping?

blueyed:
Thanks!

blueyed:
Thanks a lot again!

Do you think we should/can provide a Ubuntu PPA with that?
Do you have any experience with that?  (I have been a contributor to Ubuntu's repos  myself)

PR #83, opened 2015-01-20
menubar: handle nil Name in .desktop files

roysc:
Fixes erroring out on gsub() call, which propagated to a nil result from menu_gen.generate()

### DIFF
--- a/lib/menubar/utils.lua.in
+++ b/lib/menubar/utils.lua.in
@@ -177,6 +177,9 @@ function utils.parse(file)
     if program.Exec then
         -- Substitute Exec special codes as specified in
         -- http://standards.freedesktop.org/desktop-entry-spec/1.1/ar01s06.html
+        if program.Name == nil then
+            program.Name = '['.. file:match("([^/]+)%.desktop$") ..']'
+        end
         local cmdline = program.Exec:gsub('%%c', program.Name)
         cmdline = cmdline:gsub('%%[fuFU]', '')
         cmdline = cmdline:gsub('%%k', program.file)


roysc:
Although the vast majority of desktop files have a Name, this would result in nameless menu entries for those that don't, but I'm not sure there's a much better solution.

blueyed:
Using `[no name]` or `[unnamed]` might be better.

Where/how did you came across it?

roysc:
When I tried to open the menubar, it failed with an error about a non 
string argument to `gsub`. So it failed to generate the menu items, and 
after that, the menubar fails with a nil argument to `ipairs` at 
menubar/init.lua:152.

blueyed:
Ok. I was just curious about which desktop file might have caused it.

I think using the basename of `file`, with `.desktop` removed, in brackets, might be a good fallback.

So for /foo/bar/awesome.desktop without a Name, it would use `[awesome]`.

roysc:
Ah, it was actually a program that wasn't properly packaged anyway, but 
I figured it's a case that should be handled.

That makes sense, will update the PR.

roysc:
Updated.

blueyed:
:+1:

iffer:
How comes this patch has been it into the master-branch for a year and didn't make it into 3.5?

psychon:
Because it didn't make it into the 3.5 branch yet. If it applies, I guess it should be cherry-picked.

blueyed:
Cherry-picked in dd40e7a021dfdc3f69bf029321b7eb5fd32ac7a0.
Thanks!

psychon:
Sorry that I forgot to mention the cherry-pick here and thanks to @blueyed for fixing that oversight :)

Issue #84, opened 2015-01-20
Menubar cannot handle special pattern characters


roysc:
Entering a "magic character" for Lua patterns in the menubar causes it to freeze and issue an error, e.g., after entering only a "%":
```
W: awesome: event_handle_key:586: error running function: /usr/share/awesome/lib/menubar/init.lua:140: malformed pattern (ends with '%')
```
The menubar then can't be closed without reloading Awesome.

roysc:
Problem characters: `% [ ( )`

blueyed:
Thanks for the report!

See my PR for a (possibly rough) fix.

PR #85, opened 2015-01-21
WIP: menubar: quote special chars in `nocase`

blueyed:
This needs to get improved probably, in the sense that the quote
function should get moved to some utils module?!

Fixes https://github.com/awesomeWM/awesome/issues/84

### DIFF
--- a/lib/menubar/init.lua.in
+++ b/lib/menubar/init.lua.in
@@ -79,11 +79,19 @@ local function colortext(s, c)
     return "<span color='" .. c .. "'>" .. s .. "</span>"
 end
 
+--- Quote special chars in a pattern for gsub.
+-- Source: http://stackoverflow.com/a/20778724/15690
+local quotepattern = '(['..("%^$().[]*+-?"):gsub("(.)", "%%%1")..'])'
+string.quote = function(str)
+    return str:gsub(quotepattern, "%%%1")
+end
+
 -- Generate a pattern matching expression that ignores case.
--- @param s Original pattern matching expresion.
+-- @param s Original pattern matching expression.
 local function nocase (s)
-    s = string.gsub(s, "%a",
+    s = string.gsub(s, ".",
                     function (c)
+                        c = string.quote(c)
                         return string.format("[%s%s]", string.lower(c),
                                              string.upper(c))
                     end)


psychon:
There's a little bit too much of "WIP" in these descriptions for me to pull things, sorry. Also, I don't really understand these patches...

blueyed:
> There's a little bit too much of "WIP" in these descriptions for me to pull things, sorry.

That was the idea behind it.. it should not get pulled yet.. ;)

But @roysc improved my initial patch (via a PR on my fork), and I think it's good now.

Nevertheless, a review from you would be still appreciated of course.

> Also, I don't really understand these patches...

But I assume you understand the final diff/patch?! https://github.com/awesomeWM/awesome/pull/85/files

If it looks OK to you, I could clean it up and refresh the PR, or you could merge/cherry-pick/squash it from blueyed:menubar-quote-string yourself.
I can also merge and push it myself, when it looks sensible to you.

When squashing it, it should get credited to @roysc.

psychon:
Ok, so:

     local quotepattern = '['..("%^$().[]*+-?"):gsub("(.)", "%%%1")..']'

Took me a moment to understand this. I'm not sure if this is really easier to understand, but howe about:

    local quotepattern = '[' .. "%" .. table.concat({"%", "^", "$", "(", ")", ".", "[", "]", "*", "+", "-", "?"}, "%") .. ']'

Or, just for the sake of readability:

    local quotepattern = "["
    for _, char in pairs({"%", "^", "$", "(", ")", ".", "[", "]", "*", "+", "-", "?"}) do
        quotepattern = quotepattern .. "%" .. char
    end
   quotepattern = quotepattern .. "]"

Also, having this outside of query_to_pattern() is just a micro-optimizations. Setting this up inside of that function could make it clearer what this is for. Then you could also add some comments like "Escape all special pattern-matching characters so that lua interprets them literally instead of as a character class" for the first gsub and "Poor man's case-insensitive character matching" for the second?

Also, why does this even work? For a string like "%s" it will produce "[%%][%%][sS]", I guess. The first gsub call in your patch produces "%%s", the second one then does "that". The resulting pattern would match "%%s", but wouldn't match "%s". Or do I miss something...?

An idea on how to implement this (And hopefully this also doesn't suffer from the above bug (if it is a bug, I'm not sure....)):

    local chars_to_quote = {"%", "^", "$", "(", ")", ".", "[", "]", "*", "+", "-", "?"}
    s = string.gsub(s, ".", function(c)
        local lower, upper = string.lower(c), string.upper(c)
        if awful.util.table.hasitem(chars_to_quote, c) then
            return "[%" .. lower .. "][%" .. upper .. "]"
        end
        return "[" .. lower .. "][" .. upper .. "]"
    end

If I may make a wish:

Make this escape-query-function somehow available publicly (Move it to awful.util? Dunno) and write some unit tests for it. I guess those tests shouldn't test that the expected patterns are generated, but that the patterns really match what they should match.

blueyed:
> Make this escape-query-function somehow available publicly (Move it to awful.util? Dunno)

Agreed. That's one of the reasons for it being "WIP". It seems to required in several places, and Lua appears to make it rather hard to handle it.

@roysc 
Do you feel like incorporating / implementing the feedback provided above? That would be awesome.. :)

psychon:
Oh and UTF8 says hi:

    $ lua -e 'print(string.gsub("ä", "%a", function(c) return "[" .. string.lower(c) .. string.upper(c) .. "]" end))'
    ä	0

(Old version doesn't produce "[äÄ]")

    $ lua -e 'print(string.gsub("ä", ".", function(c) return "[" .. string.lower(c) .. string.upper(c) .. "]" end))' 
    [ÃÃ][¤¤]	2

(New version totally breaks multi-byte sequences)

roysc:
Actually the double `%` was an oversight, fixed by using `%a` rather than `.` as it was originally. This way we `gsub` only letters.

I don't have a strong feeling on the syntax of the char list, but I will refactor a bit and clean up.

roysc:
@psychon Lua doesn't have any built in Unicode support AFAICT, so I think that old version is the best we can do.

roysc:
As for unit tests, I wrote something _very_ basic but there aren't any other tests I see in the entire code base. I have just been dog-fooding my code.

roysc:
One a different note, I'm not sure how to best introduce my changes. Is it standard to just open a PR on @blueyed 's branch for this PR (what I did last time)?

Elv13:
Lua 5.3 has utf support

blueyed:
> Is it standard to just open a PR on @blueyed 's branch for this PR (what I did last time)?

No. Please create a PR against awesomeWM/awesome's master.

blueyed:
Closing this in favor of #90.
Everything from here appears to have been addressed in the new PR.

psychon:
@roysc I fixed up the unit tests in commit b4d677797a95

PR #86, opened 2015-01-21
WIP: completion: quote strings passed to the shell

blueyed:
The strings passed to the shell completion machinery need to be quoted.

I've patched this a while ago and only tested it with Zsh.


psychon:
Why is this a WIP? I don't have much clue about this magic and cannot really tell what you'd need help with.

blueyed:
> Why is this a WIP?

I've only tested it with my shell (zsh).

I've gone ahead and cherry-picked the first commit with the fix (925bdbc) into master: 704cf21.

To test this try completion after e.g. `ls '`, which should trigger the bug.

PR #87, opened 2015-01-21
awful.rules.execute: use delayed_call for emitting request::activate

blueyed:
This is a followup to commit 90fde13, which caused the client not to be
visible during the request::activate callback.

This requires an initial call to `awesome_refresh` before the main loop
to get `awesome.startup` right.

Ref: https://github.com/awesomeWM/awesome/commit/90fde139#commitcomment-9308225

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -545,6 +545,10 @@ main(int argc, char **argv)
     g_main_context_set_poll_func(g_main_context_default(), &a_glib_poll);
     gettimeofday(&last_wakeup, NULL);
 
+    /* Do all deferred work now once outside of the loop to get awesome.startup
+     * right. */
+    awesome_refresh();
+
     /* main event loop */
     globalconf.loop = g_main_loop_new(NULL, FALSE);
     g_main_loop_run(globalconf.loop);
--- a/common/luaobject.c
+++ b/common/luaobject.c
@@ -253,9 +253,14 @@ luaA_object_emit_signal(lua_State *L, int oud,
                         const char *name, int nargs)
 {
     int oud_abs = luaA_absindex(L, oud);
-    lua_object_t *obj = lua_touserdata(L, oud);
+    lua_class_t *lua_class = luaA_class_get(L, oud);
+    lua_object_t *obj = luaA_toudata(L, oud, lua_class);
     if(!obj) {
         warn("Trying to emit signal '%s' on non-object", name);
+        return;
+    }
+    else if(lua_class->checker && !lua_class->checker(obj)) {
+        warn("Trying to emit signal '%s' on invalid object", name);
         return;
     }
     signal_t *sigfound = signal_array_getbyid(&obj->signals,
--- a/lib/awful/rules.lua.in
+++ b/lib/awful/rules.lua.in
@@ -12,6 +12,7 @@ local ipairs = ipairs
 local pairs = pairs
 local aclient = require("awful.client")
 local atag = require("awful.tag")
+local timer = require("gears.timer")
 
 --- Apply rules to clients at startup.
 -- awful.rules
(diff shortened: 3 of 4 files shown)


actionless:
works for me

blueyed:
I am seeing an error when opening an attachment from an email in Thunderbird:

    stack traceback:
      [C]: in function 'isvisible'
      /home/user/.config/awesome/rc.lua:1648: in function </home/user/.config/awesome/rc.lua:1638>
      [C]: in function 'emit_signal'
      /usr/local/share/awesome/lib/awful/rules.lua:239: in function </usr/local/share/awesome/lib/awful/rules.lua:238>
      [C]: in function 'pcall'
      /usr/local/share/awesome/lib/gears/timer.lua:103: in function </usr/local/share/awesome/lib/gears/timer.lua:101>
    error: /home/user/.config/awesome/rc.lua:1648: invalid object

The client before `timer.delayed_call` has no title:

    client : window/client(): 0x2bc04d8 (client)

After that there's another error from tag.withcurrent:

    /usr/local/share/awesome/lib/awful/tag.lua:580: invalid object

Only after that the real client appears:

    client : window/client(Image Viewer): 0x2933688 (client)

The problem appears to be that the "client" is alive for a short period only
(it was named "Saved Files" once), and gets unmanaged before the delayed calls
gets triggered.

This happens when there's a default action for the attachment in Thunderbird,
and the "Opening ..." dialog gets skipped.

blueyed:
I wonder why the error from this PR's function gets displayed through naughty and with a stacktrace in the log / stderr, but the one from `tag.withcurrent` is only printed?

blueyed:
> I wonder why the error from this PR's function gets displayed through naughty

That's probably because the error comes from rc.lua vs. from internals?!

psychon:
Does commit 254e50d88c5f58cb8b736dbb4681d020ddb88919 obsolete this or is there still something left to be fixed? If no, this can be closed...?

blueyed:
For reference: see the discussion at https://github.com/awesomeWM/awesome/commit/90fde1393f697ca0d7201e385293a5ac4778d136#commitcomment-9308225.

I think that using `timer.delayed_call` is a good idea in general, and might also help for issues like [FS#1306 - Postpone/block layout (re)arrangement during applying of awful.rules / manage signal](https://awesome.naquadah.org/bugs/index.php?do=details&task_id=1306&project=1&order=id&sort=desc).

blueyed:
I've rebased this on master, and added the revert of 98aeffa.

This works better in general, e.g. for the use case of having one tiled client on a tag and opening a new floating client (which gets set to floating via rules): with this PR there will be no flickering, but without it, the client gets managed as a tiled client first, before it gets set to be floating.

blueyed:
I've added a check for a valid object to `luaA_object_emit_signal`, which solves the problem mentioned in my first comment.

I still think this should get merged, and would appreciate feedback on it.

aroig:
I've just realized one annoying change in behaviour with this pull request.

New clients do not get focus! The previously focused client keeps the focus. I'm not sure now if it is a problem with the pull request or with my configuration, but it is a change in behaviour nonetheless.

aroig:
Ok, more concretely, What changed is that I can no longer set focus to new clients by hooking to the manage signal. It works if I do it via rules.

blueyed:
@aroig 
What are you doing exactly? Preventing focus for some clients?
Are you using the "focus" property or function/callback?
Do you detach (and then re-attach) the default manage signal handler? (to get your handler in front of it)?

blueyed:
@aroig 
Hmm, looking at https://github.com/awesomeWM/awesome/pull/87/files#diff-7fd492502362f822088c030c8af688caL236 I can imagine what might happen..

Some rule says that a client should be focused (and a delayed call to raise the client gets queued).
You then change the focus manually later, and that gets overridden.

blueyed:
@aroig 
See #117, which hopefully addresses this - but then is still rough with regards to config adjustments.

Do you use an own function for the `request::activate` signal (instead and/or on top of `ewmh.activate`)?

blueyed:
@aroig 
See #118 instead.

PR #89, opened 2015-01-24
Make textbox widget respect the current theme

dunric:
Without this fix `wibox.widget.textbox` ignores current theme font setting and resets to the initial one.

It becomes handy when initial theme is tweaked during runtime, ie. in *rc.lua* . Other Awesome WM components like `awful` (taglist, tasklist, menu) are aware and do respect actual theme settings.

### DIFF
--- a/lib/wibox/widget/textbox.lua.in
+++ b/lib/wibox/widget/textbox.lua.in
@@ -136,7 +136,7 @@ local function new(text, ignore_markup)
     ret:set_wrap("word_char")
     ret:set_valign("center")
     ret:set_align("left")
-    ret:set_font(theme and theme.font)
+    ret:set_font(beautiful and beautiful.font)
 
     if text then
         if ignore_markup then
--- a/lib/wibox/widget/textbox.lua.in
+++ b/lib/wibox/widget/textbox.lua.in
@@ -7,7 +7,6 @@
 
 local base = require("wibox.widget.base")
 local beautiful = require("beautiful")
-local theme = beautiful.get()
 local lgi = require("lgi")
 local cairo = lgi.cairo
 local Pango = lgi.Pango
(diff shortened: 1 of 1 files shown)


psychon:
Sorry, but I'm not totally convinced.

For example, I have a MPD widget which displayed the currently playing song in a textbox. The code creates this textbox when it is `require`'d. However, this is before `beautiful.init` is called from `rc.lua` and thus before the theme is loaded.
Figuring out why some textbox suddenly uses a different font than all the other ones will be weird.

We didn't have this problem in 3.4, because there was only a single, global font setting (`awesome.font`).

Hm...

A small experiment shows that a PangoLayout has a FontDescription of `nil` if none is set:

    $ lua -e 'lgi = require("lgi") ; pango = lgi.Pango; pc = lgi.PangoCairo ; ctx = pc.font_map_get_default():create_context() ; lay = pango.Layout.new(ctx) ; print(lay:get_font_description())'
    nil

I would thus propose to handle this in `setup_layout()` (untested!):

    local function setup_layout(box, width, height)
        local layout = box._layout
        if layout:get_font_description() == nil and beautiful.font then
            box:set_font(beautiful.font)
        end
        layout.width = Pango.units_from_double(width)
        layout.height = Pango.units_from_double(height)
    end

What do you think? Can you replace your commit with this?

dunric:
@psychon Thanks for the response.
The idea to set font in `setup_layout` is interesting, however be aware this function is called repeatedly from various place like `:draw()` and `:fit()` so anything you'd put there will slow things down. I'd suggest to leave `set_font` call in `new` when widget is created and be able to change it explicitly with `textbox:set_font`.

blueyed:
Makes sense to me.
Should be squashed/rebased.

blueyed:
Merged in https://github.com/awesomeWM/awesome/commit/e39504a30f34654d53a8b97cc68ea92effce78c3.

Sorry @dunric, I've apparently messed up when trying to cherry-pick/am this commit, because your authorship got lost.. :(

PR #90, opened 2015-01-25
Quote special characters in user query patterns

roysc:
This should properly handle all special pattern characters.

Fixes https://github.com/awesomeWM/awesome/issues/84

Discussed in https://github.com/awesomeWM/awesome/pull/85

### DIFF
--- a/lib/awful/util.lua.in
+++ b/lib/awful/util.lua.in
@@ -478,6 +478,28 @@ function util.table.iterate(t, filter, start)
     end
 end
 
+-- Escape all special pattern-matching characters so that lua interprets them
+-- literally instead of as a character class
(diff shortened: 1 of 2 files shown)


blueyed:
:+1: 
Thanks for your work on this!

I think it's good to be merged, but should get squashed into a single commit.

@roysc 
Can you do this? Otherwise this can be done when merging, too.
Just let us know, if you prefer to do this yourself and/or if you'd like to keep the history as-is instead.

If nobody intervenes I'm going to merge this later today.

blueyed:
Committed in https://github.com/awesomeWM/awesome/commit/02bfd06038a61d14a4c31576c62105aa174e2181.

roysc:
Sorry, lost track of this with school starting, thanks for handling it.

PR #96, opened 2015-01-31
Emit tag::history::update signal from tag.history.restore

blueyed:
Restoring a previous history entry changes the tag history (in the
meaning of currently selected tags), and therefore should also emit the
tag::history::update signal.

A use case might be to automatically hide tags with no clients on them.

Signed-off-by: Daniel Hahler <git@thequod.de>


psychon:
Fine with me, the commit message makes sense.

PR #97, opened 2015-02-02
Add support for setting client.name

blueyed:
It is basically a copy of / derived from luaA_tag_set_name.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1898,6 +1898,19 @@ luaA_client_get_name(lua_State *L, client_t *c)
     return 1;
 }
 
+/** Set the client name.
+ * \param L The Lua VM state.
+ * \param client The client to name.
+ * \return The number of elements pushed on stack.
+ */
+static int
+luaA_client_set_name(lua_State *L, client_t *c)
+{
+    const char *name = luaL_checkstring(L, -1);
+    client_set_name(L, 1, a_strdup(name));
+    return 0;
+}
+
 static int
 luaA_client_get_icon_name(lua_State *L, client_t *c)
 {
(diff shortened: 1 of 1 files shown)


objects/client.c
```diff
@@ -1899,6 +1899,20 @@ luaA_client_get_name(lua_State *L, client_t *c)
...
+    size_t len;
+    const char *buf = luaL_checklstring(L, -1, &len);
+    char *name;
+    if (a_iso2utf8(buf, len, &name, NULL)) {
+        client_set_name(L, -3, name);
+        luaA_object_emit_signal(L, -3, "property::name", 0);
+        return 0;
+    }
+    return 1;
+}
```
blueyed: This probably needs improvements - I've derived it from `luaA_tag_set_name`.

luaa.h
```diff
@@ -122,6 +122,14 @@ luaA_checkboolean(lua_State *L, int n)
     return lua_toboolean(L, n);
 }
 
+static inline const char *
+luaA_checkstring(lua_State *L, int n)
```
blueyed: I am not sure if this is required, because there wasn't such a function already..

If it makes sense it's probably better to not have it `inline`, because this triggers compiling almost all `.o` files?!

psychon:
What's the use case for this? Right now I don't really see a point, because the client can still changes its own name later on and overwrite lua's decision. Depending on the use case, it might perhaps make more sense to have a lua-only implementation (A "visible_name" property in `awful.client.property`?)

blueyed:
> What's the use case for this? Right now I don't really see a point, because the client can still changes its own name later on and overwrite lua's decision.

That's fine. It's like setting the terminal's title, but it can be changed later, too.

> Depending on the use case, it might perhaps make more sense to have a lua-only implementation (A "visible_name" property in awful.client.property?)

That's a different use case, and it should be implemented separately / additionally in case it's considered to be useful.  I think that this is something so special that you could just implement it yourself via a custom client property.

psychon:
Hm, ok. If this really does what you want it to do, do so...

PR #98, opened 2015-02-02
request::activate with client.movetoscreen

blueyed:
@Elv13 
Is using the context with the signal here done correctly / as you would expect it?

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -542,8 +542,15 @@ function client.movetoscreen(c, s)
             s = sel.screen + 1
         end
         if s > sc then s = 1 elseif s < 1 then s = sc end
-        sel.screen = s
-        screen.focus(s)
+        if sel.screen ~= s then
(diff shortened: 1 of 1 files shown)


Elv13:
thanks

I like the movetoscreen idea, this will be useful, I may merge this later on. For now I still need to review some of the earlier PR

blueyed:
I think this should be improved to check if the client is currently focused before changing the screen / as early as possible.
Otherwise the focus might have changed already, e.g. because of autofocus etc.

blueyed:
I've moved the second commit (for client.swap.global_bydirection) into a separate PR: https://github.com/awesomeWM/awesome/pull/158

blueyed:
I've amended the previous commit, and fixed it to only send the signal when the screen property gets changed.

This is untested.

blueyed:
Rebased it on master.

psychon:
Fine by me

PR #102, opened 2015-02-08
WIP: Persistent tag properties

blueyed:
This brings persistent properties for tags, via X properties on the root window (via `awesome.set_xproperty` etc).

This has been discussed / sketched in https://github.com/awesomeWM/awesome/pull/29 before.

### DIFF
--- a/lib/awful/layout/init.lua.in
+++ b/lib/awful/layout/init.lua.in
@@ -58,6 +58,16 @@ function layout.get(screen)
     return tag.getproperty(t, "layout") or layout.suit.floating
 end
 
+--- Get a layout by name.
+-- @return The layout function.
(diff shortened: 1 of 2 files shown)


blueyed:
I have some local stash for this still.

I'd appreciate some feedback on it, despite that it's only a workaround for awesome restarting on screen changes (or when restarting it manually during development).
As with persistent client properties I think this is good to have in general.

Elv13:
Should this be closed now that Awesome wont restart when adding screens? If the only use case is when developing, then I propose using Xephyr. Anyhow, it wont work in Wayland and it wont ever cover all corner cases of weird stuff you can store in a tag.

Elv13:
Closing, this seems abandoned

PR #103, opened 2015-02-08
Add `has_xproperty`: no error for unregistered props

blueyed:
I came up with this while doing the persistent tag properties, but then
never actually used it there.

Is this useful to have?
I could also imagine having an optional flag with `get_xproperty` that
wouldn't cause an error.

### DIFF
--- a/objects/window.c
+++ b/objects/window.c
@@ -385,6 +385,20 @@ window_get_xproperty(lua_State *L, xcb_window_t window, int prop_idx)
     return 1;
 }
 
+int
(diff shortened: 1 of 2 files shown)


psychon:
I would think that any given xproperty should only ever be used by a single piece of code. So that piece of code should now if it already registered the property or not. I can't really see a use case for "let's see if some other lua module happened to register a property".

blueyed:
> I would think that any given xproperty should only ever be used by a single piece of code. So that piece of code should now if it already registered the property or not. I can't really see a use case for "let's see if some other lua module happened to register a property".

Ok and agreed. Closing it then.

PR #104, opened 2015-02-09
awful.util.file_readable: return false for dirs

blueyed:
This is currently only used for icon lookups / where files are expected.

### DIFF
--- a/lib/awful/util.lua.in
+++ b/lib/awful/util.lua.in
@@ -207,13 +207,18 @@ function util.geticonpath(iconname, exts, dirs, size)
     end
 end
 
---- Check if file exists and is readable.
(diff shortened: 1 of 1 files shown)


Elv13:
I don't really like the fact this is a synchronous function, but as it already exist, I guess the extra read blocking call wont make that much of a difference. Unless someone have an objection, I will merge this

blueyed:
While looking into this I've came across LuaFileSystem (https://keplerproject.github.io/luafilesystem/), which probably would provide a better way for this.

Would this be a good / feasible dependency?

blueyed:
Ok, as discussed on IRC, this can be merged as is.

PR #105, opened 2015-02-09
awful.util.deprecate: display deprecations only once

blueyed:
This remembers displayed tracebacks and skips them if they were
displayed already.

### DIFF
--- a/lib/awful/util.lua.in
+++ b/lib/awful/util.lua.in
@@ -30,13 +30,21 @@ util.table = {}
 
 util.shell = os.getenv("SHELL") or "/bin/sh"
 
+local displayed_deprecations = {}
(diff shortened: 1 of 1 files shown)


Elv13:
From IRC:

```
[17:53] <Elv1313> blueyed: about #105, are you sure storing the full backtrace is a good idea? what about https://developer.gnome.org/glib/stable/glib-Strings.html#g-string-hash
[17:55] <Elv1313> given the hash already need to be generated for storage in the array, why not only storing the hash
```

blueyed:
Tried that, waiting on https://github.com/pavouk/lgi/issues/100 now.

blueyed:
@Elv13 
The workaround (`glib.String():append(tb):hash()`) works for me.
Is it sensible?

PR #106, opened 2015-02-09
Print debug.traceback() with imagebox:set_image, when reading image failed

blueyed:
This made it easier to figure out where a file causing an error was
coming from, but I could imagine that this could become too noisy, and
that there is/should probably be a more streamlined way to enable more
verbose logging / error output.


Elv13:
I think it is a good idea, an error is an error, so better always display the stacktrace when one happen. Mind if I merge this right away?

blueyed:
:+1:

PR #112, opened 2015-02-12
TBD: Add luaA_object_tostring_idx

blueyed:
This is meant to be used when the object is not at the top of the stack.

I wanted to use the objects string representation for debugging (in
luaA_object_emit_signal), and did not manage to manipulate the stack to
make the original `luaA_object_tostring` work.

Feedback on improving / renaming etc this would be appreciated!

This is useful / needed for #129.

### DIFF
--- a/common/luaobject.c
+++ b/common/luaobject.c
@@ -313,10 +313,10 @@ luaA_object_emit_signal_simple(lua_State *L)
 }
 
 int
-luaA_object_tostring(lua_State *L)
+luaA_object_tostring_idx(lua_State *L, int oud)
 {
-    lua_class_t *lua_class = luaA_class_get(L, 1);
-    lua_object_t *object = luaA_checkudata(L, 1, lua_class);
+    lua_class_t *lua_class = luaA_class_get(L, oud);
+    lua_object_t *object = luaA_checkudata(L, oud, lua_class);
     int offset = 0;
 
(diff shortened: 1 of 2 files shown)


psychon:
Uhm, doesn't plain `lua_tostring(L, idx)` work? At least `luaA_dumpstack(L)` (which is quite handy!) uses this. Lua will then call `luaA_object_tostring()` for you.

blueyed:
`luaA_dumpstack` only uses `lua_tostring` for strings?

For a client it would use the default case:

            fprintf(stderr, "%d: %s\t#%d\t%p\n", i, lua_typename(L, t),
                    (int) luaA_rawlen(L, i),
                    lua_topointer(L, i));

..and that does not appear to use the __tostring method?!

psychon:
Oh, you are right. I never noticed that. Anyway, I still claim that plain `lua_tostring` should work and you don't need to call `luaA_object_tostring` directly.

blueyed:
As noted on IRC it does not, but returns "(null)".

blueyed:
I've been told on IRC:

    > furq | blueyed: lua 5.2 has luaL_tolstring which respects __tostring

Have not check it yet.

psychon:
Fixed by https://github.com/awesomeWM/awesome/commit/5f6bbdfc63d2b74af3d7a7bff0257526a5d2dd8b

PR #113, opened 2015-02-12
Proof of concept: do layout.arrange only once per main loop

blueyed:
This is an idea that could be integrated into #101 ("DRY off layout code"), too.

It uses `timer.delayed_call`.

The idea is to not arrange the layouts multiple times per loop iteration, but only once.


psychon:
Fine by me

blueyed:
Committed in https://github.com/awesomeWM/awesome/commit/74106462ac1f0877d4dd3791c9a983283c034169.

PR #114, opened 2015-02-12
Fully revert  90fde1393f

aroig:
The partial revert of  90fde1393f in commit 254e50d88c5f58 was not enough. It broke floating clients, since they get tiled for an instant, and the geometry gets messed up. The easiest solution is to fully revert 90fde1393f.

### DIFF
--- a/lib/awful/rules.lua.in
+++ b/lib/awful/rules.lua.in
@@ -224,6 +224,11 @@ function rules.execute(c, props, callbacks)
         end
     end
 
+    -- If untagged, stick the client on the current one.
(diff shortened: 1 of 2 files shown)


blueyed:
Please see https://github.com/awesomeWM/awesome/pull/87 instead.

If you can confirm that it fixes your issue, I'm going to finally merge this.

aroig:
Yes #87 fixes my issue!

blueyed:
Ok, I am closing your PR in favor of that one then.

Thanks for your report and contribution nonetheless, and sorry for theses issues in master currently.

PR #115, opened 2015-02-13
Use glib.idle_add(glib.PRIORITY_DEFAULT_IDLE, ..) for delayed calls

blueyed:
Ref: https://github.com/awesomeWM/awesome/commit/90fde13#commitcomment-9308957
Ref: https://developer.gnome.org/glib/2.30/glib-The-Main-Event-Loop.html#g-idle-add

### DIFF
--- a/lib/gears/timer.lua.in
+++ b/lib/gears/timer.lua.in
@@ -102,23 +102,24 @@ timer.new = function(args)
     return ret
 end
 
-local delayed_calls = {}
-capi.awesome.connect_signal("refresh", function()
-    for _, callback in ipairs(delayed_calls) do
-        local success, message = pcall(unpack(callback))
-        if not success then
-            print(message)
-        end
-    end
-    delayed_calls = {}
-end)
-
---- Call the given function at the end of the current main loop iteration
+--- Call the given function at the end of the current main loop iteration.
+-- The function will be called once by default.  You can return true to be
+-- called again.
 -- @tparam function callback The function that should be called
 -- @param ... Arguments to the callback function
 function timer.delayed_call(callback, ...)
     assert(type(callback) == "function", "callback must be a function, got: " .. type(callback))
-    table.insert(delayed_calls, { callback, ... })
+
+    local callback = { callback, ... }
+    glib.idle_add(glib.PRIORITY_DEFAULT_IDLE, function()
+        local success, message = pcall(unpack(callback))
+        if not success then
+            print("Error in delayed_call: " .. tostring(message))
+        elseif message ~= nil then
+            return message  -- Allow to control being called again.
+        end
+        return false  -- Only call it once.
+    end)
 end
 
 function timer.mt:__call(...)


psychon:
I tried this a while back locally and it felt like awesome's redraws look "worse". Before this change "things worked fine", but after this e.g. opening a menu first drew the border and only a split moment later did it draw the content of the menu. Do you see something like this, too?

I didn't investigate too closely, but the "regular" main loop calls `xcb_flush` right before going to sleep, sending out all pending X11 requests. My guess for this new behavior is:
Previously we had this order: do-work, emit "refresh" (delayed calls are run), xcb_flush (this sends out the "make window visible" and "draw window" X11 requests together in one batch), call poll() with some timeout
With this change we would have this order: do-work, emit "refresh", xcb_flush(), poll() with timeout zero, the idle signal triggers (delayed calls are run), xcb_flush only at end of next main loop.

In the later case the requests aren't sent together.

(But it's possible that I am just imagining this! Please double-check)

blueyed:
I could not really see a difference.
But maybe there's a way to exaggerate any side effects (by adding some extra delay somehow)?

Aren't the (new) delayed calls run at the end of `a_glib_poll`, and therefore similar to the behavior before, where `awesome_refresh` (with `luaA_emit_refresh` at the beginning) is run at the beginning of `a_glib_poll`?

psychon:
Perhaps we (I?) should also move the "stuff" the C code is doing into an idle handler instead of "stealing" the call to `g_main_poll()`. Pull request incoming....

psychon:
Since commit 74106462ac1f0877d4dd3791c9a983283c034169 this will cause new clients to become visible before the layout ran and arranged them. (At the end of the main loop iteration awesome unbans clients, the delayed_call into the layout will only occur in the next main loop iteration)

blueyed:
Just rebased this on master..

@psychon 
I do not fully understand your last comment.. do you mean that this delays things for an additional loop, where it's not delayed / handled then anymore?
Is it just a note to be aware of, or something bad?

psychon:
Right now:

- A client appears
- Layout starts a delayed call to do arrange
- "refresh" signal is emited
- Layout is run and client is positioned
- Main loop iteration ends and the C code makes the new client visible (at the correct size)

After this change:

- A client appears
- Layout starts a delayed call to do arrange
- Main loop iteration ends and the C code makes the new client visible (at the wrong position!)
- Idle signal fires, delayed calls are run
- Layout is run and client is positioned correctly (but only after being visible already)

This has visible effects since the client will first redraw themselves for the "wrong" size and then redraw again for the new, correct size. The "wrong" redraw is briefly visible

blueyed:
Can this be fixed by changing the order of:

 - Main loop iteration ends and the C code makes the new client visible (at the wrong position!)
 - Idle signal fires, delayed calls are run

?

psychon:
Not really. And I don't see why we should try to move everything into idle signals.

The problem with making the C code e.g. use `g_idle_add`, too, is that the last thing that `awesome_refresh` does is to call `xcb_flush`. This means "send all X11 requests to the X server now". Flushing the queue needs to be done pretty much all the time and we get no notification of when it is needed. Since `g_idle_add` causes wakeups, we do not know when no `xcb_flush` will be needed anymore.
So I guess the current hack to hijack `g_poll` is the only place where we can do this flush. And flushing requests really should be the last thing we do, after everything else is done.

blueyed:
What do you suggest then?  Is this PR/approach void?

psychon:
What problem is this PR even trying to self? The "refresh" signal works fine as-is. The only doubts where "reinvent the glib.idle_add wheel" and "slow". For (1) we have figured out that we can't easily switch over, so it's not really a "reimplement the wheel". For (2) we just shouldn't use `delayed_calls` too much. ;-)

blueyed:
Ok, fine with me.

PR #118, opened 2015-02-14
Properly fix activating an "invisible" client via rules

blueyed:
This reverts PR #87, except for f562b4a ("luaA_object_emit_signal: check
for valid object"), and allows focusing "invisible" clients (not attached to a tag yet) instead.

This is meant to finally fix the results of delaying the call to `tag.withcurrent` (https://github.com/awesomeWM/awesome/commit/90fde1393f697ca0d7201e385293a5ac4778d136).

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -545,10 +545,6 @@ main(int argc, char **argv)
     g_main_context_set_poll_func(g_main_context_default(), &a_glib_poll);
     gettimeofday(&last_wakeup, NULL);
 
-    /* Do all deferred work now once outside of the loop to get awesome.startup
-     * right. */
-    awesome_refresh();
-
     /* main event loop */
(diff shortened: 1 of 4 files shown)


psychon:
:+1:

aroig:
Hi, sorry for the delay in answring...

I've tested this now from current master. Now setting the focus works fine from the manage signal handler, but there is still a problem.

The first client I open does not get focus. Afterwards everything works as expected. I'm not sure what's going on...

aroig:
To clarify, I meant that everytime I open a client into an empty tag, it fails to set the focus, but when the tag already has some client, it sets the focus alright to the new client.

aroig:
I investigated my problem a bit, and open a new issue #134 with the data I found.

PR #122, opened 2015-02-14
Fix doxygen settings / warnings

blueyed:


psychon:
What? We have a Doxyfile laying around? Wow. Change looks good: :+1:

PR #125, opened 2015-02-14
awful.placement.no_offscreen: no negative x/y

blueyed:
If a client's width/height is larger than the screen's geometry, align
it at the screen's x/y offset.

Test case with gvim, from a maximized terminal on a floating tag:

    % gvim -u NONE -N --cmd "set lines=$((LINES + 5))"

Without this patch the top of the window is off-screen, but the bottom
should be.  Fixing it using a number of lines that fit the screen should
display the window completely.


psychon:
Fine by me (and a quite easy patch)

PR #126, opened 2015-02-14
layout.arrange: delay a call per screen

blueyed:
This is a followup / fix for 7410646, which did not handle multiple
arrange calls to different screens per main lopp.


psychon:
Whoops. I guess I should have spotted that. I have some minor stylistic nitpick, but besides that: :+1:

blueyed:
Committed in 1777605.

PR #129, opened 2015-02-15
WIP: debugging signals

blueyed:
This is a work-in-progress to help with debugging emitted signals.

For debugging purposes it would/might be useful to have an option for the awesome binary, which would dump/print all emitted signals.

From @psychon on IRC:

> there are two calls to signal_array_getbyid() in common/luaobject.c that are interesting for that
> and in lua there is gears.object which has its own :emit_signal() that would need to be changed

Edit: Uli would propose to use os.getenv() and C's getenv() for implementing this. An AWESOME_DEBUG env var that can be set to "signals"? Perhaps a comma-seperates list of things and gears.debug gets a function that gets called with an option and some strings and dumps them, if enabled?

Adding:

> if you really want to dump all emitted signals, I would suggest using an environment variable for that (because lua already has os.getenv())
> and for recursion reasons, it _might_ make sense to (optionally) print something at the beginning and the end of a signal
> so that one could figure out which signal causes which other ones to run...

Initially reported at: https://awesome.naquadah.org/bugs/index.php?do=details&task_id=1232

### DIFF
--- a/common/luaobject.c
+++ b/common/luaobject.c
@@ -224,6 +224,7 @@ signal_object_emit(lua_State *L, signal_array_t *arr, const char *name, int narg
         foreach(func, sigfound->sigfuncs)
             luaA_object_push(L, *func);
 
+        debug("Emitting class signal '%s' (%d funcs)", name, nbfunc);
         for(int i = 0; i < nbfunc; i++)
         {
             /* push all args */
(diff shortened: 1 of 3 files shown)


psychon:
Closing before this WIP reaches gets its second birthday party. Feel free to reopen later when working on it again.

PR #131, opened 2015-02-15
TBD: Use $DISPLAY variable for D-Bus request name, defaulting to :0

blueyed:
Does this make sense?  Or should "path" or something else be used?

Anyway, awesome-client should probably accept / use vars from the environment, e.g. `DBUS_DEST`.

The use case is with multiple instances of awesome, e.g. in a Xephyr server.

### DIFF
--- a/lib/awful/dbus.lua.in
+++ b/lib/awful/dbus.lua.in
@@ -8,12 +8,19 @@
 local dbus = dbus
 
 --- D-Bus module for awful.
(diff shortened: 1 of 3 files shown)


blueyed:
It's probably easier to just use `dbus-launch` for isolation.

blueyed:
Regarding the last comment (from me) I am closing this for now.

PR #133, opened 2015-02-16
Add functional tests via Xephyr/Xvfb/xdotool

blueyed:
Run it via `t/run.sh`.

Related issues: #136 (Travis integration in general).

Build status of this branch: [Build Status](https://travis-ci.org/blueyed/awesome)

Last successful build: https://travis-ci.org/blueyed/awesome/builds/51381508 (before rebasing on master, where libxcb-xkb is now missing).

It currently uses a pre-built Debian chroot to run the tests in.  A lot of packages are too old on Travis and I've figured that it might be easier to use Debian testing right away.
Using the pre-built images makes the tests run much faster.  This could be achieved by using Travis' cache facilities, but these are only available for container based environments (where `sudo` would not be available).  `sudo` is required for `chroot` (as long as `fakechroot` or something similar is not installed).

TODO:

 - [ ] This might make use of `busted` (more), instead of only using the custom runner.  Needs to be investigated / refactored then.
 - [ ] `W: awesome: a_dbus_connect:607: D-Bus session bus system failed: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory`: this could be avoided by bind-mounting `/mnt/debian/var/run/dbus` and `/mnt/debian/var/lib/dbus` probably, but it's not an issue really, is it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,16 @@ before_install:
 
 install:
  # Based on "apt-get build-depends awesome":
- - sudo apt-get install -qq libcairo2-dev xmlto cmake libxcb-xtest0-dev libxcb-icccm4-dev libxcb-randr0-dev libxcb-keysyms1-dev libxcb-xinerama0-dev libdbus-1-dev libxdg-basedir-dev libstartup-notification0-dev imagemagick libxcb1-dev libxcb-shape0-dev libxcb-util0-dev libgdk-pixbuf2.0-dev libglib2.0-dev libxcb-image0-dev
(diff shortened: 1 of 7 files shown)


blueyed:
I'd really like to get this into master. it would provide CI testing for all forthcoming pull requests.

psychon:
Small comments that have for now: Some of this looks fragile, I don't understand why this has to download a debian chroot from the internet for the tests (nor do I really know the envirionment in which travis works) and finally: These files will end up in release tarballs! IMO we should try to make stuff as small as possible.

blueyed:
> Some of this looks fragile, I don't understand why this has to download a debian chroot from the internet for the tests

Yes, it has to be done mostly because dependencies are too old on Travis, and I found it easier to do this in the end.
See #208 for a new approach on this, but mainly because I've forgotten about this PR first.. :/

> These files will end up in release tarballs

That could skip / ignore certain files.

blueyed:
I will rebuild this on top of #208, once it's in master.

blueyed:
I would appreciate feedback on the files in `t/` (tests and runner): https://github.com/awesomeWM/awesome/pull/133/files#diff-74f5ac166d27a9414d798e8fe1904e84

blueyed:
Closing this in favour of #272, which reuses the test files.

Issue #130, opened 2015-02-15
Tag is not losing "urgent" state


actionless:
After one of the patches merged yesterday tags are not losing "urgent" state after receiving it.

actionless:
can't reproduce it no more

blueyed:
Hmm, it's likely related to https://github.com/awesomeWM/awesome/commit/f2aa8d39edf1427a635a0be3cbc96c7b2f54eb5b.

But I cannot reproduce it, even using the default config.

> can't reproduce it no more

Ok, me neither.. :)

actionless:
just reproduced it again

actionless:
btw, is there any small tool which sets that urgent state on run?

like to do smth like `sleep 3 ; that-tool` and switch to another tag and wait for sleep to finish

blueyed:
sleep 3 ; echo ^G

(where `^G` is literal (bell character), typically done through Ctrl-V Ctrl-G)

You can also use awesome-client:

    echo 'client.get()[2].urgent = true' | awesome-client

actionless:
I will try with the first one. I don't think latter will help to reproduce
it.

actionless:
i did not managed to reproduce it manually, but from time to time that happens in such scenarios:
1) create a rule for some app to open on tag "2"
2) go to tag "1"
3) open that app
tag "2" receives urgent state

blueyed:
That's not a bug until there, is it?

Elv13:
maybe the c:tags() is present before the tagged signal and that result in the counter being incremented twice? Just a theory

actionless:
@blueyed 
> That's not a bug until there, is it?

sorry, i did not understood the question clearly

blueyed:
> tag "2" receives urgent state

That's not a bug, is it?
Or do you mean that a tag gets marked as urgent, when a client opens there?

@psychon reported some issue on IRC where `focus = true` with `tag = ...` behaves differently now. Is this also what you mean?

Anyway, I've created a proof of concept for functional tests: https://github.com/awesomeWM/awesome/pull/133, and started adding some tests for urgent tags / clients.

actionless:
bug is what it's not losing it. scenario was describing prerequisite, ie i am not able to reproduce the problem when tag receiving urgent state in the other way than using rules.

blueyed:
Ok. So the bug is then that the urgent state is still there after you switch the tag?

How do you switch it? Mod4+2? Clicking the tag name?
[...]

PR #137, opened 2015-02-17
awful.tag: fix call to update_urgent from client_untagged

blueyed:
Ref: #130


blueyed:
Please review and merge if it's OK.

PR #139, opened 2015-02-18
spec: fix "require" statements

blueyed:
I don't know why it works by default, but it does not on a minimal
system (Travis, Debian minbase).

### DIFF
--- a/spec/awful/util_spec.lua
+++ b/spec/awful/util_spec.lua
@@ -5,7 +5,6 @@
 
 local util = require("awful.util")
 local say = require("say")
-local assert_util = require("luassert.util")
 
 describe("awful.util", function()
     it("table.keys_filter", function()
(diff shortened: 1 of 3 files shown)


psychon:
Sorry, but this looks like a bug in busted (on IRC we figured out that this might be lua 5.1 vs lua 5.2 or rather setfenv vs no setfenv). I'd prefer this to be reported to the busted guys and see if they really think this is a bug. If not, then we can still fix this in awesome.

blueyed:
This has been changed, and what we're seeing is expected:

> Restrict luassert variables to busted environment

> The luassert variables assert/spy/mock/stub should only be visible
inside a busted context. Other modules outside of busted should be free
to reuse these variable names without interfering with the test
environment. If other modules need to use luassert, they should
explicitly `require` it. Otherwise, `assert` should default to the
builtin Lua `assert`.


(see https://github.com/Olivine-Labs/busted/issues/346#issuecomment-74961215)

I assume that's a "merge this then"?

blueyed:
Committed in c26a961.

PR #142, opened 2015-02-18
Test if cairo-xcb is usable during startup

psychon:
Create a pixmap of size 1x1 and a cairo-xcb surface for this pixmap. If the
surface ends up in an error state, awesome will refuse to start.

This turns a "awesome is unusable and prints lots of errors on X11 servers where
the root window has a depth of 8bpp and no one knows where exactly the problem
is" into "awesome refuses to start on such X11 servers and prints an error that
helps Uli to say immediately what the problem is".


blueyed:
Thanks!

PR #147, opened 2015-02-21
added middle anchors for naughty notifications

kmiller13:
top_middle and bottom_middle can be used as position arguments for new naughty.notifications


blueyed:
Thanks!

Issue #80, opened 2015-01-19
Implement useless gap support


Elv13:
Here we are again,

For the zillion-th time, someone asked for it, so I got bored of the same feature request and hacked it. Here is the code:
https://github.com/Elv13/awesome-1/commit/bb987e4842f6ce67d7723cb979fbce81b0b60914

I need some feedback before porting this to master and doing a proper PR.

![blender230](https://cloud.githubusercontent.com/assets/340384/5796038/befb19cc-9f6b-11e4-8ce9-053f2c5d23f6.png)
![blender229](https://cloud.githubusercontent.com/assets/340384/5796037/bef8d784-9f6b-11e4-93f4-e46002455d02.png)
![blender228](https://cloud.githubusercontent.com/assets/340384/5796039/befc7d4e-9f6b-11e4-89e7-4375683647d0.png)
![blender227](https://cloud.githubusercontent.com/assets/340384/5796040/bf0511ac-9f6b-11e4-82ee-b98d4c5df2cb.png)

It seem to work fine. Is it useful: no. Does it look nice: maybe. Will I recommend anyone using that: no

actionless:
i don't think what visual separation is useless

Elv13:
Oh, I forgot to talk about the code:

Done:

* Implement for all layouts
* Add methods in awful.tag like other properties

What's left:

* Find a way to hook that with beautiful (I don't want to `require()` it in awful.tag
* Remove the 10px default and use 0px
* Somehow add comments in theme or rc.lua to guide users
* Port to master
* Do more advanced testing

BenoitZugmeyer:
For what it's worth, I would love to see this in Awesome. For now, I use [lain](https://github.com/copycat-killer/lain), which is great put has one annoying [issue](https://github.com/copycat-killer/lain/issues/48).

liquiddandruff:
Neat. I'd like this.

Elv13:
@BenoitZugmeyer @liquiddandruff: Can you test the branch and give some feedback?

Thanks

BenoitZugmeyer:
Testing it right now. No issue spotted. I'll keep using it for a few days.

If anyone is interested, [here is an ArchLinux PKGBUILD](https://github.com/BenoitZugmeyer/packages/blob/master/awesome/PKGBUILD)

BenoitZugmeyer:
I've spotted an issue:

* With the layout awful.layout.suit.tile
* Open two clients
[...]

PR #148, opened 2015-02-22
PARTIAL: Add useless gap utility functions

Elv13:
This PR take the awful.tag part of my previous useless gap issue ( https://github.com/awesomeWM/awesome/issues/80 ) as this part is less or more ready. It doesn't add the actual useless gap logic based on ( https://github.com/awesomeWM/awesome/pull/101 ). This commit can be used as base for the actual implementation

### DIFF
--- a/lib/awful/layout/init.lua.in
+++ b/lib/awful/layout/init.lua.in
@@ -7,14 +7,15 @@
 -- Grab environment we need
 local ipairs = ipairs
 local type = type
-local tag = require("awful.tag")
 local util = require("awful.util")
 local ascreen = require("awful.screen")
 local capi = {
     screen = screen,
     awesome = awesome,
-    client = client
+    client = client,
+    tag = tag
 }
+local tag = require("awful.tag")
 local client = require("awful.client")
 local timer = require("gears.timer")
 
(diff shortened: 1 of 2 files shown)


psychon:
I would propose `tag.setgap` to do `gap = math.max(0, gap)` instead of silently ignoring negative arguments. Also, I'd vote for removing the "useless" everywhere and have it only called "gap".

Besides that, looks good to me.

silverhammermba:
I didn't see that you were working on this, so I also [implemented gaps](https://github.com/silverhammermba/awesome/compare/gaps). The big difference is that I made them a client property rather than a tag property.

My reasoning was that border widths are a client property, and gaps are similar in terms of how they are computed and in the resulting effect. But I can also see how the tag property makes sense, since you're most likely going to want uniform gaps on each tag (though I'd say the same for border widths...).

Elv13:
I am not sure, I think it belong to the tag, not the client. Also, having to change the `C` core for something like that is usually a sign a solution is too complicated. I Think it make more sense per tag.

Imagine a client in multiple tags, do you really think the distorted layout it will create is really what the user wish for? Any tags with client with multiple gaps will look very wierd. No?

silverhammermba:
I agree that it would look weird, but isn't the same true for border widths?

You do raise a good point about the C core, and now I realize that I could have just set the new property in Lua as you did with the tags. If there is a great outcry to have gaps per-client rather than per-tag, I'll go back and do that.

Elv13:
There is some use case where you want border per clients. For example, some clients have build in borders or special shapes. In that case, you might want to disable borders on them. This use case doesn't really apply to the gaps.

Elv13:
Deprecated by https://github.com/awesomeWM/awesome/pull/279

PR #150, opened 2015-02-22
layout: Fix delayed arrange regression

Elv13:
```lua
stack traceback:
        /usr/share/awesome/lib/awful/layout/init.lua:102: in function 'arrange'
        /usr/share/awesome/lib/awful/layout/init.lua:164: in function </usr/share/awesome/lib/awful/layout/init.lua:163>
        [C]: ?
        /usr/share/awesome/lib/awful/tag.lua:191: in function 'delete'
        /home/lepagee/.config/awesome/tyrannical/init.lua:205: in function </home/lepagee/.config/awesome/tyrannical/init.lua:198>
error: /usr/share/awesome/lib/awful/layout/init.lua:102: table index is nil
```

Tags don't have a screen for their whole lifecycle, so this happen

### DIFF
--- a/lib/awful/layout/init.lua.in
+++ b/lib/awful/layout/init.lua.in
@@ -97,7 +97,7 @@ end
 --- Arrange a screen using its current layout.
 -- @param screen The screen to arrange.
 function layout.arrange(screen)
-    if delayed_arrange[screen] then return end
+    if not screen or delayed_arrange[screen] then return end
     delayed_arrange[screen] = true
 
     timer.delayed_call(function()


blueyed:
While the fix is good in general probably, I'd like to understand why that
happened.

Is the traceback with master?

Because it seems to come through `arrange_on_tagged` then, which is not used anywhere anymore?!

lib/awful/layout/init.lua:162:

    local function arrange_on_tagged(c, tag)
        if not tag.screen then return end
        layout.arrange(tag.screen)

Elv13:
Let say "near master". But I investigated that before. Tag with no screen is normal, there is a window of time where it is the case. Any callback/signal emitted during that will have this issue.

blueyed:
> Tag with no screen is normal, there is a window of time where it is the case.

But then maybe `layout.arrange` shouldn't be called for no screen in the first place?!

Elv13:
Given the numbers of leafs from where arrange can be called, I don't think trying to safeguard against 100+ codepaths is a better idea than just ignoring the issue, it will get rearanged when screen will be set anyway.
https://github.com/Elv13/awesome-1/blob/fix_arrange_error/lib/awful/layout/init.lua.in#L176

the `tagged` one is the most problematic.

psychon:
:+1:

PR #152, opened 2015-02-24
autofocus: emit `request::activate` signal

blueyed:
After a client gets unmanaged, the newly focused client might not be
visible / raised above other windows.

Using the `request::activate` signal makes the autofocus handler more
flexible, and will raise the focused client by default.


Elv13:
+1

PR #153, opened 2015-02-24
TBD: Revert "layout.arrange only once per main loop"

blueyed:
This reverts 7410646, 1777605 and c88c51f.

The problem is that the change was meant to improve performance, but
triggers problems where a plugin / config relies on geometries being
updated/changed during the same main loop.

The code might be improved to e.g. not delay arranging sometimes, but
then it's easier to restore the old behavior for now.

We might re-add this, but then should provide a way to easily call
something on the next arrange signal, e.g.

    > screen:on_next("arrange", function() ... end)

### DIFF
--- a/lib/awful/layout/init.lua.in
+++ b/lib/awful/layout/init.lua.in
@@ -42,8 +42,6 @@ layout.layouts = {
 -- This is a special lock used by the arrange function.
 -- This avoids recurring call by emitted signals.
 local arrange_lock = false
--- Delay one arrange call per screen.
-local delayed_arrange = {}
 
 --- Get the current layout.
 -- @param screen The screen number.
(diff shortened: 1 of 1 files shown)


Elv13:
I disagree to revert this commit. The new behavior is preferable

blueyed:
I'd prefer to keep it, too.

For reference, the problem was reported for "revelation": https://github.com/guotsuan/awesome-revelation/blob/f8363cbc7443c94722755b7ae7ab8edb2202ba6b/init.lua#L160-L170

psychon:
> The problem is that the change was meant to improve performance, but

Does anyone have some details on why this needs to be improved? Benchmarks or something? Otherwise, isn't this "delay arranges" just premature optimization?

If we knew what *exactly* is slow, other things could be done. For example, the C code could delay actually committing the geometry changes to the end of the main loop iteration (similar to how banning and stacking updates are delayed).

blueyed:
I don't know what is slow exactly, I've only noticed a while ago that there are / can be a lot of calls to `arrange`, which could be avoided.

Benchmarking this might be doable, but then again there are many different use case scenarios of course.

> For example, the C code could delay actually committing the geometry changes to the end of the main loop iteration (similar to how banning and stacking updates are delayed).

Good idea.  But the to-be-avoided part is probably mainly not hitting the `arrange` processing too often.

psychon:
Alternatively, we could add a function to force any delayed arranges to occur. Revelation could then call that function before looking at client geometry (and it could even do so in a backwards compatible way: `(awful.layout.force_pending_arrange or function() end)()`)

But I'm not sure if I really like that solution, but at least that's IMO less complicated than adding some kind of "arrange is done"-signal/callback system. If you want to convince me, all that is needed is a scenario where "lots" (for a suitable high definition of "lots") arranges needlessly occur.

Elv13:
@psychon: The same can be said of `:on_next()`. With the same `or function() end)())`  code, we keep the async aspect when git master while using sync in 3.5.*.

blueyed:
> For example, the C code could delay actually committing the geometry changes to the end of the main loop iteration (similar to how banning and stacking updates are delayed).

I think this would help with #168, too - where a client shortly appears at the old position, before it gets fixed in the `property::screen` signal.

blueyed:
See #174 for only committing geometries (and other properties) once per main loop.

psychon:
Sorry, I don't see this happening.

PR #158, opened 2015-03-02
client.swap.global_bydirection: use `request::activate` signal

blueyed:
Do not set `capi.client.focus` unconditionally, but use the `request::activate` signal if the swapped client was focused currently.

### DIFF
--- a/lib/awful/client.lua.in
+++ b/lib/awful/client.lua.in
@@ -397,7 +397,9 @@ function client.swap.global_bydirection(dir, c)
         end
 
         screen.focus(sel.screen)
-        capi.client.focus = sel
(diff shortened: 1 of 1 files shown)


blueyed:
@psychon 
Yes, it's a behavior change: it allows to move clients (without automatically focusing them), while improving the focus behavior for already focused clients.

Should I clarify the commit message, and then it would be good to commit?

psychon:
I'm not really convinced the new behavior makes sense. Could this possibly be broken down into multiple patches that each only changes one thing, so that people get more useful results from `git bisect`?

Elv13:
This has been merged long time ago

PR #165, opened 2015-03-14
Make client.focus = nil work (Closes #164)

### LINKED ISSUE #164: Unfocusing a client doesn't work
```lua
/* Client module new index.
 * \param L The Lua VM state.
 * \return The number of pushed elements.
 */
static int
luaA_client_module_newindex(lua_State *L)
{
    const char *buf = luaL_checkstring(L, 2);
    client_t *c;

    if (A_STREQ(buf, "focus"))
    {
        c = luaA_checkudata(L, 3, &client_class);
        client_focus(c);
    }

    return 0;
}
```

This code prevent `capi.client.focus = nil` to work. There doesn't seem to be a non-hacky way to remove the focus. A potential solution is to check nil there and set `globalconf.focus.client` and `globalconf.focus.need_update = true;` manually

psychon:

(Disclaimer: No testing done)

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -2326,8 +2326,11 @@ luaA_client_module_newindex(lua_State *L)
 
     if (A_STREQ(buf, "focus"))
     {
-        c = luaA_checkudata(L, 3, &client_class);
-        client_focus(c);
+        c = luaA_checkudataornil(L, 3, &client_class);
+        if (c)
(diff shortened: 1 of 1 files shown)


blueyed:
:+1: 
@Elv13 
Please merge it, if it solves your issue.

Elv13:
It does

PR #168, opened 2015-03-17
Improve floating movetoscreen

blueyed:
This contains a fix and an improvement:

1. handle fullscreen/maximized clients properly: skip storing some properties for them
2. save floating property per screen: this allows to position the client differently per screen, and fixes moving it around on different screen sizes

The issue with 2. is that the client first appears at the "fixed" position and then gets moved to the old one.
I guess this could be fixed by really redrawing clients only once per main loop?

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -10,6 +10,7 @@
 -- Grab environment we need
 local util = require("awful.util")
 local tag = require("awful.tag")
+local timer = require("gears.timer")
 local pairs = pairs
 local type = type
 local ipairs = ipairs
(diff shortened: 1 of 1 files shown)


Elv13:
Can this be closed?

Elv13:
Closing for lack of activity, re-open when needed

blueyed:
@actionless 
> nice

Thanks!  Please consider finishing it.. :)

actionless:
@Elv13 what do you think, can restoring the windows (pt.2 from original description) could be implemented as `awful.placement.restore_floating`?

PR #174, opened 2015-03-20
Only commit/configure client geometries once per main loop

blueyed:
This adds client_geometry_commit and call it from awesome_refresh.
client_geometry_commit is factored out of client_resize_do.

Two new entries in client_t are used for this: need_geometry_commit and
need_geometry_commit_send_notice.  Additionally the flag
globalconf.need_geometry_commit is used for performance reasons.

This fixes the issue where changing the geometry of a client via the
property::screen signal would draw the client twice (see issue #168).

(It will also result in only one SIGWINCH signal in Zsh when (un)maximizing a terminal, instead of two.)

Given this, the delayed calls to layout.arrange are probably not needed
anymore. Ref: https://github.com/awesomeWM/awesome/pull/153#issuecomment-76051092

### DIFF
--- a/event.h
+++ b/event.h
@@ -33,13 +33,13 @@ void luaA_emit_refresh(void);
 
 /* objects/client.c */
 void client_focus_refresh(void);
-void client_geometry_commit(void);
+void client_geometry_refresh(void);
 
 static inline int
 awesome_refresh(void)
 {
     luaA_emit_refresh();
-    client_geometry_commit();
+    client_geometry_refresh();
     banning_refresh();
     stack_refresh();
     client_focus_refresh();
--- a/event.h
+++ b/event.h
@@ -38,6 +38,9 @@ void client_properties_refresh(void);
 /* objects/drawable.c */
 void drawable_geometry_refresh(void);
 
+/* objects/drawin.c */
+void drawin_geometry_refresh(void);
+
 /* objects/window.c */
 void window_properties_refresh(void);
 
@@ -48,6 +51,7 @@ awesome_refresh(void)
     client_properties_refresh();
     window_properties_refresh();
     drawable_geometry_refresh();
+    drawin_geometry_refresh();
     banning_refresh();
     stack_refresh();
     client_focus_refresh();
--- a/event.h
+++ b/event.h
@@ -33,13 +33,17 @@ void luaA_emit_refresh(void);
 
 /* objects/client.c */
 void client_focus_refresh(void);
-void client_geometry_refresh(void);
+void client_properties_refresh(void);
+
+/* objects/window.c */
+void window_properties_refresh(void);
 
 static inline int
 awesome_refresh(void)
 {
     luaA_emit_refresh();
-    client_geometry_refresh();
+    client_properties_refresh();
+    window_properties_refresh();
     banning_refresh();
     stack_refresh();
     client_focus_refresh();
--- a/event.h
+++ b/event.h
@@ -35,6 +35,9 @@ void luaA_emit_refresh(void);
 void client_focus_refresh(void);
 void client_properties_refresh(void);
 
+/* objects/drawable.c */
+void drawable_geometry_refresh(void);
+
 /* objects/window.c */
 void window_properties_refresh(void);
 
@@ -44,6 +47,7 @@ awesome_refresh(void)
     luaA_emit_refresh();
     client_properties_refresh();
     window_properties_refresh();
+    drawable_geometry_refresh();
     banning_refresh();
     stack_refresh();
     client_focus_refresh();
--- a/event.h
+++ b/event.h
@@ -33,11 +33,13 @@ void luaA_emit_refresh(void);
 
 /* objects/client.c */
 void client_focus_refresh(void);
+void client_geometry_commit(void);
 
 static inline int
 awesome_refresh(void)
 {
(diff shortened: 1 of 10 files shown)


blueyed:
@psychon 
Thanks for your feedback, I've addressed all your remarks.

I've noticed that properties like `opacity` should get handled in a similar way probably, and that it might make sense to have a more generic approach to this: something like a "needs property refresh" flag and then storing what needs to be updated in a list.

blueyed:
I've given the list approach a try in https://github.com/blueyed/awesome/commit/9693fe19316c895299a1b073144b0aa2a0e04775.  It uses a list of windows (which appears to be the base for drawins and clients, correct?), and handles the opacity attribute through it.

Does this look sane?
Can / should the list be merged into a single one?
Or should there just be a new list for clients? (instead of the flag and then looping over all clients)

psychon:
There is no point in delaying the opacity changes. Its a really cheap thing to do and the composite manager likely isn't fast enough to even get the "intermediate value". I'm not sure that this makes sense for `border_width` either. That only affects things internal to the X11-server... Premature optimization?

For adding fields to windows (aka "both drawin and clients"), you can use `WINDOW_OBJECT_HEADER` in `objects/window.h`.

blueyed:
Ok, I'll only use `window_properties_refresh` then.

I've thought that `border_width` might be a good candidate, because it often gets changed multiple times (through signals), and has a larger impact (e.g. on re-drawing layouts).

blueyed:
I've rebased this and added a list to handle clients, too.

I've changed the lazy-committing for windows to use `border_width` instead of `opacity`.
While it's not really critical for `border_width`, it can be useful in some cases, where `border_width` gets changed multiple times due to signals.

I've tried to handle both windows and clients in the same loop/list, but while it's possible to cast a client to a window, the other way does not seem to be possible, or at least seemed to cause weird effects?!

psychon:
Sorry, I'm still not really convinced for this to have any "nice" effects and it's relatively easy to break ICCCM via this (some actions from a window require an action from the WM every time this is done, even when this is done twice in short succession).

I'll be lazy and look at other things for now.

blueyed:
Seems fair, I'll cut it down to only use `client_properties_refresh` then (https://github.com/awesomeWM/awesome/pull/174/files#diff-5be9034afa9b5e045561617bb1d673dbR797), to handle the geometry only, which was the initial intention of this PR anyway.

blueyed:
@psychon 
Does it make sense to handle this for drawins, too? (via [`drawin.moveresize`](https://github.com/awesomeWM/awesome/blob/master/objects/drawin.c#L148))

I've looked at this branch again, because it might help with flickering because of multiple calls to `geometry` for tooltips (when the text changes and it gets re-placed automatically).

psychon:
If you want to, feel free to do this for drawins. I guess I would just add a boolean to drawins `need_geometry_update` and in `awesome_refresh` go to all drawins and position the ones that have this boolean set.

I'm not really sure what to do with the `drawin_update_drawing` call. Doing it immediately might cause issues if lua does a redraw before the drawin is actually resized, delaying it might have unintended consequences to lua code.

blueyed:
I guessed that the problem I'm seeing (`mouse::leave` gets emitted for a widget with a tooltip left of the mouse cursor, when the tooltip's width grows into the mouse pointer) is caused via `drawable_set_geometry`, and tried to delay that - but it seems I was wrong..

Now I've tried to add delayed geometry handling for drawins, but it is not working as expected: the drawins appear to not been drawn anymore at all.  There is some slight indication of them, but they appear to have a size of 0/1?!

That might be related to the `drawin_update_drawing` call you've mentioned: it calls `drawable_set_geometry`, which is delayed now, too.

Calling `drawin_geometry_refresh` just after adding to its array works.

Maybe the order of the calls in `awesome_refresh` is relevant?!  It is the following currently:

    client_properties_refresh();
    window_properties_refresh();
    drawable_geometry_refresh();
    drawin_geometry_refresh();

Given the order of calls it would make sense maybe to have a single list for the different types?!
(that's what the idea with generic window properties was somehow).

Apart from that, this hits the limits of my C foo constantly.. :/

Besides the newer issues, I've noticed that the stacking order was weird before already (with the PR as of yesterday/before): tooltips and naughty notifications appears behind clients.

psychon:
I added a comment on the corresponding commit for the "resizes get lost" issue. Sorry, but I also wrote a patch which does what you are trying to do (sorry for stealing your work, feel free to experiment with that and submit it as a PR / tell me to submit it as a PR). And `drawable_set_geometry()` definitely will not cause the X11 server to send us any events.

Elv13:
Is this still relevant. If yes, should it be a 3.6 blocker? If not, should this be closed? Is there really that many instances of `:geometry()` on the same drawin during the same event loop to justify this?

psychon:
Drawins were handled in commit 9408b61a974bc351e285592f36b2e325c66d5cf2. The remaining "issue" are clients and these are harder to handle, because we have to send some "you were moved without a resize"-events to clients in some cases (this is what I meant with the "resize get lost" above, I think).

But yeah, I agree that there shouldn't be that many `:geometry()` calls on a single client per main loop iteration. If someone disagrees, I'd volunteer to implement this, if needed. For now I'll close this.

PR #175, opened 2015-03-21
ewmh: screen_change: fix off-screen geometries

blueyed:
When moving a fullscreen/maximized client to a smaller screen or with
different Y offset etc, its offset was not handled properly.

This moves it first into the new area, and then makes sure that its not
off-screen.  The latter should not happen, because the width/height is
adjusted before.

### DIFF
--- a/lib/awful/ewmh.lua.in
+++ b/lib/awful/ewmh.lua.in
@@ -106,11 +106,21 @@ local function screen_change(window)
                     if data[window][reqtype].x then
                         new_x = to.x + data[window][reqtype].x - from.x
                         if new_x > to.x + to.width then new_x = to.x end
+                        -- Move window if it overlaps the new area to the right.
+                        if new_x + data[window][reqtype].width > to.x + to.width then
+                            new_x = to.x + to.width - data[window][reqtype].width
+                        end
+                        if new_x < to.x then new_x = to.x end
                         data[window][reqtype].x = new_x
                     end
                     if data[window][reqtype].y then
                         new_y = to.y + data[window][reqtype].y - from.y
                         if new_y > to.y + to.width then new_y = to.y end
+                        -- Move window if it overlaps the new area to the bottom.
+                        if new_y + data[window][reqtype].height > to.y + to.height then
+                            new_y = to.y + to.height - data[window][reqtype].height
+                        end
+                        if new_y < to.y then new_y = to.y end
                         data[window][reqtype].y = new_y
                     end
                 end


psychon:
Sorry, I'm confused. "The latter should not happen" sounds like all of this shouldn't make a difference?

Since this seems to confuse me, perhaps we can write some test cases for this? ;-) (I'm ok without them, too, I just don't get this change, sorry)

blueyed:
The first part is necessary:

    -- Move window if it overlaps the new area to the right.
    if new_x + data[window][reqtype].width > to.x + to.width then
        new_x = to.x + to.width - data[window][reqtype].width
    end

Otherwise windows are not handled properly when being moved to a smaller screen to the left.  That's the issue I was seeing.

I am not sure if the additional check below is needed:

    if new_x < to.x then new_x = to.x end

getzze:
I experience a related issue.
When moving a fullscreen client to a larger screen, the client is not resized to the large screen size, and is a small window in the upper left corner.
I monkey patched awful.client.movetoscreen and awful.client.movetotag (https://github.com/getzze/awesome_config/blob/master/lib/patch_move_client.lua) but it is not sufficient. The resizing should be done also when client.screen is set to a new value.

blueyed:
@getzze 
IIRC this should be handled already, and might have been another bug that has been fixed in master.
Which awesome version are you using?  Can you try Git master?

getzze:
I am using awesome 3.5.6.
I will try the git master when I can.

blueyed:
@psychon 

> Since this seems to confuse me, perhaps we can write some test cases for this? ;-)

That would be nice. Related: https://github.com/awesomeWM/awesome/pull/133 - but this could be done better using unit tests maybe?!

Apart from that I'm going merge this in the next days if there's no other feedback.

@getzze 
Any updates on your issue?

btw: instead of monkeypatching awesome, you might want to consider a pull request in the future directly.. :)

getzze:
The patch I created was not working actually, but I added the re-fullscreen function and connected it to the signal of screen changed in the `rc.lua` and it is working perfectly now:

    local function client_reload_max(c)
        local c = c or client.focus
        if not c then return end
        if c.maximized then
             --naughty.notify({text="Maximized ! " .. c.name, screen=c.screen})
            c.maximized = false
            c.maximized = true
        else
            if c.maximized_horizontal then
                c.maximized_horizontal = false
                c.maximized_horizontal = true
            end
            if c.maximized_vertical then
                c.maximized_vertical = false
                c.maximized_vertical = true
            end
        end
        if c.fullscreen then
            --naughty.notify({text="Fullscreen ! " .. c.name, screen=c.screen})
            c.fullscreen = false
            c.fullscreen = true
        end
    end

    -- Connect change screen signal to a resize function
    client.connect_signal("property::screen", client_reload_max)

blueyed:
@getzze 
I think https://github.com/awesomeWM/awesome/pull/168 might be related to your issue.

PR #177, opened 2015-03-23
Integrate documentation into C files

juw:
This PR moves the documentation of the C-API as good, as it is currently (to my knowledge) possible with ldoc, into the C source files. Furthermore it rips out the (in the current state) useless doxygen support and makes it possible to build awesome against Lua 5.2.


juw:
I just made a few changes to the PR, mainly some rewording in the description for the signal and xproperty functions.

blueyed:
@juw 
Thanks for this!

Can you elaborate a bit on the before and after of this PR, in terms of how the docs look?

I've tried building them, but then `make` failed:

    Linking C executable awesome
    CMakeFiles/awesome.dir/awesome.c.o: In function `awesome_refresh':
    …/awesome/event.h:40: undefined reference to `luaA_emit_refresh'
    …/awesome/event.h:43: undefined reference to `client_focus_refresh'
    ...

I've removed the existing build dir and used `make clean` without success.

I see that the source files have been moved to `*.in`, which may be problematic
for any existing pull requests and patches in terms of conflicts.
Could this be avoided, e.g. by copying/using the original files with the docs target?

juw:
Am Thu, 26 Mar 2015 11:05:45 -0700
schrieb Daniel Hahler <notifications@github.com>:

> @juw 
> Thanks for this!
> 
> Can you elaborate a bit on the before and after of this PR, in terms
> of how the docs look?
It basically looks the same, the only difference is, that
class:connect_signal etc. are not documented in the classes in which
they are used anymore, but in an extra file, which is referenced from
the classes, where these functions can be used, since it is not
possible with ldoc, to have the documentation for the same function
reused in several classes.

> I've tried building them, but then `make` failed:
> 
>     Linking C executable awesome
>     CMakeFiles/awesome.dir/awesome.c.o: In function `awesome_refresh':
>     …/awesome/event.h:40: undefined reference to `luaA_emit_refresh'
>     …/awesome/event.h:43: undefined reference to
> `client_focus_refresh' ...
> 
> I've removed the existing build dir and used `make clean` without
> success.
> 
> I see that the source files have been moved to `*.in`, which may be
> problematic for any existing pull requests and patches in terms of
> conflicts. Could this be avoided, e.g. by copying/using the original
> files with the docs target?
Sorry, I forgot to check if the whole build works and only checked
building the docs via "make ldoc". I updated the PR, now building
awesome should work again.

juw:
> I see that the source files have been moved to `*.in`, which may be
> problematic for any existing pull requests and patches in terms of
> conflicts. Could this be avoided, e.g. by copying/using the original
> files with the docs target?

Might be, that it is possible but I did not try it. Maybe if one
modifies this macro in awesomeConfig.cmake

    macro(a_configure_file file)
        string(REGEX REPLACE ".in\$" "" outfile ${file})
        message(STATUS "Configuring ${outfile}")
        configure_file(${SOURCE_DIR}/${file}
                       ${BUILD_DIR}/${outfile}
                       ESCAPE_QUOTE
                       @ONLY)

But having *.in files, makes it more clear, that these files get
modified by the buildsystem. For clarity I would prefer to have them
either all called *.in or none called *.in but not a mixture of both.

blueyed:
Still fails to build:

    [ 28%] Building C object CMakeFiles/awesome.dir/objects/button.c.o
    …/awesome/.build-localhost-x86_64-linux-gnu-4.8/objects/button.c:33:20: fatal error: button.h: No such file or directory
     #include "button.h"
                        ^
    compilation terminated.

> But having *.in files, makes it more clear, that these files get modified by the buildsystem.

Agreed, but as stated before it has its drawbacks - which also applies to `*.lua.in`.

I think that generating docs might be considered to be an exception in this regard.

In my opinion it should be avoided to rename so many files for this.

blueyed:
Re signals: I think they should be documented from the source, too.  AFAIK there's currently only [the list in the wiki](http://awesome.naquadah.org/wiki/Signals#awesome).
Can such a list be generated, i.e. a section where all signals are listed?

juw:
> Still fails to build:
> 
>     [ 28%] Building C object CMakeFiles/awesome.dir/objects/button.c.o
>     …/awesome/.build-localhost-x86_64-linux-gnu-4.8/objects/button.c:33:20:
> fatal error: button.h: No such file or directory #include "button.h"
>                         ^
>     compilation terminated.
Oups, the header files have to be copied to, this is done now.

> 
> > But having *.in files, makes it more clear, that these files get
> > modified by the buildsystem.
> 
> Agreed, but as stated before it has its drawbacks - which also
> applies to `*.lua.in`.
> 
> I think that generating docs might be considered to be an exception
> in this regard.
> 
> In my opinion it should be avoided to rename so many files for this.
Yes, that would be nicer. If so many renames (basically all lua files)
are no problem, I could have a go at it.

juw:
> Re signals: I think they should be documented from the source, too.
> AFAIK there's currently only [the list in the
> wiki](http://awesome.naquadah.org/wiki/Signals#awesome). Can such a
> list be generated, i.e. a section where all signals are listed?
It could, I will have a look at it.

juw:
> I see that the source files have been moved to *.in, which may be problematic
> for any existing pull requests and patches in terms of conflicts.
git detects renames, so I do not think, that it would be a big problem.

juw:
Regarding the documentation of the signals: Did you have something in mind, like I did in the updated patch (https://github.com/juw/awesome/commit/a3102f5570d0e8bfa00cf302925050df75f1a060)?

blueyed:
@juw 
Yes, `@signal` seems to be what I've meant (https://github.com/juw/awesome/commit/12aa6c8afca2ecba5b75ecdba5d3ceee9ac0a3a0#diff-1e8b36a227f955f3c3ff63d9598006b9R154).

It shows up in the doc like this: http://codeprobe.de/tmp/awesome-doc/classes/button.html#Signals

Is it possible to have a section / item in the navigation that would list all signals, like in the wiki?

blueyed:
Why is there a dot for the client `manage` signal (and others)? (http://codeprobe.de/tmp/awesome-doc/classes/client.html)

blueyed:
I think signals should be listed after "Tables" and "Methods" (http://codeprobe.de/tmp/awesome-doc/classes/client.html).

juw:
> I think signals should be listed after "Tables" and
> "Methods" (http://codeprobe.de/tmp/awesome-doc/classes/client.html).

This is how ldoc orders them automatically, I do not think, that one
can influence that.


> Why is there a dot for the client `manage` signal (and others)?
> (http://codeprobe.de/tmp/awesome-doc/classes/client.html)

Otherwise ldoc would automatically replace `name` with `class.name`
before it, eg. `client.press` instead of `press`. I could not find how
to stop ldoc from doing that, probably one would need to extend ldoc
for that case.


> Is it possible to have a section / item in the navigation that would
> list all signals, like in the wiki?

Only if one would add an extra page in the docs and therefore would add
an extra page in the source code, which only describes the signals. I
do not see the benefit of that, since the signals make only sense and
work if one uses the respective classes.

blueyed:
I've finally looked into this again, and rebased in on master: https://github.com/awesomeWM/awesome/pull/266

>> Why is there a dot for the client `manage` signal (and others)?
>> (http://codeprobe.de/tmp/awesome-doc/classes/client.html)

> Otherwise ldoc would automatically replace `name` with `class.name`
before it, eg. `client.press` instead of `press`. I could not find how
to stop ldoc from doing that, probably one would need to extend ldoc
for that case.

Maybe it's better to use `::manage` for now then?

juw:
> I've finally looked into this again, and rebased in on master:
> https://github.com/awesomeWM/awesome/pull/266
Looks good, cheers.

> >> Why is there a dot for the client `manage` signal (and others)?
> >> (http://codeprobe.de/tmp/awesome-doc/classes/client.html)
> 
> > Otherwise ldoc would automatically replace `name` with `class.name`
> before it, eg. `client.press` instead of `press`. I could not find how
> to stop ldoc from doing that, probably one would need to extend ldoc
> for that case.
> 
> Maybe it's better to use `::manage` for now then?
I do not know and do not think it really matters, what is used, since
the actual signal name would still be `manage` and one still would need
a hint like the one, that is there ("Some signal names are starting
with a dot. These dots are artefacts from the documentation generation,
you get the real signal name by removing the starting dot."). If you
prefer `::manage` than change it.

Cheers.

blueyed:
Closing this in favour of https://github.com/awesomeWM/awesome/pull/266, which will be merged soonish.

Thanks again!

blueyed:
@juw 

>> Why is there a dot for the client `manage` signal (and others)?

> Otherwise ldoc would automatically replace `name` with `class.name`
before it, eg. `client.press` instead of `press`. I could not find how
to stop ldoc from doing that, probably one would need to extend ldoc
for that case.

I've found that using `request::activate` as the name does not allow to link to it (via `foo` or `@see`).
Any solution to that?

We could use names that would be different from the actual signal, e.g. `request_activate`, but that does not look nice, and might also conflict with actual functions of the same name etc.
Using something like `signal.request_activate` seems even worse.

Maybe a custom ldoc handler could be used?
OTOH it is good for now to just refer to it by name, like I've done in https://github.com/awesomeWM/awesome/pull/529/files#diff-8691318d56f240840f33b4376de0713bR160.

PR #180, opened 2015-03-27
awesomeConfig.cmake: Allow setting AWESOME_DATA_DIR

heirecka:
The motivation behind this patch is my distro moving to a multiarch
layout. While binaries, libraries, etc. are are installed into
/usr/{host}/{bin,lib,...} architecture-independent data should still
go to /usr/share.


psychon:
Fine by me, but I wonder if there could be "something" more natural to handle this. What are the usual configure arguments that distros use for multi-arch and can these be translated "more naturally" into CMake arguments? (Without some special case with "AWESOME" in the name)

heirecka:
Not especially multi-arch specific but cmake's GNUInstallDir module (http://www.cmake.org/cmake/help/v3.2/module/GNUInstallDirs.html) probably can configure every relevant directory.

Elv13:
I used it in other projects I am involved with. It does indeed work as advertised for various multi arch approach (aka, /usr/lib{32,64}/, /usr/x86[_64]-pc-linux-gnu/, /usr/lib/ and others common ones)

psychon:
I guess we should eventually use that `GNUInstallDirs` module, but this is orthogonal to this PR.

psychon:
Oh, any volunteers for making the switch to that module? :-)

heirecka:
I'd be willing to do it if nobody beats me to it. It might take a few days until I have time for it, though.

psychon:
Thanks a lot!

PR #183, opened 2015-03-29
config: sloppy focus: ignore mouse::enter if mouse was not moved

blueyed:
When raising a client's transient_for window, the parent window gets
also raised.  This might trigger a `mouse::enter` signal for the parent
window, if it was not visible at the mouse pointer's position before.

The sloppy focus handler in the default config now ignores any
`mouse::enter` events, when the mouse coordinates have not changed since
the last signal.

### DIFF
--- a/awesomerc.lua.in
+++ b/awesomerc.lua.in
@@ -435,20 +435,16 @@ client.connect_signal("manage", function (c)
 end)
 
 -- Enable sloppy focus.
-local sloppyfocus_last = {x=nil, y=nil, c=nil}
+local sloppyfocus_last = {c=nil}
 client.connect_signal("mouse::enter", function(c)
-    local mcoords = mouse.coords()
     if awful.layout.get(c.screen) ~= awful.layout.suit.magnifier
             and awful.client.focus.filter(c) then
         -- Skip focusing the client if the mouse wasn't moved.
-        if (mcoords.x ~= sloppyfocus_last.x or mcoords.y ~= sloppyfocus_last.y)
-            and c ~= sloppyfocus_last.c then
+        if c ~= sloppyfocus_last.c then
             client.focus = c
             sloppyfocus_last.c = c
         end
     end
-    sloppyfocus_last.x = mcoords.x
-    sloppyfocus_last.y = mcoords.y
 end)
 
 client.connect_signal("focus", function(c) c.border_color = beautiful.border_focus end)
--- a/awesomerc.lua.in
+++ b/awesomerc.lua.in
@@ -434,12 +434,18 @@ client.connect_signal("manage", function (c)
     end
 end)
 
--- Enable sloppy focus
+-- Enable sloppy focus.
+local sloppy_last_coords = {x=nil, y=nil}
 client.connect_signal("mouse::enter", function(c)
+    local mcoords = mouse.coords()
     if awful.layout.get(c.screen) ~= awful.layout.suit.magnifier
-        and awful.client.focus.filter(c) then
-        client.focus = c
+            and awful.client.focus.filter(c) then
+        -- Skip focusing the client if the mouse wasn't moved.
+        if mcoords.x ~= sloppy_last_coords.x or mcoords.y ~= sloppy_last_coords.y then
+            client.focus = c
+        end
     end
+    sloppy_last_coords = mcoords
 end)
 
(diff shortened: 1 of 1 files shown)


blueyed:
In my own config I was remembering the client from the last signal and compared it to the one from the current signal, which might be better and/or could be used additionally.

It helped to ignore the signal when using `rofi`, which uses a temporary redirected / on-top window.

psychon:
Isn't this exactly the reason why `client_ignore_enterleave_events()` exists in the C code? To avoid spurious enter signals. This problem could be fixed by adding such calls around `stack_refresh` in the C code.

Long-term (aka "if anyone ever wants those signals to occur"), we could provide a context argument to the `mouse::enter` signal that says if the mouse was really moved or if this happened because some client was configured suitably...

r8b7xy:
Your pull request solve the problem with unclutter I've mentioned in [#202](https://github.com/awesomeWM/awesome/issues/202) but I've found one drawback - cursor unhide when I close last client under hidden cursor. Do you have any idea how to fix this issue?

blueyed:
@r8b7xy 
Can you check if the last change (https://github.com/blueyed/awesome/commit/0438133dfc3fc699368e53df98b5f8f1f31e93bc) still fixes it for you? (it only looks at the client)
That might also fix the issue mentioned in your comment above.

At least it helps with the extra popup about potentially forgotten attachments from Thunderbird after the "Send" dialog.  This wasn't focused for me.

r8b7xy:
``` lua
- run_once("unclutter")
+ run_once("unclutter -root")
```
and your last change (blueyed@ 0438133) fix the problem.
With previous configuration I've mentioned here (copycat-killer/awesome-copycats#81) and root option cursor sometimes blinked.
Thank you.

blueyed:
@psychon 
> Isn't this exactly the reason why `client_ignore_enterleave_events()` exists in the C code?

Not sure if it would help here: the problem exists with e.g. Thunderbird, too:

 - maximized Thunderbird main window as the only client on a tiled tag
 - right click on a mail, choose "Forward as => Attachment"
 - the "Compose" window pops up, which is handled via rules, and will be not below the mouse cursor

    { rule = { class = "Thunderbird", instance = "Msgcompose" },
      properties = { floating = true }},

 - the main window gets focused, instead of the new one

This is not really reproducible, but lead me to dig out this old PR again.

It seems like `mouse::enter` really should not be triggered in this case, and is not done so anymore after looking closely at it.

Any ideas why `mouse::enter` might have been triggered, or what else might have caused this?

psychon:
The only commit that I remember that messes with enter/leave would be d96c4d61a9a50b205f8b2e01635775cd4bc90fb5. However, I don't think that should make a difference here. Dunno...

Elv13:
Can this be closed or is it sitll relevant?

Elv13:
Closing for lack of activity re-open if needed.

blueyed:
Revisited my code.

What do you think?

```lua
-- Enable sloppy focus.

-- Skip focusing the client if the mouse wasn't moved (to a new client, and
-- in general).
do
  local sloppyfocus_last
  local sloppyfocus_timer
  local sloppyfocus_timeout = 0.1

    if not use_sloppy_focus then  -- NOTE: I am using this for toggling it, but could/should connect/reconnect the signal instead.
      return
    end
    if awful.layout.get(c.screen) == awful.layout.suit.magnifier or
        not awful.client.focus.filter(c) then
      return
    end

    -- Stop any existing timer.
    if sloppyfocus_timer then
      sloppyfocus_timer:stop()
      sloppyfocus_timer = nil
    end

    -- Check if focused already (when moving back and forth).
    if client.focus == c then
      sloppyfocus_last = nil
      return false
    end

    local mcoords = mouse.coords()
    sloppyfocus_last = {x = mcoords.x, y = mcoords.y}

    sloppyfocus_timer = timer.start_new(sloppyfocus_timeout, function()
      mcoords = mouse.coords()
      if (mcoords.x == sloppyfocus_last.x and mcoords.y == sloppyfocus_last.y) then
        c:emit_signal("request::activate", "mouse_enter", {raise = false})

        -- Reset state and stop timer.
        sloppyfocus_last = nil
        sloppyfocus_timer = nil

        return false
      end

      -- Update mouse position, and restart timer.
      sloppyfocus_last.x = mcoords.x
      sloppyfocus_last.y = mcoords.y

      return true
    end)
  end)
end
```

PR #186, opened 2015-03-30
Set &error in parse_command, return it from luaA_spawn

blueyed:
In case of parse errors from `g_shell_parse_argv`, this now gets
returned by `luaA_spawn`.


psychon:
:+1:

PR #187, opened 2015-03-30
Various fixes / improvements to prompt.run / completion

blueyed:
I've came up with these when looking into completion issues.

This fixes e.g. `Shift-Tab` and completion from inside of the command (not at then end).

~~It also fixes the directory detection/handling for Lua 5.2, adding `awful.util.is_dir`.~~ (extracted)

### DIFF
--- a/lib/awful/completion.lua
+++ b/lib/awful/completion.lua
@@ -81,7 +81,11 @@ function completion.shell(command, cur_pos, ncomp, shell)
     end
 
     while wend <= #command do
-        wend = command:find(" ", wstart)
+        wend, nextw = command:find(" +", wstart)
+        -- Skip escaped spaces.
+        while wend and command:sub(wend-1, wend-1) == "\\" do
+            wend, nextw = command:find(" +", wend + 1)
+        end
         if not wend then wend = #command + 1 end
         table.insert(words, command:sub(wstart, wend - 1))
         if cur_pos >= wstart and cur_pos <= wend + 1 then
@@ -89,8 +93,10 @@ function completion.shell(command, cur_pos, ncomp, shell)
             cword_end = wend
             cword_index = i
         end
-        wstart = wend + 1
-        i = i + 1
+        if nextw then
+            wstart = nextw + 1
+            i = i + 1
+        end
     end
 
     if cword_index == 1 and not string.find(words[cword_index], "/") then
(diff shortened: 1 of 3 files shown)


psychon:
I don't think this version of `awful.util.is_dir` is portable / works correctly everywhere. Besides that, I have no clue about the completion code and just trust you.

blueyed:
@psychon 
`f:seek("end")` does not work (anymore?!) - it now returns 9.2233720368548e+18 for dirs..

Could we make luafilesystem a dependency, and then just use `lfs.attributes(path, 'mode')=='directory'`?

A slower alternative would be `os.execute("test -d past")`, but that should be avoided.

Elv13:
```
Could we make luafilesystem a dependency, and then just use lfs.attributes(path, 
```

LGI already has that function from GLib, no need for new dependencies beside maybe `Gio`

blueyed:
Just for reference: the current PR introduced a few regressions/bugs that I will have to look into.

blueyed:
> LGI already has that function from GLib, no need for new dependencies beside maybe Gio

Can you provide some code that would do this, please?

Elv13:
> > LGI already has that function from GLib, no need for new dependencies beside maybe Gio
>
> Can you provide some code that would do this, please?

Not until next week, I have some work deadlines to meet, you can look at `fd_async`  from my config, there is something similar. Of course we need to wrap it in a `lua` friendly API, but `Gio` > `LuaFilesystem` as it integrate with our event loop and is asynchronous with co-routines.

psychon:
$ lua -e 'g = require("lgi").GLib ; print(g.file_test("/tmp", { g.FileTest.IS_DIR }))'
    true
    $ lua -e 'g = require("lgi").GLib ; print(g.file_test("/etc/passwd", { g.FileTest.IS_DIR }))' 
    false
    $ lua -e 'g = require("lgi").GLib ; print(g.file_test("/not-exist", { g.FileTest.IS_DIR }))' 
    false

The other possible flags (`IS_REGULAR`, `IS_SYMLINK`, `IS_EXECUTABLE` and `EXISTS`) sound interesting, too.

psychon:
Should some of this be rebased and merged (and perhaps the unrelated `is_dir` one gets its own PR)?

blueyed:
Yes.  I've extraced `is_dir` in https://github.com/awesomeWM/awesome/pull/536.

I would still need to closer at the rest, according to my previous comment:

> Just for reference: the current PR introduced a few regressions/bugs that I will have to look into.

psychon:
Closing for lack of activity

PR #191, opened 2015-04-01
added dbus.call_method

dodo:
Added the ability to call D-Bus methods.

Plus i fixed a bug with `dbus_message_new_signal` returning `NULL`.

To call a method you best do:
```lua
local name, on_method_return
on_method_return = function (method, ...)
    dbus.disconnect_signal(name, on_method_return)
    -- process method return result here
end
local serial = dbus.call_method('org.dbus.destination', '/org/dbus/path/', 'org.dbus.interface', 'Method', unpack(args))
name = string.format('reply %d', serial)
dbus.connect_signal(name, on_method_return)
```

What do you think?

### DIFF
--- a/dbus.c
+++ b/dbus.c
@@ -397,6 +397,23 @@ a_dbus_process_request(DBusConnection *dbus_connection, DBusMessage *msg)
     if(dbus_message_iter_init(msg, &iter))
         nargs += a_dbus_message_iter(L, &iter);
 
+    if (dbus_message_get_reply_serial(msg))
+    {
+        char buf[27];
+        snprintf(buf, sizeof(buf), "reply %d%c", dbus_message_get_reply_serial(msg), '\0');
+        signal_t *sigfound = signal_array_getbyid(&dbus_signals,
+                                                  a_strhash((const unsigned char *) &buf));
+        /* emit signals */
+        if(sigfound) {
+            /* copy stack */
+            for (int i = 0; i < nargs; i++) {
+                lua_pushvalue(L, -nargs);
+            }
+            signal_object_emit(L, &dbus_signals, (const char *) &buf, nargs);
+        }
+
+    }
+
     if(dbus_message_get_no_reply(msg))
     {
         signal_t *sigfound = signal_array_getbyid(&dbus_signals,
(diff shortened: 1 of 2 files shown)


actionless:
when i started to work on it @Elv13 asked to stop doing that because later dbus cmodule will be replaced with `lua-lgi` implementation

Elv13:
LGI-GDBus is indeed not very stable... I wonder if we should just port the C code to GDBus and expose it ourself... We could still use the LGI GVariant<->Lua bridge, so it would save some work.

Another issue I have with this implementation is that it isn't asynchronous and I am not sure it is flexible enough with "in" and "out" parameters. I haven't looked to deep into it yet.

We do need better DBus support, but I am still not sure of how it should be done. The libdbus API is not designed to be used by application and wont work with kDBus later this year.

psychon:
Except for the new `call_method()` function, the various fixes seem nice and I think we should merge them.
For `call_method()`:
```
to catch the method_result you have to use the serial number returned by
bdus.call_method. call dbus.connect_signal(string.format('serial %d',
serial), function (method, ...) end) to listen for the method_return.
use dbus.disconnect_signal to properly clean up signal callback.
```
That seems like a terrible API. However, I'm not really sure what a better API would be.

Elv13:
Here is my API for getting property, the method calling is similar
https://github.com/Elv13/wirefu/blob/master/demo/upower.lua

I am currently porting it to C+lua because using LGI is too unstable (it crash quite often due to bugs in LGI handling of recursive G_CALLBACK)

I made some progress this week on the asynchronous I/O api and GDK XDG icon API so @blueyed can play with it sooner. The current state is that it can be called from LUA and LUA is notified of the request completion. I still need to implement some lua bits to get the return value, fix multiple memory leaks and check for errors more often. After that (probably today or tomorrow actually) I will publish it as a module so you can review/test it before submitting it to Awesome. For the testing I done so far, it is stable (I did unit testing too). After that, it will be GDbus turn. I will port my `wirefu` module to use C for the actual GDBus handling as it suffer from the same crashes as the I/O module.

psychon:
Cherry-picked e8d209fde60333885e09d6c618b39ea9ac7c7c63, 33d4100ea0740acbd5002d340c1eba268f55ed94, and 8557a7e0e48cd94dc2565ded7145560f6bff1fda.
The rest either has unaddressed comments or didn't apply. Closing this pull request due to these problems, but feel free to reopen with an update.

PR #198, opened 2015-04-06
expose dbus signal sender

dodo:
Another addition to the dbus api.

I found it useful to filter signals by sender, specially in combination with PR #191.

### DIFF
--- a/dbus.c
+++ b/dbus.c
@@ -384,6 +384,12 @@ a_dbus_process_request(DBusConnection *dbus_connection, DBusMessage *msg)
     lua_pushstring(L, s);
     lua_setfield(L, -2, "member");
 
+    s = dbus_message_get_sender(msg);
+    if (s != NULL) {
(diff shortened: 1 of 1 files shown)


Elv13:
Thanks for this, but really, we need to replace this code very soon. KDBus is comming probably next week:

http://lkml.iu.edu/hypermail/linux/kernel/1503.1/05692.html

And given how better it look, the fall round of distributions (Ubuntu 14.10, Fedora 23) may very well use it. This code wont work anymore as soon as this happen. There was already a giant red warning in the API not to use it the day the original `dbus.c` was written, so this is no surprise.

As stated in the other thread, the options are LGI-GDBus, which is seriously unstable and a C based GDBus binding. As apparently, nobody (me included) is willing to dive into LGI and GObject introspection dark corners to fix the former, the later look more viable.

Plus, as you discovered already, the current implementation is badly lacking and is synchronous.

PR #206, opened 2015-04-18
Menu fixes

dodo:
Found some bugs in the awful.menu code :)

One is that layout.get_dir doesn't exist any more and the other one is proper wibox resizing when number of menu.items changes while menu is shown.


blueyed:
LGTM.

For reference, `get_dir` appears to have been removed in 9d33311.

Moving it to the "next" milestone.

PR #208, opened 2015-04-18
Add config for Travis CI

blueyed:
A fresh approach to Travis CI after / in parallel to https://github.com/awesomeWM/awesome/pull/133 (functional tests).

I would like to get this into master soon, so new pull requests would be tested automatically.


blueyed:
Should be good now.

Feedback would be appreciated!

It only tests Lua 5.2, because there is not much sense in having an extra job for Lua 5.1 currently (by default), or is there?

PR #221, opened 2015-05-12
Use table.unpack() with Lua5.2 instead of deprecated unpack()

anatol:


blueyed:
Thanks!
There are several other places using the same mechanism already.

PR #222, opened 2015-05-13
notifications (naughty) state

osleg:
Just a small function to expose suspended state for use in configurations


blueyed:
I think it's good.
Moving it to the "next" milestone, giving others the opportunity to provide feedback.

When merging, the single commit should be cherry-picked probably, especially since there's a merge commit in your PR, too.

psychon:
:+1: if we skip that merge commit.

PR #224, opened 2015-05-13
Add 'request::focus' signal, use it with autofocus

blueyed:
Follow-up of #194.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -161,11 +161,9 @@ mytasklist.buttons = awful.util.table.join(
                                           end),
                      awful.button({ }, 4, function ()
                                               awful.client.focus.byidx(1)
-                                              if client.focus then client.focus:raise() end
                                           end),
                      awful.button({ }, 5, function ()
                                               awful.client.focus.byidx(-1)
-                                              if client.focus then client.focus:raise() end
                                           end))
 
 for s = 1, screen.count() do
@@ -229,12 +227,10 @@ globalkeys = awful.util.table.join(
     awful.key({ modkey,           }, "j",
         function ()
             awful.client.focus.byidx( 1)
-            if client.focus then client.focus:raise() end
         end),
     awful.key({ modkey,           }, "k",
         function ()
             awful.client.focus.byidx(-1)
-            if client.focus then client.focus:raise() end
         end),
     awful.key({ modkey,           }, "w", function () mymainmenu:show() end),
 
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -55,7 +55,6 @@ file = {
         -- documentation
         '../lib/awful/autofocus.lua',
         '../lib/awful/dbus.lua',
-        '../lib/awful/ewmh.lua',
         '../lib/awful/init.lua',
         '../lib/awful/remote.lua',
         '../lib/awful/startup_notification.lua',
--- a/ewmh.c
+++ b/ewmh.c
@@ -430,7 +430,8 @@ ewmh_process_client_message(xcb_client_message_event_t *ev)
             lua_State *L = globalconf_get_lua_State();
             luaA_object_push(L, c);
             lua_pushstring(L, "ewmh");
-            luaA_object_emit_signal(L, -2, "request::activate", 1);
+            lua_pushboolean(L, true);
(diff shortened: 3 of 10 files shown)


Elv13:
<pre>[11:29] <Elv1313_> blueyed: ^ I was about to talk about that
[11:32] <blueyed> Elv1313_: cool. What's your opinion? Is the new PR sensible?
[11:34] <Elv1313_> What is the diffeence between request::activate and request::focus?
[11:35] [Absent] Vous n'êtes plus considéré comme absent.
[11:35] <Elv1313_> difference*
[12:17] <blueyed> activate won't raise. It basically only sets client.focus by default.
[12:34] <Elv1313_> blueyed: maybe the wording could be improved a little, I would assumed the opposite
[12:36] <blueyed> Where?
[12:36] <blueyed> In one of the commit messages?
[12:36] <Elv1313_> that request::activate doesn't raise while request::focus does
[12:37] <Elv1313_> I think the signal name should be changed
[13:19] <spyroboy> I don't agree with the names either
[13:31] <Elv1313_> do we really need that extra signal in the first place? Wouldn't it be cleaner just to update the handler to choose if the client need to be raised?
[13:32] <spyroboy> I'm not sure what it's trying to fix so no comment from me.
[13:32] <spyroboy> not sure if related but something I find fairly regularly is when I'm trying to resize or move a floating window
[13:32] <Elv1313_> they try to allow user to define very fine grained focus stealing policies
[13:33] <spyroboy> so I alt-click a window but it moves a different window that's focused
[13:33] <spyroboy> instead of automatically focusing the window I alt-clicked on
[13:34] <spyroboy> it seems to happen more often than not but sometimes it works properly.
[13:34] <Elv1313_> blame bob marley
[13:35] <Elv1313_> https://awesome.naquadah.org/wiki/Bob_Marley_Algorithm
[13:39] <spyroboy> lol
[13:39] <spyroboy> interesting
[14:25] <blueyed> Elv1313_: I think request::focus is pretty good, because it is about focus; and request::activate has been there before.
[14:27] <Elv1313_> blueyed: What about dealing with the difference in the handler?
[14:27] <blueyed> In the ewmh.activate handler?  Or in the signal emitter?
[14:27] <Elv1313_> the handler
[14:28] <blueyed> How can it differentiate, if there is only one signal?
[14:28] <Elv1313_> the hints
[14:29] <blueyed> Ah, the context. But then you end up with a list of things that should raise in there?!
[14:29] <Elv1313_> if hint == "autofocus.check_focus_tag" or .... or .... then c:raise() end
[14:29] <Elv1313_> or have an hash map with the hint as key and a boolean as value
[14:29] <blueyed> ..and then you need to handle all that when overwriting the handler.
[14:30] <Elv1313_> on the bright side, it give a single point of entry to control all that
[14:30] <blueyed> Not sure if that's a good thing.. :)
[14:31] <Elv1313_> I think it is. Those 2 signals are really the same thing with a very minor difference
[14:31] <spyroboy> honestly if it's two similar behaviors
[14:31] <spyroboy> I think the single point of entry is better
[14:31] <spyroboy> if it was 15 very distinctly different behaviors
[14:31] <spyroboy> I'd probably agree about separating them
[14:32] <Elv1313_> or add a second "raise hint" parameter to the signal
[14:35] <blueyed> That sounds better, yes.</pre>

blueyed:
I've changed it (in new commits).

I'm still not convinced really that having this in the same handler is better.
At least it complicated things in my custom handler, which now also needs to handle the case for  `raise=false`.

That's a backward-incompatible change btw then, in case somebody has a custom `request::activate` handler.

@psychon 
What's your opinion on this?

psychon:
Fine by me.

I wonder if it would make sense to have a table as arguments instead, because just now I thought thinking about adding some hint of the "the user directly targeted this client" vs "something selected this client automatically" kind so that you don't need a list of all possible arguments to the signal.

blueyed:
A table sounds good to me.
I would change it to use `{ raise = false }` then.
What's a good name for the table argument? `hints`?

blueyed:
I've changed it to use a `hints` table argument now, with only `raise` for now.

Moving this to the "next" milestone, and plan to squash-merge it in the next days.

Issue #219, opened 2015-05-10
awful.widget.prompt and menubar do not support russian (cyrillic) input


Temp24354:
When I trigger these widgets (Meta+R, Meta+P by default) and try to type text, input is always in English, even if I have another keyboard layout chosen (for example, if I use russian keyboard layout). Thus I cannot access russian-named files using awful.widget.prompt/menubar, for example.

blueyed:
#227 should fix this.
Can you try it?

Temp24354:
I have built awesome from https://github.com/petoju/awesome, and here is my report:
Now Cyrillic input works fine in all Awesome windows, but those keybinds, which require additional modkeys (such as Ctrl, Shift) stopped working with both layouts. For example, these ones:
awful.key({ modkey, "Shift"   }, "j", function () awful.client.swap.byidx(  1)    end),
awful.key({ modkey, "Shift"   }, "k", function () awful.client.swap.byidx( -1)    end),
awful.key({ modkey, "Control" }, "j", function () awful.screen.focus_relative( 1) end),
awful.key({ modkey, "Control" }, "k", function () awful.screen.focus_relative(-1) end),
while other programs' keybindings using these modkeys (such as Ctrl+F in text editor or file manager) work fine with both layouts.
I will add a similar report to https://github.com/awesomeWM/awesome/pull/227

Temp24354:
Any ideas?

petoju:
@Temp24354 this should be fixed in newer commits in #227

blueyed:
I am closing this in favor of #227.

PR #227, opened 2015-05-14
Support for XKB - changing keyboard layouts

petoju:
This commit adds support for writing in prompt (<Super>+R by default)
while different keymap is selected

Steps to reproduce the fixed bug:
1. Run `setxkbmap -option grp:switch,grp:alt_shift_toggle,grp_led:scroll us,sk`
2. Switch to Slovak keymap using alt+shift
3. Try running something in <Super>+R - you can write only garbage 

Signed-off-by: Peter Júnoš <petoju@gmail.com>


blueyed:
Thanks for cleaning up and improving this area!

I think it would fix a lot of long-standing issues, e.g. #219.

The Travis build failure is because of missing dependencies.  I've added them in a temporary branch: https://github.com/awesomeWM/awesome/compare/petoju-pr227-travis.

Temp24354:
I have built awesome from https://github.com/petoju/awesome, and here is my report:
Now Cyrillic input works fine in all Awesome windows, but those keybinds, which require additional modkeys (such as Ctrl, Shift) stopped working with both layouts. For example, these ones:
awful.key({ modkey, "Shift" }, "j", function () awful.client.swap.byidx( 1) end),
awful.key({ modkey, "Shift" }, "k", function () awful.client.swap.byidx( -1) end),
awful.key({ modkey, "Control" }, "j", function () awful.screen.focus_relative( 1) end),
awful.key({ modkey, "Control" }, "k", function () awful.screen.focus_relative(-1) end),
while other programs' keybindings using these modkeys (such as Ctrl+F in text editor or file manager) work fine with both layouts.

petoju:
Temp24354: it should be fixed now

blueyed:
:+1:

I've been trying this for a few hours now, and had no issues!

psychon:
I'd be nice if someone would compare this closely with e.g. https://github.com/i3/i3/commit/3f5a0f0024
I will do as time permits.

For example, that commit also adds a call to `xkb_x11_setup_xkb_extension`, why don't we? (And replace our own xkb initialization with this? Or not, that other commit also adds it).
That commit also re-grabs all keys when the keymap changes.

petoju:
psychon: thanks for comments. Includes are fixed.

That cheap hack is also partially fixed - we reload layout as suggested in http://xkbcommon.org/doc/current/group__x11.html - only in case we get NewKeyboardNotify or MapNotify.

blueyed:
@petoju 
Have you looked into and compared it to https://github.com/i3/i3/commit/3f5a0f0024?

petoju:
@psychon @blueyed yes - and:

> For example, that commit also adds a call to `xkb_x11_setup_xkb_extension`, why don't we?

That's a good question - mainly I wasn't rewriting code, that worked. This is still supported and provides a bit more debugging information (Xkb not supported vs not present). If such distiction is useless, I can replace current init with the other one.

> That commit also re-grabs all keys when the keymap changes.

It is not exactly like we are doing it.  We are using `xkb_x11_keymap_new_from_device()` and `xkb_x11_state_new_from_device()` like described in documentation (http://xkbcommon.org/doc/current/group__x11.html). i3 is asking for pressed keys.

On state change, they still ungrab and grab keys (as they don't use new keyboard notify), we just update the xkb state.

Furthermore, we don't care if someone uses qwerty, qwertz or azerty. We just use the basic layout (qwerty) for keybindings and a current keyboard for writing. They get all the symbols just to print there are more possible symbols to use - however, they end up with using the basic layout, like we do.

I noticed we cannot get XCB_XKB_NAMES_NOTIFY, as we don't register any event for it (this can be fixed by moving under the other call).

psychon:
I still haven't looked more closely into this, but @petoju sounds like he knows what he is doing, so :+1:

petoju:
As @psychon wrote about previous state, I looked into it and fixed bugs there.

Commit 11c99212c0e97a71bf5da31ea0bdbf1909f3dada solves the problem, when someone remaps e.g. CapsLock and Escape - Escape key worked like CapsLock and Escape together or like only Escape (should work as CapsLock). Now, it is correctly remapped also in awesome.

Commit c8bb5b6d324e560131d4b25fe68ea43cf5579dce adds unbinding, so we won't get more events after finishing.

Commit 9b5be28699d6069877c20c2ccc7bdd68d78b2a9e changes notification to NamesNotify, as the other one should never come and this was probably the idea of code - however, original author (Aleksey Fedotov) should be contacted.

And finally, in 6fcd80ad457db0cf7292816f0a9ebdf4cba3a7a6, I changed more init functions by xcb to one provided by libxkbcommon (like @psychon noted). It is really easier to read.

JKobyP:
I am running master with this patch applied, in order to fix a bug which caused my external keyboard to have trouble running the mod + "o" shortcut.

This fixed my problem, but only after commenting out the following two lines from my rc.lua:
    `--client.connect_signal("focus", function(c) c.border_color = beautiful.border_focus end)`
    `--client.connect_signal("unfocus", function(c) c.border_color = beautiful.border_normal end)`
With those lines in effect, I would get the following error every time I switched focus:

    error: /home/koby/.config/awesome/rc.lua:483: bad argument #-1 to '__newindex' (string expected, got nil)
	/home/koby/.config/awesome/rc.lua:482: in function </home/koby/.config/awesome/rc.lua:482>
	/usr/local/share/awesome/lib/awful/ewmh.lua:161: in function 'awful.ewmh.activate'
	/usr/local/share/awesome/lib/awful/autofocus.lua:27: in function </usr/local/share/awesome /lib/awful/autofocus.lua:22>
	/usr/local/share/awesome/lib/gears/timer.lua:110: in function </usr/local/share/awesome/lib/gears/timer.lua:108>

petoju:
@JKobyP can you please try awesome without this patch and test, whether changing focus works? It seems like the error is elsewhere.

blueyed:
@JKobyP 
Also, what is the contents of `/home/koby/.config/awesome/rc.lua:482`?

blueyed:
@JKobyP 
Please report back.  I plan to merge this PR soonish, and want to make sure that it causes no regressions.

JKobyP:
I'll try to report back tomorrow. Sorry for the delay.

psychon:
The changes look relatively good to me. I still don't have much clue about XKB, but I'll trust @petoju on this.
There is one minor thing (using some _checked request without using the cookie), but that's really minor ("just" a tiny memory leak) and also present in master. I made a note to eventually fix this up later myself.

JKobyP:
blueyed: the contents at those lines are as I reported. `client.connect_signal`... etc. in my previous post. 
petoju: i'm about to build without the patch, i'll report back

JKobyP:
Building from awesome-master, all the bugs I reported are the same - ie. this patch is fine, and has nothing to do with it.


I don't know if this is notable (I've never been near awesome development before, I just use it) but building from awesome-master, I notice: 
1. The menu icon in the top left is gone
2. awesome.desktop is not copied over to /usr/share/xsessions during install


I didn't want to file a bug report in case these are known issues, but I figured ya'll might want to know. Oh- and building from master fixes the original issue I was having in the first place, so I didn't even really need to use this patch! Go figure.

blueyed:
@JKobyP 
Thanks for reporting back.  This PR is fine then.

> 1. The menu icon in the top left is gone

Might be related to `2.` - e.g. if you are using a path via `/usr/`, uninstalled the distribution package, and now have it in `/usr/local`.

> 2. awesome.desktop is not copied over to /usr/share/xsessions during install

Haven't you used `/usr/local` as prefix?  Then it should be in `/usr/local/share/xsessions/awesome.desktop`.

PR #234, opened 2015-05-19
Give focus to restored clients

barlik:
When you are restoring previously minimized windows, you definetely want
focus on them.

Signed-off-by: Rastislav Barlik <barlik@zoho.com>

### DIFF
--- a/awesomerc.lua.in
+++ b/awesomerc.lua.in
@@ -262,7 +262,15 @@ globalkeys = awful.util.table.join(
     awful.key({ modkey,           }, "space", function () awful.layout.inc( 1) end),
     awful.key({ modkey, "Shift"   }, "space", function () awful.layout.inc(-1) end),
 
-    awful.key({ modkey, "Control" }, "n", awful.client.restore),
+    awful.key({ modkey, "Control" }, "n",
+              function ()
+                  c = awful.client.restore()
+                  -- Focus restored client
+                  if c then
+                      client.focus = c
+                      c:raise()
+                  end
+              end),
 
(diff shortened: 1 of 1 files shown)


syphoxy:
this is quite sensible.

Elv13:
Thanks for this

Please use the `request::activate` signal instead of giving the client "forced" focus. This allow focus stealing policy to be implemented in extensions or directly in `rc.lua`

barlik:
Why isn't then ```request::activate``` signal used in different places in ```rc.lua```. For example in mytasklistbuttons or titlebar buttons?

blueyed:
Probably because it's rather new.

On the other hand your pull request is for the (default) config, so that gives the users all the control they need already.

I'm not sure if `request::activate` should be used in the default config really - although I would prefer it myself.

barlik:
In my opinion ```request::activate``` shouldn't be in the default config. It isn't clear what it means. At least, I didn't have a clue what it meant without looking at the source code.

blueyed:
I think this PR is good as-is.

If we want to use `request::activate` in the default config, this should be something separate.

Moving to the "next" milestone, to be merged soonish.

Elv13:
I think `request::activate` should be used in the default config. That way external modules can implement the focus stealing policy for everything, including `rc.lua` without have to modify it further.

blueyed:
@Elv13
I see your point, but then it should be a separate PR still, changing all the other places, too.

blueyed:
Merged in eb7fba2.

Thanks!

PR #239, opened 2015-05-21
Make client.focusable writable from Lua

### LINKED ISSUE #237: Set window as unfocusable
Was trying to find a way to set specific window un-focusable. 
For example, since the video live wallpaper doesn't play nice with compositing managers I'm using mpv as follows ` mpv --no-input-default-bindings --no-osc --fs --on-all-workspaces <filename ` which puts mpv on all workspaces in fullscreen mode and I can use on top of it with transparency enabled.
The problem rises with the window being focusable - when i switch tags or switching between open clients the video getting focus and hiding everythin under it (which is pretty obvious). Though if it would be possible to make it unfocusable it will work "perfectly" for the use I need, plus will allow the gimmick (a bit hackish though) of live video wallpaper for awesome ;)

blueyed:
It uses an extra boolean to keep track of if `c.focusable` has been set.

It's possible to unset the overriding by setting it to `nil`, i.e.
`client.focus.focusable = nil`.

Fixes #237

### DIFF
--- a/lib/awful/client.lua.in
+++ b/lib/awful/client.lua.in
@@ -156,10 +156,12 @@ end
 --
 -- @client c The client that has been focused.
 function client.focus.history.add(c)
-    -- Remove the client if its in stack
-    client.focus.history.delete(c)
-    -- Record the client has latest focused
-    table.insert(client.data.focus, 1, c)
+    if client.focus.filter(c) then
+        -- Remove the client if its in stack
+        client.focus.history.delete(c)
+        -- Record the client has latest focused
+        table.insert(client.data.focus, 1, c)
+    end
 end
 
 --- Get the latest focused client for a screen in history.
--- a/objects/client.c
+++ b/objects/client.c
@@ -956,10 +956,10 @@ client_set_focusable(lua_State *L, int cidx, bool s)
 {
     client_t *c = luaA_checkudata(L, cidx, &client_class);
 
-    if(c->focusable != s)
+    if(c->focusable != s || !c->focusable_set)
     {
         c->focusable = s;
-        banning_need_update();
+        c->focusable_set = true;
         luaA_object_emit_signal(L, cidx, "property::focusable", 0);
     }
 }
@@ -1901,10 +1901,8 @@ luaA_client_set_focusable(lua_State *L, client_t *c)
 {
     if(lua_isnil(L, -1))
         c->focusable_set = false;
-    else {
-        c->focusable_set = true;
+    else
         client_set_focusable(L, -3, luaA_checkboolean(L, -1));
-    }
     return 0;
 }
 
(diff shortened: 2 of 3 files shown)


blueyed:
@psychon 
Thanks for your feedback!

I've addressed the issues in the fixup-commit.

psychon:
:+1:

osleg:
The client spawns unfocusable.
Everything works fine untill clicking on it. While it remains it's unfocusable status it does still focus on tag switch.

blueyed:
> While it remains it's unfocusable status it does still focus on tag switch.

What do  you mean exactly?
It seems to happen for me only, if it's the only client on that tag. Do you mean that?
(This does not appear to come from "autofocus").

osleg:
1. Spawn unfocusable client
2. Change to tag with this client only
3. Click on it with mouse
4. Change to another tag that have other clients
Expected behaviour on change: one of the focusable clients getting focus, unfocusable client losing it
What does happen: unfocusable client always in focus until you manualy switch to another client, and it intercept focus on every tag switch

blueyed:
@osleg 
I could reproduce it that way.  I've added a commit (c380da4) to address this.

osleg:
It does not intercept focus anymore but still remains focused if it was focused before tag switch

blueyed:
You mean when going back to the tag where it was forcefully focused?

osleg:
No, like this:
tag 1 = 2 urxvt clients
running mvp with `--on-all-workspaces` the rule on mvp is
```
    { rule = { instance = "mpvldb" },
          properties = { focusable = false,
                         below = true,
                         floating = true,
                         maximized = true
                        } },

```

Now going to tag[2], clicking on mvp window and going back to tag[1] - mvp 
still focused. Explicitly focusing another window and mvp losing focus for 
good.

psychon:
I think that "becomes focus" issue is basically https://awesome.naquadah.org/bugs/index.php?do=details&task_id=574 and doesn't have anything to do with this change.
AFAIK the problem is that the focus history (awful.client.focus.history) ignores awful.client.focus.filter and thus a sticky client is always "more recent" in the focus history than clients on another tag.
In the case at hand: Once the unfocusable client is in the focus history, autofocus (and other code paths) will happily focus it.

I'm not sure how exactly this should be fixed. When the other cycles through the history by hand, the "true" focus history should be used, I guess. But anything automatical should honor focus.filter.

Anyway: I don't think this is related to this changes.

blueyed:
I think the change to client.focus.history.add (https://github.com/blueyed/awesome/commit/c380da4ff156aec9ec08731e023e4864fed7fa93) is bad, but that autofocus should use focus.filter.

I will change that - but then in a separate PR, because it's really different / another fix.

blueyed:
@osleg 
Can you try if https://github.com/awesomeWM/awesome/pull/242 works as expected for you?

PR #242, opened 2015-05-25
autofocus: use focus.filter, via new arg to focus.history.get

blueyed:
This adds an optional `filter` arg to awful.client.focus.history.get,
and uses `awful.client.focus.filter` from the autofocus module.

This improves the behaviour when manually setting clients as
non-focusable (#237), and is considered to be the desired behaviour in
general.

### DIFF
--- a/lib/awful/autofocus.lua.in
+++ b/lib/awful/autofocus.lua.in
@@ -22,7 +22,7 @@ local timer = require("gears.timer")
 local function check_focus(obj)
     -- When no visible client has the focus...
     if not client.focus or not client.focus:isvisible() then
-        local c = aclient.focus.history.get(obj.screen, 0)
+        local c = aclient.focus.history.get(obj.screen, 0, aclient.focus.filter)
         if c then
             c:emit_signal('request::activate', "autofocus.check_focus")
         end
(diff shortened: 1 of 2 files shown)


osleg:
Still same behaviour as here https://github.com/awesomeWM/awesome/pull/239#issuecomment-104855736

psychon:
Change looks good to me, but I don't really know what exactly this fixes. Sticky clients will still steal focus. Hm :-/

osleg:
Its not exactly sticky client, unless I missunderstand the defenition of it.
On the other hand it works pretty well and this "issue" with it being focused on tag switch is not intrusive so might be announced as feature :) 
I mean, there is a reason why user focused unfocusable client so it might be one click/keypress away to make it unfocused/able again

blueyed:
@psychon 
This is not meant to fix the issue with the sticky clients.  See my comment to the issue in the old bugtracker for a possible solution about that: https://awesome.naquadah.org/bugs/index.php?do=details&task_id=574#comment4232

blueyed:
..but then it seems like `--on-all-workspaces` just behaves like or results in a sticky client AFAICS.

blueyed:
I've amended the commit, fixing the typo.

Given your feedback I'm moving it to the "next" target.

@osleg 
It appears that your remaining issue (or "feature" as you are beginning to call it ;)) is something more different.

Feel free to contribute to the issue in the old bugtracker (https://awesome.naquadah.org/bugs/index.php?do=details&task_id=574).

osleg:
@blueyed The issue was when the client was stealing focus on every tag change, now when the focus remains unless explicitly removed it's not an issue. Though I'm not sure what expected behaviour should be

blueyed:
@osleg 
Cool, I'll merge this then later.

Thanks for your feedback and requesting a nice new feature.. :)

PR #243, opened 2015-05-26
lua: Use pushinteger for int type

### LINKED ISSUE #238: os.setlocale(os.getenv("LANG")) in lua 5.3 with LANG = pl_PL.UTF-8 cause errors
I have arch linux - and now the latest lua version is 5.3, in lua 5.2.5 the same rc.lua have worked fine...

I find that line: os.setlocale(os.getenv("LANG"))
with:
echo $LANG
pl_PL.UTF-8 
Cause a lot of errors: https://bugs.archlinux.org/task/45008?getfile=13052

it seem similar to:
https://github.com/awesomeWM/awesome/issues/75
http://comments.gmane.org/gmane.comp.window-managers.awesome/8950

I also reported this bug on archlinux bugs: https://bugs.archlinux.org/task/45008#comment135576

Elv13:
Attempt to fix #238 

This is a contribution from someone on IRC, but using an anonymous email, so I will commit with my account. I tested on lua 5.1 and luajit. This has been reported to work in 5.3

### DIFF
--- a/mouse.c
+++ b/mouse.c
@@ -113,14 +113,14 @@ luaA_mouse_index(lua_State *L)
          * having lots of lua errors in this case.
          */
         if (globalconf.focus.client)
-            lua_pushnumber(L, screen_get_index(globalconf.focus.client->screen));
+            lua_pushinteger(L, screen_get_index(globalconf.focus.client->screen));
         else
-            lua_pushnumber(L, 1);
+            lua_pushinteger(L, 1);
         return 1;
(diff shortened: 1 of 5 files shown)


salorium:
It ok, good, thank you :+1:

psychon:
:+1:

Elv13:
So, do we merge that?

blueyed:
Thanks!

PR #246, opened 2015-05-28
Client padding

akdor1154:
implement basic spacing between windows (off by default, don't panic!). Currently this is very complicated to do completely in rc.lua, but with a couple of patches to awful.layout it's ridiculously simple.

Looks great, especially with a compositing manager!
http://www.zimagez.com/zimage/screenshot2015-05-2812-26-29.php

Jarrad


Elv13:
Hello, thanks for this. What do you think rebasing this on top of this patch?
https://github.com/awesomeWM/awesome/pull/148

This make it per-tag instead of global

PR #247, opened 2015-05-28
Fix Error while drawing widget

salorium:
I'm error :
```
Error while drawing widget: /usr/share/awesome/lib/wibox/widget/systray.lua:48: bad argument #2 to 'systray' (number has no integer representation)
```
I am looking : /usr/share/awesome/lib/wibox/widget/systray.lua:48 and i am printing x and y
```
function systray:draw(wibox, cr, width, height)
    local x, y, _, _ = lbase.rect_to_device_geometry(cr, 0, 0, width, height)
    local num_entries = capi.awesome.systray()
    local bg = beautiful.bg_systray or beautiful.bg_normal or "#000000"
    local spacing = beautiful.systray_icon_spacing or 0

    -- Figure out if the cairo context is rotated
    local dir_x, dir_y = cr:user_to_device_distance(1, 0)
    local is_rotated = abs(dir_x) < abs(dir_y)

    local in_dir, ortho, base
    if horizontal then
        in_dir, ortho = width, height
        is_rotated = not is_rotated
    else
        ortho, in_dir = width, height
    end
    if ortho * num_entries <= in_dir then
        base = ortho
    else
        base = in_dir / num_entries
    end
    print(x)
    print(y)
    capi.awesome.systray(wibox.drawin, x, y, base, is_rotated, bg, reverse, spacing)
end
```
Result :
```
948,88888888889
0.0
Error while drawing widget: /usr/share/awesome/lib/wibox/widget/systray.lua:48: bad argument #2 to 'systray' (number has no integer representation)
```
I am patching :
```
function systray:draw(wibox, cr, width, height)
    local x, y, _, _ = lbase.rect_to_device_geometry(cr, 0, 0, width, height)
    local num_entries = capi.awesome.systray()
    local bg = beautiful.bg_systray or beautiful.bg_normal or "#000000"
    local spacing = beautiful.systray_icon_spacing or 0

    -- Figure out if the cairo context is rotated
    local dir_x, dir_y = cr:user_to_device_distance(1,
    print()
    print(x)
    print(math.ceil(y))
    print(y)0)
    local is_rotated = abs(dir_x) < abs(dir_y)

    local in_dir, ortho, base
    if horizontal then
        in_dir, ortho = width, height
        is_rotated = not is_rotated
    else
        ortho, in_dir = width, height
    end
    if ortho * num_entries <= in_dir then
        base = ortho
    else
        base = in_dir / num_entries
    end
[...]

PR #250, opened 2015-05-30
rect_to_device_geometry: make sure to return numbers as ints

blueyed:
The following error was reported (in #247):

    Error while drawing widget:
    /usr/share/awesome/lib/wibox/widget/systray.lua:48: bad argument #2 to
    'systray' (number has no integer representation)

Something like "948,88" might be returned for x, which cannot be
represented as integer by Lua.

Fixes https://github.com/awesomeWM/awesome/pull/247

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,9 +64,9 @@ install:
   - sudo apt-get install -qq gir1.2-pango-1.0
 
   # Install luarocks (for the selected Lua version).
-  - wget http://keplerproject.github.io/luarocks/releases/luarocks-2.2.1.tar.gz
-  - tar xf luarocks-2.2.1.tar.gz
-  - cd luarocks-2.2.1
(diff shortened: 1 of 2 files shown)


psychon:
Uhm, why? Cairo is fine with doubles everywhere. You can draw some rectangle with top-left corner at (3.14159, 3.14159) if you want. `rect_to_device_geometry` is just returning the exact result.

I guess that the systray might be the only caller who can't handle floating point numbers, so I'd rather fix it there. Alternatively, we could change the whole widget system to force integer coordinates and sizes for widgets (this might make sense since otherwise you'd get blurry results). Alternatively, we could make layouts have some kind of "rounding mode"...

I just don't see why returning the exact result is wrong here, so I'd rather fix it elsewhere.

blueyed:
Makes sense. Closing.

PR #267, opened 2015-06-20
Make stdout/stderr line buffered

blueyed:
This should improve the behaviour with print()ing for debugging.

I was using `setbuf(…, 0)` initially, but it makes sense to buffer it
per line.

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -325,6 +325,10 @@ main(int argc, char **argv)
         { NULL,      0, NULL, 0 }
     };
 
+    /* Make stdout/stderr line buffered. */
+    setlinebuf(stdout);
+    setlinebuf(stderr);
+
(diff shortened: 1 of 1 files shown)


psychon:
Note that `setlinebuf` is a BSD extension and so might not be available everywhere. According to the man page, `setvbuf` should be used instead. However, let's wait for people to complain about this and fix it then. For now doing it like this is fine with me.

blueyed:
But we could also just use `setvbuf(stream, NULL, _IOLBF, 0)` right away then?

psychon:
But `setlinebuf` looks so much nicer and hopefully is common enough. :-) If you want to, feel free to do that (`setvbuf()` exists in C89 and C99 according to my man page).

PR #266, opened 2015-06-19
docs: move documentation to C source

blueyed:
This is a followup of https://github.com/awesomeWM/awesome/pull/177, rebased on master.

Discussion should continue there for now.

PR #269, opened 2015-06-21
Fix cmake config - typo

ideasman42:
This was giving a lot of errors using `-Wdev`

> CMake Warning (dev) at awesomeConfig.cmake:304 (configure_file):
>   configure_file called with unknown argument(s):
> 
>    ESCAPE_QUOTE

Should be `ESCAPE_QUOTES`, see:

http://www.cmake.org/cmake/help/v3.3/command/configure_file.html


blueyed:
Thanks.  I had fixed this for/in https://github.com/awesomeWM/awesome/pull/266 already: https://github.com/awesomeWM/awesome/commit/d2e25c3594df451b20bea195e3057d8221f2ce07.

Merged into master now.

PR #270, opened 2015-06-21
default config: raise with "maximize" and "fullscreen"

blueyed:
After using sloppy mouse focus to select a client,
maximizing/fullscreening it should raise it.

This also uses the new shortcut `c.maximized = not c.maximized`.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -291,7 +291,11 @@ globalkeys = awful.util.table.join(
 )
 
 clientkeys = awful.util.table.join(
-    awful.key({ modkey,           }, "f",      function (c) c.fullscreen = not c.fullscreen  end),
+    awful.key({ modkey,           }, "f",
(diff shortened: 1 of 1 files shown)


psychon:
Fine with me

Elv13:
I still think `:raise()` should not be called directly, even from `rc.lua`. But if it's just me, then I guess you can merge it anyway. The rational, like before, is that modules can implement "complete" focus stealing policies without having to modify all of `rc.lua`. But in that particular case, I don't even see why it would possibly make sense not to raise full screened clients. So beside the ideological thing, I am fine with this commit

blueyed:
Cool.

@Elv13 
I see your point (and also agree with that this one can be short-circuited), but then there should be a commit to change it for the default config in general (i.e. maybe without this instance then maybe even).

PR #272, opened 2015-06-22
Add functional tests via Xephyr/Xvfb/xdotool

blueyed:
Follow-up of https://github.com/awesomeWM/awesome/pull/133

 ~~- [ ] This might make use of `busted` (more), instead of only using the custom runner.  Needs to be investigated / refactored then.~~
 - [x] Rebase/squash on master (without https://github.com/blueyed/awesome/commit/260bd9c58db767e09193da98b61ecbb3ef6f147c).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,6 +90,9 @@ install:
   - tar xf cmake-3.2.3-Linux-x86_64.tar.Z
   - PATH=$PWD/cmake-3.2.3-Linux-x86_64/bin:$PATH
 
+  # Deps for functional tests.
+  - sudo apt-get install -qq dbus-x11 xterm xdotool xterm xvfb
+
 script:
   - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIB} -DLUA_INCLUDE_DIR=/usr/include/lua${LUAPKG}"
-  - make && sudo env PATH=$PATH make install && awesome --version
+  - make && sudo env PATH=$PATH make install && awesome --version && tests/run.sh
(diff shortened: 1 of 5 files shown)


blueyed:
Is the `tests` dir good for this?
Could/should it be merged with `spec/`?

blueyed:
Adding this to the "next" milestone: I'd really like to get this into master to better check PRs in the future.

blueyed:
@psychon 
I'd appreciate feeback on this, especially regarding the file locations and if this could be integrated / make use of busted.

psychon:
Well, I don't think this should go to spec, since it is something else than the lua tests.

I don't really want to stare too much at this. Let's just hope that it works good enough and fix it if something breaks? :-)

blueyed:
Ok, merged in 611438a.

btw: the existing tests would have caught #300 already..

blueyed:
I think there should be a target in the Makefile (via CMakeLists.txt) for this.

Could `test` be made to depend on `check` and this?
While `check` does not require awesome to be built, this test runner does - so that `make test` would trigger a build, while `make check` won't.

But then there's the warning from CMake about using `test` anyway, so another target could be used for this explicitly?

> CMake Warning (dev) at CMakeLists.txt:330 (add_custom_target):
  Policy CMP0037 is not set: Target names should not be reserved and should
  match a validity pattern.  Run "cmake --help-policy CMP0037" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

>  The target name "test" is reserved or not valid for certain CMake features,
  such as generator expressions, and may result in undefined behavior.

PR #274, opened 2015-06-22
Call `warn` in case of errors in luaA_spawn

blueyed:
This is useful to have, as it might indicate a missing program etc.


blueyed:
Any objections?

Elv13:
None

PR #278, opened 2015-07-02
widget.graph: use Cairo matrix for color pattern with `set_scale(true)`

blueyed:
A graph widget with a linear color pattern needs this pattern to be
scaled, too.  Otherwise the same value gets a different color after
scaling.

This makes `graph.draw` use an uncached pattern with a scale matrix.

Q: This might be wrong for non-linear pattern methods?!  I have not tested it.

### DIFF
--- a/lib/awful/widget/graph.lua
+++ b/lib/awful/widget/graph.lua
@@ -14,6 +14,7 @@ local table = table
 local type = type
 local color = require("gears.color")
 local base = require("wibox.widget.base")
+local cairo = require("lgi").cairo
 
 local graph = { mt = {} }
 
@@ -118,11 +119,16 @@ function graph.draw(_graph, wibox, cr, width, height)
             end
         end
     else
+        local scale = 1
         if data[_graph].scale then
+            local cfg_max_value = max_value
             for _, v in ipairs(values) do
                 if v > max_value then
                     max_value = v
                 end
+            end
+            if max_value ~= cfg_max_value then
+                scale = cfg_max_value / max_value
             end
         end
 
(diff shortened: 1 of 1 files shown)


Elv13:
+1

blueyed:
@Elv13 
Do you have widgets with non-linear patterns and could test this?

Elv13:
Yes, but they use my `Blind` library, so they can't really be used directly. I have very, very little time these days, but what exactly do you want me to test? Radial gradients on `awful.widget.graph`?

blueyed:
I could imagine that the scaling should be limited / needs to be adjusted for specific/other widget types.

psychon:
> Q: This might be wrong for non-linear pattern methods?! I have not tested it.

What do you mean with wrong? It will work the same way that it does for linear patterns, scaling it so that it covers "enough". For example, if you use an image as a "color" and keep the default max_value of 1, it will scale up so that only the first column of pixels is used.

psychon:
Oh and one thought:

    cr:scale(42, 5)
    cr:set_source(something)
    cr:scale(1/42, 1/5)

I *think* that the above will scale the pattern by 42 and 5. Looking into cairo's source code, I cannot fully confirm this, but it might be true.

This trick might help you avoid messing with the gradient's matrix.

blueyed:
Thanks for your feedback!

About the linear pattern's scaling I was not sure if the X axis should be scaled, but then it's not affected by `max_value`.

Elv13:
@blueyed I am trying to bring the number of PR down, can you either close or fix this? This is an useful change, I would still like to have it merged.

psychon:
Closing for lack of activity

PR #279, opened 2015-07-02
Useless gap v3

actionless:
follow up on https://github.com/awesomeWM/awesome/pull/148 and https://github.com/awesomeWM/awesome/pull/246

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -121,20 +121,21 @@ function layout.arrange(screen)
 
         local p = {}
         p.workarea = capi.screen[screen].workarea
+        local useless_gap = tag.getgap(tag.selected(screen))
         -- Handle padding
-        local padding = ascreen.padding(capi.screen[screen])
-        if padding then
-            p.workarea.x = p.workarea.x + (padding.left or 0)
-            p.workarea.y = p.workarea.y + (padding.top or 0)
-            p.workarea.width = p.workarea.width - ((padding.left or 0 ) + (padding.right or 0))
-            p.workarea.height = p.workarea.height - ((padding.top or 0) + (padding.bottom or 0))
-        end
+        local padding = ascreen.padding(capi.screen[screen]) or {}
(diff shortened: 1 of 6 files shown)


Elv13:
Other than that it look good, it's about time this get merged. Thanks for taking care of this, I really have no time these days.

Elv13:
The only little thing with this is that it doesn't handle the multi tag scenario, but I don't care and don't think anybody does, for as far as it goes, I am fine with this. Can you attach a screenshot? If it work well enough for you, I will merge this.

actionless:
2015-07-02-230026_3840x2160_scrot

Elv13:
Nice!

Do you think this can be merged now?

And offtopic you could add round corners to client and add some shadows, this make it look even nicer. Or even maybe only a round corner for the LCARS top left corner. But it does indeed look awesome! Another idea would be to add a thin titlebar on the top and right side to add some LCARS artwork to each clients. Just ideas like that

actionless:
mb use `beautiful.useless_gap` as a fallback here https://github.com/awesomeWM/awesome/pull/279/files#diff-ff605a1b9a61e6f8b011625708e86d70R407?
the rest LGTM

thanks for suggestions, i was thinking about some fancy titlebars

Elv13:
Yes, indeed, using `beautiful` is almost necessary for this to be useful to regular humans, however, it can be another commit, even maybe another PR

actionless:
already done

Elv13:
Thanks

Ok then, last call, if nobody else complain in the next 12 hours I will merge it when I get back to work tomorrow.

actionless:
and where `beautiful` props are documented?

Elv13:
Usually in `theme.lua`, I guess someone should write a wiki page about them. I guess you could add them `= 0` to all default themes

actionless:
>should write a wiki page 
mb better to place it somewhere in the generated docs?
i am really don't like idea of using wiki for such purpose because it's not tied to code revision

Elv13:
As promised

PR #284, opened 2015-07-04
Expose properties on awful.widget.graph

blueyed:
This allows to see if a widget is stacked for example.

Ref from IRC:

    blueyed | A widget does not know if it's stacked? (it's in the private data table).
    blueyed | Should graph.new maybe attach "_data", too?  Or provide getter methods for the properties, similar to the setters?
    blueyed | psychon: is this something your new-widgets branch addresses/conflicts with?
    Elv1313 | define "stacked". The psychon patches improve this in the sense it prevent full redraw of everything
    Elv1313 | blueyed: ^
    blueyed | I just want to know (for vivious) if a widget is stacked (w:set_stack(true)).
    blueyed | The property can only be set on a widget, but there appears to be no getter.
    Elv1313 | blueyed: The stack is only for the graph and allow multiple data to be shown in the same graph
    blueyed | btw: please consider using the "next" milestone for announcing something to be merged soonish.
    Elv1313 | ok
    blueyed | Elv1313: yeah, but the data format is different when stacking is used, and vicious needs to be aware of it.
    blueyed | (it calls "add_value", and needs to provide the group there, too)
    Elv1313 | right
    Elv1313 | I would favor a Radical like property system. It has a little overhead, but myobject.myproperty = true will call myobject:set_myproperty(true) and
            | local v = myobject.myproperty call :get_myproperty()
    Elv1313 | I have a branch somewhere to implement it for the capi objects based on one of psychon unmerged commit, but I really have 0 time for awesome/my pet
            | projets
    Elv1313 | projects*
    blueyed | I see.  I leave it then patched locally for now.  But the quick fix might be to just expose the "_data" entry with the widget.
    Elv1313 | wibox now does that for its widgets. Awful.widgets come from another age and was ported to the wibox.widget system

### DIFF
--- a/lib/awful/widget/graph.lua
+++ b/lib/awful/widget/graph.lua
@@ -225,6 +225,7 @@ for _, prop in ipairs(properties) do
     if not graph["set_" .. prop] then
         graph["set_" .. prop] = function(_graph, value)
             data[_graph][prop] = value
+            _graph[prop] = value
             _graph:emit_signal("widget::updated")
             return _graph
         end
@@ -254,6 +255,7 @@ function graph.new(args)
 
     for _, prop in ipairs(properties) do
         _graph["set_" .. prop] = graph["set_" .. prop]
+        _graph[prop] = data[_graph][prop]
     end
 
     return _graph
--- a/lib/awful/widget/graph.lua
+++ b/lib/awful/widget/graph.lua
@@ -226,10 +226,14 @@ for _, prop in ipairs(properties) do
         graph["set_" .. prop] = function(_graph, value)
             if data[_graph][prop] ~= value then
                 data[_graph][prop] = value
-                _graph[prop] = value
                 _graph:emit_signal("widget::updated")
             end
             return _graph
+        end
+    end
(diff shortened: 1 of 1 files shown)


psychon:
This just exposes the internals. If you need some API, then it makes more sense to just expose that API so that someone can later change the internals of the graph widget without breaking your code.

There were some people who used a textbox' ._text member (or something like that, I don't remember exactly) to get to something that wasn't available via the API. I guess I should have put that stuff into some really private place so that hacks like these don't work.

Alternatively, we don't need setters either. People can just set the fields directly and emit `widget::updated` on the graph when they are done. :-/

blueyed:
Makes sense to not being lazy here, yes.

Having setters with automatic signal handling makes sense though (if I understand you correctly).

blueyed:
@psychon 
I've updated the PR, and decided to use getters instead of exposing the properties (see the different commits).
Is this good like this?

psychon:
:+1: (But don't squash the wrong commits into each other :-) )

PR #285, opened 2015-07-04
Make use of the new `oldscreen` argument with property::screen handlers

blueyed:
I found this older local branch.


psychon:
:+1:

blueyed:
I wonder if `oldscreen == nil` should be handled, by updating all screens?!
It will be `nil` if there was no previous `c->screen`.

psychon:
The only case where `oldscreen` can be nil is during manage, I think (when a client gets assigned to a screen for the very first time). Also, the C code will force-set some screen for the client in that case. So IMO `oldscreen == nil` doesn't need any special treatment.

Grepping for `->screen = ' shows that only `screen_client_moveto` assigns to `c->screen` and I hope that nothing calls that function with NULL as the screen.

PR #289, opened 2015-07-05
xresources.theme.wallpaper: use max width/height from all screens

blueyed:
This creates the wallpaper surface based on the biggest dimensions, but
might screw up the format / aspect ratio?!

/cc @actionless

### DIFF
--- a/themes/xresources/assets.lua
+++ b/themes/xresources/assets.lua
@@ -60,6 +60,7 @@ function theme_assets.wallpaper(bg, fg, alt_fg, s)
             width = math.max(_s.workarea.width, width)
         end
     end
+    print("theme_assets.wallpaper: using width/height: ", width, height)
     local img = cairo.RecordingSurface(cairo.Content.COLOR,
         cairo.Rectangle { x = 0, y = 0, width = width, height = height })
     local cr = cairo.Context(img)
--- a/themes/xresources/assets.lua
+++ b/themes/xresources/assets.lua
@@ -48,9 +48,18 @@ end
 
 
 function theme_assets.wallpaper(bg, fg, alt_fg, s)
-    s = s or screen.primary
-    local height = s.workarea.height
-    local width = s.workarea.width
+    -- s = s or screen.primary
+    local height, width
+    if s then
+        height = s.workarea.height
+        width = s.workarea.width
+    else
+        width, height = 1, 1
+        for _s in screen do
+            height = math.max(_s.workarea.height, height)
+            width = math.max(_s.workarea.width, width)
+        end
+    end
     local img = cairo.RecordingSurface(cairo.Content.COLOR,
         cairo.Rectangle { x = 0, y = 0, width = width, height = height })
     local cr = cairo.Context(img)


actionless:
LGTM

psychon:
Only a quick idea: Does using a recording surface of size 1x1 work? `gears.wallpaper.maximized` should later scale this to a suitable size.

Sorry for not testing this myself and feel free to ignore this and just go with your patch.

actionless:
1x1 will work if it would be a solid fill, but it have letters. For them downscaling would be better than upscaling.

Elv13:
Or wallpaper could be an array for each screen? But that would make differencing an array of wallpapers from a gradient hard. Other than that, feel free to also ignore my comment and merge this

psychon:
@actionless Recording surfaces don't rasterize what is drawn to them. Think of them of just recording the drawing commands. So scaling a 1x1 recording surface up will not produce scaling artifacts.

(Put differently: A recording surface should look good on any resolution and should get rid of all scaling artifacts)

blueyed:
@psychon 
Interesting.
But keep in mind that `width` and `height` are also used to create/position the letters.

I've tried `width, height = 1, 1`, but it did not work: 
2015-07-06-203144_1366x768_scrot

(I've amended the commit to use `1, 1` as starting values instead of `0, 0`)

blueyed:
@Elv13 
Having a separate wallpaper for each screen would make sense, but then it cannot be in just `theme.wallpaper`.
But maybe `theme.wallpaper` could be a Lua object that returns the current version when accessed as `theme.wallpaper` and the versions per screen at `theme.wallpaper[s]`?! (is that possible?)

Otherwise the `wallpaper` function could be exported by the `xresources` module and accept a `screen` argument.

psychon:
The following almost works, but seems to misplace / misscale letters.

The reason for the problem with the letters is that the wallpaper is independent from the scale, but sadly depends on the aspect ratio of the screen. If I pretend that the screen has a ratio of 1:1, the code seems to do The Wrong Thing (tm).

(The problem seems to be the letter size which is height / 10 and ignores the width of the screen. In my version the letters are scaled in an ugly way.)

```diff
diff --git a/themes/xresources/theme.lua b/themes/xresources/theme.lua
index 589614c..e7710d8 100644
--- a/themes/xresources/theme.lua
+++ b/themes/xresources/theme.lua
@@ -153,9 +153,20 @@ theme.taglist_squares_unsel = taglist_squares_unsel()
 
 
 local function wallpaper()
+    --[[
     local height = screen[1].workarea.height
     local width = screen[1].workarea.width
     local img = cairo.ImageSurface(cairo.Format.ARGB32, width, height)
+    --]]
+    ---[[
+    local width, height = 1, 1
+    local rect = cairo.Rectangle()
+    rect.x = 0
+    rect.y = 0
+    rect.width = 1
+    rect.height = 1
+    local img = cairo.RecordingSurface.create(cairo.Content.COLOR, rect)
+    --]]
     local cr = cairo.Context(img)
 
     local bg = xrdb.color8
```

Perhaps we should still use a RecordingSurface since it likely needs less memory than a full image surface? (Only having the drawing commands in memory vs the value of each pixel)

blueyed:
> depends on the aspect ratio of the screen

Yes, that's what I've meant.

So allowing to call it dynamically makes sense (if you want to have it proper for all your screens), and then the default could use not `1x1` but the most common ratio across screens (or just the first one), e.g. `4x3`?!

Elv13:
Please update or close. We can't have PR open like this forever.

themes/xresources/assets.lua
```diff
@@ -48,9 +48,18 @@ end
 
 
 function theme_assets.wallpaper(bg, fg, alt_fg, s)
-    s = s or screen.primary
-    local height = s.workarea.height
-    local width = s.workarea.width
```
psychon: By the way, why is this using the workarea instead of the geometry of the screen? That seems wrong to me.

psychon:
Closing for lack of activity

PR #293, opened 2015-07-07
awful.tag: emit property:: signals only on change

blueyed:
This can have a significant performance impact in case you listen to the
property::hide signal to auto-hide tags for example.

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -937,11 +937,13 @@ function client.property.set(c, prop, value)
     if not client.data.properties[c] then
         client.data.properties[c] = {}
     end
-    if client.data.persistent_properties_registered[prop] then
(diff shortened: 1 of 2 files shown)


Elv13:
+1

psychon:
Sorry, I don't get the commit message. What's messing with the "hide" property so much? As far as I see, nothing in awesome ever sets it (we don't even provide a setter for it).

The patch itself is obviously correct, but while you are at it, you should also fix this in `awful.client.property.set`.

blueyed:
@psychon 
I've added a new commit for `awful.client.property.set`.

As for the "hide" property: I am setting it myself (to autohide empty, non-current tags), and it gets respected/queried in `lib/awful/widget/taglist.lua`

blueyed:
Merged in bd885f59 and ae7c7ff3.

PR #294, opened 2015-07-07
Add client.first_tag, as a shortcut for `c:tags()[1]`

blueyed:
This is meant to be a faster alternative in case only the first tag is
relevant/used.

I've recognized this to be slowish in a plugin.
Maybe it (`c:tags()`) could (also or instead) be optimized in general, by using some internal data structures (instead of the `for` loops).

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -141,7 +141,7 @@ mytasklist.buttons = awful.util.table.join(
                                                   -- :isvisible() makes no sense
                                                   c.minimized = false
                                                   if not c:isvisible() then
-                                                      awful.tag.viewonly(c:tags()[1])
+                                                      awful.tag.viewonly(c:first_tag())
                                                   end
                                                   -- This will also un-minimize
                                                   -- the client, if needed
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -141,7 +141,7 @@ mytasklist.buttons = awful.util.table.join(
                                                   -- :isvisible() makes no sense
                                                   c.minimized = false
                                                   if not c:isvisible() then
-                                                      awful.tag.viewonly(c:first_tag())
+                                                      awful.tag.viewonly(c.first_tag)
                                                   end
                                                   -- This will also un-minimize
                                                   -- the client, if needed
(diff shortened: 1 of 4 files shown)


Elv13:
Your right that `:tags()` is called way too often. I always wondered if it would be a good idea to have a chained list build right into the client (CAPI) object or cache them on the lua side. `:tags()` is usually one of the top 3 worst offender when profiling Awesome. With my 5 screens, I usually have 50+ tags, that make a lot of looping.

Again, I have no time to fix Awesome right now and don't want to put additional load on others for the sake of it. I propose we merge this as is and fix it eventually. So +1

Elv13:
Thinking about it, I guess it should be a property, not a method.

blueyed:
> it should be a property, not a method

Makes sense, changed.

Elv13:
To be pedantic, the commit messages should be updated ;)

But then again, look fine to me, go ahead and merge it!

blueyed:
Thanks for your feedback! :)

I've planned to squash the commits anyway.

I'll wait for some more feedback - no need to rush this.

psychon:
Sorry to always(?) stop the optimizations you two are doing, but I don't get it. What exactly is slow here? I can't really imaging that calling `c:tags()` is slow for 50 tags. The C code just has to count from 0 to 50 and do "a bit" with each entry, and lua surely isn't slow when creating a table with 10 entries either.

How exactly did you figure out that this is slow?

    $ echo 'for i=1, 1000000 do client.focus:tags() end' | time awesome-client - 
    awesome-client -  0,00s user 0,00s system 0% cpu 1,552 total

This is with 2 screens with 10 tags each and the currently focused terminal is only on a single tag. That's 1.5 micoseconds per iteration (or 0.0015 milliseconds).

After tagging the terminal with all tags of one of the screens:

    $ echo 'for i=1, 1000000 do client.focus:tags() end' | time awesome-client - 
    awesome-client -  0,00s user 0,00s system 0% cpu 4,623 total

Let's claim that this takes 5 times as long as before. That's still insignificantly fast.

I'm not saying that this change makes no sense, I just don't see why this needs improvement.
(And no, my laptop isn't super-fast).

blueyed:
> Let's claim that this takes 5 times as long as before. That's still insignificantly fast.

In my benchmark/test it was "only" twice as fast.

And you're right of course, that it's still fast - but there's no reason for it to be not faster.. :)

I've came across this when debugging where my cyclefocus plugin spends most of its time.

PR #296, opened 2015-07-07
tasklist/taglist: call update method only once per main loop

blueyed:
Currently `tasklist_update` gets triggered often, because it listens to
a lot of properties.
This patch makes it only call the last one through `timer.delayed_call`.


Elv13:
Look good to me

psychon:
Fine by me, but note that this might mean the tasklist "misses" the next redraw of the wibox and then cause another redraw. Do we need priorities between delayed calls? :-)

blueyed:
Merged in 5630ad1 and 1924e04.
I've keep it separate to help with bisecting in case of problems.

PR #297, opened 2015-07-07
awful.tag.delete: single call to #c:tags()

blueyed:
This is a minor performance optimisation I came across.


psychon:
Feels like premature optimization to me, but doesn't really hurt here, so feel free.

PR #304, opened 2015-07-12
Enhanced awesome-client

blueyed:
It now exits with a non-zero status code when `dbus-send` fails.

I've also changed the shebang to use `bash` directly, which simplifies the logic to detect/handle "bash as /bin/sh" and allows to use `set -o pipefail` (which makes handling the D-Bus error easier - without using a subshell).

It also fixes the prompt:
 - trailing whitespace
 - prompt is actually used (which wasn't the case with /bin/sh not pointing at bash)

The first and third commits will be squashed, but I've left them as-is for now, to see the difference - and in case `/bin/sh` should be kept - which I could understand when `awesome-client` is meant to be used often.

### DIFF
--- a/utils/awesome-client
+++ b/utils/awesome-client
@@ -1,13 +1,12 @@
-#!/bin/sh
-
-READ_OPTIONS="-r"
+#!/bin/bash
 
 tty 2>&1 > /dev/null
 ISATTY=$?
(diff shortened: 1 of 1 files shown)


psychon:
Looks good

PR #306, opened 2015-07-12
naughty: skip permanent notifications when there is no more room

blueyed:
This skips permanent notifications (with timeout=0) in `get_offset` when
there is not enough room for a new notification.  It will still fallback
to removing the first/oldest one.

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -209,10 +209,23 @@ local function get_offset(screen, position, idx, width, height)
         v.y = ws.y + ws.height - (naughty.config.padding + height + existing)
     end
 
+    -- Find old notification to replace in case there's not enough room.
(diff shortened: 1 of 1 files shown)


lib/naughty/core.lua
```diff
@@ -209,10 +209,24 @@ local function get_offset(screen, position, idx, width, height)
         v.y = ws.y + ws.height - (naughty.config.padding + height + existing)
     end
 
+    -- Find old notification to replace in case there is not enough room.
+    -- This tries to skip permanent notifications (without a timeout),
+    -- e.g. critical ones.
+    local find_old_to_replace = function()
+        for i = 1, idx-1, 1 do
```
psychon: Why the last `, 1`?

blueyed: Copied from above.. :)
Will remove it, since it's the default.

PR #308, opened 2015-07-13
Add beautiful.get_merged_font; fix doc / load_font

blueyed:
This is meant to get a new font with adjusted attributes, e.g.

    font_focus = beautiful.get_font_copy(theme.font, "bold")

### DIFF
--- a/lib/beautiful/init.lua
+++ b/lib/beautiful/init.lua
@@ -37,13 +37,20 @@ local descs = setmetatable({}, { __mode = 'k' })
 local fonts = setmetatable({}, { __mode = 'v' })
 local active_font
 
---- Load a font name
+--- Load a font from a string or a font description.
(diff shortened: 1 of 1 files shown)


psychon:
This doesn't actually add the function that the commit message claims is added. I don't have any other complaints.

blueyed:
@psychon 
Good catch.. :)
I've renamed it after the fact.  Will add it as `get_font_copy` then, OK?

psychon:
I liked the "merge" in the name more, because the "copy stuff" doesn't matter that much IMO. For me it seems more important that this adds some more "properties" to a font.

blueyed:
Oh sorry, I've meant to use the `get_merged_font` actually (only the Github desc was off).
Will use it as this then.

Issue #309, opened 2015-07-13
Regression with floating.mouse_resize_handler: button needs to be hold


blueyed:
https://github.com/awesomeWM/awesome/commit/18f6ab107f1b57dda947386c8c074a604ce08244 changes the behavior when resizing floating clients using the mouse (via `modkey + RMB`).

Previously, you did not have to keep the mouse button pressed during resizing, but now you'll have to.
Now the resizing gets aborted as soon as you release the mouse button.

This does not seem to be intentional?!

blueyed:
Closing this in favor of #310, which fixes this.

PR #310, opened 2015-07-13
Restore old mousegrabber behavior

blueyed:
18f6ab1 changed the behavior when resizing floating clients using the
mouse (via modkey + RMB).

Previously, you could initiate the mousegrabber using e.g. "modkey +
RMB", release the key and button, and resize the window using the left
mouse button.

This restores this behavior by canceling the mousegrabbers only if the
cursor was moved, without any mouse buttong being pressed.

Fixes https://github.com/awesomeWM/awesome/issues/309

### DIFF
--- a/lib/awful/layout/suit/floating.lua
+++ b/lib/awful/layout/suit/floating.lua
@@ -28,10 +28,12 @@ function floating.mouse_resize_handler(c, corner, x, y)
     -- Warp mouse pointer
     capi.mouse.coords({ x = x, y = y })
 
+    local prev_coords = {}
     capi.mousegrabber.run(function (_mouse)
                               for k, v in ipairs(_mouse.buttons) do
                                   if v then
                                       local ng
+                                      prev_coords = { x =_mouse.x, y = _mouse.y }
                                       if corner == "bottom_right" then
                                           ng = { width = _mouse.x - g.x,
                                                  height = _mouse.y - g.y }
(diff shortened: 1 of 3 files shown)


psychon:
Sorry, but why is the old behavior better?

Elv13:
I also think is make more sense to just hold the button. But I also think the resizing code should be totally decoupled from the mousegrabber and available without it so user could implement whatever they want with it (but that's another story).

blueyed:
> but why is the old behavior better?

It allows to initiate the grabber, and then use the primary button while resizing.  This is more comfortable - at least for me.

> I also think is make more sense to just hold the button

That's also (still) possible of course.

> the resizing code should be totally decoupled from the mousegrabber and available without it so user could implement whatever they want with it (but that's another story)

:+1:

syphoxy:
is it possible that we can make the current behavior default but the legacy behavior easily accessible? like a configuration boolean? I'm all for options.

blueyed:
Is there anything bad about keeping the old behavior, i.e. applying this patch/PR?

syphoxy:
I could be wrong but I believe majority vote appears to favour the new functionality over the old functionality.

psychon:
Intuitively, I think the new behavior makes more sense, but feel free to merge this

actionless:
IMHO, it should be optional, i like current master behavior more

blueyed:
> is that what mouse cursor is jumping to the corner is by design or a bug?

It's by design, but I agree that it would be better to not do this.

syphoxy:
I also agree that it shouldn't jump to the corner. I can open an separate issue for this if you guys agree that we should change it.

blueyed:
> I can open an separate issue for this if you guys agree that we should change it.

:+1:

blueyed:
I will merge this later - given that it does no harm and provides the old behaviour only additionally, which I like myself.

This could get enhanced/refactored later still.

Issue #301, opened 2015-07-10
Error during keyhandler results in "(null)" error message/info


blueyed:
If there's an `error` during the keyhandler, only the following appears in the log:

> W: awesome: event_handle_key:607: error running function: (null)

This gets thrown through:

    function textbox:set_markup(text)
        local attr, parsed = Pango.parse_markup(text, -1, 0)
        -- In case of error, attr is false and parsed is an error message
        if not attr then error(parsed) end

        self._layout.text = parsed
        self._layout.attributes = attr
        self:emit_signal("widget::updated")
    end

While this works:

    keygrabber.run(function(mod, key, event) error("foo") end)

it can be triggered using:

    keygrabber.run(function(mod, key, event) require("wibox").widget.textbox():set_markup("&") end)

The main problem here is that the error (`parsed`) is "userdata" and not a string [1].

However, awesome should handle the error from its `debug::error` handler better, and not just say "(null)".

The following shows the real error:

    require("wibox").widget.textbox():set_markup("&")

    W: awesome: luaA_dofunction:77: error while running function
    stack traceback:
        [C]: in ?
        [C]: in function 'error'
        /usr/local/share/awesome/lib/wibox/widget/textbox.lua:63: in function 'set_markup'
        [string "require("wibox").widget.textbox():set_markup(..."]:1: in function 'f'
        /usr/local/share/awesome/lib/awful/remote.lua:25: in function </usr/local/share/awesome/lib/awful/remote.lua:21>
    error: attempt to concatenate a userdata value

1: https://github.com/pavouk/lgi/issues/115

blueyed:
The error triggering this is fixed (https://github.com/awesomeWM/awesome/commit/d4c7a7ce04db22ff8444febe633a536f198b3056), but it seems like there's another issue, so that this would have been handled better, i.e. there should be no "(null)" error message, but a more specific one.

blueyed:
btw: would it be possible to have a Lua traceback in this case?
I've tried experimenting with http://www.lua.org/source/5.2/lauxlib.c.html#luaL_traceback, but the traceback is empty.  In `event_handle_key`:

            if(lua_pcall(L, 3, 1, 0))
            {
                const char *msg;
                msg = lua_pushfstring(L,
                        "error running function: %s. Stopping keygrabber.",
                        lua_tostring(L, -1));
                luaL_traceback(L, L, msg, 1);
                warn("%s", lua_tostring(L, -1));
[...]

PR #311, opened 2015-07-13
Better error handling via luaA_dofunction and xpcall

blueyed:
This does not change all places where `pcall` is being used currently, but it's a good start.

Ref: https://github.com/awesomeWM/awesome/issues/301


psychon:
:+1: (except for one minor comment, see above)

PR #312, opened 2015-07-13
widget.common.list_update: improve margins (dpi / imagebox)

blueyed:
Use `beautiful.xresources.apply_dpi` for `common.list_update`'s textbox
margins, and add a left margin for the imagebox, too.

The latter is useful/nicer with the default behavior of an inverted
style for the focused tasklist entry.

### DIFF
--- a/lib/awful/menu.lua
+++ b/lib/awful/menu.lua
@@ -15,6 +15,7 @@ local util = require("awful.util")
 local tags = require("awful.tag")
 local keygrabber = require("awful.keygrabber")
 local beautiful = require("beautiful")
+local dpi = require("beautiful").xresources.apply_dpi
 local object = require("gears.object")
 local surface = require("gears.surface")
 local cairo = require("lgi").cairo
@@ -546,13 +547,13 @@ function menu.entry(parent, args)
         end
         iconbox = wibox.widget.imagebox()
         if iconbox:set_image(icon) then
-            margin:set_left(2)
+            margin:set_left(dpi(2))
         else
             iconbox = nil
         end
     end
     if not iconbox then
-        margin:set_left(args.theme.height + 2)
+        margin:set_left(args.theme.height + dpi(2))
     end
     -- Create the submenu icon widget
     local submenu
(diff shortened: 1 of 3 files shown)


actionless:
looks good to me, also i remember it was a similar problem somewhere else, mb in menu

blueyed:
@actionless 
Added it to naughty and awful.menu.

But it's problematic, at least with Naughty - see #320.

blueyed:
Rebased on master, moving to the "next" milestone.

#320 is not really a direct issue anymore (after b28cda2).

actionless:
also i think it would be better to do like:
```lua
padding = beautiful.common_list_padding or dpi(2)
```
and so on

blueyed:
@actionless 

> beautiful.common_list_padding

Do you mean to add new intermediate / base settings?
Not too convinced about the name then: a list of naughty notifications is quite different from a list in the menubar for example..

actionless:
i was not meant any particular name but the ide a itself to have it configurable from the theme

blueyed:
Agreed. Something like `beautiful.padding_box` maybe?

actionless:
i am not sure if it's obvious, mb prefix settings with the module name?

blueyed:
Ah, you mean to have settings for modules in `beautiful`, not some basic setting that could be used across modules?

actionless:
no, just settings, but prefixed with the target name, ie `beautiful.common_list_padding_box` instead of `beautiful.padding_box` and so on

blueyed:
But `common_list_padding_box` could then fall back / use `padding_box`.
`common_list` is maybe not a really good example, given that it should not have specific settings probably.

actionless:
ah, i hope i've got it now, you mean to provide fallback value for missing theme variables in beautiful?

blueyed:
Yes.

actionless:
yeah, that would be great, and will help to cut few `or`s from the widgets

blueyed:
@actionless 
I've merged this PR now, without any more changes (except for doc).

You seem to have understood it like `beautiful` could have some index metamethods to handle non-existent settings?  That's not what I've meant initially, but might be a good idea.

We should have a separate issue to discuss this further probably.

actionless:
definitely! sounds very good

also mb it would be possible to have in beautiful sort of references/links, for example:
```lua
theme.widget_bg = "$theme.bg"
```
so the value will be resolved on call, ie if `theme.bg` will changed, next time it will be resolved to the new value

PR #313, opened 2015-07-13
RFC: awful.widget.common.list_update: expose textbox to label callback

blueyed:
This allows the "label" callback to adjust the textbox itself.
`tasklist_label` is changed to make use of it and supports a new style
argument `font_focus`.

I am not sure if that is the way to go - probably a table should be used
instead, allowing to enhance this more easily in the future.  For now, I
could also imagine to expose/pass on the imagebox for the icon.
Also, the taglist widget should get adjusted for this API then, too.

Together with #308 this allows a theme to easily configure the font to be bold for the focused tasklist entry.

### DIFF
--- a/lib/awful/widget/common.lua
+++ b/lib/awful/widget/common.lua
@@ -91,7 +91,7 @@ function common.list_update(w, buttons, label, data, objects)
             }
         end
 
-        local text, bg, bg_image, icon = label(o)
+        local text, bg, bg_image, icon = label(o, tb)
         -- The text might be invalid, so use pcall.
         if text == nil or text == "" then
             tbm:set_margins(0)
--- a/lib/awful/widget/tasklist.lua
+++ b/lib/awful/widget/tasklist.lua
@@ -26,7 +26,7 @@ local tasklist = { mt = {} }
 -- Public structures
 tasklist.filter = {}
 
-local function tasklist_label(c, args)
+local function tasklist_label(c, args, tb)
     if not args then args = {} end
     local theme = beautiful.get()
     local fg_normal = util.ensure_pango_color(args.fg_normal or theme.tasklist_fg_normal or theme.fg_normal, "white")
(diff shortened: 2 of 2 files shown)


blueyed:
Additionally `font_urgent` and `font_minimize` should get added.

Elv13:
/semi offtopic

It is already possible to change the font, but it's  very hacky. If you set the `bg_image` to be a function. It will give you access to the widget internal. The way `Radical` handle this is by having delegate objects for each "layer" of a menu/list_widget. You can set the widget container (standard widget, floating wibox, centered wibox), the item container (vertical menu, horizontal, radial) and the item renderer itself. This avoid adding more and more code in a centralized widget like `awful.widget.tasklist`.

I am not saying `Radical` is the perfect solution. All those abstractions and layer make it rather to understand unless you wrote it (aka, only God and me know how it works). But it allow to add new stuff like that almost without code.

I wrote the original `bg_image` hack to allow users to override the behavior and at some point used it, but I gave up on both adding more stuff into tasklist.lua/common.lua and adding other hacks to bypass it. I think the widgets/common.lua design is just a dead end.

blueyed:
@Elv13 
Interesting.. I've discovered the possibility to use `bg_image`, too - but wasn't aware that it was meant to be a hack, and then rather not wanted to "abuse" it.

Using callbacks / delegates seems to be useful, but then that's what `label` is about already - and is a bit confusing at first.

In this case we could maybe just expose `data` or `data[o]` to the label callback? (and then use more explicit names, like `textbox` instead of `tb`).

Elv13:
```
Using callbacks / delegates seems to be useful, but then that's what label is about already - and is a bit confusing at first.
```

But the label never reach the user, it is used internally by `tag/tasklist.lua` and that's the end of the road. It is why I think the current design of those widget is not flexible enough. Any changes, even minor, require to [monkey]patch the widget

blueyed:
Rebased on master, and added `font_urgent` and `font_minimized`.

actionless:
cool, also it made me think about another feature -- mb to add `:set_fg()` method to textbox to get rid of some direct pango markup usage?

actionless:
> But the label never reach the user, it is used internally by tag/tasklist.lua and that's the end of the road. It is why I think the current design of those widget is not flexible enough. Any changes, even minor, require to [monkey]patch the widget

i think the easier solution will just to expose `label` [1] as `tasklist.label`

[1] https://github.com/awesomeWM/awesome/blob/master/lib/awful/widget/tasklist.lua#L29

blueyed:
I am merging this as-is for now.
Please create a new issue or PR for the mentioned/discussed improvements.

PR #315, opened 2015-07-14
client: do not assume that wfact exists

Elv13:
Since the call can now be delayed, "windowfact" may not exist yet.

Regression from 74106462ac1f0877d4dd3791c9a983283c034169


blueyed:
What's needed to trigger this?
Are other properties affected in a similar way?

Elv13:
When doing `ctrl+mod4+r` to reload awesome, the "manage" signal will be sent before the layout is created. So if there is a rule that set `wfact`, it will run before the array is created.

blueyed:
Without having reproduced it: :+1:

PR #316, opened 2015-07-14
Improve tests/run.sh

blueyed:
- `set -x` only for Travis/CI=true.
 - improve waiting for PID and awesome-client success.
 - kill childs via trap always.


psychon:
Fine by me, but I don't claim to understand this stuff.

PR #319, opened 2015-07-15
Handle fit(-1, -1) more consistent / use base.fit_widget / use margin's fit()

blueyed:


psychon:
Did I mention that I don't like this -1 hack and would rather remove it?

blueyed:
> Did I mention that I don't like this -1 hack and would rather remove it?

Got it.. :)

psychon:
Do you agree that #535 seems saner? Can this be closed?

blueyed:
Yes, thanks!

PR #320, opened 2015-07-15
Drawin: handle non-integer margins / geometry

blueyed:
This PR is meant for discussing how to address the issue described in the 2nd commit (https://github.com/awesomeWM/awesome/commit/c49165bf7d50bb30f12a673ebb3a62edb9fdc88b).

Setting a non-integer height of e.g. 1.42 will cast it to 1 (in `area_t`).

This is a problem when using a marginbox with a textbox widget and a
non-integer margin in a fixed layout: the fixed layout won't draw the
text.  This can be reproduced by using a non-integer margin with
`naughty.notify`.

It seems that it is problematic in general to have floats in Lua, but
integers in C for the drawin geometries.  An alternative might be to use
floats in `area_t`, too?

The first commit changes `margin:fit` to use `ceil`, while the second one does so in `drawin.c`.

Related comment about Cairo being fine about doubles everywhere from @psychon: https://github.com/awesomeWM/awesome/pull/250#issuecomment-107004468

### DIFF
--- a/lib/awful/util.lua
+++ b/lib/awful/util.lua
@@ -27,6 +27,7 @@ local capi =
     awesome = awesome,
     mouse = mouse
 }
+local floor = math.floor
 
 local util = {}
 util.table = {}
@@ -524,6 +525,13 @@ function util.query_to_pattern(q)
                                              string.upper(c))
                     end)
     return s
+end
+
+--- Round a number to an integer.
+-- @tparam number x
+-- @treturn integer
+function util.round(x)
+    return floor(x + 0.5)
 end
 
 return util
--- a/lib/awful/wibox.lua
+++ b/lib/awful/wibox.lua
@@ -23,6 +23,7 @@ local table = table
 local error = error
 local wibox = require("wibox")
 local beautiful = require("beautiful")
+local round = require("awful.util").round
 
 local awfulwibox = { mt = {} }
 
@@ -154,7 +155,7 @@ function awfulwibox.align(wibox, align, screen)
         elseif align == "left" then
             wibox.y = area.y + area.height - (wibox.height + 2 * wibox.border_width)
         elseif align == "center" then
-            wibox.y = area.y + (area.height - wibox.height) / 2
+            wibox.y = area.y + round((area.height - wibox.height) / 2)
         end
     elseif position == "left" then
         if align == "right" then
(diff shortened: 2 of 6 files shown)


blueyed:
I've reverted the `ceil()` in C and added integer checking instead.
This required fixing two more callers (awful.menu and awful.wibox).

In #lua on IRC they said "to use `lua_tointegerx` and verify it yourself", but after what I had come up with - so this might get simplified / changed.  For Lua 5.3+ we might also use Lua's version instead?

To not break too much, we could also use a warning instead:

    luaA_warn(L, "Expected integer, but got float (%f).", d);
    d = ceil(d);

And when keeping the error, it should have an improved message, like the one from the warning above.

blueyed:
This also affects `border_width`.  Using a dpi of 94 will result in a `beautiful.border_width` being less than 1 and `c.border_width` becoming 1.
I've created #330 to address this.

Of course, I've thought about using `math.ceil` directly in `apply_dpi`, but I think it's good to get a float from there.  Opinions?

blueyed:
Cleaned up after b28cda2 is in master now.

I've left the two different approaches here for now - with the revert of the first one in the middle, awaiting feedback.

blueyed:
@psychon 
Can you please review this?

psychon:
The change looks good, but the CI failure looked weird. I rebased this onto master and pushed a psychon_is_experimenting branch to get a new CI build without the `expr 1-1` problem. It also failed: https://travis-ci.org/awesomeWM/awesome/builds/75225369
```
error while running function
stack traceback:
	[C]: in function 'drawin'
	lib/wibox/init.lua:109: in function <lib/wibox/init.lua:107>
	(...tail calls...)
	lib/awful/wibox.lua:253: in function <lib/awful/wibox.lua:214>
	(...tail calls...)
	...vis/build/awesomeWM/awesome/tests/../build/awesomerc.lua:187: in main chunk
error: lib/wibox/init.lua:109: bad argument #-1 to 'drawin' (integer expected, got number)
```
I guess `awful.wibox` needs to round the height/width that it calculates.

blueyed:
Yes, the values in `awful.wibox.new` need to be rounded.

Should this be fixed there, in `wibox.init`, or in `drawin`?

psychon:
In my opinion `awful.wibox.new` should do the rounding. It also does the multiplication with 1.5.

blueyed:
Added the rounding, via a new `awful.util.round`.

psychon:
:+1: (I guess I can live with the dependency between the flex layout and awful.util)

PR #321, opened 2015-07-15
awful.tooltip: wrap it in a marginbox, using a margin of dpi(2)

blueyed:
NOTE: this is affected by #320!


Elv13:
Look good to me

blueyed:
Fixed it to use `math.ceil()` (because of #320) and made the margin configurable.

psychon:
Fine by me

PR #322, opened 2015-07-15
awful.widget.layoutbox: add tooltip with the layout's name

blueyed:


Elv13:
Look good to me

psychon:
I haven't actually tested this, but I think the tooltip (actually: The textbox) will react badly on `:set_text(nil)`. Perhaps the "unknown name" case should be handled specially? (And perhaps it makes sense to also add some handling for the "couldn't find a valid image for this layout"-case)

Besides that: :+1:

blueyed:
> will react badly on `:set_text(nil)`

Indeed.  Changed it to use `[no name]` in that case.

> (And perhaps it makes sense to also add some handling for the "couldn't find a valid image for this layout"-case)

Probably, but I've left it as-is for now.

Issue #298, opened 2015-07-08
Add "grow_master" parameter to tag/layout?


actionless:
Here is a small change to standard tile layout I am using in my config, it's somehow similar to lain's centerfair/centerwork but with the simpler logic.

![myimage](https://cloud.githubusercontent.com/assets/1655669/8569861/09e45f02-257c-11e5-9b26-549f41bdbe6c.gif)

The question, if idea is looking good, should i add this to all the layouts or try to find a way to generalize it?

```patch
--- /usr/share/awesome/lib/awful/layout/suit/tile.lua   2015-07-05 15:54:46.000000000 +0200
+++ actionless/center_layout.lua        2015-07-08 13:27:27.364621634 +0200
@@ -216,13 +216,17 @@
         place_master = false
     end

+    local grow_master = false -- will use tag property for that
     -- this was easier than writing functions because there is a lot of data we need
     for d = 1,2 do
         if place_master and nmaster > 0 then
             local size = wa[width]
-            if nother > 0 then
+            if nother > 0 or not grow_master then
                 size = math.min(wa[width] * mwfact, wa[width] - (coord - wa[x]))
             end
+            if nother == 0 and not grow_master then
+              coord = coord + (wa[width] - size)/2
+            end
             if not data[0] then
                 data[0] = {}
             end
@@ -255,7 +259,7 @@
 end

 tile.right = {}
-tile.right.name = "tile"
+tile.right.name = "center"
 tile.right.arrange = do_tile
 function tile.right.mouse_resize_handler(c, corner, x, y)
     return mouse_resize_handler(c, corner, x, y)


```

PR #323, opened 2015-07-16
Add "master_fill_policy" tag property and its support in "tile" and "corner" layouts

actionless:
Follow-up on https://github.com/awesomeWM/awesome/issues/298

myimage

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -409,6 +409,36 @@ function tag.getgap(t)
     return tag.getproperty(t, "useless_gap") or beautiful.useless_gap or 0
 end
 
+--- Set sizing policy for the master client(s)
+-- @tparam string policy Can be set to
+-- "expand" (fill all the available workarea) or
+-- "mwfact" (fill only an area inside the master width factor)
+-- @tparam tag t The tag to modify, if null tag.selected() is used.
+function tag.setmfpol(enabled, t)
+    local t = t or tag.selected()
+    tag.setproperty(t, "master_fill_policy", enabled)
+end
+
+--- Toggle sizing policy for the master client(s)
+-- @tparam tag t The tag to modify, if null tag.selected() is used.
+function tag.togglemfpol(t)
+    if tag.getmfpol(t) == "expand" then
+        tag.setmfpol("mwfact", t)
+    else
+        tag.setmfpol("expand", t)
+    end
+end
+
+--- Get sizing policy for the master client(s)
+-- @tparam tag t Optional tag.
+-- @treturn string Can be set to
+-- "expand" (fill all the available workarea) or
+-- "mwfact" (fill only an area inside the master width factor)
+function tag.getmfpol(t)
+    local t = t or tag.selected()
+    return tag.getproperty(t, "master_fill_policy") or "expand"
+end
+
 --- Set the number of master windows.
 -- @param nmaster The number of master windows.
 -- @param[opt] t The tag.
@@ -702,6 +732,7 @@ capi.tag.add_signal("property::icon_only")
 capi.tag.add_signal("property::layout")
 capi.tag.add_signal("property::mwfact")
 capi.tag.add_signal("property::useless_gap")
+capi.tag.add_signal("property::master_fill_policy")
 capi.tag.add_signal("property::ncol")
 capi.tag.add_signal("property::nmaster")
 capi.tag.add_signal("property::windowfact")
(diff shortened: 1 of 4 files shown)


actionless:
i am not sure what it can be applied nicely to the rest of layouts

actionless:
i just came up this the idea how it can be generalized:
instead of `arrange` of layout [1] call custom "arranger" if :
1) there are only master client(s) on tag
AND
2) master_fill_policy set to mwfact
AND
3) layout is not `floating` or `max`

but i'm not sure if it's better than now

1) https://github.com/awesomeWM/awesome/blob/master/lib/awful/layout/init.lua#L138

lib/awful/tag.lua
```diff
@@ -409,6 +409,36 @@ function tag.getgap(t)
     return tag.getproperty(t, "useless_gap") or beautiful.useless_gap or 0
 end
 
+--- Set size fill policy for the master client(s)
+-- @tparam string policy Can be set to
+-- "expand" (fill all the available workarea) or
+-- "mwfact" (fill only an area inside the master width factor)
+-- @tparam tag t The tag to modify, if null tag.selected() is used.
```
blueyed: You could write this as `@tparam[opt=tag.selected()] tag The tag to modify`.

JFI, it's OK like it is.

lib/awful/tag.lua
```diff
@@ -409,6 +409,36 @@ function tag.getgap(t)
...
+function tag.setmfpol(enabled, t)
+    local t = t or tag.selected()
+    tag.setproperty(t, "master_fill_policy", enabled)
+end
+
+--- Toggle size fill policy for the master client(s)
+-- @tparam tag t The tag to modify, if null tag.selected() is used.
+function tag.togglemfpol(t)
```
blueyed: This should document what it toggles between.

lib/awful/tag.lua
```diff
@@ -409,6 +409,36 @@ function tag.getgap(t)
     return tag.getproperty(t, "useless_gap") or beautiful.useless_gap or 0
 end
 
+--- Set size fill policy for the master client(s)
+-- @tparam string policy Can be set to
+-- "expand" (fill all the available workarea) or
+-- "mwfact" (fill only an area inside the master width factor)
+-- @tparam tag t The tag to modify, if null tag.selected() is used.
+function tag.setmfpol(enabled, t)
```
blueyed: The argument name is misleading.. indicates to expect a boolean.

lib/awful/tag.lua
```diff
@@ -409,6 +409,36 @@ function tag.getgap(t)
...
+    if tag.getmfpol(t) == "expand" then
+        tag.setmfpol("mwfact", t)
+    else
+        tag.setmfpol("expand", t)
+    end
+end
+
+--- Get size fill policy for the master client(s)
+-- @tparam tag t Optional tag.
+-- @treturn string Can be set to
```
blueyed: s/Can bet set to/Possible values are/

And also mention the default of `expand` here.

blueyed:
Cool feature!

I think that a numerical value would be useful maybe, too.
This would behave like the "mwfact" mode, but can be different from it then.
But I see that the main benefit of using "mwfact" is that clients are not getting resized.

lib/awful/tag.lua
actionless: oops, i did not renamed it from the time when it actually was a bool

actionless:
>I think that a numerical value would be useful maybe, too.
This would behave like the "mwfact" mode, but can be different from it then.
But I see that the main benefit of using "mwfact" is that clients are not getting resized.

I did it intentionally because i wanted master to always have the same width.
But bright thing here is that what it's possible to add another policy which would behave differently or if it's set to number instead of string it could work instead of mwfact value (or that was what you did suggested?).

And i still need some feedback on generalization (https://github.com/awesomeWM/awesome/pull/323#issuecomment-122023593) and one more layout policy (https://github.com/awesomeWM/awesome/pull/323#issuecomment-121923384).

Btw, what about naming, should i change "mfpol" to smth different?

blueyed:
> if it's set to number instead of string it could work instead of mwfact value (or that was what you did suggested?).

Yes.

> Btw, what about naming, should i change "mfpol" to smth different?

I'd like a more verbose name, but I see that it's based on the current naming scheme.  As long as the docs are good, I think it's OK.

blueyed:
> And i still need some feedback on generalization (#323 (comment)) and one more layout policy (#323 (comment)).

Sorry, I cannot provide feedback for this really.. :/

blueyed:
@actionless 
Do you think this is good as it is for master now? (given that you would like to get some feedback still)

actionless:
i am using this option daily and don't see any issues with it

additional options can be added in a separate PR though

blueyed:
Merged, thanks a lot!

Issue #182, opened 2015-03-29
mod-j/k problems with HeidiSQL (main window 1x1 pixel, focuses its dialog)


blueyed:
[This was originally reported in the old bugtracker ([FS#1284](https://awesome.naquadah.org/bugs/index.php?do=details&task_id=1284)).]

1. HeidiSQL (download at http://www.heidisql.com/download.php) breaks Mod+J:
the focus cannot be moved away from the HeidiSQL window. The taskbar entry flashes, but does not move to the next client.

2. When moving to the HeidiSQL client with Mod+K, the client window does not get raised properly: it gets the focus, but stays in the background.

This requires you to use combinations of mod+j and mod+k to move away or select the client window.

The relevant properties of the dialog ("Session manager") are:

    WM_STATE(WM_STATE):
                    window state: Normal
                    icon window: 0x0
    _NET_WM_STATE(ATOM) = _NET_WM_STATE_SKIP_TASKBAR
    _NET_WM_NAME(UTF8_STRING) = "Session manager"
    WM_ICON_NAME(STRING) = "Session manager"
    WM_NAME(STRING) = "Session manager"
    WM_HINTS(WM_HINTS):
                    Client accepts input or input focus: False
                    Initial state is Normal State.
                    bitmap id # to use for icon: 0x58009dc
                    bitmap id # of mask for icon: 0x58009de
                    window id # of group leader: 0x5a00003
    _NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL
    WM_TRANSIENT_FOR(WINDOW): window id # 0x5a00003
    _MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x26, 0x1e, 0x114400, 0x10500
    WM_NORMAL_HINTS(WM_SIZE_HINTS):
                    program specified location: 371, 154
                    window gravity: Static
    _NET_WM_USER_TIME_WINDOW(WINDOW): window id # 0x5800030
    XdndAware(ATOM) = ATOM
    _NET_WM_PID(CARDINAL) = 15842
    WM_LOCALE_NAME(STRING) = "en_US.UTF-8"
    WM_CLIENT_MACHINE(STRING) = "lenny.thequod.de"
    WM_CLASS(STRING) = "heidisql.exe", "Wine"
    WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, _NET_WM_PING, WM_TAKE_FOCUS

[...]

PR #324, opened 2015-07-16
Improve transient for

blueyed:
I've came across this old branch, which is meant to improve the behavior described in #182, and provides utility functions to get the `transient_for` root.

Looking at it, I think `is_transient_for` should also return true for any intermediate clients?!


psychon:
Idea looks fine to me, thanks

psychon:
I think this can be merged

blueyed:
Yes, I've rebased it and cleaned it up.

PR #325, opened 2015-07-17
tasklist: display symbols with "above" and "below" clients

blueyed:
When windows are set as "above", it will basically behave as "ontop"
under normal circumstances and therefore an indicator should get
displayed.

### DIFF
--- a/lib/awful/widget/tasklist.lua
+++ b/lib/awful/widget/tasklist.lua
@@ -51,6 +51,8 @@ local function tasklist_label(c, args)
     -- symbol to use to indicate certain client properties
     local sticky = args.sticky or theme.tasklist_sticky or "▪"
     local ontop = args.ontop or theme.tasklist_ontop or '⌃'
+    local above = args.above or theme.tasklist_above or '▴'
+    local below = args.below or theme.tasklist_below or '▾'
     local floating = args.floating or theme.tasklist_floating or '✈'
     local maximized = args.maximized or theme.tasklist_maximized or '<b>+</b>'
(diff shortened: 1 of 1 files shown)


Elv13:
(While I still prefer the way Awesome 3.4 did it)

+1

blueyed:
> (While I still prefer the way Awesome 3.4 did it)

How was it done there?

Elv13:
There was icons on the right side of the tasklist. While the default was only for floating and the icon was unrecognizable, it was kind of nice. You can see it in this old screenshot (from 3.4, with the default tasklist)

https://imagizer.imageshack.us/v2/1715x965q90/19/blender175.png

Radical now support that.

blueyed:
> There was icons on the right side of the tasklist.

Do you mean (the window's) taskbar? (which I do not use by default)

psychon:
Fine by me (and I do know the difference between above and ontop :-) )

Elv13:
> Do you mean (the window's) taskbar? (which I do not use by default)
In the wibox tasklist itself, not the titlebar. Each entry had a butterfly icon on the right if the client was floating

blueyed:
@Elv13 

> In the wibox tasklist itself, not the titlebar. Each entry had a butterfly icon on the right if the client was floating

Sorry, it's really hard to grasp from your screenshot.
But I figured it out: a) the tasklist is at the bottom and the floating indicator is very light.

actionless:
i think the main point what it have different color than text what currently impossible because client name and status symbols are inside one textbox

psychon:
Just use `theme.tasklist_below = "<span color='#ff0000'>▾</span>"` in your theme to set a color?

actionless:
yeah, makes sense

PR #326, opened 2015-07-17
Add "visible" property to widgets

blueyed:
With `visible = false`, widgets will not be drawn.

I've looked for a quick way to hide a particular widget, which has no outer wibox.

Do you think this makes sense, or should the widget just get wrapped in a wibox (assuming that this would be enough then)?

### DIFF
--- a/lib/wibox/layout/mirror.lua
+++ b/lib/wibox/layout/mirror.lua
@@ -17,9 +17,11 @@ local mirror = { mt = {} }
 
 --- Draw this layout
 function mirror:draw(wibox, cr, width, height)
-    if not self.widget then return { width = 0, height = 0 } end
+    if not self.widget then
+        return { width = 0, height = 0 }
+    end
     if not self.horizontal and not self.vertical then
-        self.widget:draw(wibox, cr, width, height)
+        base.draw_widget(wibox, cr, self.widget, 0, 0, width, height)
         return -- nothing changed
     end
 
--- a/lib/wibox/widget/base.lua
+++ b/lib/wibox/widget/base.lua
@@ -107,6 +107,15 @@ function base.make_widget(proxy, widget_name)
         ret._fit_geometry_cache = cache.new(cb)
     end)
 
+    -- Add visible property and setter.
+    ret.visible = true
+    function ret:set_visible(b)
+        if b ~= self.visible then
+            self.visible = b
+            self:emit_signal("widget::updated")
+        end
+    end
+
     -- Add __tostring method to metatable.
     ret.widget_name = widget_name or object.modulename(3)
     local mt = {}
--- a/spec/wibox/test_utils.lua
+++ b/spec/wibox/test_utils.lua
@@ -59,6 +59,7 @@ return {
 
     widget_stub = function(width, height)
         local w = object()
+        w.visible = true
         w:add_signal("widget::updated")
 
         w.fit = function()
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -57,7 +57,7 @@ local function do_redraw(self)
 
     -- Draw the widget
     self._widget_geometries = {}
-    if self.widget then
+    if self.widget and self.widget.visible then
         cr:set_source(self.foreground_color)
         self.widget:draw(self.widget_arg, cr, width, height)
         self:widget_at(self.widget, 0, 0, width, height)
--- a/lib/wibox/layout/base.lua
+++ b/lib/wibox/layout/base.lua
@@ -34,6 +34,9 @@ end
 -- @param height The available height for the widget
 -- @return The width and height that the widget wants to use
 function base.fit_widget(widget, width, height)
+    if not widget.visible then
+        return 0, 0
+    end
     -- Sanitize the input. This also filters out e.g. NaN.
     local width = math.max(0, width)
     local height = math.max(0, height)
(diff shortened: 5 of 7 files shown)


Elv13:
+1000 for the concept

The `:fit()` hack I recommand on IRC was always ugly and the constraint layout is often overkill. This is much better. 

However I would use `get_visible` and `set_visible` to the can emit `widget::changed`. I would also use "visible" everywhere instead of mixing hidden/visible. This would less or more restore the 3.4 API for this.

syphoxy:
:+1: 

I definitely appreciate this. the last time I asked about this I was told to set width/height to zero which I felt was a bit .. less comfortable.

blueyed:
> use "visible" everywhere instead of mixing hidden/visible

I've tried to use HTML semantics here, where "visible" would not show it, but reserve the space.

>  `get_visible` and `set_visible`

Makes sense.

blueyed:
I think `self.widget.visible / .hidden` should be kept (at least internally), instead of using a getter?!

Would it make sense to use object properties, via a function (for the setter) in this case?

psychon:
`.visible` and `.hidden` should be added in two separate commits.

Sorry, what's the point of `visible`? (Although I don't really mind adding it, but I guess it can easily confuse people)

And yeah, you need to emit `widget::update` when these things change.

blueyed:
I agree that `visible` as in CSS/HTML is confusing and not really necessary (yet?).
I've only mentioned it as one of the reasons to use `hidden` in this patch, instead of `visible` (which is used elsewhere already, and what @Elv13 pointed out).

I would be OK with both alternatives, but tend to prefer `hidden`.

blueyed:
I've changed it to use `visible` instead of `hidden`, and added a `set_visible` method.
There is no need for a getter, is there?

Moving it to the "next" milestone.

lib/wibox/layout/constraint.lua
```diff
@@ -16,7 +16,7 @@ local constraint = { mt = {} }
 
 --- Draw a constraint layout
 function constraint:draw(wibox, cr, width, height)
-    if not self.widget then
+    if not self.widget or not self.widget.visible then
         return
```
psychon: This isn't necessary. Let `base.draw_widget` handle this case.

lib/wibox/layout/margin.lua
```diff
@@ -23,7 +23,7 @@ function margin:draw(wibox, cr, width, height)
     local h = self.bottom
     local color = self.color
 
-    if not self.widget or width <= x + w or height <= y + h then
+    if not self.widget or not self.widget.visible or width <= x + w or height <= y + h then
```
psychon: I'd drop this as well. Otherwise a margin around a widget of size `0,0` would look different from a margin around a widget with `visible = true` and that seems like a weird thing to me.

lib/wibox/layout/mirror.lua
```diff
@@ -17,7 +17,9 @@ local mirror = { mt = {} }
 
 --- Draw this layout
 function mirror:draw(wibox, cr, width, height)
-    if not self.widget then return { width = 0, height = 0 } end
+    if not self.widget or not self.widget.visible then
+        return { width = 0, height = 0 }
+    end
```
psychon: I'd drop this as well. Let it rotate the cairo context as needed and then... oh

lib/wibox/layout/mirror.lua
```diff
@@ -17,7 +17,9 @@ local mirror = { mt = {} }
...
     if not self.horizontal and not self.vertical then
         self.widget:draw(wibox, cr, width, height)
```
psychon: This should use `base.draw_widget` instead of calling `draw` directly.

lib/wibox/widget/base.lua
```diff
@@ -101,6 +101,15 @@ function base.make_widget(proxy)
...
 
+    -- Add visible property and setter.
+    ret.visible = true
+    ret.set_visible = function(b)
+        if b ~= ret.visible then
+            ret.visible = b
+            ret:emit_signal("widget::updated")
+        end
+    end
+
```
psychon: I'd make this a `proper method` instead of catching `ret` by an upvalue. So something like:
```
    function ret:set_visible(b)
        if b ~= self.visible then
            self.visible = b
            self:emit_signal("widget::updated")
        end
    end
```
Seems more consistent to me to other stuff "out there".

spec/wibox/test_utils.lua
```diff
@@ -22,7 +22,9 @@ end
 local function stub_draw_widget(wibox, cr, widget, x, y, width, height)
     assert.is.equal("wibox", wibox)
     assert.is.equal("cr", cr)
-    table.insert(widgets_drawn, { widget, x, y, width, height })
+    if widget.visible then
+        table.insert(widgets_drawn, { widget, x, y, width, height })
+    end
```
psychon: I'd just skip this chunk. Surely tests involving invisible widgets would still like to know if such a widget was (tried to) drawn.

psychon:
Fine by me and nope, no getter. I'd perhaps prefix the property with an underscore to mean "this is a private property which you should never have to touch". In other news, other code shouldn't have to care if a widget is visible, only `fit_widget` and `draw_widget` do.

lib/wibox/layout/mirror.lua
blueyed: > oh

That means to not drop it?
Or would/should it get handled by `base.draw_widget` anyway, too?

lib/wibox/layout/mirror.lua
blueyed: I see (regarding the previous comment).

PR #327, opened 2015-07-18
Tests

psychon:
'Nuff said. :-)

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -18,9 +18,6 @@ else
     tests=$this_dir/test*.lua
 fi
 
-# Use a separate D-Bus session; sets $DBUS_SESSION_BUS_PID.
-eval $(dbus-launch --sh-syntax)
-
 root_dir=$PWD/..
 
 # Travis.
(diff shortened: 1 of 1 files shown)


tests/run.sh
```diff
@@ -91,31 +90,13 @@ AWESOME_CLIENT="$root_dir/utils/awesome-client"
...
-            echo "Log:"
-            cat "$awesome_log"
-            exit 1
-        fi
-        sleep 0.1
-    done
+    export DISPLAY="$D"
+    cd $root_dir/build
+    # Setup xrdb, for awesome's xresources backend / queries.
+    echo "Xft.dpi: 96" | DISPLAY="$D" xrdb
```
actionless: awesome anyway defaults to the same value so i don't think it's necessary

psychon: I *guess* that this will otherwise result in a warning that the grep at the end of this file would consider an error. Just a guess. Anyway, I didn't add this and shouldn't mess with it in this unrelated change.

A quick test results in:
    ...chon/projects/awesome/build/lib/beautiful/xresources.lua:67: W: Failed to get xrdb value 'Xft.dpi'.

The `grep` wouldn't trigger on this.

blueyed: > The grep wouldn't trigger on this.

Yes, but I found it cleaner to not trigger a warning there.
And it allows for adding other resources later etc.

blueyed:
Thanks! :)

PR #329, opened 2015-07-19
default config: enhance list of default floating clients

blueyed:
This is meant to improve the default behavior.

Any more suggestions to include there by default?

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -423,12 +423,33 @@ awful.rules.rules = {
                      raise = true,
                      keys = clientkeys,
                      buttons = clientbuttons } },
-    { rule = { class = "MPlayer" },
-      properties = { floating = true } },
-    { rule = { class = "pinentry" },
-      properties = { floating = true } },
-    { rule = { class = "gimp" },
-      properties = { floating = true } },
+
+    -- Floating clients.
+    { rule_any = {
+        instance = {
+          "DTA", -- Firefox addon DownThemAll.
+        },
+        class = {
+          "Arandr",
+          "Copyq",
+          "Gpick",
+          "Kruler",
+          "MessageWin",  -- kalarm.
+          "Sxiv",
+          "Wpa_gui",
+          "pinentry",
+          "veromix",
+          "xtightvncviewer"},
+
+        name = {
+          "Event Tester", -- xev.
+        },
+        role = {
+          "AlarmWindow",  -- Thunderbird's calendar.
+          "pop-up",       -- e.g. Google Chrome's (detached) Developer Tools.
+        }
+      }, properties = { floating = true }},
+
     -- Set Firefox to always map on tags number 2 of screen 1.
     -- { rule = { class = "Firefox" },
     --   properties = { tag = tags[1][2] } },


spectromas:
Here are mine: 

    instance = "DTA" --Firefox addon DownThemAll

    class = "Qbittorrent", name = "Deletion confirmation - qBittorrent", "Torrent Upload Speed Limiting", "Torrent Download Speed Limiting", "Global Download Speed Limit", "Global Upload Speed Limit", "Choose save path", "Rename",

    class = {"mpv", "Sxiv", "Wpa_gui", "Lxappearance",}, name = {"Already in download list"}
The last one there is another qbittorrent one. I expect there is a simpler and more elegant way to do all those though.

blueyed:
@spectromas
For qbittorrent there is maybe a "role" for these windows?
I recommend / use the following script (usually run from awesome's run prompt): https://github.com/blueyed/dotfiles/blob/master/usr/bin/xpop - it displays the most relevant properties in a zenity popup.

psychon:
I can offer (note that I do not use `awful.rules` in my config):
```
	{ match = { "gimp"},			float = true },
	{ match = { "gimp[-]image[-]window"},	float = false },
	{ match = { "pinentry" },		float = true},
	{ match = { "MessageWin" },		float = true}, -- kalarm (which I don't use since years...)
	{ match = { "xtightvncviewer" },	float = true},
```

spectromas:
@blueyed thanks for that script, it's very useful. They don't appear to have a role, this is an example output from that xpop script on one of the windows: 

    _NET_WM_STATE(ATOM) = _NET_WM_STATE_MODAL
    WM_TRANSIENT_FOR(WINDOW): window id # 0x1200007
    _NET_WM_PID(CARDINAL) = 10851
    _NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DIALOG, _NET_WM_WINDOW_TYPE_NORMAL
    WM_NAME(STRING) = "Global Download Speed Limit"
    WM_CLASS(STRING) = "qbittorrent", "Qbittorrent"

blueyed:
@spectromas 
> _NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DIALOG, _NET_WM_WINDOW_TYPE_NORMAL

Wouldn't a rule `type = { "dialog" }` work then?

According to https://github.com/awesomeWM/awesome/blob/a8750fe2d59a9f2454e93319f99b4159adca3650/objects/window.c#L200-215 "dialog" should be preferred.

This might also be a very good default?!

spectromas:
@blueyed Haha oh wow yeah that does work and is much better, I didn't even know that was possible. 
Thanks a lot for pointing that out, it makes things much easier!

>This might also be a very good default?!

Absolutely, I can't think of a circumstance where you would dialog boxes to be tiled.

psychon:
@blueyed That already is a default: https://github.com/awesomeWM/awesome/blob/a8750fe2d59a9f2454e93319f99b4159adca3650/lib/awful/client.lua#L684-L688

blueyed:
@psychon 
Ah, I remember seeing this.

But why has the type-rule an effect for @spectromas then?
Is the floating property being set to false explicitly from another rule? (or from something like persistent properties etc)

blueyed:
Rebased and comments added.

@psychon 
Not sure about your `{ match = { "gimp[-]image[-]window"},   float = false },` - it is not meant to be a class?
Is it required with the "gimp" rule we had already?

actionless:
i am using most of those apps in a tiling layout so I can't really get the point of making the rule to make them float.

also, regarding `qbittorent` -- its dialog windows (like adding a new torrent) are floating without any special rules

blueyed:
@actionless 
> qbittorent

Should get removed then probably altogether, if it behaves sane.

Do you mean that you use `xev` and others in tile mode?
Which would you not define by default?

actionless:
Arandr, Gpick, Lxappearance, MPlayer/mpv, gimp (i am using single-windows layout there), xev

btw, what is `pinentry`?

i personally prefer to avoid floating windows as much as possible because i am often forgetting about them when they are overlapped by other windows

blueyed:
> btw, what is pinentry?

Used with gpg-agent to ask for passphrases.

I dislike it that windows would get moved around for clients like that, which then should also not span half of the screen.  Also depends on the number of clients on the screen already.

actionless:
btw, i think for password entry window it makes sense not only to make it float, but also to make it ontop -- that's again related to my own issues with loosing windows overlapped by other ones

blueyed:
> ontop

Good point, although pinentry is that by default.

spectromas:
I think the reason I was originally making all the exclusions for qbittorrent was because I had set it to be maximized permanently but this caused some of the dialog windows to also be maximized.

psychon:
> Not sure about your { match = { "gimp[-]image[-]window"}, float = false }, - it is not meant to be a class?
Well, this is "old stuff" back when gimp only had its multi-window mode. I guess I wanted the toolboxes to be floating, but the image window itself not to float. Now that Gimp has a single window mode, I'm not sure how much sense it makes to keep this at all...

blueyed:
@psychon 
> Now that Gimp has a single window mode, I'm not sure how much sense it makes to keep this at all...

But the class "gimp" was matched already before?!  Aren't the patterns anchored?

Anyway, it sounds like we could/should remove gimp altogether then?

blueyed:
@actionless 
[to be removed]
> Arandr, Gpick, Lxappearance, MPlayer/mpv, gimp (i am using single-window layout there), xev

I disagree with Gpick and xev at least: I use them on the current window setup/layout, and if they would be launched as a new master window you would not be able to easily use them, because the window you want to target has been resized?!

I've pushed some new commits, please review it again.

psychon:
I still don't like matching by window name, but the rest seems good to me (although I only know less than half of these)

blueyed:
awful.client.property.floating(CARDINAL) = 1
_NET_WM_DESKTOP(CARDINAL) = 7
> I still don't like matching by window name

Is there another way to match `xev`?
```
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
_NET_FRAME_EXTENTS(CARDINAL) = 2, 2, 2, 2
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW
WM_NORMAL_HINTS(WM_SIZE_HINTS):
		program specified size: 178 by 178
		program specified minimum size: 78 by 78
WM_COMMAND(STRING) = { "xev" }
WM_NAME(STRING) = "Event Tester"
```

psychon:
@blueyed Sorry, `xev` should be fine (that window title doesn't get translated). However, doesn't thunar localize its window title ("File Operation Progress")?

blueyed:
Oh, yeah.  Regarding thunar that might be fine by now apparently - the progress window is of type "dialog" with Thunar 1.6.10 (Xfce 4.12).  Will remove it.

PR #330, opened 2015-07-21
xresources theme: use math.ceil with DPI values

blueyed:
Given a DPI below 96 would result in a border_size of 0.

Ref: https://github.com/awesomeWM/awesome/pull/320

### DIFF
--- a/themes/xresources/theme.lua
+++ b/themes/xresources/theme.lua
@@ -22,8 +22,8 @@ theme.fg_focus      = theme.bg_normal
 theme.fg_urgent     = theme.bg_normal
 theme.fg_minimize   = theme.bg_normal
 
-theme.useless_gap   = dpi(5)
-theme.border_width  = dpi(1)
+theme.useless_gap   = math.ceil(dpi(5))
+theme.border_width  = math.ceil(dpi(1))
 theme.border_normal = xrdb.color0
 theme.border_focus  = theme.bg_focus
 theme.border_marked = xrdb.color10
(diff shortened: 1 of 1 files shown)


actionless:
mb wrap it all like that?
```
function dpi(val)
  return math.ceil(apply_dpi(val))
end
```

or even move math.ceil to `apply_dpi` itself, while it causes few similar issues

blueyed:
> mb wrap it all like that?

See https://github.com/awesomeWM/awesome/pull/320#issuecomment-123272071 - I think it might be good to have it as-is.
But a wrapper-function that would use ints, and maybe (a to be added) `round` function?

blueyed:
@psychon @Elv13 
I would appreciate your feedback on how to move forward in this case (also with regard to #320).

Should `apply_dpi` use `ceil` by default?
Should there be a new function to do this? Or a new one that returns a float?

I've found that there's a `round` function in wibox.layout.flex already, which could be moved to `utils`:

    local function round(x)
        return floor(x + 0.5)
    end

But then it's probably better to use `ceil` always, especially for values below 1.

Elv13:
I see no issues of using `ceil` directly. Just as for the widget layouts, computer screens display pixels. Pixels are the ultimate unit. Unless someone come up with an use case where the `float` is still desirable at that point, I think using `ceil` in `dpi()` is cleaner.

blueyed:
I agree that just using `math.ceil` is a sensible and simple solution for now.

PR #331, opened 2015-07-21
client_resize_do: prefer client's current screen

blueyed:
In case of an overlapping screen configuration, prefer the client's
current screen.

Without this, clients would be moved to the first matching screen.

I've thought about adding a `screen_getbycoord_prefer_given` function to handle this, but using the factored out `screen_coord_in_screen` appears to be easy enough, and even clearer.

For reference, the xrandr setup:

    Screen 0: minimum 8 x 8, current 2948 x 1200, maximum 32767 x 32767
    LVDS1 connected primary 1366x768+0+0 (normal left inverted right x axis y axis) 277mm x 156mm
    HDMI2 connected 1920x1200+1028+0 (normal left inverted right x axis y axis) 519mm x 324mm

Note the offset of the HDMI2 output, which overlaps into LVDS1 (while I've
experimented with scaling etc).

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -795,7 +795,10 @@ client_resize_do(client_t *c, area_t geometry, bool force_notice, bool honor_hin
     lua_State *L = globalconf_get_lua_State();
     bool send_notice = force_notice;
     bool hide_titlebars = c->fullscreen;
-    screen_t *new_screen = screen_getbycoord(geometry.x, geometry.y);
+
+    screen_t *new_screen = c->screen;
+    if(!screen_coord_in_screen(new_screen, geometry.x, geometry.y))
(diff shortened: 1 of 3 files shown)


psychon:
This change is fine with me. That '< 0' thing should either be ignored or dealt with separately. Suggestions?

psychon:
Shall we just merge this as-is? If you want, I can add code for finding the closest screen instead of closest-upper-left-corner-of-a-screen later based on this.

blueyed:
@psychon 
> Shall we just merge this as-is?

Yeah, I won't have the time currently to look deeper into this.

> If you want, I can add code for finding the closest screen instead of closest-upper-left-corner-of-a-screen later based on this.

That would be great!

PR #332, opened 2015-07-21
screen_client_moveto: only emit property::screen if it changed

blueyed:
While debugging #331, I've noticed that the call to `client_resize`
might have changed the screen (and emitted the signal) already, via
a call to `screen_client_moveto` with `doresize=False`.


psychon:
I'll just trust you on this :+1:

PR #334, opened 2015-07-22
Fixes to awful.client's lazy-loading of awful.screen

blueyed:
At least awful.client.getmaster was broken:

> lib/awful/client.lua:450: attempt to index global 'awful' (a nil value)

Follow-up to cbe684e (https://github.com/awesomeWM/awesome/pull/94).

Not tested nor proofread really yet.

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -24,9 +24,17 @@ local capi =
     awesome = awesome,
 }
 
--- we use require("awful.screen") inside functions to prevent circular dependencies.
+-- We use a metatable to prevent circular dependency loops.
 local screen
-
+do
+    screen = setmetatable({}, {
+        __index = function(t, k)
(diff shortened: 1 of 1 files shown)


psychon:
Yay for dependency loops. :-(

Fine by me, but I'd suggest adding some helper to get `awful.screen` so that delayed loading is only implemented in one place. Perhaps even some metatable magic?

```lua
local screen
do
    screen = setmetatable({}, {
        _index = function(t, k)
            screen = require("awful.screen")
            return screen[k]
        end,
        _newindex = error -- Just to be sure in case anything ever does this
     }
end
```

blueyed:
Good idea with the metatable, done in https://github.com/blueyed/awesome/commit/f4415ada682cf4e4cfd209d1b46831fda95faefc.

PR #335, opened 2015-07-22
wibox.widget.textbox: set_markup/set_text: skip unchanged

blueyed:
In case the text or markup has not changed, there's no need to update
the widget.

Not too sure about it, but came across it while hacking on widgets in general.

### DIFF
--- a/lib/wibox/widget/textbox.lua
+++ b/lib/wibox/widget/textbox.lua
@@ -58,10 +58,15 @@ end
 --   `<b>bold</b>`). You can use awful.util.escape to escape
 --   parts of it.
 function textbox:set_markup(text)
+    if self._markup == text then
+        return
+    end
+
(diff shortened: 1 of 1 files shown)


Elv13:
+1

blueyed:
@psychon 
> The fix in my branch was to also save the test with markup in the textbox widget and compare it before parsing the markup.

IIRC that's in the new_widget_system branch, isn't it?

Then this PR can be closed..

psychon:
Well, I'd like to make that branch smaller, so I'd still be nice if we get this in before.
Could you update this to do something similar to https://github.com/psychon/awesome/blob/new_widget_system/lib/wibox/widget/textbox.lua#L61-L83 ?

blueyed:
@psychon 
Ok, done in https://github.com/awesomeWM/awesome/commit/dbf8b1b5a916d93b709551a7107d181956ec440e.

PR #339, opened 2015-07-23
Get clients in stacking order (for awful.client.swap.bydirection)

blueyed:
Fixes/for https://github.com/awesomeWM/awesome/issues/178.

Clients in stacked order are returned top-to-bottom, which is what you're usually interested in (done in the fixup-commit).

PR #340, opened 2015-07-23
Keep client order across restarts

blueyed:
I've noticed that the internal order of clients gets reversed when
restarting awesome.

This patch fixes this by calling `xcb_reparent_window` in the reverse
order.

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -79,7 +79,7 @@ awesome_atexit(bool restart)
     signal_object_emit(L, &global_signals, "exit", 1);
 
     /* Move clients where we want them to be */
-    foreach(c, globalconf.clients)
+    foreach_reverse(c, globalconf.clients)
     {
         xcb_reparent_window(globalconf.connection, (*c)->window, globalconf.screen->root,
                 (*c)->geometry.x, (*c)->geometry.y);
@@ -93,7 +93,7 @@ awesome_atexit(bool restart)
     lua_close(L);
 
     /* X11 is a great protocol. There is a save-set so that reparenting WMs
-     * don't kill clients when they shut down. However, when a focused windows
+     * don't kill clients when they shut down. However, when a focused window
      * is saved, the focus will move to its parent with revert-to none.
      * Immediately afterwards, this parent is destroyed and the focus is gone.
      * Work around this by placing the focus where we like it to be.
--- a/common/array.h
+++ b/common/array.h
@@ -50,6 +50,15 @@
             (var = &(array).tab[__foreach_index_##var]);                    \
             ++__foreach_index_##var)
 
+#define foreach_reverse(var, array) \
+    for(int __foreach_index_##var = (array).len-1; \
+        __foreach_index_##var > -1; \
+        __foreach_index_##var = -1) \
+        for(typeof((array).tab) var = &(array).tab[__foreach_index_##var];  \
+            (__foreach_index_##var > -1) &&                                 \
+            (var = &(array).tab[__foreach_index_##var]);                    \
+            --__foreach_index_##var)
+
 /** Common array functions */
 #define ARRAY_COMMON_FUNCS(type_t, pfx, dtor)                               \
     static inline pfx##_array_t * pfx##_array_new(void) {                   \


psychon:
Wow. I did not know that `ReparentWindow` puts clients at the top of the stacking order.
Is it more important to keep the stacking order intact or the order that clients actually `are in`? If we want to keep both, then I'd suggest keeping the stacking order intact and saving the client order in some property on the root window at shutdown.

However, this change is also fine by me.

blueyed:
> If we want to keep both, then I'd suggest keeping the stacking order intact and saving the client order in some property on the root window at shutdown.

I've thought about the property on the root window, but then discovered that the problem at hand (keeping the order of entries in the tasklist) was easy to solve like this.

Do you mean that this change will mess with the stacking order instead?

To be clear: this patch fixes `client.get()` to be the same across restarts.

psychon:
Yeah, this change loses the stacking order which is possibly more annoying than loosing the order of the tasklist entries?

blueyed:
Are you sure?

I've just tested it with `return require("gears").debug.dump_return(client.get(nil, true))` (and #339) and it's the same order across restarts.

I only noticed that a focused floating window will get put on top of the stack, although it wasn't on top before the restart - but IIRC that's handled somewhere else in a special way.

psychon:
I opened four client, focused the second client and then the last (so that the stacking order was 4, 2, 3, 1). After restarting awesome, the stacking order ended up being 4,3,2,1.

So yes, I am sure.

Also: http://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#requests:ReparentWindow

    The window is placed on top in the stacking order with respect to siblings.

blueyed:
I do not understand this issue then really: shouldn't the output of `wmctrl -l` be the same across restarts, and not get reversed?

Also this patch touches code which involves `globalconf.clients` and not `globalconf.stack`.
Doesn't this mean that the stacking order gets messed up anyway?

Is it maybe that `globalconf.stack` was meant to be used here instead (to keep the stack order)?

Ok, `git blame` shows that it was added by you in 27988ec to solve something else:

> Remove titlebars from clients during shutdown (FS#1159)
>
> This reparents all clients back to their proper position during shutdown, so
> that their top-left corner is now where their titlebar's top-left corner was.
> Hopefully, this fixes floating clients moving around across a restart.

So given that description, the order should not matter (because it's about the position), and the stacking order was broken unintentionally back then already?!

psychon:
Oh... you are right. That should use `globalconf.stack` and thus the current code is already wrong.

blueyed:
Ok.  What's the plan then?

What about just skipping this on restart?

psychon:
Skipping this on restart: Floating clients will be shifted by `border_width` + the size of the titlebar.

I guess changing the loop to iterate over `globalconf.stacks` (and either using `foreach` or `foreach_reverse`, which ever gives the right order) is the best thing to do.

However, this will then lose the client order from e.g. the tasklist and I guess we will just have to live with that.

blueyed:
> Skipping this on restart: Floating clients will be shifted by border_width + the size of the titlebar.

True.  But only if they have one.

Given that this is the original issue, couldn't only these clients get handled here, i.e. moving them based on the offset?

> However, this will then lose the client order from e.g. the tasklist and I guess we will just have to live with that.

That was the initial issue I was trying to fix, and it's more annoying than floating clients moving around.

blueyed:
This appears to work:

    /* Move clients where we want them to be, but only when exiting. */
    foreach(c, globalconf.stack)
        for (client_titlebar_t bar = CLIENT_TITLEBAR_TOP; bar < CLIENT_TITLEBAR_COUNT; bar++)
            if ((*c)->titlebar[bar].drawable)
            {
                xcb_reparent_window(globalconf.connection, (*c)->window,
                        globalconf.screen->root, (*c)->geometry.x, (*c)->geometry.y);
                client_send_configure((*c));
                break;
            }


But I think something less "intrusive" should be used.  I've tried
`client_send_configure((*c))`, but without success.


Also, we should add `bool client_has_titlebar(client_t *c)` then probably.

blueyed:
Closing this in favour of #356.

Issue #337, opened 2015-07-23
Provide widget name (for debugging)


blueyed:
While debugging / hacking on widget related stuff, I've added something like the following to e.g. `lib/wibox/layout/fixed.lua`:

    ret.widget_name = "layout.flex." .. dir

Is this the way to go, or is there something else already to get the name of a widget/layout/wibox?

psychon:
Currently there is no way to get the name of a widget.

blueyed:
Do you think it makes sense to manually define it then?

And/Or should we add a function to `gears.debug` to get the module name of an object somehow (if that is possible)?

psychon:
I don't think we can come up with something useful for `gears.debug`. For names on widget's, you should follow your approach of adding a `widget_name` member, I guess.

Perhaps we can also use this in `wibox.widget.base.draw_widget` in case an error occurs? I'm not sure if that makes sense...

blueyed:
The following would work in `base.make_widget`, but `debug.getinfo` is documented as not being efficient (http://www.lua.org/pil/23.1.html).

    ret.widget_name = debug.getinfo(2, 'S').source:gsub('.*/lib/', ''):gsub("/", "."):gsub("\\.lua$", "")

I've tried to apply this to `gears.object`, with the idea of using `getinfo` on the object's (`self`) metatable / `__call` method, but the metatable is not available from the objects itself.
Doing so would require to change all callers (e.g. wibox/widget/background.lua) to add it probably?

psychon:
That `debug.getinfo` thing looks scary. I'd rather make this explicit (and perhaps so that the widget can change its name and run-time and e.g. add more information? Dunno how useful that would be)

blueyed:
> I'd rather make this explicit

True.
This would also eliminate the discussion about any performance impact.

PR #341, opened 2015-07-23
wibox.widget.base: add __tostring method to widgets

blueyed:
The drawback here appears to be that `debug.getinfo` is not very
efficient, and required to be run to get the call stack.

However, using "S" to limit the information being gathered helps, and it
is not that slow after all:

    % time lua -e "for i = 1, 1000000 do debug.getinfo(1) end"
    2,25s user 0,00s system 99% cpu 2,255 total
    % time lua -e "for i = 1, 1000000 do debug.getinfo(1, 'S') end"
    1,32s user 0,00s system 99% cpu 1,318 total

Fixes https://github.com/awesomeWM/awesome/issues/337.

### DIFF
--- a/lib/gears/object.lua
+++ b/lib/gears/object.lua
@@ -112,6 +112,21 @@ function object.mt:__call(...)
     return new(...)
 end
 
+--- Helper function to get the module name out of `debug.getinfo`.
+-- @usage
+--  local mt = {}
+--  mt.__tostring = function(o)
+--      return require("gears.object").modulename(debug.getinfo(1, 'S'))
+--  end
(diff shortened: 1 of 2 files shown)


blueyed:
How about https://github.com/blueyed/awesome/commit/23a166e47712da6867daf2126d461f40ff06d0ce?

This way widgets can provide their names (and the internal ones would do so), but it falls back to using `object.modulename`, which got more convenient based on your suggestion.

psychon:
You don't have to rename `debug` to `gears_debug` anymore. Perhaps also add a prefix saying "widget"?
Besides that: :+1:

PR #343, opened 2015-07-23
Travis: build and upload API doc on success for master

blueyed:
This is a test / proof-of-concept.

It uses Github pages, and the docs will be published at http://awesomewm.github.io/apidoc/.

Useful Travis config vars: http://docs.travis-ci.com/user/environment-variables/, e.g. `TRAVIS_TAG`.

This could also push the documentation to http://awesome.naquadah.org/doc/api/master.
For this, secret/encrypted environment variables can be configured (see http://docs.travis-ci.com/user/encryption-keys/).
I have created a separate Github user, which has only limited access for this purpose, and who's Github token is used for this.

An example build: https://travis-ci.org/awesomeWM/awesome/builds/72388323 (with an adjusted test, normally PRs will be skipped).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ env:
     # Disable 5.1 for now: no need to test both 5.1 and 5.2 currently.
     # - LUA=5.1 LUAPKG=5.1 LUALIB=/usr/lib/x86_64-linux-gnu/liblua5.1.so
     - LUA=5.2 LUAPKG=5.2 LUALIB=/usr/lib/x86_64-linux-gnu/liblua5.2.so
+  global:
+    # Secure token to push to gh-pages.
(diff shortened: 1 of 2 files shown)


blueyed:
It should not force-push, but add a new commit instead.

blueyed:
> What a hack to get the git repo.

Do you mean the environment vars etc?
Or the GH_TOKEN / Travis secret?

Keep in mind that it's a separate repo and user, only for the docs.  Travis should not have write access to the main repo.

> I don't think your command handles file deletions correctly

`git add --all` should take care of it.

blueyed:
Ok.

What do you think about using Github pages for this?
Should it instead/additionally upload to the main site, into a separate dir? (https://github.com/awesomeWM/awesome-www/blob/master/Makefile#L3)
This would require write access for @awesome-robot, which should be limited to this specific task.

/cc @jd

psychon:
I'm fine with Github pages. I don't look into docs anyway. ;-)

jd:
@blueyed yup, GH page looks cool.

PR #345, opened 2015-07-24
README: add reference to online doc, small improvements

blueyed:
I've looked at it for #343 (to link the up-to-date version), and noticed that http://awesome.naquadah.org/doc/ is not mentioned at all.


psychon:
:+1:

PR #347, opened 2015-07-24
spawn: improve doc, and add `sn` for `awful.util.spawn_with_shell`

blueyed:


psychon:
Besides those "minor" things: :+1:

PR #349, opened 2015-07-24
doc: distinguish between client.object and client.class

blueyed:
Instead of `client.client`, the client object is now referred to as
`client.object` and the client class as `client.class`.

This moves the documentation of `client.focus` to the class.


psychon:
You are the guy who understands the docs, so sure :-)

PR #350, opened 2015-07-25
Change "Mod1" to "Mod4" in menubar/init.lua

tigrisathon:
"Mod1" ususally stands for Alt, "Mod4" stands for Meta and "Mod4" is used in rc.lua. 
In menubar/init.lua it is said: ""C-M-Return"    execute the command in a terminal", but in code "Mod1" is used instead. It seems to me, it should be changed to "Mod4"

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -275,7 +275,7 @@ local function prompt_keypressed_callback(mod, key, comm)
     elseif key == "Return" or key == "KP_Enter" then
         if mod.Control then
             current_item = #shownitems
-            if mod.Mod1 then
+            if mod.Mod4 then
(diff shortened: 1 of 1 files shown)


blueyed:
:+1: 
Will merge/cherry-pick it later.

blueyed:
Sorry, it's good as it is: "C-M-Return" means Ctrl-Alt-Return.

"M" refers to "Meta", which is a synonym for "Alt".

alexander-yakushev:
@tigrisathon `C-M-Return` is an Emacs way of writing Ctrl-Alt-Return. Perhaps a change in docs is necessary? Because in the code Mod1 is meant exactly as Alt.

PR #354, opened 2015-07-25
Some stuff around the "check" target

psychon:
The main thing is that this includes the integration tests in "make check".


blueyed:
Awesome, pushed to master.

PR #356, opened 2015-07-26
Client order after restart

psychon:
This PR makes sure we keep the stacking order (which client is above which other client) and the client order (the order in which they appear in the tasklist) across restarts.

Note that this explicitly does not handle "quit, then start again" (the property won't be set during shutdown then).


blueyed:
:+1:

blueyed:
Merged in c945492 and 86e242b.
Thanks!

PR #357, opened 2015-07-27
Add a textbox widget that shows the output of a command and updates it

b80905:


actionless:
idea itself is nice, but can it be async?

blueyed:
See https://github.com/Elv13/awesome-configs/blob/master/utils/fd_async.lua for @Elv13's async lib.
I think it would be great to get this into awesome first.

actionless:
@blueyed on my config it segfaults awesome when doing a nested async call

actionless:
@b80905 are you going to re-open it?
otherwise i can take it

b80905:
@actionless, no, you can take it.

PR #358, opened 2015-07-27
Travis: enhanced apidoc script

blueyed:
It will now handle all branches and PRs (but only push master to/as gh-pages).
In case of PRs a comment should be added (we'll see).

TODO:

 - [x] use commit message of source's HEAD


blueyed:
It got quite evolved, although it cannot be used for PRs from another fork (because the secret API token won't be provided for security reasons).
But the main feature of updating docs on master is this there, of course.. :)

PR #362, opened 2015-07-28
Travis: add Lua 5.1 to the build matrix

blueyed:


psychon:
:+1: (Especially since I'm now also using 5.2 and thus no longer test with 5.1)

blueyed:
@mindeunix
You've reported this on IRC, right?
So, just for your information: it is fixed again.

PR #364, opened 2015-07-28
keygrabber: ignore modifiers for name of control keys

blueyed:
`xkb_state_key_get_one_sym` returns `Meta_L` for a (release) event of
`Alt_L` when Shift is pressed.

This patch changes it to use `xcb_key_symbols_get_keysym` in the same
way like normal key handling does it in `event_handle_key`.  This was
changed in 372d12e, but apparently forgotten for the keygrabber.

### DIFF
--- a/keygrabber.c
+++ b/keygrabber.c
@@ -89,14 +89,13 @@ keygrabber_handlekpress(lua_State *L, xcb_key_press_event_t *e)
 
     if (is_control(buf))
     {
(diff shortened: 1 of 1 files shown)


blueyed:
@petoju 
Does this make sense?

petoju:
@blueyed it seems to be ok.

blueyed:
@petoju 
Thanks!

PR #34, opened 2014-05-05
Startup notification callbacks

Elv13:
This pull finally complete a feature that started in my June 2013 pull. It is now possible to have pass arguments to awful.spawn() to, for example, spawn in a new tag. It also add some other things to make it easier to use.

It is a "big" pull, sorry about that. I tried to avoid making so much changes, but in the end they are required, so umm, have an happy time reviewing this.

PR #365, opened 2015-07-28
(input/utf8) key handling in mappings

blueyed:
This patch uses `xkb_state_key_get_utf8` to lookup keys that could not
get translated to a keycode or keysym during setup/definition.

This allows for using the actual key also for non-ASCII mappings:

    awful.key({ modkey }, "ü",          function() print("ü") end),
    awful.key({ modkey }, "#34",        function() print("#34") end),
    awful.key({ modkey }, "udiaeresis", function() print("udiaeresis") end),

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -131,7 +131,6 @@ pkg_check_modules(AWESOME_REQUIRED REQUIRED
     glib-2.0
     gdk-pixbuf-2.0
     cairo
-    x11
     x11-xcb
     xcb-cursor
     xcb-randr
--- a/event.c
+++ b/event.c
@@ -44,6 +44,8 @@
 #include <xcb/xcb_event.h>
 #include <xcb/xkb.h>
 
+#include <xkbcommon/xkbcommon.h>
+
 #define DO_EVENT_HOOK_CALLBACK(type, xcbtype, xcbeventprefix, arraytype, match) \
     static void \
     event_##xcbtype##_callback(xcb_##xcbtype##_press_event_t *ev, \
@@ -87,11 +89,21 @@
 static bool
 event_key_match(xcb_key_press_event_t *ev, keyb_t *k, void *data)
 {
+    char buf[MAX(MB_LEN_MAX, 32)];
+
     assert(data);
     xcb_keysym_t keysym = *(xcb_keysym_t *) data;
-    return (((k->keycode && ev->detail == k->keycode)
-             || (k->keysym && keysym == k->keysym))
-            && (k->modifiers == XCB_BUTTON_MASK_ANY || k->modifiers == ev->state));
+    if((k->modifiers != XCB_BUTTON_MASK_ANY && k->modifiers != ev->state))
+        return false;
+
+    if((k->keycode && ev->detail == k->keycode)
+            || (k->keysym && keysym == k->keysym))
+        return true;
+
+    if(k->utf8 && xkb_state_key_get_utf8(globalconf.xkb_state, ev->detail, buf, sizeof(buf)))
+        return !strcmp(k->utf8, buf);
+
+    return false;
 }
 
 static bool
--- a/objects/key.c
+++ b/objects/key.c
@@ -39,8 +39,6 @@
 #include "common/xutil.h"
 #include "xkb.h"
 
-/* XStringToKeysym() and XKeysymToString */
-#include <X11/Xlib.h>
 #include <xkbcommon/xkbcommon.h>
 
 /** Key object.
@@ -82,7 +80,7 @@ luaA_keystore(lua_State *L, int ud, const char *str, ssize_t len)
     /* Then set up the new state */
     if(*str != '#' || len == 1)
     {
-        key->keysym = XStringToKeysym(str);
+        key->keysym = xkb_keysym_from_name(str, 0);
         if(!key->keysym)
         {
             if(len > 1)
--- a/objects/key.c
+++ b/objects/key.c
@@ -55,6 +55,12 @@
  * @table key
  */
 
+static void
+key_wipe(keyb_t *key)
+{
+    p_delete(&key->utf8);
+}
+
 /** Get the number of instances.
  *
  * @return The number of key objects alive.
@@ -64,28 +70,31 @@
 static void
 luaA_keystore(lua_State *L, int ud, const char *str, ssize_t len)
 {
+    if(!len)
+        return;
     keyb_t *key = luaA_checkudata(L, ud, &key_class);
-    if(len)
+
+    /* First, unset everything */
+    p_delete(&key->utf8);
+    key->keycode = 0;
+    key->keysym = 0;
+
+    /* Then set up the new state */
+    if(*str != '#' || len == 1)
     {
-        if(*str != '#')
+        key->keysym = XStringToKeysym(str);
+        if(!key->keysym)
         {
-            key->keysym = XStringToKeysym(str);
-            if(!key->keysym)
-            {
-                if(len == 1)
-                    key->keysym = *str;
-                else
-                    warn("there's no keysym named \"%s\"", str);
-            }
-            key->keycode = 0;
+            if(len > 1)
+                warn("There is no keysym named \"%s\". Will be handled as utf8 key.", str);
+            key->utf8 = a_strdup(str);
         }
-        else
-        {
-            key->keycode = atoi(str + 1);
-            key->keysym = 0;
-        }
-        luaA_object_emit_signal(L, ud, "property::key", 0);
     }
+    else
+    {
+        key->keycode = atoi(str + 1);
+    }
+    luaA_object_emit_signal(L, ud, "property::key", 0);
 }
 
 /** Create a new key object.
@@ -205,15 +214,17 @@ luaA_key_get_key(lua_State *L, keyb_t *k)
         int slen = snprintf(buf, sizeof(buf), "#%u", k->keycode);
         lua_pushlstring(L, buf, slen);
     }
-    else
+    else if(k->keysym)
     {
         char buf[MAX(MB_LEN_MAX, 32)];
 
-        if (xkb_keysym_get_name(k->keysym, buf, countof(buf)) < 0) {
+        if(xkb_keysym_get_name(k->keysym, buf, countof(buf)) < 0)
             return 0;
-        }
-
         lua_pushstring(L, buf);
+    }
+    else
+    {
+        lua_pushstring(L, k->utf8);
     }
     return 1;
 }
@@ -252,7 +263,9 @@ key_class_setup(lua_State *L)
     };
 
     luaA_class_setup(L, &key_class, "key", NULL,
-                     (lua_class_allocator_t) key_new, NULL, NULL,
+                     (lua_class_allocator_t) key_new,
+                     (lua_class_collector_t) key_wipe,
+                     NULL,
                      luaA_class_index_miss_property, luaA_class_newindex_miss_property,
                      key_methods, key_meta);
     luaA_class_add_property(&key_class, "key",
--- a/objects/key.c
+++ b/objects/key.c
@@ -35,6 +35,7 @@
  * @classmod key
  */
 
+#include "globalconf.h"
 #include "objects/key.h"
 #include "common/xutil.h"
 #include "xkb.h"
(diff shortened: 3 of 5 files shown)


psychon:
I don't really know the libxkbcommon parts of this, but I'll just trust you on it. Seems relatively simple. Perhaps @petoju can still tell us that this is correct? :-)
(Note for the "who owns the string" question I ask: I guess the string comes from lua and so will eventually be garbage-collected by Lua. That'd be bad for us. Definitely needs to make a copy and make sure that it is freed everywhere correctly.)

petoju:
@blueyed  As far as I see, libxkbcommon usage seems to be correct.

Except the already commented issues (thx @psychon), it does not support only '#' character in luaA_keystore, if I am not mistaken. It is not a major issue, but it could be solved by checking length.

blueyed:
I am assigning it to myself, although I'd be happy for anybody to step in and fix the (mostly) C issues with this.

blueyed:
I've finally come back to this: I've tried to address the mentioned issues.

Please review again.  There's probably still a problem with (not) freeing `->utf8`?!  Please advice.

blueyed:
Rebased it on master.

psychon:
Something trivial (feel free to squash this into your stuff): https://gist.github.com/psychon/72bdf3c72551d1ab003a

I really wanted to fix the memory leak on `->utf8`, but this will be ugly. So far you can just `free()` a `keyb_t` instance and the code does so...

psychon:
In the end this wasn't really bad: https://gist.github.com/psychon/d4ff7b436d389ef6af6f

Squash this into your work, too, test the result and feel free to merge. Note that I didn't even compile-test this, so I would ask you to also test something like this (perhaps run this under `valgrind --leak-check=full --show-reachable=yes` before and after my patch and hopefully you can see some key-related difference under all the noise that is reported?):

    for i = 1, 100 do
        local k = key({ modifiers = {}, key = "ö" })
        for i = 1, 100 do
            k.key = k.key .. "ö"
        end
        collectgarbage("collect")
    end


Ohhhh... I just noticed that this PR breaks reading from `k.key`. I guess you have to modify `luaA_key_get_key()` as well so that it pushes `->utf8` if both `->keycode` and `->keysym` are zero. I guess the above code won't trigger this, but surely `k.key = "sdjfhsifzudfvuizfiuwezr" ; print(k.key)` would do.

psychon:
:+1: (Except for my comment on 6b745392d6c3cda371e860).

Another "feel free": Feel free to ignore my comment and just add this to *the* milestone. :-)

nuko8:
Isn't it possible to eliminate the field `utf8` from `keyb_t` and use the field `keysym` instead?

According to http://xkbcommon.org/doc/current/xkbcommon_8h.html#a79e604a22703391bdfe212cfc10ea007:

> any Unicode/ISO 10646 character in the range U0100 to U10FFFF can be represented by a keysym value in the range 0x01000100 to 0x0110FFFF. 

(Note: We don't need to take the values above U+10FFFF into consideration these days.)

This description indicates that, to check if a given `keysym` represents a unicode or not, only checking its 24th bit, i.e., calculating the boolean value of `keysym & 0x01000000`, is enough.

If we make use of this bit check, we don't need to convert a unicode keysym into the corresponding UTF-8 bytes right after we recognize it's a unicode, but can put off the conversion until we actually need the corresponding UTF-8 bytes, keeping the value in the field `keysym` till then.  

When we need the conversion, we check the bit first, allocate memory if it's a unicode keysym, convert the keysym info the corresponding UTF-8 bytes, push the resulting bytes into the Lua's stack, and release the memory right after that.

Conversely, at any places where we don't want to allocate memory, convert a given UTF-8 bytes into the corresponding unicode, add  0x01000000 to it, and save the resulting integer in the field `keysym`.

If this bit check scheme were made possible, the burden of memory management for UTF-8 bytes could be made far much lighter and easier to handle.

blueyed:
@nuko8 
That sounds good!
Do you want to give it a try, based on my branch / this PR?

nuko8:
@blueyed

The patch below looks woking well for each case of
```
awful.key({ modkey }, "ü", function() naughty.notify({ text = "ü" }) end)
awful.key({ modkey }, "udiaeresis", function() naughty.notify({ text = "udiaeresis" }) end)
awful.key({ modkey }, "#34", function() naughty.notify({ text = "#34" }) end)
```
which are basically the same as  in your initial post.

NOTE:  I guess keycode 34 is the one for the U umlaut key on your keyboard.  For tests, I had to use keycode 41 instead of keycode 34, because, with my keyboard, one of the keysyms of keycode 34 is ampersand and thus `Mod4 + #34` means `Mod4 + Shift + "7"`, which takes my window somewhere unless I happen to be on tag No.7 :-)   Keycode 41 corresponds to the U umlaut key for my keyboard.

The patch was made against the current master, because all the modifications were done only with `objects/key.c`; otherwise, I would have to make four of the five files revert all the way :-): https://gist.github.com/nuko8/08db43c3cbfa6b3103ca .

For convenience, the whole file is given here: https://gist.github.com/nuko8/e53c0b1523eacf6f2455 .

Since I'm not sure if you're satisfied with the patch, I'd be happy if you would give your views on that.

psychon:
Looks mostly good to me (no idea if all of those changes are really necessary, but it seems good). My only comment is on the `switch(len)`: Is this doing the same thing that `g_utf8_get_char_validated` would do? https://developer.gnome.org/glib/stable/glib-Unicode-Manipulation.html#g-utf8-get-char-validated (I guess so and I guess that one does more sanity checks; plus just by calling that function we better document what is going on)

nuko8:
Thank you for the comment.

> `g_utf8_get_char_validated()`

Oh, I completely forgot I could use glib directly in the C code:-) 

Done and tested. Good result obtained. Thanks.

> no idea if all of those changes are really necessary, but it seems good

The reason I did something seemingly unnecessary was that i was worrying about possible glitches which might be caused by subtile differences, if any, between Xlib and libxkbcommon.  If you're feeling some bad smell there, I should rethink about that again.

> if (key->keysym >= 0x100 && key->keysym <= 0x10ffff) key->keysym |= 1 << 24;, so a comment would be nice. 

This is the unicode-to-keysym conversion: http://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#keysym_encoding

Added a comment about that above the `if` statement.

> Below 256 there is a one-to-one mapping? 

Yes, no conversion.

> and above 0x10ffff... something?

They are considered invalid these days.  The glib function should/is supposed to take care for that :-)

> Should the function error in that case?

For that, the return value of the glib functions is done.

The diff and file were all updated.

blueyed:
Thanks a lot @nuko8.

Please push this as a new PR (with you commit(s) on top of mine), so we could merge and/or follow up there.

psychon:
Sorry for being annoying.

I think the following should get a temporary variable instead of these casts. Also instead, of `4`, this can use the actual length.

    key->keysym = (uint32_t)g_utf8_get_char_validated(str, 4);
    if ((int32_t)key->keysym < 0)

And now I wonder: What happens when one gives two codepoints to this function (e.g. `üä`). It will silently ignore the second one. If this should be handled, too, we seem to need `g_utf8_validate()` and `'\0' == *g_utf8_next_char(p)`...? Or something with `g_utf8_find_next_char`?

nuko8:
> Sorry for being annoying.

No, not really.

> I think the following should get a temporary variable instead of these casts. 

The first cast is for a documentation purpose: `gunichar = guint32 --> uint32_t = xkb_keysym_t`  So, it's possible to omit it.  

I'm afraid the second cast cannot be omitted because the manual of the glib function says:

> If p points to a partial sequence at the end of a string that could begin a valid character (or if max_len is zero), returns (gunichar)-2; otherwise, if p does not point to a valid UTF-8 encoded Unicode character, returns (gunichar)-1.

Perhaps, use of a `gunichar` temp variable makes the intention clearer.  I'll do that.

> instead, of 4, this can use the actual length.

The manual also says (note: the 4 was given to the parameter `max_len` in this context)

> max_len the maximum number of bytes to read, or -1, for no maximum or if p is nil-terminated

So, I thought it was not a problem because I assumed the number of unicodes fed into the function was one and that the maximum byte number of a single unicode was 4.

In my intention, the unicodes handled here are confined to those that can be seen on the top of keys.  So I never thought about the case like

> What happens when one gives two codepoints to this function (e.g. üä)

or the case where someone tries
```
awful.key({ modkey }, "üä", function() naughty.notify({ text = "üä" }) end)
```
But I think your argument makes sense.  I'll see to it in conjunction with the number 4: Read a UTF-8 encoded byte string and only take the first consecutive bytes that represents a single unicode character.

If you find that still problematic, let me know.  Always welcome :-)

nuko8:
@psychon 

Reading the glib function name closely, it's talking about `char`.

So, it now looks to me that
```
g_utf8_get_char_validated(str, 4)
```
is just the usage of the function which resolves all your concerns.  

I'd be happy if you would agree with that.

psychon:
My intention was to refuse utf8 strings with more than one codepoint. So for `öä` I'd make the code reject this string while (if I understood you correctly), you plan to have this ignore the `ä` and have this behave just like `ö`.

On the other thing about maximum length, does this bit from the docs mean that if p is nil-terminated, then one can use -1, but if one gives an explicit number, then this has to be the allowed string length? Uhm...

    max_len the maximum number of bytes to read, or -1, for no maximum or if p is nil-terminated

I don't know really. I guess I'll just wait for the right PR and do some tests. If this breaks something, then I can at least say what I'd like to work differently. :-)

nuko8:
I think the "number of bytes" in the phrase "the maximum number of bytes to read" doesn't represent the length of a byte string which is fed into the function, but the number of bytes which is used to represent a single character encoded in UTF-8.

OK, I'm going to push a PR to discuss this issue further there :-)

psychon:
Well, what's the difference between "the maximum number of bytes for a single character encoding" and "the length of the input" when we read a single byte?

http://sources.debian.net/src/glib2.0/2.42.1-1/glib/gutf8.c/?hl=662#L662

In line 612 to 619 it errors out if the length to read is larger than the length that it would have to read. So if I give a single byte with value 0xff to this function and you say max_length = 4, it will try to read 4 bytes.

nuko8:
OK.

Now I understand how useless `max_len` is.   It's not a sort of upper bound or else. 

I'll change the value from 4 to -1.

blueyed:
Closing this in favor of #538.

Thanks for picking this up!

psychon:
After #538 is merged, does it make sense to still apply the other changes from this PR that allow us to drop the xlib dependency? (rebased ontop of the then-current master)

actionless:
xlib is also used for xrdb-related stuff

nuko8:
@psychon , @actionless 

Looks the construction of #538 is causing an unnecessary confusion.  Sorry about that.

First, let me say the conclusion: If #538 is merged, the libX11 dependency is intact.

Now, let me explain about the relation between this PR (#365) and #538.  It's a little bit lengthy and troublesome story.

In appearance, PR #538 has the construction that my work is built on top of #365.  But technically and logically, it doesn't.

I adopted such a construction because @blueyed asked me to do so:

> Please push this as a new PR (with you commit(s) on top of mine), so we could merge and/or follow up there.

Prior to his comment,  I already stated in one of my comments, which was talking about my patch:

> The patch was made against the current master, because all the modifications were done only with objects/key.c; otherwise, I would have to make four of the five files revert all the way

With this, I thought @blueyed came to know my patch didn't depend on his modifications.

Naturally, I assumed that @blueyed had been tracking my comments and already reviewed my patch in gist to make sure of that.

Nonetheless, @blueyed still asked me to make a new PR "with you commit(s) on top of mine".

I concluded, therefore, that his request was based on something related to some policy of this particular development team which I was supposed to follow.  I should have asked @blueyed about that, but I was busy at that time for doing something with @psychon 's comments on my patch.

That's why I adopted the "on top of #365" construction for #538 and why my first commit (338fc8c) of #538 begun cancelling almost all modifications of #365.  That particular commit restores the libX11 dependency.

Hmm, I should not have fabricated a commit history.  That's nothing other than evil.

blueyed:
@nuko8
Sorry, I was not aware that the new PR should have been only against master from the beginning.
Please feel free to rebase/fix that.

nuko8:
@blueyed 

As I wrote in a somewhat different way, I should have asked you immediately when I felt something went wrong.  I'm responsible for that.  

Thank you for your understanding.

psychon:
Now that #538 is merged, does it make sense to do a new PR to use `xkb_keysym_from_name` instead of `XKeysymFromString` etc? (=To make us use less of Xlib)

nuko8:
Contrary to what is written in `XKeysymToString`(3):

> The returned string is in a static area and must not be modified. 

the function does allocate a buffer with `malloc` if a given keysym falls in the range `[0x01000100, 0x0110ffff]`: http://cgit.freedesktop.org/xorg/lib/libX11/tree/src/KeysymStr.c?id=libX11-1.6.3#n129.

So, we have to do something at the caller side to prevent possible memory leaks if `xib_keysym_from_name` has to be replaced with `XKeysymToString` since unicodes in that range cannot be ignored on account of #538.

Issue #185, opened 2015-03-30
Provide a helper to start a command, and return exit code, stdout and stderr


blueyed:
It would be nice to have something like the following in `awful.util`:

    exit, stdin, stdout = awful.util.pread2("echo 1; echo 2>&2; exit 3")

The following could be adopted: http://stackoverflow.com/a/16515126/15690

psychon:
Calling `fork()` from lua seems like a really bad idea...
Your example seems to assume that this wraps `bash -c foo` around the code, I wonder if that really is necessary or if at least two different functions for this could be provided
What would this `pread2` function return? Some lgi wrappers around GIO objects?
What happens to `stderr`? The combination of "stdin" and "stdout" seems weird...
If you meant `stderr` instead of `stdin`: I would like to avoid having some blocking code in awful. I have seen to many "my awesome regularly freezes" bug reports that turned out to be blocking calls to some shell scripts which did slow things. Some callback-based mechanism seems nicer...

The short version of all this: Could you tell us more about what exactly this function does? Something like API docs that explain things?

actionless:
i am using such solution, but it's very far from ideal

actionless:
i am using such solution, but it's very far from ideal:
https://github.com/actionless/awesome_config/blob/devel/utils/async.lua

Elv13:
I have one that works
https://github.com/Elv13/awesome-configs/blob/master/utils/fd_async.lua#L294

Bew78LesellB:
I use asyncshell, from lain : 
https://github.com/copycat-killer/lain/blob/master/asyncshell.lua

It doesn't return stderr, nor exit code, but is very usable to fetch data from scripts commands !

psychon:
Implemented in #406 (although that doesn't give you the exit status, do we need that one?)

blueyed:
@psychon 
> doesn't give you the exit status, do we need that one?

Yes, that can be useful, e.g. for things like `ponymix is-muted` etc.
Please re-open in case you agree.

PR #367, opened 2015-07-28
Add an asynchronous version of io.popen()

b80905:


lib/gears/async_popen.lua
```diff
@@ -0,0 +1,59 @@
...
+local util = require("awful.util")
+
+local async_popen = { mt = {} }
+
+-- Maps temporary filenames to callbacks
+async_popen.callbacks = {}
+
+--- Run a command asynchronously.
+-- @param command The command to run
+-- @param callback The function to be called when the command exits. The output of the command will be passed to it.
```
blueyed: Please use `@tparam` and `@treturn` here, i.e. `@tparam string command ...`.

Also, please wrap lines at max. 80 chars, indenting them in the documentation with two spaces.

lib/gears/async_popen.lua
```diff
@@ -0,0 +1,59 @@
...
+function async_popen.async_popen(command, callback)
```
blueyed: This could return the result from `util.spawn_with_shell`, which would also include the startup-notification id (`sn`) as of https://github.com/awesomeWM/awesome/commit/f12f9b3b97f2ab23406c3193901fbaf06bccf202.

mindeunix:
is there any reason not to use GLib async functions ?

lib/gears/async_popen.lua
```diff
@@ -0,0 +1,59 @@
...
+end
+
+--- Run a command synchronously and return its output. If the given timeout expires, kill the command and return nil.
+-- @param command The command to run
+-- @param timeout The maximum time the command will be run
+-- @return If the command did not time out, return its output, otherwise return nil.
+function async_popen.sync_popen_with_timeout(command, timeout)
+    local fh =
+        io.popen(string.format("timeout %d sh -c '(%s) || true'",
+                               -- ^without the "|| true" if the given command returns 124 this function will think that it timed out
```
blueyed: > without the "|| true" if the given command returns 124 this function will think that it timed out

The 124 comes from `timeout`, right?
You could use `--preserve-status` with it.

OTOH another approach might be to get the PID of the process (as in `sleep 10 &; echo $!`) and then use a `gears.timer` to kill the process after the timeout.
This would not require the `timeout` tool - which is in coreutils, but still.

blueyed:
`sync_popen_with_timeout` will block, won't it?
It should be more like the version in https://github.com/alexander-yakushev/awesompd/blob/master/asyncshell.lua.

blueyed:
> is there any reason not to use GLib async functions ?

I would prefer this, too - and have recommended to use / include the async library from @Elv13 instead (https://github.com/Elv13/awesome-configs/blob/master/utils/fd_async.lua).

But there were also several reports on IRC about glib behaving badly, as in segfaulting..

Therefore I think that it might be good to have a generic approach like this first, and then implement it (alternatively) using glib.

mindeunix:
@blueyed 
> I would prefer this, too - and have recommended to use / include the async library from @Elv13 instead (https://github.com/Elv13/awesome-configs/blob/master/utils/fd_async.lua).

I think it would be good idea.

> But there were also several reports on IRC about glib behaving badly, as in segfaulting..

True, as an example: 
```lua
-- Executes a child program asynchronously
GLib.timeout_add_seconds(GLib.PRIORITY_DEFAULT, 5, function()
    local argv = GLib.shell_parse_argv('/bin/touch /tmp/test')
    local flags = GLib.SpawnFlags.DEFAULT
    local pid, stdin, stdout, stderr = GLib.spawn_async_with_pipes(nil, argv, nil, flags, function() end)
    return true
end)
```
but maybe better try to fix these problems? it would be much better than creating another implementation.

blueyed:
> but maybe better try to fix these problems? it would be much better than creating another implementation.

ACK.

psychon:
Perhaps it's easier to add a function to C with wraps `spawn_async_with_pipes` and returns the PID and FDs?

psychon:
@mindeunix Your example doesn't crash here, but deadlocks. The function argument will be run in the child process and `lgi` doesn't like `fork`. The deadlock occurs in some code that tries to get a mutex.

actionless:
>Perhaps it's easier to add a function to C with wraps spawn_async_with_pipes and returns the PID and FDs?

while `lgi` implementation causes some problems for me and other users i think that's a better way

blueyed:
> your attempt at safely escaping a string for the shell via string.gsub is only a joke.

Agreed.

@b80905 
That should be changed to `string.format('%q', ...)` probably.

> while lgi implementation causes some problems

From my understanding the problems are rather severe (deadlocks and crashes).

blueyed:
There exists an older issue already: https://github.com/awesomeWM/awesome/issues/185

The mentioned solutions there:
 - https://github.com/copycat-killer/lain/blob/master/asyncshell.lua (what inspired this PR)
 - https://github.com/Elv13/awesome-configs/blob/master/utils/fd_async.lua
 - https://github.com/actionless/awesome_config/blob/devel/utils/async.lua (very similar to this PR)

mindeunix:
@psychon 
> Your example doesn't crash here, but deadlocks. The function argument will be run in the child process and lgi doesn't like fork. The deadlock occurs in some code that tries to get a mutex.

Yes, my example  just freezes Awesome WM (as you have already noticed), and this is the only one problem I have been faced with.

> Perhaps it's easier to add a function to C with wraps spawn_async_with_pipes and returns the PID and FDs?

+1

Maybe you can add load_contents_async() too ?

Elv13:
Hello,

Sorry for being unresponsive, I am traveling for the next few months and I don't really get my laptop out of its bag that much.

The issues include both deadlocks and SIGSEGV. This is even more problematic for the async version of dbus, where it will simply have tons on random crashes. Calling something async into an async callback will cause deadlocks.

Another solution may be to port fd_async and https://github.com/Elv13/wirefu to C and exposing the API to lua? That would make Awesome depend on GIO, but beside this, it is a very clean solution well integrated into Awesome event loop. However, that still required some work and I am totally unavailable to do it (this email is sent from Marrakech, the first reliable Internet connection I have in this country until now)

psychon:
This is now implemented in https://github.com/awesomeWM/awesome/pull/406 as `awful.util.spawn_with_line_callback`.

PR #215, opened 2015-05-04
Replace the default theme titlebar icons

Elv13:
> ---------- Forwarded message ----------
> From: Hajime Branko Yamasaki Vukelic <branko@brankovukelic.com>
> To: Julien Danjou <julien@danjou.info>
> Cc: 
> Date: Mon, 20 Apr 2015 22:15:47 +0200
> Subject: Titlebar icons
> Hey, continuing with icons, I've just finished the titlebar set, so
> feel free to use them if you like em.
> 
> Also attaching the SVG sources for layout and titlebar icons so people
> can edit them if they want.
> 
> --
> Branko
> branko@brankovukelic.com
> 
> /dev/blog: brankovukelic.com

PR #369, opened 2015-07-29
New default theme titlebar icons (follow-up)

blueyed:
This is a followup to https://github.com/awesomeWM/awesome/pull/215 with fixes for the build process and the xresources theme.

I wonder if the new set could be also auto-generated, based on the normal or focus variants?

For reference:

    foreach(icon ${ALL_ICONS})
        # Make unfocused icons translucent
        a_icon_convert("_focus" "_normal" ${icon}
            -colorspace rgb -gamma 0.6 -channel A -evaluate Multiply 0.4)
    endforeach()

    foreach(icon ${ALL_ICONS})
        # Make inactive icons grayscale
        a_icon_convert("_active" "_inactive" ${icon}
            -colorspace Gray)
    endforeach()


blueyed:
@actionless 
What do you think about the handling for light backgrounds in https://github.com/blueyed/awesome/commit/f4b10d741819e7dcd90260188a5d7af4ec42892b?

actionless:
for the first time we can do the same as for layout icons -- just to have normal and inverted variants for light and dark colorschemes

later on we can think on some more generalized solution, like using svg-s (cast @worron), he is using it for icons in his config: https://github.com/worron/awesome-config

blueyed:
> later on we can think on some more generalized solution, like using svg-s

That sounds like a good approach.  Therefore the source SVGs should be in the repo probably then.

actionless:
also it seems like we are missing an icon `minimize`: http://awesome.naquadah.org/doc/api/modules/awful.titlebar.html#titlebar.widget.minimizebutton

worron:
Recoloring solid icon is pretty [simple thing](https://github.com/worron/redflat/blob/master/gauge/svgbox.lua#L195L201) to add, no need svg for this. Current raster icon is ok.

Real profit form svg is true vector scaling, which unsupported by `awesome` today. I have some dirty workarounds for this in my config, but would be great if someone add a proper solution for vector scaling.

actionless:
@worron wow, i wasn't expected what it's *so* easy

@blueyed i think we can merge it without this (https://github.com/blueyed/awesome/commit/f4b10d741819e7dcd90260188a5d7af4ec42892b) commit, i will prepare recoloring using worron's method as a separate PR and add it for all the icons in the `xresources` theme

actionless:
here it is: https://github.com/awesomeWM/awesome/pull/386

blueyed:
@worron 
> Real profit form svg is true vector scaling, which unsupported by awesome today

What would be required for this?
I've noticed that you can use svgs as-is, e.g. in the menubar (https://github.com/awesomeWM/awesome/commit/dc41e62f9a58faf723c3eefba729b1e2362f2b2f), and I assume they should be automatically scaled then?

worron:
@blueyed
>What would be required for this?

To be honest I can't say exactly. May be rsvg bindings and redesigning `wibox.widget.imagebox` for a bit.

>you can use svgs as-is,

Yes, but it loaded with default sizes and scaled by cairo like any other raster image then.
Here is an example.

psychon:
If you want, I can look into better SVG support. Open an issue an assign it to me. :-)

Gnome Pixbuf has some concept of a "default size" and it uses that for SVGs. There is also an API for loading an image with a specified size (raster images are then scaled by pixbuf). I guess we should either extend the `awesome.load_image` API or drop it and instead use pixbuf through lgi. No idea how a good lua API for this would look like, but we'd have to stop passing cairo surfaces around for proper SVG support, I guess.

worron:
Thank you for hint about GdkPixbuf API. I'm definitely interested in better svg support. I will open issue after taking a closer look at this example.

PR #370, opened 2015-07-29
add missing parenthesis to a call of screen.focused

aroig:


blueyed:
Thanks!

Will pick this later.

PR #372, opened 2015-07-29
depth for gears.debug

blueyed:
This is useful with big tables and might even be necessary to prevent
infinite recursion.

To prevent the latter, the default should maybe be 10, instead of
`math.huge`?!

### DIFF
--- a/lib/gears/debug.lua
+++ b/lib/gears/debug.lua
@@ -31,12 +31,12 @@ end
 -- @param data Value to inspect.
 -- @param shift Spaces to indent lines with.
 -- @param tag The name of the value.
--- @tparam[opt] int depth Depth of recursion.
+-- @tparam[opt=10] int depth Depth of recursion.
 -- @return a string which contains tag, value, value type and table key/value
 -- pairs if data is a table.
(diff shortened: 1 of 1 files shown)


lib/gears/debug.lua
```diff
@@ -45,8 +47,11 @@ local function dump_raw(data, shift, tag)
...
 
     shift = (shift or "") .. "  "
-    for k, v in pairs(data) do
-        result = result .. "\n" .. shift .. dump_raw(v, shift, k)
+    result = result .. tostring(data)
+    if depth > 0 then
+        for k, v in pairs(data) do
+            result = result .. "\n" .. shift .. dump_raw(v, shift, k, depth - 1)
+        end
     end
```
psychon: Perhaps some indication that things continue? Or using `tostring` instead of `dump_raw` when the depth limit is hit?

blueyed: Good point.
Using `tostring` would not help with trying to keep it short though..

blueyed: Done in 4aa6c339.

psychon:
:+1:

PR #374, opened 2015-07-29
awesome_atexit: keep client order always

blueyed:
This is not only useful for soft restarts, but also when TERMinating
awesome during development/testing.

The drawback appears to be that it would leak a property on the root
window in case it is really meant to be quit.


psychon:
Fine by me

PR #375, opened 2015-07-29
Fix sporadic test failure due to lgi

psychon:
With version 2.0rc5, Busted started cleaning up between tests more thoroughly.
The result was that it managed to unload and reload lgi. However, the C part of
lgi is not safe to be reloaded. This caused sporadic errors in the test suite
runs.

Work around this via a helper script that is run before Busted starts running
tests. When we load lgi in this helper script, it is loaded before Busted starts
saving and restoring everything and thus lgi won't ever be unloaded.

@blueyed Could you test this patch some hundred times and see if this really fixes our problems?

### DIFF
--- a/spec/preload.lua
+++ b/spec/preload.lua
@@ -1,4 +1,4 @@
 -- This script is given to Busted via the --helper argument. Modules loaded here
 -- won't be cleared and reloaded by Busted. This is needed for lgi because lgi
--- is not safe to reload and yet Busted makes us try to do this.
+-- is not safe to reload and yet Busted manages to do this.
 require("lgi")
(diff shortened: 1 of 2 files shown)


spec/preload.lua
```diff
@@ -0,0 +1,4 @@
+-- This script is given to Busted via the --helper argument. Modules loaded here
+-- won't be cleared and reloaded by Busted. This is needed for lgi because lgi
+-- is not safe to reload and yet Busted makes us try to do this.
```
blueyed: > and yet Busted makes us try to do this

I don't understand this?!

blueyed:
:+1:

spec/preload.lua
psychon: Whoops, see the fixup. It now says `and yet Busted manages to do this`.

blueyed:
Committed in https://github.com/awesomeWM/awesome/commit/b75f374d9c301324b5894647deaa021e3eaaa782.  Thanks!

PR #377, opened 2015-07-29
Naughty: default to focused screen

blueyed:
The main change is to default to the focused screen for notifications: https://github.com/awesomeWM/awesome/commit/efe5358a0bdab0d62615a3580445d26529247138

I think that's more sensible, and this thread reminded me of it: http://article.gmane.org/gmane.comp.window-managers.awesome/11100

There could be also support for "all", which would display them on all screens?


psychon:
:+1:

PR #378, opened 2015-07-30
Add util.run_script_async()

b80905:


actionless:
>This should be /bin/sh probably.

 for example google in its shell style guide [1] recommends using bash by default

and `sh` is usually a symlink to either `dash` or `bash`, and that can make widgets not "portable" between, for example, ubuntu and arch

[1] https://google-styleguide.googlecode.com/svn/trunk/shell.xml#Which_Shell_to_Use

blueyed:
> that can make widgets not "portable" between, for example, ubuntu and arch

Therefore I think using `sh` and assuming that it's not bash is good.
`sh` (as in `dash`) is more lightweight.  

It should rather use `util.shell`, which is `util.shell = os.getenv("SHELL") or "/bin/sh"`.

actionless:
>It should rather use util.shell, which is util.shell = os.getenv("SHELL") or "/bin/sh".

:+1:

psychon:
I would replacing all of this with something like `awesome.spawn_with_pipes(cmd, return_stdin, return_stdout, return_stderr)`. `cmd` would be handled just like in `awesome.spawn` and is either a string or a table. The boolean flags do what their name suggests and if one of them is false, the corresponding argument to `spawn_async_with_pipes` would be NULL.

The return values would be `pid, stdin, stdout, stderr` in case of success (where pid is a number and the others are nil if the corresponding boolean argument was false, else an integer describing the file descriptor). On error, the same kind of return as `awesome.spawn` would happen (I don't remember right now what it is).

Lua can then use GLib and construct UnixSomethingStreams if it wants to.

Wouldn't this API be much more flexible?

Based on this you could then implement `awful.util.run_script_async` (why "script"? Also, the spawn functions are async, too. How about `spawn_with_output` or something like that?), but other functions would also be possible.

b80905:
@psychon I've implemented your suggestions. Can you check out the new version of the branch?

psychon:
Sorry for being so annoying. :(

psychon:
How about something like this? https://gist.github.com/psychon/8447f229fd313b5f3e3e

b80905:
@psychon Yeah. You should add that instead.

blueyed:
Re-opening to keep track of it.

blueyed:
Seems to be addressed in https://github.com/awesomeWM/awesome/pull/406.

Thanks again @b80905 for pushing this topic/issue, and sorry again in the name of the team in case we've come across as too picky etc.
Stay awesome! :)

PR #381, opened 2015-08-01
tooltip: expose timer_function and timer

blueyed:
Having access to `timer_function` allows you to change the contents of a
tooltip for when it gets displayed again.  When using the timer /
`timer_function` feature, any call to `set_text`/`set_markup` would be
overruled by the former private `timer_function`.

### DIFF
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -18,7 +18,7 @@
 --             end,
 --         })
 --
--- How to add the same tooltip to several objects?
+-- How to add the same tooltip to multiple objects?
 -- ---
 --
 --     myclock_t:add_to_object(obj1)
@@ -26,7 +26,7 @@
 --
 -- Now the same tooltip is attached to `myclock`, `obj1`, `obj2`.
 --
--- How to remove tooltip from many objects?
+-- How to remove a tooltip from several objects?
 -- ---
 --
 --     myclock_t:remove_from_object(obj1)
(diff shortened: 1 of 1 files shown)


blueyed:
Maybe `set_timeout_function` would be more in line with the current API..

blueyed:
For my config I would like to use a `timer_function`, but return text instead of markup from there, because the latter would require to escape e.g. `&`.

The current implementation does not allow for this, but with this PR you could set `tooltip.timeout_function` explicitly after creating the tooltip object.

I am not certain that's the best approach?!

psychon:
We have `awful.util.escape` for cases where you need markup instead of text.

blueyed:
Okay.  Then only the timer_function that's been used / passed in via `args` should be exposed probably?
Or it could be still exposed, but with a different name? (it's rather confusing to have the same name for the inner and outer function).

psychon:
I'm confused about this stuff, sorry. Why do we want to expose the timeout function and how can this be used to escape markup? Can't that be done "normally" by whatever uses the tooltip?

And yeah, confusing names are confusing. :-)

blueyed:
@psychon 
`timer_function` and `timer` are exposed since 3e9fdea now :)

I've rebased this and only kept small changes to the doc, and how `timer_function` is being called.  Passing in `self` there seems to be useful if you want to refer to the tooltip itself from the function.

psychon:
Ok :+1:

PR #382, opened 2015-08-01
Fix mouse::leave for drawables (titlebar)

blueyed:
Store X/Y of the last motion event in `globalconf`, to have it available
in `event_handle_leavenotify`, where only the current X/Y is available
on the event.  And since the titlebar has only a drawable, but no
drawin, it will not get handled through `drawin_getbywin(ev->event)`.

An alternative might be to use a `drawin` also for the titlebar.

### DIFF
--- a/event.c
+++ b/event.c
@@ -454,6 +454,9 @@ event_handle_motionnotify(xcb_motion_notify_event_t *ev)
         luaA_object_emit_signal(L, -3, "mouse::move", 2);
         lua_pop(L, 2);
     }
+
+    globalconf.last_motion_x = ev->event_x;
+    globalconf.last_motion_y = ev->event_y;
 }
 
 /** The leave notify event handler.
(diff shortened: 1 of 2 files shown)


blueyed:
Is there a better/lighter way?
Storing the last x/y coordinates just to handle leaving titlebars seems like too much overhead?!

psychon:
How about a `globalconf.drawable_under_mouse` that gets updated during enter / motion / leavenotifies? Via such a variable it might be simpler to ensure that `mouse::enter` and `mouse::leave` events really are paired.

Note that `a_xcb_check()` is being clever with motion notify events and drops multiple such events and only handles the last one and also "moves them behind other events" (just not past enter/leave notifies and button press/release). This re-ordering could have weird effects.

So perhaps we should do something for the drawable under the cursor like we do in lua with `wibox.drawable` `drawable._widgets_under_mouse`? Although this would need to properly be unset when the drawable is hidden...

What do you think? I'm not really sure which approach would be simpler.

blueyed:
Superseded by https://github.com/awesomeWM/awesome/pull/461 - closing.

PR #385, opened 2015-08-01
Apply left/top titlebars in event_handle_configurerequest

blueyed:
I have noticed that the HeidiSQL session manager window/dialog will be
moved down by the height of the top titlebar when selecting an entry
from the server list.

While I am not sure if that is the correct way to fix this, and what it
might break, it fixes the issue described above.


psychon:
It's stupid that that thing moves itself where it already is, but yeah, this patch seems fine. Sorry for missing this before.

PR #386, opened 2015-08-02
Xresources theme refinements

actionless:


actionless:
and a screenshot:
2015-08-02--1438524950_3553x64_scrot

blueyed:
It uses `#657b83` for my `theme.fg_normal` which does not color the titlebar icons fully:

focused
unfocused

That's probably how the pattern/mask works?
Is it meant to use a single color for the images/icons in both cases?

Then, it would be nice to move this (the recoloring of titlebar icons and layoutbox icons) to the assets module, too - so that it can be used in a derived theme more easily, e.g. by calling a function providing the color(s)?!
But there might be a better approach, e.g. by providing / wrapping it into a `setup` method that does all this, and where you can provide a `theme` object.  This basically should use `theme` then instead of `local theme = dofile("@AWESOME_THEMES_PATH@/default/theme.lua")`.

This might require to enhance the theme handling in awesome itself, so that a theme would/could return a function instead of a table, which is meant to be called, and could provide an optional base theme?!

My idea is to have a Solarized theme, based on this one, but currently it requires to copy'n'paste stuff that needs to get redone, after I've setup `theme.fg_normal` etc.

btw: `theme.font          = "sans 8"` is not necessary, since it is provided by the default theme already.

actionless:
strange, try to merge latest master, it should like that (i did screens with light and dark colorschemes):
2015-08-02-195046_3840x2160_scrot
2015-08-02-195139_3840x2160_scrot

actionless:
>That's probably how the pattern/mask works?
Is it meant to use a single color for the images/icons in both cases?
 
 
-

>A drawing operator that paints the current source using the alpha channel of pattern as a mask. (Opaque areas of pattern are painted with the source, transparent areas are not painted.)
http://www.cairographics.org/manual/cairo-cairo-t.html#cairo-mask

>Then, it would be nice to move this (the recoloring of titlebar icons and layoutbox icons) to the assets module, too - so that it can be used in a derived theme more easily, e.g. by calling a function providing the color(s)?!
i will do that, but the rest seems to be outside of scope of this PR

actionless:
@blueyed done

blueyed:
> strange, try to merge latest master, it should like that (i did screens with light and dark colorschemes):

Yes, with master it's better - given the new icons.

> @blueyed done

Awesome!

What do you think about using the border color for normal titlebar icons?

    theme = theme_assets.recolor_titlebar_normal(theme, theme.border_normal)

For inactive windows that makes them less "distracting" / outstanding.

actionless:
@blueyed can you better do some screenshot with dark or both colorschemes? it will be easier to understand the idea

lib/gears/surface.lua
```diff
@@ -70,6 +70,25 @@ function surface.get_size(surf)
     return w - x, h - y
 end
 
+
+--- @TODO
```
actionless: i have no idea what to write here

blueyed:
Removing the "next" milestone: @actionless please address/comment on the last comments.

actionless:
@blueyed on which of them? i thought i have resolved all of them already

blueyed:
@actionless 
Ok then, and sorry.  I've only skimmed the page and it looked like it.
Will move it forward again and review/merge it soon.

actionless:
oops, actually i had one unresolved todo, but now it's done

blueyed:
Merged / rebased on master.

Thanks!

PR #388, opened 2015-08-02
Add placement.next_to_mouse, use it for tooltips

blueyed:
Ref: https://github.com/awesomeWM/awesome/commit/9883d3484dd8b335f589235103fab6fe4d80c81f#commitcomment-12483313

### DIFF
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -21,6 +21,7 @@ local capi =
 local client = require("awful.client")
 local layout = require("awful.layout")
 local a_screen = require("awful.screen")
+local dpi = require("beautiful").xresources.apply_dpi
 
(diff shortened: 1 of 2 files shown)


actionless:
now placement is fine :+1: 
but with tooltip itself there is one more problem: when moving mouse cursor from the button too quickly tooltip is not disappearing unless hovering mouse in and out again

blueyed:
> when moving mouse cursor from the button too quickly tooltip is not disappearing unless hovering mouse in and out again

That's because of https://github.com/awesomeWM/awesome/pull/382: there's no leave event.

psychon:
Assuming you tested all four cases (left, right, above, below), this is fine by me. Thanks!

Issue #328, opened 2015-07-18
tests/run.sh: Dependencies and how to run?


psychon:
First:

    sh: 1: xdotool: not found

and hence I changed:

```diff
diff --git a/tests/test-urgent.lua b/tests/test-urgent.lua
index c33f01c..a9b3cf2 100644
--- a/tests/test-urgent.lua
+++ b/tests/test-urgent.lua
@@ -46,7 +46,11 @@ local steps = {
   function(count)
     if count == 1 then
       -- Setup: switch to tag.
-      os.execute('xdotool key super+2')
+      root.fake_input("key_press", 133)
+      root.fake_input("key_press", 11)
+      root.fake_input("key_release", 11)
+      root.fake_input("key_release", 133)
+      --os.execute('xdotool key super+2')
 
     elseif awful.tag.selectedlist()[1] == tags[1][2] then
       assert(#client.get() == 1)
```

However, this depends on the specific key codes for Mod and 2 and is non-portable.
Ideas?

The other issue I have can be seen here: https://gist.github.com/psychon/67f795a87a97f67d8128
Can we somehow make this work for not-installed awesome? For example, something like this can be use to "hack up" a "working" theme:

    sed -e 's:/usr/local/share/awesome/themes/default/theme:./themes/sky/theme:' -i lib/beautiful/init.lua
    sed -e 's:/usr/local/share/awesome/themes/default/theme:./themes/sky/theme:' -i awesomerc.lua
    sed -e 's:awesome_icon.*:awesome_icon = "../icons/awesome16.png":' -i themes/sky/theme.lua
    sed -e 's:/usr/local/share/awesome:..:' -i awesomerc.lua -i themes/sky/theme.lua
    sed -e '/titlebar/s:\.\./::' -i themes/sky/theme.lua

(But I guess this should be replaced by e.g. `sed -e 's:@AWESOME_THEMES_PATH@:./themes/sky/theme:' awesomerc.lua.in > build/awesome.rclua` so that it works independently of the prefix)

Ideas?

actionless:
about the latest part, mb to add a function instead of using build-time template? ie:
```lua
local themes_path = beautiful.get_themes_path()
theme.titlebar_close_button_normal = themes_path .. "/default/titlebar/close_normal.png"
```

psychon:
This makes the functional tests totally work for me (by creating a theme that points to the not-installed versions of our PNG files):

https://gist.github.com/psychon/f9f556889a30359ed83d

[...]

PR #391, opened 2015-08-04
tests: use root.fake_input instead of xdotool

blueyed:
Ref: https://github.com/awesomeWM/awesome/issues/328


psychon:
Thanks, :+1:

PR #392, opened 2015-08-05
Travis: add lgi 0.7.0 to the build matrix

blueyed:
This makes sure that the build and tests succeed with the minimal required
version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,9 @@ env:
   matrix:
     # Note: luarocks does not work with Lua 5.0.
     # - LUA=5.0 LUAPKG=50  LUALIB=/usr/lib/liblua50.so
-    - LUA=5.1 LUAPKG=5.1 LUALIB=/usr/lib/x86_64-linux-gnu/liblua5.1.so
-    - LUA=5.2 LUAPKG=5.2 LUALIB=/usr/lib/x86_64-linux-gnu/liblua5.2.so BUILD_APIDOC=true
+    - LUA=5.1 LGIVER=      LUAPKG=5.1 LUALIB=/usr/lib/x86_64-linux-gnu/liblua5.1.so
+    - LUA=5.2 LGIVER=      LUAPKG=5.2 LUALIB=/usr/lib/x86_64-linux-gnu/liblua5.2.so BUILD_APIDOC=true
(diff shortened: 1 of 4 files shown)


psychon:
Fine by me, but travis failed with lgi 0.7.0 (apparently awesome didn't start up, but we don't get any error message?)

psychon:
Thanks for this. Should we just merge this and open a new issue about "awesome does not work with lgi 0.7.0" or try to fix this here (even though this PR is about something else)?

blueyed:
I think it should be fixed here, before merging it.

In case of merging it anyway, the new, failing build should be added to Travis' "allowed failures", but then it's pointless anyway.

I've tried to reproduce it locally, but failed to do so.. :/

psychon:
Yup, I cannot reproduce this either.

You could try adding something like the following. Either it will crash even more badly or we might get some helpful information about which lua code causes this. It will try to call `debug.traceback` while dying. For lua 5.1 `lua_pushglobaltable` and the following `lua_getfield` need to be replaced with `lua_getfield(L, LUA_GLOBALSINDEX, "debug");`, I guess.
```diff
diff --git a/awesome.c b/awesome.c
index fc10280..ce1a962 100644
--- a/awesome.c
+++ b/awesome.c
@@ -299,13 +299,33 @@ a_glib_poll(GPollFD *ufds, guint nfsd, gint timeout)
 
     return res;
 }
+static int traceback (lua_State *L) {
+    lua_pushglobaltable(L);
+  lua_getfield(L, -1, "debug");
+  if (!lua_istable(L, -1)) {
+    lua_pop(L, 1);
+    return 1;
+  }
+  lua_getfield(L, -1, "traceback");
+  if (!lua_isfunction(L, -1)) {
+    lua_pop(L, 2);
+    return 1;
+  }
+  lua_call(L, 0, 1);  /* call debug.traceback */
+  return 1;
+}
 
 static void
 signal_fatal(int signum)
 {
+    lua_State *L = globalconf_get_lua_State();
     buffer_t buf;
     backtrace_get(&buf);
-    fatal("signal %d, dumping backtrace\n%s", signum, buf.s);
+    fprintf(stderr, "signal %d, dumping backtrace\n%s\n", signum, buf.s);
+    /* try to dump lua traceback? */
+    if(traceback(L))
+        fprintf(stderr, "%s\n", lua_tostring(L, -1));
+    fatal("foo");
 }
 
 /** Function to exit on some signals.
```

PR #393, opened 2015-08-05
Ci script improvements

psychon:
Some stuff, the main thing being that I can run `tests/run.sh` now, too, even though I don't have an up-to-date version of awesome installed.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -60,22 +60,20 @@ XDG_CONFIG_HOME="./"
 export LUA_PATH
 export XDG_CONFIG_HOME
 
-awesome_log=/tmp/_awesome_test.log
-echo "awesome_log: $awesome_log"
-
 cd - >/dev/null
 
-
-kill_childs() {
+# Cleanup on errors / aborting.
+cleanup() {
     for p in $awesome_pid $xserver_pid; do
         kill -TERM $p 2>/dev/null || true
     done
+    rm -rf $tmp_files || true
 }
-# Cleanup on errors / aborting.
-set_trap() {
-    trap "kill_childs" 0 2 3 15
-}
-set_trap
+trap "cleanup" 0 2 3 15
+
+tmp_files=$(mktemp -d)
+awesome_log=$tmp_files/_awesome_test.log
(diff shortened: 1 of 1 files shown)


tests/run.sh
```diff
@@ -62,47 +60,68 @@ XDG_CONFIG_HOME="./"
...
+    max_wait=60
+    while true; do
+        set +e
+        eval reply="\$($2)"
+        ret=$?
+        set -e
+        if [ $ret = 0 ]; then
+            break
+        fi
+        max_wait=$(expr $max_wait - 1)
```
blueyed: That needs an `|| true`: `expr 1 - 1` will exit with code 1.
See 938987a1.

tests/run.sh
```diff
@@ -62,47 +60,68 @@ XDG_CONFIG_HOME="./"
...
+        if [ "$max_wait" -lt 0 ]; then
```
blueyed: ..or use `-le 1` here?!

tests/run.sh
```diff
@@ -62,47 +60,68 @@ XDG_CONFIG_HOME="./"
...
-done
+wait_until_success "setup xrdb" "echo 'Xft.dpi: 96' | DISPLAY='$D' xrdb 2>&1"
 
+# Use a separate D-Bus session; sets $DBUS_SESSION_BUS_PID.
+eval $(DISPLAY="$D" dbus-launch --sh-syntax --exit-with-session)
 
-AWESOME_CLIENT="$root_dir/utils/awesome-client"
+# Not in Travis?
+if [ "$CI" != true ]; then
+    # Prepare a config files pointing to a working theme
```
blueyed: s/files/file/

psychon:
I rebased this onto your "add lgi 0.7.0"-branch. My plan is to wait until that other branch is merged and then rebase this onto master again and only merge this then.

Note that I removed your `kill -0` trick again. The timeout handles this situation as well and it does not work for my helper function when it is used via `xrdb` to wait for the X11 server to start up.

psychon:
Rebased, can this be merged?

PR #396, opened 2015-08-07
feat(lib: awful: layout: suit: floating): add 'resize_jump_to_corner'

actionless:


psychon:
Fine by me I guess, since apparently the floating layout is the only one doing this

actionless:
nope, actually cursor is jumping to the corner when resizing on `tile` layout as well, but there it seems to be a bit more complicated so i am going to add it later when figure out a better way

psychon:
Ok, in that case it would be better to have one central place to configure this, instead of per-layout, no?

actionless:
yes but i have no idea where

i was thinking about `awful.mouse....` but it seems to be a cyclic import then

actionless:
so now the only question remains, where to put common setting for 'resize_jump_to_corner' behavior (btw, feel free to suggest another name for it)

psychon:
Good question. So let's just leave it like this.

PR #398, opened 2015-08-09
Widget context

psychon:
This changes the `wibox` argument that widgets get in `:draw` to a `context` table. This contains an (optional (needed for titlebars)) wibox key, the drawable etc. This also adds a `screen` and a `dpi` entry, allowing widgets to make their look depend on the DPI.

To also allow a widget size that depends on the dpi, the `:fit` callback is also changed to get this context argument (**this is an API change that can break configs**). Based on this @blueyed might implement DPI-support for the `textbox` widget.

Since we cache the results of `:fit`, this code is careful not to break the cache. The same `context` table will be used as long as the `wibox.drawable` is not moved to another screen. This is enough so that the cache continues to work.

If we really change the widget API to allow partial redraws, another API break might follow later. Do we want bleeding edge users to live "on a bumpy road"?

@blueyed This is mostly a request for comments. How much do you want the textbox to incorporate the DPI of the screen? Is it ok to break the API for this?

@Elv13 API break. Opinions?

### DIFF
--- a/lib/awful/titlebar.lua
+++ b/lib/awful/titlebar.lua
@@ -76,7 +76,11 @@ local function new(c, args)
 
     local ret
     if not bars[position] then
-        ret = drawable(d, nil, "awful.titlebar")
+        local context = {
+            client = c,
+            position = position
+        }
+        ret = drawable(d, context, "awful.titlebar")
         local function update_colors()
             local args = bars[position].args
             ret:set_bg(get_color("bg", c, args))
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -83,7 +83,7 @@ end
 local function set_geometry(self)
     local my_geo = self.wibox:geometry()
     -- calculate width / height
-    local n_w, n_h = self.textbox:fit(-1, -1)
+    local n_w, n_h = self.textbox:fit(nil, -1, -1) -- Hack! :(
     n_w = n_w + self.marginbox.left + self.marginbox.right
     n_h = n_h + self.marginbox.top + self.marginbox.bottom
     if my_geo.width ~= n_w or my_geo.height ~= n_h then
--- a/lib/awful/widget/graph.lua
+++ b/lib/awful/widget/graph.lua
@@ -158,7 +158,7 @@ function graph.draw(_graph, context, cr, width, height)
     end
 end
 
-function graph.fit(_graph, width, height)
+function graph.fit(_graph)
     return data[_graph].width, data[_graph].height
 end
 
--- a/lib/awful/widget/graph.lua
+++ b/lib/awful/widget/graph.lua
@@ -69,7 +69,7 @@ local properties = { "width", "height", "border_color", "stack",
                      "stack_colors", "color", "background_color",
                      "max_value", "scale" }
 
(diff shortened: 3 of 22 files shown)


psychon:
I cannot reproduce the build failure with lua 5.1 locally :-(

psychon:
...however, looking at the API docs shows that `xpcall` does not pass arguments to the called function: http://www.lua.org/manual/5.1/manual.html#pdf-xpcall
```sh
$ for lua in lua5.1 lua5.2 ; do echo $lua ; $lua -e 'return xpcall(print, function(...) print("in error function", ...) end, 1, 2, 3)' ; done 
lua5.1

lua5.2
1	2	3
```
I fixed up the corresponding commit to use an anonymous function instead of trying to pass arguments through `xpcall`. This should fix the travis build.

blueyed:
API break seems to be OK for me, given the new possibilities and the generality of the new `context` argument.

Elv13:
/me still traveling and unavailable, at least I am out of Morocco now, so I have reliable Internet again

This will somewhat break some of my stuff, but overall I am happy with the change. This "wibox" parameter was rarely useful. The context one open lots of new possibilities.

psychon:
Rebased, because it conflicted with the new `widget.visible` property

blueyed:
Can the API break for e.g. `textbox:fit()` be detected from libs/plugins, apart from wrapping it in `pcall`?

I was using the following for cyclefocus:

            local _, text_h = textbox:fit(-1, -1)

Using a new method might solve this (e.g. `fit_with_context`), where you could check if it exists - but there might be a better way.

psychon:
Since this `-1, -1` is only ever used (AFAIK) on textboxes, how about adding some methods to the textbox for this?
For example `get_layout_size()` or `get_preferred_size()`? Perhaps this could also be extended with a method where a width is specified and we want to know the required height (so that e.g. naughty can apply a maximal width to notifications), but I'm not sure how useful that is (and if it is useful, do we also need "height is specified, which width is needed"?).

psychon:
The above would be the `proper fix`. I just came up with a quick hack:

    w:fit(-1, -1, -1)

In the old API, the extra argument is ignored. Since the textbox does not use the context argument (yet, didn't you want to make the textbox use the DPI of the screen? That's now possible. Want to submit a PR?) the `-1` will be fine as a context.

actionless:
> make the textbox use the DPI of the screen?
it is already, or do you mean any other changes than this?
https://github.com/awesomeWM/awesome/blob/master/lib/wibox/widget/textbox.lua#L157

psychon:
@actionless That's "one global DPI". I mean something like `ctx:set_resolution(beautiful.xressources.get_dpi(context.screen)` so that a different DPI on each screen is supported. As far as I remember, blueyed mentioned that he wanted to do this and this is the original reason for adding this `context` argument.

blueyed:
> As far as I remember, blueyed mentioned that he wanted to do this

True.  It is an older branch I have laying around and will come back to eventually.

> w:fit(-1, -1, -1)

This is good enough for me currently.  Thanks!

psychon:
@blueyed by the way, for detecting how to call fit, you could check `awesome.version` and do something based on that

PR #353, opened 2015-07-25
Add gears.matrix / make wibox.layout.base smaller

psychon:
This does two things: It adds a module to gears with some useful functions for working with cairo matrixes and it does a step in getting rid of wibox.layout.base.

PR #404, opened 2015-08-12
New widget system

psychon:
Here is the PR that brings us a new widget system which supports partial redraws and hopefully brings lots of fun. Hopefully 2d7368ca632c7df778a1fba09a23c009c8dac44c explains the new state well enough.

Looking forward to your complaints. :-)

This depends on #398 and #353. I guess I should rebase this on master after those are merged.


psychon:
Heh. CI failed badly (whoops) and I also just noticed that this doesn't handle the new `.visible` property correctly yet. Will have to fix both.

psychon:
One more issue:
Caches aren't properly cleared. In the PR as-is, `wibox.drawable` clears the caches for `:fit` and `:layout` when the corresponding signals are emitted. That means that when a widget is not part of any wibox, its caches aren't cleared.
This can likely be demonstrated by this:
```lua
local t = wibox.widget.textbox("")
local lay = wibox.background(t)
print(wibox.widget.base.fit_widget(nil, lay, 100, 100))
t:set_text("foo")
print(wibox.widget.base.fit_widget(nil, lay, 100, 100))
```
Both calls to `fit_widget` will likely produce the same results.

I guess I will just drop these caches for now.

Elv13:
I totally agree with the intended results. For a first look on my smartphone, the parts I looked at look sound. I really don't have time right now and often don't even have Internet in the first place. I will take a closer look soon. Does it handle widgets that draw outside of their area (shadow, powerarrow, glow, mouseover zoom effect, etc)?

psychon:
@Elv13 No and yes. No, widgets must not draw outside of their area and definitely should not unset the clip, because partial redraws would break stuff. However, widgets can place other widgets outside of their own bounds. So a "glow hack" would consist of a dummy widget that just places some other widget so that it gets more space allocated.

In an older version I had some extra "margin" property and widgets were automatically enlarged by their margin. However, this made the code more complicated and (as explained above) doesn't add new features.

For an example where a widget is placed outside of the parent I wrote a zoom demo on an older version of this branch: https://github.com/psychon/awesome/blob/new_widget_system/awesomerc.lua#L172-L237
This function returns a widget which demos animations (performance isn't better than with current git master) and which places a child widget "weirdly".

`get_animation_stage` returns a number between 0 and 1 which decides what the current stage should look like. The `before_draw_children` and `after_draw_children` functions fade the widget in and out (at the maximum zoom, the child widget is fully transparent aka "not visible anymore"). Here, frames are also counted. The `layout` function does the actual animation: The child widget is placed outside of the bounds of the widget that is being created. (The `fit` function is forwarded to the child widget, so the sizes of both widgets are the same)

All of this is already possible with our current layout system, but it seemed like a nice demo. Also, in that branch I have a boolean `wibox.drawable.visible_redraws` that allows to debug which areas are redrawn. This showed that only the part that really changed is redrawn.

psychon:
Superseded by #435

PR #405, opened 2015-08-12
Add "opacity" property for widgets

kindlycat:


blueyed:
:+1: (without having tested it)
Thanks!

psychon:
Fine with me, but I wonder if we should skip the group if the opacity is 1. That should result in less overhead for the common case.

psychon:
Thanks! :+1:

PR #406, opened 2015-08-13
Spawn with pipes

psychon:
This adds `awful.util.spawn_with_lines` (TODO: Suggestions for a better name?) that spawns a program and captures its standard output line by line, calling a lua function on each line. Additionally, another function is called when this pipe is closed.

Ref: https://github.com/awesomeWM/awesome/pull/378

### DIFF
--- a/lib/awful/util.lua
+++ b/lib/awful/util.lua
@@ -22,6 +22,8 @@ local pairs = pairs
 local string = string
 local lgi = require("lgi")
 local Pango = lgi.Pango
+local Gio = lgi.Gio
+local GLib = lgi.GLib
 local capi =
 {
     awesome = awesome,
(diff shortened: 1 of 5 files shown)


psychon:
I did some experiments and now added a `tostring`. Apparently older lgi versions gave us an instance of GLib.Bytes or something like that instead of a string. This should fix all CI failures.

blueyed:
Awesome! Thanks!

lib/awful/util.lua
```diff
@@ -110,6 +112,61 @@ function util.spawn_with_shell(cmd, sn)
...
+--- Spawn a program and asynchronously capture its output line by line.
+-- @tparam string|table cmd The command.
+-- @tparam function line_callback Function to call with each line of output.
+-- @tparam[opt] function done_callback Function to call when no more output is
+-- produced.
+-- @treturn[1] integer The forked PID.
+-- @treturn[2] string Error message.
+function util.spawn_with_lines(cmd, line_callback, done_callback)
+    -- Spawn without startup notification and with returning a pipe for stdout
+    local pid, _, _, stdout, _ = capi.awesome.spawn(cmd, false, false, true, false)
```
blueyed: I wonder if `stderr` should also be captured and passed to the callback function?
Could `start_read` read both stdout and stderr, and call the callback accordingly from what was read from?

lib/awful/util.lua
```diff
@@ -110,6 +112,61 @@ function util.spawn_with_shell(cmd, sn)
     end
 end
 
+--- Spawn a program and asynchronously capture its output line by line.
+-- @tparam string|table cmd The command.
+-- @tparam function line_callback Function to call with each line of output.
```
blueyed: Please document the signature of the callback function, e.g. that it gets `line_from_stdout`.

blueyed:
Waiting for an update.. please move to "next" then.

lib/awful/util.lua
```diff
@@ -110,6 +112,94 @@ function util.spawn_with_shell(cmd, sn)
     end
 end
 
+--- Spawn a program and asynchronously capture its output line by line.
```
blueyed: Missing an "and".  I will amend that.

PR #408, opened 2015-08-13
Strict fit

psychon:
This makes sure that `wibox.layout.base.fit_widget` does not return larger numbers than were given to it. Due to this, some things in the test suite are shuffled around (and we get rid of `XXX` comments!).

While looking into this, I found some weirdness in the align layout. Sadly I failed at producing a test case for it.


psychon:
Hm. The first time this failed travis because `assert(#client.get() == 1)` failed in `test-urgent.lua`. I just rebased this due to merge conflicts with current master and now everything worked. Perhaps we have some flaky test?

@blueyed Opinions on this change?

blueyed:
> Perhaps we have some flaky test?

Might be.

PR #409, opened 2015-08-14
Tooltip: re-add call to placement.no_offscreen

blueyed:


psychon:
:+1: (Although these tooltips are starting to annoy me... but that's a different issue)

blueyed:
> Although these tooltips are starting to annoy me

Which ones? The ones on the tasklist are not in master, are they?
Feel free to create a new issue for it.

psychon:
Well, I switch between screens via a key binding that puts the cursor in the top left corner (which is also where my layout box is). So after switching screens, "something" (the tooltip) appears there, I notice that something happened and get distracted for a moment. I consider this my own problem and not really a flaw with the tooltips (plus I should just switch to `awful.screen.focus`).

Issue #411, opened 2015-08-14
Add property::geometry signals to drawin/drawable


blueyed:
If any of x/y/width/height changes, `property::geometry` should be emitted for drawins and drawables, too - similar to clients.

(This might then be used e.g. by tooltips to automatically place themselves again, if their text change causes the geometry to be changed.)

PR #412, opened 2015-08-14
drawin: lazily configure drawin windows

blueyed:
Ref: https://github.com/awesomeWM/awesome/issues/411.
Source: https://github.com/awesomeWM/awesome/pull/174#commitcomment-12708326.

### DIFF
--- a/event.h
+++ b/event.h
@@ -31,6 +31,9 @@
 /* luaa.c */
 void luaA_emit_refresh(void);
 
+/* objects/drawin.c */
+void drawin_refresh(void);
+
 /* objects/client.c */
 void client_focus_refresh(void);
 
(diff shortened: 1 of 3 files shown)


psychon:
As I already wrote in the comment where I proposed this:

    (Now that I just read through the diff again, I think that drawin_moveresize() is wrong:
     It should change w->geometry first and only afterwards emit signals, so that the other
     properties are already updated when the signals occur, similar to how this is done in
     client_resize_do().)

If you want, I will write that later. I'm not on my computer right now.

Basically, I would do `area_t old_geometry = w->geometry` at the beginning of `drawin_moveresize` and at the end (after everything else that is already done) it should do e.g. `if (old_geometry.x != w->geometry.x) emit_property::x;`.

Otherwise e.g. changing the bounding or the clip shape inside a `property::x` signal won't work correctly. Sorry for not fixing this before I wrote that comment.

blueyed:
It wasn't clear to me if the comment was meant for your patch or the diff from the other PR/commit.

> If you want, I will write that later. I'm not on my computer right now.

Please do.

psychon:
Merged in 9408b61a974bc351e285592f36b2e325c66d5cf2 (with my changes)

PR #413, opened 2015-08-14
tooltips: re-place them on width/height changes

blueyed:
If the dimensions of a tooltip change, e.g. after the text has been
changed, they are now placed again.

Note that this only works properly with https://github.com/awesomeWM/awesome/pull/412.


psychon:
:+1: but I wonder why it only works with the lazy configures. Is this the problem with leave events?

blueyed:
> Is this the problem with leave events?

Yes.  It still works as-is, but causes flickering always, and might cause them to dis-appear.
But that also happens when no re-placing them.

Issue #245, opened 2015-05-26
Maximized Firefox windows not restored properly after fullscreen (F11)


blueyed:
TEST CASE:

1. Tiled layout, with a maximized Firefox window (Mod4-m).
2. Press F11 to switch Firefox into fullscreen mode.
3. Press F11 again.

There will be a gap (~35px) between awesome's taskbar and the Firefox window.
The Firefox window will not be recognized as "maximized" anymore, but only "floating".
The bottom of the Firefox window is aligned to the bottom (and not overlapping), reducing the window's height effectively (by ~35px) after leaving fullscreen mode.

The same behavior occurs when using awesome's fullscreen shortcut (`mod4+f`).

aelspire:
With Chromium something similar occur, after leaving fullscreen mode "maximized" state is lost, but without gap issue and it stay floating only if I made it floating before. I also installed Firefox and tested it - Firefox work the same too, I cannot reproduce gap issue (awesome remember previous floating client position so maybe this is cause of the gap). Losing "maximized" look to me like limitation, I think the cause of such behavior is:

1. Firefox get "maximized"
2. When toggling fulscreen on - "maximized" mark was turned off.
3. When toggling fulscreen off - "fullscreen" mark was turned off, but WM and application do not have memory of previous state so client is not marked as "maximized", so if it was floating before, awesome restore previous position and size.

I tested it with:
1. Make Firefox floating
2. Resize and move Firefox window
3. Maximize Firefox
4. Press F11 or mod4+f to switch Firefox into fullscreen mode
5. Press F11 or mod4+f again
Result: Firefox lost "maximized" state but floating window is correctly restored wit size and position like in 2. 

Losing maximized state occur with all clients, not only Chomium and Firefox.

blueyed:
The problem with Firefox is that it forgets about the window's maximized state when coming back from fullscreen.  Seems to be this (old) bug: https://bugzilla.mozilla.org/show_bug.cgi?id=425086

I cannot reproduce the additional gap issue currently anymore.

> WM and application do not have memory of previous state

Firefox should be able to keep track of this, shouldn't it?

I've added the following to my config to handle this now.
It keeps track of the "maximized" properties (delayed, and skipped when in "fullscreen"), and then restores them after "fullscreen".

    -- Restore maximized state after fullscreen (Firefox).
    -- https://github.com/awesomeWM/awesome/issues/245
    local restore_properties = {}
    local restore_timer
[...]

PR #418, opened 2015-08-18
Do not unset max/fullscreen when setting the other

blueyed:
From http://standards.freedesktop.org/wm-spec/latest/ar01s05.html:

> _NET_WM_STATE_FULLSCREEN indicates that the window should fill the
> entire screen and have no window decorations. Additionally the Window
> Manager is responsible for restoring the original geometry after a
> switch from fullscreen back to normal window. For example, a
> presentation program would use this hint.

awesome prefers fullscreen internally already.  With this patch, the
previous maximized state will be restored after leaving fullscreen mode.

Fixes https://github.com/awesomeWM/awesome/issues/245.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1085,8 +1085,6 @@ client_set_fullscreen(lua_State *L, int cidx, bool s)
         /* become fullscreen! */
         if(s)
         {
-            /* remove any max state */
-            client_set_maximized(L, cidx, false);
             /* You can only be part of one of the special layers. */
             client_set_below(L, cidx, false);
             client_set_above(L, cidx, false);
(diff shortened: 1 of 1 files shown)


blueyed:
@psychon 
What is your opinion?
You've raised some concerns in https://github.com/awesomeWM/awesome/issues/245#issuecomment-118627267 - are they still valid?

psychon:
Well. If you really want to, then merge this, but e.g. `_NET_WM_STATE` would then contain multiple values and I don't know what kind of behavior is expected in this case (if two states conflict). On the other hand, I'm not sure if maximized and fullscreen are really meant to be conflicting in EWMH (the definition for maximized doesn't say what the window is maximized to).

I still stand by "this can cause confusing behavior if you toggle a window's state and nothing changes" (which is similar to the discussion about the number of masters that we have right now in some other issue), but you don't seem to have a problem with that.

blueyed:
I see your point.
OTOH it fixes a bigger annoyance than not being able to see maximized being switched when in fullscreen.

Issue #151, opened 2015-02-24
descriptions of keybindings


actionless:
What do you think about passing hotkey description to `awful.key()` as an optional argument (and when describing all of them in default rc.lua)?

It will be useful for 3rd party widgets (i seen at least 4 different ones) which shows list of current hotkeys. And at the same time it still will be readable from config (sort of replacement to commenting each hotkey).

Later it can be agreed on design for such widget and it can be implemented as part of awesome's lua libraries.

I think to have such cheatsheet will be very useful.

blueyed:
:+1:

blueyed:
`awful.key` might get changed to accept an optional table of arguments, where `desc` would be part of (to make it more flexible for future changes).  But then, adding `desc` as a new arg by itself would make it easier to use by default (no table/key needed).

actionless:
mb it can be like:

```
function key.new(mod, _key, press, release, ...)
    allkeys[mod][_key]['keydata'] = ...
```
```
    desc = foo.keydata[1]
```

Elv13:
Just as a side note, I also would like to have categories. I made a module for that, but after implementing it, the syntax isn't as nice looking as I thought it would be:

https://github.com/Elv13/awesome-configs/blob/master/shortcuts.lua#L136

The reason for this is to implement this as the welcome page:
!(Unity)[http://upload.wikimedia.org/wikipedia/commons/9/93/Ubuntu_Unity_Keyboard_shortcuts_-_En.png]

Here is how it look:
!(Awesome)[http://i.imgur.com/aMDMeii.png]

Many complains about tiled WM is that they are hard to learn (meh...) because they lack WYSIWYG hints. This kind of welcome screen help with that. Actionless edition with the visual keyboard layout is very useful too, maybe an hybrid of both could be used. The `Shorter` module used here have many issues, so I don't think it is the way forward versus upgrading awful.key. However, maybe we could break things a little more to make this use case better covered. What do you think?

actionless:
so the solution with `...` should help with categories, ie it can be like
```
awful.key({modkey}, 'a', do_action, 'Do some action', 'Misc stuff')


category = foo.keydata[2]
```

Elv13:
maybe more like 

```
awful.key({modkey}, 'a', do_action, {desc='Do some action', category='Misc stuff'})
```
?

actionless:
it's more straight-forward, i am agree.
[...]

PR #421, opened 2015-08-20
Add hotkeys' descriptions

actionless:
follow-up on #151 

how it looks (atm i added descriptions only for 7 hotkeys):
2015-08-20-214657_3840x2160_scrot

how it looks when vim is opened:
2015-08-20-214754_3840x2160_scrot

screen with more hotkeys (from my own config):
2015-08-20-220442_3840x2160_scrot

some work on appearance is still needed, like adding margins and so on, feel free to suggest

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -10,6 +10,7 @@ local beautiful = require("beautiful")
 -- Notification library
 local naughty = require("naughty")
 local menubar = require("menubar")
+local hotkeys_popup = require("awful.hotkeys_popup.widget")
 
 -- {{{ Error handling
 -- Check if awesome encountered an error during startup and fell back to
@@ -108,6 +109,7 @@ end
 -- {{{ Menu
 -- Create a laucher widget and a main menu
 myawesomemenu = {
+   { "hotkeys", function() return false, hotkeys_popup.show_help end},
    { "manual", terminal .. " -e man awesome" },
    { "edit config", editor_cmd .. " " .. awesome.conffile },
    { "restart", awesome.restart },
@@ -227,6 +229,8 @@ root.buttons(awful.util.table.join(
 
 -- {{{ Key bindings
 globalkeys = awful.util.table.join(
+    awful.key({ modkey,           }, "s",      hotkeys_popup.show_help,
+              {description="show help", group="awesome"}),
     awful.key({ modkey,           }, "Left",   awful.tag.viewprev,
               {description = "view previous", group = "tag"}),
     awful.key({ modkey,           }, "Right",  awful.tag.viewnext,
--- a/lib/awful/hotkeys_popup/init.lua
+++ b/lib/awful/hotkeys_popup/init.lua
@@ -1,0 +1,16 @@
+---------------------------------------------------------------------------
+--- Popup widget which shows current hotkeys and their descriptions.
+--
+-- @author Yauheni Kirylau &lt;yawghen@gmail.com&gt;
+-- @copyright 2014-2015 Yauheni Kirylau
+-- @release @AWESOME_VERSION@
+-- @module awful.hotkeys_popup
+---------------------------------------------------------------------------
+
+
+local hotkeys_popup = {
+  widget = require("awful.hotkeys_popup.widget"),
+  keys = require("awful.hotkeys_popup.keys")
+}
+hotkeys_popup.show_help = hotkeys_popup.widget.show_help
+return hotkeys_popup
--- a/lib/awful/hotkeys_popup/keys/init.lua
+++ b/lib/awful/hotkeys_popup/keys/init.lua
@@ -1,0 +1,14 @@
+---------------------------------------------------------------------------
+--- Additional hotkeys for awful.hotkeys_widget
+--
+-- @author Yauheni Kirylau &lt;yawghen@gmail.com&gt;
+-- @copyright 2014-2015 Yauheni Kirylau
+-- @release @AWESOME_VERSION@
+-- @module awful.hotkeys_popup.keys
+---------------------------------------------------------------------------
+
+
+local keys = {
+  vim = require("awful.hotkeys_popup.keys.vim")
+}
+return keys
--- a/lib/awful/hotkeys_popup/keys/vim.lua
+++ b/lib/awful/hotkeys_popup/keys/vim.lua
@@ -1,0 +1,132 @@
+---------------------------------------------------------------------------
+--- VIM hotkeys for awful.hotkeys_widget
+--
+-- @author Yauheni Kirylau &lt;yawghen@gmail.com&gt;
+-- @copyright 2014-2015 Yauheni Kirylau
+-- @release @AWESOME_VERSION@
+-- @module awful.hotkeys_popup.keys.vim
+---------------------------------------------------------------------------
+
+local hotkeys_popup = require("awful.hotkeys_popup.widget")
+
+local vim_rule = {name="vim"}
+for group_name, group_data in pairs({
+    vim_motion=                 { color="#009F00", rule=vim_rule },
+    vim_command=                { color="#aFaF00", rule=vim_rule },
+    vim_command_insert= { color="#cF4F40", rule=vim_rule },
+    vim_operator=             { color="#aF6F00", rule=vim_rule },
+}) do
+    hotkeys_popup.group_rules[group_name] = group_data
+end
+
+
+local vim_keys = {
+
+    vim_motion={{
+        modifiers = {},
+        keys = {
+            ['`']="goto mark",
+            ['0']='"hard" BOL',
+            ['-']="prev line",
+            w="next word",
+            e="end word",
+            t=". 'till",
+            ['[']=". misc",
+            [']']=". misc",
+            f=". find char",
+            [';']="repeat t/T/f/F",
+            ["'"]=". goto mk. BOL",
+            b="prev word",
+            n="next word",
+            [',']="reverse t/T/f/F",
+            ['/']=". find",
+            ['~']="toggle case",
+            ["#"]='prev indent',
+            ["$"]='EOL',
+            ["%"]='goto match bracket',
+            ["^"]='"soft" BOL',
+            ["*"]='next indent',
+            ["("]='begin sentence',
+            [")"]='end sentence',
+            ["_"]='"soft" BOL down',
+            ["+"]='next line',
+            W='next WORD',
+            E='end WORD',
+            T=". back 'till",
+            ['{']="begin parag.",
+            ['}']="end parag.",
+            F='. "back" find char',
+            G='EOF/goto line',
+            H='screen top',
+            L='screen bottom',
+            B='prev WORD',
+            N='prev (find)',
+            M='screen middle',
+            ['?']='. find(rev.)',
+        }
+    }},
+
+    vim_operator={{
+        modifiers = {},
+        keys = {
+            ['=']="auto format",
+            y="yank",
+            d="delete",
+            c="change",
+            ["!"]='external filter',
+            ['&lt;']='unindent',
+            ['&gt;']='indent',
+        }
+    }},
+
+    vim_command={{
+        modifiers = {},
+        keys = {
+            q=". record macro",
+            r=". replace char",
+            u="undo",
+            p="paste after",
+            g="gg: top of file, gf: open file here",
+            z="zt: cursor to top, zb: bottom, zz: center",
+            x="delete char",
+            v="visual mode",
+            m=". set mark",
+            ['.']="repeat command",
+            ["@"]='. play macro',
+            ["&amp;"]='repeat :s',
+            Q='ex mode',
+            Y='yank line',
+            U='undo line',
+            P='paste before',
+            D='delete to EOL',
+            J='join lines',
+            K='help',
+            [':']='ex cmd line',
+            ['"']='. register spec',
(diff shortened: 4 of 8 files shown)


lib/hot/keys/vim.lua
```diff
@@ -0,0 +1,137 @@
...
+      ["'"]=". goto mk. BOL",
+      b="prev word",
+      n="next word",
+      [',']="reverse t/T/f/F",
+      ['/']=". find",
+    }
+  },{
+    modifiers = {"Shift",},
+    keys = {
+      ['`']="toggle case",
```
blueyed: Is that specific to your keyboard layout?  (Vim uses `~`, which is `AltGr++` on a German layout)

blueyed:
Awesome!

I think that https://github.com/awesomeWM/awesome/pull/365 might help here.  It should allow to use the utf8 representation of keys in the bindings.

lib/hot/keys/vim.lua
actionless: i am using en_US

actionless:
seems like only the new name and pagination remaining?

blueyed:
What do you think about getting in #365 for this first?

actionless:
from what i understood after #365 will be merged i will just need to remove some predefined labels, but seems like no other changes need to be done in hotkey widget, or i am missing something?

lib/hot/widget.lua
```diff
@@ -0,0 +1,327 @@
...
+  keygrabber = keygrabber,
+}
+local awful = require("awful")
+local common = require("awful.widget.common")
+local wibox = require("wibox")
+local beautiful = require("beautiful")
+local dpi = beautiful.xresources.apply_dpi
+
+
+local markup = {}
```
actionless: btw, what do you think about copying whole `lain.util.markup`[1] into awesome?

[1] https://github.com/copycat-killer/lain/blob/master/util/markup.lua

actionless:
i hope now it should be more readable; do i need to do more visual adjustments?
2015-08-22-123919_3840x2160_scrot

actionless:
i wish someone will make a PR with `mpv` keys later on

actionless:
seems only the last issue remains: https://github.com/awesomeWM/awesome/pull/421#commitcomment-12829419

actionless:
i hope i am done now

actionless:
can you please merge it until it rot?

actionless:
rebased

blueyed:
I'll try this out later.
However, it has a lot of changed to the default config, which makes it hard to adopt/merge into a customized one.

actionless:
it shouldn't break any custom config

psychon:
I haven't tested this and I don't really get the vim-related module, but fine with me. Looking forward to breaking this via per-screen-DPI-textboxes and also looking forward to someone making this work with "non default" keyboard layouts.

blueyed:
> it shouldn't break any custom config

I see.  But you don't get the benefits without merging.

> I don't really get the vim-related module

Me neither.  It will only reflect the default Vim config, but seems to be merely used to provide an example, and therefore appears to be fine as-is.

Elv13:
May I suggest breaking this into 2 PR? One for the `awful.key` changes and one for the widget stuff. I think the `awful.key` changes are quite uncontroversial at this point and could be merged. I think, the remaining part may need some more work. For example, the key id being hardcoded for F keys wont work for all keyboard layouts and so on. I know this commit has been squashed and I am asking to undo some of that, but I think this PR is just too large not to scare off reviewer :P

actionless:
created https://github.com/awesomeWM/awesome/pull/491

actionless:
rebased on top of just-merged #491

Elv13:
The important thing is, Awesome 3.6 will have 22 more lines of useful code sure to be available and well tested by the time it is released ;) (this keybinding story has been going on for a year already, time to start merging some of it, right? :) )

actionless:
mb let's now merge only changes to key descriptions in rc.lua?

Elv13:
Sure

Should the first letter of each comment have a upper case or lower case?

actionless:
i dunno about the case

actionless:
rebased on top of https://github.com/awesomeWM/awesome/pull/519

actionless:
rebased on top of https://github.com/awesomeWM/awesome/pull/542

actionless:
with which application i should replace the vim example?

psychon:
If this should have an application example, I'm fine with vim. It's as good as everything else (and perhaps a bit better, because it's more keyboard-controlled than e.g. gedit).

blueyed:
I agree, Vim is fine.

Elv13:
After all, why not Vim

blueyed:
The default mapping to show the help does not work on a German keyboard layout, where "/" is Shift-7: it will switch to tag 7 instead.

That's probably an issue with the numeric handling of the tag mappings?!


> awful.key({ modkey,           }, "/",      hotkeys_popup.show_help,
>           {description="show help", group="awesome"}),

actionless:
do you have any idea what would be better to do then?

psychon:
Can we use `Mod+?`? Is awesome smart enough to figure out that this is `Mod+Shift+ß` on a German keyboard and `<something else involving Shift>` on an US layout?

actionless:
yup, it's ok i think, if anybody wouldn't like to press 2 modifiers for that than it's not a problem to rebind it later on in the config

actionless:
updated

blueyed:
I use `mod+s` for sticky, but rarely use it.
I think we should figure out how to use `mod+?`.

I also just got that error when trying to call `mod+?` (which is mod-shift-ß) on a de layout) via the following:

    awful.key({ modkey, 'Shift'           }, "ß",      hotkeys_popup.show_help,
                {description="show help", group="awesome"}),
```
stack traceback:
        /usr/share/awesome/lib/menubar/utils.lua:228: in function 'menubar.utils.compute_textbox_width'
        /usr/share/awesome/lib/awful/hotkeys_popup/widget.lua:251: in local 'insert_keys'
        /usr/share/awesome/lib/awful/hotkeys_popup/widget.lua:264: in upvalue 'create_wibox'
        /usr/share/awesome/lib/awful/hotkeys_popup/widget.lua:362: in function 'awful.hotkeys_popup.widget.show_help'
        /usr/share/awesome/lib/awful/key.lua:57: in function </usr/share/awesome/lib/awful/key.lua:57>
error: /usr/share/awesome/lib/menubar/utils.lua:228: attempt to call a nil value (method 'get_preferred_size')
```
Source: https://github.com/actionless/awesome/blob/6500708225f365f4b13492d1bee3d0e187d3e8c5/lib/awful/hotkeys_popup/widget.lua#L205, called from https://github.com/actionless/awesome/blob/6500708225f365f4b13492d1bee3d0e187d3e8c5/lib/awful/hotkeys_popup/widget.lua#L251-L252.

actionless:
but it works ok for you when it's binded to an other key, right?

i can't reproduce it though:
2015-11-21-224002_3840x2160_scrot

blueyed:
@actionless 
It prints:
```
<span font="Monospace Bold 9"><span foreground="#555555">Shift+Super+</span>ß </span><span font="Monospace 8">show help</span>
```

This apparently comes from:
```
    awful.key({ modkey, 'Shift'           }, "ß",      hotkeys_popup.show_help,
              {description="show help", group="awesome"}),
```

It's really odd though.  If I set it to "foo" it still fails.

```
max_label_content = "foo"
print("wibox", tostring(wibox.widget.textbox(max_label_content)))
```
results in `wibox	wibox.widget.textbox (table: 0x1874ae0)` though.

actionless:
weird, mb it depends on your current locale?

can you try now adding to your config:
```
max_label_content = '<span font="Monospace Bold 9"><span foreground="#555555">Shift+Super+</span>ß </span><span font="Monospace 8">show help</span>'
print("wibox", tostring(wibox.widget.textbox(max_label_content)))
```

blueyed:
I've tried it in awesome-client where it worked, also in the config.

Rebasing on master did not fix it either, but after adding some prints (and removing them again), it now does not happen anymore?!

actionless:
so, is there anything i can improve in this PR?

blueyed:
>  I think we should figure out how to use mod+?.

actionless:
you've told what it could be resolved by https://github.com/awesomeWM/awesome/pull/365

anyway i don't think what handling unicode symbols as keybindings is in scope of this PR and it is not blocking its work (and widget can even display those keybindings not looking to the fact what they are not actually work)

blueyed:
#365 is in since a while, but does not help in this regard.

Anyway, you're right that this should not really block it, and can be changed later anyway.

actionless:
yes, my point was what after that will be fixed, hotkey widget won't require any modifications to display them

blueyed:
Merged finally!
Thanks!

actionless:
thank you guys for reviewing that amount of code :)

blueyed:
@actionless 
Thanks for staying with us! :octocat: :)

psychon:
I just tried this with the default config plus the sky theme:
```
stack traceback:
	[C]: in function 'error'
	...chon/projects/awesome/build/lib/wibox/widget/textbox.lua:128: in function 'set_markup'
	...chon/projects/awesome/build/lib/wibox/widget/textbox.lua:238: in function <...chon/projects/awesome/build/lib/wibox/widget/textbox.lua:215>
	(...tail calls...)
	...rojects/awesome/build/lib/awful/hotkeys_popup/widget.lua:169: in function 'group_label'
	...rojects/awesome/build/lib/awful/hotkeys_popup/widget.lua:234: in function 'create_wibox'
	...rojects/awesome/build/lib/awful/hotkeys_popup/widget.lua:362: in function 'press'
	/home/psychon/projects/awesome/build/lib/awful/key.lua:57: in function </home/psychon/projects/awesome/build/lib/awful/key.lua:57>
error: ...chon/projects/awesome/build/lib/wibox/widget/textbox.lua:128: Value of 'background' attribute on <span> tag on line 1 could not be parsed; should be a color specification, not 'rgb:80/32/32'
```
No idea where `rgb:80/32/32` comes from. `grep -r rgb:` doesn't find anything.

psychon:
Ohhhh...
```
$ cat .Xdefaults 
[....]
!black
*color0:         rgb:19/19/19
*color8:         rgb:25/25/25
!red
*color1:         rgb:80/32/32
*color9:         rgb:98/2b/2b
!green
*color2:         rgb:5b/76/2f
*color10:        rgb:89/b8/3f
!brown/yellow
*color3:         rgb:aa/99/43
*color11:        rgb:ef/ef/60
!blue
*color4:         rgb:32/4c/80
*color12:        rgb:2b/4f/98
!magenta
*color5:         rgb:70/6c/9a
*color13:        rgb:82/6a/b1
!cyan
*color6:         rgb:92/b1/9e
*color14:        rgb:a1/cd/cd
!white
*color7:         rgb:ff/ff/ff
*color15:        rgb:dd/dd/dd
```
This is a bug in `beautiful.xresources`, right? (It should convert the stuff it reads into something that awesome can understand)

blueyed:
@actionless 
I've tried to use the hotkey_popup for a subset of hotkeys (after overwriting root keys temporarily), but the widget would display all hotkeys.. there is some caching involved, and the widget itself appears to be a singleton(?!) - there's no constructor exposed.
I would like to use a separate instance for the time when the modal mode is enabled.

I've also looked into invoking `show_help` with only a specific (list of) group(s).

Let me know if we should create a separate issue for this.

actionless:
@blueyed 

i can change it to the way what it will be returning popup instance when calling some `.create()` method but it will be returning sorta default instance of the widget when calling it the way how it's called now

are you ok with that idea? it should address your feature-request

___________________

btw, currently i have fork of `menubar` with the same feature implemented (being able to have multiple `menubar` instances instead of only singleton how is it now):
https://github.com/actionless/awesome_config/blob/devel/actionless/menubar/init.lua#L400-L412
but it's too much far from the upstream, do you want to have it upstream?

blueyed:
@actionless 
> i can change it to the way what it will be returning popup instance when calling some .create() method but it will be returning sorta default instance of the widget when calling it the way how it's called now

> are you ok with that idea? it should address your feature-request

Yes! :+1: 

Regarding the menubar, I've created some fix for the current version: https://github.com/awesomeWM/awesome/pull/711.  Having a better implementation by default would be nice, of course.

actionless:
> Regarding the menubar, I've created some fix for the current version: #711. Having a better implementation by default would be nice, of course.

so the point what it's missing at least 4-5 of the latest menubar-related PRs so it could take some time to rebase my menubar fork on upstream, the question is do anyone wants that or not

blueyed:
@actionless 
I want it in general, but have not looked into details.
I am using the menubar only from time to time.

actionless:
i just realized what your solution will be easier (to add an optional argument like `enabled_groups`), no reason for having multiple instances

or it still could be useful for cases like having multiple popups with different hotkeys at the same time?

blueyed:
@actionless 
> or it still could be useful for cases like having multiple popups with different hotkeys at the same time?

Not sure.
Separate instances seem to be better, and do not limit the different choices to available groups.

In my case I would like to have the global instance, and then one for a modal mode and/or some keygrabber (have not checked how useful it's in that context, given that you typically won't use `awful.key` there).

actionless:
created: https://github.com/awesomeWM/awesome/pull/772

PR #422, opened 2015-08-21
Add finish callback to awful.mouse.client.move

timroes:
This kind of callback might come in handy for user configurations and won't break any backward compatibility.

### DIFF
--- a/lib/awful/mouse/init.lua
+++ b/lib/awful/mouse/init.lua
@@ -155,7 +155,10 @@ end
 --- Move a client.
 -- @param c The client to move, or the focused one if nil.
 -- @param snap The pixel to snap clients.
-function mouse.client.move(c, snap)
+-- @param finished_cb An optional callback function, that will be called
+--                 when moving the client has been finished. The client
+--                 that has been moved will be passed to that function.
+function mouse.client.move(c, snap, finished_cb)
     local c = c or capi.client.focus
 
     if not c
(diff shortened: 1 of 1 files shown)


psychon:
:+1: (Although I'd leave out the `type(finished) == "function"` check to make sure things break if someone uses this wrongly. However, it doesn't hurt to have this either, so just leave it as-is.)

timroes:
I removed it since it definitely is a good idea not to silently accept broken configs :-)

lib/awful/mouse/init.lua
```diff
@@ -155,7 +155,10 @@ end
 --- Move a client.
 -- @param c The client to move, or the focused one if nil.
 -- @param snap The pixel to snap clients.
-function mouse.client.move(c, snap)
+-- @param finished An optional callback function, that will be called
+--                 when moving the client has been finished. The client
+--                 that has been moved will be passed to that function.
+function mouse.client.move(c, snap, finished)
```
blueyed: I suggest using a more distinct name, like cb_finished, on_finish or finished_cb.

timroes: Any preferences on which of those three? (I would use `on_finish`, but will change it to whatever you decide.)

blueyed: `_cb` is used by naughty and gears.cache already, so I would go with `finished_cb`.

While at it, then please also squash your commits into one, and then I think it's good to be merged.

Thanks!

PR #430, opened 2015-08-29
Set an upper limit on nmaster/ncol based on number of tiled windows

alastair-robertson:
(with the default key-mappings) 

Pressing Mod+Shift+h repeatedly keeps on increasing the value of nmaster indefinitely, past the number of windows visible on the screen. This means it can take a lot of presses of Mod+Shift+l to get it back to a value where you can actually see your layout changing.

With this patch, pressing Mod+Shift+l will always reduce the number of visible master windows by 1.

Same again for ncol.

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -26,6 +26,9 @@ local capi =
     root = root
 }
 
+-- we use require("awful.client") inside functions to prevent circular dependencies.
+local client
+
 local tag = { mt = {} }
 
 -- Private data
@@ -335,9 +338,10 @@ function tag.setscreen(t, s)
 end
 
 --- Get a tag's screen
--- @param t tag object
+-- @param[opt] t tag object
 -- @return Screen number
 function tag.getscreen(t)
+    local t = t or tag.selected()
     return tag.getproperty(t, "screen")
 end
 
@@ -460,9 +464,27 @@ end
 
 --- Increase the number of master windows.
 -- @param add Value to add to number of master windows.
--- @param t The tag to modify, if null tag.selected() is used.
-function tag.incnmaster(add, t)
-    tag.setnmaster(tag.getnmaster(t) + add, t)
+-- @param[opt] t The tag to modify, if null tag.selected() is used.
+-- @tparam[opt=false] boolean sensible Limit nmaster based on the number of
+-- visible tiled windows?
+function tag.incnmaster(add, t, sensible)
+    if sensible then
+        client = client or require("awful.client")
+        local ntiled = #client.tiled(tag.getscreen(t))
+
+        local nmaster = tag.getnmaster(t)
+        if nmaster > ntiled then
+            nmaster = ntiled
+        end
+
+        local newnmaster = nmaster + add
+        if newnmaster > ntiled then
+            newnmaster = ntiled
+        end
+        tag.setnmaster(newnmaster, t)
+    else
+        tag.setnmaster(tag.getnmaster(t) + add, t)
+    end
 end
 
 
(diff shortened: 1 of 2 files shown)


psychon:
These circular dependencies are beginning to be annoying.

WIth this patch it is possible that `awful.tag.incmaster` actually decreases the number of master windows. Also, there might be valid use cases for tags with a bigger number of master windows than tiled windows, e.g. when "preparing for something later". Thus I would definitely want to make this optional.

I have two ideas for this: Either a new optional argument to these functions (but too many optional arguments makes it harder to understand their meaning) or a new function `awful.tag.sensiblenmaster()` which decreases `nmaster` if it is too large (this new function would then be called in the default config right after calling `incmaster` / `incncol`).

Any other ideas? Anyone?

blueyed:
:+1: for a new function to be used in the default config.

alastair-robertson:
Thanks for the responses!

I might be misunderstanding something here, but I think we'd have to call sensiblenmaster() both before and after incmaster() in order for this to work as expected when the number of tiled windows decreases or increases. How much of a problem do you think this is?

blueyed:
You are right.  And I misunderstood the idea: I've thought it was a new function that would behave like `incnmaster` etc, but with bounds (like using an optional arg).
I would now vote for the new arg: it makes it more concise, and would not cause the value to be changed twice, in case it hits the boundary.

actionless:
+ for opt arg

psychon:
Hm, right, the default config would have to call that function twice. That wouldn't be nice, I agree.

lib/awful/tag.lua
```diff
@@ -471,7 +486,21 @@ end
 -- @param add Value to add to number of column windows.
 -- @param t The tag to modify, if null tag.selected() is used.
 function tag.incncol(add, t)
-    tag.setncol(tag.getncol(t) + add, t)
+    client = client or require("awful.client")
+    local ntiled = #client.tiled(1)
```
blueyed: That should be the tag's screen probably (assuming that `1` is the screen here).

lib/awful/tag.lua
```diff
@@ -431,7 +434,19 @@ end
 -- @param add Value to add to number of master windows.
 -- @param t The tag to modify, if null tag.selected() is used.
 function tag.incnmaster(add, t)
-    tag.setnmaster(tag.getnmaster(t) + add, t)
+    client = client or require("awful.client")
+    local ntiled = #client.tiled(1)
```
blueyed: That should be the tag's screen probably (assuming that `1` is the screen here).

blueyed:
@alastair-robertson 
Can you make it an optional behaviour, please?

blueyed:
Also I agree with @psychon  that it's odd when `incmaster` actually would decrease the number.

lib/awful/tag.lua
alastair-robertson: Oh wow - yeah wasn't meant to leave that, sorry

alastair-robertson:
Yep, no problem

lib/awful/tag.lua
```diff
@@ -429,9 +433,23 @@ end
...
 -- @param add Value to add to number of master windows.
--- @param t The tag to modify, if null tag.selected() is used.
-function tag.incnmaster(add, t)
-    tag.setnmaster(tag.getnmaster(t) + add, t)
+-- @param[opt] t The tag to modify, if null tag.selected() is used.
+-- @tparam[opt=false] boolean sensible Limit nmaster based on the number of
+-- visible tiled windows?
+function tag.incnmaster(add, t, sensible)
+    client = client or require("awful.client")
+    local ntiled = #client.tiled(tag.getscreen(t))
```
blueyed: You only need to get/set `ntiled` for `sensible == true`.

lib/awful/tag.lua
```diff
@@ -429,9 +433,23 @@ end
...
+
+    local nmaster = tag.getnmaster(t)
```
blueyed: Same as above - only for `sensible`.

blueyed:
Looks good already, but please optimize it for the (default) `sensible == false` case.

blueyed:
@alastair-robertson 
Do you feel like adjusting the PR, or should we take over?
(I don't want to be too picky, but it'll be good to get it as good as possible before merging)
Thanks!

alastair-robertson:
Yeah I don't mind doing it at all, I just don't have much time in the evenings - I'll try to get it done tomorrow. Also, should I squash all the commits into one once this is done, or leave them separate?

blueyed:
Cool!
Yes, please squash them.

blueyed:
Thanks a lot, merged!

PR #431, opened 2015-08-30
Fix awful.rules.does_match

blueyed:
This includes a fix (https://github.com/awesomeWM/awesome/commit/eb01e969f68ca577224bdf8693c5d13c5211be04) and some refactoring / improvement.

I am not sure about the new method name `entry_matches`.
There is `does_match` already for a list of rules, although it has never been working apparently (added in adbc7ac).

Maybe we could use `matches` and `list_matches`?

### DIFF
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -130,6 +130,16 @@ function rules.match_any(c, rule)
     return false
 end
 
+--- Does a given rule entry match a client?
+-- @client c The client.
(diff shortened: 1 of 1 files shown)


psychon:
Urgh. Fine with me, but when did `awful.rules` become so complicated?

blueyed:
What is fine with you?  The supposed change?

> Maybe we could use matches and list_matches?

psychon:
Yeah

blueyed:
Done, moving to "next".

PR #433, opened 2015-09-02
keyboardlayout: Extend allowable 'group_names' pattern range

nuko8:
Extend the range of the 'group_names' patterns which is currently
restricted to typical ones like

    pc+us+ru:2+de:3+ba:4+inet

to more general ones such as

    macintosh_vndr/apple(alukbd)+macintosh_vndr/jp(usmac)+macintosh_vndr/jp(mac):2

so that the keyboardlayout widget can handles all possible patterns
returned by awesome.xkb_get_layout_group().

### DIFF
--- a/lib/awful/widget/keyboardlayout.lua
+++ b/lib/awful/widget/keyboardlayout.lua
@@ -17,58 +17,261 @@ local widget_base = require("wibox.widget.base")
 -- awful.widget.keyboardlayout
 local keyboardlayout = { mt = {} }
 
+-- As to the country-code-like symbols below, refer to the names of XKB's
+-- data files in /.../xkb/symbols/*.
+keyboardlayout.xkeyboard_country_code = {
+   ["ad"] = true,    -- Andorra
+   ["af"] = true,    -- Afganistan
+   ["al"] = true,    -- Albania
+   ["am"] = true,    -- Armenia
+   ["ara"] = true,   -- Arabic
+   ["at"] = true,    -- Austria
+   ["az"] = true,    -- Azerbaijan
+   ["ba"] = true,    -- Bosnia and Herzegovina
+   ["bd"] = true,    -- Bangladesh
+   ["be"] = true,    -- Belgium
+   ["bg"] = true,    -- Bulgaria
+   ["br"] = true,    -- Brazil
+   ["bt"] = true,    -- Bhutan
+   ["bw"] = true,    -- Botswana
+   ["by"] = true,    -- Belarus
+   ["ca"] = true,    -- Canada
+   ["cd"] = true,    -- Congo
+   ["ch"] = true,    -- Switzerland
+   ["cm"] = true,    -- Cameroon
+   ["cn"] = true,    -- China
+   ["cz"] = true,    -- Czechia
+   ["de"] = true,    -- Germany
+   ["dk"] = true,    -- Denmark
+   ["ee"] = true,    -- Estonia
+   ["epo"] = true,   -- Esperanto
+   ["es"] = true,    -- Spain
+   ["et"] = true,    -- Ethiopia
+   ["fi"] = true,    -- Finland
+   ["fo"] = true,    -- Faroe Islands
+   ["fr"] = true,    -- France
+   ["gb"] = true,    -- United Kingdom
+   ["ge"] = true,    -- Georgia
+   ["gh"] = true,    -- Ghana
+   ["gn"] = true,    -- Guinea
+   ["gr"] = true,    -- Greece
+   ["hr"] = true,    -- Croatia
+   ["hu"] = true,    -- Hungary
+   ["ie"] = true,    -- Ireland
+   ["il"] = true,    -- Israel
+   ["in"] = true,    -- India
+   ["iq"] = true,    -- Iraq
+   ["ir"] = true,    -- Iran
+   ["is"] = true,    -- Iceland
+   ["it"] = true,    -- Italy
+   ["jp"] = true,    -- Japan
+   ["ke"] = true,    -- Kenya
+   ["kg"] = true,    -- Kyrgyzstan
+   ["kh"] = true,    -- Cambodia
+   ["kr"] = true,    -- Korea
+   ["kz"] = true,    -- Kazakhstan
+   ["la"] = true,    -- Laos
+   ["latam"] = true, -- Latin America
+   ["latin"] = true, -- Latin
+   ["lk"] = true,    -- Sri Lanka
+   ["lt"] = true,    -- Lithuania
+   ["lv"] = true,    -- Latvia
+   ["ma"] = true,    -- Morocco
+   ["mao"] = true,   -- Maori
+   ["me"] = true,    -- Montenegro
+   ["mk"] = true,    -- Macedonia
+   ["ml"] = true,    -- Mali
+   ["mm"] = true,    -- Myanmar
+   ["mn"] = true,    -- Mongolia
+   ["mt"] = true,    -- Malta
+   ["mv"] = true,    -- Maldives
+   ["ng"] = true,    -- Nigeria
+   ["nl"] = true,    -- Netherlands
+   ["no"] = true,    -- Norway
+   ["np"] = true,    -- Nepal
+   ["ph"] = true,    -- Philippines
+   ["pk"] = true,    -- Pakistan
+   ["pl"] = true,    -- Poland
+   ["pt"] = true,    -- Portugal
+   ["ro"] = true,    -- Romania
+   ["rs"] = true,    -- Serbia
+   ["ru"] = true,    -- Russia
+   ["se"] = true,    -- Sweden
+   ["si"] = true,    -- Slovenia
+   ["sk"] = true,    -- Slovakia
+   ["sn"] = true,    -- Senegal
+   ["sy"] = true,    -- Syria
+   ["th"] = true,    -- Thailand
+   ["tj"] = true,    -- Tajikistan
+   ["tm"] = true,    -- Turkmenistan
+   ["tr"] = true,    -- Turkey
+   ["tw"] = true,    -- Taiwan
+   ["tz"] = true,    -- Tanzania
+   ["ua"] = true,    -- Ukraine
+   ["us"] = true,    -- USA
+   ["uz"] = true,    -- Uzbekistan
+   ["vn"] = true,    -- Vietnam
+   ["za"] = true,    -- South Africa
+}
+
 -- Callback for updaing current layout
-local function update_status (keyboardlayout)
-   keyboardlayout._current = awesome.xkb_get_layout_group();
+local function update_status (self)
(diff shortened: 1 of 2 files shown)


psychon:
So far I found http://cgit.freedesktop.org/xkeyboard-config/tree/symbols which seems to be where the list of languages comes from.

After reading half of https://wiki.archlinux.org/index.php/X_KeyBoard_extension, I'm scared. At least I now know what this `vendor/file(section):group_idx` syntax actually means and things seem to be complicated.

I'm not sure I understand the stuff about "linguistic".
Let's take your example:

        macintosh_vndr/apple(alukbd)+macintosh_vndr/jp(usmac)+macintosh_vndr/jp(mac):2

You want to ignore the "apple" part and instead get `jp(usmac)` for group 1 and `jp(mac)` for group 2, right?

nuko8:
Yes, that's right.

`jp(usmac)` is for ASCII characters, and `jp(mac)` is for Japanese phonic characters known as kana.

nuko8:
@psychon wrote:

> Do we really need this hardcoded list instead of accepting everything which "looks correct"? (this is an honest question, I don't know how these layout patterns work)
>
> If we really have such a list, who makes sure it stays up to date? Can we simplify updating it somewhow?

Roughly speaking, the file in the `symbols` directory of `xkb`'s configuration files are categorized into two classes: One is those related to keys which has nothing to do with characters (I mean, keys such as shift, lock, control, mod1—mod5, enter/return, escape, tab, F* and so on) and the other is those related to keys which has something to do with characters.

I call the latter *linguistic* conventionally.  (The former might be called *functional*.)

As you pointed out, there may be a heuristic way to tell one from another, but it’s only heuristic and, once we rely on it, we will have to be worried about exceptions which can be brought by other parties we have no control over.

So, I decided to use the country code like symbols, knowing its consequences: It make the code harder to maintain and uncertain in updating.

But they are definite in comparison with something heuristic and the number of them isn’t over 300 or so, which can be handled manually, if we will.

Thus, even with such hard-coded symbols,  we simply add new symbols to `layout_name` when we find the contents of `symbols` directory is updated,  Perhaps, it is easier to find a “logic” or “reasoning” behind a heuristic we may rely on.

That’s why I’m using the hard-coded `layout_names`, still seeking for a better way.  So I'm quite open to other ways and longing for them, if any.   Any comment on it is definitely welcome.

As to the rest of your comments, I think I revised the code in such a way you're satisfied with it.   I'll push it in a day after testing.

nuko8:
I’ve just pushed the revised version.

There have been 9 line notes on 45d78ae.  I’m going to write what I did for each.

> I don't think this was meant to happen :-)

Neither do I :-)  Fixed.

> I'd suggest leaving the check for string out. This should only ever see nil or a string, everything else should be allowed to cause errors.

Done.

> Do we really need this hardcoded list instead of accepting everything which "looks correct"? (this is an honest question, I don't know how these layout patterns work)
>
> If we really have such a list, who makes sure it stays up to date? Can we simplify updating it somewhow?

Pending.  I’m still looking for a better way…

Or, should we use an exclusion check instead of the current inclusion check?  I mean, we use `pc`, `capslock`, `shift` and so on instead of the country code like symbols.

That makes the hard-coded list shorter and thus future update burden lighter, though they still remain as a cause of headache.

> All of the above can be shortened: 
>
> (…snip…)
>
> (However, I am not 100% sure that this really does the same thing as your code. For example, "a++b" will be handled differently, I think (your code produces an empty match, my version ignores it)).

Done.  As to the “a++b” case (an empty term), we can safely ignore it.

> Perhaps anchors ^ and $ should be added around these patterns to enforce matching the whole string

Done.  That greatly helped to make the code shorter and neater.

> Hm. Instead of these magic numbers for the indices into the patterns table, how about moving this as a callback function into the table?
>
> (…snip…)
>(Note that lua wouldn't guarantee the order of entries any more, I guess that needs to be handled somehow differently...)

Done.  As to your concern about the entry order, owing to your previous comment on whole string matching, it seemingly disappears.

> Could you add a comment explaining this linguistic part? Perhaps some general explanation of these layout strings (or a link to such an explanation)?

I changed the variable names from `layout_names` and `is_linguistic` to `xkeyboard_country_code` and `is_country_specific` to clarify my intension and their relation with XKB.

> print?

Fixed.  That’s part of the remnants of debugging.  Sorry about that.

> This code seems complicated enough that it might make sense to add unit tests for this parsing, I guess. This would mean that the function get_groups_from_group_names needs to be public and a corresponding test would be placed in spec/awful/widgets/keyboardlayout_spec.lua.

For that, the function was made public and the header comment was changed accordingly.

So, the only outstanding issue is the hard-coded country code array for now…

Of course, additional comments are also welcome.

nuko8:
In addition to the last two commits, should I replace the last statement of the body of `keyboard layout.new()`, i.e., `return self` with `return setmetatable(self, keyboardlayout)` so that the function `keyboardlayout.get_groups_from_group_names()` can be used with the singleton returned from the `new` function, and change the signature of `keyboardlayout.get_groups_from_group_names()` from `(group_names)` to `(self, group_names)` accordingly?

psychon:
Nice, thanks a lot! I don't mind the list of countries anymore either. Plus, if that is moved to `keyboardlayout.IForgotTheNameOfTheVariable` we can count it as a setting with a complicated default instead of a hardcoded list. ;-)

I don't see much value in making `get_groups_from_group_names` an "instance method" instead of it being "static" like it is now. It doesn't do anything instance-specific and the only callers should be this code and the test (that I still have to write, sorry!).

psychon:
Here is a quick go for a unit test: https://gist.github.com/psychon/16920283f0b8253eddfd
Save this as `spec/awful/widget/keyboardlayout_spec.lua` and use `make check-unit` to run the test.

I guess this should be added to the PR and feel free to take over the copyright for that file (in case you add more tests to it, the tests are the hard part, the "general structure" is easy).

Note that this test currently fails: The `group_idx` values that are returned are strings, not numbers (except for group 1, that one is handled correctly).

Also, I'm not really sure what good test cases would be and what complicated stuff is possible. Does this already cover enough cases? Perhaps something like just `de` should be tested as well. And while we are at it, what happens with `de+jp` (besides this being unlikely to work with XKB)?

nuko8:
I’ve done another push to address your latest comments:

> This code would be simpler if the `xkeyboard_country_code` table used the countries as key instead of values.
> 
> Honest question (I'm not sure!): Does it make sense to change local `xkeyboard_country_code = { "de", "fr", ... }` to local `xkeyboard_country_code = { de = true, fr = true, ... }` so that this code here just becomes `if not xkeyboard_country_code[file] then break end`?

Yes, it does make sense.  That `{ key = true }` construct as well as that of `{ pattern = callback}` is very impressed on me.  It seems what I’ve learnt from you is more than my possible contribution ;-)

I rewrote the code, following your suggestion.

> And while I am thinking out loud:
> Should the local table `xkeyboard_country_code` be moved to a "global" table as `keyboardlayout.xkeyboard_country_code` so that people can add/remove entries if they really want to?

Done.

I’m sorry to have kept you saying the same thing again and again.  Now I understand that is the second best measure which I should take only after looking into any other means.

Last but not least. Thank you for making the unit test.  I need a day to get familiar with it.  So I beg your patience.

nuko8:
I’ve just pushed two commits.

@psychon wrote:

> Here is a quick go for a unit test: https://gist.github.com/psychon/16920283f0b8253eddfd
Save this as spec/awful/widget/keyboardlayout_spec.lua and use make check-unit to run the test.

Let me say thank you again!

> I guess this should be added to the PR and feel free to take over the copyright for that file (in case you add more tests to it, the tests are the hard part, the "general structure" is easy).

I added my name as coauthor and pushed a revised version of `keyboardlayout_spec.lua` as part of the PR.  

Without your help, I couldn’t complete the unit test in such a short period of time.  In fact, I haven’t had either `luarocks` or `busted` on my Mac until this morning.  So, I’d like to keep your name there.  I wish you wouldn’t mind it.

> Note that this test currently fails: The group_idx values that are returned are strings, not numbers (except for group 1, that one is handled correctly).

That’s my fault.  I should have explicitly converted `group_idx` into a value of   the number type, not relying on Lua’s auto-conversion.   I fixed it and pushed a revised version, which should pass the original part of the test as well as its newly added part.

> Also, I'm not really sure what good test cases would be and what complicated stuff is possible.

I know that’s what I must look into ;-)

> Does this already cover enough cases? Perhaps something like just `de` should be tested as well. 

I tried having the test include all cases as possible.  I think logically interesting cases are all included.  Also, I added some practically interesting cases I found here and there.

> And while we are at it, what happens with `de+jp` (besides this being unlikely to work with XKB)?

The function returns

```
{
  { file = “de”, group_idx = 1 },
  { file = “jp”, group_idx = 1 }
}
```

as expected from the new implementation.  As you pointed out, obviously indefinite settings like `de+jp` is silently ignored by XKB.  So we won’t get something like `de+jp` from XKB.

Should `keyboardlayout` receive such groups for some reason, since `group_idx == 1`, every entry would be put overly on 0th position of the table, which means that `keyboardlayout` widget does nothing with 0th position entry because it works only with entries where `index >= 1`.

So, we are always on a safer side in that particular case.

That said, I added such a case to the test for checking the robustness of the implementation.

Now, as you suggested, the PR has become lengthy.  So I also think it’s a good idea to squash the existing commits but not sure of the way.  After the review is all done, should I replace the current commits with a single or two new commits via `push -f`, or is there another way you all prefer to?

psychon:
The current "head" of this PR is commit 859bb89d3a22d2cacf31e1cfd842e33693e8d282. Via this we can always find it again later if needed (although I doubt it).

I think this can be merged. Feel free to rewrite the history if you want, but I volunteer to do that myself (I just can't make any promise when I find the time).

Thanks, it's been a pleasure to work with you. Also thanks for these test cases (I'm ashamed I forgot to test the empty string :-) ). Sorry that I forgot that `busted` is a "special" requirement that isn't satisfied out of the box everywhere.

nuko8:
> The current "head" of this PR is commit 859bb89. Via this we can always find it again later if needed (although I doubt it).
>
> I think this can be merged. Feel free to rewrite the history if you want, but I volunteer to do that myself (I just can't make any promise when I find the time).

I see.  It looks better for now that I leave the existing commits as they are.  If you find that some work on my side is convenient for you, just let me know any time.

> Thanks, it's been a pleasure to work with you. 

That's my pleasure, too.  That pleasure made me study `busted` as a matter of course.

> Sorry that I forgot that busted is a "special" requirement that isn't satisfied out of the box everywhere.

No problem at all.  Actually, that made me more confident of what I've done with you.  

Thanks a lot.

blueyed:
I've only skimmed this, but it appears to be good to be merged?!
Moving to the "next" milestone.

nuko8:
The existing 9 commits has been squashed into 3 commits, and the PR has been rebased on the latest master 4753580.

PR #437, opened 2015-09-09
placement.no_offscreen: add optional screen argument

blueyed:
This is used by `tooltip.place` then to tie the tooltip to the screen of
the mouse.  Without this, a tooltip from the tasklist might get moved to
the screen above the tasklist, if it gets considered to be on that
screen given its coordinates.


psychon:
:+1:

PR #438, opened 2015-09-09
ewmh.activate: raise a client always

blueyed:
When a client is not visible, this would adjust its stacking order still.

This is a proposal from IRC discussion with @b80905, where a `request::activate` signal from `movetotag` would not raise the client if it's not visible anymore.

This is not tested, but just a quick idea.


blueyed:
I wonder if sending the signal before moving the client to another tag would work? probably?!

blueyed:
Ref: https://github.com/awesomeWM/awesome/pull/439 (regarding the previous comment)

PR #439, opened 2015-09-09
Make client.movetotag() raise the client

b80905:
if you move two windows from one tag to another one and switch to the other tag, it might be that the focus is not on the topmost window, which looks odd if the windows are maximized. This is a fix.


blueyed:
This should use the same mechanism as in fbc72624d7cd8bc7a96d4a261e0f281be5614375 to only do this when the selected client is focused currently - you might want to move non-focused clients etc, and in this case they should not be raised (probably?!).

b80905:
OK. I've fixed that.

PR #441, opened 2015-09-10
add missing hints argument in `awful.movetoscreen`

aroig:
Otherwise the signal handler complains about the nil value.


blueyed:
:+1: 

Additionally: https://github.com/awesomeWM/awesome/pull/444.

PR #445, opened 2015-09-10
Fix `set_visible` and `set_opacity` for widgets: s/self/ret/g

blueyed:


psychon:
I haven't tested anything, but what exactly is broken? Shouldn't `function ret:set_visible(b)` get interpreted as `function ret.set_visible(self, b)` and that's where the `self` comes from?

blueyed:
Yeah, I'm not sure about this anymore - I have seen some error in this regard IIRC, and had this patch laying around since a while.
Would have need to investigate I think.

blueyed:
Closing this for now.

PR #448, opened 2015-09-13
fix #420 (clients menu does not reopen after closing it with ESC)

### LINKED ISSUE #420: [Default rc.lua] Right-click on tasklist (brining up `awful.menu.clients`) is buggy
How to reproduce:

1. Open a few clients;
2. Right-click on the tasklist to bring up `awful.menu.clients`;
3. Click on one of the entries to switch to that client;
4. Try to right-click again on the taskbar to bring up another `awful.menu.clients` -- it won't work.

This happens because of the [`instance`](https://github.com/awesomeWM/awesome/blob/master/awesomerc.lua#L153) var in rc.lua. When an entry in the menu is clicked, `menu:exec()` calls `menu:hide()` internally, however, `instance` remains non-nil, causing it to fail the if-condition the next time there is a right-click.

everslick:


blueyed:
Thanks!

blueyed:
Holding this back, until https://github.com/awesomeWM/awesome/issues/420#issuecomment-139857493 gets incorporated.

everslick:
of course it's #420 :-/

blueyed:
> of course it's #420 :-/

No problem, I will amend this into a single commit.

psychon:
You don't need that helper function just to create a new scope:

```lua
local client_menu_toggle
do
    local client_menu_instance
    client_menu_toggle = function()
        if and-all-that-other-stuff
    end
end
```

What do others think? Would it be easier to read this way? If it were ok for `client_menu_toggle` to be a global function, the `local client_menu_toggle` before this block wouldn't be needed.

blueyed:
I like the `do / end` approach.

@everslick 
What do you think?

everslick:
I think it is more confusing to Lua novices (like me!) ;-). On the other
hand it can be rather educational to have this construct in rc.lua.

On a personal level, I think everything should be local. In this regard the
do/end looks strange to me.

All in all, I have no strong opinion either way. :) I'd rather leave the
decision to you, guys.

blueyed:
Agreed.
@everslick 
Do you want to create another update of the PR then? :)

everslick:
I'd rather not. I'll mess it up somehow! ;-)

psychon:
Just a quick comment:
If everything in rc.lua is local, you can't access it via e.g. `awesome-client`. Things one might to access are the `tags` table, just to name one thing. Dunno if there are any cases for which we have no work around (tags table can just be fixed by going through awful to get tags).

everslick:
Thx, for the hint. It didn't occur to me as useful to have everything
defined globally. At the end I think these questions are less of an issue,
because i guess serious Awesome users will customize rc.lua to a very high
degree anyway. They can decide for themselves.

PR #451, opened 2015-09-15
Add a benchmark to the test suite

psychon:
This (ab)uses the integration tests to run a benchmark. This currently only
measures wibox drawing performance. To avoid wasting CPU-cycles, this does only
a quick run under travis while on "normal" runs the function under test is
executed in a loop to improve the precision of the measurement.

This benchmarks hopefully allow to optimize things in a clear fashion instead of
things like "it feels faster to me".

Results when run against the previous commit:

== Running test-benchmark.lua ==
        create wibox: 0.0788958  sec/iter ( 13 iters, 1.103 sec for benchmark)
    update textclock: 0.019493   sec/iter ( 56 iters, 2.507 sec for benchmark)
  relayout textclock: 0.0160725  sec/iter ( 63 iters, 1.029 sec for benchmark)
    redraw textclock: 0.0015601  sec/iter (647 iters, 1.875 sec for benchmark)
W: awesome: a_glib_poll:291: Last main loop iteration took 6.593912 seconds! Increasing limit for this warning to that value.

Results right before the new widget layouts were merged (commit 52154d0f15b99b):

== Running test-benchmark.lua ==
        create wibox: 0.0782874  sec/iter ( 13 iters, 1.095 sec for benchmark)
    update textclock: 0.00736755 sec/iter (136 iters, 1.346 sec for benchmark)
[...]
error: /home/psychon/projects/awesome/build/lib/gears/object.lua:30: Trying to emit non-existent signal 'widget::layout_changed'

### DIFF
--- a/tests/test-benchmark.lua
+++ b/tests/test-benchmark.lua
@@ -1,0 +1,95 @@
+-- Some benchmarks that aren't really tests, but are included here anyway so
+-- that we notice if they break.
+
+local awful = require("awful")
+local wibox = require("wibox")
+local lgi = require("lgi")
+local GLib = lgi.GLib
+local cairo = lgi.cairo
+
+local not_under_travis = not os.getenv("CI")
+
+local measure, benchmark
+do
+    local timer_measure = GLib.Timer()
+    measure = function(f, iter)
+        timer_measure:start()
+        for i = 1, iter do
+            f()
+        end
+        local elapsed = timer_measure:elapsed()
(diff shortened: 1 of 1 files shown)


blueyed:
Nice!

I first thought that it might be useful to have this in Travis in the normal way, which would show how performance changes, but then it is not relevant for most commits and could be plotted manually better anyway - we would have to send/store the metrics generated on Travis otherwise.

PR #452, opened 2015-09-15
Speed up widget hierarchy updates

psychon:
All the important stuff can be found in this commit message: https://github.com/awesomeWM/awesome/commit/716333d5d78b9e00b4765218ead6f5d15cc2020f

### DIFF
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -75,19 +75,25 @@ local function do_redraw(self)
     -- Relayout
     if self._need_relayout or self._need_complete_repaint then
         self._need_relayout = false
-        local old_hierarchy = self._widget_hierarchy
-        self._widget_hierarchy_callback_arg = {}
-        self._widget_hierarchy = self.widget and
-            hierarchy.new(get_widget_context(self), self.widget, width, height,
-                self._redraw_callback, self._layout_callback, self._widget_hierarchy_callback_arg)
+        if self._widget_hierarchy and self.widget then
+            self._widget_hierarchy:update(get_widget_context(self),
+                self.widget, width, height, self._dirty_area)
+        else
+            self._need_complete_repaint = true
+            if self.widget then
+                self._widget_hierarchy_callback_arg = {}
+                self._widget_hierarchy = hierarchy.new(get_widget_context(self), self.widget, width, height,
+                        self._redraw_callback, self._layout_callback, self._widget_hierarchy_callback_arg)
+            else
+                self._widget_hierarchy = nil
+            end
+        end
 
-        if old_hierarchy == nil or self._widget_hierarchy == nil or self._need_complete_repaint then
+        if self._need_complete_repaint then
             self._need_complete_repaint = false
             self._dirty_area:union_rectangle(cairo.RectangleInt{
                 x = 0, y = 0, width = width, height = height
             })
-        else
-            self._dirty_area:union(self._widget_hierarchy:find_differences(old_hierarchy))
         end
     end
 
--- a/lib/wibox/hierarchy.lua
+++ b/lib/wibox/hierarchy.lua
@@ -88,8 +88,8 @@ function hierarchy_update(self, context, widget, width, height, region, matrix_t
     -- Update children
     local old_children = self._children
     self._children = {}
-    for idx, w in ipairs(layout_result or {}) do
-        local r = old_children[idx]
+    for _, w in ipairs(layout_result or {}) do
+        local r = table.remove(old_children, 1)
         if not r then
             r = hierarchy_new(w._widget, self._redraw_callback, self._layout_callback, self._callback_arg)
         end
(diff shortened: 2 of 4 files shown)


blueyed:
Awesome!
No problems on my end so far - feel free to merge it.

blueyed:
With this, closing the last client on a tag will not update the tasklist properly: the client is still being displayed there (as highlighted).

psychon:
Whoops. The code handled the case of "I was moved 10px to the left", but it did not handle the cases of "there are now less children than before" nor "a different widget should now be shown here". I added two commits for handling these and hopefully this now handles all cases.

Looking at the old code for `hierarchy:find_differences()` (which this PR removes), this new code should now handle the same cases (but gets away with redrawing only a smaller area in some cases).

I guess we should test this some more before merging.

blueyed:
Since 716333d5d78b9e00b4765218ead6f5d15cc2020f (and with the latest commits, too), my cyclefocus plugin gets not updated anymore.

I have started rewriting it to use a wibox etc instead a while ago (no code online for this currently).

I am using `layout:reset()` in case of the second invocation, and that might not prune the caches maybe?

        if not wbox then
            local wa = screen[mouse.screen].workarea
            local w = math.ceil(wa.width * 0.618)
            local h = wa.height * 4 / 10
            wbox = wibox({x = math.ceil(wa.x + wa.width / 2 - w/2),
                          -- y = wa.y + wa.height / 2 - h/2,
                          width = w,
                          -- height = h,
                          ontop = true })
            -- wbox:set_bg(beautiful.bg_normal.."cc")
            wbox:set_fg(beautiful.fg_normal)
            wbox:set_bg("#ffffff00")

            local container_inner = wibox.layout.align.vertical()
            container_layout = wibox.layout.margin(
                container_inner,
                container_margin_left_right, container_margin_left_right,
                container_margin_top_bottom, container_margin_top_bottom)

            container_layout = wibox.widget.background(container_layout)
            -- container_layout:set_bg(beautiful.bg_normal..'dd')
            container_layout:set_bg(beautiful.bg_normal)

            -- constraint:set_widget(layout)
            -- constraint = wibox.layout.constraint(layout, "max", w, h/2)
            -- wbox:set_widget(constraint)
            wbox:set_widget(container_layout)
            layout = wibox.layout.flex.vertical()
            container_inner:set_middle(layout)
        else
            layout:reset()
        end

I can investigate further, if that should not be the case.

psychon:
`wibox.layout.flex:reset()` definitely does emit `widget::layout_changed` and that clears caches. Could you send me some code to experiment with (including instructions on what to do to make it break :-) ) and I'll investigate?

psychon:
I pushed a new commit that fixes the tasklist not updating when two clients are swapped. Does that fix your problem?

blueyed:
No.

blueyed:
Would my issue be fixed by https://github.com/awesomeWM/awesome/pull/458, especially since I am using/setting `wibox.visible`?

> This also fixes a bug where no dependencies were tracked for widgets after
:set_visible(false).

actionless:
the same here -- no more noticeable performance issue, but when widget change size/childs it's not redrawing, reproduces in both tasklist and taglist (blank tags are not shown, so when closing the last client on tag that tag should disappear from taglist)

psychon:
I'll close this. I'll redo this in a different way with hopefully less issues.

Thanks for mention `:set_visible()` so much. I now pushed another commit to this PR that also fixes this issue here. Please test if this fixes your issues with this branch, so that I can be a little more confident in that other branch with a slightly different approach. :-)

blueyed:
@psychon 
Sorry, the last commit also does not fix the display with cyclefocus.
But I'll try your new branch/approach, and we can then debug it from there.

psychon:
Travis broke. Reopening to see if this helps.

psychon:
@blueyed Could you try if reverting https://github.com/awesomeWM/awesome/commit/ee001ce2f0cbc01d6eeaebde50467482af7b578a fixes your issue with cyclefocus? (Was just a random idea. I don't know what cyclefocus does exactly, but it's possible that we have to emit `widget::repaint_needed` when the same image is set again)

psychon:
Superseded by #463 and this was reopened to fix travis (which also now works again)

psychon:
In case anyone ever comes across this and tries to figure out what the problem was:

Thanks to @blueyed giving me a version of his code, I figured out the problem with this patch. The cache doesn't have to be invalidated just because some "sub widget" emitted `widget::layout_changed`. In this case, the first widget in the hierarchy was a `wibox.widget.background` and thus its `:fit()` function was never called. This in turn means that there were no dependencies between this widget and other widgets and thus this widget's cache was never cleared. This is correct, because the `:layout()` callback really does not depend on the results of the child. However, of course the hierarchy still needed to be updated "down this path".

The new approach in #463 doesn't have this problem since it explicitly tracks which paths to follow with a boolean in the hierarchy and doesn't care that the cache (correctly) didn't get cleared.

PR #458, opened 2015-09-17
Explicitly track dependencies between widgets

psychon:
Before this, dependencies between widgets where implicitly discovered by
recursive calls to base.fit_widget() and base.layout_widget(). However, it is
too easy to get this wrong (just call one of these functions from outside of a
widget's :fit() / :layout() function) and the resulting mess would be hard to
debug.

Thus, this commit changes the API so that callers have to identify themselves
and we can explicitly record the dependency between the widgets involved.

This also fixes a bug where no dependencies were tracked for widgets after
:set_visible(false). Whoops...

Sorry for breaking the API for adding this.

### DIFF
--- a/lib/wibox/hierarchy.lua
+++ b/lib/wibox/hierarchy.lua
@@ -12,12 +12,13 @@
 local matrix = require("gears.matrix")
 local cairo = require("lgi").cairo
 local base = require("wibox.widget.base")
+local no_parent = base.no_parent_I_know_what_I_am_doing
 
 local hierarchy = {}
 
 local function hierarchy_new(context, widget, width, height, redraw_callback, layout_callback, callback_arg,
             matrix_to_parent, matrix_to_device)
-    local children = base.layout_widget(context, widget, width, height)
+    local children = base.layout_widget(no_parent, context, widget, width, height)
     local draws_x1, draws_y1, draws_x2, draws_y2 = 0, 0, width, height
     local result = {
         _matrix = matrix_to_parent,
--- a/lib/wibox/layout/align.lua
+++ b/lib/wibox/layout/align.lua
@@ -40,7 +40,7 @@ function align:layout(context, width, height)
     --  if the second widget doesn't exist, we will prioritise the first one
     --  instead
     if self._expand ~= "inside" and self.second then
-        local w, h = base.fit_widget(context, self.second, width, height)
+        local w, h = base.fit_widget(self, context, self.second, width, height)
         size_second = self.dir == "y" and h or w
         -- if all the space is taken, skip the rest, and draw just the middle
         -- widget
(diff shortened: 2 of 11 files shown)


blueyed:
@psychon 
Can you elaborate on how this is related to #452?

psychon:
This doesn't relate to #452 at all. The implicit tracking based on recursive calls to `fit_widget` and `layout_widget` works fine. However, for the scroll layout (#268) I needed to call `fit_widget` from a `:draw` callback and the implicit tracking can't track this correctly (we aren't currently in `scroll:fit()` or `scroll:layout()` which would be needed for the implicit tracking).

So point 1: I needed this for some code.

Then I wondered "What would happen if others write code which calls `fit_widget` or `layout_widget` in a way that would break this implicit tracking?". The answer was "We would have hard to debug issues where in some cases things weren't correctly update when a widget changes its layout". Hard to debug issues are bad and so I thought some more about this and thought about just throwing away all the magic for figuring out dependencies and adding explicit arguments for this instead.

Point 2 thus is: Makes it harder to break the dependency tracking / Avoids hard to debug issues / Allows calling some functions in more places.

As far as I know, this doesn't fix any issues. It hopefully just avoids issues in the future.

psychon:
I had the order of the dependencies messed up (it wasn't clearing caches towards the root of the hierarchy, but away from it). Force-pushed a new commit with this change squashed:
```diff
diff --git a/lib/wibox/widget/base.lua b/lib/wibox/widget/base.lua
index dbe2812..51d3cfc 100644
--- a/lib/wibox/widget/base.lua
+++ b/lib/wibox/widget/base.lua
@@ -49,9 +49,9 @@ local function record_dependency(parent, child)
     base.check_widget(parent)
     base.check_widget(child)
 
-    local deps = widget_dependencies[parent] or {}
-    deps[child] = true
-    widget_dependencies[parent] = deps
+    local deps = widget_dependencies[child] or {}
+    deps[parent] = true
+    widget_dependencies[child] = deps
 end
 
 -- Clear the caches for `widget` and all widgets that depend on it.
```

actionless:
it breaks redraw when number of childs changes

or those fixes are already implemented in https://github.com/awesomeWM/awesome/pull/462

psychon:
@actionless The tasklist works for me (with the latest commit, as explained in my previous comment), what exactly breaks for you?

actionless:
because tasklist is set in the center so it's not changing size if number of its child changes

try:
```lua
awful.widget.taglist(s, awful.widget.taglist.filter.noempty)
```

psychon:
Uhm, sorry, but works for me. Are you sure that you have the very latest version of this PR? Commit 41a8fabf4c64bf62e4

actionless:
my bad, i did not pulled your branch after force-push

PR #461, opened 2015-09-17
Correctly generate mouse::enter and mouse::leave events on drawins, titlebars and clients

psychon:
Hopefully the title is self-explanatory and all the rest is explained in the individual commit messages. All I have to add is: This took me some `printf`-debugging and reading X11 docs to figure out. These input events always turn out to be complicated...

I think that this supersedes #382.


blueyed:
Works for me! Thanks!

PR #462, opened 2015-09-20
Implement full matrix operations in gears.matrix

psychon:
Instead of using matrices from cairo through LGI, implement matrices in pure Lua. That way we don't have to go through LGI for every little thing. Apparently the overhead is quite noticeable...
```
Before:
        create wibox: 0.0826502  sec/iter ( 13 iters, 1.157 sec for benchmark)
    update textclock: 0.0186952  sec/iter ( 57 iters, 2.473 sec for benchmark)
  relayout textclock: 0.0158112  sec/iter ( 64 iters, 1.028 sec for benchmark)
    redraw textclock: 0.0015197  sec/iter (662 iters, 1.861 sec for benchmark)

After:
        create wibox: 0.0825672  sec/iter ( 13 iters, 1.154 sec for benchmark)
    update textclock: 0.00378412 sec/iter (277 iters, 4.216 sec for benchmark)
  relayout textclock: 0.00259056 sec/iter (420 iters, 1.09 sec for benchmark)
    redraw textclock: 0.00105128 sec/iter (958 iters, 1.79 sec for benchmark)
```
Relayout only takes 16% of the previous time and even redraw is faster (69% compared to before, even though we only copied one matrix per widget here, I think...?)

### DIFF
--- a/lib/gears/matrix.lua
+++ b/lib/gears/matrix.lua
@@ -11,16 +11,6 @@ local cairo = require("lgi").cairo
 local debug = require("gears.debug")
 local matrix = {}
 
---- Copy a cairo matrix
--- @param matrix The matrix to copy.
--- @return A copy of the given cairo matrix.
-function matrix.copy(matrix)
-    debug.assert(cairo.Matrix:is_type_of(matrix), "Argument should be a cairo matrix")
-    local ret = cairo.Matrix()
-    ret:init(matrix.xx, matrix.yx, matrix.xy, matrix.yy, matrix.x0, matrix.y0)
-    return ret
-end
-
 -- Metatable for matrix instances. This is set up near the end of the file.
 local matrix_mt = {}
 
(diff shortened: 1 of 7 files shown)


actionless:
now it's fine -- no eye-observable performance issues and problem with fitting new size when child items changed not reproducible no more

blueyed:
For some reason the Travis build does not start, although I've stopped and restarted it (https://travis-ci.org/awesomeWM/awesome/builds/81255650).
Trying to close and re-open the PR to trigger it.

psychon:
@actionless Could you tell me more about the previous performance issues? I'd like to look into them.
(Also, awesome seems to have worse memory usage now, but debugging claims that memory usage does not increase, so no leaks. Anyone else sees something?)
Finally: @blueyed https://travis-ci.org/awesomeWM/awesome/pull_requests #452 seems to have travis blocked.

actionless:
i have a pretty complicated layout of widgets composed from flex, fixed, constraint etc layouts on my panel and each tag switch was happened slowly because of panel redrawing

PR #464, opened 2015-09-21
fix(lib: naughty: core): make it work with textbox per-screen dpi

actionless:
was broken after https://github.com/awesomeWM/awesome/pull/457/files


psychon:
Sorry for missing this and thanks for fixing this. I guess I should finally implement https://github.com/awesomeWM/awesome/pull/398#issuecomment-134938189.

PR #465, opened 2015-09-21
test(benchmark): add for tag switch

actionless:
it's more like e2e test -- it not helps to find out what exactly became slower/faster but can help to measure "overall" rapidity of interface, because speed of tag switching strongly affects an impression of how fast awesome works, ie latest performance issues were most of all noticeable during tag switch

i think it also can be reproduced by calling manual redraw of panel wibox, i am not sure how it will be better

### DIFF
--- a/tests/test-benchmark.lua
+++ b/tests/test-benchmark.lua
@@ -87,9 +87,15 @@ local function update_textclock()
     do_pending_repaint()
 end
 
+local function e2e_tag_switch()
+    awful.tag.viewnext()
+    do_pending_repaint()
+end
+
 benchmark(create_wibox, "create wibox")
 benchmark(update_textclock, "update textclock")
 benchmark(relayout_textclock, "relayout textclock")
 benchmark(redraw_textclock, "redraw textclock")
+benchmark(e2e_tag_switch, "tag switch")
 
 require("_runner").run_steps({ function() return true end })


psychon:
Uhm. I guess your intention is to have one xterm on each tag. You do know that the code doesn't actually work like this, right? (awesome will only manage the xterms after the benchmark completed, so during this benchmark they aren't actually visited yet)

Also, this only really benchmarks `awful.tag.viewnext` and thus `awful.tag.viewidx` (and the tag history). If you want to include redrawing the taglist in the benchmark, you have to also call `do_pending_repaint` in your benchmark function.

psychon:
Fine by me.

Also, my intention with these benchmarks was also just to have some overall impression of speed so that my work on speeding things up can be backed by some numbers. For detailed timing information, a profiler should be used. So this thing fits with my intention.

psychon:
Just because, here are some numbers for this benchmark with awesome 3.5 (as shipped by debian testing):
```
    create wibox: 0.00812339 sec/iter (142 iters, 1.161 sec for benchmark)
update textclock: 1.27652    sec/iter (  1 iters, 1.277 sec for benchmark)
      tag switch: 1.5448     sec/iter (  1 iters, 1.545 sec for benchmark)
```
Since these numbers are meaningless without comparing them to something, here are my numbers for git/master (I removed all the other benchmarks which would break with 3.5):
```
    create wibox: 0.0899627  sec/iter ( 12 iters, 1.17 sec for benchmark)
update textclock: 0.00169582 sec/iter (643 iters, 2.167 sec for benchmark)
      tag switch: 0.207549   sec/iter (  5 iters, 1.242 sec for benchmark)
```
So... we are already doing a lot better than with 3.5 and I have no idea why it claims to need so long for a tag switch.

psychon:
Argh... -.-
We just created hundreds of wiboxes which are all updated on tag switch. Here are some new numbers with things re-ordered so that tag switch is measured first.
```
3.5:    tag switch: 0.0237742  sec/iter ( 45 iters, 1.092 sec for benchmark)
master: tag switch: 0.0255799  sec/iter ( 40 iters, 1.691 sec for benchmark)
```
So master needs 107.6% of the time that 3.5 needed for a tag switch including a wibox update. May I claim that this is about the same time?

Issue #466, opened 2015-09-22
Proper API for figuring out required size of a textbox


psychon:
See https://github.com/awesomeWM/awesome/pull/398#issuecomment-134938189

The idea is to add new functions to textboxes:

    -- Call the next function with the screen's DPI
    tb:get_preferred_size(s)
    -- Assume unlimited space is available and return needed width and height
    tb:get_preferred_size_at_dpi(dpi)

If we can come up with any use for them (naughty applying a maximum width to notifications?), we could also add the following functions:

    -- Call the next function with the screen's DPI
    tb:get_preferred_height_for_width(s, width)
    -- At the given width, which height is needed to show all the text?
    tb:get_preferred_height_for_width_at_dpi(dpi, width)

Is it ok to require a screen/dpi argument or can we somehow come up with a sane default value? Any other ideas/suggestions for the API?

PR #467, opened 2015-09-22
Fix tooltip.set_geometry: provide textbox context

blueyed:
This is another fix like https://github.com/awesomeWM/awesome/commit/f2e554de91d0c275292e418de59784141a9780e1.

1. I wonder if `nil` should be handled as an empty context maybe? (but it is also fine to use just `{}` then)
2. Could a missing `dpi` value in the context be handled better, e.g. by using the textbox' screen / its dpi automatically?

I've used `mouse.screen`, which is also used with `placement.no_offscreen`.

### DIFF
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -82,8 +82,9 @@ end
 -- @tparam tooltip self A tooltip object.
 local function set_geometry(self)
     local my_geo = self.wibox:geometry()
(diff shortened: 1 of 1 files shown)


psychon:
Well, long-term the plan would be that only `wibox.drawin` creates contexts and thus that certain entries can be guaranteed to exist. We'd have to document what these are somewhere. However, for the scroll layout I also needed to construct a "fake context"....

The long-term plan for fixing what this PR fixes would be the API from #466 .

The short-term plan is: :+1:

Issue #410, opened 2015-08-14
Menubar only looks for icons in /usr/share/icons/hicolor


jaysonwillson:
Hello.
I realized, that menubar only shows icon for one category: internet. I have checked and found out, that this is the only category, for which icon exists in /usr/share/hicolor. After it I started to suspect something and decided to check:
echo 'naughty.notify({text = menubar.utils.lookup_icon("applications-internet.png")})' | awesome-client
returns /usr/share/icons/hicolor/48x48/apps/applications-internet.png,
but
echo 'naughty.notify({text = menubar.utils.lookup_icon("applications-multimedia.png")})' | awesome-client
returns nothing, though
locate applications-multimedia
reports, that such icon exists in
/usr/share/icons/Adwaita/(size)/categories...
/usr/share/icons/gnome/(size)/categories...
/usr/share/icons/mate/(size)/categories...
And many other places.
It seems like menubar.utils_lookup_icon only looks for category icons is /usr/share/hicolor.
Of course, I can set full paths to icons in menu_gen.lua, but it seems to me, that there should be another way, more universal.

JulianVolodia:
I don't understand, but have time to reproduce and maybe try to fix :)

Command
```shell
echo 'naughty.notify({text = menubar.utils.lookup_icon("applications-internet.png")})' | awesome-client
```
gives me error:
> [string "naughty.notify({text = menubar.utils.lookup..."]:1: attempt to index global 'naughty' (a nil value)

Please send which verison of Awesome you use, screenshots showing actual problem.

Best regards.

jaysonwillson:
Please forgive me, if I have not explained clearly.
I will try to be more concrete now.
In https://github.com/awesomeWM/awesome/blob/master/lib/menubar/menu_gen.lua, lines 33-54, menubar categories (which you can see when you launch menubar) are declared. 
"icon_name = "applications-multimedia.png", use = true }" string defines, which icon should be used for a specific category. menubar.utils.lookup_icon() is used to look for icons (lines 56-61)
In my case, when I launch menubar, only "Internet" category has icon, other categories do not have it. 
I have checked, if there is any difference between "applications-internet.png" icon and, for example, "applications-development.png" icon. I have realized, that "applications-internet.png" is the only icon among applications-{multimedia, development, etc...} icons used by menubar, which exists in the "hicolor" icon suite.
[...]

PR #468, opened 2015-09-22
Enhance menubar icon path lookup

nuko8:
The way of icon path lookup for `menubar` is enhanced so that it is
based on a theme-oriented way as described in the specification:

  Icon Theme Specification, Ver. 0.12
  http://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html

To accomplish this:

 * Add the two new files `icon_theme.lua` and `index_theme.lua`.

   The former implements an icon lookup algorithm suggested in the URL
   above.  The latter implements a helper object to parse the cache file
   `index.theme` of which data is used by the former.

 * Modify `menu_gen.lua` to use the new algorithm.

   - The implementation of `lookup_category_icons` is changed
     accordingly.
   - The values of the field `all_categories.icon_name` are changed file
     names to icon names, i.e., file extensions which are used to
     indicate image file formats are removed.

 * Add the new file `icon_theme_spec.lua` for a unit test for checking
   if `icon_theme.lua` together with `index_theme.lua` works as
   expected.

----
I hope this fixes Issue #410.

The lookup algorithm depends on the status (i.e., defined or undefined) or the values of `$XDG_DATA_HOME` and `$XDG_DATA_DIRS`.  So, even if the modules work correctly, the unit test can fail in case those environmental variables are handled incorrectly.

~~The default value of the icon theme name is `gtk-icon-theme-name`.~~  If not set,  the value is selected among`Adwaita`, `gnome` and `hicolor` in this order.

The icon theme name is can be set via `theme.icon_theme` of `theme.lua`, and it overwrites the default value.

I'm not familiar with unit test stuff.  So, I'm quite open to rewriting it if necessary.

Of course, other feedbacks are also welcome.


nuko8:
Hmm...all checks have failed...

I'm using lgi 0.8.0 (not sure of the patch level) and all the unit tests have passed with it.

Downgrading it to 0.7.1, then I got a bunch of errors not in the new test, but in other tests `color_spec.lua`, `matrix_spec.lua` and `hierarchy_spec.lua`.

It looks a bug of lgi.

How should I do with it?   Should I take back the new unit test? :-)

psychon:
I don't think we can use Gtk through Lgi. Sadly. Its quite easy to get lockups when using some functions from a window manager. I'm not sure about why the tests failed, but they failed while loading Gdk and I think Gdk is only loaded as a dependency of Gtk. Perhaps some incompatibility with the Gdk version on ubuntu 12.04? One more reason that we cannot use Gtk. Sorry.

I looked a bit at the commit (we can use Gio.GFile to check if something is a directory), but didn't look too carefully at this. Will try to do so later.

Anyway, thanks for working on this!

nuko8:
Still I got red, but I don't feel bad so much.

There are two possible causes of the failure.  

One is that the Linux systems used by the CI test didn't have Adwaita or gnome icon suite there.  In short, that wan an expected failure and rather proved that my code was sane and faithful :-)

The other is, of course, a bug in my code.  On my Mac, moving icon suites here and there, and setting, unsetting and changing  `$XDG_DATA_DIRS`, my code worked as expected.  So I do need reviewers' stringent eyes for possible flaws in my code.

As @psychon suggested, I dropped the dependency on `lgi.Gtk` and re-implement `is_readable_directory` using `Gio.File` with `Gio.FileInfo'.

Owing to the former, I think we were able to go one step further to resolve the CI test failure. 

As to the latter, I hope @psychon's concerns are all gone.

About the magic number 21.  I used it because I happened to see it in the implementation of `awful.util.file_readable()`.   Since the reference manual of the function was so terse, I needed to make sure of the behavior of the function when it got a directory path.  I began reading the implementation and found the number there.  Sure, I was also aware of the consequences then, but...  I'm wondering why @blueyed didn't mention it :-)

nuko8:
I think this bug fix is not relevant to the CI test.

That's about a change of constructor's argument order.  Once it was `(base_directories, icon_theme_name)`, but later the order has been reversed so that the chances of omitting arguments can increase.  This bug fix is for those I forgot to change.

Since the bugs were found only in part of fallback directory calculation which is to be done when `index.theme` has a key called `inherits`, and that key is not used with Adwaita, gnome or hicolor.  So it is unlikely to cause the CI test failures.

nuko8:
Looks all checks have passed.

This was done by making the CI test self-contained, i.e.,

1 Use the icons in `icons` of the source tree.
2. Make a small icon-theme-like folder structure in `spec/menubar` and have the test work with it.

Also, the implementation of `index_theme.lua` was modified in such a way that, ..., well, ... it looks like more Lua programming, not a lunatic one as if a C programmer were making a chain of if statements, murmuring, "Lua should have `continue`s..." :-)

As you see,  the existing commits were rebased an squashed into a new single commit.

psychon:
Good idea, thanks for working on this.

It's unlikely that I will have the time to look at this closely any time soon, so I propose to use the same approach that lead us to the menubar in the first place: Trust @nuko8, merge this and wait for bug reports. :-)

lib/menubar/icon_theme.lua
```diff
@@ -0,0 +1,262 @@
...
+    table.insert(dirs, GLib.get_home_dir() .. "/.icons")
+    for _, dir in ipairs(GLib.get_system_data_dirs()) do
+        table.insert(dirs, dir .. "/icons")
+    end
+    table.insert(dirs, "/usr/share/pixmaps")
+
+    return dirs
+end
+
+local is_readable_directory = function(path)
```
blueyed: Please move this to `lib/awful/util.lua`.
And then `util.file_readable` could make use of it, too - via a common method.

nuko8:
I've just pushed a fix-up commit to resolve the issues raised by @psychon on 3af0d8a .

Both the removal of `require("beautiful")` and imposing stricter conditions on the tests are done.

lib/menubar/icon_theme.lua
nuko8: I guess you meant the following part of the code, right?
```
local Gio = require("lgi").Gio

local is_readable_directory = function(path)
    local gfile = Gio.File.new_for_path(path)
    local gfileinfo = gfile:query_info("standard::type,access::can-read",
                                       Gio.FileQueryInfoFlags.NONE)
    if not gfileinfo then return false end -- practically ENOENT
    local is_dir = (gfileinfo:get_file_type() == "DIRECTORY")
    local is_readable = gfileinfo:get_attribute_boolean("access::can-read"
    return is_dir and is_readable
end
```

As for me, I'm OK to keep waiting for the last word on this PR for a while.  If a decision is made to move the part of the code to `util.file.readable`, it's also OK for me to modify this PR accordingly.

blueyed: OK.  We could also refactor it later / separately.

nuko8:
This fix-up is a sort of cosmetic change.

It appears that GitHub's file viewer thinks an image file of zero size as invalid;  it keeps a loading icon rotating for a while and then concludes, "Sorry, something went wrong. Reload?"

For the sake of everyone's peace of mind, the `*.png` files in `spec/menubar/icons` were replaced with symlinks against the real image files in `icons`.

nuko8:
This additional commit is substantial, although it should not change the current icon lookup behavior (people hardly see behavioral difference between the current implementation and that of this commit).

The current implementation of that part of the code is based on the assumption that an empty line separates sections.  But this is not part of the specification (There's little about the format of `index.theme` there.  Who can say it's right or wrong? :-)).  

The new implementation uses the pattern `^%[(.+)%]$` as a criterion of separation so that the resulting parser can be more robust.

psychon:
Idea for "future work" if the parser really might have problems: Instead of writing our own parser, use GLib's KeyFile: https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html

```lua
glib = require("lgi").GLib
file = glib.KeyFile()
print(file:load_from_file("/usr/share/applications/pidgin.desktop", glib.KeyFileFlags.NONE))
print(file:get_start_group())
for k, v in pairs(file:get_keys("Desktop Entry")) do print(k, v) end
print(file:get_value("Desktop Entry", "Name"))
print(file:get_locale_string("Desktop Entry", "Name", nil))
```
Output:
```
true
Desktop Entry
1	Name
2	Name[de]
3	GenericName
4	GenericName[de]
5	Comment
6	Comment[de]
7	Exec
8	Icon
9	StartupNotify
10	Terminal
11	Type
12	Categories
13	X-Desktop-File-Install-Version
Pidgin Internet Messenger
Pidgin Internet-Sofortnachrichtendienst
```

nuko8:
Thank you for your note.  Since I didn't know that material, it is very interesting and helpful for me.

At the beginning of the document at the link you pointed, we see

> Groups are started by a header line containing the group name enclosed in '[' and ']', and ended implicitly by the start of the next group or the end of the file. Each key-value pair must be contained in a group.

That is just what I wanted to address with the last commit.  (What I call *section* is called *group* there, though).

> "future work"

OK, but I really think it's worth a try in another branch in my repo.  If `KeyFile` gives us a satisfactory result, use of it could improve the maintainability of the code.  Perhaps, overhead and speed are only concerns I should care about.

nuko8:
I tried `lgi.GLib.KeyFile`:
https://gist.github.com/nuko8/2fe853b7de8d4644c3b5

As you see, the code becomes shorter and readable, and I like it.  But sadly, it caused noticeable delay as predicted.

Although I didn't measure the delays mechanically,  `KeyFile` took about 3 seconds while this PR's implementation less than 1 second (almost promptly) to show the menubar of non-cached state on my relatively old machine (2.4GHz Intel Core 2 Duo) when loading the hicolor (fallback, i.e., worst case) icon theme.

As to Adawaita or Gnome icon themes (normal cases), `KeyFile` took about 1 second to load and show the menubar.

So, if there's no room for optimization we can try with the current `KeyFile` implementation, I think keeping this PR's implementation would be a better choice.

nuko8:
Squashed into a single commit and rebased onto the current master.

Elv13:
Sorry for comming late into the game here, but isn't better to simply use GTK for this, I never had any lockup with that code path and GTK provide 100% XDG icon theme support with caching https://github.com/Elv13/awesome-configs/blob/master/utils/fd_async.lua#L414

Code like this PR is still necessary because Awesome doesn't depend on GTK, but I am not sure sure about the performance compared to just using GTK (when we can)

nuko8:
Thank you for your comment.

Though I'm not sure of what you mean by "the performance" exactly, let me try to answer.

Other than retrieving directory prefixes and checking directory readability, this PR's implementation doesn't rely on any external library or runtime; its lookup algorithm is purely written in Lua.  So, I believe the performance of this PR is fast enough in comparison with other means.

nuko8:
**Performance Comparison Test between `icon_theme:find_icon_path()` and `module.icon.load()`**

I did a performance comparison test to compare `icon_theme:find_icon_path()` with `module.icon.load()` of
https://github.com/Elv13/awesome-configs/blob/master/utils/fd_async.lua#L414

In order to make the comparison easier, I modified `module.icon.load()` in such a way that it returns the icon path:
https://gist.github.com/nuko8/ed14beb61e8257e36148

I used a slightly modified version of the default `rc.lua`:
https://gist.github.com/nuko8/525334a004a661143c8c

I put the modified version of `fd_async.lua` and `rc.lua` in `~/.config/awesome` together with `icon_theme.lua` and `index_theme.lua`, and run the latest github Awesome on Mac OS X 10.9.5, 2.4GHz Intel Core 2 Duo.

The result obtained is as follows:

Trial | `icon_theme:find_icon_path()` | `module.icon.load()`
---:|---:|---:
1 | 0.047 | 0.138
2 | 0.046 | 0.138
3 | 0.046 | 0.141
4 | 0.046 | 0.135
5 | 0.046 | 0.132

As the test itself is a little bit naive, the numbers themselves are not so important, but the tendency is clear:  The overheads incurred by external libraries or runtimes (`Gtk/Gdk/GObject/lgi`) are significant in this particular case.

I hope the report above gives an satisfactory answer to the question raised by @Elv13 .

PR #470, opened 2015-09-23
Never explicitly focus the root window

psychon:
Whenever client.focus == nil, we set the input focus to the root window to
express "nothing has the input focus". However, thanks to the way X11 input
works, this means that whatever is under the mouse cursor gets keyboard input
events. This can easily be reproduced with urxvt and some small addition to the
config to unfocus things.

This commit changes things. Instead of focusing the root window, we create a
special "no focus" window that gets focused if we want nothing to have the
focus.

Edit: Note that this will not cause any change in behaviour that would be visible to Lua code. This should only make a difference for users.

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -569,23 +569,27 @@ main(int argc, char **argv)
     xkb_init();
 
     /* The default GC is just a newly created associated with a window with
-     * depth globalconf.default_depth */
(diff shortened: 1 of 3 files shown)


blueyed:
Good idea!

Issue #455, opened 2015-09-16
Awesome not changing tag on URL clicking.


osleg:
2 tags, first tag with terminal second with firefox
Expected behaviour:
In terminal I have a URL, clicking on this url should focus FF and switch the tag to second.

Current Behaviour:
Clicking on link making focus switch to FF from terminal but the tag does not switching.

blueyed:
This is with master?
Has it changed recently?
Is it related to Firefox only?

osleg:
It is with master.
I've not idea when did it change since i was using 3.5.6 stable and then moved to git instantly.
Sadly I have no idea no ideas what else should work this way so I'm not sure whether it's FF only :)

blueyed:
You could try with another browser, from the command line.
I will look into this later in more depth.

actionless:
can this one fix your issue?
https://github.com/awesomeWM/awesome/pull/438

blueyed:
I can reproduce it using the default config (regarding `ewmh.activate`, the `request::activate` handler).

I've tried to bisect it, but it seems to be an issue for longer (HEAD~100 at least).
I've also tried #470, but that did not help.

The problem here is the unconditional `client.focus = c`, which will nowadays move the focus to the non-visible client.  This was not the case a while ago, and then the tag/client would have been marked as urgent.
That would be the expected behavior currently - and there is some (not really correct) [FAQ](http://awesome.naquadah.org/wiki/FAQ#Why_are_new_clients_urgent_by_default.3F) about restoring the previous behavior in that case.

Additionally, there appears some (newer?) issue with the taglist, which won't be updated for the urgent client in it.  The `urgent_count` seems to be handled properly, and moving to another tag (not the urgent one), will then display it as urgent.
The latter issue gets fixed / worked around with #474.

blueyed:
The problem with the tag widget not highlighting the urgent tag is this code (from https://github.com/awesomeWM/awesome/blob/0414a579c280a78887bf74b754389216311511ed/lib/awful/widget/taglist.lua#L49-L94):

    local sel = capi.client.focus
    …
    if sel then
        if taglist_squares_sel then
            -- Check that the selected client is tagged with 't'.
            local seltags = sel:tags()
            for _, v in ipairs(seltags) do
                if v == t then
                    bg_image = taglist_squares_sel
                    bg_resize = taglist_squares_resize == "true"
                    is_selected = true
                    break
                end
            end
        end
    end
[...]

PR #474, opened 2015-09-24
ewmh.activate: only focus visible clients

blueyed:
Since focus can be moved to non-visible clients nowadays, this needs to be
checked in the `request::activate` handler.

Fixes https://github.com/awesomeWM/awesome/issues/455.


psychon:
Fine by me

PR #475, opened 2015-09-24
Rename client_maybevisible to client_on_selected_tags

blueyed:
This makes it more clear what it is about, and also fixes its
documentation.


psychon:
:+1:

PR #476, opened 2015-09-24
awful.widget.taglist: handle/display urgent property always

blueyed:
For non-selected tags, display the urgent bg/fg always.
Previously, this was skipped in case the tag had the focused client,
which is possible since a while.

Ref: https://github.com/awesomeWM/awesome/issues/455#issuecomment-143072993


psychon:
Seems fine to me

Issue #471, opened 2015-09-24
Add luaA_tolstring() / reminder for psychon


psychon:
See https://github.com/awesomeWM/awesome/pull/384

Also:

     #define luaL_tolstring you_should_use_luaA_tolstring_instead

:-)

blueyed:
Fixed as per https://github.com/psychon/awesome/commit/8a1db9d75510c662983ca84fc91e8c926fd26f7e?

PR #480, opened 2015-09-27
Some minor things

psychon:


lib/awful/client.lua
```diff
@@ -428,7 +428,7 @@ end
 --- Cycle clients.
 --
 -- @param clockwise True to cycle clients clockwise.
--- @param[opt] screen The screen where to cycle clients.
+-- @param[opt] s The screen where to cycle clients.
```
blueyed: Found that one, too, today.. :)

psychon:
Yup. I'll ignore the `#define` that I propose in #471 since this PR doesn't add this function to e.g. luaa.h. The plan is that we move this to the right place if we ever also need this function elsewhere.

blueyed:
Merged in https://github.com/awesomeWM/awesome/commit/5f6bbdfc63d2b74af3d7a7bff0257526a5d2dd8b and the commits before.

Issue #163, opened 2015-03-11
Implement lua __index and __newindex for capi objects


Elv13:
client, tag, screen, drawable, key and drawin should have their (unknown) property handler delegated to lua.

This API could be used to:
* Implement dynamic properties (replace awful.{...}.getproperty / setproperty)
* Forward property to awful.{client, tag,...}.set_ and get_ when they exist
* Other use

Those handlers should be per object (because someone will eventually complain if they aren't) but always fallback to a common (per class) handler.

New API for all objects:
.index_handler (property with set and get)
.newindex_handler (property with set and get)

Elv13:
```[11:30] <psychon> Elv1313: that would make the code for setting titlebars harder to use / kinda ugly
[11:30] <psychon> I wonder if we can come up with something that is "transparent" to uses in rc.lua
[11:31] <Elv1313> psychon: yes, but it will require being able to set custom metatables for client
[11:31] <Elv1313> that way we can use __index to hijack .border_width and __newindex for the setter. Then it wont break the API and just work
[11:31] <normalra> Exactly.
[11:31] <psychon> hm... might be a good idea to have
[11:32] <psychon> well, rather than setting the metatable, it might be easier to be able to set handlers for __index and __newindex for "property misses" (when the C code doesn't know how to handle things)
[11:32] <Elv1313> I tried to implement that the other day, but ran out of time without much results (I don't know the luac API that well)
[11:32] <Elv1313> psychon: yes, that would work
[11:35] <psychon> "just" requires patching luaA_class_newindex() and luaA_class_index()
[11:35] <psychon> instead of calling newindex_miss_property, they'd need to call into some lua function (that somehow lua must be able to set earlier...)
[11:38] <Elv1313> psychon: if you take care of the C side, I will implement the handlers. I already have the code in Radical to do just that
[11:38] <Elv1313> it also emulate :set_foo(x) -> object.foo = x
[11:41] <Elv1313> psychon: do you want me to create an issue for that?
[11:41] <psychon> yeah, feel free
[11:41] <psychon> and we need to come up with some API for lua to set the handlers
[11:41] <psychon> does it make sense to have them object-specific? or is it ok to have it just per-class?
[11:42] <blueyed> psychon: good reasons against such a tag.  It would be nice for "git describe", e.g. for the awesome-git version in Nix/NixOS, but I can use a custom version like "3.6-git-g4636b11" myself.
[...]

PR #482, opened 2015-09-27
Make it possible for Lua to emulate arbitrary properties

psychon:
This makes it possible to add something similar to a __index / __newindex
metamethod to all our C objects. Based on this, Lua can then easily implement
arbitrary properties on our capi objects.

blueyed told me to clean this up and submit it. Example usage (untested) would be the following (should we just dump this into awful.init so that it works on all capi objects? Anyone wants to do a PR for this?)
```lua
local client_props = setmetatable({}, { __mode = "k" })
local function get(c)
    local props = client_props[c]
    if not props then
        props = {}
        client_props[c] = props
    end
    return props
end
client.set_index_miss_handler(function(c, k)
    return get(c)[k]
end)
client.set_newindex_miss_handler(function(c, k, v)
    get(c)[k] = v
end)
```

Edit: Sorry, the above doesn't work. With Lua 5.1, the client would become un-garbage-collectable after `c.prevent_gc = function() print(c) end`.

### DIFF
--- a/common/luaclass.c
+++ b/common/luaclass.c
@@ -251,6 +251,8 @@ luaA_class_setup(lua_State *L, lua_class_t *class,
     class->parent = parent;
     class->tostring = NULL;
     class->instances = 0;
+    class->index_miss_handler = LUA_REFNIL;
+    class->newindex_miss_handler = LUA_REFNIL;
 
     signal_add(&class->signals, "new");
 
@@ -361,6 +363,40 @@ luaA_class_property_get(lua_State *L, lua_class_t *lua_class, int fieldidx)
     return NULL;
 }
 
+/** Call a registered function.
+ * \param L The Lua VM state.
+ * \param handler The function to call.
+ * \return The number of elements pushed on stack.
+ */
+static
+int luaA_class_call_handler(lua_State *L, int handler)
+{
+    /* This is based on luaA_dofunction, but allows multiple return values */
+    assert(handler != LUA_REFNIL);
+
+    int nargs = lua_gettop(L);
+
+    /* Push error handling function and move it before args */
+    lua_pushcfunction(L, luaA_dofunction_error);
+    lua_insert(L, - nargs - 1);
+    int error_func_pos = 1;
+
+    /* push function and move it before args */
+    lua_rawgeti(L, LUA_REGISTRYINDEX, handler);
+    lua_insert(L, - nargs - 1);
+
+    if(lua_pcall(L, nargs, LUA_MULTRET, error_func_pos))
+    {
+        warn("%s", lua_tostring(L, -1));
+        /* Remove error function and error string */
+        lua_pop(L, 2);
+        return 0;
+    }
+    /* Remove error function */
+    lua_remove(L, error_func_pos);
+    return lua_gettop(L);
+}
+
 /** Generic index meta function for objects.
  * \param L The Lua VM state.
  * \return The number of elements pushed on stack.
@@ -397,6 +433,8 @@ luaA_class_index(lua_State *L)
     }
     else
     {
+        if(class->index_miss_handler != LUA_REFNIL)
+            return luaA_class_call_handler(L, class->index_miss_handler);
         if(class->index_miss_property)
             return class->index_miss_property(L, luaA_checkudata(L, 1, class));
     }
@@ -427,6 +465,8 @@ luaA_class_newindex(lua_State *L)
     }
     else
     {
+        if(class->newindex_miss_handler != LUA_REFNIL)
+            return luaA_class_call_handler(L, class->newindex_miss_handler);
         if(class->newindex_miss_property)
             return class->newindex_miss_property(L, luaA_checkudata(L, 1, class));
     }
--- a/common/luaclass.h
+++ b/common/luaclass.h
@@ -74,6 +74,10 @@ struct lua_class_t
     unsigned int instances;
     /** Class tostring method */
     lua_class_propfunc_t tostring;
+    /** Function to call on index misses */
+    int index_miss_handler;
+    /** Function to call on newindex misses */
+    int newindex_miss_handler;
 };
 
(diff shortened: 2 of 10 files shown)


psychon:
Corresponding feature request is #163

Elv13:
I propose this
https://github.com/Elv13/radical/blob/master/object.lua

as a base for the implementation with whatever property table awful.client and awful.tag have as the fallback for unknown properties

xerxesbeat:
a few points of clarification:

This appears to store properties in two separate places depending on whether they are considered 'arbitrary' i.e. custom.  If this is indeed desired, the case may arise where a 'custom' property is set with a direct call to override a non-custom property; how should this be handled?

As per Elv13's comment, it seems there is a desire for a default function as a callback for get/set when a property is deemed 'custom', which may be overridden by an extension on a per-object-instance basis.  This callback seems to further default to a systematic concatenation of the property key with a standard prefix, which could likely produce inconsistent results depending on whether the default behavior is assumed (attempting a call on set_foo directly) or when operating via the handler function ( c.foo = 42 ) in the event that one or more extensions are attempting to override default behavior of custom properties (I am told this is for consistency in API formatting, and there may yet be a means to implement this behavior without the aforementioned discrepancies, if desired)

Any further details/input are most appreciated

Elv13:
Ok, let me document this properly

### Issues

 1. The API is inconsistent. You have 2 places for every `capi` objects. A method can be either `c:foo(...)` or `awful.client.method(c, ...)` and new users are missing that all the time.
2. The API naming is inconsistent. `awful.client.property.get` and `awful.tag.getproperty` are the worst examples. You also have `getfoo`, `get_foo` and `foo.get`. It is too late to break those in an API compatible way, but if the `local foo = c.foo` and `c.foo = "bar"` (and `c:set_foo("bar", "baz")`) could be standardized, then at least there would be one consistent API.
3. Extending the `capi` object is currently complicated. Modules may monkeypatch new methods to `awful`, but looking at the existing ones, they mostly add their own APIs, making them harder to use. (/me blame myself for being the worst offender for this)

### Solution

My proposed solution is to be able to add custom properties directly accessible from the `capi.client` `capi.tag` objects.

If `c.foo` is called, then I propose to check the "object table" if `get_foo()` is declared. If it is, call it and return the retval. As a fallback, just querry `awful.client.property.get`

If `c.foo = "bar` is called, then check if `set_foo` is defined. If it isn't, then uset `awful.client.property.set`. `set_foo` can take multiple argument, but only the first 2 (the `client` or `tag` and the value to be set) are required. Any additional parameter can still be accessed by calling `set_foo` without the property hack.

### Why

While a setter could be called from listening to `property::prop_name` signal, this isn't always enough. The common example is when the argument is mutated before being set. For example, `c.width` must be an integer, calling `c.width = 0.5` will "crash" Awesome and load the fallback config. The `set_width` function can take care of that before really setting the value. But the main problem is the getter. Calling  `awful.tag.getproperty` is only able to return "dumb" properties. Some properties, like client index, are computed when the accessors is called. This require the additional optional `get_foo` indirection wrapper on top of  `awful.tag.getproperty`.

Another advantage is being able to add new mutator method callable directly using `c:mymutator(...)`. It would be possible to manually add them to `awful.client.property.get` in a `capi.client.connect_signal("manage"...)` hook, but it is ugly. Having a "standardised" way of adding new method is much nicer. This way doing `awful.object.client.mymutator = function(c, args) end` would allow to globally add new client methods.

psychon:
@blueyed Why did you assign me? Is anything wrong with this PR? My understanding is that this adds the necessary stuff to the C side and that the Lua part can be added in a separate PR (and I wouldn't mind if it is added by someone else).

@Elv13 I'm not commenting at the magic/complicateness of that API, but I like it a lot that your code has the properties table as an upvalue. That should fix all the memory leak problems that I was worries about, nice!

xerxesbeat:
@psychon I apologize, I offered to help with this PR a couple months ago via IRC and have since disappeared.  I was tasked with finalizing the implementation provided that began this, and was unable to clearly define what work had yet to be done as well as the proper approach to implementing the remainder of the features requested.  I cannot speak for the others involved, but I can only imagine my lapse here has caused the delay of this 'feature' implementation.

psychon:
@xerxesbeat I don't think that you caused anything, especially not any delays. All's well! :-)

blueyed:
@psychon 
I had the feeling that @Elv13's comments need to be addressed by use?

psychon:
Ok, so I just merged this. I *think* that this adds the necessary support code to the C code for this and now just needs someone to implement the Lua side of this. For that I opened https://github.com/awesomeWM/awesome/issues/747.

(If someone comes up with a nicer way to do the C side of this, we can still change the interface later.)

xerxesbeat:
I realize you three seem to be the major contributors to awesome wm at the present, and with all due respect... as a random person who poked their head in and ended up in this particular place (I've been using awesome wm for a few years now) I'd like to offer my two cents if it is welcome (It's not exactly pretty)

psychon:
You are asking for permission to tell us something bad? I'm curious and I think I have thick skin. Plus, I'd like to improve. :-)
(And sorry that you somehow ended up in the middle of this "emulate arbitrary properties"-issue)

Oh and since this PR is merged, I don't really mind using it for some unrelated discussion.

psychon:
Partly :+1:, partly "it's already possible to have collisions, the syntax to do so just becomes shorter (`c.foo = 42` vs `awful.client.property.set(c, "foo", 42)`). :-)

Elv13:
Actually, there is *less* collisions c.minimized = true; `awful.client.property.set(c, "minimized", false)` will create a schrodinger property. They will both emit the "property::minimized" signal and confuse the hell out of signal listeners. Having a single system that, in the worst case, alias each others if way better than having 2 property systems.

Then the extensibility, API coherence and ease of use arguments still apply.

PR #483, opened 2015-09-27
Elv13 prompt hook v2

blueyed:
Continuation of #52, assumes #34 to be merged.

/cc @Elv13

### DIFF
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -224,7 +224,26 @@ end
 -- user editing.
 --
 -- @tparam table args A table with optional arguments: `fg_cursor`, `bg_cursor`,
--- `ul_cursor`, `prompt`, `text`, `selectall`, `font`, `autoexec`.
+--   `ul_cursor`, `prompt`, `text`, `selectall`, `font`, `autoexec`, `hooks`.
+-- @tparam[opt] table args.hooks The "hooks" argument uses a syntax similar to
+--   `awful.key`.  It will call a function for the matching modifiers + key.
+--   It receives the command (widget text/input) as an argument.
+--     hooks = {
+--       -- Apply startup notification properties with Shift-Return.
+--       {{"Shift"  }, "Return", function(command)
+--         mypromptbox[awful.screen.focused()]:spawn_and_handle_error(
+--           command, {floating=true})
+--       end},
+--       -- Override default behavior of "Return": launch commands prefixed
+--       -- with ":" in a terminal.
+--       {{}, "Return", function(command)
+--         if command:sub(1,1) == ":" then
+--           command = terminal .. ' -e ' .. command:sub(2)
+--         end
+--         mypromptbox[awful.screen.focused()]:spawn_and_handle_error(
+--           command)
+--       end}
+--     }
 -- @param textbox The textbox to use for the prompt.
 -- @param exe_callback The callback function to call with command as argument
 -- when finished.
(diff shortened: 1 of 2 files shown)


psychon:
Fine by me

Elv13:
+1

So, I guess you want to rebase this one first?

blueyed:
Rebased.  Let's see if it still passes.. ;)

blueyed:
@Elv13 mentioned that this would be (would have a potential for) a nice addition to the default config, but that could be done in a new PR.

blueyed:
Rebased and seems to be fine - but https://github.com/awesomeWM/awesome/pull/489 should go in first.

Elv13:
While I think about it, another PR should also add an API to add new default hooks for the prompt. This could be used to implement VI style prompt shortcuts by adding escape. For this, layered keygrabbers would also need to be properly supported, but this is just an example (EMACS mode could be implemented (well, extended given `CTRL+A/E` and some others are already implemented) directly). 

/me also wonder is generic hooks for modified text could be added to implement syntax highlight, but this is just dreaming and would required substantial changes to decouple the prompt text and the displayed one

Bew78LesellB:
@Elv13 I'm currently doing a big library, which will have (with a lot of other features) layered grabbers and dynamic keymap, for every possible use, from tag management to prompt editText with vim style keybinds.. This is in WIP, and doesn't work ATM, but it will ;) 

Maybe when finished (in a long time..) I will make a separate library with only key* stuff. For those who doesn't want the full lib.

Elv13:
@Bew78LesellB Well, I guess the issues with layered grabbers should be treated as bugs in Awesome and fixed internally rather than adding an abstraction layer on top the the current API (the internal one should handle it)

Bew78LesellB:
@Elv13 yeah this is another way to think the issue, but I still didn't progress in the study of Awesome's C code, so I'm currently making what I'm good at.. (love lua now!) 
If you can fix this internally, this will surely speed-up things!

blueyed:
@Bew78LesellB 
I am not sure that you will need too much from the C side in awesome.

I think it would be a great opportunity for you to address this where it's meant to be fixed.  Whenever you work around something, just consider fixing it in the core.. :)

Bew78LesellB:
@blueyed Oh I never thought of that !

The thing is that I added some concept in my lib, on top of the lua api (well, in fact, on top of the capi, I'm rewriting a lua api..), and I think everything with theses concepts.. 

As they are kind of specific to me, I'm not sure it's a good idea to add them in the core of awesome..

blueyed:
Rebased on master (fixing a conflict), added some doc, moving to "next" milestone.

blueyed:
I've also made the hook's return value handling a bit smarter in https://github.com/blueyed/awesome/commit/7925e6a6f5d3e42191dbcaea1958808761b604dc.  Does that make sense?

blueyed:
@psychon @Elv13 ping

psychon:
Uhm. Your latest commit looks fine to me (thanks!) and the rest was already decided to be good before, right? So I guess everything is good (sorry, I don't really want to look too deeply into the prompt code.)

blueyed:
https://github.com/blueyed/awesome/commit/7925e6a6f5d3e42191dbcaea1958808761b604dc is also new.

psychon:
I trust you on that one. :-)

Elv13:
+1

PR #485, opened 2015-09-27
cmd: Add an option to include an extra directory

Elv13:
This could be done using environement variables or some
symlinks in ~/.config/awesome, but I think it make development
easier. A future patch could also do the same for the cache
directory.


psychon:
Some quick thoughts:
- Would be nice if this property could be specified multiple times, I guess
- man pages need to be updated (what do we do with the translations?)
- You can get rid of `pathbuf`  in luaA_init() by just pushing all strings on the stack and using `lua_concat()`. That also removes the 256 bytes length limit.

Elv13:
TODO detect ';' and split the string to the suffix is added for each entries

PR #486, opened 2015-09-27
widget: Add a new container declaration syntax

Elv13:
This new syntax is inspired by the Awesome widget 3.2-3.4 API. It
allow cleaner widgets declaration. The produced code is usually much
shorted and easier to read than wibox.widget imperative syntax.


Elv13:
Ok, I fixed Travis, I broke it again when I moved the `layout` to the top, a `,` was missing

actionless:
what do you think about such approach to API? https://github.com/actionless/awesome_config/blob/devel/actionless/widgets/common.lua#L182-L208

psychon:
@actionless We could still add arguments to widget constructors (`left`, `midle`, `right` for align and a vararg `...` for fixed/flex. That's a nice idea, too.

psychon:
@Elv13 I would propose that instead of trying some function names (`set_widget`, `add`, `set_first`), instead we require e.g. a function `_accept_widget` on all widgets. This would be more work to implement (all "widget containers" need to be modified to add such a function), but it might be more future-proof. The same would apply to `add_element`. (Or both functions could do `local f = self._accept_widget or self.set_widget or self.add` and then in the align layout, `_accept_widget` would set the next `nil`-widget that it contains to its argument). The general idea being: `base` shouldn't know about individual widget's details.

psychon:
Besides the above, I like this: :+1:

psychon:
> I would propose that instead of trying some function names (set_widget, add, set_first), instead we require e.g. a function _accept_widget on all widgets. [...]

Another idea would be to pass the constructed widgets to `layout` as arguments (based on #490). E.g.:
```lua
local unpack = unpack or table.unpack
local function drill(content)
  if not content or content.is_widget then return content end
  local children = {}
  for k=1, #content do -- TODO: Make this always work with nils
    local v = content[k]
    if v then
      table.insert(children, drill(v))
    end
  end
  return content.layout(unpack(children))
end
```
This would remove the possibility to pass in a pre-constructed layout, but from the layout's point of view it would be simpler to implement, I guess. Oh and perhaps this makes it simpler to use own layouts, because the layout function is mostly "just a callback function".

Elv13:
> This would remove the possibility to pass in a pre-constructed layout

This is necessary, some "layout"s require to get/set some attributes that are not (yet) part of the constructor. Having layouts built in the array itself is also required for thinks like `background` that are useless without parameters. Finally, being able to insert an exiting layout make it possible to add/remove/change it later. Only with the declarative array, you have no pointer to the layout object, so it is "static", limiting its usefulness.

psychon:
True. Hm... perhaps if it's a function, then call it with the widgets and if not then assume it's a layout with "some suitable setter"? I don't know and will just leave this to you.

psychon:
Instead of sleeping, I had an idea yesterday evening. We could avoid the `nil`-issue by using functions. Something like the following filters `nil` (and `false`) values:
```lua
-- Filter nil values
local function f(a, ...)
  if a then
    return a, f(...)
  end
  if select("#", ...) == 0 then
    return
  end
  return f(...)
end
```
Based on this, we can use https://github.com/awesomeWM/awesome/pull/490.
For example, in the current default config this would look like this:
```lua
local l, w = wibox.layout, wibox.widget
local align, fixed, flex = l.align, l.fixed, l.flex
mywibox[s]:set_widget(align.horizontal(
  fixed.horizontal(
    mylauncher,
    mytaglist[s],
    mypromptbox[s]
  ),
  mytasklist[s],
  fixed.horizontal(f(
    mykeyboardlayout,
    s == 1 and w.systray(),
    mytextclock,
    mylayoutbox[s]
  ))
))
```
And for the titlebar:
```lua
local function center(w)
  w:set_align("center")
  return w
end
local function set_buttons(w)
  w:buttons(buttons)
  return w
end
local tw = awful.titlebar.widget
awful.titlebar(c):set_widget(align.horizontal(
  fixed.horizontal(
    set_buttons(tw.iconwidget(c))
  ),
  set_buttons(flex.horizontal(
    center(tw.titlewidget(c)
  )),
  fixed.horizontal(
    tw.floatingbutton(c),
    tw.maximizedbutton(c),
    tw.stickybutton(c),
    tw.ontopbutton(c),
    tw.closebutton(c),
  )
))
```
Does that look declarative? Should we use this in the default config?
(`set_buttons` is used in this weird way because that is what the current default config does. That should be fixed...)

Elv13:
That look fine to me. Of those 2 syntax, I guess, it come down to taste. I still don't think mine as a `nil` issue, as I use it in my (very complex) config since years without issue with both luajit 5.1 on desktop and lua5.2 on laptop. While not really documented, all lua implementations seem to behave correctly.

Also, I am not sure how the `f` function would look in a real implementation and how much it would make it look strange and (?) unintuitive.

psychon:
Another idea about the `f` function: It only exists for the systray, so perhaps we should avoid it all together via `local empty_widget = wibox.widget.base.make_widget()` and `s == 1 and w.systray() or empty_widget`.

Elv13:
So...

I just (finally) got to "Port all Awesome modules to the new widget system" part of my TODO list. Radical was a pita to port, but I made it (well, almost**,***, and it somehow is slooooww*). The new widget system is cool and will allow me to drop quite a lot of hacks and eventually make Radical simple enough to people who happen not to be me can understand the code... One thing that would help is an implementation of this ticket.

I tried the constructor syntax. It is an improvement, but as `Radical` has a lot of optional component, the "optional widget" part become an issue again. Optional widget are really an important part when it come to more complex GUI layout description. `Qt`, `Glade` and `Android` all have some (different) way to handle this.

I still suggest the approach proposed in this PR is the cleanest way to handle all types of corner case and can be extended with new features easier than the constructor one. Currently, the `layout` table element is the only type of metadata passed to the declaration, but other could be added eventually. Do you want me to ping the Lua list to ask about the possible side effects of the use of spares tables****?

* I will profile it, it only happen in sub-menus, I don't know why yet
** The new widget system doesn't allow widget overlay, I will submit a PR to fix this soon
*** As expected, the `powerline` and `shadows` effect are buggy (but sort of sometime working), I am still investigating my options
**** That's how they are called in the mailing list threads I found about them

psychon:
Here is my version of this idea: https://gist.github.com/psychon/7a75bc258de37091f7e5
What I like more: Things like `local f = self.set_widget or self.add or self.set_first` and `local f = layout.set_widget or layout.add or layout[align_elem[offset]]` aren't needed and widgets don't get these functions "forced upon them".
What you will like less: `nil` breaks for sure and the `.layout`-key has to be callable (= it cannot be a pre-constructed layout). One can work around the later by wrapping a layout with a function, but that might be a bit ugly (=`function(...) existing_layout:add(...) return existing_layout end`). The former can be worked-around via `nil or wibox.widget.base.make_widget()` to replace the nil with an empty widget.

I'm not sure if having this functionality in `check_widget` is the best approach, but from a pragmatic POV it makes sense.

How ugly do you think this is?

Elv13:
The syntax is fine and the code is sound. However, it doesn't solve all of the issues I care about. I really don't want to sound NIH, it really isn't about that. I am fully aware and acknowledge the way my patch work is a duct tape grade hack.

###### Extensibility 
 The `layout` element of the declaration is only the first of many possible metadata that could be attached to portion of the widget tree declaration.

For example, lets imagine the final version of the dynamic client layout system I have in my fork. An extended version of Tyrannical could generate layouts very easily:

```lua
-- Unpractical example of a declarative client layout
mytag.layout = {
    name = "my_magic_layout",
    {
        { -- Master
            max_client = 1, -- Maximum number of master client before pushing
                            -- them to the tiled section
            default_destination = true, -- Add new clients to master by default
            layout = wibox.layout.flex.vertical()
        },
        { -- Slave
            tyrranical.spawnable {
                -- Spawn a client when the client is first selected in the slave section
                -- This is already supported using the startup_id handler since 3.5.3
                command = "urxvt",
                border_color = "#unicorn_purple",
                spawn = "on_first_select" -- Or "exec_once", "on_init"
            },
            max_client = 5,
            class = "column", -- Declare this as a fragment that could be repeated
                              -- (somehow)
            layout = wibox.layout.flex.vertical()
        },
        layout = wibox.layout.flex.horizontal
    }
}
```
Of course, that syntax suck, I will probably find a nicer one, but you can see the concept of allowing handler for weird metadata. While this PR doesn't support adding handler, it can be extended to do so.

Another example would be for modules like vicious

```lua
mywibox.widgets = {
    taglist[s],
    nil,
    {
       { -- Vicious could catch that and use both widget to display the CPU usage
            wibox.widget.textbox(),
            wibox.widget.graph(),
            vicious = vicious.probe.cpu {
                refresh    = 2.0, --seconds
                formatting = "$1 %"
            }
        },
        mytextclock
    },
}
```
vicious

###### Layout as functions only
This wont work, I can't port my config to that new system without hacks and I wont be alone. Wrapping everything in a function is not very practical.

The `wibox.widget.background` is a simple example of a container that can't be instantiated as a function. One limitation of this style of syntax is that individual widgets are not easily accessible. As long as you are able to also use a declarative syntax to declare them, it is usually not an issue, but if you can't, it is a dead end. Either creating the container ("layout"s like widgets like background or margin) outside of the declarative tree or directly in them + constructor arguments.

With some containers, it will also look rather ugly

```lua
...
{
    mywidget,
    2, -- left
    2, -- right
    2, -- top
    2, -- bottom
    "pink", -- color
    layout = wibox.layout.margin
}
...
```

instead of:

```lua
...
{
    mywidget,
    layout = wibox.layout.margin (
        nil,
        2     , -- left
        2     , -- right
        2     , -- top
        2     , -- bottom
        "pink", -- color
    )
}
...
```

Of course, the it could also be written as
```lua
...
{
    wibox.layout.margin (
        mywidget,,
        2     , -- left
        2     , -- right
        2     , -- top
        2     , -- bottom
        "pink", -- color
    )
}
...
```

But this would be quite confusing and would not work for all containers. I think, IMO, being able to use already constructed and configured layout is still required for many corner case.

###### How to fix it?

Maybe we could just make the "add(...)" method mandatory for all widgets and be done with it? That way, an hybrid implementation between yours and mine could be created.

Instead of
```
-    for k, func in pairs({ "add_signal", "connect_signal", "disconnect_signal" }) do
-        assert(type(widget[func]) == "function", func .. " is not a function")
+    if widget.is_widget then
+        for k, func in pairs({ "add_signal", "connect_signal", "disconnect_signal" }) do
+            assert(type(widget[func]) == "function", func .. " is not a function")
+        end
+        return widget
+    else
+        return base.check_widget(widget.layout(unpack(widget)))
     end
```

It could be 
```
-    for k, func in pairs({ "add_signal", "connect_signal", "disconnect_signal" }) do
-        assert(type(widget[func]) == "function", func .. " is not a function")
+    if widget.is_widget then
+        for k, func in pairs({ "add_signal", "connect_signal", "disconnect_signal" }) do
+            assert(type(widget[func]) == "function", func .. " is not a function")
+        end
+        return widget
+    else
+        local l = type(widget.layout) == "function" and widget.layout() or widget.layout
+        if base.check_widget(l) then
+            l:add(unpack(widget))
+            return l
+        end
+        assert("dsfdsfdsfsdf")
+        -- Some more code to support metadata handlers and
+        -- "layoutless" {} (see the spawn example above)
     end
```

psychon:
Ok. So first let's add a new function for doing this instead of hacking this into `check_widget()`. Perhaps `instantiate_widget()`? That function could then look like this:

    function instantiate_widget(arg)
        assert(type(arg) == "table")
        if arg.is_widget then
            return arg
        end
        local l = arg.layout
        if type(l) == "function" then
            return l(unpack(arg))
        end
        assert(type(l.instantiate_with) == "function", "The given layout does not support on-the-fly-instantiation") -- TODO: Better message
        return l:instantiate_with(arg, unpack(arg))
    end

If it is called with a widget, it just returns its argument.
Otherwise, it assumes that this is for the declarative syntax and looks at the `.layout` key of its argument.
If that is a function, it just calls that function with the argument unpacked (this works e.g. for the fixed and flex layouts).
Otherwise, it assumes that this has an `:instantiate_with()` function that does everything else. For a fixed/flex layout, this function could just be `function foo:instantiate_with(_, ...) self:add(...) end`. The idea being that the first argument is some metadata and the rest is the widgets that should be added.

However, I don't know how vicious or `tyrranical.spawnable` could catch anything here. (Overwrite the function? If so, we could add a public function somewhere that in the default implementation just returns its arguments again. That could then be called in the right place and could be overwritten by modules.)

Elv13:
I am too lazy to update the PR, but here is a new version
https://gist.github.com/Elv13/997513f175e64b931473

Changes:

 * Always work with sparse tables, do not assume clua/luajit implementation
 * Allow attributes to be set
 * Allow custom widge creation directly
 * Add `id` property to create reference widget by name, QML style
 * Alias `widget` to `layout` as it now work with normal widget too

The end result now look like this:

```lua
    --------------------------------------------------
    --                 Top titlebar                 --
    --------------------------------------------------

    local title_layout = {
        {
            {
                beautiful.titlebar_show_icon and wibox.widget.imagebox(c.icon) or nil,
                title,
            },
            left   = 10                 ,
            right  = 10                 ,
            layout = wibox.layout.margin,
        },
        id                 = "title_bg"                                  ,
        shape              = beautiful.titlebar_title_shape              ,
        shape_border_width = beautiful.titlebar_title_border_width       ,
        shape_border_color = beautiful.titlebar_title_border_color_active,
        bg                 = beautiful.titlebar_title_bg                 ,
        buttons            = (align ~= "center") and buttons             ,
        layout             = wibox.widget.background                     ,
    }

    tb:set_widgets {
        { -- Left
            beautiful.titlebar_side_top_left and wibox.widget.imagebox(beautiful.titlebar_side_top_left),
            {
                {
                    resize_handle,
                    tag_selector ,
                },
                id     = "left_button_bg"       ,
                bg     = alt_color              ,
                layout = wibox.widget.background,
            },
            endArrow_alt,
            (not align or align == "left") and title_layout or nil,
        },
        { -- Middle
            nil,
            (align  == "center") and title_layout or nil,
            (align  == "center") and underlay_w   or nil,
            id      = "middle_section"               ,
            expand  = align == "center" and "outside",
            buttons = buttons                        ,
            layout  = wibox.layout.align.horizontal  ,
        },
        { -- Right
            align ~= "center" and underlay_w or nil,
            endArrowR2,
            {
                right_layout                    ,
                id     = "title_bg"             ,
                bg     = alt_color              ,
                layout = wibox.widget.background,
            },
            beautiful.titlebar_side_top_right and wibox.widget.imagebox(beautiful.titlebar_side_top_right),
        },
        expand = "inside"                     ,
        id     = "main_layout"                ,
    }

    set_underlay(c,underlay_w)

    -- Now, the bottom one (floating clients only)
    if awful.client.floating.get(c) or beautiful.titlebar_bottom then

        local tb = awful.titlebar(c,{size= beautiful.titlebar_bottom_height or 5,position="bottom"})

        tb:set_widgets {
            {
                draw    = draw_bottom_left             ,
                fit     = bottom_corner_fit            ,
                widget  = wibox.widget.base.make_widget,
                buttons = awful.util.table.join(
                    awful.button({ }, 1, function(geometry)
                        collision._resize.mouse_resize(c,"bl")
                end))
            },
            {
                draw    = beautiful.titlebar_bottom_draw or draw_3_dots,
                fit     = function(self,w,h)
                    return w,h
                end                                    ,
                widget  = wibox.widget.base.make_widget,
                buttons = awful.util.table.join(
                    awful.button({ }, 1, function(geometry)
                        c:raise()
                        collision._resize.mouse_resize(c,"bl")
                end))
            },
            {
                draw    = draw_bottom_right            ,
                fit     = bottom_corner_fit            ,
                widget  = wibox.widget.base.make_widget,
                buttons = awful.util.table.join(
                    awful.button({ }, 1, function(geometry)
                        collision._resize.mouse_resize(c,"br")
                end))
            },
            id     = "main_layout",
        }
```

If I implemented `set_widgets` as I implemented `get_widgets` in the other PR, I could drop the add/set_first/set_widget hack and all your concerns would be addressed.

What do you think?

actionless:
i like how it looks now

Elv13:
Rebased on top of my `upstream_dynamic_layout_p5` branch (hopefully soon to be merged). Fix the last 2 remaining @psychon complains.

(for review, ignore the commits from `upstream_dynamic_layout_p5`, there is only relevant 4 commits)

Elv13:
I removed the extra commit, now sit on `master`

lib/awful/titlebar.lua
```diff
@@ -26,6 +26,37 @@ local titlebar = {
...
+--         my_second_widget,
+--         { -- Add a background color/pattern for my_third_widget
+--               my_third_widget,
+--               layout = wibox.widget.background(nil,beautiful.bg_focus),
+--         },
+--         layout = wibox.layout.fixed.horizontal,
+--     }
+--
+-- The `layout` array member can either be the name of the layout or a layout
+-- object itself. This allow additional parameters be passed to the layout. It
```
psychon: "This allows additional parameters *to* be passed to the layout."

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,133 @@ function base.place_widget_at(widget, x, y, width, height)
...
+-- @usage local mylayout = wibox.widget.background()
+--    mylayout:set_bg("#ff0000")
+--    
+--    mywibox:set_widgets {
+--        mywidget,
+--        layout = mylayout
+--    }
+--
+function base.widget:setup(args)
+    local f,ids = self.set_widget or self.add or self.set_first,{}
```
psychon: Could you write some basic unit tests for this?
You add a `:set_children()` method to all widgets and then you do not use it?

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,133 @@ function base.place_widget_at(widget, x, y, width, height)
...
+end
+
+-- Recursively build a container from a declarative table
+local function drill(ids, content)
+    if not content then return end
+
+    -- Alias `widget` to `layout` as they are handled the same way
+    content.layout = content.layout or content.widget
+
+    -- Make sure  is not called on a function
```
psychon: Feels like some word is missing here

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,133 @@ function base.place_widget_at(widget, x, y, width, height)
...
+    local layout = type(content.layout) == "function" and  content.layout() or content.layout
+
+    -- Create layouts based on metatable __call
+    local l = layout.is_widget and layout or layout()
```
psychon: If layout is a function that returns something which has a non-true `is_widget` key, it will be called again? Seems weird to me, why?
Is this just so that both "things with `__call`" and "functions" work?

lib/awful/titlebar.lua
Elv13: Right, I also need to document the new features such as properties (see example above), new widget creation and widget identifiers.

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,133 @@ function base.place_widget_at(widget, x, y, width, height)
...
+            if not v.is_widget then
+                e, id2 = drill(ids, v)
+                widgets[k] = e
+            end
+
+            -- Place the widget in the access table
+            if id2 then
+                l  [id2] = e
+                ids[id2] = e
+                id2      = nil
```
psychon: Why unset `id2`? That doesn't really matter.

Also, you are adding random keys to random widgets (the `l[id2] = e`-part). If I give a widget the id `add`, `draw` or something like that, stuff will break.
What's the point of this `id` stuff (I don't see it documented, I think). If users should use that, wouldn't it be enough to just return the `ids` table (that's already done now) and people have to save that somewhere themselves?

Oh and all of this might break if a widget is shown in more than one place, but I guess I can live with that (it needs something made on purpose to break it)....

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,133 @@ function base.place_widget_at(widget, x, y, width, height)
...
+            end
+        end
+    end
+
+    -- Drawable don't have any set_children and it would make no sense, this
+    -- make an exeption for them
+    local f = l.set_children or l.set_widget
+    f(l, widgets)
```
psychon: This doesn't work, does it? It will do the equivalent of `drawable:set_widget({widget})`, but only `drawable:set_widget(widget)` works.

lib/wibox/widget/base.lua
Elv13: `is_widget` was proposed by you in the first revision. All widgets set it to `true` (see below). This allow:

* existing layouts to be used (they will have the `is_widget` property)
* `__call` to somehow work
* New layouts to be created

If something else happen, let it misbehave, /me blame the user. There is very little case where this wont work. Something like `widget = "potato"` or variations `local f; f=function() f() end; ...; layout = f` and some other, but I don't think we should care about those.

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,133 @@ function base.place_widget_at(widget, x, y, width, height)
...
+
+    -- Set layouts attributes
+    for attr, val in pairs(attributes) do
+        if l["set_"..attr] then
+            l["set_"..attr](l, val)
+        elseif type(l[attr]) == "function" then
+            l[attr](l, val)
+        else
+            l[attr] = val
+        end
```
psychon: Can we get rid of the last "else"-part? Again, this sets random fields on random widgets. That's not really any kind of API and could break stuff in weird ways.

I'd vote for:
```lua
for attr, val in pairs(attributes) do
    local f = l["set_" .. attr] or l[attr]
    if type(f) == "function" then
        f(l, val)
    else
        error("Don't know how to set property " .. tostring(attr) .. " on widget " .. tostring(l) .. (id and "with id " .. tostring(id) or ""))
    end
end
```

lib/wibox/widget/base.lua
Elv13: I will document it (I did not update the doc from the last version, I wanted to push the logic first so I see if you feel better about this design)

It allow javascript/QML like `mywibox.acontainer.anothercontainer.awidget:set_text("foo")`. One of the limitation (documented in the comments above) of the previous patch was it was write only. Once you created the tree, you had no pointers to the widgets you just created. This `identifier` feature allow both JS like `get_element_by_id()` and QML like `local w = widget1.widget2.widget2` style access. It will indeed cause issues if reserved names are used. I think documenting this is enough, unless you have another elaborate idea?

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,133 @@ function base.place_widget_at(widget, x, y, width, height)
...
+--        mywidget,
+--        layout = mylayout
+--    }
+--
+function base.widget:setup(args)
+    local f,ids = self.set_widget or self.add or self.set_first,{}
+    local w, id = drill(ids, args)
+    f(self,w)
+    if id then
+        self[id] = w
```
psychon: This also sets random fields on random widgets, I don't like that.

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,133 @@ function base.place_widget_at(widget, x, y, width, height)
...
+    end
+    self.by_id = ids
```
psychon: Could this be called `widget_by_id`? This also sets a random field on a random widget, but since the key is constant, I guess I can live with that.

lib/wibox/widget/base.lua
Elv13: ... umm, somehow it does work, but it should not, the screenshot I made in the other PR use it for the whole config... Good catch (/me wonder why it works, because it does)

lib/wibox/layout/fixed.lua
```diff
@@ -110,6 +110,14 @@ function fixed:get_children()
     return self.widgets
 end
 
+--- Replace the layout children
+-- @tparam table children A table composed of valid widgets
+function fixed:set_children(children)
+    if not children then return self:reset() end
```
psychon: This check isn't needed here nor in any of the other implementations of `:set_children()`. The API documentation says this must be a table, and the implementation of `setup()` below only calls this with a table.

lib/wibox/widget/base.lua
Elv13: It doesn't set random thing, this allow widgets to be created

```lua
{
   draw = my_draw_function,
   fit      = my_fit_function,
   widget = wibox.widget.base.make_widget,
}
```

And random properties to be set (`Radical` use that)

lib/wibox/widget/base.lua
Elv13: You can already do `mywidget.foobar = "chaos"`, this is no different (and extensions can make use of it)

awesomerc.lua
```diff
@@ -195,27 +195,24 @@ for s = 1, screen.count() do
...
+        },
+        mytasklist[s], -- Middle widget
+        { -- Right widgets
+            layout = wibox.layout.fixed.horizontal,
+            mykeyboardlayout,
+            s == 1 and wibox.widget.systray(),
+            mytextclock,
+            mylayoutbox[s],
+        },
+    }
```
psychon: Why not also the titlebar code below? I guess that would also be more interesting since it does weird stuff with `:button()` (it wants to bind "move client" to the whole titlebar, except for the buttons (close, maximize etc)).

Elv13: Oops, I had it in the previous version but forgot to copy/paste it.

lib/wibox/widget/base.lua
psychon: Prefix the IDs with something? So then it becomes `widget1.child_by_id_widget2.child_by_id_widget2`? We only need some less ugly prefix.

Alternatively: `widget1.child_by_id.widget2.child_by_id.widget2` (`child_by_id` is a table instead of a prefix). Actually, I think this is much nicer since only a single entry will only ever be set on random widgets.
So... name.. is `by_id` short enough?

lib/wibox/widget/base.lua
psychon: See my edit: This code here only ever sees widgets, drawables only appear in `:setup()`.

lib/wibox/widget/base.lua
psychon: Ok, nice magic... How likely is it that some user will shoot into their foot with this? We either allow this directly or do something like the following:
```lua
{
  widget = wibox.widget.base.make_widget,
  set_fields = {
    draw = draw_function,
    fit = fit_function
  }
}
```
The implementation is just a `awful.util.crush` away (except that you cannot require("awful.util") from inside of wibox, meh...) and users can't set random junk easily.

lib/wibox/widget/base.lua
psychon: Uhm, now that I read this all... The `id` is only available as `widget.foo` on the direct parent, while `widget.by_id.foo` "looks" for the id anywhere in the hierarchy, right?

lib/wibox/widget/base.lua
Elv13: Again, the user can already set random junks, setting random junks is like the whole point of this. See my `spawn` example for the client layout system above. Also, many widgets do have "random" properties already (like `visible` or `data` or `_uli_secret_api`), that `else` allow them to be used. I prefer to let that flexibility (and self-foot-shooting potential) as it is.

lib/wibox/widget/base.lua
Elv13: Right, this is on purpose. As I said in some comments above, one emulate QML and the other JavaScript DOM. Both style have their advantages. `by_id` return a single element, so if the same ID is used twice, it is lost (I could change that by having a "real" get_element_by_id implementation) while the other is more verbose, but you never miss, you will get the right widget the first time as opposed to the (well known) corner case of the DOM API (and the reason jquery was invented)

psychon: Ok, thanks. Seems like I at least understood some part. :-)

lib/wibox/layout/fixed.lua
Elv13: I am unsure about this. `set_children` is the opposite of `get_children` and it return a table, not an `unpack`. It would rather inconsistent. As for the check, it doesn't hurt. I am just being nice for users, no?

Elv13:
Updated

@psychon Sorry for my stubbornness and again ignoring some of your comments, but I wrote an extensive documentation about why everything is exactly the way it is. Let see if you agree

@blueyed I added a commit to enable the documentation have carry images. It work locally, I hope it work on Github too. Having the images in the repository allow users to access it without internet. Most projects do that, but they add up and bloat the repository size. Hopefully we wont ever have enough to actually cause issues.

awesomerc.lua
Elv13: Done

lib/wibox/layout/fixed.lua
psychon: I agree with the inconsistency... Perhaps make `get_children()` return them as a "list", too? (Callers could get a table via `{ foo:get_children() }`)
This would again simplify the implementation in most widgets and fixed can use `unpack()`, but this is quite a minor simplification, so I'm not sure if this makes a lot of sense. Feel free to say "No!".

About "I am just being nice for users": Well, is it really nice? Assuming the user meant to do something and e.g. just had a typo in the variable name, the code will silently do nothing instead of causing a Lua error whose backtrace hopefully points at the typo. I'm not sure if "being nice" is always the right thing.

Elv13: Sorry, but

>  "No!".

I think it make the API worst and given (my) the use case of `get_children`, it make no sense

> Well, is it really nice?

I think so, I see myself wanting this sort of thing to make my life easier. It does what it claim to do, the the children "to nothing"

lib/wibox/layout/margin.lua
```diff
@@ -78,6 +78,14 @@ function margin:get_children()
     return {self.widget}
 end
 
+--- Replace the layout children
+-- This layout only accept one children, all others will be ignored
+-- @tparam table children A table composed of valid widgets
+function margin:set_children(children)
+    self.widget = children and children[1]
```
psychon: Should all of these functions use `:check_widget()`? Actually... these could just use `:set_widget(children and children[1])` and then there is only one function doing all of this. Similarly, `fixed` could use `self:reset() self:add(unpack(children or {}))`.

Am I micro-optimizing for no real gain? (Does `check_widget()` make sense or am I just scared from when I started this widget stuff and had tons of errors while drawing, because invalid widgets were set?)

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+        else
+            l[attr] = val
+        end
+    end
+
+    return l, id
+end
+
+-- Only available when the declarative system is used
+local function get_children_by_id(self, name)
```
psychon: `get_child_by_id`, or does this return a table?

lib/wibox/layout/margin.lua
Elv13: This is a good point. I asked myself the same thing (and also while writing the client layout system). After thinking about it, I decided that no, it isn't worth it. My decision is debatable. `check_widget` does show up when profiling, but it isn't a bottleneck. In the end it's just that the actual implementation of `check_widget` is actually not that useful (IMO) and doesn't really enforce anything beside basic "yes, this should be a widget, at least I hope so". You can still make invalid widgets that pass the test.

I added an `is_widget` below. If this is to true and it isn't a widget, then I blame the user.

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+--        },
+--        layout = wibox.layout.fixed.horizontal,
+--    }
+--    
+--
+-- Result:
+-- ![Example2 screenshot](../../docs/images/widgetlayout1.png)
+--
+--
+-- **Example 3 :** Use an `wibox.layout.align` latout
```
psychon: latout -> layout

lib/wibox/widget/base.lua
Elv13: It return a table of children with the `id` identifier or nil

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+-- * [Pattern](http://cairographics.org/manual/cairo-cairo-pattern-t.html)
+-- * [transformation](http://cairographics.org/manual/cairo-Transformations.html)
+-- * [Operator](http://cairographics.org/operators/)
+-- * [Pango text](https://developer.gnome.org/pango/stable/)
+--
+--
+-- **Example 5 :** Externally defined widgets and layouts
+--
+-- The declarative system can also use external widgets. This is useful when the
+-- widget is provided by an external module or require complex manupilations
```
psychon: manipulations

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+-- would make the declaration unreadable.
```
psychon: Actually `This is useful when the widget is provided by an external module or *when* *it* require*s* complex manipulations *which* would make the declaration unreadable` (but I'm not totally sure)

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+-- For each widget or container, it is possible to add an `identifier` attribute
+-- so the widget can be accessed later.
+--
+-- Widgets defined using `setup` can be access by 3 means:
+--
+-- * Avoid the issue by using externally created widgets
+-- * Use `my_wibox.my_first_widget.my_second_widget` style access
+-- * Use JavaScript like `my_wibox:get_children_by_id("my_second_widget")[1]`
+--
+-- The first method mix the imperative and declarative syntax, but make the code
```
psychon: "The first method mix*es* [...] but make*s* the code"

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+-- less readable. The second is a little verbose and only work if every node in
```
psychon: work*s*

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+-- the chain have a valid identifier. The last one doesn't require long paths,
+-- but it is not easy to get a specific instance if multiple widgets have the
+-- same identifier.
+--
+-- WARNING: The widget identifier must not use reseved name. This include all
+-- method names, existing widget attributes, `layout` and `widget`. Names should
+-- also respect the lua variable name policies (cape sensitive, alphanumeric and
```
psychon: ca*s*e

lib/wibox/widget/base.lua
Elv13: 2 of Elv13 weakness

 * Making stupid mistakes when updating PR at 4 AM
 * Being totally reliant on spellcheck and not having one in its editor

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+-- catched by a lua `metatable` (operator overload).
+--
+-- Code:
+--    
+--    -- "Monkeypatch" a new function to wibox.widget.textbox to add vicious
+--    -- extension support
+--    function wibox.widget.textbox.vicious(w, args)
+--        local f = unpack or table.unpack -- Lua 5.1 compat
+--        vicious.register(w, f(args))
+--    end
```
psychon: Magic! :-)
Could you make this:
```lua
function wibox.widget.textbox:vicious(args)
    local f = unpack or table.unpack -- Lua 5.1 compat
    vicious.register(self, f(args))
end
```

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+--            widget  = wibox.widget.textbox
+--        },
+--        layout = wibox.layout.fixed.horizontal,
+--    }
+--    
+--
+-- In this example, the system is extended to that the popular
+-- [Vicious](http://awesome.naquadah.org/wiki/Vicious) extension module can be
+-- used directly in the layout declaration. This example will update the textbox
+-- every 3 seconds to set the CPU usage.
```
psychon: "to show the CPU usage"? (I'm unsure)

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+--
+--
+-- **Example 8 :** Handling sections
+--
+-- The system is allow sections to be defined externally, then composed into
```
psychon: "The system [is (remove this)] allow*s*"

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+--    mywibox[s] : setup (three_circle)
+--    
+--
+-- @param args An array containing the widgets disposition
+function base.widget:setup(args)
+    local f,ids = self.set_widget or self.add or self.set_first,{}
+    local w, id = drill(ids, args)
+    f(self,w)
+    if id then
+        -- Avoid being dropped my wibox metatable -> drawin
```
psychon: *by*

lib/wibox/widget/base.lua
```diff
@@ -321,6 +330,372 @@ function base.place_widget_at(widget, x, y, width, height)
...
+--    local three_circle = {layout = wibox.layout.align.horizontal}
+--    for i=1, 3 do
+--        table.insert(three_circle, circle)
+--    end
+--    
+--    mywibox[s] : setup (three_circle)
+--    
+--
+-- @param args An array containing the widgets disposition
+function base.widget:setup(args)
```
psychon: Thanks for these extensive docs. I like them!

lib/wibox/init.lua
```diff
@@ -36,6 +37,272 @@ function wibox:set_widget(widget)
     self._drawable:set_widget(widget)
 end
 
+--- Set a declarative widget hierarchy description.
```
psychon: Since this (and the titlebar version) is duplicate... Does it make sense to have only a single version and link to it? Or can LDoc perhaps include docs from somewhere?

Having multiple versions of this long doc comment is a guarantee for them to get out of sync.

awesomerc.lua
```diff
@@ -482,45 +479,42 @@ client.connect_signal("manage", function (c)
...
+            },
+            { -- Right
+                awful.titlebar.widget.floatingbutton (c),
+                awful.titlebar.widget.maximizedbutton(c),
+                awful.titlebar.widget.stickybutton   (c),
+                awful.titlebar.widget.ontopbutton    (c),
+                awful.titlebar.widget.closebutton    (c),
+                layout = wibox.layout.fixed.horizontal()
+            },
+            layout = wibox.layout.align.horizontal
```
psychon: Wow. I like!

lib/wibox/layout/margin.lua
psychon: So you are saying that (a) `check_widget()` should check if `.is_widget` is true (nice idea, you should add that :-) ) and (b) This doesn't count as public API and users shouldn't call it, right?

lib/wibox/widget/base.lua
psychon: Yeah, noticed that while reading the docs. Should it perhaps return an empty table for invalid IDs? That would get rid of a special case.

lib/wibox/init.lua
Elv13: Yes, I think I should make a `.md` like the user contributor guide. I could also move `awful.rule` to one too. They are very high level, very general features users are much more likely to use compared to a deep API doc page. What do you think (and what do @blueyed think)?

lib/wibox/widget/base.lua
psychon: No problem. Thanks a lot for writing all of this. (And I'm often ashamed when reading what I wrote in a commit message...)

lib/wibox/layout/margin.lua
Elv13: Right. Again, this is debatable and I have no killer argument. `check_widget` is useful in some "hot spot" in the code where all the magic may cause an apocalypse, but it is overkill to just use it everywhere.

lib/wibox/widget/base.lua
Elv13: Ok

lib/wibox/widget/base.lua
Elv13: I hope someone will read it eventually, that took a while to write ;)

But it is also the base of the unit test I will add. This (hopefully) cover all case, and all the examples are copy-pastable into the default `rc.lua` to replace `mywibox[s]:set_widget()`, so a user can try them right away (onlt the first one is quite useless, it the widgets are `nil`, I removed them as it made the example too big)

awesomerc.lua
Elv13: To give credit to @jd where it is due
https://github.com/awesomeWM/awesome/blob/3.4/awesomerc.lua.in#L170
https://github.com/awesomeWM/awesome/blob/3.4/lib/awful/titlebar.lua.in#L126

psychon: Yeah, but the button and align magic is due to you :-)

Elv13: Now you understand was I did not like the `contructor(...)` idea. In itself it did the same thing as the original version of the patch, but I had this sort of awesomeness in mind.

I should not have been lazy and implemented it all the first time, this would have avoided a long and time consuming debate. Sorry about that.

psychon: Sorry for debating so longly and time consumingly ;-)

Elv13:
Updated

psychon:
It's good enough. I might mess with some details after this is merged and I won't convince to stop setting random entries on random widgets anyway, so: :+1:

Elv13:
The future is now!

lib/awful/titlebar.lua
```diff
@@ -108,6 +114,9 @@ local function new(c, args)
     -- Make sure the titlebar has the right colors applied
     bars[position].update_colors()
 
+    -- Handle declarative/recursive widget container
+    ret.setup = base.widget.setup
```
psychon: lib/awful/titlebar.lua:117:17: accessing undefined variable base

(I think the default config should enable titlebars, not disable them; that way travis would have caught this...)

PR #489, opened 2015-09-28
Move util.spawn to a new module, add ability to spawn with properties

blueyed:
* This commit add a new module to avoid a (4 level) loop dependency
* It is now possible to call awful.spawn() with a table of properties
* awful.rules is used to execute the rules.

* Everything is public to allow alternative workflow modules such as
    Tyrannical to use their own callback implementation.

Successor of #34 to make Travis pick it up.


Elv13:
The typo has been fixed

Elv13:
Ok, I fixed Travis (forgot to include LGI in spawn.lua, as required by the new async method)

blueyed:
I agree with @nuko8 that the default config should make use of it.

@Elv13 
Do you feel like doing this?  Otherwise I could add it, but then would have to adopt the PR or move it over to the orga's repo.

PR #495, opened 2015-10-03
Improve test runner to be usable for git-bisect

blueyed:
It might be nice if it would not have to be in a subdir, e.g. by looking for `awesomerc.lua` / `awesomerc.lua.in`?!


psychon:
One inline comment (why start awesome before checking if the file to run is readable?), besides that this is fine by me

PR #502, opened 2015-10-08
Remember mouse position per screen

### LINKED ISSUE #478: Remember mouse position per screen
The (relative) screen position of the mouse on a screen should be remembered and restored when moving the focus to another screen.

Could be done for the following functions first / only / in the beginning:

    awful.key({ modkey, "Control" }, "j", function () awful.screen.focus_relative( 1) end),
    awful.key({ modkey, "Control" }, "k", function () awful.screen.focus_relative(-1) end),

Bew78LesellB:
The mouse position is saved and restored when focusing another screen

If the target screen has no saved mouse position, the relative position
of the current screen is used

This should close #478

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -24,7 +24,7 @@ local screen = {}
 local data = {}
 data.padding = {}
 
-mouse_per_screen = {}
+screen.mouse_per_screen = {}
 
 ---
 -- Return Xinerama screen number corresponding to the given (pixel) coordinates.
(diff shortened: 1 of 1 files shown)


Elv13:
This is a good addition to awful, the patch look sane to me. One other thing you usually want when doing this is moving the mouse when a client is moved from screen to screen so the mouse remain on the client (if it is focussed)

I guess you might also want to try my implementation of this, it does a few more things like drawing a red circle around the mouse cursor when it is automatically moved. This is overkill for the default implementation like yours, but I like it
https://github.com/elv13/collision

Bew78LesellB:
here you are, it should be fine now

psychon:
Don't you also have to modify all the uses of this table so that they have a `screen.` prefix?

Bew78LesellB:
aargh, how did I forgot that...

if I do the modifications, then a commit --amend, then a force push, will it break something ?

psychon:
Nah, feel free

Bew78LesellB:
okay, done !

blueyed:
Thanks!

Some nitpicking: I would have liked for this to be a single commit in master.

PR #503, opened 2015-10-08
Raise client in awful.client.movetoscreen

blueyed:
The intention of [#98] / fbc72624 was to actually raise the client,
but that was never the case apparently, and got totally lost in
[#441] / 57755b3).

[#98]: https://github.com/awesomeWM/awesome/pull/98
[#441]: https://github.com/awesomeWM/awesome/pull/441


psychon:
Fine with me, I guess

PR #506, opened 2015-10-10
Fix awful.util.spawn_with_shell wrapper

blueyed:


psychon:
While you are fixing this, I'd suggest to also add the missing `return` to both wrapper functions so that error messages / PIDs are propagated.

blueyed:
Done in #511 - noticed it only later.

blueyed:
I'll commit them all at once to master.

Issue #292, opened 2015-07-07
bad interaction between Dota 2 - Reborn and AwesomeWM


lewisdiamond:
See https://github.com/ValveSoftware/Dota-2-Reborn/issues/145 for details

vdrandom:
It is clearly a bug in Dota 2, not in Awesome WM. It also affects other window managers, I remember many people mentioned Fluxbox there.

psychon:
Uhm, I don't see what awesome would be doing wrong. (Obviously) I don't know steam's source code, but I guess they have a race where they assume that the window is managed by the WM before they try to set the input focus to this window. However, such an assumption is just always wrong.

Also, a non-reparenting WM could decide to move the steam window to a currently not selected workspace and ban the window by unmapping, in which case this error would also occur.

Finally, since this also happens with Fluxbox, what's the common bug?

benoit-pierre:
It's fixed in SDL master: https://bugzilla.libsdl.org/show_bug.cgi?id=2997

psychon:
In other news, awesome doesn't follow ICCCM 2.8. :-( (It doesn't do anything with the WM_Sn selection)

https://awesome.naquadah.org/bugs/index.php?do=details&task_id=782

PR #513, opened 2015-10-10
Wm sn selection

psychon:
Various selection stuff that ICCCM requires us to do.

Fixes: https://awesome.naquadah.org/bugs/index.php?do=details&task_id=782
Fixes: https://github.com/awesomeWM/awesome/issues/292


blueyed:
Merged with fixed punctuation for the man pages, and the French translation in 0478b94.

PR #518, opened 2015-10-11
Fix signature for recursive find_icon_path_helper

blueyed:
Ref: https://github.com/blueyed/awesome/commit/81072c042087b7dd7b214b44a7e559f09e0280f6#commitcomment-13705007.

This reverts a part of 81072c0.

### DIFF
--- a/lib/menubar/icon_theme.lua
+++ b/lib/menubar/icon_theme.lua
@@ -196,7 +196,8 @@ local lookup_icon = function(self, icon_name, icon_size)
     return nil
 end
 
-local find_icon_path_helper = function(self, icon_name, icon_size)
+local find_icon_path_helper  -- Gets called recursively.
+find_icon_path_helper = function(self, icon_name, icon_size)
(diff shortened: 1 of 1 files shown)


psychon:
I know I have written code like this myself, but I no longer know why. This kind of recursive call seems to work for me:
```
$ lua5.1 -e 'local function f(a) print(a) if a then f(false) end end ; f(1)'
1
false
$ lua5.2 -e 'local function f(a) print(a) if a then f(false) end end ; f(1)' 
1
false
```

nuko8:
According to http://www.lua.org/pil/6.2.html,
```
local function func(...)
    func(...)
end
```
is OK, but not if we write the above function definition like this:
```
local func = function(...)
    func(...)
end
```
Looks my preference to the `func = function()` style caused the confusion...

psychon:
Thanks for figuring this out. I was sure that there was a problem, but I could never figure out how it happened. :-)

PR #523, opened 2015-10-11
Some menubar / icon theme fixes

blueyed:
/cc @nuko8

### DIFF
--- a/lib/menubar/icon_theme.lua
+++ b/lib/menubar/icon_theme.lua
@@ -168,16 +168,18 @@ local lookup_icon = function(self, icon_name, icon_size)
     local minimal_size = 0xffffffff -- Any large number will do.
     local closest_filename = nil
     for _, subdir in ipairs(self.index_theme:get_subdirectories()) do
-        for _, basedir in ipairs(self.base_directories) do
-            for _, ext in ipairs(self.extensions) do
-                local filename = string.format("%s/%s/%s/%s.%s",
-                                               basedir, self.icon_theme_name, subdir,
-                                               icon_name, ext)
-                if not checked_already[filename] then
-                    local dist = directory_size_distance(self, subdir, icon_size)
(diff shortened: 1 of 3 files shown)


blueyed:
@nuko8 @psychon 
> I guess caching the index_theme instance in icon_theme could bring us a better result, if some optimization matters.

Makes sense.  For this it would be required for `gears.cache` to compare the table argument of directories by value.  Would that be possible/feasible?  It might still get garbage-collected then though, and therefore a non-weak cache should be used maybe?

nuko8:
Why is #525 cherry picked here?

Also, I'm wondering why #518 is left behind for days.

Is there some reasons not to merge them into master right now?

blueyed:
@nuko8 
> Why is #525 cherry picked here?

I've thought it fits the whole PR altogether, at least it fixes some issue discovered here, and it makes it easier to test without creating local merge branches.

But I see that it can/should go directly into master - done so now.

> Also, I'm wondering why #518 is left behind for days.

Merged by now.

blueyed:
Rebased on master, ready for merging from my side.

nuko8:
@blueyed 

Thank you for the explanation and the merge.   I understand the explanation.  If you gave it to me beforehand, I would definitely agree with you and could arrange something convenient for you.

Thanks to the merge, the convoluted non-linearity that was only seen in my local repo was resolved, which really makes my work with you easier (Remember, non-member can't merge PRs easily).  I'm happy with that.  Thank you!

blueyed:
@nuko8 
Cool!

> Remember, non-member can't merge PRs easily

JFI: You can always have a local branch with different stuff in it, and e.g. https://hub.github.com/ might help you with managing that.

nuko8:
@blueyed 

Thank you for the info.  I'll review my work flow and see how I can improve it with hub.

PR #526, opened 2015-10-13
Fix/handle focus history on startup

blueyed:
After fcccc77 the focus history is not handled correctly any more during restart.

This PR suggests two ways to address this.

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -1078,6 +1078,14 @@ capi.client.add_signal("marked")
 capi.client.add_signal("unmarked")
 
 capi.client.connect_signal("focus", client.focus.history.add)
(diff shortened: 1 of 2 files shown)


lib/awful/client.lua
```diff
@@ -1078,6 +1078,14 @@ capi.client.add_signal("marked")
 capi.client.add_signal("unmarked")
 
 capi.client.connect_signal("focus", client.focus.history.add)
+-- Add clients during startup to focus history.
+-- This used to happen through ewmh.activate, but that only handles visible
+-- clients now.
```
blueyed: This comment needs to be changed according to the outcome of this PR.

blueyed:
Done.

PR #530, opened 2015-10-14
Set the wallpaper more intelligently

### LINKED ISSUE #288: Wallpaper not properly painted after first restart (with changed screen offset)
When going to my docking station (with an external monitor) I need to manually
start awesome once more, to fix the wallpaper on the smaller / internal screen:
its offset and scaling is OK, but it does not cover the lower part, which is black.

After restarting once more, it gets painted properly.

It looks like an offset/height calculation issue: the internal screen gets an offset of 432 (when docked).

I am using `theme.wallpaper` from the xresources theme (which is a
`cairo.ImageSurface`), and set it using:

    for s = 1, screen.count() do
        gears.wallpaper.maximized(beautiful.wallpaper, s, true)
    end

(A normal image path will have the same effect.)

The xrandr setup, which triggers this:

    % xrandr --output HDMI2 --gamma 1.0:1.3:1.6 --mode 1920x1200 --panning 0x0 --pos 1366x0 --rate 60.0 --reflect normal --rotate normal --transform 1,0,0,0,1,0,0,0,1 --output LVDS1 --gamma 1.0:1.3:1.6 --mode 1366x768 --panning 0x0 --pos 0x432 --primary --rate 60.0 --reflect normal --rotate normal --transform 1,0,0,0,1,0,0,0,1
    (via [autorandr](https://github.com/phillipberndt/autorandr)).

    % xrandr -q
    Screen 0: minimum 8 x 8, current 3286 x 1200, maximum 32767 x 32767
    LVDS1 connected primary 1366x768+0+432 (normal left inverted right x axis y axis) 277mm x 156mm
    HDMI2 connected 1920x1200+1366+0 (normal left inverted right x axis y axis) 519mm x 324mm

Coming from:

    % xrandr --output HDMI2 --off
    % xrandr --output LVDS1 --gamma 1.0:0.909:0.833 --mode 1366x768 --panning 0x0 --pos 0x0 --primary --rate 60.0 --reflect normal --rotate normal --transform 1,0,0,0,1,0,0,0,1

That does not happen when using `feh` to set the wallpapers:

    awful.util.spawn('feh --randomize --bg-fill /path/to/wallpapers/')

The restart happens through `event_handle_configurenotify`, and I am seeing no
calls to `event_handle_randr_screen_change_notify`.

But it can be reproduced without the automatic restarting.

It's strange that it's fixed after another restart, so some persistent X setting/property might influence this.

I have not tested it with the default config yet, to check if it's related to my config.

psychon:
Usually users want a wallpaper on all their screens. With the current code, this
resulted in a loop of upload-wallpaper, {download-wallpaper, add-new-part,
upload-wallpaper}*.

Fix this by being more intelligent: Instead of setting the wallpaper
immediately, this now uses gears.timer.delayed_call() to set the wallpaper. All
following modifications which come in before the delayed call runs will still be
part of the current update. This should mean that during startup, there is just
a single upload of a wallpaper.

(The above is what happens if there is no wallpaper yet. If there is already
one, we use :create_similar() and thus should only upload the part of the
wallpaper that changed, but this doesn't really make a difference.)

As a side-effect, the new code no longer draws to the old wallpaper to modify
it, but always creates a copy of it. This means that:

Fixes: https://github.com/awesomeWM/awesome/issues/288

@blueyed Please check if this really fixes #288. Also, I have to admit that I didn't test this in a multi-screen environment....

### DIFF
--- a/lib/gears/wallpaper.lua
+++ b/lib/gears/wallpaper.lua
@@ -8,6 +8,7 @@
 local cairo = require("lgi").cairo
 local color = require("gears.color")
 local surface = require("gears.surface")
+local timer = require("gears.timer")
 
 local wallpaper = { mt = {} }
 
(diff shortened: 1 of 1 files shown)


blueyed:
> @blueyed Please check if this really fixes #288. Also, I have to admit that I didn't test this in a multi-screen environment....

:+1:

PR #533, opened 2015-10-16
Travis: try dist=trusty

blueyed:
http://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: required
+dist: trusty
 language: c
 env:
   matrix:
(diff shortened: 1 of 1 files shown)


blueyed:
No cmake by default?!

`cmake --version` fails.

blueyed:
We still need to build CMake:

> CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
> CMake 3.0.0 or higher is required.  You are running version 2.8.12.2

blueyed:
For reference, cmake is likely to be available by default (but probably too old): https://github.com/travis-infrastructure/packer-templates/issues/42

blueyed:
Should be good for now, but could be cleaned up more.

psychon:
@coveralls Huh?!?!?

PR #536, opened 2015-10-18
Add awful.util.is_dir

blueyed:
Extracted from https://github.com/awesomeWM/awesome/pull/187.

### DIFF
--- a/lib/awful/util.lua
+++ b/lib/awful/util.lua
@@ -214,6 +214,22 @@ function util.dir_readable(path)
         gfileinfo:get_attribute_boolean("access::can-read")
 end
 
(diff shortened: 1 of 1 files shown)


psychon:
Fine with me (with that linux-specific 21 gone, thanks for killing that)
P.S.:
```
$ grep -r EISDIR /usr/include/asm-generic 
/usr/include/asm-generic/errno-base.h:#define	EISDIR		21	/* Is a directory */
```

blueyed:
> Fine with me (with that linux-specific 21 gone, thanks for killing that)

It is still used with `file_readable` above.

Issue #348, opened 2015-07-24
timer: emit signals for "start", "pause", "stop"


blueyed:
I think the timer objects should also emit signals on "start", "pause" and "stop".

This would allow you to connect to it and e.g. show/hide widgets only while a timer is running.

psychon:
There is no "pause" and because I am lazy there is also no "again". The rest is in #537.

PR #537, opened 2015-10-18
gears.timer: Add start and stop signals (#348)

psychon:
This makes the timer emit signals for when it is started and stopped. This does
not add a signal for :again(), because that function just calls the other two
functions and thus already emits start and stop.

### DIFF
--- a/lib/gears/timer.lua
+++ b/lib/gears/timer.lua
@@ -28,6 +28,12 @@ local object = require("gears.object")
 --   started.
 -- @table timer
 
+--- When the timer is started.
+-- @signal start
+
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!
I think this should be added to the documentation, but offer to do it myself.  Just let me know - moving it to "next" anyway.

psychon:
Yeah, that's a good idea. Since I won't have much time in the near future, feel free to fix this. Thanks!

psychon:
Added some documentation myself (more than just the signals, also some more text for `:again()` and return values for the two `start_new` functions).

PR #538, opened 2015-10-18
(input/utf8) key handling in mappings -- yet another approach

nuko8:
This PR is a successor of #365.

### DIFF
--- a/objects/key.c
+++ b/objects/key.c
@@ -42,6 +42,7 @@
 /* XStringToKeysym() and XKeysymToString */
 #include <X11/Xlib.h>
 #include <xkbcommon/xkbcommon.h>
+#include <glib.h>
 
 /** Key object.
  *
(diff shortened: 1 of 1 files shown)


psychon:
Could you then use -1 instead of 4 for g_utf8_get_char_validated? Looking at its code, it almost-could read beyond the end of the string otherwise (in practice it won't, because the right bits aren't set in a null byte and thus this won't be the "middle byte" of a valid utf8 character, but still).

For testing, I ran this:
```lua
function t(a)
	local k = key({})
	k.key = a
	print(a, k.key, k.keysym)
end
t("#42")
t("42")
t("ö")
t("öü")
t("a")
```
For "4", and "öü" it ignores the second character and instead we get `odiaeresis`. I'd expect this to do something else... (fail and set the key to NoSymbol? Cause a Lua error?)

Also:

    /tmp/awesome/.build-psytux-x86_64-linux-gnu-5.2.1/objects/key.c: In function ‘luaA_keystore’:
    /tmp/awesome/.build-psytux-x86_64-linux-gnu-5.2.1/objects/key.c:108:46: warning: suggest braces around empty body in an ‘else’ statement [-Wempty-body]
                     /* This should not happen. */;

nuko8:
> For "4", and "öü" it ignores the second character and instead we get diaeresis.

This is due to the difference in handling "öü" and "diaeresis", not because of 4.

The function first asks `XStringToKeysym()` for checking if the given string is relevant to the keysym names that are defined in `keysymdef.h` and friends.  In this case, the string length doesn't  matter at all (Rather, the libxkbcommon online manual recommends 64 bytes long buffer for minimum). The string "diaeresis" falls into this category.

If that check fails, the function next checks if the string represents a single character which is supposed to be legally encoded in UTF-8.   In this case, the string length matters because the function needs to take only a single character; it's also a protection against illegal 5 or more byte length UTF-8 encoded characters.  Actually, even if 4 is specified, only read is the character "ö" and not "ü".  If 4 meant the string length, that would not happen and the both of the characters would be read just like "diaeresis".

As to the compiler warning, it's OK for me to put a pair of braces around that place.  Looks I wrongly understood the statement in `docs/02-contributing.md`:
```
- Place braces alone on new lines, and do not place braces for single
   line statement where it is not needed, ...
```
Or, you meant I should remove the branch itself?  Please clarify that.

nuko8:
@psychon 

> I'd expect this to do something else... (fail and set the key to NoSymbol? Cause a Lua error?)

Although I can't clearly see from this sentence what you really expect, I did something for that because I think it may be helpful for discussion if we have something concrete at hand.

I changed the handling of UTF-8 strings as follows:

1. Reject strings such as "öü" completely to accept for keysym conversio.  Don't take the first letter for convenience or whatever reason; the whole string is rejected, leaving a warning in syslog.

2. But that doesn't mean that every multi character strings are rejected.  As with u umlaut, there're two ways to express it in unicode and hence in UTF-8; the one is a UTF-8 string corresponding to U+00FC (a single character) and another is U+0075 U+0308 (two consecutive characters).  The new implementation in the last commit is written in such a way that the both cases are equally handled.

psychon:
Ah, thanks! I totally forgot about combining characters.

I like this version. Thanks!

For the git history: I would propose to squash all of this into a single commit. You undo pretty much all of Daniel's changes and so I don't think it's worth keeping them in here (you could acknowledge him in the commit message for starting all this). In the end, this PR only modifies key.c and I think it only does a single "thing".

nuko8:
@blueyed 

What do you think of @psychon's idea on squashing?

I know your long term efforts on this issue and do appreciate your initiative; without the both, I couldn't have done this work on my own.  Therefore, I believe it's utterly natural, fair and right for me to acknowledge your contribution to the issue.

I would highly appreciate it if you would permit me to mention your name with the contribution in the commit message of a squashed version of this PR if squashing should happen.

blueyed:
:+1: for squashing.

Thanks!

nuko8:
Rebased and squashed.

Thanks a lot for your understanding and cooperation!

psychon:
Ignore my comment on the patch, I am wrong.

This is fine with me: :+1:

PR #542, opened 2015-10-27
feat(awesomerc): add descriptions for some keybindings

Elv13:
This is an updated version of Actionless keybinding description patch


blueyed:
:+1:

PR #544, opened 2015-10-28
c.startup_id should be nil when launched without startup notification support

blueyed:
It seems like internally with use_snid=false the machinery to request it is not triggered, and then the next one with use_snid=false will get the previous one again.

I've tried to bisect it, but stopped at about HEAD~200 - seems to be an issue for longer already.

I've noticed it because I use `tag = X, switchtotag = true` with some client, and then recently changed some other default launcher to not use sn anymore, and that would get moved to the tag also (through my custom snid handler, before `spawn` got improved recently).

### DIFF
--- a/tests/test-spawn-snid.lua
+++ b/tests/test-spawn-snid.lua
@@ -1,0 +1,42 @@
+--- Tests for spawn's startup notifications.
+
+local spawn = require("awful.spawn")
+
+local manage_called, c_snid
+
+client.connect_signal("manage", function(c)
+  manage_called = true
+  c_snid = c.startup_id
+end)
+
+
+local ret, snid
+local steps = {
+  function(count)
+    if count == 1 then
+      ret, snid = spawn('urxvt', true)
+    elseif manage_called then
+      local c = client.get()[1]
+      assert(ret)
+      assert(snid)
+      assert(snid == c_snid)
+      return true
+    end
+  end,
+
+  -- Test that c.startup_id is nil for a client without startup notifications,
+  -- and especially not the one from the previous spawn.
+  function(count)
+    if count == 1 then
+      manage_called = false
+      ret, snid = spawn('urxvt', false)
+    elseif manage_called then
+      assert(ret)
+      assert(snid == nil)
+      assert(c_snid == nil, "c.startup_snid should be nil!")
+      return true
+    end
+  end
+}
+
+require("_runner").run_steps(steps)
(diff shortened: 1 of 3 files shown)


blueyed:
It seems like internally with use_snid=false the machinery to request it is not triggered, and then the next one with use_snid=false will get the previous one again.

psychon:
This startup_id stuff works by setting some environment variables which are then picked up by the spawned process. Looking at the code in `spawn.c`, I guess that `sn_launcher_context_setup_child_process(context);` sets up these environment variables. I guess the easiest fix would be to call that function after forking.

To do so, we should pass the `context` to `g_spawn_async_with_pipes` and then do `sn_launcher_context_setup_child_process(context);` in `spawn_callback`. Looking through some docs, this shouldn't cause any problems. Looking at the source code of `sn_launcher_context_setup_child_process`, this will even fix a memory leak. :-) (The environment variable is called `DESKTOP_STARTUP_ID`).

(To be exact: If `spawn_callback` is called with a non-NULL argument, it should assume that this is a context argument for `sn_launcher_context_setup_child_process` and call that function.)

@blueyed Do you want to implement this?

blueyed:
Thanks for the investigation!

> @blueyed Do you want to implement this?

Not currently.  Feel free to pick it up.

psychon:
https://gist.github.com/psychon/1f9408f68df87f203541 (Please add this commit to this PR and then add to the next milestone)

blueyed:
@psychon 
Added.  Now the first test fails: `assert(snid == c_snid)`?!
https://github.com/awesomeWM/awesome/pull/544/files#diff-8e95991a62031b5b7ffdf28da33739c4R22

psychon:
Hm. I tested this and it did work for me (Also: Yay, cmake.org now forces https, but has a certificate for *.kitware.com).

tests/test-spawn-snid.lua
```diff
@@ -0,0 +1,44 @@
+--- Tests for spawn's startup notifications.
+
+local spawn = require("awful.spawn")
+
+local manage_called, c_snid
+
+client.connect_signal("manage", function(c)
+  manage_called = true
+  c_snid = c.startup_id
+  print("manage_called: c_snid:", tostring(c_snid))
```
blueyed: @psychon 
This outputs `manage_called: c_snid:	nil` for the first test/step?!

psychon: Hm. Does the urxvt version on travis actually support startup notification? Looking at their `configure.ac`, it's an optional feature.
Ideas:
* Try without my patches so that we can agree that it's not my fault. :-)
* Try something like gnome-terminal, gedit, kate, ... (something which "surely" supports startup notification)
* These travis/cmake-fixes should be merged to master soonish, so that other branches can be rebased. This is independent from the rest of this PR (and I don't think we care much for SSL-certificates, because it's "only" a temporary travis machine that could be attacked).

psychon: Oh and: Does this work for you on your own computer? As I said, it did work on mine. :-/

blueyed: > Does this work for you on your own computer?

Yes, as with you.

I'll try with gnome-terminal instead.

blueyed: This PR is based on the old Travis infrastructure (Ubuntu Precise), where it seems to be just missing: http://packages.ubuntu.com/precise/rxvt-unicode

blueyed: I will finish and merge https://github.com/awesomeWM/awesome/pull/533, and then rebase it on top of it.

I think while urxvt is OKish, gnome-terminal would proably pull in too many deps anyway.

Issue #504, opened 2015-10-09
Add an optional filter to `awful.menu.clients`


d125q:
What does everyone think about adding a filter function as an optional argument to [`awful.menu.clients`](https://github.com/awesomeWM/awesome/blob/master/lib/awful/menu.lua#L471)?  I would find the ability to have it display, say, only visible clients, quite useful.

psychon:
It could use `awful.client.iterate` internally and let caller provide the arguments. That seems like a good idea.

d125q:
You are indeed correct about `awful.client.iterate` -- but I suppose it would have to be modified to make the filter optional. Well, it would be best if [`util.table.iterate`](https://github.com/awesomeWM/awesome/blob/master/lib/awful/util.lua#L584) was changed.

psychon:
We can just use `for c in client.filter(filter or function() return true end, start, s)`, so if no filter is specified, we just use an always-true filter.

d125q:
Of course; you are correct.  The only thing left to determine now is whether it's wise to begin iterating from the focused client, like `awful.client.iterate` by default does.

blueyed:
@d125q 
Do you feel like creating a PR for this?

d125q:
@blueyed: Yes; I will make one during the next few days.

PR #552, opened 2015-11-04
Add an optional filter to `awful.menu.clients` (#504)

d125q:
As per #504 -- this pull requests adds an optional filter to `awful.menu.clients`.

It also changes the way the clients are added to the menu: the adding begins with the focused client (if part of the filter), or as near to the right of the focused client as possible.  If the previous behavior is desired, where the clients were added from index 1 to the end, let me know so that I can revert it.

### DIFF
--- a/lib/awful/menu.lua
+++ b/lib/awful/menu.lua
@@ -15,6 +15,7 @@ local util = require("awful.util")
 local spawn = require("awful.spawn")
 local tags = require("awful.tag")
 local keygrabber = require("awful.keygrabber")
+local client_iterate = require("awful.client").iterate
 local beautiful = require("beautiful")
 local dpi = require("beautiful").xresources.apply_dpi
 local object = require("gears.object")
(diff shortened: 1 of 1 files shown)


lib/awful/menu.lua
```diff
@@ -465,14 +466,15 @@ end
 --------------------------------------------------------------------------------
 
 --- Build a popup menu with running clients and shows it.
--- @param args Menu table, see new() function for more informations
+-- @param args Menu table, see new() function for more informations.
 -- @param item_args Table that will be merged into each item, see new() for more
 --   informations.
+-- @param filter A function taking a client as an argument and returning true or
```
blueyed: Should be `@tparam func filter[opt] A function ...`.
This can be fixed when amending/merging.

blueyed:
Thanks, I've restarted the build jobs, which were apparently temporarily failing.

blueyed:
Re-submitted in #594.

Closing this in favor of the one with the higher number.. ;)

d125q:
Thank you.

blueyed:
Thank you, merged now.

PR #554, opened 2015-11-06
Add Travis for the 3.5 branch

blueyed:


psychon:
Fine by me. Feel free to merge this to the 3.5 branch (although I wonder if the apidoc stuff in there might still trigger and "do something"?).

Thanks!

blueyed:
We could maybe build the API doc for the stable / 3.5 branch also, possibly in another directory of https://github.com/awesomeWM/apidoc?

psychon:
Why would we want that? I don't think there are many documentation changes for 3.5 and the apidoc for the latest release should be "good enough".
(At least that's my opinion, but I'm a bad api-documentation-user and rather read the source code.)

blueyed:
Merged in https://github.com/awesomeWM/awesome/commit/91743f01b8a26793a780686f5896c76928120f5b.

psychon:
Thanks!

PR #558, opened 2015-11-07
Xkb detectable autorepeat

psychon:
This makes us use detectable autorepeat. This means when a key is held down, we only get press events, but the release events are skipped. Only when the key really is released do we get a release event.

Does this make sense to have? How much code will break?
I'm unsure....

### DIFF
--- a/xkb.c
+++ b/xkb.c
@@ -275,14 +275,14 @@ xkb_init(void)
                          XCB_XKB_MAP_PART_VIRTUAL_MODS | 
                          XCB_XKB_MAP_PART_VIRTUAL_MOD_MAP;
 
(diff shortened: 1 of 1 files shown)


actionless:
i like the idea very much -- i had to do pretty ugly hacks to detect continuous keypresses from the lua side

ShadOoW:
i've done a simple test on my machine, works as expected, i will do more tests soon, and report back if i find anything.

ShadOoW:
@psychon I was able to code the behavior i needed based on this branch (so all good!) @actionless i've wrote a simple function to handle this, you can have a look at it [here](https://gist.github.com/ShadOoW/c6034f6d9d9698d0591b)

thank you @psychon  hope to see this on pacman soon.

PR #559, opened 2015-11-07
Fix/silence various ldoc warnings in menubar

psychon:
I don't think `menubar.index_theme.new` really should be a public / should get the metatable for the created object via an argument, but whatever. This is the way it is....


blueyed:
Thanks!

PR #560, opened 2015-11-09
Reset layout group number to default when update_layout() is called

nuko8:
This reset is needed to keep the internal state of XKB consistent; otherwise, `awesome.xkb_get_layout_group()` returns the value for the previous XKB setting, and that confuses
`keyboardlayout.update_status()`, in particular, when `self._current` happens to be larger than `#self._layout`.

To be more specific, suppose we've just switched from a layout group to another via `setxkbmap`, and that the number of the layouts contained in the former group is larger than that in the latter.  Then, it is possible that the layout ID number we chose for the previous layout group is larger than the total number of the layouts contained in the current layout group.  

If that is the case, Awesome naturally raises a warning like
```
Oops, an error happened!
/usr/local/share/awesome/lib/awful/widget/keyboardlayout.lua:122: attempt to concatenate a nil value (field '?')
```
because `#self._layout < self._current`.

The proposed patch fixes that.

### DIFF
--- a/lib/awful/widget/keyboardlayout.lua
+++ b/lib/awful/widget/keyboardlayout.lua
@@ -250,6 +250,7 @@ local function update_layout(self)
         -- is greater by one than the real group number.
         self._layout[v.group_idx - 1] = layout_name
     end
+    awesome.xkb_set_layout_group(0);
     update_status(self)
 end
 


psychon:
Uhm. I don't like this patch. If the current state is wrong, then it's not some random widget's job to set a new state. Also, (while this is a stupid idea) what happens if the keyboard layout doesn't have a zero'th group? E.g. I could try setting `de:3`.

For this particular problem, I would suggest fixing the widget so that it e.g. displays `<i>invalid</i>` when "something goes wrong".
I do not have any good ideas for fixing the underlying problem (too few XKB understanding).

nuko8:
Thank you for the comment.

Let me clarify the issue.

Conceptually, there're two kinds of keyboard layout switching:
1.  Switchings between keyboard layouts listed in **a single group**.
2.  Switchings between **groups** of keyboard layouts.

This PR is relevant to Case 2. i.e., the case where the whole layout group is replaced with another group during an X session.

In contrast to Case 1, we need to specify explicitly which member of an updated group is the layout that is used after a layout switching takes place, because, without that specification, XKB keeps using the layout number set for the previous group prior to the switch.

So, switchings between layouts in a single group, say `de+us:2+jp:3`, causes no problem at all.  This is Case 1.

But if you change the group to `de+us:2` from `de+us:2+jp:3` and happen to set and use `jp:3` at that time , a problem arises; XKB knows the given layout group is changed and that the number of the layouts contained in the group is changed from 3 to 2, but XKB keeps assuming that the layout to use is still at Index 3 of the group.  Awesome gets a nil reference.

This PR addresses the case and tells XBK to use the first layout of the updated layout group.

psychon:
Awesome cannot do what you call (2), so it cannot leave XKB in an invalid state here. If you want to call this a bug, then the bug is in the X server / in `setxkbmap` / in something else.

I would fix the corresponding error in the keyboardlayout widget by replacing
```lua
    local text = ""
    if (#self._layout > 0) then
        text = (" " .. self._layout[self._current] .. " ")
    end
    self.widget:set_text(text)
```
with
```lua
    if (#self._layout > 0) then
        if self._layout[self._current] then
            self.widget:set_text(" " .. self._layout[self._current] .. " ")
        else
            self.widget:set_markup(" <i>unknown group</i> ")
        end
    else
        self.widget:set_text("")
    end
```
The keyboard layout currently does (1) from your list, but only when explicitly told to do so via a click. Your patch makes it do (1) on its own and I guess if I called `setxkbmap us:2` I would still get an error. My proposed patch (if you agree and want that merged) displays the "error" and doesn't "interfere" with the normal group switching.

Oh and google suggests that there are ways to set up groups that we wouldn't be able to figure out anyway:
https://bugs.kde.org/show_bug.cgi?id=160333

nuko8:
> If you want to call this a bug, then the bug is in the X server / in setxkbmap / in something else.

I didn't call it a bug.   I carefully chose the word "warning" from the beginning even if, as a matter of fact,  Awesome itself said "Oops, an error happened!" :-)

Interestingly, your proposed patch resembles my first try, except the `unknown group` block.  

I abandoned my first try because it looks illogical that  `awesome.xkb_get_layout_group()`, which is to return the current layout index,  returns a wrong number even after both XKB and Awesome know the layout group is changed.  That's why I insist that we need to specify the default index value for the new layout group when the keyboard layout group is changed.

I think we don't need `unknown group` block because the warnings are issued in case XKB can't make out possible inconsistency between the current layout index to be used and the total number of the layouts in the given group.  XKB will fall back to just what I wrote in the patch so that the keyboard won't be unusable.

`unknown group` is too much anyway.  It's a temporary inconsistency till XKB falls back to a natural layout index for a new group.

nuko8:
> Awesome cannot do what you call (2), so it cannot leave XKB in an invalid state here.

I'm feeling this perception is the cause of your misunderstanding on the PR.

As I wrote in the first comment:

> To be more specific, suppose we've just switched from a layout group to another via **setxkbmap**

Case 2 surely happens using `setxkbmap`, not by clicking the keyboardlayout widget and I know group switchings will never happen through clicks on that.

Without clicking the widget, layout switchings can be done with command line or key bindings (Rather, I rarely use the mouse/trackpad to change the layout unless I lose key bindings with unfamiliar layouts, which I often use for test purposes).

And, thanks to the original implementor,  the current keyboardlayout widget works well to display the state of the keyboard layout for almost all cases even for such non-clicking usages; the only exception is just the issue this PR raises.

> My proposed patch (if you agree and want that merged) displays the "error" and doesn't "interfere" with the normal group switching.

As for my patch, it doesn't  "interfere" with the normal **layout** switching.  Does it cause any regression or anything harmful?

>Oh and google suggests that there are ways to set up groups that we wouldn't be able to figure out anyway:
> https://bugs.kde.org/show_bug.cgi?id=160333

This appears to be a story that someone kindly wrote a patch to prevent a KDE app from crash even if a user feed a wrong configuration into it.  

This PR doesn't mean a crash-protect against wrong usages.

Rather, it is based on my experience with the development of the UTF-8 stuff for Awesome.  I wrote the code on US layout and tested it by switching the layout to DE, FR, RU etc. via `setxkbmap`.  Through the tests, I found the problem of the current keyboardlayout widget and am now proposing the patch to address it.

So, I think my patch is useful and helpful for decent usages by suppressing annoying warnings which are simply caused by failing to update the whole state to work cooperatively with XKB.

psychon:
(Sorry for the slow reply, I just don't think we'll reach any agreement and thus didn't feel like answering.)

> Does it cause any regression or anything harmful?

That's my problem. I don't really know. XKB can do lots of weird stuff and AFAIK it does specify the behaviour for these "out of range" numbers (I think one can define per-key(?) if the number should be wrapped into the valid range or just "stick" to the maximum, or something like that). So there might be some layout somewhere where this patch does actually break something.

Put differently, something that should only display the current state shouldn't change the state just because it doesn't understand it.

Elv13:
@nuko8 Ping?

Elv13:
@nuko8 last call for 3.6. I will close this in a week if there is no activity.

nuko8:
@Elv13 

Oh, looks I missed your previous message somehow.  Sorry about that and congrats on the coming new release.

PR #566, opened 2015-11-19
Remove useless information from awesome -v

psychon:
Most of this information isn't interesting. If you are getting awesome from a
distro, then the time, hostname and username of the build are likely 'random
stuff' and if you are building awesome yourself, then the hostname and username
are obviously yours and the time can still be interfered based on the awesome's
binary ctime.

The GCC version shouldn't make any difference at all.

Looking forward to your complaints. :-)


blueyed:
It basically only removes the `Build` line, right?

:+1:

psychon:
Yup. Sorry for making this so hard to see, I should have split this into multiple commits.

PR #567, opened 2015-11-19
Weak signal lifetime

### LINKED ISSUE #520: Could try redrawing an already destroyed wibox
I've been thinking about a bug for a few days and I think I finally figured it out.

The symptom can be seen in `tests/test-benchmark.lua`. With the following patch, awesome crashes when the test is run:
```diff
diff --git a/tests/test-benchmark.lua b/tests/test-benchmark.lua
index 065b8e4..ccdad52 100644
--- a/tests/test-benchmark.lua
+++ b/tests/test-benchmark.lua
@@ -66,7 +66,9 @@ local function e2e_tag_switch()
     do_pending_repaint()
 end
 
+benchmark(e2e_tag_switch, "tag switch")
 benchmark(create_and_draw_wibox, "create&draw wibox")
+benchmark(e2e_tag_switch, "tag switch")
 benchmark(update_textclock, "update textclock")
 benchmark(relayout_textclock, "relayout textclock")
 benchmark(redraw_textclock, "redraw textclock")
```
The crash occurs in `cairo_region_union_rectangle` and when you ask valgrind, it tells you that this is a use-after-free. I did not understand this.

Today I somehow came up with https://github.com/pavouk/lgi/issues/122 and then I understood it. The key is the following part from http://www.lua.org/manual/5.2/manual.html#2.5.2:

> Resurrected objects (that is, objects being finalized and objects accessible only through objects being finalized) have a special behavior in weak tables. They are removed from weak values before running their finalizers, but are removed from weak keys only in the next collection after running their finalizers, when such objects are actually freed. This behavior allows the finalizer to access properties associated with the object through weak tables.

The problem comes through `weak_connect_signal`. Internally it uses a weak table and thus the above behavior applies. This means that finalizers for objects which can only be reached via weak signals run before the weak signal is actually disconnected. This in turn means that if the weak signal gets emitted, the callback function can work with userdata which was already finalized. And this explains how the above crash happens.
In the above case this means that changing the tag will cause a taglist to be updated that is only weakly reachable and whose wibox was already finalized.

When lgi gets fixed, the crash doesn't just go away, but instead will likely be replaced with an error like "trying to work on object that was already destroyed" (I know this because I sometimes see this problem when the Lua code manages to use the drawable before some lgi object and then we get such an error).

So... I guess the code behind `weak_connect_signal` needs to be changed to use tables with weak values instead of weak keys? Or does anyone have other ideas?

psychon:
This should fix #520 and hopefully not introduce any new problems. Lots of magic and duct-tape is needed to make this work with Lua 5.1. Does anyone care that I (ab)use the function environment for this?

### DIFF
--- a/lib/gears/object.lua
+++ b/lib/gears/object.lua
@@ -40,7 +40,7 @@ function object:add_signal(name)
     if not self._signals[name] then
         self._signals[name] = {
             strong = {},
-            weak = setmetatable({}, { __mode = "k" })
+            weak = setmetatable({}, { __mode = "kv" })
         }
     end
 end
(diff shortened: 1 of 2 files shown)


psychon:
Did anyone look at this yet? Are you scared of this magic? May I use `setfenv` like this in Lua 5.1? @blueyed @Elv13 Ping? Any complaints?

blueyed:
Looks good to me.

Issue #218, opened 2015-05-09
awful.util.getdir("config") vs. `awesome -c CONFIG`


ff2000:
I ran into issues when trying to run Elv13 awesome-config. Scripts and resources get concatenated with awful.util.getdir("config"). I wanted to simply run it safely without losing my current config, which turned out to be impossible as scripts were not found, because getdir always returns "${XDG_CONFIG_HOME}/awesome".
I would expect this to return the actual location my rc.lua is stored, when I pass one. E.g. "$HOME/test/awesome/" when I run "awesome -c $HOME/test/awesome/rc.lua".

Bew78LesellB:
For my configuration, i've made a Theme Loader that find my different configuration, then try to load them, and set a global variable with the curre't theme path... 
See https://github.com/Bew78LesellB/ConfigAwesome for more information!

ff2000:
Thx for your answer! But your config doesn't work for me, either - because you load files relative to awful.util.getdir("config"), which returns ~/.config/awesome in any case ;) I really thought in the first place that it worked. It looked like the default theme, but - well - probably you like that. Then I read your rc.lua and I realized: This IS the default theme ;)

You can try what this issue is about:
```
mv ~/.config/awesome ~/.config/awesome_bkp
awesome -c ~/.config/awesome_bkp/rc.lua
```
(The default directory has to be out of place)

In the meantime I found awesome.conffile. Manipulating that in awful.util.getdir shouldn't be that hard. I can create a PR if you think this issue is a real issue at all.

And there is another related issue: LUA_PATH isn't expanded, so init.lua never is considered for require when running an rc.lua different from the default one (~/.config/awesome/rc.lua). I have to manually export LUA_PATH="?/init.lua;;" to make it work again (which isn't a big issue)

Bew78LesellB:
Oh ok, I see know what you mean! 

For information, my config is in wip, and the dev branch is much more advanced than master! 

But if you take the idea of my ThemeLoader, I think you can redefine the awful.utils.getdir function, to return another base directory!... 

Also: you could make kind of script that execute awesome, then a function passed by awesome-client, (to the theme loader for example) with the given config path!

PR #575, opened 2015-11-28
Add various directory getters to awful.util

psychon:
This started as a fix for #218, but ended up with a couple more functions.

The PR adds functions that get various useful directories. Especially, we now have a function that returns the directory containing the configuration. `awful.util.getdir("config")` retains its old (IMO misleading) behavior, just because of backwards compatibility.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -39,7 +39,7 @@ end
 
 -- {{{ Variable definitions
 -- Themes define colours, icons, font and wallpapers.
-beautiful.init("@AWESOME_THEMES_PATH@/default/theme.lua")
+beautiful.init(awful.util.get_themes_dir() .. "default/theme.lua")
 
 -- This is used later as the default terminal and editor to run.
 terminal = "xterm"
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -318,7 +318,7 @@ globalkeys = awful.util.table.join(
                   awful.prompt.run({ prompt = "Run Lua code: " },
                   mypromptbox[awful.screen.focused()].widget,
                   awful.util.eval, nil,
-                  awful.util.getdir("cache") .. "/history_eval")
+                  awful.util.get_cache_dir() .. "/history_eval")
               end,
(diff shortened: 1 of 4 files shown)


blueyed:
Would be nice to have something for `@AWESOME_THEMES_PATH@`, to resolve the following from your config:

    local theme_assets = dofile("@AWESOME_THEMES_PATH@/xresources/assets.lua")

Or is this possible already?

psychon:
Rebased to fix conflicts and added a new commit to address @blueyed's comment

blueyed:
> The sample that you gave is from themes/xresources/theme.lua

Thinking about it the `assets` module should not be in `themes` maybe?
/cc @actionless

actionless:
i think it should. mb we should just find a better way of importing themes.

mb adding themes' path to `package.path`?

blueyed:
> adding theme's path to package.path?

That would work for me.

psychon:
So... instead of giving `beautiful.init` the path to `theme.lua` as it is done now, we would give it the path to the directory containing the theme, it would add that path to `package.path` (and perhaps save it somewhere so that the theme can e.g. find images based on that path) and then it just does `require("theme")`? Dunno, sounds weird and different...

Why do we want to do that? Which problem would that solve?

blueyed:
> Which problem would that solve?

I have a theme derived from the xresources one (solarized), and would like to use functions from xresource's assets module in there.

Currently I am using this pattern to access the parent theme:
```
local this_file = debug.getinfo(1, "S").source:sub(2)
function dirname(name)
    return name:match("^(.+)[/\\].-$") or (default or "")
end
local theme = dofile(dirname(this_file) .. "/../xresources/theme.lua")
```

psychon:
Yeah... perhaps we should move assets to `beautiful.assets`? (and remove `icons/` and replace its use with the icon from assets?)

psychon:
If you want to move parts of the xresources theme elsewhere, feel free, but that's not part of this PR. Sorry.

PR #576, opened 2015-11-29
Spawn exit code

psychon:
This adds support for getting a spawned processes exit status. Just because @blueyed wanted this. :-)

Some question: Should the last two arguments to `with_line_callback` be swapped? I guess it makes more sense to have the exit callback first and then the done callback.
And before someone asks: The done callback is called when the streams for stdout and stderr are closed. This is something else than "the spawned process exits", because it can in turn start other processes that share stdout and keep it open, etc.

### DIFF
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -91,20 +91,24 @@ end
 
 --- Spawn a program and asynchronously and capture its output line by line.
 -- @tparam string|table cmd The command.
--- @tparam[opt] function stdout_callback Function that is called with each line of
---   output on stdout, e.g. `stdout_callback(line)`.
--- @tparam[opt] function stderr_callback Function that is called with each line of
---   output on stderr, e.g. `stderr_callback(line)`.
--- @tparam[opt] function done_callback Function to call when no more output is
---   produced.
--- @tparam[opt] function exit_callback Function to call when the spawned process
+-- @tab callbacks Table containing callbacks that should be
+--   invoked on various conditions.
+-- @tparam[opt] function callbacks.stdout Function that is called with each line of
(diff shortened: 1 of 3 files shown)


blueyed:
Thanks! :)

blueyed:
@Elv13 @actionless 
What's your opinion on the open question(s)?

Elv13:
Yes, I think swapping would make sense. exit look more useful. However, given the growing numbers of parameters, should it just switch to "args" instead? calling something like `foo("ls", nil, nil, nil, nil, nil, nil, function() end)` will eventually get frustrating compared to `foo { command = "ls, done_callback=function() end}`. This can be done without breaking the API by simply removing the named table elements before passing it to `capi.awesome.spawn`. I think this is one of the case where lua alternate call syntax is much better[1].

[1] Yes, calling `func({foo=2})` and `func { foo=2 }` do the same thing. This is why `print "test"` work. This is some of less known Lua syntax sugar.

blueyed:
Yes, using a args table seems sensible.

psychon:
I'll update this.

Also, since this API wasn't part of any release yet, I'm not worried about breaking it. Is there any reason why this shouldn't always get an args table?

actionless:
i like args too

psychon:
Done. Note that I didn't add the command to the callback, because that one is always required and IMO not confusing, only the various callbacks are.

blueyed:
Thanks!

psychon:
I removed commit https://github.com/psychon/awesome/commit/ec21093917e60775d3f31a992ced178fdb14a6b4 from this PR. @blueyed's comment is right and this should be done properly in a separate PR. I only noticed this while doing these edits and did something simple about it.

PR #578, opened 2015-11-30
Add an option to set the preferred client icon size

lukash:
Hi guys :)

I tried to get this merged in a (less than satisfactory) per-screen form some years ago... I figured I'm never gonna do this properly, so I simplified the implementation to just one global setting. While this doesn't take DPI into account etc., it's better than nothing... Right? :)

Adds a awesome.set_preferred_icon_size() method to set the preferred icon size of the clients.

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -543,6 +543,9 @@ main(int argc, char **argv)
     /* We have no clue where the input focus is right now */
     globalconf.focus.need_update = true;
 
+    /* set the default preferred icon size */
+    globalconf.preferred_icon_size = 0;
+
     /* XLib sucks */
     XkbIgnoreExtension(True);
 
--- a/ewmh.c
+++ b/ewmh.c
@@ -662,27 +662,45 @@ ewmh_window_icon_get_unchecked(xcb_window_t w)
 }
 
 static cairo_surface_t *
-ewmh_window_icon_from_reply(xcb_get_property_reply_t *r)
+ewmh_window_icon_from_reply(xcb_get_property_reply_t *r, uint32_t preferred_size)
 {
-    uint32_t *data;
-    uint64_t len;
+    uint32_t *data, *end, *found_data = 0;
+    uint32_t found_size = 0;
 
     if(!r || r->type != XCB_ATOM_CARDINAL || r->format != 32 || r->length < 2)
         return 0;
 
     data = (uint32_t *) xcb_get_property_value(r);
-    if (!data)
-        return 0;
+    if (!data) return 0;
 
-    /* Check that the property is as long as it should be, handling integer
-     * overflow. <uint32_t> times <another uint32_t casted to uint64_t> always
-     * fits into an uint64_t and thus this multiplication cannot overflow.
+    end = data + r->length;
+
+    /* Goes over the icon data and picks the icon that best matches the size preference.
+     * In case the size match is not exact, picks the closest bigger size if present,
+     * closest smaller size otherwise.
      */
-    len = data[0] * (uint64_t) data[1];
-    if (!data[0] || !data[1] || len > r->length - 2)
-        return 0;
+    while (data + 1 < end) {
+        /* check whether the data size specified by width and height fits into the array we got */
+        uint64_t data_size = (uint64_t) data[0] * data[1];
+        if (data_size > (uint64_t) (end - data - 2)) break;
 
-    return draw_surface_from_data(data[0], data[1], data + 2);
+        /* use the greater of the two dimensions to match against the preferred size */
+        uint32_t size = MAX(data[0], data[1]);
+        /* pick the icon if it's a better match than the one we already have */
+        if (data[0] && data[1] && (
+            (found_size < preferred_size && size > found_size) ||
+            (found_size > preferred_size && size >= preferred_size && size < found_size)))
+        {
+            found_data = data;
+            found_size = size;
+        }
+
+        data += data_size + 2;
+    }
+
+    if (!found_data) return 0;
+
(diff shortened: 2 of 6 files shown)


psychon:
I like this idea. It's easier to do than my idea of handing all icons to Lua (which will apparently never happen...).
Since commits tend to get lost: I left some comments at 4c69ca876d23542b6eb336a278482cd1b06ea2. My biggest concern right now are integer overflows.

I don't know if I got all the integer overflows right, but my code for reading all instead of just the first icon can be found here: https://github.com/psychon/awesome/commit/dbb3258472748379ac91e3c9e45f040c92773918#diff-b9fe4ccd229bc1c65dafde3891a104e0R618

lukash:
I've pushed the fixes as separate commits, meant to be squashed before merging. How to proceed with the squashing? Create a new PR or rather push -f? I'm not aware of the preferable workflow in this case...

blueyed:
You can `push -f` them, but during review it's better to keep them as-is (to see what has changed during the discussion).
Just for reference, just might be interested in reading up on Git's fixup commit support (e.g. http://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html).

psychon:
With the previous two (minor) things fixed, this looks good to me. Thanks a lot!

lukash:
I hope it's fine like this :) ready to squash and push -f it
@blueyed thanks for that, I didn't know git has fancy options for what I've been doing manually till now :) still I've used regular commits to be able to enter descriptive commit messages.

psychon:
Sorry, two more comments again.

Should we just assume that windows have well-formed `_NET_WM_ICON` properties and ignore the possibility of overflows? Who would attack a window manager anyway...?

psychon:
Here is some test program to experiment with "evil icons": https://gist.github.com/psychon/0a9491c61a3f57bd3663
It can be compiled with `gcc -lxcb invalid_icon.c -o invalid_icon`. As arguments, it gets pairs of `width height`. The final pair is `width height length`. It will then create a window with `_NET_WM_ICON` containing icons in that order where the last one is possible broken (if a wrong length was specified, it should be `width*height+2`; so length is in 4-byte-quantities, not in bytes).

Then, I added some `printf`s to your code: https://gist.github.com/psychon/79b9201d19dbe609dac9
The cast of `(uint32_t) offset` is to simulate 32 bit pointers where, I think, we get an implicit conversion to 32 bit pointers.

With arguments `1 1 65536 65536 4096` I get:
```
Looking at icon 1x1 at offset 0 (0)
Adding offset 3 (3)
Looking at icon 65536x65536 at offset 12 (3)
Adding offset 4294967298 (2)
```
If the prefered icon size were large and it would try the second icon, we'd get a crash... Possibly... I think....

Does this make sense? Would it be a good idea to somehow come up with automatic tests for this stuff? Or is the issue minor enough that we can just ignore it after everything works right?

lukash:
Ok ok :) I rewrote the whole thing. There were actually two places where the same overflow-prone check was happening, so I moved it to one place and this new check should never overflow and compare the size calculated from width and height against the size of the data that are left in the array.

I dare you try and break it :P I tested it, but not with your test program. Please check it if still in doubt :) I included some logging in a separate commit for your convenience.

I was also tempted to introduce an arbitrary constant to limit the maximum size of an icon that we'd accept, but in the end the icons aren't guaranteed to be ordered and so we couldn't just quit (or perhaps we could, if the constant was big enough, but my conscience wouldn't be clean :)) if we hit an icon that's too big. This was based on the fact that in your hypothetical situations the icons (if the data were really there) wouldn't be displayable at all, they actually most certainly wouldn't fit in RAM :) (not saying some more realistic combination couldn't trigger this)

Unittests would be nice, but I'm not that eager to write them, sorry :/

psychon:
Sorry that this takes so long. Feel free to fix this minor thing with 0xSomething / Somethingx0 icons and then squash the whole thing (except for the last commit). I agree with your "ultimate fix" commit message. :-)

lukash:
Done :)

psychon:
Thanks a lot. And an even bigger "Thank you" for bearing with me!

lukash:
You're welcome, it was me who couldn't get it right... :)

I've seen somewhere you asked for candidates to cherry-pick into the 3.5 branch, this might be one of them.

awesome.c
```diff
@@ -543,6 +543,9 @@ main(int argc, char **argv)
     /* We have no clue where the input focus is right now */
     globalconf.focus.need_update = true;
 
+    /* set the default preferred icon size */
+    globalconf.preferred_icon_size = 0;
```
blueyed: Could the default be more dynamic, e.g. depending on the DPI?

psychon: This doesn't really depend on the DPI. It depends on the size at which the icons will be visible. As a first approximation, this is the height of the wibox, but I guess that only works with the default config. (Plus, different screens can have different heights here)

If you want, we could add something to the default config which sets this to... something. But what?

blueyed: > This doesn't really depend on the DPI

Your are right in general, but if the default would be something like 24px, then for a DPI of 180 (instead of 96) the default should be higher.

After all, the default should probably refer to pt and not px?

psychon: The default is 0 which is pretty much what we are doing right now already (right now we pick the first one, now we will pick the smallest; however I only ever saw icons sorted by size and so this shouldn't make a difference).

The nice thing about 0 is that it's a DPI-independent value. :-)

I agree that "come up with something that makes us pick 'good' icons by default" is still left to do, but that is a hard problem and IMO shouldn't stop this PR.

PR #583, opened 2015-12-04
Use integer size when creating titlebar

barlik:
I stumbled upon this issue during investigation of a bug. Function `get_titlebar_function()` calls `titlebar_resize()` which requires size as an integer.

### DIFF
--- a/lib/awful/titlebar.lua
+++ b/lib/awful/titlebar.lua
@@ -64,7 +64,7 @@ end
 local function new(c, args)
     local args = args or {}
     local position = args.position or "top"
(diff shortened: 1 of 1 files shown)


psychon:
Could you use `awful.util.round` instead so that 1.9 becomes 2 instead of 1? (While 1.1 is still rounded to 1)

Besides that: Nice catch, thanks. No idea why my titlebars work, but apparently I am just lucky with me font size being an even number...

barlik:
I've created a new pull request https://github.com/awesomeWM/awesome/pull/584

PR #584, opened 2015-12-05
Use integer size when creating titlebar

barlik:
Titlebar height needs to be an integer.


psychon:
Looks good to me, thanks!

Issue #585, opened 2015-12-05
beautiful.xressources needs to convert colors `rgb:rr/gg/bb` -> `#rrggbb` and `argb:aa/rr/gg/bb` -> `#aarrggbb`


psychon:
See https://github.com/awesomeWM/awesome/pull/421#issuecomment-162161587 and the following comment.

Does anyone know the format for these things? I don't even know where I got the format that I am using from nor do I know what is allowed.

psychon:
So... rxvt-unicode apparently uses `XParseColor` and supports some non-standard extensions: http://sources.debian.net/src/rxvt-unicode/9.21-1/src/rxvttoolkit.C/#L827
The extensions being: A prefix `[50]` means "50% transparency", `rgba:rrrr/gggg/bbbb/aaaa`.
Looking at the source of `XParseColor`... things get complicated. So, `#rrggbb` works with the individual components having a length from 1 to 4 (so `#rgb` to `#rrrrggggbbbb`).
Then Xcms is queries somehow: http://insar.stanford.edu/~lharcke/programming/Xcms/
Finally, the X11 server is asked (this makes things like `red` work).

nuko8:
> Does anyone know the format for these things? I don't even know where I got the format that I am using from nor do I know what is allowed.

Perhaps, you saw them in the man page, `X(7)`, `COLOR NAMES`.

actionless:
probably, in `beautiful.xresources` we could check if color string starts from `rgb:`

prefix `[50] ` is supported only in urxvt, right? seems it could problems elsewhere if it will be used in xsetting not prefixed with `URxvt`, ie ```
*background: #0e0021
URxvt*background: [95]#0e0021
```

PR #588, opened 2015-12-07
feat(beautiful: xresources): read rgb: and argb: prefixed colors

actionless:
should address https://github.com/awesomeWM/awesome/issues/585

### DIFF
--- a/lib/beautiful/xresources.lua
+++ b/lib/beautiful/xresources.lua
@@ -56,6 +56,9 @@ function xresources.get_current_theme()
             print("W: beautiful: can't get colorscheme from xrdb (using fallback).")
             return fallback
         end
+        if colors[key]:find("rgb:") then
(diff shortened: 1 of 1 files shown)


psychon:
Sorry, but this doesn't handle `rgb:f/f/f` (should be white), `#fff` (still white) nor `rgb:/ffff/ffff/ffff` (white, too). Should this be fixed in `gears.color` so that it supports things like `#fff` properly?

Thanks for writing this. :-)

actionless:
so do you suggest moving `rgb:X/Y/Z -> #XYZ` conversion from here to `gears.color`? or only support for `#XXX` and `#XXXXXXXXXXXX` ?

psychon:
Leave this the way it is. I guess we could even merge this as-is. I only meant the `#fff`-case to be fixed in `gears.color`, but I am not really sure if that is a good idea.

PR #590, opened 2015-12-08
Add meminfo support

mindeunix:
Only as an example.

It took me about an hour to write with zero knowledge about C/C++ programming  and at least for me it works really fast.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -535,3 +535,7 @@ end)
 client.connect_signal("focus", function(c) c.border_color = beautiful.border_focus end)
 client.connect_signal("unfocus", function(c) c.border_color = beautiful.border_normal end)
 -- }}}
+
+for i=0,1000 do
+  print(awesome.sys_get_meminfo())
+end
--- a/system.c
+++ b/system.c
@@ -1,0 +1,16 @@
+#include "system.h"
+
+#include <glibtop.h>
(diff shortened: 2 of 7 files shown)


psychon:
Could you also add `libgtop2-dev` here: https://github.com/awesomeWM/awesome/blob/0d2e6d20684936035d438ea4da6a8c232f39d294/.travis.yml#L31
That should make the travis check succeed... Possibly.

awesomerc.lua
```diff
@@ -535,3 +535,7 @@ end)
 client.connect_signal("focus", function(c) c.border_color = beautiful.border_focus end)
 client.connect_signal("unfocus", function(c) c.border_color = beautiful.border_normal end)
 -- }}}
+
+for i=0,1000 do
+  print(awesome.sys_get_meminfo())
+end
```
blueyed: That should not be in the default config probably?!

system.c
```diff
@@ -0,0 +1,16 @@
...
+#include <glibtop/mem.h>
+
+#include <glib.h>
+#include <unistd.h>
+
+int luaA_sys_get_meminfo(lua_State *L)
+{
+    glibtop_mem mem;
+    glibtop_init();
+    glibtop_get_mem(&mem);
```
blueyed: Is that the memory usage of the system or the process / awesome?

PR #593, opened 2015-12-12
Vim modelines and Editorconfig

psychon:
Thanks to @blueyed's comment on https://github.com/psychon/awesome/commit/ec21093917e60775d3f31a992ced178fdb14a6b4, I know about editorconfig.org. This PR first updates various vim modelines and then adds an editorconfig config file.

For finding missing modelines, the following was used:
```sh
git ls-tree HEAD -r |
  grep -v manpages/awesome |
  cut -c 54- |
  grep -v png |
  while read file
  do
    printf "%s " "$file"
    tail -n 1 "$file"
  done  |
  grep -v '// vim: filetype=c:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80' |
  grep -v -- '-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80' |
  grep -v '# vim: filetype=sh:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80' |
  grep -v '# vim: filetype=cmake:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80:foldmethod=marker'
```
After the commits in this PR, the above produces the following output:
```
.editorconfig indent_style = tab
.gitignore /awesome
.travis.yml   - if [ "$BUILD_APIDOC" = "true" ]; then build-utils/travis-apidoc.sh; fi
LICENSE Public License instead of this License.
Makefile .PHONY: cmake-build cmake install distclean $(BUILDLN) tags
README.md or [v3](http://www.gnu.org/licenses/gpl.html)).
awesome.desktop Type=Application
awesomerc.lua -- }}}
common/atoms.list AWESOME_CLIENT_ORDER
docs/00-authors.md all copies or substantial portions of the Software.
docs/01-readme.md or [v3](http://www.gnu.org/licenses/gpl.html)).
docs/02-contributing.md   it first by sending email to yourself.
spec/menubar/icons/awesome/index.theme Size=64
themes/default/README     Licensed under CC-BY-SA-3.0
```
(I don't know why `awesomerc.lua` doesn't have a modeline, but it's always been like this and I will just keep it this way)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,26 @@
+# See http://editorconfig.org
+
+# Don't search for other .editconfig files
+root = true
+
+[*]
+
+# Unix-style newlines
+end_of_line = lf
+
+# There a newline at the end of the file
+insert_final_newline = true
+
+# Trailing whitespaces, no one needs them
+trim_trailing_whitespace = true
+
+# Why would anyone use anything else?
+charset = utf-8
+
+# This is just what we use, sorry if you disagree
+indent_style = space
+indent_size = 4
+
+# In a Makefile, the tab is significant and not "just" whitespace
+[Makefile]
+indent_style = tab
--- a/awesome-version-internal.h
+++ b/awesome-version-internal.h
@@ -5,3 +5,5 @@
 #define AWESOME_RELEASE             "@AWESOME_RELEASE@"
 
 #endif //_AWE_VERSION_INTERNAL_H_
+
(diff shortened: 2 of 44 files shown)


blueyed:
> I don't know why awesomerc.lua doesn't have a modeline, but it's always been like this and I will just keep it this way

Probably because it's a mess regarding its indentation?!
(That might also be the result of a missing modeline though)

btw: for Vim I can recommend this plugin: https://github.com/roryokane/detectindent

PR #595, opened 2015-12-14
Add padding for tag

tequillaz:
Add tag.- setpadding(...) and getpadding(...) functions
Modify layout arrange function to use tag.getpadding(...)

PR #596, opened 2015-12-14
Beautiful useless gap

tequillaz:
(sounds funny)
This will make space between windows ("useless gap") equal
to useless_gap value instead of useless_gap_2 value as
it was before. Also we removes extra space equal
useless_gap_2 value between workarea border and window
because it is not "useless gap" it is "tag padding".

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -181,14 +181,44 @@ function layout.arrange(screen)
         local p = layout.parameters(nil, screen)
 
         local useless_gap = p.useless_gap
+        local ugap_edge = 0
+        local ugap_d = math.floor(useless_gap/2)
+        local ugap_m = math.fmod(useless_gap,2)
+        local m = 0
+
 
         p.geometries = setmetatable({}, {__mode = "k"})
         layout.get(screen).arrange(p)
         for c, g in pairs(p.geometries) do
-            g.width = math.max(1, g.width - c.border_width * 2 - useless_gap * 2)
-            g.height = math.max(1, g.height - c.border_width * 2 - useless_gap * 2)
-            g.x = g.x + useless_gap
-            g.y = g.y + useless_gap
+            if g.x ~= p.workarea.x
+                then m = ugap_d
+                else m = ugap_edge
+            end
+            g.x = g.x + m
+            g.width = g.width - m
+
+            if (g.x + g.width) ~= (p.workarea.x + p.workarea.width)
+                then m = ugap_d + ugap_m
+                else m = ugap_edge
+            end
+            g.width = g.width - m
+
+            if g.y ~= p.workarea.y 
+                then m = ugap_d
+                else m = ugap_edge
+            end
+            g.y = g.y + m
+            g.height = g.height - m
+
+            if (g.y + g.height) ~= (p.workarea.y + p.workarea.height)
+                then m = ugap_d + ugap_m
+                else m = ugap_edge
+            end
+            g.height = g.height - m
+
+            g.width = g.width - c.border_width * 2 
+            g.height = g.height - c.border_width * 2
+            
             c:geometry(g)
         end
         screen:emit_signal("arrange")


lib/awful/layout/init.lua
```diff
@@ -123,14 +123,19 @@ function layout.arrange(screen)
         p.workarea = capi.screen[screen].workarea
         local useless_gap = tag.getgap(tag.selected(screen),
                                        #client.tiled(screen))
+        local ugap_edge = 0
+        local ugap_d = math.floor(useless_gap/2)
```
actionless: can't this lead to different gaps between clients if useless_gap will be set to an uneven value?

actionless:
also removing screen padding breaks useless gaps behaviour when they're enabled only on some specific tags

lib/awful/layout/init.lua
tequillaz: Nope, because we have compensation `upag_m=useless_gap mod 2`, wich will add to right or bottom client edge if useless_gap is odd

tequillaz:
I know about it. 
I was add padding API for tag in another feature (https://github.com/awesomeWM/awesome/pull/595).
You can set different padding and useless_gap for different tags

actionless:
also it breaks this feature:
https://github.com/awesomeWM/awesome/blob/master/lib/awful/tag.lua#L417-L419
(for the same reason)

tequillaz:
It is not or i dont undestend how it can break it. 
We have no useless_gap on tag with only one window, so useless_gap is zero. With this patch whether or not we have the same result.

actionless:
you've already mentioned what now it's required to manually set screen padding to have equal gaps around the clients

that means what if screen padding will be set, that codeblock which i referenced above won't be affective and single client will have a padding around

tequillaz:
Yes client will have padding around but not unnecessary.
I did this because `useless_gap` and `padding` is not the same. `useless_gap` is gap only between windows, `padding` is gap between window and workarea edge. One window on your tag is not mean you dosnt want padding around your window :) Also with this patch we dosnt have some extra space around windows that `useless_gap` add.
The value of `padding` is becoming more predictable, and `useless_gap` too.

actionless:
it's good what this PR makes a more clear differentiation between useless gaps and workspace padding, but at the same time it shouldn't break the current behavior

tequillaz:
I undestand you, @actionless, but i think this pull request changes not so much to save old behaviour as option. Also current changes have `ugap_edge` value (equal to zero) which used to add extra space between window and workarea as `useless_gap` space. May be will need to use them like this

```lua
    if padding then
        ugap_edge = 0
    else
        ugap_edge = useless_gap
    end
```

This will use `useless_gap` as space betweeb window and workarea edge if have no padding

actionless:
so, exposing `ugap_edge` to smth like `theme.useless_gap_edge`  should return the current ehavior if `theme.useless_gap == theme.useless_gap_edge`, right?

tequillaz:
Nope, theme is not best place to save options for return old behaviour. I say about `local padding` variable to check it exist or not, no option to return old behaviour

actionless:
i am not agree. having that in theme allows to select gaps behaviour in theme without adding any extra logic in rc.lua

psychon:
@actionless Somehow I feel like you should decide about this PR. If I understand you correctly, your opinion is that this should be closed, right?

PS: Sorry, but I'm trying to get the number of open issues and PRs down. :(

actionless:
ehm, i like the proposed change but i have a different vision on how that option should be configured, because the proposed way to configure it don't allows to have the same size of useless gap and window padding (the current behavior) and will force to add additional logic to set screen padding when changing useless gap

i can update the PR to set that behavior from the theme (ie it will be the option like `theme.useless_gaps_screen_padding`)

psychon:
Sounds good to me.

blueyed:
@tequillaz 
Do you agree with @actionless updating the PR?

tequillaz:
Yes, iam sorry for long wait answer

Elv13:
@actionless (or @tequillaz) Can you do the new/updated PR please. I am trying to either get the open PR merged, closed them or explicitly postponed to 3.7.

tequillaz:
updated to current master without any change

Elv13:
Also, it can't be merged until is passed the unit tests and is unit tested itself.

julien-f:
Any news on this?

lib/awful/layout/init.lua
```diff
@@ -181,14 +181,44 @@ function layout.arrange(screen)
...
+            g.height = g.height - m
+
+            if (g.y + g.height) ~= (p.workarea.y + p.workarea.height)
+                then m = ugap_d + ugap_m
+                else m = ugap_edge
+            end
+            g.height = g.height - m
+
+            g.width = g.width - c.border_width * 2 
+            g.height = g.height - c.border_width * 2
```
actionless: just a side question, when setting `width` or `height` property, signal handler will fire just after assignment or on the next evenloop?

if second, how it will handle multiple value modifications -- handle the signals in some order or will handle only the last one?

psychon: Signals are emited immediately.

actionless: then the code above will fire them multiple times, not very good

idk, mb just close it, if anybody would like to resurrect it the PR still will be in the list of closed ones and the code will not disappear

psychon: Uhm. The code above just modifies a normal Lua table. Only the call to `c:geometry()` in line 192 / 222 actually changes things and causes signals to be emitted.

actionless: my bad


so i've just merged it locally to test again -- smth got changed (in awesome itself probably) -- now `ugap_edge` is being added to the "outer" gap size

however setting `ugap_edge` to `-useless_gap`



here is the full patch with my proposal fix to make those edges configurable via `beautiful`:

```diff
diff --git a/lib/awful/layout/init.lua b/lib/awful/layout/init.lua
index 5fa5ec3d..e5ddd50f 100644
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -21,6 +21,7 @@ local client = require("awful.client")
 local ascreen = require("awful.screen")
 local timer = require("gears.timer")
 local gmath = require("gears.math")
+local beautiful = require("beautiful")
 
 local function get_screen(s)
     return s and capi.screen[s]
@@ -189,6 +190,7 @@ function layout.arrange(screen)
     delayed_arrange[screen] = true
 
     timer.delayed_call(function()
+        local theme = beautiful.get()
         if not screen.valid then
             -- Screen was removed
             delayed_arrange[screen] = nil
@@ -200,44 +202,56 @@ function layout.arrange(screen)
         local p = layout.parameters(nil, screen)
 
         local useless_gap = p.useless_gap
-        local ugap_edge = 0
+        local ugap_edge_left, ugap_edge_top, ugap_edge_right, ugap_edge_bottom = useless_gap, useless_gap, useless_gap, useless_gap
+        if theme.useless_gap_edge_left ~= nil then
+            ugap_edge_left = theme.useless_gap_edge_left
+        end
+        if theme.useless_gap_edge_top ~= nil then
+            ugap_edge_top = theme.useless_gap_edge_top
+        end
+        if theme.useless_gap_edge_right ~= nil then
+            ugap_edge_right = theme.useless_gap_edge_right
+        end
+        if theme.useless_gap_edge_bottom ~= nil then
+            ugap_edge_bottom = theme.useless_gap_edge_bottom
+        end
+        --local useless_gap = 0
         local ugap_d = math.floor(useless_gap/2)
         local ugap_m = math.fmod(useless_gap,2)
         local m = 0
 
-
         p.geometries = setmetatable({}, {__mode = "k"})
         layout.get(screen).arrange(p)
         for c, g in pairs(p.geometries) do
             if g.x ~= p.workarea.x
                 then m = ugap_d
-                else m = ugap_edge
+                else m = ugap_edge_left - useless_gap
             end
             g.x = g.x + m
             g.width = g.width - m
 
             if (g.x + g.width) ~= (p.workarea.x + p.workarea.width)
                 then m = ugap_d + ugap_m
-                else m = ugap_edge
+                else m = ugap_edge_right - useless_gap
             end
             g.width = g.width - m
 
-            if g.y ~= p.workarea.y 
+            if g.y ~= p.workarea.y
                 then m = ugap_d
-                else m = ugap_edge
+                else m = ugap_edge_top - useless_gap
             end
             g.y = g.y + m
             g.height = g.height - m
 
             if (g.y + g.height) ~= (p.workarea.y + p.workarea.height)
                 then m = ugap_d + ugap_m
-                else m = ugap_edge
+                else m = ugap_edge_bottom - useless_gap
             end
             g.height = g.height - m
 
-            g.width = g.width - c.border_width * 2 
+            g.width = g.width - c.border_width * 2
             g.height = g.height - c.border_width * 2
-            
+
             c:geometry(g)
         end
         arrange_lock = false
```

actionless:
```lua
theme.useless_gap = dpi(4)
theme.useless_gap_edge_left = 0
theme.useless_gap_edge_bottom = 0
theme.useless_gap_edge_right = 0
```

2017-05-17--1495044389_3840x2160_scrot

and3rson:
Bump... :v

actionless:
@and3rson does that mean you're volunteering to resurrect the PR? if so, you could start from applying this patch (https://github.com/awesomeWM/awesome/pull/596#discussion_r117071391) and resolving merge conflict with the current master

psychon:
Closing for lack of activity from the author for almost two years. The git branch is still there, so other PRs can be opened based on this from anyone who wants to continue this.
(Sorry @and3rson, I guess this is the exact opposite of what you wanted to achieve)

PR #597, opened 2015-12-15
feat(lib: awful: spawn: with_line_callback): make callbacks optional

actionless:


psychon:
Sorry that I missed that the `done_callback` was still necessary. Thanks for catching that.

psychon:
I'd squash the two commits, but this looks fine to me.

actionless:
rebased

Issue #602, opened 2015-12-20
Timestamp with awesome warnings/errors


blueyed:
Looking at e.g. `~/.xsession-errors` it would be useful if the messages (warnings etc) from awesome would be prefixed by a timestamp.

This would probably mean to change `_warn` and `_fatal` in `common/util.c` and `luaA_warn` in `luaa.h`.
The latter should then be used e.g. in `awful.util.deprecate` etc.

What do you think?

psychon:
Seems fine to me. I'd implement this in both C and Lua instead of exporting some function to Lua for this. Should I give this as try, or do you want to do that?

blueyed:
> Should I give this as try

Yes, please do.

PR #606, opened 2015-12-22
Add timestamps to messages on stderr (#602)

psychon:


blueyed:
The format should be %Y-%m-%d for the date.

And then it might also make sense to have it at the beginning of the line, but I am not sure about that.

psychon:
Old version: 6ef19b34cc38d30ba6ab22, new one: af63a2b31f776fb0c691d5ba8fef211b0bf66cf5

Issue #598, opened 2015-12-15
Regression: Multiple separate X11 screens setup (Zaphod, no Xinerama) fails; WORKAROUND: start multiple instances


Manfred-Knick:
REGRESSION
from
. . . x11-wm/awesome-3.4.15:0
to
. . . x11-wm/awesome-3.5.6-r1:0

My setup:
2 graphic "card"s with 1+3=4 monitors
= four separate X11 screens
( c.f. xorg.conf.d -> Attachment ).

Layout "Four":
Only on the fist screen a full working desktop appears.
The other three remain black.
In principle, all four X11 screens get addressed;
at least the "X" sign mouse cursor flows over the rest of the four.

Cross-Check:
Plain X works with (c)twm, as well as fluxbox, xfce, ... on all four screens.

Manfred-Knick:
[Device-01.intel.conf.TXT](https://github.com/awesomeWM/awesome/files/63008/Device-01.intel.conf.TXT)
[Device-11.nvidia.conf.TXT](https://github.com/awesomeWM/awesome/files/63009/Device-11.nvidia.conf.TXT)
[Monitor.conf.TXT](https://github.com/awesomeWM/awesome/files/63014/Monitor.conf.TXT)
[Screen.conf.TXT](https://github.com/awesomeWM/awesome/files/63012/Screen.conf.TXT)

[Files.conf.TXT](https://github.com/awesomeWM/awesome/files/63010/Files.conf.TXT)
[InputDevice_Keyboard.conf.TXT](https://github.com/awesomeWM/awesome/files/63016/InputDevice_Keyboard.conf.TXT)
[InputDevice_Mouse.conf.TXT](https://github.com/awesomeWM/awesome/files/63015/InputDevice_Mouse.conf.TXT)
[ServerFlags.conf.TXT](https://github.com/awesomeWM/awesome/files/63011/ServerFlags.conf.TXT)

[ServerLayout.conf.TXT](https://github.com/awesomeWM/awesome/files/63013/ServerLayout.conf.TXT)

Manfred-Knick:
Added DEBUG into rc.lua:
. . . naughty.notify({ preset = naughty.config.presets.critical,
. . . . . . title = "DEBUG : rc.lua",
. . . . . . text = "screen.count() = " .. screen.count() })
RESULT:
. . . screen.count() = 1
EXPECTED:
. . . screen.count() = 4

QUERY with xrandr:
. . . xrandr -q --screen 0.n
with
[...]

PR #607, opened 2015-12-23
Fix #599 (issue with focus on multiple instances)

### LINKED ISSUE #599: focus issue with multiple awesome instances
If I run multiple instances of awesome on separate X screens:

DISPLAY=:0.0 awesome &
DISPLAY=:0.1 awesome

it mostly works, except the focus is a little wonky. If there are multiple windows (i.e. multiple terminals) mapped on the current tag of the root window I'm moving into, if I cross one of those windows, then focus is transferred, and the expected instance of awesome gets the key events to manage windows / launch applications. But if there is only one window (or none) - basically so there isn't an enter/leave/enter chain of events (or some other event I don't know of), then the screen the mouse moves into doesn't transfer focus to that window now under the mouse. Additionally, if there are no windows on the currently displayed tag on the newly-entered screen, the hot keys are still applied to the original screen, although I can click on the awesome menu bar and launch an item, at which point focus is where you would expect.

Thinking the last would be the easiest, I tried adding a check in event_handle_enternotify in event.c and regrabbing keys as happened in the mapping notify when the enter notify window is the root window. Apparently I missed something about the internal state of things (or am testing for the wrong thing), as that had no effect. Any clues / docs about what to change, please let me know and I will try it. For the focus problem when there is a window, I suspect something similar where there's a leavenotify but because the other process gets the enternotify since the root window has changed as it's a different screen, there's some state of focus not being updated.

thanks!

kdt3rd:
This fixes #599 as far as I'm concerned - meaning it changes behavior to fix the issues I was seeing with running multiple instances of awesome on separate screens.

The awesomeConfig.cmake I accidentally checked in prior to this branch, and now github wants it unconditionally, and I can't seem to deselect it. However, if you do require lua v5.3, it does fix an issue with FindLua when there are multiple versions of lua installed on the system (since lua doesn't really do pkg config very well)

### DIFF
--- a/event.c
+++ b/event.c
@@ -580,6 +580,12 @@ event_handle_enternotify(xcb_enter_notify_event_t *ev)
         }
         lua_pop(L, 1);
     }
+    else if (ev->event == globalconf.screen->root) {
+        /* for when there are multiple X screens with awesome running
+         * separate instances, reset focus
+         */
+        globalconf.focus.need_update = true;
+    }
 }
 
 /** The focus in event handler.


psychon:
The second commit looks fine to me (glad to hear that my idea works).

The first one makes travis fail to build this, because neither lua 5.1, lua 5.2 nor luajit work anymore while they did before. :-) I consider the problem with multiple installed lua versions to be a bug in CMake / FindLua and I don't know what to do about it. See #361 and https://cmake.org/Bug/view.php?id=15756.

I think we can just cherry-pick the second commit and be done with this PR (if @blueyed agrees)?

Manfred-Knick:
@ kdt3rd : against which version of awesome was this patch created? It's not awesome-3.5.6 ...
Perhaps against a HEAD tree pulled from git as mentioned in #599 ? 
Thanks in advance for pointing this out.

blueyed:
@Manfred-Knick 
The patch is against Git master.

@kdt3rd 
Please remove 41dcdea from the PR, so we can see that it passes the builds.

Manfred-Knick:
Confirmation:
. . . back-ported into awesome-3.5.6 .
Please c.f. comment *WORKSFORME* in #598 .

kdt3rd:
I don't seem to easily be able to remove the initial patch against awesomeConfig.cmake - I wanted to remove it earlier, but realized I'd already pushed it to my remote. I guess either make the patch elsewise and reject this pull request, or if you want, I'll clean up my fork and re-submit.

thanks

blueyed:
> I'll clean up my fork and re-submit

Yes, you should be able to use `git rebase -i origin/master` to remove the patch/commit and then force-push it to your branch again.  That should update this PR.

psychon:
(Force push: I guess that this would be `git push --force origin multi_instance` for you. However, `git push git://github.com/kdt3rd/awesome.git +multi_instance` should work in any case (the `+` means "force push, even if this means that changes are lost))

kdt3rd:
OK, thanks, I've patched up that branch, and Travis seems happy now. thanks for the patience

psychon:
Cherry-picked into 3.5.
Closing this since it was apparently merged with a slightly edited commit message, as @thmo noticed.

PR #608, opened 2015-12-23
Replace various calls to warn() with luaA_warn()

psychon:
luaA_warn() prints a Lua backtrace and thus generates more useful output. warn()
should only be used in awesome-internal places (e.g. receiving an error from the
X11 server).

### DIFF
--- a/common/luaobject.c
+++ b/common/luaobject.c
@@ -254,7 +254,7 @@ signal_object_emit(lua_State *L, signal_array_t *arr, const char *name, int narg
             luaA_dofunction(L, nargs, 0);
         }
     } else
-        warn("Trying to emit unknown signal '%s'", name);
+        luaA_warn(L, "Trying to emit unknown signal '%s'", name);
 
     /* remove args */
     lua_pop(L, nargs);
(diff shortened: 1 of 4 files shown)


common/luaobject.c
```diff
@@ -304,8 +304,10 @@ luaA_object_emit_signal(lua_State *L, int oud,
             lua_remove(L, - nargs - nbfunc - 2 + i);
             luaA_dofunction(L, nargs + 1, 0);
         }
-    } else
-        warn("Trying to emit unknown signal '%s'", name);
+    } else {
+        luaA_warn(L, "Trying to emit unknown signal '%s'", name);
+        return;
```
blueyed: That is a behaviour change, but seems to be what was intended.

psychon: Well, what behavior really changes? When you do `c:emit_signal("foo")`, the "no such signal"-message is now only printed once instead of twice. I don't think this has any other effect than that. (But yeah, it's a behavior change)

blueyed: Ok.

PR #609, opened 2015-12-23
Fix client icons

psychon:
Since commit ea94f7f7e64fd2b, the code has a preferred icon size and it tries to find the closest matching icon. However, the default size of 0 and thus this never returned any icon at all (no icon matches better than an icon of size 0x0).

Fix this by saying "every icon is better than no icon at all".

@lukash Any complaints?

Edit: This fixes chromium's icon for me. However, I was thinking that before this all icons are broken, but that's not actually the case. My thunderbird does have an icon...

### DIFF
--- a/ewmh.c
+++ b/ewmh.c
@@ -694,7 +694,7 @@ ewmh_window_icon_from_reply(xcb_get_property_reply_t *r, uint32_t preferred_size
         bool better_because_bigger =  found_icon_too_small && size > found_size;
         bool better_because_smaller = found_icon_too_large &&
             size >= preferred_size && size < found_size;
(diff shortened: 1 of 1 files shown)


blueyed:
:+1: 

This fixes also the icon for urxvt for me (within cyclefocus).

lukash:
I'm not particularly convinced daff5cd makes it more readable, but that's a matter of personal preference...

Otherwise, sorry for the omission. Seems I forgot to test the default after setting it to 0. Proceed to merge :)

psychon:
My intention with daff5cd was to break up the complicated `if` into smaller parts... too much nested logic for my brain.

PR #612, opened 2015-12-29
awful.client.idx: Avoid error when called with a floating client

Elv13:
Before, it was the caller job to make sure the client wasn't floating.

This limitation is unnecessary. awful.client.idx now return nil instead
of an error. awful.rules setting the master width factor are now
foolproof.


Elv13:
Updated, fix @psychon concerns

blueyed:
:+1: from me, too.

@Elv13 
Please merge it yourself, to make sure it's ready and in a good order.

PR #613, opened 2015-12-29
typo: Replace 'wibox' with 'context' where relevant

Elv13:


psychon:
Looks fine to me. Thanks for also catching the mistakes in the docs & sorry that I introduced them.

PR #618, opened 2016-01-06
feat(awful: titlebar): make tooltips optional

actionless:


psychon:
Fine by me. What about documentation for this?
Perhaps also add an optional argument to the various functions, but that would be harder to do....

actionless:
added docstring

I think adding this option as an arg is too much.

btw what do you think about changing all the functions to accept only one table with all the args for hypothetic awesome 4.0?
it could make easier adding new things without breaking the configs

psychon:
We can do so where it makes sense. Doing that everywhere just makes it unclear what the possible arguments are and means that typos are possible.

PR #619, opened 2016-01-06
Spelling fix in comment

deavmi:


blueyed:
Thanks.
While at it, this should also add punctuation (a dot at the end), but that can me done when merging/cherry-picking.

deavmi:
Okay, great! Thanks by the way. ;)

PR #621, opened 2016-01-09
surface: Warn when loading failed

Elv13:
Otherwise, LGI will crash awesome with an unprotected call

This can happen, for example, if the file exist, but is not
readable or corrupted.

The user need to check if the surface exist before using it,
but without this patch, they wont know where to look, awesome
will have crashed with a binary backtrace.

### DIFF
--- a/draw.c
+++ b/draw.c
@@ -239,11 +239,8 @@ draw_load_image(lua_State *L, const char *path)
     GdkPixbuf *buf = gdk_pixbuf_new_from_file(path, &error);
 
     if (!buf) {
-        luaL_where(L, 1);
-        lua_pushstring(L, error->message);
-        lua_concat(L, 2);
+        luaA_warn(L, error->message);
         g_error_free(error);
-        lua_error(L);
         return NULL;
     }
 
--- a/lib/gears/surface.lua
+++ b/lib/gears/surface.lua
@@ -19,6 +19,8 @@ end
 local surface = { mt = {} }
 local surface_cache = setmetatable({}, { __mode = 'v' })
 
+local dummy_surface = nil
+
 --- Try to convert the argument into an lgi cairo surface.
 -- This is usually needed for loading images by file name.
 function surface.load_uncached(_surface)
@@ -37,6 +39,16 @@ function surface.load_uncached(_surface)
     if type(_surface) == "string" then
         file = _surface
         _surface = capi.awesome.load_image(file)
+        if not _surface then
+            io.stderr:write("W: The requested surface failed to load\n", debug.traceback())
+
+            -- Avoid calling cairo.Surface with nil, it will crash Awesome
+            if not dummy_surface then
+                dummy_surface = cairo.ImageSurface(cairo.Format.ARGB32, 1, 1)
+            end
+
+            return dummy_surface
+        end
     end
     -- Everything else gets forced into a surface
     _surface = cairo.Surface(_surface, true)


psychon:
Uhm... any example on the crash you are seeing/fixing? Adding `require("gears.surface")("does not exist")` at the end of the rc.lua results in:
```
error while running function
stack traceback:
	[C]: in function 'load_image'
	/home/psychon/projects/awesome/build/lib/gears/surface.lua:39: in function </home/psychon/projects/awesome/build/lib/gears/surface.lua:24>
	(...tail calls...)
	./awesomerc.lua:539: in main chunk
error: /home/psychon/projects/awesome/build/lib/gears/surface.lua:39: Failed to open file 'does not exist': No such file or directory
```
(And awesome falls back to the "next" config)

Elv13:
Well, given I was developing directly in awesomerc.lua, there was no fallback :P 

But anyway, I still think loading the next config just because an image failed to load is overreacting. In my case, it was a png from a .desktop where the prefix was not accessible by the user account I used to test some patches. The "unprotected lua call" is because `cairo.Surface()` is called with `nil` (without the dummy surface) and it doesn't produce a backtrace saying why (as you can see in your own example)

psychon:
Ok, so I see two different issues here.

One is the general "we are possibly overreacting on config errors", but I have no idea what to do about it. We already had this discussion somewhere about textbox complaining loudly about invalid markup by throwing an error. I guess we should come up with some general policy on when it is ok for Lua code to throw errors.
I also guess no one would complain if `gears.colors.parse_color(function() end)` throws an error "string expected, got function". Why is `gears.colors.parse_color("The house is red")` not allowed to throw errors?
I've been writing (too much?) Java code lately where such things are signaled by throwing an exception and I guess I write Lua in a similar way. By this train of thought, the issue is not "gears.surface throws errors", but "menubar does not catch errors" (no idea which code @Elv13 was using exactly, but .desktop sounds like menubar to me; I bet he was not using this, but something else). My thinking is "we can't make everything bulletproof by checking for all cases and only printing warnings, so errors can happen anyway, so 'the code'  has to handle them anyway". If we make this "image not found" thing a fatal error, too, any time this causes an issue because `pcall()` was not used, fixing that will also fix cases where e.g. a function is used instead of a file name.
Do you understand what I am saying?

For this issue, I'd propose to add `gears.surface.load_safe` (anyone has a good idea for a name?) that returns `nil` or some default surface or something like that when loading fails.

Elv13:
While this opinion is quite valid, I would argue that there is a zillion call to `gears.surface` and none is protected in `pcall`. `pcall` is annoying to use for simple case like this.

This error does not really happen because if "invalid" lua input, but deeper in the code when there is an error processing that command. It is the **only** place in all of Awesome were the `C` code was voluntarily crashing the load. `lua_error` is used nowhere else (LGI use it for type exceptions).

Can we have a middle ground for this? Like returning `nil` when the surface fail to load and returning a 1x1 surface if gears.surface is called with `nil`? Then, if the users don't check nil, then its their problem. If we start to say "all API can fail, it's your problem", then every single method call would have to be `pcall`ed, and this someone have to be totally disconnected from reality if you thing people will ever do that.

psychon:
Hm. How about a new `default` argument to `gears.surface.load()`? If that argument is `nil`, it is replaced with an ImageSurface in an error state (why would you want a `1x1` surface instead?)?

Something like the following (*NOT* the actual implementation, but just as a proposal for how it should behave):
```lua
function gears.surface.load_file(path, default) -- This should replace the "load file name"-case in the current code, the rest stays as-is
  local success, ret = pcall(function() return capi.awesome.load_image(path) end)
  if not success then
    -- Create a cairo surface in an error state of "FILE_NOT_FOUND"
    ret = require("lgi").cairo.ImageSurface.create_from_png("")
  end
  return ret
end
```
(And we should fix the imagebox widget to reject images in an error-state...)

Should I propose a PR doing this?

Elv13:
> Should I propose a PR doing this?

Feel free

psychon:
Cannot do so right now, but I'd do the following (however, right now this is completely untested):
```diff
diff --git a/draw.c b/draw.c
index d9795bf..22e695b 100644
--- a/draw.c
+++ b/draw.c
@@ -229,23 +229,17 @@ draw_dup_image_surface(cairo_surface_t *surface)
 /** Load the specified path into a cairo surface
  * \param L Lua state
  * \param path file to load
+ * \param error A place to store an error message, if needed
  * \return A cairo image surface or NULL on error.
  */
 cairo_surface_t *
-draw_load_image(lua_State *L, const char *path)
+draw_load_image(lua_State *L, const char *path, GError *error)
 {
-    GError *error = NULL;
     cairo_surface_t *ret;
     GdkPixbuf *buf = gdk_pixbuf_new_from_file(path, &error);
 
-    if (!buf) {
-        luaL_where(L, 1);
-        lua_pushstring(L, error->message);
-        lua_concat(L, 2);
-        g_error_free(error);
-        lua_error(L);
+    if (!buf)
         return NULL;
-    }
 
     ret = draw_surface_from_pixbuf(buf);
     g_object_unref(buf);
diff --git a/draw.h b/draw.h
index 7ad7941..4c7e9da 100644
--- a/draw.h
+++ b/draw.h
@@ -70,7 +70,8 @@ a_iso2utf8(const char *str, ssize_t len, char **dest, ssize_t *dlen)
 
 cairo_surface_t *draw_surface_from_data(int width, int height, uint32_t *data);
 cairo_surface_t *draw_dup_image_surface(cairo_surface_t *surface);
-cairo_surface_t *draw_load_image(lua_State *L, const char *path);
+#include <gdk-pixbuf/gdk-pixbuf.h> /* FIXME! */
+cairo_surface_t *draw_load_image(lua_State *L, const char *path, GError *error);
 
 xcb_visualtype_t *draw_find_visual(const xcb_screen_t *s, xcb_visualid_t visual);
 xcb_visualtype_t *draw_default_visual(const xcb_screen_t *s);
diff --git a/lib/awful/menu.lua b/lib/awful/menu.lua
index ca8a90f..04751cd 100644
--- a/lib/awful/menu.lua
+++ b/lib/awful/menu.lua
@@ -20,6 +20,7 @@ local beautiful = require("beautiful")
 local dpi = require("beautiful").xresources.apply_dpi
 local object = require("gears.object")
 local surface = require("gears.surface")
+local gdebug = require("gears.debug")
 local cairo = require("lgi").cairo
 local setmetatable = setmetatable
 local tonumber = tonumber
@@ -523,7 +524,7 @@ function menu.entry(parent, args)
     local margin = wibox.layout.margin()
     margin:set_widget(label)
     if args.icon then
-        icon = surface.load(args.icon)
+        icon = surface.load_and_print_errors(args.icon)
     end
     if icon then
         local iw = icon:get_width()
@@ -558,7 +559,7 @@ function menu.entry(parent, args)
     if type(args.cmd) == "table" then
         if args.theme.submenu_icon then
             submenu = wibox.widget.imagebox()
-            submenu:set_image(surface.load(args.theme.submenu_icon))
+            submenu:set_image(args.theme.submenu_icon)
         else
             submenu = wibox.widget.textbox()
             submenu:set_font(args.theme.font)
diff --git a/lib/awful/widget/taglist.lua b/lib/awful/widget/taglist.lua
index 46ac8b7..c043b26 100644
--- a/lib/awful/widget/taglist.lua
+++ b/lib/awful/widget/taglist.lua
@@ -107,11 +107,7 @@ function taglist.taglist_label(t, args)
         text = text .. "</span>"
     end
     if not taglist_disable_icon then
-        if tag.geticon(t) and type(tag.geticon(t)) == "image" then
-            icon = tag.geticon(t)
-        elseif tag.geticon(t) then
-            icon = surface.load(tag.geticon(t))
-        end
+        icon = surface.load_and_print_errors(tag.geticon(t))
     end
 
     return text, bg_color, bg_image, not taglist_disable_icon and icon or nil
diff --git a/lib/gears/surface.lua b/lib/gears/surface.lua
index c7a96d7..e25b969 100644
--- a/lib/gears/surface.lua
+++ b/lib/gears/surface.lua
@@ -9,6 +9,7 @@ local setmetatable = setmetatable
 local type = type
 local capi = { awesome = awesome }
 local cairo = require("lgi").cairo
+local gdebug = require("gears.debug")
 
 -- Keep this in sync with build-utils/lgi-check.sh!
 local ver_major, ver_minor, ver_patch = string.match(require('lgi.version'), '(%d)%.(%d)%.(%d)')
@@ -19,13 +20,26 @@ end
 local surface = { mt = {} }
 local surface_cache = setmetatable({}, { __mode = 'v' })
 
+local function get_default(default)
+    if type(default) ~= nil then
+        default = cairo.ImageSurface.create_from_png("")
+    end
+    return default
+end
+
 --- Try to convert the argument into an lgi cairo surface.
 -- This is usually needed for loading images by file name.
-function surface.load_uncached(_surface)
+-- @param _surface The surface to load
+-- @param default The default value to return on error, when nil, then a surface
+-- in an error state is returned.
+-- @return The loaded surface, or the replacement default
+-- @return An error message
+-- @see surface.load For a version caching results
+function surface.load_uncached(_surface, default)
     local file
-    -- Nil is not changed
+    -- Nil gets the default applied
     if not _surface then
-        return nil
+        return get_default(default)
     end
     -- Remove from cache if it was cached
     surface_cache[_surface] = nil
@@ -35,8 +49,12 @@ function surface.load_uncached(_surface)
     end
     -- Strings are assumed to be file names and get loaded
     if type(_surface) == "string" then
+        local err
         file = _surface
-        _surface = capi.awesome.load_image(file)
+        _surface, err = capi.awesome.load_image(file)
+        if not _surface then
+            return get_default(default), err
+        end
     end
     -- Everything else gets forced into a surface
     _surface = cairo.Surface(_surface, true)
@@ -47,14 +65,34 @@ function surface.load_uncached(_surface)
     return _surface
 end
 
-function surface.load(_surface)
+--- Try to convert the argument into an lgi cairo surface.
+-- This is usually needed for loading images by file name and uses a cache.
+-- @param _surface The surface to load
+-- @param default The default value to return on error, when nil, then a surface
+-- in an error state is returned.
+-- @return The loaded surface, or the replacement default
+-- @return An error message
+-- @see surface.load_uncached For a version not caching results
+function surface.load(_surface, default)
     if type(_surface) == "string" then
         local cache = surface_cache[_surface]
         if cache then
             return cache
         end
     end
-    return surface.load_uncached(_surface)
+    return surface.load_uncached(_surface, default)
+end
+
+--- TODO: Better name, docs
+-- @return[1] The loaded surface
+-- @return[2] nil on error
+function surface.load_and_print_errors(_surface)
+    local result, error = surface.load(_surface, false)
+    if result then
+        return result
+    end
+    gdebug.print_error("Failed to load '" .. tostring(_surface) .. "': " .. tostring(error))
+    return nil
 end
 
 function surface.mt:__call(...)
diff --git a/lib/gears/wallpaper.lua b/lib/gears/wallpaper.lua
index 277ce2e..f1cddc1 100644
--- a/lib/gears/wallpaper.lua
+++ b/lib/gears/wallpaper.lua
@@ -104,7 +104,8 @@ end
 --   gears.color. The default is black.
 function wallpaper.centered(surf, s, background)
     local geom, cr = wallpaper.prepare_context(s)
-    local surf = surface(surf)
+    local surf = surface.load_and_print_errors(surf)
+    if not surf then return end
     local background = color(background)
 
     -- Fill the area with the background
@@ -133,7 +134,9 @@ function wallpaper.tiled(surf, s, offset)
         cr:translate(offset.x, offset.y)
     end
 
-    local pattern = cairo.Pattern.create_for_surface(surface(surf))
+    local surf = surface.load_and_print_errors(surf)
+    if not surf then return end
+    local pattern = cairo.Pattern.create_for_surface(surf)
     pattern.extend = cairo.Extend.REPEAT
     cr.source = pattern
     cr.operator = cairo.Operator.SOURCE
@@ -149,7 +152,8 @@ end
 -- @param offset This can be set to a table with entries x and y.
 function wallpaper.maximized(surf, s, ignore_aspect, offset)
     local geom, cr = wallpaper.prepare_context(s)
-    local surf = surface(surf)
+    local surf = surface.load_and_print_errors(surf)
+    if not surf then return end
     local w, h = surface.get_size(surf)
     local aspect_w = geom.width / w
     local aspect_h = geom.height / h
@@ -181,8 +185,9 @@ end
 --   gears.color. The default is black.
 function wallpaper.fit(surf, s, background)
     local geom, cr = wallpaper.prepare_context(s)
-    local surf = surface(surf)
+    local surf = surface.load_and_print_errors(surf)
     local background = color(background)
+    if not surf then return end
 
     -- Fill the area with the background
     cr.operator = cairo.Operator.SOURCE
diff --git a/lib/naughty/core.lua b/lib/naughty/core.lua
index ee41e72..97051c8 100644
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -562,13 +562,7 @@ function naughty.notify(args)
         end
 
         -- is the icon file readable?
-        local success, res = pcall(function() return surface.load_uncached(icon) end)
-        if success then
-            icon = res
-        else
-            io.stderr:write(string.format("naughty: Couldn't load image '%s': %s\n", tostring(icon), res))
-            icon = nil
-        end
+        local icon = surface.load_and_print_errors(icon)
 
         -- if we have an icon, use it
         if icon then
diff --git a/lib/wibox/widget/imagebox.lua b/lib/wibox/widget/imagebox.lua
index 14eb131..5f7f882 100644
--- a/lib/wibox/widget/imagebox.lua
+++ b/lib/wibox/widget/imagebox.lua
@@ -7,6 +7,7 @@
 
 local base = require("wibox.widget.base")
 local surface = require("gears.surface")
+local gdebug = require("gears.debug")
 local setmetatable = setmetatable
 local pairs = pairs
 local type = type
@@ -72,17 +73,16 @@ end
 --- Set an imagebox' image
 -- @param image Either a string or a cairo image surface. A string is
 --   interpreted as the path to a png image file.
+-- @return true on success, false if the image cannot be used
 function imagebox:set_image(image)
     local image = image
 
     if type(image) == "string" then
-        local success, result = pcall(surface.load, image)
-        if not success then
-            print("Error while reading '" .. image .. "': " .. result)
+        image = surface.load_and_print_errors(image)
+        if not image then
             print(debug.traceback())
             return false
         end
-        image = result
     end
 
     image = surface.load(image)
diff --git a/luaa.c b/luaa.c
index 446cc7e..1ef4b0a 100644
--- a/luaa.c
+++ b/luaa.c
@@ -147,16 +147,24 @@ luaA_restart(lua_State *L)
 /** Load an image from a given path.
  *
  * @param name The file name.
- * @return A cairo surface as light user datum.
+ * @return[1] A cairo surface as light user datum.
+ * @return[2] nil
+ * @treturn[2] string Error message
  * @function load_image
  */
 static int
 luaA_load_image(lua_State *L)
 {
+    GError *error = NULL;
     const char *filename = luaL_checkstring(L, 1);
-    cairo_surface_t *surface = draw_load_image(L, filename);
-    if (!surface)
-        return 0;
+    cairo_surface_t *surface = draw_load_image(L, filename, error);
+    if (!surface) {
+        lua_pushnil(L);
+        lua_pushstring(L, error->message);
+        g_error_free(error);
+        return 2;
+    }
+
     /* lua has to make sure to free the ref or we have a leak */
     lua_pushlightuserdata(L, surface);
     return 1;
```
Sorry for doing so much stuff in one patch, I also went through all callers of `gears.surface` and tried to fix them up.
Any good ideas on how to call `surface.load_and_print_errors()`? Should `surface()` call that function instead of `load`?

PR #622, opened 2016-01-09
Add support for short/long colors

psychon:
This adds support to gears.color.parse_color to parse things like "#fff" (one
character per color component, without alpha) and "#ffff0000ffff0000" (four
characters per component, with alpha).

This makes sense on its own, but should also help with
https://github.com/awesomeWM/awesome/issues/585.


psychon:
Rebased this onto master and added a new commit which adds support for named colors (e.g. "red"). Apparently Pango offers an API for this, let's use it! :-)

PR #625, opened 2016-01-14
fix(awful: hotkeys): caching on multiscreen setup

actionless:


psychon:
Could you also change `local cached_wiboxes = {{}}` into `local cached_wiboxes = {}` (line 93).
No reason to treat the first screen special here.

actionless:
done

PR #627, opened 2016-01-15
Warnings when a surface fails to load (see #621)

psychon:
This is a step towards less Lua errors. In the end, the visible behavior of this change is quite similar to #621, even though this started differently... I keep stealing @Elv13's changes. Sorry!


Elv13:
A little more invasive than my change, but for the better

+1

Issue #616, opened 2016-01-03
"arrange" is called when focus change


Elv13:
In `awful/layout/init.lua`, there is:

```lua
capi.client.connect_signal("focus", function(c) layout.arrange(c.screen) end)
```

This cause slowdown and full redraw of some clients. For example, then using `urxvtd` (daemon with multiple terminal and internal IPC between terminals), this cause all terminal to be redrawn. On my old computer, this took more time than it took to focus the next one when moving the mouse across the screen, causing visual glitches for a few milliseconds (with urxvt/aterm `tint` option).

I am also playing with a port of the `awful.layout` system to use `wibox.hierarchy`, this cause many pointless calls (but they are delayed, some in the end don't affect performance that much).

Commenting the line cause no immediate issues for me, so is it really needed?

psychon:
Removing that line breaks the magnifier layout.

Why does re-layouting stuff cause redraws? Every window should get the position that it already had before. In the C code, `client_resize()` doesn't do anything at all if the geometry doesn't change (look at the code, only if something changed does it call `client_resize_do()`).

PR #629, opened 2016-01-16
client: Add swapped, raised and lowered signals

Elv13:
#### Hello world, beware: here I come

This is the first pull request of a rather long stream. Eventually, if everything get merged, it will add stateful layout support to Awesome, including tabs, fancy extension widgets for third party module, limitless flexibility and dynamic layout (split where you want when you want).

The change itself will at first add a parallel implementation of most `awful.layout.suit`. It will be disabled by default and only enabled when the use `require` the system in `rc.lua`. It provide a drop-in replacement of the existing system. By default, it will <s>only add bugs</s> provide nothing new compared with the current system. In the first iteration, any cool stuff will be provided by extensions. The change is already scary enough in itself, no need to add a million line of <s>cool hacks</s> code.

The first few pull requests will make low level to existing APIs to increase flexibility. The second will add new helper code for the new layout system. The last batch, once the first 2 are merged, will add the new suits. The "final" version is available in the "dynamic_layout2" branch of my fork. It is good enough to be used in production as long as you don't play to hard with it. There is still some bugs, mainly in the most neglected layout suits and in the mouse resize code. Overall, I use it, it's stable-ish for daily use if you set your expectations low. If it end up not being merged, it I will create yet another external module called `leaning` for the code. The minimal code to get stateful layouts working is about +300 line. However, with all (I hope) the corner case handled it grows a lot larger, around +2.5k. Then again, it is mostly boring new code, not critical code path changes.

#### This Pull request

This allow layout "arrange" to be called less often and react on
the cause of the change itself rather than it's consequences
(usually, the "focus" signal).

Previously, the layout were re-arranged everytime the focus changed.
Now, with "raised" and "lowered", it require less "arrange".

"swapped" allow smarted layouts. Currently, swapped cause a full
re-arrange. It re-read the "index" list from scratch and create
a "new" layout. With "swapped", incremental layout changes are
possible.

Fixes https://github.com/awesomeWM/awesome/issues/616

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1602,6 +1602,17 @@ luaA_client_swap(lua_State *L)
         *ref_swap = c;
 
         luaA_class_emit_signal(L, &client_class, "list", 0);
+
+        luaA_object_push(L, swap);
+        lua_pushboolean(L, true);
+        luaA_object_emit_signal(L, -4, "swapped", 2);
+        lua_pop(L, 2);
+
+        luaA_object_push(L, swap);
+        luaA_object_push(L, c);
+        lua_pushboolean(L, false);
+        luaA_object_emit_signal(L, -3, "swapped", 2);
+        lua_pop(L, 3);
     }
 
     return 0;
@@ -1688,7 +1699,13 @@ static int
 luaA_client_raise(lua_State *L)
 {
     client_t *c = luaA_checkudata(L, 1, &client_class);
+
+    /* Avoid sending the signal if nothing was done */
+    if (c->transient_for == NULL && globalconf.stack.tab[globalconf.stack.len-1] == c)
+        return 0;
+
     client_raise(c);
+
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


Elv13:
Updated

Elv13:
Updated again

Elv13:
Updated, add the `transient_for` hack suggested by @psychon

psychon:
Fine with me: :+1:

objects/client.c
```diff
@@ -1602,6 +1602,17 @@ luaA_client_swap(lua_State *L)
...
+        luaA_object_push(L, swap);
+        lua_pushboolean(L, true);
+        luaA_object_emit_signal(L, -4, "swapped", 2);
+        lua_pop(L, 2);
+
+        luaA_object_push(L, swap);
+        luaA_object_push(L, c);
+        lua_pushboolean(L, false);
+        luaA_object_emit_signal(L, -3, "swapped", 2);
+        lua_pop(L, 3);
```
blueyed: Wouldn't a single signal call be enough?

Elv13: Not really, there is 2 clients involved in a swap, both need to be notified or it void the whole concept

blueyed: Ah, sure..!

blueyed:
:+1: 

Please merge it yourself.

PR #631, opened 2016-01-16
tag: Refactor awful.tag.setproperty to call setters

Elv13:
*WARNING* This introduce a minor API break as awful.tag.setscreen
arguments are now swapped for consistency

This allow to introduce logic for each properties and improve
awful.tag.add and execute logic when setting properties.

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -67,7 +67,7 @@ function tag.move(new_index, target_tag)
 
     for i=new_index < rm_index and new_index or rm_index, #tmp_tags do
         local tmp_tag = tmp_tags[i]
-        tag.setscreen(tmp_tag, scr)
+        tag.setscreen(scr, tmp_tag)
         tag.setproperty(tmp_tag, "index", i)
     end
 end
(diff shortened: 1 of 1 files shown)


Elv13:
Updated

psychon:
:+1:

blueyed:
> consistency

With what?  Isn't the tag the first argument always/usually?

Elv13:
> With what? Isn't the tag the first argument always/usually?

Unfortunaly, our consistency is very inconsistent. In `awful.tag`, all other setters are `(value, t)`.

I had a discussion on IRC with @psychon about this. I will leave it as it for this commit, then do a new pull request to standardize all modules to

`set_foo(t, value, other...)` and `get_foo(t)`

Currently, there is some `foo.set(t, val)` `foo.set(val, t)` `setfoo(t, val)` `setfoo(val, t)` `set_foo(t, val)` and some more. The pull request that will fix that will be messy and large, so I prefer to to that in another PR as it is totally unrelated beside that little commit chunk.

psychon:
Sounds convincing. I'm in for this. @blueyed ?

blueyed:
Sounds good.

But `tag.setscreen` should not get changed in this PR then in the first place, right?

Elv13:
> But tag.setscreen should not get changed in this PR then in the first place, right?

I will revert this later, as it would cause regression now if I didn't

blueyed:
Because it's used in some other PR like that?

Elv13:
> Because it's used in some other PR like that?

No, because anything calling `awful.tag.setproperty(t, "screen", 1)` would call `setscreen` with the wrong parameters. So it might brake existing code, while doing the change will only cause a deprecation warning. The deprecation warning wont go away when I rename the function `set_screen`, only the warning message will change from "use setscreen(val, t)" to "use set_screen(t, val)"

blueyed:
:+1:

PR #632, opened 2016-01-16
awful.layout: Split the layout parameters computation from 'arrange'

Elv13:
It is necessary to have it beforehand when creating layout objects
for unselected layouts.

In the current layout system, there is no layout object, but to allow
tabs and dynamic tagging features like ion3, layouts cannot be stateless.

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -14,6 +14,7 @@ local util = require("awful.util")
 local ascreen = require("awful.screen")
 local capi = {
     screen = screen,
+    mouse  = mouse,
     awesome = awesome,
     client = client,
     tag = tag
@@ -79,7 +80,7 @@ function layout.inc(i, s, layouts)
         local curlayout = layout.get(s)
         local curindex
         for k, v in ipairs(layouts) do
-            if v == curlayout then
+            if v == curlayout or curlayout._type == v then
                 curindex = k
                 break
             end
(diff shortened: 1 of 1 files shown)


Elv13:
Updated

lib/awful/layout/init.lua
```diff
@@ -109,6 +110,56 @@ function layout.set(_layout, t)
     tag.setproperty(t, "layout", _layout)
 end
 
+--- Get the layout parameters used for the screen
+--
+-- This should give the same result as "arrange", but without the "geometries"
+-- paramater, as this is computed during arranging
```
blueyed: s/paramater/parameter/ - and please add punctuation (a dot at the end).

lib/awful/layout/init.lua
```diff
@@ -109,6 +110,56 @@ function layout.set(_layout, t)
...
+--
+-- If "t" is given, "screen" is ignored, if none are given, the mouse screen is
+-- used
```
blueyed: Please use "`" instead of normal quotes when referring to arguments, and `@tparam` instead of `@param`.
Also, punctuation.

lib/awful/layout/init.lua
```diff
@@ -109,6 +110,56 @@ function layout.set(_layout, t)
...
+--
+-- @param[opt] t The tag to query
+-- @param[opt] screen The screen
+-- @return A table with the workarea (x, y, width, height), the screen
```
blueyed: `@treturn table` instead of `@return`.

lib/awful/layout/init.lua
```diff
@@ -109,6 +110,56 @@ function layout.set(_layout, t)
...
+--   geometry (x, y, width, height), the clients, the screen and call
+--   the "geometries" table with client as keys and geometry as value
```
blueyed: I don't get the following:

> and call the "geometries" table with client as keys and geometry as value"

lib/awful/layout/init.lua
```diff
@@ -109,6 +110,56 @@ function layout.set(_layout, t)
...
+function layout.parameters(t, screen)
+    if not screen then
+        screen = capi.mouse.screen
+    end
```
blueyed: `screen = screen or capi.mouse.screen` ?

psychon: I like my idea more (made as a comment on the commit; why does GitHub have so many different possibilities?). Just removing this part should be enough, because `tag.selected(nil)` does the right thing and uses `awful.screen.selected()` to get a screen.

Elv13:
Updated

psychon:
:+1: (Although I don't remember what my comments were last time...)
@bl

PR #633, opened 2016-01-16
layout: Support layout with a constructor

Elv13:
This allow the most basic kind of stateful layouts to be created.
It is now possible to have layout instances instead of global
stateless layout arrange functions.

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -34,6 +34,7 @@ local tag = { mt = {} }
 -- Private data
 local data = {}
 data.history = {}
+data.dynamic_cache = setmetatable({}, { __mode = 'k' })
 data.tags = setmetatable({}, { __mode = 'k' })
 
 -- History functions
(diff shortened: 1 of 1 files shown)


psychon:
How does this "compare to" `awful.layout.set`? We now have two functions for setting the layout.
Should `awful.layout.set` be deprecated? That would feel weird to me. I guess you want this in `awful.tag` so that your "setter magic" from the other commit works.

I would propose to have this new code in `awful.layout.set()` and just add a stub-function `awful.tag.setlayout` that calls `awful.layout.set`.

Elv13:
> How does this "compare to" awful.layout.set? We now have two functions for setting the layout.
Should awful.layout.set be deprecated? That would feel weird to me. I guess you want this in awful.tag so that your "setter magic" from the other commit works.

 `awful.layout.set` magically call  `awful.tag.setlayout`. I think it conceptually belong to tag and  `awful.layout.set` should be deprecated. `awful.tag` is where the other layout related accessors are anyway (`mwfact`, `nmaster', `ncol`)

> I would propose to have this new code in awful.layout.set() and just add a stub-function awful.tag.setlayout that calls awful.layout.set.

Actually, that function just call the other one, so I don't see any urgency to change this. I will deprecate in my future "API consistency" PR

Elv13:
Updated

psychon:
Ahhh, I now understand how this "magic" works that makes `awful.layout.set` call this new code.

Fine with me: :+1: (And I'm looking forward to some docs on the API that a stateful layout has to implement)

Elv13:
> And I'm looking forward to some docs on the API that a stateful layout has to implement

You saw it 3 weeks ago, I need to update it again, but its over a page long, half the total documentation length of 3.4, should be enough ;)

Keep in mind, there is still 5 pull requests before <s>insane</s> adventurous users can turn it on.

PR #637, opened 2016-01-17
Make wibox.widget.textbox:set_markup() throw Lua errors less often

### LINKED ISSUE #546: set_markup crashes awesome with invalid text
Can't an exception be caught and logged instead. Of course, the markup needs to be done right; but it makes no sense to crash and exit the WM.

```
stack traceback:
	[C]: in function 'error'
	...usr/local/share/awesome/lib/wibox/widget/textbox.lua:73: in function 'set_markup'
	/usr/local/share/awesome/lib/awful/tooltip.lua:135: in function 'set_markup'
	/etc/xdg/awesome/rc.lua:671: in function 'xxxx_notifier'
	/etc/xdg/awesome/rc.lua:684: in main chunk
error: ...usr/local/share/awesome/lib/wibox/widget/textbox.lua:73: Error on line 6: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity - escape ampersand as &amp;
```

psychon:
This will still throw Lua errors for malformed input (not valid utf8), but at least invalid markup is now handled differently. Apparently this is what people want, so they can have it...

Edit: Fixes #546

### DIFF
--- a/lib/awful/widget/common.lua
+++ b/lib/awful/widget/common.lua
@@ -96,7 +96,7 @@ function common.list_update(w, buttons, label, data, objects)
         if text == nil or text == "" then
             tbm:set_margins(0)
         else
-            if not pcall(tb.set_markup, tb, text) then
+            if not tb:set_markup_silently(text) then
                 tb:set_markup("<i>&lt;Invalid text&gt;</i>")
             end
         end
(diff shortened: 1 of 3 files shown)


blueyed:
:+1:

PR #638, opened 2016-01-18
Upstream shape api

Elv13:
A little pull request I worked on while Uli and Daniel were reviewing the layout stuff. @psychon Wanted to know how I draw those arrow in the dynamic splitting I demoed 2 weeks ago.

Shape used on a `wibox` to make it pretty:

snapshot14

Those 2 screenshot demo the use of shape for `wibox.widget.background` and the use of the `wibox.layout.margin` `draw_empty == false` feature for @mindeunix notification applet

snapshot14

snapshot14

```lua
-- Create notifications history
local notif = wibox.layout.margin(
    wibox.widget.background(
        wibox.layout.margin(
            notifications {fg = beautiful.fg_normal},
            7,7,1,1,
            nil,
            false
        ),
        beautiful.bg_alternate or beautiful.bg_normal,
        utils.shape.rounded_bar
    )
    ,0,0,1,1,
    nil,
    false
)
```

Collision and Radical had variation of these helpers for years. I think it would be generally useful, so here is an attempt to upstream them. In the future, I would add more shape. I have about 30 in my config and modules, but lets first see if there is interest

@psychon Do you think it would be possible to store the shape_bounding surface in X11 itself rather than creating a `.A1` at every call?

### DIFF
--- a/lib/gears/init.lua
+++ b/lib/gears/init.lua
@@ -17,6 +17,7 @@ return
     timer = require("gears.timer");
     cache = require("gears.cache");
     matrix = require("gears.matrix");
+    shape = require("gears.shape");
 }
 
 -- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/lib/gears/shape.lua
+++ b/lib/gears/shape.lua
@@ -1,0 +1,58 @@
+---------------------------------------------------------------------------
+--- Module dedicated to gather common shape painters.
+--
+-- It add the concept of "shape" to Awesome. A shape can be applied to a
+-- background, a margin, a mask or a drawable shape bounding.
+--
+-- The functions exposed by this module always take a context as first
+-- parameter followed by the widget and height and additional parameters.
+--
+-- The functions provided by this module only create a path in the content.
+-- to actually draw the content, use cr:fill(), cr:mask(), cr:slip() or
+-- cr:stroke()
+--
+-- In many case, it is necessary to apply the shape using a transformation
+-- such as a rotation. The preferred way to do this is to wrap the function
+-- in another function calling `cr:rotate()` (or any other transformation
+-- matrix)
+--
+-- @author Emmanuel Lepage Vallee
+-- @copyright 2011-2016 Emmanuel Lepage Vallee
+-- @release @AWESOME_VERSION@
+-- @module gears.shape
+---------------------------------------------------------------------------
+
+local module = {}
+
+--- Add a rounded rectangle to the current path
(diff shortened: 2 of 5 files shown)


psychon:
Lots of comments on the individual commits (sorry!), but I like this in general. Thanks!

> @psychon Do you think it would be possible to store the shape_bounding surface in X11 itself rather than creating a .A1 at every call?

Not really. The Shape extension provides the shape as a list of rectangles: https://github.com/awesomeWM/awesome/blob/master/xwindow.c#L316-L318
Also: Why do you want to store it? Which problem should that solve?

Elv13:
> Also: Why do you want to store it? Which problem should that solve?

Resize speed suck on old laptop when I add rounded border. for every resize event, it create a new surface and so on (but I admit that version wasted an ARGB32 surface as I used the same surface for background _and_ shape_bounding (it work, but is a dirty trick).

psychon:
Hm. Could you open an issue about that with some instructions on what exactly I'd have to do to reproduce? Assign that to me and I will investigate (but for now I doubt that the shape re-computation is all that slow; I'd rather guess that the time until awesome updates the shape is the problem here since without any shape the X server does that for us directly and before a redraw, now it happens afterwards).

Elv13:
Updated, this will be my last update for tonight

psychon:
Fine by me like this (missing modeline can be fixed latter, a no-op draw operation doesn't hurt and neither does an unused variable)

Elv13:
Last update, if nobody complain soon, I will merge it

Changes:

* Add the new line
* Use cairo CLEAR operator instead of the no-op to make sure the surface is correct, even then, it is not really required, but still, doesn't hurt too much.
* Remove the unused variable

psychon:
May I add `gears.surface.apply_shape_clip` and make `awful.client.shape` use this code or is there a reason against it that I miss?

Elv13:
> is there a reason against it that I miss?

I try to merge a 2800 SLoC patch at the same time and I am overwhelmed? As I said, I have 30 of those shapes, maybe 15 worth upstreaming. I also need to implement shapes for `wibox.widget.imagebox`, `wibox.layout.margin` (+ border) and implement `wibox.widget.clip`. Then, there is a pattern and image API too:

```lua
local image = blind_sur(img) 
    : colorize("#a0aab5ff") : resize_center(2,taglist_height,taglist_height) : shadow() : to_img()

local pattern = blind_pat.thick_stripe("#19324E","#132946",14,default_height,true)
    : add_noise(3, 4) : add_gradient(my_grad) : to_pattern()
```
(not the final API, I have a million TODO left)

But lets not get into that yet, first, the layout stuff

PR #642, opened 2016-01-19
Add ratio and stack layouts

Elv13:
Here are the commits I removed from the other pull request. I upgraded them, but I guess there is still some unfixed issues

### DIFF
--- a/lib/wibox/layout/ratio.lua
+++ b/lib/wibox/layout/ratio.lua
@@ -1,0 +1,364 @@
+---------------------------------------------------------------------------
+--- A layout filling all the available space. Each widget is assigned a
+-- ratio (percentage) of the total space. Multiple methods are available to
+-- ajust this ratio.
+-- @author Emmanuel Lepage Vallee
+-- @copyright 2016 Emmanuel Lepage Vallee
+-- @release @AWESOME_VERSION@
+-- @classmod wibox.layout.ratio
+---------------------------------------------------------------------------
+
+local base  = require("wibox.widget.base" )
+local flex  = require("wibox.layout.flex" )
+local fixed = require("wibox.layout.fixed")
+local table = table
+local pairs = pairs
+local floor = math.floor
+local util  = require("awful.util")
+
+local ratio = {}
+
+--- Set a widget at a specific index, replace the current one
+-- @tparam number index A widget or a widget index
+-- @param widget2 The widget to take the place of the first one
+-- @treturn boolean If the operation is successful
+-- @name set
+-- @class function
+
+--- Replace the first instance of `widget` in the layout with `widget2`
+-- @param widget The widget to replace
+-- @param widget2 The widget to replace `widget` with
+-- @tparam[opt=false] boolean recursive Digg in all compatible layouts to find the widget.
+-- @treturn boolean If the operation is successful
+-- @name replace_widget
+-- @class function
+
+--- Swap 2 widgets in a layout
+-- @tparam number index1 The first widget index
+-- @tparam number index2 The second widget index
+-- @treturn boolean If the operation is successful
+-- @name swap
+-- @class function
+
+--- Swap 2 widgets in a layout
+-- If widget1 is present multiple time, only the first instance is swapped
+-- @param widget1 The first widget
+-- @param widget2 The second widget
+-- @tparam[opt=false] boolean recursive Digg in all compatible layouts to find the widget.
+-- @treturn boolean If the operation is successful
+-- @name swap_widgets
+-- @class function
+
+--- Get all children of this layout.
+-- @param layout The layout you are modifying.
+-- @return a list of all widgets
+-- @name get_children
+-- @class function
+
+--- Fit the ratio layout into the given space
+-- @param layout The layout you are modifying.
+-- @param context The context in which we are fit.
+-- @tparam number orig_width The available width.
+-- @tparam number orig_height The available height.
+-- @name fit
+-- @class function
+
+--- Reset a ratio layout. This removes all widgets from the layout.
+-- @param layout The layout you are modifying.
+-- @name reset
+-- @class function
+
+-- Compute the sum of all ratio (ideally, it should be 1)
+local function gen_sum(self, i_s, i_e)
+    local sum, new_w = 0,0
+
+    for i = i_s or 1, i_e or #self.widgets do
+        if self._ratios[i] then
+            sum = sum + self._ratios[i]
+        else
+            new_w = new_w + 1
+        end
+    end
+
+    return sum, new_w
+end
+
+-- The ratios are expressed as percentages. For this to work, the sum of all
+-- ratio must be 1. This function attempt to ajust them. Space can be taken
+-- from or added to a ratio when widgets are being added or removed. If a
+-- specific ratio must be enforced for a widget, it has to be done with the
+-- `ajust_ratio` method after each insertion or deletion
+local function normalize(self)
+    local count = #self.widgets
+    if count == 0 then return end
+
+    -- Instead of adding "if" everywhere, just handle this common case
+    if count == 1 then
+        self._ratios = { 1 }
+        return
+    end
+
+    local sum, new_w = gen_sum(self)
+    local old_count  = #self.widgets - new_w
+
+    local to_add = (sum == 0) and 1 or (sum / old_count)
+
+    -- Make sure all widgets have a ratio
+    for i=1, #self.widgets do
+        if not self._ratios[i] then
+            self._ratios[i] = to_add
+        end
+    end
+
+    sum = sum + to_add*new_w
+
+    local delta, new_sum =  (1 - sum) / count,0
+
+    -- Increase or decrease each ratio so it the sum become 1
+    for i=1, #self.widgets do
+        self._ratios[i] = self._ratios[i] + delta
+        new_sum = new_sum + self._ratios[i]
+    end
+
+    -- Floating points is not an exact science, but it should still be close
+    -- to 1.00.
+    assert(new_sum > 0.99 and new_sum < 1.01)
+end
+
+--- Layout a ratio layout. Each widget gets a share of the size proportional
+-- to its ratio
+-- @param context The context in which we are drawn.
+-- @tparam number width The available width.
+-- @tparam number height The available height.
+function ratio:layout(context, width, height)
+    local result = {}
+    local pos,spacing = 0, self._spacing
+    local num = #self.widgets
+    local total_spacing = (spacing*(num-1))
+
+    --normalize(self)
+
+    for k, v in ipairs(self.widgets) do
+        local space = nil
+        local x, y, w, h
+
+        if self.dir == "y" then
+            space = height * self._ratios[k]
+            x, y = 0, util.round(pos)
+            w, h = width, floor(space)
+        else
+            space = width * self._ratios[k]
+            x, y = util.round(pos), 0
+            w, h = floor(space), height
+        end
+
+        table.insert(result, base.place_widget_at(v, x, y, w, h))
+
+        pos = pos + space + spacing
+
+        -- Make sure all widgets fit in the layout, if they aren't, something
+        -- went wrong
+        if (self.dir == "y" and util.round(pos) >= height) or
+            (self.dir ~= "y" and util.round(pos) >= width) then
+            break
+        end
+    end
+
+    return result
+end
+
+--- Increase the ratio of "widget"
+-- If the increment produce an invalid ratio (not between 0 and 1), the method
+-- do nothing.
+-- @tparam number index The widget index to change
+-- @tparam number increment An floating point value between -1 and 1 where the
+--   end result is within 0 and 1
+function ratio:inc_ratio(index, increment)
+    if #self.widgets ==  1 or (not index) or (not self._ratios[index])
+      or increment < -1 or increment > 1 then
+        return
+    end
+
+    local widget = self.widgets[index]
+
+    assert(self._ratios[index])
+
+    self:set_ratio(index, self._ratios[index] + increment)
+end
+
+--- Increment the ratio of the first instance of `widget`
+-- If the increment produce an invalid ratio (not between 0 and 1), the method
+-- do nothing.
+-- @param widget The widget to ajust
+-- @tparam number increment An floating point value between -1 and 1 where the
+--   end result is within 0 and 1
+function ratio:inc_widget_ratio(widget, increment)
+    if not widget or not increment then return end
+
+    local index = self:index(widget)
+
+    self:inc_ratio(index, increment)
+end
+
+--- Set the ratio of the widget at position `index`
+-- @tparam number index The index of the widget to change
+-- @tparam number percent An floating point value between 0 and 1
+function ratio:set_ratio(index, percent)
+    if not percent or #self.widgets ==  1 or not index or not self.widgets[index]
+        or percent < 0 or percent > 1 then
+        return
+    end
+
+    local old = self._ratios[index]
+
+    -- Remove what has to be cleared from all widget
+    local delta = ( (percent-old) / (#self.widgets-1) )
+
+    for k, v in pairs(self.widgets) do
+        self._ratios[k] = self._ratios[k] - delta
+    end
+
+    -- Set the new ratio
+    self._ratios[index] = percent
+
+    -- As some widgets may now have a slightly negative ratio, normalize again
+    normalize(self)
+
+    self:emit_signal("widget::layout_changed")
+end
+
+--- Set the ratio of "widget"
+-- @param widget The widget to ajust
+-- @tparam number percent An floating point value between 0 and 1
+function ratio:set_widget_ratio(widget, percent)
+    local index = self:index(widget)
+
+    self:set_ratio(index, percent)
+end
+
+--- Update all widgets to match a set of a ratio.
+-- The sum of before, itself and after must be 1 or nothing will be done
+-- @tparam number index The index of the widget to change
+-- @tparam number before The sum of the ratio before the widget
+-- @tparam number itself The ratio for "widget"
+-- @tparam number after The sum of the ratio after the widget
+function ratio:ajust_ratio(index, before, itself, after)
+    if not self.widgets[index] or not before or not itself or not after then
+        return
+    end
+
+    local sum = before + itself + after
+
+    -- As documented, it is the caller job to come up with valid numbers
+    if math.min(before, itself, after) < 0 then return end
+    if sum > 1.01 or sum < -0.99 then return end
+
+    -- Compute the before and after offset to be applied to each widgets
+    local before_count, after_count = index-1, #self.widgets - index
+
+    local b, a = gen_sum(self, 1, index-1), gen_sum(self, index+1)
+
+    local db, da = (before - b)/before_count, (after - a)/after_count
+
+    -- Apply the new ratio
+    self._ratios[index] = itself
+
+    -- Equality split the delta among widgets before and after
+    for i = 1, index -1 do
+        self._ratios[i] = self._ratios[i] + db
+    end
+    for i = index+1, #self.widgets do
+        self._ratios[i] = self._ratios[i] + da
+    end
+
+    -- Remove potential negative ratio
+    normalize(self)
+
+    self:emit_signal("widget::layout_changed")
+end
+
+--- Update all widgets to match a set of a ratio
+-- @param widget The widget to ajust
+-- @tparam number before The sum of the ratio before the widget
+-- @tparam number itself The ratio for "widget"
+-- @tparam number after The sum of the ratio after the widget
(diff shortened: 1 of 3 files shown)


lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,342 @@
+---------------------------------------------------------------------------
+--- Fill all the available space. Each widget is assigned a ratio (percentage)
```
blueyed: This should probably have some prefix about being a layout - that might be missing in the others as well - but still.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,342 @@
...
+local floor = math.floor
+local util  = require("awful.util")
+
+local ratio = {}
+
+--- Layout a fixed layout. Each widget gets just the space it asks for.
+-- @param layout The layout you are modifying.
+-- @param context The context in which we are drawn.
+-- @param width The available width.
+-- @param height The available height.
```
blueyed: `@tparam`, please (with sugar on top).

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,342 @@
...
+-- @name layout
+-- @class function
+
+--- Get all children of this layout
```
blueyed: Punctuation.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,342 @@
...
+-- @param layout The layout you are modifying.
+-- @warning If the widget contain itself and recursive is true, this will cause
```
blueyed: s/contain/contains/
s/recursive/\`recursive\`/

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,342 @@
...
+--   a stack overflow
```
blueyed: Missing dot.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,342 @@
...
+-- @param[opt] recursive Also add all widgets of childrens
```
blueyed: s/of/from/
Missing dot.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,342 @@
...
+--- Reset a fixed layout. This removes all widgets from the layout.
+-- @param layout The layout you are modifying.
+-- @name reset
+-- @class function
+
+-- Compute the sum of all ratio (ideally, it should be 1)
+local function gen_sum(self, i_s, i_e)
+    local sum, new_w = 0,0
+
+    -- Get the sum of all widget ratio
```
blueyed: Can be removed - it's in the function's doc already.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,342 @@
...
+function ratio:add(...)
+    -- Clear the cache
+    for k,v in ipairs({...}) do
+        self._ratios[v] = nil
+    end
+
+    fixed.add(self,...)
+
+    --FIXME This doesn't work if the same widget is there multiple time
+    --TODO FIX IT
```
blueyed: This should be fixed probably before merging?!

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,342 @@
...
+
+    normalize(self)
+end
+
+--- Remove one or more widgets from the layout
+--- @tparam widget ... Widgets that should be removed (must at least be one)
+--- The last parameter can be a boolean, forcing a recursive seach of the
+--- widget(s) to remove.
```
blueyed: The 2nd and following lines should only have two dashes.
Also dots.. ;)

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,342 @@
...
+
+    --FIXME This doesn't work if the same widget is there multiple time
+    --TODO FIX IT
+
+    normalize(self)
+end
+
+--- Remove one or more widgets from the layout
+--- @tparam widget ... Widgets that should be removed (must at least be one)
+--- The last parameter can be a boolean, forcing a recursive seach of the
```
blueyed: s/seach/search/

lib/wibox/layout/ratio.lua
Elv13: All of these are just copy paste of the existing ones from `wibox.layout.fixed`. Can I later do a PR just to fix them? Also, is it possible to include files for these redundant interface copy/paste? It is generally an issue as methods from `wibox.widget.base` are not present in any widgets API doc.

blueyed: > Can I later do a PR just to fix them? 

Sure.  Although it might be forgotten probably.

I don't know about the other question.

Elv13:
Rebased and fixed all @blueyed issues but the one regarding using widgets as table key, this still need to be fixed

Elv13:
Updated again, ready from my POV. I am not claiming it is bug free, but I use it for the dynamic layout and it now work a vast majority of the time (and when it doesn't, it is by a couple of pixels)

lib/wibox/layout/stack.lua
```diff
@@ -0,0 +1,166 @@
...
+
+local stack = {mt={}}
+
+--- Layout a fixed layout. Each widget gets just the space it asks for.
+-- @param layout The layout you are modifying.
+-- @param context The context in which we are drawn.
+-- @tparam number width The available width.
+-- @tparam number height The available height.
+-- @name layout
+-- @class function
```
psychon: This function is present below, it doesn't need extra docs here

lib/wibox/layout/stack.lua
```diff
@@ -0,0 +1,166 @@
...
+
+--- Get all direct children widgets
+-- @param layout The layout you are modifying.
+-- @return a list of all widgets
+-- @name get_children
+-- @class function
+
+--- Add some widgets to the given fixed layout
```
psychon: This (and other docs) talk about `the given fixed layout`. Hm, and I bet the same occurs in `flex`....

lib/wibox/layout/stack.lua
```diff
@@ -0,0 +1,166 @@
...
+-- @param width The available width.
+-- @param height The available height.
+function stack:layout(context, width, height)
+    local result = {}
+    local spacing = self._spacing
+
+    for k, v in pairs(self.widgets) do
+        table.insert(result, base.place_widget_at(v, spacing, spacing, width - 2*spacing, height - 2*spacing))
+        if self._top_only then break end
+    end
```
psychon: This widget has undocumented functions `:remove_widgets()` and `:fill_space()`. The later one doesn't even have any effect, but is still inherited from fixed.

No idea what to do about `:fill_space()`.

psychon: Oh and `:set_spacing()` is missing, too! (didn't notice it at first, because it's defined in a weird place in fixed). And the function definitely has a different meaning here than for fixed.

lib/wibox/layout/stack.lua
```diff
@@ -0,0 +1,166 @@
...
+function stack:set_display_top_only(top_only)
+    self._top_only = top_only
+end
+
+--- Raise a widget at `index` to the top of the stack
+-- @tparam number index the widget index to raise
+function stack:raise(index)
+    if not index or index > #self.widgets or index < 1 then return end
+
+    if layout == self and idx ~= 1 then
```
psychon: This `if` can be simplified to `if false then` (please use your sarcasm detector when implementing this)

lib/wibox/layout/stack.lua
```diff
@@ -0,0 +1,166 @@
...
+        idx, layout = self:index(layout, recursive)
+    end
+
+    if layout == self and idx ~= 1 then
+        self:raise(idx)
+    end
+end
+
+local function new(dir, widget1, ...)
+    local ret = fixed.horizontal(...)
```
psychon: A bit of a hack, but ok...

You recently wrote something about awesome needing better docs. Could you add something for this? E.g. something somewhere should explain what the indices mean and that index 1 is the bottom-most widget (if I understand `:layout()` correctly etc.

lib/wibox/layout/stack.lua
```diff
@@ -0,0 +1,166 @@
...
+    self._top_only = top_only
+end
+
+--- Raise a widget at `index` to the top of the stack
+-- @tparam number index the widget index to raise
+function stack:raise(index)
+    if not index or index > #self.widgets or index < 1 then return end
+
+    if layout == self and idx ~= 1 then
+        self.widgets[idx], self.widgets[1] = self.widgets[1], self.widgets[idx]
```
psychon: Uhm, why 1? If I understand `:layout()` correctly, then 1 is the bottom-most widget, not top-most.

lib/wibox/layout/stack.lua
```diff
@@ -0,0 +1,166 @@
...
+-- @name remove
+-- @class function
+
+--- Fit the fixed layout into the given space
+-- @param layout The layout you are modifying.
+-- @param context The context in which we are fit.
+-- @tparam number orig_width The available width.
+-- @tparam number orig_height The available height.
+-- @name fit
+-- @class function
```
psychon: Wait, this inherits `:fit()` from fixed? That makes no sense at all, does it?
Could you stack e.g. an imagebox and a textbox ontop of each other, add some other widget to the left/right of this and put all of that in a fixed layout? The idea is to verify that each widget gets a sane amount of space assigned (due to the inherited `:fit()`, the stack should be too wide).

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+local flex  = require("wibox.layout.flex" )
+local fixed = require("wibox.layout.fixed")
+local table = table
+local pairs = pairs
+local floor = math.floor
+local util  = require("awful.util")
+
+local ratio = {}
+
+--- Layout a fixed layout. Each widget gets just the space it asks for.
```
psychon: This also talks about `fixed`.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+            sum = sum + self._ratios[i]
+        else
+            new_w = new_w + 1
+        end
+    end
+
+    return sum, new_w
+end
+
+--- Make sure "diff" the sum is 1
```
psychon: I don't get this comment.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+
+    local delta, new_sum =  (1 - sum) / count,0
+
+    -- Increase or decrease each ratio so it the sum become 1
+    for i=1, #self.widgets do
+        self._ratios[i] = self._ratios[i] + delta
+        new_sum = new_sum + self._ratios[i]
+    end
+
+    -- Floating points summation isn't an exact science, worst on 32bit systems
```
psychon: What? Why 32 bit systems? Doesn't everything use IEEE 754 for float and double?

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+            x, y = util.round(pos), 0
+            w, h = floor(space), height
+        end
+
+        table.insert(result, base.place_widget_at(v, x, y, w, h))
+
+        pos = pos + space + spacing
+
+        if (self.dir == "y" and pos-spacing >= height) or
+            (self.dir ~= "y" and pos-spacing >= width) then
```
psychon: Why `pos-spacing >= height`? Shouldn't `util.round(pos) > height` be enough and more exact?

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+-- do nothing.
+-- @tparam number index The widget index to change
+-- @tparam number increment An floating point value between -1 and 1 where the
+--   end result is within 0 and 1
+function ratio:inc_ratio(index, increment)
+    if #self.widgets ==  1 or (not index) or increment < -1 or increment > 1 then return end
+
+    local widget = self.widgets[index]
+
+    assert(self._ratios[index])
```
psychon: Why does this check if `index` is a valid index with an assert? In other code you used `return false` for this case.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+    local index = self:index(widget)
+
+    self:inc_ratio(index, increment)
+end
+
+--- Set the ratio of the widget at position `index`
+-- @tparam number index The widget to change or its index
+-- @tparam number percent An floating point value between 0 and 1
+function ratio:set_ratio(index, percent)
+    if not percent or #self.widgets ==  1 or not index or index > #self.widgets
```
psychon: Missing `index < 1`. And it just occurred to me that all of these `index < 1 or index > #self.widgets` checks (which we already have elsewhere) do not catch things like e.g. `2.5`...

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+
+    for k, v in pairs(self.widgets) do
+        self._ratios[k] = self._ratios[k] - delta
+    end
+
+    -- Set the new ratio
+    self._ratios[index] = percent
+
+    -- As some widgets may now have a slightly negative ratio, normalize again
+    normalize(self)
```
psychon: What's the idea for when to call `normalize()`? All functions seem to call `normalize()` before doing anything, so that suggests that the idea is "the widget may always be 'inconsistent'". However, by that reasoning, you wouldn't have to call `normalize` here.

However, for speed reasons (beware, micro-optimization!) I would suggest to change that rule to "the widget is always normalized". This means that nothing has to call `normalize(self)` before doing something, since the assumption is that we are already normalize. Only after changing things (e.g. here) does normalize have to be called. I guess that would mean that this is the only call to `normalize` that is left.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+-- @param widget The widget to ajust
+-- @tparam number percent An floating point value between 0 and 1
+function ratio:set_widget_ratio(widget, percent)
+    local index = self:index(widget)
+
+    self:set_ratio(index, percent)
+end
+
+--- Update all widgets to match a set of a ratio
+-- @tparam number index The widget to change or its index
```
psychon: `The index of the widget to change`

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+-- @tparam number before The sum of the ratio before the widget
+-- @tparam number itself The ratio for "widget"
+-- @tparam number after The sum of the ratio after the widget
+function ratio:ajust_ratio(index, before, itself, after)
+    if not self.widgets[index] or not before or not itself or not after then
+        return
+    end
```
psychon: In `:set_ratio` you check for things like `<0` or `>1`. Should that be done here as well?

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+    local index = self:index(widget)
+    self:ajust_ratio(index, before, itself, after)
+end
+
+--- Add some widgets to the given fixed layout
+-- @tparam widget ... Widgets that should be added (must at least be one)
+function ratio:add(...)
+    fixed.add(self,...)
+
+    normalize(self)
```
psychon: Ok, there will be two places where `normalize` has to be called. :-)

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+--- Remove one or more widgets from the layout
+-- @tparam widget ... Widgets that should be removed (must at least be one)
+-- The last parameter can be a boolean, forcing a `recursive` search of the
+-- widget(s) to remove.
+function ratio:remove_widgets(...)
+    -- Clear the cache
+    local count = #self._ratios - #({...})
+    for i=count+1, #self._ratios do
+        self._ratios[i] = nil
+    end
```
psychon: What? Why are these `caches`? You are are throwing away settings here!
I think this has to implement `remove_widgets` itself and call `table.remove` with each removed index.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+--- Remove a widget from the layout
+-- @tparam number index The widget index to remove
+-- @treturn boolean index If the operation is successful
+function ratio:remove(index)
+    if not index then return false end
+
+    -- Pop the table elements
+    for i=index, #self._ratios do
+        self._ratios[i] = self._ratios[i+1]
+    end
```
psychon: Doesn't `table.remove` do this for us?

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+
+    local ret = fixed.remove(self, index)
```
psychon: More and more do I think that this widget shouldn't be based on `fixed`....

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+end
+
+--- Insert a new widget in the layout at position `index`
+-- @param index The position
+-- @param widget The widget
+function ratio:insert(index, widget)
+
+    fixed.insert(self, index, widget)
+
+    normalize(self)
```
psychon: Yup, please do not base this on `fixed`. That just makes it a lot harder to understand, but doesn't make the code that much shorter.

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,339 @@
...
+end
+
+local function get_layout(dir, widget1, ...)
+    local ret = flex[dir](widget1, ...)
+
+    util.table.crush(ret, ratio)
+
+    ret.fill_space = nil
+
+    assert(ret.index)
```
psychon: Hu? Why?

lib/wibox/layout/ratio.lua
Elv13: I will remove some of them, them and the abusive `assert()`s were useful until it got stable-ish. Now it's not really necessary anymore.

lib/wibox/layout/ratio.lua
Elv13: It was a widget key-ed cache in previous version, but it exploded when adding the same widget multiple time. I forgot to change the comment

lib/wibox/layout/ratio.lua
Elv13: Oops, leftover code

Elv13:
I updated the ratio layout. It still based on fixed (for the 6-7 methods it inherit), but no longer call fixed methods.

Elv13:
All done (I think/hope)

lib/wibox/layout/ratio.lua
```diff
@@ -0,0 +1,362 @@
...
+-- @tparam number itself The ratio for "widget"
+-- @tparam number after The sum of the ratio after the widget
+function ratio:ajust_ratio(index, before, itself, after)
+    if not self.widgets[index] or not before or not itself or not after then
+        return
+    end
+
+    local sum = before + itself + after
+
+    if sum > 1.01 or sum < -0.01 then return end
```
psychon: What's the purpose of this? It silently does nothing when.... something?

Elv13: Well, it address your comment just above, and is documented in the doc

> -- The sum of before, itself and after must be 1 or nothing will be done

Ok, the comment is not accurate, it should be 0.99 in that case, but it allow some room for flexibility (or I can change it to .99). I guess I should also check for individual negative values too, as this is invalid.

psychon: Ah ok. Then please also add a comment here. Something like "as per the API doc, this must hold" (but the API docs say `sum=1` while you allow `-0.01<sum<1.01`; as you say, this should likely be 0.99 instead of -0.01)

Elv13:
Updated. It is good enough now?

A future update will add minimum and maximum widget ratio and a choice of resize strategies. But first thing first, get this merged (and the other PR too), so I can go ahead with PR 8/10 and 9/10, enough to turn on the dynamic layout system on the tiled layout. (this is 6/10, the "add compatibility" is 7/10, almost there)

psychon:
:+1:

PS: I'm looking forward to your PR adding unit tests for these... :-)

lib/wibox/layout/stack.lua
```diff
@@ -0,0 +1,188 @@
...
+    while layout and layout ~= self do
+        idx, layout = self:index(layout, recursive)
+    end
+
+    if layout == self and idx ~= 1 then
+        self:raise(idx)
+    end
+end
+
+local function new(dir, widget1, ...)
```
psychon: Luacheck says:

    lib/wibox/layout/stack.lua:175:20: unused argument dir
    lib/wibox/layout/stack.lua:175:25: unused argument widget1

Could you please add APIdocs to something explaining the arguments and fixing the above? (From these warnings it looks like this just shouldn't get any arguments at all...)

Issue #424, opened 2015-08-23
Mouse resize in tile layout works weird for the bottom row when `size_hints_honor = false`


actionless:
Steps to reproduce:
1) set `size_hints_honor = false` in the rules
2) switch to tile layout
3) open 3 or more clients
4) try to resize client on the most bottom row (either slave or master if there is two or more master clients, ie more than one row)

Client size will jump crazily, while on other rows resize happens as expected.

motdef:
I also have this issue. It looks like the bottom row client becomes vertically smaller with every mouse movement until it reaches zero and "rolls over" so that it fills the column.

blueyed:
Could not reproduce it with rxvt-unicode windows.
When mod-right-clicking the cursor jumps to the corner, and the window gets resized a bit already (gets ~10pt smaller, related to taskbar?! - I don't have them by default).

actionless:
i noticed too the issue with cursor moved few pixels when starting resize, but that's another one

zhrts:
I get even weirder behaviour, when resizing the lower clients of three or more `urxvtc` in a tiled layout, the one I'm clicking on will jump around and then stick to the geometry of one adjacent to it, persisting even after I close and open the clients. Adding more clients to the layout begins to cascade them, and then resizing again tiles them back to normal all at once. 0_o

larkery:
I think this is due to drag corner detection failing or mismatching somehow. When I drag from the top left corner of a window in the middle of the right column (with 3 windows in that column), the mouse warps to the bottom left position and awesome seems to try and drag the top left to that position.

This collapses the window to be very small.

Then, when I move the mouse, it snaps back to how it was before, dragging the bottom left corner (even though I clicked the top left). So, it looks like the sequence is something like:

- Click top left, resize handler starts resizing from top left
- Mouse cursor warps to bottom left position
- Resize handler attempts to resize by moving top left to bottom left
- Resize handler then starts trying to drag from bottom left
- Moving mouse 1px puts the bottom left back where it was (where cursor warped to)

dimitris-triantafyllidis:
@actionless I gave awesome a try today and stumbled upon this behaviour. Do you use gaps? I tried switching gaps off and the problem disappeared.

Elv13:
@dimitris-triantafyllidis try to disable size honoring (see the FAQ) to make this problem "go away" with gaps. A second step one could take is to forbid clients from resizing themselves.

[...]

PR #644, opened 2016-01-19
Add the wibox.layout compatibility code

Elv13:
I will eventually drop the dummy surface, I haven't done so yet, I had enough work fixing the other PR, I guess this one will be even worst from that POV.

The resize code have issues. Up to the geometry request, it work fine up the the `request::geometry` part. Then there is something wrong and in some scenario, the result is a few pixel off. One of the issue is that the way border width and `c:geometry()` don't seem to agree with each other. But even with `border_width = 0`, there is still something wrong somewhere in the ratio stuff. It is minor, but umm, still annoying.


Elv13:
I improved the `resize` commit. I now use the "live" resize mode by default (same as the current stateless system) as `wibox.layout.ratio` has been fixed, it no longer explode. I left the other commit untouched not to lose @psychon  comments, I only fixed half of them so far (I just restarted working on this now that I don't have to spend an hour rebasing them everyday)

Elv13:
> The way you check if a tag is selected (tag.selected(tag.getscreen(t)) == t) is still broken when multiple tags are selected.

Actually, no, I think it is correct. If there is multiple selected tags and it isn't the first, then the layout in suspended. As with the current system, when multiple tags are selected, the first tag layout is used. I did some tests. It seem to work

Elv13:
Rebased and updated

Elv13:
Updated, ready for review

@psychon I implemented a different solution from the one you proposed. Using a parallel implementation of hierarchy traversing wasn't really working. Many widgets expect a certain sequence, so making it work was almost re-implementing hierarchy. The solution proposed in the commit is a little hacky, but cleaner than the other (IMO).

psychon:
Here are some selected (I removed things like "unused loop variable) warnings from Luacheck:


Checking lib/awful/layout/dynamic/base.lua

    lib/awful/layout/dynamic/base.lua:13:7: unused variable util
    lib/awful/layout/dynamic/base.lua:139:28: unused argument self
    lib/awful/layout/dynamic/base.lua:164:36: unused argument self
    lib/awful/layout/dynamic/base.lua:184:23: unused argument a
    lib/awful/layout/dynamic/base.lua:186:15: unused variable cr
    lib/awful/layout/dynamic/base.lua:186:34: accessing undefined variable img
    lib/awful/layout/dynamic/base.lua:300:27: accessing undefined variable unpack

Checking lib/awful/layout/dynamic/resize.lua

    lib/awful/layout/dynamic/resize.lua:12:7: unused variable screen
    lib/awful/layout/dynamic/resize.lua:99:31: unused argument offset
    lib/awful/layout/dynamic/resize.lua:139:54: unused argument corner
    lib/awful/layout/dynamic/resize.lua:139:62: unused argument x
    lib/awful/layout/dynamic/resize.lua:139:65: unused argument y
    lib/awful/layout/dynamic/resize.lua:145:26: accessing undefined variable geo
    lib/awful/layout/dynamic/resize.lua:238:39: unused argument c

Checking lib/awful/layout/dynamic/wrapper.lua

    lib/awful/layout/dynamic/wrapper.lua:13:7: unused variable client
    lib/awful/layout/dynamic/wrapper.lua:138:34: unused argument c

lib/gears/matrix.lua
```diff
@@ -150,6 +176,43 @@ function matrix:equals(other)
...
+        end
+    end
+end
+
+--- When adding transformations to the matrix, retroactively apply them
+-- @tparam boolean mutate Enable mutation
+function matrix:set_mutating(mutate)
+    rawset(self, "_automerge", mutate)
+end
+
```
psychon: I don't like all of this change to `gears.matrix`. I made matrices immutable on purpose so that e.g. I don't have to return a copy from functions in `wibox.hierarchy`, but can just return the internal matrix. If a matrix is immutable, you don't need to save/restore it, but can just keep the old instance around (in a stack?).

Let's see what you are using this for in the other commits...

lib/awful/layout/dynamic/base.lua
```diff
@@ -0,0 +1,322 @@
...
+    clip_extents        = function(self) return {width=0, height=0} end,
+    transform           = function(self, cmatrix)
+        return self:multiply(matrix.from_cairo_matrix(cmatrix))
+    end,
+}, {__index = matrix})
+
+-- Emulate enough of cairo context to use unmodied hierarchies
+local function create_fake_context()
+    local m = matrix.identity:copy()
+    m:set_mutating(true)
```
psychon: ...to make a matrix look like a cairo context. Sounds a lot like a hack. Also, the cairo API is a lot larger than this (random examples: `:set_operator()`, `.target`).

Oh and `:clip_extents()` is supposed to return two numbers, not a table. And when you fix this (and keep returning `0,0`), this whole code will break since `wibox.hierarchy:draw` sees "the clip is empty, so I don't have to draw anything" (clip extents are the basis on how partial redraws are implemented).

All together: NO!

psychon:
```
error while running function
stack traceback:
	./awesomerc.lua:15: in main chunk
error: ./awesomerc.lua:15: attempt to index field 'dynamic' (a nil value)
```
Apparently I have to require this separately, `require("awful")` is not enough. That should change.

psychon:
This is so broken, you may only reopen this once you added unit tests to this which cover at least 50% of the new code. Meh.

psychon:
Or can you at least tell me how to test this since apparently I didn't do the right thing from looking at the docs?

psychon:
I'm sorry I were so harsh and will try to be less angry when things break.

Elv13:
> how to test this

Better take the `update_dynamic_complete` branch of my fork. It has the missing `init.lua` and modified `awesomerc.lua`. It also has the huge "work in progress" commit, ignore it

> I'm sorry I were so harsh
Thanks, it was indeed a little harsh. Ready for review ~= ready to merge, it just imply I fixed the issues you pointed in the last round. You are also right this badly need tests. I will write them. 

It is not as broken as you think it is. Considering only the tile layout is properly implemented (the others are there, but incomplete and somewhat broken). The bugs are:

 * There is a rounding error leaving a 1px gap between some clients
 * Tabbing implementation is a little cheap. It "work" fine, but if you play with it for too long it will eventually do weird things.
 * Resize only works when there is 1 layer of X and Y ratio layouts. This isn't a design issue, I just need to add "ratio strategies" to `wibox.layout.ratio` as one size wont fit all for such resizing. Note that the current layouts do not support this kind of resizing at all for this very reason, so this is not a regression.
 * There is some minor itches, I use this code without problem, but my workflow may not hit all corner cases (also, my config may hide some).
 * The clients that require size_honoring are a little broken (nothing major, it just need some more work in resize.wrapper_geometry)

Note that the "dynamic" branch of my collision module need to be up-to-date. Resizing can be done using the usual keybindings, but the "better" version is mod4+alt+arrows to increase size and mod4+alt+shift+arrows to reduce it. If you get any errors, forward them here. also note that the mouse resize implemented by this PR is more advanced than the current one (and the code is also simpler, as long as the reader did some matrix math before). It support "side" resize, like left/right/up/down instead of corner only like the old implementation. It also support proper keyboard resize, the old implementation did not. It is also layout agnostic, unlike the previous code.

Elv13:
Rebased on top of the new APIs

It hasn't changed much in 2 months. The few issues it had back then are still there. I am still waiting for a review before spending more time on this.

The requested coverage is implemented since a month or so.

widgetlayout5

Elv13:
Updated. I removed the fake context, then it still didn't always work (the same corner cases as last time, no surprises here). It also added 5 new "standard" method to the widget API for almost no reason, which is ugly and confusing.

After that I reverted that commit and created a new "domain specific" context, but with @psychon `handle_hierarchy` code as backend. So the current solution is an hybrid of mine and psychon. It can still use unmodified normal widgets[1]  while not having to shim the whole clip part of the context (that was called, but had its value unused).

[1] including those who alter the CR matrix in the before_*/after_*, but not the clip (previously, the code would not cause error, but would do nothing). So not "all" widgets/layouts/containers are compatible, a little regression compared to my previous cairo context shim. I also force the translation back, I will have to re-implement more of wibox.hierarchy.draw to avoid doing this.

psychon:
If you really want this to use a cairo context, use an unbounded cairo recording surface as the target of the cairo context. Just, please, don't try to create a fake cairo context.
(Hm, widgets altering the transformation in their before/after draw methods... That's bad, I think. That can easily break redraws badly.)

psychon:
Do you perhaps have some concrete examples of the features that get lost when not using something-like-a-cairo-context for the widgets? I think that applying transformations in before_draw_child(ren) is broken anyway, so that doesn't really convince me. Also, isn't the whole purpose of the code here to assign geometry to clients? What kind of extra magic is needed for that?

Elv13:
Rebased

* Size hints are now supported
* The resize code has been merged in my last PR, so the copy provided by this PR has been removed
* The fullscreen layout is now really fullscreen (previously it was an alias on the max layout)

As for the fake context, my argument is that this system is designed to be and feel like a normal widget hierarchy. Of course drawing stuff and clip wont work, that's normal. But the point is that the API is exactly the same. No changes are needed compared to your proposed system. To me, this is a very important design goal. Remember, as I said in the declarative widget system PR a while ago, one if the "killer" feature of this system will eventually to be able to declare layouts inline[1]. In that case, the API will be used directly by "real" users (at least some Tyrannical ones who currently have quite insane way to do it according to the issues/questions I got). In that case, having one coherent API between clients and widget layouts is even more important. So, if `before_draw_children` do some funny thing, I want those funny things to also work in client layouts. Beside, the magnifier layout kind of use that to update the mwfact.

In an ideal world, just having a real cairo A1 context and not caring about how hacky that is would be the simplest solution. This fake context is uglier, but "take less memory" (for whatever this is worth). With an A1 contact with the screen size, then the "core" part of this system get back to ~600 lines, as much as it was in January. There is not much dark magic in the PR beside that hack. Everything else is kind of clean (except the tabbing part, but that's just a proof of concept)

[1]:

```lua
my_tag.layout = awful.layout.dynamic {
    {
        {
             on_top = true,
             some_client_prop = 42,
             spawn = "my_foo_client"
        },
        {
             border_width = 50,
             some_client_prop = 1337,
             spawn = "my_bar_client"
        },
        layout = awful.layout.dynamic.base_layout.vertical
    },
    {
         on_top = true,
         some_client_prop = 42,
         spawn = "my_awesome_client"
    },
    layout = awful.layout.dynamic.base_layout.vertical
}
```

psychon:
I don't have much time right now (and I'm tempted to just say "go ahead and merge this"...), but I want to reply to one thing:

> So, if `before_draw_children` do some funny thing, I want those funny things to also work in client layouts. 

What kind of funny things?
Anything modifying the `cr`'s transformation matrix is broken, because redraws won't work properly (the `wibox.hierarchy` really, really needs to know which widget is visible where). The `:layout`-callback can specify a transformation matrix, so messing with the matrix in `:before_draw_children` is not even necessary.
Everything else (clip, current source, line width, ...) only modifies the way in which things are drawn and all those values are unused when there is just a fake cairo context and all that is done is to set the geometry of clients.

The above is the whole reason why I don't think that having a cairo context here is not necessary and that using `:draw()` is not needed (client geometry can be queried from the hierarchy directly as well, without going through `:draw()`).

However, since this is you doing the work, pick your favorite solution and merge it...

copycat-killer:
Along with tabbed, what about a stacking layout like i3?

It could replace my [cascadetile](https://github.com/copycat-killer/lain/blob/master/layout/cascade.lua).

Elv13:
> Along with tabbed, what about a stacking layout like i3?

Can you describe exactly what you expect. I think it is covered already.

copycat-killer:
Do we have [this](https://www.youtube.com/watch?v=Wx0eNaGzAZU&t=3m4s) already in this pull?

Elv13:
Yes, but they are horizontal instead of vertical. Adding a vertical option adds like 10 lines of code. Do you want it?

copycat-killer:
> Do you want it?

Yes, especially if it's just 10 lines. 😃 

This PR pratically bashes away i3, so if you need a hand, I'd happily lend two.

Elv13:
Done
pick8


```
beautiful.tabbar_vertical = true
beautiful.tabbar_bg = "#222222"
beautiful.tabbar_fg = "#ffff00"
```

Bew78LesellB:
and what if I want the tabbar to be on the right of the clients ? (kind of multiple titlebars on the right)

Elv13:
> and what if I want the tabbar to be on the right of the clients ? (kind of multiple titlebars on the right)

5-6 more lines of code and another option ;). It really is a simple wibox.layout+wibox.widget. If you can do it in a wibar, then you can do it in your client layout.

Bew78LesellB:
and do you think there must be options for every case ?

Elv13:
> and do you think there must be options for every case ?

The point of that layout system is that it scales. Creating your own layout is so simple you can do it inline in your rules. But I don't mind adding options. There is no manual geometry math anymore so the complexity remain low and don't increase much.

Elv13:
Here's an example (I just noticed I never added one)

```lua
local stack = require("awful.layout.dynamic.tabbed")
local tile = require("awful.layout.dynamic.base_layout")
local manual = require("awful.layout.dynamic.suit.manual")

local mycustomtilelayout = manual {
        {
            max_elements = 1,
            priority     = 3,
            layout       = tile.vertical
        },
        {
            reflow       = true,
            max_elements = 2,
            priority     = 1,
            layout       = tile.vertical
        },
        {
            priority = 1,
            layout = tile.vertical
        },
        fill_space = true,
        layout     = tile.horizontal
}
```

pick8

actionless:
looks completely good, is it ready for some alpha testing? last time when i tried about 2-3 months i didn't managed it to work

btw, sorry for offtopic but i can't unsee it anymore after i've noticed it just now, your client controls are appearing like `fosex`, good what you haven't followed the common practice to have only 3 window controls ;-)
https://cloud.githubusercontent.com/assets/340384/14449966/5cfeb0e2-0047-11e6-97ac-748fb9d733d8.png

Elv13:
> looks completely good, is it ready for some alpha testing? last time when i tried about 2-3 months i didn't managed it to work

I depends how you tried. The code is mostly "as-is" since last year with minor rebase and fixes. You can use it "in production". I have been using it for over a year. It works for me and the layout/setup I use are bug free and unit tested.

actionless:
```diff
diff --git a/config/layouts.lua b/config/layouts.lua
index 1f02408..9905204 100644
--- a/config/layouts.lua
+++ b/config/layouts.lua
@@ -1,4 +1,6 @@
 local awful = require("awful")
+--require("awful.layout.dynamic")
+--local suit = require("awful.layout.dynamic.suit")
 local gears = require("gears")
 local beautiful = require("beautiful")
 local lcars_layout = require("actionless.lcars_layout")
@@ -17,6 +19,9 @@ function layouts.init(context)
     awful.layout.suit.floating,
     awful.layout.suit.fair,
     awful.layout.suit.spiral,
+
+    --suit.tile,
+    --suit.treesome.horizontal,
   }
   awful.layout.layouts = context.layouts
   -- }}}
```

Elv13:
@actionless you need 

    require("awful.layout.dynamic")

After that, to check if it's enabled, try to resize the fair layout. If it works, then you use the new layout system

actionless:
sorry, i copied some wrong diff :)

those lines which are green should be **un**commented

i guess the problem was what i was trying to use them directly, anyway i will try it again tomorrow

actionless:
resize indeed is working, feels so flexible, i like it more and more :-)

this time time, i've added only `require("awful.layout.dynamic")` without using `dynamic.suit` layout directly

actionless:
it also closes https://github.com/awesomeWM/awesome/issues/424 :champagne:  :fireworks:

actionless:
also on a `fair` layout i've noticed the problem:

1) after making some crazy resizing:
2017-02-10--1486688532_3840x2160_scrot
2) start closing/minimizing clients:
2017-02-10--1486688697_3840x2160_scrot
3) open/unminize some clients:
2017-02-10--1486688698_3840x2160_scrot

Elv13:
Yes, that's a recent "regression" I added semi-on-purpose to simplify some code. I wasn't expecting a sudden spike in enthusiasm :P. This is/was to simplify some code and try to centralize the "empty layouts" garbage collection, but I didn't implement that part yet. I will give it a shot now. The tiled layout still has it's on manual GC for empty columns.

Elv13:
@actionless I implemented automatic empty columns/rows garbage collection. The fair layout should be more reliable now. I will eventually rewrite the tiled layout. It is the oldest and doesn't use most of the improvements I implemented later.

> UPD: but now layouts are ignoring master_fill_policy:

My first version of this predates it. I never bothered to implement it. My idea for this is to make it easier to have different "sublayouts" depending on the number of widgets. It's in the list of TODO earlier in this thread.

actionless:
> master_fill_policy

i was thinking, mb to reimplement it in a layout-agnostic way?
so `master_fill_policy` tag property instead of being affected in layout logic will be just setting screen padding if it's only one client on tag, and restoring previous value when it will be more clients

Elv13:
@actionless What do you think about simply making it easier to use different layouts depending on the number of clients? We already got the "can I make master_fill_policy align like the normal layout" question twice. Maybe this is one of those features @Bew78LesellB mentioned where adding an infinite number of options isn't a good idea. Having an easy way to manager multiple layout depending on the number of client solve the issue "once and for all".

actionless:
> align like the normal layout
what does it mean?

or mb you could the conversation? i am not sure if i read it

Elv13:
> or mb you could the conversation? i am not sure if i read it

Be on IRC more often ;). They want the master client to be positioned as it would be if there was 2 columns instead of centered.

actionless:
after using it for a day i also noticed some weird problems with clients order -- when restarting awesome or changing number of columns/masters clients are changing order, i can't understand yet are they reversing their order or the change is just random

Elv13:
Good question. It's not random. It has to do with client index. The current system is itself stateless, but save some states in both the client and tag object. The new system contain its own state within itself. It sometime tries to synchronize with the client index, but there is no "clear cut" right thing to do when the state is lost. Given those layouts could be serialized, maybe that's the thing to do. Right now, no layout implement the serialize() method. But they could, its planned. Also, is it a problem? Awesome never kept the layouts intact when being restarted. Given that it no longer restarts on screen changes, maybe "I don't care" is the best possible answer?

actionless:
the problem what it also happens when changing number of columns/masters

and previously that wasn't happening on restart

actionless:
i think the biggest problem now what when setslave is now replaced to `.slave` property it's not changing the index right

Elv13:
> .slave property it's not changing the index right

Confirmed, I will fix it eventually (or feel free to fork my branch and submit a pullrequest on my github fork (not the official repo))

actionless:
currently the simplest way to fix it is:
```diff
--- lib/awful/client.lua        2017-02-11 17:04:18.678698073 +0100
+++ /usr/share/awesome/lib/awful/client.lua     2017-02-12 01:28:58.957548070 +0100
@@ -300,6 +300,7 @@
 -- @legacylayout awful.client.setslave
 -- @client c The window to set as slave.
 function client.setslave(c)
+    c.slave = true
     local cls = capi.client.get(c.screen)
     for _, v in pairs(cls) do
         c:swap(v)
```

but it looks somehow weird to me

blendi-93:
my two cents:

* **this is amazing!**
* is there alrdy a doku? if not [how] can i:
  * set the width of a block
  * make them expand if theres less then 4 windows (based on ur example  Elv13 commented on 9 Feb )
  * return my own layoutname instead of ASDASD

Elv13:
> return my own layoutname instead of ASDASD

I rebased the patches and implemented this. You can now set `name` in the top-level block. It can be a function too

> is there alrdy a doku? if not [how] can i: 

yes, if you have `ldoc` installed, it will be generated and located in `awful.layout`

> set the width of a block [...] make them expand if theres less then 4 windows (based on ur example Elv13 commented on 9 Feb )

In theory, this `ratio` property, but it isn't fully implemented. The tile layout has it fully implemented. You can achieve the same result as the manual layout using the little documented split functions.

lib/awful/layout/dynamic/suit/manual.lua
```diff
@@ -0,0 +1,169 @@
...
+            end
+
+            self:remove_widgets(...)
+        end
+
+        return main_layout
+    end
+
+    -- Pick a name for this layout
+    local name = type(args.name) == "function" and name(args) or name or "manual"
```
copycat-killer: This casts a Travis `undefined` warning. I would suggest:

```diff
-local name = type(args.name) == "function" and name(args) or name or "manual"
+local ntype = type(args.name)
+local name = (ntype == "function" and args.name()) or (ntype == "string" and name) or "manual"
```

to also assert that the name can only be a function or a string.

copycat-killer:
This is the last Travis error:

```
error while running function!
stack traceback:
	[C]: in function 'require'
	lib/awful/layout/init.lua:19: in main chunk
	[C]: in function 'require'
	lib/awful/init.lua:42: in main chunk
	[C]: in function 'require'
	/home/travis/build/awesomeWM/awesome/awesomerc.lua:4: in main chunk
error: error loading module 'awful.tag' from file 'lib/awful/tag.lua':
	lib/awful/tag.lua:336: '=' expected near 'deleting'
error while running function!
stack traceback:
	/usr/local/etc/xdg/awesome/rc.lua:3: in main chunk
error: /usr/local/etc/xdg/awesome/rc.lua:3: loop or previous error loading module 'awful'
2017-05-15 07:32:32 E: awesome: main:820: couldn't find any rc file
make[3]: *** [CMakeFiles/check-integration.dir/all] Error 2
make[2]: *** [CMakeFiles/check.dir/rule] Error 2
make[1]: *** [check] Error 2
make: *** [check] Error 2
```

copycat-killer:
Can you give me an example on how to use `tabbed` to simulate suckless one?

Elv13:
> Can you give me an example on how to use tabbed to simulate suckless one?

Can you be more specific?

copycat-killer:
I can't figure out how to replicate [suckless tabbed](http://tools.suckless.org/tabbed) horizontal tabbing as a dynamic layout.

Elv13:
Assuming you ask how to take 2 clients and tab them, there is a spliting/tabbing API you can use. See how the `dynamic` branch of my `collision` module uses it (the `split.lua` file). Given there is multiple ways to split/merge a client, the API is not as pretty as I would like it to be, but it's there.

Note that you can also add the collision module and just use it (or improve it) to add tabbing capabilities to your configs.

If you asked for something else, please be even more specific and describe the behavior you desire to achieve.

copycat-killer:
Thanks for your willingness. 

What I'm trying to ask is an example of [horizontal tabbing](https://www.youtube.com/watch?v=Wx0eNaGzAZU&t=3m28s) with this PR.

Elv13:
> lib/awful/tag.lua:336: '=' expected near 'deleting'

@copycat-killer Sorry, but that error doesn't match the code from this branch. Did you add extra patches?

>     +local ntype = type(args.name)
>    +local name = (ntype == "function" and args.name()) or (ntype == "string" and name) or "manual"

@copycat-killer applied, thanks

> What I'm trying to ask is an example of horizontal tabbing with this PR.

@copycat-killer Like this:

```lua
local mycustomtilelayout3 = manual {
    {
        reflow       = true,
        max_elements = 2,
        priority     = 1,
        layout       = tile.vertical
    },
    {
        priority = 1,
        layout    = require('awful.layout.dynamic.tabbed')
    },
    inner_fill_strategy = "default",
    layout     = tile.horizontal
}
```

pick13


It was broken earlier today due a recent change in Awesome. After a long WTF time I found it.

> set the width of a block

@blendi-93 Mostly done. You can now use constraints. It fills the space by default.

> UPD: but now layouts are ignoring master_fill_policy:

@actionless Done. there is now 5 filling policies. I am sure you will want to add more. See the `inner_fill_strategy` in the `ratio` layout for the documentation. I guess you will want to add an additional "spacing" mode.

**Other news:**

* The corner.nw layout is now fully working.
* More tests were added
* Many bugs fixed
* A new example how to add margins (or size constraints) has been added to the doc

**Whats next:**

Drop the current `tile` implementation and replace it with a specialized `manual` layout.

copycat-killer:
> @copycat-killer Sorry, but that error doesn't match the code from this branch. Did you add extra patches?

I just reported current Github's Travis CI, not my local. And it still persists. Maybe some misconfiguration on the server?

Thanks for the (fix behind the) example, I tried something similar and couldn't figure out why it was not working.

I think it would be great to have this PR in v4.2, so I'll try to help reviewing and testing in the next days.

psychon:
> I just reported current Github's Travis CI, not my local. And it still persists. Maybe some misconfiguration on the server?

That error is from the built which tests all commits individually. The error occurs with just the very first commit of this PR. This commit is obviously broken: https://github.com/awesomeWM/awesome/pull/644/commits/9252ec0f10921a08b4b057afd7716d6d0b7feff1
The following commit fixes the above: https://github.com/awesomeWM/awesome/pull/644/commits/bf2a01ba0638f80a569e438876b695f360f19253
Both commit messages sound like left-overs from a rebase gone wrong.

The only real problem that Travis points out are a couple of Luacheck warnings.

Elv13:
> I just reported current Github's Travis CI, not my local. And it still persists. Maybe some misconfiguration on the server?

Ah, yea, sorry. I wont fix that one right now. Given I rebase this PR every couple of weeks, luacheck issues regularly get added and fixed. There is also many work in progress commits I don't even try to "clean" yet. This project is still farther away from being merged. However it can be packaged as a module (minus the 2 patches I added yesterday to the ratio layout, but it will be upstreamed soon). I target this PR for v5.0. 

At this point it is quite stable, but regress every now and then when upstream commits changes the behavior. 4.x have a 99% stable API and behavior, but this PR adds very sensitive code. The main design issue of this system is the fact that stateful systems are more fragile than stateless ones. This mostly true for everything in engineering. However, stateless systems are limited and don't scale to advanced use case (try to see those silly i3 like layouts for 3.5 and 4.0 on GitHub, they pile hacks on top of hacks). At this point, I have 85% test coverage and add more and more asserts to catch those regressions. But as long as it isn't merged, it might regress.

Also, there is some APIs that have yet to be written and upstreamed before this can be merged. The most important is https://github.com/awesomeWM/awesome/issues/1534 and the Z-index support for the clients+wiboxes. Without the first one, screen hotpluggin loses the state. Without the second one, the tabbed section order is tricky/hacky to keep working. https://github.com/awesomeWM/awesome/pull/1487 also needs to be merged for the serialized layouts to work, but it is stuck in design discussions.

Elv13:
I did a real commit-by-commit rebase (instead of a fast-forward) so all the rebase weirdness should be gone. I also fixed all regressions from #1776 and removed some commits like the "stop use fake context, then revert it because it broke everything and wont ever work" to simplify the commits list. Finally, I added a new `manual` example to show how to use a `mirror` container to simplify some layouts.

Asynchronous003:
Hello, forgive me if this is the wrong place. I am trying to build this  pr (#644)  but getting this error "build/lib/gears/color.lua:45: ?: 'class' cannot have multiple values; {module,module}" and do not really know what i am doing. should i put this in the 2hard basket, and move on?
Thank you for your time.

psychon:
@Asynchronous003 That's not related to this PR, but also happens on current master (at least for me). Seems to be once again some LDoc version incompatibility, only that this time I have no idea what causes it. Having CMake not find LDoc might be a way around this, but I'm not quite sure how to tell CMake to pretend that no `ldoc` binary exists.

Asynchronous003:
@psychon thank you, seems I am out of my dept. Love your work guys, Peace.

Elv13:
@Asynchronous003 You just need to uninstall `ldoc` and flush Awesome build directory.

Asynchronous003:
@Elv13 Thank you. Will keep at it :)

actionless:
@Elv13 btw i don't remember if `master_fill_layout_policy` tag property is already implemented on this branch?

Elv13:
> @Elv13 btw i don't remember if master_fill_policy tag property is already implemented on this branch?

Yes, but not in a way 100% compatible with your version. My version is offers more granularity instead of "global" tag `fill_policy`. For example, you can have different options for horizontal and vertical directions.

https://awesomewm.org/apidoc/classes/wibox.layout.ratio.html#wibox.layout.ratio.inner_fill_strategy

Today I worked to address one of the final "big ticket" from the list above to be able to use totally different layouts depending on the number of clients. However I want the API to be fluid and intuitive. Right now I am not happy about the compromise I had to make, so I wont push it yet.

Elv13:
I rebased this PR after spending some time last weekend fixing a couple issues (because I promised to do so).

@actionless There's something you will like there.

There is now an experimental `conditional` declarative container. This can be used to create layouts with `if/elseif/else` sections. This is very nice to create layouts that act differently depending on the number of clients. No need for hardcoded `master_fill_policy`. Having layouts that act depending on the number of client is now easy enough to write

https://elv13.github.io/libraries/awful.layout.html#awful.layout.suit.dynamic.manual (the 2 last examples)

screenshot_20180416_155352

```lua
local mycustomtilelayout = manual {
        {
            max_elements = 3,
            {
                {
                    priority     = 2,
                    max_elements = 1,
                    ratio        = 0.33,
                    layout       = tile.vertical
                },
                {
                    max_elements = 1,
                    priority     = 1,
                    ratio        = 0.33,
                    layout       = tile.vertical
                },
                {
                    max_elements = 1,
                    priority     = 3,
                    ratio        = 0.33,
                    layout       = tile.vertical
                },
                reflow              = true,
                inner_fill_strategy = 'spacing',
                layout              = tile.horizontal
            },
            margins = 3,
            widget  = margin
        },
        {
            reflow       = true,
            max_elements = 4,
            min_elements = 4,
            layout       = fair
        },
        {
            reflow       = true,
            min_elements = 5,
            layout       = corner
        },
        reflow = true,
        layout = cond
    }
```

actionless:
and also i see ratio in the example is hardcoded, while in the current master_fill_policy implementation it synchronized master_width_factor, which allows master window stay the same size whatever it's single or some more slave clients are opened

actionless:
also `max_elements` is not enough unless you're going to deprecate the defintion of `master` and `slave` clients at all, if not -- it will be else useful to have `max_master` and `max_slave`

Elv13:
> does it allowing to do the same as toggling master_fill_policy value on the fly?
> mb it can be implemented via allowing those custom layouts to use some custom tag properties?

The "custom" serialized layouts are not really designed to be very reactive. However you have a point that tracking only focus and the number of clients might not be enough to support all use cases. The `conditonal` layout code currently has an hardcoded list of things it tracks. I will expose the properties to the public API to allow tracking additional signals.

> also max_elements is not enough unless you're going to deprecate the defintion of master and
> slave clients at all, if not -- it will be else useful to have max_master and max_slave

I thinks that's irrelevant to `manual` layouts. If someone write a layout manually, I don't think he/she/it/they/sdfdsfsdfdsfsgfher want[s] random variables set "globally" (per tag) to affect the layout. This is especially true it's a serialized layout (in that case, you really, really want to restore it as it was when saved). The `tile` and other normal suits implement `master_count`/`master_width_factor`/`master_fill_policy` (mostly) correctly. But this is something else entirely. The `manual` layouts are not designed to use external values.

But as said earlier, I will add ways to run the `conditional` `:eval()` method on more signals, so if you use a `when` function (see documentation), then you should be able to use those variables if you wish.

> can it be defined like ratio = function(t) return t.master_width_factor end", instead of ratio = 0.33,?

I guess it could be done, but that would hardly solve your problem. It's cool to have a function, but when is it called and when is it refreshed? On the conditional layout itself, that's easy enough to fix, but on the "inner" widgets, they cannot know when to call the function, so that would be a bit pointless beside the moment where they are instantiated. Maybe that's the most important case anyway so we don't care about `t.master_width_factor` changing in the future, but I don't think that's your intention.

But again, the `awful.layout.suit.tile` works as expected, so to me it's almost a non issue. As for being able to resize dynamic layout, there is a "real" API where you can say "please add 30 pixels to the focused client on the left, thank you" instead of an inflexible API that can't even work properly when you have multiple columns.

actionless:
but it's being used not only by me, just google the mentions of `master_fill_policy` by other people

Elv13:
> but it's being used not only by me, just google the mentions of master_fill_policy by other people

I don't think you get what I am trying to say here (and I am probably to blame for that). `master_fill_policy` *is* intended to be implemented by the `tile`/`corner` suit as-is. I am not proposing to remove or deprecate it for the **tile**/**corner** layouts or any layout shipped with Awesome that uses it (which is currently limited to those 2). Neither I am intending to cause **any** change in behavior whatsoever.

The `manual` layout is a *new* feature not currently available at all and has a more fine grained way of customizing how the unused space is distributed. It is totally parallel to the existing layout suits. It aims at i3 use cases, not the current Awesome `awful.layout` ones. The `manual` layout is just another layout that allows custom definition, tabs and serialization.

barlik:
Hi, any update on this being merged?

Elv13:
> Hi, any update on this being merged?

Most probably not this year. This is a long term project and bits and pieces are merged all the time. The current focus is on the declarative control flow aspect necessary to support serialized layouts. The next iteration of this branch will split the project into a true module instead of a branch and will add the ability to spawn programs in specific location. This is now possible due to a recently merged pull request and another currently undergoing review.

There wont be a big "let's merge pr #644" event. It's all going to be parts being merged when they are mature enough.

Sinvers:
That's a really good work that you have done here. It was the only thing that was really missing from Awesomewm.
But I faced a problem, is this a known bug that the focused client is never taken in account while using the awful.layout.dynamic.suit.treesome ? Or am I doing domething wrong ? I built the project from this pull request. Everything looks fine, except that I just can't always get to split the right client. The only way I manage to make it works is maximizing the client, then open an other client and unmaximize the first client. And then, the split happened at the right place.
Anyway, great job, I'm looking forward to the project's progress.

Elv13:
@Sinvers sorry for the delay, it has been "I will find time tomorrow" for a month now.
 
> Or am I doing domething wrong

No, you are doing it right. The commit message has `[BROKEN]` in its name for that reason ;). I am very surprised this suit works at all without error. It's the first one I wrote, it never worked properly and I never ever updated it since I wrote it.

The issue is that by the time the client is added to the layout, it already has focus and as it can split itself, it falls back to the largest remaining client. So the fallback code is the only one that can be reached unless you add some `awful.rules` to disable autofocus. At that time there is no clean way to fox this. That's why I submitted, I wrote that pull request 2 years ago https://github.com/awesomeWM/awesome/pull/1031

However I never patched the `treesome` layout to use it. Until now that is! I wont update this pull request with the fix. I will probably never update it again as I broke the API in the second generation of this module. You can get it in my fork `doc_tests_and_notif` branch under the `lib/dynamite` directory. It is a temporary placeholder name until I find a better one for the module. Otherwise it is a much cleaner version of the PR. It is 99% complete and has addressed much of the TODO in the list above and much of @psychon complaints (beside the fake cairo context, that's unfixable with the current wibox API).  Only the serialization part isn't final. See doc at https://elv13.github.io/libraries/dynamite.html . As always with that branch, it is my fork "main branch" and has a megaton of incomplete stuff and I break the API all the time.

I will close this PR and open a new once I found a better name and added regression tests for minimization, maximization and master_fill_policy. @blueyed I can no longer reproduce the bug you had last year, mind trying again with the version from my `doc_tests_and_notif` branch?

Sinvers:
@Elv13 Thanks for the precise answer, I'll check this out !

actionless:
@GrayJack depends on user feedback

is it working fine for you?

GrayJack:
I don't use it, since I don't know witch branch to clone to compile it

What is the branch, cause I really want to test it!!

Elv13:
Elv13 upstream_dynamic_p7 is the branch for this PR, there is a more advanced one, but it has a couple minor regressions regarding floating clients (but has more testable code, which isn't a factor for people who want to try it)

The doc is here https://elv13.github.io/libraries/dynamite.html

If you want to help, resolve all code comments here:
https://github.com/awesomeWM/awesome/pull/2244 I have zero time and I am not working on getting this merged. The few hours I have for Awesome go toward getting the other big project merged (notification rewrite).

GrayJack:
And where I can find a example of rc.lua

GrayJack:
What is the regression for floating? if is something minor, I don't mind using the more updated branch :smile:

Elv13:
Try the `doc_tests_and_notif` branch of my fork, which has all the commits from all my ongoing projects. If you can't find bugs, then was there a regression in the first place? A word of warning, this branch has a lot of stuff in it, so random other things may have regressed. The notification maybe? It should not be that bad and you can email me bugs, but this is a very experimental branch.

GrayJack:
I want to do something like i3, dynamic split + both types of stack.

One thing I don't understand about your documentation I where I define my layouts, is it inside the `awful.layout.layouts` or is somewhere else?

Elv13:
> I want to do something like i3, dynamic split + both types of stack.

It has all the feature to do that, for now it's a bit low level and non-trivial to use them, but it's definitely there.

> One thing I don't understand about your documentation I where I define my layouts, is it inside the awful.layout.layouts or is somewhere else?

By default it uses the same layouts as Awesome. If you want to add new layouts, you can add them there is directly set them on the tag (see my tyrannical module).

In the near term,  this pull request was recently merged https://github.com/awesomeWM/awesome/pull/2360 and will make this easier to manage going forward.

GrayJack:
Ok, i put `dynamite.suit.treesome,` on the list, but I get an error in the line that I put it saying
`<name> expected near ','`

Elv13:
You are probably missing a `,` at the end of the previous line. For further issue, email me at elv1313 at gmail. That will prevent flooding this pull request with support questions.

If the error isn't a missing `,` at the previous line, please send the complete error message (and use Xephyr to debug).

skrzyp:
What's the current situation of this?

Elv13:
> What's the current situation of this?

Unchanged. It works fine and is usable in production. There another, newer, version that has some improvements regarding unit testing, but has some regressions. This project is too impractical to merge during Awesome 4.x lifetime so is maintained out-of-tree. Some parts get merged as time pass, but I have very limited time to work on these things, so other AwesomeWM related project get more attention than this. Currently finishing the notification rewrite and getting v4.4 into shape is the priority.

intrntbrn:
Idea: There should be a possibility of enforcing tabs via rules, so that e.g. every new instance of `wm_class` gets added as a tab by default.

m-kru:
@Elv13 any status update? When is it planed to be merged into master?

Narice:
Time for the yearly status update questoom x)
Is there any plans to get this merged or should someone interested in having tabbed and stacked layouts for screen real estate just stick to i3 for now?
In my research, it seems like i3 is the only tiling window manager to have tabbed and stacked layouts (If I'm wrong, that would be awesome)

Elv13:
For now there are:

 * https://github.com/xinhaoyuan/layout-machi
 * https://github.com/RobSis/treesome

Getting this PR merged cannot happen during the 4.x lifecycle and has to wait for 5.0. The code here still works and still does what it claims to do. As of now 4.4 is getting close. It was "supposed" to be the last 4.x, but I don't have enough time/energy to finish all the features I originally planned, so it looks like there is still room for a 4.5 before getting to the point where bumping the default API level to 5 becomes worthwhile.

As for "why can't you just merge it and leave it disabled by default", this PR is missing an important chunk. Right now, AwesomeWM has some "layers" for clients like ontop/below/above and a `:raise()` and `:lower()` method to move objects in those stacks. That's not enough to make tabbed layouts non-buggy. Switching tags currently messes with the tab order and it's annoying. A large refactor needs to happen in the C code and it will change some behavior, thus is hard to integrate in a minor release.

Narice:
Woaw thank you for the great explanation and the links!
I'll check back when 5.0 is released then 😄
I'm super hyped that another window manager will support these modes natively in the future, awesome!

LeducH:
5 years later still not merged. this is so funny!

PR #649, opened 2016-01-23
Make benchmarks less exact

psychon:
The benchmarks in tests/test-benchmark.lua have two modes. When CI=1 is set in
the environment, only a "quick" and less exact test is done. Otherwise, a slower
and more exact measurements is taken. This was added so that we do not waste CPU
time on travis.

However, most of the time the user running "make check" doesn't want exact
measurements either. So instead of only being quick when CI=1 is set, this
commit changes the logic to always being quick unless BENCHMARK_EXACT=1 is set.
Additionally, a message is printed next to the benchmark results so that the
user is reminded to set this var if the measurements should actually mean
something.

### DIFF
--- a/tests/test-benchmark.lua
+++ b/tests/test-benchmark.lua
@@ -5,7 +5,10 @@ local awful = require("awful")
 local GLib = require("lgi").GLib
 local create_wibox = require("_wibox_helper").create_wibox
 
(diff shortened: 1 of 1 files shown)


psychon:
Thanks for merging!

blueyed:
Thanks for all your work, too! :)

PR #652, opened 2016-01-24
Travis lua 5.3 support

psychon:
This PR adds Lua 5.3 to the travis build matrix. Since this Lua version is not available in Ubuntu Trusty, it is installed by hand (Urgh, Lua should get a proper build system).
While looking at this, I noticed some obsolete dependencies being installed in Travis and cleaned that up as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,8 @@ install:
           else
                   sudo apt-get install -y lib${LUANAME}-dev ${LUANAME} ${INSTALL_PKGS}
           fi
+  # "Create" /usr/bin/lua if needed (Yup, this is a bad hack)
+  - [[ -n "$LUAROCKS_ARGS" ]] && sudo ln -s /usr/bin/lua* /usr/bin/lua
 
   # Install luarocks (for the selected Lua version).
   - travis_retry wget https://keplerproject.github.io/luarocks/releases/luarocks-2.2.2.tar.gz
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
 
   # Install build dependencies.
   # See also `apt-cache showsrc awesome | grep -E '^(Version|Build-Depends)'`.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -28,13 +31,26 @@ install:
...
-  - sudo apt-get install -y libcairo2-dev xmlto asciidoc libpango1.0-dev gperf luadoc libxcb-xtest0-dev libxcb-icccm4-dev libxcb-randr0-dev libxcb-keysyms1-dev libxcb-xinerama0-dev libxcb-image0-dev libev-dev libimlib2-dev libdbus-1-dev libxdg-basedir-dev libstartup-notification0-dev imagemagick libxcb1-dev libxcb-shape0-dev libxcb-util0-dev libxcursor-dev libx11-xcb-dev libxcb-cursor-dev libxcb-xkb-dev libxkbcommon-dev libxkbcommon-x11-dev
+  - sudo apt-get install -y libcairo2-dev xmlto asciidoc libpango1.0-dev libxcb-xtest0-dev libxcb-icccm4-dev libxcb-randr0-dev libxcb-keysyms1-dev libxcb-xinerama0-dev libdbus-1-dev libxdg-basedir-dev libstartup-notification0-dev imagemagick libxcb1-dev libxcb-shape0-dev libxcb-util0-dev libx11-xcb-dev libxcb-cursor-dev libxcb-xkb-dev libxkbcommon-dev libxkbcommon-x11-dev
 
   # Deps for functional tests.
   - sudo apt-get install -y dbus-x11 xterm xdotool xterm xvfb rxvt-unicode
 
   # Install Lua (per env).
-  - sudo apt-get install -y lib${LUANAME}-dev ${LUANAME} ${INSTALL_PKGS}
+  # Note that Lua 5.3 is installed by hand, because not available in trusty
+  - |
```
blueyed: That works (by now?)? (multiline statements?)

psychon: Apparently it does. I was "inspired by" (= stole it from) here: https://github.com/znc/znc/blob/2bcc5c51593fd63de5f0e2757bbfbff13451b06d/.travis.yml#L60-L65 The only downside is that I need that `|| exit 1` else Travis does not fail if a subcommand fails.

I just don't know why Travis didn't try another rebuild with my additionally pushed commit which "added" `/usr/bin/lua` (previously `build-utils/lgi-check.sh` was using lua5.1 which was pulled in by Luadoc, because I removed Luadoc this fails now). Any ideas?

blueyed: > Any ideas?

Probably of an error with the `.travis.yml`.  See https://travis-ci.org/awesomeWM/awesome/requests.

> The only downside is that I need that || exit 1 else Travis does not fail if a subcommand fails.

Have you tried using `set -e`?

blueyed: Seems to work for me.

psychon: Because the line started with `[[`, travis had a parsing error (somewhere on the web page it said something like "no config found"). I'll squash the mess that I created here later.

blueyed:
:+1:

blueyed:
Squashed into 6479baf.

psychon:
Thanks!

PR #654, opened 2016-01-26
wibox: Read border_width and color from constructor

Elv13:
It was missing

### DIFF
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -146,6 +146,12 @@ local function new(args)
         __newindex = w
     })
 
+    -- Set border
+    if args.border_width then
+        ret.border_width = args.border_width
+        ret.border_color = args.border_color
+    end
+
     return ret
 end
 


psychon:
I think this must be a bug in the C code (I just don't know where, everything looks fine...). In line 110, all the `args` are passed to the `drawin` constructor and the C code should pick it up from there.
(Also, your patch fails when only a `border_color` is specified)

Elv13:
> (Also, your patch fails when only a border_color is specified)

Then there is no border, so why bother, nobody will set a `border_color` without a border. They can always call `border_color` later. No?

psychon:
But from `w = wibox({ border_color = "#ff0000" }) w.border_width = 10` I would expect a red border.

Is it ok for you if I take this over and fix this on the C side instead? Does my explanation of why I think this is a bug on the C side make sense?

Elv13:
Apparently, this is not really a CAPI problem. I recently added a metatable to the `awful.wibox` creation so it could read the "defaults" from `beautiful` (not commited). The problem is that CAPI use `lua_next()` to parse the table. This never hit the `wibox_args` fallback table. So, when I "forced" border width and color, it used the metatable, this would be true for the geometry and visible/ontop/below too.

```
    -- Create the wibox
    print("SET", wibox_args.border_color, wibox_args.border_width)
    local w = wibox(setmetatable({ position = "top"    },{__index=wibox_args}))
    print("GET", wibox_top[s].border_color, wibox_top[s].border_width)
```

```
SET	#00FF00	1
GET	#000000	0
```

Having beautiful fallback for `awful.wibox` is useful to create flexible themes, but apparently, I just did it wrong. Closing there is no "bug", just a limitation. Also, before you say I can implement more of the metatable to make it work: no, thanks to lua 5.1.

PR #655, opened 2016-01-26
Upstream shape api part 2.5 and 3

Elv13:
This replace the other pull request.

* More shapes are added
* Use `gears.matrix` directly instead of cairo.matrix
* Remove the wibox `shape_clip` commit, it is pointless, better use the wibox `after_draw_children`, it produce a much nicer effect. The shapes are usually not rectangles, so it always produce an ugly pixelated output.
* Remove the ... args from imagebox constructor
* Add `rotate_at` to `gears.matrix` and allow it to keep a `__call` function (undocumented hack used by `gears.shape`, `gears.pattern` and `gears.pixmap`)
* Add a new method to `gears.surface` to build from iconified vector fonts.

### DIFF
--- a/lib/gears/matrix.lua
+++ b/lib/gears/matrix.lua
@@ -52,6 +52,22 @@ function matrix.create_rotate(angle)
     return matrix.create(c, s, -s, c, 0, 0)
 end
 
+--- Create a matrix copy
+-- @return A new matrix identical to `other`
+function matrix:copy()
+    local m = matrix.create(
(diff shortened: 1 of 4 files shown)


Elv13:
Updated. I removed the surface from text commit from this and moved it to the part 4 as it require more non-trivial work. This version add matrix:copy() instead of hardcoding an identity matrix in `shape.lua`

psychon:
Looks good to me, thanks!

PR #657, opened 2016-01-27
Add `screen::change` signal for Xrandr output change notification events

baguette:
This lets users connect to a signal emitted when a monitor is plugged in or unplugged from the system. Here's an example of its use (from my `rc.lua`):

```lua
awesome.connect_signal("screen::change", function(output, state)
    naughty.notify{title="Output changed", text=output.." "..state}
    if output == 'HDMI1' then
        if state == 'Connected' then
            awful.util.spawn_with_shell("xrandr --output HDMI1 --auto --right-of eDP1 --primary")
        elseif state == 'Disconnected' then
            awful.util.spawn_with_shell("xrandr --output eDP1 --auto --primary")
            awful.util.spawn_with_shell("xrandr --output HDMI1 --off")
        end
    end
end)
```

### DIFF
--- a/event.c
+++ b/event.c
@@ -784,6 +784,52 @@ event_handle_randr_screen_change_notify(xcb_randr_screen_change_notify_event_t *
     awesome_restart();
 }
 
+/** XRandR event handler for RRNotify subtype XRROutputChangeNotifyEvent
+ */
+static void
+event_handle_randr_output_change_notify(xcb_randr_notify_event_t *ev)
+{
+    if(ev->subCode == XCB_RANDR_NOTIFY_OUTPUT_CHANGE) {
+        xcb_randr_output_t output = ev->u.oc.output;
+        uint8_t connection = ev->u.oc.connection;
+        char *output_name = NULL;
+        const char *connection_str = NULL;
+        xcb_randr_get_output_info_reply_t *info = NULL;
+        lua_State *L = globalconf_get_lua_State();
+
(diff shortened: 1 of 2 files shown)


psychon:
Nice idea, I like it! :+1:

baguette:
I've been looking through some other awesome code, and I wonder if I shouldn't change `free()` to `p_delete()` now that I'm using `p_dup()`, or does it matter?

blueyed:
@psychon should now.. :)

psychon:
Feel free to use `p_free(&output_name);` instead of `free(output_name);`. I always forget to use `p_free`, too, and the only difference is that it sets the variable to `NULL` after freeing.

So: No, it does not really matter, but yeah, you might want to do that.

Also, could you squash all your commits into a single one and force-push the result to your branch? The git history got a little complicated here.

baguette:
All squashed!

psychon:
Looks fine to me, thanks! :+1: 
(I just noticed that Lua has `lua_pushlstring()` which is "pushstring + a length argument", that would avoid all the `p_dup()` stuff and you could just do `lua_pushlstring(L, xcb_randr_get...(), xcb_randr_get...())`. However, this feels like overkill and the current version is fine as well.)

Oh and this also needs a `p_delete(&info)` at the end of the function (memory leak). That one should be fixed since it counts as a bug.

baguette:
Fixed.

blueyed:
Thanks!

PR #658, opened 2016-01-28
Force recalculation of coordinates. Fixes #248

### LINKED ISSUE #248: Java JMenuBar gets incorrect focus
With latest git build of awesome, my Java applications are bugged.

It looks like focus in open JMenuBar is shifted.

Steps to reproduce:
1. Find application with JMenuBar (or compile an example).
2. Center it in your screen, so that shifted focus will be visible
3. Try maximizing it or resizing it from top-left corner. Focus in menu will be shifted. Focus is restored, when you move the window

petoju:
Changing the order of actions leads to correct interpretation of offset in Java JMenuBar.

This patch was written without deep knowledge of all possible interpretations of graphic subsystem - it works for me, but I cannot say anything about programs I don't use.

With this patch, all "wmname" hacks are obsolete; JMenuBar works with them as well as without them.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -2,6 +2,7 @@
  * client.c - client management
  *
  * Copyright © 2007-2009 Julien Danjou <julien@danjou.info>
+ * Copyright © 2016 Google Inc, Peter Junos <junosp@google.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -128,7 +129,7 @@
 
 static area_t titlebar_get_area(client_t *c, client_titlebar_t bar);
 static drawable_t *titlebar_get_drawable(lua_State *L, client_t *c, int cl_idx, client_titlebar_t bar);
-static void client_resize_do(client_t *c, area_t geometry, bool force_notice);
+static void client_resize_do(client_t *c, area_t geometry, bool force_notice, bool honor_hints);
 
 /** Collect a client.
  * \param L The Lua VM state.
@@ -481,7 +482,7 @@ border_width_callback(client_t *c, uint16_t old_width, uint16_t new_width)
         geometry.x += diff_x;
         geometry.y += diff_y;
         /* force_notice = true -> inform client about changes */
-        client_resize_do(c, geometry, true);
+        client_resize_do(c, geometry, true, false);
     }
 }
 
@@ -814,13 +815,18 @@ client_apply_size_hints(client_t *c, area_t geometry)
 }
 
 static void
-client_resize_do(client_t *c, area_t geometry, bool force_notice)
+client_resize_do(client_t *c, area_t geometry, bool force_notice, bool honor_hints)
 {
     lua_State *L = globalconf_get_lua_State();
     bool send_notice = force_notice;
     bool hide_titlebars = c->fullscreen;
 
     screen_t *new_screen = c->screen;
+
+    if (honor_hints) {
+        geometry = client_apply_size_hints(c, geometry);
+    }
+
     if(!screen_coord_in_screen(new_screen, geometry.x, geometry.y))
         new_screen = screen_getbycoord(geometry.x, geometry.y);
 
@@ -933,15 +939,12 @@ client_resize(client_t *c, area_t geometry, bool honor_hints)
     if(geometry.width == 0 || geometry.height == 0)
         return false;
 
-    if (honor_hints)
-        geometry = client_apply_size_hints(c, geometry);
-
     if(c->geometry.x != geometry.x
        || c->geometry.y != geometry.y
        || c->geometry.width != geometry.width
        || c->geometry.height != geometry.height)
     {
-        client_resize_do(c, geometry, false);
+        client_resize_do(c, geometry, true, honor_hints);
 
         return true;
     }
@@ -1119,7 +1122,7 @@ client_set_fullscreen(lua_State *L, int cidx, bool s)
         luaA_object_emit_signal(L, abs_cidx, "request::fullscreen", 1);
         luaA_object_emit_signal(L, abs_cidx, "property::fullscreen", 0);
         /* Force a client resize, so that titlebars get shown/hidden */
-        client_resize_do(c, c->geometry, true);
+        client_resize_do(c, c->geometry, true, false);
         stack_windows();
     }
 }
@@ -1939,7 +1942,7 @@ titlebar_resize(lua_State *L, int cidx, client_t *c, client_titlebar_t bar, int 
     }
 
     c->titlebar[bar].size = size;
-    client_resize_do(c, geometry, true);
+    client_resize_do(c, geometry, true, false);
 
     luaA_object_emit_signal(L, cidx, property_name, 0);
 }


psychon:
My first impression was "this doesn't change anything!", but I was wrong. This should call `client_send_configure(c)` in more case. I'm curious, how did you come up with this?

Sorry, but for the patch you get a NAK, but I would like to know exactly why it fixes something. Could you test if the following patch also fixes this problem? It prints a backtrace in the situations that it does something differently than before and should cause the same change in behaviour that your patch does. Hopefully(?) this allows to figure out what exactly is going on.
```diff
--- client.c.orig	2016-01-29 08:46:00.334673006 +0100
+++ client.c	2016-01-29 08:50:30.469999521 +0100
@@ -46,6 +46,7 @@
 #include "spawn.h"
 #include "systray.h"
 #include "xwindow.h"
+#include "common/backtrace.h"
 
 #include <xcb/xcb_atom.h>
 #include <xcb/shape.h>
@@ -926,12 +927,12 @@ client_resize(client_t *c, area_t geomet
         geometry.y = 0;
 
     if(geometry.width < c->titlebar[CLIENT_TITLEBAR_LEFT].size + c->titlebar[CLIENT_TITLEBAR_RIGHT].size)
-        return false;
+        goto return_false;
     if(geometry.height < c->titlebar[CLIENT_TITLEBAR_TOP].size + c->titlebar[CLIENT_TITLEBAR_BOTTOM].size)
-        return false;
+        goto return_false;
 
     if(geometry.width == 0 || geometry.height == 0)
-        return false;
+        goto return_false;
 
     if (honor_hints)
         geometry = client_apply_size_hints(c, geometry);
@@ -946,6 +947,14 @@ client_resize(client_t *c, area_t geomet
         return true;
     }
 
+return_false:
+    if (true) {
+        buffer_t buf;
+        backtrace_get(&buf);
+        warn("Sending a configure notify due to:\n%s", buf.s);
+    }
+    client_send_configure(c);
+
     return false;
 }
 
```

JanMarvin:
@psychon, just tested your patch with awesome 3.5.7 and it seems to work fine.

petoju:
> I'm curious, how did you come up with this?

I ran simple Java JMenuBar app with xtrace multiple times, so that I could find X commands responsible for setting incorrect coordinates and another for correcting that error. Afterwards, I found the state of awesome when sending this X commands and edited it in a simplest possible way, so that correcting command is sent after the corrupting command when resizing.

I don't know protocol specification, so I don't have any general idea, why is that one X command harming Java and why should the other command work. But it works and that is necessary for me; otherwise, awesome 3.5 is close to unusable for Java folks.

Your patch also works, so it would be nice, if you could bring that or any alternative to master. Backtrace is in [log.zip](https://github.com/awesomeWM/awesome/files/110389/log.zip)

psychon:
Ok, I bite the bullet. Let's install java and compile your example (https://raw.githubusercontent.com/petoju/awesome-issue-248/master/SimpleMenuEx.java):
```
$ javac SimpleMenuEx.java
Error occurred during initialization of VM
Unable to allocate 26624KB bitmaps for parallel garbage collection for the requested 851968KB heap.
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```
My motivation raises...

Anyway, I start this program on a floating tag and open the FileA-menu. Hovering over items seems to work fine. Then I do `sleep 2 ; echo 'local g = client.focus:geometry() ; client.focus:geometry{ y = g.y + 5, x = g.x + 5, width = g.width - 5, height = g.height - 5 }'  | awesome-client` and focus the window (simulate resize be corner). Afterwards, opening the menu highlights a slightly wrong entry.

TL;DR: I can reproduce the issue.

I'll look into this, but if this turns out to be Java breaking on purpose, I won't be amused...

psychon:
The bug is in Java and IMO they should fix it on their side. A simple work-around in awesome would be to force `send_notify = true` in `client_resize_do()` (feel free to try this).
For debugging, the above SimpleMenuEx can be run like this:
```sh
( echo "handlers=java.util.logging.ConsoleHandler" ; echo "java.util.logging.ConsoleHandler.level=FINER" ; echo ".level=FINER" ) > /tmp/java-log-conf &&  java -Djava.util.logging.config.file=/tmp/java-log-conf SimpleMenuEx 2>&1  | grep 'new dimension'
```
This gives messages about the position that Java thinks its windows have.
Doing this (and then quickly focusing the Java window) does *not* cause a problem:
```
$ sleep 2 ; echo 'local g = client.focus:geometry() ; client.focus:geometry{ y = g.y + 5, x = g.x + 5 }'  | awesome-client
```
However, this does cause a problem:
```
$ sleep 2 ; echo 'local g = client.focus:geometry() ; client.focus:geometry{ y = g.y + 5, x = g.x + 5, width = g.width - 5, height = g.height - 5 }'  | awesome-client
```
The difference is "move without resize" vs "move with resize". The relevant part of the standard is ICCCM § 4.1.5 "Configuring the window" http://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#configuring_the_window. This basically says "if a window is moved without being resized, the WM has to send a synthetic configure notify to it (this is what `client_send_configure()` does), otherwise the X11 server automatically sends a non-synthetic configure notify. Also, there is an "Advice to implementors":

> Clients cannot distinguish between the case where a top-level window is resized and moved from the case where the window is resized but not moved, since a real ConfigureNotify event will be received in both cases. Clients that are concerned with keeping track of the absolute position of a top-level window should keep a piece of state indicating whether they are certain of its position. Upon receipt of a real ConfigureNotify event on the top-level window, the client should note that the position is unknown. Upon receipt of a synthetic ConfigureNotify event, the client should note the position as known, using the position in this event. If the client receives a KeyPress, KeyRelease, ButtonPress, ButtonRelease, MotionNotify, EnterNotify or LeaveNotify event on the window (or on any descendant), the client can deduce the top-level window's position from the difference between the (event-x, event-y) and (root-x, root-y) coordinates in these events. Only when the position is unknown does the client need to use the TranslateCoordinates request to find the position of a top-level window.

Of course, Java does not do this: http://hg.openjdk.java.net/jdk7/jdk7/jdk/file/9b8c96f96a0f/src/solaris/classes/sun/awt/X11/XDecoratedPeer.java#l720
For synthetic configure notifies, it updates its location correctly (line 722). For non-synthetic ones it just assumes that the position did not change (line 719 fetches the old location, 747 "moves this" to the new location). I really like the comment from lines 724 to 728 claiming that some WMs are buggy, because they do not send a synthetic configure notify when the window is resized from the top or left (doesn't this sound similar? Just like the problem we are looking at!). Java being Java, they have a list of WMs where they then use TranslateCoordinates as in the "advice to implementors" above (this is what `queryXLocation()` does internally). However, awesome is detected as a WM of type "3" which seems to be `OTHER_WM` (http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/sun/awt/X11/XWM.java#91) and `OTHER_WM` isn't in the list of WMs for which Java suddenly does the right thing.

Coming back to the patches that we have that fixes the problem. What they do is to make awesome send a synthetic configure notify in more cases and Java then gets its window position correct.

Fun fact: As explained in the "Advice to implementors" above, any input events contain the mouse position relative to the window's (0,0) and in root window coordinates and clients can use this to get their position right. However, Java ignores the relative position and uses the root window coordinates plus its idea of its own window position to come up with wrong values. I might add: needlessly! (Seriously, WHY?!?)

We could be generous and always send synthetic configure notifies like some of the patches here are doing. I would even consider doing this if this were some almost-well-behaved program. However, Java is annoying on purpose. Whoever wrote that code didn't look at standards but just made things work with whatever WM they were using. When other WMs did things differently, the Java code wasn't fixed to do the right thing, but instead the type of WM was detected and various band aids applied based on this. So no, I don't want hacks in awesome just to work around Awesome's brokeness. Hence, I'll close this and the corresponding issue. Sorry, but Java.

Feel free to report this as a bug to Java (Openjdk?). If you want, I can also write up a better explanation of the problem. If you report the bug, please leave a link here.

PR #668, opened 2016-02-05
feat(awful: spawn): add 'easy_async' wrapper

actionless:


lib/awful/spawn.lua
```diff
@@ -139,6 +139,58 @@ function spawn.with_line_callback(cmd, callbacks)
     return pid
 end
 
+--- Spawn a program and asynchronously and capture its output
```
psychon: I'm not really good at English, but shouldn't it be: `Asynchronously spawn a program and capture its output`?

psychon:
Fine by me: :+1:

psychon:
@blueyed Should look at the docs for this one. Perhaps he can come up with a nicer explanation of the callback arguments. I added this to the "next" milestone for now.

actionless:
@blueyed so what about the docstrings?

psychon:
I merged this as-is. If there is a better way to document this, we can change that later.

blueyed:
I think the documentation is fine: http://awesomewm.github.io/apidoc/libraries/awful.spawn.html#easy_async.

psychon:
Wow, that looks surprisingly good and definitely not what I expected. Sorry for my confusion.

PR #671, opened 2016-02-06
Number vs int

psychon:
This makes the C code use `lua_pushinteger` and `luaA_opt_integer` in more places where it previously used the `number`-version instead. I was careful to allow e.g. setting a client's width to 42.5. Requiring an integer here would break lots of code for no good reason.

The idea for this is from one of @actionless's recent pull requests where he noticed that the `spawn`-API provides numbers instead of integers as exit codes, but I can't find that comment right now...


actionless:
thanks for fixing this, i've also updated docstring in easy-async PR to from number to integer

PR #675, opened 2016-02-07
Remove gears.sort

psychon:
It's unused since commit 0aa4304bda4069e. Before this was a stable sorting
algorithm since table.sort is allowed to be unstable. Apparently we don't need a
stable sorting algorithm anymore.


psychon:
Adding to "next" milestone since no one complained. Now is your time to do so, if you want! ;-)

blueyed:
You could argue that it might be useful for some config, and it has tests etc.
There's not much overhead in keeping it, is there?

But feel free to remove it.

psychon:
If we need it again, we can resurrect it from git, but I don't like having lots of code just because "code". I also want to get rid of the mouse finder module....

(Plus, Bubblesort is evil ;-) )

PR #678, opened 2016-02-08
Extend bgimage support

Elv13:
This PR still need some work, but generally seem to cause no regressions

Screenshot:
http://i.imgur.com/1QVpnXG.png

In this screenshot, made last week, everything is done using programmatic `bgimage`. A titlebar like this was previously impossible, but now take only ~20 line of code. Also, The `bgimage` function allow the button background to be created with the right width and height using on top of the current API using the default tasklist widget, making the old hack obsolete. This PR doesn't remove the old `awful.widget.common` hack, I want to see how much you hate it first.

I made that code last week, but I wasn't sure myself about it, but it solve real limitations and add a very powerful API without breaking too much in the process.

Edit:

By the way, this is an example of a simple titlebar with rounded corner (+ antialiasing)

```lua
local cols = {
   [true] = color("#ff0000"),
   [false] = color("#00ff00")
}

local bgs = {
    top = function(context, cr, w, h)
        cr:translate(1,1)
        shape.rounded_rect(cr, w-2, h*3, 12)
        cr:set_source(cols[context.has_focus])
        cr:set_line_width(3)
        cr:stroke()
        cr:translate(-2,-2)
    end,
    bottom = function(context, cr, w, h)
        cr:translate(1,-3*h+ 17)
        shape.rounded_rect(cr, w-2, h*3, 12)
        cr:set_source(cols[context.has_focus])
        cr:set_line_width(3)
        cr:stroke()
        cr:translate(-1,-2)
    end,
    left = function(context, cr, w, h)
        cr:set_source_rgb(0,0,0)
        cr:paint()
        cr:set_source(cols[context.has_focus])
        cr:rectangle(0,0,2,h)
        cr:fill()
    end,
    right = function(context, cr, w, h)
        cr:set_source_rgb(0,0,0)
        cr:paint()
        cr:set_source(cols[context.has_focus])
        cr:rectangle(w-2,0,2,h)
        cr:fill()
    end,
}
theme.titlebar_bgimage = function(context, cr, w, h)
    bgs[context.position](context, cr, w, h)
end
```

### DIFF
--- a/lib/awful/titlebar.lua
+++ b/lib/awful/titlebar.lua
@@ -89,6 +89,7 @@ local function new(c, args)
             local args = bars[position].args
             ret:set_bg(get_color("bg", c, args))
             ret:set_fg(get_color("fg", c, args))
+            ret:set_bgimage(get_color("bgimage", c, args))
         end
 
         bars[position] = {
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -22,6 +22,7 @@ local timer = require("gears.timer")
 local matrix = require("gears.matrix")
 local hierarchy = require("wibox.hierarchy")
 local base = require("wibox.widget.base")
+local unpack = unpack or table.unpack
 
 local drawables = setmetatable({}, { __mode = 'k' })
 local wallpaper = nil
@@ -70,18 +71,19 @@ local function do_redraw(self)
     local cr = cairo.Context(surf)
     local geom = self.drawable:geometry();
     local x, y, width, height = geom.x, geom.y, geom.width, geom.height
+    local context = get_widget_context(self)
 
     -- Relayout
     if self._need_relayout or self._need_complete_repaint then
         self._need_relayout = false
         if self._widget_hierarchy and self.widget then
-            self._widget_hierarchy:update(get_widget_context(self),
+            self._widget_hierarchy:update(context,
                 self.widget, width, height, self._dirty_area)
         else
             self._need_complete_repaint = true
             if self.widget then
                 self._widget_hierarchy_callback_arg = {}
-                self._widget_hierarchy = hierarchy.new(get_widget_context(self), self.widget, width, height,
+                self._widget_hierarchy = hierarchy.new(context, self.widget, width, height,
                         self._redraw_callback, self._layout_callback, self._widget_hierarchy_callback_arg)
             else
                 self._widget_hierarchy = nil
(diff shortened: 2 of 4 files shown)


psychon:
> This PR doesn't remove the old awful.widget.common hack, I want to see how much you hate it first.

I wouldn't complain about a PR removing that. At all. As you say, it's a bad hack.

lib/wibox/drawable.lua
```diff
@@ -128,12 +130,25 @@ local function do_redraw(self)
 
     cr:set_source(self.background_color)
     cr:paint()
+
+    -- Paint the background image
+    if self.background_image then
+        if type(self.background_image) == "function" then
+            --TODO if opaque, cache this
```
psychon: Why? I don't think caching this will be significantly faster and if every drawable caches this, the memory usage might be significant.

I don't agree with this "TODO".

lib/awful/titlebar.lua
```diff
@@ -84,11 +85,13 @@ local function new(c, args)
             client = c,
             position = position
         }
+        args.context = context
```
psychon: ...? This sets `args.context` to nil, right?

Elv13: no, it is defined at line 84

lib/awful/titlebar.lua
```diff
@@ -101,6 +104,7 @@ local function new(c, args)
         c:connect_signal("focus", update_colors)
         c:connect_signal("unfocus", update_colors)
     else
+        args.context = {}
```
psychon: This seems wrong, why does it do this?

Elv13: Not the cleanest code on earth, but it is a placeholder until the "real" context is created.

lib/awful/titlebar.lua
```diff
@@ -34,7 +34,8 @@ local all_titlebars = setmetatable({}, { __mode = 'k' })
 -- Get a color for a titlebar, this tests many values from the array and the theme
 local function get_color(name, c, args)
     local suffix = "_normal"
-    if capi.client.focus == c then
+    args.context.has_focus = capi.client.focus == c
```
psychon: I think the idea here (and your example seems to imply so, too) is that widgets get called with a context that has a `has_focus` entry. However, this does not work.

The function `get_widget_context()` in `wibox.drawable` makes a copy of the `_widget_context_skeleton` that you gave to the `drawable` constructor. This copy will not be updated just because you change the variable here. And even if it were updated, you'd still have to cause a redraw for the whole wibox so that widgets are, well, redrawn with the new context.

Since changing the context implicitly clears the various caches, I don't think that this is a viable approach. Individual widgets already have `context.client`. Adding a `context.has_focus` doesn't *really* add any new information, but causes many issues.

lib/awful/titlebar.lua
psychon: Sorry, I was too slow at deleting my comment, but also too fast (I noticed that it is defined, but deleted the comment after you already replied) :-)

lib/awful/titlebar.lua
Elv13: ah, right, I will remove it, I was thinking it would be nicer, but updating the context, umm, context, seem like too much trouble.

lib/awful/titlebar.lua
psychon: Well, the context is already created.

Users can cause `awful.titlebar.new()` to be called multiple times on the same titlebars. The first call executes the "true"-part of this if and causes a new drawable to be created. The following calls only end up in this "false" part. No new drawable is created here (at one point I had it create multiple drawables and the results were really bad...).

So no, this line doesn't do what you think it does.

(And yes, I guess it might make sense to explain the above in a comment in the code...)

Elv13:
Updated

psychon:
:+1: Thanks!

Elv13:
thanks

PR #682, opened 2016-02-10
chore(awful: hotkeys: vim): small improvements in hotkeys descriptions

actionless:


actionless:
i think i will also add `Ctrl+` bindings too

actionless:
i will do more changes tomorrow

psychon:
While you are at it, do we have some docs/example on how to enable these Vim descriptions? Perhaps as a comment in the default config, or something?

actionless:
they're enabled automatically when the rule matches: https://github.com/actionless/awesome/blob/vim-bindings-improvements/lib/awful/hotkeys_popup/keys/vim.lua#L12

psychon:
Ah, I didn't know that. Does the rule match gvim? (Yup, I should just try this...)

actionless:
here is the fix for that: https://github.com/awesomeWM/awesome/pull/685

actionless:
i think i am done for now

psychon:
Fine by me (= doesn't look like it can break something): :+1:

blueyed:
Thanks!

PR #684, opened 2016-02-11
wibox.widget: Add method to create declarative widgets directly

Elv13:
It was possible with the previous commit, but the more of my modules I port, the uglier I find this corner case boilerplate code.

### DIFF
--- a/lib/wibox/widget/base.lua
+++ b/lib/wibox/widget/base.lua
@@ -437,6 +437,31 @@ function base.widget:setup(args)
     rawset(self, "get_children_by_id", get_children_by_id)
 end
 
+--- Create a widget from a declarative description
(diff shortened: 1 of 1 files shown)


lib/wibox/widget/base.lua
```diff
@@ -437,6 +437,31 @@ function base.widget:setup(args)
...
+
+    local w, id = drill(ids, args)
+
+    local mt = {}
+    local orig_string = tostring(ret)
+
+    rawset(w, "_by_id", ids)
+    rawset(w, "get_children_by_id", get_children_by_id)
+
+    mt.__tostring = function(o)
```
psychon: Unused argument `o` (and I only notice this because Luacheck made me fix every instance of this...)

psychon:
Fine by me: :+1:

PR #685, opened 2016-02-11
feat(awful: hotkeys): extended matching rules

actionless:


lib/awful/hotkeys_popup/keys/vim.lua
```diff
@@ -9,12 +9,12 @@
 
 local hotkeys_popup = require("awful.hotkeys_popup.widget")
 
-local vim_rule = {name="vim"}
+local vim_rule_any = {name={"vim", "VIM"}}
```
blueyed: Could this use some globbing, matching the start or end?

For me a vim instance looks like this:

> WM_NAME(COMPOUND_TEXT) = "✐ [No Name] (~/src/awesome) - vim"

psychon: That should already match, no? `awful.rules` uses `string.match()`.

blueyed: Ok then, I've feared to miss something like that.

PR #673, opened 2016-02-06
Add code coverage support

psychon:
This uses coveralls.io and integrates the collection of coverage information
with Travis.

Right now this only tests unit tests. Examples for the results are at https://coveralls.io/builds/4991426.

PR #686, opened 2016-02-11
Add code coverage support

psychon:
Having messed up https://github.com/awesomeWM/awesome/pull/673 badly, here is a new attempt that makes me look less stupid:

This runs the unit and functional tests under LuaCov and uploads the results to coveralls.io. This service provides web access to the coverage information and shows the coverage change on any PR.

### DIFF
--- a/.luacov
+++ b/.luacov
@@ -13,6 +13,10 @@ return {
 
     include = {
         escape_pattern(lib_dir) .. ".+",
+
+        -- For things already having the correct path
+        -- (happens with integration tests)
+        "^lib/",
     },
 
     -- configuration for luacov-coveralls reporter
--- a/.luacov
+++ b/.luacov
@@ -1,0 +1,26 @@
+-- Configuration file for LuaCov
+
+-- This variable is set externally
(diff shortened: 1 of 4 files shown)


psychon:
Updated for @blueyed's suggestions from https://github.com/psychon/awesome/commit/e8494fd5cfb8796766391da6a93d166cae1fcebd.
Apparently the new `test-leak-client.lua`-test fails when run under LuaCov. I will have to investigate this, but not today (at least I already know that this can be reproduced locally).

blueyed:
Have you also considered codecov.io?

psychon:
@blueyed My approach was "Oh, https://github.com/moteus/luacov-coveralls looks shiny, let's use that". So no, I haven't really considered anything. Does anyone know about a good way to find alternative projects that take output from LuaCov and upload that somewhere useful? (If we don't find anything like that, then coveralls.io wins because there are no alternatives :-) )

psychon:
I fixed the test failure (and rebased this on master). The problem was that the wibox wasn't updated yet when the leak check was run and thus the tasklist prevented garbage collection of the client.

Issue #689, opened 2016-02-12
Travis doesn't notice some errors


psychon:
Some quote from https://travis-ci.org/awesomeWM/awesome/jobs/108660087 (a successful travis build):

> == Running /home/travis/build/awesomeWM/awesome/tests/test-leak-client.lua ==
> [...]
> 2016-02-11 22:39:46 W: beautiful: can't get colorscheme from xrdb (using fallback).
> 2016-02-11 22:39:47 W: awesome: luaA_dofunction:77: error while running function
> stack traceback:
>	lib/awful/titlebar.lua:117: in function <lib/awful/titlebar.lua:72>
>	(...tail calls...)
>	[string "-- Some memory leak checks involving clients ..."]:30: in function <[string "-- Some memory leak checks involving clients ..."]:6>
> error: lib/awful/titlebar.lua:117: attempt to index global 'base' (a nil value)
> +grep -q -E ^Error|assertion failed /tmp/tmp.ogV2z6N0Cu/_awesome_test.log
> +echo == Running /home/travis/build/awesomeWM/awesome/tests/test-leaks.lua ==

The above should have made the build fail.

(This issue is not about the problem "titlebars are broken", but about the bug in the tests runner that hides this issue!)

blueyed:
The `grep` should look for `^error:` maybe, too?

psychon:
My plan is to turn the logic around. Instead of looking for errors, I want to do `print("Functional test finished successfully")` at the end of `_runner` and grep for that. The test fails unless that is present. Hopefully that should be all that is necessary.

Or do we have some kind of non-fatal errors that grepping for errors catches, but wouldn't make the test fail otherwise?

blueyed:
Sounds good!
No time for this currently though.

PR #690, opened 2016-02-12
Make the functional test runner more robust

psychon:
Currently, an error in the default config in the right place isn't noticed. Fix
this by doing two things:

- Also grep for "error" (this catches runtime errors with a stack trace)
- Make _runner print a "success" message at the end and also grep for that

Fixes: https://github.com/awesomeWM/awesome/issues/689

Note: Travis will fail due to the bug currently present in master. That's expected. If you want, I can rebase after that was fixed or this can just be merged in the hope that Travis becomes happy afterwards.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -169,9 +169,10 @@ for f in $tests; do
     # Tail the log and quit, when awesome quits.
     tail -n 100000 -f --pid $awesome_pid $awesome_log
 
-    if grep -q -E '^Error|assertion failed' $awesome_log; then
+    if ! grep -q -E '^Test finished successfully$' $awesome_log ||
+            grep -q -E '[Ee]rror|assertion failed' $awesome_log; then
         echo "===> ERROR running $f! <==="
-        grep --color -o --binary-files=text -E '^Error.*|.*assertion failed.*' $awesome_log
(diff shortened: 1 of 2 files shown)


tests/run.sh
```diff
@@ -169,9 +169,10 @@ for f in $tests; do
     # Tail the log and quit, when awesome quits.
     tail -n 100000 -f --pid $awesome_pid $awesome_log
 
-    if grep -q -E '^Error|assertion failed' $awesome_log; then
+    if ! grep -q -E '^Test finished successfully$' $awesome_log ||
+            grep -q -E '[Ee]rror|assertion failed' $awesome_log; then
```
blueyed: This one is not anchored anymore - but that might be intentional?!
But it should be the same as below.

psychon: Whoops. Yeah, that's intentional, but I forgot the one below. Better be too greedy when looking for error messages (I noticed that the output of `lua -e 'error("")'` wouldn't be matched. This would now match the function name "error".)

psychon: Done

psychon:
Rebased ontop of master (= ontop of the fix for broken titlebars). Thus, Travis doesn't complain any more.

PR #692, opened 2016-02-12
Remove awful.mouse.finder

psychon:
This module is partly broken since 2009 (the way to use it that is mentioned in
the docs doesn't actually work) because the mousefinder object doesn't have a
find()-method (the line doing "self.find = find" should do "self.finder =
finder.find"). Since no one really noticed, this module is apparently not used
much.

When someone wants to still use this, they are free to copy this to their own
config. It's not much code, but it's enough code that I am annoyed that we ship
something broken to users. Everyone who copies it to their own config will make
sure it works the way they want.


lib/awful/mouse/init.lua
```diff
@@ -25,8 +25,6 @@ local capi =
     mousegrabber = mousegrabber,
 }
 
-local finder = require("awful.mouse.finder")
```
psychon: Fun fact: I noticed all of this because Luacheck was complaining about the above line (variable `finder` is assigned a value that is never used).

PR #693, opened 2016-02-12
Enable titlebars by default

psychon:
This means that titlebars, which are an important feature, get more testing. As
was recently shown, they don't get enough testing currently. Also, new users
will likely expect titlebars these days.

Everyone who doesn't want titlebars can easily disable them.


Elv13:
+1

Issue #681, opened 2016-02-10
Support images in documentation


Elv13:
My first attempt apparently only work locally. I was kind of expecting this, but hoped it would work anyway.

http://awesomewm.github.io/apidoc/documentation/03-declarative-layout.md.html

Look like I will have to dive deeper into this

funeralismatic:
```
CMake Error at cmake_install.cmake:147 (file):
  file INSTALL cannot find
  "/var/tmp/portage/x11-wm/awesome-9999/work/awesome-9999_build/docs/images/widgetlayout1.png".
```

Elv13:
strange, I get

```
-- Installing: /tmp/dir/share/doc/awesome/images/widgetlayout1.png
-- Installing: /tmp/dir/share/doc/awesome/images/widgetlayout2.png
```

psychon:
There is no awesome-9999 in Gentoo: https://gitweb.gentoo.org/repo/gentoo.git/tree/x11-wm/awesome. @funeralismatic Where did you get the ebuild from? Also, do you have ldoc installed or are you building without docs? I guess this is the problem, because the copy-images-to-build-dir is inside of a `if(GENERATE_DOC)` while the `install()` command is executed always.

(However, the ebuild for 3.5.8 doesn't use our wrapper Makefile)

Elv13:
Indeed, patch coming

 cmake .. -DCMAKE_INSTALL_PREFIX=/tmp/cab -DGENERATE_DOC=false

sorry about that

Elv13:
Done

funeralismatic:
@psychon 

>Where did you get the ebuild from?

ebuild from [nightmare overlay](https://gitlab.com/r3lgar/nightmare/tree/master/x11-wm/awesome).

>Also, do you have ldoc installed or are you building without docs?

Without docs.

>I guess this is the problem, because the copy-images-to-build-dir is inside of a if(GENERATE_DOC)

All files are copied to build dir, only these two images are ignored.

Elv13:
@funeralismatic Is it fixed now? I pushed a commit yesterday that fixed `-DGENERATE_DOC=false`, at least on my side

funeralismatic:
@Elv13, all ok without docs.

Elv13:
@blueyed Can you update the robot to also copy the image folder? I don't seem to have access to it (and I don't need access, just add a line to copy the image folder at the right relative path).

psychon:
[...]

PR #694, opened 2016-02-13
Fix images in the documentation

psychon:
Instead of copying images to ${BUILD_DIR}/docs/images, which is the directory
with the input to ldoc, this now copies the images to ${BUILD_DIR}/doc/images,
which is where ldoc generates its output. That way, the images are together with
the html files and are automatically picked up by a lot of stuff.

Fixes: https://github.com/awesomeWM/awesome/issues/681

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,13 +34,6 @@ set(AWE_DOC_FILES
     ${AWE_DOC_DIR}/02-contributing.md
     ${SOURCE_DIR}/LICENSE)
 
-if (GENERATE_DOC)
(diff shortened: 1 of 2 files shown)


Elv13:
+1

psychon:
I'm still waiting for the travis build to finish; The diff that awesome-robot posts should show if this actually works.

psychon:
Heh, I "broke" the apidoc script. The images ended up in the "boilerplate" commit, but they are there: https://github.com/awesomeWM/apidoc/commit/be02580e5e18506633f6d7ae4075acb1589b6e42

So this should indeed fix #681.

PR #698, opened 2016-02-15
fix(awful: prompt): fix variable name after luacheck fixup

actionless:


psychon:
Sorry for breaking this and thanks for fixing my breakage.

Issue #697, opened 2016-02-14
Mplayer window drifting slowly in top left direction.


jk411:
Hi

Every time mplayer video window is set fullscreen/restored it is moved in top left direction.
It happens in floating mode.
With border_width = 0 video window stays in place.
Any other value sets the "step".
So far Ive noticed this behavior only with mplayer video window, xine and the rest of programs seems to be fine.

lib/awful/ewmh.lua:
71:    elseif data[window] and data[window].fullscreen then
72:        window:geometry(data[window].fullscreen)
73:        window.border_width = data[window].fullscreen.border_width
74:    end

Swapping line 72 with line 73 fixes problem but Ive no idea if it's the correct fix.

Awesome 3.5.8
complied from sources on slackware 14.1 x64
Thanks

psychon:
Confirmed on master. This was introduced in https://github.com/awesomeWM/awesome/commit/b2aaefd095c964491f2c93aa01ec7f0fcc9c340f. Before this, the top-left corner of a client stayed where it was when the border width changes. Now we apply window gravity and the top-left corner of the actual content window stays where it was.

So yeah, your proposed fixed is the correct one. From a quick look through `ewmh.lua`, line 77 has the same problem, but nothing else in this file. I'll have to carefully look at all places that set the border width. And I guess it would also make sense to add test cases for this....

Thanks for reporting a bug and providing the fix with it! :-)

PR #701, opened 2016-02-15
Fix awful.ewmh to handle window gravities

psychon:
Since commit b2aaefd095c964491f2c93a, we correctly handle window gravities when
the border width of a client changes. Since most windows out there have a
NorthWest gravity, this means that most windows do not have this problem.
However, e.g. mplayer uses gravity "Static" and this causes this issue (any
gravity other than NorthWest will do).

This affects the fullscreen handling in awful.ewmh. The code has to set the
border width before it changes a client's geometry so that the move when the
border width changes doesn't matter.

No new integration test for this since I didn't find anything usable with a
non-NorthWest gravity. A test would be easy to write, just test if `c.fullscreen
= true ; c.fullscreen = false` restores the previous window geometry.

Fixes: https://github.com/awesomeWM/awesome/issues/697

### DIFF
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -70,11 +70,11 @@ local function fullscreen(window, set)
     if set then
         store_geometry(window, "fullscreen")
         data[window].fullscreen.border_width = window.border_width
-        window:geometry(screen[window.screen].geometry)
         window.border_width = 0
+        window:geometry(screen[window.screen].geometry)
     elseif data[window] and data[window].fullscreen then
-        window:geometry(data[window].fullscreen)
         window.border_width = data[window].fullscreen.border_width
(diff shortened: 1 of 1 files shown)


blueyed:
> I didn't find anything usable with a non-NorthWest gravity

We could have `mplayer` for this on Travis though?! ;)

Maybe something via www.github.com/search?
E.g. https://github.com/kakaroto/e17/blob/93aef5b89a5e845d90fba2c209b548c1c1f553cf/TEST/app/e/e_place_test/src/TestGravityWindow.cpp.

However, it's good without a functional test for now though, too.

psychon:
I can easily write a small test app for this if you want. However, that would require many changes (e.g. make CMake compile this test app) and would turn this two line change into something huge, so I'll see if I can fix this without adding a regression test. :-)

PR #702, opened 2016-02-15
Fix test-urgent.lua

psychon:
Step 1 starts a client and uses awful.rules to move it to a not-selected tag.
Because this rule still has focus=true, this calls awful.ewmh.activate() via the
request::activate signal. This function makes the client urgent because it is on
a not-selected tag.

Step 3 does the same thing, but also uses switchtotag=true. Now
awful.ewmh.activate() doesn't make the client urgent because it successfully
focused this client. However, the test was wrongly assuming that the client
became urgent (copy&paste error? I don't know).

The fix is of course not to require the client to become urgent.

@blueyed Does the above sound sensible?

### DIFF
--- a/tests/test-urgent.lua
+++ b/tests/test-urgent.lua
@@ -69,7 +69,7 @@ local steps = {
   -- Step 3: tag 2 should not be urgent, but switched to.
   function(count)
     if count == 1 then  -- Setup.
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!

PR #709, opened 2016-02-27
Implement set children sanely

psychon:
This PR adds some calls to `base.check_widget()` in the code handling a declarative widget hierarchy . The idea is to check for "non-widgets" while constructing a widget hierarchy so that we get useful backtraces instead of "Some widget somewhere failed while rendering". This would have saved me some time while debugging an issue where a boolean ended up as a "widget" in a fixed layout.

(See https://github.com/psychon/awesome/commit/94419832fdd9d94f250bfdb794a6b80fcf38cfc2 / https://github.com/psychon/awesome/tree/stop-restarting-on-RandR-changes; The issue was that `s` is now a screen object and thus... I don't know, somehow a boolean (`false`?) ended up somewhere where it shouldn't be.)

CC: @Elv13 Objections?

P.S. `wibox.layout.ratio` does not implement `:set_children()`.

### DIFF
--- a/lib/wibox/layout/align.lua
+++ b/lib/wibox/layout/align.lua
@@ -175,11 +175,9 @@ end
 -- This layout only accept three children, all others will be ignored
 -- @tparam table children A table composed of valid widgets
 function align:set_children(children)
-    if not children then return self:reset() end
-    self.first  = children[1]
-    self.second = children[2]
-    self.third  = children[3]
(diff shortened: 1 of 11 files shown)


Elv13:
+1

Elv13:
> wibox.layout.ratio does not implement :set_children().

I will fix that, they were in 2 parallel PR, so there was a race condition ;)

PR #711, opened 2016-02-27
Menubar: recursively search for .desktop entries

blueyed:


lib/menubar/menu_gen.lua
```diff
@@ -90,34 +90,38 @@ function menu_gen.generate()
...
-                if program.categories then
-                    for _, category in pairs(program.categories) do
-                        local cat_key, cat_use =
-                            get_category_name_and_usage_by_type(category)
-                        if cat_key and cat_use then
-                            target_category = cat_key
-                            break
-                        end
+        for _, entry in ipairs(utils.parse_dir(dir)) do
+            entries[entry.Exec .. '\0' .. entry.Name] = entry
```
psychon: This key is unused, this is just so that you get a unique key, right?
I'd propose to use the following then instead:

    for _, dir in ipairs(menu_gen.all_menu_dirs) do
        awful.util.table.merge(entries, utils.parse_dir(dir))
    end

(This would also preserve "some" order of the elements)

blueyed: @psychon It's meant to remove duplicates.

psychon: Hm, ok. Fine by me then. (Where do duplicates come from? Why would something install the same .desktop file in different places? But ok...)

blueyed:
@psychon 
I've amended the previous commit with a (hopefully) better version.
Please review again.

psychon:
Still fine by me: :+1:

PR #712, opened 2016-02-28
Protected calls (gears.protected_call)

psychon:
This adds and uses `gears.protected_call` which is a wrapper around `pcall()` that also handles error reporting. This simplifies some code elsewhere. (Plus, this PR also has some random fixes so that `gears.fit_widget` and `gears.layout_widget` are more robust against errors from widgets).


blueyed:
Awesome!

psychon:
Thanks for merging all these pull requests! It's nice to have some look at this stuff instead of me just pushing random commits directly to master.

PR #717, opened 2016-02-28
Add callback support for client.jumpto

blueyed:
Instead of true/false `merge` can be a callback now, which allows for
more dynamic handling of the client not being visible, e.g. moving it to
the current tag, instead of merging.


Elv13:
+1

PR #719, opened 2016-02-29
xresources: Prevent tonumber(nil)

Elv13:
```lua
stack traceback:
	[C]: in function 'tonumber'
	...usr/local/share/awesome/lib/beautiful/xresources.lua:85: in function 'get_dpi'
	/usr/local/share/awesome/lib/beautiful/init.lua:55: in function 'load_font'
	/usr/local/share/awesome/lib/beautiful/init.lua:74: in function 'set_font'
	/usr/local/share/awesome/lib/beautiful/init.lua:155: in main chunk
	[C]: in function 'require'
	/home/elv13/.config/awesome/utils/tools.lua:7: in main chunk
	[C]: in function 'require'
	/home/elv13/.config/awesome/utils/init.lua:2: in main chunk
	[C]: in function 'require'
	/home/elv13/.config/awesome/rc.lua:8: in main chunk
error: ...usr/local/share/awesome/lib/beautiful/xresources.lua:85: bad argument #1 to 'tonumber' (value expected)
error while running function
```

### DIFF
--- a/lib/beautiful/xresources.lua
+++ b/lib/beautiful/xresources.lua
@@ -82,7 +82,7 @@ function xresources.get_dpi(s)
     if not xresources.dpi then
         -- Might not be present when run under unit tests
         if awesome and awesome.xrdb_get_value then
-            xresources.dpi = tonumber(awesome.xrdb_get_value("", "Xft.dpi"))
+            xresources.dpi = tonumber(awesome.xrdb_get_value("", "Xft.dpi") or 96)
         end
         if not xresources.dpi then
             xresources.dpi = 96


psychon:
Which Lua version is this? I don't get any error (and so does the Lua on Travis)
```
$ lua -e 'print(tonumber(nil))'
nil
```

Elv13:
5.1. 

    $ lua -e 'print(tonumber(nil))'
    nil

Work for me too, but not at runtime (see the copy pasted as-is error). This is strange, but umm, without DPI is has issues down the road anyway, so I guess the patch is still better than nothing. A quick git-grep show no magic hack in my config beside some `local tonumber = tonumber`.

psychon:
Ohhhh!
```
$ lua -e 'print(tonumber(nil))'
nil
$ lua -e 'print(tonumber())'   
lua: (command line):1: bad argument #1 to 'tonumber' (value expected)
stack traceback:
	[C]: in function 'tonumber'
	[C]: in ?
$ 
```
Should this be fixed in the C code so that `awesome.xrdb_get_value()` returns `nil` instead of nothing when it didn't find any value? This might be the better fix....
(And it's easy: Just add a `lua_pushnil(L)` and `return 1` instead of `return 0` at the end of xrdb.c; I can't do this right now (can't test a build of awesome here), but feel free to do this if you want).

(Oh and if the answer to the above question is "no", then this pull request wins. But only then.)

Elv13:
But it is still "better" to have a DPI (96 is the "classic" default) than none right? This simplify the code down the road. So maybe "both" is the right answer for this problem? Fix the C code **and** always have a default DPI so widgets don't have to check (unless this is set elsewhere, I did not check and I think this would be the right place to do so)

psychon:
> (unless this is set elsewhere, I did not check and I think this would be the right place to do so)

Look in the next three lines after the line that your patch touches. :-P

actionless:
actually it will be better to apply this change and remove those 3 lines

psychon:
@actionless Why? That won't work, because in the unit tests the DPI will end up being `nil` ("making `beautiful.xresources` work (=not break) when `awesome.xrdb_get_value()` is not available" is the whole reason why the `or 96` was removed and instead the next three lines were added.

Or do I miss something?

actionless:
i've updated my comment few seconds before you posted the answer :)

PR #720, opened 2016-02-29
awful.layout: Allow parameters to be called when no tags is selected

Elv13:
There seem to be a little race condition (either in my layout code or
elsewhere) when playing with multiple screens. As most properties do
not depend on the tag, there is no point in returning early anyway.

It just remove a not so necessary limitation. It isn't a bug fix.

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -132,15 +132,13 @@ function layout.parameters(t, screen)
     screen = get_screen(screen)
     t = t or tag.selected(screen)
 
-    if not t then return end
-
-    screen = get_screen(tag.getscreen(t) or 1)
+    screen = get_screen(t and tag.getscreen(t) or 1)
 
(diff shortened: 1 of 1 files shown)


psychon:
This feels like it tries to hide bugs that are elsewhere instead of fixing them. For `t` to be `nil`, the function argument has to be `nil` and `tag.selected()` has to return `nil`, so nothing is selected...

However, I agree that this function sometimes returning `nil` instead of a table is bad, so: :+1:

Elv13:
> This feels like it tries to hide bugs that are elsewhere 

Actually, there is some Tyrannical users (not me) who do that on purpose. They start with no tags at all and create them dynamically. So not only is there no selected tags, there is no tags at all. You might still want to get some parameters for some reasons. I don't know, selecting the tag layout dynamically before creating the tag object (strange, but why not)

PR #721, opened 2016-02-29
awful.screen: Fix regression

Elv13:
a missing `capi.`

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -17,7 +17,7 @@ local capi =
 local util = require("awful.util")
 
 local function get_screen(s)
-    return s and screen[s]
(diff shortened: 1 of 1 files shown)


psychon:
Sorry for missing that... but why is it a regression? The original `screen` should still be available in the global env...

Anyway: The same issue exists in this file in the `get_screen` function and the function you are fixing does `screen[i]` a few lines down. Could you fix both of these as well in this PR?

Elv13:
> The original screen should still be available in the global env...

`local screen = {}` nope ;)

Elv13:
Updated, ready?

psychon:
Sounds like we need more multi-monitor tests on Travis... Yup, looks fine.

PR #722, opened 2016-02-29
awful.widget.graph: Expose private data

Elv13:
@mindeunix If I recall, you asked for this. Please test (my testing are limited to "it seem to work")

### DIFF
--- a/lib/awful/widget/graph.lua
+++ b/lib/awful/widget/graph.lua
@@ -17,8 +17,6 @@ local base = require("wibox.widget.base")
 
 local graph = { mt = {} }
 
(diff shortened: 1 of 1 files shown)


mindeunix:
Works great,  thanks.

psychon:
I'm fine with this since it removes a pointless use of some weak table, thus making everything saner. However, "expose private data" sounds bad. If the API is missing something, the missing API should be added instead of turning all of the implementation into an API. (and yeah, I know that when one can monkeypatch something, monkeypatching seems easier than fixing something properly...)

Still: :+1:

PR #725, opened 2016-02-29
Make xrdb_get_value() behave as documented

psychon:
The documentation states that this function returns nil if the property does not
exist, but it actually returned nothing. And yes, in Lua this is a difference...

@Elv13 Could you test if this fixes your issue?

P.S.: I created this change entirely through the GitHub web interface. Let's see if this actually works...


psychon:
I updated my awesome and ran into the same problem as @Elv13, hence I merged this.

Issue #718, opened 2016-02-29
wibox.layout.mirror is broken


Elv13:
It will need some love/tests, I may do it at some point. It try to calls functions that doesn't exist (anymore?) like `wibox.widget.base.draw_widget`.

PR #727, opened 2016-02-29
Fix the mirror layout

psychon:
Issues involve:
- :layout() had the wrong signature and expected a cr argument that was left
  from when this was still the :draw() function.
- horizontal and vertical reflection were mixed up (I guess it has always been
  this way?)
- The return value should be a table of widget placements. Instead it was just a
  single widget placement.

This is broken since commit 85ab3f045b.

Fixes: https://github.com/awesomeWM/awesome/issues/718

### DIFF
--- a/lib/wibox/layout/mirror.lua
+++ b/lib/wibox/layout/mirror.lua
@@ -16,28 +16,24 @@ local matrix = require("gears.matrix")
 local mirror = { mt = {} }
 
 --- Layout this layout
-function mirror:layout(context, cr, width, height)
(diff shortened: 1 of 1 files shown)


Elv13:
It fix the code. There is no longer errors. However it doesn't really work in the quick test I did. The child widget children don't have the mirror effect. I don't mind much, as it is what I want in my case, but I guess it is not exactly working as expected.

I propose merging this, just remove the `Fixes: #718` part. Non-broken code with missing edge-case is better than broken code.

PR #730, opened 2016-03-02
fix(lib: menubar): take in account common list padding

actionless:


lib/menubar/init.lua
```diff
@@ -61,7 +61,7 @@ menubar.geometry = { width = nil,
                      y = nil }
 
 --- Width of blank space left in the right side.
-menubar.right_margin = 50
+menubar.right_margin = theme.xresources.apply_dpi(8)
```
psychon: Why change the default for this (assuming 96) DPI from 50 to 8? That seems random. I'd like to see at least some reason in the commit message.

lib/menubar/init.lua
```diff
@@ -69,6 +69,8 @@ menubar.right_label = "▶▶"
 --- Label used for "Previous page", default "◀◀".
 menubar.left_label = "◀◀"
 
+local list_interspace = theme.xresources.apply_dpi(4) * 3 -- common.list_update padding
```
psychon: Why not `apply_dpi(12)`? I guess this is about the padding that `common.list_update` adds around the textbox and imagebox, right? The comment should explain this better.

lib/menubar/init.lua
actionless: it was not dpi-aware back then and also i set it a little bit larger to avoid edge cases with pagination (which i can't reproduce after adding `list_interspace`)

lib/menubar/init.lua
actionless: because each of them is 4 px, ie for readability, i can change it to dpi(12)

psychon: Can you add a longer comment here instead? Something that explains that `common.list_update` adds three times a margin of `dpi(4)`?

actionless:
updated comment

blueyed:
Thanks!

PR #731, opened 2016-03-02
RFC awful.placement: Use math.ceil()

Elv13:
(lua 5.1)

I get a lot of "expected integer, got number" errors when using `awful.placement` (mostly with wibox). In 3.5.x, I don't have this problem.

I don't really like this PR, but umm, it work... I would prefer the C side to do it, but I open this PR instead of a bug because umm... I don't know

### DIFF
--- a/objects/drawin.c
+++ b/objects/drawin.c
@@ -396,10 +396,10 @@ luaA_drawin_geometry(lua_State *L)
         area_t wingeom;
 
         luaA_checktable(L, 2);
-        wingeom.x = luaA_getopt_integer(L, 2, "x", drawin->geometry.x);
-        wingeom.y = luaA_getopt_integer(L, 2, "y", drawin->geometry.y);
-        wingeom.width = luaA_getopt_integer(L, 2, "width", drawin->geometry.width);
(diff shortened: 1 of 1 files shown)


psychon:
This should be fixed on the C side, I think. Rounding things everywhere in Lua is ugly. However, I don't really understand why this expects a number. `luaA_client_geometry` uses ` luaA_getopt_number`, not `luaA_getopt_integer`...

Elv13:
But wiboxes don't

``
`../objects/drawin.c:        wingeom.x = luaA_getopt_integer(L, 2, "x", drawin->geometry.x);
../objects/drawin.c:        wingeom.y = luaA_getopt_integer(L, 2, "y", drawin->geometry.y);
../objects/drawin.c:        wingeom.width = luaA_getopt_integer(L, 2, "width", drawin->geometry.width);
../objects/drawin.c:        wingeom.height = luaA_getopt_integer(L, 2, "height", drawin->geometry.height);
```

PR update coming

Elv13:
Updated

psychon:
:+1: (and yeah, I do vaguely remember that it was me who made this "integer" on purpose and fixed some of the resulting issues on the Lua size, but this path leads to madness, as the original commit in this PR shows)

PR #737, opened 2016-03-04
wibox.widget: Minor enhancements

Elv13:
* Better widget names when using the declarative syntax
 * Add ratio.get_ratio to avoid using the private API
 * Also support `set_widget` when swapping widgets

### DIFF
--- a/lib/wibox/layout/fixed.lua
+++ b/lib/wibox/layout/fixed.lua
@@ -164,9 +164,17 @@ function fixed:swap_widgets(widget1, widget2, recursive)
     local idx1, l1 = self:index(widget1, recursive)
     local idx2, l2 = self:index(widget2, recursive)
 
-    if idx1 and l1 and idx2 and l2 then
-        l1:set(idx1, widget2)
-        l2:set(idx2, widget1)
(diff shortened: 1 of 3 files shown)


lib/wibox/layout/fixed.lua
```diff
@@ -164,9 +164,17 @@ function fixed:swap_widgets(widget1, widget2, recursive)
...
+        if l1.set then
+            l1:set(idx1, widget2)
+        elseif l1.set_widget then
+            l1:set_widget(widget2)
+        end
+        if l2.set then
+            l2:set(idx2, widget1)
+        elseif l2.set_widget then
+            l2:set_widget(widget1)
+        end
```
psychon: Meh. :-(

But ok, I let you merge the original code and now this has to be fine as well.

lib/wibox/layout/ratio.lua
```diff
@@ -215,6 +215,11 @@ function ratio:set_ratio(index, percent)
     self:emit_signal("widget::layout_changed")
 end
 
+function ratio:get_ratio(index)
```
psychon: The documentation is missing

psychon:
Adding to the "next" milestone assuming that the missing documentation is fixed before merging (which likely means that you'll then merge this yourself).

PR #738, opened 2016-03-04
gears.color: Add transparent constant

Elv13:
Avoid repetitive "#00000000" for transparent wiboxes and such
elements.


Elv13:
oops, wrong `git add`

psychon:
:+1:

Issue #672, opened 2016-02-06
Stop restarting on RandR configuration changes


psychon:
Same issue in the old tracker:https://awesome.naquadah.org/bugs/index.php?do=details&task_id=687
Some branch where I worked on something, but I'm not even sure on what: https://github.com/awesomeWM/awesome/tree/remove_screen_index

I think everyone agrees that restarting on RandR configuration changes is just a cheap hack since we are too lazy to handle this properly. This issue is about handling this properly. Before working on this, I want to get some input on what everyone thinks "properly" means.

For example, should monitor "coldplug" (monitor already present when awesome starts) and "hotplug" (configuration is changed on the fly) be handled differently?

### The current screen API

- Screens identified with integers from 1 to `screen.count()`
- Properties:
  - `geometry` The geometry on the root window
  - `workarea` The geometry of the screen minus things like e.g. the wibox. This is the space available for displaying clients
  - `index` The number identifying this screen
  - `outputs` Some information about the RandR outputs connected to this screen's CRTC (only if RANDR is in use)
- Signals: Only `property::workarea`

I explain this so detailed, because I think the first point is a problem: Assume you are in a situation with three screens and screen 2 is disconnected. Effectively this means that there are now screens 1 and 2, but 2 has a completely different meaning. This is why I propose to stop identifying screens with numbers and only use userdata instead.

### @psychon's idea after thinking for 0.5 seconds

- Lua can iterate over all screens via `for s in screen.iterate()` (I don't want `for i, s in screen.iterate()` or something like that since this would provide some kind of integer again)
- Properties stay the same (minus `index`)
- All the properties get signals that are emitted when needed
- Screens have signals like `connected`/`disconnected` or `enabled`/`disabled` (I'd try to stay close the RANDR's terminology, but I don't know what that is right now)

So far for simple. I'd guess that I wouldn't many complaints about this. However, now comes my evil idea:

- While parsing the config, we have no screens at all yet.
- Afterwards (and before managing any clients), the screens are introduced in the same way as monitor hotplug is handled

Pro (at least I hope so): No special magic is needed so that Lua code handles monitor hotplug correctly, because all Lua code will have to do this anyway.
Con: Quite an API break (but I guess removing screen indices already does quite a lot of damage)

### Your idea here

So, what is your opinion? Without thinking about any implementation constraints, what kind of API would you prefer?

kapsh:
I like second idea without index. But I have only one question: how it should be possible to distinguish screens? 
[...]

PR #746, opened 2016-03-06
Indexless screen iteration

psychon:
This adds a new way to iterate over screens without using a screen index:
```lua
for s in screen do
    print("I'm seeing screen " .. tostring(s))
end
```
Right now this is equivalent to `for s=1, screen.count() do`, but it can easily be changed to iterate over screen objects on the C side.

Note that this conflicts with https://github.com/awesomeWM/awesome/pull/745 (It does some changes that is also done in there). I'm only submitting this for review right now and will rebase this on master once that other PR is merged.
Edit: Done!

(I'm especially curious about complaints on the syntax `for s in screen do` and suggestions what to do instead.)

Edit: This is one more step forward for https://github.com/awesomeWM/awesome/issues/672 (and I mention this here so that the issue has a nice list of all the related pull requests embedded)

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -91,7 +91,7 @@ end
 
 -- {{{ Wallpaper
 if beautiful.wallpaper then
-    for s = 1, screen.count() do
+    for s in screen do
         gears.wallpaper.maximized(beautiful.wallpaper, s, true)
     end
 end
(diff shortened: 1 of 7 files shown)


blueyed:
I like it!

Could this be made available for all / other objects as well, e.g. clients - in a generic way?

psychon:
What do you mean with "e.g."? I think clients are the only other kind of objects for which something like this would make sense (who wants to iterate over all drawins?). For clients, we have `client.get()` right now and that gets an optional screen argument. So that doesn't completely fit into this scheme either. Also, "user" code should use something like `awful.client.tiled` instead of iterating over `client.get` directly.

All together: I don't see where this can be sensibly added for clients.
(Oh and it wouldn't really be generic either, I guess)

Random thing: I think Lua 5.0 had the `for k,v in table do`-syntax behave the same way as we have now with `pairs`...

blueyed:
Yes, clients is the only thing where it really makes sense.
Ok with me then.
Thanks for explaining!

Issue #739, opened 2016-03-04
Broken link in README.md


psychon:
![contributing-link](https://cloud.githubusercontent.com/assets/89482/13527034/65e66820-e20c-11e5-9aa2-5c254b9fc440.png)

psychon:
Thanks!

PR #750, opened 2016-03-08
Fix broken link in README.md #739

liquiddandruff:


blueyed:
Thanks, but it does not work for the HTML files built by `make ldoc` itself.

blueyed:
See https://github.com/awesomeWM/awesome/pull/752 - while it's a workaround it's also good to have in general (apart from offline usage?!).

liquiddandruff:
Good call

PR #751, opened 2016-03-08
tasklist: displayed minimized client label in italic font

blueyed:
This could/should be made configurable probably, but it's a good start
nonetheless?!


psychon:
This gives `beautiful.tasklist_plain_task_name` two meanings. Why not just use another key for this, e.g. `tasklist_no_italic_font_for_minimized` (but perhaps you can come up with a better name)?

blueyed:
It's possible already:

```
theme.font = "Ubuntu Sans 8"
…
theme.tasklist_font_minimized = "Ubuntu Sans Italic 8"
```

PR #753, opened 2016-03-10
Add shapes

Elv13:
This pull request add more `gears.shape`s and add dynamically generated examples

This slow down the compilation by 2-3 seconds

### DIFF
--- a/docs/generate_examples.cmake
+++ b/docs/generate_examples.cmake
@@ -1,0 +1,47 @@
+
+# Get and update the LUA_PATH so the scripts can be executed without Awesome
+execute_process(COMMAND lua -e print\(package.path\) OUTPUT_VARIABLE LUA_PATH_)
+set(ENV{LUA_PATH} "${SOURCE_DIR}/lib/?;${SOURCE_DIR}/lib/?.lua;${LUA_PATH_}")
+
+file(MAKE_DIRECTORY "${BUILD_DIR}/doc/images")
+
+#
+# Shape API
+#
+foreach (SHAPE_NAME "circle" "arrow" "rounded_rect" "hexagon" "infobubble"
+    "powerline" "isosceles_triangle" "cross" "octogon" "parallelogram"
+    "losange" "partially_rounded_rect" "radial_progress" "rounded_bar"
+    "rectangle" "rectangular_tag")
+    set(SHAPE_FILE "${SOURCE_DIR}/tests/shape/${SHAPE_NAME}.lua")
+    set(SHAPE_SVG  "${BUILD_DIR}/doc/images/shape_${SHAPE_NAME}.svg")
+
+    # Generate some SVG for the documentation and load the examples for the doc
+    execute_process(
+        COMMAND lua ${SOURCE_DIR}/tests/shape/test-shape.lua ${SHAPE_FILE} ${SHAPE_SVG}
+        OUTPUT_VARIABLE SHAPE_OUTPUT
+        ERROR_VARIABLE  SHAPE_ERROR
+    )
+
+    if (NOT SHAPE_ERROR STREQUAL "")
+        message(${SHAPE_ERROR})
+        message(FATAL_ERROR ${SHAPE_NAME} " SVG generation failed, bye")
+    endif()
+
+    # Set the SVG paths for the doc
+    set(SHAPE_${SHAPE_NAME}_SVG "../images/shape_${SHAPE_NAME}.svg")
+
+    # Use the .lua as code example
+    file(READ ${SHAPE_FILE} SHAPE_EXAMPLE)
+    STRING(REGEX REPLACE "\n" ";" SHAPE_EXAMPLE "${SHAPE_EXAMPLE}")
+    SET(SHAPE_COMMENTED
+        "![Shape example](../images/shape_${SHAPE_NAME}.svg)\n--\n-- @usage"
+    )
+    foreach (EXAMPLE_FILE ${SHAPE_EXAMPLE})
+        if(NOT EXAMPLE_FILE MATCHES "^.+--DOC_HIDE$")
+            SET(SHAPE_COMMENTED ${SHAPE_COMMENTED}\n--${EXAMPLE_FILE})
+        endif()
+    endforeach()
+
+    set(SHAPE_${SHAPE_NAME}_EXAMPLE ${SHAPE_COMMENTED})
+
+endforeach()
--- a/lib/gears/shape.lua
+++ b/lib/gears/shape.lua
@@ -326,6 +326,76 @@ function module.losange(cr, width, height)
     cr:close_path()
 end
 
+--- A partial rounded bar. How much of the rounded bar is visible depends on
+-- the given percentage value.
+--
+-- Note that this shape is not closed and thus filling it doesn't make much
+-- sense.
+--
+-- @SHAPE_radial_progress_EXAMPLE@
+--
+-- @param cr A cairo context
+-- @tparam number w The shape width
+-- @tparam number h The shape height
+-- @tparam number percent The progressbar percent
+-- @tparam boolean hide_left Do not draw the left side of the shape
+function module.radial_progress(cr, w, h, percent, hide_left)
+    percent = percent or 1
+    local total_length = (2*(w-h))+2*((h/2)*math.pi)
+    local bar_percent  = (w-h)/total_length
+    local arc_percent  = ((h/2)*math.pi)/total_length
+    --TODO ajust the percent (with - arc_percent) is hide_left
+
+    -- Bottom line
+    if percent > bar_percent then
+        cr:move_to(h/2,h)
+        cr:line_to((h/2) + (w-h),h)
+        cr:stroke()
+    elseif percent < bar_percent then
+        cr:move_to(h/2,h)
+        cr:line_to(h/2+(total_length*percent),h)
+        cr:stroke()
+    end
+
+    -- Right arc
+    if percent >= bar_percent+arc_percent then
+        cr:arc(w-h/2 , h/2, h/2,3*(math.pi/2),math.pi/2)
+        cr:stroke()
+    elseif percent > bar_percent and percent < bar_percent+(arc_percent/2) then
+        cr:arc(w-h/2 , h/2, h/2,(math.pi/2)-((math.pi/2)*((percent-bar_percent)/(arc_percent/2))),math.pi/2)
+        cr:stroke()
+    elseif percent >= bar_percent+arc_percent/2 and percent < bar_percent+arc_percent then
+        cr:arc(w-h/2 , h/2, h/2,0,math.pi/2)
+        cr:stroke()
+        local add = (math.pi/2)*((percent-bar_percent-arc_percent/2)/(arc_percent/2))
+        cr:arc(w-h/2 , h/2, h/2,2*math.pi-add,0)
+        cr:stroke()
+    end
+
+    -- Top line
+    if percent > 2*bar_percent+arc_percent then
+        cr:move_to((h/2) + (w-h),0)
+        cr:line_to(h/2,0)
+        cr:stroke()
+    elseif percent > bar_percent+arc_percent and percent < 2*bar_percent+arc_percent then
+        cr:move_to((h/2) + (w-h),0)
+        cr:line_to(((h/2) + (w-h))-total_length*(percent-bar_percent-arc_percent),0)
+        cr:stroke()
+    end
+
+    -- Left arc
+    if not hide_left then
+        if percent > 0.985 then
+            cr:arc(h/2, h/2, h/2,math.pi/2,3*(math.pi/2))
+            cr:stroke()
+        elseif percent  > 2*bar_percent+arc_percent then
+            local relpercent = (percent - 2*bar_percent - arc_percent)/arc_percent
+            cr:arc(h/2, h/2, h/2,3*(math.pi/2)-(math.pi)*relpercent,3*(math.pi/2))
+            cr:stroke()
+        end
+    end
+end
+
 --- Ajust the shape using a transformation object
 --
 -- Apply various transformations to the shape
--- a/lib/gears/shape.lua
+++ b/lib/gears/shape.lua
@@ -187,6 +187,30 @@ function module.isosceles_triangle(cr, width, height)
     cr:close_path()
 end
 
+--- A cross (**+**) symbol
+-- @param cr A cairo context
+-- @tparam number width The shape width
+-- @tparam number height The shape height
+-- @tparam[opt=width/3] number thickness The cross section thickness
+function module.cross(cr, width, height, thickness)
+    thickness = thickness or width/3
+    local xpadding   = (width  - thickness) / 2
+    local ypadding   = (height - thickness) / 2
+    cr:move_to(xpadding, 0)
+    cr:line_to(width - xpadding, 0)
+    cr:line_to(width - xpadding, ypadding)
+    cr:line_to(width           , ypadding)
+    cr:line_to(width           , height-ypadding)
+    cr:line_to(width - xpadding, height-ypadding)
+    cr:line_to(width - xpadding, height         )
+    cr:line_to(xpadding        , height         )
+    cr:line_to(xpadding        , height-ypadding)
+    cr:line_to(0               , height-ypadding)
+    cr:line_to(0               , ypadding       )
+    cr:line_to(xpadding        , ypadding       )
+    cr:close_path()
+end
+
 --- Ajust the shape using a transformation object
 --
 -- Apply various transformations to the shape
--- a/lib/gears/shape.lua
+++ b/lib/gears/shape.lua
@@ -211,6 +211,26 @@ function module.cross(cr, width, height, thickness)
     cr:close_path()
 end
 
+--- A similar shape to the `rounded_rect`, but with sharp corners
+-- @param cr A cairo context
+-- @tparam number width The shape width
+-- @tparam number height The shape height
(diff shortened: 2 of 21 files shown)


Elv13:
So, this PR, beside more shapes, add an early way to add tested code examples to the documentation. Those code snippets are ran at compile time and must run clean (or kill the build). For now, only test that can be ran without Awesome are used. I am not sure having compile time Xephyr is a good idea. One possibility is to add a dummy drawable/drawin on an SVG surface for code that doesn't require "live" CAPI usage. But this will be later.

This is an early version. I first need to make it pass Travis (and generate the `.svg`)

lib/gears/shape.lua
```diff
@@ -87,21 +97,41 @@ function module.infobubble(cr, width, height, corner_radius, arrow_size, arrow_p
...
+        cr:line_to(width        , 0        )
+        cr:line_to(width        , height   )
+        cr:line_to(arrow_length , height   )
+    else
+        cr:move_to(0            , 0        )
+        cr:line_to(-arrow_length, height/2 )
+        cr:line_to(0            , height   )
+        cr:line_to(width        , height   )
+        cr:line_to(width        , 0        )
+    end
```
psychon: Do I understand it correctly that a negative `arrow_length` makes the arrow point to the other direction? If so, should that be mentioned in the docs?

lib/gears/shape.lua
```diff
@@ -195,3 +484,4 @@ function module.transform(shape)
 end
 
 return module
+-- kate: space-indent on; indent-width 4; replace-tabs on;
```
psychon: Huh? This doesn't have a vim modeline (weird...) and we now have kate mode lines?

I guess everyone should somehow use the `.editconfig` file that we have in awesome, I just don't know how. :-(
(Adding special comments for every editor isn't going to fly.)

lib/gears/shape.lua
```diff
@@ -168,6 +220,243 @@ function module.isosceles_triangle(cr, width, height)
...
 end
 
+--- A cross (**+**) symbol.
+--
+-- ![Shape example](@SHAPE_cross_SVG@)
+--
+-- @SHAPE_cross_EXAMPLE@
+--
+-- @param cr A cairo context
+-- @tparam number width The shape with
```
psychon: `with` -> `width` (also in other places in this file)

lib/gears/shape.lua
Elv13: I think the last commit in this PR address this quite well

lib/gears/shape.lua
```diff
@@ -168,6 +220,243 @@ function module.isosceles_triangle(cr, width, height)
...
+--
+-- @SHAPE_octogon_EXAMPLE@
+--
+-- @param cr A cairo context
+-- @tparam number width The shape with
+-- @tparam number height The shape height
+-- @tparam number corner_radius 
+function module.octogon(cr, width, height, corner_radius)
+    local corner_radius = corner_radius or math.min(10, math.min(width, height)/4)
+    local offset = math.sqrt( (corner_radius*corner_radius) / 2 )
```
psychon: Isn't this just `offset = corner_radius / math.sqrt(2)` (assuming `corner_radius` is positive)?
That doesn't really help me understand this (why not just `corner_radius` directly?), but at least feels easier to read.

Oh and: Why `math.min(10)` in the line before this? To me it looks like just `math.min(width,height/4)` should do.

lib/gears/shape.lua
```diff
@@ -168,6 +220,243 @@ function module.isosceles_triangle(cr, width, height)
...
+--
+-- ![Shape example](@SHAPE_circle_SVG@)
+--
+-- @SHAPE_circle_EXAMPLE@
+--
+-- @param cr A cairo context
+-- @tparam number width The shape with
+-- @tparam number height The shape height
+function module.circle(cr, width, height)
+    local size = width < height and (width/2) or (height/2)
```
psychon: Please use `math.min(width, height) / 2`, I think that's easier to read.

lib/gears/shape.lua
```diff
@@ -168,6 +220,243 @@ function module.isosceles_triangle(cr, width, height)
...
+    local w_offset, h_offset = (width - size*2)/2, (height - size*2)/2
+    cr:arc(w_offset + size, h_offset + size, size, 0, 2*math.pi)
```
psychon: Uhm... so the x-coordinate of the center of the circle is `w_offset + size = (width - size*2)/2 + size = width/2` and the y-coordinate is just `height/2` as well. Wouldn't it be easer to use `cr:arc(width / 2, height / 2, size, 0, 2*math.pi)`?

lib/gears/shape.lua
```diff
@@ -168,6 +220,243 @@ function module.isosceles_triangle(cr, width, height)
...
+--
+-- @param cr A cairo context
+-- @tparam number width The shape with
+-- @tparam number height The shape height
+function module.rectangle(cr, width, height)
+    cr:rectangle(0, 0, width, height)
+end
+
+--- A diagonal parallelogram with the bottom right corner at x=0 and top left
+-- at x=width.
```
psychon: Isn't (0, 0) the top-left corner? I think this should say that the top-left corner is at x=0 and the bottom right one is at x=width. Also, you could mention that the top right corner is at (width-thickness, 0), that might simplify understanding this. (Is it usual to call this a thickness? I have no idea...)

lib/gears/shape.lua
```diff
@@ -168,6 +220,243 @@ function module.isosceles_triangle(cr, width, height)
...
+--
+-- @param cr A cairo context
+-- @tparam number width The shape with
+-- @tparam number height The shape height
+-- @tparam boolean tl If the top left corner is rounded
+-- @tparam boolean tr If the top right corner is rounded
+-- @tparam boolean br If the bottom right corner is rounded
+-- @tparam boolean bl If the bottom left corner is rounded
+-- @tparam number rad The corner radius
+function module.partially_rounded_rect(cr, width, height, tl, tr, br, bl, rad)
```
psychon: Should `rounded_rect` be replaced with a call to this, just so that the code exists only once? (Dunno if that makes much of a difference)
Also, please move the two closer together.

tests/test-shape.lua
```diff
@@ -0,0 +1,54 @@
...
+
+local function show(cr, skip_fill)
+    cr:set_source_rgba(0.380392156863,0.505882352941,1,1)
+
+    if skip_fill then
+        cr:stroke()
+    else
+        cr:stroke_preserve()
+        cr:set_source_rgba(0.380392156863,0.505882352941,1,0.5)
+        cr:fill()
```
psychon: Doesn't this fill partly draw above the stroke, which should be visible thanks to transparency? I'd do it the other way around, first fill, then stroke (which doesn't have any transparency) to avoid this. Or is that on purpose?

tests/test-shape.lua
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    cr:translate(96)
```
psychon: Please use `cr:translate(96, 0)`.

psychon:
Nice idea. I don't mind the new shapes (I wonder how many shapes you still can come up with...) and having (1) tested usage examples and (2) images of these shapes in the docs is both nice.
I don't have any suggestions on how to integrate this more nicely in the build process (having to run tests just so that the installed documentation is complete is weird...), but it's a really nice idea. Thanks!

Elv13:
```sh
CMake Error at CMakeLists.txt:271 (file):

  file COPY cannot set permissions on

  "/home/travis/build/awesomeWM/awesome/.build-testing-gce-c4a4845b-a5c0-4ca8-a50d-9bd047905b5c-x86_64-linux-gnu-4.8/doc/images/shape_arrow.svg"
```

Any idea on that error? Seem like the folder is created with a different user. I looked at `travis.yml`, but beside a couple of `sudo`, it doesn't seem to do anything special.

Btw, should we remove those `sudo` and use a local prefix and add it to `$PATH`, `$LUA_PATH` and `$XDG_*`? This would avoid this kind of issues.

Elv13:
Umm, eventually I will have to figure out how to get these into the coverage _and_ running without the test framework...

@psychon So, is this good enough to merge now? I think the unaddressed comment are easy to understand once you see the images.

lib/gears/shape.lua
```diff
@@ -125,11 +204,14 @@ function module.arrow(cr, width, height, head_width, shaft_width, shaft_length)
...
+--- A squeezed hexagon filling the rectangle.
+--
+-- @SHAPE_hexagon_EXAMPLE@
+--
 -- @param cr A cairo context
--- @tparam number width The shape with
+-- @tparam number width The shape width
 -- @tparam number height The shape height
-function module.hexagon(cr, width, height)
+function module.hexagon(cr, width, height) --TODO need to new args to control top width
```
psychon: This (part of the first commit) adds a TODO...?

lib/gears/shape.lua
psychon: I think this comment wasn't addressed yet, was it?

lib/gears/shape.back
```diff
@@ -0,0 +1,449 @@
+---------------------------------------------------------------------------
```
psychon: Uhm, I think something went wrong with this commit ("add radial_progress shape"). It adds way more stuff than what the commit messages makes me expect.

Should this PR be split into two parts, one addings lots of new shapes (basically the first few commits) and another one handles the new "magic doc-images"?

(GitHub's interface changed, I can't comment at the bottom of this commit... Let's see where this comment ends up)

tests/shape/test-shape.lua
```diff
@@ -0,0 +1,36 @@
+-- Test if shape crash when called
+-- Also generate some SVG to be used by the documentation
+-- it also "prove" that the code examples are all working
+local cairo = require( "lgi"         ).cairo
+local shape = require( "gears.shape" )
+local filepath, svgpath = ...
+
+-- This is normal when running test, it will run twice
+if (not filepath) or (not svgpath) or key then
+    os.exit(0)
```
psychon: At least print a warning that it doesn't do anything? Or, even better, don't run it twice?

docs/generate_examples.cmake
```diff
@@ -0,0 +1,48 @@
+
+# Get and update the LUA_PATH so the scripts can be executed without Awesome
+execute_process(COMMAND lua -e print\(package.path\) OUTPUT_VARIABLE LUA_PATH_)
+set(ENV{LUA_PATH} "${SOURCE_DIR}/lib/?;${SOURCE_DIR}/lib/?.lua;${LUA_PATH_}")
+
+execute_process(COMMAND mkdir -p "${BUILD_DIR}/doc/images")
```
psychon: Can `file(MAKE_DIRECTORY "${BUILD_DIR}/doc/images")` handle this instead (e.g. so that no new process is forked)?

docs/generate_examples.cmake
```diff
@@ -0,0 +1,48 @@
...
+    execute_process(
+        COMMAND lua ${SOURCE_DIR}/tests/shape/test-shape.lua ${SHAPE_FILE} ${SHAPE_SVG}
+        OUTPUT_VARIABLE SHAPE_OUTPUT
+        ERROR_VARIABLE  SHAPE_ERROR
+    )
+
+    if (NOT SHAPE_ERROR STREQUAL "")
+        message(${SHAPE_ERROR})
+        message(${SHAPE_NAME} " SVG generation failed, bye")
+        exit(1)
```
psychon: Does this work? I can't find `exit` in the CMake docs. Shouldn't this be `message(FATAL_ERROR the rest of the message here)`?
(And for the previous line: `message(SEND_ERROR ${SHAPE_ERROR})`)

psychon:
Also, could you push this as a branch to the awesomeWM/awesome repo? That way, awesome-robot should provide a link to the finished docs to (hopefully) look at.

lib/gears/shape.back
Elv13: Oops, git rebase fail... (note to myself, never use -A in aliases again)

docs/generate_examples.cmake
Elv13: Umm, good catch, . It was working, yes, but not the way it was intended to (Aka, it did call `message(FATAL_ERROR)` "indirectly".

tests/shape/test-shape.lua
Elv13: I will have to figure out how to write tests that run on both the test run and the pre-processing run. One give coverage and the other the documentation. This will be improved in future commits.

lib/gears/shape.lua
Elv13: I don't think it is worth merging, its simple code, it's a pointless micro optimization, but the rounded_rect one is called often when resizing clients with awful.shape, lets avoid a few pointless "if". It's 4 lines

tests/shape/test-shape.lua
psychon: Then I would suggest: Don't run this in the tests at all. If something breaks, the build should still fail. The only missing piece would be code coverage. If you want, feel free to open an issue about that and assign it to me (it was me who created this code-coverage stuff, after all).

docs/generate_examples.cmake
```diff
@@ -0,0 +1,47 @@
...
+    "losange" "partially_rounded_rect" "radial_progress" "rounded_bar"
+    "rectangle" "rectangular_tag")
+    set(SHAPE_FILE "${SOURCE_DIR}/tests/shape/${SHAPE_NAME}.lua")
+    set(SHAPE_SVG  "${BUILD_DIR}/doc/images/shape_${SHAPE_NAME}.svg")
+
+    # Generate some SVG for the documentation and load the examples for the doc
+    execute_process(
+        COMMAND lua ${SOURCE_DIR}/tests/shape/test-shape.lua ${SHAPE_FILE} ${SHAPE_SVG}
+        OUTPUT_VARIABLE SHAPE_OUTPUT
+        ERROR_VARIABLE  SHAPE_ERROR
```
psychon: `execute_process` has `RESULT_VARIABLE`, according to the docs. It sounds like this will (sometimes?!?) be set to the exit code of the process. That should catch Lua errors more reliable (`lua -e 'error()'` has a non-zero exit code).

lib/gears/shape.lua
psychon: Ok.

docs/generate_examples.cmake
Elv13: Currently. if there is anything on stderr, it exit. I think it is what we want, even if there would be a "0" exit code. No?

psychon: Yup. I was thinking that there could be errors that are only indicated by the exit code, but... you are right, that doesn't make any sense.

psychon:
Hm. Apparently there is no easy way to look at the resulting web pages except for cloning the git and checking out the branch. Anyway, here is the result from that travis build:

https://github.com/awesomeWM/apidoc/commits/add_shapes-491f17f

Compare view: https://github.com/awesomeWM/apidoc/compare/45a8a04...e0f16a9

Comparison against master (gh-pages): https://github.com/awesomeWM/apidoc/compare/gh-pages...e0f16a9

Elv13:
I git cloned the page, it display fine locally

tests/shape/test-shape.lua
Elv13: Ok, I pushed one last version for today, I am exhausted and I will make even more silly mistakes if I keep rebasing this stuff. The test (should) no longer run, I removed this code.

psychon:
Yup, looks great here as well. Nice work!
The docs for the parallelogram still say "A diagonal parallelogram with the bottom right corner at x=0 and top left at x=width.", but it's the bottom left corner which is at x=0 and the top right one which is at x=width.
Also, "A stroke only rounded bar with only percent set. This path is not closed." is hard to understand. How about "A partial rounded bar. How much of the rounded bar is visible depends on the given percentage value. Note that this shape is not closed and thus filling it doesn't make much sense." (by the way, what's the point of this shape? Perhaps some usage example should also be mentioned in the docs for it.)
I think that all the usage examples begin with an empty line. I guess that should be removed (either by editing the source or via some more CMake magic...).

tests/shape/rectangular_tag.lua
```diff
@@ -0,0 +1,13 @@
+local shape,cr,show = ... --DOC_HIDE
+ 
+shape.rectangular_tag(cr, 70, 70)
+show(cr) --DOC_HIDE
+ 
+-- shape.rectangular_tag(cr,20,70) --TODO broken
```
psychon: This one shows up in the generated API docs (and sounds like something that should be fixed).

Elv13:
Ok, I will do that tomorrow. Can I merge it after fixing that? I have another (actually, 5) PR that re-use this framework coming for other modules, so I this has to be merged before I submit it. The coverage issue can be addressed later (it would be nice to have it, if not every new commits I will make will regress the coverage % as I plan to always include tested code examples where relevant.

tests/shape/rectangular_tag.lua
Elv13: I will add a DOC_HIDE for now, it has a minor visual issue, nothing terrible. I will fix it later when I improve the svg generator to show more than 3 images

psychon:
The decreased code coverage isn't an issue for me. The code is actually covered, but the tool just misses that. That doesn't worry me much.
I'd still like to take another look at this tomorrow. Sorry.
(P.S.: I'll be AFK next week, so I no reviews from me then.)

psychon:
After looking at this again: I should be less of a control freak and you are doing fine work. Please take a careful look at the final commits tomorrow (mostly to spot other rebase errors). Afterwards, feel free to merge. Thanks!

PR #754, opened 2016-03-11
awful.tag: Add "volatile" property

Elv13:
Useful when using dynamic tags. The tags will be closed once
it is empty. This was part of Tyrannical for many years, but is
generally useful for other workflows too.

```lua
   local t = awful.tag.add("my_tag",{volatile=true, screen=2})
   awful.spawn("myapp", {tag=t})
```


psychon:
Fine by me

PR #757, opened 2016-03-13
C-API objects properties

Elv13:
Here is a first implementation of the property system introduced by #482

##### Testing done:

 * It does seem to work

##### TODO:

* Update the unit test to use my implementation
* Drop some deadcode like in `awful.rules`
* Create genuine classes instead of the fallback to `awful.{tag,client}` as discussed before.
* Abuse of the cmake macros to force the documentation into the "tag" and "client" CAPI classes

Please comment on the design before I address the TODO

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -10,6 +10,7 @@
 -- Grab environment we need
 local util = require("awful.util")
 local spawn = require("awful.spawn")
+local object = require("gears.object")
 local tag = require("awful.tag")
 local pairs = pairs
 local type = type
@@ -1102,6 +1103,7 @@ capi.client.add_signal("marked")
 capi.client.add_signal("unmarked")
 
 capi.client.connect_signal("focus", client.focus.history.add)
+
 -- Add clients during startup to focus history.
 -- This used to happen through ewmh.activate, but that only handles visible
 -- clients now.
@@ -1120,6 +1122,14 @@ capi.client.connect_signal("unmanage", client.floating.delete)
 
 -- Register persistent properties
 client.property.persist("floating", "boolean")
+
+-- Extend the luaobject
+object.properties(capi.client, {
+    getter_class    = client,
+    setter_class    = client,
+    getter_fallback = client.property.get,
+    setter_fallback = client.property.set,
+})
 
 return client
 
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -11,6 +11,7 @@
 local util = require("awful.util")
 local ascreen = require("awful.screen")
 local beautiful = require("beautiful")
+local object = require("gears.object")
 local pairs = pairs
 local ipairs = ipairs
 local table = table
@@ -872,6 +873,16 @@ capi.screen.connect_signal("tag::history::update", tag.history.update)
 function tag.mt:__call(...)
     return tag.new(...)
 end
+
+-- Extend the luaobject
+-- `awful.tag.setproperty` currently handle calling the setter method itself
+-- while `awful.tag.getproperty`.
+object.properties(capi.tag, {
+    getter_class    = tag,
+    setter_class    = tag,
+    getter_fallback = tag.getproperty,
+    setter          = tag.setproperty,
+})
 
 return setmetatable(tag, tag.mt)
 
--- a/lib/gears/object.lua
+++ b/lib/gears/object.lua
@@ -9,8 +9,9 @@ local setmetatable = setmetatable
 local pairs = pairs
 local type = type
 local error = error
+local properties = require("gears.object.properties")
 
-local object = { mt = {} }
+local object = { properties = properties, mt = {} }
 
 --- Verify that obj is indeed a valid object as returned by new()
 local function check(obj)
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -15,6 +15,7 @@ local capi =
     client = client
 }
 local util = require("awful.util")
+local object = require("gears.object")
 
 local function get_screen(s)
     return s and capi.screen[s]
@@ -244,6 +245,9 @@ function screen.get_bounding_geometry(s, args)
 end
 
 capi.screen.add_signal("padding")
+
+-- Extend the luaobject
+object.properties(capi.screen, {auto_emit=true})
 
 return screen
 
--- a/lib/gears/object/properties.lua
+++ b/lib/gears/object/properties.lua
@@ -1,0 +1,103 @@
+---------------------------------------------------------------------------
+--- An helper module to map userdata __index and __newindex entries to
+-- lua classes.
+--
+-- @author Emmanuel Lepage-Vallee &lt;elv1313@gmail.com&gt;
+-- @copyright 2016 Emmanuel Lepage-Vallee
+-- @release @AWESOME_VERSION@
+-- @module gears.object.properties
+---------------------------------------------------------------------------
+
+local object = {}
+
+local properties = setmetatable({}, { __mode = 'k' })
+
+local function cobj_register(cobj)
+    local fallback = {}
+
+    function fallback:rawset(_, prop, val)
+        fallback[prop] = val
+    end
+
+    function fallback:rawget(_, prop)
+        return fallback[prop]
+    end
+
+    properties[cobj] = fallback
+    return fallback
+end
+
+--- Add the missing properties handler to a CAPI object such as client/tag/screen
+-- Valid args:
+--
+-- * **getter**: A smart getter (handle property getter itself)
+-- * **getter_fallback**: A dumb getter method (don't handle individual property getter)
+-- * **getter_class**: A module with individual property getter/setter
+-- * **getter_prefix**: A special getter prefix (like "get" or "get_" (default))
+-- * **setter**: A smart setter (handle property setter itself)
+-- * **setter_fallback**: A dumb setter method (don't handle individual property setter)
+-- * **setter_class**: A module with individual property getter/setter
+-- * **setter_prefix**: A special setter prefix (like "set" or "set_" (default))
+-- * **auto_emit**: Emit "property::___" automatically (default: false). This is
+--     ignored when setter_fallback is set or a setter is found
+--
+-- @param class A standard luaobject derived object
+-- @tparam[opt={}] table args A set of accessors configuration parameters
+function object.capi_index_fallback(class, args)
+    args = args or {}
+
+    local getter_prefix = args.getter_prefix or "get_"
+    local setter_prefix = args.setter_prefix or "set_"
+
+    local getter = args.getter or function(cobj, prop)
+        -- Look for a getter method
+        if args.getter_class and args.getter_class[getter_prefix..prop] then
+            return args.getter_class[getter_prefix..prop](cobj)
+        end
+
+        -- Make sure something like c:a_mutator() works
+        if args.getter_class and args.getter_class[prop] then
+            return args.getter_class[prop]
+        end
+
+        -- In case there is already a "dumb" getter like `awful.tag.getproperty'
+        if args.getter_fallback then
+            return args.getter_fallback(cobj, prop)
+        end
+
+        local fallback = properties[cobj] or cobj_register(cobj)
+
+        -- Use the fallback property table
+        return fallback[prop]
+    end
+
+    local setter = args.setter or function(cobj, prop, value)
+        -- Look for a setter method
+        if args.setter_class and args.setter_class[setter_prefix..prop] then
+            return args.setter_class[setter_prefix..prop](cobj, value)
+        end
+
+        -- In case there is already a "dumb" setter like `awful.client.property.set'
+        if args.setter_fallback then
+            return args.setter_fallback(cobj, prop, value)
+        end
+
+        local fallback = properties[cobj] or cobj_register(cobj)
+
+        -- Use the fallback property table
+        fallback[prop] = value
+
+        -- Emit the signal
+        if args.auto_emit then
+            cobj:emit_signal("property::"..prop, value)
+        end
+    end
+
+    -- Attach the accessor methods
+    class.set_index_miss_handler(getter)
+    class.set_newindex_miss_handler(setter)
+end
+
+return setmetatable( object, {__call = function(_,...) object.capi_index_fallback(...) end})
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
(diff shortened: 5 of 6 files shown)


lib/gears/object.lua
```diff
@@ -157,6 +157,95 @@ function object.modulename(level)
...
+-- * **getter_class**: A module with individual property getter/setter
+-- * **getter_prefix**: A special getter prefix (like "get" or "get_" (default))
+-- * **setter**: A smart setter (handle property setter itself)
+-- * **setter_fallback**: A dumb setter method (don't handle individual property setter)
+-- * **setter_class**: A module with individual property getter/setter
+-- * **setter_prefix**: A special setter prefix (like "set" or "set_" (default))
+-- * **auto_emit**: Emit "property::___" automatically (default: false). This is
+--     ignored when setter_fallback is set or a setter is found
+--
+-- @param cobj A standard luaobject derived object
```
blueyed: `tparam` and correct arg name.

blueyed:
Could the extra logic from `tag.setproperty` then be removed, and the `setter` path be dropped?

I think it's better to have the look-for-setter-code only in a single place.

lib/awful/client.lua
```diff
@@ -1121,6 +1123,14 @@ capi.client.connect_signal("unmanage", client.floating.delete)
 -- Register persistent properties
 client.property.persist("floating", "boolean")
 
+-- Extend the luaobject
+object.capi_index_fallback(capi.client, {
+    getter_class    = client             ,
+    setter_class    = client             ,
```
blueyed: I personally do not like this alignment / extra whitespace..

lib/awful/tag.lua
```diff
@@ -873,6 +874,14 @@ function tag.mt:__call(...)
     return tag.new(...)
 end
 
+-- Extend the luaobject
+object.capi_index_fallback(capi.tag, {
+    getter_class    = tag             ,
+    setter_class    = tag             ,
+    getter_fallback = tag.getproperty ,
+    setter          = tag.setproperty ,
```
blueyed: 2. I personally do not like this alignment / extra whitespace..

lib/gears/object.lua
```diff
@@ -157,6 +157,95 @@ function object.modulename(level)
     return debug.getinfo(level, "S").source:gsub(".*/lib/", ""):gsub("/", "."):gsub("%.lua", "")
 end
 
+local properties = setmetatable({}, { __mode = 'k' })
+
+local function cobj_register(cobj)
+    local fallback = {
+    }
```
blueyed: Can be joined with previous line.

lib/gears/object.lua
```diff
@@ -157,6 +157,95 @@ function object.modulename(level)
...
+
+        -- Emit the signal
+        if args.auto_emit then
+            cobj:emit_signal("property::"..prop, value)
+        end
+    end
+
+    -- Attach the accessor methods
+    class.set_index_miss_handler   (getter)
+    class.set_newindex_miss_handler(setter)
```
blueyed: (I personally do not like this alignment / extra whitespace..)

lib/gears/object.lua
```diff
@@ -157,6 +157,95 @@ function object.modulename(level)
...
+end
+
```
psychon: Can this be put into a separate file instead of `gears.object`? It doesn't have much interaction with the rest of the file (I'm trying to avoid something like `awful.util` happening again...)

psychon:
Welcome to "why psychon didn't want to implement this himself". The following test highlights a memory leak with Lua 5.1 / LuaJIT with your code (that AFAIK does not occur with newer Lua versions, should we drop Lua 5.1 support?):
```lua

print("Lua version:", _VERSION)

-- This should as well work with clients, but uses keys because they are easier
-- to create.
object.capi_index_fallback(key, {
    getter_class = key,
    setter_class = key
})

function test_leak(callback)
    local k = key{}
    local objs = setmetatable({ k }, { __mode = "v" })
    callback(k)
    k = nil

    for i = 0, 10 do -- Better too often than too few
        collectgarbage("collect")
    end
    for _, v in pairs(objs) do
        print("Some object was not garbage collected!")
        error(v)
    end
    print("Test successful")
end

print("Without doing anything")
test_leak(function() end)

print("While using some properties")
test_leak(function(k)
    k.foo = 42
    print("42 == ", k.foo)
end)

print("self-referential")
test_leak(function(k)
    k.self = k
    print("k.self == ", k.self)
end)

```
(I didn't think about the API yet, I just wanted to make it clear why I only did the C part of this)

Setting `k.self = k` causes the memory leak. While doing it directly like this might seem weird, it is actually quite easy to cause something like this through tables / functions-with-upvalues. Solving this in a nice way will be interesting.

Elv13:
> Could the extra logic from tag.setproperty then be removed, and the setter code path (for the "smart 
> setter") be dropped?
> 
> I think it's better to have the look-for-setter-code only in a single place.

Eventually, yes, but that will require splitting tag.lua and client.lua in 2 files first (with tons of deprecation warnings) as tag.lua argument order for setters is different from everything else (hello consistency!). So, until this is done, only a smart setter can make it work.

lib/awful/tag.lua
Elv13: IMO, it make the code faster to read as it highlight the structure and code affinity. Just as tabs/space, this is a polarizing topic with no general consensus. I wont go religious on this, but I prefer that style and it would not be the first place where it is in use (but there is not that many). The big problem with that style is "git-blame"/patch noise unless you enable the ignore-whitespace option.

Elv13:
> Setting k.self = k causes the memory leak. 

Replacing 

```lua
local fallback = {}
```

by

```lua
local fallback = setmetatable({}, { __mode = 'v' })
```

Make the test pass, but also render the thing useless... It is also probably unavoidable, on purpose. Sorting/safekeeping of properties are the point of this after all. While it is true that "pure lua" tables would manage to GC themselves anyway, there is another layer of coupling here that seem to mess Lua 5.1 GC.

In a more interesting note, using awful.tag.get/setproperty also fail your test, so how is this a regression from the current state of things?

```lua
-- local runner = require("_runner")
local atag = require("awful.tag")

print("Lua version:", _VERSION)

-- object.capi_index_fallback(tag, {
--     getter_class    = atag            ,
--     setter_class    = atag            ,
--     getter_fallback = atag.getproperty ,
--     setter          = atag.setproperty ,
-- })

function test_leak(callback)
    local k = tag{}
    local objs = setmetatable({ k }, { __mode = "v" })
    callback(k)
    k = nil

    for i = 0, 10 do -- Better too often than too few
        collectgarbage("collect")
    end
    for _, v in pairs(objs) do
        print("\n\nFAILED: Some object was not garbage collected!")
        os.exit(1)
    end
    print("Test successful")
end

print("Without doing anything")
test_leak(function() end)

print("While using some properties")
test_leak(function(k)
    atag.setproperty(k, "foo", 42)
    print("42 == ", atag.getproperty(k, "foo"))
end)

print("self-referential")
test_leak(function(k)
--     k.self = k
    atag.setproperty(k, "self", k)
    print("k.self == ",  atag.getproperty(k, "self"))
end)

os.exit(0) --quit the mainloop
```

(modified to run with `awesome -c`)

I am not trying to argue that because it leaks now, then leaks are ok. But at the same time, it would be hard to avoid this test-case failing and the current API already ignore it with a nice "blame the user". Any large config will probably fail to GC clients unless manually fixed (mine leak clients, I know and never cared much, it would take a decade to fill the ram).

Given this PR just call awful.tag.setproperty (it skip the "fallback" implementation), then it doesn't change anything.

> should we drop Lua 5.1 support?

Dropping Lua 5.1 isn't possible. First, Lua 5.1 is like python 2.7, it is the most used and distributed version of Lua (as far as I can see). Also, LuaJIT default to 5.1 (because it is the most common target). Dropping LuaJIT is a big "no". So, if you want to deprecate "CLua" 5.1 support and print a big warning in CMake, then fine by me, but dropping the 5.1 language itself, please don't.

Elv13:
@psychon  By the way, while I think about it. What about splitting this PR in 2. The first PR could just add the boilerplate to call the getters/setters and a later one could handle random properties. Would that be uncontroversial enough to be merged (given I fix the other complains)?

psychon:
Ah, I didn't know that we already have that memleak-problem right now. That makes this less of an issue and I can live with partly ignoring this for now.

lib/gears/object.lua
```diff
@@ -157,6 +157,95 @@ function object.modulename(level)
...
+-- * **getter**: A smart getter (handle property getter itself)
+-- * **getter_fallback**: A dumb getter method (don't handle individual property getter)
+-- * **getter_class**: A module with individual property getter/setter
+-- * **getter_prefix**: A special getter prefix (like "get" or "get_" (default))
+-- * **setter**: A smart setter (handle property setter itself)
+-- * **setter_fallback**: A dumb setter method (don't handle individual property setter)
+-- * **setter_class**: A module with individual property getter/setter
+-- * **setter_prefix**: A special setter prefix (like "set" or "set_" (default))
+-- * **auto_emit**: Emit "property::___" automatically (default: false). This is
+--     ignored when setter_fallback is set or a setter is found
```
psychon: Should/Can this also have some magic to auto add signals for new properties? I don't really know how that should work, though... It's just annoying to have to add the property signal.

psychon:
Feel free to clean this up and handle your TODOs. The approach sounds good enough to me.
(It would be nice if the weak table could be avoided by using an upvalue for the storage table, like in your code that you linked to somewhere, but whatever...)

Elv13:
How should the new "classes" should be called? I guess they could go into a new module. Trying to fix awful will be a huge mess. There is a mix of accessors, mutator and helper methods in the same namespace/module. No amount of dark magic can fix this, so the mutators and accessors need to be moved out and renamed to match a standard name policy (and stick to it, this time).

Previous discussion https://github.com/awesomeWM/awesome/pull/631#issuecomment-172440330

As far the upvalue version referenced above, the code is here
https://github.com/Elv13/radical/blob/master/object.lua

It was designed for the first implementation of the dynamic layout system in 2012. Most of my modules use it for the user facing APIs. It has been reliable for a long time, I will integrate it at some point, but not yet. It had been on my TODO list for a long time, but lets start with the more simple version provided by this PR. The Radical version require some kind of constructor callback from CAPI objects while @psychon PR use a global fallback callback per class. So sequence isn't directly compatible (but that can be fixed)

As for the signals, it is totally possible to remove the need to use add_signal. I would just turn it off, my fork/branch did that for a long time. It has always been noise to me and provide no real value. Note that my object class doesn't have the "add_signal" method, it do it implicitly. All it does is to warn of typos, but consider every unknown name a typo, so the signal to noise ratio is quite high. Maybe `gears.debug` could have a method to turn it on for the tests and leave it off for regular builds, like most asserts implementations.

Elv13:
@psychon Can you give you opinion on the few questions from my previous comment? May I call the new module "core" for the lack of a better name?

psychon:
Name for the module: Uhm... `gears.object_properties`? `gears.properties`?

"The Radical version require some kind of constructor callback from CAPI objects": All CAPI objects have a `new` signal (but are not yet fully constructed and usable at the time that `new` is emitted)
If needed, I can also make this "miss-callback" per-object, but that didn't seem necessary when I wrote the current version of this.

I like the idea to have some "strict signal"-mode for our tests... but could also live with removing `add_signal` completely.

Elv13:
> but that didn't seem necessary when I wrote the current version of this.

I don't think it would help. To use the Radical implementation, the C side need call the function when the object is created and store the result (put the table into the global index and keep the reference number. In that case, the upvalue would do its job (at least, I think so) and prevent the leak.

> gears.object_properties? gears.properties?

I wasn't referring to that one (but I will use your name, it's better than the one I came up with). I was talking about all the code from `awful.client` and `awful.tag` I will have to move somewhere else. It cannot be in gears, as it will still depend on `awful`. The problem here is that `awful.*` modules are used for both helper functions and method. I want to place the methods somewhere else so `c:focus.history.get(idx, filter)` don't exist (the first argument is a screen, not a client, and it would produce a syntax error anyway)

psychon:
Uhm... so you want to separate e.g. `awful.client.run_or_raise` from e.g. `awful.client.dockable.get`? Assuming that the property-functions shouldn't be used directly anymore, you could move them to e.g. `awful.client.properties.dockable.get`? Or do you want to do it the other way around and have `awful.properties.client.dockable.get`?
Am I understanding the question correctly this time?

Elv13:
> Am I understanding the question correctly this time?

Yes

> Or do you want to do it the other way around and have awful.properties.client.dockable.get?

I was thinking about this one, but you `awful.client.properties.get_dockable` solution would work too. I am fine with either. I let you pick one. I would still prefer `get_dockable` to `dockable.get`, it save some nil checks and is already the most common getter/setter syntax in `wibox` and `gears`

psychon:
Ok, I prefer `awful.client.properties.dockable_{get,set}` (mostly so that the two functions are alphabetically close), but can also live with `{get,set}_dockable`, especially if it already is the most common syntax in `wibox` and `gears`.

raphael-costa:
02:27:13    Elv13 | Please leave a comment at https://github.com/awesomeWM/awesome/pull/757 so I don't forget to add a c.show_titlebar =     │ Enlik
                  | true/false property

Elv13:
Rebased and updated. Please review. If this part isn't too controversial anymore and it provide real value as-is, I propose to merge it and do the getter/setter/mutator refactoring in another PR (or one for each class).

lib/awful/tag.lua
```diff
@@ -873,6 +874,14 @@ function tag.mt:__call(...)
     return tag.new(...)
 end
 
+-- Extend the luaobject
+object.properties(capi.tag, {
+    getter_class    = tag,
+    setter_class    = tag,
+    getter_fallback = tag.getproperty,
+    setter          = tag.setproperty,
```
psychon: Perhaps a comment explaining why the getter is fallback, but the setter is not? I looked at the code and I think I understand the difference between the two, but I'm not sure why it's necessary to do it like this here.

psychon:
One minor comment (see above), besides that this looks fine. Adding to next.

lib/awful/tag.lua
Elv13: As @blueyed  pointed out, `awful.tag` have it's own property logic and this should be removed. Once I will have converted that to the new format, it will be possible to use the more simple:

```
 +object.properties(capi.tag, {
+    getter_class    = tag,
+    setter_class    = tag,
+    auto_emit       = true,
})
```

options, but as it is currently written, it would break stuffs, so I will do this in its own PR.

PR #762, opened 2016-03-17
Run shape-example-generation under LuaCov

psychon:
This is ugly, but at least it works. Hopefully...

(I'm looking forward to coveralls confirming the above.)

Edit: @Elv13 kind-of requested this in #753.


psychon:
This doesn't seem to work https://coveralls.io/builds/5448582/source?filename=lib%2Fgears%2Fshape.lua

Elv13:
I got an idea, but it will take a while to implement. If it is done right, it should run fine.

PR #763, opened 2016-03-17
Run shape-example-generation under LuaCov

psychon:
This is ugly, but at least it works. Hopefully...

I force-pushed the branch while #762 was closed and now it can't be reopened. So let's try this again. (And I only needed four more travis builds to get something that might actually work!)

### DIFF
--- a/.luacov
+++ b/.luacov
@@ -2,7 +2,8 @@
 
 -- This variable is set externally
 local build = os.getenv("BUILD_DIRECTORY") or error("$BUILD_DIRECTORY not set")
-local lib_dir = build .. "lib/"
+local lib_dir = os.getenv("AWESOME_LIB_DIR") or build .. "lib/"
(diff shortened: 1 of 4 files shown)


Elv13:
That's cool if it works, but again, don't waste too much time on this. While you wont like the first version of that PR (when it's ready), it will eventually be much cleaner than the pipeline I came up with (but also much larger, and it will still require to run before the pre-processor)

psychon:
It worked! https://coveralls.io/builds/5449017/source?filename=lib%2Fgears%2Fshape.lua

> (but also much larger, and it will still require to run before the pre-processor)
If you want code coverage to work, this should always run before the `@usage`-examples are expanded. coveralls.io gets the source from the git repo and thus the line numbers won't match up if there is a difference here (we already have this problem for the part of `gears.shape` that is executed by the functional / integration tests, but I guess we can just ignore this).

Looking forward to someone merging this. :-)

PR #773, opened 2016-03-23
tests: Add a new GUI testing framework

Elv13:
This framework allow widgets and user interface related code to be tested.

This should close the gap in our testing pipeline. It sits between unit testing and the full integration testing.

#### It has the following features:

 * Generate documentation examples
 * Generate SVG screenshots of widgets
 * Generate dynamic ldoc content
 * Emulate just enough of the C API for the tests to run
 * Tests corner cases like weird multi screen that are harder to test with Xephyr
 * provide a template mechanism to generate different kind of images. Each module can have their own template or inherit from the upper module.

#### Drawbacks:

 * With a thousand tests, the build get very slow, so should only be enabled on the CI. I still have to test parallel test generation. Given the bottleneck is mostly cmake and fork/pipe, I could technically generate a test list, then run 16 in parallel. With the 60 tests I already wrote, it add ~7 seconds to the build.
 * It is currently 1 test per file. This cause tons of files to be created. The reason behind this is to clean the state at every run. For example, tests can set random virtual screen setup, but I don't want to have to write code to rollback everything.

#### WIP:

 * Eventually run some regression testing on the SVG files. A simple vector path comparison should detect regressions/changes. A github bot could report them in the pull request with before and after images in the comment.

@blueyed Do you have any feature you want in this before I get too far?

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -299,6 +299,11 @@ set(AWESOME_THEMES_PATH      ${AWESOME_DATA_PATH}/themes)
 if(GENERATE_DOC)
     # Generate some images and examples
     include(docs/generate_examples.cmake)
+
+    # Use `include`, rather than `add_subdirectory`, to keep the variables
+    # The file is a valid CMakeLists.txt and can be executed directly if only
+    # the image artefacts are needed.
+    include(tests/examples/CMakeLists.txt)
 endif()
 
 # {{{ Configure files
--- a/lib/gears/surface.lua
+++ b/lib/gears/surface.lua
@@ -11,6 +11,7 @@ local capi = { awesome = awesome }
 local cairo = require("lgi").cairo
 local color = nil
 local gdebug = require("gears.debug")
+local hierarchy = require("wibox.hierarchy")
 
 -- Keep this in sync with build-utils/lgi-check.sh!
 local ver_major, ver_minor, ver_patch = string.match(require('lgi.version'), '(%d)%.(%d)%.(%d)')
@@ -203,6 +204,53 @@ function surface.apply_shape_bounding(draw, shape, ...)
   cr:fill()
 
   draw.shape_bounding = img._native
+end
+
+local function no_op() end
+
+local function run_in_hierarchy(self, cr, width, height)
+
+    local function redraw(h)
+        h:draw({dpi=96}, cr)
+    end
+
+    local h = hierarchy.new({dpi=96}, self, width, height, redraw, no_op, {})
+
+    redraw(h)
+
+    return h
+end
+
+--- Create an SVG file with this widget content.
+-- This is dynamic, so the SVG will be updated along with the widget content.
+-- because of this, the painting may happen hover multiple event loop cycles.
+-- @tparam widget widget A widget
+-- @tparam string path The output file path
+-- @tparam number width The surface width
+-- @tparam number height The surface height
+-- @return The cairo surface
+-- @return The hierarchy
+function surface.widget_to_svg(widget, path, width, height)
+    local img = cairo.SvgSurface.create(path, width, height)
+    local cr = cairo.Context(img)
+
+    return img, run_in_hierarchy(widget, cr, width, height)
+end
+
+--- Create a cairo surface with this widget content.
+-- This is dynamic, so the SVG will be updated along with the widget content.
+-- because of this, the painting may happen hover multiple event loop cycles.
+-- @tparam widget widget A widget
+-- @tparam number width The surface width
+-- @tparam number height The surface height
+-- @param[opt=cairo.Format.ARGB32] format The surface format
+-- @return The cairo surface
+-- @return The hierarchy
+function surface.widget_to_surface(widget, width, height, format)
+    local img = cairo.ImageSurface(format or cairo.Format.ARGB32, width, height)
+    local cr = cairo.Context(img)
+
+    return img, run_in_hierarchy(widget, cr, width, height)
 end
 
 return setmetatable(surface, surface.mt)
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -1,0 +1,251 @@
+# This module is designed to allow the Awesome documentation examples to be
+# tested.
+#
+# It shims enough of the Awesome C API to allow code to be executed without an
+# actual X server or running Awesome process. These tests are not genuine
+# integration tests, but they are the next best thing.
+#
+# # As secondary goals, this module also generates images of the test result where
+# relevant. Those images are used by the documentation and help the developers
+# track user interface regressions and glitches. Finally, it also helps to find
+# broken code.
+cmake_minimum_required(VERSION 3.0.0)
+
+# Get and update the LUA_PATH so the scripts can be executed without Awesome.
+execute_process(COMMAND lua -e print\(package.path\) OUTPUT_VARIABLE "LUA_PATH_")
+
+# Add the main awesome lua libraries.
+set(LUA_PATH2_ "\
+${CMAKE_SOURCE_DIR}/lib/?.lua;\
+${CMAKE_SOURCE_DIR}/lib/?/init.lua;\
+${CMAKE_SOURCE_DIR}/lib/?;"
+)
+
+# Add the C API shims.
+set(LUA_PATH3_ "\
+${CMAKE_SOURCE_DIR}/tests/examples/shims/?.lua;\
+${CMAKE_SOURCE_DIR}/tests/examples/shims/?/init.lua;\
+${CMAKE_SOURCE_DIR}/tests/examples/shims/?;"
+)
+
+# Done in 3 variables to avoid CMake from implicitly converting into a list.
+set(ENV{LUA_PATH} "${LUA_PATH3_}${LUA_PATH2_}${LUA_PATH_}")
+
+# The documentation images directory
+set(IMAGE_DIR "${CMAKE_BINARY_DIR}/doc/images")
+file(MAKE_DIRECTORY "${IMAGE_DIR}")
+
+# Escape potentially multiline strings to be part of the API doc.
+#  * add "--" in front of each lines
+#  * add a custom prefix in front of each lines
+#  * drop empty lines
+#  * convert " " lines into empty lines
+#  * drop lines ending with "--DOC_SOMETHING", they are handled elsewhere
+function(escape_string variable content escaped_content line_prefix)
+    # If DOC_HIDE_ALL is present, do nothing
+    if(variable MATCHES "--DOC_HIDE_ALL")
+        return()
+    endif()
+
+    string(REGEX REPLACE "\n" ";" var_lines "${variable}")
+
+    set(tmp_output ${content})
+    foreach (LINE ${var_lines})
+        if(NOT LINE MATCHES "^.+--DOC_[A-Z]+$")
+            set(tmp_output ${tmp_output}\n--${line_prefix}${LINE})
+        endif()
+    endforeach()
+
+    set(${escaped_content} ${tmp_output} PARENT_SCOPE)
+endfunction()
+
+# Extract lines with the --DOC_HEADER marker
+function(extract_header variable pre_output post_output)
+    string(REGEX REPLACE "\n" ";" var_lines "${variable}")
+
+    set(IS_POST 0)
+
+    foreach (LINE ${var_lines})
+        # This function doesn't escape the lines, so make sure they are
+        # already comments.
+        if(LINE MATCHES "^--.*--DOC_HEADER")
+            # Remove the header tag
+            string(REGEX REPLACE "[ ]*--DOC_HEADER" "" LINE "${LINE}")
+
+            # ldoc is picky about what happen after the first --@, so split
+            # the output between all that come before and all that come after.
+            if (NOT IS_POST AND LINE MATCHES "^--[ ]*@")
+                set(IS_POST 1)
+            endif()
+
+            if (IS_POST)
+                set(tmp_post_output ${tmp_post_output}\n${line_prefix}${LINE})
+            else()
+                set(tmp_pre_output ${tmp_pre_output}\n${line_prefix}${LINE})
+            endif()
+        endif()
+    endforeach()
+
+    set(${post_output} "${tmp_post_output}\n--" PARENT_SCOPE)
+    set(${pre_output} "${tmp_pre_output}\n--" PARENT_SCOPE)
+endfunction()
+
+# Read a code file and convert it to ldoc usage example.
+#  * add "--" in front of each lines
+#  * drop empty lines
+#  * convert " " lines into empty lines
+#  * drop lines ending with "--DOC_HIDE"
+function(escape_code path escaped_content pre_header post_header)
+    file(READ ${path} path)
+
+    escape_string("${path}\n" "" escaped_code "")
+
+    extract_header("${path}" example_pre_header example_post_header)
+
+    set(${escaped_content} ${escaped_code} PARENT_SCOPE)
+    set(${pre_header} ${example_pre_header} PARENT_SCOPE)
+    set(${post_header} ${example_post_header} PARENT_SCOPE)
+endfunction()
+
+# Execute a lua file.
+function(run_test test_path namespace template escaped_content)
+
+    # A template is required to know how to handle the output.
+    if (template STREQUAL " ")
+        message(FATAL_ERROR "No template found for " ${test_path} ", bye")
+    endif()
+
+    # Get the file name without the extension
+    get_filename_component(${test_path} TEST_FILE_NAME NAME)
+    set(IMAGE_PATH "${IMAGE_DIR}/${namespace}_${TEST_FILE_NAME}")
+
+    # Execute the script, leave the image extension decision to the test
+    # SVG is preferred, but PNG is better suited for some tests, like bitmap
+    # patterns.
+    execute_process(
+        COMMAND lua ${template} ${test_path} ${IMAGE_PATH} ${SOURCE_DIR}/.luacov
+        OUTPUT_VARIABLE TEST_OUTPUT
+        ERROR_VARIABLE  TEST_ERROR
+    )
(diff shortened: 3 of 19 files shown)


Elv13:
Here are pdf (print) output of the `widget` template  and the `screen+client+mouse` template.

[wibox_widget_background.pdf](https://github.com/awesomeWM/awesome/files/185708/wibox_widget_background.pdf)
[awful_placement.pdf](https://github.com/awesomeWM/awesome/files/185709/awful_placement.pdf)

Please ignore all the new code in these files, it is work in progress, full of typo and there is good reasons why the placement rewrite design is a little weird.

The screen template is not done yet, it is missing screen, mouse and client text labels and visual indicator of the initial states versus the result.

awesomeConfig.cmake
```diff
@@ -299,6 +299,11 @@ set(AWESOME_THEMES_PATH      ${AWESOME_DATA_PATH}/themes)
 if(GENERATE_DOC)
     # Generate some images and examples
     include(docs/generate_examples.cmake)
+
+    # Use `include` rather than `add_subdirectory` to keep the variables
```
blueyed: Punctuation!
Don't need to change it now, but please consider it in the future!

awesomeConfig.cmake
```diff
@@ -299,6 +299,11 @@ set(AWESOME_THEMES_PATH      ${AWESOME_DATA_PATH}/themes)
...
+    # The file is a valid CMakeLists.txt and can be executed directly if only
+    # the images artefacts are needed.
```
blueyed: s/images/image/

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
+# Module is designed to allow the Awesome documentation examples to be tested.
```
blueyed: s/Module/This module/

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+#
+# It shim enough of Awesome C API to allow code to be executed without an
```
blueyed: s/shim/shims/

blueyed: s/of Awesome/of the Awesome/

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+# actual X server or running Awesome process. These test are not genuine
```
blueyed: s/test/tests/

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+# integration tests, but they are the next best thing.
+#
+# As secondary goals, this module also generate images of the test result where
```
blueyed: s/generate/generates/

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+# relevant. Those images are used by the documentation and help developpers
```
blueyed: s/developpers/developers/

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+# track user interface regressions and glitches. Finally, it also help to find
```
blueyed: s/help/helps/

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+# broken code.
+cmake_minimum_required(VERSION 3.0.0)
+
+# Get and update the LUA_PATH so the scripts can be executed without Awesome
```
blueyed: dot-at-the-end.

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+execute_process(COMMAND lua -e print\(package.path\) OUTPUT_VARIABLE "LUA_PATH_")
+
+# Add the main awesome lau libraries
```
blueyed: Another missing dot.

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+# set(${LUA_PATH_} "${SOURCE_DIR}/lib/?;${SOURCE_DIR}/lib/?.lua;${LUA_PATH_}")
```
blueyed: Remove the commented code?

blueyed: Or is it useful for un-commenting locally?

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+set(LUA_PATH2_ "\
+${CMAKE_SOURCE_DIR}/lib/?.lua;\
+${CMAKE_SOURCE_DIR}/lib/?/init.lua;\
+${CMAKE_SOURCE_DIR}/lib/?;"
+)
+
+# Add the C API shims
+# set(${LUA_PATH_} "${SOURCE_DIR}/lib/?;${SOURCE_DIR}/lib/?.lua;${LUA_PATH_}")
+set(LUA_PATH3_ "\
```
blueyed: Same as above.

Elv13:
Updated, I also added some of the other WIP commits. The shims are currently terrible and full of copy pasted hacks. It's good enough to develop with, but it will have to be cleaned up before merging.

The shims can not also emulate tags enough to play with them, as long as you only have 1 per virtual screen (enough for testing most of awful code paths). Other than that, it is still an early WIP.

Elv13:
[tile.pdf](https://github.com/awesomeWM/awesome/files/189197/tile.pdf)

The new tile layout documentation (WIP)

Elv13:
I removed the `[WIP]` tag. So here it is. The shims layer is a little messy, but good enough to do its job. It doesn't have to be perfect now as long as it allow the tests to run. This PR include tests for the background widget as an example. Coverall seem to work. The CMake part is good enough, even if a little slow. About that, it is still faster than the Xephyr tests, so I guess it is fine.

@psychon I know this is a big, scary and far reaching. The logic behind investing time into this to make it happen is for what it will eventually bring. The documentation improvement are nice, but being able to cover parts of that code that was previously very hard to cover is what will eventually bring Awesome forward. Yes, all those tests represent *a lot* of code. It is not fun to write, I will tell you that, but overall I am convinced it is a step in the right direction.

The new `awful.placement` and my new client layout system are tested using this, so it has to be merged so I can advance on the other fronts.

lib/gears/surface.lua
```diff
@@ -205,6 +206,52 @@ function surface.apply_shape_bounding(draw, shape, ...)
...
+    end
+
+    local function redraw(h)
+        h:draw({dpi=96}, cr)
+    end
+
+    local h = hierarchy.new({dpi=96}, self, width, height, redraw, main_layout, {})
+
+    main_layout(h)
+    h:_redraw()
```
psychon: I'd replace this with `redraw(h)` instead. However, I don't think this method is necessary (see below).

lib/gears/surface.lua
```diff
@@ -205,6 +206,52 @@ function surface.apply_shape_bounding(draw, shape, ...)
...
+-- because of this, the painting may happen hover multiple event loop cycles.
+-- @tparam widget widget A widget
+-- @tparam string path The output file path
+-- @tparam number width The surface width
+-- @tparam number height The surface height
+function surface.widget_to_svg(widget, path, width, height)
+    local img = cairo.SvgSurface.create(path, width, height)
+    local cr = cairo.Context(img)
+
+    return run_in_hierarchy(widget, cr, width, height)
```
psychon: Replace this with `hierarchy.new({dpi=96}, widget, width, height, function() end, function() end, {}):draw({dpi=96}, cr)` (for extra niceness, you can make and use a `local function no_op() end` and move the context into a variable, too).

There will be no relayouts necessary here and no redraws either (in fact, redraws would be bad, because the widget would be drawn to the SVG twice).

(Same thing below)

psychon: Also, add `img:finish()` at the end (only then is the SVG file actually written, right now it depends on the garbage collector to finish the SVG Surface)

lib/gears/surface.lua
```diff
@@ -205,6 +206,52 @@ function surface.apply_shape_bounding(draw, shape, ...)
...
+-- This is dynamic, so the SVG will be updated along with the widget content.
+-- because of this, the painting may happen hover multiple event loop cycles.
+-- @tparam widget widget A widget
+-- @tparam number width The surface width
+-- @tparam number height The surface height
+-- @param[opt=cairo.Format.ARGB32] format The surface format
+function surface.widget_to_surface(widget, width, height, format)
+    local img = cairo.ImageSurface(format or cairo.Format.ARGB32, width, height)
+    local cr = cairo.Context(img)
+    local h = run_in_hierarchy(widget, cr, width, height)
```
psychon: Ok, perhaps the code suggested above should be moved into a helper function. Something like `layout_and_draw_widget(widget, cr, width, height)`?

lib/gears/surface.lua
```diff
@@ -205,6 +206,52 @@ function surface.apply_shape_bounding(draw, shape, ...)
...
+    return run_in_hierarchy(widget, cr, width, height)
+end
+
+--- Create a cairo surface with this widget content.
+-- This is dynamic, so the SVG will be updated along with the widget content.
+-- because of this, the painting may happen hover multiple event loop cycles.
+-- @tparam widget widget A widget
+-- @tparam number width The surface width
+-- @tparam number height The surface height
+-- @param[opt=cairo.Format.ARGB32] format The surface format
```
psychon: Return values are undocumented.

tests/examples/shims/awesome.lua
```diff
@@ -0,0 +1,27 @@
...
+function awesome.register_xproperty()
+    
+end
+
+awesome:add_signal("refresh")
+awesome:add_signal("wallpaper_changed")
+awesome:add_signal("spawn::canceled")
+awesome:add_signal("spawn::timeout")
+
+return awesome
```
psychon: Modeline is missing

tests/examples/shims/client.lua
```diff
@@ -0,0 +1,202 @@
+local gears_obj = require("gears.object")
+
+local clients = {}
+
+--TODO merge all other implementations
```
psychon: How about moving this to `awesome._shim_convert_singleton`?

tests/examples/shims/client.lua
```diff
@@ -0,0 +1,202 @@
...
+local function convert_singleton(obj)
+    obj._connect_signal = obj.connect_signal
+    function obj.connect_signal(name, func)
+        return obj._connect_signal(obj, name, func)
+    end
+    obj.weak_connect_signal = obj.connect_signal
```
psychon: The C-API doesn't have weak signal connections, so I think this part can be dropped.

tests/examples/shims/client.lua
```diff
@@ -0,0 +1,202 @@
...
+
+
+    add_signals(ret)
+
+    -- Emulate capi.client.geometry
+    function ret:geometry(new)
+        if new then
+            for k,v in pairs(new) do
+                ret:emit_signal("property::"..k, v)
+                ret[k] = v
```
psychon: I think the order is wrong, you should first change the value and then emit the signal (and only emit if the value actually changed)

tests/examples/CMakeLists.txt
```diff
@@ -0,0 +1,251 @@
...
+    endif()
+
+    # Get the file name without the extension
+    get_filename_component(${test_path} TEST_FILE_NAME NAME)
+    set(IMAGE_PATH "${IMAGE_DIR}/${namespace}_${TEST_FILE_NAME}")
+
+    # Execute the script, leave the image extension decision to the test
+    # SVG is preferred, but PNG is better suited for some tests, like bitmap
+    # patterns.
+    execute_process(
```
psychon: Should there be a `message(STATUS "Doing foo")` here, so that we know that it is executing something like this? If not, should there be such a message at the beginning and at the end of the file so that one can at least see how much time is spent here in the build?

tests/examples/awful/template.lua
```diff
@@ -0,0 +1,159 @@
+local file_path, image_path, luacovpath = ...
+
+local cairo      = require("lgi").cairo
+local pango      = require("lgi").Pango
+local pangocairo = require("lgi").PangoCairo
+
+-- Set the global shims
+-- luacheck: globals awesome
```
psychon: I think it is enough to do `luacheck: globals awesome root tag screen client mouse`, it applies from "here" to the end of the file (unless the comment is on a not-otherwise-empty line)

tests/examples/wibox/template.lua
```diff
@@ -0,0 +1,63 @@
...
+
+        -- There is an overhead that cause testbox "...", add 10 until someone
+        -- figures out the real equation
+        f_w, f_h = f_w+10, f_h+10
+
+        -- Save to the output file
+        surface["widget_to_"..image_type](container, image_path.."."..image_type, w or f_w, h or f_h)
+
+        -- Quit normally, else the .svg are going to be empty
+        main_loop:quit()
```
psychon: This will be fixed by my `:finish()` suggestion on the commit that adds these methods

tests/examples/wibox/template.lua
```diff
@@ -0,0 +1,63 @@
...
+
+-- Let the test request a size and file format
+local w, h, image_type = loadfile(file_path)(container)
+image_type = image_type or "svg"
+
+-- Emulate 10 cycles of mainloop
+local counter = 1
+
+-- Loop as fast as possible
+GLib.timeout_add_seconds(0, 0,function()
```
psychon: Can you try using `add_idle` (I'm not totally sure about the name). This function will be called whenever there is nothing else to do, which seems better than the 0 second timeout.

Actually... does anything in awesome-lua-land use the GLib main loop? I think you can replace all of this with
```
for counter = 1, 10 do
  awesome:emit_signal("refresh")
end
```
here comes the code you do on the 10th iteration

psychon:
Thanks for working on this. You are right that this looks a bit scary, but I'll live with it. My comments above are mostly minor things.

tests/examples/wibox/template.lua
```diff
@@ -0,0 +1,63 @@
...
+    else
+        -- Get the example fallback size (the tests can return a size if the want)
+        local f_w, f_h = container:fit({dpi=96}, 9999, 9999)
+
+        -- There is an overhead that cause testbox "...", add 10 until someone
+        -- figures out the real equation
+        f_w, f_h = f_w+10, f_h+10
+
+        -- Save to the output file
+        surface["widget_to_"..image_type](container, image_path.."."..image_type, w or f_w, h or f_h)
```
psychon: There is only `widget_to_svg`, or do I miss something?

lib/gears/surface.lua
Elv13: Finish was already in one of the template, now, both have it. As for removing the local function, I prefer not, I don't think it cover all corner cases such as creating new images for each "frame" of the main loop. It is currently not implemented, but as it would add some (~5 lines) of new logic, the code duplication would start to be more problematic. Also, there is a PNG version too. I just didn't add it in this commit as it was broken and I have no tests that require bitmap output yet.

tests/examples/wibox/template.lua
Elv13: There is the png version too, it's not in this PR, but it will come when I will need it. This PR is already 1k line with only the code I need...

(there is no huge second part, don't worry ;) ).

tests/examples/CMakeLists.txt
Elv13: done

lib/gears/surface.lua
Elv13: done

Elv13:
Updated. I only partially ignored the `gears.surface` comments. By the way, the SVG is not written twice, it is written on `:finish()` and the "CLEAR" operator do it's job, so it is not written on top of itself.

psychon:
Thanks. And yeah, you are right, drawing the background should handle multiple frames. It's still unneccessary here to handle re-layouts and re-draws, but if you insist...

lib/gears/surface.lua
```diff
@@ -205,6 +206,55 @@ function surface.apply_shape_bounding(draw, shape, ...)
   draw.shape_bounding = img._native
 end
 
+local function run_in_hierarchy(self, cr, width, height)
+    local function main_layout(h)
+        h:update({dpi=96}, self, width, height)
```
psychon: Hmm. This doesn't handle redraws due to relayouts, does it? That part of the code works with a `cairo.Region` that is returned from `:update()`.

Elv13: Your better placed than me to know your code, but so far it worked fine

psychon: Because no widget can emit any signals that this hierarchy should react to, so this code is never actually executed. ;-)
As I said, for the code you have right now, you could just use no-op-functions for these callbacks.

Elv13: Are you sure? I am just asking, but isn't `widget::layout_changed` supposed to end up calling this? If I create an `.svg` at main loop cycle `#1`, then do `:finish()` at cycle `#10` after chaning the widget 9 time, isn't it better to have this called? If not, then I will remove it, test and merge this PR as it is the only remaining issue.

psychon: Hm... apparently I misunderstood your intention with this function.
In the following code, I'd espect a SVG saying "foo" and not "bar":
```
local t = wibox.widget.textbox("foo")
gears.surface.widget_to_svg(t, "/tmp/t.svg", 100, 100)
t:set_text("bar")
```
Apparently, you want it to say `bar`. And mentioned something like "multiple frames" being produced. I don't really get this.

Elv13: nevermind, one way or another, it is unused. I will remove it for now. Cairo doesn't yet support multiframe SVG anyway.

Elv13:
Umm, the examples seem to be processed, but not kept by the awesome-robot. I probably missed something, I will check that tomorrow. It's probably minor as it work locally with the same script.

PR #776, opened 2016-03-26
Awful screen improvements

Elv13:
This pull request is the first in a quite long queue of enhancements to Awesome handling of various geometries. Those PRs will progressively remove hardcoded geometry code and move them in places where the code can be shared/tested/extended.

Testing done: My client layout tests still pass. It seem to work with the default config. The new `awful.placement` (to be published later) unit tests also rely on this code and pass.

Please note that get_bounding_geometry use `t.screen`. This wont work until the dynamic property work is integrated. Given it will soon be, I chose to ignore the issue. This avoid a circular dependency hack.

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -136,27 +136,18 @@ function layout.parameters(t, screen)
 
     local p = {}
 
-    p.workarea = screen.workarea
-
     local useless_gap = t and tag.getgap(t, #client.tiled(screen)) or 0
 
-    -- Handle padding
-    local padding = ascreen.padding(screen) or {}
-
-    p.workarea.x = p.workarea.x + (padding.left or 0) + useless_gap
-
-    p.workarea.y = p.workarea.y + (padding.top or 0) + useless_gap
-
-    p.workarea.width = p.workarea.width - ((padding.left or 0 ) +
-        (padding.right or 0) + useless_gap * 2)
-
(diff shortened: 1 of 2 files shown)


psychon:
Fine by me (if Travis passes)

Elv13:
Damn, the test passed but the merging ones didn't???

Elv13:
```
2016-03-26 07:45:44 W: awesome: luaA_dofunction:77: error while running function!
stack traceback:
	[string "-- Some memory leak checks as integration tes..."]:35: in function 'collectable'
	[string "-- Some memory leak checks as integration tes..."]:63: in function 'f'
	lib/awful/remote.lua:25: in function <lib/awful/remote.lua:21>
```
Only one of the target failed....

Elv13:
@psychon Any idea? The PR had "2 checks passed" (see above), but now the merged result fail on luajit 5.1. It's not obvious why. There is a leak, but beside the screen padding commit, I see no obvious new sources of leaks...

psychon:
I restarted the travis build for master+luajit. This time, the build passed. Yay.
Line 63 in the test file is the leak test for the layoutbox. Layoutboxes are "kind of singletons" that are kept in a weak table in `awful.widget.layoutbox`. There is also a tooltip referencing the layoutbox, but I can't see which problems that would cause... So: Dunno what is going on.

PR #778, opened 2016-03-26
Call collectgarbage() more often

Elv13:
I will revert all commits one by one until it passes again

### DIFF
--- a/tests/test-leaks.lua
+++ b/tests/test-leaks.lua
@@ -29,6 +29,7 @@ local function collectable(a, b, c, d, e, f, g, h, last)
     end
     collectgarbage("collect")
     collectgarbage("collect")
+    collectgarbage("collect")
     -- Check if the table is now empty
     for _, v in pairs(objs) do
         print("Some object was not garbage collected!")


blueyed:
travis-bisect? :D

Elv13:
I wasn't aware of that, but in this was it would not help, because only the last commit in the queue use the code introduced by the previous ones. So far it seem like a cosmic rays. This build just passed and the only difference is

```
    ascreen.get_bounding_geometry(screen, {
        honor_padding  = true,
        honor_workarea = true,
        margins        = useless_gap,
    })

    p.workarea = {x=0,y=0,width=100,height=100}
```

instead of 
```
    p.workarea = ascreen.get_bounding_geometry(screen, {
        honor_padding  = true,
        honor_workarea = true,
        margins        = useless_gap,
    })
```

Elv13:
Ok, so the build just passed with the exact same code as master HEAD. I only reverted the commit and cherry-picked it again. But yet, the build failed **twice** in a row with the same error.

So, my code should not have introduced a regression, but something is wrong with the test. Maybe `collectgarbage()` now need to be called more often in the test?

/me love heisenbug

Elv13:
My build pool is still small, but apparently, adding more `collectgarbage()` seem to prevent the heisenbug.

psychon:
Fine with me to add another call to `collectgarbage`. And no, I do now know why/when it is necessary...

Elv13:
> And no, I do now know why/when it is necessary...

I don't even know if it is necessary. All I can say is that the problem did not happen in the builds using this commit while it did happen 2 out of 3 builds without it earlier today. But it also doesn't happen with your new PR, so the correlation between this commit and the build failure this morning is very low.

psychon:
If you want a random guess: It depends on the speed of the host that travis runs our build on (and on the load on that host). Some things will be done in two separate main loop iteration in slow builds, but perhaps in the same iteration on fast builds. This tiny difference leads to tiny differences in the state of the garbage collector and the result is what we see here. Perhaps a new GC cycle starts before/after some specific operation and thus some garbage is only "recognized" later?

Elv13:
Well, it is not like this change has any drawbacks beside not being a solution, but rather a quick fix. So, I will merge given it may have a little chance to avoid another stress bomb at 5 AM.

PR #779, opened 2016-03-26
Return screen objects

psychon:
Another step towards the deprecation of screen indicies. Instead of numbers, the C API (e.g. `mouse.screen`, `c.screen` etc) now return screen objects. This should cause some breakage.

https://github.com/awesomeWM/awesome/issues/672

### DIFF
--- a/mouse.c
+++ b/mouse.c
@@ -133,13 +133,13 @@ luaA_mouse_index(lua_State *L)
          * having lots of lua errors in this case.
          */
         if (globalconf.focus.client)
-            luaA_pushscreen(L, globalconf.focus.client->screen);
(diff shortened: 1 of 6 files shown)


mouse.c
```diff
@@ -133,13 +133,13 @@ luaA_mouse_index(lua_State *L)
          * having lots of lua errors in this case.
          */
         if (globalconf.focus.client)
-            luaA_pushscreen(L, globalconf.focus.client->screen);
+            luaA_object_push(L, globalconf.focus.client->screen);
```
Elv13: Please update the ldoc comment at the top of this file

psychon: Done (in a new commit)

PR #780, opened 2016-03-26
Add screen add signal

psychon:
Another (big) step towards https://github.com/awesomeWM/awesome/issues/672. This depends on https://github.com/awesomeWM/awesome/pull/779 to fix the functional tests (otherwise the tests fail for unrelated reasons). Also, for me the tests fail with the following output, which however does not appear when running "just awesome" outside of the testing framework. I'll have to investigate.
```
2016-03-26 18:45:11 W: awesome: xerror:943: X error: request=(null) (major 139, minor 4), error=BadDrawable (9)
2016-03-26 18:45:11 W: awesome: xerror:943: X error: request=(null) (major 139, minor 8), error=(null) (143)
2016-03-26 18:45:11 W: awesome: xerror:943: X error: request=(null) (major 139, minor 7), error=(null) (143)
```

Edit: When/If this is merged, every config out there will break. You don't get any wibox, no wallpapers, nothing (well, popup menus on the root window should still work). That will be lots of fun!

### DIFF
--- a/lib/gears/screen.lua
+++ b/lib/gears/screen.lua
@@ -1,0 +1,29 @@
+---------------------------------------------------------------------------
+-- @author Uli Schlachter
+-- @copyright 2016 Uli Schlachter
+-- @release @AWESOME_VERSION@
+-- @classmod gears.screen
+---------------------------------------------------------------------------
+
+local screen = screen
+
+local module = {}
+
+--- Call a function for each existing and created-in-the-future screen.
+-- @tparam function func The function to call.
+function module.connect_for_each_screen(func)
+    for s in screen do
+        func(s)
+    end
+    screen.connect_signal("added", func)
+end
+
+--- Undo the effect of connect_for_each_screen.
+-- @tparam function func The function that should no longer be called.
+function module.disconnect_for_each_screen(func)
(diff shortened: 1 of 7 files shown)


lib/gears/screen.lua
```diff
@@ -0,0 +1,29 @@
...
+-- @classmod gears.screen
+---------------------------------------------------------------------------
+
+local screen = screen
+
+local module = {}
+
+--- Call a function for each existing and created-in-the-future screen.
+-- @tparam function func The function to call.
+function module.connect_for_each_screen(func)
```
Elv13: I am fine with this, but I wonder if `added` couldn't simply be sent for every screens when Awesome starts. It would avoid this helper function.

psychon: The plan is to not define any screens before loading the config, so yeah, your edit explains the reason for this. :-)
TODO: Merge this into awful.screen if no users in gears/wibox show up.

psychon:
Rebased and updated. Looking forward to the breakage that this produces...
I figured out the errors. They are unrelated to this. `wibox.drawable` uses the old, no-longer-valid wallpaper after a new one was set and thus the old one destroyed. I'll fix this up (with an API change, in another PR).

PR #781, opened 2016-03-26
Make matrices immutable again

psychon:


Elv13:
fine by me. I also dropped the other mutable methods from my `upstream_dynamic_complete` branch. So it should not break it anymore.

PR #782, opened 2016-03-27
Changes to wallpaper change

psychon:
The last commit in this series fixes the race condition that lead to the errors in https://github.com/awesomeWM/awesome/pull/780. The rest should make us perform slightly better since cairo can cache some information about the wallpaper. :-)

The first commit is completely unrelated to all of this and only part of this PR since valgrind noticed this problem while I was working on this.

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -694,6 +694,9 @@ main(int argc, char **argv)
     xcb_ungrab_server(globalconf.connection);
     xcb_flush(globalconf.connection);
 
+    /* get the current wallpaper, from now on we are informed when it changes */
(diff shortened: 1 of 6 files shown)


blueyed:
Related to https://github.com/awesomeWM/awesome/commit/b05b699129cd4cc7b488745deaa7461666a97bf3#commitcomment-16761333 also?

psychon:
If it helps: Yes, but no idea why.
If not: Nope
My guess would be "Nope", since this should only help with problems which result in awesome printing about xerrors on stderr.

PR #785, opened 2016-03-28
Geometry overhaul part 1

Elv13:
/me Ok, so here it is. This took a lot of work to prepare. It is also just the tip of the iceberg.

This pull request is the first (big) one in a series that will consolidate/rewrite geometry, placement and resizing code from:

 * `awful.menu`
 * `awful.tooltip`
 * `awful.wibox`
 * `awful.mouse`
 * `awful.layout.dynamic`
 * `ewmh`
 * `the dynamic layout splitter widget`
 * `elv13/radical`
 * `elv13/collision`
 * `client.c`
 * `drawin.c`
 * `wibox`
 * `naughty`

The goal is to create one coherent API able to handle placement and geometry across Awesome. While this create a new God class, it is necessary because:

 * This turn the awful.placement methods into a standardized API so the methods
   can be used as properties (like in `awful.rules` with a `placement` property or in `awful.wibox`/`naughty`). All methods have configuration arguments for common things like honor_workarea or screen padding.
 * Allow existing features for an object type (wibox, client, mouse) to be
    re-used by other "drawable" (loose definition also including the mouse) types.
 * Re-use existing code in other placement methods, creating a rich and flexible
   API in the process.
 * Have an unified test templates designed to test placement methods.
 * Reduce code duplication (like between awful.mouse and layout.dynamic)
 * Fix unhandled corner cases in a single places rather than once for tooltip,
    menu and client.
 * All methods can be used with a custom area, parent drawable or screen, making them very useful in a lot of (my) extensions.

This PR lay the ground work for the whole overhaul. Most placement methods are here. Some of the missing ones include "relative to parent widget", as used by `awful.menu`, `awful.tooltip` and `radical` and some `gears.matrix` integration. The next step for this is to remove all the code this PR deprecate.

This is in line with my dynamic client layout work. I found some limitations where I had to duplicate existing code (like in resize.lua, the splitting code and base.lua). There is also other work to use the `request::` API for size hints (not in this PR). Finally, as with the PR merged a few days ago that did the similar thing to `awful.screen`, this will lead to a new "signal sequence" that will replace the "god callback" `awful.layout.arrange()` by placing relevant code and events where they belong (like `tag.geometry`/`property::geometry`)

I know this is a scary one. It is not fully ready to be merged, there is a lot of typos left. However, I would like feedback as this has consequences in my other "active" projects like the property system and client layout rewrite. It rely on the currently unmerged testing framework.


lib/awful/placement.lua
```diff
@@ -29,6 +29,132 @@ end
...
+-- indices into a geometry point.
+-- @tparam table geo a geometry table
+-- @tparam number corner_i The horizontal matrix index
+-- @tparam number corner_j The vertical matrix index
+-- @tparam number n The (square) matrix dimension
+-- @treturn table A table with *x* and *y* keys
+local function rect_to_point(geo, corner_i, corner_j, n)
+    return {
+        x = geo.x + corner_i * math.floor(geo.width  / (n-1)),
+        y = geo.y + corner_j * math.floor(geo.height / (n-1)),
```
psychon: Why do you `floor` this before the multiplication? Shouldn't it be `floor(corner * width / (n-1))`?

lib/awful/placement.lua
```diff
@@ -29,6 +29,132 @@ end
...
+-- area to be smaller than the source one.
+-- @tparam table source The (larger) geometry to move `target` into
+-- @tparam table target The area to move into `source`
+-- @treturn table A table with *x* and *y* keys
+local function move_into_geometry(source, target)
+    local ret = {x = target.x, y = target.y}
+
+    -- Horizontally
+    if ret.x < source.x then
+        ret.x = source.x + 1
```
psychon: Why the `+ 1`? This is inconsistent. If `ret.x = source.x`, then nothing changes, if it's smaller, then `ret.x = source.x + 1`. Since I don't see the reason for the `+1`, I'd suggest to fix this by removing this part (the alternative being to change `<` into `<=`)
(The same applies to the other cases below)

tests/examples/awful/placement/closest_mouse.lua
```diff
@@ -0,0 +1,53 @@
+
+local awful = {placement = require("awful.placement")} --DOC_HIDE
+
+local c = client.gen_fake {x = 20, y = 20, width= 280, height=200, screen =screen[1]} --DOC_HIDE
+
+mouse.coords {x=100,y=100} --DOC_HIDE
+ -- Move the mouse to the closest corner of the focused client
+awful.placement.closest_corner(mouse, {include_sides=true, parent=c})
+mouse.push_history() --DOC_HIDE
```
psychon: - What's `push_history`?
- Shouldn't the `@foo@`-part of the previous commit only be added here?
- Wouldn't it be nicer to have "proper" unit tests in `spec/` than to have some code whose result isn't really looked at?

lib/awful/placement.lua
```diff
@@ -573,6 +573,58 @@ end
...
+        ngeo.width  = dgeo.width + (dgeo.x - ngeo.x) - 2*bw
+    elseif args.direction == "right" then
+        ngeo.width  = sgeo.width - ngeo.x - bw
+    elseif args.direction == "up" then
+        ngeo.y      = sgeo.y + bw
+        ngeo.height = dgeo.height + (dgeo.y - ngeo.y) - 2*bw
+    elseif args.direction == "down" then
+        ngeo.height = sgeo.height - dgeo.y - bw
+    else
+        assert(false)
```
psychon: How about `error("Unknown direction: " .. tostring(args.direction))`?

lib/awful/placement.lua
```diff
@@ -155,6 +159,89 @@ local function move_into_geometry(source, target)
...
+
+    d:connect_signal("property::width" , tracker)
+    d:connect_signal("property::height", tracker)
+
+    tracker()
+
+    if args.update_workarea then
+        local function tracker_struts()
+            --TODO this is too fragile and doesn't work with all methods.
+            wibox_update_strut(d, reverse_align_map[position_f])
```
psychon: `reverse_align_map` doesn't have any entries, this will always be `nil`...?

tests/examples/awful/placement/align.lua
```diff
@@ -0,0 +1,25 @@
...
+
+local c
+
+c = client.gen_fake {x = screen[2].geometry.x+10, y = screen[2].geometry.y+10, width=40, height=30, screen=screen[2]}
+placement.align(c,{position = "center_vertical", honor_workarea = true})
+
+c = client.gen_fake {x = screen[3].geometry.x+10, y = screen[3].geometry.y+10, width=40, height=30, screen=screen[3]}
+placement.align(c, {position = "center_horizontal", honor_workarea=true})
+
+return {hide_lines=true}
```
psychon: I thought this was supposed to be for docs, not for tests...

tests/examples/awful/placement/align.lua
```diff
@@ -0,0 +1,25 @@
+--DOC_HIDE_ALL
+local placement = require("awful.placement")
+screen[1]._resize {x= 50}
```
psychon: Huh?

psychon:
I'm a bit worried that this `examples` stuff goes a bit... far. Besides that, I wonder how the Travis build succeeded since the code here apparently uses methods that do not exist.
Besides that, fine by me.

tests/examples/awful/placement/align.lua
Elv13: The shims added by the other PR have some `_something` methods like those mostly for the screen stuff.

tests/examples/awful/placement/closest_mouse.lua
Elv13: > What's push_history?

The cursor have little lines between the points it visit. That method just create "groups" so unrelated mouse move are not linked. See the other PR, there is a PDF example of this (generated by this example)

> Shouldn't the @foo@-part of the previous commit only be added here?

Maybe, but is it worth it? The code is just eliminated if the variable id not found.

> Wouldn't it be nicer to have "proper" unit tests in spec/ than to have some code whose result isn't really looked at?

I will add some asserts to this one eventually. IMO why having tests in 2 places that are copy/pasted duplicate of each other when you can have only one. Most other example/tests have their asserts, I guess I have been lazy on this one. I will add them.

lib/awful/placement.lua
Elv13: Oops, I forgot a few lines when building this commit. The align part is not test-covered yet. I will have to write a rest `tests/test-*.lua` for this as my testing framework lack the code to emulate workarea (and it would be a PITA to implement)

Elv13:
> I wonder how the Travis build succeeded since the code here apparently uses methods that do not exist.

This depend on the other testing PR. It runs fine if you merge the other commits into this one. The shims have some extra methods to configure screens and add metadata to objects such as clients. Without these commits, the coverage decrease while it increase if you merge them.

tests/examples/awful/placement/align.lua
Elv13: This file only generate an image/screenshot, like this one. So it is for doc. If there was something that would only be relevant in `spec/` then I would place it there.

lib/awful/placement.lua
Elv13: i and j are integers, so it wont change anything

psychon:
Oh apropos code coverage: It occurred to me that code coverage actually breaks for any file where some `@foo@` expands to more than one line. The code coverage stuff records line numbers that were hit during execution. Coveralls looks up those lines in the files in git. When `@foo@` is expanded, the line numbers for the code afterwards changes.
Even worse, some tests (`spec/` and `tests/examples`, I think) run on the code before expansion, others (the rest of `test/`) run on the code after expansion. These results are then merged together (with some warnings that something seems wrong).

This problem is not specific to this PR and also exists without it. No idea what to do about it.

lib/awful/placement.lua
psychon: `2 * math.floor(1 / 2) = 0`, but `math.floor(2 * (1 / 2)) = 1`

Or, more generally, let width be an odd number: `2*math.floor(width / (3-1)) = 2*(width/2-0.5) = width-1 ~= width`

Elv13:
Rebased, the build still pass, but one of the worked crashed and I manager to get `[build|failed]` due to the same heisenbug again, yay!

Elv13:
so? anyone have something left to say or I merge this?

Again, this is part 1 or 3, there is probably some broken corner cases, but it has an "ok" coverage of what users might actually use. So at least most of it work. Part 2 and 3 are mostly removing the, now duplicated, code from other modules and add the few missing methods like all the "move relative to X" variants (the code already exist in `radical.placement`).

lib/awful/placement.lua
psychon: You decided against this? Ok

psychon:
I'll trust you on this: :+1:

lib/awful/placement.lua
Elv13: > You decided against this? Ok

The diff is outdated, this has been factored out of the final version. It remain in the first commit to avoid breaking the build, but later commits introduce `.align` and I use that instead and remove this method. There is a dependency loop between those 2 commits, I could have worked harder to fix it, but meh, in the end it works.

actionless:
now the build fails for me:
```
-- Testing code based on _awful

2016-04-03 19:38:16 W: Cannot reliably detect EOF on an GIOInputStream with this LGI version

CMake Error at tests/examples/CMakeLists.txt:135 (message):

  /home/lie/build/awesome-git/src/awesome-git/tests/examples/awful/placement/maximize_vertically.lua
  A test failed, bye
  tests/examples/CMakeLists.txt:216 (run_test)
  tests/examples/CMakeLists.txt:208 (digg)
  tests/examples/CMakeLists.txt:208 (digg)
  tests/examples/CMakeLists.txt:237 (digg)
  awesomeConfig.cmake:306 (include)
  CMakeLists.txt:17 (include)


```

Elv13:
Can you add more lines of the build log and some print in the test file? I
can't reproduce (gentoo), nor does travis (Ubuntu). I will disable that one
for now, but I need to know why it fails for you. Thanks for reporting this

actionless:
can test be marked as failed because of that warning?
`2016-04-03 21:23:36 W: Cannot reliably detect EOF on an GIOInputStream with this LGI version`

```lua
print("BEGIN")
print(c.height+2*c.border_width)
print(screen[1].geometry.height)
assert(c.height+2*c.border_width == screen[1].geometry.height)--DOC_HIDE
print(c.x)
assert(c.x == 45) --DOC_HIDE
print(c.width)
assert(c.width == 40) --DOC_HIDE
print("END")
```

```
BEGIN
96
96
45
40
END
```

Elv13:
@actionless oh, I missed that from the first post, sorry

So, it is https://github.com/awesomeWM/awesome/pull/787 that "broke" the build because both PR were developed in parallel and made different assumptions. Last week new test framework assumed only CAPI normally print on stderr and #787 assume printing warnings on stderr has no consequences. Both assumptions were true when the code was published, but they no longer are.

Please test #792, I will merge it soon to avoid more user reports. Its hackish, but something has to be done.

actionless:
yup, soft warning helped

PR #787, opened 2016-03-28
awful.spawn: Reliably detect end-of-file

psychon:
Now that https://github.com/pavouk/lgi/issues/125 is fixed, my old hack in `awful.spawn` to detect end-of-file will no longer work. This PR adds some auto-detection-magic to fix awesome. While loading `awful.spawn` this code does a test to figure out how to detect EOF:
- On newest LGI, this checks for a `nil`. This was tested here: https://travis-ci.org/awesomeWM/awesome/builds/118974100 (The build tests this PR + another commit making it use LGI git)
- On LGI as shipped by debian testing, this uses a hack that considers any empty line the end of the input. This is better than nothing. This was tested on my laptop (I've been patching awful.spawn for ages)
- On older LGI, the same check as we did before is done. This is tested by Travis for this PR.

### DIFF
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -29,6 +29,45 @@ local protected_call = require("gears.protected_call")
 local spawn = {}
 
 
+local end_of_file
+do
+    -- API changes, bug fixes and lots of fun. Figure out how a EOF is signalled.
+    local input
+    if not pcall(function()
+        -- No idea when this API changed, but some versions expect a string,
+        -- others a table with some special(?) entries
+        input = Gio.DataInputStream.new(Gio.MemoryInputStream.new_from_data(""))
+    end) then
+        input = Gio.DataInputStream.new(Gio.MemoryInputStream.new_from_data({}))
+    end
+    local line, length = input:read_line()
+    if not line then
+        -- Fixed in 2016: NULL on the C side is transformed to nil in Lua
+        end_of_file = function(arg)
+            return not arg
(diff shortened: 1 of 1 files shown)


psychon:
No idea why the build fails with old LGI versions. Can we bump our minimum required LGI version? :-/

blueyed:
@psychon / @Elv13 
This causes the build/tests to fail with me now - probably because something gets printed to stderr?!

```
-- Testing code based on _awful

2016-04-03 19:01:25 W: Cannot reliably detect EOF on an GIOInputStream with this LGI version

CMake Error at tests/examples/CMakeLists.txt:135 (message):
  
  …/awesome-git/tests/examples/awful/placement/stretch_left.lua
  A test failed, bye
  tests/examples/CMakeLists.txt:216 (run_test)
  tests/examples/CMakeLists.txt:208 (digg)
  tests/examples/CMakeLists.txt:208 (digg)
  tests/examples/CMakeLists.txt:237 (digg)
  awesomeConfig.cmake:306 (include)
  CMakeLists.txt:17 (include)
```

I am using lua-lgi 0.9.0-1 on Arch Linux.

blueyed:
Jup: https://github.com/awesomeWM/awesome/blob/3e5b1b3ba9b9f91f3ec78c1f547adb405c10bb6a/tests/examples/CMakeLists.txt#L131-L136

@Elv13 
This should filter out any `W:` lines probably?

PR #789, opened 2016-03-30
fixes to keyboard layout names

aroig:
These two commits do the following:

* fix pattern expression to handle layout names like `us(altgr-intl)`,
  with a `-`in them.

* extract the keyboard layout naming into a function, that can be easily
  overriden. For instance, one may want to show the layout name without
  the variant.

### DIFF
--- a/lib/awful/widget/keyboardlayout.lua
+++ b/lib/awful/widget/keyboardlayout.lua
@@ -241,10 +241,7 @@ local function update_layout(self)
         layouts[1].group_idx = 0
     end
     for _, v in ipairs(layouts) do
-        local layout_name = v.file
-        if v.section ~= nil then
-            layout_name = layout_name .. "(" .. v.section .. ")"
-        end
+        local layout_name = self.layout_name(v)
         -- Please note that numbers of groups reported by xkb_get_group_names
         -- is greater by one than the real group number.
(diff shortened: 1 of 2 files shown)


psychon:
Fine by me. CC @nuko8 What do you think? (Yup, you touched the code last and now I will CC you on every PR touching this :-) )

psychon:
One more thing: Could you add suitable tests to `spec/awful/keyboardlayout_spec.lua`? Something that was broken before and works now?

aroig:
Ok. How should I run these tests?

psychon:
The tests are automatically run by `make` and can be explicitly run via `make check-unit`. However, you need to have busted installed to do so ( http://olivinelabs.com/busted/ ) (no idea if CMake picks up busted if you install it after having build awesome, perhaps `make distclean && make` to force a full rebuild?)

aroig:
Ok, I have pushed a test for it, but have not run it. Does travis take care of that?

I'll install busted for next time...

psychon:
Yup. Travis runs the tests and it looks like the new one didn't fail. Thanks!

nuko8:
@psychon 

I tried the patch.  Built and installed successfully and no error on the tests.  Regression not found. :+1:

psychon:
Thanks

PR #790, opened 2016-03-31
Client, tag and screen API cleanup.

Elv13:
Behold, this PR start the most massive API changes in years... (but with proper deprecation, so no API break ;) )

Ok, here are the first few commits associated with this. It mostly improve the `client` API documentation page. `awful.client` now have a link to `client`. Both API are unified into documentation "chapters". This will be less confusing that 2 files.

I also add many new `ldoc` tags for `@property` `@beautiful` and `@deprecated`.

The generated document is still a mess because I am not done putting everything into their sections.


Elv13:
It is still a WIP, but I am not "mostly" done with marging `awful.client` and `capi.client` into a single API. The generated `ldoc` apidoc is getting quite nice. Net yet where I want it to be, but getting there.

Some commits will be squashed/reworked/re-ordered, the code is still a little bit messy after all the copy/paste/find&replace and I probably managed to introduce some regressions (move config seem to run once all the deprecated messages are gone... Also, I have to fix all the those luacheck problems and write a few new regression tests (not using the new framework, real integration tests)

I would like some feedbacks on my work before I do the other 2. This is really, really long and boring to do, but I think it is worth it and should have been done loooong time ago.

objects/client.c
```diff
@@ -2854,135 +3371,135 @@ client_class_setup(lua_State *L)
      * @signal mouse::move
      */
     signal_add(&client_class.signals, "mouse::move");
-    /**
+    /*
```
psychon: I only looked at the "total diff", not the individual commits and have to say: Thanks for working on this.

The only small remark that I have is on this part here. This comment exists just for ldoc. If you "un-ldoc" it, it can just be removed instead.

Elv13: > The only small remark that I have is on this part here. This comment exists just for ldoc. If you > "un-ldoc" it, it can just be removed instead.
Done

Elv13:
So, this is more like it. My config seem to run unmodified, but this isn't a good indicator as its just way too different. It still need the tests (I will write them once all classes are done). I think the commit log is nailed, there is no weird commit dependency loops anymore. I would appreciate a more in depth review.

This is still WIP and it wont be merged as is, but its kind of coherent right now, so it is good time for testing. Typo hunt is welcome too, not exactly my strength. For the PR size, most of it is actually the 1.3k line commits that rewrite the `capi.client` API doc, the "real" changes are not very large.

lib/awful/client.lua
```diff
@@ -439,72 +439,90 @@ function client.object.move_to_screen(self, s)
...
-    end
+function client.object.get_is_marked(self)
+    return client.property.get(self, "marked")
 end
 
+--- Mark a client, and then call 'marked' hook.
+-- @deprecated awful.client.mark
+-- @client c The client to mark, the focused one if not specified.
+function client.mark(c)
+    util.deprecate "Use c.is_marked = true instead of awful.client.unmark"
```
ff2000: Should be "awful.client.mark" instead of unmark.

Elv13: Fixed, thanks

Elv13:
Arg... Travis is killing me... Since yesterday most builds just lock and timeout... (note: This PR is WIP and not expected to pass tests anyway, but it's beside the point)

psychon:
Since travis builds for master still work fine: Could this be an error in your branch? E.g. the added call to `screen_scan()` in awesome.c looks fishy. Does awesome build from your branch work locally for you?

Elv13:
> E.g. the added call to screen_scan() in awesome.c looks fishy.

Woops, your commit revert and my temporary patch collided. But that isn't the issue as it doesn't prevent the tests from running locally and Travis did succeed once in a while, so it's something else.

Elv13:
I found yesterdays problem. There was in infinite loop on travis that somehow did not trigger locally (it depended on the call order when loading the tests, if `t.index` was first called by `gettags()`, the stack would pop. The other way around was fine).

However I wrote more test and now I have another problem. Apparently `c.ontop` is not very reliable. The test pass 10% of the time, but 90% of the time it fails. @psychon Is there some magic XCB async spell going on here? The problem happen both locally and on travis. As it is unrelated to my changes, I can disable that test, but I could like to know why it fails that often (I tried adding some sleeps here and there, it doesn't seem to affect the outcome).

psychon:
Got some (small, self-contained) code for me to look at? Perhaps even something I can run on current master? I don't know about any magic going on with `c.ontop`, so I'd be tempted to "blame" the test.

Elv13:
> Got some (small, self-contained) code for me to look at?

Take `test-awesomerc.lua` and delete the sections after the second one (or not), place it in the `tests/` folder and run the script

psychon:
`Error: running function for step 1/6 (@2): [string "local awful = require("awful")..."]:51: attempt to index field 'selected_tag' (a nil value)`

Elv13:
Umm, oh, right, sorry replace that by `awful.tag.selected()`, I will push a temporary version using the deprecated names.

psychon:
Ok, how about "there is a race". You don't give awesome a chance to wait for the key press to be handled. The timer from `tests/_runner.lua` fires every 0.1 seconds and if awesome is slow enough, it might fire before your code runs.

I would add the following to the second function: `if not old_c.ontop then return false end`
That way, the function will be re-called again later.

Does that help?

Elv13:
```
local glib  = require("lgi").GLib

-- luacheck: globals modkey

local function get_callback(mod, key)
    local inf = {}
    awful.key(mod, key, nil, nil, inf)

    return inf.execute
end

-- Get a tag and a client
local function get_c_and_t()
    client.focus = client.get()[1]
    local c  = client.focus

    return c
end

-- display deprecated warnings
awful.util.deprecate = function() end

local has_spawned = false
local old_c = nil

local steps = {

function(count)
    if count <= 1 and not has_spawned and #client.get() < 2 then
        awful.spawn("xterm")
        awful.spawn("xterm")
        awful.spawn("xterm")
        awful.spawn("xterm")
        awful.spawn("xterm")
        has_spawned = true
    elseif #client.get() >= 2 then
        local c, _ = get_c_and_t()
        old_c = c

        assert(c)
        glib.usleep(130000)

        awful.layout.inc(1)

        assert(not c.ontop)
        get_callback({modkey}, "t")()

        return true
    end
end,

-- Ok, no now ontop should be true
function()
    get_c_and_t()

    -- I don't know why, but this fail once in a while --FIXME
    -- The callback **is** called.
    -- It happen both locally and on travis
    glib.usleep(130000)
    assert(old_c.ontop)

    return true
end
}

require("_runner").run_steps(steps)
```

psychon:
If you want to play error message ping-pong: `Error: running function for step 1/2 (@2): [string "local awful = require("awful")..."]:47: attempt to call a nil value` (I don't have `awful.key.execute` and this one might be harder to work around than the previous error)

Elv13:
umm, it also require the first commit, I forgot. Travis is building, let see if the tests pass (or I will play ping pong with travis until it either fail or pass on that error again)

psychon:
Sorry, I remembered the "API" for `tests/_runner.lua` correctly. It should be `return`, not `return false`.

Quick recap:
- `true`: Function succeeded, continue to next function
- `false`: Function failed, fail
- nothing / `nil`: Call this function again later

psychon:
Hm :-(

Elv13:
Still happen

```
../tests/run.sh                                                        ⮂Fri 03:44 ⮀━┛
awesome_log: /tmp/tmp.dm6nsSquJ7/_awesome_test.log
_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root
== Running /home/lepagee/dev/awesome/tests/test-awesomerc.lua ==
2016-04-08 03:48:25 W: Cannot reliably detect EOF on an GIOInputStream with this LGI version
urxvt: Loading image icon failed, continuing without.
urxvt: Loading image icon failed, continuing without.
urxvt: Loading image icon failed, continuing without.
urxvt: Loading image icon failed, continuing without.
urxvt: Loading image icon failed, continuing without.
2016-04-08 03:48:26 W: awesome: a_glib_poll:373: Last main loop iteration took 0.157336 seconds! Increasing limit for this warning to that value.
Test finished successfully
== Running /home/lepagee/dev/awesome/tests/test-awful-client.lua ==
2016-04-08 03:48:27 W: Cannot reliably detect EOF on an GIOInputStream with this LGI version
urxvt: Loading image icon failed, continuing without.
urxvt: Loading image icon failed, continuing without.
Test finished successfully
^C%                                                                                       ┏━ /  ⮀ home ⮀ lepagee ⮀ dev ⮀ awesome ⮀ build ⮀───────────────────────────⮂  client_properties ⮀━┓ 
┗━: ../tests/run.sh                                                        ⮂Fri 03:48 ⮀━┛
_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root
awesome_log: /tmp/tmp.wELyXCTmSN/_awesome_test.log
== Running /home/lepagee/dev/awesome/tests/test-awesomerc.lua ==
2016-04-08 03:48:30 W: Cannot reliably detect EOF on an GIOInputStream with this LGI version
urxvt: Loading image icon failed, continuing without.
urxvt: Loading image icon failed, continuing without.
urxvt: Loading image icon failed, continuing without.
urxvt: urxvt: Loading image icon failed, continuing without.
Loading image icon failed, continuing without.
2016-04-08 03:48:30 W: awesome: a_glib_poll:373: Last main loop iteration took 0.247660 seconds! Increasing limit for this warning to that value.
Error: timeout waiting for signal in step 2/6 (@5).
===> ERROR running /home/lepagee/dev/awesome/tests/test-awesomerc.lua! <===
Error: timeout waiting for signal in step 2/6 (@5).
== Running /home/lepagee/dev/awesome/tests/test-awful-client.lua ==
2016-04-08 03:48:32 W: Cannot reliably detect EOF on an GIOInputStream with this LGI version
urxvt: Loading image icon failed, continuing without.
urxvt: Loading image icon failed, continuing without.
Test finished successfully
== Running /home/lepagee/dev/awesome/tests/test-awful-screen.lua ==
```

(from my local workstation, Travis passed a couple of times too, it's just rare)

Elv13:
I cannot seem to reproduce it using my "real" setup, mod4+t seem to be reliable. So, do we call that a mystery, disable it and move on?

psychon:
I added some prints. The callback for `mod-t` is called on another client than `old_c` (just `print(c)` / `print(old_c)` is enough to see this).

And the reason for that... I'm not really sure.

psychon:
Uhhhhh! The C code lazily bans windows and it lazily updates the focus. However, keys are faked non-lazily. So what happens is:

- "Some" `xterm` has the input focus.
- Your code runs, it focuses the "target xterm" (only internally in awesome)
- Also, it fakes some input (this is sent immediately)
- Then, awesome actually handles the tag switch and all clients are hidden
- The X server reports the key press event to awesome for the client which had the input focus "back then"
- Some random client becomes ontop and not the one that your code expects.

Elv13:
Make sense, I guess... Well too bad, being able to test all awesomerc shortcuts would have been nice to catch "grave" regressions. But given neither "space" or anything using client.focus works, it will never be reliable...

psychon:
...and (sorry for the comment spam) I even have a fix for you:
```
function(count)
    if count == 1 then
        assert(#client.get() == 0)
        for i=1, 5 do awful.spawn("xterm") end
    elseif #client.get() == 5 then
        if not old_c then
            old_c = get_c_and_t()
            assert(old_c)
            awful.layout.inc(1)
            assert(not old_c.ontop)
    else
            assert(not old_c.ontop)
            get_callback({modkey}, "t")()
           return true
    end
end,
function()
    if old_c.ontop then return true end
end
```

Elv13:
Lets hope for the best, it seem to be more reliable locally. So there is (hopefully?) only that "space" problem left.

Elv13:
Victory! (and 2.6% more coverage, so far (and yes, it seem to treat tests as part of the ratio, so it's not "really" 2.6%", but whatever))

@psychon So, there is probably mistakes left and deprecated calls left, but this is about it. Feel free to start doing reviews. I gave you access to my fork ( @blueyed already had it). So, for the typos, you can fix them inline using github web editor. The generated ldoc is really an improvement over what was there before. The only regression I caused on that from is that the return values of deprecated methods are not displayed, I have yet to figure out why. I will write more tests and try to cover more of the code I changed before this get merged. It is very large and very error prone, but not very "scary". The logic is (mostly) the same, I just moved it around and implemented a single naming convention in the process.

Elv13:
@psychon I added yet more coverage. However I hit some issues not related to my code (or even Awesome itself). One of them seem to be related to the titlebars having a negative size. The others seem to be related to xterm size being set of bad values by the layouts. I am not really sure what to do about them. Is that Awesome layouts job to make sure they don't set invalid sizes? Is it the C core job? Or do we just don't care and I find a way to mute those errors?

psychon:
> One of them seem to be related to the titlebars having a negative size. The others seem to be related to xterm size being set of bad values by the layouts. I am not really sure what to do about them. Is that Awesome layouts job to make sure they don't set invalid sizes? Is it the C core job? Or do we just don't care and I find a way to mute those errors?

Wait, what? All of those should be impossible. Titlebar resizes to negative sizes and client resizes to non-positive sizes are silently ignored by the C code (if the C code wouldn't ignore them, the X server would complain).
So... is this your "mock" of the C api that allows these negative sizes? Is this the issue that you are running in?

Elv13:
There is no mock here, its xephyr. They may not be negative, but they are
invalid. You can fetch the PR and run the tests locally. You will see all
the xephyr bad requests warnings. This is because of corner cases like zero
width mwfact that are being drawn.

Elv13:
@blueyed, @actionless Hello, I would like your opinion on this. Can you build the doc locally and take a look? There is many naming decision to double check. There is some incoherence such as `tag.mwfact` vs. `tag.master_fill_policy`. I avoided breaking the API, so I left them there, but I wonder if we should come up with a strict naming guideline and just stick it it. Another issue is about the boolean properties. Should they follow Qt style and always have an "is_" prefix like some of them already have or should we just use the property name like `floating`? Again, should we be strict on that or avoid breaking the API and letting them stay as-is?

blueyed:
> There is some incoherence such as tag.mwfact vs. tag.master_fill_policy

How about using a new name and keeping the other as a deprecated alias then? (which could be dropped in 3.7 or 3.8 later?!)

Elv13:
> How about using a new name and keeping the other as a deprecated alias then? (which could be dropped in 3.7 or 3.8 later?!)

I am very open to that, I only want to have a good discussion before doing anything. I would also vote for `is_`.

Elv13:
[tag.pdf](https://github.com/awesomeWM/awesome/files/211657/tag.pdf)

Elv13:
```
[03:56] <Elv13> psychon:  also, if you could give your opinion on the naming convention blueyed seem to prefer is_something for boolean properties, like Qt naming guidelines recommands. I don't mind one way or another, as long as we decide a naming convention and stick to it from now on.
[03:57] <psychon> so .sticky becomes .is_sticky? .ontop becomes .is_ontop? etc?
[04:00] <Elv13> psychon: right, I am personally not convinced it is worth breaking the API for that. At the same time, there is already some is_something, so it is inconsistent
[04:00] <Elv13> inconsistent APIs suck
[04:01] <kjetilho> there should be a transition period where both function names work, IMHO
[04:01] <kjetilho> (and that is preferable to waiting until 4.0 to do a sudden switch-over)
[04:02] <psychon> we don't have that many is_foo APIs, do we? wouldn't switching in the other direction be easier?
[04:02] <psychon> also, those rules should be added to something... don't we have some coding style document somewhere?
[04:03] <Elv13> kjetilho: I made sure there is proper deprecations warnings, but I did not, and doesn't plan to break the API.
[04:04] <Elv13> psychon: Given how the API was, no, there is no conventions at all. Sometime getters and setters didn't even have the same one...
[04:04] <Elv13> :getidx() -> property::index -> :move(new_idx)
[04:05] <Elv13> getmfpol -> property::master_fill_policy
[04:05] <Elv13> tag.property.get(), client.getproperty()
[04:06] <psychon> awful.tag has too many short names (mfpol, mwfact etc)
[04:08] <Elv13> that's another problem. Should I move all to the full name? mwfact -> master_width_factor? Given master_fill_policy already use the full name, it would make sense
[04:08] <Elv13> nmaster -> master_count, ncol -> column_count, mwfact -> master_width_factor
[04:09] <psychon> I think so, yeah
[04:11] <Elv13> psychon: If you don't want to build it yourself, here is the result https://github.com/awesomeWM/awesome/files/211657/tag.pdf
[04:13] <-- j605 (~j605@193.190.253.150) has left this server (Quit: WeeChat 1.4).
[04:14] <psychon> up to now, I was reading RSS feeds and mails
[04:14] <psychon> now it's time to pull your branch, take a look at its diff from master and... see
[04:15] <psychon> also, please make sure that 3.7 happens not too far after 3.6, we have way too many changes since 3.5 :-(
[04:15] <psychon> ("make sure" = push me more for releases)
[04:17] <psychon> Elv13: default config: you replace awful.tag.viewonly with awful.tag.object.view_only
[04:17] <psychon> I thought the .object was supposed to be an implementation detail that shouldn't really be used directly?
[04:18] <Elv13> I can change that
[04:18] <psychon> if so, wouldn't it be more correct to change this code to function(t) t:view_only() end?
[04:18] <psychon> or did I get something wrong?
[04:18] <Elv13> I was lazy, sorry
[04:18] <psychon> no problem, this is mostly for my understanding
[04:19] <Elv13> expect a few find&replace in your config at first, but after that it should kind of run. If it doesn't, then there is a bug somewhere. I am quite sure you will find at least one, but I spent the last 3 days testing it, so I start to feel confident about it. All find&replace are documented in the deprecation warning
[04:20] <psychon> I thought you didn't break the API? if so, my config shouldn't need any changes at all
[04:20] <Elv13> It will work, but with 200 deprecation warnings
[04:20] <psychon> I don't look at deprecation warnings anyway, my config is even still using timer instead of gears.timer
[04:21] <psychon> (back then there was a reason of "testing old commits" and "running 3.5 for testing", but those reasons mostly disappeared and now I'm just lazy)
[04:25] <psychon> duh, reading that diff is boring
[04:26] <Elv13> Imagine writing it, it took me a full week of absolute boringness, no fun at all, just copy, paste, rename, test, repeat
[04:27] <psychon> "Honor size hints, i.e. respect size ratio." <- may I change that to "e.g."? size hints are more than aspect ratio
[04:27] <psychon> and yeah, it must have been bad
[04:28] <psychon> thanks for writing that, I can see that this is big leap for the docs
[04:28] <Elv13> it was my only motivation
[04:29] <psychon>  gears.surface(gears.surface(c.content)) <- huh? why gears.surface twice?
[04:29] <Elv13> oops
[04:29] <psychon> ok
[04:30] <psychon> also, AFAIK c.snapshot is not a screenshot, it's a life reference to the window's content...
[04:30] <Elv13> but cannot use used that way, so it always end up as screenshots, right?
[04:31] <Elv13> unless you have a magic indirection +resize  widget I never heard of
[04:31] <psychon> if you do this: local s = gears.surface(c.content) gears.timer.start_new(30, function() s:write_to_png("s.png") end)
[04:31] <psychon> this will get you the content of the client in 10 seconds
[04:31] <psychon> but yeah, client resizes will actually break it
[04:32] <Elv13> ;)
[04:32] <psychon> it's just that I avoid doing a copy of the client's content in the C code since 99% of users won't need it...
[04:32] <psychon> hm.... gears.surface.widget_to_surface() doesn't really belong there... it doesn't have anything to do with the rest of gears.surface :-(
[04:33] <Elv13> Well, your right this should be documented and another example (with copy)
[04:33] <psychon> some day(tm) I should move some stuff around in gears
[04:33] <psychon> I'm working on that
[04:33] <Elv13> feel free to do it. The decision making process should first be documented and reviewed
[04:34] <Elv13> just so I don't _ever_ have to do a PR like this one again
[04:34] <psychon> the process for "what belongs in one module"? my thinking goes like this: awful is an ugly, entangled mess where everything needs everything, so for gears, split things up a lot
[04:35] <psychon> heh, yeah, the naming should also be documented :-)
[04:36] <Elv13> While your here, what do you think about moving background, margin, mirror and constraint into wibox.container instead of half of them in wibox.layout and the other half in wibox.widget? (also moving graph and progressbar into wibox, as they don't depend on awful)
[04:37] <psychon> having a wibox.container base class for widgets sounds a lot better than using fixed as base class ;-)
[04:38] <psychon> but yeah, the naming seems fine to me as well
[04:38] <Elv13> as for the awful entangled mess, I cleaned things up quite a bit. There is a lot less dependencies and I split awful.client as it was unmaintainable.
[04:38] <psychon> perhaps we should have something like awful.util.deprecate_moved that can be used like: wibox.widget.background = awful.util.deprecate_moved("awful.container.background", awful.container.background)
[04:39] <psychon> that would require less code than the current version of this
[04:39] <Elv13> that't the plan
[04:39] <Elv13> that's*
[04:40] <Elv13> And also hit ldoc with a hammer until it allows me to add a "widget" section in the left panel
[04:40] <psychon> (so something like: function deprecate_moved(old_name, new_name, new_function) return function(...) awful.util.deprecate(old_name .. " was moved to " .. new_name) return new_function(...) end end
[04:40] <psychon> heh :-)
[04:40] <psychon> # Do not put too much load on this, it will deadlock <- huh?
[04:41] <psychon> wait_until_success can deadlock? or is it awesome-client that can deadlock? how?
[04:41] <Elv13> no idea, sync isn't enough. The tests wont run on my laptop unless I add that sleep
[04:41] <Elv13> they never fail with the sleep, they never work without it
[04:42] <psychon> hm, I'd like to have that one figured out
[04:42] <psychon> first step might be to look at pstree to figure out which commands hang
[04:43] <Elv13> too late for today
[04:43] <psychon> anyway, for the PR: I think that this is unreviewable big; I will just check if my config still works and then hope for the best
[04:43] <Elv13> It will require a typo hunt
[04:43] <psychon> I only looked at some of it and collected this diff: http://sprunge.us/NEHS?diff (and yes, I feel bad just changing your git repo)
[04:45] <Elv13> ok, bye
[04:45] <psychon> lots of deprecation warnings, but from a quick check it seems like my config still works
```

psychon:
I think that this is unreviewable much. I looked at some random parts of the diff between master and this and came up with the following changes for you: http://sprunge.us/NEHS?diff
Looking at the docs (randomly...), I spotted that in the client docs, the docs for awful.client.focus.history and awful.client.urgent are at the very bottom and not together with the other functions from awful.client (right at the top). Also, the urgent ones don't have the "awful.client" prefix in there.

For how to proceed, I'm fine with just merging this (assuming that the commits are sane and not a mess of WIP-stuff). I guess I will still take a look at the individual commits when I have the time, but no idea when I do that. This PR will too easily conflict with other changes otherwise, and I trust you to have done nice work. Small errors can still be corrected later.

Elv13:
> [04:08] <Elv13> that's another problem. Should I move all to the full name? mwfact -> master_width_factor? Given master_fill_policy already use the full name, it would make sense
> [04:08] <Elv13> nmaster -> master_count, ncol -> column_count, mwfact -> master_width_factor
> [04:09] <psychon> I think so, yeah

Done

> http://sprunge.us/NEHS?diff

Done


> Looking at the docs (randomly...), I spotted that in the client docs, the docs for
> awful.client.focus.history and awful.client.urgent are at the very bottom and not
> together with the other functions from awful.client (right at the top).

Blame ldoc. I can force ldoc to merge the functions into the top section, but I can't set it as a sub module at the same time. ldoc put submodules are the end and doesn't seem to let me override that. If you really want to "squash" the submodules into the function section, I can do that, but I don't think it would make it more readable.

> Also, the urgent ones don't have the "awful.client" prefix in there.

Done

> Small errors can still be corrected later.

I expect a few to appear, however I went to great length to test and review the thing. As you said, it is too large to do that by hand and not expect to miss something. Technically, this PR can be cut in 5, all commits are "complete" and only depends on previous ones, but all it would do is "break" users configs 5 time in a row.

@blueyed I did not use `is_` as it would have caused too many deprecations. See the IRC log above.

So, everybody, speak now or remain silent forever. I will merge this tomorrow to avoid the otherwise unavoidable rebase conflicts with other people work.

lib/awful/client.lua
```diff
@@ -599,6 +599,8 @@ function client.mark(c)
     end
 end
 
+
+
 --- Unmark a client and then call 'unmarked' hook.
```
psychon: Adding empty lines in the middle of nowhere? (Well, between "mark" and "unmark", which even belong together, so this isn't some kind of separator)

(Yup, I decided to look through these commits)

docs/aliases/awful_client.lua
```diff
@@ -0,0 +1,6 @@
+---------------------------------------------------------------------------
+--- This module is deprecated, use `client`
+--  ===============================
+--
+-- @module awful.client
+---------------------------------------------------------------------------
```
psychon: Not that it is important (feel free to ignore this comment), but this commit could also add that page to the ldoc config.

lib/awful/client/focus.lua
```diff
@@ -0,0 +1,230 @@
...
+-- @client c The client that has been focused.
+-- @function awful.client.focus.history.add
+function focus.history.add(c)
+    -- Remove the client if its in stack
+    focus.history.delete(c)
+    -- Record the client has latest focused
+    table.insert(internal, 1, c)
+end
+
+-- Copy of awful.client.visible...
```
psychon: Why do you copy the method? Was this added before this code `require()`d `awful.client`?

Elv13: Right, I will change this. I tried hard to avoid require("awful.client") dep loop, then I gave up because the code is a spaghetti ball. I tried to improve/fix/mitigate your complains about awful code. It isn't a total success, but at least it make awful.client smaller...

lib/awful/client.lua
```diff
@@ -999,11 +999,22 @@ function client.is_transient_for(c, c2)
...
 capi.client.add_signal("property::dockable")
+
+--- The client marked signal.
+-- @signal .marked
 capi.client.add_signal("marked")
-capi.client.add_signal("unmarked")
+
+--- The client unmarked signal.
+-- @signal unmarked
+capi.client.add_signal(".unmarked")
```
psychon: I don't think you wanted to add the `.` here. Perhaps better the line before?
(Also, why the heck do we have marked and unmarked and not just property::marked...)

lib/awful/client.lua
```diff
@@ -540,9 +540,30 @@ end)
...
+-- @deprecated awful.client.isfixed
+-- @see is_fixed
+-- @see size_hints_honor
 function client.isfixed(c)
+    util.deprecate "Use c.is_fixed instead of awful.client.isfixed"
     c = c or capi.client.focus
+    return client.object.is_fixed(c)
+end
+
+--- Return if a client has a fixe size or not.
```
psychon: "has a fixe`d` size" (missing d)

docs/config.ld
```diff
@@ -28,8 +28,19 @@ tparam_alias('client', 'client.object')
...
 tparam_alias('tab', 'table')
+
+-- Hack to get the functions and method on top of the signals and properties
+new_type("function", "Functions")
+-- Documentation for objects properties
+new_type("property", "Object properties", false, "Type")
 -- New type for signals
 new_type("signal", "Signals", false, "Arguments")
+-- Allow objects to define a set of beautiful properties affecting them
+new_type("beautiful", "Theme variables", false, "Type")
```
psychon: I think this commit should be reordered to a earlier position, some commit before this already used `@beautiful` and `@property`

lib/awful/client.lua
Elv13: Ah, I missed one. At first I remove all the leading points form the doc, then figured out that ldoc suck and it was why they were there. Then added them back using sed, but forgot to re-re-fix this.

> (Also, why the heck do we have marked and unmarked and not just property::marked...)

the first 2 are deprecated and documented as such. At some point we need to add add_signal_deprecated to that all `connect_signal()` print a warning.

tests/run.sh
```diff
@@ -169,6 +169,9 @@ start_awesome() {
     awesome_pid=$!
     cd - >/dev/null
 
+    # Do not put too much load on this, it will deadlock
+    sleep 0.5
```
psychon: Me being me, I would add a comment "TODO: Figure out what the problem is and fix it properly". But I guess we can also live with a sleep....

Elv13: I plan to remove this commit from the final version and do a PR. For now, I still need it, otherwise I can't run the tests and have to wait 7 minutes for Travis.

lib/awful/client.lua
```diff
@@ -439,72 +439,90 @@ function client.object.move_to_screen(self, s)
...
+        c.marked = not c.marked
     end
 end
 
 --- Return the marked clients and empty the marked table.
+-- @function awful.client.getmarked
 -- @return A table with all marked clients.
 function client.getmarked()
     for _, v in pairs(client.data.marked) do
+        client.property.set(v, "marked", false)
```
psychon: I don't think this works. `pairs()` remembers the key that it returned last (1), then you remove that entry. Since this is in the array-part of the table, everything is shifted down. So what was previously entry 2 is now entry 1. Now, `pairs` looks at the next entry after 1 and the old-entry-2-now-entry-1 is missed.
(I haven't actually tested this, it's just a theory)

Also, the return value of this function will be wrong since by the time it gets `client.data.marked`, that table will already be empty.

objects/screen.c
```diff
@@ -21,6 +21,22 @@
 
 /** awesome screen API
  *
+ * Screen objects can be added and removed over time. To get a reliable callback,
+ * use `awful.screen.connect_for_each_screen`:
```
psychon: I would word this differently. Something like "To get a callback for all current and future screens", or something like that. If the user just wants the future screens, they can use `screen.connect_signal("added", f)`.

lib/awful/screen.lua
```diff
@@ -45,13 +45,25 @@ local function apply_geometry_ajustments(geo, delta)
...
 --- Get the square distance between a `screen` and a point
+-- @deprecated awful.screen.getdistance_sq
 -- @param s Screen
 -- @param x X coordinate of point
 -- @param y Y coordinate of point
 -- @return The squared distance of the screen to the provided point
 function screen.getdistance_sq(s, x, y)
-    s = get_screen(s)
-    local geom = s.geometry
+    util.deprecate "Use s:get_suqare_distance(x, y) instead of awful.screen.getdistance_sq"
```
psychon: Typo: `s:get_square_distance`

lib/awful/screen.lua
```diff
@@ -45,13 +45,25 @@ local function apply_geometry_ajustments(geo, delta)
...
+
+    return screen.object.get_suqare_distance(s, x, y)
```
psychon: Uhm... Or a failed `sed`? Same typo here.

lib/awful/tag.lua
```diff
@@ -785,6 +866,10 @@ end
 
 --- Add a signal to all attached tags and all tags that will be attached in the
 -- future. When a tag is detached from the screen, its signal is removed.
+--
+-- Do not use.
```
psychon: Huh? Perhaps some word about why? I don't see the reason (nor the replacement function)

lib/awful/tag.lua
```diff
@@ -833,11 +826,15 @@ function tag.setproperty(_tag, prop, value, ignore_setters)
     end
 end
 
---TODO move to client
-
 --- Tag a client with the set of current tags.
+-- @deprecated awful.tag.withcurrent
 -- @param c The client to tag.
 function tag.withcurrent(c)
+--     util.deprecate("c:to_selected_tags() instead of awful.tag.selectedlist")
```
psychon: There are many commented deprecation calls in this commit. Why?

lib/awful/tag.lua
```diff
@@ -833,11 +826,15 @@ function tag.setproperty(_tag, prop, value, ignore_setters)
...
+
+    -- It can't use c:to_selected_tags() because awful.tag is loaded before
+    -- awful.client
```
psychon: I don't get this, `tag.withcurrent` will only be called long after all of `awful` is loaded, I think. No?

objects/client.c
```diff
@@ -649,12 +649,12 @@
...
+/* Set a __index metamethod for all client instances.
  * @tparam function cb The meta-method
  * @function set_index_miss_handler
  */
 
-/** Set a __newindex metamethod for all client instances.
+/* Set a __newindex metamethod for all client instances.
  * @tparam function cb The meta-method
  * @function set_newindex_miss_handler
  */
```
psychon: Just remove all these "ghosts comments" then. They don't have any purpose anymore.

psychon:
Looks good enough for me. I only spotted two typos, I think (which was what you wanted me to look for most).

Look at my comments above, do something about them or not and then feel free to merge this.

:+1:

Elv13:
Thanks for spending (a probably very boring) hour reading this!

docs/config.ld
Elv13: Don't trust github for the commit order, this commit is the second one in line, whatever github say about it. Github seem to show the commit in the order they were first added, not the "git order", that's probably a bug.

Elv13:
Lets hope for the best and prepare for the worst!

docs/config.ld
psychon: Ah, I know that problem. Sorry then.

PR #792, opened 2016-04-03
Fix build

Elv13:
@psychon I will merge this soon, feel free to revert the commits with a better idea. The build need to be fixed soon.

A better idea may be to disable generating the documentation by default generate the doc tests only when an additional cmake flag is passed, but this wont fix travis, so it isn't a practical solution for now.

To quote myself:

> So, it is #787 that "broke" the build because both PR were developed in parallel and made different 
> assumptions. Last week new test framework assumed only CAPI normally print on stderr and #787 
> assume printing warnings on stderr has no consequences. Both assumptions were true when the
> code  was published, but they no longer were by the time #787 was merged.

### DIFF
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -107,6 +107,32 @@ function(escape_code path escaped_content pre_header post_header)
     set(${post_header} ${example_post_header} PARENT_SCOPE)
 endfunction()
 
+# Only gears.debug.print_warning are allowed on stderr, everything else will
+# trigger a build failure to catch regressions and problem early.
+function(check_for_problems stderr result)
+
+    # If there is nothing to check, return
+    if (TEST_ERROR STREQUAL "")
+        set(${result} 0 PARENT_SCOPE)
+        return()
+    endif()
+
+    string(REGEX REPLACE "\n" ";" error_lines "${stderr}")
+
+    foreach (LINE ${error_lines})
+
+        # gears.debug.print_warning lines look like:
+        # yyyy-mm-dd hh:mm:ss W: message content
+        if (NOT ${LINE} MATCHES "^[0-9 :-]+ W:")
+            set(${result} 1 PARENT_SCOPE)
+            return()
+        endif()
+
+    endforeach()
+
+    set(${result} 0 PARENT_SCOPE)
+endfunction()
+
 # Execute a lua file.
 function(run_test test_path namespace template escaped_content)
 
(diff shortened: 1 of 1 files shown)


blueyed:
What do you think about ignoring the warnings on stderr instead (like suggested in https://github.com/awesomeWM/awesome/pull/787#issuecomment-205013082)?

Elv13:
> What do you think about ignoring the warnings on stderr instead (like suggested in #787 (comment))?

Well, the cmake check is kind of designed to catch these. The idea is/was "the default config should run without problems" -> "if there is problems and there wasn't any a commit ago, then this is a regression: be mean and quit".

In a way, it make sense, if our tests start to just ignore problems with a broad regex, then it will bite us at some point. No?

actionless:
what the difference assumed to be between classical and soft warning?

unless it's temporary workaround that separation a bit bugging me

Elv13:
> what the difference assumed to be between classical and soft warning?

One is on stderr and the other stdout. The build now forbid code on stderr and consider it an error and stop the build. This was designed to help tests detect regressions on the CI system (travis). But as this code was written before the check was implemented, it didn't trigger the error on travis when it was last built (before the check was implemented), but did when merged (after the check was implemented)

actionless:
Sorry for being unclear, i meant what the difference will be in their usage, ie in which case which of them should be thrown.

blueyed:
I see your point, @Elv13.
But I suggest rather matching warnings (by prefix (date + "W:")) only, instead of moving them to another stream.
A warning should never cause a build failure, right?

@actionless 
What lgi version / package / distribution do you use?

blueyed:
I seems luarocks only has 0.7.2, while the current version is 0.9.0?!
(https://github.com/pavouk/lgi/releases, https://luarocks.org/modules/luarocks/lgi)

Elv13:
Maybe your right, I will add a regex mask to filter warnings instead of
this patch. I still don't like the idea of letting new code introducing
stderr content, but your solution is cleaner than mine.

actionless:
arch, lua-lgi 0.9.0-1

Elv13:
There is an unrelated error on the Lua 5.3 target, merging anyway as this change has absolutely nothing to do with that and the build already have a `[build|failed]` tag anyway.

psychon:
Where does that unrelated error come from? It confuses me. According to the error message, there is an `#error` in `/usr/include/lua5.3/luaconf.h`. Where does that file come from? Looking at `.travis.xml` and Lua's `Makefile`, I think we are installing to `/usr/include/luaconf.h`.

Elv13:
The error seems to be because someone messed up a luarock update. Should we disable the 5.3 target until it works again? If a luarock doesn't compile on a clean state container, then someone should bark loud enough sooner or later, we could disable it for now and try again next week. I got a build ~10 hours ago, then every builds after that failed.

psychon:
I don't think it's luarock-related. In each build, luarock fails three times with the same error, all in `luaconf.h`. However, I do not know what caused it. Looking at luaconf.h, it seems like `LLONG_MAX` suddenly went missing.

blueyed:
I am trying to fix it in https://github.com/awesomeWM/awesome/pull/801.

blueyed:
The issue is with lua-term https://github.com/hoelzro/lua-term/issues/16.

Issue #793, opened 2016-04-03
Segfault when using Xephyr


Elv13:
Here's the first issue found since the screen PR was merged

```
2016-04-03 16:49:15 /usr/local/share/awesome/lib/awful/screen.lua:197: W: BOOM!

Thread 1 "awesome" received signal SIGSEGV, Segmentation fault.
0x000000000043e7ae in screen_getbycoord (x=1272, y=327)
    at /home/lepagee/dev/awesome/build/objects/screen.c:367
367	    screen_t *nearest_screen = globalconf.screens.tab[0];
(gdb) bt
#0  0x000000000043e7ae in screen_getbycoord (x=1272, y=327)
    at /home/lepagee/dev/awesome/build/objects/screen.c:367
#1  0x000000000041b807 in luaA_mouse_index (L=0x663f30) at /home/lepagee/dev/awesome/build/mouse.c:142
#2  0x00007ffff4cd15c8 in luaD_precall () from /usr/lib64/liblua.so.5
#3  0x00007ffff4cd19d4 in luaD_call () from /usr/lib64/liblua.so.5
#4  0x00007ffff4cda876 in callTMres.isra.0 () from /usr/lib64/liblua.so.5
#5  0x00007ffff4cdbc0e in luaV_execute () from /usr/lib64/liblua.so.5
#6  0x00007ffff4cd1a1d in luaD_call () from /usr/lib64/liblua.so.5
#7  0x00007ffff4cd0d0e in luaD_rawrunprotected () from /usr/lib64/liblua.so.5
#8  0x00007ffff4cd1b8b in luaD_pcall () from /usr/lib64/liblua.so.5
#9  0x00007ffff4ccd648 in lua_pcall () from /usr/lib64/liblua.so.5
#10 0x00000000004277d5 in luaA_dofunction (nret=0, nargs=1, L=0x663f30)
    at /home/lepagee/dev/awesome/build/common/lualib.h:75
#11 signal_object_emit (L=L@entry=0x663f30, arr=arr@entry=0x654740 <global_signals>, 
    name=name@entry=0x447364 "debug::error", nargs=nargs@entry=1)
    at /home/lepagee/dev/awesome/build/common/luaobject.c:254
#12 0x0000000000416eff in luaA_dofunction_on_error (L=0x663f30)
    at /home/lepagee/dev/awesome/build/luaa.c:400
#13 0x00007ffff4cd15c8 in luaD_precall () from /usr/lib64/liblua.so.5
[...]

PR #798, opened 2016-04-04
Fix crash in screen_getbycoord without any screens

blueyed:
Ref: https://github.com/awesomeWM/awesome/issues/793#issuecomment-205245346


psychon:
Heh, someone is faking commits in my name. ;-)
Thanks

PR #801, opened 2016-04-04
Travis: fix the Lua 5.3 build

blueyed:
Trying to fix the Lua 5.3 build (https://travis-ci.org/awesomeWM/awesome/jobs/120692700).

Ref: https://github.com/awesomeWM/awesome/pull/792#issuecomment-205123319

The issue is caused by a lua-term update: https://github.com/hoelzro/lua-term/issues/16.


psychon:
Thanks for identifying this so quickly.
I think that 02c4cff0615493d7a02e0463a194efec2821a83f can be reverted. Lua-term was fixed and it seems to build on Travis again.

PR #803, opened 2016-04-04
doc: improve doc for awful.screen.focused

blueyed:
@Elv13 
Why was `awful.screen.default_focused_args` added?
Isn't it good enough to monkeypatch/overwrite the function itself?

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -182,16 +182,14 @@ end
 
 --- Get the focused screen.
 --
--- Valid arguments are:
---
--- * **client**: Use the client screen instead of the mouse screen.
--- * **mouse**: Use the mouse screen (default).
---
 -- It is possible to set `awful.screen.default_focused_args` to override the
 -- default settings.
 --
--- @tparam[opt={}] table args Some arguments.
--- @treturn screen The focused screen object
+-- @tparam[opt] table args
+-- @tparam[opt=false] table args.client Use the client screen instead of the
+--   mouse screen.
+-- @treturn ?screen The focused screen object, or `nil` in case no screen is
+--   present currently.
 function screen.focused(args)
     args = args or screen.default_focused_args or {}
     return get_screen(args.client and capi.client.screen or capi.mouse.screen)


Elv13:
It was added to avoid users having to monkeypatch anything and use case where you want 2 different types of focused screen.

I am currently cleaning up the API and adding a lot of documentation (see the other PR about that), so I try to make things consistent and documented. Having a comment like "oh, yeah, by the way, if you don't like what this function does, then monkeypath it" wasn't really in line with what I call a good API (there really was such comment). So I added an imaginary default field so the user don't have to monkey patch anything.

Then, it could also potentially break my collision module.

By the way, I suggest a

```
--- The default `awful.screen.focused` mode.
-- @field awful.client.default_focused_args
```

I also suggest you push this into the `client_properties` (it's now more than client) branch on my GitHub fork. The API doc changed a little. I made new types of tags and merged doc pages of awful.screen and screen into 1 API. This is to avoid having to look at 2 different places for related functions. The feedbacks on this have been positive so far. It doesn't really break the API, it just move things around. I would appreciate your help on that PR, I know you care a lot about the DOC, lets make a great one for 3.6.

blueyed:
@Elv13 
You are referring to https://github.com/awesomeWM/awesome/pull/790, right?
Or some others as well?

Is this PR here good to go as-is?
The other PR can than build on that (e.g. when the name of the setting is changed).

Personally, I do not like the indirection of some global config setting, given that the function itself was meant to be monkeypatched from the beginning.  But it's OK to go that route, of course.

Elv13:
>  do not like the indirection of some global config setting

Neither do I, but in that case, the whole "please monkeypatch" comment was already hacky enough I feel it was a minor coherence improvement, even though it still isn't exactly nice.

> You are referring to #790, right?

yes

> Is this PR here good to go as-is?
> The other PR can than build on that (e.g. when the name of the setting is changed).

It need some testing. I will do a final check to see if I didn't miss anything. I also need to add some new tests to catch potential regressions. You can use it and build on it, but I would prefer wait another couple of days before merging it. Also, there is probably a thousand typos left. I also want `awful.tag` and `tag` (and maybe `awful.mouse`) to be ported first so there is only 1 PR that break the API rather than 2 in a row.

blueyed:
My question was if this PR (here) can me merged (the doc fix), for now?

Elv13:
Oh, this one, sure, but I will have to change it again in the other one as the ldoc syntax changed a little. That's why I was suggesting to push it into #790, but I can resolve the conflict on my side

blueyed:
Cool, thanks!

Issue #796, opened 2016-04-04
Displaying startup errors fails on master (no screen)


blueyed:
In case of startup errors the following code will fail, because there's no screen available yet ([source](https://github.com/awesomeWM/awesome/blob/e24cb901cbe95718901c5b0b73b15ac0448f989d/awesomerc.lua#L18-L22)):

```lua
if awesome.startup_errors then
    naughty.notify({ preset = naughty.config.presets.critical,
                     title = "Oops, there were errors during startup!",
                     text = awesome.startup_errors })
end
```

```
error while running function!
stack traceback:
        /home/daniel/.config/awesome/rc.lua:3215: in main chunk
error: /home/daniel/.config/awesome/rc.lua:3215: attempt to index a nil value (field '?')
========================
error while running function!
stack traceback:
        /usr/share/awesome/lib/naughty/core.lua:617: in function 'naughty.core.notify'
        /etc/xdg/awesome/rc.lua:22: in main chunk
error: /usr/share/awesome/lib/naughty/core.lua:617: attempt to index a nil value (local 's')
2016-04-04 14:12:41 E: awesome: main:699: couldn't find any rc file
```

I am not sure how to address this, but it would be helpful to `print` it here also (and not rely on `naughty` alone)?!

Elv13:
Can this still happen?

psychon:
We reverted the late screen addition stuff. I still think that it would be a good idea to not have coldplug, but only hotplug of monitors, but people get mad at me when I implement that. :-)

PR #805, opened 2016-04-04
naughty.notify: handle non-existing screen, print a warning

blueyed:
Ref: https://github.com/awesomeWM/awesome/issues/796

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -421,7 +421,8 @@ end
 -- @tparam[opt] table args.actions Mapping that maps a string to a callback when this
 --   action is selected.
 -- @usage naughty.notify({ title = "Achtung!", text = "You're idling", timeout = 0 })
--- @return The notification object
+-- @treturn ?table The notification object, or nil in case a notification was
+--   not displayed.
 function naughty.notify(args)
     if naughty.config.notify_callback then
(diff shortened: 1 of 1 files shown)


Elv13:
Well, I am not sure this is very user friendly compared to the "old way", but given the problem is already occurring, better merge this.

blueyed:
@awesome-robot 
You're hopefully fixed now.  I've restarted the build for reference: https://travis-ci.org/awesomeWM/awesome/jobs/120732804

blueyed:
> The command "eval git fetch origin +refs/pull/805/merge:" failed. Retrying, 2 of 3.

Meh.

lib/naughty/core.lua
```diff
@@ -437,6 +438,11 @@ function naughty.notify(args)
     local text = args.text or preset.text
     local title = args.title or preset.title
     local s = get_screen(args.screen or preset.screen or screen.focused())
+    if not s then
+        require("gears.debug").print_warning(
+            "naughty.notify: there is no screen available to display the notification.")
```
psychon: Should this also print the text (and title?) of the notification?

blueyed: Yeah, makes sense after all!

(I've only thought about the `awesome.startup_errors` use case, where it gets printed already.)

PR #810, opened 2016-04-05
gears.screen: Merge into awful.screen

Elv13:
gears modules usually don't depend on Awesome C-API. This code has
been placed there for unclear reasons.

Also, there is ongoing work to unify each "concepts" API into one
single page. Having `gears.screen` go against this effort.

@psychon, sorry for not complaining sooner and breaking the API 48 hours after it was introduced. However, `gears.screen` isn't compatible with my work on the API cleanup. I would really like to avoid multi-file-for-the-same-concept problems again. Also, that would be the only gears code to depend on CAPI (beside a soft dependency on xresources to get the DPI and `load_image`). Increasing coupling at this point isn't something I like. Wayland is coming, and *I* will make it happen (I hacked a prototype, it's not *that* hard after all), but in my plan, the C core has to go (for Wayland only, X11 would still use it, I don't plan some funky abstraction, just expose the same CAPI where possible). It wont be for this year, I have already too many WIP code to upstream and almost nobody use Wayland yet anyway.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -91,7 +91,7 @@ end
 
 -- {{{ Wallpaper
 if beautiful.wallpaper then
-    gears.screen.connect_for_each_screen(function(s)
+    awful.screen.connect_for_each_screen(function(s)
         gears.wallpaper.maximized(beautiful.wallpaper, s, true)
     end)
 end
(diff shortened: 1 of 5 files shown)


blueyed:
Do we really need the deprecation in this case?

Elv13:
> Do we really need the deprecation in this case?

I am probably too careful. It will be removed soon, but better let it be deprecated for a week or 2, no?

psychon:
The reason that it is in gears is because `gears.wallpaper` needs it. I want to move this to `awful.screen` as well (and yes, right now I think we can do so without any backwards compatibility).

So... this is fine with me, but feel free to do so without avoiding breakage.
(I think I'll have to come up with some new C-API for `gears.wallpaper` anyway...)

Elv13:
Ok then, I will replace all the use of `gears.screen` with `awful.screen` and then merge this (tomorrow, as my record of late night big changes is not exactly error free ;) ).

psychon:
Thanks. Feel free to make `gears.wallpaper` depend on `awful.screen` for now (and add a comment complaining about this being bad and needing to be fixed). I'll head to work...

Issue #808, opened 2016-04-04
test-leaks seems to be flaky / "Some object was not garbage collected!"


blueyed:
Build: https://travis-ci.org/awesomeWM/awesome/jobs/120752692

```
== Running /home/travis/build/awesomeWM/awesome/tests/test-leaks.lua ==
+start_awesome
+export DISPLAY=:5
+cd /home/travis/build/awesomeWM/awesome/tests/../build
+awesome_pid=13416+DISPLAY=:5 /home/travis/build/awesomeWM/awesome/tests/../build/awesome -c /home/travis/build/awesomeWM/awesome/tests/../build/awesomerc.lua
+cd -
+wait_until_success wait for awesome startup via awesome-client echo 'return 1' | DISPLAY=:5 '/home/travis/build/awesomeWM/awesome/tests/../utils/awesome-client' 2>&1
+[ true = true ]
+set +x
+[ ! -r /home/travis/build/awesomeWM/awesome/tests/test-leaks.lua ]
+DISPLAY=:5 /home/travis/build/awesomeWM/awesome/tests/../utils/awesome-client
+cat /home/travis/build/awesomeWM/awesome/tests/test-leaks.lua
+tail -n 100000 -f --pid 13416 /tmp/tmp.tirTf7WQk1/_awesome_test.log
2016-04-04 22:36:06 W: Cannot reliably detect EOF on an GIOInputStream with this LGI version
Some object was not garbage collected!
2016-04-04 22:36:07 W: awesome: luaA_dofunction:77: error while running function!
stack traceback:
	[string "-- Some memory leak checks as integration tes..."]:38: in function 'collectable'
	[string "-- Some memory leak checks as integration tes..."]:66: in function 'f'
	lib/awful/remote.lua:25: in function <lib/awful/remote.lua:21>
error: @lib.wibox.widget.imagebox (table: 0x41002b90)
Error: run_steps() was never called
+grep -q -E ^Test finished successfully$ /tmp/tmp.tirTf7WQk1/_awesome_test.log
+echo ===> ERROR running /home/travis/build/awesomeWM/awesome/tests/test-leaks.lua! <===
===> ERROR running /home/travis/build/awesomeWM/awesome/tests/test-leaks.lua! <===
+grep --color -o --binary-files=text -E .*[Ee]rror.*|.*assertion failed.* /tmp/tmp.tirTf7WQk1/_awesome_test.log
[...]

PR #812, opened 2016-04-05
Travis: build luajit from source (2.0.4 as of now)

blueyed:
This tries to address https://github.com/awesomeWM/awesome/issues/808 by updating luajit to the current upstream version (2.0.2 => 2.0.4).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,13 +57,14 @@ install:
     fi
 
   # Install luarocks (for the selected Lua version).
-  - travis_retry wget https://keplerproject.github.io/luarocks/releases/luarocks-2.3.0.tar.gz
-  - tar xf luarocks-2.3.0.tar.gz
-  - cd luarocks-2.3.0
-  - ./configure --lua-version=$LUA --with-lua-include=${LUAINCLUDE} ${LUAROCKS_ARGS}
-  - make build
-  - sudo make install
-  - cd ..
+  - |
+    travis_retry wget https://keplerproject.github.io/luarocks/releases/luarocks-2.3.0.tar.gz
+    tar xf luarocks-2.3.0.tar.gz
+    cd luarocks-2.3.0
+    ./configure --lua-version=$LUA --with-lua-include=${LUAINCLUDE} ${LUAROCKS_ARGS}
+    make build
+    sudo make install
+    cd ..
 
   # lgi.
   - sudo apt-get install -y gir1.2-pango-1.0 libgirepository1.0-dev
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -46,13 +44,18 @@ install:
       (cd src && make SYSCFLAGS="-DLUA_USE_LINUX -ULUA_COMPAT_5_2" SYSLIBS="-Wl,-E -ldl -lreadline" LUA_A=liblua.so MYCFLAGS="-fPIC" RANLIB=: AR="gcc -shared -ldl -o" liblua.so) || exit 1
       sudo make INSTALL_TOP=/usr/ INSTALL_INC=${LUAINCLUDE} TO_LIB=liblua.so linux install || exit 1
       cd ..
+    elif [[ "$LUANAME" == "luajit-2.0" ]]; then
```
psychon: Could you move this into its own block (`if`) instead of using `elif`?`That way this will be easier to find when Travis folds the commands.

psychon:
Looks good, nice idea. The changelogs don't mention anything that sounds like our issues, but whatever.

I'll spend the day making Travis re-run the LuaJIT build. Perhaps I can trigger a failure eventually. Feel free to tell me to stop that.

psychon:
I managed to produce a failure (after many retries). See the current travis build: https://travis-ci.org/awesomeWM/awesome/jobs/121031551
I'm not sure if this should still be merged...

blueyed:
> I'm not sure if this should still be merged...

While it does not fix the build issue it's still better to use an explicit (most recent) upstream version than the one shipped in Ubuntu in my opinion.

blueyed:
> after many retries

Like 5, 10 or 50?

psychon:
I'm not sure, but I guess "more like 5 or 10".

Also, I installed LuaJIT locally and briefly looked into the issue. It becomes much harder to hit if I comment out half of the default config, but sometimes it still occurs. At this point I stopped investigating since I have no good idea on how to continue. Also, I saw this to not only occur with the layoutbox in `test-leaks.lua`, but also with `tooltip_delayed()`. I explicitly tried to also hit this issue with other tests in there, but failed.

Tips for reproducing the issue:
- Remove the start-Xephyr-code from `tests/run.sh`
- Add `--sleep-interval=0.1` to the call to `tail` in `tests/run.sh`
- `make && xvfb-run -w 0.5 -s "-screen 0 800x600x24" sh -c 'while tests/run.sh test-leaks.lua ; do : ; done'`

blueyed:
Merging this, if only for the simplified env/setup.

.travis.yml
But that would hit the `else` case also for `LUA=5.3`, wouldn't it?

psychon: Oh, I missed the `else`, sorry.

PR #816, opened 2016-04-08
awesomerc.lua: swap meaning of j and k keys

reinerh:
Using j for previous and k for next is more intuitive.
When using more than 2 screens, mod+ctrl+j was focusing the screen
to the right of the current one, while mod+ctrl+k was focusing the
one to the left.(though j is located to the left of k).

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -235,13 +235,13 @@ globalkeys = awful.util.table.join(
     awful.key({ modkey,           }, "Escape", awful.tag.history.restore,
               {description = "go back", group = "tag"}),
 
-    awful.key({ modkey,           }, "j",
+    awful.key({ modkey,           }, "k",
         function ()
             awful.client.focus.byidx( 1)
         end,
         {description = "focus next by index", group = "client"}
     ),
-    awful.key({ modkey,           }, "k",
+    awful.key({ modkey,           }, "j",
         function ()
             awful.client.focus.byidx(-1)
         end,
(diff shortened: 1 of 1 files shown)


Elv13:
Hello @reinerh,

Thanks for your contribution. While it may (well, it *does*) make sense, but I am quite sure many users wont appreciate breaking their muscle memory after 9 years. I will let others give their opinion, but I don't think changing some "core" default shortcuts is a good idea.

In a not so unrelated topic, my https://github.com/elv13/collision module use the normal arrow keys and doesn't have the issue you describe (it does use geometry to selected the next/previous client/tag/screen).

blueyed:
> many users wont appreciate breaking their muscle memory after 9 years

Those will have a custom config anyway, and that diff would be small.

But I also see it more in the sense of "up" and "down" instead of "left" / "right".

So I'm neutral on this change.

Regarding the patch however, it should just replace "j" and "k", to have it (way) smaller.

psychon:
> Using j for previous and k for next is more intuitive.

Just like "j" in vim moves the cursor to the previous line and "k" to the next? ;-)

I don't have these key bindings in my config (I think... I'm not sure), so I don't care much.

> When using more than 2 screens, mod+ctrl+j was focusing the screen
to the right of the current one, while mod+ctrl+k was focusing the
one to the left

I'm not sure how portable this observation is. In the code, this does "next" / "previous" in the right direction. So, this all depends on the order we give to screens internally and that depends mostly on the X server config, I guess.

I'm not convinced.

blueyed:
Closing this, given the not-too-enthusiastic feedback.

@reinerh 
Thanks for suggesting it though anyway, of course!

PR #817, opened 2016-04-09
Add C-API for root window size

psychon:
Instead of having `gears.wallpaper` do magic to get the size of the root window, it now asks the C code for the correct numbers.


blueyed:
:+1:

PR #818, opened 2016-04-09
Randr stuff

psychon:
I noticed this due to #672 and it's a small step towards in. When the size of the root window changes, we will now update our in-memory idea of how big the root window is and then restart. Right now that's useless, but it's useful for when we stop restarting on this kind of change.


blueyed:
:+1:

Issue #799, opened 2016-04-04
Default config: non-working (commented) code with tags in rules


blueyed:
https://github.com/awesomeWM/awesome/blob/e24cb901cbe95718901c5b0b73b15ac0448f989d/awesomerc.lua#L454-L456:

    -- Set Firefox to always map on tags number 2 of screen 1.
    -- { rule = { class = "Firefox" },
    --   properties = { tag = tags[1][2] } },

Accessing tags here (when setting up the rules) won't work.
For my rules, I've wrapped them in `gears.screen.connect_for_each_screen(function(s) ... end)`.

The default config could also be changed to use `screen.primary` then regardless?!

psychon:
Uhm... how could something like this even work with multiple screens? I have no good idea right now.

Elv13:
Yeah, moving the scan screen down really broken this use case for me. Tyrannical have a declarative tag syntax like the new widget system one.

```lua
...
    {
        name = "Term",
        init        = true                                           ,
        exclusive   = true                                           ,
        icon        = utils.tools.invertedIconPath("term.png")       ,
        screen      = {config.scr.pri, config.scr.sec} ,
        layout      = awful.layout.suit.tile                         ,
        focus_new   = true                                           ,
        selected    = true,
        nmaster     = 2,
        mwfact      = 0.6,
        index       = 1,
        max_clients = five_layout,
        class       = {
            "xterm" , "urxvt" , "aterm","URxvt","XTerm"
        },
    } ,
...
```

You can use `screen=` to tell where you want the tag to be created (either a screen index or an array of indices). In the case the number of screen didn't match, it would simply fallback to the lowest valid index or use `[1]`. This is how Tyrannical handled moving tags to the right screen when you connected a new one (and Awesome restarted).

The problem with this construct is that it doesn't really work with screen objects for obvious reasons. I can probably invent a fix for this, after all, it broke because Awesome got better, not the other way around. So I will find a solution myself.

But another thing is a little more problematic. Declaration like `screen = screen.count() >= 3 and 2 or 1` don't really work anymore as they are interpreted when the config is parsed and no screens are present. Again, it is a new "problem" introduced because of a feature I rather like. However this one doesn't really have a solution beside wrapping that in a function (fugly, but functional...).

[...]

PR #819, opened 2016-04-10
Accept screen and tag names in awful.rules

psychon:
This fixes https://github.com/awesomeWM/awesome/issues/799
This should not be merged before https://github.com/awesomeWM/awesome/pull/790 is in, because I don't want to make @Elv13's job harder than it already is. I guess I'll have to rebase this afterwards, but the idea can already be reviewed.

### DIFF
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -53,7 +53,7 @@ can add:
 Note that all "rule" entries need to match. If any of the entry does not
 match, the rule won't be applied.
 
-If a client matches multiple rules, their applied in the order they are
+If a client matches multiple rules, they are applied in the order they are
 put in this global rules table. If the value of a rule is a string, then the
 match function is used to determine if the client matches the rule.
 
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -211,7 +211,9 @@ end
 -- @tab props Properties to apply.
 -- @tab[opt] callbacks Callbacks to apply.
 function rules.execute(c, props, callbacks)
+    local handle_later = { focus = true, switchtotag = true }
     local switchtotag = props.switchtotag
+
     for property, value in pairs(props) do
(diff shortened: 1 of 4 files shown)


lib/awful/rules.lua
```diff
@@ -195,23 +212,29 @@ end
...
+            c:tags({ t })
         elseif property == "height" or property == "width" or
                 property == "x" or property == "y" then
             local geo = c:geometry();
             geo[property] = value
             c:geometry(geo);
-        elseif property == "focus" then
+        elseif property == "focus" or property == "switchtotag" then
             -- This will be handled below
             (function() end)() -- I haven't found a nice way to silence luacheck here
```
blueyed: Could this be with the last `if`, i.e. `elseif property =~ "focus" and property =~ "swtichtotag"` there?

psychon: This is to get rid of the "I haven't found a nice way to silence luacheck here", right?
Yeah, that could work. I'll (hopefully remember to) do a separate PR for that. I guess I would do:
```lua
local ignore = { focus = true, switchtotag = true }
....
elseif not ignore[property] then
    if type(c[property]) == "function" then
        c[property](c, value)
    else
        c[property] = value
    end
end
```

Elv13:
+1

All I would change (and I might do it myself after this is merged) is moving the documentation into a new page instead of Awful rules. This will make it more visible and easier to "wikitize". Something like "Managing clients"

psychon:
Rebased onto master. I addressed @blueyed's comment. Since I honestly don't know how, I didn't do anything about @Elv13's comment.

Elv13:
> Since I honestly don't know how, I didn't do anything about @Elv13's comment.

I have an `awful.rules` PR in the pipeline too, so I will do it. Basically, you add a nn-name.md file in the docs folder, modify config.ld and that's about it. Everything else is just markdown.

PR #823, opened 2016-04-12
screen: XRandR 1.5 support

kiryl:
XRandR 1.5 adds support for the new monitor objects.

'Monitor' is a rectangular subset of the screen which represents a
coherent collection of pixels presented to the user. Each Monitor is be
associated with a list of outputs (which may be empty).

The patch below matches 1:1 screens in AwesomeWM with XRandR's Monitors.
This way I get one screen across my 4K monitor, which is represented by
two CRTCs.

One missing thing is handling change in monitor configuration: I don't
see a way to get XCB generate event about this.

Background info: http://keithp.com/blogs/MST-monitors/

### DIFF
--- a/event.c
+++ b/event.c
@@ -399,9 +399,7 @@ event_handle_configurenotify(xcb_configure_notify_event_t *ev)
 {
     xcb_screen_t *screen = globalconf.screen;
 
-    if(ev->window == screen->root
-       && (ev->width != screen->width_in_pixels
-           || ev->height != screen->height_in_pixels))
+    if(ev->window == screen->root)
         /* it's not that we panic, but restart */
         awesome_restart();
 
--- a/objects/screen.c
+++ b/objects/screen.c
@@ -152,6 +152,8 @@
  * @function set_newindex_miss_handler
  */
 
+DO_ARRAY(xcb_randr_output_t, randr_output, DO_NOTHING);
+
 struct screen_output_t
 {
     /** The XRandR names of the output */
@@ -159,7 +161,7 @@ struct screen_output_t
     /** The size in millimeters */
     uint32_t mm_width, mm_height;
     /** The XID */
-    xcb_randr_output_t output;
+    randr_output_array_t outputs;
 };
 
 static void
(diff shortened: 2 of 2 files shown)


objects/screen.c
```diff
@@ -241,6 +243,143 @@ screens_exist(void)
     return globalconf.screens.len > 0;
 }
 
+#ifdef XCB_RANDR_GET_MONITORS
```
psychon: Could you replace this with something like the following (is there a nicer way to express this?)
```
/* Monitors were introduced in RandR 1.5 */
#if XCB_RANDR_MAJOR_VERSION > 1 || XCB_RANDR_MINOR_VERSION >= 5
```

objects/screen.c
```diff
@@ -241,6 +243,143 @@ screens_exist(void)
...
+static bool
+screen_scan_randr_monitors(void)
+{
+    xcb_randr_get_monitors_cookie_t monitors_c = xcb_randr_get_monitors(globalconf.connection, globalconf.screen->root, 1);
+    xcb_randr_get_monitors_reply_t *monitors_r = xcb_randr_get_monitors_reply(globalconf.connection, monitors_c, NULL);
+    xcb_randr_monitor_info_iterator_t monitor_iter;
+    bool found = false;
+
+    for(monitor_iter = xcb_randr_get_monitors_monitors_iterator(monitors_r); monitor_iter.rem; xcb_randr_monitor_info_next(&monitor_iter))
```
psychon: One just has to love the short syntax that XCB's iterators give you. :-) I always try to avoid them.

objects/screen.c
```diff
@@ -241,6 +243,143 @@ screens_exist(void)
...
+            output.name = strdup("unknown");
+        }
+        randr_output_array_init(&output.outputs);
+
+        randr_outputs = xcb_randr_monitor_info_outputs(monitor_iter.data);
+        for(int i = 0; i < xcb_randr_monitor_info_outputs_length(monitor_iter.data); i++) {
+            randr_output_array_append(&output.outputs, randr_outputs[i]);
+        }
+
+        screen_output_array_append(&new_screen->outputs, output);
```
psychon: I can see why you don't want the deduplication that `screen_add` does, but due to this you also skip emission of the `added` signal. That will have to be fixed somehow. I suggest renaming `screen_add` to `screen_add_deduplicate` (feel free to come up with a better name, I don't like this one) and then having a new `screen_add` that just handles the "signal stuff".

objects/screen.c
kiryl: I'm not sure why would we want to do version-based check, if we can check capability directly.

objects/screen.c
kiryl: Any justification to avoid them? I'm new to XCB.

objects/screen.c
psychon: Ok, but at least do `#ifdef XCB_RANDR_GET_MONITORS /* Monitors were introduced in RandR 1.5 */`, so that one doesn't have to figure this out from just `screen_scan_randr()`.

psychon:
This conflicts with some of my work for #672, but whatever. At least someone else figured out how monitors are supposed to work. Thanks a lot!
:+1:

objects/screen.c
psychon: The syntax requires long lines, as you see in this example. I don't have access to RandR 1.5 XCB headers right now, but I guess `for (int i = 0; i < xcb_randr_get_monitors_monitors_length(monitors_r); i++)` could work and be shorter (and then accessing that part of the reply as `xcb_randr_get_monitors_monitors(reply)[i]`.

However, feel free to do this how you want. Both approaches are equivalent and the one without iterators gets us longer lines elsewhere.

objects/screen.c
kiryl: Actually, I've just missed the screen_add() part :-/
I think it should work fine in this case too.

event.c
```diff
@@ -399,17 +399,9 @@ event_handle_configurenotify(xcb_configure_notify_event_t *ev)
...
-           || ev->height != screen->height_in_pixels))
+    if(ev->window == screen->root)
         /* it's not that we panic, but restart */
         awesome_restart();
-
-    /* Copy what XRRUpdateConfiguration() would do: Update the configuration */
-    if(ev->window == screen->root) {
-        screen->width_in_pixels = ev->width;
-        screen->height_in_pixels = ev->height;
-    }
```
psychon: Sorry, but this reverts cb46a1163e904d37deee0e069cc81320ff34ea37 from just four days ago. The goal is to get rid of the `awesome_restart()` and handle screen changes without restarting. We are not there yet, which is why we still restart. We are on the way there, which is why we have this dead code.

Could you leave that part of the code intact?

(Also: Thanks for figuring out what kind of notify is sent for monitor changes! I don't think this is documented in the randr protocol docs.)

PR #824, opened 2016-04-13
Rules refactor, part 1

Elv13:
Revive https://github.com/awesomeWM/awesome/pull/53

Now that the new property system is merged, it is possible for modules to add extra properties themselves. However, it is limited to "real" client properties. There is still no way to add "rules only" property that are not kept after the initial rules application.

This PR add the ones from #53, [plan to] add unit test and [plan to] document everything properly. It conflict with @psychon other PR and isn't ready at all, but early comments are welcome.

Edit: After adding the tests, I found out that most! properties were broken or had unhandled corner case. So... Nobody ever write complex rules or are they are those powerusers don't know how to report bugs? Whatever:

 * **x**: Broken when "border_width" is set or left titlebars are used
 * **y**: Borken when"border_width" is set or top titlebars are used
 * **width**: See above + right litlebar
 * **height**: Same as above
 * **switchtotag**: Have a race with the "manage" -> "tag.withcurrent" code in `awful.tag`
 * **tag**: Had dead code
 * **screen**: Had a race condition with switchtotag
 * **urgent**: Had a race with screen and another with switchtotag+focus
 * **focusable**: Was broken yet again when request::activate was introduced (and also because of FS1098, that I also fixed)
 * **no_overlap**: The "no_overlap" call in rc.lua "manage" section conflict with the geometry rules as the hints are not set (idk why).
 * **size_hints_honor**: If set to false, it would be applied too late, causing height and width offsets due to placement or various geometry related properties

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -458,6 +458,11 @@ awful.rules.rules = {
         }
       }, properties = { floating = true }},
 
+    -- Add titlebars to normal clients and dialogs
+    { rule_any = {type = { "normal", "dialog" }
+      }, properties = { titlebars_enabled = true }
+    },
+
     -- Set Firefox to always map on the tag named "2" on screen 1.
     -- { rule = { class = "Firefox" },
     --   properties = { screen = 1, tag = "2" } },
@@ -481,48 +486,48 @@ client.connect_signal("manage", function (c)
         -- Prevent clients from being unreachable after screen count changes.
         awful.placement.no_offscreen(c)
     end
+end)
 
-    local titlebars_enabled = true
-    if titlebars_enabled and (c.type == "normal" or c.type == "dialog") then
-        -- buttons for the titlebar
-        local buttons = awful.util.table.join(
-            awful.button({ }, 1, function()
-                client.focus = c
-                c:raise()
-                awful.mouse.client.move(c)
-            end),
-            awful.button({ }, 3, function()
-                client.focus = c
-                c:raise()
-                awful.mouse.client.resize(c)
-            end)
-        )
+-- Add a titlebar if titlebars_enabled is set to true in the rules.
+client.connect_signal("request::titlebars", function(c)
+    -- buttons for the titlebar
+    local buttons = awful.util.table.join(
+        awful.button({ }, 1, function()
+            client.focus = c
+            c:raise()
+            awful.mouse.client.move(c)
+        end),
+        awful.button({ }, 3, function()
+            client.focus = c
+            c:raise()
+            awful.mouse.client.resize(c)
+        end)
+    )
 
-        awful.titlebar(c) : setup {
-            { -- Left
-                awful.titlebar.widget.iconwidget(c),
-                buttons = buttons,
-                layout  = wibox.layout.fixed.horizontal
+    awful.titlebar(c) : setup {
+        { -- Left
+            awful.titlebar.widget.iconwidget(c),
+            buttons = buttons,
+            layout  = wibox.layout.fixed.horizontal
+        },
+        { -- Middle
+            { -- Title
+                align  = "center",
+                widget = awful.titlebar.widget.titlewidget(c)
             },
-            { -- Middle
-                { -- Title
-                    align  = "center",
-                    widget = awful.titlebar.widget.titlewidget(c)
-                },
-                buttons = buttons,
-                layout  = wibox.layout.flex.horizontal
-            },
-            { -- Right
-                awful.titlebar.widget.floatingbutton (c),
-                awful.titlebar.widget.maximizedbutton(c),
-                awful.titlebar.widget.stickybutton   (c),
-                awful.titlebar.widget.ontopbutton    (c),
-                awful.titlebar.widget.closebutton    (c),
-                layout = wibox.layout.fixed.horizontal()
-            },
-            layout = wibox.layout.align.horizontal
-        }
-    end
+            buttons = buttons,
+            layout  = wibox.layout.flex.horizontal
+        },
+        { -- Right
+            awful.titlebar.widget.floatingbutton (c),
+            awful.titlebar.widget.maximizedbutton(c),
+            awful.titlebar.widget.stickybutton   (c),
+            awful.titlebar.widget.ontopbutton    (c),
+            awful.titlebar.widget.closebutton    (c),
+            layout = wibox.layout.fixed.horizontal()
+        },
+        layout = wibox.layout.align.horizontal
+    }
 end)
 
 -- Enable sloppy focus
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -430,7 +430,10 @@ awful.rules.rules = {
                      focus = awful.client.focus.filter,
                      raise = true,
                      keys = clientkeys,
-                     buttons = clientbuttons } },
+                     buttons = clientbuttons,
+                     placement = awful.placement.no_overlap+awful.placement.no_offscreen
+     }
+    },
 
     -- Floating clients.
     { rule_any = {
@@ -472,17 +475,13 @@ awful.rules.rules = {
 -- {{{ Signals
 -- Signal function to execute when a new client appears.
 client.connect_signal("manage", function (c)
-    if not awesome.startup then
-        -- Set the windows at the slave,
-        -- i.e. put it at the end of others instead of setting it master.
-        -- awful.client.setslave(c)
+    -- Set the windows at the slave,
+    -- i.e. put it at the end of others instead of setting it master.
+    -- if not awesome.startup then awful.client.setslave(c) end
 
-        -- Put windows in a smart way, only if they do not set an initial position.
-        if not c.size_hints.user_position and not c.size_hints.program_position then
-            awful.placement.no_overlap(c)
-            awful.placement.no_offscreen(c)
-        end
-    elseif not c.size_hints.user_position and not c.size_hints.program_position then
+    if awesome.startup and
+      not c.size_hints.user_position
+      and not c.size_hints.program_position then
         -- Prevent clients from being unreachable after screen count changes.
         awful.placement.no_offscreen(c)
     end
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -1,7 +1,6 @@
 -- Standard awesome library
 local gears = require("gears")
 local awful = require("awful")
-awful.rules = require("awful.rules")
 require("awful.autofocus")
 -- Widget and layout library
 local wibox = require("wibox")
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -12,70 +12,13 @@ local client = client
 local screen = screen
 local ipairs = ipairs
 local math = math
+local util = require("awful.util")
 local aclient = require("awful.client")
+local aplace = require("awful.placement")
 
 local ewmh = {}
 
 local data = setmetatable({}, { __mode = 'k' })
-
-local function store_geometry(window, reqtype)
-    if not data[window] then data[window] = {} end
-    if not data[window][reqtype] then data[window][reqtype] = {} end
-    data[window][reqtype] = window:geometry()
-    data[window][reqtype].screen = window.screen
-end
-
---- Maximize a window horizontally.
---
--- @param window The window.
--- @param set Set or unset the maximized values.
-local function maximized_horizontal(window, set)
-    if set then
-        store_geometry(window, "maximized_horizontal")
-        local g = screen[window.screen].workarea
-        local bw = window.border_width or 0
-        window:geometry { width = g.width - 2*bw, x = g.x }
-    elseif data[window] and data[window].maximized_horizontal
-        and data[window].maximized_horizontal.x
-        and data[window].maximized_horizontal.width then
-        local g = data[window].maximized_horizontal
-        window:geometry { width = g.width, x = g.x }
-    end
-end
-
---- Maximize a window vertically.
---
--- @param window The window.
--- @param set Set or unset the maximized values.
-local function maximized_vertical(window, set)
-    if set then
-        store_geometry(window, "maximized_vertical")
-        local g = screen[window.screen].workarea
-        local bw = window.border_width or 0
-        window:geometry { height = g.height - 2*bw, y = g.y }
-    elseif data[window] and data[window].maximized_vertical
-        and data[window].maximized_vertical.y
-        and data[window].maximized_vertical.height then
-        local g = data[window].maximized_vertical
-        window:geometry { height = g.height, y = g.y }
-    end
-end
-
---- Fullscreen a window.
---
--- @param window The window.
--- @param set Set or unset the fullscreen values.
-local function fullscreen(window, set)
-    if set then
-        store_geometry(window, "fullscreen")
-        data[window].fullscreen.border_width = window.border_width
-        window.border_width = 0
-        window:geometry(screen[window.screen].geometry)
-    elseif data[window] and data[window].fullscreen then
-        window.border_width = data[window].fullscreen.border_width
-        window:geometry(data[window].fullscreen)
-    end
-end
 
 local function screen_change(window)
     if data[window] then
@@ -202,12 +145,60 @@ function ewmh.urgent(c, urgent)
     end
 end
 
+-- Map the state to the action name
+local context_mapper = {
+    maximized_vertical   = "maximize_vertically",
+    maximized_horizontal = "maximize_horizontally",
+    fullscreen           = "maximize"
+}
+
+--- Move and resize the client.
+--
+-- This is the default geometry request handler.
+--
+-- @tparam client c The client
+-- @tparam string context The context
+-- @tparam[opt={}] table hints The hints to pass to the handler
+function ewmh.geometry(c, context, hints)
+    context = context or ""
+
+    local original_context = context
+
+    -- Now, map it to something useful
+    context = context_mapper[context] or context
+
+    local props = util.table.clone(hints or {}, false)
+    props.store_geometry = props.store_geometry==nil and true or props.store_geometry
+
+    -- If it is a known placement function, then apply it, otherwise let
+    -- other potential handler resize the client (like in-layout resize or
+    -- floating client resize)
+    if aplace[context] then
+
+        -- Check if it correspond to a boolean property
+        local state = c[original_context]
+
+        -- If the property is boolean and it correspond to the undo operation,
+        -- restore the stored geometry.
+        if state == false then
+            aplace.restore(c,{context=context})
+            return
+        end
+
+        local honor_default = original_context ~= "fullscreen"
+
+        if props.honor_workarea == nil then
+            props.honor_workarea = honor_default
+        end
+
+        aplace[context](c, props)
+    end
+end
+
 client.connect_signal("request::activate", ewmh.activate)
 client.connect_signal("request::tag", ewmh.tag)
 client.connect_signal("request::urgent", ewmh.urgent)
-client.connect_signal("request::maximized_horizontal", maximized_horizontal)
-client.connect_signal("request::maximized_vertical", maximized_vertical)
-client.connect_signal("request::fullscreen", fullscreen)
+client.connect_signal("request::geometry", ewmh.geometry)
 client.connect_signal("property::screen", screen_change)
 client.connect_signal("property::border_width", geometry_change)
 client.connect_signal("property::geometry", geometry_change)
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -177,8 +177,14 @@ end
 --
 -- @client c A client to tag
 -- @tag[opt] t A tag to use. If omitted, then the client is made sticky.
-function ewmh.tag(c, t)
+-- @tparam[opt={}] table hints Extra information
+function ewmh.tag(c, t, hints) --luacheck: no unused
+    -- There is nothing to do
+    if not t and #c:tags() > 0 then return end
+
     if not t then
+        c:to_selected_tags()
+    elseif type(t) == "boolean" and t then
         c.sticky = true
     else
         c.screen = t.screen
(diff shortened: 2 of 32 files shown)


tests/test-awful-rules.lua
```diff
@@ -0,0 +1,84 @@
...
+end
+
+local function test_rules()
+    -- Test if callbacks works
+    assert(callback_called)
+
+    -- Make sure "smart" dynamic properties are applied
+    assert(get_client_by_class("rule_float").floating)
+    local geo = get_client_by_class("rule_float"):geometry()
+    --assert(geo.width == 100 and geo.height == 100) --FIXME height=100+workarea.y+border_width
```
Elv13: @psychon Is that normal? The gtk client ask for 100x100, but awesome gives 100x118 here

psychon: Does this run with titlebars enabled?

Elv13: Ah, right, sorry. I wonder how I missed that...

Elv13: Ah, I get it now, the problem is that there is no titlebar yet when the rule is executed, so when the tb is added, everything is off when you use `geometry` or `placement` properties.

Elv13:
Ok, so this will 2 PR after all. It's already too much

left to part 2
max_client property
request::* documentation
port awful.rules to the new ldoc API
create a new "manage clients" documentation/wiki page
fix the urgent test use of deprecated APIs

lib/awful/placement.lua
```diff
@@ -493,7 +496,7 @@ function placement.no_overlap(c)
     new.width = geometry.width
     new.height = geometry.height
 
-    return c:geometry({ x = new.x, y = new.y })
+    return c:geometry({ x = new.x + c.border_width, y = new.y + c.border_width })
 end
```
psychon: The problem you were seeing is that `placement.no_offscreen()` would move a client by its border width, right?
I think your patch only hides the problem, but does not fix it. The actual bug is in `area_common` which does:
```
geometry.x = geometry.x - border
geometry.y = geometry.y - border
geometry.width = geometry.width + 2 * border
geometry.height = geometry.height + 2 * border
```
Adding the border to the width is correct, but subtracting it from (x,y) is incorrect. The position is outside of the border, not inside of it.

Thus, because of the above subtraction, you get the wrong values here.

Put different: `c:geometry{ x = c:geometry().x }` does not modify the position of a client, but adding the border width does move it. So if your patch is needed, then something is messing up the position (this something is `area_common()`).

lib/awful/rules.lua
```diff
@@ -205,36 +206,97 @@ function rules.apply(c)
...
 function rules.execute(c, props, callbacks)
-    local handle_later = { focus = true, switchtotag = true }
-    local switchtotag = props.switchtotag
+
+    -- Some properties need to be handled first. For example, many properties
+    -- depend that the client is tagged, this isn't yet the case.
+    for prop, handler in pairs(rules.high_priority_properties) do
+        if props[prop] then
+            handler(c, props[prop])
+            props[prop] = nil
```
psychon: Could you add a warning to the docs that this function modifies the `props` argument? (Even better would be not to do so, but whatever...)

lib/awful/rules.lua
```diff
@@ -205,36 +206,97 @@ function rules.apply(c)
...
+        end
+    end
 
     for property, value in pairs(props) do
+
+
         if property ~= "focus" and type(value) == "function" then
             value = value(c)
         end
```
psychon: This part is missing in the previous for-loop.

Perhaps its better if this function does not modify its argument, but instead creates a copy of it "by hand" and in the process calls functions like this?

lib/awful/rules.lua
```diff
@@ -205,36 +206,97 @@ function rules.apply(c)
...
+                    value = value(c, props)
+                end
+
+                if value then
+                    c.screen = value.screen
+                    c:tags{ value }
+                end
+
+            elseif property == "switchtotag" and value and props.tag then
+                props.tag:view_only()
```
psychon: Uhm, am I missing some part of the diff? This patch removes the `switchtotag` variable above, but I don't see it removing the `if switchtotag and c.first_tag then` that follows after this for loop?

`switchtotag` seems like a candidate for delayed properties. I bet some later commit actually does this...

lib/awful/rules.lua
```diff
@@ -277,15 +277,9 @@ function rules.execute(c, props, callbacks)
...
-                if type(value) == "function" then
-                    value = value(c, props)
+                if t then
+                    c:tags{ t }
                 end
-
-                if value then
-                    c.screen = value.screen
-                    c:tags{ value }
-                end
```
psychon: This looks like a rebase gone wrong. The previous commit introduced all of this, so you should squash this into there.

lib/awful/rules.lua
```diff
@@ -266,7 +266,7 @@ function rules.execute(c, props, callbacks)
 
 
         if property ~= "focus" and type(value) == "function" then
-            value = value(c)
+            value = value(c, props)
```
psychon: This is not dead code, this is a new feature that this commit adds.

lib/awful/rules.lua
```diff
@@ -308,6 +309,32 @@ function rules.extra_properties.geometry(c, _, props)
...
+-- @tparam client c The client
+-- @tparam boolean|function|string value The value.
+-- @treturn tag The new tag
+function rules.high_priority_properties.new_tag(c, value)
+    local ty = type(value)
+    local t = nil
+
+    if ty == "boolean" then
+        -- Create a new tag named after the client class
+        t = a_tag.add(c.class or "N/A", {screen=c.screen})
```
psychon: Is it a good idea to use `c.screen` here? A non-high-priority-rule could modify the screen later.

lib/awful/rules.lua
```diff
@@ -236,17 +242,67 @@ rules.extra_properties = {}
...
 -- To add a new properties, just do:
 --
---    function awful.rules.high_priority_properties.my_new_property(c, value)
+--    function awful.rules.high_priority_properties.my_new_property(c, value, props)
 --        -- do something
 --    end
 --
+-- By default, the table has the following functions:
+--
+-- * tag
```
psychon: I should stop reviewing commits on GitHub. I saw a previous commit where this part of the code was already present. Thus, I'm again looking at  them in the wrong order....

lib/awful/rules.lua
```diff
@@ -236,17 +242,67 @@ rules.extra_properties = {}
...
+--
 -- @tfield table awful.rules.high_priority_properties
+-- @tfield function awful.rules.high_priority_properties.tag FOO
```
psychon: FOO?

lib/awful/rules.lua
```diff
@@ -236,17 +242,67 @@ rules.extra_properties = {}
...
+    for t in pairs(selected_tags) do
+        t.selected = false
+    end
+end
+
+function rules.extra_properties.geometry(c, _, props)
+    local cur_geo = c:geometry()
+
+    local new_geo = type(props.geometry) == "function"
+        and props.geometry(c, props) or props.geometry or {}
```
psychon: Why does this ignore the `value` argument and instead re-computes it? Isn't this line the same as doing `local new_geo = value or {}`?

lib/awful/rules.lua
```diff
@@ -262,32 +318,26 @@ function rules.execute(c, props, callbacks)
         end
     end
 
-    for property, value in pairs(props) do
+    -- Before requesting a tag, make sure the screen is right
```
psychon: I agree with this comment. Doesn't that mean that this code should be before the previous for-loop?

lib/awful/rules.lua
```diff
@@ -319,13 +320,13 @@ function rules.high_priority_properties.new_tag(c, value)
 
     if ty == "boolean" then
         -- Create a new tag named after the client class
-        t = a_tag.add(c.class or "N/A", {screen=c.screen})
+        t = atag.add(c.class or "N/A", {screen=c.screen})
```
psychon: Uhm. Shouldn't this be fixed in the commit that added this error instead?

lib/awful/rules.lua
```diff
@@ -356,6 +380,24 @@ function rules.execute(c, props, callbacks)
             or screen[props.screen]
     end
 
+<<<<<<< HEAD
+=======
```
psychon: .....

lib/awful/rules.lua
```diff
@@ -16,6 +16,7 @@ local type = type
 local ipairs = ipairs
 local pairs = pairs
 local atag = require("awful.tag")
+local util = require("awful.util")
```
psychon: This adds an unused import?

lib/awful/tag.lua
```diff
@@ -1370,11 +1366,6 @@ object.properties(capi.tag, {
     setter_fallback = tag.setproperty,
 })
 
--- fix a load loop
-client = require("awful.client")
-capi.client.connect_signal("manage", function(c) client.object.to_selected_tags(c) end)
```
psychon: I fear that this has to stay. Perhaps in a `gears.timer.delayed_call`.
For some reason, using `awful.rules` is still optional (`require("awful")` does not load `awful.rules`). People who don't use `awful.rules` get their new clients tagged due to this.

lib/awful/rules.lua
```diff
@@ -360,6 +360,13 @@ function rules.extra_properties.placement(c, value)
     end
 end
 
+function rules.extra_properties.tags(c, value)
+    if not type(value) == "table" then return end
```
psychon: Why a silent return instead of `error()` or `gears.debug.print_error()`? People who get this wrong would surely like some hint that something is wrong.

lib/awful/tag.lua
Elv13: It cannot stay. Delayed timer make the urgent and focus tests fail, so it doesn't work
also, it cause the "tag" property to fail due to a race condition, new_tag break too. See the comments in awful.rules and PR description.

It would rather make awful.rules mandatory. It doesn't conflict with neither Shifty nor Tyrannical, so what's the point of having this optional?

tests/_client.lua
```diff
@@ -0,0 +1,33 @@
...
+        "    application_id = 'org.awesomewm.tests.",class,
+        "'};",
+        "function app:on_activate()",
+        "    window.application = self;",
+        "    window:show_all();",
+        "end;",
+        "app:run {''}\" | lua"
+    }
+
+    spawn.with_shell(cmd)
```
psychon: I don't like `spawn.with_shell` due to (in theory) needing escaping on its argument (yeah, this is just test code, but still). I know that this is not an issue here, but could you still change this to `spawn({ "lua", "-e", code })` (and `code` is set up like you do right now).

tests/_client.lua
```diff
@@ -0,0 +1,33 @@
...
+        "local Gtk = lgi.require('Gtk');",
+        "local GdkPixbuf = lgi.GdkPixbuf;",
+        "Gtk.init();",
+        "local class = '",
+        class or 'test_app',"';",
+        "local window = Gtk.Window {",
+        "    default_width  = 100,",
+        "    default_height = 100,",
+        "    title          = '",
+            title or 'Awesome test client',
```
psychon: Could you move this to the beginning of the function? `title = title or "awesome test client"` (and same for `class`)
(Because right now you handle `class = nil` in only one of two places)

tests/_client.lua
```diff
@@ -0,0 +1,33 @@
...
+
+-- This file provide a simple, yet flexible, test client.
+-- It is used to test the `awful.rules`
+
+return function(class, title)
+    local cmd = table.concat {
+        "echo \"",
+        "local lgi = require 'lgi';",
+        "local Gtk = lgi.require('Gtk');",
+        "local GdkPixbuf = lgi.GdkPixbuf;",
```
psychon: Unused "import"

tests/test-awful-rules.lua
```diff
@@ -0,0 +1,290 @@
...
+
+local tb_height = awful.util.round(beautiful.get_font_height() * 1.5)
+-- local border_width = beautiful.border_width
+
+local function test_rule(rule)
+    rule.rule = rule.rule or {}
+
+    -- Create a random class. The number need to be large as "rule1" and "rule10" would collide
+    -- The math.ceil is necessary to remove the floating point, this create an invalid dbus name
+    rule.rule.class = rule.rule.class or "rule"..#tests..math.ceil(math.random()*100000000)
```
psychon: How about having a `local counter = 42` that you increment on each use? That would reduce the chance of collisions from "pretty unlikely" to "impossible".

Elv13: Until we get 420 tests!

objects/client.c
```diff
@@ -2834,25 +2779,6 @@ luaA_client_get_icon(lua_State *L, client_t *c)
...
-
-    if (c->focusable_set)
-        ret = c->focusable;
-
-    /* A client can be focused if it doesnt have the "nofocus" hint...*/
-    else if (!c->nofocus)
-        ret = true;
-    else
-        /* ...or if it knows the WM_TAKE_FOCUS protocol */
-        ret = client_hasproto(c, WM_TAKE_FOCUS);
```
psychon: This commit loses the above defaults. Also, you cannot check `client_hasproto(c, WM_TAKE_FOCUS)` from Lua and thus this cannot be moved into Lua. (we could export `WM_PROTOCOLS` to Lua, but for that we'd need a good use case first; I don't think this one is one)

lib/awful/ewmh.lua
```diff
@@ -202,12 +145,63 @@ function ewmh.urgent(c, urgent)
...
+
+        local honor_default = original_context ~= "fullscreen"
+
+        if props.honor_workarea == nil then
+            props.honor_workarea = honor_default
+        end
+
+        aplace[context](c, props)
+    elseif context == "fullscreen" then
+        props.honor_workarea = false
```
psychon: Doesn't this also need to check `c.fullscreen`, like in the previous case (`local state = c[original_context]`)?
Put differently: Does unfullscreening work?

lib/awful/ewmh.lua
```diff
@@ -202,12 +145,63 @@ function ewmh.urgent(c, urgent)
...
+    -- Now, map it to something useful
+    context = context_mapper[context] or context
+
+    local props = util.table.clone(hints or {}, false)
+    props.store_geometry = props.store_geometry==nil and true or props.store_geometry
+
+    -- If it is a known placement function, then apply it, otherwise let
+    -- other potential handler resize the client (like in-layout resize or
+    -- floating client resize)
+    if aplace[context] then
```
psychon: An argument of "fullscreen" is mapped to "maximize". `awful.placement.maximize` does exists. Thus, isn't the `elseif` that follows after this `if' dead code that can never be reached?

awesomerc.lua
```diff
@@ -469,17 +472,13 @@ awful.rules.rules = {
...
-
-        -- Put windows in a smart way, only if they do not set an initial position.
-        if not c.size_hints.user_position and not c.size_hints.program_position then
-            awful.placement.no_overlap(c)
-            awful.placement.no_offscreen(c)
-        end
-    elseif not c.size_hints.user_position and not c.size_hints.program_position then
+    -- Set the windows at the slave,
+    -- i.e. put it at the end of others instead of setting it master.
+    -- awful.client.setslave(c)
```
psychon: Shouldn't this comment stay inside `if not awesome.startup`? Otherwise, the client order will be messed up during restarts.

tests/test-maximize.lua
```diff
@@ -0,0 +1,91 @@
...
+    end,
+    function()
+        local c = client.get()[1]
+
+        local new_geo = c:geometry()
+        local sgeo = c.screen.workarea
+
+        --assert(new_geo.x-c.border_width==sgeo.x) --FIXME c:geometry({x=1}).x ~= 1
+
+        --assert(new_geo.width+2*c.border_width==sgeo.width) --FIXME off by 4px
```
psychon: I think these asserts should work before this is merged. Otherwise, you could just skip&remove this part of the test.

tests/test-awful-rules.lua
psychon: Ok, then use `string.format("%010d", counter)`, that will guarantee uniqueness for quite a while.

lib/awful/tag.lua
psychon: Ok, then add the require to awful and ensure that `awful.rules` even tags a client if no rules match it.

objects/client.c
Elv13: > WM_PROTOCOLS to Lua, but for that we'd need a good use case first; I don't think this one is one)

I would also like to move maximize, apply size_hints and fullscreen code to lua, but they also interact with some xprops. I missed this one, I should been more careful.. But given the large amount of boilerplate code in C. Could some kind of hidden `c:set_xproperty()` take care of that. Eventually, I could also like to move capi.tag to lua, but that again require being able to set a few workspace properties.

It also like to reduce the amount of code in the C core so I can share more with the eventual Wayland port. I know some concepts (such as this one!), wont match the behavior of the Wayland equivalent, but still.

psychon: We had size hints in Lua once. It caused problems (but I don't remember which ones) and was eventually moved back.
As for moving the tag-stuff to Lua: I'd love to have that, it would get rid of about half of `ewmh.c`. However, some of the kinds of properties required for this can not even be set via our `xproperty` code (by the way, do we still need that? should it be dropped to simplify our code base?)

tests/test-awful-rules.lua
Elv13: Is that really worth it?

psychon: The birthday paradox says "yes", doesn't it.

Elv13:
Ok, I fixed the vast majority of your comments. I still need to fix the gtk test client and the placement use of border_width offsets. Feel free to review everything else

tests/_client.lua
```diff
@@ -0,0 +1,33 @@
...
+        "'};",
+        "window:set_wmclass(class, class);",
+        "local app = Gtk.Application {",
+        "    application_id = 'org.awesomewm.tests.",class,
+        "'};",
+        "function app:on_activate()",
+        "    window.application = self;",
+        "    window:show_all();",
+        "end;",
+        "app:run {''}\" | lua"
```
psychon: By the way, right now I'm on a Fedora box where `urxvt` does not support startup-notification and thus the sn-related tests always fail. I *think* that GTK automatically supports startup-notification and so this test-app could be used instead of `urxvt` in the tests.

Remind me to actually implement this eventually...

Elv13:
all done (I think)

lib/awful/rules.lua
```diff
@@ -205,36 +206,102 @@ function rules.apply(c)
...
+                    t = atag.find_by_name(props.screen, t)
+                end
+
+                if type(value) == "function" then
+                    value = value(c, props)
+                end
+
+                if value then
+                    c.screen = value.screen
+                    c:tags{ value }
```
psychon: This breaks tag-assignment. The new code isn't equivalent to the previous code. Also, the next comment also still applies.

However, the next commit cleans this up again... So the end result is ok, the way to get there is weird.

lib/awful/rules.lua
```diff
@@ -307,6 +308,32 @@ function rules.extra_properties.geometry(c, _, props)
...
+function rules.high_priority_properties.new_tag(c, value)
+    local ty = type(value)
+    local t = nil
+
+    if ty == "boolean" then
+        -- Create a new tag named after the client class
+        t = atag.add(c.class or "N/A", {screen=c.screen})
+    elseif ty == "string" then
+        -- Create a tag named after "value"
+        t = atag.add(value, {screen=c.screen})
```
psychon: Shouldn't both this and the previous case also use `volatile=true` for the new tag?

awesomerc.lua
```diff
@@ -481,9 +482,11 @@ client.connect_signal("manage", function (c)
         -- Prevent clients from being unreachable after screen count changes.
         awful.placement.no_offscreen(c)
     end
+end)
 
-    local titlebars_enabled = true
-    if titlebars_enabled and (c.type == "normal" or c.type == "dialog") then
+-- Add a titlebar if titlebars_enabled is set to true in the rules.
+client.connect_signal("request::titlebars", function(c)
+    if (c.type == "normal" or c.type == "dialog") then
```
psychon: Should this type-check also be moved into `awful.rules`? If I'm not mistaken, adding a new rule with `{ rule_any = { type = { "normal", "dialog" } }, properties = { titlebars_enabled = true } }` should work (and of course, I did not try this).

objects/client.c
```diff
@@ -3446,17 +3446,9 @@ client_class_setup(lua_State *L)
      */
     signal_add(&client_class.signals, "request::activate");
     /**
-     * @signal request::fullscreen
+     * @signal request::geometry
```
psychon: Could you add some words about the argument(s) to the signal? The C code seems to emit this with a single string argument, but the handler function accepts extra hints.

tests/test-maximize.lua
psychon: Ping, what's with these asserts?
Is it really that `c:geometry({x=1}).x ~= 1`? That's be weird...
For the second FIXME: I think maximized windows still have size hints applied. They are only ignored for fullscreen windows. However, I'm not 100% sure that this is the problem.

tests/_client.lua
```diff
@@ -0,0 +1,33 @@
...
+
+-- This file provide a simple, yet flexible, test client.
+-- It is used to test the `awful.rules`
+
+return function(class, title)
+    title = title or 'Awesome test client'
+
+    local cmd = table.concat {
+        "lua -e \"",
+        "local lgi = require 'lgi';",
```
psychon: Could you make this:
`local cmd = { "lua", "-e", table.concat { "here is the actual code" }}`? That way we don't have useless argument splitting going on and the code doesn't need to be enclosed in `"`.

You could also make this
```
local code = [[
local lgi = require('lgi');
etc
print('class is', 
]] .. class .. [[ , 'and Uli agrees that this part is ugly ]]
```

psychon:
I took a look at the commits leading to 637e841e382236e71d69fb6d3b468aedc64f0dd0 and the results are above. Feel free to merge this when done, I think.

PR #827, opened 2016-04-15
build: Be more careful about the use of ..

Elv13:
Fix @actionless problem. It use the same hacky solution as the markdown files.


actionless:
thanks, works for me now with unmodified pkgbuild

PR #828, opened 2016-04-16
Stop restarting on RandR changes

psychon:
Here is it, the "big" PR that fixes https://github.com/awesomeWM/awesome/issues/672!

This should hopefully contain the necessary changes to the C code. Note that none of this is really tested since I couldn't find anything to test this with. I don't want to break my X session!

A bit of a problem might be commit https://github.com/awesomeWM/awesome/commit/d985c23edae96f47014e333d0a1617e95a1c6564, but as far as I know the nvidia blob doesn't hit that problem any more, so I'll just hope for the best with it and assume it works.

Note that this does not yet handle anything on the Lua side, so there are e.g. lots of leaks, tags for removed screens aren't removed, `awful.wibox`es assigned to a removed screened aren't removed etc. If you want, feel free to work on this stuff.

### DIFF
--- a/event.c
+++ b/event.c
@@ -400,8 +400,7 @@ event_handle_configurenotify(xcb_configure_notify_event_t *ev)
     xcb_screen_t *screen = globalconf.screen;
 
     if(ev->window == screen->root)
-        /* it's not that we panic, but restart */
-        awesome_restart();
+        globalconf.screen_need_refresh = true;
 
     /* Copy what XRRUpdateConfiguration() would do: Update the configuration */
     if(ev->window == screen->root) {
@@ -782,7 +781,7 @@ event_handle_randr_screen_change_notify(xcb_randr_screen_change_notify_event_t *
         globalconf.screen->height_in_pixels = ev->height;
     }
 
-    awesome_restart();
+    globalconf.screen_need_refresh = true;
 }
 
 /** XRandR event handler for RRNotify subtype XRROutputChangeNotifyEvent
--- a/event.c
+++ b/event.c
@@ -992,10 +992,8 @@ void event_init(void)
     const xcb_query_extension_reply_t *reply;
 
     reply = xcb_get_extension_data(globalconf.connection, &xcb_randr_id);
-    if (reply && reply->present) {
-        xcb_randr_select_input(globalconf.connection, globalconf.screen->root, XCB_RANDR_NOTIFY_MASK_OUTPUT_CHANGE);
+    if (reply && reply->present)
         globalconf.event_base_randr = reply->first_event;
-    }
 
     reply = xcb_get_extension_data(globalconf.connection, &xcb_shape_id);
     if (reply && reply->present)
--- a/event.h
+++ b/event.h
@@ -38,9 +38,13 @@ void drawin_refresh(void);
 void client_focus_refresh(void);
 void client_border_refresh(void);
 
+/* objects/screen.c */
+void screen_refresh(void);
+
 static inline int
 awesome_refresh(void)
 {
+    screen_refresh();
     luaA_emit_refresh();
     banning_refresh();
     stack_refresh();
--- a/globalconf.h
+++ b/globalconf.h
@@ -92,6 +92,10 @@ typedef struct
     xcb_window_t selection_owner_window;
     /** Do we have RandR 1.3 or newer? */
     bool have_randr_13;
+    /** Do we have RandR 1.5 or newer? */
+    bool have_randr_15;
+    /** Do we have a RandR screen update pending? */
+    bool screen_need_refresh;
     /** Check for XTest extension */
     bool have_xtest;
     /** Check for SHAPE extension */
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -446,12 +446,14 @@ function client.object.to_selected_tags(self)
         end
     end
 
-    if #tags == 0 then
-        tags = self.screen.selected_tags
-    end
+    if self.screen then
+        if #tags == 0 then
+            tags = self.screen.selected_tags
+        end
 
-    if #tags == 0 then
-        tags = self.screen.tags
+        if #tags == 0 then
+            tags = self.screen.tags
+        end
     end
 
     if #tags ~= 0 then
--- a/lib/awful/wibox.lua
+++ b/lib/awful/wibox.lua
@@ -285,6 +285,14 @@ end
 capi.client.connect_signal("property::struts", update_wiboxes_on_struts)
 capi.client.connect_signal("unmanage", update_wiboxes_on_struts)
 
+capi.screen.connect_signal("removed", function(s)
+    for _, wprop in ipairs(wiboxes) do
+        if wprop.screen == s then
+            wprop.wibox.visible = false
+        end
+    end
+end)
+
 function awfulwibox.mt:__call(...)
     return awfulwibox.new(...)
 end
--- a/objects/screen.c
+++ b/objects/screen.c
@@ -197,56 +197,19 @@ luaA_checkscreen(lua_State *L, int sidx)
         return luaA_checkudata(L, sidx, &screen_class);
 }
 
-static inline area_t
-screen_xsitoarea(xcb_xinerama_screen_info_t si)
+static screen_t *
+screen_add(lua_State *L, screen_array_t *screens)
 {
-    area_t a =
-    {
-        .x = si.x_org,
-        .y = si.y_org,
-        .width = si.width,
-        .height = si.height
-    };
-    return a;
-}
-
-static void
-screen_add(lua_State *L, int sidx)
-{
-    screen_t *new_screen = luaA_checkudata(L, sidx, &screen_class);
-
-    foreach(screen_to_test, globalconf.screens)
-        if(new_screen->geometry.x == (*screen_to_test)->geometry.x
-           && new_screen->geometry.y == (*screen_to_test)->geometry.y)
-            {
-                /* we already have a screen for this area, just check if
-                 * it's not bigger and drop it */
-                (*screen_to_test)->geometry.width =
-                    MAX(new_screen->geometry.width, (*screen_to_test)->geometry.width);
-                (*screen_to_test)->geometry.height =
-                    MAX(new_screen->geometry.height, (*screen_to_test)->geometry.height);
-                lua_remove(L, sidx);
-                return;
-            }
-
-    sidx = luaA_absindex(L, sidx);
-    lua_pushvalue(L, sidx);
-    luaA_object_ref(L, sidx);
-    screen_array_append(&globalconf.screens, new_screen);
-    luaA_object_emit_signal(L, -1, "added", 0);
-    lua_pop(L, 1);
-}
-
-static bool
-screens_exist(void)
-{
-    return globalconf.screens.len > 0;
+    screen_t *new_screen = screen_new(L);
+    luaA_object_ref(L, -1);
+    screen_array_append(screens, new_screen);
+    return new_screen;
 }
 
 /* Monitors were introduced in RandR 1.5 */
 #ifdef XCB_RANDR_GET_MONITORS
-static bool
-screen_scan_randr_monitors(void)
+static void
+screen_scan_randr_monitors(lua_State *L, screen_array_t *screens)
 {
     xcb_randr_get_monitors_cookie_t monitors_c = xcb_randr_get_monitors(globalconf.connection, globalconf.screen->root, 1);
     xcb_randr_get_monitors_reply_t *monitors_r = xcb_randr_get_monitors_reply(globalconf.connection, monitors_c, NULL);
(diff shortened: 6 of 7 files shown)


psychon:
If anyone is curious how much magic is needed so that removed screens can be garbage collected: https://github.com/psychon/awesome/commit/d2ed63accd149de0f5ca6b1125d979748bf11361
The change to `awful.tag.object.delete` actually fixes a kind-of-bug: It indexes `nil` if you try to delete the last tag on a screen.

blueyed:
@Elv13 
Have you tested this?
I will do so tomorrow myself, and then let's get it merged if it's OK.

Elv13:
I only have a laptop right now, so I can't really test. Also, we found no way to CI test this beside maybe adding some fake screens.

ff2000:
If it's enough to build + run (with default config) this PR and plug/unplug a second monitor I could give this a go. What should I suspect to see?

psychon:
@ff2000 When you plug in a new monitor and use the default config, everything should work as previously, just that awesome doesn't restart (e.g. the wallpaper is set, tags are created, a new wibox appears).
When you unplug a monitor... Well, lots of Lua errors will happen, because the Lua side keeps using the old, now broken screen. But that means that at least the C code did the right thing and "killed" the monitor.
In any case you could check something like `echo "local res = {} ; for s in screen do res[s] = s:geometry() end return require("gears.debug").dump_return(res)" | awesome-client` to see if the screen information changes correctly.
Also, when resizing a screen (switching to a different resolution), of course the C code should update the list of available screens, but (as always) the Lua code doesn't handle this correctly.

What would be really nice is switching between clone mode (two displays showing the same content, perhaps with them not even having the same geometry) and two screens next to each other. For clone mode, awesome should only "give" a single screen to Lua, but RandR still keeps these separate.

Disclaimer: I only did some really minimal testing on this patch. I don't know if the above is really what happens. I only tested this against some hardcoded "fake screens" that I made something like `screen_scan_randr` return and against Xvfb exporting just a single RandR screen.

ff2000:
First off I was not able to start with the default config. There was no error printed, so I can't say what actually went wrong. So I did the test with my own config.
I tried changing screen setup with xrandr. It actually did something, but it stopped updating my laptop screen. Also key bindings did not work (Meta+r for prompt, Meta+mouse for move/resize).

ff2000:
So today I gave it another go. And magically my xsession-errors-file got some content: I made a mistake when specifying CMAKE_INSTALL_PREFIX so the default theme was not found resulting in a non-complete beautiful-table :angry:
After fixing that I can also start with the default config. But again mouse doesn't work, Meta+r doesn't show the promptbox, etc. But now (magically, again...) I got a log:
```
2016-04-20 07:53:26 W: awesome: luaA_dofunction:77: error while running function!
stack traceback:
	/home/franz/src/awesome/build/lib/awful/layout/init.lua:68: in function 'awful.layout.get'
	/home/franz/src/awesome/build/lib/awful/mouse/init.lua:179: in function </home/franz/src/awesome/build/lib/awful/mouse/init.lua:176>
error: /home/franz/src/awesome/build/lib/awful/layout/init.lua:68: attempt to index a nil value
2016-04-20 07:53:26 W: awesome: event_handle_mousegrabber:123: Stopping mousegrabber.
2016-04-20 07:53:27 W: naughty.notify: there is no screen available to display the notification.
```
(this is directly after starting awesome, so no monitors got plugged in and no xrandr changes were done)
AHA!
After restarting awesome mouse and Meta+r work! And it seems I need to restart awesome after every call to xrandr.
So this looks as if it is not THAT far away from working. Re-setting up the screens seems to work, changes to screens need to be caught more reliably (probably this is the mentioned "work on the lua side"?).

ff2000:
And just forgot: awesome "lost" a window after changing to a different screen layout (actually the one I wrote the previous comment in...). But I got it back after switching to the previously used screen layout. But most likely this also is due to the missing integration on the lua side.

psychon:
Hrm. That definitely did not happen in Xephyr. Anyway, thanks a lot for testing.
```
2016-04-20 07:53:26 W: awesome: luaA_dofunction:77: error while running function!
stack traceback:
    /home/franz/src/awesome/build/lib/awful/layout/init.lua:68: in function 'awful.layout.get'
    /home/franz/src/awesome/build/lib/awful/mouse/init.lua:179: in function </home/franz/src/awesome/build/lib/awful/mouse/init.lua:176>
error: /home/franz/src/awesome/build/lib/awful/layout/init.lua:68: attempt to index a nil value
2016-04-20 07:53:26 W: awesome: event_handle_mousegrabber:123: Stopping mousegrabber.
```
This is `awful.mouse.client.move` failing because `c.screen` is nil. How the heck does this code run directly after startup? Weird.
```
2016-04-20 07:53:27 W: naughty.notify: there is no screen available to display the notification.
```
This doesn't sound good at all... No idea what went wrong, but I'll try to figure this out (next week...).

psychon:
Sigh. Problem exists between computer and chair. I screwed up a version check and thus the code tried to use RandR 1.5 even though we don't have the client side code for this which resulted in "no screens found". And I needed way too long to notice this.

I also added another commit that fixes a Lua error that I ran into while debugging this. Not completely related to this pull request, but not completely unrelated either...

juw:
Starting with one screen and adding a second one works properly but when removing the second screen, I get some errors:
```
2016-04-27 15:09:56 W: awesome: luaA_class_call_handler:390: error while running function!
stack traceback:
        [C]: in function '__index'
        /usr/share/awesome/lib/awful/screen.lua:21: in function 'get_screen'
        /usr/share/awesome/lib/awful/screen.lua:424: in function 'get_tags'
        /usr/share/awesome/lib/awful/screen.lua:449: in function 'get_selected_tags'
        /usr/share/awesome/lib/awful/screen.lua:470: in function </usr/share/awesome/lib/awful/screen.lua:469>
        (...tail calls...)
        [C]: in function '__index'
        /usr/share/awesome/lib/awful/layout/init.lua:68: in function 'get'
        /home/wolle/.config/awesome/rc.lua:673: in function </home/wolle/.config/awesome/rc.lua:672>
error: /usr/share/awesome/lib/awful/screen.lua:21: invalid object
2016-04-27 15:09:56 W: awesome: luaA_dofunction:77: error while running function!
stack traceback:
        [C]: in function '__index'
        /usr/share/awesome/lib/awful/widget/tasklist.lua:24: in function 'get_screen'
        /usr/share/awesome/lib/awful/widget/tasklist.lua:195: in function 'us'
        /usr/share/awesome/lib/awful/widget/tasklist.lua:204: in function </usr/share/awesome/lib/awful/widget/tasklist.lua:202>
        [C]: in function '__newindex'
        /home/wolle/.config/awesome/rc.lua:675: in function </home/wolle/.config/awesome/rc.lua:672>
2016-04-27 15:09:56 W: awesome: luaA_class_call_handler:390: error while running function!
stack traceback:
        [C]: in function '__index'
        /usr/share/awesome/lib/awful/screen.lua:21: in function 'get_screen'
        /usr/share/awesome/lib/awful/screen.lua:424: in function </usr/share/awesome/lib/awful/screen.lua:420>
        (...tail calls...)
        [C]: in function '__index'
        /usr/share/awesome/lib/awful/widget/taglist.lua:124: in function 'taglist_update'
        /usr/share/awesome/lib/awful/widget/taglist.lua:171: in function </usr/share/awesome/lib/awful/widget/taglist.lua:170>
        [C]: in function 'xpcall'
        /usr/share/awesome/lib/gears/protected_call.lua:37: in function </usr/share/awesome/lib/gears/protected_call.lua:36>
        (...tail calls...)
        /usr/share/awesome/lib/gears/timer.lua:158: in function </usr/share/awesome/lib/gears/timer.lua:156>
error: /usr/share/awesome/lib/awful/screen.lua:21: invalid object
2016-04-27 15:09:56 E: Error during a protected call: /usr/share/awesome/lib/awful/widget/taglist.lua:124: bad argument #1 to 'ipairs' (table expected, got nil) 
stack traceback:
        /usr/share/awesome/lib/gears/protected_call.lua:17: in function </usr/share/awesome/lib/gears/protected_call.lua:16>
        [C]: in function 'ipairs'
        /usr/share/awesome/lib/awful/widget/taglist.lua:124: in function 'taglist_update'
        /usr/share/awesome/lib/awful/widget/taglist.lua:171: in function </usr/share/awesome/lib/awful/widget/taglist.lua:170>
        [C]: in function 'xpcall'
        /usr/share/awesome/lib/gears/protected_call.lua:37: in function </usr/share/awesome/lib/gears/protected_call.lua:36>
        (...tail calls...)
        /usr/share/awesome/lib/gears/timer.lua:158: in function </usr/share/awesome/lib/gears/timer.lua:156>
2016-04-27 15:09:56 W: awesome: luaA_class_call_handler:390: error while running function!
stack traceback:
        [C]: in function '__index'
        /usr/share/awesome/lib/awful/screen.lua:21: in function 'get_screen'
        /usr/share/awesome/lib/awful/screen.lua:424: in function </usr/share/awesome/lib/awful/screen.lua:420>
        (...tail calls...)
        [C]: in function '__index'
        /usr/share/awesome/lib/awful/widget/tasklist.lua:277: in function 'filter'
        /usr/share/awesome/lib/awful/widget/tasklist.lua:130: in function 'tasklist_update'
        /usr/share/awesome/lib/awful/widget/tasklist.lua:184: in function </usr/share/awesome/lib/awful/widget/tasklist.lua:182>
        [C]: in function 'xpcall'
        /usr/share/awesome/lib/gears/protected_call.lua:37: in function </usr/share/awesome/lib/gears/protected_call.lua:36>
        (...tail calls...)
        /usr/share/awesome/lib/gears/timer.lua:158: in function </usr/share/awesome/lib/gears/timer.lua:156>
error: /usr/share/awesome/lib/awful/screen.lua:21: invalid object
2016-04-27 15:09:56 E: Error during a protected call: /usr/share/awesome/lib/awful/widget/tasklist.lua:278: bad argument #1 to 'ipairs' (table expected, got nil)
stack traceback:
        /usr/share/awesome/lib/gears/protected_call.lua:17: in function </usr/share/awesome/lib/gears/protected_call.lua:16>
        [C]: in function 'ipairs'
        /usr/share/awesome/lib/awful/widget/tasklist.lua:278: in function 'filter'
        /usr/share/awesome/lib/awful/widget/tasklist.lua:130: in function 'tasklist_update'
        /usr/share/awesome/lib/awful/widget/tasklist.lua:184: in function </usr/share/awesome/lib/awful/widget/tasklist.lua:182>
        [C]: in function 'xpcall'
        /usr/share/awesome/lib/gears/protected_call.lua:37: in function </usr/share/awesome/lib/gears/protected_call.lua:36>
        (...tail calls...)
        /usr/share/awesome/lib/gears/timer.lua:158: in function </usr/share/awesome/lib/gears/timer.lua:156>
```

psychon:
> Starting with one screen and adding a second one works properly

Good! :-)

> but when removing the second screen, I get some errors:

That's expected.

I don't know what `/home/wolle/.config/awesome/rc.lua:673` is, but the first error is that this calls `awful.layout.get()` for the removed screen. The second one is the tasklist for the removed screen trying to update, then the taglist does the same thing, etc. Some error is the tasklist querying `s.tags` for the removed screen (which apparently doesn't cause a "real" error, then `s.tags` ends up being `nil` and then the following `ipairs(nil)` causes the next error.

Seems like the stuff that should work actually does work. :-)

Anyone else wants to test? Or can this be merged?
This PR adds the `removed` signal to screens, which is needed for most of the cleanup on the Lua side after a screen is gone. #832 tries to do the stuff that is possible without having a `removed` signal.

juw:
> I don't know what `/home/wolle/.config/awesome/rc.lua:673` is

Sorry, thats
```
-- Enable sloppy focus
    if awful.layout.get(c.screen) ~= awful.layout.suit.magnifier <-line 673
        and awful.client.focus.filter(c) then
        client.focus = c
    end
end)
```

psychon:
Some clients still have `c.screen` pointing to the removed screen. That's one of the things which still need to be fixed (apparently the code for this that's included in this commit was broken by some later change).

psychon:
Rebased on master and did some changes.

For one, the "fix my broken `#if`"-commit is gone and was squashed into the commit that introduced the problem. The other thing is is that I added a special `valid` boolean to the screen struct. This fixes @juw's problem where a client's `c.screen` referred to a no longer valid screen, because the code for moving clients to other screens when their current screen goes away now actually works (the code now removes a screen from `globalconf.screens` before doing anything else).

Diff to the previous version is: https://gist.github.com/psychon/ba10e0c4c547fcbcb26f8df5285cdce1

psychon:
Does anybody mind if I add this to next? Should this be tested some more? How?

The necessary Lua changes after this don't seem all that bad. Most of this was already merged with #832. What is missing is a change to `awful.wibox` to hide/get rid of wiboxes whose screen is removed and some change to `naughty` to handle screen removal. (At least this made a "screen leak check" with my fake screen additions/removals work)

psychon:
Now that #832 is merged, I rebased this onto master (so that the fixes from there are included for testing) and added a new last commit (9a42ae87cfc683d0388c4a44199d903214d1839e) that makes `awful.wibox` hide wiboxes for removed screens.
I haven't tested this yet, but I think that now most things should also work when removing screens.

psychon:
Just because I said "most things should work":
- Moving screens isn't handled anywhere at all yet
- taglist and tasklist still cause errors after removing screens since they try to update for the removed screens
- Nothing at all removes tags yet. I'm thinking about something like https://gist.github.com/psychon/7b313885ac6d85c23175c9b0f503a29f for this, but this isn't tested at all yet.

PR #820, opened 2016-04-11
tests: Test all client layouts

Elv13:
This is a failing test extracted from the API refactor PR. It fails because of existing problems

For example, `set_master_width_factor` is defined as:

```lua
if mwfact >= 0 and mwfact <= 1 then
```

And a few other corner case (literally) where the clients are pushed to the screen edge with invalid sizes. It prints many warnings, asserts a few time if the titlebars are enables. Xephyr also complain in its own log about invalid operations.

It's rather easy to fix, but the question is where. Who's (object/module) responsibility is it to avoid those errors. Should all layouts check that themselves or should `capi.client` just reject invalid request. A third option is claim we don't care and mute the warnings.

PR #829, opened 2016-04-17
Catch under/overflows

psychon:
Previously, `c:geometry { width = -42 }` would mean that the C side assigns the value -42 to an `uint16_t` and thus the result was `UINT16_MAX-42`. This commit adds range checking to everything that I could spot. Some things (e.g. fake input) are not checked, but mostly because I didn't figure out good limits.

### DIFF
--- a/common/xutil.h
+++ b/common/xutil.h
@@ -28,6 +28,11 @@
 #include "common/atoms.h"
 #include "common/util.h"
 
+#define MAX_X11_COORDINATE INT16_MAX
+#define MIN_X11_COORDINATE INT16_MIN
+#define MAX_X11_SIZE UINT16_MAX
+#define MIN_X11_SIZE 1
+
 static inline char *
 xutil_get_text_property_from_reply(xcb_get_property_reply_t *reply)
 {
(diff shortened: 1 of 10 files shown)


psychon:
tile.lua:188: bad argument #1 to 'apply_size_hints' (integer expected, got number)

Sigh. This PR adds some "must be an integer" constraints that weren't there previously. Should this be handled on the Lua side (so that e.g. some "good" rounding can be done) or should I make the C code shut up about this and continue to silently truncate numbers?

Elv13:
> or should I make the C code shut up about this and continue to silently truncate numbers?

Given the recent problems caused by forcing wibox x/y/w/h to be integers (now reverted), I would argue that adding tons of math.ceil() everywhere is much more burdensome than doing it once.

When the "must be integer" patch was reverted, I removed 200! rounding calls from my config (yes, I have way too many modules creating way too many wiboxes).

On top of that, most of the time, only one kind of rounding make sense. For wibox and clients, it is always a `ceil()` as `floor()` will cause some "..." issues in wiboxes and titlebars textboxes.

Elv13:
Sorry for stealing your commits, but I made the changes here:
https://github.com/awesomeWM/awesome/pull/820

I think it make much more sense. See the commit messages for details. Feel free to pull my branch and squash my commits into yours if you like.

I think fixing all instances of `:geometry()` to have integers make both no sense and is very un-lua.

psychon:
Thanks! :-)

PR #832, opened 2016-04-17
Screen GC & stuff

psychon:
This started as some preparation for "make sure that screens can be garbage collected", but I got a bit carried away.

Commit https://github.com/awesomeWM/awesome/commit/e8d1c50a1db647a129e618b74da7ae142b4c7932 might be a bit controversial. It removes the `tags` table from the default config, since people should use e.g. `screen[1].tags` instead. This requires finding `.tags` in the documentation, but such is life...

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -162,6 +162,11 @@ function layout.arrange(screen)
     delayed_arrange[screen] = true
 
     timer.delayed_call(function()
+        if not screen.valid then
+            -- Screen was removed
+            delayed_arrange[screen] = nil
+            return
+        end
         if arrange_lock then return end
         arrange_lock = true
 
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -88,22 +88,9 @@ local function client_menu_toggle_fn()
 end
 -- }}}
 
--- {{{ Wallpaper
-if beautiful.wallpaper then
-    awful.screen.connect_for_each_screen(function(s)
-        gears.wallpaper.maximized(beautiful.wallpaper, s, true)
-    end)
-end
--- }}}
-
--- {{{ Tags
 -- Define a tag table which hold all screen tags.
+-- This table is filled later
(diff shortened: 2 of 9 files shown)


lib/awful/layout/init.lua
```diff
@@ -162,6 +162,11 @@ function layout.arrange(screen)
     delayed_arrange[screen] = true
 
     timer.delayed_call(function()
+        if not screen.valid then
```
Elv13: Doesn't that requires a pcall() like the clients do?

psychon: Hu? `.valid` was added specifically so that no `pcall`-related hacks are needed. It exists on all C-API objects and even works on objects that otherwise throw errors when you access them. So, no, this doesn't need a `pcall()` and whatever you mean with "like the clients do" doesn't need a `pcall()` either. What are you referring to?

Elv13: My config say otherwise, I need the `pcall` anyway or I get an assert "attempt to call an invalid object" in `if c.valid then`. At least, that was still the case not so long ago.

psychon: Hm, if you have a test for `tests/` for that, I'd like to check. I just added the following to the "catch all" `awful.rules`-rule in the default config and ran the test suite. It printed a bunch of true/false, but no errors.
```
callback = function(c) gears.timer.start_new(0.001, function() print(c.valid) return true end) end},
```

psychon:
Rebased on master. Commit 9e86f3bb0758b7a1799c725d751da02a1f6e880b is gone since this change was already applied in d736850e6058537dea4c9ec9b52ffc5ca9030458 (which is the sole reason why a rebase was necessary).

@Elv13 I'll continue to claim that `foo.valid` should never throw an error if `foo` is a C-API object.

@ Everyone Anything else that stops this from being merged?

psychon:
Adding this to next since apparently there are no complaints about this.

PR #838, opened 2016-04-18
Streamline/cleanup util.deprecate calls

blueyed:


Elv13:
Fine by me. I kind of like that syntax, but for the sake of coherence, your right, so you can merge this.

psychon:
Sure, why not

Issue #841, opened 2016-04-19
Make awful.client.focus.history.list public


blueyed:
There used to be `awful.client.data.focus`, which was moved to `awful.client.focus.history.internal`.
While the former was accessible, the latter is not.

This is useful to get a list of most recently focused clients, without having to hook into the signal yourself.

blueyed:
Closing this in favor of https://github.com/awesomeWM/awesome/pull/842.

Issue #733, opened 2016-03-03
Client with two tags always gets focus when switching between those tags


lohfu:
When i have a client belonging to two tags, when i switch between the tags focus will always be on the shared tag even if it was not focused.

i tried solving it with the following code (both with and without the timer) to no avail.

```
awful.tag.viewonly(awful.tag.gettags(i)[j]) awful.screen.focus(i) 
local focus_timer = timer({ timeout = 0.2 })
focus_timer:connect_signal("timeout", function()
    local c = awful.mouse.client_under_pointer()
    if not (c == nil) then
        client.focus = c
        c:raise()
    end
    focus_timer:stop()
end)
```

also, i noticed that moving the mouse does not change focus until it leaves the initial client

psychon:
Did you `focus_timer:start()` your timer? (= Does your code actually do anything?)

This issue is likely the same as https://awesome.naquadah.org/bugs/index.php?do=details&task_id=574.

lohfu:
my god i had indeed simply forgot to call `focus_timer:start()`. sorry.

why do i need the timer at all? setting the timeout to 0.001 makes it barely noticable though. would you recommend any other fix?

and yes, it seems it is a duplicate of that issue... was not aware there were 2 issue trackers.

psychon:
That other issue tracker is the old, read-only issue tracker from before we moved to GitHub.

The issue here is that the focus history is global. When you switch to a tag and no client has the input focus, then `awful.autofocus` picks the newest entry from the focus history that is currently visible. Thus, if the client that is on both tags was focused on the "old" tag (even when you later moved the focus elsewhere), it will also be picked on the "new" tag.

So the issue here is that you think of the focus history "per tag" while it really is global.

rogorido:
> So the issue here is that you think of the focus history "per tag" while it really is global.

And is there any workaround for this? This is maybe one of the few things which are very very annoyable in awesome....

Elv13:
Nothing technically prevent you from implementing per-tag and per-screen histories. It just require connecting a bunch of signals (mostly "tagged", "untagged", "manage" "focus" and "unfocus") then use that as the source for tag switching focus and mod4+tab

rogorido:
Thanks for your answer. I'm using awesome only since 2-3 days... could you maybe only point out in which source files I should look in order to have an idea?

Elv13:
[...]

PR #842, opened 2016-04-19
Make awful.client.focus.history.list public

blueyed:
There used to be `awful.client.data.focus`, which was moved to
`awful.client.focus.history.internal`.
While the former was accessible, the latter is not.

This is useful to get a list of most recently focused clients, without
having to hook into the signal yourself.

Closes https://github.com/awesomeWM/awesome/issues/841.

TODO:
 - [ ] make this read-only?!
 - [ ] better name?

### DIFF
--- a/lib/awful/client/focus.lua
+++ b/lib/awful/client/focus.lua
@@ -37,9 +37,7 @@ do
     })
 end
 
-local focus = {history = {}}
-
-local internal = {}
+local focus = {history = {list = {}}}
 
 local function get_screen(s)
     return s and capi.screen[s]
@@ -50,9 +48,9 @@ end
 -- @client c The client that must be removed.
 -- @function awful.client.focus.history.delete
 function focus.history.delete(c)
-    for k, v in ipairs(internal) do
+    for k, v in ipairs(focus.history.list) do
         if v == c then
-            table.remove(internal, k)
+            table.remove(focus.history.list, k)
             break
         end
     end
@@ -96,7 +94,7 @@ function focus.history.add(c)
     -- Remove the client if its in stack
     focus.history.delete(c)
     -- Record the client has latest focused
-    table.insert(internal, 1, c)
+    table.insert(focus.history.list, 1, c)
 end
 
 --- Get the latest focused client for a screen in history.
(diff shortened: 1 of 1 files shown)


Elv13:
This looks fine. The history will eventually need a better implementation, as discussed in #733 and on IRC.

 * Be able to "pause" history recording to be able to implement alt+tab like cycling without altering the history. Add `.next` in addition to `.previous` and make them wraparound.
 * Support fine grained history per tag and per screen
 * Exclude clients (such as sticky ones) from history (`history.get` already kind of support this, but not the other methods)
 * Set a default filter

Also, for the list being private. Isn't that the `awful.client.focus.history.get` job to return the list? It can be used in a `while` loop to get the list. I guess it would be nice if it was wrapper into a `awful.client.focus.history.iterate` method so it can be used as a for loop with a filter. Maybe this is a better way to go?

If you disagree with that, then this patch is fine and can be merged.

blueyed:
I agree with all your remarks, but would still merge this as-is for now (with minimal changes like making it read-only (is this possible?)), and agree on a name that does not have to change in the future.

As for using `awful.client.focus.history.get`: this would not really be efficient with the current implementation (it does not really matter, but shows that it's not the way to go).

Elv13:
> making it read-only (is this possible?))

yes `readonly = setmetatable({},{__newindex=function()end, __index=real_table})`

> and agree on a name that does not have to change in the future.

Well, if the "fixed" API only has iterators navigation and no "raw" access, then the name will have to change again anyway.

blueyed:
I've tried the following, but it does not support `#` (for length), which is 0:

    local internal = {}
    local focus = {
        int = internal,
        history = {
            list = setmetatable({}, {
                                __newindex=function() end,
                                __index=internal
                            })
        }}

Elv13:
There is `__len`, but I think it isn't very reliable in Lua 5.1. Just give read/write access for now, proper iterators is still the way to go.

blueyed:
@psychon 
What's your opinion?
Can this be merged as-is for now?

psychon:
Sure, feel free. And yeah, it might be nice if people wouldn't have to write their own history "stuff"...

PR #844, opened 2016-04-21
Geometry overhaul part 2.5

Elv13:
I added some commits originally intended for this PR in my awful.rules refactor (to avoid fixing the bugs twice). Here is most of the remaining ones.

This PR port the mouse resizing pipeline to use `request::geometry` and `awful.placement`. The code was already in `awful.layout.dynamic.resize`, but is now properly integrated.

**The geometry overhaul series:**
 
The goal of these patches is to implement a new API (`awful.placement`) that's flexible enough to replace all the other geometry related code. This allow many features to be re-used across many other components, reduce code duplication, add more flexibility to existing components and base the Awesome (floating) geometry code on top of a well documented and unit tested API. Finally, it merge many features found in my Radical, Collision and Tyrannical modules, reducing the thick layer of hacky abstractions my modules/config (and its users) are based on. It is a 4 part series:

 * Part 1: Add the new API [MERGED]
 * Part 2: Integrate the placement API with the client, rules and request:: APIs (40% merged, 60% here)
 * Part 3: Integrate with awful.wibox, awful.menu and awful.tooltip (the code exist in my Radical awesome3.6 branch, it still need some integration). Also support screen resize while at it because someone has to do it anyway (TODO)
 * Part 4: Integrate with other components, add documentation and tutorials. (TODO)

**New features:**

 * Mouse resize now support the sides
 * Windows 7 style snap to screen edges (also implemented by KDE, Unity and Gnome)
 * New placement functions and options
 * Support mouse grabber extensions (such as snap to grid, snap to edges, snap to clients)
 * The mouse API and documentation have been ported to the new API standards introduced in the last few months.
 * Many features previously mutually exclusive (such as dragtotag) can now be enabled at the same time

**Why:**

 * To conform to the "vision" of the request:: system
 * To share well tested code between all geometry related features instead of hardcoded copies everywhere in the code.
 * To have floating clients, stateless layouts and statefull layouts share the same resizing code[1]
 * Make move and resize operation work in more scenarios
 * Allow extensions to the move/resize code
 * Add custom "modes" such as outline resize or TWM like resize.
 * So layouts stop implementing mousegrabbers themselves.
 * The first half of this system is already merged anyway

**TODO:**

 * Also use the new pipeline for the "mouse move"
 * Fix the ldoc, I changed the method name a few time without updating the doc
 * Write some more tests
 * Fix the stateless layout resize (not very hard)
 * Turn the "mouse snap" code into a resize extension
 * Deprecate more code from `awful.mouse`
 * Clean the patches, fix luacheck and so on

[1] My dynamic layout already managed to hook into the stateless resize codepath, but it was a hack


psychon:
So far I only looked at the "introductionary text", but I already have to say that "The first half of this system is already merged anyway" is one of the best answers to "why?" that I ever saw. :-)

Elv13:
This is no longer a work in progress. I updated the PR description with the new features. <s>The tests will probably fail due to the test-leak issue, but merging the other PR should fix this.</s>

For part 3, I wont do one huge PR again, I will probably do "awful.wibox", "awful.tooltip", "wibox" and "fix add and remove screens" as 4 smaller PR. They will be smaller than this one anyway. After that, I will finish the shape/pattern/pixmap APIs and my layout stuff. Then test test test, doc doc doc. Then that will be it for me for 3.6! (and yes, I really do have a TODO list and I have been round-robin it since a while)

lib/awful/placement.lua
```diff
@@ -434,6 +447,24 @@ local function area_remove(areas, elem)
...
+        width  = p2x - p1x,
+        height = p2y - p1y,
+    }
+end
+
+-- Convert a rectangle and matrix info into a point
+local function rect_to_point(rect, corner_i, corner_j, n)
+    return {
+        x = rect.x + corner_i * math.floor(rect.width  / (n-1)),
+        y = rect.y + corner_j * math.floor(rect.height / (n-1)),
```
psychon: Do you need the `n` argument? In the entire PR, only this commit adds a call to this function and this commit has a hardcoded `n=2` for that call. Then, `math.floor(rect.width / (2-1))` can be simplified to `rect.width`, I think.

lib/awful/placement.lua
```diff
@@ -622,6 +653,65 @@ function placement.next_to_mouse(c, offset)
     return c:geometry({ x = x, y = y })
 end
 
+--- Resize the drawable to the cursor.
+--
+-- Valid args:
+--
+-- * *axis*:The axis (vertical or horizontal). If none is
+--    specified, then the drawable will be maximized on both axis.
```
psychon: maximized? Why maximized?

lib/awful/placement.lua
```diff
@@ -622,6 +653,65 @@ function placement.next_to_mouse(c, offset)
...
+
+    local h_only = args.axis == "horizontal"
+    local v_only = args.axis == "vertical"
+
+    -- To support both growing and shrinking the drawable, it is necessary
+    -- to decide to use either "north or south" and "east or west" directions.
+    -- Otherwise, the result will always be 1x1
+    local closest_corner = placement.closest_corner(capi.mouse, {
+        parent        = d,
+        pretend       = true,
```
psychon: The `pretend` option is only added two commits later...

lib/awful/placement.lua
Elv13: Oops, copy paste issue

lib/awful/placement.lua
```diff
@@ -679,7 +687,7 @@ function placement.resize_to_mouse(d, args)
     -- To support both growing and shrinking the drawable, it is necessary
     -- to decide to use either "north or south" and "east or west" directions.
     -- Otherwise, the result will always be 1x1
-    local closest_corner = placement.closest_corner(capi.mouse, {
+    local _, closest_corner = placement.closest_corner(capi.mouse, {
```
psychon: Unrelated change? Should this be part of the previous commit instead?

lib/awful/mouse/resize.lua
```diff
@@ -0,0 +1,175 @@
...
+        -- Execute the move callbacks. This can be used to add features such as
+        -- snap or adding fancy graphical effects.
+        for _, cb in ipairs(callbacks.move[context] or {}) do
+            -- If something is returned, assume it is a modified geometry
+            geo = cb(client, geo, args) or geo
+        end
+
+        if args.move_callback then
+            args.move_callback(client, geo, args)
+        end
```
psychon: Why no `geo = call() or geo` here, like you do in the `for`-loop above?

lib/awful/mouse/init.lua
```diff
@@ -26,7 +26,9 @@ local capi =
     mousegrabber = mousegrabber,
 }
 
-local mouse = {}
+local mouse = {
+    resize = require("awful.mouse.resize")
+}
```
psychon: Should this be part of the previous commit?

lib/awful/mouse/init.lua
```diff
@@ -154,10 +152,15 @@ end
 --- Move a client.
 -- @param c The client to move, or the focused one if nil.
 -- @param snap The pixel to snap clients.
--- @param finished_cb An optional callback function, that will be called
+-- @param finished_cb Deprecated, do not use
 --   when moving the client has been finished. The client
 --   that has been moved will be passed to that function.
```
psychon: This only removes half of the documentation for `finished_cb`.

lib/awful/mouse/snap.lua
```diff
@@ -67,7 +67,7 @@ end
 -- @param fixed_x True if the client isn't allowed to move in the x direction.
 -- @param fixed_y True if the client isn't allowed to move in the y direction.
 function module.snap(c, snap, x, y, fixed_x, fixed_y)
-    snap = snap or 8
+    snap = snap or snap.default_distance or 8
```
psychon: Shouldn't this be `module.default_distance`?

Elv13: Oops, rebase gone wrong

psychon: And: Can't you move the `8` default value into an explicit setting of this? So somewhere here (next to the documentation of it) you do `module.default_distance = 8` and in this function you remove the `or 8`.

lib/awful/mouse/snap.lua
```diff
@@ -119,5 +120,15 @@ function module.snap(c, snap, x, y, fixed_x, fixed_y)
     return geom
 end
 
+-- Enable edge snapping
+resize.add_move_callback(function(c, geo, args)
+    if args and (args.snap == nil or args.snap) then
+        local ngeo = module.snap(c, args.snap, geo.x, geo.y)
+        ngeo.x = ngeo.x + (2 * c.border_width)
+        ngeo.y = ngeo.y + (2 * c.border_width)
```
psychon: Huh? Why does this have to add the border width? That smells fishy to me.

lib/awful/mouse/snap.lua
```diff
@@ -20,6 +21,54 @@ local capi = {
...
+    end
+
+    if math.abs((sg.y + sg.height) - coords.y) <= snap then
+        v = "bottom"
+    end
+
+    return v, h
+end
+
+local current_snap = nil
```
psychon: Hrm. I don't really know how, but can we somehow get rid of this global variable? This is used as a magic way to get some argument from the move callback below into the leave callback...

lib/awful/mouse/init.lua
```diff
@@ -35,6 +35,18 @@ mouse.wibox = {}
...
 
+--- The snap outline background color.
+-- @beautiful beautiful.snap_bg
+-- @tparam color|string|gradient|pattern color
+
+--- The snap outline width.
+-- @beautiful beautiful.snap_border_width
+-- @param integer
+
+--- The snap outline width.
```
psychon: width -> shape

lib/awful/mouse/init.lua
```diff
@@ -276,6 +277,25 @@ capi.client.connect_signal("request::geometry", mouse.resize_handler)
 -- Set the cursor at startup
 capi.root.cursor("left_ptr")
 
+-- Implement the custom property handler
+local props = setmetatable({}, {__mode = "v"})
```
psychon: Why is this a weak table? `mouse.foo = {}` is allowed to silently be removed?

mouse.c
```diff
@@ -237,7 +237,9 @@ luaA_mouse_pushstatus(lua_State *L, int x, int y, uint16_t mask)
     return 1;
 }
 
-/** Get or set the mouse coords.
+/* Get or set the mouse coords.
+ *
+ *@DOC_awful_mouse_coords_EXAMPLE@
```
psychon: Can't you just replace this whole comment with `documented in lib/awful/mouse/init.lua`?

tests/examples/awful/mouse/coords.lua
```diff
@@ -0,0 +1,13 @@
+screen[1]._resize {x = 175, width = 128, height = 96} --DOC_HIDE
+mouse.coords {x=175+60,y=60} --DOC_HIDE
+
+ -- Get the position
+print(mouse.coords().x)
+
```
psychon: I know that I'm annoying, but: trailing whitespace

lib/awful/mouse/init.lua
```diff
@@ -280,6 +280,32 @@ function mouse.resize_handler(c, context, hints)
...
+--- Get the wibox currently under the mouse cursor.
+-- @property current_wibox
+-- @tparam wibox|nil The wibox
+
+function mouse.object.get_current_wibox()
+    local obj = capi.mouse.object_under_pointer()
+    if type(obj) == "drawin" then
+        return obj
+    end
+end
```
psychon: Sorry, but a drawin is not a wibox. For example, try calling `get_current_wibox():setup()` or `:set_widget()`. This really returns a drawin and thus this function is useless in 99% of cases.

Can't we just deprecate it without a replacement? :-(
If not, then this has to be called `get_current_drawin()` and needs a warning about a drawin not being a wibox etc.

tests/test-resize.lua
```diff
@@ -0,0 +1,412 @@
...
+            width  = 300,
+            height = 300,
+        }
+
+        return true
+    end
+end)
+
+table.insert(steps, function()
+    root.fake_input("button_press",1)
```
psychon: Why this button press? I think it's so that `client.resize` works (it wants a button to be down). Could you add a comment explaining this (and perhaps move it down a bit to right before the call to `mouse.coords`)?

tests/test-resize.lua
```diff
@@ -0,0 +1,412 @@
...
+-- Shirnk the client by 100px from the top right
+table.insert(steps, function()
+
+    local c = client.get()[1]
+
+--             print(c:geometry().width)
+    assert(c:geometry().x == 100)
+    assert(c:geometry().y == 200)
+--     assert(c:geometry().width == 200-2*c.border_width) --FIXME off by border width...
+--     assert(c:geometry().height == 200-2*c.border_width) --FIXME off by border width...
```
psychon: Hm? Is this left-over from some debugging and can be removed?

tests/test-resize.lua
```diff
@@ -0,0 +1,412 @@
...
+
+    mouse.coords {x = 300, y= 200}
+
+    return true
+end)
+
+--TODO test bottom left
```
psychon: What's with this?

tests/test-resize.lua
```diff
@@ -0,0 +1,412 @@
...
+
+-- Stop the resize
+table.insert(steps, function()
+    root.fake_input("button_release",1)
+    root.fake_input("button_release",1)
+    root.fake_input("button_release",1)
+
+--     if not mousegrabber.isrunning then --FIXME it should work, but doesn't
+--         return true
+--     end
```
psychon: Why releasing three times? Why this FIXME?

Elv13: There is something wrong. The resize should stop when the button is released, but it doesn't. I forgot about that FIXME, but it may be that the mousgrabber callback is called only when the mouse move, not when there is press/release (untested theory)

psychon:
I looked through the commits leading to fa08a46a8db41669be014b8f6cbeaac1a42f4b21. I hope that I don't have to do this again. Tell me if I should look at something specifically again.

tests/test-resize.lua
psychon: Well, `fake_input` only causes some input to be generated, but doesn't wait for it to be handled (it's asynchronous). Only the next event loop iteration will see the mouse release and then stop the keygrabber. Thus, I guess this needs another step / another entry in the `steps` array.

Elv13:
Yeah, sorry about some of the obvious mistakes. I should have done one last review after hitting my coverage targets instead of relying on them. Also, this PR has grown way too big, sorry about that.

tests/test-resize.lua
Elv13: I don't think so, a previous iteration of this step didn't return anything unless it hit the "return true" in the `if`, so it caused a new loop iteration and it still didn't release the grabber. I may have missed something. I will look into this later.

tests/examples/awful/mouse/coords.lua
Elv13: Actually, this is done on purpose. CMake drop empty lines and convert lines with trailing spaces into empty lines (not my idea, its how it behave).

psychon: That's... bad. Eventually this trailing whitespace will be removed by someone.
However, feel free to try to leave it in for now.

Elv13:
Updated

Commits 688021d, b5a1a8d and 3dab42e are new. I actually had those things in the TODO above, but almost forgot to commit them. It would have been bad...

Other than that, all issues raised above have been fixed but:

 * The local variable used to keep the current snapping position (no idea how to make it *really* cleaner)
 * Some comments/FIXME in `test-resize.lua` (they need some more investigation)

psychon:
No (real) complaints about those three commits, so :+1:

Elv13:
I did more tests today. It seem to work. I am sure I missed a bunch of corner cases and maybe an obvious one, but that can be fixed as issue appear. (note to myself, stop doing 2.5k SLOC PRs, they are a pita)

PR #847, opened 2016-04-24
Add clear() function to awful.widget.graph and doc fixes

popeye123:
1. There hasn't been an easy way to clear a graph widget. `awful.widget.graph.clear()` provides an easy way to clear the graph.
2. `awful.widget.graph.add_value()` did not show up in the generated HTML documentation.

### DIFF
--- a/lib/awful/widget/graph.lua
+++ b/lib/awful/widget/graph.lua
@@ -161,8 +161,9 @@ function graph.fit(_graph)
 end
 
 --- Add a value to the graph
--- @param _graph The graph.
--- @param value The value between 0 and 1.
+--
+-- @function add_value
+-- @param value The value to be added to the graph
 -- @param group The stack color group index.
 local function add_value(_graph, value, group)
     if not _graph then return end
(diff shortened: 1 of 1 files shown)


lib/awful/widget/graph.lua
```diff
@@ -197,6 +198,16 @@ local function add_value(_graph, value, group)
     return _graph
 end
 
+--- Clear the graph.
+--
+-- @function clear
+local function clear(_graph)
```
blueyed: Should/can this use the same style as below? (`function graph:clear()`)

lib/awful/widget/graph.lua
```diff
@@ -161,8 +161,9 @@ function graph.fit(_graph)
 end
 
 --- Add a value to the graph
--- @param _graph The graph.
--- @param value The value between 0 and 1.
+--
+-- @function add_value
+-- @param value The value to be added to the graph
 -- @param group The stack color group index.
 local function add_value(_graph, value, group)
```
blueyed: Could/should also get changed to `function graph:add_value(value, group)`.

psychon:
I didn't look too closely, but seems fine to me: :+1:

blueyed:
Thanks!

PR #849, opened 2016-04-25
fix(awful: rules): floating property

actionless:


psychon:
Heh, that's an obvious error, but only once you know about it :+1:

Elv13:
whoops, I missed one

blueyed:
Would be nice to have a test for this.

@actionless 
Do you feel like diving into the test framework to add one?

Issue #813, opened 2016-04-07
awful.placement.no_overlap does not consider current screen (was: mouse.screen can be wrong)


blueyed:
I've noticed today, that `mouse.screen` (used via `awful.screen.focused()`, but also in `awesome-client`) is often wrong for me: it returns 2 for the index, although the mouse pointer is on screen 1 (screen.primary).

I have not investigated further.

blueyed:
@psychon 
Any idea / hints for me to look at/for?

psychon:
Nope, nothing good really. Some bad ifrsd:
- How easy can you reproduce this?
- What are `mouse.coords()` when `mouse.screen` is wrong? (plus what's the geometry of all the screens)

blueyed:
Seems to be reproducible at work always, but it does not seem to be about `mouse.screen` being wrong.

`screen.primary.index` is 1 (the laptop screen).
The other screen is above.
```
mouse.coords():
  x : 2181 (number),
  y : 2345 (number)
mouse.screen.index = 1
```

Therefore that's OK.

But programs will open at screen 2 (the one above), and not where the mouse is.

`xrandr -q`:
```
Screen 0: minimum 8 x 8, current 2688 x 2592, maximum 32767 x 32767
eDP1 connected 1920x1080+768+1512 (normal left inverted right x axis y axis) 276mm x 156mm
DP2-2 connected 2688x1512+0+0 (normal left inverted right x axis y axis) 521mm x 293mm panning 2688x1512+0+0
```

psychon:
The "place new client's on mouse's screen"-stuff is handled by `lib/awful/tag.lua`, connecting to the "manage" signal. It does `c.screen = ascreen.focused()`. Perhaps you can add some `print(c.screen)` before and after this to see if it worked? Also, `client.connect_signal("property::screen", function(c) print(debug.traceback(c.name .. " now on " .. (c.screen and c.screen.index or "no screen")) end)` might be really spammy, but might also be helpful.

blueyed:
See https://github.com/awesomeWM/awesome/pull/852 for a fix and the explanation.

PR #852, opened 2016-04-25
awful.placement.no_overlap: use screen.workarea as fallback

blueyed:
When there is a maximized and floating client,
`awful.placement.no_overlap` would end up with an empty list of areas to
place the client into.

This patch fixes it to use the default `screen.workarea` in that case.

Closes https://github.com/awesomeWM/awesome/issues/813.

### DIFF
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -555,8 +555,13 @@ function placement.no_overlap(c)
     end
 
     -- We did not find an area with enough space for our size:
-    -- just take the biggest available one and go in
+    -- just take the biggest available one and go in.
+    -- This makes sure to have the whole screen's area in case it has been
+    -- removed.
     if not found then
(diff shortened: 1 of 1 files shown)


psychon:
Fine by me. Let's see what @Elv13 thinks.
(Also: Nice find!)

psychon:
How does this explain #813? When all the space is used, `no_overlap` will not move the client but just leave it where it already is. Thus, this doesn't seem like an explanation for #813.

blueyed:
Good point.
I cannot reproduce it at home with two screens of the same size (1920x1080).
I could imagine that `awful.placement.no_offscreen` is also involved.
IIRC from the `print` statements the original position was on the other screen though (number 2, but above the laptop).

Elv13:
I also had issues with that method. +1 from me.

PR #853, opened 2016-04-27
feat(lib: menubar): replace io.popen to lgi

actionless:


psychon:
All the API docs still need to be updated (see the travis failure) and my error handling with `error(err)` and `print(tostring(err))` might need to be done better. Perhaps `gears.debug.print_error()`?
Thanks for writing this!
(The other comment I'll leave as an inline comment)

actionless:
@psychon updated

actionless:
just realized what it was breaking recursive scan of dirs, like `/usr/share/applications/kde`, fixed it now

lib/menubar/menu_gen.lua
```diff
@@ -83,48 +83,60 @@ local function trim(s)
 end
 
 --- Generate an array of all visible menu entries.
--- @treturn table All menu entries.
-function menu_gen.generate()
+-- @tparam function callback Will be fired when all menu entries will be parsed.
```
psychon: ...when all menu entries *were* parsed.

Also, I wouldn't mind adding `with the resulting list of menu entries as argument`.

lib/menubar/menu_gen.lua
```diff
@@ -83,48 +83,60 @@ local function trim(s)
...
+                            unique_entries[unique_key] = true
+                        end
                     end
                 end
             end
-        end
+            dirs_parsed = dirs_parsed + 1
+            if dirs_parsed == #menu_gen.all_menu_dirs then
+                callback(result)
+            end
```
psychon: This part is missing on the initial `if not entries then dirs_parsed = dirs_parsed + 1`.
I suggest replacing this initial check with `entries = entries or {}`. That would get rid of this special case.

lib/menubar/utils.lua
```diff
@@ -243,17 +246,49 @@ end
...
 -- @tparam string dir The directory.
--- @treturn table Paths of found .desktop files.
-function utils.parse_dir(dir)
-    local programs = {}
-    local files = io.popen('find '.. dir .." -name '*.desktop' 2>/dev/null")
-    for file in files:lines() do
-        local program = utils.parse_desktop_file(file)
-        if program then
-            table.insert(programs, program)
+-- @tparam function callback Will be fired when all the files will be parsed.
```
psychon: Same comment as before: *were* parsed and it might be nice to explain the argument a bit (or at least mention it)

lib/menubar/utils.lua
```diff
@@ -243,17 +246,49 @@ end
...
+            debug.print_error(err)
+            return callback(nil)
         end
-    end
-    return programs
+        local subdirs_to_parse = 0
+        local files_per_call = 100 -- Actual value is not that important
+        local programs = {}
+        while true do
+            local list, enum_err = enum:async_next_files(files_per_call, glib.PRIORITY_DEFAULT)
```
psychon: Can you remove this `glib.PRIORITY_DEFAULT` argument and see if this still works? Thanks to you I read the lgi docs yesterday and they say that cancellation and priority arguments are skipped on these `async_foo` functions. Thus, I guess that this part isn't actually needed here.

lib/menubar/utils.lua
```diff
@@ -243,17 +246,49 @@ end
...
+                local file_type = info:get_file_type()
+                local file_path = f:get_child(info:get_name()):get_path()
+                if file_type == 'REGULAR' then
+                    local program = utils.parse_desktop_file(file_path)
+                    if program then
+                        table.insert(programs, program)
+                    end
+                elseif file_type == 'DIRECTORY' then
+                    subdirs_to_parse = subdirs_to_parse + 1
+                    utils.parse_dir(file_path, function(more_programs)
```
psychon: Hm, could you restructure this? This `utils.parse_dir` starts another async operation. It would be nice if this were done inside of this async operation instead. Otherwise, we could run out of file descriptors and things will start to fail with "too many open FDs" (I had this yesterday when I wrote something that recursively lists files).

To do this, I would suggest to move the function that you call `gio.Async.start` on into a local function of this file that gets `dir` as an argument (and another argument with the table where entries should be added to).

That way, you can replace `utils.parse_dir` with `gio.Async.start(function() local result = {} your_local_function(dir, result) callback(result) end)()` and this recursive call can be replaced with a plain `your_local_function(file_path, result)`.

Another benefit: I think you no longer have to count `subdirs_to_parse` and the invocation of the callback is a lot simpler, too.

lib/menubar/utils.lua
```diff
@@ -243,17 +246,49 @@ end
...
+                        subdirs_to_parse = subdirs_to_parse - 1
+                        if more_programs then
+                            awful_util.table.merge(programs, more_programs)
+                        end
+                    end)
+                end
+            end
+            if #list == 0 and subdirs_to_parse == 0 then
+                return callback(programs)
```
psychon: For the above suggestion (and so that the following works): Please replace the `return`s inside of the loop with `break`.

lib/menubar/utils.lua
```diff
@@ -243,17 +246,49 @@ end
...
+            end
+        end
```
psychon: Please add `enum:async_close()` here (after the `while true`-loop). When writing the code yesterday I thought "the garbage collector will clean up". However, this file enumeration keeps a file descriptor open and there is an upper limit on the number of file descriptors that can be open. Thus, this should be closed when done.

psychon:
Sorry for having so many "complaints" about this. I should have been more careful yesterday (I only noticed later the problem with the open file descriptors when recursively and parallely listing all files in `/proc`).

lib/menubar/utils.lua
```diff
@@ -243,17 +246,49 @@ end
...
+        local programs = {}
+        while true do
+            local list, enum_err = enum:async_next_files(files_per_call, glib.PRIORITY_DEFAULT)
+            if enum_err then
+                debug.print_error(enum_err)
+                return callback(nil)
+            end
+            for _, info in ipairs(list) do
+                local file_type = info:get_file_type()
+                local file_path = f:get_child(info:get_name()):get_path()
```
psychon: Oops. I just found https://developer.gnome.org/gio/stable/GFileEnumerator.html#g-file-enumerator-get-child which sounds like the above code can be replaced with `local file_path = enum:get_child(info):get_path()`

actionless:
@psychon hopefully done

lib/menubar/menu_gen.lua
psychon: This comment still applies....

lib/menubar/menu_gen.lua
```diff
@@ -83,48 +83,61 @@ local function trim(s)
...
-                        for _, category in pairs(entry.categories) do
-                            local cat_key, cat_use =
-                            get_category_name_and_usage_by_type(category)
-                            if cat_key and cat_use then
-                                target_category = cat_key
-                                break
+        utils.parse_dir(dir, function(entries)
+            if not entries then
+                dirs_parsed = dirs_parsed + 1
+                return
```
psychon: ....here

lib/menubar/utils.lua
```diff
@@ -242,18 +245,51 @@ function utils.parse_desktop_file(file)
...
+        end
+        enum:async_close()
     end
-    return programs
+
+    gio.Async.start(function()
+        local result = {}
+        parser(dir_path, result)
+        callback(result)
+    end)()
```
psychon: Looks good, thanks!

lib/menubar/menu_gen.lua
actionless: my bad, missed this one

psychon:
Thanks. :+1:

PR #854, opened 2016-04-27
refactor(awful: util): use lgi.Gio for is_dir

actionless:


psychon:
:+1:

lib/awful/util.lua
```diff
@@ -253,16 +253,7 @@ end
...
-    if file then
-        if not file:read(0) -- Not a regular file (might include empty ones).
-            and file:seek("end") ~= 0 then  -- And not a file with size 0.
-            io.close(file)
-            return true
-        end
-        io.close(file)
-    end
-    return false
+    return Gio.File.new_for_path(path):query_file_type({}) == "DIRECTORY"
```
actionless: should i free it before returning?
https://developer.gnome.org/gio/stable/GFile.html#g-file-new-for-path
> Returns
> a new GFile for the given path . Free the returned object with g_object_unref().

psychon: No, that's ok. I don't know if lgi even let's you access this function. However, a `GFile` only contains a path name. It doesn't keep the file open or anything like that. Thus, this problem doesn't apply here.
(It's similar to "always remember to call `io.close` on file handles in Lua)

actionless: but how did you get what i have to free only enum? is that from lgi code or docs?

psychon:
Try and error. After closing (well, closing is something else than freeing/unref!) I stopped getting the "all FDs used" error.

I guess some key point is the following from https://developer.gnome.org/gio/stable/GFileEnumerator.html#g-file-enumerator-close (no such text is present elsewhere):

> This will be automatically called when the last reference is dropped, but you might want to call this function to make sure resources are released as early as possible.

lib/awful/util.lua
psychon: Try and error. After closing (well, closing is something else than freeing/unref!) I stopped getting the "all FDs used" error.

I guess some key point is the following from https://developer.gnome.org/gio/stable/GFileEnumerator.html#g-file-enumerator-close (no such text is present elsewhere):

> This will be automatically called when the last reference is dropped, but you might want to call this function to make sure resources are released as early as possible.

PR #855, opened 2016-04-28
Fix leak test

Elv13:
So, I did some tests locally and could reproduce the issue. It is not just luajit, the normal lua also eventually hit the bug that has been causing build issues for some weeks.

Apparently, it was just a question of time. if you rename the test `test-aaaaaaa-leaks.lua`, it will **never** hit the problem, but if it is being run later, it gets more and more common (~15% for luajit, ~3% for normal lua).

The problem is simply that the widget is a singleton and is for some reasons kept (it __is__ leaking). There was also a new leak when screen are removed. Finally, this was yet another useless `attached_connect_signal`.

So, this is it. I think (after 100 test run) that this is fixed once and for all.

Edit: The clear method is left undocumented on purpose, it has no value beyond the tests as the "kv" mode should now handle the only relevant use case (removing screens).

### DIFF
--- a/lib/awful/widget/layoutbox.lua
+++ b/lib/awful/widget/layoutbox.lua
@@ -8,10 +8,9 @@
 ---------------------------------------------------------------------------
 
 local setmetatable = setmetatable
-local capi = { screen = screen }
+local capi = { screen = screen, tag = tag }
 local layout = require("awful.layout")
 local tooltip = require("awful.tooltip")
-local tag = require("awful.tag")
 local beautiful = require("beautiful")
 local imagebox = require("wibox.widget.imagebox")
 
@@ -47,9 +46,9 @@ function layoutbox.new(screen)
 
     -- Do we already have the update callbacks registered?
     if boxes == nil then
-        boxes = setmetatable({}, { __mode = "v" })
-        tag.attached_connect_signal(nil, "property::selected", update_from_tag)
-        tag.attached_connect_signal(nil, "property::layout", update_from_tag)
+        boxes = setmetatable({}, { __mode = "kv" })
+        capi.tag.connect_signal("property::selected", update_from_tag)
+        capi.tag.connect_signal("property::layout", update_from_tag)
         layoutbox.boxes = boxes
     end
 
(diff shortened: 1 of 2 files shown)


lib/awful/widget/layoutbox.lua
```diff
@@ -47,9 +46,9 @@ function layoutbox.new(screen)
 
     -- Do we already have the update callbacks registered?
     if boxes == nil then
-        boxes = setmetatable({}, { __mode = "v" })
-        tag.attached_connect_signal(nil, "property::selected", update_from_tag)
-        tag.attached_connect_signal(nil, "property::layout", update_from_tag)
+        boxes = setmetatable({}, { __mode = "kv" })
```
psychon: Sigh, I did the same change in #832 and thus will have to do another rebase if this is merged....

tests/test-leaks.lua
```diff
@@ -63,6 +62,12 @@ collectable(wibox.layout.align.horizontal())
 collectable(awful.widget.launcher({ image = cairo.ImageSurface(cairo.Format.ARGB32, 20, 20), command = "bash" }))
 collectable(awful.widget.prompt())
 collectable(awful.widget.textclock())
+
+-- Test the layout box. As its widgets are a singleton (per screen), clear them
+-- first.
+awful.widget.layoutbox.clear()
+collectgarbage("collect")
+collectgarbage("collect")
```
psychon: Why the extra `collectgarbage` here? Also, does this really fix the leak? Apparently the old layoutbox is un-GC-able (perhaps something in the widget stuff keeps a reference to it?). This fixes the test by just creating a new layoutbox and ignoring that the previous one couldn't be collected. I'd rather figure out what references the layoutbox.

The only other test that does something with layoutboxes seems to be `test-benchmark.lua`, but it uses the same `create_wibox()` that this test already uses, so that shouldn't cause any leaks....

psychon:
Sorry, but this feels more like hiding the problem than fixing it. Sadly, I don't have any good suggestions on where that "fake" reference is coming from.

Thanks for the hint that this seems to be some interaction with other tests, I'll try to investigate based on that. For the time being: Sorry, I don't like this fix.

psychon:
I didn't manage to reproduce the leak here in 2000 runs of the test (running things like `./run.sh $(for i in $(seq 1 20) ; do echo test-leaks.lua ; done) test*.lua $(for i in $(seq 1 1000) ; do echo test-leaks.lua ; done)`). Hrm.... Ideas?

Issue #814, opened 2016-04-07
BUG: Reference not found: -1 via luaA_class_disconnect_signal_from_stack


blueyed:
The following showed up when resuming from a two-monitor setup into a single-monitor/undocked state:

The "SCREEN_CONNECT" is printed by a custom signal handler.

```
SCREEN_CONNECT: 1
2016-04-07 21:09:33 W: awful: function disconnect_for_each_screen is deprecated, see Use awful.screen.disconnect_for_each_screen.
stack traceback:
        /usr/share/awesome/lib/awful/util.lua:42: in function 'awful.util.deprecate'
        /usr/share/awesome/lib/gears/screen.lua:23: in function 'gears.screen.disconnect_for_each_screen'
        /home/user/.config/awesome/rc.lua:1247: in main chunk
2016-04-07 21:09:33 W: awesome: luaA_object_decref:130: BUG: Reference not found: -1 0x21c73d0
/usr/bin/awesome(backtrace_get+0x54) [0x425445]
/usr/bin/awesome(luaA_object_decref+0x6a) [0x4278c1]
/usr/bin/awesome(luaA_class_disconnect_signal_from_stack+0x3ff) [0x426f2c]
/usr/bin/awesome() [0x43f851]
/usr/lib/liblua.so.5.3(+0xddba) [0x7fe90686ddba]
/usr/lib/liblua.so.5.3(+0x1ad6d) [0x7fe90687ad6d]
/usr/lib/liblua.so.5.3(+0xe1af) [0x7fe90686e1af]
/usr/lib/liblua.so.5.3(+0xe201) [0x7fe90686e201]
/usr/lib/liblua.so.5.3(+0xd5c2) [0x7fe90686d5c2]
/usr/lib/liblua.so.5.3(+0xe48d) [0x7fe90686e48d]
/usr/lib/liblua.so.5.3(lua_pcallk+0x81) [0x7fe906869c41]
/usr/bin/awesome() [0x4180cf]
/usr/bin/awesome(luaA_parserc+0x60) [0x41b8f0]
/usr/bin/awesome(main+0xbb3) [0x40f91a]
/usr/lib/libc.so.6(__libc_start_main+0xf0) [0x7fe90628e710]
/usr/bin/awesome(_start+0x29) [0x40e819]
```
master (6e8337e) with the following patch only:

```diff
diff --git i/lib/gears/wallpaper.lua w/lib/gears/wallpaper.lua
index ca4e4d8..0b25906 100644
[...]

PR #856, opened 2016-04-28
Fix wallpaper setting after RandR monitor changes

psychon:
[Please note: This pull request targets the 3.5 branch, not master!]

A co-worker of mine has a laptop with a small display and two larger displays.
When enabling the larger displays (and disabling the built-in display), the
wallpaper was set incorrectly. Only the size of the old, smaller screen actually
contained the wallpaper and the left was black.

The reason for this is that the code directly draws the new wallpaper to the
pixmap containing the old wallpaper. However, because the old display was
smaller, this pixmap isn't as large as the screen and so couldn't contain all of
the new wallpaper.

Fix this by explicitly creating a surface of the needed size and copying the old
wallpaper into it (because we might be setting only the wallpaper of some
screen, not all of them).

Special-thanks-to: vsp

### DIFF
--- a/lib/gears/wallpaper.lua.in
+++ b/lib/gears/wallpaper.lua.in
@@ -40,6 +40,14 @@ local function prepare_wallpaper(s)
         -- No wallpaper yet, create an image surface for just the part we need
         img = cairo.ImageSurface(cairo.Format.RGB24, geom.width, geom.height)
         img:set_device_offset(-geom.x, -geom.y)
+    else
+        -- Some wallpaper yet, copy it to a new surface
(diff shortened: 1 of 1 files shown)


blueyed:
So master is not affected?
Sounds like the issue I am seeing (where the patch mentioned in https://github.com/awesomeWM/awesome/issues/814 helps).
See the comments at https://github.com/awesomeWM/awesome/commit/b05b699 also.

psychon:
A similar change (and some more) was done for master in 112ce3b419245bdae8e12252415c0da9d3059349. It also uses `create_similar()` etc.

PR #858, opened 2016-04-28
Wallpaper function

psychon:
This changes the default config so that `beautiful.wallpaper` can be a function that gets the screen as its argument (so that it can e.g. choose a different wallpaper for VGA1 and LVDS1 or so that a wallpaper with the correct resolution is used).

The following commit changes the xresources theme to make use of that and create a wallpaper only when needed (no permanent references to the wallpaper) and to use a RecordingSurface instead of an ImageSurface.

CC @actionless

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -168,7 +168,12 @@ mytasklist.buttons = awful.util.table.join(
 awful.screen.connect_for_each_screen(function(s)
     -- Wallpaper
     if beautiful.wallpaper then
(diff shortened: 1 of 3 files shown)


actionless:
oops, just realized what github reactions are not very useful for PR reviewing since submitter is not receiving the notification :)

blueyed:
I could imagine this being enhanced in the future by counting number of up/down thumbs on the original comment though.

psychon:
Rebased on master

Issue #826, opened 2016-04-14
How to override get_gap and set_gap methods after objects' properties refactoring?


actionless:
i was using this code to control useless gaps behavior:

```
  awful.tag.getgap = function(t)
    t = t or awful.tag.selected()
    if #awful.client.tiled(awful.tag.getscreen(t)) == 1 and awful.tag.getmfpol(t) == "expand" then
        return 0
    end
    return awful.tag.getproperty(t, "useless_gap") or beautiful.useless_gap or 0
  end
```

or mb we can just merge that behaviour as a default one?

psychon:
I think that setting `awful.tag.object.get_gap` to your function should work.

actionless:
thanks!

PR #860, opened 2016-04-30
awful.tag: Add `gap_count` property

Elv13:
This restore a feature removed during the API cleanup.

This still need a better name

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -41,6 +41,16 @@ data.tags = setmetatable({}, { __mode = 'k' })
 tag.history = {}
 tag.history.limit = 20
 
+-- Default values
+tag.defaults = {}
+
+tag.defaults.gap                 = 0
+tag.defaults.gap_count           = 0
+tag.defaults.master_fill_policy  = "expand"
+tag.defaults.master_width_factor = 0.5
+tag.defaults.column_count        = 1
+tag.defaults.master_count        = 1
+
 -- screen.tags depend on index, it cannot be used by awful.tag
 local function raw_tags(scr)
     local tmp_tags = {}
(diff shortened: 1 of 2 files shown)


Elv13:
@zainin @actionless This is totally untested and will require a better name please test.

> and what do you think about also checking awful.tag.getmfpol(t) == "expand" ?

@actionless Can you elaborate? You mean being able to disable the gaps differently when the mfpol is set? In that case, wont that also affect the "vertical" spacing in an unwanted way?

lib/awful/tag.lua
```diff
@@ -795,6 +797,32 @@ function tag.incgap(add, t)
...
+-- @beautiful beautiful.useless_gap_count
+-- @param integer (default: 0)
+-- @see gap
+-- @see gap_count
+
+--- The minimum number of client at which to enable gaps.
+--
+-- **Signal:**
+--
+-- * *property::gap\_count*
```
psychon: Isn't it `property::useless_gap_count` due to using a different name internally?

lib/awful/tag.lua
```diff
@@ -55,6 +65,39 @@ end
 
 --- The number of elements kept in the history.
 -- @tfield integer awful.tag.history.limit
+-- @tparam[opt=20] integer limit
+
+--- The gap between clients (in points).
+-- @tfield[opt=0] integer awful.tag.defaults.gap
+-- @tparam[opt=0] integer gap
+-- @see gap
+-- @see gap_count
```
psychon: Can't you move these docs directly to the variable definition above?
```
--- Foo is a great value
-- @field[opt=42] awful.tag.defaults.foo
tag.foo = 42
```

psychon:
Name... uhm... this does `minimum_clients_for_useless_gaps`, but that's quite a long name. When does anyone want to set this to something else than 0 or 1? (Actually, the implementation uses `>`, not `>=`, so the only interesting values would be -1 and 0, but -1 cannot be set, right?) If this were turned into a boolean, it could be `useless_gaps_ignore_single_client` or something like that. Some more random rambling: `useless_gaps_require_multiple_clients`, `useless_gaps_for_single_client` (this one being an integer that is used when just one client is visible).... hm, this is hard

actionless:
@Elv13 
> Can you elaborate? 
see example here: https://github.com/awesomeWM/awesome/issues/826 or mb better copy it to the config and see it in action -- the idea is what when master_fill_policy is set to mwfact when window have big gaps on the sides, and since globally gaps are enabled it looks a bit illogical what that single window will have only side gaps but not upper/bottom small ones

zainin:
@Elv13 
Works fine, I've found no issues.

@psychon 
A boolean sounds like the best choice. Like you've said, it's unlikely someone would want >1 minimum clients.

Elv13:
Obseleted by #1066

PR #862, opened 2016-04-30
Import some docs

psychon:
For https://github.com/awesomeWM/awesome-www/issues/7, @Elv13 suggested to move some parts from the wiki into ldoc format. This pull request moves all the usable stuff that I found into awesome's source. That's just the FAQ (but cleaned up and fixed) and the "how to do a release"-page (but this one is not hooked up with ldoc since it is mainly useful to me).

Is this the right direction?

For some slightly off-topic question: What other wiki pages would you like to keep?

Edit: As an overview, here is a list of all pages in the wiki: https://gist.github.com/psychon/fb068540d02aa84741a255d8598ad68c (sorted into three groups: "everything else", "how to install on distro X" and "translations")

### DIFF
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -21,6 +21,7 @@ topics={
     '02-contributing.md',
     '03-declarative-layout.md',
     '04-new-widgets.md',
+    '90-FAQ.md',
 }
 
 -- Setup @client to be an alias for "@tparam client.object"
--- a/docs/90-FAQ.md
+++ b/docs/90-FAQ.md
@@ -1,0 +1,392 @@
+# FAQ
+
+## General
+
+### Why call it awesome?
+
+The name *awesome* comes from the English word *awesome* often used by the
+character [Barney Stinson](http://en.wikipedia.org/wiki/Barney_Stinson)
+from the TV series HIMYM.
+
+
+## Configuration
+
+### How to change the default window management layout?
+
+In the default configuration file one layout is set for all tags, it happens to
+be the `floating` layout. You can change that by editing your tag creation
+loop in the `rc.lua`:
+
+    -- Each screen has its own tag table.
+    awful.tag({ "1", "2", "3", "4", "5", "6", "7", "8", "9" }, s, awful.layout.layouts[1])
+
+Notice that all tags will use the 1st layout from your layouts table, which is
+defined right before tags are created. Just change the layout number in order to
+use another window management layout.
+
+#### How to change the name and layout per tag?
+
+You can modify your tag section, there are many possible implementations, here
+is a simple one:
+
+At the beginning of `rc.lua`:
+
+    layouts = awful.layout.layouts
+    tags = {
+      names  = { "www", "editor", "mail", "im", "rss", 6, 7, "rss", "media"},
+      layout = { layouts[2], layouts[1], layouts[1], layouts[4], layouts[1],
+                 layouts[6], layouts[6], layouts[5], layouts[6]
+    }}
+
+Then later to create tags:
+
+    tags[s] = awful.tag(tags.names, s, tags.layout)
+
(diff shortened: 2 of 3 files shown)


docs/config.ld
```diff
@@ -21,6 +21,7 @@ topics={
     '02-contributing.md',
     '03-declarative-layout.md',
     '04-new-widgets.md',
+    '99-FAQ.md',
```
blueyed: Better use 90 here to have some room for later?!

psychon: Sure, I can do that. However, if we need to add something later we can just rename this file then. No idea what a good order would actually be...

psychon:
I re-read this and the only things I found are https://gist.github.com/psychon/9fde97ceb489034b34801da4f430cca7 (already squashed into the commit)

Elv13:
This depend how many of these we want. Should every `tag`, `screen` and `client` (and `tag`s and `client`s) references should be a link? This would create a lot of them, but also, as the FAQ is more useful to new users than you and I, it may make sense anyway.

blueyed:
More references are good in general, yes.
But then it's probably enough to only do so for the first entry in a section/paragraph?!

I've only skimmed it, but think it's good.
I've restarted the apidoc Travis job, hopefully the bot will kick in (and the diff will be useful).

From https://github.com/psychon/awesome/blob/import-some-docs/docs/90-FAQ.md it looks good.

For keyboard input the `<kbd>` tag could be used, e.g. <kbd>Mod4+Ctrl+r</kbd> or <kbd>Mod4-Ctrl-r</kbd>.

Elv13:
This can be done later. I will merge this now before it bitrot.

psychon:
Thanks & sorry, but I didn't yet have the time to try `<kbd>`. I didn't even find any docs saying that it should work.

Also: Which other pages from the wiki are useful and should be moved into our docs? Which pages are useful and should be moved to $NEW_WIKI?

Elv13:
Umm, http://awesomewm.github.io/apidoc/ wasn't updated. Nor for my last PR. It still show last week doc.

@blueyed any idea?

psychon:
https://travis-ci.org/awesomeWM/awesome/jobs/128145153

> [...]
patching file libraries/wibox.widget.base.html
[merged-update 139b3d4] [relevant] Update docs for master-g2d4b8d8 via Travis
 105 files changed, 1557 insertions(+), 858 deletions(-)
 delete mode 100644 classes/drawin.html
 create mode 100644 documentation/90-FAQ.md.html
[merged-update 62df4b6] [boilerplate] Update docs for master-g2d4b8d8 via Travis
 102 files changed, 186 insertions(+), 186 deletions(-)
HEAD is now at db76232 Update docs for master-g612756f via Travis
[merged-update bb4b684] [boilerplate] Update docs for master-g2d4b8d8 via Travis
 102 files changed, 186 insertions(+), 186 deletions(-)
[merged-update fd8da88] [relevant] Update docs for master-g2d4b8d8 via Travis
 105 files changed, 1557 insertions(+), 858 deletions(-)
 delete mode 100644 classes/drawin.html
 create mode 100644 documentation/90-FAQ.md.html
Deleted tag '_old' (was 62df4b6)
Switched to branch 'gh-pages'
Your branch is up-to-date with 'origin/gh-pages'.

Elv13:
Well, its not there http://awesomewm.github.io/, the FAQ URL still 404 in North America (and probably everywhere else).

psychon:
The "relevant" commit says that 105 files were changed, so something goes wrong later, I guess...?

blueyed:
Yes, something broke: https://github.com/awesomeWM/apidoc/branches are not being updated.
I'll look into it later.

blueyed:
The `set -e` caused it to abort with the new calls to `grep` which have not found a pattern.. :/

PR #866, opened 2016-05-01
Travis: use new GH_APIDOC_TOKEN

blueyed:
While still being used via Travis' `secure` env mechanism, we could
additionally set it through an env for the repo (not displayed by
default), and revoke/regenerate it when it gets revealed.
The token provides access only to the `apidoc` repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ env:
     # Lua 5.3 isn't available in Ubuntu Trusty, so some magic below installs it.
     - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so
   global:
-    # Secure token to push to gh-pages.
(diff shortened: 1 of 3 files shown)


psychon:
Sorry, what? I don't understand this, you are just changing the name of the variable?

blueyed:
The name and contents was changed.
It is still used via Travis' `secure` which makes it available in the build.
But I have also set it for all builds in general, which makes it available for PRs.
This could be exploited, but we should be able to see it from a PRs diff.
The worst that could happen would be somebody changing the apidoc repo in the name of awesome-robot.
I've even removed awesome-robot from the awesomeWM Github organization to keep it more separate.

psychon:
Oh, ok. :+1:
How did you make it available to all PRs? This doesn't seem to be visible in the diff.

blueyed:
https://travis-ci.org/awesomeWM/awesome/settings

Issue #861, opened 2016-04-30
Testsuite leaves running processes behind


psychon:
After running `make check`:
```
$ ps aux | grep lua
psychon  17137  0.4  1.9 423616 38684 ?        SNsl 09:15   0:00 lua -e local lgi = require 'lgi';local Gtk = lgi.require('Gtk');Gtk.init();local class = 'rule0000000048';local window = Gtk.Window {    default_width  = 100,    default_height = 100,    title          = 'Foo'};window:set_wmclass(class, class);local app = Gtk.Application {    application_id = 'org.awesomewm.tests.rule0000000048'};function app:on_activate()    window.application = self;    window:show_all();end;app:run {''}
psychon  17143  0.4  1.9 423616 38688 ?        SNsl 09:15   0:00 lua -e local lgi = require 'lgi';local Gtk = lgi.require('Gtk');Gtk.init();local class = 'rule0000000051';local window = Gtk.Window {    default_width  = 100,    default_height = 100,    title          = 'Foo'};window:set_wmclass(class, class);local app = Gtk.Application {    application_id = 'org.awesomewm.tests.rule0000000051'};function app:on_activate()    window.application = self;    window:show_all();end;app:run {''}
psychon  17145  0.4  1.9 423744 39100 ?        SNsl 09:15   0:00 lua -e local lgi = require 'lgi';local Gtk = lgi.require('Gtk');Gtk.init();local class = 'rule0000000052';local window = Gtk.Window {    default_width  = 100,    default_height = 100,    title          = 'Foo'};window:set_wmclass(class, class);local app = Gtk.Application {    application_id = 'org.awesomewm.tests.rule0000000052'};function app:on_activate()    window.application = self;    window:show_all();end;app:run {''}
psychon  17147  0.4  1.9 423608 38936 ?        SNsl 09:15   0:00 lua -e local lgi = require 'lgi';local Gtk = lgi.require('Gtk');Gtk.init();local class = 'rule0000000053';local window = Gtk.Window {    default_width  = 100,    default_height = 100,    title          = 'Foo'};window:set_wmclass(class, class);local app = Gtk.Application {    application_id = 'org.awesomewm.tests.rule0000000053'};function app:on_activate()    window.application = self;    window:show_all();end;app:run {''}
[...]

PR #868, opened 2016-05-01
Some more screen changes, add tests

psychon:
Hi everyone,

this adds a new API to fake screen additions and removals. Based on this, a new functional test is added that does actually test this. The rest are various fixes.

However, since tags are leaked (hrm, `tag:delete()` should clean up better), 17dab5446ef59933601dc4bf9bd706cc793a6f0d is necessary to make the test pass. Also, https://github.com/awesomeWM/awesome/commit/17dab5446ef59933601dc4bf9bd706cc793a6f0d#diff-f9bfa3f1498dea81c7ec01763ba182a8R30 is really ugly.

This last commit is the WIP part of this PR and I'd be nice to hear some ideas from you what to do about this. If you want, I can remove the "add a new test" and the WIP commits and the rest can be merged.

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -295,6 +295,7 @@ function tag.object.delete(self, fallback_tag, force)
 
     -- delete the tag
     data.tags[self].screen = nil
+    data.tags[self] = nil
     self.activated = false
 
     -- Update all indexes
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -283,8 +283,7 @@ function tag.object.delete(self, fallback_tag)
 
         -- If a client has only this tag, or stickied clients with
         -- nowhere to go, abort.
-        if (not c.sticky and nb_tags == 1) or
-                                    (c.sticky and fallback_tag == nil) then
+        if (not c.sticky and nb_tags == 1) then
             return
         -- If a client has multiple tags, then do not move it to fallback
         elseif nb_tags < 2 then
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -1352,6 +1352,7 @@ capi.tag.add_signal("property::urgent")
 
 capi.tag.add_signal("property::urgent_count")
 capi.tag.add_signal("property::volatile")
+capi.tag.add_signal("request::screen")
 capi.tag.add_signal("removal-pending")
 
 capi.screen.add_signal("tag::history::update")
@@ -1359,9 +1360,18 @@ capi.screen.add_signal("tag::history::update")
 capi.screen.connect_signal("tag::history::update", tag.history.update)
 
 capi.screen.connect_signal("removed", function(s)
-    -- First give other code a chance to clean up
+    -- First give other code a chance to move the tag to another screen
+    for _, t in pairs(s.tags) do
+        t:emit_signal("request::screen")
+    end
+    -- Everything that's left: Tell everyone that these tags go away (other code
+    -- could e.g. save clients)
     for _, t in pairs(s.tags) do
         t:emit_signal("removal-pending")
+    end
+    -- Give other code yet another change to save clients
+    for _, c in pairs(capi.client.get(s)) do
+        c:emit_signal("request::tag", nil, { reason = "screen-removed" })
     end
     -- Then force all clients left to go somewhere random
     local fallback = nil
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -1352,10 +1352,34 @@ capi.tag.add_signal("property::urgent")
 
 capi.tag.add_signal("property::urgent_count")
 capi.tag.add_signal("property::volatile")
+capi.tag.add_signal("removal-pending")
 
 capi.screen.add_signal("tag::history::update")
 
 capi.screen.connect_signal("tag::history::update", tag.history.update)
+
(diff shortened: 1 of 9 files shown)


psychon:
Oh and cb6ce838a3fe6f8b53ecec22760ae4ed7ca9a504 might fix #861!
@Elv13 Any particular reason why you use `GtkApplication` besides "all the lgi examples do so as well"?

Elv13:
> Any particular reason why you use GtkApplication besides "all the lgi examples do so as well"?

It is a copy pasted (and trimmed) LGI example

lib/awful/tag.lua
```diff
@@ -1355,6 +1355,32 @@ capi.screen.add_signal("tag::history::update")
...
+    local old_tags = s.tags
+    -- First try to get delete tags gently
+    for _, t in pairs(s.tags) do
+        t:delete()
+    end
+    -- Then be a little more fierce and force clients to go somewhere random
+    local fallback = nil
+    for other_screen in capi.screen do
+        if #other_screen.tags > 0 then
+            fallback = other_screen.tags[1]
```
Elv13: This should be done in a request::screen handler. Tyrannical may want to move "volatile" tags to existing screens instead of just deleting them. This is also true for the `:delete()` above. The default handler should delete the tags. This make total sense for the default config where tags are less or more used as virtual workspaces.

objects/screen.c
```diff
@@ -1016,6 +1016,57 @@ luaA_screen_count(lua_State *L)
...
+/** Add a fake screen.
+ * @tparam integer x X-coordinate for screen.
+ * @tparam integer y Y-coordinate for screen.
+ * @tparam integer width width for screen.
+ * @tparam integer height height for screen.
+ * @return The new screen.
+ * @function fake_add
+ */
+static int
+luaA_screen_fake_add(lua_State *L)
```
Elv13: Could there be a fake_resize too? It would be useful to be able to split Xephyr screen into "virtual sub screens"[1] for the tests (or use the Xinerama support and just virtually enable and disable the other screens. Also, this is actually useful for people using buggy hardware (like the Radeon HD5450[2]), MOSAIC or other unsupported ways to define screens.

[1] Virtually resize the real screen in 2, then fake_add the second part of the real screen as an imaginary second screen.
[2] That one support 3 screens per card, but expose them as 2 screens because of an (hardware?) bug.

psychon: Yup, that's planned, but only after "just disabling" screens works. One problem at a time. :-)

lib/awful/tag.lua
psychon: Oh, I totally forgot to say "I don't like this part at all and I'm open for suggestions".
How about some "special" signal just for "this tag must go away, do something suitably nice" that is emitted at the beginning of this function? That way tyrannical can hook into that signal and do everything it needs to do and the rest of this function won't be necessary.

Thanks for this idea!

So, what do you think about:
```
screen.connect_signal("removed", function(s)
    for _, t in pairs(s:tags) do
        t:emit_signal("delete-for-screen-removal")
    end
    [Now here the rest of the code]
end)
```
(Also: Ideas for a good name for the signal?)

psychon:
No longer a WIP, I hope. The diff to the previous version of this PR is https://gist.github.com/psychon/676ec08155cee1dd44aaedcd56342471 (but hand-edited, so the patch might be invalid).

Does this look mergable?

lib/awful/tag.lua
```diff
@@ -1350,11 +1352,35 @@ capi.tag.add_signal("property::urgent")
...
+capi.tag.add_signal("removal-pending")
 
 capi.screen.add_signal("tag::history::update")
 
 capi.screen.connect_signal("tag::history::update", tag.history.update)
 
+capi.screen.connect_signal("removed", function(s)
+    -- First give other code a chance to clean up
+    for _, t in pairs(s.tags) do
+        t:emit_signal("removal-pending")
```
Elv13: I know I could implement `request::tag` myself using this signal, but I would prefer it to be integrated into the "default" sequence.

First, tags should emit a `request::screen` without any handler by default. If the screen changed, then return. After that, the client should each emit `request::tag`. The default handler will already do less or more what's your doing here. This way, the tag can be salvaged with its layout and all and just moved to another screen.

psychon: So you are saying:

1. I should rename `removal-pending` to `request::screen` (well, how about `request::new_screen`?)
2. I should do another loop afterwards with `for _, c in ipairs(s.clients) do c:emit_signal("request::tag") end` (how about `request::new_tag`?)

(And after the above, I keep the code that I already have)

Right?

Elv13:
`removal-pending` is still required if `request::screen` end up with no screen. If no handler decided to move the tag, then the code should go ahead and start to delete it.

> c:emit_signal("request::tag") end (how about request::new_tag?)

`request::tag` already exist and does that. As implemented, there is already an hint. So "reason" should be "screen_removed" or somehting like that. Here is how it is currently used:

```lua
 c:emit_signal("request::tag", nil, {reason="rules"})
```

psychon:
Rebased on master for resolving a conflict and added https://github.com/awesomeWM/awesome/pull/868/commits/5b1f8cd35c2b6ea88778238ef13b1e96ef5f32c1 to hopefully make Elv13 happy

Elv13:
:+1:

Issue #848, opened 2016-04-24
make check fails on test-spawn-snid.lua


Jajauma:
Doing `make check` of the current master (2ccc7c8) fails on my system on a step which tests the startup notifications. The complete build and run logs are in the attach.  

Adding a `print()` statement to the Step 1 of the offending test reveals that [c_snid](https://github.com/awesomeWM/awesome/blob/master/tests/test-spawn-snid.lua#L22) happens to be nil for some reason, hence the failure.

[test-failure.txt](https://github.com/awesomeWM/awesome/files/233635/test-failure.txt)

Elv13:
Hello,

This test is known to be currently broken on some distributions because of the compilation flags they use on our test program. I will change it for a less fragile one. Thanks for reporting this.

Jajauma:
As of 6b6cbf4 I've got yet another test case failing. This time it is test-resize.lua, at the lines [here](https://github.com/awesomeWM/awesome/blob/master/tests/test-resize.lua#L279) and [here](https://github.com/awesomeWM/awesome/blob/master/tests/test-resize.lua#L297). Assertions fail because they try to compare the numbers like 395 and 394,5. Wrapping the divisions with a `math.ceil()` call makes this test pass for me. 
```patch
diff --git a/tests/test-resize.lua b/tests/test-resize.lua
index a9eb716..18613b1 100644
--- a/tests/test-resize.lua
+++ b/tests/test-resize.lua
@@ -276,7 +276,7 @@ table.insert(steps, function()
     local c = client.get()[1]
 
     assert(c:geometry().x      == c.screen.workarea.x                                     )
-    assert(c:geometry().y      == c.screen.workarea.y+c.screen.workarea.height/2          )
+    assert(c:geometry().y      == math.ceil(c.screen.workarea.y+c.screen.workarea.height/2))
     assert(c:geometry().width  == c.screen.workarea.width - 2*c.border_width              )
     assert(c:geometry().height == math.ceil(c.screen.workarea.height/2 - 2*c.border_width))
 
@@ -294,7 +294,7 @@ table.insert(steps, function()
     local c = client.get()[1]
 
     assert(c:geometry().x       == c.screen.workarea.x                                     )
-    assert(c:geometry().y       == c.screen.workarea.y+c.screen.workarea.height/2          )
+    assert(c:geometry().y       == math.ceil(c.screen.workarea.y+c.screen.workarea.height/2))
[...]

PR #870, opened 2016-05-01
Use an uneven screen width/height (1921x1079) in one Travis job

blueyed:
This is meant to trigger issues that would need `math.ceil()` fixes etc.

Ref: https://github.com/awesomeWM/awesome/issues/848#issuecomment-216068831


blueyed:
Failed as expected: https://travis-ci.org/awesomeWM/awesome/jobs/127105442.
Should now pass after 612756f has been merged.

psychon:
Assuming that https://github.com/awesomeWM/awesome/pull/882 fixes the problem that this PR highlights, we really should get this merged. Adding to next (under this assumption...).

Issue #834, opened 2016-04-18
Weird changes in API docs


psychon:
https://github.com/awesomeWM/apidoc/commit/e3619cc30000a3462fb87575245fc62524bd3285#diff-ac83493a098e53ca8086890f2845a7d5L3711
https://github.com/awesomeWM/apidoc/compare/dba588b20c...e3619cc300#diff-7d61cae60cd4bcbe5e620947dc757a9aL796

This changes the target of a link in a a build that didn't touch anything in awful (it's the build of my recent naughty PR). The actual link target (`<a id="foo">`) is not modified. So either before or after this change, the link was broken.

What is ldoc doing here? Do we have docs with broken links?

blueyed:
The first change is:

```diff
-         <a href="../classes/client.html#awful.client.get_transient_for_matching">client.get_transient_for_matching</a>
+         <a href="../classes/client.html#client:get_transient_for_matching">client.get_transient_for_matching</a>
```

I've seen some changes like this before, and thought it was because of some
(ldoc?) update during the build.

Both links seem to work after all:

http://awesomewm.github.io/apidoc/classes/client.html#awful.client.get_transient_for_matching
http://awesomewm.github.io/apidoc/classes/client.html#client:get_transient_for_matching

But one appears to be for the (shortened) instance, and the other for the longer one.
But after all those should not get linked to `client`, but `awful.client`.

/cc @Elv13

Elv13:
`awful.client` is "gone" (from ldoc), `awful.client.get_transient_for_matching is deprecated` `client:get_transient_for_matching()` is the non deprecated function. Maybe ldoc use a hashtable and the first element matching happen to change. In that case it would need to be fixed.

blueyed:
Ok, got it.

The issue seems to be though that there are too many changes (and not just the ones from the update/merge).

Relevant log: https://travis-ci.org/awesomeWM/awesome/jobs/123780619#L3360
> 5 files changed, 11 insertions(+), 11 deletions(-)

There are also changes for https://github.com/awesomeWM/awesome/compare/3fe85a8f050a...8cfdc0c59334: https://github.com/awesomeWM/apidoc/commit/ca56189e5ab1696595e18253632f9a7cc9d236c6 ?!

I guess it's time for some review of https://github.com/awesomeWM/awesome/blob/master/build-utils/travis-apidoc.sh.. :)

blueyed:
[...]

PR #871, opened 2016-05-02
doc: fix usage of "mouse.move" as a string

blueyed:
Ref: https://github.com/awesomeWM/awesome/issues/834#issuecomment-216141389.


psychon:
Thanks :+1:

PR #872, opened 2016-05-02
tests: fail if $AWESOME does not exist / is not executable

blueyed:
Without this the stdout/stderr rediraction when launching it makes the
error regarding the "program not found" disappear(?!).

[ci skip]


psychon:
:+1:

PR #873, opened 2016-05-03
Geometry overhaul p3.01

Elv13:


Elv13:
This will be it for this one. I kept the doc in drawin.c for the potential readers of the code. It's not that big anyway.

@psychon One of the rare leak test failure on non-luajit https://travis-ci.org/awesomeWM/awesome/jobs/127416601

lib/wibox/init.lua
```diff
@@ -31,6 +31,174 @@ function wibox:set_widget(widget)
...
+-- @param integer
+
+--- Border color.
+--
+-- **Signal:**
+--
+--  * *property::border_color*
+--
+-- @property border_color
+-- @tparam string|color|gradient|pattern A color
```
psychon: Nope, only strings forming a hexadecimal color (without any alpha component). Welcome to X11! :-)

lib/wibox/init.lua
```diff
@@ -31,6 +31,174 @@ function wibox:set_widget(widget)
...
+-- @see gears.color
```
psychon: Oh, this isn't correct either

lib/wibox/init.lua
```diff
@@ -324,6 +329,8 @@ local function new(args)
     return ret
 end
 
+capi.drawin.add_signal("property::wibox")
+
```
psychon: Did you intend to have this and the previous change in this commit? If so, why and what does it do?

lib/awful/mouse/init.lua
```diff
@@ -298,7 +298,7 @@ function mouse.object.set_current_client() end
 function mouse.object.get_current_wibox()
     local obj = capi.mouse.object_under_pointer()
     if type(obj) == "drawin" then
-        return obj
+        return obj.wibox()
```
psychon: Please use `local wibox = obj.wibox if wibox then return wibox() end` (just in case anyone dares to use a drawin directly, without going through the wibox code)

lib/awful/mouse/init.lua
```diff
@@ -298,12 +298,42 @@ function mouse.object.set_current_client() end
 function mouse.object.get_current_wibox()
     local obj = capi.mouse.object_under_pointer()
     if type(obj) == "drawin" then
-        return obj.wibox()
+        return obj.wibox
```
psychon: Another bad cherry pick? (This should actually be in the "fix `current_wibox`"-commit and I didn't even notice this error...)

lib/awful/mouse/init.lua
```diff
@@ -298,12 +298,42 @@ function mouse.object.set_current_client() end
...
     end
 end
 
 function mouse.object.set_current_wibox() end
 
+--- Get the widgets currently under the mouse cursor.
+--
+-- @property current_widgets
+-- @tparam widget|nil The widget
```
psychon: There is no such parameter, is there?

lib/awful/mouse/init.lua
```diff
@@ -298,12 +298,42 @@ function mouse.object.set_current_client() end
...
+-- @treturn table The list of widgets.The first element is the biggest
+-- container while the last is the topmost widget. The table contains *x*, *y*,
+-- *width*, *height* and *widget*.
+-- @treturn table The list of geometries.
+-- @see wibox.find_widgets
+
+function mouse.object.get_current_widgets()
+    local obj = capi.mouse.object_under_pointer()
```
psychon: Why don't you use `get_current_wibox()` here?

lib/awful/mouse/init.lua
Elv13: ldoc has a funny syntax, "param" here is turned into "type". I could make a "@type" alias, but it would also be misleading.

But there really is an error, it should be `@tparam widget|nil widget The widget`

Elv13:
all done

lib/wibox/init.lua
psychon: Ping? Do you want to move these two changes to the next commit?

lib/awful/mouse/init.lua
```diff
@@ -297,8 +297,8 @@ function mouse.object.set_current_client() end
 
 function mouse.object.get_current_wibox()
     local obj = capi.mouse.object_under_pointer()
-    if type(obj) == "drawin" then
-        return obj
+    if type(obj) == "drawin" and obj.wibox then
+        return obj.wibox()
```
psychon: This shouldn't be a function call

psychon:
Two comments left: No function call in `get_current_wibox`, but return `.wibox` directly (sorry, I was a bit confused yesterday) and some change is still in the IMO wrong commit (also the addition of `.wibox`).

Feel free to merge after looking at these (and e.g. deciding to keep it the way it is)

lib/awful/mouse/init.lua
Elv13: That will be problematic, it is that way to avoid a leak in Lua 5.1. The only way I found to fix it is by using an upvalue weak table. I will rename this to `get_wibox()`. Otherwise, it would look like `.wibox[1]` and that's way worst than `wibox()`

psychon: Sorry then. Apparently I was confused. I thought this used $MAGIC to then call your `get_wibox` function which directly returned the wibox, instead of a function returning the wibox. Apparently I'm wrong....

Elv13: I can't do that. As we talked about when I implemented that system, the only way to avoid leaks in 5.1 is to use upvals to store the properties. Given this is exactly this type of leaks, I had no choice but to do it like this (at least of the solutions I tried). Otherwise, the leaks test would fail, much, much more often (maybe all the time for 5.1 and the vast majority of the time for luajit, my test pool is too small to confirm).

psychon: Ok, so this confirms that I misremembered this when writing the above comment. Just make sure this works (and the code change is in the right commit) and merge it then. Sorry for my confusion.

PR #874, opened 2016-05-04
Awful.widget.button buttons are not released when the mouse leave the widget

Elv13:
This is in turn an issue with `awful.button`. Here is a test that trigger the problem

### DIFF
--- a/lib/awful/widget/button.lua
+++ b/lib/awful/widget/button.lua
@@ -39,6 +39,9 @@ function button.new(args)
     end
     w:set_image(args.image)
     w:buttons(abutton({}, 1, function () orig_set_image(w, img_press) end, function () orig_set_image(w, img_release) end))
+
+    w:connect_signal("mouse::leave", function(self) orig_set_image(self, img_release) end)
+
     return w
 end
 
--- a/tests/test-awful-widget-button.lua
+++ b/tests/test-awful-widget-button.lua
@@ -1,0 +1,119 @@
+local runner    = require( "_runner"   )
+local wibox     = require( "wibox"     )
+local awful     = require( "awful"     )
+local beautiful = require( "beautiful" )
+
+local steps = {}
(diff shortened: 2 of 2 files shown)


psychon:
What's the semantics we want here? X11 provides a passive grab: As long as the mouse button is pressed, all mouse events "belong to us", even when the mouse cursor moves elsewhere. So options are:
- Pretend the button is released when it leaves the widget (not really correct, would likely also need to fake a button press on the neighboring widget)
- Remember the widget and send it the release event when the button is actually released (which will give a position possible wide outside of the widget for the button release)

Elv13:
Buttons in GTK and Qt seem to start a mousegrabber and "accept" the button action only when the mouse button is released within the widget area.

psychon:
Hm, no idea how that could be implemented.

How about keeping things like they are now and widgets who want to track something while button is pressed will have to assume themselves that `mouse::leave` is a button release? After all, the events that we are sending right now could be interpreted as being correct: Everything that happens inside the widget causes an event.

Elv13:
Updated with the proposed "fix". It isn't correct, but make the bug itself go away (and replace it by another much less visible bug...)

psychon:
The unrelated test failure is being fixed in #890. What do you mean with "another much less visible bug"? What solution to this issue would you prefer?

Elv13:
> What do you mean with "another much less visible bug"?

The button image is set to "released" when the mouse leave or the click is released. But if you leave the widget, then enter again without click, it is not set to "pressed". This is very minor. The current bug is that the button stay in "pressed" mode forever until it is clicked and released correctly again.

Elv13:
Rebased

psychon:
Fine by me and not doing anything special for "pressed button leaves a widget" is also fine by me, mostly since I have no good idea on what to do....

PR #875, opened 2016-05-05
Startup-Notification fix

### LINKED ISSUE #848: make check fails on test-spawn-snid.lua

psychon:
This started as "let's fix #848" and turned into "Oh, wow, GTK does startup notification in a way incompatible to what we expect". Could someone test SN with Qt/KDE, please? (@Elv13 How about you?)

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1241,8 +1241,19 @@ HANDLE_GEOM(height)
         xcb_get_property_reply(globalconf.connection, startup_id_q, NULL);
     /* Say spawn that a client has been started, with startup id as argument */
     char *startup_id = xutil_get_text_property_from_reply(reply);
-    c->startup_id = startup_id;
     p_delete(&reply);
+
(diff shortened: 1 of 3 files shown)


blueyed:
Does this enable SN for all GTK apps?

psychon:
I guess so, but I only tested it with what is in this pull request. (Well, let's say "fix" instead of "enable")
According to GTK docs, it automatically does "SN startup complete" when the first window is shown, unless you tell it explicitly "this is just a startup splash, real startup is still going on"

PR #877, opened 2016-05-07
doc(readme: faq): add some hints

actionless:


docs/90-FAQ.md
```diff
@@ -100,6 +100,8 @@ On zsh, any changes to $PATH you do in `~/.zshrc` will not be picked up (because
 this is only run for interactive shells). Use `~/.zshenv` instead to make
 additions to the path you want to use in awesome.
 
+If you need to execute command asynchronously (not blocking awesome while long command is running) you can use `awful.spawn.easy_async`.
```
psychon: I think it makes more sense to introduce this with "If you need the output of a program". The stuff above (`spawn_with_shell`) is also async. :-)

docs/90-FAQ.md
```diff
@@ -283,6 +291,18 @@ you don't want new clients to be urgent by default put this in your rc.lua:
...
     client.connect_signal("request::activate", awful.ewmh.activate)
+    
+### How to prevent mouse cursor from jumping to the corner when resizing or moving a client
+
+You can specify the behavior for these layouts:
+
+    awful.layout.suit.floating.resize_jump_to_corner = false
+    awful.layout.suit.tile.resize_jump_to_corner = false
+    
+### How to create a simple panel widget
```
psychon: What's a panel widget? I think this should be "How to create a simple widget showing the output of a shell command", or something like that.

psychon:
Thanks for this. These are really nice additions! :+1:

docs/90-FAQ.md
```diff
@@ -283,6 +291,18 @@ you don't want new clients to be urgent by default put this in your rc.lua:
         end
     end
     client.connect_signal("request::activate", awful.ewmh.activate)
+    
+### How to prevent mouse cursor from jumping to the corner when resizing or moving a client
+
+You can specify the behavior for these layouts:
+
+    awful.layout.suit.floating.resize_jump_to_corner = false
+    awful.layout.suit.tile.resize_jump_to_corner = false
```
Elv13: Isn't this broken right now? I thought I kind of broke that a while back... ( The feature itself still exist, but not that way)

    awful.placement.resize_to_mouse(client.focus)

Kind of does something similar. Else, it would need a `pretend = true` in `awful.mouse.client.resize`. That's not possible right now.

actionless: in my config it's defined like that and i am using the latest master, could you re-check? mb some of your branches is not merged yet

docs/90-FAQ.md
blueyed: Also s/execute command/execute a command/

docs/90-FAQ.md
blueyed: s/prevent mouse/prevent the mouse/

docs/90-FAQ.md
```diff
@@ -283,6 +291,18 @@ you don't want new clients to be urgent by default put this in your rc.lua:
         end
     end
     client.connect_signal("request::activate", awful.ewmh.activate)
+    
+### How to prevent mouse cursor from jumping to the corner when resizing or moving a client
```
blueyed: s/prevent mouse/prevent the mouse/

docs/90-FAQ.md
actionless: my bad, the problem was on my side

actionless: how that feature can be re-enabled in the config then?

blueyed: @Elv13 
ping

blueyed:
@actionless 
Can you please address the typos / other minor fixes here?

It seems like this is also waiting on feedback on the "How to prevent mouse cursor from jumping to the corner when resizing or moving a client", but maybe that's even not relevant anymore?
Seems to me like a bug and not something to put into a doc anyway?!

actionless:
@blueyed, no that's not a bug, but the different behavior when resizing.
in awesome wm mouse cursor jumps to the closest edge of the client when resizing, while in majority of other wm-s that not happens

yup, i will address the comments together with that part

docs/90-FAQ.md
actionless: pinging @Elv13 again :)

Elv13: `awful/mouse/init.lua`, line 213 need to be adapted with offsets like the one and pretends like like the one just above in `.move()`. It currently isn't exposed externally, so it will need a config option.

actionless: (linking the original PR for myself to find it https://github.com/awesomeWM/awesome/pull/396)

psychon:
Closing for lack of activity (getting the number of open issues down seems a bit out of reach, so I am targeting open PRs now)

PR #878, opened 2016-05-07
Geometry overhaul p3.02 (awful.wibox)

Elv13:
This remove most of the `awful.wibox` code

TODO:

 * <s>Rename to `awful.wibox` to avoid confusion (with proper deprecation)</s>
 * <s>Turn into a class</s>
 * <s>Finish the tests</s>
 * <s>Track the workarea locally to avoid a a dependency on it's own workarea offset.</s>
 * <s>Test screen resize</s>
 * <s>Test multiple stacked wibars</s>
 * <s>Test multiple screens</s>


Elv13:
I am not done writing the tests, but I would appreciate a review as everything else is finished. Don't be scared by the SLoC changes, the "real changes" are like ~250 lines, everything else is doc, tests and a file renaming.

psychon:
I'll take a look at the "total diff", but I'll skip the individual commits for now.

So... since "wibox" could refer to two different things, you want to rename `awful.wibox` to `awful.wibar` (widget bar?). Ok, fine by me.

I like the comment "When the parent geometry (like the screen) changes, re-apply the placement function". This sounds like you are solving a problem which I was going to have and didn't yet know how to solve. Nice!

No "real" comments from me right now. I guess I might come up with something when looking at the individual commits, but the "big change" looks nice. (Well, I don't really like "import some wibox documentation" because that means that this documentation is duplicated and exists in several different positions. They will get out of sync eventually. My preferred fix would be to have some nice way to properly implement this, my second idea would be to just say 'also look here'.

aroig:
I tested this PR as @Elv13 suggested in #892. There, I reported a problem with struts, I'll repeat it here, so it is easier to keep track of.

The thing is, running with this PR I get struts of the same size both on the top and bottom of the screen

    $ xprop  _NET_WM_STRUT_PARTIAL
    _NET_WM_STRUT_PARTIAL(CARDINAL) = 0, 0, 18, 18, 0, 0, 0, 0, 0, 0, 0, 0

However there is only one wibar on top.

By the way, I've been using this on a 2-screen setup for about 2 days, and I'm happy to say that this is the only problem I found! I'm going back to master though...

Elv13:
@aroig I know, it's a regression from an earlier version. The older version had issues with stacking many wibars/wiboxes. This version kind of fix it, but has this issue. It's why the tests fails.

Elv13:
This is no longer a work in progress. This is finally good enough. I went a little overboard with the handling of some obscure corner cases. Many of them make no sense to support (like moving a horizontal wibar to a vertical position), but I support them anyway. All the tests also found some issues with the placement code, so at least it really has great coverage now. @aroig it should now work on multiple screens again. Please test (I don't have the hardware to do it right now). The problem was that `awful.screen.getbycoords` returns the wrong screen when calling the function with a [screen.geometry.x+0,screen.geometry.x+0] position. So I added a `+1`. It's a bit cheap, but its (I think) good enough to solve the issue.

lib/awful/widget/layoutbox.lua
```diff
@@ -60,7 +59,7 @@ function layoutbox.new(screen)
         w._layoutbox_tooltip = tooltip({ objects = {w}, delay_show = 1 })
 
         update(w, screen)
-        boxes[screen] = w
+        boxes[screen] = setmetatable({}, {__index = w, __newindex = w})
```
psychon: How does wrapping the widget in an empty table change anything? This new entry keeps the widget alive, so just using `= w` here should have the same effect.

But whatever, if it works, I'll not question it.

lib/wibox/init.lua
```diff
@@ -347,10 +356,16 @@ local function new(args)
...
+                return self["get_"..k](self)
+            else
+                return w[k]
+            end
+        end,
         __newindex = function(self, k,v)
-            if wibox["set_"..k] then
-                wibox["set_"..k](v)
+            if rawget(self, "set_"..k) then
+                self["set_"..k](self, v)
```
psychon: This seems like a fixup for the commit that added this `wibox["set_"..k]`, no? Perhaps this part of the change should be squashed into that.

(Why does this need/use rawget?)

lib/awful/placement.lua
```diff
@@ -540,6 +555,33 @@ attach = function(d, position_f, args)
...
+                end
+            end
+        end
+
+        if args.update_workarea then
+            d:disconnect_signal("property::geometry" , tracker_struts)
+            d:disconnect_signal("property::visible"  , tracker_struts)
+            capi.client.disconnect_signal("property::struts", tracker_struts)
+        end
+    end
```
psychon: Does this warrant a new leak check in some test?

lib/wibox/init.lua
```diff
@@ -8,14 +8,16 @@
...
+    awesome = awesome,
+    screen = screen
 }
 local setmetatable = setmetatable
 local pairs = pairs
 local type = type
 local object = require("gears.object")
 local beautiful = require("beautiful")
 local base = require("wibox.widget.base")
+local ascreen = require("awful.screen")
```
psychon: I'm afraid that `wibox` shouldn't depend on functions from `awful`. We already have enough "weird dependency loops" (My hack around this was to have a copy called `screen_getbycoord` in `lib/wibox/drawable.lua`). Perhaps this should be worked around by moving some helper function into `gears`?

lib/wibox/init.lua
```diff
@@ -268,6 +270,18 @@ function wibox:find_widgets(x, y)
...
 
+function wibox:get_screen()
+    return capi.screen[ascreen.getbycoord(self.x, self.y)]
+end
+
+function wibox:set_screen(s)
+    s = capi.screen[s or 1]
+    if s ~= self:get_screen() then
+        self.x = s.geometry.x + 1
+        self.y = s.geometry.y + 1
```
psychon: Assuming that this is the "+1" that you refer to in your commit: I think you can get rid of this again and have it fixed by #909.

lib/wibox/init.lua
```diff
@@ -357,7 +371,7 @@ local function new(args)
     -- If a value is not found, look in the drawin
     setmetatable(ret, {
         __index = function(self, k)
-            if wibox["get_"..k] then
+            if rawget(self, "get_"..k) then
```
psychon: Huh? I'll just ask: Is this on purpose and does it really belong into this commit? Further, I'll assume that the answer is "yes" and won't try to understand the details.

lib/awful/wibox.lua
```diff
@@ -32,77 +32,157 @@ local awfulwibox = { mt = {} }
...
---- Reset all wiboxes positions.
-local function update_all_wiboxes_position()
-    for _, wprop in ipairs(wiboxes) do
-        awfulwibox.set_position(wprop.wibox, wprop.position, wprop.screen)
+-- Create the placement function
+local function gen_placement(position, stretch)
+    local maximize = (position == "right" or position == "left") and
+        "maximize_vertically" or "maximize_horizontally"
+
+    return placement[position] + (stretch and placement[maximize] or nil)
```
psychon: Let's assume `stretch` is false. This does e.g. `placement.left + nil`. I don't think `nil` in an addition ever works, does it?

psychon:
No real complaints by me. I'm really, really unhappy about `wibox` depending on `awful` and would like someone to be done about this, but I don't have any good suggestions. That sounds like a move to `gears` then...
The addition with `nil` looks fishy, the `+1` should be fixed by #909, the rest is fine by me. I'll trust you on this.

Feel free to merge when you think my comments are addressed and/or wrong.

lib/wibox/init.lua
Elv13: > (Why does this need/use rawget?)

To avoid getting handled by the metatable in an infinite loop. Eventually, it will hit the drawin __index, then the drawin miss handler. This is not what we need. That would prevent wibox polymorphism as used by wibars, tooltips, notifications and menus.

lib/awful/placement.lua
Elv13: Your screen-change test cover this for the screens. Given everything else use the same code, then I assume it works. There is a "leak" if something is not detached. But it actually make sense not to collect those wiboxes/clients/screens until then, so I don't plan to fix it, it's a feature.

Elv13:
Ok, all done. @psychon I consolidated rectangle related functions into gears. I hope the result is good enough for you. I also removed `gears.screen`. It was already deprecated and "existed" only 3 days. After this is merged, do you want me to use my virtual chainsaw on `awful.util` and move it to `gears` (with proper deprecation and separation of concerns)? This way, `wibox` would stop depending on `awful` (almost).

Elv13:
There is a leak when using fake_resize (the test is also broken, but I already have a patch to fix it). I suggest to merge this first and fix it later (read: tomorrow). It wont be too bad.

lib/wibox/drawable.lua
```diff
@@ -19,30 +19,26 @@ local color = require("gears.color")
...
     local geom = self.drawable:geometry()
-    local s = screen_getbycoord(geom.x, geom.y)
+
+    local sgeos = {}
+
+    for s in capi.screen do
+        sgeos[s] = s.geometry
+    end
+
+    local s = grect.get_by_coord(sgeos, geom.x, geom.y)
```
psychon: This needs an `or capi.screen.primary` at the end (same fallback as the previous code uses)

lib/gears/geometry.lua
```diff
@@ -76,4 +76,76 @@ function gears.geometry.rectangle.get_by_coord(list, x, y)
...
+    local gBy = _gB.y
+
+    if dir == "up" then
+        gBy = _gB.y + _gB.height
+    elseif dir == "down" then
+        gAy = _gA.y + _gA.height
+    elseif dir == "left" then
+        gBx = _gB.x + _gB.width
+    elseif dir == "right" then
+        gAx = _gA.x + _gA.width
```
psychon: For future-me: I think this needs a `-1` at the end (however, this won't ever make a difference, I think)

psychon:
I think I took a look at the new commits (those with "Commits on May 16, 2016"), but with GitHub's UI, one can never be sure...

:+1:

PR #879, opened 2016-05-08
Priority heap to sort menubar entries by execution count

modk:
Using menubar as my primary means of starting applications, I was weary going through the menu the same way all the time (i.e. type two characters go right three times etc.). Therefore I hacked together a mechanism that displays the menubar entries sorted by the number times they have been executed before:

A file `menu_count_file` is created containing the application name and the number of times it has been executed. These values are used as input for a priority heap. The file is currently recreated every time an application is executed and re-read every time the menu is opened. This is certainly not optimal, a memory-cached version of the heap would be better. 

I shamelessly copied the heap implementation from here http://lua-users.org/lists/lua-l/2007-07/msg00482.html.  Luis Carvalho, the original author, should be asked for permission before including this. 

In addition, neither am I familiar with lua nor awesome idioms so there is certainly room for improvement. It's a start that works well for me (and has been working for some weeks), however.


psychon:
> I shamelessly copied the heap implementation from here http://lua-users.org/lists/lua-l/2007-07/msg00482.html. Luis Carvalho, the original author, should be asked for permission before including this.

Yup, this is likely not going to fly. However, do you need a heap at all? Can't a simple `table.sort` be used for this? (And before someone says "slooooow!", I'd like to see some actual evidence for this first).

Also, you seem to have written this based on an older version of menubar (you are re-introducing the `nocase` function, you are using `awful.util.spawn` etc). Could you rewrite this for the current version of the file without reverting other changes? If you do so, please move the loading and saving of the cache file to separate `local` functions to improve readability (and you might want to install `luacov`, the travis build failed due to 40 warnings from Luacov).

modk:
I used the heap in order to always have a sorted list. In hindsight, an unnecessary/premature optimization. You're probably right that performance is not major concern here as the number of applications is unlikely to exceed say 1000.

psychon:
Sigh. I always mix up luacov and luacheck. Sorry. I meant luacheck.

You could also ignore luacheck for now and update this pull request with more changes. Travis will run luacheck for you and tell you about failures. However, travis needs a couple of minutes for this, so it's a bit slow.

blueyed:
Thanks!

Issue #756, opened 2016-03-13
Screen object property::workarea is emitted too often


Elv13:
It will need some caching. It is emitted many times without the workarea actually changing. This cause some useless client layout redraw. Also, the `workarea` screen CAPI property doesn't scale very well when using large number of wiboxes.

psychon:
Grepping through were we emit that property... there might be a case on client unmanage where we emit that signal needlessly, but that's so unimportant that I do not care.
The only real place where this is emitted needlessly is when the struts of some window changes. We needlessly emit `property::workarea` on all screens and not just the one where the window is visible. However, struts shouldn't change often, I guess.

Do you have a lua traceback for these needless signals without the workarea actually changing?

Also, what do you mean with "the `workarea` screen CAPI property doesn't scale very well when using large number of wiboxes"?

Elv13:
> Do you have a lua traceback for these needless signals without the workarea actually changing?

In my upstream_dynamic_complete branch, there is a tag geometry change callback in wrapper.lua. The backtrace isn't really useful. It mostly happen when restarting Awesome with many clients. Lot of noise, then it is less noisy (when client come and go).

I opened this because I implemented some `property::geometry` for tags to avoid relying on `awful.layout.arrange()`. The problem with arrange in a stateful scenario is that there is different type of "refresh". The hierarchy main layout method doesn't need to be called everytime (thanks to your layout cache), so I had to check for WA changes from `arrange()` output manually. When I implemented this, I found that `property::workarea` was emitted 10-20 time without actual changes when starting Awesome. Sorry for not being clean the first time. My config has a bunch of docked wiboxes too, so it might increase the count, but for many of the calls, the value is the same.

PR #880, opened 2016-05-08
Cache a screen's workarea

psychon:
Instead of computing the workarea whenever some Lua code asks for it, it is now
remembered explicitly as a property on a screen. This allows us to only emit
property::workarea if the workarea actually changed.

Fixes: https://github.com/awesomeWM/awesome/issues/756


Elv13:
Thanks :+1:

PR #882, opened 2016-05-08
tests: test-resize: use math.ceil to handle uneven screen width

Jajauma:
See https://github.com/awesomeWM/awesome/issues/848#issuecomment-217361136.

This is actually a leftover from 612756f and is meant to fix the failure
at https://github.com/awesomeWM/awesome/pull/870.


blueyed:
Thanks!

Issue #427, opened 2015-08-26
Superfluous mouse event during move block (KiCad)


Araeos:
This may be specific to KiCAD, but the behaviour below does only apply to awesome wm.

# Description
When dragging components in KiCad, the viewport recenters immediately on placing them (see [video](https://www.dropbox.com/s/53ci9u9nnu933ss/kicad-block-move.avi?dl=0))

_Expected behaviour:_ KiCad does not change the viewport or mouse position during the move action, specifically when placing the object using left mouse click.

_Current behaviour:_ KiCad seems to recenter the screen on click before components are acutally placed, places the components on the (now incorrect) position of the mouse and then moves the mouse to the screen's center.
In addition to this movement, this results in an offset `diff` from the expected placement where `diff` is the initial difference of the mouse pointer to the center of the viewport.

When doing the above block movement while holding any modifier key (only during placement, *not* selection) this issue does not manifest itself.

# Analysis
This behavour is a known [issue on launchpad](https://bugs.launchpad.net/kicad/+bug/1480584), but the underlying cause seems to lie in awesome's handling of input events.
Clicking (without modifiers) seems to invoke a mouse movement or mouse leaving event of some sort, at least in the context of dragging the block.
As I am not familiar with wxWidgets/KiCad/awesome internals this is partially speculation.

# Software used
- awesome wm (v3.5.6), default config (rc.lua)
- KiCad product build (2015-08-24 BZR 6119)
- Platform: Arch Linux

psychon:
Could you try removing the mouse binding for the left click from the default config?

     awful.button({ }, 1, function (c) client.focus = c; c:raise() end),

(Note that after this focusing a client becomes harder, so better plan before. The tasklist and titlebars should still work.)

If this really is the problem, then... I don't know.

Araeos:
Yes removing that line fixes the described behaviour in revisions of kicad before 6130.
Revision 6130 introduces a patch that avoids the recentering when the mouse did not leave the viewport.

Interestingly changing the above line to

     awful.button({ }, 1, function (c) end),

did _not_ fix this issue. Only completely removing helps.
This seems very mysterious to me, since I thought the focus or `raise()` call was the issue.

psychon:
Well, awesome will create a grab for the mouse button when you ask it to do so. And when this grab activates, the X11 server will inform KiCAD that it won't be getting all mouse events by telling it via a special kind of leave. Hm...

osleg:
[...]

PR #883, opened 2016-05-08
Grab buttons on the child window

psychon:
Since commit 102063dbbd, awesome is a reparenting WM. That means that we put our
own frame windows around child windows. This means that we have the option of
grabbing input events on the frame window or the child window. This commit chose
the frame window for this.

For keyboard events, this decision was already reverted in 532ec0cd90. This
commit does the same thing for mouse events.

This fixes the spurious leave/enter events that were visible on mouse clicks.
They occurred because the click activated a passive grab (all mouse events now
"belonged" to awesome). This passive grab caused the X server to inform clients
that they "lost" the mouse pointer (with the detail field set to "a grab
activated").

Fixes: https://github.com/awesomeWM/awesome/issues/427

### DIFF
--- a/event.c
+++ b/event.c
@@ -223,7 +223,7 @@ event_handle_button(xcb_button_press_event_t *ev)
                              XCB_ALLOW_ASYNC_POINTER,
                              XCB_CURRENT_TIME);
     }
-    else if((c = client_getbyframewin(ev->event)))
+    else if((c = client_getbywin(ev->event)))
     {
         luaA_object_push(L, c);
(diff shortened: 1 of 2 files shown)


gldrk:
This is not ICCCM-compliant btw.
>Clients should establish button and key grabs only on windows that they own. 
>
>In particular, this convention means that a window manager that wishes to establish a grab over the client's top-level window should either establish the grab on the root or reparent the window and establish the grab on a proper ancestor.

PR #884, opened 2016-05-08
gears.wallpaper: Handle concurrent screen changes

psychon:
The code in gears.wallpaper currently sets a wallpaper in a deferred fashion.
Only a while after it is told to do something does it actually do the wallpaper
change. This is to incorporate many wallpaper changes right after another. These
changes happens during startup where the wallpaper for each screen is set one
after another.

However, since we no longer restart on RandR changes, the screen configuration
could change while we have a pending wallpaper. In this case, part of the
wallpaper could be "chopped off", because the surface that we draw the wallpaper
to is too small.

This commit makes gears.wallpaper track the size of the pending wallpaper and
create a new surface if the already-pending one is too small.

@blueyed Nope, I still have no idea what might be going wrong with your wallpapers, but thinking about them made me come up with this.

### DIFF
--- a/lib/gears/wallpaper.lua
+++ b/lib/gears/wallpaper.lua
@@ -18,7 +18,7 @@ local function root_geometry()
     return { x = 0, y = 0, width = width, height = height }
 end
 
--- A cairo surface that we still want to set as the wallpaper
+-- Information about a pending wallpaper change, see prepare_context()
 local pending_wallpaper = nil
 
 --- Prepare the needed state for setting a wallpaper.
@@ -29,34 +29,48 @@ local pending_wallpaper = nil
 -- @return[1] The available geometry (table with entries width and height)
 -- @return[1] A cairo context that the wallpaper should be drawn to
 function wallpaper.prepare_context(s)
+    local root_width, root_height = root.size()
     local geom = s and screen[s].geometry or root_geometry()
-    local cr
+    local source, target, cr
 
     if not pending_wallpaper then
         -- Prepare a pending wallpaper
-        local wp = surface(root.wallpaper())
-
-        pending_wallpaper = wp:create_similar(cairo.Content.COLOR, root.size())
-
-        -- Copy the old wallpaper to the new one
-        cr = cairo.Context(pending_wallpaper)
-        cr:save()
-        cr.operator = cairo.Operator.SOURCE
-        cr:set_source_surface(wp, 0, 0)
-        cr:paint()
-        cr:restore()
+        source = surface(root.wallpaper())
+        target = source:create_similar(cairo.Content.COLOR, root_width, root_height)
 
         -- Set the wallpaper (delayed)
         timer.delayed_call(function()
             local paper = pending_wallpaper
             pending_wallpaper = nil
-            wallpaper.set(paper)
-            paper:finish()
+            wallpaper.set(paper.surface)
+            paper.surface:finish()
         end)
+    elseif root_width > pending_wallpaper.width or root_height > pending_wallpaper.height then
+        -- The root window was resized while a wallpaper is pending
+        source = pending_wallpaper.surface
+        target = source:create_similar(cairo.Content.COLOR, root_width, root_height)
     else
         -- Draw to the already-pending wallpaper
-        cr = cairo.Context(pending_wallpaper)
+        source = nil
+        target = pending_wallpaper.surface
     end
+
+    cr = cairo.Context(target)
+
+    if source then
+        -- Copy the old wallpaper to the new one
+        cr:save()
+        cr.operator = cairo.Operator.SOURCE
+        cr:set_source_surface(source, 0, 0)
+        cr:paint()
+        cr:restore()
+    end
+
+    pending_wallpaper = {
+        surface = target,
+        width = root_width,
+        height = root_height
+    }
 
     -- Only draw to the selected area
     cr:translate(geom.x, geom.y)


blueyed:
It sounds like it could solve the issue I am seeing, will try it tomorrow.

blueyed:
@psychon 
Did not help.

But the following patch still helps:
```diff
diff --git i/lib/gears/wallpaper.lua w/lib/gears/wallpaper.lua
index 540df80..3c95e64 100644
--- i/lib/gears/wallpaper.lua
+++ w/lib/gears/wallpaper.lua
@@ -174,7 +174,7 @@ function wallpaper.maximized(surf, s, ignore_aspect, offset)
     cr:set_source_surface(surf, 0, 0)
     cr.operator = cairo.Operator.SOURCE
     cr:paint()
-    surf:finish()
+    -- surf:finish()
 end
 
 --- Set a fitting wallpaper.
```

psychon:
@blueyed I think it's finally time that you open an issue about this (perhaps with some hints on how to reproduce this, but apparently this is hard to do for others...).
Also, could you try if the following patch says anything but `SUCCESS`? (Oh and: Did I already ask you about your cairo version? Could you try something newer?)
```patch
index fe6609a..37f0549 100644
--- a/lib/gears/wallpaper.lua
+++ b/lib/gears/wallpaper.lua
@@ -13,6 +13,15 @@ local root = root
 
 local wallpaper = { mt = {} }
 
+local function check(...)
+    for k, v in ipairs{ ... } do
+        print(k, v, v.status)
+        timer.delayed_call(function()
+            print("delayed", k, v, v.status)
+        end)
+    end
+end
+
 local function root_geometry()
     local width, height = root.size()
     return { x = 0, y = 0, width = width, height = height }
@@ -52,6 +61,7 @@ function wallpaper.prepare_context(s)
             pending_wallpaper = nil
             wallpaper.set(paper)
             paper:finish()
+            check(paper)
         end)
     else
         -- Draw to the already-pending wallpaper
@@ -106,6 +116,7 @@ function wallpaper.centered(surf, s, background)
     cr:set_source_surface(surf, 0, 0)
     cr:paint()
     surf:finish()
+    check(surf, cr)
 end
 
 --- Set a tiled wallpaper.
@@ -127,6 +138,7 @@ function wallpaper.tiled(surf, s, offset)
     cr.operator = cairo.Operator.SOURCE
     cr:paint()
     surf:finish()
+    check(surf, cr)
 end
 
 --- Set a maximized wallpaper.
@@ -161,6 +173,7 @@ function wallpaper.maximized(surf, s, ignore_aspect, offset)
     cr.operator = cairo.Operator.SOURCE
     cr:paint()
     surf:finish()
+    check(surf, cr)
 end
 
 --- Set a fitting wallpaper.
@@ -192,6 +205,7 @@ function wallpaper.fit(surf, s, background)
     cr:set_source_surface(surf, 0, 0)
     cr:paint()
     surf:finish()
+    check(surf, cr)
 end
 
 return wallpaper
```

blueyed:
With your patch I am getting:

```
1       lgi.rec 0x1600a10:cairo.Surface SUCCESS
delayed 1       lgi.rec 0x13bb070:cairo.RecordingSurface        SUCCESS
delayed 2       lgi.rec 0x1600e70:cairo.Context SUCCESS
delayed 1       lgi.rec 0x13bb070:cairo.RecordingSurface        SUCCESS
delayed 2       lgi.rec 0x13c1820:cairo.Context INVALID_MATRIX
delayed 1       lgi.rec 0x1600a10:cairo.Surface SUCCESS
```

But it still (kept) working - I've fixed it before by re-moving and re-attaching to the external screen.

When I then disable the external screen I get:
```
1       lgi.rec 0x282e940:cairo.Surface SUCCESS
delayed 1       lgi.rec 0x25e8910:cairo.RecordingSurface        SUCCESS
delayed 2       lgi.rec 0x282eda0:cairo.Context SUCCESS
delayed 1       lgi.rec 0x282e940:cairo.Surface SUCCESS
```


Re-enabling the screen again then retriggers it somehow:

```
1       lgi.rec 0x2075750:cairo.RecordingSurface        SUCCESS
2       lgi.rec 0x22bcac0:cairo.Context SUCCESS
1       lgi.rec 0x2075750:cairo.RecordingSurface        SUCCESS
2       lgi.rec 0x207bb30:cairo.Context INVALID_MATRIX
```
```
1       lgi.rec 0x22bc660:cairo.Surface SUCCESS
delayed 1       lgi.rec 0x2075750:cairo.RecordingSurface        SUCCESS
delayed 2       lgi.rec 0x22bcac0:cairo.Context SUCCESS
delayed 1       lgi.rec 0x2075750:cairo.RecordingSurface        SUCCESS
delayed 2       lgi.rec 0x207bb30:cairo.Context INVALID_MATRIX
delayed 1       lgi.rec 0x22bc660:cairo.Surface SUCCESS
```

I think the issue is related to panning, which I use to get more space on the external screen / match the high DPI of the laptop screen.

```
% xrandr -q
Screen 0: minimum 8 x 8, current 2688 x 2592, maximum 32767 x 32767
eDP1 connected 1920x1080+768+1512 (normal left inverted right x axis y axis) 276mm x 156mm
DP2-2 connected 2688x1512+0+0 (normal left inverted right x axis y axis) 521mm x 293mm panning 2688x1512+0+0
```

It seems that the wallpaper is applied to the not panned region only (i.e. it is 1920x1080 but the screen is panned to 2688x1512)!

These are the `xrandr` commands (via autorandr):
```sh
xrandr --output DP2-2 --transform none
xrandr --output DP2-2 --gamma 1.0:0.909:0.833 --mode 1920x1080 --panning 2688x1512+0+0 --pos 0x0 --rate 60.00 --reflect normal --rotate normal --transform 1.399994,0.000000,0.000000,0.000000,1.399994,0.000000,0.000000,0.000000,1.000000 --output eDP1 --gamma 1.0:0.909:0.833 --mode 1920x1080 --panning 0x0 --pos 768x1512 --rate 60.04 --reflect normal --rotate normal
```

blueyed:
@psychon 
I am using cairo-infinality-ultimate 1.14.6-2 on Arch Linux.

psychon:
Uhm. Why is it a recording surface?!?!? It should either be an XCB surface or an image surface. That INVALID_MATRIX also sounds weird. Could you try checking where that comes from? Does something call `cr:scale(0, 0)` or something like that? (This delayed stuff was apparently useless and can be removed)

psychon:
Never mind my previous comment. I guess you are using the default wallpaper from the xresources themes and that's a recording surface since recently. And lgi does not know XCB surfaces and thus this is just a normal cairo.Surface.

So only "add `print("I'm here", surface.status)` in lots of places and try to figure out where the INVALID_MATRIX error comes from, please" is left.

PR #886, opened 2016-05-08
Fake screen resize

psychon:
This addresses https://github.com/awesomeWM/awesome/pull/868#discussion_r61692761. Only the last two commits are new, the rest is just a duplicate of https://github.com/awesomeWM/awesome/pull/868.

I somehow feel that https://github.com/awesomeWM/awesome/pull/878 will help with this, because it might make the wibox update its geometry when the screen is moved. Right now this part of the test fails.

Edit: Rebased on master now that #868 is merged. @Elv13 can you confirm that #878 is enough to make this test pass? (a `awful.wibox` / `awful.wibar` is automatically re-positioned when the geometry of its screen changes)

### DIFF
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -63,6 +63,11 @@ function wibox:find_widgets(x, y)
 end
 
 function wibox:get_screen()
+    if self.screen_assigned and self.screen_assigned.valid then
+        return self.screen_assigned
+    else
+        self.screen_assigned = nil
+    end
     local sgeos = {}
 
     for s in capi.screen do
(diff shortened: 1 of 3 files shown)


Elv13:
> @Elv13 can you confirm that #878 is enough to make this test pass?

If the screen correctly emit `property::geometry`, then it will do something. Whether or not it will do the right thing remain to be seen. Right now I think it would re-align the wibox correctly, but not stretch it. I will make sure it works before it's merged.

psychon:
Rebased on master and added 9912a8fd5fa4677b3caa318634aa17226beda9f9 to make the test succeed.
@Elv13 Any idea on how to fix this problem nicely? Should we count it as a bug in the "closest screen" function and/or say that awesome does not support configurations with overlapping screens? Should this be fixed by doing something similar to this commit, but only done nicely?

Elv13:
Well, my commit was
https://github.com/Elv13/awesome-1/commit/a66d0b68fb870b76bf7cf3581ab3245867f3a5f4

Sorry, I forgot about it. If I recall, when I last tried, the test almost passed. The only problem was the leak.

psychon:
I updated this again. I made `wibox` explicitly track the screen that is assigned to so that things also work with overlapping screens. This feels safer to me than the surprises that could otherwise happen (e.g. a possible scenario: while the screens are reconfigured, awesome sees an intermediate state with overlapping screens and lots of Lua state gets broken/lost).

Feel free to review this.

PR #888, opened 2016-05-09
doc: Add a system to share ldoc content across multiple files

Elv13:
This avoid copy-pasting

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -297,6 +297,9 @@ set(AWESOME_THEMES_PATH      ${AWESOME_DATA_PATH}/themes)
 
 
 if(GENERATE_DOC)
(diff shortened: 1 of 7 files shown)


Elv13:
This is a possible "fix" to address @psychon concern in the geometry PR that I shamelessly copy-paste documentation that will eventually get out of sync. It isn't exactly pretty, but it works. At some point, I will probably add a new "inherited method" category to the class own methods stand out.

I prefer this over a link to the parent class, as ldoc lack a clear inheritance hierarchy visualization, so most users would overlook it and fail to notice. If someone have a better idea, I am open to suggestion. This is (I think) a simple solution to the "repetitive doc out of sync" problem.

psychon:
Yay! :+1:

PR #890, opened 2016-05-09
Fix screen.fake_add()

psychon:
Commit 8a6787bd54 added screen.fake_add(). Commit 08845c7a4b made us cache a
screen's workarea in the struct screen_t. This new member needs to be
initialized to the screen's geometry when a new screen is added. Since both
these commits were developed concurrently, the workarea was not initialized in
screen.fake_add().

Fix this by calling in fake_add() the helper function added in 08845c7a4b.

### DIFF
--- a/objects/screen.c
+++ b/objects/screen.c
@@ -1042,9 +1042,9 @@ luaA_screen_fake_add(lua_State *L)
     s->geometry.y = y;
     s->geometry.width = width;
     s->geometry.height = height;
-    s->valid = true;
+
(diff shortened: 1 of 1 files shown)


blueyed:
Could there be a test for this?

psychon:
There kind-of is already a test, because travis builds for master currently fail. They fail because the fake screen's workarea is lots of 0s and thus things end up with negative sizes, but they fail. I'd say that this is enough.

PR #894, opened 2016-05-11
screen_update_primary: break when the primary screen is found

blueyed:


psychon:
...why? The `break` only breaks the inner loop, so it doesn't even skip going through all other outputs and screens. Even if it would, I don't think that this makes any difference runtime-wise, because the lists that are iterated here are really, really short.

blueyed:
Yeah.  Is there `break 2` in C?

While the lists are short, it's really a matter of "found, let's go on".

Elv13:
> Yeah. Is there break 2 in C?

Yes, `goto`, but please don't

psychon:
The best option would be to have a static helper function to do the search. That would provide `return` to do what you want.

Still, I think that this is highly irrelevant...

PR #895, opened 2016-05-11
gears.surface: add a traceback for "Failed to load" errors

blueyed:
This is useful to see where the error is coming from.

But maybe we should have a traceback with `gears.debug.print_error` in general?


psychon:
Looking at how `print_error` is used currently, `gears.protected_call` already adds a traceback and most other cases don't really need a traceback (perhaps `textbox:set_markup()`?). Perhaps it still makes sense to have `print_error` and `print_error_without_traceback`...

However, that doesn't stop this PR: :+1:

PR #898, opened 2016-05-12
tests: Don't use arbitrary build dir path

Elv13:
Calling the test script by hand from the build dir stopped working.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -17,15 +17,20 @@ if [ "$CI" = true ]; then
     set -x
 fi
 
-source_dir="$PWD"
+# Either the build dir is passed in $CMAKE_BINARY_DIR or we guess based on $PWD
 build_dir="$CMAKE_BINARY_DIR"
 if [ -z "$build_dir" ]; then
-    build_dir="$PWD/build"
+    if [ -d "$PWD/build" ]; then
+        build_dir="$PWD/build"
+    else
+        build_dir="$PWD"
+    fi
 fi
 
 # Change to file's dir (POSIXly).
 cd -P -- "$(dirname -- "$0")"
 this_dir=$PWD
+source_dir="$PWD/.."
 
 # Get test files: test*, or the ones provided as args (relative to tests/).
 if [ $# != 0 ]; then


Elv13:
Any idea why this fail on 5.2? It's caused by this patch because it happen all the time and PRs without it passes.

tests/run.sh
```diff
@@ -17,10 +17,14 @@ if [ "$CI" = true ]; then
     set -x
 fi
 
-source_dir="$PWD"
+source_dir="$(dirname "$(realpath "$(type -p $0)")")"/..
```
psychon: I don't think `type -0` is portable. My ZSH has different output than bash (zsh says e.g. `/usr/bin/test is /usr/bin/test`). Since this shell script runs with `/bin/sh`, that's a problem.

Also, right now I don't really get the problem that this tries to fix. AFAIK my patch to this script was "if $CMAKE_BINARY_DIR is set, use that, else do exactly the same thing as before my patch"

psychon:
Oh, I just noticed you say "from the build dir". Why are you calling the script from the build dir? How did that work before?

The script always used `$PWD/..` as (what was previously called) $root_dir. It then assumed that `$root_dir` is where the source exists (e.g. `$root_dir/utils/awesome-client`) and `$root_dir/build` is where the build is (e.g. it ran `$root_dir/build/awesome`).

How did it previously find `awesome_client` for you? That script is not copied to the build dir and thus this already shouldn't have worked before.

tests/run.sh
blueyed: FWIW: I tend to use this snippet:

```sh
# Change to current dir (POSIXly).
unset CDPATH
cd -P -- "$(dirname -- "$0")"
```

In this case this would include a `/..`.

blueyed:
@Elv13 
ping

Elv13:
> Why are you calling the script from the build dir?

Because it is where my terminal point to when I run `make`?

> How did that work before?
> How did it previously find awesome_client for you? That script is not copied to the build dir and thus this already shouldn't have worked before.

It did, but the real Awesome is also installed, so that may have helped

psychon:
Does this patch help you?
```
index 21909e1..9917659 100755
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -17,7 +17,6 @@ if [ "$CI" = true ]; then
     set -x
 fi
 
-source_dir="$PWD"
 build_dir="$CMAKE_BINARY_DIR"
 if [ -z "$build_dir" ]; then
     build_dir="$PWD/build"
@@ -26,6 +25,7 @@ fi
 # Change to file's dir (POSIXly).
 cd -P -- "$(dirname -- "$0")"
 this_dir=$PWD
+source_dir="$PWD/.."
 
 # Get test files: test*, or the ones provided as args (relative to tests/).
 if [ $# != 0 ]; then
(END)
```
Instead of assuming we are in the source dir when the script is started, this now assumes that the source dir is ".." to the dir in which `tests/run.sh` resides.

btw: I never cd into the build dir, I think...

psychon:
Ahh... and now I understood your patch. :-)
The following does basically the same and might work better than the patch I proposed above.
```
index 21909e1..3e0027b 100755
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -17,15 +17,20 @@ if [ "$CI" = true ]; then
     set -x
 fi
 
-source_dir="$PWD"
 build_dir="$CMAKE_BINARY_DIR"
 if [ -z "$build_dir" ]; then
-    build_dir="$PWD/build"
+    if [ -d "$PWD/build" ]; then
+        build_dir="$PWD/build"
+    else
+        # Uhm, ok... let's guess
+        build_dir="$PWD"
+    fi
 fi
 
 # Change to file's dir (POSIXly).
 cd -P -- "$(dirname -- "$0")"
 this_dir=$PWD
+source_dir="$PWD/.."
 
 # Get test files: test*, or the ones provided as args (relative to tests/).
 if [ $# != 0 ]; then
```

Elv13:
Yes, this work perfectly, thanks. Feel free to force push this PR and merge it.

psychon:
I changed this into my patch but kept @Elv13 as author since he did more of this patch. Since this is now mine and Emmanuel's patch, @blueyed gets to :+1: this. :-)

blueyed:
👍

PR #899, opened 2016-05-12
[RFC] naughty: display improvement of the notification client actions

Jajauma:
As the commit messages have got an ample background introduction let's put the long story short: a notification popup looks like this ATM:
action-id
while it could look like that:
action-text
Note the bottom line of a text which renders a stack of available actions along with the notification being sent by the notification client (over DBus and directly using the naughty API). The former picture has some action identifier (`disable-connected-notifications`) exposed directly to the user, while the latter one has the action ID replaced with a localized formal text passed by the notification client which clearly states that the user can opt out of any notifications regarding that kind of event. The action markup has been changed as well to emphasize the fact that the bottom area is clickable (hence the unicode hand and the styled link).

I guess this slight modification is going to be seen controversial since 
- ~~it introduces some naughty API breakage (the layout of the `actions` parameter of `naughty.notify` was altered to pass the description)~~
- someone might not like the new action markup.

I am ready to fix my work as you consider appropriate if it worths.

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -543,7 +543,7 @@ function naughty.notify(args)
             actionmarginbox:set_widget(actiontextbox)
             actiontextbox:set_valign("middle")
             actiontextbox:set_font(font)
-            actiontextbox:set_markup(string.format('<b>%s</b>', action))
+            actiontextbox:set_markup(string.format('☛ <u>%s</u>', action))
             -- calculate the height and width
             local w, h = actiontextbox:get_preferred_size(s)
             local action_height = h + 2 * margin
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -137,24 +137,20 @@ capi.dbus.connect_signal("org.freedesktop.Notifications", function (data, appnam
         if actions then
             args.actions = {}
 
-            local actionid
-            -- create actions callbacks
-            for i , v in ipairs(actions) do
-                if i % 2 == 1 then
-                    actionid = v
-                elseif actionid == "default" then
+            for i = 1,#actions,2 do
+                local action_id = actions[i]
+                local action_text = actions[i + 1]
+
+                if action_id == "default" then
                     args.run = function()
                         sendActionInvoked(notification.id, "default")
                         naughty.destroy(notification, naughty.notificationClosedReason.dismissedByUser)
                     end
-                    actionid = nil
-                elseif actionid ~= nil then
-                    local action = actionid
-                    args.actions[actionid] = function()
-                        sendActionInvoked(notification.id, action)
+                elseif action_id ~= nil and action_text ~= nil then
+                    args.actions[action_text] = function()
+                        sendActionInvoked(notification.id, action_id)
                         naughty.destroy(notification, naughty.notificationClosedReason.dismissedByUser)
                     end
-                    actionid = nil
                 end
             end
         end


lib/naughty/core.lua
```diff
@@ -428,8 +428,11 @@ end
 -- @tparam[opt] func args.callback Function that will be called with all arguments.
 --   The notification will only be displayed if the function returns true.
 --   Note: this function is only relevant to notifications sent via dbus.
--- @tparam[opt] table args.actions Mapping that maps a string to a callback when this
---   action is selected.
+-- @tparam[opt] table args.actions Table describing all of the actions that are
+--   supported by the notification client. The table key is an action identifier.
+--   The table value is a record table indexed by two keys: `text` is a localized
+--   description string which is displayed to a user, and `callback` is a function
+--   called when this action is selected.
```
psychon: Why do you have to change this? Are you worried that two actions will have the same text?
If not, then the old `text = callback` is just as expressive as `{ text = "foo", callback = callback }` (although it may be more future proof and extendable...

I'm not really sure about what is better. Feel free to leave it like this if no one else complains.

psychon:
:+1: by me, but I'll give others a chance to complain.
@Elv13 do you care about backwards compatibility of an API that (as far as I remember) is not present in 3.5?

lib/naughty/core.lua
Jajauma: The naughty [dbus handling code](https://github.com/awesomeWM/awesome/blob/master/lib/naughty/dbus.lua#L188) is the one who uses public API to display a received notification over DBus so this change is required to be able to pass three pieces (action-id, action-text, action-callback) instead of just two (action-id, action-callback). But it definitely seems (since you mention it) that the code in the core.lua doesn't care about action-id, which serves solely for the needs of the DBus dispatching.

So I guess now I can alter the PR to just pass the action nice description and callback in the old fashion `actions[text] = callback` table. I'm not sure if multiple actions with the same title make sense anyway. So let's wait some for the others and decide then.

Jajauma:
Somehow I reverted the naughty.notify API breakage initially as a PoC but it ended up to be in PR automatically :(. This leaves it with one extra commit and me with absolutely no idea what should I do about them (squash/rebase/push -f/drop and start from the scratch...)

psychon:
Just leave it like this for now. Let's first figure out the idea and then handle the detail (also, now one can look at both versions of the idea: The individual commits show version A, the "files changes"-tab at the top of this pull request shows version B :-) ).

psychon:
Apparently no one else has any strong opinions about this (or everyone just didn't find the time yet...).
@Jajauma Could you squash the "revert the API-breakage"-commit into the first commit (so that only two commits are left: Use user-readable text instead of ID, nice icon unicode for the texts) and then force-push the result to your naughty-action-text branch?

Jajauma:
I've finished with squashing, comparing, building and testing on my side. All of the API breakage notes were cleaned out.

psychon:
Thanks! :+1:

Elv13:
> @Elv13 do you care about backwards compatibility of an API that (as far as I remember) is not present in 3.5?

Not at all. Go ahead and break the API. I am not against breaking the API (I do it all the time too). I just want to save some pain to users upgrading from the previous official API.

:+1:

blueyed:
Thanks!  (I am using dunst for my notifications myself - but it's good to make naughty less naughty regarding the standards)

Issue #891, opened 2016-05-10
awesome systray triggers Gtk+2 crash?


orebokech:
Hi,

For a few years now, awesome has been known to trigger a crash in Gtk+2 related to the systray implementation. This was reported before against several apps like Google Chrome or Thunar, before finally landing on the Gtk+ bugtracker, but the maintainers seem unwilling to do anything about it:

https://bugzilla.gnome.org/show_bug.cgi?id=747280#c19

Other reports:
https://bugs.chromium.org/p/chromium/issues/detail?id=430910
https://bugzilla.xfce.org/show_bug.cgi?id=11035
https://awesome.naquadah.org/bugs/index.php?do=details&task_id=1286 (in the old awesome bugtracker)

I think we can all agree that it's not a good thing that Gtk crashes its users because of something unexpected with the systray. However, given that the only reports of this bug are from awesome users, it's not entirely unreasonable to think that awesome's systray implementation does do things wrong, or at least not like the other systray implementations.

It would be very helpful if the awesome maintainers could take a closer look at this issue and perhaps work with the Gtk+ maintainers to get to the bottom of this. Having Chrome crash every time the file dialog opens gets tiring very quickly (and maintaining a patched Gtk package just for this is annoying).

Thank you!

psychon:
I don't think I ever heard about this before nor did I notice it. So what is my system doing differently from yours? Weird...

orebokech:
Do you use Google Chrome with apps or extensions that create a systray icon, like Hangouts?

From the activity on the Chrome and Gtk bug trackers it's apparent that many users are affected, although I guess if you don't use any Gtk2 apps that create a tray icon you could never see this bug...

psychon:
Nope

psychon:
Ok, staring at GTK2's code just makes me say "please switch to GTK3". Since I doubt that you will do so: Can you run a small program that crashes under xtrace (best use `xtrace --relative-timestamps`), make it crash and show me the resulting output? And no, you really do not want to run chromium under xtrace, the output will be huge.

What I learned from the code: Somehow `gtk/gtkplug-x11.c` works together with `gtk/gtktrayicon-x11.c`. I still don't know why the latter crashes (the crash is in a line where nothing "interesting" happens), but the former is apparently responsible for "things". Also, I'm not sure if the suggested patch works by just breaking systray support completely...

psychon:
> Do you use Google Chrome with apps or extensions that create a systray icon, like Hangouts?

[...]

PR #904, opened 2016-05-14
Some Xembed fixes

psychon:
Here are two things that I noticed while looking into https://github.com/awesomeWM/awesome/issues/891. Neither helps with the issue described there (I think...), but hopefully conforming to the spec somewhere somewhen helps someone.

As mentioned in the second commit message, this isn't properly tested. Let's just see if things still work...

### DIFF
--- a/common/xembed.c
+++ b/common/xembed.c
@@ -24,6 +24,9 @@
 #include "common/util.h"
 #include "common/atoms.h"
 
+/* I should really include the correct header instead... */
+void luaA_systray_invalidate(void);
+
 /** Send an XEMBED message to a window.
  * \param connection Connection to the X server.
  * \param towin Destination window
@@ -141,6 +144,7 @@ xembed_property_update(xcb_connection_t *connection, xembed_window_t *emwin,
             xembed_window_deactivate(connection, emwin->win);
             xembed_focus_out(connection, emwin->win);
         }
+        luaA_systray_invalidate();
     }
 }
 
--- a/event.c
+++ b/event.c
@@ -770,14 +770,6 @@ event_handle_unmapnotify(xcb_unmap_notify_event_t *ev)
 
     if((c = client_getbywin(ev->window)))
         client_unmanage(c, true);
-    else
-        for(int i = 0; i < globalconf.embedded.len; i++)
-            if(globalconf.embedded.tab[i].win == ev->window)
-            {
-                xembed_window_array_take(&globalconf.embedded, i);
-                xcb_change_save_set(globalconf.connection, XCB_SET_MODE_DELETE, ev->window);
-                luaA_systray_invalidate();
-            }
 }
 
 /** The randr screen change notify event handler.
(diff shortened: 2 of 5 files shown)


psychon:
The previous version broke Skype's tray icon (there are specifications and there is the reality... Skype (Qt?) lives in the reality). This is fixes now.

Adding to next.

common/xembed.c
```diff
@@ -24,6 +24,9 @@
 #include "common/util.h"
 #include "common/atoms.h"
 
+/* I should really include the correct header instead... */
```
blueyed: Will you do it? :)

event.c
```diff
@@ -724,10 +725,13 @@ event_handle_maprequest(xcb_map_request_event_t *ev)
     if(wa_r->override_redirect)
         goto bailout;
 
-    if(xembed_getbywin(&globalconf.embedded, ev->window))
+    if((em = xembed_getbywin(&globalconf.embedded, ev->window)))
     {
         xcb_map_window(globalconf.connection, ev->window);
         xembed_window_activate(globalconf.connection, ev->window);
+        /* This is not the correct way to set this, but let's be gentle... */
+        em->info.flags |= XEMBED_MAPPED;
```
blueyed: What would be the correct way?

common/xembed.c
psychon: Nope. At least I call myself out for this uglyness. ;-)

The problem is that the stuff in `common/` should be self-contained and this dependency goes against that and i have no good idea what to do about it.

event.c
psychon: Actually set an `_XEMBED_INFO` property on the tray icon, like the xembed specification say (which Qt/Skype does not set at all). Neither of the xembed spec nor the systray spec mention anything about MapWindow.

blueyed: Would be nice to have this in the comment with the code then.

blueyed:
:+1: (just quickly tested it locally)

psychon:
Ok, I'll squash the following diff into this when I find the time:
```diff
diff --git a/event.c b/event.c
index 74591b2..c4e2ea3 100644
--- a/event.c
+++ b/event.c
@@ -729,7 +729,11 @@ event_handle_maprequest(xcb_map_request_event_t *ev)
     {
         xcb_map_window(globalconf.connection, ev->window);
         xembed_window_activate(globalconf.connection, ev->window);
-        /* This is not the correct way to set this, but let's be gentle... */
+        /* The correct way to set this is via the _XEMBED_INFO property. Neither
+         * of the XEMBED not the systray spec talk about mapping windows.
+         * Apparently, Qt doesn't care and does not set an _XEMBED_INFO
+         * property. Let's simulate the XEMBED_MAPPED bit.
+         */
         em->info.flags |= XEMBED_MAPPED;
         luaA_systray_invalidate();
     }
```

psychon:
Ok. I updated this and it could now be merged.

Issue #572, opened 2015-11-25
Temporarily disable awful.client.focus.history?


blueyed:
I am thinking about fixing an issue, where the raising/focusing of clients through https://github.com/blueyed/awesome-cyclefocus (when cycling through its history) causes those to be added to `awful.client.focus`' history, and will then be used by the autofocus handler.

I see two fixes:
1) temporarily allow for the `awful.client.focus` history handler / tracking to be disabled
2) overwrite the `awful.client.focus.history.get` handler (from the `cyclefocus` module)

The issue itself can be seen with this test case:
1. there are two clients on a tag, both maximized
2. client A is on top of the window stack and therefore the only visible one
3. You alt-tab through your windows, which brings client B on top (I use the ontop property, and not `raise`), but you do not select/activate/raise it.  You end alt-tabbing on another client on another tag.
4. When going back to the tag with clients A and B, the autofocus handler will select client B (because it is the last in the history stack of that tag).

The autofocus handler will use `raise=false` in the hints (see https://github.com/awesomeWM/awesome/issues/217).  As said in #217 already, I could imagine to use `raise=true` in case the client is not (at least partly) visible.

This is what I am using in cyclefocus currently:
```
-- Override awesome's internal history handler to handle `ignore_focus_signal`.
capi.client.disconnect_signal("focus", awful.client.focus.history.add)
cyclefocus.wrapped_awful_client_focus_history_add = function(c)
    if ignore_focus_signal then
        return
    end
    return awful.client.focus.history.add(c)
end
capi.client.connect_signal("focus", cyclefocus.wrapped_awful_client_focus_history_add)
```
The issue itself is resolved/fixed by that, but I wondered if there's a better method, and only came up with replacing the signal handler after having written most of this already.

psychon:
Hm. Tough question...
Doing (1) in a safe way seems hard. If this is just a boolean, "something random" could forget to re-set the boolean to its proper value. However, I don't know how to make this safer, because the resulting API would be ugly (having a function that focuses a client without adding it to the history? That doesn't work if there are more than one place needing such an exception). Perhaps some signal-with-hints thing to focus a client? Doesn't seem too nice either....

psychon:
Another idea: If something forgets to enable the focus history again, it's their fault, so do the following in `client.lua` (this does the right thing if two different modules want to disable this at the same time):
```lua
do
    local disabled_count = 1
    function client.focus.history.disable_tracking()
        disabled_count = disabled_count + 1
[...]

PR #906, opened 2016-05-14
awful.client.history: allow to disable history tracking

blueyed:
Based on the code from Uli Schlachter at
https://github.com/awesomeWM/awesome/issues/572#issuecomment-160080378.

Fixes https://github.com/awesomeWM/awesome/issues/572.

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -1130,8 +1130,6 @@ capi.client.add_signal("marked")
 -- @signal unmarked
 capi.client.add_signal("unmarked")
 
-capi.client.connect_signal("focus", client.focus.history.add)
-
 -- Add clients during startup to focus history.
 -- This used to happen through ewmh.activate, but that only handles visible
 -- clients now.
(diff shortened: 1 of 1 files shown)


blueyed:
Does the same make sense for autofocus, too?

lib/awful/client.lua
```diff
@@ -1142,9 +1140,50 @@ capi.client.connect_signal("manage", function (c)
...
+        end
+        return disabled_count == 0
+    end
+
+    --- Is history tracking enabled?
+    -- @treturn[0] true True is history tracking is enabled.
+    -- @treturn[1] int The number of times that tracking has been disabled.
+    -- @function awful.client.focus.history.is_enabled
+    function client.focus.history.is_enabled()
+        return disabled_count == 0 or disabled_count
```
psychon: Did you mean `,` instead of `or`?

blueyed: No.. it should have been either `true` (`disabled_count` == 0), or `disabled_count`.
What you mean is returning two values `bool` and the number, right?
Makes more sense and seems to be less magical that way - amended.

psychon: Apparently I still have problems understanding `@return[42]`.

Yup, you are right and yup, I am right. :-)

PR #909, opened 2016-05-15
Fix awful.screen.getbycoord

psychon:
This started as a fix for the issue that @Elv13 reported here: https://github.com/awesomeWM/awesome/pull/878#issuecomment-219272864
Then I noticed that this function is also broken when no screens are defined and decided to fix this as well.


Elv13:
Sorry for not doing this myself. It was very late and I wanted a reviewable version of my work immediately, so I went for a cheap workaround (the +1).

psychon:
No problem.

Issue #170, opened 2015-03-19
Focus problems in multi monitor setup with Google Chrome / Chromium


HerrBerg:
**Problem:**
I use awesomewm v3.5.6 and have the following problem in a multi-monitor setup:
When I use chromium browser I cannot rearrange tabs using the mouse. When I try to drag a tab I instantly get a new window with that tab. After that it is not possible to move this tab back into the main window.
Another issue: in the bookmark manager (of chromium) it is not possible to reagange the bookmarks using drak and drop either.

**Why I think this is an awesomewm issue:**
Today I found, that these issues only occur on screen two and three.
On screen one everything works fine. 
More precisely: it only works on the primary monitor (see below) 

**What I did to setup monitors:**
`xrandr --output eDP1 --auto --primary --output DP4 --auto --right-of eDP1 --output HDMI1 --auto --right-of DP4`

the screen where I can move tabs moves with the `--primary` flag

blueyed:
Thanks for filing this.  It is a known issue AFAIK.

Elv13:
Yes, reported multiple time over the years on IRC.

vdrandom:
It is the case with most chromium based applications, including github's Atom editor. The only exclusion among chromium based browsers I know is Opera. So if you are desperate to use a chromium-based browser for some reason, you can stick with Opera, at least for now.

chpill:
I'm experiencing the exact same bug. The drag and drop inside a chrome web page works only on the primary monitor. 

Does anyone have any idea why chrome behaves this way?

berlam:
Just for the docs. This bug was reported to the chromium team also:
https://code.google.com/p/chromium/issues/detail?id=132706

Already closed unfortunately.

schmellow:
I wonder where one should start looking to pinpoint the problem. Since tab dragging works in i3,  i was skipping through both awesome and i3 source code, trying to find significant differences  in window and mouse handling code to get some clues, but to no avail (even though significant differences are aplenty, haha).

So, tab dragging works in i3 both for primary and secondary displays by default. BUT, while playing with chromium i was able to reproduce exactly same erroneous behaviour in i3.

Steps:
1) Open chromium on primary screen
2) Make it floating
3) Drag it to secondary screen

Dragged window and all its children (created by ctrl+n command) will exhibit the behaviour in question while being on secondary monitor.

Alternative steps:
1) Open chromium on secondary screen (tab dragging works)
2) Make it floating
[...]

PR #910, opened 2016-05-15
Attempt at more correct handling of NET_CURRENT_DESKTOP property.

schmellow:
First attempt at fixing issue #170 

Update NET_CURRENT_DESKTOP property on:
1) Tag select;
2) Focus change;
3) Client screen change (mostly to support the case when client is dragged between screens by mouse).

NET_CURRENT_DESKTOP is assigned value equal to index of the first tag that is:
1) Selected;
2) Belongs to currently focused screen.

Caveats:
1) "Focused screen" is screen under the mouse cursor. Judging by awful.screen.focused that may not always be the case, but pulling lua state properties and functions that are not already there into the C layer is beyond my understanding of how to do it right for now.
2) For the same reason as previous: tag is considered belonging to screen if it has at least one client, belonging to that screen. This means that when mouse is on empty tag, it won't be set as current desktop. Also if no tags found this way at all - default to old behaviour (just first selected tag). Hacky...
3) I don't like really calling ewmh_update_net_current_desktop each time in client_resize_do: if we move windows by mouse from screen to screen, then in screen_client_moveto it should get refocused and refocusing does the update, so update effectively gets called twice. In reality, for some reason, it only works when moving from secondary to primary screen, not the other way around. I'm not sure on performance implications, maybe it's ok to let it be.

### DIFF
--- a/ewmh.c
+++ b/ewmh.c
@@ -228,6 +228,12 @@ ewmh_init(void)
     luaA_class_connect_signal(L, &client_class, "property::titlebar_left" , ewmh_client_update_frame_extents);
     luaA_class_connect_signal(L, &client_class, "property::border_width" , ewmh_client_update_frame_extents);
     luaA_class_connect_signal(L, &client_class, "manage", ewmh_client_update_frame_extents);
+    /* NET_CURRENT_DESKTOP handling */
+    luaA_class_connect_signal(L, &client_class, "focus", ewmh_update_net_current_desktop);
+    luaA_class_connect_signal(L, &client_class, "unfocus", ewmh_update_net_current_desktop);
+    luaA_class_connect_signal(L, &client_class, "tagged", ewmh_update_net_current_desktop);
+    luaA_class_connect_signal(L, &client_class, "untagged", ewmh_update_net_current_desktop);
+    luaA_class_connect_signal(L, &tag_class, "property::selected", ewmh_update_net_current_desktop);
 }
 
 /** Set the client list in stacking order, bottom to top.
@@ -256,14 +262,15 @@ ewmh_update_net_numbers_of_desktop(void)
 			_NET_NUMBER_OF_DESKTOPS, XCB_ATOM_CARDINAL, 32, 1, &count);
 }
 
-void
-ewmh_update_net_current_desktop(void)
+int
+ewmh_update_net_current_desktop(lua_State *L)
 {
-    uint32_t idx = tags_get_first_selected_index();
+    uint32_t idx = tags_get_current_or_first_selected_index();
 
     xcb_change_property(globalconf.connection, XCB_PROP_MODE_REPLACE,
                         globalconf.screen->root,
                         _NET_CURRENT_DESKTOP, XCB_ATOM_CARDINAL, 32, 1, &idx);
+    return 0;
 }
 
 void
--- a/objects/tag.c
+++ b/objects/tag.c
@@ -91,23 +91,6 @@
  * @function set_newindex_miss_handler
  */
 
-/** Tag type */
-struct tag
-{
-    LUA_OBJECT_HEADER
-    /** Tag name */
-    char *name;
-    /** true if activated */
-    bool activated;
-    /** true if selected */
-    bool selected;
-    /** clients in this tag */
-    client_array_t clients;
-};
-
-static lua_class_t tag_class;
-LUA_OBJECT_FUNCS(tag_class, tag_t, tag)
-
 
 void
 tag_unref_simplified(tag_t **tag)
@@ -139,7 +122,6 @@ tag_view(lua_State *L, int udx, bool view)
     {
         tag->selected = view;
         banning_need_update();
-        ewmh_update_net_current_desktop();
 
         luaA_object_emit_signal(L, udx, "property::selected", 0);
     }
@@ -220,15 +202,29 @@ is_client_tagged(client_t *c, tag_t *t)
     return false;
 }
 
-/** Get the index of the first selected tag.
- * \return Its index.
+/** Get the index of the tag with focused client or first selected 
+ * \return Its index
  */
 int
-tags_get_first_selected_index(void)
+tags_get_current_or_first_selected_index(void)
 {
+    /* Consider "current desktop" a tag, that has focused window,
+     * basically a tag user actively interacts with.
+     * If no focused windows are present, fallback to first selected.
+     */
+    if(globalconf.focus.client)
+    {
+        foreach(tag, globalconf.tags)
+        {
+            if((*tag)->selected && is_client_tagged(globalconf.focus.client, *tag))
+                return tag_array_indexof(&globalconf.tags, tag);
+        }
+    }
     foreach(tag, globalconf.tags)
+    {
         if((*tag)->selected)
             return tag_array_indexof(&globalconf.tags, tag);
+    }
     return 0;
 }
 
--- a/ewmh.h
+++ b/ewmh.h
@@ -31,7 +31,7 @@ typedef struct client_t client_t;
 
 void ewmh_init(void);
 void ewmh_update_net_numbers_of_desktop(void);
-void ewmh_update_net_current_desktop(void);
+int ewmh_update_net_current_desktop(lua_State *);
 void ewmh_update_net_desktop_names(void);
 int ewmh_process_client_message(xcb_client_message_event_t *);
 void ewmh_update_net_client_list_stacking(void);
--- a/objects/tag.h
+++ b/objects/tag.h
@@ -24,13 +24,30 @@
 
 #include "client.h"
 
-int tags_get_first_selected_index(void);
+int tags_get_current_or_first_selected_index(void);
 void tag_client(lua_State *, client_t *);
 void untag_client(client_t *, tag_t *);
 bool is_client_tagged(client_t *, tag_t *);
 void tag_unref_simplified(tag_t **);
 
 ARRAY_FUNCS(tag_t *, tag, tag_unref_simplified)
+
+/** Tag type */
+struct tag
+{
+    LUA_OBJECT_HEADER
+    /** Tag name */
+    char *name;
+    /** true if activated */
+    bool activated;
+    /** true if selected */
+    bool selected;
+    /** clients in this tag */
+    client_array_t clients;
+};
+
+lua_class_t tag_class;
+LUA_OBJECT_FUNCS(tag_class, tag_t, tag)
 
 void tag_class_setup(lua_State *);
 


objects/client.c
```diff
@@ -969,6 +969,8 @@ client_focus_update(client_t *c)
         luaA_object_emit_signal(L, -1, "focus", 0);
 
     lua_pop(L, 1);
+    
+    ewmh_update_net_current_desktop();
```
psychon: Instead of having callbacks everywhere (e.g. here), could you connect to signals? In function `ewmh_init` various callbacks are connected and this even already connects a callback to the "focus" signal. You can just copy that line and connect `ewmh_update_net_current_desktop`. (same for other places where you add callbacks: Connect to `property::screen` on clients,

objects/tag.c
```diff
@@ -220,16 +221,52 @@ is_client_tagged(client_t *c, tag_t *t)
...
+}
+
+/** Check if tag has clients on specified screen.
+ * \param t the tag
+ * \param s the screen
+ * \return true if the are client is tagged with the tag, false otherwise.
+ */
+bool
+is_tag_on_screen(tag_t **t, screen_t *s)
+{
```
psychon: Can you make this `static inline` and remove it from the header? It's only used internally.

(Even better would be to get rid of it completely, but I don't think that you'll like that idea...)

psychon:
I can't say that I really like this, but I don't really know how to work around Chromium's requirements in a better way...
Let's see what others say.

schmellow:
>Instead of having callbacks everywhere (e.g. here), could you connect to signals? 

Will do. Signature has to be changed to static int with lua_State argument, am i right?
I can't connect to tag_class though, as it is static in objects/tag.c, so original callback in tag_view will have to stay.

>Can you make this static inline and remove it from the header? It's only used internally.
>(Even better would be to get rid of it completely, but I don't think that you'll like that idea...)

I can get rid of separate function and move code inside, no problem.


>I can't say that I really like this, but I don't really know how to work around Chromium's requirements in a better way...

Actually i can think of at least two alternative solutions, that may be (or may be not) cleaner.
The common idea is to do most of the work from inside lua, because getting target tag here is a matter of "awful.screen.focused().selected_tag".

Then first alternative would be:
1) Get focused screen as screen.focused, get tag as this screen selected_tag
2) I guess there is no equivalent of globalconf.tags, so to get index we will have to sum all screen.tags lengts up to focused screen and add screen-local index of our target tag.
3) Somehow set NETWM property, either directly from lua (if we can do it without depending on external tools like xprop), or call a C function (but i guess you don't do that, and you don't do call to lua functions from C, as far as i get it - its all objects and properties and signals only)

Second alternative:
1) Get target tag
2) Drop it down to C level in form of pointer to tag in globalconf.tags

This one, in fact, i tried to do first: i've added a signal "update_current" to tag_class, then made a function that gets the tag and emits this signal from it. The function itself was hooked to client.focus and tag.selected signals. But it did not quite work, i've messed up and got wrong indices in the end. Still if it is viable solution and if it would be preferable, i can make a second pass on it.

schmellow:
I've done requested changes and reviewed the whole alghorithm, since i've encountered several corner cases that were producing troubles with the first version (sloppy focus and switching to empty tag on separate desktop generated some wonky behaviour, and some keyboard-only window interactions worked wrong).

First and foremost, there is actually no need to look for "focused" screen. I really can't think of examples where current desktop won't be the desktop where the focused window user interacting with is. So we can just use tag that has currently focused window as reference.

Now, EWMH_CURRENT_DESKTOP updates are triggered only on client focus and client screen change. I've removed update callback from tag_view, and unless there is a some specific reason i don't know of, we don't really need it to be here explicitely: if we switch to tag with windows, then focus event will take care of the current desktop update, and if there are no clients then there is no one to complain, i think.

So far it seems to be working for all types of interactions, both keyboard-only and mouse-driven. 

The only thing i have to note is that apparently client screen is detected by upper left coordinate (as opposed to {(x+width)/2, (y+height)/2} middlepoint commonly used in other window managers). So if you are dragging tab from window on left screen to window on right screen, you got to drag it all the way, so client screen actually switches. Or one can just switch detection to middle point in source code.

PS. Oh, travis failed for some reason, gonna look for why.

ewmh.c
```diff
@@ -256,14 +259,15 @@ ewmh_update_net_numbers_of_desktop(void)
 			_NET_NUMBER_OF_DESKTOPS, XCB_ATOM_CARDINAL, 32, 1, &count);
 }
 
-void
-ewmh_update_net_current_desktop(void)
+int
+ewmh_update_net_current_desktop(lua_State *L)
```
blueyed: It does not use lua_State (anymore)?!

objects/client.c
```diff
@@ -969,7 +969,7 @@ client_focus_update(client_t *c)
         luaA_object_emit_signal(L, -1, "focus", 0);
 
     lua_pop(L, 1);
-
+
```
blueyed: Please revert this.

objects/tag.c
```diff
@@ -220,16 +219,45 @@ is_client_tagged(client_t *c, tag_t *t)
...
+            {
+                if((*tag)->clients.tab[i] == client)
+                {
+                    found_current = true;
+                    break;
+                }                
+            }
+            if(found_current)
+            {
+                current_tag = tag;
```
blueyed: No need for found_current - just assign to current_tag directly?!

blueyed: And it could even return directly?!

blueyed:
Thanks for this!
Left some comments.

ewmh.c
schmellow: >It does not use lua_State (anymore)?!

Uh, it did not, and does not. Still it should have appropriate signature to be used as signal handler, or so i thought...correct me if i'm wrong.

objects/tag.c
```diff
@@ -139,7 +139,6 @@ tag_view(lua_State *L, int udx, bool view)
     {
         tag->selected = view;
         banning_need_update();
-        ewmh_update_net_current_desktop();
```
psychon: This line is still needed, I think, for the case when no client is currently focused and you are switching between tags (and the tag you switch to also doesn't have any clients).

objects/tag.c
```diff
@@ -220,16 +219,32 @@ is_client_tagged(client_t *c, tag_t *t)
...
+            if(!first_tag)
+                first_tag = tag;
+            for(int i = 0; i < (*tag)->clients.len; i++)
+            {
+                if((*tag)->clients.tab[i] == client)
+                    return tag_array_indexof(&globalconf.tags, tag);
+            }
+        }
+     }
+     return tag_array_indexof(&globalconf.tags, first_tag);
```
psychon: I would suggest:
```C
if (globalconf.focus.client)
{
    foreach(tag, globalconf.tags)
    {
        if((*tag)->selected && is_client_tagged(globalconf.focus.client, *tag))
            return tag_array_indexof(&globalconf.tags, tag);
    }
}
foreach(tag, globalconf.tags)
{
    if ((*tag)->selected)
        return tag_array_indexof(&globalconf.tags, tag);
}
return 0;
```
I consider this easier to read (I can immediately see that it first tries to find a tag for the focused client, then just any selected tag) and it also works "correctly" if no tag at all is currently selected.

ewmh.c
```diff
@@ -228,6 +228,9 @@ ewmh_init(void)
     luaA_class_connect_signal(L, &client_class, "property::titlebar_left" , ewmh_client_update_frame_extents);
     luaA_class_connect_signal(L, &client_class, "property::border_width" , ewmh_client_update_frame_extents);
     luaA_class_connect_signal(L, &client_class, "manage", ewmh_client_update_frame_extents);
+    /* NET_CURRENT_DESKTOP handling */
+    luaA_class_connect_signal(L, &client_class, "property::screen", ewmh_update_net_current_desktop);
```
psychon: I don't think that this needs to connect to `property::screen`. The function `ewmh_update_net_current_desktop` does not use a client's screen. However, the `tagged` and `untagged` signals on `client_class` are necessary (if the currently focused client is moved to another tag).

objects/tag.c
schmellow: It depends on what for. First focused client will trigger the update, so actual clients depending on this atom should be fine. Pagers depend on NET_CURRENT_DESKTOP, but i doubt they worked correctly before with how it was handled.

This, of course, can be reverted, but then we also might need to update on mouse::enter events, to behave correctly.
Example: let's say we have chromium focused on one screen, then we switch to empty tag on second screen. Client on first screen is still focused, but CURRENT_DESKTOP now points to tag on the second screen. Now to move tabs we have to explicitely refocus chromium window (and with sloppy focus, for example, just moving mouse back to already focused window does not trigger focus event).

So we will need to update on screen change, on focus (to work properly with keyboard driven interactions), on tag switch AND on mouse enter event. Again i'm not sure about performance implications there.
No really good solutions, i'm afraid.

ewmh.c
blueyed: Oh, that might be the case, of course.

objects/tag.c
psychon: Sorry, but I don't understand. What you are saying is:
- For simplicity: Screen 1 has tag 1-1, screen 2 has tag 2-1 and tag 2-2
- Chromium is on 1-1, tags 1-1 and 2-2 are selected
- Thus, `_NET_CURRENT_DESKTOP` is 1-1 currently (it contains the focused client)
- You deselect 2-2 and select 2-1 (switch to another tag on the second screen)
- `_NET_CURRENT_DESKTOP` does not change, because the focused client is still on 1-1 (well, awesome will set the value that this property already has again, but that does not matter)

Since `ewmh_update_net_current_desktop` does not look at the mouse pointer, mouse moves do not matter to it. What the function does look at is focused client and selected tag, thus it needs to connect to "focus" and "unfocus" on client (oh, unfocus is missing!), "tagged" and "untagged" also on client and "property::selected" on tags.

schmellow: Yeah...you are right, i've confused myself it seems - what i've said would stand only for old version of the solution, not the current. 

Do we really need both unfocus and untagged? 

Regarding property::selected, i'll have to revert to the direct call to ewmh_update_current_desktop(NULL) in tag.c, since tag class is not accessible outside of tag.c

psychon: Why we need "unfocus": Tags 1 and 2 are selected. The focused client is on tag 2, so `_NET_CURRENT_DESKTOP` has value 2. You do `client.focus = nil`. Now tag 1 is the first selected tag and `_NET_CURRENT_DESKTOP` needs to be updated.

Why we need "untagged": No client is selected. Tags 1 and 2 are selected, so the property has value 1. You deselect tag 1. Now the property should have value 2.

For accessing `tag_class`: Perhaps it is easier to move it into `objects/tag.h`, just like `objects/client.h` contains `client_class`.

schmellow: Got it, thank you.

I'll move tag class to header then and push the next set of changes later today probably.

psychon:
Thanks. The total diff looks fine to me. I think the individual commits can be squashed into a single one and this can be merged. We can do that for you if this sentence confuses you. :-)

Also thanks for bearing with us and our comments.

@blueyed Any remarks?

blueyed:
I dislike PRs without tests, we need to improve this.. ;/

Looks fine to me otherwise - except for the punctuation at the end of the commit subject.. :D

psychon:
@blueyed Here is a unit test for this https://gist.github.com/psychon/78c16b23b9a88b178dbe08430588fa16
Should I open a separate PR for this once this is merged or do you want to add this here after this was squashed? (Or does @schmellow want to `git am -s` this patch to this PR after squashing the other commits?)

schmellow:
I've squashed previous commits (hope i got it right). Regarding unit test, i'd say its your call. I will pull it into my branch if you want.

psychon:
Thanks. I merged this and also added my test. I hope the test is good enough. If not, feel free to be angry at me and tell me which part of it is wrong.

syphoxy:
I just want to unnecessarily confirm that this patch fixed the issue for me. I compiled and installed master today.

thanks for fixing this, everyone!

psychon:
The thanks go to @schmellow, he came up with a working approach for a fix, implemented it and wasn't demotivated by the several rounds of review we force upon him.

Thanks!

PR #911, opened 2016-05-16
Travis: redirect stderr for git-clone/git-push/curl

blueyed:
git-clone for example might reveal the token on network errors.


blueyed:
@psychon 
Changed it like for awesome-www now.

Having no errors is really unhelpful, the `sed` should now take care of not revealing the token.

psychon:
Still fine with me

PR #912, opened 2016-05-16
tag: Avoid a race condition when deleting tags.

Elv13:
Since the screen removal patchset, the tags properties were
cleaned too early. This caused code connecting to "property::activated"
to be called with the tag already partially deleted. For code depending
on those properties, such as radical.impl.taglist, this caused errors.

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -295,7 +295,6 @@ function tag.object.delete(self, fallback_tag, force)
 
     -- delete the tag
     data.tags[self].screen = nil
-    data.tags[self] = nil
     self.activated = false
 
     -- Update all indexes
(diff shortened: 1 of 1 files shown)


psychon:
What if some code wants to show/hide/show/hide/... a tag? Clearing all Lua-properties just because `activated` was set to `false` is weird for that kind of usage. Doing a delayed call to delete the properties in `awful.tag.object.delete` is weird for the same reason. (Although it might be harder to hit a bug due to this since it requires a tag to be deactivated and reactivated in the same main loop iteration)

Doing this properly might be ugly... The properties could be moved to a `pending_removal` table in `:delete()` and a delayed call is started. The delayed call clears the entry in this table is the tag is still deactivated. If the tag is re-activated, the properties are moved back.

I really, really want to avoid this complexity. Does someone have a better idea?

Elv13:
What about just removing this and hoping it will get garbage collected anyway?

psychon:
At least that would fix various issues, yeah.
I have no good idea on how to make sure that no leaks happen for sure.

Elv13:
Ok, I merged the simplified version. As it solve a bunch of real errors now, it's better than waiting for a perfect solution. So far, it doesn't seem to leak.

PR #913, opened 2016-05-17
Gear.object properties

Elv13:
This close the loop and will help bring coherent API across the other relevant classes. Similar systems have already been introduced in other Awesome area. This one will eventually deprecate the one found in `wibox.widget.base_widget`

### DIFF
--- a/lib/gears/object.lua
+++ b/lib/gears/object.lua
@@ -122,10 +122,57 @@ function object:emit_signal(name, ...)
     end
 end
 
---- Returns a new object. You can call :emit_signal(), :disconnect_signal,
--- :connect_signal() and :add_signal() on the resulting object.
-local function new()
+local function get_miss(self, key)
+    local class = rawget(self, "_class")
+
+    if rawget(self, "get_"..key) then
+        return rawget(self, "get_"..key)(self)
+    elseif class and class["get_"..key] then
+        return class["get_"..key](self)
+    elseif class then
+        return class[key]
+    end
+
+end
+
+local function set_miss(self, key, value)
+    local class = rawget(self, "_class")
+
+    if rawget(self, "set_"..key) then
+        return rawget(self, "set_"..key)(self, value)
+    elseif class and class["set_"..key] then
+        return class["set_"..key](self, value)
+    elseif rawget(self, "_enable_auto_signals") then
+        local changed = class[key] ~= value
+        class[key] = value
+
+        if changed then
+            self:emit_signal("property::"..key, value)
+        end
+    else
+        return rawset(self, key, value)
+    end
+end
+
+--- Returns a new object. You can call `:emit_signal()`, `:disconnect_signal()`,
+-- `:connect_signal()` and `:add_signal()` on the resulting object.
+--
+-- Note that `args.enable_auto_signals` is only supported when
+-- `args.enable_properties` is true.
+--
+-- @tparam[opt={}] table args The arguments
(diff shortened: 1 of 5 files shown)


lib/gears/object.lua
```diff
@@ -122,9 +122,47 @@ function object:emit_signal(name, ...)
...
+    local class = rawget(self, "_class")
+
+    if rawget(self, "get_"..key) then
+        return rawget(self, "get_"..key)(self)
+    elseif class and class["get_"..key] then
+        return class["get_"..key](self)
+    elseif class and class[key] then
+        return class[key]
+    else
+        return rawget(self, key)
```
psychon: If we are in an `__index`-handler, this `rawget` will always return `nil`. Otherwise, the handler wouldn't be called in the first place. You can just remove this.

lib/gears/object.lua
```diff
@@ -122,9 +122,47 @@ function object:emit_signal(name, ...)
...
+local function set_miss(self, key, value)
+    local class = rawget(self, "_class")
+
+    if rawget(self, "set_"..key) then
+        return rawget(self, "set_"..key)(self, value)
+    elseif class and class["set_"..key] then
+        return class["set_"..key](self, value)
+    elseif rawget(self, "_enable_auto_signals") then
+        self:emit_signal("property::"..key, value)
+        class[key] = value
```
psychon: You should set the new value before emitting the signal, no? (And, if you want, emit the signal with the old value as argument). You always cite Qt for this. :-)

Apropos Qt: Only do this if `new_value ~= old_value`.

psychon: Oh and: Why the heck does this set the property on the class?

lib/gears/object.lua
```diff
@@ -136,7 +174,27 @@ local function new()
...
+    if args.enable_properties then
+        -- Check got existing get_xxxx and set_xxxx
+        mt.__index    = get_miss
+        mt.__newindex = set_miss
+    elseif args.class then
+        -- Use the class table a miss handler
+        mt.__index = ret._class
+    end
+
+    return setmetatable(ret, mt)
```
psychon: Could you add a test to `spec/` testing the various cases?

lib/gears/object.lua
```diff
@@ -21,10 +21,10 @@ local function check(obj)
 end
 
 --- Find a given signal
```
psychon: The commit message lies a bit. `@param` works just fine in ldoc as well. How about "Improve documentation"?

psychon:
Sigh. Since the can of worms is already open: Fine, let's implement our own Lua-OOP-mechanism. Mike Pall would complain. ;-)

lib/gears/object.lua
Elv13: > Oh and: Why the heck does this set the property on the class?

Because it would stop working after the first change (__newindex would no longer be called). Below, the code make sure the `class` variable is wrapper in an additional table. I could change the variable name to `proxy` if you think this would make it more obvious.

Elv13:
All done

lib/gears/object.lua
psychon: Either that or a comment explaining this (in some cases `class` really is the class, in others it isn't...)

psychon:
Thanks! I didn't look too closely, but instead will just trust you on this. (Feel free to ignore my confusion about "this sets a property on the class?!" if you think the code is fine like this.)

PR #915, opened 2016-05-18
layoutbox: Revert 25f4f24791

### LINKED ISSUE #914: Wibar indicator for window layout does not change
Hi,

since 32eeaa9513 the indicator, which shows the window layout the tag is currently using, does not change if I change the layout. E.g. it indicates, that the tag has floating layout, when in reality it has tiling layout.

Cheers.

Elv13:
Apparently, there is such thing as not leaking enough...

Fixes #914, unfixes #808

### DIFF
--- a/lib/awful/widget/layoutbox.lua
+++ b/lib/awful/widget/layoutbox.lua
@@ -56,10 +56,10 @@ function layoutbox.new(screen)
     local w = boxes[screen]
     if not w then
         w = imagebox()
-        w._layoutbox_tooltip = tooltip({ objects = {w}, delay_show = 1 })
+        w._layoutbox_tooltip = tooltip {objects = {w}, delay_show = 1}
 
         update(w, screen)
-        boxes[screen] = setmetatable({}, {__index = w, __newindex = w})
+        boxes[screen] = w
     end
 
     return w
--- a/tests/test-leaks.lua
+++ b/tests/test-leaks.lua
@@ -14,6 +14,7 @@ end
 -- Make the layoutbox in the default config GC'able
 -- luacheck: globals mywibox mylayoutbox
 for s in screen do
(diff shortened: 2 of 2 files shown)


psychon:
Seems like a random experiment, but ok.
Seems like just a partly revert, partly new changes, but whatever.

:+1:

psychon:
(Oh and: I have no idea why that commit broke the layoutbox...)

Elv13:
> Seems like a random experiment, but ok.

I ran the test 100 time and none failed. It's still in the "lucky" territory, so maybe there is no causality, but it would make at least some sense. What if the `layoutbox` is actually correctly GCable, but the `wibar` is the one actually failing to be GCed. In that case, we looked at the wrong thing for months. If this experiment turn out to get rid of the problem, then we will have a clue.

psychon:
Might be. I ran the leaks test 100 times with the following patch and sadly it never failed (if it had failed, it would be interesting if it failed for the layoutbox of the "main wibox").
```diff
diff --git a/tests/test-leaks.lua b/tests/test-leaks.lua
index 247524f..26fa689 100644
--- a/tests/test-leaks.lua
+++ b/tests/test-leaks.lua
@@ -16,7 +16,7 @@ end
 for s in screen do
     mywibox[s].visible = false
 end
-mywibox = nil
+--mywibox = nil
 mylayoutbox = nil
 emit_refresh()
 
@@ -53,6 +53,14 @@ local function tooltip_now()
     return w, t
 end
 
+local function get_foo()
+    local result = mywibox
+    mywibox = nil
+    return result
+end
+
+collectable(get_foo())
+
 -- First test some basic widgets
 collectable(wibox.widget.base.make_widget())
 collectable(wibox.widget.textbox("foo"))
```

Issue #916, opened 2016-05-18
Wibar ignores some arguments for wibox construction


actionless:
for example, `visible` is hardcoded to `true` so you have to set it after creating a widget

and the rest of args is just skipped:
```
border_width
border_color
ontop 	
cursor 	
opacity 
x 	
y 	
drawable 	
widget 	
window 	
shape_bounding 
shape_clip     
bg 	
bgimage 
fg 	
```

mb in `wibar.new()` we can iterate through args and assign them all to a newly created wibar?

psychon:
Looking at the code I thought "this has to work". Then I thought: Could someone please try changing line 344 of lua/awful/wibar.lua from `function awfulwibar.mt:__call(...)` into `function awfulwibar.mt:__call(_, ...)` and check if this fixes this issue?

actionless:
sorry, my bad, the problem was only with visible, so `if arg.visible == nil then w.visible = true end` should address the problem

funeralismatic:
bgimage not work.

Elv13:
@funeralismatic That one isn't a regression, it just wasn't a feature. Anyway, I fixed that.

actionless:
sorry, closed by mistake

PR #918, opened 2016-05-18
fix(awful: wibar): visible arg was ignored

actionless:
partly addresses #916


Elv13:
Thanks for the patch, sorry about the regression. I will add a regression test for it.

psychon:
(AFAIK this isn't actually a regression. `visible=true` was also forced before.)

PR #919, opened 2016-05-19
wibox: Add more constructor arguments

### LINKED ISSUE #916: Wibar ignores some arguments for wibox construction

Elv13:
Fixes #916

### DIFF
--- a/lib/awful/wibar.lua
+++ b/lib/awful/wibar.lua
@@ -261,13 +261,34 @@ end
 -- @see stretch
 
 --- Create a new wibox and attach it to a screen edge.
--- @see wibox
--- @param arg A table with standard arguments to wibox() creator.
 -- You can add also position key with value top, bottom, left or right.
 -- You can also use width or height in % and set align to center, right or left.
 -- You can also set the screen key with a screen number to attach the wibox.
--- If not specified, 1 is assumed.
--- @return The wibox created.
+-- If not specified, the primary screen is assumed.
+-- @see wibox
+-- @tparam[opt=nil] table arg
+-- @tparam string arg.position The position.
+-- @tparam string arg.stretch If the wibar need to be stretched to fill the screen.
+-- @tparam integer arg.border_width Border width.
+-- @tparam string arg.border_color Border color.
+-- @tparam boolean arg.ontop On top of other windows.
+-- @tparam string arg.cursor The mouse cursor.
+-- @tparam boolean arg.visible Visibility.
+-- @tparam number arg.opacity The opacity of the wibox, between 0 and 1.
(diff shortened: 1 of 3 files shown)


tests/run.sh
```diff
@@ -17,10 +17,14 @@ if [ "$CI" = true ]; then
...
-source_dir="$PWD"
+source_dir="$(dirname "$(realpath "$(type -p $0)")")"/..
 build_dir="$CMAKE_BINARY_DIR"
 if [ -z "$build_dir" ]; then
-    build_dir="$PWD/build"
+    if [ -d "$source_dir/build" ]; then
+        build_dir="$source_dir/build"
+    else
+        build_dir="$PWD"
+    fi
```
psychon: Did this commit end up in this PR on purpose? My comments from the last time that I saw this patch still apply (e.g. bashism that doesn't work in zsh; I don't see how this worked before my patch, ...)

Elv13: No, it wasn't supposed to be there. I still need this to run the tests using my setup (and the sleep 0.5), I accidentally pushed it again.

Elv13:
Ok, this should be it

psychon:
:+1:

actionless:
so now i can assign `.align` property instead of calling placement function explicitly, right?

lib/awful/wibar.lua
```diff
@@ -205,11 +205,14 @@ function awfulwibar.set_position(wb, position, screen) --luacheck: no unused arg
 end
 
 --- Attach a wibox to a screen.
--- If a wibox is attached, it will be automatically be moved when other wiboxes
--- will be attached.
+--
+-- This function has been moved to the `awful.placement` module. Calling this
+-- no longer do anything.
```
blueyed: s/do/does/

Elv13:
> so now i can assign .align property instead of calling placement function explicitly, right?

The `placement` API is the recommended way to place things (wibox/client/tooltip/menu/mouse/...). `awful.wibar.align` is still deprecated. Those placement function are shared across many modules. The goal of this API is to pack together all geometric placing/resizing function so they can be composed into more complex ones and share helper code to support corner cases like screen resize and workarea changes that would have been otherwise implemented in a "zillion" places.

PR #921, opened 2016-05-20
Add pause() and continue() methods to clients

Elv13:
This was asked many time, there is a few implementations floating around using spawn, but this implement is directly in Awesome. The main advantage is to be able to use this in `awful.rules` or create some kind of modules to pause hidden tags clients.

This cannot be added to the default config as some clients have multiple client per `pid`, so there is no generic/magic rules for this.

Edit: the pause and continue methods have been tested manually, so are the usage examples.

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -1205,6 +1205,22 @@ function client.object.is_transient_for(self, c2)
     return nil
 end
 
+--- Pause a client.
+-- This force the client to stop using power. This is the equivalent of `CTRL+Z`
+-- in a terminal.
+-- @function client.pause
+-- @see client.continue
+function client.object.pause(self)
+    self:kill(client.posix_signal.SIGSTOP)
+end
+
+--- Restore a paused client.
+-- @function client.continue
+-- @see client.pause
+function client.object.continue(self)
+    self:kill(client.posix_signal.SIGCONT)
+end
+
 -- Register standards signals
 
 --- The last geometry when client was floating.
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -1208,6 +1208,14 @@ end
 --- Pause a client.
 -- This force the client to stop using power. This is the equivalent of `CTRL+Z`
 -- in a terminal.
+--
+-- To add this to your `rc.lua`, add to the `clientkeys` section:
+--
+--    awful.key({ modkey,           }, "z", function (c) c:pause() end,
+--              {description = "pause", group = "client"}),
+--    awful.key({ modkey, "Shift"   }, "z", function (c) c:continue() end,
+--              {description = "continue", group = "client"}),
+--
 -- @function client.pause
 -- @see client.continue
 function client.object.pause(self)
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -59,6 +59,56 @@ client.property = {}
 client.shape = require("awful.client.shape")
 client.focus = require("awful.client.focus")
 
+--- The POSIX signal table.
+--
+-- This table was extracted from
+--
+--    man 7 signal
+--
+-- @table awful.client.posix_signal
+-- @field SIGHUP Hangup detected on controlling terminal or death of controlling process
+-- @field SIGINT Interrupt from keyboard
+-- @field SIGQUIT Quit from keyboard
+-- @field SIGILL Illegal Instruction
+-- @field SIGABRT Abort signal from abort(3)
+-- @field SIGKILL Kill signal
+-- @field SIGSEGV Invalid memory reference
+-- @field SIGALRM Timer signal from alarm(2)
+-- @field SIGTERM Termination signal
+-- @field SIGUSR1 User-defined signal 1
+-- @field SIGUSR2 User-defined signal 2
+-- @field SIGCONT Continue if stopped
+-- @field SIGSTOP Stop process
+-- @field SIGTSTP Stop typed at terminal
+-- @field SIGURG Urgent condition on socket (4.2BSD)
+-- @field SIGXCPU CPU time limit exceeded (4.2BSD)
+-- @field SIGXFSZ File size limit exceeded (4.2BSD)
+-- @field SIGIO I/O now possible (4.2BSD)
+-- @field SIGWINCH Window resize signal (4.3BSD, Sun)
+-- @see kill
+client.posix_signal = {
+    SIGHUP   = 1,
+    SIGINT   = 2,
+    SIGQUIT  = 3,
+    SIGILL   = 4,
+    SIGABRT  = 6,
+    SIGKILL  = 9,
+    SIGSEGV  = 11,
+    SIGALRM  = 14,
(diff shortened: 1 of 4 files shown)


lib/awful/client.lua
```diff
@@ -1155,6 +1205,30 @@ function client.object.is_transient_for(self, c2)
     return nil
 end
 
+--- Pause a client.
+-- This force the client to stop using power. This is the equivalent of `CTRL+Z`
+-- in a terminal.
+--
+-- To add this to your `rc.lua`, add to the `clientkeys` section:
```
blueyed: s/add to the/add the following to the/

lib/awful/client.lua
```diff
@@ -1155,6 +1205,30 @@ function client.object.is_transient_for(self, c2)
...
+--              {description = "continue", group = "client"}),
+--
+-- @function client.pause
+-- @see client.continue
+function client.object.pause(self)
+    self:kill(client.posix_signal.SIGSTOP)
+end
+
+--- Restore a paused client.
+-- @function client.continue
```
blueyed: Mention `SIGCONT` in the description, please.

objects/client.c
```diff
@@ -1985,27 +1986,34 @@ client_unmanage(client_t *c, bool window_valid)
...
+            xcb_send_event(globalconf.connection, false, c->window,
+                        XCB_EVENT_MASK_NO_EVENT, (char *) &ev);
+        }
+        else
+            xcb_kill_client(globalconf.connection, c->window);
     }
     else
-        xcb_kill_client(globalconf.connection, c->window);
+    {
+        kill(c->pid, sig);
```
blueyed: Is `c->pid` always a given?

objects/client.c
```diff
@@ -2153,13 +2161,40 @@ client_set_icon_from_pixmaps(client_t *c, xcb_pixmap_t icon, xcb_pixmap_t mask)
 
 /** Kill a client.
  *
+ * By default, this will use the normal X11 way of closing a client (the `X`
```
blueyed: better:
> the "X" / close button

objects/client.c
```diff
@@ -2153,13 +2161,40 @@ client_set_icon_from_pixmaps(client_t *c, xcb_pixmap_t icon, xcb_pixmap_t mask)
...
+ * button). If a signal number is provided, it will be used instead.
+ *
+ * This can be used to pause or wakeup some clients or to save energy.
```
blueyed: > This can be used to pause or wakeup some clients or to save energy.

Better:

> This can be used to pause or wakeup some clients, which can be used to save energy.

objects/client.c
```diff
@@ -2153,13 +2161,40 @@ client_set_icon_from_pixmaps(client_t *c, xcb_pixmap_t icon, xcb_pixmap_t mask)
...
     client_t *c = luaA_checkudata(L, 1, &client_class);
-    client_kill(c);
+    int sig = -1;
+
+    /* Allow a signal number */
+    if(lua_gettop(L) == 2)
+    {
+        luaA_checkinteger(L, 2);
+
+        sig = lua_tointeger(L, 2);
```
blueyed: No need for a newline.

lib/awful/client.lua
```diff
@@ -1155,6 +1205,30 @@ function client.object.is_transient_for(self, c2)
     return nil
 end
 
+--- Pause a client.
+-- This force the client to stop using power. This is the equivalent of `CTRL+Z`
```
blueyed: s/force/forces/

Also mention here that it will send `SIGSTOP`.

blueyed:
Awesome!

barlik:
I've been using this snippet, works without a problem. It can also be called from `awful.rules` with a callback. 
```
function process_continue(c)
    awful.util.spawn(string.format("kill -CONT %d", c.pid))
end
function process_stop(c)
    awful.util.spawn(string.format("kill -STOP %d", c.pid))
end
```

Replacing it with just `c:pause()` and `c:continue()` sounds good.

objects/client.c
```diff
@@ -2153,13 +2161,29 @@ client_set_icon_from_pixmaps(client_t *c, xcb_pixmap_t icon, xcb_pixmap_t mask)
...
  * @function kill
  */
 static int
 luaA_client_kill(lua_State *L)
 {
     client_t *c = luaA_checkudata(L, 1, &client_class);
-    client_kill(c);
+    int sig = -1;
+
+    /* Allow a signal number */
```
psychon: No. Just no!

Feel free to deprecate `c:kill()`, rename it to `c:close()` and add `c:send_signal(number)` (or `c:signal(number)`?), but having the same function do two completely different things seems like a really, really bad idea. I guess that the only reason why you want to do it like this is because it's called `kill`.

objects/client.c
psychon: Nope. `c->pid` will be 0 for clients which don't have `_NET_WM_PID` (I just tested with `xcalc`).
`man 2 kill`:
> If pid equals 0, then sig is sent to every process in the process group of the calling process.

Yay, awesome will stop itself if you run xcalc and have the auto-SIGSTOP-"feature" in your config.

Also, just checking `_NET_WM_PID` is not enough since X11 is soooo great at being a remote protocol. Could you please also make this check `WM_CLIENT_MACHINE` and only try to send signals to random processes if the client claims to be coming from the local machine?
https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#KILLINGWINDOWS

lib/awful/client.lua
```diff
@@ -59,6 +59,56 @@ client.property = {}
 client.shape = require("awful.client.shape")
 client.focus = require("awful.client.focus")
 
+--- The POSIX signal table.
+--
+-- This table was extracted from
+--
+--    man 7 signal
```
psychon: Sorry, but this is non-portable. POSIX doesn't define the numbers for the signals, this is Linux-specific.

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html

lib/awful/client.lua
```diff
@@ -59,6 +59,56 @@ client.property = {}
...
+    SIGILL   = 4,
+    SIGABRT  = 6,
+    SIGKILL  = 9,
+    SIGSEGV  = 11,
+    SIGALRM  = 14,
+    SIGTERM  = 15,
+    SIGUSR1  = 10,
+    SIGUSR2  = 12,
+    SIGCONT  = 18,
+    SIGSTOP  = 19,
```
psychon: Also, this one isn't even correct on all Linux versions. Some use 19 for SIGSTOP, others have 17 or 23. `man 7 signals` explains this.
You'll have to do something build-time to get the signal numbers, if you really want to export these to Lua.

objects/client.c
psychon: Better idea: Export a wrapper around `kill` and another around `gethostname` to Lua, then this functionality can be implemented in Lua and someone will eventually find other uses for this.

PR #923, opened 2016-05-23
Widget api refactor

Elv13:
As discussed in IRC and some Github issues, this PR will:

 * Create a container module [DONE]
 * Move some awful.widgets into wibox if they don't require other awful code (but `util`) [DONE]
 * Fix all constructor documentation [DONE]
 * Add documentation [DONE]
 * Remove methods used internally from the "user visible" API (aka, ldoc) [DONE]
 * Enable the property system (and update documentation) [DONE]
 * Add usage examples [WIP]
 * Add images [DONE]
 * Document any `beautiful` variables used by the widgets and add new where relevant. [DONE]
 * Add header description. [WIP]
 * <s>Fix coverall input so the coverage is actually correctly computed, otherwise it will take a serous dive (while actually being up)</s>

**Questions:**

 * Should I move the `systray` widget into `awful.widget`?


psychon:
> Should I move the systray widget into awful.widget?

I guess the reason being "wibox contains 'lightweight' widgets and awful the 'heavy'  ones?". Fine with me. (And getting rid of awful.widget doesn't work; tasklist and taglist are too...something for that)

Elv13:
The new container documentation and layout table of contents
widgetlayout6


widgetlayout6

Elv13:
Ok, this is it. It was also boring, but not as bad as when I did that for the capi objects.

The first few commits move modules around, then I add some tests/screenshot to catch regressions, then I enable the object properties and add a lot of doc. The object property part is a little  bit scaryer, as it forces me to also change the internal API. I used these regexes to find places to change:

```sh
([a-z_]+)\.([^_ .:]|[^p .:])[^p\.:\ =+-<>~*/]+[.: =+-<>~*/]

self\.[^_]+
self\._[^p]+

```

They are not really accurate, so I may have missed some, but so far it look to work. I know it wont be very fun to review, sorry about that. Maybe most commits can be skipped and only the "real changes" are worth taking a look. There is a lot of the same in there commit, done 16 time (for each widgets).

I will move the systray and awful.buttons in a future PR, this is way to large already...

Elv13:
I added a 'rawset' option to `awful.util.table.crush`. After doing a little experiment by logging all miss from the widgets, I found that doing that and adding some other `rawset()` reduced the miss by 93%. I don't like the look of `rawset`s, but for larger config it would eventually become noticeable. I also fixed a typo (provite->private) introduced by one of the commit.

lib/wibox/layout/constraint.lua
```diff
@@ -1,140 +1,18 @@
...
 -- @release @AWESOME_VERSION@
 -- @classmod wibox.layout.constraint
 ---------------------------------------------------------------------------
 
-local pairs = pairs
-local type = type
-local setmetatable = setmetatable
-local base = require("wibox.widget.base")
-local math = math
+local util = require("awful.util")
```
psychon: Sigh. I hate dependencies from wibox into awful, but I guess I'll have to live with it.

tests/examples/wibox/container/defaults/template.lua
```diff
@@ -0,0 +1,96 @@
...
+for _ = 1, 10 do
+    awesome:emit_signal("refresh")
+end
+
+-- Get the example fallback size (the tests can return a size if the want)
+local f_w, f_h = container:fit({dpi=96}, 9999, 9999)
+
+-- There is an overhead that cause testbox "...", add 10 until someone
+-- figures out the real equation
+f_w, f_h = f_w+10, f_h+10
```
psychon: Uhm, all textboxes are inside of `constraint` containers, so their `fit` result is ignored. I guess a single space is just wider than how much space you give it here? My suggested fix would be to use the empty string instead of a single space. Even better would be not to use a textbox at all since you really only want a widget doing nothing. I thought the declarative layout allows to construct widgets on the fly? All callbacks should be optional by now, so could that be used instead of a textbox here?

btw: spelling: testbox -> textbox

tests/examples/wibox/container/defaults/template.lua
```diff
@@ -0,0 +1,96 @@
...
+
+-- Save to the output file
+local img = surface["widget_to_svg"](container, image_path..".svg", f_w, f_h)
```
psychon: Any reason not to use `surface.widget_to_svg`?

lib/wibox/container/background.lua
```diff
@@ -167,7 +167,7 @@ function background:get_fg()
     return self._private.foreground
 end
 
---- The background shape.
+--- The background shap     e.
```
psychon: Please fix this

lib/wibox/container/background.lua
```diff
@@ -22,7 +22,7 @@ local background = { mt = {} }
 
 -- Draw this widget
 function background:draw(context, cr, width, height)
-    if not self._private.widget or not self._private.widget.visible then
+    if not self._private.widget or not self._private.widget._private.visible then
```
psychon: Shouldn't this use `:get_visible()` instead? Looking into something called `_private` on another widget seems wrong. (Same comment for other places)

docs/common/widget.ldoc
```diff
@@ -39,3 +39,64 @@
 --- Set/get a widget's buttons.
 -- @param _buttons The table of buttons that should bind to the widget.
 -- @function buttons
+
+
+--- When the layout (size) change.
+-- @signal widget::layout_changed
```
psychon: Could you add a comment that unless this signal is emitted, `:layout()` and `:fit()` must return the same result when called with the same arguments?
(And layout is more than just size, because it also involved `:layout()`. However, I don't have a good suggestion for the wording.)

docs/common/widget.ldoc
```diff
@@ -39,3 +39,64 @@
...
+
+--- When the widget content changed.
+-- @signal widget::redraw_needed
```
psychon: Here I'd say something about the result of `:draw()`, similar to the previous point...
The difference between the two signals should become clear.

docs/common/widget.ldoc
```diff
@@ -39,3 +39,64 @@
...
+
+--- When a mouse button is pressed over the widget.
+-- @signal button::press
+-- @tparam table widget The widget
+-- @tparam number lx The horizontal position.
+-- @tparam number ly The vertical position.
```
psychon: The "l" means "local", I think. This is the position of the mouse press relative to the widget while `geometry` contains the geometry of the widget relative to the wibox (as long as you don't rotate by non-multiple of 90°....).
(Same for button::release)

docs/common/widget.ldoc
```diff
@@ -39,3 +39,64 @@
...
+-- @signal mouse::leave
+-- @tparam table widget The widget
+-- @tparam table mods The modifiers (mod4, mod1 (alt), Control, Shift)
+-- @tparam table geometry
+-- @tparam number geometry.x The vertical position
+-- @tparam number geometry.y The horizontal position
+-- @tparam number geometry.width The widget
+-- @tparam number geometry.height The height
+-- @tparam drawable geometry.drawable The `drawable`
+-- @see mouse
```
psychon: Sorry, but we/I am inconsistent. :-)

enter/leave only have a single argument which is the result of `:find_widget` which is what you call `geometry`here. There are no modifiers (and I don't think there should be...)

docs/common/object.ldoc
```diff
@@ -0,0 +1,31 @@
+
+--- Connect to a signal weakly. This allows the callback function to be garbage
+-- collected and automatically disconnects the signal when that happens.
```
psychon: Add something like: Only use this function if you really, really, *really* know what you are doing.

docs/common/object.ldoc
```diff
@@ -0,0 +1,31 @@
...
+--- Add a signal to an object. All signals must be added before they can be used.
+--
+-- @tparam string name The name of the new signal.
+-- @function add_signal
+
+--- Connect to a signal.
+-- @tparam string name The name of the signal
+-- @tparam function func The callback to call when the signal is emitted
+-- @see add_signal 
+-- @function connect_signal
```
psychon: Why is this so far away from its weak version? I'd keep them closer together in this file.

lib/wibox/container/mirror.lua
```diff
@@ -71,30 +77,32 @@ end
...
-            self[ref] = reflection[ref]
+            self._private[ref] = reflection[ref]
         end
     end
     self:emit_signal("widget::layout_changed")
 end
 
 --- Get the reflection of this mirror layout.
---  @return a table of booleans with the keys "horizontal", "vertical".
+-- @property reflexion
```
psychon: reflexion vs reflection. The docs use one word, the code another

lib/wibox/widget/textbox.lua
```diff
@@ -25,37 +24,37 @@ local textbox = { mt = {} }
 
 --- Set the DPI of a Pango layout
 local function setup_dpi(box, dpi)
-    if box.dpi ~= dpi then
-        box.dpi = dpi
-        box._ctx:set_resolution(dpi)
-        box._layout:context_changed()
+    if box._private.dpi ~= dpi then
```
psychon: ....here

lib/wibox/widget/textbox.lua
```diff
@@ -239,17 +262,13 @@ end
...
-        if type(v) == "function" then
-            ret[k] = v
-        end
-    end
+    util.table.crush(ret, textbox, true)
 
-    ret._dpi = -1
-    ret._ctx = PangoCairo.font_map_get_default():create_context()
-    ret._layout = Pango.Layout.new(ret._ctx)
+    ret._dpi = -1 --FIXME this is unused, but .dpi is used but not defined
```
psychon: This is supposed to be used above (see the other comment), but apparently I typo'd. Feel free to fix that.

lib/awful/titlebar.lua
```diff
@@ -204,8 +204,8 @@ function titlebar.widget.button(c, name, selector, action)
     local ret = imagebox()
 
     if titlebar.enable_tooltip then
-        ret.tooltip = atooltip({ objects = {ret}, delay_show = 1 })
-        ret.tooltip:set_text(name)
+        ret._private.tooltip = atooltip({ objects = {ret}, delay_show = 1 })
+        ret._private.tooltip:set_text(name)
```
psychon: According to the commit message, this shouldn't be in this commit.

spec/wibox/test_utils.lua
```diff
@@ -93,7 +93,7 @@ return {
                 return width or 10, height or 10
             end
         end
-        w._widget_caches = {}
+        w._private.widget_caches = {}
```
psychon: Should this really be in this commit?

psychon:
I'll skip trying to look at the generated docs. I have a feeling like you did enough of that already. The stuff in `docs/widget_lists.cmake` seems complicated enough so that I'll just trust you on it.

For future stuff: Could we do such things in e.g. Lua instead of CMake? And no, I don't see how exactly this should be done either, but CMake seems to be wrong to me.

Some of the commits that move e.g. `.visible` into `.private.visible` are half-wrong, because the widget shim used by the unit tests is changed before the actual code is changed. There is multiple such things. I'll let that slip for this time and only left comments for the "worst offenders".

Feel free to merge this after looking through my comments (and hopefully fixing most of them).

lib/wibox/layout/constraint.lua
Elv13: Then move `util.table` into gears.table and `util.deprecate` into `gears.deprecate`. Those are the 2 being used everywhere.

PR #926, opened 2016-05-27
Signals and hostname

psychon:
Follow up to #921. This should be more general (and I'll just ignore that this partly reimplements stuff that is already available in luaposix).
@Elv13


blueyed:
@Elv13 any objections/comments?

blueyed:
There we go.. :)

PR #929, opened 2016-05-29
Geometry overhaul part 3.03 (final part for 3.6)

Elv13:
* Add the last planned placement function [DONE]
* Turn tooltip into a class [DONE]
* Add shape support to tooltips [DONE]
* Support "next to drawable" tooltip mode like Max OS X [DONE]
* Add more tooltip properties and theme options [DONE]
* <s>Remove awful.menu geometry code and use awful.placement</s> later
* <s>Add hot corner support</s> later
* <s>Find a way to get the "smart wibox" (wibox that resize itself to fit a widget and update its own shape) from radical into awful or wibox (not sure yet)</s> later
* Test all this [DONE]

What wont make it (from the original TODO list of this PR serie):

 * Notification shape/placement (this require the dynamic layout system and I wont push it into 3.6)

Closing note:

This is the last PR of the geometry overhaul. After this PR, my Radical module will be 3K line shorted than the 3.5 version. My collision module will be much smaller too (partly because of the shape API too). I think so far it went well. Beside the placement code and the doc, awful also got smaller, but not by much. The first 4 PR were quite big and intrusive, but there was only a few regression reports so far. `awful.placement` may be a huge pile of complex and coupled/spaghetti code, but (I think/hope) it made everything else simpler and less coupled. Plus, this serie gave many cool new features I hope users will love. I hope this code will be the basis of tons of Awesomeness.

### DIFF
--- a/lib/awful/mouse/init.lua
+++ b/lib/awful/mouse/init.lua
@@ -284,7 +284,6 @@ end
 -- @treturn table The list of widgets.The first element is the biggest
 -- container while the last is the topmost widget. The table contains *x*, *y*,
 -- *width*, *height* and *widget*.
--- @treturn table The list of geometries.
 -- @see wibox.find_widgets
 
 function mouse.object.get_current_widgets()
@@ -308,8 +307,8 @@ end
 -- @property current_widget
 -- @tparam widget|nil widget The widget
 -- @treturn ?widget The widget
--- @treturn ?table The geometry.
 -- @see wibox.find_widgets
+-- @see current_widget_geometry
 
 function mouse.object.get_current_widget()
     local wdgs, geos = mouse.object.get_current_widgets()
@@ -317,6 +316,28 @@ function mouse.object.get_current_widget()
     if wdgs then
         return wdgs[#wdgs], geos[#geos]
     end
+end
+
+--- Get the current widget geometry.
+-- @property current_widget_geometry
+-- @tparam ?table The geometry.
+-- @see current_widget
+
+function mouse.object.get_current_widget_geometry()
+    local _, ret = mouse.object.get_current_widget()
+
+    return ret
+end
+
+--- Get the current widget geometries.
+-- @property current_widget_geometries
+-- @tparam ?table A list of geometry tables.
+-- @see current_widgets
+
+function mouse.object.get_current_widget_geometries()
+    local _, ret = mouse.object.get_current_widgets()
+
+    return ret
 end
 
 --- True if the left mouse button is pressed.
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -96,6 +96,8 @@ local a_screen = require("awful.screen")
 local grect = require("gears.geometry").rectangle
 local util = require("awful.util")
 local dpi = require("beautiful").xresources.apply_dpi
+local cairo = require( "lgi" ).cairo
+local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
 
 local function get_screen(s)
     return s and capi.screen[s]
@@ -258,6 +260,19 @@ local resize_to_point_map = {
     right        = {p1={0,0} , p2= nil , x_only=true , y_only=false, align="top_left"    },
     top          = {p1= nil  , p2={1,1}, x_only=false, y_only=true , align="bottom_left" },
     bottom       = {p1={0,0} , p2= nil , x_only=false, y_only=true , align="top_left"    },
+}
+
+-- Outer position matrix
+-- 1=best case, 2=fallback
+local outer_positions = {
+    left1   = function(r, w, _) return {x=r.x-w        , y=r.y            }, "down"  end,
+    left2   = function(r, w, h) return {x=r.x-w        , y=r.y-h+r.height }, "up"    end,
+    right1  = function(r, _, _) return {x=r.x          , y=r.y            }, "down"  end,
+    right2  = function(r, _, h) return {x=r.x          , y=r.y-h+r.height }, "up"    end,
+    top1    = function(r, _, h) return {x=r.x          , y=r.y-h          }, "right" end,
+    top2    = function(r, w, h) return {x=r.x-w+r.width, y=r.y-h          }, "left"  end,
+    bottom1 = function(r, _, _) return {x=r.x          , y=r.y            }, "right" end,
+    bottom2 = function(r, w, _) return {x=r.x-w+r.width, y=r.y            }, "left"  end,
 }
 
 --- Add a context to the arguments.
@@ -553,8 +568,9 @@ attach = function(d, position_f, args)
         end
     end
 
-    -- If there is a parent drawable, screen or mouse, also track it
-    if parent then
+    -- If there is a parent drawable, screen, also track it.
+    -- Note that tracking the mouse is not supported
+    if parent and parent.connect_signal then
         parent:connect_signal("property::geometry" , tracker)
     end
 
@@ -601,6 +617,158 @@ local function rect_to_point(rect, corner_i, corner_j)
         x = rect.x + corner_i * math.floor(rect.width ),
         y = rect.y + corner_j * math.floor(rect.height),
     }
+end
+
+-- Create a pair of rectangles used to set the relative areas.
+-- v=vertical, h=horizontal
+local function get_cross_sections(abs_geo, mode)
+    if not mode or mode == "cursor" then
+        -- A 1px cross section centered around the mouse position
+        local coords = capi.mouse.coords()
+        return {
+            h = {
+                x      = abs_geo.drawable_geo.x     ,
+                y      = coords.y                   ,
+                width  = abs_geo.drawable_geo.width ,
+                height = 1                          ,
+            },
+            v = {
+                x      = coords.x                   ,
+                y      = abs_geo.drawable_geo.y     ,
+                width  = 1                          ,
+                height = abs_geo.drawable_geo.height,
+            }
+        }
+    elseif mode == "geometry" then
+        -- The widget geometry extended to reach the end of the drawable
+
+        return {
+            h = {
+                x      = abs_geo.drawable_geo.x     ,
+                y      = abs_geo.y                  ,
+                width  = abs_geo.drawable_geo.width ,
+                height = abs_geo.height             ,
+            },
+            v = {
+                x      = abs_geo.x                  ,
+                y      = abs_geo.drawable_geo.y     ,
+                width  = abs_geo.width              ,
+                height = abs_geo.drawable_geo.height,
+            }
+        }
+    elseif mode == "cursor_inside" then
+        -- A 1x1 rectangle  centered around the mouse position
+
+        local coords = capi.mouse.coords()
+        coords.width,coords.height = 1,1
+        return {h=coords, v=coords}
+    elseif mode == "geometry_inside" then
+        -- The widget absolute geometry, unchanged
+
+        return {h=abs_geo, v=abs_geo}
+    end
+end
+
+-- When a rectangle is embedded into a bigger one, get the regions around
+-- the outline of the bigger rectangle closest to the smaller one (on each side)
+local function get_relative_regions(geo, mode, is_absolute)
+
+    -- Use the mouse position and the wibox/client under it
+    if not geo then
+        local draw   = capi.mouse.current_wibox
+        geo          = draw and draw:geometry() or capi.mouse.coords()
+        geo.drawable = draw
+    elseif is_absolute then
+        -- Some signals are a bit inconsistent in their arguments convention.
+        -- This little hack tries to mitigate the issue.
+
+        geo.drawable = geo -- is a wibox or client, geometry and object are one
+                           -- and the same.
+    elseif (not geo.drawable) and geo.x and geo.width then
+        local coords = capi.mouse.coords()
+
+        -- Check if the mouse is in the rect
+        if coords.x > geo.x and coords.x < geo.x+geo.width and
+          coords.y > geo.y and coords.y < geo.y+geo.height then
+            geo.drawable = capi.mouse.current_wibox
+        end
+
+        -- Maybe there is a client
+        if (not geo.drawable) and capi.mouse.current_client then
+            geo.drawable = capi.mouse.current_client
+        end
+    end
+
+    -- Get the drawable geometry
+    local dpos = geo.drawable and (
+        geo.drawable.drawable and
+            geo.drawable.drawable:geometry()
+            or geo.drawable:geometry()
+    ) or {x=0, y=0}
+
+    -- Compute the absolute widget geometry
+    local abs_widget_geo = is_absolute and geo or {
+        x            = dpos.x + geo.x              ,
+        y            = dpos.y + geo.y              ,
+        width        = geo.width                   ,
+        height       = geo.height                  ,
+        drawable     = geo.drawable                ,
+    }
+
+    abs_widget_geo.drawable_geo = geo.drawable and dpos or geo
+
+    -- Get the point for comparison.
+    local center_point = mode:match("cursor") and capi.mouse.coords() or {
+        x = abs_widget_geo.x + abs_widget_geo.width  / 2,
+        y = abs_widget_geo.y + abs_widget_geo.height / 2,
+    }
+
+    -- Get widget regions for both axis
+    local cs = get_cross_sections(abs_widget_geo, mode)
+
+    -- Get the 4 closest points from `center_point` around the wibox
+    local regions = {
+        left   = {x = cs.h.x           , y = cs.h.y            },
+        right  = {x = cs.h.x+cs.h.width, y = cs.h.y            },
+        top    = {x = cs.v.x           , y = cs.v.y            },
+        bottom = {x = cs.v.x           , y = cs.v.y+cs.v.height},
+    }
+
+    -- Assume the section is part of a single screen until someone complains.
+    -- It is much faster to compute and getting it wrong probably has no side
+    -- effects.
+    local s = geo.drawable and geo.drawable.screen or a_screen.getbycoord(
+                                                        center_point.x,
+                                                        center_point.y
+                                                      )
+
+    -- Compute the distance (dp) between the `center_point` and the sides.
+    -- This is only relevant for "cursor" and "cursor_inside" modes.
+    for _, v in pairs(regions) do
+        local dx, dy = v.x - center_point.x, v.y - center_point.y
+
+        v.distance = math.sqrt(dx*dx + dy*dy)
+        v.width    = cs.v.width
+        v.height   = cs.h.height
+        v.screen   = capi.screen[s]
+    end
+
+    return regions
+end
+
+-- Check if the proposed geometry fits the screen
+local function fit_in_bounding(obj, geo, args)
+    local sgeo   = get_parent_geometry(obj, args)
+    local region = cairo.Region.create_rectangle(cairo.RectangleInt(sgeo))
+
+    region:intersect(cairo.Region.create_rectangle(
+        cairo.RectangleInt(geo)
+    ))
+
+    local geo2 = region:get_rectangle(0)
+
+    -- If the geometry is the same then it fits, otherwise it will be cropped.
+    return geo2.width == geo.width and geo2.height == geo.height
 end
 
 --- Move a drawable to the closest corner of the parent geometry (such as the
@@ -1087,6 +1255,10 @@ for _, v in ipairs {"vertically", "horizontally"} do
     end
 end
 
+---@DOC_awful_placement_maximize_vertically_EXAMPLE@
+
+---@DOC_awful_placement_maximize_horizontally_EXAMPLE@
+
 --- Scale the drawable by either a relative or absolute percent.
 --
 -- Valid args:
@@ -1143,9 +1315,94 @@ function placement.scale(d, args)
     return fix_new_geometry(ngeo, args, true)
 end
 
----@DOC_awful_placement_maximize_vertically_EXAMPLE@
+--- Move a drawable to a relative position next to another one.
+--
+-- The `args.preferred_positions` look like this:
+--
+--    {"top", "right", "left", "bottom"}
+--
+-- In that case, if there is room on the top of the geomtry, then it will have
+-- priority, followed by all the others, in order.
+--
+-- @tparam drawable d A wibox or client
+-- @tparam table args
+-- @tparam string args.mode The mode
+-- @tparam string args.preferred_positions The preferred positions (in order)
+-- @tparam string args.geometry A geometry inside the other drawable
+-- @treturn table The new geometry
+-- @treturn string The choosen position
+-- @treturn string The choosen direction
+function placement.next_to(d, args)
+    args = add_context(args, "next_to")
+    d    = d or capi.client.focus
 
----@DOC_awful_placement_maximize_horizontally_EXAMPLE@
+    local preferred_positions = {}
+
+    if #(args.preferred_positions or {}) then
+        for k, v in ipairs(args.preferred_positions) do
+            preferred_positions[v] = k
+        end
+    end
+
+    local dgeo = geometry_common(d, args)
+    local pref_idx, pref_name = 99, nil
+    local mode,wgeo = args.mode
+
+    if args.geometry then
+        mode = "geometry"
+        wgeo = args.geometry
+    else
+        local pos = capi.mouse.current_widget_geometry
+
+        if pos then
+            wgeo, mode = pos, "cursor"
+        elseif capi.mouse.current_client then
+            wgeo, mode = capi.mouse.current_client:geometry(), "cursor"
+        end
+    end
+
+    if not wgeo then return end
+
+    -- See get_relative_regions comments
+    local is_absolute = wgeo.ontop ~= nil
+
+    local regions = get_relative_regions(wgeo, mode, is_absolute)
+
+    -- Check each possible slot around the drawable (8 total), see what fits
+    -- and order them by preferred_positions
+    local does_fit = {}
+    for k,v in pairs(regions) do
+        local geo, dir = outer_positions[k.."1"](v, dgeo.width, dgeo.height)
+        geo.width, geo.height = dgeo.width, dgeo.height
+        local fit = fit_in_bounding(v.screen, geo, args)
+
+        -- Try the other compatible geometry
+        if not fit then
+            geo, dir = outer_positions[k.."2"](v, dgeo.width, dgeo.height)
+            geo.width, geo.height = dgeo.width, dgeo.height
+            fit = fit_in_bounding(v.screen, geo, args)
+        end
+
+        does_fit[k] = fit and {geo, dir} or nil
+
+        if fit and preferred_positions[k] and preferred_positions[k] < pref_idx then
+            pref_idx  = preferred_positions[k]
+            pref_name = k
+        end
+
+        -- No need to continue
+        if fit and preferred_positions[k] == 1 then break end
+    end
+
+    local pos_name = pref_name or next(does_fit)
+    local ngeo, dir = unpack(does_fit[pos_name] or {}) --FIXME why does this happen
+
+    geometry_common(d, args, ngeo)
+
+    attach(d, placement.next_to, args)
+
+    return fix_new_geometry(ngeo, args, true), pos_name, dir
+end
 
 --- Restore the geometry.
 -- @tparam[opt=client.focus] drawable d A drawable (like `client` or `wibox`)
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -95,7 +95,6 @@ local layout = require("awful.layout")
 local a_screen = require("awful.screen")
 local grect = require("gears.geometry").rectangle
 local util = require("awful.util")
-local dpi = require("beautiful").xresources.apply_dpi
 local cairo = require( "lgi" ).cairo
 local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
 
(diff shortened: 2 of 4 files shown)


lib/awful/mouse/init.lua
```diff
@@ -326,6 +325,19 @@ end
 
 function mouse.object.set_current_widget() end
 
+--- Get the current widget geometry.
+-- @property current_widget_geometry
+-- @tparam ?table The geometry.
```
blueyed: s/tparam/treturn ?!

lib/awful/placement.lua
```diff
@@ -603,6 +617,150 @@ local function rect_to_point(rect, corner_i, corner_j)
     }
 end
 
+-- Create a pair of rectangles used to set the relative areas.
+-- v=vertical, h=horizontal
+local function get_cross_sections(abs_geo, mode)
+print("\n\nLA!",mode)
```
blueyed: Leftover print.

lib/awful/placement.lua
```diff
@@ -603,6 +617,150 @@ local function rect_to_point(rect, corner_i, corner_j)
...
+        local coords = capi.mouse.coords()
+        coords.width,coords.height = 1,1
+        return {h=coords, v=coords}
+    elseif mode == "geometry_inside" then
+        -- The widget absolute geometry, unchanged
+
+        return {h=abs_geo, v=abs_geo}
+    end
+
+    assert(false)
```
blueyed: Include the invalid mode here?

lib/awful/placement.lua
```diff
@@ -603,6 +617,150 @@ local function rect_to_point(rect, corner_i, corner_j)
...
+
+    -- Use the mouse position and the wibox/client under it
+    if not geo then
+        local draw   = capi.mouse.current_wibox
+        geo          = draw and draw:geometry() or capi.mouse.coords()
+        geo.drawable = draw
+    elseif (not geo.drawable) and geo.x and geo.width then
+        local coords = capi.mouse.coords()
+
+        -- Check id the mouse is in the rect
```
blueyed: s/id/if

lib/awful/mouse/init.lua
Elv13: properties documentation is using param. ldoc is configured to turn that into a "type" section. I could create an alias to make that less confusing...

lib/awful/placement.lua
Elv13: This is a local function, I will remove it once the the WIP todos are done. This is still work in progress Not all features from the prototype (and Radical.tooltip) have been imported yet.

Elv13:
I will stop this here. There is already too much features and breakages from me. Of course, you can't change the API as much as I did without breaking some eggs. I am sorry about that. I really believe the work I did to standardize and document the API will be beneficial for the project. If I had to do it again, I probably would not. It took _a lot_ of time, but addressed issues that have been bugging me for a long time. It also added some "cool" new features and flexibility. But this endeavor was way too large for a single release, let alone with 1200 other commits since 3.5.0.

So I am done for this release. This is it. I will fix whatever regressions I caused (that are not due to fragile hacks in modules/configs), but I wont submit new code until 3.7 is branched.

mindeunix:
@Elv13 Good work, as always.

Elv13:
> @Elv13 Good work, as always.

Thanks. This PR doesn't fully merge Radical tooltips and I didn't merge the other Blind PRs after all. I will turn them back into modules. At least, we (/me + @mindeunix) will have a better base to build our configs. I once this is merged, I will push a Radical update for the recent API changes. This will drop some files, but it's not quite enough to drop smart_wibox, hot_corner and tooltip Radical components as I hoped.

lib/awful/mouse/init.lua
```diff
@@ -326,6 +325,19 @@ end
...
+-- @see current_widget
+
+function mouse.object.get_current_widget_geometry()
+    local _, ret = mouse.object.get_current_widget()
+
+    return ret
+end
```
psychon: This handles `get_current_widget`. What about `get_current_widgets` (plural)?
Perhaps it is easier to make `get_current_widgets`  just return the result from `:find_widgets` instead of extracting the individual widgets (and perhaps renaming it to add `_geometries`)? I don't know.

lib/awful/mouse/init.lua
```diff
@@ -326,6 +325,19 @@ end
...
+
+function mouse.object.set_current_widget_geometry() end
```
psychon: Hm. Perhaps we should try getting rid of these "doing-nothing-setters" (and yes, this new idea doesn't belong into this PR).
What do you think about not allowing properties if there is a getter but not a setter in `gears.object.properties`?
Something like the following (untested):
```diff
diff --git a/lib/gears/object/properties.lua b/lib/gears/object/properties.lua
index 9b98868..a35239d 100644
--- a/lib/gears/object/properties.lua
+++ b/lib/gears/object/properties.lua
@@ -83,6 +83,12 @@ function object.capi_index_fallback(class, args)
             return args.setter_fallback(cobj, prop, value)
         end
 
+        -- If a getter exists but not a setter, then the property is read-only
+        if args.getterer_class and args.getter_class[getter_prefix..prop] then
+            -- error("Tried to set read-only-property (psychon always adds error()s, Elv13 doesn't like them")
+            return
+        end
+
         local fallback = properties[cobj] or cobj_register(cobj)
 
         -- Use the fallback property table
```

lib/awful/placement.lua
psychon: Still a leftover print a week later and when "WIP" is removed. I'll stop reading commits here and wait for you to specifically say "I handled all existing comments, please review again".

psychon: Uhm. Github GUI says that this is an outdated diff. However, I reviewed this by `git pull`ing your branch and looking at this locally and noticed the `print` there independently from Daniel. What's going on?!?

Elv13: The fixup was accidently rebased in the wrong commit. I reviewed the pr as
a whole, so i missed it. Github ui fail quite hard to make that obvious.
On Jun 3, 2016 3:37 AM, "Uli Schlachter" <notifications@github.com> wrote:

In lib/awful/placement.lua
<https://github.com/awesomeWM/awesome/pull/929#discussion_r65667165>:

> @@ -603,6 +617,150 @@ local function rect_to_point(rect, corner_i, corner_j)
>      }
>  end
>
> +-- Create a pair of rectangles used to set the relative areas.
> +-- v=vertical, h=horizontal
> +local function get_cross_sections(abs_geo, mode)
> +print("\n\nLA!",mode)

Uhm. Github GUI says that this is an outdated diff. However, I reviewed
this by git pulling your branch and looking at this locally and noticed the
print there independently from Daniel. What's going on?!?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/awesomeWM/awesome/pull/929/files/405f522b13aa51399e6088c39ecd71086619684f#r65667165>,
or mute the thread
<https://github.com/notifications/unsubscribe/AAUxoNo82wOBni-FNogX9IBBW3F2uQ-Xks5qH9mdgaJpZM4IpNLU>
.

lib/awful/mouse/init.lua
blueyed: JFI: `args.getterer_class` seems wrong in the patch from @psychon ?!

lib/awful/placement.lua
```diff
@@ -552,8 +566,9 @@ attach = function(d, position_f, args)
         end
     end
 
-    -- If there is a parent drawable, screen or mouse, also track it
-    if parent then
+    -- If there is a parent drawable, screen, also track it
+    -- note that tacking the mouse is not supported
```
blueyed: Please use punctuation.  These appear to be two comments, right?!

lib/awful/placement.lua
```diff
@@ -603,6 +618,160 @@ local function rect_to_point(rect, corner_i, corner_j)
...
+    elseif mode == "geometry_inside" then
+        -- The widget absolute geometry, unchanged
+
+        return {h=abs_geo, v=abs_geo}
+    end
+
+    assert(false)
+end
+
+-- When a rectangle is embedded into a bigger one, get the regious around
```
blueyed: s/regious/region(s)/

lib/awful/placement.lua
```diff
@@ -603,6 +618,160 @@ local function rect_to_point(rect, corner_i, corner_j)
...
+-- the outline of the bigger rectangle closest to the smaller one (on each sides)
```
blueyed: s/sides/side/

lib/awful/placement.lua
```diff
@@ -603,6 +618,160 @@ local function rect_to_point(rect, corner_i, corner_j)
...
+        x            = dpos.x + geo.x              ,
+        y            = dpos.y + geo.y              ,
+        width        = geo.width                   ,
+        height       = geo.height                  ,
+        drawable     = geo.drawable                ,
+    }
+
+    abs_widget_geo.drawable_geo = geo.drawable and dpos or geo
+
+    -- Get the comparaison point
```
blueyed: "Get the point for comparison."

lib/awful/placement.lua
```diff
@@ -603,6 +618,160 @@ local function rect_to_point(rect, corner_i, corner_j)
...
+
+    -- Get the 4 closest points from `center_point` around the wibox
+    local regions = {
+        left   = {x = cs.h.x           , y = cs.h.y            },
+        right  = {x = cs.h.x+cs.h.width, y = cs.h.y            },
+        top    = {x = cs.v.x           , y = cs.v.y            },
+        bottom = {x = cs.v.x           , y = cs.v.y+cs.v.height},
+    }
+
+    -- Assume the section is part of a single screen until someone complain.
```
blueyed: s/complain/complains/

lib/awful/placement.lua
```diff
@@ -603,6 +618,160 @@ local function rect_to_point(rect, corner_i, corner_j)
...
+    -- It is much faster to compute and getting it wrong probably have no side
```
blueyed: s/have/has/

lib/awful/placement.lua
```diff
@@ -603,6 +618,160 @@ local function rect_to_point(rect, corner_i, corner_j)
...
+        v.distance = math.sqrt(dx*dx + dy*dy)
+        v.width    = cs.v.width
+        v.height   = cs.h.height
+        v.screen   = capi.screen[s]
+    end
+
+    return regions
+end
+
+-- Check if the proposed geometry fit in the screen
```
blueyed: s/fit in the screen/fits the screen/

lib/awful/placement.lua
```diff
@@ -603,6 +618,160 @@ local function rect_to_point(rect, corner_i, corner_j)
...
+    local sgeo   = get_parent_geometry(obj, args)
+    local region = cairo.Region.create_rectangle(cairo.RectangleInt(sgeo))
+
+    region:intersect(cairo.Region.create_rectangle(
+        cairo.RectangleInt(geo)
+    ))
+
+    local geo2 = region:get_rectangle(0)
+
+    -- If the geometry is the same, then it fit, else, it will be cropped
```
blueyed: s/fit/fits/

lib/awful/placement.lua
```diff
@@ -1143,9 +1322,96 @@ function placement.scale(d, args)
     return fix_new_geometry(ngeo, args, true)
 end
 
----@DOC_awful_placement_maximize_vertically_EXAMPLE@
+--- Move a drawable to a relative position next to another one.
+--
+-- The `args.preferred_positions` looks like:
```
blueyed: s/looks/look/

lib/awful/placement.lua
```diff
@@ -1143,9 +1322,96 @@ function placement.scale(d, args)
...
+        for k, v in ipairs(args.preferred_positions) do
+            preferred_positions[v] = k
+        end
+    end
+
+    local dgeo = geometry_common(d, args)
+
+    local pref_idx, pref_name = 99, nil
+
+    local mode,wgeo = args.mode
```
blueyed: Remove newlines between the assignments?!

lib/awful/tooltip.lua
```diff
@@ -37,41 +37,160 @@
...
+local capi = {mouse=mouse, awesome=awesome}
 
---- Tooltip object definition.
--- @table tooltip
--- @tfield wibox wibox The wibox displaying the tooltip.
--- @tfield boolean visible True if tooltip is visible.
 local tooltip = { mt = {}  }
 
-local instance_mt = {}
+-- The mouse point is 1x1, so anything aligned besed on it as parent
```
blueyed: s/besed/based/

lib/awful/tooltip.lua
```diff
@@ -37,41 +37,160 @@
...
+-- geometry will go out of bound. To get the desired placement, it is
+-- necessary to swap left<-> and top<->bottom
```
blueyed: s/swap left<-> and top<->bottom/swap left with right and top with bottom/

lib/awful/tooltip.lua
```diff
@@ -37,41 +37,160 @@
...
+local align_convert = {
+    top_left     = "bottom_right",
+    left         = "right",
+    bottom_left  = "top_right",
+    right        = "left",
+    top_right    = "bottom_left",
+    bottom_right = "top_left",
+    top          = "bottom",
+    bottom       = "top",
+}
```
blueyed: This could use string substitution maybe?! (but feel free to ignore)

lib/awful/tooltip.lua
```diff
@@ -37,41 +37,160 @@
...
+-- @beautiful beautiful.tooltip_font
+
+--- The tooltip border width.
+-- @beautiful beautiful.tooltip_border_width
+
+--- The tooltip opacity.
+-- @beautiful beautiful.tooltip_opacity
+
+--- The tooltip shape.
+-- @beautiful beautiful.tooltip_shape
```
blueyed: "shape" is not very specific.  Could need some more details.

lib/awful/tooltip.lua
```diff
@@ -37,41 +37,160 @@
...
+    local img = cairo.ImageSurface(cairo.Format.A1, w, h)
+    local cr = cairo.Context(img)
+
+    cr:set_source_rgba(1,1,1,1)
+
+    s(cr, w, h, unpack(self._tooltip._private.shape_args or {}))
+    cr:fill()
+    wb.shape_bounding = img._native
+
+    -- The wibox background uses ARGB32 border so tooltip anti-aliasing work
```
blueyed: s/work/works only/

lib/awful/tooltip.lua
```diff
@@ -37,41 +37,160 @@
...
+    -- when an external compositor is used. This will look better than
+    -- the capi.drawin own border support.
```
blueyed: s/capi.drawin/capi.drawin's/

lib/awful/tooltip.lua
```diff
@@ -37,41 +37,160 @@
...
+    img = cairo.ImageSurface(cairo.Format.ARGB32, w, h)
+    cr  = cairo.Context(img)
+
+    -- Draw the border ( multiply by 2, then mask the inner part to save a path)
```
blueyed: s/( /(/

lib/awful/tooltip.lua
```diff
@@ -37,41 +37,160 @@
...
+    -- the -1
```
blueyed: Leftover?

lib/awful/tooltip.lua
```diff
@@ -37,41 +37,160 @@
...
+
+    wb:set_bgimage(img)
+end
+
+local function apply_mouse_mode(self)
+    local w = self:get_wibox()
+
+    local align  = self._private.align
+
+    local real_placement = align_convert[align]
```
blueyed: Remove newlines between assignments.

lib/awful/tooltip.lua
```diff
@@ -110,46 +244,190 @@ end
...
+function tooltip:get_wibox()
+    if self._private.wibox then
+        return self._private.wibox
+    end
+
+    local wb = wibox(self.wibox_properties)
+    wb:set_widget(self.marginbox)
+
+    -- Close the tooltip when clicking it.  This gets done on release, to not
+    -- emit the release event on an underlying object, e.g. the titlebar icon.
```
blueyed: Awesome comment! :cherries:

lib/awful/tooltip.lua
```diff
@@ -110,46 +244,190 @@ end
...
+    wb:buttons(abutton({}, 1, nil, self.hide))
+
+    self._private.wibox = wb
+    wb._tooltip = self
+
+    return wb
+end
+
+--- If the tooltip is visible.
```
blueyed: "Is the tooltip visible?"

lib/awful/tooltip.lua
```diff
@@ -110,46 +244,190 @@ end
...
+-- @tparam gears.shape s The shape
+-- @see shape
+function tooltip:set_shape(s, ...)
+    self._private.shape = s
+    self._private.shape_args = {...}
+    apply_shape(self.wibox)
+end
+
+--- Set the tooltip positioning mode.
+-- This affect how the tooltip is placed. By default, the tooltip is `align`ed
```
blueyed: s/affect/affects/

lib/awful/tooltip.lua
```diff
@@ -110,46 +244,190 @@ end
...
+-- close to the mouse cursor. It is also possible place the tooltip relative to
```
blueyed: s/possible place/possible to place/

lib/awful/tooltip.lua
```diff
@@ -110,46 +244,190 @@ end
...
+    self:emit_signal("property::mode")
+end
+
+function tooltip:get_mode()
+    return self._private.mode or "mouse"
+end
+
+--- The preferred positions when in `outside` mode.
+--
+-- If the tooltip fits on different side of the drawable, then pick the
```
blueyed: s/different side/multiple sides/

lib/awful/tooltip.lua
```diff
@@ -110,46 +244,190 @@ end
...
+-- priorities.
```
blueyed: s/then pick the priorities/then this defines the priority/

lib/awful/tooltip.lua
```diff
@@ -110,46 +244,190 @@ end
...
+--
+-- The default is:
+--
+--    {"top", "right", "left", "bottom"}
+--
+-- @property preferred_positions
+-- @tparam table preferred_positions The position, orderred by priorities
```
blueyed: s/orderred/ordered/

lib/awful/tooltip.lua
```diff
@@ -170,11 +449,21 @@ end
...
 --   `mouse::leave` signals.
-tooltip.remove_from_object = function(self, obj)
+-- @function remove_from_object
+function tooltip:remove_from_object(obj)
     obj:disconnect_signal("mouse::enter", self.show)
     obj:disconnect_signal("mouse::leave", self.hide)
 end
 
+-- Tooltip can be applied to both widgets, wibox and client, their geometry
+-- work differently.
```
blueyed: s/work/works/

tests/test-tooltip.lua
```diff
@@ -0,0 +1,124 @@
...
+local steps = {}
+
+local w = wibox {
+    width   = 250,
+    height  = 250,
+    visible = true,
+    ontop   = true
+}
+
+w: setup {
```
blueyed: This looks weird with the space inbetween.

blueyed:
Left some comments and tried it actually.

It breaks both my custom tasklist_update, which uses the private `tb._layout` (which is arguably a fault on my side), but also the tooltips for my widgets do not show anymore.

I only see "SHOW" through https://github.com/awesomeWM/awesome/blob/42382bb5f6dfdb27cdc2b92637110909973bad47/lib/awful/tooltip.lua#L529-L535.

actionless:
> Notification shape/placement (this require the dynamic layout system and I wont push it into 3.6)

that's really sad, because i like how it looks (made with smart_wibox from radical):

2016-06-04-113645_3840x2160_scrot

also there is some minor issue with inconsistency between clipping and border shape but that's out of the scope of this conversation

Elv13:
>  but the text is missing.

@blueyed Can you investigate a bit. All I can say is that the tests display text and the titlebar tooltips also display correctly. Given your config isn't public, there is little else I can do.

> that's really sad, because i like how it looks (made with smart_wibox from radical):

Sad, yes, but very time consuming to fix. Feel free to do it yourself. Your screenshot is a fine example of why I don't want to upstream those parts yet. You see the black pixels in the corners? Those would go away if everything was properly implemented (and a compositing manager was used). Getting anti-aliased border (both external and internal) wont be that easy. In fact, it is the same issue as #954. Wiboxes should use a `background` widget and drop the internal border support. However, the list of nasty side effects that would have to be fixed (including tons of geometry code) is too much for now.

blueyed:
@Elv13 
The tooltip also appears empty (or with the same fg/bg color) for the default titlebar with the icons there.
It is related to my custom theme however, which is derived from the xresources theme.  This is affected by the same issue.

Elv13:
Can you set beautiful.tooltip_bg and fg to confirm this? Also, is it a regression? what was the color before?

lib/awful/tooltip.lua
```diff
@@ -244,27 +550,26 @@ tooltip.new = function(args)
         self.timer:connect_signal("timeout", self.timer_function)
     end
 
+    local fg = beautiful.tooltip_fg or beautiful.fg_focus or "#000000"
```
blueyed: @Elv13 
This breaks for the xresources theme: it seems to use the same fg as bg.
Forcing this to be "#000000" makes the text of tooltips appear.

lib/awful/tooltip.lua
```diff
@@ -244,27 +550,26 @@ tooltip.new = function(args)
...
+
     -- Set default properties
     self.wibox_properties = {
         visible = false,
         ontop = true,
-        border_width = beautiful.tooltip_border_width or beautiful.border_width or 1,
-        border_color = beautiful.tooltip_border_color or beautiful.border_normal or "#ffcb60",
+        border_width = 0,
+        fg = fg,
+        bg = color.transparent,
```
blueyed: @Elv13 
I have also tried using "#000000" here, but it was not applied.

blueyed:
@Elv13 
> Can you set beautiful.tooltip_bg and fg to confirm this?

Yes, forcing this to red and green works.

> Also, is it a regression? what was the color before?

Based on your xrdb settings.

"fg_color" is still mentioned there: https://github.com/Elv13/awesome-1/blob/geometry_overhaul_p3.03/themes/xresources/theme.lua#L40 - but it does not seem to be used?! (assets: https://github.com/Elv13/awesome-1/blob/geometry_overhaul_p3.03/themes/xresources/assets.lua).

blueyed:
@actionless 
Does the xresources theme work for you with this PR?

actionless:
yup, tooltips are black on black then

i think in order to fix that we should just update the theme to have tooltip_{f,b}g explicitly defined

Elv13:
@actionless or @blueyed Can you provide a patch or propose a change inline here. Which xresources color do you want for the tooltip fg and bg?

blueyed:
@actionless 
Can you look into it, please?
I am using some workaround, but am not sure what exactly.

blueyed:
@Elv13 
Shouldn't the defaults cover this use case also, or is the xresources theme weird/wrong in this regard?

Elv13:
cpu1

They work in all my themes and the default one. It "should" work with all themes with `fg_focus` set. However I just noticed that the old code use `bg_focus` as the fallback while the new one use `bg_normal`, so this part is wrong on my side (so is not correctly handle the old color name)

blueyed:
So, given that the xresources-theme gets adjusted I think this can be merged then, too?!

blueyed:
There are still some unaddressed comments in the file/diff view.
Assigning to @Elv13.

Elv13:
Rebased, the comments have been addressed

lib/awful/tooltip.lua
```diff
@@ -189,14 +483,28 @@ end
...
+function tooltip.new(args)
+    local self = object {
+        enable_properties = true,
+    }
+
+    rawset(self,"_private", {})
+
+    self:add_signal("property::visible")
+    self:add_signal("property::align")
+    self:add_signal("property::mode")
```
psychon: I haven't looked at this at all. I just saw a metric ton of warnings in the travis build (the LuaJIT-build failed because of `tests/test-screen-changes.lua`, our unstable test, running into a timeout).
Since commit 0857f6f1b5535a862c840311f5bdad849f42e95d, `add_signal` doesn't do anything except for printing a warning, so these three calls can be removed.

Elv13:
I removed the `add_signal`s. Is this ok now? I am waiting for this before pushing some commits in my modules.

lib/awful/mouse/init.lua
```diff
@@ -319,6 +318,32 @@ function mouse.object.get_current_widget()
     end
 end
 
+function mouse.object.set_current_widget() end
```
psychon: Empty setters are no longer necessary since cdf16d4660d73c7db988723aaee40076f7eaa9fe. Please remove (all of them).

lib/awful/tooltip.lua
```diff
@@ -441,7 +527,10 @@ function tooltip.new(args)
             hide(self)
         end
     else
-        function self.show()
+        function self.show(other, geo)
```
psychon: I don't really understand how this works. What type of object are `other` and `geo` here?

psychon:
I looked through this and left my comments in the form of fixup commits. I had to do a rebase so should I could add a fixup commit in the right place (commit A had a `print`, commit B removed it; I wanted to remove it from commit A). So, this is now rebased on lastest master and commit 3831d29cd69d34a7a2f0c93 is gone.

Please look through my changes, decide if they are good and then apply them if you are agree (`git checkout geometry_overhaul_p3.03 ; git rebase -i --autosqash master`). In case this whole fixup commits idea is wrong: `git reset --hard 0a9ea10710b928975bf4767d4b5a8f1ce94dcf25` (this is the branch before I changed something).

Please look at the commit message from 1c492ec96b86fb41068d6823b8b1eb23999f4b6e and 3555a26c2cb914d2e2ce6af72ab2d1a1edb5060d, because I am not 100% sure about these changes.

The only commit that I left here at github and that is still relevant: How are the arguments to `self.show` supposed to work? What is their type?

Elv13:
Everything looks alright. Thanks!

> The only commit that I left here at github and that is still relevant: How are the arguments to self.show supposed to work? What is their type?

See `mouse::enter` arguments. It is a little magic, I will give you that.

Feel free to merge this.

blueyed:
Before merging this should be interactively rebased.. (to remove the fixup commits).

psychon:
> See mouse::enter arguments. It is a little magic, I will give you that.

On a drawin, there are no arguments.
On a wibox, there are no arguments.
On a widget, it's a table with entries `x, y, width, height, drawable, widget`.

None of this has two arguments and I cannot see how the code you add in this PR can work on anything but widgets.

Elv13:
> None of this has two arguments and I cannot see how the code you add in this PR can work on anything but widgets.

But does it even make sense to use those on anything but widgets? Why would you want a tooltip on a wibox directly?

psychon:
I bet someone can come up with a reason why they want a tooltip on their conky window.
Anyway, changing this code to work only on widgets should be documented better. The code looks like it tries to work on multiple types of objects. E.g. the whole `get_parent_geometry` exists, but serves no purpose...? Also, the API docs talk only about "object", not "widget".

How about changing `add_to_object` so that things work on all kind of objects? It would require some magic to detect the kind of argument and creating some new function with the object as upvalue. (which makes `remove_from_object` harder to implement...)
For example:
```lua
function tooltip:add_from_object(obj)
    if self._private.connected_objects[obj] then return end
    self._private.connected_objects[obj] = {
        show = function(...) self.show(obj, ...) end,
        hide = function(...) self.hide(obj, ...) end,
    }
    obj:connect_signal("mouse::enter", self._private.connected_objects[obj].show)
    obj:connect_signal("mouse::leave", self._private.connected_objects[obj].hide)
end
function tooltip:remove_from_object(obj)
    if not self._private.connected_objects[obj] then return end
    obj:disconnect_signal("mouse::enter", self._private.connected_objects[obj].show)
    obj:disconnect_signal("mouse::leave", self._private.connected_objects[obj].hide)
    self._private.connected_objects[obj] = nil
end
```

Elv13:
Rebased and fixed. I also have a new commit for anti-aliased tooltips to fix the issue in @actionless screenshot. I didn't add it to the PR and will probably submit it once it is merged. The current tooltip isn't affected by the issue, but lack outer anti-aliasing.

Elv13:
@psychon Is the changes I made are good enough for this to be merged?

lib/awful/mouse/init.lua
```diff
@@ -319,6 +318,28 @@ function mouse.object.get_current_widget()
...
+-- @see current_widget
+
+function mouse.object.get_current_widget_geometry()
+    local _, ret = mouse.object.get_current_widget()
+
+    return ret
+end
+
+--- Get the current widget geometries.
+-- @property current_widget_geometriesgeometriesy
```
psychon: I cannot really call this a typo, but only because it is worse than that. :-P

lib/awful/tooltip.lua
```diff
@@ -83,7 +86,87 @@ local offset = {
...
+-- they can default to rounded rectangle or stretched octogons.
+-- @beautiful beautiful.tooltip_shape
+-- @tparam[opt=gears.shape.rectangle] function shape A `gears.shape` compatible function
+-- @see shape
+-- @see gears.shape
+
+local function apply_shape(self)
+    local s = self._private.shape
+
+    if not s then return end
```
psychon: I think this has to do things like `self.wibox.shape_bounding = nil` and and `self.wibox:set_bg_image(nil)` (or something like this, I'm not completely sure).

Put differently: Does unsetting the shape work correctly?

lib/awful/tooltip.lua
```diff
@@ -37,41 +37,162 @@
...
+local function apply_outside_mode(self)
+    local w = self:get_wibox()
+
+    local _, position = a_placement.next_to(w, {
+        geometry            = self._private.widget_geometry,
+        preferred_positions = self.preferred_positions,
+        honor_workarea      = true,
+    })
+
+    if position ~= self._private.direction then
```
psychon: Correct me if I'm wrong, but `self._private.direction` is always `nil`?

psychon:
The above comments are just minor things (and I will soon be offline for another week), so feel free to merge this after you are done.

PR #932, opened 2016-05-31
"Fix" client properties which can be absent

psychon:
A client c could have no c.machine or no c.pid because the corresponding
properties are not set on its window. Previously, the C code would return an
empty string or 0 for these values. This commit makes the C code give Lua no
value instead (not even a nil).

### DIFF
--- a/common/luaobject.h
+++ b/common/luaobject.h
@@ -195,6 +195,16 @@ int luaA_object_emit_signal_simple(lua_State *);
         return 1; \
     }
 
+#define LUA_OBJECT_EXPORT_OPTIONAL_PROPERTY(pfx, type, field, pusher, empty_value) \
+    static int \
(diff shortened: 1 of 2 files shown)


common/luaobject.h
```diff
@@ -195,6 +195,16 @@ int luaA_object_emit_signal_simple(lua_State *);
         return 1; \
     }
 
+#define LUA_OBJECT_EXPORT_PROPERTY2(pfx, type, field, pusher, empty_value) \
```
actionless: sorry for bikeshedding proposal but mb better name it LUA_OBJECT_EXPORT_PROPERTY_WFALLBACK or somehow like that?

psychon: I squashed the following into this commit, is that ok? https://gist.github.com/psychon/d738c616db977eafff8149b1a1eff95f

actionless: looks good

PR #936, opened 2016-05-31
Travis: test each commit

blueyed:
TODO:
 - [ ] would be nice if we could set the status for previous commits in the PR then (like you have it currently for commits that where built/tested)  -- I've asked the Travis support team about this, but let's not hold this back for this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,10 @@ language: c
 env:
   matrix:
     - LUA=5.2 LUANAME=lua5.2 BUILD_APIDOC=true DO_COVERAGE=coveralls
+    # luajit: installed from source.
+    - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.4 TEST_PREV_COMMITS=1
     # Note: luarocks does not work with Lua 5.0.
     - LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build
-    # luajit: installed from source.
-    - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.4
     # Lua 5.2 with older lgi and screen size not divisible by 2.
     - LUA=5.2 LUANAME=lua5.2 LGIVER=0.7.1 TESTS_SCREEN_SIZE=1921x1079
     # Lua 5.3 isn't available in Ubuntu Trusty, so some magic below installs it.
@@ -102,16 +102,42 @@ install:
       fi
     }
 
+  # Functions for custom Travis folds.
+  - |
+    travis_fold_start() {
+      echo -n "travis_"
+      echo "fold:start:$1"
+      travis_fold_current="$1"
+    }
+    travis_fold_end() {
+      echo -n "travis_"
+      echo "fold:end:$travis_fold_current"
+    }
+    travis_run_in_fold() {
+      travis_fold_start "$1"
+      shift
+      "$@"
+      local result="$?"
+      travis_fold_end
+      return $result
+    }
 script:
   - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -DLUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -D DO_COVERAGE=${DO_COVERAGE}"
   - |
     if [ -n "$BUILD_IN_DIR" ]; then
-        SOURCE_DIRECTORY="$PWD"
-        mkdir "$BUILD_IN_DIR"
-        cd "$BUILD_IN_DIR"
-        cmake $CMAKE_ARGS "$SOURCE_DIRECTORY"
+      # Explicitly remove the Makefile to not build from the src dir accidentally.
+      rm Makefile
+      SOURCE_DIRECTORY="$PWD"
+      mkdir "$BUILD_IN_DIR"
+      cd "$BUILD_IN_DIR"
+      travis_run_in_fold "build_in_dir" cmake $CMAKE_ARGS "$SOURCE_DIRECTORY"
     fi
-  - make && sudo env PATH=$PATH make install && awesome --version
+  - travis_run_in_fold "make" make
+  - |
+    if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
+      travis_run_in_fold "make.install" sudo env PATH=$PATH make install
+      awesome --version
+    fi
   - |
     if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
       do_codecov samples
@@ -119,13 +145,44 @@ script:
   - |
     if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
       if [ "$DO_COVERAGE" != "0" ]; then
-        make check-coverage || exit 1
-        do_codecov unittests
-        sed -i "1 i\\require('luacov.runner')('"$PWD"/.luacov')" build/awesomerc.lua || exit 1
-        BUILD_DIRECTORY="" tests/run.sh || exit 1
-        do_codecov functionaltests
+        travis_fold_start "DO_COVERAGE"
+        (make check-coverage \
+         && do_codecov unittests \
+         && sed -i "1 i\\require('luacov.runner')('"$PWD"/.luacov')" build/awesomerc.lua \
+         && BUILD_DIRECTORY="" tests/run.sh \
+         && do_codecov functionaltests)
+        ret=$?
+        travis_fold_end
+        exit $ret
       else
-        make check
+        travis_run_in_fold "make.check" make check
+      fi
+    fi
+  - |
+    if [ "$TEST_PREV_COMMITS" = 1 ] && ! [ "$TRAVIS_PULL_REQUEST" = false ]; then
+      # Check each commit separately (to make git-bisect less annoying).
+      echo "Testing previous commits based on TRAVIS_COMMIT_RANGE=$TRAVIS_COMMIT_RANGE"
+      rev_list="$(git rev-list --bisect-all $TRAVIS_COMMIT_RANGE)"
+      echo "rev-list: $rev_list"
+      commits="$(echo $rev_list | grep -v 'dist=0' | cut -d\  -f 1)"
+      echo "Testing commits: $commits"
+      n="$(echo commits | wc -l)"
+      i=0
+      failed=
+      for commit in $commits; do
+        i=$((i+1))
+        travis_fold_start "test_commit_${commit}_.$i.$n"
+        echo "Testing commit $commit"
+        git checkout "$commit"
+        git --no-pager show --stat
+        if ! make all check; then
+          failed="$failed $commit"
+        fi
+        travis_fold_end
+      done
+      if [ -n "$failed" ]; then
+        echo "Checks failed for these commits:$failed"
+        exit 1
       fi
     fi
 


blueyed:
@psychon 
What do you think?
It needs to be cleaned up and the broken test commit has to be removed, of course.

.travis.yml
```diff
@@ -90,21 +90,73 @@ install:
...
+  - |
+    travis_fold_start "make"
+    make
+    travis_fold_end
+    if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
+      travis_fold_start "make.install"
+      sudo env PATH=$PATH make install && awesome --version
+      travis_fold_end
+    fi
+    if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
```
psychon: Why the many new checks for `$TRAVIS_TEST_RESULT`?

.travis.yml
```diff
@@ -90,21 +90,73 @@ install:
...
+      commits="$(git rev-list --abbrev-commit "$TRAVIS_COMMIT_RANGE")"
+      echo "Commits: $commits"
+      n="$(echo comiits | wc -l)"
+      i=0
+      failed=
+      for commit in $commits; do
+        i=$((i+1))
+        travis_fold_start "test_commit_${commit}_.$i.$n"
+        git checkout "$commit"
+        if ! make check; then
```
psychon: I don't think just `make check` is enough. We don't have the dependencies set up correctly so that this e.g. rebuilds the `awesome` binary if needed. I guess you need to run `make all check` for this.

.travis.yml
```diff
@@ -90,21 +90,73 @@ install:
...
+  - |
+    travis_fold_start() {
+        echo -n "travis_"
+        echo "fold:start:$1"
+        travis_fold_current="$1"
+    }
+    travis_fold_end() {
+        echo -n "travis_"
+        echo "fold:end:$travis_fold_current"
+    }
```
psychon: Perhaps it is nice to additionally do something like the following:
```sh
travis_run_in_fold() {
   travis_fold_start "$@"
   "$@"
   local result="$?"
   travis_fold_end "$@"
   ( exit "$result" )
}
```
Usage would be `travis_run_in_fold make`. Would this simplify some of the rest of the file.

psychon: Oh, I guess it should be `return $result` instead of `( exit "$result" )`. That avoids spawning a subshell. :-)

psychon:
I tried `autosquash`ing this myself, but the result is still confusing. I just looked at the total diff instead and left some comments. I guess this is fine with me, but it will surely blow up from time to time... I guess we can live with that (it will only happen on complicated, breaking changes and I don't yet know how it will happen exactly)

.travis.yml
blueyed: "script" dies not stop if one command fails.

psychon: But we already had it this way before, didn't we? And it worked...?

blueyed: Well, for the `make check` I changed it explicitly to use `&&` in between.
It is now necessary, because I've gone wild with adding manual Travis fold markers, and therefore the `&&` trick does not work.  Although I could also skim the calls to `travis_fold_*` into it..

Please note that the checking of each commit does not have this check: it should get done always.

.travis.yml
blueyed: Ok, changed.

.travis.yml
blueyed: I've gone that route, but it seemed like it cannot be (sanely) used everywhere, so I've reverted it.
It makes sense for the simple `make` though, of course.  We can keep that in mind.

blueyed: Also, you need a "clean" marker for the folder, you cannot use an abbreviate string there.  So that would be the (optional) first argument for longer commands.

blueyed:
I've removed the fail-for-testing commit and cleaned it up.  Should be good from my side.

blueyed:
I'd like to see this being merged before https://github.com/awesomeWM/awesome/pull/929 - not to annoy @Elv13, but to see how it behaves.

blueyed:
I assume/fear that this will run the test twice for a PR with a single commit, too: once for the merge commit and then for the non-merge commit that shows up via `git rev-list --abbrev-commit "$TRAVIS_COMMIT_RANGE"` - I've not really investigated that area.

blueyed:
I've amended it into a single commit, we should see it then from the Travis logs.

blueyed:
@psychon 
@coveralls is a bit annoying altogether with adding new comments all the time, even if coverage stays the same (and therefore also if it improves probably).  Could we try @codecov instead?!  I don't remember the details.

blueyed:
I've changed it to only run against PRs, and then remove the top-most commit from `rev-list` against the target branch.

psychon:
> Could we try @codecov instead?! I don't remember the details.

Sure, feel free. I went with coveralls since there was an already existing LuaCov -> coveralls project.

.travis.yml
```diff
@@ -4,10 +4,10 @@ language: c
 env:
```
psychon: commit message: s/amd/and/
Also: The travis build failed complaining about a syntax error.

.travis.yml
```diff
@@ -90,21 +90,75 @@ install:
...
-        mkdir "$BUILD_IN_DIR"
-        cd "$BUILD_IN_DIR"
-        cmake $CMAKE_ARGS "$SOURCE_DIRECTORY"
+      travis_fold_start "build_in_dir"
+      SOURCE_DIRECTORY="$PWD"
+      mkdir "$BUILD_IN_DIR"
+      cd "$BUILD_IN_DIR"
+      cmake $CMAKE_ARGS "$SOURCE_DIRECTORY"
+      cd -
+      travis_fold_end
```
psychon: How does this `build_in_dir`-fold interact with the fold that travis automatically creates for each separate command?

.travis.yml
```diff
@@ -90,21 +90,75 @@ install:
...
-        SOURCE_DIRECTORY="$PWD"
-        mkdir "$BUILD_IN_DIR"
-        cd "$BUILD_IN_DIR"
-        cmake $CMAKE_ARGS "$SOURCE_DIRECTORY"
+      travis_fold_start "build_in_dir"
+      SOURCE_DIRECTORY="$PWD"
+      mkdir "$BUILD_IN_DIR"
+      cd "$BUILD_IN_DIR"
+      cmake $CMAKE_ARGS "$SOURCE_DIRECTORY"
+      cd -
```
psychon: This `cd -` actually breaks the "build in dir" stuff.

What happens after your change:
- CMake is run in `$BUILD_IN_DIR`
- The script `cd`s back into the source dir
- CMake is run against for the "normal" `build/` stuff and the `$BUILD_IN_DIR` is ignored.

.travis.yml
```diff
@@ -90,21 +90,75 @@ install:
...
+      BUILD_DIRECTORY="" tests/run.sh || exit 1
+      travis_fold_end
     fi
-  - make && sudo env PATH=$PATH make install && awesome --version && make check
   - |
-    if [ "$DO_COVERAGE" = "true" ]; then
-        make check-coverage || exit 1;
-        sed -i "1 i\\require('luacov.runner')('"$PWD"/.luacov')" build/awesomerc.lua || exit 1
-        BUILD_DIRECTORY="" tests/run.sh || exit 1
+    if [ "$TEST_PREV_COMMITS" = 1 ] && { "$TRAVIS_PULL_REQUEST" = false }; then
```
psychon: I don't think you meant to use `{` here (at least this doesn't work)

.travis.yml
psychon: This does not work!

`.travis.yml`:
```
script:
        - |
          echo $TRAVIS_TEST_RESULT
          true
          echo $TRAVIS_TEST_RESULT
          false
          echo $TRAVIS_TEST_RESULT
```
Build: https://travis-ci.org/psychon/awesome/builds/135220421
(Relevant part is: `$TRAVIS_TEST_RESULT` always stays 0 and the build is successfull)

So this variable is only updated at the end of each "command block".

.travis.yml
psychon: Actually... can we somehow ensure that this isn't broken accidentally again later? How about adding `rm Makefile` as the first thing in this block to make sure that the "default magic" will not be used later (together with adding a comment explaining why this is done).

I'd do a PR adding this, but it would conflict with your PR.

.travis.yml
blueyed: There are no automatic folds for the items in `script`..

psychon: Oh, right. Sorry!

.travis.yml
```diff
@@ -90,24 +90,81 @@ install:
...
+      rm Makefile
+      SOURCE_DIRECTORY="$PWD"
+      mkdir "$BUILD_IN_DIR"
+      cd "$BUILD_IN_DIR"
+      travis_run_in_fold "build_in_dir" cmake $CMAKE_ARGS "$SOURCE_DIRECTORY"
+    fi
+  - travis_run_in_fold "make" make
+  - |
+    if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
+      travis_run_in_fold "make.install" sudo env PATH=$PATH make install
```
psychon: Does this need an `|| exit 1` in case installation fails (for whatever reason)?

.travis.yml
```diff
@@ -90,24 +90,81 @@ install:
...
+    fi
+  - |
+    if [ "$TEST_PREV_COMMITS" = 1 ] && ! [ "$TRAVIS_PULL_REQUEST" = false ]; then
+      # Check each commit separately (to make git-bisect less annoying).
+      echo "Testing previous commits based on TRAVIS_COMMIT_RANGE=$TRAVIS_COMMIT_RANGE"
+      # Get the commits diff to the target branch, removing the merge commit at
+      # the top, which has already been tested.
+      commits="$(git rev-list --abbrev-commit "...$TRAVIS_BRANCH" | head -n1)"
+      echo "Commits: $commits"
+      n="$(echo comiits | wc -l)"
```
psychon: comiits -> commits

psychon:
Looks fine to me. Adding to next assuming that some minor issues are fixed (and that the travis build actually works afterwards; no idea why no travis run was done for your fixup commit).

.travis.yml
blueyed: OMG, thanks!

.travis.yml
blueyed: Not really.. then then `awesome --version` should fail afterwards.

blueyed:
@psychon 
Thanks for your review.

I will merge it once it's green again.

blueyed:
The list of commits was (totally) buggy.

Changed it:
```diff
--- a/.travis.yml
+++ b/.travis.yml
@@ -146,8 +146,10 @@ script:
       echo "Testing previous commits based on TRAVIS_COMMIT_RANGE=$TRAVIS_COMMIT_RANGE"
       # Get the commits diff to the target branch, removing the merge commit at
       # the top, which has already been tested.
-      commits="$(git rev-list --abbrev-commit "...$TRAVIS_BRANCH" | head -n1)"
-      echo "Commits: $commits"
+      rev_list="$(git rev-list --bisect-all $TRAVIS_COMMIT_RANGE)"
+      echo "rev-list: $rev_list"
+      commits="$(echo $rev_list | grep -v 'dist=0' | cut -d\  -f 1)"
+      echo "Testing commits: $commits"
       n="$(echo commits | wc -l)"
       i=0
       failed=
```

Elv13:
Umm, the coverage took a large hit from the last version of this patch...

blueyed:
@Elv13 
Yeah, I've resolved the conflict in a wrong way when rebasing.. hopefully better now / in the next build.

blueyed:
Still down by 11%.
@psychon 
What am I missing / doing wrong?

psychon:
`do_codecov functionaltests` is gone? Aka the coverage from `tests/run.sh` is never uploaded.

blueyed:
Travis so broken currently, we can merge this also.. ;)

PR #937, opened 2016-05-31
Travis: kill awesome after max. 60s per test file

blueyed:
This is required for when the test setup fails already, e.g. because of
an assertion error at the top of a test file.

Example of a failing build: https://travis-ci.org/awesomeWM/awesome/jobs/134291905.

Should also help with hangs in general: https://api.travis-ci.org/jobs/134295087/log.txt?deansi=true (struts test)

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -189,6 +189,8 @@ start_awesome() {
 
 # Count errors.
 errors=0
+# Seconds after when awesome gets killed.
+timeout_stale=60
 
 for f in $tests; do
     echo "== Running $f =="
@@ -203,6 +205,13 @@ for f in $tests; do
 
     # Send the test file to awesome.
     cat $f | DISPLAY=$D "$AWESOME_CLIENT" 2>&1
+
+    # Kill awesome after 1 minute (e.g. with errors during test setup).
+    (sleep $timeout_stale
+      if [ "$(ps -o comm= $awesome_pid)" = "${AWESOME##*/}" ]; then
+        echo "Killing (stale?!) awesome (PID $awesome_pid) after $timeout_stale seconds."
+        kill $awesome_pid
+      fi) &
 
     # Tail the log and quit, when awesome quits.
     tail -n 100000 -f --pid $awesome_pid $awesome_log
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -103,7 +103,7 @@ wait_until_success() {
     if [ "$CI" = true ]; then
         set +x
     fi
-    max_wait=60
+    wait_count=60  # 60*0.05s => 3s.
     while true; do
         set +e
         eval reply="\$($2)"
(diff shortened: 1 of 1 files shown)


tests/run.sh
```diff
@@ -103,7 +103,7 @@ wait_until_success() {
     if [ "$CI" = true ]; then
         set +x
     fi
-    max_wait=60
+    max_wait=30
```
psychon: This is the maximum number of loop iteration, not the number of seconds to wait. In the loop you will see a `sleep 0.05`, so the old time is `0.05s * 60 = 3 seconds` while the new time is 1.5 seconds. Not much of a difference. (The commit message talks about seconds)

tests/run.sh
```diff
@@ -204,6 +204,12 @@ for f in $tests; do
     # Send the test file to awesome.
     cat $f | DISPLAY=$D "$AWESOME_CLIENT" 2>&1
 
+    # Kill awesome after 1 minute (e.g. with errors during test setup).
+    (sleep 60; if kill -0 $awesome_pid 2>/dev/null; then
+        echo "Killing (stale?!) awesome (PID $awesome_pid) after 10 seconds."
+        kill $awesome_pid
+      fi) &
+
```
psychon: Perhaps it would be easier to do `echo 'require("gears.timer").start_new(60, function() awesome.quit() end)' | DISPLAY=$D "$AWESOME_CLIENT"`. Less code and less likely that the script kills a random, other PID.

blueyed: Seems like a good idea, but on the other hand that might fail already somehow.

I think the `kill` is safe, since it's really supposed to be the PID of awesome.

blueyed: @psychon 
I will add a check using `ps -o comm= $awesome_pid` here to make sure that it's awesome.

Does that sound good to you?

blueyed:
Moving to next.. but I'd like to see some ACK before merging it.
Thanks!

tests/run.sh
psychon: Sigh. Just keep it as a `kill`. It's unlikely that the PID is reused so quickly, so we should not over-engineer this.

tests/run.sh
```diff
@@ -204,6 +204,12 @@ for f in $tests; do
     # Send the test file to awesome.
     cat $f | DISPLAY=$D "$AWESOME_CLIENT" 2>&1
 
+    # Kill awesome after 1 minute (e.g. with errors during test setup).
+    (sleep 60; if [ "$(ps -o comm= $awesome_pid)" = "${AWESOME##*/}" ]; then
+        echo "Killing (stale?!) awesome (PID $awesome_pid) after 10 seconds."
```
psychon: The message here talks about 10 seconds. You might want to fix that (and perhaps move the timeout into a variable (at the top of the file?) so that this cannot happen again).

psychon: Oh and apropos what this commit actually does: I tried to fix this with e18bece3df8911102c92f60fc06fa5da62438c35, but I guess the problem was re-introduced since then? Am I understanding this correctly? If so, your fix (while ugly) should fix this once and for all.

tests/run.sh
```diff
@@ -103,7 +103,7 @@ wait_until_success() {
     if [ "$CI" = true ]; then
```
psychon: The link in the commit message is broken

tests/run.sh
blueyed: Oh, yeah.. I've noticed this with `tests/test-awful-client.lua` where the runner is not included at the beginning.

psychon:
Assuming the travis build passes: :+1:

PR #944, opened 2016-06-04
Silence warnings in examples

psychon:
There was a problem that the examples were considered to have failed as soon as
they produced any kind of output, but there were legitimate cases of warnings
being printed that triggered these checks. Commit 4819be4f4f3037ef3f used a
regular expression to detect and ignore this warnings.

This commit reverts the above commit and instead silences the warnings by
monkey-patching the function that prints the warnings into a no-op.

CC @Elv13

### DIFF
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -106,32 +106,6 @@ function(escape_code path escaped_content pre_header post_header)
     set(${post_header} ${example_post_header} PARENT_SCOPE)
 endfunction()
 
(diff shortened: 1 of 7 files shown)


blueyed:
I think the monkey-patching should be done in a common include / setup file?!

psychon:
I thought so, too, but in each of these files the stuff before the part that is included is also something that could be shared (loading all the shims). I think moving this into some common file is a task left for the future (and if needed, I'll volunteer to code that).

PR #945, opened 2016-06-04
Tests runner

psychon:
This makes the test runner for `tests/examples/` non-recursive and hopefully a bit easier to understand. Both of my original goals (run all tests in a single Lua process, run tests during the build step and not during the configure step) seems quite hard to reach and so I came up with this instead...

CC @Elv13


blueyed:
@Elv13 
Any feedback on this, please?

Apart from that it needs to be rebased.

Elv13:
And it does solve the scope issue. +1

PR #946, opened 2016-06-04
Readonly property

psychon:
Some readonly properties and getting rid of useless no-op setters.

### DIFF
--- a/tests/test-miss-handlers.lua
+++ b/tests/test-miss-handlers.lua
@@ -40,4 +40,8 @@ local w = wibox()
 w.foo = "bar"
 assert(w.foo == "bar")
 
+-- Test if read-only properties really are read-only
+screen[1].clients = 42
+assert(screen[1].clients ~= 42)
+
 require("_runner").run_steps({ function() return true end })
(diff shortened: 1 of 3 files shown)


tests/test-miss-handlers.lua
```diff
@@ -40,4 +40,8 @@ local w = wibox()
 w.foo = "bar"
 assert(w.foo == "bar")
 
+-- Test if read-only properties really are read-only
+screen[1].clients = 42
+assert(screen[1].clients ~= 42)
```
blueyed: Should there be a warning or even error in this case?
Something like `cannot set read-only attribute: clients`?

psychon: I could do that in the other cases, but in this case the C code "eats" all `error()`s and turns them into warnings. Should I do this anyway?

blueyed: Thanks!

psychon: I guess that means "please squash the fixup commits into the related commits". Ok. :-)

blueyed:
Yeah, please squash and then merge.

psychon:
Done. However, I first pushed awesome master and then the PR for this branch and now github turned this into the empty pull request... Whatever.

PR #948, opened 2016-06-04
Deprecate add_signal

psychon:
Based on some recent comment from @Elv13 (Sorry, I don't remember where exactly) and his work with `gears.object` allowing arbitrary properties to be changed and signaled (i.e.. `property::name`), it became clear that having to add signals before they can be used has outlived its usefulness. Catching typos is nice, but making it a lot harder to write code is bad. Thus, this PR deprecated `add_signal` and makes other functions automatically add signals when needed.

### DIFF
--- a/common/luaclass.h
+++ b/common/luaclass.h
@@ -123,7 +123,7 @@ luaA_checkudataornil(lua_State *L, int udx, lua_class_t *class)
     static inline int                                                          \
     luaA_##prefix##_class_add_signal(lua_State *L)                             \
     {                                                                          \
-        signal_add(&(lua_class).signals, luaL_checkstring(L, 1));              \
+        luaA_deprecate(L, "signal usage without add_signal()");                \
         return 0;                                                              \
     }                                                                          \
                                                                                \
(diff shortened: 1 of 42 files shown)


common/luaclass.h
```diff
@@ -123,7 +123,7 @@ luaA_checkudataornil(lua_State *L, int udx, lua_class_t *class)
     static inline int                                                          \
     luaA_##prefix##_class_add_signal(lua_State *L)                             \
     {                                                                          \
-        signal_add(&(lua_class).signals, luaL_checkstring(L, 1));              \
+        luaA_deprecate(L, "signal usage without add_signal()");                \
```
blueyed: The deprecation message looks strange.  How about `Use signals without add_signal()` or `Explicitly calling add_signal is not required anymore`?

psychon: The problem is the same one that we have with `awful.util.deprecate`: The implementation does "This function is deprecated and will be removed, see <the argument here>". That's why I'm using this weird wording.

Should I instead change `luaA_deprecate`?

blueyed: Oh, I've forgotten about that.
I think it's good then.

PR #949, opened 2016-06-05
try menu_* colors before regular normal/focus

### LINKED ISSUE #941: Add menubar customization options
In `lib/menubar/init.lua`, labels are hard-coded to use `theme.bg_normal` if they are not focused. Elsewhere in awesome, fg/bg colors are usually selected from several alternatives, falling back to the `_normal` variants as a last resort. For example `lib/awful/menu.lua` tries `menu_bg_normal` first.

I just ran into a situation where I wanted to set the menubar differently from the taskbar and currently that seems to be impossible. It would be simple (and consistent) to add alternatives e.g. `menu_bg_normal`.

silverhammermba:
This resolves #941. This lets users customize the menubar separately from the tasklist in the same way the popup menu is.

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -102,9 +102,9 @@ end
 -- @return item name, item background color, background image, item icon.
 local function label(o)
     if o.focused then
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!

PR #950, opened 2016-06-05
Fix disconnecting not connected signals

psychon:
When a function is disconnected from a signal ("disconnect_signal") that is not
actually connected to the function, two things happened:

1. The attempt to remove the function from the signal array didn't do anything
2. Unreferencing the function noticed that the function wasn't referenced

The second step printed a big, fat scary warning.

Actually, this has the possibility of causing errors. For example, in the
following code, awesome would wrongly unreference the function at the
disconnect_signal() call and might later still try to call it when the
"refresh" signal is emitted:

do
    local function f() end
    awesome.connect_signal("refresh", f)
    awesome.disconnect_signal("debug::error", f)
end

Fix this by making signal_disconnect() return a boolean value indicating if it
actually did something. All callers are fixed to use this value and only update
the reference counts if something was actually disconnected.

Fixes: https://github.com/awesomeWM/awesome/issues/814

### DIFF
--- a/common/luaclass.c
+++ b/common/luaclass.c
@@ -278,8 +278,8 @@ luaA_class_disconnect_signal_from_stack(lua_State *L, lua_class_t *lua_class,
 {
     luaA_checkfunction(L, ud);
     void *ref = (void *) lua_topointer(L, ud);
-    signal_disconnect(&lua_class->signals, name, ref);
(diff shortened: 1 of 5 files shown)


blueyed:
Needs to be rebased.

psychon:
Rebased on master

PR #955, opened 2016-06-06
mouse: Fix moving clients between screens

Elv13:
Reported on IRC. I am not sure why swap() is not enough, but the
old code removed before the mouse refactor did this, so apparently
it is necessary.

The fix has been reported to work by spyroboy on IRC, thanks!


syphoxy:
I've confirmed that this works.

psychon:
On why this is necessary: `c:swap` only modifies the order that `client.get()` / `awful.client.visible()` / `awful.client.tiled()` returns its elements when called with "no specific screen" (`s=nil`). It doesn't move things between screens.

For the patch: Hm... My complaint would also apply to the previous code, so I'll skip that and just say "fine with me".

PR #957, opened 2016-06-09
resize: Restore the fleur cursor and add a beautiful option

### LINKED ISSUE #956: Client move/resize cursor changes?
Previously, when you drag or resize the window, it was a human-readable cursor (arrows), instead now simply crosshair. What are the reasons for this change?

(It's just a question.)

Elv13:
Fixes #956


blueyed:
Thanks!

PR #959, opened 2016-06-11
doc: Add a Cairo/LGI introduction page

Elv13:
Try to make @funeralismatic a happy user

### DIFF
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -21,6 +21,7 @@ topics={
     '02-contributing.md',
     '03-declarative-layout.md',
     '04-new-widgets.md',
+    '16-using-cairo.md',
     '90-FAQ.md',
 }
 
--- a/docs/16-using-cairo.md
+++ b/docs/16-using-cairo.md
@@ -1,0 +1,163 @@
+# Using Cairo and LGI
+
+These days, Awesome interface is mostly based on a library called LGI. It allows
+to access C libraries such as GTK, GLib, Cairo, Pango, PangoCairo and RSVG using
+Lua code without having to write actual "glue" C code.
+
+This is done using the GObject-introspection framework. The main advantage is
+the time saved and large number of features exposed for free. The downside is
+the lack of proper Lua centric documentation and examples. Some examples can be
+found in [LGI's own documentation](https://github.com/pavouk/lgi/tree/master/docs),
+but this does not directly explain how to use a concrete API.
+Using other APIs requires some trial and error, and can be even impossible if
+the introspection data is missing or inaccurate.
+Using low-level APIs directly can easily cause crashes.
+It is the programmer's responsibility to properly check return and error
+values.
+
+## Using LGI in Awesome
+
+GObject and Gnome centric libraries tend to use the common C practice of
+emulating namespaces using underscores in function names.
+LGI exposes a proper namespace based API. For example, if the C function is:
+
+    cairo_image_surface_create()
+
+Then the LGI equivalent is:
+
+    lgi.cairo.ImageSurface.create()
+
+The same goes for enums:
+
+    CAIRO_FORMAT_ARGB32
+
+becomes:
+
+    lgi.cairo.Format.ARGB32
+
+LGI is also object oriented while the C API is function based. When those
+functions take the "class" "object", then this:
+
+    cairo_line_to(cr, x, y)
+
+can be expressed as:
+
+    cr:line_to(x, y)
+
+It is however important to note some inconsistencies. For example, Cairo is
+called `lgi.cairo` while GLib is called `lgi.GLib`. Figuring this out will
+require some experimenting. The best way to do this without actually
+reloading Awesome is to open the `lua` command in a terminal and use `print`:
+
+    print("This will print a table address:", require("lgi").cairo)
+    print("This will print an error:", require("lgi").Cairo)
+
+It is recommended to avoid using `require` always when using a function, but
+include the libraries at the top of your `rc.lua` or Lua module instead:
+
+    local cairo = require("lgi").cairo
+
+## The Cairo API
+
+Cairo is a 2D graphic library used by Awesome, Gnome and XFCE. It allows to
+e.g. paint paths on a `surface`.
+Awesome uses it internally and being able to call it directly is a powerful
+feature.
+
+The following concepts are necessary to be able to use Cairo:
+
+**Surface**:
+
+A surface is the area where the painting will be done. There are multiple types
+of surfaces including:
+
+ * Color images with transparency (`ARGB32`) or without (`RGB24`)
+ * Monochrome image surfaces with transparency (`A8`) or without (`A1`)
+ * SVG vectorial surfaces
+ * Native (XCB) surfaces
+ * Framebuffers and other less interesting ones (from an Awesome's point of
+   view)
+
+For more details see [Surfaces](https://cairographics.org/manual/cairo-Image-Surfaces.html).
+
+**Sources**:
+
+Sources are elements like colors, patterns or gradients. See `gears.color` for
+common sources.
+
+For more details see [Pattern](http://cairographics.org/manual/cairo-cairo-pattern-t.html).
+
+**Context and paths**:
+
+A context is the proxy between the program and the surface, and holds a path.
+Paths are something like a line, circle or rectangle, which may or
+may not be closed (a shape).
+
+All drawing operations on a surface are done via a context.
+The current path is extended until it is used and reset (see next section).
+Until then nothing will be drawn to the surface.  For example:
+
+    cr:rectangle(0 , 0 , 10, 10)
+    cr:rectangle(10, 10, 10, 10)
+
+will not do anything until the operation is applied to the context.
+
+For more details, read:
+
+* [Path](http://cairographics.org/manual/cairo-Paths.html)
+* [Context](http://cairographics.org/manual/cairo-cairo-t.html)
+* [Transformation](http://cairographics.org/manual/cairo-Transformations.html)
+
+A context also holds a transformation matrix (see `gears.matrix`), which is
+used when applying an operation.
+
+**Operations**:
+
+Multiple operations can be done with the paths. The most common are:
+
+ * *fill*: Fill the path with the current source.
+ * *stroke*: Paint the path outline with the current source.
+ * *mask*: Use the current source as an alpha mask while painting with the
+   current operator.
+ * *clip*: Crop the surface's workarea so nothing outside of the clip will be
+   affected by all following operations.
+
+**Operators**:
+
+[Operators](http://cairographics.org/operators/) are modifiers used when
+applying operations.
+
+### Cairo in Awesome
+
+All of Awesome's `wibox`es, `awful.wibar`s, `gears.wallpaper`s and
+`awful.titlebar`s contain a Cairo surfaces, which can be accessed through the
+`drawin` API.  This allows widgets to use the Cairo context directly.
+See the
+[declarative layout system](../documentation/03-declarative-layout.md.html)
+and [new widgets](../documentation/04-new-widgets.md.html) articles for more
+information and examples on how widgets work.
+
+It is also possible to create surfaces manually. See `gears.surface` for
+some examples. Here is the most simple example you can get:
+
+    -- Create a surface
+    local img = cairo.ImageSurface.create(cairo.Format.ARGB32, 50, 50)
+
+    -- Create a context
+    local cr  = cairo.Context(img)
+
+    -- Set a red source
+    cr:set_source(1, 0, 0)
+    -- Alternative:
+    cr:set_source(gears.color("#ff0000"))
+
+    -- Add a 10px square path to the context at x=10, y=10
+    cr:rectangle(10, 10, 10, 10)
+
+    -- Actually draw the rectangle on img
+    cr:fill()
+
+This can then be used as `bgimage` for a `wibox`, `awful.wibar` or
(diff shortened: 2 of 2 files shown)


docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
+# Using Cairo and LGI
+
+These days, Awesome interface is mostly based on a library called LGI. It allows
+to access C libraries such as GTK, GLib, Cairo, Pango, PangoCairo and RSVG using
+Lua code without having to write actual "glue" C code.
+
+This is done using the GObject-introspection framework. The main advantage is
+the time saved and large number of features exposed for free. The downside is
+the lack of proper Lua contric documentation and examples. Some examples can be
+found in [LGI own documentation](https://github.com/pavouk/lgi/tree/master/docs)
```
psychon: LGI -> LGI's

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+, but this doesn't even cover 1% of the actual API. Using other APIs require
```
psychon: I think the comma is mis-misplaced. I'm not 100% sure, but doesn't this mean there will be a space before the comma?
Put differently: The comma should be at the end of the previous line.

psychon: Also, the 'not even 1%' might be a bit... colloquial. How about:

but this doesn't directly explain how to use a concrete API.

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+some trial and errors. It can also prove impossible if the introspection data
+is missing in inacurate. Using those low level APIs directly is also crash
+prone. It is the programmer responsability to properly check return and errors
```
psychon: "Those" in "those low level APIs" is an unclear reference. I would suggest "Using low-level APIs directly can easily cause crashes."

"return and errors values": errors -> error

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+values.
+
+## Using LGI in Awesome
+
+GObject and Gnome centric libraries tend to use the common C practive of
+emulating namespaces using underscores in funciton name. LGI expose a proper
```
psychon: expose -> exposes

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+namespace based API. For example, if the C function is:
+
+    cairo_image_surface_create()
+
+Then LGI equivalent is:
```
psychon: Then *the* LGI equivalent is

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+
+    lgi.cairo.ImageSurface.create()
+
+The same goes for enums:
+
+    CAIRO_FORMAT_ARGB32
+
+become:
```
psychon: become*s*

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+
+It is recommanded to avoid using `require` everytime you use a function and
+instead include the libraries at the top of your `rc.lua` or lua module
+
+    local cairo = require("lgi").cairo
+
+## The Cairo API
+
+Cairo is a 2D graphic library used by Awesome, Gnome and XFCE. It allows to
+paint paths on a `surface`. Awesome use it internally and being able to call
```
psychon: "paint paths" might be a bit narrow. Cairo can do much more, but ok.

"use it internally" -> "uses it internally"

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+is directly should be considered a power user feature.
```
psychon: is -> it

(Is it "power user" or "power-user"?)

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+
+Understanding these few important concepts is necessary to be able to use Cairo:
+
+**Surface**:
+
+A surface is the area where the painting will be done. There is multiple types
```
psychon: There is -> There are

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+of surfaces including:
+
+ * Color pixmap with transparency (`ARGB32`) or without (`RGB24`)
```
psychon: Can we call these images instead of pixmaps? That's what cairo actually calls them. Also, pixmap in an X11-context has a very specific meaning that is different from this one.
(Same for the following lines)

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+common sources.
+
+For more details, read:
+
+* [Pattern](http://cairographics.org/manual/cairo-cairo-pattern-t.html)
+
+**Context and paths**:
+
+A context is the proxy between the program and the surface. The context hold a
+set of paths. Paths are something like a line, circle or rectangle. They may or
```
psychon: Technically (in cairo's term), it's a single path and not a set of paths...

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+may not be closed (a shape). All operations on the surfaces has to be told to
+the context. Until the context state is flushed (see next section), those paths
```
psychon: "All drawing operations on a surface are done via a context"
(I added "drawing" because there are operations that can be done on a surface; "has to be told to the context" sounds weird to me)

"Until the context state is flushed" sounds weird to me as well. How about "The current path is extended until it is used and reset (see next section)"?

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+add up, until that, nothing will be written to the surface. For example:
```
psychon: "." instead of "," between here? ("add up. Until that")
Also: that -> then
Also: I suggest written -> drawn

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+
+    cr:rectangle(0 , 0 , 10, 10)
+    cr:rectangle(10, 10, 10, 10)
+
+wont do anything until and operation is applied on the context.
```
psychon: What does "and operation is applied on the context" mean? I don't understand.

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+
+For more details, read:
+
+* [Path](http://cairographics.org/manual/cairo-Paths.html)
+* [Context](http://cairographics.org/manual/cairo-cairo-t.html)
+* [Transformation](http://cairographics.org/manual/cairo-Transformations.html)
+
+A context also hold a transformation matrix (see `gears.matrix`) used when
```
psychon: hold -> holds

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+applying an operation.
+
+**Operations**:
+
+Multiple operations can be done with the paths. The most common are:
+
+ * *fill*: Fill the path with the current source
+ * *stroke*: Paint the path outline with the current source
```
psychon: Perhaps point at the line width for stroking? I'm not sure, the line might become weirdly longer than the other lines.

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+ * *mask*: Use the current source as an alpha mask while painting with the
+   current operator.
+ * *clip*: Crop the surfaces workarea so nothing outside of the clip will be
+   affected by the next opetations.
```
psychon: Typo in "operations"
Instead of "next" I suggest to use "all following"

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+
+**Operators**:
+
+[Operators](http://cairographics.org/operators/) are modifiers used when
+applying operations.
+
+### Cairo in Awesome
+
+Each Awsome `wibox`es, `awful.wibar`s, `gears.wallpaper`s and `awful.titlebar`s
```
psychon: Awsome -> Aw*e*some
Instead of "Each" I suggest "All `wibox`es, `awful.wibar`s..."

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+are in fact glorified Cairo surfaces. They can be used as such using an hidden
```
psychon: Technically they are not, they contain one... but whatever...

an hidden -> a hidden
(Is it really hidden? The API docs contain it)

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+API called `drawin`. Howver, they let widgets use the Cairo context directly.
```
psychon: Howver -> However

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+See the
+[declarative layout system](../documentation/03-declarative-layout.md.html)
+and [new widgets](../documentation/04-new-widgets.md.html) articles for more
+information and examples.
```
psychon: for more information and examples on how widgets work / are implemented / something like this

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+some examples. Here is the most simple example you can get:
+
+    -- Create a surface
+    local img = cairo.ImageSurface.create(cairo.Format.ARGB32, 50, 50)
+
+    -- Create a context
+    local cr  = cairo.Context(img)
+
+    -- Set a red source
+    cr:set_source(gears.color("#ff0000"))
```
psychon: Since you mention "most simple": Can you make this `cr:set_source_rgb(1, 0, 0)`?

docs/06-using-cairo.md
```diff
@@ -0,0 +1,160 @@
...
+
+    -- Add a 10px square path to the context at x=10, y=10
+    cr:rectangle(10, 10, 10, 10)
+
+    -- Fill the path and write it to `img`
+    cr:fill()
```
psychon: This comment seems weird. At first I didn't understand what "write it to `img`" refers to. How about:
"Actually draw the rectangle on `img`"

psychon:
Thanks, this sounds good. Sadly I left a metric ton of comments. Sorry!

blueyed:
Really nice.

I'm going through this as well right now, and will provide a fixup commit - including @psychon's changes.

docs/06-using-cairo.md
blueyed: @psychon 
Seems to refer to "Until the context is flushed"?!

I have changed it to the following now:

```
All drawing operations on a surface are done via a context.
The current path is extended until it is used and reset (see next section).
Until then nothing will be drawn to the surface.  For example:

    cr:rectangle(0 , 0 , 10, 10)
    cr:rectangle(10, 10, 10, 10)

will not do anything until the operation is applied to the context.
```

Still unclear though / the same for that part.

Is my change good as-is?  What do you suggest for the line in question?

docs/06-using-cairo.md
blueyed: @psychon 
I don't understand your comment.  Is it about adding a reference to the `stroke` bullet point entry?

psychon: * *stroke*: Paint the path outline with the current source and line width

or something like that

docs/06-using-cairo.md
psychon: instead of "the operation", how about "a drawing operation" or "a drawing operation using a path"?
And instead of "is applied to the context", how about "is done"?

docs/06-using-cairo.md
blueyed: I'll keep the other as an alternative, I think it is good to have a reference to `gears.color` there.

blueyed:
@psychon 
Here are some helper functions to checkout a PR from Github, in case you want to do inline edits the next time for something like that: https://github.com/blueyed/oh-my-zsh/commit/a8e38c2de3280e302da77c67b7cde0cea0b3da5d

docs/config.ld
```diff
@@ -21,6 +21,7 @@ topics={
     '02-contributing.md',
     '03-declarative-layout.md',
     '04-new-widgets.md',
+    '06-using-cairo.md',
     '90-FAQ.md',
```
blueyed: We should have this more spread out (to allow for new pages inbetween), since it appears to be used in links (which should be stable).

Elv13:
> Thanks, this sounds good. Sadly I left a metric ton of comments. Sorry!

Sorry to have learned English chatting in MMORGPs

@blueyed Thanks for the fixup, I will squash it and rename the file (I skipped 05 because I wanted a spot for another widget related page, but I should leave many more)

So, if we are to come up with a convention, what about:

1-100: Widgets and painting
101-200: Window management
201-300: Development process
301+: User contribution and external wiki pages (assuming we use ldoc for the wiki)

blueyed:
I've thought more about keeping it at 2 digits (and the FAQ) where it is, but using 3 digits right away is also good I suppose?!

psychon:
> Sorry to have learned English chatting in MMORGPs

No problem, but that's really a nice way to learn English. :-)

I don't have much of an opinion about the number of digits. Feel free to go with three digits.

@blueyed I thought you didn't like coveralls for leaving comments like codecov just did? :-)

blueyed:
@psychon 
Codecov will update an existing comment AFAIK.

blueyed:
AFAIK the numbering is only required as a hack for luadoc, right?

@Elv13 
Feel free to merge this.

PR #960, opened 2016-06-11
.travis.yml: Add codecov support

### LINKED ISSUE #952: Switch to codecov
Based on https://github.com/awesomeWM/awesome/pull/936#issuecomment-223718847, I assume that @blueyed wants to switch us from coveralls.io to codecov.io (and I have to say there web page looks way more awesome). And apparently they even have working Lua support out of the box!
I'm opening this issue for tracking that and so that we have a place for discussion.

So:
- Anyone against this?
- @blueyed Do you want to do this? Should I look into it?

psychon:
Fixes #952

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 language: c
 env:
   matrix:
-    - LUA=5.2 LUANAME=lua5.2 BUILD_APIDOC=true DO_COVERAGE=true
+    - LUA=5.2 LUANAME=lua5.2 BUILD_APIDOC=true DO_COVERAGE=coveralls
     # Note: luarocks does not work with Lua 5.0.
     - LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build
     # luajit: installed from source.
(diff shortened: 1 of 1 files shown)


psychon:
Coveralls result: https://coveralls.io/jobs/15323082
Codecov result: https://codecov.io/github/awesomeWM/awesome/commit/602a24cdbd84f702a4472f87411b993755696b13
Note that coveralls seems to be done immediately when the build finishes while codecov still needs some minutes to "crunch the data".

.travis.yml
```diff
@@ -83,12 +83,23 @@ install:
...
 
   # Determine custom version.
   - export AWESOME_VERSION="${TRAVIS_BRANCH}-g$(git rev-parse --short HEAD)"
   - 'if [ "$TRAVIS_PULL_REQUEST" != false ]; then AWESOME_VERSION="${AWESOME_VERSION}-PR${TRAVIS_PULL_REQUEST}"; fi'
+  # function for codecov support
+  - |
+    do_codecov() {
+      if [ "$DO_COVERAGE" = "codecov" ]; then
+        (cd build && luacov) || exit 1
+        bash <(curl -s https://codecov.io/bash) -F "$1" || exit 1
```
blueyed: Add `travis_retry` here?
Maybe also fetch the script only once into `/tmp`, and then use it from there?!

.travis.yml
```diff
@@ -102,10 +113,16 @@ script:
   - make && sudo env PATH=$PATH make install && awesome --version
   - |
     if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
-      if [ "$DO_COVERAGE" = "true" ]; then
+      do_codecov samples
+    fi
+  - |
+    if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
+      if [ "$DO_COVERAGE" != "0" ]; then
```
blueyed: You could test for `-n "$DO_COVERAGE` here, and then just not export it as `0` by default above.

psychon: Well, the 0 value is needed by cmake, AFAIK. I didn't add the `export DO_COVERAGE=0`, it was already there before (needed for the code for `tests/samples` in CMake).

.travis.yml
psychon: I don't think it matters much, but whatever. Added as a fixup commit.

blueyed: Thanks!

blueyed:
:+1:

PR #961, opened 2016-06-11
.travis.yml: Stop installing rxvt-unicode

psychon:
The last place where we used urxvt for testing was removed in 6b4a2625cb1.


blueyed:
Thanks!

I've added `[ci skip]` to the squash-merge's commit message to not trigger another CI run on master for this.

PR #963, opened 2016-06-12
tag: Fix restoring invalid tags from history

Elv13:
If:

 1) An app is open in a tag
 2) A new tag is created
 3) The app is closed in the first tag
 4) The first tag is deleted
 5) The new tag is deleted

Then the history would try to restore an inactive tag without
a screen. Bad things will then happen.


psychon:
I'm trying to come up with a more general fix for this that doesn't require touching all code that touches tags... but I have no good idea. Apparently deleting tags never worked correctly. Perhaps some (more?) callbacks on e.g. `property::screen` should check `.activated` and do nothing for deactivated tags?

Anyway, :+1: for this patch.

Issue #962, opened 2016-06-11
CMake configure_file should be disabled for the CodeCov build 


Elv13:
This way, the line number will be correct again. Until that, the coverage report is just gamification, it's technically useless when there is expanded documentation. And yes, I just took a look, CodeCov is Awesome!

Also, it seems, at least from the look of the front page, that the test code is included in the coverage report.

I may do that eventually, but I wont have much time for Awesome in the coming months.

psychon:
The line numbers were already off for coveralls and apparently it wasn't a problem there...? This part of the problem isn't new, but I agree that fixing it would be nice.

For the test code: Yup, it's included and I don't understand why it is not included in coveralls. This can be fixed with a suitable exclude rule. I'll handle that.

PR #964, opened 2016-06-12
Exclude tests/ on codecov

psychon:
I _think_ that this fixes the part of #962 that is not part of the title, namely that codecov also showed code coverage for the tests runners. Let's see what codecov says after this ran...


psychon:
Hm.. this seems to break coverage for the "examples tests". `gears.shape` is completely uncovered in both coveralls and codecov.

psychon:
Heh, I guess that this found the problem that #982 tracks before the new luacov release brought this problem to master. :-)

blueyed:
This needs to be rebased.

blueyed:
Closing this as per the discussion in https://github.com/awesomeWM/awesome/pull/1239#issuecomment-270336596: it appears to be a workaround only, and I think it is better to track coverage of tests also instead.

PR #970, opened 2016-06-18
Correct name variable

salorium:


psychon:
Could you mention in the commit message that this typo was introduced in 03e2310c? Thanks!

How did you find this problem?

salorium:
When i update my awesome. I use a scroll for display mpd and i have a error message.

PR #972, opened 2016-06-20
prompt: Fix hooks when number or caps lock are set

Elv13:
I don't know why I never noticed this. Prompt hooks are not working when number lock was set.

### DIFF
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -22,6 +22,7 @@ local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility
 local keygrabber = require("awful.keygrabber")
 local util = require("awful.util")
 local beautiful = require("beautiful")
+local akey = require("awful.key")
 
 local prompt = {}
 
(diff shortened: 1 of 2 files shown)


lib/awful/prompt.lua
```diff
@@ -362,13 +363,24 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
...
+            -- Remove caps and num lock
+            local mods_mask_inv, filtered_modifiers = {}, {}
+            for _, m in ipairs(akey.ignore_modifiers) do
+                mods_mask_inv[m] = true
+            end
+            for _, m  in ipairs(modifiers) do
+                if not mods_mask_inv[m] then
+                    table.insert(filtered_modifiers, m)
+                end
+            end
```
psychon: The above can be simplified, I think:
```lua
for _, m in ipairs(modifiers) do
    if not awful.util.table.hasitem(akey.ignore_modifiers, m) then
        table.insert(filtered_modifiers, m)
    end
end
```

Elv13:
> The above can be simplified, I think:

At first I did, however I prefer not using `util.table` for this kind of tasks, one is O(N) and the other O(N^2). In "reality", it wont change anything, but the little early optimizer evil in me say to keep it the way it is.

psychon:
Yeah, because iterating over a table which by default has two entries... well, whatever. Ok.

blueyed:
I would prefer the simplified version suggested by @psychon.
@Elv13 
Can you change it, please?
Otherwise I could do it also when merging - let me/us know.

Elv13:
I wont have time for awesome until next friday, sorry

blueyed:
Thanks for letting us know.
I've amended the commit, but unfortunately have no Num Lock and my Caps Lock is remapped, so I cannot test this myself easily.

Elv13:
@blueyed Thanks! I fixed a little problem where you added an `awful.util` call while `awful` wasn't imported. Beside that, your changes worked.

Issue #982, opened 2016-07-04
Travis is not very stable as of late.


Elv13:
Some build passes, some build fails. I didn't investigate much, but I have seen the "build::passed" and "build::failed" change state for no apparent reasons.

```
-- All test passed!

-- Running tests again with coverage

-- Running tests/examples/awful/mouse/coords.lua...

Error: couldn't run config file /home/travis/build/awesomeWM/awesome/.luacov: line 4: $BUILD_DIRECTORY not set

CMake Error at tests/examples/CMakeLists.txt:162 (message):

  /home/travis/build/awesomeWM/awesome/tests/examples/awful/mouse/coords.lua

  A test failed, bye

Call Stack (most recent call first):

  tests/examples/CMakeLists.txt:238 (run_test)

  awesomeConfig.cmake:359 (include)

  CMakeLists.txt:17 (include)

```

I wont have time this week, nor the next. I should have some time the week after that, but if someone want to step in, it is quite obvious which commit broke this (and I have a big part of the partial blame for it).

blueyed:
> it is quite obvious which commit broke this

Do you have a reference / more information regarding this?

psychon:
I don't think I saw `$BUILD_DIRECTORY not set` before, but two builds in https://travis-ci.org/awesomeWM/awesome/builds/142325384 failed due to this (and another one failed because `Error: timeout waiting for signal in step 4/4 (@5).` in `test-screen-changes.lua`; let's see if restarting the builds fixes things there).

I'm not sure where this was introduced either.

psychon:
It just occurred to me: "Not very stable" is wrong. Both builds with coverage information consistently fail, so they *are* stable. The only unstable build is (once again) the LuaJIT build where two functional tests randomly fail.

psychon:
A fix for the part of this that is caused by the new luacov release is in #964, but afterwards coverage for `tests/examples/` seems to be missing. Perhaps I got some path wrong, or something like that?

psychon:
I give up. Someone else can look into this. I just don't understand Travis.

Before https://github.com/psychon/awesome/commit/a39f5ff71e68fd705072f0763adb6900026ea6ec, the travis build output showed that the `$SOURCE_DIRECTORY` environment variable was passed on into `.luacov`: https://travis-ci.org/psychon/awesome/builds/143392915 Relevant output is
```
[...]

PR #983, opened 2016-07-04
fix(themes: xresources): explicit bg and fg for tooltip

actionless:
should address https://github.com/awesomeWM/awesome/pull/929#issuecomment-229145404

### DIFF
--- a/themes/xresources/theme.lua
+++ b/themes/xresources/theme.lua
@@ -42,6 +42,9 @@ theme.border_marked = xrdb.color10
 -- Example:
 --theme.taglist_bg_focus = "#ff0000"
 
+theme.tooltip_fg = theme.fg_normal
+theme.tooltip_bg = theme.bg_normal
+
 -- Variables set for theming the menu:
 -- menu_[bg|fg]_[normal|focus]
 -- menu_[border_color|border_width]


psychon:
The travis failure seems unrelated (but annoying), so: :+1: 
I think @blueyed "started this" and gets to decide if this is the right fix.

blueyed:
LGTM, given that the default was changed (accidentally(?) - but I think it makes sense).

blueyed:
@actionless 
You were just added to the team (after speaking to @psychon and @Elv13).. :boom: 

Please (squash-)merge it yourself.. :)

actionless:
thanks :)

i rebased it on master, but those two travis setups are still failing

psychon:
Ignore that, it's not yet fixed. It's tracked in #982 and worked around by #987 (which wasn't merged yet and I don't know if it should be merged). The problem is that all code-coverage builds suddenly fail.

actionless:
then it's probably a permission problem, i first can't merge because of the failing travis

psychon:
Heh. Thanks GitHub. You set us up a so-called "Team". Membership in which gains you... no access rights.

@actionless Can you try again? You should now be able to push to awesome (and awesome-www, but that's not really important, I guess)

Issue #969, opened 2016-06-18
undefined reference to symbol 'round@@GLIBC_2.2.5


nasedo:
When trying to build awesome (GCC 6.1.1) from git I'm getting the following errors:
`/usr/bin/ld: CMakeFiles/awesome.dir/mouse.c.o: undefined reference to symbol 'round@@GLIBC_2.2.5'`

Reverting commits 9991f9cc and 5ebe073c makes the build possible but then when running a red message window often appears:
`(integer expected, got number)`

psychon:
How are you building awesome? Anything special? What system are you on (Which distro, any special packages, ...)? Could you do a fresh build, but with `make V=1` so that all the arguments that are passed to the compiler are shown (and of course provide the resulting output somehow)?

nasedo:
Weird, apparently this happens only if linking to luajit. I'm on Arch Linux and this [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=awesome-luajit-git) describes the build process. It boils down to [this](https://aur.archlinux.org/cgit/aur.git/tree/0001-add-luajit.patch?h=awesome-luajit-git) patch and:
`cmake -DCMAKE_INSTALL_PREFIX=/usr -DSYSCONFDIR=/etc -DLUA_LIBRARIES=/usr/lib/libluajit-5.1.so`
[Here](http://ix.io/Ubi) is `make VERBOSE=1` output.

psychon:
Hm, seems like `FindLua51` does indeed make us link against `libm`. For now I'd say that the `PKGBUILD` from AUR is broken since it makes awesome fail to build. However, we really should check for `libm` ourselves if we use it (and of course this isn't actually easy with CMake... why are we using CMake again?).

PR #984, opened 2016-07-05
CMake: Check if we need libm for round()

psychon:
Fixes: https://github.com/awesomeWM/awesome/issues/969

P.S.: Who said that CMake is easier than autotools? Can we switch to meson instead? :-(

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -209,6 +209,20 @@ else()
     message(STATUS "checking for __builtin_clz -- no")
 endif()
 
+# Do we need libm for round()?
+check_function_exists(round HAS_ROUND_WITHOUT_LIBM)
+if(NOT HAS_ROUND_WITHOUT_LIBM)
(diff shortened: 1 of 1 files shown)


blueyed:
> Who said that CMake is easier than autotools?

Yeah, it often looks similar, if the magic fails.

> Can we switch to meson instead?

Your call.. :)

Elv13:
I still prefer CMake to othre obscure alternative (and I hate autotools). While I agree that when the magic fails, your on your own, I still argue that CMake is the most popular system right now and have plenty of third party modules and support.

blueyed:
@psychon 
You were right: it conflicts..  please feel free to cherry-pick into master directly.

psychon:
Done (and once again in a way so that GitHub doesn't notice -- sorry!)

blueyed:
You would have to add "Closes: …" to it.

blueyed:
Thanks! :)

psychon:
...or I would have to push to the branch of the PR before pushing to master.

PR #987, opened 2016-07-05
Dirty hack to make travis build again

psychon:
I'm not really sure what is going on here... the coverage builds on travis set
DO_COVERAGE and thus our CMake machinery tries to re-run the sample-tests with
coverage information. However, $BUILD_DIRECTORY is not set and our .luacov file
bails out. I do not know how this ever worked. Did it ever work?

This half-address #982 by just disabling the broken code. However, a better fix would be to actually fix this instead.

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -352,7 +352,7 @@ endforeach()
 # The examples coverage need to be done again after the configure_file has
 # inserted the additional code. Otherwise, the result will be off, rendering
 # the coverage useless as a tool to track untested code.
-if(GENERATE_DOC AND DO_COVERAGE)
+if(GENERATE_DOC AND DO_COVERAGE AND FALSE) # FIXME: This is broken and causes build failures on travis
     message(STATUS "Running tests again with coverage")
     set(USE_LCOV 1)
 


blueyed:
Too dirty for me..

I don't know what `BUILD_DIRECTORY` is used for (CMake?), but it looks like we need to set it in a similar way like it's done for `check-coverage` (https://github.com/awesomeWM/awesome/blob/0c909e8d315dc2f2c80652c1f09f56df465a0433/CMakeLists.txt#L381) when running the `.luacov` file (https://github.com/awesomeWM/awesome/blob/0c909e8d315dc2f2c80652c1f09f56df465a0433/tests/examples/CMakeLists.txt#L152-L156)?!

psychon:
My problem is: How did this ever work and when was it broken? I don't really want to try to "randomly fix it", because there should be something identifiable that broke it...

`BUILD_DIRECTORY` is only used in `.luacov` and it is used for three things:
- Filename of the file that coverage results are saved to
- Pattern specifying for which code should coverage be collected (e.g. we don't care about code coverage inside of lgi)
- coveralls' pathcorrect feature (Replace the prefix `$BUILD_DIR/lib` with `lib/` in the gathered coverage data; codecov doesn't need this)

CMake doesn't care about it.

blueyed:
See https://travis-ci.org/awesomeWM/awesome/builds.
2122 seems to be the first failing in that way: https://travis-ci.org/awesomeWM/awesome/builds/142278594 - but the diff is small and seems unrelated?!

psychon:
Well... let's try what happens if we restart the build before that (= build 2121): https://travis-ci.org/awesomeWM/awesome/builds/140592536
Now it failed with the same error.

So... something outside of awesome must have changed. How about a new luacov release? https://luarocks.org/modules/mpeterv/luacov

psychon:
https://github.com/keplerproject/luacov/commit/3ec7bcfa595c22d55b564da9213affc9aee78b0e

We load luacov inside a `pcall()`. Before the above commit, the pcall caught the error that we are seeing now. After the commit we get a call to `os.exit()` which `pcall()` does not protect against.

blueyed:
Good investigation!
What do you suggest doing now, given that we know what the issue is?  (and that it did not work before, but was silent)

blueyed:
How about pinning luacov and other deps to prevent such mysteries in the future?

Issue #986, opened 2016-07-05
wibox.widget.systray: missing application icons


Jajauma:
The oddball problem I came across lately is that inherent systray widget doesn't show some application icons (transmission-gtk, pasystray) while others (nm-applet, keepassx) are shown just fine. Here is a screenshot with interrogation marks I put where applications icons are supposed to be drawn. The empty areas are clickable with mouse, producing respective application menus or other actions.
![awm-systray-icons-interrogation](https://cloud.githubusercontent.com/assets/11892063/16597694/55c71d8a-430b-11e6-834f-fab250da0533.png)
Compare with an external implementation of the systray protocol (http://stalonetray.sourceforge.net/), all four applications icons are sheer working.
![stalonetray-icons](https://cloud.githubusercontent.com/assets/11892063/16597848/0b3cbbde-430c-11e6-90d2-93e63e649ac5.png)
I wonder if the issue could be connected to our incurring icon frenzy https://github.com/awesomeWM/awesome/issues/908 somehow.

blueyed:
Is this on master?
`pasystray` and others work for me.

But I've also seen a missing icon once in a while lately.

Jajauma:
Yes, it is git master at 441587a on the pictures.

psychon:
Could you gather some information about the missing icons?

Start with `xwininfo -tree` and click on the wibox. This should list the "Awesome systray window" and its children. Show the output of this. For each direct child of this window (especially for the missing icons) I'd be interested in the output of `xwininfo -id 0x42` and `xprop -id 0x42` where `0x42` is the ID that you got in the previous step.

Jajauma:
OK, here is the gist https://gist.github.com/Jajauma/0bde81381a8169f86bd62267103e7b82.

Jajauma:
Apparently the number of systray applications and their launch order does matter. If I kill nm-applet, the tray is redrawn and transmission icon with a gear handle is left lonely sitting in the shrinked systray as if nothing had happened.

psychon:
Thanks. I can't spot any difference between them and the values that I was interested in (`Map State: IsViewable` and `_XEMBED_INFO = 0x1, 0x1`) both look good.
However, I noticed that for transmission (which seems to be the missing icon), the size is below its minimum size. Width/Height is 21, `WM_NORMAL_HINTS` specifies a minimum size of `24x24`.

[...]

PR #990, opened 2016-07-07
Correctly deny ConfigureRequests for embedded windows

psychon:
Since commit 4daa6e8, we are denying resizes and moves for embedded windows
(=tray icons). However, the Xembed spec says that the embedding client acts like
a WM (as specified by ICCCM) to the embedded window. Thus, when denying a
configure request, we have to inform the window by sending it a synthetic
configure notify. Otherwise, GTK seems to sometimes not draw its tray icon.

Fixes: https://github.com/awesomeWM/awesome/issues/986

### DIFF
--- a/event.c
+++ b/event.c
@@ -403,7 +403,31 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
     {
         /* Ignore this so that systray icons cannot resize themselves.
          * We decide their size!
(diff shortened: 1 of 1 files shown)


blueyed:
:+1: 
Thanks!
Merging as it really seems to work according to https://github.com/awesomeWM/awesome/issues/986#issuecomment-231181495.

PR #991, opened 2016-07-08
Travis: Install luacov 0.11.0-1

psychon:
With luacov 0.12.0-1 our travis build fails. It was always kinda broken, but now
this breakage is revealed.

What likely happens: Our .luacov errors out because some needed environment
variable to figure out where to put the results is not set. Due to the way we
load luacov, it runs with its default config which luckily happens to work. With
luacov 0.12.0-1 the error is propagated instead and the build fails.


blueyed:
Thanks!

PR #993, opened 2016-07-09
tests/examples: Move common template code into a common file

psychon:
This is just the first commit from #964 as a separate PR. The other changes in there are "oh, I discovered #982 before it was reported/understood and got confused".

### DIFF
--- a/tests/examples/awful/template.lua
+++ b/tests/examples/awful/template.lua
@@ -1,29 +1,9 @@
-local file_path, image_path, luacovpath = ...
+local file_path, image_path = ...
+require("_common_template")(...)
 
 local cairo      = require("lgi").cairo
 local pango      = require("lgi").Pango
 local pangocairo = require("lgi").PangoCairo
-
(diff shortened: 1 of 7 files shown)


blueyed:
The codecov/patch output is weird: it says that lines are not covered, but says so for lines below?!
https://codecov.io/gh/awesomeWM/awesome/compare/4cc6a8155f311fe684e285a5466cba0a9c66e3e2...7410a53568589d74b05eeca9b1732167940c9ddf

psychon:
Well, it is right. Luacov is loaded by `_common_template` and everything that is run before that will not show up in the coverage report. However, we don't want to track code coverage for the test templates anyway. The only reason this shows up is that this is not using our `.luacov` / all of this is broken.

What amazes me more are the 259 new hits and the 188 less misses. However, I'll just go with "ignore the broken coverage output" until it is fixed.

psychon:
@blueyed Thanks for handling all these PRs :-)

Issue #387, opened 2015-08-02
theme/titlebar: missing icon for minimize button


blueyed:
Via https://github.com/awesomeWM/awesome/pull/369#issuecomment-126942662, from @actionless:

> also it seems like we are missing an icon for minimize: http://awesome.naquadah.org/doc/api/modules/awful.titlebar.html#titlebar.widget.minimizebutton

actionless:
after you'll merge that PR i can add it

sixshotx:
The latest titlebar.lua in lib/awful directory might still have the problem.
` local widget = titlebar.widget.button(c, "minimize", function() return c.minimized end, function(c) c.minimized = not c.minimized end)`

parameter "c" is missing in the first function().

PR #995, opened 2016-07-09
Fix the titlebar minimize button

psychon:
Commit 0318c6132840413b1f added an image for the minimize button which was
missing before. However, only the default theme and xresources where fixed to
contain the path to the image.

This commit adds the path to all themes.

Also, minimized clients aren't visible, just as closed (=killed) clients aren't
visible. Thus, we don't need an "active" version of this image.

This commit makes us handle the image for the minimize button just like we
handle the close button: There is no difference between "active" and "inactive"
and the file path in the theme doesn't get any path suffix.

Fixes: https://github.com/awesomeWM/awesome/issues/387


actionless:
LGTM

PR #996, opened 2016-07-09
Print LuaJIT version in --version

psychon:
When called with "--version", awesome prints the Lua version that it was
compiled against and the one that it is running against. This commit makes the
code detect LuaJIT and makes it print the LuaJIT version instead of an
unspecific "Lua 5.1".

### DIFF
--- a/common/version.c
+++ b/common/version.c
@@ -36,8 +36,14 @@ void
 eprint_version(void)
 {
     lua_State *L = luaL_newstate();
-    luaopen_base(L);
+    luaL_openlibs(L);
     lua_getglobal(L, "_VERSION");
+    lua_getglobal(L, "jit");
+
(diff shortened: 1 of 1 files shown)


psychon:
TODO: If this is merged, I think it should be cherry-picked to the 3.5 branch

blueyed:
Works: https://travis-ci.org/awesomeWM/awesome/jobs/143534777#L2920

psychon:
Almost more important is: https://travis-ci.org/awesomeWM/awesome/jobs/143534778#L2756
The first version of this patch worked on LuaJIT, but broke normal Lua (it would just print `(null)`). ;-)

blueyed:
@psychon 
Tried to cherry-pick it into "3.5", but it conflicts - I think you know better how to resolve it.. :)

Issue #184, opened 2015-03-30
Wrong usage of io.popen?!


blueyed:
I've noticed that there are a few places, where `io.popen` gets used like this:

    local f, err = io.popen(cmd, 'r')
    if f then
        local s = f:read("*all")
        f:close()
        return s
    else
        return err
    end

([source](https://github.com/awesomeWM/awesome/blob/master/lib/awful/util.lua.in#L106))

But `io.popen` always only returns a file handle, as far as I can see, also for `io.popen("DOESNOTEXIST")`.

Am I missing something?  Will the second return value be used in case of more serious errors?

Elv13:
Popen should be banned, glib provide an async version that wont block

blueyed:
@Elv13 
Can you provide a PR for this?  Shouldn't be too many places IIRC.

actionless:
if `spawn.pread` will became async it will be no difference between it and `spawn.easy_async`

btw tonight i've updated `menubar` to use async spawn, will submit PR tomorrow during the day after i will update docstrings accordingly

psychon:
You mean async spawn for `find lots of arguments here`? Can't we use something sane for that instead? Assuming that `g_dir_open()` and friends are available through lgi, that can be used.
https://developer.gnome.org/glib/stable/glib-File-Utilities.html

actionless:
are planning to do anything on that? i still haven't deleted my branch with async refactoring for those menubar helpers

psychon:
What exactly do you mean? Am I planning to make "something for iterating over directories" work? If so, then I have just opened https://github.com/pavouk/lgi/issues/135 which contains some hints on how to do this.

actionless:
https://github.com/actionless/awesome/commit/7db9b39aef85f7a3c8775dadfb1637a6f8d882b4

psychon:
I'd just replace "spawn `find`" with using `lgi.GLib.Dir` to iterate over directories. This will likely have more overhead and be slower, but not noticeably much. That's be even simpler code.

psychon:
Ok, apparently `GDir` is the wrong API to use. The following can be put into a file and run as Lua code to list the contents of `/tmp` and supposedly this is even async:
```lua
local lgi = require("lgi")
local glib = lgi.GLib
local gio = lgi.Gio
local main_loop = glib.MainLoop()

local path = "/tmp"

gio.Async.start(function()
	local f = gio.File.new_for_path("/tmp")
[...]

PR #998, opened 2016-07-09
Remove awful.{spawn.,}pread (#184)

psychon:

### DIFF
--- a/lib/awful/init.lua
+++ b/lib/awful/init.lua
@@ -29,9 +29,9 @@ util.spawn_with_shell = function(...)
    return spawn.with_shell(...)
 end
 
-util.pread = function(...)
-   util.deprecate("awful.spawn.pread")
-   return spawn.pread(...)
+util.pread = function()
+    util.deprecate("Use io.popen() directly or look at awful.spawn.easy_async() "
+            .. "for an asynchronous alternative")
 end
 
 return
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -276,23 +276,6 @@ function spawn.read_lines(input_stream, line_callback, done_callback, close)
     start_read()
 end
 
---- Read a program output and return its output as a string.
--- @tparam string cmd The command to run.
--- @treturn string A string with the program output, or the error if one
---   occured.
-function spawn.pread(cmd)
-    if cmd and cmd ~= "" then
-        local f, err = io.popen(cmd, 'r')
-        if f then
-            local s = f:read("*all")
-            f:close()
-            return s
-        else
-            return err
-        end
-    end
-end
-
 capi.awesome.connect_signal("spawn::canceled" , spawn.on_snid_cancel   )
 capi.awesome.connect_signal("spawn::timeout"  , spawn.on_snid_cancel   )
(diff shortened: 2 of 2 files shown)


blueyed:
Hmm, I think it's a good convenience wrapper, but am using it anyway wrapped with `rtrim`:

```lua
--- Remove trailing whitespace from string.
local rtrim = function(s)
  local n = #s
  while n > 0 and s:find("^%s", n) do n = n - 1 end
  return s:sub(1, n)
end
```

blueyed:
I support your comment at https://github.com/awesomeWM/awesome/issues/184#issuecomment-215947469 though: I can write my own wrapper and/or should make this async anyway.

Maybe the documentation could have a section regarding how to convert your `pread` calls (in case it's not there already)?

psychon:
How about making the deprecation message point to `awful.spawn.easy_async` and `with_line_callback`? That way people can find the async alternatives more easily.

blueyed:
Good idea!

blueyed:
What do you think about an even easier helper, which would automatically print `stderr` for example (in case of errors)?

Something like `function spawn_async(cmd, success_handler, error_handler)`, where the handlers would get `stdout` and `stderr`?
This way you could just call `spawn_async({'sh', '-c', '…'}, function(stdout) … end)`.

Without `error_handler` it would default to printing the error to stderr and/or create a notification.

blueyed:
Something like this:
```lua
my_easy_async = function(cmd, cb, cb_error, cb_signal)
  if not cb_error then
    cb_error = function(stdout, stderr, exitcode)
      gears.debug.print_error(string.format(
        '"%s" exited with %d. stderr: %s', cmd, exitcode, stderr))
    end
  end
  if not cb_signal then
    cb_signal = function(stdout, stderr, exitcode)
      gears.debug.print_error(string.format(
        '"%s" exited due to signal %d. stderr: %s', cmd, exitcode, stderr))
    end
  end

  awful.spawn.easy_async(cmd, function(stdout, stderr, exitreason, exitcode)
    if exitreason == 'exit' then
      if exitcode == 0 then
        if cb then
          cb(stdout, stderr)
        end
      else
        cb_error(stdout, stderr, exitcode)
      end
    else
      cb_signal(stdout, stderr, exitcode)
    end
  end)
end
```

psychon:
Fine by me, but I don't know what a good name for that would be. If we have too many functions in `awful.spawn` it will become confusing at some point.
Also: I'd print both stdout and stderr (perhaps only if non-empty?) because someone will find something which prints on stdout instead of stderr.

blueyed:
What about changing `easy_async` to make it that easy?

> Also: I'd print both stdout and stderr (perhaps only if non-empty?) because someone will find something which prints on stdout instead of stderr.

Agreed.

blueyed:
Let's discuss further on https://github.com/awesomeWM/awesome/pull/1001.

psychon:
How about this fixup commit? Does it make the deprecation warning clear enough?

blueyed:
Very good!
Thanks.
Please squash-merge yourself - not moving to "next" since there can only be one milestone.
Should we better use a label for "next" instead of a milestone?
Might also be worth checking out https://www.zenhub.com/ for this and in general (free for OSS).

PR #1001, opened 2016-07-11
Make spawn.easy_async more easy

blueyed:
Based on https://github.com/awesomeWM/awesome/pull/998#issuecomment-231548644.

The code feels very verbose, please criticize!

### DIFF
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -318,21 +318,58 @@ function spawn.with_line_callback(cmd, callbacks)
     return pid
 end
 
---- Asynchronously spawn a program and capture its output.
--- (wraps `spawn.with_line_callback`).
+--- Asynchronously spawn a program and (optionally) execute callbacks.
+-- `cmd` is being run and in case `cb`, `cb_error` or `cb_signal` are provided
+-- they will be run, according to the exit code of `cmd`.
+--
+-- This wraps `spawn.with_line_callback`.
 -- @tparam string|table cmd The command.
--- @tab callback Function with the following arguments
--- @tparam string callback.stdout Output on stdout.
--- @tparam string callback.stderr Output on stderr.
--- @tparam string callback.exitreason Exit Reason.
--- The reason can be "exit" or "signal".
--- @tparam integer callback.exitcode Exit code.
--- For "exit" reason it's the exit code.
--- For "signal" reason — the signal causing process termination.
--- @treturn[1] Integer the PID of the forked process.
+-- @tparam[opt] function cb Function to be called on success.  It will receive
+--   `stdout` and `stderr` as arguments.
+-- @tparam[opt] function cb_error Callback on errors. It will receive
+--   `stdout`, `stderr` and `exitcode` as arguments.
+-- @tparam[opt] function cb_signal Callback on signals. It will receive
+--   `stdout`, `stderr` and `signal` as arguments.
+-- @treturn[1] integer The PID of the forked process.
 -- @treturn[2] string Error message.
 -- @see spawn.with_line_callback
-function spawn.easy_async(cmd, callback)
+function spawn.easy_async(cmd, cb, cb_error, cb_signal)
+    if not cb_error or not cb_signal then
+        local function get_cmd_string(cmd)
+            return type(cmd) == 'table' and table.concat(cmd, ' ') or cmd
+        end
+        if not cb_error then
+            cb_error = function(stdout, stderr, exitcode)
+                require("gears.debug").print_error(string.format(
+                    '"%s" exited with %d. output: %s',
+                    get_cmd_string(cmd), exitcode,
+                    util.concat_table({stdout=stdout, stderr=stderr})))
+            end
+        end
+        if not cb_signal then
+            cb_signal = function(stdout, stderr, exitcode)
+                require("gears.debug").print_error(string.format(
+                    '"%s" exited due to signal %d. output: %s',
+                    get_cmd_string(cmd), exitcode,
+                    util.concat_table({stdout=stdout, stderr=stderr})))
+            end
+        end
+    end
+
+    local function callback(stdout, stderr, exitreason, exitcode)
+        if exitreason == 'exit' then
+            if exitcode == 0 then
+                if cb then
+                    cb(stdout, stderr)
+                end
+            else
+                cb_error(stdout, stderr, exitcode)
+            end
+        else
+            cb_signal(stdout, stderr, exitcode)
+        end
+    end
+
     local stdout = ''
     local stderr = ''
     local exitcode, exitreason
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -319,24 +319,39 @@ function spawn.with_line_callback(cmd, callbacks)
 end
 
 --- Asynchronously spawn a program and (optionally) execute callbacks.
+--
 -- `cmd` is being run and in case `cb`, `cb_error` or `cb_signal` are provided
 -- they will be run, according to the exit code of `cmd`.
 --
 -- This wraps `spawn.with_line_callback`.
+--
 -- @tparam string|table cmd The command.
--- @tparam[opt] function cb Function to be called on success.  It will receive
---   `stdout` and `stderr` as arguments.
--- @tparam[opt] function cb_error Callback on errors. It will receive
---   `stdout`, `stderr` and `exitcode` as arguments.
--- @tparam[opt] function cb_signal Callback on signals. It will receive
---   `stdout`, `stderr` and `signal` as arguments.
+--
+-- @tparam[opt] function cb Function to be called on success.
+-- @tparam string cb.stdout The output on stdout.
+-- @tparam string cb.stderr The output on stderr.
+--
+-- @tparam[opt] function cb_error Callback on errors.
+--   By default a message will be printed to awesome's stderr using
+--   `gears.debug.print_error`.
+-- @tparam string cb_error.stdout The output on stdout.
+-- @tparam string cb_error.stderr The output on stderr.
+-- @tparam string cb_error.exitcode The exit code (will be != 0).
+--
+-- @tparam[opt] function cb_signal Callback on signals.
+--   By default a message will be printed to awesome's stderr using
+--   `gears.debug.print_error`.
+-- @tparam string cb_signal.stdout The output on stdout.
+-- @tparam string cb_signal.stderr The output on stderr.
+-- @tparam string cb_signal.signal The signal.
+--
 -- @treturn[1] integer The PID of the forked process.
 -- @treturn[2] string Error message.
 -- @see spawn.with_line_callback
 function spawn.easy_async(cmd, cb, cb_error, cb_signal)
     if not cb_error or not cb_signal then
-        local function get_cmd_string(cmd)
-            return type(cmd) == 'table' and table.concat(cmd, ' ') or cmd
+        local function get_cmd_string(command)
+            return type(command) == 'table' and table.concat(command, ' ') or command
         end
         if not cb_error then
             cb_error = function(stdout, stderr, exitcode)
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -337,6 +337,7 @@ end
 -- @tparam string cb_error.stdout The output on stdout.
 -- @tparam string cb_error.stderr The output on stderr.
 -- @tparam string cb_error.exitcode The exit code (will be != 0).
+-- @tparam[opt='error'] string cb_signal.type 'error' always
 --
 -- @tparam[opt] function cb_signal Callback on signals.
 --   By default a message will be printed to awesome's stderr using
@@ -344,6 +345,7 @@ end
 -- @tparam string cb_signal.stdout The output on stdout.
 -- @tparam string cb_signal.stderr The output on stderr.
 -- @tparam string cb_signal.signal The signal.
+-- @tparam[opt='signal'] string cb_signal.type 'signal' always
 --
 -- @treturn[1] integer The PID of the forked process.
 -- @treturn[2] string Error message.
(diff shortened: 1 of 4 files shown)


lib/awful/spawn.lua
```diff
@@ -181,21 +181,72 @@ function spawn.with_line_callback(cmd, callbacks)
...
+-- @tparam[opt] function cb_signal Callback on signals. It will receive
+--   `stdout`, `stderr` and `signal` as arguments.
+-- @treturn[1] integer The PID of the forked process.
 -- @treturn[2] string Error message.
 -- @see spawn.with_line_callback
-function spawn.easy_async(cmd, callback)
+function spawn.easy_async(cmd, cb, cb_error, cb_signal)
+    if not cb_error or not cb_signal then
+        local function get_cmd_string(cmd)
+            return type(cmd) == 'table' and table.concat(cmd) or cmd
```
psychon: Don't you want to at least add spaces between the entries of `cmd`?

psychon:
Fine by me, but you broke `tests/test-awful-widget-watch.lua`.

blueyed:
@psychon 
Added some fixes (not sure if it was finished).

@Elv13 
Do you agree with this API?
My idea was for `awful.widget.watch` to hook into the error callback and display any error in the widget.

Elv13:
> Do you agree with this API?

I am not sure I like function with that many optional parameters, but beside that, fine by me.

tests/test-awful-widget-watch.lua
```diff
@@ -9,22 +9,20 @@ local steps = {
...
-                    assert(exitcode == 0, exitcode)
+                    assert(stdout == "hi\n", tostring(stdout))
+                    assert(stderr == "stderr\n", tostring(stderr))
                     callbacks_done = callbacks_done + 1
                 end,
                 "i_am_widget_mock"
             )
         end
-        if callbacks_done > 1 then  -- timer fired at least twice
+        if callbacks_done == 1 then
```
psychon: This comparison should stay the way it was before. The watch widget could have updated itself twice between two invocations of this function.

blueyed: How about `>= 1` then?
Because previously there were multiple callbacks (output and exit).

psychon: Oh, I didn't notice that part. That was indeed confusing. Yup, it should be `>= 1` or `> 0`.

psychon: Actually... no. I think this was part of the test. It tested that the callback was called more than just once on purpose, because `awful.widget.watch` should behave that way. Perhaps it should be `>= 2`?

blueyed:
Updated, but `awful.widget.watch` is not finished, e.g. `callback_error` is not used but sneaked in for some reason..
What do you think about having multiple (optional) callbacks for `watch`?
Should it just have one that behaves as before (getting the exit code), and then sets the error/stderr on the widget by default?

psychon:
Without looking at the code, what would be a good behaviour for the watch widget?
I'm not sure. Simplest approach might be to just copy the API signature from `easy_async` and have all the callbacks.
From a user's point of view... I don't know. The best case would be the user not having to specify a callback.

actionless:
Watch widget is still not finished, right? Because I can't understand the latest change (https://github.com/awesomeWM/awesome/pull/1001/commits/6e5c46e57915107ba20e0eeba779ed6d4acb6401).

blueyed:
@actionless 
> Watch widget is still not finished, right? Because I can't understand the latest change (https://github.com/awesomeWM/awesome/pull/1001/commits/6e5c46e57915107ba20e0eeba779ed6d4acb6401).

It was a fixup in the sense that on errors the output from "stderr" should be displayed on the widget.
The idea is that if you have a widget producing an error, this would be displayed in the widget already.  Not sure about this though - it just probably makes you really want to fix it?! :)

actionless:
ah, i've got the idea now

what do you think about concatenating stdout and stderr or showing stderr separately via naughty?

blueyed:
> what do you think about concatenating stdout and stderr or showing stderr separately via naughty?

Ok, so we'd display stdout always (and not in red?), but the stderr via naughty?  (possibly trimmed then)

actionless:
i was just thinking what some of the applications could print to both stderr and stdout in case of some problem (however i can't come up with any examples ATM)

Elv13:
@blueyed or @actionless Please do something about this PR. I would like 3.6 to happen (very) soon. If this isn't done this week, I will have to remove this from the 3.6 milestone.

actionless:
it's fine by me, i was just sharing some related ideas

blueyed:
@Elv13 
Won't have much time probably, but will try.

Elv13:
@blueyed, please do something about it now or remove this from the release blockers. As we both want a release, then it makes no sense to keep this as a release blocker.

blueyed:
Worked a bit on it.  Feel free to review it / leave comments.

TODO:
 - [ ] really stop commands, once they failed?  (could be done after X (consecutive) failures only)
   - [ ] when not stopping the command: once a notification is dismissed, it should not be displayed again probably?!

lib/awful/spawn.lua
```diff
@@ -178,21 +178,75 @@ function spawn.with_line_callback(cmd, callbacks)
...
+-- @tparam string cb.stdout The output on stdout.
+-- @tparam string cb.stderr The output on stderr.
+--
+-- @tparam[opt] function cb_error Callback on errors.
+--   By default a message will be printed to awesome's stderr using
+--   `gears.debug.print_error`.
+-- @tparam string cb_error.stdout The output on stdout.
+-- @tparam string cb_error.stderr The output on stderr.
+-- @tparam string cb_error.exitcode The exit code (will be != 0).
+-- @tparam[opt='error'] string cb_signal.type 'error' always
```
psychon: `cb_signal.type`? Should this be `cb_error.type`?

lib/awful/spawn.lua
```diff
@@ -178,21 +178,75 @@ function spawn.with_line_callback(cmd, callbacks)
...
+--
+-- @tparam[opt] function cb_signal Callback on signals.
+--   By default a message will be printed to awesome's stderr using
+--   `gears.debug.print_error`.
+-- @tparam string cb_signal.stdout The output on stdout.
+-- @tparam string cb_signal.stderr The output on stderr.
+-- @tparam string cb_signal.signal The signal.
```
psychon: `The signal` is a bit unspecific, I think. Signal number? Signal name? Should people be pointed to `awesome.signal_I_forgot_the_exact_name` that allows to map between signal numbers and names?

lib/awful/util.lua
```diff
@@ -582,6 +582,24 @@ function util.round(x)
...
+-- and trimmed on the right).
+-- @tab t Table to format, with key/value pairs as strings.
+-- @tparam string[opt=', '] delim Delimiter.
+-- @treturn[opt] string The concatenated string, or nil.
+function util.concat_table(t, delim)
+    local r = {}
+    for k,v in pairs(t) do
+        v = tostring(v):gsub('%s*$', '')
+        if #v > 0 then
+            table.insert(r, string.format('<b>%s:</b> %s', k, v))
```
psychon: This returns a string with HTML-markup? That doesn't fit well with the use above (printing to stderr) and isn't mentioned in the docs. Couldn't the code just use `gears.debug.dump_return`?

lib/awful/widget/watch.lua
```diff
@@ -60,22 +62,48 @@ local watch = { mt = {} }
...
+            output, (type == 'error' and "Exit code" or "Signal"), exitcode)
+        notification_id = naughty.notify({preset = naughty.config.presets.critical,
+                        title = title,
+                        text = text,
+                        replaces_id = notification_id,
+                    }).id
+        base_widget:set_text(stdout)
     end
-    local t = timer { timeout = timeout }
+    callbacks.signal = callbacks.signal or callbacks.error
```
psychon: So, if the caller provides an `error` callback, it will also be used for signals? That seems surprising.

lib/awful/widget/watch.lua
```diff
@@ -60,22 +62,48 @@ local watch = { mt = {} }
 -- For "exit" reason it's the exit code.
 -- For "signal" reason — the signal causing process termination.
 --
--- @param[opt=wibox.widget.textbox()] base_widget Base widget.
+-- @tparam[opt=wibox.widget.textbox()] widget base_widget Base widget.
 --
 -- @return The widget used by this watch
-function watch.new(command, timeout, callback, base_widget)
-    timeout = timeout or 5
+function watch.new(command, interval, callbacks, base_widget)
```
psychon: The actual parameter name and its documentation no longer matches: `callback` / `callbacks`.
Also, the whole table-stuff is undocumented...?

lib/awful/widget/watch.lua
```diff
@@ -60,22 +62,48 @@ local watch = { mt = {} }
...
+        }
+    end
+
+    local notification_id
+    callbacks.error = callbacks.error or function(stdout, stderr, exitcode, type)
+        -- widget:set_text(stderr)
+        local output = util.concat_table({stdout=stdout, stderr=stderr}, "\n") or "No output"
+        local title = (type == 'error'
+            and "An error occurred with a watch widget. Stopping it."
+            or  "A signal was received for a watch widget. Stopping it.")
```
psychon: I guess this needs a way to re-start the widget again. Sigh...

lib/awful/util.lua
blueyed: The idea was to beautify it for the naughty notification.
I've planned to allow for passing in the format here.

lib/awful/widget/watch.lua
blueyed: Signals are errors somehow, aren't they?
But we could fallback to some default here, too (instead of using the `error` cb).

lib/awful/widget/watch.lua
blueyed: Yes, that was the last part I've changed - and since ldoc warns about it I've thought that I would not forget to fix it later.. ;)

lib/awful/widget/watch.lua
blueyed: Yeah.
Maybe we should not stop it automatically at all, but instead ensure that we display notifications only once?!

lib/awful/widget/watch.lua
Elv13: > Signals are errors somehow, aren't they?

No, signals are process level interrupts. SIGPIPE is how "|" is implemented and some signals are sometime used as an RPC mechanism (for example, `dd` has some to print status. Other signals, such as SIGINT, SIGSTOP and SIGCONT are used in userspace scheduling. *Some* signals are errors, but not all of them, not even most of them.

psychon: Well, all signals that end up appearing here. If this code is run, then the process exited as a result of the default action of some signal (so SIGSTOP and SIGCONT will never show up here). What I was referring to was more like "this mixes things". All the other code in this file keeps "exit by `exit`" and "exit by deadly signal" separate.

blueyed: > So, if the caller provides an error callback, it will also be used for signals? That seems surprising.

Let's use it then only when no explicit error handler was used?!

lib/awful/widget/watch.lua
blueyed: But the callback is responsible for restarting it, and therefore the timer needs to be passed in there.

blueyed: ..or rather do it like with the success callback, which wraps the user's callback.

lib/awful/spawn.lua
```diff
@@ -318,21 +318,75 @@ function spawn.with_line_callback(cmd, callbacks)
...
+-- `cmd` is being run and in case `cb`, `cb_error` or `cb_signal` are provided
+-- they will be run, according to the exit code of `cmd`.
+--
+-- This wraps `spawn.with_line_callback`.
+--
 -- @tparam string|table cmd The command.
--- @tab callback Function with the following arguments
--- @tparam string callback.stdout Output on stdout.
--- @tparam string callback.stderr Output on stderr.
--- @tparam string callback.exitreason Exit Reason.
```
Elv13: If this was removed, then the API is broken and this can't be merged until v5 is branched. Please make sure the API remains compatible. Feel free to add `--TODO v5 deprecate this`

lib/awful/util.lua
Elv13: psychon is right, the HTML is rather unexpected and undocumented. How is this different from `gears.debug`? Why can't `gears.debug` be extended instead of adding this function?

Elv13:
Ok, so I did the review. It will have to be API/ABI compatible with existing config before it can be merged. Also I don't think the concat function should stay. Finally, I am not convinced the `easy_async` changes make it easier to use as the PR title seem to claim. I don't want to feel rude, but honestly, what's the point of this?

lib/awful/spawn.lua
blueyed: I guess that's why I wanted to have it before the release already then.
Initially we said that it would be OK to change in 4.1, but well.

blueyed:
> what's the point of this?

The main point seems to be having good default callbacks, and an easier interface.

Elv13:
> and an easier interface.

This is the detail I was never so sure about and why I tried to ignore the PR until now (hoping someone else would review/merge it). I mean, I am all for good API, but is having multiple callback really better? A single callback with all the arguments seems simpler to use, isn't it? Does handling the error in the callback really harder than having 2 callbacks?

blueyed:
> Does handling the error in the callback really harder than having 2 callbacks?

I'd say so: the default error handler can be sophisticated, and you would not want to do this all by yourself when using this.

blueyed:
It needs an update, but some more review(s) before.

psychon:
Closing for lack of activity in over one year

PR #1004, opened 2016-07-17
Fix entry use count for categories

### LINKED ISSUE #999: Fulltext search in the XDG menu is not working anymore

awesome v3.5.2-1867-g45d555d (The Fox)
 • Compiled against Lua 5.1.4 (running with Lua 5.1)
 • D-Bus support: ✔

Hit `Mod4 + p`, type `v`, `n`, `c`.

**Actual result:**

Nothing found.
45d555d

**Expected result:**

`X11VNC Server` and `TigerVNC Viewer` to be displayed as it used to be up to 45d555d.
b33fb2a

NB: If I type `x`, `1`, `1` then `X11VNC Server` is displayed at once. Again, if I type `t`, `i`, `g`, `e`, `r`, then `TigerVNC Viewer` is displayed. So the menu seems to match the user input just against the start of its items names for some reason.

modk:
Attempt to fix #999. Pretty untested :| feedback welcome.

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -113,12 +113,12 @@ local function load_count_table()
     local count_file = io.open (count_file_name, "r")
     local count_table = {}
 
-    -- read count file
+    -- read weight file
     if count_file then
         io.input (count_file)
         for line in io.lines() do
             local name, count = string.match(line, "([^;]+);([^;]+)")
-            if name ~= nil and count ~= nil then
+            if name ~= nil and weight ~= nil then
                 count_table[name] = count
             end
         end
@@ -135,7 +135,7 @@ local function write_count_table(count_table)
     if count_file then
         io.output (count_file)
 
-        for name,count in pairs(count_table) do
+        for name, count in pairs(count_table) do
             local str = string.format("%s;%d\n", name, count)
             io.write(str)
         end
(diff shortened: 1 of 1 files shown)


lib/menubar/init.lua
```diff
@@ -231,47 +231,74 @@ local function menulist_update(query, scr)
...
+                        -- use count from count_table if present
+                        if string.len(pattern) > 0 and count_table[v.name] ~= nil then
+                            v.count = tonumber(count_table[v.name])
+                        end
+
+                        if string.match(v.name, "^" .. pattern)
+                            or string.match(v.cmdline, "^" .. pattern) then
+                            v.prio = PRIO_HIGH
+                        else
+                            v.prio = PRIO_NORMAL
```
psychon: This is dead code. There is an `if` above which makes sure that we only end up here when `"^" .. pattern` matches.

Also, this patch is wrong since it drops the `match_inside` case. This is the same problem that the original patch brought: We now only consider matches at the beginning of the category name and ignore "interior matches" (but the original patch introduced that problem for applications, not for categories).


How about introducing a helper function to avoid all this code duplication:
It begins with the `string.match` check above and ends with the `table.insert` below. It gets two arguments: The priority if we have a beginning-match and the priority for "interior matches".

lib/menubar/init.lua
```diff
@@ -231,47 +231,74 @@ local function menulist_update(query, scr)
...
     if menubar.show_categories then
         for _, v in pairs(menubar.menu_gen.all_categories) do
             v.focused = false
             if not current_category and v.use then
                 if string.match(v.name, pattern) then
                     if string.match(v.name, "^" .. pattern) then
-                        table.insert(shownitems, v)
-                    else
-                        table.insert(match_inside, v)
+                        v.count = PRIO_NONE
```
psychon: This should be 0, not `PRIO_NONE`. Count is not a priority.

lib/menubar/init.lua
```diff
@@ -231,47 +231,74 @@ local function menulist_update(query, scr)
...
+                table.insert (command_list, v)
             end
         end
     end
 
     local function compare_counts(a,b)
+        if a.count == b.count then
+            return a.prio > b.prio
+        end
         return a.count > b.count
```
psychon: This means that first the counts are compared and only then the priorities. What do others think? Should the priority or the count be more important?
I think that it should be the other way around and the priority should win.

psychon:
CC @Jajauma 
You can get this code with `git pull https://github.com/modk/awesome.git menubar-priority-category-fix`

psychon:
I didn't look at this much, I just wanted to mention that the Travis build failed with the following error message from luacheck:
>    lib/menubar/init.lua:121:32: accessing undefined variable weight
>    lib/menubar/init.lua:228:11: unused variable match_inside

Jajauma:
The categories items on the first page of the menu are back now but got reordered compared with pre-45d555d versions though. Are they subject to usage counting or do we have another side effect here? Either case I can confirm this fixes https://github.com/awesomeWM/awesome/issues/999 at last (i.e. restores the fulltext search feature and categories icons).

psychon:
@Jajauma Yup, the code orders them according to counts ("weights").

This is fine by me (but the commits should be squashed, I guess?). What do other's think?

actionless:
i like this weights feature, later on it could be very helpful for implementing some dmenu-like launcher (ie which will behave the same as Launch Prompt (mod+R) but will appear like menubar)

Elv13:
I don't mind merging this, but I want to point that this feature isn't tested. The whole menubar implementation seems regression prone. "Crazylemon" on IRC told me this he/she was working on this. However as we keep adding code, this feels like a ticking bomb right now. Testing the menubar will require support for XDG DATA directories to add a test folder full of test `.desktop`. It won't be funny to write, but unfortunately, there is no way around it.

I don't use this, but I think it is a very important feature to have. So it should be better tested than "Pretty untested :| feedback welcome.".

Elv13:
About to be merged in #1083

PR #1005, opened 2016-07-17
Fix awful.util.pread() return value

psychon:
The deprecation wrapper that we still have for this function didn't return
anything. However, awful.util.pread() used to return strings. This breaks
script.

Work around this by returning an empty string. That way code will still break,
but at least it should not error out.


blueyed:
Thanks!

PR #1008, opened 2016-07-17
tests/run: change to script dir first

blueyed:
Then use `$source_dir` for detecting the build dir: this way you can run
`./run.sh some-test-file.lua` from `./tests/`.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -17,20 +17,20 @@ if [ "$CI" = true ]; then
     set -x
 fi
 
-# Either the build dir is passed in $CMAKE_BINARY_DIR or we guess based on $PWD
-build_dir="$CMAKE_BINARY_DIR"
-if [ -z "$build_dir" ]; then
-    if [ -d "$PWD/build" ]; then
-        build_dir="$PWD/build"
-    else
-        build_dir="$PWD"
-    fi
-fi
-
 # Change to file's dir (POSIXly).
 cd -P -- "$(dirname -- "$0")"
 this_dir=$PWD
 source_dir="$PWD/.."
+
+# Either the build dir is passed in $CMAKE_BINARY_DIR or we guess based on $PWD
+build_dir="$CMAKE_BINARY_DIR"
+if [ -z "$build_dir" ]; then
+    if [ -d "$source_dir/build" ]; then
+        build_dir="$source_dir/build"
+    else
+        build_dir="$source_dir"
+    fi
+fi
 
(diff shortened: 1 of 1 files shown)


blueyed:
> Merging #1008 into master will decrease coverage by 9.20%

Codecov issue, or did I break something here?

tests/run.sh
```diff
@@ -17,21 +17,21 @@ if [ "$CI" = true ]; then
...
+
 # Either the build dir is passed in $CMAKE_BINARY_DIR or we guess based on $PWD
 build_dir="$CMAKE_BINARY_DIR"
 if [ -z "$build_dir" ]; then
-    if [ -d "$PWD/build" ]; then
-        build_dir="$PWD/build"
+    if [ -d "$source_dir/build" ]; then
+        build_dir="$source_dir/build"
     else
         build_dir="$PWD"
```
psychon: Shouldn't this also be `build_dir="$source_dir"`, just like the change you did to the previous line?
(No idea which of these cases is run on travis, thinks are confusing and complicated...)

psychon:
The travis build log says " X> Failed to upload to S3". I restarted the job that does "codecov stuff" (which will likely break stuff even more, because now codecov gets duplicate reports...?)

So yup, seems to be "the internet randomly failed" and not related to the patch.

tests/test-awful-widget-watch.lua
```diff
@@ -21,7 +21,7 @@ local steps = {
                 "i_am_widget_mock"
             )
         end
-        if callbacks_done > 1 then  -- timer fired at least twice
+        if callbacks_done >= 1 then
```
psychon: This seems to have ended up in the wrong branch (I'm seeing this on #1008)?

tests/run.sh
blueyed: Not sure, but could make sense.  Will change it.

tests/test-awful-widget-watch.lua
blueyed: Woops, sorry - will fix it.

psychon:
Looks fine to me :+1: , but we should check @Elv13's use of running this script from the build dir so that we don't break that again. I'm not sure any more from where he called it... (and don't have the time to investigate)

blueyed:
@Elv13 
Please merge in case you have no objections.

Elv13:
Yes, it works, sorry for the delay

PR #1009, opened 2016-07-23
drawable: Add more data to the find_widgets metadata table

Elv13:
The metadata commit is necessary to correctly re-implement @kindlycat slider widget using linear algebra. The previous version didn't work when used in a mirror or reflection widget. It also implemented `width` and `vertical` itself instead of using containers. I wont publish my fixed version now, I did say "no more code for 3.6" and I will keep my word on this*. However, it wasn't possible to make it work without this change.

Basically, you have to convert the `{x,y}` coordinate of the mouse to the relative ones compared to the **transformed** origin point of the widget.

\* The widget will be part of my Radical module for now so early adopters can test it.

### DIFF
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -170,7 +170,9 @@ local function find_widgets(_drawable, result, _hierarchy, x, y)
             0, 0, width, height)
         table.insert(result, {
             x = x3, y = y3, width = w3, height = h3,
-            drawable = _drawable, widget = _hierarchy:get_widget()
+            drawable = _drawable, widget = _hierarchy:get_widget(),
+            matrix_to_device = _hierarchy:get_matrix_to_device(),
+            matrix_to_parent = _hierarchy:get_matrix_to_parent(),
         })
     end
     for _, child in ipairs(_hierarchy:get_children()) do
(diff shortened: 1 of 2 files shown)


lib/wibox/drawable.lua
```diff
@@ -170,7 +170,9 @@ local function find_widgets(_drawable, result, _hierarchy, x, y)
             0, 0, width, height)
         table.insert(result, {
             x = x3, y = y3, width = w3, height = h3,
-            drawable = _drawable, widget = _hierarchy:get_widget()
+            drawable = _drawable, widget = _hierarchy:get_widget(),
+            matrix = _hierarchy:get_matrix_to_parent(),
+            full_path = result,
```
psychon: Why do you need the full path? What do you even mean by full path here? The `result` table will not be a path. If you have two overlapping widgets, it will e.g. include both, even though neither includes the other.

Do you perhaps want `matrix = _hierarchy:get_matrix_to_device()`?
The matrix to the parent cannot really be useful here, because code using this knows nothing about where the parent widget is. The device (=drawable coordinates) seems a lot more useful.

Elv13:
Updated, I removed the "full path" thing as it can be replaced by having both matrices.

psychon:
:+1:

PR #1011, opened 2016-07-24
Commandline option to add a directory to Lua's search path.

psychon:
Here is my version of https://github.com/awesomeWM/awesome/pull/485. This uses option `-s` / `--search`, but I'm not really sure if this is better than `--include` (I was thinking of C's `#include` and that this option does something else than that, which is why I changed the name).


blueyed:
Just a quick thought: I don't think we need a short option for this, do we?  I think it's better to have it explicit with `--search` always/only.

psychon:
Sure, here would be the corresponding patch: https://gist.github.com/psychon/227d63cff0183bd71448c2f3ea3a3069 (I'll add this as fixup commits when I find the time)

blueyed:
@psychon 
> I'll add this as fixup commits when I find the time

I see some opportunities to improve your Git workflow right there - since you're able to upload a gist for a patch and plan to add it to your own PR later: just push a fixup commit (see https://github.com/marcgibbons/django-rest-swagger/pull/523#issuecomment-240541259 for my setup and how it can go wrong (but you know about that part already).
The question from this being: why do you post it into a gist (via some command hopefully), but not just commit and push it?  (just curious :))

psychon:
I was at work when I created that gist. The main difference there being that my .gitconfig is set up with a different mail address (but I think awesome already contains commits from me from that mail address...).

psychon:
> why do you post it into a gist (via some command hopefully)

Does copy&paste from `git diff | gvim -R -` count as a command? ;-)

Anyway, is this change now ok by you? Do others have complaints about this?

blueyed:
Yes, it's nice now.
And thanks for explaining your workflow!

awesome.c
```diff
@@ -514,6 +512,12 @@ main(int argc, char **argv)
             else
                 fatal("-c option requires a file name");
             break;
+          case 's':
+            if(a_strlen(optarg))
+                string_array_append(&searchpath, a_strdup(optarg));
+            else
+                fatal("-s option requires a directory name");
+            break;
```
blueyed: This should refer to --search, not -s.

psychon: Whoops, yeah. However, this whole error should just be removed instead:
https://github.com/awesomeWM/awesome/pull/1063
https://github.com/awesomeWM/awesome/pull/1063/commits/4cbf87b441bf36961827a7372fcf8172bc03a75a

PR #1017, opened 2016-07-26
Change move and resize cursors

juw:
Hi,

the fleur should actually be used for the resize cursor (as it was on awesome 3.5.9), therefore, this pull request just switches both cursors.

Cheers.


psychon:
I don't even remember what the cursors look like, but I trust you that they are swapped. Unless @Elv13 disagrees: :+1:

PR #1020, opened 2016-07-27
restore client.border_width after fullscreen

cmertz:
client.border_width is not reset to its original value after fullscreen which results in the client borders effectively beeing removed after toggling fullscreen.

This should fix this behaviour (i.e. restore client.border_width to the original value before fullscreen).

### DIFF
--- a/tests/test-maximize.lua
+++ b/tests/test-maximize.lua
@@ -45,7 +45,9 @@ local steps = {
     -- Test restoring client.border_width
     function()
         local c = client.get()[1]
-        local test_width = 2
+
+        -- pick an arbitrary border_width distinct from the default one
+        local test_width = c.border_width + 1
 
         c.border_width = test_width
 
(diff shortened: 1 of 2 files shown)


tests/test-maximize.lua
```diff
@@ -41,6 +41,22 @@ local steps = {
         c.maximized_horizontal = false
         return true
     end,
+
+    -- Test restoring client.border_width
+    function()
+        local c = client.get()[1]
+        local test_width = 2
```
psychon: How about making this `c.border_width * 2` so that the test also works if the default border width is changed to 2?

cmertz: didn't think so far, thanks :)
```c.border_width + 1``` works for default border width of 0 as well

psychon: Sure

psychon:
:+1:

blueyed:
@Elv13 
I think this should have been squashed into a single commit.

PR #1025, opened 2016-07-31
Travis: allow failures with LuaJIT

blueyed:
This is not nice and we should still try to fix it, but it seems better
to have green PRs without triggering a rebuild etc.

See https://github.com/awesomeWM/awesome/issues/982#issuecomment-231522318.


Elv13:
Well, I think #982 is actually a real race condition (and a real bug), that only happen on LuaJIT because (or not) it's faster. I am not really against this PR, but we should make sure there is no races in the screen code first, right?

blueyed:
> we should make sure there is no races in the screen code first, right?

Agreed.

Still merging this for now.

PR #1026, opened 2016-08-01
Misc fixes

Elv13:


lib/wibox/widget/base.lua
```diff
@@ -522,6 +524,11 @@ function base.make_widget_declarative(args)
     local mt = getmetatable(w) or {}
     local orig_string = tostring(w)
 
+    -- Add the main id (if any)
+    if id then
+        ids[id] = ids[id] or {}
+        table.insert(ids[id], 1, w)
+    end
```
blueyed: FYI: The code in the `if` block is not covered by tests.
(you can use https://github.com/codecov/browser-extension to see it in the browser (this view))

blueyed:
LGTM.

PR #1031, opened 2016-08-08
Add recursive signal for widgets trees.

Elv13:
This was necessary to (sanely) fix some issues with client tabbing when used with the dynamic client layout system. So now, there is a way to bubble up some signals. I am happy with this solution, it's simple and well integrated. My other solution to the problem was very intrusive and made the hierarchy ugly.

The other commit add more signals.

Edit: I will add tests before merging if @psychon agrees with this change.

### DIFF
--- a/docs/common/widget.ldoc
+++ b/docs/common/widget.ldoc
@@ -40,6 +40,13 @@
 -- @param _buttons The table of buttons that should bind to the widget.
 -- @function buttons
 
+--- Emit a signal and ensure all parent widgets in the hierarchies also
+-- forward the signal. This is useful to track signals when there is a dynamic
+-- set of containers and layouts wrapping the widget.
+-- @tparam string signal_name
+-- @param ... Other arguments
+-- @function emit_signal_recursive
+
 --- When the layout (size) change.
 -- This signal is emited when the previous results of `:layout()` and `:fit()`
 -- are no longer valid.
--- a/docs/common/widget.ldoc
+++ b/docs/common/widget.ldoc
@@ -40,7 +40,6 @@
 -- @param _buttons The table of buttons that should bind to the widget.
 -- @function buttons
 
-
 --- When the layout (size) change.
 -- This signal is emited when the previous results of `:layout()` and `:fit()`
 -- are no longer valid.
--- a/lib/awful/widget/button.lua
+++ b/lib/awful/widget/button.lua
@@ -1,4 +1,12 @@
 ---------------------------------------------------------------------------
+-- A simple button widget.
+-- @usage local button = awful.widget.button()
+-- button:buttons(awful.util.table.join(
+--     button:buttons(),
+--     awful.button({}, 1, nil, function ()
+--         print("Mouse was clicked")
+--     end)
+-- ))
 -- @author Julien Danjou &lt;julien@danjou.info&gt;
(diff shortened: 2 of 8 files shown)


docs/common/widget.ldoc
```diff
@@ -40,6 +40,12 @@
 -- @param _buttons The table of buttons that should bind to the widget.
 -- @function buttons
 
+--- Emit a signal and ensure all parent widgets in the hierarchies also
+-- forward the signal. This is useful to track signals when there is a dynamic
+-- set of containers and layouts wrapping the widget.
+-- @tparam string signal_name
+-- @param ... Other arguments
+-- @function emit_signal_recursive
```
psychon: This part seems to be in the wrong commit. The next commit adds this function.

lib/wibox/hierarchy.lua
```diff
@@ -51,6 +51,15 @@ local function hierarchy_new(redraw_callback, layout_callback, callback_arg)
         end
         layout_callback(result, callback_arg)
     end
+    function result._emit_recursive(_, name, ...)
+        local cur = result
```
psychon: I'd propose to do
```
function result._emit_recursive(self, name, ...)
  local cur = self
```
That way, this function still has to be called with the correct argument.

psychon:
Technically fine by me, but the implementation emits excessive signals and I do not really know how to avoid that. A simple example would be if you have two screens. The wibars share e.g. the textclock widget and each wibar's hierarchy will forward the recursive signal up to the parents. This can lead to a signal being emitted multiple times on some widget.

The other issue is "this only works after the wibox updated which needs an indeterminate amount of time", but I can ignore that.

I guess that a proper implementation of this would need us to throw out the current "a widget can be visible in multiple places at once" and instead have a single position for a widget. That way, widget's could finally known their own size...

Should we add this functionality even though we have the above limitations? I'm not really a fan of this.

lib/wibox/hierarchy.lua
Elv13: Nope, wrong self. `self` is the widget, we need the hierarchy. I added an assert instead

Elv13:
Updated. I tested the recursive part in a random test suit. The ratio/fixed parts have tests, but they are in the big "lets rewrite the client layout system" PR. When both of them are there, it has nearly full coverage.

lib/wibox/hierarchy.lua
psychon: Oh, right. Sorry & thanks.

lib/awful/widget/button.lua
```diff
@@ -1,4 +1,12 @@
 ---------------------------------------------------------------------------
+-- A simple button widget.
+-- @usage local button = awful.widget.button()
+-- button:buttons(awful.util.table.join(
+--     button:buttons(),
+--     awful.button({}, 1, nil, function ()
+--         button:emit_signal_recursive("test::recursive")
```
psychon: How about `print("Mouse was clicked")`? That might be "less weird" for users.

psychon:
If you really want to have this, ok. However, could you add the problems of this approach to the docs (Only works if the widget is shown in a drawin which already repainted since the last (relevant) layout change; if the widget is visible in multiple places, the signal is "forwarded" multiple times)?

PR #1035, opened 2016-08-10
Revert "screen: Fix potential nil index"

aroig:
This reverts commit facf676b1367e6915235a45ce3e7ef11790b0411.

Using capi.client.focus.screen to decide which screen is focused breaks
a multiscreen setup. At least makes it extremely annoying to use.

In particular, if you have a focused client on screen 1, move the mouse
to screen 2 and launch a new client, the new client appears in screen 1,
since screen.focused reports that current focused screen is 1, not 2
because of the focused client.

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -230,11 +230,7 @@ end
 --   present currently.
 function screen.focused(args)
     args = args or screen.default_focused_args or {}
-    return get_screen(
-        (args.client and args.client.screen) or
-        (capi.client.focus and capi.client.focus.screen) or
-        capi.mouse.screen
-    )
+    return get_screen(args.client and capi.client.screen or capi.mouse.screen)
 end
 
 --- Get a placement bounding geometry.


Elv13:
This cannot be reverted as the original logic is flawed and will cause runtime errors if args.client is specified (this is an issue I accidentally introduced in a messed up rebase).

This doesn't mean the PR is without merit. The issue is genuine and ages old. The underlying problem isn't simple to fix. On one hand, some users that don't use the mouse and navigate between [terminal] clients will get the "wrong" result out of the "mouse.screen is the truth" algorithm. However, you are right that my fix did break 2 important use cases.

 * People not using focus follow the mouse will have issues
 * Screens without a focusable client cannot have focus

For the first scenario, I honestly have no good idea how to fix it. For the second, it can be auto-detected by checking `mouse.current_client`. If that returns `nil`, then mouse.screen could "win". This behavior, while not perfect, would be more acceptable than the current one.

In a bit of history, there was previously 2 attempts at fixing this.

 * Awesome 3.4 and 3.5 documentation tell the user: "if you are not happy with this, monkeypatch this function". This is in itself a valid workaround, but has limitations. First, it is hacky. Second, it requires non trivial plumbing from rc.lua. Third, I am against this kind of API, it is a bad design. Finally, it makes third party modules harder to write as the result is unpredictable.

* Commit 83df30d3a436f0842c5c6b159a519dcaefa0e6d3q tried to introduce and argument and a global configuration variable to configure the default mode. However, the code was flawed and caused an error. Your pull request here revert the commit that fixes this error (and re-introduce the issue it tried to solve, even making it slightly worse for your usage scenario)

So I propose yet another try at getting this "right" for the majority of users. I suggest adding `args.mode` with these:

* "auto" (default): If a client is focused and the screen is mouse screen is the same, return that screen. If the mouse is over a focusable client on a screen, but the focussed client is on another, choose the client screen. If the mouse is over no client or a non focusable one, choose the mouse screen. If there is no focused client, choose the mouse screen.
* "mouse": Always pick the mouse screen
* "client": Always pick the client screen, then fallback to the mouse screen

What do you think?

aroig:
> This cannot be reverted as the original logic is flawed and will cause runtime errors if args.client is specified (this is an issue I accidentally introduced in a messed up rebase).

Ah, thanks for the lengthy explanation! I did not understand the issue you tried to fix. However facf676 still seems broken to me, if the documentation is valid.

1. Shouldn't `args.client` be a boolean? why do you use `args.client.screen` then?
2. It completely ignores the case `args.mouse = true`. It still lets `capi.client.focus.screen` win over `capi.mouse.screen`.
3. In any case, I'd prefer to have `args.mode` as you suggested, instead of two booleans for mutually exclusive options.

I'd be ok if `screen.focused` respected a global `screen.default_focused_args.mouse=true`, or equivalent setting, I think. Maybe it should even be default in order to not break tons of workflows. Mine is broken, and I don't think I have a particularly weird usage pattern.

I admit I still do not understand the use of `args.client=true` though. How do you launch clients on an empty tag on a second screen with such setting?

Also note that in addition to the two problems you pointed out

* People not using focus follow the mouse will have issues,
* Screens without a focusable client cannot have focus,

the navigation with `screen.focus_bydirection` `client.focus.global_bydirection`, and client-moving counterparts is also broken due to the second point above.

Elv13:
Ok, fine. You are right. I made a mess here and instead of just fixing a bug shoot myself in the foot. I made a new version of your commit that doesn't re-introduce the broken code, but restore the old behavior intact.

aroig:
Thanks @Elv13, for the quick fix and for the rest of your work on awesome!

Regarding this focus mess, I think the deeper issues are not yet resolved. I'll open a new issue for discussing that.

PR #1037, opened 2016-08-11
Revert "screen: Fix potential nil index"

### LINKED ISSUE #1029: Can't screen.focus() from empty tag
 • Build: Mar 12 2016 01:11:40 for x86_64 by gcc version 5.3.0 (builduser@rw)
 • Compiled against Lua 5.3.2 (running with Lua 5.3)
 • D-Bus support: ✔

**Reproduce:**

1. Setup multiple screens with xrandr one by the other
2. Bind F1, F2 to `awful.screen.focus()`
3. Go to empty tag without clients.
4. Try switch between screens.

**Actual result:**

So it doesn't switch, you have to move to populated tag before that.

**Expected result:**

Really annoys sometimes. Is there a workaround?

Elv13:
This reverts commit facf676b1367e6915235a45ce3e7ef11790b0411.

Using capi.client.focus.screen to decide which screen is focused breaks
a multiscreen setup. At least makes it extremely annoying to use.

In particular, if you have a focused client on screen 1, move the mouse
to screen 2 and launch a new client, the new client appears in screen 1,
since screen.focused reports that current focused screen is 1, not 2
because of the focused client.

Close #1035
Fix #1029

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -231,9 +231,7 @@ end
 function screen.focused(args)
     args = args or screen.default_focused_args or {}
     return get_screen(
-        (args.client and args.client.screen) or
-        (capi.client.focus and capi.client.focus.screen) or
-        capi.mouse.screen
+        args.client and capi.client.focus and capi.client.focus.screen or capi.mouse.screen
     )
 end
 


Elv13:
@blueyed Feel free to merge this directly once reviewed

PR #1042, opened 2016-08-13
Travis: fix return/exit usage

blueyed:
Return 0 (success) in do_codecov function.  It is used in "&&" chains,
and should just do nothing for "coverall", but not break it.

This also makes use of "set -ev" instead of "exit" in general.

The "exit" caused the whole build to stop, but it was still green!
https://travis-ci.org/awesomeWM/awesome/jobs/151943526

"exit" is not really supported:
https://docs.travis-ci.com/user/customizing-the-build/#How-does-this-work%3F-(Or%2C-why-you-should-not-use-exit-in-build-steps)

This also wraps (most of) the "cd" commands in a subshell, to ensure
being in the repo directory all the time (except for $BUILD_IN_DIR).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,18 +43,17 @@ install:
   # Install Lua (per env).
   # Note that Lua 5.3 is installed manually, because it is not available in Ubuntu Trusty.
   - |
+    set -ev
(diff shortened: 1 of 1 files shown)


blueyed:
Followup: #1043.

PR #1044, opened 2016-08-16
Upstream shape api part 4 (of 4)

Elv13:
Ok, here is  giant PR I once said I would not submit before a release. But here it is anyway. I first said it existed back in January, so it should not be such a big surprise. I cleaned and completed the various parts since May.

So first, it does complete the shape API

-- \* `client`s (see `gears.surface.apply_shape_bounding`)
 -- \* `wibox`es (see `wibox.shape`)
 -- \* All widgets (see `wibox.container.background`)
 -- \* The progressbar (see `wibox.widget.progressbar.bar_shape`)
 -- \* The graph (see `wibox.widget.graph.step_shape`)
 -- \* The checkboxes (see `wibox.widget.checkbox.check_shape`)
 -- \* Images (see `wibox.widget.imagebox.clip_shape`)
 -- \* Sliders (see `wibox.widget.slider.handle_shape`)
 -- \* The taglist tags (see `awful.widget.taglist`)
 -- \* The tasklist clients (see `awful.widget.tasklist`)
 -- \* The tooltips (see `awful.tooltip`)

It also adds a lot of doc. Actually, 50% of the PR is doc related. It adds tens of examples, dozens of images and 2 whole new documentation pages. As @psychon asked, I didn't try to implement the new auto-ganarated pages using CMake. I used lua scripts instead. So now all beautiful variables are documented and an index is auto-generated. The other pages is to documented `rc.lua`. It could help meta-document any user file too (in case the config is eventually splitted). Beside this, I manually read the whole ldoc documentation and fixed the issues I saw.

And yeah, this one is **HUGE**. Sorry about that. I could make some iterations. But why hiding the whole picture.

This commit also add the grid layout, just because I had it ready (it was in Radical since a while). I know luacheck still find issues and that there is a few obvious bugs, but as far as development goes, this is "done".

slider1

Edit: And thanks to @actionless for testing this a little and contributing the notification part.


Elv13:
So, just to confirm what I said. Here is the breakdown:

1627 lines of doc
1304 lines of tests / examples
563 lines of meta-documentation / linters
996 empty lines
925 lines of new code

So it is actually not _that_ big.

lib/wibox/widget/progressbar.lua
```diff
@@ -181,16 +182,24 @@ function progressbar:set_value(value)
 end
 
 --- Set the progressbar height.
+-- This method is deprecated.
+-- Use a `wibox.container.constraint` widget or `forced_height`.
 -- @param height The height to set.
-function progressbar:set_height(height)
+-- @deprecated set_height
+function progressbar:set_height(height) --luacheck: no unused_args
```
psychon: I don't think that this luacheck comment is needed, is it? (Same comment elsewhere here)

(I tried looking at the "complete diff" for this for a quick check, but that's just too big and mixes lots of things. Thus, I'll take a quick look through the individual commits).

lib/wibox/widget/progressbar.lua
```diff
@@ -144,10 +148,31 @@ function progressbar:set_value(value)
...
 end
 
+function progressbar:set_max_value(max_value)
+
+    self._private.max_value = max_value
+
+    -- Restore the value that was trimmed due to a race condition
+    if self._private.value_cache and self._private.value_cache <= max_value then
+        self._private.value = self._private.value_cache
+    end
```
psychon: Hm... How about having `set_value` only do `self._private.value = value` and move the `math.min(max_value, math.max(0, value))` into `:draw()`? I.e. we never trim the value that is saved, but only apply the triming while drawing the widget. That might be simpler to understand than this version.

lib/wibox/container/radialprogressbar.lua
```diff
@@ -0,0 +1,264 @@
...
+
+function radialprogressbar:set_widget(widget)
+    if widget then
+        base.check_widget(widget)
+    end
+    self._private.widget = widget
+    self:emit_signal("widget::layout_changed")
+end
+
+--- Get the number of children element
```
psychon: `Get the number of children element` doesn't sound like it describes what this function does, does it?

lib/wibox/container/radialprogressbar.lua
```diff
@@ -0,0 +1,264 @@
...
+
+--- Replace the layout children
+-- This layout only accept one children, all others will be ignored
+-- @tparam table children A table composed of valid widgets
+function radialprogressbar:set_children(children)
+    self._private.widget = children and children[1]
+    self:emit_signal("widget::layout_changed")
+end
+
+--- Reset this layout. The widget will be removed and the rotation reset.
```
psychon: Which rotation is being reset?

lib/wibox/container/radialprogressbar.lua
```diff
@@ -0,0 +1,264 @@
...
+function radialprogressbar:reset()
+    self:set_widget(nil)
+end
+
+for k,v in ipairs {"left", "right", "top", "bottom"} do
+    radialprogressbar["set_"..v.."_padding"] = function(self, val)
+        self._private.paddings = self._private.paddings or {}
+        self._private.paddings[v] = val
+        self:emit_signal("widget::redraw_needed")
```
psychon: Shouldn't it (also) be `layout_changed`? The padding is used by `:fit` and `:layout` (and `:before_draw_children`).

lib/wibox/container/radialprogressbar.lua
```diff
@@ -0,0 +1,264 @@
...
+-- @property value
+-- @tparam number value Between min_value and max_value
+
+function radialprogressbar:set_value(val)
+    if not val then self._percent = 0; return end
+
+    if val > self._private.max_value then
+        self:set_max_value(val)
+    elseif val < self._private.min_value then
+        self:set_min_value(val)
```
psychon: Does this work? As far as I can tell, there is no `max/min`, but `maximum/minimum`.

lib/wibox/container/radialprogressbar.lua
```diff
@@ -0,0 +1,264 @@
...
+
+--- The maximum value.
+-- @property max_value
+
+for _, prop in ipairs {"maximum_value", "minimum_value", "border_color", "color",
+    "border_width", "paddings"} do
+    radialprogressbar["set_"..prop] = function(self, value)
+        self._private[prop] = value
+        self:emit_signal("property::"..prop)
+        self:emit_signal("widget::redraw_needed")
```
psychon: `"paddings"` also influences the layout and thus needs `widget::layout_changed`.

lib/wibox/container/radialprogressbar.lua
```diff
@@ -0,0 +1,264 @@
...
+        right  = val,
+        top    = val,
+        bottom = val,
+    } or val or {}
+    self:emit_signal("property::paddings")
+    self:emit_signal("widget::redraw_needed")
+end
+
+--- Returns a new radialprogressbar layout. A radialprogressbar layout radialprogressbars a given widget. Use
+-- :set_widget() to set the widget and :set_direction() for the direction.
```
psychon: There is no `set_direction`, is there?

lib/wibox/widget/checkbox.lua
```diff
@@ -0,0 +1,253 @@
...
+local function new(checked, args)
+    checked, args = checked or false, args or {}
+
+    local ret = base.make_widget(nil, nil, {
+        enable_properties = true,
+    })
+
+    util.table.crush(ret, checkbox)
+
+    ret._private.style = args.style or beautiful.checkbox_style or "default"
```
psychon: This `style` value is never used

lib/wibox/widget/checkbox.lua
```diff
@@ -0,0 +1,253 @@
...
+    return ret
+end
+
+--@DOC_widget_COMMON@
+
+--@DOC_object_COMMON@
+
+return setmetatable({}, { __call = function(_, ...) return new(...) end})
+
+-- kate: space-indent on; indent-width 4; replace-tabs on;
```
psychon: Heh, someone is working against vim modelines :-)

lib/wibox/widget/checkbox.lua
```diff
@@ -0,0 +1,253 @@
...
+    wa.x      = offset + wa.x + (padding.left or 1)
+    wa.y      = offset + wa.y + (padding.top  or 1)
+    wa.width  = wa.width  - (padding.left or 1) - (padding.right  or 1) - 2*offset
+    wa.height = wa.height - (padding.top  or 1) - (padding.bottom or 1) - 2*offset
+
+    return wa
+end
+
+local function draw(self, context, cr, width, height)
+    local size = math.min(width, height)
```
psychon: Could `draw` "really use" the provided width/height instead of using the minimum of them? If some layout ignores the fit result and gives this widget a non-rectangular area, some part would otherwise be left "empty" and this could be confusing to users.

tests/examples/wibox/widget/checkbox/custom.lua
```diff
@@ -0,0 +1,35 @@
...
+            cr:line_to( 0  , rs )
+            cr:move_to( 0  , rs )
+            cr:line_to( rs , rs )
+            cr:move_to( rs , 0  )
+            cr:line_to( rs , rs )
+            cr:move_to( 0  , 0  )
+            cr:line_to( rs , rs )
+            cr:move_to( 0  , rs )
+            cr:line_to( rs , 0  )
+            cr:close_path()
```
psychon: Why this `close_path`? I'd leave it out. It doesn't really do much here and seems weird to me. The other subpaths aren't explicitly closed either.

lib/gears/shape.lua
```diff
@@ -376,6 +376,37 @@ function module.losange(cr, width, height)
...
+-- @tparam number height The shape height
+-- @tparam[opt=0] number start_angle The start angle (in radian)
+-- @tparam[opt=math.pi/2] number end_angle The end angle (in radian)
+-- @tparam[opt=math.min(width height)/2] number radius The shape height
+function module.pie(cr, width, height, start_angle, end_angle, radius)
+    radius = radius or math.floor(math.min(width, height)/2)
+    start_angle, end_angle = start_angle or 0, end_angle or math.pi/2
+
+    -- If the shape is a circle, then avoid the lines
+    if math.abs(start_angle + end_angle) == 2*math.pi then
```
psychon: How about `>=` instead of `==`?

lib/gears/shape.lua
```diff
@@ -376,6 +376,37 @@ function module.losange(cr, width, height)
...
+        cr:arc(width/2, height/2, radius, 0, 2*math.pi)
+    else
+        cr:move_to(width/2, height/2)
+        cr:line_to(
+            width/2 + math.cos(start_angle)*radius,
+            height/2 + math.sin(start_angle)*radius
+        )
```
psychon: Is this `:line_to` really needed? I'm not 100% sure, but I think `:arc` does this automatically.
Also, you could use `rel_line_to` (or whatever the function is called exactly) to get rid of the `width/2+` and `height/2+`.

lib/wibox/container/arcchart.lua
```diff
@@ -0,0 +1,317 @@
...
+
+function arcchart:set_widget(widget)
+    if widget then
+        base.check_widget(widget)
+    end
+    self._private.widget = widget
+    self:emit_signal("widget::layout_changed")
+end
+
+--- Get the number of children element
```
psychon: `number of`?

lib/wibox/container/arcchart.lua
```diff
@@ -0,0 +1,317 @@
...
+-- @tparam number value Between min_value and max_value
+
+function arcchart:set_value(val)
+    if not val then self._percent = 0; return end
+
+--     if val > 0 then
+--         self:set_max_value(val)
+--     elseif val < 0 then
+--         self:set_min_value(val)
+--     end
```
psychon: Any reason not to remove this code?

lib/wibox/container/arcchart.lua
```diff
@@ -0,0 +1,317 @@
...
+
+    local delta = 0 - 0
+
+    self._percent = val/delta
```
psychon: ...or this whole function if it just does a division by zero. Huh?

lib/wibox/container/arcchart.lua
```diff
@@ -0,0 +1,317 @@
...
+-- @tparam[opt=3] number border_width
+
+--- The minimum value.
+-- @property min_value
+
+--- The maximum value.
+-- @property max_value
+
+for _, prop in ipairs {"border_width", "border_color", "paddings", "colors",
+    "rounded_edge", "bg", "thickness", "values", "min_value", "max_value" } do
```
psychon: Consistency: Here it is called `min/max_value`, some previous widget called it `minimum/maximum_value` (see the comment that I left there). It would be nice if both used the same name.

lib/wibox/container/arcchart.lua
```diff
@@ -0,0 +1,317 @@
...
+        top    = val,
+        bottom = val,
+    } or val or {}
+    self:emit_signal("property::paddings")
+    self:emit_signal("widget::redraw_needed")
+end
+
+--- Returns a new arcchart layout. A arcchart layout arccharts a given widget. Use
+-- :set_widget() to set the widget and :set_direction() for the direction.
+-- The default direction is "north" which doesn't change anything.
```
psychon: There is no direction on this widget either.

Also, "A arcchart layout archarts a given widget" doesn't really say anything...

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,548 @@
...
+
+function event.mouse_enter(self)
+    self:emit_signal("widget::redraw_needed") --TODO
+    local cursor = self._private.cursor or "fleur"
+    capi.root.cursor(cursor)
+end
+
+function event.mouse_leave(self)
+    self:emit_signal("widget::redraw_needed") --TODO
+--     capi.root.cursor("fleur")
```
psychon: This seems unfinished... The cursor is changed, but never reset? Also, why is the cursor on the root window changed? That seems useless and not related to this widget at all. (And no, right now I have no good idea nor much interest to implement per-widget mouse cursors)

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,548 @@
...
+end
+
+local rect_order = {"x", "y", "width", "height"}
+
+-- Move the handle to the correct location
+local function move_handle(self, x, y, geo)
+    -- Fix the X position when used in a rotate or mirror widget
+
+    -- Remove the translation to bring {x, y} and the rect on the same origin
+    local rem_translate = geo.matrix_to_device*gmatrix.identity
```
psychon: Multiplication with the identity matrix is a no-op...?

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,548 @@
...
+
+    -- If the transformation moved the rectangle away from {0,0}, compensate
+    if rect.x < 0 then
+        px = rect.width + px
+    end
+
+    --HACK Cairo 0,0 is top left, a matrix is bottom left, funny...
+    if util.round(rem_translate:transform_point(0, 100)) ~= 0 then
+        px = -(px - rect.width)
+    end
```
psychon: I don't get the math here and it doesn't seem to me like the author really understood it either. A matrix doesn't have a "bottom left"...?!?

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,548 @@
...
+            widget = wibox.widget.slider,
+        },
+        direction = "east",
+        widget =  wibox.container.rotate
+    },
+    reflection = {
+        horizontal = true
+    },
+    widget = wibox.container.mirror
+}]]
```
psychon: ...????

lib/awful/widget/common.lua
```diff
@@ -84,7 +84,9 @@ function common.list_update(w, buttons, label, data, objects)
             }
         end
 
-        local text, bg, bg_image, icon = label(o, tb)
+        local text, bg, bg_image, icon, args = label(o, tb)
+        args = args or {}
```
psychon: :-(
(At least we now have this thing "meta-extensible"!)

lib/awful/widget/taglist.lua
```diff
@@ -91,15 +94,53 @@ function taglist.taglist_label(t, args)
...
+            shape = args.shape_urgent or theme.taglist_shape_urgent
+        end
+
+        if args.shape_border_width_urgent or theme.taglist_shape_border_width_urgent then
+            shape_border_width = args.shape_border_width_urgent or theme.taglist_shape_border_width_urgent
+        end
+
+        if args.shape_border_color_urgent or theme.taglist_shape_border_color_urgent then
+            shape_border_color = args.shape_border_color_urgent or theme.taglist_shape_border_color_urgent
+        end
```
psychon: These new values need to be added to the documentation. You do so for the tasklist, but apparently the taglist is forgotten (this is only how I noticed it).

docs/common/widget.ldoc
```diff
@@ -40,6 +40,12 @@
 -- @param _buttons The table of buttons that should bind to the widget.
 -- @function buttons
 
+--- Emit a signal and ensure all parent widgets in the hierarchies also
+-- forward the signal. This is useful to track signals when there is a dynamic
+-- set of containers and layouts wrapping the widget.
+-- @tparam string signal_name
+-- @param ... Other arguments
+-- @function emit_signal_recursive
```
psychon: This part is (still) in the wrong commit. It needs to be in the following commit.

lib/wibox/hierarchy.lua
```diff
@@ -51,6 +51,15 @@ local function hierarchy_new(redraw_callback, layout_callback, callback_arg)
         end
         layout_callback(result, callback_arg)
     end
+    function result._emit_recursive(_, name, ...)
+        local cur = result
```
psychon: Didn't you previously have this code in some other PR? My comments from there still apply (I'd make use of the `self` argument instead of ignoring it).

lib/wibox/drawable.lua
```diff
@@ -144,6 +144,48 @@ local function do_redraw(self)
...
+            -- bounding own anti-alising.
+            cr:set_line_width(1 + (bw % 2)/2)
+
+            cr:set_operator(cairo.Operator.CLEAR)
+            cr:set_source_rgba(0,0,0,1)
+            cr:stroke()
+
+            cr:restore()
+        end
+    end
```
psychon: I'd prefer to have this wibox-related code in `wibox.lua`. How about adding an after-draw-callback that the wibox code could use?

Then again, what does this actually do and how/why does it work? You take the bounding shape of the wibox and somehow apply it to the drawing after it was done? I do not really see how this can work or does work. Also, this CLEARs some pixels, which means that we now get some black where previously there was no black (unless a compositing manager is running...).

So... how does this work? Should it perhaps only be done when a compositing manager is running?

psychon: Actually... could we move this part (hacks(?) to make a wibox shape look nicer) out of this PR and into its own PR? I feel like I'll have to stare a lot at this code.

lib/wibox/widget/checkbox.lua
blueyed: I would argue to not use any modelines, but if so, only Vim ones - which probably can be parsed by other editors, too?!
FWIW, we have an `.editorconfig` already.

awesomeConfig.cmake
```diff
@@ -340,6 +340,13 @@ foreach(file ${AWESOME_CONFIGURE_FILES})
 endforeach()
 #}}}
 
+# {{{ Generate some aggregated documentation from lua script
+execute_process(
+        COMMAND lua ${SOURCE_DIR}/docs/06-appearance.md.lua
+        ${BUILD_DIR}/docs/06-appearance.md
```
psychon: Thanks for not doing this in CMake!

docs/06-appearance.md.lua
```diff
@@ -0,0 +1,222 @@
...
+-- Recursive file scanner
+local function get_all_files(path, ext, ret)
+    ret = ret or {}
+    local enumerator = gio.File.new_for_path(path):enumerate_children(
+        "FILE_ATTRIBUTE_STANDARD_NAME", 0, nil, nil
+    )
+
+    for file in function() return enumerator:next_file() end do
+        local file_name = file:get_attribute_as_string(name_attr)
+        local file_type = file:get_file_type()
```
psychon: Please look at `lib/menubar/utils.lua`, function `utils.parse_dir`. You have to explicitly request the file type, since otherwise some filesystems (XFS?) do not provide it. Also, instead of giving a string to `enumerate_children`, I would prefer the approach used in menubar since it catches typos (it still provides a string argument, but the string is not hardcoded but comes from gio).

docs/06-appearance.md.lua
```diff
@@ -0,0 +1,222 @@
...
+        if file_type == "REGULAR" and file_name:match(ext or "") then
+            table.insert(ret, path..file_name)
+        elseif file_type == "DIRECTORY" then
+            get_all_files(path..file_name.."/", ext, ret)
```
psychon: I don't think it matters much, but portability says you shouldn't just concatenate strings, but do what `utils.parse_dir` does: `local child_path = enumerator:get_child(file):get_path()`

docs/06-appearance.md.lua
```diff
@@ -0,0 +1,222 @@
...
+    end
+
+    f = io.open("doc/libraries/".. mod ..".html","r")
+
+    if f~=nil then
+        f:close()
+        return "../libraries/".. mod ..".html"
+    end
+
+    return ""
```
psychon: Should this `error()` out instead of silently producing broken(?) output? (Same for the previous function)

docs/06-appearance.md.lua
```diff
@@ -0,0 +1,222 @@
...
+        module_to_html(path_to_module(file)),
+        "#",
+        element,
+        "'>",
+        element:match("[. ](.+)"),
+        "</a>"
+    }
+end
+
+local all_files = get_all_files("/home/lepagee/dev/awesome/lib/", "lua")
```
psychon: ....

docs/05-awesomerc.md.lua
```diff
@@ -0,0 +1,241 @@
...
+flexibility. However, it also allows the user to write invalid code. There is
+multiple "levels" of problems:
+
+ * Syntax: There is an `awesome -k` option available in the command line to
+   check this. Awesome cannot start with an invalid `rc.lua`
+ * Invalid APIs and type errors: Lua is a dynamic language. It doesn't have much
+   support for static/compile time checks. There is the `luacheck` utility to
+   help find some categories of errors. Those errors will cause Awesome to
+   "drop" the current call stack and start over. Note that if the config cannot
+   reach the end of the `rc.lua` without errors. It will fallback to the
```
psychon: `, i` instead of `. I`

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+
+        row, column = get_next_empty(self, row, column)
+
+        base.check_widget(w)
+
+        self._private.rows[row][column] = w
+        table.insert(self._private.children, w)
+    end
+
+    self:emit_signal("widget::layout_changed")
```
psychon: Doesn't this also need `widget::redraw_needed` in case a new row was added and there is a border between widgets that has to be redrawn?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+        grid:set_row_count(row)
+    end
+
+    if column < self._private.column_count then
+        grid:set_column_count(column)
+    end
+
+    self._private.rows[row] = self._private.rows[row] or {}
+
+    self:emit_signal("widget::layout_changed")
```
psychon: This function doesn't work, does it? You need to add the widget to the row (and perhaps call check_widget())

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+    if #children > 0 then
+        self:add(unpack(children))
+    end
+end
+
+--- Remove a widget from the layout
+-- @tparam number index The widget index to remove
+-- @treturn boolean index If the operation is successful
+function grid:remove(index)
+    local TODO_TOTAL = #self:get_children()
```
psychon: Why the creative variable name?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+    self:emit_signal("widget::layout_changed")
+end
+
+--- Re-pack all the widget according to the current `row_count` and `column_count`.
+function grid:reflow()
+    local children = self:get_children()
+    self:reset()
+    if #children > 0 then
+        self:add(unpack(children))
+    end
```
psychon: How about implementing this function as `self:set_children(self:get_children())`? Seems simpler to me.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+
+    return false
+end
+
+function grid:set(index, widget2)
+    if (not widget2) or (not self._private.rows[index]) then return false end
+
+    base.check_widget(widget2)
+
+    self._private.rows[index] = widget2
```
psychon: No idea what this function should do (it is not documented), but it seems wrong. `rows[index]` should be a table containing widgets (all the widgets in that row), and not just some widget. Is there a test for this function (that also tests if the grid layout can still be drawn afterwards)?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+
+    return true
+end
+
+--- Insert a new widget in the layout at position `index`
+-- @tparam number index The position
+-- @param widget The widget
+-- @treturn boolean If the operation is successful
+function grid:insert(index, widget)
+    local TODO_TOTAL = math.huge --FIXME
```
psychon: Huh? Please fix O.o

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+    if not index or index < 1 or index > TODO_TOTAL then return false end
+
+    base.check_widget(widget)
+    table.insert(self._private.rows, index, widget) --FIXME
```
psychon: This also looks broken. Also, I lost track of the number of ways to add widgets. I would prefer if this function turned the index into row, cell numbers and then used the existing function for setting a widget based on that.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+    local max_col, max_row = get_cell_sizes(self, context, orig_width, orig_height)
+
+    -- Now that all fit is done, get the maximum
+    local used_max_h, used_max_w = 0, 0
+
+    for row_idx = 1, #max_row do
+        -- The other widgets will be discarded
+        if used_max_h + max_row[row_idx] > orig_height then
+            break
+        end
```
psychon: I don't think that the other widgets will be discarded. Instead, I read the code so that they will be "forced into the available space", won't they?

In other words, I think that this `if` and the `break` can be removed without problems (same for the next for loop; no it is "not really" a problem if `:fit` returns numbers that are larger than the available space, `base.fit_widget` will clean that up and force the result into the available space).

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+        used_max_w = used_max_w + max_col[col_idx]
+    end
+
+    return used_max_w, used_max_h
+end
+
+function grid:reset()
+    self._private.rows = {}
+    self._private.children = {}
+    self:emit_signal("widget::layout_changed")
```
psychon: Also: Redraw needed, due to the borders (if there are any)

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+--- Set the layout's fill_space property. If true, the remaining space will
+-- be distributed equally across all widgets in the grid.
+-- @DOC_wibox_layout_grid_fill_space_EXAMPLE@
+-- @property fill_space
+-- @param boolean
+
+function grid:set_fill_space(val)
+    if self._private.fill_space ~= val then
+        self._private.fill_space = not not val
+        self:emit_signal("widget::layout_changed")
```
psychon: I think that this property also influences the border between widgets, so: `redraw_needed`

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+    end
+end
+
+--- The the border (grid) width.
```
psychon: The the

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+
+--- Set the layout's fill_space property. If true, the remaining space will
+-- be distributed equally across all widgets in the grid.
+-- @DOC_wibox_layout_grid_fill_space_EXAMPLE@
+-- @property fill_space
+-- @param boolean
+
+function grid:set_fill_space(val)
+    if self._private.fill_space ~= val then
+        self._private.fill_space = not not val
```
psychon: Shouldn't you do the "not not dance" before the `~=`? Not that it matters much, just saying...
So something like:
```
val = not not val
if self._private.fill_space == val then return end
self._private.fill_space = val
etc
```

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+-- @DOC_wibox_layout_grid_border_EXAMPLE@
+-- @property border_width
+-- @param number
+
+function grid:set_border_width(val)
+    if self._private.border_width ~= val then
+        self._private.border_width = math.max(0, val)
+
+        -- This count as spacing, so the layout really change
+        self:emit_signal("widget::layout_changed")
```
psychon: Redraw needed?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+    ret._private.column_count = 1
+    ret:set_spacing(0)
+    ret:set_fill_space(false)
+
+    if widget1 then
+        ret:add(widget1, ...)
+    end
+
+    return ret
+end
```
psychon: Could you move this function to the end of the file?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,484 @@
...
+-- @param color
+-- @see gears.color
+
+function grid:set_border_color(val)
+    self._private.border_color = color(val)
+
+    self:emit_signal("widget::redraw_needed")
+end
+
+local function get_layout(dir, widget1, ...)
```
psychon: The `dir` argument should go away, this will be the first widget.

lib/wibox/container/align.lua
```diff
@@ -0,0 +1,158 @@
...
+local align_fct = {
+    left   = function(_  , _   ) return 0                         end,
+    center = function(wdg, orig) return math.max(0, (orig-wdg)/2) end,
+    right  = function(wdg, orig) return math.max(0, orig-wdg    ) end,
+}
+align_fct.top, align_fct.bottom = align_fct.left, align_fct.right
+
+-- Layout this layout
+function align:layout(context, width, height)
+    if not self._private.widget or not self._private.widget._private.visible then
```
psychon: I don't think you should check `visible` here. One reason is because it won't make a difference on the visual result. The second reason is "I don't think things will be properly redrawn if the child widget is made visible" (because the widget code doesn't know that this code tries to display the widget). The third reason is "because you are using other code's `_private`".

lib/wibox/container/align.lua
```diff
@@ -0,0 +1,158 @@
...
+--- Replace the layout children
+-- This layout only accept one children, all others will be ignored
+-- @tparam table children A table composed of valid widgets
+function align:set_children(children)
+    self:set_widget(children[1])
+end
+
+--- Reset this layout. The widget will be removed and the rotation reset.
+function align:reset()
+    self._private.direction = nil
```
psychon: There is no direction, is there? Instead this should reset valign and halign.

lib/wibox/container/align.lua
```diff
@@ -0,0 +1,158 @@
...
+    self:set_widget(nil)
+end
+
+--- Get the direction of this rotating layout
+function align:get_direction()
+    return self._private.direction or "north"
+end
```
psychon: Remove this whole function

lib/wibox/container/align.lua
```diff
@@ -0,0 +1,158 @@
...
+    util.table.crush(ret, align, true)
+
+    ret:set_widget(widget)
+    ret:set_halign(halign)
+    ret:set_valign(valign)
+
+    return ret
+end
+
+function align.mt:__call(...)
```
psychon: I think this needs `_, ` as a first argument

lib/wibox/widget/graph.lua
```diff
@@ -142,19 +168,43 @@ function graph.draw(_graph, _, cr, width, height)
...
-            cr:stroke()
+
+            if step_shape or step_width > 1 then
+                cr:fill()
+
+                -- Undo all the translates
+                if step_shape then
+                    local mat = cr:get_matrix()
+                    mat.x0 = x0
+                    cr:set_matrix(mat)
```
psychon: Uhm. Could you somehow use `:save()` / `:restore()` instead? For one thing: Because your code wont work inside of a rotating widget. For another thing: You aren't undoing *all*  the translations, but only the translation in x direction.

lib/wibox/widget/graph.lua
```diff
@@ -142,19 +168,43 @@ function graph.draw(_graph, _, cr, width, height)
...
+                    local x = i*step_width + ((i-1)*step_spacing) + 0.5
                     value = (value - min_value) / max_value
-                    cr:move_to(i + 0.5, height * (1 - value))
-                    cr:line_to(i + 0.5, height)
+                    cr:move_to(x, height * (1 - value))
+
+                    if step_shape then
+                        cr:translate(step_width + (i>1 and step_spacing or 0), height * (1 - value))
+                        step_shape(cr, step_width, height)
+                        cr:translate(nil, -(height * (1 - value)))
```
psychon: I *bet*  that lgi / cairo do not let you call this function with a `nil` argument. This should likely be 0 and needs some test which catches this breakage.

lib/wibox/container/align.lua
```diff
@@ -0,0 +1,158 @@
+---------------------------------------------------------------------------
+-- A container used to align smaller widgets into larger space.
```
psychon: This gives us a container called align and a layout called align, making the name "align" ambiguous. This needs another name (but I don't have any good suggestions, sorry).

lib/naughty/core.lua
```diff
@@ -107,14 +107,20 @@ naughty.config.presets = {
 -- @tfield[opt=apply_dpi(5)] int margin
 -- @tfield[opt=apply_dpi(1)] int border_width
 -- @tfield[opt="top_right"] string position
+-- @tfield[opt] gears.color border_color
+-- @tparam[opt] gears.shape shape Widget shape.
+-- @tparam[opt] table shape_args Arguments passed to widget shape function.
```
psychon: Why this mix of tfield and tparam? I think all of this should be tfield...?

awesomeConfig.cmake
```diff
@@ -347,7 +347,7 @@ execute_process(
 
 execute_process(
         COMMAND lua ${SOURCE_DIR}/docs/05-awesomerc.md.lua
-        ${BUILD_DIR}/05-awesomerc.md.lua ${SOURCE_DIR}/awesomerc.lua
+        ${BUILD_DIR}/docs/05-awesomerc.md ${SOURCE_DIR}/awesomerc.lua
```
psychon: This looks like a change/fix that should be squashed into the original commit instead. Same for the other changes in this commit?

docs/06-appearance.md.lua
```diff
@@ -10,7 +10,7 @@ local name_attr = gio.FILE_ATTRIBUTE_STANDARD_NAME
 -- Recursive file scanner
 local function get_all_files(path, ext, ret)
     ret = ret or {}
-    local enumerator = gio.File.new_for_path(path):enumerate_children(
+    local enumerator, _ = gio.File.new_for_path(path):enumerate_children(
```
psychon: ...why?

psychon:
I read through this up to 2b3c2d746979c86a2d23e46ce12f426224a0ccbe (just so that I know which version of this I looked at). I know that this is still titled WIP, but I took a look anyway.

lib/wibox/widget/checkbox.lua
psychon: Removing vim modelines would also be fine with me if many people are against them. We only have them because JD added them "Back Then" (tm).

lib/wibox/widget/graph.lua
Elv13: bet taken :P but I will change it anyway.

lib/wibox/layout/grid.lua
Elv13: I guess you saw what I used for base at this point. I should cleanup my template.

lib/wibox/widget/graph.lua
psychon: Wow, you are right. That's a surprise. I always saw lgi as being quite strict...

lib/wibox/widget/checkbox.lua
Elv13: No, the problem is when the background shape and check shape isn't the same, it creates a mess as different shapes resize themselves differently. Being a square force some kind of linearity, so the result is predictable(-ish).

Elv13:
Ok, I addressed 5-6 of @psychon comments. I fixed all luachecks, made it work with lua5.3 and improved the `grid` layout to fix some more broken corners cases (and wrote the tests for them). This is a longterm-ish PR, so don't expect all issues to be fixed by tomorrow. However, it is progressing well. And the 93% coverage is getting nice too! For some reason Codecov is missing the integration test coverage (`test-wibox.lua`). By the "old" way codecov worked before being broken(?) with a commit yesterday, I did hit 75% (and yes, it is because it include the tests themselves in the percentage, so adding 1.5k lines of tests did artificially boost the perceived coverage).

lib/wibox/widget/checkbox.lua
psychon: I'd call that a bug in the shape (assuming they also use math.min then the shape should be fixed not to do that), but ok...

lib/wibox/widget/slider.lua
Elv13: Not since you remove the matrix `:copy()`, even if it's not the cleanest way to get a new matrix to play with. I could revert the revert that deleted the `:copy()` function.

lib/awful/widget/taglist.lua
Elv13: This is documented in a later commit along with some other undocumented things. Mind if it stays that way?

Elv13:
Ok, I addressed most issues above (I think).

The slider need to work I guess. I think I will remove from the PR is @psychon think it still suck too much. 

@actionless Do you mind if your patches go into a new PR instead of this one along with the rounded corner one? You can open that pull request yourself so you can work on it.

I removed the [WIP] tag. I guess it will need another row of review, but I think it is close enough to be merged. About the "new" 1.1K lines since the last update, they are all tests or doc.

lib/wibox/widget/slider.lua
psychon: Why do you need a new matrix to play with? The idea is that matrixes are immutable (= cannot be modified), so there is no reason to copy them. (And this means that modifying a matrix does not accidentally influence some other code that happens to reference the same matrix)

lib/awful/widget/taglist.lua
psychon: Ok, fine with me...

psychon:
One of the travis build segfaulted in `cairo_region_union_rectangle`. Scary, but I can't see how this PR could cause that. Hm... :-(

> The slider need to work I guess. I think I will remove from the PR is @psychon think it still suck too much. 

Without having looked at the slider widget:
This PR is titled "shape api" and adds a slider widget.

> I guess it will need another row of review,

I'll try to find some time to go through this on Sunday, but I cannot promise yet.

actionless:
@Elv13 yup, i don't mind, anyway i am using my fork for building the local package instead of upstream repo

Elv13:
> This PR is titled "shape api" and adds a slider widget.

The PR add widgets using the shape API and retrofit existing ones to support it. Beside all the documentation stuff (that was kind of necessary to be able to properly document those widgets), the only unrelated code is the grid layout (as said in the first message).

lib/wibox/container/radialprogressbar.lua
```diff
@@ -0,0 +1,264 @@
...
+end
+
+-- Set the clip
+function radialprogressbar:before_draw_children(_, cr, width, height)
+    cr:save()
+    local wa = content_workarea(self, width, height)
+    cr:translate(wa.x, wa.y)
+    shape.rounded_bar(cr, wa.width, wa.height)
+    cr:clip()
+    cr:translate(-wa.x, -wa.y)
```
psychon: I don't think this works correctly. The `wibox.hierarchy` now has a wrong idea on where the child widget is and so partial redraws will not redraw the correct part of the screen.

However, your `:layout()` already applies a translation (in fact, the opposite translation) to the child widget, so why is this needed at all?

lib/wibox/container/radialprogressbar.lua
```diff
@@ -0,0 +1,264 @@
...
+            self._private.widget, wa.x, wa.y, wa.width, wa.height
+        ) }
+    end
+end
+
+-- Fit this layout into the given area
+function radialprogressbar:fit(context, width, height)
+    if self._private.widget then
+        local wa = content_workarea(self, width, height)
+        return base.fit_widget(self, context, self._private.widget, wa.width, wa.height)
```
psychon: I'm not totally sure, but I think this should be:
```lua
local w, h = base.fit_widget(self, context, self._private.widget, wa.width, wa.height)
return wa.x + w, wa.y + h
```

lib/wibox/container/radialprogressbar.lua
```diff
@@ -0,0 +1,264 @@
...
+--- Reset this container.
+function radialprogressbar:reset()
+    self:set_widget(nil)
+end
+
+for _,v in ipairs {"left", "right", "top", "bottom"} do
+    radialprogressbar["set_"..v.."_padding"] = function(self, val)
+        self._private.paddings = self._private.paddings or {}
+        self._private.paddings[v] = val
+        self:emit_signal("widget::redraw_needed")
```
psychon: Also `widget::layout_changed`, because the `content_workarea()` depends on the padding and is used in `:layout()`

lib/wibox/container/radialprogressbar.lua
```diff
@@ -0,0 +1,264 @@
...
+    util.table.crush(ret, radialprogressbar)
+    ret._private.max_value = 1
+    ret._private.min_value = 0
+
+    ret:set_widget(widget)
+
+    return ret
+end
+
+function radialprogressbar.mt:__call(...)
```
psychon: This should be `_, ...` instead of `...` since the first argument is just `wibox.container.radialprogressbar`

lib/wibox/widget/checkbox.lua
```diff
@@ -0,0 +1,252 @@
...
+    wa.x      = offset + wa.x + (padding.left or 1)
+    wa.y      = offset + wa.y + (padding.top  or 1)
+    wa.width  = wa.width  - (padding.left or 1) - (padding.right  or 1) - 2*offset
+    wa.height = wa.height - (padding.top  or 1) - (padding.bottom or 1) - 2*offset
+
+    return wa
+end
+
+local function draw(self, _, cr, width, height)
+    local size = math.min(width, height)
```
psychon: This isn't needed (at least in `draw`), because this only ever uses `size, size` which can easily be replaced with `width, height`. Only `fit` needs this. And yes, I know that you disagree.

lib/wibox/widget/piechart.lua
```diff
@@ -0,0 +1,212 @@
...
+local pie       = require( "gears.shape"       ).pie
+local unpack    = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
+
+local module = {}
+
+local piechart = {}
+
+local function draw_label(cr,angle,radius,center_x,center_y,text)
+    cr:move_to(center_x+(radius/2)*math.cos(angle),center_y+(radius/2)*math.sin(angle))
+    cr:line_to(center_x+(1.5*radius)*math.cos(angle),center_y+(1.5*radius)*math.sin(angle))
```
psychon: Perhaps `rel_line_to` helps here? The above `line_to` can be replaced with (at least I think so):
```lua
cr:rel_line_to(radius*math.cos(angle), radius*math.sin(angle))
```

lib/wibox/widget/piechart.lua
```diff
@@ -0,0 +1,212 @@
...
+    local x,y = cr:get_current_point()
+    cr:line_to(x+(x>center_x and radius/2 or -radius/2),y)
```
psychon:
```lua
cr:rel_line_to(x > center_x and radius/2 or -radius/2, 0)
```

lib/wibox/widget/piechart.lua
```diff
@@ -0,0 +1,212 @@
...
+    local extents
+
+    -- Support older LGI versions
+    if type(cr.font_extents) == "function" then
+        extents = cr:font_extents()
+    else
+        extents = cr.font_extents
+    end
```
psychon: I would suggest:
```lua
local extents = cr.font_extents
-- Support older LGI versions
if type(extents) == "function" then
  extents = extents(cr)
end

File: lib/wibox/widget/piechart.lua [hunk 68]
```diff
@@ -0,0 +1,212 @@
...
+
+    cr:move_to(x+(x>center_x and radius/2 + 5 or (-radius/2 - cr:text_extents(text).width - 5)),y+(extents.height/4))
```
psychon: I'd like to have more explanation for this. Why 5? (likely just some distance between the text and the line, perhaps replacing this with a space character in the text?) Why `extents.height/4`? This feels like some magic numbers that hide some incorrectness.

lib/wibox/widget/piechart.lua
```diff
@@ -0,0 +1,212 @@
...
+
+    return ret
+end
+
+--@DOC_widget_COMMON@
+
+--@DOC_object_COMMON@
+
+return setmetatable(module, { __call = function(_, ...) return new(...) end })
+-- kate: space-indent on; indent-width 4; replace-tabs on;
```
psychon: Nice try, but I saw it :-)

lib/gears/shape.lua
```diff
@@ -331,6 +331,7 @@ end
 -- @tparam[opt=math.min(width  height) / 2)] number radius The radius
 function module.circle(cr, width, height, radius)
     radius = radius or math.min(width, height) / 2
+    cr:move_to(width/2+radius, height/2)
```
psychon: What happens without this? Looking at cairo's implementation (`_cairo_default_context_arc()`), the first thing that it does is a `line_to` to the start point of the arc. If you just want to get rid of this `line_to`, I would suggest `cr:new_sub_path()` instead (easier to read/understand).

lib/wibox/container/arcchart.lua
```diff
@@ -0,0 +1,343 @@
...
+--- Reset this layout. The widget will be removed and the rotation reset.
+function arcchart:reset()
+    self:set_widget(nil)
+end
+
+for _,v in ipairs {"left", "right", "top", "bottom"} do
+    arcchart["set_"..v.."_padding"] = function(self, val)
+        self._private.paddings = self._private.paddings or {}
+        self._private.paddings[v] = val
+        self:emit_signal("widget::redraw_needed")
```
psychon: Doesn't this also need `layout_changed`?

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,528 @@
...
+    return width, height
+end
+
+local event = {}
+
+local rect_order = {"x", "y", "width", "height"}
+
+-- Move the handle to the correct location
+local function move_handle(self, x, y, geo)
+    -- Fix the X position when used in a rotate or mirror widget
```
psychon: Uhm, I'll take this as a bug report. x and y should already be in the widget's own space, but I see that the code in `drawable` for this is wrong (completely ignoring rotations). Do you want to fix this or should I do? Something like the following might work (untested):
```lua
index 0212efa..6b450bb 100644
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -376,8 +376,7 @@ function drawable.new(d, widget_context_skeleton, drawable_name)
             local widgets = ret:find_widgets(x, y)
             for _, v in pairs(widgets) do
                 -- Calculate x/y inside of the widget
-                local lx = x - v.x
-                local ly = y - v.y
+                local lx, ly = v.matrix_to_device:invert():transform_point(x, y)
                 v.widget:emit_signal(name, lx, ly, button, modifiers,v)
             end
         end)
```
(find_widgets should just be taught to also return the matrix_from_device so that this `invert()`-call isn't needed, but whatever...)

lib/wibox/widget/slider.lua
psychon: ...??????

psychon:
I read up to f4b32faac45b268d998ccd149e2a68a3dc4157e5 and now ran out of review-time. Sorry, but could this be split up into at least one other PR? For example, the new shapes and widgets seem relatively independent from the docs-related changes. Simplest way would be to just open a separate PR for half of the commits here so that this doesn't cause much work for you.
Thanks & sorry.

actionless:
i am trying to get this (https://github.com/awesomeWM/awesome/pull/1044/commits/0ef77e268fc130b26df93263f53b142fe877af83) working on master branch again, could you hint me which parts of this changeset are currently missing for that?

Elv13:
The client shape commit, just about everything else is merged

actionless:
weird, when i am re-applying naughty-shape commit on top of master i don't see any warnings but notification shape stays rectangular

also in documentation wibox do not have `shape_border_width`, `shape` and `shape_args`: http://new.awesomewm.org/apidoc/classes/wibox.html

Elv13:
Let me rephrase myself. Almost have commits have been merged *at the exception* of the mentioned wibox shape commit. It has not been merged because it needs work. Work I would like you to do if you wish to get your notification commit merged. It isn't much, but it is delicate and I already spent enough time getting the thing to "work".

actionless:
that's why i was asking you to hint me from where to start :)

Elv13:
Look at the comments above (in the PR) related to this commit, that's
pretty much it ;)

Elv13:
this PR was finally fully merged in 4.1

Issue #821, opened 2016-04-11
Build error


hari-rangarajan:
/home/hari/tmpbuilds/awesome1/awesome/tests/examples/awful/placement/align.lua
  A test failed, bye
Call Stack (most recent call first):
  tests/examples/CMakeLists.txt:243 (run_test)
  tests/examples/CMakeLists.txt:235 (digg)
  tests/examples/CMakeLists.txt:235 (digg)
  tests/examples/CMakeLists.txt:264 (digg)
  awesomeConfig.cmake:306 (include)
  CMakeLists.txt:17 (include)

Elv13:
Can you add more lines please?

hari-rangarajan:
This is the entire log. Please let me know if you need specific details from the logs. BTw, 3.5.9 builds fine on the same machine. The error is with the latest pull.

Running cmake…
-- git -> /usr/bin/git
-- asciidoc -> /usr/bin/asciidoc
-- xmlto -> /usr/bin/xmlto
-- gzip -> /bin/gzip
-- ldoc -> /usr/bin/ldoc
-- convert -> /usr/bin/convert
-- checking for execinfo -- found
-- checking for __builtin_clz -- yes
-- Testing code based on _awful

2016-04-11 16:07:53 W: Cannot reliably detect EOF on an GIOInputStream with this LGI version
Fontconfig error: "~/.fonts.conf", line 1: XML or text declaration not at start of entity
Fontconfig warning: "/etc/fonts/conf.d/50-user.conf", line 14: reading configurations from ~/.fonts.conf is deprecated. please move it to /home/hari/.config/fontconfig/fonts.conf manually

CMake Error at tests/examples/CMakeLists.txt:162 (message):
  
  /home/hari/tmpbuilds/awesome1/awesome/tests/examples/awful/placement/align.lua
  A test failed, bye
Call Stack (most recent call first):
  tests/examples/CMakeLists.txt:243 (run_test)
  tests/examples/CMakeLists.txt:235 (digg)
  tests/examples/CMakeLists.txt:235 (digg)
  tests/examples/CMakeLists.txt:264 (digg)
  awesomeConfig.cmake:306 (include)
  CMakeLists.txt:17 (include)


-- Configuring incomplete, errors occurred!
See also "/home/hari/tmpbuilds/awesome1/awesome/.build-lini-me-x86_64-linux-gnu-5.2.1/CMakeFiles/CMakeOutput.log".
[...]

PR #1045, opened 2016-08-20
Be less strict about warnings from "example tests"

psychon:
The build is no longer aborted when one of the "example tests" produces a
message on stderr. However, on Travis this requirement is still made. This
should catch "bad errors" via Travis while not breaking the build for users.

Fixes: https://github.com/awesomeWM/awesome/issues/821


blueyed:
LGTM, thanks!

PR #1051, opened 2016-08-22
tag: Fix index calculation.

Elv13:
The index was updated on an unordered table. As the elements
order did not match the relative indices once they have been
changed, further calls to set_index produced garbage.

The default taglist didn't notice because it use screen.tags
table index instead of the tag index. A debug using

echo 'for _,t in ipairs(mouse.screen.tags) do
  print("INDEX:", _, t.index, t.name) end' | awesome-client

Would have shown two or more elements with the same index. To
debug issues related to tag indices, this bash script can be
enabled:

while true; do
  echo 'for _,t in ipairs(mouse.screen.tags) do
  assert( _==t.index) end' | awesome-client
  sleep 0.5
done

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -74,6 +74,12 @@ function tag.object.set_index(self, idx)
     -- screen.tags cannot be used as it depend on index
     local tmp_tags = raw_tags(scr)
 
+    -- sort the tags by index
+    table.sort(tmp_tags, function(a, b)
+        local ia, ib = tag.getproperty(a, "index"), tag.getproperty(b, "index")
+        return (ia or math.huge) < (ib or math.huge)
+    end)
+
     if (not idx) or (idx < 1) or (idx > #tmp_tags) then
         return
     end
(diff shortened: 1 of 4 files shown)


Elv13:
The build is broken, see https://github.com/stevedonovan/LDoc/issues/248

@blueyed Can we stop using random luarocks packages. The whole idea of these unreliable package manager is bad. Be it luarocks, pip, npm, AUR, brew or anything else. If Ubuntu doesn't have recent enough version, we should consider adding a PPA to Travis. That way we would stop having broken builds because someone on the internet had a bad day?

blueyed:
@Elv13 
We could pin versions with luarocks also.

Elv13:
Ok, the ldoc guy pushed a new released too, but it is also broken. Can I merge this? I think @psychon is away from keyboard for the week and this quite an obvious bug.

lib/awful/tag.lua
```diff
@@ -74,6 +74,12 @@ function tag.object.set_index(self, idx)
     -- screen.tags cannot be used as it depend on index
     local tmp_tags = raw_tags(scr)
 
+    -- sort the tags by index
+    table.sort(tmp_tags, function(a, b)
+        local ia, ib = tag.getproperty(a, "index"), tag.getproperty(b, "index")
+        return (ia or 9999) < (ib or 9999)
```
blueyed: What about `math.huge` instead of 9999?

blueyed:
Thanks, @Elv13!
Do you think it's possible to have a test for this easily?

Elv13:
Here you go

Elv13:
@blueyed Beside the coverage being wrong, it's ready. If I put `print()` and run the tests, it is being reached. So it is just one of those codecov issue we had lately.

blueyed:
Thanks!
:+1:

PR #1055, opened 2016-08-27
client: Fix a potential crash due to a race condition.

Elv13:
If raise was called from the awful.rule `manage` callback at
startup with a client already present it crashed due to a
buffer underflow.


blueyed:
Thanks! :)

Issue #1028, opened 2016-08-02
Spawning a client with callback properties on screen 1 will tag it on screen 2


blueyed:
The following code, run through `awesome-client` on screen 1 (internal monitor) will spawn rxvt-unicode on that screen, but it gets tagged with the selected tag on screen 2 (external display):

    require("awful").spawn('urxvt', {})

The following works as a workaround, but should probably be the default behavior:

    require("awful").spawn('urxvt', {floating=true, screen=require("awful").screen.focused})

For programs launched without properties `c:tags()` seems to be empty when `rules.execute`

I've attached to signals for debugging:

    client.connect_signal("tagged", tag_screen_debug)
    client.connect_signal("property::screen", tag_screen_debug)

This is the output:
```
Client got tagged:
c.screen: 2, awful.tag.getscreen(t): 2, c window/client(urxvt): 0x418ff378, c.name urxvt, t tag: 0x40bcc3b0
stack traceback:
        /home/daniel/.config/awesome/rc.lua:2231: in function </home/daniel/.config/awesome/rc.lua:2227>
        [C]: in function 'tags'
        /usr/share/awesome/lib/awful/client.lua:461: in function 'to_selected_tags'
        /usr/share/awesome/lib/awful/ewmh.lua:138: in function </usr/share/awesome/lib/awful/ewmh.lua:133>
        [C]: in function 'emit_signal'
        /usr/share/awesome/lib/awful/rules.lua:432: in function 'execute'
        /usr/share/awesome/lib/awful/rules.lua:497: in function </usr/share/awesome/lib/awful/rules.lua:496>
        [C]: in function 'emit_signal'
        /usr/share/awesome/lib/awful/spawn.lua:76: in function </usr/share/awesome/lib/awful/spawn.lua:73>
Client got tagged:
c.screen: 1, awful.tag.getscreen(t): nil, c window/client(urxvt): 0x418ff378, c.name urxvt, t screen: 0x402e1698
stack traceback:
        /home/daniel/.config/awesome/rc.lua:2231: in function </home/daniel/.config/awesome/rc.lua:2227>
        [C]: in function '__newindex'
        /usr/share/awesome/lib/awful/tag.lua:1284: in function </usr/share/awesome/lib/awful/tag.lua:1273>
```

Apparently the tag gets set to screen 2 from the beginning, via `to_selected_tags`.

Will have to investigate further.

psychon:
Sigh.

So:
[...]

PR #1056, opened 2016-08-28
Test multiple screens on the CI + fixes

Elv13:
**Edit:** This PR introduce multi-screen testing because I accidentally broke this enough time already. Time for more "deterministic" testing.

**Original:** There was still a problem that caused the "old" tags to be
inserted in the wrong position when "saved" from a screen being
removed.

Also, this use a :get_tags(true) to save an uneeded sorting pass.

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -12,6 +12,7 @@ local util = require("awful.util")
 local ascreen = require("awful.screen")
 local beautiful = require("beautiful")
 local object = require("gears.object")
+local timer = require("gears.timer")
 local pairs = pairs
 local ipairs = ipairs
 local table = table
@@ -1277,21 +1278,25 @@ function tag.attached_connect_signal(screen, ...)
 end
 
 -- Register standard signals.
-capi.client.connect_signal("manage", function(c)
-    -- If we are not managing this application at startup,
-    -- move it to the screen where the mouse is.
-    -- We only do it for "normal" windows (i.e. no dock, etc).
-    if not awesome.startup and c.type ~= "desktop" and c.type ~= "dock" then
-        if c.transient_for then
-            c.screen = c.transient_for.screen
-            if not c.sticky then
-                c:tags(c.transient_for:tags())
+capi.client.connect_signal("property::screen", function(c)
+    -- First, the delayed timer is necessary to avoid a race condition with
+    -- awful.rules. It is also messing up the tags before the user have a chance
+    -- to set them manually.
+    timer.delayed_call(function()
+        local tags, new_tags = c:tags(), {}
+
+        for _, t in ipairs(tags) do
+            if t.screen == c.screen then
+                table.insert(new_tags, t)
             end
-        else
-            c.screen = ascreen.focused()
         end
-    end
-    c:connect_signal("property::screen", function() c:to_selected_tags() end)
+
+        if #new_tags == 0 then
+            c:emit_signal("request::tag", nil, {reason="screen"})
+        elseif #new_tags < #tags then
+            c:tags(new_tags)
+        end
+    end)
 end)
 
 -- Keep track of the number of urgent clients.
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -465,6 +465,7 @@ function tag.object.set_screen(t, s)
 
     -- Change the screen
     tag.setproperty(t, "screen", s)
+    tag.setproperty(t, "index", #s:get_tags(true))
 
     -- Make sure the client's screen matches its tags
     for _,c in ipairs(t:clients()) do
@@ -473,10 +474,8 @@ function tag.object.set_screen(t, s)
     end
 
     -- Update all indexes
-    for _,screen in ipairs {old_screen, s} do
-        for i,t2 in ipairs(screen.tags) do
-            tag.setproperty(t2, "index", i)
-        end
+    for i,t2 in ipairs(old_screen.tags) do
+        tag.setproperty(t2, "index", i)
     end
 
     -- Restore the old screen history if the tag was selected
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -312,7 +312,7 @@ function tag.object.delete(self, fallback_tag, force)
     -- view one. The > 1 is because ntags is no longer synchronized with the
     -- current count.
     if target_scr.selected_tag == nil and ntags > 1 then
-        tag.history.restore(nil, 1)
+        tag.history.restore(target_scr, 1)
         if target_scr.selected_tag == nil then
             local other_tag = tags[tags[1] == self and 2 or 1]
             if other_tag then
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -259,15 +259,14 @@ end
 --  stickied tags to.
 -- @tparam[opt=false] boolean force Move even non-sticky clients to the fallback
 -- tag.
--- @return Returns true if the tag is successfully deleted, nil otherwise.
+-- @return Returns true if the tag is successfully deleted.
 -- If there are no clients exclusively on this tag then delete it. Any
 -- stickied clients are assigned to the optional 'fallback_tag'.
 -- If after deleting the tag there is no selected tag, try and restore from
 -- history or select the first tag on the screen.
 function tag.object.delete(self, fallback_tag, force)
-
     -- abort if the taf isn't currently activated
-    if not self.activated then return end
+    if not self.activated then return false end
 
     local target_scr = get_screen(tag.getproperty(self, "screen"))
     local tags       = target_scr.tags
@@ -275,7 +274,7 @@ function tag.object.delete(self, fallback_tag, force)
     local ntags      = #tags
 
     -- We can't use the target tag as a fallback.
-    if fallback_tag == self then return end
+    if fallback_tag == self then return false end
 
     -- No fallback_tag provided, try and get one.
     if fallback_tag == nil then
@@ -284,7 +283,7 @@ function tag.object.delete(self, fallback_tag, force)
 
     -- Abort if we would have un-tagged clients.
     local clients = self:clients()
-    if ( #clients > 0 and ntags <= 1 ) or fallback_tag == nil then return end
+    if #clients > 0 and fallback_tag == nil then return false end
 
     -- Move the clients we can off of this tag.
     for _, c in pairs(clients) do
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -308,8 +308,10 @@ function tag.object.delete(self, fallback_tag, force)
         tag.setproperty(tags[i], "index", i-1)
     end
 
-    -- If no tags are visible, try and view one.
-    if target_scr.selected_tag == nil and ntags > 0 then
+    -- If no tags are visible (and we did not delete the lasttag), try and
+    -- view one. The > 1 is because ntags is no longer synchronized with the
+    -- current count.
+    if target_scr.selected_tag == nil and ntags > 1 then
         tag.history.restore(nil, 1)
         if target_scr.selected_tag == nil then
             local other_tag = tags[tags[1] == self and 2 or 1]
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -59,6 +59,8 @@ local function get_widget_context(self)
 end
 
 local function do_redraw(self)
+    if type(self.drawable) ~= "drawable" then return end --FIXME See #1070
+
     local surf = surface.load_silently(self.drawable.surface, false)
     -- The surface can be nil if the drawable's parent was already finalized
     if not surf then return end
--- a/objects/client.c
+++ b/objects/client.c
@@ -2304,6 +2304,10 @@ luaA_client_swap(lua_State *L)
  *
  * Use the `first_tag` field to access the first tag of a client directly.
  *
+ * **Signal:**
+ *
+ *  * *property::tags*
+ *
  * @tparam table tags_table A table with tags to set, or `nil` to get the
  *   current tags.
  * @treturn table A table with all tags.
@@ -2342,7 +2346,10 @@ luaA_client_tags(lua_State *L)
         lua_pushnil(L);
         while(lua_next(L, 2))
             tag_client(L, c);
+
         lua_pop(L, 1);
+
+        luaA_object_emit_signal(L, -1, "property::tags", 0);
     }
 
     lua_newtable(L);
--- a/tests/_multi_screen.lua
+++ b/tests/_multi_screen.lua
@@ -1,0 +1,512 @@
+-- Helper module to replicate a set of steps across multiple screen scenarios
+-- Ideas for improvements:
+--  * Export an iterator so the suits can call the screen change step by step
+--    themselves instead of the current all or nothing
+--  * Offer multiple scenarios such as screen add, remove, resize/swap instead
+--    of having to test all corner cases all the time.
+local wibox = require("wibox")
+local surface = require("gears.surface")
+local wallpaper = require("gears.wallpaper")
+local color = require("gears.color")
+local shape = require("gears.shape")
+
+local module = {}
+
+-- Get the original canvas size before it is "lost"
+local canvas_w, canvas_h = root.size()
+
+local half_w, half_h = math.floor(canvas_w/2), math.floor(canvas_h/2)
+local quarter_w = math.floor(canvas_w/4)
+
+-- Create virtual screens.
+-- The steps will be executed on each set of screens.
+local dispositions = {
+    -- A simple classic, 2 identical screens side by side
+    {
+        function()
+            return {
+                x = 0,
+                y = 0,
+                width  = half_w,
+                height = canvas_h,
+            }
+        end,
+        function()
+            return {
+                x = half_w,
+                y = 0,
+                width = half_w,
+                height = canvas_h,
+            }
+        end,
+    },
+
+    -- Take the previous disposition and swap the screens
+    {
+        function()
+            return {
+                x = half_w,
+                y = 0,
+                width = half_w,
+                height = canvas_h,
+            }
+        end,
+        function()
+            return {
+                x = 0,
+                y = 0,
+                width  = half_w,
+                height = canvas_h,
+            }
+        end,
+        keep = true,
+    },
+
+    -- Take the previous disposition and resize the leftmost one
+    {
+        function()
+            return {
+                x = quarter_w,
+                y = 0,
+                width = 3*quarter_w,
+                height = canvas_h,
+            }
+        end,
+        function()
+            return {
+                x = 0,
+                y = 0,
+                width  = quarter_w,
+                height =  half_h,
+            }
+        end,
+        keep = true,
+    },
+
+    -- Take the previous disposition and remove the leftmost screen
+    {
+        function()
+            return {
+                x = quarter_w,
+                y = 0,
+                width = 3*quarter_w,
+                height = canvas_h,
+            }
+        end,
+        function()
+            return nil
+        end,
+        keep = true,
+    },
+
+    -- Less used, but still quite common vertical setup
+    {
+        function()
+            return {
+                x = 0,
+                y = 0,
+                width  = canvas_w,
+                height = half_h,
+            }
+        end,
+        function()
+            return {
+                x = 0,
+                y = half_h,
+                width  = canvas_w,
+                height = half_h,
+            }
+        end
+    },
+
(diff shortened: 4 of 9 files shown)


Elv13:
@psychon:

    2016-08-29 04:03:02 E: Error during a protected call: lib//wibox/drawable.lua:62: attempt to index field 'drawable' (a userdata value)
    2016-08-29 04:03:02 E: Error during a protected call: lib//wibox/drawable.lua:62: attempt to index field 'drawable' (a userdata value)

I can't reproduce this locally. It looks like some kind of messed up wibox without its `__index` metatable. Any idea how that's happening? I used `surface.widget_to_surface()` in the test. But I don't think its related. It may be something that happens when you remove too many screens (their drawin get GCed, but not the wibox)?

@blueyed This should solve the #1028 bug. Please test.

lib/awful/tag.lua
```diff
@@ -465,6 +465,7 @@ function tag.object.set_screen(t, s)
 
     -- Change the screen
     tag.setproperty(t, "screen", s)
+    tag.setproperty(t, "index", #s:get_tags(true))
```
psychon: I'm not sure if this matters much, but perhaps the index should be updated before the screen is changed so that the index is already correct when `property::screen` occurs?

lib/awful/tag.lua
```diff
@@ -284,7 +283,7 @@ function tag.object.delete(self, fallback_tag, force)
 
     -- Abort if we would have un-tagged clients.
     local clients = self:clients()
-    if ( #clients > 0 and ntags <= 1 ) or fallback_tag == nil then return end
+    if #clients > 0 and fallback_tag == nil then return false end
```
psychon: The old logic was "don't allow deleting the last tag of a screen". I think that's quite a good idea since screens without tags will surely cause many problems.

How about making this also depend on the `force` argument to this function?
So:
```lua
if fallback_tag == nil then return false end
local clients = self:clients()
if #clients > 0 and ntags <= 1 and not force then return false end
```
(By the way, did you change the `or` in the old code into an `and` on purpose?)

lib/awful/tag.lua
```diff
@@ -309,7 +309,7 @@ function tag.object.delete(self, fallback_tag, force)
     end
 
     -- If no tags are visible, try and view one.
-    if target_scr.selected_tag == nil and ntags > 0 then
+    if target_scr.selected_tag == nil and ntags > 1 then
```
psychon: Could you also change the comment above? "If no tags are visible (and we did not delete the last tag), try and view one". This would actually describe this line of code a bit better. (At first I was confused here)

objects/client.c
```diff
@@ -2342,7 +2342,10 @@ luaA_client_tags(lua_State *L)
         lua_pushnil(L);
         while(lua_next(L, 2))
             tag_client(L, c);
+
         lua_pop(L, 1);
+
+        luaA_object_emit_signal(L, -1, "property::tags", 0);
```
psychon: Should this somehow be added to the API docs?

tests/_multi_screen.lua
```diff
@@ -0,0 +1,509 @@
...
+local wibox = require("wibox")
+local surface = require("gears.surface")
+local wallpaper = require("gears.wallpaper")
+local color = require("gears.color")
+local shape = require("gears.shape")
+
+local module = {}
+
+-- Get the original canvas size before it is "lost"
+local canvas_w, canvas_h = screen[1].geometry.width, screen[1].geometry.height
```
psychon: I think you are reinventing `local canvas_w, canvas_h = root.size()` here (which exists so far just for wallpaper setting in `gears.wallpaper`).

tests/_multi_screen.lua
```diff
@@ -0,0 +1,509 @@
...
+            return {
+                x = 0,
+                y = 0,
+                width  = math.floor(canvas_w/2),
+                height = canvas_h,
+            }
+        end,
+        function()
+            return {
+                x = math.floor(canvas_w/2)+1,
```
psychon: Perhaps just `x = math.ceil(canwas_w/2)`? Or why do you want to force a one pixel gap?

tests/_multi_screen.lua
```diff
@@ -0,0 +1,509 @@
...
+                y = 0,
+                width = math.floor(canvas_w/2),
```
psychon: But this one should be `math.ceil`, I guess :-)

tests/_multi_screen.lua
```diff
@@ -0,0 +1,509 @@
...
+local shape = require("gears.shape")
+
+local module = {}
+
+-- Get the original canvas size before it is "lost"
+local canvas_w, canvas_h = screen[1].geometry.width, screen[1].geometry.height
+
+-- Create virtual screens.
+-- The steps will be executed on each set of screens.
+local dispositions = {
```
psychon: Why is this an array of arrays of *functions` returning screen layouts?
Wouldn't it be easier to get rid of the functions? E.g.
```lua
local dispositions = {
  {
    { x = 0, y = 0, width = math.floor(canvas_w/2), height = canvas_h },
    { x = math.floor(canvas_w/2), y = 0, width = math.ceil(canvas_w/2), height = canvas_h },
  },
  {
etc
  }
}
```

tests/_multi_screen.lua
```diff
@@ -0,0 +1,509 @@
...
+        keep = true,
+    },
+
+    -- Take the previous disposition and resize the leftmost one
+    {
+        function()
+            return {
+                x = math.floor(canvas_w/4)+1,
+                y = 0,
+                width = math.floor(3*(canvas_w/4)),
```
psychon: How about having some "global variables" `local canvas_w_half = math.floor(canvas_w/2)` etc (`canvas_w_quarter`, `canvas_w_three_quarter`, ...). I'm not totally sure, but it might improve readability by getting of all the rounding that is needed here.

tests/_multi_screen.lua
```diff
@@ -0,0 +1,509 @@
...
+    for s in screen do
+        local sur = surface.widget_to_surface(
+            wibox.widget {
+                bg     = "#000000",
+                widget = wibox.container.background
+            },
+            s.geometry.width,
+            s.geometry.height
+        )
+        wallpaper.fit(sur, s, "#000000")
```
psychon: Uhm.... why the widget? How about just `wallpaper.set("#000000")` instead? This would even allow getting rid of the `for s in screen`, it would set the whole root window to black in a single operation. 

If you just want to draw black to actually-existing screens:
```lua
for s in screen do
  local _, cr = wallpaper.prepare_context(s)
  cr:set_source_rgb(0, 0, 0)
  cr:paint()
end
```

tests/_multi_screen.lua
```diff
@@ -0,0 +1,509 @@
...
+    end
+end
+
+-- Make it easier to debug the tests by showing the screen geometry when the
+-- tests are executed.
+local function show_screens()
+    wallpaper.set(color("#000000")) -- Should this clear the wallpaper? It doesn't
```
psychon: Sounds like a todo-comment. Yup, it should. do that. However, how about replacing this whole function:
```lua
local _, cr = wallpaper.prepare_context()
cr:set_source_rgb(0, 0, 0)
cr:paint()
cr.line_width = 1
for s in screen do
  cr:set_source_rgb(1, 1, 1)
  cr:fill_preserve()
  cr:set_source_rgb(1, 0, 0)
  cr:stroke()
end
```

tests/_multi_screen.lua
```diff
@@ -0,0 +1,509 @@
...
+                for i=#old_screens + 1, #geos do
+                    local geo = geos[i]
+                    screen.fake_add(geo.x, geo.y, geo.width, geo.height)
+                end
+
+                check_tag_indexes()
+            else
+                -- Move all existing screens out of the way (to avoid temporary overlapping)
+                for s in screen do
+                    s:fake_resize(canvas_w, canvas_h, 1, 1)
```
psychon: Doesn't this cause the screens to overlap if more than one is moved out of the way?!?

tests/test-awful-client.lua
```diff
@@ -79,4 +77,87 @@ end
...
+table.insert(steps, function()
+    -- Make sure there is no extra callbacks that causes double screen changes
+    -- (regress #1028 #1052)
+    client.connect_signal("property::screen",function(c)
+        if c.class and c.class ~= "screen"..c.screen.index then
+            has_error = table.concat {
+                "An unexpected screen change did occur\n",
+                debug.traceback(),"\n",
+                c.class, " Screen: ",c.screen.index
+            }
```
psychon: Why not just call `error("Foo")` directly instead of this `has_error` variable?

tests/test-awful-tag.lua
```diff
@@ -128,4 +128,117 @@ end
...
+                assert(t.screen == nil)
+            end
+        end
+
+        return true
+    end
+
+    for _, c in ipairs(client.get()) do
+        c:kill()
+    end
```
psychon: Isn't the killing already being done by `_multi_screen` and thus unnecessary here?

psychon:
> 2016-08-29 04:03:02 E: Error during a protected call: lib//wibox/drawable.lua:62: attempt to index field 'drawable' (a userdata value)
> 2016-08-29 04:03:02 E: Error during a protected call: lib//wibox/drawable.lua:62: attempt to index field 'drawable' (a userdata value)
> I can't reproduce this locally. It looks like some kind of messed up wibox without its __index metatable. Any idea how that's happening?

Look at 1408e8b9522186746137e90c672595c546323176. Yup, to avoid crashes in the C code, the metatable of all userdata objects is unset in the `__gc` metafunction. Thus, e.g. `__index` disappears.

The question rather is: What causes an pretty-much-unreachable drawable to be redrawn? By the time the delayed call for the redraw is pending, the drawable is strongly reachable and thus cannot be `__gc`'d. Thus, whatever triggered the redraw must have done so on an already-gc'd drawable. I'd suggest patching lib/wibox/drawable.lua, line 349 (function `draw()` on lua-drawable-wrappers) to see what causes this. Add something like the following:
```lua
if type(ret.drawable) == "userdata" then error("Type is not drawable: " .. ret.drawable .. "\n" .. debug.traceback()) end
```
Alternatively, just a simple `surface.load_silently(self.drawable.surface, false)` at this place should be enough to trigger this.

If the above doesn't help in figuring out what is wrong, then
- either my theory is wrong and this is not what is going on
- the drawable is `__gc`'d between the redrawn being made pending and actually happening. This can AFAIK happen, because Lua keeps a list of have-to-finalize objects after a GC round and only finalizes them step by step. Thus, Lua could have already decided that it will have to `__gc` the drawable, but didn't yet actually do so.

Neither of these two theories really help for debugging...

lib/awful/tag.lua
Elv13: While both methods causes race conditions, what you propose is worst as it will cause 2 tags to have to same index for a short time interval. Also, before you ask, setting the screen to `nil` to solve this also isn't a good idea have have its fair share of problems. There is already a possibility that this line does nothing at all because some other code will query the index and it will be re-generated elsewhere (see `get_index()`). However, this only happen if something explicitly query the index, so this line is required for the signal to be emitted.

lib/awful/tag.lua
Elv13: > The old logic was "don't allow deleting the last tag of a screen". I think that's quite a good idea since screens without tags will surely cause many problems.

The old logic was inconsistent with Awesome default state. No tags is currently supported and is heavily used by the Shifty extension. By default, Shifty has no tags and they are generated on the fly. This logic bug caused it to be unable to return to its original state. Tyrannical is also mildly affected by this (it isn't the default, but this behavior can be defined). But mostly, it bugs when you remove the screens as `:delete()` doesn't do its job anymore.

 > (By the way, did you change the or in the old code into an and on purpose?)

Yes,  returning early prevented important `:delete()` logic from being executed. When the tag _has_ to be deleted, then it _has_ to be deleted. Removing screens is a good example where the `:delete()` method has to work. Anyway, if `#clients > 0` is true, it still return, the fallback tag is irrelevant if there is nothing looking for a fallback, the previous code was wrong.

tests/_multi_screen.lua
Elv13: It is what I did at first. Then I imagined a future improvements where there is multiple scenarios for different unit tests. For example, some tests only care about the number of screen changing, they don't need the full disposition testing (it is quite long and slow).

To push things further, then functions are required, as the same could be used in multiple "stages"* and change its output based on the current state. It would save quite a lot of code. The functions are currently necessary, but this will change. If I ever finish Tyrannical tests**, I will need this.

\* A screen configuration change.
\** I just `ln -s` test-tyrannical-screen.lua and test-tyrannical-rules.lua into Awesome `awesome/tests/` folder and `ln -s` tyrannical itself into `awesome/lib` locally.

tests/_multi_screen.lua
Elv13: As they are going to be deleted anyway, it doesn't really matter much. On the plus side, we need to make sure overlapping screens don't cause any "hard" errors. I don't mean to **support** that use case, but as it happens, at least make sure it doesn't cause unrecoverable errors.

So far it looks to cause no issues, I did it on purpose.

tests/test-awful-client.lua
Elv13: Because it will happen in normal test execution. It is only an error if it happens during certain steps. I know in *this* PR, it isn't the case anymore. I have more tests, and then it became a problem and its why I did it. However, the final version of this PR doesn't include them (well, I moved one of them to the tag tests and the other isn't ready**). I could just disconnect the signal. It would work too.

\** test titlebar drag and drop (move or resize) across screens with tags selected, no tags selected and no tags at all.

psychon:
Another idea for that error: `lib/wibox/drawable.lua` connects to "wallpaper_changed" and force-redraws all drawables there (needed when fake transparency is used, but for some reason we do it always...). Perhaps this is what triggers the repaint for an already GC'd drawable?

Elv13:
> Perhaps this is what triggers the repaint for an already GC'd drawable?

I saw the error locally when setting a wallpaper in Xephyr using feh, so this looks like the culprit. I still can't reproduce using the tests.

tests/_multi_screen.lua
Elv13: Is it really worth re-inventing the wheel when we have a very capable widget system that can do just that without all the manual placement stuff?

Elv13:
Updated (unaddressed comments need more discussions)

I tried to fix the drawable bug. However, it is quite nasty. Now I get

    awesome: /build/buildd/cairo-1.13.0~20140204/src/cairo-pattern.c:1062: cairo_pattern_reference: Assertion `((*&(&pattern->ref_count)->ref_count) > 0)' failed.

So it isn't much better. Also, I am sure @psychon will hate my hack. I tried many different things before that. Commenting the "forced redraw on wallpaper change" does **not** make the tests pass on lua5.2 and 5.3. They pass on 5.1 and luajit.

tests/_multi_screen.lua
psychon: Nope, it's not worth reinventing the wheel. I forgot/missed the needed text part. Without the text, I'd say yes. ;-)

psychon:
> Commenting the "forced redraw on wallpaper change" does not make the tests pass on lua5.2 and 5.3.

Doesn't that basically mean that something else causes this issue? Hm...
I think we had the problem "already gc'd drawable being repainted" a while ago already and back then it was a bug in `gears.object.weak_connect_signal`. But that should be fixed... :-(

objects/drawable.c
```diff
@@ -225,6 +233,7 @@ drawable_class_setup(lua_State *L)
     static const struct luaL_Reg drawable_meta[] =
     {
         LUA_OBJECT_META(drawable)
+        { "__gc", luaA_collect },
```
psychon: I don't think this works. `common/luaclass.c`, `luaA_class_setup()` sets up its own `__gc` handler, so only one of the two will be executed (I'm not sure which one wins; if your "collect" signal is being emitted, then apparently your new code won). Instead, you want to add this `emit_signal` to `drawable_wipe()` which is called by the other `__gc`-handler.

lib/wibox/drawable.lua
```diff
@@ -408,6 +410,11 @@ function drawable.new(d, widget_context_skeleton, drawable_name)
         ret:draw()
     end
 
+    d:connect_signal("collect", function()
+        print("\n\nCOLLECT!")
+        ret._redraw_pending = false
```
psychon: I'd suggest `ret.was_GCd = true` and in the `:draw()` and `:_do_redraw()`-functions then `if ret.was_GCd then print("redraw after GC caused by", debug.traceback()) end`.

Elv13:
```lua
Redrawing on an invalid drawable	stack traceback:
	lib//wibox/drawable.lua:353: in function 'draw'
	lib//wibox/drawable.lua:413: in function 'layout_callback'
	lib//wibox/hierarchy.lua:52: in function 'func'
	lib//gears/object.lua:120: in function 'emit_signal'
	lib//wibox/widget/imagebox.lua:123: in function 'set_image'
	lib//awful/titlebar.lua:246: in function <lib//awful/titlebar.lua:218>
```

Here we go

Elv13:
So, there is 2 of them. The first is because "unfocus" is emitted in the same event look iteration as "umanage". This causes the titlebar imageboxes to be redrawn after the client is unmanaged.

The second one, as you expected, is the wallpaper one. So more of the same. If `c:kill()` is called in the same iteration as the wallpaper, then the error happen. No amount of delayed call can fix this. I added a few commits to make delayed redraw cancelable. It imply the "collect" signal is no longer for debug only, but permanent. I wasn't very lucky sending it from wipe (and yes, I added it to the stack and poped it after), so I call the function manually instead. I didn't checked if it has deeper implications. It seems to work.

The last main issue is the fact that starting so many new clients over all those screens configuration makes `test-awful-client.lua` too slow for the CI (up to 110 seconds, but sometime below 30). Is it possible to unmanage, then re-manage all those clients instead of spawning new ones?

psychon:
> So, there is 2 of them. The first is because "unfocus" is emitted in the same event look iteration as "umanage". This causes the titlebar imageboxes to be redrawn after the client is unmanaged.
> The second one, as you expected, is the wallpaper one.

I'll have to think about this. For now, I don't understand this.
Unmanaging a client queues a pending redraw. However, it does *not* invalidate the drawable, so the following redraw (even though it is pointless) should still work correctly. And the drawable cannot be garbage collected, because the pending redraw keeps a strong reference to it...

I'll see if I can find a simple(r) test case for this on the weekend (likely Sunday) (if no one is faster on handling this; does this issue deserve its own PR even though there are no tests yet for it?).

For this PR I'd suggest using the `if type(self.drawable) == "userdata" then return end -- TODO: Fix this in a non-ugly way` hack. That way this doesn't block this PR. Feel free to open another issue about "fix this properly" and mark it as a 3.6 blocker, if you want.

> The last main issue is the fact that starting so many new clients over all those screens configuration makes test-awful-client.lua too slow for the CI (up to 110 seconds, but sometime below 30). Is it possible to unmanage, then re-manage all those clients instead of spawning new ones?

If you want me to guess, then half of the slowness is due to starting so many Lua processes which all have to load lgi etc again. You *could* try coming up with a way to avoid this. For example, `_test_client.lua` could spawn a single process when it is loaded and then communicate with it via stdin. Whenever a new test client is needed, the information about it is sent over stdin and the process then opens a new window.
I do not know if this would really help a lot, but I don't think we can come up with anything better than the above. If you want to implement this, I'd suggest to do it in another pull request.

Elv13:
ok then, so you want to fast-track this with both `--FIXME` suggested above? Is there any other comments above you want more discussions? There is quite a few comments above I "ignored", I just want to make sure. Of course, I will clean it up and remove the temporary commits. Of course, having those tests and fixes is already a good step forward, it close some of the remaining 3.6 blocker (it's why I did it). Also, I think I have a clue about the coverage issues. The cmake (and probably unit) tests are executed 3 times, it should be 2 and 1 respectively. This is probably messing up something in the coverage data (this is an untested theory).

psychon:
> ok then, so you want to fast-track this with both --FIXME suggested above?

Dunno if I would call it "fast-track", let's just call it "normal". :-) And yeah, the FIXMEs should become release blockers, but since they are about unrelated problems that already existed before, sure.

Just to be sure: One of the FIXMEs is the slow spawning of test clients and the other FIXME is the temporary `if type(foo) == "userdata" then`-hack, right?

> Is there any other comments above you want more discussions? There is quite a few comments above I "ignored", I just want to make sure. 

For the ones that you replied to, your answer sounded convincing enough. I'm too lazy to go through them again right now, so if that's all, ok.

PR #1058, opened 2016-08-29
awesome-client: send command line arguments to awesome

bucciarati:
In some languages, invoking commands with command-line arguments is
significantly simpler than setting up pipes for writing to a command's
standard input.

This adds an additional way for sending commands through awesome-client;
so all of these will be equivalent:

  $ awesome-client 'awesome.restart()'
  $ echo 'awesome.restart()' | awesome-client
  $ awesome-client <<<'awesome.restart()'
  $ awesome-client
  awesome# awesome.restart()
  awesome# ^D
  $

Note that this sends each command line argument as a separate message
over dbus.

### DIFF
--- a/utils/awesome-client
+++ b/utils/awesome-client
@@ -17,7 +17,7 @@ then
     then
         if [ "$A_RERUN" = "" ]
         then
-            A_RERUN="no" exec $RLWRAP $0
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!

PR #1061, opened 2016-08-30
tooltip: Stop using a fallback font

Elv13:
First, that font isn't very common and it also causes a tiny
tooltip on normal screens. 6px on 1080p laptop screen (not
using HiDPI) is totally unreadable.

Let the default font be used, just as everywhere else.


psychon:
It's not 6 pixels, it's 6...something font size related (6/72 inch? No...). Anyway, this should be the physical sizes and if a 1080p displays makes it display too small, then your dpi is off.

Still: :+1:

PR #1062, opened 2016-08-30
Make use of awesome-client commands

psychon:
This uses the new support introduced in f0f31bc30588 / https://github.com/awesomeWM/awesome/pull/1058 in the docs and in tests/run.sh, removing an useless use of cat/echo.

### DIFF
--- a/utils/awesome-client
+++ b/utils/awesome-client
@@ -44,14 +44,16 @@ a_dbus_send()
     $DBUS_SEND --dest=$DBUS_DEST --type=method_call --print-reply \
         $DBUS_PATH $DBUS_METHOD string:"$1" | tail -n +2
     ret=$?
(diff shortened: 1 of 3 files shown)


psychon:
...and now this also fixes the return value for things like `awesome-client "error('foo')"`!

utils/awesome-client
```diff
@@ -44,14 +44,16 @@ a_dbus_send()
     $DBUS_SEND --dest=$DBUS_DEST --type=method_call --print-reply \
```
blueyed: commit message: with "when sending a command that it for from stdin failed": s/for/got/ ?

psychon: Yup, done.

Elv13:
👍

actionless:
:+1:

PR #1066, opened 2016-09-01
awful.tag: Add `gap_single_client` property and user changeable defau…

Akulen:
This is a reimplementation of [https://github.com/awesomeWM/awesome/pull/860], with `gap_count` being renamed to `gap_single_client` and changed to a boolean (true meaning that useless gaps are allowed even when only one client is present)

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -137,7 +137,13 @@ function layout.parameters(t, screen)
 
     local p = {}
 
-    local useless_gap = t and t.gap or 0
+    local clients           = client.tiled(screen)
+    local gap_single_client = true
+    if(t and not (t.gap_single_client == nil)) then
+        gap_single_client = t.gap_single_client
+    end
+    local min_clients       = gap_single_client and 1 or 2
+    local useless_gap       = t and (#clients >= min_clients and t.gap or 0) or 0
 
     p.workarea = screen:get_bounding_geometry {
         honor_padding  = true,
@@ -146,7 +152,7 @@ function layout.parameters(t, screen)
     }
 
     p.geometry    = screen.geometry
-    p.clients     = client.tiled(screen)
+    p.clients     = clients
     p.screen      = screen.index
     p.padding     = screen.padding
     p.useless_gap = useless_gap
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -139,9 +139,11 @@ function layout.parameters(t, screen)
 
     local clients           = client.tiled(screen)
     local gap_single_client = true
-    if(t and not (t.gap_single_client == nil)) then
+    
+    if(t and t.gap_single_client ~= nil) then
         gap_single_client = t.gap_single_client
     end
+    
     local min_clients       = gap_single_client and 1 or 2
     local useless_gap       = t and (#clients >= min_clients and t.gap or 0) or 0
 
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -41,6 +41,41 @@ data.tags = setmetatable({}, { __mode = 'k' })
 tag.history = {}
 tag.history.limit = 20
 
+-- Default values
+tag.defaults = {}
+--- The gap between clients (in points).
+-- @tfield[opt=0] integer tag.defaults.gap
+-- @tparam[opt=0] integer gap
+-- @see gap
+-- @see gap_single_client
+tag.defaults.gap                 = 0
+--- The default gap_count.
+-- @tfield[opt=true] boolean tag.defaults.gap_single_client
+-- @tparam[opt=true] boolean gap_single_client
+-- @see gap
+-- @see gap_single_client
+tag.defaults.gap_single_client   = true
+--- The default master fill policy.
+-- @tfield[opt="expand"] string tag.defaults.master_fill_policy
+-- @tparam[opt="expand"] string master_fill_policy
+-- @see master_fill_policy
+tag.defaults.master_fill_policy  = "expand"
+--- The default master width factor.
+-- @tfield[opt=0.5] number tag.defaults.master_width_factor
+-- @tparam[opt=0.5] number master_width_factor
+-- @see master_width_factor
+tag.defaults.master_width_factor = 0.5
+--- The default master count.
+-- @tfield[opt=1] integer tag.defaults.master_count
+-- @tparam[opt=1] integer master_count
+-- @see master_count
+tag.defaults.master_count        = 1
+--- The default column count.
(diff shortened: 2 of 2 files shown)


lib/awful/layout/init.lua
```diff
@@ -137,7 +137,10 @@ function layout.parameters(t, screen)
 
     local p = {}
 
-    local useless_gap = t and t.gap or 0
+    local clients           = client.tiled(screen)
+    local gap_single_client = t and t.gap_single_client or true
```
actionless:
```
$ lua
Lua 5.3.3  Copyright (C) 1994-2016 Lua.org, PUC-Rio
> true and false or true
true
```

actionless:
also i think it could useful to add for a single client some logic similar to the one presented in this code sample:
https://github.com/awesomeWM/awesome/issues/826

Elv13:
I changed my mind a little since I made the original versions. The `.defaults` variables should become hardcoded again, but each should have a `@beautiful` property. This way themes can set their own `master_width_factor`

Akulen:
I've fixed the value attribution for `gap_single_client` in init.lua and each of the defaults variable are overriden by their beautiful counterparts. I dont know it the defaults value should be hardcoded (maybe just not documented)

lib/awful/tag.lua
```diff
@@ -41,6 +41,41 @@ data.tags = setmetatable({}, { __mode = 'k' })
 tag.history = {}
 tag.history.limit = 20
 
+-- Default values
+tag.defaults = {}
+--- The gap between clients (in points).
+-- @tfield[opt=0] integer tag.defaults.gap
+-- @tparam[opt=0] integer gap
+-- @see gap
+-- @see gap_single_client
```
Elv13: This will need some empty lines between the block

Elv13:
> I dont know it the defaults value should be hardcoded (maybe just not documented)

I was thinking about turning the vars into `beautiful` theme variables. Then hardcoding the definitive fallback.

In some other unrelated issue, I discussed the possibility of adding hardcoded fallbacks to all theme variables so incomplete themes become supported again. I see this feature as something a theme might want to set.

Akulen:
> I was thinking about turning the vars into beautiful theme variables. Then hardcoding the definitive fallback.

I've already added a beautiful counterpart to all the defaults variables, but i was thinking that they could still be centralized in a defaults array (maybe just in `local defaults = {}` ...) to make it easier to see and edit the default values if ever needed.

actionless:
>> I've already added a beautiful counterpart to all the defaults variables, but i was thinking that they could still be centralized in a defaults array (maybe just in local defaults = {} ...) to make it easier to see and edit the default values if ever needed.

good idea but i would suggest using `awful.layout.defaults = {...}` or smth like that instead of `local defaults = {...}` since such approach is already used in some awesome modules (but they're documented, so i would recommend to document those as well)

Akulen:
>  i would suggest using `awful.layout.defaults = {...}`

Yeah, it seems better than to use a local array, but these are properties for `awful.tag`, so i don't think `awful.layout.defaults` is the correct location. What I think should be used is `awful.tag.defaults`, but that's already the case in my PR and i've documented them as well

lib/awful/layout/init.lua
```diff
@@ -137,7 +137,13 @@ function layout.parameters(t, screen)
 
     local p = {}
 
-    local useless_gap = t and t.gap or 0
+    local clients           = client.tiled(screen)
+    local gap_single_client = true
+    if(t and not (t.gap_single_client == nil)) then
```
Elv13: Please use `~=` and add a new line before and after for readability

lib/awful/tag.lua
```diff
@@ -41,6 +41,21 @@ data.tags = setmetatable({}, { __mode = 'k' })
 tag.history = {}
 tag.history.limit = 20
 
+-- Default values
+local defaults = {}
+-- The gap between clients (in points).
```
Elv13: Please add new lines between the variables for consistency with other files

lib/awful/tag.lua
```diff
@@ -570,7 +593,7 @@ function tag.object.set_master_width_factor(t, mwfact)
 end
 
 function tag.object.get_master_width_factor(t)
-    return tag.getproperty(t, "master_width_factor") or 0.5
+    return tag.getproperty(t, "master_width_factor") or beautiful.master_width_factor or defaults.master_width_factor
```
Elv13: we have a soft 80 line policy (most of the time)

```lua
return tag.getproperty(t, "master_width_factor")
    or beautiful.master_width_factor
    or defaults.master_width_factor
```

lib/awful/tag.lua
```diff
@@ -835,7 +892,7 @@ end
 -- @param string "expand" or "master_width_factor"
 
 function tag.object.get_master_fill_policy(t)
-    return tag.getproperty(t, "master_fill_policy") or "expand"
+    return tag.getproperty(t, "master_fill_policy") or beautiful.master_fill_policy or defaults.master_fill_policy
```
Elv13: see the previous comment (line length)

lib/awful/tag.lua
```diff
@@ -1005,7 +1074,7 @@ function tag.object.set_column_count(t, ncol)
 end
 
 function tag.object.get_column_count(t)
-    return tag.getproperty(t, "column_count") or 1
+    return tag.getproperty(t, "column_count") or beautiful.column_count or defaults.column_count
```
Elv13: same

lib/awful/tag.lua
```diff
@@ -878,9 +935,15 @@ function tag.getmfpol(t)
     util.deprecate("Use t.master_fill_policy instead of awful.tag.getmfpol")
 
     t = t or ascreen.focused().selected_tag
-    return tag.getproperty(t, "master_fill_policy") or "expand"
+    return tag.getproperty(t, "master_fill_policy") or beautiful.master_fill_policy or defaults.master_fill_policy
```
Elv13: this one too

Elv13:
👍

lib/awful/tag.lua
```diff
@@ -41,6 +41,27 @@ data.tags = setmetatable({}, { __mode = 'k' })
 tag.history = {}
 tag.history.limit = 20
 
+-- Default values
+local defaults = {}
```
actionless: but `defaults` is still local

Elv13: Yes, I request that change on IRC. Sorry about that, but I don't think having 2 places to change the default is a good idea. You can set beautiful variables in rc.lua if you want. I didn't wanted to have to deal with the confusion having an exported default table would ensue.

PR #1067, opened 2016-09-02
Fix a bug moving between a tiled and a floating screen.

BlueUsername:
When dragging a client with the mouse from a tiled screen on the left to a floating screen on the right the client will flicker when the mouse is on the right screen but the left side of the client is on the left.

This is because the screen is changed when the mouse moves screen, but is then reset if the left most edge isn't on this screen. This fix checks if any part of the client is on the new screen instead.

--
Check if any of the window is on the screen it is moved to, not just
the topmost leftmost edge.

Signed-off-by: blueusername <blueusername@github>

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1506,7 +1506,7 @@ client_resize_do(client_t *c, area_t geometry, bool force_notice)
     bool java_is_broken = true;
 
     screen_t *new_screen = c->screen;
-    if(!screen_coord_in_screen(new_screen, geometry.x, geometry.y))
+    if(!screen_area_in_screen(new_screen, geometry))
         new_screen = screen_getbycoord(geometry.x, geometry.y);
 
     if(c->geometry.width == geometry.width
--- a/objects/screen.c
+++ b/objects/screen.c
@@ -731,6 +731,20 @@ screen_coord_in_screen(screen_t *s, int x, int y)
            && (y >= s->geometry.y && y < s->geometry.y + s->geometry.height);
 }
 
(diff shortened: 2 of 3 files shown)


blueyed:
Are you sure this does not revert 50982fc?

BlueUsername:
It doesn't revert it but it changes it slightly as now it stays on the current screen until ALL of the client is off the first screen, rather than just one edge. Going off the commit message for that commit this sounds like what is meant to happen but I don't know.

blueyed:
Yeah, sounds like it's an improvement to my previous commit.

Thanks!

:+1:

actionless:
i am not too familiar with C part of awesome but looks good to me

psychon:
> This is because the screen is changed when the mouse moves screen, but is then reset if the left most edge isn't on this screen.

Uhm, I'm confused. So, awesome decides the client's screen based on the top-left corner. Why does any other part of the client matter?

Which part of the code does `c.screen = mouse.screen` when the mouse changes screen? Isn't that code the problem (Or at least part of it)?

psychon:
I think that my question is still open.

Elv13:
Oh, sorry

BlueUsername:
> Which part of the code does c.screen = mouse.screen when the mouse changes screen?

https://github.com/awesomeWM/awesome/blob/master/lib/awful/layout/init.lua#L264

> Isn't that code the problem (Or at least part of it)? 

I guess that depends on what you believe should be the intended functionality, changing screen when the  mouse moves or changing screen when the top-left corner would move.

PR #1070, opened 2016-09-06
Titlebars "draw after free" errors

Elv13:
Here are the few commits from the multi screen PR. It isn't the best "solution" and will require more investigation.

Those commits force the code to behave even when such issue happen. It doesn't fix the root cause. However, I think we also have to ask ourselves if this can happen due to an user "mistake". In that case, maybe those commits are worth merging anyway (even if the 2 errors triggering this are fixed).

### DIFF
--- a/lib/awful/titlebar.lua
+++ b/lib/awful/titlebar.lua
@@ -216,6 +216,10 @@ function titlebar.widget.button(c, name, selector, action)
     end
 
     local function update()
+        -- This is necessary as this callback can be triggerred from a delayed
+        -- call anywhere in the code.
+        if c.valid then return end
+
         local img = selector(c)
         if type(img) ~= "nil" then
             -- Convert booleans automatically
@@ -256,6 +260,9 @@ function titlebar.widget.button(c, name, selector, action)
     -- connect to the corresponding signal here.
     c:connect_signal("focus", update)
     c:connect_signal("unfocus", update)
+    c:connect_signal("unmanage", function()
+        c:disconnect_signal("unfocus", update)
+    end)
 
     return ret
 end
--- a/lib/gears/timer.lua
+++ b/lib/gears/timer.lua
@@ -150,7 +150,9 @@ end
 local delayed_calls = {}
 capi.awesome.connect_signal("refresh", function()
     for _, callback in ipairs(delayed_calls) do
-        protected_call(unpack(callback))
+        if #callback > 0 then
+            protected_call(unpack(callback))
+        end
     end
     delayed_calls = {}
 end)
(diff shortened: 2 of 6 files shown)


objects/client.c
```diff
@@ -1988,8 +1988,16 @@ client_unmanage(client_t *c, bool window_valid)
...
     if(globalconf.focus.client == c)
+    {
         client_unfocus(c);
 
+        /* Avoid unmanage and unfocus to be in the same lua iteration. This
+         causes a race condition where delayed redraw are registered for an
+         already unmanaged client
+         */
+        awesome_refresh();
+    }
```
psychon: Doesn't this just make it a lot harder to hit the problem, but it can still happen if awesome just happens to be slow, or something like that? I'm not a fan (but I guess you already knew).

lib/gears/timer.lua
```diff
@@ -150,17 +150,24 @@ end
...
+        end
     end
     delayed_calls = {}
 end)
 
 --- Call the given function at the end of the current main loop iteration
 -- @tparam function callback The function that should be called
 -- @param ... Arguments to the callback function
+-- @treturn table All callbacks. This table can be modified during the current
+--  event loop. If all callbacks are removed, then it is cancelled.
```
psychon: Uhm... this returns a table with entry 1 being the functions to call and the other (array) entries are its arguments. That's (a) a bad thing to return (as an API) and (b) not really what the docs describe.

If you want something unique to return, I'd do something like this:
```lua
function timer.delayed_call(callback, ...)
  local args = { callback, ..., _internal_unique_id = { _internal_cancelled = false } }
  table.insert(delayed_calls, args)
  return args._internal_unique_id
end

function timer.cancel_delayed_call(identifier)
  identifier._internal_cancelled = true
end

capi.awesome.connect_signal("refresh", function()
  for _, callback in ipairs(delayed_calls) do
    if not callback._internal_unique_id._internal_cancelled then
      protected_call(unpack(callback))
    end
  end
  delayed_calls = {}
end)
```

lib/awful/titlebar.lua
```diff
@@ -256,6 +260,9 @@ function titlebar.widget.button(c, name, selector, action)
     -- connect to the corresponding signal here.
     c:connect_signal("focus", update)
     c:connect_signal("unfocus", update)
+    c:connect_signal("unmanage", function()
+        c:disconnect_signal("unfocus", update)
+    end)
```
psychon: This is dead code. The C code unfocuses the client before it emits the unmanage signal, so this can never occur.

lib/awful/titlebar.lua
```diff
@@ -216,6 +216,10 @@ function titlebar.widget.button(c, name, selector, action)
     end
 
     local function update()
+        -- This is necessary as this callback can be triggerred from a delayed
+        -- call anywhere in the code.
+        if c.valid then return end
```
psychon: Shouldn't this be `not c.valid` instead of `c.valid`?

objects/drawable.c
```diff
@@ -225,6 +237,7 @@ drawable_class_setup(lua_State *L)
     static const struct luaL_Reg drawable_meta[] =
     {
         LUA_OBJECT_META(drawable)
+        { "__gc", luaA_collect },
```
psychon: This causes leaks, because the "real" __gc handler isn't run. Also, I don't like the idea, but I guess you already know...

psychon:
Previous discussions of / some more information about the issue that this addresses: https://github.com/awesomeWM/awesome/pull/1056#issuecomment-244802384 (just so that I don't have to search for it again)

blueyed:
@Elv13 
What's the reason for closing it?

Elv13:
Yes, the fix was merged in  #1139. Uli just forgot to add the "Fix #1070" in the commit.

blueyed:
Cool, thanks!

Issue #1071, opened 2016-09-07
Only a single listener allowed per name using dbus.connect_signal


hoelzro:
Output of `awesome --version`:

```
awesome v3.5.9 (Mighty Ravendark)
 • Build: Sep  7 2016 08:27:17 for x86_64 by gcc version 6.1.1 (rob@eridanus)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔
```

**How to reproduce the issue:**

Call this twice: `dbus.connect_signal('org.freedesktop.DBus.Properties', function() end)`

**Actual result:**

`dbus.connect_signal` returns `nil` both times, a warning is emitted to Awesome's output, and the second callback function is not invoked for `org.freedesktop.DBus.Properties` DBus events.

**Expected result:**

Ideally, I would like for multiple functions to be able to listen for the same DBus event.  

Barring that, it would be nice if `dbus.connect_signal` returned `true` upon success and `nil, string` upon failure, like many other Lua functions, so that scripters could know from their script whether or not they were able to register the callback.  In addition, it would be nice if there were a way to get the callback currently associated with the name so that the user could manually chain callbacks, like so:

```lua
local interface_name = 'org.freedesktop.DBus.Properties'
local function callback(...)
end
if not dbus.connect_signal(interface_name, callback) then
  local current_callback = dbus.get_connection(interface_name)
  dbus.disconnect_signal(interface_name, current_callback)
  dbus.connect_signal(interface_name, function(...)
    pcall(current_callback, ...)
    pcall(callback, ...)
  end)
end
```

psychon:
Awesome's dbus bindings are quite... basic. Also, writing awesome-specific dbus Lua bindings limits their use to awesome. The long-term goal is that someone comes up with generic dbus Lua bindings (based on lgi) that also work in awesome. When this happens, all of the dbus-based code in awesome('s C code) would be dropped.

Thus, I'm not really enthusiastic for improving the existing bindings.

Anyway:
- Multiple functions connected to the same signal does not work. Even though these are called "signal", this is not a signal in the dbus sense. This is also used for method calls and here the return value of the function is the return value for the dbus call (see e.g. `lib/awful/remote.lua`). For this reason, we cannot just allow multiple functions to be connected.
- Chaining callbacks: I don't really see the point...? What would you do with the return value of the existing callback? Just ignore it? That's not good.
[...]

PR #1078, opened 2016-09-09
Return success/failure for dbus.connect_signal

hoelzro:
dbus.connect_signal may only have a single function bound to
a given name, but the caller has no way of knowing if their
function was bound or not.

This change has dbus.connect_signal adopt the standard Lua error
convention of returning a truthy value upon success, or nil and an error
message upon failure.

Ties into #1071

### DIFF
--- a/dbus.c
+++ b/dbus.c
@@ -761,6 +761,8 @@ luaA_dbus_remove_match(lua_State *L)
  *
  * @param interface A string with the interface name.
  * @param func The function to call.
+ * @return true on success, nil + error if the signal could not be connected
+ * because another function is already connected.
  * @function connect_signal
  */
 static int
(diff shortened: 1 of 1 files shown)


dbus.c
```diff
@@ -770,11 +772,16 @@ luaA_dbus_connect_signal(lua_State *L)
...
+    if(sig) {
         luaA_warn(L, "cannot add signal %s on D-Bus, already existing", name);
-    else
+        lua_pushnil(L);
+        lua_pushfstring(L, "cannot add signal %s on D-Bus, already existing", name);
+        return 2;
+    } else {
         signal_connect(&dbus_signals, name, luaA_object_ref(L, 2));
-    return 0;
+        lua_pushboolean(L, 1);
```
psychon: Minor nitpick: We include stdbool.h, so you could just say `true` instead of `1` here, but that's not highly important.

PR #1085, opened 2016-09-11
ewmh: Remove dead code

Elv13:
The geometry storage has been moved into awful.placement. This
code was never executed as data[] was never populated.

There is some behavior that is indeed lost, but it is unlikely
someone will ever notice (it has been broken for 6 months)

(whoops, I cause that when streamlining the ewmh code to play nice with the request:: API)


psychon:
Could you specify what this "some behaviour" that was lost is? (Perhaps even in the commit message)

PR #1087, opened 2016-09-11
Add a list of build dependencies to the readme

### LINKED ISSUE #426: build guide should be improved
Hi,
I tried to build awesome today which worked finally. Anyway, some things were a bit annoying:

1. No list of dependencies is given. I had to test-install some packages to find out whether they provided what was missing. After doing //mk-build-deps awesome -i// only the packages //libxkbcommon(-x11)?-dev// were missing on Debian/testing.
2. The documentation says that after calling make one has to change the directory and call cmake. That's not correct since make did this for me. Thats great! But not according to the documentation :-)

So maybe one could improve a bit on that documentation. I agree, that this points are minor.

psychon:
The list was more or less generated in an ad-hoc fashion by reading through the CMake files, `.travis.yml`, `build-utils/lgi-check.sh` and some other things. The links were (mostly) generated by picking the first Google result. No idea where e.g. the X11 should go to.

This is intended to fix #426, but I guess people actually want the names of distro packages that they need. I'm ignoring this here. (Possible issues: What if some dependency is not available in a distro? How "large" should a distro be so that it can be added here? Who keeps this thing up to date? ....)

### DIFF
--- a/docs/01-readme.md
+++ b/docs/01-readme.md
@@ -24,6 +24,53 @@ or by auto-generating a .deb or .rpm package, for easy removal later on:
     # or
     sudo rpm -Uvh awesome-x.y.z.rpm
 
+### Build dependencies
+
+Awesome has the following dependencies (besides a more-or-less standard POSIX
+environment):
+
+- [CMake >= 3.0.0](https://cmake.org)
+- [Lua >= 5.1.0](https://www.lua.org) or [LuaJIT](http://luajit.org)
+- [LGI >= 0.7.1](https://github.com/pavouk/lgi)
+- [xproto >= 7.0.15](https://www.x.org/archive//individual/proto/)
+- [libxcb >= 1.6](https://xcb.freedesktop.org/) with support for the RandR, XTest, Xinerama, SHAPE and
+  XKB extensions
+- [libxcb-cursor](https://xcb.freedesktop.org/)
+- [libxcb-util >= 0.3.8](https://xcb.freedesktop.org/)
+- [libxcb-keysyms >= 0.3.4](https://xcb.freedesktop.org/)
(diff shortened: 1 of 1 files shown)


psychon:
I proof read this yesterday, but apparently didn't actually look at this. At least I don't know how those kinds of typos and copy&paste errors could happen. Sorry.
The two commits should be squashed into one before merging, but I cannot do so currently.

docs/01-readme.md
```diff
@@ -24,6 +24,53 @@ or by auto-generating a .deb or .rpm package, for easy removal later on:
...
+- [GLib](https://wiki.gnome.org/Projects/GLib) with support for GObject
+  introspection
+- [GIO](https://developer.gnome.org/gio/stable/) with support for GObject
+  introspection
+- [GdkPixbuf](https://wiki.gnome.org/Projects/GdkPixbuf)
+- libX11 with xcb support
+- [Imagemagick's convert utility](http://www.imagemagick.org/script/index.php)
+- [libxdg-basedir >= 1.0.0](https://github.com/devnev/libxdg-basedir)
+
+Additionally, the following optional dependencies exists:
```
blueyed: s/exists/exist/

blueyed:
:+1: 

> want the names of distro packages that they need

What about mentioning e.g. `apt-get build-dep` for Debian/Ubuntu?  (although that would be incomplete for master, but is a good start)

docs/01-readme.md
psychon: Done

psychon:
`make package` only helps you after you have all the build dependencies. And yeah, `build-dep` is only half the answer. I don't really know what to do for those.
Besides that: Updated

PR #1088, opened 2016-09-11
Allow relative path in theme

zhustec:
For example:

Before, we can only use absolute path or `~/` in theme.lua file

``` lua
theme.wallpaper = "/usr/local/share/awesome/themes/default/background.png"
```

Now, we can use relative path

``` lua
theme.wallpaper = "./background.png"
```

### DIFF
--- a/lib/beautiful/init.lua
+++ b/lib/beautiful/init.lua
@@ -133,11 +133,11 @@ function beautiful.init(config)
                             theme[k] = v:gsub("^%./", configdir .. "/")
                         end
                     end
-                end
-            -- expand '~'
-            elseif type(config) == 'table' then
-                for k, v in pairs(theme) do
-                    if type(v) == "string" then theme[k] = v:gsub("^~/", homedir .. "/") end
+                -- expand '~'
+                elseif type(config) == 'table' then
+                    for k, v in pairs(theme) do
+                        if type(v) == "string" then theme[k] = v:gsub("^~/", homedir .. "/") end
+                    end
                 end
             end
 
--- a/lib/beautiful/init.lua
+++ b/lib/beautiful/init.lua
@@ -111,26 +111,27 @@ end
 function beautiful.init(config)
     if config then
         local homedir = os.getenv("HOME")
+        local theme_path = ""
 
         -- If `config` is the path to a theme file, run this file,
         -- otherwise if it is a theme table, save it.
         if type(config) == 'string' then
             -- Expand the '~' $HOME shortcut
             config = config:gsub("^~/", homedir .. "/")
-            configdir = config:gsub("/[^/]+$", "")
             theme = protected_call(dofile, config)
+            theme_path = config:gsub("/[^/]+$", "")
         elseif type(config) == 'table' then
             theme = config
         end
 
         if theme then
-            -- expand '~' and './'
             if homedir then
+                -- expand '~' and './'
                 if type(config) == "string" then
                     for k, v in pairs(theme) do
                         if type(v) == "string" then
                             theme[k] = v:gsub("^~/", homedir .. "/")
-                            theme[k] = v:gsub("^%./", configdir .. "/")
+                            theme[k] = v:gsub("^%./", theme_path .. "/")
                         end
                     end
                 -- expand '~'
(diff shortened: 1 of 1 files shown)


lib/beautiful/init.lua
```diff
@@ -117,14 +117,25 @@ function beautiful.init(config)
         if type(config) == 'string' then
             -- Expand the '~' $HOME shortcut
             config = config:gsub("^~/", homedir .. "/")
+            configdir = config:gsub("/[^/]+$", "")
```
psychon: Start awesome with arguments `-c relative/path/to/some/rc.lua`. This code will break.

I like this idea, but right now I don't know how to make it work reliably. How about instead giving the relative path to the config to the theme, so it can use it where needed?

farkasity:
Please try this time, it works for me. If it still break, please you show more information, I will trying to fix it.

lib/beautiful/init.lua
Elv13: Please avoid global variables. This is bad and this causes the Travis CI to fail.

farkasity: @Elv13 My bad, fix it now.

@psychon Maybe it's them problem of your custom `rc.lua` file, since it works well for me.

My desktop entry file.

```
% cat /usr/share/xsessions/awesome-test.desktop 
[Desktop Entry]
Encoding=UTF-8
Name=awesome-test
Comment=Highly configurable framework window manager
Exec=/usr/local/bin/awesome -c /usr/local/etc/xdg/awesome/rc.lua
Type=Application
```

psychon:
Makes sense to me to have paths relative to the theme in the theme: :+1: 
(And this still won't work with relative theme paths, but I can live with that since it already was like this before)

lib/beautiful/init.lua
psychon: Just for completeness, something like `sh -c 'cd /usr/local/etc && /usr/local/bin/awesome -c awesome/rc.lua` would have the issue I meant, because here the path to the config is a relative path. However, as I wrote below, the code already assumes absolute paths elsewhere and thus we can ignore this here.

farkasity:
@psychon Maybe we can do more? Just use

```lua
theme.wallpaper = "background.png"
```

Elv13:
👍

lib/beautiful/init.lua
```diff
@@ -111,22 +111,34 @@ end
...
-            -- expand '~'
             if homedir then
-                for k, v in pairs(theme) do
-                    if type(v) == "string" then theme[k] = v:gsub("^~/", homedir .. "/") end
+                -- expand '~' and './'
+                if type(config) == "string" then
+                    for k, v in pairs(theme) do
+                        if type(v) == "string" then
+                            theme[k] = v:gsub("^~/", homedir .. "/")
+                            theme[k] = v:gsub("^%./", theme_path .. "/")
```
blueyed: This will not expand "./" in case there is no homedir..
While this is not a typical case, the `if homedir` from above should be moved probably around the expansion of `~/`?

lib/beautiful/init.lua
```diff
@@ -111,22 +111,34 @@ end
...
+                        end
+                    end
+                -- expand '~'
+                elseif type(config) == 'table' then
+                    for k, v in pairs(theme) do
+                        if type(v) == "string" then theme[k] = v:gsub("^~/", homedir .. "/") end
```
blueyed: Why is `./` not expanded here?

blueyed:
@farkasity 
Why have you closed it?  Do you don't want to carry it on?

farkasity:
@blueyed
I'm busy at some other stuff these days, I do not have time to do it.
Maybe you can finishing it?

lib/beautiful/init.lua
farkasity: In this case, the `config` is a table instead of a file path, it cannot decide the base path unless specify one explicitly in the table.

blueyed:
@farkasity 
Did you have time by now or in the future for this?
Would be great if you could finish this (with out help, of course).

Elv13:
Addressed by  #1791, even if this would have been a little nicer.

PR #1090, opened 2016-09-12
awful.spawn: Use LGI's Async facilities

psychon:
Normally, writing asynchronous code with Gio results in an awful mess of
spaghetti code, because every operation needs a callback that is called when
this operation is done. To improve upon this, LGI provides some magic with
coroutines that hide the asyncronous nature of things and you end up with
"normal looking" code.

Use this Gio.Async facility in awful.spawn to make the code easier to follow.

Signed-off-by: Uli Schlachter <uli.schlachter@informatik.uni-oldenburg.de>

### DIFF
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -22,7 +22,6 @@ local capi =
 }
 local lgi = require("lgi")
 local Gio = lgi.Gio
-local GLib = lgi.GLib
 local util   = require("awful.util")
 local protected_call = require("gears.protected_call")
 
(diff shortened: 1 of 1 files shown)


psychon:
Sigh.

Lgi 0.7.1 causes `Error: running function for step 1/3 (@1): lib//awful/spawn.lua:271: attempt to index field 'Async' (a nil value)`. Can we raise our minimum lgi requirement?

Other Lua versions crashed with coroutine-related assertion failures and I do not have the time to debug them right now (nor can I reproduce anything locally).

Elv13:
> Can we raise our minimum lgi requirement?

yes

> Other Lua versions crashed with coroutine-related assertion failures

And I bet some other will just crash with a segmentation fault.

psychon:
Ok, I'll (hopefully remember to) look into this again after the release. We can still bump out minimum requirement then.

psychon:
Updated. Please review and decide if this should go in.

One of the Travis builds will fail, because this needs at least LGI 0.8.0. That will be fixed/handled by #1350.

blueyed:
I like the changed in general.

actionless:
as far as i remember lgi was failing when trying to initiate async call from a callback of an other async call

psychon:
This PR likely runs into an LGI issue: https://github.com/pavouk/lgi/pull/156
The backtraces look like it (function names fit), but it is not so easy to see, because the bug is basically "random memory corruption". I wonder if `menubar` in current master already runs into this problem...

PR #1096, opened 2016-09-15
Test & fix a bug with dbus.disconnect_signal

psychon:
The code was written so that it assumes that disconnecting the last signal also
removed the corresponding entry in the signal array. This lead e.g. to an
index-out-of-bounds access in some cases.


blueyed:
Thanks!

PR #1097, opened 2016-09-15
Lazy client geometry update

psychon:
This is my own version of  https://github.com/awesomeWM/awesome/pull/174 that should hopefully not cause any problems.

CC @blueyed


blueyed:
Thanks!

PR #1099, opened 2016-09-15
Fix #1091

### LINKED ISSUE #1091: Clients always open on first screen
Issue was originally reported [here](http://www.mail-archive.com/awesome@naquadah.org/msg08282.html), but I can reproduce it

```
awesome v3.5.2-1968-ge86ddaa-dirty (The Fox)
 • D-Bus support: ✔
```

- Move mouse pointer to some other screen than screen 1
- Open a terminal

**Actual result:**

Terminal opens on screen 1

**Expected result:**

Terminal opens on screen `awful.screen.focused()`

Looking at the diff of 21787444e4f1, this removes a call to `awful.screen.focused()`, but doesn't add one. Which part of the code is now responsible for assigning a default screen?

Elv13:
The test error will be easy enough to fix. Just disable the connect_sighnal("property::screen") in the first commit, then add a second commit to undo that and monkey-patch awful.screen.preferred to nil for the current tests, then put it back to its original value for the new tests we will need to prevent this from regressing. The connect_signal("property::screen") check is still necessary for the tests to be useful, as what they really test is that c.screen is set only once (to prevent the other use cases from regressing, again...). I don't have time for it now.

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -12,7 +12,8 @@ local capi =
 {
     mouse = mouse,
     screen = screen,
-    client = client
+    client = client,
+    awesome = awesome,
 }
 local util = require("awful.util")
 local object = require("gears.object")
@@ -211,6 +212,15 @@ function screen.object.set_padding(self, padding)
         data.padding[self] = padding
         self:emit_signal("padding")
     end
+end
+
+--- Return the currently screen for new clients.
+-- This is exactly the same as `awful.screen.focused` exept it avoids clients
+-- being moved when Awesome is restarted. This is used by the default `rc.lua`.
+-- @tparam client c A client
+-- @treturn screen The preferred screen.
+function screen.preferred(c)
+    return capi.awesome.startup and c.screen or screen.focused()
 end
 
 --- The defaults arguments for `awful.screen.focused`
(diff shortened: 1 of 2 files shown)


lib/awful/screen.lua
```diff
@@ -213,6 +214,15 @@ function screen.object.set_padding(self, padding)
     end
 end
 
+--- Return the currently screen for new clients.
```
blueyed: s/currently/current/

lib/awful/screen.lua
```diff
@@ -213,6 +214,15 @@ function screen.object.set_padding(self, padding)
...
+-- This is exactly the same as `awful.screen.focused` exept it avoids clients
```
blueyed: s/exept/except/

blueyed:
Thanks!

Elv13:
I will fix the build later. As said in the first comment, this commit break an assumption enforced by the multi-screen tests. Enforcing that assumption is still necessary to prevent your startup notification screen bug from regressing.

blueyed:
> I will fix the build later.

Thanks!
I somehow oversaw that, and did not pay attention apparently: I did not want to break master.

Elv13:
Don't worry, it isn't a bug, just a test failure due to the way the test are written. Master is not broken. Only the build is reported to have a failed minor test.

blueyed:
yeah, that's what I've meant - it also runs fine for me.. :)

Issue #1089, opened 2016-09-12
Implement a faster "client deamon" for the unit tests


psychon:
With commit 35c0476c86a6913208e84733d8f339f0c886c96c, a FIXME was added that basically says that `tests/_client.lua` is slow.

Here is my proposal for a new, faster implementation:
Child Lua process (awful.spawn.read_lines should likely be re-written in a similar way):
```lua
local lgi = require 'lgi'
local Gtk = lgi.require('Gtk')
local Gio = lgi.require('Gio')
Gtk.init()

local function open_window(class, title)
	local window = Gtk.Window {
		default_width  = 100,
		default_height = 100,
		title          = title
	}
	window:set_wmclass(class, class)
	window:show_all()
end

local function read_line(stream)
	local line, length = stream:async_read_line()
	if type(length) ~= "number" then
		error("Error reading line: " .. tostring(length))
	end
	-- Either return the line that was read or nil for EOF
	return line
end

local function read_stdin()
	local stdin = Gio.UnixInputStream.new(0, false)
	stdin = Gio.DataInputStream.new(stdin)
	
	for class in read_line, stdin do
		local title = read_line(stdin)
		if not title then break end
		open_window(class, title)
	end

	Gtk.main_quit()
end

local function error_handler(err)
	print(debug.traceback("Error: " .. tostring(err)))
	Gtk.main_quit()
end

Gio.Async.start(xpcall)(read_stdin, error_handler)

Gtk:main{...}
```
Code for `tests/_client.lua` (untested):
```lua
local lgi = require("lgi")
local Gio = lgi.require("Gio")

local initialized = false
local pipe
local function init()
  if initialized then return end
  initialized = true
  local _, _, stdin = capi.awesome.spawn(the_above_lua_code, false, true)
  stdin = Gio.UnixOutputStream.new(stdin, true)
  pipe = Gio.DataOutputStream.new(stdin)
end
return function(class, title, use_sn)
  init()
  if use_sn then error("No idea how to implement this right now") end
  local data = class .. "\n" .. title .. "\n"
  pipe:write_all(data, #data) -- TODO: Error handling?
end
```
For SN, something like the following is needed (but no idea how to implement that on the awesome-side...): https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-set-startup-id

psychon:
Adding to v3.6 milestone per https://github.com/awesomeWM/awesome/issues/1091#issuecomment-247447145

PR #1100, opened 2016-09-16
Rewrite tests/_client.lua

psychon:
The new client is hopefully faster. Why, you ask?

Instead of spawning a new Lua process each time a test asks for a new window,
there is a "daemon process" which gets commands to open new windows from its
standard input. That way, Lua doesn't have to load LGI all the time and lots of
pointless work is skipped. The daemon process exits when its stdin is closed and
thus should automatically exit when awesome exits.

Fixes: https://github.com/awesomeWM/awesome/issues/1089

### DIFF
--- a/tests/_client.lua
+++ b/tests/_client.lua
@@ -3,27 +3,101 @@ local spawn = require("awful.spawn")
 -- This file provide a simple, yet flexible, test client.
 -- It is used to test the `awful.rules`
 
-return function(class, title, use_sn)
-    class = class or 'test_app'
-    title = title or 'Awesome test client'
+local test_client_source = [[
+local lgi = require 'lgi'
+local Gtk = lgi.require('Gtk')
+local Gio = lgi.require('Gio')
+Gtk.init()
 
-    local cmd = {"lua" , "-e", table.concat {
-        "local lgi = require 'lgi';",
-        "local Gtk = lgi.require('Gtk');",
-        "Gtk.init();",
-        "local class = '",class,"';",
-        "local window = Gtk.Window {",
-        "    default_width  = 100,",
-        "    default_height = 100,",
-        "    on_destroy     = Gtk.main_quit,",
-        "    title          = '",title,
-        "'};",
-        "window:set_wmclass(class, class);",
-        "window:show_all();",
-        "Gtk:main{...}"
-    }}
+local function open_window(class, title, snid)
+    local window = Gtk.Window {
+        default_width  = 100,
+        default_height = 100,
+        title          = title
+    }
+    if snid ~= "" then
+        window:set_startup_id(snid)
+    end
+    window:set_wmclass(class, class)
+    window:show_all()
+end
 
-    return spawn(cmd, use_sn)
+-- Start a coroutine for nicer input handling
+local coro = coroutine.wrap(function()
+    while true do
+        local class = coroutine.yield()
+        local title = coroutine.yield()
+        local snid = coroutine.yield()
+        open_window(class, title, snid)
+    end
+end)
+coro()
+
+-- Read lines from stdin and feed them to the coroutine
+local stdin = Gio.UnixInputStream.new(0, false)
+stdin = Gio.DataInputStream.new(stdin)
+
+local read_start, read_finish
+read_start = function()
+    stdin:read_line_async(0, nil, read_finish)
+end
+read_finish = function(...)
+    local line, length = stdin.read_line_finish(...)
+    if type(length) ~= "number" then
+        error("Error reading line: " .. tostring(length))
+    end
+
+    local eof = not line -- Behaviour of up-to-date LGI
+            or (tostring(line) == "" and #line ~= length) -- (Ab)use uninitialized variable
+    if eof then
+        Gtk.main_quit()
+    else
+        coro(tostring(line)) -- tostring() needed for older LGI versions
+        read_start()
+    end
+end
+
+read_start()
+Gtk:main{...}
+]]
+
+local lgi = require("lgi")
+local Gio = lgi.require("Gio")
+
+local initialized = false
+local pipe
+local function init()
+    if initialized then return end
+    initialized = true
+    local cmd = { "lua", "-e", test_client_source }
+    local _, _, stdin, stdout, stderr = awesome.spawn(cmd, false, true, true, true)
+    pipe = Gio.UnixOutputStream.new(stdin, true)
+    stdout = Gio.UnixInputStream.new(stdout, true)
+    stderr = Gio.UnixInputStream.new(stderr, true)
+    spawn.read_lines(stdout, function(...) print("_client", ...) end)
+    spawn.read_lines(stderr, function(...) print("_client", ...) end)
+end
+
+-- Hack needed for awesome's Startup Notification machinery
+local function get_snid(sn_rules)
+    local success, snid = spawn({ "/bin/true" }, sn_rules)
+    assert(success)
+    assert(snid)
+    return snid
+end
+
+return function(class, title, sn_rules)
+    class = class or "test_app"
+    title = title or "Awesome test client"
+
+    init()
+    local snid = sn_rules and get_snid(sn_rules) or ""
+    local data = class .. "\n" .. title .. "\n" .. snid .. "\n"
+    local success, msg = pipe:write_all(data)
+    assert(success, msg)
+
+    -- TODO: Fix the API of this function
+     return true, sn_rules and snid or nil
 end
 
 -- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80


psychon:
Whoops. `use_sn` isn't a boolean, but can be a table.... intuitive. Let's rename that.
```diff
diff --git a/tests/_client.lua b/tests/_client.lua
index fc0378f..1f1252f 100644
--- a/tests/_client.lua
+++ b/tests/_client.lua
@@ -81,25 +81,25 @@ local function init()
 end
 
 -- Hack needed for awesome's Startup Notification machinery
-local function get_snid()
-    local success, snid = awesome.spawn({ "/bin/true" }, true)
+local function get_snid(sn_rules)
+    local success, snid = spawn({ "/bin/true" }, sn_rules)
     assert(success)
     assert(snid)
     return snid
 end
 
-return function(class, title, use_sn)
+return function(class, title, sn_rules)
     class = class or "test_app"
     title = title or "Awesome test client"
 
     init()
-    local snid = use_sn and get_snid() or ""
+    local snid = sn_rules and get_snid(sn_rules) or ""
     local data = class .. "\n" .. title .. "\n" .. snid .. "\n"
     local success, msg = pipe:write_all(data)
     assert(success, msg)
 
     -- TODO: Fix the API of this function
-     return true, use_sn and snid or nil
+     return true, sn_rules and snid or nil
 end
 
 -- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
```
(Yup, I'm talking to myself)

Elv13:
Thanks for this. One of the CI target did hang on the unrelated layout test. That test is supposed to spawn 5 clients. But those are supposed to be xterm...

tests/_client.lua
```diff
@@ -3,27 +3,103 @@ local spawn = require("awful.spawn")
...
-
-    return spawn(cmd, use_sn)
+local test_client_source = [[
+local lgi = require 'lgi'
+local Gtk = lgi.require('Gtk')
+local Gio = lgi.require('Gio')
+Gtk.init()
+
+local function open_window(class, title, snid)
+    print("open", class, title, snid)
```
psychon: Whoops. That print shouldn't be there, I think

psychon:
I tried restarting the build a couple of times, but the failure stays (I managed to get rid of the LuaJIT failure that way). So, something is weird in the land of Lua 5.2 that makes [this assert](https://github.com/psychon/awesome/blob/3e15585a026324c1f40541fef3a9d23dd719752d/tests/test-awful-client.lua#L150) always fail. No idea what that could be, but I guess I'll have to install Lua 5.2 on my computer.

psychon:
@Elv13 I don't understand "test-awful-client.lua". The test asserts that `property::screen` is only emitted once. However, how is this supposed to work? There is a race between `awful.rules` applying the "normal" rules and the `sn`-specific rules. Both are applied when the "manage" signal comes in and both are applied independently of each other.
Thus, the unexpected screen change that causes a failure on Travis right now is `awful.rules` applying the default rule.

However, I have no idea why this works before my patch and on current master...

Elv13:
See https://github.com/awesomeWM/awesome/pull/1099 comments. There was a little misunderstanding and we actually merged this too soon. Originally, I planned this PR to be merged first. For now, please monkey-patch awful.screen.preferred to nil so the tests pass again.

psychon:
Nope. I did not rebase this on master. The builds here do not include that code yet, but fail independently of it (for the same reason? dunno).

psychon:
Hrm. Building https://github.com/awesomeWM/awesome/pull/1100/commits/486f9202d572dc7235a890221017a2cc0122f2a4 produces https://travis-ci.org/awesomeWM/awesome/builds/162435971.
The daemon process for spawning new windows gets the correct strings (as shown by the results of some print()s that it does), but Gtk ends up creating windows with e.g. name `Awesome test client��+` or class `screen2ce-79dcf9A`. This seems to only happen with LGI 0.7.1, but I have no idea why (I didn't spot anything obvious in the LGI commits). The class string looks like it partly contains the startup id, but I don't see how this makes any sense...
The test fails in the end, because the wrong class means that the client doesn't match the rule that it is expected to match and thus isn't assigned to the right screen.

psychon:
Looks like before 0.7.2-25-g9240e4e, instead of producing strings, LGI produced its own "byte buffer" type for e.g. reading lines and this did not guarantee null-terminated strings. I pushed a WIP commit which fixes this (well, works around) via a call to `tostring()`. This hopefully means that everything works correctly now. I'll do some experiments...

psychon:
I think this can be merged now.

------

I also did some benchmarks: `tests/_client.lua` really is faster now. The time for calling it stays roughly the same, but the time until the "manage" signal is smaller. However, this does not reduce the time that `tests/run.sh` takes (on my not-slow computer), apparently because the old code was also fast enough so that the new window is managed before the next iteration of the test runner.

So I took a the test runner: It uses a timer with a timeout of 0.1 seconds and calls the current step each time that this timer fires. All together, it calls a step function 498 times for all tests. Ignoring things like "time needed to start awesome", this means just the timer needs about 50 seconds. The test run took 90 seconds wall clock time.
The worst offender for "many step-calls" is `test-awful-layout.lua` with 183 step-calls. Second place goes to both `test-awful-tag.lua` and `test-awful-client.lua` each with 63 calls. Both of these later tests also have awesome spitting out warnings like "Last main loop iteration took 0.7 seconds". Next is `test-resize.lua` with 40 calls and then the numbers get rapidly smaller.
Also, I did some timing on individual tests via `cd tests ; for x in test-*lua ; do /bin/time ./run.sh $x ; done` (note that this includes time needed to start Xvfb and awesome!). `test-awful-layout.lua` took 20 seconds, `test-awful-client.lua` took 11 seconds, `test-awful-tag.lua` took 10 seconds and `tests-resize.lua` took 5 seconds. The rest are around two or one seconds.

It looks like slow tests are those with many steps. Oh and those with many steps are exactly those using the new `_multi_screen.lua` (`test-awful-client.lua` and `test-awful-tag.lua`).

Perhaps the best approach to speeding things up is to make `_multi_screen.lua` a bit more harmless. The multi-screen part of `test-awful-tag.lua` only cares about the number of screens and not their geometry and `test-awful-client.lua` can surely be a good test without having to go through 12 screen configurations.

Another idea would be to speed up `_runner`. For example, the following patch halves its runtime (48 seconds instead of 90 seconds, but again that's on my fast computer):
```diff
diff --git a/tests/_runner.lua b/tests/_runner.lua
index 56860de..d6d3c7d 100644
--- a/tests/_runner.lua
+++ b/tests/_runner.lua
@@ -34,8 +34,9 @@ end)
 runner.run_steps = function(steps)
   -- Setup timer/timeout to limit waiting for signal and quitting awesome.
   -- This would be common for all tests.
-  local t = timer({timeout=0.1})
-  local wait=20
+  local t = timer({timeout=0.01})
+  local initial_wait=200
+  local wait=initial_wait
   local step=1
   local step_count=0
   assert(not running, "run_steps() was called twice")
@@ -65,7 +66,7 @@ runner.run_steps = function(steps)
         -- Next step.
         step = step+1
         step_count = 0
-        wait = 5
+        wait = initial_wait
         t:again()
         return
       end
```

tests/_client.lua
```diff
@@ -3,27 +3,103 @@ local spawn = require("awful.spawn")
...
+    else
+        coro(tostring(line)) -- tostring() needed for older LGI versions
+        read_start()
+    end
+end
+
+read_start()
+Gtk:main{...}
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
```
blueyed: Huh?  Is this useful in some context?

blueyed:
Thanks for your investigations!

btw: we could always have a normal / lighter test run by default, and then Travis still doing some more work, i.e. testing those 12 screens.

I think your last patch makes sense - the `0.1` has been taken probably arbitrarily, and it has not mattered much when there were fewer tests.
But we can add it with some other PR, too.

Elv13:
Actually, this isn't the main problem. This is (well, I think and have statistical evidences to back it)

```lua
local test_wallpaper_scaling = nil

local w,h = root.size()
local real = screen[1]
screen[1]:fake_resize( 0, 0, 1, 1 )
local start = os.time()

local function add_all()
    for i=0, 15 do
        screen.fake_add(i*w/16, 0, w, h)
    end
end

local function rem_all()
    local scrs = {}
    for s in screen do
        table.insert(scrs, s)
    end

    for k,s in ipairs(scrs) do
        if s.valid and s ~= real then
            s:fake_remove()
        end
    end

    add_all()
end

local counter = 1
local delta = 0

test_wallpaper_scaling = function ()
    for s in screen do
        local sur = surface.widget_to_surface(
            wibox.widget {
                {
                    widget = wibox.widget.imagebox
                },
                bg     = "#ff0000",
                        forced_width =s.geometry.width,
                        forced_height =s.geometry.height,
            },
            s.geometry.width,
            s.geometry.height
        )
        gears.wallpaper.fit(sur, s, "#0000ff")
    end

    print("run:", counter, os.time() - delta)
    delta = os.time()
    counter = counter + 1

    -- 15 minutes should be enough
    if delta - start >= 900 then
        awesome.quit()
    end

    gears.timer.delayed_call(test_wallpaper_scaling)
    rem_all()
end

test_wallpaper_scaling()
```

On a slow computer (it takes like 200 iteration to reach 7 seconds on my desktop):

```
run:	2	2
run:	3	5
run:	4	5
run:	5	7
run:	6	7
run:	7	7
run:	8	7
run:	9	8
run:	10	8
run:	11	9
run:	12	9
run:	13	9
run:	14	8
run:	15	10
run:	16	10
run:	17	10
run:	18	10
run:	19	11
run:	20	11
run:	21	12
```

It matches the "mainloop iteration time" messages from the tests. It grows at the same pace. I think there is a strong correlation. Letting it run the full time for 15 minutes keep growing in a linear pattern like this.

From heaptrack memory consumption and object leaked memory after 5 minutes:

slider8

image

(@psychon and yes, now you understand why I don't recommend to use massif anymore, viva heaptrack)

tests/_client.lua
psychon: Nope. :-)
I wrote that file by hand at first and out of a habbit added a vim modeline. Then I copied the whole file into this Lua file. It could be useful if anyone copies out this chunk of code in its own file to play with it and edit it...
Do you want me to remove this modeline?

psychon:
> First, is it possible to re-use the wallpaper surface so there is only 1 full allocation per wallpaper

I don't really know how (outside of the test runner). The `gears.wallpaper` functions that get a surface cannot assume ownership of this surface and definitely must not modify some random surface that they got. Even when loading stuff from files, `gears.surface` keeps a cache (in a weak table) which means that these surfaces may not be modified or the cache starts returning wrong data.

> So, long story short, the screens leaks. I didn't investigate why yet. 

I think I can tell you why. Or at least partly (this shouldn't cause sooo much garbage).

Imagine you have a wibox that becomes garbage (no more references, `.visible = false`). This wibox is still weakly reachable via e.g. `wibox.drawable.drawables_force_complete_repaint`. When the wallpaper is changed, we force a repaint via this table and the wibox becomes strongly reachable. This means that it cannot be GC'd for this GC cycle.
Now, if the wallpaper changes faster than the garbage collector can do its thing, a wibox will never be considered garbage. This will then keep all the widgets that are specific to this wibox alive.

This would also (partly) explain why your test runs slower and slower: There are more and more repaints to handle on each wallpaper change.

We could optimize this if either a compositing manager is running or the wibox has a non-transparent background (because in these cases the repaint isn't actually needed). Back then I was too lazy for this and left a comment in the code saying that this could be optimised. Perhaps the time has come?

Elv13:
Make sense. Now we have a clearer picture of why the test is so slow. And yes, this need to be fixed and regression tested, if only for the tests themselves. Over time, more and more tests will have a multi-screen section. Of course, as said in some _multi_screen.lua, not all scenarios need the full disposition check, so multiple "disposition suits" should be made available to relevant tests.

psychon:
One more micro-fake-test: Add the following to the end of the default config
```lua
local max = {}
local objs = { screen = screen, button = button, tag = tag, drawable = drawable, drawin = drawin, client = client, key = key }
gears.timer.start_new(0.01, function()
	for k, v in pairs(objs) do
		local instances = v.instances()
		if not max[k] or max[k] < instances then
			max[k] = instances
			print(string.format("There are now %d %s alive", max[k], k))
		end
	end
	return true
end)
```
This provides the maximum count on how many objects of a certain type are alive.
After a full test suite run, the results over all tests:
At most 315 tags, 28 screens, 312 keys, 66 drawables, 29 drawins, 37 clients and 7972 buttons alive. That doesn't sound all that bad...  Now for the slow tests:
`test-awful-client.lua` peaks at 252 tags, 28 screens, 288 keys, 66 drawables, 29 drawins, 37 clients and 7880 buttons.
`test-awful-tag.lua` peaks at 315 tags, 28 screens, 288 keys, 31 drawables, 29 drawins, 2 clients and 7972 buttons.

psychon:
@Elv13 Could you rerun your test with the following patch applied: https://gist.github.com/psychon/c262f1ba6923a47b7f906ab2bef5f275 Does the patch help? (I wouldn't submit this patch as a review, because this might mean that a wibox with a transparent background that is hidden, the wallpaper is changed and then it is made visible again might still contain the old background for a tiny moment.)
Also, should we move this discussion to a dedicated issues for tracking it?

Issue #1049, opened 2016-08-21
Travis: codecov coverage went down by 10%


Elv13:
See https://codecov.io/gh/awesomeWM/awesome/commits

Many files are missing from (for example) `lib/awful` and coverage for other files went for ~90% to ~60% (like placement.lua, and this one really has 90%, for all the **** time it took me to get there). Coverall seems unaffected.

I didn't investigate yet.

psychon:
Looks like the functional tests went missing. I cannot find any file which has any kind of coverage from functional tests. I don't have much time to investigate either.

blueyed:
The "do_codecov functionaltests" part seems to run.  Search for "functionaltest" in e.g. https://s3.amazonaws.com/archive.travis-ci.org/jobs/153899494/log.txt.

blueyed:
Some example from #1088:
Searching for "&flags" in the build output (https://s3.amazonaws.com/archive.travis-ci.org/jobs/159271723/log.txt) shows that there are three uploads.
In https://codecov.io/gh/awesomeWM/awesome/tree/45e8c6083c047f0278ca8cf4cb8f22450a2bcc67 however I cannot see the flags (as documented in https://codecov.readme.io/docs/flags#section-usage).

While I can imagine that we end up in the wrong directory through https://github.com/awesomeWM/awesome/blob/0014b275a80bf35ab5646e7d5b840a675cbd8962/.travis.yml#L159-L160, I've contacted the codecov support already, since I would assume that flags should show up anyway, and there's maybe (another) issue with that.

blueyed:
This seems to be caused by e723d18 (#1011), by looking at https://codecov.io/gh/awesomeWM/awesome/commits?page=3.

psychon:
Thanks for looking into that! So I guess this means that more than just awesome was using `$LUA_PATH`....

stevepeak:
Hey @blueyed which flag is missing? I'm still [seeing 3 uploads](https://codecov.io/gh/awesomeWM/awesome/commit/78300f24d5025630ed5f655db64d1e5c7f123d4e/builds) and [three flags here](https://codecov.io/gh/awesomeWM/awesome/src/78300f24d5025630ed5f655db64d1e5c7f123d4e/tests/examples/text/template.lua)

Thanks!

blueyed:
@stevepeak 
Thanks for looking into it!  I was expecting to see the flags also in the dir view apparently.
[...]

PR #1105, opened 2016-09-18
Travis: fix codecov

blueyed:
Trying to fix https://github.com/awesomeWM/awesome/issues/1049

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -82,11 +82,10 @@ fi
 
 cd $build_dir
 
-AWESOME_OPTIONS="$AWESOME_OPTIONS --search lib/"
+AWESOME_OPTIONS="$AWESOME_OPTIONS --search lib"
 # Add test dir (for _runner.lua).
 AWESOME_OPTIONS="$AWESOME_OPTIONS --search $this_dir"
-XDG_CONFIG_HOME="./"
-export XDG_CONFIG_HOME
+export XDG_CONFIG_HOME="./"
 
 cd - >/dev/null
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,7 +100,7 @@ install:
     do_codecov() {
       if [ "$DO_COVERAGE" = "codecov" ]; then
         (cd build && luacov) || return 1
-        travis_retry bash /tmp/codecov-bash -F "$1" || return 1
+        travis_retry bash /tmp/codecov-bash -X coveragepy -c -F "$1" || return 1
         rm build/luacov.stats.out
       fi
       return 0


blueyed:
@psychon 
I assume this to be ready, without knowing why it's needed.
Do  you have a clue?

Elv13:
👍 Thanks for taking care of this

tests/run.sh
```diff
@@ -85,8 +85,11 @@ cd $build_dir
 AWESOME_OPTIONS="$AWESOME_OPTIONS --search lib/"
 # Add test dir (for _runner.lua).
 AWESOME_OPTIONS="$AWESOME_OPTIONS --search $this_dir"
-XDG_CONFIG_HOME="./"
-export XDG_CONFIG_HOME
+export XDG_CONFIG_HOME="./"
+
+# Set LUA_PATH for codecov.
+# (https://github.com/awesomeWM/awesome/issues/1049#issuecomment-247876195)
+export LUA_PATH="$(lua -e 'print(package.path)');lib/?.lua;lib/?/init.lua"
```
psychon: Unrelated, but I just stumbled upon the following:

> Any ";;" in the value of the environment variable is replaced by the default path. 

https://www.lua.org/manual/5.1/manual.html#pdf-LUA_PATH

I think we should make use of that. :-)

psychon:
Nope, no idea. But `tests/run.sh` only runs awesome, so the difference must be in luacov's "statistics collector" part. However, the idea of this `--search`-flag was to have the same end result in `package.path`.

I have two wishes before this is merged (sorry, I cannot easily test this myself right now):
* Check if the following patch also fixes the problem
* `print(package.path)` at the beginning of rc.lua for a working/non-working build and compare. The important difference *must* be in there.
```diff
diff --git a/tests/run.sh b/tests/run.sh
index b04d62c..1288771 100755
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -82,7 +82,7 @@ fi
 
 cd $build_dir
 
-AWESOME_OPTIONS="$AWESOME_OPTIONS --search lib/"
+AWESOME_OPTIONS="$AWESOME_OPTIONS --search lib"
 # Add test dir (for _runner.lua).
 AWESOME_OPTIONS="$AWESOME_OPTIONS --search $this_dir"
 XDG_CONFIG_HOME="./"
```

blueyed:
@psychon 
Looks like removing the slash fixes it!

psychon:
@blueyed Thanks. Sounds like we should use that as a fix instead. Can you squash the first two commits and remove the third one in this PR?

I guess this means that codecov really fails to see that "lib//awful/util.lua" and "lib/awful/util.lua" is the same file and silently drops coverage information for "lib//awful/util.lua".
CC @stevepeak Just for your information

stevepeak:
@psychon our backend would fix `lib//awful/util.lua` to `lib/awful/util.lua` and merge and duplicate coverage details. Have you discovered otherwise?

psychon:
@stevepeak Yup, looks like that fails for us: https://github.com/awesomeWM/awesome/issues/1049#issuecomment-247928575 (but note that the file names are more complicated than just `/` vs `//`, see the link for working and non-working example)

psychon:
@stevepeak More specifically: No file seems to have `functionaltests`-flag coverage when the upload uses `lib//`, but it works correctly when AFAIK the only difference is turning `lib//` into `lib/`.

stevepeak:
I see talk about `luacov-coveralls` but we are Codecov and do not use this package. Maybe that is the issue?

stevepeak:
I think I found the issue. I'll ping you back here shortly

psychon:
@stevepeak Let's just say "the situation there is complex". ;-) (One travis build handles codecov, another one handles coveralls; together with using different Lua versions etc). But yeah, coveralls is a lot more picky about the path names that it gets and caused me a lot of lost hair to get working.

stevepeak:
Ok. I was wrong. It is not fixing `lib//awful/util.lua` properly. I'll make a server side patch now to fix 👍

psychon:
Thanks a lot!

For this PR: I'm still in favor of removing that one slash (plus the other nice python-"disablation" that Daniel found)

blueyed:
@psychon 
As for this PR, we can still use it?  (it is squashed)

Let's wait / see if codecov handles the extra slash now correctly, before merging it?

stevepeak:
@blueyed deployed that fix yesterday 👍

psychon:
@blueyed :+1: (but feel free to wait a bit for some other PR if you want. Any "random" PR should see a coverage jump from 60% to 70% due to codecov's server side fix.)

PR #1106, opened 2016-09-19
The example test runner now unsets version-specific LUA_PATH before proceeding

Crazylemon64:
When lua version-specific `LUA_PATH` values were set, they would take precedence over plain `LUA_PATH`, and the example test suites would fail to find `common_template`. By unsetting them during the tests, we are able to safely run the tests with the path as we expect it.


Elv13:
Should we check if they are set instead, the just update the proper one(s)? Otherwise, fine by me.

Issue #1107, opened 2016-09-21
Huge mouse-click delays


madduck:
Output of `awesome --version`:
awesome v3.5.2-2001-gb4d6bfa (The Fox)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔

**How to reproduce the issue:**

Start it, use the mouse a bit, observe.

**Actual result:**

Between click of a mouse button and reaction in X client, there's an increasing delay up to several seconds.

**Expected result:**

Instantaneous reaction.

Here's why I think this is a problem with awesome:

In short, because if I replace awesome with fvwm2, the problems go away. If I replace fvwm with awesome again, the problem comes back.

The issue appears with evdev, as well as libinput, with the internal trackpad, as well as an external mouse. xev and evemu-record both see the event immediately. All clients experience the delay, no matter what framework, including pure xterm. Curiously, I don't  know why xev (which has its own window) doesn't experience the delay, but I suppose it hooks into the driver or hardware much higher up. Neither X server nor any drivers were updated before the problem appeared, but the issues did start in the wake up replacing awesom 3.4.15 with current Git master, and they exist too with stock rc.lua.

I'll try to record a video as soon as I am back in a work environment.

psychon:
Could you try running awesome with `--no-argb`? The theory is that not the "actual input" is slow, but updating the screen is slow. The reason for this theory is that `xev` doesn't have delays, but it really should have...
Also, could you check your `~/.xsession-errors` (or wherever awesome's stderr goes to; `ls -l /proc/$(pidof awesome)/fd/2` might help you (or not)) for messages like the following:
```
2016-09-21 14:31:16 W: awesome: a_glib_poll:373: Last main loop iteration took 0.159048 seconds! Increasing limit for this warning to that value.
```
Such messages (with times of "several seconds") would really point towards awesome.
If so, what does your `rc.lua` look like? Are you doing anything slow in there?

madduck:
Hey,

I am already running with `no-argb`. And unfortunately, there is no message of the type you're looking for anywhere… :/

The `rc.lua` is from the Git repo, i.e. I can reproduce this from stock.

Here is the video (250Mb!): http://scratch.madduck.net/.tmp__20160921_133528.mp4

juw:
Hi,

[...]

Issue #1103, opened 2016-09-18
magnifier layout: attempt to index local `focus` (a nil value)


psychon:
Output of `awesome --version`: Sorry, it's on another computer. But it's commit 1be0964

https://github.com/awesomeWM/awesome/blob/master/lib/awful/layout/suit/magnifier.lua#L71

`focus` was nil here at some random time when I really wanted to experiment with RandR changes...
I guess this `if` should be merged with the previous if.

Bug was introduced here: https://github.com/awesomeWM/awesome/commit/82342f055cc5c46fecdb82b324563f428ce929a9#diff-13d00ab11942bb012af76654ff3e06dfL70

Issue #1102, opened 2016-09-18
Default config: Re-set wallpaper on property::geometry of a screen


psychon:
I just swapped the position of two monitors (Before: 1|3|2; After: 1|2|3). 2 and 3 have different resolutions and so I noticed that the wallpaper isn't updated and still displays "the old state of things".

The default config has to set a wallpaper:
- When a new screen appears
- When a screen changes its geometry

blueyed:
Sounds like the issue I was seeing already - but could not find it now.

PR #1111, opened 2016-09-23
Some assorted fixes

psychon:
Lots of small fixes that hopefully do not introduce any regressions. Instead of opening several PRs for all of them individually, I put all of them here.

This should close at least #1103 and #1102, should help with #982 and also fixes some issues that weren't even reported yet (assign different DPis to your screens, move a client from one screen to the other, notice that the titlebar isn't updated; however, the size of the titlebar does still not react to DPI changes...).

This conflicts with #1070. @Elv13 do we really need a `__gc` callback or should this approach of making `.valid` work on GC'd objects be enough?

Edit: Now also has a patch for #1107.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -165,8 +165,7 @@ mytasklist.buttons = awful.util.table.join(
                                               awful.client.focus.byidx(-1)
                                           end))
 
-awful.screen.connect_for_each_screen(function(s)
-    -- Wallpaper
+local function set_wallpaper(s)
     if beautiful.wallpaper then
         local wallpaper = beautiful.wallpaper
         -- If wallpaper is a function, call it with the screen
@@ -175,6 +174,14 @@ awful.screen.connect_for_each_screen(function(s)
         end
         gears.wallpaper.maximized(wallpaper, s, true)
     end
+end
+
+-- Re-set wallpaper when a screen's geometry changes (e.g. different resolution)
+screen.connect_signal("property::geometry", set_wallpaper)
+
+awful.screen.connect_for_each_screen(function(s)
+    -- Wallpaper
+    set_wallpaper(s)
 
     -- Each screen has its own tag table.
     awful.tag({ "1", "2", "3", "4", "5", "6", "7", "8", "9" }, s, awful.layout.layouts[1])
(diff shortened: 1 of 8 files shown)


Elv13:
> This conflicts with #1070. @Elv13 do we really need a __gc callback or should this approach of making .valid work on GC'd objects be enough?

#1070 isn't really intended to be merged. It is more an "issue" than a pull request. I just attached some of the test commits for further inspection. It is still a bad hack and while it could be turned into a solution, something better is welcome.

awesomerc.lua
```diff
@@ -175,6 +174,14 @@ awful.screen.connect_for_each_screen(function(s)
         end
         gears.wallpaper.maximized(wallpaper, s, true)
     end
+end
+
+-- Re-set wallpaper when a screen's geometry changes (e.g. different resolution)
+screen.connect_signal("property::geometry", set_wallpaper)
```
Elv13: Should we add a "connect_for_each_screen_change" into `awful.screen`? It seems like it will become a common use case.

Elv13:
Also, hooray! Cocecov broke again...

blueyed:
> Cocecov broke again (nothing to do with your commits)

I thought it's because of build failures, but have not looked to closely?!

awesomerc.lua
blueyed: @Elv13 
Do you mean to have a generic wrapper to connect to different signals then?  Do you have an example?
In this case I think it is clearer to specify it explicitly like it's done here.

psychon: Does removing a screen count as a screen change? You definitely don't want to set a wallpaper there (quite a wasted effort and it might actually cause "the wrong thing" to happen, if some other screen takes the place of the old one). So I'm not really sure about the semantics of `awful.screen.connect_for_each_screen_change`.

psychon:
Yup. I think the build is aborted when `tests/run.sh` fails and thus no coverage info is uploaded for the functional tests.

blueyed:
Thanks a lot! :+1: 
I cannot use the new review functionality for some reason here.

blueyed:
:fireworks:

PR #1112, opened 2016-09-24
Upstream shape API p4.01

Elv13:
Some commits from #1044 with fixes suggested by @psychon. I only ignored the checkbox one, because, yeah, I disagree. If the `draw` function has some non rectangular parameters (for example, `fit` being ignored by a `flex` layout), then it look corrupted because (1) not all shape scale the same way on the *x* and *y* axis and (2) if a different shape is used for the "checked" part, then it just doesn't work. By forcing the square area, then at least it makes the ratio (and output) predictable.


psychon:
Thanks! :+1:

PR #1113, opened 2016-09-24
Upstream shape API p4.02

Elv13:
Some documentation related commits from #1044 that could be cherry-picked without much work.


docs/06-appearance.md.lua
```diff
@@ -0,0 +1,224 @@
...
+    for _, module in ipairs {
+        "awful", "wibox", "gears", "naughty", "menubar", "beautiful"
+    } do
+        local match = path:match("/"..module.."/([^.]+).lua")
+        if match then
+            return module.."."..match:gsub("/",".")
+        end
+    end
+
+    error()
```
psychon: How about `error("Cannot figure out module for " .. tostring(path))`?

docs/06-appearance.md.lua
```diff
@@ -0,0 +1,224 @@
...
+
+local beautiful_vars = {}
+
+-- Find all @beautiful doc entries
+for _,file in ipairs(all_files) do
+    local f = io.open(file)
+
+    local buffer = ""
+
+    for line in function() return f:read("*line") end do
```
psychon: I think `for line in f:lines() do` should work

awesomeConfig.cmake
```diff
@@ -340,6 +340,18 @@ foreach(file ${AWESOME_CONFIGURE_FILES})
 endforeach()
 #}}}
 
+# {{{ Generate some aggregated documentation from lua script
+execute_process(
+        COMMAND lua ${SOURCE_DIR}/docs/06-appearance.md.lua
+        ${BUILD_DIR}/docs/06-appearance.md
+        ERROR_VARIABLE  BEAUTIFUL_ERROR
+)
```
psychon: Could this use `add_custom_command()` instead so that CMake does not become even slower? (and so that this can be run in parallel to other things).
Something like the following might do:
```CMake
add_custom_target(OUTPUT docs/06-appearence.md
    COMMAND lua "${SOURCE_DIR}/docs/06-appearance.md.lua" "${BUILD_DIR}/docs/06-appearance.md")
```
(or one of the alternative forms of `add_custom_target()` that CMake has)

awesomeConfig.cmake
```diff
@@ -352,6 +351,20 @@ if(BEAUTIFUL_ERROR)
     message(FATAL_ERROR "Count not generate the theme index")
 endif()
 
+execute_process(
+        COMMAND lua ${SOURCE_DIR}/docs/05-awesomerc.md.lua
+        ${BUILD_DIR}/docs/05-awesomerc.md ${SOURCE_DIR}/awesomerc.lua
+        ${BUILD_DIR}/awesomerc.lua
+        ERROR_VARIABLE  RC_ERROR
+)
```
psychon: Could this also be moved to the build phase? (add_custom_target)

docs/05-awesomerc.md.lua
```diff
@@ -0,0 +1,241 @@
...
+
+-- Parse the default awesomerc.lua
+
+local rc = io.open(rcfile)
+
+local doc_block = false
+
+local output = {}
+
+for line in function() return rc:read("*line") end do
```
psychon: `for line in rc:lines() do`?

psychon:
Looks ok to me besides the above, minor comments.

PR #1114, opened 2016-09-24
tests: Fix awful.screen tests

Elv13:
To ensure that some features such as SNID rules work, we need
to ensure that the screen isn't set by other code paths. Only
a single algorithm can be executed for the screen. As soon
as many algorithms are executed on events such as "manage", it
will most likely regress again.

This commit make sure of that by disabling the default normal source
of c.screen. After that, any other c.screen changes can be
considered bugs.

### DIFF
--- a/tests/test-awful-client.lua
+++ b/tests/test-awful-client.lua
@@ -79,6 +79,9 @@ end
 
 local has_error
 
+-- Disable awful.screen.preferred(c)
+awful.rules.rules[1].properties.screen = nil
+
 table.insert(steps, function()
     -- Make sure there is no extra callbacks that causes double screen changes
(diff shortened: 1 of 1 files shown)


psychon:
I can't say I'm happy with this, but it gets the job done, so :+1:

psychon:
@Elv13 What do you think about something like https://github.com/psychon/awesome/commit/aba28fc9c ? The idea is to move SN-based rules into `awful.rules` and merge them with the to-be-applied rules before anything is done. That way `awful.spawn` does not have to connect to `manage`.
I think that should fix things enough so that this commit can be reverted (but testing on Travis said otherwise for the old-lgi-build).

PR #822, opened 2016-04-12
tests: Prevent locking

Elv13:
I need to figure out why I need this on my laptop

PR #1116, opened 2016-09-24
tests/run.sh: Use dbus-send directly with low timeouts

psychon:
run.sh waits for awesome's startup to be done by having awesome execute "return
1" via its dbus interface. However, by default dbus has a 25 second timeout
before it fails a dbus-send invocation. This defeats the purpose of this
exercise.

So instead of using awesome-client, this commit makes the code use dbus-send
directly and specifies a relative low reply timeout (which should still be
plenty so that this doesn't erroneously fail).

This hopefully supersedes #822.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -187,7 +187,8 @@ start_awesome() {
     cd - >/dev/null
 
     # Wait until the interface for awesome-client is ready (D-Bus interface).
-    wait_until_success "wait for awesome startup via awesome-client" "DISPLAY=$D '$AWESOME_CLIENT' 'return 1' 2>&1"
+    # Do this with dbus-send so that we can specify a low --reply-timeout
+    wait_until_success "wait for awesome startup via awesome-client" "dbus-send --reply-timeout=100 --dest=org.naquadah.awesome.awful --print-reply / org.naquadah.awesome.awful.Remote.Eval 'string:return 1' 2>&1"
 }
 
 # Count errors.


blueyed:
In this regard: what about different awesome instances?

We could use different dbus paths per instance, with $DISPLAY maybe?

Setting AWESOME_DBUS_DEST in the awesome session might help with that, and awesome-client would use the path with AWESOME_DBUS_DEST in it by default?!

Elv13:
> In this regard: what about different awesome instances?

We already had `malloc()`/Quotas failures in the logs, I don't think it will make it any better. I guess it will just cause more random build failures on some Travis nodes.

blueyed:
@Elv13 
I was thinking about usage of awesome-client in general.

IIRC I am using a separate dbus session inside of awesome for that, but just wondered recently about that again.

Elv13:
I also "kind of" do that. I use a whole user account testing. So a separate session, user and config. I can `git pull` my main config from my devel one when it's ready.

I also have a bunch of tools and some zsh functions to list clients remotely and such. I also have scripts that call awful.spawn with SNID support.

```sh
#!/bin/bash

# If Xephyr is started with a command like "Xephyr :1", this will be set
# to 1
XE_SCREEN=$(ps aux | grep Xephyr | grep -Eo ' :[0-9.]+' | grep -Eo '[0-9.]+')

for AW_PID in $(pidof awesome); do
    AW_DIS=$(grep -Eoa 'DISPLAY=:[0-9.]+' /proc/$AW_PID/environ | grep -Eo '[0-9.]+')
    if [ "$AW_DIS" == "$XE_SCREEN" ]; then
        echo FOUND PID $AW_PID at DISPLAY $AW_DIS
        echo 'awesome.restart()' | DISPLAY=:$AW_DIS awesome-client
        exit 0
    fi
done

echo Awesome not found

exit 1
```

blueyed:
@Elv13 
>         echo 'awesome.restart()' | DISPLAY=:$AW_DIS awesome-client

How does `DISPLAY` get used there?  Does dbus make use of it?

Elv13:
> How does DISPLAY get used there? Does dbus make use of it?

Long story. Back in the early days, there was something called DCOP. It was replaced by DBus when a common IPC had to be chosen for XDG standardization. DCOP was tied to X11 session and used X messages to communication between clients and the WM. It was a sound design for GUI apps, but could not support headless ones (without a running X session). I guess this is one of the compromise that was made. It ends up in the right "Session" dbus session. Maybe its not safe, but as I have a different user anyway, it can't touch my real Awesome instance anyway unless I use `xhost +`.

blueyed:
I still don't get if/how DISPLAY is used by what part of `awesome-client`.  dbus?

psychon:
> I still don't get if/how DISPLAY is used by what part of awesome-client. dbus?

I thought dbus would set the session bus address as a property of the root window. However, `xprop` doesn't print anything, so dunno. I didn't remove it here on purpose, only accidentally. Should I add it again?

Why do we set `$DISPLAY` in so many places in `run.sh`? Shouldn't it be enough to set it once in the beginning and be done?

On how the right dbus session is found:
`tests/run.sh` does `eval $(DISPLAY="$D" dbus-launch --sh-syntax --exit-with-session)`. This results in the right env vars being set for everything and I don't think the `$DISPLAY` is then necessary to find the dbus:
```
$ dbus-launch --sh-syntax --exit-with-session
DBUS_SESSION_BUS_ADDRESS='unix:abstract=/tmp/dbus-iHT9G7rdwX,guid=b37c9f3ad9527559ed6bac7357e74752';
export DBUS_SESSION_BUS_ADDRESS;
DBUS_SESSION_BUS_PID=10088;
```

blueyed:
> I didn't remove it here on purpose, only accidentally. Should I add it again?

No, I think it's fine.
My question/comments were more general and specific to @Elv13's snippet.

PR #1117, opened 2016-09-25
Speedup ignore enterleave

psychon:
This is yet another fix to #1107. According to that bug report, `client_ignore_enterleave_events()` is slow. We already reacted to that by calling this function less often. This pull request also makes that function less slow.

Unrelated to that, this also contains a commit that stops forcing `CMAKE_BUILD_TYPE` to be set to `Release`. This means that people can now actually chose a value for this variable and it means that `assert()` starts to work, because we are no longer compiled with `NDEBUG` defined. Well, it is partly related, since this adds some `assert()`s...

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -8,8 +8,6 @@ set(CODENAME "The Fox")
 
 project(${PROJECT_AWE_NAME} C)
 
-set(CMAKE_BUILD_TYPE RELEASE)
-
 option(WITH_DBUS "build with D-BUS" ON)
 option(GENERATE_MANPAGES "generate manpages" ON)
 option(COMPRESS_MANPAGES "compress manpages" ON)
--- a/event.c
+++ b/event.c
@@ -991,9 +991,42 @@ xerror(xcb_generic_error_t *e)
     return;
 }
 
+static bool
+should_ignore(xcb_generic_event_t *event)
+{
+    uint8_t response_type = XCB_EVENT_RESPONSE_TYPE(event);
+
+    /* Remove completed sequences */
+    uint32_t sequence = event->full_sequence;
+    while (globalconf.ignore_enter_leave_events.len > 0) {
+        uint32_t end = globalconf.ignore_enter_leave_events.tab[0].end.sequence;
+        /* Do if (end >= sequence) break;, but handle wrap-around: The above is
+         * equivalent to end-sequence > 0 (assuming unlimited precision). With
+         * int32_t, this would mean that the sign bit is cleared, which means:
+         */
+        if (end - sequence < UINT32_MAX / 2)
+            break;
+        sequence_pair_array_take(&globalconf.ignore_enter_leave_events, 0);
+    }
+
+    /* Check if this event should be ignored */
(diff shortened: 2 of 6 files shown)


psychon:
Apparently "someone" broke test-tooltip.lua...

psychon:
Previously, only enter/leave events for clients were ignored. Now, *all* enter/leave events are ignored. I guess the new behavior makes more sense, but it wasn't intended.

Anyway, I think `test-tooltip.lua` is broken, is broken. The difference is that [this](https://github.com/awesomeWM/awesome/blob/9b7e655afea8f68343365708d388dcbe532621ce/tests/test-tooltip.lua#L59-L61) previously generated a leave and an enter event and now it only generates an enter event. I don't know why this makes a difference, but apparently it does.
[No idea which magic happens now, but:]
I can fix the test with the following patch. I do not know why it fixes the test, but the following patch seems to make sense to me: The code warps the mouse cursor and afterwards makes an assertion. This looks a lot like the assert is meant to check the result of the mouse warp and changing the code accordingly fixes things...
```patch
diff --git a/tests/test-tooltip.lua b/tests/test-tooltip.lua
index ab25a56..03c9ce1 100644
--- a/tests/test-tooltip.lua
+++ b/tests/test-tooltip.lua
@@ -93,8 +93,12 @@ table.insert(steps, function()
         x = w.x + w.width - 20 - 12.5,
         y = w.y + 125 + 12.5,
     }
+    return true
+end)
 
-    assert(tt.current_position == "top")
+-- Test that the above move had the intended effect
+    assert(tt.current_position == "top", tt.current_position)
 
     return true
 end)
```
@Elv13 Sorry for the above random randomness. What do you say about all this?

madduck:
I applied your branch on top of 8dc98e8c51580f33608a93591a8aa373e7102b3a and it looks like there are no delays. I'll use it a bit longer and would update the issue if there's a regression from the status quo.

madduck:
Been using the computer quite a bit today without any problems, so it seems this patch "works".

Elv13:
> @Elv13 Sorry for the above random randomness. What do you say about all this?

As long as https://awesome.naquadah.org/bugs/index.php?do=details&task_id=1129 (equivalent) doesn't regress, fine by me.

psychon:
Thanks.
I think that this could be merged then.

blueyed:
Thanks!
It would be nice to have the test fix squashed into the failing commit or rather before it.
I think it's easier if you do it yourself.

psychon:
Which one is the failing commit? Do you mean squasing "Fix test-tooltip.lua: Run the event loop" into "Speed up client_ignore_enterleave_events()"? I thought they were unrelated changes, but if you want, I can squash them.
Should I just merge this after squashing?

Elv13:
You will need to cherry-pick the luacheck commit or the build will fail once merged. You can squash the test-tooltip into the "offending" commit so bisect keep working.

psychon:
How about changing their order? First fix the test, then optimise enter/leave-ignore. That would keep bisect happy.

Elv13:
Fine by me, up to you

PR #1119, opened 2016-09-26
Upstream shape api p4.03

Elv13:
Some more commits cherry-picked from #1044 now that their dependencies are merged

### DIFF
--- a/lib/awful/widget/taglist.lua
+++ b/lib/awful/widget/taglist.lua
@@ -177,6 +177,10 @@ function taglist.new(screen, filter, buttons, style, update_function, base_widge
     local uf = update_function or common.list_update
     local w = base_widget or fixed.horizontal()
 
+    if w.set_spacing and (style and style.spacing or beautiful.taglist_spacing) then
+        w:set_spacing(style and style.spacing or beautiful.taglist_spacing)
+    end
+
     local data = setmetatable({}, { __mode = 'k' })
 
     local queued_update = {}
--- a/lib/awful/widget/taglist.lua
+++ b/lib/awful/widget/taglist.lua
@@ -30,6 +30,143 @@ end
 local taglist = { mt = {} }
 taglist.filter = {}
 
+--- The tag list main foreground (text) color.
+-- @beautiful beautiful.taglist_fg_focus
+-- @param[opt=fg_focus] color
+-- @see gears.color
+
+--- The tag list main background color.
+-- @beautiful beautiful.taglist_bg_focus
+-- @param[opt=bg_focus] color
+-- @see gears.color
+
+--- The tag list urgent elements foreground (text) color.
+-- @beautiful beautiful.taglist_fg_urgent
+-- @param[opt=fg_urgent] color
+-- @see gears.color
+
+--- The tag list urgent elements background color.
+-- @beautiful beautiful.taglist_bg_urgent
+-- @param[opt=bg_urgent] color
+-- @see gears.color
+
+--- The tag list occupied elements background color.
+-- @beautiful beautiful.taglist_bg_occupied
+-- @param color
+-- @see gears.color
+
+--- The tag list occupied elements foreground (text) color.
+-- @beautiful beautiful.taglist_fg_occupied
+-- @param color
+-- @see gears.color
+
+--- The tag list empty elements background color.
+-- @beautiful beautiful.taglist_bg_empty
+-- @param color
+-- @see gears.color
+
+--- The tag list empty elements foreground (text) color.
+-- @beautiful beautiful.taglist_fg_empty
+-- @param color
+-- @see gears.color
+
+--- The selected elements background image.
+-- @beautiful beautiful.taglist_squares_sel
+-- @param surface
+-- @see gears.surface
+
+--- The unselected elements background image.
+-- @beautiful beautiful.taglist_squares_unsel
+-- @param surface
+-- @see gears.surface
+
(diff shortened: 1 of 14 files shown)


lib/awful/widget/taglist.lua
```diff
@@ -177,6 +177,10 @@ function taglist.new(screen, filter, buttons, style, update_function, base_widge
     local uf = update_function or common.list_update
     local w = base_widget or fixed.horizontal()
 
+    if w.set_spacing and (style and style.spacing or beautiful.taglist_spacing) then
```
psychon: Doesn't this now also need a doc comment? Something like `@beatiful taglist_spacing`?

lib/awful/widget/taglist.lua
```diff
@@ -91,15 +94,53 @@ function taglist.taglist_label(t, args)
...
+        if args.shape_urgent or theme.taglist_shape_urgent then
+            shape = args.shape_urgent or theme.taglist_shape_urgent
+        end
+
+        if args.shape_border_width_urgent or theme.taglist_shape_border_width_urgent then
+            shape_border_width = args.shape_border_width_urgent or theme.taglist_shape_border_width_urgent
+        end
+
+        if args.shape_border_color_urgent or theme.taglist_shape_border_color_urgent then
+            shape_border_color = args.shape_border_color_urgent or theme.taglist_shape_border_color_urgent
```
psychon: Don't all of these also need `@beautiful` docs?

lib/awful/widget/tasklist.lua
```diff
@@ -168,6 +214,18 @@ end
...
+-- @tparam[opt=nil] string|color style.shape_border_color
+-- @tparam[opt=nil] gears.shape style.shape_focus
+-- @tparam[opt=nil] number style.shape_border_width_focus
+-- @tparam[opt=nil] string|color style.shape_border_color_focus
+-- @tparam[opt=nil] gears.shape style.shape_minimized
+-- @tparam[opt=nil] number style.shape_border_width_minimized
+-- @tparam[opt=nil] string|color style.shape_border_color_minimized
+-- @tparam[opt=nil] gears.shape style.shape_urgent
+-- @tparam[opt=nil] number style.shape_border_width_urgent
+-- @tparam[opt=nil] string|color style.shape_border_color_urgent
```
psychon: I think something like this is also missing from the taglist commit.

lib/awful/widget/taglist.lua
```diff
@@ -30,6 +30,143 @@ end
...
+--- The shape used for the urgent elements border width.
+-- @beautiful beautiful.taglist_shape_border_width_urgent
+-- @param[opt=0] number
+-- @see wibox.container.background
+
+--- The urgents elements shape border color.
+-- @beautiful beautiful.taglist_shape_border_color_urgent
+-- @param color
+-- @see gears.color
+
```
psychon: Ok, I see. I guess you have everything covered...

psychon:
Just some minor (and likely wrong) comments above. Besides that :+1:.
From my POV, feel free to merge after concluding that my comments are irrelevant.

Elv13:
Well, I talked about that in the other PR (or IRC?). I honestly don't think it was worth fixing, and I still don't. We're talking about comments here. And those comments are in dedicated commits along without other similar comments. So "wrong commit" is debatable and consequences free. So, yeah, I see why you wrote those PR comments, but I do think, in this case, it is indeed pretty irrelevant.

Thanks for the review. #1044 is almost done merging. There is only the wibox/notification shape with the compositing aliasing PITA and the slider widget (due to the widget matrix bug I have yet to investigate).

psychon:
> There is only the wibox/notification shape with the compositing aliasing PITA and the slider widget (due to the widget matrix bug I have yet to investigate).

Does something like this help? (It took forever to find this and I'm always surprised by cairo's matrix multiplication order... and of course, none of the existing unit tests catched either of these bugs)
```diff
diff --git a/lib/wibox/drawable.lua b/lib/wibox/drawable.lua
index e7df582..e3d850f 100644
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -386,8 +386,7 @@ function drawable.new(d, widget_context_skeleton, drawable_name)
             local widgets = ret:find_widgets(x, y)
             for _, v in pairs(widgets) do
                 -- Calculate x/y inside of the widget
-                local lx = x - v.x
-                local ly = y - v.y
+                local lx, ly = v.matrix_to_device:invert():transform_point(x, y)
                 v.widget:emit_signal(name, lx, ly, button, modifiers,v)
             end
         end)
diff --git a/lib/wibox/hierarchy.lua b/lib/wibox/hierarchy.lua
index ffdc992..69f0377 100644
--- a/lib/wibox/hierarchy.lua
+++ b/lib/wibox/hierarchy.lua
@@ -126,7 +126,7 @@ function hierarchy_update(self, context, widget, width, height, region, matrix_t
             r = hierarchy_new(self._redraw_callback, self._layout_callback, self._callback_arg)
             r._parent = self
         end
-        hierarchy_update(r, context, w._widget, w._width, w._height, region, w._matrix, matrix_to_device * w._matrix)
+        hierarchy_update(r, context, w._widget, w._width, w._height, region, w._matrix, w._matrix * matrix_to_device)
         table.insert(self._children, r)
     end
 
```
With the above, clicks on a rotated widget that is not at (0,0) in the wibar/wibox/drawin started to work.

PR #1120, opened 2016-09-26
Add a grid layout and a placement container

Elv13:
More from #1044.

### DIFF
--- a/lib/wibox/container/place.lua
+++ b/lib/wibox/container/place.lua
@@ -1,0 +1,187 @@
+---------------------------------------------------------------------------
+-- A container used to place smaller widgets into larger space.
+--
+--@DOC_wibox_container_defaults_place_EXAMPLE@
+-- @author Emmanuel Lepage Vallee &lt;elv1313@gmail.com&gt;
+-- @copyright 2016 Emmanuel Lepage Vallee
+-- @release @AWESOME_VERSION@
+-- @classmod wibox.container.place
+---------------------------------------------------------------------------
+
+local setmetatable = setmetatable
+local base = require("wibox.widget.base")
+local util = require("awful.util")
+
+local place = { mt = {} }
+
+-- Take the widget width/height and compute the position from the full
+-- width/height
+local align_fct = {
+    left   = function(_  , _   ) return 0                         end,
+    center = function(wdg, orig) return math.max(0, (orig-wdg)/2) end,
+    right  = function(wdg, orig) return math.max(0, orig-wdg    ) end,
+}
+align_fct.top, align_fct.bottom = align_fct.left, align_fct.right
+
+-- Layout this layout
+function place:layout(context, width, height)
+
+    if not self._private.widget then
+        return
+    end
+
+    local w, h = base.fit_widget(self, context, self._private.widget, width, height)
+
+    if self._private.content_fill_horizontal then
+        w = width
+    end
+
+    if self._private.content_fill_vertical then
+        h = height
+    end
+
+    local valign = self._private.valign or "center"
+    local halign = self._private.halign or "center"
+
+    local x, y = align_fct[halign](w, width), align_fct[valign](h, height)
+
+    return { base.place_widget_at(self._private.widget, x, y, w, h) }
+end
+
+-- Fit this layout into the given area
+function place:fit(context, width, height)
+    if not self._private.widget then
+        return 0, 0
+    end
+
+    local w, h = base.fit_widget(self, context, self._private.widget, width, height)
+
+    return (self._private.fill_horizontal or self._private.content_fill_horizontal)
+        and width or w,
+    (self._private.fill_vertical or self._private.content_fill_vertical)
+        and height or h
+end
+
+--- The widget to be placed.
+-- @property widget
+-- @tparam widget widget The widget
+
+function place:set_widget(widget)
+    if widget then
+        base.check_widget(widget)
+    end
+    self._private.widget = widget
+    self:emit_signal("widget::layout_changed")
+end
+
+function place:get_widget()
+    return self._private.widget
+end
+
+--- Get the number of children element
+-- @treturn table The children
+function place:get_children()
+    return {self._private.widget}
+end
+
+--- Replace the layout children
+-- This layout only accept one children, all others will be ignored
+-- @tparam table children A table composed of valid widgets
+function place:set_children(children)
+    self:set_widget(children[1])
+end
+
+--- Reset this layout. The widget will be removed and the rotation reset.
+function place:reset()
+    self:set_widget(nil)
+end
+
+--- The vertical alignment.
+--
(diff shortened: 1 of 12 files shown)


lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,607 @@
...
+    for row_idx = 1, #max_row do -- Vertical
+        local x = 0
+        for col_idx = 1, #max_col do -- Horizontal
+            local v = self._private.rows[row_idx][col_idx]
+
+            if v then
+                table.insert(result, base.place_widget_at(
+                    v, x, y, max_col[col_idx], max_row[row_idx]
+                ))
+                x = x + max_col[col_idx] + spacing
```
psychon: Shouldn't this addition be outside of the `if v then`? Otherwise there is e.g. no spacing around empty cells.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,607 @@
...
+
+            if counter == index then
+                return i, j
+            end
+        end
+    end
+
+    -- This can have bad side effects, but is the best fallback
+    return math.floor(index / data._private.row_count) + 1,
+        index % data._private.column_count + 1
```
psychon: The only caller of this function is `:remove`. I have to suggestions:
- Get rid of this function as a separate function and instead inline it into `:remove`
- Handle this "index out of bounds" case via `return` and doing nothing instead of this bad code (every widget has exactly one index... unless you specify an invalid index that is too large, then some random other widget is hit)

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,607 @@
...
+        row, column = get_next_empty(self, row, column)
+
+        base.check_widget(w)
+
+        self._private.rows[row][column] = w
+        table.insert(self._private.children, w)
+    end
+
+    self:emit_signal("widget::redraw_needed")
+    self:emit_signal("widget::layout_changed")
```
psychon: Assuming `check_widget` raises an error, then these signals aren't emitted. I guess these signals either have to be emitted inside of the `for`-loop, or the `check_widget` is moved into a separate `for` loop that runs first before anything else is done.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,607 @@
...
+        self._private.row_count = self._private.row_count + new_rows_count
+
+        for ii=1, self._private.row_count do
+            self._private.rows[ii] = self._private.rows[ii] or {}
+        end
+    end
+
+    -- This ignore if there is a widget or not in the cell
+    local i = math.floor((index - 1) / self._private.column_count) + 1
+    local j = ((index - 1) % self._private.column_count) + 1
```
psychon: Hm... Why does this layout work with two different definitions of `index`? The one that is used by `:remove` and the one that is used everywhere else. That seems dangerous...

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,607 @@
...
+    self._private.row_count = count
+
+    --TODO if count is smaller, check if there is widgets in the part that need
+    -- to be deleted and reflow() after (maybe?) forcing the column count
+    self._private.rows = self._private.rows or {}
+    for i=1, self._private.row_count do
+        self._private.rows[i] = self._private.rows[i] or {}
+    end
+
+    self:emit_signal("widget::layout_changed")
```
psychon: Doesn't this also need a redraw due to the border?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,607 @@
...
+end
+
+function grid:set_column_count(count)
+    self._private.column_count = count
+    self:emit_signal("widget::layout_changed")
```
psychon: Same as for `:set_row_count()`

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,607 @@
...
+    if index > max_count then
+        local new_rows_count = math.floor(
+            (index - max_count)/self._private.column_count
+        ) + 1
+
+        self._private.row_count = self._private.row_count + new_rows_count
+
+        for ii=1, self._private.row_count do
+            self._private.rows[ii] = self._private.rows[ii] or {}
+        end
```
psychon: I'd suggest calling just `set_row_count()`. Otherwise: This has to (later?) emit `redraw_needed`.

lib/wibox/layout/grid.lua
psychon: I'm not sure if it is a good idea, but I'll suggest it anyway:
Instead of having both `self._private.children` and `self._private.rows`, how about only having one of them?
Either just `children`, but with `nil`s at cells that are empty (and then having to fix things up if the column count changes), or just having `rows` and... Hm, both ideas make it harder to have a "linear set of indicies".

Ok, how about `local empty_widget = wibox.widget.base.make_widget()` at the beginning and instead of using `nil` for empty cells, this widget is used? This means that the layout "automagically adds entries to itself" when e.g. the row count changes, but it seems like it could simplify the code a lot. The code could then store its children in a "linear table" (`children`) and `i`/`j` can easily be "transformed" into an index into this table.

Anything like this...?

lib/wibox/container/place.lua
```diff
@@ -0,0 +1,187 @@
...
+local util = require("awful.util")
+
+local place = { mt = {} }
+
+-- Take the widget width/height and compute the position from the full
+-- width/height
+local align_fct = {
+    left   = function(_  , _   ) return 0                         end,
+    center = function(wdg, orig) return math.max(0, (orig-wdg)/2) end,
+    right  = function(wdg, orig) return math.max(0, orig-wdg    ) end,
```
psychon: I'd argue that the `math.max` calls here are wrong: The child should be centered/right-aligned even when not enough space is available.

lib/wibox/container/place.lua
```diff
@@ -0,0 +1,187 @@
...
+
+function place:set_content_fill_vertical(value)
+    self._private.content_fill_vertical = value
+    self:emit_signal("widget::layout_changed")
+end
+
+function place:set_content_fill_horizontal(value)
+    self._private.content_fill_horizontal = value
+    self:emit_signal("widget::layout_changed")
+end
```
psychon: ...all of these functions are undocumented?

lib/wibox/container/place.lua
```diff
@@ -0,0 +1,187 @@
...
+    return new(_, ...)
+end
+
+--@DOC_widget_COMMON@
+
+--@DOC_object_COMMON@
+
+return setmetatable(place, place.mt)
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
```
psychon: I like this layout. It's simple.

PR #1121, opened 2016-09-27
Add a slider widget

Elv13:
Thanks to @psychon for fixing the matrix issue. I removed the --FIXME and --HACK part from the commit first proposed in #1044. It look cleaner now and there is no "WTF pseudo math until it works" code ;).

To test:

```lua
local w2 = wibox {
    visible = true,
    width = 150,
    height = 30,
    x= 200,
    y=125,
    ontop = true,
    bg = color("#0000ff")
}
w2:setup{
    {
        {
            widget = wibox.widget.slider,
        },
        direction = "south",
        widget =  wibox.container.rotate
    },
    reflection = {
        vertical = false
    },
    widget = wibox.container.mirror
}

local w = wibox {
    visible = true,
    width = 30,
    height = 150,
    x= 100,
    y=125,
    ontop = true,
    bg = color("#0000ff")
}
w:setup{
    {
        {
            widget = wibox.widget.slider,
        },
        direction = "east",
        widget =  wibox.container.rotate
    },
    reflection = {
        horizontal = false,
        vertical   = true
    },
    widget = wibox.container.mirror
}

local w3 = wibox {
    visible = true,
    width = 30,
    height = 150,
    x= 100,
    y=325,
    ontop = true,
    bg = color("#0000ff")
}
w3:setup{
    {
        {
            widget = wibox.widget.slider,
        },
        direction = "west",
        widget =  wibox.container.rotate
    },
    reflection = {
        horizontal = false,
        vertical   = true
    },
    widget = wibox.container.mirror
}

local w4 = wibox {
    visible = true,
    width = 30,
    height = 150,
    x= 400,
    y=325,
    ontop = true,
    bg = color("#0000ff")
}
w4:setup{
    {
        {
            widget = wibox.widget.slider,
        },
        direction = "east",
        widget =  wibox.container.rotate
    },
    reflection = {
        horizontal = true
    },
    widget = wibox.container.mirror
}


local w5 = wibox {
    visible = true,
    width = 150,
    height = 30,
    x= 370,
    y=125,
    ontop = true,
    bg = color("#0000ff")
}
w5:setup{
    widget = wibox.widget.slider,
}

```

### DIFF
--- a/lib/wibox/widget/slider.lua
+++ b/lib/wibox/widget/slider.lua
@@ -1,0 +1,463 @@
+---------------------------------------------------------------------------
+-- An interactive mouse based slider widget.
+--
+--@DOC_wibox_widget_defaults_slider_EXAMPLE@
+--
+-- @author Grigory Mishchenko &lt;grishkokot@gmail.com&gt;
+-- @author Emmanuel Lepage Vallee &lt;elv1313@gmail.com&gt;
+-- @copyright 2015 Grigory Mishchenko, 2016 Emmanuel Lepage Vallee
+-- @classmod wibox.widget.slider
+---------------------------------------------------------------------------
+
+local setmetatable = setmetatable
+local type = type
+local color = require("gears.color")
+local util = require("awful.util")
+local beautiful = require("beautiful")
+local base = require("wibox.widget.base")
+local shape = require("gears.shape")
+local capi = {
+    mouse        = mouse,
+    mousegrabber = mousegrabber,
+    root         = root,
+}
+
+local slider = {mt={}}
+
+--- The slider handle shape.
+--
+--@DOC_wibox_widget_slider_handle_shape_EXAMPLE@
+--
+-- @property handle_shape
+-- @tparam[opt=gears shape rectangle] gears.shape shape
+-- @see gears.shape
+
+--- The slider handle color.
+--
+--@DOC_wibox_widget_slider_handle_color_EXAMPLE@
+--
+-- @property handle_color
+-- @param color
+
+--- The slider handle margins.
+--
+--@DOC_wibox_widget_slider_handle_margins_EXAMPLE@
+--
+-- @property handle_margins
+-- @tparam[opt={}] table margins
+-- @tparam[opt=0] number margins.left
+-- @tparam[opt=0] number margins.right
+-- @tparam[opt=0] number margins.top
+-- @tparam[opt=0] number margins.bottom
+
+--- The slider handle width.
+--
+--@DOC_wibox_widget_slider_handle_width_EXAMPLE@
+--
+-- @property handle_width
+-- @param number
+
+--- The handle border_color.
+--
+--@DOC_wibox_widget_slider_handle_border_EXAMPLE@
+--
+-- @property handle_border_color
+-- @param color
+
+--- The handle border width.
+-- @property handle_border_width
+-- @param[opt=0]  number
+
+--- The bar (background) shape.
+--
+--@DOC_wibox_widget_slider_bar_shape_EXAMPLE@
+--
+-- @property bar_shape
+-- @tparam[opt=gears shape rectangle] gears.shape shape
+-- @see gears.shape
+
+--- The bar (background) height.
+--
+--@DOC_wibox_widget_slider_bar_height_EXAMPLE@
+--
+-- @property bar_height
+-- @param number
+
+--- The bar (background) color.
+--
+--@DOC_wibox_widget_slider_bar_color_EXAMPLE@
+--
+-- @property bar_color
+-- @param color
+
+--- The bar (background) margins.
+--
+--@DOC_wibox_widget_slider_bar_margins_EXAMPLE@
+--
+-- @property bar_margins
+-- @tparam[opt={}] table margins
+-- @tparam[opt=0] number margins.left
+-- @tparam[opt=0] number margins.right
+-- @tparam[opt=0] number margins.top
+-- @tparam[opt=0] number margins.bottom
+
+--- The bar (background) border width.
+-- @property bar_border_width
+-- @param[opt=0] numbergb
+
+--- The bar (background) border_color.
+--
+--@DOC_wibox_widget_slider_bar_border_EXAMPLE@
+--
+-- @property bar_border_color
+-- @param color
+
+--- The slider value.
+--
+--@DOC_wibox_widget_slider_value_EXAMPLE@
+--
+-- @property value
+-- @param[opt=0] number
+
+--- The slider minimum value.
+-- @property minimum
+-- @param[opt=0] number
+
+--- The slider maximum value.
+-- @property maximum
+-- @param[opt=100] number
+
+--- The bar (background) border width.
+-- @beautiful beautiful.slider_bar_border_width
+-- @param number
+
+--- The bar (background) border color.
+-- @beautiful beautiful.slider_bar_border_color
+-- @param color
+
+--- The handle border_color.
+-- @beautiful beautiful.slider_handle_border_color
+-- @param color
+
+--- The handle border width.
+-- @beautiful beautiful.slider_handle_border_width
+-- @param number
+
+--- The handle .
+-- @beautiful beautiful.slider_handle_width
+-- @param number
+
+-- @beautiful beautiful.slider_handle_color
+-- @param color
+
+--- The handle shape.
+-- @beautiful beautiful.slider_handle_shape
+-- @tparam[opt=gears shape rectangle] gears.shape shape
+-- @see gears.shape
+
+--- The bar (background) shape.
+-- @beautiful beautiful.slider_bar_shape
+-- @tparam[opt=gears shape rectangle] gears.shape shape
+-- @see gears.shape
+
+--- The bar (background) height.
+-- @beautiful beautiful.slider_bar_height
+-- @param number
+
+--- The bar (background) margins.
+-- @beautiful beautiful.slider_bar_margins
+-- @tparam[opt={}] table margins
+-- @tparam[opt=0] number margins.left
+-- @tparam[opt=0] number margins.right
+-- @tparam[opt=0] number margins.top
+-- @tparam[opt=0] number margins.bottom
+
+--- The slider handle margins.
+-- @beautiful beautiful.slider_handle_margins
+-- @tparam[opt={}] table margins
+-- @tparam[opt=0] number margins.left
+-- @tparam[opt=0] number margins.right
+-- @tparam[opt=0] number margins.top
+-- @tparam[opt=0] number margins.bottom
+
+--- The bar (background) color.
+-- @beautiful beautiful.slider_bar_color
+-- @param color
+
+local properties = {
+    -- Handle
+    handle_shape         = shape.rectangle,
+    handle_color         = false,
+    handle_margins       = {},
+    handle_width         = false,
+    handle_border_width  = 0,
+    handle_border_color  = false,
+
+    -- Bar
+    bar_shape            = shape.rectangle,
+    bar_height           = false,
+    bar_color            = false,
+    bar_margins          = {},
+    bar_border_width     = 0,
+    bar_border_color     = false,
+
+    -- Content
+    value                = 0,
+    minimum              = 0,
+    maximum              = 100,
+}
+
+-- Create the accessors
+for prop in pairs(properties) do
+    slider["set_"..prop] = function(self, value)
+        local changed = self._private[prop] ~= value
+        self._private[prop] = value
+
+        if changed then
+            self:emit_signal("property::"..prop)
+            self:emit_signal("widget::redraw_needed")
+        end
+    end
+
+    slider["get_"..prop] = function(self)
+        -- Ignoring the false's is on purpose
+        return self._private[prop] == nil
+            and properties[prop]
+            or self._private[prop]
+    end
+end
+
+-- Add some validation to set_value
+function slider:set_value(value)
+    value = math.min(value, self:get_maximum())
+    value = math.max(value, self:get_minimum())
+    local changed = self._private.value ~= value
+
+    self._private.value = value
+
+    if changed then
+        self:emit_signal( "property::value"       )
+        self:emit_signal( "widget::redraw_needed" )
+    end
+end
+
+local function get_extremums(self)
+    local min = self._private.minimum or properties.minimum
+    local max = self._private.maximum or properties.maximum
+    local interval = max - min
+
+    return min, max, interval
+end
+
+function slider:draw(_, cr, width, height)
+    local bar_height = self._private.bar_height
+
+    -- If there is no background, then skip this
+    local bar_color = self._private.bar_color
+        or beautiful.slider_bar_color
+
+    if bar_color then
+        cr:set_source(color(bar_color))
+    end
+
+    local margins = self._private.bar_margins
+        or beautiful.slider_bar_margins
+
+    local x_offset, right_margin, y_offset = 0, 0
+
+    if margins then
+        if type(margins) == "number" then
+            bar_height = bar_height or (height - 2*margins)
+            x_offset, y_offset = margins, margins
+            right_margin = margins
+        else
+            bar_height = bar_height or (
+                height - (margins.top or 0) - (margins.bottom or 0)
+            )
+            x_offset, y_offset = margins.left or 0, margins.top or 0
+            right_margin = margins.right or 0
+        end
+    else
+        bar_height = bar_height or beautiful.slider_bar_height or height
+        y_offset   = (height - bar_height)/2
+    end
+
+
+    cr:translate(x_offset, y_offset)
+
+    local bar_shape = self._private.bar_shape
+        or beautiful.slider_bar_shape
+        or properties.bar_shape
+
+    local bar_border_width = self._private.bar_border_width
+        or beautiful.slider_bar_border_width
+        or properties.bar_border_width
+
+    bar_shape(cr, width - x_offset - right_margin, bar_height or height)
+
+    if bar_color then
+        if bar_border_width == 0 then
+            cr:fill()
+        else
+            cr:fill_preserve()
+        end
+    end
+
+    -- Draw the bar border
+    if bar_border_width > 0 then
+        local bar_border_color = self._private.bar_border_color
+            or beautiful.slider_bar_border_color
+            or properties.bar_border_color
+
+        cr:set_line_width(bar_border_width)
+
+        if bar_border_color then
+            cr:save()
+            cr:set_source(color(bar_border_color))
+            cr:stroke()
+            cr:restore()
+        else
+            cr:stroke()
+        end
+    end
+
+    cr:translate(-x_offset, -y_offset)
+
+    -- Paint the handle
+    local handle_color = self._private.handle_color
+        or beautiful.slider_handle_color
+
+    -- It is ok if there is no color, it will be inherited
+    if handle_color then
+        cr:set_source(color(handle_color))
+    end
+
+    local handle_height, handle_width = height, self._private.handle_width
+        or beautiful.slider_handle_width
+        or height/2
+
+    local handle_shape = self._private.handle_shape
+        or beautiful.slider_handle_shape
+        or properties.handle_shape
+
+    -- Lets get the margins for the handle
+    margins = self._private.handle_margins
+        or beautiful.slider_handle_margins
+
+    x_offset, y_offset = 0, 0
+
+    if margins then
+        if type(margins) == "number" then
+            x_offset, y_offset = margins, margins
+            handle_width  = handle_width  - 2*margins
+            handle_height = handle_height - 2*margins
+        else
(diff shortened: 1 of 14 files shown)


lib/wibox/hierarchy.lua
```diff
@@ -126,7 +126,7 @@ function hierarchy_update(self, context, widget, width, height, region, matrix_t
             r = hierarchy_new(self._redraw_callback, self._layout_callback, self._callback_arg)
             r._parent = self
         end
-        hierarchy_update(r, context, w._widget, w._width, w._height, region, w._matrix, matrix_to_device * w._matrix)
+        hierarchy_update(r, context, w._widget, w._width, w._height, region, w._matrix, w._matrix * matrix_to_device)
```
psychon: Bad psychon, fixing bugs without introducing tests that make sure that the bug is not re-introduced! ;-)
Could you squash the following into this commit?
```diff
diff --git a/spec/wibox/hierarchy_spec.lua b/spec/wibox/hierarchy_spec.lua
index 92541bc..cb04451 100644
--- a/spec/wibox/hierarchy_spec.lua
+++ b/spec/wibox/hierarchy_spec.lua
@@ -124,7 +124,7 @@ describe("wibox.hierarchy", function()
         before_each(function()
             child = make_widget(nil)
             intermediate = make_widget({
-                make_child(child, 10, 20, matrix.create_translate(0, 5))
+                make_child(child, 10, 20, matrix.create_translate(0, 5):scale(2, 2))
             })
             parent = make_widget({
                 make_child(intermediate, 5, 2, matrix.create_translate(4, 0))
@@ -151,33 +151,33 @@ describe("wibox.hierarchy", function()
         end)
 
         it("get_matrix_to_parent", function()
-            assert.is.equal(hierarchy_child:get_matrix_to_parent(), matrix.create_translate(0, 5))
+            assert.is.equal(hierarchy_child:get_matrix_to_parent(), matrix.create(2, 0, 0, 2, 0, 5))
             assert.is.equal(hierarchy_intermediate:get_matrix_to_parent(), matrix.create_translate(4, 0))
             assert.is.equal(hierarchy_parent:get_matrix_to_parent(), matrix.identity)
         end)
 
         it("get_matrix_to_device", function()
-            assert.is.equal(hierarchy_child:get_matrix_to_device(), matrix.create_translate(4, 5))
+            assert.is.equal(hierarchy_child:get_matrix_to_device(), matrix.create(2, 0, 0, 2, 4, 5))
             assert.is.equal(hierarchy_intermediate:get_matrix_to_device(), matrix.create_translate(4, 0))
             assert.is.equal(hierarchy_parent:get_matrix_to_device(), matrix.identity)
         end)
 
         it("get_matrix_from_parent", function()
-            assert.is.equal(hierarchy_child:get_matrix_from_parent(), matrix.create_translate(0, -5))
+            assert.is.equal(hierarchy_child:get_matrix_from_parent(), matrix.create(0.5, 0, 0, 0.5, 0, -2.5))
             assert.is.equal(hierarchy_intermediate:get_matrix_from_parent(), matrix.create_translate(-4, 0))
             assert.is.equal(hierarchy_parent:get_matrix_from_parent(), matrix.identity)
         end)
 
         it("get_matrix_from_device", function()
-            assert.is.equal(hierarchy_child:get_matrix_from_device(), matrix.create_translate(-4, -5))
+            assert.is.equal(hierarchy_child:get_matrix_from_device(), matrix.create(0.5, 0, 0, 0.5, -2, -2.5))
             assert.is.equal(hierarchy_intermediate:get_matrix_from_device(), matrix.create_translate(-4, 0))
             assert.is.equal(hierarchy_parent:get_matrix_from_device(), matrix.identity)
         end)
 
         it("get_draw_extents", function()
             assert.is.same({ hierarchy_child:get_draw_extents() }, { 0, 0, 10, 20 })
-            assert.is.same({ hierarchy_intermediate:get_draw_extents() }, { 0, 0, 10, 25 })
-            assert.is.same({ hierarchy_parent:get_draw_extents() }, { 0, 0, 15, 25 })
+            assert.is.same({ hierarchy_intermediate:get_draw_extents() }, { 0, 0, 20, 45 })
+            assert.is.same({ hierarchy_parent:get_draw_extents() }, { 0, 0, 24, 45 })
         end)
 
         it("get_size", function()
```
(Also, technically this is fixing two bugs at once: The hierarchy mis-computing the matrix and the drawable almost completely ignoring the transformation when handling button events. I'll let that one slip for now)

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,508 @@
...
+                (margins.left or 0) - (margins.right  or 0)
+            handle_height = handle_height -
+                (margins.top  or 0) - (margins.bottom or 0)
+        end
+    end
+
+    local value = self._private.value or self._private.min or 0
+
+    -- Get the widget size back to it's non-transfored value
+    --local matrix = gmatrix.from_cairo_matrix(cr:get_matrix())
```
psychon: Some left-overs that should be removed?

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,508 @@
...
+    return width, height
+end
+
+local event = {}
+
+local rect_order = {"x", "y", "width", "height"}
+
+-- Move the handle to the correct location
+local function move_handle(self, x, y, geo)
+    -- Fix the X position when used in a rotate or mirror widget
```
psychon: Nope, not needed. Right?

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,508 @@
...
+
+    local px = rem_translate:transform_point(x, y)
+
+    -- If the transformation moved the rectangle away from {0,0}, compensate
+    if rect.x < 0 then
+        px = rect.width + px
+    end
+
+    self:set_value(get_value(self, rect.width, px))
+    self:emit_signal("widget::redraw_needed")
```
psychon: `:set_value` handles `redraw_needed`

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,508 @@
...
+
+local rect_order = {"x", "y", "width", "height"}
+
+-- Move the handle to the correct location
+local function move_handle(self, x, y, geo)
+    -- Fix the X position when used in a rotate or mirror widget
+
+    -- Remove the translation to bring {x, y} and the rect on the same origin
+    local rem_translate = geo.matrix_to_device*gmatrix.identity
+    rem_translate.x0, rem_translate.y0 = 0, 0
```
psychon: Now I know why you need a copy of the matrix....

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,508 @@
...
+    end
+
+    local px = rem_translate:transform_point(x, y)
+
+    -- If the transformation moved the rectangle away from {0,0}, compensate
+    if rect.x < 0 then
+        px = rect.width + px
+    end
+
+    self:set_value(get_value(self, rect.width, px))
```
psychon: So this function does....
- Create a matrix that translates from widget coordinates to device coordinates, but ignoring/removing the translation part of it (so just rotations and scaling is left)
- Calculates the rounded(????) extents of this widget in non-translated-device-space
- Calculates the x-coordinate of the passed-in point in non-translated device coordinates
- If this turns out to be negative, the width of the widget is added to the result

I'm not even sure which coordinate system the result is supposed to be in. Especially the "if it's negative, just add something" seems like it transforms into the black magic coordinate system.
I'd expect the following to work, but will now try it for myself and see why it apparently does not work
```lua
local function move_handle(self, x, y, geo)
    self:set_value(get_value(self, geo.width, x))
end
```

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,508 @@
...
+        abs_x = abs_x > (wx+geo.width) and wx+geo.width or (
+            abs_x < wx and wx
+        ) or abs_x
+        abs_y = abs_y > (wy+geo.height) and wy+geo.height or (
+            abs_y < wy and wy
+        ) or abs_y
+
+        -- Get the relative point
+        local rx, ry = abs_x - wx, abs_y - wy
+        move_handle(self, rx, ry, geo)
```
psychon: Ok, I might need a while to figure out the needed magic O.o

Elv13:
Your right, I did a lousy job at cleaning this. It will need another round. Some of the magic can go away.

psychon:
I got it to work (your "test like this"-snippet doesn't set any useful colors and so I didn't see anything at first) and surprisingly it also works when scaled.

Anyway, I suggest the following patch, but it is only a start. More cleanup is still needed/possible. Also, we need to fix our API for widgets (see below).
```diff
diff --git a/lib/wibox/widget/slider.lua b/lib/wibox/widget/slider.lua
index 9741c7b..8716198 100644
--- a/lib/wibox/widget/slider.lua
+++ b/lib/wibox/widget/slider.lua
@@ -414,45 +414,24 @@ local event = {}
 local rect_order = {"x", "y", "width", "height"}
 
 -- Move the handle to the correct location
-local function move_handle(self, x, y, geo)
-    -- Fix the X position when used in a rotate or mirror widget
-
-    -- Remove the translation to bring {x, y} and the rect on the same origin
-    local rem_translate = geo.matrix_to_device*gmatrix.identity
-    rem_translate.x0, rem_translate.y0 = 0, 0
-
-    local rect = {rem_translate:transform_rectangle(
-        0,
-        0,
-        geo.width,
-        geo.height
-    )}
-
-    -- Floating point precision is here both too much and not enough
-    for k, v in ipairs(rect) do
-        rect[rect_order[k]] = util.round(v)
-    end
-
-    local px = rem_translate:transform_point(x, y)
-
-    -- If the transformation moved the rectangle away from {0,0}, compensate
-    if rect.x < 0 then
-        px = rect.width + px
-    end
-
-    self:set_value(get_value(self, rect.width, px))
-    self:emit_signal("widget::redraw_needed")
+local function move_handle(self, x, y, width)
+    self:set_value(get_value(self, width, x))
 end
 
 function event.mouse_press(self, x, y, button_id, _, geo)
     if button_id ~= 1 then return end
 
-    move_handle(self, x, y, geo)
+    local matrix_from_device = geo.matrix_to_device:invert()
+
+    -- Sigh. geo.width/geo.height is in device space. We need it in our own
+    -- coordinate system
+    local _, _, width, _ = matrix_from_device:transform_rectangle(geo.x, geo.y, geo.width, geo.height)
+
+    move_handle(self, x, y, width)
 
-    -- Get the "real" position of the widget
+    -- Calculate a matrix transforming from screen coordinates into widget coordinates
     local wgeo = geo.drawable.drawable:geometry()
-    local wx = wgeo.x + geo.x
-    local wy = wgeo.y + geo.y
+    local matrix = matrix_from_device:translate(-wgeo.x, -wgeo.y)
 
     capi.mousegrabber.run(function(mouse)
         if not mouse.buttons[1] then
@@ -460,18 +439,9 @@ function event.mouse_press(self, x, y, button_id, _, geo)
             return false
         end
 
-        -- Get the closest point to the widget
-        local abs_x, abs_y = mouse.x, mouse.y
-        abs_x = abs_x > (wx+geo.width) and wx+geo.width or (
-            abs_x < wx and wx
-        ) or abs_x
-        abs_y = abs_y > (wy+geo.height) and wy+geo.height or (
-            abs_y < wy and wy
-        ) or abs_y
-
-        -- Get the relative point
-        local rx, ry = abs_x - wx, abs_y - wy
-        move_handle(self, rx, ry, geo)
+        -- Calculate the point relative to the widget
+        local x, y = matrix:transform_point(mouse.x, mouse.y)
+        move_handle(self, x, y, width)
 
         return true
     end,"fleur")
```
Tested with the following patch ontop of the default config (I added w6 to your snippet to test scaling):
```diff
diff --git a/awesomerc.lua b/awesomerc.lua
index 287a916..46def71 100755
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -570,3 +570,139 @@ end)
 client.connect_signal("focus", function(c) c.border_color = beautiful.border_focus end)
 client.connect_signal("unfocus", function(c) c.border_color = beautiful.border_normal end)
 -- }}}
+local base = wibox.widget.base
+local function scale(widget)
+	local ret = base.make_widget()
+	function ret:layout(_, width, height)
+		return { base.place_widget_via_matrix(widget, gears.matrix.create_scale(2, 2), width/2, height/2) }
+	end
+	function ret:fit(context, width, height)
+		return base.fit_widget(self, context, widget, width/2, height/2)
+	end
+	return ret
+end
+
+local color = gears.color
+beautiful.slider_handle_color = "#ff0000"
+beautiful.slider_bar_color = "#00ff00"
+beautiful.slider_border_color = "#0000ff"
+
+local w2 = wibox {
+    visible = true,
+    width = 150,
+    height = 30,
+    x= 200,
+    y=125,
+    ontop = true,
+    bg = color("#0000ff")
+}
+w2:setup{
+    {
+        {
+            widget = wibox.widget.slider,
+        },
+        direction = "south",
+        widget =  wibox.container.rotate
+    },
+    reflection = {
+        vertical = false
+    },
+    widget = wibox.container.mirror
+}
+
+local w = wibox {
+    visible = true,
+    width = 30,
+    height = 150,
+    x= 100,
+    y=125,
+    ontop = true,
+    bg = color("#0000ff")
+}
+w:setup{
+    {
+        {
+            widget = wibox.widget.slider,
+        },
+        direction = "east",
+        widget =  wibox.container.rotate
+    },
+    reflection = {
+        horizontal = false,
+        vertical   = true
+    },
+    widget = wibox.container.mirror
+}
+
+local w3 = wibox {
+    visible = true,
+    width = 30,
+    height = 150,
+    x= 100,
+    y=325,
+    ontop = true,
+    bg = color("#0000ff")
+}
+w3:setup{
+    {
+        {
+            widget = wibox.widget.slider,
+        },
+        direction = "west",
+        widget =  wibox.container.rotate
+    },
+    reflection = {
+        horizontal = false,
+        vertical   = true
+    },
+    widget = wibox.container.mirror
+}
+
+local w4 = wibox {
+    visible = true,
+    width = 30,
+    height = 150,
+    x= 400,
+    y=325,
+    ontop = true,
+    bg = color("#0000ff")
+}
+w4:setup{
+    {
+        {
+            widget = wibox.widget.slider,
+        },
+        direction = "east",
+        widget =  wibox.container.rotate
+    },
+    reflection = {
+        horizontal = true
+    },
+    widget = wibox.container.mirror
+}
+
+
+local w5 = wibox {
+    visible = true,
+    width = 150,
+    height = 30,
+    x= 370,
+    y=125,
+    ontop = true,
+    bg = color("#0000ff")
+}
+w5:setup{
+    widget = wibox.widget.slider,
+}
+
+
+local w6 = wibox {
+    visible = true,
+    width = 150,
+    height = 30,
+    x= 570,
+    y=125,
+    ontop = true,
+    bg = color("#0000ff")
+}
+w6:set_widget(scale(wibox.widget.slider()))
```
So, API for widgets on button events:
These events get some argument which you called `geo` in your example. It contains in `geo.x`, `geo.y`, `geo.width` and `geo.height` an upper approximation of the widget's geometry *in* *device* *space* (try rotating a widget by 45° and you will see what I mean with "upper approximation"). As this example shows, this is kind of useless, because widgets will likely want to know their size in their local coordinate system. However, x/y in local coordinate system is always (0,0), so also useless.
Can't we just replace this with the instance of `wibox.hierachy` that manages the widget?
For example, this provides `:get_size()` which returns the size in the widget-local coordinate system. Also, this provides all the info which one might need, because right now `wibox.drawable` is computing everything based on this. However, of course, this would be an API break. Since the current API is weird and broken (as soon as you start rotating or scaling things), I'd argue that this is needed and unavoidable.

What do you think?

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,508 @@
...
+
+    util.table.crush(ret, slider, true)
+
+    -- Change the cusor and redraw (in case there is hover and pressed colors)
+    ret:connect_signal("button::press", event.mouse_press)
+
+    return ret
+end
+
+function slider.mt:__call(...)
```
psychon: This needs `_` as first argument (first argument being the metatable on which this function is called)

Elv13:
Rebased and fixed

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,473 @@
...
+    local max = self._private.maximum or properties.maximum
+    local interval = max - min
+
+    return min, max, interval
+end
+
+local function get_value(self, width, x)
+    local _, _, interval = get_extremums(self)
+
+    return math.floor((x*interval)/width)
```
psychon: I'd suggest inlining `get_value()` into its only caller `move_handle`. Also, why the `math.floor`? This is (by default) a percentage value that you are rounding, so something between 0 and 100. Having this as an integer shouldn't make any difference unless the slider is exactly 100 pixel wide. I suggest getting rid of `math.floor`.

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,473 @@
...
+
+        cr:set_line_width(handle_border_width)
+        cr:stroke()
+    end
+end
+
+function slider:fit(_, width, height)
+    -- Use all the space, this should be used with a constraint widget
+    return width, height
+end
```
psychon: Kind-of-unrelated to this patch:
What do you think of the following patch? It would allow to get rid of all these "do nothing" `:fit()` methods.
```diff
diff --git a/lib/wibox/widget/base.lua b/lib/wibox/widget/base.lua
index 218f60f..765c153 100644
--- a/lib/wibox/widget/base.lua
+++ b/lib/wibox/widget/base.lua
@@ -279,17 +279,12 @@ function base.fit_widget(parent, context, widget, width, height)
     width = math.max(0, width)
     height = math.max(0, height)
 
-    local w, h = 0, 0
+    local w, h
     if widget.fit then
         w, h = get_cache(widget, "fit"):get(context, width, height)
     else
-        -- If it has no fit method, calculate based on the size of children
-        local children = base.layout_widget(parent, context, widget, width, height)
-        for _, info in ipairs(children or {}) do
-            local x, y, w2, h2 = matrix.transform_rectangle(info._matrix,
-                0, 0, info._width, info._height)
-            w, h = math.max(w, x + w2), math.max(h, y + h2)
-        end
+        -- If it has no fit method, use all the available space
+        w, h = width, height
     end
 
     -- Apply forced size and handle nil's
```

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,473 @@
...
+
+local event = {}
```
psychon: Why this local table? Its only entry is `mouse_press`. I suggest getting rid of it and making `mouse_press` itself local.

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,473 @@
...
+
+-- Move the handle to the correct location
+local function move_handle(self, x, _, width)
+    self:set_value(get_value(self, width, x))
+end
+
+function event.mouse_press(self, x, y, button_id, _, geo)
+    if button_id ~= 1 then return end
+
+    local matrix_from_device = geo.hierarchy:get_matrix_to_device():invert()
```
psychon: Please use `get_matrix_from_device()` instead of inverting the to-device-matrix.

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,473 @@
...
+
+    -- Sigh. geo.width/geo.height is in device space. We need it in our own
+    -- coordinate system
+    local _, _, width, _ = matrix_from_device:transform_rectangle(geo.x, geo.y, geo.width, geo.height)
```
psychon: Please replace with `local width = geo.widget_width`. This slider widget is exactly why this field was added in 6d8e91f5e21b6083afd2a55df68de1b15f7bcc69.

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,473 @@
...
+
+    move_handle(self, x, y, width)
+
+    -- Calculate a matrix transforming from screen coordinates into widget coordinates
+    local wgeo = geo.drawable.drawable:geometry()
+    local matrix = matrix_from_device:translate(-wgeo.x, -wgeo.y)
+
+    capi.mousegrabber.run(function(mouse)
+        if not mouse.buttons[1] then
+            self:emit_signal("widget::redraw_needed")
```
psychon: Why is this redraw needed? I don't think it is.

lib/wibox/widget/slider.lua
```diff
@@ -0,0 +1,473 @@
...
+            return false
+        end
+
+        -- Calculate the point relative to the widget
+        local x_prime, y_prime = matrix:transform_point(mouse.x, mouse.y)
+        move_handle(self, x_prime, y_prime, width)
```
psychon: I'm exaggerating, but: Change `move_handle()`'s argument order into `self, width, x, y` and the above can become `move_handle(self, width, matrix:transform_point(mouse.x, mouse.y))`. (Feel free to ignore this comment, I was only wondering about the variable names)

PR #1123, opened 2016-09-28
Support focus stealing filters

Elv13:
After seeing the "Fixing Firefox focus stealing" mailing list thread.

the goal of this is "simply to make it simple".

### DIFF
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -15,7 +15,26 @@ local aclient = require("awful.client")
 local aplace = require("awful.placement")
 local asuit = require("awful.layout.suit")
 
-local ewmh = {}
+local ewmh = {
+    generic_activate_filters    = {},
+    contextual_activate_filters = {},
+}
+
+--- The list of all registered generic request::activate (focus stealing)
+-- filters. If a filter is added to only one context, it will be in
+-- `ewmh.contextual_activate_filters`["context_name"].
+-- @table[opt={}] generic_activate_filters
+-- @see ewmh.activate
+-- @see ewmh.add_activate_filter
+-- @see ewmh.remove_activate_filter
+
+--- The list of all registered contextual request::activate (focus stealing)
+-- filters. If a filter is added to only one context, it will be in
+-- `ewmh.generic_activate_filters`.
+-- @table[opt={}] contextual_activate_filters
+-- @see ewmh.activate
+-- @see ewmh.add_activate_filter
+-- @see ewmh.remove_activate_filter
 
 --- Update a client's settings when its geometry changes, skipping signals
 -- resulting from calls within.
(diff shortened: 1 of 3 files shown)


lib/awful/ewmh.lua
```diff
@@ -68,6 +85,55 @@ function ewmh.activate(c, context, hints) -- luacheck: no unused args
...
+-- The callback takes the following parameters:
+--
+-- * **c** (*client*) The client requesting the activation
+-- * **context** (*string*) The activation context.
+-- * **hints** (*table*) Some additional hints (depending on the context)
+--
+-- If the callback returns `true`, the client will be activated unless the `force`
+-- hint is set. If the callback returns `false`, the activation request is
+-- cancelled. If the callback returns `nil`, the previous callback will be
+-- executed. This will continue until either a callback handle the request or
```
blueyed: s/handle/handles/

lib/awful/ewmh.lua
```diff
@@ -68,6 +85,55 @@ function ewmh.activate(c, context, hints) -- luacheck: no unused args
...
+-- when it run out of callbacks. In that case, the request will be granted if
```
blueyed: s/run/runs/

lib/awful/ewmh.lua
```diff
@@ -68,6 +85,55 @@ function ewmh.activate(c, context, hints) -- luacheck: no unused args
...
+--    end)
+--
+-- @tparam function f The callback
+-- @see activate_filters
+-- @see remove_activate_filter
+function ewmh.add_activate_filter(f)
+    table.insert(ewmh.activate_filters, f)
+end
+
+--- Remove an activate (focus steanling) filter function.
```
blueyed: s/steanling/stealing/

tests/test-awful-client.lua
```diff
@@ -77,6 +77,41 @@ end
...
+
+    -- This should still be the case
+    assert(client.focus == c1)
+
+    c2:emit_signal("request::activate", "i_said_so")
+
+    return true
+end)
+
+-- Check if writing a focus stealing filter work
```
blueyed: s/work/works./

Elv13:
Done

> I am using some advanced stuff in this regard myself, where I wrap firefox to write a hint to a file, so that it gets whitelisted/raised when executed explicitly (e.g. via xdg-open). This will help me in that regard.

At some point you should share your config... I think @mindeunix uses an SQLite database for this.

Personally, I wrap this into awesome-client "awful.spawn(cmd, {rules})". No need to serialize any hints. They are pushed directly into Awesome.

Elv13:
I added an unrelated commit to fix the tests. Maybe `luacheck` was updated? The warning is "real", so it has to be fixed anyway.

lib/awful/ewmh.lua
```diff
@@ -68,6 +85,55 @@ function ewmh.activate(c, context, hints) -- luacheck: no unused args
...
+-- @treturn boolean If the callback existed
+-- @see activate_filters
+-- @see add_activate_filter
+function ewmh.remove_activate_filter(f)
+    for k, v in ipairs(ewmh.activate_filters) do
+        if v == f then
+            table.remove(ewmh.activate_filters, k)
+
+            -- In case the callback is there multiple time.
+            ewmh.remove_activate_filter(f)
```
psychon: Is the order of these filters really relevant? I'd expect them to be side-effect-free aka "their order is relevant". If so, I'd simplify the filter management:
```lua
function add(f)
  ewmh.activate_filters[t] = true
end
function remove(f)
  local old = ewmh.activate_filters[t]
  ewmh.activate_filters[t] = nil
  return old == true
end
```

Elv13: > Is the order of these filters really relevant?

Yes, just like other rules. It goes from the specific to the generic

tests/test-awful-client.lua
```diff
@@ -77,6 +77,61 @@ end
...
+
+    return true
+end)
+
+
+table.insert(steps, function()
+    -- Now that the filter is gone, it should work again.
+
+    return client.focus == c1
+end)
```
psychon: I think that you can merge all of this into a single step. While it is not always clear which operations are instantaneous and which aren't, setting the input focus works fine (except for some corner cases where the client then forwards the focus elsewhere).
Less steps = faster tests. :-)

.luacheckrc
```diff
@@ -23,13 +23,14 @@ read_globals = {
...
     "root",
-    "screen",
     "selection",
     "tag",
     "window",
 }
+
 -- May not be read-only due to client.focus
 globals = {
+    "screen",
```
psychon: Could add to the commit message and the comment before this array the reason why `screen` can not be read-only?

psychon:
> I think @mindeunix uses an SQLite database for this.

Wtf? Are there usable SQLite-Lua bindings? And... wtf?

> Personally, I wrap this into awesome-client "awful.spawn(cmd, {rules})". No need to serialize any hints. They are pushed directly into Awesome.

Just to be sure: You use the above to add a callback and that callback then adds a "may no longer steal focus" entry? How exactly does that code look like?

> I added an unrelated commit to fix the tests. Maybe luacheck was updated? The warning is "real", so it has to be fixed anyway.

I'd be curious to know which warning that is exactly :-)

.luacheckrc
Elv13: Well, because the new version of luacheck complain if you do `screen[1].foo = "bar"` while `local s = screen[1]; s.foo = "bar"` still work. I will add it to the commit message.

Elv13:
> Wtf? Are there usable SQLite-Lua bindings? And... wtf?
https://github.com/mindeunix/clientdb

Elv13:
> > Personally, I wrap this into awesome-client "awful.spawn(cmd, {rules})". No need to serialize any 
> hints. They are pushed directly into Awesome.
> 
> Just to be sure: You use the above to add a callback and that callback then adds a "may no longer
> steal focus" entry? How exactly does that code look like?

The code above is about complete. Try it for yourself

    awesome-client  "require('awful.spawn')('urxvt', {new_tag=true})"

Tyrannical has some focus stealing code of its own based on monkey-patched hacks and luck. The 3.6 version will use request::activate to implement it using this PR and the SNID based spawn.

madduck:
I hope I am not thick, but I can't find @Elv13's branch for this…?

psychon:
@madduck GitHub tells me I can merge it via `git pull https://github.com/Elv13/awesome-1.git focus_stealing_filter`. Does that info help you?

madduck:
@psychon Indeed. I was looking at `awesome.git`, not `awesome-1.git`. Weird how this information doesn't seem to be apparent from the pull request.

madduck:
Applied this on top of 7f74c783cd174edb241fcfa2a695cdc78927982d and it seems to work nicely!

madduck:
Actually, @Elv13, there's a bug: with this applied, I can no longer use `mod4-j/k` to activate the Firefox window. Maybe this is because your filters do not check for `context == evmh`?

Elv13:
> Maybe this is because your filters do not check for context == evmh? 

Right, the filter I proposed is too aggressive. I will have to improve this pull request to properly document request::activate and the different contexts (and fix the example).

madduck:
For completeness, and since I noticed that you do make `context` available:

```
awful.ewmh.add_activate_filter(function(c, context, hints)
    if context == "ewmh" and c.class == "Firefox" then return false end
end
```

works fine.

Elv13:
Thanks @madduck 

@blueyed Do you think we should have a second parameter to add "per context" filters? This seems like a common use case?

blueyed:
@Elv13 
Providing `context` here makes sense to me, and I am also using it in my current `client.disconnect_signal("request::activate", awful.ewmh.activate)` overwrite.

psychon:
I cherry-picked 2cd587d77e6d0aa8c043ee4c79295688ee91dc56 into master amended with a change adding a comment explaining why this is needed.

Elv13:
All done and rebased

Issue #1104, opened 2016-09-18
awful.prompt.run: hooks: use the unmodified command in the history?


blueyed:
You can hook into `awful.prompt.run` to change the command to be run (e.g. to use a terminal when prefixed with ":"): )https://github.com/awesomeWM/awesome/blob/1be0964d8a6bd25f85dfce29e52c14fa6303a8c4/lib/awful/prompt.lua#L381-L393

However, the modified command is saved then to the history, and not the original one.
The current workaround is to handle the exec yourself (and not return a new command to be handled automatically):

``` lua
       {{}, "Return", function(command)
         if command:sub(1,2) == "::" then
           -- will keep history clean
           local s=awful.screen.focused()
           command = terminal .. ' -e zsh -i -c "' .. command:sub(3):gsub('"', '\\"') .. '"'
           mypromptbox[s]:spawn_and_handle_error(command, {floating=true})
           return
         end
         if command:sub(1,1) == ":" then
           -- will use the expanded version in the history
           return terminal .. ' -e ' .. command:sub(2)
         end
         return command
       end}
```

I think it would be better to save the unmodified (usually shorter) version there by default, and optionally provide a method to save the modified version.
This could be done with an extra return value (where nil would not change the command for the history).

The hooks were added in ce2d31c by @Elv13, but I wonder now why the already existing `keypressed_callback` was not enough already?!
Now we have two methods to customize this, which is confusing..  should we deprecate `keypressed_callback` (given that its functionality can be achieved using the new hooks)?

Elv13:
Is this really a release blocker? I think this is something that can wait until 3.6.*, no?

blueyed:
@Elv13 
Sure, it's not a blocker, but it is about new API, and I think it is good to get it right before a new major release.

Elv13:
> The hooks were added in ce2d31c by @Elv13, but I wonder now why the already existing keypressed_callback was not enough already?!

I still need it to handle some context aware "hooks". For example, `Alt+[0-9a-z]` to open the client in a specific tag. As the number of tags changes, I prefer not to register all possible hooks right away.

PR #1124, opened 2016-09-29
Prompt args refactor

Elv13:
Inspired by #1104

Eventually, this will need some tests, but AFAIK, it works fine, so "eventually".


psychon:
Fine by me. I'll leave this to @blueyed since I feel like he has more clue about this.

Elv13:
Rebased

@blueyed ping?

Elv13:
It can be changed later, I so I will merge this now to clear the v3.6 queue

PR #1130, opened 2016-09-30
WIP: Default config: use screen properties instead of global tables

### LINKED ISSUE #1125: Fix the default config so that screens become garbage-collectable
Currently when a screen is removed, the screen, its wibox and all the contained widgets cannot be garbage collected. The following patch hopefully describes the issue by itself.
```patch
index 287a916..b121d07 100755
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -237,6 +237,15 @@ awful.screen.connect_for_each_screen(function(s)
         },
     }
 end)
+
+screen.connect_signal("removed", function(s)
+    -- TODO: Figure out how to do this nicely
+    mypromptbox[s] = nil
+    mylayoutbox[s] = nil
+    mytaglist[s] = nil
+    mytasklist[s] = nil
+    mywibox[s] = nil
+end)
 -- }}}
 
 -- {{{ Mouse bindings
diff --git a/tests/test-screen-changes.lua b/tests/test-screen-changes.lua
index 7afb703..66f8548 100644
--- a/tests/test-screen-changes.lua
+++ b/tests/test-screen-changes.lua
@@ -8,6 +8,7 @@ local max = require("awful.layout.suit.max")
 local real_screen = screen[1]
 local fake_screen = screen.fake_add(50, 50, 500, 500)
 local test_client1, test_client2
+local wibox_for_fake_screen
 
 local steps = {
     -- Step 1: Set up some clients to experiment with and assign them as needed
@@ -53,6 +54,8 @@ local steps = {
         assert(wb.y == 110, wb.y)
         assert(wb.width == 600, wb.width)
 
+        wibox_for_fake_screen = wb
+
         return true
     end,
 
@@ -62,15 +65,8 @@ local steps = {
 
         -- Now that the screen is invalid, the wibox shouldn't refer to it any
         -- more
-        assert(mywibox[fake_screen].screen ~= fake_screen)
-
-        -- TODO: This is a hack to make the test work, how to do this so that it
-        -- also works "in the wild"?
-        mypromptbox[fake_screen] = nil
-        mylayoutbox[fake_screen] = nil
-        mytaglist[fake_screen] = nil
-        mytasklist[fake_screen] = nil
-        mywibox[fake_screen] = nil
+        assert(wibox_for_fake_screen.screen ~= fake_screen)
+        wibox_for_fake_screen = nil
 
         -- Wrap in a weak table to allow garbage collection
         fake_screen = setmetatable({ fake_screen }, { __mode = "v" })
```
The only alternative that I see to the about patch is using a weak table in the default config for these tables. I like the above patch more (it's explicit about what is going on). Still, the above patch is also ugly.

Ideas?
(This also kind of blocks speeding up the multi screen tests, because for that we need to make sure that "old" wiboxes are not repainted and our current approach to that is having them GC'd.)

psychon:
This fixes #1125, but depends on #1126 to be merged first. Only the last commit is new. All the other commits are the current state of #1126.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -124,12 +124,8 @@ mykeyboardlayout = awful.widget.keyboardlayout()
 mytextclock = wibox.widget.textclock()
 
 -- Create a wibox for each screen and add it
-mywibox = {}
-mypromptbox = {}
-mylayoutbox = {}
-mytaglist = {}
 -- @TAGLIST_BUTTON@
-mytaglist.buttons = awful.util.table.join(
+local taglist_buttons = awful.util.table.join(
                     awful.button({ }, 1, function(t) t:view_only() end),
                     awful.button({ modkey }, 1, function(t)
                                               if client.focus then
@@ -146,9 +142,8 @@ mytaglist.buttons = awful.util.table.join(
                     awful.button({ }, 5, function(t) awful.tag.viewprev(t.screen) end)
                 )
 
-mytasklist = {}
 -- @TASKLIST_BUTTON@
-mytasklist.buttons = awful.util.table.join(
+local tasklist_buttons = awful.util.table.join(
                      awful.button({ }, 1, function (c)
                                               if c == client.focus then
                                                   c.minimized = true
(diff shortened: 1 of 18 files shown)


Elv13:
You can merge #1126, and this one too 👍

PR #1133, opened 2016-10-01
Widget mouse signals

psychon:
This is fixing up the documentation of widget signals, but also making the API a bit saner for some cases that Elv13 found and that are currently quite weird.
I'm open to suggestions on how to improve these docs, especially the docs for `wibox.drawable:find_widgets()`.

This will likely conflict with #1121 by making a change that is also in there. This also effectively reverts c159ad3ab24b96d383fb0a6a11d2a395df9655d5. For these two reasons: CC @Elv13


Elv13:
I am sorry to say that, but I mostly disagree with this one. I don't mind you reverting c159ad3 and adding the hierarchy, but I don't think the v3.5 API should be broken for that. A "hard" API break has to provide so much value it is worth doing. This isn't one of them. I would propose to keep the x,y,width,height table elements (device geometry) and replace matrix_* with the hierarchy, even if it provide "duplicate" data.

The use case for the geometry used to be quite simple, it was so you could move stuff when the mouse hover/click. It is used by the tooltip and my Radical module and at least 2 known users. I get that your changes here don't "lose" this information and just pack it differently, but it makes it more complicated to use with little benefits. Sure, advanced widget need the "raw" information to transpose inputs correctly, but this wasn't the primary use case of this callback argument.

psychon:
@Elv13 I squashed [this](https://gist.github.com/psychon/accd93144cdb20e27732cb722ae7ca05) into the "Change API for wibox.drawable:find_widgets"-commit. Is it ok like this? (And `widget_width` would be the value that the slider widget wants/needs)

Elv13:
Ok, it is good now 👍 thanks for this

docs/common/widget.ldoc
```diff
@@ -48,77 +48,51 @@
...
 
 --- When a mouse button is pressed over the widget.
--- The position of the mouse press relative to the widget while geometry
--- contains the geometry of the widget relative to the wibox.
 -- @signal button::press
--- @tparam table widget The widget
--- @tparam number lx The relative horizontal position.
--- @tparam number ly The relative vertical position.
+-- @tparam number lx The horizontal position relative to the (0,0) position in
+-- the widget.
```
blueyed: Needs to / should be indented by two spaces?!

blueyed:
:+1:

docs/common/widget.ldoc
psychon: press
Seems to be ok.

Elv13: Well, there is the little problem that the find_widgets() link isn't detected. Also, are you sure about this? I think just documenting the output is "better" in the sense the user doesn't have to navigate the documentation to get the answers (on top of that, there is no link).

psychon: Hm, yeah. The link might be a bit "ungood".
My reason for not listing everything again was that it's hard to keep up to date if the result changes. I think I almost missed one position and here we had the result of `find_widgets()` documented, but the function itself didn't say much about the return value. My hope is that by having this piece documented in only one place, it won't go stale / forgotten to be updated.

Should I change this back to what it was like before?

psychon: Via this change, we get working links and LDoc errors out if it cannot produce a link:
```diff
--- `wibox.drawable:find_widgets()` for the position that the mouse hit.
+-- @{wibox.drawable:find_widgets} for the position that the mouse hit.
```
I squashed this into the corresponding commit. Should I still duplicate `find_widgets`'s docs here? Alternatively, should I introduce some kind of fake type `widget_info` that can be linked to everywhere?

blueyed: > Should I still duplicate find_widgets's docs here?

Does not sound reasonable to me.

So we should probably stick to the `@{wibox.drawable:find_widgets}` style by default in the future?!

blueyed:
@Elv13 
Please merge it in case it's OK for you (given that conflicts are expected for you).

Elv13:
Honestly, given  my experience on IRC supporting Awesome for 9 years, the more explicit the better. So yea, I would prefer the doc to just document the whole thing or I will have to answer questions about it.

As for how to avoid the duplication. The most simple way is not to do it and copy paste. The "better" way is to also use `configure_file` on the `.ldoc` file and scan them in alphabetical order. So if you create a `01-find_widget_args.ldoc`, then you can include it in `widget.ldoc`. Wont affect the performance much and will allow use to recursively meta-document things like that.

psychon:
New attempt including the suggested copy&paste (well, it's easier to document table arguments instead of table return values, so the docs here look nicer). How is this version?

Elv13:
👍

PR #1136, opened 2016-10-04
build: Fix make install when ldoc is missing

### LINKED ISSUE #1131: "make install" fails if ldoc is not available
Originally reported here: https://github.com/awesomeWM/awesome/issues/1127#issuecomment-250840247

I can reproduce this. After `sudo chmod 0 /usr/bin/ldoc`, running `make && make install DESTDIR=/tmp/awesome-install` fails with
```
CMake Error at cmake_install.cmake:76 (file):
  file INSTALL cannot find
  "/tmp/awesome/.build-ZweiLED-x86_64-linux-gnu-6.1.1/awesomerc.lua".
```

Elv13:
When ldoc is missing, nothing does depend on `awesomerc.lua` (apparently, `install()` isn't smart enough to update the dependency graph). So it wasn't generated... This PR make the dependency explicit. Sorry about missing that earlier.

Fix #1131

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,6 +106,9 @@ set(AWESOME_C_FLAGS
 mark_as_advanced(AWESOME_C_FLAGS)
 target_compile_options(${PROJECT_AWE_NAME} PRIVATE ${AWESOME_C_FLAGS})
 
+# Make sure awesomerc.lua is generated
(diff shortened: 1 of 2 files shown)


psychon:
Oh and thanks for figuring our what the problem is. I gave this issue a quick look back then and had no idea.

Elv13:
> This feels a bit hackish to me, but so is CMake...

It is the officially documented solution to this...

PR #1138, opened 2016-10-04
Fix short flicker when new clients appear

psychon:
We are doing more and more things lazily in the C code. The newest
addition is lazily configuring clients, which means that geometry
changes are only applied later.

However, this caused a short flicker when a new client appears: We were
first making the client visible and then moving it to its "proper"
position.

Since unbanning is also done lazily, we just have to change the order in
which we apply these operations to fix this.

### DIFF
--- a/event.h
+++ b/event.h
@@ -46,10 +46,10 @@ awesome_refresh(void)
 {
     screen_refresh();
     luaA_emit_refresh();
-    banning_refresh();
-    stack_refresh();
     drawin_refresh();
     client_refresh();
+    banning_refresh();
+    stack_refresh();
     return xcb_flush(globalconf.connection);
 }
 


blueyed:
Just tried this and it's better!

Can this be improved to also fix the "flicker" when closing a client: I see the wallpaper for a short time (the "cornerse" layout seems to take longer than the "tile" layout in this regard).

psychon:
Uhm. I guess that that flicker occurs because on unmanage we immediately destroy the frame window for the client while updating the other window's position is delayed until the end of the current main loop iteration.
Fixing that would mean we also need to delay destroying the frame window. Oh what ugly, ugly fun...
Something like the following? (ignore the compiler warnings)
```diff
diff --git a/event.h b/event.h
index 47ea428..6afbc8f 100644
--- a/event.h
+++ b/event.h
@@ -50,6 +50,12 @@ awesome_refresh(void)
     stack_refresh();
     drawin_refresh();
     client_refresh();
+    /* Ignore all spurious enter/leave notify events */
+    client_ignore_enterleave_events();
+    foreach(w, globalconf.delayed_destroy_windows)
+        xcb_destroy_window(globalconf.connection, *w);
+    client_restore_enterleave_events();
+    globalconf.delayed_destroy_windows.len = 0;
     return xcb_flush(globalconf.connection);
 }
 
diff --git a/globalconf.h b/globalconf.h
index 0d4057f..3237d6b 100644
--- a/globalconf.h
+++ b/globalconf.h
@@ -68,6 +68,7 @@ ARRAY_TYPE(client_t *, client)
 ARRAY_TYPE(drawin_t *, drawin)
 ARRAY_TYPE(xproperty_t, xproperty)
 DO_ARRAY(sequence_pair_t, sequence_pair, DO_NOTHING)
+DO_ARRAY(xcb_window_t, window, DO_NOTHING)
 
 /** Main configuration structure */
 typedef struct
@@ -189,6 +190,7 @@ typedef struct
     /** List of enter/leave events to ignore */
     sequence_pair_array_t ignore_enter_leave_events;
     xcb_void_cookie_t pending_enter_leave_begin;
+    window_array_t delayed_destroy_windows;
 } awesome_t;
 
 extern awesome_t globalconf;
diff --git a/objects/client.c b/objects/client.c
index 0b81bdc..b8cbcb9 100644
--- a/objects/client.c
+++ b/objects/client.c
@@ -2088,12 +2088,9 @@ client_unmanage(client_t *c, bool window_valid)
                 c->geometry.x, c->geometry.y);
     }
 
-    /* Ignore all spurious enter/leave notify events */
-    client_ignore_enterleave_events();
     if (c->nofocus_window != XCB_NONE)
-        xcb_destroy_window(globalconf.connection, c->nofocus_window);
-    xcb_destroy_window(globalconf.connection, c->frame_window);
-    client_restore_enterleave_events();
+        window_array_append(&globalconf.delayed_destroy_windows, c->nofocus_window);
+    window_array_append(&globalconf.delayed_destroy_windows, c->frame_window);
 
     if(window_valid)
     {
```

blueyed:
Very nice, thank you!
Do you want to add it to this PR, or a new one?
(The patch does not apply for `event.h` on this PR, but I've added it before the `xcb_flush`.)

psychon:
Well, if it actually works (does it?), I have to decide that this is not too ugly and should really be done and then it needs a nicer implementation. Nothing that should end up in this PR.

blueyed:
Yes, it actually works!

Merging this PR then, please don't forget the other patch.. :)

PR #1139, opened 2016-10-04
Drawable visibility

psychon:
This adds explicit tracking of visible to `wibox.drawable` and uses that to fix a use-after-free related crash that occasionally occurred on Travis (and I think @Elv13 ran into this while developing some PR).

The downside of this is that it might add unnecessary redraws when a wibox is made visible after being hidden for a while. This might be fixable via various kinds of magic, but I'd live with it for now. (Example for such magic: Have a global variable which is initially 0 and incremented whenever the wallpaper changes; have drawables track the value of this counter when they were redrawn last; that way we can detect if the wallpaper changed while a drawable was invisible) (Another example: Have a single boolean tracking if the dirty area is empty or not to avoid redrawns when nothing changed at all)

### DIFF
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -266,6 +266,9 @@ function drawable:set_fg(c)
     self._do_complete_repaint()
 end
 
+function drawable:_inform_visible(visible)
+end
+
 local function emit_difference(name, list, skip)
     local function in_table(table, val)
         for _, v in pairs(table) do
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -266,6 +266,7 @@ function drawable:set_fg(c)
 end
 
 function drawable:_inform_visible(visible)
+    self._visible = visible
     if visible then
         visible_drawables[self] = true
         -- The wallpaper or widgets might have changed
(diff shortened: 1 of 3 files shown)


lib/wibox/drawable.lua
```diff
@@ -407,13 +408,8 @@ function drawable.new(d, widget_context_skeleton, drawable_name)
...
     ret._redraw_callback = function(hierar, arg)
-        -- XXX: lgi will lead us into memory-corruption-land when we use an
-        -- object after it was GC'd. Try to detect this situation by checking if
-        -- the drawable is still valid. This is only a weak indication, but it
-        -- seems to be the best that we can do. The problem is that the drawable
-        -- could not yet be GC'd, but is pending finalisation, while the
-        -- cairo.Region below was already GC'd. This would still lead to corruption.
-        if not ret.drawable.valid then
+        -- Avoid crashes when a drawable was partly finalized and dirty_area is broken.
+        if not ret._visible then
```
psychon: Should something similar also be done in `_layout_callback` to avoid relayouts of partly finalised widgets?
Something like the following:
```diff
@@ -426,7 +434,9 @@ function drawable.new(d, widget_context_skeleton, drawable_name)
             return
         end
         ret._need_relayout = true
-        ret:draw()
+        if ret._visible then
+            ret:draw()
+        end
     end
 
     -- Add __tostring method to metatable.
```
(If the drawable becomes visible again, the relayout will be done. If it never becomes visible again, the relayout will just keep being pending until the drawable is garbage collected.)

psychon:
I added one more commit to this PR (as per my comment above).

Elv13:
Does this totally solve #1070?

psychon:
Yes, I'd expect so.

We ignore all callbacks from `wibox.hierarchy` for not-visible `wibox.drawable`s. Only such drawables can be partly garbage collected and cause issues.

If some code tried badly, it could still wait for an invisible drawable to be partly destroyed and then make it visible again, but I will ignore that possibility.

PR #1143, opened 2016-10-04
Doc fixes for wibox.widget.base

blueyed:


lib/wibox/widget/base.lua
```diff
@@ -19,42 +19,41 @@ local base = {}
...
         self._private.widget_buttons = _buttons
     end
-
     return self._private.widget_buttons
 end
 
---- Set a widget's visible property
--- @tparam boolean b Wether the widget is visible at all
+--- Set a widget's visibility.
+-- @tparam boolean b Wether the widget is visible.
```
psychon: Unrelated to this PR, but: Wether -> Whether

lib/wibox/widget/base.lua
```diff
@@ -174,17 +182,16 @@ function base.widget:emit_signal_recursive(signal_name, ...)
...
 end
 
---- Get a widex index.
--- @param widget The widget to look for
--- @param[opt] recursive Also check sub-widgets
--- @param[opt] ... Aditional widgets to add at the end of the "path"
--- @return The index
--- @return The parent layout
--- @return The path between "self" and "widget"
+--- Get a widget' index.
```
psychon: I don't understand the apostrophe. Should this be `widget's`?

.travis.yml
```diff
@@ -1,6 +1,13 @@
 sudo: required
 dist: trusty
 language: c
+
+# Build only master and stable branches.  Other branches got through PRs.
```
psychon: got -> go ?

The `3.5` value won't really work since the 3.5 branch does not contain this commit and thus does not forbid itself to be built. I guess it is still a good idea to have this, I just wanted to mention it.

lib/wibox/widget/base.lua
blueyed: Yes.  Now: "Get the index of a widget"

blueyed:
Rebased, and addressed comments.
Will merge it when it's green.

PR #1147, opened 2016-10-05
Remove some weak tables

psychon:
Here are some commits removing uses of weak tables. For example, after this `t.foo = t` for a tag object `t` should no longer cause a leak (however, I did not actually test/verify this).

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -46,7 +46,6 @@ local client = {object={}}
 -- Private data
 client.data = {}
 client.data.marked = {}
-client.data.properties = setmetatable({}, { __mode = 'k' })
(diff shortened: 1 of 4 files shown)


lib/awful/client.lua
```diff
@@ -986,17 +984,17 @@ end
...
             local value = c:get_xproperty("awful.client.property." .. p)
             if value ~= nil then
                 client.property.set(c, p, value)
             end
         end
     end
-    if client.data.properties[c] then
-        return client.data.properties[c][prop]
+    if c.data.awful_client_properties then
+        return c.data.awful_client_properties[prop]
```
Elv13: You might as well just return `c[prop]` and print a deprecated message now that this function is "really" deprecated. Until now, it was still used in the background by the miss handler.

PR #1148, opened 2016-10-06
Delay client frame window destruction

psychon:
Daniel sees a short flicker of his wallpaper when he closes a client.
This happens because the window is destroyed immediately, but other
clients are re-arranged only shortly later. In the mean time, the X
server updates the display and repaints the root window (= wallpaper
becomes visible).

Work around this by delaying the destruction of frame windows to the end
of the current main loop iteration. This means that we first update the
position of all other windows and later destroy the window that was
actually closed.

See #1138.

### DIFF
--- a/event.h
+++ b/event.h
@@ -37,6 +37,7 @@ void drawin_refresh(void);
 /* objects/client.c */
 void client_refresh(void);
 void client_focus_refresh(void);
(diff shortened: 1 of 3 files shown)


blueyed:
Thanks!

PR #1150, opened 2016-10-06
awesome-client: allow override of rlwrap

blueyed:
This is necessary when being run with a tty that has no width, which
happens when using https://github.com/metakirby5/codi.vim/ in Vim 8 (but
not Neovim), where rlwrap complains as follows:

> rlwrap: error: My terminal reports width=0 (is it emacs?) I can't
> handle this, sorry!

Then `awesome-client` can be used like this in codi.vim:

```vim
    function! s:pp_awesome_remove_spaces(line)
      return substitute(a:line, '^   ', '', 'g')
    endfunction
    let g:codi#interpreters = {
          \ 'awesome': {
          \   'bin': ['env', 'AWESOME_RLWRAP=', 'awesome-client'],
          \   'prompt': '^awesome# ',
          \   'preprocess': function('s:pp_awesome_remove_spaces')
          \ }}
```
(Requires this fix: https://github.com/metakirby5/codi.vim/pull/48)

### DIFF
--- a/utils/awesome-client
+++ b/utils/awesome-client
@@ -12,12 +12,15 @@ if [ "$ISATTY" = 0 ]
 then
     # rlwrap provides readline functionality for "read", which is more enhanced
     # than bash's "read" itself.
-    RLWRAP=$(which rlwrap 2>/dev/null)
-    if [ "$RLWRAP" != "" ]
+    # It can be disabled/overridden using 'AWESOME_RLWRAP= awesome-client'.
+    if [ -z "${AWESOME_RLWRAP+x}" ]; then
+        AWESOME_RLWRAP="$(which rlwrap 2>/dev/null)"
+    fi
+    if [ -n "$AWESOME_RLWRAP" ]
     then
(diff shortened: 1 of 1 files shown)


utils/awesome-client
```diff
@@ -12,12 +12,15 @@ if [ "$ISATTY" = 0 ]
 then
     # rlwrap provides readline functionality for "read", which is more enhanced
     # than bash's "read" itself.
-    RLWRAP=$(which rlwrap 2>/dev/null)
-    if [ "$RLWRAP" != "" ]
+    # It can be disabled/overridden using 'AWESOME_RLWRAP= awesome-client'.
+    if [ -z "${AWESOME_RLWRAP+x}" ]; then
```
psychon: I had to actually look this up to understand it, but yeah, seems to work....

blueyed: There is also `${FOO-value-if-unset}`, but that caused an issue with Vim's highlighting.  Also using an `if` here is less magical.

PR #1151, opened 2016-10-06
awesome-client: simplify/fix ISATTY usage

blueyed:
Followup to 4ce1608.

### DIFF
--- a/utils/awesome-client
+++ b/utils/awesome-client
@@ -5,10 +5,7 @@
 # when used with "|tail".
 set -o pipefail
 
-tty 2>&1 > /dev/null
-ISATTY=$?
-
-if [ "$ISATTY" = 0 ]
+if [ -t 0 ]  # is a tty.
 then
     # rlwrap provides readline functionality for "read", which is more enhanced
     # than bash's "read" itself.
@@ -42,6 +39,7 @@ DBUS_PATH=/
 DBUS_DEST=org.naquadah.awesome.awful
 DBUS_METHOD=${DBUS_DEST}.Remote.Eval
 
+FATAL_ERRORS=1
 a_dbus_send()
 {
     $DBUS_SEND --dest=$DBUS_DEST --type=method_call --print-reply \
@@ -53,15 +51,14 @@ a_dbus_send()
     fi
 }
 
-FATAL_ERRORS="$ISATTY"
 if [ $# -ne 0 ]
 then
-    FATAL_ERRORS="1"
     for arg in "$@" ; do
         a_dbus_send "$arg"
     done
-elif [ "$ISATTY" = 0 ]
+elif [ -t 0 ]
 then
+    FATAL_ERRORS=0
     while $READ_CMD -p "awesome# " -r line
     do
         if [ "$line" = "" ]; then


utils/awesome-client
```diff
@@ -39,6 +36,7 @@ DBUS_PATH=/
 DBUS_DEST=org.naquadah.awesome.awful
 DBUS_METHOD=${DBUS_DEST}.Remote.Eval
 
+FATAL_ERRORS=1
```
psychon: Should this be moved down to....

utils/awesome-client
```diff
@@ -50,15 +48,14 @@ a_dbus_send()
     fi
 }
 
-FATAL_ERRORS="$ISATTY"
```
psychon: ...here?

utils/awesome-client
```diff
@@ -5,10 +5,7 @@
 # when used with "|tail".
 set -o pipefail
 
-tty 2>&1 > /dev/null
-ISATTY=$?
-
-if [ "$ISATTY" = 0 ]
+if [ -t 1 ]  # is a tty.
```
psychon: I keep opening man pages today. So much magic. :-)

blueyed:
Something is wrong according to Travis, where all build jobs fail..

utils/awesome-client
psychon: Before: Is NOT a tty
After: Is a tty

I think this line should be `if [ ! -t 1 ]`

psychon:
...and there are load of Lua parse errors ("Expected end" etc)

So either this is a quoting issue, but you didn't touch that, or there is some negation missing and instead of sending all lines from stdin in one DBus call (`"$(cat)"`), it is now sending each line of input on its own (`while $READ_CMD -p "awesome# " -r line`). The latter would also explain why there are multiple, different errors from Lua.

utils/awesome-client
```diff
@@ -53,15 +51,14 @@ a_dbus_send()
...
 
-FATAL_ERRORS="$ISATTY"
 if [ $# -ne 0 ]
 then
-    FATAL_ERRORS="1"
     for arg in "$@" ; do
         a_dbus_send "$arg"
     done
-elif [ "$ISATTY" = 0 ]
+elif [ -t 1 ]
```
psychon: Same logic error here: Missing negation

blueyed: Are you sure?
It was $? before.. so 0 meant that it is a tty.

psychon: Hm. Good point. But Travis is definitely mixing up the cases here and using the wrong one.

psychon:
I have a theory: `tty` looks at stdin, `[ -t 1 ]` looks at fd 1 which is stdout.
```
$ dash -c 'tty ; echo $? ; [ -t 1 ] && echo t || echo not t'       
/dev/pts/0
0
t
$ echo | dash -c 'tty ; echo $? ; [ -t 1 ] && echo t || echo not t' 
kein Terminal
1
t
$ dash -c 'tty ; echo $? ; [ -t 1 ] && echo t || echo not t' | cat
/dev/pts/0
0
not t
$ echo | dash -c 'tty ; echo $? ; [ -t 1 ] && echo t || echo not t' | cat 
kein Terminal
1
not t
$ 
```

blueyed:
Trying that..

it's strange though since the outcome appears to be the same (keeping in mind that `$ISATTY` is 0 for the true case)?!

```
+tty
/dev/pts/0
+ISATTY=0
+'[' -t 1 ']'
+echo '-t 1'
-t 1
```

blueyed:
Seems to work..
My simple test case apparently does not cover the real context awesome-client gets used in.

psychon:
You were testing if stdout (FD 1) is connected to a terminal. It pretty much always is.
`tty` looks at stdin (according to its man page). So if you do `echo | SOMETHING`, then stdin is not connected to a tty, but stdout still is.

PR #1152, opened 2016-10-07
Fix #1108: crash with custom keyboard layout

### LINKED ISSUE #1108: Crash when using a custom keyboard layout

```
awesome  (The Fox)
 • D-Bus support: ✔
```

I am using a custom keyboard layout (a [variant of Dvorak with accents](https://github.com/hishamhm/dvorak_alt2); for the record, in my system `awesome.xkb_get_group_names()` returns `pc+dvorak_alt2+inet(evdev)`).

**Actual result:**

Whenever I restart Awesome with an error in my ~/.config/awesome/rc.lua, which should have caused the default config to load, instead I get a fatal error: X terminates and I'm back to the console. Piping `awesome` to a log file, I found that I got the following Lua error:

`Failed to get list of keyboard groups`

Triggered at [awful/widget/keyboardlayout.lua](https://github.com/awesomeWM/awesome/blob/c86c873779660cea655c43e4b2f879fed3ea55c2/lib/awful/widget/keyboardlayout.lua#L237):

```lua
local function update_layout(self)
    self._layout = {};
    local layouts = keyboardlayout.get_groups_from_group_names(awesome.xkb_get_group_names())
    if layouts == nil or layouts[1] == nil then
        error("Failed to get list of keyboard groups")
        return;
    end
    if #layouts == 1 then
        layouts[1].group_idx = 0
    end
    for _, v in ipairs(layouts) do
        local layout_name = self.layout_name(v)
        -- Please note that numbers of groups reported by xkb_get_group_names
        -- is greater by one than the real group number.
        self._layout[v.group_idx - 1] = layout_name
    end
    update_status(self)
end
```

Interestingly, loading Awesome with the default config works, but restarting always shuts down X.

**Expected result:**

I simply commented out the `error("Failed to get list of keyboard groups")` line and everything works as expected: Awesome doesn't terminate and my keyboard works the way I configured it; it seems to me that not finding a standard keyboard layout should not be a fatal error.

I can provide more info if desired. Thank you!

blueyed:
Ref: https://github.com/awesomeWM/awesome/issues/1108#issuecomment-248904455.

This should cause a build failure?!

### DIFF
--- a/spec/awful/keyboardlayout_spec.lua
+++ b/spec/awful/keyboardlayout_spec.lua
@@ -14,6 +14,8 @@ describe("awful.widget.keyboardlayout get_groups_from_group_names", function()
         -- possible worst cases
         [""] = {
         },
+        ["pc+dvorak_alt2+inet(evdev)"] = {
+        },
         ["empty"] = {
         },
         ["empty(basic)"] = {


blueyed:
Closing in favor of https://github.com/awesomeWM/awesome/issues/1154.

Issue #1149, opened 2016-10-06
Minor widgets improvments


Elv13:
Just to keep in mind some details discussed on IRC, I will do that eventually

* background container with forced width/height should be drawn even if there is no child widget
* margin containers should allow functions(w,h,context) for left/right/top/bottom so it is easier to auto-scale them with proportion based margins.
* The parallelogram shape default base_width is terrible
* the parallelogram, rect_tag and powerline shape should have easy way to revert their directions. Having to use matrix for that is too complex for normal users without linear algebra knowledge
* Find a way to build a powerline that always work and document it
* Look into the feasibility of extending the fixed, ratio, grid and flex layouts to have separator widgets. This would address the "outline" powerline version

Elv13:
Here's a cheap separator based powerline layout https://gist.github.com/Elv13/e04e2b0f1e0292eaf7702ff2e6197ddf

Obviously, it has all the issues and bugs I known it would have, but it is good enough for static layouts. I added an example about how to use it for the taglist. It kind of work

![slider10](https://cloud.githubusercontent.com/assets/340384/19180588/fb62f8be-8c34-11e6-99f2-1ff26379c62d.png)

For this to work properly, it will need a full blown layout implementation. Meh, another widget... But umm, it *is* a popular request and according to reddit /r/unixporn, that widget style has a dedicated fanbase. My config implementation derserve to be burned down to ashes for abuse of dark magic. So a new implementation has to be written.

* separator based impl have painting issues with sub-pixel widgets
* separator based impl can't be used for the taglist because you can't control the separator color
* shape API based impl work fine, but is limited to 2 colors and one direction

psychon:
To highlight the issues that you know about, modify like this:
```lua
s.mywibox:setup{
	{
		widget = wibox.widget.base.make_widget(),
		forced_width= 0.5
	},
       [the other widgets here]
}
```
Result is: ![not_good](https://cloud.githubusercontent.com/assets/89482/19181349/cf843d92-8c6c-11e6-9585-953f44324ee8.png)
Oh, the joy of anti-aliasing...

Elv13:
@psychon What is the best way to get a child widget area from `before_draw_child`? It seems to only provide the index

psychon:
[Here](https://gist.github.com/psychon/c51187b0909dd8d05b03d9ae1d00e622) is my hack for a powerline layout (based on yours). It just calls `self:layout()` to get information about the children and then access implementation details. Not good...
[...]

Issue #1073, opened 2016-09-08
One of the Travis pass should use an empty theme


Elv13:
The title say it all. Should a "complete" theme be an absolute requirement? Currently, Awesome has fallback for "most" values if they are not found in the theme, but not all. Should there be a documented "required" set of variables or should Awesome had a fallback for everything?

(this issue doesn't have a high priority)

PR #1072, opened 2016-09-07
Menubar tests

Crazylemon64:
This is currently a complete skeleton of a test suite for the menubar. I'm having trouble actually inspecting the contents of the menubar's state, but then again maybe I'm looking at this absolutely boneheadedly.

PR #1153, opened 2016-10-09
[WIP] Next batch of features

Elv13:
This is a long term PR. It doesn't block 3.6 and will replace #1044 as the "random stuff" collection of proof of concepts and work in progress features and requests.

So far it mostly add the `shaped` and `free` layouts (both already discussed elsewhere and in gists). It also fix a couple of drawin related bugs I found when writing some integration tests for the slider (unpublished).

On the doc/test side, this add a new test utility to create SVG from widgets, wibox and screens during the integration. This is for #1149, #1072, #1073, notifications and @actionless shortcut popup. We currently don't have tests, coverage or images from the menubar/non-default-theme/shortcut popup. This is the most overkill way to fix this  (_, but it is so cool_). Once those get covered and documented, Awesome will be at ~77-79% coverage (estimate) and all\* major parts will have good coverage.

Again, this is early work in progress. Reviews are welcome, but some important parts are missing (like documentation) and corners were cut short to get the basic working. It may be better to wait until I remove the [WIP] before spending too much time inspecting the code

See #1149.

\* but the menu and prompt. Other modules will all be in the ~70% range, so the initial back catalog testing endeavor will be completed!

### DIFF
--- a/lib/awful/mouse/init.lua
+++ b/lib/awful/mouse/init.lua
@@ -124,7 +124,7 @@ end
 -- @function awful.mouse.wibox.move
 --@tparam wibox w The wibox to move, or none to use that under the pointer
 function mouse.wibox.move(w)
-    w = w or mouse.wibox_under_pointer()
+    w = w or mouse.current_wibox
     if not w then return end
 
     if not w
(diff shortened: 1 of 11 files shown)


Elv13:
followed up in #1990 and beyond renders the remaining commits of this PR obsolete.

PR #1155, opened 2016-10-09
gears.surface.widget_to_*: Ignore repaints

psychon:
For some reason, the code here tried to handle widget::redraw_needed
signals even though it should apparently/obviously only produce a
current snapshot of the widget's look.

Fix this by just removing the redraw code.

While here, also factor out the widget context table into a local
variable and re-use it for the initial layout and for the later draw.


blueyed:
The hardcoded dpi=96 seems strange, but has been there before already.

PR #1156, opened 2016-10-09
Shape rotations

psychon:
This adds helper functions to `gears.shape` that simplify the rotation of shapes.

### DIFF
--- a/lib/gears/shape.lua
+++ b/lib/gears/shape.lua
@@ -652,6 +652,54 @@ function module.transform(shape)
     return result
 end
 
+--- Rotate the shape right by 90 degree
+--
+-- @DOC_gears_shape_rotate_right_EXAMPLE@
+--
+-- @param shape A shape function
+-- @return A rotated version of the shape; this is also a shape function
+function module.rotate_right(shape)
+    return function(cr, width, height)
+        cr:save()
+        local matrix = g_matrix.create_rotate_at(width/2, height/2, math.pi/2)
+        cr:transform(matrix:to_cairo_matrix())
+        shape(cr, height, width)
+        cr:restore()
+    end
+end
+
+--- Rotate the shape left by 90 degree
+--
+-- @DOC_gears_shape_rotate_left_EXAMPLE@
+--
+-- @param shape A shape function
+-- @return A rotated version of the shape; this is also a shape function
+function module.rotate_left(shape)
+    return function(cr, width, height)
+        cr:save()
+        local matrix = g_matrix.create_rotate_at(width/2, height/2, -math.pi/2)
+        cr:transform(matrix:to_cairo_matrix())
+        shape(cr, height, width)
+        cr:restore()
+    end
+end
+
+--- Rotate the shape by 180 degree
+--
+-- @DOC_gears_shape_rotate_180_EXAMPLE@
+--
+-- @param shape A shape function
+-- @return A rotated version of the shape; this is also a shape function
+function module.rotate_180(shape)
+    return function(cr, width, height)
+        cr:save()
+        local matrix = g_matrix.create_rotate_at(width/2, height/2, math.pi)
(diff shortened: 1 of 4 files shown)


psychon:
Question: What would be good names for these functions? Is left/right/180° good? Is something like "clockwise" better?

Elv13:
Thanks for taking care of this. It was overdue. The reason while I didn't made this earlier (if I recall, there was some discussions in one of the shape API PR) is that I at some point wanted to include this in the `transform` method so you could do `shape.transform(shape.powerline) : flip_left() : shrink_by(4,4) : translate_by(2,2)`. This is a cool API, but it does create a very, very expensive function call...

(I am well aware you know all that) It is important to point out that this is classical CS theory. Just like RegEx and some other algorithm, affine transformation composition is a really cool math application. It is "compiled" once and then have an `O(1)` execution time, even if there is 9999 transformations. However with this API, each transformation does it own matrix thing (cairo does some multiplication of its own in the background, and we do some on our side). Each transformation become an iteration and is executed everytime the shape is applied/drawn (because it depend on the width/height/current_matrix inputs values). So if we give a nice API like this (and it is really, really useful), don't we kind of shoot the users in the foot if their wiboxes get repainted too often? Game engine use OpenCL/CUDA for this kind of operations for good reasons. Of course, we don't do as many of those as they do (and we do so in 2D), but still. In theory, it could be possible to "compile" this into an `O(1)` matrix and cache it into the hierarchy (just like `fit`). I investigated this a little, but it is non-trivial to implement.

What's your opinion of this? I am in the micro-optimization territory or is it something we should care about? I totally agree this is very useful. I shared similar code a few time on IRC and some of my module have equivalent functions (because I need centroid based rotation, origin rotation are mostly useless).

psychon:
> (if I recall, there was some discussions in one of the shape API PR) 

Yup, that plus some question about it on IRC that I recently saw is why I wrote this (but I didn't try to find that discussion again).

> I am in the micro-optimization territory

Yup, you are.

--------

But if you want, let's think about a more useful API: 
Needed features are likely: Margins (your `shrink_by(4,4):translate_by(2,2)`-example is just a margin of 2 on each side) and rotations by multiples of 90 degrees.

Arbitrary translation/scaling/rotation will likely move the shape outside of the area assigned to it and this will likely not work (e.g. due to clipping). So, this is not needed. Also, the API is intended as a convenience API, not a "make everything possible" API (people who want that can define their own wrapper-shape-functions).

So I guess we should not merge this PR and instead I propose an API break:
Instead of `:transform()` enabling all of the matrix-API, it just provides few functions:
- rotate_left
- rotate_right
- rotate_180 (I still don't like that name, but "flip" doesn't really fit either)
- mirror_horizontally (or is flip_left / flip_right better?)
- mirror_vertically
- set_margin_{left,right,top,bottom}
- set_margin (setting all margins at once)

This provides three features: Rotation, mirroring and margins. Rotation and mirroring are independent (the order in which they are applied does not matter) and for margins we just say that they should be applied first (before rotation/mirroring).
Also, this can be implemented in O(1) time and space (although the constant is slightly higher in both cases than it previously was, but whatever).

What do you think? Is the current "make everything possible `:transform`" needed or can we remove that and switch to the above?

psychon:
@Elv13 (Untested) but how about something like https://gist.github.com/psychon/09cb6511766b28ef0bb272bde41397af ? (Feel free to pick this up, if you want)

psychon:
I vote for merging this. If we latter add a "more complete" convenience-API, this can be changed accordingly or deprecated. For now, this provides the most-often-needed feature in a simpler form.
(And I'll pass on the task for writing that "more complete" convenience-API to someone else)

Elv13:
The PR or the gist? As I will take a closer look tomorrow

psychon:
This PR.

Issue #1159, opened 2016-10-09
awful.spawn calls callback function twice (patch included) 


ptrxyz:
Output of `awesome --version`:
```
awesome v3.5.2-2035-g8dc98e8 (The Fox)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
```
**How to reproduce the issue:**
Simply use `awful.spawn("somecommand", false, somecallback)` somewhere.

**Actual result:**

The callback function gets executed **twice**.

**Expected result:**

Callback gets executed **once**.

ptrxyz:
I did some research on this. It seems that `awful.spawn` sets the callback function as a property of a table that is then passed to `rules.execute`. Then, there is a loop that simply applies all properties to the client object **but then**, in addition, there is a loop that executes all callbacks.

Suggested fix:
Either filter out the callback property in rules.execute or before the properties table gets passed to the rules.execute function. I'd go for the 2nd version, since I don't know where else callbacks are used as properties and I am pretty sure the `completed_with_payload_callback` callback is only used for spawned clients.

I'll try to send a merge request asap.

psychon:
Untested, but how about something like this? This makes `awful.spawn` keep the callback separated instead of merging it with the "normal properties table".
```diff
diff --git a/lib/awful/rules.lua b/lib/awful/rules.lua
index f899685..7f8bdfc 100644
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -492,9 +492,8 @@ function rules.execute(c, props, callbacks)
     end
 end
 
-function rules.completed_with_payload_callback(c, props)
-    rules.execute(c, props, type(props.callback) == "function" and
-        {props.callback} or props.callback )
+function rules.completed_with_payload_callback(c, props, callbacks)
+    rules.execute(c, props, callbacks)
 end
 
 client.connect_signal("spawn::completed_with_payload", rules.completed_with_payload_callback)
diff --git a/lib/awful/spawn.lua b/lib/awful/spawn.lua
index a59d52d..c06776e 100644
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -71,9 +71,11 @@ end
 spawn.snid_buffer = {}
 
 function spawn.on_snid_callback(c)
-    local props = spawn.snid_buffer[c.startup_id]
-    if props then
-        c:emit_signal("spawn::completed_with_payload", props)
[...]

PR #1160, opened 2016-10-09
Fixed double execution of callback functions when applying rules

ptrxyz:
See #1159


psychon:
This now conflicts, because we used #1186 instead.

Should we still special-case `callback` in rules? Perhaps by explicitly ignoring it and printing a warning about it? I have to admit, this "callback is not part of a rule" often confuses me as well.

Elv13:
Closed by #1186

Issue #1098, opened 2016-09-15
Blacklist broken ldoc versions


Elv13:
If `ldoc` is installed, we need to check its version and ignore it if it's one of the broken one. If we don't, then there is build "errors". So far, we forced the version on the CI, but that doesn't help the end users who have it installed.

See https://github.com/stevedonovan/LDoc/issues/248

psychon:
Can someone try the following and turn it into a PR?
```patch
diff --git a/awesomeConfig.cmake b/awesomeConfig.cmake
index 4208d67..075d465 100644
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -42,6 +42,14 @@ a_find_program(LDOC_EXECUTABLE ldoc FALSE)
 if(NOT LDOC_EXECUTABLE)
     a_find_program(LDOC_EXECUTABLE ldoc.lua FALSE)
 endif()
+if(LDOC_EXECUTABLE)
+    execute_process(COMMAND sh -c "ldoc --sadly-ldoc-has-no-version-option 2>&1  | grep ' vs 1.4.4'"
+                    OUTPUT_VARIABLE LDOC_VERSION_RESULT)
+    if(NOT LDOC_VERSION_RESULT STREQUAL "")
+        message(WARNING "Ignoring LDoc, because version 1.4.4 is known to be broken")
+        unset(LDOC_EXECUTABLE)
+    endif()
+endif()
 # theme graphics
 a_find_program(CONVERT_EXECUTABLE convert TRUE)
 # pkg-config
```
Good night

blueyed:
Shouldn't this check for 1.4.5? (as per https://github.com/awesomeWM/awesome/issues/1127#issuecomment-250826494)

psychon:
If I understand https://github.com/actionless/awesome/commit/74669f4bc3eaa0a0a15c76a524797340fee5583d correctly, then this is not actually fixed?

wangjiezhe:
I build successfully with the newest git version of ldoc.

psychon:
Yup, newest version seems to work fine. I guess only rejecting version 1.4.5 remains to be done then? (Or: Shown that this is not needed)

blueyed:
Yes, finally fixed (since ~20 days already, see https://github.com/stevedonovan/LDoc/issues/248#issuecomment-258492278).

> I guess only rejecting version 1.4.5 remains to be done then?

Yes.

PR #1161, opened 2016-10-10
Ignore warnings from ldoc by default

psychon:
A while ago, we made errors from ldoc fatal by default. Then a new ldoc
release appeared and caused problems for all of our users, because
awesome failed to work.

This patch reverts the previous fix so that we ignore ldoc warnings by
default again. However, to catch ldoc warnings on Travis, another
ldoc-building-target is added that fails on warnings. This new target is
included in our "check" target.

Helps-with: https://github.com/awesomeWM/awesome/issues/1098

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,8 @@ include_directories(
     ${AWESOME_REQUIRED_INCLUDE_DIRS}
     ${AWESOME_OPTIONAL_INCLUDE_DIRS})
 
+set(CHECK_TARGETS check-integration)
+
 set(AWE_CONF_FILE_DEFAULT ${BUILD_DIR}/awesomerc.lua)
 set(AWE_CONF_FILE rc.lua)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -272,17 +274,24 @@ if(GENERATE_DOC)
     # Copy the aliases to the build directory
     file(COPY ${SOURCE_DIR}/docs/aliases DESTINATION ${BUILD_DIR}/docs)
 
+    add_custom_target(ldoc ALL
+        COMMAND ${LDOC_EXECUTABLE} .
+        WORKING_DIRECTORY ${AWE_DOC_DIR}
+        DEPENDS ${AWE_LUA_FILES} ${AWE_MD_FILES} ${BUILD_DIR}/docs/06-appearance.md
+        ${BUILD_DIR}/docs/05-awesomerc.md
+    )
```
blueyed: I think we should use `.../docs/*.md.lua` (with the `.lua` extension removed) in the DEPENDS here.

Elv13:
Please also add `Fix #1098` in the commit message. While it doesn't "blacklist" anything, it does solve the intent of the issue (users not being able to install awesome).

actionless:
this solves the build issue for me

CMakeLists.txt
blueyed: @psychon 
Can you address this please, if it makes sense?

psychon: (1) CMake is ugly. (2) Done.

psychon:
> Please also add Fix #1098 in the commit message.

Done (although I still prefer full URLs instead of this short form; we already have lots of "FS#1234"-references in the git history that are now that we do not use Flyspray anymore a lot harder to understand; but I seem to be alone with this opinion)

blueyed:
> but I seem to be alone with this opinion

No, I also prefer the long form.

Thanks!

Issue #1173, opened 2016-10-16
Menubar freeze and causing awesome to restart


hexchain:
Output of `awesome --version`:

```
$ awesome --version
awesome v3.5.2-2142-g6b90ed7 (The Fox)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
```

**How to reproduce the issue:**
1. Freeze:
   1. start awesome
   2. `echo 'require("menubar").show()' | awesome-client`
2. Crash:
   1. start awesome
   2. `echo 'require("menubar").refresh()' | awesome-client`

**Actual result:**
1. Menubar does not react on keyboard input. Input goes to the focused window.
2. Awesome crashes:

```
2016-10-16 13:14:47 E: Error opening directory '/usr/local/share/applications': No such file or directory
2016-10-16 13:14:47 W: awesome: luaA_panic:449: unprotected error in call to Lua API (/usr/share/awesome/lib/menubar/init.lua:199: attempt to index a nil value (field 'prompt'))
2016-10-16 13:14:47 W: awesome: luaA_panic:452: dumping backtrace
awesome(backtrace_get+0x54) [0x4218b5]
awesome() [0x417610]
/usr/lib/liblua.so.5.3(+0xd2fd) [0x7f4468f852fd]
/usr/lib/liblua.so.5.3(+0xcdcd) [0x7f4468f84dcd]
/usr/lib/liblua.so.5.3(+0x9e59) [0x7f4468f81e59]
/usr/lib/lua/5.3/lgi/corelgilua51.so(+0x9942) [0x7f4460f90942]
/usr/lib/libffi.so.6(ffi_closure_unix64_inner+0x1cf) [0x7f4467991fbf]
/usr/lib/libffi.so.6(ffi_closure_unix64+0x46) [0x7f4467992330]
/usr/lib/libgio-2.0.so.0(+0x53160) [0x7f446803b160]
/usr/lib/libgio-2.0.so.0(+0x8a323) [0x7f4468072323]
/usr/lib/libgio-2.0.so.0(+0x8a359) [0x7f4468072359]
/usr/lib/libglib-2.0.so.0(g_main_context_dispatch+0x15a) [0x7f446b81043a]
/usr/lib/libglib-2.0.so.0(+0x4a7f0) [0x7f446b8107f0]
/usr/lib/libglib-2.0.so.0(g_main_loop_run+0xc2) [0x7f446b810b12]
[...]

PR #1175, opened 2016-10-18
menubar: Fix missing exe_callback in args table

hexchain:
In https://github.com/awesomeWM/awesome/commit/07f3a178fa3de7dc0de0b5c39661dcaf4ed75d0b the `exe_callback` argument was left out, but at [lib/awful/prompt.lua#L367](https://github.com/awesomeWM/awesome/blob/master/lib/awful/prompt.lua#L367) it is checked for existence, and in this case it causes `prompt.run` to return before actually showing the prompt.

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -444,6 +444,7 @@ function menubar.show(scr)
     awful.prompt.run(setmetatable({
         prompt              = "Run: ",
         textbox             = instance.prompt.widget,
+        exe_callback        = function() end,
         completion_callback = awful.completion.shell,
         history_path        = awful.util.get_cache_dir() .. "/history_menu",
         done_callback       = menubar.hide,


Elv13:
Ah, ok, thanks. I would however prefer `if not textbox or not (exe_callback or args.hooks) then` to be changed so the callback isn't necessary anymore. I don't like these "do nothing" arguments. If there is use cases where it isn't required, then it should not be required. But really, thanks for looking into this.

hexchain:
Totally agreed. I only add them back because they are missing and causing problems.

psychon:
I'm reading @Elv13's comment as "there is some further changes to be done to make things nicer". This leaves this PR as "fix something that was broken in the simplest way possible". Hence, adding this to next. (How the heck were I unable to reproduce this?!?)

hexchain:
I'm wondering too, as both of you are unable to reproduce this, but looking from the code it's sure (I think?) to fail.
Did you see the "Run:" prompt when calling menubar? And could you type in menubar?

actionless:
@hexchain could you then make the change as Elv13 is suggesting?

Bew78LesellB:
It seems that my comment has been ignored in #1173 : 

As `Enter` key is already handled in `keypressed_callback`, there is no `exe_callback`.
Also, I'm wondering why special keys are handled in that `keypressed_callback` instead of having `hooks`.. Isn't it why `hooks` exists ?

To have a nicer fix : 
- change how the keys are handled in menubar (by using `hooks` ?)
- just put `Enter` in a hook ? or in `exe_callback` (and other special keys in `hooks` ?
- change the `prompt.run` to accept not having a `exe_callback` or a `hooks` argument as suggested by @Elv13 

The first solution seems to be the best to me, what do you think ?

hexchain:
@Bew78LesellB No, your comment is not ignored. I think it's up to Elv13 and psychon's decision to choose a nicer fix.

psychon:
> It seems that my comment has been ignored in #1173 :

Do I understand correctly that you are saying "I wrote a comment in #1173, but it was ignored"? If so: Where? I cannot find any comment in there from you (I wanted to link to it in #1180).

hexchain:
https://github.com/awesomeWM/awesome/issues/1173#issuecomment-254563863 ?

psychon:
Sorry, I must have been blind.

PR #1181, opened 2016-10-23
gears.matrix: Add create_rotate_at()

psychon:
We already have a variant of this function for transforming an actual
matrix. This adds the corresponding static factory.

I'm just giving the first, kind-of-unrelated commit from #1156 its own PR. This is that PR.


actionless:
will try my luck merging the first PR :)

psychon:
Thanks :)

PR #1182, opened 2016-10-23
Screen swap

### LINKED ISSUE #1122: Allow to change screen order in configuration (to support 3+ monitors setups)
If I understand things correctly, currently the order of the screens in awesome is determined by the (rather arbitrary) order in which they are returned by xrandr. And if I am correct, it cannot be changed.

This causes issues in 3 (or more) monitors setups, when the actual position of screens does not correspond to the order returned by xrandr. Some commands, such as `awful.screen.focus_relative`, or `awful.client.movetoscreen(c, c.screen+1)`, do not behave as expected by the user.

I have such an issue with my current setup (1 laptop screen + 2 externals screens).
The "physical" order of the screens is A, B, C, but the order they appear when listed by xrandr is A, C, B. When using commands such as `focus_relative` or `movetoscreen`, the focus will jump from A to C then to B, instead of the "normal" order.

My proposal would be to allow changing the order of the screens manually from the lua API. This would open the possibility for users to at least hard-code the screen layout in their config (after all, my screen setup does not change much), as well as enabling the possibility to define automatic "screen layout" strategies (left-to-right then top-to-bottom) via custom lua function.

Not sure how easy that would be however...

(as a sidenote, this it is _not_ a X configuration issue. The screens are correctly configured, and the mouse moves from A->B->C normally)

 • Build: Mar 27 2016 02:23:55 for x86_64 by gcc version 5.3.1 (pbuilder@minjo)
 • D-Bus support: ✔

psychon:
Implement `s1:swap(s2)` to change the order of screens. This fixes #1122.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -2343,13 +2343,11 @@ luaA_client_swap(lua_State *L)
         luaA_object_push(L, swap);
         lua_pushboolean(L, true);
         luaA_object_emit_signal(L, -4, "swapped", 2);
-        lua_pop(L, 2);
 
         luaA_object_push(L, swap);
         luaA_object_push(L, c);
         lua_pushboolean(L, false);
         luaA_object_emit_signal(L, -3, "swapped", 2);
-        lua_pop(L, 3);
     }
 
     return 0;
--- a/objects/screen.c
+++ b/objects/screen.c
@@ -79,6 +79,10 @@
  * @signal removed
  */
 
+/** This signal is emitted when the list of available screens changes.
+ * @signal .list
+ */
+
  /**
   * The primary screen.
   *
@@ -603,6 +607,7 @@ screen_refresh(void)
 
     screen_array_t new_screens;
     lua_State *L = globalconf_get_lua_State();
+    bool list_changed = false;
 
     screen_array_init(&new_screens);
     if (globalconf.have_randr_15)
@@ -624,6 +629,8 @@ screen_refresh(void)
              * globalconf.screens reference this screen now */
             luaA_object_push(L, *new_screen);
             luaA_object_ref(L, -1);
+
+            list_changed = true;
         }
     }
 
@@ -642,6 +649,8 @@ screen_refresh(void)
             lua_pop(L, 1);
             luaA_object_unref(L, old_screen);
             old_screen->valid = false;
+
+            list_changed = true;
         }
     }
 
@@ -656,6 +665,9 @@ screen_refresh(void)
     screen_array_wipe(&new_screens);
 
     screen_update_primary();
+
+    if (list_changed)
+        luaA_class_emit_signal(L, &screen_class, "list", 0);
 }
 
 /** Return the squared distance of the given screen to the coordinates.
@@ -1088,6 +1100,7 @@ luaA_screen_fake_add(lua_State *L)
     s->geometry.height = height;
 
     screen_added(L, s);
+    luaA_class_emit_signal(L, &screen_class, "list", 0);
     luaA_object_push(L, s);
 
     return 1;
@@ -1109,6 +1122,7 @@ luaA_screen_fake_remove(lua_State *L)
     luaA_object_push(L, s);
     screen_removed(L, -1);
     lua_pop(L, 1);
+    luaA_class_emit_signal(L, &screen_class, "list", 0);
     luaA_object_unref(L, s);
     s->valid = false;
 
--- a/objects/screen.c
+++ b/objects/screen.c
@@ -83,6 +83,12 @@
  * @signal .list
  */
 
+/** When 2 screens are swapped
+ * @tparam screen screen The other screen
+ * @tparam boolean is_source If self is the source or the destination of the swap
+ * @signal .swapped
+ */
+
  /**
   * The primary screen.
   *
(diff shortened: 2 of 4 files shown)


Elv13:
Maybe just a little addition to test-screen-change.lua just so the code is called, then 👍

blueyed:
Thanks!

What do you think of providing a default hook / implementation to specify the order?
(Ref: https://github.com/awesomeWM/awesome/issues/1122#issuecomment-250736450)

psychon:
@blueyed You mean something like `awful.util.screen.order_by(some-function-similar-to-what-table.sort-gets)`? Sure, I can do that.

blueyed:
@psychon yes, that's what I've understood from it.
@tomjorquera what do you think?

tomjorquera:
Obviously I think that's really great! :smile: 

Nothing much more to add to my previous comment. Having some kind hook like this to override the default order seems a rather handy addition for people with similar (or more complicated) setups to me.

psychon:
Sigh. Sorting screens by default breaks `tests/_multi_screen.lua`. It verifies that it ended up with the screens that it expected, but by then the screens were already sorted into a different order....

psychon:
Updated to add a test for the originally added functionality, as requested by @Elv13.
The following would be nice to have to help @tomjorquera, but for now it breaks the test suite.
```diff
diff --git a/lib/awful/screen.lua b/lib/awful/screen.lua
index 5a056a3..caab42f 100644
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -30,6 +30,48 @@ local screen = {object={}}
 local data = {}
 data.padding = {}
 
+--- Define a total order on screens.
+-- This function compares two screens. It returns true if the first screen
+-- should be before the second screen. This is used to define an order on
+-- screens.
+--
+-- The default implementation compares screens first by their x-coordinate and
+-- then their y-coordinate.
+--
+-- If you want a different screen order, then you can just overwrite this
+-- function.
+-- @usage awful.screen.order_function = function(s1, s2)
+--   local geo1, geo2 = s1:geometry(), s2:geometry()
+--   return geo1.x < geo2.x or (geo1.x == geo2.x and geo1.y < geo2.y)
+-- end
+-- @function awful.screen.order_function
+-- @tparam screen s1 The first screen to compare
+-- @tparam screen s2 The second screen to compare
+-- @treturn bool a truish value if 's1 < s2', else a falseish value.
+function screen.order_function(s1, s2)
+    local geo1, geo2 = s1.geometry, s2.geometry
+    return geo1.x < geo2.x or (geo1.x == geo2.x and geo1.y < geo2.y)
+end
+
+function screen.order_screens()
+    print("sorting screens")
+    -- Just a bubble sort to order screens
+    local stop = false
+    while not stop do
+        stop = true
+        for i=1, capi.screen.count()-1 do
+            if screen.order_function(capi.screen[i+1], capi.screen[i]) then
+                print("swapping", i, i+1)
+                capi.screen[i]:swap(capi.screen[i+1])
+                stop = false
+            end
+        end
+    end
+end
+
+capi.screen.connect_signal("list", screen.order_screens)
+capi.screen.connect_signal("property::geometry", screen.order_screens)
+
 --- Take an input geometry and substract/add a delta.
 -- @tparam table geo A geometry (width, height, x, y) table.
 -- @tparam table delta A delta table (top, bottom, x, y).
diff --git a/spec/awful/screen_spec.lua b/spec/awful/screen_spec.lua
index 7a9a31e..55b3b1d 100644
--- a/spec/awful/screen_spec.lua
+++ b/spec/awful/screen_spec.lua
@@ -6,6 +6,7 @@
 local fake_screens = {}
 
 _G.screen = setmetatable({
+    connect_signal = function() end,
     set_index_miss_handler = function() end,
     set_newindex_miss_handler = function() end,
 }, {
```

Elv13:
Look good once the lua 5.3 issue has been fixed

psychon:
Fixed the Lua 5.3 issue and also added the final version #1195 into this, since they are loosely related (this PR made me find the issue in the code that I copied).

blueyed:
As for tests/_multi_screen.lua: couldn't it unset/overwrite the order_function and/or disconnect the signals temporarily?

psychon:
That would fix the immediate test suite failure, yeah. However, `_multi_screen.lua` for example explicitly does "screen 1 left, screen 2 right" and later "screen 2 left, screen 1 right". Such things would be silently removed if we do that. So, the question more or less is "how random should the screen orders be that we test against?".

blueyed:
I don't know..
@Elv13 ?

Elv13:
> I don't know..
> @Elv13 ?

The multi screen test was designed to test each possible corner case I could either think of, or caused breakages in the past. Having screen 2 at 0x0 caused quite many issues in my modules (and probably in Awesome too). This is why my modules often kept their own screen array. If it can no longer happen, then what's the point of testing it. However we have to be careful before automagically messing with the screen order. For example, of a grid of screen have 2 row. screen.primary/screen[1] is most probably the bottom left, but auto-messing with the order would probably pick the top left (I didn't read the patch above carefully, this is beside the point). Also, I don't like to have "please monkey patch this function" APIs. It's bad API and it make creating modules harder because they don't know what to expect. I am not against adding an helper to re-order screens, but I am against doing so by default and proposing a default implementation. This will turn into the "current screen" mess all over again. Sometime, it is better to leave things as-is.

psychon:
How about getting this PR merged and thinking about an enforcable screen order later? The current state of this PR only provides `s1:swap(s2)`. The auto-magic Lua side of this is only available in a comment above.

blueyed:
> How about getting this PR merged and thinking about an enforcable screen order later? 

Yeah, push the button and :sheep: :it:.. ;)

PR #1186, opened 2016-10-24
awful.spawn: Separate rules from callbacks

psychon:
When adding callbacks as a `callback` entry in a property, the callback
is run by `awful.rules`, because it does `c.callback =
result_of_function`. This is obviously not intended. Also, this causes
the callbacks to run twice, because the code already handled this
`callback` property specially.

Fix this by just not merging callbacks with the normal rules at all.

Fixes: https://github.com/awesomeWM/awesome/issues/1159

This conflicts with #1160. You'll have to decide between this and the other PR.

### DIFF
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -491,9 +491,8 @@ function rules.execute(c, props, callbacks)
     end
 end
 
-function rules.completed_with_payload_callback(c, props)
-    rules.execute(c, props, type(props.callback) == "function" and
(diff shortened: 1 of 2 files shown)


ptrxyz:
This patch changes the function's behavior:
Previously `client.spawn("cmd", false, callback)` was executing callback which is no longer the case with this patch. However I have no idea if this is intended just wanted to point it out.

psychon:
@ptrxyz That should finally be fixed by #1240.

PR #1187, opened 2016-10-24
awful.util.file_readable: Use Gio

psychon:
Instead of doing Linux-specific magic with error codes and trying to
read the first byte of a file, just use Gio to check if a file exists
and is readable.

### DIFF
--- a/lib/awful/util.lua
+++ b/lib/awful/util.lua
@@ -8,7 +8,6 @@
 
 -- Grab environment we need
 local os = os
-local io = io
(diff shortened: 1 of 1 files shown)


blueyed:
Wasn't there some crash related issues with that approach?

psychon:
I don't remember anything. I just noticed that `dir_readable` and `file_readable` do quite different things. If there is a crash-issue, than that already affects the existing `dir_readable`.

PR #1188, opened 2016-10-24
wibox: Add gears.shape support.

actionless:
see https://github.com/awesomeWM/awesome/pull/1044/commits/1cc613338c8e0c645f4dc835388e85fc62cf4276

i don't mind refactoring this for moving the code to wibox part, however i don't understand the rest of the problem from @psychon's comment

### DIFF
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -145,6 +145,48 @@ local function do_redraw(self)
         self._widget_hierarchy:draw(context, cr)
     end
 
+    -- If a `gears.shape` is set and the border_width is greater than 2,
+    -- re-apply the border to fix anti-aliasing.
+    if self._widget_context_skeleton.wibox then
+        local wb = self._widget_context_skeleton.wibox
+
+        local shape = wb._shape
+        local bw = wb.border_width
+
+        -- The "real" size is width+2*bw and height+2*bw. This code only exist
+        -- to draw the inner anti-aliased border just 1px inside of the clip.
+        if shape then
+            local bc = wb.border_color or beautiful.border_normal or beautiful.fg_normal
+
+            cr:save()
+
+            -- First, draw the actual border. This will also include an
+            -- anti-aliased inner border that the clip shape could not provide.
+            cr:move_to(0, 0)
+            cr:translate(-bw, -bw)
+            cr:set_source(color(bc))
+            shape(cr, width+2*bw, height+2*bw)
+
+            -- Add 1px when the border_width is 1 to compensate for the
+            -- outer border anti-alising. Yes, this does steal half a pixel
+            -- that should allow visible content.
+            cr:set_line_width(bw*2 + (bw == 1 and 1 or 0))
+            cr:stroke_preserve()
+
+            -- Finally, clear the outer pixel to create an anti-alised effect
+
+            -- Odd numbers are odds, but this is mostly due to the A1 shape
+            -- bounding own anti-alising.
+            cr:set_line_width(1 + (bw % 2)/2)
+
+            cr:set_operator(cairo.Operator.CLEAR)
+            cr:set_source_rgba(0,0,0,1)
+            cr:stroke()
+
+            cr:restore()
+        end
+    end
+
     self.drawable:refresh()
 
     assert(cr.status == "SUCCESS", "Cairo context entered error state: " .. cr.status)
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -15,8 +15,10 @@ local pairs = pairs
 local type = type
 local object = require("gears.object")
 local grect =  require("gears.geometry").rectangle
+local gshape =  require("gears.shape")
 local beautiful = require("beautiful")
 local base = require("wibox.widget.base")
+local cairo = require("lgi").cairo
 
 --- This provides widget box windows. Every wibox can also be used as if it were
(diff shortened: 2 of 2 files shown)


psychon:
Half of my problem is "there is no `wibox`-specific code in `wibox.drawable`". Such code should live in `wibox`. However, let's consider this an unimportant implementation detail for now.

The other half is "I don't understand what this code does". I can look at the individual function calls and see that it does some magic with "twice the border width". What I don't understand is why it does this. Perhaps it would be best to have a big comment (next to `apply_shape`? Dunno) explaining how all this works and how compositing managers come into play.

If I read this correctly, the drawing code draws the wibox' border, too, so that it can (badly, if no compositing manager is running*) simulate anti-aliasing on the border. Is that the whole point of this code?

*: Ok, so you now get anti-aliasing on the inside of the border (parts of the border appear in the area of the actual content). However, there is no anti-aliasing in the order direction (parts of the content appear in the area of the border). However, for proper anti-aliasing, I think that both things should happen.

P.S.: Oh and this sounds like a TODO:

> Note that this code doesn't handle compositing manager being turned on and off.

P.P.S.: Feel free to point at any code that I wrote which needs such an "explain the following magic"-comment.

lib/wibox/init.lua
```diff
@@ -61,6 +63,106 @@ function wibox:find_widgets(x, y)
...
+        cr:set_operator(cairo.Operator.SOURCE)
+        cr:set_source_rgba(1,1,1,1)
+        cr:paint()
+    end
+
+
+    self.shape_clip = img._native
+
+    -- Force the drawable to redraw the inner border
+    self:draw()
```
psychon: Oh and this doesn't actually work. `draw()` will start a `delayed_call` to do a redraw. That call will then see "there is no need to re-layout widgets and the dirty area (which pixels in the wibox need to be redrawn?) is empty, so I don't have anything to do". Thus, this will not work.

Elv13:
image

It does 2 passes because it has to create 2 different anti-aliased sections. One for the X11 mask and the other for the cairo surface

Elv13:
As for some of the "twice" and "half" border part. Remember that cairo line_width is applied half outside of the path and half inside. So if you want a 4px border and your current path is the outer pixels of your mask, then 4px == 8px/2

Elv13:
@actionless Can you integrate my previous comment into the patch (or grant me access so I can do it, but I don't have time for this right now)?

actionless:
@Elv13 
i've added you, from the comment i don't understand what should be added/fixed

Elv13:
Apparently @psychon will take care of that (we spoke on IRC), you can also add him to the repository whitelist.

actionless:
i've also added @blueyed just in case :)

psychon:
Quick test: With
```diff
diff --git a/awesomerc.lua b/awesomerc.lua
index 87d9da7..b2f857e 100755
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -210,6 +210,7 @@ awful.screen.connect_for_each_screen(function(s)
 
     -- @DOC_WIBAR@
     -- Create the wibox
+    --[[
     s.mywibox = awful.wibar({ position = "top", screen = s })
 
     -- @DOC_SETUP_WIDGETS@
@@ -231,6 +232,15 @@ awful.screen.connect_for_each_screen(function(s)
             s.mylayoutbox,
         },
     }
+    --]]
+    s.mywibox = wibox {
+        bg = "#ff0000",
+        border_width = 3,
+        border_color = "#00ff00",
+    }
+    s.mywibox:geometry{ x=10, y=10, width=30, height = 30}
+    s.mywibox.shape = gears.shape.octogon
+    s.mywibox.visible = true
 end)
 -- }}}
 
diff --git a/lib/wibox/init.lua b/lib/wibox/init.lua
index f081e09..abe176f 100644
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -92,6 +92,8 @@ local function apply_shape(self)
     cr:fill()
 
     self.shape_bounding = img._native
+    img:write_to_png("/tmp/bounding.png")
+    print("wrote bounding")
 
     -- Now, set an empty clip. The wibox clip is drawn using cairo in
     -- drawable.lua so anti-alising work.
@@ -129,6 +131,8 @@ local function apply_shape(self)
 
 
     self.shape_clip = img._native
+    img:write_to_png("/tmp/clip.png")
+    print("wrote clip")
 
     -- Force the drawable to redraw the inner border
     self:draw()
```
I get the following. Notice the "weird spots of dark-ish pixels". That doesn't look good. These dark-ish pixels aren't present in the clip and bounding shape, so must be due to the code in `wibox.drawable`.
Screenshot: shaped Bounding: bounding Clip: clip

lib/wibox/drawable.lua
```diff
@@ -145,6 +145,48 @@ local function do_redraw(self)
...
+
+            -- Finally, clear the outer pixel to create an anti-alised effect
+
+            -- Odd numbers are odds, but this is mostly due to the A1 shape
+            -- bounding own anti-alising.
+            cr:set_line_width(1 + (bw % 2)/2)
+
+            cr:set_operator(cairo.Operator.CLEAR)
+            cr:set_source_rgba(0,0,0,1)
+            cr:stroke()
```
psychon: I guess this is where the black-ish pixels come from. Why should there be black (without compositing manager) or transparent (with compositing manager) pixels between the content and the border?

PR #1189, opened 2016-10-25
feat(naughty: core): add optional border colors, shape and shape_args

actionless:
depends on https://github.com/awesomeWM/awesome/pull/1188

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -22,7 +22,7 @@ local bt = require("beautiful")
 local wibox = require("wibox")
 local surface = require("gears.surface")
 local cairo = require("lgi").cairo
-local dpi = require("beautiful").xresources.apply_dpi
+local dpi = bt.xresources.apply_dpi
 
 local function get_screen(s)
     return s and capi.screen[s]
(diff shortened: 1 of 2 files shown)


psychon:
Mostly fine with me, but I can already see this needing a margin inside the notification...

actionless:
@psychon there is already `naughty.config.defaults.margin` for that

psychon:
...and I even looked for something, but yay for undocumented properties.

actionless:
it's documented as a default param https://github.com/awesomeWM/awesome/pull/1189/files#diff-488c2f784e0805010c93f1c3a4803557R107 but indeed not documented as an notification argument, i will add this to the PR

lib/naughty/core.lua
```diff
@@ -603,6 +613,10 @@ function naughty.notify(args)
                                bg = bg,
                                border_color = border_color,
                                border_width = border_width,
+                               shape_border_color = shape and border_color,
+                               shape_border_width = shape and border_width,
+                               shape = shape,
+                               shape_args = shape_args,
```
actionless: btw, shape_args seems to be ignored now by wibox, so in case if it's not going to be changed, in naughty it should be changed like:

```lua
shape = function(c, w, h) return shape(c, w, h, table.unpack(shape_args)) end
```

Elv13:
sooooo

The dependencies have been resolved. `shape_border_color` and `shape_border_width` don't really exist, so could be removed into inner_border_width is implemented (not part of this PR).

Also, while there is `presets`, this is inconsistent with the other APIs. Can another commit add "all" beautiful variables (including the shape)?

actionless:
but isn't `config.default` is a default preset? border_width and other stuff is also defined there

Elv13:
Yes, I know, but as we add more appearance features into this, we end up with a new problem. Themes cannot "really" set the notification look&feel without creating a "forbidden"[1] `require()`... So the cleanest way to solve that is to add beautiful variables for the look&feel related properties.

[1] The new require policy introduced by a new PR from today.

psychon:
> [1] The new require policy introduced by a new PR from today.

That one does not check themes. Also, a theme is loaded outside of any active call to `require()`, so it cannot cause dependency loops.

Elv13:
> That one does not check themes. Also, a theme is loaded outside of any active call to require(), so it cannot cause dependency loops.

Ok, but it doesn't change the fact that `beautiful` initialization (if the user were to put the look&feel prefer there) would depend on naughty and naughty depends on beautiful.

Currently, the "semi-loop" is "solved" using `local beautiful = bt.get()`

actionless:
updated, let's get it merged before the release :grinning:

Issue #171, opened 2015-03-19
RFC: Set c.border_width to 0 when maximized


Elv13:
Someone asked for how to do this on IRC. I think it is a good idea, should we do this by default?

Addition to `rc.lua`:

``` lua
client.connect_signal("property::maximized", function(c) 
    c.border_width = c.maximized and 0 or beautiful.border_width
end)
```

actionless:
IMO it makes sense

blueyed:
:+1:

I am doing this myself, with additional rules for maximized and fullscreen for example, and therefore it gets done in the `arrange` handler.
This could be done via `property::fullscreen` and `property::maximized`, too.

syphoxy:
on this note, is there a patch for the maximized layout to disable borders?

r8b7xy:
This is my way to disable borders for max layout. It's based on [copycat's config](https://github.com/copycat-killer/awesome-copycats).
```lua
-- No border for maximized clients

client.connect_signal("focus",
  function(c)
    if c.maximized_horizontal == true and c.maximized_vertical == true then
      c.border_color = beautiful.border_normal
    else
      c.border_color = beautiful.border_focus
    end
  end)

client.connect_signal("unfocus",
  function(c) c.border_color = beautiful.border_normal
  end)

-- Arrange signal handler
for s = 1, screen.count() do screen[s]:connect_signal("arrange", 
  function ()
    local clients = awful.client.visible(s)
    local layout  = awful.layout.getname(awful.layout.get(s))

    if #clients > 0 then -- Fine grained borders and floaters control
      for _, c in pairs(clients) do -- Floaters always have borders
        if awful.client.floating.get(c) or layout == "floating" then
          c.border_width = beautiful.border_width

        -- No borders with only one visible client
        elseif #clients == 1 or layout == "max" then
          c.border_width = 0
        else
          c.border_width = beautiful.border_width
        end
      end
    end
  end)
end

syphoxy:
that's really good, I think. I was thinking about making it so that any layout removes the borders with clients == 1 so this really goes above and beyond my original question and above my wishful thinking.

I vote to include this in the default awesome config as well.

blueyed:
My config is also based on copycat's:

    -- Handle border sizes of clients.
    for s = 1, screen.count() do screen[s]:connect_signal("arrange", function ()
      local clients = awful.client.visible(s)
      local layout = awful.layout.getname(awful.layout.get(s))

[...]

PR #1191, opened 2016-10-25
Emit screen::arrange signal outside of arrange_lock

blueyed:
This will handle changes in the layout recursively, e.g. when changing
the border_width of clients.

Ref: https://github.com/awesomeWM/awesome/issues/171#issuecomment-256146578

The patch is from Uli, but he does not want to be blamed: https://github.com/awesomeWM/awesome/issues/171#issuecomment-256150715 ;)

As for problems this might cause: should there be some "maximum recursion" guard?  Or do we have it implicitly already?

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -190,10 +190,10 @@ function layout.arrange(screen)
             g.y = g.y + useless_gap
             c:geometry(g)
         end
-        screen:emit_signal("arrange")
-
         arrange_lock = false
         delayed_arrange[screen] = nil
+
+        screen:emit_signal("arrange")
     end)
 end
 


psychon:
> As for problems this might cause: should there be some "maximum recursion" guard? Or do we have it implicitly already?

Lua has a maximum stack size. Since (non-tail)-calls surely use some Lua stack, this is effectively a maximum recursion depth that ignores tail calls.  So eventually whatever causes the recursion will be stopped with a Lua error.
```
$ lua -e 'local function f(a) return 1+f(a) end f()'
lua: (command line):1: stack overflow
[...]
```

psychon:
Oh, even better: Layouts are arranged in a delayed call. So what I wrote above is wrong and instead you get an endless loop (main loop iteration never finishes, because there are always pending delayed calls).

blueyed:
> Oh, even better

Wouldn't an infinite loop be bad, and should be detected/aborted?

blueyed:
OTOH you might want this in some cases even?!

psychon:
I don't really see what we could do about it and it would always be a bug in the config. Causing a new arrange from the `arrange` signal is never a good idea and should be avoided, but it will work in most cases...

PR #1195, opened 2016-10-28
WIP: Test swapping clients

psychon:
I accidentally noticed that this is broken (and the current tests never run `c:swap()`). Good that we are compiling Lua 5.3 in debug mode!


psychon:
Yay. Calling `c:swap()` reproducibly crashes awesome when run with a debug build of Lua:
> awesome: lapi.c:182: lua_settop: Assertion `(-(idx+1) <= (L->top - (func + 1))) && "invalid new top"' failed.
2016-10-28 17:03:11 E: awesome: signal_fatal:407: signal 6, dumping backtrace

PR #1196, opened 2016-10-29
Add more information to the --version output

psychon:
This should now handle all #ifdef's that we react to.

### DIFF
--- a/common/version.c
+++ b/common/version.c
@@ -28,6 +28,7 @@
 
 #include <lualib.h>
 #include <lauxlib.h>
+#include <xcb/randr.h> /* for XCB_RANDR_GET_MONITORS */
 
 /** \brief Print version message and quit program.
  * \param executable program name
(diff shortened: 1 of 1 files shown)


Elv13:
Maybe also LGI, Gio and glib/gobject_introspection versions? Beside that, 👍

psychon:
> Maybe also LGI, Gio and glib/gobject_introspection versions? Beside that, :+1:

That'd be a tiny bit harder, since it requires a little more setup in the Lua context. For now I just did the simple stuff, but something like the following would be needed `print(require("lgi.version"))`. Dunno how to get the exact Gio version or things like that.

psychon:
Well, ok. Here is a new commit that also adds the LGI version number. If `lgi.version` is not found, we print the error message from `require` as the version. Looks weird, but I can live with that. :-)
For example, replacing `lgi.version` with `foobar` results for me in:
```
awesome 9999 (The Fox)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: [string "return require('foobar')"]:1: module 'foobar' not found:
	no field package.preload['foobar']
	no file '/usr/local/share/lua/5.2/foobar.lua'
	no file '/usr/local/share/lua/5.2/foobar/init.lua'
	no file '/usr/local/lib/lua/5.2/foobar.lua'
	no file '/usr/local/lib/lua/5.2/foobar/init.lua'
	no file '/usr/share/lua/5.2/foobar.lua'
	no file '/usr/share/lua/5.2/foobar/init.lua'
	no file './foobar.lua'
	no file '/usr/local/lib/lua/5.2/foobar.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/5.2/foobar.so'
	no file '/usr/lib/lua/5.2/foobar.so'
	no file '/usr/local/lib/lua/5.2/loadall.so'
	no file './foobar.so'
```

PR #1198, opened 2016-10-29
Wibox: Add gears.shape support

psychon:
Since I suffer from NIH-syndrome, here is my version of #1188. I'm looking forward to you telling me which magic parts that worked in #1188 now no longer work.

For the complaints that I expect, please check if the following patch ontop on this branch would make you happy:

``` diff
diff --git a/lib/wibox/init.lua b/lib/wibox/init.lua
index 92a2c88..622734a 100644
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -72,18 +72,15 @@ local function draw_hook(self, cr)
     if not shape then return end

     local geo = self:geometry()
-    cr:set_source(gcolor(self.border_color))
+    local bw = self.border_width

-    -- We want to invert the shape (draw to anything that is not inside of it).
-    -- To fake this, we use EVEN_ODD and a rectangle containing the full
-    -- surface. Pixels in the border will be inside the big rectangle, so they
-    -- are contained once in the path. Pixels in the "content area" are
-    -- additionally contained in the shape, so twice.
-    cr:set_fill_rule(cairo.FillRule.EVEN_ODD)
-    cr:rectangle(0, 0, geo.width, geo.height)
-    shape(cr, geo.width, geo.height)
+    cr:translate(-bw, -bw)
+    shape(cr, geo.width + 2*bw, geo.height + 2*bw)

-    cr:fill()
+    -- 2*, because half of them is outside of the visible area
+    cr:set_line_width(2*bw)
+    cr:set_source(gcolor(self.border_color))
+    cr:stroke()
 end

 function wibox:_apply_shape()
```

Hopefully this version is easier to understand than #1188.

### DIFF
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -17,6 +17,7 @@ local object = require("gears.object")
 local grect =  require("gears.geometry").rectangle
 local beautiful = require("beautiful")
 local base = require("wibox.widget.base")
+local cairo = require("lgi").cairo
 
 --- This provides widget box windows. Every wibox can also be used as if it were
 -- a drawin. All drawin functions and properties are also available on wiboxes!
@@ -59,6 +60,53 @@ end
 
 function wibox:find_widgets(x, y)
     return self._drawable:find_widgets(x, y)
+end
+
+function wibox:_apply_shape()
+    local shape = self._shape
+
+    if not shape then
+        self.shape_bounding = nil
+        self.shape_clip = nil
+        return
+    end
+
+    local geo = self:geometry()
+    local bw = self.border_width
+
+    -- First handle the bounding shape (things including the border)
+    local img = cairo.ImageSurface(cairo.Format.A1, geo.width + 2*bw, geo.height + 2*bw)
+    local cr = cairo.Context(img)
+
+    shape(cr, geo.width + 2*bw, geo.height + 2*bw)
+    cr:set_operator(cairo.Operator.SOURCE)
+    cr:fill()
+    self.shape_bounding = img._native
+    img:finish()
+
+    -- Now handle the clip shape (things excluding the border)
+    img = cairo.ImageSurface(cairo.Format.A1, geo.width, geo.height)
+    cr = cairo.Context(img)
+
+    shape(cr, geo.width, geo.height)
+    cr:set_operator(cairo.Operator.SOURCE)
+    cr:fill()
+    self.shape_clip = img._native
+    img:finish()
+end
+
+--- Set the wibox shape.
+-- @property shape
+-- @tparam gears.shape A gears.shape compatible function.
+-- @see gears.shape
+
+function wibox:set_shape(shape)
+    self._shape = shape
+    self:_apply_shape()
+end
+
+function wibox:get_shape()
+    return self._shape
 end
 
 function wibox:get_screen()
@@ -196,6 +244,9 @@ local function new(args)
     -- Make sure the wibox is drawn at least once
     ret.draw()
 
+    ret:connect_signal("property::geometry", ret._apply_shape)
+    ret:connect_signal("property::border_width", ret._apply_shape)
+
     -- If a value is not found, look in the drawin
     setmetatable(ret, {
         __index = function(self, k)
@@ -228,6 +279,8 @@ local function new(args)
     if args.screen then
         ret:set_screen ( args.screen  )
     end
+
+    ret.shape = args.shape
 
     return ret
 end
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -78,6 +78,7 @@ function wibox:_apply_shape()
     local img = cairo.ImageSurface(cairo.Format.A1, geo.width + 2*bw, geo.height + 2*bw)
     local cr = cairo.Context(img)
 
+    -- We just draw the shape in its full size
     shape(cr, geo.width + 2*bw, geo.height + 2*bw)
     cr:set_operator(cairo.Operator.SOURCE)
     cr:fill()
@@ -88,9 +89,18 @@ function wibox:_apply_shape()
     img = cairo.ImageSurface(cairo.Format.A1, geo.width, geo.height)
     cr = cairo.Context(img)
 
-    shape(cr, geo.width, geo.height)
+    -- We give the shape the same arguments as for the bounding shape and draw
+    -- it in its full size (the translate is to compensate for the smaller
+    -- surface)
+    cr:translate(-bw, -bw)
+    shape(cr, geo.width + 2*bw, geo.height + 2*bw)
     cr:set_operator(cairo.Operator.SOURCE)
-    cr:fill()
+    cr:fill_preserve()
+    -- Now we remove an area of width 'bw' again around the shape (We use 2*bw
+    -- since half of that is on the outside and only half on the inside)
+    cr:set_source_rgba(0, 0, 0, 0)
+    cr:set_line_width(2*bw)
+    cr:stroke()
     self.shape_clip = img._native
     img:finish()
 end
--- a/tests/test-wibox-shape.lua
+++ b/tests/test-wibox-shape.lua
@@ -1,0 +1,31 @@
+-- A quick-and-dirty test of wibox shapes ("Does it error out?")
+
+local runner = require("_runner")
+local wibox = require("wibox")
+local shape = require("gears.shape")
+
+local was_drawn
+local widget = wibox.widget.base.make_widget()
+function widget.draw()
+    was_drawn = true
+end
+
+local wb = wibox {
+    shape = shape.powerline,
+    widget = widget,
+    border_width = 42,
+}
+wb:geometry(screen[1].geometry)
+wb.visible = true
+
+runner.run_steps({
+    function()
+        assert(wb.shape == shape.powerline)
+        assert(wb.shape_bounding) -- This is a memory leak! Don't copy!
+        if was_drawn then
+            return true
+        end
+    end
+})
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80


actionless:
this looks cleaner to me

also i've just tested, i am not using compositor and there is no noticeable difference between both implementations in my setup

lib/wibox/init.lua
```diff
@@ -61,6 +63,72 @@ function wibox:find_widgets(x, y)
...
+    end
+
+    local geo = self:geometry()
+    local bw = self.border_width
+
+    -- First handle the bounding shape (things including the border)
+    local img = cairo.ImageSurface(cairo.Format.A1, geo.width + 2*bw, geo.height + 2*bw)
+    local cr = cairo.Context(img)
+
+    shape(cr, geo.width + 2*bw, geo.height + 2*bw)
```
Elv13: Technically, the shape should be called with the same parameter both time. Nothing guarantee shape scaling is linear. Some, like the infobubble have corner cases where they are not. This is why I had that hack you commented on in the other PR. You can work around the issue by adding a bigger border around the smaller shape.

psychon: Should I read this as "I want you to squash the diff that you mentioned in the first comment"? That diff does what you say, I think.

Elv13: Umm, eh, isn't this my first comment here? I just pointed out a little bug in this patch. Note that my Radical module implementation of this has the bug (because I am lazy) and someone complained on IRC about it (a long time ago). It does look weird:

slider10

(not using your patch). See the `\` part of the `''\/''` I hope you see the issue with doing this. My version of this PR had this corner case covered.

psychon: At least now I know why your code did it like this. :-)
The next complain that I expect is "this does not do magic anti-aliasing on the outside when a compositing manager is running".

I'll squash that diff into this commit.

Elv13: My next complain is:

> "this does not do magic anti-aliasing on the outside when a compositing manager is running".

psychon: Ok, here is the patch and a screenshot:
```diff
diff --git a/awesomerc.lua b/awesomerc.lua
index 87d9da7..06ecdb4 100755
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -210,6 +210,7 @@ awful.screen.connect_for_each_screen(function(s)
 
     -- @DOC_WIBAR@
     -- Create the wibox
+    --[[
     s.mywibox = awful.wibar({ position = "top", screen = s })
 
     -- @DOC_SETUP_WIDGETS@
@@ -231,6 +232,18 @@ awful.screen.connect_for_each_screen(function(s)
             s.mylayoutbox,
         },
     }
+    --]]
+    s.mywibox = wibox {
+        x = 100,
+        y = 100,
+        width = 300,
+        height = 300,
+        bg = "#ff0000",
+        border_color = "#00ff00",
+        border_width = 30,
+        shape = gears.shape.circle,
+    }
+    s.mywibox.visible = true
 end)
 -- }}}
 
diff --git a/lib/wibox/init.lua b/lib/wibox/init.lua
index 622734a..24af541 100644
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -74,13 +74,28 @@ local function draw_hook(self, cr)
     local geo = self:geometry()
     local bw = self.border_width
 
+    -- Draw the border shape
     cr:translate(-bw, -bw)
     shape(cr, geo.width + 2*bw, geo.height + 2*bw)
+    cr:translate(bw, bw)
 
     -- 2*, because half of them is outside of the visible area
     cr:set_line_width(2*bw)
     cr:set_source(gcolor(self.border_color))
-    cr:stroke()
+    cr:stroke_preserve()
+
+    -- Now do magic to simulate anti-aliasing for the bounding shape. We
+    -- "invert" the shape via EVEN_ODD and adding a rectangle around everything.
+    -- Pixels inside of the shape will be contained twice in the path
+    -- (rectangle+shape), so an even number of times. Pixels only inside the
+    -- rectangle get an odd number.
+    -- TODO: Only if compositing manager is running
+    cr:rectangle(0, 0, geo.width, geo.height)
+    cr:set_fill_rule(cairo.FillRule.EVEN_ODD)
+    cr:set_operator(cairo.Operator.CLEAR)
+    cr:fill()
+
+    cr:new_path()
 end
 
 function wibox:_apply_shape()
```
foo
If you zoom, you can see that this does have some effect, but I wouldn't say that the result really is good:
foo-zoomed
I don't like the hack from your version were you try magic with the line width, as this causes the artifacts that are visible in the screenshot I posted there.

Should we really try to make X11 do something that it is not meant to do?

Elv13: > Should we really try to make X11 do something that it is not meant to do?

Yes, because everybody else does it and because for rounded rectangles, hexagons and octogons (the 3 most popular use cases), it has no issue.

Also, here's something to help you test:

```lua
local cairo = require( "lgi" ).cairo

-- Shape with hardcoded values or values based on only 1 axis don't necessarely
-- support linear radial scaling from the shape centroid. This forces the shape
-- to always be called with the same with and height or the result will miss
-- part of its border on one side
local function hardcoded_shape(cr, w, h)
    --actually, I didn't test it, but I think it triggers the bug
    cr:move_to(0,10)
    cr:line_to(10, 0)
    cr:line_to(w, 0)
    cr:line_to(w, h-10)
    cr:line_to(w-10, h)
    cr:line_to(0, h)
    cr:line_to(0, 10)
    cr:close_path()
end

-- Set a checkboard wallpaper
local img = cairo.ImageSurface(cairo.Format.ARGB32, 20, 20)
local cr = cairo.Context(img)
cr:set_source_rgb(1,1,1)
cr:paint()
cr:set_source_rgb(0.5,0.5,0.5)
cr:rectangle(0,0,10,10)
cr:rectangle(10,10,10,10)
cr:fill()
local pat = cairo.Pattern.create_for_surface(img)
gears.wallpaper.tiled(img, 1)

local w5 = wibox {
    visible = true,
    width = 200,
    height = 200,
    x= 50,
    y=50,
    ontop = true,
    shape = function(cr,w,h) gears.shape.octogon(cr,w,h, 30) end,
    bg = "#ff0000",
    border_width = 4,
    border_color = "#00ff00",
}
local w6 = wibox {
    visible = true,
    width = 200,
    height = 200,
    x= 270,
    y=50,
    ontop = true,
    shape = hardcoded_shape,
    bg = "#ff0000",
    border_width = 4,
    border_color = "#00ff00",
}
local w7 = wibox {
    visible = true,
    width = 200,
    height = 200,
    x= 490,
    y=50,
    ontop = true,
    shape = gears.shape.circle,
    bg = "#ff0000",
    border_width = 4,
    border_color = "#00ff00",
}
local w8 = wibox {
    visible = true,
    width = 200,
    height = 200,
    x= 50,
    y=270,
    ontop = true,
    shape = function(cr,w,h) gears.shape.octogon(cr,w,h, 30) end,
    bg = "#ff0000",
    border_width = 0,
}
local w9 = wibox {
    visible = true,
    width = 200,
    height = 200,
    x= 270,
    y=270,
    ontop = true,
    shape = hardcoded_shape,
    bg = "#ff0000",
    border_width = 0,
}
local w10 = wibox {
    visible = true,
    width = 200,
    height = 200,
    x= 490,
    y=270,
    ontop = true,
    shape = gears.shape.circle,
    bg = "#ff0000",
    border_width = 0,
}
```

Elv13: Oh, and

```lua

local w11 = wibox {
    visible = true,
    width = 200,
    height = 200,
    x= 50,
    y=490,
    ontop = true,
    shape = function(cr,w,h) gears.shape.octogon(cr,w,h, 30) end,
    bg = "#ff0000",
    border_width = 4,
    border_color = "#00ff00",
}
w11:setup {
    {
        text = "should be pure green",
        widget = wibox.widget.textbox
    },
    bg = "#00ff00",
}
local w11 = wibox {
    visible = true,
    width = 200,
    height = 200,
    x= 270,
    y=490,
    ontop = true,
    shape = function(cr,w,h) gears.shape.octogon(cr,w,h, 30) end,
    bg = "#ff0000",
    border_width = 4,
    border_color = "#00ff00",
}
w11:setup {
    {
        text = "no red at all",
        widget = wibox.widget.textbox
    },
    bg = "#ffffff",
}

(check for the inner black pixel issue of my implementation)
```

Elv13: Yes, that's an issue of my version. Can you take a screenshot using compton if you implemented the compositing anti-aliasing? Also, I added 2 tests for the other corner cases, but I thing your version handle them just fine. This code should be added to the test, while the CI can't check the result, we manually can. As you can see, there is plenty of corner case someone in the far future could say "that's overcomplicated, lets drop that part"

psychon: The current state of this PR + the patch that I gave some comments before with your test code and compton.
own-with-compton
I'm curious how this would look like with your version, but I'm too lazy to test. :-)

Side note: The issue that I mentioned above seems to be a bug in the X11 server. Sigh...

Elv13: Looks great! I think this is it.

It's a little too much to ask given my own version didn't bothered, but can you make it work with clients too (with or without turning them into full systray like drawable widgets)? `awful.client.shape` was added by you a while back in the 4.0 release cycle. It could still be dropped in favor of an API identical to this one without causing an official API break.

Elv13: After closer inspection with a magnifier, the anti-aliasing isn't at 100% with the compositing. This can be fixed by removing 0.5px from the shape. 

Bottom right:
slider10


Top left:
slider10

psychon:
I kind-of figured out the problem that I was running into and reported it upstream: https://bugs.freedesktop.org/show_bug.cgi?id=98499
My first does not set a clip shape while Elv13's code painted a cairo surface completely white and set it as the clip shape. My code should be "better", but seems to run into a bug in Xephyr.

lib/wibox/init.lua
psychon: > It's a little too much to ask given my own version didn't bothered, but can you make it work with clients too (with or without turning them into full systray like drawable widgets)?

How could you sanely do anti-aliasing here? We don't draw the client's content and so cannot anti-alias it against the shape. Doing tricks with transparent pixels if a compositing manager is running won't help here either, because the client content is not "behind" our frame window. This could only ever work where there is a titlebar around a client and that's quite a limited... thing.

> After closer inspection with a magnifier, the anti-aliasing isn't at 100% with the compositing. 

I pass. I don't even intend to add that anti-aliasing hack to this PR. As shown in the screenshot with the "big cycle" with the "big border", this cannot always work. With a border, you only get anti-aliasing for the part not inside the "real X11 border". I don't like that.

actionless:
i've just tried with compton (it was running before awesome restart) and outer border is not AA:

2016-11-02--1478114302_352x271_scrot

psychon:
Yup, I think the [hack for AA at outer border](https://github.com/awesomeWM/awesome/pull/1198#discussion_r85655947) can not work properly (see end of the same comment plus [this](https://github.com/awesomeWM/awesome/pull/1198#discussion_r85657068)). I do not plan to add it to this PR.

Looking at your screenshot and adding some lines:
950d4118-a139-11e6-8de8-097168a63f34
We draw the part inside of the green line. Here we can do AA. The part between the red and green line is drawn by the X11 server. We cannot do AA here (unless we stop using the X11 window border). The original code by Elv13 does not do AA here either. 
This is the same problem as here (but here it is more easy to see due to the large border width):
bar
foo

actionless:
what about using `margin` and `background` widgets combination instead of X11 border?

however it seems like that won't be applicable to fix not-AA border of clients

actionless:
also  today i was using awesome wm for a while on a "standard" resolution display and that mixture of AA and not-AA borders looks not really sweet here

what do you think about doing AA optionally?

psychon:
I can live without simulating AA on borders that do not really support it, you and @Elv13 want it. :-)

actionless:
or instead of having it optional we can now merge it without AA but leave PR with AA version in case if someone would like to improve it

psychon:
I'll move the AA part of this PR to another PR when I find the time.

psychon:
Here are the non-AA parts.

Elv13:
IMO the outer AA is still useful even if it's limited, but anyway. I guess this can be merged as its a big step in the right direction.

Elv13:
Also, when I said it would be nice for clients, I was totally aware it would not be AAed, I was talking mainly about making the API closer to this one instead of `awful.client.shape`. Also, in theory, the titlebar inner part could be AAsed the same way as this PR do it, but it is of secondary importance.

actionless:
i can't understand the usecase when inner border is AA and outer is not-AA

or you have some ideas how to get outer border AA-ed with compositing?

Elv13:
@psychon There is a more recent version of this, right? This version seems to call `shape()` (with the issue raised in the comments)

Elv13:
@psychon There is a more recent version of this, right? This version seems to call shape() (with the issue raised in the comments)

Elv13:
ping

psychon:
No, there is no more recent version of this. The next commit was removing part of this code again and added anti-aliasing stuff. That code did things the way you remember. This first commit always looked like this.

My GitHub-mail-folder currently has about 200 unread mails and I only check things on the web page. Put differently: Currently I do not have much time to fix this up.

psychon:
I rebased this on master and added a fixup commit to address @Elv13's complaint. The result looks like this for the same test as was used previously (ignore the funny wibar, it has a border width of 10 and a rounded rectangles shape):

shapes

Elv13:
The screenshot from the last message seems to have lost the inner aliasing. That part worked fine in some previous iterations.

psychon:
> The screenshot from the last message seems to have lost the inner aliasing.

https://github.com/awesomeWM/awesome/pull/1198#issuecomment-258145061

> or instead of having it optional we can now merge it without AA but leave PR with AA version in case if someone would like to improve it

https://github.com/awesomeWM/awesome/pull/1198#issuecomment-258472497

> Here are the non-AA parts.

psychon:
And one more:

https://github.com/awesomeWM/awesome/pull/1198#issuecomment-258129577

> also today i was using awesome wm for a while on a "standard" resolution display and that mixture of AA and not-AA borders looks not really sweet here

psychon:
And in case this is ever merged, here is the additional diff that adds anti-aliasing of the clip shape:
```diff
diff --git a/lib/wibox/drawable.lua b/lib/wibox/drawable.lua
index 330edc31c..6f05605ff 100644
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -149,6 +149,13 @@ local function do_redraw(self)
         self._widget_hierarchy:draw(context, cr)
     end
 
+    -- Call a hook if we have one
+    if self._draw_hook then
+        cr:save()
+        self._draw_hook(cr)
+        cr:restore()
+    end
+
     self.drawable:refresh()
 
     assert(cr.status == "SUCCESS", "Cairo context entered error state: " .. cr.status)
@@ -352,10 +359,11 @@ local function setup_signals(_drawable)
     clone_signal("property::y")
 end
 
-function drawable.new(d, widget_context_skeleton, drawable_name)
+function drawable.new(d, widget_context_skeleton, drawable_name, draw_hook)
     local ret = object()
     ret.drawable = d
     ret._widget_context_skeleton = widget_context_skeleton
+    ret._draw_hook = draw_hook
     ret._need_complete_repaint = true
     ret._need_relayout = true
     ret._dirty_area = cairo.Region.create()
diff --git a/lib/wibox/init.lua b/lib/wibox/init.lua
index 757b2473a..5da5f2599 100644
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -14,6 +14,7 @@ local setmetatable = setmetatable
 local pairs = pairs
 local type = type
 local object = require("gears.object")
+local gcolor = require("gears.color")
 local grect =  require("gears.geometry").rectangle
 local beautiful = require("beautiful")
 local base = require("wibox.widget.base")
@@ -62,6 +63,26 @@ function wibox:find_widgets(x, y)
     return self._drawable:find_widgets(x, y)
 end
 
+-- This function is called on every redraw after the background has been drawn,
+-- but before the actual widget was drawn. We use this to draw an anti-aliased
+-- clipping shaped (the edge between the border and the background is
+-- anti-aliased)
+local function draw_hook(self, cr)
+    local shape = self._shape
+    if not shape then return end
+
+    local geo = self:geometry()
+    local bw = self.border_width
+
+    cr:translate(-bw, -bw)
+    shape(cr, geo.width + 2*bw, geo.height + 2*bw)
+
+    -- 2*, because half of them is outside of the visible area
+    cr:set_line_width(2*bw)
+    cr:set_source(gcolor(self.border_color))
+    cr:stroke()
+end
+
 function wibox:_apply_shape()
     local shape = self._shape
 
@@ -85,24 +106,11 @@ function wibox:_apply_shape()
     self.shape_bounding = img._native
     img:finish()
 
-    -- Now handle the clip shape (things excluding the border)
-    img = cairo.ImageSurface(cairo.Format.A1, geo.width, geo.height)
-    cr = cairo.Context(img)
-
-    -- We give the shape the same arguments as for the bounding shape and draw
-    -- it in its full size (the translate is to compensate for the smaller
-    -- surface)
-    cr:translate(-bw, -bw)
-    shape(cr, geo.width + 2*bw, geo.height + 2*bw)
-    cr:set_operator(cairo.Operator.SOURCE)
-    cr:fill_preserve()
-    -- Now we remove an area of width 'bw' again around the shape (We use 2*bw
-    -- since half of that is on the outside and only half on the inside)
-    cr:set_source_rgba(0, 0, 0, 0)
-    cr:set_line_width(2*bw)
-    cr:stroke()
-    self.shape_clip = img._native
-    img:finish()
+    -- Now handle the clip shape (things excluding the border).
+    -- We don't actually set a shape here, but instead provide anti-aliasing via
+    -- our draw_hook().
+    self._drawable._do_complete_repaint()
+    self.shape_clip = nil
 end
 
 --- Set the wibox shape.
@@ -222,7 +230,8 @@ local function new(args)
 
     ret.drawin = w
     ret._drawable = wibox.drawable(w.drawable, { wibox = ret },
-        "wibox drawable (" .. object.modulename(3) .. ")")
+        "wibox drawable (" .. object.modulename(3) .. ")",
+        function(cr) draw_hook(ret, cr) end)
 
     ret._drawable:_inform_visible(w.visible)
     w:connect_signal("property::visible", function()
@@ -256,6 +265,7 @@ local function new(args)
 
     ret:connect_signal("property::geometry", ret._apply_shape)
     ret:connect_signal("property::border_width", ret._apply_shape)
+    ret:connect_signal("property::border_color", ret._drawable._do_complete_repaint)
 
     -- If a value is not found, look in the drawin
     setmetatable(ret, {
```
Anyone is free to pick this up and do something with it. I will not pursue this further.

Elv13:
Would you accept to merge a PR with the above diff applied with an additional wibox property (disabled by default)?. Without AA, it does look ugly with the infobubble shape.

 * The outer AA works fine as long as you don't use a border
 * The inner AA always works fine
 * The outer AA works almost fine with 1 or 2px borders and the "rounded rect" shape. It's only ugly with larger borders and shapes like the circle.

If you agree to that, then this PR can be merged. So far I prefer the "nothing at all" solution over the non-AA one because if this was merged as it, then AA support will never be merged.

psychon:
> Would you accept to merge a PR with the above diff applied 

You wouldn't accept that, I think: https://github.com/awesomeWM/awesome/pull/1198#discussion_r85657068

Besides that, I am not really sure. Trying to come up with some magic to toggle between with-aa and without-aa seems weird. 

> * The outer AA works fine as long as you don't use a border

...and you use a compositing manager and you never stop it (the code doesn't handle that case). Plus, no it does not: https://github.com/awesomeWM/awesome/pull/1198#discussion_r85657068

> * The inner AA always works fine

...but looks weird when there is inner AA, but no outer AA, according to actionless

> * The outer AA works almost fine with 1 or 2px borders and the "rounded rect" shape. It's only ugly with larger borders and shapes like the circle.

Make a screenshot and zoom in. I'm 5% sure that one can see the missing AA. It's just that one does not notice the ugliness normally.

> So far I prefer the "nothing at all" solution over the non-AA one because if this was merged as it, then AA support will never be merged.

Fine with me. I only created this PR as a response to #1188. My version avoids some of the black pixels from that PR.

actionless:
i think we still have to choose one of the solutions to have that functionality working

and since it seems to be not breaking anything more it should be a pretty safe decision to make

psychon:
Which solution would you prefer? Elv13 doesn't like non-antialised shapes and I don't like "sometimes-antialiased" shapes.

actionless:
from my point of view, we should have either both outer and inner sides of the border anti-aliased either both not-AA

Elv13:
What about adding an `inner_border_width` property and only allow AA when border_width == 0? It isn't pretty, but sidestep the issue. Tooltips and notification could use client side borders and have their shape. This is what the tooltip code already do, so it would simplify existing code. There cannot be "border modes" as the geometry is calculated differently.

psychon:
So, let's merge this?

Elv13:
> So, let's merge this?

Not so fast ;). First, what do you think about having both "native" and "soft" border concepts alongside each other? If it's implemented (yes, I will do it, unless you want to do it yourself or feel that you would rewrite it anyway), will it be merged and will you allow the AA code path to be turned on for it? Also, once this is done, would you accept the patch that also adds it for the clients? As you said many times, the X11 border is limited to a single solid color, no transparency*, rectangle shapes** and no AA. Also, I think Awesome is almost the last WM to use it. Finally, should both border types be mutually exclusive to avoid confusion***

\* You once said that one could be fixed
\*\* As in the sense the non-rectangle section isn't part of the border from X11 PoV (as demonstrated above) (and yes, I know about the `clip` mask, but it isn't the same as the border clip)
\*\*\* If both are used, then outer AA would break and we would be back to square one.

psychon:
I assume "native borders" are what we have now. Are "soft" borders a wibox with `border_width = 0` containing a widget that draws some borders internally? And you would want to add more magic ontop of that so that it is still controlled by the `border_width` property?

So far this sounds orthogonal to outer AA. Sure, sounds doable (except that the resulting code will be messy, but sigh....)

> will you allow the AA code path to be turned on for it? 

So this question is about my earlier patch for doing kind-of-outer-AA for `border_width = 0` windows? Fine with me.
But let me point out that is only half-AA: Let us assume we have values from 0 (transparent) to 255 (opaque). Without AA you just get these two values. With proper AA, anything in between can appear. However, here we "cut off" the window shape at 127 and so only values 0 and anything between 127 and 255 can appear. So, it's not proper AA and will never be.
If you are fine with that (and won't try to e.g. shrink/enlarge windows by one pixel to overcome that), then that's also fine with me.

Elv13:
> shrink/enlarge windows by one pixel to overcome that), then that's also fine with me.

My version of this PR did that, but I agree it's the ugliest part of it. I am fine with 0-127, its better than nothing.

> And you would want to add more magic ontop of that so that it is still controlled by the border_width property?

Or not. It can be another property like `inner_border_width`. The use cases here are menu/notifications/tooltips/wibars (and someday clients/titlebars). It doesn't have to be that nice API-wise, it just has to work.

psychon:
Half-fine with me. I just took a look at Gtk and Qt. They provide only "pixel-perfect" window shapes. However, the examples that I found don't bother with that. They just set a transparent background and draw ontop of that. I did not run any of the examples and so I do not know what they do without a compositing manager...
What do you think about following the paths of these toolkits?

(I'll start working on input shape support so that the transparent part can at least be made click-through)

Elv13:
Things can never be simple... As much as I would like to say we could do that, only about 10% (?) of users run a compositor... So we still need the `shape_bounding`. It could only be enabled when `awesome.compositor_manager_running == false` and we can ignore the corner case where users start and stop compositing. This would allow to have pixel perfect shapes for users with a compositor and only inner AA for users without and no AA at all with `border_width > 0`. That's fine by me.

This issue can only be resolved with compromises on both side. There is no perfect solution or we would never had all these back and forth arguments over 3 PR and 5 issues.

psychon:
@Elv13 For you to play with: https://gist.github.com/psychon/551eea50ad2133e057bc88d91f1ea07c https://gist.github.com/psychon/972c5fc0dba2b052f894a1bf084b7bd7 (the blinking is to work around an X server issue: https://bugs.freedesktop.org/show_bug.cgi?id=98499 )

actionless:
wow, i can't believe that finally happened 🎆 🍾

actionless:
now it's time to merge this one:  https://github.com/awesomeWM/awesome/pull/1189

which is depends on this PR

Issue #1176, opened 2016-10-19
Switch to xcb-util-xrm


psychon:
We can replace our use of Xlib with https://github.com/Airblader/xcb-util-xrm. That should get rid of quite some weird code.

PR #1202, opened 2016-10-30
Use xcb-util-xrm

psychon:
Instead of using Xlib for parsing resource files, this now uses the
dedicated xcb-based library that is meant for exactly this task.

Fixes: https://github.com/awesomeWM/awesome/issues/1176

CC @AirBlader just for your information

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,10 @@ install:
   # Deps for functional tests.
   - sudo apt-get install -y dbus-x11 xterm xdotool xterm xvfb
 
+  # Need xorg-macros
+  - sudo apt-get install -y xutils-dev
+  - git clone --recursive https://github.com/Airblader/xcb-util-xrm.git && cd xcb-util-xrm && ./autogen.sh --prefix=/usr && make && sudo make install && cd -
+
   # Install Lua (per env).
   # Note that Lua 5.3 is installed manually, because it is not available in Ubuntu Trusty.
   - |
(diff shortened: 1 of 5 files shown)


Airblader:
Completely forgot about putting this into awesome. Glad to see it happen now. :-)

psychon:
In that case I won't mention cairo ;-)

Airblader:
If only I knew someone working on cairo... :-P Perhaps I'll find the time at some point to clone cairo and get that done as well. :-)

actionless:
works for me

blueyed:
The package is quite new (https://packages.debian.org/search?searchon=sourcenames&keywords=xcb-util-xrm, https://launchpad.net/ubuntu/+source/xcb-util-xrm), so maybe we should keep the old custom code with `#ifdef`s?!

Airblader:
i3 and rofi have been using it for a while now. I'm happy to say that not one bug has since been found in the library. Most distros should also have added the package by now.

actionless:
i think if for somebody it will be a showstopper that person will create issue/PR on that problem and we can return to it then

blueyed:
I see.
It came up on IRC, that's why I've wanted to mention it here.
And while most distros include it, it will not be available in LTS ones etc.
(sorry to be annoying.. ;))

Airblader:
> And while most distros include it, it will not be available in LTS ones etc

Neither will this version of awesome be. And if they use a PPA they can use one for the library.

PR #1204, opened 2016-11-01
Improve DPI per screen setting

blueyed:


lib/beautiful/xresources.lua
```diff
@@ -101,13 +101,15 @@ end
...
 function xresources.set_dpi(dpi, s)
-    s = get_screen(s)
     if not s then
         xresources.dpi = dpi
     else
-        dpi_per_screen[s] = dpi
+        s = get_screen(s)
+        if s then
+            dpi_per_screen[s] = dpi
+        end
```
psychon: Documentation-wise: Sure, thanks for catching this.
However, the code-change shouldn't make any real difference, should it? (Well, you can no longer use e.g. "primary" or "HDMI-1" as arguments, but that's fine with me)
Only thing I see is that using e.g. a function as argument no longer throws an error.

What kind of invalid arguments were you thinking of here? Should that be mentioned in the commit message?

lib/beautiful/xresources.lua
```diff
@@ -70,13 +70,21 @@ local function get_screen(s)
...
---- Get global or per-screen DPI value falling back to xrdb.
--- @tparam[opt] integer|screen s The screen.
+--- Get global or per-screen DPI value, falling back to xrdb.
+-- @screen[opt] s The screen.
 -- @treturn number DPI value.
 function xresources.get_dpi(s)
-    s = get_screen(s)
-    if dpi_per_screen[s] then
-        return dpi_per_screen[s]
+    if s then
```
psychon: Uhm, this means `nil` is now a valid argument? Why? I know that this might make things easier for some callers, but doesn't that also mean it becomes more easy to break per-screen-DPI?

lib/beautiful/xresources.lua
```diff
@@ -70,13 +70,21 @@ local function get_screen(s)
...
+        s = get_screen(s)
+        if dpi_per_screen[s] then
+            return dpi_per_screen[s]
+        end
+        local width = 0
+        for _,v in pairs(s.outputs) do
+            width = width + v.mm_width
+        end
+        local inch = width / 25.4
+        return math.floor(s.workarea.width / inch)
```
psychon: This now ignores `Xft.dpi` even though the documentation says otherwise. Also, it's wrong. Adding the width of outputs is wrong.

Imagine clone mode: Multiple monitors (= outputs) display the same part of the screen. This is the case where a screen has multiple outputs. Of course adding their width together makes no sense here. The only possible thing might be to pick one of them arbitrarily, and that's bad ("arbitrary").

(Oh and the 25.4 magic number might need an explanation)

actionless: `25.4` looks to me like conversion from pixels to inch with hardcoded `100` dpi

blueyed: Well, a inch is 2.54cm, therefore the mm from `mm_width` get divided by `25.4` and you get inches.
I thought the variable name is good documentation already, but apparently it needs a comment like `25.4mm = 1inch`.

lib/beautiful/xresources.lua
blueyed: That part should be reverted/kept then probably.
IIRC I wanted to distinguish between the argument and the return value from `get_screen` in some way.

psychon: Well, previously calling with `nil` means that `get_screen` returned `nil` and then `dpi_per_screen[nil]` would cause a Lua error "table index is nil".
Why do you want to distinguish between the argument and the return value?

blueyed:
Might be irrelevant after https://github.com/awesomeWM/awesome/pull/2055 - or at least needs adjustments.
Closing for now.

PR #1206, opened 2016-11-04
Implement some fallbacks if XKB is unavailable

psychon:


actionless:
just as a side note, xlib is still used here: https://github.com/awesomeWM/awesome/blob/master/objects/key.c#L42

psychon:
> just as a side note, xlib is still used here: https://github.com/awesomeWM/awesome/blob/master/objects/key.c#L42

Sigh.

PR #1207, opened 2016-11-05
Blacklist broken ldoc versions

psychon:
Since apparently none of the affected persons wanted to test my patch from https://github.com/awesomeWM/awesome/issues/1098#issuecomment-250851514, here it is now as a PR. I will test it via additional commits that are then run via Travis.

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -42,6 +42,14 @@ a_find_program(LDOC_EXECUTABLE ldoc FALSE)
 if(NOT LDOC_EXECUTABLE)
     a_find_program(LDOC_EXECUTABLE ldoc.lua FALSE)
 endif()
+if(LDOC_EXECUTABLE)
+    execute_process(COMMAND sh -c "${LDOC_EXECUTABLE} --sadly-ldoc-has-no-version-option 2>&1  | grep ' vs 1.4.5'"
+                    OUTPUT_VARIABLE LDOC_VERSION_RESULT)
+    if(NOT LDOC_VERSION_RESULT STREQUAL "")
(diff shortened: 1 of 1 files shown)


psychon:
Wow.
LDoc 1.4.5 rejected: https://github.com/awesomeWM/awesome/pull/1207/commits/734d47d7f3612efb2e58e691e2040ac1c51a3b81 https://travis-ci.org/awesomeWM/awesome/builds/173454219
During test builds, LDoc 1.4.6 also turned out to be apparently broken. However, it identifies itself as version 1.4.5 (I checked the source download, it really does!) and thus is also rejected by this patch: https://github.com/awesomeWM/awesome/pull/1207/commits/a7a53d3143ddea1d7c1c25028b59f3e20c0b10cd https://travis-ci.org/awesomeWM/awesome/builds/173454259

I'll now get rid of those two extra commits and that will hopefully show that LDoc 1.4.4 is correctly accepted.

actionless:
last patch was working fine until ldoc updated again (or something else changed in my setup), so i wanted to clarify the problem first before commenting on it

this patch fixes the build itself, but ldoc is not running at all so no docs generated. i think it could be better just to ignore ldoc exit code if we are not in travis

blueyed:
> During test builds, LDoc 1.4.6 also turned out to be apparently broken

So is there still an issue?
Is it the version from current Git master?

psychon:
>>During test builds, LDoc 1.4.6 also turned out to be apparently broken
>So is there still an issue?

Nope. LDoc 1.4.6 (which is available on luarocks) identifies itself as 1.4.5 in its help output. Thus, this patch ignores that one out.

> Is it the version from current Git master?

I don't think so. I didn't explicitly test current git master.

blueyed:
Thanks!

PR #1215, opened 2016-11-09
Minor doc and code style fixes

blueyed:


lib/awful/client.lua
```diff
@@ -900,19 +900,18 @@ function client.setwfact(wfact, c)
     t:emit_signal("property::windowfact")
 end
 
---- Increment a client's window factor
+--- Change a client's window factor.
 --
 -- @legacylayout awful.client.incwfact
--- @param add amount to increase the client's window
+-- @tparam number add Amount to increase/decrease the client's window.
+--   Should be between 0.0 and 1.0.
```
actionless: how can it decrease if it can't be less than zero?

psychon: The range is [-current_window_factor, infinity]. You don't really want this in the docs, I guess.

The code in this function does the increment and then does normalize(). This function gets all window factors back in the range [0,1].

If the argument to `incwfact` is `-current_factor`, the window ends up with factor zero (and things will break when the layout actually tries to assign the window a size of zero).
If the argument is something huge, all other windows will end up with a factor of zero (and the window we are called on ends up with value 1).
In between these extremes, sane things happen. Further outside of these extremes, breakage ensures (passing in a NaN will permanently  break window factors).

psychon:
:+1: if `incwfact`'s docs get fixed somehow.

lib/awful/client.lua
blueyed: @psychon 
Thanks for explaining it.  After all, I consider this pattern / range a bit weird.  It seems to be the cause for incrementing and then decrementing by the same value (e.g. `0.1`, using a key binding) will result in a different size than before.
Could this be fixed somehow to be more "predictable"?  /cc @Elv13

blueyed: @Elv13 
But what about making it idem-potent, so that calling it with -0.1 and 0.1 (multiple times) results in the same layout?
What is the new API for this?

lib/awful/client.lua
```diff
@@ -900,19 +900,18 @@ function client.setwfact(wfact, c)
     t:emit_signal("property::windowfact")
 end
 
---- Increment a client's window factor
+--- Change a client's window factor.
```
blueyed: "Change window factor of a client"

lib/awful/client.lua
```diff
@@ -900,11 +900,11 @@ function client.setwfact(wfact, c)
     t:emit_signal("property::windowfact")
 end
 
---- Change a client's window factor.
+--- Change window factor of a client.
 --
 -- @legacylayout awful.client.incwfact
 -- @tparam number add Amount to increase/decrease the client's window.
---   Should be between 0.0 and 1.0.
+--   Should be between -1.0 and 1.0 (all window factors get normalized to 1).
```
psychon: Isn't this still wrong? I would just drop this sentence.

As said in the other argument, the argument should be between `-current_window_factor` and `something close to infinite`. The normalisation then ensures that the sum of all factors is 1.

blueyed: Used your phrasing now.  Leaving it out is not really an option: the documentation should explain what it is about after all - which it is still not great at, but better than before.

blueyed: The whole concept of window factors is not explained well I'd say.

PR #1219, opened 2016-11-14
Dpi in default theme

### LINKED ISSUE #1214: Active tab indicator square is not scaled with DPI

```
awesome v3.5.2-2174-g588cbc1 (The Fox)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1
```

1. Download [`awesome-git`](https://aur.archlinux.org/packages/awesome-git)
1. Unpack it
1. Apply the following diff to `PKGBUILD`:

        20c20
        <          'lua' 'lua-lgi-git' 'pango' 'startup-notification' 'xcb-util-cursor'
        ---
        >          'lua' 'lua-lgi' 'pango' 'startup-notification' 'xcb-util-cursor'
        22c22
        < makedepends=('asciidoc' 'cmake' 'docbook-xsl' 'doxygen' 'imagemagick' 'ldoc-git' 'xmlto')
        ---
        > makedepends=('asciidoc' 'cmake' 'docbook-xsl' 'doxygen' 'imagemagick' 'xmlto')
        47c47,48
        <     -DLUA_LIBRARY=/usr/lib/liblua.so.5.3
        ---
        >     -DLUA_LIBRARY=/usr/lib/liblua.so.5.3 \
        >     -DGENERATE_DOC=False
1. Compile and install using `makepkg --syncdeps --install`
1. Set up your X resources:

        $ cat /etc/X11/Xresources
        Xft.dpi:        276
1. Restart awesome

**Actual result:**

The tab bar font scales up, but the tab state indicator square is still tiny: 2016-11-09-202242_810x71_scrot

**Expected result:**

The tab state indicator square should scale with DPI. Possibly it could be a glyph rather than an image?

actionless:
closes #1214


themes/default/theme.lua
```diff
@@ -2,6 +2,11 @@
 -- Default awesome theme --
 ---------------------------
 
+local beautiful = require("beautiful")
+local xresources = beautiful.xresources
+local dpi = xresources.apply_dpi
+local theme_assets = beautiful.theme_assets
```
psychon: Isn't this another cyclic dependency waiting to become a problem? It's no problem right now, but can become one in the future (beautiful depends on theme which depends on beautiful).

I'd propose not to `require("beautiful")` here, but directly the needed parts.

themes/sky/theme.lua
```diff
@@ -47,15 +53,22 @@ theme.awesome_icon           = "@AWESOME_THEMES_PATH@/sky/awesome-icon.png"
...
+    taglist_square_size, theme.fg_normal
+)
 
 -- MISC
 theme.wallpaper             = "@AWESOME_THEMES_PATH@/sky/sky-background.png"
 theme.taglist_squares       = "true"
 theme.titlebar_close_button = "true"
-theme.menu_height           = 15
-theme.menu_width            = 100
+theme.menu_height = dpi(16)
```
psychon: Is the change 15 -> 16 intended?

themes/sky/theme.lua
```diff
@@ -4,6 +4,12 @@
 -------------------------------
 -- If you want SVGs and extras, get them from garoth.com/awesome/sky-theme
 
+local beautiful = require("beautiful")
+local xresources = beautiful.xresources
+local xrdb = xresources.get_current_theme()
```
psychon: Why xrdb?

themes/sky/theme.lua
actionless: nope, by mistake

actionless:
@psychon fixed

actionless:
done

psychon:
Now LDoc is unhappy:
```
/home/travis/build/awesomeWM/awesome/.build-testing-gce-702733d5-ff7b-4572-b0b3-9bf87a234d0d-x86_64-linux-gnu-4.8/lib/beautiful/theme_assets.lua:1: contains no items
```

actionless:
any ideas how this travis failure can be related to these changes?
https://travis-ci.org/awesomeWM/awesome/jobs/185041642

Elv13:
"awesome" is normally a global variable. If it isn't set, then something overwrote it with `nil`, but nothing in your patchset does that. Did you rebase this correctly on top of the master branch?

psychon:
`beautiful` now requires `beautiful.theme_assets` which it previously did not do. `theme_assets` require()s `gears` and so `gears.timer` is loaded while it previously was not loaded. The unit tests run by `busted` are supposed to be *unit*  tests and so do not mock the whole API. Specifically, there is no (and never was) `awesome` global available.

Simple work around would be to change `assets.lua` to `require("gears.color")` instead of all of `gears`. It doesn't use anything but `gears.color` any way.

actionless:
@psychon 
oh, right, i was fixing that already in the original commits but lost it accidentally while resolving merge conflict, i will update it later on

actionless:
updated and rebased (again :) )

let's wait for travis

blueyed:
@actionless 
Tests/CI checks every commit by now (to help with git-bisect), and currently it fails for the following two still: 3053669b07b11efc30d37973366e694502f54720 4f64a00774c054546e76dc00983a4a3c6f49dd06.
You should probably rebase it to squash the fix, or move it before those commits.

actionless:
oh, my bad, squashed

actionless:
:tada:

Elv13:
Just a little question before merging. Should them_asset be a private API instead of a public one? We can't just break its API as it would break people themes, but still.

psychon:
What would you count as "private API"? An undocumented one?
That doesn't really work since people will copy the themes we ship and do modifications, so it cannot just break.
Or are you referring to "this commit changes API" by breaking "old" versions of the `xresources` theme?

psychon:
> will have to wait for the v5 API break to be merged

I know you want to avoid a "4 years of development on master", but "collecting" pull requests until we decide its time to break the API is not going to fly either. We can either have a "v4" branch and a "master" branch and try to backport everything to the "v4" branch or we can (at least initially, might want to switch to a different strategy later) have a "v5" branch for API breaks and continue development on master (and then regularly (after releases?)) merge master into v5).

Anyway, thanks for catching this API break.

> This time I guess @actionless can add some compatibility alias to avoid breaking the existing themes. 

So, we need a file `themes/xresources/assets.lua` containing something like the following:
```lua
require("awful.util").deprecate("foo")
return require("beautiful.xresources")
```
Right?

Elv13:
> I know you want to avoid a "4 years of development on master", but "collecting" pull requests until we decide its time to break the API is not going to fly either. We can either have a "v4" branch and a "master" branch and try to backport everything to the "v4" branch or we can (at least initially, might want to switch to a different strategy later) have a "v5" branch for API breaks and continue development on master (and then regularly (after releases?)) merge master into v5).

This isn't going to work. This is what we did last time. Doing the opposite (working on master and having a v5 branch) is already a better option. But just working harder to keep the API compatible, even if it requires dummy files, FIXME and an improved deprecation framework requires *less* work than dealing with 2 branches. 95% of the breakage can either be avoided or handled transparently.

actionless:
hopefully this time everything will be fine

actionless:
@Elv13 , what do you think?

blueyed:
Thanks!

PR #1220, opened 2016-11-15
doc: Document tasklist_plain_task_name

Elv13:
It was missing.

also add a missing `-`

### DIFF
--- a/lib/awful/widget/tasklist.lua
+++ b/lib/awful/widget/tasklist.lua
@@ -1,6 +1,29 @@
 ---------------------------------------------------------------------------
 --- Tasklist widget module for awful.
 --
+-- <a name="status_icons"></a>
+-- **Status icons:**
+--
+-- By default, the tasklist prepends some symbols in front of the client name.
+-- This is used to notify that the client has some specific properties that are
+-- currently enabled. This can be disabled using
+-- `beautiful.tasklist_plain_task_name`=true in the theme.
+--
+-- <table class='widget_list' border=1>
+-- <tr style='font-weight: bold;'>
+--  <th align='center'>Icon</th>
+--  <th align='center'>Client property</th>
+-- </tr>
+-- <tr><td>▪</td><td><a href="./client.html#client.sticky">sticky</a></td></tr>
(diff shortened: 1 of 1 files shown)


lib/awful/widget/tasklist.lua
```diff
@@ -1,6 +1,27 @@
 ---------------------------------------------------------------------------
 --- Tasklist widget module for awful.
 --
+-- By default, the tasklist prepend some symbols in front of the client name.
```
psychon: preprend -> prepends

lib/awful/widget/tasklist.lua
```diff
@@ -1,6 +1,27 @@
...
+-- This is used to notify that the client has some specific properties are
+-- currently enabled. This can be disabled using
```
psychon: properties are -> properties that are

lib/awful/widget/tasklist.lua
```diff
@@ -83,10 +104,17 @@ local instances
 -- @beautiful beautiful.tasklist_bg_image_minimize
 -- @tparam[opt=nil] string bg_image_minimize
 
--- Disable the tasklist clients icons.
+--- Disable the tasklist clients icons.
```
blueyed: s/clients/client/

lib/awful/widget/tasklist.lua
```diff
@@ -83,10 +104,17 @@ local instances
...
 -- @beautiful beautiful.tasklist_tasklist_disable_icon
 -- @tparam[opt=false] boolean tasklist_disable_icon
 
+--- Disable the extra tasklist client property notification icons.
+--
+-- See <a href="#client_icons">the top of this page</a> for more details.
```
blueyed: Try to rename the link target, i.e. "Client Icons".

psychon: I don't think spaces work well in anchors, do they?

blueyed: Sorry, I've meant instead of "the top of this page", so the link text, not `href`.  (but you could use `%20` there to encode spaces)
But feel free to ignore that: it is not as bad as the typical "here", which is unspecific.

Elv13:
done

PR #1221, opened 2016-11-15
doc: improve client class

blueyed:


psychon:
Why didn't codecov run for this? (Not that it matters much....)

PR #1222, opened 2016-11-16
Use unmodified command for the command history.

morethanoneanimal:
Closes awesomeWM/awesome#1104.

### DIFF
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -389,9 +389,9 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
         cursor_pos = cur_pos, cursor_ul = cur_ul, selectall = selectall,
         prompt = prettyprompt })
 
-    local function exec(cb)
+    local function exec(cb, command_to_history)
         textbox:set_markup("")
-        history_add(history_path, command)
+        history_add(history_path, command_to_history)
         keygrabber.stop(grabber)
         cb(command)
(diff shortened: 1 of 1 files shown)


Elv13:
Thanks a lot. I guess this could use a test, but then again, this file isn't well tested yet, so it might be a little too much to ask given there is some internal data that will need to be exposed. In turn, that will require and API talk and so on.

I will wait to see what @blueyed think about it. But really, thanks for attacking the blocking bugs list, this will make 4.0 so much earlier.

lib/awful/prompt.lua
```diff
@@ -461,17 +461,18 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
                     for _,v2 in ipairs(v[1]) do
                         match = match and mod[v2]
                     end
-                    if match or #filtered_modifiers == 0 then
+                    if match then
```
psychon: Is the `or`-part removed on purpose? If so, I don't see why.

The rest looks good to me.

morethanoneanimal: On purpose, it's an impossible condition. If `#filtered_modifiers == 0` then `#v[1] == 0` and we don't enter the previous loop, therefore `local match` remains true.

psychon: Ok, convinced. Would be nicer in a separate commit, but whatever.

Issue #1144, opened 2016-10-04
doc: gears.color entrypoint?!


blueyed:
In e.g. https://github.com/awesomeWM/awesome/blob/26d3c5eb1aa3a9f5930dcae609580649f62289ad/lib/wibox/widget/progressbar.lua#L39 `@tparam gears.color` is used, but gets not resolved.

There is no single `color` type that can be used as reference, and apparently ldoc does not link to modules in this case.

Can/should we add some base type in `gears.color` that can be used (not only) for documentation?

psychon:
Not having much clue about ldoc, I have to ask: What is / counts as a base type? How would such a change look like?
Having links in these types would surely be useful.

blueyed:
@Elv13 
Any idea?

Elv13:
> Any idea

I am not sure this is a big issue.

First, it needs something like:

```lua
--- Convert something to a color.
-- @tparam table|string|pattern arg Something that can be converted to a color
-- @function gears.color
```

The "single type" of `gears.color` is a cairo pattern, but this user isn't aware of that, and should not be. Most (all?) APIs just take the unconverted values and pass them to `gears.color`. There isn't much to "do" (from an API doc PoV) about this, right? Sure, having confusing types all over the place need some cleanup. Maybe a `string|pattern|table` + "(See \`gears.color\`)" in the argument description, but there _is_ no types involved here.

blueyed:
I am more interested in linking the user to a good entrypoint about what `gears.color` is: currently it's just not being linked, but could go to the module for example at least.

Elv13:
A `@see` or "(See `gears.color`)" is about as close as you will get without patching ldoc (unless I missed something). And if we get into that, this probably isn't the top thing to patch. The ".manage" signal type issue is the worst, the properties are also weird (we want `value`, not `wibox.widget.progressbar.value`). It's the same limitation that causes both.

PR #1227, opened 2016-11-21
gears.color refactoring

morethanoneanimal:
I allowed myself to refactor `parse_color` function to more readable form which fails for invalid inputs.
Closes awesomeWM/awesome#1144.

### DIFF
--- a/lib/gears/color.lua
+++ b/lib/gears/color.lua
@@ -22,51 +22,50 @@ local pattern_cache
 
 --- Parse a HTML-color.
 -- This function can parse colors like `#rrggbb` and `#rrggbbaa` and also `red`.
--- Thanks to #lua for this. :)
+-- Max 4 chars per channel. Thanks to #lua for this. :)
 --
 -- @param col The color to parse
--- @return 4 values which each are in the range [0, 1].
+-- @return 4 values representing color in RGBA format (each of them in [0, 1]
+-- range) or nil if input is incorrect.
 -- @usage -- This will return 0, 1, 0, 1
 -- gears.color.parse_color("#00ff00ff")
 function color.parse_color(col)
     local rgb = {}
-    -- Is it a HTML-style color?
     if string.match(col, "^#%x+$") then
-        -- Get all hex chars
-        for char in string.gmatch(col, "[^#]") do
-            table.insert(rgb, tonumber(char, 16) / 0xf)
+        local hex_str = col:sub(2, #col)
+        local channels
+        if #hex_str % 3 == 0 then
+            channels = 3
+        elseif #hex_str % 4 == 0 then
+            channels = 4
+        else
+            return nil
         end
-        -- Merge consecutive values until we have at most four groups (rgba)
-        local factor = 0xf
-        while #rgb > 4 do
-            local merged = {}
-            local key, value = next(rgb, nil)
-            local next_factor = (factor + 1)*(factor + 1) - 1
-            while key do
-                local key2, value2 = next(rgb, key)
-                local v1, v2 = value * factor, value2 * factor
-                local new = v1 * (factor + 1) + v2
-                table.insert(merged, new / next_factor)
-                key, value = next(rgb, key2)
-            end
-            rgb = merged
-            factor = next_factor
+        local chars_per_channel = #hex_str / channels
+        if chars_per_channel > 4 then
+            return nil
+        end
+        local dividor = 0x10^chars_per_channel - 1
+        for idx=1,#hex_str,chars_per_channel do
+            local channel_val = tonumber(hex_str:sub(idx,idx+chars_per_channel-1), 16)
+            table.insert(rgb, channel_val / dividor)
+        end
+        if channels == 3 then
+            table.insert(rgb, 1)
         end
     else
-        -- Let's ask Pango for its opinion (but this doesn't support alpha!)
         local c = Pango.Color()
-        if c:parse(col) then
-            rgb = {
-                c.red / 0xffff,
-                c.green / 0xffff,
-                c.blue / 0xffff,
-            }
+        if not c:parse(col) then
+            return nil
         end
+        rgb = {
+            c.red / 0xffff,
+            c.green / 0xffff,
+            c.blue / 0xffff,
+            1.0
+        }
     end
-    -- Add missing groups (missing alpha)
-    while #rgb < 4 do
-        table.insert(rgb, 1)
-    end
+    assert(#rgb == 4, col)
     return unpack(rgb)
 end
 
--- a/lib/gears/color.lua
+++ b/lib/gears/color.lua
@@ -20,13 +20,38 @@ local surface = require("gears.surface")
 local color = { mt = {} }
 local pattern_cache
 
+--- Create a pattern from a given string.
+-- This function can create solid, linear, radial and png patterns. In general,
+-- patterns are specified as strings formatted as "type:arguments". "arguments"
+-- is specific to the pattern being used. For example, one can use
+-- "radial:50,50,10:55,55,30:0,#ff0000:0.5,#00ff00:1,#0000ff".
+-- Alternatively, patterns can be specified via tables. In this case, the
+-- table's 'type' member specifies the type. For example:
+-- {
+--   type = "radial",
+--   from = { 50, 50, 10 },
+--   to = { 55, 55, 30 },
+--   stops = { { 0, "#ff0000" }, { 0.5, "#00ff00" }, { 1, "#0000ff" } }
+-- }
+-- Any argument that cannot be understood is passed to @{create_solid_pattern}.
+--
+-- Please note that you MUST NOT modify the returned pattern, for example by
+-- calling :set_matrix() on it, because this function uses a cache and your
+-- changes could thus have unintended side effects. Use @{create_pattern_uncached}
+-- if you need to modify the returned pattern.
+-- @see create_pattern_uncached, create_solid_pattern, create_png_pattern,
+--   create_linear_pattern, create_radial_pattern
+-- @tparam string col The string describing the pattern.
+-- @return a cairo pattern object
+-- @function gears.color
+
 --- Parse a HTML-color.
 -- This function can parse colors like `#rrggbb` and `#rrggbbaa` and also `red`.
--- Max 4 chars per channel. Thanks to #lua for this. :)
+-- Max 4 chars per channel.
 --
 -- @param col The color to parse
--- @return 4 values representing color in RGBA format (each of them in [0, 1]
--- range) or nil if input is incorrect.
+-- @treturn table 4 values representing color in RGBA format (each of them in
+-- [0, 1] range) or nil if input is incorrect.
 -- @usage -- This will return 0, 1, 0, 1
 -- gears.color.parse_color("#00ff00ff")
 function color.parse_color(col)
@@ -45,7 +70,7 @@ function color.parse_color(col)
         if chars_per_channel > 4 then
             return nil
         end
-        local dividor = 0x10^chars_per_channel - 1
+        local dividor = (0x10 ^ chars_per_channel) - 1
         for idx=1,#hex_str,chars_per_channel do
             local channel_val = tonumber(hex_str:sub(idx,idx+chars_per_channel-1), 16)
             table.insert(rgb, channel_val / dividor)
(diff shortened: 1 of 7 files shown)


Elv13:
thanks for that, can you add a `@function gears.color` section so the links to `@see gears.color` end up there? 99% of users will want that. Almost nobody call the methods. This `@function` block should document the allowed arguments and their syntax

Something like

```lua
--- Convert the input into a valid pattern object.
-- Valid inputs are:
--
-- * `string` FOO
-- * table (gradient) BAR
-- * patterns BAZ
--
-- @function gears.color
```

Another improvment, but this is a little much to ask, is to add examples backed by the documentation example test suits. Those are in `awesome/tests/examples/`. But @psychon will hate me for saying this because it makes the build slower until we implement some kind of servers to run them in parallel. The issue is that they write the result back into the code, so it has to be done before compilation or cmake always think all files have changed. It also messes with the code coverage. There is plenty of solutions, its just that none of them are trivial. However, from an user point of view, it makes the doc pretty and useful.

morethanoneanimal:
I can add `@function gears.color` section, but links `@see gears.color` still point to module, not this artificially defined function. In other words they still point to `doc/libraries/gears.color.html`, not to `doc/libraries/gears.color.html#gears.color`

Also, it causes some inconsistency. Right now in doc list of functions looks like this:
```
- parse_color(col)
- create_solid_pattern(col)
- create_png_pattern(col)
...
```
After function section:
```
- gears.color(col)
- parse_color(col)
- create_solid_pattern(col)
- create_png_pattern(col)
...
```
Lack of _gears_ prefix may be a little confusing. As far as I can tell there's no general rule how to document those module functions added with _setmetatable_. Probably it's good idea to standardize it along all modules.

I think the best solution here (and for other modules) would be to have this function section and have  actual implementation function (in this case `parse_color`) to point to proper description (in this case something like "same as `gears.color`").

psychon:
Besides the 86 new warnings from LuaCheck that still need to be fixed (see Travis results): 👍 
I'd additionally do the following change, but that might just be a matter of taste:
```diff
diff --git a/lib/gears/color.lua b/lib/gears/color.lua
index 0a6be97..659bd6d 100644
--- a/lib/gears/color.lua
+++ b/lib/gears/color.lua
@@ -50,25 +50,23 @@ function color.parse_color(col)
             channel_val = tonumber(hex_str:sub(idx,idx+chars_per_channel-1), 16)
             table.insert(rgb, channel_val / dividor)
         end
-        if #rgb == 3 then
+        if channels == 3 then
             table.insert(rgb, 1)
         end
     else
         local c = Pango.Color()
-        if c:parse(col) then
-            rgb = {
-                c.red / 0xffff,
-                c.green / 0xffff,
-                c.blue / 0xffff,
-                1.0
-            }
+        if not c:parse(col) then
+            return nil
         end
+        rgb = {
+            c.red / 0xffff,
+            c.green / 0xffff,
+            c.blue / 0xffff,
+            1.0
+        }
     end
-    if #rgb == 4 then
-        return unpack(rgb)
-    else
-        return nil
-    end
+    assert(#rgb == 4, col)
+    return unpack(rgb)
 end
 
 --- Find all numbers in a string
```
Thanks for writing this!

Elv13:
I would still argue that there is code that call "gears.color(foo)" and `gears.color` is not (in the doc) a function. I agree it would be better is `ldoc` handled that itself, but given it doesn't, I prefer having `@function gears.color`.

All of `wibox` modules already do that (because otherwise it would be "impossible" to create widgets given their constructors are local functions)

psychon:
Seems like some of the new warnings from luacheck are due to a new release of luacheck. Sorry that I previously thought that it was related to this PR.

morethanoneanimal:
Alright, I'll do documentation like `wibox` does.

Some of the warnings are caused by this PR, I forgot to use local variables. I'll update it soon.

blueyed:
btw: I am fixing the whitespace in https://github.com/awesomeWM/awesome/pull/1229.

morethanoneanimal:
Oh damn it, I mess something up with rebasing...

Elv13:
I am sure you know this, but there is a million way to fix that, but the easiest is `git fetch origin; git rebase -i origin/master;git push --force morethanoneanimal/gears-color-refactoring`. `git pull` causes issues like this.

morethanoneanimal:
I didn't know that, so thanks a lot!

lib/gears/color.lua
```diff
@@ -29,6 +29,9 @@ local pattern_cache
 -- range) or nil if input is incorrect.
 -- @usage -- This will return 0, 1, 0, 1
 -- gears.color.parse_color("#00ff00ff")
+-- @function gears.color
+
+--- Parse a HTML-color, same as gears.color
 function color.parse_color(col)
```
psychon: Uhm, calling `gears.color("foo")` goes to `gears.color.create_pattern`, not `gears.color.parse_color`. Or do I mis-understand this?

lib/gears/color.lua
```diff
@@ -53,25 +53,23 @@ function color.parse_color(col)
...
+        }
     end
-    if #rgb == 4 then
-        return unpack(rgb)
-    else
-        return nil
-    end
+    assert(#rgb == 4, col)
+    return unpack(rgb)
 end
```
psychon: Thanks! :-)

psychon:
👍 if everyone else says the "same as gears.color"-comment is fine and I'm just misunderstanding this.

I propose we squash some of the commits when merging this (result would be one that adds all the `@see` and another one that cleans up `parse_color`).

lib/gears/color.lua
morethanoneanimal: Yep, you're right, I gonna fix.

lib/gears/color.lua
```diff
@@ -205,15 +213,8 @@ color.types = {
...
--- `color.create_pattern`.  The difference between `color.create_pattern`
--- and this function is that this function does not insert the generated
--- objects into the pattern cache. Thus, you are allowed to modify the
--- returned object.
--- @see create_pattern
--- @param col The string describing the pattern.
--- @return a cairo pattern object
+--- Create a pattern from a given string, same as gears.color.
+-- @see gears.color
 function color.create_pattern_uncached(col)
```
morethanoneanimal: I'm gonna nail it on my third attempt!

Elv13:
Thanks for this. Please squash your PR into a relevant changeset.

Another git note, just in case. In `git rebase -i origin/master`, move the rows in the relevant order and replace the "pick" prefix with "f" (for fixup). Git will take care of rewriting and change set into something more useful. We favor small commits exept when they add something that only makes sense as a whole. As an example, if you were to add a new widget, that would be 1 commit, because smaller ones are irrelevant for tracking regressions. On the other hand, if you fix multiple issues, such as "there is missing gears.color links", "there is a bug", "better test gears.color" and "gears.color isn't documented well enough", then multiple commits are welcome. Tests commit can be part of the fix commit or independent. I usually prefer independent as it could be reverted without reverting the test, but this is just my personal opinion.

psychon:
(And if you are thinking "This is complicated": Just say so and one of us will squash the commits.)

lib/gears/color.lua
```diff
@@ -20,53 +20,73 @@ local surface = require("gears.surface")
 local color = { mt = {} }
 local pattern_cache
 
+--- Create a pattern from a given string.
+-- This function can create solid, linear, radial and png patterns. In general,
+-- patterns are specified as strings formatted as"type:arguments". "arguments"
```
blueyed: s/as"/as "/

lib/gears/color.lua
```diff
@@ -20,53 +20,73 @@ local surface = require("gears.surface")
...
+-- is specific to the pattern used. For example, one can use
```
blueyed: s/pattern used/pattern being used/ ?

lib/gears/color.lua
```diff
@@ -20,53 +20,73 @@ local surface = require("gears.surface")
...
+-- "radial:50,50,10:55,55,30:0,#ff0000:0.5,#00ff00:1,#0000ff"
```
blueyed: Dot at the end, please.

lib/gears/color.lua
```diff
@@ -20,53 +20,73 @@ local surface = require("gears.surface")
...
+-- Alternatively, patterns can be specified via tables. In this case, the
+-- table's 'type' member specifies the type. For example:
+-- { type = "radial", from = { 50, 50, 10 }, to = { 55, 55, 30 },
+--   stops = { { 0, "#ff0000" }, { 0.5, "#00ff00" }, { 1, "#0000ff" } } }
```
blueyed: Maybe have all the entries on separate line here?
```
type = "radial",
from = { 50, 50, 10 },
to = { 55, 55, 30 },
…
```

lib/gears/color.lua
```diff
@@ -20,53 +20,73 @@ local surface = require("gears.surface")
...
+-- Any argument that cannot be understood is passed to @{create_solid_pattern}.
+--
+-- Please note that you MUST NOT modify the returned pattern, for example by
+-- calling :set_matrix() on it, because this function uses a cache and your
+-- changes could thus have unintended side effects. Use @{create_pattern_uncached}
+-- if you need to modify the returned pattern.
+-- @see create_pattern_uncached, create_solid_pattern, create_png_pattern,
+--   create_linear_pattern, create_radial_pattern
+-- @param col The string describing the pattern.
```
blueyed: We could use `@tparam string col The string describing the pattern.` here.

lib/gears/color.lua
```diff
@@ -20,53 +20,73 @@ local surface = require("gears.surface")
...
+-- @return a cairo pattern object
+-- @function gears.color
+
 --- Parse a HTML-color.
 -- This function can parse colors like `#rrggbb` and `#rrggbbaa` and also `red`.
--- Thanks to #lua for this. :)
+-- Max 4 chars per channel. Thanks to #lua for this. :)
```
blueyed: Let's drop the "Thanks to #lua for this. :)".

lib/gears/color.lua
```diff
@@ -20,53 +20,73 @@ local surface = require("gears.surface")
...
 --
 -- @param col The color to parse
--- @return 4 values which each are in the range [0, 1].
+-- @return 4 values representing color in RGBA format (each of them in [0, 1]
+-- range) or nil if input is incorrect.
```
blueyed: I prefer two spaces for indentation of continuation lines.
`@treturn` would be nice here, too - but feel free to skip it.

lib/gears/color.lua
```diff
@@ -20,53 +20,73 @@ local surface = require("gears.surface")
...
-                table.insert(merged, new / next_factor)
-                key, value = next(rgb, key2)
-            end
-            rgb = merged
-            factor = next_factor
+        local chars_per_channel = #hex_str / channels
+        if chars_per_channel > 4 then
+            return nil
+        end
+        local dividor = 0x10^chars_per_channel - 1
```
blueyed: Space around `^`? (and also brackets)

lib/gears/color.lua
```diff
@@ -234,27 +254,9 @@ function color.create_pattern_uncached(col)
...
---
--- Please note that you MUST NOT modify the returned pattern, for example by
--- calling :set_matrix() on it, because this function uses a cache and your
--- changes could thus have unintended side effects. Use @{create_pattern_uncached}
--- if you need to modify the returned pattern.
--- @see create_pattern_uncached, create_solid_pattern, create_png_pattern,
---   create_linear_pattern, create_radial_pattern
--- @param col The string describing the pattern.
--- @return a cairo pattern object
+--- Create a pattern from a given string., same as gears.color
```
blueyed: s/string.,/string/
Punctuation/dot at the end, please.  (move the dot ;)
You should also refer to `gears.color` (not only by `@see` - but also add backticks in the doc).

psychon:
@blueyed How many of your comments are unrelated to this PR? It's a bit unfair to force someone to fix mistakes that were already there before. (Also: I'd say this should end up as two commits, one adding lots of `@see` and one changing the code in `gears.surface`)

morethanoneanimal:
No problem, I can fix those mistakes along the way and squash commits, but no earlier than on Monday.

morethanoneanimal:
Updated and squashed.

PR #1229, opened 2016-11-21
Fix whitespace warnings reported by luacheck

blueyed:


Elv13:
The "test examples" whitespaces are "necessary" because CMake otherwise remove empty lines. With this, the doc isn't generated "correctly". It can be fixed, but given the lack of real problems caused by such whitespace, I would suggest to fix the non test example ones then disable the check.

blueyed:
@Elv13 
Shit.  And sorry for merging it all by myself so quickly.

> With this, the doc isn't generated "correctly"

Can you provide links to before/after?  I would look at the diff output from @awesome-robot, but that it probably still kind of useless.

> if the test example is rewritten as a daemon instead of a pre-processing target 

I would really appreciate this.. it is really bad that all the examples are processed when using `make` from the top-level dir after all.

Do we have to revert something from here for now?  How bad is the issue with not "correctly" generated docs?

Elv13:
http://new.awesomewm.org/apidoc/libraries/gears.shape.html all the new lines are gone (and from all other files)

Btw, the reason why the test example were done that way was because once upon a time, there was some other mandatory macros such as the theme path and version string in each lua and C files (remement the `.in` thing?). Those were inserted using `configure_file`. The test example had to be inserted at the same time as those variables/macros because it is an all or nothing operation. Changing those files later would tag them as "dirty" and force an eternal loop of full recompilation.

Now that most of those are gone, it isn't really necessary to do it this way. We could just make a copy of all C and Lua files after the compilation is completed, modify the copy and running ldoc on the copies. The tests themselves could be regular cmake targets or a daemon. This would still mess the code coverage, so that Travis target would still have to run the test examples first.

blueyed:
> all the new lines are gone

Could this be instead marked as a kind of code block (probably not).
But what about replacing newlines with ` ` (a single space) during the pre-processing?

blueyed:
@Elv13 
Should we revert this for now, and rather pin luacheck?

Elv13:
I tried

```lua
files = {
   ["tests/example"] = {
        ignore = {"^%s*\n", "^[ ]+$", "line contains only whitespace"}
    }
}

```

But `luacheck` doesn't seem to want to match anything about this particular check.

psychon:
> Of course, if the test example is rewritten as a daemon instead of a pre-processing target then this kind of awkward limitations would go away (and it could be much faster). But this is a rather large project if it has to restore feature parity with the current solution, so disabling the check is an easier fix for now.

This is a bit of-topic, but I'd rather propose to have the test source in the actual source code (instead of in extra files under `tests/examples`) and have some "magic" to extract them from the source. That would solve the code-coverage issue, because then the source code wouldn't actually be modified any more. However, this means that it is no longer possible to have some text generated by these tests (well, one could have some "expected output" in the source code and compilation would fail if this is not the actual output of the script....).

> I tried

No idea if it helps you, but according to https://github.com/mpeterv/luacheck/issues/79#issuecomment-260630188, `--ignore 6` is the right command line flag.

Elv13:
I still argue the test output is relevant and its already used in many examples. Your suggestion would also force to hardcode the image path, make the source unreadable, break syntax highlight when developing examples, make luacheck and coverage a nightmare to use (the line numbers wont match if you use anonymous files for the extracted doc) and many more severe regressions. I am not saying the status quo is perfect. They are slow, I know. This is an issue for the ~5 people who compile more than once a day. I am saying most possible paths going forward either take a lot of time or are inferior to the current CMake solution (for usability _and_ feature set)

psychon:
> I still argue the test output is relevant and its already used in many examples.

```
$ grep -r 'Usage example output' build/lib
build/lib/awful/mouse/init.lua:--**Usage example output**:
build/lib/awful/placement.lua:--**Usage example output**:
build/lib/awful/placement.lua:--**Usage example output**:
build/lib/gears/object.lua:--**Usage example output**:
```
That's four and the only non-trivial one is in `gears.object`. I could live without these if we fix all of the other problems instead. Hardcoding the expected output and verifying it matches would even allow to keep this feature.

> Your suggestion would also force to hardcode the image path,

Yup. I don't really see this as much of a problem. The current state hardcodes the name of the test (`@DOC_awful_placement_no_offscreen_EXAMPLE@`). That's about equivalent.

> make the source unreadable,

Which source? The actual library code or the example? Why does it become unreadable?

> break syntax highlight when developing examples,

Well, yes. I could live with that. Examples could be developed outside of the source and then copied over, or we use some block-comment-thingie that can easily be removed when one wants to "play" with the example.

> make luacheck and coverage a nightmare to use (the line numbers wont match if you use anonymous files for the extracted doc)

Why do you want code coverage information for examples? If it weren't for other issues in our complex build system preventing that, we would already have a (working) exclude-rule for `tests/examples`. Also, examples should be trivial enough that we do not need code coverage information for them.

Also, the whole point of this exercise is to solve the current nightmare that is coverage. My proposal tries to do that and simplify the build system at the same time (there is just one source code, all automatic replacements done on it are quite trivial).

And yeah, `luacheck` wouldn't check the examples. As long as they are actually working, I can live with that. I bet 90% of user configs wouldn't pass `luacheck` either.

> and many more severe regressions.

I guess we should document all of the requirements. My favourite requirement would be "can be explained on a napkin" / KISS.

> I am not saying the status quo is perfect. [...] I am saying most possible paths going forward either take a lot of time or are inferior to the current CMake solution (for usability and feature set)

Yup, I agree.


We might have to have this discussion at some point, but I don't think that we should do so now and certainly not here. Sorry for bringing it up again.

Elv13:
> Why do you want code coverage information for examples?

Because ~12% of the code coverage from from there. Including both "hits" and "real test"

    cd dev/awesome/tests/examples/
    git grep assert | wc -l 
    > 144

I wrote all my tests there because it has a re-usable output that double as a visual test and a code example. Not doing that would require having **both** these tests and copy paste `spec` ones. IMO that's a pure waste of time. Write once, use twice, that was the whole point why I wrote this framework.

Also, my layout system also has its test and coverage based on this, so do some of my (unpublished) test code for my modules.

Elv13:
Also, one last thing. If you "just" want to fix the coverage, disable the `configure_file` call when the coverage is enabled.

psychon:
>> Why do you want code coverage information for examples?
> Because ~12% of the code coverage from from there. Including both "hits" and "real test"

Ah, ok. This is just a misunderstanding. I thought you wanted code coverage information for files in `tests/examples` (Something like "75% of lines in `tests/examples/foo.lua` were executed). However, you actually want code coverage information *generated*/*caused* by files in `tests/examples`. Of course, this should stay.

> (the line numbers wont match if you use anonymous files for the extracted doc)

I just don't understand this problem then. We do not care about the line numbers in these extracted files anywhere.

blueyed:
@Elv13 @psychon 
Sorry, I've only skimmed this now again, because I got the impression that this should be reverted/addressed somehow.
Are we good with this for now and addressing this at other fronts from here?

psychon:
I won't claim that I really understand the magic here, but couldn't this be fixed in CMake? For example, something like the following untested patch?
```diff
diff --git a/tests/examples/CMakeLists.txt b/tests/examples/CMakeLists.txt
index 148cee0..c8db09e 100644
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -56,7 +56,7 @@ function(escape_string variable content escaped_content line_prefix)
     set(tmp_output ${content})
     foreach (LINE ${var_lines})
         if(NOT LINE MATCHES "^.+--DOC_[A-Z]+$")
-            set(tmp_output ${tmp_output}\n--${line_prefix}${LINE})
+            set(tmp_output "${tmp_output}\n--${line_prefix}${LINE} ")
         endif()
     endforeach()
 
```

blueyed:
Sounds like a plan.
@Elv13 ?

Elv13:
Needs testing

PR #1230, opened 2016-11-21
Add check-qa target and only run it once on Travis

blueyed:
This just also help to easier spot failures because of this when looking
at the build matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ branches:
 
 env:
   matrix:
-    - LUA=5.2 LUANAME=lua5.2 BUILD_APIDOC=true DO_COVERAGE=coveralls
+    - LUA=5.2 DO_CHECKQA=1 LUANAME=lua5.2 BUILD_APIDOC=true DO_COVERAGE=coveralls
     # luajit: installed from source.
     - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.4 TEST_PREV_COMMITS=1
     # Note: luarocks does not work with Lua 5.0.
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -151,12 +151,13 @@ script:
...
   - |
     set -ev
     if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
       if [ "$DO_COVERAGE" != "0" ]; then
         travis_fold_start "DO_COVERAGE"
-        (make check-coverage \
+        (make check-unit-coverage \
          && do_codecov unittests \
          && sed -i "1 i\\require('luacov.runner')('"$PWD"/.luacov')" build/awesomerc.lua \
          && BUILD_DIRECTORY="" tests/run.sh \
```
blueyed: This is basically `make check-integration`.  Not sure if we could do `BUILD_DIRECTORY="" make check-integration` here to clarify it?!

.travis.yml
```diff
@@ -165,9 +166,15 @@ script:
         travis_fold_end
         [ "$ret" = 0 ]
       else
-        travis_run_in_fold "make.check" make check
+        travis_run_in_fold "make.check-unit" make check-unit
+        travis_run_in_fold "make.check-integration" make check-integration
       fi
     fi
+  # Run check-qa.
```
blueyed: This comment seems to be superfluous?!

blueyed:
I have hoped that we could improve the general build time with this, but the `DO_COVERAGE` builds are still both above 10 minutes..

When looking into it quickly, ac7ea4c might be causing it after all (the re-run therein)  /cc @Elv13.

blueyed:
Assigning to @psychon for review.

blueyed:
> Unrelated: I wonder if `travis_run_in_fold` really is such a great thing; sometimes looking at Travis ends up being confusing.

Yeah!  I think it was good once, but Travis might have changed in this regard.

Squashing and addressing the comment now.

PR #1233, opened 2016-11-22
Guarantee stable output from 06-appearence.md.lua

psychon:
This function was iterating over a table with pairs() to generate output (the
sample theme file). Since pairs() does not guarantee any iteration order, this
lead to a different order each time this file was generated. This is, for
example, visible in the diffs in the generate api documentation repository.

Fix this by using a self-made iteration function which behaves like pairs(), but
guarantees an iteration order sorted by keys.

### DIFF
--- a/docs/06-appearance.md.lua
+++ b/docs/06-appearance.md.lua
@@ -8,6 +8,27 @@ local glib = require("lgi").GLib
 local name_attr = gio.FILE_ATTRIBUTE_STANDARD_NAME
 local type_attr = gio.FILE_ATTRIBUTE_STANDARD_TYPE
 
+-- Like pairs(), but iterate over keys in a sorted manner. Does not support
+-- modifying the table while iterating.
+local function sorted_pairs(t)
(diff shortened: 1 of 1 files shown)


actionless:
i am using almost the same helper (but with ability to easily do reverse-sort) in my config, mb we would better move it to some `awful.utils` or so to make it importable from elsewhere?

blueyed:
@actionless 
Good points!

Please create a new PR to improve this.

actionless:
as it was mointed in my update number two, i think it's not possible unfortunately

psychon:
We could always split up generic helpers into another package which can then be installed e.g. via luarocks (/me hides).
(Well, and we could use some already-existing things from out there instead of reinventing all the wheels that others already invented)

actionless:
like this one? https://mirven.github.io/underscore.lua/

Elv13:
Please don't pull random lua rocks as "hard" dependencies. We already have enough issues with the ones in the tests.

PR #1236, opened 2016-11-27
rules.new_tag: Fix when the tag screen doesn't match the client

Elv13:
Also add the documented `props` argument to the high priority
rules.

### DIFF
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -312,8 +312,9 @@ end
 --- Create a new tag based on a rule.
 -- @tparam client c The client
 -- @tparam boolean|function|string value The value.
+-- @tparam table props The properties.
 -- @treturn tag The new tag
-function rules.high_priority_properties.new_tag(c, value)
+function rules.high_priority_properties.new_tag(c, value, props)
     local ty = type(value)
     local t = nil
 
@@ -324,8 +325,16 @@ function rules.high_priority_properties.new_tag(c, value)
         -- Create a tag named after "value"
         t = atag.add(value, {screen=c.screen, volatile=true})
     elseif ty == "table" then
-        -- Assume a table of tags properties
-        t = atag.add(value.name or c.class or "N/A", value)
+        -- Assume a table of tags properties. Set the right screen, but
+        -- avoid editing the original table
+        local values = value.screen and value or util.table.clone(value)
+        values.screen = values.screen or c.screen
+
+        t = atag.add(value.name or c.class or "N/A", values)
+
+        -- In case the tag has been forced to another screen, move the client
+        c.screen = t.screen
+        props.screen = t.screen -- In case another rule query it
     else
         assert(false)
     end
@@ -409,7 +418,7 @@ function rules.execute(c, props, callbacks)
                 value = value(c, props)
             end
 
-            handler(c, value)
+            handler(c, value, props)
         end
 
     end
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -195,7 +195,7 @@ start_awesome() {
 # Count errors.
 errors=0
 # Seconds after when awesome gets killed.
-timeout_stale=120 # FIXME This should be no more than 60s
+timeout_stale=180 # FIXME This should be no more than 60s
 
 for f in $tests; do
     echo "== Running $f =="
--- a/tests/test-awful-client.lua
+++ b/tests/test-awful-client.lua
@@ -209,6 +209,36 @@ table.insert(multi_screen_steps, function()
     return true
 end)
 
+-- Test the `new_tag` rule
+table.insert(multi_screen_steps, function()
+    for _, c in ipairs(client.get()) do
+        c:kill()
+    end
+
+    for i=1, screen.count() do
+        local s = screen[i]
+        test_client("screen"..i, nil, {
+            new_tag = {
+                name = "NEW_AT_"..i,
+                screen = s,
+            }
+        })
+    end
+
+    return true
+end)
+
+table.insert(multi_screen_steps, function()
+    if #client.get() ~= screen.count() then return end
+
+    for _, c in ipairs(client.get()) do
+        assert(#c:tags() == 1)
+        assert(c.first_tag.name == "NEW_AT_"..c.screen.index)
+    end
+
+    return true
+end)
+
 require("_multi_screen")(steps, multi_screen_steps)
 
 require("_runner").run_steps(steps)


Elv13:
... The new test causes to bust the 120 seconds limit of `test-awful-client.lua` again...

lib/awful/rules.lua
```diff
@@ -324,8 +325,16 @@ function rules.high_priority_properties.new_tag(c, value)
...
+        -- Assume a table of tags properties. Set the right screen, but
+        -- avoid editing the original table
+        local values = value.screen and value or util.table.clone(value)
+        values.screen = values.screen or c.screen
+
+        t = atag.add(value.name or c.class or "N/A", values)
+
+        -- In case the tag has been forced to another screen, move the client
+        c.screen = t.screen
+        props.screen = t.screen -- In case another rule query it
```
psychon: NAK. You can't modify rules just because some tag had a different screen. Following executions will see this modification.

psychon: How about the following?
```lua
t = atag.add(value.name or c.class or "N/A", value)
if props.screen then t.screen = props.screen end
```

Elv13: >  Following executions will see this modification.

That's exactly the idea. The `screen` property has higher priority than this, so it has already been executed. If a further rule query the `props` array to check the screen, it will get the wrong one.

This table is generated by the code, it doesn't add garbage anywhere else. So better fix it to limit the side effects of the other rules.

Elv13: > How about the following?

That's kind of the bug this PR try to fix. If the `new_tag` is `{name="Foo", screen=capi.mouse.screen}` using `awful.spawn`, then you don't want the default rule overrideing that. The spawn arguments always have priority over the default fallbacks (actually, this work because of a race condition, but whatever, works as long as `awful.spawn` is loaded after `awful.rules`).

psychon:
How about this (ontop of your patches)?
```diff
diff --git a/lib/awful/rules.lua b/lib/awful/rules.lua
index c027d4a..c1d099a 100644
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -327,14 +327,24 @@ function rules.high_priority_properties.new_tag(c, value, props)
     elseif ty == "table" then
         -- Assume a table of tags properties. Set the right screen, but
         -- avoid editing the original table
-        local values = value.screen and value or util.table.clone(value)
-        values.screen = values.screen or c.screen
-
-        t = atag.add(value.name or c.class or "N/A", values)
+        local s = value.screen and screen[value.screen]
+        if props.screen then
+            -- props.screen might be a function, but it was already applied.
+            if s and s ~= c.screen then
+                require("gears.debug").print_warning(
+                    "Contradictory screen assignment in awful.rule-rule: "..
+                    "Created tag should have a different screen than client. "..
+                    "Ignoring tag's screen")
+            end
+            s = c.screen
+        else
+            s = s or c.screen
+            c.screen = s
+        end
 
-        -- In case the tag has been forced to another screen, move the client
-        c.screen = t.screen
-        props.screen = t.screen -- In case another rule query it
+        value = util.table.clone(value)
+        value.screen = s
+        t = atag.add(value.name or c.class or "N/A", value)
     else
         assert(false)
     end
```
Your idea is to ignore `props.screen` if `props.new_tag.screen` is also set. My patch tries to ignore `props.new_tag.screen` in this case (and prints a warning). This is why `props.screen` does not need to be modified.

lib/awful/rules.lua
psychon: I didn't see your comment before writing my "top-level" comment, but... now I wonder why my patch works. You are right that the default rule should mean that the wrong screen is assigned. And yeah, I missed that this is actually a generated table that may be modified. Sorry for that.

Elv13:
But that's the bug, `new_tag.screen` should takes priority. It makes otherwise no sense to set it. The use case for this is `awful.spawn`. If this new_tag table is stored in a variable, then this forces the user to manually add `awful.spawn(foo, {new_tag=mytag_constrctor_function, screen = mytag_constrctor().screen})` then delete the tag. It just cannot work that way.

Elv13:
Ok, we commented at the same time. So the next question is, should I bump the timeout to 600 seconds? If I bump it to 180, then it will break again. If I add a new test suits, then it will take _more_ total time. While the forced screen test needs to be executed against all possible screen layout, the `new_tag` only need to check against 1 multi-screen scenario. But supporting 2 multi-screen layout suit isn't related to this bug fix, so given I have little time, I suggest to go ahead with a ridiculous timeout and handle this later. After all, we almost never hit that timeout. It either so a stack overflow or is killed by the kernel memory quota watchdog unless there really is a `while 1 do end` type of loop.

psychon:
Feel free to bump the timeout to 600 seconds.

Here is another new approach for fixing this, what do you think (it first reverts your patch, sorry for the noise)? Test seems to be happy with this.
```diff
diff --git a/lib/awful/rules.lua b/lib/awful/rules.lua
index c027d4a..ba225fc 100644
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -312,9 +312,8 @@ end
 --- Create a new tag based on a rule.
 -- @tparam client c The client
 -- @tparam boolean|function|string value The value.
--- @tparam table props The properties.
 -- @treturn tag The new tag
-function rules.high_priority_properties.new_tag(c, value, props)
+function rules.high_priority_properties.new_tag(c, value)
     local ty = type(value)
     local t = nil
 
@@ -325,16 +324,8 @@ function rules.high_priority_properties.new_tag(c, value, props)
         -- Create a tag named after "value"
         t = atag.add(value, {screen=c.screen, volatile=true})
     elseif ty == "table" then
-        -- Assume a table of tags properties. Set the right screen, but
-        -- avoid editing the original table
-        local values = value.screen and value or util.table.clone(value)
-        values.screen = values.screen or c.screen
-
-        t = atag.add(value.name or c.class or "N/A", values)
-
-        -- In case the tag has been forced to another screen, move the client
-        c.screen = t.screen
-        props.screen = t.screen -- In case another rule query it
+        -- Assume a table of tags properties.
+        t = atag.add(value.name or c.class or "N/A", value)
     else
         assert(false)
     end
@@ -374,9 +365,26 @@ end
 
 --- Apply properties and callbacks to a client.
 -- @client c The client.
--- @tab props Properties to apply.
+-- @tab props Properties to apply. This table may be modified!
 -- @tab[opt] callbacks Callbacks to apply.
 function rules.execute(c, props, callbacks)
+    -- Clean-up the passed in argument: new_tag.screen and screen should agree.
+    if props.new_tag then
+        local s = props.screen or props.new_tag.screen
+        if not s then
+            s = screen.preferred()
+        end
+        if s ~= props.new_tag.screen then
+            require("gears.debug").print_warning(
+                "Contradictory screen assignment in awful.rule-rule: "..
+                "Created tag should have a different screen than client. "..
+                "Ignoring tag's screen")
+        end
+        s = type(s) == "function" and screen[s(c, props)] or screen[s]
+        props.screen = s
+        props.new_tag.screen = s
+    end
+
     -- This has to be done first, as it will impact geometry related props.
     if props.titlebars_enabled then
         c:emit_signal("request::titlebars", "rules", {properties=props})
```

Elv13:
> Here is another new approach for fixing this

I avoided doing that on purpose. It has a couple of issues. First, it re-add hardcoded property logic to `rules.execute` after I tried hard to remove it so Tyrannical like modules can re-use it (they could monkey patch the individual rule funtions). The second isssue is purely separation of concern. My patch act more like a chain of resposability and is mostly self contained while yours "wont scale" is other issues like this arise. Finally, I don't think the warning is sueful, it can and will happen under normal usage if there is a global screen rule.

psychon:
...I still don't really understand why that warning didn't fire in the tests.

Anyway: How about introducing some other chain of callback functions that should do nothing but patch the properties before they are applied? The problem is, after all, a consistency one: The inconsistency that you are trying to overcome is `props.screen ~= props.new_tag.screen`.
That sounds saner than patching the properties in the middle of being applied.

Elv13:
> How about introducing some other chain of callback functions that should do nothing but patch the properties before they are applied?

It would be just as prone to race conditions. It would move the problem and somewhat be better, but it would solve nothing. All properties can be callbacks, all callbacks can mess the state. All of those extra properties handler can also mess the state. To implement your idea, all property function -> values would have to be converted first, then the rules applied. But really, it doesn't solve anything.

Elv13:
Also (sorry for the noise, I just don't want to do an "Edit999" in the previous comment as this is somewhat important. To reduce the issues and probably the reason why the warning wasn't printed for you. I propose removing the `connect_signal("manage")` from the rule and extending the one present in `awful.spawn`. Right now the rules can be applied twice (unless I missed something). This is what the Tyrannical link above fixes by merging all 3 "property sources" into 1.

psychon:
Another random idea: Why do we support `props.new_tag.screen` at all? I propose to ignore it (perhaps print a warning if it is present) and just have the new tag on screen `c.screen`. That would solve all problems, wouldn't it?

Elv13:
No, it is important, that's what I use the most of `new_tag`. I store some "futuretag" on shortcuts.

Elv13:
sooo? I still propose to merge this PR as-is.

psychon:
Sigh. I still think that this is a bad band-aid that papers around the problem that having two different ways to specify a screen cannot work (and I don't really buy the "but I use it like this in my config", you could always move/copy `props.new_tag.screen` to `props.screen` in your config). However, since this is basically your code, feel free.

Elv13:
I agree it is a band-aid and I agree it cannot work 100% of the time, ever. However my argument is that it is unavoidable. The `tag` and `tags` rules have similar issues and in their case, the same band-aid will be necessary. It is worst when the tag already exist, because you can't even ignore the screen issue. This PR doesn't even try (yet) to fix that one. It fixes `new_tag` because the bug was just very, very bad. A future PR will need more and more band-aids against the other similar corner cases (and those simply cannot be ignored/mitigated like you propose). So given the issue has be resolved using band-aids for `tag` and `tags`, `new_tag` has to do it the same way for the sake of consistency.

Issue #1237, opened 2016-11-27
We have some slow tests


psychon:
I was just reminded by [this](https://github.com/awesomeWM/awesome/pull/1236#issuecomment-263113855) that we have this issue. However, I cannot find an issue about this so far. Here it is.

The problem is that `tests/test-awful-client.lua` and `tests/test-awful-tag.lua` are slow. A while ago (in some comment somewhere; we really should have had this issue a long time ago) we figured out that the problem was that some of the intermediate screens are not being garbage collected.

Today I looked into that problem again and here is what I found:
The problem seems to be wibars in connected with screens. When a screen is removed, its wibar (s.mywibar) is [updated](https://github.com/awesomeWM/awesome/blob/333cd6491f57d8ce9b7d665f2b48d8bc8e51682d/lib/wibox/drawable.lua#L475) and now its `_widget_context` refers to another, still existing screen. When that screen is removed, well, our wibar will now keep it alive.

I'm still not sure what the root of this "chain of wibar-screen-wibar-screen-...." is, but this is how the amount of "garbage wibox" keeps increasing...
Ok, the root of this seems to be `mymainmenu.wibox._drawable._widget_context`. So I guess this isn't really relevant "in the real world"?

psychon:
Looking at `test-awful-client.lua` with a profiler says:
- 90% of time in `gears.timer` (so I guess this means inside of `delayed_call`s).
- 66% in `do_redraw` from `wibox.drawable` (this is both relayout and redraw!)
- 38% inside of `hierarchy_update` (so actual relayouts)
- 41% inside of `wibox.hierarchy:draw` (so actual redraws)

The numbers don't quite add up, but I think it's safe to conclude that something is redrawing a lot...

psychon:
Let's look at some relayout and redraw numbers:
```diff
diff --git a/awesomerc.lua b/awesomerc.lua
index f329240..7105b38 100755
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -1,3 +1,11 @@
+laycounter, drawcounter = {}, {}
+actlaycounter, actdrawcounter = {}, {}
+function redraw(drawable, hierarchy, x, y, width, height)
+    drawcounter[drawable] = (drawcounter[drawable] or 0) + 1
+end
+function relayout(drawable, hierarchy)
+    laycounter[drawable] = (laycounter[drawable] or 0) + 1
+end
 -- @DOC_REQUIRE_SECTION@
 -- Standard awesome library
[...]

PR #1238, opened 2016-11-27
wibox.drawable: Support forced screens

psychon:
Up to now, a drawable always figured out the screen that it is on by
looking at its position. This causes memleak-like problems with wibars:
A wibar has a screen assigned, but its underlying drawable will end up
referring to another screen. Via this, we were managing to build a long
reference chain of screens and drawable that meant that none of the fake
screens that our test suite added could be garbage collected.

To fix this, add wibox.drawable._force_screen(s). After this function is
called, the normal screen detection based on the position is skipped and
instead the given screen is always used. This breaks the above reference
chain and things become garbage-collectable.

Also, this chains the drawable to the life time of the screen: When the
screen becomes invalid (.valid == false), the drawable will stop
redrawing.

Fixes: https://github.com/awesomeWM/awesome/issues/1237

### DIFF
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -30,13 +30,16 @@ local visible_drawables = {}
 local function get_widget_context(self)
     local geom = self.drawable:geometry()
 
-    local sgeos = {}
+    local s = self._forced_screen
+    if not s then
+        local sgeos = {}
 
-    for s in capi.screen do
-        sgeos[s] = s.geometry
+        for s in capi.screen do
+            sgeos[s] = s.geometry
+        end
+
+        s = grect.get_by_coord(sgeos, geom.x, geom.y) or capi.screen.primary
     end
-
-    local s = grect.get_by_coord(sgeos, geom.x, geom.y) or capi.screen.primary
 
     local context = self._widget_context
     local dpi = beautiful.xresources.get_dpi(s)
@@ -59,6 +62,7 @@ end
 
 local function do_redraw(self)
     if not self.drawable.valid then return end
+    if self._forced_screen and not self._forced_screen.valid then return end
 
     local surf = surface.load_silently(self.drawable.surface, false)
     -- The surface can be nil if the drawable's parent was already finalized
@@ -272,6 +276,10 @@ function drawable:set_fg(c)
     end
     self.foreground_color = c
     self._do_complete_repaint()
+end
+
+function drawable:_force_screen(s)
+    self._forced_screen = s
 end
 
 function drawable:_inform_visible(visible)
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -86,6 +86,7 @@ function wibox:set_screen(s)
     -- Remember this screen so things work correctly if screens overlap and
     -- (x,y) is not enough to figure out the correct screen.
     self.screen_assigned = s
+    self._drawable:_force_screen(s)
 end
 
 for _, k in pairs{ "buttons", "struts", "geometry", "get_xproperty", "set_xproperty" } do


psychon:
Sigh, builds are still slow. I'll try sticking some `collectgarbage` calls in.

Elv13:
```
+[ ! -r /home/travis/build/awesomeWM/awesome/tests/test-awful-tag.lua ]
+DISPLAY=:5 /home/travis/build/awesomeWM/awesome/tests/../utils/awesome-client
+cat /home/travis/build/awesomeWM/awesome/tests/test-awful-tag.lua
+tail -n 100000 -f --pid 24659 /tmp/tmp.kLkahENMMN/_awesome_test.log
2016-11-27 17:24:28 W: awesome: a_glib_poll:387: Last main loop iteration took 1.052843 seconds! Increasing limit for this warning to that value.
2016-11-27 17:24:30 W: awesome: a_glib_poll:387: Last main loop iteration took 1.881237 seconds! Increasing limit for this warning to that value.
2016-11-27 17:24:33 W: awesome: a_glib_poll:387: Last main loop iteration took 2.204811 seconds! Increasing limit for this warning to that value.
2016-11-27 17:24:50 W: awesome: a_glib_poll:387: Last main loop iteration took 3.345098 seconds! Increasing limit for this warning to that value.
2016-11-27 17:25:02 W: awesome: a_glib_poll:387: Last main loop iteration took 4.500988 seconds! Increasing limit for this warning to that value.
```

Indeed...

psychon:
Sigh.

psychon:
So my WIP-commit shows that screens are now correctly garbage-collected, so I guess that there is still something else lingering around...

Elv13:
It's the tags too, not only the screens. At the end of the run, there is hundreds of tags

psychon:
@Elv13 I'm confused. Here you say:

> It's the tags too, not only the screens. At the end of the run, there is hundreds of tags
[...]
> With your patch it is a lot better. 

Four hours later you write at https://github.com/awesomeWM/awesome/pull/1239#issuecomment-267741422 that 

> @psychon Your patch to stop the screens for leaking is totally working. It also prevents the tags from leaking. 

So... Is this patch enough or not?

Note that I didn't look at your drill-speed-holes-PR yet and so I don't know what is in there.

Elv13:
I was confused too when I started again to fix the CI. The other PR has
better explanations and includes the commits from this one

psychon:
I did a force-push to get rid of commits fe5dbc21649b4a3be19478eddc8f292645b36313 and 3f30a11bfbd9d5f67ca107eef35c7d6f7417f0b9. According to @Elv13 the remaining commit is mergable (?).

psychon:
Ping? @Elv13

Elv13:
Yes, see https://github.com/awesomeWM/awesome/pull/1239#issuecomment-267741422 . (for #1239) My best guess for the coveralls issue a an hidden file is in the wrong place. This is how I fixed CodeCov. After that, the whole PR could be merged. Coveralls could also be disabled given it has little value?

blueyed:
> Coveralls could also be disabled given it has little value?

I don't think so: it's good to see what code is covered after all.

Let's merge this as a next step?

psychon:
Sigh. This now caused a luacheck warning.

Elv13:
> I don't think so: it's good to see what code is covered after all.

@blueyed Sorry for not being clear. My point was that *a fixed codecov* has more value than a *both broken coverage tools*. As far as I can see, CodeCov is always better than Coveralls. The main problem is that currently, both coverage tools show the right coverage percentage, but you can't see what is actually covered because the lines don't match (giving you pseudo random coverage data). That PR fixes CodeCov by moving around some files, but as found earlier, Coveralls doesn't play well with that trick and lose either all integration test coverage+specs or all pre-integration coverage. The integration+specs have an 8.3% lower coverage (64%). The pre-integration covers about 53% of the code. Both of them cover 73%. But again, the main

blueyed:
Ok, let's remove coveralls then for now?!

PR #1239, opened 2016-11-27
Improve the CI speed and accuracy

### LINKED ISSUE #962: CMake configure_file should be disabled for the CodeCov build

### LINKED ISSUE #982: Travis is not very stable as of late.

### LINKED ISSUE #1142: "make ldoc" should be smarter/faster
`make ldoc` does too much:

As @psychon mentioned on IRC it seems to be an issue of cmake doing too much
when invoked through the main Makefile.

```
Running cmake…
-- git -> /usr/bin/git
-- asciidoc -> /usr/bin/asciidoc
-- xmlto -> /usr/bin/xmlto
-- gzip -> /usr/bin/gzip
-- ldoc -> /home/user/.luarocks/bin/ldoc
-- convert -> /usr/bin/convert
-- checking for execinfo -- found
-- checking for round -- in libm
-- Running tests/examples/awful/mouse/coords.lua...
-- Running tests/examples/awful/placement/align.lua...
-- Running tests/examples/awful/placement/bottom.lua...
-- Running tests/examples/awful/placement/bottom_left.lua...
-- Running tests/examples/awful/placement/bottom_right.lua...
-- Running tests/examples/awful/placement/center_horizontal.lua...
-- Running tests/examples/awful/placement/center_vertical.lua...
-- Running tests/examples/awful/placement/centered.lua...
-- Running tests/examples/awful/placement/closest_mouse.lua...
-- Running tests/examples/awful/placement/compose.lua...
-- Running tests/examples/awful/placement/compose2.lua...
-- Running tests/examples/awful/placement/left.lua...
-- Running tests/examples/awful/placement/maximize.lua...
-- Running tests/examples/awful/placement/maximize_horizontally.lua...
-- Running tests/examples/awful/placement/maximize_vertically.lua...
-- Running tests/examples/awful/placement/next_to_mouse.lua...
-- Running tests/examples/awful/placement/no_offscreen.lua...
-- Running tests/examples/awful/placement/no_overlap.lua...
-- Running tests/examples/awful/placement/resize_to_mouse.lua...
-- Running tests/examples/awful/placement/right.lua...
-- Running tests/examples/awful/placement/stretch.lua...
-- Running tests/examples/awful/placement/stretch_down.lua...
-- Running tests/examples/awful/placement/stretch_left.lua...
-- Running tests/examples/awful/placement/stretch_right.lua...
-- Running tests/examples/awful/placement/stretch_up.lua...
-- Running tests/examples/awful/placement/top.lua...
-- Running tests/examples/awful/placement/top_left.lua...
-- Running tests/examples/awful/placement/top_right.lua...
-- Running tests/examples/awful/placement/under_mouse.lua...
-- Running tests/examples/gears/shape/arc.lua...
-- Running tests/examples/gears/shape/arrow.lua...
-- Running tests/examples/gears/shape/circle.lua...
-- Running tests/examples/gears/shape/cross.lua...
-- Running tests/examples/gears/shape/hexagon.lua...
-- Running tests/examples/gears/shape/infobubble.lua...
-- Running tests/examples/gears/shape/isosceles_triangle.lua...
-- Running tests/examples/gears/shape/losange.lua...
-- Running tests/examples/gears/shape/octogon.lua...
-- Running tests/examples/gears/shape/parallelogram.lua...
-- Running tests/examples/gears/shape/partially_rounded_rect.lua...
-- Running tests/examples/gears/shape/pie.lua...
-- Running tests/examples/gears/shape/powerline.lua...
-- Running tests/examples/gears/shape/radial_progress.lua...
-- Running tests/examples/gears/shape/rectangle.lua...
-- Running tests/examples/gears/shape/rectangular_tag.lua...
-- Running tests/examples/gears/shape/rounded_bar.lua...
-- Running tests/examples/gears/shape/rounded_rect.lua...
-- Running tests/examples/text/gears/object/properties.lua...
-- Running tests/examples/text/gears/object/signal.lua...
-- Running tests/examples/wibox/container/arcchart/bg.lua...
-- Running tests/examples/wibox/container/arcchart/border_width.lua...
-- Running tests/examples/wibox/container/arcchart/paddings.lua...
-- Running tests/examples/wibox/container/arcchart/rounded_edge.lua...
-- Running tests/examples/wibox/container/arcchart/start_angle.lua...
-- Running tests/examples/wibox/container/arcchart/thickness.lua...
-- Running tests/examples/wibox/container/arcchart/value.lua...
-- Running tests/examples/wibox/container/background/bg.lua...
-- Running tests/examples/wibox/container/background/clip.lua...
-- Running tests/examples/wibox/container/background/fg.lua...
-- Running tests/examples/wibox/container/background/shape.lua...
-- Running tests/examples/wibox/container/defaults/arcchart.lua...
-- Running tests/examples/wibox/container/defaults/background.lua...
-- Running tests/examples/wibox/container/defaults/constraint.lua...
-- Running tests/examples/wibox/container/defaults/margin.lua...
-- Running tests/examples/wibox/container/defaults/mirror.lua...
-- Running tests/examples/wibox/container/defaults/radialprogressbar.lua...
-- Running tests/examples/wibox/container/defaults/rotate.lua...
-- Running tests/examples/wibox/container/radialprogressbar/border_color.lua...
-- Running tests/examples/wibox/container/radialprogressbar/border_width.lua...
-- Running tests/examples/wibox/container/radialprogressbar/color.lua...
-- Running tests/examples/wibox/container/radialprogressbar/padding.lua...
-- Running tests/examples/wibox/container/radialprogressbar/value.lua...
-- Running tests/examples/wibox/container/rotate/angle.lua...
-- Running tests/examples/wibox/layout/defaults/align.lua...
-- Running tests/examples/wibox/layout/defaults/fixed.lua...
-- Running tests/examples/wibox/layout/defaults/flex.lua...
-- Running tests/examples/wibox/layout/defaults/ratio.lua...
-- Running tests/examples/wibox/layout/defaults/stack.lua...
-- Running tests/examples/wibox/widget/checkbox/bg.lua...
-- Running tests/examples/wibox/widget/checkbox/check_shape.lua...
-- Running tests/examples/wibox/widget/checkbox/custom.lua...
-- Running tests/examples/wibox/widget/checkbox/shape.lua...
-- Running tests/examples/wibox/widget/defaults/checkbox.lua...
-- Running tests/examples/wibox/widget/defaults/graph.lua...
-- Running tests/examples/wibox/widget/defaults/imagebox.lua...
-- Running tests/examples/wibox/widget/defaults/piechart.lua...
-- Running tests/examples/wibox/widget/defaults/progressbar.lua...
-- Running tests/examples/wibox/widget/defaults/textbox.lua...
-- Running tests/examples/wibox/widget/graph/step.lua...
-- Running tests/examples/wibox/widget/piechart/border_color.lua...
-- Running tests/examples/wibox/widget/piechart/border_width.lua...
-- Running tests/examples/wibox/widget/piechart/label.lua...
-- Running tests/examples/wibox/widget/progressbar/bar_shape.lua...
-- Running tests/examples/wibox/widget/progressbar/clip.lua...
-- Running tests/examples/wibox/widget/progressbar/shape.lua...
-- Running tests/examples/wibox/widget/progressbar/text.lua...
-- Running tests/examples/wibox/widget/progressbar/vertical.lua...
-- Example tests passed.
-- busted -> /home/user/.luarocks/bin/busted
-- luacheck -> /home/user/.luarocks/bin/luacheck
-- Package version will be set to 3.5.60~git2055-g771e6e9-dirty.
-- rpmbuild not found.
-- The 'package' target will not build a RPM
-- dpkg not found.
-- The 'package' target will not build a DEB
-- Configuring done
-- Generating done
-- Build files have been written to: …/a.example.com-x86_64-pc-linux-gnu-6.2.1
Running make Makefile…
Running make ldoc…
reading configuration from config.ld
found master module …/a.example.com-x86_64-pc-linux-gnu-6.2.1/luaa.c
found master module …/a.example.com-x86_64-pc-linux-gnu-6.2.1/luaa.c
found master module …/a.example.com-x86_64-pc-linux-gnu-6.2.1/objects/client.c
found master module …/a.example.com-x86_64-pc-linux-gnu-6.2.1/mouse.c
found master module …/a.example.com-x86_64-pc-linux-gnu-6.2.1/objects/screen.c
found master module …/a.example.com-x86_64-pc-linux-gnu-6.2.1/objects/tag.c
found master module …/a.example.com-x86_64-pc-linux-gnu-6.2.1/lib/naughty/core.lua
format: discount not found, using markdown
output written to …/a.example.com-x86_64-pc-linux-gnu-6.2.1/doc
[100%] Built target ldoc
  8.93s user 0.49s system 97% cpu 9.652 total
```

Running it in the `build` dir is much faster:

```
reading configuration from config.ld
found master module     …/a.example.com-x86_64-pc-linux-gnu-6.2.1/luaa.c
found master module     …/a.example.com-x86_64-pc-linux-gnu-6.2.1/luaa.c
found master module     …/a.example.com-x86_64-pc-linux-gnu-6.2.1/objects/client.c
found master module     …/a.example.com-x86_64-pc-linux-gnu-6.2.1/mouse.c
found master module     …/a.example.com-x86_64-pc-linux-gnu-6.2.1/objects/screen.c
found master module     …/a.example.com-x86_64-pc-linux-gnu-6.2.1/objects/tag.c
found master module     …/a.example.com-x86_64-pc-linux-gnu-6.2.1/lib/naughty/core.lua
format: discount not found, using markdown
output written to …/a.example.com-x86_64-pc-linux-gnu-6.2.1/doc
[100%] Built target ldoc
```

Elv13:
image

Fix #1142 
Fix #982 
Fix #962
Fix #964 
Close #1238

### DIFF
--- a/.luacov
+++ b/.luacov
@@ -3,7 +3,6 @@
 -- This variable is set externally
 local source = os.getenv("SOURCE_DIRECTORY") or os.getenv('PWD')
 local build = os.getenv("BUILD_DIRECTORY") or (source .. "/build")
-local lib_dir = os.getenv("AWESOME_LIB_DIR") or (source .. "/lib")
 
 local function escape_pattern(str)
     return string.gsub(str, "%W", "%%%1")
@@ -13,17 +12,15 @@ return {
     statsfile = build .. "/luacov.stats.out",
 
     include = {
-        escape_pattern(lib_dir) .. "/.+",
-
-        -- For things already having the correct path
-        -- (happens with integration tests)
-        "^lib/",
+        escape_pattern(source) .. "/.+",
+        -- Relative paths (to source/build dir).
+        '^[^/].+',
     },
 
     -- configuration for luacov-coveralls reporter
     coveralls = {
         pathcorrect = {
-            { escape_pattern(lib_dir..'/'), "lib/"},
+            { escape_pattern(source .. '/'), ''},
         },
     },
 }
--- a/.luacov
+++ b/.luacov
@@ -1,33 +1,29 @@
 -- Configuration file for LuaCov
 
 -- This variable is set externally
-local build = os.getenv("BUILD_DIRECTORY") or error("$BUILD_DIRECTORY not set")
-local lib_dir = os.getenv("AWESOME_LIB_DIR") or build .. "lib/"
-local source = os.getenv("SOURCE_DIRECTORY")
+local source = os.getenv("SOURCE_DIRECTORY") or os.getenv('PWD')
+local build = os.getenv("BUILD_DIRECTORY") or (source .. "/build")
+local lib_dir = os.getenv("AWESOME_LIB_DIR") or (source .. "/lib")
 
 local function escape_pattern(str)
     return string.gsub(str, "%W", "%%%1")
 end
 
 return {
-    statsfile = build .. "luacov.stats.out",
+    statsfile = build .. "/luacov.stats.out",
 
     include = {
-        escape_pattern(lib_dir) .. ".+",
+        escape_pattern(lib_dir) .. "/.+",
 
         -- For things already having the correct path
         -- (happens with integration tests)
         "^lib/",
-
-        -- For the shape-API auto-generated images
-        source and escape_pattern(source .. "lib/"),
     },
 
     -- configuration for luacov-coveralls reporter
     coveralls = {
         pathcorrect = {
-            { escape_pattern(lib_dir), "lib/"},
-            source and { escape_pattern(source .. "lib/"), "lib/" },
+            { escape_pattern(lib_dir..'/'), "lib/"},
         },
     },
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,13 @@ branches:
 
 env:
   matrix:
-    - LUA=5.2 DO_CHECKQA=1 LUANAME=lua5.2 BUILD_APIDOC=true DO_COVERAGE=coveralls
+    - LUA=5.2 DO_CHECKQA=1 LUANAME=lua5.2 DO_COVERAGE=coveralls
     # luajit: installed from source.
     - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.4 TEST_PREV_COMMITS=1
     # Note: luarocks does not work with Lua 5.0.
     - LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build
     # Lua 5.2 with older lgi and screen size not divisible by 2.
-    - LUA=5.2 LUANAME=lua5.2 LGIVER=0.8.0 TESTS_SCREEN_SIZE=1921x1079
+    - LUA=5.2 LUANAME=lua5.2 LGIVER=0.8.0 TESTS_SCREEN_SIZE=1921x1079 BUILD_APIDOC=true
     # Lua 5.3 isn't available in Ubuntu Trusty, so some magic below installs it.
     - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -168,7 +168,6 @@ script:
 
         (make check-unit-coverage \
          && do_codecov unittests \
-         && sed -i "1 s~^~require('luacov.runner')('$PWD/.luacov'); \0~" build/awesomerc.lua \
          && tests/run.sh \
          && do_codecov functionaltests)
         ret=$?
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,7 +110,7 @@ install:
     do_codecov() {
       echo "== do_codecov in $PWD: $*: build/luacov.stats.out: =="
       if [ "$DO_COVERAGE" = "codecov" ]; then
-        cat build/luacov.stats.out
+        test -f build/luacov.stats.out || { echo 'build/luacov.stats.out does not exist.'; return 1; }
         luacov || return 1
         travis_retry bash /tmp/codecov-bash -X coveragepy -c -F "$1" || return 1
         rm build/luacov.stats.out
@@ -226,6 +226,6 @@ after_success:
   - |
     set -e
     if [ "$DO_COVERAGE" = "coveralls" ]; then
-      cat build/luacov.stats.out
+      test -f build/luacov.stats.out || { echo 'build/luacov.stats.out does not exist.'; return 1; }
       luacov-coveralls --verbose --merge
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -111,7 +111,7 @@ install:
       echo "== do_codecov in $PWD: $*: build/luacov.stats.out: =="
       if [ "$DO_COVERAGE" = "codecov" ]; then
         cat build/luacov.stats.out
-        (cd build && luacov) || return 1
+        luacov || return 1
         travis_retry bash /tmp/codecov-bash -X coveragepy -c -F "$1" || return 1
         rm build/luacov.stats.out
       fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -168,7 +168,7 @@ script:
 
         (make check-unit-coverage \
          && do_codecov unittests \
-         && sed -i "1 irequire('luacov.runner')('$PWD/.luacov')" build/awesomerc.lua \
+         && sed -i "1 s~^~require('luacov.runner')('$PWD/.luacov'); \0~" build/awesomerc.lua \
          && tests/run.sh \
          && do_codecov functionaltests)
         ret=$?
--- a/.travis.yml
+++ b/.travis.yml
@@ -201,13 +201,18 @@ script:
         echo "Testing commit $commit"
         git checkout "$commit"
         git --no-pager show --stat
-        if ! make all check; then
+        if ! DO_COVERAGE=0 make all check; then
           failed="$failed $commit"
         fi
         travis_fold_end
       done
+
+      git checkout -qf FETCH_HEAD
       if [ -n "$failed" ]; then
-        echo "Checks failed for these commits:$failed"
+        echo "Checks failed for these commits:"
+        for c in $commits; do
+          git log -1 --pretty="%h %s (%an, %ad)" "$c"
+        done
         false
       fi
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,8 +90,11 @@ install:
   - if [ "$DO_CHECKQA" = 1 ]; then travis_retry sudo luarocks install luacheck; fi
 
   # Install ldoc for building docs.
-  - travis_retry sudo luarocks install ldoc
-  - travis_retry sudo luarocks install lua-discount
+  - |
+    if [ "$BUILD_APIDOC" = "true" ]; then
+      travis_retry sudo luarocks install ldoc
+      travis_retry sudo luarocks install lua-discount
+    fi
 
   # Install dependencies for code coverage testing.
   - if [ "$DO_COVERAGE" = "" ]; then export DO_COVERAGE=0; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ env:
     - secure: "R/HYDclnws1I1+v9Yjt+RKa4CsFhbBT9tiwE3EfPhEj2KCYX4sFRMxuZvLf5sq0XWdrQaPhQ54fgAZGr3f054JKRXcTB0g9J6nhSHz9kIjPh446gafUhEeDQcZRwM/MeCWiwFIkiZm6smYoDFE9JTWu6quNV+lQ4kcVDOp2ibEc="
 
 before_install:
+  - if [ "$BUILD_APIDOC" = true ] && [ -n "$DO_COVERAGE" ]; then echo "BUILD_APIDOC and DO_COVERAGE are not meant to be used together." >&2; exit 1; fi
   - if [ -z $LUAINCLUDE ]; then LUAINCLUDE=/usr/include/${LUANAME}; fi
   - if [ -z $LUALIBRARY ]; then LUALIBRARY=/usr/lib/x86_64-linux-gnu/lib${LUANAME}.so; fi
   - cmake --version
@@ -49,7 +50,7 @@ install:
   # Install Lua (per env).
   # Note that Lua 5.3 is installed manually, because it is not available in Ubuntu Trusty.
   - |
-    set -ev
+    set -e
     if [[ "$LUA" == "5.3" ]]; then
       wget http://www.lua.org/ftp/lua-5.3.3.tar.gz -O lua.tar.gz
       tar -xvzf lua.tar.gz
@@ -94,17 +95,19 @@ install:
 
   # Install dependencies for code coverage testing.
   - if [ "$DO_COVERAGE" = "" ]; then export DO_COVERAGE=0; fi
-  - if [ "$DO_COVERAGE" != "0" ]; then sudo luarocks install luacov 0.11.0-1; fi
+  - if [ "$DO_COVERAGE" != "0" ]; then sudo luarocks install luacov 0.12.0-1; fi
   - if [ "$DO_COVERAGE" = "coveralls" ]; then sudo luarocks install luacov-coveralls; fi
 
   # Determine custom version.
   - export AWESOME_VERSION="${TRAVIS_BRANCH}-g$(git rev-parse --short HEAD)"
   - 'if [ "$TRAVIS_PULL_REQUEST" != false ]; then AWESOME_VERSION="${AWESOME_VERSION}-PR${TRAVIS_PULL_REQUEST}"; fi'
   # function for codecov support
-  - travis_retry wget -O /tmp/codecov-bash https://codecov.io/bash
+  - if [ "$DO_COVERAGE" = "codecov" ]; then travis_retry wget -O /tmp/codecov-bash https://codecov.io/bash; fi
   - |
     do_codecov() {
+      echo "== do_codecov in $PWD: $*: build/luacov.stats.out: =="
       if [ "$DO_COVERAGE" = "codecov" ]; then
+        cat build/luacov.stats.out
         (cd build && luacov) || return 1
         travis_retry bash /tmp/codecov-bash -X coveragepy -c -F "$1" || return 1
         rm build/luacov.stats.out
@@ -130,9 +133,9 @@ install:
       return $result
     }
 script:
-  - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -DLUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -D DO_COVERAGE=${DO_COVERAGE} -DSTRICT_TESTS=true"
+  - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -DLUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -DSTRICT_TESTS=true"
   - |
-    set -ev
+    set -e
     if [ -n "$BUILD_IN_DIR" ]; then
       # Explicitly remove the Makefile to not build from the src dir accidentally.
       rm Makefile
@@ -147,20 +150,23 @@ script:
       travis_run_in_fold "make.install" sudo env PATH=$PATH make install
       awesome --version
     fi
-  - |
-    if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
-      do_codecov samples
-    fi
   # Run check-unit{,-coverage} and check-integration.
   - |
-    set -ev
+    set -ex
     if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
       if [ "$DO_COVERAGE" != "0" ]; then
         travis_fold_start "DO_COVERAGE"
+
+        # Run tests/examples explicitly.
+        (mkdir -p tests/examples/build \
+         && cd tests/examples/build \
+         && cmake $CMAKE_ARGS ..)
+        do_codecov samples
+
         (make check-unit-coverage \
          && do_codecov unittests \
-         && sed -i "1 i\\require('luacov.runner')('"$PWD"/.luacov')" build/awesomerc.lua \
-         && BUILD_DIRECTORY="" tests/run.sh \
+         && sed -i "1 irequire('luacov.runner')('$PWD/.luacov')" build/awesomerc.lua \
+         && tests/run.sh \
          && do_codecov functionaltests)
         ret=$?
         travis_fold_end
@@ -176,7 +182,7 @@ script:
       travis_run_in_fold "make.check-qa" make check-qa
     fi
   - |
-    set -ev
+    set -e
     if [ "$TEST_PREV_COMMITS" = 1 ] && ! [ "$TRAVIS_PULL_REQUEST" = false ]; then
       # Check each commit separately (to make git-bisect less annoying).
       # Fix Travis' commit range (https://github.com/travis-ci/travis-ci/issues/4596).
@@ -210,4 +216,9 @@ after_success:
   # Push updated API docs for relevant branches, e.g. non-PRs builds on master.
   - if [ "$BUILD_APIDOC" = "true" ]; then build-utils/travis-apidoc.sh; fi
   # Push code coverage information
-  - if [ "$DO_COVERAGE" = "coveralls" ]; then BUILD_DIRECTORY="$(pwd)/$(readlink build)/" SOURCE_DIRECTORY="$(pwd)/" luacov-coveralls --verbose --merge; fi
+  - |
+    set -e
+    if [ "$DO_COVERAGE" = "coveralls" ]; then
+      cat build/luacov.stats.out
+      luacov-coveralls --verbose --merge
+    fi
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -389,7 +389,7 @@ if(BUSTED_EXECUTABLE)
     # Keep the arguments in sync with the version below!
     add_custom_target(check-unit ALL
         ${BUSTED_EXECUTABLE} "--helper=${CMAKE_SOURCE_DIR}/spec/preload.lua"
-        "--lpath=${CMAKE_BINARY_DIR}/lib/?.lua;${CMAKE_BINARY_DIR}/lib/?/init.lua;spec/?.lua"
+        "--lpath=lib/?.lua;lib/?/init.lua;spec/?.lua"
         WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
         COMMENT "Running unit tests"
         VERBATIM)
@@ -397,9 +397,8 @@ if(BUSTED_EXECUTABLE)
 
     # Same as above, but with --coverage argument
     add_custom_target(check-unit-coverage
-        "BUILD_DIRECTORY=${CMAKE_BINARY_DIR}/"
         ${BUSTED_EXECUTABLE} "--helper=${CMAKE_SOURCE_DIR}/spec/preload.lua"
-        "--lpath=${CMAKE_BINARY_DIR}/lib/?.lua;${CMAKE_BINARY_DIR}/lib/?/init.lua;spec/?.lua"
+        "--lpath=lib/?.lua;lib/?/init.lua;spec/?.lua"
         "--coverage"
         WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
         COMMENT "Running unit tests under LuaCov"
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -12,6 +12,9 @@ option(WITH_DBUS "build with D-BUS" ON)
 option(GENERATE_MANPAGES "generate manpages" ON)
 option(COMPRESS_MANPAGES "compress manpages" ON)
 option(GENERATE_DOC "generate API documentation" ON)
+if (GENERATE_DOC AND ENV{DO_COVERAGE})
+    message(STATUS "Not generating API documentation with DO_COVERAGE")
+endif()
 
 # {{{ Endianness
 include(TestBigEndian)
@@ -37,18 +40,26 @@ a_find_program(GIT_EXECUTABLE git FALSE)
 a_find_program(ASCIIDOC_EXECUTABLE asciidoc FALSE)
 a_find_program(XMLTO_EXECUTABLE xmlto FALSE)
 a_find_program(GZIP_EXECUTABLE gzip FALSE)
-# lua documentation
-a_find_program(LDOC_EXECUTABLE ldoc FALSE)
-if(NOT LDOC_EXECUTABLE)
-    a_find_program(LDOC_EXECUTABLE ldoc.lua FALSE)
-endif()
-if(LDOC_EXECUTABLE)
-    execute_process(COMMAND sh -c "${LDOC_EXECUTABLE} --sadly-ldoc-has-no-version-option 2>&1  | grep ' vs 1.4.5'"
-                    OUTPUT_VARIABLE LDOC_VERSION_RESULT)
-    if(NOT LDOC_VERSION_RESULT STREQUAL "")
-        message(WARNING "Ignoring LDoc, because version 1.4.5 is known to be broken")
-        unset(LDOC_EXECUTABLE CACHE)
+# Lua documentation
+if(GENERATE_DOC)
+    a_find_program(LDOC_EXECUTABLE ldoc FALSE)
+    if(NOT LDOC_EXECUTABLE)
+        a_find_program(LDOC_EXECUTABLE ldoc.lua FALSE)
     endif()
+    if(LDOC_EXECUTABLE)
+        execute_process(COMMAND sh -c "${LDOC_EXECUTABLE} --sadly-ldoc-has-no-version-option 2>&1  | grep ' vs 1.4.5'"
+                        OUTPUT_VARIABLE LDOC_VERSION_RESULT)
+        if(NOT LDOC_VERSION_RESULT STREQUAL "")
+            message(WARNING "Ignoring LDoc, because version 1.4.5 is known to be broken")
+            unset(LDOC_EXECUTABLE CACHE)
+        endif()
+    endif()
+    if(NOT LDOC_EXECUTABLE)
+        message(STATUS "Not generating API documentation. Missing: ldoc.")
+        set(GENERATE_DOC OFF)
+    endif()
+else()
+    message(STATUS "Not generating API documentation.")
 endif()
 # theme graphics
 a_find_program(CONVERT_EXECUTABLE convert TRUE)
@@ -73,13 +84,6 @@ if(GENERATE_MANPAGES)
 
         message(STATUS "Not generating manpages. Missing: " ${missing})
         set(GENERATE_MANPAGES OFF)
-    endif()
-endif()
-
-if(GENERATE_DOC)
-    if(NOT LDOC_EXECUTABLE)
-        message(STATUS "Not generating API documentation. Missing: ldoc")
-        set(GENERATE_DOC OFF)
     endif()
 endif()
 # }}}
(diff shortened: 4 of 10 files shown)


Elv13:
@psychon codecov somehow fails, but looking at coverall, this PR makes the build ~4 minutes faster. Of course, its ugly. On the other hand, I think this fix the coverage line delta, but I can't be sure until codecov runs again.

psychon:
Well, the message is:
```
Could not load stats file luacov.stats.out.
Run your Lua program with -lluacov and then rerun luacov.
```
It's not codecov that fails, but luacov.

coveralls seems to have the same problem: Coverage data from `tests/examples` is missing. This is just not a fatal error for coveralls.

I think you just rediscovered the problem that https://github.com/awesomeWM/awesome/pull/964 / https://github.com/awesomeWM/awesome/issues/982 already found.

Elv13:
/me Down the rabbit hole I went, many issues I found, many issues I fixed, many issues remains.

@psychon Your patch to stop the screens for leaking is totally working. It also prevent the tags from leaking. In returns, on my local computer, it does fully fix the exponential memory usage. It doesn't fix it when the coverage is enabled. Therefor I blame luacov for it. The more code is executed, the more memory luacov takes. I propose to close both slow CI related issues as they are not caused by Awesome.

Also, coveralls isn't happy at all to lose the test example coverage. CodeCov don't care much because of some miracle circumstances where the number of lines removed is exactly proportional to the number of misses remove due to the coverage being correct (WTF). The CodeCov lines are now correct, the offset caused by the example tests is now corrected. I added an attempted fix for Coveralls, but it doesn't help, so I leave this to someone who knows more about it.

* The coverage no longer include the tests
* The build is 3.5-4 minutes faster by trimming some fat where we can
* The screens and tags are no longer leaking in multi screen tests
* The coverage data no longer have an offset due to the CMake `configure_file` insertions
* Work has been removed from the slower builds and added to the faster ones
* The investigation why 2 of the builds are very slow is completed
* Open issues have been fixed

Other possible speed holes:

 * Disable the wibar on the client multi-screen tests, it should trim 45s
 * Install less dependencies TBD
 * Don't install/use ldoc on the slower targets (save the install time of penlight and ldoc, save 45s)
 * Create our own debian repository and stop pulling from git and luarock all the time (save 1.5 minute)
 * Explore using Docker.io images on Travis and see its smart enough to cache the first few build steps (replace the 3 previous bullet points, save 3 minutes)

blueyed:
What needs to be done here?
"coverage/coveralls" complains about coverage loss: -8.3%.

psychon:
Should I take a look before the release? I thought the priorities were different.

blueyed:
I don't think it is a release blocker - I came across it when waiting for Travis, and remembered that this is supposed to make it faster.

.travis.yml
```diff
@@ -89,8 +89,11 @@ install:
   - if [ "$DO_QUECKQA" = 1 ]; then travis_retry sudo luarocks install luacheck; fi
 
   # Install ldoc for building docs.
-  - travis_retry sudo luarocks install ldoc 1.4.4
-  - travis_retry sudo luarocks install lua-discount
+  - |
+    if [ "$DO_COVERAGE" == "0" || "$DO_COVERAGE" == "" ]; then
```
blueyed: Should check for `$BUILD_APIDOC` here instead?!

awesomeConfig.cmake
```diff
@@ -77,7 +77,7 @@ if(GENERATE_MANPAGES)
 endif()
 
 if(GENERATE_DOC)
-    if(NOT LDOC_EXECUTABLE)
+    if(NOT LDOC_EXECUTABLE OR DO_COVERAGE)
```
blueyed: Should not check for DO_COVERAGE here and above: the build should (and does) not install ldoc instead.

.travis.yml
```diff
@@ -149,16 +156,16 @@ script:
...
-    fi
   # Run check-unit{,-coverage} and check-integration.
   - |
     set -ev
     if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
       if [ "$DO_COVERAGE" != "0" ]; then
         travis_fold_start "DO_COVERAGE"
+
+        (mkdir -p tests/examples/build && cd tests/examples/build && cmake $CMAKE_ARGS -DUSE_LCOV=1 "..")
+        do_codecov samples tests/examples/build
```
blueyed: Maybe there is a problem with relative paths (in `luacov.stats.out`, which gets copied).
What about `cd`ing here instead?!

.travis.yml
```diff
@@ -10,9 +10,9 @@ branches:
 
 env:
   matrix:
-    - LUA=5.2 DO_CHECKQA=1 LUANAME=lua5.2 BUILD_APIDOC=true DO_COVERAGE=coveralls
+    - LUA=5.2 DO_CHECKQA=1 LUANAME=lua5.2 DO_COVERAGE=coveralls
     # luajit: installed from source.
-    - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.4 TEST_PREV_COMMITS=1
+    - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so BUILD_APIDOC=true LUAROCKS_ARGS=--lua-suffix=jit-2.0.4 TEST_PREV_COMMITS=1
```
blueyed: Might easily be the slowest though, because of TEST_PREV_COMMITS.

awesomeConfig.cmake
```diff
@@ -329,19 +328,25 @@ if(GENERATE_DOC)
 endif()
 
 # {{{ Configure files
```
blueyed: Changes here are hard to understand.
Should have some comment(s), and I dislike using "DO_COVERAGE" here, but it cannot be avoided probably.

awesomeConfig.cmake
```diff
@@ -352,6 +357,27 @@ foreach(file ${AWESOME_CONFIGURE_FILES})
                    ESCAPE_QUOTES
                    @ONLY)
 endforeach()
+
+set(AWESOME_CONFIGURE_NO_COV_FILES
+    ${awesome_c_configure_files}
+    ${awesome_lua_configure_files}
+)
+
+if(NOT DO_COVERAGE)
```
blueyed: Switch the `if` blocks to have the positive one first.
Also a comment here would help.

awesomeConfig.cmake
```diff
@@ -352,6 +357,27 @@ foreach(file ${AWESOME_CONFIGURE_FILES})
                    ESCAPE_QUOTES
                    @ONLY)
 endforeach()
+
+set(AWESOME_CONFIGURE_NO_COV_FILES
```
blueyed: Better name: `AWESOME_CONFIGURE_COPY_ONLY_FILES`?

awesomeConfig.cmake
```diff
@@ -376,6 +376,12 @@ else()
                     ${BUILD_DIR}/${file}
                     COPYONLY)
     endforeach()
+
+    # There is no way to avoid that one
```
blueyed: What for / why?
Could not find "@@" in there.

And if required, this part could be moved into a separate file maybe?!

.travis.yml
psychon: Then I propose to drop `TEST_PREV_COMMITS`. The idea is nice, but its slow down is a problem for larger pull requests.

blueyed: Seems to be not working as expected currently anyway: looks like it tests only one of them.

> its slow down is a problem for larger pull requests.

Well, I think that is OK, given the benefit of ensuring that git-bisect works (better).
And given that it uses the existing build (cache), it is probably not so slow after all.

My point in the comment above was just that this should be taken into account when shuffling the build matrix.

awesomeConfig.cmake
```diff
@@ -352,6 +357,33 @@ foreach(file ${AWESOME_CONFIGURE_FILES})
...
+        configure_file(${SOURCE_DIR}/${file}
+                    ${BUILD_DIR}/${file}
+                    COPYONLY)
+    endforeach()
+
+    # There is no way to avoid that one
+    configure_file(${SOURCE_DIR}/lib/awful/util.lua
+                    ${BUILD_DIR}/lib/awful/util.lua
+                    ESCAPE_QUOTES
+                    @ONLY)
```
blueyed: What is it for / why can't it be avoided?
Could not find "@@" in there.

And if required, this part could be moved into a separate file maybe?!

blueyed: Ah:
```lua
--- Get the path to the directory where themes are installed.
-- @return A string with the requested path with a slash at the end.
function util.get_themes_dir()
    return "@AWESOME_THEMES_PATH@" .. "/"
end

--- Get the path to the directory where our icons are installed.
-- @return A string with the requested path with a slash at the end.
function util.get_awesome_icon_dir()
    return "@AWESOME_ICON_PATH@" .. "/"
end
```

Should/can we move this to a separate file?
Or is it not too much of a hack to handle the whole of `util.lua` in a special way here?

blueyed:
As per the discussion in https://github.com/awesomeWM/awesome/pull/1238#issuecomment-269131480 we might want to remove coveralls (or ignore it for now?!).
But I've left a comment above which might explain it (and we could `cd` instead).
Will look into it later.

Elv13:
Thanks for taking a look. I guess you now know why I put that image at the top. Fixing this has been only frustration for @psychon and me for a while now... Do you think Coveralls is worth all that trouble? I got it to work fine with CodeCov. IMO your time is better spent elsewhere than trying yet more changes. My own commits were already rebased a hundred time to get to this point.

blueyed:
I'll poke a bit more.
It seems like the current luacov failed to work with codecov also (that's why it was pinned): it was not generating coverage output for the functional tests: https://travis-ci.org/awesomeWM/awesome/jobs/186857911#L4089
9abe68c hopefully fixes this.

I agree that we do not really need coveralls also, but it's very nice to compare the coverage from two different tools / processors though, to see if something goes wrong.

blueyed:
@Elv13 @psychon 
It worked out finally..
Will clean it up a bit more now.

CMakeLists.txt
```diff
@@ -388,17 +388,16 @@ if(BUSTED_EXECUTABLE)
...
         COMMENT "Running unit tests"
         VERBATIM)
     list(APPEND CHECK_TARGETS check-unit)
 
     # Same as above, but with --coverage argument
     add_custom_target(check-unit-coverage
-        "BUILD_DIRECTORY=${CMAKE_BINARY_DIR}/"
         ${BUSTED_EXECUTABLE} "--helper=${CMAKE_SOURCE_DIR}/spec/preload.lua"
-        "--lpath=${CMAKE_BINARY_DIR}/lib/?.lua;${CMAKE_BINARY_DIR}/lib/?/init.lua;spec/?.lua"
+        "--lpath=lib/?.lua;lib/?/init.lua;spec/?.lua"
```
blueyed: This might be wrong, but is the easiest way to get the coverage from there.
It can be done by setting `SOURCE_DIRECTORY` though instead (probably), or rather by adjusting `.luacov` for it.

.luacov
```diff
@@ -20,12 +20,20 @@ return {
         -- (happens with integration tests)
         "^lib/",
         "^awesomerc.lua",
+
+        -- Include awesomerc.lua (in build dir with integration tests).
+        escape_pattern(build) .. "/awesomerc.lua",
```
blueyed: This might not be required when using a relative path for the RC file (awesome gets run from the build dir already)?!

blueyed: But `awesomerc.lua` is where the `luacov.runner` gets added, so that might be irrelevant..

blueyed:
awesomewmrc.lua gets not included.
I will remove it from this PR for now.

Elv13:
👍

tests/examples/CMakeLists.txt
```diff
@@ -14,22 +14,44 @@ cmake_minimum_required(VERSION 3.0.0)
...
+${SOURCE_DIR}/tests/examples/shims/?;"
 )
 
+# Use the example testing as coverage source.
+if (USE_LCOV)
+    set(LCOV_PATH ${SOURCE_DIR}/.luacov)
+
+    if (NOT EXISTS "${LCOV_PATH}")
+        message(WARNING "The luacov configuration file was not found")
+        unset(LCOV_PATH)
```
psychon: Why a warning instead of an error?

tests/examples/CMakeLists.txt
```diff
@@ -14,22 +14,44 @@ cmake_minimum_required(VERSION 3.0.0)
...
+if(NOT SOURCE_DIR AND ${CMAKE_CURRENT_SOURCE_DIR} MATCHES "/tests/examples")
+    set(SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../")
+
+    # .luacov use these variables
+    set(ENV{BUILD_DIRECTORY} ${CMAKE_CURRENT_BINARY_DIR}/)
+    set(ENV{AWESOME_LIB_DIR} ${SOURCE_DIR}/lib/)
+    set(ENV{SOURCE_DIRECTORY} ${SOURCE_DIR})
+else()
+    set(SOURCE_DIR ${CMAKE_SOURCE_DIR})
+endif()
```
psychon: "if `SOURCE_DIR` is set, overwrite it". That seems odd to me. How about giving this variable a different name instead (dunno, `TOP_SOURCE_DIR`)?

tests/examples/CMakeLists.txt
```diff
@@ -14,22 +14,44 @@ cmake_minimum_required(VERSION 3.0.0)
 # Get and update the LUA_PATH so the scripts can be executed without Awesome.
 execute_process(COMMAND lua -e print\(package.path\) OUTPUT_VARIABLE "LUA_PATH_")
```
psychon: For the commit message: How about some more details. The commit message does not even say that this commit fixes running the doc tests without compiling. I'd like to know what exactly was broken and was fixed (and perhaps also that "running the doc tests without compiling" means running `cmake` in the `tests/examples` directory (does it mean that?))

awesomeConfig.cmake
psychon: So... this can be removed?

awesomeConfig.cmake
```diff
@@ -355,6 +360,33 @@ foreach(file ${AWESOME_CONFIGURE_FILES})
...
+else()
+    foreach(file ${AWESOME_CONFIGURE_COPYONLY_WITHCOV_FILES})
+        configure_file(${SOURCE_DIR}/${file}
+                    ${BUILD_DIR}/${file}
+                    ESCAPE_QUOTES
+                    @ONLY)
+    endforeach()
+endif()
+
```
psychon: Sigh. More bad hacks and band-aids. My long-term plan is still to change things so that bug reports with line numbers are useful again, but for now if this hack makes you happy...

.travis.yml
```diff
@@ -147,16 +151,16 @@ script:
...
   - |
     set -ev
     if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
       if [ "$DO_COVERAGE" != "0" ]; then
         travis_fold_start "DO_COVERAGE"
+
+        (mkdir -p tests/examples/build && cd tests/examples/build && cmake $CMAKE_ARGS -DUSE_LCOV=1 "..")
+        do_codecov samples tests/examples/build
+
         (make check-unit-coverage \
```
psychon: Could someone please explain this commit? For example in the commit message? I don't like the copy of `luacov.stats.out` (why??), I don't like most of it.

.travis.yml
```diff
@@ -89,8 +89,11 @@ install:
   - if [ "$DO_CHECKQA" = 1 ]; then travis_retry sudo luarocks install luacheck; fi
 
   # Install ldoc for building docs.
-  - travis_retry sudo luarocks install ldoc 1.4.4
-  - travis_retry sudo luarocks install lua-discount
+  - |
+    if [ "$BUILD_APIDOC" == "true" ]; then
+      travis_retry sudo luarocks install ldoc 1.4.4
+      travis_retry sudo luarocks install lua-discount
+    fi
```
psychon: As Daniel already pointed out: Doesn't this make changes in the CMake files from earlier commits obsolete?

.luacov
```diff
@@ -1,33 +1,29 @@
 -- Configuration file for LuaCov
 
 -- This variable is set externally
-local build = os.getenv("BUILD_DIRECTORY") or error("$BUILD_DIRECTORY not set")
-local lib_dir = os.getenv("AWESOME_LIB_DIR") or build .. "lib/"
-local source = os.getenv("SOURCE_DIRECTORY")
+local source = os.getenv("SOURCE_DIRECTORY") or os.getenv('PWD')
+local build = os.getenv("BUILD_DIRECTORY") or (source .. "/build")
+local lib_dir = os.getenv("AWESOME_LIB_DIR") or (source .. "/lib")
```
psychon: According to the commit message, this should be more than one commit. I have a hard time following all of this. Why the switch to a different version of luacov? Should that perhaps just be unpinned instead? What does `set -v` do and what is its connection with this PR?

.travis.yml
```diff
@@ -92,7 +92,7 @@ install:
   # Install ldoc for building docs.
   - |
     if [ "$BUILD_APIDOC" == "true" ]; then
-      travis_retry sudo luarocks install ldoc 1.4.4
+      travis_retry sudo luarocks install ldoc 1.4.6
```
psychon: How does this help with speed and accuracy? Why is ldoc still pinned to a specific version?

awesomeConfig.cmake
blueyed: In this case line numbers should match the source?!

tests/examples/CMakeLists.txt
blueyed: It is not meant to be overwritten, I guess.
But another name makes sense!

awesomeConfig.cmake
psychon: The line numbers that the coverage tool sees, yes. But line numbers from bug reports will still be different from what we have in Git.

.travis.yml
blueyed: It was pinned for some workaround - mentioned in the commit message.
It can be upgraded now, but is still pinned to avoid unexpected changes.

.luacov
blueyed: Yeah, I ended up squashing everything.
And there are reverts for previous commits (from this PR) in there..
I could look into cleaning that up again.

`set -v` prints the executed commands, but less verbose than `-x`. Travis seems to do that already / and/or it annoyed me.

blueyed: Do you mean ldoc instead of luacov?

tests/examples/CMakeLists.txt
blueyed: This part has been removed by me later.. should amend the previous commit here..

awesomeConfig.cmake
blueyed: This part was changed by me, too.
Should amend the previous commit.

blueyed:
@psychon 
Thanks for the review.
You might want to look at the overall diff again for now, and not the individual commits.

tests/_multi_screen.lua
```diff
@@ -411,6 +411,7 @@ local function add_steps(real_steps, new_steps)
     for _, dispo in ipairs(dispositions) do
         -- Cleanup
         table.insert(real_steps, function()
+            collectgarbage("collect")
             if #client.get() == 0 then return true end
```
blueyed: Pulled out here, and changed accordingly: https://github.com/awesomeWM/awesome/pull/1333.

.luacov
blueyed: Oh..

>      - Use latest luacov (0.12.0-1) again                                                                            
>       This reverts parts of 4cc6a815.                                                    
>       I think it is better to fix any failure that 4cc6a815 tried to work around.

.travis.yml
blueyed: I've thought of luacov with my previous comment.
Pulled out the update of ldoc into https://github.com/awesomeWM/awesome/pull/1334.

awesomeConfig.cmake
blueyed: Ah, of course.
I agree that we should get back to that state in the long run.

blueyed:
Updated/cleaned it a bit.

I suggest squashing the commits from @Elv13 into my big one.

Elv13:
> I suggest squashing the commits from @Elv13 into my big one.

Do as you wish

Elv13:
Ping? Also, I don't mind if you just redo the commits and remove me as the author (and merge your changes with mines) so the commits are nicer. I also don't mind if this is squashed into a giant commit.

tests/run.sh
```diff
@@ -23,7 +23,7 @@ fi
 # Change to file's dir (POSIXly).
 cd -P -- "$(dirname -- "$0")"
 this_dir=$PWD
-source_dir="$PWD/.."
+source_dir="${this_dir%/*}"
```
psychon: This assumes that `$this_dir` (thus `$PWD`) contains a slash (and it was not all that easy to figure out what this does). I'm not really sure if this always works correctly...?

tests/examples/CMakeLists.txt
```diff
@@ -14,20 +14,45 @@ cmake_minimum_required(VERSION 3.0.0)
...
 
+# Make sure the system can be called from the test directory
+if(NOT SOURCE_DIR AND ${CMAKE_CURRENT_SOURCE_DIR} MATCHES "/tests/examples")
+    get_filename_component(SOURCE_DIR
+                           "${CMAKE_CURRENT_SOURCE_DIR}/../.." ABSOLUTE)
+    # Used by .luacov.
+    set(ENV{SOURCE_DIRECTORY} ${SOURCE_DIR})
+else()
+    set(SOURCE_DIR ${CMAKE_SOURCE_DIR})
```
psychon: Still overwrites `$SOURCE_DIR`

tests/examples/CMakeLists.txt
```diff
@@ -14,20 +14,45 @@ cmake_minimum_required(VERSION 3.0.0)
...
+    OUTPUT_VARIABLE TEST_OUTPUT
+    ERROR_VARIABLE  TEST_ERROR)
+  if (TEST_ERROR EQUAL 0)
+    set(ENV{DO_COVERAGE} 1)
+  else()
+    message(WARNING "Failed to run luacov.runner, skipping DO_COVERAGE.")
+    message("${TEST_ERROR}")
+    set(ENV{DO_COVERAGE} 0)
+  endif()
+endif()
```
psychon: ....why? I would just let it fail instead instead of semi-silently just not generating coverage data.

awesomeConfig.cmake
```diff
@@ -332,19 +335,25 @@ if(GENERATE_DOC)
 endif()
 
 # {{{ Configure files
-file(GLOB awesome_c_configure_files RELATIVE ${SOURCE_DIR}
+file(GLOB awesome_base_c_configure_files RELATIVE ${SOURCE_DIR}
     ${SOURCE_DIR}/*.c
-    ${SOURCE_DIR}/*.h
+    ${SOURCE_DIR}/*.h)
+
+file(GLOB awesome_c_configure_files RELATIVE ${SOURCE_DIR}
```
psychon: Why split this up in two? We do not generate coverage information for the C code anyway (should we?).

psychon:
Codecov currently [says](https://github.com/awesomeWM/awesome/pull/1239#issuecomment-267736185):
```
@@             master      #1239   diff @@
==========================================
  Files           242        116    -126   
  Lines         16511      12107   -4404   
  Methods           0          0           
  Messages          0          0           
  Branches          0          0           
==========================================
- Hits          12072       8837   -3235   
+ Misses         4439       3270   -1169   
  Partials          0          0           
```
Why do 126 files go missing? (And I guess someone will tell me the 4404 files are documentation, but even including the DOC_HIDE lines, templates and CMake files, we only have 3955 such lines according to `cat $(find tests/examples -type f) | wc -l`).

tests/run.sh
blueyed: Yes, I think "$PWD" will always be absolute, and therefore contain a slash.

awesomeConfig.cmake
blueyed: Yes, we should eventually.. currently it makes no sense to split them, but it's also not a problem really.  I'll leave it as-is for now.

blueyed:
About the missing files: in master, `tests` and `themes/xresources/assets.lua` is also included: https://codecov.io/gh/awesomeWM/awesome/branch/master.

I will bring them back.. we should probably just include everything from our build and source dirs.  Would fix https://github.com/awesomeWM/awesome/issues/1315 then also.

blueyed:
FWIW, there was some timeout just now with the coveralls build:
```
2017-01-03 16:41:24 W: awesome: a_glib_poll:389: Last main loop iteration took 1.272816 seconds! Increasing limit for this warning to that value.
2017-01-03 16:41:28 W: awesome: a_glib_poll:389: Last main loop iteration took 3.037822 seconds! Increasing limit for this warning to that value.
2017-01-03 16:41:44 W: awesome: a_glib_poll:389: Last main loop iteration took 3.144219 seconds! Increasing limit for this warning to that value.
2017-01-03 16:41:57 W: awesome: a_glib_poll:389: Last main loop iteration took 3.237561 seconds! Increasing limit for this warning to that value.
2017-01-03 16:42:35 W: awesome: a_glib_poll:389: Last main loop iteration took 4.407701 seconds! Increasing limit for this warning to that value.
2017-01-03 16:43:12 W: awesome: a_glib_poll:389: Last main loop iteration took 6.451608 seconds! Increasing limit for this warning to that value.
2017-01-03 16:44:07 W: awesome: a_glib_poll:389: Last main loop iteration took 7.415468 seconds! Increasing limit for this warning to that value.
Awesome was killed due to timeout after 180 seconds
===> ERROR running /home/travis/build/awesomeWM/awesome/tests/test-awful-client.lua <===
```

Elv13:
> FWIW, there was some timeout just now with the coveralls build:

They happen from time to time. This PR doesn't touch the integration tests, so it probably isn't to blame.

blueyed:
Hmm, luacov did not update its comment?!

Elv13:
> Why do 126 files go missing?

@psychon #964

psychon:
So... we make @blueyed clean up the commits and then this gets merged?

Elv13:
Yes, once the regression from the last fixup is fixed. Namely only 2 of the 5 targets remains ( see https://travis-ci.org/awesomeWM/awesome/builds/188582885 ).

Elv13:
@blueyed Is it possible that your last changes made it show coverage for the tests again? The -126 files is gone from the new coverage. This was done on purpose to fix #964.

blueyed:
The TEST commit will be removed, of course.
What is bad about covering everything, especially the tests itself?
I wonder about things like https://codecov.io/gh/awesomeWM/awesome/src/8af82ef7fd1430b701eaf1fbc49e317070871305/tests/examples/shims/mouse.lua though, which should be covered 100% shouldn't it?

Elv13:
> What is bad about covering everything, especially the tests itself?

I personally don't mind. @psychon said otherwise at some point. I also agree that the coverage percentage with the tests is a little inflated and misrepresent the reality. As for the link you provided, the shims are not expected to have full coverage. They mimic CAPI and are fixed only when a function is required. Some other documentation examples (such as the prompt ones) are also not fully covered and are not intended to be[1].

[1] It would be cool, but is impractical and not worth the effort.

blueyed:
@Elv13 
> As for the link you provided, the shims are not expected to have full coverage. They mimic CAPI and are fixed only when a function is required.

But shouldn't line 1 there be covered (`local screen = require("screen")`)?!
Also only the function body is covered, but not the function definition..
(is this really expected like that)

blueyed:
Similar: https://codecov.io/gh/awesomeWM/awesome/src/8af82ef7fd1430b701eaf1fbc49e317070871305/tests/examples/shims/_common_template.lua - if I add some `error()` there, it gets executed with `DO_COVERAGE=1 cmake tests/examples` certainly.
So it looks like an issue with `luacov` after all?!

I am getting the following in `luacov.stats.out`:

> 24:/home/user/src/awesome/tests/examples/shims/_common_template.lua
0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 153 151

.travis.yml
```diff
@@ -161,7 +161,7 @@ script:
 
         (make check-unit-coverage \
          && do_codecov unittests \
-         && sed "1 s~^~require('luacov.runner')('$PWD/.luacov'); \0~" build/awesomerc.lua \
+         && sed -i "1 s~^~require('luacov.runner')('$PWD/.luacov'); \0~" build/awesomerc.lua \
```
Elv13: The "original line numbers" would require to use `../awesomerc.lua`. `build/awesomerc.lua` have been modified by the build system to remove the section markers used when generating the documentation.

blueyed: Ok.  I will change tests/run.sh to handle this (instead of here), which would allow for `DO_COVERAGE=1 make check` to handle it there.

blueyed:
Looking good by now again.

Codecov has 252 files covered (only `preload.lua` in `spec/`, because `luacov` could not `stat` the relative `spec/` files).  This should be fixed in the commit I've just pushed.
Coveralls has 268 files covered (16 additional files in `spec/`).

blueyed:
Rebased.
Coverage increased to 78.99% on both.
I'll merge this tonight probably.

Elv13:
It's still a bit sad that all the speedups went away. The first build after both coveralls and codecov was fixed was in the 8.x minutes range. Now, there is "5% more coverage gamification tokens", but its 11m long. Just as the coverage for the C code, I will have to ask: is there real value from having the test covered. Is it worth waiting 2:30m more for each builds? @psychon, what do you think about this?

I am not saying this PR isn't a great job: it is. Thanks a lot for it, I know how long it took you to fix all the corner case my version had. It's just IMO that the last few changes went in favor of the "accuracy" part but did hurt the "speed" part a lot.

blueyed:
Well, it is only reporting more coverage, i.e. not excluding it - the example tests are/should run with coverage, and therefore also the tests are covered.
AFAIK once coverage is enabled, it will be tracked across all files, doesn't it?

The initial 8m runs seem to not have generated coverage after all?!
I could try another build with adding the excludes for tests/ again, but this probably not what you've meant?!

Elv13:
> The initial 8m runs seem to not have generated coverage after all?!

Codecov was correctly generated (with only the `lib/` file, that was done on purpose) and took 8:35m. When you initially fixed coveralls, it was in the same ballpark too.

> AFAIK once coverage is enabled, it will be tracked across all files, doesn't it?

I see the logic in that, but if codecov was getting the correct (as in, the same coverage numbers as coveralls in other unrelated PRs)percentage in 8:35m (average over 5 runs), then why is it always at 9:45m (+/- 30 sec) right now? This isn't due to bad luck with the Travis nodes, not this many time in a row.

If by some luck going back to 188 (at 73%) covered files put it back below 9 minutes, then I argue it is worth more than having 268 files covered (the tests & shims) and a shiny 79% coverage badge.

Elv13:
Run1: coveralls 9:20, codecov 8:57

(I will update this post)

.travis.yml
```diff
@@ -226,6 +225,5 @@ after_success:
   - |
     set -e
     if [ "$DO_COVERAGE" = "coveralls" ]; then
-      cat build/luacov.stats.out
```
blueyed: Will change it to `test -f`, since this is where missing coverage info used to bail out.

blueyed:
@Elv13 
Thanks.
What have you tried?  Changing the `include` in `.luacov`?

blueyed:
@Elv13
I've tried `time (DO_COVERAGE=1 cmake ..)` in `tests/examples/build` with and without including lib/ and awesomerc.lua.  And it was actually a bit slower when not including everything (although it should have less (disk) cache misses):

Full (1st)
43.24s user 0.57s system 99% cpu 43.995 total

Only lib/ + awesomerc.lua (2nd)
45.52s user 0.63s system 99% cpu 46.373 total

I think we should address performance/speed of the builds in a new PR, based on the suggestions you've provided already, e.g. by using (pre-built) Docker base images, which have all the deps installed already.

blueyed:
🎆

Issue #1242, opened 2016-12-01
Travis builds fail


psychon:
```
$ wget https://github.com/lua/lua/releases/download/5.3.3/lua-5.3.3.tar.gz 
--2016-12-01 15:47:00--  https://github.com/lua/lua/releases/download/5.3.3/lua-5.3.3.tar.gz
Auflösen des Hostnamens »github.com (github.com)« … 192.30.253.112, 192.30.253.113
Verbindungsaufbau zu github.com (github.com)|192.30.253.112|:443 … verbunden.
HTTP-Anforderung gesendet, auf Antwort wird gewartet … 404 Not Found
2016-12-01 15:47:01 FEHLER 404: Not Found.
```
Due to commit cf96296f2795a0932ff227d074e3a81c1aa771d2, I'll leave this one for @blueyed to decide. (The tag is now called `5-3-3` instead of `5.3.3`; the only working URL that I found is  https://github.com/lua/lua/archive/v5-3-3.tar.gz ).

This caused a failed travis build for #1240 (and every coming build).

blueyed:
https://github.com/awesomeWM/awesome/pull/1243

psychon:
Thanks! :-)

psychon:
Again: Thank you. You managed to fix this before I even got home and had time to take a look!

PR #1243, opened 2016-12-01
Travis: fix URL for Lua 5.3.3

blueyed:
Fixes https://github.com/awesomeWM/awesome/issues/1242.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,9 +51,9 @@ install:
   - |
     set -ev
     if [[ "$LUA" == "5.3" ]]; then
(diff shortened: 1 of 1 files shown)


psychon:
Oh, wow. What the heck? Those "new" tarballs do not contain the same things that a normal Lua tarball would contain. It looks like they only have the `src/`-directory.

blueyed:
Yes, seems to be a source snapshot.
I think the releases where actually mirrored before.  The GitHub repo might have changed in that regard, and maybe is not just a mirror anymore?!

Let's revert https://github.com/awesomeWM/awesome/commit/cf96296f2795a0932ff227d074e3a81c1aa771d2 then.

PR #1250, opened 2016-12-02
Make the resize_to_mouse example test deterministic

psychon:
This test has a list of "things" that should be present in the output. This
table is iterated over via pairs(), which means that the output is
non-deterministic and the order of the entries is basically random.

Fix this by using ipairs() to get a deterministic iteration order. This requires
some slight change to the table that is iterated over.

This handles/fixes the issue identified at https://github.com/awesomeWM/awesome/issues/834#issuecomment-264448537.


blueyed:
Thanks a lot!

PR #1251, opened 2016-12-02
More deterministic test output

psychon:
Another part-fix for a problem tracked in #834.


Elv13:
Could the example be copied into the ldoc instead of removed? I don't mind if ti isn't unit tested, but there is still some value that would be loss beyond that.

psychon:
I removed the commit which removes `tests/examples/text/gears/object/signal.lua` from this PR.

blueyed:
Thanks!

PR #1247, opened 2016-12-02
doc: Document the client layouts.

Elv13:
This commit doesn't add any useful documentation, but adds
previously hidden documentation variables. It can be the basis
of a better layout documentation.

Fix #1246

PR #1252, opened 2016-12-02
doc: awful.tag.layouts moved to awful.layout.layouts

blueyed:
Might conflict with https://github.com/awesomeWM/awesome/pull/1247 already?!

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -33,7 +33,33 @@ local layout = {}
 -- @fixme Add documentation on available layouts as all of them are hidden
 layout.suit = require("awful.layout.suit")
 
--- The default list of layouts
+--- The default list of layouts.  Order matters.
+--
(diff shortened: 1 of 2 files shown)


psychon:
So.. .should #1247 be merged before we look at this one, or what?

Elv13:
Another question worth thinking about before merging either one. Should this says in the tag documentation? Should the whole layout doc just be in the tag page? The extra info created by #1247 is still rather useless as a whole documentation page.

 + The tag already has the master_width_factor/column_count/master_count/fill_policies
 + The tag page have the information how to set a layout
 + The layout page (currently) has only 2 useful function, the others are intended for internal uses
 - A tag and a layout are not the same concepts (a tag has a layout, not the other way around)
 - It would require another of those confusing alias pages

I am mostly neutral on this. Either can make sense if work is done to improve xor integrate it.

actionless:
#1247 is already merged, what should be the faith of this one?

PR #1253, opened 2016-12-02
cmake: Check dependencies one by one

Elv13:
It was currently done all at once. Theirfor, the error message was:

CMake Error at /usr/share/cmake/Modules/FindPkgConfig.cmake:415 (message):
  A required package was not found

With no further information beside the full dependency list. With this
change, it will print the missing package.

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -115,7 +115,7 @@ endif()
 pkg_check_modules(AWESOME_COMMON_REQUIRED REQUIRED
     xcb>=1.6)
 
-pkg_check_modules(AWESOME_REQUIRED REQUIRED
+set(AWESOME_DEPENDENCIES
     glib-2.0
     gdk-pixbuf-2.0
     cairo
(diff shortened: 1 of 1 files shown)


awesomeConfig.cmake
```diff
@@ -137,11 +137,20 @@ pkg_check_modules(AWESOME_REQUIRED REQUIRED
...
+foreach(dependency ${AWESOME_DEPENDENCIES})
+    pkg_check_modules(TMP_DEP REQUIRED ${dependency})
+
+    if(NOT TMP_DEP_FOUND)
+        message(FATAL_ERROR)
+    endif()
+endforeach()
+
+# Do it again, but this time with the CFLAGS/LDFLAGS
+pkg_check_modules(AWESOME_REQUIRED REQUIRED ${AWESOME_DEPENDENCIES})
```
blueyed: Nice, thanks!
I assume that it lists all missing deps?
And in the end you would get an error for all of them when one fails with CFLAGS/LDFLAGS?

blueyed:
This should get reverted, see https://github.com/awesomeWM/awesome/issues/1920.

Issue #1249, opened 2016-12-02
The piechart-widget draws its "pie pieces" in a non-deterministic order.


psychon:
I'll copy from https://github.com/awesomeWM/awesome/issues/834#issuecomment-264442256

> And here are the two images:
> https://github.com/awesomeWM/apidoc/blob/9963bf794e719c6937d3bd342ec78cc66b1cdb3b/images/AUTOGEN_wibox_widget_piechart_label.svg
> https://github.com/awesomeWM/apidoc/blob/21583d2133edab6adf6dfce7fb5105d9e16d6233/images/AUTOGEN_wibox_widget_piechart_label.svg
> 
> As you can see, the order in which the piechart draws things changes non-deterministically. And that's a serious bug. This means that a redraw might not actually work!
> Imagine if only half the widget is redrawn, but the widget now decides to change the order of things. The result will be broken.

So I think the piechart widget at least needs a change to get a deterministic iteration order of `self._private.data`. It would also be nice to expose an API so that the data can be given to the widget in a deterministic order, i.e. "first display 'foo' with 33%, then display 'bar' with 67% and finally 'foobar' with 33%". As far I see, that's not yet possible either.

- [ ] piechart needs deterministic iteration order
- [ ] piechart should provide API for "working with" this deterministic order.

psychon:
Assigning to @Elv13 under the assumption that he would want to work on this. Feel free to un-assign if you don't want this.

psychon:
This issue seems to affect all tests for the piechart widget:
![changes](https://cloud.githubusercontent.com/assets/89482/20834710/fcb69228-b896-11e6-9c00-2f5132799bb6.png)

Elv13:
Already discussed in #897. It's worth fixing, but it isn't really an user facing issues. The widget don't flicker and/or change at runtime unless you add/remove pies

PR #1258, opened 2016-12-07
Make the piechart deterministic

psychon:
Previously, the API to set the data that should be displayed was
:set_data(t) where t is a table. This table has the labels to use as its
keys and the numbers as its values. With this API, it was not possible
to influence the order in which the "pie pieces" were drawn.

This commit adds and uses a new API called :set_data_list(t). Here, t is
a table with integer keys and tables as values, thus one can iterate
over this with ipairs() and the order is well-defined. The tables used
as values contain the label as their first entry and the number as their
second entry.

Fixes: https://github.com/awesomeWM/awesome/issues/1249

-----

The WIP-part of this is that it makes the API a bit ugly. Can we break the API of this widget and remove the old ".data" property and replace it with what I am calling ".data_list" right now? If we can, do we want to do that?

### DIFF
--- a/lib/wibox/widget/piechart.lua
+++ b/lib/wibox/widget/piechart.lua
@@ -51,18 +51,18 @@ end
 
 local function compute_sum(data)
     local ret = 0
-    for _,v in pairs(data) do
-        ret = ret + v
+    for _, entry in ipairs(data) do
+        ret = ret + entry[2]
     end
 
     return ret
 end
 
 local function draw(self, _, cr, width, height)
-    if not self._private.data then return end
+    if not self._private.data_list then return end
 
     local radius = (height > width and width or height) / 4
-    local sum, start, count = compute_sum(self._private.data),0,0
+    local sum, start, count = compute_sum(self._private.data_list),0,0
     local has_label = self._private.display_labels ~= false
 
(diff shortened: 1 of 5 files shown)


Elv13:
Thanks for this. I am fine with this PR. As I said before, the issue was known when this piechart widget was merged (I used it as an example in one of the non deterministic doc discussion). I was under the impression the API was simpler and the issue negligible, but obviously I was wrong with this. Given the recent work in the deterministic doc, you made enough progress to proved me wrong: it can be done with limited effort.

So go ahead and merge this. As far as deleting `.data`, up to you. I think it is a simpler, more user friendly API. The bug don't really happen between redraws, so isn't visible to the average user, but your patch totally fix it anyway. So removing `.data` isn't necessary. On the other hand, having only 1 way to do things is also a better API design practice. I am neutral on this, do as you think is the best.

psychon:
Ok, if you don't mind, then I'll just go with this version of the patch. The user-friendlier API is the right thing to do. Only issue is that the doc examples won't be able to use this simpler API any more (sorry!)...

Elv13:
👍

blueyed:
Thanks once again!

Issue #244, opened 2015-05-26
Improve handling of moving foo.lua to module (foo/init.lua)


blueyed:
With `beautiful` being moved to a module (`beautiful.lua` =>  `beautiful/init.lua` etc) (#229), `make install` will not remove the now obsolete `beautiful.lua`, which results in this file being preferred.

Since there is no `make uninstall` target, this has to be handled manually.

Is there a better way?

Can we make `make install` to remove the obsolete file, or add a `make uninstall` command?
The latter would have to be run before a new `make install` would update the list of installed files though probably.

Changing Lua's `package.path` would also work, but this appears to be contrary to Lua conventions:

```
diff --git i/luaa.c w/luaa.c
index 2c16152..3842108 100644
--- i/luaa.c
+++ w/luaa.c
@@ -462,8 +462,8 @@ luaA_init(xdgHandle* xdg)
     }

     /* add Lua lib path (/usr/share/awesome/lib by default) */
-    lua_pushliteral(L, ";" AWESOME_LUA_LIB_PATH "/?.lua");
     lua_pushliteral(L, ";" AWESOME_LUA_LIB_PATH "/?/init.lua");
+    lua_pushliteral(L, ";" AWESOME_LUA_LIB_PATH "/?.lua");
     lua_concat(L, 3); /* concatenate with package.path */
     lua_setfield(L, 1, "path"); /* package.path = "concatenated string" */
```

This happened with the dbus split already, and now with https://github.com/awesomeWM/awesome/pull/229.

psychon:
Urgh. I like people who use package managers. :-(
http://www.cmake.org/Wiki/CMake_FAQ#Can_I_do_.22make_uninstall.22_with_CMake.3F

psychon:
Could we print a warning during `make install`?

blueyed:
Possibly.

I've used the following in a custom script myself , before `make install`:

    # Handle migrations of Lua files to modules.
    sudo rm -r /usr/local/share/awesome/lib/beautiful* .build-*/lib/beautiful* || true

Elv13:
Can we close this? It isn't really a bug. It is the same for any project. If a C project remove a public `.h`, then the same thing happen. `make uninstall` exist for a reason.

psychon:
We do not have a `make uninstall`, do we?
But I'd say "package managers exist for a reason". I'd conclude that we cannot do anything about this and close this, yeah.

Elv13:
[...]

PR #1259, opened 2016-12-08
Add pointless naughty.lua and beautiful.lua files

psychon:
People who use a plain "make install" to install awesome will likely
also use the same approach to update their installation. However, this
does not actually work, because in awesome 3.5 there are beautiful.lua
and naughty.lua. These modules have since been split up into multiple
files and we now have beauitful/init.lua and naughty/init.lua instead.
The result is that the newer awesome will use some code from an older
version of awesome.

This commit has a work-around for this: We add "empty" beautiful.lua and
naughty.lua files whose only purpose is to load the real file. These
"empty" files will then overwrite files from an older installation and
everything works.

Sadly, this bad hack will have to be kept around forever and in the
future we will only have more instances of it. I would like to just say
people to fix their system, but apparently this should be worked around
instead.

Fixes: https://github.com/awesomeWM/awesome/issues/244

### DIFF
--- a/lib/beautiful.lua
+++ b/lib/beautiful.lua
@@ -1,0 +1,7 @@
+-- Work-around for broken systems which are updated by overwriting the awesome
+-- installation. This would not remove beautiful.lua from older awesome versions
+-- and thus breakage follows.
+-- The work-around is to use a pointless beautiful.lua file.
+return require("beautiful.init")
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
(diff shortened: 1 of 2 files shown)


psychon:
(I'd say "let's just move `beautiful/init.lua` back to `beautiful.lua` instead, but we cannot even do that, because there might be people who installed a git/master version of awesome...)

Elv13:
Keep in mind we can eventually remove those files, once 95% percent of the user base has migrated, then they can be removed. Those files wont hurt if they are on existing systems, they will keep doing what they're intended for even after we removed them.

blueyed:
Thanks!

Issue #1205, opened 2016-11-02
awesome crashes on vncserver session


snemiro514:
Output of `awesome --version`:

awesome v3.5.9 (Mighty Ravendark)
 • Build: Nov  2 2016 11:28:12 for x86_64 by gcc version 5.4.0 (root@pd-qs-16)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✘

(I'm trying the compiled version from the last sources. Same behaviour with the package in Ubuntu repo)

**How to reproduce the issue:**

Install Ubuntu 16.04
Install vnc4server
Install awesome 
Configure your xstartup with:
---

echo starting ... >&2
xterm +sb &
exec awesome
---
run 
$ vncserver :22

**Actual result:**
vnc session log: (session 22)

Xvnc Free Edition 4.1.1 - built Feb 25 2015 23:02:21
Copyright (C) 2002-2005 RealVNC Ltd.
See http://www.realvnc.com for information on VNC.
Underlying X server release 40300000, The XFree86 Project, Inc

Wed Nov  2 12:15:11 2016
 vncext:      VNC extension running!
 vncext:      Listening for VNC connections on port 5922
 vncext:      created VNC server for screen 0
error opening security policy file /etc/X11/xserver/SecurityPolicy
Could not init font path element /usr/X11R6/lib/X11/fonts/Type1/, removing from list!
Could not init font path element /usr/X11R6/lib/X11/fonts/Speedo/, removing from list!
Could not init font path element /usr/X11R6/lib/X11/fonts/misc/, removing from list!
Could not init font path element /usr/X11R6/lib/X11/fonts/75dpi/, removing from list!
Could not init font path element /usr/X11R6/lib/X11/fonts/100dpi/, removing from list!
starting ...
set mod4
E: awesome: signal_fatal:318: signal 11, dumping backtrace
awesome(backtrace_get+0x54) [0x423e07]
awesome() [0x40c183]
/lib/x86_64-linux-gnu/libc.so.6(+0x354b0) [0x7f03ffa884b0]
awesome(screen_scan+0xa0) [0x41e8c5]
awesome(main+0x6fb) [0x40cdfd]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0) [0x7f03ffa73830]
awesome(_start+0x29) [0x40be79]

**Expected result:**
awesome should run and not crash

psychon:
Since you are compiling yourself already:
[...]

PR #1260, opened 2016-12-09
Handle unexpected XCB failures

psychon:
We have many places where we are sending an XCB request and expect an
answer where the protocol guarantees that no error can occur and we are
sure to get an answer. However, for example if the X11 server crashes,
these places can still fail. This commit tries to handle failures at all
these places.

I went through the code and tried to add missing error checking (well,
NULL-pointer-checking) to all affected places.

In most cases these errors are just silently ignored. The exception is
in screen querying during startup. If, for example, querying RandR info
fails, we will fall back to Xinerama or zaphod mode. This is serious
enough that it warrants a warning. In most cases, we should exit shortly
afterwards anyway, because, as explained above, these requests should
only fail when our connection to the X11 server breaks.

References: https://github.com/awesomeWM/awesome/issues/1205#issuecomment-265869874

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -199,7 +199,7 @@ scan(xcb_query_tree_cookie_t tree_c)
 
         state = xwindow_get_state_reply(state_wins[i]);
 
-        if(!attr_r || attr_r->override_redirect
+        if(!geom_r || !attr_r || attr_r->override_redirect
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks! 🍒

PR #1262, opened 2016-12-10
4.0 Changelog

### LINKED ISSUE #967: Document API changes since 3.5
Inspired by #966 I just remembered that the C code now returns screen objects in places where it previously provided screen numbers. We should add this (and other changes) to the docs.

Feel free to edit-add points to the following list:
- Screen objects instead of screen numbers everywhere(?)
  - `mouse.screen`, `c.screen` etc are now screen objects instead of numbers! (This breaks shutdown(!) in https://github.com/pw4ever/awesome-wm-config). A compatibility layer might look like this: `local function fix_screen(s) if type(s) == "screen" then return s.index end return s end`
- Widget API changed to support partial redraws
- `menubar.menu_gen.generate` is now asynchronous and needs a callback as an argument (this breaks https://github.com/copycat-killer/awesome-copycats)
  (All the user configs that I tried produce metric tons of deprecation warnings, but I guess that is expected...)

Elv13:
This one was a long time coming. As for some weird \`, we abused of ldoc, that's the consequences.

Fix #967

### DIFF
--- a/docs/00-authors.md
+++ b/docs/00-authors.md
@@ -3,29 +3,38 @@
 
 ## awesome - awesome window manager
 
-**Author**:
+**Maintainer**:
+
+ * © 2009- Uli Schlachter &lt;psychon@znc.in>
+
+**Main contributors**:
+
+ * © 2013- Emmanuel Lepage-Vallee &lt;elv1313@gmail.com>
+ * © 2014- Daniel Hahler &lt;https://daniel.hahler.de>
+ * © 2014- Yauhen Kirylau &lt;actionless.loveless@gmail.com>
+
+**Original author**:
 
  * © 2007-2009 Julien Danjou &lt;julien@danjou.info>
-
-**Contributors**:
-
- * © 2007-2009 Nikos Ntarmos &lt;ntarmos@ceid.upatras.gr>
- * © 2007-2009 Arnaud Fontaine &lt;arnau@debian.org>
- * © 2008-2009 Damien Leone &lt;damien.leone@gmail.com>
- * © 2008-2009 Gregor Best &lt;farhaven@googlemail.com>
- * © 2008-2009 Maarten Maathuis &lt;madman2003@gmail.com>
- * © 2008-2009 koniu &lt;gkusnierz@gmail.com>
- * © 2009 Uli Schlachter &lt;psychon@znc.in>
 
 **Past contributors**:
 
  * © 2007-2008 Marco Candrian &lt;mac@calmar.ws>
+ * © 2007 Aldo Cortesi &lt;aldo@nullcube.com>
  * © 2008 Lucas de Vries &lt;lucasdevries@gmail.com>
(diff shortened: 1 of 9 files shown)


docs/00-authors.md
```diff
@@ -3,29 +3,38 @@
...
 
- * © 2007-2009 Nikos Ntarmos &lt;ntarmos@ceid.upatras.gr>
- * © 2007-2009 Arnaud Fontaine &lt;arnau@debian.org>
- * © 2008-2009 Damien Leone &lt;damien.leone@gmail.com>
- * © 2008-2009 Gregor Best &lt;farhaven@googlemail.com>
- * © 2008-2009 Maarten Maathuis &lt;madman2003@gmail.com>
- * © 2008-2009 koniu &lt;gkusnierz@gmail.com>
- * © 2009 Uli Schlachter &lt;psychon@znc.in>
+ * © 2013- Emmanuel Lepage-Vallee &lt;elv1313@gmail.com>
+ * © 2014- Daniel Hahler &lt;git@thequod.de>
```
blueyed: I would rather not have my email in there. But a link to GitHub, Twitter, or Daniel.Hahler.de instead.

docs/89-NEWS.md
```diff
@@ -0,0 +1,839 @@
...
+### Custom xproperties support
+
+Awesome can now save some data in the X11 server itself. This allows to
+communicate with external applications or so save state across restarts.
+
+There is also supported/used for persistent client properties.
+
+### Better XKB keyboard layout support
+
+Awesome now has native support for keyboard layout detection and setting. Using
```
psychon: Can we really set the keyboard layout? I think we can only set the active "group" of the layout. So in fact `setxkbmap` is still needed.

psychon:
I only took a quick look. The above is the result. If needed, I can take a more careful look, but if no one asks me to, I'll just believe that this is ok to go in.

Elv13:
Updated

> I can take a more careful look, but if no one asks me to, I'll just believe that this is ok to go in.

It can be done in other PRs anyway. I suggest merging it

docs/89-NEWS.md
```diff
@@ -0,0 +1,839 @@
+# NEWS
+
+![Logo](../../images/icons/aw_64_shadow.png)
```
psychon: The API doc can e.g. also be installed to `/usr/share/doc/awesome` (I think Debian does something like this?), so yeah, please add this file.

Elv13: Ok, should I merge this afterward?

Elv13:
I added 3 commits to re-use the logo generator @actionless implemented for the xresource theme.

pick5

psychon:
I guess: Feel free to merge

actionless:
i will rebase https://github.com/awesomeWM/awesome/pull/1219 later on, i am now in the process of relocating to the different country so don't have enough free time in the next month or so

PR #1263, opened 2016-12-11
doc: Change the working of the module aliases

### LINKED ISSUE #1146: doc: awful.screen marked as deprecated?!
`awful.screen` is documented as being deprecated (http://new.awesomewm.org/apidoc/libraries/awful.screen.html), but then in http://new.awesomewm.org/apidoc/classes/screen.html there are a lot of references to `awful.screen` functions.
I think this is confusing.

As for this (and other) deprecated aliased docs: they should not be used in the main navigation: you tend to click on `awful.screen` only to see that it's deprecated.

Elv13:
Fixes #1146

### DIFF
--- a/docs/aliases/awful_client.lua
+++ b/docs/aliases/awful_client.lua
@@ -1,6 +1,5 @@
 ---------------------------------------------------------------------------
---- This module is deprecated, use `client`
---  ===============================
+--- This module documentation has been merged with the `client` module.
 --
 -- @module awful.client
 ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


docs/aliases/awful_client.lua
```diff
@@ -1,6 +1,6 @@
 ---------------------------------------------------------------------------
---- This module is deprecated, use `client`
---  ===============================
+--- This module documentation has been merged with the `client` module
+--  ==================================================================
```
blueyed: Please add a dot at the end of the sentence (also below).

And I think it might render better without the underlining?!

Elv13:
Done

PR #1265, opened 2016-12-11
Improve deprecation messages

Elv13:


psychon:
Sigh. Not really all that pretty, but it gets the job done. My suggestions would amount to bike shedding, so 👍

PR #1267, opened 2016-12-12
Fix more rules screen mismatch

Elv13:
I almost forgot. As I said in the PR that fixed `new_tag`, the bug is also present for `tag` and `tags`. This fixes it and also improve `tags` so it support the tag-by-name thing @psychon added a while back (9cc28a8da). It is also consistent with the screen selection added by @psychon in 2e13c120e8

As with #1236, this isn't pretty, but it is also unavoidable.

No tests because it would require 5 of them and they are exponentially slow.

### DIFF
--- a/docs/90-FAQ.md
+++ b/docs/90-FAQ.md
@@ -371,6 +371,16 @@ xkb keyboard description xkb_symbols line.
     };
     EOF
 
+### I upgraded from Awesome 3 to Awesome 4 and multiscreen broke. Why is that?
+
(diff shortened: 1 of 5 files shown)


Elv13:
I was uncomfortable with adding this change without tests. So I added some, but I also made a commit I am really not proud of to disable the wallpaper on the multi-screen tests. This both make debugging harder and hide a bug.

Disabling the wallpaper free enough resources to offset the extra time the 5 new tests takes. The build is now 1 minute faster. I would like to revert this commit, so I left the code commented just to make sure it stays annoying enough so we don't forget.

PR #1269, opened 2016-12-13
Remove docs/doing_a_release.md

psychon:
I am no longer sure why I added this file. It doesn't really belong into
the awesome source code. It is better suited to be added to our web
page, I think.


psychon:
> As long as its somewhere and you don't get hit by a bus

If I get hit by bus, you still have this version of the document in git history.
If I don't get hit by a bus (and finally have a weekend of so that I can actually do something awesome-related which takes more than 2 minutes), I'll try writing something up, make a WIP pull request with it on awesome-www, wait for my next weekend (so that others can complain/improve), then do a 4.0-rc1 release and finally fix up the doing-a-release document with the lessons learnt from trying to apply it.
At least that's the plan.

Issue #1273, opened 2016-12-14
Restore last position of windows


llatigid:
Hello Awesome! It's not an issue, it's an idea -- the option, that remembers a window last position and its next re-opening will be exactly at the same place. Could you support the brilliant feature?

Elv13:
@mindeunix has a module (at least in theory) capable of doing this. 

https://github.com/mindeunix/clientdb

Awesome is mostly used as a tiled WM. Therefor this feature makes little sense for the majority of the user base. KDE and some others applications also implement this themselves. Having our own system would just fight against this.

Finally, if you want this and still doesn't want to use @mindeunix module, then this is something that belongs to `rc.lua` or an external module, not the Awesome core. It is one of these "it wont ever work perfectly"* features we avoided adding to Awesome.

I will close this for now. If another member of the team thinks I am wrong, then they will re-open it.

* We already have input grabbing, the systray, startup-notification and the current screen algorithm. Each of them is a nightmare to maintain.

Elv13:
The documentation (for 4.0, but still) has been updated to better guide you if you wish to enable such feature:
https://awesomewm.org/apidoc/libraries/awful.spawn.html

PR #1274, opened 2016-12-15
More spawn doc

Elv13:
Inspired by #1273, https://github.com/mindeunix/clientdb/issues/, https://www.reddit.com/r/awesomewm/comments/5icvga/sending_notification_after_entering_the_password/ and some IRC questions over the last few months.

### DIFF
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -6,6 +6,61 @@
 -- program after it has been launched.  This requires currently that the
 -- applicaton supports them.
 --
+-- **Understanding clients versus PID versus commands versus class**:
+--
+-- A *process* has *PID* (process identifier). It can have no, one or many *window*s.
+--
+-- A *command* if what is used to start *process*(es). It has no direct relation
+-- with *process*, *client* or *window*. When a command is executed, it will
+-- usually start a *process* which keeps running until it exits. This, however,
+-- isn't always true as some applications use scripts as command and some other
+-- use various single instance mechanism (usually client/server) and merge with
+-- an existing process.
+--
+-- A *client* correspond to a *window*. It is owned by a process. It can have
+-- both a parent and one or many children. A *client* has a *class*, an
+-- *instance*, a *role*, and a *type*. See `client.class`, `client.instance`,
+-- `client.role` and `client.type` for more information about these properties.
+--
+-- **The startup notification protocol**:
+--
+-- The startup notification protocol is an optional specification implemented
+-- by X11 applications to bridge the chain of knowledge between the moment a
+-- program is launched to the moment its window (client) is shown. It can be
+-- found [on the FreeDesktop.org website](https://www.freedesktop.org/wiki/Specifications/startup-notification-spec/).
+--
+-- Awesome has support for the various events that are part of the protocol, but
+-- the most useful is the identifier, usually identified by its `SNID` acronym in
+-- the documentation. It isn't usually necessary to even know it exists, as it
+-- is all done automatically. However, if more control is required, the
+-- identifier can be specified by an environment variable called
+-- `DESKTOP_STARTUP_ID`. For example, let us consider execution of the following
+-- command:
+--
+--    DESKTOP_STARTUP_ID="something_TIME$(date '+%s')" my_command
+--
+-- This should (if the program correctly implements the protocol) results in
+-- `c.startup_id` to at least match `something`.
+-- This identifier can then be used in `awful.rules` to configure the client.
+--
+-- Awesome can automatically set the `DESKTOP_STARTUP_ID` variable. This is used
+-- by `awful.spawn` to specify additional rules for the startup. For example:
+--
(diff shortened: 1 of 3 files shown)


lib/awful/spawn.lua
```diff
@@ -6,6 +6,61 @@
...
+-- documentation. It isn't usually necessary to even know it exists, as it is all
+-- done automatically. However, if more control is required, the identifier
+-- can be specified in by an environment variable called `DESKTOP_STARTUP_ID`.
+-- For example, if a command such as:
+--
+--    DESKTOP_STARTUP_ID="something" my_command
+--
+-- Is executed in a terminal, it should (if the program correctly implements the
+-- protocol) results in `c.startup_id` to at least match `something`. In case it
+-- is already used, some implementations will happen some suffixes to the string.
```
psychon: Huh? Any examples for this?

As far as I remember, the `DESKTOP_STARTUP_ID` should have a special format and contain a timestamp. If there is no timestamp, GTK adds one.
How could anything even check if the startup id is already used?

lib/awful/spawn.lua
```diff
@@ -6,6 +6,24 @@
...
+-- * When the command is a shell alias
+--
+-- Note there a shell is **not** a terminal emulator. A terminal emulator is
+-- something like XTerm, Gnome-terminal or Konsole. A shell is something like
+-- `bash`, `ZSH`, `busybox sh` or `Debian ash`.
+--
+-- If you wish to open a process in a terminal window, check that yours support
+-- the common `-e` option. If it does, then something like this should work:
+--
+--    awful.spawn(terminal.." -e 'my command'")
```
psychon: Sadly, this is a problem.
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=600429

With urxvt, you need the quotes. With gnome-terminal and terminator the quotes may not be present. Thus, you cannot portably execute a command containing spaces in a terminal.

I think this means we cannot describe this in the docs.

lib/awful/spawn.lua
```diff
@@ -79,6 +79,52 @@
...
+-- First, do **not** use `io.popen` **ever**. It is synchronous. Synchronous
+-- function **stop everything** until they are done. All visual applications
+-- locks (as Awesome no longer respond), you will probably lose some keyboard
+-- and mouse events and will have higher latency when playing games. This is
+-- also true about reading files, but this is another topic.
+--
+-- Awesome provides a few ways of getting command results. One is to use the
+-- `Gio` libraries directly. This is usually very complicated, but gives a lot
+-- on control over the command execution, including the ability to create
+-- custom pipes and link many process.
```
psychon: Uhm, I'm not sure if `Gio` really provides this, and if it does, then I bet that this is not accessible through `lgi`. Could we remove this part? (And just say that using `Gio` directly is complicated)

psychon:
I added some fixup commits. If you agree with them, `git pull` to get them locally and `git rebase -i --autosquash master` to apply them. The comments that I left above are still open, I think (even though the GitHub interface disagrees).

lib/awful/spawn.lua
```diff
@@ -130,20 +130,19 @@
...
--- to use the standard command. How the default application is selected is
+-- If the intent is to open a document, then it is recommended to use the
+-- following standard command. How the default application is selected is
 -- defined in the [Shared MIME-info Database](https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html)
 -- specification. The `xdg-utils` package provided by most distributions
--- will provide the `xdg-open` command:
+-- provides the `xdg-open` command:
 --
---    awful.spawn("xdg-open /path/to/file")
+--    awful.spawn({"xdg-open", "/path/to/file"})
```
psychon: Is this change ok?

I know that everyone thinks of commands as strings. However, the kernel (the `execve` syscall) wants an array of strings instead. Directly giving such an array to awesome avoids all kinds of word-splitting problems.

Sadly, this cannot really be used in the other examples...
(`noisy` could be turned into `noisy = { "bash", "-c", [[lots of text]] }`, but perhaps in this case that is more confusing than helping)

lib/awful/spawn.lua
```diff
@@ -79,6 +79,52 @@
 --        placement = awful.placement.bottom_right,
 --    })'
 --
+-- **Getting a command result**:
+--
+-- First, do **not** use `io.popen` **ever**. It is synchronous. Synchronous
+-- function **stop everything** until they are done. All visual applications
+-- locks (as Awesome no longer respond), you will probably lose some keyboard
+-- and mouse events and will have higher latency when playing games. This is
```
Elv13: > (but some key-bindings and focus-changing-keybindings will have non-intuitive results).

Or not happen at at. I have this issue when I compile and move the mouse too fast (yes, some modules I use, like Vicious and Drawer, still use `popen`...). This is the kind of issues I am referring to. I found awesome to be unreliable when it is slowed down by blocking calls. My Awesome 3.4 config had serious issues with that.

Elv13:
done

PR #1277, opened 2016-12-18
Fix luacheck installation on Travis

psychon:
A trivial typo in .travis.yml caused luacheck to never run.

The Travis build will fail, because we have luacheck warnings (this is how I found this: I had luacheck warnings locally and wondered how that could be). I do not have time to fix this, so it would be nice if someone can pick this up and do something about it.

Edit: And of course I had still a typo in the first version of this PR's commit -.-

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,7 +86,7 @@ install:
   # Install busted for "make check-unit".
   - travis_retry sudo luarocks install busted
   # Install luacheck for "make check-qa".
-  - if [ "$DO_QUECKQA" = 1 ]; then travis_retry sudo luarocks install luacheck; fi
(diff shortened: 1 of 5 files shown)


psychon:
Found some time to work on this. be5934af7892809ca28898c4395d1901f666aea8 is ugly, so CC @Elv13

psychon:
I understand that as "the change is mostly ok", so I changed the variable name into `skip_wallpapers_due_to_weird_slowdowns`. Sorry for the inappropriate comment, but I didn't have any good ideas for the name instead.

Elv13:
Thanks, sorry for adding a couple of luacheck issues, I should not have trusted the CI to do that for me (even if it's kind of the whole point of the thing)

PR #1279, opened 2016-12-20
Tag & keybindings doc

Elv13:
Another row of "I was more than once asked this, lets write it down" now that the old wiki is closed down.

### DIFF
--- a/objects/tag.c
+++ b/objects/tag.c
@@ -54,6 +54,43 @@
  *        screen = s,
  *    }
  *
+ * **Accessing tags**:
+ *
+ * To access the "current tags", use
+ *
+ *    local tags = awful.screen.focused().selected_tags
+ *
+ * See: `awful.screen.focused`
+ *
+ * See: `screen.selected_tags`
+ *
+ * To ignore the corner case where multiple tags are selected:
+ *
+ *    local t = awful.screen.focused().selected_tag
+ *
+ * See: `screen.selected_tag`
+ *
+ * To get all tags for the focused screen:
+ *
+ *    local tags = awful.screen.focused().tags
+ *
+ * See: `screen.tags`
+ *
+ * To get all tags:
+ *
+ *    local tags = root.tags()
+ *
+ * To get the current tag of the focused client:
+ *
+ *    local t = client.focus and client.focus.first_tag or nil
+ *
+ * See: `client.focus`
+ * See: `client.first_tag`
+ *
+ * To get a tag from its name:
+ *
+ *    local t = awful.tag.find_by_name(awful.screen.focused(), "name")
+ *
  * Some signal names are starting with a dot. These dots are artefacts from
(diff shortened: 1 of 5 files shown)


objects/tag.c
```diff
@@ -91,6 +91,83 @@
...
+ *
+ * Here is a few useful shortcuts not part of the default `rc.lua`. Add these
+ * functions above `-- {{{ Key bindings`:
+ *
+ * Delete the current tag
+ *
+ *    local function delete_tag()
+ *        local t = awful.screen.focused().selected_tag
+ *        if not t then return end
+ *        t:delete()
```
psychon: Why not `if t then t:delete() end`? But whatever, I'll stop the bike-shedding here.

objects/tag.c
```diff
@@ -91,6 +91,83 @@
...
+ *                end
+ *            end
+ *        }
+ *    end
+ *
+ * Move the focused client to a new tag
+ *
+ *    local function move_to_new_tag()
+ *        local c = client.focus
+ *        if not c then return add_tag() end
```
psychon: Should this really call `add_tag()` here? I would expect this to do nothing if no key is focused.

psychon:
I also pushed two fixup commits for spelling things, please squash if you agree to them.

objects/tag.c
Elv13: Just me following my usual precondition programming style. In this case its a little silly since there is only one line, but these days I prefer to stick to that style. It avoids a lot of indentation, reduce NULL dereference risk and (but not in this case) make the code more readable.

Elv13:
done

psychon:
Sorry, but I added another two fixup commits. Feel free to merge afterwards.

objects/tag.c
```diff
@@ -26,6 +26,148 @@
...
+ *    awful.tag(
+ *      { "1", "2", "3", "4", "5", "6", "7", "8", "9" },
+ *      s,
+ *      awful.layout.layouts[1]
+ *    )
+ *
+ * If you wish to have tags with different properties, then `awful.tag.add` is
+ * a better choice:
+ *
+ *    awful.tag.add("First tag", {
```
Wouterbeets: "(" is not closed

objects/tag.c
```diff
@@ -26,6 +26,148 @@
...
+ *        icon               = "/path/to/icon1.png",
+ *        layout             = awful.layout.suit.tile,
+ *        master_fill_policy = "master_width_factor",
+ *        gap_single_client  = true,
+ *        gap                = 15,
+ *        screen             = s,
+ *    }
+ *
+ *    awful.tag.add("Second tag", {
```
Wouterbeets: "(" is not closed

psychon:
@Wouterbeets Thanks, but that was already fixed in commit 160ea0fd66629d60597096040c3f7f32a6182018 18 months ago

PR #1280, opened 2016-12-20
Remove traces of the old URL

Elv13:
Lets hope not too many people have scripts using the DBUS URI directly

### DIFF
--- a/docs/01-readme.md
+++ b/docs/01-readme.md
@@ -130,9 +130,7 @@ for any coding, documentation or patch guidelines.
 
 ## Documentation
 
-Online documentation is available at http://awesome.naquadah.org/doc/ for the
-stable branch and at http://awesomewm.github.io/apidoc/ for the master branch.
-It can be built using `make ldoc`.
+Online documentation is available at https://awesomewm.org/apidoc/.
 
 ## License
 
--- a/manpages/awesome-client.1.fr.txt
+++ b/manpages/awesome-client.1.fr.txt
@@ -38,7 +38,7 @@ Julien Danjou <julien@danjou.info>
 WWW
 ---
 
-http://awesome.naquadah.org
+https://awesomewm.org
 
 TRADUCTION
 ----------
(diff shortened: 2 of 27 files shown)


docs/01-readme.md
```diff
@@ -130,9 +130,7 @@ for any coding, documentation or patch guidelines.
 
 ## Documentation
 
-Online documentation is available at http://awesome.naquadah.org/doc/ for the
-stable branch and at http://awesomewm.github.io/apidoc/ for the master branch.
-It can be built using `make ldoc`.
+Online documentation is available at http://awesome.org/apidoc/.
```
psychon: Currently we have both `/doc` (part of the awesome-www repo and build via Travis) and `/apidoc` (via the `apidoc` repo, always points to master).

Since this file is the README both for our GitHub repo (should point to `/apidoc`, I guess) and for the built documentation that gets installed (shouldn't point at any web pages at all, since the docs are already available), I'm not really sure what to do here.
Thoughts?

manpages/awesome-client.1.fr.txt
```diff
@@ -1,3 +1,4 @@
+
 awesome-client(1)
```
psychon: Why the empty line?

psychon:
Get rid of the empty line in the man page and 👍 
(The README-issue is unrelated to this PR, I think, but it might still make sense to do something about it)

docs/01-readme.md
blueyed: Agreed, but also no idea.
@Elv13 
Please use https.

manpages/awesomerc.5.txt
```diff
@@ -54,4 +54,4 @@ Written by Julien Danjou <julien@danjou.info>.
 
 WWW
 ---
-http://awesome.naquadah.org
+http://awesomewm.org
```
blueyed: https

Elv13:
done

blueyed:
Thanks!

> **WARNING**: This breaks the API

Because of the DBUS change?

Elv13:
Yes, and if people try to use the installed `awesome-client` with a git build, it will fail.

PR #1283, opened 2016-12-21
doc: Add a porting guide.

Elv13:
Also thanks to @syphoxy for his contribution to this guide. This isn't a real diff, I originally reverted some cosmetic commits to `rc.lua`, then removed the new features new re-converted it to the imperative widget API. Hopefully this should reduce the porting effort down to the "just make it works" territory.

Done because of @sim590 questions on IRC.

### DIFF
--- a/docs/17-porting-tips.md
+++ b/docs/17-porting-tips.md
@@ -1,0 +1,341 @@
+# Porting tips
+
+## From 3.5 to 4.0
+
+The
+**best advice is to start from the default `rc.lua` and backport any changes**
+you might have added to your `rc.lua`.
+This avoids most of the possible errors due to missing important changes.
+
+If you still wish to ignore this advice, first read the
+<a href="./89-NEWS.md.html">NEWS</a> section about the breaking changes. This
+document assumes you did.
+
+Be warned, even if it looks like it's working after changing some lines, **it wont
+supports dynamic screen changes and will have many subtle bugs**. The changes
+mentioned below **are important for the stability of your window manager**.
+
+Here is a diff of the 3.5.9 `rc.lua` with the 4.0 one. All changes due to
+new features and new syntaxes have been removed. A `-` in front of the line
+correspond to content of the 3.5 `rc.lua` and `+` its replacement in 4.0.
+
+This document does not cover the new features added in the Awesome 4 `rc.lua`,
+it only covers the minimal required changes to have a properly behaving config.
+
+To test during the port, we recommend the `Xephyr` X11 server.
+
+If Awesome4 **is not installed yet**, we recommand to install it in its own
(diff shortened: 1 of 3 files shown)


docs/17-porting-tips.md
```diff
@@ -0,0 +1,333 @@
...
+mentioned below **are important for the stability of your window manager**.
+
+Here is a diff of the 3.5.9 `rc.lua` with the 4.0 one. All changes due to
+new features and new syntaxes have been removed. A `-` in front of the line
+correspond to content of the 3.5 `rc.lua` and `+` its replacement in 4.0.
+
+To test during the port, we recommend the `Xephyr` X11 server:
+
+    Xephyr :1 -screen 1280x800 &
+    DISPLAY=:1 path/to/awesome -c /path/to/rc.lua --search /path/to/awesome/lib
```
psychon: Why the `--search`? I am not sure which path I am supposed to use here and I bet most users would have similar problems. Perhaps it makes sense to assume a prefix (e.g. `/usr/local`) and then give the full command based on that? When this is mentioned in the text, users should have a chance to adapt to their local paths.

docs/17-porting-tips.md
```diff
@@ -0,0 +1,333 @@
...
+     -- Create a wibox for each screen and add it
+     -mywibox = {}
+     -mypromptbox = {}
+     -mylayoutbox = {}
+     -mytaglist = {}
+     -mytasklist = {}
+
+Many functions have been converted to methods. The old functions are deprecated,
+they are still supported, but will be removed in the next release. Also note
+that the name of the tables is now in a local variables instead of the table
```
psychon: I'm lazy right now, so no fixup commit: `in a local variables` -> `in local variables`.

Also, I do not understand this sentence. The name of the table is in a local variable? So if the table is called `foo`. its name in a local variable would mean `local table_name = "foo"`. Then I would access the table as `_G[table_name]`. Obviously this is wrong, but this is how I understand the sentence.

docs/17-porting-tips.md
```diff
@@ -0,0 +1,333 @@
...
+                                                   `awful.client.focus.byidx`(1)
+    -                                              if `client.focus` then `client.focus:raise`() end
+                                               end),
+                          awful.button({ }, 5, function ()
+                                                   `awful.client.focus.byidx`(-1)
+    -                                              if `client.focus` then `client.focus:raise`() end
+                                               end))
+
+
+This section is **very important**. This is the section that handle when screens
```
psychon: handle*s*

However, I would instead propose to replace the sentence: This is where adding and removing screens is handled.

psychon:
Feel free to merge after handling the above

Issue #1266, opened 2016-12-11
Define a new release strategy (post-4.0)


Elv13:
*First, everything described below is **my** opinion and **not** an announcement. I would like feedbacks and other proposals to solve the core issues highlighted by the 4 years delta between our last 2 major releases. The final decision isn't mine to make anyway.*

This issue is to discuss how releases will be done going forward. The past 4 years was a total failure in this regard as vast amount of useful code, documentation and bugfix failed to reach 95% of our userbase. There is also the fact that we have been promoting the use of the master branch for a while, but failed to communicate this correctly.

I propose following guidelines to *embrace our current practices* and make them official

### Create one release per commit in the master branch

We already do that anyway. Try `awesome --version` on the Git master branch and you already see that version number. Also, distribution such as Arch and Debian want the release string to be comparable using `>`, `==` and `<`. This doesn't mean we should officially announce them all. Also, I don't think we should even tag them anymore. We don't have the manpower to keep 2 supported branches.

So, no more 4.0.x releases. just 4.X where `X` can become very large. Chrome and many other projects already do this.

### If it isn't stable and tested, then it doesn't belong in master

The SVN days are long gone, branches are cheap and easy to merge. Most Agile development already follow that rule. We also unofficially follow it. Lets make it official.

Sure, some changes are scarier than other and breakages will happen. Both 3.5.8 and 3.5.9 existed to fix breakages from 3.5.7. If I recall, the same did happen with 3.5.6 and 3.5.1/2. It can't be fully prevented. Lets be cautious, but not overly concerned. Keeping bug fixes away form the releases is often worst than the consequence of the small regressions they can cause. Release early, release often. Even Microsoft now embraced this rule. We're late to the party, lets catch up.

### An official tarball release can be generated from the master branch without further process

If there is a bugfix worthy of doing an "official" snapshot release, then merging master into the release branch should automatically generate an official release. The CI could tag it (or not, TBD). A commit or "back merge" should be done in the master branch to keep the commit count aligned. This can be done manually at first (TBD: PGP).

[...]

PR #1284, opened 2016-12-21
Suggest a codename for 4.x and add a controversial commit

Elv13:
Well, here's some more pressure to release v4

The rolling release commit will be controversial (for you), I know. I already added my little manifesto in #1266, discussions should happen there. The commit bellow assumes the tag is  called "v4". Maybe it should also check that the branch is "master".

As for the codename, many were discussed on IRC with @blueyed. That's my final pick. I think it represents v4 well. Options are still open.

It is *harder* because of the multi screen support
It is *better* because of the multi screen support
It is *faster* because of the partial redraw
It is *stronger* because... HULK!

Other codenames discussed have a common theme:

 * "an ancient sign of coming storm"
 * "time of my life"
 * "It's about time"
 * "Time"
 * "Time Has Come Today"

By the way, happy 4th anniversary to Awesome 3.5.0

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -4,7 +4,7 @@ set(PROJECT_AWE_NAME awesome)
 # `git describe` later.
 set(VERSION devel)
 
-set(CODENAME "The Fox")
+set(CODENAME "Harder, Better, Faster, Stronger")
 
 project(${PROJECT_AWE_NAME} C)
 
--- a/build-utils/git-version-stamp.sh
+++ b/build-utils/git-version-stamp.sh
@@ -17,7 +17,8 @@ if [ -z "$STAMP" ]; then
     die "Missing STAMP: $1"
 fi
 
-CURRENT=`git describe --dirty 2>/dev/null`
+CURRENT=`git describe 2>/dev/null | awk -F- '{print(($2=="")?($1".0"):($1"."$2))}'`
+
 if [ -z "$CURRENT" ]; then
     die "git describe failed: $(git describe --dirty)."
 fi


build-utils/git-version-stamp.sh
```diff
@@ -17,7 +17,8 @@ if [ -z "$STAMP" ]; then
     die "Missing STAMP: $1"
 fi
 
-CURRENT=`git describe --dirty 2>/dev/null`
+CURRENT=`git describe 2>/dev/null | awk -F- '{print(($2=="")?($1".0"):($1"."$2))}'`
```
psychon: Since `awk` is, well, awk:
"If no `-` appears in the input, append `.0`. Otherwise replace the first `-` with `.`".

I am not a big fan of this idea. It is purely cosmetic, but previously you could just take the output of `awesome --version` and feed it to `git show` and things to figure out the exact source code version used (well, and `--dirty` informs you about the existence of changes that you would not find this way). After this, finding the right git commit-ish becomes a lot harder.

awesomeConfig.cmake
```diff
@@ -4,7 +4,7 @@ set(PROJECT_AWE_NAME awesome)
 # `git describe` later.
 set(VERSION devel)
 
-set(CODENAME "The Fox")
+set(CODENAME "Harder, Better, Faster, Stronger")
```
psychon: I remember you asking for a RC. Is this codename for the RC or for the final release? (I assume for the final release)

psychon:
> Well, here's some more pressure to release v4

It was my birthday about two weeks ago. The weekend was busy with christmas shopping and preparing things. Following weekend was a family party. Next weekend... well, guess ;-)
Outside of weekends I am at work where I should, well, work. I am doing more awesome-y stuff there already than I should.

I'm sorry that my time is so limited, but I'm really trying my best.

>  The commit bellow assumes the tag is called "v4".

Uhm, it should  be v4.0, I think. This might be an exercise in bike-shedding, but "v4.0, v4.1, v4.2" looks better to me.

> It is harder because of the multi screen support
> It is better because of the multi screen support

At this point I could see where this was heading to and I was sold. ;-)

awesomeConfig.cmake
Elv13: I would have liked an RC if it was released 2 weeks ago, but  it was more of a marketing move than anything else. I did a the marketing anyway using "soon" instead of "RC". It produced the few bug reports we had, pointed to flaws in the doc and got us a new screenshot, so I am happy.

In the rolling scenario, "Harder, Better, Faster, Stronger" would be the codename for Awesome 4.

Elv13:
> Uhm, it should be v4.0, I think. This might be an exercise in bike-shedding, but "v4.0, v4.1, v4.2" looks better to me.

It isn't bike shedding. It is important to discuss a new release strategy. This should be done in the other issue. 3.5 is now over 4 years old and this is one (but not the main) reason why we have lost so many users (the main reason simply being having more competition filling niches we were never, and still are not, good at filling anyway). This PR is about the codename, the other commit was just to make you talk about the release ;), sorry for the evilness.

blueyed:
"Harder, Better, Faster, Stronger" seems to be agreed upon.
But the other commit should be removed?!

Let's close this PR and use the codename when releasing?

Issue #1254, opened 2016-12-05
Provide signatures


sonicou:
Please provide pgp signatures for future releases.
Thank you very much for your work.

psychon:
The [release announcement mails](http://www.mail-archive.com/awesome@naquadah.org/msg07828.html) are PGP-signed and contain the md5 and sha1 hashes of the released filed (newer announcement use sha2 hashes, but I can't find any such mail right now). So I just want to mention that there already is a chain of trust towards the released files, but I agree that it is inconvenient and providing a detached signature is a good idea.

The step-by-step list that is used for doing releases is available (but will soon need to be updated). 
https://github.com/awesomeWM/awesome/blob/master/docs/doing_a_release.md
I guess adding a `gpg --output $FILE.sig --detach-sig $FILE` step is all that is needed to handle this.

sonicou:
Thanks.
For distro maintainers it is important to have signatures for easy and automated verification.

PR #1286, opened 2016-12-23
Sign release tarballs

psychon:
Fixes: https://github.com/awesomeWM/awesome/issues/1254


Elv13:
I guess it should check if `gpg` is installed and if there it's capable of signing anything. Beside that 👍

blueyed:
`set -e` takes care of it, and it is not really user facing anyway.

psychon:
Well, if no `gpg` or no key is available, things will fail and no release should be made in the first place. ;-)
(And no, I'm still low on "time available")

Issue #1292, opened 2016-12-25
Local failures with tests/run.sh


blueyed:
I am getting the following failures locally:

```
== Running …/src/awesome/tests/test-awesomerc.lua ==
Error: timeout waiting for signal in step 1/11 (@20).
===> ERROR running …/src/awesome/tests/test-awesomerc.lua! <===
Error: timeout waiting for signal in step 1/11 (@20).

===> ERROR running …/src/awesome/tests/test-awful-layout.lua! <===
Error: timeout waiting for signal in step 1/182 (@20).

===> ERROR running …/src/awesome/tests/test-awful-tag.lua! <===
Error: timeout waiting for signal in step 1/61 (@20).
== Running …/src/awesome/tests/test-awful-widget-button.lua ==
Test finished successfully
== Running …/src/awesome/tests/test-awful-widget-watch.lua ==

== Running …/src/awesome/tests/test-resize.lua ==
Error: running function for step 11/39 (@1): [string "local test_client = require("_client")..."]:188: assertion failed!
stack traceback:
        [C]: in function 'assert'
        [string "local test_client = require("_client")..."]:188: in function <[string "local test_client = require("_client")..."]:186>
        [C]: in function 'xpcall'
        …/src/awesome/tests/_runner.lua:49: in function <…/src/awesome/tests/_runner.lua:43>
        [C]: in function 'xpcall'
        lib/gears/protected_call.lua:30: in function 'protected_call'
        lib/gears/timer.lua:167: in function <lib/gears/timer.lua:165>!
===> ERROR running …/src/awesome/tests/test-resize.lua! <===
Error: running function for step 11/39 (@1): [string "local test_client = require("_client")..."]:188: assertion failed!

== Running …/src/awesome/tests/test-urgent.lua ==
Error: timeout waiting for signal in step 2/4 (@5).
===> ERROR running …/src/awesome/tests/test-urgent.lua! <===
Error: timeout waiting for signal in step 2/4 (@5).
```

The first one should be fixed by https://github.com/awesomeWM/awesome/pull/1289, but I am wondering why this is only an issue for me locally, and not Travis and you?!

psychon:
No time to look much into it, but my guess is that the previous test still had some clients open that didn't exit fast enough (`c:kill()` takes some time). Dunno about the other tests. Do things also fail if you run the tests individually? Dunno, other ideas...?

blueyed:
Yes, IIRC it also happens when running the file alone.
[...]

PR #1293, opened 2016-12-25
tests: fix test-awful-layout

blueyed:
TODO:
 - [ ] wait for investigation on https://github.com/awesomeWM/awesome/issues/1292.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -224,6 +224,7 @@ for f in $tests; do
             grep -q -E '[Ee]rror|assertion failed' "$awesome_log"; then
         echo "===> ERROR running $f <==="
         grep --color -o --binary-files=text -E '.*[Ee]rror.*|.*assertion failed.*' "$awesome_log" || true
+        ((errors + 1))
     fi
 done
 
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -222,9 +222,8 @@ for f in $tests; do
 
     if ! grep -q -E '^Test finished successfully$' "$awesome_log" ||
             grep -q -E '[Ee]rror|assertion failed' "$awesome_log"; then
-        echo "===> ERROR running $f! <==="
+        echo "===> ERROR running $f <==="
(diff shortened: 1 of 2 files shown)


tests/run.sh
```diff
@@ -222,9 +222,8 @@ for f in $tests; do
 
     if ! grep -q -E '^Test finished successfully$' "$awesome_log" ||
             grep -q -E '[Ee]rror|assertion failed' "$awesome_log"; then
-        echo "===> ERROR running $f! <==="
+        echo "===> ERROR running $f <==="
         grep --color -o --binary-files=text -E '.*[Ee]rror.*|.*assertion failed.*' "$awesome_log" || true
-        ((errors + 1))
```
psychon: What happened to this?

tests/run.sh
```diff
@@ -224,6 +224,7 @@ for f in $tests; do
             grep -q -E '[Ee]rror|assertion failed' "$awesome_log"; then
         echo "===> ERROR running $f <==="
         grep --color -o --binary-files=text -E '.*[Ee]rror.*|.*assertion failed.*' "$awesome_log" || true
+        ((errors + 1))
```
psychon: A fixup gone wrong?

psychon:
Besides the fixup gone wrong: :+1: 

I guess the issue is that clients are too slow at exiting (`c:kill()` takes some time to work) and we start running the next tests already before all clients are gone?

blueyed:
There is no `c:kill()` involved here, is it?
It looks more like spawning 5 in the same loop does not work (for me).
Might be better after / fixed by https://github.com/awesomeWM/awesome/commit/fb874122ba1ac06a3af29c39c5abe6415e77bbf8.

psychon:
The previous test might have opened clients. The test runner `c:kill()`s them after the test is done. However, I think that nothing makes sure the clients really close before the next test is started, hence the `if` in this test sees `#client.get() ~= 0` and doesn't start the `xterm`s.

At least that's my theory.

Issue #1281, opened 2016-12-21
awesome.spawn does not handle errors well


blueyed:
> a,b = require("awful").spawn("FOO=1 bar"); return a
> boolean true

On stderr:

> 2016-12-21 01:32:30 …/ib/awful/spawn.lua:107: W: Failed to execute child process "FOO=1" (No such file or directory)

I think it should not return a trueish value!?

```
% awesome-client
awesome# return awesome.spawn("FOO=1 bar")
   string "FOO=1 bar"
```

psychon:
Sigh. The error handling in `awesome.spawn` is just broken. It returns its last argument on error. -.-
Breakage comes from 5e6a8932078e3352c972dfb811669f3b96cb3978 and no one noticed it in the last year.

The docs for `awesome.spawn` say that it just returns a string on errors. Do we really want that, or wouldn't `false, "error message"` make more sense?
Anyway, patch could be (this also gets rid of the warnings, I do not like them anyway):
```diff
diff --git a/spawn.c b/spawn.c
index f65c68d..9dd626b 100644
--- a/spawn.c
+++ b/spawn.c
@@ -414,11 +414,11 @@ luaA_spawn(lua_State *L)
     {
         g_strfreev(argv);
         if (error) {
-            luaA_warn(L, "spawn: parse error: %s", error->message);
+            lua_pushfstring(L, "parse error: %s", error->message);
             g_error_free(error);
         }
         else
-            luaA_warn(L, "spawn: There is nothing to execute");
+            lua_pushliteral(L, "There is nothing to execute");
         return 1;
     }
 
@@ -443,7 +443,7 @@ luaA_spawn(lua_State *L)
     g_strfreev(argv);
     if(!retval)
     {
-        luaA_warn(L, "%s", error->message);
+        lua_pushstring(L, error->message);
         g_error_free(error);
         if(context)
             sn_launcher_context_complete(context);
```
Due to my usual not-having-much-time: Anyone is free to pick up the above patch and submit a pull request, as long as we also get a test for this.

blueyed:
> The docs for `awesome.spawn` say that it just returns a string on errors. Do we really want that, or wouldn't `false, "error message"` make more sense?

Yes, it would make more sense, but is 
Should `luaA_spawn` be changed then, too?
(https://github.com/awesomeWM/awesome/blob/0444bf30cf57e6f222ad1dc6d82e5342037f1b93/spawn.c#L372-L377)
[...]

PR #1296, opened 2016-12-25
Fix error handling in awesome.spawn()

psychon:
Commit 5e6a893 broke error handling in awesome.spawn(): Instead of
returning an error message, it would just return its last argument.

This commit fixes that, removes some not-so-helpful warnings, and adds
lots of tests for this code.

Fixes: https://github.com/awesomeWM/awesome/issues/1281

### DIFF
--- a/tests/test-spawn.lua
+++ b/tests/test-spawn.lua
@@ -29,6 +29,9 @@ local steps = {
     error_message = spawn(" ")
     assert(string.find(error_message, 'Text was empty'), error_message)
 
+    error_message = spawn("")
+    assert(string.find(error_message, 'Text was empty'), error_message)
+
     error_message = spawn{}
     assert(string.find(error_message, 'There is nothing to execute'), error_message)
 
@@ -89,17 +92,6 @@ local steps = {
       assert(exit_snd == 42)
       return true
     end
-  end,
-  function()
-    -- Test empty command table
-    spawn{}
-    return true
-  end,
-  function()
-    -- Test empty command string
-    spawn("")
-    assert(#client.get() == 0)
-    return true
   end,
 }
 
--- a/tests/test-spawn.lua
+++ b/tests/test-spawn.lua
@@ -10,6 +10,31 @@ local exit_yay, exit_snd = nil, nil
 -- * spawn with startup notification is covered by test-spawn-snid.lua
 
 local steps = {
+  function()
+    -- Test various error conditions. There are quite a number of them...
+    local error_message
+
+    error_message = spawn("this_does_not_exist_and_should_fail")
+    assert(string.find(error_message, 'No such file or directory'), error_message)
+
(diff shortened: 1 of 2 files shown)


tests/test-spawn.lua
```diff
@@ -10,6 +10,31 @@ local exit_yay, exit_snd = nil, nil
...
 local steps = {
+  function()
+    -- Test various error conditions. There are quite a number of them...
+    local error_message
+
+    error_message = spawn("this_does_not_exist_and_should_fail")
+    assert(string.find(error_message, 'No such file or directory'), error_message)
+
+    error_message = spawn{"this_does_not_exist_and_should_fail"}
+    assert(string.find(error_message, 'No such file or directory'), error_message)
```
blueyed: Duplicate test?!

tests/test-spawn.lua
```diff
@@ -10,6 +10,31 @@ local exit_yay, exit_snd = nil, nil
...
+
+    error_message = spawn()
+    assert(string.find(error_message, 'No command to execute'), error_message)
+
+    error_message = spawn(" ")
+    assert(string.find(error_message, 'Text was empty'), error_message)
+
+    error_message = spawn{}
+    assert(string.find(error_message, 'There is nothing to execute'), error_message)
+
```
blueyed: There are some tests below already (https://github.com/awesomeWM/awesome/pull/1296/files#diff-4b6e42b8881ecad8f5d1bc29b4550517L68), maybe remove those now?

blueyed: And maybe move the `assert(#client.get() == 0)` from there to here?!

tests/test-spawn.lua
psychon: One is with a string as argument, the other with a table. They take different code paths in the C code (one of them wants to turn the string into a "list of strings", the other directly takes the table to get that "list of strings").

Any idea how to make this difference more prominent?

tests/test-spawn.lua
psychon: Oh, thanks. I missed those. But what's the point of the `assert(#client.get() == 0)`? This test never spawns anything that might open a window.

tests/test-spawn.lua
blueyed: Oh..
> Any idea how to make this difference more prominent?

By using parenthesis? 
`error_message = spawn({"this_does_not_exist_and_should_fail"})`

tests/test-spawn.lua
psychon: Done in a new commit (already pushed)

blueyed: Yes.. but that ensures this.. it's a bit pointless though, I agree.

tests/test-spawn.lua
psychon: That works. Did a force-push to add parenthesis.

PR #1297, opened 2016-12-25
tests: fix incrementing errors

blueyed:
D'oh.


psychon:
So many bashisms. ;-)

Sure: :+1:

psychon:
Sorry for the breakage that I caused. PR is on its way.

Issue #1300, opened 2016-12-25
API: awful.client.visible not fully supported by screen.clients


blueyed:
[`awful.client.visible`](https://github.com/awesomeWM/awesome/blob/69915fb6585ec483844635d02992d7c75cc6fc25/lib/awful/client.lua#L108-L124) is deprecated in favor of the [`screen.clients`](https://github.com/awesomeWM/awesome/blob/69915fb6585ec483844635d02992d7c75cc6fc25/lib/awful/screen.lua#L305-L327) property, but the latter does not provide an option to get the clients in non-stacking order.

Should there be a new property or some method for this on screens?

The same applies to [`awful.client.tiled`](https://github.com/awesomeWM/awesome/blob/69915fb6585ec483844635d02992d7c75cc6fc25/lib/awful/client.lua#L126-L146).

Elv13:
you could add the stacked clients to `screen.object.get_clients`. It will have to be called "the old way" `s:get_clients(false)`. Please note that `.clients` *is* using the stacking order.

PR #1301, opened 2016-12-25
doc: awful.screen: document internal functions for (some) props

blueyed:
Fixes https://github.com/awesomeWM/awesome/issues/1300


Elv13:
(I am away from a computer)

Please use @function client:get_clients

Don't expose the .object in the API. It is an implicit alias of the client
class

blueyed:
@Elv13 
Amended.

PR #1302, opened 2016-12-25
build: output 'Using version from …' message

blueyed:


psychon:
Argh, nice catch. I guess another side-effect of switching to a newer disto and thus a newer version of Git. Thanks for figuring this out!

PR #1303, opened 2016-12-25
tests: fix style/indenting

blueyed:


psychon:
Why just two spaces? Everywhere else (except for some tests) we use four spaces for indentation...
Besides that: :+1:

blueyed:
> Why just two spaces?

My preferred setting, and this file did not have a modeline.

I've now processed all files in `tests/` (`for f in **/*.lua; do vim '+set sw=4' '+norm gg=G' '+wq' "$f"; done`).

PR #1309, opened 2016-12-27
Some improvements for the docs

psychon:
So far this saves large parts of the "my first awesome"-page from the wiki and makes the index page that LDoc generates more useful.

### DIFF
--- a/docs/07-my-first-awesome.md
+++ b/docs/07-my-first-awesome.md
@@ -1,0 +1,179 @@
+# My first awesome
+
+This little tutorial is for people who never used a tiling window manager before
+and don't have fancy hacking skills. We will explore the awesome world of
+awesome step by step and finally come up with a working desktop, providing some
+basic knowledge on how to configure awesome in the process. We will start with
+the default `rc.lua` that comes with awesome, play around a little and try to
+make it more beautiful and useful with widgets. For this tutorial you don't need
+any programming skills besides using a text editor.
+
+## Explore awesome
+
+I will assume that you properly installed awesome through your distribution's
+package manager or compiled it from source. You should now add `exec awesome` to
+your `~/.xinitrc`. As already mentioned, awesome provides a default config file
+which will be our starting point. You will usually find this `rc.lua` file in
+`/etc/xdg/awesome/`. Copy it to `~/.config/awesome/`, and then start X (for
+example by running `startx`).
+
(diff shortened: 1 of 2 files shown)


docs/07-my-first-awesome.md
```diff
@@ -34,46 +34,47 @@ between Ctrl and Alt. You can change the modkey if you want, but we'll get to
 that later. An `xterm` window will pop up. You can of course use your favourite
 terminal if you like. Again, more on that later. Now press *Mod4+Shift+c* to
 close the terminal. This is the command which lets you close any application.
-Awesome has a very elegant way to launch programs: press *Mod4+r*. Now you will
+Awesome has a elegant way to launch programs: press *Mod4+r*. Now you will
```
psychon: an elegant?

docs/07-my-first-awesome.md
```diff
@@ -163,7 +164,7 @@ yourself some trouble by using `Xephyr` to test your changes.
 
 Open a terminal and run the following command:
 
-    Xephyr :1 & sleep 1 ; DISPLAY=:1 awesome
+    Xephyr :5 & sleep 1 ; DISPLAY=:5 awesome
```
psychon: What's so special about display 5?

docs/07-my-first-awesome.md
```diff
@@ -0,0 +1,177 @@
...
+awesome step by step and finally come up with a working desktop, providing some
+basic knowledge on how to configure awesome in the process. We will start with
+the default `rc.lua` that comes with awesome, play around a little and try to
+make it more beautiful and useful with widgets. For this tutorial you don't need
+any programming skills besides using a text editor.
+
+## Explore awesome
+
+I will assume that you properly installed awesome through your distribution's
+package manager or compiled it from source. You should now add exec awesome to
```
psychon: :+1:

psychon:
Updated, squashed your fixup commit and added another new one of my one. The only remaining question is out of curiosity: Why is display 5 better than 1?

docs/07-my-first-awesome.md
blueyed: Yep.

blueyed:
> Why is display 5 better than 1?

I've thought that :1 is more likely to be in use already.. (we're also using :5 for the tests)

Issue #1257, opened 2016-12-06
03-declarative-layout.md has non-deterministic HTML result


psychon:
Ref: #834 

The file `docs/03-declarative-layout.md` contains inclusions like `@DOC_container_WIDGET_LIST@`. These are filed in by `docs/widget_lists.cmake`. However, this does not guarantee a stable (e.g. alphabetically sorted) order of entries. Instead, the order is just what CMake's `file(GLOB` produces, so

An example of this can be seen here: https://github.com/awesomeWM/apidoc/commit/4f897d4796c4f3db8483b87d0e0acdc16fd6f7d7#diff-4505e595f1fd6592002000f5a5a92ed0

Untested patch:
```diff
diff --git a/docs/widget_lists.cmake b/docs/widget_lists.cmake
index b527037..d0fe506 100644
--- a/docs/widget_lists.cmake
+++ b/docs/widget_lists.cmake
@@ -19,6 +19,7 @@ endfunction()
 function(generate_widget_list name)
     file(GLOB ex_files RELATIVE "${SOURCE_DIR}/tests/examples/wibox/${name}/defaults"
     "${SOURCE_DIR}/tests/examples/wibox/${name}/defaults/*")
+    list(SORT ex_files)
 
     # Add the table header
     set(MY_LIST "<table class='widget_list' border=1>\n\
```

blueyed:
Makes sense.
Please create a PR or push it directly.

@Elv13 ?

psychon:
Hm. I cannot really test this. Apparently my file system always produces the same order when iterating over a directory. At least I cannot produce a "random change" due to this, but since we see that (sometimes) on Travis, the patch should still be needed, right?

blueyed:
@psychon 
I think so, yes.
I cannot reproduce it either using the method in https://github.com/awesomeWM/awesome/issues/834#issuecomment-216059802.

psychon:
@blueyed This is about non-determinism in CMake, not in LDoc. I used the following to test for this:
```
cmake -DSOURCE_DIR=$HOME/projects/awesome/ -P ~/projects/awesome/docs/widget_lists.cmake > expect ; for x in $(seq 1 50) ; do cmake -DSOURCE_DIR=$HOME/projects/awesome/ -P ~/projects/awesome/docs/widget_lists.cmake > actual ; cmp expect actual || {echo fail ; break} ; done
```

PR #1312, opened 2016-12-27
Guarantee stable order for docs/03-declarative-layout.md

psychon:
Fixes: https://github.com/awesomeWM/awesome/issues/1257

Still no idea how to test this locally, but the idea sounds good to me...


blueyed:
Thanks!

PR #1329, opened 2016-12-30
Prompt documentation & fixes

Elv13:
I still have a bigger menubar refactor to add to the PR + fix luacheck and travis.

 * Add syntax highlight to the prompt
 * Add more timer documentation (unrelated, but whatever)
 * Add many prompt examples
 * Will eventually add a generic `awful.widget.common` based completion bar

### DIFF
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -273,6 +273,10 @@ local function prompt_text_with_cursor(args)
     local cursor_color = util.ensure_pango_color(args.cursor_color)
     local text_color = util.ensure_pango_color(args.text_color)
 
+    if args.highlighter then
+        text_start, text_end = args.highlighter(text_start, text_end)
+    end
+
     ret = _prompt .. text_start .. "<span background=\"" .. cursor_color ..
         "\" foreground=\"" .. text_color .. "\" underline=\"" .. underline ..
         "\">" .. char .. "</span>" .. text_end .. spacer
@@ -334,6 +338,15 @@ end
 -- @tparam string key The key name.
 -- @tparam string command The current command.
 
+--- A function to add syntax highlighting to the command.
+-- @usage local function my_highlighter(before_cursor, after_cursor)
+--    -- do something
+--    return before_cursor, after_cursor
+-- end
+-- @callback highlighter
+-- @tparam string before_cursor
+-- @tparam string after_cursor
+
 --- Run a prompt in a box.
 --
 -- @tparam[opt={}] table args A table with optional arguments
@@ -346,6 +359,8 @@ end
 -- @tparam[opt] string args.font
 -- @tparam[opt] boolean args.autoexec
 -- @tparam widget args.textbox The textbox to use for the prompt.
+-- @tparam[opt] function args.highlighter A function to add syntax highlighting to
+--  the command.
 -- @tparam function args.exe_callback The callback function to call with command as argument
 -- when finished.
 -- @tparam function args.completion_callback The callback function to call to get completion.
@@ -416,6 +431,7 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
     local text = args.text or ""
     local font = args.font or theme.font
     local selectall = args.selectall
+    local highlighter = args.highlighter
     local hooks = {}
 
     -- A function with 9 parameters deserve to die
@@ -500,7 +516,7 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
     textbox:set_markup(prompt_text_with_cursor{
         text = text, text_color = inv_col, cursor_color = cur_col,
         cursor_pos = cur_pos, cursor_ul = cur_ul, selectall = selectall,
-        prompt = prettyprompt })
+        prompt = prettyprompt, highlighter =  highlighter})
 
     local function exec(cb, command_to_history)
         textbox:set_markup("")
@@ -516,7 +532,7 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
         textbox:set_markup(prompt_text_with_cursor{
                                text = command, text_color = inv_col, cursor_color = cur_col,
                                cursor_pos = cur_pos, cursor_ul = cur_ul, selectall = selectall,
-                               prompt = prettyprompt })
+                               prompt = prettyprompt, highlighter =  highlighter })
     end
 
     grabber = keygrabber.run(
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -83,6 +83,12 @@
 --
 -- @DOC_wibox_awidget_prompt_keypress_EXAMPLE@
 --
+-- **highlighting**:
+--
+-- The prompt also support custom highlighters:
+--
+-- @DOC_wibox_awidget_prompt_highlight_EXAMPLE@
+--
 -- @author Julien Danjou &lt;julien@danjou.info&gt;
 -- @copyright 2008 Julien Danjou
 -- @module awful.prompt
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -1,25 +1,17 @@
 ---------------------------------------------------------------------------
 --- Prompt module for awful.
 --
+-- **Basic usage**:
+--
 -- By default, `rc.lua` will create one `awful.widget.prompt` per screen called
 -- `mypromptbox`. It is used for both the command execution (`mod4+r`) and
 -- Lua prompt (`mod4+x`). It can be re-used for random inputs using:
 --
---    -- Create a shortcut function
---    local function echo_test()
---        awful.prompt.run {
---            prompt       = "Echo: ",
---            textbox      = mouse.screen.mypromptbox.widget,
---            exe_callback = function(input)
---                if not input or #input == 0 then return end
---                naughty.notify{ text = "The input was: "..input }
---            end
---        }
---    end
+-- @DOC_wibox_awidget_prompt_simple_EXAMPLE@
 --
---    -- Then **IN THE globalkeys TABLE** add a new shortcut
---    awful.key({ modkey }, "e", echo_test,
---        {description = "Echo a string", group = "custom"}),
+--     -- Then **IN THE globalkeys TABLE** add a new shortcut
+--     awful.key({ modkey }, "e", echo_test,
+--         {description = "Echo a string", group = "custom"}),
 --
 -- Note that this assumes an `rc.lua` file based on the default one. The way
 -- to access the screen prompt may vary.
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -279,6 +279,61 @@ local function prompt_text_with_cursor(args)
     return ret
 end
 
+--- The callback function to call with command as argument when finished.
+-- @usage local function my_exe_cb(command)
+--    -- do something
+-- end
+-- @callback exe_callback
+-- @tparam string command The command (as entered).
+
+--- The callback function to call to get completion.
+-- @usage local function my_completion_cb(command_before_comp, cur_pos_before_comp, ncomp)
+--    return command_before_comp.."foo", cur_pos_before_comp+3, 1
+-- end
+--
+-- @callback completion_callback
+-- @tparam string command_before_comp The current command.
+-- @tparam number cur_pos_before_comp The current cursor position.
+-- @tparam number ncomp The number of completion elements.
+-- @treturn string command
+-- @treturn number cur_pos
+-- @treturn number matches
+
+--- The callback function to always call without arguments, regardless of
+-- whether the prompt was cancelled.
+-- @usage local function my_done_cb()
+--    -- do something
+-- end
+-- @callback done_callback
+
+---  The callback function to call with command as argument when a command was
+-- changed.
+-- @usage local function my_changed_cb(command)
+--    -- do something
+-- end
+-- @callback changed_callback
+-- @tparam string command The current command.
+
+--- The callback function to call with mod table, key and command as arguments
+-- when a key was pressed.
+-- @usage local function my_keypressed_cb(mod, key, command)
+--    -- do something
+-- end
+-- @callback keypressed_callback
+-- @tparam table mod The current modifiers (like "Control" or "Shift").
+-- @tparam string key The key name.
+-- @tparam string command The current command.
+
+--- The callback function to call with mod table, key and command as arguments
+-- when a key was released.
+-- @usage local function my_keyreleased_cb(mod, key, command)
+--    -- do something
+-- end
+-- @callback keyreleased_callback
+-- @tparam table mod The current modifiers (like "Control" or "Shift").
+-- @tparam string key The key name.
+-- @tparam string command The current command.
+
 --- Run a prompt in a box.
 --
 -- @tparam[opt={}] table args A table with optional arguments
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -485,16 +485,33 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
                     end
                     if match then
                         local cb
-                        local ret = v[3](command)
+                        local ret, quit = v[3](command)
                         local original_command = command
-                        if ret then
-                            command = ret
+
+                        -- Support both a "simple" and a "complex" way to
+                        -- control if the prompt should quit.
+                        quit = quit == nil and (ret ~= true) or (quit~=false)
+
+                        -- Allow the callback to change the command
+                        command = (ret ~= true) and ret or command
+
+                        -- Quit by default, but allow it to be disabled
+                        if ret and type(ret) ~= "boolean" then
                             cb = exe_callback
-                        else
+                            if not quit then
+                                cur_pos = ret:wlen() + 1
+                                update()
+                            end
+                        elseif quit then
                             -- No callback.
                             cb = function() end
                         end
-                        exec(cb, original_command)
+
+                        -- Execute the callback
+                        if cb then
+                            exec(cb, original_command)
+                        end
+
                         return
                     end
                 end
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -55,6 +55,17 @@
 -- Note that this assumes an `rc.lua` file based on the default one. The way
 -- to access the screen prompt may vary.
 --
+-- **Extra key hooks**:
+--
+-- The Awesome prompt also supports adding custom extensions to specific
+-- keyboard keybindings. Those keybindings have precedence over the built-in
+-- ones. Therefor, they can be used to override the default ones.
+--
+-- *[Example one] Adding pre-configured `awful.spawn` commands:*
+--
+-- @DOC_wibox_awidget_prompt_hooks_EXAMPLE@
+--
+--
 -- @author Julien Danjou &lt;julien@danjou.info&gt;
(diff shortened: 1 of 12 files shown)


psychon:
> I still have a bigger menubar refactor to add to the PR + fix luacheck and travis.

You mean "I still have several other PRs to open so I do not create any more monster-PRs", right?

Elv13:
> You mean "I still have several other PRs to open so I do not create any more monster-PRs", right?

As you wish. This PR doesn't have a lot of code, its all doc/tests and some minor things.

Here's the menubar core rewrite prototype https://gist.github.com/Elv13/bfe38dc8d8bc0a02a3ae273df9f65100 . It will need to be split in 2 so the awful.widget.common support is decoupled from the wibar code.

This PR is now ready

Elv13:
@blueyed Can you review the prompt command hook example. It is inspired by a description of your system (I don't have the code, so I may be totally off). Also, there is a lot of content you have a lot of experience with, so I guess your well placed to review it.

About the decreased coverage, it should go away once the accuracy PR is merged. It is due to all the example tests callbacks not being called (there is no way to test them anyway, even if I improved the shims to support take inputs).

lib/awful/prompt.lua
```diff
@@ -1,21 +1,52 @@
...
+--   <tr><td><kbd>CTRL+F</kbd></td><td></td></tr>
+--   <tr><td><kbd>CTRL+H</kbd></td><td>backward-delete-char</td></tr>
+--   <tr><td><kbd>CTRL+K</kbd></td><td>kill-line</td></tr>
+--   <tr><td><kbd>CTRL+U</kbd></td><td>unix-line-discard</td></tr>
+--   <tr><td><kbd>CTRL+W</kbd></td><td>unix-word-rubout</td></tr>
+--   <tr><td><kbd>CTRL+BACKSPACE</kbd></td><td>unix-word-rubout</td></tr>
+--   <tr><td><kbd>SHIFT+INSERT</kbd></td><td>paste</td></tr>
+--   <tr><td><kbd>HOME</kbd></td><td>beginning-of-line</td></tr>
+--   <tr><td><kbd>END</kbd></td><td>end-of-line</td></tr>
+-- </table>
```
blueyed: Can this be a markdown table?

Elv13: Discount supports table. Some time ago I tried to make ldoc display them properly, it didn't. This may be fixed (or not). Also, ldoc fallback to its own implementation if discount isn't installed. That one doesn't support table.

That being said, yes, I would like to use markdown there, but it has to work too...

tests/examples/wibox/awidget/prompt/highlight.lua
```diff
@@ -0,0 +1,66 @@
+local parent    = ... --DOC_NO_USAGE --DOC_HIDE
+local wibox     = require( "wibox"     ) --DOC_HIDE
+local awful     = { prompt = require("awful.prompt") }--DOC_HIDE
+local beautiful = require( "beautiful" ) --DOC_HIDE
+
+    local amp = "&amp"..string.char(0x3B)
+    local quot = "&quot"..string.char(0x3B)
```
blueyed: Why is it indented?
Is it required for doc rendering?

Elv13: Yes, markdown requires 4 spaces for code. Some examples get away without it if they use `@usage`, but `@usage` is used differently for headers (as their documentation explains). This is why that example has a "DISABLE_USAGE" near the top. All lines with DOC_HIDE are hidden, so they don't need the indentation.

Also, before you ask, the ugly `string.char(0xFF)` is a bad hack around a bug. When I will rewrite this system to make @psychon happy, it wont use CMake, I can guarantee you that.

blueyed: Hasn't my tests-cleanup broke those then?!

What about having the examples in the source files, like Python doctests?
And they would be extracted and run from there?
I don't think it is bad to have the requires in there btw, since it makes it easier to adopt, doesn't it?

Elv13: At first I tried to have them in the doc, but ldoc isn't advanced enough to do that. Also, the little magic syntax to add the boilerplate is necessary, or it would require an external template for each test, making things worst instead of better.

I am not saying it is impossible to improve the current system. It is. What I am saying is that before saying "lets ditch it and start over", the replacement has to be *better* than the current solution.

> Hasn't my tests-cleanup broke those then?!

I hope not, some documentation pages are entirely written using it.

Once that "improve CI accuracy" PR is merged, I will open a new one to fix the other @psychon issue about the lines being off in users backtraces. This will also solve the speed issue as it wont run everytime cmake is called. The next steps would be to generate the test list using cmake, then have a single lua process generate all of them. I have a prototype of that, but it has its own share of issues.

As for having them "in" the lua file. It isn't impossible (and it is probably a good thing), but it will take a lot more work than you seem to think to get to the "status quo".

lib/gears/timer.lua
```diff
@@ -105,6 +105,8 @@ local timer_instance_mt = {
 -- @tparam table args Arguments.
 -- @tparam number args.timeout Timeout in seconds (e.g. 1.5).
 -- @tparam[opt=false] boolean args.autostart Automatically start the timer.
+-- @tparam function args.callback Shortcut to avoid having to
+--  use `connect_signal`.
```
psychon: That is not really a good documentation. How about `Callback function to register. Equivalent to connecting to the "timeout" signal of the returned object.` or something like that.

lib/gears/timer.lua
```diff
@@ -1,6 +1,47 @@
...
+--            widget       = wibox.widget.progressbar,
+--        },
+--        {
+--            id           = "mytb",
+--            text         = "100%",
+--            widget       = wibox.widget.textbox,
+--        },
+--        layout      = wibox.layout.stack,
+--        set_battery = function(self, val)
+--            self.mytb.text, self.mypb.value = tonumber(val).."%", tonumber(val)
```
psychon: I'd prefer this to be done in two lines instead of one. Easier to read.

lib/gears/timer.lua
```diff
@@ -1,6 +1,47 @@
...
+--            -- instead of spawning a shell. This is only an example.
+--            awful.spawn.easy_async(
+--                {"sh", "-c", "acpi | sed -n 's/^.*, \([0-9]*\)%/\1/p'"},
+--                function(out)
+--                    mybatterybar.battery = out
+--                end
+--            )
+--        end
+--    }
+--
```
psychon: It's a nice example. :-)

For the commit message ("cannot be tested"):

I suggest adding an `:emit_signal("timeout")`. That way the widget is updated right at startup and half of this could be tested, I guess. Hm, perhaps not tested, but that `emit_signal` is still a good idea.

tests/examples/shims/_common_template.lua
```diff
@@ -1,3 +1,4 @@
+-- luacheck: globals string
```
psychon: Wrong fixup? This change should be in the previous commit.

lib/awful/prompt.lua
```diff
@@ -485,16 +485,33 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
...
                         local original_command = command
-                        if ret then
-                            command = ret
+
+                        -- Support both a "simple" and a "complex" way to
+                        -- control if the prompt should quit.
+                        quit = quit == nil and (ret ~= true) or (quit~=false)
+
+                        -- Allow the callback to change the command
+                        command = (ret ~= true) and ret or command
```
psychon: Sigh, the joy of API compatibility. I don't really understand this and I'm lazy right now, so I'll let @blueyed decide. My only question is: API documentation?

lib/awful/prompt.lua
```diff
@@ -485,16 +485,33 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
...
+
+                        -- Quit by default, but allow it to be disabled
+                        if ret and type(ret) ~= "boolean" then
                             cb = exe_callback
-                        else
+                            if not quit then
+                                cur_pos = ret:wlen() + 1
```
psychon: Why `wlen`? (Almost) nothing else here uses it. I guess you found some test case that requires this?

tests/examples/wibox/awidget/prompt/simple.lua
```diff
@@ -2,6 +2,7 @@ local parent    = ... --DOC_NO_USAGE --DOC_HIDE
 local wibox     = require( "wibox"     ) --DOC_HIDE
 local awful     = { prompt = require("awful.prompt") }--DOC_HIDE
 local beautiful = require( "beautiful" ) --DOC_HIDE
+local naughty   = {} --DOC_HIDE
```
psychon: Unrelated change?

psychon: ...that should be in another commit

lib/awful/prompt.lua
```diff
@@ -65,6 +65,13 @@
 --
 -- @DOC_wibox_awidget_prompt_hooks_EXAMPLE@
 --
+-- *[Example two] Modifiing the command (+ vi like input)*:
+--
+-- The hook system also allows to modify the command before interpreting it in
+-- the `exe_callback`.
+--
+-- @DOC_wibox_awidget_prompt_vilike_EXAMPLE@
+--
```
psychon: Why "WIP" in the commit message?

lib/awful/prompt.lua
```diff
@@ -65,6 +65,13 @@
 --
 -- @DOC_wibox_awidget_prompt_hooks_EXAMPLE@
 --
+-- *[Example two] Modifiing the command (+ vi like input)*:
```
psychon: Less "i", please

tests/examples/wibox/awidget/prompt/vilike.lua
```diff
@@ -0,0 +1,55 @@
...
+local beautiful = require( "beautiful" ) --DOC_HIDE
+local terminal  = "xterm"  --DOC_HIDE
+
+    local atextbox = wibox.widget.textbox()
+
+    -- Store a list of verbs characters in a hash
+    local verbs = {
+        -- Spawn in a terminal
+        t = function(adjs, count, cmd) return {terminal, "-e", cmd} end,  --luacheck: no unused args
+        -- Spawn width a shell
```
psychon: width -> with

lib/awful/prompt.lua
```diff
@@ -72,6 +72,16 @@
 --
 -- @DOC_wibox_awidget_prompt_vilike_EXAMPLE@
 --
+-- *[Example three] Key listener*:
+--
+-- The 2 previous examples were focused on changing the prompt behavior. This
+-- one explains or to "spy" on the prompt events. This can be used for
```
psychon: Should the "or" be a "how"?

tests/examples/wibox/awidget/prompt/keypress.lua
```diff
@@ -0,0 +1,33 @@
...
+    awful.prompt.run {
+        prompt               = "<b>Run: </b>",
+        keypressed_callback  = function(mod, key, cmd) --luacheck: no unused args
+            if key == "Shift_L" then
+                notif = naughty.notify { text = "Shift pressed" }
+            end
+        end,
+        keyreleased_callback = function(mod, key, cmd) --luacheck: no unused args
+            if notif then
+                notif.destroy_cb(naughty.notificationClosedReason.undefined)
```
psychon: Are you trying to destroy the notification? That's not how it works. This only called the `destroy = foo`-callback that you gave to `naughty.notify`, so should be `nil` here.

I think you want `naughty.destroy(notify)` (the reason argument can be nil).

lib/awful/prompt.lua
Elv13: > Sigh, the joy of API compatibility. 

Indeed, we entered the era of "don't break the API" again.

> My only question is: API documentation?

My nice example wasn't working ;) (the completion bar one posted in the gist). The first version of this PR had it (see the https://github.com/awesomeWM/awesome/pull/52 example), but it was already broken by the time the last revision was merged.

lib/awful/prompt.lua
Elv13: Right. To stay compliant with how the other code paths behave.

lib/gears/timer.lua
Elv13: Ok, I will add more magic to be able to add shim commands to `$PATH`. That way we will have a stable output.

> It's a nice example. :-)

(as you know) coming from a genuine question. He was kind of right, there was bits of answer everywhere, but putting them together was non trivial.

The signal/connect part of the code is still undocumented (there is no `client:connect_signal` documentation because I have yet to create a shared version for CAPI, the widgets have one, but without an example)

Elv13:
All done except the timer test. I propose to do that in another PR as it requires changes to the test framework[1] and you said this one is already big enough.

[1] To have stable `easy_async` output.

tests/examples/wibox/awidget/prompt/highlight.lua
```diff
@@ -0,0 +1,66 @@
...
+        awful.prompt.run {
+            prompt       = "<b>Echo: </b>",
+            text         = 'a_very "nice" $SHELL && command', --DOC_HIDE
+            bg_cursor    = "#ff0000",
+            -- To use the default `rc.lua` prompt:
+            --textbox      = mouse.screen.mypromptbox.widget,
+            textbox      = atextbox,
+            highlighter  = function(b, a)
+                -- Add a random marker to delimitate the cursor
+                local cmd = b.."ZZZZZ"..a
```
psychon: By the way: Bad hack! (I just had to mention it, feel free to ignore this comment)

psychon:
Pushed two fixup commits. Squash them and get @blueyed to say that https://github.com/awesomeWM/awesome/pull/1329/commits/e1a58dabff0d8c63d7a5cce03d37f18c95bb7fe2 is good and this is good to go in.

blueyed:
e1a58da is gone, which one was it?

I will review it altogether, but not today anymore.

psychon:
Heh. Your link works. Just in case, it points here: https://github.com/awesomeWM/awesome/commit/e1a58dabff0d8c63d7a5cce03d37f18c95bb7fe2

Elv13:
@blueyed ping? (or just merge it)

The patch basically restore the ability not to exit. We discussed those use cases in Berlin last year.

Elv13:
BUMP (again, sorry for being annoying)

lib/gears/timer.lua
```diff
@@ -104,9 +146,14 @@ local timer_instance_mt = {
 --- Create a new timer object.
 -- @tparam table args Arguments.
 -- @tparam number args.timeout Timeout in seconds (e.g. 1.5).
+-- @tparam[opt=false] boolean args.autostart Automatically start the timer.
+-- @tparam[opt=nil] function args.callback Callback function to connect to the
+--  "timeout" signal.
+-- @tparam[opt=false] boolean args.single_shot Run only once then stop.
```
blueyed: Maybe `once` instead of `single_shot`?

tests/examples/shims/_common_template.lua
blueyed: @Elv13 
addressed?

tests/examples/wibox/awidget/prompt/highlight.lua
blueyed: Yes, it should probably not be used in the docs like that.
Can we do better here?

tests/examples/wibox/awidget/prompt/vilike.lua
```diff
@@ -0,0 +1,55 @@
...
+                    util   = require("awful.util")}--DOC_HIDE
+local beautiful = require( "beautiful" ) --DOC_HIDE
+local terminal  = "xterm"  --DOC_HIDE
+
+    local atextbox = wibox.widget.textbox()
+
+    -- Store a list of verbs characters in a hash
+    local verbs = {
+        -- Spawn in a terminal
+        t = function(adjs, count, cmd) return {terminal, "-e", cmd} end,  --luacheck: no unused args
```
blueyed: It was mentioned somewhere else that the `-e` works differently on terminals.
Have we addressed this somehow already?

tests/examples/wibox/awidget/prompt/vilike.lua
```diff
@@ -0,0 +1,55 @@
...
+        -- Spawn with a shell
+        s = function(adjs, count, cmd) return {awful.util.shell, '-c', cmd} end, --luacheck: no unused args
```
blueyed: Might also behave differently wrt to quoting `cmd` or not.

tests/examples/wibox/awidget/prompt/highlight.lua
Elv13: Any idea? Given the cursor is added to the text and the size of the content can be modified on both sides, there is not a million ways to "preserve" the cursor position and style. I agree it's ugly, but it's also simple and has a comment. Other solutions I can think of would either make the API worst or the example overcomplicated (but cleaner).

tests/examples/wibox/awidget/prompt/vilike.lua
Elv13: At one place, `terminal` was replaced with `urxvt`. This is not a solution, but could be done here too (or just ignore the problem and leave it as-is).

tests/examples/wibox/awidget/prompt/vilike.lua
Elv13: When? This is an argument array, not a parsed string. Both `bash` and `zsh` should handle this.

tests/examples/shims/_common_template.lua
Elv13: yes. GitHub doesn't check if the commit changed (as it would create too many problems). This was done not long after it was reported. This line is now part of `shims: Add the string.wlen function`

tests/examples/wibox/awidget/prompt/highlight.lua
psychon: Only suggestion that I have to offer is to treat both parts separately. That means less highlight in some cases, but also less hacks...?

Elv13: Trading a mild hack for an obvious bug doesn't seems like such a good idea to me. Maybe I should simply replace `ZZZZZ` by a better name (like `ZZZCURSORZZZ`[1])?

[1] I prefer to limit it to a single char category (like upper caps) to avoid too many of the parser expressions breakage. `ZZZ` because it has the lowest risk of collision (beside a PIZZZA typo).

blueyed: `ZZZCURSORZZZ` seems good to me then.

lib/gears/timer.lua
Elv13: Qt call it `singleShot` and glib doesn't have it. My naming process is usually glib->qt->random. `once` isn't a bad name, but so is `single_shot`. If you *really* want me to change it, I will, but right now I left it as it is.

Elv13:
Ok, rebased again with the ZZZZZ change. I think I addressed most issues. The last one is the terminal `-e`. There is much to do for this one. It *is* a problem, but there is no obvious solution beside extending `awful.spawn` to support it. This is out-of-scope for this PR. I could add a comment warning users about the issue, or just ignore it.

blueyed:
@Elv13 
👍

PR #1330, opened 2016-12-30
Add error checking in 05-awesomerc.md.lua

psychon:
io.open returns nil, an error message and an error code on failure. This
perfectly fits assert which will make the script fail with the error
message. Previously it would fail with "attempted to index a nil value".

Reference: https://www.reddit.com/r/awesomewm/comments/5kyqji/cant_build_git_awesome/


blueyed:
Thanks!

PR #1333, opened 2016-12-30
_multi_screen: Help the garbage collector

blueyed:
Tests using _multi_screen end up with lots of screens, which also means
that they end up with lots of wiboxes, widgets and so on. Since
drawables are kept around in a weak table and are forced to redraw when
screens change, this causes slowdowns later on when we end up with lots
of drawables.

Help the garbage collector in cleaning up by giving it a slight kick in
the right moment.

Via https://github.com/awesomeWM/awesome/pull/1239#discussion_r94220308.

### DIFF
--- a/tests/_multi_screen.lua
+++ b/tests/_multi_screen.lua
@@ -411,7 +411,10 @@ local function add_steps(real_steps, new_steps)
     for _, dispo in ipairs(dispositions) do
         -- Cleanup
         table.insert(real_steps, function()
-            if #client.get() == 0 then return true end
+            if #client.get() == 0 then
+                collectgarbage("collect")
+                return true
+            end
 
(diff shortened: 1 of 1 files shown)


blueyed:
@psychon 
Can you tell if it helps like this?

psychon:
Well...

> 2016-12-30 14:17:07 W: awesome: a_glib_poll:389: Last main loop iteration took 6.095839 seconds! Increasing limit for this warning to that value.

I'm not sure if this commit helps anything. I thought it does not, but Elv13 it fixes the problem for him locally. Adding it "just in case" seems wrong.

Elv13:
Better close this PR. @psychon is right that this is a hack. It doesn't change anything on the CI, so there is no point. When luacov is not running, there is no problem.

PR #1334, opened 2016-12-30
Travis: unpin ldoc

blueyed:
Via #1239.


psychon:
:+1:
This way we notice failures due to new releases early. It's annoying when that happens at a random change, but better than not happening at all.

blueyed:
Agreed.

PR #1342, opened 2016-12-31
Fix missing workarea updates with client struts

### LINKED ISSUE #1335: Docking windows affects other tags

    awesome devel (Harder, Better, Faster, Stronger)
     • Compiled against Lua 5.2.3 (running with Lua 5.2)
     • D-Bus support: ✔
     • execinfo support: ✔
     • RandR 1.5 support: ✔
     • LGI version: 0.7.2

(Build from current master branch)

* make window dockable on tag 2
* dock window on tag 2
* switch to tag 1
* open window there

**Actual result:**

* the window ignores the spaces taken up by a docked window on tag 2

**Expected result:**

* The fact that something is docked on tag 2 doesn't affect tag 1 at all

**Note:**
I'm not really sure that my idea of what should happen is correct but the current behaviour seems strange to me.

psychon:
This fixes #1335 and all other, similar issues that I could find. The `TEST_PREV_COMMITS=1` build will fail, but I cannot really do much about it, I guess. My commits are small and each one fixes some partial bug.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1797,6 +1797,8 @@ client_set_sticky(lua_State *L, int cidx, bool s)
     {
         c->sticky = s;
         banning_need_update();
+        if(strut_has_value(&c->strut))
+            screen_update_workarea(c->screen);
         luaA_object_emit_signal(L, cidx, "property::sticky", 0);
     }
 }
--- a/objects/tag.c
+++ b/objects/tag.c
@@ -321,6 +321,7 @@ tag_client(lua_State *L, client_t *c)
     client_array_append(&t->clients, c);
     ewmh_client_update_desktop(c);
     banning_need_update();
+    screen_update_workarea(c->screen);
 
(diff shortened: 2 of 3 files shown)


Elv13:
Coverall seems MIA. Beside that 👍

blueyed:
> The `TEST_PREV_COMMITS=1` build will fail, but I cannot really do much about it, I guess. My commits are small and each one fixes some partial bug.

Hmm, but it could be fixed by squashing the fixes and the changed test?!

blueyed:
@Elv13 
> MIA

Please elaborate.

Elv13:
> Please elaborate.

It is not in the "check" section. Chances are it crashed on their cloud and we have nothing to do about it. I just noticed it wasn't listed. I restarted the build, lets see how this goes.

Elv13:
Still missing (I don't see how this PR could have broken it, but other PRs from today have it, so it isn't a breakage that's already introduced in `master`)

PR #1343, opened 2016-12-31
WIP: Test runner speed

psychon:
This is my trying to speed up the test runner by getting of its 0.1 seconds time interval. The PR only exists so that I get a Travis build. Perhaps someone will even check if the build is faster than our current builds so that I do not have to do that? :-)
@blueyed Feel free to re-do https://github.com/awesomeWM/awesome/commit/88db9023005d7204f69469c341ed5eaf559f640a and merge it to master, or something.

P.S.: Everyone, have a happy new year.

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -374,6 +374,8 @@ a_glib_poll(GPollFD *ufds, guint nfsd, gint timeout)
         lua_settop(L, 0);
     }
 
+    xcb_aux_sync(globalconf.connection);
+
     /* Don't sleep if there is a pending event */
     assert(globalconf.pending_event == NULL);
     globalconf.pending_event = xcb_poll_for_event(globalconf.connection);
--- a/tests/_runner.lua
+++ b/tests/_runner.lua
@@ -1,5 +1,7 @@
 local timer = require("gears.timer")
 local awful = require("awful")
+local glib = require("lgi").GLib
+local protected_call = require("gears.protected_call")
 
 local runner = {
     quit_awesome_on_error = os.getenv('TEST_PAUSE_ON_ERRORS') ~= '1',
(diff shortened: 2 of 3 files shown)


blueyed:
@psychon 
> P.S.: Everyone, have a happy new year.

Thanks!  Same for you and all awesome contributors! :)

Elv13:
> Perhaps someone will even check if the build is faster than our current builds so that I do not have to do that? :-)

After 3 iterations (compared that with that structs PR), it doesn't seem to have a statistically significant difference, but I guess it can't hurt. Thanks for this. As proven by all the "slow event loop iterations" warnings, the time is spent mostly in the lua code (and luacov itself), not in the sleeps. Cutting the idle time can only do so much.

blueyed:
Restarted the  build, let's see if it is really random.

Elv13:
I restarted it a few time and it passed, the failing target isn't always the same. So yes, it is really "random" (for some "this is a race condition" definition of "random")

psychon:
>> Feel free to re-do 88db902 and merge it to master, or something.

> What do you mean?
> Seems weird to require this?!

We have `set -e`. `((errors++))` returns the old value. So the first time this is done, the command is considered to have "failed" and `run.sh` aborts. Thus, after the first test failure, the runner aborts and following tests are not executed.
Fix: The usual (and ugly) `|| true`.

psychon:
Superseded by #1345. There I just use a timer with timeout zero instead of doing magic with a glib idle callback.

psychon:
Oh and: Bummer that this does not help on Travis (I guess the other PR then also does not help). Thanks for checking and sorry that I did not have the time yesterday to look into it myself! For the new PR I at least made sure that it helps on my local computer.

PR #1345, opened 2017-01-01
Test runner speedup

psychon:
The quick summary: Running `tests/run.sh` now only takes 60 seconds on my computer while previously it needed 100 seconds to finish.

This supersedes #1343.

### DIFF
--- a/tests/_runner.lua
+++ b/tests/_runner.lua
@@ -32,7 +32,7 @@ end)
 runner.run_steps = function(steps)
     -- Setup timer/timeout to limit waiting for signal and quitting awesome.
     -- This would be common for all tests.
-    local t = timer({timeout=0.1})
+    local t = timer({timeout=0})
     local wait=20
     local step=1
     local step_count=0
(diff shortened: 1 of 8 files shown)


tests/run.sh
```diff
@@ -228,7 +228,7 @@ for f in $tests; do
             grep -q -E '[Ee]rror|assertion failed' "$awesome_log"; then
         echo "===> ERROR running $f <==="
         grep --color -o --binary-files=text -E '.*[Ee]rror.*|.*assertion failed.*' "$awesome_log" || true
-        ((errors++))
+        ((errors++)) || true
```
blueyed: `(( ++errors ))` would not require `|| true` here.

tests/_runner.lua
```diff
@@ -32,7 +32,7 @@ end)
 runner.run_steps = function(steps)
```
blueyed: Commit message (ff4ad53af8): s/we are done to 60 seconds/we are down to 60 seconds/

psychon:
Thanks. Both comments should be addressed.

PR #1347, opened 2017-01-01
Revert some of the indentation changes

Elv13:
The doc wasn't generated correctly anymore. The tests examples are using markdown, some spaces are required by the RFC. Also the DOC_HIDE macro doesn't count in the indentation. Plus some other spaces were also necessary. I fixed that in `CMakeLists.txt` instead of the code.

The "end of line whitespace" commit from a while ago also broke the empty line support. This is still not fixed.

### DIFF
--- a/tests/examples/wibox/widget/progressbar/vertical.lua
+++ b/tests/examples/wibox/widget/progressbar/vertical.lua
@@ -3,18 +3,17 @@ local wibox  = require("wibox") --DOC_HIDE
 
 parent:add( --DOC_HIDE
 
-           wibox.widget {
-               {
-                   max_value     = 1,
-                   value         = 0.33,
-                   widget        = wibox.widget.progressbar,
-               },
-               forced_height = 100,
-               forced_width  = 20,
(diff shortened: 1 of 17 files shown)


tests/examples/wibox/widget/progressbar/vertical.lua
```diff
@@ -3,16 +3,16 @@ local wibox  = require("wibox") --DOC_HIDE
...
-                   max_value     = 1,
-                   value         = 0.33,
-                   widget        = wibox.widget.progressbar,
-               },
-               forced_height = 100,
-               forced_width  = 20,
-               direction     = "east",
-               layout        = wibox.container.rotate,
-           }
+wibox.widget {
```
blueyed: Why isn't this indented?

Elv13: Everything with DOC_HIDE isn't part of the example. It is removed before being added to the doc. In This case, as the indentation is uniform, it doesn't change the output, but in some other places it did.

PR #1355, opened 2017-01-02
Remove menubar.get [API break]

psychon:
The function seems useless and its documentation is wrong. It does not
return a wibox, but a widget. Also, the widget cannot really be used on
its own since it depends on the size of the wibox.

So menubar.get() and its wrapper through the metatables __call should
just be removed.

DO NOT MERGE THIS
It conflicts with #1354 which should be merged first. I only have this open already to start some discussion.

@Elv13 What should our strategy regarding API breaks be? Do we create a branch with the great name "4" in git and then we can happily break API in master?

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -35,7 +35,7 @@ local function get_screen(s)
 end
 
 -- menubar
-local menubar = { mt = {}, menu_entries = {} }
+local menubar = { menu_entries = {} }
 menubar.menu_gen = require("menubar.menu_gen")
 menubar.utils = require("menubar.utils")
 local compute_text_width = menubar.utils.compute_text_width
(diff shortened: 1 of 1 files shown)


psychon:
Oh and of course: Can someone please point out the obvious use case for `menubar.get()` that I am missing?

Elv13:
I propose extending `awful.util.deprecate` to have an "API level" that auto-triggers the deprecation messages and/or an error. Everything currently deprecated should be deprecated as of v4, error in v5. Everything currently not deprecated would either be "deprecated in v5" or just "error in v5" in case the API is useless and/or broken by design.

This way we can add some new APIs that deprecates others while keeping the API stable for now. I think it is a good compromise between "not merging breaking changes to keep using master" and "fork, then let stable rot for 4 years".

This will also make it easier to track when time comes to drop the old code.

blueyed:
#1354 is merged, so this could be rebased.
`deprecated_in` is ready in https://github.com/awesomeWM/awesome/pull/1586.

psychon:
Superseded by https://github.com/awesomeWM/awesome/pull/1620

PR #1363, opened 2017-01-05
Remove obsolete include and Xlib comment.

Airblader:


objects/key.c
```diff
@@ -38,7 +38,7 @@
 #include "common/xutil.h"
 #include "xkb.h"
 
-/* XStringToKeysym() and XKeysymToString */
+/* XStringToKeysym() */
 #include <X11/Xlib.h>
```
psychon: Some day (tm) I will also get rid of that one. But today is not that day...

Airblader: One of the main reasons xcb doesn't have this yet, BTW, was the fact that xcb had no xrm equivalent. It should now be portable to XCB according to a discussion from 2008. :-)

Elv13:
Thanks!

Welcome to the AwesomeWM contributors club ;) <sub>(yes, I know)</sub>

Airblader:
@psychon helped us out over at i3 so many times, it's the least I can do. :-)

PR #1366, opened 2017-01-08
Add taglist theme variables for volatile tags

thekrampus:
If a user wants to apply some taglist theming specifically for volatile tags, there currently isn't any nice way of doing that. Ergo, this.

Adds beautiful variables for taglist volatile elements, mirroring those for empty and urgent elements. Selected and urgent tag theming overrides volatile theming, which in turn overrides occupied theming.

(long-time user, first-time contributor. i don't usually like to show up out of nowhere with a feature addition, but this seems self-explanatory...)

### DIFF
--- a/lib/awful/widget/taglist.lua
+++ b/lib/awful/widget/taglist.lua
@@ -69,6 +69,16 @@ taglist.filter = {}
 -- @param color
 -- @see gears.color
 
+--- The tag list volatile elements background color.
+-- @beautiful beautiful.taglist_bg_volatile
+-- @param color
+-- @see gears.color
+
+--- The tag list volatile elements foreground (text) color.
+-- @beautiful beautiful.taglist_fg_volatile
+-- @param color
+-- @see gears.color
+
 --- The selected elements background image.
 -- @beautiful beautiful.taglist_squares_sel
 -- @param surface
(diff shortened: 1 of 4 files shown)


Elv13:
> (long-time user, first-time contributor. i don't usually like to show up out of nowhere with a feature addition, but this seems self-explanatory...)

Well, you had to start somewhere. People starting with patches instead of whining is always a big plus in my book ;)

As for the missing test case, I wont blame you for it. The taglist has tests (by accident, given it is used in the default config), but don't have its own suit in `spec/` or (most likely) `tests/example/wibox/awidget`. So there is currently no good place for it to go anyway.

thekrampus:
Right on. I'd written this for my own use after moving to version 4 and figured I ought to make a PR on a whim -- hadn't even thought about coverage. *Mea culpa.*

Elv13:
> Right on. I'd written this for my own use after moving to version 4 and figured I ought to make a PR on a whim

Those PRs are welcome. The legends are true, my 3.1-3.5 config indeed had a (almost) full rewrite of the whole Awesome Lua userspace. Tens of thousands of lines of it. For 4.0 I finally decided to develop upstream. It's worth it and our improvements can reach the other users. The volatile tags was one of those feature my config had[1] while the mainstream `awful` didn't ;).

[1] Provided by the Tyrannical extension

Issue #1340, opened 2016-12-31
size_hints_honor issue


vitek:
Output of `awesome --version`:

awesome v4.0-18-gb4387d3 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.2.4 (running with Lua 5.2)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✘
 • LGI version: 0.9.0

**How to reproduce the issue:**

With global `size_hints_honor = false` and `awful.layout.suit.tile` layout open two `gnome-terminals` then try to resize them with `Mod-L`.

**Actual result:**

Window content is corrupt. Left terminal's right edge is painted with parts of other teminal content.

**Expected result:**

Both terminals painted as expected.

psychon:
Got a screenshot if this issue? Can you try if it also happens when starting awesome with `--no-argb`?

vitek:
The same behaviour with `--no-argb` here is how it looks like:

![2016-12-31 15-40-18](https://cloud.githubusercontent.com/assets/489439/21577687/26853254-cf75-11e6-80d3-d1773cc021d0.png)

vitek:
Same problem on both ubuntu 14.04 and 16.04 while with awesome 3.4 and 3.5 it's ok.

I'm also experience problems with emacs in tile layout (size_hints_honor=true) sometimes it starts like this:

![emacs1](https://cloud.githubusercontent.com/assets/489439/21748189/4b1776ac-d58f-11e6-86c2-288f40fdcc8c.png)

If I open terminal emacs window is ok now:

![emacs2](https://cloud.githubusercontent.com/assets/489439/21748194/66f829ca-d58f-11e6-8ef3-dc92035e1413.png)

psychon:
So the failed ConfigureWindow request seems to be the problem (possibly?), but so far I have no idea why it would fail...

vitek:
Here is xtrace output for emacs issue:

```
000:<:5ef9: 28: Request(12): ConfigureWindow window=0x0280005e values={x=0 y=17 width=0 height=0}
000:<:5efa: 28: Request(12): ConfigureWindow window=0x01e000ab values={x=0 y=0 width=0 height=0}
000:<:5efb: 44: Request(25): SendEvent propagate=false(0x00) destination=0x01e000ab event-mask=StructureNo
[...]

PR #1367, opened 2017-01-08
Fix some possible errors with resizes

psychon:
This fixes at least the X11 errors part from #1340 (at least I think so, original reporter didn't test this patch yet).

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1678,6 +1678,17 @@ client_resize(client_t *c, area_t geometry, bool honor_hints)
     if(geometry.y + geometry.height < 0)
         geometry.y = 0;
 
+    if (honor_hints) {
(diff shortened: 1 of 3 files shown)


vitek:
Created its own issue for this issue https://github.com/awesomeWM/awesome/issues/1368

vitek:
Cannot reproduce the issue with this PR

psychon:
> Cannot reproduce the issue with this PR

Which one exactly? #1368?

vitek:
Yes, #1368. I've created its own issue for each problem

blueyed:
Sounded like this should have `Fixes: https://github.com/awesomeWM/awesome/issues/1368` when merging - so I did that.

Issue #1234, opened 2016-11-22
"make package" produces a non-installable RPM


psychon:
Quote from https://github.com/awesomeWM/awesome/issues/1225#issuecomment-262190743
```
file /etc/xdg from install of awesome-3.5.2~git2189_gd59f943-1.x86_64 conflicts with file from package filesystem-3.2-37.fc24.x86_64
file /usr/share/xsessions from install of awesome-3.5.2~git2189_gd59f943-1.x86_64 conflicts with file from package filesystem-3.2-37.fc24.x86_64
```

psychon:
Google to the rescue!
```diff
diff --git a/Packaging.cmake b/Packaging.cmake
index 102c033..6710948 100644
--- a/Packaging.cmake
+++ b/Packaging.cmake
@@ -60,6 +60,7 @@ endif(NOT CPACK_GENERATOR)
 set(CPACK_PACKAGE_NAME "awesome")
 set(CPACK_DEBIAN_PACKAGE_MAINTAINER "devnull@example.com")
 set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "A tiling window manager")
+set(CPACK_RPM_EXCLUDE_FROM_AUTO_FILELIST_ADDITION "/etc/xdg;/usr/share/xsessions")
 
 if(CPACK_GENERATOR)
        include(CPack)
```

PR #1370, opened 2017-01-08
Fix "make package" with RPMs

psychon:
In contrast do dpkg, rpm tracks which package created a directory. No
idea what happens if you remove the package which owns the directory,
but another package still owns files in there. Anyway, this behaviour
produced the following problem:

    file /etc/xdg from install of awesome-[snip].x86_64 conflicts with
    file from package filesystem-3.2-37.fc24.x86_64

Fix this by telling CPack to exclude these directories from the file
list.

Fixes: https://github.com/awesomeWM/awesome/issues/1234


blueyed:
Thanks for tracking this down!

PR #1371, opened 2017-01-08
Modify package.cpath

psychon:
This makes us do similar things to `package.cpath` as we already do to `package.path`.


blueyed:
i trust you on this!

Sorry, accidentally squashed..!

PR #1372, opened 2017-01-08
awful.tag: handle invalid client in delayed_call

blueyed:
A short-lived client (e.g. `urxvt -e false`) might be invalid already
when the delayed callback is called.

TODO:
 - [ ] test


blueyed:
> Do you intend to add a test before merging?

Yes, got stuck with that when trying it right away.

blueyed:
I might come back to finishing the test, but let's get it fixed already.

Issue #1374, opened 2017-01-09
A way to speed up tail --pid in the test suite


psychon:
Running `tests/run.sh` on this computer takes about 55 seconds. After the following patch, this is down to 25 seconds:
```diff
diff --git a/tests/run.sh b/tests/run.sh
index ed2a604..fef8782 100755
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -210,7 +210,7 @@ for f in $tests; do
     DISPLAY=$D "$AWESOME_CLIENT" 2>&1 < "$f"
 
     # Tail the log and quit, when awesome quits.
-    tail -n 100000 -f --pid "$awesome_pid" "$awesome_log"
+    tail -n 100000 -s 0.01 -f --pid "$awesome_pid" "$awesome_log"
 
     set +e
     wait $awesome_pid
```
I cannot quite explain the speed-up by 30 seconds, but I'll try to explain parts of it:

`tail --pid 1234` works by checking every `n` seconds if the process with the given PID is still alive. If not, it exits. Assuming "exit of the process" is a uniformly distributed event (it is not), the average delay that `tail` needs to notice that the process exited is `n/2`.
According to the man page, the default is `n=1` and we have 30 tests, so the expected delay due to `tail` is `30/2 = 15 seconds`.
However, this obviously does not explain the 30 seconds speed-up that I saw in a quick experiment.
(Random idea: Most tests are "immediately done" (e.g. `test-leaks.lua`) and for these there is a high probability that almost a whole second is needed before `tail` notices)

Before something like this is merged, I guess we should make some (rough) measurements about what is the maximal argument for the `--sleep-interval` argument which still provides quite a good speed-up (on some random computer). I bet that `0.01` is way too low.

Anyone wants to find the sweet spot for this parameter?

blueyed:
Great find.
I think we should start with `0.05`, but `0.01` is also find: after all it's probably quite cheap for `tail` to check if the process is still there?!

On CI we could also skip the `tail` completely probably, and just cat it afterwards?!

psychon:
I did some measurements with the following command and patch:
```sh
export CI=true LC_ALL=C ; for x in $(seq 1 100) ; do TAIL_ARG=$(printf "-s %.2f " $(($x/100.0))) ; /bin/time -f "${TAIL_ARG} %E" sh -c 'tests/run.sh >/dev/null 2>/dev/null' ; done
```
```diff
diff --git a/tests/run.sh b/tests/run.sh
[...]

PR #1376, opened 2017-01-10
[TEST] tests: Add a very dirty hack to break wibox for multiscreen tests

Elv13:
This commit add an option to shim the whole wibox module when
running multi-screen tests. This is intended to lower the test
runtime when coverage is enabled.

In theory, most of that code is already covered by the
test-screen-changes suit.

It is a reaction to #1374. My issue with #1374 is that while it saves 30 seconds and is a low hanging fruit, it fixes a linear issue depending on the number of test suits. This commit is an ugly hack, but I am curious to see how well it fares in Travis. In theory, the slowdown is exponential depending on how slow the target is.

The implementation is crap, it is intended to see the effect it has on the CI time.

### DIFF
--- a/tests/_multi_screen.lua
+++ b/tests/_multi_screen.lua
@@ -517,6 +517,29 @@ local function add_steps(real_steps, new_steps)
     end
 end
 
+-- This is a very ugly hack to speed up the test. Luacov get exponentially
+-- slower / more memory hungry when it has more work to do in a single test.
+-- A lot of that work is building wibars and widgets for those screen. This
+-- has value when tried once (already covered by the test-screen-changes suit),
+-- but not done 180 times in a row. This code monkey-patch `wibox` with the
+-- intent of breaking it without causing errors. This way it stops doing too
+-- many things (resulting in a faster luacov execution)
+function module.disable_wibox()
+    local awful = require("awful")
+
+    setmetatable(wibox, {
+        __call = function() return {
+            geometry = function()
+                return{x=0, y=0, width=0, height=0}
+            end,
+            set_widget = function() end,
+            setup = function() return {} end
+        }
+    end })
+
+    awful.wibar = wibox
+end
+
 return setmetatable(module, {
                     __call = function(_,...) return add_steps(...) end
                 })
--- a/tests/test-awful-client.lua
+++ b/tests/test-awful-client.lua
@@ -146,6 +146,8 @@ table.insert(steps, function()
     return true
 end)
 
+
+
 local multi_screen_steps = {}
 
 -- Add a test client on each screen.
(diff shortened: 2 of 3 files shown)


Elv13:
num | coveralls | codecov
----- | ------------ | ---------
Run 1 | 7:21 | 7:40

Elv13:
Ok, so, the results are in:

 * it is up to ~50% faster
 * variance has been reduced[1]
 * test-awful-tag is no longer a bottleneck

[1] Before, the multi-screen tests had a tendency to page-fault or something that made their runtime very unpredictable, it is now exactly the same each time (over 3 runs). test-awful-layout is now equally slow with test-awful-client.

The runtime is now less or more this:

20% setup (104/431)
11% cmake (26/431)
14% busted / luacheck / make / example tests (61/431)
55% Integration tests (240/431)

So, here is what can be done next

 * Use docker to maybe save most of those 104 seconds (assuming the AWS node already has the docket image cached)
 * Save the 30 seconds using #1374 runner changes
 * Save maybe 15 seconds by using less process for the doc tests, but really, with the current stats it isn't even worth it (until I start to add more of these again, then it will be relevant again, I will have to do it eventually)

So, this test is a success. The question now is if my terrible patch is good enough or is it worth baking this directly into wibox to have something "cleaner".

psychon:
I don't have the time right now, but the following patch might have a similar effect (less code is short-circuited, but hopefully all the slowness goes away):
```diff
diff --git a/lib/wibox/drawable.lua b/lib/wibox/drawable.lua
index 330edc3..caba397 100644
--- a/lib/wibox/drawable.lua
+++ b/lib/wibox/drawable.lua
@@ -376,7 +376,7 @@ function drawable.new(d, widget_context_skeleton, drawable_name)
 
     -- Connect our signal when we need a redraw
     ret.draw = function()
-        if not ret._redraw_pending then
+        if not ret._redraw_pending and not drawable.disable_all_drawing then
             timer.delayed_call(ret._do_redraw)
             ret._redraw_pending = true
         end
```
(Obviously this variable still has to be set in the right place; its result is that all "I need a redraw"-requests are just ignored)

Elv13:
Given we had multiple build failures over the last few days due to this. I will bring up this issue again. The advantage of this hack compared to psychon counter-proposal is that it's both faster and less intrusive. This is really a band-aid and nothing else. It aims to fix an issue that isn't really ours. While it will eventually break in our face, we have failing build right now and this fixes it. Nobody likes this patch, but something "good enough now" is better than "maybe a magical solution will arise later".

Elv13:
I will merge this in 2 days if nothing else is merged first to solve the problem. Now over 80% of the builds fail due to this except for a few hours in the morning (UTC-5). It's pointless to keep this out of tree. The code doesn't touch the "release code" at all.

Elv13:
/me sad about the hack, but relieved about having not having build failures due to limited resources anymore.

PR #1379, opened 2017-01-10
FreeBSD doesn't have SIGPOLL

rbgarga:


luaa.c
```diff
@@ -612,7 +613,9 @@ setup_awesome_signals(lua_State *L)
 
     /* POSIX.1-2001, according to man 7 signal */
     SETUP_SIGNAL(SIGBUS);
+#ifndef __FreeBSD__
     SETUP_SIGNAL(SIGPOLL);
+#endif
```
blueyed: Shouldn't this check for if `SIGPOLL` is defined?!  (and then leave a comment about FreeBSD there only)

rbgarga: Yeah, it should be fine and we will avoid extra include. Let me change it

psychon:
Sigh. And I even checked the man page... Does this mean that FreeBSD does not follow SUSv2 / POSIX.1-2001?

Anyway, thanks for fixing this! I hope this is the worst problem that we bring to you and building onFreeBSD.

blueyed:
Thanks!

rbgarga:
@psychon I didn't follow all the history, but there is an old discussion here:
https://lists.freebsd.org/pipermail/freebsd-standards/2011-January/002098.html

Thanks!

PR #1381, opened 2017-01-10
Fix build on FreeBSD

rbgarga:
Include <sys/wait.h> on FreeBSD to fix build, it's required for
WIFEXITED and another couple of macros

### DIFF
--- a/spawn.c
+++ b/spawn.c
@@ -53,6 +53,11 @@
 
 #include "spawn.h"
 
+#include <sys/param.h>
+#ifdef __FreeBSD__
+# include <sys/wait.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


psychon:
My man page for `waitpid` (which also explains e.g. `WIFEXITED`) mentions these two headers:
```C
#include <sys/types.h>
#include <sys/wait.h>
```
Could you check if unconditionally including these two is enough and if so, replace your patch with that? (There must be *some* reason why the man page mentions them :-) )

rbgarga:
It's on to add them unconditionally, I'll update the patch soon

blueyed:
Thanks!

Issue #1369, opened 2017-01-08
Emacs menubar sometimes broken in tile layout


vitek:
Output of `awesome --version`:

awesome v4.0-65-g8941c21 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.2.4 (running with Lua 5.2)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✘
 • LGI version: 0.9.0

**How to reproduce the issue:**

Start emacs few times in tile layout. With ``` '(menu-bar-mode nil)``` in emacs.el to disable menubar. Sometimes it starts with menubar shown but unusable. It looks like awesome didn't handle emacs window layout change for some reason. 

**Actual result:**

Right after emacs started:
![emacs-menubar2](https://cloud.githubusercontent.com/assets/489439/21749822/3066b63c-d5b8-11e6-87fc-139b7a9d8299.png)

After workspace switch and back:
![emacs-menubar](https://cloud.githubusercontent.com/assets/489439/21749716/53763114-d5b5-11e6-8dd0-c31d416c27e4.png)

**Expected result:**

Normal emacs window without menubar.

vitek:
Another way to reproduce:
* start emacs with `emacs -q`
* menubar works as expected
* Issue `M-x menu-bar-mode RET` to disable menubar
* menubar is still there but does not work just like on original screenshots

PR #1388, opened 2017-01-12
Reply on configurerequest even if geometry is the same

vitek:
Becuse `ConfigureRequest` are handled in async way some requests may be left without reply in case lua code disacrds new geometry. This PR makes sure such requests are replied.

Fixes https://github.com/awesomeWM/awesome/issues/1369 and https://github.com/awesomeWM/awesome/issues/1340

### DIFF
--- a/event.c
+++ b/event.c
@@ -395,9 +395,8 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
                 geometry.y += diff_y;
         }
 
-        if(!client_resize(c, geometry, false))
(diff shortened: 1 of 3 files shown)


vitek:
Our versions are very similar. The difference is `got_configure_request` instead of `geometry_requested` I think you name is better. Also you moved `client_send_configure()` out of `event.c` which is also a good idea.

So I'll switch to your version.

vitek:
I've update PR

blueyed:
Thanks a lot!

PR #1390, opened 2017-01-12
build-utils/travis-apidoc.sh: ignore README.md in diff

blueyed:


psychon:
This works as long as nothing else "touches this". It would be nice if the scripts would just always make sure we have the right "state".

But whatever...

blueyed:
Men, it was removed again: https://github.com/awesomeWM/apidoc/commit/e1ee8a444f173fe6bccace55a0b9845979e0a39d

PR #1428, opened 2017-01-22
Travis: install lua-discount before ldoc

blueyed:
This should not pull in "markdown" and might fix
https://github.com/awesomeWM/awesome-www/issues/68#issuecomment-274354906.


blueyed:
Creating a PR from the main repo.

PR #1432, opened 2017-01-23
import vim hotkeys from user config, not hotkeys_popup module

actionless:
i remember one of the users was complaining what they are enabled by default


actionless:
i forecast wht it will reduce the coverage for hotkeys module to 0% :)

Elv13:
I am not sure adding vim related stuff to rc.lua is such a good idea, it's out of scope and confusing.

actionless:
IMO it's better than loading vim hotkeys by default when loading widget itself

actionless:
any more opinions?

actionless:
let's see if any of users will complain what vim hotkeys disappeared in their local configs, mb we should put that into release notes for 4.1

Issue #1400, opened 2017-01-15
Avoid cyclic dependencies between modules


Elv13:
gears.color->beautiful.init->beautiful.theme_assets->gears.color = BOOM

How this got past the CI...

Simplest solution, don't require theme_assets in beautiful.init

Elv13:
nevermind, it can't happen on master

psychon:
Should we define some hard rules on what might depend on what to make sure we have no cyclic dependencies? For example, I would propose that gears must not use beautiful...

Elv13:
Sure. It will however require some changes as gears depends on awful.util.

gears: depends on nothing
wibox: depends on gears
beautiful: depends on gears
awful: depends on everything

`awful.util` deprecation support can be moved into `gears.deprecation`. `awful.util.table` should be moved to `table` itself. `awful.util` string utils should move into `string`. This can be done without breaking the API (with aliases).

actionless:
why not to move `awful.util` to just `util`?

psychon:
I don't think we should start monkey-patching standard Lua things. We already do that too much. That leads to too much confusion. Also 'util' is too much of a generic name that eventually some name collision will occur.

psychon:
(My plan is to implement a tool that tries to automatically check the above rules and complains if they are not followed.)

psychon:
Following program run in the awesome source directory:
```lua
local depgraph = require("depgraph")

local allowed_deps = {
	gears = {
		lgi = true,
	},
	beautiful = {
		gears = true,
		lgi = true,
	},
	wibox = {
		beautiful = true,
		gears = true,
		lgi = true,
	},
	awful = {
		beautiful = true,
		gears = true,
		lgi = true,
		wibox = true,
	},
	naughty = {
		awful = true,
		beautiful = true,
		gears = true,
		lgi = true,
		wibox = true,
	},
	menubar = {
		awful = true,
		beautiful = true,
		gears = true,
		lgi = true,
		wibox = true,
	}
}

local graph = assert(depgraph.make_graph({"lib"}, {}, "lib", nil, nil))

for _, module in ipairs(graph.modules) do
	local base_name = string.match(module.name, "%a*")
	local allowed = allowed_deps[base_name] or {}
	for _, dep in ipairs(module.deps) do
		local dep_base_name = string.match(dep.name, "%a*")
		if base_name ~= dep_base_name and not allowed[dep_base_name] then
[...]

PR #1433, opened 2017-01-23
avoid cyclic dependency between awful and menubar

actionless:
related to #1400

### DIFF
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -15,7 +15,6 @@ local awful = require("awful")
 local wibox = require("wibox")
 local beautiful = require("beautiful")
 local dpi = beautiful.xresources.apply_dpi
-local compute_textbox_width = require("menubar").utils.compute_textbox_width
 
 
 -- Stripped copy of this module https://github.com/copycat-killer/lain/blob/master/util/markup.lua:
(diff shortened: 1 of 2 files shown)


lib/awful/hotkeys_popup/widget.lua
```diff
@@ -290,8 +290,8 @@ function widget.new()
                     joined_labels = joined_labels .. rendered_hotkey .. (i~=#_keys and "\n" or "")
                     end
                 current_column.layout:add(wibox.widget.textbox(joined_labels))
-                local max_width = compute_textbox_width(wibox.widget.textbox(max_label_content), s) +
-                                                    widget_instance.group_margin
+                local max_width, _ = wibox.widget.textbox(max_label_content):get_preferred_size(s)
+                max_width = max_width + widget_instance.group_margin
                 if not current_column.max_width or max_width > current_column.max_width then
```
psychon: This part of the change seems unnecessary? The file can just be left as-is. (Also, I think luacheck will now complain that `compute_textbox_width` is set, but unused)

psychon:
Ok, I'm confused. So this says "avoid cyclic dependency between awful and menubar". However, the change does not actually remove any kind of dependency, does it?

lib/awful/hotkeys_popup/widget.lua
actionless: my  bad, forgot to remove the require line

blueyed:
Thanks!

actionless:
@psychon do you mean table.pack() or there is some different way?

psychon:
You mean the single line?
Isn't
```lua
local max_width, _ = wibox.widget.textbox(max_label_content):get_preferred_size(s)
max_width = max_width + widget_instance.group_margin
```
equivalent to
```
local max_width = wibox.widget.textbox(max_label_content):get_preferred_size(s) + widget_instance.group_margin
```
The only difference should be that the variable with name `_` isn't set.

Put differently: Doesn't Lua ignore all following values from multiple return values in binary operations?

actionless:
thanks, i didn't actually knew about that:
```lua
> function f() return 1,2,3 end ; a,b=f() ; print(a,b,c)
1       2       nil
```

i was expecting it to throw an error in that case

Issue #1411, opened 2017-01-17
Transparent background isn't properly cleared between draws


silverhammermba:
With awesome v3.5.9, fully transparent background colors aren't cleared between draws, meaning that each draw call seems to just stack on top of the last, until everything is a blurry mess. For example, here is what my run dialog looks like:

![transparency bug](http://i.imgur.com/dI32P5A.png)

To reproduce:
1. Set `theme.bg_normal = "#00000000"` i.e. full transparency.
2. Do some stuff that changes the wiboxes e.g. change layouts, open/close windows

If I set the background color to an opaque color everything looks fine. I believe transparency worked as expected in v3.5.8.

Elv13:
You need a background *or* a compositor for this to work. Set your background to black and it will work fine.

silverhammermba:
Ugh, yeah. I just realizing my compositor wasn't running. Disregard.

silverhammermba:
Okay, turns out the compositor wasn't running because of some change in how the background is set. This is how i use to start it:

1. startx runs `compton -b`
2. then `exec awesome`
3. awesome runs `gears.wallpaper.maximized(beautiful.wallpaper, 1, true)`

This was setting the wallpaper, but the compositor was failing with the error `error 143: BadPicture request 139 minor 8`.

If I instead set the wallpaper with feh then the compositor no longer gets an error.

psychon:
> You need a background or a compositor for this to work. Set your background to black and it will work fine.

Well, no? Isn't awesome supposed to do fake transparency in this case?

Elv13:
> Well, no? Isn't awesome supposed to do fake transparency in this case?

It is reproducible, it as always been that way and still is as of git-HEAD. If you have no compositor **and** no root/wallpaper, then semi transparent content is plain garbage. The workaround is simply not to do that. I never saw that as a "bug", but more as a "just don't do this" kind of limitation.

Start a new Xephyr, do something that messes the buffers (but without setting a wallpaper), the start awesome with an `rc.lua` showing a single semi-transparent wibox => garbage.

silverhammermba:
Yeah, that part isn't weird. The odd behavior is that the wallpaper set by awesome causes the compositor to crash.

psychon:
I would still consider this a bug if awesome "draws things wrong" if there is no background.

@silverhammermba Can you run compton with -S and try to get a backtrace of the problem? (With gdb)
[...]

PR #1437, opened 2017-01-23
wibox.drawable: Draw black if no wallpaper is available

psychon:
I-hope-this-fixes: https://github.com/awesomeWM/awesome/issues/1411


psychon:
I now actually tested this with `theme.bg_normal = "#00000000"` and `theme.wallpaper = nil`. I can reproduce the original problem and this patch does indeed fix the problem. Merge?

blueyed:
@psychon
Yes, please adjust the commit message then.

psychon:
Updated the commit message (Removed the "I-hope-this-"-part).

blueyed:
Thanks!

Issue #1445, opened 2017-01-23
naughty notifications close on mouse press rather than mouse release


Veratil:
With notifications closing on a mouse press rather than the mouse release, this can cause unwanted release events to happen on an underlying client.

For instance, if a notification is over a titlebar clicking near the top right will close the notification (the mouse button is still pressed). When the button is released this will cause the close button of a client to fire off, closing the client. Not what is wanted to happen.

Granted this is a rare occurrence given a notifications lifetime, this has happened to me before.

blueyed:
i think it might be here: https://github.com/awesomeWM/awesome/blob/2627f087c5ab4ce3dcb8f9fa7c128cf2b31f85ff/lib/naughty/core.lua#L668-L672.

```diff
diff --git i/lib/naughty/core.lua w/lib/naughty/core.lua
index 764afe20..b3a56a8d 100644
--- i/lib/naughty/core.lua
+++ w/lib/naughty/core.lua
@@ -666,8 +666,8 @@ function naughty.notify(args)
     notification.box:set_widget(completelayout)
 
     -- Setup the mouse events
-    layout:buttons(util.table.join(button({ }, 1, run),
-                                   button({ }, 3, function()
+    layout:buttons(util.table.join(button({ }, 1, nil, run),
+                                   button({ }, 3, nil, function()
                                         die(naughty.notificationClosedReason.dismissedByUser)
                                     end)))
```

Veratil:
That works!

PR #1446, opened 2017-01-23
naughty: mouse: only react on release events

blueyed:
Fixes https://github.com/awesomeWM/awesome/issues/1445.

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -666,8 +666,8 @@ function naughty.notify(args)
     notification.box:set_widget(completelayout)
 
     -- Setup the mouse events
-    layout:buttons(util.table.join(button({ }, 1, run),
-                                   button({ }, 3, function()
+    layout:buttons(util.table.join(button({ }, 1, nil, run),
+                                   button({ }, 3, nil, function()
                                         die(naughty.notificationClosedReason.dismissedByUser)
                                     end)))
 
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -666,8 +666,8 @@ function naughty.notify(args)
     notification.box:set_widget(completelayout)
 
     -- Setup the mouse events
-    layout:buttons(util.table.join(button({ }, 1, nil, run),
-                                   button({ }, 3, nil, function()
+    layout:buttons(util.table.join(button({}, 1, nil, run),
+                                   button({}, 3, nil, function()
                                         die(naughty.notificationClosedReason.dismissedByUser)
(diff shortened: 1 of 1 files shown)


blueyed:
Tested by @Veratil.

Elv13:
Wont that just move the problem from "things that act on mouse release events" to "things that act on mouse press events"? I think both press and release events need to be set. Also, #1447 might happen/

Veratil:
> Wont that just move the problem from "things that act on mouse release events" to "things that act on mouse press events"? I think both press and release events need to be set. Also, #1447 might happen/

This actually causes a notification to act like it should, destroying itself when the button is released rather than pressed and causing whatever is below the mouse after the notification is destroyed getting a mouse release event.

I've tested every notification I have in various places so far. This makes things act as expected without passing events anywhere else. It doesn't interfere with dbus notifications with internal commands (tested with Pithos popups for skip/pause/play). The notifications are working exactly as expected with this patch.

As for #1447, I haven't had propagated events from notifications yet. I'm not discounting the possibility, just noting that that has not happened to me throughout all the testing with notifications I've done.

psychon:
> Wont that just move the problem from "things that act on mouse release events" to "things that act on mouse press events"? I think both press and release events need to be set. 

That's not how the C code works here. It does not care if a callback is set. You could even set both to `nil` and the C code would still do the same things, X11-wise.

What happens is:

- The button is pressed over a notification
- The X11 server sends an event to awesome about the press
- Awesome reacts by removing the notification window
- The mouse cursor is now over whatever was below the notification
- The button is released
- The release event is sent to whatever window is below the mouse cursor

You could get the same behaviour by having a wibox destroyed via a timer. If you press the button before the wibox is destroyed and release it afterwards, the release event gets sent to the window below the wibox. What I am trying to say is: This has nothing to do with "is there a press/release callback".

psychon:
Merged with dd339fd62a835e1f5899ff73a1d8837b9cbf1c27, but I seem to have mixed something up while rebasing.

blueyed:
@psychon 
> but I seem to have mixed something up while rebasing.

The commit looks fine to me?!

psychon:
>> but I seem to have mixed something up while rebasing.

>The commit looks fine to me?!

...but I didn't make it point back here like you always do.

blueyed:
@psychon 
Oh.. then you did not use GitHub's squash UI probably.

psychon:
yup

blueyed:
I find this very useful - but it will use your GitHub email then..

Issue #1449, opened 2017-01-24
awful.placement not updating workarea


bryanalves:
**Output of `awesome --version`:**

awesome v4.0-139-g508614df (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1

**How to reproduce the issue:**

with the following rules
```
awful.rules.rules = {
  { rule = { },
    properties = {
      border_width = beautiful.border_width,
      border_color = beautiful.border_normal,
      callback = awful.client.setslave,
      focus = awful.client.focus.filter,
      keys = clientkeys,
      buttons = clientbuttons,
      screen = awful.screen.preferred,
      placement = awful.placement.no_overlap+awful.placement.no_offscreen
    }
  },

  { rule = { class = "Pidgin", role = "buddy_list" },
    properties = {
      floating = true,
      screen = 1,
      tag = "3",
      width = 250,
      callback = function (c)
          awful.placement.right(c, { attach = true, update_workarea = true })
      end
    }
  }
}
```

Launch pidgin, and then launch a terminal window.

**Actual result:**

Pidgin launches, and is aligned to the right of the screen as it should be.  The terminal also loads, but is full-screen and overlaps with the pidgin window.

**Expected result:**

The 2 windows should appear side by side and not overlap.  Subsequent new windows should stack and never overlap with pidgin.  The 250 pixel area to the right should be reserved only for pidgin -- other windows shouldn't stack or overlap with it.

A couple of other maybe interesting configurations:

**Manual struts**
```
  { rule = { class = "Pidgin", role = "buddy_list" },
    properties = {
      floating = true,
      screen = 1,
      tag = "3",
      width = 250,
      callback = function (c)
          c:struts({right = 250})
      end
    }
  }
```

Here, the strut appears 250 pixels wide on the right, and windows don't enter that.  The pidgin window initially launches on the left of the screen and can be manually dragged into place

**Manual struts with placement parameter**
```
  { rule = { class = "Pidgin", role = "buddy_list" },
    properties = {
      floating = true,
      screen = 1,
      tag = "3",
      width = 250,
      placement = awful.placement.right,
      callback = function (c)
          c:struts({right = 250})
      end
    }
  }
```

[...]

PR #1450, opened 2017-01-24
[RFC] Fix dockable rules

Elv13:
I am not very happy with the PR. The problem is "deeper" than that and the fix is a hack. Awesome keep a "workarea" and uses it for "no_offscreen". Since 4.0, the code handle "all corner cases" and thus detect the docked client as being offscreen in the rules because of the already added struts (from the X properties).

This commit flush the struts early in the rules to mitigate the issue, but it sucks. @psychon could the clients use a `request::struts` and have a `c.preferred_struts` property instead of using the X struts directly?

### DIFF
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -466,7 +466,7 @@ end
 -- Update the workarea
 wibox_update_strut = function(d, position, args)
     -- If the drawable isn't visible, remove the struts
-    if not d.visible then
+    if d.visible == false then
         d:struts { left = 0, right = 0, bottom = 0, top = 0 }
         return
     end
@@ -480,7 +480,6 @@ wibox_update_strut = function(d, position, args)
     local struts = { left = 0, right = 0, bottom = 0, top = 0 }
 
     local m = get_decoration(args)
-
     if vertical then
         for _, v in ipairs {"right", "left"} do
             if (not position) or position:match(v) then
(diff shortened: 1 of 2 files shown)


lib/awful/placement.lua
```diff
@@ -499,6 +498,37 @@ wibox_update_strut = function(d, position, args)
...
+            wibox_update_strut(c, edge, {})
+        else
+            c:struts { left = 0, right = 0, bottom = 0, top = 0 }
+        end
+    end
+end
+
+-- Listen to dockable clients geometry change and update the workarea.
+capi.client.connect_signal("property::geometry", detect_and_update_struts)
+capi.client.connect_signal("property::dockable", detect_and_update_struts)
```
psychon: So, any client which is moved to the edge of the screen (for example by a layout) gets docked there? That doesn't seem right.

Also, I never really understood the magic here, but I thought the idea was to first compute the new geometry and only apply it in the end. So, shouldn't the struts only be applied in the end, too? That way they would not interfere with `no_offscreen`, because no struts are applied yet.
I don't really understand why #1449 happens then, but from your rules it is wrong, since it uses a callback to call into `awful.placement` even though it should use the `placement` key. No?

psychon:
> @psychon could the clients use a request::struts and have a c.preferred_struts property instead of using the X struts directly?

I don't understand the question. So, `c:struts{foo}` shouldn't do anything except emitting a `request::struts` signal. What's `c.preferred_struts` then?

psychon:
Closing for lack of activity in 2.5 years

Issue #1406, opened 2017-01-16
Eclipse detached windows cannot be moved to different screen


crater2150:
Eclipse can detach panels as separate windows (by dragging them out of the main window). In Awesome 3.5 I could launch Eclipse on the main screen and move some panels to the second screen.

Since the upgrade to awesome 4.0, the panels can no longer be moved to a screen different from the main window. Trying to move a panel to the other screen by keyboard does nothing, making it floating and dragging causes it to jump back as soon as the last pixel leaves the screen with the main window. If the main window was on another tag, the subwindow is even placed on that tag, instead of the one it was before.
What **does** work is moving the main window to a different screen, this does not move the detached windows. This workaround is limited, as I can still only place the subwindows on one screen.

[xprop output for main window](https://github.com/awesomeWM/awesome/files/708029/xprop_main_window.txt)
[xprop output for detached panel window](https://github.com/awesomeWM/awesome/files/708030/xprop_detached_window.txt)

Output of `awesome --version`:

awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.2.4 (running with Lua 5.2)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1

**How to reproduce the issue:**

1. Start eclipse
2. Detach a panel by dragging
3. try to move panel to a different screen

**Actual result:**

detached panel jumps back to screen of main eclipse window

**Expected result:**

detached panel should be shown on different screen

Elv13:
Can you create a new user account on your system (to avoid messing up your "real" one) and do a `git bisect` of this? You will have to use the default config for each steps as it changed many, many time in the 2300 commits delta. Use a `LUA_PATH` environment variable (the newer revisions have `--search`, but 3.5 doesn't) it wont take much time.

You can't use Xephyr for this (even with xinerama enabled, it wont reproduce the bug correctly), so use `X :1` to start a second X server (ctrl+alt+f8 to access it)

crater2150:
A lot of revisions fail to link for me with 

    /usr/sbin/ld: CMakeFiles/awesome.dir/awesome.c.o: undefined reference to symbol 'XSetEventQueueOwner'
    /lib64/libX11-xcb.so.1: error adding symbols: DSO missing from command line
    collect2: error: ld returned 1 exit status

[...]

PR #1452, opened 2017-01-24
ewhm: fix some transient windows not movable to other screens (#1406)

crater2150:
After the last comment in the related bug (#1406) I was not sure, if I should create a PR for this. This change fixes the bug for me and I have not yet noticed any side effects, but I use very few programs with windows on several screens.


psychon:
Too much hints magic for me, @Elv13 ?

Elv13:
Probably Ok, I will test tomorrow. I should really write a couple of flowchart for this. Since the code was unified into the request system, it became harder to make_sense_of, but much more flexible and coherent.

blueyed:
Thanks!

Issue #758, opened 2016-03-14
Maximized windows


rshadow:
In version 3.5 are broken maximized window:
1. They are in what is a separate layer. And not the tile with the other windows.
2. They do not adapt to the screen size. If you create a maximized window on a screen and move to another smaller, the window simply override everything and be cropped to fit the screen.

This behavior was not in version 3.4. Maximize the window behave like the rest of the window. These properties are simply ignored.

psychon:
What do you mean by (1)? How could a maximized window tile normally with the other windows? What would "maximized" mean if it did that? Why do you want it to be ignored? I'm confused.

For (2): I thought this was already fixed. And I'm surprised that 3.4 did not have that problem. Weird.

rshadow:
1. I say about windows with X state:
```
$ xprop | grep -i max
_NET_WM_STATE(ATOM) = _NET_WM_STATE_MAXIMIZED_VERT, _NET_WM_STATE_MAXIMIZED_HORZ
```

You can set current window maximized like:
```
wmctrl -r :ACTIVE: -b toggle,maximized_vert,maximized_horz
```

rshadow:
2. This is only for maximized windows. I hope if (1) it will be fixed, and (2) too.

rshadow:
I use awesome 3.5.6

blueyed:
Can you try awesome master, please?

(2) should be fixed there, and (1) maybe, too?!

rshadow:
I try master. Bug exists.

rshadow:
I try 3.5.9 from debian. Bug exists.

Elv13:
The code has been rewritten, most of these issues are fixed, for what remains, better open new issues.

PR #1453, opened 2017-01-24
Fix #1321 and #1459 and #171 and #1320 and #1456

### LINKED ISSUE #1320: Moved to other screen maximized window un-maximizes on first screen

     • D-Bus support: ✔
     • execinfo support: ✔
     • RandR 1.5 support: ✔
     • LGI version: 0.9.1

1. Have two screens. Layout doesn't matter.
2. `S-Ret S-m S-o` on first screen to:
    * create terminal window (actually any window will do)
    * maximize it, and
    * send to other screen.
3. `S-m` on other screen when focused on terminal window from previous step.

**Actual result:**
Terminal window un-maximizes *on the first screen*.
Focus remains on the second screen.

**Expected result:**
I don't really know, but I personally expect window to un-maximize on the screen it currently belongs to.

### LINKED ISSUE #1321: Float mode un-maximization doesn't restore window width
```
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1
```
1. Have tag in floating layout.
2. `S-Ret S-m S-m` to create terminal (e.g., any window will do), maximize and un-maximize it.

**Actual result:**
Window's height is restored to what it was before maximization, but width is not.

**Expected result:**
Width is also returned to what it was before.

Elv13:
Before: to everything twice including the request::geometry
Now: do it once

Edit: also create a real `maximized` property, spliting the state across 2 properties is a source of corner cases.

@taptap @cagelight can you test?

Fix #1320 #1321 #171 #1459

### DIFF
--- a/ewmh.c
+++ b/ewmh.c
@@ -49,9 +49,9 @@ ewmh_client_update_hints(lua_State *L)
         state[i++] = _NET_WM_STATE_MODAL;
     if(c->fullscreen)
         state[i++] = _NET_WM_STATE_FULLSCREEN;
-    if(c->maximized_vertical)
+    if(c->maximized_vertical || c->maximized)
         state[i++] = _NET_WM_STATE_MAXIMIZED_VERT;
-    if(c->maximized_horizontal)
+    if(c->maximized_horizontal || c->maximized)
         state[i++] = _NET_WM_STATE_MAXIMIZED_HORZ;
     if(c->sticky)
         state[i++] = _NET_WM_STATE_STICKY;
@@ -220,6 +220,7 @@ ewmh_init_lua(void)
     luaA_class_connect_signal(L, &client_class, "property::fullscreen" , ewmh_client_update_hints);
     luaA_class_connect_signal(L, &client_class, "property::maximized_horizontal" , ewmh_client_update_hints);
     luaA_class_connect_signal(L, &client_class, "property::maximized_vertical" , ewmh_client_update_hints);
+    luaA_class_connect_signal(L, &client_class, "property::maximized" , ewmh_client_update_hints);
     luaA_class_connect_signal(L, &client_class, "property::sticky" , ewmh_client_update_hints);
     luaA_class_connect_signal(L, &client_class, "property::skip_taskbar" , ewmh_client_update_hints);
     luaA_class_connect_signal(L, &client_class, "property::above" , ewmh_client_update_hints);
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -37,26 +37,24 @@ local ewmh = {
 
 --- Update a client's settings when its geometry changes, skipping signals
 -- resulting from calls within.
-local geometry_change_lock = false
-local function geometry_change(window)
-    if geometry_change_lock then return end
-    geometry_change_lock = true
+local repair_geometry_lock = false
+local function repair_geometry(window)
+    if repair_geometry_lock then return end
+    repair_geometry_lock = true
 
-    -- Fix up the geometry in case this window needs to cover the whole screen.
-    local bw = window.border_width or 0
-    local g = window.screen.workarea
-    if window.maximized_vertical or window.maximized then
-        window:geometry { height = g.height - 2*bw, y = g.y }
-    end
-    if window.maximized_horizontal or window.maximized then
-        window:geometry { width = g.width - 2*bw, x = g.x }
-    end
-    if window.fullscreen then
-        window.border_width = 0
-        window:geometry(window.screen.geometry)
+    -- Re-apply the geometry locking properties to what they should be.
+    for _, prop in ipairs {
+        "fullscreen", "maximized", "maximized_vertical", "maximized_horizontal"
+    } do
+        if window[prop] then
+            window:emit_signal("request::geometry", prop, {
+                store_geometry = false
+            })
+            break
+        end
     end
 
-    geometry_change_lock = false
+    repair_geometry_lock = false
 end
 
 --- Activate a window.
@@ -265,7 +263,10 @@ function ewmh.geometry(c, context, hints)
         -- If the property is boolean and it correspond to the undo operation,
         -- restore the stored geometry.
         if state == false then
+            local original = repair_geometry_lock
+            repair_geometry_lock = true
             aplace.restore(c,{context=context})
+            repair_geometry_lock = original
             return
         end
 
@@ -283,11 +284,11 @@ client.connect_signal("request::activate", ewmh.activate)
 client.connect_signal("request::tag", ewmh.tag)
 client.connect_signal("request::urgent", ewmh.urgent)
 client.connect_signal("request::geometry", ewmh.geometry)
-client.connect_signal("property::border_width", geometry_change)
-client.connect_signal("property::geometry", geometry_change)
+client.connect_signal("property::border_width", repair_geometry)
+client.connect_signal("property::screen", repair_geometry)
 screen.connect_signal("property::workarea", function(s)
     for _, c in pairs(client.get(s)) do
-        geometry_change(c)
+        repair_geometry(c)
     end
 end)
 
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -13,11 +13,16 @@ local util = require("awful.util")
 local aclient = require("awful.client")
 local aplace = require("awful.placement")
 local asuit = require("awful.layout.suit")
+local beautiful = require("beautiful")
 
 local ewmh = {
     generic_activate_filters    = {},
     contextual_activate_filters = {},
 }
+
+--- Honor the screen padding when maximizing.
+-- @beautiful beautiful.maximized_honor_padding
+-- @tparam[opt=true] boolean maximized_honor_padding
 
 --- The list of all registered generic request::activate (focus stealing)
 -- filters. If a filter is added to only one context, it will be in
@@ -274,6 +279,10 @@ function ewmh.geometry(c, context, hints)
 
         if props.honor_workarea == nil then
             props.honor_workarea = honor_default
+        end
+
+        if props.honor_padding == nil and props.honor_workarea and context:match("maximize") then
+            props.honor_padding = beautiful.maximized_honor_padding ~= false
         end
 
         aplace[context](c, props)
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -224,6 +224,7 @@ end
 local context_mapper = {
     maximized_vertical   = "maximize_vertically",
     maximized_horizontal = "maximize_horizontally",
+    maximized            = "maximize",
     fullscreen           = "maximize"
 }
 
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -24,6 +24,10 @@ local ewmh = {
 -- @beautiful beautiful.maximized_honor_padding
 -- @tparam[opt=true] boolean maximized_honor_padding
 
+--- Hide the border on fullscreen clients.
+-- @beautiful beautiful.fullscreen_hide_border
+-- @tparam[opt=true] boolean fullscreen_hide_border
+
 --- The list of all registered generic request::activate (focus stealing)
 -- filters. If a filter is added to only one context, it will be in
 -- `ewmh.contextual_activate_filters`["context_name"].
@@ -285,7 +289,19 @@ function ewmh.geometry(c, context, hints)
             props.honor_padding = beautiful.maximized_honor_padding ~= false
         end
 
+        if original_context == "fullscreen" and beautiful.fullscreen_hide_border ~= false then
+            props.ignore_border_width = true
+        end
+
         aplace[context](c, props)
+
+        -- Remove the border to get a "real" fullscreen.
+        if original_context == "fullscreen" and beautiful.fullscreen_hide_border ~= false then
+            local original = repair_geometry_lock
+            repair_geometry_lock = true
+            c.border_width = 0
+            repair_geometry_lock = original
+        end
     end
 end
 
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -45,10 +45,10 @@ local function geometry_change(window)
     -- Fix up the geometry in case this window needs to cover the whole screen.
     local bw = window.border_width or 0
     local g = window.screen.workarea
-    if window.maximized_vertical then
+    if window.maximized_vertical or window.maximized then
         window:geometry { height = g.height - 2*bw, y = g.y }
     end
-    if window.maximized_horizontal then
+    if window.maximized_horizontal or window.maximized then
         window:geometry { width = g.width - 2*bw, x = g.x }
     end
     if window.fullscreen then
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -291,6 +291,7 @@ local function store_geometry(d, reqtype)
     if not data[d][reqtype] then data[d][reqtype] = {} end
     data[d][reqtype] = d:geometry()
     data[d][reqtype].screen = d.screen
+    data[d][reqtype].sgeo = d.screen and d.screen.geometry or nil
     data[d][reqtype].border_width = d.border_width
 end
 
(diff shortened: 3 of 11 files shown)


objects/client.h
```diff
@@ -47,6 +47,12 @@ typedef enum {
     CLIENT_TITLEBAR_COUNT = 4
 } client_titlebar_t;
 
+typedef enum {
+    CLIENT_MAXIMIZED_NONE = 0 << 0,
+    CLIENT_MAXIMIZED_V    = 1 << 0,
+    CLIENT_MAXIMIZED_H    = 1 << 1,
+} client_maximized_t;
```
psychon: Not needed in the header. Can't this be right in the C file?

Elv13: Done

Elv13:
Ah, as I think of it, its a behavior change for the 0.1 users who set both manually. I should change this to be additive instead of overwriting.

objects/client.c
```diff
@@ -1919,38 +1920,68 @@ client_get_maximized(client_t *c)
...
+void
+client_set_maximized_common(lua_State *L, int cidx, bool s, const char* type, const int val)
+{
+    client_t *c = luaA_checkudata(L, cidx, &client_class);
+
+    /* Store the current and next state on 2 bit */
+    const client_maximized_t current = (
+        (c->maximized_vertical   ? CLIENT_MAXIMIZED_V : CLIENT_MAXIMIZED_NONE)|
+        (c->maximized_horizontal ? CLIENT_MAXIMIZED_H : CLIENT_MAXIMIZED_NONE)
+    );
```
psychon: Well, ok. One thing: Why two bits instead of just using two different variables for this? GCC can likely figure out nice and fast code anyway (not that it matters much here...) and I think the result would be more readable.

taptap:
It still restores on the monitor where it was maximized.
However, it's not widened to whole width anymore.

objects/client.c
Elv13: The property macro like booleans

Bew78LesellB:
It works well for me, this fix #1459, thanks !

objects/client.c
```diff
@@ -1919,38 +1926,68 @@ client_get_maximized(client_t *c)
...
+    const client_maximized_t next = s ? val : CLIENT_MAXIMIZED_NONE;
```
psychon: Likely makes no difference, but this could be moved into the caller. Instead of having a boolean and a mask, this function could directly get the `next` value.
</bikeshedding>

tests/test-maximize.lua
```diff
@@ -99,6 +100,47 @@ local steps = {
...
+        c.maximized = false
+
+        return true
+    end,
+    function()
+        local c = client.get()[1]
+
+        if c.maximized then
+            return
+        end
```
psychon: Is this `if` necessary? Setting `c.maximized` should be instantaneous (so this could even be done in the previous step)

ewmh.c
```diff
@@ -216,6 +216,7 @@ ewmh_init(void)
     luaA_class_connect_signal(L, &client_class, "property::fullscreen" , ewmh_client_update_hints);
     luaA_class_connect_signal(L, &client_class, "property::maximized_horizontal" , ewmh_client_update_hints);
     luaA_class_connect_signal(L, &client_class, "property::maximized_vertical" , ewmh_client_update_hints);
+    luaA_class_connect_signal(L, &client_class, "property::maximized" , ewmh_client_update_hints);
```
psychon: So... a client can be maximized vertically, maximized horizontally and at the same time not be maximized? Put differently, these are now three completely independent booleans? And all code that wants to check for `maximized_vertical` should now check `maximized_vertical or maximized`, because the C code no longer does so for you? That seems rather odd and I don't like it at all.

How about deprecating the horizontal/vertical maximization instead and only have a single "maximized" state instead?

lib/awful/placement.lua
```diff
@@ -1422,6 +1422,11 @@ function placement.restore(d, args)
 
     if not memento then return false end
 
+    -- Apparently, 0x0 did happen to at least 1 person. Officially can't
+    -- happen, but can't be too careful.
+    if memento.width  and memento.width < 1 then return end
+    if memento.height and memento.height < 1 then return end
```
psychon: How about moving this into the creation of a memento instead, so that such an invalid memento cannot be created at all?

However, I am curious how this could happen. This seems to require `c:geometry()` to return 0 for width/height. This can only happen while the client is still being managed, because there is a time frame while the value is still 0 before the values coming from the X11 server are applied. This can only happen during the following signals: `new`, `property::x`, `property::y`, `property::width`, `property::screen`, `property::window`.

An (in my opinion) even better proposal: Instead of working around this in Lua, change the C code so that it emits the signals later (except for `new`, but no one should use that one anyway)?
Basically:
```diff
diff --git a/objects/client.c b/objects/client.c
index ec4ab424b..9f6bd138b 100644
--- a/objects/client.c
+++ b/objects/client.c
@@ -1405,8 +1405,6 @@ client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_at
      * (Else, reparent could cause an UnmapNotify) */
     xcb_change_window_attributes(globalconf.connection, w, XCB_CW_EVENT_MASK, select_input_val);
 
-    luaA_object_emit_signal(L, -1, "property::window", 0);
-
     /* The frame window gets the border, not the real client window */
     xcb_configure_window(globalconf.connection, w,
                          XCB_CONFIG_WINDOW_BORDER_WIDTH,
@@ -1429,15 +1427,16 @@ client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_at
 
     /* Store initial geometry and emits signals so we inform that geometry have
      * been set. */
-#define HANDLE_GEOM(attr) \
-    c->geometry.attr = wgeom->attr; \
-    luaA_object_emit_signal(L, -1, "property::" #attr, 0);
-HANDLE_GEOM(x)
-HANDLE_GEOM(y)
-HANDLE_GEOM(width)
-HANDLE_GEOM(height)
-#undef HANDLE_GEOM
-
+    c->geometry.x = wgeom->x;
+    c->geometry.y = wgeom->y;
+    c->geometry.width = wgeom->width;
+    c->geometry.height = wgeom->height;
+
+    luaA_object_emit_signal(L, -1, "property::x", 0);
+    luaA_object_emit_signal(L, -1, "property::y", 0);
+    luaA_object_emit_signal(L, -1, "property::width", 0);
+    luaA_object_emit_signal(L, -1, "property::height", 0);
+    luaA_object_emit_signal(L, -1, "property::window", 0);
     luaA_object_emit_signal(L, -1, "property::geometry", 0);
 
     /* Set border width */
```

lib/awful/ewmh.lua
```diff
@@ -270,6 +275,10 @@ function ewmh.geometry(c, context, hints)
             props.honor_workarea = honor_default
         end
 
+        if props.honor_padding == nil and props.honor_workarea then
+            props.honor_padding = beautiful.maximized_honor_padding ~= false
+        end
```
psychon: The commit message says that this is specific to maximisation, but I fail to see that in the code...?

lib/awful/ewmh.lua
```diff
@@ -284,7 +279,6 @@ client.connect_signal("request::tag", ewmh.tag)
 client.connect_signal("request::urgent", ewmh.urgent)
 client.connect_signal("request::geometry", ewmh.geometry)
 client.connect_signal("property::border_width", geometry_change)
-client.connect_signal("property::geometry", geometry_change)
```
psychon: ...sounds like the function name no longer matches its job (although I am not quite sure what that job is). Could you clean that up?

lib/awful/ewmh.lua
```diff
@@ -42,18 +42,13 @@ local function geometry_change(window)
...
-        window:geometry { width = g.width - 2*bw, x = g.x }
-    end
-    if window.fullscreen then
-        window.border_width = 0
-        window:geometry(window.screen.geometry)
+    for prop in ipairs {
+        "fullscreen", "maximized", "maximized_vertical", "maximized_horizontal"
+    } do
+        if window[prop] then
+            window:emit_signal("request::geometry", prop, {})
```
psychon: Should the last argument somehow indicate that this is just re-applying the way things should be?

tests/test-maximize.lua
```diff
@@ -82,10 +82,11 @@ local steps = {
...
         assert(c.fullscreen)
-        assert(new_geo.x-bw==sgeo.x)
-        assert(new_geo.y-bw==sgeo.y)
-        assert(new_geo.x+new_geo.width+bw==sgeo.x+sgeo.width)
-        assert(new_geo.y+new_geo.height+bw==sgeo.y+sgeo.height)
+
+        assert(new_geo.x==sgeo.x)
+        assert(new_geo.y==sgeo.y)
+        assert(new_geo.x+new_geo.width+2*bw==sgeo.x+sgeo.width)
+        assert(new_geo.y+new_geo.height+2*bw==sgeo.y+sgeo.height)
```
psychon: Do I understand correctly that the code doing ` window:geometry(window.screen.geometry)` was actually wrong (since it ignored the border width) and the wrongness managed to end up in a test? Yuck.

Elv13: That code path wasn't of my making ;) (but the test is). Actually, after thinking about it, it was probably done on purpose. Done this way, you get "real" fullscreen. This was initially why I left it that way.  But now it conflicts with the request API, so it has to use the standard code path or these issues will keep piling up.

ewmh.c
Elv13: The previous version of the patch kept the 2bit==maximized concept, but it doesn't work after all.

*problem 1:*
When you maximized something, all 3 signals were sent. If some code handled "maximized" a "maximized_horizontally" (or all 3), then it was executed 3 times.

*problem 1.1:*
`request::geometry` was sent twice (before this PR) when you did a single operation. This is why it failed to serialize the size.

*problem 2:*
If you had the 3 maximize button in the titlebar, than where you toggled maximize twice, it would override the other 2 and do->undo would not restore the state as it was.

So, having a split state made most operation *more* completed than having 3 independent states. Also, have 3 booleans ensure that when changing one of them, only 1 signal is emitted. The previous code worked, I am not denying that, but it had the "design quirks build-in" while the request API blueyed and me implemented is designed to unify "concepts" into a single handler. It *could* be fixed in the handler, but I would rather have a simplified pipeline. This is what this patch does. Most users use maximize and not the other 2, so simplifying that one while making the other 2 a little more complex doesn't seem such a bad idea.

> How about deprecating the horizontal/vertical maximization instead and only have a single "maximized" state instead?

I don't think its necessary. Once this patch becomes stable, it wont need to be changed. I probably should merge `fullscreen` into this function as well. I will if I ever attack #758 and [insert z-index limitations issue I can't find right now].

tests/test-maximize.lua
psychon: ...so, does that mean that the next issue will be the "fullscreen" is no longer fullscreen? That's not good. Fullscreen should hide the border (=set border width to zero).

Elv13: Done

Elv13:
I addressed many comments and added the fix for #1320. I don't expect this to be approved without another round of changes, but at least this help the pile of open PRs I have right now.

Again, I want to point out that adding another maximized mode was made to simplify the code, even if it makes the C code is little bit uglier.

objects/client.c
```diff
@@ -1962,21 +1953,28 @@ client_set_maximized_common(lua_State *L, int cidx, bool s, const char* type, co
...
+        (c->maximized_vertical   ? CLIENT_MAXIMIZED_V    : CLIENT_MAXIMIZED_NONE)|
+        (c->maximized_horizontal ? CLIENT_MAXIMIZED_H    : CLIENT_MAXIMIZED_NONE)|
+        (c->maximized            ? CLIENT_MAXIMIZED_BOTH : CLIENT_MAXIMIZED_NONE)
     );
-    const client_maximized_t next = s ? val : CLIENT_MAXIMIZED_NONE;
+    client_maximized_t next = s ? (val | current) : (current & (~val));
+
+    /* When both are already set during startup, assume `maximized` is true*/
+    if (next == (CLIENT_MAXIMIZED_H|CLIENT_MAXIMIZED_V) && !globalconf.loop)
+        next = CLIENT_MAXIMIZED_BOTH;
```
psychon: Shouldn't this by `|=` instead of `=`?
Or should we invent our own `MAXIMIZED` state for EWMH and use that one here? If so, feel free to open a todo-issue assigned to me telling me to get rid of this again.

lib/awful/ewmh.lua
```diff
@@ -280,7 +284,18 @@ function ewmh.geometry(c, context, hints)
...
+            props.ignore_border_width = true
+        end
+
         aplace[context](c, props)
+
+        -- Remove the border to get a "real" fullscreen.
+        if original_context == "fullscreen" and beautiful.fullscreen_hide_border ~= false then
+            repair_geometry_lock = true
+            c.border_width = 0
+            repair_geometry_lock = false
```
psychon: shouldn't this be:
```
local original = repair_geometry_lock
repair_geometry_lock = true
c.border_width = 0
repair_geometry_lock = original
```
After all, I think that this function here will be called by `repair_geometry()`.

lib/awful/rules.lua
```diff
@@ -363,8 +363,8 @@ function rules.extra_properties.placement(c, value)
     local ty = type(value)
 
     local args = {
-        honor_workarea = true,
-        honor_padding  = true
+        honor_workarea = props.honor_workarea ~= false,
+        honor_padding  = props.honor_padding ~= false
```
psychon: Undocumented?

lib/awful/placement.lua
```diff
@@ -1422,11 +1422,33 @@ function placement.restore(d, args)
...
-    memento.screen = nil --TODO use it
+    local x, y = memento.x, memento.y
+
+    -- Some people consider that once moved to another screen, then
+    -- the memento needs to be upgraded. For now this is only true for
+    -- maximization until someone complains.
+    if memento.screen and memento.screen.valid and args.context == "maximize"
+      and d.screen and get_screen(memento.screen) ~= get_screen(d.screen) then
+        -- Use the absolute geometry as the memento also does
+        local oldsgeo = memento.screen.geometry
```
psychon: Should the screen's geometry be saved inside the memento? That way this would also work if the original screen has since be removed.
Also, the screen's geometry could have changed, so the `x`,`y` saved in there need to be fixed up anyway (the `get_screen(memento.screen) ~= get_screen(d.screen) ` should be dropped), I think.

psychon:
Tentative :+1: besides the comments above.

objects/client.c
Elv13: > Shouldn't this by |= instead of =?

Nah, this is one of the possible "bad side effect" of this. If someone manually set both `maximized_[*]`, then it's interpreted as maximized on on restart. The opposite, if `=|` was used and the user had no shortcut for `maximized_[*]`, then he/she/it would be stuck with a triple maximized client with no way to turn it off. The "issue" itself already exist, but isn't "as bad" since the default maximized shortcut is currently "destructive" in the sense toggling it twice will reset both maximization. From now on, if one of the `maximize_*` is set and not the other, then it will be harder for the user to turn it off. The `=` just makes sure it doesn't happen because of this change.

At first I was thinking your idea of another maximized xprop was a good idea, but thinking about it again, I think a better course of action might be to reset fullscreen and maximization properties when restarting. It is a behavior change compared to previous version and someone, somewhere, will notice. However it simplifies the algorithm and users don't restart Awesome as often as they used to, so it wont be as noticeable as before. Anyway, when you lose your layouts and stack order, keeping maximization cause a confusing result.

lib/awful/rules.lua
Elv13: Nothing new in the module. I still have my `managing-clients.md.lua` doc pages, but it's incomplete. It started from the same code that ended up being `appearance.md.lua`. It parses the code (client.c and client.lua) to build a property database then add the "custom" ones.

But for now I added a new commit with a crappy header doc. It also adds all the other undocumented properties.

Elv13:
This is ready for another review and/or merge+open_issue for the remaining "problem" (that isn't really a regression)

Elv13:
/me I think I will hear about these changes for some time... Hopefully they get us closer to something that works.

objects/client.c
```diff
@@ -805,9 +805,17 @@
  * @function set_newindex_miss_handler
  */
 
+typedef enum {
+    CLIENT_MAXIMIZED_NONE = 0 << 0,
+    CLIENT_MAXIMIZED_V    = 1 << 0,
+    CLIENT_MAXIMIZED_H    = 1 << 1,
+    CLIENT_MAXIMIZED_BOTH = 1 << 2, /* V|H == BOTH, but ~(V|H) != ~(BOTH)... */
```
blueyed: @Elv13 
The comment is not true (or I do not get it).. CLIENT_MAXIMIZED_BOTH is 4, but V|H is 3 (V being 1 and H being 0)?!
So `V|H == BOTH` is not true.

Elv13: I was referring to the concepts. When you maximize both H and V, then you are *in effect* maximized. But when you undo the operation, you are left in a potentially inconsistent state.

I agree the comment is unhelpful and misleading. I should copy/paste the [discussion comment](https://github.com/awesomeWM/awesome/pull/1453#discussion_r97938954) into the code and remove this comment.

PR #1462, opened 2017-01-25
beautiful.theme_assets: Use screen geometry instead of workarea

psychon:


psychon:
Ping?

Issue #1417, opened 2017-01-20
awesome crash when there is a syntax error in rc.lua


Raphi111:
Output of `awesome --version`:

awesome v4.0-105-gbfb35349 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1

**How to reproduce the issue:**

make a random syntax error in rc.lua and restart awesome.

**Actual result:**

awesome crash.

**Expected result:**

Awesome fallsback to a default rc.lua and prints some errors 
********************

I've been looking for errors messages in Xorg's logs but I haven't been able to find any (yet). My tty is configured to restart X as soon as it crash so I can't see any error there either but I can disable it to try to debug it more if needed.

PS: awesome 4.0 starts so much faster than version 3.5 ! It's amazing.

aajjbb:
Do your distribution is a fresh install on pure awesome, or you had some DE installed before ? I'm experiencing similar issue with Manjaro (KDE by default). 

Instead of falling back on default `rc.lua` and show the errors, the system just freezes. If you're in KDE base too, I think it might have something to do with SDDM.

psychon:
I can't test right now, so @aajjbb what do you mean exactly with "crash"? (@Raphi111 already wrote that he cannot provide more information)

aajjbb:
By crash, I mean the system just freezes. And I need to open other tty and restart. 

@Raphi111 do you have a main DE installed as well ? I do only experience it in my Desktop, which has ManjaroKDE. I do not experience it with ManjaroMATE, so I suspect it may have something to do with KDE or SDDM (it's display manager)

Raphi111:
I've been able to test it on another laptop that doesn't auto-restart X.
By crash I mean that when I run X (startx) from a console, the screen goes black for one second, then back to the console. I think I got the error:
`awesome: main:728: couldn't find any rc file`
I spotted other errors before that line but I've got to check if they are related or not.

psychon:
> awesome: main:728: couldn't find any rc file

This is the error message you get after all configs error out. So the question should be: What error did the default config die with (after your user config already errored out)?

Note that you can run awesome in Xephyr by running something like this in a terminal:

Xephyr :5 & sleep 1 ; DISPLAY=:5 awesome

[...]

Issue #1346, opened 2017-01-01
awesome 4.0 debian installs default rc.lua to /usr/local/etc/xdg/awesome/ not /etc/xdg/awesome/


wright8191:
Output of `awesome --version`:

```
↪ awesome --version
awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✘
 • LGI version: 0.9.0
```

**How to reproduce the issue:**

build and install awesome v4 debian package (was done on ubuntu 16.04.1)

```
make
make package
sudo dpkg -i build/awesome-4.0.0.0-Linux.deb
```

**Actual result:**

default `rc.lua` is installed to `/usr/local/etc/xdg/awesome/rc.lua` not `/etc/xdg/awesome/rc.lua`

```
↪ dpkg -c build/awesome-4.0.0.0-Linux.deb | grep rc.lua
-rw-r--r-- root/root     22027 2016-12-30 21:44 ./usr/local/etc/xdg/awesome/rc.lua
```

so there is no default `rc.lua` to fallback on

**Expected result:**

`rc.lua` is installed to `/etc/xdg/awesome/rc.lua`

legostax:
I would like to add the following:

**Expected result:**

The folders `icons`, `lib`, and `themes` should be installed to `/usr/share/awesome` (at least, according to the docs)

**Actual result:**

The folders `icons`, `lib`, and `themes` are installed to `/usr/local/share/awesome`

This was found on the same version of awesome, (only difference is no D-Bus support) and on the same version of Ubuntu (16.04.1).  I too compiled from source and created a `.deb`.

Elv13:
Note that the workaround is calling CMake yourself instead of using the Makefile wrapper.

```sh
cd path/to/awesome
mkdir build
cmake .. -DCMAKE_INSTALL_PREFIX=/usr
make -j4
make package
```

This is, of course, not a solution to the deeper issue (our "make package" target should check and fix the prefix), but it should generate a correct .deb.

wright8191:
This workaround does not work for me. It installs rc.lua to `/usr/etc/xdg/awesome/rc.lua` which awesome does not fall back to.

```
↪ dpkg -c awesome-4.0.0.0-Linux.deb| grep rc.lua
[...]

PR #1463, opened 2017-01-25
Find fallback config

psychon:
This hardcodes the path to the default config into the binary. This will make awesome try the default config twice for most people, but will make things actually work for people who do not have the default config in `/etc/xdg/awesome`.


Elv13:
Still wont fix the issue if they modified the config or have an older one and didn't do make install but added `--search ./lib` (or $LUA_PATH). I still think putting a copy of `rc.lua` in the binary is the safest bet. +/- 25kB wont be noticed by most.

psychon:
#1346 is titled "awesome 4.0 debian installs default rc.lua to /usr/local/etc/xdg/awesome/ not /etc/xdg/awesome/", so yes, this does fix that issue.

For #1417 I am still waiting for someone to actually show the error that makes the default config fail. I have a feeling that it might be "the user config messes up the Lua state in a way that also makes the default config fail" and so putting the config into the binary still would not help.

psychon:
Putting the default config into the binary is not that easy. It is generated by `docs/05-awesomerc.md.lua` during `make`, so we would need something that afterwards takes the generated file, escapes it properly and puts it into a generated header or source file. All of that while getting the make-dependencies right so that the code is not compiled before the default config is added.
Also: If we want to protect against someone editing the default config, should we also protect against someone editing the libs and put e.g. all of `awful` into the binary?

But this is all off topic. This PR is about: When we install the config to some non-xdg-basedir-compliant directory, we should also look there and load the config from there. It is a bit amazing that we never did so.

psychon:
Any reason against this?

PR #1467, opened 2017-01-25
WIP: Just a quick test

psychon:
Just FYI @actionless (but this is not what I am actually trying out here):

2017-01-25 10:01:54 W: awful: function ? is deprecated. Use 'width, _ = textbox:get_preferred_size(s)' directly..
stack traceback:
	lib/awful/util.lua:45: in function 'deprecate'
	lib/menubar/utils.lua:297: in function <lib/menubar/utils.lua:296>
	(...tail calls...)
	lib/menubar/init.lua:214: in function 'get_current_page'
	lib/menubar/init.lua:345: in function 'menulist_update'
	lib/menubar/init.lua:354: in function 'callback'
	lib/menubar/menu_gen.lua:133: in function <lib/menubar/menu_gen.lua:97>
	[C]: in function 'xpcall'
	(...tail calls...)
	lib/menubar/utils.lua:292: in function <lib/menubar/utils.lua:289>

I just need a Travis build of this. Sorry for the spam.

I want to check if this crashes since (in theory) it should run into a memory corruption bug in LGI.

### DIFF
--- a/tests/test-menubar.lua
+++ b/tests/test-menubar.lua
@@ -1,0 +1,69 @@
+local awful = require("awful")
+
+require("_runner").run_steps({
+    function()
+        require("menubar").show()
+        return true
+    end,
+    function(count)
+        if count == 5 then
+            return true
+        end
+    end,
+    function(count)
+        if count == 5 then
+            return true
+        end
+    end,
+    function(count)
+        if count == 5 then
+            return true
+        end
+    end,
+    function(count)
+        if count == 5 then
+            return true
+        end
+    end,
+    function(count)
+        if count == 5 then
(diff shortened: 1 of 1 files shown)


psychon:
Lua 5.3:
```
awesome: lapi.c:915: lua_callk: Assertion `(L->status == 0) && "cannot do calls on non-normal thread"' failed.
```
Lua 5.2:
```
awesome: luaA_panic:464: unprotected error in call to Lua API ((null))
```
So yes, we are already affected by that LGI bug, but just lucky enough not to see it happen all that often in practice. Yay for memory-corruption bugs!

PR #1471, opened 2017-01-25
Test that beautiful is not used too early

psychon:
For loading a theme, beautiful.init() has to be called. We do not handle
dynamic changes of the theme, so if the theme is changed after something
was already queried, this change will not apply. This also means that
before the theme is loaded, nothing useful can be returned. In fact,
code that tries to access the theme during require() will never get a
useful value from beautiful, but just nils.

To catch this, one Travis build is modified so that an error is raised
if the theme is accessed before beautiful.init() was called.

----

The only reason that this is WIP is because it already catches things that do not work. Ideas what to do about it?

Edit: Here is the backtrace:
```
error: lib/beautiful/init.lua:32: May not access theme before beautiful.init():
stack traceback:
	lib/beautiful/init.lua:32: in function '__index'
	lib/beautiful/init.lua:214: in function '__index'
	lib/awful/hotkeys_popup/widget.lua:58: in function 'new'
	lib/awful/hotkeys_popup/widget.lua:463: in function 'get_default_widget'
	lib/awful/hotkeys_popup/widget.lua:480: in function 'add_hotkeys'
	lib/awful/hotkeys_popup/keys/vim.lua:171: in main chunk
	[C]: in function 'require'
	lib/awful/hotkeys_popup/keys/init.lua:11: in main chunk
	[C]: in function 'require'
	lib/awful/hotkeys_popup/init.lua:12: in main chunk
	[C]: in function 'require'
	/home/psychon/projects/awesome/awesomerc.lua:13: in main chunk
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
   matrix:
     - LUA=5.2 DO_CHECKQA=1 LUANAME=lua5.2 DO_COVERAGE=coveralls
     # luajit: installed from source.
-    - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.4 TEST_PREV_COMMITS=1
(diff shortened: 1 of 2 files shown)


actionless:
could you rebase on top of the PR which got merged already?

psychon:
Rebase ontop of master: Done
Squash the fixup commits: Done

PR #1475, opened 2017-01-26
Input shape support

psychon:
This adds input shape support on drawins and clients to the C code (and touches some Lua docs to add it where needed). The input shape allows (as I understand it, I have to admit that I did not actually test this) to make parts of a window click-through. So if you have a compositing manager running, have parts of a drawin fully transparent and set the input shape correctly, you can actually interact with whatever is below.

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -661,6 +661,16 @@ main(int argc, char **argv)
     /* check for shape extension */
     query = xcb_get_extension_data(globalconf.connection, &xcb_shape_id);
     globalconf.have_shape = query && query->present;
+    if (globalconf.have_shape)
+    {
+        xcb_shape_query_version_reply_t *reply =
+            xcb_shape_query_version_reply(globalconf.connection,
+                    xcb_shape_query_version_unchecked(globalconf.connection),
+                    NULL);
+        globalconf.have_input_shape = reply && (reply->major_version > 1 ||
+                (reply->major_version == 1 && reply->minor_version >= 1));
+        p_delete(&reply);
+    }
 
     event_init();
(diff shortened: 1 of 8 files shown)


Elv13:
Awesome!

One easy change I would see here is a `passthrough` (better name welcome) property. It could be implemented in Lua or C.

I also have a rather silly question. If a wibox/client have an empty input mask (on purpose), will it have `mouse::enter` / `mouse::leave` signals? This is kind of purpose defying, but at the same time, the Unity/Ubuntu notification change opacity when you hover it. I wonder how they know when to restore opacity when the mouse leave. I actually don't want this myself, but I am curious enough to ask.

psychon:
No idea and Debian codesearch so far did not help me finding the right source code. There is always polling as a possibility....

psychon:
Experiments say that you don't get any mouse events:
```lua
local lower_w = wibox{ visible = true, x = 20, y = 20, width = 100, height = 100, bg = "#0f0" }
local upper_w = wibox{ visible = true, x = 45, y = 45, width =  50, height =  50, bg = "#f00" }
for name, w in pairs { lower = lower_w, upper = upper_w } do
    local function connect(signal)
        w:connect_signal(signal, function()
            print(string.format("Signal %s on %s", signal, name))
        end)
    end
    connect("mouse::enter")
    connect("mouse::leave")
    connect("mouse::move")
    connect("button::press")
    connect("button::release")
end
local s = cairo.ImageSurface.create(cairo.Format.A1, 0, 0)
upper_w.shape_input = s._native
s = nil
```

PR #1477, opened 2017-01-26
Fall back to reported DPI if Xft.dpi is not present

Oblomov:


Elv13:
As discussed on IRC, please change this patch so it doesn't mess the DPI if it's below 200 (or 160, or random number in the wild). This http://dpi.lv/ websites show no "official" HiDPI monitors below 200.

(This is a parallel issue and doesn't mean I approve or disprove  the patch itself.)

lib/beautiful/xresources.lua
```diff
@@ -83,8 +83,16 @@ function xresources.get_dpi(s)
         if awesome and awesome.xrdb_get_value then
             xresources.dpi = tonumber(awesome.xrdb_get_value("", "Xft.dpi"))
```
psychon: What does "link in comment" in the commit message refer to? What kind of comment can a commit message have?

lib/beautiful/xresources.lua
```diff
@@ -83,8 +83,16 @@ function xresources.get_dpi(s)
...
+        -- https://keithp.com/~keithp/talks/xtc2001/paper/xft.html#sec-editing
+        -- the proper fallback for Xft.dpi is the vertical DPI reported by
+        -- the X server. This will generally be 96 on Xorg, unless the user
+        -- has configured it differently
         if not xresources.dpi then
-            xresources.dpi = 96
+            local root = require("root")
+            _, h = root.size()
+            _, hmm = root.size_mm()
+            xresources.dpi = round(h*25.4/hmm)
```
psychon: I do not like 25.4-magic-values and you never even tested this.

Oblomov:
Elv13:
> As discussed on IRC, please change this patch so it doesn't mess the DPI if it's below 200 (or 160, or random number in the wild). This http://dpi.lv/ websites show no "official" HiDPI monitors below 200.

[snip]
> On the other hand, I own an HiDPI device for testing my apps and it is truly unusable when DPI is set to 96. So Awesome has to fix itself as well as it can. But only when not doing so renders it unusable.

> (This is a parallel issue and doesn't mean I approve or disprove the patch itself.)

I also feel that this is a matter which is orthogonal to my patch, since funky not-really-HiDPI scaling can also affect current awesome if the user sets `Xft.dpi` resource to a value such a 140. However, if you feel that this should be solved _first_, I could work on that before getting back to the fallback aspect.

> Rationals: It is a behavior change and a very unexpected one for 95% of the userbase.

Well, unless 95% of the userbase actually changes their Xorg configuration to report anything other than 96 (which is what Xorg does unless told otherwise), they won't really be affected by the change, but I do get your point about it being a breaking change.

Oblomov:
psychon:

> If you want to compute some DPI values: How about using the average of the screen's outputs?

As someone who frequently has to work with mixed DPI setups, I would recommend against the solution. The result is that things end up appearing too small in high-DPI monitors, and too large in low-DPI monitors. Picking just the  highest (or lowest, but in my experience highest is better) gives better results.

Of course, the ideal solution is handling different DPI monitors separately, so that things do not appear significantly larger/smaller in any of them. However, this would require extensive engine to the theme engine, as all the properties would have to become a function of screen at least.

root.c
```diff
@@ -448,6 +448,20 @@ luaA_root_size(lua_State *L)
...
+ * @return Width of the root window, in millimeters.
+ * @return height of the root window, in millimeters.
+ * @function size_mm
+ */
+static int
+luaA_root_size_mm(lua_State *L)
+{
+    lua_pushinteger(L, globalconf.screen->width_in_millimeters);
+    lua_pushinteger(L, globalconf.screen->height_in_millimeters);
+    return 2;
```
Oblomov: This can be handled in the same way RANDR itself handles it internally: updating the `*_in_millimeters` too, keeping the ratio constant. I can add a commit that does this if deemed appropriate.

lib/beautiful/xresources.lua
Oblomov: I meant the comment added to the function in the actual commit. I can rephrase the commit message to include the link itself to make it clearer.

lib/beautiful/xresources.lua
Oblomov: I can define the constants `inches_to_millimeter` to explain the 25.4

As for testing, I'm actually running awesome with this code now, did I miss something?

psychon: Uhm, `require("root")` failed quite badly on Travis since no such module was found. It's just a global variable.

Oblomov: That's interesting, how the heck did it work here then? I don't even get an error in the log. Very odd. I'll get rid of it.

psychon: Oh, actually it fails on Travis while running unit tests, which is half-unrelated. Wow, I did not know that the C code makes things like `root` `require`able....

root.c
Oblomov: Scratch that, we actually get the new physical size via RRChangeNotifyEvent, we just need to keep track of the update.

actionless:
Of course, the ideal solution is handling different DPI monitors separately, so that things do not appear significantly larger/smaller in any of them. However, this would require extensive engine to the theme engine, as all the properties would have to become a function of screen at least.

do you mean this https://awesomewm.org/apidoc/libraries/beautiful.xresources.html#set_dpi + file:///usr/share/doc/awesome/doc/libraries/beautiful.xresources.html#apply_dpi?

if you'll feed them screen as argument it should do the thing

however not all the theme arguments can be functions, but i would be thinking in that direction (like it's already implemented for `theme.wallpaper` in order to handle different wallpaper size on different screen)

blueyed:
Just some drive-by comments:
 - `set_dpi` / `apply_dpi` work quite well already for me
 - my "HiDPI" screen is the laptop's internal one, but only with ~170 - still almost twice the default of 96.
 - I've experimented with some `start-with-dpi` wrapper that would be used by awesome to start new programs, and would temporarily change my DPI settings (via `xsettingsd` and `xrdb`), launch the program and set it back.  It uses/used `beautiful.xresources.get_dpi(s)` - and I'm setting it per screen (but mainly hardcoded currenty).

Some snippet from that:  (yes, I do not know why I've used `workarea` - I think the documentation can be improved/clearer on the differences).
```lua
for s in screen do
  local inch = width / 25.4
  local dpi = math.floor(s.workarea.width / inch)
  beautiful.xresources.set_dpi(dpi, s)
end
```

Oblomov:
@actionless : the thing is, themes in general do no use `apply_dpi` with a screen argument. If not everything can be made into a function, some other way would have to be envisioned. An option would be to have themes only report unscaled lengths and sizes (i.e. with lengths at the reference 96 DPI), and then have widgets scale them depending on the screen they get rendered on.

However, I think that this is a separate topic of discussion that has its merits independently from this changeset (whose only aim is to fallback to what the user set up as core DPI, rather than what to do with it). I'll look into opening a separate issue to track the discussion about how to handle this.

@blueyed : I have had cursory discussions with @psychon on IRC about that, and we could even have per-screen DPI autoconfiguration as well, by using RANDR (of course without removing the user's option to override it) and keeping it dynamically update on RRChangeNotifyEvent. The biggest hurdle is choosing what DPI to use when outputs overlap (my idea is to use the highest DPI, since such a setup is generally used to provide a 'zoomed in' view anyway, but this should be decided anyway).

There is another thing that should be handled: again, it's not directly related to this changeset, but if we pick the native DPIs, it's going to be more important, as in most case the actual DPIs are not integer multiples of the reference one. And we should probably look into separating UI scaling from 'proper' DPI scaling. For example, @blueyed monitor has 170 DPI, which means a scaling of 1.77 over the reference 96 DPI. My eDP has 235 DPI, which would mean a scaling of 2.45 over the reference 96 DPI. in my experience, non-integer scaling produces sub-optimal results. The UI scaling should preferably follow integer steps (unless the user wants differently), although we should still have the option to use the 'true' scaling (to draw 'exact' lengths). I'll open a discussion about this too.

root.c
psychon: Oh right, I actually missed that.

actionless:
>  and then have widgets scale them depending on the screen they get rendered on.

after widget system was remade all widgets should have screen in their `context`, however i am not that familiar with widget system's internals to say how complicated it will be to implement

Elv13:
@psychon any comments on the current version of this? I am trying to see if we can clear the backlog a bit (I guess you already notified ;) )

Elv13:
Thanks

psychon:
This broke the build on Travis: https://travis-ci.org/awesomeWM/awesome/builds/236151601
```
    lib/beautiful/xresources.lua:94:17: setting non-standard global variable _
    lib/beautiful/xresources.lua:94:20: setting non-standard global variable h
    lib/beautiful/xresources.lua:95:17: setting non-standard global variable _
    lib/beautiful/xresources.lua:95:20: setting non-standard global variable hmm
    lib/beautiful/xresources.lua:96:20: accessing undefined variable hmm
    lib/beautiful/xresources.lua:97:44: accessing undefined variable h
    lib/beautiful/xresources.lua:97:57: accessing undefined variable hmm
```

lib/beautiful/xresources.lua
```diff
@@ -83,6 +83,22 @@ function xresources.get_dpi(s)
...
         end
+        -- Following Keith Packard's whitepaper on Xft,
+        -- https://keithp.com/~keithp/talks/xtc2001/paper/xft.html#sec-editing
+        -- the proper fallback for Xft.dpi is the vertical DPI reported by
+        -- the X server. This will generally be 96 on Xorg, unless the user
+        -- has configured it differently
+        if not xresources.dpi then
+            if root then
+                local mm_to_inch = 25.4
+                _, h = root.size()
```
actionless: yup, those should be `local`

but i am wondering why those weren't catched by travis before merging

psychon: Travis caught it. Just no one looked at the result: https://travis-ci.org/awesomeWM/awesome/builds/195847851
foo

Elv13:
That one's on me, sorry.

psychon:
I guess I didn't notice it either.

actionless:
i didn't mean to blame anybody, i just thought what the tests were false positive (wasn't looking on the status when asking)

Issue #1482, opened 2017-01-27
Can't spawn program in another screen


martinber:
Output of `awesome --version`:

~~~~~
awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.2.3 (running with Lua 5.2)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✘
 • LGI version: 0.8.0
~~~~~

**How to reproduce the issue:**

I have two monitors. I've created some tags with unique names. I want to start a program in a specific tag, the problem it's that I can't spawn on a tag located on an unfocused tag. I'm not a native English speaker so sorry if I make some mistakes, also I don't know if what I'm trying to do it's possible.

Starting from a default rc.lua, I've commented out this line (around line 183):
~~~~~
awful.tag({ "1", "2", "3", "4", "5", "6", "7", "8", "9" }, s, awful.layout.layouts[1])
~~~~~

And at the end of the file, I've created my own tags and tried to spawn apps on my tags using `awful.spawn()`. AFAIK by reading the docs I can set the tag and screen from there

~~~~~
-- Set names for tags
mytags = {}
tags_names = {}
tags_layouts = {}

layout_tile = awful.layout.layouts[2]

tags_names[1] = {'editor',
                'terminal',
                'files1',
                'view2',
                'etc2',
                'etc3'}

tags_names[2] = {'www',
                'config',
                'view1',
                'files2',
                'etc1'}

for s = 1, screen.count() do
    mytags[s] = awful.tag(tags_names[s], s, layout_tile)
end

-- Returns the screen and tag object from a tag name
-- I can't use awful.tag.find_by_name() because I need to previously know his screen
function get_screen_and_tag(name)
    for s in screen do
        for i, tag in ipairs(s.tags) do
            if name == tag.name then
                return s, tag
            end
        end
    end
end

-- Spawns my program in a tag
function spawn_in_tag(program, tag_name)
    s, t = get_screen_and_tag(tag_name)
    awful.spawn(program, {
        tag = t,
        screen = s,
    })
end

spawn_in_tag("mate-terminal", "view1") -- in screen 2
spawn_in_tag("mate-terminal", "files2") -- in screen 2
spawn_in_tag("mate-terminal", "etc1") -- in screen 2
spawn_in_tag("mate-terminal", "view2") -- in screen 1
spawn_in_tag("mate-terminal", "etc2") -- in screen 1
[...]

PR #1487, opened 2017-01-28
Add an awful.rules chain of responsibility

### LINKED ISSUE #1457: New clients are added to the seleceted tag instead of the one they were spawned from.
```
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1
```

Run a slower program like firefox and then switch tags before the window is created.

**Actual result:**
The program spawns on the currently selected tag.

**Expected result:**
The program spawns on the tag it was spawned from.

my "fix" that is probably unreliable:
```
function ()
                  awful.prompt.run {
                      prompt       = "Open: ",
                      textbox      = awful.screen.focused().mypromptbox.widget,
                      exe_callback = function(...)
                        awful.screen.focused().mypromptbox:spawn_and_handle_error(..., {tag=mouse.screen.selected_tag})
                      end,
                      history_path = awful.util.get_cache_dir() .. "/history"
                  }
              end
```

I also noticed that `awful.widget.prompt()` has some args like `completion_callback` from `awful.prompt` but not others like `exe_callback`, it's unclear which are deprecated...

Elv13:
Fix #1457 #1482 #1937 #2159

Description:
Allow workflow extensions to affect the rule table before it is applied

Use cases:

 * normal awful.rules
 * restore client state/position when started
 * dynamic tags
 * spawn with arguments

As for the questions there:

> Is the order important

Totally crucial and unavoidable. Rules have a dependency graph. Just reversing the order of the 2 default source would break #1482 again if there was a `tag` rule in `rc.lua` (something most users will totally do). Tyannical cannot work if it's not between awful.rules (it depends on it) and awful.spawn source (that always must override everything else). @mindeunix clientDB can only be just before `awful.spawn`, nowhere else or it wont work.

> Can't you just re-order them yourself? let's face it. you will be the only user of this

No, the functions are `local` and exposing them is pointless and hacky.

> Isn't the API over complicated

Not really. Letting the `table` or `string` API is used elsewhere and is for convenience. Not a lot of people will use this API directly. It's really something for ClientDB/Tyrannical/Shifty and other "workflow" extensions. Even then, most source will have only 1 dependency. However not all. ClientDB must be after[1] Tyrannical if it's installed.

Also @mindeunix, this concerns you, I would like your review.

[1] after == higher priority


lib/awful/rules.lua
```diff
@@ -188,23 +190,164 @@ function rules.matches_list(c, _rules)
...
+function rules.remove_rule_source(name)
+    for k, v in ipairs(rules.rule_sources) do
+        if v.name == name then
+            table.remove(rules.rule_sources, k)
+            return true
+        end
+    end
+
+    return false
+end
```
psychon: How about dropping the above two function? People can do `table.insert(awful.rules.rule_source, my_source)` directly.  You can even have the complicated dependencies-stuff in one of your modules.

psychon: Also, so far we have a list of Tyannical and clientDB which might want to use this. One can always use `table.insert(awful.rules.rule_sources, foo)` while the other can always use `table.insert(awful.rules.rule_sources, 2, foo)`. So, nothing really needs the above complicated functions.

Elv13: I would argue it's bad design

1. It exposes internal structures [1]
2. One of the undisputed advantage of encapsulation is to increase the flexibility of internal structures by exposing a stable interface
3. It moves what's clearly boilerplate code into modules for no good reasons
4. It's only a few lines of code
5. It would force the `foo` in your example to be encoded in an undefined table format (that shall remains undefined)

[1] I really should make it fully ptivate, but I left it undocumented, so it is still private

lib/awful/rules.lua
```diff
@@ -188,23 +190,164 @@ function rules.matches_list(c, _rules)
...
+--  priority over.
+-- @treturn boolean The position in the list, `nil` if none was found.
+function rules.add_rule_source(name, callback, dependencies, prioritize_over)
+    local min_index, max_index = 1, #rules.rule_sources+1
+
+    dependencies = type(dependencies) == "string"
+        and {dependencies} or dependencies or {}
+
+    prioritize_over = type(prioritize_over) == "string"
+        and {prioritize_over} or prioritize_over or {}
```
psychon: How about at least getting rid of this and require the argument to be a string?

Elv13: sure

lib/awful/rules.lua
```diff
@@ -532,12 +674,11 @@ function rules.execute(c, props, callbacks)
     end
 end
 
+-- TODO v5 deprecate this
 function rules.completed_with_payload_callback(c, props, callbacks)
     rules.execute(c, props, callbacks)
 end
 
-client.connect_signal("spawn::completed_with_payload", rules.completed_with_payload_callback)
```
psychon: Can you also add a deprecated-todo to the corresponding bits in `awful.spawn` (as a separate commit)?

tests/test-awful-client.lua
```diff
@@ -264,6 +264,12 @@ table.insert(multi_screen_steps, function()
                 screen = screen[1],
             })
 
+    -- Same as previous, but switched
```
psychon: For the commit message: How does this fix any issue? Should that be in the previous commit's commit message instead?

tests/test-awful-rules.lua
```diff
@@ -282,10 +282,38 @@ test_rule {
...
     end
 }
 
+-- Test the custom sources
+assert(awful.rules.add_rule_source("before", function(c, props, _)
+    assert(type(c) == "client")
+    assert(not props.random1)
+
+    props.random2 = true
+end, {"awful.rules"}, {"awful.spawn"}))
```
psychon: Now that I see this: How about changing the argument order of `add_rule_source` so that the callback is last?

tests/test-awful-rules.lua
```diff
@@ -282,10 +282,38 @@ test_rule {
...
+
+assert(awful.rules.add_rule_source("before2", function(c, props, _)
+    assert(type(c) == "client")
+    props.random3 = true
+    assert(false)
```
psychon: How about replacing these three lines with `error("This function should not be called")`? That way the reader is not confused by so much dead code (and I prefer `error` over `assert(false)`).

tests/test-awful-rules.lua
```diff
@@ -282,10 +282,38 @@ test_rule {
...
+
+assert(awful.rules.remove_rule_source("before2"))
+
+assert(awful.rules.add_rule_source("after", function(c, props, _)
+    assert(type(c) == "client")
+    assert(props.random2)
+
+    props.random1 = true
+end, {"awful.spawn", "void"}))
+
```
psychon: Could you move the `test_rule` call to down here? I was going to complain that `random1` is unused, but instead it is checked by code that appears earlier in this file...

psychon:
Most important comment: https://github.com/awesomeWM/awesome/pull/1487#discussion_r98684652
The rest should be relatively trivial.

psychon:
I have a proposal what to do with that dependency systems: I still think that using strings will not work, because this means that modules must "know each other" so that they are ordered correctly. How about turning the priority into a single integer? `awful.rules` gets priority 0, `awful.spawn` gets 100 and the function to add a new entry just checks for a duplicate name and then `table.insert`s at the right position (or uses `table.sort()` with a callback function).

Elv13:
> How about turning the priority into a single integer? [...] must "know each other"

So each module has to know the integer use by the other modules? Sound like moving the problem around and making it worst (what if that module changes its number). This is a low-use, but necessary API. Strings will be good enough. I will do the fixup soon-ish, probably the weekend.

psychon:
> Sound like moving the problem around and making it worst (what if that module changes its number)

Then it likely has a reason for changing its number.

The idea was more meant as a simplification for the implementation (single `table.sort` (or even hand-written insertion sort) is enough while right now you need to implement a [topological sorting](https://en.wikipedia.org/wiki/Topological_sorting); also: numbers half-solve the "modules must know each other's names"-problem).

blueyed:
Numbers/ranges could also be documented.  It's easier that using names there then.
It's a bit like `.d` dirs, where files are prefixed with a number.

tests/test-awful-rules.lua
Elv13: Because they are optional and the callback isn't.

Elv13:
Rebased and fixes. Please review so the bug it fix doesn't end up in v4.1

> Numbers/ranges could also be documented. It's easier that using names there then.
> It's a bit like .d dirs, where files are prefixed with a number.

I disagree with numbers. Systems that used to use them (like SysVInit) now use names. Numbers simply wont work. I really don't see why we are arguing with this. It isn't 1000 lines of hard to maintain code. The line count different of numbers vs. names is 50 lines. That 50 lines prevent a design bug from going into the code.

tests/test-awful-client.lua
Elv13: done

lib/awful/rules.lua
Elv13: done

tests/test-awful-rules.lua
Elv13: done

lib/awful/rules.lua
```diff
@@ -204,23 +206,164 @@ function rules.matches_list(c, _rules)
...
         end
+
+        for _, v2 in ipairs(run_before) do
+            if v.name == v2 then
+                max_index = math.min(max_index, k)
+            end
+        end
+    end
+
+    if max_index + 1 < min_index then return false end
```
psychon: I fail to see how you addressed my counter-example showing that just using indices is not enough:
(For simplicity I'll assume there are no other rule sources)

- Add "a" without dependencies (this gets index 1)
- Add "b" without dependencies (this gets index 2)
- Add "c" with `run_before={"a"}` and `run_after={"b"}`. This addition is incorrectly rejected (it can be satisfied by reordering "a" and "b").

Reference to what you would need to implement: https://en.wikipedia.org/wiki/Topological_sorting

psychon:
Oh and one more thing:

> I disagree with numbers. Systems that used to use them (like SysVInit) now use names. Numbers simply wont work. I really don't see why we are arguing with this. It isn't 1000 lines of hard to maintain code. The line count different of numbers vs. names is 50 lines. That 50 lines prevent a design bug from going into the code.

Numbers have an intuitive total order while your dependency system still requires a topological sort. Topological sort is kind-of-orthogonal to the rest of this, so I wonder if that should be implemented in `gears`? (Although I doubt any other users of this API will show up... also, I am not sure how complicated an implementation of topological sorting would be)

lib/awful/rules.lua
Elv13:
```

        assert(awful.rules.add_rule_source(
            "a", function() end)
        )
        assert(awful.rules.add_rule_source(
            "b", function() end)
        )
        assert(awful.rules.add_rule_source(
            "c", function() end, {"a"}, {"b"})
        )
```

works

```

        assert(awful.rules.add_rule_source(
            "a", function() end)
        )
        assert(awful.rules.add_rule_source(
            "b", function() end)
        )
        assert(awful.rules.add_rule_source(
            "c", function() end, {"b"}, {"a"})
        )
```

doesn't.

It's even worst for number ordering. `priority 1 (a)` + `priority 1 (b)` and you can't insert `c` between `a` and `b`. Solving the problem with names is simple, just change the `require()` order in `rc.lua` and you are done.  Solving the problem with number is impossible without some hacky manual table re-order using function pointer comparison. Solving the name dependency problem with code *does* require solving a dependency graph. But the simple code used here is good enough a vast majority of the time and the workaround for the corner case where it isn't (and that probably wont ever happen to anyone) is trivial. So IMO the issue you raise is real, but in practice it is negligible. The worst thing with it is that it can silently not work. If you wish, I can add an `error()` message and improve the doc about this issue. I think it makes more sense than adding generic graph resolving code.

psychon: > It's even worst for number ordering. priority 1 (a) + priority 1 (b) and you can't insert c between a and b. 

I'd say that `awful.rules` gets priority (or value or whatever) 0 and `awful.spawn` gets 10000. That should leave a lot of space so that people can pick non-colliding numbers where necessary. This does not solve all problems, as you point out, but at least it should solve all issues that depend on the order in which the rule sources are added.

Elv13:
Are your opinion set in stone for the numbering indices? This PR fix a rather bad issue, so I would like it get it merged. If you insist, I will use number, even if I think they don't solve the problem and are confusing. However, I still prefer the current solution over that.

(ignore the most recent commit, it will be removed from the PR before it gets merged, I am just bored of carrying it around, so I put it there).

psychon:
> Are your opinion set in stone for the numbering indices? 

No, I am also fine with a version using strings that works. As far as I remember (I might misremember something here...), I always came up with a way where the code would incorrectly claim that there is a dependency loop. I am just arguing for numbers so much because then the whole issue of "we need a topological sort" goes away.

From a superficial look at the current state of this, there was still some cases where the code saw dependency loops were none are. Tell me if I should take a closer look again.

Elv13:
Ok, I am "done" creating a version of the sort+awful.rules combo that's good enough for another round of review. It still has the Linux specific hack commit and isn't ready to be merged as-is. After a round of review, I will split the PR in 3 (sort, awful.rules, Linux extension recipe).

I made quite a lot of changes to the sort code to make it work across the different Lua versions, not depend on `pairs()` order, write the spec tests and the doc tests. I also renamed all the "before" and "after" to something less ambiguous. The problem was that since some comments and code had different point of view, "before" and "after" switched meaning and that was very confusing.

lib/awful/rules.lua
```diff
@@ -207,23 +211,169 @@ function rules.matches_list(c, _rules)
...
---- Apply awful.rules.rules to a client.
--- @client c The client.
-function rules.apply(c)
 
-    local props = {}
+-- Contains the sources.
+-- The elements are ordered "first in, first executed". Thus, the higher the
+-- index, the higher the priority. Each entry is a table with a `name` and a
+-- `callback` field. This table is exposed for debugging purpose. The API
+-- is private and should be modified using the public accessors.
```
psychon: 1) This comment talks about index-based priorities. I thought the idea was to get rid of that?
2) If this is private, then why is it public (i.e.: not `local`) *and* documented?

lib/awful/rules.lua
```diff
@@ -207,23 +211,169 @@ function rules.matches_list(c, _rules)
...
+    end
+
+    rules._rule_source_sort:prepend(name, precede    )
+    rules._rule_source_sort:append (name, depends_on )
+
+    local res, err = rules._rule_source_sort:sort()
+
+    if err then
+        gdebug.print_warning("Failed to add the rule source: "..err)
+        return false
```
psychon: At this point the topological sort instance will always find the same cycle again. Thus, all following calls to `add_rule_source` will fail as well. I fear that caching the instance does not work.

lib/awful/rules.lua
```diff
@@ -207,23 +211,169 @@ function rules.matches_list(c, _rules)
...
+
+--- The default `awful.rules` source.
+--
+-- **Depends on:**
+--
+-- *nothing*
+--
+-- @rulesources awful.rules
+
+rules.add_rule_source("awful.rules", apply_awful_rules, {"awful.spawn"}, {})
```
psychon: The comment and the code disagrees. This passes `{"awful.spawn"}` for the `depends_on` parameter.
However, I have to admit that I am not quite sure what this means for the priorities... I guess that `awful.spawn` should overwrite rules from `awful.rules`, but then the dependency should be the other way around... Or not? I'm confused by all this wording (the fundamental problem: higher priority callbacks have to be called later so that they are not overwritten, so there is some kind of inversion going on...).

lib/awful/rules.lua
```diff
@@ -207,23 +211,169 @@ function rules.matches_list(c, _rules)
...
+
+--- The rule source for clients spawned by `awful.spawn`.
+--
+-- **Depends on:**
+--
+-- * `awful.rules`
+--
+-- @rulesources awful.spawn
+
+rules.add_rule_source("awful.spawn", apply_spawn_rules, {}, {"awful.rules"})
```
psychon: Why is this in this file and not in `awful.spawn`?

lib/awful/rules.lua
```diff
@@ -207,23 +211,169 @@ function rules.matches_list(c, _rules)
...
+    return true
+end
+
+--- Remove a source.
+-- @tparam string name The source name.
+-- @treturn boolean If the source was removed
+function rules.remove_rule_source(name)
+    for k, v in ipairs(rules.rule_sources) do
+        if v.name == name then
+            table.remove(rules.rule_sources, k)
```
psychon: This needs to update (recreate?) `rules._rule_source_sort` as well.

lib/awful/rules.lua
```diff
@@ -342,10 +342,41 @@ end
...
+    f:close()
+
+    local snid = s:match("STARTUP_ID=(.*)")
+
+    -- There is an extra \0 in the string
+    snid = snid and snid:sub(1, -2) or nil
+
+    -- They can be used only once. Otherwise it is an inherited environ or an
+    -- app with multiple windows.
+    snid = snid and (not blacklisted_snid[snid]) and snid or nil
```
psychon: `blackliste_snid` is always an empty table, so this condition will always evaluate to `true`...?

lib/awful/rules.lua
```diff
@@ -342,10 +342,41 @@ end
...
+    if not c.pid then return end
+
+    local f = io.open("/proc/"..c.pid.."/environ", "rb")
+
+    -- It will only work on Linux, that's already 99% of the userbase.
+    if not f then return end
+
+    local s = f:read("*all")
+
+    f:close()
```
psychon: What are the chances of this blocking?

lib/awful/rules.lua
```diff
@@ -342,10 +342,41 @@ end
...
+
+    local snid = s:match("STARTUP_ID=(.*)")
```
psychon: Sigh, this does not work since you also capture all following variables:
```
$ lua -e 'local c = "SOMETHING=ELSE" .. string.char(0) .. "FOO=bar" ..string.char(0) .. "NEXT_VAR=42" print(c:match("FOO=(.*)")) '                     
barNEXT_VAR=42
```
Luckily, it seems like `environ` always ends with a zero byte, so `s:match("STARTUP_ID=(.*)" .. string.char(0))` should do the trick.

lib/awful/rules.lua
```diff
@@ -342,10 +342,41 @@ end
...
+
+--- Make startup notification work for some clients like XTerm. This is ugly
+-- but works often enough to be useful.
+local function get_startup_id_fallback(c)
+    if not c.pid then return end
+
+    local f = io.open("/proc/"..c.pid.."/environ", "rb")
+
+    -- It will only work on Linux, that's already 99% of the userbase.
+    if not f then return end
```
psychon: For at least FreeBSD there seems to be `procstat`: https://forums.freebsd.org/threads/environment-variables-of-a-running-process.49190/

lib/awful/rules.lua
```diff
@@ -342,10 +342,41 @@ end
 
 rules.add_rule_source("awful.rules", apply_awful_rules, {"awful.spawn"}, {})
 
+local blacklisted_snid = {}
+
+--- Make startup notification work for some clients like XTerm. This is ugly
+-- but works often enough to be useful.
```
psychon: The commit message also talks about KF5, which seems to be part of KDE. How about fixing this in KDE / Qt, too? (Or is it at least reported as a bug somewhere?)

lib/gears/sort/topological.lua
```diff
@@ -0,0 +1,96 @@
...
+
+function tsort.topological()
+    return setmetatable({
+        edges = {},
+    }, mt)
+end
+
+return setmetatable(tsort, {__call = function(_, ...)
+    return tsort.topological(...)
+end})
```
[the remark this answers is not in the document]
Elv13: I wrote some based on you `rc.lua` diff. They are in their own commit, as usual in my PR.

lib/awful/rules.lua
Elv13: > This comment talks about index-based priorities.

Because it's the *result* of the sort, so it's index based ;)

> then why is it public (i.e.: not local) 

done

> and documented?

It is not publicly documented, it's `--` instead of `---` ;). The `awful.rules` code is already a spagatti mess just to be able to do everything in the right order, at least it's a well documented (and necessary...) mess.

lib/awful/rules.lua
Elv13: Fixed and added to the tests

lib/awful/rules.lua
Elv13: I changed the wording. This is why I remove the `after`/`before` names the original version had.

lib/awful/rules.lua
Elv13: Because it creates an indirect circular module dependencies. `awful.rules` can require `awful.spawn`, but not the opposite.

lib/awful/rules.lua
Elv13: Wont happen. The X11 support is frozen, the maintainer isn't interested in fixing this bug or messing with that code. I did talk to him directly about this (we also worked for the same employer for a while).

It's also not a simple fix. The Qt (X11Extra module) has support for this, but is not enabled by default. Then the KDE "QPA" (Qt Platform Plugin) for Linux also has code for this but it's under a dozen layer of abstraction and is never called due to a bunch of things that are unlikely to change.

lib/awful/rules.lua
Elv13: done

Elv13:
I removed (as mentioned since I added it) the "hacky Linux specific rule source" as it was here only has an example of relevant 3rd party sources along with the serialized client database.

lib/awful/rules.lua
psychon: For future reference, the cycle is `spawn` -> `rules` -> `placement` -> `client` -> `spawn`.
Sigh. :-(

PR #1486, opened 2017-01-28
doc: porting guide corrections, addition, and cleanup

boxofrox:
- added note that `awesome.quit` must be wrapped in an anonymous
  function when used as a menu command.
- corrected tasklist diff to include missing lines from rc.lua.  the
  porting guide code doesn't make much sense otherwise.
- cleaned up diff columns on a couple other code blocks.

These are couple points of confusion that tripped me up when porting my awesome config to v4.0.

PR #1488, opened 2017-01-29
Fix/porting guide code blocks

boxofrox:
In #1486 @Elv13 found that ldoc glitched when rendering code blocks I modified.  This PR resolves that issue and cleans up the porting guides code blocks a bit more.

I found that non-blocking-whitespace (0xa0) had the same effect on ldoc as Elv13's braille space (0x2800), but the nbsp has the same width as a space (0x32), where braille space does not.  If we'd rather have braille space, I'm happy to adjust this patch.

The first patch adds nbsp to correct #1486.  The second patch replaces the remaining braille space with nbsp, and adds nbsp to better align the code blocks.

I installed ldoc 1.4.6, reproduced the bug, and verified the patches correctly render.

Steps I used [modified for your perspective]
```
git clone git@github.com:awesomeWM/awesome.git
git fetch origin pull/1488/head:pr-1488
cd awesome
git checkout 420f6726  # PR 1486
cmake .
make ldoc
(cd doc && chromium documentation/17-porting-tips.md.html)
# review code blocks and observe missing diff columns
git checkout .
git checkout pr-1488
make ldoc
(cd doc && chromium documentation/17-porting-tips.md.html)
# review code blocks and observe all diff columns appear
```

### DIFF
--- a/docs/17-porting-tips.md
+++ b/docs/17-porting-tips.md
@@ -97,7 +97,7 @@ work. Remove this section.
 The `quit` menu command must be wrapped in a function, otherwise an error
 occurs due to mismatched argument types from the v4.0 `awful.menu` library.
 
-     -- {{{ Menu
+     -- {{{ Menu
      -- Create a laucher widget and a main menu
      myawesomemenu = {
        { "manual", terminal .. " -e man awesome" },
@@ -109,7 +109,7 @@ occurs due to mismatched argument types from the v4.0 `awful.menu` library.
 
 The textclock is now part of the `wibox` library, rename it.
 
-     -- {{{ Wibar
+     -- {{{ Wibar
      -- Create a textclock widget
     -mytextclock = `awful.widget.textclock`()
     +mytextclock = `wibox.widget.textclock`()


blueyed:
Thanks.
Having/requiring magic whitespace seems to be really bad, and prone to breakage in the future, isn't it?

Elv13:
But it solves the problem now, rather than next year. Users are upgrading now, so waiting for a better solution defies the purpose of that document ;)

Elv13:
@boxofrox Why did you add so many of them. The problem is there only once, right?

boxofrox:
Consistency I guess.  I can drop the second patch if desired.

After realigning the other code blocks in the markdown, the problem reappears.  I think ldoc gobbles up all N spaces at the beginning of a code block and throws away N characters for all other lines in the code block.  ldoc didn't see nbsp as a space, so it solved the problem in those cases.

boxofrox:
@blueyed I agree, it's definitely prone to breakage in the future.  Given ldocs behavior, my ocd favors perfectly rendered docs, so the price seemed small.

I can use a period or some other symbol to circumvent ldoc if you prefer.

Elv13:
> Consistency I guess. I can drop the second patch if desired.

I would prefer to keep the changes to a minimum. If it's not broken, don't fix it ;). You can also add extra spaces to fix the problem. Avoid having "---" and "strange indentation". 

@stevedonovan any idea about this? 

The "code" is a patch and has a 4 spaces indentation, but renders like

pick7

(it lacks the `-` in front of  `-    { "quit"`

boxofrox:
Retracted the second patch as requested.

boxofrox:
ldoc is a bit wacky.  If I take the nbsp out of my patch and use 6 spaces on the first line (4 for code block, 2 as code indent), `make ldoc` rendered correct once, and subsequent renders behave as I described above; ldoc is eating N characters from each line in the block.

```
diff --git a/docs/17-porting-tips.md b/docs/17-porting-tips.md
index 500fb432..e3f85542 100644
--- a/docs/17-porting-tips.md
+++ b/docs/17-porting-tips.md
@@ -97,7 +97,7 @@ work. Remove this section.
 The `quit` menu command must be wrapped in a function, otherwise an error
 occurs due to mismatched argument types from the v4.0 `awful.menu` library.
 
-     -- {{{ Menu
+      -- {{{ Menu
      -- Create a laucher widget and a main menu
      myawesomemenu = {
        { "manual", terminal .. " -e man awesome" },
```

This patch produced this image after a few runs of `make ldoc`.  Removing the nbsp from the `-- {{{ Wibar` line may have also had some effect.  It's wacky.

image

boxofrox:
Playing a bit more with this, I can add a blank line at the start of the code block (so long as only 4 spaces start the block, though this does add a blank line in the html)...

```diff
+    
      -- {{{ Menu
```

...or use an underscore...

```diff
-     -- {{{ Menu
+    _-- {{{ Menu
```

...or use a hyphen.  It's not critical if someone mistakes the first line as a diff removal, since it's only a comment.  And despite the `---`, it renders properly.  Perhaps 3 hyphens are only special for ldoc if not prefixed by spaces.

```diff    
-     -- {{{ Menu
+    --- {{{ Menu
```

PR #1492, opened 2017-01-30
fix(awful: key): clone table with user data

### LINKED ISSUE #1401: User data for awful.key cannot be shared
```
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✘
 • LGI version: 0.9.0
```

Somewhere in your config where you define keys, replace a local table with description/group with a variable that contains this table. Use the same variable in some other key.
E.g.
```lua
local aud = { description = "Raise the volume", group = "audio" }
local globalkeys = awful.util.table.join(
    awful.key({}, "XF86AudioRaiseVolume", apw.up, aud),
    awful.key({ modkey }, "Right", apw.up, aud),
)
```
Restart with the new config, open hotkeys popup.

**Actual result:**
image

**Expected result:**
There should have been two records for different keys with the same description/group. Or at least there should be a documentation notice on `awful.key` that you can't reuse `data` argument.

actionless:
closes #1401 

i've accidentially noticed what the change is not for the hotkey widget itself so no need to wait for its tests

### DIFF
--- a/lib/awful/key.lua
+++ b/lib/awful/key.lua
@@ -95,7 +95,7 @@ function key.new(mod, _key, press, release, data)
     end
 
     -- append custom userdata (like description) to a hotkey
-    data = data or {}
+    data = data and util.table.clone(data) or {}
     data.mod = mod
     data.key = _key
     table.insert(key.hotkeys, data)
--- a/tests/test-awesomerc.lua
+++ b/tests/test-awesomerc.lua
@@ -4,12 +4,6 @@ local awful = require("awful")
 
 local old_c = nil
 
-local function get_callback(mod, key)
-    local inf = {}
-    awful.key(mod, key, nil, nil, inf)
-
-    return inf.execute
-end
 
 -- Get a tag and a client
 local function get_c_and_t()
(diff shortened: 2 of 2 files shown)


psychon:
No idea why, but this causes:
```
== Running /home/travis/build/awesomeWM/awesome/tests/test-awesomerc.lua ==

Error: running function for step 2/11 (@1): [string "local awful = require("awful")..."]:60: attempt to call a nil value
```

@Elv13 Unrelated, but I noticed it due to this (next step would be to inline this so that no anonymous function is created):
```diff
diff --git a/tests/test-awesomerc.lua b/tests/test-awesomerc.lua
index 9076945..5db2646 100644
--- a/tests/test-awesomerc.lua
+++ b/tests/test-awesomerc.lua
@@ -5,10 +5,9 @@ local awful = require("awful")
 local old_c = nil
 
 local function get_callback(mod, key)
-    local inf = {}
-    awful.key(mod, key, nil, nil, inf)
-
-    return inf.execute
+    return function()
+        awful.key.execute(mod, key)
+    end
 end
 
 -- Get a tag and a client
```

psychon:
Actually... it's trivial why this breaks this `execute` callback. The better question is: Why does `awful.key` add an `execute` entry to the data table that it gets? I can't find any (reasonable) use of this (but I only did a quick grep). This is not even documented, so would it be an API break to just get rid of that `execute` entry?

actionless:
https://awesomewm.org/apidoc/libraries/awful.key.html#execute

i think we shoud drop get_callback helper from the test and replace `get_callback({modkey}, "t")()` to  `awful.key.execute({"modkey"}, "t")`

psychon:
> or do you mean what it's not documented as field of data table arg?

Yup. That's what your patch breaks, but the test depends on that.

> i think we shoud drop get_callback helper from the test and replace get_callback({modkey}, "t")() to awful.key.execute({"modkey"}, "t")

Yup.

actionless:
now travis seems happy

Elv13:
Thanks

PR #1499, opened 2017-01-31
client: add shape settings

copycat-killer:
Lets add `set_shape` to `client` as well.

Commits needed for this PR:

- [x] squash commit on size signals;
- [x] add a `property::size` signal in `client.c` to avoid the shape being applied twice;
- [x] add documentation to <s>`wibar.lua` and</s> the API docs.
- [x] add a test.

### DIFF
--- a/lib/awful/key.lua
+++ b/lib/awful/key.lua
@@ -95,7 +95,7 @@ function key.new(mod, _key, press, release, data)
     end
 
     -- append custom userdata (like description) to a hotkey
-    data = data or {}
+    data = data and util.table.clone(data) or {}
     data.mod = mod
     data.key = _key
     table.insert(key.hotkeys, data)
--- a/tests/test-awesomerc.lua
+++ b/tests/test-awesomerc.lua
@@ -4,12 +4,6 @@ local awful = require("awful")
 
 local old_c = nil
 
-local function get_callback(mod, key)
-    local inf = {}
-    awful.key(mod, key, nil, nil, inf)
-
-    return inf.execute
-end
 
 -- Get a tag and a client
 local function get_c_and_t()
@@ -42,14 +36,12 @@ local steps = {
         assert(old_c)
 
         -- Test layout
-        -- local cb = get_callback({modkey}, " ")
-        -- assert(cb)
 
         --TODO use the key once the bug is fixed
         local l = old_c.screen.selected_tag.layout
         assert(l)
 
-        -- cb()
+        --awful.key.execute({modkey}, " ")
         awful.layout.inc(1)
 
         assert(old_c.screen.selected_tag.layout ~= l)
@@ -57,7 +49,7 @@ local steps = {
         -- Test ontop
 
         assert(not old_c.ontop)
-        get_callback({modkey}, "t")()
+        awful.key.execute({modkey}, "t")
         awesome.sync()
 
         return true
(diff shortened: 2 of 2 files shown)


Elv13:
Thanks **a lot** for this and *finally :)* enter the official commit log. As discussed on IRC:

 * create a `property::size` for clients in `client.c` to avoid the code being called twice
 * connect the `border_width` and size signals to `client.object._apply_shape`
 * add a documentation block in `wibar.lua`
 * add a test to make CodeCov happy, this can be done in `test-struts.lua` by adding a shape to the wibars there. As they are already programmed to move and resize, the code should be stressed enough

psychon:
Sorry, but this conflicts with `awful.client.shape` which already tries to do *some* shape handling on clients (this code is what makes `xeyes` slightly less ugly). I think a new solution would need to incorporate this handling. How "bad" would it be to have the "actual applying of a shape" in `awful.client.shape` and `awful.client` just calls functions from there? (Or `awful.client.shape` monkey-patches `awful.client` to avoid a `require()`-dependency-loop)

Elv13:
>  How "bad" would it be to have the "actual applying of a shape" in awful.client.shape

Right. `_apply_shape` could become a `local function` and be called in `if c._shape then apply_shape(c); return`. Adding the `property::border_width` will also be required, as found by copycat-killer.

copycat-killer:
Moved the shaping to `awful.client.shape.set`.

@Elv13 I don't understand the relation with `wibar.lua`: why (and where) do I have to add documentation in it?

Elv13:
> @Elv13 I don't understand the relation with wibar.lua: why (and where) do I have to add documentation in it?

Nevermind. I was confused with wibox shapes again. However the problem is real, but forget it. I will create a PR to fix that.

copycat-killer:
So, it seems that documentation is grabbed directly from the code, and I added the proper comments.

Any suggestions about the test?

lib/awful/client/shape.lua
```diff
@@ -51,8 +51,54 @@ function shape.get_transformed(c, shape_name)
     return result
 end
 
+--- Set the client shape.
+-- @function awful.client.shape.set
+-- @client c The client whose shape should be retrieved
+-- @tparam gears.shape A gears.shape compatible function.
+-- @see gears.shape
+function shape.set(c, shape)
```
Elv13: `function client.object.set_shape` is still required to eixst in `client.lua`. Otherwise `c.shape = ...` wont work

Elv13:
> Any suggestions about the test?

What about `client.get()[1].shape = gears.shape.rounded_rect` in `test-resize.lua:18`. This should apply the shape and resize it in all sorts of ways.

lib/awful/client/shape.lua
copycat-killer: Added.

lib/awful/client/shape.lua
```diff
@@ -87,6 +133,9 @@ capi.client.connect_signal("property::shape_client_bounding", shape.update.bound
 capi.client.connect_signal("property::shape_client_clip", shape.update.clip)
 capi.client.connect_signal("property::width", shape.update.all)
```
Elv13: If `size` works, then `width` and `height` can be removed, right?

Elv13: Oh, and it should call `shape.update.all`, not `set_shape`.

copycat-killer: It seems so.

copycat-killer:
Added test too.

Elv13:
I think you accidentally removed all commits

copycat-killer:
Haha yeah, I moved my mods to a dedicated branch, but forgot that this PR was bound to my master.

Will open a new one.

PR #1501, opened 2017-02-01
Move the wibar:remove() call outside of the loop over all wiboxes.

### LINKED ISSUE #1498: After multiple quick screen disconnects (from xrandr --off) some widgets get confused

```
awesome v4.0-154-gefeb0ad (Harder, Better, Faster, Stronger)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✘
 • LGI version: 0.9.1
```

I have 3 external monitors connected through laptop docking station, two left of laptop, one right (since setting above causes other issues). When I run my script to run `xrandr --output ${xxx} --off` for each monitor (except the laptop), I seem to lose the systray and get an extra bottom wibar which isn't snapped to the bottom of the screen.

**Actual result:**

[screenshot](http://imgur.com/NhiY3uV)
Looking at my bottom wibar the tasklist gets cut off where the systray would normally start, systray is missing, and there's an extra? wibar floating oddly.

**Expected result:**

I don't get weird widget artifacts.

Veratil:
Removing a wibar mid-loop will corrupt the indexing of the for loop and can cause other wibars from being removed.

Resolves #1498.

### DIFF
--- a/lib/awful/wibar.lua
+++ b/lib/awful/wibar.lua
@@ -359,10 +359,14 @@ function awfulwibar.new(arg)
 end
 
 capi.screen.connect_signal("removed", function(s)
+    local wibars = {}
     for _, wibar in ipairs(wiboxes) do
         if wibar._screen == s then
-            wibar:remove()
+            table.insert(wibars, wibar)
         end
+    end
+    for _, wibar in ipairs(wibars) do
+        wibar:remove()
     end
 end)
 


Elv13:
> @Elv13 Do we need a new test for this? (regression tests, regression tests, regression tests :-) )

I don't have time right now, but I already have some code in my stash for this

https://github.com/awesomeWM/awesome/pull/1153/commits/9c8576a198390b8b0fefb7ce462ff992f96a518c

```lua
+    -- Draw all normal wiboxes
+    for _, d in ipairs(drawin.get()) do
+        local w = d.get_wibox and d:get_wibox() or nil
+
+        assert((not w._screen) and w._screen.valid)
+    end
```

in test-screen-changes.lua

Elv13:
@blueyed  I don't have time for this anytime soon. Better just merge this and open an issue for the test

blueyed:
Done.

Thanks for the report and fix, @Veratil!

PR #1502, opened 2017-02-01
Add aweful.util.pread() so API docs have deprication notice

alerque:
I spent a while chasing my tail because the API docs don't mention that `aweful.util.pread()` got deprecated.


lib/awful/util.lua
```diff
@@ -344,6 +344,13 @@ function util.get_rectangle_in_direction(dir, recttbl, cur)
     return grect.get_in_direction(dir, recttbl, cur)
 end
 
+-- @deprecated awful.util.pread
+function util.pread ()
+    util.deprecate("Use io.popen() directly or look at awful.spawn.easy_async() "
+            .. "for an asynchronous alternative")
+    return ""
+end
```
psychon: This kind-of-conflicts with the same thing done in `lib/awful/init.lua`. No time to test right now, but would it also work if there were no actual lua-function after the doc comment?
(Actually: Does the proposed patch work? The doc comment begins with `--` but I thought `---` is needed?)
(Also: Typo in the commit message: awEful)

alerque: I added the `---` and fixed the typo. I wasn't sure how the thing marked as a hack in `init.lua` worked. I assume it's to get a round a race condition or something, but is seems like _this_ would be the correct place for it and the other one should be dropped. Given Lua's namespacing I figured re-declaring it couldn't hurt but not having it at all might create a problem.

actionless: but why not to drop another workaround in this PR?

Elv13:
Rebased and simplified to avoid the duplication so this PR can be merged.

PR #1504, opened 2017-02-01
Add custom keys to existing groups

getzze:
So the keydocs for the keybindings to numbered tags can be removed and put this instead:
```
hotkeys_popup.add_hotkeys({
    ["tag"]      = {{modifiers = {modkey,                   }, keys = {["#"]="view tag #(1-9)"}},
                    {modifiers = {modkey, "Control"         }, keys = {["#"]="toggle tag #(1-9)"}},
                    {modifiers = {modkey, "Shift"           }, keys = {["#"]="move client to tag #(1-9)"}},
                    {modifiers = {modkey, "Control", "Shift"}, keys = {["#"]="toggle client on tag #(1-9)"}},
    }
})
```
Otherwise, these additional hotkeys cannot be added to the `tag` group.

### DIFF
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -227,7 +227,7 @@ function widget.new()
         local max_height_px = height - group_label_height
         local column_layouts = {}
         for _, group in ipairs(available_groups) do
-            local keys = cached_awful_keys[group] or widget_instance.additional_hotkeys[group]
+            local keys = awful.util.table.join(cached_awful_keys[group], widget_instance.additional_hotkeys[group])
             local joined_descriptions = ""
             for i, key in ipairs(keys) do
                 joined_descriptions = joined_descriptions .. key.description .. (i~=#keys and "\n" or "")


lib/awful/hotkeys_popup/widget.lua
```diff
@@ -227,7 +227,7 @@ function widget.new()
         local max_height_px = height - group_label_height
         local column_layouts = {}
         for _, group in ipairs(available_groups) do
-            local keys = cached_awful_keys[group] or widget_instance.additional_hotkeys[group]
+            local keys = awful.util.table.join(cached_awful_keys[group], widget_instance.additional_hotkeys[group])
```
Elv13: If possible, avoid the `awful.util.table.join`. It's slow and unnecessary. Better add a second level to the 2 for loops that use it.

getzze: It's complicated, because the rest of the function uses `#keys` and split the keys table in case of multiple pages. It is handled better (at least more easily readable) if all the keys are in a single table.

actionless: but whole function is cached in `cached_wiboxes` so i think it's fine

blueyed:
Thanks!

Issue #1503, opened 2017-02-01
Layout box does not update when swapping tags


mphe:
Output of `awesome --version`:

awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1

**How to reproduce the issue:**

* Add a layoutbox widget

* Add this snippet to rc.lua (it preserves tags when adding and removing a screen at the same time with the same position and size).

```lua
tag.connect_signal("request::screen", function(t)
    for s in screen do
        if s ~= t.screen and
           s.geometry.x == t.screen.geometry.x and
           s.geometry.y == t.screen.geometry.y and
           s.geometry.width == t.screen.geometry.width and
           s.geometry.height == t.screen.geometry.height then
            local t2 = awful.tag.find_by_name(s, t.name)
            if t2 then
                t:swap(t2)
            else
                t.screen = s
            end
            return
        end
    end
end)
```

Then replace a display with another display of the same size and position. The tags of the previous screen will be swapped with the tags on the new screen.

**Actual result:**

The layoutbox doesn't update until switching tags. It still shows the layout of the selected tag on the new screen, before it was swapped.

**Expected result:**

The layoutbox would update accordingly to the layout of the selected tag.

**Example**

Display A: 0, 0, 1920x1080 (active)
Display B: 0, 0, 1920x1080 (inactive)

Tag 9 is selected and has the corner layout. Every other tag has the tile layout (default).
Layoutbox shows corner layout.

`xrandr --output Display-A --off --output Display-B --auto`

Tag 9 is selected and has the corner layout. Every other tag has the tile layout.
Layoutbox shows tile layout.

Elv13:
The problem is much more simple than that: the layoutbox code currently doesn't handle this at all. There is literally no code to track screen changes, so it doesn't.

psychon:
Ok. That's (almost) easy. @Elv13 Any way to get the old screen in a tag's `property::screen` signal? Otherwise, the code will have to update all layoutboxes when a tag is moved to another screen (which wouldn't be bad either, because nothing *actually* changes so the update would be cheap).

Elv13:
So far, no:

```lua
function tag.setproperty(_tag, prop, value)
    if not _tag.data.awful_tag_properties then
        _tag.data.awful_tag_properties = {}
[...]

PR #1505, opened 2017-02-01
layoutbox: Update when a tag's screen changes

psychon:
In theory it would be enough to only update the old and new screen of
the tag whose screen changed. However, we cannot easily get the old
screen, so just update all layoutboxes.

Fixes: https://github.com/awesomeWM/awesome/issues/1503

### DIFF
--- a/lib/awful/widget/layoutbox.lua
+++ b/lib/awful/widget/layoutbox.lua
@@ -48,6 +48,13 @@ function layoutbox.new(screen)
         boxes = setmetatable({}, { __mode = "kv" })
         capi.tag.connect_signal("property::selected", update_from_tag)
         capi.tag.connect_signal("property::layout", update_from_tag)
+        capi.tag.connect_signal("property::screen", function()
+            for s, w in pairs(boxes) do
+                if s.valid then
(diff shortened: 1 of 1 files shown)


psychon:
Sigh. First attempt didn't work. It caused lots of the following: 
```
error: lib/awful/widget/layoutbox.lua:17: invalid object
```
I now added the following change to handle this:
```diff
diff --git a/lib/awful/widget/layoutbox.lua b/lib/awful/widget/layoutbox.lua
index 855c430d9..5c1df49e0 100644
--- a/lib/awful/widget/layoutbox.lua
+++ b/lib/awful/widget/layoutbox.lua
@@ -50,7 +50,9 @@ function layoutbox.new(screen)
         capi.tag.connect_signal("property::layout", update_from_tag)
         capi.tag.connect_signal("property::screen", function()
             for s, w in pairs(boxes) do
-                update(w, s)
+                if s.valid then
+                    update(w, s)
+                end
             end
         end)
         layoutbox.boxes = boxes
```
(Yay, the tests work and catch problems!)

blueyed:
Thanks!

PR #1507, opened 2017-02-01
Allow clients to set shape

copycat-killer:
*The squasher strikes back*

This is #1499 , which I accidentally deleted in my fork.

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -9,6 +9,7 @@
 -- Grab environment we need
 local util = require("awful.util")
 local spawn = require("awful.spawn")
+local set_shape = require("awful.client.shape").update.all
 local object = require("gears.object")
 local grect = require("gears.geometry").rectangle
 local pairs = pairs
@@ -1153,6 +1154,15 @@ function client.object.is_transient_for(self, c2)
         tc = tc.transient_for
     end
     return nil
+end
+
+--- Set the client shape.
+-- @property shape
+-- @tparam gears.shape A gears.shape compatible function.
+-- @see gears.shape
+function client.object.set_shape(self, shape)
+    client.property.set(self, "_shape", shape)
+    set_shape(self)
 end
 
 -- Register standards signals
--- a/lib/awful/client/shape.lua
+++ b/lib/awful/client/shape.lua
@@ -18,7 +18,7 @@ local shape = {}
 shape.update = {}
 
 --- Get one of a client's shapes and transform it to include window decorations.
--- @function awful.shape.get_transformed
+-- @function awful.client.shape.get_transformed
 -- @client c The client whose shape should be retrieved
 -- @tparam string shape_name Either "bounding" or "clip"
 function shape.get_transformed(c, shape_name)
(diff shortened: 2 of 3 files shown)


Elv13:
You can run the tests locally if you have Xephyr installed. There is a "run.sh" script. In this case, `local gears =require("gears")` is missing

copycat-killer:
There's an assertion fail with `test-awful-widget-button.lua`. Is it related?

Elv13:
I don't see such assertion in the log. Maybe it happens locally on your system (or maybe you moved the mouse during the tests). WhatI see is

```

    lib/awful/client/shape.lua:55:11: shadowing upvalue shape on line 17

    lib/awful/client/shape.lua:102:23: shadowing upvalue shape on line 17

```

You can install `luacheck` and add it as a commit hook to spot those before they get into Travis (to save time)

copycat-killer:
Travis shows this error which I can't understand:

```shell
2017-02-01 21:31:49 W: awesome: a_glib_poll:389: Last main loop iteration took 1.015218 seconds! Increasing limit for this warning to that value.
2017-02-01 21:31:52 W: awesome: a_glib_poll:389: Last main loop iteration took 2.403711 seconds! Increasing limit for this warning to that value.
2017-02-01 21:32:08 W: awesome: a_glib_poll:389: Last main loop iteration took 2.995378 seconds! Increasing limit for this warning to that value.
2017-02-01 21:32:51 W: awesome: a_glib_poll:389: Last main loop iteration took 3.511933 seconds! Increasing limit for this warning to that value.
2017-02-01 21:33:22 W: awesome: a_glib_poll:389: Last main loop iteration took 5.520004 seconds! Increasing limit for this warning to that value.
2017-02-01 21:34:11 W: awesome: a_glib_poll:389: Last main loop iteration took 6.302974 seconds! Increasing limit for this warning to that value.
2017-02-01 21:34:35 W: awesome: a_glib_poll:389: Last main loop iteration took 6.333373 seconds! Increasing limit for this warning to that value.
Awesome was killed due to timeout after 180 seconds
===> ERROR running /home/travis/build/awesomeWM/awesome/tests/test-awful-client.lua <===
```

Elv13:
Ignore that one and its sibling in test-awful-tag. The Travis node didn't have enough memory and started swapping. We have a hack to "fix" it in another PR, but it isn't merged yet (and is very ugly).

copycat-killer:
This PR introduces `property::size`. What about introducing `property::position` for x,y as well?

psychon:
Sure.
How about adding those new signals in a separate PR to get things in quicker and to have a smaller PR?

copycat-killer:
Do we have to wait for the Travis fix before deciding on this PR, or is there something else I can do?

psychon:
Travis failed this time because:
```
lib/awful/client/shape.lua:59:23: shadowing upvalue shape on line 17
```
Besides that, I don't have enough time to look much into this right now. I can take a look at the weekend, I think.

copycat-killer:
Solved.

Elv13:
This needs to be rebase. This time it may be easier to just re-apply the patch without the commit. Also make sure to merge the 2 code paths. XEye still has to work if no shape is set.

actionless:
but this version of shape routine (`shape.update.all`) will change the shape only of an outer part of a client's border

i would recommend to use the later revision of that code, which can be found in my local config:
https://github.com/actionless/awesome_config/blob/devel/config/signals.lua#L25-L78

Elv13:
@actionless the copycat version looks alright to me
pick8

actionless:
how did you configured it?

i did only `c.shape = gears.shape.rounded_rect`:

2017-02-03--1486134400_1386x1330_scrot

Elv13:
Uhh, I did the same (I added it to awful.rules). Does it only happens with GTK3 apps?

Elv13:
Merged with some fixes by psychon and me

actionless:
no, it happened the same with any apps, i will later on test the latest commit which got merged, mb there it will be fine for me

actionless:
i think the problem what i am using 4 titlebars instead of borders

however the only reason why i am using titlebars there is because i can't subscribe to mouse clicks on a client's borders

so i am not even sure to which of two directions i should move with that

psychon:
@actionless Did #1521 help with that? I think it should/might...

actionless:
no, it didn't

actionless:
so, let me rephrase my previous question:
should i try to fix mouse-event on a real client's border or should i update the current client_shape implementation to bring the changes from my local ones to support inner border shaping for the titlebars?

psychon:
> should i try to fix mouse events on a real client's border

Feel free. However, X11 won't let you do that. Feel free to try with `xev` (that's what I did :-) ).

psychon:
Today while showering I thought: This cannot work.

We apply the shape at the outside of the border and its inside. Since you have `border_width = 0`, the two are "at the same position".

For shaping titlebars: Each pixels can have one of three states:
* Outside of the window (not contained in the bounding shape)
* In the border (inside the bounding shape, outside the clip shape)
* Inside the window (inside of both shapes)

Since you have no border, there are no pixels inside of the border (and it wouldn't make sense to add a border "in the middle of the window"; Plus mouse clicks wouldn't be registered here). Of course, the area were you want the shape to be applied should not be outside of the window, because then the window would just have a hole. That leaves "inside the window" as the only alternative and this is just what we are doing.

actionless:
but with the apply_shape code from my config it's working asexpected:

2017-02-07--1486472668_370x292_scrot

psychon:
You have `border_width = 0` and use titlebars instead, right? What is visible in the "inside" of the rounded corner there? Put differently: How do the two shapes look like here?

actionless:
yup, border width is 0, but it seems to be unrelated because i still can't reproduce the problem with any other border width

i think the main difference is about coordinates of the shape -- in my example the square is smaller, ie:

upstream:
`_shape(cr, geom.width + 2*c.border_width, geom.height + 2*c.border_width)`

mine:
`shape(cr, geo.width, geo.height, shape_args)`

Elv13:
> shape(cr, geo.width, geo.height, shape_args)

For the shape_args part, it can be added later without breaking the API. If you open an issue for it, I can take care of it.

actionless:
sorry, shape_args wasn't the point, the point was a smaller geometry

Elv13:
I know, it was a side note. It's why I said to create an issue if you wan it

PR #1510, opened 2017-02-02
Add position and size property signals

copycat-killer:
Add two new client properties: `property::size` and `property::position`.

Useful to avoid doubling signals, and to #1507 in particular.

PR #1513, opened 2017-02-02
object/client.c: added position and size properties

Elv13:
added client signals connection

added comments doc

Close #1510


Elv13:
Thanks!

Issue #1495, opened 2017-01-31
Tests fail: An unexpected screen change did occur


qwelyt:
Output of `awesome --version`:

Awesome 4.0 source. Building with lua 5.3.

**How to reproduce the issue:**

Download source. 
Run `make`. 
`cd tests` `./run.sh`

**Actual result:**
```
qwelyt@machine ~/awesome-4.0/tests % ./run.sh         
awesome_log: /tmp/tmp.2ka1KoC7rr/_awesome_test.log
== Running /home/qwelyt/awesome-4.0/tests/test-awesomerc.lua ==
Test finished successfully
== Running /home/qwelyt/awesome-4.0/tests/test-awful-client.lua ==
2017-01-31 08:28:50 W: awesome: a_glib_poll:389: Last main loop iteration took 0.221708 seconds! Increasing limit for this warning to that value.
An unexpected screen change did occur
stack traceback:
	[string "local awful = require("awful")..."]:140: in function <[string "local awful = require("awful")..."]:136>
	[C]: in metamethod '__newindex'
	lib/awful/rules.lua:440: in function 'awful.rules.execute'
	lib/awful/rules.lua:536: in function 'awful.rules.completed_with_payload_callback'
	[C]: in method 'emit_signal'
	lib/awful/spawn.lua:217: in function 'awful.spawn.on_snid_callback'
screen3 Screen: 1
Error: running function for step 9/96 (@1): [string "local awful = require("awful")..."]:169: assertion failed!
stack traceback:
	[C]: in function 'assert'
	[string "local awful = require("awful")..."]:169: in function <[string "local awful = require("awful")..."]:165>
	(...tail calls...)
	[C]: in function 'xpcall'
	/home/qwelyt/awesome-4.0/tests/_runner.lua:47: in function </home/qwelyt/awesome-4.0/tests/_runner.lua:41>
	[C]: in function 'xpcall'
	lib/gears/protected_call.lua:36: in function <lib/gears/protected_call.lua:35>
	(...tail calls...)
	lib/gears/timer.lua:167: in function <lib/gears/timer.lua:165>!
===> ERROR running /home/qwelyt/awesome-4.0/tests/test-awful-client.lua! <===
Error: running function for step 9/96 (@1): [string "local awful = require("awful")..."]:169: assertion failed!
```

**Expected result:**

I expect all tests to pass.

blueyed:
Does it also happen on Git master?

psychon:
[...]

PR #1516, opened 2017-02-02
test-current-desktop: make wait_for_current_desktop fail directly

blueyed:
We should get the expected state before already.

Ref: https://github.com/awesomeWM/awesome/issues/1495#issuecomment-277098634

### DIFF
--- a/tests/test-current-desktop.lua
+++ b/tests/test-current-desktop.lua
@@ -23,6 +23,7 @@ local function wait_for_current_desktop(tag)
             return true
         end
         print(string.format("Got _NET_CURRENT_DESKTOP = '%s', expected %d", value, idx))
+        return false
     end
 end
 


psychon:
The test failures say that the theory about "the loop is not needed" is wrong. If adding a call to `awesome.sync()` before the `io.popen("xprop....")` helps, I'm fine with that, but without this there is definitely a race between the X11 server handling the awesome's request to update the property on the root window and `xprop` reading said property.

blueyed:
The DO_COVERAGE=codecov build failed sometimes.

But the strange thing is that the non-DO_COVERAGE builds error, but do not fail.  That should only happen when the non-script sections fail.
I think it's related to using `travis_fold_end` after a failure, and should be fixed by now.

blueyed:
It seems we were missing only a single extra step, which is fixed in master by now.
Restarted all jobs to be sure.

PR #1531, opened 2017-02-06
Add a utility checking uses of require()s

psychon:
Currently, "everything can require everything". It's an unstructured mess which sometimes causes problems.
    
This commit adds a tool that enforces a white-list of require() uses. It uses depgraph to scan the source code and then each use of require() that is found is checked. If any violations are found, the tool returns
a failure.
    
This tool is wired up to a new target "make check-requires" which is included in "make check". Thus, Travis will run this.

Edit: Forgot to reference #1400 and https://github.com/mpeterv/depgraph/issues/1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,6 +88,8 @@ install:
   - travis_retry sudo luarocks install busted
   # Install luacheck for "make check-qa".
   - if [ "$DO_CHECKQA" = 1 ]; then travis_retry sudo luarocks install luacheck; fi
+  # Install depgraph for "make check-qa".
+  - if [ "$DO_CHECKQA" = 1 ]; then travis_retry sudo luarocks install depgraph; fi
 
   # Install ldoc for building docs.
   - |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -381,6 +381,13 @@ add_custom_target(check-integration
     COMMENT "Running integration tests"
     USES_TERMINAL
     VERBATIM)
+add_custom_target(check-requires
+    lua "${CMAKE_SOURCE_DIR}/build-utils/check_for_invalid_requires.lua"
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    COMMENT "Checking use of require()"
+    USES_TERMINAL
+    VERBATIM)
+list(APPEND CHECK_QA_TARGETS check-requires)
 a_find_program(BUSTED_EXECUTABLE busted FALSE)
 if(BUSTED_EXECUTABLE)
     # Keep the arguments in sync with the version below!
--- a/build-utils/check_for_invalid_requires.lua
+++ b/build-utils/check_for_invalid_requires.lua
@@ -1,0 +1,119 @@
+#!/usr/bin/env lua
+
+-- This uses the "depgraph" package to parse all source code and look for calls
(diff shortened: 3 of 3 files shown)


psychon:
Travis build result for https://github.com/awesomeWM/awesome/pull/1531/commits/a493a156bd690e1d330c92e792f63e669b76cfe1 (that is, with https://github.com/awesomeWM/awesome/pull/1531/commits/588110e1ceca98d8764e432173dd022368ca0ad0): https://travis-ci.org/awesomeWM/awesome/builds/198922922

I will now do a force-push to remove that "break things" commit.

.travis.yml
```diff
@@ -88,6 +88,8 @@ install:
   - travis_retry sudo luarocks install busted
   # Install luacheck for "make check-qa".
   - if [ "$DO_CHECKQA" = 1 ]; then travis_retry sudo luarocks install luacheck; fi
+  # Install depgraph for "make check-requires".
+  - travis_retry sudo luarocks install depgraph
```
Elv13: Please add a `CHECK_DEPGRAPH` variable and only instal/use depgraph if it's set. Only 1 target should run this. Until Docker is working, those luarocks are "costly".

blueyed: I would argue that we could also just include it in `check-qa`?!

Elv13: Yes, and `check-qa` could then be moved to `LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build` instead of the coverage build (the second slowest)

psychon: At least I now know that this code (currently) works with all relevant Lua versions.

psychon: Rebased on master and added a fixup commit to address the above (moving `check-qa` to another travis target is outside the scope of this PR)

actionless:
so what about moving `awful.util` to be a first-level module? any volunteers? :)

Elv13:
If it moves, it will probably be into `gears` (the generic utils module). It also need to be broken down and have all the aliases into `awful.util` with future deprecation comment and/or improve `gears.depracate` (currently `awful.util.deprecated`) to support future deprecation based on the API level (lets call the current one "v4"). Feel free to do so :P

actionless:
i was thinking to do it easier:
1) `mv awful/util.lua new_name.lua`
2) replace imports
3) in `awful/util.lua` write smth like:
`return new_name.deprecate_class(new_name, "awful.utils", "new_name")`

actionless:
also regarding new name, mb `helpful`

Elv13:
We already have module dedicated for helpers, we really don't need 2. Also, the problem with `awful.util` is that it's all over the place. The scope needs to be redefined otherwise moving it is not worth the trouble.

actionless:
`awful.utils` -> `gears.utils` (`gears.misc`?)

Elv13:
I think you miss the point I am trying to make. There is multiple *types* of functions in `awful.util`. Every time something becomes complex, such as `awful.util.spawn`, it has to be moved elsewhere to prevent cyclic dependencies. The module need to be broken down into concerns, like `gears.table`, `gears.string`, `gears.deprecate` and so on. Modules that currently use `awful.util.deprecated` don't care about `awful.util.ensure_pango_color`.

PR #1533, opened 2017-02-06
refactor(awful: hotkeys_popup): expose configuration options for a widget instance and use more object-oriented structure for the widget

### LINKED ISSUE #1352: Why does awful.hotkeys_popup use notification_opacity
This seems at least surprising. I wouldn't consider this a notification. Can it use its own beautiful key instead?
https://github.com/awesomeWM/awesome/blob/3fea2db8c5f8cd5a2f3ed463fe57dcdc673c3afc/lib/awful/hotkeys_popup/widget.lua#L338

### LINKED ISSUE #1497: [awful.hotkey_widget] get theme parameters when creating widget's wibox, not when initializing widget object itself
should fix the problem found by #1471 and previous psychons's attempts to figure out the mess with beautiful import order/cyclic deps

actionless:
closes #1352
closes #1497

### DIFF
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -55,36 +55,50 @@ widget.hide_without_description = true
 widget.merge_duplicates = true
 
 
---- Widget background color.
+--- Hotkeys widget background color.
 -- @beautiful beautiful.hotkeys_bg
+-- @tparam color hotkeys_bg
 
---- Widget foreground color.
+--- Hotkeys widget foreground color.
 -- @beautiful beautiful.hotkeys_fg
+-- @tparam color hotkeys_fg
 
---- Border width.
+--- Hotkeys widget border width.
 -- @beautiful beautiful.hotkeys_border_width
+-- @tparam int hotkeys_border_width
 
---- Border color.
+--- Hotkeys widget border color.
 -- @beautiful beautiful.hotkeys_border_color
+-- @tparam color hotkeys_border_color
 
---- Widget shape.
+--- Hotkeys widget shape.
 -- @beautiful beautiful.hotkeys_shape
+-- @tparam[opt] gears.shape hotkeys_shape
 -- @see gears.shape
 
 --- Foreground color used for hotkey modifiers (Ctrl, Alt, Super, etc).
 -- @beautiful beautiful.hotkeys_modifiers_fg
+-- @tparam color hotkeys_modifiers_fg
 
---- Foreground color used for group and other labels.
+--- Background color used for miscellaneous labels of hotkeys widget.
+-- @beautiful beautiful.hotkeys_label_bg
+-- @tparam color hotkeys_label_bg
+
+--- Foreground color used for hotkey groups and other labels.
 -- @beautiful beautiful.hotkeys_label_fg
+-- @tparam color hotkeys_label_fg
 
---- Main widget font.
+--- Main hotkeys widget font.
 -- @beautiful beautiful.hotkeys_font
+-- @tparam string|lgi.Pango.FontDescription hotkeys_font
 
 --- Font used for hotkeys' descriptions.
 -- @beautiful beautiful.hotkeys_description_font
+-- @tparam string|lgi.Pango.FontDescription hotkeys_description_font
 
 --- Margin between hotkeys groups.
 -- @beautiful beautiful.hotkeys_group_margin
+-- @tparam int hotkeys_group_margin
 
 
 --- Create an instance of widget with hotkeys help.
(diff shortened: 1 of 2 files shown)


Elv13:
Didn't really do a review yet, but the first this I notice is that the `@beautiful` documentation blocks seem to be missing. They are required for the page that list all possible theme variables (and the hotkey own doc page, of course).

actionless:
@Elv13 submitted

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -44,22 +55,71 @@ widget.hide_without_description = true
 widget.merge_duplicates = true
 
 
+--- Widget background color.
+-- @beautiful beautiful.hotkeys_bg
```
Elv13: I really feel like I am nitpicking right now, but you should use the "hotkeys" or "hotkeys widget" in the description and add the type to all of them. See https://awesomewm.org/apidoc/documentation/06-appearance.md.html 

Not really important, but coherency, boring predictability and conventions goes a long way when you document a library.

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -100,55 +160,77 @@ function widget.new()
...
+        self.border_width = args.border_width or
+            beautiful.hotkeys_border_width or beautiful.border_width
+        self.border_color = args.border_color or
+            beautiful.hotkeys_border_color or self.fg
+        self.shape = args.shape or beautiful.hotkeys_shape
+        self.modifiers_fg = args.modifiers_fg or
+            beautiful.hotkeys_modifiers_fg or beautiful.bg_minimize or "#555555"
+        self.label_fg = args.label_fg or
+            beautiful.hotkeys_label_fg or self.bg
+        self.opacity = beautiful.notification_opacity or 1
```
Elv13: As a couple of people already told us on IRC, this still make little sense. Please add a proper variable for it. You can fallback to `beautiful.notification_opacity` is you thing it's worth it. I would not mind the behavior break if this was changed.

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -100,55 +160,77 @@ function widget.new()
...
+            beautiful.hotkeys_bg or beautiful.bg_normal
+        self.fg = args.fg or
+            beautiful.hotkeys_fg or beautiful.fg_normal
+        self.border_width = args.border_width or
+            beautiful.hotkeys_border_width or beautiful.border_width
+        self.border_color = args.border_color or
+            beautiful.hotkeys_border_color or self.fg
+        self.shape = args.shape or beautiful.hotkeys_shape
+        self.modifiers_fg = args.modifiers_fg or
+            beautiful.hotkeys_modifiers_fg or beautiful.bg_minimize or "#555555"
```
Elv13: beautiful.bg_minimize

This also looks like a random choice

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -357,21 +437,21 @@ function widget.new()
...
-            self.current_page = self.current_page - 1
-            self.wibox:set_widget(pages[self.current_page])
+        function widget_obj.page_prev(_self)
+            if _self.current_page == 1 then return end
+            _self.current_page = _self.current_page - 1
+            _self.wibox:set_widget(pages[_self.current_page])
         end
-        function widget_obj:show()
-            self.wibox.visible = true
+        function widget_obj.show(_self)
```
Elv13: Why the change? If it was about shadowing `self`, then you should consider moving these methods out of the constructor.

themes/default/theme.lua
```diff
@@ -39,6 +39,7 @@ theme.border_marked = "#91231c"
 -- tooltip_[font|opacity|fg_color|bg_color|border_width|border_color]
 -- mouse_finder_[color|timeout|animate_timeout|radius|factor]
 -- prompt_[fg|bg|fg_cursor|bg_cursor|font]
+-- hotkeys_[bg|fg|border_width|border_color|shape|opacity|modifiers_fg|label_fg|group_margin|font|description_font]
```
Elv13: Just to note that `opacity` doesn't seem to exist in this revision

Elv13:
Overall it's good. Only some minor comments here and there.

lib/awful/hotkeys_popup/widget.lua
actionless: actually i've documented it in the default theme.lua but forgot to add actually :)

lib/awful/hotkeys_popup/widget.lua
actionless: that value is usually looking the best with most of the themes i've tried

in beautiful we don't have anything like `theme.bg_disabled` and that `bg_minimize` is the most closer to represent some inactive/disabled info (or like in this case to make the labels with that color less accentiated)

lib/awful/hotkeys_popup/widget.lua
actionless: yup, because of that, i'll move them then

lib/awful/hotkeys_popup/widget.lua
actionless: for some reason those beautiful docstrings are displayed in apidoc for the hotkeys_popup.widget but not present in 06-appearance

could it be because they're missing `@param`?

lib/awful/hotkeys_popup/widget.lua
actionless: mm, i think in the place it's more clear

actionless:
is it fine now?

actionless:
@blueyed, @psychon any objections?

psychon:
I did not look too closely, but seems sensible. I trust Elv13 here.

PR #1540, opened 2017-02-08
awful.widget.watch: return timer too

copycat-killer:
It would be useful to be able to control the widget timer too. 

For instance, when you want to pause the watcher because you don't need it for some time.

If there is already an easy way to get every `gears.timer` connected to a widget, you can close this PR.

### DIFF
--- a/lib/awful/widget/watch.lua
+++ b/lib/awful/widget/watch.lua
@@ -62,7 +62,8 @@ local watch = { mt = {} }
 --
 -- @param[opt=wibox.widget.textbox()] base_widget Base widget.
 --
--- @return The widget used by this watch
+-- @return The widget used by this watch.
+-- @return Its gears.timer.
 function watch.new(command, timeout, callback, base_widget)
     timeout = timeout or 5
     base_widget = base_widget or textbox()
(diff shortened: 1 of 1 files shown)


lib/awful/widget/watch.lua
```diff
@@ -79,7 +80,7 @@ function watch.new(command, timeout, callback, base_widget)
     end)
     t:start()
     t:emit_signal("timeout")
-    return base_widget
+    return { widget = base_widget, timer = t }
```
psychon: This changes the API and thus will be blocked by @Elv13. How about `return base_widget, t` instead? (For the API doc: You would then just have multiple `@return`-tags, in the order in which the values are returned).
If this is a good idea is for others to decide...

copycat-killer: > If this is a good idea is for others to decide...

Starting from `vicious`, every widget library enables you to control your widget timers. I'll give you an example: if I have a network resources fetcher with a timeout of a couple of seconds and I don't want it to watch when I am away from my machine, I could disable it and spare resources. Now imagine if one has multiple watchers.

If you don't find it useful, I do. 

Plus, returning multiple arguments, if you don't want the timer, you can just skip it.

copycat-killer:
Updated the commit in order to not break the API.

If the user wants the timer, he can just grab it as a second argument.

Issue #1532, opened 2017-02-06
awful.spawn.easy_async Doesn't work properly


geripgeri:
Output of `awesome --version`:
```
awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1
```

**How to reproduce the issue:**
Before version 4.0 the command  `"TZ=Europe/London date +'%R'"` returns with correct data, after 4.0 there is no return, no stdout, stderr, reason, exit_code nothing. This also happens to other complex commands.

For example:

```
awful.spawn.easy_async("TZ=Europe/London date +'%R'", function(stdout, stderr, reason, exit_code)
    naughty.notify { text = stdout .. " , " .. stderr .. " , " .. reason .. " , " .. exit_code }
end)
```

**Actual result:**

```
nothing
```

**Expected result:**

```
20:38
```

Elv13:
You need a shell. Does:

```
awful.spawn.easy_async({'bash', '-c, "TZ=Europe/London date +'%R'"}, function(stdout, stderr, reason, exit_code)
    naughty.notify { text = stdout .. " , " .. stderr .. " , " .. reason .. " , " .. exit_code }
end)
```

helps?

geripgeri:
@Elv13 It's working! Thank you :)

Elv13:
@blueyed Should we (read: you 😄) add an `easy_async_with_shell`?

blueyed:
I think https://github.com/awesomeWM/awesome/pull/1001 is still related in this regard.

psychon:
(Oh and: No one checks the return value of anything! We should try not to add too many more of these)

psychon:
Even `awful.widget.watch` gets this wrong...

geripgeri:
Should I reopen this issue?

actionless:
i think we could either wrap command inside sh/bash -c as it was proposed already or we could check (probably via glib) what file is exists and executable

blueyed:
@geripgeri @Elv13 
`env TZ=Europe/London date +'%R'` should also work.

geripgeri:
@blueyed  Thanks with env it's working, but as I mentioned, I have other "complex" commands with wrong outputs, these also work with  @Elv13's workaround, but all of my commands were working before v4.0.

[...]

PR #1541, opened 2017-02-08
awful.spawn: add easy_async_with_shell

copycat-killer:
Related to: https://github.com/awesomeWM/awesome/issues/1532#issuecomment-278263856

### DIFF
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -321,14 +321,14 @@ end
 --- Asynchronously spawn a program and capture its output.
 -- (wraps `spawn.with_line_callback`).
 -- @tparam string|table cmd The command.
--- @tab callback Function with the following arguments
--- @tparam string callback.stdout Output on stdout.
--- @tparam string callback.stderr Output on stderr.
--- @tparam string callback.exitreason Exit Reason.
--- The reason can be "exit" or "signal".
--- @tparam integer callback.exitcode Exit code.
--- For "exit" reason it's the exit code.
--- For "signal" reason — the signal causing process termination.
+-- @function callback Function with the following arguments:
+--   @tparam string callback.stdout Output on stdout.
+--   @tparam string callback.stderr Output on stderr.
+--   @tparam string callback.exitreason Exit Reason.
+--   @tparam integer callback.exitcode Exit code.
+--   The exitreason argument can either be "exit" or "signal".
+--   For "exit" reason it's the exit code.
+--   For "signal" reason it's the signal causing process termination.
 -- @treturn[1] Integer the PID of the forked process.
 -- @treturn[2] string Error message.
 -- @see spawn.with_line_callback
@@ -368,6 +368,24 @@ function spawn.easy_async(cmd, callback)
         exit=exit_callback,
         output_done=output_done_callback
     })
+end
+
+--- Call spawn.easy_async with a shell.
+-- This calls `cmd` with `$SHELL -c` (via `awful.util.shell`).
+-- @tparam string|table cmd The command.
+-- @function callback Function with the following arguments:
+--   @tparam string callback.stdout Output on stdout.
+--   @tparam string callback.stderr Output on stderr.
+--   @tparam string callback.exitreason Exit Reason.
+--   @tparam integer callback.exitcode Exit code.
+--   The exitreason argument can either be "exit" or "signal".
+--   For "exit" reason it's the exit code.
+--   For "signal" reason it's the signal causing process termination.
+-- @treturn[1] Integer the PID of the forked process.
+-- @treturn[2] string Error message.
+-- @see spawn.with_line_callback
+function spawn.easy_async_with_shell(cmd, callback)
+    return spawn.easy_async({ util.shell, "-c", cmd or "" }, callback)
 end
 
 --- Read lines from a Gio input stream
--- a/tests/test-spawn.lua
+++ b/tests/test-spawn.lua
@@ -4,6 +4,7 @@ local runner = require("_runner")
 local spawn = require("awful.spawn")
 
 local spawns_done = 0
+local async_spawns_done = 0
 local exit_yay, exit_snd = nil, nil
 
 -- * Using spawn with array is already covered by the test client.
@@ -40,6 +41,16 @@ local steps = {
 
     function(count)
         if count == 1 then
+            spawn.easy_async("echo yay", function(stdout)
+                if stdout:match("yay") then
+                    async_spawns_done = async_spawns_done + 1
+                end
+            end)
+            spawn.easy_async_with_shell("true && echo yay", function(stdout)
+                if stdout:match("yay") then
+                    async_spawns_done = async_spawns_done + 1
+                end
+            end)
             local steps_yay = 0
             spawn.with_line_callback("echo yay", {
                                      stdout = function(line)
@@ -90,9 +101,10 @@ local steps = {
         if spawns_done == 2 then
             assert(exit_yay == 0)
             assert(exit_snd == 42)
+            assert(async_spawns_done == 2)
             return true
         end
-    end,
+    end
 }
 
 runner.run_steps(steps)


lib/awful/spawn.lua
```diff
@@ -370,6 +370,18 @@ function spawn.easy_async(cmd, callback)
     })
 end
 
+--- Call spawn.easy_async with a shell.
+-- This calls `cmd` with `$SHELL -c` (via `awful.util.shell`).
+-- @tparam string cmd The command.
+-- @tab callback Callback function.
```
Elv13: This is wrong, it's a function. `@tab` is for table. Also, the callback parameters are missing. Please copy/paste the parameters from `easy_async`.

copycat-killer: `@tab` is what is currently defined for `spawn.easy_async`. Guess we have to fix that then.

blueyed: See https://github.com/awesomeWM/awesome/pull/1001 for improving easy_async by itself first.

copycat-killer:
Fixed code docs. For `callback` arguments, I wanted to put a reference to `easy_async` instead of copy/pasting, since `easy_async_with_shell` should show up right after it in the documentations. 

But I'll copy/paste, if you prefer.

lib/awful/spawn.lua
copycat-killer: Even if related, `easy_async_with_shell` shouldn't affect or be affected by #1001, I believe. I'll give it a look, anyway.

lib/awful/spawn.lua
```diff
@@ -370,6 +370,18 @@ function spawn.easy_async(cmd, callback)
     })
 end
 
+--- Call spawn.easy_async with a shell.
+-- This calls `cmd` with `$SHELL -c` (via `awful.util.shell`).
+-- @tparam string|table cmd The command.
+-- @function callback The callback function.
```
Elv13: You forgot to copy/paste the arguments here. Sorry about being picky, it's just that correct/consistent doc is something we try to take seriously now.

lib/awful/spawn.lua
```diff
@@ -370,6 +370,18 @@ function spawn.easy_async(cmd, callback)
...
+-- @see spawn.easy_async for details about `callback`.
+function spawn.easy_async_with_shell(cmd, callback)
+    if cmd and cmd ~= "" then
+        cmd = { util.shell, "-c", cmd }
+        return spawn.easy_async(cmd, callback)
+    end
+end
```
Elv13: I would remove the if and use `return spawn.easy_async(cmd or "", callback)`. The rational is that the return values should be consistent with `easy_async`.

copycat-killer: Done.

lib/awful/spawn.lua
```diff
@@ -370,6 +370,25 @@ function spawn.easy_async(cmd, callback)
...
+--   @tparam string callback.exitreason Exit Reason.
+--   @tparam integer callback.exitcode Exit code.
+--   The exitreason argument can either be "exit" or "signal".
+--   For "exit" reason it's the exit code.
+--   For "signal" reason it's the signal causing process termination.
+-- @treturn[1] Integer the PID of the forked process.
+-- @treturn[2] string Error message.
+-- @see spawn.with_line_callback
+function spawn.easy_async_with_shell(cmd, callback)
+    local cmd = { util.shell, "-c", cmd or "" }
```
actionless: you should remove `local`:
https://travis-ci.org/awesomeWM/awesome/jobs/199737866#L3492

blueyed: @actionless 
That seems to be from recent updates to master?!

I've rebased this PR now.

blueyed: Whoops.. I thought it was for my easier-async branch/PR..  ignore it.

copycat-killer: @actionless Done.

copycat-killer:
I don't understand codecov: does this need a test?

Elv13:
> I don't understand codecov: does this need a test?

In a perfect world, yes. The first codecov row is the relative coverage, the number of code lines front he patch covered by tests. the second one is the absolute coverage. We usually try to get code with a higher coverage than the current absolute coverage percentage (otherwise it will go down). Feel free to add a `test-awful-spawn.lua` integration suit and try a fast command. The roadblock is obviously that testing network related sequences or async IO is very hard. On one hand a shim/mock unit test wont prove or disprove that the code work, so it's useless and an integration test is hard to write.

If you want to do it, then it's welcome (and we would have had to do it anyway), but I can understand if you think it's too much to ask. This is why neither of us commented on that so far.

The integration test suits are executed in a "real" Awesome instance. They have multiple steps. If the step return `true`, then the next is executed. If it returns `false`, the test fail. If it returns `nil`, then the step is executed again 100ms later. It can only return `nil` 5 times (500ms). After that it is considered a failure. So pick a command like `MARVAR=foo echo $MYVAR` that is almost guaranteed to take less than 0.5s.

copycat-killer:
Thanks for your explanation.

Little by little I'm understanding how things work here, the process is fun and also educative.

I'll try to do a proper coverage.

Elv13:
@copycat-killer Any news on this?

copycat-killer:
@Elv13 No, I still have to start working on this.

copycat-killer:
@Elv13 What do you think about something like this?

```diff
diff --git a/tests/test-spawn.lua b/tests/test-spawn.lua
index a859214f..74f5fa08 100644
--- a/tests/test-spawn.lua
+++ b/tests/test-spawn.lua
@@ -86,8 +86,18 @@ local steps = {
                                          exit_snd = code
                                      end
                                  })
+            spawn.easy_async("ping -c1 localhost", function(stdout, stderr, reason, exit_code)
+                if stdout:match("0%% packet loss") then
+                    spawns_done = spawns_done + 1
+                end
+            end)
+            spawn.easy_async_with_shell("ping -c1 localhost", function(stdout, stderr, reason, exit_code)
+                if stdout:match("0%% packet loss") then
+                    spawns_done = spawns_done + 1
+                end
+            end)
         end
-        if spawns_done == 2 then
+        if spawns_done == 4 then
             assert(exit_yay == 0)
             assert(exit_snd == 42)
             return true
```

actionless:
i think it will be better to have in a test for shell variant of this function some more shell-specific command, for example iteration with printing, and/or some pipes usage

copycat-killer:
What about putting the code in `test-spawn.lua`? Is it OK, or should I write a dedicate test?

actionless:
looks fine to me

Elv13:
@copycat-killer If you plan to do it, please do it ASAP. v4.1 is imminent. Otherwise I can merge it as-is, but obviously we try hard to avoid adding features without regression tests.

copycat-killer:
@actionless @Elv13 would this be OK as test?

```diff
diff --git a/tests/test-spawn.lua b/tests/test-spawn.lua
index a859214f..74f5fa08 100644
--- a/tests/test-spawn.lua
+++ b/tests/test-spawn.lua
@@ -86,8 +86,18 @@ local steps = {
                                          exit_snd = code
                                      end
                                  })
+            spawn.easy_async("ping -c1 localhost", function(stdout, stderr, reason, exit_code)
+                if stdout:match("0%% packet loss") then
+                    spawns_done = spawns_done + 1
+                end
+            end)
+            spawn.easy_async_with_shell("ping -c1 localhost | grep 'packet loss' | cut -d' ' -f6", function(stdout, stderr, reason, exit_code)
+                if stdout == "0%" then
+                    spawns_done = spawns_done + 1
+                end
+            end)
         end
-        if spawns_done == 2 then
+        if spawns_done == 4 then
             assert(exit_yay == 0)
             assert(exit_snd == 42)
             return true
```

actionless:
i think you should do separate spawn counters -- otherwise the test could return falsepositive result

copycat-killer:
Done.

tests/test-spawn.lua
```diff
@@ -86,13 +87,24 @@ local steps = {
...
+                    async_spawns_done = async_spawns_done + 1
+                end
+            end)
+            spawn.easy_async_with_shell("ping -c1 localhost | grep 'packet loss' | cut -d' ' -f6", function(stdout, stderr, reason, exit_code)
+                if stdout:match("0%%") then
+                    async_spawns_done = async_spawns_done + 1
+                end
+            end)
         end
         if spawns_done == 2 then
```
actionless: If spawns_done >= 2 and async_spawns_done >= 2

copycat-killer: Because it's asynchronous, I preferred to put is as `assert` below.

actionless: because it's asynchronous you can't guarantee what async_spawn counter will always reach exactly 2 in the same time when spawn counter did that

actionless: i think the current test results (4/5 pass) are caused by exactly this problem

https://travis-ci.org/awesomeWM/awesome/builds/211464746

Elv13: Also, `ping` is not a "base" command. `echo` is. In some sandboxes runners, `ping` wont be available.

copycat-killer: > you can't guarantee what async_spawn counter will always reach exactly 2 in the same time when spawn counter did that

Indeed that was my reason for putting it below.

Travis is facing a timeout, but it doesn't happen with my machine.

How to fix that? Probably we need an ad hoc test.

Elv13: > Is there a "base" command for local networking check?

Ping is probably installed, but don't count on it. Never write unit tests that use the network, this is, in general, a very bad practice. A test should be as atomic as possible. `echo` is as atomic a command will get without having to prepend our own directory to $PATH and use our own test commands (which would be the best solution, but takes longer to implement).

copycat-killer: OK, switched to `echo`.

Elv13:
@copycat-killer ping

copycat-killer:
@Elv13 34cf619 is OK to me.

blueyed:
Thank you, @copycat-killer.

You might want to review / comment on https://github.com/awesomeWM/awesome/pull/1001, since you seem to like easy-async.. ;)

copycat-killer:
I'll give it a look, ASAP.

PR #1544, opened 2017-02-08
Let env find bash

qbit:
> Applications should note that the standard PATH to the shell cannot be assumed to be either /bin/sh or /usr/bin/sh, and should be determined by interrogation of the PATH returned by getconf PATH, ensuring that the returned pathname is an absolute pathname and not a shell built-in.

POSIX and SUS both say the same thing about this. This will also fix awesome-client on *BSD.


blueyed:
Thanks!

PR #1550, opened 2017-02-10
Add wibox grid layout (gtk style)

getzze:
Using ideas form #1120 and how the grid layout works in qt and gtk.
Widgets can span several columns/rows. They can be added automatically (depending of the `orientation` property: "y" from left to right, "x" from up to down).
One can define the `spacing` between columns/rows and the `padding`, which is the gap with the outer limit. There are two options, `expand` so the grid extends to all the available space and `fixed_size` which defines if the columns/rows all have the same size.
I have a test file to test dynamical resizing of the widget with the mouse, but I didn't test dynamical removal/addition of a widget (only static).
I use it for a calendar widget (like blingbling had).

### DIFF
--- a/lib/wibox/layout/grid.lua
+++ b/lib/wibox/layout/grid.lua
@@ -1,0 +1,896 @@
+---------------------------------------------------------------------------
+--- A grid layout.
+--
+-- Widgets spanning several columns or rows cannot be included using the declarative system.
+-- Instead, create the grid layout and call the `add_widget_at` method.
+--
+--@DOC_wibox_layout_grid_imperative_EXAMPLE@
+--
+-- Using the declarative system, widgets are automatically added next to each other spanning only one cell.
+--
+--@DOC_wibox_layout_defaults_grid_EXAMPLE@
+-- @author getzze
+-- @copyright 2017 getzze
+-- @classmod wibox.layout.grid
+---------------------------------------------------------------------------
+
+local setmetatable = setmetatable
+local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
+local table = table
+local pairs = pairs
+local ipairs = ipairs
+local math = math
+local util = require("awful.util")
+local base = require("wibox.widget.base")
+
+local grid = { mt = {} }
+
+local properties = {
+                    "orientation", "superpose",
+                    "forced_num_rows", "forced_num_cols",
+                    "min_cols_size", "min_rows_size",
+                   }
+
+local dir_properties = { "spacing", "homogeneous", "expand" }
+
+
+
+--- Set the preferred orientation of the grid layout.
+--
+-- Allowed values are "horizontal" and "vertical".
+-- When calling `get_next_empty`, empty cells are browsed differently:
+--
+-- * for "horizontal", the grid can be extended horizontally. The current column is filled first; if no empty cell is found up to `forced_num_rows`, the next column is filled, creating it if it does not exist.
+--
+-- * for "vertical", the grid can be extended vertically. The current row is filled first; if no empty cell is found up to `forced_num_cols`, the next row is filled, creating it if it does not exist.
+--
+--@DOC_wibox_layout_grid_orientation_EXAMPLE@
+-- @param[opt="vertical"] string Preferred orientation: "horizontal" or "vertical".
+-- @property orientation
+
+
+--- Allow to superpose widgets in the same cell.
+-- If false, check before adding a new widget if it will superpose with another widget and prevent from adding it.
+--
+--@DOC_wibox_layout_grid_superpose_EXAMPLE@
+-- @param[opt=false] boolean
+-- @property superpose
+
+
+--- Force the number of rows of the layout.
+-- @property forced_num_rows
+-- @tparam[opt=nil] number|nil number Forced number of rows (`nil` for automatic).
+
+--- Force the number of columns of the layout.
+-- @property forced_num_cols
+-- @tparam[opt=nil] number|nil number Forced number of columns (`nil` for automatic).
+
+--- Set the minimum size for the columns.
+--
+--@DOC_wibox_layout_grid_min_size_EXAMPLE@
+-- @param[opt=0] number Minimum size of the columns.
+-- @property min_cols_size
+
+--- Set the minimum size for the rows.
+-- @param[opt=0] number Minimum size of the rows.
+-- @property min_rows_size
+
+
+--- The spacing between columns.
+--
+-- see `spacing`
+--
+-- @param[opt=0] number The spacing
+-- @property horizontal_spacing
+
+--- The spacing between rows.
+--
+-- see `spacing`
+--
+-- @param[opt=0] number The spacing
+-- @property vertical_spacing
+
+--- The spacing between rows and columns.
+-- Set both `horizontal_spacing` and `vertical_spacing` to the same value.
+-- Get the value `horizontal_spacing` or `vertical_spacing` defined by the preferred `orientation`.
+--
+--@DOC_wibox_layout_grid_spacing_EXAMPLE@
+-- @param[opt=0] number The spacing
+-- @property spacing
+
+
+--- Controls if the columns are expanded to use all the available width.
+--
+-- see `expand`
+--
+-- @param[opt=false] boolean Expand the grid into the available space
+-- @property horizontal_expand
+
+--- Controls if the rows are expanded to use all the available height.
+--
+-- see `expand`
+--
+-- @param[opt=false] boolean Expand the grid into the available space
+-- @property vertical_expand
+
+--- Controls if the columns/rows are expanded to use all the available space.
+-- Set both `horizontal_expand` and `vertical_expand` to the same value.
+-- Get the value `horizontal_expand` or `vertical_expand` defined by the preferred `orientation`.
+--
+--@DOC_wibox_layout_grid_expand_EXAMPLE@
+-- @param[opt=false] boolean Expand the grid into the available space
+-- @property expand
+
+--- Controls if the columns all have the same width or if the width of each column depends on the content.
+--
+-- see `homogeneous`
+--
+-- @param[opt=true] boolean All the columns have the same width.
+-- @property horizontal_homogeneous
+
+--- Controls if the rows all have the same height or if the height of each row depends on the content.
+--
+-- see `homogeneous`
+--
+-- @param[opt=true] boolean All the rows have the same height.
+-- @property vertical_homogeneous
+
+--- Controls if the columns/rows all have the same size or if the size depends on the content.
+-- Set both `horizontal_homogeneous` and `vertical_homogeneous` to the same value.
+-- Get the value `horizontal_homogeneous` or `vertical_homogeneous` defined by the preferred `orientation`.
+--
+--@DOC_wibox_layout_grid_expand_EXAMPLE@
+-- @param[opt=true] boolean All the columns/rows have the same size.
+-- @property homogeneous
+
+--- Child widget position. Return of `get_widget_position`.
+-- @field row Top row index
+-- @field col Left column index
+-- @field row_span Number of rows to span
+-- @field col_span Number of columns to span
+-- @table position
+
+
+-- Return the maximum value of a table.
+local function max_value(t)
+    local m = 0
+    for _,v in ipairs(t) do
+        if m < v then m = v end
+    end
+    return m
+end
+
+-- Return the sum of the values in the table.
+local function sum_values(t)
+    local m = 0
+    for _,v in ipairs(t) do
+        m = m + v
+    end
+    return m
+end
+
+
+-- Find a widget in a widget_table, by matching the coordinates.
+-- Using the  `row`:`col` coordinates, and the spans `row_span` and `col_span`
+-- @tparam table widgets_table Table of the widgets present in the grid
+-- @tparam number row Row number for the top left corner of the widget
+-- @tparam number col Column number for the top left corner of the widget
+-- @tparam number row_span The number of rows the widget spans (default to 1)
+-- @tparam number col_span The number of columns the widget spans (default to 1)
+-- @treturn table Table of index of widget_table
+local function find_widgets_at(widgets_table, row, col, row_span, col_span)
+    if not row or row < 1 or not col or col < 1 then return nil end
+    row_span = (row_span and row_span > 0) and row_span or 1
+    col_span = (col_span and col_span > 0) and col_span or 1
+    local ret = {}
+    for index, data in ipairs(widgets_table) do
+        -- If one rectangular widget is on left side of other
+        local test_horizontal = not (row > data.row + data.row_span - 1 or data.row > row + row_span - 1)
+        -- If one rectangular widget is above other
+        local test_vertical   = not (col > data.col + data.col_span - 1 or data.col > col + col_span - 1)
+        if test_horizontal and test_vertical then
+            table.insert(ret, index)
+        end
+    end
+    -- reverse sort for safe removal of indices
+    table.sort(ret, function(a,b) return a>b end)
+    return #ret > 0 and ret or nil
+end
+
+
+-- Find a widget in a widget_table, by matching the object.
+-- @tparam table widgets_table Table of the widgets present in the grid
+-- @param widget The widget to find
+-- @treturn number|nil The index of the widget in widget_table, `nil` if not found
+local function find_widget(widgets_table, widget)
+    for index, data in ipairs(widgets_table) do
+        if data.widget == widget then
+            return index
+        end
+    end
+    return nil
+end
+
+--- Get the number of rows and columns occupied by the widgets in the grid.
+-- @treturn number,number The number of rows and columns
+function grid:get_dimension()
+    return self._private.num_rows, self._private.num_cols
+end
+
+-- Update the number of rows and columns occupied by the widgets in the grid.
+local function update_dimension(self)
+    local num_rows, num_cols = 0, 0
+    if self._private.forced_num_rows then
+        num_rows = self._private.forced_num_rows
+    end
+    if self._private.forced_num_cols then
+        num_cols = self._private.forced_num_cols
+    end
+
+    for _, data in ipairs(self._private.widgets) do
+        num_rows = math.max(num_rows, data.row + data.row_span - 1)
+        num_cols = math.max(num_cols, data.col + data.col_span - 1)
+    end
+    self._private.num_rows = num_rows
+    self._private.num_cols = num_cols
+end
+
+
+--- Find the next available cell to insert a widget.
+-- The grid is browsed according to the `orientation`.
+-- @tparam[opt=1] number hint_row The row coordinate of the last occupied cell.
+-- @tparam[opt=1] number hint_column The column coordinate of the last occupied cell.
+-- @return number,number The row,column coordinate of the next empty cell
+function grid:get_next_empty(hint_row, hint_column)
+    local row    = (hint_row and hint_row > 0) and hint_row or 1
+    local column = (hint_column and hint_column > 0) and hint_column or 1
+
+    local next_field
+    if self._private.orientation == "vertical" then
+        next_field = function(x, y)
+            if y < self._private.num_cols then
+                return x, y+1
+            end
+            return x+1,1
+        end
+    elseif self._private.orientation == "horizontal" then
+        next_field = function(x, y)
+            if x < self._private.num_rows then
+                return x+1, y
+            end
+            return 1,y+1
+        end
+    end
+    while true do
+        if find_widgets_at(self._private.widgets, row, column, 1, 1) == nil then return row, column end
+        row, column = next_field(row, column)
+    end
+end
+
+
+--- Add some widgets to the given grid layout.
+-- The widgets are assumed to span one cell
+-- @param ... Widgets that should be added (must at least be one)
+function grid:add(...)
+    local args = { n=select('#', ...), ... }
+    assert(args.n > 0, "need at least one widget to add")
+    local row, column
+    for i=1, args.n do
+        local w = args[i]
+        -- Get the next empty coordinate to insert the widget
+        row, column = self:get_next_empty(row, column)
+        self:add_widget_at(w, row, column, 1, 1)
+    end
+end
+
+--- Add a widget to the grid layout at specific coordinate.
+--
+--@DOC_wibox_layout_grid_add_EXAMPLE@
+-- @param child Widget that should be added
+-- @tparam number row Row number for the top left corner of the widget
+-- @tparam number col Column number for the top left corner of the widget
+-- @tparam[opt=1] number row_span The number of rows the widget spans.
+-- @tparam[opt=1] number col_span The number of columns the widget spans.
+-- @treturn boolean index If the operation is successful
+function grid:add_widget_at(child, row, col, row_span, col_span)
+    if not row or row < 1 or not col or col < 1 then return false end
+    row_span = (row_span and row_span > 0) and row_span or 1
+    col_span = (col_span and col_span > 0) and col_span or 1
+
+    -- check if the object is a widget
+    base.check_widget(child)
+
+    -- test if the new widget superpose with existing ones
+    local superpose = find_widgets_at(self._private.widgets, row, col, row_span, col_span)
+    if not self._private.superpose and superpose then
+        return false
+    end
+
+    -- Add grid information attached to the widget
+    local child_data = {
+        widget = child,
+        row = row,
+        col = col,
+        row_span = row_span,
+        col_span = col_span
+    }
+    table.insert(self._private.widgets, child_data)
+
+    -- Update the row and column numbers
+    self._private.num_rows = math.max(self._private.num_rows, row + row_span - 1)
+    self._private.num_cols = math.max(self._private.num_cols, col + col_span - 1)
+
+    self:emit_signal("widget::layout_changed")
+    self:emit_signal("widget::redraw_needed")
+    return true
+end
+
+--- Remove one or more widgets from the layout.
+-- @param ... Widgets that should be removed (must at least be one)
+-- @treturn boolean If the operation is successful
+function grid:remove(...)
+    local args = { ... }
+    local ret = false
+    for _, rem_widget in ipairs(args) do
+        local index = find_widget(self._private.widgets, rem_widget)
+        if index ~= nil then
+            table.remove(self._private.widgets, index)
+            ret = true
+        end
+    end
+    if ret then
+        -- Recalculate num_rows and num_cols
+        update_dimension(self)
+        self:emit_signal("widget::layout_changed")
+        self:emit_signal("widget::redraw_needed")
+    end
+    return ret
+end
+
+
+--- Remove widgets at the coordinates.
+--
+--@DOC_wibox_layout_grid_remove_EXAMPLE@
+-- @tparam number row The row coordinate of the widget to remove
+-- @tparam number col The column coordinate of the widget to remove
+-- @tparam[opt=1] number row_span The number of rows the area to remove spans.
+-- @tparam[opt=1] number col_span The number of columns the area to remove spans.
+-- @treturn boolean If the operation is successful (widgets found)
+function grid:remove_widgets_at(row, col, row_span, col_span)
+    local widget_indices = find_widgets_at(self._private.widgets, row, col, row_span, col_span)
+    if widget_indices == nil then return false end
+
+    for _,index in ipairs(widget_indices) do
+        table.remove(self._private.widgets, index)
+    end
+    -- Recalculate num_rows and num_cols
+    update_dimension(self)
+    self:emit_signal("widget::layout_changed")
+    self:emit_signal("widget::redraw_needed")
+    return true
+end
+
+--- Return the coordinates of the widget.
+-- @param widget The widget
+-- @treturn table The `position` table of the coordinates in the grid, with fields `row`, `col`, `row_span` and `col_span`.
+function grid:get_widget_position(widget)
+    local index = find_widget(self._private.widgets, widget)
+    if index == nil then return nil end
+    local data = self._private.widgets[index]
+    local ret = {}
+    ret["row"] = data.row
+    ret["col"] = data.col
+    ret["row_span"] = data.row_span
+    ret["col_span"] = data.col_span
+    return ret
+end
+
+
+--- Return the widgets at the coordinates.
+-- @tparam number row The row coordinate of the widget
+-- @tparam number col The column coordinate of the widget
+-- @tparam[opt=1] number row_span The number of rows to span.
+-- @tparam[opt=1] number col_span The number of columns to span.
+-- @treturn table The widget(s) found at the specific coordinates, nil if no widgets found
+function grid:get_widgets_at(row, col, row_span, col_span)
+    local widget_indices = find_widgets_at(self._private.widgets, row, col, row_span, col_span)
+    if widget_indices == nil then return nil end
+
+    local ret = {}
+    for _,index in ipairs(widget_indices) do
+        local data = self._private.widgets[index]
+        table.insert(ret, data.widget)
+    end
+
+    return #ret > 0 and ret or nil
+end
+
+--- Replace old widget by new widget, spanning the same columns and rows.
+-- @param old The widget to remove
+-- @param new The widget to add
+-- @treturn boolean If the operation is successful (widget found)
+function grid:replace_widget(old, new)
+    -- check if the new object is a widget
+    local status = pcall(function () base.check_widget(new) end)
+    if not status then return false end
+    -- find the old widget
+    local index = find_widget(self._private.widgets, old)
+    if index == nil then return false end
+
+    -- get old widget position
+    local data = self._private.widgets[index]
+    local row, col, row_span, col_span = data.row, data.col, data.row_span, data.col_span
+
+    table.remove(self._private.widgets, index)
+    return self:add_widget_at(new, row, col, row_span, col_span)
+end
+
+-- Update position of the widgets when inserting, adding or removing a row or a column.
+-- @tparam table table_widgets Table of widgets
+-- @tparam string orientation Orientation of the line: "horizontal" -> column, "vertical" -> row
+-- @tparam number index Index of the line
+-- @tparam string mode insert, extend or remove
+-- @tparam boolean after Add the line after the index instead of inserting it before.
+-- @tparam boolean extend Extend the line at index instead of inserting an empty line.
+local function update_widgets_position(table_widgets, orientation, index, mode)
+    local t = orientation == "horizontal" and "col" or "row"
+    local to_remove = {}
+    -- inc   : Index increment or decrement
+    -- first : Offset index for top-left cell of the widgets to shift
+    -- last  : Offset index for bottom-right cell of the widgets to resize
+    local inc, first, last
+    if mode == "remove" then
+        inc, first, last = -1, 1, 1
+    elseif mode == "insert" then
+        inc, first, last = 1, 0, 0
+    elseif mode == "extend" then
+        inc, first, last = 1, 1, 0
+    else
+        return
+    end
+    for i, data in ipairs(table_widgets) do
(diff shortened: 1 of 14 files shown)


actionless:
i think it will be better to implement that layout on top of new API which will be brought by this PR:
https://github.com/awesomeWM/awesome/pull/644

getzze:
It's not an awful layout, but a wibox layout

actionless:
good then, need to try it later on, mb it will allow to simplify code of the hotkeys widget

Elv13:
Thanks for this. I will review this at some point. A few things I can see right away that are easy to fix is the doc examples. It should be `@DOC_path_to_test_EXAMPLE@`. When you have ldoc installed, it will auto replace the macro with the example code an the image. That way we are sure the examples are in sync. Lines with `--DOC_HIDE` are removed and if `--DOC_HIDE_ALL` is present, then only the image is added. Another one is that the first line of the doc block need a "." at the end. This is what `ldoc` needs to generate the table of content properly.

psychon:
This new module, `wibox.layout.grid`, does `require("awful")`. This leads to the following dependencies (only one example, there are also other loops):

`wibox.layout.grid` -> `awful` -> `awful.wibar` -> `wibox` -> `wibox.layout` -> `wibox.layout.grid`

This leads to `require` going into an endless loop.

Since this actually only uses `awful.util`, you can "fix" this problem via `local util = require("awful.util")` instead of requiring all of `awful`.

(I did not look at more of this than the above)

getzze:
I fixed these and updated the documentation of the file.
Also fixed a problem that the min_[cols|rows]_size was not correctly updated

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
+---------------------------------------------------------------------------
```
psychon: Commit message: Whoops? ;-)

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+local base = require("wibox.widget.base")
+local beautiful = require("beautiful")
+
+
+local grid = { mt = {} }
+
+local properties = {
+                    "orientation", "forced_num_rows", "forced_num_cols",
+                    "min_cols_size", "min_rows_size",
+                    --"max_cols_size", "max_rows_size",
```
psychon: What's up with `max_cols_size` and `max_rows_size`?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+    ret._private.num_cols = 0
+    ret._private.rows_size = {}
+    ret._private.cols_size = {}
+
+    ret._private.forced_num_rows = nil
+    ret._private.forced_num_cols = nil
+    ret:_set_grid_default("min_rows_size", 0)
+    ret:_set_grid_default("min_cols_size", 0)
+    --ret:_set_grid_default("max_rows_size", 0)
+    --ret:_set_grid_default("max_cols_size", 0)
```
psychon: What's up with `max_cols_size` and `max_rows_size`?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+local properties = {
+                    "orientation", "forced_num_rows", "forced_num_cols",
+                    "min_cols_size", "min_rows_size",
+                    --"max_cols_size", "max_rows_size",
+                   }
+
+local dir_properties = { "spacing", "padding", "fixed_size", "expand" }
+
+
+--- Reset a ratio layout. This removes all widgets from the layout.
```
psychon: "ratio layout"? Sounds like a copy&paste.

psychon: Also, to me this sounds like it just removes all widgets. How about `Among other things, this removes all widgets from the layout`?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+-- @class function
+
+
+--- The preferred orientation for the grid.
+-- @beautiful beautiful.grid_orientation
+
+--- The spacing between cells.
+-- @beautiful beautiful.grid_[horizontal|vertical]_spacing
+
+--- The padding with outer limits.
```
psychon: This sounds weird to me. Is it just me or do others agree?
I would propose `The space between the grid and the surrounding area`?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+-- @beautiful beautiful.grid_[horizontal|vertical]_padding
+
+--- Controls if the columns/rows are expanded to fit all the available area.
```
psychon: `space` instead of `area`? I'm not sure (also: `fit`? How about `cover` or `use`?)

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+-- @beautiful beautiful.grid_[horizontal|vertical]_expand
+
+--- Controls if the columns/rows have the same size.
+-- @beautiful beautiful.grid_[horizontal|vertical]_fixed_size
```
psychon: I don't understand... so, this is a boolean and if it is true, all of them have the same size. And if it is false?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+
+
+--- The spacing between cells.
+-- If direction is not specified, set both directions and get the preferred orientation
```
psychon: Which direction?!?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+-- @param number row Row number for the top left corner of the widget
+-- @param number col Column number for the top left corner of the widget
+-- @param number row_span The number of rows the widget spans (default to 1)
+-- @param number col_span The number of columns the widget spans (default to 1)
+-- @treturn boolean index If the operation is successful
+local function matrix_add(matrix, object, row, col, row_span, col_span)
+    if row < 1 then return false end
+    if col < 1 then return false end
+    row_span = row_span > 0 and row_span or 1
+    col_span = col_span > 0 and col_span or 1
```
psychon: I would leave these two lines out. The only caller already makes sure that these "span" values are sane.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+-- @param matrix Matrix containing the widgets
+-- @param widget Widget to add to the matrix
+-- @param number row Row number for the top left corner of the widget
+-- @param number col Column number for the top left corner of the widget
+-- @param number row_span The number of rows the widget spans (default to 1)
+-- @param number col_span The number of columns the widget spans (default to 1)
+-- @treturn boolean index If the operation is successful
+local function matrix_add(matrix, object, row, col, row_span, col_span)
+    if row < 1 then return false end
+    if col < 1 then return false end
```
psychon: Doesn't this check rather belong into the only caller instead? (`add_widget_at`)

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+--- Add a widget to the matrix
+-- the widget is set in coordinates `row`:`col`, spanning several cells
+-- given by `row_span` and `col_span`
+-- @param matrix Matrix containing the widgets
+-- @param widget Widget to add to the matrix
+-- @param number row Row number for the top left corner of the widget
+-- @param number col Column number for the top left corner of the widget
+-- @param number row_span The number of rows the widget spans (default to 1)
+-- @param number col_span The number of columns the widget spans (default to 1)
+-- @treturn boolean index If the operation is successful
```
psychon: The only caller doesn't check the return value, so the return value can as well be removed. Also, you have a `@param widget`, but the actual parameter's name is `object`. (LDoc ignores documentation for non-public functions, I guess)

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+local function matrix_remove(matrix, row, col, row_span, col_span)
+    if row < 1 then return false end
+    if col < 1 then return false end
+    row_span = row_span > 0 and row_span or 1
+    col_span = col_span > 0 and col_span or 1
```
psychon: Same comments as to `matrix_add`

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+-- @treturn table of index of widget_table
+local function find_widgets_by_coordinates(widgets_table, row, col, orig_row_span, orig_col_span)
+    local row_span = (orig_row_span and orig_row_span > 0) and orig_row_span or 1
+    local col_span = (orig_col_span and orig_col_span > 0) and orig_col_span or 1
+    local ret = {}
+    for index, data in ipairs(widgets_table) do
+        -- If one rectangular widget is on left side of other
+        local test_horizontal = not (row > data.row + data.row_span - 1 or data.row > row + row_span - 1)
+        -- If one rectangular widget is above other
+        local test_vertical   = not (col > data.col + data.col_span - 1 or data.col > col + col_span - 1)
```
psychon: These conditions are giving me a headache...
Since the only caller only wants to check "is there any overlap with existing widgets", how about looping over the matrix instead and checking if all the needed entries do not exist? That sounds a lot simpler to me.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+    end
+
+    for _, data in ipairs(self._private.widgets) do
+        num_rows = math.max(num_rows, data.row + data.row_span - 1)
+        num_cols = math.max(num_cols, data.col + data.col_span - 1)
+    end
+    self._private.num_rows = num_rows
+    self._private.num_cols = num_cols
+    return num_rows, num_cols
+end
```
psychon: Since this is just a getter, how about:
```Lua
function grid:get_shape() -- By the, the name is really bad
    return self._private.num_rows, self._private.num_cols
end

local function update_shape(self)
     All of the above code; This function would then be called to update these variables
end
```

psychon: How about `get_dimension` instead of `get_shape`?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+    local rspan = (row_span and row_span > 0) and row_span or 1
+    local cspan = (col_span and col_span > 0) and col_span or 1
+
+    -- check if the object is a widget
+    base.check_widget(child)
+
+    -- test if the new widget superpose with existing ones
+    local superpose = find_widgets_by_coordinates(self._private.widgets, row, col, rspan, cspan)
+    if superpose then
+        -- TODO: add message, remove other widgets?
```
psychon: This TODO should be handled somehow. Personally, I'm fine with rejecting the addition.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+    table.insert(self._private.widgets, child_data)
+    matrix_add(self._private.matrix, child_data, row, col, rspan, cspan)
+
+    -- Update the row and column numbers
+    if self._private.num_rows < row + rspan - 1 then
+        self._private.num_rows = row + rspan - 1
+    end
+    if self._private.num_cols < col + cspan - 1 then
+        self._private.num_cols = col + cspan - 1
+    end
```
psychon: How about `self._private.num_rows = math.max(self._private.num_rows, row + rspan - 1)`?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+        end
+    -- Add new widget to next row. Create new column if needed
+    elseif self._private.orientation == "x" then
+        -- Scan cols until the end of time
+        local current_row = row
+        if type(matrix[current_row]) ~= "table" then
+            matrix[current_row] = {}
+        end
+        if next(matrix[current_row]) == nil then
+            return current_row, column
```
psychon: I do not really like the code duplication between `x` and `y`. What I actually want to say: The other case has another outer `for` loop and returns `i,1` in this case. Why the (completely?) different behaviour here?

psychon: How about something like this instead:
```lua
local next_field
if self._private.orientation == "y" then
  next_field = function(x, y)
    if y < self._private.num_cols then
      return x, y+1
    end
    return x+1,y
  end
else
  -- Something similar here
end
while true do
  row, column = next_field(row, column)
  if not matrix[row] then matrix[row] = {} end
  if not matrix[row][column] then return row, column end
end
```
(Or something like that)

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+    -- Update the row and column numbers
+    if self._private.num_rows < row + rspan - 1 then
+        self._private.num_rows = row + rspan - 1
+    end
+    if self._private.num_cols < col + cspan - 1 then
+        self._private.num_cols = col + cspan - 1
+    end
+
+    self:emit_signal("widget::layout_changed")
+    self:emit_signal("widget::redraw_needed")
```
psychon: Why `widget::redraw_needed`? Removing a widget does not emit that signal and by symmetry both cases should emit the same signals.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+    if not row or row < 1 or row > self._private.num_rows then return false end
+    if not col or col < 1 or col > self._private.num_cols then return false end
+
+    local ret = false
+    for index, data in ipairs(self._private.widgets) do
+        if row >= data.row and row < data.row + data.row_span and
+           col >= data.col and col < data.col + data.col_span then
+            matrix_remove(self._private.matrix, data.row, data.col, data.row_span, data.col_span)
+            table.remove(self._private.widgets, index)
+            ret = true
```
psychon: Can there actually be multiple matches here? There should be only a single entry in `self._private.widgets` matching these coordinates, right? If so, there is no reason to continue the iteration.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+--- Remove one or more widgets from the layout
+-- @param widget ... Widgets that should be removed (must at least be one)
+-- @treturn boolean index If the operation is successful
+function grid:remove_widgets(...)
+    local args = { ... }
+    local ret = false
+    for _, rem_widget in ipairs(args) do
+        -- loop over the existing widgets
+        for index, data in ipairs(self._private.widgets) do
+            if data.index ~= index then data.index = index end
```
psychon: What? Why!?!? That seems quite dubious to me. If the index is not trustworthy, how about just dropping it? I know that it is listed as "for debugging", but I cannot imagine a single good debugging reason for it.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+-- @param number row The row coordinate of the widget
+-- @param number col The column coordinate of the widget
+-- @return widget The widget found at the specific coordinates, otherwise nil
+function grid:get_widget_at(row, col)
+    if not row or row < 1 or row > self._private.num_rows then return nil end
+    if not col or col < 1 or col > self._private.num_cols then return nil end
+    for _, data in ipairs(self._private.widgets) do
+        if row > data.row - 1 and row < data.row + data.row_span and
+           col > data.col - 1 and col < data.col + data.col_span then
+            return data.widget
```
psychon: Why doesn't this use `self._private.matrix`?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+        self:emit_signal("widget::layout_changed")
+    end
+end
+
+--- Force the number of rows of the layout
+-- @property forced_num_rows
+function grid:set_forced_num_rows(val)
+    if self._private.forced_num_rows ~= val then
+        self._private.forced_num_rows = val
+        self._private.num_rows = val
```
psychon: This is wrong. It should call `self:get_shape()` instead, because the grid could already have more rows than the new argument dictates. Otherwise, the next call to `get_shape` will suddenly increase the number of rows (without emitting any signals)

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+--- Helper function set default value
+-- @param string name Property name, suffix in beautiful
+-- @param object def Return value if not defined in beautiful
+function grid._set_grid_default(self, name, def)
+    local ret = def
+    if beautiful["grid_" .. name] ~= nil then
+        ret = beautiful["grid_" .. name]
+    end
+    self._private[name] = ret
+end
```
psychon: Why is this public instead of just a `local` function?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+    end
+    for j = 1,self._private.num_cols do
+        cols_size[j] = self._private.min_cols_size
+    end
+
+    -- Calculate cell sizes
+    local line
+    for row = 1, self._private.num_rows do
+        line = matrix[row]
+        if type(line) == "table" then -- check that the row exists
```
psychon: Why not just `if line then`? Same thing for other places where `type` is used to check existence.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+            for col = 1, self._private.num_cols do
+                local data = matrix[row][col]
+                if data then
+                    -- fit the widget to calculate the cell size
+                    local w, h = base.fit_widget(self, context, data.widget, orig_width, orig_height)
```
psychon: Is it really a good idea to use `orig_width, orig_height` here? For example, an `imagebox` with resizing allowed will just use all available space.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+    local used_width_max = 0
+    local used_height_max = 0
+
+    -- fit matrix cells
+    local rows_size, cols_size = get_matrix_sizes(self, context, width, height)
+
+    -- compute the width
+    if self._private.horizontal_fixed_size then
+        -- all the columns have the same size
+        used_width_max = #cols_size * max_value(cols_size) + (#cols_size - 1) * self._private.horizontal_spacing + 2 * self._private.horizontal_padding
```
psychon: Why does this widget implement its own padding, by the way? A `margin` widget around the grid could just do so as well.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+    if self._private.vertical_fixed_size then
+        -- all the rows have the same size
+        used_height_max = #rows_size * max_value(rows_size) + (#rows_size - 1) * self._private.vertical_spacing + 2 * self._private.vertical_padding
+    else
+        -- sum the rows size
+        for _,h in ipairs(rows_size) do
+            used_height_max = used_height_max + h + self._private.vertical_spacing
+        end
+        used_height_max = used_height_max - self._private.vertical_spacing + 2 * self._private.vertical_padding
+    end
```
psychon: Code duplication: How about creating a local function inside of this function and calling or once for the horizontal computations and another time for the vertical computations?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+
+    -- Fit matrix cells
+    local rows_size, cols_size = get_matrix_sizes(self, context, width, height)
+    local total_expected_width, total_expected_height = 0, 0
+    for _,w in ipairs(cols_size) do
+        total_expected_width = total_expected_width + w
+    end
+    for _,h in ipairs(rows_size) do
+        total_expected_height = total_expected_height + h
+    end
```
psychon: This might almost need a helper function like `max_value`...

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+
+    -- Figure out the maximum size we can give out to sub-widgets
+    local sub_width = max_value(cols_size)
+    local sub_height = max_value(rows_size)
+    if self._private.horizontal_expand then
+        sub_width  = (width - 2*hpadding - (self._private.num_cols-1)*hspacing) / self._private.num_cols
+    end
+    if self._private.vertical_expand then
+        sub_height = (height - 2*vpadding - (self._private.num_rows-1)*vspacing) / self._private.num_rows
```
psychon: What if these values become negative?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+        cumul_height[i] = ch
+        if self._private.vertical_fixed_size then
+            rows_size[i] = math.max(self._private.min_rows_size, sub_height)
+        elseif self._private.vertical_expand then
+            local vpercent = self._private.num_rows * sub_height * rows_size[i] / total_expected_height
+            rows_size[i] = math.max(self._private.min_rows_size, vpercent)
+        end
+        ch = ch + rows_size[i] + vspacing
+    end
+    cumul_height[#rows_size + 1] = ch
```
psychon: Ok, I'm lost...

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+    -- Place widgets
+    local fill_space = true  -- should be fill_space property?
+    for _, v in pairs(self._private.widgets) do
+        local x, y, w, h
+        -- Round numbers to avoid decimals error, force to place tight widgets
+        -- and avoid redraw glitches
+        x = math.floor(cumul_width[v.col])
+        y = math.floor(cumul_height[v.row])
+        w = math.floor(cumul_width[v.col + v.col_span] - hspacing - x)
+        h = math.floor(cumul_height[v.row + v.row_span] - vspacing - y)
```
psychon: What's a decimal error? Also, why are you `floor`ing the position, shouldn't that be `ceil`?

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,712 @@
...
+        end
+        -- Recalculate the height so the last widget fits
+        if (fill_space or self._private.vertical_expand) and y + h > math.floor(height - vpadding) then
+            h = math.floor(math.max(self._private.min_rows_size, height - vpadding - y))
+        end
+        -- Place the widget if it fits in the area
+        if x + w <= math.floor(width - hpadding) and y + h <= math.floor(height - vpadding) then
+            table.insert(result, base.place_widget_at(v.widget, x, y, w, h))
+        end
+    end
```
psychon: I know that `:layout` and `:fit` are always quite harry, but still I barely understand this one. Somehow it feels to me like this has too many options, or something like that.

lib/wibox/layout/grid.lua
getzze: I'll do that

lib/wibox/layout/grid.lua
getzze: I didn't think of that, maybe that's where a `max_[cols|rows]_size` could be useful. Otherwise grids with `imagebox` with resizing should have the `fixed_size` and `expand` properties set to true

lib/wibox/layout/grid.lua
getzze: Yes, i didn't realize that

lib/wibox/layout/grid.lua
getzze: Much better indeed :)

lib/wibox/layout/grid.lua
getzze: Well, it's a security in case the code is updated and this part is used in some other place. It could be removed, but is it really a problem?

lib/wibox/layout/grid.lua
getzze: It's not so complicated, it checks if the first widget (new widget) and second widget (loop of all present widgets) superpose by checking if they are not (first widget is left of second widget) or (second widget left of first widget)

getzze:
Several points:
 - `matrix` is used only for fit and layout, all the work of checking, adding and removing widgets only uses the `widgets` table, so there is less risk of error in case there is an an unwanted mismatch between the data in `matrix` and in `widgets`. Also `matrix` is 2D and `widgets` is 1D, so one loop less.

 - Maybe it could be useful to add a property `superpose`, where several widgets could be in the same cell. I thought about that at the beginning, so some part of the code was written to leave this possibility open.

Elv13:
I just compiled this and none of the tests and examples actually work. They don't produce errors, but result in garbage pixmap.
pick8

getzze:
Sorry about that, but I don't know how the tests work, I made minimal changes from #1120.
These are some screenshots I made, putting the grid layout in a wibox:
wibox-grid

Elv13:
You need to install ldoc, then open the doc in your browser. The files are in the build directory `doc` directory.

getzze:
Ok, the tests are all wrong indeed, now the images are correct.

getzze:
Thanks to @psychon comments and suggestions, I realized I could remove the `matrix` table and use only `widgets`. I also added the `superpose` property ;).
As for the `padding` property, I don't think it adds much complexity and in usage, the code is more compact by setting a property than using a container. But I understand that the more options, the more difficult to maintain.

psychon:
> I realized I could remove the matrix table and use only widgets.

Yay! Not much time and too sick right now for another round of review, but already thanks for your work on this and thanks that you take my criticism so well. Feel free to ping me if I forget to look at this again.

lib/wibox/layout/grid.lua
psychon: Ok. I know that this is a difficult problem (and things like Gtk sidestep it by having a "natural size" for a widget). Keeping it like this is fine with me for now (mostly because I have no good ideas on what to do instead).

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,629 @@
...
+local beautiful = require("beautiful")
+
+
+local grid = { mt = {} }
+
+local properties = {
+                    "orientation", "superpose",
+                    "forced_num_rows", "forced_num_cols",
+                    "min_cols_size", "min_rows_size",
+                    --"max_cols_size", "max_rows_size",
```
Elv13: Commented code

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,629 @@
...
+-- @beautiful beautiful.grid_orientation
+
+--- If widget superposition is allowed.
+-- @beautiful beautiful.grid_superpose
+
+--- The spacing between cells.
+-- @beautiful beautiful.grid_[horizontal|vertical]_spacing
+
+--- The space between the grid and the surrounding area.
+-- @beautiful beautiful.grid_[horizontal|vertical]_padding
```
Elv13: You can't do that, it doesn't render correctly. Copy paste isn't fun, but has to be done. Also true for all other `@beautiful`  and `@property`.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,629 @@
...
+
+--- Controls if the columns/rows are expanded to use all the available width/height.
+-- @beautiful beautiful.grid_[horizontal|vertical]_expand
+
+--- Controls if the columns/rows have the same size or if the size depend on the content.
+-- @beautiful beautiful.grid_[horizontal|vertical]_fixed_size
```
Elv13: Also, I don't think all properties should have theme variables. This isn't very useful and may break some extensions that uses the grid but don't expect a theme to change it.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,629 @@
...
+
+
+--- If false, check before adding a new widget if it will superpose with other widget and prevent from adding it.
+--@DOC_wibox_layout_grid_superpose_EXAMPLE@
+-- @property superpose
```
Elv13: Each properties need a `@param typename` or optionally `@param[opt=default_value]` below them. It will render as `type:` in the final doc.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,629 @@
...
+
+
+--- Find a widget in a widget_table, by matching the coordinates
+-- using the  `row`:`col` coordinates, and the spans `row_span` and `col_span`
+-- @param widgets_table Table of the widgets present in the grid
+-- @param number row Row number for the top left corner of the widget
+-- @param number col Column number for the top left corner of the widget
+-- @tparam number row_span The number of rows the widget spans (default to 1)
+-- @tparam number col_span The number of columns the widget spans (default to 1)
+-- @treturn table of index of widget_table
```
Elv13: If you document a local function that is then turned into a method, you need to add `@function grid:find_widgets_at` or ldoc wont see it.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,629 @@
...
+            table.insert(ret, index)
+        end
+    end
+    -- reverse sort for safe removal of indices
+    table.sort(ret, function(a,b) return a>b end)
+    return #ret > 0 and ret or nil
+end
+
+--- Get the number of rows and columns occupied by the widgets in the grid
+-- @treturn number,number the number of rows and columns
```
Elv13: Getters are auto-magically turned into properties. Please use `@property direction`. If a property is read-only, say so in the doc.

lib/wibox/layout/grid.lua
```diff
@@ -0,0 +1,629 @@
...
+        end
+        for j = data.col, data.col + data.col_span - 1 do
+            if w > cols_size[j] then cols_size[j] = w end
+        end
+    end
+    return rows_size, cols_size
+end
+
+
+--- Fit the grid layout into the given space
```
Elv13: Please use `--` for fit/draw/draw_*/layout as they are not "user facing" methods and don't need to be documented in the public doc over and over again.

lib/wibox/layout/grid.lua
getzze: You are right. The only property that could be interesting to set in the theme is `direction`, to make automatic vertical or horizontal wibars for example.

lib/wibox/layout/grid.lua
getzze: This function is only local, the methods that use it are also documented

lib/wibox/layout/grid.lua
getzze: I am not sure how to document that: below this line, or at the beginning of the file with the other property comments?

lib/wibox/layout/grid.lua
getzze: Maybe `min_rows_size` and `min_cols_size` also

lib/wibox/layout/grid.lua
psychon: Yeah. The idea is not to document the getter/setter functions at all and instead have something like:
```lua
--- The number of rows occupied by widgets in the grid.
-- @property rows
-- @tparam integer
```
(Note that this magic of turning getters into properties does not support multiple return values, so `get_dimension` would need to be split into two)

lib/wibox/layout/grid.lua
psychon: Since currently themes cannot just do this and setting this theme variable can have lots of unintended consequences, I would not recommend to add theme variables for `direction`. I'm also not sure that minimum size is a good idea...

getzze: Ok, so I removed all the theme variable

lib/wibox/layout/grid.lua
getzze: What about keeping the `get_dimension` method then, without creating a `direction` property?

getzze:
I have corrected the documentation.
I see few questions/problems:
 - the names of the properties are not perfect:
   * `fixed_size` is not really explicit, in GTK it's called `homogeneous` but I find it a bit long. 
   * `min_cols_size` or `min_columns_size` or `horizontal_min_size` or else
   * `forced_num_cols` or `forced_column_count`, idem for the rows
 - remove the `padding` property? I agree that the same result can be achieved using a margin container, but it is easier and shorter to use the `padding` property.
 - clean up the `layout` code so it is easier to maintain.
 - add some methods, for now I see two method that could be useful, `replace_widget` and `insert_column|row`.
 - more testing is needed. I have an ugly test function that resizes a grid and it's working well so far. I also have a working (and nice) calendar widget.

Elv13:
I built the new version. It's much better already. I will have to actually really tests it before having a final "lets merge it", but it's getting very close (API wise).

> remove the padding property? I agree that the same result can be achieved using a margin container, but it is easier and shorter to use the padding property.

Given it's the external padding and they do exactly the same thing as the margin container, I would prefer to have them removed. I removed the custom "vertical mode" and other duplicated features from other widgets. I prefer not to go "lets add custom margins to everything" route again. Some widgets have "real" padding, like the arcchart, but this was because of the limitations introduced by the shapes and clips. For the grid, you have inner spacing. This is totally ok. It's a great example of where having custom padding/spacing/margin makes total sense. But the external padding in also an example where no real value is added.

> `min_cols_size` and `horizontal_fixed_size` (and vertical)

Why not `max_cols_size` too? Right now it's a little hard without using constraint containers in each cells. It's strange to have a limit for small elements and no limit for big ones. I would argue the upper limit is even more useful than the lower one.

> The screenshot above

This layout is hard-ish to use only with the declarative syntax. That's ok, it's not an issue. My version wasn't supporting colspan/rowspan, so I could get away with it by using `nil` for empty cells. But it will take a code example for it. Please add another `tests/examples` file with the code from your screenshot above. Use `--DOC_NO_USAGE` and **don't** use `--DOC_HIDE_ALL`. Then place the example in the header with some explanations. See https://awesomewm.org/apidoc/classes/wibox.widget.progressbar.html for a similar page.

> wibox.layout.grid ([orientation="y"])

Please use `wibox.layout.grid.vertical` and `wibox.layout.grid.horizontal` "sub layouts" like the `fixed`, `flex` and `align` layouts. Not because it's better, just because it's more consistent.

> Other

The border support is missing. It can be implemented later. So feel free to ignore this comment.

getzze:
Ok I agree, so I will remove padding. Border cannot be supported, it's outside the padding. It can be done with a background container anyway.
I can add sub-layouts `wibox.layout.grid.vertical|horizontal`, but we should keep the "normal" layout, because most people will not care about the `orientation` property, so default is fine.

I added the support for `min_cols_size` to give a size to empty rows (like in the screenshot) and Qt has this option also, but doesn't have `max_cols_size`. It can be added, but I think most widgets have a limited size (with the exception of images with resize that can be handled with a `constraint` container), so you usely need to be sure that the cells are big enough to display them, not the contrary. And how is this going to work with `expand`? I have to think about it.

I will add some more examples. And also I wrote the code to insert, add and remove rows/columns and to replace widgets.

getzze:
I changed the `fixed_size` property name to `homogeneous`.

getzze:
The CI error is not related to the PR (I think):
Error → spec/wibox/widget/textbox_spec.lua @ 10
wibox.widget.textbox before_each
lib/beautiful/init.lua:32: May not access theme before beautiful.init()

psychon:
> The CI error is not related to the PR (I think):

That's not why CI fails. At the end, travis tests each individual commit (instead of just the result of merging this PR) and something fails there. Anyway: Yup, this should be ignored for now.

Elv13:
So, where are we at with this? There is still many "x" and "y" orientations. "x"/"y" are axis name conventions. Everything else in awesome use the words "horizontal" and "vertical".

Otherwise the doc mostly build fine and the tests pass.

getzze:
I changed the orientation to "horizontal" and "vertical".
What else can I do?

Elv13:
It looks about right, at least from an API point of view, which is the most important part to get this merged.  The commits will need to be squashed into 1 or 2 patches. I can do that when merging. I will let other devs review this, but I built it and tested the examples, it works. I could be silly and find problems in the code, but this can be fixed later. It's useful enough as-is. Thanks for this.

getzze:
I squashed the commits into two (layout and tests) but the CI tests are failing

Elv13:
The Internet was down earlier today[1], the build failed due to this. Now it failed due to a new integrity check added a few commits ago. It somehow passed back then but doesn't pass anymore. This is not related to this patch.

[1] AWS East Coast 1 (the most important data center in the world) shut down, creating a cascading effect on everything relying on something relying on something relying on something hosted there.

Elv13:
Merged! All I changed was the line lenght. Since a couple of days the new version of LuaCheck enforces 120 char hard limit (80 recommended). Otherwise, thanks a lot. It must have been frustrating with all these reviews and CI and tests and all. Getting any 1.3k line PR merged is quite a challenge.

PR #1552, opened 2017-02-10
fix(menubar): close count_file after opening

### LINKED ISSUE #1548: menubar/init.lua:  `count_file` not closed.

```
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1
```

While on a walking tour of the menubar code, I found in init.lua the functions `load_count_table()` and `write_count_table()`, which read and write a statistical history for commands invoked by menubar.  I also noted by inspection that, while the file holding the count table is opened:

    local count_file = io.open (count_file_name, "w")

It is never closed.  This leads to a bunch of stale filehandles which only get cleaned up when Lua's garbage collector decides to get around to it (verified by running `lsof | grep menu_count_file` between invocations of the menubar).

actionless:
closes #1548

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -119,14 +119,10 @@ end
 
 local function load_count_table()
     local count_file_name = awful.util.getdir("cache") .. "/menu_count_file"
-
-    local count_file = io.open (count_file_name, "r")
     local count_table = {}
-
-    -- read weight file
+    local count_file = io.open (count_file_name, "r")
     if count_file then
-        io.input (count_file)
-        for line in io.lines() do
+        for line in count_file:lines() do
             local name, count = string.match(line, "([^;]+);([^;]+)")
             if name ~= nil and count ~= nil then
                 count_table[name] = count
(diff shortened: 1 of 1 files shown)


lib/menubar/init.lua
```diff
@@ -150,6 +151,7 @@ local function write_count_table(count_table)
             io.write(str)
         end
         io.flush()
```
psychon: What the heck is this code doing? Doesn't `io.output(foo)` redirect all output to that file? And the `io.flush()` seems quite obviously unnecessary. (Yes, I know that this is unrelated to this pull request)

lib/menubar/init.lua
```diff
@@ -132,6 +132,7 @@ local function load_count_table()
                 count_table[name] = count
             end
         end
+        count_file:close()
     end
```
psychon: The whole file handling here could be replaced with `for line in io.lines(filename) do` (and that function even automatically closes the file when it hits its end, so much simpler!)

actionless: it seems like that way is no handling the case when file is not created yet

psychon: True... but it would be simpler ;-)

actionless: but shouldn't we check then the existance of that file before calling :lines on it?

actionless:
actually i've missed the review of count_table PR so only now noticed this (locally i am using pretty outdated fork of menubar which implements dmenu-like launcher, ie based on `$PATH` instead of `.desktop` files and w/o icons)

@psychon please review again after i fixed few more related problems

lib/menubar/init.lua
psychon: Yeah. What I meant (and did not write was): It would be simpler if my idea worked, but you are right that it does not and thus `file.lines(foo)` cannot be used here.

actionless:
it would be also nice to have some option or even hardcoded value for maximum records stored and mb append a date, in order to use last called date as a part of item's weight (ie more recently called item will have a bigger weight/priority) and be able to maintain size of the menu_count list, otherwise it could grow up more and more without any limits, however i can't say how many records you will need to make it a noticeable slowdown

actionless:
squashed

actionless:
i think it's neither too coo complicated nor too critical thing so i think one review is enough

PR #1555, opened 2017-02-11
Add another declarative example

Elv13:
Requested by rnabinger on IRC

aAA


psychon:
I find the resulting image confusing. How about adding different colors to make it a bit easier to tell which progressbar is which? (I was actually thinking in the wrong direction here and wondered where the free space comes from...)

actionless:
in that casse it would be better to use not just 3 different colors, but 3 different gradients, to represent the direction

Elv13:
Done

pick8

Elv13:
pick8

as requested by actionless

PR #1557, opened 2017-02-11
Add place container

Elv13:
Now that there is a new grid layout, I extracted the place container from the old PR and added the missing doc.

Close #1120

### DIFF
--- a/lib/wibox/container/place.lua
+++ b/lib/wibox/container/place.lua
@@ -1,0 +1,203 @@
+---------------------------------------------------------------------------
+-- A container used to place smaller widgets into larger space.
+--
+--@DOC_wibox_container_defaults_place_EXAMPLE@
+-- @author Emmanuel Lepage Vallee &lt;elv1313@gmail.com&gt;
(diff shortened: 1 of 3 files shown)


lib/wibox/container/place.lua
```diff
@@ -0,0 +1,203 @@
...
+local util = require("awful.util")
+
+local place = { mt = {} }
+
+-- Take the widget width/height and compute the position from the full
+-- width/height
+local align_fct = {
+    left   = function(_  , _   ) return 0                         end,
+    center = function(wdg, orig) return math.max(0, (orig-wdg)/2) end,
+    right  = function(wdg, orig) return math.max(0, orig-wdg    ) end,
```
psychon: https://github.com/awesomeWM/awesome/pull/1120/commits/0a681f23603f8aadc1fad45c65f113d2062e3df9#r80425803

psychon: Well, ok, that might not actually be possible...

blueyed:
@Elv13 
btw: when you push branches to the main repo, we would get a comment for the new apidoc automatically.

PR #1561, opened 2017-02-13
Fix #1523

### LINKED ISSUE #1523: wibars in 4.0 cover each other, inconsistent doc
```
awesome devel (Harder, Better, Faster, Stronger)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.0
```

Sup guys 😄 

In 4.0 when you have two wibars that meet in the corner, they cover each other. Something along these lines:
```lua
    local panel1 = awful.wibar{position = "left", screen = s, width = 64}
    panel1:setup{
        layout = wibox.layout.align.vertical,
        wibox.widget.textbox("FOO")
    }

    local panel2 = awful.wibar{position = "top", screen = s, height = 15}
    panel2:setup{
        widget = wibox.widget.textbox("BAR")
    }
```

It used to place `panel2` with an offset so that it wouldn't cover `panel1`. I've tried to make it work by passing some extra stuff to the 'constructor' to no avail.

EDIT: I've just found out it works if you switch the order of defining the panels, ie. the top one first. The outcome is obviously not the same though.

- The doc mentions `args.x` and `args.y`, but they don't seem to work and I don't see them used in code at first look.

- `align` is mentioned in the 'constructor', but it is not in `arg`, might be a good idea to formulate better how to set it.

Elv13:
Plus a lot of tests. Apparently there was a couple other corner case that would have been caught if the original tests were also checking the wibars geometry.

### DIFF
--- a/lib/awful/wibar.lua
+++ b/lib/awful/wibar.lua
@@ -108,7 +108,7 @@ local function get_position(wb)
     return wb._position or "top"
 end
 
-local function set_position(wb, position)
+local function set_position(wb, position, skip_reattach)
     -- Detach first to avoid any uneeded callbacks
     if wb.detach_callback then
         wb.detach_callback()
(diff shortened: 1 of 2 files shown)


lib/awful/wibar.lua
```diff
@@ -350,10 +354,16 @@ function awfulwibar.new(arg)
 
     if arg.visible == nil then w.visible = true end
 
-    w:set_position(position)
+    -- `w` needs to be inserted in `wiboxes` before reattach or it's own offset
```
blueyed: s/it's/its/

tests/test-struts.lua
```diff
@@ -99,6 +143,9 @@ table.insert(steps, function()
     assert(parent:geometry().width == small:geometry().width )
     assert(small:geometry().height == 24)
 
+    -- Hide (and hopefully GC) then old "small"
```
blueyed: s/then/the/

lib/awful/wibar.lua
```diff
@@ -350,10 +354,16 @@ function awfulwibar.new(arg)
...
+    -- wont be taken into account by the "older" wibars when `reattach` is
```
blueyed: s/wont/will not/

lib/awful/wibar.lua
```diff
@@ -137,16 +137,20 @@ local function set_position(wb, position)
...
-    -- or right wibars. To solve, this, they need to be re-attached.
-    reattach(wb)
-
     -- Set the new position
     wb._position = position
 
     -- Attach to the new position
     attach(wb, position)
+
+    -- A way to skip re-attach is required when first adding a wibar as it's not
```
blueyed: s/re-attach/reattach/ (assuming it refers to the method).

PR #1568, opened 2017-02-15
luacheck: configure it for .luacheckrc itself

blueyed:


psychon:
Uhm, why? Which problem does this fix?

actionless:
@psychon it's not enabled in travis but if you have autolinter enabled in your editor then luacheck raises:

```
  1 .luacheckrc|2 col 1 warning| setting non-standard global variable 'std'
  2 .luacheckrc|5 col 1 warning| setting non-standard global variable 'self'
  3 .luacheckrc|8 col 1 warning| accessing undefined variable 'files'
  4 .luacheckrc|11 col 1 warning| accessing undefined variable 'files'
  5 .luacheckrc|14 col 1 warning| setting non-standard global variable 'read_globals'
  6 .luacheckrc|36 col 1 warning| setting non-standard global variable 'globals'
```

actionless:
some offtopic:


`X    codecov/project — 80.66% (-0.01%) compared to 52011a6`

mb we could configure codecov not to react on such small values?

blueyed:
Yeah, Neomake made me do this.. ;)

PR #1571, opened 2017-02-15
mouse.resize_handler: remove hints with maximized states

blueyed:


Elv13:
I think it should be in `ewmh.lua`. Otherwise other `request::geometry` context could still trigger the bug

blueyed:
Done.

lib/awful/mouse/init.lua
```diff
@@ -231,19 +231,16 @@ end
...
 
         if (lay and lay == layout.suit.floating) or c.floating then
-            c:geometry {
-                x      = hints.x,
-                y      = hints.y,
-                width  = hints.width,
-                height = hints.height,
-            }
+            hints.from_mouse_resize_handler = true
+            c:emit_signal("request::geometry", context, hints)
```
Elv13: I think you created a race condition. If you allow ewmh.geometry to understand mouse.resize/mouse.move, then it will handle both request::geometry and the size will be applied twice. Better use a different context name to avoid that.

blueyed: `ewmh.geometry` uses `apply_geometry` then, which calls `c:geometry()` (just once), no?

Elv13: both `awful.mouse.resize_handler` and `awful.ewmh.geometry` are connected to the `request::geometry` signal. So they get called twice

Issue #1559, opened 2017-02-12
Handle maximize_h and maximize_v in manage


Elv13:
I mentioned the possibility it might happen in the maximize PR, well, it happened to someone.

Here is a snippet from Awesome 3.5.9 `rc.lua` (also true in 4.0 and git-master, but not as obvious)

```lua
    awful.key({ modkey,           }, "m",
        function (c)
            c.maximized_horizontal = not c.maximized_horizontal
            c.maximized_vertical   = not c.maximized_vertical
    end)
```

That thing only works if both have the same maximization state. Otherwise you "dance" between max_h and max_v without being able to unmaximize the client. In git-master, it is a little better, as you can maximize it properly, but not unmaximize. But it can also be worst and you can end up dancing between (max_h+max_x) and max_hv.

I conclude it's stupid to preserve partial maximization if the default config offers no way of unmaximizing it (with the keyboard, I know you could using the titlebar, but not anymore). So I propose to add some code in `awful.rules` to auto-remove it. Any opinons?

actionless:
that's indeed very annoying when happens to me

psychon:
...why not just change the default config to only use `c.maximize` (I guess this already happened) and tell people not to use the "partial maximization" if they run problems with it?
Alternatively: I do not really see the point in partial maximisation. How about dropping that from the C code (yay, more deprecations!).

actionless:
my bad, i've just realized wjay i changed it in hotkeys, but not in titlebar

sudo-nice:
>I do not really see the point in partial maximisation. How about dropping that from the C code

Sometimes it's handy to maximize a window in one direction only, to keep other client(s) visible. If it's possible, I would like to preserve it (if this can't be imitated with new `placement` or such features).

Elv13:
```
[15:28] <blueyed> FWIW: when you maximized kmymoney before, the max=false, max_h=false, max_v=false will make it take the whole screen (covering the task list) when starting it again in a floating layout.
[15:29] <Elv13_> The default rc.lua has a placement=no_overlap+no_offscreen. If the client size is maximized, then no_offscreen moves it back to 0x0
[15:30] <Elv13_> (well, 0x0+wibar height)
[15:30] <Elv13_> This is because KDE apps based on KXmlGuiWindow save their position and state in ~/.local
[15:31] <Elv13_> Awesome honor the size, but not the position. It should also not honor maximization status
[...]

PR #1572, opened 2017-02-15
config: add bindings to (un)maximize horizontally/vertically

blueyed:


blueyed:
I have this in my config myself, and it certainly helps with debugging https://github.com/awesomeWM/awesome/issues/1559#issuecomment-280130269.. ;)

PR #1574, opened 2017-02-15
awful.placement: unwrap the maximize_{horizontally,vertically} funcs

blueyed:
This makes it easier to find them.

### DIFF
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -1251,13 +1251,16 @@ function placement.maximize(d, args)
     return fix_new_geometry(ngeo, args, true)
 end
 
--- Add the alias functions
-for _, v in ipairs {"vertically", "horizontally"} do
-    placement["maximize_"..v] = function(d2, args)
-        args = add_context(args, "maximize_"..v)
-        args.axis = v
-        return placement_private.maximize(d2, args)
-    end
+function placement.maximize_vertically(d2, args)
+    args = add_context(args, "maximize_vertically")
+    args.axis = "vertically"
+    return placement_private.maximize(d2, args)
+end
+
+function placement.maximize_horizontally(d2, args)
+    args = add_context(args, "maximize_horizontally")
+    args.axis = "horizontally"
+    return placement_private.maximize(d2, args)
 end
 
 ---@DOC_awful_placement_maximize_vertically_EXAMPLE@


lib/awful/placement.lua
```diff
@@ -1251,13 +1251,16 @@ function placement.maximize(d, args)
...
+    args = add_context(args, "maximize_vertically")
+    args.axis = "vertically"
+    return placement_private.maximize(d2, args)
+end
+
+function placement.maximize_horizontally(d2, args)
+    args = add_context(args, "maximize_horizontally")
+    args.axis = "horizontally"
+    return placement_private.maximize(d2, args)
 end
```
blueyed: Should have proper doc then maybe, too?!

Elv13: They already have a proper doc https://awesomewm.org/apidoc/libraries/awful.placement.html#maximize_vertically

Elv13:
I am not for or against this PR. I will let psychon decide. Given it's only copy/pasted once I guess it wont hurt. But there's a lot of those in that file, so unrolling one and not unrolling the other isn't exactly making anything easier to find. Unrolling all generation loops will create a lot of copy/paste and increase the chances of accidentally fixing one but not its clones.

psychon:
I agree with the goal "making it easier to find". I will rephrase @Elv13's comment as "the whole module is hard to navigate through, read and understand" (yeah, I formulated it a bit more negative). Sorry, but I don't think this helps much with that. For example, the `stretch_`-stuff is still "done like this".

Elv13:
So, do you want to unwrap all aliases with copy/paste or leave it as-is. For the complexity and spaghetti, it's definitely there.

[[offtopic]]
The reason it's there is often to make sure all function handle `border_width` and `padding` the exact same way everywhere. This allows for very well unit tested and unified code, but readability is badly damaged in the process. Plus, some of the more obscure corner case, like the workarea changes + composite placement functions add to the overall complexity. The worst part (maintainability wise) is the pretend code required for most composite functions. It solves real bug and is really necessary, but it takes a math proof to show it because it ain't obvious at all. Also, the composite functions themselves are necessary because otherwise you can't express some configurations in multiple independent steps (same proof as the pretend code[1]).

On the other hand, beside the issues in the `no_offset` and `no_overlap` inherited from 3.5 and never really fixed, the code is well tested and "make sense" if you "get" why it's designed that way. I still think having an unified geometry subsystem is better than having geometry code all over the place. The code it replaces wasn't exactly pretty in the first place[2][3]

[1] A layman version of that goes by: Some compositions are logical `add` and some other logical `multiply` and that `a*(b+c*d) != (a*b)+(a'*c)*(a''*d)`
[2] https://github.com/awesomeWM/awesome/pull/1453#discussion_r97786651
[3] https://github.com/awesomeWM/awesome/blob/3.5/lib/awful/mouse/init.lua.in#L176 - 537

psychon:
> So, do you want to unwrap all aliases with copy/paste or leave it as-is. For the complexity and spaghetti, it's definitely there.

Yup, so let's leave it as-is. Sorry, but this PR does not improve this much.

> The worst part (maintainability wise) is the pretend code required for most composite functions. 

So... how about this: Move, for example, `placement.under_mouse` to `placement.raw.under_mouse` and at the end of the module do something like `for k, f in pairs(placement.raw) do placement[k] = wrap_client(f, k) end`. That might remove one of the "magics" in the module (but I am not really sure that it is so much better than the current metatable magic).

And, if I understand correctly, the only difference with `pretend=true` is that the resulting geometry is not applied, but only returned. So have everything always behave as if `pretend=true` were used and the caller (`wrap_client`, in the above example) would then apply the geometry (yup, this does add more metatable magic).

Should I try doing something like the above?

> [1] A layman version of that goes by: Some compositions are logical add and some other logical multiply and that a*(b+c*d) != (a*b)+(a'*c)*(a''*d)

I won't even ask what `*` and `+` are supposed to mean (besides "addition/multiplication do not distribute")...

psychon:
Sorry, I do not think this helps that much.

blueyed:
What about adding the full names as "hidden" (i.e. not appearing in the apidoc) comments?
While `ctags` would still not find it then, you would find it with regular grepping at least.

psychon:
Fine with me. That would at least help navigation through this file.

Issue #492, opened 2015-10-01
Extend the clipboard API


Elv13:
This issue is just some idea for some improvements to the clipboard selection API

* The the selection buffer (aka, middle click paste buffer)
* The the secondary clipboard buffer
* Have a signal then the clipboard selection change
* Be able to set the selection
* Get the mime type(s)

I had an `xsel` command wrapper at some point to do this, but a native API would be nicer. This would also make possible to have a clipboard overview widget. I might do this at some point, but if anyone have other ideas, please comment here.

psychon:
Can the API be asynchronous? This selections stuff just isn't synchronous and currently `selection()` kind-of starts its own main loop (but ignores everything except the XCB connection). Events are still handled normally and so lua could call `selection()` again in response to some other event. The result won't be pretty (it would be good if only the responses were mixed up; I guess instead this will never return to the normal main loop).

Elv13:
Ok, I updated the description, if you have an idea of what has to be (technically) done, feel free to elaborate, you know XCB better than me

cj1324:
+1

psychon:
@cj1324 What exactly do you want from the clipboard API? What's missing?

cj1324:
Special conditions?

https://code.google.com/p/chromium/issues/detail?id=395077

Temporary use `xsel`

psychon:
I took a quick look at https://tronche.com/gui/x/icccm/sec-2.html and I already do not want to work on this anymore...
Anyway, what should the API on the Lua-side look like?
I propose some new selection object that can be used to request a selection. That way things can be asynchronous (signals can be emitted on the object) and I have something to store some state on in the C side.
@Elv13 can you come up with some example code showing how things could work like? ICCCM specifies so much stuff that I am not sure how much of that is useful.

From you list in the initial posting here:

> The the selection buffer (aka, middle click paste buffer)
> The the secondary clipboard buffer

So, just allow a "random name" to be given as the name of the selection? Can't be more generic than that.

> Have a signal then the clipboard selection change

As far as I see, X11 does not allow to do this?!?

> Be able to set the selection

Hm. Ok, but I guess that can be done with a different API then. How about having this issue for querying the selection and someone opens a new issue for an API to set the selection.

> Get the mime type(s)

I do not see any MIME types in the protocol.

psychon:
Some random attempt at playing with selections: https://gist.github.com/psychon/91a1456eb86fca7735b1c988cbeaa138
[...]

PR #1578, opened 2017-02-18
WIP: Extend the clipboard API

psychon:
This is my today's work on #492. So far you cannot monitor for selection changes and `INCR`-transfers are not supported. Besides that, the API more-or-less works, I guess (and there are more roundtrips involved than I anticipated).


psychon:
This can now do `INCR` transfers, too. From a quick look at #492, this still needs some "the selection's content changed"-signal. However, right now I have no idea how that is supposed to work and (as always when touching low-level X11 stuff) not much motivation to figure it out.

psychon:
Oh and of course the API is currently awful:
```lua
selection("CLIPBOARD"):request("UTF8_STRING", function(part)
    if part then
        print("Next part of the clipboard is:", part)
    else
        print("Transfer done")
    end
end)
```
The C code hands each part to Lua as soon as it is received.

psychon:
I figured out how to get notifications that the clipboard changes. Now, there is a `selection_changed` signal that is emitted on selection objects. However, this now introduces some kind of memory leak: The C code pretty much leaks a reference to each selection object so that it can later find the object when an event comes in and the notification has to be emitted. If anyone has a good idea on how a better API for this would look like, I'd welcome it. The current version is:
```lua
selection("CLIPBOARD"):connect_signal("selection_changed", function() print("clipboard changed") end)
```

actionless:
seems like it should be rebased

Issue #1565, opened 2017-02-15
Widgets displayed conditionally disappear after adding/removing screens


doctaweeks:
In addition to the documentation issue confirmed in #1562 there also appears to be a bug when using conditionals in the declarative method. When adding or removing screens the widgets disappear from every screen until awesome is restarted.

awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.0

psychon:
Well, the `s.index == 1` is not really that much of a good idea, since screen index might change when you add/remove screens. When you remove a screen and another screen becomes screen 1, the code here does not run again and so the widget is not visible on the "new screen 1".

What is it that you really want to achieve? For example "Only ever show the widget on VGA1" or "if VGA1 is disconnected, show the widget on the first screen. Otherwise, show it on VGA1" (the latter would require some way to remove the widget from elsewhere when VGA1 is added).

doctaweeks:
I'll clarify the situation: This is a laptop. I have widgets set to display only on screen one so when I have it docked with multiple monitors the widgets only show up on the primary display instead of all of them.

From my perspective, this is a regression or at least a significant change in behavior that wasn't documented. If my understanding is correct, it's a side-effect of the change to no longer restart on xrandr changes (since previously the config was just processed over again with a new screen 1).

I recommend, to support the old behavior, that there is an automatic pass over the widgets during a refresh that corrects placement if it was conditional.

PR #1580, opened 2017-02-18
Add a only_on_screen container widget

psychon:
Fixes: https://github.com/awesomeWM/awesome/issues/1565

### DIFF
--- a/lib/awful/widget/init.lua
+++ b/lib/awful/widget/init.lua
@@ -19,6 +19,7 @@ return
     textclock = require("awful.widget.textclock");
     keyboardlayout = require("awful.widget.keyboardlayout");
     watch = require("awful.widget.watch");
+    only_on_screen = require("awful.widget.only_on_screen");
 }
 
 -- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/lib/awful/widget/only_on_screen.lua
+++ b/lib/awful/widget/only_on_screen.lua
@@ -1,0 +1,147 @@
+---------------------------------------------------------------------------
+--
+-- A container that makes a widget display only on a specified screen.
+--
+-- @author Uli Schlachter
+-- @copyright 2017 Uli Schlachter
+-- @classmod awful.widget.only_on_screen
+---------------------------------------------------------------------------
+
+local type = type
+local pairs = pairs
+local setmetatable = setmetatable
+local base = require("wibox.widget.base")
+local util = require("awful.util")
+local capi = {
+    screen = screen,
+    awesome = awesome
+}
+
+local only_on_screen = { mt = {} }
+
+local instances = setmetatable({}, { __mode = "k" })
+
+local function should_display_on(self, s)
+    if not self._private.widget then
+        return false
+    end
+    local success, own_s = pcall(function()
+        return capi.screen[self._private.screen]
+    end)
+    return success and own_s == s
+end
+
+-- Layout this layout
+function only_on_screen:layout(context, ...)
+    if not should_display_on(self, context.screen) then return end
+    return { base.place_widget_at(self._private.widget, 0, 0, ...) }
+end
+
+-- Fit this layout into the given area
+function only_on_screen:fit(context, ...)
+    if not should_display_on(self, context.screen) then
+        return 0, 0
+    end
+    return base.fit_widget(self, context, self._private.widget, ...)
+end
+
+--- The widget to be displayed
+-- @property widget
+-- @tparam widget widget The widget
+
+function only_on_screen:set_widget(widget)
+    if widget then
+        base.check_widget(widget)
+    end
+    self._private.widget = widget
+    self:emit_signal("widget::layout_changed")
+end
+
+function only_on_screen:get_widget()
+    return self._private.widget
+end
+
+--- Get the number of children element
+-- @treturn table The children
+function only_on_screen:get_children()
+    return {self._private.widget}
+end
+
+--- Replace the layout children
+-- This layout only accept one children, all others will be ignored
+-- @tparam table children A table composed of valid widgets
+function only_on_screen:set_children(children)
+    self:set_widget(children[1])
+end
+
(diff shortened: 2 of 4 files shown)


lib/wibox/container/only_on_screen.lua
```diff
@@ -0,0 +1,145 @@
...
+
+-- Handle lots of cases where a screen changes and thus this widget jumps around
+
+capi.screen.connect_signal("primary_changed", function()
+    for widget in pairs(instances) do
+        if widget._private.widget and widget._private.screen == "primary" then
+            widget:emit_signal("widget::layout_changed")
+        end
+    end
+end)
```
Elv13: Is this the right place for such code? Maybe the wiboxes should always be redraw when the screen change. The DPI might also change (assuming that multi-DPI code ever get merged). The context have the screen. So if the drawin get redraw, then this code become unnecessary. Also, the taglist/tasklist widgets could eventually make the `s` argument optional and track their own screens/instances internally.

lib/wibox/container/only_on_screen.lua
```diff
@@ -0,0 +1,145 @@
+---------------------------------------------------------------------------
+--
```
Elv13: Need doc

lib/wibox/container/only_on_screen.lua
```diff
@@ -0,0 +1,145 @@
...
+-- @author Uli Schlachter
+-- @copyright 2017 Uli Schlachter
+-- @classmod wibox.container.only_on_screen
```
Elv13: Should this be in `awful.widget`? This is on the line between Awesome specific and generic. If we abide to the rule where "wibox is independent from Awesome", then it makes little sense there. On the other hand, the context have the screen, so the coupling is technically already present.

I am fine both way, but I think it's worth thinking about it.

Elv13:
A simple test should be added in `test-screen-changes.lua`. Beside that and the comments above 👍

lib/wibox/container/only_on_screen.lua
psychon: Well, if the DPI does not change, then the widget context will not change and thus the wibox does not relayout/redraw everything... if I understand you correctly.
Put differently: Something still has to empty the layout caches on the widgets and that can only be done by the widget emitting the right signal.

lib/wibox/container/only_on_screen.lua
psychon: > "wibox is independent from Awesome"

Huh? Wibox is based on drawin which is part of awesome. It might have low number of interdependences, but it is still there.

psychon:
> A simple test should be added in test-screen-changes.lua.

Where I write a widget which keeps track on which screens it was drawn and use that for `assert`s? Or how else would I verify that The Right Thing (tm) happens?

lib/wibox/container/only_on_screen.lua
psychon: Yup. I think I copied this from the `mirror` container, but that one does have some example code (which is not really possible here).

Elv13:
> Or how else would I verify that The Right Thing (tm) happens?

By adding a custom property on the widget and mining the hierarchies to find it, then checks the size on that hierarchy element?

Not exactly easy[1], but it could be added to an utility module and re-used elsewhere.

Untested

```lua
 local function find_widget(h, prop)
     local w = h:get_widget()
     if widget._private and widget._private[prop] then return h end

     for _, child in ipairs(h:get_children()) do
         return find_widget(child, prop)
     end
 end
```

psychon:
> By adding a custom property on the widget and mining the hierarchies to find it,

Well, there is no public function to get the hierarchy. However, I just remembered `:find_widgets(x,y)` which might be enough for this...

The next thing would be: What to test? Setting this to `2` and `screen[2]` would work, but that's not all-that-dynamic. Would that be enough?

lib/wibox/container/only_on_screen.lua
Elv13: > It might have low number of interdependences, but it is still there.

It can be used on it's own with some minimal efforts (either from GTK+cairo or SVG like the doc tests). This is in line with your dependency work from a couple weeks back. Should `wibox` keep this "minimal interdependences" and exists on its own or should it keeps getting more dependent on Awesome and WMs concepts?

I think given it already is almost a different project, better keeping it that way.

Elv13:
> Well, there is no public function to get the hierarchy.

But there is a private one.

> The next thing would be: What to test? Setting this to 2 and screen[2] would work, but that's not all-that-dynamic. Would that be enough?

trowing ideas:

 * if it work when having 2 screens and only want the widget on one
 * does it works when the screens are swapped
 * does it work when the primary screen changes
 * does it works by screen index
 * does it works by screen name

psychon:
> if it work when having 2 screens and only want the widget on one

Added

> does it works when the screens are swapped

Huh? As in "screen's position changed"? That should not influence the test.
However, the test that I added tests both "only on screen 1" and "only on screen 2".

> does it work when the primary screen changes

The primary screen cannot be set from Lua. I could only try removing the current primary screen. Should this be added?

> does it works by screen index

Added.

> does it works by screen name

Not possible. Screen names come from RANDR, but cannot be set via `fake_add`.

blueyed:
From the coverage diff it looks like not everything is tested actually?!
https://codecov.io/gh/awesomeWM/awesome/compare/17fe135a9f3f94d70395730d8c3e95b9e8ffe688...5e82d754d6eda80ab828618967fba98ec7551232/diff

psychon:
Enhanced the test a bit more and fixed a bug in the C code that no one with a real X11 server would ever run into.

> From the coverage diff it looks like not everything is tested actually?!

I cannot test RandR output names without RandR, so the `property::outputs`-signal-stuff is not tested. Also, I'm not going to write tests for `get_children` and `set_children`. Besides that, everything should be covered now (but Travis is still building, so the results are not yet in).

psychon:
Coverage looks good. I will now do some squash, rebase and a force-push.

blueyed:
Awesome! :)

Issue #1549, opened 2017-02-09
Split up awful.util


psychon:
See https://github.com/awesomeWM/awesome/pull/1531#issuecomment-278449926

Currently `awful.util` contains functionality

* To print deprecation warnings (`util.deprecate` and `util.deprecate_class`)
* To do something to handle colors that Pango does not like (`util.ensure_pango_color`).
* To have a modulo-operator which always returns a positive result (turns result 0 into the maximum value) (`util.cycle`)
* To create directories and do shell injections (`util.mkdir`)
* To evaluate code (`util.eval`)
* To XML-(un)escape a string (`util.escape`, `util.unescape`)
* To Lua-pattern-escape a string (`util.quote_pattern`)
* To do some (ugly?) magic to make a pattern case-insenstive (`util.query_to_pattern`)
* To have a thin (useless?) wrapper around `loadfile` (`util.checkfile`)
* To restart, but only if the config does not contain any Lua syntax errors (`util.restart`)
* To get various paths (`util.get_xdg_config_home`, `util.get_xdg_cache_home`, `util.get_configuration_dir`, `util.get_cache_dir`, `util.get_themes_dir`, `util.get_awesome_icon_dir`, `util.getdir`)
* To do some icon lookup (`util.geticonpath`, see #908)
* To check some things with the file system (`file_readable`, `dir_readable`, `is_dir`)
* To produce a powerset (`subsets`)
* To do lots of weird things to tables (all the various functions in `table`)
* To do poor-man's line wrapping and line counting (`linewrap`, `linecount`)
* To round a number (`round`)

I tried to add one bullet point to the above list per "functionality". The idea is to move each of them into a new submodule of `gears` (that will be quite a number). For QA reasons, I will require tests for each of these new modules. For speed reasons, these tests should be `busted` unit tests, i.e. in `spec` (unless some module really, really, *really*  requires to be run in awesome). At the same time, we could think about dropping some of the functionality without having a replacement, if that turns out to be sensible.

Veratil:
I was curious about this, so I started with `util.ensure_pango_color` and everything else sort of branches from file hits using it. It looks like it could easily be moved to `gears.color` without issue.

Hits for `util.ensure_pango_color` (and any other `util.*` hits in the file):
* `awful/widget/taglist.lua` (1 hit)
  * `util.escape` (2 hits)
* `awful/widget/tasklist.lua` (4 hits)
  * `util.escape` (5 hits)
[...]

PR #1581, opened 2017-02-19
Move filesystem functions out of awful.util into new gears.filesystem

Veratil:
For #1549.

### DIFF
--- a/lib/gears/filesystem.lua
+++ b/lib/gears/filesystem.lua
@@ -1,0 +1,109 @@
+---------------------------------------------------------------------------
+--- Filesystem module for gears
+--
+-- @module gears.filesystem
+---------------------------------------------------------------------------
+
+-- Grab environment we need
+local Gio = require("lgi").Gio
+
+local filesystem = {}
+
+--- Create a directory
+-- @tparam string dir The directory.
+-- @return (true, nil) on success, (false, err) on failure
+function filesystem.mkdir(dir)
+	local gfile = Gio.File.new_for_path(dir)
+	local success, err = gfile:make_directory_with_parents()
+	if success then
+		return true
+	end
+	if err.domain == Gio.IOErrorEnum and err.code == "EXISTS" then
+		-- Direcotry already exists, let this count as success
+		return true
+	end
+	return false, err
+end
+
+--- Check if a file exists, is readable and not a directory.
+-- @tparam string filename The file path.
+-- @treturn boolean True if file exists and is readable.
+function filesystem.file_readable(filename)
+    local gfile = Gio.File.new_for_path(filename)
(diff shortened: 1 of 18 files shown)


lib/gears/filesystem.lua
```diff
@@ -0,0 +1,48 @@
...
+-- Grab environment we need
+local Gio = require("lgi").Gio
+
+local filesystem = {}
+
+--- Create a directory
+-- @tparam string dir The directory.
+-- @return mkdir return code
+function filesystem.mkdir(dir)
+    return os.execute("mkdir -p " .. dir)
```
psychon: Could you replace this with:
```lua
local gfile = Gio.File.new_for_path(filename)
gfile:make_directory_with_parents()
```
And, actually, some return values might be nice.
The following is untested, but might work (API documentation is missing):
```lua
local gfile = Gio.File.new_for_path(filename)
local success, err = gfile:make_directory_with_parents()
if success then
    return true
end
if err.domain = Gio.IOErrorEnum and err.code == "EXISTS" then
    -- Directory already exists, let this count as success
    return true
end
return false
```

Veratil: Tested with new code, works fine with my tests. 👍

Veratil:
Rebased off #1584 to make Travis happy

lib/gears/filesystem.lua
```diff
@@ -0,0 +1,109 @@
...
+function filesystem.get_awesome_icon_dir()
+    return (os.getenv('AWESOME_ICON_PATH') or awesome.icon_path) .. "/"
+end
+
+--- Get the user's config or cache dir.
+-- It first checks XDG_CONFIG_HOME / XDG_CACHE_HOME, but then goes with the
+-- default paths.
+-- @param d The directory to get (either "config" or "cache").
+-- @return A string containing the requested path.
+function filesystem.getdir(d)
```
Elv13: `get_dir` please (for consistency)

Elv13:
Good, so there's this one to rebase/merge, then the last one to move awful.util.deprecate to `gears.debug` and we're done! Thanks a lot for this work so far, this is greately appreciated. It takes a **lot** of patience to get these PR merged, I don't think anybody else would have gotten this far :)

lib/gears/filesystem.lua
Veratil: Sounds good, I'll update all the places used.

Veratil:
I'm going to delay rebasing until Wednesday so I can use this as an example for my team at work (we're transitioning to Git from CVS).

@Elv13 So final say is deprecate to debug? I'll update my list in #1549 with that info.

Veratil:
Rebased off 7c775e9

PR #1582, opened 2017-02-19
Move table functions out of awful.util into new gears.table

Veratil:
For #1549.

Note: I did not include `util.subsets` in this move. While it does deal specifically with tables I feel it's more a mathematics function than a table function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,7 +105,12 @@ install:
 
   # Determine custom version.
   - export AWESOME_VERSION="${TRAVIS_BRANCH}-g$(git rev-parse --short HEAD)"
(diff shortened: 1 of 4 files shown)


Veratil:
Oops, missed that util.cycle call. Thought I checked for all the function calls.

Since this will create a cyclic dependency by requiring `awful.util`, I will update this after the future `gears.math` PR is done and update accordingly.

Elv13:
The CI fails quite badly, maybe a simple rebase will do, but all the tests will need to be updated anyway, so the errors are "real" and should be fixed.

Veratil:
Rebased off #1583 to make Travis happy

Elv13:
I added a commit as discussed on IRC to force the CI to use v9999 as version string for pull requests. This way functions pending deprecations will now trigger a warning.

Veratil:
Rebased on #1637

Veratil:
All deprecation warnings are fixed in #1641

PR #1583, opened 2017-02-19
Move math functions out of awful.util into new gears.math

Veratil:
For #1549.

This leaves the old functions there with deprecation warnings while redirecting to new gears functions.

Note: This will need to be merged first before #1582, and a slight change to `gears.table.iterate` will be made afterwards.

### DIFF
--- a/lib/awful/util.lua
+++ b/lib/awful/util.lua
@@ -27,7 +27,7 @@ local capi =
     mouse = mouse
 }
 local gears_debug = require("gears.debug")
-local floor = math.floor
+local gmath = require("gears.math")
 
 local util = {}
 util.table = {}
@@ -112,16 +112,15 @@ function util.ensure_pango_color(color, fallback)
 end
 
 --- Make i cycle.
+-- @deprecated util.cycle
 -- @param t A length. Must be greater than zero.
 -- @param i An absolute index to fit into #t.
 -- @return An integer in (1, t) or nil if t is less than or equal to zero.
+-- @see gears.math
 function util.cycle(t, i)
-    if t < 1 then return end
-    i = i % t
-    if i == 0 then
-        i = t
-    end
-    return i
+    util.deprecate("gears.math.cycle", {deprecated_in=5})
+
+    return gmath.cycle(t, i)
 end
 
 --- Create a directory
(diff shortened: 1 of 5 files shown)


lib/awful/util.lua
```diff
@@ -314,8 +313,10 @@ end
 -- For example, if we consider a set with value { 10, 15, 34 },
 -- it will return a table containing 2^n set:
 -- { }, { 10 }, { 15 }, { 34 }, { 10, 15 }, { 10, 34 }, etc.
+-- @deprecated util.subsets
 -- @param set A set.
 -- @return A table with all subset.
+-- @see gears.math
 function util.subsets(set)
```
psychon: No deprecation warning?

Veratil: Gah! Copied the function, forgot to deprecate...

Elv13:
It still needs the  "deprecate after v5 is released" feature we talked about.

To get the verasion

    awesome.version

It looks like:

`v4.0-180-g0de026b91`

Thanks to the way string are compared (as used in Arch and Debian for packages version), this works:

    "v4.0-180-g0de026b91" >= "v5.0"

`false`

    "v4.0-180-g0de026b91" >= "v4.0"

`true`

Now, `awful.util.deprecate()` has an `args` argument. You need to add `deprecated_in` with an integer value (here: `5`).

The new code will look (untested)

```
if args and args.deprecated_in then
    loval cur_ver, dep_ver = awesome.version,  "v"..tostring(args.deprecated_in)..".0"
    if cur_ver < dep_ver then
        return
    end
end
```

Veratil:
> It still needs the "deprecate after v5 is released" feature we talked about.

I remembered it, but I didn't want to mess with the deprecate function just yet. It would be an easy change in `util.lua` after these update though.

Elv13:
> It would be an easy change in util.lua after these update though.

Not really, right now you are about to add new deprecations in the stable version. I wont allow that and will revert the PRs if they accidentally new merged. The 4.x API doc states `awful.util` exist and isn't deprecated. It has to stay this way until the v5 development begins. ABI/API stability sucks for developer and power users, but breaking the API sucks for the regular users. It has to be balanced.

Also, if we start adding new deprecation now without a way to track when to drop the code, we will either never drop the code or accidentally drop code for function that were never officially deprecated.

Veratil:
Ah okay, I understand. I'll see about updating the deprecate function then.

Elv13:
> Ah okay, I understand. I'll see about updating the deprecate function then.

I really appreciate the work you do here and it will be the cornerstone of a much better API. I am sorry for adding the extra work on you. It just happens that you are the first to deprecate something since 4.0 was released 2 months ago. I don't have time to hack on Awesome for the next few weeks.

Veratil:
> I really appreciate the work you do here and it will be the cornerstone of a much better API. I am sorry for adding the extra work on you. It just happens that you are the first to deprecate something since 4.0 was released 2 months ago. I don't have time to hack on Awesome for the next few weeks.

Not a problem! I'm learning more and more of the API as I dig deeper so it's all good learning for me.

spec/preload.lua
```diff
@@ -3,6 +3,9 @@
 -- is not safe to reload and yet Busted manages to do this.
 require("lgi")
 
+-- Prevent deprecated messages from showing in the Travis output
```
Elv13: Well, it's the opposite, isn't it? This does *force* all deprecation messages to show up. However the line itself is necessary. If it's not there, then there is build errors as `capi.awesome` is `nil`. I would chnage the comment to

> -- Display deprecated messages even for pending deprecation

Elv13:
/me expect the build to fail due to new luacheck

blueyed:
Thanks already!

PR #1584, opened 2017-02-19
Move string functions out of awful.util into new gears.string

Veratil:
For #1549.


lib/gears/string.lua
```diff
@@ -0,0 +1,86 @@
...
+-- @class function
+-- @name escape
+-- @param text Text to escape.
+-- @return Escape text.
+function gstring.escape(text)
+    return text and text:gsub("['&<>\"]", xml_entity_names) or nil
+end
+
+local xml_entity_chars = { lt = "<", gt = ">", nbsp = " ", quot = "\"", apos = "'", ndash = "-", mdash = "-", amp = "&" };
+--- Unescape a string from entities.
```
Bew78LesellB: I think you should add `XML` to make it more clear that it unescape from XML entities

Veratil: The more I think of pros/cons the more I agree with this.

@Elv13 what's your say on this? `escape` is used in the tasklist, taglist, prompt, menu, and menubar (from what I can tell). The name change wouldn't be hit until we actually remove it from `awful.util`. By then we'd have updated these to the new `gears.string` functions.

Veratil:
Rebased off #1619 to make Travis happy

PR #1587, opened 2017-02-20
Replace improperly used asserts with warnings in prompt.lua

morethanoneanimal:
`assert('foo') `never fails because assert treats 'foo` like a condition, not as an error message. Replacing with warnings not to break existing configs.

Try:
`lua -e 'assert("Critical error!")'`

### DIFF
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -125,6 +125,7 @@ local keygrabber = require("awful.keygrabber")
 local util = require("awful.util")
 local beautiful = require("beautiful")
 local akey = require("awful.key")
+local debug = require('gears.debug')
 
 local prompt = {}
 
@@ -518,10 +519,10 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
                 hooks[key] = hooks[key] or {}
                 hooks[key][#hooks[key]+1] = v
             else
-                assert("The hook's 3rd parameter has to be a function.")
+                debug.print_warning("The hook's 3rd parameter has to be a function.")
             end
         else
-            assert("The hook has to have 3 parameters.")
+            debug.print_warning("The hook has to have 3 parameters.")
         end
     end
 
--- a/tests/examples/wibox/awidget/prompt/hooks.lua
+++ b/tests/examples/wibox/awidget/prompt/hooks.lua
@@ -1,7 +1,8 @@
 local parent    = ... --DOC_NO_USAGE --DOC_HIDE
 local wibox     = require( "wibox"     ) --DOC_HIDE
 local awful     = { prompt = require("awful.prompt"),--DOC_HIDE
-                    util   = require("awful.util")}--DOC_HIDE
+                    util   = require("awful.util"), --DOC_HIDE
+                    spawn  = function () end } --DOC_HIDE
 local beautiful = require( "beautiful" ) --DOC_HIDE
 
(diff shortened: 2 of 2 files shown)


tests/examples/wibox/awidget/prompt/hooks.lua
```diff
@@ -1,7 +1,8 @@
 local parent    = ... --DOC_NO_USAGE --DOC_HIDE
 local wibox     = require( "wibox"     ) --DOC_HIDE
 local awful     = { prompt = require("awful.prompt"),--DOC_HIDE
-                    util   = require("awful.util")}--DOC_HIDE
+                    util   = require("awful.util"),
```
Elv13: the `--DOC_HIDE` is missing and this seems unrelated to the commit

Elv13:
Otherwise 👍 

As discussed on IRC, the first version of that patch had `print()` instead of `assert()`. This was replaced at some point, but not correctly.

tests/examples/wibox/awidget/prompt/hooks.lua
morethanoneanimal: Fixed.

Elv13: Not really, it's still unrelated to the commit. Spawn is never *used* in the test. While the example code has it, it's never called. Also, it should not be called in documentation test. We currently have no way to shim the commands.

morethanoneanimal: It's not used but it needs to be declared, otherwise `awful.spawn` is nil.
Lua ignores nil values at the end of the table (`#{1,2,3,nil,nil,nil}` is 3). Without this declaration we end up passing hook with 2 arguments and printing warning.

lib/awful/prompt.lua
```diff
@@ -518,10 +519,10 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
                 hooks[key] = hooks[key] or {}
                 hooks[key][#hooks[key]+1] = v
             else
-                assert("The hook's 3rd parameter has to be a function.")
+                debug.print_warning("The hook's 3rd parameter has to be a function.")
             end
         else
-            assert("The hook has to have 3 parameters.")
+            debug.print_warning("The hook has to have 3 parameters.")
         end
```
blueyed: Should be an error / return here IMHO?!

actionless: it was stated in PR message what the previous behavior wanted to be preserved

blueyed: I've read that, but would consider it to be a bug that it was not stopping here before.
Anyway, feel free to ignore it.

actionless: mb then fix it properly but revert to this version (with warning) if few people will be complaining on a broken behavior?

morethanoneanimal: I'm fine with asserts too, I wonder how much of an issue it would be, passing wrong parameters simply doesn't work or doesn't working as intended, so maybe the impact would be minimal?
I'd go with @actionless 's solution.

Elv13: My opinion is still that code that didn't cause an error in 4.0 should not start to cause errors in 4.1. Beside, as I said on IRC, a warning is enough. It won't break anything if the table element count is wrong. Having an error is over-reacting.

blueyed:
OK, thanks!

PR #1590, opened 2017-02-20
docs: Update for awesome 4.0 changes

jmccrohan:
s is now a screen object.

Signed-off-by: Jonathan McCrohan <jmccrohan@gmail.com>

### DIFF
--- a/docs/03-declarative-layout.md
+++ b/docs/03-declarative-layout.md
@@ -33,17 +33,19 @@ configurable rules.
 
 ### A simple layout
 
-* Display `my_first_widget` only on screen one
-* Add a background color to `my_third_widget`
+* Display `my_first_widget` only on primary screen
+* Display `my_second_widget` only on screen two
+* Add a background color to `my_fourth_widget`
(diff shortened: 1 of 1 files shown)


docs/03-declarative-layout.md
```diff
@@ -40,7 +40,7 @@ configurable rules.
 Code:
 
     s.mywibox : setup {
-        s == 1 and my_first_widget, -- Only display on screen 1
+        s.index == 1 and my_first_widget, -- Only display on screen 1
```
actionless: mb better to replace it to `s == screen.primary`?

jmccrohan: @actionless Personally, I find the screen.index notation more powerful, but I can see where other would be useful.

How about I add another widget to that imaginary wibox, one using s == screen.primary, and one using s.index == 2? This would demonstrate both?

blueyed: @jmccrohan 
Please do.

jmccrohan: @blueyed 
Done. See 4719726.

PR #1592, opened 2017-02-22
doc/images/mouse.svg: fix indexes

copycat-killer:
The mouse image [here](https://awesomewm.org/doc/api/libraries/mouse.html) is reporting wrong indexes for RMB and scroll click, that is 2 and 3, while it's currently the opposite. This PR fix it.

Otherwise, should we change (force) the indexes to this more (human) logical view?

Is there any work on this done? I can't seem find it.

### DIFF
--- a/docs/images/mouse.svg
+++ b/docs/images/mouse.svg
@@ -14,7 +14,7 @@
    viewBox="0 0 546.85725 295.98017"
    id="svg2"
    version="1.1"
-   inkscape:version="0.91 r13725"
+   inkscape:version="0.92.1 r"
    sodipodi:docname="mouse.svg">
   <defs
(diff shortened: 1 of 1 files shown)


Elv13:
> Otherwise, should we change (force) the indexes to this more (human) logical view?
> Is there any work on this done? I can't seem find it.

Not without undesirable API breaks. I added

```
mouse.is_left_mouse_button_pressed
mouse.is_right_mouse_button_pressed
mouse.is_middle_mouse_button_pressed
```

Properties to the mouse module. I also played with a couple of intex-to-name helper, but none of them were really helpful.

blueyed:
Thanks!

PR #1597, opened 2017-02-25
awful.screen: extract screen:get_next_in_direction()

cmertz:
This extracts the code for finding the next screen
from focus_bydirection to a separate method on
the screen object.

The main reason was to use the finding code without
actually changing the screen focus but this should
incidentally make the code slightly easier to test
as well since both concerns can be tested in
isolation.

Signed-off-by: Christoph Mertz <chris@nimel.de>

### DIFF
--- a/lib/gears/geometry.lua
+++ b/lib/gears/geometry.lua
@@ -11,10 +11,6 @@
 ---------------------------------------------------------------------------
 local math = math
 
-local pow = math.pow or function(a,b) -- lua 5.3 compat
-    return a^b
-end
-
 local gears = {geometry = {rectangle = {} } }
 
 --- Get the square distance between a rectangle and a point.
@@ -121,7 +117,7 @@ local function calculate_distance(dir, _gA, _gB)
         gAx = _gA.x + _gA.width
     end
 
-    return math.sqrt(pow(gBx - gAx, 2) + pow(gBy - gAy, 2))
+    return math.sqrt((gBx - gAx) ^ 2 + (gBy - gAy) ^ 2)
 end
 
 --- Get the nearest rectangle in the given direction. Every rectangle is specified as a table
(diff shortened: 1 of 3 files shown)


blueyed:
Thanks!
I've pushed a fixup.  Is it OK?

blueyed:
@Elv13 
I think we should cover this with tests.
Where should this go?  A new file, or some existing one?

cmertz:
Thx for the ldoc fix, it should however be `@function screen:get_next_in_direction` if i'm not mistaken.

If you want i can write tests (probably busted specs?) for it as well.

blueyed:
@cmertz 
Good catch - amended.

As for the tests, I think it needs / should be integration tests (via `tests/run.sh`).

psychon:
If possible, I always prefer unit tests (Stuff in `spec/`) over the integration-unit-mix that is `tests/` (main reason would be: It's a lot faster). However, `spec/` does not have access to a running `awesome` instance and so anything that needs the C-API cannot be tested there.

The tests in `spec/awful/screen_spec.lua` mock the `screen` C-API to do tests. From a quick look, it should be possible to test this there.

cmertz:
Looks like it broke the specs under lua 5.3 due to `math.pow` beeing deprecated (https://www.lua.org/manual/5.3/manual.html#6.7).

cmertz:
Anyways i think it makes more sense to spec `gears.geometry` instead of implicitly specing its behaviour through its client code.

Thoughts on this?

lib/gears/geometry.lua
```diff
@@ -11,6 +11,10 @@
 ---------------------------------------------------------------------------
 local math = math
 
+local pow = math.pow or function(a,b) -- lua 5.3 compat
+    return a^b
```
Elv13: Is this necessary? Is there a Lua version without `^`? I tried with Lua 5.1 and LuaJIT2

cmertz: sorry, i effectively did not check if `^` would work below 5.3

Elv13:
Looks good to me. @blueyed squash/merge if you agree

Elv13:
thanks

PR #1600, opened 2017-02-25
Corrected selected check in awful.tag.set_screen()

Drauthius:
Previously it would check if the awful.tag.selected function existed,
and not whether the tag was actually selected.


blueyed:
Thanks for this fix!

We should add a regression test for this.  /cc @Elv13

PR #1601, opened 2017-02-25
Improvements on wibox.container.scroll thanks to Unode

psychon:
@Unode made me do this!


Unode:
"made" is such a strong word! 👍

Elv13:
> I think it should be squashed.

The commits make sense individually, so I don't think squashing them will add value. PRs a piling up while Uli is on "sick leave", so better start merging things again before it goes out of control. I propse you either squash them or merge this as-is.

PR #1602, opened 2017-02-25
wibar: Add beautiful variables

Elv13:
This was done a few weeks ago for the notifications. This was
requested on IRC a while ago to have different font for the wibars.

Now that shapes are supported, it also makes more sense for the
border.

This was in my stash since more than a year.


blueyed:
@Elv13 
Please squash-merge when you're ready / agree.

PR #1603, opened 2017-02-25
Add rc.lua and theme.lua to ldoc

### LINKED ISSUE #1576: doc: Porting tips: add link to old (3.5) rc.lua and tell people to diff from that
The 3.5 -> 4.0 porting tips at https://awesomewm.org/doc/api/documentation/17-porting-tips.md.html, as written now, could be a bit more helpful. Specifically, after the first paragraph, which goes 

"""
The **best advice is to start from the default rc.lua and backport any changes** you might have added to your rc.lua. This avoids most of the possible errors due to missing important changes.
"""

I would like to add another paragraph that instructs the user how to do so (the main point is having links to the new and old defaults). As an example:

"""
To do this, you can download the default rc.lua for 3.5.9 from <insert_permalink>, and then compare your existing rc.lua with the 3.5.9 default using your diff tool of choice. Write down the changes, then apply these to the 4.0 default rc.lua, which you can find at /etc/xdg/awesome/rc.lua after the upgrade, or at <this_permalink> if you have not yet performed the upgrade.
"""

Before the "I warned you, but here is the difficult way" text that is most of the page today, I would then like to see included a list of changes that are needed for the new rc.lua to go back to the 3.5.9 defaults. In my case, that meant digging up how to disable the titlebars (now on by default, I don't like them personally). Probably there are a few more such "frequent requests".

Elv13:
Fix  #1576

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -389,6 +389,9 @@ endif()
 #}}}
 
 # {{{ Generate some aggregated documentation from lua script
+
+file(MAKE_DIRECTORY ${BUILD_DIR}/script_files/)
+
 add_custom_command(
         OUTPUT ${BUILD_DIR}/docs/06-appearance.md
         COMMAND lua ${SOURCE_DIR}/docs/06-appearance.md.lua
@@ -398,9 +401,11 @@ add_custom_command(
 
 add_custom_command(
         OUTPUT ${BUILD_DIR}/awesomerc.lua ${BUILD_DIR}/docs/05-awesomerc.md
+            ${BUILD_DIR}/script_files/rc.lua
         COMMAND lua ${SOURCE_DIR}/docs/05-awesomerc.md.lua
         ${BUILD_DIR}/docs/05-awesomerc.md ${SOURCE_DIR}/awesomerc.lua
         ${BUILD_DIR}/awesomerc.lua
+        ${BUILD_DIR}/script_files/rc.lua
 )
 
 # Create a target for the auto-generated awesomerc.lua
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -408,8 +408,23 @@ add_custom_command(
         ${BUILD_DIR}/script_files/rc.lua
 )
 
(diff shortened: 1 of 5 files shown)


awesomeConfig.cmake
```diff
@@ -398,13 +401,30 @@ add_custom_command(
 
 add_custom_command(
         OUTPUT ${BUILD_DIR}/awesomerc.lua ${BUILD_DIR}/docs/05-awesomerc.md
+        ${BUILD_DIR}/script_files/rc.lua
```
blueyed: This belongs to `OUTPUT`, right?
Should be indented then maybe..

Also for the `COMMAND` lines below then.

docs/05-awesomerc.md.lua
```diff
@@ -242,7 +242,16 @@ local rc = assert(io.open(rcfile))
 
 local doc_block = false
 
-local output = {}
+local output, output_script = {}, {[[
+---------------------------------------------------------------------------
+--- The default rc.lua file.
+--
+-- A copy is this file is usually installed in `/etc/xdg/awesome/`.
```
blueyed: s/is/of/

docs/05-awesomerc.md.lua
```diff
@@ -242,7 +242,16 @@ local rc = assert(io.open(rcfile))
...
+--
+-- See [The declarative layout system](../documentation/05-awesomerc.md.html)
+-- for a more documented version of this file.
```
blueyed: s/for a more documented version/for a version with additionally (auto-generated) comments/

docs/sample_theme.lua
```diff
@@ -0,0 +1,29 @@
...
+[[---------------------------------------------------------------------------
+--- The default theme file.
+--
+-- If you wish to create a custom theme, copy this file to
+-- `~/.config/awesome/mytheme.lua`
+-- and replace:
+--
+--    beautiful.init(awful.util.get_themes_dir() .. "default/theme.lua")
+--
+-- by
```
blueyed: s/by/with/

Elv13:
done

PR #1604, opened 2017-02-25
Add naughty presets `ok`, `warn` and `info`.

Lawful-Lazy:
Inspired by [Bootstrap's Alerts](http://getbootstrap.com/components/#alerts), I decided to make three extra presets for naughty's system:

* **ok** - Notification of success or sanity. 5 second timeout.
* **warn** - Notification of unwanted state, but not critical. 10 second timeout.
* **info** - Notification the user should be aware of, but unimportant. 5 second timeout.

screenshot


Elv13:
thanks

Issue #1554, opened 2017-02-11
Use request::geometry in event.c


Elv13:
<s>A regression was introduced by the maximize refactor PR</s>. Before that, ewmh.lua was listening to property::geometry to re-apply the maximize/fullscreen geometry (edit: but was then bypassed again). Right now, if a client resize itself, `client_resize` is called directly from event.c. This bypass the safeguard that limits the geometry changes of maximized/fullscreen clients. Before the PR, it would flicker, but nobody cared. Now, as the handler was bypassed, the geometry is wrong.

This happens, for example, when a terminal changes its font size or the corner size handle is used.

The `request::` API exists to let Lua handler filters the "raw" events. This allows custom behavior, such as maximize or focus stealing filters, to be implemented (and replaced) outside of the C core. Apparently this one was overlooked when the API was implemented.

```
[20:22] <blueyed> Similar: when resizing the font in a maximized urxvt, it will not occupy the whole screen afterwards.
[20:22] <Elv13> possible, there was a lot of changes in that code
[20:23] <Elv13> ah, that one I know why
[20:23] <blueyed> Yeah, I feel already bad about typing them down here instead of reporting them properly.. :/
[20:23] <blueyed> why?
[20:23] * Elv13 reading event.c
[20:24] <Elv13> event.c: 399
[20:24] <Elv13> client_resize(c, geometry, false);
[20:24] <Elv13> that should be request::geometry
[20:24] <Elv13> but not all the time
[20:25] <Elv13> I found that while debugging an unrelated problem this morning (that was a bug in the application after closer inspection)
[20:26] <Elv13> blueyed: one of the change my PR did was to only listen to border_width and maximize change in the callback
[20:26] <Elv13> 3.5 and 4.0 also used property::geometry, but this created a race condition
[20:27] <Elv13> There was/is an ugly mutex in the code to prevent an infinite loop, but it wasn't enough
[20:27] <Elv13> This is why removing the border on fullscreen clients and other trick were not reliable
[20:28] <Elv13> so I removed the property::geometry and "trusted" that request::geometry would be used, but int he case you change the font size in urxvt, the lua side is never notified, the C side just applies it
[20:28] <Elv13> This is why you have that bug
[20:29] <Elv13> blueyed: can you fix it?
[20:30] <blueyed> Elv13: I don't think so - at least currently.. :)
```

Elv13:
According to blueyed on IRC, the bug is much older after all

blueyed:
[...]

PR #1606, opened 2017-02-26
emit ewmh geometry requests as signals

wacossusca34:
This patch will emit `request::geometry` signals when clients reposition or resize themselves, instead of simply applying the new geometry straight from C. `awful.ewmh` has been tweaked to always apply geometry requests to dockable clients (allowing panels and other programs to position themselves correctly).

Further changes are probably nessecary, notably changing the provided layouts to handle the new geometry requests. Tested (somewhat) under my own configuration, and the default config.

Note: some applications expect the request to be applied (eg. firefox when it restores a session, assigning the saved position and size of its windows), and will render incorrectly if the WM does not apply the new geometry. In the scenario where the layout does not allow the client to resize itself, a simple fix would be to update the client's size with its existing geometry.

Related to #1554 - "Yes, removing `property::geometry` did cause that."

Thanks to @Elv13 for providing the rough initial patch.

### DIFF
--- a/event.c
+++ b/event.c
@@ -340,6 +340,8 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
         uint16_t deco_top = bw + tb_top;
         uint16_t deco_bottom = bw + tb_bottom;
         int16_t diff_w = 0, diff_h = 0, diff_border = 0;
+        
+        lua_State *L = globalconf_get_lua_State();
 
         if(ev->value_mask & XCB_CONFIG_WINDOW_X)
         {
@@ -373,8 +375,6 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
         }
         if(ev->value_mask & XCB_CONFIG_WINDOW_BORDER_WIDTH)
         {
-            lua_State *L = globalconf_get_lua_State();
-
             diff_border = ev->border_width - bw;
             diff_h += diff_border;
             diff_w += diff_border;
@@ -396,7 +396,16 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
         }
 
         c->got_configure_request = true;
-        client_resize(c, geometry, false);
+        
+        /* Request the changes to be applied */
+        luaA_object_push(L, c);
+        lua_pushstring(L, "ewmh");     /* context */
+        lua_newtable(L);               /* props */
+        lua_pushstring(L, "geometry");
+        luaA_pusharea(L, geometry);    /* geometry request */
+        lua_rawset(L, -3);
+        luaA_object_emit_signal(L, -3, "request::geometry", 2);
+        lua_pop(L, 1);
     }
     else if (xembed_getbywin(&globalconf.embedded, ev->window))
     {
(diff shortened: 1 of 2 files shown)


lib/awful/ewmh.lua
```diff
@@ -246,6 +246,12 @@ local context_mapper = {
 function ewmh.geometry(c, context, hints)
     local layout = c.screen.selected_tag and c.screen.selected_tag.layout or nil
 
+    -- For dockable clients, always apply new geometry
+    if c.dockable and context == "ewmh" then
+        c:geometry(hints.geometry)
+        return
+    end
+
```
Elv13: This wont work as it changes the behavior compared to 4.0 and the next release is a stable one. I propose

```
diff --git a/lib/awful/ewmh.lua b/lib/awful/ewmh.lua
index 817a69bdb..c30fe563d 100644
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -305,10 +305,25 @@ function ewmh.geometry(c, context, hints)
     end
 end
 
+--- Allow the client to move itself.
+--
+-- This is the default geometry request handler when the context is `ewmh`.
+--
+-- @signalhandler awful.ewmh.client_geometry_requests
+-- @tparam client c The client
+-- @tparam string context The context
+-- @tparam[opt={}] table hints The hints to pass to the handler
+function ewmh.client_geometry_requests(c, context, hints)
+    if context == "ewmh" then
+        c:geometry(hints.geometry)
+    end
+end
+
 client.connect_signal("request::activate", ewmh.activate)
 client.connect_signal("request::tag", ewmh.tag)
 client.connect_signal("request::urgent", ewmh.urgent)
 client.connect_signal("request::geometry", ewmh.geometry)
+client.connect_signal("request::geometry", ewmh.client_geometry_requests)
 client.connect_signal("property::border_width", repair_geometry)
 client.connect_signal("property::screen", repair_geometry)
 screen.connect_signal("property::workarea", function(s)
```

event.c
```diff
@@ -396,7 +396,16 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
...
         c->got_configure_request = true;
-        client_resize(c, geometry, false);
+        
+        /* Request the changes to be applied */
+        luaA_object_push(L, c);
+        lua_pushstring(L, "ewmh");     /* context */
+        lua_newtable(L);               /* props */
+        lua_pushstring(L, "geometry");
+        luaA_pusharea(L, geometry);    /* geometry request */
+        lua_rawset(L, -3);
```
Elv13: The mouse-resize `request::geometry` handler put `x`/`y`/`width`/`height` directly in the hints. This should be done here too for consistency.

lib/awful/ewmh.lua
wacossusca34: @Elv13 That will work for preserving behaviour, but ideally you would want to handle requests according to the layout. Bug #1554 will persist if this approach is taken.

event.c
wacossusca34: @Elv13 Sounds good.

Elv13: Do you plan to do it?

Elv13:
Merged

Issue #1607, opened 2017-02-26
Fullscreening a window leaves a gap of size <border_width> for some clients


psychon:
Output of `awesome --version`:

As you wish: :-)
```
awesome 9999 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.2.4 (running with Lua 5.2)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1
```

**How to reproduce the issue:**

* Play a video with mpv/mplayer
* Fullscreen the video player (either through awesome via mod4+f or through the client by pressing f)

**Actual result:**

To quote `xwininfo`:
```
  Absolute upper-left X:  1
  Absolute upper-left Y:  1
```

**Expected result:**

The window should end up in the upper left corner of the screen.

**Some more information on what is going on:**

ICCCM brings us this great thing that is window gravities[*]. For `mplayer` the specified value can be seen here:
```
WM_NORMAL_HINTS(WM_SIZE_HINTS):
		program specified location: 640, 404
		program specified size: 640 by 272
		program specified minimum size: 4 by 4
		program specified minimum aspect ratio: 640/272
		program specified maximum aspect ratio: 640/272
		program specified base size: 0 by 0
		window gravity: Static
```
For example Chromium does not have a `window gravity`-entry in its `WM_NORMAL_HINTS` and the bug is not reproducible with Chromium.

What happens is that `awful.ewmh` first calls `awful.placement.maximize` with `ignore_border_width = true` and only later sets `c.border_width = 0`. Changing the border width invokes behaviour specified by ICCCM (changing the size without specifying a new position). `mplayer` having a `Static`-gravity means that we should move it so that "it stays where it was before we added/removed window decorations", which means the window is shifted to the bottom/right by its previous border width (well, stays where the border previously moved it...).

No idea how this should be fixed. Anyone wants to hack the C code so that the gravity is more-or-less ignored for changes done from Lua and only applied in places where ICCCM requires us to do so?

[*]: Actually, gravities are an X11-thing, but ICCCM adds them to one more use and this bug happens to be about this part...

P.S.: Assigning to the v4.1 milestone under the assumption that this might be a regression. I haven't actually checked. Anyone who thinks "this is not a regression" or "this shouldn't be in the milestone" is free to remove it (assuming he has the necessary access rights for that...).
[...]

PR #1608, opened 2017-02-26
Add gravity awful.placement + tests [WIP]

Elv13:
Begin work on #1607, the current patch isn't correct and it's only the beginning. It will need a *lot* more tests before I change all the placement code to support gravity.

The good news:

 * It can reproduce the problem
 * It isn't a "true" regression compared to 4.0, only a different bug (they both misbehave differently)

The bad news:

* If it ever "worked", it was because the code to remove the border was broken/deadcode
* I ran the tests with other gravities and saying "is isn't pretty" would be an euphemism...
* Fixing all the corner cases will take a lot of work
* It never worked, not even remotely
* `git grep -En '(border|bw)' | grep 2 | grep '*'` on Awesome 3.5 show many similar issues, so this is nothing new... but it's worst as I added many new placement function and none of them handle gravity correctly

### DIFF
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -1276,13 +1276,13 @@ function placement.maximize(d, args)
     local bw   = (not args.ignore_border_width) and d.border_width or 0
 
     if (not args.axis) or args.axis :match "vertical" then
-        ngeo.y      = sgeo.y
-        ngeo.height = sgeo.height - 2*bw
+        ngeo.y      = apply_border_width(sgeo, 'y'     , bw, d.size_hints, -1)
+        ngeo.height = apply_border_width(sgeo, 'height', bw, d.size_hints, -1)
     end
 
     if (not args.axis) or args.axis :match "horizontal" then
-        ngeo.x      = sgeo.x
-        ngeo.width  = sgeo.width - 2*bw
+        ngeo.x      = apply_border_width(sgeo, 'x'    , bw, d.size_hints, -1)
+        ngeo.width  = apply_border_width(sgeo, 'width', bw, d.size_hints, -1)
     end
 
     geometry_common(d, args, ngeo)
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -273,6 +273,19 @@ local outer_positions = {
     bottom2 = function(r, w, _) return {x=r.x-w+r.width, y=r.y            }, "left"  end,
 }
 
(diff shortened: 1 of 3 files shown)


lib/awful/placement.lua
```diff
@@ -273,6 +273,19 @@ local outer_positions = {
...
+local gravity_map = {
+    north_west = {x = 0, y = 0, width = 2, height = 2},
+    north_east = {x = 2, y = 2, width = 0, height = 0}, --FIXME
+    south_east = {x = 2, y = 2, width = 0, height = 0}, --FIXME
+    south_west = {x = 2, y = 2, width = 0, height = 0}, --FIXME
+    center     = {x = 1, y = 1, width = 1, height = 1},
+    north      = {x = 1, y = 1, width = 1, height = 1}, --FIXME
+    south      = {x = 1, y = 1, width = 1, height = 1}, --FIXME
+    east       = {x = 1, y = 1, width = 1, height = 1}, --FIXME
+    west       = {x = 1, y = 1, width = 1, height = 1}, --FIXME
```
psychon: static gravity is missing from this list.

psychon:
This PR tries to "work around" gravities in the Lua code. If we go down that route, we might end up with lots of complicated code, so I'll try to step back a bit:

There are two kinds of operations that need to be differentiated:
- Adding decorations to a client (gravity needs to be handled)
- Placing a client (gravity is not important)

"Adding decorations" is traditionally just titlebars, but for us that's also window borders since we use them as "some kind of" titlebar.

Current behaviour:
- Changing titlebar size and border size: The C code corrects for window gravity
- Placing the window (`c:geometry`): Gravity does not matter

As #1607 shows: Border size changes is not always related to gravity.

So, should gravity handling perhaps be removed from the C code (at least the API accessible to Lua) and into rc.lua instead?
Another idea: Move "border width is set to 0 for fullscreen clients" into the C code?
What other options do we have?

Elv13:
@psychon The less changes to `awful.placement`, the better. If the C side could abstract all the corner cases, then that's good.

Do you want me to merge the 3 first commits, remove the other 2 and close this?

psychon:
I tried to do something "like this" in: https://github.com/awesomeWM/awesome/pull/1614 (but incomplete since no Lua code was touched)
Dunno about merging this yet. That would add unused code to the test suite, wouldn't it?

Elv13:
This is a dead end

PR #1610, opened 2017-02-27
ewmh.tag: Fix the transient_for clients without parent client tags

### LINKED ISSUE #1594: Windows spawned from hidden clients will be hidden as well

```
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1
```

1. Create a terminal
2. Run something like `sleep 5; zenity --info --title=foo --text=bar`
3. Hide the terminal by setting `client.hidden = true`

I have a script that creates and hides a "scratchpad" terminal on a keypress:
https://github.com/majutsushi/etc/blob/master/awesome/scratch/uniqdrop.lua

**Actual result:**

The window gets created, but is hidden and inaccessible

This worked fine in Awesome 3.5, so something relevant seems to have changed in 4.0.

I tried running this to get the new window to show up, but it didn't do anything:
```
echo 'for _,c in ipairs(client.get()) do c.hidden = false end' | awesome-client
```

**Expected result:**

The window gets created as if the originating client wasn't hidden

Elv13:
Fix #1594 
Made in the GitHub UI

### DIFF
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -204,7 +204,8 @@ function ewmh.tag(c, t, hints) --luacheck: no unused
         if c.transient_for and not (hints and hints.reason == "screen") then
             c.screen = c.transient_for.screen
             if not c.sticky then
-                c:tags(c.transient_for:tags())
+                local t = c.transient_for:tags()
+                c:tags(#t > 0 and t or c.transient_for.screen.selected_tags)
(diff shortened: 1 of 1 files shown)


Elv13:
It was tested by an user and +1 by blueyed in the other issue, so merging this to stop PR from piling up

PR #1613, opened 2017-03-01
Add beautiful property to control busy cursor during spawn

SammysHP:


lib/awful/startup_notification.lua
```diff
@@ -14,13 +14,17 @@ local capi =
...
     root = root
 }
+local beautiful = require("beautiful")
 
 local app_starting = {}
 
 local cursor_waiting = "watch"
 
+--- Show busy mouse cursor during spawn.
+-- @beautiful beautiful.enable_spawn_cursor
```
Elv13: +-- @beautiful beautiful.enable_spawn_cursor
+-- @tparam[opt=true] boolean enable_spawn_cursor

lib/awful/startup_notification.lua
```diff
@@ -14,13 +14,17 @@ local capi =
...
+
 local function update_cursor()
-    if #app_starting > 0 then
+    if #app_starting > 0 and (beautiful.enable_spawn_cursor or beautiful.enable_spawn_cursor == nil) then
```
Elv13: `beautiful.enable_spawn_cursor ~= false` will do

SammysHP: Sure! Too late and had too much work today. :)

blueyed:
What about controlling the cursor being used (a default of "watch" then)?
Using false would then use "left_ptr".

SammysHP:
@blueyed I had the same idea but couldn't think of any other cursor that a user might want to use. That should be changed in the cursor theme.

Elv13:
Why not, but I think this belong to it's own beautiful variables. Something like `spawn_cursor`. But this means changing the name to `spawn_cursor_enabled` to have them both in the same namespace

Elv13:
Will someone ever use another cursor than the current one or left_ptr? After thinking about it, I am not sure anybody will ever use such option. Did I miss something? Otherwise, if there is no further activity, I will merge this as-is in 2 days.

PR #1614, opened 2017-03-02
WIP: Force a client's border width to be zero while fullscreen

psychon:
This is WIP because I did exactly no testing at all of this. Also, someone still has to remove the code from `awful.placement` that tries to set the border width to 0.

Ref: https://github.com/awesomeWM/awesome/issues/1607
Ref: https://github.com/awesomeWM/awesome/pull/1608#issuecomment-283414465

### DIFF
--- a/event.c
+++ b/event.c
@@ -380,7 +380,7 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
             diff_w += diff_border;
 
             luaA_object_push(L, c);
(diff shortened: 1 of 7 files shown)


psychon:
Closing because:
- superseded by #1764 
- this doesn't work anyway according to https://github.com/awesomeWM/awesome/issues/1607#issuecomment-300020880

PR #1618, opened 2017-03-02
Added _hover & _click to client buttons

legostax:
Every client button type (e.g. minimize, maximize_inactive, maximize_active, close) has the option to show a different icon when the mouse hovers over it or a "button::press" signal is sent.
Hopefully my documentation isn't too repetitive...
Signed-off-by: Lego Stax legostax@gmail.com

### DIFF
--- a/lib/awful/titlebar.lua
+++ b/lib/awful/titlebar.lua
@@ -89,8 +89,8 @@ local titlebar = {
 -- @param surface
 -- @see gears.surface
 
---- minimize_button_normal_click.
--- @beautiful beautiful.titlebar_minimize_button_normal_click
+--- minimize_button_normal_press.
+-- @beautiful beautiful.titlebar_minimize_button_normal_press
 -- @param surface
 -- @see gears.surface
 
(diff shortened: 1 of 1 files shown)


Elv13:
Thanks for this. The CI wants you to fix some unused function arguments

```
    lib/awful/titlebar.lua:648:56: unused argument mod
    lib/awful/titlebar.lua:648:61: unused argument x
    lib/awful/titlebar.lua:648:64: unused argument y
    lib/awful/titlebar.lua:663:50: unused argument mod
    lib/awful/titlebar.lua:663:55: unused argument x
    lib/awful/titlebar.lua:663:58: unused argument y
```

Otherwise `titlebar_floating_button_normal_active_click` sigh... How long before we get 16 layers of options `5*2*2^16` titlebar variables, yay! Anyway, can't argue with hover/pressed as other WMs have them

I would however rename `click` to `press`

legostax:
> lib/awful/titlebar.lua:648:56: unused argument mod, x, y

Fixed this one, but I'm pretty lost as far as using `mouse.is_left_mouse_button_pressed` to fix the second one.  I've tried it in my `rc.lua` by running a timer that fires every second.  I hold the left mouse button and it still returns false.  Any ideas?

Elv13:
Just name the arguments `_` if you don't use them (like in Go)

Elv13:
Ok, good, the linting now works. We currently don't have integration tests for this, so I guess it can be merged. Testing it isn't impossible, but it wont be funny.

legostax:
I tested it for close, maximize, and minimize buttons.  I imagine ontop, sticky, and floating should work too.  However, I don't know what would happen if the _inactive/_active or _normal/_focus attributes were omitted.

Elv13:
Thanks

PR #1619, opened 2017-03-03
Move ensure_pango_color out of awful.util to gears.color

Veratil:
For #1549.

The change to `wibox/widget/base.lua` is to prevent a cyclic require chain. Details [here](https://github.com/awesomeWM/awesome/issues/1549#issuecomment-283810356).


Veratil:
Rebased off #1582 to make Travis happy

Issue #1621, opened 2017-03-03
New Luacheck release breaks our Travis builds


psychon:
3 hours ago we got: https://github.com/mpeterv/luacheck/releases/tag/0.19.0
The result is breakage: https://travis-ci.org/awesomeWM/awesome/jobs/207446828
```
[...]
Checking lib/awful/completion.lua                 1 warning
    lib/awful/completion.lua:109:1: line is too long (126 > 120)
Checking lib/awful/dbus.lua                       OK
Checking lib/awful/ewmh.lua                       1 warning
    lib/awful/ewmh.lua:101:9: indirectly setting read-only field focus of global client
[...]
Checking lib/awful/layout/suit/tile.lua           3 warnings
    lib/awful/layout/suit/tile.lua:120:1: line is too long (139 > 120)
    lib/awful/layout/suit/tile.lua:268:1: line is too long (126 > 120)
    lib/awful/layout/suit/tile.lua:288:1: line is too long (137 > 120)
Checking lib/awful/menu.lua                       2 warnings
    lib/awful/menu.lua:635:1: line is too long (219 > 120)
    lib/awful/menu.lua:636:1: line is too long (225 > 120)
[...]
Checking lib/awful/placement.lua                  1 warning
    lib/awful/placement.lua:98:26: indirectly accessing undefined field unpack of global table
Checking lib/awful/prompt.lua                     1 warning
    lib/awful/prompt.lua:123:26: indirectly accessing undefined field unpack of global table
Checking lib/awful/remote.lua                     1 warning
    lib/awful/remote.lua:15:26: indirectly accessing undefined field unpack of global table
[...]
Checking lib/awful/spawn.lua                      1 warning
    lib/awful/spawn.lua:135:1: line is too long (143 > 120)
[...]
Checking lib/awful/titlebar.lua                   3 warnings
    lib/awful/titlebar.lua:474:1: line is too long (132 > 120)
    lib/awful/titlebar.lua:488:1: line is too long (133 > 120)
    lib/awful/titlebar.lua:496:1: line is too long (136 > 120)
Checking lib/awful/tooltip.lua                    1 warning
    lib/awful/tooltip.lua:56:26: accessing undefined field unpack of global table
Checking lib/awful/util.lua                       1 warning
    lib/awful/util.lua:149:1: line is too long (122 > 120)
[...]
Checking lib/awful/widget/button.lua              1 warning
    lib/awful/widget/button.lua:48:1: line is too long (123 > 120)
[...]

PR #1622, opened 2017-03-03
Fix Luacheck 0.19.0 release Travis breaks

Veratil:
For #1621


lib/awful/spawn.lua
```diff
@@ -1,4 +1,5 @@
 ---------------------------------------------------------------------------
+-- luacheck: no check comment
```
Elv13: That one will end up in the doc, right? If it does, that's not good... I guess the whole file will need to have a luacheck suppression comment...

lib/naughty/core.lua
```diff
@@ -5,6 +5,7 @@
 -- @copyright 2008 koniu
 -- @module naughty
 ----------------------------------------------------------------------------
+-- luacheck: no check comment
```
Elv13: I think this one will show up in the doc too, an extra line should fix that

lib/naughty/dbus.lua
```diff
@@ -105,7 +105,8 @@ local function convert_icon(w, h, rowstride, channels, data)
     return cairo.ImageSurface.create_for_data(tcat(rows), format, w, h, stride)
 end
 
-capi.dbus.connect_signal("org.freedesktop.Notifications", function (data, appname, replaces_id, icon, title, text, actions, hints, expire)
+capi.dbus.connect_signal("org.freedesktop.Notifications",
+function (data, appname, replaces_id, icon, title, text, actions, hints, expire)
```
Elv13: Indentation?

lib/awful/spawn.lua
Veratil: Ah yeah probably so. Could it be placed on top of file?

themes/default/theme.lua
```diff
@@ -1,6 +1,7 @@
 ---------------------------
 -- Default awesome theme --
 ---------------------------
+-- luacheck: no check comment
```
Elv13: I am not sure what to do about that one. I think there is a way to omit some warnings in the `.luacheckrc`, maybe this is a better option for this one? @blueyed: opinion?

Veratil: Correction needed at least:
`--luacheck: no max line length`

Besides that, I don't see a code to ignore line lengths we can set in `.luacheckrc` for the file.

Veratil: Found it, code 631.

Adding to `.luacheckrc`:
```diff
diff --git a/.luacheckrc b/.luacheckrc
index cb3293c..8641cd6 100644
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -13,6 +13,9 @@ files["awesomerc.lua"].allow_defined_top = true
 -- This file itself
 files[".luacheckrc"].ignore = {"111", "112", "131"}
 
+-- Theme files, ignore max line length
+files["themes/*"].ignore = {"631"}
+
 -- Global objects defined by the C code
 read_globals = {
     "awesome",
```

Would that be preferred?

.luacheckrc
```diff
@@ -13,11 +13,13 @@ files["awesomerc.lua"].allow_defined_top = true
 -- This file itself
 files[".luacheckrc"].ignore = {"111", "112", "131"}
 
+-- Theme files, ignore max line length
+files["themes/*"].ignore = {"631"}
```
psychon: Sorry, I don't understand? Instead of this, you want to add a line to themes that except them from all checks?

Yesterday I learnt that the themes in awesome 3.5 set a global variable `theme` and some people made us of this in their config. This was clearly(?) unintended, so luacheck coverage of themes is a good thing.

blueyed: I've meant to explicitly ignore lines that are too long, but I think configuring it like above is better.

Issue #1562, opened 2017-02-14
Declarative guide mistake?


doctaweeks:
It appears there is either a bug in awesome or an error in the declarative guide. Adding a screen conditional as suggested by the guide (s == 1) causes the widget to disappear from every screen. However, using s.index instead (which seems more explicit) works as expected.

awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.0

Elv13:
Indeed. This was valid when added to the guide, but isn't anymore. I added a commit a few days ago that enable unit test to be executed on documentation articles. The other examples can now be ported to the documentation test framework (v4.1 https://awesomewm.org/apidoc/documentation/03-declarative-layout.md.html).

Elv13:
Fixed in 052cda939

PR #1624, opened 2017-03-04
Use only_on_screen widget in docs

psychon:
This is a follow-up to 052cda939b8a5ce1. Expressions like 's.index == 2'
and 's == screen.primary' might work at first. However, they are not
dynamic: If screens are added/removed or the primary screen changes,
then the widget is not updated to follow this.

Instead, use the new awful.widget.only_on_screen container which
provides the needed dynamic behaviour.

To do:

- [x] Bring back the needed trailing whitespace which my editorconfig plugin killed thanks to our `.editorconfigrc`
- [x] Actually test if the example works correctly
- [x] Mention that this is "the real fix" to https://github.com/awesomeWM/awesome/issues/1562 in the commit message
- [x] Find the time to do the above

### DIFF
--- a/docs/03-declarative-layout.md
+++ b/docs/03-declarative-layout.md
@@ -41,8 +41,16 @@ configurable rules.
 Code:
 
     s.mywibox : setup {
-        s == screen.primary and my_first_widget, -- Only display on primary screen
-        s.index == 2 and my_second_widget, -- Only display on screen 2
+        {
+            layout = awful.widget.only_on_screen,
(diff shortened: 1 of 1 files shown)


psychon:
Updated

Elv13:
Also, it would be good/necessary to add a `only_on_screen.lua` file to `tests/examples/wibox/container/defaults/` so it appears on the container list (`scroll` to, btw). As with scroll, it's a bit pointless to have an "example" for something invisible, but not being in the list and being referenced just below is a bit confusing.

PR #1629, opened 2017-03-05
Some small-ish work on gears.geometry

psychon:
possibly WIP since this adds `pending()` calls to the unit tests


Elv13:
Should this be merged now even with the `pending()`? We have too many PRs open. They start to have conflict with each other (obviously not this one, but 30 open PRs is too much).

psychon:
Updated

PR #1631, opened 2017-03-05
Handle missing trailing slash in $XDG_DATA_HOME

majutsushi:


blueyed:
Without it there will be two slashes then though.

Elv13:
Most parsers (as is almost all) handle that internally.

majutsushi:
In fact, this is even specified by POSIX:

http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_266

blueyed:
Well, but why add it on purpose?
I mean in https://github.com/awesomeWM/awesome/pull/1631/files#diff-123b9113b71ac42b91d0bad8c68ddf81L23.

blueyed:
Thanks!

PR #1632, opened 2017-03-06
Fix typo in awful.tag.setmwfact

### LINKED ISSUE #1627: tag.setmwfact() no longer has any effect

```
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1
```

Add this line to `rc.lua`:

    awful.tag.setmwfact(0.2, screen[1].tags[1])

Create windows and see if the master width factor has changed.

**Actual result:**

No change, it doesn't matter what the value is set to, two windows with one master are split 50/50.

**Expected result:**

The master window would be much narrower, as it was in earlier versions of Awesome.

Veratil:
Fix #1627

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -610,7 +610,7 @@ end
 function tag.setmwfact(mwfact, t)
     util.deprecate("Use t.master_width_factor = mwfact instead of awful.tag.setmwfact")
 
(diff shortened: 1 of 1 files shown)


Elv13:
Thanks for looking into this!

PR #1635, opened 2017-03-06
awful.screen.focus_bydirecttion: fix setting screen focus

cmertz:
This was broken in 9cb60b8 in PR #1597 (from myself).

focus() is not defined on the screen instance as method
but on the screen module as function.

Signed-off-by: Christoph Mertz <chris@nimel.de>


cmertz:
Sorry for breaking it ...

I can do a followup PR to add specs/tests for the area in question if that is welcome.

Elv13:
Can you add a regression test for this? There is a `test-screen-changes.lua` integration suit that can be used for this.

PR #1637, opened 2017-03-06
Fix some deprecation warnings

psychon:
This spam while running unit tests was annoying me...

### DIFF
--- a/lib/awful/util.lua
+++ b/lib/awful/util.lua
@@ -512,7 +512,7 @@ function util.table.iterate(t, filter, start)
     return function ()
         while count < length do
             local item = t[index]
-            index = util.cycle(#t, index + 1)
+            index = gmath.cycle(#t, index + 1)
             count = count + 1
             if filter(item) then return item end
         end
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -12,6 +12,7 @@ local spawn = require("awful.spawn")
 local set_shape = require("awful.client.shape").update.all
 local object = require("gears.object")
 local grect = require("gears.geometry").rectangle
+local gmath = require("gears.math")
 local pairs = pairs
 local type = type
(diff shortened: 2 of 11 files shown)


Elv13:
the `gears.table` PR have an extra commit to ensure this never happens again.

lib/awful/util.lua
```diff
@@ -512,7 +512,7 @@ function util.table.iterate(t, filter, start)
     return function ()
         while count < length do
             local item = t[index]
-            index = util.cycle(#t, index + 1)
+            index = gmath.cycle(#t, index + 1)
```
Veratil: FYI: If #1582 gets merged before this one, this will fail.

blueyed: @Veratil 
What do you recommend to to?
Should we get your PRs in first, or do you not mind rebasing them?
Some of them are not green?!

Veratil: @blueyed Either way mine or this PR is merged something has to be changed. I'd like to get the tables PR in first (and @Elv13 added an extra commit for the warnings). After that I'm free to start moving everything over to the new gears modules. Small PRs at a time.

I'm not sure why some aren't green anymore, they were all green when I rebased them to be merged in order. Overall I'm okay with rebasing if necessary. I don't know which would be considered easier though, removing that line from the commit or rebasing the PR. :)

Elv13: But if I merge your PR first, it will breaks the CI. Right now the CI doesn't care about "future deprecation". So this one is safe to merge now. But I can't merge your PR without breaking the CI. I propose merging this PR now. It will be less work for you, not more.

Veratil: Fine by me! Whichever way is best.

Elv13:
/me whoops, had the wrong merge selected, more fixup in the commit log...

PR #1639, opened 2017-03-07
menubar: consider XDG_HOME_DIR and XDG_DATA_DIRS

romildo:
The freedesktop specifications let desktop files to be stored in
different directories indicated by the environment variables
XDG_DATA_HOME and XDG_DATA_DIRS. Only use the default value for these
variables if the variables are not defined.

This is important for systmes like NixOS which does not follow the LFS
and installs files differently.

### DIFF
--- a/lib/gears/table.lua
+++ b/lib/gears/table.lua
@@ -212,5 +212,20 @@ function gtable.merge(t, set)
     return t
 end
 
+--- Map a function to a table. The function is applied to each value
+-- on the table, returning a table of the results.
+-- @class function
+-- @name map
+-- @tparam function f the function to be applied to each value on the table
+-- @tparam table tbl the container table whose values will be operated on
+-- @treturn table Return a table of the results
+function gtable.map(f, tbl)
+    local t = {}
+    for k,v in pairs(tbl) do
+        t[k] = f(v)
+    end
+    return t
+end
+
 
 return gtable
--- a/lib/gears/filesystem.lua
+++ b/lib/gears/filesystem.lua
@@ -6,6 +6,8 @@
 
 -- Grab environment we need
 local Gio = require("lgi").Gio
+local gstring = require("gears.string")
+local gtable = require("gears.table")
 
 local filesystem = {}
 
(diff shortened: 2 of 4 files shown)


Veratil:
Your `split` and `map` functions could be useful outside of this, along with determining the `xdg_data_*` variables.

For modules in PR waiting:
`split` could move to `gears.string`
`map` could move to `gears.table`
`xdg_data_home` and `xdg_data_dirs` could move to functions in `gears.filesystem`

Then add the few requires() and this would be as simple as:
```lua
local gstring = require("gears.string")
menu_gen.all_menu_dirs = gtable.map (
    function(dir) return dir .. '/applications/' end,
    gstring.split(gfs.get_xdg_data_home() .. ':' .. gfs.get_xdg_data_dirs(), ':'))
```
with the added benefit of everybody else getting to use it if they need. :)

Veratil:
@romildo `gears.string` and `gears.table` have been merged if you want to add those functions to their respective modules. `gears.filesystem` is approved and going to be merged soon, too.

Don't forget the documentation! :grin:

romildo:
@Veratil I have reworked the PR so that the helper functions are defined in the appropriate module in the `gears` library.

Elv13:
The CI found 1 problem

```
Checking lib/gears/string.lua                     1 warning

    lib/gears/string.lua:89:38: trailing whitespace in a comment
```

lib/gears/table.lua
```diff
@@ -212,5 +212,20 @@ function gtable.merge(t, set)
...
 end
 
+--- Map a function to a table. The function is applied to each value
+-- on the table, returning a table of the results.
+-- @class function
+-- @name map
+-- @tparam function f the function to be applied to each value on the table
+-- @tparam table tbl the container table whose values will be operated on
+-- @treturn table Return a table of the results
+function gtable.map(f, tbl)
```
Elv13: Could you swap both parameters for consistency with the other function?

Veratil: In defense of the argument order, this is traditionally the way the map function arguments are in all languages.
Python: `map(function_to_apply, list_of_inputs)`
Haskell: `map :: (a -> b) -> [a] -> [b]`
PHP: `array array_map ( callable $callback , array $array1 [, array $... ] )`
Even the [Lua example](http://lua-users.org/wiki/FunctionalLibrary) has `(func, tbl)`

Elv13: ok, then

Elv13:
Merged

PR #1640, opened 2017-03-07
Add some non-helpful examples

psychon:
This makes the scroll and only_on_screen widgets appear in the widget
lists. The examples are not really helpful, but Elv13 told me to add
them. Also, only_on_screen is in awful.widget, but now something in
tests/examples/wibox/container refers to it...


psychon:
Oh and: See https://github.com/awesomeWM/awesome/pull/1624#issuecomment-284171110
foo

Elv13:
I will add a new section for awful widget, eventually

PR #1641, opened 2017-03-07
Update deprecated table calls

Veratil:
Based on #1582 

This updates all the `awful.util.table.*` function calls to the new `gears.table.*` functions to remove all the deprecation messages that will be produced by the base PR. And one change in `awful/button.lua` for a `gears.math.subsets` call.

Quite a few removed the need to `require("awful.util")`.


Elv13:
Here goes the scary one. Squashing to avoid breaking `git-bisect` even more.

PR #1647, opened 2017-03-11
Add a manual (free floating) layout

Elv13:
Was in a "misc" pull request since a while. All that was missing was the doc, so I added it.

### DIFF
--- a/lib/wibox/layout/manual.lua
+++ b/lib/wibox/layout/manual.lua
@@ -1,0 +1,217 @@
+---------------------------------------------------------------------------
+-- A layout with widgets added at specific positions.
+--
+-- Use cases include desktop icons, complex custom composed widgets, a floating
+-- client layout and fine grained control over the output.
+--
+--@DOC_wibox_layout_defaults_manual_EXAMPLE@
+-- @author Emmanuel Lepage Vallee
+-- @copyright 2016 Emmanuel Lepage Vallee
+-- @classmod wibox.layout.manual
+---------------------------------------------------------------------------
+local gtable = require("gears.table")
+local base = require("wibox.widget.base")
+local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
+
+local manual_layout = {}
+
+--- Add some widgets to the given stack layout
+-- @param layout The layout you are modifying.
+-- @tparam widget ... Widgets that should be added
+-- @name add
+-- @class function
+
+--- Remove a widget from the layout
+-- @tparam index The widget index to remove
+-- @treturn boolean index If the operation is successful
+-- @name remove
+-- @class function
+
+--- Insert a new widget in the layout at position `index`
+-- @tparam number index The position
+-- @param widget The widget
+-- @treturn boolean If the operation is successful
+-- @name insert
+-- @class function
+
+--- Remove one or more widgets from the layout
+-- The last parameter can be a boolean, forcing a recursive seach of the
+-- widget(s) to remove.
+-- @param widget ... Widgets that should be removed (must at least be one)
+-- @treturn boolean If the operation is successful
+-- @name remove_widgets
+-- @class function
+
+
+function manual_layout:fit(_, width, height)
+    return width, height
+end
+
+local function geometry(self, new)
+    self._new_geo = new
+    return self._new_geo or self
+end
+
+function manual_layout:layout(context, width, height)
+    local res = {}
+
+    for k, v in ipairs(self._private.widgets) do
+        local pt = self._private.pos[k] or {x=0,y=0}
+        local w, h = base.fit_widget(self, context, v, width, height)
+
+        -- Make sure the signature is compatible with `awful.placement`. `Wibox`,
+        -- doesn't depend on `awful`, but it is still nice not to have to code
+        -- geometry functions again and again.
+        if type(pt) == "function" or (getmetatable(pt) or {}).__call then
+            local geo = {
+                x      = 0,
+                y      = 0,
+                width  = w,
+                height = h,
+                geometry = geometry,
+            }
+            pt = pt(geo, {
+                parent = {
+                    x=0, y=0, width = width, height = height, geometry = geometry
+                }
+            })
+            -- Trick to ensure compatibility with `awful.placement`
+            gtable.crush(pt, geo._new_geo or {})
+        end
+
+        assert(pt.x)
+        assert(pt.y)
+
+        table.insert(res, base.place_widget_at(
+            v, pt.x, pt.y, pt.width or w, pt.height or h
+        ))
+    end
+
+    return res
+end
+
+function manual_layout:add(...)
+    local wdgs = {...}
+    local old_count = #self._private.widgets
+    gtable.merge(self._private.widgets, {...})
+
+    -- Add the points
+    for k, v in ipairs(wdgs) do
+        if v.point then
+            self._private.pos[old_count+k] = v.point
+        end
+    end
+
+    self:emit_signal("widget::layout_changed")
+end
+
+--- Add a widget at a specific point.
+--
+-- The point can either be a function or a table. The table follow the generic
+-- geometry format used elsewhere in Awesome.
+--
+-- * *x*: The horizontal position.
+-- * *y*: The vertical position.
+-- * *width*: The width.
+-- * *height*: The height.
+--
+-- If a function is used, it follows the same prototype as `awful.placement`
+-- functions.
+--
+-- * *geo*:
+--     * *x*: The horizontal position (always 0).
+--     * *y*: The vertical position (always 0).
+--     * *width*: The width.
+--     * *height*: The height.
+--     * *geometry*: A function to get or set the geometry (for compatibility).
+--       The function is compatible with the `awful.placement` prototype.
+-- * *args*:
+--     * *parent* The layout own geometry
+--         * *x*: The horizontal position (always 0).
+--         * *y*: The vertical position (always 0).
+--         * *width*: The width.
+--         * *height*: The height.
+--         * *geometry*: A function to get or set the geometry (for compatibility)
+--           The function is compatible with the `awful.placement` prototype.
+--
+--@DOC_wibox_layout_manual_add_at_EXAMPLE@
+-- @tparam widget widget The widget.
+-- @tparam table|function point Either an `{x=x,y=y}` table or a function
+--  returning the new geometry.
+function manual_layout:add_at(widget, point)
+    assert(not widget.point, "2 points are specified, only one is supported")
+
+    -- Check is the point function is valid
+    if type(point) == "function" or (getmetatable(point) or {}).__call then
+        local fake_geo = {x=0,y=0,width=1,height=1,geometry=geometry}
+        local pt = point(fake_geo, {
+            parent = {
+                x=0, y=0, width = 10, height = 10, geometry = geometry
+            }
+        })
+        assert(pt and pt.x and pt.y, "The point function doesn't seem to be valid")
+    end
+
+    self._private.pos[#self._private.widgets+1] = point
+    self:add(widget)
+end
+
+--- Move a widget (by index).
+-- @tparam number index The widget index.
+-- @tparam table|function point A new point value.
+-- @see add_at
+function manual_layout:move(index, point)
+    assert(self._private.pos[index])
+    self._private.pos[index] = point
+    self:emit_signal( "widget::layout_changed" )
+end
+
+--- Move a widget.
+--
+--@DOC_wibox_layout_manual_move_widget_EXAMPLE@
+-- @tparam widget widget The widget.
+-- @tparam table|function point A new point value.
+-- @see add_at
+function manual_layout:move_widget(widget, point)
+    local idx, l = self:index(widget, false)
+
+    if idx then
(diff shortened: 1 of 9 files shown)


lib/wibox/layout/free.lua
```diff
@@ -0,0 +1,215 @@
...
+                width  = w,
+                height = h,
+                geometry = geometry,
+            }
+            pt = pt(geo, {
+                parent = {
+                    x=0, y=0, width = width, height = height, geometry = geometry
+                }
+            })
+            gtable.crush(pt, geo._new_geo or {})
```
psychon: There's some non-trivial and non-intuitive (and non-documented?) amount of magic here?!?
What's going on?

Elv13: `awful.placement` compat (private API)

lib/wibox/layout/free.lua
```diff
@@ -0,0 +1,215 @@
...
+--     * *geometry*: A function to get or set the geometry (for compatibility)
+-- * *args*:
+--     * *parent* The layout own geometry
+--         * *x*: The horizontal position (always 0).
+--         * *y*: The vertical position (always 0).
+--         * *width*: The width.
+--         * *height*: The height.
+--         * *geometry*: A function to get or set the geometry (for compatibility)
+--
+--@DOC_wibox_layout_free_add_at_EXAMPLE@
```
psychon: I do not really understand the above. If `args` are the arguments, what is `geo`?!?

However, why is the API not just `:add(widget, x, y, width, height)`. Anything involving a callback function means that the result could change over time. Thus, the right signals need to be emitted somewhere.

Put differently, how about the following instead:
```lua
local foo = {}
function foo:layout() -- No point in looking at the arguments, I think
     return foo._private.layout
end
function foo:add(widget, x, y, width, height)
    table.insert(self._private.layout,  base.place_wiget_at(widget, x, y, width, height)
end
```

psychon: (Feel free to use a table instead of individual arguments, if you think that makes a difference)

Elv13:
@blueyed Do you will plan to add fixups? If so, I will wait for yours to avoid useless rebase conflicts. Otherwise please say so, I am fine either way. If you have no time, there is absolutely nothing wrong with that. There is no hurry either, 4.2 is a long way off (unless there is a major problem with 4.1, of course)

blueyed:
@Elv13 
Please go ahead.
I have local changes, but nothing important. I will handle it after your fixes.

lib/wibox/layout/free.lua
```diff
@@ -0,0 +1,215 @@
...
+-- @copyright 2016 Emmanuel Lepage Vallee
+-- @release @AWESOME_VERSION@
+-- @classmod wibox.layout.free
+---------------------------------------------------------------------------
+local gtable = require("gears.table")
+local base = require("wibox.widget.base")
+
+local free_layout = {}
+
+--- Add some widgets to the given stack layout
```
copycat-killer: Maybe he meant "free" as in "freedesktop"? Because that seems rather appropriate to me.

Elv13: It was intended as *free*floating. But it would be used to create desktop icons. It's one of the potential use case.

psychon: How about calling it `freefloating` then? Doesn't make me completely happy, but provides more of a meaning than `free` (Oh, it's free? I do not have to pay?)

Elv13:
Rebased, renamed and fixed

Elv13:
@psychon ping?

psychon:
Whoops.

Ok, high-level view first and I try not to look too much at the code for now:

Use case: You want to be able to arbitrarily place child widgets in a declarative specification, which is why you need this layout (and I guess it might also make sense when clients are layouted via widget layouts).
Specifically, this means that widgets now can have a `point` property.
Also, just a position is not ok, but you also want e.g. "center" or "lower right corner minus (10, 42)". Hence the callback style when adding widgets. And the callbacks also decide about a size.
However, always having to add a callback is cumbersome, so instead of a callback, a table can also be given directly. If only a position and no size is given, then `:fit` is used to come up with a size.
This already leads to some complicated code which I would like to avoid. On top of this, you want to make this compatible with the existing `awful.placement` functions which expect to get clients and more weird/ugly code results.
(Sorry for already complaining so much)

Is this (minus the complaints - sorry) more or less a correct high-level description of this?

My major high-level comment is that the callback functions can do anything. They are not required to yield the same results when called with the same arguments. For example, people might try to move widgets to create an animation. I would very much like to avoid this (or alternatively: Position callbacks are objects which are required to emit some signal when their return value changes).

For example, instead of arbitrary callback functions, can this get a string that is used to index `awful.placement` so that a well-behaved placement function can be assumed? Or is this a direction you definitely do not want to go into.

Elv13:
> Is this (minus the complaints - sorry) more or less a correct high-level description of this?

Rather exact actually

> Use case: You want to be able to arbitrarily place child widgets in a declarative specification,

**Use case 1:**
Desktop wiboxes. Currently doing desktop icons (and yes, a couple of configs I found did that...) require multiple wiboxes and that's slow. Other desktop use case included udisk based devices list of conky like info boxes. It would be cool if those could be managed in a single layout instead of many hardcoded wiboxes. [one](https://github.com/awesomeWM/awesome/issues/1395#issuecomment-292734685) [two](http://i.imgur.com/fGW2TlH.png) [three](https://github.com/awesomeWM/awesome/issues/1395#issuecomment-272591203)

**Use case 2:**
Creating "drag and drop" widgets less or more easily. `QtQuick.Controls2` (the "current" Qt widget toolkit) works this way. Interactive widgets are just compilation of simpler widgets (all the way down). It then uses some anchors, layouts and callbacks to place everything.

**Use case 3:**
The client layouts. Currently its floating layout does nothing and it "just works". I don't plan to change that, too much useless complexity. However, some more advanced floating layouts would be implemented using this and preserve the ability to tab clients in the floating layout.

**Use case 4:**
My "Collision" navigation widgets place a lot of elements on the screen. [One of the more advanced widget](https://github.com/awesomeWM/awesome/issues/1395#issuecomment-273400444) requires some synchronization between all the elements to avoid overlapping. It is already using a hierarchy in the background, but this widget would be much simpler. 

> can this get a string that is used to index awful.placement so that a well-behaved placement
> function can be assumed? Or is this a direction you definitely do not want to go into.

That would not work very well, someone would then complain place1+place2 don't behave as documented. Also, the magic is hidden from view, so it is transparent to the user, why add the "string" complexity.

> (Sorry for already complaining so much)

I understand the worries, but given the size of the code, let alone of the size of the hacks, I don't think it matters much. The code part of the widget is about 50 lines. It's high level API is flexible and hide the complexity **in** the layout rather than expose a confusing and hard to use API. Hiding the placement hack in the code was done on purpose as it simplify the user API. Given that the code has tests, is very small and wont influence the overall maintenance cost over time, I don't think the couple hacks are worth worrying about.

tests/examples/wibox/layout/template.lua
```diff
@@ -22,15 +22,20 @@ local function generic_widget(text, col)
...
+                text   = text or "foobar",
                 widget = wibox.widget.textbox
             } or nil,
             widget = wibox.layout.stack
         },
         margins = 5,
+        set_text = function(self, text2)
+            self:get_children_by_id("text")[1]:set_text(text2)
+        end,
+        is_widget = true,
```
psychon: Huh? This `is_widget` seems fishy to me, why is it needed?
Also, you might want to rebase this on master. Half of the commit is already applied.

psychon: Sorry, ignore the above comment. Obviously, I fail to read diffs.

tests/examples/wibox/layout/defaults/manual.lua
```diff
@@ -0,0 +1,25 @@
+local generic_widget = ... --DOC_HIDE
+local wibox     = require("wibox") --DOC_HIDE
+
+local w1 = generic_widget() --DOC_HIDE
+w1.point  = {x=75,y=5} --DOC_HIDE
+w1.text   = "first" --DOC_HIDE
+w1.forced_width = 50 --DOC_HIDE
```
psychon: Why the many doc hides? It might be confusing enough that `w1` and `w2` come out of nowhere while `third` is created via a call to `generic_widget()` (as the reader of the docs: Whatever that might be). But you also hide the `point` for `w1`.
Should perhaps more of this be done in a declarative way?

tests/examples/wibox/layout/manual/add_at.lua
```diff
@@ -0,0 +1,48 @@
...
+
+local w3 = generic_widget()
+w3.text  = "third"
+w3.point = awful.placement.bottom_right
+l:add(w3)
+
+--
+-- Option 4: Use `:add_at` instead of using the `.point` property. This works
+-- with all 3 ways to define the point.
+-- function.
```
psychon: Can this fourth option to do the same thing be removed, so that only the `.point` property stays available?

lib/wibox/layout/manual.lua
```diff
@@ -0,0 +1,211 @@
...
+--         * *height*: The height.
+--         * *geometry*: A function to get or set the geometry (for compatibility)
+--
+--@DOC_wibox_layout_manual_add_at_EXAMPLE@
+-- @tparam widget widget The widget.
+-- @tparam table|function point Either an `{x=x,y=y}` table or a function
+--  returning the new geometry.
+function manual_layout:add_at(widget, point)
+    self._private.pos[#self._private.widgets+1] = point
+    self:add(widget)
```
psychon: This does not work if `widget.point ~= nil`, because `:add` will overwrite with the point.

psychon: Also, should `:add` and `:add_at` somehow verify the point that they got? E.g. if it is a function, call it with some dummy values to see that it is well-behaved and otherwise throw an error. That's better than the whole widget not being drawn because of an error that made `:layout()` return by error.

lib/wibox/layout/manual.lua
```diff
@@ -0,0 +1,211 @@
...
+                }
+            })
+            -- Trick to ensure compatibility with `awful.placement`
+            gtable.crush(pt, geo._new_geo or {})
+        end
+
+        assert(pt.x)
+        assert(pt.y)
+        assert(pt.width or w)
+        assert(pt.height or h)
```
psychon: The last two asserts do not test anything. `w, h` are set via `base.fit_widget()` and if that returns `nil` or `false`, something is seriously wrong.

lib/wibox/layout/manual.lua
```diff
@@ -0,0 +1,211 @@
...
+end
+
+function manual_layout:get_children()
+    return self._private.widgets
+end
+
+function manual_layout:set_children(children)
+    self:reset()
+
+    self._private.widgets = children
```
psychon: How about `:add(children)` here instead? Otherwise, the caller and this widget now both think they own the `children` table and are allowed to modify it.

lib/wibox/layout/manual.lua
```diff
@@ -0,0 +1,211 @@
...
+        end
+
+        assert(pt.x)
+        assert(pt.y)
+        assert(pt.width or w)
+        assert(pt.height or h)
+
+        table.insert(res, base.place_widget_at(
+            v, pt.x, pt.y, pt.width or w, pt.height or h
+        ))
```
psychon: Random thought (feel free to ignore): Could it be that eventually people also want to be able to do rotations and scaling here? Should there be a way to supply a matrix and a size instead of just position and size?

lib/wibox/layout/manual.lua
```diff
@@ -0,0 +1,211 @@
...
+--     * *width*: The width.
+--     * *height*: The height.
+--     * *geometry*: A function to get or set the geometry (for compatibility)
+-- * *args*:
+--     * *parent* The layout own geometry
+--         * *x*: The horizontal position (always 0).
+--         * *y*: The vertical position (always 0).
+--         * *width*: The width.
+--         * *height*: The height.
+--         * *geometry*: A function to get or set the geometry (for compatibility)
```
psychon: Can you mention that `geo` and `args` are the two arguments to the functions? At first I was confused what this is supposed to be.

lib/wibox/layout/manual.lua
```diff
@@ -0,0 +1,211 @@
...
+--
+--@DOC_wibox_layout_manual_add_at_EXAMPLE@
+-- @tparam widget widget The widget.
+-- @tparam table|function point Either an `{x=x,y=y}` table or a function
```
psychon: Instead of `{x=x,y=y}`, I would say `{x=x,y=y,width=w,height=h}` to highlight that. The default size of "whatever the widget's `:fit` returns for some more-or-less-random arguments" might not be what people usually want.

tests/examples/wibox/layout/manual/add_at.lua
Elv13: That would break the "a widget can be in multiple places" corner code.

lib/wibox/layout/manual.lua
Elv13: Maybe in a future iteration. It makes sense. However none of us have unlimited time to do everything that would make sense.

Elv13:
Rebased, fixed

PR #1648, opened 2017-03-12
ldoc: fixes in lib/gears/color.lua and objects/client.c

copycat-killer:
Fixes a couple of malformed lines `ldoc` was complaining about.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -387,11 +387,11 @@
  *
  * **Signal:**
  *
- *  * *property::icon*
+ *  * *property::icon\_sizes*
  *
  * @property icon_sizes
  * @tparam table sizes
- * @see `get_icon`
+ * @see get_icon
  */
 
 /**
--- a/lib/gears/color.lua
+++ b/lib/gears/color.lua
@@ -333,7 +333,7 @@ function color.recolor_image(image, new_color)
 end
 
 --- Get a valid color for Pango markup
--- @param color The color.
+-- @param check_color The color to check.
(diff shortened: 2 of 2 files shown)


Elv13:
Thanks

objects/client.c
```diff
@@ -387,11 +387,11 @@
  *
  * **Signal:**
  *
- *  * *property::icon*
+ *  * *property::icon\_sizes*
```
psychon: This is actually incorrect, but I agree that it should be this way. Anyone wants to submit a PR with the following?
```diff
diff --git a/objects/client.c b/objects/client.c
index 3adfe3ac8..297971136 100644
--- a/objects/client.c
+++ b/objects/client.c
@@ -2344,6 +2344,7 @@ client_set_icons(client_t *c, cairo_surface_array_t array)
     lua_State *L = globalconf_get_lua_State();
     luaA_object_push(L, c);
     luaA_object_emit_signal(L, -1, "property::icon", 0);
+    luaA_object_emit_signal(L, -1, "property::icon_sizes", 0);
     lua_pop(L, 1);
 }
 
```

copycat-killer: [Done](https://github.com/awesomeWM/awesome/pull/1654).

What about `property::icon_name`?

psychon: Thanks!

What about it? `c.icon_name` is unrelated to `c.icon`.

The "icon" in "icon_name" refers to "iconified" which basically means "minimized". So `c.name` is the normal name and `c.icon_name` is the name while iconified.

PR #1653, opened 2017-03-13
Added a scale factor to gears.wallpaper.centered

bioxz:
I've added a scale factor to gears.wallpaper.centered to scale
the wallpaper the be centered and a small check for valid values.

### DIFF
--- a/lib/gears/wallpaper.lua
+++ b/lib/gears/wallpaper.lua
@@ -117,7 +117,7 @@ function wallpaper.centered(surf, s, background, scale)
     background = color(background)
 
     -- Set default scale if unset
-    if not scale then
+    if not scale or scale <= 0 then
         scale = 1
     end
 
--- a/lib/gears/wallpaper.lua
+++ b/lib/gears/wallpaper.lua
@@ -108,12 +108,18 @@ end
 --   all screens are set.
 -- @param background The background color that should be used. Gets handled via
 --   gears.color. The default is black.
+-- @param scale The scale factor for the wallpaper. Default is 1 (original size).
 -- @see gears.color
-function wallpaper.centered(surf, s, background)
+function wallpaper.centered(surf, s, background, scale)
     local geom, cr = wallpaper.prepare_context(s)
     local original_surf = surf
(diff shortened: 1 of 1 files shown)


lib/gears/wallpaper.lua
```diff
@@ -108,23 +108,32 @@ end
...
 
     -- Now center the surface
     local w, h = surface.get_size(surf)
-    cr:translate((geom.width - w) / 2, (geom.height - h) / 2)
-    cr:rectangle(0, 0, w, h)
+    cr:translate((geom.width - (w * scale)) / 2, (geom.height - (h * scale)) / 2)
+    cr:rectangle(0, 0, (w * scale), (h * scale))
+
     cr:clip()
+    cr:scale(scale, scale)
```
blueyed: Should this be skipped for scale 1?

Veratil: I took a scroll through the cairo source (which is confusing as hell btw), and from what I can find it only checks to see if scale == 0 or is infinite. Checking if scale == 1 would save a bit of processing in the normal case it seems.

psychon: Guys, don't exaggerate. Even if this call would take 1 millisecond (it does not), how often do you have to set a wallpaper before that becomes noticable? Even if this would then save one second of time, I bet setting the wallpaper so often takes quite a lot longer just because so much data has to be transfered back and forth.

We also do not skip the call to `cr:transform(matrix)` in `wibox.hierarchy:draw` if the `matrix` is the identity matrix, for example.

> I took a scroll through the cairo source (which is confusing as hell btw),

Well, "flexible". :-)
For `cairo_scale`, you need to look at `_cairo_default_context_scale` (I think, I haven't checked right now). I bet this function uses `cairo_matrix_init_scale` to initialise a suitable scale matrix and then does the equivalent of `cairo_translate()` to apply it.

Anyway: From cairo's POV a call to `cairo_scale(cr, 1, 1)` should be fast and have no effect.
The only reason that this might be a bit slow here would be LGI.

blueyed:
Thanks!

PR #1656, opened 2017-03-15
Move awful.util.deprecate and awful.util.deprecate_class to gears.debug

Veratil:
Change all awful.util function calls to gears.debug function calls

For #1549 

Note: It's fun to see deprecate is deprecated. :grin:


Elv13:
I don't know how advanced your script (or will power) is, but could you also add the deprecated_in=4 or whatever it ended up being called in this PR (while changing all those lines anyway)

Veratil:
Done.

PR #1674, opened 2017-03-25
Add null check to set imagebox to blank state

firefish5000:


Elv13:
Added a totally untested test copy/pasted from `textbox_spec.lua` and modified in the GitHub UI.

PR #1677, opened 2017-03-26
parse_desktop_file: Trim lines before parsing

hexchain:
`.desktop` files may have CRLF endings (such as Vuze), so it's better to trim the line before parsing.

### DIFF
--- a/lib/menubar/menu_gen.lua
+++ b/lib/menubar/menu_gen.lua
@@ -13,7 +13,6 @@ local utils = require("menubar.utils")
 local icon_theme = require("menubar.icon_theme")
 local pairs = pairs
 local ipairs = ipairs
-local string = string
 local table = table
 
 local menu_gen = {}
@@ -75,16 +74,6 @@ local function get_category_name_and_usage_by_type(app_type)
     end
 end
 
---- Remove CR\LF newline from the end of the string.
--- @param s string to trim
-local function trim(s)
-    if not s then return end
-    if string.byte(s, #s) == 13 then
-        return string.sub(s, 1, #s - 1)
-    end
-    return s
-end
-
 --- Generate an array of all visible menu entries.
 -- @tparam function callback Will be fired when all menu entries were parsed
 -- with the resulting list of menu entries as argument.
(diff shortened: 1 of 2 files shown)


lib/menubar/menu_gen.lua
```diff
@@ -71,16 +71,6 @@ local function get_category_name_and_usage_by_type(app_type)
...
---- Remove CR\LF newline from the end of the string.
--- @param s string to trim
-local function trim(s)
-    if not s then return end
-    if string.byte(s, #s) == 13 then
-        return string.sub(s, 1, #s - 1)
-    end
-    return s
-end
-
```
blueyed: `string` is now unused:

> lib/menubar/menu_gen.lua:14:7: unused variable string

lib/menubar/utils.lua
```diff
@@ -119,6 +119,16 @@ local function get_icon_lookup_path()
     return icon_lookup_path
 end
 
+--- Remove CR\LF newline from the end of the string.
+-- @param s string to trim
+function utils.trim(s)
+    if not s then return end
+    if string.byte(s, #s) == 13 then
+        return string.sub(s, 1, #s - 1)
+    end
```
blueyed: This only removes `CR`, so the doc should be updated.

lib/menubar/menu_gen.lua
hexchain: Oops, my bad. Done.

lib/menubar/utils.lua
hexchain: Done.

hexchain:
I'm wondering, do we need to make `trim` a real trim function (like `str.strip` in Python), or just keep it as is?

blueyed:
I think it can be enhanced later.

blueyed:
Maybe rename it to `rtrim` though for now?

I am using this in my rc.lua:
```lua
--- Remove trailing whitespace from string.
local rtrim = function(s)
  return s:gsub("%s*$", "")
end
```

Veratil:
@blueyed In efforts to reduce, reuse, and recycle, should we provide trim functions in `gears.string`? Do you think `gsub` is faster than `[g]match`?

hexchain:
Just came across [this](http://lua-users.org/wiki/StringTrim).

Veratil:
> Just came across this.

Yep. I think trim7 is the best choice based on timings (without going to C code obviously), but curious about gsub/gmatch viability (I haven't tested it). We could have a full trim, ltrim, and rtrim depending on needs.

hexchain:
Renamed to `rtrim`.

Elv13:
@blueyed ping, you started to review this, there has been changes since that, can you do the follow up? Thanks

blueyed:
Merged.
Thanks for the PR and pinging me about it.

Issue #1494, opened 2017-01-30
Using xmodmap prevent awesome from receiving (at least) mod4-combinations for about a second per event


juhanipaasonen:
Output of `awesome --version`:

awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1

**How to reproduce the issue:**

Run an xmodmap command, e.g.
`$ xmodmap -e "keycode  49 = grave             asciitilde          dead_tilde              dead_caron"`

**Actual result:**

xmodmap returns immediately. For a bit less than a second awesome does not receive (at least) commands sent with mod4, such as mod4+number to switch workspace. The delay increases linearly with the number of events run.

There is a [backtrace](https://gist.github.com/Elv13/1ba6b5204a98f13248a7f7eb7eaa079e) available.

**Expected result:**

awesome should receive commands all the time.

psychon:
Just since no one said so yet: I cannot reproduce this.

For the backtrace: Awesome is handling the event telling it that the keymap changed and now seems to be querying the new keymap from the server (`xcb_connection_has_error()` cannot block).

Elv13:
I can reproduce. Here's my (messy) Xmodmap file

(setxkbmap us) + 

```
keycode 118 = exclam Insert
keycode 107 = parenleft
keycode 78 = parenright
keycode 127 = quotedbl
!;#keycode 77 = dollar
!;#keycode 82 = ampersand
!;#keycode 109 = braceright
!;#keycode 117 = braceleft
!;#keycode 116 = greater greater
!;#keycode 113 = less less
keycode 61 = slash question slash question eacute Eacute
keycode 48 = apostrophe quotedbl apostrophe quotedbl egrave Egrave
keycode 35 = bracketright braceright bracketright braceright ccedilla Ccedilla
keycode 51 = backslash bar backslash bar agrave Agrave
keycode 136 = F13
keycode 220 = F14
keycode 209 = F15
keycode 210 = F16
keycode 143 = F17
keycode 153 = XF86AudioNext
keycode 176 = XF86AudioPrev
remove Lock = Caps_Lock

! Move caps lock to use shift to prevent accidental trigger
clear Lock
!keycode 66  = Shift_L Caps_Lock
keycode 66  = F22 Caps_Lock

[...]

PR #1679, opened 2017-03-28
Asynchronously update the keyboard state

psychon:
When the keyboard layout is modified via xmodmap, each single "change"
(line of input to xmodmap) causes an "the keyboard configuration
changed"-event to be sent. Awesome reacted to each of these events by
reloading the keyboard layout. Thus, awesome reloaded the keyboard
layout a lot and appeared to freeze.

Fix this by asynchronously update the keyboard state: When such an event
comes in, instead of reloading things immediately, we set a flag which
makes us update the state at the end of the main loop iteration. This
means that many events still cause only a single (or at least few)
re-quering of the layout. Thus, a lot of time is saved.

This commit removes the argument to the (undocumented!) signal
xkb::group_changed. Previously, the argument was the active group
number. Since this argument was unused and I'm lazy, I just removed it.
The alternative would be that it might be visible to Lua that some "the
active group changed"-events are dropped.

Fixes: https://github.com/awesomeWM/awesome/issues/1494

P.S.: The new test takes 20 seconds here without the fix. Since the test runner does not allow a test step to take that long, the test fails. Thus, this new test is testing "is awesome finishing in the time allowed by the test runner?".

### DIFF
--- a/xkb.c
+++ b/xkb.c
@@ -297,13 +297,29 @@ xkb_reload_keymap(void)
     }
 }
 
+void
+xkb_refresh(void)
+{
+    lua_State *L = globalconf_get_lua_State();
+
+    if (globalconf.xkb_reload_keymap)
+        xkb_reload_keymap();
+    if (globalconf.xkb_map_changed)
+          signal_object_emit(L, &global_signals, "xkb::map_changed", 0);
+    if (globalconf.xkb_group_changed)
+          signal_object_emit(L, &global_signals, "xkb::group_changed", 0);
+
+    globalconf.xkb_reload_keymap = false;
+    globalconf.xkb_map_changed = false;
+    globalconf.xkb_group_changed = false;
+}
+
 /** The xkb notify event handler.
  * \param event The event.
  */
 void
 event_handle_xkb_notify(xcb_generic_event_t* event)
 {
-    lua_State *L = globalconf_get_lua_State();
     assert(globalconf.have_xkb);
 
     /* The pad0 field of xcb_generic_event_t contains the event sub-type,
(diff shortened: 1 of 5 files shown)


psychon:
> Gio.UnixOutputStream: no `write_all_async'

Sigh. Adding "WIP" to the title...

psychon:
Using `awk`, I found a way around that LGI problem. Please review and merge.

xkb.c
```diff
@@ -297,13 +297,29 @@ xkb_reload_keymap(void)
...
+    if (globalconf.xkb_reload_keymap)
+        xkb_reload_keymap();
+    if (globalconf.xkb_map_changed)
+          signal_object_emit(L, &global_signals, "xkb::map_changed", 0);
+    if (globalconf.xkb_group_changed)
+          signal_object_emit(L, &global_signals, "xkb::group_changed", 0);
+
+    globalconf.xkb_reload_keymap = false;
+    globalconf.xkb_map_changed = false;
+    globalconf.xkb_group_changed = false;
```
Elv13: The inner C dev in me say this should be using enum flags instead of 3 booleans. This would be faster because you could do `if (!xkb_reload_component) return;` and shorter because the reset line would be `xkb_reload_component = XKB_RELOAD_NONE;`. But as you would say, this is bikeshedding. Feel free to ignore this. Thanks for the patch.

psychon: > This would be faster because you could do if (!xkb_reload_component) return;

Code optimisation is GCC's job. Also, a cache miss will fetch the whole cache line and there is a high probability that these three booleans will end up in the same cache line. So, sorry, but I do not let "faster" count as a reason here.

> But as you would say, this is bikeshedding. 

Thanks anyway. Sorry for disagreeing with so many of your optimisations...

Also: Wow, you are not complaining about breaking some undocumented API? (One of the signal loses its argument)

Elv13:
> Also: Wow, you are not complaining about breaking some undocumented API? (One of the signal loses its argument)

`undocumented` is the word.

PR #1682, opened 2017-03-29
Speed up tests/run.sh on Travis

psychon:
tests/run.sh waits for awesome to exit via tail's "--pid" option. This
makes tail check once per second if the process still exists and if not,
tail will exit. However, the default of "once per second" causes lots of
wasted time for us.

This commit adds the argument "-s 0.1" to tail which makes tail check
once every 0.1 seconds. This commit changes the time that "tests/run.sh"
runs on Travis from about 60 seconds to about 35 seconds, which is a big
improvement.

Closes: https://github.com/awesomeWM/awesome/issues/1374

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -210,7 +210,7 @@ for f in $tests; do
     DISPLAY=$D "$AWESOME_CLIENT" 2>&1 < "$f"
 
     # Tail the log and quit, when awesome quits.
-    tail -n 100000 -f --pid "$awesome_pid" "$awesome_log"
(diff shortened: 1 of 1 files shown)


psychon:
The latest master build on Travis needed 8:42/3:50/3:18/4:02/7:43 (each entry is the time in minutes:seconds that one of the five builds took). The Travis build for this PR took 8:01/3:26/2:54/3:34/7:46. That's 41/24/68/-3 seconds faster. (Dunno why Lua 5.3 became slower, but I would guess that this is just a coincidence)

blueyed:
Thanks! :)

PR #1687, opened 2017-03-30
#1685 fix

### LINKED ISSUE #1685: windows spawning on the wrong screen/moving to another screen after layout change

awesome --version
awesome v4.1-19-gb81b214-dirty (Technologic)
 • Compiled against Lua 5.3.1 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.4
 • LGI version: 0.9.1

I have two monitors configuration with default rc.lua. When i trying to open window on the right monitor, it appears on the first screen (not always). If window is already on the second monitor, it moves to first one when i changing layout to tile.  

I found that the issue is in `client_resize` function. `display_area_get` returns geometry for first screen only (not sure if it should return all screens geometry together, or only current screen). I patched it to get current screen geometry, so it works for me now. You can find patch here: [https://pastebin.com/dyRYkngQ](https://pastebin.com/dyRYkngQ)

paul-axe:
#1685 fix

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1741,19 +1741,17 @@ client_resize_do(client_t *c, area_t geometry)
 bool
 client_resize(client_t *c, area_t geometry, bool honor_hints)
 {
-    area_t area;
-
     /* offscreen appearance fixes */
-    area = display_area_get();
+    area_t area = c->screen->geometry;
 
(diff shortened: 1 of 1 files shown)


psychon:
Superseded by #1780.

PR #1689, opened 2017-03-30
Add calendar widgets (wibox and popup)

getzze:
There are two widgets: 
* wibox.widget displaying a month or a year.
The output is very similar to the UNIX `cal` command.
The code is based on the calendar implementations from blingbling and lain and it uses `wibox.widget.grid`. This widget doesn't have many properties, but it provides a wrapper property (`fn_embed`) to tweak all the individual cells.

* awful.widget is a wibox to display the calendar (year or month) as a popup. It has a lot of properties (too many maybe) but it is usable directly in `rc.lua`. Maybe this is too complicated to maintain.

### DIFF
--- a/lib/wibox/widget/calendar.lua
+++ b/lib/wibox/widget/calendar.lua
@@ -1,0 +1,385 @@
+---------------------------------------------------------------------------
+-- A calendar widget
+--
+-- This module defines two widgets: a month calendar and a year calendar
+--
+-- The two widgets have a `date` property, in the form of
+-- a table {day=[number|nil], month=[number|nil], year=[number]}.
+--
+-- The `year` widget displays the whole specified year, e.g. {year=2006}.
+--
+-- The `month` widget displays the calendar for the specified month, e.g. {month=12, year=2006},
+-- highlighting the specified day if the day is provided in the date, e.g. {day=22, month=12, year=2006}.
+--
+-- Cell and container styles can be overridden using the `fn_embed` callback function
+-- which is called before adding the widgets to the layouts. The `fn_embed` function
+-- takes three arguments, the original widget, the flag (`string` used to identified the widget)
+-- and the date (`table`).
+-- It returns another widget, embedding (and modifying) the original widget.
+--
+--@DOC_wibox_widget_defaults_calendar_EXAMPLE@
+--
+-- @author getzze
+-- @copyright 2017 getzze
+-- @classmod wibox.widget.calendar
+---------------------------------------------------------------------------
+
+
+local setmetatable = setmetatable
+local string = string
+local gtable = require("gears.table")
+local vertical = require("wibox.layout.fixed").vertical
+local grid = require("wibox.layout.grid")
+local textbox = require("wibox.widget.textbox")
+local bgcontainer = require("wibox.container.background")
+local base = require("wibox.widget.base")
+local beautiful = require("beautiful")
+
+local calendar = { mt = {} }
+
+local properties = { "date", "font", "spacing", "week_numbers", "start_sunday", "long_weekdays", "fn_embed" }
+
+
+--- The calendar font.
+-- @beautiful beautiful.calendar_font
+-- @tparam string font Font of the calendar
+
+--- The calendar spacing.
+-- @beautiful beautiful.calendar_spacing
+-- @tparam number spacing Spacing of the grid (twice this value for inter-month spacing)
+
+--- Display the calendar week numbers.
+-- @beautiful beautiful.calendar_week_numbers
+-- @param boolean Display week numbers
+
+--- Start the week on Sunday.
+-- @beautiful beautiful.calendar_start_sunday
+-- @param boolean Start the week on Sunday
+
+--- Format the weekdays with three characters instead of two
+-- @beautiful beautiful.calendar_long_weekdays
+-- @param boolean Use three characters for the weekdays instead of two
+
+--- The calendar date.
+--
+-- A table representing the date {day=[number|nil], month=[number|nil], year=[number]}.
+--
+-- E.g.. {day=21, month=2, year=2005}, {month=2, year=2005}, {year=2005}
+-- @tparam date table Date table.
+-- @tparam number date.year Date year
+-- @tparam number|nil date.month Date month
+-- @tparam number|nil date.day Date day
+-- @property date
+
+--- The calendar font.
+--
+-- Choose a monospace font for a better rendering.
+--@DOC_wibox_widget_calendar_font_EXAMPLE@
+-- @param[opt="Monospace 10"] string Font of the calendar
+-- @property font
+
+--- The calendar spacing.
+--
+-- The spacing between cells in the month.
+-- The spacing between months in a year calendar is twice this value.
+-- @param[opt=5] number Spacing of the grid
+-- @property spacing
+
+--- Display the calendar week numbers.
+--
+--@DOC_wibox_widget_calendar_week_numbers_EXAMPLE@
+-- @param[opt=false] boolean Display week numbers
+-- @property week_numbers
+
+--- Start the week on Sunday.
+--
+--@DOC_wibox_widget_calendar_start_sunday_EXAMPLE@
+-- @param[opt=false] boolean Start the week on Sunday
+-- @property start_sunday
+
+--- Format the weekdays with three characters instead of two
+--
+--@DOC_wibox_widget_calendar_long_weekdays_EXAMPLE@
+-- @param[opt=false] boolean Use three characters for the weekdays instead of two
+-- @property long_weekdays
+
+--- The widget encapsulating function.
+--
+-- Function that takes a widget, flag (`string`) and date (`table`) as argument
+-- and returns a widget encapsulating the input widget.
+--
+-- Default value: function (widget, flag, date) return widget end
+--
+-- It is used to add a container to the grid layout and to the cells:
+--
+--@DOC_wibox_widget_calendar_fn_embed_cell_EXAMPLE@
+-- @param function Function to embed the widget depending on its flag
+-- @property fn_embed
+
+
+--- Make a textbox
+-- @tparam string text Text of the textbox
+-- @tparam string font Font of the text
+-- @tparam boolean center Center the text horizontally
+-- @treturn wibox.widget.textbox
+local function make_cell(text, font, center)
+    return base.make_widget_declarative {
+        markup = text,
+        align  = center and "center" or "right",
+        valign = 'center',
+        font   = font,
+        widget = textbox
+    }
+end
+
+--- Create a grid layout with the month calendar
+-- @tparam table props Table of calendar properties
+-- @tparam table date Date table
+-- @tparam number date.year Date year
+-- @tparam number date.month Date month
+-- @tparam number|nil date.day Date day
+-- @treturn widget Grid layout
+local function create_month(props, date)
+    local num_rows    = 8
+    local num_columns = props.week_numbers and 8 or 7
+
+    -- Create grid layout
+    local layout = base.make_widget_declarative{
+        layout      = grid,
+        expand      = true,
+        homogeneous = true,
+        spacing     = props.spacing,
+        forced_num_rows = num_rows,
+        forced_num_cols = num_columns,
+    }
+
+    local start_row    = 3
+    local start_column = num_columns - 6
+    local week_start   = props.start_sunday and 1 or 2
+    local last_day     = os.date("*t", os.time{year=date.year, month=date.month+1, day=0})
+    local month_days   = last_day.day
+    local column_fday  = (last_day.wday - month_days + 1 - week_start ) % 7
+
+    --local flags = {"header", "weekdays", "weeknumber", "normal", "focus"}
+    local cell_date, t, i, j, w, flag, text
+
+    -- Header
+    flag = "header"
+    t = os.time{year=date.year, month=date.month, day=1}
+    if props.subtype=="monthheader" then
+        flag = "monthheader"
+        text = os.date("%B", t)
+    else
+        text = os.date("%B %Y", t)
+    end
+    w = props.fn_embed(make_cell(text, props.font, true), flag, date)
+    layout:add_widget_at(w, 1, 1, 1, num_columns)
+
+    -- Days
+    i = start_row
+    j = column_fday + start_column
+    local current_week = nil
+    local drawn_weekdays = 0
+    for d=1, month_days do
+        cell_date = {year=date.year, month=date.month, day=d}
+        t = os.time(cell_date)
+        -- Week number
+        if props.week_numbers then
+            text = os.date("%V", t)
+            if tonumber(text) ~= current_week then
+                flag = "weeknumber"
+                w = props.fn_embed(make_cell(text, props.font), flag, cell_date)
+                layout:add_widget_at(w, i, 1, 1, 1)
+                current_week = tonumber(text)
+            end
+        end
+        -- Week days
+        if drawn_weekdays < 7 then
+            flag = "weekday"
+            text = os.date("%a", t)
+            if not props.long_weekdays then
+                text = string.sub(text, 1, 2)
+            end
+            w = props.fn_embed(make_cell(text, props.font), flag, cell_date)
+            layout:add_widget_at(w, 2, j, 1, 1)
+            drawn_weekdays = drawn_weekdays +1
+        end
+        -- Normal day
+        flag = "normal"
+        text = string.format("%2d", d)
+        -- Focus day
+        if date.day == d then
+            flag = "focus"
+            text = "<b>"..text.."</b>"
+        end
+        w = props.fn_embed(make_cell(text, props.font), flag, cell_date)
+        layout:add_widget_at(w, i, j, 1, 1)
+
+        -- find next cell
+        i,j = layout:get_next_empty(i,j)
+        if j < start_column then j = start_column end
+    end
+    return props.fn_embed(layout, "month", date)
+end
+
+
+--- Create a grid layout for the year calendar
+-- @tparam table props Table of year calendar properties
+-- @param date Year to display (number or string)
+-- @treturn widget Grid layout
+local function create_year(props, date)
+    -- Create a grid widget with the 12 months
+    local in_layout = base.make_widget_declarative{
+        layout      = grid,
+        expand      = true,
+        homogeneous = true,
+        spacing     = 2*props.spacing,
+        forced_num_cols = 4,
+        forced_num_rows = 3,
+    }
+
+    local month_date
+    local current_date = os.date("*t")
+
+    for month=1,12 do
+        if date.year == current_date.year and month == current_date.month then
+            month_date = {day=current_date.day, month=current_date.month, year=current_date.year}
+        else
+            month_date = {month=month, year=date.year}
+        end
+        in_layout:add(create_month(props, month_date))
+    end
+
+    -- Create a vertical layout
+    local flag, text = "yearheader", string.format("%s", date.year)
+    local year_header = props.fn_embed(make_cell(text, props.font, true), flag, date)
+    local out_layout = base.make_widget_declarative{
+        year_header,
+        in_layout,
+        spacing = 2*props.spacing, -- separate header from calendar grid
+        layout  = vertical
+    }
+    return props.fn_embed(out_layout, "year", date)
+end
+
+
+--- Set the container to the current date
+-- @param self Widget to update
+local function fill_container(self)
+    local date = self._private.date
+    if date then
+        -- Create calendar grid
+        if self._private.type == "month" then
+            self._private.container:set_widget(create_month(self._private, date))
+        elseif self._private.type == "year" then
+            self._private.container:set_widget(create_year(self._private, date))
+        end
+    else
+        self._private.container:set_widget(nil)
+    end
+    self:emit_signal("widget::layout_changed")
+end
+
+
+-- Set the calendar date
+function calendar:set_date(date)
+    if date ~= self._private.date then
+        self._private.date = date
+        -- (Re)create calendar grid
+        fill_container(self)
+    end
+end
+
+
+-- Build properties function
+for _, prop in ipairs(properties) do
+    -- setter
+    if not calendar["set_" .. prop] then
+        calendar["set_" .. prop] = function(self, value)
+            if (string.sub(prop,1,3)=="fn_" and type(value) == "function") or self._private[prop] ~= value then
+                self._private[prop] = value
+                -- (Re)create calendar grid
+                fill_container(self)
+            end
+        end
+    end
+    -- getter
+    if not calendar["get_" .. prop] then
+        calendar["get_" .. prop] = function(self)
+            return self._private[prop]
+        end
+    end
+end
+
+
+--- Return a new calendar widget by type.
+--
+-- @tparam string type Type of the calendar, `year` or `month`
+-- @tparam table date Date of the calendar
+-- @tparam number date.year Date year
+-- @tparam number|nil date.month Date month
+-- @tparam number|nil date.day Date day
+-- @tparam[opt="Monospace 10"] string font Font of the calendar
+-- @treturn widget The calendar widget
+local function get_calendar(type, date, font)
+    local ct = bgcontainer()
+    local ret = base.make_widget(ct, "calendar", {enable_properties = true})
+    gtable.crush(ret, calendar, true)
+
+    ret._private.type = type
+    ret._private.container = ct
+
+    -- default values
+    ret._private.date = date
+    ret._private.font = font or beautiful.calendar_font or "Monospace 10"
+
+    ret._private.spacing       = beautiful.calendar_spacing or 5
+    ret._private.week_numbers  = beautiful.calendar_week_numbers or false
+    ret._private.start_sunday  = beautiful.calendar_start_sunday or false
+    ret._private.long_weekdays = beautiful.calendar_long_weekdays or false
+    ret._private.fn_embed      = function (w, _) return w end
+
+    -- header specific
+    ret._private.subtype = type=="year" and "monthheader" or "fullheader"
+
+    fill_container(ret)
+    return ret
+end
+
+--- A month calendar widget.
+--
+-- A calendar widget is a grid containing the calendar for one month.
+-- If the day is specified in the date, its cell is highlighted.
+--
+--@DOC_wibox_widget_calendar_month_EXAMPLE@
+-- @tparam table date Date of the calendar
+-- @tparam number date.year Date year
+-- @tparam number date.month Date month
+-- @tparam number|nil date.day Date day
+-- @tparam[opt="Monospace 10"] string font Font of the calendar
+-- @treturn widget The month calendar widget
+-- @function wibox.widget.calendar.month
+function calendar.month(date, font)
+    return get_calendar("month", date, font)
+end
+
+--- A year calendar widget.
+--
+-- A calendar widget is a grid containing the calendar for one year.
+--
+--@DOC_wibox_widget_calendar_year_EXAMPLE@
+-- @tparam table date Date of the calendar
+-- @tparam number date.year Date year
+-- @tparam number|nil date.month Date month
+-- @tparam number|nil date.day Date day
+-- @tparam[opt="Monospace 10"] string font Font of the calendar
+-- @treturn widget The year calendar widget
+-- @function wibox.widget.calendar.year
+function calendar.year(date, font)
+    return get_calendar("year", date, font)
+end
+
+
+return setmetatable(calendar, calendar.mt)
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/tests/examples/wibox/widget/calendar/fn_embed_cell.lua
+++ b/tests/examples/wibox/widget/calendar/fn_embed_cell.lua
@@ -1,0 +1,84 @@
+local parent    = ... --DOC_HIDE
+local wibox     = require("wibox") --DOC_HIDE
+local gears     = require("gears") --DOC_HIDE
+local beautiful = require( "beautiful" ) --DOC_HIDE
+local Pango     = require("lgi").Pango --DOC_HIDE
+
+-- Beautiful fake get_font --DOC_HIDE
+local f = Pango.FontDescription.from_string("monospace 10") --DOC_HIDE
+beautiful.get_font = function() return f end --DOC_HIDE
+
+-- Fake beautiful theme --DOC_HIDE
+beautiful.fg_focus = "#ff9800" --DOC_HIDE
+beautiful.bg_focus = "#b9214f" --DOC_HIDE
+
+    local styles = {}
+    local function rounded_shape(size, partial)
+        if partial then
+            return function(cr, width, height)
+                       gears.shape.partially_rounded_rect(cr, width, height,
+                            false, true, false, true, 5)
+                   end
+        else
+            return function(cr, width, height)
+                       gears.shape.rounded_rect(cr, width, height, size)
+                   end
+        end
+    end
+    styles.month   = { padding      = 5,
+                       bg_color     = "#555555",
+                       border_width = 2,
+                       shape        = rounded_shape(10)
+    }
+    styles.normal  = { shape    = rounded_shape(5) }
+    styles.focus   = { fg_color = "#000000",
+                       bg_color = "#ff9800",
+                       markup   = function(t) return '<b>' .. t .. '</b>' end,
+                       shape    = rounded_shape(5, true)
+    }
+    styles.header  = { fg_color = "#de5e1e",
+                       markup   = function(t) return '<b>' .. t .. '</b>' end,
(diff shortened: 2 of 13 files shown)


Elv13:
Oh nice, you will make a lot of people happy with this. I don't have time to do a full review right now, so I will just note some high level stuff.

 * There seems to be an issues found by `luacheck`
 * The "default" screenshot seem to have some extra unused pixels at the bottom
 * There is some line length issues in the code and example. For the grid layout I fixed those myself, but as I said, I currently don't have time for Awesome. Try to make sure the examples never create an horizontal scrollbar and keep the line length close to 80 char.
 * The coverage is already quite good (congratulation for that, this is appreciated), but could be a bit better. An integration test for the `awful.widget.calendar` should bump it in the green. Those are the files in `tests/test-*.lua`. You can run them locally using `run.sh`.
 * Your style system create cool looking widgets, but the API look a bit awkward. Could that callback be converted into a property for each "flag". Some kind of template properties?

I only reviewed some details of the documentation pages generated by your patch. I didn't check the code yet.

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+--    local mytextclock = wibox.widget.textclock()
+--    local month_calendar = calendar.month()
+--    month_calendar:attach(mytextclock, 'tr')
+--
+-- @param widget Widget to attach the calendar
+-- @tparam[opt="tr"] string position Two characters string defining the position on the screen
+-- @treturn wibox The wibox calendar
+function calendar_popup:attach(widget, position)
+    local position = position or "tr"
+    widget:buttons(util.table.join(abutton({ }, 1, function ()
```
Veratil: Change `util.table.join` to `gears.table.join`

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+-- @treturn wibox A wibox containing the calendar
+local function get_cal_wibox(caltype, args)
+    if caltype ~= "month" and caltype ~= "year" then return end
+    local args = args or {}
+
+    local ret = wibox{ ontop   = true,
+                       opacity = args.opacity or 1,
+                       bg      = args.bg
+    }
+    util.table.crush(ret, calendar_popup, false)
```
Veratil: Change `util.table.crush` to `gears.table.crush`

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+        font         = args.font,
+        spacing      = args.spacing,
+        week_numbers = args.week_numbers,
+        start_sunday = args.start_sunday,
+        fn_embed     = embed(parse_all_options(args)),
+        widget = wibox.widget.calendar(caltype)
+    }
+    ret:set_widget(widget)
+
+    ret:buttons(util.table.join(
```
Veratil: Change `util.table.join` to `gears.table.join`

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+
+
+--- A year calendar wibox.
+--
+-- It is highly customizable using the same options as for the widgets.
+-- The options are set once and for all at creation, though.
+--
+--@DOC_wibox_awidget_calendar_year_wibox_EXAMPLE@
+--
+--    globalkeys = awful.util.table.join(globalkeys, awful.key({ modkey, "Control" }, "c",  function () year_calendar:toggle() end))
```
Veratil: Change `awful.util.table.join` to `gears.table.join`

Also luacheck line too long (132 > 120)

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+--
+-- @author getzze
+-- @copyright 2017 getzze
+-- @classmod awful.widget.calendar
+---------------------------------------------------------------------------
+
+local setmetatable = setmetatable
+local string = string
+local ipairs = ipairs
+local util = require("awful.util")
```
Veratil: Remove ipairs and this line. Almost all of `awful.util` is deprecated.

tests/examples/wibox/widget/calendar/week_numbers.lua
```diff
@@ -0,0 +1,22 @@
+local parent    = ... --DOC_HIDE
+local wibox     = require("wibox") --DOC_HIDE
+local beautiful = require( "beautiful" ) --DOC_HIDE
+local lgi = require("lgi") --DOC_HIDE
+local Pango = lgi.Pango --DOC_HIDE
```
Veratil: You can probably consolidate this and the previous line to:
`local Pango = require("lgi").Pango`
since lgi isn't used anywhere except here.

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+-- The calendar style can be tweaked by providing tables of style properties at creation:
+-- `style_year` and `style_month` (see `container_properties`); `style_yearheader`, `style_header`,
+-- `style_weekdays`, `style_weeknumber`, `style_normal`, `style_focus` (see `cell_properties`).
+--
+-- The wibox accepts arguments for the calendar widget: "font", "spacing", "week_numbers", "start_sunday".
+-- It also accepts the extra arguments `opacity`, `bg`, `screen` and `position`.
+-- `opacity` and `bg` apply to the wibox itself, they are mainly useful to manage opacity
+-- by setting `opacity` for the false opacity or setting `bg="#00000000"` for compositor opacity.
+-- The `screen` argument forces the display of the wibox to this screen (instead of the focused screen by default).
+-- The `position` argument is a two-characters string describing the screen alignment "[vertical][horizontal]", e.g. "cc", "tr", "bl", ...
```
Veratil: luacheck line too long (138 > 120)

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+local ascreen = require("awful.screen")
+local abutton = require("awful.button")
+local gears = require("gears")
+local wibox = require("wibox")
+local base = require("wibox.widget.base")
+local beautiful = require("beautiful")
+
+local calendar_popup = { offset = 0, mt = {} }
+
+local properties = { "fg_color", "bg_color", "shape", "padding", "border_width", "border_color", "opacity" }
```
Veratil: luacheck says this is unused

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+    return {x=x, y=y, width=width, height=height}
+end
+
+--- Call the calendar with offset
+-- @tparam number inc_offset Offset with respect to current date
+-- @tparam table props Properties of the calendar (screen, position, ...)
+-- @treturn wibox The wibox calendar
+function calendar_popup:call_calendar(inc_offset, props)
+    local props = props or {}
+    local offset, position, s = inc_offset or 0, props.position or self.position, props.screen or self.screen or ascreen.focused()
```
Veratil: luacheck line too long (130 > 120)

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+    self.position = position  -- remember last position when changing offset
+
+    self.offset = self.offset + offset
+    local is_current = (offset == 0 or self.offset == 0)
+    if is_current then
+        self.offset = 0
+    end
+
+    local widget = self:get_widget()
+    local date, hint_width, hint_height
```
Veratil: hint_width and hint_height are unused

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+--
+--    local mytextclock = wibox.widget.textclock()
+--    local month_calendar = calendar.month()
+--    month_calendar:attach(mytextclock, 'tr')
+--
+-- @param widget Widget to attach the calendar
+-- @tparam[opt="tr"] string position Two characters string defining the position on the screen
+-- @treturn wibox The wibox calendar
+function calendar_popup:attach(widget, position)
+    local position = position or "tr"
```
Veratil: luacheck is complaining about this, forgot what the fix was.

    lib/awful/widget/calendar.lua:271:11: variable position was previously defined as an argument on line 270

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,360 @@
...
+
+--- Return a new calendar wibox by type.
+--
+-- A calendar widget displaying a `month` or a `year`
+-- @tparam string caltype Type of calendar `month` or `year`
+-- @tparam table args Properties of the widget
+-- @treturn wibox A wibox containing the calendar
+local function get_cal_wibox(caltype, args)
+    if caltype ~= "month" and caltype ~= "year" then return end
+    local args = args or {}
```
Veratil: Another luacheck complaint:

    lib/awful/widget/calendar.lua:290:11: variable args was previously defined as an argument on line 288

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+
+
+
+-- Parse a date for the day, month and year
+-- @tparam string date Date in the "%d %m %Y" format
+-- @treturn {number,number,number} List of day, month, year (nil if undefined)
+local function parse_date(date)
+    -- Split a string by a given `separator`
+    local function split(str, sep)
+        local sep, fields = sep or ":", {}
```
Veratil: Another luacheck complaint:

    lib/wibox/widget/calendar.lua:113:15: variable sep was previously defined as an argument on line 112

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+
+    local start_row    = 3
+    local start_column = num_columns - 6
+    local week_start   = props.start_sunday and 1 or 2
+    local d            = os.date("*t", os.time{year=year, month=month+1, day=0})
+    local month_days   = d.day
+    local column_fday  = (d.wday - d.day + 1 - week_start ) % 7
+
+    --local flags = {"header", "weekdays", "weeknumber", "normal", "focus"}
+    local t, i, j, w, flag, text, markup
```
Veratil: luacheck: `markup` is unused

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+    -- Header
+    flag = "header"
+    t = os.time{year=year, month=month, day=1}
+    text = os.date(props.hide_year and "%B" or "%B %Y", t)
+    w = props.fn_embed(make_cell(text, props.font, flag=="yearheader" or flag=="header"), flag)
+    layout:add_widget_at(w, 1, 1, 1, num_columns)
+
+    -- Week days
+    i = 2
+    for j = start_column, num_columns do  -- columns
```
Veratil: luacheck complaint:

    lib/wibox/widget/calendar.lua:184:9: shadowing definition of variable j on line 173

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+        text = string.sub(os.date("%a", t), 1, 2)
+        w = props.fn_embed(make_cell(text, props.font, flag=="yearheader" or flag=="header"), flag)
+        layout:add_widget_at(w, i, j, 1, 1)
+    end
+
+    -- Days
+    i = start_row
+    j = column_fday + start_column
+    local current_week = nil
+    for d=1, month_days do
```
Veratil: luacheck complaint:

    lib/wibox/widget/calendar.lua:196:9: shadowing definition of variable d on line 168

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+    local in_layout = base.make_widget_declarative{
+        layout      = grid,
+        expand      = true,
+        homogeneous = true,
+        spacing     = 2*props.spacing,
+        forced_num_cols = 4,
+        forced_num_rows = 3,
+    }
+
+    local month_date = ""
```
Veratil: luacheck complaint:

    lib/wibox/widget/calendar.lua:241:11: value assigned to variable month_date is unused

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+    end
+end
+
+-- Override container functions
+function calendar:get_children()
+    fill_container(self, false)
+    return self._private.container or {}
+end
+
+function calendar:set_children(...)
```
Veratil: luacheck complaint:

    lib/wibox/widget/calendar.lua:323:32: unused variable length argument

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+end
+
+-- Layout widget
+function calendar:layout(context, width, height)
```
Veratil: luacheck complaint:

    lib/wibox/widget/calendar.lua:327:26: unused argument context

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+
+
+--- Return a new calendar widget by type.
+--
+-- @tparam string t Type of the calendar, `year` or `month`
+-- @tparam string date Date of the calendar
+-- @tparam[opt="Monospace 10"] string font Font of the calendar
+-- @treturn widget The calendar widget
+local function get_calendar(t, date, font)
+    args = args or {}
```
Veratil: luacheck complaint:

    lib/wibox/widget/calendar.lua:352:5: setting non-standard global variable args
    lib/wibox/widget/calendar.lua:352:12: accessing undefined variable args

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+
+    -- default values
+    ret._private.date = date
+    ret._private.font = font or beautiful.calendar_font or "Monospace 10"
+
+    ret._private.spacing      = beautiful.calendar_spacing or 5
+    ret._private.week_numbers = beautiful.calendar_week_numbers or false
+    ret._private.start_sunday = beautiful.calendar_start_sunday or false
+    ret._private.hide_year    = t == "year" and true or false
+    ret._private.fn_embed     = function (w, flag) return w end
```
Veratil: luacheck complaint:

    lib/wibox/widget/calendar.lua:368:46: unused argument flag

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+--
+-- @tparam string t Type of the calendar, `year` or `month`
+-- @tparam string date Date of the calendar
+-- @tparam[opt="Monospace 10"] string font Font of the calendar
+-- @treturn widget The calendar widget
+local function get_calendar(t, date, font)
+    args = args or {}
+    if t ~= "month" and t ~= "year" then return end
+    local ret = base.make_widget(nil, "calendar_"..t, {enable_properties = true})
+    util.table.crush(ret, calendar, true)
```
Veratil: Change `util.table` to `gtable`

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,405 @@
...
+--
+-- @author getzze
+-- @copyright 2017 getzze
+-- @classmod wibox.widget.calendar
+---------------------------------------------------------------------------
+
+local setmetatable = setmetatable
+local string = string
+local ipairs = ipairs
+local util = require("awful.util")
```
Veratil: Change to:
`local gtable = require("gears.table")`

tests/examples/wibox/widget/calendar/month.lua
```diff
@@ -0,0 +1,50 @@
+local parent    = ... --DOC_HIDE_ALL
+local wibox     = require("wibox") --DOC_HIDE
+local beautiful = require( "beautiful" ) --DOC_HIDE
+local lgi = require("lgi") --DOC_HIDE
+local Pango = lgi.Pango --DOC_HIDE
```
Veratil: You can probably consolidate this and the previous line to:
`local Pango = require("lgi").Pango`
since lgi isn't used anywhere except here.

tests/examples/wibox/widget/calendar/year.lua
```diff
@@ -0,0 +1,24 @@
+local parent    = ... --DOC_HIDE
+local wibox     = require("wibox") --DOC_HIDE
+local beautiful = require( "beautiful" ) --DOC_HIDE
+local lgi = require("lgi") --DOC_HIDE
+local Pango = lgi.Pango --DOC_HIDE
```
Veratil: You can probably consolidate this and the previous line to:
`local Pango = require("lgi").Pango`
since lgi isn't used anywhere except here.

tests/examples/wibox/widget/defaults/calendar.lua
```diff
@@ -0,0 +1,17 @@
+local parent    = ... --DOC_HIDE
+local wibox     = require("wibox") --DOC_HIDE
+local beautiful = require( "beautiful" ) --DOC_HIDE
+local lgi = require("lgi") --DOC_HIDE
+local Pango = lgi.Pango --DOC_HIDE
```
Veratil: You can probably consolidate this and the previous line to:
`local Pango = require("lgi").Pango`
since lgi isn't used anywhere except here.

tests/examples/wibox/widget/calendar/font.lua
```diff
@@ -0,0 +1,18 @@
+local parent    = ... --DOC_HIDE
+local wibox     = require("wibox") --DOC_HIDE
+local beautiful = require( "beautiful" ) --DOC_HIDE
+local lgi = require("lgi") --DOC_HIDE
+local Pango = lgi.Pango --DOC_HIDE
```
Veratil: You can probably consolidate this and the previous line to:
`local Pango = require("lgi").Pango`
since lgi isn't used anywhere except here.

tests/examples/wibox/widget/calendar/fn_embed_cell.lua
```diff
@@ -0,0 +1,67 @@
+local parent    = ... --DOC_HIDE
+local wibox     = require("wibox") --DOC_HIDE
+local gears     = require("gears") --DOC_HIDE
+local beautiful = require( "beautiful" ) --DOC_HIDE
+local lgi = require("lgi") --DOC_HIDE
+local Pango = lgi.Pango --DOC_HIDE
```
Veratil: You can probably consolidate this and the previous line to:
`local Pango = require("lgi").Pango`
since lgi isn't used anywhere except here.

lib/awful/widget/calendar.lua
Elv13: This is shadowing the original `position`. Remove the `local` to fix it.

getzze:
I corrected the `luacheck` errors and added a `test-awful-widget-calendar.lua` for the awful.widget.
 @Elv13:
* the calendar has a fixed size of 8 rows, so for some months the bottom row (can even be two rows) is empty.
* The properties are a problem. First, I had put all the properties of the `awful.widget` in the `wibox.widget`, but that is a lot (8 flag x 8 properties ...). So I came up with the `fn_embed` callback which is quite handy because it gives a lot of liberty. However it is slower, displaying the year calendar makes Awesome lag.
If somebody has a suggestion to deal with the properties...

Elv13:
Thanks for the stellar code coverage! For one of the current CI error, you should squash some of your commits back to "add a cal"  and "test the cal". For the other, I didn't investigate, but it looks like it found something:

```
== Running /home/travis/build/awesomeWM/awesome/tests/test-awful-widget-calendar.lua ==

2017-03-31 23:19:25 W: Cannot reliably detect EOF on an GIOInputStream with this LGI version

Error: running function for step 1/5 (@2): [string "--- Test for awful.widget.calendar..."]:43: assertion failed!

stack traceback:

	[C]: in function 'assert'

	[string "--- Test for awful.widget.calendar..."]:43: in function <[string "--- Test for awful.widget.calendar..."]:36>

	(...tail calls...)

	[C]: in function 'xpcall'

	/home/travis/build/awesomeWM/awesome/tests/_runner.lua:48: in function </home/travis/build/awesomeWM/awesome/tests/_runner.lua:42>

	[C]: in function 'xpcall'

	(...tail calls...)

	lib/gears/timer.lua:226: in function <lib/gears/timer.lua:224>!

===> ERROR running /home/travis/build/awesomeWM/awesome/tests/test-awful-widget-calendar.lua <===

Error: running function for step 1/5 (@2): [string "--- Test for awful.widget.calendar..."]:43: assertion failed!
```

lib/awful/widget/calendar.lua
```diff
@@ -0,0 +1,381 @@
...
+--@DOC_wibox_awidget_defaults_calendar_EXAMPLE@
+--
+-- @author getzze
+-- @copyright 2017 getzze
+-- @classmod awful.widget.calendar
+---------------------------------------------------------------------------
+
+local setmetatable = setmetatable
+local string = string
+local gtable = require("gears.table")
```
Veratil: You already import gears below, you could remove this line and expand `gtable` to `gears.table`

Veratil: Actually looking through everything, I think you only use `gears.table` and `gears.shape`.

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,401 @@
...
+
+
+
+
+-- Parse a date for the day, month and year
+-- @tparam string date Date in the "%d %m %Y" format
+-- @treturn {number,number,number} List of day, month, year (nil if undefined)
+local function parse_date(date)
+    -- Split a string by a given `separator`
+    local function split(str, sep)
```
Veratil: [gears.string.split](https://awesomewm.org/apidoc/libraries/gears.string.html#split)

Veratil: As a note, if gsub is faster than gmatch then you could update the [split function here](https://github.com/awesomeWM/awesome/blob/master/lib/gears/string.lua#L95). Just make sure it still functions the same where it's used in `gears.filesystem` and anywhere else.

Elv13:
@actionless @psychon @blueyed Can someone else review this, I really have no time

@getzze Really sorry about the delay here, this is very good work

psychon:
@Elv13 Sorry, but I also seem to fail to find the time to do a review. All I figured out so far is that this does not seem to spawn external processes to do the work (good!) and that there is a split into `wibox.widget.calender` and `awful.widget.calender`, but I haven't even figured out the difference between the two (might make sense to rename one of them so that the name "calender" is unique).

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,401 @@
...
+-- @param function Function to embed the widget depending on its flag
+-- @property fn_embed
+
+
+
+
+-- Parse a date for the day, month and year
+-- @tparam string date Date in the "%d %m %Y" format
+-- @treturn {number,number,number} List of day, month, year (nil if undefined)
+local function parse_date(date)
```
psychon: How about changing the API so that users already provide `day, month, year` so that no such string parsing is needed?

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,401 @@
...
+    t = os.time{year=year, month=month, day=1}
+    text = os.date(props.hide_year and "%B" or "%B %Y", t)
+    w = props.fn_embed(make_cell(text, props.font, flag=="yearheader" or flag=="header"), flag)
+    layout:add_widget_at(w, 1, 1, 1, num_columns)
+
+    -- Week days
+    i = 2
+    for jj = start_column, num_columns do  -- columns
+        flag = "weekdays"
+        t = os.time{year=2006, month=1, day=jj-start_column+week_start}
```
psychon: What's so special about january 2006? Why does this not use the actual year and month that were provided?

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,401 @@
...
+        text = string.sub(os.date("%a", t), 1, 2)
```
psychon: Why do you abbreviate the abbreviated weekday name? AFAIR there are locales where this makes the day non-unique.

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,401 @@
...
+-- @tparam string date Date of the calendar
+-- @tparam[opt="Monospace 10"] string font Font of the calendar
+-- @treturn widget The calendar widget
+local function get_calendar(t, date, font)
+    if t ~= "month" and t ~= "year" then return end
+    local ret = base.make_widget(nil, "calendar_"..t, {enable_properties = true})
+    gtable.crush(ret, calendar, true)
+
+    ret.type = t
+    ret._private.calendar_cache = {}
```
psychon: Why does this have a cache at all? I am not convinced that something that displays a calender has to cache things.
(Also, I don't think that the cache should be keyed via a string, but let's ignore this)

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,401 @@
...
+
+    -- default values
+    ret._private.date = date
+    ret._private.font = font or beautiful.calendar_font or "Monospace 10"
```
psychon: Why the hardcoded `Monospace 10`? Just use `nil` as the default.

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,401 @@
...
+function calendar.year(date, font)
+    return get_calendar("year", date, font)
+end
+
+
+function calendar.mt:__call(caltype, ...)
+    return get_calendar(caltype, ...)
+end
+
+return setmetatable(calendar, calendar.mt)
```
psychon: Our usual vim modeline is missing.

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,401 @@
...
+        end
+    end
+end
+
+-- Layout widget
+function calendar:layout(_, width, height)
+    if self._private.container then
+        return { base.place_widget_at(self._private.container, 0, 0, width, height) }
+    end
+    return {}
```
psychon: This widget just wraps another widget. As I see it, that other widget is a `grid` layout. How about just creating the grid up-front and passing it as the first argument to `base.make_widget`, so that the `:fit` and `:layout` methods become unnecessary here.

lib/wibox/widget/calendar.lua
```diff
@@ -0,0 +1,401 @@
...
+
+
+--- Return a new calendar widget by type.
+--
+-- @tparam string t Type of the calendar, `year` or `month`
+-- @tparam string date Date of the calendar
+-- @tparam[opt="Monospace 10"] string font Font of the calendar
+-- @treturn widget The calendar widget
+local function get_calendar(t, date, font)
+    if t ~= "month" and t ~= "year" then return end
```
psychon: Replace `return` by an `error()` call with a useful error message. That makes debugging a lot easier, if this ever goes wrong.

lib/wibox/widget/calendar.lua
actionless: i would add just `beautiful.font` before Monospace but i think monospace is good one, because if not unset it could use sans and make calendar unreadable

lib/wibox/widget/calendar.lua
getzze: I used the code from another calendar widget, but the creation of the 'weekdays' cells can be done in the next for-loop indeed.

lib/wibox/widget/calendar.lua
getzze: This is the behavior of the `cal` command-line in linux, so the cell size is two characters, like for the days numbers. I didn't find another way to ouput a two-characters weekday with the LUA libraries.

lib/wibox/widget/calendar.lua
getzze: There were two problems with this solution:
1-the api of the `grid` widget was exposed as the api of the calendar widget, so names could collide.
2-to change the date of the calendar widget, you have to destroy the last grid and create a new one (I think) and it is not so straightforward to have a cache (see other comment).

lib/wibox/widget/calendar.lua
getzze: Creating a year calendar is slow, so the cache helps if you want to check the next year and come back to the current year. Also I didn't understand how the widget cache works so I did this.

lib/wibox/widget/calendar.lua
getzze: Because of the grid alignment, a monospace font is nicer. But it works also with other fonts, so as you want

lib/wibox/widget/calendar.lua
getzze: maybe I could remove the test and make the function completely local (by removing calendar.mt:__call(caltype, ...)).

lib/wibox/widget/calendar.lua
getzze: `date` is a single property, otherwise there are three. but maybe a table would be ok.

lib/wibox/widget/calendar.lua
psychon: The widget cache in `wibox.widget.base`? It just caches the result of `:fit()` and `:layout()` and is cleared when the respective signals are emitted. So that would not help here.

lib/wibox/widget/calendar.lua
psychon: 1) I don't mean to replace `local my_widget = base.create_widget() my_widget._private.real_widget  = grid()` with `local my_widget = grid()`. Instead I mean: Replace this with: `local real_widget = grid() local my_widget = base.create_widget(real_widget) my_widget._private.real_widget = real_widget`. That way you do not have to implement `:fit()` and `:layout` at all, but instead they are automatically redirected to the target widget. The only problem is that you then cannot replace the grid layout any more and always have to use the same. However, that should not be a problem because of (2):
(2): There is a `:reset()` method on basically all layouts, also on the `grid` layout. That way you do not have to replace the grid.

And for "not so straightforward to have a cache": Well, I am not that much of a fan of this cache so far anyway. ;-)

lib/wibox/widget/calendar.lua
psychon: Hm... I am not sure `cal` should do that either...

lib/wibox/widget/calendar.lua
psychon: > I used the code from another calendar widget,

Copyright-wise, could this be a problem? Even if that other implementation is GPL'd (is it?), then this would violate the attribution clause in the GPL.

lib/wibox/widget/calendar.lua
psychon: I don't have much of a problem with three properties either, but sure, I don't mind a table either. Just string parsing should be avoided where possible. ;-)

lib/wibox/widget/calendar.lua
actionless: i would use max length of abbreviated weekday name as a column width (for all columns)

lib/wibox/widget/calendar.lua
getzze: I took inspiration - not exact copy - from this one:
https://github.com/deficient/calendar
Otherwise, for calendar_popup (awful widget), the code is very similar to lain.calendar.
These two should be mentioned I guess.

lib/wibox/widget/calendar.lua
getzze: I will add an option to show three characters

lib/wibox/widget/calendar.lua
getzze: Actually, `real_widget` is not necessarily a `grid`. It can be embedded in another widget, see the last lines of the `create_month` and `create_year` functions.
The code for the `:fit` and `:layout` functions are the same as for others `wibox.widget.container`, it does not require much maintenance.

getzze:
Different changes:
 - rename awful.widget.calendar -> calendar_popup
 - use table {year=..., month=..., day=...} instead of strings for the date
 - rename the flag `weekdays` to `weekday`
 - add `long_weekdays` property to display the weekdays with 3 chars instead of 2
 - remove the cache
 - use a container.background as proxy for the calendar widget, to be able to remove the `:fit` and `:layout` functions.
 - remove the global calendar(type, ...) function
 - add a `date` argument to `fn_embed` function so you can decorate the days depending on the date (change colors for weekend days, add a list of bank holidays...)

About the default font, should I leave "Monospace"?
I also realized about a non-consistent behavior between  calendar.year and calendar.month: the year calendar always highlights the current day, whereas calendar.month highlights the specified day. It is the same behavior as the cal command-line. I thought about changing it by adding a `no_highlight` property. If it is set, no day is highlighted, otherwise, if a day is specified, it is the one that is highlighted, otherwise the current day is highlighted. But it's maybe a bit complicated.

Elv13:
> About the default font, should I leave "Monospace"?

I think so, it would otherwise look ugly.

Also, could you rebase on top of master so the CI shows up again? (it did run, but isn't visible)

psychon:
@Elv13 How unlikely is it that you can find the time to look at this in the near future?

Elv13:
@psychon I built it, tested it and looked at the code and API. I think for a first version, this is getting good enough to merge. One way or another there is no way this could have been built in 10 lines of code, so the size and sometime non-trivial code are to be expected. I am fine with merging this as-is (once luacheck is fixed)

@getzze Thanks for this. Do you want this to be merged? I can fix the luacheck regression when I merge this.

Elv13:
Thanks a lot to everybody who contributed to this PR, be it code, inspiration, testing or reviews.

PR #1690, opened 2017-03-31
Fix some deprecated calls in code and docs

Veratil:
Add missing newline in test runner

---

* Missed changing `awful.util.get_themes_dir` in `rc.lua` to `gears.filesystem.get_themes_dir`
* Updating `90-FAQ` to not use deprecated calls
* Updating `keygrabber` comments to not use deprecated calls
* Add missing newline in tests `_runner`


Elv13:
Thanks

PR #1693, opened 2017-03-31
Improve doc for creating new widget

getzze:
The function to use to implement new widget was not specified...
Also `set_children` is very important because it is called automatically by the `drill` internal function.
It is good to mention it in the doc.

The modification to `base` is because I had bugs with a new widget "inherited" from a container. When I used it as the last widget in a declarative table, because `widgets==nil`, it was setting the internal widget to `nil`. I had to override `set_children` to do nothing to solve the pb.


docs/04-new-widgets.md
```diff
@@ -111,3 +111,10 @@ looks like this:
         widget:after_draw_child(context, cr, child_index, child, width, height)
     end
     widget:after_draw_children(context, cr, width, height)
+
+The `:get_children()` and `:set_children()` methods are called when setting a
+widget with the declarative layout system, therefore these two methods should be
+well defined. Specifically, `:get_children()` should return all the children to
+be able to draw them. `:set_children()` is called by the declarative system to
```
psychon: I don't think anything in awesome currently ever calls `get_children` (and for drawing them, you also need a position and that's `:layout`'s job). @Elv13 ?

getzze: Yes, I realized and removed this part, sorry. Only `set_children` is used

PR #1697, opened 2017-04-02
awful.autofocus: Discriminate sticky clients

psychon:
A relatively common problem with awesome is with mixing sticky clients
and the focus history. Once a sticky client ever had the focus, it will
always get the focus after a tag switch. This is because the focus
history is global and the sticky client is always the most recently
focused and currently visible client in the list.

Work around this by discriminating sticky clients: First try to find a
client to focus, but ignore sticky clients. When this does not find
anything, try again, but this time also consider sticky clients.

(Basically the same issue exists with clients that are on multiple tags,
but I guess that one can be ignored.)

Fixes: https://github.com/awesomeWM/awesome/issues/733

### DIFF
--- a/lib/awful/autofocus.lua
+++ b/lib/awful/autofocus.lua
@@ -13,6 +13,10 @@ local client = client
 local aclient = require("awful.client")
 local timer = require("gears.timer")
 
+local function filter_sticky(c)
+    return not c.sticky and aclient.focus.filter(c)
+end
+
 --- Give focus when clients appear/disappear.
 --
 -- @param obj An object that should have a .screen property.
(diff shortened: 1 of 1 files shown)


Maxlufs:
I'm not sure where went wrong but i'm still having this problem.
To reprod:
step 0: new login
step 1: open google-chrome
step 2: open terminal (slave)
step 3: focus on terminal, press ctrl+shift+mod+2 (terminal now has multiple tags 1 and 2)
step 4: focus on google-chrome (master) 
step 5: switch to tag 2 by pressing mod+2
step 6: switch back to tag 1 by pressing mod + 1
Expected output: google-chrome gets focused
Actual output: terminal gets focused

Could you triage this? thanks!

awesome v4.2-106-g8923df4-dirty (Human after all)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.4
 • LGI version: 0.9.0

psychon:
@Maxlufs You have a (slightly) different problem. This PR does something about sticky clients. In your example there is no sticky client, so the normal "focus the client which last had the focus"-mechanism applies. Since this is a global list of clients and not a per-tag one, you get the behaviour that you are seeing. Dunno if there is an open issue about this.

PR #1698, opened 2017-04-02
awful.ewmh.add_activate_filter: Fix doc example

psychon:
I didn't actually test this, but at least this now looks like valid Lua
code to me, so this is definitely an improvement.


Elv13:
Thanks

Issue #1663, opened 2017-03-19
~/.cache/awesome/ does not exist => menubar crash while saving menu_count_file


bew:
Output of `awesome --version`:
```
awesome v4.1-dirty (Technologic)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1
```

**How to reproduce the issue:**

Open the menubar, type anything and press enter

**Actual result:**

```
2017-03-19 18:08:33 W: awesome: luaA_dofunction:77: error while running function!
stack traceback:
	/usr/local/share/awesome/lib/menubar/init.lua:152: in upvalue 'write_count_table'
	/usr/local/share/awesome/lib/menubar/init.lua:176: in function </usr/local/share/awesome/lib/menubar/init.lua:160>
	(...tail calls...)
	/usr/local/share/awesome/lib/awful/prompt.lua:755: in local 'keygrabber_function'
	/usr/local/share/awesome/lib/awful/keygrabber.lua:24: in function </usr/local/share/awesome/lib/awful/keygrabber.lua:21>
error: /usr/local/share/awesome/lib/menubar/init.lua:152: attempt to index a nil value (local 'count_file')
```

**Expected result:**

No error ( _obviously_ )

**Investigations**

As I'm using a tmpfs for my `~/.cache` directory, it is empty on computer start and the directory `~/.cache/awesome` doesn't exists.

The line at `/usr/local/share/awesome/lib/menubar/init.lua:152` is attempting to index a file opened inside this cache directory, thus crash.

Elv13:
Note for myself. The prompt has:

```lua
        if not f then
            local i = 0
            for d in id:gmatch(".-/") do
                i = i + #d
            end
            gfs.mkdir(id:sub(1, i - 1))
            f = assert(io.open(id, "w"))
        end
```

for this. However copy/pasting the code in the menubar isn't such a good idea. GIO has been ways to ensure a file path exist.

mvdan:
> As I'm using a tmpfs for my ~/.cache directory

Have the same setup, experiencing the exact same errors as of recently since switching to v4.1 (not sure if it was present in 4.0). Any workaround until a bugfix is out?

psychon:
@mvdan `io.popen("mkdir -p ~/.cache/awesome")` as the first line in your `rc.lua`.

mvdan:
[...]

PR #1699, opened 2017-04-02
Some fixes and changes to cache dirs

psychon:
This is what happened when I tried to fix https://github.com/awesomeWM/awesome/issues/1663.

### DIFF
--- a/lib/gears/filesystem.lua
+++ b/lib/gears/filesystem.lua
@@ -103,7 +103,9 @@ end
 --- Get the path to a directory that should be used for caching data.
 -- @return A string with the requested path with a slash at the end.
 function filesystem.get_cache_dir()
-    return filesystem.get_xdg_cache_home() .. "awesome/"
+    local result = filesystem.get_xdg_cache_home() .. "awesome/"
+    filesystem.mkdir(result)
+    return result
 end
 
 --- Get the path to the directory where themes are installed.
--- a/lib/gears/filesystem.lua
+++ b/lib/gears/filesystem.lua
@@ -23,11 +23,16 @@ local function make_directory(gfile)
     return false, err
 end
 
---- Create a directory
+--- Create a directory, including all missing parent directories.
 -- @tparam string dir The directory.
 -- @return (true, nil) on success, (false, err) on failure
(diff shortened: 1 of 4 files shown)


lib/gears/filesystem.lua
```diff
@@ -11,20 +11,30 @@ local gtable = require("gears.table")
 
 local filesystem = {}
 
+local function make_parent(gfile)
```
psychon: I should drink more water. This should be `make_directory`, not `make_parent`.

lib/gears/filesystem.lua
```diff
@@ -11,20 +11,30 @@ local gtable = require("gears.table")
...
+        -- Directory already exists, let this count as success
+        return true
+    end
+    return false, err
+end
+
 --- Create a directory
 -- @tparam string dir The directory.
 -- @return (true, nil) on success, (false, err) on failure
 function filesystem.mkdir(dir)
```
Veratil: What do you think about merging `mkdir` and `make_parent_dir` since they do essentially the same thing (making directories)? Adding an extra parameter to `mkdir` wouldn't break API.

Something like:
```lua
function filesystem.mkdir(path, parents_only)
  if parents_only then
    return make_parent(Gio.File.new_for_path(path):get_parent())
  end
  return make_parent(Gio.File.new_for_path(path))
end
```

psychon: I'm not sure about code readability. `mkdir("foo/bar/baz", true)` isn't as self-explanatory as `make_parent_dir("foo/bar/baz")`.

Hm. Should `mkdir` be renamed to `make_directory` or something like that for consistency?

Veratil: That's true. I do see that `Gio.File` has `make_directory` and `make_directory_with_parents`.

I think that would be a good change since we're not calling `os.mkdir` anymore. Also, to fit with the other `get_*_directory` functions perhaps expand `make_parent_dir` to `make_parent_directories`?

psychon:
Updated

psychon:
@blueyed Thanks for going through the pending pull requests!

PR #1705, opened 2017-04-04
Add iconv library detection by using code for macOS

tomoyat1:
The workaround for finding iconv on macOS is also necessary on FreeBSD,
because libiconv needs to be explicitly linked against on this platform.

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -160,11 +160,11 @@ endforeach()
 # Do it again, but this time with the CFLAGS/LDFLAGS
 pkg_check_modules(AWESOME_REQUIRED REQUIRED ${AWESOME_DEPENDENCIES})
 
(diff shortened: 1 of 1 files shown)


awesomeConfig.cmake
```diff
@@ -164,7 +164,7 @@ pkg_check_modules(AWESOME_REQUIRED REQUIRED ${AWESOME_DEPENDENCIES})
 # explicitly.  Unfortunately, libiconv doesn't have its pkg-config file,
 # and CMake doesn't provide a module for looking up the library.  Thus, we
 # have to do everything for ourselves...
-if(APPLE)
+if(APPLE OR CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
```
psychon: Please also change the `endif(APPLE)`

tomoyat1: Fixed.
Also edited the commit message because it made no sense.

PR #1706, opened 2017-04-04
Add support for debugging signals

blueyed:
This is #129, rebased.

Re-opening (which could not be done for the initial PR), since I've used this to debug some 100% CPU loop in awesome's `refresh`.

I think we should provide a way to easily enable this somehow.

There has not been much discussion/review on https://github.com/awesomeWM/awesome/pull/129, but please look at it before reviewing.

### DIFF
--- a/common/luaobject.c
+++ b/common/luaobject.c
@@ -287,13 +287,22 @@ luaA_object_emit_signal(lua_State *L, int oud,
         int nbfunc = sigfound->sigfuncs.len;
         luaL_checkstack(L, nbfunc + nargs + 2, "too much signal");
 
-        const char *objname = lua_tostring(L, oud_abs);
-        debug("Emitting signal '%s' on %s (%d funcs)", name, objname, nbfunc);
+        /* if( name != "refresh" ) { */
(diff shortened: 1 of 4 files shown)


Elv13:
In any case, it can't be merged as-is or it would spam the logs. I propose adding an `awesome.debug_core_signals` CAPI property to toggle this debug output (default: `false`)

blueyed:
Of course it should not be enabled by default.

> I propose adding an `awesome.debug_core_signals` CAPI property to toggle this debug output (default: false)

What about `awesome.debug_signals`? (shorter, and the namespace is `core` already)

Elv13:
> What about awesome.debug_signals? (shorter, and the namespace is core already)

IMO that would be confusing, because your patch only print message for ~5% of signals. Most signals are implemented using `gears.object` and there is *a lot* of them. Often hundreds per main loop iteration.

blueyed:
Then those should get added probably here, too.

psychon:
Another option would be to use an environment variable that enables "all this spam". No idea if that's better than other options.

actionless:
@blueyed  is it still alive?

PR #1708, opened 2017-04-05
Added different wallpapper support for multiscreen

nemux:
Added code for load different wallpapers in multi screen configuration adding a table in theme.lua files 
theme.wallpaper = { "/home/nemux/images/wallpapers/screen_1.jpg", "/home/nemux/Images/wallpapers/screen_2.jpg"}
Signed-off-by: Josue Araujo <nemux.1@gmail.com>

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -179,6 +179,9 @@ local function set_wallpaper(s)
         if type(wallpaper) == "function" then
             wallpaper = wallpaper(s)
         end
+        if type(wallpaper) == "table" then
+            wallpaper = wallpaper[s.index]
+          end
         gears.wallpaper.maximized(wallpaper, s, true)
     end
 end
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -181,7 +181,7 @@ local function set_wallpaper(s)
         end
         if type(wallpaper) == "table" then
             wallpaper = wallpaper[s.index]
-          end
+        end
         gears.wallpaper.maximized(wallpaper, s, true)
     end
 end


Elv13:
Hello,

Thanks for this. However I am not sure I agree. First of all, there is an indentation issue. But also, given you can already pass a function taking the screen as argument. You can already do this by wrapping your table in a function. So the patch doesn't really add any feature that aren't already there. May I propose improving the `rc.lua` documentation instead of merging this patch?

Veratil:
@Elv13 is there a reason the set_wallpaper function is local to `rc.lua`? Why wouldn't we have it in `beautiful`?

Elv13:
@Veratil it grew up "organically". It started in the `for s in screen do` / `connec_for_each_screen`. It was then factored out to be handle screen resize. Then some more code was added. It ended up the way it currently is. It wasn't added to `rc.lua` "on purpose". It "code-rotted" to its current state. I don't think it passed some unacceptable threshold yet.

psychon:
And before someone suggests that `beautiful` automatically should set the wallpaper: In my rc.lua here at work, I call `gears.wallpaper.maximized` with `nil` for the screen argument. I.e. I have one big image that I want to span all my screens. Right now that's easily doable, but if setting the wallpaper were to move into `beautiful`, that would become a lot harder.

Veratil:
@psychon I was thinking more of expanding the set_wallpaper function to take more arguments and be a little more versatile. I didn't know setting screen to `nil` would span all the screens though, that's cool.

Elv13:
> I was thinking more of expanding the set_wallpaper function to take more arguments and be a little more versatile.

Then users will require more features and more argument and we will have 2 copies of `gears.wallpaper`. Sometime a full module is just better than a god function. Sure, standalone themes need to copy `set_wallpaper` in their `theme.lua` to add such features, but they can, so I prefer to leave it as-is too.

Veratil:
> Then users will require more features and more argument and we will have 2 copies of gears.wallpaper.

Isn't that always the case with everything? ;)

Fair enough though! Good points all around.

blueyed:
What about moving the function from the config to either beautiful, or gears.wallpaper, and using it in the default config (I.e. not calling it automatically, but have it factored out)?

Veratil:
> What about moving the function from the config to either beautiful, or gears.wallpaper, and using it in the default config (I.e. not calling it automatically, but have it factored out)?

That's what I was thinking before. Something that could take a few arguments to be a little more generic. If we put something in `gears.wallpaper` (and change the other mode functions to take an `args` table) then we could get something like this (crude and untested obviously):
```lua
function wallpaper.set(wallpaper, s, mode, modeargs)
  if wallpaper == nil then
    if beautiful.wallpaper then wallpaper = beautiful.wallpaper end
  end
  if type(wallpaper) == "function" then wallpaper = wallpaper(s) end
  mode(wallpaper, s, modeargs)
end
```

Then our set_wallpaper line in `rc.lua` just changes to:
```lua
-- nil argument means use beautiful.wallpaper
gears.wallpaper.set(nil, s, gears.wallpaper.maximized, {ignore_aspect=true})

-- non-nil could be strings from a table for multi-monitors
gears.wallpaper.set(wallpapers[s.index], s, gears.wallpaper.maximized, {ignore_aspect=true})

-- custom wallpaper function if someone wrote one
gears.wallpaper.set(wallpaper, s, mycustomfunc, {customarg=false})
```

Elv13:
I will close this for now. This version of the patch is not the way to resolve the issue it fixes, feel free to open an issue to track it.

Issue #1710, opened 2017-04-08
The check-ldoc-warnings target is not run on Travis


psychon:
If `GENERATE_DOC` is true, `CMakeLists.txt` adds a `check-ldoc-warnings`-target to the list of targets run by `check-qa`.

On Travis, we have `DO_CHECKQA=1` for the Lua 5.2 target and `BUILD_APIDOC=true` for the Lua-5.2-with-LGI-0.8.0 target. Thus, when `check-qa` is run, LDoc is not installed / `GENERATE_DOC` is false and thus this target is never run.

Possible options:

- Merge `BUILD_APIDOC` and `DO_CHECKQA` into the same built (Elv13 will not like this)
- Explicitly run `make check-ldoc-warnings` when `BUILD_APIDOC` is true.
- Other ideas that I am missing right now

In other news, our build configuration is getting quite complex and hard to figure out (at first I thought version differences where the reason why it fails locally for me, but works on Travis).

PR #1711, opened 2017-04-08
Fix ldoc

psychon:
This should fix the errors that were not noticed due to #1710.

@barjo Could you please check if this change is OK with you?


barjo:
@psychon It works just fine, sorry about that.

Issue #1599, opened 2017-02-25
Chromium starts maximized


psychon:
No time right now to investigate, so I'll just report this:

Output of `awesome --version`:

```
awesome 9999 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.2.4 (running with Lua 5.2)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1
```
Perhaps some actually useful "version" information:
```
$ ls -l /usr/bin/awesome
-rwxr-xr-x 1 root root 2263216 Feb 18 15:13 /usr/bin/awesome
```

**How to reproduce the issue:**

* Start chromium

**Actual result:**

Chromium opens maximized. This means `xprop` says `_NET_WM_STATE(ATOM) = _NET_WM_STATE_MAXIMIZED_VERT, _NET_WM_STATE_MAXIMIZED_HORZ` and the tasklist entry looks like this:
![foo](https://cloud.githubusercontent.com/assets/89482/23331975/a5556834-fb70-11e6-9ab6-964e0af55c1e.png)

Oh and:
```
$ sleep 1 ; awesome-client 'local c = client.focus ; return c.maximized, c.maximized_vertical, c.maximized_horizontal'                                                                                                     
   boolean false
   boolean true
   boolean true
```

**Expected result:**

Chromium does not start maximized.
No idea why it previously did not do this.

psychon:
First observations: I'm bad at not trying to debug things myself.

Second observations: Chromium maximizes itself. The maximisation is coming from `ewmh_process_state_atom()` handling a request to add the maximized properties.

However, since I am quite sure that this did not happen before, I did a git bisect between `v4.0` and my current master. The result is:
```
3b1599bd990cc6f9d73f232faca580ce79b7cf53 is the first bad commit
commit 3b1599bd990cc6f9d73f232faca580ce79b7cf53
Author: Emmanuel Lepage Vallee <elv1313@gmail.com>
Date:   Tue Jan 24 04:19:45 2017 -0500

    maximize: Stop doing it in 2 operations.
    
    Before 4.0, maximizing could only be done in 2 operations.
    
    4.0 add an unified way, but kept doing 2 operations. The old
    Lua EWMH code to serialize the 2 operations was dropped when
    the codepath was simplified and replaced by a generic version
    in awful.placement. However this version never implemented
    combining multiple mementos into 1.
    
[...]

PR #1712, opened 2017-04-08
WIP: Setting c.maximized now also sets max_vert/horz

psychon:
Today at 4 AM someone on IRC once again had the problem "Chrome is maximized, but toggling its maximization status does not change anything" (aka "Our current handling of maximized/maximized_vertical/maximized_horizontal is confusing and wrong"). Here is a minimal patch attempting to change this.
Testing done: None
Fixes: #1599 (I think...??)

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -2008,7 +2008,7 @@ void
 client_set_maximized(lua_State *L, int cidx, bool s)
 {
     return client_set_maximized_common(
-        L, cidx, s, "maximized", CLIENT_MAXIMIZED_BOTH
+        L, cidx, s, "maximized", CLIENT_MAXIMIZED_BOTH | CLIENT_MAXIMIZED_H | CLIENT_MAXIMIZED_V
     );
 }
 


Elv13:
I don't think this is the solution. It brings back "split states" and the bugs attached to them. I attempted to prevent this from happening by merging `max_h+max_v -> max_both` when the client is added at startup, but it doesn't seem to work right. There is some other case where max_v+max_h accidentally happen. I think those cases need to be fixed first. The bug exist and it's annoying to some people and not other. However this patch goes it in the opposite direction I think is right in the long run. If an extreme solution is preferred, I would prefer to remove max_h/max_v rather than bringing back split states and signal hell.

Veratil:
I had the same problem with Atom. For some reason it had the `_NET_WM_STATE_MAXIMIZED_[HORZ|VERT]` set which Awesome didn't detect upon client startup and therefore didn't set the maximized state flag. When I maximized it, it set the maximized flag properly, but didn't remove when unmaximizing so the flags were still set, causing it to not tile with the layout. It took unsetting with `xprop` to finally get it to play correctly.

blueyed:
btw: we have https://github.com/awesomeWM/awesome/pull/1572, so it's possible to change those states in the default config.

I think the main issue here is that for some reason clients might set max_h/max_v during startup (I've noticed it with KMyMoney), and we might be good by ignoring this during startup (but not the maximized state itself).
See also https://github.com/awesomeWM/awesome/issues/1559.

Elv13:
Closing as it should be "fixed"

Issue #1683, opened 2017-03-29
popup widget / widget in tooltip


jandob:
I am currently trying to make a volume control widget for pulse audio.
On mouse over it would be nice to show volume control sliders for all sinks, sources and clients. 

Is there any possibility to have widgets in an `awful.tooltip`? 
If not, can someone provide any pointers on how to place a widget outside the wibar?

psychon:
> Is there any possibility to have widgets in an awful.tooltip?

Nope, sorry.

> If not, can someone provide any pointers on how to place a widget outside the wibar?

```lua
-- Preparation
local tooltip_widget = whatever
local w = wibox{ width = 42, height = 42, widget = tooltip_widget }

-- Making visible
awful.placement.left(w, { parent = mouse, offset = { x=1, y=0 }})
w.visible = true

-- Hiding
w.visible = false
```
And if you want to implement your own tooltip on something:
```lua
local widget = whatever
local tooltip_widget = whatever
local w = wibox{ width = 42, height = 42, widget = tooltip_widget }
widget:connect_signal("mouse::enter", function()
  awful.placement.left(w, { parent = mouse, offset = { x=1, y=0 }})
  w.visible = true
end)
widget:connect_signal("mouse::leave", function()
  w.visible = false
end)
```
If you wonder "how do I pick good values for width and height?", then you found out why `awful.tooltip` cannot display arbitrary widgets (yet). A text can be queried for its size (more or less...), but arbitrary widgets cannot.

(Memo to myself: `awful.tooltip` could provide a function `set_widget(widget, width, height)`, I think....)

jandob:
Thanks for your quick response, i have got it working now!

In case you are curious, here is my current implementation:

```
volWidget:connect_signal("mouse::enter", function(other, geo)
    pulseaudio.get_devices(function (deviceList)
        local wibox_height = 0
        local slider_height = 20
        local slider_width = 100
        local sliders = wibox.widget {
            forced_width = slider_width,
            layout  = wibox.layout.flex.vertical
        }
        for _, device in ipairs(deviceList) do
            local slider = wibox.widget(slider_properties)
            slider.value = device.volume
            slider.pulseaudio_device = device
            slider:connect_signal("widget::redraw_needed", function ()
                -- called when the slider value changes
                if not slider.pulseaudio_device then return end
                pulseaudio.set_volume(slider.pulseaudio_device, slider.value)
            end)
            sliders:add(slider)
[...]

PR #1713, opened 2017-04-08
tooltip: Add set_widget() API

psychon:
Fixes: https://github.com/awesomeWM/awesome/issues/1683

### DIFF
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -209,11 +209,19 @@ end
 -- @tparam tooltip self A tooltip object.
 local function set_geometry(self)
     -- calculate width / height
-    local n_w, n_h = self.textbox:get_preferred_size(mouse.screen)
-    n_w = n_w + self.marginbox.left + self.marginbox.right
-    n_h = n_h + self.marginbox.top + self.marginbox.bottom
-
+    local n_w, n_h
     local w = self:get_wibox()
+
+    if self._private.widget then
+        w:set_widget(self._private.widget[1])
+        n_w, n_h = self._private.widget[2], self._private.widget[3]
+    else
+        n_w, n_h = self.textbox:get_preferred_size(mouse.screen)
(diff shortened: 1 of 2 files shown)


lib/awful/tooltip.lua
```diff
@@ -405,6 +412,21 @@ function tooltip:set_preferred_positions(value)
     set_geometry(self)
 end
 
+--- Set a widget to display.
+-- @tparam widget widget The widget to display
+-- @tparam integer width The width of the tooltip
+-- @tparam integer height The height of the tooltip
+function tooltip:set_widget(widget, width, height)
```
Elv13: I think it should be 3 properties, but a function with 3 parameters. The size concerns and widgets concerns seems rather different. One might want different size with the text and others might be ok with the size returned by `:fit()` as-is (with the screen width/height as inputs)

psychon: The size of the screen is wrong. The correct value would be the size of the screen minus the placement of the tooltip, i.e. only the "really remaining space". But then we have a dependency loop between placement and size.

Also, the following widgets/layouts cannot be used in this scenario since they just take all available space:
- checkbox
- imagebox (mostly (allow_resize))
- piechart
- progressbar
- slider
- flex
- systray (mostly (base size needs to be unset), but still: ewh, no!)

These would work (sometimes, if they don't contain something from the above list):
- textbox
- textclock
- align
- constraint
- fixed
- grid

So... basically only a textbox works in the scenario you describe. And even then it is easy to get a really long tooltip where multiple lines would be better.

(I'm ignoring containers since they mostly pass on the fit result of some other widget)

Elv13: > Also, the following widgets/layouts cannot be used in this scenario since they just take all 
> available space:

They would work with a constraint widget and/or forced_height/width. This is exactly how they work in a `wibar`. For the sake of consistency, I would still argue my previous comment stands.

PR #1715, opened 2017-04-09
[API break] Remove vertical/horizontal maximisations

psychon:
See https://github.com/awesomeWM/awesome/pull/1712#issuecomment-292742298. Places to change where found via `git grep 'maximized_[vh]`.

However, I don't think this is v4-worthy since it not only deprecates a feature, but breaks it (`c.maximized_horizontal = true` now also sets the other two maximization properties). Time for a v5 milestone/branch?


blueyed:
I don't like this: max_h/max_v is a nice feature by itself, and should get fixed properly instead.
The last time I've looked into this @Elv13 had some suggestions where to look for fixing it, I could try to find the gists from the IRC logs.

psychon:
Summary:
- Removing maximized_vertical/_horizontal is veto'd by @blueyed since he likes that feature. https://github.com/awesomeWM/awesome/pull/1715#issuecomment-292795334
- Removing maximized is veto'd by @Elv13 since it breaks `awful.placement`. https://github.com/awesomeWM/awesome/pull/1712#issuecomment-292742298
- The current situation is veto'd by me, since EWMH only has maximized_vertical/horizontal while too many people only have maximized in their config. This results in too many people getting confused since a window is maximized, but they cannot unmaximize it (See #1599: Even I got confused. What should "Average Joe" do?). Right now we are trying to work around it by assigning "some of the states" based on "some other states", but that's just a band-aid and IMO there will always be remaining problems down this path.

I'm out of ideas. The best idea would be to fix `awful.placement` so that `maximized` can be removed again, but that code is so complex that I don't want to make it more complex. Something like "make awful.placement only create mementos once per main loop iteration" might work, dunno...

getzze:
What about keeping maximized_vertical/_horizontal (they are part of EWMH and it allows fine tuning) and instead of defining maximized_both as a different state in the C-api, define it as V&H (meaning that c.maximized is not stored but is a function that checks for V&H).
Therefore, checking for maximized means checking for (max_v&max_h), and setting `maximized` to true means set max_v=true, max_h=true, setting `maximized` to false means set max_v=false, max_h=false.

psychon:
@getzze That's what was veto'd in https://github.com/awesomeWM/awesome/pull/1712#issuecomment-292742298 (as I mentioned in my previous comment)

getzze:
@psychon it's not exactly the same I think.
Because `max=max_v&max_h` visually, not implementing it like that in the c-api can creates bugs (and it does). So only `max_v` and `max_h` should have their value stored. And then define `max` as I said makes `toggle_max` to work.
Then for the signals, we can have the four `property::max_both`, `property::max_h`, `property::max_v` and ``property::max_none` (or a single signal with two bits/four states). They can be sent through a filtering function before actually being emitted, so they are not sent twice.

PR #1716, opened 2017-04-09
Travis: merge BUILD_APIDOC and DO_CHECKQA

blueyed:
Fixes https://github.com/awesomeWM/awesome/issues/1710.

Should fail now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,15 @@ branches:
 
 env:
   matrix:
-    - LUA=5.2 DO_CHECKQA=1 LUANAME=lua5.2 DO_COVERAGE=coveralls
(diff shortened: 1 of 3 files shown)


blueyed:
Works (i.e. fails): https://travis-ci.org/awesomeWM/awesome/builds/220295560

psychon:
:+1: thanks!

psychon:
Should I just merge #1711 to handle that dependency?

blueyed:
Yeah, I thought GitHub would notice that it's a single commit now and then I would have squashed it from the UI.

psychon:
Thanks for handling this!

PR #1722, opened 2017-04-14
naughty get_offset() bug fix and some notification object change

paul-axe:
get_offset() function returned incorrect values for multi-monitor configurations and notification position set to "*_middle". 

Also i think it's a good idea to save actions and run function into notificaiton object, so users can call this actions from keyboard shortcuts (like in dunst).

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -546,7 +546,7 @@ function naughty.notify(args)
         beautiful.notification_margin
     local opacity = args.opacity or preset.opacity or
         beautiful.notification_opacity
-    local notification = { screen = s, destroy_cb = destroy_cb, timeout = timeout }
+    local notification = { screen = s, destroy_cb = destroy_cb, timeout = timeout, actions = actions }
 
     -- replace notification if needed
     if args.replaces_id then
@@ -583,6 +583,7 @@ function naughty.notify(args)
             die(naughty.notificationClosedReason.dismissedByUser)
         end
     end
+    notification.run = run
 
     local hover_destroy = function ()
         if hover_timeout == 0 then
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -274,7 +274,7 @@ local function get_offset(s, position, idx, width, height)
     if position:match("left") then
         v.x = ws.x + naughty.config.padding
     elseif position:match("middle") then
-        v.x = (ws.width / 2) - (width / 2)
+        v.x = ws.x + (ws.width / 2) - (width / 2)
     else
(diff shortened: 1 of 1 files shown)


lib/naughty/core.lua
```diff
@@ -583,6 +583,7 @@ function naughty.notify(args)
             die(naughty.notificationClosedReason.dismissedByUser)
         end
     end
+    notification.run = run
```
psychon: Uhm, why? Both of these could be done by the caller. Instead of `local foo = naughty.run{ actions = actions, run = run }`, just do `local foo = { actions = actions, run = run, notification = naughty.run{ actions = actions, run = run }}` and fix up all surrounding code.
(Well, ok, if you do not use a run callback, you need to call `naughty.destroy(notification, naughty.notificationClosedReason.dismissedByUser)` directly. Still, I fail to see the point.)

lib/naughty/core.lua
```diff
@@ -274,7 +274,7 @@ local function get_offset(s, position, idx, width, height)
     if position:match("left") then
         v.x = ws.x + naughty.config.padding
     elseif position:match("middle") then
-        v.x = (ws.width / 2) - (width / 2)
+        v.x = ws.x + (ws.width / 2) - (width / 2)
```
psychon: :+1: for this commit

lib/naughty/core.lua
paul-axe: I think it's useful when you want to call actions for notifications sent through dbus from other apps. For example, we have notificaton sent from some app with some action in run argument (can be triggered by clicking on it) and also several actions in actions argument. I want to be able to draw menu by hotkey (when is see notification on the screen) with all these actions and call some of them using my keyboard. I can get notification from naughty.notificatoins table, but don't know other way to get available actions from it

psychon: Uhm... that makes me wonder "why is `naughty.notifications` accessible?" That makes a terrible API since it just exposes internals. @Elv13 Opinions?

lib/naughty/core.lua
blueyed: Let's cherry-pick that one then already?!

blueyed: @paul-axe 
Or can you create a separate PR for it, please?

paul-axe: sure, #1727

lib/naughty/core.lua
paul-axe: Actually `naughty.notifications` is documented in official wiki: [https://awesomewm.org/doc/api/libraries/naughty.html#notifications](https://awesomewm.org/doc/api/libraries/naughty.html#notifications).

Issue #1658, opened 2017-03-16
When running Chrome 57.0.2987.98  under awesome, clicks don't always work


chadcatlett:
Output of `awesome --version`:

```
$ awesome --version
awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1
```
**How to reproduce the issue:**

1. Browse to http://memset.com/
2. Hover over Services
3. Move mouse to Virtual Servers under the Servers heading on the left
4. Click Virtual Servers link

**Actual result:**

Click doesn't seem to have made it to the browser.

**Expected result:**

Click should have made it to the browser and the browser loads Virtual Servers link.

**Notes:**

It's worth noting in a previous version of Chrome this did not occur. Also worth noting that switching to XFCE4, Openbox, i3, etc. and Chrome works correctly.

I'm not stating the issue is actually awesome, but it's strange that works fine under other WMs but not awesome. 

I opened an issue with the Chromium project, it is located at https://bugs.chromium.org/p/chromium/issues/detail?id=702255 . I'll update the Chromium issue with this issue as well.

0x041E:
Found what causes the issue it is in awesome. In rc.lua clientbuttons bind for left mouse
`awful.button({ }, 1, function (c) client.focus = c; c:raise() end),`
If you comment out this line and reload config then clicking on that site works correctly although mouse focus for windows is obviously partially broken since you unbound window focus so you have to use keyboard binds to focus windows.

psychon:
Same issue as #1479

psychon:
Fixed in Chromium according to https://github.com/awesomeWM/awesome/issues/1479#issuecomment-287410587, could someone test?

Philosoft:
dunno how to verify if this commit belongs to a specific buuild. tried [latest beta](https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Linux_x64%2F464322%2Fchrome-linux.zip?alt=media) - still the same

AranelSurion:
It seems this bug is fixed in the latest unstable Chrome version, at least it worked fine on my awesome3.5 debian setup.

See: https://www.google.com/chrome/browser/desktop/index.html?platform=linux&extra=devchannel

psychon:
[...]

Issue #1479, opened 2017-01-26
Unable to move clips around in FlowBlade Movie Editor


kacperski1:
Output of `awesome --version`:

awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1

**How to reproduce the issue:**

1. Create a new project in [FlowBlade Movie Editor](https://github.com/jliljebl/flowblade),

2. Add some clips to the project (can be anything),

3. Try to move clips around the timeline.

**Actual result:**

Sometimes everything works correctly, but most of the time instead of, well, clips moving around as expected, awesome starts spawning file icons at mouse position instead. 
This bug only happens with awesome window manager and, [according to FlowBlade developers](https://github.com/jliljebl/flowblade/blob/master/flowblade-trunk/docs/KNOWN_ISSUES.md), cannot be fixed in the program itself. 

"The problem causing the bug is that awesome fires 'leave-notify-event' signals even when mouse has not left the timeline area." (quoting the FlowBlade devs)

**Expected result:**

Clips should be moving around the timeline.

blueyed:
Can you try it with Git master, please?

psychon:
Sigh. X11. Remove `awful.button({ }, 1, function (c) client.focus = c; c:raise() end),` from `clientbuttons`  and this will work (but focusing clients by mouse click will stop working).

kacperski1:
> Can you try it with Git master, please?

Tried it with awesome v4.0-165-ga1a5e18b, no change.

> Sigh. X11. Remove awful.button({ }, 1, function (c) client.focus = c; c:raise() end), from clientbuttons and this will work (but focusing clients by mouse click will stop working).

This hacky solution works, would be great to keep the client raise functionality though.

psychon:
Some X11 explanation: Titlebars work by creating an own window and reparenting the window that (for example) Firefox into this window. The titlebar-window is a bit larger and this extra space can be used to draw titlebars.

What we do for clientbuttons: We grab the input event on our titlebar-window. Thus, the input focus shortly switches to the titlebar-window for the click event, which is why the "real window" gets informed that the input focus kind-of-shortly moved elsewhere.

What to do about this: Dunno.

blueyed:
Is / can this be short-circuited / skipped when not using titlebars?

[...]

PR #1724, opened 2017-04-15
Stop grabbing mouse buttons on frame windows

psychon:
(@Airblader @stapelberg Thanks for i3 and thanks for getting this right there)

I'm stupid. I failed to see the obvious. Thanks to mlen from IRC to make
me look into i3's source code so that I can figure out the obvious.

First, what is the problem? We want to be informed about mouse clicks on
client windows, e.g. for client-to-focus. For this we have to grab mouse
buttons on the client window, since only a single client at a time can
ask for information about all mouse clicks (I think...) and that right
is reserved for the actual application and not the WM.

We also want to be informed about mouse clicks on the titlebar, so that
clicking there actually does something (and also e.g. click-to-focus).
Obviously, we can just grab mouse buttons on the titlebar as well.

However, this causes lots of problems. The X11 window containing the
titlebar is not just the size of the titlebar, but also includes all of
the actual client window. This means that clicking into the client also
activates our button grab on the titlebar. This, in turn, causes the
input focus to briefly shift to the window for the titlebar. This shift
of focus can be detected by clients and caused various issues.

This fix is so obvious that I missed it. We don't have to grab buttons
on the titlebar window. We can just use the "normal" event propagation
without grabs there, i.e. we just include the event masks for button
press and button release and get informed about all mouse events. In
fact, we already have this event mask in place, so the whole use of
grabs is redundant.

That's what this commit does. It also partially reverts commit
394ff06589676c which is where this unnecessary grabbing was introduced.

Fixes: https://github.com/awesomeWM/awesome/issues/1479
Fixes: https://github.com/awesomeWM/awesome/issues/1658

### DIFF
--- a/objects/window.c
+++ b/objects/window.c
@@ -88,7 +88,6 @@ luaA_window_buttons(lua_State *L)
     {
         luaA_button_array_set(L, 1, 2, &window->buttons);
         luaA_object_emit_signal(L, 1, "property::buttons", 0);
-        xwindow_buttons_grab(window_get(window), &window->buttons);
         xwindow_buttons_grab(window->window, &window->buttons);
     }
 


Airblader:
Looking at the awesome code helped us with i3 often enough. I'm happy to see it also works the other way around. :-)

PR #1725, opened 2017-04-15
Remove most uses of XCB_CURRENT_TIME

psychon:
The ICCCM and XEMBED specifications agree that CurrentTime really should be avoided.


Elv13:
@blueyed Just a note, last year I merged some patches after you approved them, but for the next few months that wont happen, so feel ok to merge PRs right away.

PR #1729, opened 2017-04-15
Add the ignore_suspend flag to naughty

### LINKED ISSUE #1728: Feature Request: flag to ignore naughty suspend mode
It would be nice to have a possibility to ignore the suspend state for single notifications, something like:

```lua
naughty.notify({
  -- ...
  ignore_suspend = true
})
```

**Use case:** the suspend mode is great to create a Do Not Disturb mode for you window manager. But some notifications are triggered in direct reaction to a user interaction (e.g. I have a shortcut to switch the keyboard layout, and the currently selected keyboard is shown in a notification). These notification often would make sense to also show during do not disturb mode.

timroes:
Add an `ignore_suspend` flag to `naughty.notify` so that you can choose for single notifications to ignore the global suspend state.

This implements the behavior wished for in #1728

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -494,6 +494,8 @@ end
 --   Note: this function is only relevant to notifications sent via dbus.
 -- @tparam[opt] table args.actions Mapping that maps a string to a callback when this
 --   action is selected.
+-- @bool[opt=false] args.ignore_suspend If set to true this notification
+--   will be shown even if notifications are suspended via `naughty.suspend`.
 -- @usage naughty.notify({ title = "Achtung!", text = "You're idling", timeout = 0 })
 -- @treturn ?table The notification object, or nil in case a notification was
 --   not displayed.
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks.
I wonder if `ignore_suspended` would be a better name?!

timroes:
@blueyed I have no strong feelings regarding the name. If you wish for something else, I will happily rename it.

@psychon Don't need this in the commit message. Thought, that it might close the other issue once merged automatically. But I will rename it and remove it.

blueyed:
@timroes 
I'm not sure myself, just thought that `ignore_suspend` could be improved upon.. e.g. `ignore_dnd_mode` would be better, but then it's not named "do not disturb mode" currently.

Veratil:
`suspended_override` (or `override_suspended`) ?

timroes:
How will we get to a decision regarding the variable name? :-) Does anyone just want to make this now, or should I kind of open a comment for each variant and we all want to vote?

psychon:
To ensure some progress (and to test whether this actually works):
With the powers that I gave myself just now, I rule that the argument is to be called `ignore_suspend`.

Also, it would be nice to mention the possible use case from #1728 in the commit message (A Do-Not-Disturb mode where some notifications are triggered e.g. via some keybindings and thus should ignore the Do-Not-Disturb mode). Oh and `Fixes: https://github.com/awesomeWM/awesome/issues/1728` might also be a nice addition for the commit message.

timroes:
Apparently it worked!

blueyed:
Thanks!

Issue #1734, opened 2017-04-17
The auto-completion of mod+r luncher doesn't work with zsh


YuanPre:
Output of `awesome --version`:

awesome v4.1 (Technologic)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1

**How to reproduce the issue:**

I use`·mod + r` to open a program and want to type the tab key for auto-completion.

**Actual result:**
When my home directory does not have anything, it will not be completed, and output error log in `$HOME/.xsession-errors`.
the error log is :` zsh:1: no matches found: /home/tlmf/*(:t)`
**Expected result:**

It should output complete information

**Other imformation**
I use the Z shell.This problem does not occur when using bash.
But I use zsh + i3wm will not have this problem
The `$HOME/.xsession-errors` is in here

[xsession-errors.txt](https://github.com/awesomeWM/awesome/files/925135/xsession-errors.txt)

blueyed:
This should fix it:
```diff
diff --git i/lib/awful/completion.lua w/lib/awful/completion.lua
index 2777cd43..a550988f 100644
--- i/lib/awful/completion.lua
+++ w/lib/awful/completion.lua
@@ -108,7 +108,7 @@ function completion.shell(command, cur_pos, ncomp, shell)
             "res=( "..
             "\"${(k)commands[@]}\" \"${(k)aliases[@]}\" \"${(k)builtins[@]}\" \"${(k)functio
             "\"${(k)reswords[@]}\" "..
-            "${PWD}/*(:t)"..
+            "${PWD}/*(N:t)"..
             "); "..
             "print -ln -- ${(M)res[@]:#" .. string.format('%q', words[cword_index]) .. "*}'"
         end
```

blueyed:
@YuanPre 
Can you test if #1736 works for you, please?

YuanPre:
@blueyed I dont't use the GitHub version. I used Arch Linux to package the binary version. Which file should I change?

YuanPre:
@blueyed  Thank you !
I changed you said part of `/usr/share/awesome/lib/awful/completion.lua`. Then, it works.
If this change does not raise other bugs, I recommend adding this change to the next stable version.
After all, the $HOME directory is not always a file every time ah.
I live in China, because the time difference is not timely reply to you, I am sorry.
[...]

PR #1736, opened 2017-04-17
awful.completion: zsh: fix command completion from PWD

blueyed:
Fixes https://github.com/awesomeWM/awesome/issues/1734.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,8 @@ install:
   # See also `apt-cache showsrc awesome | grep -E '^(Version|Build-Depends)'`.
   - sudo apt-get install -y libcairo2-dev gir1.2-gtk-3.0 xmlto asciidoc libpango1.0-dev libxcb-xtest0-dev libxcb-icccm4-dev libxcb-randr0-dev libxcb-keysyms1-dev libxcb-xinerama0-dev libdbus-1-dev libxdg-basedir-dev libstartup-notification0-dev imagemagick libxcb1-dev libxcb-shape0-dev libxcb-util0-dev libx11-xcb-dev libxcb-cursor-dev libxcb-xkb-dev libxkbcommon-dev libxkbcommon-x11-dev
 
-  # Deps for functional tests.
-  - sudo apt-get install -y dbus-x11 xterm xdotool xterm xvfb
+  # Deps for tests.
+  - sudo apt-get install -y dbus-x11 xterm xdotool xterm xvfb zsh
 
   # Need xorg-macros
   - sudo apt-get install -y xutils-dev
--- a/lib/awful/completion.lua
+++ b/lib/awful/completion.lua
@@ -156,7 +156,7 @@ function completion.shell(command, cur_pos, ncomp, shell)
     end
 
     local str = command:sub(1, cword_start - 1) .. output[ncomp] .. command:sub(cword_end)
-    cur_pos = cword_end + #output[ncomp] + 1
+    cur_pos = cword_start + #output[ncomp]
 
     return str, cur_pos, output
 end
--- a/lib/awful/completion.lua
+++ b/lib/awful/completion.lua
@@ -103,15 +103,17 @@ function completion.shell(command, cur_pos, ncomp, shell)
             -- NOTE: ${~:-"..."} turns on GLOB_SUBST, useful for expansion of
             -- "~/" ($HOME).  ${:-"foo"} is the string "foo" as var.
             shell_cmd = "/usr/bin/env zsh -c 'local -a res; res=( ${~:-"
-                .. string.format('%q', words[cword_index]) .. "}* ); "
+                .. string.format('%q', words[cword_index]) .. "}*(N) ); "
                 .. "print -ln -- ${res[@]}'"
         else
-            -- check commands, aliases, builtins, functions and reswords
-            shell_cmd = "/usr/bin/env zsh -c 'local -a res; "..
+            -- Check commands, aliases, builtins, functions and reswords.
+            -- Adds executables and non-empty dirs from $PWD (pwd_exe).
+            shell_cmd = "/usr/bin/env zsh -c 'local -a res pwd_exe; "..
+            "pwd_exe=(*(N*:t) *(NF:t)); "..
             "res=( "..
             "\"${(k)commands[@]}\" \"${(k)aliases[@]}\" \"${(k)builtins[@]}\" \"${(k)functions[@]}\" "..
             "\"${(k)reswords[@]}\" "..
-            "${PWD}/*(:t)"..
+            "./${^${pwd_exe}} "..
             "); "..
             "print -ln -- ${(M)res[@]:#" .. string.format('%q', words[cword_index]) .. "*}'"
         end
--- a/lib/awful/completion.lua
+++ b/lib/awful/completion.lua
@@ -93,7 +93,7 @@ function completion.shell(command, cur_pos, ncomp, shell)
         i = i + 1
     end
 
-    if cword_index == 1 and not string.find(words[cword_index], "/") then
+    if cword_index == 1 then
         comptype = "command"
     end
 
--- a/spec/awful/completion_spec.lua
+++ b/spec/awful/completion_spec.lua
@@ -1,0 +1,63 @@
+local compl = require("awful.completion")
+local shell = function(...)
+    local command, pos, matches = compl.shell(...)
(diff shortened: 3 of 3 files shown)


lib/awful/completion.lua
```diff
@@ -100,15 +100,17 @@ function completion.shell(command, cur_pos, ncomp, shell)
...
-            shell_cmd = "/usr/bin/env zsh -c 'local -a res; "..
+            -- Check commands, aliases, builtins, functions and reswords.
+            -- Adds executables and non-empty dirs from $PWD (pwd_exe).
+            shell_cmd = "/usr/bin/env zsh -c 'local -a res pwd_exe; "..
+            "pwd_exe=(*(N*:t) *(NF:t)); "..
             "res=( "..
             "\"${(k)commands[@]}\" \"${(k)aliases[@]}\" \"${(k)builtins[@]}\" \"${(k)functions[@]}\" "..
             "\"${(k)reswords[@]}\" "..
-            "${PWD}/*(:t)"..
+            "./${^${pwd_exe}} "..
```
psychon: I'll now assume that you understand this code...  No idea what this does and we have no tests for this, so I'll just believe you.

blueyed: I've looked into adding tests, but there does not seem to be an easy way to change the cwd in busted tests, is there?
As for the Zsh completion itself, I've looked into adopting the approach from https://github.com/Valodim/zsh-capture-completion, which provides much better completion.

blueyed: Added some tests.

spec/awful/completion_spec.lua
```diff
@@ -0,0 +1,52 @@
...
+    return {command, pos, matches}
+end
+local gfs = require("gears.filesystem")
+
+describe("awful.completion.shell in empty directory", function()
+    local orig_popen = io.popen
+
+    setup(function()
+        gfs.mkdir('/tmp/awesome-tests/empty')
+        io.popen = function(...)
```
blueyed: @psychon 
`luacheck` complains here rightly: "setting read-only field popen of global io"

How should this be mocked correctly?

psychon: The following causes no warnings here:
```lua
io.popen = function() end --luacheck: ignore
```
But since you are asking for "properly": You could do `completion.popen = io.popen` in the module and use that instead of `popen`. That way the tests could overwrite this directly instead of messing with the `io` table. (But I think this module-field should not be documented)

lib/awful/completion.lua
psychon: > I've looked into adding tests, but there does not seem to be an easy way to change the cwd in busted tests, is there?

Since (by that other comment)  you seem to be mocking `io.popen` now, I guess that this is no longer required? Otherwise: Nope, no easy way. You could use `lgi.GLib.chdir("/tmp")`. That would allow splitting things into two parts: One where `popen` is mocked and produces "known good" results and another part where we test that zsh produces the output we want from it. For this it might make sense to change the directory in `before_test(function() end)` and change back in `after_test()` (`lgi.GLib.get_current_dir()`).

spec/awful/completion_spec.lua
```diff
@@ -0,0 +1,57 @@
...
+        assert.same(shell('./localcomm', 12, 1, 'bash'), {'./localcommand', 15, {'./localcommand'}})
+    end)
+
+    it("completes local file", function()
+        assert.same(shell('ls ', 4, 1, 'zsh'), {'ls localcommand', 16, {'localcommand'}})
+        assert.same(shell('ls ', 4, 1, 'bash'), {'ls localcommand', 16, {'localcommand'}})
+    end)
+end)
+
+-- vim: ft=lua:et:sw=4:ts=8:sts=4:tw=80
```
psychon: I only looked at the complete diff and it seems nice. Thanks already!

Elv13:
There is still luacheck issues to fix

Elv13:
@blueyed If you want this to be in `v4.2`, can you fix the build or would you prefer if I do it?

spec/awful/completion_spec.lua
blueyed: Hmm.. instead it would overwrite / restore the `completion` module, so there is not much gained after all.
Will add ignore statements.

lib/awful/completion.lua
blueyed: I've decided to keep wrapping `io.popen` (with the `cd` in there).
Apart from performance reasons there is no need to really mock io.popen/zsh, and since it should really test against zsh also, there would only be the benefit of testing against expected output from zsh additionally.

blueyed:
@psychon 
Please review it again.

lib/awful/completion.lua
```diff
@@ -90,7 +90,7 @@ function completion.shell(command, cur_pos, ncomp, shell)
         i = i + 1
     end
 
-    if cword_index == 1 and not string.find(words[cword_index], "/") then
+    if cword_index == 1 then
```
psychon: CC @Asido Any thoughts on this and https://github.com/awesomeWM/awesome/commit/3986409e70a3ec5602512c0d9a148ba0a57cb015 ?

lib/awful/completion.lua
```diff
@@ -153,7 +153,7 @@ function completion.shell(command, cur_pos, ncomp, shell)
     end
 
     local str = command:sub(1, cword_start - 1) .. output[ncomp] .. command:sub(cword_end)
-    cur_pos = cword_end + #output[ncomp] + 1
+    cur_pos = cword_start + #output[ncomp]
```
psychon: Perhaps some words on the commit message on what is going on here? This will now return a smaller number than previously, I think. Why?

.travis.yml
```diff
@@ -41,7 +41,7 @@ install:
   - sudo apt-get install -y libcairo2-dev gir1.2-gtk-3.0 xmlto asciidoc libpango1.0-dev libxcb-xtest0-dev libxcb-icccm4-dev libxcb-randr0-dev libxcb-keysyms1-dev libxcb-xinerama0-dev libdbus-1-dev libxdg-basedir-dev libstartup-notification0-dev imagemagick libxcb1-dev libxcb-shape0-dev libxcb-util0-dev libx11-xcb-dev libxcb-cursor-dev libxcb-xkb-dev libxkbcommon-dev libxkbcommon-x11-dev
 
   # Deps for functional tests.
```
psychon: This line is now a bit wrong...

blueyed: Fixed.

lib/awful/completion.lua
blueyed: If I remember correctly it is a hack: if there's no '/' then it's considered a command, but that fails for `/usr/bin/ls` etc, of course.

lib/awful/completion.lua
blueyed: This was done so long ago (and is now covered by a test IIRC), that I do not remember really, sorry..

blueyed: Yes, reverting it breaks 3 tests:
```
Failure → spec/awful/completion_spec.lua @ 23
awful.completion.shell in empty directory handles completion of true
spec/awful/completion_spec.lua:24: Expected objects to be the same.
Passed in:
(table) {
  [1] = 'true'
 *[2] = 5
  [3] = {
    [1] = 'true' } }
Expected:
(table) {
  [1] = 'true'
 *[2] = 10
  [3] = {
    [1] = 'true' } }

Failure → spec/awful/completion_spec.lua @ 52
awful.completion.shell completes local command
spec/awful/completion_spec.lua:53: Expected objects to be the same.
Passed in:
(table) {
  [1] = './localcommand'
 *[2] = 15
  [3] = {
    [1] = './localcommand' } }
Expected:
(table) {
  [1] = './localcommand'
 *[2] = 27
  [3] = {
    [1] = './localcommand' } }

Failure → spec/awful/completion_spec.lua @ 57
awful.completion.shell completes local file
spec/awful/completion_spec.lua:58: Expected objects to be the same.
Passed in:
(table) {
  [1] = 'ls localcommand'
 *[2] = 16
  [3] = {
    [1] = 'localcommand' } }
Expected:
(table) {
  [1] = 'ls localcommand'
 *[2] = 17
  [3] = {
    [1] = 'localcommand' } }
```

lib/awful/completion.lua
Asido: 5+ years later I honestly don't know what was the reason for that patch. What @blueyed says might be right ;)

PR #1742, opened 2017-04-20
Add UTC option for textclock widget

matt-h:
In awesome 3.x textclock widget used `os.date` which allowed you to start the format with `!` to use UTC time. In 4.x it now uses `GLib.DateTime` which has its own function to get the time in UTC.

This change adds a third parameter "utc" to the textclock widget that is optional
(defaults to false).
If set to true, the textclock is formatted with UTC time.

PR #1743, opened 2017-04-20
Add timezone option for textclock widget

matt-h:
This pull request replaces https://github.com/awesomeWM/awesome/pull/1742 with code from a [comment](https://github.com/awesomeWM/awesome/pull/1742#issuecomment-295765755) by @psychon support any timezone instead of just UTC.

In awesome 3.x textclock widget used `os.date` which allowed you to start the format with `!` to use UTC time. In 4.x it now uses `GLib.DateTime` which has its own function to get the time in UTC.

This change adds a third parameter "timezone" to the textclock widget that is
optional. Defaults to local timezone if nil.

### DIFF
--- a/lib/wibox/widget/textclock.lua
+++ b/lib/wibox/widget/textclock.lua
@@ -10,7 +10,9 @@ local setmetatable = setmetatable
 local os = os
 local textbox = require("wibox.widget.textbox")
 local timer = require("gears.timer")
-local DateTime = require("lgi").GLib.DateTime
+local glib = require("lgi").GLib
+local DateTime = glib.DateTime
+local TimeZone = glib.TimeZone
 
 local textclock = { mt = {} }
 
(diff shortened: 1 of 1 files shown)


matt-h:
@psychon I was impressed it was untested.

I tested it and it worked perfectly for me.

lib/wibox/widget/textclock.lua
```diff
@@ -24,16 +26,18 @@ end
 --
 -- @tparam[opt=" %a %b %d, %H:%M "] string format The time format.
 -- @tparam[opt=60] number timeout How often update the time (in seconds).
+-- @tparam[opt] string timezone The timezone to display, or localtime if nil.
```
blueyed: Better?!
```
@tparam[opt=local timezone] string timezone The timezone to use, e.g. "Z" for UTC, "±hh:mm" or "Europe/Amsterdam". See https://developer.gnome.org/glib/stable/glib-GTimeZone.html#g-time-zone-new.
```

psychon: Fine by me. I was not sure how to document the default value since it's not like "we use the string `foo` if none is given".

blueyed:
Pushed a fixup.  Can be squash-merged when it's still green on Travis.

psychon:
Thanks

Issue #1746, opened 2017-04-21
Crash when trying to focus window


petersohn:
Output of `awesome --version`:

awesome v4.1 (Technologic)
 • Compiled against Lua 5.2.4 (running with Lua 5.2)
 • D-Bus support: ✘
 • execinfo support: ✔
 • xcb-randr version: 1.4
 • LGI version: 0.9.0

**How to reproduce the issue:**

1. Put this into `rc.lua`:
```lua
client.connect_signal("unfocus",
        function(c)
            if c.minimized then
                naughty.notify({text="unfocus "..c.name})
                gears.timer.delayed_call(function() client.focus = c end)
            end
        end)
```
2. Start a new window.
3. Minimize it.
4. Awesome crashes.

**Actual result:**

Awesome crashes. Unfortunately, I could not start a standalone X server, so I could not produce a terminal output.

**Expected result:**

I would expect it to restore the minimized window, as it does when the notify is not there. While this exact functionality is normally not very useful, and the reproduction is indeed fabricated, it should not crash in any case like this.

PR #1747, opened 2017-04-21
Fix recursive enterleave ignore

psychon:
Every call to client_ignore_enterleave_events() must be paired with a
following call to client_restore_enterleave_events(). In between these
two calls, no other calls to client_ignore_enterleave_events() is
allowed.

The code in banning_refresh() sometimes broke these rules. This can
happen because the code causes signals to be emitted and Lua code can do
basically anything.

Fix this by moving the calls into the called functions.

Fixes: https://github.com/awesomeWM/awesome/issues/1746

(Memo to myself: Don't mention that this does not have a regression test because then someone will make me write one and I do not have the time right now)

### DIFF
--- a/banning.c
+++ b/banning.c
@@ -52,8 +52,6 @@ banning_refresh(void)
 
     globalconf.need_lazy_banning = false;
 
-    client_ignore_enterleave_events();
-
     foreach(c, globalconf.clients)
         if(client_isvisible(*c))
             client_unban(*c);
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks!
Since CI is green we can just merge this then.. ;)

blueyed:
btw: https://github.com/awesomeWM/awesome/issues/1746 would have a test case, which could be adopted.

PR #1748, opened 2017-04-23
Fix maximize again

Elv13:
Fix 2 bugs, introduce a lesser one.

Prevent `max_h` and `max_v` to be set by the atoms and bypass the safeguards. <s>Also starts to ignore requests *from the clients* after `manage` has been called.</s> There is some `request::geometry` signals if the user want to honor them, but it's their problem, not mine.

### DIFF
--- a/event.c
+++ b/event.c
@@ -341,6 +341,8 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
         uint16_t deco_bottom = bw + tb_bottom;
         int16_t diff_w = 0, diff_h = 0, diff_border = 0;
 
+        lua_State *L = globalconf_get_lua_State();
+
         if(ev->value_mask & XCB_CONFIG_WINDOW_X)
         {
             int16_t diff = 0;
@@ -373,8 +375,6 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
         }
         if(ev->value_mask & XCB_CONFIG_WINDOW_BORDER_WIDTH)
         {
-            lua_State *L = globalconf_get_lua_State();
-
             diff_border = ev->border_width - bw;
             diff_h += diff_border;
             diff_w += diff_border;
@@ -396,7 +396,33 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
         }
 
         c->got_configure_request = true;
-        client_resize(c, geometry, false);
+
+        /* Request the changes to be applied */
+        luaA_object_push(L, c);
+        lua_pushstring(L, "ewmh");     /* context */
+        lua_newtable(L);               /* props */
+
+        /* area, it needs to be directly in the `hints` table to comply with
+           the "protocol"
+         */
+        lua_pushstring(L, "x");
+        lua_pushinteger(L, geometry.x);
+        lua_rawset(L, -3);
+
+        lua_pushstring(L, "y");
+        lua_pushinteger(L, geometry.y);
+        lua_rawset(L, -3);
+
+        lua_pushstring(L, "width");
+        lua_pushinteger(L, geometry.width);
+        lua_rawset(L, -3);
+
+        lua_pushstring(L, "height");
+        lua_pushinteger(L, geometry.height);
+        lua_rawset(L, -3);
+
+        luaA_object_emit_signal(L, -3, "request::geometry", 2);
+        lua_pop(L, 1);
     }
     else if (xembed_getbywin(&globalconf.embedded, ev->window))
     {
--- a/ewmh.c
+++ b/ewmh.c
@@ -554,6 +554,8 @@ ewmh_client_check_hints(client_t *c)
     void *data = NULL;
     xcb_get_property_cookie_t c0, c1, c2;
     xcb_get_property_reply_t *reply;
+    bool is_h_max = false;
+    bool is_v_max = false;
 
     /* Send the GetProperty requests which will be processed later */
     c0 = xcb_get_property_unchecked(globalconf.connection, false, c->window,
@@ -578,7 +580,34 @@ ewmh_client_check_hints(client_t *c)
     {
         state = (xcb_atom_t *) data;
         for(int i = 0; i < xcb_get_property_value_length(reply) / ssizeof(xcb_atom_t); i++)
-            ewmh_process_state_atom(c, state[i], _NET_WM_STATE_ADD);
+            if (state[i] == _NET_WM_STATE_MAXIMIZED_HORZ)
+                is_h_max = true;
+            else if (state[i] == _NET_WM_STATE_MAXIMIZED_VERT)
+                is_v_max = true;
+            else
+                ewmh_process_state_atom(c, state[i], _NET_WM_STATE_ADD);
+    }
+
+    /* Check maximization manually */
+    if (is_h_max && is_v_max) {
+        lua_State *L = globalconf_get_lua_State();
+        luaA_object_push(L, c);
+        client_set_maximized(L, -1, true);
+        lua_pop(L, 1);
+    }
+    else if(is_h_max)
+    {
+        lua_State *L = globalconf_get_lua_State();
+        luaA_object_push(L, c);
+        client_set_maximized_horizontal(L, -1, true);
+        lua_pop(L, 1);
+    }
+    else if(is_v_max)
+    {
+        lua_State *L = globalconf_get_lua_State();
+        luaA_object_push(L, c);
+        client_set_maximized_vertical(L, -1, true);
+        lua_pop(L, 1);
     }
 
     p_delete(&reply);
--- a/ewmh.c
+++ b/ewmh.c
@@ -207,6 +207,28 @@ ewmh_init(void)
                         father, _NET_WM_PID, XCB_ATOM_CARDINAL, 32, 1, &i);
 }
 
+static void
+ewmh_update_maximize(bool h, bool status, bool toggle)
+{
+    lua_State *L = globalconf_get_lua_State();
+
+    if (h)
+        lua_pushstring(L, "client_maximize_horizontal");
+    else
+        lua_pushstring(L, "client_maximize_vertical");
+
+    /* Create table argument with raise=true. */
+    lua_newtable(L);
+    lua_pushstring(L, "toggle");
+    lua_pushboolean(L, toggle);
+    lua_settable(L, -3);
+    lua_pushstring(L, "status");
+    lua_pushboolean(L, status);
+    lua_settable(L, -3);
+
+    luaA_object_emit_signal(L, -3, "request::geometry", 2);
+}
+
 void
 ewmh_init_lua(void)
 {
@@ -333,20 +355,20 @@ ewmh_process_state_atom(client_t *c, xcb_atom_t state, int set)
     else if(state == _NET_WM_STATE_MAXIMIZED_HORZ)
     {
         if(set == _NET_WM_STATE_REMOVE)
-            client_set_maximized_horizontal(L, -1, false);
+            ewmh_update_maximize(true, false, false);
         else if(set == _NET_WM_STATE_ADD)
-            client_set_maximized_horizontal(L, -1, true);
+            ewmh_update_maximize(true, true, false);
         else if(set == _NET_WM_STATE_TOGGLE)
-            client_set_maximized_horizontal(L, -1, !c->maximized_horizontal);
+            ewmh_update_maximize(true, false, true);
     }
     else if(state == _NET_WM_STATE_MAXIMIZED_VERT)
     {
         if(set == _NET_WM_STATE_REMOVE)
-            client_set_maximized_vertical(L, -1, false);
+            ewmh_update_maximize(false, false, false);
         else if(set == _NET_WM_STATE_ADD)
-            client_set_maximized_vertical(L, -1, true);
+            ewmh_update_maximize(false, true, false);
         else if(set == _NET_WM_STATE_TOGGLE)
-            client_set_maximized_vertical(L, -1, !c->maximized_vertical);
+            ewmh_update_maximize(false, false, true);
     }
     else if(state == _NET_WM_STATE_ABOVE)
     {
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -364,11 +364,27 @@ function ewmh.merge_maximization(c, context, hints)
     end
 end
 
+--- Allow the client to move itself.
+--
+-- This is the default geometry request handler when the context is `ewmh`.
+--
+-- @signalhandler awful.ewmh.client_geometry_requests
+-- @tparam client c The client
+-- @tparam string context The context
+-- @tparam[opt={}] table hints The hints to pass to the handler
+function ewmh.client_geometry_requests(c, context, hints)
+    if context == "ewmh" and hints then
+        c:geometry(hints)
+    end
+end
+
+
 client.connect_signal("request::activate", ewmh.activate)
 client.connect_signal("request::tag", ewmh.tag)
 client.connect_signal("request::urgent", ewmh.urgent)
 client.connect_signal("request::geometry", ewmh.geometry)
 client.connect_signal("request::geometry", ewmh.merge_maximization)
+client.connect_signal("request::geometry", ewmh.client_geometry_requests)
 client.connect_signal("property::border_width", repair_geometry)
 client.connect_signal("property::screen", repair_geometry)
 screen.connect_signal("property::workarea", function(s)
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -9,6 +9,7 @@
 local client = client
 local screen = screen
 local ipairs = ipairs
+local timer = require("gears.timer")
 local gtable = require("gears.table")
 local aclient = require("awful.client")
 local aplace = require("awful.placement")
@@ -247,10 +248,6 @@ local context_mapper = {
 function ewmh.geometry(c, context, hints)
     local layout = c.screen.selected_tag and c.screen.selected_tag.layout or nil
 
-    --TODO handle `client_maximize_horizontal` and `client_maximize_vertical`
-    -- so the clients can request being maximized. Currently they can set the
-    -- atom, but it is ignored.
-
     -- Setting the geometry will not work unless the client is floating.
     if (not c.floating) and (not layout == asuit.floating) then
         return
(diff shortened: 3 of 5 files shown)


ewmh.c
```diff
@@ -578,7 +580,34 @@ ewmh_client_check_hints(client_t *c)
     {
         state = (xcb_atom_t *) data;
         for(int i = 0; i < xcb_get_property_value_length(reply) / ssizeof(xcb_atom_t); i++)
-            ewmh_process_state_atom(c, state[i], _NET_WM_STATE_ADD);
+            if (state[i] != _NET_WM_STATE_MAXIMIZED_HORZ)
+                is_h_max = true;
+            else if (state[i] != _NET_WM_STATE_MAXIMIZED_VERT)
+                is_v_max = true;
```
psychon: ....huh? If the client wants to be fullscreen, we maximize it horizontally?

Shouldn't this be `==` instead of `!=`? Now I'm worried about this PR...

tests/_client.lua
```diff
@@ -124,6 +130,13 @@ return function(class, title, sn_rules, callback, resize_increment, args)
     if resize_increment then
         options = options .. "resize_increment,"
     end
+    if args.maximize_before then
+        print("ADDED")
```
psychon: Left over debugging `print`.

lib/awful/ewmh.lua
```diff
@@ -247,6 +247,10 @@ local context_mapper = {
 function ewmh.geometry(c, context, hints)
     local layout = c.screen.selected_tag and c.screen.selected_tag.layout or nil
 
+    --TODO handle `client_maximize_horizontal` and `client_maximize_vertical`
+    -- so the clients can request being maximized. Currently they can set the
+    -- atom, but it is ignored.
```
psychon: This just breaks client-induced maximisation and is thus a regression (well, replaces one regression with another regression). Sorry.

ewmh.c
```diff
@@ -208,6 +208,28 @@ ewmh_init(void)
 }
 
 void
+ewmh_update_maximize(bool h, bool v, bool status, bool toggle)
```
psychon: The `v` argument is just unused and can be dropped (or replaced by an enum)

psychon:
The problem with partial maximisation was that the memento used by `awful.placement` can only undo the last change and so when setting both partial maximisation flags, undoing that only undoes half the maximisation. That's the only problem, right?

How about teaching `awful.placement` to only update the memento once per main loop or once per 0.1 seconds or something like that (and afterwards dropping `c.maximize` as a separate boolean from the C code again)? Is there anything that I miss (besides the slight ugliness of this approach) or could that fix all problems? Should I try to investigate/implement this direction?

Elv13:
> That's the only problem, right?

No, far from it. The bug isn't in partial maximization, that works just fine. It's the "other"/"both" maximization that has always been broken since its inception. Lets go back in history.

#### Awesome 3.0-3.5

It had only partial maximization, it worked fine-ish. The default `rc.lua` had an obvious bug where you could end up toggling between max_h and max_v without the ability to unmaximize. Beside that it worked just fine. The current `placement` code would work just fine if that was left as-is.

#### Awesome "3.6-pre" / release hiatus

It introduced a new "unified" maximization *on top* of the partial maximization. All it did was to do 3 operations. In itself, that was a hack, but kind of worked as long as nobody dug too deep.

It however had a huge design bug. By doing the "maximize" + max_h + max_v, you ended up with 3 iteration of signals being emitted. This had some nice side effects, like "fixing" that gVim bug because the client would flicker very fast. But it started the slippery slope. But the damage was done, now signal handler would fire 3 time per operation.

It has a second design issue where it broke the "undo-ability" of the code. Doing `c.maximized = not c.maximized` and `c.maximized = not c.maximized` again would not restore the client as it was. The old memento broke there, it isn't due to the PRs I introduced later, even if I made it worst.

#### request:: API and Geometry overhaul PR series

Daniel started to implement the request API in 2015 and I took over and "completed" it in 2016. The goal of those many PRs was to get rid of all code paths where Awesome would change the state internally without letting the user have a chance to block it. For the `request::activate` (focus), this has little consequences. But for the position changes (`request::geometry`), this was a much scarier set of changes. There was code all over the place that changed the client position. Many of those code path were "fighting" each other and that caused a bunch of "invisible" position flickering.

The new code path introduced the concept of flexible placement "object" that could (and did) replace all (almost) the drawable geometry code in Awesome. It also implemented a single pipeline where "plugins" could be added to do changes while avoiding the geometry to be changed many time in a row (the `request::geometry` `context`).

For the next 9 months, nobody complained too loudly about any bugs regarding maximization.

#### Awesome 4.0 bug reports

When 4.0 was released, many bugs were reported. Some, indeed, about the new memento implementation. Other about the fact maximization was just kind of broken, which it was. I originally did what you propose, but the patch was an atrocity and I deleted it. Instead, I re-implemented the maximization in the core to mimic how GTK and Qt treat it in their API (as a single operation). The intent was to separate maximize h/v from maximize_both and let the 2 concepts drift apart. I hope(d) this would remove the need to handle max_h and max_v in the code and, in the long run, simplify everything. 

I still think this is the way to go, but I botched 2 important corner case in my implementation. I had no time and I still don't, but there is some pressure form the repeated bug reports so I took care of it. The problem with the first implementation was that it only checked for the "both" case at startup instead of when the atom changed. So basically local testing worked fine but real world did not. This caused the exact bug it tried to fix: maximization state mismatch. Then again, nobody complained until 4.1 was released...


#### This PR

First, sorry for coding anything at 4am, I should not do that and did an obvious copy/paste mistake between 2 version of the patch. But what the patchset does is to fix the 2 corner case I missed. The bug isn't in the `placement` as you think it is. The bug is that the placement is **overwritten** due to the too many signals. Even if I taught (hacked) placement to keep a list it would *still* not work correctly due to some signal misordering. Only making sure all maximizations changes are "atomic" from the Lua point of view will fix the bug. Atomic in the sense when you set one property, only one changes.

As for the regression, I will add the missing commit, it was just too late tomorrow when I stopped being lazy. The timer code isn't very pretty, but at least it hides the X11 nastiness from the higher level code paths. I will also pull the #1606 commits into this as they are spiritually semi-related.

event.c
```diff
@@ -396,7 +396,27 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
...
+        lua_pushstring(L, "y");
+        lua_pushinteger(L, geometry.y);
+        lua_pushstring(L, "width");
+        lua_pushinteger(L, geometry.width);
+        lua_pushstring(L, "height");
+        lua_pushinteger(L, geometry.height);
+
+        lua_rawset(L, -3);
+        luaA_object_emit_signal(L, -3, "request::geometry", 2);
+        lua_pop(L, 1);
```
psychon: ....

lib/awful/ewmh.lua
```diff
@@ -310,10 +307,60 @@ function ewmh.geometry(c, context, hints)
...
+-- @tparam string context The context
+-- @tparam[opt={}] table hints The hints to pass to the handler
+function ewmh.merge_maximization(c, context, hints)
+    if context ~= "client_maximize_horizontal" and context ~= "client_maximize_vertical" then
+        return
+    end
+
+    if not c._delay_maximization then
+        c._delay_maximization = function()
+            -- This ignores unlikely corner cases like mismatching toggles.
```
psychon: No, this ignores `toggles` completely! Instead, it unconditionally sets the state to `false` since the C code always emits `status = false` for toggles (why??).

ewmh.c
```diff
@@ -333,20 +355,20 @@ ewmh_process_state_atom(client_t *c, xcb_atom_t state, int set)
...
     {
         if(set == _NET_WM_STATE_REMOVE)
-            client_set_maximized_horizontal(L, -1, false);
+            ewmh_update_maximize(true, false, false);
         else if(set == _NET_WM_STATE_ADD)
-            client_set_maximized_horizontal(L, -1, true);
+            ewmh_update_maximize(true, true, false);
         else if(set == _NET_WM_STATE_TOGGLE)
-            client_set_maximized_horizontal(L, -1, !c->maximized_horizontal);
+            ewmh_update_maximize(true, false, true);
```
psychon: How about setting status (second argument) to `!c->maximized_horizontal` here? Otherwise... how about just leaving it out since it is meaningless?

tests/test-maximize.lua
```diff
@@ -190,13 +190,35 @@ local steps = {
...
+    function()
+        local c = client.get()[1]
+
+        if not c then return end
+
+        assert(not c.maximized_horizontal)
+        assert(not c.maximized_vertical)
+
         --FIXME for now it is simply ignored. Eventually some delayed execution
         -- based hack could merge both max_h and max_v requests.
```
psychon: This `FIXME` is old and can be removed no?

event.c
```diff
@@ -396,7 +396,27 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
         }
 
         c->got_configure_request = true;
-        client_resize(c, geometry, false);
+
+        /* Request the changes to be applied */
+        luaA_object_push(L, c);
+        lua_pushstring(L, "ewmh");     /* context */
```
psychon: It doesn't matter much and we already get this wrong in other places, but: This is either core X11 or ICCCM, but not EWMH. But whatever...

Elv13:
Rebased, fixed and I added a new commit to test against the copy/paste issue of the last revision.

ewmh.c
```diff
@@ -208,6 +208,28 @@ ewmh_init(void)
 }
 
 void
+ewmh_update_maximize(bool h, bool status, bool toggle)
```
psychon: Can this be made static instead?

tests/test-maximize.lua
```diff
@@ -142,6 +153,90 @@ local steps = {
...
+        -- There is a small 1 frame delay to lower the risk of a race condition
+        if c.maximized then
+            c:kill()
+            return true
+        end
+    end,
+    function()
+        if #client.get() > 0 then return end
+
+        test_client(nil,nil,nil,nil,nil,{maximize_after=true})
```
psychon: Why are you testing `maximize_after` twice? This and the following function does the same thing as the previous two function (modulo something that seems to just be preparing for the following test).

tests/_client.lua
```diff
@@ -136,10 +141,20 @@ return function(class, title, sn_rules, callback, resize_increment, args)
...
     end
+    if args.resize then
+        options = table.concat {
+            options,
+            "resize_after_x=",
+            args.resize.x, ",",
+            "resize_after_y=",
+            args.resize.y, ","
+        }
+    end
```
psychon: Can these values pleased be called `width` and `height` instead of `x` and `y`? (Also in the following commits)

tests/test-maximize.lua
```diff
@@ -219,6 +219,25 @@ local steps = {
...
+        test_client(nil,nil,nil,nil,nil,{resize={x=400, y=400}})
+
+        return true
+    end,
+    function()
+        if #client.get() ~= 1 then return end
+
+        local c = client.get()[1]
+
+        if c.width ~= 400 or c.height < 400 then return end
```
psychon: Why this particular combination of `~=` and `<`?

tests/test-maximize.lua
```diff
@@ -142,6 +153,90 @@ local steps = {
...
+        assert(not c.maximized_vertical)
+
+        -- It might happen in the second try
+        if not c.maximized then return end
+
+        c:kill()
+
+        client.disconnect_signal("request::geometry", awful.ewmh.geometry)
+        client.disconnect_signal("request::geometry", awful.ewmh.merge_maximization)
+        client.connect_signal("request::geometry", geometry_handler)
```
psychon: Can/should this "change some signals"-part be moved into the next function to keep the distinction between "what is a single test" clearer?

tests/test-maximize.lua
Elv13: The titlebar offset

psychon: `select(2, c:titlebar_top())` should give you the size of the titlebar
(or perhaps in a more readable form: `local _, size = c:titlebar_top()`)

Elv13:
rebased/fixed

event.c
blueyed: Good point.  We should probably rename those for 4.2?!

ewmh.c
blueyed: @Elv13 
Please respond to this.

lib/awful/ewmh.lua
blueyed: @Elv13 
Please respond.

lib/awful/ewmh.lua
```diff
@@ -302,10 +303,88 @@ function ewmh.geometry(c, context, hints)
...
+            -- That's likely to be an accident anyway.
+            if c._delayed_max_h and c._delayed_max_v then
+                c.maximized = c._delayed_max_h or c._delayed_max_v
```
blueyed: Isn't this `true` always?

blueyed: And the comment (@psychon is commenting on?!) does not seem to match: this is the case where both were set, and likely to be expected, no?

tests/_client.lua
```diff
@@ -32,8 +32,19 @@ local function open_window(class, title, options)
         }
         window:set_geometry_hints(nil, geom, Gdk.WindowHints.RESIZE_INC)
     end
+    if options.maximize_before then
+        window:maximize()
+    end
```
blueyed: Could/should this also test max_h/max_v by themselves here (and afterwards)?

tests/test-maximize.lua
blueyed: This seems to be addressed.

lib/awful/ewmh.lua
Elv13: This was improved in a later version. Now it only ignores when the requests come from *the client* and mismatch. As far as I know, there there is exactly zero mainstream apps that does that. Neither GTK or Qt allow it. Maybe some weird 90's apps may have some funny behaviors, but none I know of.

ewmh.c
Elv13: It would not work

tests/test-maximize.lua
Elv13: it is

tests/_client.lua
Elv13: No, neither GTK nor Qt support it in their `GtkWindow`/`QWindow` classes. It would require a custom C client to trigger such code path, but it would be a limited (read: none) value as no client I know of ever trigger partial maximization by themselves.

Elv13:
@blueyed Please test this PR for a few days before it's merged. I don't use maximizing so I don't really dogfood this beside the tests.

So now, the following case are supposed to work:

 * Maximization from Awesome (all 4 types: max_h, max_v, max, fullscreen)
 * Maximization from the client **before** it is managed
 * Maximization from the client **after** it is managed
 * Fullscreeen from the client **before** it is managed
 * Fullscreeen from the client **after** it is managed
 * Maximization moving across screens
 * Unmaximization to new screens
 * Partial -> full maximization -> partial == original

blueyed:
I have seen some weird issue today, where dialogs where tiled.
Adding `'type': 'dialog'` to the default list of the floating-rule fixed it, but that might also have been the restart of awesome?!
I could not reproduce it right now anymore.. affected windows where pinentry and `zenity --question`.
Might it be related to this PR?!

Elv13:
Depending when you last updated, https://github.com/awesomeWM/awesome/pull/1776 might also be a candidate. I also seen similar issues, yet I don't have this maximize PR merged on my desktop yet.

blueyed:
Thanks for the pointer to #1776 (which is included here).
I was using b6e6a0489 (with that being just merged) since 5 days - I cannot say for sure, but it was not a problem until more recently, so my guess is that #1776 is not what is causing this, but maybe together with this PR then?!  You also mentioned that all the new signals might cause some trouble before merging it.

I will keep an eye on it, and maybe try with just master then also again.

But apart from that, I am 👍 on merging this already, since it addresses a really annoying bug.

blueyed:
Rebased, with my fixup - let's merge it when it is green.

Elv13:
Don't you want to wait to see if you can reproduce your issue? At some point the cycle of fixes introducing bugs requiring fixes introducing bugs has to stop...

blueyed:
@Elv13 
Agreed, but I cannot really tell, and it can be git-bisect also after this has been merged.

blueyed:
Here we go.. 🎆

psychon:
> I have seen some weird issue today, where dialogs where tiled.
> Adding 'type': 'dialog' to the default list of the floating-rule fixed it, but that might also have been the restart of awesome?!

The workaround really sounds a lot like #1776 to me. However, I am not sure. Why would the dialogs become floating? I guess they are neither fullscreen nor maximized. Do they have minimum size = maximum size in size hints?
Also, assuming #1776 works correctly, it should only make the problem more prominent. Previously, the next re-arrange would have picked up the new non-floating state anyway.
Thus, I would rather guess that #1776 does not work as intended... but no idea why that would be.

If you manage to reproduce, I would be interested in something like `return c.floating, awful.client.property.get(c, "_implicitly_floating")` for the affected client (and perhaps also maximized, fullscreen and `c.fixed`).

blueyed:
> Why would the dialogs become floating?

Aren't they floating / non-tiled by default?
Did you mean "Why would the dialogs become tiled?"?

Thanks for the hints for further investigation.

blueyed:
@psychon 

> actually, why are dialogs floating by default? 

That was my assumption, but could not find it in the source now, too..
Why is `zenity --question` ending up as floating then?  (in case it works)

It happened again (noticed after coming back to awesome running on another X server):
```
_NET_WM_PID(CARDINAL) = 10149
WM_CLASS(STRING) = "zenity", "Zenity"
WM_NAME(STRING) = "Timeout after 1800 seconds"
WM_WINDOW_ROLE:  not found.
WM_TRANSIENT_FOR:  not found.
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DIALOG
_NET_WM_STATE:  not found.
---
class: Zenity
name: Timeout after 1800 seconds
instance: zenity
type: dialog
floating: false
_implicitly_floating: false
maximized: false
fullscreen: false
fixed: false
transient_for: nil
x: 0
y: 35
sticky: false
ontop: false
```

When running `zenity --question` again, it was floating though:
```
floating: true
_implicitly_floating: true
maximized: false
fullscreen: false
fixed: false
```

Should we move this to a new issue?

psychon:
> Should we move this to a new issue?

I think I figured it out. Let's start piece-wise:

> why are dialogs floating by default?

"Old" awesome version: https://github.com/awesomeWM/awesome/blob/088584c8120ba3c93a40bfe8017bfcfc27b6fb82/lib/awful/client.lua#L670
"Current" awesome version: Well, it doesn't have the above so this is why something breaks. To be more specific: If https://github.com/awesomeWM/awesome/blob/c1bcad5f5e1d8911653cd3c84ba56aa16d74034d/lib/awful/client.lua#L678-L694 never runs, then we do not initialize the implicitly floating state correctly.

Proposed patch: Also connect to the "manage" signal here: https://github.com/awesomeWM/awesome/blob/c1bcad5f5e1d8911653cd3c84ba56aa16d74034d/lib/awful/client.lua#L696-L701

No idea why you can only sometimes reproduce this with zenity. Does it change its window type or size hints sometimes a short while after becoming visible / before becoming visible? (=Is there a race between awesome and zenity?)

Let's come back to the original question:

> Should we move this to a new issue?

Feel free. But you can also feel free to test the above patch idea and submit it as a pull request if it works.

blueyed:
It does not happen currently when using `zenity --question`, but with a small script using it, invoked through `spawn`:

```lua
    awful.key({ modkey, "Shift"   }, "e",     function ()
      spawn('awesome-logout')
    end),
```

```sh
% cat =awesome-logout
#!/bin/sh

if [ "$1" != '--force' ]; then
  if ! zenity --question --text "Logout?"; then
    exit
  fi
fi

# Use custom exit code to indicate ~/.dotfiles/usr/bin/session-run-awesome.sh to exit.
echo 'awesome.quit(99)' | awesome-client
```

Your suggested fix helps.  I have created https://github.com/awesomeWM/awesome/pull/1803.

blueyed:
Using `zenity --question` from the Run prompt also triggers it (i.e. it does not become floating).
It is related to my config however, since it does not happen with the default config.

PR #1750, opened 2017-04-23
Don't call gears.wallpaper.maximized with function

### LINKED ISSUE #1749: Default rc.lua calls gears.wallpaper.maximized with a wallpaper function

awesome v4.1 (Technologic)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1

In a theme, set theme.wallpaper to a function (which should probably attempt to set the wallpaper).

**Actual result:**

An error appears at startup, saying cairo has an INVALID_MATRIX. The default rc.lua calls wallpaper if it is a funciton, but then unconditionally passes it to gears.wallpaper.maximized.

**Expected result:**

Wallpaper function is invoked, gears.wallpaper.maximized is not.

ChrisMondok:
If theme.wallpaper is a function, call it, but don't call gears.maximized.wallpaper with it.

Fixes #1749

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -178,8 +178,9 @@ local function set_wallpaper(s)
         -- If wallpaper is a function, call it with the screen
         if type(wallpaper) == "function" then
             wallpaper = wallpaper(s)
+        else
+            gears.wallpaper.maximized(wallpaper, s, true)
         end
-        gears.wallpaper.maximized(wallpaper, s, true)
     end
 end
 


Veratil:
I think the original intent of `theme.wallpaper` being a function was to return a surface for `gears.wallpaper.xxx` to use.

ChrisMondok:
Good catch, I should have gathered that from line 180 saving a reference to the function's return value. I'll close this PR (and associated bug), although I think the comment on 178 could be more clear.

PR #505, opened 2015-10-10
Fix window gravities

psychon:
This makes metacity's `test-gravity.c` behave correctly. The code can be found at 
https://github.com/ccg/metacity/blob/master/src/wm-tester/test-gravity.c.

Related commits which this reverts (mostly): f0ab2aebebfaa391de3efdc9598e5fb23a52e483 ecddee44cb72c2211e188b97886f0a2ca6925ab5 (and likely others)

Issue #1364, opened 2017-01-05
Coverage for C code


blueyed:
We should track coverage of the C code also (after #1239).

I have found the following references when briefly looking into it:
https://github.com/codecov/example-c
https://github.com/bilke/cmake-modules/blob/master/CodeCoverage.cmake
https://github.com/RWTH-ELP/CMake-codecov

I have added `-g;-O0;-fprofile-arcs;-ftest-coverage` to `AWESOME_C_FLAGS`, `CMAKE_CXX_FLAGS` and `CMAKE_C_FLAGS`.
`.gcno`/`.gcda` files were then written to `build/CMakeFiles/awesome.dir/**/*.gcno`.

Elv13:
Open question: Should/can it use a different codecov/coveralls account/profile and run in another CI target?

Running the tests already takes too long. C tests wont slow it down as much as Lua ones, but they will make the slowest targets slower.

blueyed:
> Should/can it use a different codecov/coveralls account/profile and run in another CI target?

It might be possible to upload/merge coverage from another build job for the same build..

PR #1760, opened 2017-04-30
Add a test for gravity handling

psychon:
This adds a C program which tests if the window manager handles
gravities correctly. This program is loosely based on metacity's
test-gravity.c, but completely rewritten and this version does automatic
tests instead of allowing the user to perform testing by hand.

By having this as a self-contained C program, it is possible to compare
awesome's behaviour with the behaviour of other WMs.

In my testing, only metacity and awesome pass this test. This is not
that much of a big surprise since awesome was fixed in
https://github.com/awesomeWM/awesome/pull/505 to work correctly with
metacity's test-gravity.c. However, I am surprised that e.g. Fluxbox
gets this wrong.

I'm not sure how this relates to #1608. This new program didn't actually uncover any new bugs while #1608 seems to be fixing something, but I'm not sure what exactly (maximized? how does that interact with gravities? Do I want to know?)...
My intention was not to cause conflicts or something like that, I was just thinking about #1607 and noticed that we have no tests for gravities. I forgot all about #1608.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -375,12 +375,16 @@ endif()
 # }}}
 
 # {{{ Tests
+add_executable(test-gravity tests/test-gravity.c)
+target_link_libraries(test-gravity
+    ${AWESOME_COMMON_REQUIRED_LDFLAGS} ${AWESOME_REQUIRED_LDFLAGS})
 add_custom_target(check-integration
     sh -c "CMAKE_BINARY_DIR='${CMAKE_BINARY_DIR}' ${CMAKE_SOURCE_DIR}/tests/run.sh"
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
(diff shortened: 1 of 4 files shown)


blueyed:
> Merging #1760 into master will not change coverage.

?

psychon:
> ?

Why should it change coverage? From awesome's POV this is just a new client that resizes and moves itself a bit and then disappears again. Nothing in our Lua code looks at gravity so far. If #1364 were a thing, there might be like 10 lines of C code which would now be covered.

blueyed:
Rebased, let's see if coverage increases now as expected.

psychon:
Thanks for rebasing. The coverage diff now looks more or less as expected (expected: gravity-related things gets covered; unexpected: `tests/gravity.c` is coverage-tracked and this now also tests clients which e.g. have no `WM_CLASS` propperty).

PR #1764, opened 2017-05-01
Fix fullscreen clients with gravity != NorthWest

psychon:
Once upon a time, 4b9584fdb14fc3 already fixed this problem: We have to
set the border width to zero before applying the new geometry to the
client, because changing the border width makes the client move
according to its gravity.

Then came e54387904b16e3 and made this code use awful.placement instead
of just fullscreening the client itself (without explaining why in the
commit message!). After this commit, the border width was just ignored
and left as-is. This was then fixed in 0bf8bb6a64638dec0e (no idea which
callback the commit message refers to, the old code was basically just
c.border_width=0, c:geometry(screen_geo)). However, now the border width
was again changed after the geometry and the bug that was fixed by
4b9584fdb14fc3 was back.

This commit fixes this regression again by making sure that the border
width is set to zero before the geometry is set. This becomes slightly
more complicated, because now it is also awful.placement's job to
restore the old border width.

This is why this commit adds a new option to awful.placement so that it
sets the border width to zero after creating its memento of the old
border width.

Fixes: https://github.com/awesomeWM/awesome/issues/1607

No idea how this PR related to https://github.com/awesomeWM/awesome/pull/1608, but that might be because I am not even sure what that PR tries to do. Also, I don't understand `awful.placement` and even needed some minutes just to figure out where it calls `c:geometry()`. Thus requesting review by @Elv13 to see which part of `awful.placement` I am misunderstanding that would make it possible to do this a lot nicer...

### DIFF
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -292,17 +292,10 @@ function ewmh.geometry(c, context, hints)
 
         if original_context == "fullscreen" and beautiful.fullscreen_hide_border ~= false then
             props.ignore_border_width = true
+            props.zap_border_width = true
         end
 
         aplace[context](c, props)
-
-        -- Remove the border to get a "real" fullscreen.
-        if original_context == "fullscreen" and beautiful.fullscreen_hide_border ~= false then
-            local original = repair_geometry_lock
-            repair_geometry_lock = true
-            c.border_width = 0
-            repair_geometry_lock = original
-        end
     end
 end
 
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -295,7 +295,10 @@ function ewmh.geometry(c, context, hints)
             props.zap_border_width = true
         end
 
+        local original = repair_geometry_lock
+        repair_geometry_lock = true
         aplace[context](c, props)
+        repair_geometry_lock = original
     end
 end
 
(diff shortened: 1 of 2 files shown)


psychon:
> Some documentation about zap in the header would be nice

This is intentionally undocumented. I saw that some other property (I think it was `ignore_border_width`) was basically undocumented, so not documenting something seems to be fine here. Also, this is more of a bad hack than API and as I understood, we are allowed to break undocumented API. Thus, I'm trying to hide my hack.

> Also, the: [geometry_lock] will be required

Done as a fixup.

> A couple tests would be nice. There is a gravity commit in some PR you can cherry pick to add the gravity support to the lua GTK client (or use your new client)

This is the only remaining TODO-item and I agree that some kind of regression test would be nice. I'll look into it when I have some more time (which will not be the coming weekend, nor the weekend afterwards, and possibly not even the weekend in 2.5 weeks).

Elv13:
> This is the only remaining TODO-item and I agree that some kind of regression test would be nice. I'll look into it when I have some more time (which will not be the coming weekend, nor the weekend afterwards, and possibly not even the weekend in 2.5 weeks).

Then you can go ahead and merge and open a new issue or add a comment my "fix this thing again" so I eventually do it. I guess at *some* point we will have enough test to cover this seemingly (but really not) trivial thing...

PR #1766, opened 2017-05-01
Also track coverage information for C code

psychon:
Fixes: https://github.com/awesomeWM/awesome/issues/1364

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,6 +124,14 @@ install:
       fi
       return 0
     }
+  - |
+    do_codecov_gcov() {
+      echo "== do_codecov_gcov in $PWD: $*: =="
+      if [ "$DO_COVERAGE" = "codecov" ]; then
+        travis_retry bash /tmp/codecov-bash -X coveragepy -c -F "$1" || return 1
+      fi
+      return 0
+    }
 
   # Functions for custom Travis folds.
   - |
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!
I think we can just add the `gcov` call to `do_codecov`?!
The first change will upload it for the existing Codecov flags, and the second change as `*_c`.

psychon:
I'm just not sure if your change means that coverage is "accumulated" or if $SOMETHING deletes/overwrites the coverage information files when needed...

Also, your changes broke travis: /home/travis/.travis/job_stages: eval: line 71: syntax error near unexpected token `else'

.travis.yml
```diff
@@ -195,9 +188,7 @@ script:
         (make check-unit-coverage \
          && do_codecov unittests \
          && tests/run.sh \
-         && do_codecov functionaltests \
-         && gcov $(find -name '*.o') \
-         && do_codecov_gcov c_code)
+         && do_codecov functionaltests
```
psychon: Missing `)`, I think.

blueyed:
Should be finally fixed.
I have kept the two commits, so we should be able to compare the outcome(s).
There should be `*_c`, so that the C coverage is also grouped/flagged like the Lua one.

psychon:
(Codecov has some weird problem with the C files; I already reported that two days ago).

Fine by me, but I'm not sure how useful `samples_c` and `unittests_c` is since both do not run the awesome binary (and thus are much faster than functional tests...) (and for some weird reason, `unittests_c` and `samples_c` still get 0.74% coverage?!?).

blueyed:
Let's use your initial approach.  I'll cherry-pick it to master.

psychon:
Thanks for cherry-picking and thanks for looking into this.

PR #1771, opened 2017-05-05
Fix the keyboardlayout widget for a single layout

laomaiweng:
Currently, awful's `keyboardlayout` widget doesn't display anything for setups with a single layout in the XKB map.

This is due to the fact that in `keyboardlayout.update_layout()`, `self._layout` is assigned at index `-1` and so `#self._layout == 0` in `keyboardlayout.update_status()`.
The fix consists in working the other way around: fill `self._layout` from index `1` so that `#self._layout` returns the actual length, and adjust `self._current` by `1` when indexing `self._layout`.

This behavior seems more consistent, and I think it makes more sense for the widget to always display the current layout. The user can always disable/remove the widget if he/she only uses a single layout and doesn't want to have it displayed in the wibar.

### DIFF
--- a/lib/awful/widget/keyboardlayout.lua
+++ b/lib/awful/widget/keyboardlayout.lua
@@ -118,7 +118,9 @@ local function update_status (self)
     self._current = awesome.xkb_get_layout_group();
     local text = ""
     if (#self._layout > 0) then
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!

Issue #1753, opened 2017-04-25
cygwin: proper iconv detection is missing


ncaldwell:
Compiling from the 4.1 stable tarball in cygwin I was getting linker reallocate issues when building awesome. I found IF(APPLE) block at awesomeConfig.cmake:167 and made it true, and was able to compile it.

psychon:
I guess https://github.com/awesomeWM/awesome/blob/2bb07343c6893d29222c87785c063ea15b1e05b2/awesomeConfig.cmake#L163-L198 has to be replaced with a proper check for libiconv (that tests if its available in libc, then in libiconv, then fails). Hopefully CMake was improved by now and our comment is wrong ("CMake does not provide a module for looking up the library").

ncaldwell:
Perhaps instead of checking for the system you could call pkg-config libiconv, and see if it returns a string or returns nothing. Not sure if this possible with cmake.

blueyed:
@ncaldwell 
Would be great if you could investigate there, since you are able to reproduce the issue yourself.

btw: I wonder how good awesome behaves in general on cygwin.. what do you use to run it?

ncaldwell:
Can you confirm for me that `pkg-config --libs libiconv` is what you call to get libiconv packages?

Apparently I can do it with execute_process, which will put stdout into a cmake variable. I can also put stderr in a variable, which would then not get used. It would ensure that changes to cmake are less likely to cause it to break (it looks like some people seem to think that stderr should go in with stdout unless specified otherwise.) Currently not specifying where stderr should go sends it to cmake's stderr, so this shows up in the output:

```
Package libiconv was not found in the pkg-config search path.
Perhaps you should add the directory containing `libiconv.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libiconv' found
```

Calling pkg-config should be fine, as cmake is checking for it earlier in the script... though I've made the assumption that that check stops cmake if it failes.

I'm running a Surface Pro 4, which is why I'm using cygwin. My use case is that work has me living in a terminal and I end up having many open. I've not actually used AwesomeWM before, so I can't provide a comparison, but I've been wanting to try a tiling window manager for a while. At first glance it works pretty well.

psychon:
> Can you confirm for me that pkg-config --libs libiconv is what you call to get libiconv packages?

[...]

PR #1772, opened 2017-05-07
Drop our use of iconv

psychon:
We were only using this for tag names. This means we are assuming that
everything is UTF8, but tag names are provided in the local locale and
need to be translated into UTF8? That makes no sense, so just drop this.

Fixes: https://github.com/awesomeWM/awesome/issues/1753


blueyed:
Thanks!

Issue #318, opened 2015-07-15
Screen ordering can cause issues with where clients spawn


syphoxy:
I have an issue where I can't spawn clients on one of my screens. my screen layout is illustrated below.

```
+-----------+
| 3 | 1 | 2 |
+-----------+
```

the screen I can't spawn windows on is `2`. they always appear on `1`. the reason for this ordering is because I configured X to treat `1` as my primary display. awesome counts forward from there it seems and loops around. that being said if I spawn clients on `3` it works fine.

I've tried using the version of awesome that shipped with Fedora 22 and it was exhibiting the issue. I tried using master. I even tried moving `.config/awesome` to see if it was a configuration issue.

```
~ % awesome --version
awesome v3.5.2-530-g1e9da09 (The Fox)
 • Build: Jul 14 2015 23:13:10 for x86_64 by gcc version 5.1.1 (root@britestar.dariorusso.ca)
 • Compiled against Lua 5.3.0 (running with Lua 5.3)
 • D-Bus support: ✔
```

let me know if I missed anything or if there's something I can do to help troubleshoot this.

thanks!

syphoxy:
I just had a thought and tested around with the ordering of my displays. I moved screen `3` so that it appears after screen `2` in X. the new layout illustrated below:

```
+-----------+
| 1 | 2 | 3 |
+-----------+
```

I no longer have the issue it seems! strange!

psychon:
Could you add this to the end of the default config, try to open a single client and see what it reports? (Code untested)
This should report lots and lots of stuff, especially about how it gets moved between screens.
```lua
function connect(signal)
    client.connect_signal(signal, function(c)
        local g = c:geometry()
        print(signal, c, c.name, c.screen, c.border_width, g.x,g.y,g.width,g.height, debug.traceback())
    end)
end
connect("manage")
connect("property::screen")
connect("property::border_width")
connect("tagged")
connect("untagged")
connect("focus")
connect("unfocus")
connect("list")
for s=1, screen.count() do
    local g = screen[s].geometry
    print("screen",s,g.x,g.y,g.width,g.height)
end
```

rashkov:
I am intermittently having the same problem, I think. I loaded your code psychon, and it actually seemed to jog it into working again. Actually the code had an error in it, complaining that you are trying to index c which is a nil value in line "local g = c:geometry()". I removed it and reloaded the config and the problem was fixed, strangely!

Only two screens set up on my system:
[...]

PR #1780, opened 2017-05-13
Fix setting screen

psychon:
This PR contains the result of the discussion in #318.


psychon:
> One easy way to get rid of the build failure would be to squash. Otherwise it needs to be fixed

I re-ordered the commits to fix that. For reference, this PR previously pointed at fe5c7d1128aebd964ad67ccf39e4bdfab82880f1.

psychon:
Thanks!

PR #1783, opened 2017-05-15
Some changes to our wrapper Makefile

psychon:
This turns our `Makefile` basically into a no-op and tells people to use CMake directly. People who try in-tree builds will end up having CMake overwrite this Makefile, so it might make sense to remove this completely.

My reasoning from the commit message:

>     To build awesome, just use CMake directly. This change also gets rid of
>    the phenomenon that every time the compiler version changes, a new
>    hidden build directory is created. Sometimes even the "build" symlink
>    got out of sync with the directory that the Makefile used for building.

Also, I don't know of any other CMake-using project that does something like this.

(And another commit ended up in here more or less accidentally and removes a broken recommendation from our README)

Edit: Well, of course this breaks Travis a lot. I'll fix this up, but only after I am "clear to go" and no one was a lot against this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,8 @@ $(BUILDLN):
 	test -e $(BUILDLN) || ln -s -f ${builddir} $(BUILDLN)
 
 cmake ${builddir}/CMakeCache.txt:
+	$(ECHO) "Creating build directory and running cmake in it. You can also run CMake directly, if you want."
+	$(ECHO)
 	mkdir -p ${builddir}
 	$(ECHO) "Running cmake…"
 	cd ${builddir} && cmake $(CMAKE_ARGS) "$(@D)" ..
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ all: $(TARGETS) ;
 $(TARGETS): cmake-build
 	ln -s -f $(BUILDDIR)/$@ $@
 
-cmake $(BUILDDIR)/CMakeCache.txt:
+$(BUILDDIR)/CMakeCache.txt:
 	$(ECHO) "Creating build directory and running cmake in it. You can also run CMake directly, if you want."
 	$(ECHO)
 	mkdir -p $(BUILDDIR)
@@ -36,8 +36,8 @@ distclean:
 	$(RM) -r $(BUILDDIR) $(TARGETS)
 	$(ECHO) " done"
 
-%: cmake
+%: $(BUILDDIR)/CMakeCache.txt
 	$(ECHO) "Running make $@…"
 	$(MAKE) -C $(BUILDDIR) $@
 
-.PHONY: cmake-build cmake install distclean tags
+.PHONY: cmake-build install distclean tags
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,3 @@
-builddir=.build-$(shell hostname)-$(shell gcc -dumpmachine)-$(shell gcc -dumpversion)
-
 ifeq (,$(VERBOSE))
     MAKEFLAGS:=$(MAKEFLAGS)s
     ECHO=echo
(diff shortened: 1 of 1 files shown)


blueyed:
I am -0 on this: I think it's nice to have a wrapper Makefile, since just `make` is easier than `mkdir build && cd build && cmake .. && make`.

I would like try to fix it instead, e.g. only run the `cmake` if `CMakeCache.txt` is missing, which basically means to change the `%` target probably.

Regarding the different `.build-*` dirs: we could just use a single `build` dir instead, and remove the symlink?!

Elv13:
> People who try in-tree builds will end up having CMake overwrite this Makefile

IMO we should print a fatal error early in the cmakelists when the binary_dir==source_dir. There isn't only the makefile to worry about. Many other files are overwritten (awesomerc.lua, some lua files). That break subsequent builds due to the source file being gone.

psychon:
Ok, so no removal of that Makefile. Instead, I added a commit that adds a message saying that one can use CMake directly, if wanted. I also tried to do the other Makefile-changes that @blueyed suggested. I'm curious how much stuff will break.
I didn't do the CMake changes suggested by @Elv13, since this PR is (so far...) only touching the Makefile.

Makefile
```diff
@@ -8,40 +6,38 @@ else
 endif
 
 TARGETS=awesome
-BUILDLN=build
+BUILDDIR=build
 
-all: $(TARGETS) $(BUILDLN) ;
+all: $(TARGETS) ;
```
blueyed: This should depend on the build dir.

Makefile
```diff
@@ -8,40 +6,38 @@ else
...
 
 $(TARGETS): cmake-build
-	ln -s -f ${builddir}/$@ $@
-
-$(BUILDLN):
-	test -e $(BUILDLN) || ln -s -f ${builddir} $(BUILDLN)
+	ln -s -f $(BUILDDIR)/$@ $@
```
blueyed: This makes no sense here?!

Makefile
blueyed: And another rule should mkdir it.

Makefile
psychon: Why doesn't this make sense? This is the rule for `./awesome` and it creates `./awesome`.

Makefile
psychon: Why? The dependency chain is `all` -> `awesome` -> `cmake-build` -> `build/CMakeCache.txt`. If `build` does not exist, then `build/CMakeCache.txt` does not exist and so the rule for `build/CMakeCache.txt` is run and `build` is created and `cmake` is run.

psychon:
Rebased on master

Makefile
psychon: And why does it make a difference what `all` depends on? If `all` depends on `./build`, then `all` needs to run if `build` does not exist. However, as I indicated in the previous comment, that's the case anyway. If `build` exists, then that dependency makes no difference.

psychon: Hm... but you are right, there is a problem: If you just change the source code, then I guess that `make` won't be run in `build/`. Or does the wildcard pattern match anyway?

psychon: My last comment is wrong: `cmake-build` is phony and the dependency chain is `all -> awesome -> cmake-build`, so `all` will always cause `make` to run in `build`.

Makefile
psychon: Or are you referring to something else?

psychon:
@blueyed ping

Makefile
blueyed: Ok.

Makefile
blueyed: Ok.

blueyed:
Sorry for the delay, giving it a spin now - running into other unrelated issues with pyenv currently.. (via asciidoc)

blueyed:
Thank you very much!

psychon:
Thanks (for checking this and for merging)

Issue #1786, opened 2017-05-18
Error out when BUILD_DIR = SOURCE_DIR


psychon:
Currently the build fails because we do something like `some_script < awesomerc.lua > awesomerc.lua` and so we end up with an empty default config. (There might also be other things going wrong)

PR #1788, opened 2017-05-24
Abort on in-source-builds

psychon:
They do not work since we overwrite things in such a way that the build
will fail.

Fixes: https://github.com/awesomeWM/awesome/issues/1786


blueyed:
Thanks!

PR #1800, opened 2017-05-29
Emit wallpaper_changed less often

psychon:
Commit 222f0a133cb73a9 optimised the case where we change the wallpaper
so that we can give the new wallpaper to Lua faster. However, contrary
to what was intended, it also caused another wallpaper update later when
the server told us that the wallpaper actually changed. This was because
the juggling with multiple X11 connections went wrong.

Fix this by using the right connection to actually change the wallpaper.


blueyed:
Thanks!

PR #1804, opened 2017-05-30
Run awesome through a shell script

### LINKED ISSUE #1193: AwesomeWM leaves zombie processes
If you asynchronously spawn a program and restart Awesome then started process is left as zombie.

```
awful.spawn.easy_async("sleep 10")
```

restart Awesome.

**Expected result:**
Awesome should not leave these processes.

petersohn:
Fixes #1193.


awesome.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+RESULT_CODE_RESTART=8
+
+result=$RESULT_CODE_RESTART
+while [ $result == $RESULT_CODE_RESTART ]; do
+    awesome
+    result=$?
+done
```
blueyed: Could do `while true; do … ` with a `break` to check the result only after running awesome.

blueyed: Also, `bash` is not really needed.
```diff
diff --git i/awesome.sh w/awesome.sh
index 685501ca..4ef54e1e 100755
--- i/awesome.sh
+++ w/awesome.sh
@@ -1,10 +1,12 @@
-#!/bin/bash
+#!/bin/sh
 
 RESULT_CODE_RESTART=8
 
-result=$RESULT_CODE_RESTART
-while [ $result == $RESULT_CODE_RESTART ]; do
+while true; do
     awesome
     result=$?
+    if ! [ "$result" = "$RESULT_CODE_RESTART" ]; then
+        break
+    fi
 done
 exit $result
```

psychon:
Sigh. Are zombies really that much of a problem? Do they happen so often for you?
How about doing `while (waitpid(-1, NULL, WNOHANG) > 0) {}` once at the beginning of `main` to reap all already-existing zombies? That won't help with processes which become zombies later, but at least this would then mean that a restart later gets rid of all zombies and they do not accumulate. Would that be a satisfying solution for @petersohn and @mindeunix ?

If we really go down the road proposed by this patch (I really do not like the idea), how about implementing this in C instead of as a wrapper shell script? That way we prevent everyone from having to switch to `awesome.sh` instead of `awesome`. Something like `int status; pid_t pid = fork(); if (pid == 0) return; /* continue normally */ if (pid > 0) { waitpid(pid, &status, 0); /* handle status */ } else { perror("fork"); exit(1); }`. This would then make running awesome under gdb or valgrind so much harder...

Elv13:
> Sigh. Are zombies really that much of a problem? Do they happen so often for you?

I had to reboot a few time because I ran out of PIDs and FDs due to that. However note that my config does just about every calamities (I have to reload it /a lot/ for it to happen).

That being said, I am not very warm to the idea of adding such a script. It's a hack around the problem.

And that OpenBSD guy will come on IRC and complain it broke BSD again due to the `#!/bin/bash`.

psychon:
> I had to reboot a few time because I ran out of PIDs and FDs due to that. H

Zombies do not cause FD leaks.

> (I have to reload it /a lot/ for it to happen).

So I guess collecting all already existing zombies once during startup would be enough then. It does not really fix the zombies, but should make sure they do not become a problem.

Elv13:
> Zombies do not cause FD leaks.

You can find creative ways to make it happen nevertheless ;) (however you can kill the subprocess/daemonscript that holds the FDs). Anyway, thanks for the PR

psychon:
> You can find creative ways to make it happen nevertheless ;) (however you can kill the subprocess/daemonscript that holds the FDs). Anyway, thanks for the PR

Could it be that we are talking about different problems here? Zombie processes cannot be killed, because they are already dead. The only reason they exist is because the parent process supposedly wants to be informed about their exit code.

Elv13:
Awesome starts -> start zsh scripts -> spawn various stuff with unrelated ppid -> die in unclean state -> awesome restarts -> start zsh scripts -> spawn additional instances of various stuff with unrelated ppid -> repeat -> ... -> hit the system limits

So, as I said, my bug. I was "creative" and ended up with too many open files. I will eventually get rid of that module (https://github.com/AxelTB/drawer). Please ignore the issue and concentrate on the PID depletion.

petersohn:
I agree that this is a hack, but I found no better solution. To avoid creating a wrapper shell script, I rewrote the same in C. See #1808.

PR #1805, opened 2017-05-31
Collect zombies once at startup

psychon:
When restarting awesome, the new process inherits children processes,
but does not know about them. Thus, it cannot ask glib to inform it
about the children's exit status. This means that when the children
stop, they become zombies that are sitting around for as long as awesome
runs. Over time (with many restarts), such zombies can accumulate and
e.g. fill up the process table.

We cannot do anything about this at runtime (glib doesn't allow us to
catch SIGCHLD and calling waitpid(-1,...) would mean that glib doesn't
get informed when some child that it waits for exits), but at least at
startup we can collect all already-existing zombie children.

This does not really fix the problem, but should make sure that people
who often restart awesome and who also use
awful.spawn.with_line_callback with an exit callback (I think the
problem does not occur without) do not fill up their process table over
time.

Helps-with: https://github.com/awesomeWM/awesome/issues/1193

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -636,6 +636,12 @@ main(int argc, char **argv)
         }
     }
 
+    /* Collect any zombies that we inherited. It would be nice if we also
+     * collected zombies at runtime, but that does not really work with glib.
+     */
+    while (waitpid(-1, NULL, WNOHANG) > 0) {
+    }
+
     /* register function for signals */
     g_unix_signal_add(SIGINT, exit_on_signal, NULL);
     g_unix_signal_add(SIGTERM, exit_on_signal, NULL);


psychon:
Oh, I forgot to mention #1804. Hence: CC @mindeunix @petersohn

petersohn:
I don't think this really solves #1193, as any process that dies after restart leaves behind zombies. Collecting them at startup only addresses a few of them, that happened to die while awesome was restarting.

psychon:
@petersohn My idea was more "prevent this from becoming a problem" since I have no idea how we can do "prevent zombies" without breaking glib.

Any child process that the current awesome instance started will be collected by it. Thus, at runtime the number of zombies is bounded by the number of inherited child process. Any child that was started before restarting awesome, but did not stop will become a zombie. However, there is a fixed number of these and so they shouldn't cause any problems (besides "looking weird" in `top`). Any zombie that awesome already had before restarting will be collected, so eventually zombies will go away (assuming you are restarting awesome often).

> Collecting them at startup only addresses a few of them, that happened to die while awesome was restarting.

...and all the zombies that the previous awesome instance had will be collected as well. So yes, the problem will still show up after one restart, but if you then restart the next time 5 minutes later, the idea is that all these zombies go away.

So this should fix the "huge number of zombies fill up the process table"-problem without making debugging harder.

petersohn:
I still rather not have any zombie processes left behind, so I prefer solving the problem instead. See #1808, which is I think a more elegant solution than #1804.

PR #1806, opened 2017-05-31
Updated porting tips

asmunder:
Fixes issue #1576 - updated porting tips with a section on how to obtain clean 3.5.9 and 4.0 rc.lua files, and using diff with existing config and these files to create new rc.lua

### DIFF
--- a/docs/17-porting-tips.md
+++ b/docs/17-porting-tips.md
@@ -8,6 +8,13 @@ The
 you might have added to your `rc.lua`.
 This avoids most of the possible errors due to missing important changes.
 
+To do this, you can download the default rc.lua for 3.5.9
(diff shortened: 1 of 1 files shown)


psychon:
> I'm not sure I can properly test the markdown (relative link in particular) myself, 

I pulled the branch, built it and looked at the resulting doc. Both links work fine. However, from looking at how other parts of the docs link to this page, I learned that just
```
`rc.lua`
```
works and is turned into the right link by LDoc. Right now I have no idea why that works, but apparently it does. :-)

Keeping this link the way it is is fine with me, so 👍

Elv13:
> works and is turned into the right link by LDoc. Right now I have no idea why that works, but apparently it does. :-)

That's the part I added in my partial fix to #1576 (this is the other half)

PR #1809, opened 2017-06-01
Reap children even across a restart

psychon:
In Unix, so that a process can learn about the exit status of the child
processes that it started, children become zombie processes until the
parents collects their exit information. We use glib both for starting
and for collecting processes. However, when awesome is restarted, the
new instance inherits children, but does not know about them and does
not inherit them.

Fix this by explicitly tracking a list of running child processes and by
serialising them across a restart via an environment variable. The new
awesome instance can then watch for these child processes, but besides
that it ignores them and does not use their exit status in any way.

Thanks to Colin Walters for the hint with serialising the list of processes.

Fixes: https://github.com/awesomeWM/awesome/issues/1193

---

I tested this via the following code appended to the default config:
```lua
gears.timer.start_new(1, function()
	awful.spawn.easy_async("sleep 10", function() end)
	return true
end)
gears.timer.start_new(5, function() awesome.restart() end)
```

CC @mindeunix @petersohn

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -115,6 +115,8 @@ awesome_atexit(bool restart)
     /* Disconnect *after* closing lua */
     xcb_cursor_context_free(globalconf.cursor_ctx);
     xcb_disconnect(globalconf.connection);
+
+    spawn_before_exit(restart);
 }
 
 /** Restore the client order after a restart */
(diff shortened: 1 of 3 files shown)


petersohn:
Awesome. I'll check with my use case soon.

petersohn:
Great, it works for me.

mindeunix:
works for me too, zombie apocalypse ended! :tada:

Elv13:
Thanks!

PR #1823, opened 2017-06-08
Honor DPI for awful.menu

psychon:


lib/awful/menu.lua
```diff
@@ -116,7 +116,7 @@ local function load_theme(a, b)
     ret.border = a.border_color or b.menu_border_color or b.border_normal or
                  fallback.menu_border_color or fallback.border_normal
     ret.border_width= a.border_width or b.menu_border_width or b.border_width or
-                      fallback.menu_border_width or fallback.border_width or 0
+                      fallback.menu_border_width or fallback.border_width or dpi(0)
```
actionless: why?

blueyed: @actionless 
Do you mean why `0` is wrapped?
I understood that it is for consistency, i.e. that it does not get missed when copying/changing this to something non-zero.

actionless: yup, sounds reasonable

PR #1826, opened 2017-06-09
wibar: Fix with width and height beautiful properties

### LINKED ISSUE #1824: beautiful variable `wibar_height` does not affect the height of wibars

awesome v4.1 (Technologic)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1

1. Remove any `height` args from the `awful.wibar({ position = "top", screen = s })` call.
2. Set beautiful theme variable `wibar_height`.

**Actual result:**

`wibar_height` does not affect the height of the wibars.

**Expected result:**

The height of wibars to be the same as `wibar_height`.

Looks like this [line](https://github.com/awesomeWM/awesome/blob/master/lib/awful/wibar.lua#L393) in wibar.lua is settting the height dynamically and the [line](https://github.com/awesomeWM/awesome/blob/master/lib/awful/wibar.lua#L412) here is checking for `arg.height` to be `nil` before getting the value from `wibar_height`, which in this case will never happen.

Elv13:
There was code that always set the values. The fallback was never
executed.

Fix #1824

### DIFF
--- a/lib/awful/wibar.lua
+++ b/lib/awful/wibar.lua
@@ -379,7 +379,8 @@ function awfulwibar.new(arg)
 
     -- Set default size
     if position == "left" or position == "right" then
-        arg.width = arg.width or math.ceil(beautiful.get_font_height(arg.font) * 1.5)
+        arg.width = arg.width or beautiful["wibar_width"]
+            or math.ceil(beautiful.get_font_height(arg.font) * 1.5)
         if arg.height then
             has_to_stretch = false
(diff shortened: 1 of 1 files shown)


psychon:
How about also removing `width` and `height` from the table in line 410?

Elv13:
> How about also removing width and height from the table in line 410?

Done in the github UI, use the squash option if you decide to merge this

PR #1829, opened 2017-06-10
hotkeys_popup: keys for qutebrowser

sim590:
Based on the qutebrowser [cheatsheet](https://github.com/qutebrowser/qutebrowser/blob/master/doc/img/cheatsheet-big.png), I have produced this popup keys file for qutebrowser.

### DIFF
--- a/lib/awful/hotkeys_popup/keys/qutebrowser.lua
+++ b/lib/awful/hotkeys_popup/keys/qutebrowser.lua
@@ -1,0 +1,187 @@
+---------------------------------------------------------------------------
+--- Qutebrowser hotkeys for awful.hotkeys_widget
+--
+-- @author Simon Désaulniers sim.desaulniers@gmail.com
+-- @copyright 2017 Simon Désaulniers
+-- @module awful.hotkeys_popup.keys.qutebrowser
+---------------------------------------------------------------------------
+
+local hotkeys_popup = require("awful.hotkeys_popup.widget")
+
+local qutebrowser_rule = {class="qutebrowser"}
+for group_name, group_data in pairs({
+    ["Qutebrowser: copying/yanking"]    = { rule = qutebrowser_rule },
+    ["Qutebrowser: pasting"]            = { rule = qutebrowser_rule },
+    ["Qutebrowser: navigation"]         = { rule = qutebrowser_rule },
+    ["Qutebrowser: scrolling"]          = { rule = qutebrowser_rule },
+    ["Qutebrowser: in prompt mode"]     = { rule = qutebrowser_rule },
+    ["Qutebrowser: opening"]            = { rule = qutebrowser_rule },
+    ["Qutebrowser: back/forward"]       = { rule = qutebrowser_rule },
+    ["Qutebrowser: extended hint mode"] = { rule = qutebrowser_rule },
+    ["Qutebrowser: misc. commands"]     = { rule = qutebrowser_rule },
+    ["Qutebrowser: modifier commands"]  = { rule = qutebrowser_rule },
+    ["Qutebrowser: in insert mode"]     = { rule = qutebrowser_rule },
+    ["Qutebrowser: in command mode"]    = { rule = qutebrowser_rule },
+}) do
+    hotkeys_popup.add_group_rules(group_name, group_data)
+end
+
+local qutebrowser_keys = {
+    ["Qutebrowser: copying/yanking"] = {{
+        modifiers = {},
+        keys = {
+            yy = "copy/yank URL",
+            yY = "copy URL to selection",
+            yt = "copy title to clipboard",
+            yT = "copy title to selection"
+        }
+    }},
+
+    ["Qutebrowser: pasting"] = {{
+        modifiers = {},
+        keys = {
+            pp = "open URL from clipboard",
+            pP = "open URL from slection",
+            Pp = "like pp, in new tab",
+            PP = "like pP, in new tab",
+            wp = "like pp, in new window",
+            wP = "like pP, in new window"
+        }
+    }},
+
+    ["Qutebrowser: navigation"] = {{
+        modifiers = {},
+        keys = {
+            ["[["]  = "click previous link on page",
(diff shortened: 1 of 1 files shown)


sim590:
I have made a change in one of the key text entries. It seems like the text of a field `key` cannot contain the char `&`. Otherwise, all of the keys in the same section are ignored.

lib/awful/hotkeys_popup/keys/qutebrowser.lua
```diff
@@ -0,0 +1,175 @@
...
+            ["}}"]  = "like ]], in new tab",
+            ["c-a"] = "increment no. in URL",
+            ["c-x"] = "decrement no. in URL"
+        }
+    }},
+
+    ["Qutebrowser: scrolling"] = {{
+        modifiers = {},
+        keys = {
+            ["c-f"] = "page down",
```
actionless: use `modifiers = {"Control"}`
and `f  = "page down"`

sim590: Done.

lib/awful/hotkeys_popup/keys/qutebrowser.lua
```diff
@@ -0,0 +1,186 @@
...
+        modifiers = {"Control"},
+        keys = {
+            p = "prev. history item",
+            n = "next history item"
+        }
+    }},
+}
+
+hotkeys_popup.add_hotkeys(qutebrowser_keys)
+
```
blueyed: Extra newline.

Should have our Vim modeline instead/additionally?!

lib/awful/hotkeys_popup/keys/qutebrowser.lua
```diff
@@ -0,0 +1,186 @@
...
+--- Qutebrowser hotkeys for awful.hotkeys_widget
+--
+-- @author Simon Désaulniers sim.desaulniers@gmail.com
+-- @copyright 2017 Simon Désaulniers
+-- @module awful.hotkeys_popup.keys.qutebrowser
+---------------------------------------------------------------------------
+
+local hotkeys_popup = require("awful.hotkeys_popup.widget")
+
+local qutebrowser_rule_any = {class={"__main__.py", "qutebrowser"}}
```
blueyed: This will match any `__main__.py` class, no?
When is `__main__.py` used in qutebrowser?
I get `WM_CLASS(STRING) = "qutebrowser", "qutebrowser"` for the main window.

sim590: You're right. I bet I get `__main__.py` because I'm running the qutebrowser from the repository instead of using out of date debian packages. I'm changing the class line to the one you suggested.

lib/awful/hotkeys_popup/keys/qutebrowser.lua
sim590: Okay. I'll add the vim modeline instead.

lib/awful/hotkeys_popup/keys/qutebrowser.lua
blueyed: I am getting these:
```
class: qutebrowser
name: DuckDuckGo - qutebrowser
instance: qutebrowser
```

Is instance also not `qutebrowser` for you?

Matching `name` with `- qutebrowser` at the end might also work?!

btw: found this in the source / changelog
> - New `--qt-name` argument to qutebrowser which is passed to Qt to set `WM_CLASS`.

lib/awful/hotkeys_popup/keys/qutebrowser.lua
```diff
@@ -0,0 +1,187 @@
...
+--- Qutebrowser hotkeys for awful.hotkeys_widget
+--
+-- @author Simon Désaulniers sim.desaulniers@gmail.com
+-- @copyright 2017 Simon Désaulniers
+-- @module awful.hotkeys_popup.keys.qutebrowser
+---------------------------------------------------------------------------
+
+local hotkeys_popup = require("awful.hotkeys_popup.widget")
+
+local qutebrowser_rule_any = {class={"qutebrowser", "qutebrowser"}}
```
blueyed: You just need it once here.
And not `rule_any` if it should just match a single class.

I suggest using `instance = "qutebrowser"`, if that works for you.
(and not `rule_any` then)

blueyed: btw: https://github.com/blueyed/dotfiles/blob/master/usr/bin/xpop

blueyed: See https://awesomewm.org/doc/api/libraries/awful.rules.html#rules for the documentation.

I've meant to use:
```
["Qutebrowser: copying/yanking"]    = { rule = {instance = "qutebrowser"} },
```

or

```
local qutebrowser_rule = {instance={"qutebrowser"}}
…
["Qutebrowser: copying/yanking"]    = { rule = qutebrowser_rule},
```

(basically matching the WM "instance" instead of "class" (both are in WM_CLASS, and usually different, but for qutebrowser apparently the same))

sim590: On my end, using `instance` keyword doesn't work. However, if I do the last thing you suggested, but using the keyword `class`, it's fine.

sim590: I have tried your script and I get 
```
class: qutebrowser
name: https://raw.githubusercontent.com/blueyed/dotfiles/master/usr/bin/xpop - qutebrowser
instance: __main__.py
```

sim590: Therefore, I guess the way to go is to use the `class` field.

blueyed: Cool!
Use a normal `rule` then instead of `rule_any`, and just the class (and not a table/list).

actionless:
@blueyed looks good for you now?

PR #1830, opened 2017-06-10
hotkeys_popup: keys for tmux

nahsi:
More hotkeys!

### DIFF
--- a/lib/awful/hotkeys_popup/keys/tmux.lua
+++ b/lib/awful/hotkeys_popup/keys/tmux.lua
@@ -1,0 +1,90 @@
+---------------------------------------------------------------------------
+--- tmux hotkeys for awful.hotkeys_widget
+--
+-- @author nahsi nashi@airmail.cc
+-- @copyright 2017 nahsi
+-- @module awful.hotkeys_popup.keys.tmux
+---------------------------------------------------------------------------
+
+local hotkeys_popup = require("awful.hotkeys_popup.widget")
+
+local tmux = {}
+
+--- Add rules to match tmux session.
+-- For example:
+--
+--     tmux.add_rules_for_terminal({ rule = { name = { "tmux" }}})
+--
+-- will show tmux hotkeys for any window that has 'tmux' in it's title.
+-- If no rules passed then tmux hotkeys will always be shown!
+-- @function add_rules_for_terminal
+-- @see awful.rules.rules
+-- @param rule Rules to match a window with tmux session
+function tmux.add_rules_for_terminal(rule)
+    for group_name, group_data in pairs({
+        ["tmux: sessions"] = rule,
+        ["tmux: windows"]  = rule,
+        ["tmux: panes"]    = rule,
+        ["tmux: misc"]     = rule,
+    }) do
+        hotkeys_popup.add_group_rules(group_name, group_data)
+    end
+end
+
+local tmux_keys = {
+    ["tmux: sessions"] = {{
+        modifiers = {},
+        keys = {
+            s     = "show all sessions",
+            ['$'] = "rename the current session",
+            ['('] = "move to previous session",
+            [')'] = "move to next session",
+            d     = "detach from current session"
+        }
+    }},
+
+    ["tmux: windows"] = {{
+        modifiers = {},
+        keys = {
+            c         = "create window",
+            f         = "find window",
+            [',']     = "rename current window",
+            --['&']     = "close current window",
+            p         = "previous window",
+            n         = "next window",
+            ['0...9'] = "select window by number"
+        }
+    }},
+
+    ["tmux: panes"] = {{
+        modifiers = {},
+        keys = {
+            [';']       = "toggle last active pane",
+            ['%']       = "split pane vertically",
+            ['"']       = "split pane horizontally",
+            ['{']       = "move the current pane left",
+            ['}']       = "move the current pane right",
+            ['q 0...9'] = "select pane by number",
+            o           = "toggle between panes",
+            z           = "toggle pane zoom",
+            ['space']   = "toggle between layouts",
+            ['!']       = "convert pane into a window",
+            x           = "close current pane"
+        }
+    }},
+
+    ["tmux: misc"] = {{
+        modifiers = {},
+        keys = {
+            [':'] = "enter command mode",
+            ['?'] = "list shortcuts",
+            t     = "show clock"
+        }
+    }}
+}
+
+hotkeys_popup.add_hotkeys(tmux_keys)
+
+return tmux
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80


lib/awful/hotkeys_popup/keys/tmux.lua
```diff
@@ -0,0 +1,73 @@
...
+---------------------------------------------------------------------------
+
+local hotkeys_popup = require("awful.hotkeys_popup.widget")
+
+local tmux_rule_any = {name={"tmux"}}
+for group_name, group_data in pairs({
+    ["tmux: sessions"] = { rule_any=tmux_rule_any },
+    ["tmux: windows"]  = { rule_any=tmux_rule_any },
+    ["tmux: panes"]    = { rule_any=tmux_rule_any },
+    ["tmux: misc"]     = { rule_any=tmux_rule_any },
```
blueyed: Use a simple "rule" here instead of "rule_any", see https://github.com/awesomeWM/awesome/pull/1829/files#r121262482.

blueyed: /cc @actionless 
I hope that makes sense?!

lib/awful/hotkeys_popup/keys/tmux.lua
```diff
@@ -0,0 +1,73 @@
...
+--- tmux hotkeys for awful.hotkeys_widget
+--
+-- @autor nahsi nashi@airmail.cc
+-- @copyright 2017 nahsi
+-- @module awful.hotkeys_popup.keys.tmux
+---------------------------------------------------------------------------
+
+local hotkeys_popup = require("awful.hotkeys_popup.widget")
+
+local tmux_rule = { name = "tmux" }
```
actionless: but it's never like that even with default tmux config

as a solution to this a have in mind only to turn lines 11-19 into a function which will have rule as an argument, so any user can call it with any arguments to match the terminal

for example, in my case it would be like that:

```
rc.lua
<...>
local tmux_keys = require("awful.hotkeys_popup.keys.tmux")
tmux_keys.add_terminal_rule({class="st-256color"})
<...>
```

nahsi: > but it's never like that even with default tmux config
Really? I copied that from vim.lua hotkeyfile. (i.e it's not a problem there). Also I have a default 'tmux' config and 'termite' terminal, 'tmux' always shows up in a title. Tried xfce4-terminal also and 'tmux' showed up. 

```
if tmux_rule == nil then
    t_rule = { name = "tmux" }
else
    t_rule = tmux_rule
end
for group_name, group_data in pairs({
    ["tmux: sessions"] = { rule = t_rule },
    ["tmux: windows"]  = { rule = t_rule },
    ["tmux: panes"]    = { rule = t_rule },
    ["tmux: misc"]     = { rule = t_rule },
}) do
    hotkeys_popup.add_group_rules(group_name, group_data)
end

```

Is this acceptable? The user needs to define  variable with rule, if no variable defined then `name = "tmux" is used`.

actionless: if you mean global variable then it's not acceptable

nahsi: Oh well, then I'll do some research on how to implement what you suggested (I don't know lua or any other programming language) + some sort of default option if there is no function call.

actionless: smth like that:

```lua
local tmux_keys = {}

<...>
function tmux_keys.add_rules_for_terminal(t_rule)
    for group_name, group_data in pairs({
        ["tmux: sessions"] = { rule = t_rule },
        ["tmux: windows"]  = { rule = t_rule },
        ["tmux: panes"]    = { rule = t_rule },
        ["tmux: misc"]     = { rule = t_rule },
    }) do
        hotkeys_popup.add_group_rules(group_name, group_data)
    end
end

<...>

return tmux_keys
```

blueyed: I think it should pass in the whole of `group_data` then, so that you can use `rule_any` etc.

I am using the following to match the main tmux session for the current desktop / X session with my run-or-raise:
```lua
        { rule_any = {
          name = {string.format('^%%[%s:.*:tmux%%].*', tmux_session),
                  string.format('^%%[%s:.*:tmux%%].*', 'some-project')}
          },
          except_any = { name = { "^%[weechat:",
                                  "^%[IGNORE:" } },
        },
        matcher=awful.rules.matches
```

That might not be really relevant, but would allow for more flexibility in general.

For my use case `{ rule = { name = "^%[tmux: .*" }` would work in general, but would not match the "weechat" tmux session then.

actionless: @blueyed  btw since you're using the same naming convention for you client names with tmux sessions, mb you will find this helprer useful :)
https://github.com/actionless/awesome_config/blob/devel/utils/tmux.lua

actionless:
Good!

Now add a docstring to your new function.

Example:
https://github.com/awesomeWM/awesome/blob/master/lib/awful/tag.lua#L276-L287

lib/awful/hotkeys_popup/keys/tmux.lua
```diff
@@ -0,0 +1,77 @@
+---------------------------------------------------------------------------
+--- tmux hotkeys for awful.hotkeys_widget
+--
+-- @autor nahsi nashi@airmail.cc
```
actionless: should be `author`

that's causing Travis build to fail

after fixing that please also squash all your commits

nahsi:
I think I'm done! Hope it's alright. 
Thanks for your patience and all the help, by the way!

nahsi:
If I want to add a link to awful rules table, how should I do it?
`@see awful.rules.rules` ?
Something causing travis build to fail again. I guess it's a mistake in docstring?

blueyed:
@nahsi 
it's a luacheck issue: https://travis-ci.org/awesomeWM/awesome/jobs/242060826#L4039
```
Checking lib/awful/hotkeys_popup/keys/tmux.lua    1 error

    lib/awful/hotkeys_popup/keys/tmux.lua:80:13: expected '}' (to close '{' on line 77) near 't'
```

actionless:
looking good now 👍

actionless:
@blueyed ?

blueyed:
Thanks!

Merged with some amendments:
```diff
diff --git c/lib/awful/hotkeys_popup/keys/tmux.lua i/lib/awful/hotkeys_popup/keys/tmux.lua
index 091cf7b4..16d93a1e 100644
--- c/lib/awful/hotkeys_popup/keys/tmux.lua
+++ i/lib/awful/hotkeys_popup/keys/tmux.lua
@@ -11,15 +11,16 @@ local hotkeys_popup = require("awful.hotkeys_popup.widget")
 local tmux = {}
 
 --- Add rules to match tmux session.
+--
 -- For example:
 --
 --     tmux.add_rules_for_terminal({ rule = { name = { "tmux" }}})
 --
--- will show tmux hotkeys for any window that has 'tmux' in it's title.
--- If no rules passed then tmux hotkeys will always be shown!
+-- will show tmux hotkeys for any window that has 'tmux' in its title.
+-- If no rules are provided then tmux hotkeys will be shown always!
 -- @function add_rules_for_terminal
 -- @see awful.rules.rules
--- @param rule Rules to match a window with tmux session
+-- @tparam table rule Rules to match a window containing a tmux session.
 function tmux.add_rules_for_terminal(rule)
     for group_name, group_data in pairs({
         ["tmux: sessions"] = rule,
```

blueyed:
@nahsi 
JFI: would have pushed it as a fixup first to make sure CI passes still, but you created the PR from your `master` branch, which I did not want to push to.
Please use dedicated branches for PRs in the future.

nahsi:
Cool! Will do! Thank you very much, I've learned a lot.

blueyed:
@nahsi 
Cool!
Thanks for your contribution - awesome is happy to be one of your firsts.. ;)

PR #1832, opened 2017-06-11
debug: Add a signal for deprecation warnings

### LINKED ISSUE #1612: `awful.util.deprecate` should submit signal "debug::deprecate"
## The Problem

AWM library functions `awful.util.deprecate` and `awful.util.deprecate_class` does not emit the signal `debug::deprecation`.

## The Motive

[The description](https://awesomewm.org/doc/api/libraries/awesome.html#debug::deprecation) in the AWM docs on `debug::deprecation` is:

> **debug::deprecation**
>
> A deprecated lua function was called

Since users might want a `naughty` notification or a similar means of communications to know whenever a deprecated function was called, they may use this signal to detect a deprecated library function. Alas, this would not be the case.

## The Solution

`awful.util.deprecate` and `awful.util.deprecate_class` will signal either the existing `debug::deprecation` signal or a new signal called `debug::awesome_deprecation`. This signal will have **three arguments**:

* The full name of the deprecated function (eg. `'awful.util.spawn'`)
* A suggestion of what to use instead (eg. `'awful.spawn'` or `'awful.spawn.easy_async'`)
* The location of where it was called if possible (eg. `{24, 'rc.lua'}` or `'24: rc.lua'` or `nil` if impossible to find)

Thank you for your time for reading this.

Elv13:
Fix #1612

### DIFF
--- a/lib/gears/debug.lua
+++ b/lib/gears/debug.lua
@@ -75,6 +75,10 @@ end
 
 local displayed_deprecations = {}
 --- Display a deprecation notice, but only once per traceback.
+--
+-- This function also emits the `debug::deprecate` signal on the `awesome`
+-- global object.
+--
 -- @param[opt] see The message to a new method / function to use.
 -- @tparam table args Extra arguments
 -- @tparam boolean args.raw Print the message as-is without the automatic context
@@ -108,6 +112,10 @@ function debug.deprecate(see, args)
         end
     end
     debug.print_warning(msg .. ".\n" .. tb)
+
+    if awesome then
+        awesome.emit_signal("debug::deprecation", msg, see, args)
+    end
 end
 
 --- Create a class proxy with deprecation messages.
--- a/luaa.c
+++ b/luaa.c
@@ -78,8 +78,10 @@ extern const struct luaL_Reg awesome_mouse_meta[];
 
 /** A deprecated Lua function was called.
  *
- * @param hint String with a hint on what to use instead of the
+ * @tparam string hint String with a hint on what to use instead of the
  * deprecated functionality.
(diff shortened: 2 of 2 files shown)


psychon:
Doesn't the signal need some `@signal` documentation somehow? (I'm not sure)
The C code already has `debug::deprecation` with just a string argument (macro `luaA_deprecate`). Can we somehow merge `debug::deprecate` and `debug::deprecation`? (btw: I don't like the (current) name `debug::deprecation` at all, but the signal is documented... can we deprecate the deprecation signal? :-) ).

lib/gears/debug.lua
```diff
@@ -75,6 +75,10 @@ end
 
 local displayed_deprecations = {}
 --- Display a deprecation notice, but only once per traceback.
+--
+-- This function also emit the `debug::deprecate` signal on the `awesome`
```
blueyed: s/emit/emits/

lib/gears/debug.lua
```diff
@@ -108,6 +112,10 @@ function debug.deprecate(see, args)
         end
     end
     debug.print_warning(msg .. ".\n" .. tb)
+
+    if awesome then
+        awesome.emit_signal("debug::deprecate", msg, see, args)
+    end
```
blueyed: Why the `if` check?
Might it be called too early?

Elv13: In theory, `wibox` doesn't depend on `capi.awesome`, I just don't want to add dependencies that makes the code harder to use/test on it's own. For example, one day we might want to run some wibox in their own process to improve latency. For now that's mostly to make the tests shims simpler.

Elv13:
Done

blueyed:
Thanks!

PR #1833, opened 2017-06-11
quality: Remove trailing white spaces

### LINKED ISSUE #1129: We use trailing whitespaces, but do not allow them
Hi,

I can no longer edit files like `lib/awful/widget/watch.lua`. Whenever I do that, the trailing whitespaces in there disappear. I don't remember where I read it, but I think @Elv13 somewhere wrote that these trailing whitespaces are necessary so that LDoc doesn't think a new paragraph starts (or something like that). So I had to figure out why vim kills these trailing whitespaces.

The reason is here: https://github.com/awesomeWM/awesome/blob/502b4139bad5f91306505b3ca148ae00deafae89/.editorconfig#L14-L15

So, am I right that we do not allow trailing whitespaces, but require them?
What should we do about this?

Elv13:
Fix #1129

### DIFF
--- a/docs/03-declarative-layout.md
+++ b/docs/03-declarative-layout.md
@@ -175,7 +175,7 @@ Code:
 
     local tb = wibox.widget.textbox()
     tb:set_markup("Hello world! ")
-    
+
     -- Repeat "tb" 3 times
     s.mywibox : setup {
         tb,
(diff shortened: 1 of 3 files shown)


blueyed:
That was also my impression.
@Elv13 
Please confirm that it will not break the docs, and then feel free to merge it directly.

Elv13:
It create issues in the `tests/examples` files, not elsewhere. The space was used as market to create empty lines in the markdown result. Otherwise the newlines are deleted. My rewrite of that system does a less hacky thing, but it still require lots of time to fix and isn't my top priority.

PR #1834, opened 2017-06-11
layoutbox: Use a textbox as the fallback when no images are found

### LINKED ISSUE #896: layoutbox: fallback to textbox in case the image cannot be loaded
If the layoutbox' imagebox fails to load the image(s) (because the path in `theme.layout_*` is wrong), you end up with a layoutbox of 0x0 px.

I think that there should be a placeholder / indicator in this case - maybe falling back to displaying some text?! (e.g. by replacing the imagebox with a textbox)

What do you think?

Elv13:
Fix #896

### DIFF
--- a/lib/awful/widget/layoutbox.lua
+++ b/lib/awful/widget/layoutbox.lua
@@ -11,7 +11,8 @@ local capi = { screen = screen, tag = tag }
 local layout = require("awful.layout")
 local tooltip = require("awful.tooltip")
 local beautiful = require("beautiful")
-local imagebox = require("wibox.widget.imagebox")
+local wibox = require("wibox")
+local surface = require("gears.surface")
 
 local function get_screen(s)
     return s and capi.screen[s]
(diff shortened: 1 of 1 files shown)


lib/awful/widget/layoutbox.lua
```diff
@@ -25,7 +26,12 @@ local function update(w, screen)
     screen = get_screen(screen)
     local name = layout.getname(layout.get(screen))
     w._layoutbox_tooltip:set_text(name or "[no name]")
-    w:set_image(name and beautiful["layout_" .. name])
+
+    local exists = name and beautiful["layout_" .. name]
+        and fs.file_readable(beautiful["layout_" .. name])
+
+    w.imagebox:set_image(exists and beautiful["layout_" .. name] or nil)
+    w.textbox:set_text(exists and  "" or name)
```
psychon: This version is fine with me, but how about the following instead (could catch more problems, but is untested):
```lua
local img = gears.surface.load_silently(beautiful["layout_" .. name], false)
if img then
    w.imagebox.image = img
    w.textbox.text = ""
else
    w.imagebox.image = nil
    w.textbox.text = name
end
```

Elv13:
Done

Issue #1837, opened 2017-06-11
menubar no longer works with Lua 5.1


psychon:
Sigh.

Just seen on IRC:
```
2017-06-10 09:47:30 E: Error during a protected call: attempt to yield across metamethod/C-call boundary
stack traceback:
        [C]: in function 'yield'
        /usr/share/lua/5.1/lgi/override/Gio.lua:180: in function 'async_enumerate_children'
        /usr/share/awesome/lib/menubar/utils.lua:269: in function 'parser'
        /usr/share/awesome/lib/menubar/utils.lua:303: in function </usr/share/awesome/lib/menubar/utils.lua:301>
        (tail call): ?
        [C]: in function 'xpcall'
        /usr/share/awesome/lib/gears/protected_call.lua:30: in function </usr/share/awesome/lib/gears/protected_call.lua:28>
        (tail call): ?
2017-06-10 09:47:30 W: awesome: luaA_panic:529: unprotected error in call to Lua API (attempt to call a nil value)
2017-06-10 09:47:30 W: awesome: luaA_panic:532: dumping backtrace
awesome(backtrace_get+0x47) [0x55db1217d40b]
awesome(+0x1acd6) [0x55db12172cd6]
/usr/lib/x86_64-linux-gnu/liblua5.1.so.0(+0xbd3a) [0x7f36020c0d3a]
/usr/lib/x86_64-linux-gnu/liblua5.1.so.0(lua_error+0x9) [0x7f36020bd6a9]
/usr/lib/x86_64-linux-gnu/lua/5.1/lgi/corelgilua51.so(+0x94c6) [0x7f35fca524c6]
/usr/lib/x86_64-linux-gnu/libffi.so.6(ffi_closure_unix64_inner+0x1cf) [0x7f3600acfe2f]
/usr/lib/x86_64-linux-gnu/libffi.so.6(ffi_closure_unix64+0x46) [0x7f3600ad01a0]
/usr/lib/x86_64-linux-gnu/libgio-2.0.so.0(+0x8a503) [0x7f36011ad503]
/usr/lib/x86_64-linux-gnu/libgio-2.0.so.0(+0x8a539) [0x7f36011ad539]
/lib/x86_64-linux-gnu/libglib-2.0.so.0(g_main_context_dispatch+0x15a) [0x7f3604b3a6aa]
/lib/x86_64-linux-gnu/libglib-2.0.so.0(+0x4aa60) [0x7f3604b3aa60]
[...]

PR #1838, opened 2017-06-11
Fix menubar under Lua 5.1

psychon:
LGI's async support was trying to yield inside a protected call. Lua 5.1
cannot do that. Work around this by reverting to the behaviour before
commit 50cfa6c: Only call the callback in a protected context.

Fixes: https://github.com/awesomeWM/awesome/issues/1837

---

The only reason that this is WIP because I haven't tested this in any way. And for that, I'd like to have support by Travis. And for that, I guess I am blocked by #1072.

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -43,6 +43,28 @@ utils.wm_name = "awesome"
 
 -- Private section
 
+local do_protected_call, call_callback
+do
+    -- Lua 5.1 cannot yield across a protected call. Instead of hardcoding a
+    -- check, we check for this problem: The following coroutine yields true on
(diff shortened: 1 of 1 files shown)


blueyed:
It is mostly about getting https://github.com/awesomeWM/awesome/pull/1072 back on track for the tests, isn't it?

Elv13:
> It is mostly about getting #1072 back on track for the tests, isn't it?

Yes, it's already the 3rd (I think) regression since 4.0. We really really need coverage on this. It's the ugliest code of the project. I still suggest we should drop our implementation and use the GTK once. It is much faster and more reliable. It would solve almost all open issues and the code already exists.

PR #1840, opened 2017-06-11
fix(awful: hotkeys_popup: keys): require all modules in init.lua

actionless:
allows to enable all the available hotkey maps by `require("awful.hotkeys_popup.keys")`


actionless:
i'll also update this one after this https://github.com/awesomeWM/awesome/pull/1830/files will get merged

and mb update rc.lua

PR #1842, opened 2017-06-13
Port lgi-check.sh to C

psychon:
Previously, the lgi check used the normal Lua interpreter to check if
lgi is installed. However, nothing ensures/requires that awesome is
built against the same Lua version as the Lua interpreter. This means
that if lgi is only available for some Lua version, then the check could
succeed even though awesome would later fail to start. Also, the check
might have failed even though awesome would not have any problems
finding lgi.

This commit replaces lgi-check.sh by a small C program which does the
same thing. This ensures that the same Lua version is used as awesome
will be using.

There are some places that still use the Lua interpreter: Example tests
(run through the Lua interpreter directly) and unit tests (run through
busted). For unit tests, this should not make much of a difference and
example tests might later get similar treatment.


blueyed:
What about having this in `awesome` itself directly, e.g. via `--check`?

psychon:
`--check` is already taken for the (rather useless) "check if the config file is parsable", but I like your idea.

How about adding it to `--version`? We already try to load `lgi.version` there and we could just silently try to do the rest of the checks there as well. Running `awesome --version` during the build should be ok, too (but may hurt parallel builds a bit since the awesome binary then has to be built first so that `awesome --version` can be run...). Oh and `awesome --version` would need to have a non-zero exit status on error (fine by me). Also, we can just drop the "continue besides errors"-thing, because since a while we require lgi for the integration tests during the build as well (when this code was added, we did not actually use lgi during the build).

Elv13:
> --check is already taken for the (rather useless) "check if the config file is parsable", but I like your idea.

`--version` already checks multiple component versions. `--check` could do the same and check multiple components too. It would make it a lot more useful.

 * Is `rc.lua` parsable
 * Is LGI working
 * Can `awful`/`wibox` be loaded
 * Does the fallback `rc.lua` exists
 * Are the default themes anywhere to be found

psychon:
But then I'll have to update the man pages :-(

Also, `--check` would not work during the build since nothing was installed yet, would it? So it would need to be called with `--config ${BUILD_DIR}/rc.lua` and only the first two items from your list are doable.

Elv13:
> and only the first two items from your list are doable.

I don't think so. Why, yes, they are a bit pointless when awesome isn't installed, they are very doable. Checking if `awful`/`wibox` exists when you know the value of `--search`, `package.path` and `LUA_PATH` is trivial. Checking if the fallback `rc.lua` exists and contains "valid lua" (as in the valid syntax) is also easy. As for the theme, that's also easy. If the point of `--check` is to check if, assuming the code doesn't do dumb things, can run, then it is normal to expect `--check` to say it wont if necessary components are not working and/or not found.

psychon:
> Checking if the fallback rc.lua exists and contains "valid lua" (as in the valid syntax) is also easy.

Yes, but my point was meant to be: When the current `lgi-check.sh` runs, awesome was not yet installed. So there is no fallback rc.lua yet. So the new test will always fail.

psychon:
So, besides "there can be further improvements, e.g. this could eventually be merged into awesome itself", is there anything holding this back? There was no activity here for two months and as far as I see, no one said (yet) that as-is these patches are wrong/bad.

Elv13:
Thanks, sorry for the delay

PR #1844, opened 2017-06-14
Add a very basic test for the menubar

psychon:
This does not actually test anything. It just makes sure that the code
runs. A proper test still seems to need some time and some hacks, but
this basic test is enough to check that the menubar does not cause any
Lua errors.

### DIFF
--- a/tests/test-menubar.lua
+++ b/tests/test-menubar.lua
@@ -1,0 +1,36 @@
+-- Test the menubar
+
+local runner = require("_runner")
+local menubar = require("menubar")
+
+runner.run_steps {
+    function(count)
+        -- Just show the menubar and hide it.
+        -- TODO: Write a proper test. But for the mean time this is better than
+        -- nothing (and tells us when errors are thrown).
+
+        if count == 1 then
+            -- Remove the entry for $HOME/.... from the search path. We run with
+            -- HOME=/dev/null and this would cause an error message to be
+            -- printed which would be interpreted as the test failing
+            local entry = menubar.menu_gen.all_menu_dirs[1]
+            assert(string.find(entry, "/dev/null"), entry)
+            table.remove(menubar.menu_gen.all_menu_dirs, 1)
+
+            menubar.show()
+        end
+
+        -- Test that the async population of the menubar is done
+        if #menubar.menu_entries > 0 then
+            menubar.hide()
+            awesome.sync()
+            return true
+        end
+    end,
+
+    function()
+        return true
+    end
+}
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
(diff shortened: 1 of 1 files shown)


psychon:
When reverting https://github.com/awesomeWM/awesome/commit/bcc3de469e6ae09489e1d6fb8bad4cdc0e1192cd, the build looks like this: https://travis-ci.org/awesomeWM/awesome/builds/242724312?utm_source=github_status&utm_medium=notification

I will now force-push that revert away.

psychon:
> awesome: lapi.c:915: lua_callk: Assertion `(L->status == 0) && "cannot do calls on non-normal thread"' failed.

I think the Travis build failed because this still runs into https://github.com/pavouk/lgi/issues/155.

Should we all support @hoelzro with https://github.com/pavouk/lgi/issues/155#issuecomment-307791703 and open a "we need a new release" issue at lgi where we all can ask for one? That would mean bumping our minimum lgi requirement (or at least make `menubar` require that `lgi` version and print a warning if only something older is available).

psychon:
Another option could be to disable the garbage collector in the menubar tests. No idea if that would help, but that would also make the test so much worse... "Look, it works if we are extra-careful to avoid its brokeness!".

Elv13:
> That would mean bumping our minimum lgi requirement

To play the devil advocate, that also means losing any hopes of ever shipping 4.2+ in existing distributions. Maybe a more moderate approach (read: more work on our side...) to detect if the system LGI has been patched would be reasonable. Getting commits into Debian stable (and Ubuntu) isn't that hard when it solves a crash. But for the tests we can just bump LGI. It isn't a perfect solution. But bumping the minimum LGI version to one that doesn't exist isn't helping either.

psychon:
Adding `collectgarbage("stop")` does not help (and actually causes one more build to fail): https://github.com/awesomeWM/awesome/pull/1844/commits/f01d6afad9fd16f54cba80f710acdb346ab130f4 https://travis-ci.org/awesomeWM/awesome/builds/245698005?utm_source=github_status&utm_medium=notification

psychon:
Added cf4bd7e5ad5e5446b218b1e29d13ae98bb9b056c to make the Travis build work. Should this be merged now? I think the Travis error can be ignored and will be "fixed" by #1863. If needed I can wait for that commit and rebase.

Elv13:
Too bad this doesn't show in the coverage. So all that's needed is a new LGI release? I rebased the commits and fixed a typo in the message (`text`->`test`). Merge if you agree.

psychon:
> Too bad this doesn't show in the coverage. 

Oh right, this test is disabled when measuring coverage since it crashes. :-(

> So all that's needed is a new LGI release?

As far as I know: Yes. The only crash that I know was fixed in January by https://github.com/pavouk/lgi/pull/156.

Issue #1819, opened 2017-06-06
Table index is nil


Kkaminski93:
/usr/share/awesome/lib/menubar/icon_theme.lua:95: Table index is nil

![error](https://user-images.githubusercontent.com/28407207/26844689-e0955e34-4ab9-11e7-9629-59b09767b91d.png)

psychon:
None of the following files exists on your system (I guess this means that `get_default_icon_theme_name()` should just return some non-working string instead of `nil` if it cannot find anything):
```
~/.icons/Adwaita/index.theme
~/.icons/gnome/index.theme
~/.icons/hicolor/index.theme
~/.local/share/icons/Adwaita/index.theme
~/.local/share/icons/gnome/index.theme
~/.local/share/icons/hicolor/index.theme
/usr/share/icons/Adwaita/index.theme
/usr/share/icons/gnome/index.theme
/usr/share/icons/hicolor/index.theme
/usr/local/share/icons/Adwaita/index.theme
/usr/local/share/icons/gnome/index.theme
/usr/local/share/icons/hicolor/index.theme
/usr/share/pixmaps/Adwaita/index.theme
/usr/share/pixmaps/gnome/index.theme
/usr/share/pixmaps/hicolor/index.theme
/usr/local/share/pixmaps/Adwaita/index.theme
/usr/local/share/pixmaps/gnome/index.theme
/usr/local/share/pixmaps/hicolor/index.theme
```
As a work-around, set  `icon_theme` to `"foobar"` in your theme (e.g. via (bad hack) `beautiful.get().icon_theme = "foobar"` in your config).

Alternatively, can you test the following patch?
```diff
diff --git a/lib/menubar/icon_theme.lua b/lib/menubar/icon_theme.lua
index 1bc32bd..b9becb4 100644
--- a/lib/menubar/icon_theme.lua
+++ b/lib/menubar/icon_theme.lua
@@ -70,7 +70,7 @@ local get_default_icon_theme_name = function()
             end
         end
     end
-    return nil
+    return "no_theme_found"
 end
 
 local icon_theme = { mt = {} }
```

PR #1846, opened 2017-06-14
menubar.icon_theme: Never use nil as icon theme name

psychon:
The return value for this function is used as an index in a table and
Lua does not like nil as an index.

The function that actually looks for icons, find_icon_theme() already
falls back to "hicolor" if it does not find anything via the current
theme, so fix this issue by just falling back to "hicolor" here as well.

Fixes: https://github.com/awesomeWM/awesome/issues/1819

### DIFF
--- a/lib/menubar/icon_theme.lua
+++ b/lib/menubar/icon_theme.lua
@@ -70,7 +70,7 @@ local get_default_icon_theme_name = function()
             end
         end
     end
-    return nil
+    return "hicolor"
 end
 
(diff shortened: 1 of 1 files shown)


Elv13:
Horray! We are (but not for long) below 150 open issues again! 130 issues resolved since v4.1, good job everyone!

psychon:
I feel like that was mostly you with your recent efforts to bring down the number of open issues, so: Good job @Elv13 :-)

PR #1847, opened 2017-06-14
Fix documentation of activation filter

SammysHP:
Previously the `force` hint was described incorrectly. **Please verify that I have understood the code correctly.** I am not sure if the force hint is used correctly in the code because I would expect that the client gets the focus if `ret == false and hints.force`.


psychon:
I think we can ignore the Travis failure with Lua 5.3, lua.org disappeared:

> Connecting to www.lua.org (www.lua.org)|148.251.24.173|:80... failed: Connection refused.

Elv13:
Thanks

PR #1850, opened 2017-06-15
Add missing awful.widget.common require

timroes:
You currently have to require that module directly. I don't know if this was on purpose, since it's kind of helper methods for other utils, but since it has a public documentation, I would consider it API and would also make it available via `require('awful').widget.common`.


blueyed:
Thanks!

Issue #1849, opened 2017-06-14
Text not shown in notification with screen dpi and specific text


timroes:
Output of `awesome --version`:

```
awesome v4.1 (Technologic)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1
```

**How to reproduce the issue:**

Set a screen dpi per `beautiful.xreserouces.set_dpi(192, <screen>)` and after that show a notification that only contains block drawing characters, e.g. using `notify-send`:

```
$ notify-send Title '█░'
```

**Actual result:**

The text of the notification is missing.

**Expected result:**

The text of the notification should be shown.

**More details:**

This bug only appears if you have set the dpis. If not the text is shown perfectly fine.

As soon as you put another character within the text (e.g. `notify-send Title '█░x'`) the text (including all 3 characters) is shown perfectly fine. So it's only broken if you only use block chars (e.g. to draw a progress bar).

Also this problem seem to affect only the text of a notification not the title. Everything works as expected there.

psychon:
....what?!?

psychon:
> Also this problem seem to affect only the text of a notification not the title. Everything works as expected there.

Both the text and the title are given to a textbox like this: `return textbox:set_markup_silently(string.format('<b>%s</b>%s', title, text:gsub(pattern, replacements)))`. So all of this makes no sense at all.

Could you do some experiments with `/usr/bin/pango-view` and your block drawing character? Something like this should get you started: `pango-view --markup --dpi=192 --font="Whatever you use in awesome" --backend=cairo --text="foo<b>bar</b>"`

Elv13:
I can reproduce without having the DPI set

psychon:
![notification](https://user-images.githubusercontent.com/89482/27169502-7193f28c-51aa-11e7-826b-404c5e1e1ea5.png)


Works here (but I have a fixed notification width which might make a difference here? Also, now I wonder where the line break in notifications comes from...)

Elv13:
The notification with only the blocks has 1px less (in the height) than the one with `█░x`. Looks like a rounding issue and no text is shown when the height is too little.

Elv13:
![pick13](https://user-images.githubusercontent.com/340384/27169636-a615a4ba-5178-11e7-87d7-0fcd982e27bb.png)

[...]

PR #1854, opened 2017-06-20
notification: Fix `font` and use `beautiful` directly

Elv13:
Using `beautiful.get()` has the drawback of not supporting theme
variables set from `rc.lua`. It is also used less often than
direct theme access, making it a bit inconsistent with how other
modules behave.

Found while investigating #1849

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -22,11 +22,11 @@ local screen = require("awful.screen")
 local util = require("awful.util")
 local gtable = require("gears.table")
 local gfs = require("gears.filesystem")
-local bt = require("beautiful")
+local beautiful = require("beautiful")
 local wibox = require("wibox")
 local surface = require("gears.surface")
 local cairo = require("lgi").cairo
-local dpi = bt.xresources.apply_dpi
+local dpi = beautiful.xresources.apply_dpi
 
 local function get_screen(s)
     return s and capi.screen[s]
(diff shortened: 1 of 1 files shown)


lib/naughty/core.lua
```diff
@@ -591,8 +591,7 @@ function naughty.notify(args)
     local destroy_cb = args.destroy
 
     -- beautiful
-    local beautiful = bt.get()
-    local font = args.font or preset.font or
+    local font = args.font or preset.font or beautiful.notification_font or
```
psychon: Since the commit message doesn't mention it and I do not see an added `@beautiful`: Why does this add a new theme variable?

Elv13: https://github.com/Elv13/awesome-1/blob/0829cc2d76b5af2702511a989275751e909f6437/lib/naughty/core.lua#L147

psychon: Sigh... our docs :-)

Elv13:
> Well... ok. Will you submit a PR for that?

Eventually, for now I will open an issue. I would prefer to merge this now and fix everything later.

     git grep 'beautiful.get' -- .. | cut -f1 -d':' | grep -v tests | sort | uniq | wc -l

`10`

    git grep 'beautiful' -- .. | cut -f1 -d':' | grep -v tests | sort | uniq | wc -l

`67` (`-10 ==> 57`)

So directly using `beautiful` is 6 time more common than using `.get()`. This means that setting values directly already worked for 85% of the modules using theme variables. It is also simpler. So I thinking fixing the metatable is more useful than a warning.

PR #1855, opened 2017-06-20
tag: Warn when the number of `awful.tag.new` layout is too small.

Elv13:
Just a note on how not to fix this:

```
(Julien Danjou      2009-08-21) --- Create a set of tags and attach it to a screen.
(Emmanuel L-V       2016-04-11) -- @function awful.tag.new
(Julien Danjou      2009-08-21) -- @param names The tag name, in a table
(Julien Danjou      2009-08-21) -- @param screen The tag screen, or 1 if not set.
(Zsolt Udvari       2009-10-26) -- @param layout The layout or layout table to set for this tags by default.
(Julien Danjou      2009-08-21) -- @return A table with all created tags.
(Arvydas Sidorenko  2012-06-12) function tag.new(names, screen, layout)
(Uli Schlachter     2016-02-26)     screen = get_screen(screen or 1)
(Julien Danjou      2009-08-21)     local tags = {}
(Julien Danjou      2009-08-21)     for id, name in ipairs(names) do
(Arvydas Sidorenko  2012-06-12)         table.insert(tags, id, tag.add(name, {screen = screen,
(Perry Hargrave     2010-05-28)                                             layout = (layout and layout[id]) or
(Perry Hargrave     2010-05-28)                                                         layout}))
(Julien Danjou      2009-08-21)         -- Select the first tag.
(Julien Danjou      2009-08-21)         if id == 1 then
(Perry Hargrave     2010-05-28)             tags[id].selected = true
(Julien Danjou      2009-08-21)         end
(Julien Danjou      2009-08-21)     end
(Perry Hargrave     2010-05-28) 
(Julien Danjou      2009-08-21)     return tags
(Julien Danjou      2009-08-21) end
```

Note that the code has always been this way. I admit I was also a little shocked the first time I saw it and never used `awful.tag.new` (Tyrannical uses `awful.tag.add`, a saner alternative). However the current behavior needs to "keep working", of course it doesn't mean the bug cannot be fixed, but the code might be a little ugly. The best solution is simply

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -240,11 +240,20 @@ end
 -- @return A table with all created tags.
 function tag.new(names, screen, layout)
     screen = get_screen(screen or 1)
-    local tags = {}
+
+    if layout and layout[1] and #names > #layout then
+        gdebug.print_warning("The number of tags is greater than the number of layouts")
+    end
+
+    local tags, overflow = {}, false
     for id, name in ipairs(names) do
-        table.insert(tags, id, tag.add(name, {screen = screen,
-                                            layout = (layout and layout[id]) or
-                                                        layout}))
+        table.insert(tags, id, tag.add(name, {
+            screen = screen,
+            layout = (layout and layout[id]) or (overflow and layout[1] or layout)
+        }))
+
+        overflow = overflow or (layout and layout[id])
+
         -- Select the first tag.
         if id == 1 then
             tags[id].selected = true


lib/awful/tag.lua
```diff
@@ -240,11 +240,20 @@ end
...
     for id, name in ipairs(names) do
-        table.insert(tags, id, tag.add(name, {screen = screen,
-                                            layout = (layout and layout[id]) or
-                                                        layout}))
+        table.insert(tags, id, tag.add(name, {
+            screen = screen,
+            layout = (layout and layout[id]) or (overflow and layout[1] or layout)
+        }))
+
+        overflow = overflow or (layout and layout[id])
```
psychon: This `overflow` logic confuses me. So `layout` should always be trueish (if someone passes in `nil` or `false` here, then... things would work. Oh boy.).

New attempt:
If `layout` is `nil`, then all created tags end up without a layout. Good.
If `layout` is a `layout` without `layout[1]`, then... `(layout and layout[id])` would evaluate to `nil` and `(overflow and layout[1] or layout)` would evaluate to `layout`. Good, but complicated.
If `layout` is a sufficiently long list of layouts, then the first case in the tag-creation always fires and `overflow` stays falsish.
If `layout` has too few layouts... let's say `names = { "foo", "bar", "baz" }` and `layouts = { }`, then the first created tag will end up with... an empty table as its layout.... right? `layout and layout[id]` is `nil`, `overflow` is `false` and so the second part evaluates to `layout`.

So this patch is (a) too complicated and (b) does not fix things, I think.

Elv13:
feel free to push your revised patch directly into master, it is easier to read.

lib/awful/tag.lua
```diff
@@ -240,11 +240,20 @@ end
 -- @return A table with all created tags.
 function tag.new(names, screen, layout)
     screen = get_screen(screen or 1)
-    local tags = {}
+
+    if layout and layout[1] and #names > #layout then
```
actionless: won't it throw an exception if `layout` is a single layout, not array?

PR #1859, opened 2017-06-20
Make beautiful.foo = bar work properly

psychon:
Without this, users would modify the beautiful table directly instead of
the theme. This made a difference for code using beautiful.get() to get
the theme.

Reference: https://github.com/awesomeWM/awesome/pull/1854


Elv13:
Thanks

PR #1860, opened 2017-06-22
Automatically make LuaRocks modules available in awesome

psychon:
The "preferred way" to use LuaRocks is to run `eval $(luarocks path --bin)`, but for a window manager this is a bit hard to do before starting it. Thus, this PR is intended to make this easier to do.

This is WIP since I did not yet actually test if this work, but I just remembered that this is actually easy to do:
In `awesomerc.lua`:
```diff
print(pcall(require, "term"))
print(pcall(require("gears").require_luarocks.require, "term"))
```
Output:
```
false	module 'term' not found:
	no field package.preload['term']
	no file '/usr/share/lua/5.3/term.lua'
	no file '/usr/share/lua/5.3/term/init.lua'
	no file '/usr/lib64/lua/5.3/term.lua'
	no file '/usr/lib64/lua/5.3/term/init.lua'
	no file './term.lua'
	no file './term/init.lua'
	no file 'lib/term.lua'
	no file 'lib/term/init.lua'
	no file '/home/psychon/awesome/tests/term.lua'
	no file '/home/psychon/awesome/tests/term/init.lua'
	no file '/home/psychon/awesome/build/awesome/term.lua'
	no file '/home/psychon/awesome/build/awesome/term/init.lua'
	no file '/etc/xdg/awesome/term.lua'
	no file '/etc/xdg/awesome/term/init.lua'
	no file '/usr/local/share/awesome/lib/term.lua'
	no file '/usr/local/share/awesome/lib/term/init.lua'
	no file '/usr/lib64/lua/5.3/term.so'
	no file '/usr/lib64/lua/5.3/loadall.so'
	no file './term.so'
	no file 'lib/term.so'
	no file '/home/psychon/awesome/tests/term.so'
	no file '/home/psychon/awesome/build/awesome/term.so'
	no file '/etc/xdg/awesome/term.so'
	no file '/usr/local/share/awesome/lib/term.so'
Failed to require("term"), checking if this module is available through LuaRocks
true	table: 0xde01d0
```
Thus, this works just as intended.

----- 

Sigh. While writing the above I looked up `eval $(luarocks path --bin)` and found it here on this page: https://github.com/luarocks/luarocks/wiki/Using-LuaRocks#Multiple_versions_using_the_LuaRocks_package_loader
If you click that link, you are told that LuaRocks make available a `luarocks.loader` package that replaces `require` so that modules found through LuaRocks are found. I tested it and it seems to work just fine. Thus, this whole PR can be replaced with:
```diff
diff --git a/awesomerc.lua b/awesomerc.lua
index 00d1b60..5f31086 100644
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -1,3 +1,7 @@
+-- If LuaRocks is installed, make sure that packages installed through it are
+-- found (e.g. lgi). If LuaRocks is not installed, do nothing.
+pcall(require, "luarocks.loader")
+
 -- @DOC_REQUIRE_SECTION@
 -- Standard awesome library
 local gears = require("gears")
```
So... what now? Opinions on the preferred approach? Can we be sure that the new patch works everywhere where the old patch worked? I can only see reasons why the new patch is better: No `popen`, so less resource intensive. It is a Lua module, so you do not get errors due to version incompatibilities (awesome running with Lua 5.2 and LuaRocks with Lua 5.3 and we mix modules for the two).

-----

CC  @hishamhm Just for your information, plus I'll continue to CC you on all LuaRocks-related things.

### DIFF
--- a/build-utils/check_for_invalid_requires.lua
+++ b/build-utils/check_for_invalid_requires.lua
@@ -18,6 +18,9 @@ local allowed_deps = {
     gears = {
         lgi = true,
     },
+    ["gears.require_luarocks"] = {
+        ["*"] = true,
+    },
     beautiful = {
         gears = true,
         lgi = true,
--- a/build-utils/lgi-check.sh
+++ b/build-utils/lgi-check.sh
@@ -25,7 +25,7 @@ die()
 lua -e 'require("lgi")' || die
 
 # Check the version number.
-# Keep this in sync with lib/gears/surface.lua and .travis.yml (LGIVER)!
+# Keep this in sync with lib/gears/init.lua and .travis.yml (LGIVER)!
 lua -e '_, _, major_minor, patch = string.find(require("lgi.version"), "^(%d%.%d)%.(%d)");
 	if tonumber(major_minor) < 0.8 or (tonumber(major_minor) == 0.8 and tonumber(patch) < 0) then
 		error(string.format("lgi is too old, need at least version %s, got %s.",
--- a/lib/gears/init.lua
+++ b/lib/gears/init.lua
@@ -4,6 +4,8 @@
 -- @module gears
 ---------------------------------------------------------------------------
 
+local require_luarocks = require("gears.require_luarocks")
+require_luarocks.require("lgi")
 
 return
 {
@@ -22,6 +24,7 @@ return
     table = require("gears.table");
     string = require("gears.string");
     filesystem = require("gears.filesystem");
+    require_luarocks = require_luarocks;
 }
 
 -- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/lib/gears/init.lua
+++ b/lib/gears/init.lua
@@ -7,6 +7,12 @@
 local require_luarocks = require("gears.require_luarocks")
 require_luarocks.require("lgi")
 
+-- Keep this in sync with build-utils/lgi-check.sh and .travis.yml (LGIVER)!
+local ver_major, ver_minor, ver_patch = string.match(require_luarocks.require('lgi.version'), '(%d)%.(%d)%.(%d)')
+if tonumber(ver_major) <= 0 and (tonumber(ver_minor) < 8 or (tonumber(ver_minor) == 8 and tonumber(ver_patch) < 0)) then
+    error("lgi too old, need at least version 0.8.0")
+end
+
 return
 {
     color = require("gears.color");
--- a/lib/gears/require_luarocks.lua
+++ b/lib/gears/require_luarocks.lua
@@ -1,0 +1,63 @@
+---------------------------------------------------------------------------
+-- @author Uli Schlachter
+-- @copyright 2017 Uli Schlachter
+-- @module gears.require_luarocks
+---------------------------------------------------------------------------
+
+local require = require
+local io = io
+
+local require_luarocks = {}
+
+local luarocks_set_up = false
+
+--- Make sure that libraries installed through LuaRocks can be require()d.
+-- This function does the equivalent of running `luarocks path` before starting
+-- Lua. If LuaRocks is not installed, this might print an error on stderr.
+function require_luarocks.add_luarocks_path()
+    if luarocks_set_up then
+        return
+    end
+    luarocks_set_up = true
+
+    -- Run the given command and add its output to package[key]
+    local function read_and_add(command, key)
+        -- Using io.popen blocks awesome and is almost always a bad idea,
+        -- because lgi offers better alternatives. Here, we might not have lgi
+        -- yet and so we must resort to popen.
+        local out = io.popen(command)
+        if not out then
+            return
+        end
+        -- Read output and remove trailing newline
+        local result = out:read("*a"):sub(1, -2)
+        out:close()
+        if result ~= "" then
+            package[key] = package[key] .. ";" .. result
+        end
+    end
+    read_and_add("luarocks path --lr-path", "path")
+    read_and_add("luarocks path --lr-cpath", "cpath")
+end
+
+--- Try to load a module and if that fails, try to load it through LuaRocks.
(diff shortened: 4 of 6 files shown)


psychon:
Related: https://github.com/awesomeWM/awesome-www/pull/98#issuecomment-308070675 (and following discussion)

psychon:
Heh, that's a funny Travis failure:
```
[100%] Checking use of require()
Dependency from gears.require_luarocks onto * is not allowed
make[4]: *** [CMakeFiles/check-requires] Error 1
```
I'll handle that if needed.

timroes:
I think using the `luarocks.loader` makes a lot of sense, when using luarock packages. I could get rid of [several lines of code](https://github.com/timroes/awesome-config/blob/9ca93ddee7093329d93f81e76cb09cb8b2aa17f1/rc.lua#L21:L27) and just replace it with the one `require` statement. Thanks for @psychon pointing this out again to me.

But regarding adding it to the `awesomerc.lua`: Since you don't use any luarocks packages in that file (correct me if I'm wrong) this would only be an advantage for people actually wanting to customize the config and adding luarocks packages. Since that people anyway touch their config, couldn't they also add this `require` themselves? Maybe it would be a better option to just add some visible documentation, that this method exists?

psychon:
@timroes This idea started from a bug report with "I installed lgi through LuaRocks, but awesome does not find it". The idea is to make this case work better.

timroes:
Okay. If people actually report that issue, it would make sense to add it to the default config. Also it's not much harm to include it, even if people doesn't need it. Also since it's a one liner (excluding documentation) it doesn't clutter the config too much in my opinion.

psychon:
@blueyed Could you please un-approve this and instead "request changes"? This PR is actually bad since, as far as I understand, `pcall(require, "luarocks.loader")` is just as good, but requires a lot less code (sadly I only found that after creating these commits and no one from LuaRocks told me about it before)...

Looking at LuaRocks `make install` target, they require superuser privileges for installation. `require("luarocks.loader")` would not work for anyone installing to e.g. `$HOME/luarocks`. Do we care about that? Should it be enough to have `luarocks` in `$PATH` for awesome to find it? Or do we take the easier way of having `luarocks.loader` in lua's search path? I don't really know...

hishamhm:
@psychon @blueyed Sorry about not being responsive in these threads lately (I just moved to another state this week, I set up my new home-office today). 

Both approaches should work, but they assume different things from the system. The one with that execs `luarocks path --lr-path` assumes that the `luarocks` binary is in the $PATH; the one that does `pcall(require, "luarocks.loader")` assumes that `luarocks.loader` is in the $LUA_PATH.

With vanilla installs of the Lua and LuaRocks tarballs, both assumptions will be true (since both install into `/usr/local`).

With package-manager-based installs of Lua and LuaRocks (e.g. `apt-get` and friends), I _hope_ both assumptions will be true, if the packagers for the distro configured both packages to use the same paths for Lua modules. (If `require("luarocks.loader")` doesn't work after installing Lua and LuaRocks from the package manager, I'd dare say this is a downstream packaging bug.)

Both _can_ work with a user-installed LuaRocks under `$HOME/luarocks`, if the user also has their environment variables set up accordingly (which is a usual necessity for $HOME-installed software anyway?). In any case, I think users who install stuff in custom paths are more likely to troubleshoot module installation paths and the like.

The most delicate scenario is probably a package-manager-installed Lua (in `/usr`) and a user-built LuaRocks (in `/usr/local` or elsewhere). This is a somewhat common scenario because the LuaRocks packages in distros are often severely outdated. I'm afraid distros don't configure Lua to look for modules under `/usr/local`, so `luarocks.loader` won't be found out-of-the-box. However, I believe they do configure $PATH to include `/usr/local/bin`, so the first approach with `luarocks path --lr-path` will work.

So, in short, I like `pcall(require, "luarocks.loader")` better too, but it assumes a properly setup environment. The `luarocks path --lr-path` seems a bit more resilient.

(I hate this mess as much as you do, but it's due to having lots of moving parts managed and configured by different people. :\ If there's anything I can do on my side to alleviate this situation, short of maintaining downstream packages for various distros myself, I'm all ears.)

psychon:
@hishamhm Thanks for your answer and I hope the move went fine. Sadly: No, I do not see what LuaRocks can do differently. I might have some wishes for Lua upstream, but since Lua is meant to be embedded these ideas won't be accepted.

The last case from your list is especially interesting. Yup, my Fedora-Lua does not look into `/usr/local`, so this way shorter and better patch actually helps less people. Also, modifying `$PATH` correctly might be a little easier than messing with `$LUA_PATH` and `$LUA_CPATH`.

I'll un-WIP this PR and I guess it can be merged then.

blueyed:
Unit tests are failing in one job due to missing zsh?! (why only there?)

> /usr/bin/env: zsh: No such file or directory

blueyed:
Rebased on master, out of curiosity and to trigger a rebuild.

psychon:
Rebased on master again and fixed the error from `check_for_invalid_requires` by allowing `gears.require_luarocks` to require `*` (that's the only change I did).

blueyed:
I have tried this to make "pulseaudio_dbus" installed via `lua --local install pulseaudio_dbus` available, but it does not appear to work.

It works in `lua` itself after `eval $(luarocks path)`.

blueyed:
Oh, I was under the impressions that this would adjust the path automatically to allows packages from luarocks to be required by default..

psychon:
> Oh, I was under the impressions that this would adjust the path automatically to allows packages from luarocks to be required by default..

Problem with that is: If LuaRocks is not available, "luarocks: command not found" ends up visible on stderr. So anyone with any kind of problem will assume that awesome needs luarocks.

You'll have to add `require("gears.require_luarocks").add_luarocks_path()` to the beginning of your `rc.lua` to make it available unconditionally.

blueyed:
Superseded by https://github.com/awesomeWM/awesome/pull/1999.

PR #1864, opened 2017-06-24
awful.prompt: fix Shift-Tab

blueyed:
1. before this patch the "Shift" key(s) to get to Shift-Tab would reset
   ncomp already.
2. "Shift-Tab" will come in as "Tab" with mod.Shift typically.
3. cur_pos and ncomp have to get reset in case of going back to the
   original entry.


blueyed:
Coverage should be better after https://github.com/awesomeWM/awesome/pull/1736.

blueyed:
I've started writing some tests to cover this, but do not feel like finishing them currently.

blueyed:
There we go for v4.2.

Issue #1427, opened 2017-01-22
build: docs get generated twice with "make && make install"


blueyed:
I think for packages the `make` could be just skipped (when `cmake` was called already).

```
build() {
  mkdir -p build
  cd build

  cmake ../$pkgname \
    -DCMAKE_BUILD_TYPE=RELEASE \
    -DCMAKE_INSTALL_PREFIX=/usr \
    -DSYSCONFDIR=/etc \
    -DLUA_INCLUDE_DIR=/usr/include/luajit-2.0 \
    -DLUA_LIBRARY=/usr/lib/libluajit-5.1.so
  # XXX: remove this?!
  # make
}

package() {
  cd build
  make DESTDIR="$pkgdir" install

  install -Dm644 "$srcdir"/$pkgname/awesome.desktop \
    "$pkgdir/usr/share/xsessions/awesome.desktop"

  install -Dm644 "$srcdir"/awesomeksm.desktop \
    "$pkgdir/usr/share/apps/ksmserver/windowmanagers/awesome.desktop"
}
```

Our README mentions `make && make install` though..

blueyed:
We might improve the build speed after by just calling `make install` directly / fixing this issue:
```
  - travis_run_in_fold "make" make
  - |
    if [ "$TRAVIS_TEST_RESULT" = 0 ]; then
      travis_run_in_fold "make.install" sudo env PATH=$PATH make install
      awesome --version
    fi
```

psychon:
I'm not sure that just running `install` directly always works fine. It does not run the `all` target, does it? I would feel more comfortable if we keep this separate.

However, we could have `cd build` or something like that in the right places on Travis. That should avoid rerunning cmake, I think.

blueyed:
See https://github.com/awesomeWM/awesome/pull/1867 for some fix regard.

Closing this issue as unclear / not valid.

PR #1867, opened 2017-06-24
CMakeLists.txt: remove check-unit from ALL

blueyed:
This makes "make install" not run the unit tests (again).

Instead "check-unit" is added to the top-level wrapper Makefile to be
run for its "all" task.

Ref: https://github.com/awesomeWM/awesome/issues/1427

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ endif
 TARGETS=awesome
 BUILDDIR=build
 
-all: $(TARGETS) ;
+all: check-unit $(TARGETS) ;
 
 $(TARGETS): cmake-build
 	ln -s -f $(BUILDDIR)/$@ $@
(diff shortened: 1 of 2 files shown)


blueyed:
Why did @codecov[bot] report three times here?  /cc @codecov

Makefile
```diff
@@ -8,7 +8,7 @@ endif
 TARGETS=awesome
 BUILDDIR=build
 
-all: $(TARGETS) ;
+all: check-unit $(TARGETS) ;
```
psychon: I don't like this part, sorry. The top-level Makefile is intended for convenience. Using CMake directly should not have different results.

However, I'll let this be.

blueyed: Do you see a way around this?
Should `make install` from the wrapper Makefile get specialized then maybe?

psychon: I already :+1:'d this, so let's just merge this. ;-)

I do not see another way to do this.

PR #1868, opened 2017-06-24
Makefile: add busted tests as target

blueyed:
TODO: I is bad to trigger the "wildcard" expansion on every Makefile use
probably.  Suggestions?

TODO:
- [ ] I've expected this to work with Zsh's completion,, but it does not. Need to investigate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,12 @@ distclean:
 	$(RM) -r $(BUILDDIR) $(TARGETS)
 	$(ECHO) " done"
 
+BUSTED_TESTS:=$(wildcard spec/**/*.lua)
+$(BUSTED_TESTS):
+	busted --lpath="lib/?.lua;lib/?/init.lua;spec/?.lua" \
+	  --helper spec/preload.lua $@
+.PHONY: $(BUSTED_TESTS)
+
 %: $(BUILDDIR)/CMakeCache.txt
 	$(ECHO) "Running make $@…"
 	$(MAKE) -C $(BUILDDIR) $@


Makefile
```diff
@@ -36,6 +36,12 @@ distclean:
 	$(RM) -r $(BUILDDIR) $(TARGETS)
 	$(ECHO) " done"
 
+BUSTED_TESTS:=$(wildcard spec/**/*.lua)
```
psychon: I don't see this recursive use of wildcards in https://www.gnu.org/software/make/manual/html_node/Wildcards.html
There is no `**`, I fear.

psychon:
Why do you want to just run a single test file? Are the unit tests slow for you?

> I is bad to trigger the "wildcard" expansion on every Makefile use probably.

Well, this is how Makefiles work...

Makefile
blueyed: Works for me.
Just try `$(error $(wildcard spec/**/*.lua))` at the top of the Makefile.

blueyed:
As for not expanding it always, I've tried to use `%` for the pattern here, too, but it did not work.

> Why do you want to just run a single test file? Are the unit tests slow for you?

They are not that slow, but there is still a difference, of course - at that gets bigger the more tests are being added.

Makefile
psychon: Doesn't work here:
```
Makefile:11: *** spec/menubar/icon_theme_spec.lua spec/awful/screen_spec.lua spec/awful/keyboardlayout_spec.lua spec/gears/geometry_spec.lua spec/gears/protected_call_spec.lua spec/gears/string_spec.lua spec/gears/table_spec.lua spec/gears/object_spec.lua spec/gears/cache_spec.lua spec/gears/color_spec.lua spec/gears/matrix_spec.lua spec/wibox/test_utils.lua spec/wibox/hierarchy_spec.lua.  Schluss.
```
It just does `$(wildcard spec/*/*.lua)`, the second `*` doesn't change anything. For example, `spec/preload.lua` is missing. This is not recursive.

psychon:
> They are not that slow, but there is still a difference, of course - at that gets bigger the more tests are being added.

`time` says it needs 0.173 seconds here, so "not much" and that is for 248 tests. So for one second of run time, we'd need 1.5 million tests. :-)

blueyed:
Closing this for now.

blueyed:
It is not only about time though, but only focusing on some specific tests instead, without being interrupted by other failures / output.
In those cases you can easily use the command from this Makefile literally though.

PR #1871, opened 2017-06-26
Add naughty.destroy_all_notifications functions.

morethanoneanimal:
Update connect_signal in naughty lib to use destroy_all_notifications().

I have this function in my config and it comes in handy. I think it's a good candidate to an API.

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -217,11 +217,7 @@ end)
 
 capi.screen.connect_signal("removed", function(scr)
     -- Destroy all notifications on this screen
-    for _, list in pairs(naughty.notifications[scr]) do
-        while #list > 0 do
-            naughty.destroy(list[1])
-        end
-    end
+    naughty.destroy_all_notifications({scr})
     naughty.notifications[scr] = nil
 end)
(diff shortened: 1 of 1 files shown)


Elv13:
Hello, thanks for this, this is a little issue to fix:

     lib/naughty/core.lua:375:12: shadowing upvalue screen on line 21

lib/naughty/core.lua
```diff
@@ -364,6 +360,27 @@ function naughty.destroy(notification, reason, keep_visible)
     end
 end
 
+--- Destroy all notifications on given screens.
+--
+-- @param screens Screens on which notifications should be destroyed. If nil,
+-- destroy notifications on all screens.
+-- @param reason One of the reasons from notificationClosedReason.
```
blueyed: Please use `@tparam`.

morethanoneanimal: Done. Also, I added return value to be consistent with naughty.destroy function.

morethanoneanimal:
Fixed var shadowing issue.

blueyed:
Thanks!

PR #1873, opened 2017-06-26
Update the changelog for Awesome 4.2

Elv13:


docs/89-NEWS.md
```diff
@@ -1,10 +1,78 @@
...
+ * Another instance where clients ended up in the wrong screen has been fixed
+ * Awesome will no longer generate zombie processes when restarted
+ * All official themes now support HiDPI screens
+ * The `magnifier` layout has been fixed
+ * The menubar has been fixed for Lua 5.1 users
+
+## New features
+
+ * The hotkey popup has been extended to support Firefox and TMUX
+ * Naughty, the notification system, as a new `ignore_suspend` flag
```
psychon: has

docs/89-NEWS.md
```diff
@@ -1,10 +1,78 @@
...
+ * The `textclock` widget now supports timezone
```
psychon: Should this be "timezones"?

docs/89-NEWS.md
```diff
@@ -1,10 +1,78 @@
...
+ * New utility functions have been added:
+   * `gears.string.split`
+   * `gears.table.map`
+   * `gears.filesystem.make_parent_directories`
+ * The new widgets functions have been added:
```
psychon: These / Some?

psychon: Also `widget` instead of `widgets`? This is somewhat weird.

psychon: Also.. this is not really new, the functions were just moved out of gears.

docs/89-NEWS.md
```diff
@@ -1,10 +1,78 @@
...
+The new calendar widgets are very flexible and can be themed down to the very small details:
+
+[![The calendar widget](../images/AUTOGEN_wibox_widget_calendar_fn_embed_cell.svg)](../classes/wibox.widget.calendar.html)
+[![The calendar widget](../images/AUTOGEN_wibox_widget_calendar_font.svg)](../classes/wibox.widget.calendar.html)
+
+
+## Behavior changes
+
+ * The client `property::floating` is now also emitted when the client is
+   maximized or fullscreen.
```
psychon: Should this say something like "when the floating state changes implicitly, e.g. because the client gets fullscreened"?

docs/89-NEWS.md
blueyed: "New widget functions (moved out of the `gears` module)".

docs/89-NEWS.md
blueyed:
```diff
- * The client `property::floating` is now also emitted when the client is
-   maximized or fullscreen.
+ * The client `property::floating` is now also emitted when when the floating
+   state changes implicitly, e.g. because the client gets maximized or
+   fullscreened.
```

docs/89-NEWS.md
```diff
@@ -1,10 +1,78 @@
...
+ * Building Awesome from its root source directory is no longer supported and
+   will print an error.
```
blueyed: That is not true, is it?
Just remove this?

blueyed:
Pushed some fixups, mainly based on @psychon's comments.

blueyed:
I think we should have 4.2 as soon as possible, if not only for the maximize issue.

I think there might be some kind of regression, but still that could be fixed in 4.2.1.

docs/89-NEWS.md
psychon: I also tripped over this. I think this refers to commit 27eba6c034fa4844cdc5272620ee3128862dec08 and is largely irrelevant to users. So 👍 for just removing this, but for a slightly different reason. :-)

psychon:
> Pushed some fixups, mainly based on @psychon's comments.

Thanks and sorry that I was lazy and just left comments.

> I think we should have 4.2 as soon as possible, if not only for the maximize issue.

Yes, but sigh... I'll try to find the time to think a bit harder about the calendar widget.

Elv13:
> Yes, but sigh... I'll try to find the time to think a bit harder about the calendar widget.

**Solution one (you wont like):** Fix LGI

**Solution two (you wont like):** Allow mutable contexts and use a single widget instance for all days "containers". This is what Qt does (`QStyledItemDelegate` + Qt Model/View). You already said "no" something like 5 times, but it still seems like a logical solution to me. It makes the calendar and `awful.widgets.common` much, much faster/simpler.

**Solution three (I wont like):** Hardcode everything and avoid using textbox+container+layout. This will make the widget code more complex and an order magnitude harder to make as customization as the current one is.

**Solution four (you wont laike):** Just live with it as-is.

psychon:
> Solution one (you wont like): Fix LGI

Fine by me. I even have write access to the repository. The only problem is that LGI is too much black magic to me and I do not know what would break.

> Solution two (you wont like): 

Well, I do not see how. If the `context` is modifable, we can no longer cache `:layout` and `:fit` results and we are back to the awesome 3.5 days of exponentially expensive `:fit` methods. You would not like that.

Solve this problem and I have no complaints. (But I do not yet see how this would help here...)

> > I'll try to find the time to think a bit harder about the calendar widget.

I was actually referring to its API, not to the "its slow part". I can live with the "its slow part" and will just ignore it. On the API: For example, the `fn_embed` callback (besides the bad name) is kind of a sledge hammer. It is very powerful, yes, but I feel like it is also hard to use.
For example, why does it get an already-configured textbox widget as its argument? Just have the default implementation be public and remove this argument.
Instead of the `flag` a-argument, I think I would prefer multiple callbacks, one per what is currently the flag. That also makes it more obvious which flag values / callbacks exist.

That's my current list and that is only from a quick look. (Not API-related, but: I'd also like to get rid of all the `make_widget_declarative`-calls).

Elv13:
> On the API: For example, the fn_embed callback (besides the bad name) is kind of a sledge hammer. 

I agree and noted the same thing in the PR. However there is no quick fix. My delegate+datamodel  concept would allow a simple delegate with access to pre-defined `id` with pre-defined API and would solve the problem. However, as you stated, making the mutable context work is not trivial to keep the caches. However I do think it can be solved and will in one blow solve the complexity of `awful.widget.common` (a metric ton of config option), `awful.menu` (performance scaling issue), the menubar (async issues), #1874 (inflexible hardcoded layout), my `Radical` module (lolifying complexity) and this (ridiculous customization API and constant performance issues).

The `fn_embed` API could be deprecated in v5 once every peices for the widget-as-delegate API is implemented. As @blueyed mentioned, we are overdue for a release.

psychon:
I'll try to do a release next weekend (likely Sunday). Is this okay? How do I know if the changelog is up-to-date then?

psychon:
Ping? Any reason this cannot be merged tomorrow? I looked through `git log --since="20 days ago"` and the following things caught my eye (this does NOT mean that all of these should be mentioned; anyone with enough editing-rights should feel free to tick items which should not be mentioned):

- [ ]  0430f55257d6fb12c89a45719017b963e8018bb1 ratio: Add a fill strategy for empty entries
- [ ]  fb367c3f9e043021023f508a408aac47aaf56d74  stack: Add offset support
- [ ]  547c0e4b834edd6d7ed9fa4985413c5702eb9229 awful.spawn: add easy_async_with_shell and fixed easy_async doc (#1541
- [ ]  3d1b2c2a5eb3585ee1377b7933a87299101d0959 Pass freedesktop hints to naughty.notify (#1875)
- [ ]  956ac3c50db7c139086c7f4dd62ec6a1727d2008 2a69ffa374c1429e10f91aa80ceda06d6accbec4 4c792b49a977b8b4ae1830996e691e72235ce0ac completion-related changes
- [ ]  5dcd5969be5f4b6c5f8f8c7c8bc947ea2264a893  awful.prompt: fix Shift-Tab
- [ ]  b9d9587afa58784497c35889eb0510ac9ee7383c menubar: Produce better errors for broken .desktop files (but we already had something similar for non-utf8-directory-names, one changelog entry for both likely is enough)
- [ ]  cbb8fd7ac408d229233a3b5490adf35d58a85d79 Add naughty.destroy_all_notifications (#1871)
- [ ]  d4924988dffdf6e5dcba64318837219e2fd356a3  rules: switchtotag: fix usage of awful.tag.viewmore (#1895)
- [ ]  ab559a6f36a8683167be00020373e595ab748a33  Provide better error message for unknown screen names
- [ ]  5fdb2c629847a3c393fb211ce17d8b4eac02d35f feat(beatiful: theme_assets; themes: xresources): recolor titlebar icons for `hover` and `press` states
- [ ]  4eddfacf5121a8d51a329aea7dd500e022809c9e lib/awful/completion.lua: handle missing SHELL environ var (#1897)
- [ ] There is still a comment on this PR about some item that should be removed

Elv13:
I approve the release of Awesome 4.2

psychon:
I removed the embedded images since the links obviously did not work: https://github.com/awesomeWM/awesome/releases/tag/v4.2

PR #1876, opened 2017-06-26
completion: Show command prior to directory

FloGa:
When using tab completion in the "Run:" prompt and there is a directory
with the same name as a command in the $HOME directory (or wherever
awesome was started from), then the current completion.lua would append
a slash to it and add it to the output list.

With this patch it is first tested whether there is an actual command
for the current completion item, and add it (without slash) to the
output list. Afterwards the directory test will still be executed, so
there might be an un-slashed item as well as a slashed item in the list.
This makes it possible to get completion for `foo` (which might reside
in /usr/local/bin) as well as `foo/bar` (with directory foo residing in
$HOME). The un-slashed item will be listed prior to the slashed item
since it is more common to execute commands from $PATH than from
arbitrary folders in the current working directory.

### DIFF
--- a/lib/awful/completion.lua
+++ b/lib/awful/completion.lua
@@ -9,6 +9,8 @@
 -- @module awful.completion
 ---------------------------------------------------------------------------
 
+local gfs = require("gears.filesystem")
+
 -- Grab environment we need
 local io = io
 local os = os
--- a/lib/awful/completion.lua
+++ b/lib/awful/completion.lua
@@ -160,7 +160,7 @@ function completion.shell(command, cur_pos, ncomp, shell)
         while true do
             local line = c:read("*line")
             if not line then break end
-            if gfs.is_dir(line) then
+            if str_starts(line, "./") and gfs.is_dir(line) then
                 line = line .. "/"
             end
             table.insert(output, bash_escape(line))
--- a/lib/awful/completion.lua
+++ b/lib/awful/completion.lua
@@ -79,6 +79,10 @@ function completion.shell(command, cur_pos, ncomp, shell)
     local i = 1
     local comptype = "file"
 
+    local function str_starts(str, start)
+        return string.sub(str, 1, string.len(start)) == start
+    end
+
     -- do nothing if we are on a letter, i.e. not at len + 1 or on a space
     if cur_pos ~= #command + 1 and command:sub(cur_pos, cur_pos) ~= " " then
         return command, cur_pos
--- a/lib/awful/completion.lua
+++ b/lib/awful/completion.lua
@@ -160,7 +160,7 @@ function completion.shell(command, cur_pos, ncomp, shell)
         while true do
             local line = c:read("*line")
             if not line then break end
-            if os.execute("test -d " .. string.format('%q', line)) == 0 then
+            if gfs.is_dir(line) then
                 line = line .. "/"
             end
             table.insert(output, bash_escape(line))
--- a/spec/awful/completion_spec.lua
+++ b/spec/awful/completion_spec.lua
@@ -95,7 +95,18 @@ describe("awful.completion.shell", function()
     setup(function()
         test_dir = get_test_dir()
         gfs.make_directories(test_dir .. '/true')
+        Gio.File.new_for_path(test_dir .. '/true/with_file'):create(Gio.FileCreateFlags.NONE);
         gfs.make_directories(test_dir .. '/just_a_directory')
+        Gio.File.new_for_path(test_dir .. '/just_a_directory/with_file'):create(Gio.FileCreateFlags.NONE);
+        -- Chaotic order is intended!
+        gfs.make_directories(test_dir .. '/ambiguous_dir_a')
+        gfs.make_directories(test_dir .. '/ambiguous_dir_e')
+        Gio.File.new_for_path(test_dir .. '/ambiguous_dir_e/with_file'):create(Gio.FileCreateFlags.NONE);
+        gfs.make_directories(test_dir .. '/ambiguous_dir_c')
+        Gio.File.new_for_path(test_dir .. '/ambiguous_dir_c/with_file'):create(Gio.FileCreateFlags.NONE);
+        gfs.make_directories(test_dir .. '/ambiguous_dir_d')
+        gfs.make_directories(test_dir .. '/ambiguous_dir_b')
+        gfs.make_directories(test_dir .. '/ambiguous_dir_f')
         os.execute(string.format(
             'cd %s && touch localcommand && chmod +x localcommand', test_dir))
         lfs.chdir(test_dir)
@@ -103,8 +114,18 @@ describe("awful.completion.shell", function()
     end)
 
     teardown(function()
+        assert.True(os.remove(test_dir .. '/ambiguous_dir_a'))
+        assert.True(os.remove(test_dir .. '/ambiguous_dir_b'))
+        assert.True(os.remove(test_dir .. '/ambiguous_dir_c/with_file'))
+        assert.True(os.remove(test_dir .. '/ambiguous_dir_c'))
+        assert.True(os.remove(test_dir .. '/ambiguous_dir_d'))
+        assert.True(os.remove(test_dir .. '/ambiguous_dir_e/with_file'))
+        assert.True(os.remove(test_dir .. '/ambiguous_dir_e'))
+        assert.True(os.remove(test_dir .. '/ambiguous_dir_f'))
+        assert.True(os.remove(test_dir .. '/just_a_directory/with_file'))
         assert.True(os.remove(test_dir .. '/just_a_directory'))
         assert.True(os.remove(test_dir .. '/localcommand'))
+        assert.True(os.remove(test_dir .. '/true/with_file'))
         assert.True(os.remove(test_dir .. '/true'))
         assert.True(os.remove(test_dir))
         remove_test_path_dir(test_path)
--- a/spec/awful/completion_spec.lua
+++ b/spec/awful/completion_spec.lua
@@ -137,12 +137,12 @@ describe("awful.completion.shell", function()
 
     if has_bash then
         it("completes local file (bash)", function()
-            assert.same(shell('ls ', 4, 1, 'bash'), {'ls localcommand', 16, {'localcommand'}})
+            assert.same(shell('ls l', 5, 1, 'bash'), {'ls localcommand', 16, {'localcommand'}})
         end)
     end
     if has_zsh then
         it("completes local file (zsh)", function()
-            assert.same(shell('ls ', 4, 1, 'zsh'), {'ls localcommand', 16, {'localcommand'}})
+            assert.same(shell('ls l', 5, 1, 'zsh'), {'ls localcommand', 16, {'localcommand'}})
         end)
     end
 end)
--- a/spec/awful/completion_spec.lua
+++ b/spec/awful/completion_spec.lua
@@ -6,6 +6,7 @@ end
 local gfs = require("gears.filesystem")
 local Gio = require("lgi").Gio
 local GLib = require("lgi").GLib
+local lfs = require("lfs")
 
 local has_bash = GLib.find_program_in_path("bash")
 local has_zsh = GLib.find_program_in_path("zsh")
@@ -60,12 +61,14 @@ end
 
 describe("awful.completion.shell in empty directory", function()
     local orig_popen = io.popen
+    local orig_dir = lfs.currentdir()
 
     setup(function()
         test_dir = get_test_dir()
         io.popen = function(...)  --luacheck: ignore
             return orig_popen(string.format('cd %s && ', test_dir) .. ...)
         end
+        lfs.chdir(test_dir)
         test_path = get_test_path_dir()
     end)
 
@@ -73,6 +76,7 @@ describe("awful.completion.shell in empty directory", function()
         assert.True(os.remove(test_dir))
         io.popen = orig_popen  --luacheck: ignore
         remove_test_path_dir(test_path)
+        lfs.chdir(orig_dir)
     end)
 
     if has_bash then
@@ -92,6 +96,7 @@ end)
 
 describe("awful.completion.shell", function()
     local orig_popen = io.popen
+    local orig_dir = lfs.currentdir()
 
     setup(function()
         test_dir = get_test_dir()
@@ -102,6 +107,7 @@ describe("awful.completion.shell", function()
         io.popen = function(...)  --luacheck: ignore
             return orig_popen(string.format('cd %s && ', test_dir) .. ...)
         end
+        lfs.chdir(test_dir)
         test_path = get_test_path_dir()
     end)
 
@@ -112,6 +118,7 @@ describe("awful.completion.shell", function()
         assert.True(os.remove(test_dir))
         io.popen = orig_popen  --luacheck: ignore
         remove_test_path_dir(test_path)
+        lfs.chdir(orig_dir)
     end)
 
     if has_bash then
--- a/spec/awful/completion_spec.lua
+++ b/spec/awful/completion_spec.lua
@@ -60,21 +60,16 @@ local function get_test_dir()
 end
 
 describe("awful.completion.shell in empty directory", function()
-    local orig_popen = io.popen
     local orig_dir = lfs.currentdir()
 
     setup(function()
         test_dir = get_test_dir()
-        io.popen = function(...)  --luacheck: ignore
-            return orig_popen(string.format('cd %s && ', test_dir) .. ...)
-        end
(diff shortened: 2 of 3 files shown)


blueyed:
This might be covered already in https://github.com/awesomeWM/awesome/pull/1736.
Can you test that, please?

btw: I do not like the extra `os.execute` there - it is already too slow.

FloGa:
The referenced patch does not solve the problem. Without completely understanding the code, you mention Zsh in the issue, but I use Bash. Maybe your patch just works for Zsh?

If you don't like the extra syscall, can we access $PATH via Lua, so we can search the path for the desired executable?

blueyed:
@FloGa 
It also has non-zsh changes, e.g. https://github.com/awesomeWM/awesome/pull/1736/files#diff-9ceb43126295a0b6fc91e89b488b2b15L96, but it is only partly related.

I would say to get #1736 merged, and then improve from there regarding this PR.

btw: shouldn't you have to use `./foo` to refer to the command `foo` in the current dir?
(and then the linked change above kicks in, since without the change it would not be considered a command).

But I see that this issue might be something else, and we should cover it through a test after #1736 then.

blueyed:
So, to build upon #1736 some easy test instructions would be required.
I.e. do you expect `foo` to be completed in the current dir?
Do you have `.` in PATH?
etc.

FloGa:
Thank you for your feedback! I applied the whole patch, but my described problem persists, so it appears to be something else.

> I would say to get #1736 merged, and then improve from there regarding this PR.

Yeah, that would be best, so we don't accidently work against each other.

> btw: shouldn't you have to use ./foo to refer to the command foo in the current dir?

That is true if `foo` is in the current working directory. What I described was that `foo` sits somewhere in $PATH, like /usr/local/bin/foo, so basic bashcomp finds it. Now, if there is also a directory foo in the current working directory, then this code kicks in (https://github.com/awesomeWM/awesome/blob/3d1b2c2a5eb3585ee1377b7933a87299101d0959/lib/awful/completion.lua#L137):

    if os.execute("test -d " .. string.format('%q', line)) == 0 then
        line = line .. "/"
    end

And now "foo" becomes "foo/". I have to manually delete the trailing slash if I want to call `foo`. Sure, it is not a critical bug, but it's annoying nevertheless and it differs from the behavior you get in a terminal. Furthermore, dealing with directories is something that bashcomp should handle, not a script afterwards.

As I mentioned in IRC this morning, my guess is, that completion.lua does not work well with current versions of bashcomp. Some time ago, they introduced on-demand loading of completion scripts and therefore they changed the folder structure. The lua script tries to source `/etc/bash_completion`, but (at least in Gentoo) this file does not exist anymore, it has been deleted some years ago.

I will do some research and testing in the next days, maybe I can come up with something. In the optimal case, bashcomp generates the completions and the Lua script just displays them, without any filtering afterwards.

FloGa:
Regarding your comment just before my latest one: If `.` or `$PWD` for this sake is *not* in `$PATH`, then I would *not* expect an executable in the current dir to be completed (except when I start it with `./`). I do not have `.` in my `$PATH`.

blueyed:
My first impression is: let's get  rid of `os.execute("test -d "` (an existing os.execute) also.

As for zsh, I have some pretty nice WIP which calls into the actual shell completion system to get the same output like from a shell.
IIRC there is nothing pushed on GitHub, but only IRC/sprunge/my mind.

psychon:
> let's get rid of os.execute("test -d " (an existing os.execute) also.

Easy: Either `gears.filesystem.dir_readable` or `gears.filesystem.is_dir` (technically the later is equivalent to the current test)

FloGa:
That would be a nice optimization compared to the os.execute call, but in the long term, it would just be another workaround to the actual problem, namely bashcomp not working correctly. But maybe that should be treated as a separate issue.

blueyed:
@FloGa 
Yeah.  As for Zsh I have some proof of concept that uses a real shell in the background for completions (based on https://github.com/Valodim/zsh-capture-completion).  Maybe something similar would be possible with Bash then, too?

blueyed:
And since #1736 is merged now, you might want to rebase this and add some test(s) based on the ones I've added there?!

FloGa:
@blueyed: Sure, I will do it in the next days.

FloGa:
Sorry, that's my first time rebasing within a Pull Request on Github: When I force push my rebased branch to *my* fork, will this PR get updated automatically or do I have to do something afterwards?

psychon:
@FloGa When you force-push to your `FloGa-patch-complete-directory` branch it automatically updates this PR.

FloGa:
I hope I haven't jumped the shark now. I got rid of the `os.execute` in favor of `gears.filesystem` and added some test cases in the specs file. Please note that I don't use zsh and I can't manage to get the directory completion working as it does in bash (hence the commented out assertion).

spec/awful/completion_spec.lua
```diff
@@ -4,6 +4,7 @@ local shell = function(...)
     return {command, pos, matches}
 end
 local gfs = require("gears.filesystem")
+local lfs = require("lfs")
```
psychon: So far awesome does not use `lfs`...

FloGa: `busted` depends on `lfs`. Since `busted` is used for spec testing in awesome, `lfs` has to be present, too.

blueyed:
btw: https://github.com/awesomeWM/awesome/pull/1889 has also some fixes for the tests.

spec/awful/completion_spec.lua
blueyed: Nice.

spec/awful/completion_spec.lua
```diff
@@ -55,8 +60,20 @@ describe("awful.completion.shell", function()
...
+    it("completes command together with local directory", function()
+        assert.same(shell('true', 5, 1, 'zsh'), {'true', 5, {'true', 'true/'}})
+        assert.same(shell('true', 5, 1, 'bash'), {'true', 5, {'true', 'true/'}})
+        assert.same(shell('true', 5, 1, nil), {'true', 5, {'true', 'true/'}})
+    end)
+
+    it("completes local directory", function()
+        -- TODO: Does not work with zsh yet!
+        --assert.same(shell('just_a', 7, 1, 'zsh'), {'just_a_directory/', 18, {'just_a_directory/'}})
+        assert.same(shell('just_a', 7, 1, 'bash'), {'just_a_directory/', 18, {'just_a_directory/'}})
```
blueyed: I can look into Zsh here, but think that it should be handled by a proper Zsh completion daemon (https://github.com/awesomeWM/awesome/issues/1902) in the long run.
This would then just use the user's config for completion(s).

FloGa:
Rebased on current master, after #1889. Changed my tests to match your template with `if has_*`.

lib/awful/completion.lua
```diff
@@ -154,10 +161,25 @@ function completion.shell(command, cur_pos, ncomp, shell)
         while true do
```
blueyed: Maybe we should also get rid of the `sort -u`, since the shell should specify the order already?

lib/awful/completion.lua
```diff
@@ -154,10 +161,25 @@ function completion.shell(command, cur_pos, ncomp, shell)
...
+                        end
+                    end
+                end
+            end
+            if gfs.is_dir(line) then
+                table.insert(output, bash_escape(line .. "/"))
+            elseif comptype == "file" then
+                table.insert(output, bash_escape(line))
             end
-            table.insert(output, bash_escape(line))
```
blueyed: I'm not sure that I like the added complexity here..
After all the shell completion should provide all/most of it already, and if e.g. "zsh" would complete an alias or function, this would now not get added anymore?! (since it is not executable)

lib/awful/completion.lua
FloGa: I thought so, too, but:

    $ mkdir true && /usr/bin/env bash -c 'compgen -A command -o dirnames true'
    true
    true
    true

- builtin
- /bin/true
- directory

Also note that the directory does *not* get a slash via bashcomp. Apparently this is done by readline afterwards ... I haven't found a way around this yet. Maybe we can somehow interact with readline in a non-interactive context, but that will involve more research and testing. But anyways, I'm afraid for the moment we cannot ditch the `sort -u` ...

lib/awful/completion.lua
FloGa: Sadly, bashcomp does not provide everything ... The main reason I use this overly complex construct is because we cannot know if the current item is "just" a directory or if there is also an executable somewhere in the path.

Regarding aliases: With `-A command`, we already exclude aliases and functions (at least with bash). If you want to complete aliases, too, then the bash-code needs to be updated.

I think, the whole confusion arises from the requirement to complete local directories, which is by the way not even consistent in bash itself. A directory search for example is only done if absolutely no other match has been found. And if a directory is found, a slash is only added if there is no command/executable/alias with the same name. So maybe we should first work out, how much of the (sometimes strange) behavior we really want to re-produce. Wouldn't commands (and aliases for this sake) be enough? Is it really necessary to add directories to the completion results? If we just want to complete commands, then the executable check as well as the directory check can be ditched altogether in favor of a slightly modified completion command.

lib/awful/completion.lua
blueyed: I see.  I am more concerned about the `sort`, not the `unique`.
Could we use `uniq` instead?

lib/awful/completion.lua
blueyed: > Wouldn't commands (and aliases for this sake) be enough? Is it really necessary to add directories to the completion results?

Making it simpler would be good IMHO.
Directories should get completed after `./` then.

Current master does not complete (local) directories for me in Zsh, i.e. `Down` does not complete "Downloads", but `./Down` does.
This is what I would expect for bash then, too.

lib/awful/completion.lua
FloGa: $ /usr/bin/env bash -c 'compgen -A command tru'
    true
    trust
    truncate
    true

Damn ... Also note, with or without a directory `./true`, the previous command yields *two* `true`s. However the command I used in my previous post yields *three* `true`s if the directory exists. It's crazy how I didn't realize these inconsistencies before. Obviously bash / readline does much more post-processing than I was aware of.

To summarize: It seems we still need `sort -u`.

lib/awful/completion.lua
FloGa: Yes, this sounds reasonable.

Do you want the trailing slash on directories? Then we need a check for `starts_with "./"` and `is_dir`, but it's way shorter than my current construct.

I will push these changes as separate commits to keep this discussion, but I can squash it after review.

blueyed: > Do you want the trailing slash on directories? 

Yes, I like it.

lib/awful/completion.lua
blueyed: The `sort -u` will certainly get removed/skipped with #1902, so it is (maybe already?) specific to bash.

Would be nice to have your example as a test, in case you have not added it already.

blueyed: btw: in my Zsh config I get three completions for `tru` (external command, builtin command, local directory), but only because it could also be `truncate` or `trust`.
`true` completes to just `true` as a command then.
If there is also `truedir` I get completions for `true` (external command, builtin) and `true/`, `truedir/`.

blueyed:
I think the overall changes/diff looks good.
Should I look at the zsh test issue?

blueyed:
There is one luacheck issue:

   spec/awful/completion_spec.lua:148:16: empty if branch

FloGa:
Yes, the empty branch is the commented out ZSH test. If you could look at that it would be great. My main shell is Bash and I am yet to understand the ZSH basics ...

lib/awful/completion.lua
FloGa: > Would be nice to have your example as a test, in case you have not added it already.

That could be difficult, because we can't possibly know what commands are on the users' systems. What if they don't have command "trust" or they have a command "truecrypt"?

However, we could create a small directory tree with some executable files and use a custom `PATH` while executing the consistency / sorting tests. I'll try something and commit it.

> in my Zsh config [...]

I get the impression that ZSH's completion might be a bit more intelligent (and more powerful, as I have heard ...) than bashcomp. Should I switch shells? ;-)

blueyed: > However, we could create a small directory tree with some executable files and use a custom `PATH` while executing the consistency / sorting tests.

Yes, or patch the executable-check in case we use a custom function.

> Should I switch shells? ;-)

Only after this PR is merged.. ;)

FloGa: Huh, I accidently mentioned a problem that incorrectly breaks the tests. If there is a command `truecrypt` (or `true_foobar`, or `true_*` ...) in `PATH`, then most (if not all) tests regarding `true` will fail, because you assume to be just one completion item, namely `true`. It isn't directly related with the current PR, so I'll start another one for that (as soon as I have found out how to deal with it, maybe sublist matching instead of list equality?).

FloGa:
Just in case you haven't been notified about the latest commits two weeks ago: I added another test which checks for correct ordering of completion items, as discussed in a subthread. Also, I replaced another popen-override with `lfs.chdir()`, which I seem to have missed the first time. At last, I extended the comment block on the ZSH cases, so there no longer is an "empty branch" luacheck issue.

As I understood I am now waiting for you to fix/implement the missing ZSH checks which I am unable to get right. If I can do anything else to help, just let me know!

blueyed:
@FloGa 
Sorry for being unresponsive.
This needs to be rebased now.

I've just tried it manually, which resulted in two failing tests, so I probably did it wrong.. ;)
```
Failure → spec/awful/completion_spec.lua @ 147
awful.completion.shell completes local file (bash)
spec/awful/completion_spec.lua:148: Expected objects to be the same.
Passed in:
(table) {
 *[1] = 'ls localcommand'
  [2] = 16
  [3] = {
    [1] = 'localcommand' } }
Expected:
(table) {
 *[1] = 'ls just_a_directory/'
  [2] = 21
  [3] = {
    [1] = 'just_a_directory/'
    [2] = 'localcommand'
    [3] = 'true/' } }

Failure → spec/awful/completion_spec.lua @ 152
awful.completion.shell completes local file (zsh)
spec/awful/completion_spec.lua:153: Expected objects to be the same.
Passed in:
(table) {
 *[1] = 'ls localcommand'
  [2] = 16
  [3] = {
    [1] = 'localcommand' } }
Expected:
(table) {
 *[1] = 'ls just_a_directory/'
  [2] = 21
  [3] = {
    [1] = 'just_a_directory/'
    [2] = 'localcommand'
    [3] = 'true/' } }
```

I will look into this again once you've updated/rebased it.

FloGa:
I rebased this PR to current master. With it, I resolved the fixups and squashes and also deleted commits along with their revert-commits. `make check-unit` is all green on my end.

blueyed:
@FloGa 
Fixed the tests for Zsh: it only completes non-empty dirs..! :)

Feel free to squash this in.

Then, we should also ensure that tests pass for every commit (this fails currently: https://travis-ci.org/awesomeWM/awesome/jobs/262370707#L6276).

FloGa:
> Fixed the tests for Zsh: it only completes non-empty dirs..! :)

Huh, interesting. But also reasonable, what should I want to call in an empty directory anyway? :-P
I will squash / reorder the commits tonight after work.

> Then, we should also ensure that tests pass for every commit

Sure, I can do this, probably by just reordering the commits. But in Test Driven Development I am used to failing tests. If you plan a new feature or you find a bug, you first write test cases for exactly this specific feature or bug. Of course these tests will fail, that's the reason why you created them in the first place. It's the following commits that modify your actual code in order for the tests to pass. Only then will the branch be merged.

It's your decision after all and I respect it, but I find it an overly strict policy for *every commit* to pass the tests.

blueyed:
> Test Driven Development
> I find it an overly strict policy for every commit to pass the tests.

It is meant to make git-bisect work more easily: you do not want to have failing (unrelated) tests when looking for a regression etc.
While you would typically then run only your new test with git-bisect, you might still want to run all tests when bisecting based on changed deps etc.

I would typically squash the fix and the commit together: it makes sense to have them in the same commit anyway,

FloGa:
> git-bisect

Ah, you're right, I always forget about that one. So yes, re-ordering makes sense then.

However, I would argue that squashing them together might not be the best thing to do. If you messed up the fix and want to revert it some time later, then you do not want to lose your test suite with it. By just reverting the fix, you will be left with broken tests again, but you could then squash your better_fix with the revert commit afterwards.

But I think we're going off-topic here. ;-)

FloGa:
I squashed your fixes into my commits and re-ordered the commits so that every single commit passes the tests now.

Elv13:
Thanks

PR #1877, opened 2017-06-26
awful.spawn: fix doc for spawn.easy_async{,_with_shell}

blueyed:
Using `@function` in 547c0e4b was wrong.

/cc @copycat-killer

### DIFF
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -323,14 +323,12 @@ end
 --- Asynchronously spawn a program and capture its output.
 -- (wraps `spawn.with_line_callback`).
 -- @tparam string|table cmd The command.
--- @function callback Function with the following arguments:
+-- @tab callback Function with the following arguments
 --   @tparam string callback.stdout Output on stdout.
 --   @tparam string callback.stderr Output on stderr.
---   @tparam string callback.exitreason Exit Reason.
---   @tparam integer callback.exitcode Exit code.
---   The exitreason argument can either be "exit" or "signal".
---   For "exit" reason it's the exit code.
---   For "signal" reason it's the signal causing process termination.
+--   @tparam string callback.exitreason Exit reason ("exit" or "signal").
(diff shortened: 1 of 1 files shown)


blueyed:
This is required to fix master.

lib/awful/spawn.lua
```diff
@@ -323,14 +323,12 @@ end
 --- Asynchronously spawn a program and capture its output.
 -- (wraps `spawn.with_line_callback`).
 -- @tparam string|table cmd The command.
--- @function callback Function with the following arguments:
+-- @tab callback Function with the following arguments
```
copycat-killer: Could you explain why is it correct to use `@tab` for a function?

copycat-killer: Also, you reverted what I believed were necessary corrections.

psychon: `@tab` is apparently our own invention (`docs/config.ld` defines it) and it is just a short-hand for `@tparam table`.
`@function` is wrong since it documents a (global) function, for example `math.max`. It does not document parameters to a function.
(`@table` would also be wrong here, which is likely what you were thinking of: It documents a global table and not a table argument.)

copycat-killer: Yeah, I thought `@tab` was a synonymous of `@table`.

psychon: ...so did I until 15 minutes ago

blueyed: > Also, you reverted what I believed were necessary corrections.

Do you think this still after the discussion?

Should we rather use `@tparam table` for explicitness?

Elv13: > Should we rather use @tparam table for explicitness?

It's what I personally did everywhere so far for this reason. I also found it quite confusing when I started to work on the doc 18 months ago.

blueyed: I see, agreed.
From 289bc26a (where I've added it) I was under the impression that it even should be there by default.

Issue #1888, opened 2017-06-30
tests: cleanup temporary dirs


blueyed:
https://github.com/awesomeWM/awesome/blob/cbb8fd7ac408d229233a3b5490adf35d58a85d79/spec/awful/completion_spec.lua#L34 and https://github.com/awesomeWM/awesome/blob/cbb8fd7ac408d229233a3b5490adf35d58a85d79/spec/awful/completion_spec.lua#L12 should clean up after them.

What is the way to go here, @psychon?

psychon:
I thought the `teadown` calsl in there already handles the needed cleanup...?

blueyed:
Oh, yeah - missed that.
However, it currently fails since the dir is not empty.

We do not have a helper for deleting dirs recursively already, do we?
Does not seem to be supported by Gio directly - I've found https://github.com/cgwalters/glib-async-rm-rf/blob/master/rm-rf-async.c.

PR #1889, opened 2017-07-01
tests: spec/awful/completion_spec.lua: fix teardown and missing bash/zsh

blueyed:
Fixes https://github.com/awesomeWM/awesome/issues/1888.

### DIFF
--- a/spec/awful/completion_spec.lua
+++ b/spec/awful/completion_spec.lua
@@ -4,25 +4,55 @@ local shell = function(...)
     return {command, pos, matches}
 end
 local gfs = require("gears.filesystem")
+local Gio = require("lgi").Gio
+local GLib = require("lgi").GLib
+
+local has_bash = GLib.find_program_in_path("bash")
+local has_zsh = GLib.find_program_in_path("zsh")
+
+if not has_bash and not has_zsh then
+    print('Skipping spec/awful/completion_spec.lua: bash and zsh are not installed.')
+    return
+end
+
+local test_dir
+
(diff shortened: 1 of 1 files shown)


spec/awful/completion_spec.lua
```diff
@@ -5,18 +5,29 @@ local shell = function(...)
 end
 local gfs = require("gears.filesystem")
 
+local test_dir
+
+--- Get and create a temporary test dir based on `pat`, where %d gets replaced by
+-- the current PID.
+local function get_test_dir(pat)
+    local pid = io.open("/proc/self/stat", "r"):read():gmatch("%S+")()
```
psychon: Linuxism, not good.

I suggest `https://developer.gnome.org/gio/stable/GFile.html#g-file-new-tmp` and `https://developer.gnome.org/gio/stable/GFile.html#g-file-get-path`. However, that only gets you a file and not a directory... remove that file and create a directory in its place?

blueyed: Pushed a fixup for this.

psychon: Yup, looks good. However, your `fixup!` is for the wrong commit, I think. Besides that, 👍 .

blueyed:
Squashed.
Would like to get feedback from @FloGa, since it conflicts with https://github.com/awesomeWM/awesome/pull/1876.

FloGa:
@blueyed: Looks good from my side. I don't think I will get problems rebasing my branch on yours.

blueyed:
Awesome!

Issue #740, opened 2016-03-04
Build Lua 5.3 with -DLUA_USE_APICHECK on Travis


psychon:
This define would have helped figuring out #735 (However, that issue was never reproduced on Travis). Still, since we are building our own version of Lua 5.3 on Travis anyway, it might make sense to define `LUA_USE_APICHECK` there.

Quick hack to Lua 5.3.2's source code to show that this would have helped:
```patch
diff --git a/src/lua.c b/src/lua.c
index 545d23d..c2a58db 100644
--- a/src/lua.c
+++ b/src/lua.c
@@ -593,6 +593,9 @@ static int pmain (lua_State *L) {
 int main (int argc, char **argv) {
   int status, result;
   lua_State *L = luaL_newstate();  /* create state */
+  printf("%d\n", lua_gettop(L));
+  lua_pop(L, 1);
+  printf("%d\n", lua_gettop(L));
   if (L == NULL) {
     l_message(argv[0], "cannot create state: not enough memory");
     return EXIT_FAILURE;
diff --git a/src/luaconf.h b/src/luaconf.h
index b4ed500..d0b9819 100644
--- a/src/luaconf.h
+++ b/src/luaconf.h
@@ -680,9 +680,12 @@
 @@ LUA_USE_APICHECK turns on several consistency checks on the C API.
 ** Define it as a help when debugging C code.
 */
+#define LUA_USE_APICHECK
 #if defined(LUA_USE_APICHECK)
 #include <assert.h>
 #define luai_apicheck(l,e)	assert(e)
+#else
+#error
 #endif
 
 /* }================================================================== */
```
```
$ make -s linux && ./src/lua
0
lua: lapi.c:182: lua_settop: Assertion `(-(idx+1) <= (L->top - (func + 1))) && "invalid new top"' failed.
[1]    24001 abort (core dumped)  ./src/lua
```

PR #1890, opened 2017-07-01
[WIP] Travis: use debian-sid source / try to use sudo: false

blueyed:
TODO:
- [ ] not building Lua 5.3 from source will not use `LUA_USE_APICHECK` (cc8e10e4, https://github.com/awesomeWM/awesome/issues/740)
- [ ] provide/host an own .deb that would configure APT?!
  - 'APT::Install-Recommends "false";' in /etc/apt/apt.conf.d/no-recommends
  - depend on the build dependencies?!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,35 +19,32 @@ addons:
     - lua-lgi
     - luarocks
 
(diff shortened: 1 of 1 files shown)


blueyed:
Going with Docker instead.

PR #1892, opened 2017-07-01
cmake: abort if Lua was not found

blueyed:
This will usually result in errors during linking, and the error message
itself is not obvious to find.
Therefore just abort in this case.

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -67,6 +67,12 @@ a_find_program(CONVERT_EXECUTABLE convert TRUE)
 include(FindPkgConfig)
 # lua
 include(FindLua)
+if (NOT LUA_FOUND)
+    message(FATAL_ERROR
+        "Could not find Lua. See the error above.\n"
+        "You might want to hint it using the LUA_DIR environment variable, "
+        "or set the LUA_INCLUDE_DIR / LUA_LIBRARY CMake variables.")
+endif()
 # }}}
 
 # {{{ Check if documentation can be build


awesomeConfig.cmake
```diff
@@ -67,6 +67,12 @@ a_find_program(CONVERT_EXECUTABLE convert TRUE)
 include(FindPkgConfig)
 # lua
 include(FindLua)
+if (NOT LUA_FOUND)
+    message(FATAL_ERROR
+        "Could not find Lua. See the error above.\n"
+        "You might want to hint it using the LUA_DIR environment variable, "
+        "or set the LUA_INCLUDE_DIR / LUA_LIBRARY CMake variables.")
+endif()
```
psychon: Can you try replacing `include(FindLua)` with `find_package(Lua MODULE REQUIRED)` to see if it has the same effect?
https://cmake.org/cmake/help/v3.0/command/find_package.html

psychon: ...so that this whole `if` is not required

blueyed: That would work:
```
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.3.7") 
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Lua (missing: LUA_LIBRARIES) (found version "5.2.4")
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:377 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake/Modules/FindLua.cmake:193 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  awesomeConfig.cmake:69 (find_package)
  CMakeLists.txt:32 (include)


```

However, I think it is useful to have the extra message from the custom `if` block:
```
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.3.7") 
-- Could NOT find Lua (missing:  LUA_LIBRARIES) (found version "5.2.4")
CMake Error at awesomeConfig.cmake:71 (message):
  Could not find Lua.  See the error above.

  You might want to hint it using the LUA_DIR environment variable, or set
  the LUA_INCLUDE_DIR / LUA_LIBRARY CMake variables.
  CMakeLists.txt:32 (include)


```

What do you think?

psychon: If so, shouldn't that be done in CMake instead of us working around CMake? (I'm not sure how well it fits into this PackageHandleStandardArgs-macro, however...).
Also, I don't think the hint to `LUA_DIR` works. If I read `FindLua.cmake` correctly, it looks for `${subdir}/lua.h` where `${subdir}` is `include/luaXY`, `include/luaX.Y` and `include/lua-X.Y`, and `X` is always 5 and `Y` ranges from 0 to 3. So, `LUA_DIR` is useless...?

Thus, this would only help users who have a non-system (or non-package-manager?) installation of Lua. That's basically no one, is it?

The more common error is version mismatches (5.2 headers, but 5.3 library, or something like that) and we cannot do much about that.

blueyed: LUA_DIR helped me on Alpine Linux.
It is required to find the actual lib there, and gets used as HINT somewhere.

psychon: Oh right, I missed that it is also used as a HINT for the library. Sorry.

PR #1897, opened 2017-07-01
lib/awful/completion.lua: handle missing SHELL environ var

blueyed:


spec/awful/completion_spec.lua
```diff
@@ -60,4 +60,22 @@ describe("awful.completion.shell", function()
     end)
 end)
 
+describe("awful.completion.shell without $SHELL", function()
+    local orig_popen = io.popen
+
+    setup(function()
+        io.popen = function(...)  --luacheck: ignore
+            return orig_popen('env -u SHELL ' .. ...)
```
psychon: Uhm... this unsets `$SHELL` inside of the shell. This does not have any effects inside of awesome, no?

You could monkey-patch `os.getenv` instead, but doesn't `awful.completion` cache the result of looking at `os.getenv`?

blueyed: >  but doesn't awful.completion cache the result of looking at os.getenv?

Not really: https://github.com/awesomeWM/awesome/pull/1897/files#diff-9ceb43126295a0b6fc91e89b488b2b15R101.

It could be stored on the module, and then that could be monkey-patched, but the test was meant to actually test the change in this PR (the call to `os.getenv` was chained previously).

So monkey-patching `os.getenv` is the way to go then?

psychon: > So monkey-patching os.getenv is the way to go then?

Yup, because your commit really does not do what you wanted it to do. :-)

PR #1898, opened 2017-07-01
tests: fix test-menubar for when there are no desktop files

blueyed:
This happens with Docker based tests, where no (relevant) .desktop files are installed.

### DIFF
--- a/tests/test-menubar.lua
+++ b/tests/test-menubar.lua
@@ -3,6 +3,12 @@
 local runner = require("_runner")
 local menubar = require("menubar")
 
+local menubar_refreshed = false
+
+function menubar.refresh(scr)
+    menubar_refreshed = scr or true
+end
+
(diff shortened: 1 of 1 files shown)


tests/test-menubar.lua
```diff
@@ -3,6 +3,12 @@
 local runner = require("_runner")
 local menubar = require("menubar")
 
+local menubar_refreshed = false
+
+function menubar.refresh(scr)
+    menubar_refreshed = scr or true
+end
```
psychon: How about
```lua
local orig_refresh = menubar.refresh
function menubar.refresh(...)
    menubar_refreshed = true -- What was your scr or true supposed to do?
    orig_refresh(...)
end
```
That way more of the original code runs and gets checked for "does it cause errors"?

blueyed: Adjusted.

> What was your scr or true supposed to do?

Silencing luacheck.. using `...` is much better.

PR #1904, opened 2017-07-03
Improve tests/run.sh

blueyed:
- add colors
- count test files
- display error summary
- filter out a_dbus_connect warnings and "Test finished successfully." message

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -82,7 +82,8 @@ else
 fi
 
 # Add test dir (for _runner.lua).
-awesome_options=($AWESOME_OPTIONS --search lib --search $this_dir)
+# shellcheck disable=SC2206
+awesome_options=($AWESOME_OPTIONS --search lib --search "$this_dir")
 export XDG_CONFIG_HOME="$build_dir"
 
 # Cleanup on errors / aborting.
@@ -187,13 +188,22 @@ start_awesome() {
     wait_until_success "wait for awesome startup via awesome-client" "dbus-send --reply-timeout=100 --dest=org.awesomewm.awful --print-reply / org.awesomewm.awful.Remote.Eval 'string:return 1' 2>&1"
 }
 
-# Count errors.
-errors=0
+if command -v tput >/dev/null; then
+    color_red() { tput setaf 1; }
+    color_reset() { tput sgr0; }
+else
+    color_red() { :; }
+    color_reset() { :; }
+fi
+
+count_tests=0
+errors=()
 # Seconds after when awesome gets killed.
 timeout_stale=180 # FIXME This should be no more than 60s
 
 for f in $tests; do
     echo "== Running $f =="
+    (( ++count_tests ))
 
     start_awesome
 
@@ -202,7 +212,7 @@ for f in $tests; do
             f=${f#tests/}
         else
             echo "===> ERROR $f is not readable! <==="
-            ((errors++))
+            errors+=("$f is not readable.")
             continue
         fi
     fi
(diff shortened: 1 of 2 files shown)


blueyed:
The change reported by codecov for tag.lua looks weird?!
https://codecov.io/gh/awesomeWM/awesome/compare/4eddfacf5121a8d51a329aea7dd500e022809c9e...e45b7dcf83b1067221765c9ac8cddf04a5bf8252/changes#L1446

tests/run.sh
```diff
@@ -82,7 +82,8 @@ else
 fi
 
 # Add test dir (for _runner.lua).
-awesome_options=($AWESOME_OPTIONS --search lib --search $this_dir)
+# shellcheck disable=SC2206
```
psychon: Huh? What's this? And what's SC2206?

tests/run.sh
```diff
@@ -211,7 +221,10 @@ for f in $tests; do
     DISPLAY=$D "$AWESOME_CLIENT" 2>&1 < "$f"
 
     # Tail the log and quit, when awesome quits.
-    tail -n 100000 -s 0.1 -f --pid "$awesome_pid" "$awesome_log"
+    tail -n 100000 -s 0.1 -f --pid "$awesome_pid" "$awesome_log" \
+        | grep -vE '^.{19} W: awesome: a_dbus_connect:[0-9]+: Could not
+    connect to D-Bus system bus:' \
```
psychon: Is the line break in the middle of the string intended? Does the `grep` still work? Really?

tests/run.sh
```diff
@@ -211,7 +221,10 @@ for f in $tests; do
...
+        | grep -v '^Test finished successfully\.' || true
```
psychon: Why `\.` instead of `.`? There is no `-E` here.

psychon:
> The change reported by codecov for tag.lua looks weird?!

`test-awful-rules` uses `new_tag=true` in a rule. That makes `awful.rules` create a tag with `volatile=true`. However, the test does not `:kill` its test client and wait for it to be done and so we end up with a race: When `_runner` `:kill`s the client and it closes quickly enough so that awesome still handles the event in this main loop iteration, then the tag is deleted and this part of the code is run. If the client loses the race, that code is not executed.

At least that would be my guess. So if you want, feel free to add a new step to `test_awful_rules` that `:kill`s all visible clients and waits for them to be gone.

tests/run.sh
blueyed: https://www.shellcheck.net/
You really want this for editing shell code.
Neomake has integration for it.

blueyed: https://github.com/koalaman/shellcheck/wiki/SC2206

tests/run.sh
blueyed: The dot is meant to be matched literal, and needs to be escaped then (also without -E), no?

psychon: Okay, I obviously do not know `grep`...

tests/run.sh
blueyed: Good catch.. need to retest this.

blueyed:
Will look into that.
The steps handling seems to be a bit uncommon in https://github.com/awesomeWM/awesome/blob/9938fc8a05fb19d2d12111f6a239b75081131da7/tests/test-awful-rules.lua#L299?!

psychon:
@blueyed Well, that code immediately spawns all the clients that it wants to test, wait for them to finish spawning and then generates steps for checking if they spawned as expected. That does make some sense, I think.

blueyed:
@psychon 
Any idea why the last commit (i.e. only killing it separately with test-awful-rules) is necessary to not make the coverage drop?
Why is there more coverage if the runner kills them, without the extra step?
Should I try reverting the last commit to ensure that it is not something else, e.g. codecov acting up?

psychon:
There are two kinds of "less code covered" that I could spot:

The code that preserves client stacking order across restarts no longer runs. This is fine since nothing tested that it worked correctly anyway. Plus, this coverage was racy: The client(s) were already `:kill()`d, so if they managed to exit quickly enough then this code wouldn't run anway.

The other changes are related to shaped clients. As far as I can see, we do not have any shaped clients in the tests (well, we do have `client.get()[1].shape = rounded_rect` in `test-resize.lua`, but I think this is not the kind of shape that is no longer covered. Rather, the diff says that the client previously set a shape on itself, and not we set one on it through Lua). No idea what is going on here.

I would say to just ignore this coverage changes. These code paths are not really tested anyway, just "accidentally covered".

psychon:
> Why is there more coverage if the runner kills them, without the extra step?

Because previously clients were killed without waiting for them to really disappear.

blueyed:
OK.
Removed de101acb again, and will merge when CI finished.

psychon:
Thanks

PR #1907, opened 2017-07-04
[WIP] Docker based builds/tests

blueyed:


docker/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+# Cleanup, needs to happen after build only, and get chained in a single RUN.
+# RUN apk del awesome-build
+
+ENV LUA_PKG=${LUA_PKG:-lua5.3}
+# For CMake's FindLua.
+ENV LUA_DIR /usr/lib/${LUA_PKG}
+
+# RUN apk add build-base
+# TODO: Why is g++ required?!  Why does cmake "Check for working CXX compiler"?
+RUN apk --no-cache add gcc g++ make musl-dev
```
psychon: Google found https://cmake.org/pipermail/cmake/2010-August/039245.html, but we already call `project` with just `C` as the languages in `awesomeConfig.cmake`...?

Some quick testing suggests that `CMakeLists.txt` must contain the call to `project()`. Even if I move it after `include(awesomeConfig.cmake)` does CMake stop checking for a C++ compiler. Magic?

psychon: Yup, CMake scans `CMakeLists.txt` for `cmake_minimum_required` and `project` early on. If no `project` is found, it synthesises a `PROJECT(Project)` command and adds it before the first line.
https://gitlab.kitware.com/cmake/cmake/blob/43c3afa74538fd7b78bcf534bfb3fa93c3c1f191/Source/cmMakefile.cxx#L1442-1458

blueyed: Fixed by https://github.com/awesomeWM/awesome/pull/1909.

blueyed:
Yay, first (partly) coverage upload succeeded.

docker/Dockerfile
```diff
@@ -0,0 +1,81 @@
+# Dockerfile for awesomeWM.
```
Elv13: Could you move the Dockerfile into a subdirectory? At some point we might/should have many Dockerfiles for jobs like a PPA upload and older distribution support (executed on a timely base instead of each PR changes).

blueyed: It is already in a subdir?!

Inititally I had multiple, e.g. `.dep` for the base dependencies - might come back to this..

Elv13: Yes, but wouldn't another level welcome? I usually see Dockerfile being built with a directory path instead of the Dockerfile absolute path. This allows for image specific assets to be "encapsulated" into the sub directory. Something like:

```sh
$BASE_DIR/docker/docker-awesome_coverage/Dockerfile # The current Alpine base 
$BASE_DIR/docker/docker-awesome_ppa/Dockerfile # Ubuntu LTS with PPA upload (no build)
$BASE_DIR/docker/docker-awesome_dotdeb/Dockerfile # "make package" for Debian stable and testing
$BASE_DIR/docker/docker-awesome_coverity/Dockerfile # Static analysis 
$BASE_DIR/docker/docker-awesome_scanbuild/Dockerfile # Static and dynamic analysis
$BASE_DIR/docker/docker-awesome_bleeding_edge/Dockerfile # Use all the AUR -git / luarocks dependencies to get early breakage warnings.
```

Note that those Dockerfiles can be written over time. I already have a base for the Coverity one (still need to ask for freebies from them. It is usually very expensive, but they it away as long as the code is Free).

blueyed: Yes, makes sense, although I'd prefer shorter names.. ;)
I think for now we can have it at the top level though for our main one.

blueyed:
What is the best way to pass arguments to `check-integration` / `tests/run.sh`?
I've used `$TEST_RUN_ARGS` for now - should that get added to `CMakeLists.txt` then, as (renamed) `$ENV{TEST_INTEGRATION_ARGS}`?

blueyed:
As for caching: we could use the Docker Hub for this in general, i.e. having tagged base images maybe, that could be used by others then, too?
Currently I am trying to `docker save/load` it to Travis' cache.

blueyed:
Still at 10mins with the cached Docker image (not optimized for size though): https://travis-ci.org/awesomeWM/awesome/builds/254663200

blueyed:
Might be relevant for gcov: https://gcc.gnu.org/onlinedocs/gcc/Cross-profiling.html (currently running it in `build` appears to fix it).

Is there a reason why we run two builds with coverage?
We could just report to codecov and coveralls from the same build job?! (this would also make their numbers more comparable)

blueyed:
btw: I've edited the coveralls settings (https://coveralls.io/github/awesomeWM/awesome/settings) to post comments again (which is only possible globally?!), and also to include branch coverage (what codecov is doing with this PR already).

psychon:
> Is there a reason why we run two builds with coverage?

Codecov supports tags ("this code was executed by functional tests"). For this to work, we have to delete the coverage files after uploading. However, for coveralls we need an "all in one" coverage file.

blueyed:
@psychon 
That should be handled by now.

psychon:
Okay, good to know.

So, should some parts of this be done in separate PRs? For example, moving "all the mess in `.travis.yml`" into a shell script without changing anything else could be a PR. Or merging the two coverage builds into one. Those are just two random ideas from the top of my head.
I don't want to dictate how you work, I am just afraid that eventually the `[WIP]` here disappears and we end up with a huge PR. I really do not like huge PRs.

psychon:
> btw: I've edited the coveralls settings (https://coveralls.io/github/awesomeWM/awesome/settings) to post comments again (which is only possible globally?!),

Uhm... why? I like that codecov edits its existing comment instead of posting new ones (well, mostly).

blueyed:
>> I've edited the coveralls settings 

> Uhm... why?

For debugging/feedback.

IIRC I've seen that you could configure it through yml/branch eventually, so I might change that.

> So, should some parts of this be done in separate PRs? 

Yes, and a lot of the recent build changes originated from here already.

> moving "all the mess in .travis.yml" into a shell script without changing anything else could be a PR

Might be hard, since it is refactored / based on the Docker build already.
I think regarding that it is not worth backporting that first, but I will keep it in mind.

> Or merging the two coverage builds into one.

Still WIP and also not trivial to "backport" probably.

> I really do not like huge PRs.

:+1:

blueyed:
The coverage changes (especially for the C code) are coming from covering luajit (Lua 5.1) now only, while we covered Lua 5.2 and 5.3 before.
Should we re-add it for one or both even?

Elv13:
🥇

psychon:
Could we disable coveralls' comments function again? Alternatively: Can I somehow tell GitHub to ignore all comments from coveralls? This generates just too much noise.

blueyed:
Sorry.
Disabled them again.

blueyed:
Rebases, but still not squashed.

PR #1909, opened 2017-07-09
Move project() call to CMakeLists.txt

psychon:
CMake "fakes" a call to "project(Project C CXX)" if it does not see a
project() call in CMakeLists.txt. Since we had this call in a different
file, this default applied. This meant that CMake unnecessarily required
a C++ compiler.

Fix/work-around this by moving the call to project() into CMakeLists.txt

Reference: https://github.com/awesomeWM/awesome/pull/1907#discussion_r125581273

@blueyed This should allow you to get rid of the C++ compiler in your docker build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.0.0)
+project(awesome C)
 
 # Require an out-of-source build. We generate an awesomerc.lua in the build dir
 # from the one in the source dir and this does not work otherwise.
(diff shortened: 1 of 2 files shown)


blueyed:
Yes, it helps.
Thanks!

blueyed:
Also required for the example tests.

psychon:
You mean "`tests/examples/CMakeLists.txt` does not have a `project()-line and so ends up with CMake's default project(Project C CXX)"? I'll take this as a feature request for me to handle.

PR #1911, opened 2017-07-09
CMake: only run ldoc with changed input

blueyed:
This adds a custom command to build the documentation, where you can
specify the generated output then. "ldoc" then depends on this output.

This way "ldoc" is not being run every time, but only if the input files
have changed.  ldoc's config is added therefore to the list of
dependencies to make config changes trigger a rebuild.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,10 +294,14 @@ if(GENERATE_DOC)
     file(COPY ${SOURCE_DIR}/docs/aliases DESTINATION ${BUILD_DIR}/docs)
 
     add_custom_target(ldoc ALL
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -294,10 +294,16 @@ if(GENERATE_DOC)
     file(COPY ${SOURCE_DIR}/docs/aliases DESTINATION ${BUILD_DIR}/docs)
 
     add_custom_target(ldoc ALL
-        COMMAND ${LDOC_EXECUTABLE} .
         WORKING_DIRECTORY ${AWE_DOC_DIR}
```
psychon: I'd also remove the `WORKING_DIRECTORY` here.

PR #1913, opened 2017-07-09
Fix tests/examples/CMakeLists.txt

blueyed:
This still runs all the tests in the configure phase, since the tests
report back what output file (png/svg/none) will be generated.. :/

This could be improved by having this table in CMakeLists directly, or
by having a callback into the tests that only reports back that
information.  The latter would still execute all the tests (via a Lua
process).

### DIFF
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -16,8 +16,9 @@ endif()
 
 cmake_minimum_required(VERSION 3.0.0)
 
-# Get and update the LUA_PATH so the scripts can be executed without Awesome.
-execute_process(COMMAND lua -e print\(package.path\) OUTPUT_VARIABLE "LUA_PATH_")
+# Get and update the LUA_PATH so the scripts can be executed without awesome.
+execute_process(COMMAND lua -e "p = package.path:gsub(';', '\\\\;'); io.stdout:write(p)"
+    OUTPUT_VARIABLE "LUA_PATH_")
 
 # Make sure the system can be called from the test directory
 if(NOT SOURCE_DIR AND ${CMAKE_CURRENT_SOURCE_DIR} MATCHES "/tests/examples")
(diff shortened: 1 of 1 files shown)


Elv13:
I am (slowly) working on having those tests executed as CMake targets and handle this in a cleaner way to fix those issues, but as there is no PR yet, I guess you win. Also, it's summer, so I would rather spend my time under the sun, not in front of a computer ;). There *is* code and it *does* run the tests "fine", but the output is still a bit wrong the shims don't reset themselves properly yet (so the test order affect the output...)

psychon:
> the shims don't reset themselves properly yet (so the test order affect the output...)

How about just causing the shims to be reloaded on the next `require()` for them? Seems quite simple and not tooooooooo ugly to me (just a bit).
`for _, v in ipairs{ "awesome", "client", "whatever_else_you_shim" } do package.loaded[v] = nil end`

blueyed:
Updated, so that it caches the list of source input files.
This finally make `cmake` not run the tests all the time.

It would be nice if it could detect updated source input files (timestamp).  Any idea?
Otherwise I think this is a huge improvement already, and think it is ready for merging.

blueyed:
The comments about the shims are not really related, are they?
Or is it a problem now, if only some tests are being run (via `make`)?

Elv13:
> Or is it a problem now, if only some tests are being run (via make)?

No, it is a problem once you run all test files in the same process. The problem is that loading Lua and LGI 300 times (more so when the coverage is enabled) takes a lot of time. As I document more and more part of the APIs that require shims for more and more parts of Awesome, the new "top level" tests get slower and slower (each). So the end result is that we can no longer add more of them because it gets way, way too slow on the CI.

Until now, the bottleneck was mostly on the integration tests, but all the micro optimizations done since a year improved them to the point the bottleneck is shifting to the documentation examples and images. Also, once your Docker CI is ready, I hope we can finally get the build time below 5 minutes again. But that's impossible if my documentation tests/examples/images take 3 minutes to run.

tests/examples/CMakeLists.txt
```diff
@@ -40,28 +41,22 @@ else()
...
+${TOP_SOURCE_DIR}/tests/examples/shims/?\\;\
+${LUA_PATH_}")
 
-# Done in 3 variables to avoid CMake from implicitly converting into a list.
-set(ENV{LUA_PATH} "${LUA_PATH3_}${LUA_PATH2_}${LUA_PATH_}")
-
-# Unset environment variables that would get in the way of evaluating LUA_PATH.
-unset(ENV{LUA_PATH_5_1})
-unset(ENV{LUA_PATH_5_2})
-unset(ENV{LUA_PATH_5_3})
```
psychon: I don't think we want to drop these unsets. If one of these variables is set, then `$LUA_PATH` is ignored.

tests/examples/CMakeLists.txt
```diff
@@ -216,17 +214,18 @@ function(run_test test_path namespace escaped_content)
...
-        )
+        set(OUTPUT_IMAGE_PATH "${IMAGE_PATH}.png")
+    else()
+        set(OUTPUT_IMAGE_PATH "")
     endif()
 
+    escape_string(
+        "![Usage example](../images/AUTOGEN${namespace}_${TEST_FILE_NAME}.svg)\n"
+        "${TEST_DOC_CONTENT}" TEST_DOC_CONTENT ""
+    )
```
psychon: The hardcoded `.svg` seems wrong and doesn't this now also add `Usage examples` images to tests that do not actually produce an image?

psychon:
I'll defer to Emmanuel here. Besides the things that I commented, I would look at the output of `git diff -w` since I am not completely sure what actually changed besides indentation at the end. Oh and I didn't look at the individual commits since that's even more confusing.

tests/examples/CMakeLists.txt
blueyed: Moved them into the `env` call (using `-u`).

blueyed:
@psychon 
Squashed, and here is `git show -w`: http://sprunge.us/JOgW

blueyed:
Continued in https://github.com/awesomeWM/awesome/pull/1962.

PR #1914, opened 2017-07-10
Treat max client as single client for gaps

majutsushi:
When gap_single_client is set to 'false', no gap is used when only one client is shown in a tag. Since the max layout only ever shows one client at a time it makes sense to apply this setting for this layout regardless of the actual number of clients in that tag.

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -164,8 +164,15 @@ function layout.parameters(t, screen)
         gap_single_client = t.gap_single_client
     end
 
-    local min_clients       = gap_single_client and 1 or 2
-    local useless_gap       = t and (#clients >= min_clients and t.gap or 0) or 0
+    local useless_gap = 0
+    if t then
+        local skip_gap = layout.get(screen).skip_gap or function(nclients)
+            return nclients < 2
+        end
+        if gap_single_client or not skip_gap(#clients, t) then
+            useless_gap = t.gap
+        end
+    end
 
     p.workarea = screen:get_bounding_geometry {
         honor_padding  = true,
--- a/lib/awful/layout/suit/corner.lua
+++ b/lib/awful/layout/suit/corner.lua
@@ -163,13 +163,18 @@ end
 local corner = {}
 corner.row_privileged = false
 
+function corner.skip_gap(nclients, t)
+    return nclients == 1 and t.master_fill_policy == "expand"
+end
+
 --- Corner layout.
 -- Display master client in a corner of the screen, and slaves in one
 -- column and one row around the master.
 -- @clientlayout awful.layout.suit.corner.nw
 corner.nw = {
         name = "cornernw",
-        arrange = function (p) return do_corner(p, "NW") end
+        arrange = function (p) return do_corner(p, "NW") end,
+        skip_gap = corner.skip_gap
     }
 
 --- Corner layout.
@@ -178,7 +183,8 @@ corner.nw = {
 -- @clientlayout awful.layout.suit.corner.ne
 corner.ne = {
         name = "cornerne",
-        arrange = function (p) return do_corner(p, "NE") end
+        arrange = function (p) return do_corner(p, "NE") end,
+        skip_gap = corner.skip_gap
     }
 
 --- Corner layout.
@@ -187,7 +193,8 @@ corner.ne = {
 -- @clientlayout awful.layout.suit.corner.sw
 corner.sw = {
         name = "cornersw",
-        arrange = function (p) return do_corner(p, "SW") end
+        arrange = function (p) return do_corner(p, "SW") end,
+        skip_gap = corner.skip_gap
     }
 
 --- Corner layout.
@@ -196,7 +203,8 @@ corner.sw = {
 -- @clientlayout awful.layout.suit.corner.se
 corner.se = {
         name = "cornerse",
-        arrange = function (p) return do_corner(p, "SE") end
+        arrange = function (p) return do_corner(p, "SE") end,
+        skip_gap = corner.skip_gap
     }
 
 return corner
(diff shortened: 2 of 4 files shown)


actionless:
this looks fine however i think it would be also better to take in account `master_fill_policy`.
for example, in my config there is a part of logic like this:
```lua
  if get_num_tiled(t) == 1 and t.master_fill_policy == "expand" then
    return 0
  end
```

Elv13:
The resulting change makes sense. However the implementation increases the coupling between the default suits and `awful.layout`. I would prefer the option to be added to `awful.layout.suit.max`. The `magnifier` layout already has a `magnifier.need_focus_update`, so this kind of `layout.suit` configuration variables would not be a first. It doesn't need to be official or documented. It could also be a function to implement @actionless fill policy logic.

actionless:
@Elv13 
in my config i am monkey-patching get_gap:
```lua
awful.tag.object.get_gap = function(t)
  t = t or awful.screen.focused().selected_tag
  if get_num_tiled(t) == 1 and t.master_fill_policy == "expand" then
    return 0
  end
  return awful.tag.getproperty(t, "useless_gap") or beautiful.useless_gap or 0
end
```

but i don't have any good idea how to bring it upstream (there are several places there this code could be injected)

majutsushi:
I've changed the logic to use a function defined in layouts (if it exists) to determine whether a gap should be applied, and modified the two layouts that use the `master_fill_policy` property to mimic the logic @actionless proposed.

Alternatively the layouts could just have a boolean property to determine that and people could use monkey-patching just like @actionless to modify the behaviour more.

blueyed:
There are some luacheck issues:

    lib/awful/layout/suit/max.lua:59:23: unused argument nclients
    lib/awful/layout/suit/max.lua:59:33: unused argument t

majutsushi:
The arguments are unused because they are not needed in that layout. I have amended the commit to make luacheck ignore the unused args for that function.

psychon:
Fine by me, but I feel like others have taken over maintenance of the layout code and let them take a careful look and decide. Thanks to everyone for your work on awesome!

lib/awful/layout/init.lua
```diff
@@ -164,8 +164,10 @@ function layout.parameters(t, screen)
         gap_single_client = t.gap_single_client
     end
 
-    local min_clients       = gap_single_client and 1 or 2
-    local useless_gap       = t and (#clients >= min_clients and t.gap or 0) or 0
+    local useless_gap = 0
+    if (t and (gap_single_client or (#clients >= 2 and layout.get(screen).name ~= "max"))) then
```
Anarky: It would make sense to also match the `fullscreen` layout, I found this **really** ugly way (it seems lua doesn't support general alternation):
```(#clients >= 2 and not string.find(awful.layout.get(client.screen).name, "[mf][au][xl]"))```

**Comment by Anarky** [comment]
I just left a comment in the change, it would make sense to also match the `fullscreen` layout.

**Comment by Elv13** [comment]
> I just left a comment in the change, it would make sense to also match the fullscreen layout.

You seem to have commented to an older version of the patchset. The newer one delegate the decision to the layout object. Setting the skip_gap for the fullscreen layout can be done when this is merged. I will do it tonight if nobody has anything else to add until then.

File: lib/awful/layout/suit/tile.lua [hunk 2]
```diff
@@ -317,6 +324,9 @@ end
 function tile.left.mouse_resize_handler(c, corner, x, y)
     return mouse_resize_handler(c, corner, x, y, "left")
 end
+function tile.left.skip_gap(nclients, t)
+    return tile.skip_gap(nclients, t)
+end
```
psychon: `tile.left.skip_gap = tile.skip_gap` seems simpler to me (same for similar places here)

psychon: However, this is not really important and the current version is "basically okay" as well

lib/awful/layout/suit/corner.lua
```diff
@@ -196,7 +203,8 @@ corner.sw = {
 -- @clientlayout awful.layout.suit.corner.se
 corner.se = {
         name = "cornerse",
-        arrange = function (p) return do_corner(p, "SE") end
+        arrange = function (p) return do_corner(p, "SE") end,
+        skip_gap = function (nclients, t) return corner.skip_gap(nclients, t) end
```
psychon: Here, `skip_gap = corner.skip_gap` really might be simpler

lib/awful/layout/init.lua
psychon: It would make sense to squash the two commits

lib/awful/layout/init.lua
```diff
@@ -164,8 +164,14 @@ function layout.parameters(t, screen)
...
 
-    local min_clients       = gap_single_client and 1 or 2
-    local useless_gap       = t and (#clients >= min_clients and t.gap or 0) or 0
+    local useless_gap = 0
+    if (t) then
+        local l = layout.get(screen)
+        if (gap_single_client or not l.skip_gap and #clients >= 2
+                or l.skip_gap and not l.skip_gap(#clients, t)) then
+            useless_gap = t.gap
+        end
```
psychon: Since we are proposing not-all-that-important code changes:
```lua
local useless_gap = 0
if t then
  local skip_gap = layout.get(screen).skip_gap or function(nclients)
    return nclients >= 2
  end 
  if gap_single_client or not skip_gap(#clients, t) then
    useless_gap = t.gap
  end
end
```
I'll leave it for others to decide which version is simpler.

majutsushi:
I forgot that you can do that simple function assignment in Lua, so I've changed things as suggested to make them simpler. I've also squashed the commits since that's what I've been wanting to do anyway.

blueyed:
Thank you!

Issue #1496, opened 2017-01-31
Slow menubar startup


Architektor:
Hey there I'm using fresh awesome v4 and I'm curious why the prompt (or whatever the mod+p thing name is) is starting awfully slow. I love the feature but when i press mod+p for the first time after reboot my whole machine is paralyzed for about ten seconds. After the prompt appears, it's quite laggy and every symbol i type renders in the text box with a huge pain. Then the whole thing caches to ram (i dunno actually but feels like) and the magic prompt is working as expected, with a speed of an eagle. Do someone know what causes this slow behaviour? I'm bad at lua/c code, but i'm starting my investigation, cause now it feels like this monster is grepping my whole filesystem on the first launch

Elv13:
This is because it does ~10k file system requests looking for icons of the right size and `.desktop`. I guess you have a 4200 RPM hard drive. I have a version that does the same thing in < 500ms, but it depends on GTK and its cache, so it cannot be used in the "standard version" of the menubar.

Architektor:
@Elv13 I still have the old laptop without SSD. So i'll try to somehow turn off icons at all in my setup. Ty for response.

yejingchen:
Hello, I also experienced this issue since awesome 4.0. My system is Arch Linux, `awesome --version` gives:

```
awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1
```

I tried running `journalctl -f` and wait until there is no log appearing, then I press Super+P, and the following lines appeared:

```
E: Error opening directory '/usr/local/share/applications': No such file or directory
W: awesome: a_glib_poll:389: Last main loop iteration took 3.056006 seconds! Increasing limit for this warning to that value.
```

I wonder if there exists another way to improve this without changing to SSD?

Elv13:
Of course. My config has a version of the menu that's as fast as Gnome2 one. In theory, it's code could be merged into the menubar and solve the problem. However it's a lot of work and I wont do it as less and less people have the issue and it adds a new dependency (GTK) on Awesome. *But*, if you are willing to do it, then I can see this being merged if the GTK code path is optional

https://github.com/Elv13/awesome-configs/blob/master/utils/fd_async.lua
https://github.com/Elv13/awesome-configs/blob/master/customMenu/appmenu.lua

[...]

Issue #908, opened 2016-05-14
XDG icon specs implementation woes: Introduction to the missing icons


Jajauma:
Judging by those parts of the Lua code which I have already got myself acquainted with, you maintain at least 3 (three!) competing facilities to map the icon names (possibly along with its screen size) into the file names on a disk. Let's introduce our players briefly:
- `icon_theme.find_icon_path()` of [lib/menubar/icon_theme.lua](https://github.com/awesomeWM/awesome/blob/master/lib/menubar/icon_theme.lua#L225). Used solely to extract 10 (ten) paths to the categories icons (Office, Development, etc) in the menubar.
- `utils.lookup_icon()` of [lib/menubar/utils.lua](https://github.com/awesomeWM/awesome/blob/master/lib/menubar/utils.lua#L158). Used by the menubar implementation to extract the vast majority of application icons set in their respective *.desktop files.
- `util.geticonpath()` of [lib/awful/util.lua](https://github.com/awesomeWM/awesome/blob/master/lib/awful/util.lua#L200). Used by the naughty module to extract the full icon path using the icon name being received from a notification client over DBus.

They all three do work differently. Even worse they all three fail differently. This results in some application icons missing from the menubar and the icon missing from the notifications popup sent over DBus. I've tried switching them back and forth (using the menubar as a guinea pig) and from the three of those `icon_theme.find_icon_path()` being the most sophisticated seems to be a noticeable favorite as well. It was able to find most of the icons (with or without icon theme set in beautiful) compared to `utils.lookup_icon()`,  albeit there are failures too (e.g. "nvidia-settings.png" - note the extension - could not be found while it is happily sitting under /usr/share/pixmaps).

Given that bestiary the real question reveals: is it worth to propose fixes for individual issues or is there already some undergoing black op plan of merging them/dropping them/adding 4th to the party/etc? From my personal point of view adding all the corner cases handling to the `icon_theme.find_icon_path()` and dropping the rest of a company would be the preferable and cleanest path but that might render a new problem - it is inherently _slow_. Recent making of the menubar asynchronous can mititage it for some degree though.

Just in case you prefer leaving it all as it is I can open an issue for each and every missing icon I have, possibly along with a "fix".

psychon:
I tried to stay out of this as much as possible...
I knew that the menubar had two similar implementations, but I wasn't aware that what naughty does is based on the same standard and somehow belongs to the rest.

[...]

PR #1915, opened 2017-07-10
More icons

Oblomov:
Aside from the first sidecar commit that fixes a couple of typos, the rest of the set extend both the utility function that assembles icon paths, and the default search path and extension list, to ensure that icons found in common but less standard places can still be found.


lib/naughty/core.lua
```diff
@@ -66,11 +66,19 @@ notifications, e.g.
...
     notify_callback = nil,
 }
 
+if beautiful.icon_theme then
+    local theme = beautiful.icon_theme
+    local theme_dir = "/usr/share/icons/" .. theme .. "/"
+    table.insert(naughty.config.icon_dirs, 1, theme_dir)
+    table.insert(naughty.config.icon_dirs, 1, theme_dir .. "base/")
+    table.insert(naughty.config.icon_dirs, 1, theme_dir .. "symbolic/")
+end
```
psychon: As Travis points out, this does not work: https://travis-ci.org/awesomeWM/awesome/jobs/252054577
```
error: lib/beautiful/init.lua:33: May not access theme before beautiful.init()
```
This code runs during `require()` and `beautiful.init` is only called later (look at the default config). Thus, `beautiful.icon_theme` will always be `nil` when this code runs.

Oblomov: Duh, my bad.

I was wondering about that, actually, but I was stupid enough to only test on my configuration. I'm guessing the best approach here is to have `beautiful.init` update `naughty.config`.

Oblomov:
Thanks for pointing me to #908, I had completely missed it. In some sense this contribution of mine actually makes things worse, because it increases the number of filesystem stats() when looking for icons.

I'll look into at the very least building some sort of cache for icon lookup.

psychon:
> I'll look into at the very least building some sort of cache for icon lookup.

I think that this is unrelated to this PR and should best be done separately. I am a big fan of small PRs. :-)

Oblomov:
That's even better (for me). But I'll really try and find the time to make something about that issue too ;-)

Elv13:
See for a possible implementation:

https://github.com/awesomeWM/awesome/issues/1496 (comments).

It needs work before a version of the cached loading can be merged. As for this PR, I also think it makes it too slow for hard drive users. It is already near unusable. Having the system freeze for 3+ seconds when a notification arrives might even cause the notification to timeout before it shows up. It has nothing to do about your PR and I agree it does fix things for a majority of users. However as you can see in both issues linked above, the solution isn't that simple if we want it to work for everybody. Could you split the PR in 2 so we can merge the trivial commits and leave the problematic ones? That way if an user is annoyed of the issue she/he can merge your pull request into their systems until someone takes the time to build a proper solution?

Oblomov:
@Elv13 thank you for your comment.

Aside from the typo fix (which maybe can be cherry-picked without doing a full merge, or should I do  a PR just for that?), I'm not sure which could be considered 'trivial' patches, in terms of disk usage.

For example, the simplest ones are probably the naughty config patches, but even that alone more than triplicates the workload on icon search, since we get three base paths instead of one, and four extensions instead of two (I would actually argue that .gif isn't needed, but I don't feel like removing it). I'm guessing this would already be enough to kill performance on rotating media, without some kind of optimization (like, checking which base directories actually exist).

PR #1916, opened 2017-07-13
Tests for awful.prompt.

morethanoneanimal:


morethanoneanimal:
`insulate` is described [here](https://olivinelabs.com/busted/#defining-tests). It insulates environment (global packages, loaded modules) so that they don't interfere in one test file. 
I used that because I thought I will use different environments, but I haven't. Every test file runs in a separate `insulate` by default. so I should probably skip it here.

blueyed:
Nice learning about `insulate` after all.
I think it is OK to keep it.

Thanks for these tests!

PR #1920, opened 2017-07-15
cmake: check (and cache) check of required modules

blueyed:
This makes the CMake run much faster in general, and does not re-check
the dependencies every time again (since it gets cached).

There are several reports where the reference to the missing module was
not provided, and they all appear to have happened with CMake 3.6.1.

A missing "foobar" module gets reported there, but if you remove
/usr/lib/pkgconfig/libstartup-notification-1.0.pc, this does not get
reported:

    -- Checking for modules '…'
    --
    CMake Error at /tmp/cmake/share/cmake-3.6/Modules/FindPkgConfig.cmake:424 (message):
      A required package was not found

This apparently affects modules that have version requirements. After
removing that from "libstartup-notification-1.0>=0.10", it gets reported
also: "No package 'libstartup-notification-1.0' found".

The workaround is to require the package without version information
before the requirement with the version information.

This reverts https://github.com/awesomeWM/awesome/pull/1253.

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -127,6 +127,7 @@ pkg_check_modules(AWESOME_COMMON_REQUIRED REQUIRED
     xcb>=1.6)
 
 set(AWESOME_DEPENDENCIES
+    glib-2.0
     glib-2.0>=2.40
     gdk-pixbuf-2.0
     cairo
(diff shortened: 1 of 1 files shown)


blueyed:
I am going to bisect this in CMake, but with the workaround it seems to be good already.

blueyed:
I guess I have to rm the CMake cache to get the coverage back up?!

blueyed:
I've thought about this part, but there should be no cache?!
https://github.com/awesomeWM/awesome/blob/6f5ea507d960f6512ea30654bec1f53fe9f70e74/.travis.yml#L190-L193

blueyed:
Coverage also dropped in https://github.com/awesomeWM/awesome/pull/1921 - which seems to be unrelated?!

blueyed:
It was fixed in CMake v3.7.0-618-gcbccebbac:

    FindPkgConfig: Fix missing error text when library version is specified
    
    Calls like `pkg_check_modules(somelibrary>=3.22)` that specify a version
    requirement should still display an informative error when the package
    is not found.  Fix our logic accordingly.

psychon:
> Coverage also dropped in #1921 - which seems to be unrelated?!

Looks like somewhere something "only sometimes" uses a shaped window. I think that none of our tests are using window shapes, or do I miss something? Does Gtk try to do rounded corners or something?

PR #1922, opened 2017-07-15
Debugging codecov coverage drop

blueyed:


blueyed:
@psychon 
is this expected/OK?

psychon:
The most impacted file is `lib/awful/client/shape.lua`, so this is again some shape-thing. No idea why that file ever had coverage since (as far as I know) we don't have any shaped windows in tests. A "fire and forget" test would be to just spawn `xeyes` in some test and wait a bit to make sure that it is visible and awesome handled its shape, but that would not really actually test much...

I tend to just ignore random coverage changes.

PR #1924, opened 2017-07-16
Packaging.cmake: cache CPACK_GENERATOR

blueyed:


akdor1154:
Looks nicer than my code did haha.

psychon:
So, the idea here is to make `CPACK_GENERATOR` cached so that people can override it?
Does `append` on cached variables work as intended? I mean, the new value really ends up in the cache?

blueyed:
@psychon 
> So, the idea here is to make `CPACK_GENERATOR` cached so that people can override it?

No, it is more about using the cached version in the first place - we do not need to recheck this all the time.

> Does `append` on cached variables work as intended? I mean, the new value really ends up in the cache?

The `list(APPEND CPACK_GENERATOR {DEB,RPM})` is what defines the (cached) variable in the first place.

Do you mean that the `set(CPACK_GENERATOR CACHE STRING "Include CPack if non-empty (DEB and/or RPM).")` should happen afterwards?

@akdor1154 
Thanks for the review!
Would be nice if you could test if it works as expected for you.. (I've only tested the case where both `dpkg` and `rpm` are not available ;))

blueyed:
Pushed a fixup.

As for me it is not required to have `dpkg` installed for CPACK_GENERATOR=DEB?!
Not sure if the produced `.deb` is valid though.

Using RPM did not work though:
```
CMake Error at /usr/share/cmake-3.8/Modules/CPackRPM.cmake:1611 (message):
  RPM package requires rpmbuild executable
  /usr/share/cmake-3.8/Modules/CPackRPM.cmake:2690 (cpack_rpm_generate_package)
```

psychon:
> As for me it is not required to have dpkg installed for CPACK_GENERATOR=DEB?!

Yup, we just use that as poor man's "detect debian".
Feel free to squash-merge this.

blueyed:
Looks like codecov is rather confused lately?!
There should not be coverage changes in here..

blueyed:
Seems like a debase would fix it probably.

PR #1925, opened 2017-07-16
awesomeConfig.cmake: D-Bus: use cache from pkg_check_modules

blueyed:


psychon:
Uhm, what? Shouldn't `pkg_check_modules` do that check itself?

blueyed:
It seems to do so only when used with `REQUIRED` it seemed to me.

psychon:
Looking at `FindPkgConfig.cmake`, the condition in `pkg_check_modules` is
```
# check cached value                                                                                                                                                                   
if (NOT DEFINED __pkg_config_checked_${_prefix} OR __pkg_config_checked_${_prefix} LESS ${PKG_CONFIG_VERSION} OR NOT ${_prefix}_FOUND OR NOT ${__pkg_config_arguments_${_prefix}}" STREQUAL "${_module0};${ARGN}")
```
That is terribly unreadable, but I see:
Do the check if any of the following is true
- The variable `__pkg_config_checked_${_prefix}` is undefined (this variable is CACHEd)
- The version of `pkg-config` got higher
- The variable `DBUS_FOUND` is not set
- The module is called with different arguments this time

Nothing in there looks at `REQUIRED`, so I'm not sure what is going on. This commit should be unnecessary, shouldn't it?

blueyed:
Looking at `strace` it calls `pkg-config`:

> execve("/usr/bin/pkg-config", ["/usr/bin/pkg-config", "--exists", "--print-errors", "--short-errors", "dbus-1"]

The condition that fails is `OR NOT "${__pkg_config_arguments_${_prefix}}" STREQUAL "${_module0};${ARGN}"`, where the args are `dbus-1` and `dbus-1;`.

If you add `REQUIRED` (or `QUIET`) it will end up being `REQUIRED;dbus-1` in both cases.

This appears to be fixed in CMake v3.9.0-rc5-244-g3cf267cfb though.

Let's not work around it here though.

PR #1926, opened 2017-07-17
tests/examples/CMakeLists.txt: Add a project() call

psychon:
project() automatically sets the PROJECT_NAME variable. Thus, when this
CMake code is run due to being included by awesome, this commit has no
effect.

When this CMake code is run on its own, CMake will no longer check for a
C and C++ compiler.

Reference: https://github.com/awesomeWM/awesome/pull/1909#issuecomment-315646440


blueyed:
Thanks!

PR #1932, opened 2017-07-19
Fix awful.widget.calendar_popup documentation

mqus:
There were multiple things which stood out to me, as I was trying to setup a simple calendar popup for the default textclock in my rc.lua:
1. The functions in https://awesomewm.org/doc/api/classes/awful.widget.calendar_popup.html#Functions were not named correctly (calendar vs calendar_popup) which (naturally) produced an error when used as awful.widget.calendar.month() .
2. The example for calendar_popup.month() was obviously missing a line (where does the month_calendar come from?) . Resolved by copying the line from the attach-example
3. The same examples also only refer to the function as calendar.month() which should be changed to awful.widget.calendar_popup.month()

### DIFF
--- a/lib/awful/widget/calendar_popup.lua
+++ b/lib/awful/widget/calendar_popup.lua
@@ -241,7 +241,7 @@ end
 --- Attach the calendar to a widget to display at a specific position.
 --
 --    local mytextclock = wibox.widget.textclock()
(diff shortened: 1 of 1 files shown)


Elv13:
Thanks!

PR #1934, opened 2017-07-20
Travis: move usage of xresources theme to tests/run.sh

blueyed:
This makes the coverage more consistent, and is done like that in the
docker branch already (where a read-only filesystem is used).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
   matrix:
     - LUA=5.2 LUANAME=lua5.2 DO_COVERAGE=coveralls
     # Lua 5.3 isn't available in Ubuntu Trusty, so some magic below installs it.
-    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov AWESOME_THEME=xresources
+    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov
     # luajit: installed from source.
     - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.5 TEST_PREV_COMMITS=1
     # Note: luarocks does not work with Lua 5.0.
@@ -148,10 +148,6 @@ install:
     }
 script:
   - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -DLUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -DSTRICT_TESTS=true"
-  - |
-    if [ ! -z "$AWESOME_THEME" ]; then
-      sed awesomerc.lua -i -e "s/default\/theme/$AWESOME_THEME\/theme/g"
-    fi
   - |
     if [ "$DO_COVERAGE" = "codecov" ]; then
       export CXXFLAGS="-fprofile-arcs -ftest-coverage"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -12,7 +12,7 @@ env:
   matrix:
     - LUA=5.2 LUANAME=lua5.2 DO_COVERAGE=coveralls
     # Lua 5.3 isn't available in Ubuntu Trusty, so some magic below installs it.
-    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov AWESOME_THEME=xresources
```
actionless: i would leave this and the `if` condition (also later when i will prepare a `gtk` theme it will be good to set it for other travis profile) but i'm fine with with moving sed to `run.sh`

tests/run.sh
```diff
@@ -174,6 +174,9 @@ if [ -n "$DO_COVERAGE" ] && [ "$DO_COVERAGE" != 0 ]; then
     sed "1 s~^~require('luacov.runner')('$source_dir/.luacov'); \0~" \
         "$RC_FILE" > "$tmp_files/awesomerc.lua"
     RC_FILE=$tmp_files/awesomerc.lua
+
+    # Use xresources theme for more coverage.
+    sed -i 's~default/theme~xresources/theme~' "$RC_FILE"
```
psychon: Doesn't this cause less coverage since now only the xresources theme is measured?
Also, if for some reason some error only appears under coverage, things might get confusing ("why does it break? it worked just a moment before without coverage?!?")

blueyed: Yes, true.

But previously we had differrent things covered anyway.

So I think we should actually run it twice (again), but report to both coverage providers, or rather only run specific tests/files twice.

actionless: what do you think about a test which will set each available theme one by one and just reloading awesome and making sure nothing paniced?

psychon: That doesn't really fit into the current testing infrastructure, but besides that this is fine with me

blueyed: It could be an integration test, looping over the themes, `sed`ing the config, and reload awesome?!

Is there anything easy to iterate of the them dirs?
Could we have something a bit more generic than https://github.com/awesomeWM/awesome/blob/140481a3e566fd2b7c99191fd9e0da9daa3d26f4/docs/06-appearance.md.lua#L33 in `gears`?

actionless: i am more curious, if such all-theme-iteration test can be easily integrated to the current testsuite so its execution will be traced by coverage measurement tool


so mb instead of sed-ing just do everything on lua side:
for theme in all-themes
1) flush beautiful theme table
2)  beautiful.init(theme)
nothing raised exceptions -- we are good
3) awful.util.restart()

or awesome can't be reloaded during testsuite execution? if so i still think what just flushing beautiful theme table then could be enough

psychon: Nope, you cannot restart awesome between tests. After the restart, awesome would not know that it is supposed to continue executing some text. Also, changing the theme at runtime is not a good idea since nothing will use the changed theme values for anything.

If you just want to test that the themes themselves contain valid Lua code, you can `dofile()` them.

actionless: >> dofile()

will it be added to the coverage?
and does it makes any sense at all?

psychon: > will it be added to the coverage?

Yup, coverage tracing uses debug hooks and those also run from `dofile()`'d things. I'm just not sure if there could be problems because of relative vs absolute paths.

> and does it makes any sense at all?

I'm not sure. Perhaps not...?

blueyed:
Continued in https://github.com/awesomeWM/awesome/pull/1967.

Issue #1933, opened 2017-07-19
NPE in keyboardlayout


boris-petrov:
[This PR](https://github.com/awesomeWM/awesome/pull/1771) breaks things for me. I'm getting:

```
/usr/share/awesome/lib/awful/widget/keyboardlayout.lua:123: attempt to concatenate a nil value (field '?')
```

Reverting the commit fixes the issue.

Output of `awesome --version`:

```
%  awesome --version
awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1
```

psychon:
Could you e.g. add `require("gears.debug").dump(self._layout)` or `require("naughty").notify{ text = require("gears.debug").dump_return(self._layout)` before line 123 and figure out what layouts are detected for you?
Put differently: What exactly is going on here?

mbuelte:
Hi, I've got the same problem after an upgrade, here is the output of 'require("gears.debug").dump(self._layout)' above line 123:

```
table: 0x585afeaa0c0
  1 : de (string)
```

boris-petrov:
@psychon - sorry for the delay. The output:

```
table: 0x55e959041ec0
  0 : En (string)
  1 : Bg (string)
```

Note, however, that I've changed the original names to `En` and `Bg`, but I don't think this matters.

PR #1941, opened 2017-07-25
Make keyboardlayout widget more robust

psychon:
I am not sure what exactly goes on in the below bug report, so this will
just paper over the problem. I still think that this patch is a good
idea even when it is not a proper fix, since keyboard layouts are
complicated and so this code should be robust and hard to break.

Fixes: https://github.com/awesomeWM/awesome/issues/1933

### DIFF
--- a/lib/awful/widget/keyboardlayout.lua
+++ b/lib/awful/widget/keyboardlayout.lua
@@ -115,12 +115,15 @@ keyboardlayout.xkeyboard_country_code = {
 
 -- Callback for updating current layout.
 local function update_status (self)
-    self._current = awesome.xkb_get_layout_group();
+    self._current = awesome.xkb_get_layout_group()
     local text = ""
-    if (#self._layout > 0) then
+    if #self._layout > 0 then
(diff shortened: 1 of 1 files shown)


kacejot:
Hi, guys!

Sorry about offtop, but I don't know if I need to create the new issue.
How I can change the font of the keyboard layout widget?

psychon:
Uhm... Poor man's hack would be to do:
```lua
function keyboardlayout_with_font(font)
    local theme = require("beautiful").get()
    local old_font, theme.font = theme.font, font
    local result = awful.widget.keyboardlayout()
    theme.font = old_font
    return result
end
```
And yes, I guess it would make sense to make this easier...

> Sorry about offtop, but I don't know if I need to create the new issue.

Yes, you should just create a new issue.

kacejot:
Hi, guys!

Sorry about offtop, but I don't know if I need to create the new issue.
How I can change the font of the keyboard layout widget?

actionless:
that was a nice one, never considered monkey-patching beautiful for styling purposes :smiley: :+1:

PR #1944, opened 2017-07-26
Travis: exit with 1 when check-unit-coverage fails

blueyed:
Same as done in 1cdd7ef8 already, but missed this one.


blueyed:
This should fix https://travis-ci.org/awesomeWM/awesome/jobs/257577996.

PR #1948, opened 2017-07-28
WIP: Replace awful.util.geticonpath with menugen.icon_theme function

GinoBombino:
A dropin replacement for the slower geticonpath() function

- Cons: API needs to be changed
- TODO: icon_theme_name needs to be defined and passed into awful.util
into some reasonable way, also extensions are no longer needed
- Pros: much faster, misses less icons

Another reference to #908

Questions:
Maybe menubar.icon_theme should be moved out of menubar and into awful?
If its moved into awful.util then should geticonpath function be deprecated and a new function created or make a replacement that takes in different parameters?

### DIFF
--- a/lib/awful/util.lua
+++ b/lib/awful/util.lua
@@ -11,7 +11,6 @@ local os = os
 local assert = assert
 local load = loadstring or load -- luacheck: globals loadstring (compatibility with Lua 5.1)
 local loadfile = loadfile
-local pairs = pairs
 local type = type
 local gtable = require("gears.table")
 local string = string
@@ -26,6 +25,7 @@ local capi =
 }
 local gdebug = require("gears.debug")
 local gmath = require("gears.math")
+local icon_theme = require("menubar.icon_theme")
 
 local util = {}
 util.table = {}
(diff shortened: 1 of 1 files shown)


Elv13:
> Maybe menubar.icon_theme should be moved out of menubar and into awful?

Probably into `gears`, otherwise you will still have module inter-dependency issues.

> then should geticonpath function be deprecated

Yes, it should be deprecated, but everything has to keep working as good (or better) than the current state. Hard API breaks are not allowed during the 4.x lifecycle. We have some increasingly good APIs to handle deprecation itself, see `gears.debug`. The deprecation notice should kick in for 5.0 (it is *allowed* to deprecate functions in the documentation, but it not allowed to add new deprecation notices in users logs (until the next API breaking release, 5.0).

As for all the work on icons itself, as stated to all the other discussions about it, my vote it still to use GDK/GTK own APIs for it. It can be accessed using LGI and is much faster/better/compatible/maintained/async/cached/supported than anything we can come up with. It is also the reference implementation for the spec itself.

Elv13:
@GinoBombino ping?

Both of your open pull requests are waiting for comments/updates.

GinoBombino:
Sorry for the delay guys, so I'm trying to work on using lgi to render these icons, then I got distracted by including functions for not just icons but also XDG menu related functions..

So I'll cancel this pull request and put in a new one for altering gears to add 
 Gtk.gtk_icon_theme_load_icon() function as a way to get a pixbuf for a particular theme

Any comments would be great,on what you guys would like to see here.
https://github.com/GinoBombino/awesome/tree/xdg-icon-lgi


Thanks!

PR #1950, opened 2017-07-29
tags: specify layout in add_tag doc snippet

j605:
Adding a tag without layout causes errors while using the tag. For
example resizing with mouse causes indexing a null error.

### DIFF
--- a/objects/tag.c
+++ b/objects/tag.c
@@ -111,7 +111,9 @@
  * Create a new tag at the end of the list
  *
  *    local function add_tag()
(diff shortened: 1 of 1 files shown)


psychon:
Is https://github.com/awesomeWM/awesome/pull/1952 a better solution to this problem?

j605:
Your change makes more sense.

blueyed:
Closing this one in favor of https://github.com/awesomeWM/awesome/pull/1952 then.

Elv13:
@blueyed The reason the change was necessary is "fixed", however it's still a better example (more useful) than the current one. I suggest we merge this anyway. What do you think?

Issue #1862, opened 2017-06-22
No way to restrict the icon size of dbus notification


bew:
Output of `awesome --version`:

```
awesome v4.1-184-gba78882c-dirty (Technologic)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1
```

**How to reproduce the issue:**

Execute the following shell commands:
```sh
$ wget http://via.placeholder.com/800x600 -O big-image.png
$ notify-send hello world -i $PWD/big-image.png
```

**Actual result:**

You should see a notification with a big icon (800x600 px)

**Expected result:**

I would expect to have `dbus_notification_icon_max_(width|height)` in beautiful to control the icon max size (and maybe another one to specify if I want to keep aspect ratio or not).

nickdiego:
+1

bew:
#1951 fixed my main issue, can I close this ? or do you want to keep it open for later reference during the naughty rewrite (initiated by #1990) ?

Elv13:
Sure. Anyway, I currently don't have enough time to merge all patches from the rewrite. I don't have an ETA, but I will work meet/work with @blueyed in a month or so. Hopefully we can make progress on this project (among other long standing pending work stalled in the pipeline such as the new Docker CI and maybe even the unification of the layout engines).

terinjokes:
Setting `beautiful.notification_icon_size` doesn't seem to effect the size at all. Can this be reopened?

psychon:
@terinjokes What's your awesome version? If it is 4.2 or older, then you do not yet have the fix for this issue. Put differently: This was closed in September while the latest release of awesome was in June.

terinjokes:
@psychon Yep. I noticed that after I commented and started looking at the code on the latest tag. I edited the above comment for a workaround.

actionless:
https://awesomewm.org/apidoc/libraries/naughty.html#beautiful.notification_icon_size

actionless:
and also see `https://awesomewm.org/apidoc/libraries/naughty.html#beautiful.notification_max_height` for restricting geometry of notification not regarding the icon


from your version it seems what you're running very old build from git master, you need to rebuild it to have all those nice notification options

qwelyt:
@actionless Waiting for the next release.

olekthunder:
[...]

PR #1951, opened 2017-07-29
Add notification_icon_size as a beautiful variable

tomoyat1:
This PR adds a new beautiful variable `notification_icon_size` that allows users to specify icon sizes in notifications.

This was inspired by #1862, but I thought instead of specifying a maximum size, specifing an absolute size would lead to more consistent notifications.

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -189,6 +189,10 @@ naughty.notificationClosedReason = {
 -- @beautiful beautiful.notification_height
 -- @tparam int notification_height
 
+--- Notifications icon size.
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!

KhaimovMR:
Didn't find this patch in the newest awesome wm version (4.2).

Applied it myself. Thanks for the effort, guys!

psychon:
@KhaimovMR The newest version is v4.3 and was released in January. It has this patch. v4.2 is from Juli 15th 2017 and thus older than this PR by a few days.

PR #1953, opened 2017-07-30
Specifies the wibox dependency on the menubar utils

Crazylemon64:
When including the menubar utils, you would drag in the wibox library, which would proceed to drag in all sorts of other libraries that expected awesome to be fully running. By doing this, we can use the utils file in tests without needing to monkey patch anything.

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -17,7 +17,7 @@ local theme = require("beautiful")
 local lgi = require("lgi")
 local gio = lgi.Gio
 local glib = lgi.GLib
-local wibox = require("wibox")
+local w_textbox = require("wibox.widget.textbox")
 local gdebug = require("gears.debug")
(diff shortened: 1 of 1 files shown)


psychon:
Sigh:
```
Checking lib/menubar/utils.lua                    1 warning
    lib/menubar/utils.lua:334:38: shadowing upvalue textbox on line 20
```
I guess the parameter to `compute_textbox_width` should be renamed from `textbox` to `tb`?

Crazylemon64:
Oh, I took a different strategy, and prefixed the textbox widget import to be `w_textbox` - should I change the textbox var there to `tb` instead?

psychon:
> Oh, I took a different strategy, and prefixed the textbox widget import to be w_textbox - should I change the textbox var there to tb instead?

Nah, either version is fine.

Elv13:
thanks

Issue #1923, opened 2017-07-16
Command starting with "true" breaks completion tests


FloGa:
While working on #1876, I noticed the following issue: Many tests rely on the fact that there is only one command that can be completed with "true", namely `true`. If however there is another command in `PATH` that starts with "true", then six tests fail because of that (tested with current master).

Example:

    $ touch ~/bin/truebar && chmod +x ~/bin/truebar
    $ cd ~/Projects/awesome
    $ make clean check-unit
    # (Output cropped, there are five identical messages from other tests)
    # ...
    Failure → spec/awful/completion_spec.lua @ 46
    awful.completion.shell in empty directory handles completion of true (bash)
    spec/awful/completion_spec.lua:47: Expected objects to be the same.
    Passed in:
    (table) {
      [1] = 'true'
      [2] = 5
     *[3] = {
        [1] = 'true' } }
    Expected:
    (table) {
      [1] = 'true'
      [2] = 5
     *[3] = {
        [1] = 'true'
       *[2] = 'truebar' } }
    # ...

I tried to change the related tests to not use `assert.same` but instead some sort of a "contains" check, but to give meaningful pass/fail messages I would have to rebuild parts of luassert, including several import requirements, which would be a huge overkill for this issue.

Any ideas how it could be solved otherwise?

blueyed:
I think we should set PATH completely to something we control.

FloGa:
I considered this possibility, but I couldn't find a good point in the command chain where I could modify the PATH reliably, without breaking functionality. We would have to modify the completion command so that it accepts environment variables to be passed down to `/usr/bin/env`. In the testing area, we would have to copy the Bash executable into a local directory, so it is found with the modified PATH.

Would this be acceptable for you?

blueyed:
Sounds good to me.

The EDIT basically creates a symlink?!

FloGa:
The (pseudo-)command I posted creates a tiny script that calls the actual bash, but a symlink might be an even better idea. I'm just careful with symlinks, because some programs behave differently when called via symlink. I'll test what works best.

FloGa:
Since issue-to-pullrequest seems to be deprecated, I created a new PR for this feature:  #1957

PR #1957, opened 2017-08-01
Set up isolated PATH for completion test suites

FloGa:
As discussed in #1923, the test cases will fail if there is a command in PATH which starts with "true". This PR fixes this problem by creating a temporary directory, symlinking needed commands, and using this directory as the only item in the PATH variable.

### DIFF
--- a/spec/awful/completion_spec.lua
+++ b/spec/awful/completion_spec.lua
@@ -46,6 +46,7 @@ local function remove_test_path_dir(test_path)
 end
 
 local test_dir
+local test_path
 
 --- Get and create a temporary test dir based on `pat`, where %d gets replaced by
 -- the current PID.
@@ -65,11 +66,13 @@ describe("awful.completion.shell in empty directory", function()
         io.popen = function(...)  --luacheck: ignore
             return orig_popen(string.format('cd %s && ', test_dir) .. ...)
         end
+        test_path = get_test_path_dir()
     end)
 
     teardown(function()
         assert.True(os.remove(test_dir))
         io.popen = orig_popen  --luacheck: ignore
+        remove_test_path_dir(test_path)
     end)
 
     if has_bash then
@@ -97,12 +100,14 @@ describe("awful.completion.shell", function()
         io.popen = function(...)  --luacheck: ignore
             return orig_popen(string.format('cd %s && ', test_dir) .. ...)
         end
+        test_path = get_test_path_dir()
     end)
 
     teardown(function()
         assert.True(os.remove(test_dir .. '/localcommand'))
         assert.True(os.remove(test_dir))
         io.popen = orig_popen  --luacheck: ignore
+        remove_test_path_dir(test_path)
     end)
 
     if has_bash then
@@ -159,11 +164,14 @@ describe("awful.completion.shell handles $SHELL", function()
         gdebug.print_warning = function(message)
             print_warning_message = message
         end
+
+        test_path = get_test_path_dir()
     end)
 
     teardown(function()
         os.getenv = orig_getenv  --luacheck: ignore
         gdebug.print_warning = orig_print_warning
+        remove_test_path_dir(test_path)
     end)
 
     before_each(function()
--- a/spec/awful/completion_spec.lua
+++ b/spec/awful/completion_spec.lua
@@ -15,6 +15,36 @@ if not has_bash and not has_zsh then
     return
 end
 
+local orig_path_env = GLib.getenv("PATH")
+local required_commands = {"bash", "zsh", "sort"}
+
+-- Change PATH
+local function get_test_path_dir()
+    local test_path = Gio.File.new_tmp("awesome-tests-path-XXXXXX")
+    test_path:delete()
+    test_path:make_directory()
+
+    for _, cmd in ipairs(required_commands) do
+        local target = GLib.find_program_in_path(cmd)
+        if target then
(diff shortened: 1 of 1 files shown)


spec/awful/completion_spec.lua
```diff
@@ -15,7 +15,38 @@ if not has_bash and not has_zsh then
...
+    local test_path = Gio.File.new_tmp("awesome-tests-path-XXXXXX")
+    test_path:delete()
+    test_path = test_path:get_path()
+    gfs.make_directories(test_path)
+
+    for _, cmd in ipairs(required_commands) do
+        local target = GLib.find_program_in_path(cmd)
+        if target then
+            os.execute(string.format('ln -s %s %s/%s', target, test_path, cmd))
+        end
```
blueyed: Could have an error in case it is not found?

blueyed: Is there something in GLib to create the symlink maybe?

spec/awful/completion_spec.lua
```diff
@@ -15,7 +15,38 @@ if not has_bash and not has_zsh then
     return
 end
 
+local orig_path = GLib.getenv("PATH")
```
blueyed: Name this `orig_PATH` maybe for clarity, or `orig_path_env`?

spec/awful/completion_spec.lua
FloGa: According to https://developer.gnome.org/glib/unstable/glib-Miscellaneous-Utility-Functions.html#g-find-program-in-path, if file is not found, NULL is returned, hence the `if target`. Since you also check in every test if bash/zsh is available, a missing file would not cause an error.

Regarding the symlink, I now triple checked the API and haven't found a `link` method or anything that resembles this functionality ... But it has a `chmod` method. Should I stick to my original plan and create executable files that call the appropriate command?

    local link = io.open(test_path .. "/" .. cmd, "w")
    link:write("exec " .. target .. ' "$@"')
    link:close()
    GLib.chmod(test_path .. "/" .. cmd, 0700)

(Not sure about the octal representation of the file mode, I would test it if you want to go with this solution.)

spec/awful/completion_spec.lua
FloGa: I would go with `orig_path_env` to keep it lowercase.

spec/awful/completion_spec.lua
blueyed: IMHO the `ln -s` is fine then.

spec/awful/completion_spec.lua
blueyed: OK.

spec/awful/completion_spec.lua
psychon: https://developer.gnome.org/gio/stable/GFile.html#g-file-make-symbolic-link

psychon:
```
lua -e 'print(require("lgi").Gio.File.new_for_path("/tmp/foo"):make_symbolic_link("target", nil))'
```

FloGa: Nice, I didn't see this one. I guess I'm going with this then.

FloGa:
I changed the variable name and replaced the os call with a Gio method.

As before, to keep the discussion history I just did fixup commits. If you approve the PR, I will rebase it onto the current master and resolve all fixups.

FloGa:
Thank you for approval. I rebased on current master and squashed the fixup commits.

spec/awful/completion_spec.lua
```diff
@@ -15,6 +15,37 @@ if not has_bash and not has_zsh then
...
+    local test_path = Gio.File.new_tmp("awesome-tests-path-XXXXXX")
+    test_path:delete()
+    test_path = test_path:get_path()
+    gfs.make_directories(test_path)
+
+    for _, cmd in ipairs(required_commands) do
+        local target = GLib.find_program_in_path(cmd)
+        if target then
+            local link = string.format('%s/%s', test_path, cmd)
+            Gio.File.new_for_path(link):make_symbolic_link(target)
```
psychon: It does not really make a difference, but:
```lua
local test_path = Gio.file.new_tmp("awesome-tests-path-XXXXXX")
test_path:delete()
test_path:make_directory()

for _, cmd in ipairs(required_commands) do
    local target = GLib.find_program_in_path(cmd)
    if target then
        target:get_child(cmd):make_symbolic_link(target)
    end
```

FloGa: `target` is a string, not a GFile object, hence there is no `get_child()`. Shall I instantiate `target` as a new GFile object, then?

psychon: Whoops, my fault. Let's try it step-by-step this time:
```lua
local link = string.format('%s/%s', test_path, cmd)
Gio.File.new_for_path(link):make_symbolic_link(target)
```
This constructs `cmd` as a child of `test_path` and then creates a GFile instance for it. So what I should have written is `test_path:get_child(cmd):make_symbolic_link(target)`. I'll update my above comment for this.

FloGa: Ah, now I see. I will keep returning the path as a string however, since it is needed in other parts for easier deletion of this whole directory (like in the existing `get_test_dir` method).

FloGa:
Added the requested changes to simplify creation of testing path environment and symlinking needed commands.

blueyed:
Thank you!

PR #1962, opened 2017-08-08
Fix tests/examples: generate deps for output files

blueyed:
This still runs all the tests in the configure phase, since the tests
report back what output file (png/svg/none) will be generated.. :/

This could be improved by having this table in CMakeLists directly, or
by having a callback into the tests that only reports back that
information.  The latter would still execute all the tests (via a Lua
process).

Continuation of https://github.com/awesomeWM/awesome/pull/1913 to get feedback from the doc-bot.

### DIFF
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -16,8 +16,9 @@ endif()
 
 cmake_minimum_required(VERSION 3.0.0)
 
-# Get and update the LUA_PATH so the scripts can be executed without Awesome.
-execute_process(COMMAND lua -e print\(package.path\) OUTPUT_VARIABLE "LUA_PATH_")
+# Get and update the LUA_PATH so the scripts can be executed without awesome.
+execute_process(COMMAND lua -e "p = package.path:gsub(';', '\\\\;'); io.stdout:write(p)"
+    OUTPUT_VARIABLE "LUA_PATH_")
 
 # Make sure the system can be called from the test directory
 if(NOT SOURCE_DIR AND ${CMAKE_CURRENT_SOURCE_DIR} MATCHES "/tests/examples")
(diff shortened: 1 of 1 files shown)


blueyed:
Generated images look OK from the "relevant changes" diff - assuming that the calendar ones have been "unstable" before already.

psychon:
> assuming that the calendar ones have been "unstable" before already.

Well, as you say: "unstable"
They use the current month and, well, it's August now.

blueyed:
> They use the current month and, well, it's August now.

Since a while, so it should have been used on master before.. but likely the current day is used, too.

blueyed:
@Elv13 
Please review.

Elv13:
The big issue I have with this is that if this is merged, then the system stops doing its job. If I modify "acrchart.lua", then none of its tests are executed when I run `make` or `make check`. Otherwise it seem to work fine. On the other hand, running these tests every single time `cmake` run is a waste of time and it gets worst as I add more and more of them to cover more part of the API add more examples to the doc.

The alternative is/was to rewrite the runner in Lua and run it was a daemon + worker threads model. I wrote some code last year to do it, but never really had time to finish everything. Nor do I plan to have much time to do it soon. I spent all the time I had in the last few months on the notification rework to address its limitations. I am still submitting patches from 2016. In that view, this pull request is superior to "wait for Elv13 forever" scenario. I don't oppose it. I just don't like it makes development of the doc harder without the ability to rebuild all examples when `make check` is executed.

blueyed:
> If I modify "acrchart.lua", then none of its tests are executed when I run make or make check

We could trace the included files somehow (e.g. by printing them to stderr with some prefix maybe; or using luacov for this), and use this as dependencies then?

What about an explicit check-examples target for now? (added)
The downside with this is that it gets run twice on a clean build with `make check` now.

blueyed:
Rebased.

@Elv13 
Good to merge like this from your side?

Elv13:
Sure

blueyed:
What a relief! ;)

PR #1963, opened 2017-08-09
build: use --fatalwarnings with ldoc, remove check-ldoc-warnings

blueyed:
Instead of having a custom target with redirection magic, just use
`--fatalwarnings` with ldoc.  This requires ldoc 1.4.5.


blueyed:
> Just the first commit, right?

Sure.. :)

PR #1964, opened 2017-08-09
Travis: use cluacov for improved performance

blueyed:
Luacheck uses it and it improves performance of `busted -c` there.
https://github.com/mpeterv/cluacov

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,8 +97,15 @@ install:
 
   # Install dependencies for code coverage testing.
   - if [ "$DO_COVERAGE" = "" ]; then export DO_COVERAGE=0; fi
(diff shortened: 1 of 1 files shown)


blueyed:
Much faster indeed! :fireworks: 
(4:15 and 3:29 for the coverage builds; vs. 10:50 and 9:31 for the last build on master)

The main coverage change is `spawn_monitor_timeout` not being called anymore - which appears to kick in after 20s.
Maybe during the tests `AWESOME_SPAWN_TIMEOUT` could be patched to be shorter? (assuming that it was really covered before)

blueyed:
Agreed.  Merged/cherry-picked only the first commit into master.

PR #1967, opened 2017-08-10
Test all themes

blueyed:
Continuation/rewite of https://github.com/awesomeWM/awesome/pull/1934.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,7 +317,7 @@ endif()
 # }}}
 
 # {{{ Theme icons
-file(GLOB icon_sources RELATIVE ${SOURCE_DIR} ${SOURCE_DIR}/themes/*/titlebar/*.png)
+file(GLOB_RECURSE icon_sources RELATIVE ${SOURCE_DIR} ${SOURCE_DIR}/themes/*.png)
 
 foreach(icon ${icon_sources})
     # Copy all icons to the build dir to simplify the following code.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -405,6 +405,15 @@ add_custom_target(check-integration
     USES_TERMINAL
     VERBATIM)
 add_dependencies(check-integration test-gravity)
+add_custom_target(check-themes
+    ${CMAKE_SOURCE_DIR}/tests/themes/run.sh
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+    COMMENT "Testing themes"
+    USES_TERMINAL
+    DEPENDS generated_icons generate_awesomerc ${PROJECT_AWE_NAME}
+    VERBATIM)
+add_dependencies(check-integration check-themes)
+
 add_custom_target(check-requires
     lua "${CMAKE_SOURCE_DIR}/build-utils/check_for_invalid_requires.lua"
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -161,9 +161,9 @@ wait_until_success "setup xrdb" "printf 'Xft.dpi: 96
 # Use a separate D-Bus session; sets $DBUS_SESSION_BUS_PID.
 eval "$(DISPLAY="$D" dbus-launch --sh-syntax --exit-with-session)"
 
-RC_FILE=${source_dir}/awesomerc.lua
-export AWESOME_THEMES_PATH="$source_dir/themes"
-export AWESOME_ICON_PATH="$source_dir/icons"
+RC_FILE=${AWESOME_RC_FILE:-${source_dir}/awesomerc.lua}
+AWESOME_THEMES_PATH="${AWESOME_THEMES_PATH:-${source_dir}/themes}"
+AWESOME_ICON_PATH="${AWESOME_ICON_PATH:-${source_dir}/icons}"
 
 # Inject coverage runner via temporary RC file.
 if [ -n "$DO_COVERAGE" ] && [ "$DO_COVERAGE" != 0 ]; then
@@ -181,7 +181,10 @@ start_awesome() {
     cd "$build_dir"
     # Kill awesome after $timeout_stale seconds (e.g. for errors during test setup).
     # SOURCE_DIRECTORY is used by .luacov.
-    DISPLAY="$D" SOURCE_DIRECTORY="$source_dir" timeout "$timeout_stale" "$AWESOME" -c "$RC_FILE" "${awesome_options[@]}" > "$awesome_log" 2>&1 &
+    DISPLAY="$D" SOURCE_DIRECTORY="$source_dir" \
+        AWESOME_THEMES_PATH="$AWESOME_THEMES_PATH" \
+        AWESOME_ICON_PATH="$AWESOME_ICON_PATH" \
+        timeout "$timeout_stale" "$AWESOME" -c "$RC_FILE" "${awesome_options[@]}" > "$awesome_log" 2>&1 &
     awesome_pid=$!
     cd - >/dev/null
 
--- a/tests/themes/run.sh
+++ b/tests/themes/run.sh
@@ -1,0 +1,31 @@
+#!/bin/sh
+#
+# Test all themes.
+# This should be run via `make check-themes` (or manually from the build
+# directory).
+
+set -e
+
+if ! [ -f CMakeCache.txt ]; then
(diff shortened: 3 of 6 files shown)


CMakeLists.txt
```diff
@@ -405,6 +405,15 @@ add_custom_target(check-integration
...
     VERBATIM)
 add_dependencies(check-integration test-gravity)
+add_custom_target(check-themes
+    tests/themes/run.sh
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    COMMENT "Testing themes"
+    USES_TERMINAL
+    DEPENDS generated_icons generate_awesomerc ${PROJECT_AWE_NAME}
+    VERBATIM)
+add_dependencies(check-integration check-themes)
```
blueyed: If `check-themes` fails, the other integration tests are not being run.
That applies to `test-gravity` also.
Should we rather only run them explicitly (on Travis)?
There's no point really to run `check-themes` locally most of the time.

psychon: I don't really like "only on Travis" tests. If I tell CMake "run tests", I want it to run tests and not be surprised 10 minutes later when the build on Travis fails.

If `check-themes` fails, then tests failed. I agree that it might be nice to get all test failures, but then people will have to use `make -k`.

blueyed: Ok, makes sense.
`make -k` seems useful in this context.

tests/run.sh
```diff
@@ -161,9 +161,9 @@ wait_until_success "setup xrdb" "printf 'Xft.dpi: 96
 # Use a separate D-Bus session; sets $DBUS_SESSION_BUS_PID.
 eval "$(DISPLAY="$D" dbus-launch --sh-syntax --exit-with-session)"
 
-RC_FILE=${source_dir}/awesomerc.lua
-export AWESOME_THEMES_PATH="$source_dir/themes"
-export AWESOME_ICON_PATH="$source_dir/icons"
+RC_FILE=${AWESOME_RC_FILE:-${source_dir}/awesomerc.lua}
+AWESOME_THEMES_PATH="${AWESOME_THEMES_PATH:-$source_dir/themes}"
+AWESOME_ICON_PATH="${AWESOME_ICON_PATH:-$source_dir/icons}"
```
psychon: Why no longer `export`?

tests/run.sh
```diff
@@ -162,8 +162,8 @@ wait_until_success "setup xrdb" "printf 'Xft.dpi: 96
 eval "$(DISPLAY="$D" dbus-launch --sh-syntax --exit-with-session)"
 
 RC_FILE=${AWESOME_RC_FILE:-${source_dir}/awesomerc.lua}
-AWESOME_THEMES_PATH="${AWESOME_THEMES_PATH:-$source_dir/themes}"
-AWESOME_ICON_PATH="${AWESOME_ICON_PATH:-$source_dir/icons}"
+AWESOME_THEMES_PATH="${AWESOME_THEMES_PATH:-${source_dir}/themes}"
+AWESOME_ICON_PATH="${AWESOME_ICON_PATH:-${source_dir}/icons}"
```
psychon: Fixup applied to the wrong commit? (Should be in the previous one)

tests/themes/run.sh
```diff
@@ -26,6 +27,7 @@ for theme in themes/*/theme.lua; do
   output=$(AWESOME_RC_FILE="$config_file" \
     AWESOME_THEMES_PATH="$PWD/themes" \
     AWESOME_ICON_PATH="$PWD/icons" \
+    VERBOSE=1 \
```
psychon: Why this commit? Why shouldn't `tests/themes/run.sh` use a `$VERBOSE` env var in the same way that `tests/run.sh` does?

tests/themes/run.sh
```diff
@@ -12,30 +12,20 @@ if ! [ -f CMakeCache.txt ]; then
...
     VERBOSE=1 \
-    "$source_dir/tests/run.sh" themes/tests.lua \
-    | grep -Ev '^(awesome_log:|== Running|1 tests finished.|.{19} W: awesome: (a_glib_poll|Cannot reliably detect EOF)|xterm: cannot load font|Running step)' || true)
-  if [ -n "$output" ]; then
-    echo "== Unexpected output: =="
-    echo "$output"
-    echo "========================"
-    exit_code=1
-  fi
+    "$tests_dir/run.sh" themes/tests.lua
```
psychon: Should this be squashed with the original commit which added `tests/themes/run.sh`?

tests/themes/run.sh
```diff
@@ -23,6 +23,8 @@ for theme in themes/*/theme.lua; do
 
   sed "s~default/theme~$theme/theme~g" "awesomerc.lua" > "$config_file"
 
+  # Set CMAKE_BINARY_DIR for out-of-tree builds.
+  CMAKE_BINARY_DIR="$PWD" \
```
psychon: Is this necessary here? Shouldn't this be done in `tests/run.sh` instead? What uses this env var anyway? I don't think anything here is using CMake?

blueyed: `tests/run.sh` uses it.
It could use something else maybe.
As for `check-integration` there it gets set probably, but when run via `tests/themes/run.sh` (via `check-integration` => `check-themes`) it does not seem to be available?!
So CMake maybe sets it, but not exported?!

tests/run.sh
blueyed: Why?
I think it is better to not export if it is not necessary.
Makes it clearer that it is also used here.

tests/themes/tests.lua
```diff
@@ -0,0 +1,18 @@
...
+-- This is used from tests/themes/run.sh.
+
+local awful = require("awful")
+
+local steps = {
+    function(count)
+        if count <= 2 then
+            awful.spawn("xterm")
+        elseif #client.get() == 2 then
+            return true
```
actionless: looks good however i am not sure about assertion criteria -- it will test what the theme hasn't paniced, but i am wondering mb it could make sense to also observe stderr for errors/warnings? or parsing text output sounds too much fragile for the tests?

tests/themes/run.sh
blueyed: For debugging, will remove it.

tests/themes/tests.lua
blueyed: Errors will cause `tests/run.sh` to fail already (and we have `set -e` here).
With https://github.com/awesomeWM/awesome/pull/1973 also warnings (e.g. deprecations) will cause a failure.

tests/run.sh
blueyed: Good catch, fixed.

tests/themes/tests.lua
actionless: yup, #1973 actually addresses my concern, thanks for clarifying!

tests/themes/run.sh
blueyed: Done.

blueyed:
@psychon 
Please review again / see if my comments make sense.

blueyed:
Needs a fixup, it failed on master:
```
== Running themes/tests.lua ==
xterm: cannot load font '-misc-fixed-medium-r-semicondensed--13-120-75-75-c-60-iso10646-1'
xterm: cannot load font '-misc-fixed-medium-r-semicondensed--13-120-75-75-c-60-iso10646-1'
1 tests finished.
== Testing themes/sky/theme.lua ==
awesome_log: /tmp/tmp.LSRwPI2DCr/_awesome_test.log
(EE) 
Fatal server error:
(EE) Server is already active for display 5
	If this server is no longer running, remove /tmp/.X5-lock
	and start again.
(EE) 
Error: failed to setup xrdb!
Last reply: xrdb: Connection refused
xrdb: Can't open display ':5'.
```

Looks like the previous server was not stopped.

psychon:
Did you restart the build? I don't see a failure on master: https://travis-ci.org/awesomeWM/awesome/builds/263807371

To me it looks like a race condition, as we see so often.
`tests/run.sh` uses a trap (line 96) to kill its X11 server when it is done. This just sends a SIGTERM to the server. Immediately afterwards this exits. If you want me to guess: The next call of the script was fast enough to try to start a new server before the already-existing one was done shutting down.
This would explain why the new server failed to start, but `xrdb` still reports "Can't open display".

blueyed:
> Did you restart the build?

Yes.

Your explanation makes sense, and seems to be triggered now by using the `tests/run.sh` script twice.

PR #1968, opened 2017-08-10
gears.debug.print_{warning_error}: add "awesome: " prefix

blueyed:
This makes it consistent with the C functions (`_warn` and `_fatal`).


psychon:
Fine by me

PR #1970, opened 2017-08-10
gears.timer.delayed_call: copy queue for processing

blueyed:
This allows for re-queuing a delayed call from the delayed call itself,
without causing an infinite loop.

### DIFF
--- a/tests/test-timers.lua
+++ b/tests/test-timers.lua
@@ -1,0 +1,51 @@
+local gears = require("gears")
+
+local delayed_call
+local cb_count = 0
+
+local steps = {
+    -- Basic functionality of gears.timer.delayed_call.
+    function(_)
+        gears.timer.delayed_call(function()
+            delayed_call = true
+        end)
+        return true
+    end,
+    function(_)
+        assert(delayed_call)
+        return true
+    end,
+
+    -- gears.timer.delayed_call: re-queue function.
+    function(_)
+        local function cb()
+            cb_count = cb_count + 1
+            if cb_count < 3 then
+                gears.timer.delayed_call(function()
+                    cb()
+                end)
+            end
+        end
+
+        gears.timer.delayed_call(function()
+            cb()
+        end)
+        return true
+    end,
+    function(_)
+        assert(cb_count == 2, 'Count should be 2, but is: ' .. cb_count)
+        return true
+    end,
+    function(_)
+        assert(cb_count == 3, cb_count)
+        return true
+    end,
+    function(_)
+        assert(cb_count == 3, cb_count)
+        return true
+    end,
+}
+
+require("_runner").run_steps(steps)
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/lib/gears/timer.lua
+++ b/lib/gears/timer.lua
@@ -60,6 +60,7 @@ local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility
 local glib = require("lgi").GLib
 local object = require("gears.object")
 local protected_call = require("gears.protected_call")
+local gtable = require("gears.table")
 
 --- Timer objects. This type of object is useful when triggering events repeatedly.
 -- The timer will emit the "timeout" signal every N seconds, N being the timeout
@@ -227,10 +228,12 @@ end
 
 local delayed_calls = {}
 capi.awesome.connect_signal("refresh", function()
-    for _, callback in ipairs(delayed_calls) do
+    -- Copy delayed_calls before processing them, to allow for re-queueing.
+    local callbacks = gtable.clone(delayed_calls, false)
+    delayed_calls = {}
+    for _, callback in ipairs(callbacks) do
         protected_call(unpack(callback))
     end
-    delayed_calls = {}
 end)
 
 --- Call the given function at the end of the current main loop iteration


Elv13:
+1

Just a note that the test can use the shims by calling `awesome:emit_signal("refresh")` at the end of the example. This would save an integration testsuit and can contain a code example for the documentation. The `tests/example/text/gears/timer` would be a good place.

blueyed:
@Elv13 
How is it supposed to look like then?
I've tried `tests/examples/text/gears/timer.lua`:
```lua
local gears = require("gears") --DOC_HIDE

local delayed_call --DOC_HIDE
local cb_count = 0 --DOC_HIDE

    delayed_call = true
end)

awesome:emit_signal("refresh")  --DOC_HIDE
assert(delayed_call)  --DOC_HIDE
```

But it fails with:
```
Result: 1
Output: emit

Error:  lua: ...e/daniel/src/awesome/tests/examples/text/gears/timer.lua:11: assertion failed!
stack traceback:
	[C]: in function 'assert'
	...e/daniel/src/awesome/tests/examples/text/gears/timer.lua:11: in main chunk
	/home/daniel/src/awesome/tests/examples/text/template.lua:5: in main chunk
	[C]: in ?

-- Example tests passed.
```

The prints are coming from:
```diff
--- i/tests/examples/shims/awesome.lua
+++ w/tests/examples/shims/awesome.lua
@@ -27,8 +27,10 @@ local function _shim_fake_class()
     end
 
     function obj.emit_signal(name, c, ...)
+        print("emit")
         local conns = obj._signals[name] or {strong={}}
         for func in pairs(conns.strong) do
+            print(func)
             func(c, ...)
         end
     end
```

btw: re "Example tests passed.": I see that `cmake -DSTRICT_TESTS=1 .` is required for them to fail, but it is a bit misleading as of now; I think it should not say that they pass by default?!

blueyed:
At least with e.g. `make check-integration` on a clean build it should cause a failure in case one example test fails, no?

psychon:
What's the intended semantics here? A delayed call happens at the end of the current main loop iteration. Inside a delayed call that so far meant "immediately". Your patch changes that, but the new semantics seems to be "a random time later" (whenever awesome happens to wake up for whatever reason). I don't really like that.

So, why do you need this? Should your code perhaps just instead connect to the `refresh` signal on `awesome`?

Elv13:
My bad, it's `.`, not `:`, sorry (and yes, one of the existing test do have a "bug")

```
diff --git a/lib/gears/timer.lua b/lib/gears/timer.lua
index da9f4de71..750a09e4a 100644
--- a/lib/gears/timer.lua
+++ b/lib/gears/timer.lua
@@ -228,7 +228,8 @@ capi.awesome.connect_signal("refresh", function()
     delayed_calls = {}
 end)
 
---- Call the given function at the end of the current main loop iteration
+--- Call the given function at the end of the current main loop iteration.
+--@DOC_text_gears_timer_delayed_call_EXAMPLE@
 -- @tparam function callback The function that should be called
 -- @param ... Arguments to the callback function
 -- @function gears.timer.delayed_call

```

```
local gears = require("gears") --DOC_HIDE

local delayed_call --DOC_HIDE
local cb_count = 0 --DOC_HIDE

    delayed_call = true
end)

awesome.emit_signal("refresh")  --DOC_HIDE
assert(delayed_call)  --DOC_HIDE
```

This one work

blueyed:
@psychon 
I've noticed this causing an infinite loop, when what I've meant to do was to retry it again on the next event loop.

> A delayed call happens at the end of the current main loop iteration. Inside a delayed call that so far meant "immediately". 

So what is the point then of using a delayed call from within a delayed call?  If you expect it to be called immediately, you would just call it right away.
(Currently it is not really immediately if there are other queued calls however)

Should I go ahead and apply the suggested changes from @Elv13 ?

psychon:
> I've noticed this causing an infinite loop, when what I've meant to do was to retry it again on the next event loop.

What is it that you want to do at the end of "some random event loop iteration"? Why shouldn't it use the refresh signal? That would be more explicit which could avoid problems where things were not meant to be delayed.

blueyed:
> What is it that you want to do at the end of "some random event loop iteration"?

I've used it to delay sending `SIGSTOP` to a process in case the mouse was moved (comparing current coords and the coords at the end of the loop).  If it was changed (and other conditions are met) it gets re-queued (but should not happen at the end of the current loop).

```lua
  local function delayed_cont(c, mouse_coords)
    if client.focus == c then
      local coords = mouse.coords()
      if mouse_coords.x == coords.x and mouse_coords.y == coords.y then
        sigcont(c)
        sigstopped_clients[c] = nil
      else
        timer.start_new(0.1, function()
          timer.delayed_call(function() delayed_cont(c, coords) end)
        end)
      end
    end
  end
```

> Why shouldn't it use the refresh signal?

The delayed calls get executed through the "refresh" signal, i.e. `gears.timer.delayed_calls` is just a wrapper around it.

psychon:
Do I understand it correctly that you want to get rid of the 0.1 seconds timer in your code and want to just use `timer.delayed_call` directly? And the intended behaviour is "sigcont a client if mouse does not move".

That won't work. Since the mouse is not moving, awesome does not run a main loop iteration. So your delayed call will be delayed "infinitely" (well, I guess after a minute when the textclock updates it will also update).

So I don't think that just allowing to insert a delayed call for the "current+1" main loop iteration really is useful for anything and it can actually cause hard to debug problems ("why isn't my client SIGCONT'd?").

blueyed:
> Do I understand it correctly that you want to get rid of the 0.1 seconds timer in your code and want to just use timer.delayed_call directly? 

Yes.

> And the intended behaviour is "sigcont a client if mouse does not move".

No, but "sigcont a client if mouse did not move since the last check": it is meant to not send SIGCONT in case you're moving your mouse quickly from one client to another, crossing the SIGSTOPped client, but not staying there.

> So I don't think that just allowing to insert a delayed call for the "current+1" main loop iteration really is useful for anything and it can actually cause hard to debug problems ("why isn't my client SIGCONT'd?").

Hmm.  I've found the case of "awesome becoming inresponsive" quite bad (SIGKILL was required).

psychon:
> Hmm. I've found the case of "awesome becoming inresponsive" quite bad (SIGKILL was required).

Well, yeah, but that only happens to the person writing the code. Debugging "phase of the moon bugs" (things that seem to happen only randomly) is so much worse.

So all together I prefer not to have this patch.

tests/test-timers.lua
```diff
@@ -0,0 +1,52 @@
...
+        end)
+        return true
+    end,
+    function(_)
+        assert(delayed_call)
+        return true
+    end,
+
+    -- gears.timer.delayed_call: re-queue function.
+    function(_)
```
actionless: this test is failing without the patch to `timer.lua` ,right?

blueyed: Yes.

blueyed:
I prefer not to have since when @psychon is not convinced.

We could however have an test for the current behavior then?!

actionless:
mb also to add some very short note to the docstring of `delayed_call` to make debugging things related to such unobvious behavior a bit easier?

blueyed:
Rebased for now.
I think I've just hit this again in some way (where awesome needs to be kill-9'ed).

Common stack from 4 gdb `bt`s:
```0x00007f1538972316 in  () at /usr/lib/libluajit-5.1.so.2
0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
0x000055ef75214d0e in luaA_dofunction (nret=0, nargs=0, L=0x40c08378) at …/src/awesome/src/build/common/lualib.h:76
0x000055ef75214d0e in signal_object_emit (L=0x40c08378, arr=arr@entry=0x55ef7523a600 <global_signals>, name=name@entry=0x55ef7522a5fb "refresh", nargs=nargs@entry=0)
at …/src/awesome/src/build/common/luaobject.c:253
0x000055ef75209ca7 in luaA_emit_refresh () at …/src/awesome/src/build/luaa.c:1024
0x000055ef751feca4 in awesome_refresh () at …/src/awesome/src/build/event.h:53
0x000055ef751feca4 in a_glib_poll (ufds=0x55ef7591e4b0, nfsd=5, timeout=596) at …/src/awesome/src/build/awesome.c:417
0x00007f153acd8ee0 in  () at /usr/lib/libglib-2.0.so.0
0x00007f153acd9f62 in g_main_loop_run () at /usr/lib/libglib-2.0.so.0
0x000055ef752007c6 in main (argc=<optimized out>, argv=<optimized out>) at …/src/awesome/src/build/awesome.c:884
```

Current stack:
```
#0  0x00007f1538974516 in  () at /usr/lib/libluajit-5.1.so.2
#1  0x00007f153899b65c in  () at /usr/lib/libluajit-5.1.so.2
#2  0x00007f153899bb74 in  () at /usr/lib/libluajit-5.1.so.2
#3  0x00007f15389a3071 in  () at /usr/lib/libluajit-5.1.so.2
#4  0x00007f15389713e5 in  () at /usr/lib/libluajit-5.1.so.2
#5  0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
#6  0x000055ef75213c79 in luaA_call_handler (handler=11, L=0x40c08378) at /home/daniel/src/awesome/src/build/common/lualib.h:110
        nargs = 2
        error_func_pos = 1
        __FUNCTION__ = "luaA_call_handler"
        class = <optimized out>
        attr = <optimized out>
        prop = <optimized out>
#7  0x000055ef75213c79 in luaA_class_index (L=0x40c08378) at /home/daniel/src/awesome/src/build/common/luaclass.c:440
        class = <optimized out>
        attr = <optimized out>
        prop = <optimized out>
#8  0x00007f1538972316 in  () at /usr/lib/libluajit-5.1.so.2
#9  0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
#10 0x000055ef75213c79 in luaA_call_handler (handler=9, L=0x40c08378) at /home/daniel/src/awesome/src/build/common/lualib.h:110
        nargs = 2
        error_func_pos = 1
        __FUNCTION__ = "luaA_call_handler"
        class = <optimized out>
        attr = <optimized out>
        prop = <optimized out>
#11 0x000055ef75213c79 in luaA_class_index (L=0x40c08378) at /home/daniel/src/awesome/src/build/common/luaclass.c:440
        class = <optimized out>
        attr = <optimized out>
        prop = <optimized out>
#12 0x00007f1538972316 in  () at /usr/lib/libluajit-5.1.so.2
#13 0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
#14 0x000055ef75214d0e in luaA_dofunction (nret=0, nargs=0, L=0x40c08378) at /home/daniel/src/awesome/src/build/common/lualib.h:76
        error_func_pos = 1
        __FUNCTION__ = "luaA_dofunction"
        i = 0
        nbfunc = 1
        sigfound = <optimized out>
#15 0x000055ef75214d0e in signal_object_emit (L=0x40c08378, arr=arr@entry=0x55ef7523a600 <global_signals>, name=name@entry=0x55ef7522a5fb "refresh", nargs=nargs@entry=0)
    at /home/daniel/src/awesome/src/build/common/luaobject.c:253
        i = 0
        nbfunc = 1
        sigfound = <optimized out>
#16 0x000055ef75209ca7 in luaA_emit_refresh () at /home/daniel/src/awesome/src/build/luaa.c:1024
#17 0x000055ef751feca4 in awesome_refresh () at /home/daniel/src/awesome/src/build/event.h:53
        res = <optimized out>
        now = {tv_sec = 0, tv_usec = 0}
        length_time = <optimized out>
        length = <optimized out>
        saved_errno = <optimized out>
        L = 0x40c08378
        __FUNCTION__ = "a_glib_poll"
#18 0x000055ef751feca4 in a_glib_poll (ufds=0x55ef7591e4b0, nfsd=5, timeout=596) at /home/daniel/src/awesome/src/build/awesome.c:417
        res = <optimized out>
        now = {tv_sec = 0, tv_usec = 0}
        length_time = <optimized out>
        length = <optimized out>
        saved_errno = <optimized out>
        L = 0x40c08378
        __FUNCTION__ = "a_glib_poll"
#19 0x00007f153acd8ee0 in  () at /usr/lib/libglib-2.0.so.0
#20 0x00007f153acd9f62 in g_main_loop_run () at /usr/lib/libglib-2.0.so.0
#21 0x000055ef752007c6 in main (argc=<optimized out>, argv=<optimized out>) at /home/daniel/src/awesome/src/build/awesome.c:884
        confpath = 0x0
        searchpath = {tab = 0x0, len = 2, size = 24}
        xfd = <optimized out>
        opt = <optimized out>
        xdg = {reserved = 0x55ef755e8440}
        no_argb = <optimized out>
        run_test = <optimized out>
        replace_wm = <optimized out>
        tree_c = {sequence = 0}
        long_options = 
            {{name = 0x55ef7522957a "help", has_arg = 0, flag = 0x0, val = 104}, {name = 0x55ef7522957f "version", has_arg = 0, flag = 0x0, val = 118}, {name = 0x55ef75229587 "config", has_arg = 1, flag = 0x0, val = 99}, {name = 0x55ef7522958e "check", has_arg = 0, flag = 0x0, val = 107}, {name = 0x55ef75229594 "search", has_arg = 1, flag = 0x0, val = 115}, {name = 0x55ef7522959b "no-argb", has_arg = 0, flag = 0x0, val = 97}, {name = 0x55ef752295a3 "replace", has_arg = 0, flag = 0x0, val = 114}, {name = 0x55ef752295ab "reap", has_arg = 1, flag = 0x0, val = 1}, {name = 0x0, has_arg = 0, flag = 0x0, val = 0}}
        __FUNCTION__ = "main"
        xdgconfigdirs = <optimized out>
        sa = 
          {__sigaction_handler = {sa_handler = 0x55ef751feed7 <signal_child>, sa_sigaction = 0x55ef751feed7 <signal_child>}, sa_mask = {__val = {0 <repeats 16 times>}}, sa_flags = 268435457, sa_restorer = 0x0}
        channel = <optimized out>
        query = <optimized out>
```

Where:
```
#11 luaA_class_index (L=0x40c08378) at …/src/awesome/src/build/common/luaclass.c:440
440	            return luaA_call_handler(L, class->index_miss_handler);
(gdb) l
435	            return prop->index(L, luaA_checkudata(L, 1, class));
436	    }
437	    else
438	    {
439	        if(class->index_miss_handler != LUA_REFNIL)
440	            return luaA_call_handler(L, class->index_miss_handler);
441	        if(class->index_miss_property)
442	            return class->index_miss_property(L, luaA_checkudata(L, 1, class));
443	    }
444	
```

blueyed:
I think this one is interesting:
```
#0  0x00007f1538842bb1 in poll () at /usr/lib/libc.so.6
#1  0x00007f153ae17180 in  () at /usr/lib/libxcb.so.1
#2  0x00007f153ae18b3f in  () at /usr/lib/libxcb.so.1
#3  0x00007f153ae18c51 in xcb_wait_for_reply () at /usr/lib/libxcb.so.1
#4  0x000055ef752116c4 in xwindow_get_shape (win=35651593, kind=kind@entry=XCB_SHAPE_SK_CLIP) at …/src/awesome/src/build/xwindow.c:293
#5  0x000055ef752171e3 in luaA_client_get_client_shape_clip (L=L@entry=0x40c08378, c=<optimized out>) at …/src/awesome/src/build/objects/client.c:3539
#6  0x000055ef75213d42 in luaA_class_index (L=0x40c08378) at …/src/awesome/src/build/common/luaclass.c:435
#7  0x00007f1538972316 in  () at /usr/lib/libluajit-5.1.so.2
#8  0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
#9  0x000055ef75214d0e in luaA_dofunction (nret=0, nargs=1, L=0x40c08378) at …/src/awesome/src/build/common/lualib.h:76
#10 0x000055ef75214d0e in signal_object_emit (L=L@entry=0x40c08378, arr=<optimized out>, name=name@entry=0x55ef7522a276 "property::border_width", nargs=nargs@entry=1)
    at …/src/awesome/src/build/common/luaobject.c:253
#11 0x000055ef75213a05 in luaA_class_emit_signal (L=L@entry=0x40c08378, lua_class=<optimized out>, name=name@entry=0x55ef7522a276 "property::border_width", nargs=nargs@entry=1)
    at …/src/awesome/src/build/common/luaclass.c:321
#12 0x000055ef752150e0 in luaA_object_emit_signal (L=L@entry=0x40c08378, oud=oud@entry=-3, name=name@entry=0x55ef7522a276 "property::border_width", nargs=nargs@entry=0)
    at …/src/awesome/src/build/common/luaobject.c:309
#13 0x000055ef75227608 in window_set_border_width (L=L@entry=0x40c08378, idx=idx@entry=-3, width=0) at …/src/awesome/src/build/objects/window.c:226
#14 0x000055ef7522769a in luaA_window_set_border_width (L=L@entry=0x40c08378, c=<optimized out>) at …/src/awesome/src/build/luaa.h:170
#15 0x000055ef75213dc0 in luaA_class_newindex (L=0x40c08378) at …/src/awesome/src/build/common/luaclass.c:467
#16 0x00007f1538972316 in  () at /usr/lib/libluajit-5.1.so.2
#17 0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
#18 0x000055ef75215054 in luaA_dofunction (nret=0, nargs=1, L=0x40c08378) at …/src/awesome/src/build/common/lualib.h:76
#19 0x000055ef75215054 in luaA_object_emit_signal (L=L@entry=0x40c08378, oud=oud@entry=1, name=0x4093faf8 "arrange", nargs=nargs@entry=0)
    at …/src/awesome/src/build/common/luaobject.c:302
#20 0x000055ef7521521a in luaA_object_emit_signal_simple (L=0x40c08378) at …/src/awesome/src/build/common/luaobject.c:329
#21 0x00007f1538972316 in  () at /usr/lib/libluajit-5.1.so.2
#22 0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
#23 0x000055ef75214d0e in luaA_dofunction (nret=0, nargs=0, L=0x40c08378) at …/src/awesome/src/build/common/lualib.h:76
#24 0x000055ef75214d0e in signal_object_emit (L=0x40c08378, arr=arr@entry=0x55ef7523a600 <global_signals>, name=name@entry=0x55ef7522a5fb "refresh", nargs=nargs@entry=0)
    at …/src/awesome/src/build/common/luaobject.c:253
#25 0x000055ef75209ca7 in luaA_emit_refresh () at …/src/awesome/src/build/luaa.c:1024
#26 0x000055ef751feca4 in awesome_refresh () at …/src/awesome/src/build/event.h:53
#27 0x000055ef751feca4 in a_glib_poll (ufds=0x55ef7591e4b0, nfsd=5, timeout=596) at …/src/awesome/src/build/awesome.c:417
#28 0x00007f153acd8ee0 in  () at /usr/lib/libglib-2.0.so.0
#29 0x00007f153acd9f62 in g_main_loop_run () at /usr/lib/libglib-2.0.so.0
#30 0x000055ef752007c6 in main (argc=<optimized out>, argv=<optimized out>) at …/src/awesome/src/build/awesome.c:884
```

I assume something in awesome itself is causing the recursion here.

blueyed:
Breaking at `signal_object_emit` shows these (repeating):
```
Thread 1 "awesome" hit Breakpoint 2, signal_object_emit (L=L@entry=0x40c08378, arr=0x55ef7523a3a8 <screen_class+8>, name=name@entry=0x4093faf8 "arrange", nargs=nargs@entry=1)
    at …/src/awesome/src/build/common/luaobject.c:232
Thread 1 "awesome" hit Breakpoint 2, signal_object_emit (L=L@entry=0x40c08378, arr=0x55ef7523a6a8 <client_class+8>, name=name@entry=0x55ef7522a276 "property::border_width", 
    nargs=nargs@entry=1) at …/src/awesome/src/build/common/luaobject.c:232
Thread 1 "awesome" hit Breakpoint 2, signal_object_emit (L=L@entry=0x40c08378, arr=0x55ef7523a6a8 <client_class+8>, name=name@entry=0x55ef7522bd96 "property::shape_bounding", 
    nargs=nargs@entry=1) at …/src/awesome/src/build/common/luaobject.c:232
Thread 1 "awesome" hit Breakpoint 2, signal_object_emit (L=L@entry=0x40c08378, arr=0x55ef7523a6a8 <client_class+8>, name=name@entry=0x55ef7522bd81 "property::shape_clip", 
    nargs=nargs@entry=1) at …/src/awesome/src/build/common/luaobject.c:232
Thread 1 "awesome" hit Breakpoint 2, signal_object_emit (L=L@entry=0x40c08378, arr=0x55ef7523a6a8 <client_class+8>, name=name@entry=0x55ef7522cbac "property::border_color", 
    nargs=nargs@entry=1) at …/src/awesome/src/build/common/luaobject.c:232
Thread 1 "awesome" hit Breakpoint 2, signal_object_emit (L=L@entry=0x40c08378, arr=0x55ef7523a6a8 <client_class+8>, name=name@entry=0x55ef7522a276 "property::border_width", 
    nargs=nargs@entry=1) at …/src/awesome/src/build/common/luaobject.c:232
Thread 1 "awesome" hit Breakpoint 2, signal_object_emit (L=L@entry=0x40c08378, arr=0x55ef7523a6a8 <client_class+8>, name=name@entry=0x55ef7522bd96 "property::shape_bounding", 
    nargs=nargs@entry=1) at …/src/awesome/src/build/common/luaobject.c:232
Thread 1 "awesome" hit Breakpoint 2, signal_object_emit (L=L@entry=0x40c08378, arr=0x55ef7523a6a8 <client_class+8>, name=name@entry=0x55ef7522bd81 "property::shape_clip", 
    nargs=nargs@entry=1) at …/src/awesome/src/build/common/luaobject.c:232
Thread 1 "awesome" hit Breakpoint 2, signal_object_emit (L=L@entry=0x40c08378, arr=0x55ef7523a3a8 <screen_class+8>, name=name@entry=0x4093faf8 "arrange", nargs=nargs@entry=1)
    at …/src/awesome/src/build/common/luaobject.c:232
```

`property::border_width` is emitted recursively:
```
#0  0x000055ef75214bb1 in signal_object_emit (L=L@entry=0x40c08378, arr=0x55ef7523a6a8 <client_class+8>, name=name@entry=0x55ef7522bd96 "property::shape_bounding", nargs=nargs@entry=1)
    at …/src/awesome/src/build/common/luaobject.c:232
#1  0x000055ef75213a05 in luaA_class_emit_signal (L=L@entry=0x40c08378, lua_class=<optimized out>, name=name@entry=0x55ef7522bd96 "property::shape_bounding", nargs=nargs@entry=1)
    at …/src/awesome/src/build/common/luaclass.c:321
#2  0x000055ef752150e0 in luaA_object_emit_signal (L=L@entry=0x40c08378, oud=oud@entry=-3, name=name@entry=0x55ef7522bd96 "property::shape_bounding", nargs=nargs@entry=0)
    at …/src/awesome/src/build/common/luaobject.c:309
#3  0x000055ef75217454 in luaA_client_set_shape_bounding (L=L@entry=0x40c08378, c=0x41c6f8c8) at …/src/awesome/src/build/objects/client.c:3527
#4  0x000055ef75213dc0 in luaA_class_newindex (L=0x40c08378) at …/src/awesome/src/build/common/luaclass.c:467
#5  0x00007f1538972316 in  () at /usr/lib/libluajit-5.1.so.2
#6  0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
#7  0x000055ef75214d0e in luaA_dofunction (nret=0, nargs=1, L=0x40c08378) at …/src/awesome/src/build/common/lualib.h:76
#8  0x000055ef75214d0e in signal_object_emit (L=L@entry=0x40c08378, arr=<optimized out>, name=name@entry=0x55ef7522a276 "property::border_width", nargs=nargs@entry=1)
    at …/src/awesome/src/build/common/luaobject.c:253
#9  0x000055ef75213a05 in luaA_class_emit_signal (L=L@entry=0x40c08378, lua_class=<optimized out>, name=name@entry=0x55ef7522a276 "property::border_width", nargs=nargs@entry=1)
    at …/src/awesome/src/build/common/luaclass.c:321
#10 0x000055ef752150e0 in luaA_object_emit_signal (L=L@entry=0x40c08378, oud=oud@entry=-3, name=name@entry=0x55ef7522a276 "property::border_width", nargs=nargs@entry=0)
    at …/src/awesome/src/build/common/luaobject.c:309
#11 0x000055ef75227608 in window_set_border_width (L=L@entry=0x40c08378, idx=idx@entry=-3, width=4) at …/src/awesome/src/build/objects/window.c:226
#12 0x000055ef7522769a in luaA_window_set_border_width (L=L@entry=0x40c08378, c=<optimized out>) at …/src/awesome/src/build/luaa.h:170
#13 0x000055ef75213dc0 in luaA_class_newindex (L=0x40c08378) at …/src/awesome/src/build/common/luaclass.c:467
#14 0x00007f1538972316 in  () at /usr/lib/libluajit-5.1.so.2
#15 0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
#16 0x000055ef75214d0e in luaA_dofunction (nret=0, nargs=1, L=0x40c08378) at …/src/awesome/src/build/common/lualib.h:76
#17 0x000055ef75214d0e in signal_object_emit (L=L@entry=0x40c08378, arr=<optimized out>, name=name@entry=0x55ef7522a276 "property::border_width", nargs=nargs@entry=1)
    at …/src/awesome/src/build/common/luaobject.c:253
#18 0x000055ef75213a05 in luaA_class_emit_signal (L=L@entry=0x40c08378, lua_class=<optimized out>, name=name@entry=0x55ef7522a276 "property::border_width", nargs=nargs@entry=1)
    at …/src/awesome/src/build/common/luaclass.c:321
#19 0x000055ef752150e0 in luaA_object_emit_signal (L=L@entry=0x40c08378, oud=oud@entry=-3, name=name@entry=0x55ef7522a276 "property::border_width", nargs=nargs@entry=0)
    at …/src/awesome/src/build/common/luaobject.c:309
#20 0x000055ef75227608 in window_set_border_width (L=L@entry=0x40c08378, idx=idx@entry=-3, width=0) at …/src/awesome/src/build/objects/window.c:226
#21 0x000055ef7522769a in luaA_window_set_border_width (L=L@entry=0x40c08378, c=<optimized out>) at …/src/awesome/src/build/luaa.h:170
#22 0x000055ef75213dc0 in luaA_class_newindex (L=0x40c08378) at …/src/awesome/src/build/common/luaclass.c:467
#23 0x00007f1538972316 in  () at /usr/lib/libluajit-5.1.so.2
#24 0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
#25 0x000055ef75215054 in luaA_dofunction (nret=0, nargs=1, L=0x40c08378) at …/src/awesome/src/build/common/lualib.h:76
#26 0x000055ef75215054 in luaA_object_emit_signal (L=L@entry=0x40c08378, oud=oud@entry=1, name=0x4093faf8 "arrange", nargs=nargs@entry=0)
    at …/src/awesome/src/build/common/luaobject.c:302
#27 0x000055ef7521521a in luaA_object_emit_signal_simple (L=0x40c08378) at …/src/awesome/src/build/common/luaobject.c:329
#28 0x00007f1538972316 in  () at /usr/lib/libluajit-5.1.so.2
#29 0x00007f15389b7fe0 in lua_pcall () at /usr/lib/libluajit-5.1.so.2
#30 0x000055ef75214d0e in luaA_dofunction (nret=0, nargs=0, L=0x40c08378) at …/src/awesome/src/build/common/lualib.h:76
#31 0x000055ef75214d0e in signal_object_emit (L=0x40c08378, arr=arr@entry=0x55ef7523a600 <global_signals>, name=name@entry=0x55ef7522a5fb "refresh", nargs=nargs@entry=0)
    at …/src/awesome/src/build/common/luaobject.c:253
#32 0x000055ef75209ca7 in luaA_emit_refresh () at …/src/awesome/src/build/luaa.c:1024
```

And breaking at `window_set_border_width` shows that it is called for two windows over and over again (`window` => `width`):
```
0x41c6f8c8 => 4
0x40b44128 => 0
0x41c6f8c8 => 0
0x41c6f8c8 => 4
0x40b44128 => 0
```

psychon:
Okay, so you have delayed call which... indexes some object with a key that is not set in C, but in Lua, so ends up in the index handler... that one again indexes some object and ends up calling another Lua handler. This now does a few calls in libluajit.

Your second backtrace is a refresh signal that emits `arrange` (in awesome, nothing connects to the `arrange` signal, so the following code must be in your config). The function connected to the arrange signal then changes the border width of some window, which in turns causes some Lua code to run (through the `property::border_width` signal) which then queries the clip shape of some window. This last part is most likely `awful.client.shape.update.all`.

So, what are you doing in response to `arrange` signals? What of these steps could make another arrange necessary... oh, changing the border width of a window causes another arrange. Perhaps something in there is "fluctuating" and not reaching a stable end state?

blueyed:
Likely related to this part of my config, which has an older pastebin already:

```
-- Handle border sizes of clients.
-- https://github.com/awesomeWM/awesome/issues/171#issuecomment-256143831
-- XXX: DBG: 100% CPU with two screens and cyclefocus with this?!
--      http://sprunge.us/dgLh
--      Elv13: Probably some signal -> connect -> emit loops. You can add a
--             printf() in the C code emit_signal or logs the calls using gdb
--             python breakpoints. Also try to log Lua emit_signal too if the C
--             ones aren't enough.
local skip = false
if not skip then
    local layout = require("awful.layout")
    local screen = require("awful.screen")
    local client = require("awful.client")

    function client.automatic_border_width(c, s)
        s = s or c.screen
        local layout_name = layout.getname(layout.get(s))
        -- No borders with only one humanly visible client
        -- print("automatic_border_width: " .. c.name)
        if c.marked then
            -- Handled separately.
            return
        elseif c.maximized then
            -- NOTE: also handled in focus, but that does not cover maximizing from
            -- a tiled state (when the client had focus).
            c.border_width = 0
        elseif c.floating or layout_name == "floating" then
            c.border_width = beautiful.border_width
        elseif layout_name == "max" or layout_name == "fullscreen" then
            c.border_width = 0
        else
            -- Get a list of other floating/non-tiled clients, without ones
            -- that are transient for the current client.
            -- This will not put a border around the tiled Thunderbird main
            -- window, for a "Confirm Deletion" floating (transient)
            -- confirmation window.
            local other_floating = {}
            for _, other_c in ipairs(s.clients) do
              if other_c ~= c then
                local p = other_c
                while true do
                  if p.transient_for == c then
                    break
                  end
                  if not p.transient_for then
                    other_floating[#other_floating] = other_c
                    break
                  end
                  p = p.transient_for
                end
              end
            end
            -- bnote("other_floating: "..#other_floating)

            -- No border for a single tiled client.
            -- But not in case of  "master_fill_policy" == "mwfact".
            --
            -- XXX: does not put border around only tiled client with (two) maxizized clients below!
            -- if #s.tiled_clients == 1 and (#other_floating == 0 or awful.tag.getproperty(c.first_tag, 'master_fill_policy') ~= 'mwfact') then
            if #s.tiled_clients == 1 and (awful.tag.getproperty(c.first_tag, 'master_fill_policy') ~= 'mwfact') then
                c.border_width = 0
                -- XXX: SLOW!  Not necessary anymore?!
                -- awful.client.moveresize(0, 0, 2, 0, tiled[1])
            else
                c.border_width = beautiful.border_width
            end
        end
        -- bnote('automatic_border_width: '..c.border_width..': '..c.name)
    end

    local client_border_width_for_s = function(s)
        -- timer.delayed_call(function (c)
          local clients = client.visible(s)
          for _, c in pairs(clients) do
              if c.valid then
                  client.automatic_border_width(c, s)
              end
          end
        -- end)
    end
        capi.screen[s]:connect_signal("arrange", client_border_width_for_s)
        -- XXX: might cause 100% CPU loop?!
        -- s:emit_signal("arrange")
    end)

    local client_border_maximized = function(c)
        c.border_width = not c.maximized and beautiful.border_width or 0
    end
    capi.client.connect_signal('property::maximized', client_border_maximized)

    -- Required when using (my) put_client, where a client becomes floating.
    local client_border_floating = function(c)
        c.border_width = not c.floating and beautiful.border_width or 0
    end
    capi.client.connect_signal('property::floating', client_border_floating)
end
```

psychon:
Uhm, which piece of code changes the border width in response to `property::border_width` and why would doing that be a good idea? This last back trace is: border width was changed to 0 and then is immediately changed back to 4.
(oh and all of this is getting quite a bit off-topic)

Hm... this might be `awful.placement` trying to enforce a border width of zero for fullscreen windows? And some code of yours that reacts to `arrange` by changing to a non-zero border width?

psychon:
That last piece of code will try to force `beautiful.border_width` if a window is fullscreened on a tag with a floating layout. I bet that `awful.placement` will fight against that.

So the loop would look like:

- (`arrange`) your code sets a non-zero border width (this triggers another arrange as a delayed call)
- (`property::border_size`) `awful.placement` forces the fullscreen client back to border size zero
- next `arrange`
- etc [...]

(And now: Good night)

blueyed:
Thanks for your help already.

I think this is quite on-topic, because it shows that internals might cause a dead-lock via nested delayed calls.
(It is also bad that awesome-client cannot be used on this, and does `kill -9` has to be used to stop it.  That is likely though a different issue, i.e. awesome should never get into an endless loop like this. )

`luaA_class_newindex` is emitted for/with `luaA_client_set_shape_bounding`, `luaA_client_set_shape_clip`, `luaA_window_set_border_color`, `luaA_window_set_border_width`.
(`shape.update.all` is connected to the `property::border_width` signal, and does not look at the current property/value, but sets it always - which is fine I guess assuming that `property::border_width` is only emitted on changes).
But `luaA_window_set_border_width` should not get called more than once probably, since then it should be set for the 2nd call already?  (I've just noticed that my default `awful.rules.rules` had basically `nil` for border_width, via some undefined variable - so that might explain why it gets called once at least; actually it looks like it never gets set on the object really?! (https://github.com/blueyed/awesome/blob/7fe4a8a102c83690deb9e31d9f288d3b88913599/objects/window.c#L541-L544) - so that might be good as-is / expected).

There are only 3 windows in the affected session: two on tag 5 (which is the visible/frozen one), which is a "maximized" layout, and one on tag 1 (which I cannot see).

It was triggered when using awesome-cyclefocus (alt-tab), and going from tag 1 to tag 5, and releasing it - the wibox from cyclefocus is still visible, but the screen only shows the tmux window from tag 1 still.  It is common for this issue to happen with regard to using cyclefocus.

The client for which this happens is another tmux window, which is focused in the titlebar.

None of the two windows on tag 5 is floating though, and therefore my handler should cause it to use `border_width = 0`.

`width=4` is used in the recursive `property::width` signal.

Maybe due to `awful.layout`'s handler?  (which uses `timer.delayed_call`)
https://github.com/blueyed/awesome/blob/7fe4a8a102c83690deb9e31d9f288d3b88913599/lib/awful/layout/init.lua#L239-L254
It uses a lock, but that should maybe get moved to the end / after emitting the arrange signal itself?  (https://github.com/blueyed/awesome/blob/7fe4a8a102c83690deb9e31d9f288d3b88913599/lib/awful/layout/init.lua#L224)

It is noteworthy that the second/inner `property::border_width` also triggers `luaA_window_set_border_color`, where `#268bd2` is used ("blue" from the solarized theme, also references as `theme.accent_color`), which I am using for `beautiful.border_focus` (set on `client::focus`, but the focus signal is not involved/emitted in this loop - it does not break for it).

So it must be some `property::border_width` handler that also sets `border_color` then.
I've found that cyclefocus itself does something in a callback to show the client, but from there 1 should be used for the border_width (https://github.com/blueyed/awesome/blob/7fe4a8a102c83690deb9e31d9f288d3b88913599/lib/awful/layout/init.lua#L224).

blueyed:
@psychon 
JFI: I am pretty certain that this is actually due to cyclefocus decorating the focused client (and it will use its original width which is 4).  Will investigate further later.

psychon:
> I think this is quite on-topic, because it shows that internals might cause a dead-lock via nested delayed calls.

We could deprecate (and later turn into an error) the whole mess:
```diff
diff --git a/lib/gears/timer.lua b/lib/gears/timer.lua
index c01b49f47..9fb231012 100644
--- a/lib/gears/timer.lua
+++ b/lib/gears/timer.lua
@@ -226,10 +226,13 @@ function timer.weak_start_new(timeout, callback)
 end
 
 local delayed_calls = {}
+local in_delayed_calls = false
 capi.awesome.connect_signal("refresh", function()
+    in_delayed_calls = true
     for _, callback in ipairs(delayed_calls) do
         protected_call(unpack(callback))
     end
+    in_delayed_calls = false
     delayed_calls = {}
 end)
 
@@ -239,6 +242,11 @@ end)
 -- @function gears.timer.delayed_call
 function timer.delayed_call(callback, ...)
     assert(type(callback) == "function", "callback must be a function, got: " .. type(callback))
+    if in_delayed_calls then
+        require("gears.debug").deprecate(
+            "Recursive delayed calls can cause various issues. You should not schedule a delayed call from inside a delayed call.",
+            { raw = true })
+    end
     table.insert(delayed_calls, { callback, ... })
 end
 
```
However, from a quick run through the test suite... well, no. "Everything explodes".
I also found a good reason against this PR in there:

- Taglist is updated "delayedly" which in turn causes a delayed call for the wibox redraw

Thus, with the commit from this PR, the taglist would not be updated until some "random later point in time" (whenever the next main loop iteration happens to be).

Similarly:

- `awful.autofocus` uses a delayed call to change the input focus. This in turn changes the background color of the titlebar, causing a redraw via a delayed call.

Again, with the commit from this PR, the titlebar is only redrawn at "some random later point in time".
(But I guess changing the input focus causes enough X11 events so that awesome actually wakes up shortly after)

psychon:
Another idea could be to stop (ab)using the `refresh` signal and move delayed calls into idle handler, i.e. `g_idle_add`. That would (if done correctly) make sure that awesome does not freeze if there are "endless delayed call loops". However, that would be an API break since there is code out there that assumes that delayed calls are run when emitting `refresh`.
Actually, I like that idea. I wanted to get rid of `refresh` anyway, but deprecating a signal is hard...

psychon:
Something like this (which causes a lot of test suite failures; I guess because lots of stuff actually does not want its delayed calls to be delayed):
```diff
diff --git a/lib/gears/timer.lua b/lib/gears/timer.lua
index c01b49f47..6ee439ecb 100644
--- a/lib/gears/timer.lua
+++ b/lib/gears/timer.lua
@@ -225,13 +225,8 @@ function timer.weak_start_new(timeout, callback)
     end)
 end
 
+local delayed_pending = false
 local delayed_calls = {}
-capi.awesome.connect_signal("refresh", function()
-    for _, callback in ipairs(delayed_calls) do
-        protected_call(unpack(callback))
-    end
-    delayed_calls = {}
-end)
 
 --- Call the given function at the end of the current main loop iteration
 -- @tparam function callback The function that should be called
@@ -240,6 +235,19 @@ end)
 function timer.delayed_call(callback, ...)
     assert(type(callback) == "function", "callback must be a function, got: " .. type(callback))
     table.insert(delayed_calls, { callback, ... })
+
+    if not delayed_pending then
+        glib.idle_add(glib.PRIORITY_DEFAULT, function()
+            delayed_pending = false
+            local calls = delayed_calls
+            delayed_calls = {}
+            for _, callback in ipairs(calls) do
+                protected_call(unpack(callback))
+            end
+            return false
+        end)
+        delayed_pending = true
+    end
 end
 
 function timer.mt.__call(_, ...)
```

blueyed:
> g_idle_add

Sounds good!

I've worked a bit on cyclefocus to make it behave better (and used e.g. a delayed_call therein via a normal timer with timeout=0 to detect if there is (not) a "arrange" loop).

I could not reproduce the hard freeze initially (but still have the gdb session from where it happened): maybe even just due to printing/emitting notifications - it was looping there endlessly.  Maybe the moment where the keygrabber is exited is important etc.

actionless:
what is the current status of this issue?

PR #1973, opened 2017-08-10
Tests: check-integration: error on warnings

blueyed:
Does it make sense to have the `-W` option, or should we just adjust the pattern always?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -395,7 +395,7 @@ add_executable(test-gravity tests/test-gravity.c)
 target_link_libraries(test-gravity
     ${AWESOME_COMMON_REQUIRED_LDFLAGS} ${AWESOME_REQUIRED_LDFLAGS})
 add_custom_target(check-integration
-    sh -c "CMAKE_BINARY_DIR='${CMAKE_BINARY_DIR}' ${CMAKE_SOURCE_DIR}/tests/run.sh"
+    sh -c "CMAKE_BINARY_DIR='${CMAKE_BINARY_DIR}' ${CMAKE_SOURCE_DIR}/tests/run.sh -W"
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
     COMMENT "Running integration tests"
     USES_TERMINAL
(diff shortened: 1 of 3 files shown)


blueyed:
@psychon 
What do you think about having this by default, without the flag?
But apart from that, it is nice to have getopts parsing here, which could replace `VERBOSE` etc later (but does not need to have `-W` here).

psychon:
> What do you think about having this by default, without the flag?

The last time we did "all warnings are fatal", someone reported that his fontconfig config was broken, fontconfig complained and our test machinery counted that as a failure. What are the odds that something like this will happen again?

> But apart from that, it is nice to have getopts parsing here, which could replace VERBOSE etc later (but does not need to have -W here).

Why is `getopt` parsing nicer than env vars? Either one are pretty much equally easy to use, but env vars are easier to implement.

blueyed:
> The last time we did "all warnings are fatal", someone reported that his fontconfig config was broken, fontconfig complained and our test machinery counted that as a failure.

That was about generic warnings / output, wasn't it?
In this case we're matching only awesome's warnings.
Also, it happened with CMake (example tests) IIRC, which makes the build / generating the build files fail - vs. making `check-integration` (to be run explicitly) fail.

> Why is `getopt` parsing nicer than env vars?

It feels more natural to me to pass options instead of env variables, where you never know if they are really used.
This is more the case with something generic like `VERBOSE`, whereas `AWESOME_TESTS_VERBOSE` would be more explicit.

I am OK with merging this as-is for now later.

psychon:
> In this case we're matching only awesome's warnings.

Okay.

> where you never know if they are really used.

Your argument parsing silently ignores unknown arguments, doesn't it? ;-)

blueyed:
> Your argument parsing silently ignores unknown arguments, doesn't it? ;-)

Indeed.  Fixed and improved.

blueyed:
I've thought about improving this to have a single test runner again (removing `tests/themes/run.sh`) - possibly by looking for `test-*.lua` also in subdirs, adding support for `setup-test.sh` in there (that would change the config) and then having a separate subdir for each of the themes.
The downside is that themes are not globbed anymore (and new ones have to be added).

The main motivation for this is to being able to only run a single test file again, without running the theme tests always.. for this we could also make `check-integration` not depend on `check-themes` anymore, but call `check-themes` explicitly on Travis?!

psychon:
> without running the theme tests always

Uhm, why is it a bad thing to run these tests?

blueyed:
> Uhm, why is it a bad thing to run these tests?

It is distracting / not necessary when you only want to run a single file.

psychon:
Yes indeed & okay, sorry. I misunderstood your intention then.

blueyed:
I think it is easier to keep `check-themes` separate for now.

PR #1974, opened 2017-08-11
CMakeLists.txt: add DEPENDS for check-integration

blueyed:
This makes `make check-integration` work on a clean build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -387,6 +387,7 @@ add_custom_target(check-integration
     ${CMAKE_COMMAND} -E env CMAKE_BINARY_DIR='${CMAKE_BINARY_DIR}' ./tests/run.sh \$\${TEST_RUN_ARGS:--W}
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
     COMMENT "Running integration tests"
+    DEPENDS ${PROJECT_AWE_NAME}
     USES_TERMINAL)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -403,6 +403,7 @@ add_custom_target(check-integration
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
     COMMENT "Running integration tests"
     USES_TERMINAL
+    DEPENDS ${PROJECT_AWE_NAME}
```
blueyed: Should have more probably, similar to `check-themes`?
https://github.com/awesomeWM/awesome/pull/1967/files#diff-af3b638bc2a3e6c650974192a53c7291R413

blueyed:
Rebased.
Any feedback on https://github.com/awesomeWM/awesome/pull/1974#pullrequestreview-55731228?

CMakeLists.txt
psychon: Yup, `generated_icons generate_awesomerc` definitely are needed by the integration tests.
I'd propose to just merge this one commit and if we stumble over more missing deps in the future, we will add them. (And perhaps refactor everything so that there is a target that depends on "everything needed" and where check-themes, check-integration etc just depend on)

actionless:
reopened to re-trigger stucked codacy status

blueyed:
This was added in 579159a8c already.

PR #1978, opened 2017-08-12
Add an alternative awful.tooltip example

nicodebo:
Added an awful.tooltip example that shows how to use mouse::enter signal
to update the text properties of the tooltip.

Signed-off-by: nicodebo <nico.debo@openmailbox.org>

Note: I didn't use the same args as the first exemple because they don't seem to be valid anymore and return some errors. I was not sure if I should fix or if it was in the scope of this pull request so I left it.

### DIFF
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -18,6 +18,16 @@
 --             end,
 --         })
 --
+--     -- alternatively, you can use mouse::enter signal
+--     myclock = wibox.widget.textclock("%T", 1)
+--     myclock_t = awful.tooltip({
+--         objects = { myclock }
+--         })
+--
+--     myclock:connect_signal("mouse::enter", function()
+--             myclock_t.text = os.date("Today is %A %B %d %Y\nThe time is %T")
+--     end)
+--
 -- How to add the same tooltip to multiple objects?
 -- ---
 --


psychon:
How about the following instead:
```lua
myclock = wibox.widget.textclock("%T", 1)
myclock_t = awful.tooltip({
    objects = { myclock }
})

myclock_t:connect_signal("property_visible", function()
    if myclock_t.visible then
        myclock_t.text = os.date("Today is %A %B %d %Y\nThe time is %T")
    end
end)
```
Compared to your proposed version, this connects to a signal on the tooltip and thus this also works if the tooltip is connected to more than one widget. I'm not sure if this version would really be much better and in fact I cannot remember a tooltip which was connected to more than one widget....

nicodebo:
I couldn't make your version work, so I assume you meant "property::visible".
The api is so rich, yet, from my point of view, lacks examples. Maybe both versions could be added and explain the difference between the original, yours and mine ?

Elv13:
yes, it's `property::visible`. I agree it needs more examples. I am working on this in a branch.

https://elv13.github.io/libraries/awful.popup.html#

(in that branch, popup is the tooltip superclass)

Once that work lands into Awesome, it will have a much higher code example density. Previously, our automated code example framework (the thing that validates the example and generate the images) could not handle wibox based widgets.

nicodebo:
Thanks for the heads-up. I'm looking forward to it.
Let me know if you want me to enhance this pull request like proposed in the fourth message. Otherwise feel free to close it.

actionless:
@Elv13 ?

actionless:
thanks for the clarification!

Issue #1667, opened 2017-03-19
awful.spawn callback not called


bang-uin:
Output of `awesome --version`:

awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.0

**How to reproduce the issue:**

Add following code and execute.
`awful.spawn("xterm", true, function(c) naughty.notify({text = "Callback"}) end)`

**Actual result:**

xterm executes but the callback function is not called.

**Expected result:**

Callback function should be called.

https://github.com/awesomeWM/awesome/blob/v4.0/lib/awful/spawn.lua#L251
Adds the callback to snid_buffer which in the end should be executed here
https://github.com/awesomeWM/awesome/blob/v4.0/lib/awful/rules.lua#L510
and (for Lua 5.1) here
https://github.com/awesomeWM/awesome/blob/v4.0/lib/gears/protected_call.lua#L31

Not sure where the problem is as it looks ok to me.

Elv13:
Right, `xterm` doesn't support the necessary protocols for callbacks to work. Please use `urxvt` or any `libvte` based terminals (terminator, gnome-terminal, XFCE terminal, etc) instead of `xterm`. `urxvt` is better in every way compared to `xterm`. It is faster, better maintained and has more options/extensions.

There is nothing we can do about xterm, the problem is on "their" side (where "their" is no one, because it's not under active development since 20 years). Closing this.

psychon:
> Not sure where the problem is as it looks ok to me

To add a bit more information here: xterm has `c.startup_id == nil`, i.e. it does not read the necessary environment variable and turns it into a property on its window. That way, we cannot detect that "this new window" belongs to your call to `spawn`.

bang-uin:
Hmm, i tested different applications they all didn't work.
But i see your point. Relying on startup_id is not reliable.

Maybe one of you could give me a hint.
I try to implement a session save/restore feature.
The idea is to save all clients (screens, tags, layout properties) on client appearance/logout and restore them after login. Just like a simple traditional session manager.
Starting the programs works as expected (as long as they have a valid PID) but to restore their layout positions after spawn i need an stable identifier before calling spawn which gets passed to the client and therefore can be used in the manage callback or in awful rules.
[...]

PR #1979, opened 2017-08-12
Allow setting environment vars for spawned processes

psychon:
See #1667 for how this happened (more or less, I'm not completely sure?!?).

Note that this PR only does the necessary changes on the C side so that an environment for spawned processes can be set. Nothing is done on the Lua side and somehow I feel that just "tacking things" onto the existing spawn API ends up being complicated. Also note that *only* the provided environment is used, so nothing else is inherited. Thus, the following snipped of code might be needed:
```lua
glib = require("lgi").GLib
local env = {}
for _, k in ipairs(glib.listenv()) do
    env[k] = glib.getenv(k)
end
```
(The above does the equivalent of lines 441 to 463 in https://github.com/bang-uin/awesome/commit/9e8ea1bce201be624999e527b9102e7bf73ad74c#diff-7d165d612dceb08d572a0ad1fb250565R441 but instead of 20 lines of codes (with memory leaks), it's just 5 (and we can be sure that there are no memory leaks))

### DIFF
--- a/spawn.c
+++ b/spawn.c
@@ -369,6 +369,48 @@ spawn_callback(gpointer user_data)
         unsetenv("DESKTOP_STARTUP_ID");
 }
 
+/** Convert a Lua table of string to a char** array.
+ * \param L The Lua VM state.
+ * \param idx The index of the table that we should parse
+ * \return The argv array.
+ */
+static gchar **
+parse_table_array(lua_State *L, int idx, GError **error)
+{
+    gchar **argv = NULL;
+    size_t i, len;
+
+    luaL_checktype(L, idx, LUA_TTABLE);
+    idx = luaA_absindex(L, idx);
+    len = luaA_rawlen(L, idx);
+
+    /* First verify that the table is sane: All integer keys must contain
+     * strings. Do this by pushing them all onto the stack.
+     */
+    for (i = 0; i < len; i++)
+    {
+        lua_rawgeti(L, idx, i+1);
+        if (lua_type(L, -1) != LUA_TSTRING)
+        {
+            g_set_error(error, G_SPAWN_ERROR, 0,
+                    "Non-string argument at table index %zd", i+1);
+            return NULL;
+        }
+    }
+
+    /* From this point on nothing can go wrong and so we can safely allocate
+     * memory.
+     */
+    argv = g_new0(gchar *, len + 1);
+    for (i = 0; i < len; i++)
+    {
+        argv[len - i - 1] = g_strdup(lua_tostring(L, -1));
+        lua_pop(L, 1);
+    }
+
+    return argv;
+}
+
 /** Parse a command line.
  * \param L The Lua VM state.
  * \param idx The index of the argument that we should parse
@@ -378,7 +420,6 @@ static gchar **
 parse_command(lua_State *L, int idx, GError **error)
 {
     gchar **argv = NULL;
-    idx = luaA_absindex(L, idx);
 
     if (lua_isstring(L, idx))
     {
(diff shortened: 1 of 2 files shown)


bang-uin:
Sorry, but 

> (The above does the equivalent of lines 441 to 463 in bang-uin/awesome@9e8ea1b#diff-7d165d612dceb08d572a0ad1fb250565R441 but instead of 20 lines of codes (with memory leaks), it's just 5 (and we can be sure that there are no memory leaks))

is not correct. 
1. The code does not the same as lines 441 to 463. The code above replaces 10 lines by 5, not 20. 
2. There was no leak introduced by my code. As envp is freed up in all cases. Please show me the introduced leak.

spawn.c
```diff
@@ -369,6 +369,48 @@ spawn_callback(gpointer user_data)
         unsetenv("DESKTOP_STARTUP_ID");
 }
 
+/** Convert a Lua table of string to a char** array.
```
blueyed: s/string/strings

spawn.c
```diff
@@ -369,6 +369,48 @@ spawn_callback(gpointer user_data)
...
+ * \param L The Lua VM state.
+ * \param idx The index of the table that we should parse
```
blueyed: Trailing dot.

spawn.c
```diff
@@ -388,31 +429,13 @@ parse_command(lua_State *L, int idx, GError **error)
...
-            argv[len - i - 1] = g_strdup(lua_tostring(L, -1));
-            lua_pop(L, 1);
-        }
+        argv = parse_table_array(L, idx, error);
     }
     else
     {
-        luaL_error(L, "Invalid argument to spawn(), expect string or table");
+        g_set_error_literal(error, G_SPAWN_ERROR, 0,
+                "Invalid argument to spawn(), expect string or table");
```
blueyed: s/expect/expected/ (not new, but you are changing it anyway)

spawn.c
```diff
@@ -452,18 +475,20 @@ child_exit_callback(GPid pid, gint status, gpointer user_data)
  * @tparam[opt=nil] function exit_callback Function to call on process exit. The
  * function arguments will be type of exit ("exit" or "signal") and the exit
  * code / the signal number causing process termination.
+ * @tparam[opt=nil] table cmd The environment to use for the spawned program.
+ * Without this, the spawned process inherits awesome's environment.
```
blueyed: I prefer to have two spaces of extra indentation for continued lines.
It's not the case for other params, but could be done for the new one.

I think there should be no comma after "Without this".

spawn.c
psychon: Also fixed the same problem on `parse_command`.

psychon:
Done. Did this as a new commit since these were not new problems, but already existing ones.

psychon:
@bang-uin Sorry, I did not mean to badmouth you. I was just trying to point out why doing this in Lua might be easier/better than doing it in C. Sorry if you feel offended.

> There was no leak introduced by my code. As envp is freed up in all cases. Please show me the introduced leak.

Okay, call `awesome.spawn({"foo", "bar, "baz"}, false, false, false, false, false, { {} })` (No idea if all the `false` are right, but the two tables are what I want to get to). First, `parse_command()` is called for the first table and copies it into `argv`. Parsing `{ {} }` then fails because `{}` is not a string (call to `luaL_error`). Lua `longjmp()`s away and the copy of `{ "foo", "bar", "baz" }` in `gchar **argv` is leaked.

bang-uin:
> Lua `longjmp()`s away and the copy of { "foo", "bar", "baz" } in `gchar **argv` is leaked.

That's indeed true. I apologize, my fault.

blueyed:
How do you feel about covering the error cases in a test?
https://codecov.io/gh/awesomeWM/awesome/compare/8166e2ff8b567059459076f052cfe3863b616839...eeebdd8c985c2070239d900b66001afe23bf9567/diff

psychon:
Does "feel free" count as an answer? ;-)
Now that this no longer throws Lua errors, but returns error messages instead, that would be a lot easier.

blueyed:
I'll look into adding tests.

tests/test-spawn.lua
```diff
@@ -71,6 +71,19 @@ local steps = {
                                      end
                                  })
 
+            -- Test that setting env vars works and that the env is cleared
+            local read_line = false
+            local pid, err, _, stdout = awesome.spawn({ "sh", "-c", "echo $AWESOME_SPAWN_TEST_VAR $HOME $USER" },
```
blueyed: `awesome.spawn` returns:
```
 * @treturn[1] integer Process ID if everything is OK.
 * @treturn[1] string Startup-notification ID, if `use_sn` is true.
 * @treturn[1] integer stdin, if `stdin` is true.
 * @treturn[1] integer stdout, if `stdout` is true.
 * @treturn[1] integer stderr, if `stderr` is true.
 * @treturn[2] string An error string if an error occurred.
```

psychon: Whoops. Thanks for noticing and fixing!

blueyed:
Ok?

I would squash https://github.com/awesomeWM/awesome/pull/1979/commits/0c4129579092f1e06f5c5d943b743ccf91a6457a, https://github.com/awesomeWM/awesome/pull/1979/commits/c68674d3e4321b271854fc56a3b6dff15cdda50a and the new tests - also the fixup of course.

psychon:
Ok!

PR #1981, opened 2017-08-12
Fix an old todo in the docs

psychon:
This new paragraph might not be great and it certainly is less
informative than what the wiki said here (it suggested to use some
library for widgets and explained how to use it). However, this new
paragraph is definitely better than the todo that was here for far too
long.

### DIFF
--- a/docs/07-my-first-awesome.md
+++ b/docs/07-my-first-awesome.md
@@ -171,7 +171,28 @@ A new window will appear and a new awesome instance runs inside of it.
 
 ## Add widgets
 
-TODO: Write this section.
+Before the following line:
+
+    mylauncher = awful.widget.launcher({ image = beautiful.awesome_icon,
+
+Add this to create a new widget:
+
+    praisewidget = wibox.widget.textbox()
+    praisewidget.text = "You are great!"
+
+The widget still needs to be added to a wibar. For this, find the definition of
+your wibar and change it as follows:
+
+    { -- Left widgets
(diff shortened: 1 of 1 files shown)


docs/07-my-first-awesome.md
```diff
@@ -171,7 +171,28 @@ A new window will appear and a new awesome instance runs inside of it.
...
 
-TODO: Write this section.
+Before the following line:
+
+    mylauncher = awful.widget.launcher({ image = beautiful.awesome_icon,
+
+Add this to create a new widget:
+
+    praisewidget = wibox.widget.textbox()
+    praisewidget:set_text("You are great!")
```
Elv13: Could it use `.text`? The issue I see is that `:set_text()` is not explicitly documented and this could cause some confusion to new users (the intended audience of this article)

psychon: Amended in the following:
```diff
diff --git a/docs/07-my-first-awesome.md b/docs/07-my-first-awesome.md
index 7a2aac8d4..b5845add1 100644
--- a/docs/07-my-first-awesome.md
+++ b/docs/07-my-first-awesome.md
@@ -178,7 +178,7 @@ Before the following line:
 Add this to create a new widget:
 
     praisewidget = wibox.widget.textbox()
-    praisewidget:set_text("You are great!")
+    praisewidget.text = "You are great!"
 
 The widget still needs to be added to a wibar. For this, find the definition of
 your wibar and change it as follows:
```

PR #1983, opened 2017-08-13
build-utils/dist.sh: fix shellcheck warnings and add usage

blueyed:
[ci skip]

### DIFF
--- a/build-utils/dist.sh
+++ b/build-utils/dist.sh
@@ -4,20 +4,23 @@
 
 set -e
 
+if [ -z "$1" ]; then
+    echo "Usage: $0 VERSION"
+    exit 64
+fi
+
 VERSION=$1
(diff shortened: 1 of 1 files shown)


psychon:
Could you try re-building an old release tarball and check that you get (pretty much) the same result? If so: :+1:

blueyed:
@psychon 
I see your valid concern, but raise an "it would be better if a 3rd party would do just that" - to keep errors from my side out of the equation.. ;)

psychon:
Well, okay. However, then you'll have to ping me here in a month since I won't get to do it before then (needs a computer with a pgp key, thus I can only do it at home and I'll be travelling for the next weeks).

blueyed:
Looked into it a bit more, and now it will produce the same md5 sum for the archives.

PR #1984, opened 2017-08-13
luacov: codefromstrings=true

blueyed:


blueyed:
This was just experimental; I've expected it to increase if anything.
Coveralls does not seem to handle it at all?!

PR #1990, opened 2017-08-14
Naughty rewrite, part 1

### LINKED ISSUE #1683: popup widget / widget in tooltip

### LINKED ISSUE #1874: Suggestion for new naughty actions design
I wanted to change the design of actions in naughty notifications to a for me more pleasant looking design;

Naughty with Actions

Using the default theme without any modifications it might look like:

out

Unfortunately naughty doesn't provide a hook to create own wiboxes and let naughty take care of the rest.
That's why I needed to copy over naughty completely to my config and modify the config. Since I would really like not to copy core config over to my config, I wanted to ask for your opinion on the design, and whether you think this could be part of the core config? In that case I would create a pull request for it.

If this couldn't make it in the core config, do you alternatively think a hook to create the wiboxes might be a good idea? That hook could get the preprocessed args and need to return the wibox for that notification. That way the user would have full control over the look, but still get the placement, timeouting, etc. from the naughty core.

Elv13:
So, this is it. I am mostly done with the new implementation. This pull request is almost unrelated, but has some commits that are necessary at some point for future parts of this pull request series.

* Part 1: This pull request, all the unrelated commits (16 commits)
* Part 2: Improve `wibox.layout` spacing (3 commits)
* Part 3: Add an `awful.popup` widget (2 commits, fix #1683)
* Part 4: Refactor `awful.widget.common` to make it externally extensible (3 commits)
* Part 5: The `naughty` rewrite/refactoring (~10 commits or one huge commit if squashed)
* Part 6: Add new notification widgets and use them in `rc.lua` (7 commits, fix #1874, #1862)

(I dropped the `tests/examples` rewrite from this series since the recent CI improvements introduced by @blueyed make it tolerable for little while longer)

Why all those changes:

The old notification system is a black box and it's very hard to change its behavior/look without forking the whole module (as @mindeunix and @timroes figured out). The `taglist` and `tasklist` also have the same issue. This pull request also adds more documentation and code examples around the `wiboxes` derived widgets.

The goal of the new system is to be as easy to customize as the titlebar and wibars are in the current `rc.lua`.

This pull request:

 * Add a simple `separator` widget. It will make sense in part 2, but had no dependencies, so I decided to put it in this PR
 * Update some images templates to support wiboxes and use more "real" code instead of shims
 * Allow to turn wibox into widgets (for testing and examples)
 * Improve the shims to cover a larger part of CAPI
 * Improve `gears.object` signal system to behave like the CAPI version
 * Add some minimal tests to `naughty` before f**cking everything up again.

### DIFF
--- a/lib/gears/object.lua
+++ b/lib/gears/object.lua
@@ -56,6 +56,11 @@ function object:connect_signal(name, func)
     sig.strong[func] = true
 end
 
+-- Register a global signal receiver.
+function object:_connect_everything(callback)
+    table.insert(self._global_receivers, callback)
+end
+
 local function make_the_gc_obey(func)
     if _VERSION <= "Lua 5.1" then
         -- Lua 5.1 only has the behaviour we want if a userdata is used as the
@@ -119,6 +124,9 @@ function object:emit_signal(name, ...)
     end
     for func in pairs(sig.weak) do
         func(self, ...)
+    end
+    for _, func in ipairs(self._global_receivers) do
+        func(name, self, ...)
     end
 end
 
@@ -187,6 +195,8 @@ local function new(args)
     end
 
     ret._signals = {}
+
+    ret._global_receivers = {}
 
     local mt = {}
 
--- a/tests/examples/naughty/colors.lua
+++ b/tests/examples/naughty/colors.lua
@@ -1,0 +1,19 @@
+
+local beautiful = require("beautiful") --DOC_HIDE
+
+local text = [[An <b>important</b>
+<i>notification</i>
+]]
+
+require("naughty").notify {
+    title        = "Hello world!",
+    text         = text,
+    icon         = beautiful.icon,
+    bg           = "#0000ff",
+    fg           = "#ff0000",
+    fond         = "verdana 14",
+    border_width = 1,
+    border_color = "#ff0000"
+}
+
+--DOC_HIDE vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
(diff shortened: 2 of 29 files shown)


psychon:
Since I am imaging that my review is requested for this PR: Sorry, this looks too time consuming currently. Feel free to ping me, but for now I'll have to ignore this PR.

tests/examples/naughty/colors.lua
```diff
@@ -0,0 +1,19 @@
...
+<i>notification</i>
+]]
+
+require("naughty").notify {
+    title        = "Hello world!",
+    text         = text,
+    icon         = beautiful.icon,
+    bg           = "#0000ff",
+    fg           = "#ff0000",
+    fond         = "verdana 14",
```
actionless: typo in `fond` and also Verdana could be easily missing in the system, mb better just Sans?

Elv13: Done

Elv13:
Closing this so in favor of smaller PRs

lib/gears/object.lua
```diff
@@ -56,6 +56,11 @@ function object:connect_signal(name, func)
     sig.strong[func] = true
 end
 
+-- Register a global signal receiver.
+function object:_connect_everything(callback)
+    table.insert(self._global_receivers, callback)
+end
```
Elv13: It would be cool, but this would require to add the code to `emit_signal`[1]. That's already one of the "hottest" code path in Awesome. Wildcard parsing, even with a compiled state machine, is very slow compared to doing nothing at all. That function is called up to a thousand time per event loop. Also, this isn't really related to this patch.

[1] There is no "list of signals", every object can emit anything.

psychon:
> Closing this so in favor of smaller PRs

Thanks a lot!

PR #1991, opened 2017-08-14
tests: keep check-themes separate from check-integration

blueyed:
Ref: https://github.com/awesomeWM/awesome/pull/1973#issuecomment-322183782

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -185,7 +185,7 @@ script:
       travis_run_in_fold "make.install" sudo env PATH=$PATH make install
       awesome --version
     fi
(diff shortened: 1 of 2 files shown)


psychon:
Okay, now I'm confused again. I thought this was about running `tests/run.sh test-foo.lua` by hand, but this PR only touches `.travis.yml` and CMake things...?!?

blueyed:
> I thought this was about running tests/run.sh test-foo.lua by hand

More about `make check-integration TEST_RUN_ARGS=test-foo.lua`.

PR #1992, opened 2017-08-14
Add two new gears.string functions: startswith, endswith

Veratil:
This adds the `startswith` and `endswith` functions to `gears.string`, and changes `awful.completion` to use `gears.string.startswith` instead of the local function `str_starts`.

Allows more exposure to a useful common string function.

### DIFF
--- a/lib/gears/string.lua
+++ b/lib/gears/string.lua
@@ -106,4 +106,22 @@ function gstring.split(str, delimiter)
     return result
 end
 
+--- Check if a string starts with another string
+-- @class function
+-- @name startswith
+-- @tparam string str String to search
+-- @tparam string sub String to check for
+function gstring.startswith(str, sub)
+	return string.sub(str, 1, string.len(sub)) == sub
+end
+
+--- Check if a string ends with another string
+-- @class function
+-- @name endswith
+-- @tparam string str String to search
+-- @tparam string sub String to check for
+function gstring.endswith(str, sub)
+	return sub == "" or string.sub(str,-string.len(sub)) == sub
+end
+
 return gstring
--- a/lib/awful/completion.lua
+++ b/lib/awful/completion.lua
@@ -21,6 +21,7 @@ local pairs = pairs
 local string = string
 
 local gears_debug = require("gears.debug")
+local gstring = require("gears.string")
 
 local completion = {}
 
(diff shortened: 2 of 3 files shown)


lib/gears/string.lua
```diff
@@ -106,4 +106,22 @@ function gstring.split(str, delimiter)
...
+	return string.sub(str, 1, string.len(sub)) == sub
+end
+
+--- Check if a string ends with another string
+-- @class function
+-- @name endswith
+-- @tparam string str String to search
+-- @tparam string sub String to check for
+function gstring.endswith(str, sub)
+	return sub == "" or string.sub(str,-string.len(sub)) == sub
```
psychon: I'd like to see some tests for this (I'm writing this comment instead of asking why `sub == ""` is necessary, which made me notice that things are not as simple as I initially thought)

Veratil: `endswith("something", "") == true`

Without the `sub == ""` check then `string.sub(str,-string.len(sub)) == sub` returns false.

Veratil: That's honestly the only reason why it's there.

actionless:
mb to use `lua-penlight` for such things? we depend on it for `luadoc` anyway

http://stevedonovan.github.io/Penlight/api/libraries/pl.stringx.html#startswith

Veratil:
> mb to use lua-penlight for such things? we depend on it for luadoc anyway

`lua-penlight` is useful only to those who build from source, though, it's not a runtime requirement.

actionless:
yup, you right
but also, since penlight is pure lua, it's also possible to ship static copy of it as `awful.pl` or whatever
however idk how much use of it we can make

please count my comment not as objection to merging your PR but just a topic to discuss unconditionally to this PR, it only gave me that idea

Veratil:
Oh I'm not counting it as an objection. Just saying not everyone has lua-penlight actually installed. :)

blueyed:
How do you feel about adding some tests for it then?

Veratil:
> How do you feel about adding some tests for it then?

I don't see why we need to, honestly. There's literally one edge case for `endswith` which just requires the `sub == ""`.

```lua
> gstr = require('gears.string')
> gstr.startswith("something", "")
true
> gstr.startswith("something", "some")
true
> gstr.startswith("something", "none")
false
> gstr.endswith("something", "")
true
> gstr.endswith("something", "thing")
true
> gstr.endswith("something", "that")
false
```

PR #1993, opened 2017-08-15
[WIP] Travis: add cpp-coveralls

blueyed:
Taken out of https://github.com/awesomeWM/awesome/pull/1907, to see how it behaves on master.

https://github.com/eddyxu/cpp-coveralls
https://github.com/moteus/luacov-coveralls#test-lua-module-written-on-lua-and-c-using-cpp-coveralls

TODO:

- [ ] revisit/rebase after https://github.com/awesomeWM/awesome/pull/2574
- [ ] location for build/themes is off (https://coveralls.io/builds/21000624), also with Code Climate (#2564, https://codeclimate.com/repos/5c35e985236f14026e00ceb2/settings/test_reports/5c376a59f06c8848640004a3)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,7 +98,11 @@ install:
 
   # Install dependencies for code coverage testing.
   - if [ -n "$DO_COVERAGE" ]; then sudo luarocks install cluacov; fi
-  - if [ "$DO_COVERAGE" = "coveralls" ]; then sudo luarocks install luacov-coveralls; fi
+  - |
+    if [ "$DO_COVERAGE" = "coveralls" ]; then
+      sudo luarocks install luacov-coveralls
+      pip3 install cpp-coveralls
+    fi
 
   # Determine custom version.
   - export AWESOME_VERSION="${TRAVIS_BRANCH}-g$(git rev-parse --short HEAD)"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -203,7 +203,19 @@ script:
         make check-integration || exit 1
         make check-themes || exit 1
         do_codecov functionaltests
-        do_codecov_gcov c_code
+
+        # Run gcov.  Upload for codecov directly, gets done later for coveralls.
+        # NOTE: files for the main target have the build dir as prefix, but
+        #       otherwise it is the source directory?!
```
blueyed: Any idea about this?
Is it relevant where the binary is run from?
Probably rather where it is built from?!

psychon: I'm not really sure what this comment means. What is "otherwise"? Which files is this about?

According to https://gcc.gnu.org/onlinedocs/gcc-4.7.3/gcc/Gcov-Data-Files.html#Gcov-Data-Files, the same directory as the object file is used, so build dir. However, I guess that if the compiler is invoked with a relative path, then the program ends up using a relative path as well...? I'm not really sure since I don't fully grasp what the "symptoms" are.

On an unrelated note: You recently opened quite a number of PRs and it felt to me that some of them were still WIP, so I ignored them. If you feel like some PR is "finally done", I would suggest to add a comment saying so, so that everyone knows about it. I'm mostly working based on GitHub notifications and so for me PRs can easily "fall through the cracks".

blueyed: > the same directory as the object file is used, so build dir. However, I guess that if the compiler is invoked with a relative path, then the program ends up using a relative path as well...?

The object files in `build//CMakeFiles/awesome.dir` seem to have `build/` prefix, while the files in `build//CMakeFiles/{lgi-check.dir,test-gravity.dir}` do not have it.

> I'm mostly working based on GitHub notifications and so for me PRs can easily "fall through the cracks".

I see.  I'll request your review in case it takes to long.
Also, I will make sure to add WIP preffixed more often.

psychon: `lgi-check` is created via `add_executable(lgi-check build-utils/lgi-check.c)`, i.e. a source-relative path. Instead, for the awesome executable we use absolute paths to the pre-processed `.c` files in the build dir (so relative to the source dir this means: there is an additional `build/` prefix).

Also: Thanks :-)

blueyed:
Integrating into https://github.com/awesomeWM/awesome/pull/2672.

PR #1994, opened 2017-08-15
tests/themes/run.sh: use themes from source dir

blueyed:
This makes it easier to track their coverage.

### DIFF
--- a/tests/themes/run.sh
+++ b/tests/themes/run.sh
@@ -6,14 +6,13 @@
 
 set -e
 
-if ! [ -f CMakeCache.txt ]; then
-  echo 'This should be run from the build directory (expected CMakeCache.txt).' >&2
+if ! [ -f CMakeLists.txt ]; then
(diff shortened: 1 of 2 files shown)


tests/themes/run.sh
```diff
@@ -6,14 +6,13 @@
...
+if ! [ -f CMakeLists.txt ]; then
+  echo 'This should be run from the source directory (expected CMakeLists.txt).' >&2
   exit 64
 fi
-build_dir="$PWD"
-tests_dir="$(dirname -- "$0")/.."
+source_dir="$PWD"
 
-config_file="$build_dir/test-themes-awesomerc.lua"
+config_file="$(mktemp)"
```
psychon: Shouldn't this temporary file be deleted somewhere?

blueyed: > Shouldn't this temporary file be deleted somewhere?

Done, thanks.

psychon:
(Yes, this means I also looked at this, but e.g. I did not read up on what `VERBATIM` does)

PR #1995, opened 2017-08-15
Travis: build luarocks in /tmp also

blueyed:


blueyed:
> Before this, luarocks was build in the source dir, right?

Yes.

PR #1998, opened 2017-08-16
Improve docs for emit_signal_recursive

psychon:


lib/wibox/widget/base.lua
```diff
@@ -162,14 +162,16 @@ end
...
 --
--- Note that this function has two flaws:
+-- Note that this function has some flaws:
 --
 -- 1. The signal is only forwarded once the widget tree has been built. This
 --    happens after all currently scheduled functions have been executed.
 --    Therefore, it will not start to work right away.
 -- 2. In case the widget is present multiple times in a single widget tree,
---    this function will also forward the signal multiple time (one per upward
+--    this function will also forward the signal multiple time (once per upward
```
blueyed: s/time/times/

psychon: Done, thanks & good catch.

PR #2001, opened 2017-08-16
build: make DO_COVERAGE a CMake variable (and fix flags)

blueyed:


.travis.yml
```diff
@@ -159,9 +159,8 @@ script:
   - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -DLUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -DSTRICT_TESTS=true"
   - |
     if [ "$DO_COVERAGE" = "codecov" ]; then
-      export CXXFLAGS="-fprofile-arcs -ftest-coverage"
       export CFLAGS="-fprofile-arcs -ftest-coverage"
-      export LDFLAFS="-fprofile-arcs -ftest-coverage"
+      export LDFLAGS="-fprofile-arcs"
```
psychon: Uhm, why? (And I think that should be part of the commit message)

For `$CXXFLAGS` I guess the reason is that we have no C++ code.
For `$LDFLAGS` I don't know. If I had to guess, I would say that this is not actually a valid linker flag.

Please improve the commit message.

.travis.yml
```diff
@@ -159,8 +159,8 @@ script:
   - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -DLUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -DSTRICT_TESTS=true"
   - |
     if [ "$DO_COVERAGE" = "codecov" ]; then
-      export CFLAGS="-fprofile-arcs -ftest-coverage"
-      export LDFLAGS="-fprofile-arcs"
+      export CFLAGS="-fprofile-arcs -ftest-coverage -O0"
+      export LDFLAGS="-fprofile-arcs -O0"
```
psychon: Are you sure this works? I just asked Google and it seems like this is only added to CMake's default flags (AFAIR this includes `-O2`). However, I could not figure out if these values are appended or prepended to CMake's default flags. If `-O0` ends up being before the default `-O2`, it won't have any effect.

Also, the commit message could say a bit more about why this is done. I guess the reason is to have more meaningful coverage data (so that things that the optimizer removes or moves around don't end up confusing us), but I feel like that could still be in there.

.travis.yml
```diff
@@ -156,7 +155,7 @@ install:
       return 0
     }
 script:
-  - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -DLUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -DSTRICT_TESTS=true"
+  - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -DLUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -DSTRICT_TESTS=true -D DO_COVERAGE=$DO_COVERAGE"
```
psychon: For consistency I would remove the space after `-D`.

.travis.yml
```diff
@@ -236,7 +235,7 @@ script:
         echo "Testing commit $commit"
         git checkout "$commit"
         git --no-pager show --stat
-        if ! DO_COVERAGE=0 make all check; then
+        if ! make all check CMAKE_ARGS+="-D DO_COVERAGE=0"; then
```
psychon: Why the space after `-D`?

awesomeConfig.cmake
```diff
@@ -332,6 +334,7 @@ if(DO_COVERAGE)
                     ${BUILD_DIR}/${file}
                     COPYONLY)
     endforeach()
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -O0 --coverage -fprofile-arcs -ftest-coverage")
```
psychon: Shouldn't the commit also remove the part of `.travis.yml` which added these flags? Also, why are these different flags than what `.travis.yml` set? What happened to `$LDFLAGS`?

.travis.yml
blueyed: > Please improve the commit message.

Ok.

It looks like we might not need the flag at all - it had a typo before, but somewhere I've read that it should use -fprofile-arcs for linking, but could not find it later again.

.travis.yml
blueyed: I would add it to the others (one has it already): otherwise it is harder to grep (no full WORD).

.travis.yml
blueyed: See above.

awesomeConfig.cmake
blueyed: LDFLAGS does not seem to be necessary (as the coverage report shows).
Should I re-add it here?

.travis.yml
psychon: Oh right, I missed that part for `OVERRIDE_VERSION`.

awesomeConfig.cmake
psychon: In that case this should perhaps be one commit changing the flags in `.travis.yml` and another that just moves (remaining) things to `CMakeLists.txt` without actually changing anything.

blueyed:
Rebased and squashed.
Hopefully a bit clearer now?!

blueyed:
Found these references: https://gitlab.kitware.com/cmake/cmake/blob/master/Utilities/cmlibarchive/build/cmake/LibarchiveCodeCoverage.cmake
https://stackoverflow.com/a/6090147/15690

blueyed:
Only merged the first commit.

PR #2003, opened 2017-08-17
KeyboardLayout widget: Add EurKey

SpiritCroc:
Fixes eurkey keyboard layout not being shown in widget (either keeping old layout indicator or not showing any at all).

Also, switching to eurkey with a button combination, e.g. after using
setxkbmap -layout "eu,de" -option "grp:shift_caps_toggle"
caused the error message
"Oops, an error happened!
/usr/share/awesome/lib/awful/widget/keyboardlayout.lua:124: attempt to concatenate a nil value (field '?')


SpiritCroc:
Ah OK, sorry, didn't test using the latest code base

blueyed:
Thank you!

PR #2004, opened 2017-08-17
luacheck: enable cache

blueyed:


.luacheckrc
```diff
@@ -46,4 +46,7 @@ globals = {
     "client"
 }
 
+-- Enable cache (uses .luacheckcache relative to this rc file).
+cache = true
```
psychon: Sorry, I seem to be wrong. It seems like the `cache` option silently does nothing if `lfs` is not available.

blueyed: That was my impression, too - but have not verified it.

Ok to merge then?

PR #2005, opened 2017-08-17
doc: improve/fix doc for awful.rules

blueyed:


lib/awful/rules.lua
```diff
@@ -1,5 +1,8 @@
 ---------------------------------------------------------------------------
---- Apply rules to clients at startup.
+--- Rules for clients.
+--
+-- This module applies @{rules} to clients during startup (via @{client.manage},
```
Elv13: This is just a random comment, not a hard change request. Why `@{}` instead of "`". IMHO, the latter is much more readable and end up the same in the HTML output.

psychon: `@{}` ends up as an error from LDoc if the reference cannot be found while `` just creates the same "look", but without the link in that case.

blueyed: Yes.  This might be less readable, but ensures that it actually becomes a link.

PR #2008, opened 2017-08-19
feat(awful: widget: calendar_popup: attach): implement 'on_hover' option

actionless:


lib/awful/widget/calendar_popup.lua
```diff
@@ -246,9 +246,12 @@ end
 --
 -- @param widget Widget to attach the calendar
 -- @tparam[opt="tr"] string position Two characters string defining the position on the screen
+-- @tparam table args Additional options
+-- @tparam bool args.on_hover Show popup while mouse hover
```
Elv13: `@tparam[opt=false] bool`

lib/awful/widget/calendar_popup.lua
```diff
@@ -246,9 +246,12 @@ end
 --
 -- @param widget Widget to attach the calendar
 -- @tparam[opt="tr"] string position Two characters string defining the position on the screen
+-- @tparam[opt={}] table args Additional options
+-- @tparam[opt=false] bool args.on_hover Show popup while mouse hover
```
blueyed: s/while/during ?

blueyed: Isn't `true` a better default?

actionless: i thought the idea not to break the defaults, but as for me it's better

blueyed: It's a nice feature after all - I think enabling by default is good.

actionless: also, what do you think the fact what click is still binds to the widget? it feels a bit illogical to me -- what both left click and hover are toggling the widget visibility

actionless:
i think it wil be better to clarify this question: https://github.com/awesomeWM/awesome/pull/2008#discussion_r134330709 before mering

lib/awful/widget/calendar_popup.lua
blueyed: I've not yet used it myself.
What about clicking makes it stay visible until you click it again?
Is that the case already?

actionless: i am not sure i getting the question so better if you try on your own:

```lua
  awful.widget.calendar_popup.month():attach(WIDGET, nil, {on_hover=true})
```


actually the part of scenario which seems a bit weird to me:

1) mouse hovered on widget - calendar appears
2) left mouse down (still hovered) -- calendar disappers

actionless: now i think i've got your point, we'll need to introduce additional state like `was_clicked` in order to handle it in `mouse_over` signal like `if not was_clicked then w.visible=false end`

i will update the PR tomorrow

actionless: in this calendar all parts are made like widgets, so it's possible to, for example, configure shape and spacing between days and so on

before i was using `cal` inside `naughty`

actionless:
@blueyed now i think it's fine

psychon:
Travis failed:

> Error: Failed installing dependency: https://luarocks.org/say-1.3-1.rockspec - Error fetching file: Failed downloading https://github.com/Olivine-Labs/say/archive/v1.3-1.tar.gz - v1.3-1.tar.gz

The error seems to persist (yup, I'm using a German locale):
```
$ wget https://github.com/Olivine-Labs/say/archive/v1.3-1.tar.gz 
--2017-08-23 14:34:52--  https://github.com/Olivine-Labs/say/archive/v1.3-1.tar.gz
Auflösen des Hostnamens »github.com (github.com)« … 192.30.253.113, 192.30.253.112
Verbindungsaufbau zu github.com (github.com)|192.30.253.113|:443 … verbunden.
HTTP-Anforderung gesendet, auf Antwort wird gewartet … 302 Found
Platz: https://codeload.github.com/Olivine-Labs/say/tar.gz/v1.3-1 [folgend]
--2017-08-23 14:34:53--  https://codeload.github.com/Olivine-Labs/say/tar.gz/v1.3-1
Auflösen des Hostnamens »codeload.github.com (codeload.github.com)« … 192.30.253.120, 192.30.253.121
Verbindungsaufbau zu codeload.github.com (codeload.github.com)|192.30.253.120|:443 … verbunden.
HTTP-Anforderung gesendet, auf Antwort wird gewartet … 500 Internal Server Error
2017-08-23 14:35:08 FEHLER 500: Internal Server Error.
```

actionless:
@psychon i've restarted and this time `say` installed fine but not `luassert`:
https://travis-ci.org/awesomeWM/awesome/jobs/267557973

actionless:
today the commuinication between travis and luarocks is fine :-)

blueyed:
On hover, I think it should be kept visible as long as the popup is hovered, i.e. when moving the mouse from the widget to the popup it should still be kept visible?!

actionless:
@blueyed mouse-leave event will fire and hide calendar before calendar will catch mouse-enter

a possible solution could be to hide calendar with some delay (but such behavior sounds quite annoying to me)

blueyed:
Ok then.
Should it get squashed?
Feel free to merge it yourself then.

actionless:
thanks for looking into this :-)

PR #2009, opened 2017-08-19
fix(tests: run.sh): add one more error log pattern

actionless:
`themes/tests.lua` silently fails here:
https://travis-ci.org/awesomeWM/awesome/jobs/266367133#L2845

@blueyed i had that concern during review of #1967 but unfortunately after #1973 was merged the problem still persists

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -290,14 +290,14 @@ function utils.parse_dir(dir_path, callback)
         local query = gio.FILE_ATTRIBUTE_STANDARD_NAME .. "," .. gio.FILE_ATTRIBUTE_STANDARD_TYPE
         local enum, err = file:async_enumerate_children(query, gio.FileQueryInfoFlags.NONE)
         if not enum then
-            gdebug.print_error(err)
+            gdebug.print_warning(dir_path .. ": " .. tostring(err))
             return
         end
         local files_per_call = 100 -- Actual value is not that important
         while true do
             local list, enum_err = enum:async_next_files(files_per_call)
             if enum_err then
-                gdebug.print_error(enum_err)
+                gdebug.print_warning(dir_path .. ": " .. tostring(enum_err))
                 return
             end
             for _, info in ipairs(list) do
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -285,19 +285,23 @@ end
 -- @tparam table callback.programs Paths of found .desktop files.
 function utils.parse_dir(dir_path, callback)
 
+    local function get_readable_path(file)
+        return file:get_path() or file:get_uri()
+    end
+
     local function parser(file, programs)
         -- Except for "NONE" there is also NOFOLLOW_SYMLINKS
         local query = gio.FILE_ATTRIBUTE_STANDARD_NAME .. "," .. gio.FILE_ATTRIBUTE_STANDARD_TYPE
         local enum, err = file:async_enumerate_children(query, gio.FileQueryInfoFlags.NONE)
         if not enum then
-            gdebug.print_warning(dir_path .. ": " .. tostring(err))
+            gdebug.print_warning(get_readable_path(file) .. ": " .. tostring(err))
             return
         end
         local files_per_call = 100 -- Actual value is not that important
         while true do
             local list, enum_err = enum:async_next_files(files_per_call)
             if enum_err then
-                gdebug.print_warning(dir_path .. ": " .. tostring(enum_err))
+                gdebug.print_error(get_readable_path(file) .. ": " .. tostring(enum_err))
                 return
             end
             for _, info in ipairs(list) do
--- a/tests/themes/run.sh
+++ b/tests/themes/run.sh
@@ -12,6 +12,17 @@ if ! [ -f CMakeLists.txt ]; then
 fi
 source_dir="$PWD"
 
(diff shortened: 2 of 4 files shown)


actionless:
yup, now it catches broken image paths in the themes ^__^
https://travis-ci.org/awesomeWM/awesome/jobs/266371155

blueyed:
Thanks for this.

The first issue should be fixed by my fixups.

Re 2. 
> also some unexpected error in test-menubar:
> https://travis-ci.org/awesomeWM/awesome/jobs/266371322#L2393
> "Not a directory"

This comes via https://github.com/awesomeWM/awesome/blob/67b236f1573fefe544aa20c52423028598c1477b/tests/test-menubar.lua#L26-L38

@psychon 
It looks like your code is trying to ignore this kind of error?!
Where is it coming from initially?

Wouldn't it be good to have the stacktrace there anyway?

blueyed:
Only case 2 ("E: awesome: Not a directory") is left now, and commits need to get ge-orderd / squashed.

psychon:
Uhm, how do you know that its this part of `test-menubar.lua`? That code only ignores some errors, but doesn't "cause" new ones, doesn't it?

> It looks like your code is trying to ignore this kind of error?!

Yup, `$HOME` is set to `/dev/null` and opening `/dev/null/.share/.......` caused lots of false positives for the error detection. The other part was about `/usr/local/something` AFAIK. Basically, these errors do not indicate a problem.

> Where is it coming from initially?

Uhm... From the menubar trying to open a non-existent directory to scan for `.desktop` files? Or what do you mean?

> Wouldn't it be good to have the stacktrace there anyway?

Why stacktrace? `gears.debug.print_error` never produces tracebacks, does it?

Somehow I feel like there is a misunderstanding here, but I cannot really put my finger on it...?

tests/themes/run.sh
```diff
@@ -12,6 +12,17 @@ if ! [ -f CMakeLists.txt ]; then
...
+# Either the build dir is passed in $CMAKE_BINARY_DIR or we guess based on $PWD
+# Same as in tests/run.sh.
+build_dir="$CMAKE_BINARY_DIR"
+if [ -z "$build_dir" ]; then
+    if [ -d "$source_dir/build" ]; then
+        build_dir="$source_dir/build"
+    else
+        build_dir="$source_dir"
+    fi
+fi
```
psychon: Didn't I already see this commit in some other PR...?
Anyway, I'm not really sure what to review here. `tests/run.sh` is @blueyed's playground. ;-) If it helps: 👍

actionless:
could we turn that menubar message into a warning and also mention the path which has been failed to open?

psychon:
> could we turn that menubar message into a warning and also mention the path which has been failed to open?

Sure, feel free. The error happens in `lib/menubar/utils.lua`, function `parse_dir`. `async_enumerate_children` returns it: https://github.com/awesomeWM/awesome/blob/cdd58cc355470c307bd39956b4516bfadd248fc7/lib/menubar/utils.lua#L291-L295

actionless:
now instead of 

```
E: awesome: Not a directory
```

it's 

```
W: awesome: /dev/null/.local/share/applications/: Not a directory
```

however 2 of 5 test targets are also checking for warnings, so idk how to proceed next

lib/menubar/utils.lua
```diff
@@ -290,14 +290,14 @@ function utils.parse_dir(dir_path, callback)
...
-            gdebug.print_error(err)
+            gdebug.print_warning(dir_path .. ": " .. tostring(err))
             return
         end
         local files_per_call = 100 -- Actual value is not that important
         while true do
             local list, enum_err = enum:async_next_files(files_per_call)
             if enum_err then
-                gdebug.print_error(enum_err)
+                gdebug.print_warning(dir_path .. ": " .. tostring(enum_err))
```
psychon: Could this use `local path = file:get_path() or file:get_uri()` instead of just `dir_path`? `dir_path` is just the directory that we started with, but some subdirectory could be throwing some errors, I think (although I'm not really sure what kind of errors... `chmod 0` would already be caught earlier...?)

actionless:
so, any idea how to proceed with warning coming from menubar test?

psychon:
> so, any idea how to proceed with warning coming from menubar test?

Fix the already existing code that tries to handle them (untested):
```diff
diff --git a/tests/test-menubar.lua b/tests/test-menubar.lua
index 6c31c35..0eb5d43 100644
--- a/tests/test-menubar.lua
+++ b/tests/test-menubar.lua
@@ -27,12 +27,12 @@ end
 -- word "error" appears in our output, the test is considered to have failed.
 do
     local gdebug = require("gears.debug")
-    local orig_error = gdebug.print_error
-    function gdebug.print_error(msg)
+    local orig_warning = gdebug.print_warning
+    function gdebug.print_warning(msg)
         msg = tostring(msg)
-        if (msg ~= "Error opening directory '/dev/null/.local/share/applications': Not a directory") and
+        if (not msg:match("/dev/null/.local/share/applications")) and
                 (not msg:match("No such file or directory")) then
-            orig_error(msg)
+            orig_warning(msg)
         end
     end
 end
```

actionless:
@blueyed what do you think now?

actionless:
so, if anything else need to be done?

blueyed:
Thanks!

PR #2010, opened 2017-08-20
feat(awful: widget: calendar_popup): implement 'margin' widget property

actionless:


actionless:
i think also need to do smth with shape because currently it's not applied to wibox itself but only to its background

actionless:
done

lib/awful/widget/calendar_popup.lua
```diff
@@ -272,6 +273,7 @@ end
 -- @tparam string args.bg Wibox background color
 -- @tparam string args.font Calendar font
 -- @tparam number args.spacing Calendar spacing
+-- @tparam number args.margin Margin around calendar widget
```
blueyed: Why don't you like punctuation with docs?
In this case it matches the surrounding code, but I would really like this to be more enforced, i.e. also change it for the surrounding code (in a separate commit if you like to).

actionless: i just did like in the other params of this function

actionless: maybe it can be checked in a similar way like luacheck, or mb even there is a way to plug external checks there

actionless:
but naughty have similar properties -- padding and margin, the only caveat, what in naughty those two does mean the opposite :D

psychon:
By "random" I meant doing `widget.foobar = 42` where the widget does not have a `foobar` property. It works, but feels ugly to me. So this was more meant for the implementation and not the feature itself.

actionless:
i see the point, should i rename the property from `.margin` to `._calendar_margin`?

actionless:
changed it in case if widget will have property with the same name eventually

PR #2011, opened 2017-08-20
Update the shims

Elv13:
The shims are *not* made to be accurate, but rather to have the minimal/simplest of code required for the higher level code to work.

### DIFF
--- a/tests/examples/shims/client.lua
+++ b/tests/examples/shims/client.lua
@@ -14,6 +14,14 @@ end
 --     CURRENTLY UNUSED
 -- end
 
+local function titlebar_meta(c)
+    for _, position in ipairs {"top", "bottom", "left", "right" } do
+        c["titlebar_"..position] = function(size) --luacheck: no unused
+            return drawin{}
+        end
+    end
+end
+
 -- Create fake clients to move around
 function client.gen_fake(args)
     local ret = gears_obj()
@@ -22,6 +30,7 @@ function client.gen_fake(args)
     ret.valid = true
     ret.size_hints = {}
     ret.border_width = 1
+    ret.icon_sizes = {{16,16}}
 
     -- Apply all properties
     for k,v in pairs(args or {}) do
@@ -71,6 +80,20 @@ function client.gen_fake(args)
         return nil
     end
 
+    function ret:get_icon(_)
+        return require("beautiful").awesome_icon
+    end
+
+    function ret:raise()
+        --TODO
+    end
+
+    function ret:lower()
+        --TODO
+    end
+
+    titlebar_meta(ret)
+
     function ret:tags(new) --FIXME
         if new then
             ret._tags = new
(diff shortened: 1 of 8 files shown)


blueyed:
I think the diff for this should be covered 100%?!

Elv13:
> I think the diff for this should be covered 100%?!

https://github.com/awesomeWM/awesome/pull/1990 was. Same for all the other PRs I created yesterday. As a whole it had good coverage. But as all the test related commits depends on more than 1 of these PR, I could not push those commits. The whole 30 commit branch has >90% coverage.

tests/examples/shims/client.lua
```diff
@@ -14,6 +14,15 @@ end
 --     CURRENTLY UNUSED
 -- end
 
+local function titlebar_meta(c)
+    for _, position in ipairs {"top", "bottom", "left", "right" } do
+        c["titlebar_"..position] = function(size)--luacheck: no unused
```
blueyed: Space before `--luacheck`.

tests/examples/shims/client.lua
```diff
@@ -14,6 +14,15 @@ end
...
+            local d = drawin{}
+            return d
```
blueyed: `return drawin({})` ?

tests/examples/shims/client.lua
```diff
@@ -71,6 +81,20 @@ function client.gen_fake(args)
...
+        return require("beautiful").awesome_icon
+    end
+
+    function ret:raise()
+        --TODO
+    end
+
+    function ret:lower()
+        --TODO
+    end
```
blueyed: Address the TODOs? / or leave it out for now?!

tests/examples/shims/dbus.lua
```diff
@@ -0,0 +1,24 @@
...
+        "org.freedesktop.Notifications",
+        data,
+        appname,
+        replaces_id,
+        icon,
+        title,
+        text,
+        actions,
+        hints,
+        expire
```
blueyed: Use `...`, so it becomes `dbus.emit_sigan("org.freedesktop.Notifications", ...)` ?

tests/examples/shims/screen.lua
```diff
@@ -94,8 +98,23 @@ function screen._setup_grid(w, h, rows, args)
...
 
+local function iter_scr(_, _, s)
+    if not s then
+        assert(screen[1])
+        return screen[1], 1
+    end
+
+    local i = s.index
+
+    if i +1  < #screen then
```
blueyed: s/+1 /+ 1/

tests/examples/shims/client.lua
Elv13: The functions are called by some of the newer tests, but the behavior is not (yet) required. To keep things simple, I would rather not implement z-index support in this commit.

PR #2012, opened 2017-08-20
widget: Add a flexible generic constructor

Elv13:
This avoid copy pasting this kind of code in many places. It will eventually help make some parts of the documentation less confusing and allow multiple instances to be created with "lazy loading".

### DIFF
--- a/lib/wibox/widget/base.lua
+++ b/lib/wibox/widget/base.lua
@@ -584,6 +584,34 @@ function base.make_widget_declarative(args)
     return setmetatable(w, mt)
 end
 
+--- Create a widget from an undetermined value.
+--
+-- The value can be:
+--
+-- * A widget (in which case nothing new is created)
+-- * A declarative construct
+-- * A constructor function
+-- * A metaobject
+--
+-- @param wdg The value.
+-- @param[opt=nil] ... Arguments passed to the contructor (if any).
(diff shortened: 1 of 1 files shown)


lib/wibox/widget/base.lua
```diff
@@ -584,6 +584,34 @@ function base.make_widget_declarative(args)
...
+-- * A constructor function
+-- * A metaobject
+--
+-- @param wdg The value.
+-- @param[opt=nil] ... Arguments passed to the contructor (if any).
+-- @treturn The new widget.
+function base.make_widget_from_value(wdg, ...)
+    local is_table = type(wdg) == "table"
+    local is_function = ((not is_table) and type(wdg) == "function")
+        or (is_table and getmetatable(wdg) and getmetatable(wdg).__call)
```
psychon: Why the `(not is_table)` in the first case? `type(wdg) ~= "table" and type(wdg) == "function"` is the same thing as just `type(wdg) == "function"`? (Yes, I read this code twice before understanding it).

I can't say that I'm happy with this, but... sigh.

blueyed: @Elv13 
Please address.

Elv13: Premature optimization?

PR #2013, opened 2017-08-20
drawin: Add a `get()` function.

Elv13:
Identical to `client.get()`. Same use cases.

### DIFF
--- a/objects/drawin.c
+++ b/objects/drawin.c
@@ -238,6 +238,25 @@ drawin_refresh(void)
     }
 }
 
+/** Get all drawins into a table.
+ * @treturn table A table with drawins.
+ * @function get
+ */
+static int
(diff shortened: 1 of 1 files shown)


Elv13:
> Same use cases.

The use case I was referring to was to get the list of existing wibox just you might want to get the list of existing clients.

It's used in some later PR to render the `wibar`s, `notification`s and `tooltip`s in the documentation example images without having to invent a secret API. It's also used in one of the debug utility I wrote last year to dump the whole screen as an introspectable HTML document. With that tool, you get a complete copy of the current state, with all clients, bars, wallpaper in a layered SVG document that can be dissected using the regular browser tools. This allow to see all widgets and how they affect each other.

Elv13:
> For the implementation: Could you call the variable d instead of c inside the loop

done

PR #2015, opened 2017-08-20
object: Add support for signal forwarding

Elv13:
The CAPI signal system already has this mechanism. Same use case `naughty.connect_signal("property::foo", notif) end` <==> `notif:connect_signal("property::foo", notif) end`

### DIFF
--- a/lib/gears/object.lua
+++ b/lib/gears/object.lua
@@ -56,6 +56,11 @@ function object:connect_signal(name, func)
     sig.strong[func] = true
 end
 
+-- Register a global signal receiver.
+function object:_connect_everything(callback)
+    table.insert(self._global_receivers, callback)
+end
+
 local function make_the_gc_obey(func)
     if _VERSION <= "Lua 5.1" then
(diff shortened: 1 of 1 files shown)


actionless:
i like the idea but indeed it missing docs

Elv13:
> How about adding a parent argument in new instead? The implementation would then be something like

That's what Qt does for it's object tree, but so far I just want to bring the `capi.object` and `gears.object` based systems to the same capability. I would rather only have the smallest possible change and have both behave the same way rather than improving one while now having `capi.object` to be less capable.

Elv13:
> i like the idea but indeed it missing docs

This is a private API for now. It will be documented in the classes that actually supports it.

psychon:
> but so far I just want to bring the capi.object and gears.object based systems to the same capability. I would rather only have the smallest possible change and have both behave the same way rather than improving one while now having capi.object to be less capable.

The C code has a list of classes and supports instantiating these classes. So, if you want this, then you'll have to add a concept of class to `gears.object`.

PR #2017, opened 2017-08-20
grid: Fix an unavoidable assert

Elv13:
`:add()` was called with the optional `...` arguments, but the
function required at least one argument.


blueyed:
@psychon 
> Doesn't this break the API docs? Somehow I don't see LDoc complaining...?

Can you elaborate, please?

Elv13:
I see why you are worried, but no, `ldoc` doesn't complain and renders fine

nova_pluto4

psychon:
> Can you elaborate, please?

There is a new `widget` argument that is undocumented. LDoc cannot know that it belongs to the `...`, but apparently it just silently ignores this, as @Elv13 shows.

PR #2018, opened 2017-08-20
clienticon: Add a client setter.

Elv13:
It's required to work with the declarative syntax and be used
in the `tasklist`.

### DIFF
--- a/lib/awful/widget/clienticon.lua
+++ b/lib/awful/widget/clienticon.lua
@@ -31,7 +31,7 @@ end
 
 function clienticon:draw(_, cr, width, height)
     local c = self._private.client
-    if not c.valid then
+    if not c or not c.valid then
         return
     end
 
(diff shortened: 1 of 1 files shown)


lib/awful/widget/clienticon.lua
```diff
@@ -80,6 +80,21 @@ function clienticon:fit(_, width, height)
...
+--
+-- @property client
+-- @param client
+
+function clienticon:get_client()
+    return self._private.client
+end
+
+function clienticon:set_client(c)
+    self._private.client = c
```
psychon: I'd add an `if self._private.client == c then return end`.

blueyed: Sounds good.

blueyed:
Can we cover this with a test, please?

Elv13:
> Can we cover this with a test, please?

https://github.com/awesomeWM/awesome/pull/2011#issuecomment-323619989

lib/awful/widget/clienticon.lua
Elv13: done

PR #2019, opened 2017-08-20
Add a separator widget.

Elv13:
It was requested a bunch of time (#1149), is simple and will be useful in some future PR to implement the changes proposed in #1874.

### DIFF
--- a/lib/wibox/widget/separator.lua
+++ b/lib/wibox/widget/separator.lua
@@ -1,0 +1,205 @@
+---------------------------------------------------------------------------
+-- A flexible separator widget.
+--
+-- By default, this widget display a simple line, but can be extended by themes
+-- (or directly) to display much more complex visuals.
+--
+-- This widget is mainly intended to be used alongside the `spacing_widget`
+-- property supported by various layouts such as:
+--
+-- * `wibox.layout.fixed`
+-- * `wibox.layout.flex`
+-- * `wibox.layout.ratio`
+--
+-- When used with these layouts, it is also possible to provide custom clipping
+-- functions. This is useful when the layout has overlapping widgets (negative
+-- spacing).
+--
+--@DOC_wibox_widget_defaults_separator_EXAMPLE@
+--
+-- @author Emmanuel Lepage Vallee &lt;elv1313@gmail.com&gt;
+-- @copyright 2014, 2017 Emmanuel Lepage Vallee
+-- @classmod wibox.widget.separator
+---------------------------------------------------------------------------
+local beautiful = require( "beautiful"         )
+local base      = require( "wibox.widget.base" )
+local color     = require( "gears.color"       )
+local gtable    = require( "gears.table"       )
+
+local separator = {}
+
+--- The separator's orientation.
+--
+-- Valid values are:
+--
+-- * *vertical*: From top to bottom
+-- * *horizontal*: From left to right
+-- * *auto*: Decide depending on the widget geometry (default)
+--
+-- The default value is selected automatically. If the widget is taller than
+-- large, it will use vertical and vice versa.
+--
+--@DOC_wibox_widget_separator_orientation_EXAMPLE@
+--
+-- @property orientation
+-- @param string
+
+--- The separator's thickness.
+--
+-- This is used by the default line separator, but ignored when a shape is used.
+--
+-- @property thickness
+-- @param number
+
+--- The separator's shape.
+--
+--@DOC_wibox_widget_separator_shape_EXAMPLE@
+--
+-- @property shape
+-- @tparam function shape A valid shape function
+-- @see gears.shape
+
+--- The relative percentage covered by the bar.
+-- @property span_ratio
+-- @tparam[opt=1] number A number between 0 and 1.
+
+--- The separator's color.
+-- @property color
+-- @param string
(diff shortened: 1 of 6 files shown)


lib/wibox/widget/separator.lua
```diff
@@ -0,0 +1,188 @@
...
+    if orientation == "horizontal" then
+        local w = width*span_ratio
+        cr:rectangle((width-w)/2, height/2 - thickness/2, w, thickness)
+    else
+        local h = height*span_ratio
+        cr:rectangle(width/2 - thickness/2, (height-h)/2, thickness, h)
+    end
+
+    cr:fill()
+end
```
psychon: Can't this line be done as a shape? I.e. when no shape is set, this function (without the `cr:fill()` at the end) is just used instead? That would then also add a border color for the line.

lib/wibox/widget/separator.lua
```diff
@@ -0,0 +1,188 @@
...
+    separator["set_"..prop] = function(self, value)
+        self._private[prop] = value
+        self:emit_signal("property::"..prop)
+        if prop == "data_list" then
+            self:emit_signal("property::data")
+        end
+        self:emit_signal("widget::redraw_needed")
+    end
+    separator["get_"..prop] = function(self)
+        return self._private[prop] or beautiful["piechart_"..prop]
```
psychon: `piechart`?

psychon: Also, the actual implementation uses `._private.foo` directly, so any default values that you implement here will silently be ignored by the widget.

lib/wibox/widget/separator.lua
```diff
@@ -0,0 +1,188 @@
...
+        local h = height*span_ratio
+        cr:rectangle(width/2 - thickness/2, (height-h)/2, thickness, h)
+    end
+
+    cr:fill()
+end
+
+local function draw(self, _, cr, width, height)
+    -- In case there is a specialized.
+    local draw_custom = self._private.draw or beautiful.separator_draw
```
psychon: So, this is a widget whose `draw` function can just be replaced in a theme? Meh... okay.

lib/wibox/widget/separator.lua
```diff
@@ -0,0 +1,188 @@
...
+    elseif self._private.border_color then
+        cr:fill_preserve()
+        cr:set_source(color(self._private.border_color))
+        cr:set_line_width(bw)
+        cr:stroke()
+    end
+end
+
+local function draw_line(self, _, cr, width, height)
+    local thickness = self._private.thickness or beautiful.separator_thickness or 1
```
psychon: `separator_thickness` is undocumented.

Also, I'm a bit confused about `separator_thickness` and `span_ratio`. One is a percentage of the "width" (in a horizontal line) while the other is a fixed size in pixels. I can see why the width should get a percentage, but still this feels inconsistent to me.

Elv13: > One is a percentage of the "width" (in a horizontal line) while the other is a fixed size in pixels

I don't have much to say beside "it scales better when you resize the wibar", but it's a tiny corner case. When you have a larger wibar/notification, you don't want to have a thicker separator. But keeping the same number of pixel for the length just looks wrong (to me).

lib/wibox/widget/separator.lua
Elv13: It could, but I think it would make things more complex for the same end result. At least this code is rather obvious and isn't using all the shape API indirections.

lib/wibox/widget/separator.lua
Elv13: That's basically how QtWidgets style works, less or more.

Elv13:
done

lib/wibox/widget/separator.lua
```diff
@@ -0,0 +1,207 @@
...
+    separator["set_"..prop] = function(self, value)
+        self._private[prop] = value
+        self:emit_signal("property::"..prop)
+        if prop == "data_list" then
+            self:emit_signal("property::data")
+        end
+        self:emit_signal("widget::redraw_needed")
+    end
+    separator["get_"..prop] = function(self)
+        return self._private[prop] or beautiful["separator_"..prop]
```
psychon: Could/should all the code in the widget be changed to use this getter? So, for example, in line 163, instead of `self._private.shape or beautiful.separator_shape`, just use `self.shape`?
For example, right now the getter will "lie" about a separator's orientation if `beautiful.separator_orientation` is set, because the actual code using it doesn't check the beautiful value. Just using the getter everywhere would ensure nothing like this can happen.

lib/wibox/widget/separator.lua
```diff
@@ -0,0 +1,207 @@
...
+end
+
+for _, prop in ipairs {"orientation", "color", "thickness", "span_ratio",
+                       "border_width", "border_color", "shape" } do
+    separator["set_"..prop] = function(self, value)
+        self._private[prop] = value
+        self:emit_signal("property::"..prop)
+        if prop == "data_list" then
+            self:emit_signal("property::data")
+        end
```
psychon: Uhm... dead code? There is no `data_list` property?!

Elv13:
Fixed

PR #2023, opened 2017-08-26
default config: add Pinentry to floating windows

blueyed:
We have class "pinentry" there already, but it seems to have changed to
"Pinentry"?!

pinentry 1.0.0-1, gnupg 2.1.23-1


awesomerc.lua
```diff
@@ -488,6 +488,7 @@ awful.rules.rules = {
           "Gpick",
           "Kruler",
           "MessageWin",  -- kalarm.
+          "Pinentry",  -- new
           "Sxiv",
           "Wpa_gui",
           "pinentry",
```
actionless: mb put it one after another

Issue #2026, opened 2017-09-01
The calendar popup skips two months ahead instead of one if it is the 31th and the next month has fewer days


psychon:
Just stumbled upon this. The build for master after merging #2001 worked, but the merge of #1983 caused all four builds to fail. Thanks to `[ci skip]` there was no previous build for that branch.

The failure is the following:
```
== Running /home/travis/build/awesomeWM/awesome/tests/test-awful-widget-calendar_popup.lua ==
Error: running function for step 5/6 (@2): [string "--- Test for awful.widget.calendar..."]:60: assertion failed!
stack traceback:
	[C]: in function 'assert'
	[string "--- Test for awful.widget.calendar..."]:60: in function <[string "--- Test for awful.widget.calendar..."]:53>
	(...tail calls...)
	[C]: in function 'xpcall'
	/home/travis/build/awesomeWM/awesome/tests/_runner.lua:73: in function </home/travis/build/awesomeWM/awesome/tests/_runner.lua:64>
	[C]: in function 'xpcall'
	lib/gears/protected_call.lua:36: in function <lib/gears/protected_call.lua:35>
	(...tail calls...)
	lib/gears/timer.lua:226: in function <lib/gears/timer.lua:224>!
===> ERROR running /home/travis/build/awesomeWM/awesome/tests/test-awful-widget-calendar_popup.lua <===
```
Since #1983 did not touch this and since the failure is in the calendar widget, I guess this is date-related. (Also, the last successful build was three days ago and today all builds seem to fail)

The failing assertion is
```lua
assert( mdate.day==nil and
((mdate.month==month+1 and mdate.year==year) or (mdate.month==1 and mdate.year==year+1)) )
```
where `mdate` comes from `cmonth` after `call_calender(1)` was done. Uhm... yeah, that assertion might need a second argument explaining what its values were.

psychon:
The build for #2027 succeeded ?!

psychon:
Ah, the build failure was yesterday (slightly before midnight, not after).

The following patch reproduces the problem:
```diff
diff --git a/lib/awful/widget/calendar_popup.lua b/lib/awful/widget/calendar_popup.lua
index e204b10..ee79d9e 100644
--- a/lib/awful/widget/calendar_popup.lua
+++ b/lib/awful/widget/calendar_popup.lua
@@ -214,6 +214,7 @@ function calendar_popup:call_calendar(offset, position, screen)
 
     local widget = self:get_widget()
[...]

PR #2027, opened 2017-09-01
Mention minimum ldoc version in the readme

psychon:
See commit 2c741f8e1d934da32a which introduced this dependency.

There I was, trying to reproduce #2026 locally and my build failed for yet another reason. :-(
Do we need some kind of LDoc version check in CMake or is this supposed to be enough?


psychon:
Sigh. And now I can no longer build the docs on my Debian machine.

blueyed:
@psychon 
That is bad indeed.
See https://github.com/awesomeWM/awesome/pull/2035.

PR #2028, opened 2017-09-01
calendar_popup: Avoid skipping a month if the day is 31

getzze:
Correct bug #2026


psychon:
Just from reading the title I have on thing to say: "February" (sorry, I'm in a hurry right now)

Issue #344, opened 2015-07-23
textclock: should update immediately after suspending more than 1 minute


blueyed:
I've noticed that the textclock is not being updated immediately after resuming from suspend, although the maximum of 59 seconds for the timer have passed.

It seems like the glib timer does not use absolute time, but is relative?
Can we do something about this?

I think the relative behavior is good in general, but not for the textclock widget.

ff2000:
Daniel Hahler wrote:
> I've noticed that the textclock is not being updated immediately after resuming from suspend, although the maximum of 59 seconds for the timer have passed.
> 
> It seems like the glib timer does not use absolute time, but is relative?
> Can we do something about this?
Yes ;)
Catch the resume and do the update. For systemd (and consolekit?) it should be simply connecting to a dbus signal.
Triggering a signal would be really great because it would simplify updating other widgets. I personally would like to get current battery state directly after resume.

blueyed:
@ff2000 
Sounds like a plan: awesome would connect to the D-Bus signal(s) and then emit one itself internally, e.g. `resume` (according to the already existing `startup`).

Do you feel like creating a PR for this?

ff2000:
On Fri, 24 Jul 2015 06:43:07 -0700, Daniel Hahler <notifications@github.com> wrote:
> @ff2000 
> Sounds like a plan: awesome would connect to the D-Bus signal(s) and then emit one itself internally, e.g. `resume` (according to the already existing `startup`).
> 
> Do you feel like creating a PR for this?
Sorry for the late reply...
My lua is still bad as hell (my dbus is even worse), so it would be
great if you could do that.

I just had the idea of a complete "system" module, that would
encapsulate system calls like "reboot" or "hibernate" and add signals
for system actions like "resumed" or "about-to-shutdown" and so on.
(that's when I remembered this issue...)

Do you think that's doable?

blueyed:
I guess it's more on the C side, which would "just" need to hook into dbus and then emit the signal.

The "system module" would be just a particular set of signals then, e.g. `system::resumed`.

I've came across `xss-lock`, which provides calling a screenlocker when the system gets resumed.  That code might be relevant: https://bitbucket.org/raymonad/xss-lock/src/1e158fb20108058dbd62bd51d8e8c003c0a48717/src/xss-lock.c?at=master.

psychon:
Do we really want systemd-specific code in awesome? :-(
[...]

PR #2034, opened 2017-09-10
Add a force_update() method to the textclock

psychon:
Reference: https://github.com/awesomeWM/awesome/issues/344

I pushed this to the awesome repo instead of my fork so that the docs are build and I can check if LDoc picks this up correctly (https://github.com/awesomeWM/awesome/pull/2027#issuecomment-328354095).


blueyed:
Thank you!

PR #2039, opened 2017-09-15
Display desktop entries with no categories

### LINKED ISSUE #1007: menubar does not contain items without Categories
The following .desktop file (generated by Chromium for a web app) will be skipped by the menubar (Super-p), because it has not category:

```
#!/usr/bin/env xdg-open
[Desktop Entry]
Version=1.0
Terminal=false
Type=Application
Name=Signal Private Messenger
Exec=/usr/bin/chromium --profile-directory=Default --app-id=bikioccmkafdpakkkcpdbppfkghcmihk
Icon=chrome-bikioccmkafdpakkkcpdbppfkghcmihk-Default
StartupWMClass=crx_bikioccmkafdpakkkcpdbppfkghcmihk
```

Related code: https://github.com/awesomeWM/awesome/blob/f5f845bb73ce405a5c297ab0ac3816186daaa8d1/lib/menubar/menu_gen.lua#L109-L119

I wonder if we should just add all `Type=Application` entries, or/and look for other indicators, e.g. `StartupWMClass`?

lePerdu:
Uncategorized entries are displayed along side categorized ones when no category is selected, and are searchable.

Fixes #1007

### DIFF
--- a/lib/menubar/menu_gen.lua
+++ b/lib/menubar/menu_gen.lua
@@ -108,16 +108,15 @@ function menu_gen.generate(callback)
                                 end
                             end
                         end
-                        if target_category then
-                            local name = utils.rtrim(entry.Name) or ""
-                            local cmdline = utils.rtrim(entry.cmdline) or ""
-                            local icon = entry.icon_path or nil
-                            table.insert(result, { name = name,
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks again!

Issue #2040, opened 2017-09-21
Bug with naughty.notify and replaces_id: foreground colour not updated


spamn:
Output of `awesome --version`:
awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1

Please note that I have not tested against master branch but only with package from official archlinux repository. I hope this is not a duplicate or that the bug is not already fixed as I tried to find out if there was an issue already opened for it or if it was fixed by a recent commit.

**How to reproduce the issue:**
Simple way: add the following lines in your rc.lua file (provided you have required naughty properly) and restart awesome
local notif = naughty.notify({ text = "test", fg = '#ff0000'})
notif = naughty.notify({ text = "test 2", replaces_id = notif.id, fg = '#00ff00' })

**Actual result:**
Notification with red text 'test 2'.

**Expected result:**
Notification with green text 'test 2'.

spamn:
Please note that I have witnessed the same issue with background colour.

PR #2041, opened 2017-09-22
naughty: Fix replaces_id + changing colors

psychon:
This fixes a regression introduced in be29ee67688. This commit changed
naughty to reuse an already-existing wibox when replaces_id is used,
instead of creating a new wibox. However, some of the properties that
are set only when creating a wibox were ignored due to this.

Fixes: https://github.com/awesomeWM/awesome/issues/2040


blueyed:
> The diff coverage is 100%

I like that!

PR #2043, opened 2017-09-22
Unescape strings when parsing desktop entries

lePerdu:
This is based of the [spec](https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s03.html) and works with all of my desktop files (including those created by wine).

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -188,7 +188,10 @@ function utils.unescape(s, list)
         end
 
         local segments = {}
-        local i, j = 0, 0
+        -- Starting and ending indices of the match (optional backslashes before
+        -- a semicolon)
+        local i
+        local j = 0
         while true do
             i, j = string.find(s, '\\*;', j+1)
             if not i then
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -41,6 +41,23 @@ local default_icon = nil
 --- Name of the WM for the OnlyShownIn entry in the .desktop file.
 utils.wm_name = "awesome"
 
+--- Possible escape characters (not including the preceding backslash) in
+--.desktop files and their equates.
+local escape_chars = {
+    [ [[\]] ] = [[\]],
+    [ 'n' ]   = '\n',
+    [ 'r' ]   = '\r',
+    [ 's' ]   = ' ',
+    [ 't' ]   = '\t',
+}
+
+--- The keys in desktop entries whose values are lists of strings.
+local list_keys = {
+    Categories = true,
+    OnlyShownIn = true,
+    NotShowIn = true,
+}
+
 -- Private section
 
 local do_protected_call, call_callback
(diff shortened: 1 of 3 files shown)


lib/menubar/utils.lua
```diff
@@ -151,6 +168,68 @@ function utils.rtrim(s)
...
+-- the usual meanings.
+-- @tparam string s String to unescape
+-- @tparam boolean list Whether the string should be escaped as a list
+-- @treturn string|table The unescaped string or a table of unescaped strings if
+-- list is true.
+function utils.unescape(s, list)
+    if not s then return end
+    -- We can't just use string.gsub because something like [[\\s]] would become
+    -- either [[ ]] or [[\ ]] (depending on the order) while the correct result
+    -- is [[\s]].
```
psychon: How about (basically untested):
```
return string.gsub(s, "\\.", function(sequence)
    local c = sequence:sub(2, 2)
    return escape_chars[sequence:sub(2, 2)] or sequence
end)
```
The list case is a bit harder (should it get its own, separate function? this does more than just handling escape sequences). Again, some untested idea:
```
local segments = {}
local part = ""
-- Iterate over sub-strings between semicolons
for word in string.gmatch(s, "[^;]+") do
  if word:sub(-1, -1) == "\\" then
    -- The semicolon was escaped, remember this part for later
    part = part .. word:sub(1, -2) .. ";"
  else
    table.insert(segments, utils.unescape(part .. word))
    part = ""
  end
end
```

tests/test-menubar.lua
```diff
@@ -55,6 +55,56 @@ runner.run_steps {
...
+                if returned[i] ~= unescaped[i] then
+                    return false
+                end
+            end
+        end
+
+        return true
+    end,
+
     function()
```
psychon: I'd prefer this as unit tests instead of integration tests, i.e. in `spec/menubar/utils_spec.lua`.

lePerdu: Yeah, it fits better there. Didn't know about that directory.

lib/menubar/utils.lua
```diff
@@ -169,68 +169,48 @@ function utils.rtrim(s)
...
+    if not string.match(s, ';%s*$') then
+        s = s .. ';'
+    end
 
-        return s
+    local segments = {}
+    local part = ""
+    -- Iterate over sub-strings between semicolons
+    for word, backslashes in string.gmatch(s, "([^;]-(\\*));") do
+        if #backslashes % 2 ~= 0 then
```
psychon: Oh right, I missed that part. Nice catch & nice fix.

lib/menubar/utils.lua
```diff
@@ -169,68 +169,48 @@ function utils.rtrim(s)
...
+-- Semi-colons in lists are escaped as '\;'.
+-- @tparam string s String to parse
+-- @treturn table A table containing the separated strings. Each element is
+-- unescaped by utils.unescape().
+function utils.parse_list(s)
+    if not s then return end
+
+    -- Append terminating semi-colon if not already there. Ignore whitespace
+    -- at the end
+    if not string.match(s, ';%s*$') then
```
psychon: Wouldn't `if string.sub(s, -1, -1) ~= ';'` be better? The spec does not seem to talk about ignoring trailing whitespace (so I guess we should have a test for that case).

lePerdu: I took "trailing empty strings" to mean whitespace also, but I guess that's not technically correct and `desktop-file-utils` from freedesktop.org does not ignore trailing whitespace. I'll change that and add the test.

blueyed:
Thank you!

PR #2044, opened 2017-09-23
Travis: codecov: it expects a list of flags

blueyed:


blueyed:
Only true for codecov-python, but not codecov-bash.

Issue #2045, opened 2017-09-26
another magnifier layout bug?


WhosCounting:
Output of `awesome --version`:

awesome v4.2 (Human after all)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✘
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1

**How to reproduce the issue:**

1. Set a rule for a client to obey certain size and placement. For example like this:

  ```
  { rule = { instance =  "something"  },
       properties = { 
        floating = true, 
        width = 200, 
        height = 100 },              
      callback = function (c)
          awful.placement.centered(c, nil)
      end}  
```

2. Bind a client to a key.
3. Set magnifier layout on a given tag.
4. Invoke the client by pressing the key.

**Actual result:**

Client opens maximized (but can't be "unmaximized").

**Expected result:**

Client should obey its rules (i.e. be centered and of a given size).

Note also that the same thing happens when the main client is maximized on a tag with magnifier layout (when the main client is not maximized everything works fine).
This does not happen when using max or fullscreen layouts.

**Comment by psychon** [comment]
Please test the following patch:
```diff
diff --git a/lib/awful/layout/suit/magnifier.lua b/lib/awful/layout/suit/magnifier.lua
index 9417722..9faffdc 100644
--- a/lib/awful/layout/suit/magnifier.lua
+++ b/lib/awful/layout/suit/magnifier.lua
@@ -71,7 +71,7 @@ function magnifier.arrange(p)
     if focus and focus.screen ~= get_screen(p.screen) then focus = nil end
 
     -- If no window is focused or focused window is not tiled, take the first tiled one.
-    if (not focus or focus.floating) and #cls > 0 then
+    if not focus or focus.floating then
         focus = cls[1]
         fidx = 1
     end
```

WhosCounting:
Hey, that works! Good catch, thanks.

PR #2046, opened 2017-09-28
magnifier layout: Never manage floating windows

psychon:
The magnifier layout handles the currently focused client specially.
However, if the currently floating client is floating, it should not be
handled by the layout at all. A bug caused the magnifier layout to
handle a focused and floating client anyway if it was the only tiled
client.

Fix this by removing the '#cls > 0'-case. If #cls == 0, then no client
is available to be managed. Thus, cls[1] will be nil, which is fine
since, well, no client is available to be managed. This only made a
difference in the specific bug that I described above. Thus, drop this
case.

Fixes: https://github.com/awesomeWM/awesome/issues/2045


blueyed:
Makes sense, thanks!

PR #2049, opened 2017-09-29
Sort list of files used in generated table

reinerh:
File order depends on file system, kernel version etc...
To build the documentation reproducibly, the file list used for generating the table needs to be sorted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
   matrix:
     - LUA=5.2 LUANAME=lua5.2 DO_COVERAGE=coveralls
     # Lua 5.3 isn't available in Ubuntu Trusty, so some magic below installs it.
-    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov
+    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov EXPORT_SDE=true
     # luajit: installed from source.
     - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.5 TEST_PREV_COMMITS=1
     # Note: luarocks does not work with Lua 5.0.
@@ -156,6 +156,7 @@ install:
     }
 script:
   - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -DLUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -DSTRICT_TESTS=true -D DO_COVERAGE=$DO_COVERAGE"
+  - if [ "$EXPORT_SDE" = "true" ]; then export SOURCE_DATE_EPOCH=$(date -u -r docs/89-NEWS.md +%s); fi
   - |
     if [ "$EMPTY_THEME_WHILE_LOADING" = 1 ]; then
       # Break beautiful so that trying to access the theme before beautiful.init() causes an error
--- a/docs/06-appearance.md.lua
+++ b/docs/06-appearance.md.lua
@@ -95,6 +95,7 @@ local function get_link(file, element)
 end
 
 local all_files = get_all_files("./lib/", "lua")
+table.sort(all_files)
 
 local beautiful_vars = {}
 
--- a/tests/examples/shims/_date.lua
+++ b/tests/examples/shims/_date.lua
@@ -1,0 +1,11 @@
+local source_date_epoch = os.getenv("SOURCE_DATE_EPOCH")
+if source_date_epoch then
+    local old_osdate = os.date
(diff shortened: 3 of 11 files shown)


reinerh:
Added another commit, because generated svg images of the calendar widget were also not building reproducibly.

blueyed:
This is awesome, @awesome-robot will be pleased!

But your last commit caused all builds to fail - any idea why?

blueyed:
Likely a test that needs to be adjusted.

reinerh:
Thanks @blueyed, I didn't notice the failure. I'll try to fix it.

reinerh:
The updated commit should work now (the tests run locally are passing now).
Though now I still see an unreproducible date in the head AUTOGEN_wibox_widget_calendar_month.svg (the file wasn't there before because of the bug).

reinerh:
Hm, I can't find how the blue header in AUTOGEN_wibox_widget_calendar_month.svg is generated.
Anyone has an idea?

lib/wibox/widget/calendar.lua
```diff
@@ -117,6 +117,19 @@ local properties = { "date", "font", "spacing", "week_numbers", "start_sunday",
...
+    if not date then return nil end
+
+    local source_date_epoch = os.getenv("SOURCE_DATE_EPOCH")
+    if not source_date_epoch then return date end
+
+    local ret_date = os.date("!*t", source_date_epoch)
+    if date.month == nil then ret_date.month = nil end
+    if date.day == nil then ret_date.day = nil end
+    return ret_date
+end
```
psychon: Do we really want this in `lib/`? It would also make awesome itself honor `SOURCE_DATE_EPOCH`.
Could you instead please try to get something like the following to work?
```diff
diff --git a/tests/examples/wibox/widget/defaults/calendar.lua b/tests/examples/wibox/widget/defaults/calendar.lua
index ea57daad9..39f9e9255 100644
--- a/tests/examples/wibox/widget/defaults/calendar.lua
+++ b/tests/examples/wibox/widget/defaults/calendar.lua
@@ -1,3 +1,13 @@
+local source_date_epoch = os.getenv("SOURCE_DATE_EPOCH") --DOC_HIDE
+if source_date_epoch then --DOC_HIDE
+    local gtable = require("gears.table") -- DOC_HIDE
+    local ret_date = os.date("!*t", source_date_epoch)
+    os.date = function(arg) --
+        assert(arg == "*t", "Expected '*t', but got: " .. arg) -- DOC_HIDE
+        return gtable.clone(ret_date) -- DOC_HIDE
+    end --
+end --DOC_HIDE
+
 local parent    = ... --DOC_HIDE
 local wibox     = require("wibox") --DOC_HIDE
 local beautiful = require( "beautiful" ) --DOC_HIDE
```
I guess since many tests need this, it would be best to have this code in some central file that is then `require()`d by the other files.

What do you think? Do you want to do this?

(Also: Thanks for showing how easy it is to use this variable in Lua! What's your context? Do you come from the debian reproducible build crowd? Since you are doing this work, it must be something like this, right?)

psychon:
> Hm, I can't find how the blue header in AUTOGEN_wibox_widget_calendar_month.svg is generated.
Anyone has an idea?

https://github.com/awesomeWM/awesome/blob/a6613c08a9eedc716036bcb7fc917c9cf992aba5/tests/examples/wibox/widget/calendar/month.lua#L19-L26
https://github.com/awesomeWM/awesome/blob/a6613c08a9eedc716036bcb7fc917c9cf992aba5/tests/examples/wibox/widget/calendar/month.lua#L33-L42

reinerh:
Thanks for your hints. I like your approach better, as it touches only tests/ and not the running code.
I had to modify it a bit, as os.date is also called in the calendar widget with different formats and timestamps as parameter.
I hope the place where I put the code is okay. Feel free to move it somewhere else.

> What's your context? Do you come from the debian reproducible build crowd?

Yes, I do, good guess. :-) I'm also currently updating the package to 4.2.

blueyed:
Great.
Do you think you can cover https://codecov.io/gh/awesomeWM/awesome/compare/a6613c08a9eedc716036bcb7fc917c9cf992aba5...46ffd1622bf0140453032fd8d240e01775463a00/src/tests/examples/shims/_date.lua also?

reinerh:
@blueyed Hm, I'm not sure how to cover that. As far as I can see, the tests in tests/test*.lua are testing/covering real functionality (widgets etc.).
Should the test code go to one of the example files (I didn't see any assertions there), but stay hidden from documentation with the DOC_HIDE tag? Though this doesn't feel like the right place...

psychon:
@blueyed Are you just hoping that `_date.lua` runs on Travis so that we notice when/if it breaks? Something like that? For that, shouldn't it be enough to add `export SOURCE_DATE_EPOCH=$(date +%s)` to `travis.yml` (well, perhaps only to some of the builds so that both "variable there" and "variable not there" are tested)?

blueyed:
@psychon 
Yes, makes sense.

@reinerh 
Can you update the Travis config to add the env for one of the jobs, please?

blueyed:
Thank you!

Issue #2050, opened 2017-09-29
High memory usage with client.shape


renatofdds:
Output of `awesome --version`:

awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1

**How to reproduce the issue:**

Add to awesome configuration:
```lua
	client.connect_signal(
		"manage",
		function (c)
			c.shape = function(cr,w,h)
			    gears.shape.rounded_rect(cr,w,h,3)
			end
		end
	)
```

**Actual result:**

Awesome consumes about 35MB of memory with every client spawned independent from the client program, e.g. with 15 clients memory usage is 700MB.

**Expected result:**

Memory usage close to the usage without rounded borders, e.g. with 15 clients i get about 40MB.

psychon:
Does running `awesome-client 'local a = collectgarbage("count") collectgarbage("collect") collectgarbage("collect") return a, collectgarbage("count")'` in a terminal help? What does it output?

renatofdds:
It doesn't fix it, although it show an interesting behaviour. When i run it, the memory doesn't drop, but if i spawn new clients it doesn't go up unless i spawn more clients then i used to get to that point. 

Example:
- Start awesome: ~40MB
- Spawn 5 clients: ~400MB
- Run the command, output:
```
   double 13515.3
   double 6169.28
```
- Spawn up to 5 *more* clients: ~400MB
- Spawn 6th(11th) client: 422MB
- Spawn 7th(12th) client: 447MB

And so on...

psychon:
Hm, okay. Sorry that I didn't do this yesterday (didn't have much time), but now I actually tested it. I opened a terminal and ran `for x in $(seq 0 40) ; do urxvt -e sleep 1 & sleep 0.2 ; done`. Awesome's peak memory usage was about 50 MiB residential according to top, so I'll claim that I cannot reproduce this.

In fact, I now remembered #1960 and thus I think this was already fixed / improved.

psychon:
I ran `awesome` under `valgrind --tool=massif` with `DISPLAY=:5 ; for x in $(seq 0 1000) ; do urxvt -e sleep 5 & sleep 0.3 ; done` (note: not the full 1000 iterations were done; in fact I have no idea how many clients were opened). The resulting memory usage graph looks like this:
```
    MB
[...]

PR #2051, opened 2017-10-01
Use memory shorter with awful.client.shape

psychon:
To apply the shape of a client, we have to create an image and draw the
shape we want to it. Since clients can be quite large, we have to make
sure that we do not keep this image alive unnecessarily long.

The code in awful.client.shape.get_transformed() however needs another
temporary surface in case the client has its own shape and another one
was set in Lua (side note: currently the code also creates this extra
temporary surface if the client does not have its own shape; that might
be worth fixing). This temporary surface is then used as the source of a
cairo context to draw it to the image that will be used as the client's
final source.

After we are done, the temporary surface is still kept alive since it is
the current source of the cairo context. The cairo context in turn is
only freed when Lua's garbage collector collects it, which may take
quite a while.

Improve this by setting a different source to the cairo context. Thus,
it now releases the temporary surface as soon as possible and it is only
allocated for a short time.

Fixes: https://github.com/awesomeWM/awesome/issues/2050

### DIFF
--- a/lib/awful/client/shape.lua
+++ b/lib/awful/client/shape.lua
@@ -80,6 +80,10 @@ function shape.get_transformed(c, shape_name)
         cr:pop_group_to_source()
         cr:set_operator(cairo.Operator.IN)
         cr:paint()
+
+        -- 'cr' is kept alive until Lua's GC frees it. Make sure it does not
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!

PR #2053, opened 2017-10-07
Mixed-DPI groundwork

Oblomov:
Some more groundwork to improve support for mixed-DPI setups.

It pre-computes the per-screen DPI, while still allowing overrides, it allows choosing a rounding method for the used DPI, and font loading now allows an associated screen, to be used for height computation.

(Some of these make more sense view in terms of the subsequent fixes spread around the lib part, the work is pushed to my `mixed-dpi` branch, but still needs a more thorough review.)

### DIFF
--- a/lib/beautiful/init.lua
+++ b/lib/beautiful/init.lua
@@ -103,8 +103,9 @@ local active_font
 -- @see https://developer.gnome.org/pango/stable/pango-Fonts.html#pango-font-description-from-string
 -- @tparam string|lgi.Pango.FontDescription name Font, which can be a
 --   string or a lgi.Pango.FontDescription.
+-- @tparam screen (optional) screen to get DPI information about
 -- @treturn table A table with `name`, `description` and `height`.
-local function load_font(name)
+local function load_font(name, screen)
     name = name or active_font
     if name and type(name) ~= "string" then
         if descs[name] then
@@ -113,14 +114,15 @@ local function load_font(name)
             name = name:to_string()
         end
     end
-    if fonts[name] then
-        return fonts[name]
+    local fonts_key = {name, screen}
+    if fonts[fonts_key] then
+        return fonts[fonts_key]
     end
 
     -- Load new font
     local desc = Pango.FontDescription.from_string(name)
     local ctx = PangoCairo.font_map_get_default():create_context()
-    ctx:set_resolution(beautiful.xresources.get_dpi())
+    ctx:set_resolution(beautiful.xresources.get_dpi(screen))
 
     -- Apply default values from the context (e.g. a default font size)
     desc:merge(ctx:get_font_description(), false)
@@ -136,7 +138,7 @@ local function load_font(name)
     end
 
     local font = { name = name, description = desc, height = height }
-    fonts[name] = font
+    fonts[fonts_key] = font
     descs[desc] = name
     return font
 end
@@ -174,8 +176,9 @@ end
 --- Get the height of a font.
 --
 -- @param name Name of the font
-function beautiful.get_font_height(name)
-    return load_font(name).height
+-- @param screen (optional) screen to compute height for
+function beautiful.get_font_height(name, screen)
+    return load_font(name, screen).height
 end
 
 --- Init function, should be runned at the beginning of configuration file.
--- a/lib/beautiful/xresources.lua
+++ b/lib/beautiful/xresources.lua
@@ -75,8 +75,8 @@ end
 -- @treturn number DPI value.
 function xresources.get_dpi(s)
     s = get_screen(s)
-    if dpi_per_screen[s] then
-        return dpi_per_screen[s]
+    if s then
+        return dpi_per_screen[s] or s.dpi
     end
     if not xresources.dpi then
         -- Might not be present when run under unit tests
--- a/lib/beautiful/xresources.lua
+++ b/lib/beautiful/xresources.lua
@@ -145,6 +145,7 @@ end
 -- @tparam[opt] integer|screen s The screen.
 -- @treturn integer Resulting size (rounded to integer).
 function xresources.apply_dpi(size, s)
+    if not size then return size end
     local scale = xresources.get_dpi(s)/96
     return round(size * scale)
 end
--- a/lib/beautiful/xresources.lua
+++ b/lib/beautiful/xresources.lua
@@ -140,6 +140,9 @@ local function recalc_auto_dpi()
     end
 end
 
+screen.connect_signal("property::geometry", recalc_auto_dpi)
+screen.connect_signal("property::outputs", recalc_auto_dpi)
+
 
 --- Get global or per-screen DPI value falling back to xrdb.
 -- @tparam[opt] integer|screen s The screen.
--- a/lib/beautiful/xresources.lua
+++ b/lib/beautiful/xresources.lua
@@ -66,8 +66,19 @@ end
 
 local dpi_per_screen = {}
 
+local dpi_scale_rounding = nil
+
 local function get_screen(s)
     return s and screen[s]
+end
+
+local function rounded_dpi(dpi)
+    if dpi_scale_rounding then
+        local rounded = 96*dpi_scale_rounding(dpi/96)
+        return rounded
+    else
+        return dpi
+    end
 end
 
 --- Get global or per-screen DPI value falling back to xrdb.
@@ -76,7 +87,7 @@ end
 function xresources.get_dpi(s)
     s = get_screen(s)
     if s then
-        return dpi_per_screen[s] or s.dpi
+        return dpi_per_screen[s] or rounded_dpi(s.dpi)
     end
     if not xresources.dpi then
         -- Might not be present when run under unit tests
@@ -94,7 +105,7 @@ function xresources.get_dpi(s)
                 local _, h = root.size()
                 local _, hmm = root.size_mm()
                 if hmm ~= 0 then
-                    xresources.dpi = round(h*mm_to_inch/hmm)
+                    xresources.dpi = rounded_dpi(h*mm_to_inch/hmm)
                 end
             end
         end
@@ -119,13 +130,23 @@ function xresources.set_dpi(dpi, s)
     end
 end
 
+--- Set a rounding mode for autodetected DPIs
+--
+-- If set, beautiful will round the screen DPI to a multiple of the reference
+-- 96 DPI, according to the function used
+-- @tparam function a function to be applied to the DPI scaling factor
+function xresources.set_dpi_rounding(func)
+    dpi_scale_rounding = func
+end
+
 
 --- Compute resulting size applying current DPI value (optionally per screen).
 -- @tparam number size Size
 -- @tparam[opt] integer|screen s The screen.
 -- @treturn integer Resulting size (rounded to integer).
 function xresources.apply_dpi(size, s)
-    return round(size / 96 * xresources.get_dpi(s))
+    local scale = xresources.get_dpi(s)/96
+    return round(size * scale)
 end
 
 return xresources
--- a/lib/beautiful/xresources.lua
+++ b/lib/beautiful/xresources.lua
@@ -64,6 +64,9 @@ function xresources.get_current_theme()
 end
 
 
+--- Store the cached or user-set per-screen DPI
+-- Values are tables with key dpi (the dpi) and set_from (user, auto, xrdb)
+-- The same is done for xresources.dpi
 local dpi_per_screen = {}
 
 local dpi_scale_rounding = nil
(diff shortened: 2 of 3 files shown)


tests/test-screen-changes.lua
```diff
@@ -93,9 +93,7 @@ local steps = {
         assert(test_client2.screen == real_screen, test_client2.screen)
 
         collectgarbage("collect")
-        if #fake_screen == 0 then
-            return true
-        end
+        return #fake_screen == 0
```
psychon: The test runner (`tests/_runner.lua`) does the following after a step ran:
- If it returned true, it continues with the next step
- If it returns false, the whole test is considered to have failed
- If it returns nil, this step is retried after 0.1 seconds, but only 5 times. Then the test is considered to have failed.

So, before this commit, the garbage collector had up to 5 calls to `collectgarbage("collect")` time to get rid of the old screen. Now the screen must be removed on the first collection cycle.

I guess this change is the reason why the Travis build failed. Also, this cannot stall, since the test runner only tries up to five times and then gives up.

Please remove this commit.

lib/beautiful/xresources.lua
```diff
@@ -75,8 +75,8 @@ end
 -- @treturn number DPI value.
 function xresources.get_dpi(s)
     s = get_screen(s)
-    if dpi_per_screen[s] then
-        return dpi_per_screen[s]
+    if s then
+        return dpi_per_screen[s] or s.dpi
```
psychon: If I query the DPI of some screen (which has no outputs), then the result will always be 96, since this is what the `get_dpi` function above returns. Shouldn't this rather use some of the calculations below instead if no outputs are available?
So, my proposal:
```lua
if dpi_per_screen[s] then
    return dpi_per_screen[s]
end
if #s.outputs > 0 then
    return s.dpi
end
```

psychon: And the next problem: People who have wrong dimensions in their EDID data and overrode this via `Xft.dpi` will now have `Xft.dpi` being ignored. Since `Xft.dpi` is some explicit configuration, I think it should have higher priority than auto-configuration.

Meaning that I would propose the following order of settings:
- Forced settings via Lua: `dpi_per_screen[s]`
- Forced settings via xrdb: `Xft.dpi`
- Automatically computed from outputs: `s.outputs`
- Automatically computed from the size of the root window (the code below)
- Good old `96`

lib/beautiful/xresources.lua
```diff
@@ -119,13 +130,23 @@ function xresources.set_dpi(dpi, s)
     end
 end
 
+--- Set a rounding mode for autodetected DPIs
+--
+-- If set, beautiful will round the screen DPI to a multiple of the reference
+-- 96 DPI, according to the function used
+-- @tparam function func a function to be applied to the DPI scaling factor
+function xresources.set_dpi_rounding(func)
+    dpi_scale_rounding = func
```
psychon: Sorry, but this function is useless. It gets as argument the value `dpi/96`. If you want to e.g. ceil this, you could just do `set_dpi(s, 96*math.ceil(get_dpi(s)/96))` in your rc.lua (or just compute the wanted DPI value by hand). So, this does not add any new possibilities.

Put differently: Isn't it enough to be able to override DPI per-screen?

If you want a rounding function, I would more expect the user to override the `round` call in `apply_dpi`. However, that function must return integers, so this might not be all that useful either.

lib/beautiful/xresources.lua
```diff
@@ -145,6 +145,7 @@ end
 -- @tparam[opt] integer|screen s The screen.
 -- @treturn integer Resulting size (rounded to integer).
 function xresources.apply_dpi(size, s)
+    if not size then return size end
```
psychon: Why is this needed? Which code would pass a `nil` (or `false`) here?

lib/beautiful/xresources.lua
```diff
@@ -64,6 +64,9 @@ function xresources.get_current_theme()
 end
 
 
+--- Store the cached or user-set per-screen DPI
+-- Values are tables with key dpi (the dpi) and set_from (user, auto, xrdb)
+-- The same is done for xresources.dpi
```
psychon: What this commit actually does is to add a cache to DPI values. Why? Isn't the existing state of recomputing this "all the time" good enough? Do you have some example where adding a cache here speeds something up significantly?

lib/beautiful/init.lua
```diff
@@ -113,14 +115,15 @@ local function load_font(name)
             name = name:to_string()
         end
     end
-    if fonts[name] then
-        return fonts[name]
+    local fonts_key = {name, screen}
+    if fonts[fonts_key] then
+        return fonts[fonts_key]
```
psychon: Whoops. I totally forgot about this function ignoring DPI.

Anyway, the above cache check does not work. Lua tables are only equal to themselves, i.e. the index operator does not do a deep comparison:
```
$ lua -e 't={} t[{}] = 42 print(t[{}])'
nil
```
Thus, the above doesn't cache check does not work.

Also, I wonder if it makes sense to have a version of this function that gets DPI as argument and another that gets a screen object, gets the screen's DPI and then just calls the other function (same for `get_font_height`). Thoughts?

(At least the cache only needs the DPI, not the screen itself)

tests/test-screen-changes.lua
Oblomov: The Travis build was failing even before I introduced this commit, due to (I think) references to screen being held in the new font cache. But thanks for the explanation, I'll remove the commit.

lib/beautiful/xresources.lua
Oblomov: in my `mixed-dpi` branch I remove a number of uses of the global `apply_dpi`, shifting the computation of the dpi scaling at use time rather than define time. There are a few cases in which the value is optional with a number of fallbacks that can all be nil (`local somevar = args.stuff or default.stuff or blah.stuff`). Rather than leaving that and then adding a new `somevar = somevar and dpi(somevar, s) or nil` line I decided to just wrap everything in a dpi(..., s), which then needs to pass through nil values.

psychon: Okay, but I guess then this behaviour should be added to the documentation of this function.

Oblomov:
Ping @psychon @blueyed @Elv13 review?

lib/gears/dpi.lua
```diff
@@ -0,0 +1,32 @@
...
+    if root then
+        local mm_to_inch = 25.4
+        local _, h = root.size()
+        local _, hmm = root.size_mm()
+        if hmm ~= 0 then
+            dpi = h*mm_to_inch/hmm
+        end
+    end
+    return dpi
+end
```
psychon: Is this a good idea to have? If we want to have per screen DPIs, there should be no "global" DPI that could accidentally be used in code instead.

lib/beautiful/init.lua
```diff
@@ -98,6 +100,41 @@ local active_font
 --- The current theme path (if any)
 -- @tfield string beautiful.theme_path
 
+--- Font DPI API
+-- @section font_dpi
```
psychon: This might be a good idea why for fonts the number of pixels per inch of your monitor is different than for non-fonts.

lib/beautiful/init.lua
```diff
@@ -133,6 +133,21 @@ function beautiful.set_font_dpi(dpi)
...
+-- On uniform DPI environments, this is a no-op, but on mixed-DPI environments
+-- it ensures that font scaling is uniform across different DPIs.
+-- (This approach follows the one used in Qt.)
+--
+-- @tparam[opt] screen scr The screen to get font DPI information about.
+--     Defaults to the primary screen
+function beautiful.screen_font_dpi(scr)
+    scr = scr and screen[scr] or screen.primary
+    return beautiful.get_font_dpi()*scr.dpi/screen.primary.dpi
+end
```
psychon: @Elv13 What do you think of this?

lib/beautiful/init.lua
```diff
@@ -255,7 +256,7 @@ end
 -- @tparam[opt] screen scr Screen to get the DPI information from.
 --   If unspecified, the global font DPI setting will be used.
 function beautiful.get_font_height(name, scr)
-    return load_font(name, beautiful.xresources.get_dpi(scr)).height
+    return load_font(name, beautiful.screen_font_dpi(scr)).height
```
psychon: Hm, this opens quite a can of worms.
Primary screen changed? Font height changed on all screens!
Primary screen changed DPI value? Font height changed on all screens!
Some random screen changed DPI value? Font height changed for that screen.

I bet that nothing will handle these correctly.

lib/gears/dpi.lua
```diff
@@ -7,8 +7,87 @@
...
+-- reference pixel density (96). In such cases, users' preferences vary:
+-- some prefer using the DPI values as-is, resulting in fractional scaling,
+-- others prefer rounding the scaling factor (so that e.g. anything above 144
+-- DPI and below 240 should result in a scaling factor of 2).
+-- While this could be achieved by the users by manually forcing specific DPI
+-- values for the screens, it's rather annoying in dynamic configurations.
+-- So we offer the possibility for the user to choose a global policy,
+-- and all DPI usage will take this into consideration.
+-- Note that rounding is only applied to autocomputed DPI values, never
+-- when the user sets the DPI themselves.
```
psychon: I can't say I'm convinced that this is a good idea, but at least the default is `identity`.
@Elv13 What do you think?

tests/test-dpi.lua
```diff
@@ -0,0 +1,100 @@
...
+            assert(dpi.core_dpi() >= 192)
+        end
+
+        return true
+    end
+}
+
+runner.run_steps(steps)
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
```
psychon: Yay for tests! :-)

I think this should be done as an unit test instead of integration (i.e. as `spec/gears/dpi_spec.lua`). `gears.dpi` does not use much APIs, so that can easily be mocked (it just needs `root`, I think).

lib/wibox/widget/textbox.lua
```diff
@@ -73,13 +73,10 @@ end
...
 function textbox:get_preferred_size(s)
-    local dpi
-    if s then
-        dpi = screen[s].dpi
-    else
+    if not s then
         gdebug.deprecate("textbox:get_preferred_size() requires a screen argument", {deprecated_in=5, raw=true})
-        dpi = beautiful.xresources.get_dpi()
     end
+    local dpi = beautiful.screen_font_dpi(s)
```
psychon: This changes the semantics of which DPI value is used when this function is called with `s=nil`. Previously, it used some concept of "core DPI" while now it uses the DPI of the primary screen.
@Elv13 Does that count as a regression?

lib/beautiful/xresources.lua
```diff
@@ -103,7 +103,11 @@ end
 -- @tparam[opt] integer|screen s The screen.
 -- @treturn integer Resulting size (rounded to integer).
 function xresources.apply_dpi(size, s)
-    return round(size / 96 * xresources.get_dpi(s))
+    if s then
+        return s:dpi_scale(size)
+    else
+        return round(size * require("beautiful").screen_font_dpi()/96)
```
psychon: Why does this use font DPI? If you just use the font DPI everywhere, then `s.dpi` is not needed.

lib/beautiful/init.lua
psychon: Since I fail to parse my own sentence: The docs here could explain the concept for "fonts get a different DPI" a bit here.

lib/wibox/widget/textbox.lua
```diff
@@ -73,13 +73,10 @@ end
 -- @treturn number The preferred width.
 -- @treturn number The preferred height.
 function textbox:get_preferred_size(s)
-    local dpi
-    if s then
-        dpi = screen[s].dpi
-    else
+    if not s then
```
psychon: What is with `:fit` and `:draw`? This commits make the "non-widget" functions use a different DPI value than the actual drawing does.

lib/beautiful/xresources.lua
Oblomov: There's an issue about “what has this call being used for in legacy code” and “how do we transition from there to the API we want now”.

Among other things, this means that the only case where you access the DPI without a screen specification is if you want to know the user-specified font scaling, since there is (or should not be) any concept of “global DPI” (otherwise).

tests/test-dpi.lua
Oblomov: I'll look into this.

psychon: Feel free to ping me if needed.

PR #2055, opened 2017-10-08
screen.dpi: Use RandR for per-screen DPI

psychon:
This commit makes awesome automatically compute the DPI of a screen
based on its RandR outputs. If multiple outputs exist, the lowest DPI is
used.

CC @blueyed This pretty much supersedes your #1204, is that okay with you?
CC @Oblomov

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -525,6 +525,18 @@ function screen.object.get_dpi(s)
         return xft_dpi
     end
 
+    -- Try to compute DPI based on outputs (use the minimum)
+    local dpi = nil
+    local geo = s.geometry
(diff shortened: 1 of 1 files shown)


Elv13:
Seems fine, but I wonder if some widgets start abusing this function if it could not be cached better. The loop is always rather small (not many users have 6+ screens), but still. If this ever gets "popular", then it might get called in very event loop/redraw.

In any case, it can be done when it becomes a problem instead of right now, so feel free to merge this.

Oblomov:
@Elv13 keep in mind that the loop is on all outputs _for that screen_, which is typically exactly 1. It's only in clone mode that one might get more outputs per screen.

PR #2056, opened 2017-10-08
Fix menubar screen handling

psychon:
Running test-menubar.lua just failed for me locally. The reason was that
the textbox was indexing a nil value when doing 's.dpi'.

This commit fixes the menubar so that its optional screen arguments are
really handled correctly. In fact, a default screen is now chosen way
earlier than before, so that no nil values are used as screens later on.

In other news: I guess test-menubar.lua is not run on any of our Travis
targets...

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -340,6 +340,7 @@ end
 --- Refresh menubar's cache by reloading .desktop files.
 -- @tparam[opt] screen scr Screen.
 function menubar.refresh(scr)
+    scr = get_screen(scr or awful.screen.focused() or 1)
     menubar.menu_gen.generate(function(entries)
         menubar.menu_entries = entries
         if instance then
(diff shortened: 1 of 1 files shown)


blueyed:
> In other news: I guess test-menubar.lua is not run on any of our Travis
targets...

Ref: https://github.com/awesomeWM/awesome/blob/d5f3315163c5725dc5fe474a9294be404ce2aa89/tests/test-menubar.lua#L20-L24

blueyed:
Ah, seems like you've tried to enable it via https://github.com/awesomeWM/awesome/pull/2058.

psychon:
Yup, so the test is not run on Lua 5.3, when collecting coverage, or on LuaJIT. I thought that we had some build which was not excluded (yet), but apparently not...? Looking at `.travis.yml`, I think this one should actually run the tests, no? `- LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build`

PR #2057, opened 2017-10-08
textbox: Make some screen arguments optional again

psychon:
Since beautiful.xresources.get_dpi(s) allows nil as argument to query
some "global DPI", the functions to query the size of a textbox also
allowed to use nil as the screen, even though this was never documented
to work. Commit a1376557917a88 broke this by assuming a valid screen
object.

This commit makes a nil argument work again, but will cause a
deprecation warning in awesome 5.

### DIFF
--- a/lib/wibox/widget/textbox.lua
+++ b/lib/wibox/widget/textbox.lua
@@ -73,7 +73,14 @@ end
 -- @treturn number The preferred width.
 -- @treturn number The preferred height.
 function textbox:get_preferred_size(s)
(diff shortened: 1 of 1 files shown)


Elv13:
Maybe add a regression test?

psychon:
Would that test then be marked as something like deprecated and just be removed in awesome v6, or what is the plan?

psychon:
I took a look: Both of these function currently have no tests and I'm not really sure how to test them. They depend on the available fonts and thus we could just check that the result is a number... is that worth it?

PR #2060, opened 2017-10-15
Send notification when error occurs in prompt widget

### LINKED ISSUE #1282: eternal message in prompt
I've just switched to the git version of awesome (I was using v3.5.9) and I realized that when I open the prompt box and press enter (with the box empty), the message "Error: No command to execute" stays forever in the prompt (unless I try to type something else there).

Is it done on purpose ? I remember that this error message in prompt used to be gone in a few seconds in the older version. 

How to reproduce: `modkey-r + Enter`

reinerh:
If there was an error, it was stuck in the prompt textbox
until a new command was entered.
By sending a notification the error will time out and disappear
after a few seconds.
A notification also makes it more visible that something went wrong.

Fixes #1282


lib/awful/widget/prompt.lua
```diff
@@ -46,10 +46,11 @@ local function run(promptbox)
     }
 end
 
-local function spawn_and_handle_error(self, ...)
+local function spawn_and_handle_error(_, ...)
```
actionless: mb better turn it from local function to `prompt.spawn_and_handle_error`?
so other users would be able to redefine the error handler behavior (if they want, let's say, to return the old behavior)

PR #2066, opened 2017-10-20
Add is_immobilized property to ignore resize requests

blueyed:


lib/awful/client.lua
```diff
@@ -634,6 +634,24 @@ function client.object.is_fixed(c)
...
+-- This property is read only.
+-- @property is_immobilized
+-- @param boolean The immobilized state
+-- @see maximized
+-- @see maximized_horizontal
+-- @see maximized_vertical
+-- @see fullscreen
+
+function client.object.is_immobilized(c)
+    return c.fullscreen or c.maximized or c.maximized_horizontal or c.maximized_vertical
```
psychon: Horizontally maximized clients should only be horizontally immobilized, I think. So, this needs to be split up into two properties?

tests/test-maximize.lua
```diff
@@ -38,8 +38,10 @@ for _, gravity in ipairs { "NORTH_WEST", "NORTH", "NORTH_EAST", "WEST",
         assert(not c.maximized_vertical  )
         assert(not c.maximized           )
         assert(not c.fullscreen          )
+        assert(not c.is_immobilized      )
```
psychon: This assertion fails on Travis (I only checked one built, but I guess all of them).

blueyed: `c.is_immobilized` is a function and not a boolean when printing it.
Guess I will figure it out with @Elv13 later.

psychon: Uhm, right. `is_` is one of the prefixes that is recognized just like `get_` and `set_`. So, the function needs to be called `is_is_immobilized`. But actually, stop using `is_immobilized` and call the property `immobilized` since the prefix `is_` seems unnecessary.

blueyed:
> test-gravity fails on Travis

@Elv13 mentioned something with regard to the C test client (or this test), where it waits endlessly now, since the geometry is not changed anymore due to this change.

blueyed:
@psychon 
I could some help with tests for this.. initially I've thought about using tests/test-maximize.lua for this, but then looked into using busted instead.
Requiring `awful.ewmh` requires a lot of mocking though (incomplete):

```lua
describe("awful.ewmh.client_geometry_requests", function()
    _G.awesome = {
        connect_signal = function() end,
        register_xproperty = function() end,
    }
    _G.client = {
        connect_signal = function() end,
        get = function() return {} end,
    }
    local ewmh = require("awful.ewmh")

    it("removes x and width when immobilized_horizontal", function()
        local c = stub.new()

        local context = {}
        ewmh.client_geometry_requests(c, "ewmh", context)
    end)
end)
```

psychon:
>> test-gravity fails on Travis
>
> @Elv13 mentioned something with regard to the C test client (or this test), where it waits endlessly now, since the geometry is not changed anymore due to this change.

Uhm, why? The test should run on a floating layout and the window shouldn't be maximized/fullscreen.

> Requiring awful.ewmh requires a lot of mocking though (incomplete):

As far as I see (I only scrolled through the file quickly), `awful.ewmh` itself only requires `client` to be mocked. Is the other part that you need for other modules, e.g. in `awful.placement`? If so, you could prevent them being loaded via `package.loaded["awful.placement"] = {}` (and similarly get rid of / mock other dependencies). Would this simplify things?

blueyed:
Thanks @psychon, that helped.

https://github.com/awesomeWM/awesome/pull/2066/commits/e63d3e11cfc792fa6fc18c3faee4009738cad71d

blueyed:
@psychon 
Please review again.

lib/awful/client.lua
```diff
@@ -634,6 +634,42 @@ function client.object.is_fixed(c)
     return false
 end
 
+--- Is the client immobilized horizontally, i.e. fullscreen or maximized?
```
psychon: I'd change this to `Does the client have a fixed horizontal position and width, i.e. is it fullscreen, maximized, or horizontally maximized?`

(I guess: same for `is_immobilized_horizontal`).

lib/awful/client.lua
```diff
@@ -634,6 +634,42 @@ function client.object.is_fixed(c)
...
+--
+-- **Signal:**
+--
+--  * *property::immobilized_horizontal*
```
psychon: Correct me if I'm wrong, but this claim of a signal is a lie?

lib/awful/ewmh.lua
```diff
@@ -373,7 +373,7 @@ end
 -- @tparam string context The context
 -- @tparam[opt={}] table hints The hints to pass to the handler
 function ewmh.client_geometry_requests(c, context, hints)
-    if context == "ewmh" and hints then
+    if context == "ewmh" and hints and not c.immobilized then
```
psychon: Huh? What happened to the code that checked the individual properties? Something like
```lua
local geo = {}
if not c.immobilized_horizontal then
    geo.x = hints.x
    geo.width = hints.width
end
if not c.immobilized_vertical then
    geo.y = hints.y
    geo.height = hints.height
end
c:geometry(geo)
```

lib/awful/client.lua
blueyed: Moved to the long desc.

lib/awful/client.lua
blueyed: Isn't this for all properties?  (I assumed, but apparently there is also no test for it).

blueyed: You are right - there is no setter here.  Will remove it.

blueyed:
Squashed everything.

blueyed:
Thanks for your help!

lib/awful/client.lua
psychon: Well, things like "fullscreen" are explicitly changed (`c.fullscreen = true`). For that case, there is some magic in place that automatically emits a signal, I think. However, this immobilized property is really a combination of other signals (fullscreen, maximized, maximized_h/v) and the magic cannot know that the getter will now suddenly return another value just because some other property changed.

Elv13: Do you want such magic? There is a few other places where it could be used (like `awful.layout.init`). Called `depends_on` or something. It would still not be able to detect if the property *didn't* change, but it would always send the signal if it *might* have changed.

psychon: Only if we have several places that need it. So far, `immobilized_v/h` are the only two examples that I think of and right now they do not even have a signal (and I can't imagine anyone requiring it). So I'd keep things simple for now.

Issue #1373, opened 2017-01-09
List of small documentation issues


Elv13:
Missing/misleading documentation. Gathered from IRC/Reddit/StackOverflow/Issues. This issue keeps a list until they are fixed.

 * <s>`wibox.layout.align` is missing `:remove()`, swap and others</s> (fix: #2067)
 * the wallpaper doc is not noob friendly
    * how to use a solid color
    * how and when to disable the default callback to do it manually
    * use words new users will understand (as they might look there first)
 * <s>client.icon is confusing and talks only about surfaces</s> (fix: #2067)
 * <s>screen.index: "An integer greater than 1 and smaller than screen.count()." (Jarcode) doesn't like it.</s>  (fix: #2067)
 * lack of header explainations in some classes
 * <s>document that `awful.wibox` is deprecated, but it isn't the same as `wibox` itself.</s> (fix: #2067)
 * add more example of imperative widgets declaration
 * add more examples of setters vs. properties
 * add wibox/wibars examples
 * <s>CAPI objects lack the connect/emit signal doc</s> (fix: #2067)
 * all those `object.*` pages are references are useless and should be removed
 * there is no nice list of properties for `awful.rules`
 * add a glossary
 * say it is possible to change the theme variables from `rc.lua`
 * Add the widget section to 05-awesomerc about what is what (with a screenshot)
 * what is a tag (concept, comparison, cardinalities, different usages)
    * as static workspaces
    * as dynamic workspaces
    * as activities
    * as ephemeral views
 * `gears.shape` function have "module" in their name
 * fake_input need 2 or 3 examples
 * `awful.key` doc should be merged with `key`
 * `awful.button` Lack parameter doc and info about pass though
 * `wibox`/init.lua is missing the signals section 
 * a signal is called `screen.removed` instead of .`removed`
 * document that `awful.tag.add` doesn't select anything while `awful.tag.new` does (#1423, #1509)
 * the whole xproperty thing is half undocumented and has misleading
 * how to re-apply rules like 3.5 used to do when screen changes without restarting (awful.rules.apply)
 * Be more explicit that `connect_for_each_screen` is a relative to `connect_signal`, not a way to iterate screens in a function
 * The shape API talks too much about cairo contexts and width and height. It need to be clear that those are rarely necessary and will be auto-passed to the function.
 * Add a FAQ / "Read this when distressed" section to the declarative widget doc
    * Getter and setters create implicit properties
[...]

PR #2067, opened 2017-10-20
Small batch of doc fixes

blueyed:
Ref: https://github.com/awesomeWM/awesome/issues/1373

### DIFF
--- a/docs/common/cobject.ldoc
+++ b/docs/common/cobject.ldoc
@@ -1,0 +1,24 @@
+*/
+
+/** Disconnect from a signal.
+ * @tparam string name The name of the signal.
+ * @tparam function func The callback that should be disconnected.
+ * @function disconnect_signal
+ */
+
+/** Emit a signal.
+ *
+ * @tparam string name The name of the signal.
+ * @param ... Extra arguments for the callback functions. Each connected
+ *   function receives the object as first argument and then any extra
+ *   arguments that are given to emit_signal().
+ * @function emit_signal
+ */
+
+/** Connect to a signal.
+ * @tparam string name The name of the signal.
+ * @tparam function func The callback to call when the signal is emitted.
+ * @function connect_signal
+ */
+
+/*
--- a/docs/common/object.ldoc
+++ b/docs/common/object.ldoc
@@ -1,28 +1,30 @@
 
---- Disconnect to a signal.
--- @tparam string name The name of the signal
--- @tparam function func The callback that should be disconnected
+--- Disconnect from a signal.
+-- @tparam string name The name of the signal.
+-- @tparam function func The callback that should be disconnected.
 -- @function disconnect_signal
 
 --- Emit a signal.
 --
--- @tparam string name The name of the signal
+-- @tparam string name The name of the signal.
 -- @param ... Extra arguments for the callback functions. Each connected
---   function receives the object as first argument and then any extra arguments
---   that are given to emit_signal()
+--   function receives the object as first argument and then any extra
+--   arguments that are given to emit_signal().
 -- @function emit_signal
 
 --- Connect to a signal.
--- @tparam string name The name of the signal
--- @tparam function func The callback to call when the signal is emitted
+-- @tparam string name The name of the signal.
+-- @tparam function func The callback to call when the signal is emitted.
 -- @function connect_signal
 
---- Connect to a signal weakly. This allows the callback function to be garbage
--- collected and automatically disconnects the signal when that happens.
+--- Connect to a signal weakly.
+--
+-- This allows the callback function to be garbage collected and
+-- automatically disconnects the signal when that happens.
 --
 -- **Warning:**
 -- Only use this function if you really, really, really know what you
 -- are doing.
--- @tparam string name The name of the signal
--- @tparam function func The callback to call when the signal is emitted
+-- @tparam string name The name of the signal.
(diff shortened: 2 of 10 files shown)


blueyed:
I'd say to merge this as-is for now.

docs/common/object.ldoc
```diff
@@ -1,28 +1,30 @@
 
---- Disconnect to a signal.
--- @tparam string name The name of the signal
--- @tparam function func The callback that should be disconnected
+--- Disconnect a signal.
```
psychon: Perhaps "disconnect from a signal"? Is `property::foo` the signal, or is the connection between `property::foo` and the provided callback function the signal? I'd say the former, so `from a signal` would be correct.

objects/screen.c
```diff
@@ -112,14 +112,29 @@
...
- * The screen number.
+ * The internal screen number.
  *
- * An integer greater than 1 and smaller than `screen.count()`. Please note that
- * the screen order isn't always mirroring the screen logical position.
+ * * The indeces are a continuous sequence from 1 to `screen.count()`.
+ * * It is **NOT** related to the actual screen position relative to each
+ *   other.
+ * * 1 is **NOT** necessarily the primary screen.
+ * * When screens are added and removed indexes **CAN** change.
```
psychon: The plural from `index` is `indices`, I think. And I'd add a comma between `removed` and `indexes/indicies`.

objects/client.c
```diff
@@ -371,14 +371,33 @@
  */
 
 /**
- * The client icon.
+ * The client icon as a surface (mapped memory).
```
psychon: What is `mapped memory` supposed to mean? It confuses me here. It's a normal image surface, but I'd consider this an implementation detail.

objects/client.c
```diff
@@ -371,14 +371,33 @@
...
+ *
+ * This property holds the client icon closest to the size configured via
+ * @{awesome.set_preferred_icon_size}.
+ *
+ * It is a bitmap mapped in memory and not a file path nor a widget.
```
psychon: It is a cairo image surface?!

objects/client.c
```diff
@@ -371,14 +371,33 @@
...
+ *
+ * Typically you would want to use @{awful.widget.clienticon} to get this as a
+ * widget.
+ *
+ * If you want to copy it into a cairo surface you should do something along,
+ * since it would otherwise crash awesome in case it gets accessed after the
+ * client is removed.
```
psychon: Nope. You can just do `local s = gears.surface(c.icon)` and are good.
However, this `gears.surface`-call (must be called on the returned value to avoid memory leaks, must not be called more than once to avoid double-frees (= crashes)) is quite important and I'd rather add this more prominently here.

Elv13:
@blueyed I think you forgot to `git add` the second `.ldoc`. Once you have done that, I can take care of the fixups. Don't forget to `git pull` the branch for the fixup commit.

blueyed:
> I think you forgot to git add the second .ldoc

Snap.
I've accidentally `rm -rf`ed awesome yesterday, and only had an older backup.
So you also have to re-create this yourself, sorry.

objects/client.c
Elv13: Umm, yes, I remembered there was something to say, but that was a very long time ago when you explained that one to me. I forgot that peculiarity and instead used the copy to avoid getting the issue.

I am not sure however how relevant saying it must not be done twice is for an end user. There is no way to controls this for the user. If he/she/it uses a tasklist and an 3rd party alt-tab menu that happens to also create a surface, then the copy seems a better idea than a dead-end.

In the end both option suck and maybe the real one is to implement a `awful.client.object.get_icon` and rename the "real" `c.icon` to `c._icon` and handle this in Lua? The average user should not be able to double-free Awesome accidentally.

For now I will add both in the doc, but I consider this a bug.

docs/common/cobject.ldoc
```diff
@@ -0,0 +1,38 @@
...
+ * automatically disconnects the signal when that happens.
+ *
+ *
+ * **Warning:**
+ * Only use this function if you really, really, really know what you
+ * are doing.
+ * @tparam string name The name of the signal.
+ * @tparam function func The callback to call when the signal is emitted.
+ * @function weak_connect_signal
+ */
```
psychon: Uhm, the C API does not have `weak_connect_signal`...?

objects/screen.c
```diff
@@ -117,14 +117,29 @@
...
+ * * It is **NOT** related to the actual screen position relative to each
+ *   other.
+ * * 1 is **NOT** necessarily the primary screen.
+ * * When screens are added and removed indices **CAN** change.
+ *
+ * If you really want to keep an array of screens you should use something
+ * along:
+ *
+ *     local myscreens = setmetatable({}. {__mode="k"})
+ *     myscreens[ screen[1] ] = "mydata"
```
psychon: Why not just use `screen[1].mydata` directly instead? That has way less chance of causing problems with the garbage collector. (I guess this PR is so old already, that back then we did not yet have this feature?)

Elv13: > I guess this PR is so old already

Yes, ancient, this is a copy of object.ldoc with `/* */` instead of `--`. I will ajust both.

Before this PR, client:connect_signal was never in the documented API

blueyed:
Thank you!

PR #2069, opened 2017-10-21
doc: new theme

blueyed:
Via https://github.com/stevedonovan/LDoc/pull/287.

2017-10-21-212800_1251x1045_scrot

2017-10-21-214008_1251x1045_scrot

### DIFF
--- a/docs/ldoc.css
+++ b/docs/ldoc.css
@@ -288,3 +288,11 @@ pre .global { color: #c040c0; }
 pre .user-keyword { color: #800080; }
 pre .prompt { color: #558817; }
 pre .url { color: #272fc2; text-decoration: underline; }
+
+#navigation h1 {
+  background-image: url(images/AUTOGEN_wibox_logo_logo_and_name.svg);
+  height:160px;
+  background-position: 10px 10px;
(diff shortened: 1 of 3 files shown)


docs/ldoc.css
```diff
@@ -0,0 +1,298 @@
+body { 
+    color: #47555c;
+    font-size: 16px;
+    font-family: "Open Sans", sans-serif;
```
actionless: why not just 'Sans' to use the system font?

and the same for monospace above

blueyed: Makes sense to me.
Should be forwarded to https://github.com/stevedonovan/LDoc/pull/287 then.

actionless: copied the comment there

blueyed: Thanks.

Does 9480706 address your comment?

blueyed:
Canceled the build, so ignore the failure from there, please.

blueyed:
Will merge this soon also - we can then improve it from there, and/or roll it back altogether.

Elv13:
@blueyed Merge this if you feel it's ready, otherwise I will press the button later today.

PR #2071, opened 2017-10-22
Makefile: add explicit rule for Makefile to not "update" it

blueyed:
Typically the implicit rule below will trigger a "make -C build
Makefile" run, whenever build/Makefile is newer than our Makefile.

Ref: https://www.gnu.org/software/make/manual/html_node/Remaking-Makefiles.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,9 @@ distclean:
 	$(RM) -r $(BUILDDIR) $(TARGETS)
 	$(ECHO) " done"
 
+# Use an explicit rule to not "update" the Makefile via the implicit rule below.
+Makefile: ;
+
 %: $(BUILDDIR)/Makefile
 	$(ECHO) "Running make $@ in $(BUILDDIR)…"
 	$(MAKE) -C $(BUILDDIR) $@


psychon:
Does this have any effect? As far as I can see, the CMake-generated Makefile does not have a Makefile-rule. So since there is no rule for this file and it already exists, nothing actually happens...?
```
$ LC_ALL=C make --debug=a -C  .build-ZweiLED-x86_64-linux-gnu-6.3.0 Makefile
GNU Make 4.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Reading makefiles...
Reading makefile 'Makefile'...
Updating makefiles....
 Considering target file 'Makefile'.
  Looking for an implicit rule for 'Makefile'.
  Trying pattern rule with stem 'Makefile'.
  Trying implicit prerequisite 'Makefile,v'.
  Trying pattern rule with stem 'Makefile'.
  Trying implicit prerequisite 'RCS/Makefile,v'.
  Trying pattern rule with stem 'Makefile'.
  Trying implicit prerequisite 'RCS/Makefile'.
  Trying pattern rule with stem 'Makefile'.
  Trying implicit prerequisite 's.Makefile'.
  Trying pattern rule with stem 'Makefile'.
  Trying implicit prerequisite 'SCCS/s.Makefile'.
  No implicit rule found for 'Makefile'.
  Finished prerequisites of target file 'Makefile'.
 No need to remake target 'Makefile'.
Updating goal targets....
Pruning file 'Makefile'.
```

blueyed:
It causes less confusion at least.. :)

PR #2073, opened 2017-10-22
tests: examples: remove caching, add depends for ldoc

blueyed:


CMakeLists.txt
```diff
@@ -289,7 +289,7 @@ if(GENERATE_DOC)
     file(COPY ${SOURCE_DIR}/docs/aliases DESTINATION ${BUILD_DIR}/docs)
 
     add_custom_target(ldoc ALL
-        DEPENDS ${BUILD_DIR}/doc/index.html)
+        DEPENDS generate-examples ${BUILD_DIR}/doc/index.html)
```
psychon: Why exactly is this needed? And why is this dependency here and not on the `add_custom_command` in line 300?

tests/examples/CMakeLists.txt
```diff
@@ -287,43 +287,34 @@ endfunction()
...
+        # variable during the pre-processing.
+        # While at it, replace \" created by CMake by ',
+        # &quot; wont work in <code>.
+        string(REPLACE "\"" "'" ${TEST_NAME} ${ESCAPED_CODE_EXAMPLE})
+    endif()
+endforeach()
 
-add_custom_target(generate-examples ALL
-    DEPENDS ${EXAMPLE_DOC_GENERATED_FILES})
+add_custom_target(generate-examples DEPENDS ${EXAMPLE_DOC_GENERATED_FILES})
```
psychon: Since `EXAMPLE_DOC_GENERATED_FILES` is no longer set, this seems quite wrong, I think...?

(Is there a single commit adding all this that can just be reverted instead?)

tests/examples/CMakeLists.txt
```diff
@@ -287,43 +287,34 @@ endfunction()
...
 
 add_custom_target(check-examples
-    ${CMAKE_COMMAND} -D EXAMPLE_DOC_SOURCE_FILES= ${CMAKE_SOURCE_DIR}
+    ${CMAKE_COMMAND} ${CMAKE_SOURCE_DIR}
```
psychon: This looks wrong, too (actually, the version before this patch already looks weird). The `check-examples` command just reruns CMake?!?

blueyed: Yes, it is meant to bypass the caching, which is removed here, but I've thought to keep it until it stabilized at least.
@Elv13 needed it the other day.

CMakeLists.txt
blueyed: It is not really needed currently, but it is a dependency.
I plan to e.g. not enable/run ldoc by default, so this describes the depency in this case (when not everything is added to ALL).
Will move it to line 300 though.

tests/examples/CMakeLists.txt
blueyed: `EXAMPLE_DOC_GENERATED_FILES` is still used, but not cached etc.
`generate-examples` can be used to generate them, and via the `add_custom_command` that generates the output (and its OUTPUT/DEPENDS) this actually becomes lazy.

blueyed: It is not about reverting the previous changes, but improving/fixing them.

psychon:
Ping? I just came across a case of "the wrong docs were generated" due to this. Basically, add a new, empty line to `lib/awful/placement.lua` and the next `make` run will generate `build/lib/awful/placement.lua` with all the `@FOO@` expanded to the empty string. In other news, this stuff cannot be cached without also caching these text expansions!

blueyed:
Good from my side, waiting for feedback from @Elv13.

This removes the caching, and given other changes it works much better: the example tests are not re-run always.

Elv13:
@psychon Yes, the caching is removed because of the issue you describe.

@blueyed I wont have time until a few more days, but I guess you can go ahead and merge it, that's already better than the current state.

PR #2074, opened 2017-10-22
build-utils/dist.sh: remove "-dirty" suffix

blueyed:
Calling `make dist` on a dirty tree caused an error when trying to use
`git log` with the `-dirty` suffix.

This patch removes any "-dirty" suffix from the argument.

[ci skip]

### DIFF
--- a/build-utils/dist.sh
+++ b/build-utils/dist.sh
@@ -9,7 +9,7 @@ if [ -z "$1" ]; then
     exit 64
 fi
 
-GIT_TAG="$1"
+GIT_TAG="${1%-dirty}"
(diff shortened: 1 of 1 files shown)


psychon:
Wouldn't running `make dist` on a dirty tree cause problems anyway since you are building a tarball which contains changes that are not in git?

blueyed:
It uses `git archive`, which archives the rev, not the tree (AFAIK)?!

psychon:
Oh, you are right. Sorry!

blueyed:
No problem - better to ask!
Good to go then I assume..

Issue #2070, opened 2017-10-22
build: use a fixed date for calendar examples


blueyed:
There are differences in the generated calendar images, apparently because they are using the current date:

https://github.com/awesomeWM/apidoc/blob/d932f153bfbc1d3ae79b65c399dd02b369ef4e13/images/AUTOGEN_wibox_widget_calendar_font.svg
https://github.com/awesomeWM/apidoc/blob/762037462d89763934a4448826d1f6114b28b08b/images/AUTOGEN_wibox_widget_calendar_font.svg

They should use a fixed date instead.

@Elv13 said this should have been fixed, but apparently it is not.

Via https://github.com/awesomeWM/apidoc/commit/d932f153bfbc1d3ae79b65c399dd02b369ef4e13#diff-c82e6b5d90c6a03924f68894e30f21ee

psychon:
The difficult part of this would be: What date should be picked? Something far in the past? Something far in the future? Anything that is not obviously fake would "feel" bad to readers, I think.

For the actual patch, I would suggest:
```diff
diff --git a/.travis.yml b/.travis.yml
index 47edc3c1f..d69270e04 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,13 @@ env:
   matrix:
     - LUA=5.2 LUANAME=lua5.2 DO_COVERAGE=coveralls
     # Lua 5.3 isn't available in Ubuntu Trusty, so some magic below installs it.
-    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov EXPORT_SDE=true
+    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov
     # luajit: installed from source.
     - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.5 TEST_PREV_COMMITS=1
     # Note: luarocks does not work with Lua 5.0.
     - LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build
     # Lua 5.2 with older lgi and screen size not divisible by 2.
[...]

PR #2075, opened 2017-10-22
Travis: use stable SOURCE_DATE_EPOCH for predictable calendar examples

blueyed:
Fixes https://github.com/awesomeWM/awesome/issues/2070

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,14 @@ env:
   matrix:
     - LUA=5.2 LUANAME=lua5.2 DO_COVERAGE=coveralls
     # Lua 5.3 isn't available in Ubuntu Trusty, so some magic below installs it.
-    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov EXPORT_SDE=true
+    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov
     # luajit: installed from source.
     - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.5 TEST_PREV_COMMITS=1
     # Note: luarocks does not work with Lua 5.0.
     - LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build
     # Lua 5.2 with older lgi and screen size not divisible by 2.
-    - LUA=5.2 LUANAME=lua5.2 LGIVER=0.8.0 TESTS_SCREEN_SIZE=1921x1079 BUILD_APIDOC=true DO_CHECKQA=1 EMPTY_THEME_WHILE_LOADING=1
+    # SOURCE_DATE_EPOCH: used for stable dates during build.
+    - LUA=5.2 LUANAME=lua5.2 LGIVER=0.8.0 TESTS_SCREEN_SIZE=1921x1079 BUILD_APIDOC=true DO_CHECKQA=1 EMPTY_THEME_WHILE_LOADING=1 SOURCE_DATE_EPOCH=1893456000
   global:
     # Secure GH_APIDOC_TOKEN to push to awesomeWM/apidoc.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,13 +12,14 @@ env:
...
     # Lua 5.3 isn't available in Ubuntu Trusty, so some magic below installs it.
-    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov EXPORT_SDE=true
+    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov
     # luajit: installed from source.
     - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.5 TEST_PREV_COMMITS=1
     # Note: luarocks does not work with Lua 5.0.
     - LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build
     # Lua 5.2 with older lgi and screen size not divisible by 2.
-    - LUA=5.2 LUANAME=lua5.2 LGIVER=0.8.0 TESTS_SCREEN_SIZE=1921x1079 BUILD_APIDOC=true DO_CHECKQA=1 EMPTY_THEME_WHILE_LOADING=1
+    # SOURCE_DATE_EPOCH: used for stable dates during build.  Should be the unix timestamp of the last release (git log --pretty=format:%cd -n 1 --date=unix v4.2).
```
Added a comment.  Could be added to the how-to-release doc - where is that again?

psychon: `awesome-www/doing_a_release.mdwn`

.travis.yml
```diff
@@ -12,13 +12,13 @@ env:
...
     # Lua 5.3 isn't available in Ubuntu Trusty, so some magic below installs it.
-    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov EXPORT_SDE=true
+    - LUA=5.3 LUANAME=lua5.3 LUALIBRARY=/usr/lib/liblua.so DO_COVERAGE=codecov
     # luajit: installed from source.
     - LUA=5.1 LUANAME=luajit-2.0 LUALIBRARY=/usr/lib/libluajit-5.1.so LUAROCKS_ARGS=--lua-suffix=jit-2.0.5 TEST_PREV_COMMITS=1
     # Note: luarocks does not work with Lua 5.0.
     - LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build
     # Lua 5.2 with older lgi and screen size not divisible by 2.
-    - LUA=5.2 LUANAME=lua5.2 LGIVER=0.8.0 TESTS_SCREEN_SIZE=1921x1079 BUILD_APIDOC=true DO_CHECKQA=1 EMPTY_THEME_WHILE_LOADING=1
+    - LUA=5.2 LUANAME=lua5.2 LGIVER=0.8.0 TESTS_SCREEN_SIZE=1921x1079 BUILD_APIDOC=true DO_CHECKQA=1 EMPTY_THEME_WHILE_LOADING=1 SOURCE_DATE_EPOCH=1502906828
```
psychon: What's so special about 16th of August? Could we take a date from before 2000 or after 2100 instead?

.travis.yml
psychon: But for releases it would just take the date of the release, I think. Shouldn't that be enough?

.travis.yml
blueyed: It was from the NEWS file.
Amended to use the last tag (`4.2`).

.travis.yml
blueyed: Yes, but it is fixed (I did not want to have a magic `git-log` that would look for the newest tag and use that.
So we should update the .travis.yml after a release - although it is not really important IMHO.
Would you prefer this to be automatic, i.e. using git-log magic?

psychon: I mean: For releases, the doc is built by Travis on a build of the awesome-www repo. The date we use here is only used for the builds of the current devel version docs.

I suggest to use 1893456000 here. That's the output of `date -u  +%s -d '2030-01-01'`. That way, we can just leave it at that and do not have to update it.

blueyed: Ok, amended.

PR #2079, opened 2017-10-22
[RFC] awesomeConfig.cmake: GENERATE_DOC/GENERATE_MANPAGES: default to OFF

blueyed:
Generating documentation is slow, since it runs/configures all the
example tests, and the typical user does not care about generating the
docs when building.

Generating man pages is also rather slow, and is typically not needed.
For CMAKE_BUILD_TYPE=RELEASE generating man pages is still enabled by default.

This adds fallback `ldoc` and `man` targets to explain how to enable it,
with wrapping targets in our main Makefile.

TODO:
- [ ] mention in docs how distributions should build it to include man pages, i.e. that they should use cmake with `-D CMAKE_BUILD_TYPE=RELEASE`, or `make CMAKE_ARGS=-DCMAKE_BUILD_TYPE=RELEASE`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,6 +263,10 @@ if(GENERATE_MANPAGES)
     endforeach()
 
     add_custom_target(man ALL DEPENDS ${MAN_FILES})
+else()
+    add_custom_target(man false
+        COMMENT "man target not generated. Use GENERATE_MANPAGES=1 to enable it, e.g. via `cmake -D GENERATE_MANPAGES=1 build/` or `make distclean man`."
+        VERBATIM)
 endif()
 # }}}
 
@@ -302,6 +306,10 @@ if(GENERATE_DOC)
         WORKING_DIRECTORY ${AWE_DOC_DIR}
         DEPENDS ${AWE_SRCS} ${AWE_LUA_FILES} ${AWE_MD_FILES} ${BUILD_DIR}/docs/config.ld
         COMMENT "Generating API documentation${ldoc_desc_suffix}")
+else()
+    add_custom_target(ldoc false
+        COMMENT "ldoc target not generated. Use GENERATE_DOC=1 to enable it, e.g. via `cmake -D GENERATE_DOC=1 build/` or `make distclean ldoc`."
+        VERBATIM)
 endif()
 # }}}
 
--- a/Makefile
+++ b/Makefile
@@ -34,14 +34,6 @@ distclean:
 	$(ECHO) -n "Cleaning up build directory…"
 	$(RM) -r $(BUILDDIR)
 
-# Wrapper targets to enable building man pages / doc.
-# Should do the same as the implicit target at the end.
-man: CMAKE_ARGS+=-D GENERATE_MANPAGES=1
-ldoc: CMAKE_ARGS+=-D GENERATE_DOC=1
-ldoc man: $(BUILDDIR)/Makefile
-	$(ECHO) "Running make $@ in $(BUILDDIR)…"
-	$(MAKE) -C $(BUILDDIR) $@
-
 # Use an explicit rule to not "update" the Makefile via the implicit rule below.
 Makefile: ;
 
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -302,6 +306,10 @@ if(GENERATE_DOC)
         WORKING_DIRECTORY ${AWE_DOC_DIR}
         DEPENDS ${AWE_SRCS} ${AWE_LUA_FILES} ${AWE_MD_FILES} ${BUILD_DIR}/docs/config.ld
         COMMENT "Generating API documentation${ldoc_desc_suffix}")
+else()
+    add_custom_target(ldoc false
+        COMMENT "ldoc target not generated. Use GENERATE_DOC=1 to enable it, e.g. via `cmake -D GENERATE_DOC=1 build/` or `make distclean ldoc`."
+        VERBATIM)
```
psychon: I'd still add these two targets to `ALL` (and then have them not just always fail; perhaps just run true?. Since the `man` and `ldoc` targets are included in `ALL`, I guess no one runs them directly anyway.

Makefile
```diff
@@ -35,6 +35,14 @@ distclean:
 	$(RM) -r $(BUILDDIR) $(TARGETS)
 	$(ECHO) " done"
 
+# Wrapper targets to enable building man pages / doc.
+# Should do the same as the implicit target at the end.
+man: CMAKE_ARGS+=-D GENERATE_MANPAGES=1
+ldoc: CMAKE_ARGS+=-D GENERATE_DOC=1
+ldoc man: $(BUILDDIR)/Makefile
+	$(ECHO) "Running make $@ in $(BUILDDIR)…"
+	$(MAKE) -C $(BUILDDIR) $@
```
psychon: `man` and `ldoc` should be added to `.PHONY` at the end of the file.

Also, does this actually work in a non-clean build? CMake was already run, so it will not be rerun and the new `CMAKE_ARGS` are just ignored.

CMakeLists.txt
blueyed: Well, I use `make ldoc` when working on the docs.
This should not silently (although with a comment) pass, without actually building the docs then.

blueyed:
> I'd say "the typical user also does not care much about build time". He'll just wait a little longer for awesome to build, and that's it.

Well, if they can decide they likely want a faster build.
This is also very much relevant when using it through e.g. Arch's AUR.

> This will more likely affect those contribution to awesome, I guess.

Even more so - and that's why I am trying to fix it, since it affects myself.. :)

Makefile
blueyed: You are right.. We could have a `cmake-configure` target for them, but then again this adds some overhead, so I will just remove the helpers: the note from the failing target should be clear enough  then.

psychon:
> Even more so - and that's why I am trying to fix it, since it affects myself.. :)

So what's your workflow? You have a build directory with docs enabled and another one without and then always use the one that you currently need? Or do you go just wait for the additional time that CMake needs to re-run?

Myself, I'm using CMake's ninja backend and I felt like the resulting parallel builds hide most of the additional time needed quite well. I only sometimes notice the time that LDoc needs.

blueyed:
> So what's your workflow? You have a build directory with docs enabled and another one without and then always use the one that you currently need?

No, only a single build dir typically.

> Or do you go just wait for the additional time that CMake needs to re-run?

Yes, when working on docs I would just clean and rebuild it then (if docs where not enabled before).

I am using `ccache` with a tmpfs, so awesome is really fast to build after having it done once (I reboot not often, but tend to suspend only).

blueyed:
> I felt like the resulting parallel builds hide most of the additional time needed quite well

Unfortunately not running the example tests nor build the docs itself unfortunately.
`ldoc` itself is not that slow, but previously it would also re-run all the example tests (that should have been improved by now, i.e. only new / changed images should get rebuild).

Elv13:
Does #2322 makes this obsolete?

blueyed:
@Elv13 
Likely, at least it reminded me of this / my work in this regard.
Have not checked both in detail though.

actionless:
@blueyed, so can you confirm what #2322 solves this issue?

blueyed:
Assuming so.

PR #2093, opened 2017-10-30
Add call_now argument to gears.timer

necauqua:
This is a very simple patch which adds `immediate_start` boolean argument to timer creation.
This argument works as autostart, but also calls the callback at 'zero time' - at timer creation.
It is useful for stuff like watch widgets, i have below construct repeated like 4 times across my code:
```lua
local function update() ... end
update() -- first (or zeroeth) time call here!
gears.timer { timeout = 5, autostart = true, callback = update }
```
And so i couldn't inline the callback in timer creation, which would have looked like this
```lua
gears.timer {
    timeout = 5, immediate_start = true,
    callback = function() ... end
}
```
Single statement timer looks more consice for me and there is no 'N/A' (or not any) text on widgets for `timeout` seconds.

### DIFF
--- a/lib/gears/timer.lua
+++ b/lib/gears/timer.lua
@@ -30,6 +30,7 @@
 --
 --    gears.timer {
 --        timeout   = 10,
+--        call_now  = true,
 --        autostart = true,
 --        callback  = function()
 --            -- You should read it from `/sys/class/power_supply/` (on Linux)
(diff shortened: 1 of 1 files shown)


necauqua:
Yes, at first i wanted it to be separate, but then realized that nobody wants `call_now = true, autostart = false`, because if start will not happen immediately then time gap between first call(from call_now) and next(from timer finally being started from somewhere) is not equal to `timeout` so this is inconsistent.
And so `call_now = true, autostart = true` is redundant and i came up with `immediate_start`.

blueyed:
I agree with @Elv13.
Two distinct properties seem better.

necauqua:
Changed it to `call_now` with appropriate behavior

Elv13:
Thanks!

Issue #2087, opened 2017-10-26
Rule 'tag' with tag name errors if that tag is not on same screen with client


necauqua:
Output of `awesome --version`:

awesome v4.2-142-g57b18df5 (Human after all)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.2

**How to reproduce the issue:**

On two screens have tags 'a' and 'b' respectively.
Move some client to tag 'b'.
Add rule for client to land on tag 'a' (string rule, so `tag = 'a'`).
Reload awesome.

**Actual result:**

Client will still sit on tag 'b' and this line is notified with `debug::error` signal:
```/usr/share/awesome/lib/awful/rules.lua:299: attempt to index a nil value (local 'value')```

This is because is could not find any tag named 'a' on second screen.

**Expected result:**
Either it should search for 'a' tag on all screens and move the client to it, or if such tag is not found on current screen it should ignore it.
Anyway, it should not produce an error, or that error has to be more specific(like `tag 'a' not found on screen #1`)

**Comment by psychon** [comment]
> Either it should search for 'a' tag on all screens and move the client to it,

This still needs the patch below in case there really is no such tag. Plus, ontop of that:
```diff
diff --git a/lib/awful/rules.lua b/lib/awful/rules.lua
index aefd8fbc..0b0bdfa7 100644
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -292,6 +292,7 @@ local force_ignore = {
 function rules.high_priority_properties.tag(c, value, props)
     if value and type(value) == "string" then
         value = atag.find_by_name(c.screen, value)
+             or atag.find_by_name(nil, value)
     end
 
     if value then
```

> or if such tag is not found on current screen it should ignore it.

```diff
diff --git a/lib/awful/rules.lua b/lib/awful/rules.lua
index 10a7bb86..aefd8fbc 100644
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -290,11 +290,11 @@ local force_ignore = {
 }
 
 function rules.high_priority_properties.tag(c, value, props)
-    if value then
-        if type(value) == "string" then
-            value = atag.find_by_name(c.screen, value)
-        end
+    if value and type(value) == "string" then
+        value = atag.find_by_name(c.screen, value)
+    end
 
+    if value then
[...]

PR #2094, opened 2017-10-31
awful.rules: Handle non-existing tags

psychon:
If a tag is specified by name, but no such tags exist, awful.rules would
cause an error (attempt to index a nil value). Fix this and add a test
for this case.

Fixes: https://github.com/awesomeWM/awesome/issues/2087

-----
Note that this only fixes the error in #2087. I am not sure if we want to do "fallback to `awful.tag.find_by_name(nil, value)`". @Elv13 ?


Elv13:
Thanks

PR #2099, opened 2017-11-02
Fix doc: get_clients methods belong to screen, not to client

stloewen:
The three `get_[|all_|tiled_]clients()` methods are incorrectly attributed to `client` although they belong to `screen`.


Elv13:
thanks

PR #2104, opened 2017-11-05
Fix the resize/move "after" mode

Elv13:


tests/test-resize.lua
```diff
@@ -113,19 +113,70 @@ table.insert(steps, function()
...
+
+    -- local c = client.get()[1]
 
     --     if not mousegrabber.isrunning then --FIXME it should work, but doesn't
     --         return true
     --     end
 
+    --FIXME, the mousegrabber callback says the mouse buttons are not pressed,
+    -- theirfor the test is broken
+    -- Nothing should have changed until button_release is done
```
psychon: If you want me to guess: This is because of the `button_release` done above. Add an `awesome.sync()` after it and move the setup of the next test into a separate step. That should make sure that that button release is "out of the way".

(And 👍 for the first commit)

PR #2107, opened 2017-11-08
Update menubar generator to acknowledge application translations

kitsunenokenja:
This patch revises the generation of the program entry list by
considering the current locale and matching against the program
name's translation, if there is one, for the designated locale.

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -261,6 +261,7 @@ end
 function utils.parse_desktop_file(file)
     local program = { show = true, file = file }
     local desktop_entry = false
+    local locale = string.sub(os.setlocale(nil), 1, 2) or "en"
 
     -- Parse the .desktop file.
     -- We are interested in [Desktop Entry] group only.
@@ -278,9 +279,12 @@ function utils.parse_desktop_file(file)
             end
 
             -- Grab the values
-            for key, value in line:gmatch("(%w+)%s*=%s*(.+)") do
+            for key, value in line:gmatch("([%w%[%]]+)%s*=%s*(.+)") do
                 if list_keys[key] then
                     program[key] = utils.parse_list(value)
+                elseif key == "Name[" .. locale .. "]" then
+                    -- Overwrite Name with a translation of the current locale.
+                    program["Name"] = utils.unescape(value)
                 else
                     program[key] = utils.unescape(value)
                 end


Elv13:
If you could update `.travil.yml`, set the locale on one of the runner and add a `.desktop` in `tests/data/desktop/` (need to be created, we have another PR with that, but it isn't progressing very much) then read it from `test-menubar.lua` (by checking if the right locale is set first), that would be appreciated.

Note that `test-menubar.lua` is currently not executed in all runner due to a crash.

lib/menubar/utils.lua
```diff
@@ -261,6 +261,7 @@ end
 function utils.parse_desktop_file(file)
     local program = { show = true, file = file }
     local desktop_entry = false
+    local locale = string.sub(os.setlocale(nil), 1, 2) or "en"
```
psychon: I think we should follow the spec here:
https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s04.html
https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s05.html

This means that we cannot use lua to query the locale since that doesn't allow to query `LC_MESSAGES`, it only has the following: "all", "collate", "ctype", "monetary", "numeric", or "time".
Also, the matching rules are more complicated than what you use here and I'm not quite sure if `lang` really is limited to two characters.

Plus, this does not only apply to `Name`, but to
- Name
- GenericName
- Comment
- Icon
- Keywords

Currently we only use Name and Icon, but I think the code should still read desktop files correctly.

Given the above complications, why do we parse `desktop` files ourselves anyway? I think it's time to switch to https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html which has `g_key_file_get_locale_string` that Does The Right Thing (tm) and is documented as implementing the Desktop Entry Specification.

What do others think about this?

Elv13: > What do others think about this?

I guess I made my opinion rather explicit about this topic in other issues: yes. And also use the icon loader if gtk is available, but that's not related to this issue.

kitsunenokenja: Using the proposed established library routine to properly extract information from the `.desktop` file will make code maintenance for this project simpler and potentially provide benefit in the future, should the sped for that file format change, and a corresponding adjustment to the library doesn't require changes to its consumers.

There's no point to proceeding with this patch if it's fated to be removed entirely anyway.

psychon:
Superseded by #2111 which (tries to) use(s) an existing implementation for parsing `desktop` files instead of improving and enlarging our own parser even more.

Thanks for bringing this up and sorry that your patch didn't get merged.

kitsunenokenja:
Glad to be of assistance. No need to apologise; the code change that did go in is a much better improvement.

Issue #2091, opened 2017-10-30
What's the high-DPI plan for awesome?


psychon:
I just had a quick discussion on IRC with @Elv13 about what the plan for DPI support is. As I see it, there are the following levels of DPI support (with some possibly refinement):

(a) Don't support DPI. Everything is specified in pixels and people will have to tweak sizes when their monitor is high-DPI.
(b) Only use a single DPI value for fonts. Everything except fonts is specified in pixels, fonts scale with a single, global DPI value.
(c) Only use a single DPI value for everything. Like the above, but also scale all lengths.
(d) Use per-screen DPI values for fonts. Everything except fonts is specified in pixels.
(e) Use per-screen DPI values for everything. Like the above, but also scale everything.

I am not quite sure what the current state in awesome is. We do (d) at least, but with some bugs (e.g. titlebar height is computed based on global DPI, but drawn based on per-screen DPI). We do a bit of (c) by having some random calls to `beautiful.xresources.apply_dpi` in some places, but not all of them.

We have #2053 and I think it wants to move us towards (e), but I am not quite sure.
@Elv13 correctly said that doing (e) is quite complicated since all widgets will now have to be made DPI-aware and it is relatively easy to forget just some numbers here. Testing this and making sure it works correctly (and does not regress over time) is hard.

So, @Oblomov what is your plan? How much of it are you wiling to implement yourself?

If we decide that awesome supports only a global DPI, I think we should remove per-screen-DPI since it does not work correctly anyway. If we decide that awesome supports per screen DPI, then it makes no sense to also have a global DPI value.

Another (perhaps unrelated to this issue) thing from #2053 is the concept of a separate font DPI: Historically, people who wanted larger fonts just increased their DPI value since it was only used for fonts anyway (not just in awesome). If we now do (c) or (e), then "making fonts larger" needs a separate knob and #2053 does this by introducing a per-screen font DPI which is distinct from the screen's DPI itself. This is calculated based on `Xft.dpi` to make fonts larger/smaller. Do we want that?

I think we should decide which of these options we actually want and then actually do that, instead of doing all the options badly. The current state of things is not really good (thus, marking this with the `v4.3` milestone).

CC @Elv13 @blueyed @actionless @Oblomov

Elv13:
[...]

PR #2109, opened 2017-11-13
screen: Revert the DPI per screen changes.

Elv13:
The changes to how DPI is calculated breaks existing config and thus
have to be reverted to a saner bahavior. If a theme or code in rc.lua
specify a size, it is in pixel. If the DPI is calculated differently,
then the assumption that the size in pixel corresponds to the user
settings (aka, 96dpi or whatever is in ~/.Xresources) is to be honored
was previously true and is now false. This is a behavior change and
cannot go into a stable branch release.

As suggested by Uli, this introduces a version check to change the
behavior when Awesome 5 is released. This assumes all other changes
to make Awesome auto-magically apply the DPI to all elements are
merged until then (I seriously doubt this will happen).

Compared to the patch draft proposed by @psychon, this patch invert the `get_dpi` logic to use `beautiful.xresources` as the primary source again, restoring the previous DPI choice. 

Ref: #2091

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -504,25 +504,78 @@ function screen.object.get_selected_tag(s)
     return screen.object.get_selected_tags(s)[1]
 end
 
+--- Enable the automatic calculation of the screen DPI (experimental).
+--
+-- This will cause many elements such as the font and some widgets to be scaled
+-- so they look the same (physical) size on different devices with different
+-- pixel density.
+--
+-- It is calculated using the information provided from `xrandr`.
+--
+-- When enabled, the theme and configuration must avoid using pixel sizes for
+-- different elements as this will cause misalignment or hidden content on some
+-- devices.
+--
+-- Note that it has to be called early in `rc.lua` and requires restarting
+-- awesome to take effect. It is disabled by default and changes introduced in
+-- minor releases of Awesome may slightly break the behavior as more components
+-- gain support for HiDPI.
+--
+-- When disabled the DPI is acquired from the `Xft.dpi` X resource (xrdb),
+-- defaulting to 96.
+--
+-- @tparam boolean enabled Enable or disable automatic DPI support.
+function screen.set_auto_dpi_enabled(enabled)
+    for s in capi.screen do
+        s.data.dpi_cache = nil
+    end
+    data.autodpi = enabled
+end
+
+
 --- The number of pixels per inch of the screen.
 -- @property dpi
 -- @treturn number the DPI value.
 
 local xft_dpi, fallback_dpi
 
+local function get_fallback()
+    local mm_per_inch = 25.4
+
+    -- Following Keith Packard's whitepaper on Xft,
+    -- https://keithp.com/~keithp/talks/xtc2001/paper/xft.html#sec-editing
+    -- the proper fallback for Xft.dpi is the vertical DPI reported by
+    -- the X server. This will generally be 96 on Xorg, unless the user
+    -- has configured it differently
+    if root and not fallback_dpi then
+        local _, h = root.size()
+        local _, hmm = root.size_mm()
+        fallback_dpi = hmm ~= 0 and h * mm_per_inch / hmm
+    end
+
+    return fallback_dpi or 96
+end
+
 function screen.object.get_dpi(s)
     local mm_per_inch = 25.4
 
-    if s.data.dpi then
-        return s.data.dpi
+    if s.data.dpi or s.data.dpi_cache then
+        return s.data.dpi or s.data.dpi_cache
     end
 
     -- Xft.dpi is explicit user configuration, so honor it
     if not xft_dpi and awesome and awesome.xrdb_get_value then
         xft_dpi = tonumber(awesome.xrdb_get_value("", "Xft.dpi")) or false
     end
+
     if xft_dpi then
-        return xft_dpi
+        s.data.dpi_cache = xft_dpi
+        return s.data.dpi_cache
+    end
+
+    if not data.autodpi then
+        s.data.dpi_cache = get_fallback()
+        return s.data.dpi_cache
     end
 
     -- Try to compute DPI based on outputs (use the minimum)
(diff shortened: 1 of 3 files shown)


Elv13:
Note that in itself this also breaks the behavior since there was an "hardcoded" DPI per screen system before the "automatic" one. So I guess the final patch will have to be something in between fully reverting the recent DPI per screen work or adding more "if" to make sure it doesn't try to invent its own DPI.

lib/wibox/widget/textbox.lua
```diff
@@ -107,6 +108,8 @@ end
 -- @treturn number The preferred width.
 -- @treturn number The preferred height.
 function textbox:get_preferred_size_at_dpi(dpi)
+    assert(dpi, "No DPI provided")
```
psychon: Huh? Why these new asserts? Does this just provide better error messages than what happened before? If so, should this assert perhaps be moved to `setup_dpi`?

psychon:
How about we just start an v4 branch (any suggestions for good names?), revert the commits in question there and are done? That seems saner than trying to add lots of version checks...
Backward-compatible development should target the v4 branch, breaking changes go to master and from time to time v4 is merged into master.

Elv13:
I am against that because it failed twice already:

* 3.4-3.5 took 3 years, 3.5 to 4 took 4 years. I don't see any clue that
5.0 would make the cut faster. Even if it did, it would be too soon to
break the API, the userbase would revolt, again.
* During the dual branch time (3.4.x/3.5.x), no major contributors used the
stable branch, letting it decay and bugs stay unfixed. Nobody backported
[most] compatible changes during the 3.5 era (and worst, during 3.4, it had
a separate dev base and diverged). Worst still, the changes that were
backported ended up mostly untested before releases were made because
nobody used the "git" HEAD of the stable branch. We don't have the manpower
to maintain branches. You already spend a lot of time on Awesome, Daniel
and me are very active for some time, but only once in a while, not
continuously.
* We lost 80% of our users due to, among other things, the lack of
perceived progress and the big+bad API breaks once it happened. Letting the
branch diverge will just recreate that exact same situation again. Keeping
the same branch with me complaining about any changes prevents the slippery
slope from starting again. I sucks for me too, I like changes. However I
prefer Awesome to thrive by leveraging stability and keeping the code
developed by our users compatible for an extended time frame. Yes, it does
induce minor regressions from time to time, but given we are able to
release more often than 2-3 time per decade, it's a cost I consider
acceptable.
* We are still stuck getting questions about 3.4.x (8.4 years old) due to
the inability to release often. If we branch v5 and it takes 5 years to
release, then we will have to support 4.x into the 2030's. Having a clean
"release from master" + deprecation system forces us to keep rolling. Even
if it adds many `--TODO v5` and `gears.debug.deprecate` in the code.
That's, again, a cost I consider acceptable compared to forking.
* There is often very cool stuff being developed, it's disrespectful not to
let our users have them within 4 months or so. Having a single branch
enforces that new compatible features get released.
* I am profoundly skeptical of the capacity to fully support multi-DPI.
This kind of bug is the tip of the iceberg. Any changes will only make
things worst. X11 doesn't support it for apps and it can be done using the
config already. It is still unclear to me why it could not be developed
using some kind of smart zooming at the widget scale. If that happens to be
possible, then it should be done first. And if it's done, then maybe that's
enough. Letting the little minority of users who really need this implement
it in their config themselves using the tools we provide instead of making
a gigantic number of changes affecting everybody and breaking the
API/behavior. It will turn out like a dog chasing its tail or the video
game "uncanny valley" syndrome. IMO, while having per screen DPI is cool,
it's not worth all the troubles. Having the global DPI + font scaling is a
good compromise because it keeps Awesome usable for HiDPI laptop users and
moves the burden away from us.

As for the patch here, I will do it another way. As I mentioned in the
comment when I saw luacheck failed, it indeed could not work this way. I
think I will add an `awful.screen.set_multi_dpi_enabled(bool)` (default:
false) and move the remaining part of `beautiful.xresources.get_dp(s)` into
`awful/screen.lua` to avoid the dependency loop.

Elv13:
I improved the issues in this new version

lib/wibox/widget/textbox.lua
psychon: Ping

lib/awful/screen.lua
```diff
@@ -537,16 +590,12 @@ function screen.object.get_dpi(s)
...
-    -- We have no outputs, so guess based on the size of the root window.
-    if root and not fallback_dpi then
-        local _, h = root.size()
-        local _, hmm = root.size_mm()
-        fallback_dpi = hmm ~= 0 and h * mm_per_inch / hmm
-    end
-    return fallback_dpi or 96
+    s.data.dpi_cache = get_fallback()
+    return s.data.dpi_cache
 end
```
psychon: :+1: for the second commit

lib/wibox/widget/textbox.lua
Elv13: done

actionless:
sorry, i've noticed the notification only now, will have a look now

blueyed:
@Elv13 
How do you feel about covering this with tests? :)

Elv13:
> How do you feel about covering this with tests? :)

As stated in the related issues, one of my big issue with the DPI project is how hard it is to properly test. At some point we will need a new test suit that compare each widgets at different DPI to see if the images match (as @actionless proposes). We will also have to port every single unit of measurement currently mapping to pixels (such as wibox/client/widget geometry, margins, paddings, spacings) to an abstract units. Then implement either the proposed auto-widget-scaling or the less maintainable per-widget adjustments. Then (and only then) can we start testing DPI seriously. For now testing it would only help notice how broken it is. Once all the testing is done, then maybe this could be enabled by default.

This change is a compromise. I revert the DPI per screen changes because they were (currently) untestable and started introducing regressions as soon as they got merged. On the other hand, if an user want to test the per-screen DPI support, it can still be enabled. But then, that user has to deal with the pixel vs. dpi issues him/herself. I want to state it clearly, I am *not* against DPI scaling. But I am against user visible regressions and see this as a prime example of unstable changes with little benefits.

actionless:
@blueyed mb if remove fake ~/.Xresources from the testsuite `get_fallback()` function and related code branches will became covered?

blueyed:
@actionless 
Yes, https://github.com/awesomeWM/awesome/pull/2121.

@Elv13 
I agree with your comment, but think that having it covered already a bit will help with developing this further.  #2121 hopefully helps here.

PR #2111, opened 2017-11-15
Menubar: Use GlibKeyFile

psychon:
This supersedes #2107. Testing done: Basically none. Menubar can still open, (is still slow to open) and still finds Firefox.

@Elv13 This now returns booleans as actual booleans. API break or not?

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -7,7 +7,6 @@
 ---------------------------------------------------------------------------
 
 -- Grab environment
-local io = io
 local table = table
 local ipairs = ipairs
 local string = string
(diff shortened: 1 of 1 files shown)


psychon:
@kitsunenokenja Could you test this? You seem to have a `.desktop` file with a locale-dependant `Name` entry laying around.

kitsunenokenja:
I'll return with confirmation after having a chance to test the change.

Elv13:
> @Elv13 This now returns booleans as actual booleans. API break or not?

It's fine. I don't think anyone will notice. While there is some known users of these APIs (and thankfully, for once it's not me), they are not widespread enough and could almost be considered internal given how unusable they are to create more serious menus.

Plus, this gets rid of real bugs, so it's welcome. Maybe test is one of @lcpz config explode because of this. Otherwise I am 95% certain nobody else relies on that specific code.

lcpz:
> Maybe test is one of @lcpz config explode because of this. 

The only thing potentially affected is [`freedesktop.menu`](https://github.com/lcpz/awesome-freedesktop/blob/master/menu.lua), which is a < 100 LOC exploit of `menubar.utils`, therefore easily fixable if necessary.

kitsunenokenja:
Odd that the icons for the categories do not appear, but translations definitely load as expected.

psychon:
Thanks for testing!

Issue #522, opened 2015-10-11
menubar: no category icons with Adwaita theme


blueyed:
The Adwaita theme gets used by default, but provides no icons for the categories, e.g. applications-accessories.

These would be provided by "gnome".

Maybe fallback themes should be used in this case?

Should the lookup for categories only look into `categories` / `Context=Categories` subdirs?

/cc @nuko8

blueyed:
btw: I am heavily trying to optimize the menubar/theme handling code, but there are e.g. still ~9300 calls to `file_readable` just for handling the category icons, that cannot be found..!

Elv13:
There is already a solution for that by using the GTK implementation, it is asynchronious and very fast. It work without crash for me (compared to some other LGI async API), probably because it never perform recursive async operations. I am also currently porting that whole code to C (with some C++ features mixed in) and plan to use dlopen to avoid adding a dependency on GTK at install time (but use it if it's installed). Here is the (very early, and mostly written today) code https://gist.github.com/Elv13/d5d2cf7d2fa23d75a35c (original https://github.com/Elv13/awesome-configs/blob/master/utils/fd_async.lua ) . It doesn't yet plug with lua or have the request object logic. This code is also planned to be a private API used by a lua frontend.

So, if you want my opinion on this, rather than implement the proper XDG icon spec and fallback themes and so on, lets just use a solution that's already implemented and much faster than lua lookups anyway. I still need to port the socket, download, dbus, icons and INI parsing code to "big-C" and write the lua side, but it is progressing well. The code is also not release quality, it's looking ugly, leak memory as hell and don't have error handlers at all.

blueyed:
> So, if you want my opinion on this, rather than implement the proper XDG icon spec and fallback themes and so on, lets just use a solution that's already implemented and much faster than lua lookups anyway. 

:+1: 

But it will take some time until the icon lookup / theme handling could be used through this, right?

Elv13:
Oh, yes and by the way, I already optimized that code in my 3.4 config
https://github.com/Elv13/awesome-configs/commit/ddb34e83ec0ac6cb3c8555b440208504b101a681#diff-32f0db718b66dfb55ada04bd075a0c77L1

However, I dropped this a year ago when I switched to the much faster GTK implementation. The Awesome `menubar` module is a fork of the previous `freedesktop` module I was using in my config, they diverged long time ago and the maintainer of freedesktop never cared to merge my fixes.

Elv13:
[...]

PR #2112, opened 2017-11-16
Deprecate menubar.icon_theme

psychon:
The code in menubar.icon_theme naively implements the algorithm from the
base dir specification. This is a problem: On this system,
/usr/share/icons/{Adwaita,hicolor}/index.theme list 91, respectively 649
subdirectories. Since we check for three file extensions (png, svg,
xpm), this means that a failing icon lookup for the Adwaita theme checks
for (91+649)*3 = 2220 files (in practice it might be a bit better since
the directories have specific meanings, but still). That's insane.

Since we only use this code for looking up category icons anyway, just
deprecate this mess. Category icons are now looked up in the same way
that icons for individual applications are looked up.

Since menubar.init does not require("menubar.icon_theme"), this means
that menubar.icon_theme is no longer actually loaded. That's bad.

(Hopefully) Fixes: https://github.com/awesomeWM/awesome/issues/1496

-----

My long-term plan would be to "reinvent the wheel" and write some module(s) for `gears` to do icon lookups. In a way that doesn't suck. Somehow...

CC @unode Can you test this PR?

### DIFF
--- a/lib/menubar/icon_theme.lua
+++ b/lib/menubar/icon_theme.lua
@@ -1,5 +1,5 @@
 ---------------------------------------------------------------------------
---- Class module for icon lookup for menubar
+--- (Deprecated) class module for icon lookup for menubar
 --
 -- @author Kazunobu Kuriyama
 -- @copyright 2015 Kazunobu Kuriyama
@@ -78,6 +78,7 @@ local icon_theme = { mt = {} }
 local index_theme_cache = {}
 
 --- Class constructor of `icon_theme`
+-- @deprecated menubar.icon_theme.new
 -- @tparam string icon_theme_name Internal name of icon theme
 -- @tparam table base_directories Paths used for lookup
 -- @treturn table An instance of the class `icon_theme`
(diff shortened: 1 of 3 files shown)


unode:
Looks good here. Doesn't seem to find category icons on my system but I don't see that as a blocker (the old implementation didn't either).
I agree that this solves #1496 . Here it opens in a few miliseconds.

psychon:
@unode Thanks!

psychon:
> However, please merged only once the replacement is merged.

Which replacement? So far we only (really) use the code in `menubar.utils`. That we use all this code just to look up categories is a joke, and as #1496 shows it is a bad joke. So, for the time being, I'd consider `menubar.utils.whatever_the_function_was_called` to be the replacement.

Elv13:
> look up categories is a joke, and as #1496 shows it is a bad joke.

As said many time before, I agree on that one. Funnier still, I fixed that bug on Sep 9 2012 back when I used the ancestor of this code https://github.com/Elv13/awesome-configs/blob/002e6028448f0b74e2fd1ebd6c1ed5d2ee7003d7/extern/freedesktop/utils.lua

blueyed:
@psychon 

> Since menubar.init does not require("menubar.icon_theme"), this means
that menubar.icon_theme is no longer actually loaded. That's bad.

What do you mean?  That it is not covered now anymore?

blueyed:
In memory of https://github.com/awesomeWM/awesome/issues/522#issuecomment-147250932.

psychon:
> What do you mean? That it is not covered now anymore?

Awesome does not load those files anymore. You could write invalid Lua in there and nothing would notice.
(These files were never implicitly loaded by `menubar.init`, only explicit `require("menubar.icon_theme")` would load them)

unode:
Thanks guys, that was Awesome (badum tss)

PR #2113, opened 2017-11-16
Add a (recursive) subdirectory cache

psychon:
This adds a new module to gears to scan some directory recursively and uses this in the menubar. The module itself also has some feature to detect changes to the directory over time, but this is currently unused.

### DIFF
--- a/lib/gears/filesystem/subdirectory_cache.lua
+++ b/lib/gears/filesystem/subdirectory_cache.lua
@@ -1,0 +1,125 @@
+---------------------------------------------------------------------------
+-- Check if some file exists inside a directory.
+--
+-- For performance reasons this module once recursively scans the directory to
+-- collect information about all files and then uses this cache to answer
+-- queries. If the directory is modified, it is rescanned. However,
+-- subdirectories are not checked for modification.
+--
+-- Note that this is not re-entrant in the sense that you may not use instances
+-- of this class from multiple async contexts at once.
+--
+-- @author Uli Schlachter
+-- @copyright 2017 Uli Schlachter
+-- @classmod gears.filesystem.subdirectory_cache
+---------------------------------------------------------------------------
+
+local gio = require("lgi").Gio
+local gdebug = require("gears.debug")
+
+local cache = {}
+
+local function async_get_mtime(file)
+    local info = file:async_query_info("time::modified", gio.FileQueryInfoFlags.NONE)
+    return info and info:get_attribute_uint64("time::modified")
+end
+
+local function get_readable_path(gfile)
+    return gfile:get_path() or gfile:get_uri()
+end
+
+local function scan(self, directory)
+    local query = "standard::name,standard::type"
+    local enum, err = directory:async_enumerate_children(query, gio.FileQueryInfoFlags.NONE)
+    if not enum then
+        gdebug.print_warning(get_readable_path(directory) .. ": " .. tostring(err))
+        return
+    end
+    local per_call = 1000 -- Random value
+    local children = {}
+    while true do
+        local list, enum_err = enum:async_next_files(per_call)
+        if enum_err then
+            gdebug.print_warning(get_readable_path(directory) .. ": " .. tostring(enum_err))
+            return
+        end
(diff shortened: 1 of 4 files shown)


lib/menubar/utils.lua
```diff
@@ -364,53 +364,16 @@ end
...
-            end
-        end
-        enum:async_close()
-    end
-
     gio.Async.start(do_protected_call)(function()
         local result = {}
-        parser(gio.File.new_for_path(dir_path), result)
+        for path in pairs(gfs.subdirectory_cache.async_new(dir_path).known_paths) do
+        local success, program = pcall(utils.parse_desktop_file, path)
```
Veratil: Needs one more tab.

lib/gears/filesystem/subdirectory_cache.lua
```diff
@@ -0,0 +1,125 @@
...
+local function scan(self, directory)
+    local query = "standard::name,standard::type"
+    local enum, err = directory:async_enumerate_children(query, gio.FileQueryInfoFlags.NONE)
+    if not enum then
+        gdebug.print_warning(get_readable_path(directory) .. ": " .. tostring(err))
+        return
+    end
+    local per_call = 1000 -- Random value
+    local children = {}
+    while true do
```
Veratil: Should we include a break condition or is hardcoded infinite loop okay here?

lib/gears/filesystem/subdirectory_cache.lua
```diff
@@ -0,0 +1,125 @@
...
+        timeout = timeout or 5,
+        last_check = 0,
+        directory_mtime = 0,
+        known_paths = {},
+    }, { __index = cache })
+    result:async_update()
+    return result
+end
+
+--- Contains all known subfiles
```
Elv13: Need a `.` at the end

lib/gears/filesystem/subdirectory_cache.lua
```diff
@@ -0,0 +1,125 @@
...
+    end
+    self:async_update()
+    return self.known_paths[sane_path] or false
+end
+
+--- Asynchronously creates a new cache object.
+-- @tparam string directory The directory to check.
+-- @tparam[opt=5] int timeout Number of seconds that cache might be out of date.
+-- @return A new cache object.
+function cache.async_new(directory, timeout)
```
Elv13: May I suggest `@function gears.subdirectory_cache(directory, args)` and keep tracks (using weak tables) of the existing caches internally? It could return existing ones or create a new one. If, for example, a module implements different types of applications menu, it would suck for 4-5 cache to exists for the same directory.

There `args` include the timeout and (eventually) `max_depth`? We had issues in the past with these secondary arguments accumulating over time, let's avoid it.

lib/gears/filesystem/subdirectory_cache.lua
```diff
@@ -0,0 +1,125 @@
...
+    local result = setmetatable({
+        directory = gio.File.new_for_path(directory),
+        timeout = timeout or 5,
+        last_check = 0,
+        directory_mtime = 0,
+        known_paths = {},
```
Elv13: Please add the missing `@property` and signals for these attributes.

Being able to do `mycache:connect_signal("property::known_paths", function() end)` seems to be an important use case. It's what I use on my version of this code. My menu is shown in multiple steps. The categories widgets are first, then their icons are loaded. Once that's done, the async loading of the `.desktop` begin. Then once parsed, the submenu are allowed to be created. Icons are loaded asynchronously while the menu widget is being created for the category. I also have support custom `.desktop` directories and use that to manage my `dock` widget. I would make use of such cache for this.

lib/gears/filesystem/subdirectory_cache.lua
```diff
@@ -0,0 +1,125 @@
+---------------------------------------------------------------------------
+-- Check if some file exists inside a directory.
+--
```
Elv13: If it's a file (and subdir) cache, then I think calling this `directory_cache` would make more sense, no?

PR #2116, opened 2017-11-19
Cleanup doc

Elv13:
Question: Is it necessary to create a redirection system so the old links don't break?

### DIFF
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -116,6 +116,17 @@ file = {
 
         -- Ignore some parts of the widget library
         '../lib/awful/widget/init.lua',
+
+        -- Deprecated classes for one years or more don't deserve entries
+        -- in the index
+        '../lib/awful/widget/graph.lua',
+        '../lib/awful/widget/progressbar.lua',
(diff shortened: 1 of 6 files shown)


docs/config.ld
```diff
@@ -109,9 +116,6 @@ file = {
 
         -- Ignore some parts of the widget library
         '../lib/awful/widget/init.lua',
-
-        -- Ignore the drawing window, users should only care about wiboxes
-        '../objects/drawin.c',
```
psychon: Why? What was wrong with this? Also, I'd preserve the comment. You could e.g. do
```
-- Ignore the ....
-- '../objects/drawin.c',
```

Elv13: I stopped using a wildcard, so it's no longer necessary

psychon: Okay, but I'd still like to have this more explicit. Why do we have these docs in the file at all if we don't want them?

Elv13: image

Because ldoc produces those dummy entries if the wildcard is used.

Issue #2118, opened 2017-11-20
wibox.widget.textclock + %W Week Number not working


adebruyn-abit:
Hi Everybody,

Since 4.0, mytextclock is not working anymore. The week number of the year seems to be broken.

With mytextclock = wibox.widget.textclock(" %a %d %b (W%W/D%j), %H:%M:%S ")  nothing is displayed.
If I remove the %W, the textclock is displayed.

Thanks for the feedback.

psychon:
This comes from the fix for #249. Previously, we used Lua's `os.date` which is documented to use the C function `strftime`. The fix for #249 switched to `g_date_time_format`. The [documentation for this function](https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format) states:

> The format strings understood by this function are a subset of the strftime() format language as specified by C99. The %D, %U and %W conversions are not supported, nor is the 'E' modifier. The GNU extensions %k, %l, %s and %P are supported, however, as are the '0', '_' and '-' modifiers.

I don't know why they do not support `%W`, but this is where the problem comes from. Dunno what should be done about this. My preferred solution would be for someone to add support for `%W` to glib, but I guess that will not happen...

psychon:
@adebruyn-abit Do you perhaps want `%V` instead of `%W`? `LC_ALL=C man strftime` describes `%V` as:

 The ISO 8601 week number (see NOTES) of the current year as a decimal number, range 01 to 53, where week 1 is the first week that has at least 4 days in the new year.  See also %U and %W.  (Calculated from tm_year, tm_yday, and tm_wday.)  (SU)

and `%W` as:

The  week  number  of  the  current  year  as  a  decimal number, range 00 to 53, starting with the first Monday as the first day of week 01.  (Calculated from tm_yday and tm_wday.)

and the NOTES say:

[...]

PR #2119, opened 2017-11-21
textclock: Detect failures from g_date_time_format

psychon:
If g_date_time_format() fails, it returns a NULL pointer. This change
makes the textclock detect this and print a warning. This warning
contains 'g_date_time_format()' so that people can ask their favorite
search engine for the supported formats of this function. This warning
also contains the actual format string so that all the information is
available in one place and possible bug reports hopefully include the
format string.

Reference: https://github.com/awesomeWM/awesome/issues/2118

### DIFF
--- a/lib/wibox/widget/textclock.lua
+++ b/lib/wibox/widget/textclock.lua
@@ -44,7 +44,13 @@ function textclock.new(format, timeout, timezone)
     local w = textbox()
     w.force_update = textclock.force_update
     function w._private.textclock_update_cb()
-        w:set_markup(DateTime.new_now(timezone):format(format))
+        local str = DateTime.new_now(timezone):format(format)
+        if str == nil then
+            require("gears.debug").print_warning("textclock: "
+                    .. "g_date_time_format() failed for format "
+                    .. "'" .. format .. "'")
+        end
+        w:set_markup(str)
         w._timer.timeout = calc_timeout(timeout)
         w._timer:again()
         return true -- Continue the timer


Elv13:
```
-- Running tests/examples/wibox/widget/calendar/year.lua…
Result: Child aborted
Error:  lua: lvm.c:436: luaV_equalobj: Assertion `(((t1)->value_).gc)->tt == 5' failed.
CMake Error at CMakeLists.txt:210 (message):

  /home/travis/build/awesomeWM/awesome/tests/examples/wibox/widget/calendar/year.lua

  An example test failed, aborting.
  CMakeLists.txt:295 (run_test)
```

psychon:
To add a bit more context to that: https://travis-ci.org/awesomeWM/awesome/jobs/306592346#L2533 (that's the Travis build for merging this PR into master)

It seems clear that this PR does not cause that problem. Commit 5f5bf24aa8b3b5e8f90bf9ad23f1a53d14798f8b seems like an obvious reason for where this comes from. However, that would only expose an already existing problem...
Line 436 of `lvm.c` is `if (hvalue(t1) == hvalue(t2)) return 1;`. The assertion is... well, there are lots of assertions hidden behind various `#define`s here. So far I traced the wrong one for five minutes. That wrong assert checked that `t1->tt_ == 5 | (1 << 6)` and apparently succeeded. The next assert is in the definition of `gco2t` and is actually the one that failed.

Anyway, no idea what is going on and there were some successful builds on master since the above commit was merged, so.... meh?

PR #2121, opened 2017-11-22
tests: do not set Xft.dpi with integration tests

blueyed:
This should help with covering the methods for detection/fallback during
tests.

Ref: https://github.com/awesomeWM/awesome/pull/2109#issuecomment-346224956

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -150,33 +150,8 @@ wait_until_success() {
     if (( verbose )); then set -x; fi
 }
 
-# Wait for DISPLAY to be available, and setup xrdb,
-# for awesome's xresources backend / queries.
-wait_until_success "setup xrdb" "printf 'Xft.dpi: 96
-    *.background: #002b36
-    *.foreground: #93a1a1
-    *.color0:     #002b36
-    *.color1:     #dc322f
-    *.color10:    #859900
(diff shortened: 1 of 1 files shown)


actionless:
but do we still need xrdb at all? i think other values have fallbacks as well

blueyed:
good point, let's see.

tests/run.sh
```diff
@@ -150,34 +150,6 @@ wait_until_success() {
     if (( verbose )); then set -x; fi
 }
 
-# Wait for DISPLAY to be available, and setup xrdb,
```
psychon: What happened to the `wait for DISPLAY to be available`-part of this? There is a non-polling magic approach to do this instead, but I'm not sure how well that works from a shell script.

blueyed: I hoped it was only necessary for `xrdb` maybe.
We could trap `SIGUSR1` in the shell script - assuming this is the parent process then.
Otherwise looping while awesome's error is `2017-11-22 15:51:47 E: awesome: main:655: cannot open display (error 1)` might also work?!

blueyed: Changed to use `xrdb -q` now.

actionless:
@psychon  i think what to test xrdb it should be a separate testcase

PR #2122, opened 2017-11-24
wibox: Add an input_passthrough property

Elv13:
Also improve the wibox documentation

### DIFF
--- a/docs/common/wibox.ldoc
+++ b/docs/common/wibox.ldoc
@@ -180,6 +180,22 @@
 -- @property shape
 -- @tparam gears.shape shape
 
+--- Forward the inputs to the client below the wibox.
+--
+-- This replace the `shape_input` mask with an empty area. All mouse and
+-- keyboard events are sent to the object (such as a client) positioned below
+-- this wibox. When used alongside compositing, it allows, for example, to have
+-- a subtle transparent wibox on top a fullscreen client to display important
+-- data such as a low battery warning.
+--
+-- **Signal:**
+--
+--  * *property::input_passthrough*
+--
+-- @property input_passthrough
+-- @param[opt=false] boolean
+-- @see shape_input
+
 --- Get or set mouse buttons bindings to a wibox.
 --
 -- @param buttons_table A table of buttons objects, or nothing.
--- a/docs/load_ldoc.cmake
+++ b/docs/load_ldoc.cmake
@@ -13,8 +13,11 @@ foreach(doc_file_name ${doc_files})
     # Remove the file extension
     string(REGEX REPLACE "\\.ldoc" "" DOC_FILE_NAME ${doc_file_name})
 
+    # There is a trailing \n, remove it or it cannot be included in existing blocks
+    string(REGEX REPLACE "\n$" "" doc_file_content "${doc_file_content}")
+
     # Create a new variable usable from lua files
-    set(DOC_${DOC_FILE_NAME}_COMMON "Imported documentation\n\n${doc_file_content}")
+    set(DOC_${DOC_FILE_NAME}_COMMON "${doc_file_content}")
 endforeach()
 
 # vim: filetype=cmake:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80:foldmethod=marker
--- a/lib/awful/wibar.lua
+++ b/lib/awful/wibar.lua
@@ -339,37 +339,19 @@ end
 -- You can also set the screen key with a screen number to attach the wibox.
 -- If not specified, the primary screen is assumed.
 -- @see wibox
--- @tparam[opt=nil] table arg
--- @tparam string arg.position The position.
--- @tparam string arg.stretch If the wibar need to be stretched to fill the screen.
--- @tparam integer arg.border_width Border width.
--- @tparam string arg.border_color Border color.
--- @tparam boolean arg.ontop On top of other windows.
--- @tparam string arg.cursor The mouse cursor.
--- @tparam boolean arg.visible Visibility.
--- @tparam number arg.opacity The wibar's opacity, between 0 and 1.
--- @tparam string arg.type The window type (desktop, normal, dock, …).
--- @tparam integer arg.x The x coordinates.
--- @tparam integer arg.y The y coordinates.
--- @tparam integer arg.width The wibar's width.
--- @tparam integer arg.height The wibar's height.
--- @tparam screen arg.screen The wibox screen.
--- @tparam wibox.widget arg.widget The widget that the wibox displays.
--- @param arg.shape_bounding The wibox’s bounding shape as a (native) cairo surface.
--- @param arg.shape_clip The wibox’s clip shape as a (native) cairo surface.
--- @param arg.shape_input The wibox’s input shape as a (native) cairo surface.
--- @tparam color arg.bg The wibar's background.
--- @tparam surface arg.bgimage The background image of the drawable.
--- @tparam color arg.fg The wibar's foreground (text) color.
+-- @tparam[opt=nil] table args
+-- @tparam string args.position The position.
+-- @tparam string args.stretch If the wibar need to be stretched to fill the screen.
+--@DOC_wibox_constructor_COMMON@
 -- @return The new wibar
 -- @function awful.wibar
-function awfulwibar.new(arg)
-    arg = arg or {}
-    local position = arg.position or "top"
+function awfulwibar.new(args)
+    args = args or {}
+    local position = args.position or "top"
     local has_to_stretch = true
-    local screen = get_screen(arg.screen or 1)
+    local screen = get_screen(args.screen or 1)
 
-    arg.type = arg.type or "dock"
+    args.type = args.type or "dock"
 
     if position ~= "top" and position ~="bottom"
             and position ~= "left" and position ~= "right" then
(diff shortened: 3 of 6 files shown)


Elv13:
Umm, there seems to be an unexpected side effect in `test-struts.lua`

```
Error: running function for step 20/21 (@2): /home/travis/build/awesomeWM/awesome/tests/test-struts.lua:403: 0 + 64 == 0 called from line 513

stack traceback:

	[C]: in function 'assert'

	/home/travis/build/awesomeWM/awesome/tests/test-struts.lua:403: in function 'test_workarea'

	/home/travis/build/awesomeWM/awesome/tests/test-struts.lua:513: in function </home/travis/build/awesomeWM/awesome/tests/test-struts.lua:475>

	(...tail calls...)

	[C]: in function 'xpcall'

	/home/travis/build/awesomeWM/awesome/tests/_runner.lua:73: in function </home/travis/build/awesomeWM/awesome/tests/_runner.lua:64>

	[C]: in function 'xpcall'

	(...tail calls...)

	lib/gears/timer.lua:231: in function <lib/gears/timer.lua:229>!

===> ERROR running /home/travis/build/awesomeWM/awesome/tests/test-struts.lua <===

Error: running function for step 20/21 (@2): /home/travis/build/awesomeWM/awesome/tests/test-struts.lua:403: 0 + 64 == 0 called from line 513
```

It doesn't seem to happen locally.

docs/common/wibox.ldoc
```diff
@@ -180,6 +180,15 @@
 -- @property shape
 -- @tparam gears.shape shape
 
+--- Forward the inputs to the client below the wibox.
```
psychon: I would mention that this is implemented by setting `shape_input` so that it becomes clearer that the two properties interact.

lib/wibox/init.lua
```diff
@@ -137,10 +137,23 @@ function wibox:_apply_shape()
...
+    if not self._input_passthrough then
+        self.shape_input = nil
+        return
+    end
+
+    local geo = self:geometry()
+    local bw = self.border_width
+
+    -- First handle the bounding shape (things including the border)
+    local img = cairo.ImageSurface(cairo.Format.A1, geo.width + 2*bw, geo.height + 2*bw)
```
psychon: Is it enough to use size `0x0` for the image surface? I would think so.

lib/wibox/init.lua
```diff
@@ -137,10 +137,23 @@ function wibox:_apply_shape()
...
+function wibox:_apply_input_shape()
+    if not self._input_passthrough then
+        self.shape_input = nil
+        return
+    end
+
+    local geo = self:geometry()
+    local bw = self.border_width
+
+    -- First handle the bounding shape (things including the border)
```
psychon: Copy&paste

lib/wibox/init.lua
```diff
@@ -151,6 +164,16 @@ function wibox:get_shape()
     return self._shape
 end
 
+function wibox:set_input_passthrough(value)
+    rawset(self, "_input_passthrough", value)
+    self:_apply_input_shape()
```
psychon: Why the separate function? Just do
```lua
if value then
    local img = cairo.ImageSurface(cairo.Format.A1, 0, 0)
    self.shape_inut = img._native
    img:finish()
else
    self.shape_input = nil
end
rawset(self, "_input_passthrough", value)
self:emit_signal("property::input_passthrough", value)
```

lib/wibox/init.lua
```diff
@@ -325,6 +348,7 @@ local function new(args)
     end
 
     ret.shape = args.shape
+    ret.input_passthrough = args.input_passthrough
```
psychon: I'd wrap this in an `if args.input_passthrough then`.

tests/test-wibox-shape.lua
```diff
@@ -18,14 +18,103 @@ local wb = wibox {
...
     end
-})
+}
+
+local function click(x, y)
+    table.insert(steps, function()
+        mouse.coords{x=x, y=y}
+        return true
+    end)
+    table.insert(steps, function()
```
psychon: You can merge these two into one function, I think. At least I do not see why not.

tests/test-wibox-shape.lua
```diff
@@ -18,14 +18,103 @@ local wb = wibox {
...
+    end)
+end
+
+-- Check each corner
+click(0 , 100)
+click(99, 100)
+click(99, 199)
+click(0 , 199)
+
+-- Batch the count check because they fail a couple time each
```
psychon: Just add a call to `awesome.sync()` after all the clicks and one "idle loop" of the test runner afterwards (= `check_count` must stay in a separate step, the other steps sending the clicks can be merged) and this should not need a loop anymore.

tests/test-wibox-shape.lua
```diff
@@ -18,14 +18,103 @@ local wb = wibox {
...
+check_count(4)
+
+table.insert(steps, function()
+    wb.input_passthrough = true
+    count = 0
+    awesome.sync()
```
psychon: This sync should not be necessary. I assume it is for making sure that `input_passthrough` was handled by the X server, but that should not be necessary. I think you just need a `sync()` when you wait for events generated by the X server.

tests/test-wibox-shape.lua
```diff
@@ -18,14 +18,103 @@ local wb = wibox {
...
+end)
+
+
+-- Do it again
+click(0 , 100)
+click(99, 100)
+click(99, 199)
+click(0 , 199)
+
+-- Batch the count check because they fail a couple time each
```
psychon: Same as above.

tests/test-wibox-shape.lua
```diff
@@ -18,14 +18,103 @@ local wb = wibox {
...
+check_count(0)
+
+table.insert(steps, function()
+    wb.input_passthrough = false
+    count = 0
+    return true
+end)
+
+-- It takes a while for the shape to be applied, waste time
```
psychon: No it does not. Well, okay, it does, but that should not be the problem here. The only thing that needs a `sync()` is when you fake input and the next step of the test expects that input to be handled already.

lib/awful/wibar.lua
```diff
@@ -339,37 +339,19 @@ end
...
+-- @tparam string args.position The position.
+-- @tparam string args.stretch If the wibar need to be stretched to fill the screen.
+--@DOC_wibox_constructor_COMMON@
 -- @return The new wibar
 -- @function awful.wibar
-function awfulwibar.new(arg)
-    arg = arg or {}
-    local position = arg.position or "top"
+function awfulwibar.new(args)
+    args = args or {}
```
psychon: This seems a lot like an (almost) unrelated change that should be in its own commit, but okay, whatever....

docs/load_ldoc.cmake
```diff
@@ -13,8 +13,11 @@ foreach(doc_file_name ${doc_files})
     # Remove the file extension
     string(REGEX REPLACE "\\.ldoc" "" DOC_FILE_NAME ${doc_file_name})
 
+    # There is a trailing \n, remove it or it cannot be included in existing blocks
+    string(REGEX REPLACE "\n$" "" doc_file_content "${doc_file_content}")
+
     # Create a new variable usable from lua files
-    set(DOC_${DOC_FILE_NAME}_COMMON "Imported documentation\n\n${doc_file_content}")
+    set(DOC_${DOC_FILE_NAME}_COMMON "${doc_file_content}")
```
psychon: This should cause lots of changes in the docs. Is this really okay? (I have honestly no idea)
At least the commit message could mention/explain this... (and also why it is needed)

Elv13: Its necessary to be able to do

```
--- Foobar.
-- @tparam table args
-- @tparam args.new_property
-- @DOC_wibox_constructor_COMMON@
-- @treturn boolean
function new(args) end
```

Without this, the previous result was

```
--- Foobar.
-- @tparam table args
-- @tparam args.new_property

-- Imported documentation

-- @tparam args.shared_property

-- @treturn boolean
function new(args) end
```

instead of :

```
--- Foobar.
-- @tparam table args
-- @tparam args.new_property
-- @tparam args.shared_property
-- @treturn boolean
function new(args) end
```

It doesn't change anything for existing doc beside the lack of newlines as long as there is a new line between the macros

Ok:

```
--@DOC_widget_COMMON@

--@DOC_object_COMMON@
```

Potentially not ok, but nevertheless mostly ok:

```
--@DOC_widget_COMMON@
--@DOC_object_COMMON@
```

Elv13:
All done

docs/common/wibox.ldoc
```diff
@@ -180,6 +180,22 @@
...
+--
+-- This replace the `shape_input` mask with an empty area. All mouse and
+-- keyboard events are sent to the object (such as a client) positionned below
```
psychon: `positionned` -> `positioned`

lib/wibox/init.lua
```diff
@@ -137,11 +137,6 @@ function wibox:_apply_shape()
     img:finish()
 end
 
---- Set the wibox shape.
--- @property shape
--- @tparam gears.shape A gears.shape compatible function.
--- @see gears.shape
```
psychon: Huh? Why is this removed?

tests/test-wibox-shape.lua
```diff
@@ -18,14 +18,107 @@ local wb = wibox {
...
+            root.fake_input("button_press", 1)
+        end
+
+        return true
+    end)
+end
+
+local function check_count(cnt)
+    table.insert(steps, function()
+        awesome.sync()
```
psychon: I think this `sync` should be before the `return true` in `click` and then this function can just `return cnt == count`, but feel free to ignore me here....

lib/wibox/init.lua
Elv13: It's in `docs/common/wibox.ldoc` too

Issue #1235, opened 2016-11-25
Use GdkPixbuf directly


psychon:
Currently, the C api provides `awesome.load_image` to load an image. Internally, this uses `GdkPixbuf` to load the image and then has some hand-written code to convert this into a cairo surface. This was necessary, because back when this code was written (AFAIR), there was no API entry point that provided this functionality.

However, [since Gdk 3.10](https://developer.gnome.org/gdk3/stable/gdk3-Cairo-Interaction.html#gdk-cairo-surface-create-from-pixbuf) there is now a function that does just this, as noticed here: https://github.com/awesomeWM/awesome/issues/908#issuecomment-262906787.
This means that if we are ok with adding a dependency on Gdk (are we?), we could drop `awesome.load_image` and do "all of this" in Lua instead.

psychon:
Apparently using Gdk comes with lots of problems, so this won't work. But still, we could just export our own `gdk_cairo_surface_create_from_pixbuf`. That way, Lua can actually use GdkPixbuf for stuff, which means that animations come a step closer....

Issue #650, opened 2016-01-24
Use full hierarchy for clients drawable


Elv13:
Instead of hardcoding set_titlebar, I propose having a fully hierarchy for clients

Use case:
- Implement tabs in a sane way
- Making titlebars more flexible for themes
- Drop the shape_clip API, it doesn't support antialiasing or complex borders.
- Drop clients border_width, use widgets instead

Complex border examples:
- http://i.imgur.com/th92K1o.png
- https://liquidat.files.wordpress.com/2007/10/plasma.png

Elv13:
Something like:

```lua
c:setup {
    {
        {
            awful.titlebar(c, {args}).
            c.widget,
            { --[[some bottom widget]] },
            layout = wibox.layout.align.vertical
        },
        margins = 12,
        widget = wibox.container.margins,
    },
    fg = "unicorn pink",
    widget = wibox.container.background
}
```

The current titlebar API could be implemented on top of that to keep the compatibility, then deprecate it for v5 (the current titlebar API is a mess)

psychon:
@Elv13 Is this something like what you have in mind? https://gist.github.com/psychon/ed4436cd5ad36da57c6dd81c69ac8e1c

psychon:
@Elv13 What would the `awful.titlebar` widget in your example actually do? It cannot modify the client's geometry (e.g. feedback loop, but also: bad idea), so it can only change the size of the titlebars. However, that in turn requires a full relayout, which means that the `awful.titlebar`-widget gets assigned a new size which means.... and we end up with a feedback loop again.

In a wibox, the size of the wibox is given and the code "just" has to place widgets in there.
In a client, the geometry of the client is given, but the size of the titlebars is free. My example code from my previous comment does not resize the titlebars, but assumes that their size is given so that the same thing as in the wibox case can be done.

So, any idea how the size of a titlebar is computed? And what would your `awful.titlebar` widget actually do (it can only resize titlebars and that doesn't fit well with the widget API).

Elv13:
> @Elv13 Is this something like what you have in mind?

With some love, it will do just fine, thanks a lot. I feel ashamed I didn't see the obvious earlier. I will work on it later. I am currently busy finishing a PR serie that stretch back to october (the improved doc)

> What would the awful.titlebar widget in your example actually do?

Don't worry, I think with some work on the hierarchy handler, it can behave exactly like the titlebar currently does.

psychon:
[...]

PR #2123, opened 2017-11-24
doc: Add a titlebar example

Elv13:
Another few commits from the ever expanding cache of unsubmitted commits...

For the surface hack: sorry, but it works...

screenshot_20171123_234113

### DIFF
--- a/lib/awful/titlebar.lua
+++ b/lib/awful/titlebar.lua
@@ -513,6 +513,9 @@ local function new(c, args)
     -- Handle declarative/recursive widget container
     ret.setup = base.widget.setup
 
+    c._private = c._private or {}
+    c._private.titlebars = bars
+
     return ret
 end
 
--- a/lib/awful/titlebar.lua
+++ b/lib/awful/titlebar.lua
@@ -1,9 +1,15 @@
 ---------------------------------------------------------------------------
 --- Titlebars for awful.
+--**Create a titlebar:**
+--
+-- This example reproduces what the default `rc.lua` does. It shows how to
+-- handle the titlebars on a lower level.
+--
+-- @DOC_awful_titlebar_default_EXAMPLE@
 --
 -- @author Uli Schlachter
 -- @copyright 2012 Uli Schlachter
--- @module awful.titlebar
+-- @classmod awful.titlebar
 ---------------------------------------------------------------------------
 
 local error = error
@@ -447,8 +453,7 @@ local function get_titlebar_function(c, position)
     end
 end
 
---- Get a client's titlebar
--- @class function
+--- Get a client's titlebar.
 -- @tparam client c The client for which a titlebar is wanted.
 -- @tparam[opt={}] table args A table with extra arguments for the titlebar.
 -- @tparam[opt=font.height*1.5] number args.size The height of the titlebar.
@@ -461,7 +466,7 @@ end
 -- @tparam[opt=top] string args.fg_normal
 -- @tparam[opt=top] string args.fg_focus
 -- @tparam[opt=top] string args.font
--- @name titlebar
+-- @function awful.titlebar
 local function new(c, args)
     args = args or {}
     local position = args.position or "top"
--- a/lib/gears/surface.lua
+++ b/lib/gears/surface.lua
@@ -51,7 +51,13 @@ function surface.load_uncached_silently(_surface, default)
         if not pixbuf then
             return get_default(default), tostring(err)
         end
-        _surface = capi.awesome.pixbuf_to_surface(pixbuf._native)
+        _surface = capi.awesome.pixbuf_to_surface(pixbuf._native, _surface)
+
+        -- The shims implement load_image() to return a surface directly,
+        -- instead of a lightuserdatum.
+        if cairo.Surface:is_type_of(_surface) then
+            return _surface
+        end
     end
     -- Everything else gets forced into a surface
     return cairo.Surface(_surface, true)
--- a/tests/examples/awful/titlebar/default.lua
+++ b/tests/examples/awful/titlebar/default.lua
@@ -1,0 +1,60 @@
+--DOC_NO_USAGE
+local place = require("awful.placement") --DOC_HIDE
+local awful = { titlebar = require("awful.titlebar"), --DOC_HIDE
+                button   = require("awful.button"), --DOC_HIDE
+              } --DOC_HIDE
+local wibox = require("wibox") --DOC_HIDE
+local gears = {table = require("gears.table")} --DOC_HIDE
+
+local c = client.gen_fake {hide_first=true} --DOC_HIDE
+place.maximize(c, {honor_padding=true, honor_workarea=true}) --DOC_HIDE
+
+    -- Create a titlebar for the client.
+    -- By default, `awful.rules` will create one, but all it does is to call this
+    -- function.
+
+    local top_titlebar = awful.titlebar(c, {
+        height    = 20,
+        bg_normal = "#ff0000",
+    })
+
+    -- buttons for the titlebar
+    local buttons = gears.table.join(
+        awful.button({ }, 1, function()
+            client.focus = c
+            c:raise()
+            awful.mouse.client.move(c)
+        end),
+        awful.button({ }, 3, function()
+            client.focus = c
+            c:raise()
+            awful.mouse.client.resize(c)
+        end)
+    )
+
+    top_titlebar : setup {
+        { -- Left
+            awful.titlebar.widget.iconwidget(c),
+            buttons = buttons,
+            layout  = wibox.layout.fixed.horizontal
+        },
+        { -- Middle
+            { -- Title
+                align  = "center",
+                widget = awful.titlebar.widget.titlewidget(c)
+            },
+            buttons = buttons,
+            layout  = wibox.layout.flex.horizontal
+        },
+        { -- Right
+            awful.titlebar.widget.floatingbutton (c),
+            awful.titlebar.widget.maximizedbutton(c),
+            awful.titlebar.widget.stickybutton   (c),
+            awful.titlebar.widget.ontopbutton    (c),
+            awful.titlebar.widget.closebutton    (c),
+            layout = wibox.layout.fixed.horizontal()
+        },
+        layout = wibox.layout.align.horizontal
+    }
+
+--DOC_HIDE vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/tests/examples/shims/awesome.lua
+++ b/tests/examples/shims/awesome.lua
@@ -1,6 +1,4 @@
 local lgi       = require("lgi")
-local GdkPixbuf = lgi.GdkPixbuf
-local Gdk       = lgi.Gdk
 local gears_obj = require("gears.object")
 
 -- Emulate the C API classes. They differ from C API objects as connect_signal
@@ -48,27 +46,10 @@ awesome.startup = true
 function awesome.register_xproperty()
 end
 
-local init, surfaces = false, {}
+awesome.load_image = lgi.cairo.ImageSurface.create_from_png
 
-function awesome.load_image(file)
-    if not init then
-        Gdk.init{}
-        init = true
-    end
-
-    local _, width, height = GdkPixbuf.Pixbuf.get_file_info(file)
-
-    local pixbuf = GdkPixbuf.Pixbuf.new_from_file_at_size(file, width, height)
-
-    if not pixbuf then
-        return nil, "Could not load "..file
-    end
-
-    local s = Gdk.cairo_surface_create_from_pixbuf( pixbuf, 1, nil )
-
-    table.insert(surfaces, s)
-
-    return s._native, not s and "Could not load surface from "..file or nil, s
+function awesome.pixbuf_to_surface(_, path)
+    return awesome.load_image(path)
 end
 
 -- Always show deprecated messages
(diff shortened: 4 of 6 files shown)


Elv13:
umm... it will take some work to mute the GDK warning...

Elv13:
Updated

tests/examples/shims/awesome.lua
```diff
@@ -68,7 +69,11 @@ function awesome.load_image(file)
 
     table.insert(surfaces, s)
 
-    return s._native, not s and "Could not load surface from "..file or nil, s
+    return s._native, not s and "Could not load surface from "..file or nil, s]]
+
+    local image = lgi.cairo.ImageSurface.create_from_png(file)
+
+    return image._native, nil, image
```
psychon: How's the ownership here? `awesome.load_image` returns a cairo surface with a reference that must be freed by Lua. Here, no extra reference is taken. Will this not lead to a double free? I tried to reproduce this problem, but I don't understand why the following does not crash: `lua -e 'c = require("lgi").cairo img = c.ImageSurface.create_from_png("does_not_exist") img2 = c.Surface(img._native, true) print(img == img2)'`

tests/examples/shims/awesome.lua
```diff
@@ -48,16 +46,19 @@ awesome.startup = true
 function awesome.register_xproperty()
 end
 
-local init, surfaces = false, {}
+-- local init, surfaces = false, {}
 
 function awesome.load_image(file)
+    -- The code below is actually the closest to what Awesome really does, but
+    -- it is creating a GDK warning when running without a display server (but
+    -- still works)
```
psychon: Oh and: How did this end up in git? Shouldn't the commit that added this code already have failed?

lib/gears/surface.lua
```diff
@@ -45,17 +45,23 @@ function surface.load_uncached_silently(_surface, default)
...
+        _surface, err, s = capi.awesome.load_image(file)
+
         if not _surface then
             return get_default(default), err
         end
     end
+
     -- Everything else gets forced into a surface
-    return cairo.Surface(_surface, true)
+    return s or cairo.Surface(_surface, true)
```
psychon: Can't the shims just override `gears.surface` instead? I.e. `require("gears.surface").foo = my_shim`?
That way, no hacks end up "in the real code".

lib/awful/titlebar.lua
```diff
@@ -513,6 +513,8 @@ local function new(c, args)
     -- Handle declarative/recursive widget container
     ret.setup = base.widget.setup
 
+    c.titlebars = bars
```
psychon: Can this commit be squashed with whatever makes use of this property?
Put differently: I don't see any use of `c.titlebars` in the following commits.

lib/gears/surface.lua
psychon: Also, why does the commit message refer to #908? That seems quite unrelated.

tests/examples/shims/awesome.lua
Elv13: The ownership is "tricked" to work in a weird way:

1) The surface cache prevents the reference count from going to zero, so it prevents the use after free
2) The `c.Surface(img._native)` is never called due to the `gears.surface` hack, so there is no double free

But umm, it isn't very pretty, isn't it? I can't require the shims are you propose from `capi.awesome` as this would create a circular dependency. I also can't just do that in every test file that calls this as it would be horrible.

The other option would be to add `capi.awesome._is_test = true` and check that from the `gears.surface` and never call `load_image` (and put the png code there). It isn't that much better as both "solution" introduce test code into awesome. They are both equally ugly, so I let you device.

lib/gears/surface.lua
Elv13: https://github.com/awesomeWM/awesome/issues/1235

tests/examples/shims/awesome.lua
Elv13: An attempt at breaking up a huge PR you (rightfully) rejected as too big ended up creating a semi tangled mess of smaller PRs. It was merged as part of a squashed set of other shims commit but wasn't used at that time. As it works perfectly when an X11 server is present, it went unnoticed.

tests/examples/shims/awesome.lua
psychon: > The surface cache prevents the reference count from going to zero, so it prevents the use after free

Uhm, the cache is a weak table, so it does not really prevent anything?!? Also, when the Lua state is closed, then "stuff will be freed" anyway.

> I can't require the shims are you propose from capi.awesome as this would create a circular dependency.

How about something like this?
```lua
local surface_shim = {}
function surface_shim:load()
end
package.loaded["gears.surface"] = surface_shim
```

lib/gears/surface.lua
psychon: I don't get it. O.O

tests/examples/shims/awesome.lua
Elv13: > Uhm, the cache is a weak table, so it does not really prevent anything?!? Also, when the Lua state is closed, then "stuff will be freed" anyway.

The shims has it's own cache to prevent this. It was my first "trick" to make it work, but it was racy and would sometime cause a double free depending on the GC lottery. However it wasn't enough and I added the other hack.

> How about something like this?

You want me to shim the *whole* `gears.surface` API instead of a 2 line workaround? That seems counter intuitive if the point of not using the proposed hack is to make it more maintainable...

lib/gears/surface.lua
Elv13: Long version:

This hack is required because we have something called `capi.awesome.load_image()` that should not exist and should have used LGI all along, but could not. Now it could (https://github.com/awesomeWM/awesome/issues/908#issuecomment-262906787) but that would (maybe?) cause us to bump the minimum LGI/GDK version.

The second problem is that GDK prints a big fat warning on stderr when X11 isn't present and this causes the tests to fail on the CI. Implementing a REGEX to ignore that warning could be a solution and the `gears.surface` part of the hack could be avoided. But it would also set a precedent and pollute the already messy code that generates the images.

So this PR adds a couple of ugly commits to work around the issue. It has to do so to be able to display the titlebar **and** the layoutlist in the other PR. It is also required to "complete" the shims to be able to render arbitrary screenshots and even potentially animated SVG.

tests/examples/shims/awesome.lua
psychon: Well, so far my point was to make it work at all, but... dunno, does it work?
Also, I just wondered why this shim is necessary at all. Which code uses it and what images are being loaded? By the time the shims run, "nothing" was installed yet, so e.g. titlebar icons most likely will not be found. I don't remember code in this PR that made references to uninstalled images...?

Elv13: The titlebar and the layoutlist require this code. It is called and without it errors are printed

psychon: How do the titlebar and layoutlist find their icons in the source directory? Doesn't `$SOMETHING` have to patch in the correct pahts?
(Yes, I just switched issues to discuss, sorry)

Elv13: Sure, and it does set those variables ;)

Elv13:
Sooooo

This is going a bit in circle so I will summarize what's above:

### Option 1 (As some in this PR)

Add the smallest possible changes to `gears.surface` to bypass the `_native` and use the surface directly. It isn't pretty, but the problem is that `_native` should not be used there. but have to. I vote for this solution.

### Option 2

Add `_from_png` code directly into `gears.surface` in case `capi.awesome._is_test` is set. It's ugly because it add test specific code in `gears`. On the other hand it's not a hack. It's just 2 clean code paths. That's my second choice

### Option 3

Use GDK in gears and add a REGEX to the test to ignore the warnings being printed on stderr. It may or may not break Awesome for some users having older version (not investigated). It is also a very intrusive change and has risks of side effects.

### Option 4

Shim the whole `gears.surface`. I really don't want to do that.

psychon:
How about this option (basically your Option 1, but smaller and hopefully less magical (by getting rid of the "have to"))?
```diff
diff --git a/lib/gears/surface.lua b/lib/gears/surface.lua
index 3e506b9b..7a10cc28 100644
--- a/lib/gears/surface.lua
+++ b/lib/gears/surface.lua
@@ -53,6 +53,11 @@ function surface.load_uncached_silently(_surface, default)
         if not _surface then
             return get_default(default), err
         end
+        -- The shims implement load_image() to return a surface directly,
+        -- instead of a lightuserdatum.
+        if cairo.Surface:is_type_of(_surface) then
+            return _surface
+        end
     end
     -- Everything else gets forced into a surface
     return cairo.Surface(_surface, true)
```
(Sorry for the circles; I really should use "GitHub comments on specific lines" less, because the resulting order of things is just a mess.)

Elv13:
I updated the PR with the changes you proposed. Travis is working again, it was down for ~30 hours. Feel free to squash the commits if you wish.

Elv13:
@psychon One of the Travis target crashed. I will leave it as-is so you can take a look. It's not the first time it happens. It's always the Lua 5.3 builder.  If you restart the build there is 95%+ chances it will work.

psychon:
```
awesome: lvm.c:436: luaV_equalobj: Assertion `(((t1)->value_).gc)->tt == 5' failed.
```
That's due to 5f5bf24aa8b3b5e8f90bf9ad23f1a53d14798f8b. I don't know much more than this and feel like lgi is perhaps a bit too broken. :-(

tests/examples/shims/awesome.lua
```diff
@@ -68,7 +69,11 @@ function awesome.load_image(file)
 
     table.insert(surfaces, s)
 
-    return s._native, not s and "Could not load surface from "..file or nil, s
+    return s._native, not s and "Could not load surface from "..file or nil, s]]
+
+    local image = lgi.cairo.ImageSurface.create_from_png(file)
+
+    return image, nil
```
psychon: Just do `awesome.load_image = lgi.cairo.ImageSurface.create_from_png`. The extra `nil` value isn't needed (the C code doesn't return it in the success case either).

lib/awful/titlebar.lua
psychon: Where is this used? I don't think I saw any use in this PR, so just drop this commit from this PR.

lib/gears/surface.lua
```diff
@@ -45,15 +45,28 @@ function surface.load_uncached_silently(_surface, default)
...
+
         if not _surface then
             return get_default(default), err
         end
+
+        -- The shims implement load_image() to return a surface directly,
+        -- instead of a lightuserdatum.
+        if cairo.Surface:is_type_of(_surface) then
+            return _surface
+        end
```
psychon: Please take a look at the third commit and remove all changes except for this one (especially the addition of a wrong comment).

tests/examples/awful/titlebar/default.lua
```diff
@@ -0,0 +1,60 @@
...
+local awful = { titlebar = require("awful.titlebar"), --DOC_HIDE
+                button   = require("awful.button"), --DOC_HIDE
+              } --DOC_HIDE
+local wibox = require("wibox") --DOC_HIDE
+local gears = {table = require("gears.table")} --DOC_HIDE
+
+local c = client.gen_fake {hide_first=true} --DOC_HIDE
+place.maximize(c, {honor_padding=true, honor_workarea=true}) --DOC_HIDE
+
+    -- Create a titlebar for the client. 2
```
psychon: 2?

lib/awful/titlebar.lua
Elv13: It is used in the awful/template.lua

Elv13:
updated

lib/awful/titlebar.lua
psychon: Ah, some already-merged commit uses this. Okay, so I'll have to take a look at how this stuff works... The drawin shim doesn't actually do anything and certainly cannot be drawn to.... Can't you do something similar to baf290a8b22560d3d860bb75ec1429442417093b in `wibox.drawable`? I.e. add a `d.get_wibox_drawable()` function in there? That way the shim client can keep a reference to the titlebar drawin that it returns, the drawin has a reference to its drawable, the new function allows you to get the wibox.drawable from this and this is (basically) all that `wrap_titlebar()` from `awful/template.lua` needs (widget, bg and fg color).

Thus, instead of making this implementation detail of `awful.titlebar` public, we just need "something similar" in `wibox.drawable` to what you already did in `wibox`.

What do you think?

Elv13: > so I'll have to take a look at how this stuff works...

Ping?

Please note that it  is a private API. It can be changed later. Maybe it's better to keep it that way given #650 might cause us to break the low level titlebar API once either you or me decide to complete the implementation. Also, it's blocking the `popup` PR progress and, in turn, the v4.3 release.

psychon: The `...` was meant to be the time that I used for looking at the shim's code. I did not actually mean something like "tomorrow".

Well, undocumented does not mean that people will not use it. And if the name does not start with e.g. `_private`, I wouldn't even really blame them for that.

What do you think about my proposal above to instead have the shims actually track the titlebar drawables that they hand out, instead of later trying to reconstruct that relation? Is it really better, or is it mostly "meh, whatever"?

psychon: Ping?

Elv13:
rebased and updated.

 * Rebase / fix / adapt to the lua GDK pixbuf changes
 * Add `c._private` to move the titlebars away from the undocumented "public API". Keep that way to avoid interfering with #650

Elv13:
Ping? https://github.com/awesomeWM/awesome/pull/2123#discussion_r167015035 was addressed in last week update.

lib/gears/surface.lua
psychon: This comment still applies but now to the second commit and in a different form: It only adds empty lines.

psychon:
Oh and: Sorry that I missed the update previously.

PR #2124, opened 2017-11-24
Add spacing widgets to layouts

Elv13:
This follows the separator pull request from last summer. Someone emailed me earlier today about the status of this, so lets not sit on these commits for longer.

Why:

Mostly because people want to do powerline stuff and have widgets between the taglist/tasklist and their right side monitor widgets.

screenshot_20171123_235407

This is part 2 of 3 for this series. The last one makes the tasklist/taglist more customizable, including spacing widgets (and cleanup the code to bring back some sanity to it).

### DIFF
--- a/lib/wibox/layout/fixed.lua
+++ b/lib/wibox/layout/fixed.lua
@@ -23,10 +23,15 @@ local fixed = {}
 function fixed:layout(context, width, height)
     local result = {}
     local pos,spacing = 0, self._private.spacing
+    local spacing_widget = self._private.spacing_widget
+    local is_y = self._private.dir == "y"
+    local is_x = not is_y
+    local abspace = math.abs(spacing)
+    local spoffset = spacing < 0 and 0 or spacing
 
     for k, v in pairs(self._private.widgets) do
         local x, y, w, h, _
-        if self._private.dir == "y" then
+        if is_y then
             x, y = 0, pos
             w, h = width, height - pos
             if k ~= #self._private.widgets or not self._private.fill_space then
@@ -42,9 +47,17 @@ function fixed:layout(context, width, height)
             pos = pos + w + spacing
         end
 
-        if (self._private.dir == "y" and pos-spacing > height) or
-            (self._private.dir ~= "y" and pos-spacing > width) then
+        if (is_y and pos-spacing > height) or
+            (is_x and pos-spacing > width) then
             break
+        end
+
+        -- Add the spacing widget
+        if k > 1 and abspace > 0 and spacing_widget then
+            table.insert(result, base.place_widget_at(
+                spacing_widget, is_x and (x - spoffset) or x, is_y and (y - spoffset) or y,
+                is_x and abspace or w, is_y and abspace or h
+            ))
         end
         table.insert(result, base.place_widget_at(v, x, y, w, h))
     end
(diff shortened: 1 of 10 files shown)


lib/wibox/layout/fixed.lua
```diff
@@ -23,10 +23,15 @@ local fixed = {}
 function fixed:layout(context, width, height)
     local result = {}
     local pos,spacing = 0, self._private.spacing
+    local spacing_widget = self._private.spacing_widget
+    local is_y = self._private.dir == "y"
+    local is_x = not is_y
+    local abspace = math.abs(spacing)
```
psychon: Wait, I can have negative spacing between widgets and can show a spacing widget in this non-existent space. Why??

lib/wibox/layout/fixed.lua
```diff
@@ -42,8 +47,16 @@ function fixed:layout(context, width, height)
...
+        if k > 1 and abspace > 0 and spacing_widget then
+            table.insert(result, base.place_widget_at(
+                spacing_widget, is_x and (x - spoffset) or x, is_y and (y - spoffset) or y,
+                is_x and abspace or w, is_y and abspace or h
+            ))
+        end
+
+        if (is_y and pos-spacing > height) or
+            (is_x and pos-spacing > width) then
             break
```
psychon: Why is this check after "add a spacing widget" and not before it? If the layout contains so many widgets that not all of them can be displayed, there will be a spacing widget after the last visible widget and that spacing widget will be outside of the space assigned to this `fixed` layout.
Also, you might want to think about what should happen in this situation. I'd say a spacing widget should only be shown if there is `spacing` space for it (or if `spacing` is negative...). If e.g. only half of `spacing` is available, then no widget should be shown. But that's just an idea, feel free to come up with other ones. (Dunno if we should have tests for the behaviour in this case.)

lib/wibox/layout/flex.lua
```diff
@@ -50,14 +50,35 @@ local flex = {}
...
+--
+-- @property spacing_widget
+-- @param widget
+
+--- Add spacing between each layout widgets.
+--
+--@DOC_wibox_layout_flex_spacing_EXAMPLE@
+--
+-- @property spacing
+-- @tparam number spacing Spacing between widgets.
```
psychon: Why is this not in `@DOC_fixed_COMMON@`?

lib/wibox/layout/flex.lua
```diff
@@ -79,10 +100,17 @@ function flex:layout(_, width, height)
...
+        end
+
         pos = pos + space_per_item + spacing
 
-        if (self._private.dir == "y" and pos-spacing >= height) or
-            (self._private.dir ~= "y" and pos-spacing >= width) then
+        if (is_y and pos-spacing >= height) or
+            (is_x and pos-spacing >= width) then
             break
         end
```
psychon: Same comments as for fixed, I think.

lib/wibox/layout/ratio.lua
```diff
@@ -167,6 +188,13 @@ function ratio:layout(context, width, height)
...
 
+        if k > 1 and abspace > 0 and spacing_widget then
+            table.insert(result, base.place_widget_at(
+                spacing_widget, is_x and (x - spoffset) or x, is_y and (y - spoffset) or y,
+                is_x and abspace or w, is_y and abspace or h
+            ))
+        end
+
         table.insert(result, base.place_widget_at(v, x, y, w, h))
     end
```
psychon: Same comments as for fixed and flex.

lib/wibox/layout/fixed.lua
Elv13: See the image above. The powerline example uses it. A while ago we had a discussion and 3 or 4 prototypes how to best create powerlines. However the negative spacing trick is actually from the screenshot page. I found how he/she was quite right, it does work rather well expect for the mouse area issue. The last PR of this series have an example in the doc for the "true" powerline taglist using negative spacing.

screenshot_20171124_035338

lib/wibox/layout/flex.lua
Elv13: The examples are not the same

Elv13:
Updated

psychon:
A shower thought: Should `get_children` include separators?

PR #2127, opened 2017-11-27
awful.widget.common overhaul

Elv13:
First pull request of the third stage of the giant branch I completed last summer. These PRs add the 
`awful.widget.common` delegate system. They also create the `naughty.notification` object (not in this PR) and add a few common notification helper widgets (also not in this PR).

Why (from the commit message):

```
The old function didn't scale at all. As no replacements are going
to be merged anytime soon, start to make it meta-extensible.

This is the first step to be able to let the widget be extended
directly from rc.lua without adding yet more silly parameters.

It never scaled and has reached the point of no return a very long
time ago.

My first ever contribution to Awesome was to attempt to fix this,
but the solution was a bad hack. The radical module later solved
this by delegating the style, layout, theme, item layout and item
style to various "visitor" objects. While this is superior to this
commit, it was also a very large and complicated codebase. After
5 years, it is now obvious it will never be merged "whole".
```

Ref: #1874

### DIFF
--- a/lib/awful/widget/common.lua
+++ b/lib/awful/widget/common.lua
@@ -9,7 +9,9 @@ local type = type
 local ipairs = ipairs
 local capi = { button = button }
 local wibox = require("wibox")
+local gdebug = require("gears.debug")
 local dpi = require("beautiful").xresources.apply_dpi
+local base = require("wibox.widget.base")
 
 --- Common utilities for awful widgets
 local common = {}
@@ -36,6 +38,52 @@ function common.create_buttons(buttons, object)
     end
 end
 
+local function default_template()
+    local ib  = wibox.widget.imagebox()
+    local tb  = wibox.widget.textbox()
+    local bgb = wibox.container.background()
+    local tbm = wibox.container.margin(tb, dpi(4), dpi(4))
+    local ibm = wibox.container.margin(ib, dpi(4))
+    local l   = wibox.layout.fixed.horizontal()
+
+    -- All of this is added in a fixed widget
+    l:fill_space(true)
+    l:add(ibm)
+    l:add(tbm)
+
+    -- And all of this gets a background
+    bgb:set_widget(l)
+
+    return {
+        ib      = ib,
+        tb      = tb,
+        bgb     = bgb,
+        tbm     = tbm,
+        ibm     = ibm,
+        primary = l,
+    }
+end
+
+local function custom_template(args)
+    local l = base.make_widget_from_value(args.widget_template)
+
+    -- The template system requires being able to get children elements by ids.
+    -- This is not optimal, but for now there is no way around it.
+    assert(l.get_children_by_id,"The given widget template did not result in a"..
+        "layout with a 'get_children_by_id' method")
+
+    return {
+        ib              = l:get_children_by_id( "icon_role"        )[1],
+        tb              = l:get_children_by_id( "text_role"        )[1],
+        bgb             = l:get_children_by_id( "background_role"  )[1],
+        tbm             = l:get_children_by_id( "text_margin_role" )[1],
+        ibm             = l:get_children_by_id( "icon_margin_role" )[1],
+        primary         = l,
+        update_callback = l.update_callback,
+        create_callback = l.create_callback,
+    }
+end
+
 --- Common update method.
 -- @param w The widget.
 -- @tab buttons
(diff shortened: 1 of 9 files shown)


lib/awful/widget/taglist.lua
```diff
@@ -365,55 +367,93 @@ local function taglist_update(s, w, buttons, filter, data, style, update_functio
...
+    end
+
+    assert(type(args) == "table")
+
+    local deprecated_param = {filter, buttons, style, update_function, base_widget}
+    local deprecated_names = {"filter", "buttons", "style",
+                              "update_function", "layout"}
+
+    for k=1, 5 do
+        local v = deprecated_param[k]
```
psychon: I'd do the above five lines as `for k, v in pairs { filter = filter, buttons = buttons, style = style, update_function = update_function, layout = base_widget } do` (perhaps with some line breaks. Oh, the following `if v then` can then also be dropped (tables do not have `nil` values). What do you think, would this version be easier?
(My main motivation is getting rid of the hardcoded `5` magic number`)

lib/awful/widget/tasklist.lua
```diff
@@ -363,62 +365,100 @@ local function tasklist_update(s, w, buttons, filter, data, style, update_functi
...
+    local deprecated_names = {"filter", "buttons", "style",
+                              "update_function", "layout"}
+
+    for k=1, 5 do
+        local v = deprecated_param[k]
+        if v then
+            gdebug.deprecate("The `awful.widget.taglist()` `"..deprecated_names[k]
+                .."` paramater is deprecated, use `args."..deprecated_names[k].."`.",
+            {deprecated_in=5})
+            args[deprecated_names[k]] = v
```
psychon: Both comments for the tasklist also apply here

lib/awful/widget/tasklist.lua
```diff
@@ -461,16 +461,19 @@ function tasklist.new(args, filter, buttons, style, update_function, base_widget
...
     end
 
+    -- For the tests
+    function w._do_tasklist_update_now()
+        queued_update = false
+        if screen.valid then
+            tasklist_update(screen, w, args.buttons, args.filter, data, args.style, uf)
+        end
+    end
+
```
psychon: For the commit message: I would replace "a reload" with "an update" (I think that refers to the third commit)

lib/awful/widget/common.lua
```diff
@@ -36,6 +37,51 @@ function common.create_buttons(buttons, object)
...
+        primary = l,
+    }
+end
+
+local function custom_template(args)
+    local l = base.make_widget_from_value(args.widget_template)
+
+    -- For now, only there is no better way. If template could implement this
+    -- function even if it doesn't use the declarative construct.
+    assert(l.get_children_by_id)
```
psychon: Perhaps `assert(l.get_children_by_id, "The given widget template did not result in a layout with a 'get_children_by_id' method")`, or something like that?

lib/awful/widget/common.lua
```diff
@@ -44,73 +90,62 @@ end
...
-        if icon then
-            ib:set_image(icon)
-        else
-            ibm:set_margins(0)
+
+        if cache.bgb then
+            cache.bgb:set_bg(bg)
+
+            --TODO v5 remove this if, it existed only for a removed and
+            -- undocumented API
```
psychon: `gears.debug.deprecate` with `deprecated_in`?

lib/awful/widget/tasklist.lua
```diff
@@ -24,6 +24,50 @@
...
+-- **Customizing the tasklist:**
+--
+-- The `tasklist` created by `rc.lua` use the default values for almost
+-- everything. However, it is possible to override each aspects to create a
+-- very different widget. Here's an example that create a tasklist similar to
+-- the default one, but with an explicit layout and some spacing widgets:
+--
+--@DOC_wibox_awidget_tasklist_rounded_EXAMPLE@
+--
+-- As demonstrated in the example above, there is a few "shortcuts" to avoid
```
psychon: "there *are* a few shortcuts"

lib/awful/widget/tasklist.lua
```diff
@@ -24,6 +24,50 @@
...
+-- Both callback have the same parameters:
+--
+-- * `self`: The widget instance (*widget*).
+-- * `c`: The client (*client*)
+-- * `index`: The widget position in the list (*number*)
+-- * `clients`: The list of client, in order (*table*)
+--
+-- It is also possible to omit some roles and create an icon only tasklist.
+-- Notice that this example use the `awful.widget.clienticon` widget instead
+-- of an `imagebox`. This allows higher resoluton icons to be loaded. This
```
psychon: Uhm, that sounds like something we should somehow make the default config do

lib/awful/widget/tasklist.lua
```diff
@@ -461,15 +511,16 @@ function tasklist.new(args, filter, buttons, style, update_function, base_widget
         w:set_spacing(args.style and args.style.spacing or beautiful.tasklist_spacing)
     end
 
+    local queued_update = false
```
psychon: Oh, I didn't even see this. This is in the wrong commit. The variable `queueued_update` should be moved in the commit that adds `_do_tasklist_update_now`.

lib/awful/widget/tasklist.lua
```diff
@@ -24,6 +24,50 @@
...
 -- </table>
 --
+-- **Customizing the tasklist:**
+--
+-- The `tasklist` created by `rc.lua` use the default values for almost
+-- everything. However, it is possible to override each aspects to create a
+-- very different widget. Here's an example that create a tasklist similar to
+-- the default one, but with an explicit layout and some spacing widgets:
+--
+--@DOC_wibox_awidget_tasklist_rounded_EXAMPLE@
```
psychon: Shouldn't these examples be added in the same commit that refers to them? Otherwise, the commit is not really self-contained.

lib/awful/widget/taglist.lua
```diff
@@ -1,6 +1,37 @@
...
 --
+-- Here is a more advanced example of how to extent the `taglist`. It provides:
+--
+-- * mouse "hover" color
+-- * an extra index field
+-- * a powerline look and feel
+--
+--@DOC_wibox_awidget_taglist_indexed_EXAMPLE@
+--
+-- As demonstrated in the example above, there is a few "shortcuts" to avoid
```
psychon: Again: is -> are

lib/awful/widget/taglist.lua
```diff
@@ -459,15 +491,19 @@ function taglist.new(args, filter, buttons, style, update_function, base_widget)
     local data = setmetatable({}, { __mode = 'k' })
 
     local queued_update = {}
+
+
```
psychon: One empty line too much

lib/awful/widget/taglist.lua
```diff
@@ -459,15 +491,19 @@ function taglist.new(args, filter, buttons, style, update_function, base_widget)
...
+    function w._do_taglist_update_now()
+        if screen.valid then
+            taglist_update(screen, w, args.buttons, args.filter, data, args.style, uf, args)
+        end
+        queued_update[screen] = false
+    end
```
psychon: Heh, the tasklist got a separate commit for this, the taglist does not?

tests/examples/wibox/awidget/taglist/indexed.lua
```diff
@@ -0,0 +1,106 @@
...
+                left  = 18,
+                right = 18,
+                widget = wibox.container.margin
+            },
+            id     = "background_role",
+            widget = wibox.container.background,
+
+            -- Add support for hover colors and an index label
+            create_callback = function(self, c3, index, objects) --luacheck: no unused args
+                self:get_children_by_id("index_role")[1].markup = "<b> "..index.." </b>"
```
psychon: Can this be replaced with `self:update_callback(........)` or something like that? I guess not...

lib/awful/widget/taglist.lua
```diff
@@ -365,55 +367,93 @@ local function taglist_update(s, w, buttons, filter, data, style, update_functio
...
+    local deprecated_names = {"filter", "buttons", "style",
+                              "update_function", "layout"}
+
+    for k=1, 5 do
+        local v = deprecated_param[k]
+        if v then
+            gdebug.deprecate("The `awful.widget.taglist()` `"..deprecated_names[k]
+                .."` paramater is deprecated, use `args."..deprecated_names[k].."`.",
+            {deprecated_in=5})
+            args[deprecated_names[k]] = v
```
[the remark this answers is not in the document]
Elv13: Umm. I guess so... However given this can only affect the users who *both* have migrated to the new `args` syntax *and* still have deprecated parameters, I would argue nobody will ever be realistically affected. What do you think?

lib/awful/widget/tasklist.lua
Elv13: Probably, but it can wait. I would prefer not to add stuff to this PR.

lib/awful/widget/tasklist.lua
Elv13: It doesn't change anything since the macro expands to nothing (thus ldoc works). I mostly ignored this type of "issue" in each of the PRs I submitted, I don't think it affects anyone and makes the commits overall smaller.

I would have to add the documentation in the first commit (otherwise I add undocumented features) and then edit the description in the second to refer to the example, modifying the same line in 2 commits in a row. That sounds like a little bikeshedding compared to just doing it this way. It also makes rebasing harder.

Elv13:
Updated

lib/awful/widget/tasklist.lua
psychon: Ok, I'm convinced.

psychon:
Do you want me to look at this closely again?

Elv13:
Up to you, otherwise please put the `pull requests to be merged` tag.

Elv13:
@worron @legostax @warlock90000 @timroes @lcpz @mindeunix Hi, sorry to ping you. This pull request merges the ability to implement custom taglist/tasklist widgets. Given your config seem to either abuse of some "non-feature" or downright forked the whole widget to make it work, you should be made aware that there is now an official API for this kind of modifications. Have a good day/night.

https://awesomewm.org/apidoc/classes/awful.widget.tasklist.html
https://awesomewm.org/apidoc/classes/awful.widget.taglist.html

lcpz:
@Elv13 Never forked & got rid of hacks since 4.1. Anyway, thanks for the ping, I didn't check in a while, so it's nice to see some good work from you, which is always appreciated.

worron:
@Elv13 Thanks for notification, this updated taglist/tasklist design seems very promising adding exact feature I want. Very appreciate this work.

legostax:
@Elv13 Thanks for this feature!  Now I can clean up my code and make it much less 'hacky'.

actionless:
after merging this it seems it's breaking all tasklist and taglist bg_color customizations

Elv13:
> after merging this it seems it's breaking all tasklist and taglist bg_color customizations

Indeed. Fixed and pushed directly to minimize the number of affected users (the patch is a trivial one liner)

actionless:
thanks, now it works

btw, i didn't noticed what things like `beautiful.tasklist_shape_minimized` and so on were already implemented even before this PR, i think i can use those shaping theme options in this new theme: https://github.com/awesomeWM/awesome/pull/2129

psychon:
@Elv13 Your one-liner fix seems wrong. I don't want to figure out the whole hierarchy right now. The `if cache.bgb` shortly before suggests that this was supposed to be optional, but your fix definitely doesn't want `bgb` to be optional. Don't you want to revert that commit and instead change `default_template` to return `primary = bgb` instead of `primary = l`?

Issue #442, opened 2015-09-10
Configuration improvement


silverhammermba:
`rc.lua` is one of the most awesome things about awesome. Just by changing one file you can turn awesome into an almost entirely different window manager. But despite this extreme flexibility, if you browse Linux desktop screenshots online it's generally pretty easy to pick out which ones use awesome. They all have that distinctive wibox with the taglist, tasklist, systray, clock, and a flat color theme. So it appears that even though awesome has the capability to completely change its appearance and behavior, there is a de facto "standard theme" which involves only certain superficial changes to the default `rc.lua`.

The problem with this is that `rc.lua` controls many different aspects of awesome and contains a mix of hardcoded drop-in values (e.g. tag names, menu items) and program logic (e.g. constructing window decorations, defining bind actions). Trying to merge features from someone else's `rc.lua` into your own can be incredibly difficult since they might have custom logic that you need to decipher in order to adapt it or determine it to be unnecessary. Take, for example, [awesome-copycats](https://github.com/copycat-killer/awesome-copycats) which is a bunch of different visual themes for awesome, where _each theme_ includes its own `rc.lua`! Why should changing the visual theme of awesome involve replacing the file that contains fundamental logic like showing errors on startup and binding the same shortcuts to each client? I think this really hurts the ease of customization of awesome by making it extremely difficult to share configs (or parts of configs) in a way that won't break someone else's setup. This problem comes up even for users who don't share configs, because if awesome's API changes in any way, your `rc.lua` will probably break unless you can incorporate the latest changes from the default one.

I think this situation can be improved in several ways:

1. Decide which features of the default `rc.lua` could be considered "standard" and add them to the Lua API so that we can have less logic in `rc.lua`. I think good candidates here are stuff like the error handling code, having common keybindings for tags/clients, having the same layouts per tag, common actions like task switching or viewing a single tag, etc. Basically the kind of stuff that pretty much everyone doesn't change.
2. To go one step further, add methods to the Lua API for easily modifying the config after the fact. So rather than making a copy of the default `rc.lua` and changing a few lines of code, you can just load the default as-is and then make a few more API calls to tweak the parts you don't like. For example, customizing key binds per-client, modifying the main wibox of one of your screens, defining different error handlers, etc. Those who want a more custom experience would just skip loading the default and build it all from scratch.
[...]

PR #2129, opened 2017-11-29
feat(themes; beautiful): add partial support for GTK+3 themes

### LINKED ISSUE #1560: RFC: make awesome look like a current GTK+ 3 theme
I finished the prototype.
Here is the library for reading current GTK+3 theme: https://github.com/actionless/awesome_config/blob/devel/utils/gtk.lua
Here is an awesome wm theme based on that lib (not sure if it will work correctly with the default config):
https://github.com/actionless/awesome_config/blob/devel/themes/gtk/theme.lua

And few screenshots:

Adwaita:
2017-02-12--1486918298_3840x2160_scrot

Breeze:
2017-02-12--1486918449_3840x2160_scrot

Arc-Dark:
2017-02-12--1486918399_3840x2160_scrot

2 themes generated by oomox:
2017-02-12--1486918564_3840x2160_scrot
2017-02-12--1486918510_3840x2160_scrot

actionless:
Closes #1560

### DIFF
--- a/docs/ldoc.css
+++ b/docs/ldoc.css
@@ -191,10 +191,6 @@ table th, table td {
     padding: 2px;
 }
 
-#content p, #content table, #content ol, #content ul, #content dl {
-    max-width: 900px;
-}
-
 #about {
     padding: 15px;
     padding-left: 16em;
--- a/lib/beautiful/gtk.lua
+++ b/lib/beautiful/gtk.lua
@@ -1,0 +1,293 @@
+---------------------------------------------------------------------------
+--- Querying current GTK+ 3 theme via GtkStyleContext.
+--
+-- @author Yauheni Kirylau &lt;yawghen@gmail.com&gt;
+-- @copyright 2016-2017 Yauheni Kirylau
+-- @module beautiful.gtk
+---------------------------------------------------------------------------
+local get_dpi = require("beautiful.xresources").get_dpi
+local gears_debug = require("gears.debug")
+local gears_math = require("gears.math")
+local join = require("gears.table").join
+local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
+
+
+local gtk = {
+    cached_theme_variables = nil
+}
+
+
+local function convert_gtk_channel_to_hex(channel_value)
+    return string.format("%02x", gears_math.round(channel_value * 255))
+end
+
+local function convert_gtk_color_to_hex(gtk_color)
+    return "#" ..
+        convert_gtk_channel_to_hex(gtk_color.red) ..
+        convert_gtk_channel_to_hex(gtk_color.green) ..
+        convert_gtk_channel_to_hex(gtk_color.blue) ..
+        convert_gtk_channel_to_hex(gtk_color.alpha)
+end
+
+local function lookup_gtk_color_to_hex(_style_context, color_name)
+    local gtk_color = _style_context:lookup_color(color_name)
+    if not gtk_color then
+        return nil
+    end
+    return convert_gtk_color_to_hex(gtk_color)
+end
+
+local function get_gtk_property(_style_context, property_name)
+    local state = _style_context:get_state()
+    local property = _style_context:get_property(property_name, state)
+    if not property then
+        return nil
+    end
+    return property.value
+end
+
+local function get_gtk_color_property_to_hex(_style_context, property_name)
+    return convert_gtk_color_to_hex(
+        get_gtk_property(_style_context, property_name)
+    )
+end
+
+local function read_gtk_color_properties_from_widget(gtk_widget, properties)
+    local _style_context = gtk_widget:get_style_context()
+    local result = {}
+    for result_key, style_context_property in pairs(properties) do
+        result[result_key] = get_gtk_color_property_to_hex(
+            _style_context, style_context_property
+        )
+    end
+    return result
+end
+
+
+-- luacheck: max comment line length 300
+
+--- Get GTK+3 theme variables from GtkStyleContext
+-- @treturn table Key-value table with the following structure:
+-- <table class='widget_list' border=1>
+-- <tr style='font-weight: bold;'> <th align='center'>Result key</th> <th align='center'>StyleContext key</th> <th align='center'>StyleContext fallback #1</th> <th align='center'>StyleContext fallback #2</th> <th align='center'>GTK Widget fallback</th> </tr>
+-- <tr> <td>`font_size`</td> <td></td> <td></td> <td></td> <td>Label font-size</td> </tr>
+-- <tr> <td>`font_family`</td> <td></td> <td></td> <td></td> <td>Label font-family</td> </tr>
+-- <tr> <td>`bg_color`</td> <td>`theme_bg_color`</td> <td></td> <td></td> <td>Window bg</td> </tr>
+-- <tr> <td>`fg_color`</td> <td>`theme_fg_color`</td> <td></td> <td></td> <td>Window fg</td> </tr>
+-- <tr> <td>`base_color`</td> <td>`theme_base_color`</td> <td></td> <td></td> <td>Entry bg</td> </tr>
+-- <tr> <td>`text_color`</td> <td>`theme_text_color`</td> <td></td> <td></td> <td>Entry fg</td> </tr>
+-- <tr> <td>`button_bg_color`</td> <td>`theme_button_bg_color`</td> <td>`theme_bg_color`</td> <td></td> <td>Button bg</td> </tr>
+-- <tr> <td>`button_fg_color`</td> <td>`theme_button_fg_color`</td> <td>`theme_fg_color`</td> <td></td> <td>Button fg</td> </tr>
+-- <tr> <td>`button_border_color`</td> <td></td> <td></td> <td></td> <td>Button border-color</td> </tr>
+-- <tr> <td>`button_border_radius`</td> <td></td> <td></td> <td></td> <td>Button border-radius</td> </tr>
+-- <tr> <td>`button_border_width`</td> <td></td> <td></td> <td></td> <td>Button border-top-width</td> </tr>
+-- <tr> <td>`selected_bg_color`</td> <td>`theme_selected_bg_color`</td> <td></td> <td></td> <td>ToggleButton bg</td> </tr>
+-- <tr> <td>`selected_fg_color`</td> <td>`theme_selected_fg_color`</td> <td></td> <td></td> <td>ToggleButton fg</td> </tr>
+-- <tr> <td>`menubar_bg_color`</td> <td>`menubar_bg_color`</td> <td>`theme_bg_color`</td> <td></td> <td>HeaderBar bg</td> </tr>
+-- <tr> <td>`menubar_fg_color`</td> <td>`menubar_fg_color`</td> <td>`theme_fg_color`</td> <td></td> <td>HeaderBar fg</td> </tr>
+-- <tr> <td>`header_button_bg_color`</td> <td>`header_button_bg_color`</td> <td>`menubar_bg_color`</td> <td>`theme_bg_color`</td> <td>HeaderBar > Button bg</td> </tr>
+-- <tr> <td>`header_button_fg_color`</td> <td>`header_button_fg_color`</td> <td>`menubar_fg_color`</td> <td>`theme_fg_color`</td> <td>HeaderBar > Button fg</td> </tr>
+-- <tr> <td>`header_button_border_color`</td> <td></td> <td></td> <td></td> <td>HeaderBar > Button border-color</td> </tr>
+-- <tr> <td>`error_color`</td> <td>`error_color`</td> <td>`error_bg_color`</td> <td></td> <td>destructive Button bg</td> </tr>
+-- <tr> <td>`error_bg_color`</td> <td>`error_bg_color`</td> <td>`error_color`</td> <td></td> <td>destructive Button bg</td> </tr>
+-- <tr> <td>`error_fg_color`</td> <td>`error_fg_color`</td> <td>`theme_selected_fg_color`</td> <td></td> <td>destructive Button fg</td> </tr>
+-- <tr> <td>`warning_color`</td> <td>`warning_color`</td> <td>`warning_bg_color`</td> <td></td> <td></td> </tr>
+-- <tr> <td>`warning_bg_color`</td> <td>`warning_bg_color`</td> <td>`warning_color`</td> <td></td> <td></td> </tr>
+-- <tr> <td>`warning_fg_color`</td> <td>`warning_fg_color`</td> <td>`theme_selected_fg_color`</td> <td></td> <td></td> </tr>
+-- <tr> <td>`success_color`</td> <td>`success_color`</td> <td>`success_bg_color`</td> <td></td> <td></td> </tr>
+-- <tr> <td>`success_bg_color`</td> <td>`success_bg_color`</td> <td>`success_color`</td> <td></td> <td></td> </tr>
+-- <tr> <td>`success_fg_color`</td> <td>`success_fg_color`</td> <td>`theme_selected_fg_color`</td> <td></td> <td></td> </tr>
+-- <tr> <td>`tooltip_bg_color`</td> <td>`theme_tooltip_bg_color`</td> <td>`theme_bg_color`</td> <td></td> <td></td> </tr>
+-- <tr> <td>`tooltip_fg_color`</td> <td>`theme_tooltip_fg_color`</td> <td>`theme_fg_color`</td> <td></td> <td></td> </tr>
+-- <tr> <td>`osd_bg_color`</td> <td>`osd_bg`</td> <td>`theme_tooltip_bg_color`</td> <td>`theme_bg_color`</td> <td></td> </tr>
+-- <tr> <td>`osd_fg_color`</td> <td>`osd_fg`</td> <td>`theme_tooltip_fg_color`</td> <td>`theme_fg_color`</td> <td></td> </tr>
+-- <tr> <td>`osd_border_color`</td> <td>`osd_borders_color`</td> <td>`osd_fg_color`</td> <td></td> <td></td> </tr>
+-- <tr> <td>`wm_bg_color`</td> <td>`wm_bg`</td> <td>`menubar_bg_color`</td> <td>`theme_bg_color`</td> <td>HeaderBar bg</td> </tr>
+-- <tr> <td>`wm_border_focused_color`</td> <td>`wm_border_focused`</td> <td>`theme_selected_bg_color`</td> <td></td> <td>ToggleButton bg</td> </tr>
+-- <tr> <td>`wm_border_unfocused_color`</td> <td>`wm_border_unfocused`</td> <td>`wm_border`</td> <td>`menubar_bg_color`</td> <!--<td>`theme_bg_color`</td>--> <td>HeaderBar bg</td> </tr>
+-- <tr> <td>`wm_title_focused_color`</td> <td>`wm_title_focused`</td> <td>`wm_title`</td> <td>`theme_selected_fg_color`</td> <td>ToggleButton fg</td> </tr>
+-- <tr> <td>`wm_title_unfocused_color`</td> <td>`wm_title_unfocused`</td> <td>`wm_unfocused_title`</td> <td>`menubar_fg_color`</td> <!--<td>`theme_fg_color`</td>--> <td>HeaderBar fg</td> </tr>
+-- <tr> <td>`wm_icons_focused_color`</td> <td>`wm_icons_focused`</td> <td>`wm_title_focused`</td> <td>`theme_selected_fg_color`</td> <td>ToggleButton fg</td> </tr>
+-- <tr> <td>`wm_icons_unfocused_color`</td> <td>`wm_icons_unfocused`</td> <td>`wm_title_unfocused`</td> <td>`menubar_fg_color`</td> <!--<td>`theme_fg_color`</td>--> <td>HeaderBar fg</td> </tr>
+-- </table>
+--
+function gtk.get_theme_variables()
+    if gtk.cached_theme_variables then
+        return gtk.cached_theme_variables
+    end
+
+    local result = {}
+    local _gtk_status, Gtk = pcall(function()
+        return require('lgi').Gtk
+    end)
+    if not _gtk_status or not Gtk then
+        gears_debug.print_warning(
+            "Can't load GTK+3 introspection. "..
+            "Seems like GTK+3 is not installed or `lua-lgi` was built with an incompatible GTK+3 version."
+        )
+        return nil
+    end
+    local _window_status, window = pcall(function()
+        return Gtk.Window.new(Gtk.WindowType.TOPLEVEL)
+    end)
+    if not _window_status or not window then
+        gears_debug.print_warning(
+            "Can't create GTK+3 window. "..
+            "Seems like GTK+3 theme is not set correctly or `lua-lgi` was built with an incompatible GTK+3 version."
+        )
+        return nil
+    end
+    local style_context = window:get_style_context()
+
+    result = join(result, read_gtk_color_properties_from_widget(
+        window, {
+            bg_color="background-color",
+            fg_color="color",
+        }
+    ))
+    local label = Gtk.Label()
+    local label_style_context = label:get_style_context()
+    local gdk_scale = tonumber(os.getenv("GDK_SCALE") or 1.0)
+    local xrdb_scale = get_dpi() / 96
+    local pt_to_px_ratio = 1+1/3
+    result["font_size"] = get_gtk_property(
+        label_style_context, "font-size"
+    ) * gdk_scale / xrdb_scale / pt_to_px_ratio
+    result["font_family"] = get_gtk_property(
+        label_style_context, "font-family"
+    )[1]
+
+    result = join(result, read_gtk_color_properties_from_widget(
+        Gtk.Entry(), {
+            base_color="background-color",
+            text_color="color",
+        }
+    ))
+    result = join(result, read_gtk_color_properties_from_widget(
+        Gtk.ToggleButton(), {
+            selected_bg_color="background-color",
+            selected_fg_color="color",
+        }
+    ))
+    local button = Gtk.Button()
+    result = join(result, read_gtk_color_properties_from_widget(
+        button, {
+            button_bg_color="background-color",
+            button_fg_color="color",
+            button_border_color="border-color",
+        }
+    ))
+    local button_style_context = button:get_style_context()
+    for result_key, style_context_property in pairs({
+        button_border_radius="border-radius",
+        button_border_width="border-top-width",
+    }) do
+        local state = style_context:get_state()
+        local property = button_style_context:get_property(style_context_property, state)
+        result[result_key] = property.value
+        property:unset()
+    end
+
+    local headerbar = Gtk.HeaderBar()
+    window:set_titlebar(headerbar)
+    result = join(result, read_gtk_color_properties_from_widget(
+        headerbar, {
+            menubar_bg_color="background-color",
+        }
+    ))
+
+    if result.menubar_bg_color and result.menubar_bg_color ~= "#00000000" then
+        headerbar:add(label)
+        result = join(result, read_gtk_color_properties_from_widget(
+            label, {
+                menubar_fg_color="color",
+            }
+        ))
+    end
+
+    headerbar:add(button)
+    result = join(result, read_gtk_color_properties_from_widget(
+        button, {
+            header_button_bg_color="background-color",
+            header_button_border_color="border-color",
+        }
+    ))
+    if result.header_button_bg_color and result.header_button_bg_color ~= "#00000000" then
+        result = join(result, read_gtk_color_properties_from_widget(
+            button, {
+                header_button_fg_color="color",
+            }
+        ))
+    end
+
+    local error_button = Gtk.Button()
+    error_button:get_style_context():add_class("destructive-action")
+    result = join(result, read_gtk_color_properties_from_widget(
+        error_button, {
+            error_bg_color="background-color",
+            error_fg_color="color",
+        }
+    ))
+
+    for _, color_data in ipairs({
+        {"bg_color", "theme_bg_color"},
+        {"fg_color", "theme_fg_color"},
+        {"base_color", "theme_base_color"},
+        {"text_color", "theme_text_color"},
+        {"selected_bg_color", "theme_selected_bg_color"},
+        {"selected_fg_color", "theme_selected_fg_color"},
+        --
+        {"tooltip_bg_color", "theme_tooltip_bg_color", "bg_color"},
+        {"tooltip_fg_color", "theme_tooltip_fg_color", "fg_color"},
+        {"osd_bg_color", "osd_bg", "tooltip_bg_color"},
+        {"osd_fg_color", "osd_fg", "tooltip_fg_color"},
+        {"osd_border_color", "osd_borders_color", "osd_fg_color"},
+        {"menubar_bg_color", "menubar_bg_color", "bg_color"},
+        {"menubar_fg_color", "menubar_fg_color", "fg_color"},
+        --
+        {"button_bg_color", "button_bg_color", "bg_color"},
+        {"button_fg_color", "button_fg_color", "fg_color"},
+        {"header_button_bg_color", "header_button_bg_color", "menubar_bg_color"},
+        {"header_button_fg_color", "header_button_fg_color", "menubar_fg_color"},
+        --
+        {"wm_bg_color", "wm_bg", "menubar_bg_color"},
+        {"wm_border_focused_color", "wm_border_focused", "selected_bg_color"},
+        {"wm_border_unfocused_color", "wm_border_unfocused", "wm_border", "menubar_bg_color"},
+        {"wm_title_focused_color", "wm_title_focused", "wm_title", "selected_fg_color"},
+        {"wm_title_unfocused_color", "wm_title_unfocused", "wm_unfocused_title", "menubar_fg_color"},
+        {"wm_icons_focused_color", "wm_icons_focused", "wm_title_focused_color", "selected_fg_color"},
+        {"wm_icons_unfocused_color", "wm_icons_unfocused", "wm_title_unfocused_color", "menubar_fg_color"},
+        --
+        {"error_color", "error_color"},
+        {"error_bg_color", "error_bg_color", "error_color"},
+        {"error_fg_color", "error_fg_color", "selected_fg_color"},
+        {"error_color", "error_color", "error_bg_color"},
+        {"warning_color", "warning_color"},
+        {"warning_bg_color", "warning_bg_color", "warning_color"},
+        {"warning_fg_color", "warning_fg_color", "selected_fg_color"},
+        {"warning_color", "warning_color", "warning_bg_color"},
+        {"success_color", "success_color"},
+        {"success_bg_color", "success_bg_color", "success_color"},
+        {"success_fg_color", "success_fg_color", "selected_fg_color"},
+        {"success_color", "success_color", "success_bg_color"},
+    }) do
+        local result_key, style_context_key, fallback_key, fallback_key2 = unpack(color_data)
+        result[result_key] = lookup_gtk_color_to_hex(style_context, style_context_key) or (
+            result[result_key] ~= "#00000000" and
+            result[result_key] or
+            result[fallback_key] or
+            result[fallback_key2] or
+            result[result_key]    -- <-- here is for case if it was meant to be a fully transparent color on purpose
+        )
+        if not result[result_key] then
+            gears_debug.print_warning("Can't read color '" .. style_context_key .. "' from GTK+3 theme.")
+        end
+    end
+
+    window:destroy()
+    gtk.cached_theme_variables = result
+    return result
+end
+
+
+return gtk
--- a/themes/gtk/theme.lua
+++ b/themes/gtk/theme.lua
@@ -1,0 +1,350 @@
+----------------------------------------------
+-- Awesome theme which follows GTK+ 3 theme --
+--   by Yauhen Kirylau                      --
+----------------------------------------------
+
+local theme_assets = require("beautiful.theme_assets")
+local dpi = require("beautiful.xresources").apply_dpi
+local gfs = require("gears.filesystem")
+local themes_path = gfs.get_themes_dir()
+local gears_shape = require("gears.shape")
+local wibox = require("wibox")
+local awful_widget_clienticon = require("awful.widget.clienticon")
+local gtk = require("beautiful.gtk")
+
+
+-- Helper functions for modifying hex colors:
+--
+local hex_color_match = "[a-fA-F0-9][a-fA-F0-9]"
+local function darker(color_value, darker_n)
+    local result = "#"
+    local channel_counter = 1
+    for s in color_value:gmatch(hex_color_match) do
+        local bg_numeric_value = tonumber("0x"..s)
+        if channel_counter <= 3 then
+            bg_numeric_value = bg_numeric_value - darker_n
+        end
+        if bg_numeric_value < 0 then bg_numeric_value = 0 end
+        if bg_numeric_value > 255 then bg_numeric_value = 255 end
+        result = result .. string.format("%02x", bg_numeric_value)
+        channel_counter = channel_counter + 1
+    end
+    return result
+end
+local function is_dark(color_value)
+    local bg_numeric_value = 0;
+    local channel_counter = 1
+    for s in color_value:gmatch(hex_color_match) do
+        bg_numeric_value = bg_numeric_value + tonumber("0x"..s);
+        if channel_counter == 3 then
+            break
+        end
+        channel_counter = channel_counter + 1
+    end
+    local is_dark_bg = (bg_numeric_value < 383)
+    return is_dark_bg
+end
+local function mix(color1, color2, ratio)
+    ratio = ratio or 0.5
+    local result = "#"
+    local channels1 = color1:gmatch(hex_color_match)
+    local channels2 = color2:gmatch(hex_color_match)
+    for _ = 1,3 do
+        local bg_numeric_value = math.ceil(
+          tonumber("0x"..channels1())*ratio +
+          tonumber("0x"..channels2())*(1-ratio)
+        )
+        if bg_numeric_value < 0 then bg_numeric_value = 0 end
+        if bg_numeric_value > 255 then bg_numeric_value = 255 end
+        result = result .. string.format("%02x", bg_numeric_value)
+    end
+    return result
+end
+local function reduce_contrast(color, ratio)
+    ratio = ratio or 50
+    return darker(color, is_dark(color) and -ratio or ratio)
+end
+
+local function choose_contrast_color(reference, candidate1, candidate2)  -- luacheck: no unused
+    if is_dark(reference) then
+        if not is_dark(candidate1) then
+            return candidate1
+        else
+            return candidate2
+        end
+    else
+        if is_dark(candidate1) then
+            return candidate1
+        else
+            return candidate2
+        end
+    end
+end
+
+
+-- inherit xresources theme:
+local theme = dofile(themes_path.."xresources/theme.lua")
+-- load and prepare for use gtk theme:
+theme.gtk = gtk.get_theme_variables()
+if not theme.gtk then
+    local gears_debug = require("gears.debug")
+    gears_debug.print_warning("Can't load GTK+3 theme. Using 'xresources' theme as a fallback.")
+    return theme
+end
+theme.gtk.button_border_radius = dpi(theme.gtk.button_border_radius or 0)
+theme.gtk.button_border_width = dpi(theme.gtk.button_border_width or 1)
+theme.gtk.bold_font = theme.gtk.font_family .. ' Bold ' .. theme.gtk.font_size
+theme.gtk.menubar_border_color = mix(
+    theme.gtk.menubar_bg_color,
+    theme.gtk.menubar_fg_color,
+    0.7
+)
+
+
+theme.font          = theme.gtk.font_family .. ' ' .. theme.gtk.font_size
+
+theme.bg_normal     = theme.gtk.bg_color
+theme.fg_normal     = theme.gtk.fg_color
+
+theme.wibar_bg      = theme.gtk.menubar_bg_color
+theme.wibar_fg      = theme.gtk.menubar_fg_color
+
+theme.bg_focus      = theme.gtk.selected_bg_color
+theme.fg_focus      = theme.gtk.selected_fg_color
+
+theme.bg_urgent     = theme.gtk.error_bg_color
+theme.fg_urgent     = theme.gtk.error_fg_color
+
+theme.bg_minimize   = mix(theme.wibar_fg, theme.wibar_bg, 0.3)
+theme.fg_minimize   = mix(theme.wibar_fg, theme.wibar_bg, 0.9)
+
+theme.bg_systray    = theme.wibar_bg
+
+theme.border_normal = theme.gtk.wm_border_unfocused_color
+theme.border_focus  = theme.gtk.wm_border_focused_color
+theme.border_marked = theme.gtk.success_color
+
+theme.border_width  = dpi(theme.gtk.button_border_width or 1)
+theme.border_radius = theme.gtk.button_border_radius
+
+theme.useless_gap   = dpi(3)
+
+local rounded_rect_shape = function(cr,w,h)
+    gears_shape.rounded_rect(
+        cr, w, h, theme.border_radius
+    )
+end
+
+-- There are other variable sets
+-- overriding the default one when
+-- defined, the sets are:
+-- taglist_[bg|fg|shape|shape_border_color|shape_border_width]_[focus|urgent|occupied|empty|volatile]
+-- tasklist_[bg|fg|shape|shape_border_color|shape_border_width]_[focus|urgent|minimized]
+-- titlebar_[bg|fg]_[normal|focus]
+-- tooltip_[font|opacity|fg_color|bg_color|border_width|border_color]
+-- mouse_finder_[color|timeout|animate_timeout|radius|factor]
+
+theme.tasklist_fg_normal = theme.wibar_fg
+theme.tasklist_bg_normal = theme.wibar_bg
+theme.tasklist_fg_focus = theme.tasklist_fg_normal
+theme.tasklist_bg_focus = theme.tasklist_bg_normal
+
+theme.tasklist_font_focus = theme.gtk.bold_font
+
+theme.tasklist_shape_minimized = rounded_rect_shape
+theme.tasklist_shape_border_color_minimized = mix(
+    theme.bg_minimize,
+    theme.fg_minimize,
+    0.85
+)
+theme.tasklist_shape_border_width_minimized = theme.gtk.button_border_width
+
+theme.tasklist_spacing = theme.gtk.button_border_width
+
+--[[ Advanced taglist and tasklist styling: {{{
+
+--- In order to get taglist and tasklist to follow GTK theme you need to
+-- modify your rc.lua in the following way:
+
+diff --git a/rc.lua b/rc.lua
+index 231a2f68c..533a859d2 100644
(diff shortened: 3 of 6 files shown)


actionless:
weird, locally test is passing however in travis `style_context:get_font()` and `:get_property()` are working as expected but not `:lookup_color()`

psychon:
```
[...]
awesome: lvm.c:436: luaV_equalobj: Assertion `(((t1)->value_).gc)->tt == 5' failed.
2017-11-29 21:41:24 E: awesome: signal_fatal:448: signal 6, dumping backtrace
```
Uhm... I've seen that error before recently, on some other PR. Ah, https://github.com/awesomeWM/awesome/pull/2119#issuecomment-346743140
So... yay, we got something randomly failing again. :-(

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,178 @@
...
+
+
+local function lookup_gtk_color_to_hex(_style_context, color_name)
+  local color = _style_context:lookup_color(color_name)
+  if not color then
+      return nil
+  end
+  local result = convert_gtk_color_to_hex(color)
+  --color:free()
+  return result
```
psychon: I see a trivial way to simplify the above three lines...

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,178 @@
...
+end
+
+local function get_gtk_color_property_to_hex(_style_context, property_name, state)
+    local property = _style_context:get_property(property_name, state)
```
psychon: I don't really know this API, but the documentation says:
> Note that passing a state other than the current state of context is not recommended unless the style context has been saved with gtk_style_context_save().

https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#gtk-style-context-get-property

Does that matter here? Why not?

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,178 @@
...
+    if not property then
+      return nil
+    end
+    local result = convert_gtk_color_to_hex(property.value)
+    property:unset()
+    return result
```
psychon: First: I think that `:unset` is unnecessary and pointless.
Second: Wait, what is the difference between this function and the function before? They seem to do the same thing to me, except that this one seems much more scare when looking at the API docs for `GValue`.
Are these two difference ways really necessary? That sounds a lot like this code tries to do something that should not be done.

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,178 @@
...
+  local lgi = require('lgi')
+  local Gtk = lgi.Gtk
+  local window
+  pcall(function()
+    window = Gtk.Window{
+      --on_destroy = Gtk.main_quit,
+    }
+  end)
+  if not window then
+    gears_debug.print_warning("Seems like GTK+3 is not installed or theme is not set correctly.")
```
psychon: 1) How about actually mentioning the error message from the `pcall`? (You could just use `gears.protected_call` instead)
2) The error message is wrong. If Gtk is not installed, `lgi.Gtk` will already throw an error. Also, I have a hard time seeing "I failed to create a GtkWindow" being equivalent to "theme is not set correctly".

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,178 @@
...
+    return nil
+  end
+  --window:set_override_redirect(true)
+  local style_context = window:get_style_context()
+
+  local font = style_context:get_font(Gtk.StateFlags.NORMAL)
```
psychon: This function is deprecated: https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#gtk-style-context-get-font
(And no, I don't understand the replacement nor how to use it)

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,178 @@
...
+  --result.font_family = font:get_family()
+  --result.font_size = font:get_size()/1024
```
psychon: There seems to be a lot of "you should clean this up" around here...

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,178 @@
...
+  end
+  for result_key, style_context_property in pairs({
+    base_color="background-color",
+    text_color="color",
+  }) do
+    result[result_key] = get_gtk_color_property_to_hex(
+        entry_style_context, style_context_property, Gtk.StateFlags.NORMAL
+    )
+  end
+  entry:destroy()
```
psychon: I think the docs for this module should say what this actually does... So the `bg/fg_color`s come from the background-color and color of a GtkWindow, the border width is the border-top-width of a GtkEntry and the base_color and text_color come from the background and foreground color of a GtkEntry?!?

psychon: Also, I doubt that this `entry:destroy()` is necessary. And from the docs for it... I doubt this even more, I think this does not actually do anything (besides setting a `destroyed` flag).
https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-destroy

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,178 @@
...
+  end
+  for result_key, style_context_property in pairs({
+    button_bg_color="background-color",
+    button_fg_color="color",
+  }) do
+    result[result_key] = get_gtk_color_property_to_hex(
+        button_style_context, style_context_property, Gtk.StateFlags.NORMAL
+    )
+  end
+  button:destroy()
```
psychon: Can the above code be made more readable by e.g. moving it into a table?
```lua
{
object = Gtk.Button(),
properties = { 
  Gtk.StateFlags.NORMAL = {
    border_radius = "border-radiues",
    etc
  }
}
```
That way, all that code is only needed once instead of copy&pasted three times.

psychon:
Perhaps this makes more sense as a self-contained tool? It does all the magic that is done here and then writes (either on stdout or into a file) a theme for awesome? That way people can hand-edit the result if they want.

lib/beautiful/gtk.lua
actionless: it's actually related to your other coment regarding deprecation of `:get_font()`

lib/beautiful/gtk.lua
actionless: because properties are the real properties of object, while stylecontext colors are just variables which are defined inside stylecontext.

however if theme is using pixmaps or gradient the properties will be not very useful (especially in case of pixmaps) so here come stylecontext

lib/beautiful/gtk.lua
actionless: right, if you take a look on gtk3-widget-factory (or any other gtk3 app) you will see what stylecontext values bg/fg are used for windows, text/base - for entries

however if the same values are defined in stylecontext, stylecontext have a higher prority (because of pixmaps/gradients issue described in the previous comment)

themes/gtk/theme.lua
```diff
@@ -0,0 +1,173 @@
...
+
+theme.xrdb = xresources.get_current_theme()
+theme.gtk = gtk.get_theme_variables()
+
+
+theme.font          = theme.gtk.font
+
+theme.bg_normal     = theme.gtk.menubar_bg_color
+theme.bg_focus      = theme.gtk.selected_bg_color
+theme.bg_urgent     = theme.xrdb.color9
```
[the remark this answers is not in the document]
actionless: so far it's taking from xrdb only colors for errors, i can also try to render some kind of GtkButton with "Destroyable" role (ie red one) and use color from it

themes/gtk/theme.lua
```diff
@@ -0,0 +1,173 @@
...
+
+-- Generate wallpaper:
+local wallpaper_shade = 50
+local wallpaper_bg = theme.gtk.base_color
+local wallpaper_fg = theme.gtk.bg_color
+local wallpaper_alt_fg = theme.gtk.selected_bg_color
+if not is_dark(theme.bg_normal) then
+    wallpaper_bg, wallpaper_fg = wallpaper_fg, wallpaper_bg
+end
+wallpaper_bg = darker(wallpaper_bg, is_dark(wallpaper_bg) and -wallpaper_shade or wallpaper_shade)
```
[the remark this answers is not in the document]
actionless: it works only with hex colors and refactoring it to support all possible types of gears.color could be too much effort

themes/gtk/theme.lua
```diff
@@ -0,0 +1,173 @@
...
+wallpaper_fg = darker(wallpaper_fg, is_dark(wallpaper_fg) and -wallpaper_shade or wallpaper_shade)
+wallpaper_fg = mix(wallpaper_fg, wallpaper_bg, 0.4)
```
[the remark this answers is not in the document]
actionless: yup, that all magic around wallpaper is actually to generate some sort of palette based on current gtk colors, but being less contrast (so windows won't look on the same surface as desktop wallpaper)

actionless:
actually if you take a look on chrome or firefox, they handling gtk+3 "support" in literally the same way as it done here, i really see no reason why they can "just take some random colors from gtk context and mix them randomly" while awesome wm can't and need to force users who want seamless gtk3 experience to install some extra modules

themes/gtk/theme.lua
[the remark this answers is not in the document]
psychon: Sorry, I meant "helper function in here", similar to your `darken` function. It's just meant to improve readability of the code around here.

psychon:
> actually if you take a look on chrome or firefox, they handling gtk+3 "support" in literally the same way as it done here,

Since those are quite some big code bases: Got a link to that code? I'd try to find that via debian codesearch, but perhaps you know something more direct here?

actionless:
i think in apidoc i will also add a table of all return values with some description, it should address some of the comments above

actionless:
> Since those are quite some big code bases: Got a link to that code? I'd try to find that via debian codesearch, but perhaps you know something more direct here?

i'll try to find some examples tonight, but basically you can grep -i by `stylecontext`

psychon:
Debian codesearch finds "stylecontext" in the following packages: libreoffice, klavaro, firefox, firefox-esr, blueman, gtk-d, libgtksourceviewmm, gtkatlantic, gnome-applets, gtk+3.0, blhc, meld, goocanvasmm-2.0, mate-applets, w1retap, ruby-gnome2, synapse, scribus, nemo-fileroller, icedove

The list of matches for firefox does not look all that interesting:
https://codesearch.debian.net/search?q=GtkStyleContext+package%3Afirefox

And some of the code seems to use Gtk in the way it is intended to be used for drawing: http://sources.debian.net/src/firefox/55.0-2/widget/gtk/gtk3drawing.cpp/#L781

actionless:
@psychon one example of black magic similar to `beautiful/gtk.lua`:
https://github.com/mozilla/gecko-dev/blob/dbf72d85a448c6a570d70a9bed23a689c0d4abb4/widget/gtk/nsLookAndFeel.cpp#L132-L165

and this similar to things in `themes/gtk.lua`:
https://github.com/mozilla/gecko-dev/blob/dbf72d85a448c6a570d70a9bed23a689c0d4abb4/widget/gtk/nsLookAndFeel.cpp#L238-L434

psychon:
Some random findings from those links that might be relevant to this PR:
https://github.com/mozilla/gecko-dev/blob/dbf72d85a448c6a570d70a9bed23a689c0d4abb4/widget/gtk/nsLookAndFeel.cpp#L812-L815
```
    // Gtk manages a screen's CSS in the settings object so we
    // ask Gtk to create it explicitly. Otherwise we may end up
    // with wrong color theme, see Bug 972382
GtkSettings *settings = gtk_settings_get_for_screen(gdk_screen_get_default());
```

Detection of dark themes:
https://github.com/mozilla/gecko-dev/blob/dbf72d85a448c6a570d70a9bed23a689c0d4abb4/widget/gtk/nsLookAndFeel.cpp#L817-L840


Allow to override the theme name:
https://github.com/mozilla/gecko-dev/blob/dbf72d85a448c6a570d70a9bed23a689c0d4abb4/widget/gtk/nsLookAndFeel.cpp#L842-L851

How Firefox looks up style contexts (different code paths for before/after Gtk 3.20.0!):
https://github.com/mozilla/gecko-dev/blob/dbf72d85a448c6a570d70a9bed23a689c0d4abb4/widget/gtk/WidgetStyleCache.cpp#L1293

Function used for looking up colors (well, background color; there are lots of other functions in there):
https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#gtk-style-context-get-background-color

Okay, starting around line 919, that code also creates lots of widgets and queries styles from them, but in a different way than your code does, I think (e.g. `gtk_style_context_get_color`, the style context is created/queries differently).

Basically, firefox seems to have a similar, but different, mess. Dunno what to think of this...

@Elv13 @blueyed ?

actionless:
thanks for finding the code samples, i'll inspect them tonight and mb they will help me to fix some things (however i was doing manual test with a number of themes and it was fine, but i was testing only on GTK 3.14 and 3.22)

as a side note: too bad what gtk doesn't have such kind of method like 'get_systempalette' or strict rules for themes to have at least some basic colors defined in the stylecontext

lib/beautiful/gtk.lua
actionless: 1) i didn't understood that part
2) also i was able to reproduce that problem after updating gtk but before rebuilding lua-lgi

lib/beautiful/gtk.lua
actionless: also in the docs it's recommended to unset it[1]:

```
When value is no longer needed, g_value_unset() must be called to free any allocated memory.
```

[1] https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#gtk-style-context-get-property

themes/gtk/theme.lua
```diff
@@ -0,0 +1,173 @@
...
+    taglist_square_size, theme.fg_normal
+)
+
+-- Generate wallpaper:
+local wallpaper_shade = 50
+local wallpaper_bg = theme.gtk.base_color
+local wallpaper_fg = theme.gtk.bg_color
+local wallpaper_alt_fg = theme.gtk.selected_bg_color
+if not is_dark(theme.bg_normal) then
+    wallpaper_bg, wallpaper_fg = wallpaper_fg, wallpaper_bg
```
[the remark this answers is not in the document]
actionless: because we can't know if base_color or bg_color will be lighter

themes/gtk/theme.lua
```diff
@@ -0,0 +1,173 @@
...
+theme.gtk = gtk.get_theme_variables()
+
+
+theme.font          = theme.gtk.font
+
+theme.bg_normal     = theme.gtk.menubar_bg_color
+theme.bg_focus      = theme.gtk.selected_bg_color
+theme.bg_urgent     = theme.xrdb.color9
+theme.bg_minimize   = theme.gtk.wm_border_unfocused_color
+theme.bg_systray    = theme.bg_normal
```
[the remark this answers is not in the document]
actionless: i think it's still leaving for user a lot of opportunity to customize, in this particular example -- to choose for systray gtk.bg_color, gtk.base_color or gtk.menubar_bg_color, etc

actionless:
squashed to pass the CI

lib/beautiful/gtk.lua
psychon: Well, yeah, in C code. Since Lua is supposed to be garbage collected, I would expect that lgi does the call to `:unset()` for us. From a quick `git grep unset` through lgi, I would guess that line 54 in `lgi/override/GObject-Value.lua` sets this up: `Value._uninit = core.record.value_unset`
Since colors are not huge objects whose size is invisible to the GC (in contrast to some of the cairo surface that causes us lots of problems and made us add calls to `:finish()` in various places), I really think these `:unset()`-calls can be dropped.

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,169 @@
...
+  local window
+  pcall(function()
+    window = Gtk.Window{
+      --on_destroy = Gtk.main_quit,
+    }
+  end)
+  if not window then
+    gears_debug.print_warning("Seems like GTK+3 theme is not set correctly or luajit was built with an incomptible GTK+3 version.")
+    return nil
+  end
```
psychon: How about replacing the above with (since your answer to this comment last time said you did not understand what I meant):
```lua
local lgi = require("lgi")
local Gtk, window = require("gears.protected_call").call(function()
    local Gtk = lgi.Gtk
    return Gtk, Gtk.Window{}
end)
if not Gtk then
    gears.debug.print_warning("Failed to load GTK and instantiate a GtkWindow.")
    return nil
end
```
(Can/should the `require`s be moved to the beginning of the file?)
(No, I will not ask since when luajit depends against Gtk in some way)

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,169 @@
...
+  --window:set_override_redirect(true)
```
psychon: That comment can go, I think

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,169 @@
...
+      selected_fg_color="color",
+    }
+  ))
+  local button = Gtk.Button()
+  result = join(result, read_gtk_color_properties_from_widget(
+    button, {
+      button_bg_color="background-color",
+      button_fg_color="color",
+    }
+  ))
```
psychon: Thanks, I find it a lot easier to figure out what the above code does than in the old version

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,169 @@
...
+    {"wm_border_focused_color", "wm_border_focused", "selected_bg_color"},
+    {"wm_title_focused_color", "wm_title_focused", "menubar_bg_color"},
+    {"wm_icons_focused_color", "wm_icons_focused", "menubar_fg_color"},
+    {"wm_border_unfocused_color", "wm_border_unfocused", "wm_bg_color"},
+    {"wm_title_unfocused_color", "wm_title_unfocused", "menubar_bg_color"},
+    {"wm_icons_unfocused_color", "wm_icons_unfocused", "menubar_fg_color"},
+  }) do
+    local result_key, style_context_key, fallback_key = unpack(color_data)
+    result[result_key] = lookup_gtk_color_to_hex(style_context, style_context_key) or
+      (result[result_key] ~= "#000000" and result[result_key] or result[fallback_key] or result[result_key])
```
psychon: Uhm, replacing the actual values in the expression above with `a`, `b` etc, the above does
```
a = b or (a ~= "#000000" and a or c or a)
```
I don't know what is so special about black, so perhaps a comment should be added to explain this (*). Anyway, I want to try to simplify the expression above. Is the following easier to read? Is it equivalent to the original expression? (I'm not quite sure, which is part of why I think it should be simplified):
```
a = b or (a == "#000000" and c) or a
```
Well, no... the above is different when `result[result_key]` (aka `a`) is `nil` to begin with...
How about
```lua
local value
value = lookup_gtk_color_to_hex(style_context, style_context_key)
value = value or (result[result_key] ~= "#000000" and result[result_key])
value = value or result[fallback_key]
if value then
    result[result_key] = value
end
```
Again, I am again not sure if the above is equivalent or better... Let's wait for what @Elv13 and @blueyed say.

(*): If this black happens to be the result of `read_gtk_color_properties_from_widget` when the color property is not found: Why is that code so careful in checking for `nil` all the time?

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,169 @@
...
+    if not result[result_key] then
+        gears_debug.print_warning("Can't read color '" .. style_context_key .. "' from GTK+3 theme.")
+    end
+  end
+
+  window:destroy()
```
psychon: Is this `:destroy()` necessary? If not, I'd drop it.

actionless: i think it does the same as closing the window (if it would be visible)

lib/beautiful/gtk.lua
actionless: null color property returns #000000, null stylecontext returns null

i can rewrite it in multiline way if it's easier to read

lib/beautiful/gtk.lua
actionless: >  since when luajit depends against Gtk in some way
i was very sleepy when writing that, i meant `lua-lgi`, not `luajit`

i thought not to move requires to the beginning of file in order not to fire them unless function is called explicitly

lib/beautiful/gtk.lua
psychon: > null color property returns #000000, null stylecontext returns null

What exactly do you mean here? How/where would I get a null color property or a null stylecontext? I tried the following and neither gave me `#000000`, but instead `nil`:
```
lua -e 'gtk = require("lgi").Gtk sc = gtk.Window{}:get_style_context() print(sc:get_property("foo", sc:get_state()).value)
lua -e 'gtk = require("lgi").Gtk sc = gtk.Window{}:get_style_context() print(sc:lookup_color("foo"))'
```

actionless: in your first example you're trying to retrieve `non-existing` property, not proverty with `null` value

i was able to reproduce it with `button`'s `background-color` in the themes where no background-color is set but `background: <gradient>` or just a pixmap

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,171 @@
...
+local function convert_gtk_channel_to_hex(channel_value)
+    return string.format("%02x", gears_math.round(channel_value * 255))
+end
+
+local function convert_gtk_color_to_hex(gtk_color)
+  return "#" ..
+    convert_gtk_channel_to_hex(gtk_color.red) ..
+    convert_gtk_channel_to_hex(gtk_color.green) ..
+    convert_gtk_channel_to_hex(gtk_color.blue) --..
+  --convert_gtk_channel_to_hex(gtk_color.alpha)
```
actionless: btw  pixmap or i commented it out because tasklist and taglist were not working properly with such colors, i'll try mb after https://github.com/awesomeWM/awesome/pull/2127 is merged it will work fine

lib/beautiful/gtk.lua
psychon: Ah, okay. I haven't found were this is set in the code, but I found it in the docs: https://developer.gnome.org/gtk3/stable/chap-css-properties.html#id-1.5.3.3.34
The initial value for `background-color` is `rgba(0, 0, 0, 0)`. And apparently nothing else has this default initial value. This means two things:
- This default value should only be ignored for `background-color`, not for all fields.
- This default value likely should not be ignored at all...? Since apparently that is indeed the color that will be used by GTK here, no?

Oh, okay, I already found it in the code: https://sources.debian.org/src/gtk+3.0/3.22.26-2/gtk/gtkcssstylepropertyimpl.c/#L1114

Oh and (sorry, unrelated): There seems to be a `font-size` property: https://sources.debian.org/src/gtk+3.0/3.22.26-2/gtk/gtkcssstylepropertyimpl.c/#L1075

This all still "feels" weird....

lib/beautiful/gtk.lua
psychon: Uhm, which colors do you mean exactly? The background color uses a `wibox.container.background`, but the foreground color uses Pango markup (`<span font_desc='foo'><span color='bar'>etc`) and Pango... uhm, should support alpha, too, according to https://developer.gnome.org/pango/unstable/PangoMarkupFormat.html ...? This seems like something that we should text&fix.

lib/beautiful/gtk.lua
actionless: actually after uncommenting `.alpha` part of code it should be fine -- `#00000000` is fully transparent color

actionless: also thanks, good idea re: font-size/font-family

lib/beautiful/gtk.lua
actionless: weird, could be some `--no-rgba`-related kind of thing -- i've just tried it on other machine and it works as expected

actionless: so i am now on the workstation with radeon again, the issue is still reproduciable and happens only to taglist and tasklist `fg` color, `#888a85ff` is rendered as white and `#acafa9ff` as black

actionless: found the problem:
```
$ awesome-client 'return require("gears.color").ensure_pango_color("#888a85")'
   string "#888a85"

$ awesome-client 'return require("gears.color").ensure_pango_color("#888a85ff")'
   string "black"

```

psychon: Uhm, so, why does this work on one computer but not the other?
Anyway, [PangoColor](https://developer.gnome.org/pango/unstable/pango-Text-Attributes.html#PangoColor-struct) has just red/green/blue and [alpha](https://developer.gnome.org/pango/unstable/pango-Text-Attributes.html#pango-color-parse) is [implemented](https://developer.gnome.org/pango/unstable/pango-Text-Attributes.html#pango-attr-background-alpha-new) via two other, special properties. Pango markup hides that difference, but `pango_color_parse` can, of course, not. Meh...

actionless:
previous commit's test failed with a strange (and probably unrelated) reason:
https://travis-ci.org/awesomeWM/awesome/jobs/316735213

themes/gtk/theme.lua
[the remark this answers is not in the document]
McSinyx: That's not necessary, as [error and warning colors are exported](https://git.gnome.org/browse/gtk+/tree/gtk/theme/Adwaita/_colors-public.scss#n82).

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,174 @@
...
+    return Gtk.Window{}
+  end)
+  if not _window_status or not window then
+    gears_debug.print_warning(
+      "Can't create GTK+3 window. "..
+      "Seems like GTK+3 theme is not set correctly or `lua-lgi` was built with an incompatible GTK+3 version."
+    )
+    return nil
+  end
+  local style_context = window:get_style_context()
```
McSinyx: This should be `Gtk.StyleContext()` instead.

McSinyx:
I found [a list of exported GTK+ 3 colors](https://git.gnome.org/browse/gtk+/tree/gtk/theme/Adwaita/_colors-public.scss) that might help refining `lib/beautiful/gtk.lua`.

lib/beautiful/gtk.lua
actionless: window is used not only for retrieving the style context

actionless:
@McSinyx that's just a list of the ones exported by `Adwaita` gtk theme 😉

themes/gtk/theme.lua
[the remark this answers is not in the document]
actionless: since in majority of other gtk themes stylecontext values are partially missing, `destructive/suggested/etc-action`-approach would be more robust

lib/beautiful/gtk.lua
McSinyx: Please read your own code again and tell me where (don't count [this](https://github.com/awesomeWM/awesome/pull/2129/files/43b17d65d281f01b0ab4d65e5778c9268a4440e3#diff-45ad409b45603061b8a7d3bd863c9555R95) as `{b,f}g_color` are exported).

McSinyx:
And Adwaita is the default (read: conventional) theme (actually, those colors are exported from HighContrast too). I understand that GTK+ documentation for this is poor and the theming API is not stable, but we shouldn't sacrifice efficiency just because some guys couldn't catch up to the new convention. I see you have lots of work related to GTK+ theming, so I think you already know most (if not all) gtk3 themes survive till today are based on Adwaita.

I'm just saying I'd love to see if you can use the standard exported values first, and fallback to ugly hacks later. Thank you guys for keeping maintaining and developing such an awesome window manager.

actionless:
@McSinyx see here: https://github.com/awesomeWM/awesome/pull/2129/files/43b17d65d281f01b0ab4d65e5778c9268a4440e3#diff-45ad409b45603061b8a7d3bd863c9555R161

McSinyx:
Excuse my English but I don't understand your reply clear enough. I was asking about the use of the local variable `window` which is a `Gtk.Window` instance and suggested creating a style context by excuting `Gtk.StyleContext()` instead of `window.get_style_context()`.

> the thing is what it's already implemented in that way

I don't really get which you meant to say implemented: this PR or carelessly contructed GTK+ themes?

actionless:
the line which referenced is implementing this:

>  if you can use the standard exported values first, and fallback to ugly hacks

actionless:
i did some improvements, here how it looks currently:

2018-01-12-231222_3840x2160_scrot
2018-01-12-231306_3840x2160_scrot
2018-01-12-231402_3840x2160_scrot

actionless:
the remaining part: for gtk widgets which doesn't have `background-color` but `background-image` with gradient (`cairo.SurfacePattern`) to get average color from the gradient and expose surface to beautiful.gtk.get_theme() output to use it later as `beautiful.*_bgimage`

actionless:
idk what's wrong with the regex in the last commits -- for some reason they not filtering out warnings from gtk theme

actionless:
@Elv13 mb you have a moment now to briefly look into this topic?

also i think i'll need some help dealing with gradient patterns

actionless:
> idk what's wrong with the regex in the last commits -- for some reason it not filtering out warnings from the gtk module

also i forgot to mention what running the tests locally resulting in pass (i've copied all options from travis log to exactly reproduce the case and prepared a case when gtk theme will have the same value missing)

so i still think there's something wrong in my regex but for some reason it works differently in different grep version or locale or smth else

actionless:
CI is passing except for PREV_COMMITS (that will go after squash)

let's do some reviewing?

lib/awful/widget/taglist.lua
```diff
@@ -149,9 +150,27 @@ taglist.filter = {}
 -- @beautiful beautiful.taglist_spacing
 -- @tparam[opt=0] number spacing The spacing between tags.
 
+--- The shape used for taglist widget container itself.
```
Elv13: I am not sure I like this. You are just adding random code to the taglist (but not the tasklist) that has no advantage over the user just wrapping her/his/it `taglist` in a `background` container. I see this the same way as adding `set_width/set_height/set_margin` to all widgets instead of using the idiomatic `margin` container and `forced_*` .

I understand your theme looks nicer with it, but I disagree it is a reason to extent the widgets. A future proof solution would be to add a `beautiful.main_wibar` (needs a better name) and use `beautiful.main_wibar or wibox.wibar {...}` in `rc.lua`. This way all those "but, but, it looks better!" changes could be implemented per theme instead of adding yet more options to the `taglist`/`tasklist`. It this point, I think I see the taglist/tasklist has in perpetual feature freeze.

lib/gears/color.lua
```diff
@@ -338,6 +338,11 @@ end
 -- @treturn string color if it is valid, else fallback.
 function color.ensure_pango_color(check_color, fallback)
     check_color = tostring(check_color)
+    -- Pango markup supports alpha, PangoColor does not. Thus, check for this.
```
Elv13: These commits are merged, please rebase

lib/awful/widget/taglist.lua
actionless: that will make theme not really reusable in configs with any custom widgets involved

actionless: mb smth like `beautiful.taglist_base_widget` ?

Elv13: And adding unlimited options to taglist/tasklist *already* make it unmaintainable. Also, people could just modify the themes instead of `rc.lua`. In the long run, this is a **good** thing since we might have more themes usable with the default config instead of LCPZ/Copycat-killer `rc.lua` forks for every configs.

Visual code should go in the theme, window management logic in `rc.lua`. If we can find the right balance and have a nice separation of concern between the view and the logic, everybody wins.

Elv13: > beautiful.taglist_base_widget

It would not help, it will even make it worst. Think of use case such as the `tasklist` being used for both the wibar and in an alt+tab like popup. In that case adding too many options will make it impossible for module to reuse the tasklist without making sure to override every single style variable just to ensure it works.

actionless: that approach makes harder to just switch themes with having single rc.lua

btw we `widget_template` in tasklist, having something like that there will be useful for more uniform API across default wibar widgets

actionless: so, speaking about balance, i think the right one will be instead of providing only options in beautiful add more shortcuts for overriding standard widgets' contents

actionless: but that's already a documented API in tasklist itself

Elv13: > but that's already a documented API in tasklist itself 

Maybe we are not talking about the same thing, can you provide a documentation link?

> Visual code should go in the theme

Yes, but I am thinking about higher level code (the whole wibar, notification and client decoration). Style options on the widget level are required to reuse them for different purpose (see the layoutlist in the other PR, it is a better example than the taglist). But IMO this is a longer term feature, not something for this PR. For now I am just against adding `widget_template` in the taglist beautiful namespace and wraping the taglist into a background for what I consider the wrong reasons. However, `taglist_spacing` is ok. While it contradict what I just said a bit, you are right that it is in line with what's already exposed to `beautiful`. Adding the background and widget_template, on the other hand will break things for many valid use cases.

actionless: 1) https://awesomewm.org/apidoc/classes/awful.widget.tasklist.html#awful.tasklist, second argument

2) i still think what mooving thing like a whole wibar to the theme will be not really reusable because wibar layout itself is not a matter of styling only but a matter of functionality, so this proposa means what any theme customization made on wibar will be skipped if user want to have custom wibar layout (ie wants to add one more custom widget to default wibar layout)

actionless:
bumping this since it have an unfinished discussion regarding extra styling of default panel elements via beautiful

lib/awful/widget/taglist.lua
actionless: btw, that's one more example why `beautiful.main_wibar` approach is bad (in the example it's called `beautiful.at_screen_connect(s)` but seems to be doing the same job as beautiful.main_wibar would do):

https://github.com/actionless/awesome_config/issues/24

actionless: so that will be a painful decision, to make tasklist and taglist code more complicated, but we have to make to avoid even harder decisions in beautiful or leaving users to modify rc.lua for things which are obviousle part of the theme (like widgets' rounding)

actionless:
i'll be bumping this every months :laughing: :cherry_blossom:

lib/awful/widget/taglist.lua
Elv13: I really don't think doing changes to the `awful.widget.common` based widgets just to add more theme variables is a good idea. You are right that fixing it using higher level means is more complicated and will require long discussions and such. However doing the changes you propose here is a dead end. It will always be "but I only need one more" over and over again.

https://github.com/awesomeWM/awesome/issues/442 is "right" in the sense 99% of users `rc.lua` are either the default one + few changes or based on @lcpz templates. This hurt the project in the long run because people assume Awesome is inflexible while our very mission is to be flexible. At some point the issue have to be addressed one way or another. Burying our head in the sand and having the core developers adding more and more hardcoded things in those widgets to fit their current taste or the latest trends is not solving anything. It just proves those who think Awesome is inflexible right. If we have to modify core components to get anything done, then we failed somewhere.

The reason I pushed the "delegate" system was to put a definitive end on those options. What you propose here duplicates existing functionalities (the `wibox.container.background` container. You can say the `textclock`/`layoutbox`/`menu`/`systray` also lacks a shape, do you really think it's a good idea to add such options and theme variables to every single things? I admit I myself went a little too far with them and if I had implemented the delegates earlier I would not have added them to the taglist/tasklist. It was my mistake and my lazyness of not addressing the problem and mitigating the consequences.

Also the other argument that adding this options *breaks* the widget for alternate use cases is a big and definitive blocking issue. The `beautiful.tasklist_widget_template` **cannot** exist.

If you want this PR to ever be merged, only add the gtk theme and its helper function (and maybe more integration tests for them). Otherwise I don't see this going anywhere. Don't get me wrong, I am not against the concept of having a theme that follow the GTK palette. I think it is useful and worthy of inclusion. My issue are that the way it adds theme options is a no go even if there is no currently available viable alternatives.

actionless: I see how your comment explains why implementing `beautiful.taglist_shape_*` is bad.

However i still can't undestand why `widget_template` can exist as argument to tasklist constructor but not as `beautiful.tasklist_widget_template`.

Elv13: Because if someone want to build an `alt+tab` widget using the tasklist like an example proposed in my `popup` PR does, then having a template potentially "damaged" by the theme variables forces any implementation to copy/paste the default delegate just in case the theme messed up the template in a way that renders such alt+tab widget useless.

actionless: ok, that makes sense

i'll try to similify the theme and update the PR during the week

actionless:
updated according to @Elv13's point of view

actionless:
is there still anyathing wrong? :pouting_cat:

Elv13:
Please remove all the modifications to `rc.lua` and the extra template related beautiful variables. This deserve better APIs and more work/discussions. The GTK theme part is (to me) rather uncontroversial. And should be reviewed and merged independently. I know you want it to look a certain way and and it cannot be done without the current API. However it wont get merged as is, so please just remove the controversial part so the uncontroversial one can be reviewed/merged without the distractions.

actionless:
@Elv13 i did the changed which you wanted however i think what now taglist/tasklist improvements should be addressed as soon as possible because the latest changes are making gtk theme unusable for the most of the users

actionless:
and also without the changes which you rejected the tasklist looks just ridiculously bad on a modern display:
2018-03-12-030603_286x171_scrot

actionless:
i mean if you don't like my approach i expect you fix it all "the right way" asap otherwise that's not really good attitude

Elv13:
> i mean if you don't like my approach i expect you fix it all "the right way" asap otherwise that's not really good attitude

I know... It's just sometime that going nothing is a better solution that having to fix the consequences of doing something. That's even worst when you see issues upfront and you are "oh well, I should have avoided this back when I could".

What do you think about new signals for `screen` and `client` called `request::decoration`. Then if `myscreen.is_decorated` or `myclient.is_decorated` is set, the default `rc.lua` section for this just `return`? It isn't perfect, but maybe we can improve on that? Maybe even add a `connect_signal_if` kind of API for objects where something is connected only if there isn't already something connected to it? I am just throwing some ideas about potential solutions to allow the theme to create the wibars with whatever they want. I really think this is the only way to avoid further feature creep in the current `rc.lua` or `awful.common` based widgets.

psychon:
A wild Uli appears!

How about having a function in the GTK-based theme that can be called by rc.lua, but is not actually used by default. Then you just need to document that this specific theme is best used by replacing [this part] of the default config with a call to `beautiful.foo_bar_baz()`.

So what are we talking about? Some magic to the taglist and tasklist? How about this?
```diff
diff --git a/lib/awful/widget/common.lua b/lib/awful/widget/common.lua
index 62b791c8..4b4eb92b 100644
--- a/lib/awful/widget/common.lua
+++ b/lib/awful/widget/common.lua
@@ -101,7 +101,7 @@ function common.list_update(w, buttons, label, data, objects, args)
 
         if not cache then
             cache = (args and args.widget_template) and
-                custom_template(args) or default_template()
+                custom_template(args) or require("beautiful").get().default_no_idea_what_to_call_this_template() or default_template()
 
             cache.primary:buttons(common.create_buttons(buttons, o))
 
```
It means that the config still "wins" over the theme, but themes can override the default template.

...and this Uli disappears again before he can ask too many "why????" questions.

Elv13:
This brings back the previous problem I highlighted above. If something like an "alt-tab" popup extension module uses the tasklist and the magic beautiful variables create a templates that render such popup unusable (like replacing the task by fancy colors or something equally weird), then the `awful.widget.tasklist` is "broken" in the sense modules cannot rely on it's behavior anymore. It's akin to the old monkeypatching nightmare that broke all my modules and then I got angry emails complaining my modules had bugs.

I am not saying there is no instances of such issues, but I would rather avoid adding APIs or theme variables that render the core Awesome APIs unpredictable when used from modules. The Awesome APIs should be deterministic from a module point of view. Otherwise all modules need to re-invent the wheel all the time to avoid being shot in the foot by variable that happen to be impossible to introspect (like the functions you just proposed).

actionless:
>  allow the theme to create the wibars with whatever they want

@Elv13 i still think it's bad idea (i was already justifying that quite a lot during the discussion here), and i even gave you a real example of an exactly the same issue in the wild:

https://github.com/awesomeWM/awesome/pull/2129#discussion_r169763340


>  If something like an "alt-tab" popup extension module uses the tasklist 

that sounds a bit questionable on regard if it's a hack or not. if it would be called common.AbstractListOfTasks not widget.tasklist that won't be so much questionable as it is now

actionless:
however, since that seems to be already out of the scope of this PR we could create some ticket like `More styling options for default widgets` (?) to discuss the better approach on API more purposefully

actionless:
i've created the new ticket for that: https://github.com/awesomeWM/awesome/issues/2226

now please review what 's left in scope of this PR

Elv13:
I will do it soon-ism. I just really want to try it first instead of just proof reading the code.

For now and until I find some time, can you fix the indentation issue (2 spaces vs. 4 spaces) and the some of the commented code / TODO looks like they can be addressed now.

> i've created the new ticket for that: #2226

Thanks

actionless:
btw do we have some nie way to draw a table in the luadoc?
i was thinking what documentation for output values and all their fallbacks (some of them have more than one fallback with some priorities) will be much more readable than just a list

Elv13:
I use HTML. The problem is that while discount-markdown supports tables, luadoc supports multiple MD parser and they have different syntax... Also, since we updated the CSS, they look horrible https://awesomewm.org/apidoc/classes/awful.widget.tasklist.html

actionless:
i am feeling like i can address the table styles as part of the PR, what do you think?

Elv13:
Or as another PR. That solves the "@Elv13 has no time to try this PR" issues and can get merged much more quickly. By quickly, I mean minutes if you include a screenshot.

actionless:
2018-03-20--1521563604_2674x1588_scrot

actionless:
or mb with no border at all

2018-03-20--1521563732_3365x1462_scrot

Elv13:
I prefer with light border and the same color as the normal page BG? Just an opinion. Everything better than the current one is welcome.

Elv13:
Oh, an also, while you are at it 

https://awesomewm.org/apidoc/classes/client.html#client.type
https://awesomewm.org/doc/api/classes/client.html#client.type

There's another regression where the bullet point list is broken

actionless:
i thought it was purposely removed

Elv13:
It looks very confusing for some of the lists. I think a bullet point list should have bullet points. Otherwise why was it declared a bullet point in the first place?

actionless:
i know it could look nicer but at least it's very detailed:

2018-05-05--1525539365_1418x970_scrot


i am done, will squash after review

actionless:
i've made travis happy again (except for per-commit check):
https://travis-ci.org/awesomeWM/awesome/builds/375327553?utm_source=github_status&utm_medium=notification

actionless:
now i like it more:

2018-05-05--1525547503_1606x1038_scrot

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,282 @@
...
+-- @treturn table Key-value table with the following structure:
+-- <table class='widget_list' border=1>
+-- <tr style='font-weight: bold;'> <th align='center'>Result key</th> <th align='center'>StyleContext key</th> <th align='center'>StyleContext fallback #1</th> <th align='center'>StyleContext fallback #2</th> <th align='center'>GTK Widget fallback</th> </tr>
+-- <tr> <td>`font_size`</td> <td></td> <td></td> <td></td> <td>Label font-size</td> </tr>
+-- <tr> <td>`font_family`</td> <td></td> <td></td> <td></td> <td>Label font-family</td> </tr>
+-- <tr> <td>`bg_color`</td> <td>`theme_bg_color`</td> <td></td> <td></td> <td>Window bg</td> </tr>
+-- <tr> <td>`fg_color`</td> <td>`theme_fg_color`</td> <td></td> <td></td> <td>Window fg</td> </tr>
+-- <tr> <td>`base_color`</td> <td>`theme_base_color`</td> <td></td> <td></td> <td>Entry bg</td> </tr>
+-- <tr> <td>`text_color`</td> <td>`theme_text_color`</td> <td></td> <td></td> <td>Entry fg</td> </tr>
+-- <tr> <td>`button_bg_color`</td> <td>`theme_buton_bg_color`</td> <td>`theme_bg_color`</td> <td></td> <td>Button bg</td> </tr>
```
alfunx: There's a small typo in `theme_buton_bg_color`.

actionless: thanks for your eagle eye!

actionless:
some fresh screenshots of gtk library in use with the different gtk themes (panel is using my config but subj theme):

2018-05-09--1525821066_1920x1080_scrot
2018-05-09--1525821077_1920x1080_scrot

2018-05-09--1525821178_1920x1080_scrot
2018-05-09--1525821203_1920x1080_scrot
2018-05-09--1525821276_1920x1080_scrot
2018-05-09--1525821292_1920x1080_scrot
2018-05-09--1525821292_1920x1080_scrot
2018-05-09--1525821318_1920x1080_scrot
2018-05-09--1525821058_1920x1080_scrot

actionless:
also found a regression in pixmap-rich themes which i've introduced while removing the debug code, now it considerably fine:

2018-05-09--1525822763_1920x1080_scrot

2018-05-09--1525822777_1920x1080_scrot

Elv13:
> @Elv13 anything more to be done?

I am sorry, I just really don't have time for Awesome right now. It's not in bad faith, I just don't. It's always "oh, I will do that next weekend" then there is other things and it ends up not being done. It's not specific to your PRs either, even my own PRs are equally neglected.

actionless:
@blueyed your review is also pending... :P

docs/ldoc.css
```diff
@@ -179,10 +179,6 @@ table.index td { text-align: left; vertical-align: top; }
     background-color: white;
 }
 
-#content p, #content table, #content ol, #content ul, #content dl {
-    max-width: 900px;
-}
```
psychon: Huh? Why?

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,293 @@
...
+    if not _gtk_status or not Gtk then
+        gears_debug.print_warning(
+            "Can't load GTK+3 introspection. "..
+            "Seems like GTK+3 is not installed or `lua-lgi` was built with an incompatible GTK+3 version."
+        )
+        return nil
+    end
+    local _window_status, window = pcall(function()
+        return Gtk.Window{}
+    end)
```
psychon: I don't see any way that this can fail in the docs....?

psychon: How about `Gtk.Window.new(Gtk.WindowType.TOPLEVEL)`? That seems to be the documented way to create `GtkWindow`s and it makes sure that the `type` property is set up correctly (it can only be set while constructing a window).

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,293 @@
...
+
+    result = join(result, read_gtk_color_properties_from_widget(
+        window, {
+            bg_color="background-color",
+            fg_color="color",
+        }
+    ))
+    local label = Gtk.Label()
+    local label_style_context = label:get_style_context()
+    local gdk_scale = os.getenv("GDK_SCALE") or 1.0
```
psychon: Doesn't this need a `tonumber()`?

lib/beautiful/gtk.lua
```diff
@@ -0,0 +1,293 @@
...
+    local xrdb_scale = dpi(10000) / 10000
```
psychon: Uhm. Why not just `beautiful.xresources.get_dpi() / 96`?

themes/gtk/theme.lua
```diff
@@ -0,0 +1,353 @@
...
+        result = result .. string.format("%02x", bg_numeric_value)
+    end
+    return result
+end
+local function reduce_contrast(color, ratio)
+    ratio = ratio or 50
+    return darker(color, is_dark(color) and -ratio or ratio)
+end
+
+local function choose_contrast_color(reference, candidate1, candidate2)  -- luacheck: no unused
```
psychon: Why is the `no unused` needed?

themes/gtk/theme.lua
```diff
@@ -0,0 +1,353 @@
...
+-        s.mytaglist = background_shape_wrapper
+-    end
+
+     -- Create a tasklist widget
+     s.mytasklist = awful.widget.tasklist {
+         screen  = s,
+         filter  = awful.widget.tasklist.filter.currenttags,
+-        buttons = tasklist_buttons,
+-        widget_template = beautiful.tasklist_widget_template
++        buttons = tasklist_buttons
```
psychon: Uhm, is the patch backwards?

themes/gtk/theme.lua
```diff
@@ -0,0 +1,353 @@
...
+            },
+            layout = wibox.layout.fixed.horizontal,
+        },
+        left  = dpi(2),
+        right = dpi(4),
+        widget = wibox.container.margin
+    },
+    id     = 'background_role',
+    widget = wibox.container.background,
+    create_callback = function(self, c, index, objects) --luacheck: no unused args
```
psychon: Just make this `function(_, c)` and you do not need the luacheck comment

themes/gtk/theme.lua
```diff
@@ -0,0 +1,353 @@
...
+    theme.gtk.header_button_bg_color,
+    0.3
+)
+theme.taglist_fg_empty = mix(
+    theme.gtk.menubar_bg_color,
+    theme.gtk.header_button_fg_color
+)
+
+theme.wibar_bgimage = theme.gtk.menubar_bg_image
+--print(theme.wibar_bgimage)
```
psychon: Huh?

lib/beautiful/gtk.lua
actionless: it was hapenning to me when lgi was compiled with an older gtk version (so it was failing to do anything with gtk)

themes/gtk/theme.lua
actionless: oops, those are leftovers from the next version, which support pixmap-ish themes better, but it wil take some time to come.. :-D

lib/beautiful/gtk.lua
psychon: LGI's C code does not mention GTK outside of comments...?!? What exactly does it mean to compile LGI against an older GTK version?
Is this perhaps because this code bypassed `gtk_window_new` (as I mentioned in some other comment) and does magic with GObject directly?

actionless: mb gobject-introspection was upadted on my machine at relatively the same time

actionless:
squashed + sanitized the commit message

actionless:
@Elv13 @blueyed it still needs one more review

Elv13:
@actionless oh, an you might want to update the optional dependency section of README.md

I have updated the NEWS changelog for 4.3, does this description is fine with you (or propose something better):

```
### A new GTK color palette based theme

A new theme has been added. It reads the GTK theme colors and use them in the
`wibar`, `menu` and `titlebar`. It helps create an uniform look and feel
between the window manager and client applications with minimal efforts.
```

actionless:
actually it reads not only colors but also border thickness and shape and few more things, however for those things support is very partial so mb your description will be fine (so users won't be expecting too much from it :-) )


yup, i'll update the readme. idk which is the minimal supported gtk3 version but the one installed inside travis seems to be compatible enough :-)

actionless:
>  actionless dismissed stale reviews from Elv13 and psychon via c660893 just now

fuck :-(

Elv13:
Do you want me to merge it now? Psychon is away until mid August so the review wont be updated.

actionless:
i've been testing gtk.lua code (not theme itself) for more than a year so it should be quite robust

last issue which i found and resolved (about 1-2 months ago) was what it wasn't working properly in some edge cases of GTK+3 scaling + xrdb dpi

Elv13:
> i've been testing gtk.lua code (not theme itself) for more than a year so it should be quite robust

I have been testing the new layout stuff for years too, we all have heaps of unmerged code ;)

So do I press the button or not? Your call.

actionless:
yes, it's good to merge

Elv13:
:1st_place_medal:

actionless:
thanks! :tada:

PR #2134, opened 2017-12-07
Make ensure_pango_color support #rgba

psychon:
Reference: https://github.com/awesomeWM/awesome/pull/2129#discussion_r155397507

### DIFF
--- a/lib/gears/color.lua
+++ b/lib/gears/color.lua
@@ -338,6 +338,11 @@ end
 -- @treturn string color if it is valid, else fallback.
 function color.ensure_pango_color(check_color, fallback)
     check_color = tostring(check_color)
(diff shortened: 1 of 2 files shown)


lib/gears/color.lua
```diff
@@ -338,6 +338,11 @@ end
 -- @treturn string color if it is valid, else fallback.
```
actionless: maybe mention in docs what it skip checking the certain "types" of colors?

psychon: Sorry, what?

actionless: my bad, i didn't noticed `%x` in the regexp so i thought it's only checking the length and if length seems like argb value it just returns it back

actionless:
i think one week of conservation is enough for such a small change covered with the tests

Issue #2150, opened 2018-01-04
beautiful xresources only defaults


ransomts:
Output of `awesome --version`:

> awesome v4.2 (Human after all)
 > • Compiled against Lua 5.3.4 (running with Lua 5.3)
 > • D-Bus support: ✔
 > • execinfo support: ✔
 > • xcb-randr version: 1.5
 > • LGI version: 0.9.2

Also fwiw, my kernel is 4.14.10-1-ARCH and I got my builds from the official repo.

**How to reproduce the issue:**

Just adding in the normal use cases for the xresources require.
Running this through terminal:
```
awesome-client -v 'awful=require("awful");
 xresources=require("beautiful.xresources");
 xrdb = xresources.get_current_theme();
 awful.spawn("notify-send _"..xrdb.color1)'
```

**Actual result:**

I just get the #cb1578 from the fallbacks in xresources.lua sent back to me.

**Expected result:**

> *.color1: #38B1C6
> *color1:  #38B1C6

This is the relevent section of my .Xresources, so I would like to see this value instead.

**Temp Solution**
Here is my drop-in for similar functionality, but I would still prefer not to make a clone if the original works. Any ideas where my .xresources might be going wrong?
```lua
function os.capture(cmd, raw)
  local f = assert(io.popen(cmd, 'r'))
  local s = assert(f:read('*a'))
  f:close()
  if raw then return s end
  s = string.gsub(s, '^%s+', '')
  s = string.gsub(s, '%s+$', '')
  return s
end

function getXrdbTable()
   x = os.capture("xrdb -query")
   data = {}
   for line in string.gmatch(os.capture("xrdb -query"), "[^\n]*") do
	  tt = {}
	  for element in string.gmatch(line, "[^:]*") do
		 table.insert(tt, element)
	  end
	  data[tt[1]] = tt[2]:gsub("%s+", "")
   end
   return data
   --for key,value in pairs(data) do print(key,value) end
end

xrdb = getXrdbTable()
```

actionless:
what is the output of?
```sh
awesome-client 'return awesome.xrdb_get_value("", "color1")'
awesome-client 'return awesome.xrdb_get_value("", ".color1")'
```

ransomts:
```
$ awesome-client 'return awesome.xrdb_get_value("", "color1")'
   string "#38B1C6"
$ awesome-client 'return awesome.xrdb_get_value("", ".color1")'
   string "#38B1C6
```
[...]

PR #2151, opened 2018-01-04
fix(beautiful: xresources): don't return fallback colorscheme if only some of the values are missing

actionless:
related to #2150


actionless:
i can't merge with 75% diff hit

Elv13:
Ah, you should have said so earlier. @blueyed As our expert on this, is there a way to turn off the requirements for patches with less than 50 lines of change?

PR #2153, opened 2018-01-11
Add set/get for textclock attributes

otommod:
It's not very pretty the way it's implemented but hey, it works.  My
reason for wanting this was to allow "declarative" construction of
clocks.

If you'd prefer it to be more "class-like", I'd be happy to change it.

### DIFF
--- a/lib/wibox/widget/textclock.lua
+++ b/lib/wibox/widget/textclock.lua
@@ -40,15 +40,15 @@ function textclock:get_timezone()
     return self._private.tzid
 end
 
---- Set the clock's timeout
+--- Set the clock's refresh rate
 -- @tparam number How often the clock is updated
-function textclock:set_timeout(timeout)
-    self._private.timeout = timeout or self._private.timeout
+function textclock:set_refresh(refresh)
+    self._private.refresh = refresh or self._private.refresh
     self:force_update()
 end
 
-function textclock:get_timeout()
-    return self._private.timeout
+function textclock:get_refresh()
+    return self._private.refresh
 end
 
 --- Force a textclock to update now.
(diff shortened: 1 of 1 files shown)


actionless:
looks fine but it's missing docstrings

otommod:
I added them.  Where are the docs build though?

actionless:
default make target will build the docs

lib/wibox/widget/textclock.lua
```diff
@@ -55,6 +55,39 @@ function textclock.new(format, timeout, timezone)
...
+        timezone = tzid and TimeZone.new(tzid) or TimeZone.new_local()
+        w:force_update()
+    end
+    function w.get_timezone()
+        return tzid
+    end
+
+    --- Set the clock's timeout
+    -- @tparam number How often the clock is updated
+    function w.set_timeout(_, new_timeout)
```
Elv13: Maybe call it refresh rate?

lib/wibox/widget/textclock.lua
```diff
@@ -55,6 +55,39 @@ function textclock.new(format, timeout, timezone)
         w._timer:again()
         return true -- Continue the timer
     end
+
+    --- Set the clock's format
+    -- @tparam string format The new time format.  This can contain pango
+    -- markup
+    function w.set_format(_, new_format)
```
Elv13: Please do not add functions in the constructor, add them outside and store the attributes in `self._private`

Also, use `@property` for properties and add an extra empty line under the documentation block.

lib/wibox/widget/textclock.lua
otommod: Sure.

lib/wibox/widget/textclock.lua
otommod: I thought as much.  Fair enough.  However, how should I then add the methods?  I mean, right now we're just returning a regular ol' textbox with a timer attached, not a textclock "class".

psychon: Elv13's method seems to be:
```lua
local foo = {}
function foo:set_format(new_format)
    self._private.format = new_format
end

function constructor()
    local w = textbox()
    gears.table.crush(w, foo)
    return w
end
```

otommod:
Updated.

Elv13:
Thanks!

PR #2154, opened 2018-01-13
Add the popup widget

Elv13:
This is the popup widget, a wibox that handles most resizing and placement without
boilerplate code.

This was originally designed to be the final part of the "geometry overhaul" PR
series in 2015-2016, but was never finished.

I rewrote it a few time since then and I am finally less or more happy with the
API. As @psychon would have said, the first few revisions were sledgehammer APIs
that had too many options and did too many things. Thankfully some of these
things, like the shape, are already merged and no longer part of the popup class.
I also dropped a lot of dark magic to automatically move the popup when its parent
moves. It added too much logic for too little gain.

Half of the popup was already merged as `awful.placement.next_to`. The code still
heavily rely on it. I fixed the border_width issue in `next_to` that went
unnoticed since both the original popup (as used in Radical) and the tooltip used
widget  borders.

There is also some more work to make the shims good enough to emulate most of Awesome. It has a a few issues. For some reasons `:find_widgets(x, y)` doesn't work. I didn't investigate much, but it's not so simple to fix (and no, adding more "refresh" doesn't help) and while it messes with some popup examples a bit, I think it can be left for later. This PR is already a monster as-is.

screenshot_20180113_040418

Finally, the PR also adds the layoutlist widget as it makes a good example use
case. I had this one for ages (all the way back to Awesome 3.3). As the
awful.widget.common template is now merged, it makes more sense to upstream it.

NOTE THAT IT PULLS #2123 commits until they are merged to avoid the GDK warning
on Travis.

### DIFF
--- a/docs/03-declarative-layout.md
+++ b/docs/03-declarative-layout.md
@@ -1,7 +1,6 @@
-# The declarative layout system
+# The AwesomeWM widget system
 
-The declarative layout system provides an alternative to the imperative system.
-It is inspired by the one used by Awesome 3.2-3.4 and the Qt QML style.
+This document explains how to define, place and manage widgets.
 
 ## The default widgets
 
@@ -29,7 +28,127 @@ configurable rules.
 
 @DOC_layout_WIDGET_LIST@
 
-## Placing widgets
+### The different type of widget boxes (Wibox)
+
+The Awesome API uses the word "wibox" (widget box) to describe an area of the
+screen filled with widgets. There are many subvariants of wiboxes with
+specialized role such as widget bars or tooltips. All variants mostly share the
+same characteristics, but add some extra features to make those specialized
+widget boxes easier to work with.
+
+@DOC_awful_popup_wiboxtypes_EXAMPLE@
+
+The normal `wibox` is the base class for each of these type. It is extremely
+flexible and allows to place just about anything on the screen. However it
+requires a lot of repetitive boiler plate code to use directly. For example,
+the user need to compute the optimal size by hand or use `awful.placement`.
+
+The `awful.wibar` specialization allows to attach a `wibox` to a screen edge
+and prevent clients from using this area when tiled.
+
+The `awful.popup` allows to easily place widgets on the screen. It automatically
+resize itself to fit the optimal widget size. It also has helper properties
+and methods to make it easy to place it on the screen. It both supports absolute
+positioning, relative positioning and manual positioning.
+
+The `awful.tooltip` is a very simple `wibox` that allows to display text next
+to an object such as the mouse.
+
+Finally, the `awful.titlebar`, while not technically a real `wibox`, acts
+exactly the same way and allows to attach widgets on each side of clients.
+
+## The different syntaxes to initiate widgets
+
+Awesome provides 2 totally different API access styles to manage widgets. Both
+suit different use case are are equivalent. Under the hood, both produces the
+exact same code. Consider the declarative API to be compiled into the
+imperative syntax when loaded. Also note that in contract to technologies such
+as QML, it is interpreted only once and isn't automatically updated when
+value changes.
+
+The **imperative** widget initialization is similar to QtWidgets, GTK and Win32.
+You create the object, then set the property and add the widget as a child to
+another already declared widget. It is quite simple to use but very verbose
+and full of boilerplate code. The imperative API also offers properties both
+with accessors or directly. It is useful when creating highly dynamic layouts
+where widgets are added and removed over the course of their lifecycle.
+
+The **declarative** syntax resembles HTML style code
+written in JSON or YAML. The widgets instances are created automatically and
+the hierarchy is related to the table nesting (indentation). It is preferred
+when creating static layouts that won't change over the course of their
+lifecycle.
+
+Here is the same code written in both the imperative and declarative style
+
+**Imperative with accessors**
+
+Code:
+
+    local bg = wibox.container.background()
+    bg:set_bg("#ff0000")
+
+    local tb1 = wibox.widget.textbox()
+    local tb2 = wibox.widget.textbox("bar")
+
+    tb1:set_text("foo")
+    tb2:set_text("bar")
+
+    local l = wibox.layout.fixed.vertical()
+    l:add(tb1)
+    l:add(tb2)
+
+    bg:set_widget(l)
+
+**Imperative with properties**
+
+Code:
+
+    local bg = wibox.container.background()
+    bg.bg = "#ff0000"
+
+    local tb1 = wibox.widget.textbox("foo")
+    local tb2 = wibox.widget.textbox("bar")
+
+    tb1.text = "foo"
+    tb2.text = "bar"
+
+    local l = wibox.layout.fixed.vertical()
+    l:add(tb1)
+    l:add(tb2)
+
+    bg.widget = l
+
+**Declarative**
+
+Code:
+
+    local bg = wibox.widget {
+        {
+            {
+                text = "foo",
+                widget = wibox.widget.textbox
+            },
+            {
+                text = "bar",
+                widget = wibox.widget.textbox
+            },
+            layout = wibox.layout.fixed.vertical
+        },
+        bg = "#ff0000",
+        widget = wibox.container.background
+    }
+
+
+The Awesome documentation mostly uses the declarative style for consistency,
+but both are **always** available. Note that each style can be mixed with other
+styles, but this creates very confusing code and should be avoided.
+
+## Creating and placing widgets using the declarative style
+
+The examples below explain in details how to use the declarative layout system.
+The imperative system is quite self explanatory and the respective widget API
+documentation should be enough for most.
 
 ### A simple layout
 
--- a/lib/awful/popup.lua
+++ b/lib/awful/popup.lua
@@ -1,0 +1,467 @@
+---------------------------------------------------------------------------
+--- An auto-resized, free floating or modal wibox built around a widget.
+--
+-- This type of widget box (wibox) is auto closed when being clicked on and is
+-- automatically resized to the size of its main widget.
+--
+-- Note that the widget itself should have a finite size. If something like a
+-- `wibox.layout.flex` is used, then the size would be unlimited and an error
+-- will be printed. The `wibox.layout.fixed`, `wibox.container.constraint`,
+-- `forced_width` and `forced_height` are recommended.
+--
+--@DOC_awful_popup_simple_EXAMPLE@
+--
+-- Here is an example of how to create an alt-tab like dialog by leveraging
+-- the `awful.widget.tasklist`.
+--
+--@DOC_awful_popup_alttab_EXAMPLE@
+--
+-- @author Emmanuel Lepage Vallee
+-- @copyright 2016 Emmanuel Lepage Vallee
+-- @classmod awful.popup
+---------------------------------------------------------------------------
+local wibox     = require( "wibox"           )
+local util      = require( "awful.util"      )
+local placement = require( "awful.placement" )
+local xresources= require("beautiful.xresources")
+local timer     = require( "gears.timer"     )
+local capi      = {mouse = mouse}
+
+
+local module = {}
+
+local main_widget = {}
+
+-- Get the optimal direction for the wibox
+-- This (try to) avoid going offscreen
+local function set_position(self)
+    -- First, if there is size to be applied, do it
+    if self._private.next_width then
+        self.width = self._private.next_width
+        self._private.next_width = nil
+    end
+
+    if self._private.next_height then
+        self.height = self._private.next_height
+        self._private.next_height = nil
+    end
+
+    local pf = self._private.placement
+
+    if pf == false then return end
+
+    if pf then
+        pf(self, {bounding_rect = self.screen.geometry})
+        return
+    end
+
+    local geo = self._private.widget_geo
+
+    if not geo then return end
+
+    local _, pos_name, anchor_name = placement.next_to(self, {
+        preferred_positions = self._private.preferred_directions,
+        geometry            = geo,
+        preferred_anchors   = self._private.preferred_anchors,
+        offset              = self._private.offset or { x = 0, y = 0},
+    })
+
+    if pos_name ~= self._private.current_position then
+        local old = self._private.current_position
+        self._private.current_position = pos_name
+        self:emit_signal("property::current_position", pos_name, old)
+    end
+
+    if anchor_name ~= self._private.current_anchor then
+        local old = self._private.current_anchor
+        self._private.current_anchor = anchor_name
+        self:emit_signal("property::current_anchor", anchor_name, old)
+    end
+end
+
+-- Set the wibox size taking into consideration the limits
+local function apply_size(self, width, height, set_pos)
+    local prev_geo = self:geometry()
+
+    width  = math.max(self._private.minimum_width  or 1, math.ceil(width  or 1))
+    height = math.max(self._private.minimum_height or 1, math.ceil(height or 1))
+
+    if self._private.maximum_width then
+        width = math.min(self._private.maximum_width, width)
+    end
+
+    if self._private.maximum_height then
+        height = math.min(self._private.maximum_height, height)
+    end
+
+    self._private.next_width, self._private.next_height = width, height
+
+    if set_pos or width ~= prev_geo.width or height ~= prev_geo.height then
+        set_position(self)
+    end
+end
+
+-- Layout this widget
+function main_widget:layout(context, width, height)
+    if self._private.widget then
+        local w, h = wibox.widget.base.fit_widget(
+            self,
+            context,
+            self._private.widget,
+            self._wb._private.maximum_width  or 9999,
+            self._wb._private.maximum_height or 9999
+        )
+        timer.delayed_call(function()
+            apply_size(self._wb, w, h, true)
+        end)
+        return { wibox.widget.base.place_widget_at(self._private.widget, 0, 0, width, height) }
+    end
+end
+
+-- Set the widget that is drawn on top of the background
+function main_widget:set_widget(widget)
+    if widget then
+        wibox.widget.base.check_widget(widget)
+    end
+    self._private.widget = widget
+    self:emit_signal("widget::layout_changed")
+end
+
+function main_widget:get_widget()
+    return self._private.widget
+end
+
+function main_widget:get_children_by_id(name)
+    return self._wb:get_children_by_id(name)
+end
+
+local popup = {}
+
+--- Set the preferred popup position relative to its parent.
+--
+-- This allows, for example, to have a submenu that goes on the right of the
+-- parent menu. If there is no space on the right, it tries on the left and so
+-- on.
+--
+-- Valid directions are:
+--
+-- * left
+-- * right
+-- * top
+-- * bottom
+--
+-- The basic use case for this method is to give it a parent wibox:
+--
+-- @DOC_awful_popup_position1_EXAMPLE@
+--
+-- As demonstrated by this second example, it is also possible to use a widget
+-- as a parent object:
+--
+-- @DOC_awful_popup_position2_EXAMPLE@
+--
+-- @property preferred_positions
+-- @tparam table|string preferred_positions A position name or an ordered
+--  table of positions
+-- @see awful.placement.next_to
+-- @see awful.popup.preferred_anchors
+
+function popup:set_preferred_positions(pref_pos)
+    self._private.preferred_directions = pref_pos
+    set_position(self)
+end
+
+--- Set the preferred popup anchors relative to the parent.
+--
+-- The possible values are:
+--
+-- * front
+-- * middle
+-- * back
+--
+-- For details information, see the `awful.placement.next_to` documentation.
+--
+-- In this example, it is possible to see the effect of having a fallback
+-- preferred anchors when the popup would otherwise not fit:
+--
+-- @DOC_awful_popup_anchors_EXAMPLE@
+--
+-- @property preferred_anchors
+-- @tparam table|string preferred_anchors Either a single anchor name or a table
+--  ordered by priority.
+-- @see awful.placement.next_to
+-- @see awful.popup.preferred_positions
+
+function popup:set_preferred_anchors(pref_anchors)
+    self._private.preferred_anchors = pref_anchors
+    set_position(self)
+end
+
+--- The current position relative to the parent object.
+--
+-- If there is a parent object (widget, wibox, wibar, client or the mouse), then
+-- this property returns the current position. This is determined using
+-- `preferred_positions`. It is usually the preferred position, but when there
+-- isn't enough space, it can also be one of the fallback.
+--
+-- @property current_position
+-- @param string Either "left", "right", "top" or "bottom"
+
+function popup:get_current_position()
+    return self._private.current_position
+end
+
+--- Get the current anchor relative to the parent object.
+--
+-- If there is a parent object (widget, wibox, wibar, client or the mouse), then
+-- this property returns the current anchor. The anchor is the "side" of the
+-- parent object on which the popup is based on. It will "grow" in the
+-- opposite direction from the anchor.
+--
+-- @property current_anchor
+-- @tparam string current_anchor Either "front", "middle", "back"
+
+function popup:get_current_anchor()
+    return self._private.current_anchor
+end
+
+--- Move the wibox to a position relative to `geo`.
+-- This will try to avoid overlapping the source wibox and auto-detect the right
+-- direction to avoid going off-screen.
+--
+-- @param[opt=mouse] obj An object such as a wibox, client or a table entry
+--  returned by `wibox:find_widgets()`.
+-- @see awful.placement.next_to
+-- @see awful.popup.preferred_positions
+-- @see awful.popup.preferred_anchors
+-- @treturn table The new geometry
+function popup:move_next_to(obj)
+    if self._private.is_relative == false then return end
+
+    self._private.widget_geo = obj
+
+    obj = obj or capi.mouse
+
+    if obj._apply_size_now then
+        obj:_apply_size_now(false)
+    end
+
+    self.visible = true
+
+    self:_apply_size_now(true)
+
+    self._private.widget_geo = nil
+end
+
+--- Bind the popup to a widget button press.
+--
+-- @tparam widget widget The widget
+-- @tparam[opt=1] number button The button index
+function popup:bind_to_widget(widget, button)
+    if not self._private.button_for_widget then
+        self._private.button_for_widget = {}
+    end
+
+    self._private.button_for_widget[widget] = button or 1
+    widget:connect_signal("button::press", self._private.show_fct)
+end
+
+--- Unbind the popup from a widget button.
+-- @tparam widget widget The widget
+function popup:unbind_to_widget(widget)
+    widget:disconnect_signal("button::press", self._private.show_fct)
+end
+
+--- Hide the popup when right clicked.
+--
+-- @property hide_on_right_click
+-- @tparam[opt=false] boolean hide_on_right_click
+
+function popup:set_hide_on_right_click(value)
+    self[value and "connect_signal" or "disconnect_signal"](
+        self, "button::press", self._private.hide_fct
+    )
+end
+
+--- The popup minimum width.
+-- @property minimum_width
+-- @tparam[opt=1] number The minimum width
+
+--- The popup minimum height.
+-- @property minimum_height
+-- @tparam[opt=1] number The minimum height
+
+--- The popup minimum width.
+-- @property maxmimum_width
+-- @tparam[opt=1] number The maxmimum width
+
+--- The popup maximum height.
+-- @property maximum_height
+-- @tparam[opt=1] number The maximum height
+
+for _, orientation in ipairs {"_width", "_height"} do
+    for _, limit in ipairs {"minimum", "maximum"} do
+        popup["set_"..limit..orientation] = function(self, value)
+            self._private[limit..orientation] = value
+            self._private.container:emit_signal("widget::layout_changed")
+        end
+    end
+end
+
+--- The distance between the popup and its parent (if any).
+--
+-- Here is an example of 5 popups stacked one below the other with an y axis
+-- offset (spacing).
+--
+-- @DOC_awful_popup_position3_EXAMPLE@
+-- @property offset
+-- @tparam table|number offset An integer value or a `{x=, y=}` table.
+-- @tparam[opt=offset] number offset.x The horizontal distance.
+-- @tparam[opt=offset] number offset.y The vertical distance.
+
+function popup:set_offset(offset)
+
+    if type(offset) == "number" then
+        offset = {
+            x = offset or 0,
+            y = offset or 0,
+        }
+    end
+
+    local oldoff = self._private.offset or {x=0, y=0}
+
+    if oldoff.x == offset.x and oldoff.y == offset.y then return end
+
+    offset.x, offset.y = offset.x or oldoff.x or 0, offset.y or oldoff.y or 0
+
+    self._private.offset = offset
+
+    self:_apply_size_now(false)
+end
+
+--- Set the placement function.
+-- @tparam[opt=next_to] function|string|boolean The placement function or name
+-- (or false to disable placement)
+-- @property placement
+-- @param function
+
+function popup:set_placement(f)
+    if type(f) == "string" then
+        f = placement[f]
+    end
+
+    self._private.placement = f
+    self:_apply_size_now(false)
+end
+
+-- For the tests and the race condition when 2 popups are placed next to each
+-- other.
+function popup:_apply_size_now(skip_set)
+    if not self.widget then return end
+
+    local w, h = wibox.widget.base.fit_widget(
+        self.widget,
+        {dpi= self.screen.dpi or xresources.get_dpi()},
+        self.widget,
+        self._private.maximum_width  or 9999,
+        self._private.maximum_height or 9999
+    )
+
+    -- It is important to do it for the obscure reason that calling `w:geometry()`
+    -- is actually mutating the state due to quantum determinism thanks to XCB
+    -- async nature... It is only true the very first time `w:geometry()` is
+    -- called
+    self.width  = math.max(1, math.ceil(w or 1))
+    self.height = math.max(1, math.ceil(h or 1))
+
+    apply_size(self, w, h, skip_set ~= false)
+end
+
+function popup:set_widget(wid)
+    self._private.widget = wid
+    self._private.container:set_widget(wid)
+end
+
+function popup:get_widget()
+    return self._private.widget
+end
+
+--- Create a new popup build around a passed in widget.
+-- @tparam[opt=nil] table args
+--@DOC_wibox_constructor_COMMON@
+-- @tparam function args.placement The `awful.placement` function
+-- @tparam string|table args.preferred_positions
+-- @tparam string|table args.preferred_anchors
+-- @tparam table|number args.offset The X and Y offset compared to the parent object
+-- @tparam boolean args.hide_on_right_click Whether or not to hide the popup on
+--  right clicks.
+-- @function awful.popup
+local function create_popup(_, args)
+    assert(args)
+
+    -- Temporarily remove the widget
+    local original_widget = args.widget
+    args.widget = nil
+
+    assert(original_widget, "The `awful.popup` requires a `widget` constructor argument")
+
+    local child_widget = wibox.widget.base.make_widget_from_value(original_widget)
+
+    local ii = wibox.widget.base.make_widget(child_widget, "awful.popup", {
+        enable_properties = true
+    })
+
+    util.table.crush(ii, main_widget, true)
+
+    -- Create a wibox to host the widget
+    local w = wibox(args or {})
+
+    rawset(w, "_private", {
+        container            = ii,
+        preferred_directions = { "right", "left", "top", "bottom" },
+        preferred_anchors    = { "back", "front", "middle" },
+        widget = child_widget
+    })
+
+    util.table.crush(w, popup)
+
+    ii:set_widget(child_widget)
+
+    -- Create the signal handlers
+    function w._private.show_fct(wdg, _, _, button, _, geo)
+        if button == w._private.button_for_widget[wdg] then
+            w:move_next_to(geo)
+        end
+    end
+    function w._private.hide_fct()
+        w.visible = false
+    end
+
+    -- Restore
+    args.widget = original_widget
+
+    -- Cross-link the wibox and widget
+    ii._wb = w
+    wibox.set_widget(w, ii)
+
+    --WARNING The order is important
+    -- First, apply the limits to avoid a flicker with large with or height
+    -- when set_position is called before the limits
+    for _,v in ipairs{"minimum_width", "minimum_height", "maximum_height",
+      "maximum_width", "offset", "placement","preferred_positions",
+      "preferred_anchors", "hide_on_right_click"} do
+        if args[v] ~= nil then
+            w["set_"..v](w, args[v])
+        end
+    end
+
+    -- Default to visible
+    if args.visible ~= false then
+        w.visible = true
+    end
+
+    return w
+end
+
+--@DOC_wibox_COMMON@
+
+return setmetatable(module, {__call = create_popup})
--- a/lib/awful/widget/layoutlist.lua
+++ b/lib/awful/widget/layoutlist.lua
@@ -1,0 +1,354 @@
+----------------------------------------------------------------------------
+--- Display the available client layouts for a screen.
+--
+-- This is what the layoutlist looks like by default with a vertical layout:
+--
+--@DOC_awful_widget_layoutlist_default_EXAMPLE@
+--
+-- In the second example, it is shown how to create a popup in centered in
+-- the screen:
+--
+--@DOC_awful_widget_layoutlist_popup_EXAMPLE@
+--
+-- This example show how to extend the `awful.widget.layoutbox` to show a layout
+-- list popup:
+--
+--@DOC_awful_widget_layoutlist_bar_EXAMPLE@
+--
+-- This example show how to add a layout subset to the default wibar:
+--
+--@DOC_awful_widget_layoutlist_wibar_EXAMPLE@
+--
+-- @author Emmanuel Lepage Vallee &lt;elv1313@gmail.com&gt;
+-- @copyright 2010, 2018 Emmanuel Lepage Vallee
+-- @classmod awful.widget.layoutlist
+----------------------------------------------------------------------------
+
+local wibox    = require("wibox")
+local awcommon = require("awful.widget.common")
+local abutton  = require("awful.button")
+local gtable   = require("gears.table")
+local beautiful= require("beautiful")
+local alayout  = require("awful.layout")
+local surface  = require("gears.surface")
+
+local module = {}
+
+local default_buttons = gtable.join(
+    abutton({ }, 1, function(a) a.callback() end)
+)
+
+local function wb_label(item, _, tb)
+    local selected = alayout.get(item.screen) == item.layout
+
+    -- Apply the built-in customization
+    local bg, fg, shape, shape_bw, shape_bc
+
+    -- The layout have only 2 states: normal and selected
+    if selected then
+        bg = item.style.bg_selected or
+            beautiful.layoutlist_bg_selected or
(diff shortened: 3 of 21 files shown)


docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
 
 @DOC_layout_WIDGET_LIST@
 
-## Placing widgets
+### The different type of widget boxes (Wibox)
+
+Awesome API the word "wibox" to describe an area of the screen filled with
```
psychon: This sentence... reads weirdly, dunno. Should it be `The Awesome API`, `In the` or something like that? `uses the word`?

How about just mentioning that `wibox` means `widget box`, so a box that contains widgets?

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+widgets. There is many subvariants of wiboxes with specialized role such as
```
psychon: There _are_ many

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+The normal `wibox` is the base class for each of these type. It is extremely
+flexible and allows to place just about anything on the screen. However it
+requires a lot of repetitive boiler plate code to use directly. For example,
+the user need to compute the optimal size by hand or use `awful.placement`.
+
+The `awful.wibar` specialization allows to attach a `wibox` to a screen edge
+and prevent clients from using this area when tiled.
+
+The `awful.popup` allows to easily place widgets on the screen. It automatically
+resize itself to fit the optimal widget size. It also have helper properties
```
psychon: have -> has

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+and method to make it easy to place it on the screen. It both absolute
```
psychon: method -> methods

psychon: it both _supports_ absolute

(Also, I wouldn't use `both` on a list with three entries, how about `It supports all of` or just `It supports absolute, relative and manual positioning`?)

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+positioning, relative positioning and manual positioning.
+
+The `awful.tooltip` is a very simple `wibox` that allows to display text next
+to an object such as the mouse.
+
+Finally, the `awful.titlebar`, while not technically a real `wibox`, act
```
psychon: act_s_

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+exactly the same way and allows to attach widgets on each side of the clients.
```
psychon: of clients

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+
+## The different syntaxes to initiate widgets
+
+Awesome provides 2 totally different API access style to manage widgets. Both
```
psychon: style_s_

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+suit different use case and are execute exactly the same code under the hood.
```
psychon: What does "are execute" mean? `and are based on the same code`?

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
 
 @DOC_layout_WIDGET_LIST@
 
-## Placing widgets
```
psychon: If you want, I can produce a fixup commit for the following comments, although some of them are questions instead of just plain suggestions.

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+Finally, the `awful.titlebar`, while not technically a real `wibox`, act
+exactly the same way and allows to attach widgets on each side of the clients.
+
+## The different syntaxes to initiate widgets
+
+Awesome provides 2 totally different API access style to manage widgets. Both
+suit different use case and are execute exactly the same code under the hood.
+
+The **imperative** widget initialization is similar to QtWidgets, GTK and Win32.
+You create the object, the set the property and add the widget as a child to
```
psychon: the -> then
(Also `the property` "feels" weird to me, `set properties`?)

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+another already declared widget. It is quite simple to use but very verbose
+and full of boilerplate code. The imperative API also offers property both with
```
psychon: propert_ies_

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+accessors or directly. It is useful when creating highly dynamic layouts where
+widgets are added and removed over the course of their lifecycle.
+
+The **declarative** syntax resembles HTML style code
+written in JSON or YAML. The widgets instances are created automatically and
```
psychon: HTML written in JSON or YAML?!?!?!?!?!?!?!?!?!?

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+the hierarchy is related to the indentation. It is preferred when creating
```
psychon: I can also use the imperative system together with "funny" indentation. :-)
How about `table nesting` instead? I'm nut quite sure...

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+static layouts that wont change over the course of their lifecycle.
```
psychon: won't

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+
+**Imperative with accessors**
+
+Code:
+
+    local bg = wibox.container.background()
+    bg:set_bg("#ff0000")
+
+    local tb1 = wibox.widget.textbox("foo")
+    local tb2 = wibox.widget.textbox("bar")
```
psychon: How about using `tb1:set_text("foo")` instead, just so that there are more differences between the styles?

docs/03-declarative-layout.md
```diff
@@ -29,7 +28,116 @@ configurable rules.
...
+        bg = "#ff0000",
+        widget = wibox.container.background
+    }
+
+
+The Awesome documentation mostly uses the declarative style for consistency,
+but both are **always** available. Note that each style can be mixed with other
+styles, but this creates very confusing code and should be avoided.
+
+## Creating and placing widgets using the declarative
```
psychon: using the declarative what? style?

tests/examples/awful/popup/wiboxtypes.lua
```diff
@@ -0,0 +1,237 @@
...
+    },
+    border_color = beautiful.border_color,
+    border_width = beautiful.border_width,
+    placement    = awful.placement.centered,
+    ontop        = true,
+    shape        = gears.shape.rounded_rect
+}
+
+-- poor copy of awful.widget.calendar_widget until I fix its API to be less
+-- insane.
```
psychon: I love how everyone loves this :-)

lib/awful/widget/layoutlist.lua
```diff
@@ -0,0 +1,151 @@
...
+-- @tparam screen args.screen A screen (mandatory)
+-- @treturn widget The action widget.
+-- @function naughty.widget.layoutlist
+
+local function new(_, args)
+    args = args or {}
+
+    local wdg = wibox.widget.base.make_widget(nil, nil, {
+        enable_properties = true,
+    })
```
psychon: Shouldn't this set `_private.layout` somewhere? The only place setting that seems to be `set_base_layout`?!?

lib/awful/widget/layoutlist.lua
```diff
@@ -0,0 +1,151 @@
...
+
+    gtable.crush(wdg, layoutlist, true)
+
+    reload_cache(wdg)
+
+    wdg._private.data = {}
+    gtable.crush(wdg, args)
+
+    tag.connect_signal("property::layout", function() reload_cache(wdg); update(wdg) end)
```
psychon: Please wire up this widget to `test-leaks.lua` and then make the widget somehow GC'able.

lib/awful/widget/layoutlist.lua
```diff
@@ -0,0 +1,151 @@
...
+end
+
+local function update(self)
+    if not self._private.layout then return end
+    awcommon.list_update(
+        self._private.layout,
+        default_buttons,
+        wb_label,
+        self._private.data,
+        self._private.cache,
```
psychon: Why is the cache necessary here? `gears.surface.load` has its own cache and besides that `reload_cache` doesn't do anything expensive. Just use `alayout.layouts` directly here and do "the necessary stuff" in `wb_label`.

lib/awful/widget/layoutlist.lua
```diff
@@ -0,0 +1,151 @@
...
+local gtable   = require("gears.table")
+local beautiful= require("beautiful")
+local alayout  = require("awful.layout")
+local surface  = require("gears.surface")
+
+local module = {}
+
+local default_buttons = gtable.join(
+    abutton({ }, 1, function(a) a.callback() end)
+)
```
psychon: Correct me if I'm wrong, but this is not the default, but the hardcoded list of buttons that cannot be overridden. Plus, what does this button do anyway? `callback` is always an empty function?!?

lib/awful/widget/layoutlist.lua
```diff
@@ -0,0 +1,151 @@
...
+
+local function wb_label(item)
+    local selected = alayout.get(item.screen) == item.layout
+    return item.layout.name, selected and beautiful.bg_focus or nil, nil, item.icon
```
psychon: Shouldn't there be a `beautiful.bg_layoutlist` in here?

lib/awful/widget/layoutlist.lua
```diff
@@ -0,0 +1,151 @@
...
+    return wibox.widget.base.fit_widget(self, context, self._private.layout, width, height)
+end
+
+--- Create an action list.
+--
+-- @tparam table args
+-- @tparam widget args.layout The widget layout (not to be confused with client
+--  layout).
+-- @tparam[opt] table args.widget_template A custom widget to be used for each action.
+-- @tparam screen args.screen A screen (mandatory)
```
psychon: Some `assert` should check for this mandatory argument, I think

tests/examples/awful/widget/layoutlist/bar.lua
```diff
@@ -0,0 +1,62 @@
+local awful = { --DOC_HIDE --DOC_NO_USAGE
```
psychon: Could you please also do a test showing the "default behaviour" of the layoutlist, i.e. only the minimum required arguments are given (`screen = 1`, I think). That way, (a) I'm more convinced that the widget actually works by default and (b) people get an idea what the widget actually looks like by default.

tests/examples/shims/beautiful.lua
```diff
@@ -34,6 +34,8 @@ local module = {
     border_color = "#6181FF"  ,
     border_width = 1.5        ,
 
+    prompt_bg_cursor = "#00FF7D",
+
```
psychon: Huh? Got some details for what this commit fixes exactly? For example, an error message with traceback? How is this related with 8cc469322f8043b262b36dd7743bb330bdde753d?

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+
+    if not geo then return end
+
+    local preferred_positions = self._private.preferred_directions or {
+        "right", "left", "top", "bottom"
+    }
+
+    local preferred_anchors = self._private.preferred_anchors or {
+        "back", "front", "middle"
+    }
```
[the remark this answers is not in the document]
Elv13: done

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+    })
+
+    if pos_name ~= self._private.current_position then
+        self:emit_signal("property::current_position", pos_name)
+        self._private.current_position = pos_name
+    end
+
+    if anchor_name ~= self._private.current_anchor then
+        self:emit_signal("property::current_anchor", anchor_name)
+        self._private.current_anchor = anchor_name
```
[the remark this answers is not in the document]
Elv13: whoops, copy paste gone wrong, done

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+        return 0, 0
+    end
+
+    return wibox.widget.base.fit_widget(self, context, self.widget, width, height)
+end
+
+-- Layout this widget
+function main_widget:layout(context, width, height)
+    if self.widget then
+        local w, h = wibox.widget.base.fit_widget(self, context, self.widget, 9999, 9999)
```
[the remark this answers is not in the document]
Elv13: > but I feel like they should do so way "harder"

What do you have in mind? I don't feel like adding memes to the doc ;)

> How about using context.screen.geometry.width/height here instead?

I could add a new `awful.screen` function to get the maximum width/height of the screens. I can't just use the current screen because it can change in some async events.

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+        glib.idle_add(glib.PRIORITY_HIGH_IDLE, function()
+
+            local prev_geo = self._wb:geometry()
+            self._wb.width  = math.max(1, math.ceil(w or 1))
+            self._wb.height = math.max(1, math.ceil(h or 1))
+
+            if self._wb.width ~= prev_geo.width and self._wb.height ~= prev_geo.height then
+                set_position(self._wb)
+            end
+        end)
```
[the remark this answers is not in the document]
Elv13: > Why does it need to reach for idle_add instead of gears.timer.delayed_call?

The code is very old, I updated it.

> Why does it need to be delayed at all and cannot be done immediately?

An image is worth a thousand word (and you might actually not believe me)

screenshot_20180115_021433

My theory is that the xcb surface is not reallocated yet

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+    return {self.widget}
+end
+
+-- Replace the layout children
+-- This layout only accept one children, all others will be ignored
+-- @tparam table children A table composed of valid widgets
+function main_widget:set_children(children)
+    self.widget = children and children[1]
+    self:emit_signal("widget::layout_changed")
+end
```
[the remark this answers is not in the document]
Elv13: I think it would add some boilerplate of it's own. There's only the `:fit()` boilerplate. I don't see this as an issue.

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+-- @see awful.placement.next_to
+-- @see awful.popup.preferred_positions
+-- @see awful.popup.preferred_anchors
+-- @treturn table The new geometry
+function popup:move_next_to(obj)
+    if self._private.is_relative == false then return end
+
+    self._private.widget_geo = obj
+
+    if not obj then return end
```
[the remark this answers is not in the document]
Elv13: done

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+
+    if obj._apply_size_now then
+        obj:_apply_size_now(false)
```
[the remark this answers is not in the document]
Elv13: It is necessary because popups can be placed in relation to each other (like the notifications) and at some point it is necessary to collapse all the async stuff. But the async stuff are also necessary for some other corner case like the screenshot above. Without the async part, it can actually crash Awesome.

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+    if type(offset) == "number" then
+        offset = {
+            x = offset,
+            y = offset,
+        }
+    end
+
+    local oldoff = self._private.offset or {}
+    local oldx   = oldoff.x or 0
+    local oldy   = oldoff.y or 0
```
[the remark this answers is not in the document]
Elv13: Because X could be set and not Y? Anyway, I refactored it so it cannot happen. It renders the comment below invalid.

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+
+    self._private.placement = f
+    self:_apply_size_now(false)
+
+    -- Update the position
+    set_position(self)
+end
+
+-- For the tests and the race condition when 2 popups are placed next to each
+-- other.
```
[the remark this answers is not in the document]
Elv13: It will crash

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+    self._private.widget = wid
+    self._private.container:set_widget(wid)
+end
+
+function popup:get_widget()
+    return self._private.widget
+end
+
+--- A brilliant idea to totally turn the whole hierarchy on its head
+-- and create a widget that own a wibox...
```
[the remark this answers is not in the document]
Elv13: Whoops, leftover from the prototype (it was very, very hacky back then, way more than the still rather hacky code)

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+    if args.visible ~= false then
+        w.visible = true
+    end
+
+    return w
+end
+
+--@DOC_wibox_COMMON@
+
+return setmetatable(module, {__call = create_auto_resize_widget})
```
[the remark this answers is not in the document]
Elv13: done

Elv13:
I added a couple commits to fix some of the `popup` issues reported by @psychon. I also added `minimum` and `maximum` properties to better address the "this will explode if the widget size diverge" issue.

lib/awful/popup.lua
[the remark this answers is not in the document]
psychon: Uhm... Do you happen to know an easy way to reproduce this? At least the following does not work.
```lua
gears.timer.start_new(2, function()
	local w = wibox{ x = 50, y = 50, width = 100, height = 10, bg = "#ff0000", fg = "#000000" }
	w:set_widget(wibox.widget.textbox("Foo"))
	w.visible = true
		w.height = 50
	end)
end)
```

lib/awful/popup.lua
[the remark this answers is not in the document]
psychon: Well, that and `set_children`, `get_children`. Even if it just saves 10 lines, that's 10 lines less code. ;-)

lib/awful/popup.lua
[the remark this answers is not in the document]
Elv13: If you apply the patches, it is quite easy to reproduce by commenting the `delayed_call` line. I am sorry for not having a smaller test case. I know about this issue for a years, but I always considered it irrelevant given the kind of hack you have to use to cause it (the popup is such an hack, Radical was another).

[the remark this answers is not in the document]
psychon: Ah. Stderr says (however, I did not get visual corruption, only this)
```
2018-01-15 10:03:12 E: awesome: Error during a protected call: lib/wibox/drawable.lua:165: Cairo context entered error state: SURFACE_FINISHED
stack traceback:
	[C]: in function 'assert'
	lib/wibox/drawable.lua:165: in upvalue 'do_redraw'
	lib/wibox/drawable.lua:390: in function <lib/wibox/drawable.lua:388>
	[C]: in function 'xpcall'
	(...tail calls...)
	lib/gears/timer.lua:231: in function <lib/gears/timer.lua:229>
```
This is the opposite of your theory: The problem is that the xcb surface is reallocated while Lua code still wants to draw to it. Thus, at the end after all the drawing it noticed that something is wrong. (The above happens just by doing `awful.popup{ widget = wibox.widget.textbox("textbox") }`).

So, basically: You may not resize the wibox that is currently drawn while it is being drawn.

...which brings us back to this ugly delayed call and all of its problems.

How much would you be against the popup creating its own `wibox.hierarchy` instance and using that for catching relayouts instead of this magic in `:layout`? That way, this would be disconnected from the widget drawing code instead of being a side effect of it. (And then both your `:fit` and `:layout` are boilerplate and using the argument to `wibox.widget.base.make_widget` makes even more sense... well, no, your whole `main_widget` becomes unnecessary and instead you can just directly give the widget to the wibox)

I wanted to hack this idea into your code, but got lost, so instead here is it in self-contained:
```lua
local widget = wibox.widget.textbox("textbox")
-- Resize the textbox in 5 seconds
gears.timer.start_new(5, function() widget:set_text("This is a longer text than what is was before") end)

local set_widget
do
        local wrapper = wibox.widget.base.make_widget() -- This only needs :layout(), other functions are never used
        local widget = widget
        function wrapper:layout(context, width, height)
                return { wibox.widget.base.place_widget_at(widget, 0, 0,
                        wibox.widget.base.fit_widget(self, context, widget, width, height)) }
        end
        local w = wibox{ x = 50, y = 50, visible = true, widget = widget } -- This displays widget directly, not wrapper!
        local ctxt = { meh_need_to_fake_a_context = true, dpi = screen.primary.dpi }
        local h
        local function layout_callback()
            h:update(ctxt, wrapper, 2^16-1, 2^16-1)
            print(h:get_children()[1]:get_size())
            w.width, w.height = h:get_children()[1]:get_size()
        end
        h = wibox.hierarchy.new(ctxt, wrapper, 2^16-1, 2^16-1, function() end, layout_callback, {})
        layout_callback()
        function set_widget(arg)
                w.widget = arg
                widget = arg
                wrapper:emit_signal("widget::layout_changed")
        end
end

gears.timer.start_new(10, function() set_widget(wibox.widget.textbox("Show a different widget")) end)
```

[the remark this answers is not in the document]
psychon: Or alternatively with `:layout` resizing the wibox again directly:
```lua
local widget = wibox.widget.textbox("textbox")
-- Resize the textbox in 5 seconds
gears.timer.start_new(5, function() widget:set_text("This is a longer text than what is was before") end)

local set_widget
do
        local wrapper = wibox.widget.base.make_widget() -- This only needs :layout(), other functions are never used
        local widget = widget
        local w = wibox{ x = 50, y = 50, visible = true, widget = widget } -- This displays widget directly, not wrapper!
        function wrapper:layout(context, width, height)
                local wi, h = wibox.widget.base.fit_widget(self, context, widget, width, height)
                w.width = wi
                w.height = h
                return { wibox.widget.base.place_widget_at(widget, 0, 0, wi, h) }
        end
        local ctxt = { meh_need_to_fake_a_context = true, dpi = screen.primary.dpi }
        local h
        local function layout_callback()
            h:update(ctxt, wrapper, 2^16-1, 2^16-1)
        end
        h = wibox.hierarchy.new(ctxt, wrapper, 0, 0, function() end, layout_callback, {})
        layout_callback()
        function set_widget(arg)
                w.widget = arg
                widget = arg
                wrapper:emit_signal("widget::layout_changed")
        end
end

gears.timer.start_new(10, function() set_widget(wibox.widget.textbox("Show a different widget")) end)
```

lib/awful/widget/layoutlist.lua
Elv13: ah, indeed, I forgot to actually implement it, done

lib/awful/widget/layoutlist.lua
Elv13: done (doubling the file size... whatever)

lib/awful/widget/layoutlist.lua
Elv13: The other `awful.widget.common` widgets work like that. I could be changed later on.

lib/awful/widget/layoutlist.lua
Elv13: done

lib/awful/widget/layoutlist.lua
Elv13: done

tests/examples/awful/widget/layoutlist/bar.lua
Elv13: done

Elv13:
Note that I added a new commit https://github.com/awesomeWM/awesome/pull/2154/commits/3a58777b8066d70a33464b65eef9e072f5b8fbee to try to fix the build on Lua 5.1. But I still can't reproduce and it doesn't fix it. Anyhow, not using `next()` and `pairs()` is probably a good thing in itself so I will leave it there.

docs/03-declarative-layout.md
psychon: Okay:

The **declarative** syntax resembles HTML-style code in that an explicit nesting is present and is used to express "boxes containing boxes" (the box model). However, instead of requiring (heavy-weight) XML, a lighter JSON-like syntax is used.

Elv13:
Rebased to remove the duplicate commits from the titlebar doc PR (as they were required to pass the test). Otherwise there is still the `awful.popup` way to updating it's size that remains to be done. There is no really valuable way to split this as all tests of other components require the popup class.

lib/awful/popup.lua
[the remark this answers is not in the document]
Elv13: Umm, I kind of like the idea, but my limited attempts created a result more problematic than the current version. I guess "I got lost" the same way you did when you tried to implement this.

tests/examples/awful/template.lua
```diff
@@ -164,43 +187,45 @@ end
...
 -- Loop each clients geometry history and paint it
 for _, c in ipairs(client.get()) do
-    local pgeo = nil
-    for _, geo in ipairs(c._old_geo) do
-        if not geo._hide then
-            total_area:add_at(
-                client_widget(c, c.color or geo._color or beautiful.bg_normal, geo._label),
-                {x=geo.x, y=geo.y}
-            )
+    if not c.minimized then
```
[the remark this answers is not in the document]
Elv13: done (sorry for all the "done" comments I will add, it just hard to keep track of what's done in GitHub given how arbitrary it's "outdated comment" feature is)

tests/examples/shims/client.lua
```diff
@@ -120,6 +120,20 @@ function client.gen_fake(args)
     -- Set the attributes
     ret.screen = args.screen or screen[1]
 
+    -- Good enough for the geometry and border
+    ret.drawin = ret
+    ret.drawable = ret
+
+    -- Make sure the layer properties are not `nil`
```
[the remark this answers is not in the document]
Elv13: Everything that use `if foo == false then ... end` code doesn't work, as simple as that

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+    wdg = wibox.widget.base.make_widget_from_value(wdg)
+
+    self._private.widget = wdg
+
+    self._private.container:set_widget(wdg)
+end
+
+function popup:set_widget(wid)
+    self._private.widget = wid
+    self._private.container:set_widget(wid)
```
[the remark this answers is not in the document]
Elv13: done

lib/awful/popup.lua
```diff
@@ -0,0 +1,434 @@
...
+-- @tparam table|number args.offset The X and Y offset compared to the parent object
+-- @tparam boolean args.hide_on_right_click Whether or not to hide the popup on
+--  right clicks.
+-- @function awful.popup
+local function create_auto_resize_widget(_, args)
+    assert(args)
+
+    -- Temporarily remove the widget
+    local original_widget = args.widget
+    args.widget = nil
```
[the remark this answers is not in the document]
Elv13: done

lib/awful/widget/layoutlist.lua
Elv13: done

lib/wibox/init.lua
```diff
@@ -253,6 +253,10 @@ local function new(args)
     ret._drawable = wibox.drawable(w.drawable, { wibox = ret },
         "wibox drawable (" .. object.modulename(3) .. ")")
 
+    function ret._drawable.get_wibox()
+        return ret
+    end
```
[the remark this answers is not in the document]
Elv13: > (And I fail to see how this commit has anything to do with find_widgets even though the commit 
> summary says that the commit is all about that)

Because the clickable or mouse_overable widgets callback information comes from `find_widgets` and if that's used to move a wibox then it has to know the wibox. I agree it's a bit hacky, but fixing that would be a bit complex unless I missed something obvious when I wrote that (didn't look recently). Feel free to propose a better solution, but I don't have much AwesomeWM time and a backlog of > 500 commits to finish, so it's a bit hopeless to expect me working on this in any kind of timely fashion.

Elv13:
@psychon Any idea about the race condition that causes the popup to move to the wrong side (the one that makes the test fail)?

I added some `print()` in some older revision of the PR and it seems random. However since it doesn't happen without this PR, I guess it has to be related to the `awful.placement.next_to` changes. It happens in some commit, not others and not all the same ones each time. It also only seem to happen on luajit since I 1) can't reproduce at all 2) had this PR run 35+ times on Travis and never saw the issue in the other builders.

Also, I think I am mostly done with the other changes. A fresh review would be welcome. I didn't change much since 3 months, it's mostly a rebase with minor documentation comment changes.

psychon:
> @psychon Any idea about the race condition that causes the popup to move to the wrong side (the one that makes the test fail)?

On Travis, only the LuaJIT build failed. And in fact, it failed because of `TEST_PREV_COMMITS=1`. And here things start to go weird. The summary at the end claims that the tests failed for commit aaa0d92, but when looking at the attempted build with that commit, it succeeded. Oh, great, a bug in the output of this thing... #2261 fixes that (hopefully, I cannot really test it).
Uhm, and from a quick look, this PR does not even touch that test...? Sorry, but debugging `awful.placement` is beyond my abilities.

tests/examples/awful/template.lua
```diff
@@ -154,6 +159,24 @@ local function client_widget(c, col, label)
 end
 
 -- Add all wiboxes
+
+-- Fix the wibox geometries that have a dependency on their content
+for _, d in ipairs(drawin.get()) do
+    local w = d.get_wibox and d:get_wibox() or nil
+    if w then
+        -- Force a full layout first as widgets with as the awful.popup have
+        -- interdependencies between the content and the container
```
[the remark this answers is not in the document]
psychon: I still fail to parse this comment as English.
- "Force a full layout first": Okay, so before anything else is done, something(?) is (re-?)layouted first.
- "as widgets": Because of widgets
- "with as the": Sorry, I'm lost

This is the first commit that I look at and just the second hunk in the diff...

tests/examples/awful/template.lua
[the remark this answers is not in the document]
psychon: I don't mind "done" comments. In fact, I think they are a really good idea and should be used more often. :-)

psychon:
Okay... so, I built lgi against LuaJIT:
```diff
diff --git a/lgi/Makefile b/lgi/Makefile
index 08419bd..e1da776 100644
--- a/lgi/Makefile
+++ b/lgi/Makefile
@@ -43,6 +43,7 @@ endif
 ifeq ($(HOST_OS),darwin)
 CFLAGS += -DGOBJECT_INTROSPECTION_LIBDIR=\"$(GOBJECT_INTROSPECTION_LIBDIR)\"
 endif
+LUA_CFLAGS = -I /usr/include/luajit-2.1/
 ALL_CFLAGS = $(CCSHARED) $(COPTFLAGS) $(LUA_CFLAGS) $(shell $(PKG_CONFIG) --cflags $(PKGS)) $(CFLAGS)
 LIBS += $(shell $(PKG_CONFIG) --libs $(PKGS))
 ALL_LDFLAGS = $(LIBFLAG) $(LDFLAGS)
```
Then I built awesome against LuaJIT (we might want to simplify this...):
```diff
diff --git a/awesomeConfig.cmake b/awesomeConfig.cmake
index 892b88e6..454c272e 100644
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -358,7 +358,7 @@ file(MAKE_DIRECTORY ${BUILD_DIR}/script_files/)
 
 add_custom_command(
         OUTPUT ${BUILD_DIR}/docs/06-appearance.md
-        COMMAND lua ${SOURCE_DIR}/docs/06-appearance.md.lua
+        COMMAND luajit ${SOURCE_DIR}/docs/06-appearance.md.lua
         ${BUILD_DIR}/docs/06-appearance.md
         DEPENDS lgi-check-run
 )
diff --git a/tests/examples/CMakeLists.txt b/tests/examples/CMakeLists.txt
index 124e5b1b..88ab9810 100644
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -17,7 +17,7 @@ endif()
 cmake_minimum_required(VERSION 3.0.0)
 
 # Get and update the LUA_PATH so the scripts can be executed without awesome.
-execute_process(COMMAND lua -e "p = package.path:gsub(';', '\\\\;'); io.stdout:write(p)"
+execute_process(COMMAND luajit -e "p = package.path:gsub(';', '\\\\;'); io.stdout:write(p)"
     OUTPUT_VARIABLE "LUA_PATH_")
 
 # Make sure the system can be called from the test directory
@@ -32,7 +32,7 @@ endif()
 
 if (DO_COVERAGE)
   execute_process(
-    COMMAND lua -e "require('luacov.runner')('${TOP_SOURCE_DIR}/.luacov')"
+    COMMAND luajit -e "require('luacov.runner')('${TOP_SOURCE_DIR}/.luacov')"
     RESULT_VARIABLE TEST_RESULT
     ERROR_VARIABLE  TEST_ERROR
     ERROR_STRIP_TRAILING_WHITESPACE)
@@ -40,9 +40,9 @@ if (DO_COVERAGE)
     message(${TEST_ERROR})
     message(FATAL_ERROR "Failed to run luacov.runner.")
   endif()
-  set(LUA_COV_RUNNER lua "-erequire('luacov.runner')('${TOP_SOURCE_DIR}/.luacov')")
+  set(LUA_COV_RUNNER luajit "-erequire('luacov.runner')('${TOP_SOURCE_DIR}/.luacov')")
 else()
-  set(LUA_COV_RUNNER lua)
+  set(LUA_COV_RUNNER luajit)
 endif()
 
 # Add the main awesome lua libraries.
```
Plus: `LUA_CPATH="/tmp/lgi/?.so;;" LUA_PATH="/tmp/lgi/?.lua;/tmp/lgi/?/init.lua;;" make CMAKE_ARGS="-DLUA_LIBRARY=/usr/lib64/libluajit-5.1.so -DLUA_INCLUDE_DIR=/usr/include/luajit-2.1"`

According to Travis, the following should fail now:
```
git checkout 84c2c961f22bc48423874818f3a64568f66be092
LUA_CPATH="/tmp/lgi/?.so;;" LUA_PATH="/tmp/lgi/?.lua;/tmp/lgi/?/init.lua;;"  tests/run.sh /tmp/awesome/tests/test-tooltip.lua
```
And... well, it does fail. So now comes the debugging fun, which takes forever thanks to having to run `make`, which rebuilds all the docs all the time.

Okay, so here is a patch that adds lots of debugging output:
```diff
diff --git a/lib/awful/placement.lua b/lib/awful/placement.lua
index 4499c5eb..7fd9c72b 100644
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -1386,24 +1386,46 @@ function placement.next_to(d, args)
 
     local regions = get_relative_regions(wgeo, mode, is_absolute)
 
+    require("gears.debug").dump({
+        input_geo = wgeo,
+        is_absolute = is_absolute,
+        regions = regions
+    }, nil, 2)
+    require("gears.debug").dump(regions)
+
     -- Check each possible slot around the drawable (8 total), see what fits
     -- and order them by preferred_positions
     local does_fit = {}
     for k,v in pairs(regions) do
+        print("Checking", k, v)
         local geo, dir = outer_positions[k.."1"](v, dgeo.width, dgeo.height)
         geo.width, geo.height = dgeo.width, dgeo.height
         local fit = fit_in_bounding(v.screen, geo, args)
 
+        require("gears.debug").dump {
+            geo = geo,
+            dir = dir,
+            fit = fit
+        }
+
         -- Try the other compatible geometry
         if not fit then
+            print("Trying compatible geometry")
             geo, dir = outer_positions[k.."2"](v, dgeo.width, dgeo.height)
             geo.width, geo.height = dgeo.width, dgeo.height
             fit = fit_in_bounding(v.screen, geo, args)
+
+            require("gears.debug").dump {
+                geo = geo,
+                dir = dir,
+                fit = fit
+            }
         end
 
         does_fit[k] = fit and {geo, dir} or nil
 
         if fit and preferred_positions[k] and preferred_positions[k] < pref_idx then
+            print("Preferred!")
             pref_idx  = preferred_positions[k]
             pref_name = k
         end
diff --git a/lib/awful/tooltip.lua b/lib/awful/tooltip.lua
index 36f86933..6b37f9f9 100644
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -181,6 +181,11 @@ local function apply_mouse_mode(self)
     local align          = self._private.align
     local real_placement = align_convert[align]
 
+    require("gears.debug").dump{ mode = "mouse",
+    align = self._private.align,
+    real_placement = align_convert[align]
+}
+
     a_placement[real_placement](w, {
         parent = capi.mouse,
         offset = offset[align]
@@ -190,12 +195,17 @@ end
 local function apply_outside_mode(self)
     local w = self:get_wibox()
 
+    require("gears.debug").dump({ mode = "outside",
+    geometry = self._private.widget_geometry,
+    prefered = self.preffered_positions
+}, nil, 2)
     local _, position = a_placement.next_to(w, {
         geometry            = self._private.widget_geometry,
         preferred_positions = self.preferred_positions,
         honor_workarea      = true,
     })
 
+    print("Chose position:", position)
     if position ~= self.current_position then
         -- Re-apply the shape.
         apply_shape(self)
diff --git a/tests/test-tooltip.lua b/tests/test-tooltip.lua
index d545ec56..e9159cb5 100644
--- a/tests/test-tooltip.lua
+++ b/tests/test-tooltip.lua
@@ -89,17 +89,20 @@ table.insert(steps, function()
 
     tt:add_to_object(imb)
 
+    print("Moving the mouse")
     -- Move the mouse over the widget
     place.top_left(mouse)
     mouse.coords {
         x = w.x + w.width - 20 - 12.5,
         y = w.y + 125 + 12.5,
     }
+    print("Moved mouse to x=", w.x + w.width - 20 - 12.5, " and y=", w.y + 125 + 12.5)
     return true
 end)
 
 -- Test that the above move had the intended effect
 table.insert(steps, function()
+    print("Checking current position is top")
     assert(tt.current_position == "top", tt.current_position)
 
     return true
```
The relevant part of the output is:
```
Moving the mouse
Moved mouse to x=	604.5	 and y=	396.5
table: 0x41d8ad60
  geometry : table: 0x41d8a5b0
    x : 205 (number)
    widget : @lib.wibox.widget.imagebox (table: 0x41f8da10) (table) […]
    y : 125 (number)
    widget_height : 25 (number)
    hierarchy : table: 0x41f2a0a0 (table) […]
    widget_width : 25 (number)
    width : 25 (number)
    drawable : wibox drawable (@.tmp.awesome.tests.test-tooltip) (table: 0x405938f0) (table) […]
    height : 25 (number)
  mode : outside (string)
table: 0x41d8cb30
  input_geo : table: 0x41d8a5b0
    x : 205 (number)
    widget : @lib.wibox.widget.imagebox (table: 0x41f8da10) (table) […]
    y : 125 (number)
    widget_height : 25 (number)
    hierarchy : table: 0x41f2a0a0 (table) […]
    widget_width : 25 (number)
    width : 25 (number)
    drawable : wibox drawable (@.tmp.awesome.tests.test-tooltip) (table: 0x405938f0) (table) […]
    height : 25 (number)
  is_absolute : false (boolean)
  regions : table: 0x41d8c9b8
    right : table: 0x41d8caa8 (table) […]
    left : table: 0x41d8ca48 (table) […]
    bottom : table: 0x41d8cb68 (table) […]
    top : table: 0x41d8cb08 (table) […]
table: 0x41d8c9b8
  right : table: 0x41d8caa8
    x : 1 (number)
    screen : screen: 0x419df8e0 (screen)
    y : 125 (number)
    distance : 216.86055427394 (number)
    height : 25 (number)
    width : 25 (number)
  left : table: 0x41d8ca48
    x : 0 (number)
    screen : screen: 0x419df8e0 (screen)
    y : 125 (number)
    distance : 217.85889929034 (number)
    height : 25 (number)
    width : 25 (number)
  bottom : table: 0x41d8cb68
    x : 205 (number)
    screen : screen: 0x419df8e0 (screen)
    y : 1 (number)
    distance : 137.07114940789 (number)
    height : 25 (number)
    width : 25 (number)
  top : table: 0x41d8cb08
    x : 205 (number)
    screen : screen: 0x419df8e0 (screen)
    y : 0 (number)
    distance : 138.06701271484 (number)
    height : 25 (number)
    width : 25 (number)
Checking	right	table: 0x41d8caa8
table: 0x41d904b0
  geo : table: 0x41d900a0
    y : 125 (number)
    x : 1 (number)
    height : 14 (number)
    width : 62 (number)
  fit : true (boolean)
  dir : down (string)
Checking	left	table: 0x41d8ca48
table: 0x41d90e28
  geo : table: 0x41d90a18
    y : 125 (number)
    x : -62 (number)
    height : 14 (number)
    width : 62 (number)
  fit : false (boolean)
  dir : down (string)
Trying compatible geometry
table: 0x41d91a28
  geo : table: 0x41d913a0
    y : 136 (number)
    x : -62 (number)
    height : 14 (number)
    width : 62 (number)
  fit : false (boolean)
  dir : up (string)
Checking	bottom	table: 0x41d8cb68
table: 0x41d92508
  geo : table: 0x41d920f8
    y : 1 (number)
    x : 205 (number)
    height : 14 (number)
    width : 62 (number)
  fit : false (boolean)
  dir : right (string)
Trying compatible geometry
table: 0x41d930a0
  geo : table: 0x41d92c90
    y : 1 (number)
    x : 168 (number)
    height : 14 (number)
    width : 62 (number)
  fit : false (boolean)
  dir : left (string)
Checking	top	table: 0x41d8cb08
table: 0x41d93be8
  geo : table: 0x41d937d8
    y : -14 (number)
    x : 205 (number)
    height : 14 (number)
    width : 62 (number)
  fit : false (boolean)
  dir : right (string)
Trying compatible geometry
table: 0x41d94540
  geo : table: 0x41d94130
    y : -14 (number)
    x : 168 (number)
    height : 14 (number)
    width : 62 (number)
  fit : false (boolean)
  dir : left (string)
Chose position:	right
Checking current position is top
Error: running function for step 13/16 (@1): /tmp/awesome/tests/test-tooltip.lua:106: right
```
So, top was not chosen because that would lead to `y=-14`, aka partly offscreen.
The same test with Lua 5.3 produces:
Sigh. The same error and pretty much the same output.

@Elv13 Why do you think exactly that this is LuaJIT-specific? How about this being `TEST_PREV_COMMITS=1` specific instead? In fact, I just tried and now `test-tooltip.lua` also fails for the HEAD of this branch with either Lua 5.3 or LuaJIT. But something is definitely wrong, here is a screenshot taken after letting the test run under `TEST_PAUSE_ON_ERRORS=1`:
foo
That tooltip looks a bit... off?

Elv13:
Rebased on top of the merged commits. This PR has 97% coverage.

As said in the v4.3 roadmap comment in another PR, the big remaining item here is the `delayed_call` instead of the hierarchy. I propose to skip this for now even if I agree it is conceptually better. Integrating this change has so far produced regressions and I would appreciate for this to be in v4.3 to allow the code to be used before I propose the `naughty` pull request that makes heavy use of it. I think it can be made to work without the inconsistencies, so I propose to open an issue and do it for v4.4.

Tracking this over such a large PR is hard but I think most comments were handled earlier this year. The tooltip regression that caused some builds to pass and some to fail is still troubling and the root cause wasn't identified. The bug itself was pretty trivial to fix using git-bisect, but I am not really happy about it. There is a chicken and eggs that causes the tooltip to require this pull request before it can have documentation tests to prevent future heisenbugs. I don't want to add more code to this PR because if its huge size. However it can be added to the v4.3 milestone. It's trivial to write once the template and placement commits are merged. The new improvements you added to the documentation artifacts will allow to detect future changes/regressions.

So, @psychon how do you want to go forward about this. I can split it in chunks but they are interdependent so doing this would force all the tests to go in the last PR. So it would leave some `@DOC_...EXAMPLE@` that point to nowhere at first (and no, I wont create commits to add them in that PR, it makes rebasing a mess). Or you can review the whole thing again and (hopefully) be done with it. It hasn't changed beside being rebased a hundred time, so the old review applies.

psychon:
@Elv13 By the way, what happened to the sometimes-broken test? According to the screenshot from my last comment, `awful.placement` is broken. And I think "sometimes randomly broken" is worse than "always broken".

Elv13:
It now exists in smaller PRs

PR #2159, opened 2018-01-17
Add new gears.sort module

Veratil:
After talking to @Elv13 about some things waiting for v4.3+, he mentioned
for #1487 a topological sort is needed.

This adds a new gears module, sort, to the library. This is pulled from
https://github.com/bungle/lua-resty-tsort, which is under the BSD license,
and slightly modified to conform to AwesomeWM code style while adding some
documentation comments.

For future sorting algorithms, if we wish to add them, I've made sort
return a table instead of just having a `gears.tsort`.

### DIFF
--- a/lib/gears/sort.lua
+++ b/lib/gears/sort.lua
@@ -1,0 +1,117 @@
+---------------------------------------------------------------------------
+--- Topological sort function on input data.
+--
+-- @author Aapo Talvensaari
+-- @copyright 2016 Aapo Talvensaari
+-- @module gears.tsort
+---------------------------------------------------------------------------
+
+-- Copyright (c) 2016, Aapo Talvensaari
+-- All rights reserved.
+--
+-- Redistribution and use in source and binary forms, with or without
+-- modification, are permitted provided that the following conditions are met:
+--
+-- * Redistributions of source code must retain the above copyright notice, this
+--   list of conditions and the following disclaimer.
+--
+-- * Redistributions in binary form must reproduce the above copyright notice,
+--   this list of conditions and the following disclaimer in the documentation
+--   and/or other materials provided with the distribution.
+--
+-- THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+-- AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+-- IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+-- DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+-- FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+-- DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+-- SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+-- CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+-- OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+-- OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+-- tsort.lua pulled from GitHub repository:
+-- https://github.com/bungle/lua-resty-tsort
+
+local setmetatable = setmetatable
+local pairs = pairs
+local type = type
+
+local function visit(k, n, m, s)
+    if m[k] == 0 then return 1 end
+    if m[k] == 1 then return end
+    m[k] = 0
+    local f = n[k]
+    for i=1, #f do
+        if visit(f[i], n, m, s) then return 1 end
+    end
+    m[k] = 1
+    s[#s+1] = k
+end
+
+local tsort = {}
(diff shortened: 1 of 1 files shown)


psychon:
From a quick look: Lua 5.3.4 does not have a maximum recursion depth (older versions do), but limits the maximum size of the stack:
```C
/*
@@ LUAI_MAXSTACK limits the size of the Lua stack.
** CHANGE it if you need a different limit. This limit is arbitrary;
** its only purpose is to stop Lua from consuming unlimited stack
** space (and to reserve some numbers for pseudo-indices).
*/
#if LUAI_BITSINT >= 32
#define LUAI_MAXSTACK		1000000
#else
#define LUAI_MAXSTACK		15000
#endif
```
Without using much stack 333k recursive calls seem to be possible:  `lua -e 'function f(i) local l = i + 1 print(l) f(l) return 42 end f(0)`.

So I guess this being recursive is not actually much of an issue.

psychon:
After a quick look at wikipedia (oh, now I see why the code uses this algorithm) and a look at #1487 for what kind of API Elv13 actually needs here:
```lua
local tsort

do
    tsort = {}
    local mt = { __index = tsort }

    function tsort:add_node(node)
        if not self.edges[node] then
            self.edges[node] = {}
        end
    end

    --- Add the information that some nodes must appear before a given node.
    -- @param node The node that edges are added to.
    -- @tparam table dependencies List of nodes that have to appear before `node`.
    function tsort:before(node, dependencies)
        self:add_node(node)
        for _, dep in ipairs(dependencies) do
            self:add_node(dep)
            self.edges[node][dep] = true
        end
    end

    --- Add the information that some nodes must appear after a given node.
    -- @param node The node that edges are added to.
    -- @tparam table dependencies List of nodes that have to appear after `node`.
    function tsort:after(node, dependencies)
        self:add_node(node)
        for _, dep in ipairs(dependencies) do
            self:before(dep, { node })
        end
    end

    local HANDLING, DONE = 1, 2

    local function visit(result, self, state, node)
        if state[node] == DONE then
            -- This node is already in the output
            return
        end
        if state[node] == HANDLING then
            -- We are handling this node already and managed to visit it again
            -- from itself. Thus, there must be a loop.
            result.BAD = node
            return true
        end

        state[node] = HANDLING
        -- Before this node, all nodes that it depends on must appear
        for dep in pairs(self.edges[node]) do
            if visit(result, self, state, dep) then
                return true
            end
        end
        state[node] = DONE
        table.insert(result, node)
    end

    --- Try to sort the nodes.
    -- @treturn[1] table A sorted list of nodes
    -- @treturn[2] nil
    -- @return[2] A node around which a loop exists
    function tsort:sort()
        local result = {}
        local state = {}
        for node in pairs(self.edges) do
            if visit(result, self, state, node) then
                return nil, result.BAD
            end
        end
        return result
    end

    function tsort.new()
        return setmetatable({
            edges = {},
        }, mt)
    end
end

do
    local tsort = tsort:new()
    tsort:after(1, { 2, 3, 4 })
    tsort:after(2, { 3, 4 })
    tsort:after(3, { 4 })
    local res = assert(tsort:sort())
    assert(#res == 4)
    for k, v in pairs(res) do
        assert(k == v)
    end
end

do
    local tsort = tsort:new()
    tsort:before(4, { 1, 2, 3 })
    tsort:before(3, { 1, 2 })
    tsort:before(2, { 1 })
    local res = assert(tsort:sort())
    assert(#res == 4)
    for k, v in pairs(res) do
        assert(k == v)
    end
end

do
    local tsort = tsort:new()
    tsort:after('a', { 'b' })
    tsort:after('b', { 'c' })
    tsort:after('c', { 'd' })
    local res = assert(tsort:sort())
    assert(#res == 4)
    assert(res[1] == 'a')
    assert(res[2] == 'b')
    assert(res[3] == 'c')
    assert(res[4] == 'd')
end

do
    local tsort = tsort:new()
    tsort:before(11, { 5, 7 })
    tsort:after(11, { 2, 9, 10 })
    tsort:before(8, { 7, 3 })
    tsort:after(8, { 9 })
    tsort:after(3, { 10 })

    local res, err = tsort:sort()
    if not res then
        print("Loop around:", err)
    else
        for k, v in ipairs(res) do
            print(k, v)
        end
    end
end
```

I won't claim that the above is better than what this PR adds, but at least I think that the API here better matches what is needed (#1487 has nodes and a list of before/after; with the API in this PR that would require iterating and adding individual edges).

Veratil:
@psychon I do agree your code is much easier to read and understand. @Elv13 needed a topological sort so I found one that was simple instead of writing my own (Elv's suggestion :).

The problem with tests is that topological sort doesn't guarantee the same output, only that the order is correct. If there are multiple nodes without `before` dependencies then any of them could be first in the list. I'm not saying it's impossible to write, just a little bit of a pain. I can work on adding tests soon.

That being said, this was a first step to help Elv out. @psychon, we can replace the entire thing with your code and I'd be fine with that, too. I'd rather not merge this PR until we get all the proper things in place. We can expand upon your code, too, if we want to add a few other ways for adding nodes.

Veratil:
I see @Elv13 has merged @psychon's code to his PR. Shall we close this one?

psychon:
Nah, I'd keep this open. Smaller PRs are easier to handle and if/when this one is merged, that other PR can be rebased and the duplicate commits (hopefully) go away. Of course, if you do not want to continue on this, you are free to close the PR.

Elv13:
Also, the updated #1487 was just to test the unit tests on travis (Busted refuses to work on my computer because of how my Lua is installed). The `awful.rules` part isn't even half ported. I got a newer version on my laptop, but it regresses a bunch of other stuff when abusing the new sorting feature. There is also the issue that the order is non deterministic, unless all graph paths are specified, different Lua version run things in different order. I need to solve that before even deciding if @Veratil or @psychon code is going to get my vote. I don't have unlimited time right now and try to spend it away from my computer ;), don't expect 100 PRs per hour in the coming months. Things are progressing in parallel, but not very fast.

psychon:
> don't expect 100 PRs per hour in the coming months

Thank you very much for that :anguished:

PR #2171, opened 2018-01-29
Use clockget_time and archchart use min_val

treere:
Use clockget_time insted of gettimeofday because it will be removed.

Use min_val in arcchart.

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -61,11 +61,13 @@ awesome_t globalconf;
 /** argv used to run awesome */
 static char **awesome_argv;
 
+#ifndef NDEBUG
 /** time of last main loop wakeup */
 static struct timespec last_wakeup;
 
 /** current limit for the main loop's runtime */
 static float main_loop_iteration_limit = 0.1*1e9;
+#endif
 
 /** Call before exiting.
  */
@@ -402,8 +404,10 @@ static gint
 a_glib_poll(GPollFD *ufds, guint nfsd, gint timeout)
 {
     guint res;
+#ifndef NDEBUG
     struct timespec now;
-    long length;
+		long length;
+#endif
     lua_State *L = globalconf_get_lua_State();
 
     /* Do all deferred work now */
@@ -422,6 +426,7 @@ a_glib_poll(GPollFD *ufds, guint nfsd, gint timeout)
     if (globalconf.pending_event != NULL)
         timeout = 0;
 
+#ifndef NDEBUG
     /* Check how long this main loop iteration took */
     clock_gettime(CLOCK_MONOTONIC, &now);
     length = (now.tv_nsec - last_wakeup.tv_nsec) ;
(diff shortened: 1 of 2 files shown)


treere:
Is `warn("Last main loop iteration took %.6f seconds! Increasing limit for this warning to that value.", length);` still usefull?

Elv13:
Thanks for this

> Is warn("Last main loop iteration took %.6f seconds! Increasing limit for this warning to that value.", length); still useful?

As useful as it ever was. Awesome is single threaded. If an event loop takes more than 1/60 of a second, then you "drop frames" (conceptually, Awesome isn't a compositor). Remember, X11 blocks everything when the WM is busy since the clients are waiting for the result (not exactly, not always, but as a rule of the thumb).

As for the patch, I think the `--TODO` can be removed since you basically did it. I didn't look into the details (it has been a while since I wrote it), but `math.max(min, 0)` isn't really relevant. I think it should be `math.min(min_val, max_val)`. Values smaller than zero make sense in many case. This would also get rid of the `max_val = max_val+1` `if` that's also not universally what people expect. Of course it needs to be tested with negative values first. It can be done either using the `spec/` unit tests or as a documentation example (you need to install `ldoc` to generate them, see `tests/examples/wibox/widget/arcchart` for the ones I already wrote).

treere:
Ok, now it is clear why that warning is there but maybe the float is a bit too heavy.

I will add negative values and some tests/examples.

psychon:
> Use clockget_time insted of gettimeofday because it will be removed.

Wait, what? I seriously doubt that. From `man 2 gettimeofday`:

> CONFORMING TO
>        SVr4, 4.3BSD.  POSIX.1-2001 describes gettimeofday() but not settimeofday().  POSIX.1-2008 marks gettimeofday() as obsolete, recommending the use of clock_gettime(2) instead.

Some interface that is so old will not just be dropped. Either that, or someone is seriously going nuts. O.O.

Also, the commit that does this is wrong, because it ignores the second part of the time. For example, your code will compute that from 2.9 seconds to 3.1 seconds, 0.1-0.9 = -0.8 seconds elapsed, while the actual "answer" is 0.2 seconds.

Oh and according to its man page, using `clock_gettime`requires linking against librt for glibc before 2.17, so this commit breaks compatibility with I-dont-know-how-old systems. I wonder what the situation on e.g. the BSDs here is....

The second commit sounds a lot like premature optimisation (the multiplication will not have any measurable overhead) and makes me notice that `main_loop_iteration_limit` is still a float, even though you turned length into a long.

The third commit looks to me like it should just use `v-min_val` and skip all the calls to max. Garbage in, garbage out. ;-)

And the `if` added in the last commit makes no sense to me. If someone sets min_val to be `>=` max_val, then that's their own fault, isn't it? (And you misspelt Bugfix in the commit message)

treere:
OK, I revert!
New pull request later!

awesome.c
```diff
@@ -402,8 +404,10 @@ static gint
 a_glib_poll(GPollFD *ufds, guint nfsd, gint timeout)
 {
     guint res;
-    struct timeval now, length_time;
-    float length;
+#ifndef NDEBUG
```
Elv13: No, that information is intended for the regular users, not the developers. If you get that warning, fix your config.

treere: Sorry, it was for my PC!

PR #2176, opened 2018-02-03
Add support for resizing notification icon with respect to aspect

Watkurem:
Previously, if an icon was not exactly square, an icon size set in configuration would cause the notification to pad the icon with empty space so dimensions are equal.

Now behaviour is different: the bigger dimension of the icon is scaled to fit the icon_size value, while smaller is scaled same amount to preserve aspect.

Also, ImageSurface is now created not as a fixed size square, but with dimensions computed in similar way.

P.S. I actually don't know Lua, so feel free to educate me on mistakes that I most likely managed to make even in these 3 statements.
illustration

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -759,9 +759,14 @@ function naughty.notify(args)
             iconbox = wibox.widget.imagebox()
             iconmargin = wibox.container.margin(iconbox, margin, margin, margin, margin)
             if icon_size then
-                local scaled = cairo.ImageSurface(cairo.Format.ARGB32, icon_size, icon_size)
+                local scale_factor = icon_size / math.max(icon:get_height(),
+                                                          icon:get_width())
+                local scaled =
+                    cairo.ImageSurface(cairo.Format.ARGB32,
+                                       icon:get_width() * scale_factor,
+                                       icon:get_height() * scale_factor)
                 local cr = cairo.Context(scaled)
-                cr:scale(icon_size / icon:get_height(), icon_size / icon:get_width())
+                cr:scale(scale_factor, scale_factor)
                 cr:set_source_surface(icon, 0, 0)
                 cr:paint()
                 icon = scaled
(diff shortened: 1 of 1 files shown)


lib/naughty/core.lua
```diff
@@ -759,9 +759,14 @@ function naughty.notify(args)
             iconbox = wibox.widget.imagebox()
             iconmargin = wibox.container.margin(iconbox, margin, margin, margin, margin)
             if icon_size then
-                local scaled = cairo.ImageSurface(cairo.Format.ARGB32, icon_size, icon_size)
+                local scale_factor = icon_size / math.max(icon:get_height(),
+                                                          icon:get_width())
+                local scaled =
+                    cairo.ImageSurface(cairo.Format.ARGB32,
+                                       icon:get_width() * scale_factor,
+                                       icon:get_height() * scale_factor)
```
psychon: How likely is this to have rounding issues? E.g. if the number turns out to be 41.9999999999, I am not sure if lgi will translate this into 41 or 42.
It might make sense to add a call to `gears.math.round` around these two arguments, i.e. add a `local gmath = require("gears.math")` at the beginning of the file and add `gmath.round(...)` here.

psychon: It seems to truncate instead of round:
```
$ lua -e 'cairo = require("lgi").cairo s = cairo.ImageSurface(cairo.Format.A1, 0.9999999999, 1) print(s.width, s.height)' 
0	1
```
So, the question remains whether this could be a problem in practice? Of course, a one pixel difference will likely not be noticeable, but still...

Watkurem: Sure, I'll use round here. Also assuming `gears.math` is proper way to use 'math' methods here, so I'll also use `gears.math.max`

psychon: Thanks. :-)
Just for your information: `gears.math` is an awesome-specific thing. Lua has `math.ceil` and `math.floor`, but it has no "round to nearest". `gears.math.round` just does `math.floor(x+0.5)` to get the expected rounding. It does however not have `gears.math.max` since `math.max` is already provided by Lua. Sorry for being so complicated here. ;-)

Watkurem: Got it when trying to use `gears.math.max` gave me an error and I looked into `gears.math` to get an idea of how it's set up. Thanks for trouble explaining though!

Elv13:
Thanks

PR #2180, opened 2018-02-07
dbus: fix nil notifications

gabreal:
when mangling notifications via naughty.config.notify_callback it is
advised (1) to return nil for rejecting a notification. The underlaying dbus
library tries to access a field of that notification table and fails.

(1) https://awesomewm.org/doc/api/libraries/naughty.html#config

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -421,7 +421,7 @@ function naughty.getById(id)
 end
 
 --- Increase notification ID by one
-function get_next_notification_id()
+function naughty.get_next_notification_id()
   counter = counter + 1
   return counter
 end
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -420,6 +420,12 @@ function naughty.getById(id)
     end
 end
 
+--- Increase notification ID by one
+function get_next_notification_id()
+  counter = counter + 1
+  return counter
+end
+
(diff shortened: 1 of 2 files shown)


psychon:
The patch is fine by me 👍 , but I just want to mention that it violates the protocol specifications: https://developer.gnome.org/notification-spec/

> The returned ID is always greater than zero. Servers must make sure not to return zero as an ID. 

No idea what the possible effects of this are. The protocol does not really allow the server to reject notifications...

actionless:
from the spec I would suggest creating a function to expose notification id counter, ie to add to `naughty.core` smth like:
```lua
function get_next_notification_id()
  counter = counter + 1
  return counter
end
```
and use it in this patch:
```diff
+                if notification ~= nil then
+                    return "u", core.get_next_notification_id()
+                end
```

gabreal:
Thanks for your comments but I have not entirely understood the point. You say and it's written in the docs that it should never return 0 but then the statement 3 lines below should also be changed accordingly.

And for this particular place of the code it is not always just incremented but could also be set to the replaces_id i.e. https://github.com/awesomeWM/awesome/blob/master/lib/naughty/core.lua#L658

Shouldn't it be the other way around?

blueyed:
@gabreal 
Not sure about your last comment - you appear to have updated the PR after it already.
Is this good from your side?

PR #2182, opened 2018-02-07
doc: improve gears.table.iterate

blueyed:


lib/gears/table.lua
```diff
@@ -179,10 +179,13 @@ end
...
--- @param filter a function that returns true to indicate a positive match
--- @param start  what index to start iterating from.  Default is 1 (=> start of
--- the table)
+-- @tparam table t The table to iterate.
+-- @tparam func  filter A function that returns true to indicate a positive
+--   match.
+-- @param func.item The item to filter.
+-- @tparam[opt=1] int start Index to start iterating from.
+--   Default is 1 (=> start of the table).
+-- @treturn func
```
blueyed: Should also improve/add a short desc probably?!

Elv13:
Can you add an `@usage` example (if possible in `tests/examples/text/gears/table/`)? Feel free to ignore this comment if you don't have time for this.

blueyed:
@Elv13 
Makes sense.  Will look into it when addressing my comment.

lib/gears/table.lua
actionless: if you know good words please submit as a new PR

actionless:
it's still makes doc better than the current and additional improvements can be done as new PR so i am merged to avoid it to stall

PR #2183, opened 2018-02-07
[RFC] Add gears.table.slice

blueyed:
Had this locally.
Not sure if it should get changed to use `ipairs` and/or to include entries with keys?!


Elv13:
Open question: Should this be a proxy or should by shallow copy (as it is right now)?

Lua allows golang style slice/view using metatable

blueyed:
> Lua allows golang style slice/view using metatable

Please elaborate.
Sounds like it might be good.

blueyed:
Negative indexing would also be good.

Elv13:
If you leave the "proxy" slice table empty but implement the `__index`, `__len` and `__newindex` (and optionally `__pairs` and `__ipairs`) using lambdas, it *should* work in most cases. This can also be used for negative indices.

Veratil:
@blueyed @Elv13 correct me if I'm wrong, but most slice functions in languages are `[start,end)`. `end_` in the function should be exclusive, not inclusive.

Javascript's [is exclusive](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice).

Java's [is exclusive](https://docs.oracle.com/javase/6/docs/api/java/util/Arrays.html#copyOfRange(T[],%20int,%20int)).

Python's is exclusive.
```python
>>> t = [1,2,3]
>>> t[1:1]
[]
>>> t[1:2]
[2]
```

Ruby and PHP use `length` instead of an ending index for their slice methods.

Currently the function doesn't conform to either the exclusive or length options.

blueyed:
Updated to be exclusive.

blueyed:
> Also, is there a valid use case for non-1 stepping in slices?

Likely not.

Proxy might be good, but is not clear to me - it does not change the original, does it?
Then this could be done on top/later (without a new name), no?

PR #2188, opened 2018-02-11
stub support for _NET_WM_FULLSCREEN_MONITORS atom for VMware multiple monitors

alexeicolin:
*NOTE*: I don't expect you will merge this because it's a fake non-compliant claim of supporting the atom. I just want to document the workaround for VMware multiple monitors feature, so that others can find and use it.

Workaround for the VMware multi-monitor function. Without
this patch, VMware complains that "Current host window manager does not
support multiple monitors awesome". Simply claiming support
is enough to trick VMware into opening multiple display windows.

Detailed discussion:

https://github.com/i3/i3/issues/2116
https://communities.vmware.com/thread/198341
https://www.mail-archive.com/awesome-devel@naquadah.org/msg09025.html
[dead] https://awesome.naquadah.org/bugs/index.php?do=details&task_id=1254


psychon:
Do you happen to know how this protocol is actually supposed to work? The specification at https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472531472 talks about indices as seen by the Xinerama which basically makes no sense to me. Since awesome uses RandR, it does not use Xinerema and mapping between the two seems to be basically guesswork.

Also, what should be done when the property contains nonsense, e.g. the right edge is to the left of the left one? Should the WM just ignore the request? If so, would we implement the protocol correctly if we claim to support it and just ignore all requests?

Oh and:

> In the event of a change in monitor configuration, the application is responsible for re-computing the monitors on which it wants to appear. The window manager may continue using the same monitor indices as before or simply clear the list, returning to "normal" fullscreen.

Hahaha! Xinerama does *not* support dynamic monitor configuration. Fixing that is the whole reason why RandR exists.

Oh and some criticism on the protocol was as well given here: https://github.com/i3/i3/issues/2116#issuecomment-165705565

alexeicolin:
Yes, I don't expect that implementing this protocol is a possibility at all. The point is that apparently some software (VMware Workstation) is asking for this support but never actually using it (or if it is trying to use it, there's no error, and the multimonitor feature continues to work fine.)

I just wanted to document this trivial patch out there so that whoever looks for 'VMware multimonitor does not work with awesome' can find it, apply it, and use VMware with multiple monitors in awesome.

actionless:
if that protocol defined so unclear but just marking its support but ignoring any request is fixing some real usecases i think it's fine to do so

the only thing i would suggest to add is to print a warning on such requests. that would make debugging easier for someone who could face possible issues of having this protocol supported but not really implementing it

marksolaris:
Per my fvwm patch here: https://github.com/i3/i3/issues/2116 just the act of advertizing the atom is enough to make the window manager and VMWare Workstation work together and allow multiple monitors. X11 already knows the real monitor setup, but VMWare doesn't go looking for it unless it sees that atom.

Issue #2178, opened 2018-02-03
input is delayed


1096283585:
Output of `awesome --version`:

awesome v4.2 (Human after all)
• Compiled against Lua 5.3.4 (running with Lua 5.3)
• D-Bus support: ✔
• execinfo support: ✔
• xcb-randr version: 1.5
• LGI version: 0.9.2

**How to reproduce the issue:**

this is the case after the installation.

**Actual result:**

system: arch linux
input: fcitx
especially the terminal,the input is often delayed.Other programs have similar phenomena,but there is no this phenomenon in the 3.X version

**Expected result:**

none

Bekcpear:
Hi, I use the same version and the same system as you, but I dot not see the input delay when using fcitx.

1096283585:
Thank you very much, it seems only to reinstall the system, is currently using the i3wm, next only to try awesome

aparaatti:
I had this with same version output (debian buster).

There where also issues with eclipse not taking shortcut keys. I had to switch to another tag and back to get the shortcut keys working again.

fcitx fixed slow input and I have not noticed strange behaviour (related to input) from eclipse yet.

Elv13:
Most likely unrelated, but who knows:

```
[14:41] <rindolf> Hi all! On Mageia Linux v7 x86-64, with awesome-4.2-1.mga7 , it takes the keyboard keys a short time to be typed on xterm and kate. Plasma 5, JWM, and i3 on the same sys are fine. This also happens in a new user. I am using startx. Can it be fixed?
[14:43] <Elv13> what is "a short time"?
[14:44] <rindolf> Elv13: about 0.3 to 0.5 seconds - noticable
[14:44] <Elv13> what is the hardware and display driver?
[14:45] <rindolf> Elv13: intel sandy bridge core i3
[14:45] <Elv13> Use the EXA, SNA or GLAMOR acceleration?
[14:46] <rindolf> Elv13: how do i check?
[14:47] <Elv13> `egrep -i " connected|card detect|primary dev|Setting driver" /var/log/Xorg.0.log` and  and `egrep -i backend /var/log/Xorg.0.log `
[14:48] <rindolf> Elv13: [   154.051] (II) intel(0): Using Kernel Mode Setting driver: i915, version 1.6.0 20170818
[14:49] <rindolf> [   154.284] (II) intel(0): SNA initialized with Sandybridge (gen6, gt1) backend
[14:50] <Elv13> ok, that's fine
[...]

PR #2198, opened 2018-02-19
Add a common problem section

Elv13:
Ref #2178, #1373 

I could probably find 100 instances of each of those questions in my IRC logs, github, reddit and stackoverflow.

Why the Arch Linux wiki links: Because it will always be better than the answer we come up with and is low level enough to be a generic answer for all distributions.

To reviewer, just press edit for the typo / proposed change then merge. Otherwise it will sit there for months.

### DIFF
--- a/docs/90-FAQ.md
+++ b/docs/90-FAQ.md
@@ -8,6 +8,49 @@ The name *Awesome* comes from the English word *awesome* often used by the
 character [Barney Stinson](http://en.wikipedia.org/wiki/Barney_Stinson)
 from the TV series HIMYM.
 
+## Common issues
+
+### My screens are not configured correctly
+
+Awesome is a window manager, not a desktop environment. To configure your
+screens, the `xrandr` tool or various GUI application exists. For more
+information, refer to the
+[Arch Linux wiki](https://wiki.archlinux.org/index.php/xrandr).
+
+### There is a noticeable delay with inputs
+
+There is 2 common causes for this:
+
+The first is a display driver issue where painting on the screen takes long
+time. All input events (keyboard and mouse) are processed in the main thread.
+The drawing is also locking the main thread of avoid artifacts and race
+contiditions. If there is a delay with the painting, it will delay the inputs.
+The solution to this problem is using a compositing manager such as 
+[compton](https://github.com/chjj/compton) or the older `xcompmgr`. This will
+move the painting input another process and fully mitigate the issue.
+
+The second one is when using `io.popen` or other blocking functions in `rc.lua`.
+Do **not** use such functions and prefer `awful.spawn.easy_async`,
+`awful.widget.watch` or the GIO async API. Even if you *think* a command is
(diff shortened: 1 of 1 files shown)


Elv13:
> (e.g. iowait queue? What's that?)

Install `iotop` on a laptop with a slow HDD, then trash the disk I/O with whatever uncached random read and write. When it hits "99%", the time it takes to execute a simple command such as `time cat /prox/cpuinfo` or `time ps` will go from <2ms "total" to >200ms. There is no gradual degradation, it just hits a cliff. If there is many such commands (let say `vicious` is in uncached mode and you have the same widget on 3 screens), it will totally freeze Awesome on commands that are generally very quick. On my mac, for some unknown (driver issue) reason, reading the battery status from the VFS slow down dramatically (2-3 order of magnitude) even on lite I/O load (like copying a file).

docs/90-FAQ.md
```diff
@@ -8,6 +8,50 @@ The name *Awesome* comes from the English word *awesome* often used by the
...
+[Arch Linux wiki](https://wiki.archlinux.org/index.php/xrandr).
+
+### There is a noticeable delay with inputs
+
+There are two common causes for this:
+
+The first is a display driver issue where painting on the screen takes long
+time. All input events (keyboard and mouse) are processed in the main thread.
+The drawing is also locking the main thread to avoid artifacts and race
+contiditions. If there is a delay with the painting, it will delay the inputs.
```
nickdiego: typo: _contiditions_

docs/90-FAQ.md
```diff
@@ -8,6 +8,50 @@ The name *Awesome* comes from the English word *awesome* often used by the
...
+The second one is when using `io.popen` or other blocking functions in `rc.lua`
+and most commonly manifests itself as occasional freezes instead of a generic delay.
+Do **not** use such functions and prefer `awful.spawn.easy_async`,
+`awful.widget.watch` or the GIO async API. Even if you *think* a command is
+fast enough and won't impact the main event loop iteration time, you are wrong.
+*Every* calls to `io.open` are impacted by the system `iowait` queue and can
+spend hundreds of milliseconds blocked *before* being executed. Note that
+some common widget or probe libraries such as
+[Vicious](https://github.com/Mic92/vicious) do not follow this
+advice currently and are know to cause input lag on some systems (but not all).
```
nickdiego: _are known to..._?

docs/90-FAQ.md
```diff
@@ -8,6 +8,50 @@ The name *Awesome* comes from the English word *awesome* often used by the
 character [Barney Stinson](http://en.wikipedia.org/wiki/Barney_Stinson)
 from the TV series HIMYM.
 
+## Common issues
+
+### My screens are not configured correctly
+
+Awesome is a window manager, not a desktop environment. To configure your
+screens, the `xrandr` tool or various GUI application exists. For more
```
nickdiego: _various GUI applications exist.._ ?

PR #2199, opened 2018-02-21
awesomerc.lua: moved to local variables

spk:
Hi, just moved variables to local instead of global in awesomerc.lua, cheers

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -52,16 +52,16 @@ beautiful.init(gears.filesystem.get_themes_dir() .. "default/theme.lua")
 
 -- @DOC_DEFAULT_APPLICATIONS@
 -- This is used later as the default terminal and editor to run.
-terminal = "xterm"
-editor = os.getenv("EDITOR") or "nano"
-editor_cmd = terminal .. " -e " .. editor
+local terminal = "xterm"
+local editor = os.getenv("EDITOR") or "nano"
+local editor_cmd = terminal .. " -e " .. editor
 
 -- Default modkey.
 -- Usually, Mod4 is the key with a logo between Control and Alt.
 -- If you do not like this or do not have such a key,
 -- I suggest you to remap Mod4 to another key using xmodmap or other tools.
 -- However, you can use another modifier like Mod1, but it may interact with others.
-modkey = "Mod4"
+local modkey = "Mod4"
 
 -- @DOC_LAYOUT@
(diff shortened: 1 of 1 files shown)


spk:
Look like you have this kind of setup :troll:
I can remove the local for terminal, editor, editor_cmd and modkey; I don't think lots of people need myawesomemenu, mymainmenu etc as global, wdyt ? if not feel free to close the PR

actionless:
if i recall correct vicious widgets were encouraging using globals too

so i think some users which were based their lua codestyle on them or rc.lua could use a lot of globals around in the config

psychon:
@Elv13 Well, people who split up their config, already have their own config, no? So, they would have their own `rc.lua` and likely would not just blindly copy over the default config. This is just the default config, so either people copied it and changing the default does not bother them, or they did not make a copy and nothing breaks again.
Or am I wrong here?

If this is merged, we should remove the luacheck exception for `rc.lua` that allows it to set global variables.

Elv13:
Fair enough. That leaves only the shared modules as potential points to worry about. A quick "github wide search" found ~10 of these. But most are not very important

Some are intended for `dofile()` instead of `require` so they are not as problematic.

https://github.com/esn89/searchPrompt

I still have some issues with `modkey` and `clientkeys`. As seen in

https://github.com/Elv13/tyrannical/blob/master/shortcut.lua#L135

`modkey` and clientkeys are quite required in modules that wants to add new keybindings automagically. 

@psychon I will let you decide on this. It *is* going to break some people configs if they use their modules on top of the default `rc.lua` to autopatch Awesome, but as you said, that's not a lot of people. Then it will break a few modules, including 3 of mine. I prefer an harder stance on "don't break the API". Maybe "graduating" modkey and clientkeys into `awful` just as I did for `layouts` would be a good enough compromise as I could update my modules and use it and existing users would mostly not be affected because they use custom `rc.lua`

psychon:
@Elv13 How about making everything but `modkey` and `clientkeys` local then? A follow-up patch could then make this more explicit by replacing `modkey = ` with `_G.modkey = ` and AFAIK that would then allow to remove the luacheck exception, too (but I didn't check if this is really true).

actionless:
i like `awful.key.modkey` approach more

Veratil:
What about tagging this for v5? Since this is a breaking change (and I'm all for semantic versioning, including what we're attempting to do moving forward) it should be slated for the next major version rather than minor.

PR #2201, opened 2018-02-24
Make fake screens permanent

psychon:
Before this, screen_scan() made sure that awesome's information about
available screens conforms to what the X11 server reports. Since fake
screens are, well, fake, this meant that they were deleted.

This commit marks fake screen and handles them specially in
screen_scan() so that they are not deleted.

Note that I did not test this commit at all since I still haven't
figured out how to test RandR stuff without actually messing with
hardware.

Thanks to @madduck for bringing this up and making me fix this long
standing "huh? is that really supposed to work like that?".

----------------

THIS IS UNTESTED!

### DIFF
--- a/objects/screen.c
+++ b/objects/screen.c
@@ -57,6 +57,11 @@
 #include <xcb/xinerama.h>
 #include <xcb/randr.h>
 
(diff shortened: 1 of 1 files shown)


Elv13:
I added the commit to my branch, let see how it goes

Issue #162, opened 2015-03-11
Alpha channel ignored on borders


phryk:
When specifying border colors in an awesome theme, the alpha channel is ignored.
Far as I can tell, only the border colors are affected, but I have only played around
with the values supplied in the default theme.
#### Current behavior:

Consider this screenshot: https://paste.xinu.at/ckn2U/

For the titlebar, the alpha channel works, with the color being "#7fff00b4".
The border, being assigned a color of "#7fff0000", should be invisible, but isn't transparent at all.
#### Expected behavior:

For the borders to honor the alpha channel, which for the supplied example values would mean
the border would be fully transparent.

PS: Someone in the IRC channel implied that the alpha channel is actively stripped/otherwisely ignored in the C part of awesome; And while the lua part of awesome does have a function to strip the alpha channel from a color, I haven't seen it applied to any border property.

psychon:
Welcome to X11. A color has a red, green and blue components, each having a depth of 16 (!) bits.

The composite and render extension jump through some hoops to work around this. It *might* be possible to set a border color with an alpha channel. However, I bet that this will run into video driver bugs. It might be more realistic to remove the border properties from a client and instead implement client borders in lua as "fake titlebars" (if you know what I mean).

Anyone interested to write that code?

actionless:
i am using compton for that, it also adds nice blur on those borders

Elv13:
```
[11:25] <psychon> phryk: are you using a composite manager?
[11:25] <phryk> Elv1313: So I should probably set up the current git master in ~/.local, right?
[11:26] <phryk> psychon: Yes. Because of the alpha bug?
[11:26] <psychon> yeah
[11:26] <phryk> psychon: Yeah, look at the linked screen shot, alpha works fine with xcompmgr for the tasklist, titlebar, etc
[11:27] <psychon> hm... I could move borders into lua
[11:27] <Elv1313> you can already implement border as titlebars
[11:27] <psychon> Elv1313: what do you think about removing c.border_color and c.border_width and instead implement them in awful (similar to how awful.titlebar does things)
[11:28] <Elv1313> psychon: or just use titlebars as border, it will be good for gradient and multi color borders anyway
[11:28] <normalra> psychon: would rules also have to change?
[11:28] <phryk> psychon: If it is easier to do while you're moving anyways, would it be possible to somehow set multiple borders with different width and colors? :3
[...]

PR #2207, opened 2018-03-02
Make alpha work on window borders

psychon:
Up to now, we always asked the X11 server for color allocation ("which
pixel value corresponds to (r,g,b)?", an AllocCollor request).

This commit adds direct support for TrueColor and DirectColor visuals.
In such a visual, the X11 server gives tells us where the red, green,
and blue color components are in a pixel value and we can then just
directly compute the pixel value.

Additionally, this commit adds code that assumes that in a depth=32
visual, the remaining values (after handling red, green, blue) is the
alpha channel for colors. Thus, this adds support for transparent client
borders.

This commit also touches code for the systray. However, the systray must
always use the X11 server's default visual and that one always(?) has
depth=24, i.e. does not support an alpha channel. Thus, the systray
background still cannot be transparent.

Also, in theory this commit should support visuals where some color
component does not have 8 bits, for example RGB565. However, this is
just theoretic and I have no idea how to actually test this (without
jumping through too many hoops).

Fixes: https://github.com/awesomeWM/awesome/issues/162

### DIFF
--- a/color.c
+++ b/color.c
@@ -156,6 +156,7 @@ color_init_unchecked(color_t *color, const char *colstr, ssize_t len, xcb_visual
         req.color->red   = RGB_8TO16(red);
         req.color->green = RGB_8TO16(green);
         req.color->blue  = RGB_8TO16(blue);
+        req.color->alpha = RGB_8TO16(alpha);
         req.color->initialized = true;
         return req;
     }
@@ -190,6 +191,7 @@ color_init_reply(color_init_request_t req)
         req.color->red   = hexa_color->red;
         req.color->green = hexa_color->green;
         req.color->blue  = hexa_color->blue;
+        req.color->alpha = 0xffff;
         req.color->initialized = true;
         p_delete(&hexa_color);
         return true;
@@ -210,9 +212,14 @@ luaA_pushcolor(lua_State *L, const color_t c)
     uint8_t r = RGB_16TO8(c.red);
     uint8_t g = RGB_16TO8(c.green);
     uint8_t b = RGB_16TO8(c.blue);
+    uint8_t a = RGB_16TO8(c.alpha);
 
-    char s[10];
-    int len = snprintf(s, sizeof(s), "#%02x%02x%02x", r, g, b);
+    char s[1 + 4*2 + 1];
+    int len;
+    if (a >= 0xff)
+        len = snprintf(s, sizeof(s), "#%02x%02x%02x", r, g, b);
+    else
+        len = snprintf(s, sizeof(s), "#%02x%02x%02x%02x", r, g, b, a);
     lua_pushlstring(L, s, len);
     return 1;
 }
--- a/color.c
+++ b/color.c
@@ -39,15 +39,17 @@
  */
 static bool
 color_parse(const char *colstr, ssize_t len,
-            uint8_t *red, uint8_t *green, uint8_t *blue)
+            uint8_t *red, uint8_t *green, uint8_t *blue, uint8_t *alpha)
 {
     unsigned long colnum;
     char *p;
 
+    *alpha = 0xff;
+
     colnum = strtoul(colstr + 1, &p, 16);
     if(len == 9 && (p - colstr) == 9)
     {
-        /* We ignore the alpha component */
+        *alpha = colnum & 0xff;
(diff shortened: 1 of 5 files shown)


psychon:
Sigh.

- [ ] I broke querying a color from C. `color_t`'s entries have depth 16, not 8.
- [ ] I never added support for querying colors with a non-`0xff` alpha value. This requires adding an `uint16_t alpha;` field to `color_t` and adding support to `luaA_pushcolor`.

psychon:
https://github.com/awesomeWM/awesome/pull/2207/commits/09f96da20c1fee11b6658dbcb24c784189cf92e5 was amended into the first commit and two more commits fixing things were added. This Should Now Work (tm).

color.c
```diff
@@ -65,19 +67,58 @@ color_parse(const char *colstr, ssize_t len,
...
+    if (width < 8)
+        result >>= (8 - width);
+    if (width > 8)
+        result <<= (width - 8);
+    return result << shift;
+}
+
 /** Send a request to initialize a X color.
  * If you are only interested in the rgba values and don't need the color's
  * pixel value, you should use color_init_unchecked() instead.
```
actionless: i know it's not related to the PR but this line looks weird

psychon:
Does anyone have an opinion or this, or shall I just merge this?

actionless:
what about clarifying that line in the doc? it looks not logical, probably was copied from another similar function

actionless:
i was also thinking, does this change mean what now we can get anti-aliased client-shapes?

Elv13:
> i was also thinking, does this change mean what now we can get anti-aliased client shapes?

AA borders need a 8bit alpha mask. The X11 border mask is 1bit. The 4 titlebars is still the only way to do it until https://github.com/awesomeWM/awesome/issues/650 gets implemented. If/Once the clients has a nice unified API to add decoration instead of 4 individual drawable and lots of painstaking manual Cairo code, it will be easier to implement the shape "properly".

actionless:
but at least we can get AA-ed titlebars, right?

CalebVanTassel:
I've spoke to Elv13 over the past couple of days about just this topic! They helped me implement anti-aliased rounded corners via 4 title bars and some Cario. After editing some of [meriadec's](https://github.com/meriadec/dotfiles/blob/master/config/awesome/modules/smart-borders.lua) code, I was able to successfully end up with this:
image

actionless:
@CalebVanTassel
that looks nice! could you share the resulting code snippet?

CalebVanTassel:
Sure, (here)[https://github.com/CalebVanTassel/dotfiles/blob/master/awesome/smart-borders.lua] it is.

actionless:
thanks!

psychon:
So this is getting more and more offtopic, but whatever `bg_normal = "transparent"` exactly does, it does not do what you think it does. How does `gears.color` handle unparsable strings? I think it falls back to either opaque white or black. Then, the `wibox.layout.align.horizontal` widget in your titlebars seems useless since there are no children. Just leave the titlebar without any widgets? The left and right titlebars seem to have no rounded parts...? Some other parts of this confuse me as well, but... well, that's offtopic.

> i was also thinking, does this change mean what now we can get anti-aliased client shapes?

A client's border is a single, solid color. That does not really allow any kind of anti-aliasing (well, okay, it does, you could set a 1px border that is 50% transparent and pretend that this is anti-aliasing, but I guess that is not what you want). As @Elv13 said, client shapes are still only "on" or "off" per pixel.

actionless:
> A client's border is a single, solid color.

this what i was wondering about

sorry for the offtopic

Elv13:
btw, there is a bunch of places in the doc that say the alpha is not supported

https://awesomewm.org/apidoc/classes/wibox.html#wibox.border_color

I think they are all copy/paste so it should be easy to fix

psychon:
Sorry, but the link that you provide does not sound like "alpha is not supported" to me. The text is "Please note that this property only support string based 24 bit or 32 bit colors:" and I'd say that '32 bit color' refers to "with alpha".

Also, `client.border_color`'s documentation is clearly wrong, but does not talk about alpha either: "pattern Any string, gradients and patterns will be converted to a cairo pattern. "

PR #2208, opened 2018-03-04
Add support for gaps between snapped contents

### LINKED ISSUE #2205: Gaps for window snapping
Here's my lame attempt at hacking something in:

```
diff --git a/lib/awful/mouse/snap.lua b/lib/awful/mouse/snap.lua
index e212d123..54f4ebab 100644
--- a/lib/awful/mouse/snap.lua
+++ b/lib/awful/mouse/snap.lua
@@ -227,8 +227,8 @@ function module.snap(c, snap, x, y, fixed_x, fixed_y)
     for _, snapper in ipairs(aclient.visible(c.screen)) do
         if snapper ~= c then
             local snapper_geom = snapper:geometry()
-            snapper_geom.width = snapper_geom.width + (2 * snapper.border_width)
-            snapper_geom.height = snapper_geom.height + (2 * snapper.border_width)
+            snapper_geom.width = snapper_geom.width + (2 * snapper.border_width) + (2 * (beautiful.snapper_gap or 0))
+            snapper_geom.height = snapper_geom.height + (2 * snapper.border_width) + (2 * (beautiful.snapper_gap or 0))
             geom = snap_outside(geom, snapper_geom, snap)
         end
     end
```

I noticed that this only works when snapping to the right and bottom edges, though. Any idea why? Also, could there be a way to retain both behaviors?

ma9e:
Resolves #2205

### DIFF
--- a/lib/awful/mouse/init.lua
+++ b/lib/awful/mouse/init.lua
@@ -57,6 +57,10 @@ mouse.wibox = {}
 -- @beautiful beautiful.snap_shape
 -- @tparam function shape A `gears.shape` compatible function
 
+--- The gap between snapped contents.
+-- @beautiful beautiful.snapper_gap
+-- @tparam number (default: 0)
+
 --- Get the client object under the pointer.
 -- @deprecated awful.mouse.client_under_pointer
 -- @return The client object under the pointer, if one can be found.
--- a/lib/awful/mouse/snap.lua
+++ b/lib/awful/mouse/snap.lua
@@ -197,11 +197,12 @@ function module.snap(c, snap, x, y, fixed_x, fixed_y)
     c = c or capi.client.focus
     local cur_geom = c:geometry()
     local geom = c:geometry()
-    geom.width = geom.width + (2 * c.border_width)
-    geom.height = geom.height + (2 * c.border_width)
+    local snapper_gap = beautiful.snapper_gap or 0
     local edge
-    geom.x = x or geom.x
-    geom.y = y or geom.y
+    geom.x = (x or geom.x) - snapper_gap
+    geom.y = (y or geom.y) - snapper_gap
+    geom.width = geom.width + (2 * c.border_width) + (2 * snapper_gap)
+    geom.height = geom.height + (2 * c.border_width) + (2 * snapper_gap)
 
     geom, edge = snap_inside(geom, c.screen.geometry, snap)
     geom = snap_inside(geom, c.screen.workarea, snap)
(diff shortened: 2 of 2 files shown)


actionless:
i did some testing and it seems what during the real usage that snapping gap is not really intuitive (because not having any visual hint) so i deny my previous suggestion to use useless_gap for it as default (but mb it can be done again if it will be implemented some visual hint for snappe_gap if it's bigger than `0`)

and just as a sidenote:
 it seems it's not affects that kind of snaping when you drag a floating window to the edge of the screen (like hotcorners / manual tiling)

Elv13:
As for the idea of extending `placement`, I like it. However I acknowledge it's not easy to modify that code. While it has good coverage and everything, there is so many options and abstraction it's confusing to read. It's designed for the people who use the API, not the one who modify it (blame me). The trick is to add another abstraction layer (sigh) for the `border_width` so it can optionally include a gap.

psychon:
> "We can solve any problem by introducing an extra level of indirection, except for the problem of too many levels of indirection"

ma9e:
> i did some testing and it seems what during the real usage that snapping gap is not really intuitive (because not having any visual hint)

This could be an action item. I agree it would be nice to have a thin line that indicates the snap point graphically. I'd also like to be able to snap non-adjacently so the other edges of windows could be lined up as well.

>so i deny my previous suggestion to use useless_gap for it as a fallback (but mb it can be done again if it will be implemented some visual hint for snapper_gap if it's bigger than 0)

I don't see how this follows. Someone might want tiling gaps but not snapping gaps, but then they could just do `theme.snapper_gap = 0` in their theme.rc. There's going to be a UX tradeoff no matter what.

>it seems it's not affects that kind of snaping when you drag a floating window to the edge of the screen (like hotcorners / manual tiling)

This would require the placement filter, no? (I'm using words and I'm not sure what they mean. :P)

My only real concern for this (I think we're bikeshedding) is whether the patch should go into `snap_inside`/`snap_outside` instead. We're making it appear from their perspective as if the snapper/snappee is geometrically larger, but maybe that logic should just be internal to the functions.

actionless:
> I don't see how this follows.

so the point is what since it behaves a bit counter-intuitive without visual hint it's better not to fallback it to useless_gap, better let user set it explicitly

ma9e:
Are you testing with the old patch? The spacing should be the same for equal values now (just not clear on what's unintuitive). I don't mind taking the fallback out for the time being.

actionless:
i've tested again now -- it behave better now, however i still not recommending using useless_gap as fallback, at least until either visual hint either more consistent gaping (in placing and 'floating tiling') will be implemented

actionless:
@Elv13 btw what is the "official" name of that snapping that happens when you drag a client to the side of the screen (which also drawing a red border before applying new geometry)?

psychon:
> How would we fix the coverage checks?

This is usually dealt with by "just ignore coverage". Sadly, there is some randomness in the code so that the coverage is sometimes a bit higher and sometimes a bit lower. Of course, having good coverage is nice, but I wouldn't block this PR for that.

> Anything else that needs to be done for this in the near term?

I'll stay out of this and let @Elv13 decide.

actionless:
@Elv13 that's actually the reason why i told to make it use useless_gap value by default -- until i installed the branch locally i was sure we are talking about that aero snap edge tiling

actionless:
but the ongoing discussion made me to propose the same visual hints for 'inter-client snapping' as we currently have for 'aero snap edge tiling' -- to make them more consistent and feel like one feature

in that case having one gap size as in useless gap wil lmake a lot of sense and in overall give some new nice experience of manual clients arranging on floating tag layout

Elv13:
> to make them more consistent and feel like one feature

Well, client snapping and edge tiling are not exactly the same feature, more like distant cousin in the sense they are both floating mode features.

> but the ongoing discussion made me to propose the same visual hints for 'inter-client snapping' as we currently have for 'aero snap edge tiling'

That would be rather cool. I think @blueyed also once proposed showing a label when moving/resizing the client with the size. There is some APIs to allow this to be done (my Collision module does it). I can take a look at what needs to be done for this to work. However I think this is out of scope of this PR. I also don't have time to do it myself as my AwesomeWM time is locked in the "get all the notification rewrite components merged" and "complete the new layout system" and those 2 alone involve like 10k-15k LoC worth of patches and will probably take another full year at this rate.

My plan is less or more (for the next year):

 * Push the update to the layout system I worked on last week (involves 1 mid size PR)
 * Push the next iteration of the `popup` widget and hopefully get it merged
 * Propose the new version of the keygrabber (really cool stuff in there) (necessary for keyboard control of the notifications). Maybe get it merged for 4.3, but given it's size, it's unlikely.
 * Fix all the stuff to be able to release v4.3
 * The last bunch of commits before getting to the core of the notification stuff
 * Push the PR that create the under-the-hood notification objects and add a "legacy" version of the current widget on top of that API
 * Add all the notification related widgets (icon, label, description, action list, etc)
 * Add the new `rc.lua` notification configuration section (modeled on the titlebar section)
 * Release v4.4

@actionless That's assuming there is noone else to step in and get v4.3 in shape for release. If you have some time, that would be appreciated. ( https://github.com/awesomeWM/awesome/milestone/8 )

As for this PR, I will test it during the weekend. If it works, I will approve/merge it.

PR #2213, opened 2018-03-10
`gears.filesystem`: Add XDG_RUNTIME_DIR

doronbehar:
Just like other `get_xdg`* functions.

### DIFF
--- a/lib/gears/filesystem.lua
+++ b/lib/gears/filesystem.lua
@@ -100,6 +100,12 @@ function filesystem.get_xdg_data_home()
     return (os.getenv("XDG_DATA_HOME") or os.getenv("HOME") .. "/.local/share") .. "/"
 end
 
+--- Get the runtime directory according to the XDG basedir specification.
+-- @treturn string the data home (XDG_RUNTIME_DIR) with a slash at the end.
+function filesystem.get_xdg_runtime_dir()
+    return (os.getenv("XDG_RUNTIME_DIR") or '/run/user/' .. os.getenv("UID")) .. "/"
+end
+
 --- Get the data dirs according to the XDG basedir specification.
 -- @treturn table the data dirs (XDG_DATA_DIRS) with a slash at the end of each entry.
 function filesystem.get_xdg_data_dirs()


lib/gears/filesystem.lua
```diff
@@ -100,6 +100,12 @@ function filesystem.get_xdg_data_home()
     return (os.getenv("XDG_DATA_HOME") or os.getenv("HOME") .. "/.local/share") .. "/"
 end
 
+--- Get the runtime directory according to the XDG basedir specification.
+-- @treturn string the data home (XDG_RUNTIME_DIR) with a slash at the end.
+function filesystem.get_xdg_runtime_dir()
+    return (os.getenv("XDG_RUNTIME_DIR") or '/run/user/' .. os.getenv("UID")) .. "/"
```
psychon: What's your use case for this wrapper? I have a hard time coming up with one since the spec talks about thinks like placing `AF_UNIX` sockets there.
Also, the base dir spec does not really talk about `/run/` at all and I do not see why you can assume that `$UID` is set to anything sensible. For example my awesome does not:
```
$ grep UID /proc/$(pidof awesome)/environ
$
```

doronbehar: I wonder why `UID` is `grep`ed in `/proc/$(pidof awesome)/environ`, it isn't `grep`ed on my machine as well.. Are you sure it means `os.getenv("UID")` won't be evaluated correctly? I've tested it on my machine and I encountered no problem at all.

As for the use case of this wrapper, I'm writing a small Lua module for my AwesomeWM and I use this directory for placing `pid` files of programs I run during my AwesomeWM session. I just thought that this change should complement the other xdg related functions in this module.

psychon: Well, `awesome-client 'return os.getenv("UID")'` doesn't find anything either and
```
$ awesome-client 'return type(os.getenv("UID"))'
   string "nil"
```
Hm, okay, placing PID files there seems to match the intention of this directory. However, no idea how to come up with a good fallback path that always works...

doronbehar: Perhaps we could use the command `id --user`?

psychon: That's ugly and not portable according to POSIX, I guess.

However, even then I do not really like the fallback of `/run/user/$UID/`. That seems to be quite a systemd-ism to me and I'd be surprised if that works on e.g. FreeBSD or when running Linux with upstart.

actionless: https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

>> If $XDG_RUNTIME_DIR is not set applications should fall back to a replacement directory with similar capabilities and print a warning message. Applications should use this directory for communication and synchronization purposes and should not place larger files in it, since it might reside in runtime memory and cannot necessarily be swapped out to disk. 

the thing with /run/user and uid seems much like guessing

doronbehar: Perhaps we can use a fallback similar to what `pyxdg` is using:
https://github.com/takluyver/pyxdg/blob/69707442963112f92d72ad39f8fda93d7760e437/xdg/BaseDirectory.py#L110

Elv13:
Closing this since so far the list of possible use case for this don't seem to be worth having a special wrapper for this. It can be re-opened if needed.

PR #2217, opened 2018-03-12
Remove a weird comment

psychon:
No idea what this is meant to tell us, but color_init_unchecked()
pointing at itself with "use this instead" makes no sense.

@actionless Is this what you meant with the comment being weird?


actionless:
thanks!

Issue #2163, opened 2018-01-21
Travis test is crashing sometimes for `-Dlua_assert=assert`


actionless:
https://travis-ci.org/awesomeWM/awesome/jobs/330944659

```
awesome: lvm.c:436: luaV_equalobj: Assertion `(((t1)->value_).gc)->tt == 5' failed.
```

Related to https://github.com/awesomeWM/awesome/pull/2161#issuecomment-359233666

Also I found a (probably) similar issue in a different lua project:
https://github.com/nodemcu/nodemcu-firmware/issues/802

psychon:
Possibly related: http://lua-users.org/lists/lua-l/2018-06/msg00225.html http://lua-users.org/lists/lua-l/2018-07/msg00003.html https://github.com/pavelxdd/luabug_nccalls (At some point I tried to debug some problems (but it might have been a different bug) and saw these overflows as well)
So that GitHub's search has something to find: These links seem to deal with the error "C Stack Overflow".

psychon:
I just saw a new assertion failure that I had never seen before: https://travis-ci.org/awesomeWM/awesome/jobs/441582772
(Note that this dies while running tests!)
```
[  8%] Running tests/examples/text/gears/object/signal.lua (via tests/examples/text/template.lua)
Result: 134
Error:
lua: lgc.c:986: atomic: Assertion `g->ephemeron == ((void *)0) && g->weak == ((void *)0)' failed.
Aborted (core dumped)
make[2]: *** [CMakeFiles/run-tests-examples-text-gears-object-signal.lua] Error 1
```

blueyed:
Should we try to get the core files out of it?  (e.g. post them to some ("pastebin")
Would it help you to debug it?

psychon:
I doubt it. I do not know anything about Lua internals and diagnosing memory corruption is a mess (assuming that this is memory corruption). Also, there is (AFAIR) still a "smoking gun" in that lgi does "stuff" on a yielded coroutine. I do not really know what kind of symptoms this may cause (but possibly this was already fixed in a release of lgi...).

blueyed:
Just for reference, a more complete copy'n'paste:
```
== Running /home/travis/build/awesomeWM/awesome/tests/test-gravity.lua ==
2018-12-23 17:46:36 W: awesome: a_glib_poll:438: Last main loop iteration took 0.228736 seconds! Increasing limit for this warning to that value.
[...]

PR #2219, opened 2018-03-13
Add "focused" signal to screen.

and3rson:
This PR adds a missing `focused` signal to a screen.

I've ensured there's currently no way to execute some code when active screen is changed.

Personally I use this to change the wibar color to visually distinguish the active screen.

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -114,8 +114,10 @@ function screen.focus(_screen)
     -- save pointer position of current screen
     s.mouse_per_screen = capi.mouse.coords()
 
-   -- move cursor without triggering signals mouse::enter and mouse::leave
+    -- move cursor without triggering signals mouse::enter and mouse::leave
     capi.mouse.coords(pos, true)
+
+    _screen:emit_signal('focused')
 
     local c = client.focus.history.get(_screen, 0)
     if c then
@@ -605,6 +607,9 @@ end
 
 --- When the tag history changed.
 -- @signal tag::history::update
+
+--- When the screen is focused.
+-- @signal focused
 
 -- Extend the luaobject
 object.properties(capi.screen, {


and3rson:
Codecod has shown over 9000 changes. What an I doing wrong?

psychon:
> Codecod has shown over 9000 changes. What am I doing wrong?

I guess the Travis build doing code coverage failed for some reason before it did all the code coverage steps. The crash that I saw so far is a duplicate of #2163.

actionless:
that's a good idea to have such signal (for example when using multiscreen setup i had to catch client.focus and based on client.screen make the decision if the screen have changed) however i wil agree with @psychon what the implementation is incomplete

and3rson:
Thanks for the comments!

I've been digging through the sources and it seems like we should somehow track the change of `mouse.screen` which seems to happen from C. Not sure about how intensive this operation will be.

psychon:
@and3rson The problem with making that change to the C code is that currently the C code does not notice the mouse moving to another screen either. We would have to tell the X11 server to send us a notification on every mouse move, which means selecting for `MotionNotify` on the root window. However, if another window contains the mouse pointer (e.g. a floating window that is halfway between two adjacent screens), then we would not get an event for that, since we did not select `MotionNotify` on that window. Thus, we have to recursively select `MotionNotify` events on all windows, even windows that we do not interact with in any other way (e.g. menus, as in "File->Open").

I bet that this can be done somehow, but right now I do not know how. Also, asking for a notification on every mouse move is a bit overkill when we are just interested in mouse movements between screens...

blueyed:
Maybe this could be based instead of what @actionless does (using the client focus signal and emitting a  "focused" signal for its screen if it is different from the last/current one)?

Elv13:
I will close this, the current code can't be fixed so it's going to need a much more complex implementation before we can get there. Such signal would be very useful, but right now only workaround can exist.

Issue #2221, opened 2018-03-15
Crash on Resume - nil Screen Pointer Ref in awful/ewmh.lua


ewhac:
Output of `awesome --version`:

awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.2

**How to reproduce the issue:**

 * Close laptop lid; wait for LED to show entry into suspend mode.
 * Open laptop lid.

**Actual result:**

awesome has crashed; printing the following error messages.  The timestamps indicate the crash occurred during resume, not suspend:

```
2018-03-15 13:14:13 W: naughty.notify: there is no screen available to display the following notification: title='Oops, an error happened!' text='/usr/share/awesome/lib/awful/ewmh.lua:249: attempt to index a nil value (field 'screen')'
2018-03-15 13:14:13 W: awesome: luaA_dofunction:77: error while running function!
stack traceback:
	/usr/share/awesome/lib/awful/ewmh.lua:249: in function 'awful.ewmh.geometry'
	[C]: in method 'emit_signal'
	/usr/share/awesome/lib/awful/ewmh.lua:60: in function </usr/share/awesome/lib/awful/ewmh.lua:51>
error: /usr/share/awesome/lib/awful/ewmh.lua:249: attempt to index a nil value (field 'screen')
2018-03-15 13:14:13 W: naughty.notify: there is no screen available to display the following notification: title='Oops, an error happened!' text='/usr/share/awesome/lib/awful/ewmh.lua:249: attempt to index a nil value (field 'screen')'
2018-03-15 13:14:13 W: awesome: luaA_dofunction:77: error while running function!
stack traceback:
	/usr/share/awesome/lib/awful/ewmh.lua:249: in function 'awful.ewmh.geometry'
	[C]: in method 'emit_signal'
	/usr/share/awesome/lib/awful/ewmh.lua:60: in function </usr/share/awesome/lib/awful/ewmh.lua:51>
error: /usr/share/awesome/lib/awful/ewmh.lua:249: attempt to index a nil value (field 'screen')
2018-03-15 13:14:13 W: naughty.notify: there is no screen available to display the following notification: title='Oops, an error happened!' text='/usr/share/awesome/lib/awful/ewmh.lua:249: attempt to index a nil value (field 'screen')'
[...]

PR #2223, opened 2018-03-17
screen_refresh(): Fake a screen when the last one is removed

psychon:
Thanks to @ewhac for the suggestion.

Fixes: https://github.com/awesomeWM/awesome/issues/2221

### DIFF
--- a/objects/screen.c
+++ b/objects/screen.c
@@ -652,6 +652,10 @@ screen_refresh(void)
 
     screen_deduplicate(L, &new_screens);
 
+    /* Running without any screens at all is no fun. */
+    if (new_screens.len == 0)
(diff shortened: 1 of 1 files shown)


psychon:
...ping? Anyone wants to review this?

Elv13:
> ...ping? Anyone wants to review this?

I left it unreviewed mostly because it's not the solution I would have used, but my solution patches are nowhere to be found. So while I don't think it's as clean as moving all `screen` objects to be virtual/fake screen and adding the concept of monitor output area to CAPI, this will do for now. Feel free to merge it given it solves a real problem for our users.

blueyed:
Let's try / go with this for now.
Thanks!

PR #2224, opened 2018-03-17
Add categories path to icon lookup

reinerh:
In #2112 the way of loading category icons was changed to use
utils.lookup_icon. However this function does not look in the
categories paths where the categories icons are located, and
so the category icons are no longer being displayed.

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -208,6 +208,9 @@ local function get_icon_lookup_path()
             table.insert(app_in_theme_paths,
                          glib.build_filenamev({ icon_theme_directory,
                                                 size, 'apps' }))
+            table.insert(app_in_theme_paths,
+                         glib.build_filenamev({ icon_theme_directory,
+                                                size, 'categories' }))
         end
     end
     add_if_readable(icon_lookup_path, app_in_theme_paths)


actionless:
i've tried your patch (and i also have `beautiful.icon_theme` set) but i still doesn't have icons in the menubar

reinerh:
@actionless, have you set theme.icon_path in your theme.lua?
You need to set it to a theme that actually ships category icons (e.g. Adwaita). The default hicolor theme does not.

psychon:
I'd tend to just merge this.

The patch is wrong. Our implementation is wrong. Everything is awful. The patch seems to help someone in some situation, so meh, whatever.

Any reason against just merging this?

> i've tried your patch (and i also have beautiful.icon_theme set) but i still doesn't have icons in the menubar

You need files with basenames beginning with applications- like applications-multimedia, applications-development, applications-science. And they need to be in the categories subfolder. I don't have anything like this here (there is an extra layer of directories like `48x48` for me).

reinerh:
For example I have the following files:
```
find /usr/share/icons/ | grep Adwaita.*categories.*development
/usr/share/icons/Adwaita/48x48/categories/applications-development.png
/usr/share/icons/Adwaita/16x16/categories/applications-development.png
/usr/share/icons/Adwaita/24x24/categories/applications-development.png
/usr/share/icons/Adwaita/32x32/categories/applications-development.png
/usr/share/icons/Adwaita/22x22/categories/applications-development.png
/usr/share/icons/Adwaita/256x256/categories/applications-development.png
```
This icon theme is installed by default on Debian systems when gtk3 is installed.
When using this for the icon_theme, I see the category icons.

psychon:
Oh, I have the same paths here. I didn't know/remember that the remaining implementation also searched in those paths. Thus, this patch would also give me category icons.

actionless:
i think the problem what the icontheme with which i tested it have only `scalable` dir inside `categories`, but no precise sizes

Issue #2218, opened 2018-03-12
Must restart for new timezone to take effect


garrison:
Output of `awesome --version`:

```
awesome v4.2 (Human after all)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.4
 • LGI version: 0.9.0
```

**How to reproduce the issue:**

I have a clock made using `wibox.widget.textclock("%a %b %e, %l:%M %P")`.  I change the timezone on my system, using `sudo dpkg-reconfigure tzdata` (or whatever mechanism the distribution at hand uses).

**Actual result:**

The clock does not update its timezone without restarting awesomewm.

**Expected result:**

The clock should update its timezone immediately or soon thereafter.

I am not certain, but I suspect this may be a regression due to #1743.  The local timezone is being determined outside the callback, but it should be determined within it.

psychon:
Since I do not want to mess with my system's timezone for testing: Do you want to test a fix for this? Do you perhaps even want to propose a PR to fix this yourself? Your analysis seems correct to me.

If not, could you at least test the following command? It prints one line per second, the first entry should not follow timezone changes while the second one does: `lua -e 'local glib = require("lgi").GLib local tz = glib.TimeZone.new_local() while true do print(glib.DateTime.new_now(tz):format("%a %b %d, %H:%M"), glib.DateTime.new_now(glib.TimeZone.new_local()):format("%a %b %d, %H:%M")) glib.usleep(1000*1000) end '`

A fix should be as simple as
```diff
diff --git a/lib/wibox/widget/textclock.lua b/lib/wibox/widget/textclock.lua
index f7c8bf78..d4c5765e 100644
--- a/lib/wibox/widget/textclock.lua
+++ b/lib/wibox/widget/textclock.lua
@@ -36,7 +36,7 @@ end
 
 function textclock:set_timezone(tzid)
     self._private.tzid = tzid
-    self._private.timezone = tzid and TimeZone.new(tzid) or TimeZone.new_local()
+    self._private.timezone = tzid and TimeZone.new(tzid)
     self:force_update()
 end
 
@@ -84,10 +84,10 @@ local function new(format, refresh, tzid)
     w._private.format = format or " %a %b %d, %H:%M "
     w._private.refresh = refresh or 60
     w._private.tzid = tzid
-    w._private.timezone = tzid and TimeZone.new(tzid) or TimeZone.new_local()
[...]

PR #2229, opened 2018-03-21
Move timezone determination to callback

garrison:
This is the patch provided [here](https://github.com/awesomeWM/awesome/issues/2218#issuecomment-372579051) by @psychon.  I have tested a backport of this patch to awesome 4.2, and it solved the issue (#2218).  I have compiled but have not tested this patch directly.  [After working around [this issue](https://github.com/awesomeWM/awesome/pull/644#issuecomment-364730185), I found that Xephyr segfaults on startup for me, preventing me from testing this.]

### DIFF
--- a/lib/wibox/widget/textclock.lua
+++ b/lib/wibox/widget/textclock.lua
@@ -36,7 +36,7 @@ end
 
 function textclock:set_timezone(tzid)
     self._private.tzid = tzid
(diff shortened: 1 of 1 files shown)


psychon:
Thanks for testing that patch!
Testing with awesome 4.2 should be enough here since this code does not interact with the C code.

PR #2232, opened 2018-03-27
naughty: Add notification max width and height

alfunx:
Add a theme variable, as well as an argument for `naughty.notify()`, to set the maximum width and height of a notification. This is useful if you want to have automatic sized notifications but with an upper bound. With this, you can prevent for example:

Before:

before

After:

after

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -190,6 +190,14 @@ naughty.notificationClosedReason = {
 -- @beautiful beautiful.notification_height
 -- @tparam int notification_height
 
+--- Notifications maximum width.
+-- @beautiful beautiful.notification_max_width
+-- @tparam int notification_max_width
+
+--- Notifications maximum height.
+-- @beautiful beautiful.notification_max_height
+-- @tparam int notification_max_height
+
 --- Notifications icon size.
 -- @beautiful beautiful.notification_icon_size
 -- @tparam int notification_icon_size
@@ -495,6 +503,10 @@ local function update_size(notification)
         width = s.actions_max_width
     end
 
+    if s.max_width then
+        width = math.min(width, s.max_width)
+    end
+
     -- calculate the height
     if not height then
         local w = width - (n.iconbox and s.icon_w + 2 * margin or 0) - 2 * margin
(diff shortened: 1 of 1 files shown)


nickdiego:
I think notifications and popup widgets API is currently being reworked by @Elv13 

See #2154 and some others..

Elv13:
> I think notifications and popup widgets API is currently being reworked by @Elv13

Yes, but only 45% is merged and 60% "officially" in the pipeline. That's a huge amount of code and requires an enormous amount of time for the review process alone (which I don't have). The bug is there now and it would be a mean to prevent people to have a totally adequate fix right now ;)

actionless:
i've applied your patch but for some reason i see no difference:

2018-04-02-023835_3840x2160_scrot

nickdiego:
@actionless Make sure you have these variables set in your theme:

```lua
theme.notification_max_height = 400
theme.notification_max_width  = 600
theme.notification_icon_size  = 100
```
Here, it worked fine.

alfunx:
@actionless Can you please try this too:

`awesome-client 'require("naughty").notify { text="some not so long text to test awesome notifications"; max_width=150 }'`

@nickdiego Thanks for testing!

actionless:
@alfunx 
2018-04-02--1522666223_3840x2160_scrot

actionless:
seems it was a problem when i was resolving conflict on my local branch which i use for builds

it works as expected, sorry for misleading you

actionless:
@blueyed i remember we were already discussing under other ticket the possibility to loose coverage requrements a little, especially for small patches

actionless:
and actually in this particular case it seems what the reason of coverage is some flaky test related to  `spawn.c | 73.14% <0%> (-3.31%)`

blueyed:
Coverage decrease appears to be mainly because the diff/patch is not covered: https://codecov.io/gh/awesomeWM/awesome/compare/98e59225d74b026a4a655997c4bc1bcb423e6369...143ccaa9a38c7fba850f37f172fe7259832360b0/diff.

blueyed:
Thanks!

actionless:
but it's all green on that diff

blueyed:
No, lines 507 and 524 are red (on the left).  Compare line 506 and 507.
The lines itself are green since they were added in the patch.

actionless:
yeah, my bad, i was looking to a wrong marker

PR #2234, opened 2018-04-03
Move from asciidoc to asciidoctor to generate man pages

aerostitch:
Hi,

Asciidoc is EOL and only supports python2.
Upstream clearly encourages people to move to an equivalent such as asciidoctor, see: https://github.com/asciidoc/asciidoc/releases/tag/8.6.10
As the maintainer of asciidoc in Debian I'm working with the different projects to help the transition, hence the current PR.

Thanks
Joseph

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -37,8 +37,7 @@ endmacro()
 
 a_find_program(GIT_EXECUTABLE git FALSE)
 # programs needed for man pages
-a_find_program(ASCIIDOC_EXECUTABLE asciidoc FALSE)
-a_find_program(XMLTO_EXECUTABLE xmlto FALSE)
+a_find_program(ASCIIDOCTOR_EXECUTABLE asciidoctor FALSE)
 a_find_program(GZIP_EXECUTABLE gzip FALSE)
 # Lua documentation
 if(GENERATE_DOC)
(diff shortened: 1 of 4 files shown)


psychon:
Basically :+1: except for the following on the Travis build:
```
-- asciidoctor not found.
-- Not generating API documentation.
```
Could you replace `asciidoc` by `asciidoctor` in `.travis.yml`? Thanks.

And thanks for this PR. I have to admit that I do not understand this magic process where we get manpages at the end.

aerostitch:
Good catch, thanks! :)
Just amended my commit.

aerostitch:
There's an issue in the travis build that says that asciidoctor cannot load the asciidoctor gem. I'll have a look, it's probably an environment issue.

aerostitch:
So it seems there might be a trick in the way the non-ruby travis environment is setup because it cannot load gems if it is installed through apt but it works if installed through the gem command.
Anyway, that works now.

aerostitch:
I figured out why: the package installs the gem in the global ruby environment but in travis they force rvm, so it cannot be found because the ruby in question does not have access to the global gem library.

aerostitch:
Tried:
- `rvm system` to switch to the system ruby but that didn't go well
- `rvm system ; rvm gemset export system.gems ; rvm default ; rvm gemset import system` to import the system gems into the current rvm environment but that didn't work either
- `export PATH=${PATH}:/usr/lib/ruby` to force the ruby path at the end still no good

So I reverted to use gem for the install of asciidoctor in the travis.yaml for now until someone has a better idea

awesomeConfig.cmake
```diff
@@ -37,8 +37,7 @@ endmacro()
 
 a_find_program(GIT_EXECUTABLE git FALSE)
 # programs needed for man pages
-a_find_program(ASCIIDOC_EXECUTABLE asciidoc FALSE)
-a_find_program(XMLTO_EXECUTABLE xmlto FALSE)
+a_find_program(ASCIIDOC_EXECUTABLE asciidoctor FALSE)
```
blueyed: Shouldn't this be renamed to `ASCIIDOCTOR_EXECUTABLE`?
Although keeping it picks up any custom user config/setting.

actionless: i think that's fine to rename since any custom build which was relying on ASCIIDOC_EXUTABLE could also break because of the removing of XMLTO_EXECUTABLE

aerostitch: Thanks, I updated the variable name

blueyed:
Thank you!

PR #2236, opened 2018-04-04
Allow use of custom completion callback in prompt widget

funtaps:
In my case, I use promtbox to start almost all of my apps, but there is too much noise in basic shell completion.
For example "chro" completes in "chromedriver", "chroot" and only then to "chromium".
I want to use generic completion with my own dictionary, but I need to copy this whole file to do it.

This change will allow usages like
```lua
local custom_completion_dictionary = {"chromium", "firefox", "pcmanfm"}
    s.mypromptbox = awful.widget.prompt({
        completion_callback = function(text, cur_pos, ncomp)
            return awful.completion.generic(text, cur_pos, ncomp, custom_completion_dictionary)
        end
    })
end)
```

### DIFF
--- a/lib/awful/widget/prompt.lua
+++ b/lib/awful/widget/prompt.lua
@@ -60,7 +60,7 @@ end
 -- @tparam[opt=`beautiful.prompt_bg` or `beautiful.bg_normal`] color args.bg Prompt background color.
 -- @tparam[opt=`beautiful.prompt_fg` or `beautiful.fg_normal`] color args.fg Prompt foreground color.
 -- @tparam[opt=`awful.completion.shell`] function args.completion_callback
---   The callback function to call to get completion.
+--   The callback function to call to get completion. See @{awful.prompt.run} for details.
 -- @return An instance of prompt widget, inherits from `wibox.container.background`.
 -- @function awful.widget.prompt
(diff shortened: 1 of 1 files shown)


lib/awful/widget/prompt.lua
```diff
@@ -59,6 +59,8 @@ end
 -- @tparam[opt="Run: "] string args.prompt Prompt text.
 -- @tparam[opt=`beautiful.prompt_bg` or `beautiful.bg_normal`] color args.bg Prompt background color.
 -- @tparam[opt=`beautiful.prompt_fg` or `beautiful.fg_normal`] color args.fg Prompt foreground color.
+-- @tparam[opt=`awful.completion.shell`] function args.completion_callback
+--   The callback function to call to get completion.
```
actionless: i think you should also document here callback function arguments

psychon: How about just pointing to `awful.prompt.run` here?
`See @{awful.prompt.run} for details.` (I'm not sure if the link works like this)

psychon:
Why do you use `awful.widget.prompt` at all then? It does not really do much.

Put differently, what is the problem with
>  but I need to copy this whole file to do it.

PR #2241, opened 2018-04-10
awful.ewmh: doc improvements / TODOs

blueyed:
I've used a custom `ewmh.activate` function myself, and converted this to a filter now.

Firefox focus stealing appears to have been fixed (or I could not trigger it yet again), so the only thing I like is activating the (first) tag of a client if it's not visible with "ewmh" request (e.g. when opening a web page via `xdg-open` or using Slack's notification bar icon; by default non-visible clients get marked as urgent only.

### DIFF
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -30,16 +30,22 @@ local ewmh = {
 -- @tparam[opt=true] boolean fullscreen_hide_border
 
 --- The list of all registered generic request::activate (focus stealing)
--- filters. If a filter is added to only one context, it will be in
--- `ewmh.contextual_activate_filters`["context_name"].
+-- filters.
+--
+-- If a filter is added to a specific context, it gets added to
+-- `ewmh.contextual_activate_filters`["context_name"] instead.
+--
 -- @table[opt={}] generic_activate_filters
 -- @see ewmh.activate
 -- @see ewmh.add_activate_filter
 -- @see ewmh.remove_activate_filter
 
 --- The list of all registered contextual request::activate (focus stealing)
--- filters. If a filter is added to only one context, it will be in
--- `ewmh.generic_activate_filters`.
+-- filters.
+--
+-- If a filter is added without a specific context, it gets added to
+-- `ewmh.generic_activate_filters` instead.
+--
 -- @table[opt={}] contextual_activate_filters
 -- @see ewmh.activate
 -- @see ewmh.add_activate_filter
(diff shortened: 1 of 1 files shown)


lib/awful/ewmh.lua
```diff
@@ -74,18 +80,22 @@ end
 -- It is the default signal handler for `request::activate` on a `client`.
 --
 -- @signalhandler awful.ewmh.activate
--- @client c A client to use
+-- @client c The client to activate.
 -- @tparam string context The context where this signal was used.
 -- @tparam[opt] table hints A table with additional hints:
 -- @tparam[opt=false] boolean hints.raise should the client be raised?
-function ewmh.activate(c, context, hints) -- luacheck: no unused args
+-- @tparam[opt=false] boolean hints.force TODO
```
blueyed: I could not find where `force` gets used internally.

lib/awful/ewmh.lua
```diff
@@ -133,7 +137,19 @@ end
...
 --    end, "ewmh")
 --
+-- Or to activate the first tag of not visible clients (to not make the default
+-- handler mark them urgent only):
+--
+--    awful.ewmh.add_activate_filter(function(c, _context, hints)
+--        if hints.raise and not c:isvisible() then
+--            c.first_tag:view_only()
+--        end
+--    end, "ewmh")
```
blueyed: Does this example make sense like this?
This does not return anything on purpose, but in theory another callback might return `false` later.

Could `hints` be used for this, and then this filter would enforce `hints.make_visibible`?

Elv13: Umm, that's an interesting question. I don't think this is the right place to do this kind of things, but I found no better place either.

What do you think about emitting a `require::tag {raise=true}` (or `request::activate`,  from both `awful.rules.delayed_properties.switchtotag` and `awful.ewmh.activate` and extending `awful.ewmh.tag` to handle raising tags? This would be more consistent with the intend behind your idea considering how it's done for other assets.

Elv13:
@blueyed Ping?

lib/awful/ewmh.lua
blueyed: @Elv13 
Can you comment on that TODO?

lib/awful/ewmh.lua
blueyed: Ok, will likely remove / move it to a new issue then - not now though.

Elv13: v4.3 is approaching, the issues tagged with the v4.3 milestone are about to be fixed. If you want this and some of your other PRs to be part of the release, please update them.

PR #2244, opened 2018-04-22
Add a conditional container.

Elv13:
It was mentioned in #644 for some years but it was stashed away on my computer (incomplete).

This container allow `if`/`elseif`/`else` to be created using the declarative syntax. It is similar to QML `Loader` element but without the reactive syntax tricks.

The main purpose of this widget, beside basic use base like a battery icon or volume icon is to be extended into higher level container like for the client layout system, titlebars or notifications.

Instead of letting it rot in #644 for a million years, let's upstream this now.

Note that there is some shims fix that are not really required here, but are required for the the client layout system conditional examples. It's again to prevent them from rotting away in the branch.


lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
...
+--
+function conditional:eval()
+
+    -- Avoid called eval multiple timer per event loop cycle due to the
+    -- attached signals. This creates and destroy widgets for nothing
+    self._private.exec_once = false
+
+    local old_selected, selected = self._private.selected ,nil
+
+    for _, v in ipairs(self._private.templates or {}) do
```
psychon: Why the `or {}`? This widget does not do anything without templates.

lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
...
+        if v.when then
+            if v.when(self._private.argument_callback(self)) then
```
psychon: The documentation said that `argument_callback` would only ever be called once while this calls it for each iteration of the `for` loop...?

lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
...
+                selected = v
+                break
+            end
+        elseif self._private.default_when then
+            local ret = self._private.default_when(v, self._private.argument_callback(self))
+
+            -- The function must return something or is considered broken
+            assert(type(ret) == "boolean", "The conditional layout `default_when`"..
+                " must return a boolean"
+            )
```
psychon: Why? This feels a bit un-lua-y to me. Lua allows `if 42 then` while e.g. Java does not.

lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
...
+            assert(false, "A conditional template has neither a `when` property"..
+                " nor a default_when"
+            )
+        end
+    end
+
+    -- Nothing changed, move on
+    if selected == self._private.selected then
+        if selected and selected.updated_callback then
+            selected.updated_callback(self._private.argument_callback())
```
psychon: What would be a use case for this callback? I'd say that if nothing changed, then nothing changed.

lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
...
+
+    if self._private.selected and self._private.selected.selected_callback then
+        self._private.selected:selected_callback()
+    end
+
+    local old_widget = self._private.widget
+
+    -- Create the widget unless it's already one.
+    self._private.widget = selected and
+        base.make_widget_from_value(selected) or nil
```
psychon: So that's the place that does this... why? Why not simply create all widgets upfront?

lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
...
+
+    self:emit_signal("property::selected_template",
+        selected,
+        old_selected,
+        self._private.widget,
+        old_widget
+    )
+
+    self:emit_signal("widget::layout_changed")
+    self:emit_signal("widget::redraw_needed")
```
psychon: There is no `redraw_needed`, this widget does not even implement `:draw()`.

lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
...
+
+--- Remove a signal source to automatically call `eval()`.
+function conditional:detach_signal(object, signal)
+    object:disconnect_signal(signal, self._private.eval_handler)
+    for k, v in ipairs(self._private.attached_signal) do
+        if v[1] == object and v[2] == signal then
+            table.remove(self._private.attached_signal, k)
+        end
+    end
+end
```
psychon: I'd move this code closer to the global signal stuff.

lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
...
+    gtable.crush(ret, conditional, true)
+
+    ret._private.exec_once = false
+    ret._private.attached_signal = {}
+
+    ret._private.eval_handler = function()
+        if not ret._private.exec_once then
+            gtimer.delayed_call(function()
+                ret:eval()
+            end)
```
psychon: `gtimer.delayed_call(ret.eval, ret)` avoids creation of the anonymous function, but is less readable. Your call.
(And of course this optimisation is no longer possible if you were to add `exec_once = false` here.

tests/examples/wibox/container/conditional/extend.lua
```diff
@@ -0,0 +1,54 @@
...
+local c = client.gen_fake {x = 45, y = 35, width=40, height=30} --DOC_HIDE
+client.focus = c --DOC_HIDE
+local t = require("awful.tag").add("Test", {screen=screen[1]}) --DOC_HIDE
+c:tags{t} --DOC_HIDE
+
+    local function extended_conditional()
+        return wibox.container.conditional {
+            argument_callback = function()
+                return client.focus, awful.screen.focused().selected_tag
+            end,
```
psychon: So, this is an example for the `argument_callback`. Why not simply access `client.focus` and `awful.screen.focused().selected_tag` in the callback functions below directly? That would shrink the code of the widget by at least 10 lines.

tests/examples/wibox/container/conditional/extend.lua
```diff
@@ -0,0 +1,54 @@
...
+            text   = "There is a client!",
+            when   = function(focused_client, current_tag) -- luacheck: no unused args
+                assert(focused_client and focused_client.name) --DOC_HIDE
+                assert(current_tag and current_tag.name) --DOC_HIDE
+                return true
+            end,
+
+            widget = wibox.widget.textbox
+        },
+        layout = extended_conditional
```
psychon: Hm... wouldn't something like this not be better done by single widget that updates itself when one of the three signals above is emitted? That would even be a lot less code (but would not be declarative, I guess...).

vn-ki:
What happened to this? I feel like this would be a useful container to have.

Elv13:
> What happened to this? I feel like this would be a useful container to have.

Hi. If someone could take over this and fix some of @psychon comments, it would be appreciated. I didn't had time in the last 2 years and I definitively don't have time right now.

As for the whole "do we need this" style of comment, I think that yes, we do need this. As-is, this container isn't providing huge value compared to using imperative code. However, when it is extended for more specialized use case like `only_on_screen` or my dynamic client layout system, it becomes a lot more useful.

vn-ki:
I will try to take a stab at it whenever I can. Can you please rebase it onto the current master if it's not too much trouble?

lib/wibox/widget/base.lua
```diff
@@ -485,6 +496,14 @@ local function drill(ids, content)
         end
     end
 
+    -- Let some containers handle the template themselves.
+    -- This can be used for use cases such as lazy loading, repeaters or
+    -- conditional loaders.
+    if l._accept_templates and l.set_templates then
```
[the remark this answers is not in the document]
Elv13: The historical reason why this PR was written was for the client layout system. See https://elv13.github.io/client_tiling/dynamite.html

For the last couple years, this was mentioned over and over for other reasons. You added the `only_on_screen` container, which is very useful. But the "screen exist" isn't the only use use. This is a more generic version.

It is correct that this doesn't "really really" have to use the template hook API. However that API will also be used for the `gears.connection` API. It has been in my git stash for some years. It solves the problem where it is often hard to access other objects/widgets from within a declarative construct. It's a 1:1 port of the Qt-QML equivalent (same use case, same syntax). So, since this mechanism is needed anyway, better use it and make the API 10% better than a "pure" self contained one.

In retrospect, I should probably have added the `gears.connection` part first. This would have made the `connected_global_signals` API nicer. I will try to address your comments first, then see if I can revive this zombie code and add it to the PR. If I recall, it had many unsolved problems...

tests/examples/wibox/container/conditional/eval.lua
```diff
@@ -0,0 +1,25 @@
...
+    {
+        text   = "There is a client",
+        when   = function() return client.focus ~= nil end,
+        widget = wibox.widget.textbox
+    },
+
+    layout  = wibox.container.conditional
+}
+
+client.connect_signal("focus", function(c) -- luacheck: no unused args
```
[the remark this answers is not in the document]
Elv13: > You can just do function() and remove the comment? Or do function(_), I guess.

So far I often added the "unused" arguments in the API documentation to make the examples easier to copy/paste/extend. Back in 2018, the callback and signals arguments were not documented. This is much better now, but I think having all args, even unused in the examples is still beneficial.

lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
+---------------------------------------------------------------------------
+--
+-- A container displaying a child widget when a condition is met.
+--
+-- This widget accept multiple widget templates (or instances) and will display
+-- the first one matching the condition. If templates are provided (instead
+-- of widget instances), the widget will be created only when the condition is
+-- met.
```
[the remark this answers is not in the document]
Elv13: This is required for the client layouts. See the "Use different layouts depending on the number of clients:" example (https://elv13.github.io/client_tiling/dynamite.html).

lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
...
+-- @property argument_callback
+-- @param function
+
+function conditional:set_argument_callback(callback)
+    self._private.argument_callback = callback
+end
+
+function conditional:get_argument_callback()
+    return self._private.argument_callback
+end
```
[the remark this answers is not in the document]
Elv13: An implementation of `awul.screen.only_on_screen` based on this widget could add a `screen` argument to the when callback. So you could have a theoretical `screens = {foo, bar}` field in the extended widget constructor and always pass that screen.

`wibox.container.conditional` "sucks" when used directly. It was always meant to be a base class of higher level widgets. The only one so far (because it's not merged...) is the `dynamite` declarative client layout module. It uses this mechanism to add the tag and number of clients to the `when` callback.

lib/wibox/container/conditional.lua
Elv13: For the extended widget, and, again, for the client layout system, no template change doesn't always means nothing to do. In the case of the clients, the z-position was the reason for this...

lib/wibox/container/conditional.lua
Elv13: Answered above.

lib/wibox/container/conditional.lua
Elv13: I will rework this in the next iteration once these fixes are approved (the `gears.connection` stuff mentioned above)

tests/examples/wibox/container/conditional/extend.lua
Elv13: Answered above

lib/wibox/container/conditional.lua
```diff
@@ -0,0 +1,362 @@
...
+                        assert(false, "The conditional templates require a `when` function")
+                        return
+                    end
+                end
+            end)
+        end
+    end
+
+    self._private.templates = templates
+    self:eval()
```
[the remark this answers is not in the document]
Elv13: Yeah, a bit. However in the end it is more permissive. It *probably* does the right thing even if a `when` is missing while still printing the warning.

Elv13:
@psychon I don't know if you still care your 2+ years old questions have answers, but they now have. If you are fine with the answer (or don't care anymore), I will add more code to this to add the `gears.connection` code and refactor the `wibox.container.conditional` to use it rather than hardcode the global_signal stuff.

If this get merged, I will use it to fix #3178. A nice side effects is that my `dynamite` module will work with unpatched AwesomeWM again. It required this PR since 2018.

psychon:
> I don't know if you still care your 2+ years old questions have answers, 

Well... I do not remember the questions and I won't remember the answers. For "would it be better to do X", I guess it does not matter what the answer is (either you change the code or you do not). For "I do not understand X", you can make it clearer to future readers of the code (e.g. add comments or docs or stuff like this; "It was always meant to be a base class of higher level widgets." sounds like a case for that), but again it does not really matter to me what the answer is. I will forgot anyway.

>  I will add more code to this to add the gears.connection code and refactor the wibox.container.conditional to use it rather than hardcode the global_signal stuff.

Feel free to do so.

PR #2252, opened 2018-04-25
flex: Remove exit condition

### LINKED ISSUE #2251: tasklist spacing_widget not displaying between last two items when number of windows is even

```
awesome v4.2-304-g681025dc (Human after all)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.1
```

1. Copy the [first custom tasklist example](https://awesomewm.org/apidoc/classes/awful.widget.tasklist.html) (copied below for ease)
```lua
s.mytasklist = awful.widget.tasklist {
    screen   = s,
    filter   = awful.widget.tasklist.filter.currenttags,
    buttons  = tasklist_buttons,
    style    = {
        shape_border_width = 1,
        shape_border_color = '#777777',
        shape  = gears.shape.rounded_bar,
    },
    layout   = {
        spacing = 10,
        spacing_widget = {
            {
                forced_width = 5,
                shape        = gears.shape.circle,
                widget       = wibox.widget.separator
            },
            valign = 'center',
            halign = 'center',
            widget = wibox.container.place,
        },
        layout  = wibox.layout.flex.horizontal
    },
    -- Notice that there is *NO* wibox.wibox prefix, it is a template,
    -- not a widget instance.
    widget_template = {
        {
            {
                {
                    {
                        id     = 'icon_role',
                    },
                    margins = 2,
                },
                {
                    id     = 'text_role',
                },
            },
            left  = 10,
            right = 10,
            widget = wibox.container.margin
        },
        id     = 'background_role',
    },
}
```

**Actual result:**

Spacing widget is not displaying between the last two items when number of windows is an even number.

**Expected result:**

Spacing widget displayed between all tasklist items.

alfunx:
Fixes #2251

The last `spacing_widget` was not inserted, even though there is enough space. The removed condition was formulated incorrectly:

+ the condition should've be e.g. `pos-spacing > width`
+ this exit condition is evaluated based on the next iteration's `pos` (which isn't incorrect, but it does look fishy)
+ the concrete drawing coordinate `x` is calculated with `gmath.round(pos)`, whereas the conditional check is done with plain `pos`, this may have caused the behavior described in #2251

Moreover, this conditional exit misses the case where `space_per_item` is smaller than `spacing`. So if this check indeed has another purpose, it seems like it is incomplete.

Moved up calculation of new `pos` to be analog to `fixed` and `ratio` layouts' codes.

### DIFF
--- a/lib/wibox/layout/flex.lua
+++ b/lib/wibox/layout/flex.lua
@@ -98,14 +98,9 @@ function flex:layout(_, width, height)
             w, h = floor(space_per_item), height
         end
 
-        table.insert(result, base.place_widget_at(v, x, y, w, h))
-
         pos = pos + space_per_item + spacing
 
-        if (is_y and pos-spacing >= height) or
-            (is_x and pos-spacing >= width) then
-            break
-        end
+        table.insert(result, base.place_widget_at(v, x, y, w, h))
 
         if k > 1 and spacing ~= 0 and spacing_widget then
             table.insert(result, base.place_widget_at(


alfunx:
The link you posted points to `fixed.lua`, I guess you meant [#2124 (diff)](https://github.com/awesomeWM/awesome/pull/2124#discussion-diff-152918368R112), which says `pos-spacing >= height`. Though you're right, the image was correct there, as compared to the current [docs](https://awesomewm.org/apidoc/classes/wibox.layout.flex.html#wibox.layout.flex.spacing_widget).

Anyways, I still managed to reproduce #2251 with the conditions set to `pos-spacing > height`:

screenshot

This only happens to me if there are many widgets though.

psychon:
From a quick look: :+1:

blueyed:
Thanks!

PR #2272, opened 2018-05-26
Fix incwfact() error with maximized windows

barlik:
When the window is maximized, calling `awful.client.incwfact()` results in an error because `w` is nil.


actionless:
would you like to extend the test suite to cover the PR? https://github.com/awesomeWM/awesome/blob/master/tests/test-awful-client.lua

barlik:
@actionless sure, I've added a test for the fix.

PR #2273, opened 2018-05-28
Fix doc signals

### LINKED ISSUE #965: The signal and property names documentation are incorrect
While this is well known, there is no (?) issue to track this yet.

Some signals (http://new.awesomewm.org/apidoc/classes/client.html#Signals) have a leading `.` in front of them to avoid ldoc adding an unwanted prefix in front of the name (for example, turn the `manage` signal into `client.manage`. This is also true for properties, but we don't add any prefix, so for example, a background property is displayed as `wibox.container.background.widget` instead of `widget`.

There is some possible solutions for this
- Add an hardcoded prefix like `DOC_AUTO_REMOVE` then use sed in a post-processing script to remove all of them (with the dot). This is easy enough, but a really bad hack.
- Fix `ldoc`. However, `ldoc` doesn't seem to be developed anymore, so I am not sure adding new code will be welcome.

Any other ideas?

Elv13:
While looking at the doc for an unrelated project, I found that there *was* a way to fix #965 in newer versions of ldoc. Given that "newer" versions of ldoc are already ancient, lets use it.

Fix  #965

screenshot_20180528_141803

screenshot_20180528_141834

Before this PR, the signal name would wither have been `awful.client.mysignal`, `client.mysignal` or using the "." hack `.mysignal`. Now it's `mysignal`. Same for the properties. The deprecated functions now have a nice `[deprecated]` label. Otherwise it wasn't clear they were deprecated if you used `ctrl+f` to search the page.

### DIFF
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -140,6 +140,10 @@ custom_display_name_handler = function(item, default_handler)
         return name ~= "" and name or item.name
     end
 
+    if item.type == "deprecated" then
+        return default_handler(item) .. "</a> <i class=\"deprecated_label\">[deprecated]</i><a>"
+    end
+
     return default_handler(item)
 end
 
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -132,4 +132,15 @@ file = {
     }
 }
 
+custom_display_name_handler = function(item, default_handler)
+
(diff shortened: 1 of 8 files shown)


psychon:
Fine by me, but I did not look too closely

PR #2275, opened 2018-05-28
[RFC] client: Make startup_id writable

Elv13:
Following https://www.reddit.com/r/awesomewm/comments/8m6b1h/why_is_it_that_some_programs_ignore_properties/ , I submit this commit again (it was parked in the rules chain of responsibility PR).

I know @psychon hates it and I agree it is a giant hack, however it makes life better by allowing nice things like `mod4+r` + `...` + `shift+return` to spawn client in floating mode. Whether the comment deserve to be in the `recipe` section or in the documentation is up to debate.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -723,12 +723,59 @@
  * identifier remain the same. This allow to match a spawn event to an actual
  * client.
  *
+ * This is used to display a different mouse cursor when the application is
+ * loading and also to attach some properties to the newly created client (like
+ * a `tag` or `floating` state).
+ *
+ * Some applications, like `xterm`, don't support startup notification. While
+ * not perfect, the addition the following code to `rc.lua` will mitigate the
+ * issue. Please note that this code is Linux specific.
+ *
+ *    local blacklisted_snid = setmetatable({}, {__mode = "v" })
+ *
+ *    --- Make startup notification work for some clients like XTerm. This is ugly
+ *    -- but works often enough to be useful.
+ *    local function fix_startup_id(c)
+ *        -- Prevent "broken" sub processes created by `c` to inherit its SNID
+ *        if c.startup_id then
+ *            blacklisted_snid[c.startup_id] = blacklisted_snid[c.startup_id] or c
+ *            return
+ *        end
+ *
+ *        if not c.pid then return end
+ *
+ *        -- Read the process environment variables
+ *        local f = io.open("/proc/"..c.pid.."/environ", "rb")
+ *
+ *        -- It will only work on Linux, that's already 99% of the userbase.
+ *        if not f then return end
+ *
+ *        local value = _VERSION <= "Lua 5.1" and "([^\z]*)\0" or "([^\0]*)\0"
+ *        local snid = f:read("*all"):match("STARTUP_ID=" .. value)
+ *        f:close()
+ *
+ *        -- If there is already a client using this SNID, it means it's either a
+ *        -- subprocess or another window for the same process. While it makes sense
+ *        -- in some case to apply the same rules, it is not always the case, so
+ *        -- better doing nothing rather than something stupid.
+ *        if blacklisted_snid[snid] then return end
+ *
+ *        c.startup_id = snid
+ *
+ *        blacklisted_snid[snid] = c
+ *    end
+ *
+ *    awful.rules.add_rule_source(
+ *        "snid", fix_startup_id, {}, {"awful.spawn", "awful.rules"}
+ *    )
+ *
  * **Signal:**
  *
  *  * *property::startup\_id*
  *
  * @property startup_id
  * @param string
+ * @see awful.spawn
  */
 
 /**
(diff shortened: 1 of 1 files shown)


objects/client.c
```diff
@@ -3046,6 +3092,15 @@ luaA_client_get_icon_name(lua_State *L, client_t *c)
...
 }
 
+static int
+luaA_client_set_startup_id(lua_State *L, client_t *c)
+{
+    const char *startup_id = luaL_checkstring(L, -1);
+    c->startup_id = a_strdup(startup_id);
+    luaA_object_emit_signal(L, -3, "property::startup_id", 0);
+    return 0;
+}
```
psychon: Please use `DO_CLIENT_SET_STRING_PROPERTY(startup_id)` instead of hand-writing this function. That fixes the memory leak of not freeing the old `startup_id` in here, it only emits the signal if the value actually changed and it needs so much less lines of code. Look at the handling of `name`, `role`, `machine` etc.

psychon: Well, okay, that macro only generates a function for setting the string property. A short function similar to `luaA_client_set_name` is still needed. Still, you get the idea.

objects/client.c
```diff
@@ -723,12 +723,58 @@
  * identifier remain the same. This allow to match a spawn event to an actual
  * client.
  *
+ * This is used to display a different mouse cursor when the application is
+ * loading and also to attach some properties to the newly created client (like
+ * a `tag` or `floating` state).
+ *
+ * Some applications, like `xterm`, don't support startup notification. While
+ * not perfect, the adding the following code to `rc.lua` will mitigate the
```
psychon: `the adding`?
Just `adding` (my preferred version) or `the addition of`

objects/client.c
```diff
@@ -723,12 +723,58 @@
...
+ * issue. Please note that this code is Linux specific.
+ *
+ *    local blacklisted_snid = setmetatable({}, {__mode = "v" })
```
psychon: I agree that some way for this not to grow unboundedly is needed, but why does tying this to the speed of the garbage collector make sense? However, I do not really have a better suggestion for this as we cannot know when the last process with the environment variable exits.

objects/client.c
```diff
@@ -723,12 +723,58 @@
...
+ *
+ *    --- Make startup notification work for some clients like XTerm. This is ugly
+ *    -- but works often enough to be useful.
+ *    local function fix_startup_id(c)
+ *        -- Prevent "broken" sub processes created by `c` to inherit its SNID
+ *        if c.startup_id then
+ *            blacklisted_snid[c.startup_id] = blacklisted_snid[c.startup_id] or c
```
psychon: Why not just ` = true`? The value in this table is never used.

objects/client.c
```diff
@@ -723,12 +723,58 @@
...
+ *
+ *        if not c.pid then return end
+ *
+ *        -- Read the process environment variables
+ *        local f = io.open("/proc/"..c.pid.."/environ", "rb")
+ *
+ *        -- It will only work on Linux, that's already 99% of the userbase.
+ *        if not f then return end
+ *
+ *        local snid = f:read("*all"):match("STARTUP_ID=(.*)\0")
```
psychon: Correct me if I am wrong, but Lua patterns are greedy, so this actually gets you the value of `STARTUP_ID` plus a weird mess of all following environment variables. Sigh, such fun...

objects/client.c
Elv13: done

objects/client.c
psychon: Ah, okay, so this relies on the C code mis-handling strings and truncating them at embedded zero bytes in the call to `a_strdup()`. That's... surprising and could arguably be called a bug, but whatever. Weird magic...

objects/client.c
Elv13: done

objects/client.c
Elv13: Ok, If you insist

done

objects/client.c
```diff
@@ -723,12 +723,58 @@
...
+ *
+ *        if not c.pid then return end
+ *
+ *        -- Read the process environment variables
+ *        local f = io.open("/proc/"..c.pid.."/environ", "rb")
+ *
+ *        -- It will only work on Linux, that's already 99% of the userbase.
+ *        if not f then return end
+ *
+ *        local snid = f:read("*all"):match("STARTUP_ID=([^\0]*)\0")
```
psychon: Sigh. Greetings from Lua 5.1:
```
$ lua5.1 -e 'print(io.open("/proc/self/environ"):read("*all"):match("STARTUP_ID=([^\0]*)\0"))' 
lua5.1: (command line):1: malformed pattern (missing ']')
stack traceback:
	[C]: in function 'match'
	[C]: ?
```

psychon: Oh. In Lua 5.1, `\z` works for matching a zero byte. However, it does not work in Lua 5.2 or 5.3. (I did not test LuaJIT)

Elv13: So, can I revert to letting `strdup` do its well documented job or should I add code to find the `\0` and cut the string in lua for the sake of pedantism?

psychon: Pick what you prefer. It feels like all versions are awful :-(.
(You could also do the following, which is still awful)
```lua
local value = _VERSION <= "Lua 5.1" and "([^\z]*)\0" or "([^\0]*)\0"
local snid = f:read("*all"):match("STARTUP_ID=" .. value)
```

Elv13:
Updated

Elv13:
thanks

PR #2276, opened 2018-05-29
Add new beautiful.colorful module

Veratil:
~~Add two new gears.color functions:~~
* ~~interpolate~~
* ~~stringify~~

~~Add one new gears.math function:~~
* ~~roundf~~
*~~ sign~~

~~I've used an interpolation function for a while now to calculate colors where a pattern isn't really viable. I figured this might be useful in gears.color as well. Additionally, I convert those colors to strings for markup, so I figured a stringify function would be nice to have as well.~~

~~I use `roundf` for my vicious.widgets.net to show a rounded value to 1 decimal point, I pulled it from lua-users.org's [SimpleRound](http://lua-users.org/wiki/SimpleRound). I figured it would be useful in gears.math alongside `round`.~~

[Updated PR description](https://github.com/awesomeWM/awesome/pull/2276#issuecomment-456575410)

### DIFF
--- a/lib/beautiful/colorful.lua
+++ b/lib/beautiful/colorful.lua
@@ -102,14 +102,11 @@
 -- @module beautiful.colorful
 ---------------------------------------------------------------------------
 
-local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
-
 local colorful = {}
 
 local color = {
     -- set the from and as tables
     from = {},
-    as = {}
 }
 
 -- Metatable for color instances. This is set up near the end of the file.
@@ -145,12 +142,15 @@ local function between(v, l, h)
     return false
 end
 
+-- The formulas in hsl_to_rgb and rgb_to_hsl are described at the
+-- Wikipedia page https://en.wikipedia.org/wiki/HSL_and_HSV
+
 -- Expects Hue [0,360], Saturation [0,1], Lightness [0,1]
 -- Does NOT check bounds
 local function hsl_to_rgb(hue, saturation, lightness)
     -- If s = 0 then it's a grey
     if saturation == 0 then
-        return unpack({lightness, lightness, lightness})
+        return lightness, lightness, lightness
     end
     local C = (1 - math.abs(2 * lightness - 1)) * saturation
     local X = C * (1 - math.abs((hue / 60) % 2 - 1))
@@ -169,13 +169,13 @@ local function hsl_to_rgb(hue, saturation, lightness)
     elseif 300 <= hue and hue <= 360 then
         r,g,b = C,0,X
     end
-    return unpack({r+m, g+m, b+m})
+    return r+m, g+m, b+m
 end
 
 -- Expects Red, Blue, and Green [0,1]
 -- Does NOT check bounds
 local function rgb_to_hsl(red, green, blue)
-    local h,s,l = 0,0,0 -- luacheck: ignore
+    local h,s,l
     local M = math.max(red, green, blue)
     local m = math.min(red, green, blue)
     local C = M - m
@@ -198,7 +198,7 @@ local function rgb_to_hsl(red, green, blue)
     else
         s = C / (1 - math.abs((2 * l) - 1))
     end
-    return unpack({h, s, l})
+    return h, s, l
 end
 
 -- Create a color object
@@ -208,10 +208,6 @@ local function create_color_object(r,g,b,h,s,l,a)
         hue = h, saturation = s, lightness = l,
         alpha = a
     }, color_mt)
-    -- allow as functions to use color object variables
-    setmetatable(o.as, {
-        __index = o
-    })
     return o
 end
 
@@ -299,7 +295,6 @@ end
 
 local css = {
     integer = "%d?%d?%d",   -- 0-999 (but 0-255 for values)
-    --percent = "%d?%d?%d%%", -- 0-999% (but 0-100% for values)
     float   = "%d?%.?%d+",  -- this is _very_ crude, but probably fast
     comma   = "%s*,%s*",    -- whitespace optional comma
     space   = "%s+",        -- whitespace required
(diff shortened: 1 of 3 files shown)


psychon:
>  I pulled it from lua-users.org's [SimpleRound](http://lua-users.org/wiki/SimpleRound). 

Quick first comment: That page does not have any license statement, so must be considered "all rights reserved", thus GPL incompatible.
(And now: Good night)

alfunx:
They do mention something related to licensing [here](http://lua-users.org/MiniCharter.html)

> content guidelines
> 
> Any software running on the site such as CGI scripts should be under BSD or GPL or similar free license. Any data related to the site content (for example, wiki database) should be backed up and made publicly available at regular intervals. One reason for this is if the individuals controlling the lua-users.org site should go astray (from the community's point of view), it will be possible to install the content at another location and the Lua authors can redirect the domain appropriately.

Maybe you can ask them for details.

Veratil:
Here's the replies I got:
> <zy]x[yz> surely it's under the if-you-post-something-on-a-wiki-expect-people-to-rip-off-it license
> <[xyzzy]> this comes up every once in a while and i'm pretty sure the answer is "nobody knows"
> <jdarnley> Without an explicit statement your lawyer will tell you that they are All Rights Reserved.

Veratil:
Granted the formula for rounding isn't exactly copyrighted, it's just math. I figured I'd be honest though and say I found that specific code on the wiki. :slightly_frowning_face:

lib/gears/math.lua
```diff
@@ -79,5 +79,16 @@ function gmath.round(x)
...
 end
 
+--- Round a number to a 10s place
+-- @class function
+-- @name roundf
+-- @tparam number v
+-- @tparam[opt=1] number bracket 10s place to round to
+function gmath.roundf(v, bracket)
+	-- From lua-users.org SimpleRound
+	local function sign(v) return (v >= 0 and 1) or -1 end
```
psychon: The `sign` function is used just once and thus does not need to be a function.

Also, unit tests would be nice (also for the other functions in this PR)

lib/gears/math.lua
```diff
@@ -79,5 +79,16 @@ function gmath.round(x)
     return math.floor(x + 0.5)
 end
 
+--- Round a number to a 10s place
+-- @class function
+-- @name roundf
+-- @tparam number v
+-- @tparam[opt=1] number bracket 10s place to round to
```
psychon: So.... this is supposed to be 1, 10, 100, 0.001(?), or what?
I am not sure, but I feel like 0, 1, 2, 3 is more common for this (meaning: x'd place after the comma)

lib/gears/math.lua
```diff
@@ -79,5 +79,16 @@ function gmath.round(x)
...
+function gmath.roundf(v, bracket)
+	-- From lua-users.org SimpleRound
+	local function sign(v) return (v >= 0 and 1) or -1 end
+	bracket = bracket or 1
```
psychon: Is the default of 1 sensible? In that case this just rounds away from zero (while gears.math.round rounds towards positive infinity).
Which raises the question if `gears.math.round` is wrong and should be fixed. C99's `round()` rounds away from zero as well and this behaviour would allow to implement this function as `gears.math.round(v / bracket) * bracket`.

lib/gears/math.lua
Veratil: Yes, you can round to any place you want. Rounding to 1 is default. If we change to 0, 1, 2, 3, etc. then we add an exponent operation in (which isn't slow, but just an extra op we can avoid).

Elv13:
@psychon what do we do with this?

Veratil:
I forgot about this one! If there's anything you need me to change let me know.

Elv13:
> I forgot about this one! If there's anything you need me to change let me know.

Apparently, everybody forgot about this one ;). Fact is, none of us had much time in H2 2018.

lib/gears/color.lua
```diff
@@ -107,6 +108,50 @@ function color.parse_color(col)
...
 end
 
+--- Interpolate between two colors
+-- Note: This does interpolate the alpha channel if there is one in a or b.
+-- @tparam number val Value between 0 and 1
+-- @param a Starting color point
+-- @param b Ending color point
+-- @treturn table 4 values representing color in RGBA format (each of them in
+-- [0, 1] range) or nil if input is incorrect.
+function color.interpolate(val, a, b)
```
actionless: i would name it `mix` (as for example in scss)

Veratil:
@actionless Alright! I've updated the `gears.color` functions (and added some new), added tests, replaced local functions in the `gtk` and `xresources` themes and `lib/beautiful/gtk.lua` with the gears.color functions. :)

themes/gtk/theme.lua
```diff
@@ -44,25 +30,9 @@ local function is_dark(color_value)
...
-        if bg_numeric_value > 255 then bg_numeric_value = 255 end
-        result = result .. string.format("%02x", bg_numeric_value)
-    end
-    return result
-end
 local function reduce_contrast(color, ratio)
-    ratio = ratio or 50
-    return darker(color, is_dark(color) and -ratio or ratio)
+    ratio = (ratio or 50) / 100
+    return gcolor.darken(color, is_dark(color) and ratio or 0)
```
actionless: it was before `-ratio or ratio` but after your change `ratio or 0`

themes/xresources/theme.lua
```diff
@@ -64,21 +65,11 @@ theme = theme_assets.recolor_layout(theme, theme.fg_normal)
...
-        result = result .. string.format("%2.2x", bg_numeric_value)
-    end
-    return result
-end
 theme = theme_assets.recolor_titlebar(
     theme, theme.fg_normal, "normal"
 )
 theme = theme_assets.recolor_titlebar(
-    theme, darker(theme.fg_normal, -60), "normal", "hover"
+    theme, gcolor.darken(theme.fg_normal, 0.6), "normal", "hover"
```
actionless: here and next i see you swapped direction from making lighter color into making darker one

themes/gtk/theme.lua
Veratil: Ah, I was thinking if it was dark then don't do anything, I guess if it's dark then the original `darker` made it lighter?

themes/xresources/theme.lua
```diff
@@ -64,21 +65,11 @@ theme = theme_assets.recolor_layout(theme, theme.fg_normal)
...
-        result = result .. string.format("%2.2x", bg_numeric_value)
-    end
-    return result
-end
 theme = theme_assets.recolor_titlebar(
     theme, theme.fg_normal, "normal"
 )
 theme = theme_assets.recolor_titlebar(
-    theme, darker(theme.fg_normal, -60), "normal", "hover"
+    theme, gcolor.lighten(theme.fg_normal, (60 / 255)), "normal", "hover"
```
actionless: also, `lighten` returns array of RGB channels, while everywhere it's expected in a different format

given this particular example, `theme_assets.recolor_titlebar` is passing color value to `gears.color.recolor_image` which is passing at the end to `gears.color.create_pattern`

spec/gears/color_spec.lua
```diff
@@ -291,6 +292,233 @@ describe("gears.color", function()
...
+                assert.is.same(col[7] .. "ff", color.stringify({color.mix(col[1], col[5], 0.75)}, true, 8))
+            end)
+        end
+    end)
+
+    -- 100 of first, 0 of second
+    describe("mix 100%", function()
+        for _,col in ipairs(colors) do
+            it(col[1], function()
+                assert.is.same(col[1] .. "ff", color.stringify({color.mix(col[1], col[5], 1)}, true, 8))
```
actionless: does this repetition of stringify every time looks fine for you? i would suggest having `color.hex_mix` , `hex_darker` and other similar proxies or find some other solution to have more usable output from those new functions

themes/xresources/theme.lua
Veratil: D'oh! You're right. It was obviously a late night for me. Have to stringify the values returned.

spec/gears/color_spec.lua
Veratil: That's not a bad idea.

I was actually thinking about creating a complete `Color` object which we could use similar to how SCSS is put together. I noticed in the GTK helper that the color was accessible through `gtk_color.red`, `.green`, `.blue`, and `.alpha`. That is my idea for a good object including the additional HSL code. Ideally I wanted to be able to do something like `color:lighten(0.25)` and do all the magic behind the scenes, afterwards doing something like `color:str()` to get the stringify'd output.

actionless: i like the idea with object, @Elv13 do you have any opinion on this?

psychon: So, similar to `gears.matrix`, I guess.

actionless: btw i think `:hex()` would be better than `:str()` since it's more explicit

Veratil: > btw i think `:hex()` would be better than `:str()` since it's more explicit

For me, I'd expect `:str()` to return a string representation of the color in `#rrggbb[aa]` format, and `:hex()` to return a decimal (that's actually how I have it right now). But I'll leave that for the group to decide since it's just a function name change. :)

I have a working Color object right now that I've played with based upon how `gears.matrix` is built. I'm looking at making it `gears.color.Color`. @psychon @actionless I'll open a different PR for this direction when I get it fixed up.

Veratil: > my point what there are several possible string representations for color (let's say `rgba(1, 2, 3, 4)`) so i was thinking from position what it will be the most understandable which exactly text representation is used

We can discuss how to name the different output representations in the other PR once I get it created.

> > :hex() to return a decimal
> 
> that sounds not logical to me: decimal is base of 10, hex is base of 16

When you print a hexadecimal number in lua do you see "0xf" or "15"? ;) Just pulling your leg.
But I know what you mean. They'll just be function names we can rename, so it's all good.

actionless: so your whole idea to now create functions and in new PR create an object which will be using those functions to implement color operations on it?

Veratil: Ideally everything will be done within the `Color` object itself.

I sort of have it set up like this right now (we can rename output functions later):
```lua
Color = (require 'gears.color').Color
c = Color.new("#ffcc99")
print(c:as_str()) -- prints "#ffcc99"
print(c:as_hex()) -- prints "4291598847"
print(c.red) -- prints 1
print(c.green) -- prints 0.8
print(c.blue) -- prints 0.6
print(c.hue) -- prints 30.0
print(c.saturation) -- prints 1.0
print(c.lightness) -- prints 0.8
print(c.opacity) -- prints 1
c2 = Color.new("#99ccff")
print(c1:mix(c2)) -- defaults to 0.5 mix, half and half; prints #ccccccff
print(c1:mix(c2, 0)) -- prints #ffcc99ff
print(c1:mix(c2, 1)) -- prints #99ccffff
```

You can do a function on the object itself and it will return a new color (unless we want to make something like `c1:mix(c2)` destroy the old color). So we'd have `color:lighten`, `color:darken`, etc.

The other PR I'll make (unless I just wipe this one and rename it) will essentially replace this one if we want to go that route.

actionless: i would suggest implementing object right in this PR

especially because release is soon and if we merge one API and next will replace it with another API that will mean deprecating old API and providing backward compatibility

actionless: and regarding naming, you could just google "hex colors" and observe what it's common name of "#0088ff" representation

actionless: other than points above : i really like color object idea and appreciate the time you investing into making it happen :+1:

Veratil: Alright! Sounds good! I'll do it here. :)

Veratil:
I have a few more things I want to add to the Color.String parsing (you can see the commented out lines in `lib/gears/color/Color.lua`). Right now it only parses the `#hex` strings. I'm working on adding the `rgb[a](r,g,b[,a])` and `hsl[a](h,s,l[,a)` types as well.

Elv13:
Good question. `gears.color.builder`? (just an idea)

actionless:
`gears.color.object`?

Veratil:
> Good question. `gears.color.builder`? (just an idea)

That may work.

> `gears.color.object`?

I feel like `object` doesn't fit well...

`gears.color.object.new()` vs `gears.color.builder.new()`
Actually I forgot about the `__call` metamethod, so `gears.color.object()` vs `gears.color.builder()` ?

Any other ideas?

actionless:
i think using the same name for module and its submodule will  end up very confusing when playing with it

Veratil:
> i think using the same name for module and its submodule will end up very confusing when playing with it

You're probably right.

I wonder if we reorganize a bit and put the current `gears/color.lua` to `gears/color/cairo.lua`?

actionless:
i think in scope of this PR you can just do the second part:

> then we rename gears/color/color.lua to gears/color/builder.lua or object.lua?

Elv13:
What about moving the color_builder (assuming that name stick) to `beautiful` like the other @actionless theme asserts modules? Technically `wibox` depends on `beautiful`, so it would not change the dependency situation and solve the problem? This has props and cons, but it is rather theme specific and has no use in let say, within `gears` itself.

Elv13:
> What about keeping it named color and having beautiful.color? Think that would be okay?

I would rather not. I worked hard to deprecate all instances of the same name used for different thing over the years. See the open `capi.button` vs `awful.button` PR or the Wibar (was awful.wibox) vs. Wibox (the library). Using a single word for many concepts makes the API documentation confusing.

Veratil:
> > What about keeping it named color and having beautiful.color? Think that would be okay?
> 
> I would rather not. I worked hard to deprecate all instances of the same name used for different thing over the years. See the open `capi.button` vs `awful.button` PR or the Wibar (was awful.wibox) vs. Wibox (the library). Using a single word for many concepts makes the API documentation confusing.

Well... if we're talking names, `gears/color.lua` isn't named correctly as it creates patterns not colors. ;)

Elv13:
> Well... if we're talking names, gears/color.lua isn't named correctly as it creates patterns not colors. ;)

Yes, I agree with that. If I had to pick a name now it would be called something else.

Veratil:
> > Well... if we're talking names, gears/color.lua isn't named correctly as it creates patterns not colors. ;)
> 
> Yes, I agree with that. If I had to pick a name now it would be called something else.

So let's make this PR `beautiful.color`. I'll create a second PR which does the unthinkable and renames `gears.color` to `gears.pattern`, keeping `gears.color` for future deprecation?

Elv13:
I diasgree with changing `gears.color` name after so many years just so we can add an unrelated module with the same name. First of all because `gears.color` can only be deprecated in 5.x so it would still cause 2 modules called `color` to exist for many years. Also because we can't always fix the past. Each API changes cost users. The benefits of changing things have to outweigh the disruption they cause in this cause. In my humble opinion this isn't worth changing in this case. Other can argue differently. I think `beautiful.color_builder` or `beautiful.color_mixer` are also more representative of what the module does.

Veratil:
I don't like the idea of having a two word module where one word could work.

Here's some ideas that are closest I think:
* `beautiful.dye`
* `beautiful.paint`
* `beautiful.pigment`

A few stretches:
* `beautiful.makeup`
* `beautiful.mixer`

It could be used as such, using `COLOR` as a placeholder:
* `beautiful.COLOR.function_on_a_color(value)` e.g. `beautiful.COLOR.mix`, `beautiful.COLOR.darken`
* `beautiful.COLOR.color.new(value)` in the situation you want to create a color object to reuse

I'm fond of `paint` and `pigment` myself.

Elv13:
> i would choose between mixer, color_mixer and color_builder

I concur

Veratil:
Does anyone have a one word idea?

Veratil:
@Elv13 @actionless what do you think about these?
* `colorful` (I like `beautiful.colorful` the most as a name because they both end in `ful` 🙂)
* `colorizer`
* `colorsys` (python's color lib is called this)
* `colorlab`

actionless:
from those the first i like the most

Veratil:
Changed the PR to just be for `beautiful.colorful`. I've pulled out the original commits for `gears.math` and put them in another branch to make a PR for. I've squashed all the other commits (testing and updating local) into one.

themes/gtk/theme.lua
```diff
@@ -61,8 +62,9 @@ local function mix(color1, color2, ratio)
     return result
 end
 local function reduce_contrast(color, ratio)
-    ratio = ratio or 50
-    return darker(color, is_dark(color) and -ratio or ratio)
+    ratio = (ratio or 50) / 255
```
Veratil: Just noticed this should probably be 127 instead of 50.

actionless:
great job, thanks!
i briefly looked and it seems fine, i'll click accept button after you'll push the fixes for lua inter-version compatibility (since they will discard review anyway)

Veratil:
> great job, thanks!
> i briefly looked and it seems fine, i'll click accept button after you'll push the fixes for lua inter-version compatibility (since they will discard review anyway)

Yeah, latest fixup is to support Lua 5.1 bitshift specifically... I found out that Lua 5.1 doesn't have any bitshift operators, so I had to adjust a few functions.

actionless:
i just recalled what the same problem was in this PR: https://github.com/awesomeWM/awesome/pull/2383

mb we need some more genera solution inside `gears`? cc @Elv13

Veratil:
@actionless @Elv13 I think adding `lshift` and `rshift` (and the removed from this PR `roundf` and `sign`), to `gears.math` would be a good idea. I can update a little more documentation on them, too. I would then add `gears.math` in `colorful` to clean up the local functions a bit. Maybe adding the `clamp` and `between` would make simple functions in `gears.math` as well?

actionless:
yup, nice idea

Veratil:
@actionless @Elv13 I'm not sure how to fix the error I'm getting now with `unpack` in the Travis test builds

lib/beautiful/gtk.lua
```diff
@@ -7,7 +7,7 @@
 ---------------------------------------------------------------------------
 local get_dpi = require("beautiful.xresources").get_dpi
 local gears_debug = require("gears.debug")
-local gears_math = require("gears.math")
+local colorful = require("beautiful.colorful")
 local join = require("gears.table").join
 local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
```
actionless: make the same comment hint as here

Elv13:
Nice work. I will do a full review of next weekend if time permit. On the surface it looks like it's getting there.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+    if l <= v and v <= h then return true end
+    return false
+end
+
+-- Expects Hue [0,360], Saturation [0,1], Lightness [0,1]
+-- Does NOT check bounds
+local function hsl_to_rgb(hue, saturation, lightness)
+    -- If s = 0 then it's a grey
+    if saturation == 0 then
+        return unpack({lightness, lightness, lightness})
```
psychon: (Same comment applies to all other uses of `unpack` in this file)
Why is it done like this?
The above is equivalent to `return lightness, lightness, lightness`.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+    elseif 300 <= hue and hue <= 360 then
+        r,g,b = C,0,X
+    end
+    return unpack({r+m, g+m, b+m})
+end
+
+-- Expects Red, Blue, and Green [0,1]
+-- Does NOT check bounds
+local function rgb_to_hsl(red, green, blue)
+    local h,s,l = 0,0,0 -- luacheck: ignore
```
psychon: Why the `luacheck: ignore`? I guess it is warning about unused values that are assigned here and in fact it is right. What's wrong with just writing `local h,s,l` without an assignment?

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+    elseif 240 <= hue and hue <= 300 then
+        r,g,b = X,0,C
+    elseif 300 <= hue and hue <= 360 then
+        r,g,b = C,0,X
+    end
+    return unpack({r+m, g+m, b+m})
+end
+
+-- Expects Red, Blue, and Green [0,1]
+-- Does NOT check bounds
```
psychon: Maybe this function and `hsl_to_rgb` should have a comment on where the algorithms that are used here come from. Are the formulas just copied from Wikipedia?

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+local function create_color_object(r,g,b,h,s,l,a)
+    local o = setmetatable({
+        red = r, green = g, blue = b,
+        hue = h, saturation = s, lightness = l,
+        alpha = a
+    }, color_mt)
+    -- allow as functions to use color object variables
+    setmetatable(o.as, {
+        __index = o
+    })
```
psychon: Uhm. I'm not quite sure what you expect this to do, but it does not work. This just does `setmetatable(colorful.as, { __index = 0 })`. Thus, the metatable of `colorful.as` will just be whatever color was created last. Surely, that's not what you intend this to do, right?

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+        elseif val.hue ~= nil and val.saturation ~= nil and val.lightness ~= nil then
+            local alpha = val.alpha or 1
+            return color.from.hsla(val.hue, val.saturation, val.lightness, alpha)
+        -- do we have a table of 3 or 4 values?
+        elseif #val == 3 or #val == 4 then
+            -- if val[1] [0,360] and val[2] and val[3] [0,1] we deduce HSL
+            if (between(val[1], 0, 360) and
+                between(val[2], 0, 1)   and
+                between(val[3], 0, 1)      )
+            then
```
psychon: This also does not work. If I specify 100% red, 0% blue, 0% green, this code will interpret it as hsl instead.

Why is this "guessing stuff" even needed?

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+                )
+            end
+            -- if values aren't in range, escape
+        end
+        -- don't know the table type, escape
+    end
+    -- check for C userdata (GdkRGBA) with values
+    if type(val) == "userdata" then
+        if val.red ~= nil and val.green ~= nil and val.blue ~= nil then
+            local alpha = val.alpha or 1
```
psychon: `GdkRGBA` always has a `.alpha` field. This `or 1` is dead code.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+    -- not a valid type, quit
+    return nil
+end
+
+-- }}} color.new()
+
+-- {{{ String
+-- {{{ Helper tables and functions
+
+local css = {
```
psychon: Sigh. Why do we need a (partial) CSS parser? Who are the intended users and why can they not be just told to use one thing, instead of "pick one out of the following 100 possibilities"? :-(
I would guess that this is needed for the Gtk theme, but so far that theme seemed to have survived without parsing CSS color specifications.

Would it make sense to have that in a separate file to make this huge file smaller?

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+    integer = "%d?%d?%d",   -- 0-999 (but 0-255 for values)
+    --percent = "%d?%d?%d%%", -- 0-999% (but 0-100% for values)
```
psychon: Why this commented line?

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+    eol     = "%s*%)$",     -- functional EOL
+}
+-- capture a string
+local function capt(s)
+    return "(" .. s .. ")"
+end
+-- match string followed by %
+local function perc(s)
+    return s .. "%%"
+end
```
psychon: The above two functions are meant to improve readability, right? I am not sure if it would be easier to just write the strings directly in the patterns, but I'll be quiet on this...

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+local color_patterns = {
+    -- {{{ Hexadecimal patterns
+    -- shorthand #rgb
+    { "hexs", "^#(%x)(%x)(%x)$" },
+    -- shorthand #rgba
+    { "hexsa", "^#(%x)(%x)(%x)(%x)$" },
+    -- longform #rrggbb
+    { "hexd", "%#(%x%x)(%x%x)(%x%x)$" },
+    -- longform #rrggbbaa
+    { "hexda", "%#(%x%x)(%x%x)(%x%x)(%x%x)$" },
```
psychon: Why does this not use `capt()`?

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+    -- {{{ rgb(d, d, d) integers 0-255
+    { "rgbi", table.concat({
+        css.rgb, -- functional header
+        capt(css.integer), -- red
+        css.comma, -- separator
+        capt(css.integer), -- green
+        css.comma, -- separator
+        capt(css.integer), -- blue
+        css.eol -- EOL
+        }, "")
```
psychon: Is the `table.concat` really necessary? I'd just create the string directly.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+                end
+                -- check for alpha
+                if ta ~= nil then
+                    a = tonumber(m[4])
+                    if ta == "p" then
+                        a = a/100
+                    end
+                    a = a * 255 -- both float and percentage need this
+                end
+                if r == nil or g == nil or b == nil or a == nil then return nil end
```
psychon: Why this line? The `hex` case does not have it.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+    end
+    return nil
+end
+
+-- }}} String
+
+-- {{{ Decimal
+
+--- Create a new color instance from an RGB decimal value.
+-- Will bitshift decimal 2 bytes left and call `color.from.hex4`
```
psychon: This line is not documentation, it describes the implementation. The user should not have to care about that, I think.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+-- @tparam number hex A decimal describing an RGB color
+-- @return A new color object, or nil if invalid
+-- @usage local c = colorful.color.from.hex3(0xffcc99)
+function color.from.hex3(hex)
+    return color.from.hex4(lshift(hex, 8) + 0xff)
```
psychon: Uhm... this comment is on a random place, but what happens when I specify a color with more than 50% red as a hex number with Lua 5.3 on a 32 bit system? In this case, the number would be stored in an `int` and we get an integer overflow. I'm not quite sure what would happen next, but I would guess "not the right thing".

I'm not quite sure what to do about this. The only option that I see is: `hex3` (in the current implementation) and `hex4` (in general) cannot work everywhere.

(Sigh, I'd really like to stay outside of the realm of worrying about integer overflows...)

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+        while h < 0 do
+            h = h + 360
+        end
+        while h > 360 do
+            h = h - 360
+        end
+    end
+    s = clamp(s, 0.0, 1.0)
+    l = clamp(l, 0.0, 1.0)
+    if s == nil or l == nil then return nil end
```
psychon: I think this is dead code. If one of them is `nil`, then `clamp` or `between` would have errored out. (Same for `from.rgba`)

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+end
+
+--- Returns the inverse of a color
+-- @return A new color object representing the inverted color, or nil if invalid
+-- @usage local inverted_color = c:invert()
+function color:invert()
+    local red = clamp(1 - self.red, 0, 1) * 255
+    local green = clamp(1 - self.green, 0, 1) * 255
+    local blue = clamp(1 - self.blue, 0, 1) * 255
+    return color.from.rgba(red, green, blue, self.alpha * 255)
```
psychon: Why the `clamp`? If this actually changes anything, then the passed-in color was invalid. All the color-creation functions seem to be checking for that and clamping happens there.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+-- @return A new color object representing the complementary color, or nil if invalid
+-- @usage local complemented_color = c:complement()
+function color:complement()
+    local hue = self.hue + 180
+    return color.from.hsla(hue, self.saturation, self.lightness, self.alpha)
+end
+
+--- Returns the left and right analogous colors (30deg hue shift)
+-- @return A new color object representing a -30 degree hue shift of the color, or nil if invalid
+-- @return A new color object representing a +30 degree hue shift of the color, or nil if invalid
```
psychon: Wait, why `invalid`? Isn't hue an angle that "wraps around"?

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+            self.alpha),
+        color.from.hsla(
+            right,
+            self.saturation,
+            self.lightness,
+            self.alpha)
+end
+
+--- Returns three colors making up the tetradic colors (base complement, base
+--+ 60deg, base + 60deg complement)
```
psychon: So, complement is a 120 deg rotation... So we have here `+180`, `+60`, and `+60+180=+240`... Right?
Is this worth mentioning explicitly in the docs?

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+-- @treturn number The decimal represention of the color including the alpha
+-- channel
+-- @usage local deca_color = c.as:deca()
+function color.as:deca()
+    return math.floor(
+          lshift(round(self.red   * 0xFF), 24)
+        + lshift(round(self.green * 0xFF), 16)
+        + lshift(round(self.blue  * 0xFF), 8)
+        +        round(self.alpha * 0xFF)
+    )
```
psychon: This is also one of the functions which could have problems with 32 bit signed integers... I'm not sure.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+color_mt.__unm = color.invert
+
+--- Add the RGBA values of two colors
+-- @function color.__add
+-- @tparam lhs beautiful.colorful.color A color object
+-- @tparam rhs beautiful.colorful.color A color object
+-- @treturn beautiful.colorful.color A new color object with RGBA values added
+-- @usage colorful.color("#123") + colorful.color("#321") -- returns #444444
+color_mt.__add = function(lhs, rhs)
+    if type(rhs) ~= "table" then return nil end
```
psychon: Would `error` be more appropriate than `nil`? Actually... how about just dropping this check? There are enough tables which still cannot be added, so why special case this?

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+    )
+end
+
+--- Compare the lightness value of two colors
+-- @function color.__lt
+-- @tparam lhs beautiful.colorful.color A color object
+-- @tparam rhs beautiful.colorful.color A color object
+-- @treturn boolean True if lhs lightness is less than rhs lightness
+-- @usage colorful.color("#ffcc99") < colorful.color("#aabbcc") -- returns false (0.8 < 0.73333)
+color_mt.__lt = function(lhs, rhs)
```
psychon: Using the lightness for comparsion seems arbitrary... Why the lightness? People can just compare `.lightness` directly, if they want.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+-- @tparam number amount A number [0,1] of how much of color1 should be mixed
+-- with color2.
+-- @treturn string The string representation of the mixed color in `#rrggbbaa`
+-- format.
+-- @see beautiful.colorful.color.mix
+-- @usage local mix_str = colorful.mix("#ffcc99", "#99ccff", 0.5)
+function colorful.mix(color1, color2, amount)
+    local c1 = color.new(color1)
+    local c2 = color.new(color2)
+    if c1 == nil or c2 == nil then return nil end
```
psychon: Would it make sense to have this cause errors instead of returning `nil`? `nil`s might mean that some background color is just silently ignored, which is a bit more confusing than "I get an error with a traceback".

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+-- @param val A CSS color compatible string, decimal color value, or a table
+-- containing 3 or 4 values.
+-- @return A new color object, or nil if invalid
+-- @usage local c = colorful.color.new({30, 0.8, 0.6})
+-- -- For more examples see the top of this page
+function color.new(val)
+    if type(val) == "string" then return color.from.string(val) end
+    if type(val) == "number" then
+        if val > 0xFFFFFF then return color.from.hex4(val) end
+        return color.from.hex3(val)
```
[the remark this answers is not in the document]
actionless: mb also add such case to the testsuite?

lib/beautiful/colorful.lua
actionless: i think it's fine, since for user such details could be important (different color libs could be doing that slightly differently)

lib/beautiful/colorful.lua
Veratil: I thought I remembered seeing it used in other places in the API, I could be mistaken. Not a problem to change it.

lib/beautiful/colorful.lua
Veratil: Yeah luacheck was complaining about it for some reason, so I put it there. I think I tried removing the assignment and the unused error was still there. I'll try and see what it says (I just recently updated luacheck to the newest version from 0.19, this was done before that).

lib/beautiful/colorful.lua
Veratil: The formulas are converted from the definitions on Wikipedia, correct.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+    end
+    if type(val) == "table" then
+        -- do we have red/green/blue/alpha fields?
+        if val.red ~= nil and val.green ~= nil and val.blue ~= nil then
+            local alpha = val.alpha or 255
+            return color.from.rgba(val.red, val.green, val.blue, alpha)
+        -- do we have hue/saturation/lightness/alpha fields?
+        elseif val.hue ~= nil and val.saturation ~= nil and val.lightness ~= nil then
+            local alpha = val.alpha or 1
+            return color.from.hsla(val.hue, val.saturation, val.lightness, alpha)
```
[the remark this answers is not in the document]
Veratil: So it's a little bit of mix between hexadecimal definition (RGBA) and HSLA is defined, and I wanted to try to keep parity with that as best I could. Internally the numbers will all be `[0,1]` because that's how the RGB to HSL function works (RGB needs to be `[0,1]` and vice versa).

lib/beautiful/colorful.lua
Veratil: I wanted a single entry point into defining colors, and this is a way to handle that.

lib/beautiful/colorful.lua
Veratil: Correct, but this isn't checking for the specific `GdkRGBA` table, just if it has red, green, and blue. I had to add this because I forgot that in the GTK module the color is a userdata type. This will allow any userdata struct with red, green, and blue defined in some weird corner case we have no way to foresee. It may be never be hit, but it doesn't hurt in my opinion.

lib/beautiful/colorful.lua
Veratil: I figured it would be helpful to have. It never hurts to be able to support things, right?

I have no problem splitting it up if you want though.

lib/beautiful/colorful.lua
Veratil: I forgot to delete it after I changed to creating the strings through some extra descriptor functions.

lib/beautiful/colorful.lua
Veratil: Readability is correct. Since the integer and percent is the same, only adding the `%%` I could keep integer and capture that one but require the percent sign afterwards.

lib/beautiful/colorful.lua
Veratil: They are short enough to me that they're readable this way.

lib/beautiful/colorful.lua
Veratil: I wasn't sure what would be easier to read, to me this format is. I have no problem changing it to string concat though.

lib/beautiful/colorful.lua
Veratil: You're right! Mistake on my part, I can add the check to the `hex` section.

lib/beautiful/colorful.lua
Veratil: Describing what it does helps the user understand what they get. 🙂

If we feel it's not necessary I can remove it. Not a problem.

lib/beautiful/colorful.lua
Veratil: If you passed in `0x80ffff` then that's 24 bits right? `0x80ffff << 8 == 0x80ffff00`, `0x80ffff00 + 0xff == 0x80ffffff`. I don't see overflow?

lib/beautiful/colorful.lua
Veratil: You're right. I left it there as a holdover before I changed it to use `clamp`. Can be removed.

lib/beautiful/colorful.lua
```diff
@@ -0,0 +1,1395 @@
...
+-- @section color_change
+
+--- Update the HSL values after editing any of the RGB values
+function color:hsl_update()
+    self.hue, self.saturation, self.lightness = rgb_to_hsl(self.red, self.green, self.blue)
+end
+
+--- Update the RGB values after editing any of the HSL values
+function color:rgb_update()
+    self.red, self.green, self.blue = hsl_to_rgb(self.hue, self.saturation, self.lightness)
```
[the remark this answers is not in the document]
Veratil: I stuck this in there thinking about the ability to change the object directly. I wanted to have it update automagically, but I couldn't figure out the metamethods to do it right. I was hoping to get feedback on it. I wasn't sure if we want to keep the color immutable or allow changes (which would update the other colorspace fields automagically).

lib/beautiful/colorful.lua
Veratil: So this was in case somehow the red/green/blue values were updated to be outside `[0,1]`. Since the fields can be edited directly (right now), this was to prevent invalid colors from being created.

lib/beautiful/colorful.lua
Veratil: If somehow the color creation goes wrong then it's a fallback. That's what it means.

lib/beautiful/colorful.lua
Veratil: Yeah that could be clarified.

lib/beautiful/colorful.lua
Veratil: It's a simple check just to make sure you're not adding a number or string.

lib/beautiful/colorful.lua
Veratil: It was just a thought. I figured comparing colors with < and > would be comparing lightness. Can be removed.

lib/beautiful/colorful.lua
Veratil: To me I'd like it to just return nil as it shows something went wrong and the user can catch that and use a fallback or error on their own if it's important enough. But if you think error'ing is better here then that's fine.

lib/beautiful/colorful.lua
psychon: I'll leave this for others to decide. @Elv13 ?

Sorry for the huge comment dump. Initially, I expected just one comment and then another one showed up and anther one and...

lib/beautiful/colorful.lua
psychon: Those 24 bits get turned into `color.from.hex4(0x80ffffff)`. Gdb says that's `-2130706433` (oh! not only 32 bit Linux has 32 bit int, but 64 bit has, too!). So, the overflow will be in the call to `lshift`.

lib/beautiful/colorful.lua
psychon: Well... the function first creates the object as a table and uses `color_mt` as the meta table. No `.as` field is created here. Then it access `.as` and sets the meta-table on that. Since there is no `.as` field, the `__index` meta"function" is used. Thus, we look at `color_mt.__index = color`. So, this just indexes `color`. So we end up with the value of `color.as`, which is the table that is created in line 112: `local color = { from = {}, as = {}}`. At no point is this table copied, so there is no per-object `.as`, but only the global one. So, this must modify the global `.as` table.

What exactly do you mean with "the locally created table field 'as'"?

Veratil: I thought I understood how the metatables work here. I just tested and you're right it's not working as I had it before. This happened when I tried to pull the code out of the `from.rgba` and `from.hsla` to make it more generalized. I guess that didn't work.

I'm not sure how to make the `as:output()` functions work on the object. They give me the error that it attempts to perform arithmetic on a nil value, `red`, for example, since it's not in the `as` table fields. I don't know how to make that work with the metatables.

lib/beautiful/colorful.lua
Veratil: No worries about the comments, it's just giving me more things to look at and consider as I try to help and create/fix things. :)

lib/beautiful/colorful.lua
Veratil: If an integer overflow happens, does it throw a signal that awesome catches or what? How does Lua handle that?

If Lua ignores and just returns the negative number, then it should be okay since the bits are all we're looking at and those should be okay. We're not worried about the whole number itself as we do more bit shifting afterwards to get the values.

psychon: There is no signal. Also, I am not aware of any guarantees that Lua makes. In C, integer overflows are undefined behaviour, meaning that anything can happen when one occurs. There are examples "out there in the internet" were the compiler optimised away some security check due to an integer overflow, i.e. opened up an exploitable security issue. Since Lua is implemented in C, I guess that this also means that integer overflows are undefined behaviour, but there is way less opportunity for the compiler to wreck havoc here...

Also, are you sure that we are okay since this code just looks at the bit? Shifting right by `r` bits is implemented by division through `2^r`. This does not work for negative numbers since the result will still be negative. I vaguely remember some comment in this code that assumed that shifting right results in zeros in the high bits...
I cannot say that anything is definitely wrong, but I am just feeling uneasy about these bit operations.

Veratil: I think you're right. The division will keep the high bits.

If we keep this (which I would like to do) then we should put a large warning sign for these functions then when using on 32bit systems.

Veratil: And the reason it uses division is because that's how Lua 5.2 defined the operation in their docs.

Veratil: Looking through the source code for Lua 5.1, it looks like the default type for `LUA_NUMBER` is `double`. Do you think we're safe with that?

psychon: Everything but Lua 5.3 uses `double`. If I'm not mis-reading Wikipedia, a `double` has a 52 bit mantissa. That should be quite enough. :-)

Veratil:
Bump for attention. Not sure why Travis never worked.

psychon:
From a quick look (assuming I already took a better look before): Looks basically fine to me. :+1: 

Could you rebase this on master to get codecov to work (so one can look at the coverage of this) and to incorporate the "Travis rename"?

Aire-One:
I through this would be merged long ago. :open_mouth: 

What is the status?

psychon:
> What is the status?

From a quick look: I asked for a rebase to master half a year ago and then Emmanual wrote

> I started working on an updated version and will submit a pull request to this pull request when I am done (yes, GitHub supports that)

So, this PR is now waiting on either Emmanual or @Veratil to do something (but I guess it is mostly blocked by Emmanual and @Veratil cannot do much).

SethBarberee:
>     * Add a `beautiful.colorful` template to the `tests/examples` framework
> 
>     * Remove the hardcoded examples from the public doc, unit test them, add images and insert them back into the doc.

Don't know how far you are with this, @Elv13 but I have a very basic one that's working for me... 

I can PR whichever but just don't want all the effort to go to waste.

Elv13:
> I can PR

Please do so, I am ashamed of how much time I cannot find to finish this. Right now I have literally no time at all. I am already way, way late on this and this PR is a very nice contribution to AwesomeWM.

Veratil:
> Don't know how far you are with this, @Elv13 but I have a very basic one that's working for me...
> 
> I can PR whichever but just don't want all the effort to go to waste.

Let me know if you can't make a PR against my fork and I'll try to adjust permissions.

I've been pretty busy with new job, and unfortunately not using Awesome since I can't get it to play well with XFCE at work, and just don't have the time to work on it yet.

SethBarberee:
> > Don't know how far you are with this, @Elv13 but I have a very basic one that's working for me...
> > I can PR whichever but just don't want all the effort to go to waste.
> 
> Let me know if you can't make a PR against my fork and I'll try to adjust permissions.
> 
> I've been pretty busy with new job, and unfortunately not using Awesome since I can't get it to play well with XFCE at work, and just don't have the time to work on it yet.

Awesome, I'll try to do a PR to the branch later this week, @Veratil.

SethBarberee:
Totally let this slide away. Finally pushed my basic test framework for colorful. Feel free to comment on it, @Elv13, over here: https://github.com/Veratil/awesome/pull/1

Issue #2260, opened 2018-05-05
Enable Mergify for dev workflow


jd:
Hey team!

The current pull request configuration is pretty nice and has coverage, CI run, etc and uses code review. It seems to me there could be a large piece of automation that could be setup, and that's automatic merging based on those criteria.

I'd suggest to use Mergify for that on this repository: https://mergify.io/ which allows to do that easily.

WDYT?

(Full disclosure: I'm behind the service :)

actionless:
can the service also take a PR label into account?

> Is this documented?
do you know any such tools which could be appliable to the project? it could be quite nice to see not only tests- but also documentation-coverage

actionless:
> What would you like to do with a label?

right now we have a label "PRs to be merged soon", i think having that label on a PR could be a good condition, meant what a PR confirmed by a human-review but pending some auto-checks or so

> If doc are covered by the current jobs, 
that's what i was asking -- if you know any tools for doc coverage which we could apply to awesome. it will be really amazing to start using something like that

jd:
> right now we have a label "PRs to be merged soon", i think having that label on a PR could be a good condition, meant what a PR confirmed by a human-review but pending some auto-checks or so

You actually don't need that. Those conditions are already configured on the Mergify side: you'd just have to write some rules like "2 human should review + all CI tests must pass" and that'll be enough.

> that's what i was asking -- if you know any tools for doc coverage which we could apply to awesome. it will be really amazing to start using something like that

I don't – though I don't think it's related to Mergify usage though here.

actionless:
> I don't – though I don't think it's related to Mergify usage though here.

it's indeed unrelated, i just thought you did some kinds of research of the "market" of such tools so could give some guidance on that

actionless:
let's wait for other feedback, my role in the project is too minimal for making such kind of decisions

jd:
Ping @awesomeWM/write-access :)

blueyed:
+0 from me.
Often I think it's good for somebody with write access to edit and then squash-merge the PR, i.e. some minimal human intervention often makes sense.

psychon:
+0.5 from me since I imagine this makes me less important.

[...]

PR #2277, opened 2018-05-30
Add configuration for https://mergify.io

psychon:
Fixes: https://github.com/awesomeWM/awesome/issues/2260

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,12 @@
+rules:
+  default:
+    protection:
+      required_status_checks:
+        strict: True
+        contexts:
+          - continuous-integration/travis-ci
+          - Codacy/PR Quality Review
+          - codecov/patch
+          - coverage/coveralls
+      required_pull_request_reviews:
+        required_approving_review_count: 2


psychon:
Seems like Codacy recently-ish renamed the status that it reports and I copied the name from some too old PR, will do a force push to update to `Codacy/PR Quality Review`.

blueyed:
@jd 

From the statuses:

> mergify/pr — .mergify.yml is missing

Would it make sense to consider / use the one from this (or rather the PR itself in general) already?

blueyed:
> Would it make sense to consider / use the one from this (or rather the PR itself in general) already?

Scratch that - thought it would be bad myself, and then read up on the issue.

jd:
I've opened https://github.com/Mergifyio/mergify-engine/issues/31 as per @Elv13 request.
If there's any other feature you'd like to see feel free to open issues.
(Not sure at this stage if there's anything left to prevent merging this?)

jd:
@psychon Your understand is correct.

jd:
@Elv13 you're good with that now?

Elv13:
@jd Have all issues raised in the other thread are addressed or at least have concrete and scheduled plan to fix them? Some of them are more theoretical and not an immediate threat. However the one like having your own code to detect if a PR is mergeable instead of trusting GitHub wont let you merge it otherwise it one I consider more serious given its a behavior known to change from time to time. I don't have time to do another review, so I will trust you do making a summary of the progress.

jd:
@Elv13 Yeah, they have been addressed or bugs have been opened but I think we've done most of them by now.

GitHub data are only used to know what's the state of the branches and the what the status check (CI) reports, and that's what actually is reported by the UI you see. The way you talk about that makes me think that you imagined Mergify would continuously try to merge until GitHub lets him succeed? If that's what you understood, that's not definitely never has been the case. :)

jd:
Let's do this. Ping me if you need any help.

blueyed:
:fireworks: :)

PR #2278, opened 2018-05-31
Default config: Unify mouse dragging behavior

psliwka:
When moving or resizing a client by dragging its titlebar with mouse,
the client is automatically focused and raised. However, this behavior
is not replicated by tne other dragging method, i.e. grabbing it with
mouse while pressing the modkey. Sometimes, this causes a client to
disappear below another one while moving.

This commit unifies handling of dragging clients with mouse, by ensuring
that the client is focused and raised, whichever dragging method is
used. The fix applies both to moving and resizing clients.

Signed-off-by: Piotr Śliwka <contact@psliwka.info>

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -458,7 +458,10 @@ end
 
 -- @DOC_CLIENT_BUTTONS@
 clientbuttons = gears.table.join(
-    awful.button({ }, 1, function (c) client.focus = c; c:raise() end),
+    awful.button({ }, 1, function (c)
+        client.focus = c;
+        c:raise()
+    end),
     awful.button({ modkey }, 1, function (c)
         client.focus = c
         c:raise()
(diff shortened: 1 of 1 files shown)


awesomerc.lua
```diff
@@ -459,8 +459,17 @@ end
 -- @DOC_CLIENT_BUTTONS@
 clientbuttons = gears.table.join(
     awful.button({ }, 1, function (c) client.focus = c; c:raise() end),
```
alfunx: Not important, but you could format the first binding like the two below.

psliwka: Yeah, I've been thinking about it, then I decided to follow this contributing guideline:

> do not try to fix style of code you are not writing, it's just adding noise for no gain;

Still, I can fix this one, provided that's fine with maintainers.

alfunx: Generally I'd agree, if one does lots of unrelated style fixes, it adds noise. Though in this case, it's related to the default config, which is mostly the only file that users will see and modify, so it should be as nice and consistent as possible IMO.

psliwka: Addressed in a9f871b.

PR #2280, opened 2018-06-03
Add a container window around clients

psychon:
Apparently, Gtk started giving its windows a ParentRelative background.
This means that Gtk's windows can only be reparented into windows with
the same depth. Since we want working transparency in titlebars, we
however use a window with depth 32 to reparent clients into. Thus,
ReparentWindow would fail and Gtk's window would just be always visible.

We fix this by creating a useless container window of the same depth as
the client window. Instead of reparenting client windows directly into
the window used for titlebars, they are reparented into this container
window and the container window is put into the window used for
titlebars.

Let's see how long it takes someone to find some program that gets
confused when the WM reparents its window into a window contained in
another window, instead of just one window...

@alfunx Could you test this? Sorry, but it has some differences to the earlier patch you tested (aka "it should hopefully be better").

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1409,6 +1409,7 @@ pick_visual_for_depth (uint8_t depth)
 void
 client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_attributes_reply_t *wattr)
 {
+    xcb_void_cookie_t reparent_cookie;
     lua_State *L = globalconf_get_lua_State();
     const uint32_t select_input_val[] = { CLIENT_SELECT_INPUT_EVENT_MASK };
 
@@ -1502,9 +1503,11 @@ client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_at
                                  XCB_CW_EVENT_MASK,
                                  no_event);
     if (c->container_window != XCB_NONE)
-        xcb_reparent_window(globalconf.connection, w, c->container_window, 0, 0);
+        reparent_cookie =
+            xcb_reparent_window_checked(globalconf.connection, w, c->container_window, 0, 0);
     else
-        xcb_reparent_window(globalconf.connection, w, c->frame_window, 0, 0);
+        reparent_cookie =
+            xcb_reparent_window_checked(globalconf.connection, w, c->frame_window, 0, 0);
     xcb_map_window(globalconf.connection, w);
     xcb_change_window_attributes(globalconf.connection,
                                  globalconf.screen->root,
@@ -1601,6 +1604,14 @@ client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_at
     luaA_object_emit_signal(L, -1, "manage", 0);
     /* pop client */
     lua_pop(L, 1);
+
+    xcb_generic_error_t *error = xcb_request_check(globalconf.connection, reparent_cookie);
+    if (error != NULL) {
+        warn("The following error happened while reparenting a window, window will not be managed properly!");
+        event_handle((xcb_generic_event_t *) error);
+        xcb_unmap_window(globalconf.connection, w);
+        p_delete(error);
+    }
 }
 
 static void
--- a/objects/client.c
+++ b/objects/client.c
@@ -1270,9 +1270,19 @@ client_geometry_refresh(void)
         xcb_configure_window(globalconf.connection, c->frame_window,
                 XCB_CONFIG_WINDOW_X | XCB_CONFIG_WINDOW_Y | XCB_CONFIG_WINDOW_WIDTH | XCB_CONFIG_WINDOW_HEIGHT,
                 (uint32_t[]) { geometry.x, geometry.y, geometry.width, geometry.height });
-        xcb_configure_window(globalconf.connection, c->window,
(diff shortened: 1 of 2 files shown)


psychon:
I added another commit which would make the original problem easier to debug.
Also, Travis points out that somehow this change already breaks `test-maximize.lua` and `test-gravity.lua`. Meh...

psychon:
Argh, I forgot to select `SubstructureRedirect` on the container window, so windows were allowed to resize themselves without any involvement of the WM. Fixed by the following patch which I will amend into the first commit.
```diff
diff --git a/objects/client.c b/objects/client.c
index 6ad830507..4b1c55797 100644
--- a/objects/client.c
+++ b/objects/client.c
@@ -1477,10 +1477,11 @@ client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_at
         c->container_window = xcb_generate_id(globalconf.connection);
         xcb_create_window(globalconf.connection, wgeom->depth, c->container_window, c->frame_window,
                 0, 0, wgeom->width, wgeom->height, 0, XCB_COPY_FROM_PARENT,
-                visual, XCB_CW_BACK_PIXMAP | XCB_CW_BORDER_PIXEL | XCB_CW_COLORMAP,
+                visual, XCB_CW_BACK_PIXMAP | XCB_CW_BORDER_PIXEL | XCB_CW_EVENT_MASK | XCB_CW_COLORMAP,
                 (const uint32_t[]) {
                     XCB_BACK_PIXMAP_NONE,
                     globalconf.screen->black_pixel,
+                    XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT,
                     colormap
                 });
         xcb_map_window(globalconf.connection, c->container_window);
```

alfunx:
@psychon Just wondering if I understood this correctly: If the GTK people consider gitlab.gnome.org/GNOME/gtk/issues/1134, this fix will not be needed, right?

psychon:
Yup, the fix would not be needed. At least the second commit in here is useful in general since it improves the failure mode. And for the first commit, someone could argue that this makes awesome more compatible with weird apps. I do not know how much of a performance or memory usage difference these patches would make...

alfunx:
> could argue that this makes awesome more compatible with weird apps.

Oh, that's probably right.

> I do not know how much of a performance or memory usage difference these patches would make...

That was my concern, but I guess the overhead is really not much/measurable.

psychon:
What's the way forward for this? Would it sound arrogant to say that I am the only one around here who could actually review this, hence waiting for proper review might be a bit hard?

I guess the real problem is whether we want to fix / work-around this, or if we just hope that this is handled in GTK instead.

Thoughts? Life signs? Anything?

Elv13:
Looks like it wont be necessary. While the commit isn't merged upstream, it seems to be somewhere in the pipeline. Feel free to merge it if you think it makes Awesome better or more robust. But if it doesn't really do anything for non-broken clients, then maybe not because there is no point to support a few GTK commits with an acknowledged regression.

actionless:
isn't this commit fixes this already? https://gitlab.gnome.org/GNOME/gtk/commit/01d1bc3c75fd0eff5665f5b9c690c5e1e6c65f13

Elv13:
> isn't this commit fixes this already?

It doesn't seems to me merged and/or in a pull request right now

actionless:
@Elv13 it's written what the commit is currently on the following branches:

wip/gtk324.win.egl
324.win32.build
gtk-3-24
issue1170

and, what is more important, tagged as `3.23.0 `

Elv13:
> gtk-3-24

Ah, ok that makes sense. `master` is probably GTK4 and 3-24 is the unstable branch of GTK3

actionless:
3.23 is unstable and 3.24 is gonna be stable

Elv13:
Is this PR still valid or should it be closed now that the invalid windows never create a client? Or is there any other cases where this is relevant?

psychon:
Well... I wouldn't say "never". In fact, I failed at finding some point in ICCCM to point at and say "Gtk master is violating the spec!". So, technically, this might still be necessary. However, "that other" PR of mine made this problem easier to debug by making it more prominent and printing an error message on stderr, so let's hope that this will be easy to figure out if it ever happens again.

Thus, closing this under the assumption that it is not currently required.

objects/client.c
```diff
@@ -1424,6 +1455,41 @@ client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_at
...
+                (const uint32_t[]) {
+                    XCB_BACK_PIXMAP_NONE,
+                    globalconf.screen->black_pixel,
+                    XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT,
+                    colormap
+                });
+        xcb_map_window(globalconf.connection, c->container_window);
+
+        if (colormap != globalconf.screen->default_colormap)
+            xcb_free_colormap(globalconf.connection, colormap);
```
xzfc: I see you free the colormap before the window is destroyed. Is it really allowed?

https://tronche.com/gui/x/xlib/color/XFreeColormap.html says the following:
> If the specified colormap is defined as the colormap for a window (by XCreateWindow(), [...]), XFreeColormap() changes the colormap associated with the window to None [...]. X does not define the colors displayed for a window with a colormap of None.

psychon: Well, it is definitely allowed, since you linked to documentation explaining what happens. ;-)

However, a window with colormap `None` and a depth different from its parent is indeed... "weird".

To be honest, I was expecting this to be similar to creating a pixmap, creating a render picture, and then destroying the pixmap. The picture keeps the pixmap alive for as long as needed.

PR #2294, opened 2018-06-28
awesome.kill: allow negative PIDs (and 0) to support process groups

blueyed:
From `kill(3p)`:

> If pid is greater than 0, sig shall be sent to the process whose
> process ID is equal to pid.

> If pid is 0, sig shall be sent to all processes (excluding an
> unspecified set of system processes) whose process group ID is equal to
> the process group ID of the sender, and for which the process has
> permission to send a signal.

> If pid is −1, sig shall be sent to all processes (excluding an
> unspecified set of system processes) for which the process has
> permission to send that signal.

> If  pid  is  negative,  but not −1, sig shall be sent to all processes
> (excluding an unspecified set of system processes) whose process group
> ID is equal to the absolute value of pid, and for which the process has
> permission to send a signal.

IIRC @Elv13 mentioned this somewhere, but I could not find it now.

### DIFF
--- a/luaa.c
+++ b/luaa.c
@@ -254,7 +254,8 @@ luaA_restart(lua_State *L)
 
 /** Send a signal to a process identified by its process id. See
  * `awesome.unix_signal` for a list of signals.
- * @tparam integer pid Process identifier
+ * @tparam integer pid Process identifier.  0 and negative values have special
+ *   meaning.  See `man 3 kill`.
  * @tparam integer sig Signal number
  * @treturn boolean true if the signal was successfully sent, else false
  * @function kill
@@ -262,7 +263,7 @@ luaA_restart(lua_State *L)
 static int
 luaA_kill(lua_State *L)
 {
-    int pid = luaA_checknumber_range(L, 1, 1, INT_MAX);
+    int pid = lua_tonumber(L, 1);
     int sig = luaA_checknumber_range(L, 2, 0, INT_MAX);
 
     int result = kill(pid, sig);


blueyed:
Hmm, without `client.pgid` you still need to shell out:

```lua
            {'sh', '-c',
             'pkill -CONT -g $(ps -o pgid= '..tostring(c.pid)..')'},
            done_cb)
```

Does it make sense to add `pgid` next to `pid`?
Or is there a way in awesome/Lua already?

luaa.c
```diff
@@ -262,7 +262,7 @@ luaA_restart(lua_State *L)
 static int
 luaA_kill(lua_State *L)
 {
-    int pid = luaA_checknumber_range(L, 1, 1, INT_MAX);
+    int pid = luaA_checknumber_range(L, 1, -INT_MAX, INT_MAX);
```
psychon: Just replace this with `lua_tonumber`?

psychon:
> Hmm, without client.pgid you still need to shell out:

So, are you saying you do not want this anymore?

> Does it make sense to add pgid next to pid?

Where would we get the process group id from? It's not in a property.

> Or is there a way in awesome/Lua already?

I don't know one. I do not even know a portable way to get the process group id in C. A linux-specific way would be the `/proc` filesystem.

psychon:
>   'pkill -CONT -g $(ps -o pgid= '..tostring(c.pid)..')'},

Uhm, does this work? According to its man page, `pkill` wants a comma-separated list, but this instead produces separate arguments / a whitespace separated list.

blueyed:
> 'pkill -CONT -g $(ps -o pgid= '..tostring(c.pid)..')'

`ps -o pgid= PID` returns a single number: the process pgid.
The pkill is used to select processes in this process group (-g/--pgroup), and sends SIGCONT to them.

> Plus, it only works if the client is actually the process group leader. This command, if I understand it correctly, lists all process ids for processes with the given process group id. It does not list all processes in the process group of the given pid.

No.. the pgid is the same for the leader and its childs.
I am using this with qutebrowser, which has QtWebEngineProc childs:

```
% chpids() { for p in $(ps --ppid $1 -o pid=); do echo $p; chpids $p; done; }
% ps -o pid,pgid,comm,state= 15418 $(chpids 15418)                                                                                                                  (velodrome@3.6.5)
  PID  PGID COMMAND
15418 15188 qutebrowser     S
15665 15188 QtWebEngineProc S
15692 15188 QtWebEngineProc S
..
```

You might have been confused by the `=` in the `ps` command?
`ps -o pgid= X` does not mean that pgid is X, but to output the pgid of X without the column header.

> So, are you saying you do not want this anymore?

I think it still makes sense - there are other use cases than this.

blueyed:
Using lua_tonumber now and added docs.

psychon:
> ps -o pgid= PID returns a single number: the process pgid.

```
$ ps -o pgid=$$ 
28549
28549
28606
```

psychon:
Oh wait, you have a space in there that I missed. `ps` is confusing...

blueyed:
I think this is good then?

blueyed:
> What's up with the merge commit?

Used the "Update branch" button on Github.

luaa.c
```diff
@@ -254,15 +254,16 @@ luaA_restart(lua_State *L)
 
 /** Send a signal to a process identified by its process id. See
  * `awesome.unix_signal` for a list of signals.
- * @tparam integer pid Process identifier
+ * @tparam integer pid Process identifier.  0 and negative values have special
+ *   meaning.  See `man 3 kill`.
```
jd: Rather 2. ;)

blueyed: I have it as `KILL(3P)` here (also), which I found more appropriate (given its more descriptive information IIRC).

PR #2297, opened 2018-07-12
Correct descriptions of border colors

sambanks:
The docs seem to incorrectly say that border_normal, border_focus and border_marked relate to width. These settings actually control color.

### DIFF
--- a/lib/beautiful/init.lua
+++ b/lib/beautiful/init.lua
@@ -75,15 +75,15 @@ local active_font
 --- The client border width.
 -- @beautiful beautiful.border_width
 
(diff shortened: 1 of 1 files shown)


Elv13:
@jd Shouldn't `Mergify.io` do something right now?

jd:
@Elv13 it will, you just felt in a very bad timing https://twitter.com/mergifyio/status/1017312898849308672 :)

jd:
This PR seems to hit a GitHub bug since it says "Required statuses must pass before merging" whereas everything is green…

Issue #2279, opened 2018-06-01
Firefox 60 / Thunderbird 52 weird positioning


cars10:
Output of `awesome --version`:

awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.2

**How to reproduce the issue:**

Start firefox 60 or thunderbird 52 in tile layout

**Actual result:**

Firefox/Thunderbird positions itself above the menubar

**Expected result:**

Window should be below the menu bar

Setting/Unsetting floating does not help either. Also it is present on all tags at once.

psychon:
With "menubar", do you mean the bar at the top of the screen in the default config?
Could you show the output of `/usr/bin/xprop` for one of these windows? (run `xprop` in a terminal, click on firefox' window)

> Also it is present on all tags at once.

Uhm, what?

cars10:
Yes, i mean the bar at the top.

And yes, when i open firefox 60 an switch to a different tag then the firefox window is still in the forground.

xprop:
```
GDK_TIMESTAMP_PROP(GDK_TIMESTAMP_PROP) = 0x61
_NET_WM_DESKTOP(CARDINAL) = 3
_NET_FRAME_EXTENTS(CARDINAL) = 2, 2, 2, 2
WM_STATE(WM_STATE):
		window state: Normal
		icon window: 0x0
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: True
		Initial state is Normal State.
		bitmap id # to use for icon: 0x2400014
		bitmap id # of mask for icon: 0x240001a
		window id # of group leader: 0x2400001
_GTK_THEME_VARIANT(UTF8_STRING) = 
_NET_STARTUP_ID(UTF8_STRING) = "rofi/|usr|lib|firefox|firefox/13129-0-cars10-thinkpad_TIME4397938"
WM_WINDOW_ROLE(STRING) = "browser"
_NET_WM_BYPASS_COMPOSITOR(CARDINAL) = 2
XdndAware(ATOM) = BITMAP
_NET_WM_ICON(CARDINAL) = 	Icon (32 x 32):
	                                
	                                
	                                
	           ░▒▒▒▒▒▒▒             
	     ▒   ▒▒▒▒▒▒▒▒▒░      ░      
	    ▒▒░░▓▒▒▒▒░▒▒▒▒░             
[...]

PR #2298, opened 2018-07-13
client_manage(): Check that reparenting works

psychon:
Testing done: Basically none (`tests/run.sh` did not complain)

------

When a new client appears, we reparent its window into a window of our
choice so that we can add decorations. However, windows can be difficult
making the reparenting fail. For example, some snapshots of GTK+ set a
parent-relative background, which means that its parent windows have to
have the same depth as the window itself. This makes reparenting fail if
transparency is used/supported.

Make the failure mode... different, by checking if reparenting
succeeded. If it failed, we print an error message and unmanage the
window immediately. This means that those GTK+ windows just do not
appear at all, instead of behaving weirdly.

This idea of this patch came from i3.

Related-to: https://github.com/awesomeWM/awesome/issues/2279

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1379,6 +1379,7 @@ client_update_properties(lua_State *L, int cidx, client_t *c)
 void
 client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_attributes_reply_t *wattr)
 {
+    xcb_void_cookie_t reparent_cookie;
     lua_State *L = globalconf_get_lua_State();
     const uint32_t select_input_val[] = { CLIENT_SELECT_INPUT_EVENT_MASK };
 
@@ -1436,7 +1437,7 @@ client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_at
                                  globalconf.screen->root,
                                  XCB_CW_EVENT_MASK,
                                  no_event);
-    xcb_reparent_window(globalconf.connection, w, c->frame_window, 0, 0);
+    reparent_cookie = xcb_reparent_window_checked(globalconf.connection, w, c->frame_window, 0, 0);
     xcb_map_window(globalconf.connection, w);
     xcb_change_window_attributes(globalconf.connection,
                                  globalconf.screen->root,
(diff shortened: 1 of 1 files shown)


objects/client.c
```diff
@@ -1533,6 +1534,14 @@ client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_at
     luaA_object_emit_signal(L, -1, "manage", 0);
     /* pop client */
     lua_pop(L, 1);
+
+    xcb_generic_error_t *error = xcb_request_check(globalconf.connection, reparent_cookie);
+    if (error != NULL) {
+        warn("Failed to manage some window, because reparenting failed.");
```
blueyed: Could we be more specific here instead of "some window"?

psychon: The window id will not help anybody. Should `c->name` be included in the message?

I think that when this triggers, it is obvious that something is wrong (you start firefox, no window appears, but a firefox process is running). The harder part would be for people to actually find this warning. But yes, that's no excuse to produce a bad warning.

blueyed: Ok with me then.

psychon:
> psychon dismissed stale reviews from blueyed and actionless via f3e62b1 just now

Diff to previous version:
```diff
diff --git a/objects/client.c b/objects/client.c
index a4d3c42ab..e1c82e2de 100644
--- a/objects/client.c
+++ b/objects/client.c
@@ -1532,16 +1532,18 @@ client_manage(xcb_window_t w, xcb_get_geometry_reply_t *wgeom, xcb_get_window_at
 
     /* client is still on top of the stack; emit signal */
     luaA_object_emit_signal(L, -1, "manage", 0);
-    /* pop client */
-    lua_pop(L, 1);
 
     xcb_generic_error_t *error = xcb_request_check(globalconf.connection, reparent_cookie);
     if (error != NULL) {
-        warn("Failed to manage some window, because reparenting failed.");
+        warn("Failed to manage window with name '%s', class '%s', instance '%s', because reparenting failed.",
+                NONULL(c->name), NONULL(c->class), NONULL(c->instance));
         event_handle((xcb_generic_event_t *) error);
         p_delete(&error);
         client_unmanage(c, true);
     }
+
+    /* pop client */
+    lua_pop(L, 1);
 }
 
 static void
```

objects/client.c
Elv13: Ah, right, I forgot about those

PR #2300, opened 2018-07-19
mergify: use merge_strategy.method = rebase

blueyed:
This should avoid having unnecessary merge commits.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -10,3 +10,5 @@ rules:
           - coverage/coveralls
       required_pull_request_reviews:
         required_approving_review_count: 2
+    merge_strategy:
+      method: rebase


blueyed:
@psychon 
I have a function for this: https://github.com/blueyed/oh-my-zsh/blob/b2d46dbcf1a4e7b012dbc5d38fe7e2ea53be0839/plugins/git/git.plugin.zsh#L234 - usually run it as `gbmcleanup -m` on the master branch.

@Elv13 
It is not squashing, but rebasing.
What meta data would be missing then?
I can see that it is good for bigger things (that go together) though.

@jd 
Is it possible to configure this to apply to single commits only?

Elv13:
> What meta data would be missing then?

Commits associated with a PR (without using external tools such as *name your web browser*) which is quite important to me. Sometime the commit message isn't obvious enough as to *why* a change was done, but if you look at it from its PR, then it makes more sense.

> I can see that it is good for bigger things (that go together) though.

That's the general idea, one line / one commit PR don't need merge commits, anything bigger does.

Anyhow, I opened issues on @mergifyio to talk about improvements related to this issue. Let's discuss that there as feedback from other mergify users is useful.

Please close this PR if you agree it needs a more fine grained solution instead of the one proposed by this config change. For now, I personally prefer noisy merge commits over losing metadata I often rely on. @psychon seems to also make use of the branching history.

blueyed:
I think we should disable mergify as it currently stands, since it creates even more merge commits now (due to merging master into PRs): https://github.com/awesomeWM/awesome/pull/2306#issuecomment-407007040

blueyed:
Like @jd pointed out (in https://github.com/awesomeWM/awesome/pull/2306#issuecomment-407062907) this is due to using strict mode.
Should we disable strict mode then for now?

psychon:
Feel free to submit a PR doing just that, if you want.

blueyed:
Done: #2319

PR #2293, opened 2018-06-20
Return nil for screen string index lookup instead of throwing error

Veratil:
There's often times when I'll have code to use a specific screen, and it always stops when the string lookup throws an error. This just makes it return nil instead so you can continue with the script (and handle nil return).

For example, in my rules I have a callback (which errors if the screen doesn't exist instead of putting it to the focused screen):
```lua
callback = function(c)
  c:move_to_tag(awful.tag.find_by_name(screen['DP-2'] or awful.screen.focused(), "EMAIL"))
end,
```

Also, I wasn't sure what value to return from this function, so I assumed 1.

PR #2304, opened 2018-07-20
fix(awful: widget: only_on_screen): update for screen API changes from #2293

### LINKED ISSUE #2302: tests/test-awful-widget-only_on_screen.lua causes warnings
See https://github.com/awesomeWM/awesome/pull/2293#issuecomment-404887235:

Since #2293, the code in `awful.widget.only_on_screen` is kind of broken and prints warnings. That needs to be fixed.

CC @Veratil @actionless @Elv13 @blueyed (Everyone who touched #2293)

You do not want me to go ahead and fix this. My fix would be a revert.

actionless:
fixes #2302

### DIFF
--- a/lib/awful/widget/only_on_screen.lua
+++ b/lib/awful/widget/only_on_screen.lua
@@ -25,10 +25,12 @@ local function should_display_on(self, s)
     if not self._private.widget then
         return false
     end
-    local success, own_s = pcall(function()
-        return capi.screen[self._private.screen]
-    end)
-    return success and own_s == s
+    local own_s = self._private.screen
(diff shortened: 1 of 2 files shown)


actionless:
@psychon  thanks for writing it :-)

psychon:
...but it definitely is. `tests/examples/wibox/container/defaults/only_on_screen.lua` breaks without that change. So this either Lua coverage reporting being wrong, or not working for the generate-the-doc-example-images tests.

Elv13:
I can attest that the coverage generally works for those files because my PRs are mostly using it as the sole source of coverage data (because I want the users to have example and duplicated coverage is a waste of time for everybody) and I usually get high coverage marks. So if it isn't hit by codecov, it's not a black or white situation.

Issue #1418, opened 2017-01-20
gaps in layouts caused by not calling or mis-calling client:apply_size_hints


dzhu:
This is a response to #1308 (but that's closed, so I figured it might be better to not respond directly there).

I believe I know the underlying reason for that issue. The `tile` layout calls `client:apply_size_hints` assuming that it accounts for the border size, which it does not. When `layout.arrange` calls `client:geometry`, it subtracts off the border size (and useless gap) first, which can cause hinted clients to jump down to a smaller size in the actual layout than was reported when `tile` called `apply_size_hints`, leaving a gap.

Meanwhile, the other layouts don't account for hints at all, so there are even more gaps.

Setting `honor_size_hints = false` sort of fixes the problem by forcing each client to fill the space originally computed by `tile`, but the old behavior lets the clients be their natural sizes and starts each client directly after the end of the previous one, which seems preferable.

Something like the patch linked below (based on bfb3534) might be useful; I've been testing it and it looks like it makes `tile` work properly as it used to. It doesn't deal with a nonzero `useless_gap`, though.

https://gist.github.com/dzhu/1d0889ad1419eb2b971ca5eb6b57d493

psychon:
Thanks for figuring this out and writing such a good bug report!
(Note that 3.5 didn't have this "try to work well with size hints" code at all, so there should also have been some kind of issue before)

dzhu:
Previously, `tile` would just call `client:geometry` directly, which would take care of the hints and return the appropriate size. But the change to delaying the actual sizing to later created a mismatch between what size `tile` thought would result and what actually happened, because now one place subtracts borders and one doesn't.

Elv13:
> Can we change apply_size_hints so that it does the subtraction itself or is that too much of an API break @Elv13.

It's an ABI ([B]ehavior, not [B]inary) break, but then again: it's a bug. So yes, I am fine with changing the existing function. However, `awful.placement.resize_to_mouse` (generalized code taken from `awful/mouse/init.lua`) also uses this. It will require to see if this breaks (fast look: probably) then add the offset there, add an unit test and update the documentation of the CAPI function. This is more work... @dzhu already made a great investigation here. Fixing the existing API would be nicer, but it might be a little too much to ask?

dzhu:
Hi, any updates on this? I've been running with my patch since opening the issue, and it's worked well, as far as I can tell. I can make a PR, if that would help, though it sounds like there might be more to consider first.

PR #2305, opened 2018-07-20
tile: Apply size hints correctly

psychon:
Layouts work with the client's geometry in "space on screen that is
assigned to this client". This means that the geometry should include
decoration (titlebar and borders) and useless gaps.

Everything else (especially the C code) works with client's geometry in
"space that the client can draw on". This means that the titlebar,
borders and the useless gaps are not included into this size.

Thus, when applying size hints, the tile layout has to convert between
these two representations. Otherwise, size hints are applied incorrectly
and to a wrong geometry.

Fixes: https://github.com/awesomeWM/awesome/issues/1418

CC @dzhu Just for your information (and sorry that it took so long, I have about 50 mails from GitHub marked as "eventually someone should do something about this". Pinging usually helps to boost priorities here. :-) Also, I did not go with your proposed patch since that one does not handle the useless gap and this could not be handled in C.)

-----

Testing done: Well, the test suite did not complain... (after I stopped misspelling `useless_gap`)

### DIFF
--- a/lib/awful/layout/suit/tile.lua
+++ b/lib/awful/layout/suit/tile.lua
@@ -158,7 +158,14 @@ local function mouse_resize_handler(c, _, _, _, orientation)
                           end, cursor)
 end
 
-local function tile_group(gs, cls, wa, orientation, fact, group)
+local function apply_size_hints(c, width, height, useless_gap)
+    local bw = c.border_width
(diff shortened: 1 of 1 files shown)


actionless:
@jd 
>  @mergify mergify/pr Pending — Waiting for status checks success

dzhu:
Great, thanks!

Issue #2299, opened 2018-07-17
Compiler warning in client_class_setup: cast between incompatible function types


blueyed:
Have just seen this warning when upgrading awesone-luajit-git (Arch Linux):

```
[ 15%] Building C object CMakeFiles/awesome.dir/objects/client.c.o
/home/user/.cache/pacaur/awesome-luajit-git/src/build/objects/client.c: In function ‘client_class_setup’:
/home/user/.cache/pacaur/awesome-luajit-git/src/build/objects/client.c:3662:29: warning: cast between incompatible function types from ‘int (*)(lua_State *)’ {aka ‘int (*)(struct lua_State *)’} to ‘int (*)(lua_State *, lua_object_t *)’ {aka ‘int (*)(struct lua_State *, struct <anonymous> *)’} [-Wcast-function-type]
                             (lua_class_propfunc_t) luaA_client_get_icon_sizes,
                             ^
/home/user/.cache/pacaur/awesome-luajit-git/src/build/objects/client.c:3722:29: warning: cast between incompatible function types from ‘int (*)(lua_State *)’ {aka ‘int (*)(struct lua_State *)’} to ‘int (*)(lua_State *, lua_object_t *)’ {aka ‘int (*)(struct lua_State *, struct <anonymous> *)’} [-Wcast-function-type]
                             (lua_class_propfunc_t) luaA_client_get_first_tag,
                             ^
```

psychon:
Completely untested, but I guess:
```diff
diff --git a/client.c.orig b/client.c
index e8a3641..b40b752 100644
--- a/client.c.orig
+++ b/client.c
@@ -2544,10 +2544,8 @@ luaA_client_tags(lua_State *L)
 /** Get the first tag of a client.
  */
 static int
-luaA_client_get_first_tag(lua_State *L)
+luaA_client_get_first_tag(lua_State *L, client_t *c)
 {
-    client_t *c = luaA_checkudata(L, 1, &client_class);
-
     foreach(tag, globalconf.tags)
         if(is_client_tagged(c, *tag))
         {
@@ -3436,7 +3434,7 @@ luaA_client_keys(lua_State *L)
 }
 
 static int
-luaA_client_get_icon_sizes(lua_State *L)
+luaA_client_get_icon_sizes(lua_State *L, client_t *c)
 {
     int index = 1;
     client_t *c = luaA_checkudata(L, 1, &client_class);
[...]

PR #2306, opened 2018-07-21
Fix function pointer casts in client.c

psychon:
The getters for properties already get the object as their second
argument, so there is no need to get the object again from Lua.

Fixes: https://github.com/awesomeWM/awesome/issues/2299

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -2549,10 +2549,8 @@ luaA_client_tags(lua_State *L)
 /** Get the first tag of a client.
  */
 static int
-luaA_client_get_first_tag(lua_State *L)
+luaA_client_get_first_tag(lua_State *L, client_t *c)
 {
-    client_t *c = luaA_checkudata(L, 1, &client_class);
-
     foreach(tag, globalconf.tags)
         if(is_client_tagged(c, *tag))
         {
@@ -3441,10 +3439,9 @@ luaA_client_keys(lua_State *L)
 }
 
 static int
-luaA_client_get_icon_sizes(lua_State *L)
+luaA_client_get_icon_sizes(lua_State *L, client_t *c)
 {
     int index = 1;
-    client_t *c = luaA_checkudata(L, 1, &client_class);
 
     lua_newtable(L);
     foreach (s, c->icons) {


actionless:
Merge fail?

blueyed:
Looks like "Codacy/PR Quality Review" is required and does not report?!

Elv13:
> Would be nice to have the "Details" linked there.

https://github.com/Mergifyio/mergify-engine/issues/115

Look like we will be able to do the (saner) opposite and rebase the branch onto masted instead of merging master into the PR. This will get rid of some wierdness.

Elv13:
I will do Mergifyio/mergify-engine#115 manually for now (git rebase -i origin master). Mergify should not try to merge master into PR, that's fragile compared to rebasing on master. And even if rebasing fails, at least you know *why*.

alfunx:
@Elv13 It looks like you rebased master on psychon/fix-2299... ([git log](https://i.imgur.com/o0eVpxw.png))

Elv13:
> @Elv13 It looks like you rebased master on psychon/fix-2299... (git log)

? I didn't touch `master`? I put the 4 commits on this PR above master HEAD, which was the idea. Mergify was trying to do something funny by merging master into the PR, and kept failing due to that. All I did was to undo that and rebased so it can be merged again

Unless I missed something, that git-log screenshot isn't master, but all rather pull request commits, right? I git-cloned a full copy of awesomeWM/awesome and none of the commits in you screenshots are in master. Plus, AFAIK, we have branch protection on master to prevent accidental force-push

Elv13:
Ah nvm, I see what you mean, I misunderstood the comment as me rebasing the real master branch. Anyway, that would have merged correctly anyway, the important part was removing the part that messed mergify

blueyed:
It's unfortunate that now also the merge-master-branch commit (that mergify created in this PR) is being merged into master.

jd:
@blueyed @Elv13 This is only the case because you enabled strict mode. If you disable it, you won't have those merge (but you'll also be open to a small window of merging patch that only *appears* to work).

PR #2307, opened 2018-07-22
Keyboard and keygrabbing handling overhaul

Elv13:
At last...

So since yesterday I took a break from my hiking vacations to finish this pull request. I previously wanted submit it after the popup and conditional ones, but I am too lazy to fix them, so here is even more code!

Joke aside, this is now quite mature compared to the other 2, so while I am sure @psychon will do his usual magic and find ways to make it better, it's not that bad already.

First of all, this squash all the shims commit I have from all the branches into one because rebasing them is now madness. That commit can be safely ignored since currently nothing uses most of that. In the event something (like any of my other PR) starts using them, then any bugs can be fixed there, Yes it will cause the covarage to go down, I don't think it's an issue here. Plus, it makes my other PR easier to merge since they will be smaller.

The second thing I want to note is that adding and removing root keybindings at runtime sucks and this pull request does not fix it. At some point it will require a better system but for now the hack is enough. It's ugly, yes. It's also fragile. However fixing that is more of an Awesome5 project than v4.3.

Finally, @actionless suggestion to cram in auto_repeat handling, pass-through events and other goodness are not part of the PR. Making it work wasn't trivial and this is already quite big even by my standards.

As usual, most of the PR lines of code is tests and documentation. The actual "new code" isn't that big.

Finally, I added some "doc examples" CMakeLists.txt things I had on my stash to fix the regression where newlines were no longer working since luacheck said it was a hack (18 months ago).

In theory I think I didn't break the API. Please prove me wrong if it is the case. One very minor change is that `capi.keygraber.run` and `awful.keygrabber.run` are now the same function. It fixes a bug where `awful.prompt` would stall if another keygrabber started when it was in input mode.

### DIFF
--- a/lib/awful/keygrabber.lua
+++ b/lib/awful/keygrabber.lua
@@ -1,22 +1,162 @@
 ---------------------------------------------------------------------------
---- Keygrabber Stack
+--- A keyboard grabbing and transaction object.
+--
+-- This module allows to grab all keyboard inputs until stopped. It is used
+-- to gather input data (such as in `awful.prompt`), implement VI-like keybindings
+-- or multi key transactions such as emulating the Alt+Tab behavior.
+--
+-- Note that this module has been redesigned in Awesome 4.3 to be object oriented
+-- and stateful. The use of the older global keygrabbing API is discouraged
+-- going forward since it had problem with recursive keygrabbers and required
+-- a lot of boiler plate code to get anything done.
+--
+-- Using keygrabber as transactions
+-- --------------------------------
+--
+-- The transactional keybindings are keybindings that start with a normal
+-- keybindings, but only end when a (mod)key is released. In the classic
+-- "Alt+Tab" transaction, first pressing "Alt+Tab" will display a popup listing
+-- all windows (clients). This popup will only disappear when "Alt" is released.
+--
+-- @DOC_text_awful_keygrabber_alttab_EXAMPLE@
+--
+-- In that example, because `export_keybindings` is set to `true`, pressing
+-- `alt+tab` or `alt+shift+tab` will start the transaction even if the
+-- keygrabber is not (yet) running.
+--
+-- Using keygrabber for modal keybindings (VI like)
+-- ------------------------------------------------
+--
+-- VI-like modal keybindings are trigerred by a key, like `Escape`, followed by
+-- either a number, an adjective (or noun) and closed by a verb. For example
+-- `<Escape>+2+t+f` could mean "focus (f) the second (2) tag (t)".
+-- `<Escape>+2+h+t+f` would "focus (f) two (2) tags (t) to the right (h)".
+--
+-- Here is a basic implementation of such a system. Note that the action
+-- functions themselves are not implemented to keep the example size and
+-- complexity to a minimum. The implementation is just if/elseif of all action
+-- and the code can be found in the normal `rc.lua` keybindings section:
+--
+-- @DOC_text_awful_keygrabber_vimode_EXAMPLE@
+--
+-- Using signals
+-- -------------
+--
+-- When the keygrabber is running, it will emit signals on each event. The
+-- format is "key_name".."::".."pressed_or_released". For example, to attach
+-- a callback to `Escape` being pressed, do:
+--
+--    mykeygrabber:connect_signal("Escape::pressed", function(self, modifiers, modset)
+--        print("Escape called!")
+--    end)
+--
+-- The `self` argument is the keygrabber instance. The `modifiers` is a list of
+-- all the currently pressed modifiers and the `modset` is the same table, but
+-- with the modifiers as keys instead of value. It allow to save time by
+-- checking `if modset.Mod4 then ... end` instead of looping into the `modifiers`
+-- table or using `gears.table`.
 --
 -- @author dodo
+-- @author Emmanuel Lepage Vallee &lt;elv1313@gmail.com&gt;
 -- @copyright 2012 dodo
--- @module awful.keygrabber
+-- @copyright 2017 Emmanuel Lepage Vallee
+-- @classmod awful.keygrabber
 ---------------------------------------------------------------------------
 
 local ipairs = ipairs
 local table = table
-local capi = {
-    keygrabber = keygrabber }
+local gdebug = require('gears.debug')
+local akey = require("awful.key")
+local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
+local gtable = require("gears.table")
+local gobject = require("gears.object")
+local gtimer = require("gears.timer")
+local glib = require("lgi").GLib
+local capi = { keygrabber = keygrabber, root = root }
 
-local keygrabber = {}
+local keygrab = {}
 
 -- Private data
 local grabbers = {}
 local keygrabbing = false
 
+local keygrabber = {
+    object = {}
+}
+
+-- Instead of checking for every modifiers, check the key directly.
+--FIXME This is slightly broken but still good enough for `mask_modkeys`
+local conversion = {
+    Super_L   = "Mod4",
+    Control_L = "Control",
+    Shift_L   = "Shift",
+    Alt_L     = "Mod1",
+    Super_R   = "Mod4",
+    Control_R = "Control",
+    Shift_R   = "Shift",
+    Alt_R     = "Mod1",
+}
+
+--BEGIN one day create a proper API to add and remove keybindings at runtime.
+-- Doing it this way is horrible.
+
+-- This list of keybindings to add in the next event loop cycle.
+local delay_list = {}
+
+local function add_root_keybindings(self, list)
+    assert(
+        list, "`add_root_keybindings` needs to be called with a list of keybindings"
+    )
+
+    local was_started = #delay_list > 0
+
+    -- When multiple `awful.keygrabber` objects are created in `rc.lua`, avoid
+    -- unpacking and repacking all keys for each instance and instead merge
+    -- everything into one operation. In not so extreme cases, not doing so
+    -- would slow down `awesome.restart()` by a small, but noticeable amount
+    -- of time.
+    gtable.merge(delay_list, list)
+
+    -- As of Awesome v4.3, `root.keys()` is an all or nothing API and there
+    -- isn't a standard mechanism to add and remove keybindings at runtime
+    -- without replacing the full list. Given `rc.lua` sets this list, not
+    -- using a delayed call would cause all `awful.keygrabber` created above
+    -- `root.keys(globalkeys)` to be silently overwritten. --FIXME v5
+    if not was_started then
+        gtimer.delayed_call(function()
+            local ret = {}
+
+            for _, v in ipairs(delay_list) do
+                local mods, key, press, release, description = unpack(v)
+
+                if press then
+                    local old_press = press
+                    press = function(...)
+                        self:start()
+                        old_press(...)
+                    end
+                end
+
+                if release then
+                    local old_release = release
+                    release = function(...)
+                        self:start()
+                        old_release(...)
+                    end
+                end
+
+                table.insert(ret, akey(mods, key, press, release, description))
+            end
+
+            -- Wow...
+            capi.root.keys(gtable.join( capi.root.keys(), unpack(ret) ))
+
+            delay_list = {}
+        end)
+    end
+end
+
+--END hack
 
 local function grabber(mod, key, event)
     for _, keygrabber_function in ipairs(grabbers) do
@@ -27,17 +167,119 @@ local function grabber(mod, key, event)
     end
 end
 
+local function runner(self, modifiers, key, event)
+    -- Stop the keygrabber with the `stop_key`
+    if key == self.stop_key
+        and event == self.release_event and self.stop_key then
+        self:stop(key, modifiers)
+        return false
+    end
+
+    -- Stop when only a subset of keys are allowed and it isn't one of them.
+    if self._private.allowed_keys and not self._private.allowed_keys[key] then
+        self:stop(key, modifiers)
+        return false
+    end
+
+    -- Support multiple stop keys
+    if type(self.stop_key) == "table" and event == self.release_event then
+        for _, k in ipairs(self.stop_key) do
+            if k == key then
+                self:stop(k, modifiers)
+                return false
+            end
+        end
+    end
+
+    local is_modifier = conversion[key] ~= nil
+
+    -- Reset the inactivity timer on each events.
+    if self._private.timer and self._private.timer.started then
+        self._private.timer:again()
+    end
+
+    -- Lua strings are bytes, to handle UTF-8, use GLib
+    local seq_len = glib.utf8_strlen(self.sequence, -1)
+
+    -- Record the key sequence
+    if key == "BackSpace" and seq_len > 0 then
+        self.sequence = glib.utf8_substring(self.sequence, 0, seq_len - 2)
+    elseif glib.utf8_strlen(key, -1) == 1 and  event == "release" then
+        self.sequence = self.sequence..key
+    end
+
+    -- Convert index array to hash table
+    local mod = {}
+    for _, v in ipairs(modifiers) do mod[v] = true end
+
+    -- Emit some signals so the user can connect to a single type of event.
+    self:emit_signal(key.."::"..event, modifiers, mod)
+
+    local filtered_modifiers = {}
+
+    -- User defined cases
+    if self._private.keybindings[key] and event == "press" then
+        -- Remove caps and num lock
+        for _, m in ipairs(modifiers) do
+            if not gtable.hasitem(akey.ignore_modifiers, m) then
+                table.insert(filtered_modifiers, m)
+            end
+        end
+
+        for _,v in ipairs(self._private.keybindings[key]) do
+            if #filtered_modifiers == #v[1] then
+                local match = true
+                for _,v2 in ipairs(v[1]) do
+                    match = match and mod[v2]
+                end
+                if match then
+                    v[3](self)
+
+                    if self.mask_event_callback ~= false then
+                        return
+                    end
+                end
+            end
+        end
+    end
+
+    -- Do not call the callbacks on modkeys
+    if is_modifier and self.mask_modkeys then
+        return
+    end
+
+    if self.keypressed_callback and event == "press" then
+        self.keypressed_callback(self, modifiers, key, event)
+    elseif self.keyreleased_callback and event == "release" then
+        self.keyreleased_callback(self, modifiers, key, event)
+    end
+end
+
 --- Stop grabbing the keyboard for the provided callback.
+--
 -- When no callback is given, the last grabber gets removed (last one added to
 -- the stack).
--- @param g The key grabber that must be removed.
-function keygrabber.stop(g)
+--
+-- @param[opt] g The key grabber that must be removed.
+-- @deprecated awful.keygrabber.stop
+function keygrab.stop(g)
+    -- If `run` is used directly and stop() is called with `g==nil`, get the
+    -- first keygrabber.
+    g = g
+        or keygrab.current_instance and keygrab.current_instance.grabber
+        or grabbers[#grabbers]
+
     for i, v in ipairs(grabbers) do
         if v == g then
             table.remove(grabbers, i)
             break
         end
     end
+
+    if g and keygrab.current_instance and keygrab.current_instance.grabber == g then
+        keygrab.current_instance = nil
+    end
+
     -- Stop the global key grabber if the last grabber disappears from stack.
     if #grabbers == 0 then
         keygrabbing = false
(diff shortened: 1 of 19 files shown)


Elv13:
* Added a commit to mock capi.keygrabber for busted tests
* Fixed a `ldoc` error I didn't get on my desktop
* Removed a few leftover TODO

tests/examples/shims/awesome.lua
```diff
@@ -43,9 +43,26 @@ awesome._shim_fake_class = _shim_fake_class
...
 end
 
+function awesome.xkb_get_group_names()
+    return "pc+us+inet(evdev)"
+end
+
+function awesome.xkb_get_layout_group()
+    return 0
+end
+
```
psychon: Uhm... Really? ;-)

tests/examples/shims/client.lua
```diff
@@ -31,6 +31,7 @@ function client.gen_fake(args)
     ret.size_hints = {}
     ret.border_width = 1
     ret.icon_sizes = {{16,16}}
+    ret.name = "client"
```
psychon: How about `Fake Client` or `Example Client` or something like that? `client` might be a bit too generic.

tests/examples/shims/keygrabber.lua
```diff
@@ -1,11 +1,60 @@
...
+
+local function stop()
+    keygrabber._current_grabber = nil
+end
+
+keygrabber = {
+    _run             = run,
+    stop             = stop,
+    isrunning        = function() return keygrabber._current_grabber ~= nil end,
+    _current_grabber = nil
```
psychon: Why the `rawset` and why the mix between inline functions and separate functions? Couldn't `_current_grabber` simply be saved in a local?

tests/examples/shims/keygrabber.lua
```diff
@@ -1,11 +1,60 @@
...
+    rawset(keygrabber, "_ni_handler", h)
+end
+
+function keygrabber.set_index_miss_handler(h)
+    rawset(keygrabber, "_i_handler", h)
+end
+
+function keygrabber.set_call_handler(h)
+    rawset(keygrabber, "_c_handler", h)
+end
```
psychon: What's that? I don't think the C API has this function.

tests/examples/shims/keygrabber.lua
```diff
@@ -1,11 +1,60 @@
...
+        local h = rawget(keygrabber,"_i_handler")
+        if h then
+            return h(self, key)
+        end
+    end,
+    __newindex = function(...)
+        local h = rawget(keygrabber,"_ni_handler")
+        if h then
+            h(...)
+        end
```
psychon: Why once with explicit arguments and the other time with `...`?

tests/examples/shims/mouse.lua
```diff
@@ -17,11 +17,35 @@ function mouse.coords(args)
...
     table.insert(mouse.old_histories, mouse.history)
     mouse.history = {}
 end
 
-return mouse
+return setmetatable(mouse, {
+    __index = function(self, key)
+        if key == "screen" then
+            return screen[1]
+        end
```
psychon: ...why? There already is a `screen` property being set above, so this should be dead code (`__index` is not called if the requested member exists).
In case I am wrong:
...why not remove the (apparently non-working) `screen = screen[1]` above?

tests/examples/shims/root.lua
```diff
@@ -1,9 +1,126 @@
...
+
+function root.cursor() end
+
+-- GLOBAL KEYBINDINGS --
+
+local keys = {}
+
+-- Note: the tests are not expected to create 1000 of them, so to keep the
+-- implementation minimal, it doesn't try to hash them and simply loop over
+-- every entry.
```
psychon: Uhm, the C code does not hash them either. It just does a linear search for matching keys. So, I do not see the point of this comment and would just remove it.

tests/examples/shims/root.lua
```diff
@@ -1,9 +1,126 @@
...
+local function match_modifiers(mods1, mods2)
+    if #mods1 ~= #mods2 then return false end
+
+    for _, mod1 in ipairs(mods1) do
+        local found = false
+        for _, mod2 in ipairs(mods2) do
+            if mod1 == mod2 then
+                found = true
+                break
+            end
```
psychon: `gears.table.hasitem`?

tests/examples/shims/root.lua
```diff
@@ -1,9 +1,126 @@
...
+        end
+        if not found then return false end
+    end
+
+    return true
```
psychon: Note that awesome also supports the modifier `Any` which is not handled correctly in the code above. Might be worth mentioning this in a comment?
In case you are wondering where the C code does this matching: `event_key_match` in `event.c`.

tests/examples/shims/root.lua
```diff
@@ -1,9 +1,126 @@
...
+end
+
+local function execute_press_keybindings(key, event)
+    -- It *could* be extracted from gears.object private API, but it's equally
+    -- ugly as using the list used by the hotkey widget.
+    if not hotkeys then
+        hotkeys = require("awful.key").hotkeys
+    end
```
psychon: Huh?!? Shouldn't this be using `keys`, which is the list saved by `root.keys()`?

tests/examples/shims/root.lua
```diff
@@ -1,9 +1,126 @@
...
+
+    for _, v in ipairs(hotkeys) do
+        if key == v.key and match_modifiers(v.mod, get_mods()) and v[event] then
```
psychon: I don't understand this line. Okay, it does not try to handle all the keycode vs keysym magic, but... We are checking
- for same key
- matching modifiers for some weird definition of "matching"
- has a callback

And then we only run the first matching function.

Why just the first match that is found?
Why are the key's modifiers matched against a list of all modifiers that exist / are known?

Edit: ARGH. I added a comment to the definition of `mods`.

tests/examples/shims/root.lua
```diff
@@ -1,9 +1,126 @@
...
+local conversion, mods = {
+    Super_L   = "Mod4",
+    Control_L = "Control",
+    Shift_L   = "Shift",
+    Alt_L     = "Mod1",
+    Super_R   = "Mod4",
+    Control_R = "Control",
+    Shift_R   = "Shift",
+    Alt_R     = "Mod1",
+}, {}
```
psychon: Could you pretty please split this up? I searched for `mods`, found this, and did not notice the syntactic weirdness that is going on here and thus was confused. Just do `local mods = {}`. That makes this a lot more readable and only costs you one more line of code.

tests/examples/shims/root.lua
```diff
@@ -1,9 +1,126 @@
...
+        table.insert(ret, mod)
+    end
+
+    return ret
+end
+
+local function remove_modkey(key)
+    if not conversion[key] then return end
+    mods[conversion[key]] = nil
+end
```
psychon: I suggest moving `add_modkey` and `remove_modkey` next to each other. They belong together.

Also, technically these functions are wrong. Which modifiers are active when I
- press `Alt_L`
- press `Alt_R`
- relese `Alt_L`.

Your code says that no modifiers are active, my intuition tells me that `Mod1` should be active.

tests/examples/shims/root.lua
```diff
@@ -1,9 +1,126 @@
...
+        else
+            execute_press_keybindings(key, "press")
+        end
+    end,
+    key_release    = function(key)
+        remove_modkey(key)
+        if keygrabber._current_grabber then
+            keygrabber._current_grabber(get_mods(), key, "release")
+        else
+            execute_press_keybindings(key, "release")
```
psychon: What exactly does the `press` in the name of the function refer to? How about just `execute_keybindings`?

tests/examples/shims/root.lua
```diff
@@ -1,9 +1,126 @@
...
+            return
+        end
+    end
+end
+
+local fake_input_handlers = {
+    key_press      = function(key)
+        add_modkey(key)
+        if keygrabber._current_grabber then
+            keygrabber._current_grabber(get_mods(), key, "press")
```
psychon: Depending on how compatible to the C API you want to be:
Keygrabbers should be run in a protected context and are automatically stopped when they throw a Lua error.

tests/examples/shims/root.lua
```diff
@@ -1,9 +1,126 @@
...
+            execute_press_keybindings(key, "release")
+        end
+    end,
+    button_press   = function() --[[TODO]] end,
+    button_release = function() --[[TODO]] end,
+    motion_notify  = function() --[[TODO]] end,
+}
+
+function root.fake_inputs(event_type, detail, x, y)
+    assert(fake_input_handlers[event_type], "Unknown event_type")
```
psychon: For C API compatibility: Instead of an `assert`, this should just return.

tests/examples/shims/screen.lua
```diff
@@ -24,8 +24,18 @@ local function create_screen(args)
         geo.height = args2.height or geo.height
     end
 
+    s.outputs = {{
+        mm_width  = 0,
+        mm_height = 0,
+    }}
```
psychon: The key of the table should be the name of the output, e.g. `LVDS1`, `VGA-1` or things like that. Dunno if it really matters here, but with that argument it would also be okay just not to add this at all...?

tests/examples/shims/screen.lua
```diff
@@ -24,8 +24,18 @@ local function create_screen(args)
...
+
     local wa = args.workarea_sides or 10
 
+    -- This will happen if `clear()` is called
+    if mouse and not mouse.screen then
+        mouse.screen = s
```
psychon: Uhm... this does not work. Your `__newindex` in `mouse` will catch this instead.
So, how come you are fixing this "`mouse.screen` can refer to an already-deleted screen" problem three times and each time is half-broken?

You wrote yourself that this commit is the result of several (old?) commits that were through many rounds of rebasing and were just now squashed. Perhaps it makes sense to rewrite this commit in sensible, small parts guided by what is actually needed by some of the doctests?

tests/examples/shims/screen.lua
```diff
@@ -82,15 +93,20 @@ function screen._clear()
         screen[i] = nil
     end
     screens = {}
+
+    if mouse then
+        mouse.screen = nil
+    end
```
psychon: This will also lead to fun with `__newindex` when trying to set a new screen.

tests/examples/shims/keygrabber.lua
Elv13: It is added by one of the other commit

tests/examples/shims/root.lua
Elv13: Technically I have used `gears` elsewhere in the shims so it is a bit hypocritical to argue about it, but technically `gears` depends on `capi.awesome`, so in that case, I made that on purpose to avoid future circular conflicts. However it could be used here as it stands now.

Then again, it ain't a lot of code, so I don't think bikeshedding over a simple loop versus extra dependencies is well spent time.

tests/examples/shims/root.lua
Elv13: Not currently, at least not cleanly. The callbacks are stored in the signal connections, so getting them is potentially going to run into "for some reason there is now many signals on "press" and I don't know which to use.

In any case, this is part of the reason I said in the header description that the way Awesome handle keys has to change as it is not extremely inflexible and race-prone. It doesn't matter as long as the user never change `root.keys()`, but as soon as you do, things get messy, fast.

Using `hotkeys` here is a workaround to avoid having to add even more code to fix the "real" issue. At some point, `capi.key` and `awul.key` should be merged into an unified object with a clean API. However it can't be done without breaking the API because an `awful.key` can have multiple `capi.key`.

Which bring us to the limitation of the current design and why `root.keys()` can't be used. @actionless `awful.hotkey_popup` ran into the same issue and added a parallel way to store the key properties. The shims now makes use of it instead of duplicating it. Also note that another commit add the `press` and `release` property to the previously mentioned parallel way to store key properties.

tests/examples/shims/root.lua
Elv13: I know, but the goal of the shims is to be the most minimal implementation that pass the currently existing tests. It was never intended to behave correctly and handle corner cases that are not reached using the documentation tests.

Adding all these corner cases (and also the keycode vs. keysym equivalent of this bug) would make the code harder to read for no practical reasons.

tests/examples/shims/root.lua
Elv13: If we mute failures, that would defy the purpose of having tests. `pcall()` makes no sense here.

tests/examples/shims/root.lua
Elv13: > If we mute failures, that would defy the purpose of having tests. pcall() makes no sense here.

This was done of purpose.

tests/examples/shims/screen.lua
Elv13: It is currently not used by `awful` or any of my tests, so it doesn't need to be in the shims (yet).

tests/examples/shims/screen.lua
Elv13: > Perhaps it makes sense to rewrite this commit in sensible, small parts guided by what is actually needed by some of the doctests?

That was the original idea, however I gave up because I have a "main branch" where all my PRs (submitted or yet to be submitted) are to be able to dogfood everything at once. However it got to the point I could no longer " `rebase -i`+remove the old merge commit and merge the PR again" without breaking everything.

This is in part because I have too many branches and too many ideas half implemented, but it is also because I am an active AwesomeWM developer and it will inevitably happen. Not all ideas are good ideas and not all ideas are worth spending time to get the past your review process ;).

So in the end I squashed this and now try to get it merged all at once so I can stop wasting time rebasing conflicts every single time I update one of the PR.

It ain't perfect and I prefer smaller commits, but it is no longer manageable with a day job, hobbies, outdoor activities and life itself. Better go though the review now and get over it. Now that the "upstream" shims are more complete, then I can do smaller changes to them again with decreased risk of rebase issues.

tests/examples/shims/mouse.lua
Elv13: Rebase issue...

The `if key == "screen" then ... end` is the right one. The other one breaks when you `fake_remove` the screen.

tests/examples/shims/root.lua
Elv13: It's out of scope for now

tests/examples/shims/keygrabber.lua
Elv13: Nope, `root` need to have a pointer to send the fake_input to the grabber

Elv13:
Rebased on the new shims commit and without the `awful.keyboard` usage in the tests (I moved the code to the shims with `_` in front to denote private tests API). I will rebase it once again when the shims will be merged.

lib/awful/keygrabber.lua
```diff
@@ -1,22 +1,96 @@
...
+-- overwrite them if the keygrabber is created directly from `rc.lua`.
+local delay_list = {}
+
+local function add_root_keybindings(self, list)
+    if not list then return end
+
+    local was_started = #delay_list > 0
+
+    -- Changing all keybindings is very slow and freezes Awesome for a visible
+    -- amount of timer. Make sure to batch as much as possible.
```
psychon: Benchmark:
```diff
diff --git a/tests/test-benchmark.lua b/tests/test-benchmark.lua
index def3c01f..a6f5cbff 100644
--- a/tests/test-benchmark.lua
+++ b/tests/test-benchmark.lua
@@ -71,11 +71,30 @@ local function e2e_tag_switch()
     do_pending_repaint()
 end
 
-benchmark(create_and_draw_wibox, "create&draw wibox")
-benchmark(update_textclock, "update textclock")
-benchmark(relayout_textclock, "relayout textclock")
-benchmark(redraw_textclock, "redraw textclock")
+local keys
+local function setup_keys(n)
+    keys = {}
+    local gears = require("gears")
+    for i=1, n do
+        keys = gears.table.join(keys, awful.key({"mod4"}, "x", function() end, { description = "foo", group="bar" }))
+    end
+end
+
+local function change_keybindings()
+    root.keys(keys)
+end
+
+--benchmark(create_and_draw_wibox, "create&draw wibox")
+--benchmark(update_textclock, "update textclock")
+--benchmark(relayout_textclock, "relayout textclock")
+--benchmark(redraw_textclock, "redraw textclock")
 benchmark(e2e_tag_switch, "tag switch")
+setup_keys(10)
+benchmark(change_keybindings, "change keys 10")
+setup_keys(100)
+benchmark(change_keybindings, "change keys 100")
+setup_keys(1000)
+benchmark(change_keybindings, "change keys 1000")
 
 runner.run_steps({ function() return true end })
 
```
Result:
```
          tag switch: 0.00571348 sec/iter (180 iters, 1.354 sec for benchmark)
      change keys 10: 0.00265768 sec/iter (380 iters, 1.013 sec for benchmark)
     change keys 100: 0.0261727  sec/iter ( 39 iters, 2.046 sec for benchmark)
    change keys 1000: 0.261728   sec/iter (  4 iters, 1.31 sec for benchmark)
```
That's less than 0.3 milliseconds per key and this seems to scale linearly with the number of keys. Sorry, but what are you doing that this turns into visible freezes?!?

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+-- * `callback` A function that will be called when the key combination is
+--  pressed.
+-- * `description` A table various metadata to be used for `awful.hotkeys_popup`.
+--
+-- @property keybindings
+-- @param table
+-- @see export_keybindings
+-- @see root_keybindings
+
+--- Define a list of keys that are allowed instead of using `release_key`.
```
psychon: I fail to understand what this means (and I only noticed that because the code implementing this confused me).

Does this mean that when any key that is not allowed is pressed, the keygrabber is automatically stopped? I think that should be spelt out more explicitly.
`If any key is pressed that is not in this list, the keygrabber is stopped`

lib/awful/keygrabber.lua
```diff
@@ -27,11 +101,99 @@ local function grabber(mod, key, event)
...
+    -- Stop when only a subset of keys are allowed and it isn't one of them.
+    if self._private.allowed_keys and not self._private.allowed_keys[key] then
+        self:stop(key, modifiers)
+        return false
+    end
+
+    -- Support multiple release keys
+    if type(self.release_key) == "table" then
+        for _, k in ipairs(self.release_key) do
+            if (k == key or conversion[key] == k) and event == self.release_event then
```
psychon: I'd move the check for `release_event` out of the `for` loop up the `type(self.release_key) == "table"` since it's loop-invariant.

lib/awful/keygrabber.lua
```diff
@@ -27,11 +101,99 @@ local function grabber(mod, key, event)
...
+    -- Reset the inactivity timer on each events.
+    if self._private.timer and self._private.timer.started then
+        self._private.timer:again()
+    end
+
+    -- Record the key sequence
+    if key == "BackSpace" and self.sequence:len() > 0 then
+        self.sequence = self.sequence:sub(
+            1, self.sequence:len() - 1
+        )
```
psychon: This breaks with UTF8. I vaguely remember that we have ugly code in `awful.prompt` that handles this via some magic, but I cannot find it right now... :-/

So, no idea what to do about this.

Edit: I just tested this, and, well, I need to press backspace twice in a prompt to get rid of an `ä` again.

lib/awful/keygrabber.lua
```diff
@@ -27,11 +101,99 @@ local function grabber(mod, key, event)
...
+    elseif key:len() == 1 and  event == "release" then
```
psychon: I guess this also breaks with UTF8. I can enter `ä` into a prompt and
```
$ lua -e 'print(("ä"):len())'
2
```

lib/awful/keygrabber.lua
```diff
@@ -27,11 +101,99 @@ local function grabber(mod, key, event)
...
+        self.sequence = self.sequence..key
+    end
+
+    -- Convert index array to hash table
+    local mod = {}
+    for _, v in ipairs(modifiers) do mod[v] = true end
+
+    -- Emit some signals so the user can connect to a single type of event.
+    self:emit_signal(key.."::"..event.."ed", modifiers, mod)
```
psychon: This signal format is at least uncommon in awesome. Buttons have `button::press` and keys have `press`.
So:
- why past tense?
- how about `event .. "::" .. key` instead?

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+-- @DOC_text_awful_keygrabber_timeout_EXAMPLE@
+--
+-- @property timeout
+-- @see gears.timer
+-- @see timeout_callback
+
+--- The key on which the keygrabber listen to terminate itself.
+--
+-- When this is set, the running keygrabber when quit when [one of] the release
+-- key event occurs.
```
psychon: s/when/will/ ...and something else that I fail to figure out right now. It's too hot.

lib/awful/keygrabber.lua
```diff
@@ -1,22 +1,96 @@
...
+local conversion = {
+    Super_L   = "Mod4",
+    Control_L = "Control",
+    Shift_L   = "Shift",
+    Alt_L     = "Mod1",
+    Super_R   = "Mod4",
+    Control_R = "Control",
+    Shift_R   = "Shift",
+    Alt_R     = "Mod1",
+}
```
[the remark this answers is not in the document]
psychon: > -- Please note that modkeys are accepted as release keys as long as the
> -- keyboard layouts use the IBM PC 104-key naming convention for the key names.

Sigh... so you just document this and call it a day?
Okay, but any complaints about this get assigned to you.

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+-- @see mask_event_callback
+
+--- Export all keygrabber keybindings as root (global) keybindings.
+--
+-- When this is enabled, calling all of the keygrabber object `keybinding`s will
+-- will create root `awful.key` and will automatically starts the grabbing.
+--
+-- Use this with caution. If many keygrabber or "real" root keybindings are set
+-- on the same key combination, they are all executed and there is almost no
+-- safe way to undo that. Also, while `is_enabled` can be used to temporarely
```
psychon: According to Ctrl+F, there is no `is_enabled`.

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+-- turn off this behavior, it currently cannot be detached completely. Make sure
+-- the `keygrabber` that use this option have a single instance.
+--
+-- @property export_keybindings
+-- @param[opt=false] boolean
```
psychon: If it's so unsafe, then why support it at all? :-(

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+
+--- The root (global) keybinding to start this keygrabber.
+--
+-- Instead of adding an entry to `root.keys` or `rc.lua` `globalkeys` section,
+-- this property can be used to take care of everything. This way, it becomes
+-- easier to package modules using keygrabbers.
```
psychon: Hm... so that's your use case? Hm...
Hm...

Hm...
Doesn't this basically mean that you encourage packages with poor customisability? Users might not ever see the keybindings that are added. The packages that I saw so far expire functions to call in the right places, I think, which makes them more complicated to set up, true, but also makes it clearer what is going on.
If I remember correctly, I saw instructions like "add this to your global keybindings definitions". Is that too hard for users?

Hm... :-/
I'm not sure what to think about this.

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+--
+-- When defined, this property allows to define an implicit way to release the
+-- keygrabber. It helps save some boilerplate code in the handler callbacks.
+--
+-- @DOC_text_awful_keygrabber_allowed_keys_EXAMPLE@
+--
+-- @property allowed_keys
+-- @tparam[opt=nil] table|nil allowed_keys The list of keys.
+
+--- The sequence recorded from the current acquisition.
```
psychon: "current acquisition" is a bit... opaque? How about "the sequence recorded since the start of the keygrabber"?

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+--
+-- @signal property::current_instance
+
+--- If a keygrabber is currently running.
+-- @tfield boolean is_running
+
+--- Start the keygrabber.
+--
+-- Note that only a single keygrabber can be started at any one time. If another
+-- keygrabber (or this one) is currently running. This method returns false.
```
psychon: I thought one of the goals was to make it possible to stack keygrabbers and have multiple currently running ones?!?
(But supporting that might not actually be a good idea...)

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+        __newindex = function(tab, key, value)
+            if keygrabber["set_"..key] then
+                self[key](self, value)
+            else
+                rawset(tab, key, value)
+            end
+        end
+    })
+
+    self.sequence = ""
```
psychon: If the goal of `.current_instance` is not to pollute the original keygrabber, why is this polluting the original keygrabber? :-/
In fact, I cannot find anything that actually uses `.current_instance`.

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+    -- Ease making keygrabber that wont hang forever if no action is taken.
+    if self.timeout and not self._private.timer then
+        self._private.timer = gtimer {
+            timeout     = self.timeout,
+            single_shot = true,
+            callback    = function()
+                if self.timeout_callback then
+                    self.timeout_callback(self)
+                end
+                self:stop()
```
psychon: This means the keygrabber won't be stopped if `timeout_callback` causes an error.

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+            end
+        }
+    end
+
+    if self._private.timer then
+        self._private.timer:start()
+        assert(self._private.timer.data.source_id)
```
psychon: Why this `assert`? Are you testing that `:start()` works correctly?!?
Why not use `.is_running` instead?

(Slightly unrelated: Should `.data` be renamed to `._private`?)

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+        )
+    end
+
+    self.current_instance = nil
+
+    keygrab.current_instance = nil
+    keygrab.emit_signal("property::current_instance", nil)
+
+    self.grabber = nil
+    self:emit_signal("stopped")
```
psychon: What's the intended difference between `start_callback` vs `started` signal and `stop_callback` vs `stopped` signal?

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+end
+
+--- Add a keybindings to this keygrabber.
```
psychon: `s/keybindings/keybindin/`

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+
+--- When the keygrabber stops.
+-- @signal stopped
+
+--- A function called when a keygrabber starts.
+-- @callback start_callback
+-- @tparam keygrabber keygrabber The keygrabber.
+
+--- The callback function to always call without arguments, regardless of
+-- whether the prompt was cancelled.
```
psychon: Huh? Copy&paste from `awful.prompt`, or am I missing something?
Also, you document this function as `call[ed] without arguments` and then describe the arguments?

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+-- @tparam table self The current transaction object.
+-- @tparam string release_key The release key (if any)
+-- @tparam table release_mods The release modifiers key (if any)
+-- @tparam sting sequence The recorded key sequence.
+-- @callback stop_callback
+-- @see sequence
+-- @see stop
+
+--- The function called when the keygrabber stops because of a `timeout`.
+-- @callback timeout_callback
```
psychon: Should this say explicitly that `stop` is still emitted and the `stop_callback` is still called?

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+        if #v >= 3 and #v <= 4 then
+            local _,key,callback = unpack(v)
+            if type(callback) == "function" then
+                ret._private.keybindings[key] = ret._private.keybindings[key] or {}
+                ret._private.keybindings[key][#ret._private.keybindings[key]+1] = v
+            else
+                gdebug.print_warning("The hook's 3rd parameter has to be a function.")
+            end
+        else
+            gdebug.print_warning("The hook has to have 4 parameters.")
```
psychon: This warning does not really help finding the problem. How about `gears.debug.dump(v)` afterwards and changing the message accordingly?
Also, why "has to have 4 parameters", but above 3 is fine as well?

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+
+    -- Build the hook map
+    for _,v in ipairs(args.keybindings or {}) do
+        if #v >= 3 and #v <= 4 then
+            local _,key,callback = unpack(v)
+            if type(callback) == "function" then
+                ret._private.keybindings[key] = ret._private.keybindings[key] or {}
+                ret._private.keybindings[key][#ret._private.keybindings[key]+1] = v
+            else
+                gdebug.print_warning("The hook's 3rd parameter has to be a function.")
```
psychon: Same as below.

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+    ret._private.keybindings = {}
+    ret.release_event = args.release_event or "press"
+
+    -- Build the hook map
+    for _,v in ipairs(args.keybindings or {}) do
+        if #v >= 3 and #v <= 4 then
+            local _,key,callback = unpack(v)
+            if type(callback) == "function" then
+                ret._private.keybindings[key] = ret._private.keybindings[key] or {}
+                ret._private.keybindings[key][#ret._private.keybindings[key]+1] = v
```
psychon: (This is a matter of taste, I guess, but) why not `table.insert`?

lib/awful/keygrabber.lua
```diff
@@ -45,7 +207,411 @@ function keygrabber.stop(g)
...
+        end
+    end
+
+    if args.export_keybindings then
+        add_root_keybindings(ret, args.keybindings)
+    end
+
+    local mt = getmetatable(ret)
+
+    -- So it can be used in other keybindings directly
```
psychon: Huh? What does "directly" mean and why can other keybindings not do `:start()`?
And where is it documented that calling the object directly is the same as calling `:start()`? (I'm not saying that this is not documented, I just can't recall seeing that)

lib/awful/keygrabber.lua
```diff
@@ -79,19 +661,80 @@ end
...
     if not keygrabbing then
         keygrabbing = true
         capi.keygrabber.run(grabber)
     end
+
     return g
 end
 
-return keygrabber
+-- Implement the signal system for the keygrabber.
```
psychon: Why? We already have such an implementation. It is in `gears.object`.

lib/awful/keygrabber.lua
```diff
@@ -79,19 +661,80 @@ end
...
+    for _, cb in ipairs(signals[name]) do
+        cb(...)
+    end
+end
+
+function keygrab.get_is_running()
+    return keygrab.current_instance ~= nil
+end
+
+--@DOC_object_COMMON@
```
psychon: Correct me if I'm wrong, but this line brings in another round of documentation for the signal APIs, plus it brings in `weak_connect_signal`, which you did not actually implement for your reimplementation.

tests/examples/text/awful/keygrabber/autostart.lua
```diff
@@ -0,0 +1,26 @@
+
+local awful = { keygrabber = require("awful.keygrabber") } --DOC_HIDE
+
+local naughty = { notify = function() end } --DOC_HIDE
+
+local autostart_works = false --DOC_HIDE
+
+awful.keygrabber {
+    autostart      = true,
+    release_key    = "Return",
```
psychon: Hm... perhaps `release_key` should be renamed to `stop_key`? I feel like that could make the function more intuitive.

lib/awful/keygrabber.lua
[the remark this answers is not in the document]
actionless: actually i've just realized what even on my layout Alt_L ~= Alt_R

psychon:
I bet that the XKEYBOARD extension allow retrieving this somehow. However, no idea about the details of "somehow".
A quick look at https://www.x.org/docs/XKB/XKBproto.pdf and https://xkbcommon.org/doc/current/ (specifically https://xkbcommon.org/doc/current/group__components.html) did not find anything obviously useful.

Elv13:
Thanks for the review. Most issues seem straight-forward to address. I am semi-AFK for a few more days then I will take care of the issues. I will remove the modkey <-> release key conversion for now until a proper implementation is added. As @actionless points out, Alt_Gr != Alt_L and that needs a proper solution.

As for bringing `keygrabber` and `key` closer and having the keygrabber export `key`s, yes, I do think it is a net improvement in usability and worth the trouble. Back when Awesome was at its peak popularity and my `Tyrannical` module was commonly used, I got these issues time and time again. Users just failed to setup the keybindings to the point I wrote the code found in this PR to auto-add them to `root.keys()`. Either they added them to the client keybinding section for copy-pasted the provided shortcut examples at the end of `rc.lua` (which is an invalid syntax). If keybindings where more object-like and adding/removing them from the `root.keys()` was implicitly done under the hood, then this kind of issue would happen less. I don't blame the users for making this mistake. It just happen our current API is confusing to them. This PR starts the journey to fix this. As said elsewhere, fixing this properly is an Awesome v5 project and would break the API. The current workaround is, in my opinion, good enough from an user point of view. Yes, there is a race condition, it is documented and can be "exploited", but I don't think the users will realistically run into it. Plus, it's a nice fit because it would otherwise introduce yet another incompatible keybinding format (`root`, `client`, `prompt` and `awful.keygrabber`). With this PR, the relation between root and awful.keygrabber keybindings is very clear and smooth.

psychon:
Random, semi-related thought: Make `root.keys("add", a_table_containing_keys)` work and make it add the given keys. The same could be done with `"delete"`. This API is ugly and shouldn't be documented (better ideas?), but might be less ugly than the `delayed_call`-approach. Only "big" downside is that this hack would no longer be self-contained, but would bleed into the C API.

(How to generalise this API and make it less ugly? Keys can be set on root, on clients and on drawins in the C API, if I remember correctly. A solution working for all three would be nice... A property on the `key` might be nice, but would be quite non-trivial to implement and would "reverse the natural dependencies" from the C POV.)

I'm not saying that this is a good idea and I still don't really like that this means keys are added in random places, but I'm also trying to be constructive here.

For the "`root.keys` is too slow"-comment: As I showed, this scales linearly. However, could it be that at some point you only added individual keys? I.e., a loop of querying all keys, adding one entry, setting the keys. That would lead to quadratic behaviour.

Elv13:
> Random, semi-related thought: Make root.keys("add", a_table_containing_keys) work and make it add the given keys.

Unfortunately, it wont be enough. The problem isn't the `O(N)` operation performance. The little performance issue is in the same way as https://github.com/awesomeWM/awesome/issues/1494 and the issue you mentioned where each keygrabber object would add their keys individually. But it isn't a problem with the "bundling" of all keygrabber keybindings in a single `root.keys()` call. It could even be called a premature optimization.

That's not why there is a `delayed_call`. That delay is because  `rc.lua` set the keybindings with `root.keys(globalkeys)` and it overwrite the keygrabbers ones if they were set before that line. The "performance issue" isn't really the main problem here. The code just makes sure it doesn't call `root.keys(values)` multiple time for no reasons (which otherwise does freeze Awesome for a few milliseconds and makes reloading slower).

> How to generalise this API and make it less ugly? [...] A property on the key might be nice, but would be quite non-trivial to implement and would "reverse the natural dependencies" from the C POV.)

That's mostly my long, long term plan. The general idea is to make `awful.key` an "higher level object" that handle these things internally. After that `root.keys()` would be deprecated because of the issues highlighted above and the `delayed_call` hack removed. The general concept of a group of keybindings would still exist because a) it's what the keygrabber does and b) it still need to integrate with `awful.rules`. So that imaginary API would be `awful.key` with a `group` (or something along this line?) property.

But again, that's not for tomorrow. When the waycooler backend becomes mature enough, then we can open the API breaking season again and release Awesome v5. That's the time to fix this issue. Right now doing so without breaking the API would be a heap of hacks (such as this `delayed_call`!). So my plan is to use the ugly "solution" in this PR for now, then fix `awful.keygrabber` and `awful.hotkey_popup` (and my Collision, Repetitive and Tyrannical modules) once we can break the API.

I still have many long term projects going on for 4.x, I wont start working on v5 until I am done with them (I posted that roadmap in multiple places, wont repeat it here). This PR is part of the "fix the notifications" series and is later used to create a keyboard based way to trigger actions and dismiss notifications. It is parallel to the `popup` one and with it are the last 2 (of 8) groundwork PRs before the `naughty.notification`/`naughty.widget.*` objects PR is proposed (you already partially reviewed that one last year and I implemented your requests. It wont be proposed until after v4.3 is released).

tests/examples/text/root/fake_click.lua
```diff
@@ -0,0 +1,9 @@
+--DOC_NO_USAGE --DOC_ASTERISK
+
+  local function click(button_id, x, y)
+      mouse.coords {x = x, y = y}
+      root.fake_inputs("button_press" , button_id)
+      root.fake_inputs("button_release", button_id)
+  end
+--DOC_NEWLINE
+  click(1, 42, 42)
```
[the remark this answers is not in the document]
Elv13: There is no image, it's just code and in one case, text.

tests/examples/text/awful/keygrabber/autostart.lua
Elv13: done

tests/examples/text/root/fake_string.lua
```diff
@@ -0,0 +1,23 @@
...
+awful.keygrabber{autostart=true, release_key = "!", --DOC_HIDE
+    stop_callback = function(_,_,_, seq) works=seq=="Hello world" end} --DOC_HIDE
+
+  local function send_string_to_client(s, c)
+      local old_c = client.focus
+      client.focus = c
+      for i=1, #s do
+          local char = s:sub(i,i)
+          root.fake_inputs("key_press"  , char)
+          root.fake_inputs("key_release", char)
```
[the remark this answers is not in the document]
Elv13: done

root.c
```diff
@@ -211,10 +211,52 @@ _string_to_key_code(const char *s)
...
+/** Send fake keyboard or mouse events.
  *
- * @param event_type The event type: key_press, key_release, button_press,
- *  button_release or motion_notify.
+ * Usually the currently focused client or the keybindings will receive those
+ * events. If a `keygrabber` or `mousegrabber` is running, then it will get them.
+ *
+ * Some keys have different names compared to the ones generally used in
+ * Awesome. For example, here is the modifiers Awesome names versus the ones
+ * that need to be used by `fake_inputs`:
```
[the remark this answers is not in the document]
Elv13: done

root.c
```diff
@@ -211,10 +211,52 @@ _string_to_key_code(const char *s)
...
+ *
+ * <table>
+ *  <tr style='font-weight: bold;'>
+ *   <th align='center'>Modifier name </th>
+ *   <th align='center'>Key name</th>
+ *   <th align='center'>Alternate key name</th>
```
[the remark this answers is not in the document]
Elv13: done

Elv13:
```
CMake Error at tests/examples/CMakeLists.txt:241 (file):
  file failed to open for reading (No such file or directory):

    /home/lepagee/dev/awesome/tests/examples/text/awful/keygrabber/root_keybindings.output.txt
  tests/examples/CMakeLists.txt:305 (run_test)
  awesomeConfig.cmake:285 (include)
  CMakeLists.txt:33 (include)


CMake Error at tests/examples/CMakeLists.txt:243 (message):
  Wrong output from
  /home/lepagee/dev/awesome/tests/examples/text/awful/keygrabber/root_keybindings.lua:


  Expected:

  

  Got:

  Is now active!	nil

  A key was pressed:	a	with	0	modifier!

  A key was pressed:	i	with	1	modifier!

  Called again!

  tests/examples/CMakeLists.txt:305 (run_test)
  awesomeConfig.cmake:285 (include)
  CMakeLists.txt:33 (include)

```

@psychon ^^ Look like your asserts are a little trigger happy when adding new tests

psychon:
> @psychon ^^ Look like your asserts are a little trigger happy when adding new tests

At least the "no such file or directory" gives a good hint on what is going wrong. How else should this fail? (Note that #2333, once merged, will make the failure a bit less annoying since that one fixes `STRICT_TESTS`, i.e. makes sure that failing tests are ignored by default).

Elv13:
Rebased on master to give the proper error message in the CI instead if `no such file or directory`. I also added the `output.txt`. I didn't notice it was now mandatory to add them to the repository.

A bit offtopic, we can't do it for `svg` because the output changes depending on the Cairo version. We will need to pull them from the `awesome-www` if we want to compare them.

If I didn't mess things up with the rebase, this should be up to review again.

psychon:
I didn't do that change to catch wrong output, but to be able to move running the tests from the configure phase to the build phase. So I'm fine with this not checking the images for changes.

> If I didn't mess things up with the rebase, this should be up to review again.

Okay, it's on my list. Especially since I forgot to remove the tag after I last took a look at it.

psychon:
Oh and thanks for the two new commits in here (there are at least two new ones, possibly more...)

Elv13:
>  If you are pushing for something and I just say "no", then that leaves everyone unhappy.

I know you are not a fan of the global keybindings in many place change, but I still think in the long run it allows better encapsulation and is generally necessary. Imagine if for each WebExtension Firefox/Chrome/IE has you would need to modify `userChrome.css`/`firefox.js`. However I agree with you the current `root.keys()` solution is not elegant and a proper redesign of this aspect would go a long way to find a better middle ground about how we handle the various keybindings in clients/root/prompt/keygrabber. I see your concerns about this aspect as hitting a limitation of our current design rather than something conceptually broken. Both Qt/KDE QAction, QtQuick.Action and GAction/GtkAction have this abstraction to allow a single action to be enabled/disabled and used in various contexts.

Thanks for accepting this PR anyway.

@blueyed  Do you have some time to fix the doc typos and improve the wording on the final version? It changed a bit since you last went over it.

Elv13:
I rebased the PR and added the commit to remove `capi.keygrabber` from the public API as discussed somewhere. It should prevent people using `capi.keygrabber.run` by accident.

Also fixes an existing bug discovered by porting `hotkey_popup` to `awful.keygrabber` instead of `capi.keygrabber`. The bug itself was already present in `awful.keygrabber` and isn't a regression from this PR. `awful.keygrabber.stop()` was documented to be able to take no argument, but it didn't work as documented. This caused the `hotkey_popup` to eat all inputs, forever.

actionless:
@Elv13  have you opened an issue to fix mask_modkeys?

Elv13:
> @Elv13 have you opened an issue to fix mask_modkeys?

No, I waited until it is merged. It makes little sense to report an bug about unmerged code ;)

PR #2311, opened 2018-07-24
Example tests: Make it explicit what kind of output is expected

psychon:
This modifies the tests in `tests/examples` so that we already know before running them what kind of output is expected. I expect this to be necessary for hopefully some times moving running these tests from the configure phase to the test phase (but I might be wrong?).


blueyed:
Thank you!

PR #2310, opened 2018-07-24
.travis.yml: Enable -Werror

psychon:
Signed-off-by: Uli Schlachter <psychon@znc.in>

----

I expect the Travis build to fail since we currently have a warning. Once #2309 is fixed, I expect that build failure to go away. Thus, this PR will even be tested!

Issue #2309, opened 2018-07-24
Compiler warning on current master


psychon:
Commit 7e17bcd026a39403f8da81172826faca75df2efb introduced the following warning:
```
[  3%] Building C object CMakeFiles/awesome.dir/objects/client.c.o
/home/psychon/awesome/build/objects/client.c:938:5: Warnung: kein vorheriger Prototyp für »client_set_startup_id« [-Wmissing-prototypes]
     client_set_##prop(lua_State *L, int cidx, char *value) \
     ^
/home/psychon/awesome/build/objects/client.c:956:1: Anmerkung: bei Substitution des Makros »DO_CLIENT_SET_STRING_PROPERTY2«
 DO_CLIENT_SET_STRING_PROPERTY2(startup_id, startup_id)
 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I tried to get this error in English, but setting `LC_ALL=C` did not help, so... "no previous prototype for `client_set_startup_id`" is what the message say.

(Now that I looked at the commit again: Why `DO_CLIENT_SET_STRING_PROPERTY2(startup_id, startup_id)` instead of `DO_CLIENT_SET_STRING_PROPERTY(startup_id)`?)

(Should we build with `-Werror` in Travis?)

blueyed:
> Should we build with -Werror in Travis?

Yes, for one of the build jobs I'd say.

PR #2312, opened 2018-07-24
Another attempt at making -Werror work on Travis

psychon:
Follow up for #2310 (apparently I cannot reopen a PR after a force-push).


psychon:
This now fails in the intended way:
```
[ 79%] Building C object CMakeFiles/awesome.dir/objects/client.c.o
/home/travis/build/awesomeWM/awesome/build/objects/client.c:938:5: error: no previous prototype for ‘client_set_startup_id’ [-Werror=missing-prototypes]
     client_set_##prop(lua_State *L, int cidx, char *value) \
     ^
/home/travis/build/awesomeWM/awesome/build/objects/client.c:956:1: note: in expansion of macro ‘DO_CLIENT_SET_STRING_PROPERTY2’
 DO_CLIENT_SET_STRING_PROPERTY2(startup_id, startup_id)
 ^
cc1: all warnings being treated as errors
```

actionless:
but we can't merge it anyway until the warnings causes themselves will fixed, right?

psychon:
Right. It's just nice to know that this works as intended. :-)

The idea is to rebase this on master when #2309 is fixed and hope that the build then just works.

actionless:
now it's fine but squashing is needed for PREV_COMMITS

psychon:
Rebased on master. The only change is to fix the typo that I always do and that @blueyed found.

@actionless The order of the commits is already fine for `TEST_PREV_COMMITS` (I think). GitHub just displays the commits in a different order, because I used a rebase to re-order them. The earlier commit now has a latter commit date, and I guess GitHub orders by commit date and not by parent relationship.

blueyed:
> mergify/pr — Merge fail

actionless:
should we postpone the merging because of https://github.com/awesomeWM/awesome/pull/2320#issuecomment-408193408 ?

actionless:
@psychon i was asking if current config will produce any warnings on gcc 8

just checked now on gcc 8.1 -- no warnings seen

Elv13:
> @psychon i was asking if current config will produce any warnings on gcc 8

GCC has 4 level of warnings:

 * Default (really bad stuff)
 * Wall (still critical stuff like non-void not retuning)
 * Wextra (less important warnings)
 * All the other (not always relevant or prone to false positive)

And then there is the pedantic mode, but that's something else.

The default and Wall sets of warning rarely changes because it annoys distributions like OpenSUSE that have Werror default for most of their packages. If GCC would add too many warnings by default in each release, the effort required from these distro rebuild the world would not be manageable.

It is why #2320 exists. With GCC, past Wextra, you have to curate your warnings suits. Clang has a better system where it has -Weverything, then you must curate the ones you wish to ignore. I like it better since new versions require no "RTFM" to get the new good ones. And it's also explicit that you didn't forget one because it's opt-out instead of opt-in.

PR #2315, opened 2018-07-24
awful: Add a keyboard module.

Elv13:
Extracted from the keygrabber PR

It will eventually have more goodness, but for now is just a wrapper
on top of `fake_inputs`.

It has tests, but they depend on the other PR. It can be merged as-is if approved, otherwise I will add the other 2 commits in the second row of review.

### DIFF
--- a/lib/awful/keyboard.lua
+++ b/lib/awful/keyboard.lua
@@ -1,0 +1,77 @@
+---------------------------------------------------------------------------
+--- Utilities related to the keyboard and text entry.
+--
+-- @author Emmanuel Lepage Vallee &lt;elv1313@gmail.com&gt;
+-- @copyright 2018 Emmanuel Lepage Vallee
+-- @module awful.keyboard
+---------------------------------------------------------------------------
+local capi = {root = root, client = client}
+
+local module = {}
+
+-- The hash order isn't consistent when the content changes, but it is
+-- constant when the table doesn't change.
+local conversion = {
+    Super_L   = "Mod4",
+    Control_L = "Control",
+    Shift_L   = "Shift",
+    Alt_L     = "Mod1",
+    Super_R   = "Mod4",
+    Control_R = "Control",
+    Shift_R   = "Shift",
+    Alt_R     = "Mod1",
+}
+
+--- Send an artificial set of key events to trigger a key combination.
+--
+-- This can be used to call application actions or Awesome own keybindings.
+--
+-- @tparam table modifiers The modifiers (like `Control` or `Mod4`).
+-- @tparam string|number key The key name or keycode.
+-- @function awful.keyboard.execute_keybindings
+function module.execute_keybinding(modifiers, key)
+    for _, mod in ipairs(modifiers) do
+        for real_key, mod_name in pairs(conversion) do
+            if mod == mod_name then
+                capi.root.fake_inputs("key_press", real_key)
+                break
+            end
+        end
+    end
+
+    capi.root.fake_inputs("key_press"  , key)
+    capi.root.fake_inputs("key_release", key)
+
+    for _, mod in ipairs(modifiers) do
+        for real_key, mod_name in pairs(conversion) do
+            if mod == mod_name then
+                capi.root.fake_inputs("key_release", real_key)
+                break
+            end
+        end
+    end
+end
+
+--- Send artificial key events to write a string.
+-- @tparam string string The string.
+-- @tparam[opt=nil] client c A client to send the string to.
+-- @function awful.keyboard.write_string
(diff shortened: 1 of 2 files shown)


lib/awful/keyboard.lua
```diff
@@ -0,0 +1,77 @@
...
+-- @author Emmanuel Lepage Vallee &lt;elv1313@gmail.com&gt;
+-- @copyright 2018 Emmanuel Lepage Vallee
+-- @module awful.keyboard
+---------------------------------------------------------------------------
+local capi = {root = root, client = client}
+
+local module = {}
+
+-- The hash order isn't consistent when the content changes, but it is
+-- constant when the table doesn't change.
```
psychon: I have no idea what this comment talks about. I guess it refers to the iteration order with `pairs`, but I do not see where this is important (and I'd be surprised if the Lua documentation actually gives this guarantee, but whatever...).

lib/awful/keyboard.lua
```diff
@@ -0,0 +1,77 @@
...
+local conversion = {
+    Super_L   = "Mod4",
+    Control_L = "Control",
+    Shift_L   = "Shift",
+    Alt_L     = "Mod1",
+    Super_R   = "Mod4",
+    Control_R = "Control",
+    Shift_R   = "Shift",
+    Alt_R     = "Mod1",
+}
```
psychon: Sorry, but no. This is (or at least: can be) keyboard layout specific. This is not true in general. Please don't make me hunt a keyboard layout where this actually breaks, but if you really want this kind of information, you'll have to figure out how XKB specifies keyboard layouts and how to query this information from the layout. I bet everything you'd need is in libxkbcommon, but I do not really know the details.

lib/awful/keyboard.lua
```diff
@@ -0,0 +1,77 @@
...
+
+--- Send an artificial set of key events to trigger a key combination.
+--
+-- This can be used to call application actions or Awesome own keybindings.
+--
+-- @tparam table modifiers The modifiers (like `Control` or `Mod4`).
+-- @tparam string|number key The key name or keycode.
+-- @function awful.keyboard.execute_keybindings
+function module.execute_keybinding(modifiers, key)
```
psychon: I don't think this function works when some "unwanted modifier" is currently pressed. E.g., I won't be able to bind mod4+Y to press Alt+L, because it will actually do mod4+Alt+L.

lib/awful/keyboard.lua
```diff
@@ -0,0 +1,77 @@
...
+    local old_c = capi.client.focus
+
+    if c then
+        capi.client.focus = c
+    end
+
+    for i=1, #string do
+        local char = string:sub(i,i)
+        capi.root.fake_inputs("key_press"  , char)
+        capi.root.fake_inputs("key_release", char)
```
psychon: Uhm.
- Unicode? (E.g. there are at least two ways to encode `ä` and at most one of these will work, perhaps even none).
- UTF8? (This assumes that every byte is a character).

lib/awful/keyboard.lua
```diff
@@ -0,0 +1,77 @@
...
+
+--- Send artificial key events to write a string.
+-- @tparam string string The string.
+-- @tparam[opt=nil] client c A client to send the string to.
+-- @function awful.keyboard.write_string
+function module.write_string(string, c)
+    local old_c = capi.client.focus
+
+    if c then
+        capi.client.focus = c
```
psychon: This assumes that the client is using the I-don't-know-its-name focus model. I bet this only works for the mostly used focus model out of the four that are specified by ICCCM.

psychon:
Again, sorry, but just no. I do not see how this can ever work correctly without tons and tons of work.

Elv13:
Fair enough, some of these issues were known at the time of posting. But one might want to point out that something that works for 95% of the userbase and provide value should not be dismissed because some UTF+8 string are written like `e and some client steal input when they don't have focus. While those corner case exist, it could be said it's a documentation issue rather than a technical one. Awesome i18n support is non-existant beside half translated manpages and the textclock. As for the `Alt_L` issue, the default `rc.lua` uses keycode and latin chars, so obviously those hypothetical layouts are already broken phonetic asian keyboards. Plus, they are terminal unfriendly so the local users probably don't use them along with a niche Linux tiled WMs anyway. Which bring me back that it's a documentation issue more than anything else.

Anyhow, you don't want it, so be it. I added it because some tests used it. I will move it to the shims to avoid having the boilerplate it removes in each input related documentation examples. I will also better document the `fake_inputs` example to list the limitations mentioned above.

PR #2316, opened 2018-07-24
keygrabber: Turn into an object.

Elv13:
Previously, the capi.keygrabber was global with some kind of incomplete
wrapper called awful.keygrabber that would attempt to turn it into a
self contained object.

The API was also primitive and hard to use. Given the use case for the
keygrabber is almost limited to transactions or sequences, then the API
should focus on making this trivial and idiot proof.

In an ideal world, it should be merged before the shims because the shims implement the changes from this PR. But given nothing uses it yet, its irrelevant.

This commit is a port from the `capi.mouse` commit I did 2 years ago, but for the keygrabber.

### DIFF
--- a/keygrabber.c
+++ b/keygrabber.c
@@ -31,6 +31,75 @@
 
 #include "keygrabber.h"
 #include "globalconf.h"
+#include "common/luaclass.h"
+
+static int miss_index_handler    = LUA_REFNIL;
+static int miss_newindex_handler = LUA_REFNIL;
+static int miss_call_handler     = LUA_REFNIL;
+
+/**
+* Add a custom call handler.
+*/
+static int
+luaA_keygrabber_set_call_handler(lua_State *L)
+{
+    return luaA_registerfct(L, 1, &miss_call_handler);
+}
+
+/**
+* Add a custom property handler (getter).
+*/
+static int
+luaA_keygrabber_set_index_miss_handler(lua_State *L)
+{
+    return luaA_registerfct(L, 1, &miss_index_handler);
+}
+
+/**
+* Add a custom property handler (setter).
+*/
+static int
+luaA_keygrabber_set_newindex_miss_handler(lua_State *L)
+{
+    return luaA_registerfct(L, 1, &miss_newindex_handler);
+}
+
+/** Keygrabber library.
+ * \param L The Lua VM state.
+ * \return The number of elements pushed on stack.
+ * \luastack
+ */
+static int
+luaA_keygrabber_index(lua_State *L)
+{
+    if (miss_index_handler != LUA_REFNIL)
+        return luaA_call_handler(L, miss_index_handler);
+
+    return luaA_default_index(L);
+}
+
+/** Newindex for keygrabber.
+ * \param L The Lua VM state.
+ * \return The number of elements pushed on stack.
+ */
+static int
+luaA_keygrabber_newindex(lua_State *L)
+{
+    /* Call the lua keygrabber property handler */
+    if (miss_newindex_handler != LUA_REFNIL)
+        return luaA_call_handler(L, miss_newindex_handler);
+
+    return luaA_default_newindex(L);
+}
+
+static int
+luaA_keygrabber_new(lua_State *L)
+{
+    if (miss_call_handler != LUA_REFNIL)
+        return luaA_call_handler(L, miss_call_handler);
+
+    return 0;
+}
 
 /** Grab the keyboard.
  * \return True if keyboard was grabbed.
@@ -174,13 +243,23 @@ luaA_keygrabber_isrunning(lua_State *L)
     return 1;
 }
 
-const struct luaL_Reg awesome_keygrabber_lib[] =
+const struct luaL_Reg awesome_keygrabber_methods[] =
 {
     { "run", luaA_keygrabber_run },
     { "stop", luaA_keygrabber_stop },
     { "isrunning", luaA_keygrabber_isrunning },
-    { "__index", luaA_default_index },
-    { "__newindex", luaA_default_newindex },
+    { "__call", luaA_keygrabber_new },
+    { "__index", luaA_keygrabber_index },
+    { "__newindex", luaA_keygrabber_newindex },
+    { "set_index_miss_handler", luaA_keygrabber_set_index_miss_handler},
+    { "set_call_handler", luaA_keygrabber_set_call_handler},
+    { "set_newindex_miss_handler", luaA_keygrabber_set_newindex_miss_handler},
+
+    { NULL, NULL }
+};
+
+const struct luaL_Reg awesome_keygrabber_meta[] =
+{
     { NULL, NULL }
 };
 
--- a/luaa.c
+++ b/luaa.c
@@ -71,7 +71,8 @@
 #ifdef WITH_DBUS
 extern const struct luaL_Reg awesome_dbus_lib[];
 #endif
-extern const struct luaL_Reg awesome_keygrabber_lib[];
+extern const struct luaL_Reg awesome_keygrabber_methods[];
+extern const struct luaL_Reg awesome_keygrabber_meta[];
 extern const struct luaL_Reg awesome_mousegrabber_lib[];
 extern const struct luaL_Reg awesome_root_lib[];
 extern const struct luaL_Reg awesome_mouse_methods[];
@@ -843,9 +844,8 @@ luaA_init(xdgHandle* xdg, string_array_t *searchpath)
     lua_pop(L, 1); /* luaA_registerlib() leaves the table on stack */
 #endif
 
-    /* Export keygrabber lib */
-    luaA_registerlib(L, "keygrabber", awesome_keygrabber_lib);
-    lua_pop(L, 1); /* luaA_registerlib() leaves the table on stack */
+    /* Export keygrabber */
+    luaA_openlib(L, "keygrabber", awesome_keygrabber_methods, awesome_keygrabber_meta);
 
     /* Export mousegrabber lib */
     luaA_registerlib(L, "mousegrabber", awesome_mousegrabber_lib);


keygrabber.c
```diff
@@ -31,6 +31,108 @@
...
+        return 0;
+    }
+    /* Remove error function */
+    lua_remove(L, error_func_pos);
+
+    return lua_gettop(L);
+}
+
+/**
+* Add a custom property handler (getter).
```
blueyed: not a setter?

keygrabber.c
```diff
@@ -31,6 +31,108 @@
...
+luaA_keygrabber_set_call_handler(lua_State *L)
+{
+    return luaA_registerfct(L, 1, &miss_call_handler);
+}
+
+/**
+* Add a custom property handler (getter).
+*/
+static int
+luaA_keygrabber_set_index_miss_handler(lua_State *L)
```
blueyed: not a setter?

keygrabber.c
Elv13: copy paste mistake

keygrabber.c
Elv13: This is a getter metatable handler, lua call the setter `newindex`

actionless: so should it be `luaA_keygrabber_get_index_miss_handler`?

Elv13: no, it *sets* the *getter* metatable handler

This is the C equivalent of `setmetatable({}, __index = function() end})`

keygrabber.c
Elv13: fixed

psychon:
I'm a bit confused by this. What is "an object" for you?
For me, I would expect this code to be moved to `objects/keygrabber.c` and behave similar to the other C objects. Instead... I'm not quite sure. This creates an unused and empty metatable for `keygrabber` objects and basically delegates everything to Lua.

What does this difference do / make possible, that you cannot do by the following:
```lua
local orig_keygrabber = _G.keygrabber
_G.keygrabber = my_own_implementation_that_uses_orig_keygrabber_internally
```
Because to me it looks like you are trying to do the above, but do it through the C code.

Also, I fail to find the difference between `luaA_keygrabber_call_handler`, `luaA_mouse_call_handler`, and `luaA_class_call_handler`. All three of them (two which are already in awesome) look identical to me. Why this code duplication?

Elv13:
> I'm a bit confused by this. What is "an object" for you?

The idea was at first to extend it like `capi.mouse`, then later merge it with `awful.keygrabber` "objects" (and make them public). It could in theory be done as above, however not without some complications. If `keygrabber` was renamed `_keygrabber` and for some reason the user don't end up including `awful.keygrabber` which set `keygrabber`, then that's an API break. If `keygrabber` isn't renamed, then that's `awful` doing some monkey-patching on capi and that should be prohibited in the sake of sanity.

So adding another metatable "handler" function just like we have plenty in other places in the C code seem to me an acceptable compromise given it has little direct consequences.

As for the code duplication, you addressed it in the other PR.

Elv13:
Rebased on master and removed the now shared handler copy/pasted code

psychon:
> If keygrabber isn't renamed, then that's awful doing some monkey-patching on capi and that should be prohibited in the sake of sanity.

I'm not sure about that. All these handlers feel a lot like monkey-patching to me. Almost nothing of the original API stays (or rather: all the original API is deprecated). Or... is there any part of the actual C API that users should still use? If not, then why should users still go through the C API? E.g. LuaJIT is good at optimising Lua code, but it cannot do anything about these kinds of indirections through C

Got some pointers to what your intentions with this are exactly? So far I only saw the code that makes it possible for Lua to "change everything", but not the Lua code that actually uses this.

Elv13:
I see your point and generally agree. However if you read the current `awful.keygrabber` code, you get a better picture of the current state.

Since 2012, using `capi.keygrabber` has been unofficially deprecated. `awful.keygrabber` has been added to supersede `capi.keygrabber` with additional features like supporting multiple grabbers. `awful.prompt` and other code have moved to this superset and using `capi.keygrabber` directly alongside these module is know to cause issues. One such issue is the prompt displaying "Run:" even when it's not running anymore.

So the first part of #2307 is to officially deprecate the direct use of `capi.keygrabber`. But again, it was unofficially deprecated for 6 years, but the documentation didn't reflect it. At the time, there was no way to deprecate stuff anyway, so it's not surprising.

The second part of the change is to apply the rules I started to use in https://github.com/awesomeWM/awesome/pull/790 to `awful.keygrabber`. tl;dr; it does the following:

 * Use the same naming convention everywhere (snake case, get/set/is)
 * When something is (or has) an object, then use methods instead of C-style function with the object as the first parameter for consistency
 * Use the getter and setter to create the property concept
 * Move methods concepts closer to the object they belong to instead of their module concept (aka, `s.tags` instead of `awful.tag.by_screen(s)`

And then it apply the rules I use since https://github.com/awesomeWM/awesome/pull/1124 (also https://github.com/awesomeWM/awesome/pull/2127 and some other). It adds the following convention:

 * Objects constructor should have a single `args` parameter that reflect all object properties

Finally, that pull request adds a default `grabber` function that implements keybindings and some generally useful tricks.

But the final part is out of scope for your question.

> all the original API is deprecated

No, `keygrabber.run` redirect to `awful.keygrabber.run`, but that part of the old API is not deprecated. See https://elv13.github.io/classes/keygrabber.html

So, what is deprecated is that 2 "copy" of the same concept were in the API. The lowest level ones (`capi.keygrabber`) are deprecated. `awful.keygrabber.stop` is renamed to `my_keygrabber_object:stop()` for consistency with the rest of the API, but is otherwise not conceptually deprecated.

> LuaJIT is good at optimising Lua code

Do you have a solution for the "new" keygrabber module (the current `awful.keygrabber`) to be in `_G` even when `awful.keygrabber` is not `require`d? If we can come up with a solution for this, then this PR can be closed as it would be a superior solution to the problem I try to solve.

Elv13:
@psychon Please comment, this is blocking the shim PR which is blocking the popup and keygrabber PRs. If the outcome is "that's too many handler, just rename `capi.keygrabber` and add a new automagical system to force-load some Lua modules in `_G`", then this can be closed. Or we can use this imperfect solution for now and add a lesser solution in v5.0 since minor API breaks would make it easier to do?

Given I (we?) have a little bit more time right now, I guess it's the right time to make a push for them then release v4.3?

psychon:
TL;DR: Why does all of this have to be in `keygrabber` instead of `awful.keygrabber`? We could remove the `keygrabber` global from the API docs or document it as "internal, use awful.keygrabber instead".

> No, keygrabber.run redirect to awful.keygrabber.run,

Uhm, how? The patch here does not make that part possible, does it?

> I guess it's the right time to make a push for them then release v4.3?

I've learnt that "let's push this feature that I am working on and then do a release" is never a good idea. Before I learnt this, XCB gained `xcb_send_fd` which cannot be used in a thread-safe manner. Most things are not that important.

So first, I'd like to suggest to do v4.3 first. I don't mind dropping the gravity stuff from the v4.3 if needed. Most other stuff in that milestone is already on a good track.

> See https://elv13.github.io/classes/keygrabber.html

Okay... so things I see (and that are not really relevant right now): 
- you want to move towards having key bindings defined "everywhere" ("root_keybindings" is magic to add a keybinding to start the keygrabber") instead of a single, central place (root.buttons() in rc.lua).

Let's try to focus on the issue at hand:

Why do you want to "force this" into the `keygrabber` global? How about just documenting the `keygrabber` object as an internal function or removing it from the API docs completely (e.g. `drawin` is not in the docs either (but `drawable` is!?)).

You could then add all your feature to `awful.keygrabber` and people can use that, somehow.

Perhaps it helps to add an optional second argument to `keygrabber.run` in the C API which is a callback function for when the keygrabber is stopped? I'm not quite sure about the various issues that there are, but this should be enough to fix the following:

>  One such issue is the prompt displaying "Run:" even when it's not running anymore.

Elv13:
> TL;DR: Why does all of this have to be in keygrabber instead of awful.keygrabber?

It can be done that way. If you feel it's the way forward, then I can live with that.

> So first, I'd like to suggest to do v4.3 first. I don't mind dropping the gravity stuff from the v4.3 if needed. Most other stuff in that milestone is already on a good track.

Could be done, I don't care about the `wibox.container.conditional` PR not being part of the release. It sucks a bit because it means I wont be able to ship the dynamic client layout as a module for v4.3, but otherwise it's fine. 

The popup would have been nice to merge for a single reason. I didn't push the remaining commits of the "configurable" naughty widgets because I wanted 1 release with the popup first so it was dogfooded a bit before using it for critical components (tooltip and notification popups). Yesterday I rebased it and "fixed" the issue about the tooltip regression. I say "fixed" because I am still uneasy about why only some builds/commits failed and not other. They should have all failed. The "use the hierarchy" comment, which is the last important one to handle is still problematic. I mean, I made it, but it doesn't seem to be better than the delayed call for now. It can totally be made to work and it is indeed a better solution than coupling the wibox size with the drawing code. May I propose to skip this for now and open an issue milestones for v4.4 to make it work properly?

It depends on the shims PR which I will update later today without the `keygrabber` object part. 

After that the `awful.keygrabber` PR will be easy to review. It will now be below 300 lines of code (+400 of doc and +300 of tests). All it really does is adding a default grabber callback with some generally useful properties. It also attempt to converge the keybindings `root.keys()` and `awful.keygrabber` as a more coherent package. Just like the client keybindings are mostly inter-operable with the global ones. The idea being to have a coherent API between transactional keybindings, global keybindings and client keybindings.

Also, I have the changelog ready, it just need to remove the parts that are not part of the release.

https://github.com/Elv13/awesome-1/blob/doc_tests_and_notif/docs/89-NEWS.md

Elv13:
> Well, we do have drawin, wibox, awful.wibar which are all also "basically the same". ;-)

`drawin` is hidden, so it doesn't count. `wibox` and `wibar` are very different thing from a user point of view. We know that they are in fact the same thing. But if you ask a random user and ask them if a tooltip and their desktop shell is the same thing, they will say no ;)

psychon:
> It can be done that way. If you feel it's the way forward, then I can live with that.

Seems less weird to me than trying to monkey-patch `keygrabber` without monkey-patching this. Also, that would be more consistent with other things, since (most) of our Lua code is accessed that way while the C code is accessed "directly".

Would that be okay with you? Or do you think making this all work through `keygrabber` directly would be much better?

About your wish list of things to have in the release: I'll leave this to you. I do not have that much time at my hands (but more than in the last few months), but just beware that this goes against "release early, release often" ;-)

Elv13:
> Or do you think making this all work through keygrabber directly would be much better?

It solves the "bug" where things built on `awful.keygrabber` break when mixed with `capi.keygrabber` from the user config or 3rd party modules. The idea behind merging them was to get rid of those issues. However I wont go to war for that. If you wish to reduce the scope and "just" improve `awful.keygrabber` and remove `capi.keygrabber` from the doc, it at least prevent the problem from getting worst. At this point it's more about compromise to start merging/releasing things again instead of having them pile up in my branch and not having a release in 13 months.

psychon:
> It solves the "bug" where things built on awful.keygrabber break when mixed with capi.keygrabber from the user config or 3rd party modules.

So, how exactly does this solve that bug? As I wrote above:

>> No, keygrabber.run redirect to awful.keygrabber.run,
>
> Uhm, how? The patch here does not make that part possible, does it?

To explain more: This patch adds things like `__index` meta-methods, but as far as I see, these methods are only used if no entry is found. Thus, if something calls `keygrabber.run` directly, it will still, well, call that function directly. Also, `awful.keygrabber` still needs some C function to call to actually start a grabber, so actually replacing that function is not possible.

Put differently: As far as I can see, code doing `keygrabber.run()` will still behave just the way it did before these commits.

Elv13:
The other PR renames keygrabber.run to keygrabber._run and do `capi.keygrabber.run = awful.keygrabber.run`. But that solution has issues like what happen if `awful.keygrabber` is never `require()`d. It also needs to handle `stop` to really solve the issue. However I kind of expected the current backlash so I didn't invest too much time into "really" doing everything right due to the high probability of you not liking it. Some of the comments you made were expected, I still tried to have what I see as a better solution, but really, lets not bikeshed about this, in the end it's a minor implementation detail where both potential paths to a solution have issues.

PR #2319, opened 2018-07-24
Mergify: do not use strict mode

blueyed:
It has the drawback of having intermediate merge commits for when
Mergify merges master before because of this, and then does not
squash-merge or rebases it when merging.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -2,7 +2,6 @@ rules:
   default:
     protection:
       required_status_checks:
-        strict: True
         contexts:
           - continuous-integration/travis-ci
           - Codacy/PR Quality Review


Elv13:
Note that https://github.com/Mergifyio/mergify-engine/issues/115 is about fixing this issue upstream so we can have the strict mode, no extra merge commit **and** enforce a branch to always be on top of master/HEAD, witch make `tig` and `git log --graph` output a **lot** saner (ei. no multi level branches and no overlap/crosspoints)

actionless:
will the new commits discard the review state or not?

blueyed:
@actionless 
You mean new commits pushed to a branch?
I would assume so - AFAIK it is not related to strict mode.
Strict mode is more about ensuring that it is up-to-date with master.

Ah, but mergify could require that Travis was run with the state of latest master (since Travis tests the merge commit already) - Github then says "This branch is out-of-date with the base branch".
I do not think Mergify can trigger a re-build (for Travis in our case), can it?
Because this would then cause Travis to test another (updated) merge commit then.

actionless:
so it's fine by me if it's so

PR #2322, opened 2018-07-25
Some more steps towards getting the examples tests out of the configure phase

psychon:
The expected output of a test is now in a .txt file next to the actual Lua code of the text. Thus, CMake can expand various `@VAR@`-references without having to run the test. The second commit moves the actual execution of the tests into a shell script, so that less CMake is involved.

In theory it should now be possible to run the example tests during `make` instead of during `cmake`. However, after reading through the documentation for `add_custom_command` and `add_custom_target', I'm just submitting a pull request instead.

Question: How do I sanely reference a shell script that is next to the `CMakeLists.txt` currently being processed? `"${TOP_SOURCE_DIR}/tests/examples/runner.sh"` seems a bit verbose.


tests/examples/CMakeLists.txt
```diff
@@ -188,6 +188,14 @@ function(run_test test_path namespace escaped_content)
         set(DOC_BLOCK_PREFIX "@usage")
     endif()
 
+    # Does the text generate text output?
+    if(tmp_content MATCHES "--DOC_GEN_OUTPUT")
+        # The expected output is next to the .lua file in a .txt file
+        string(REPLACE ".lua" ".txt" expected_output_path ${test_path})
```
blueyed: Should this maybe use a more distinct extension, e.g. appending `.output.txt` (or just `.output`) to the file name?

tests/examples/runner.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/sh
+# Todo: Give this shell script a sane calling convention. Right now it has a
+# file containing the expected output as first argument and just executes
+# everything else.
+
+# Cleanup on errors / aborting
+cleanup() {
+    rm -rf "$file_stderr" "$file_stdout" || true
+}
+trap "cleanup" 0 2 3 15
```
blueyed: Put this after setting `file_stdout`/`file_stderr` to not call `rm -rf` with something empty.

tests/examples/CMakeLists.txt
psychon: I don't know. I used a `.txt` extension to make it clearer that this is no fancy binary format. I do not really care about the color of the bikeshed here.

What do others say?

tests/examples/runner.sh
psychon: This is the same order as in `tests/run.sh` and calling `rm -rf` with an empty argument is harmless (it just fails). However, with the other way around, the same race will instead mean that the temporary file is not deleted even though it should.
Thus, I think it should stay the way it is.

tests/examples/CMakeLists.txt
blueyed: My point is that it easier / more specific to find `*.output.txt` instead of `*.txt` when looking for generated output.

Elv13: I don't think it's really worth arguing about and can be changed later. `.txt` is fine, `.output.txt` is better. I don't think it should block this PR.

tests/examples/runner.sh
actionless: what about just standard `set -ueo pipefail` in the beginning of the shell script?
specifically `set -u` will guard from commands with unexpectedly blank args

psychon:
Amended with the result of `find tests/examples -name '*.txt' | while read file ; do git mv $file ${file/txt/output.txt} ; done ; mv tests/examples/CMakeLists{.output,}.txt` plus the following patch
```diff
diff --git a/tests/examples/CMakeLists.txt b/tests/examples/CMakeLists.txt
index 18315e41..2f45b84b 100644
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -190,8 +190,8 @@ function(run_test test_path namespace escaped_content)
 
     # Does the text generate text output?
     if(tmp_content MATCHES "--DOC_GEN_OUTPUT")
-        # The expected output is next to the .lua file in a .txt file
-        string(REPLACE ".lua" ".txt" expected_output_path ${test_path})
+        # The expected output is next to the .lua file in a .output.txt file
+        string(REPLACE ".lua" ".output.txt" expected_output_path ${test_path})
     else()
         set(expected_output_path "/dev/null")
     endif()
```
(Actually, the above is a lie, because the rebase broke when I tried to get the fixup into the right commit, but you get the idea)

tests/examples/runner.sh
psychon: I'm not sure if that really is necessary here. Plus, so far this is a shell script and not a bash script, but this `set` assumes bash, I think.

actionless: $ cat test.sh
set -u
A="test A"
echo $A $B

$ bash test.sh
test.sh: line 3: B: unbound variable

$ dash test.sh
test.sh: 3: test.sh: B: parameter not set

tests/examples/CMakeLists.txt
blueyed: Sure, but once it is in there is at least overhead of doing a 2nd PR, and nobody might care then anyway anymore - that's what reviews are for.. ;)

blueyed: Done now, thanks @psychon!

tests/examples/runner.sh
psychon: Okay, so, should I amend the following into the last commit? I skipped pipefail since it is not in POSIX and the script does not use a single pipe (see http://pubs.opengroup.org/onlinepubs/009695399/utilities/set.html)
```diff
diff --git a/tests/examples/runner.sh b/tests/examples/runner.sh
index 8e5e3df5..ff209763 100755
--- a/tests/examples/runner.sh
+++ b/tests/examples/runner.sh
@@ -3,6 +3,8 @@
 # file containing the expected output as first argument and just executes
 # everything else.
 
+set -ue
+
 # Cleanup on errors / aborting
 cleanup() {
     rm -rf "$file_stderr" "$file_stdout" || true
@@ -16,8 +18,10 @@ expected_output=$1
 shift
 
 # Run the command that we were given
+set +e
 "$@" > ${file_stdout} 2> ${file_stderr}
 exit_code=$?
+set -e
 
 # If exit code is not zero or anything was produced on stderr...
 if [ $exit_code -ne 0 -o -s "${file_stderr}" ]
@@ -40,7 +44,7 @@ fi
 if ! cmp --silent "${file_stdout}" "${expected_output}"
 then
     echo "Expected text from ${expected_output}, but got:"
-    diff -u "${expected_output}" "${file_stdout}"
+    diff -u "${expected_output}" "${file_stdout}" || true
     exit 1
 fi
 
```
(Is there a nicer way to make `set -e` not break the actual execution of the target program?)

actionless: change
```bash
 # Run the command that we were given
 set +e
 "$@" > ${file_stdout} 2> ${file_stderr}
 exit_code=$?
 set -e
```

to 

```
 # Run the command that we were given
 "$@" > ${file_stdout} 2> ${file_stderr} || exit_code=$?
```

psychon:
Amended with
```diff
diff --git a/tests/examples/runner.sh b/tests/examples/runner.sh
index 8e5e3df5..2660f6f0 100755
--- a/tests/examples/runner.sh
+++ b/tests/examples/runner.sh
@@ -16,8 +16,8 @@ expected_output=$1
 shift
 
 # Run the command that we were given
-"$@" > ${file_stdout} 2> ${file_stderr}
-exit_code=$?
+exit_code=0
+"$@" > ${file_stdout} 2> ${file_stderr} || exit_code=$?
 
 # If exit code is not zero or anything was produced on stderr...
 if [ $exit_code -ne 0 -o -s "${file_stderr}" ]
```

psychon:
Oh, whoops. And amended again, this time with:
```diff
diff --git a/tests/examples/runner.sh b/tests/examples/runner.sh
index 2660f6f0..d3eb820c 100755
--- a/tests/examples/runner.sh
+++ b/tests/examples/runner.sh
@@ -3,6 +3,8 @@
 # file containing the expected output as first argument and just executes
 # everything else.
 
+set -ue
+
 # Cleanup on errors / aborting
 cleanup() {
     rm -rf "$file_stderr" "$file_stdout" || true
@@ -40,7 +42,7 @@ fi
 if ! cmp --silent "${file_stdout}" "${expected_output}"
 then
     echo "Expected text from ${expected_output}, but got:"
-    diff -u "${expected_output}" "${file_stdout}"
+    diff -u "${expected_output}" "${file_stdout}" || true
     exit 1
 fi
 
```

psychon:
@jd Mergify claims to have merged this one

> mergify/pr — Merged 

But both this PR's status and https://github.com/awesomeWM/awesome/commits/master disagree

jd:
cc @sileht

sileht:
Github is bugged here, all required checks are green but when we click on "merge button" we got a 405 'Required status check "coverage/coveralls" is pending.'

PR #2323, opened 2018-07-25
Fix gravity handling when reparenting a window back to the root window

### LINKED ISSUE #2308: Window gravity not handled properly when a window is unmapped and then mapped again
Awesome seems to handle all window gravity values correctly when the window is mapped the first time. However, if a program which uses `StaticGravity` (or any other gravity except `NorthWestGravity`) unmaps its window and then maps it again, the window moves from the position where it was placed initially.

Tested with:
```
awesome v4.2-345-g94d3918d (Human after all)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

- Modify the default configuration: remove `placement` from window rules, so that Awesome would always honor the client-specified window geometry.
- Compile (`gcc -o map_unmap_test -g -O2 -Wall -W map_unmap_test.c -lX11`) and run the following test program:
```
#include <stddef.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>

#include <X11/Xlib.h>
#include <X11/Xutil.h>

int main(void)
{
        Display *dpy;
        Window win;
        XSetWindowAttributes win_attrs;
        XSizeHints *size_hints;
        int i;

        dpy = XOpenDisplay(getenv("DISPLAY"));
        win = XCreateWindow(dpy, DefaultRootWindow(dpy), 100, 100, 100, 100,
                            0, CopyFromParent, InputOutput, CopyFromParent,
                            0, &win_attrs);

        size_hints = XAllocSizeHints();
        size_hints->flags = USPosition | USSize | PWinGravity;
        size_hints->win_gravity = StaticGravity;
        XSetWMNormalHints(dpy, win, size_hints);
        XFree(size_hints);

        for (i = 0; i < 5; ++i) {
                XMapWindow(dpy, win);
                XSync(dpy, True);
                sleep(1);

                XUnmapWindow(dpy, win);
                XSync(dpy, True);
                sleep(1);
        }

        return 0;
}
```

**Expected result:**

A 100×100 window should appear, the top left corner of the client area should be placed at (100, 100); the window should flash 5 times, but remain at the same position.

**Actual result:**

The window initially appears at (100, 100), but after every unmap and map the window position shifts up (by border_width+titlebar.height) and to the left (by border_width).

The real world case when this problem appears is https://github.com/telegramdesktop/tdesktop/issues/4837 (that report is for a different window manager, but the problem is apparenty the same — `telegram-desktop` sets `StaticGravity` for its window, and the “minimize to tray” action unmaps the window, then restoring just maps that window again).

I have found a Lua-level workaround for this problem — remove titlebars and border in the "unmanage" callback:
```
client.connect_signal("unmanage", function (c)
    awful.titlebar.hide(c, "top")
    awful.titlebar.hide(c, "left")
    awful.titlebar.hide(c, "right")
    awful.titlebar.hide(c, "bottom")
    c.border_width = 0
end)
```

psychon:
Fixes #2308 
CC @sigprof (I was faster than "coming weekend", whoooh; the most complicated part of this was that I broke `test-gravity` at first, but did not notice the error (this now has to wait for a specific `PropertyNotify` to know that unmapping is done, but it also listened for other `PropertyNotify` messages; the result was that that the state machine got really, really confused)).


psychon:
Meh. :-(
This works when running just `test-gravity.lua`, but Travis points out that this fails when run together with the other tests. What the heck!?!
```
Read on stderr: ERROR: For window with gravity SouthWest in state MAPPED, expected y = 43+-1, but got 43
Read on stderr: ERROR: For window with gravity South in state MAPPED, expected y = 43+-1, but got 43
Read on stderr: ERROR: For window with gravity SouthEast in state MAPPED, expected y = 43+-1, but got 43
```

psychon:
Uhm... okay. For some reason, the titlebars have a different size when just running the test vs when running all tests (22px for "make check", 16px for just the one test). I don't know why, nor do I know if this is a red herring or the actual problem (since the problem is a difference of one pixel, I would guess that some rounding in the division is involved, but both these sizes are even...)

sigprof:
Not sure what exactly breaks the test, but the code in `titlebar_resize()` is definitely incorrect for gravities which require centering. The problem is that `titlebar_resize()` passes only the size delta into `xwindow_translate_for_gravity()`, and if that delta is divided by 2 and truncated, the end result may be wrong.

To test this, I added some bindings to `clientkeys`:
```
    awful.key({ modkey,           }, "KP_Add",
        function (c)
            local _, size = c.titlebar_top(c)
            c.titlebar_top(c, size + 1)
        end,
        {description = "increase titlebar height", group = "client"}),
    awful.key({ modkey,           }, "KP_Subtract",
        function (c)
            local _, size = c.titlebar_top(c)
            if size > 0 then
                c.titlebar_top(c, size - 1)
            end
        end,
        {description = "decrease titlebar height", group = "client"}),
```
Then if I run https://github.com/ccg/metacity/blob/master/src/wm-tester/test-gravity.c and use Mod4+KP_Add and Mod4+KP_Subtract on windows with `WestGravity`, `CenterGravity` and `EastGravity`, I see that those windows do not remain vertically centered when the titlebar height is changed. Note that clicking with button 1 inside those windows will make `test-gravity` reposition them properly.

Changing the titlebar size in steps of 2 instead of 1 makes the problem disappear, because in this case truncation to 0 does not happen. Changing the border width also does not cause problems, because the total width occupied by borders on opposite sizes is always even.

The proper way to calculate the coordinates is to call `xwindow_translate_for_gravity()` twice — first with the old titlebar sizes, then with the new ones — and then use the difference of those results.

However, all of this should not matter if the titlebar size is changed from 0 to the final value in one step, which is what should usually happen — so the problem with that test is most likely caused by something else.

sigprof:
If I force the titlebar size to an even value (e.g., 22px) in `test-gravity.lua`:
```
diff --git a/tests/test-gravity.lua b/tests/test-gravity.lua
index 087a5abc..e406a41a 100644
--- a/tests/test-gravity.lua
+++ b/tests/test-gravity.lua
@@ -6,6 +6,10 @@ local spawn = require("awful.spawn")
 local todo = 2
 local had_error = false
 
+client.connect_signal("request::titlebars", function(c)
+    c.titlebar_top(c, 22)
+end)
+
 local function wait_a_bit(count)
     if todo == 0 or count == 5 then
         return true
```
I see the following test breakage:
```
Read on stderr: ERROR: For window with gravity West in state CREATED, expected y = 70+11, but got 82
Read on stderr: ERROR: For window with gravity West in state RESIZED1, expected y = 70+23, but got 94
Read on stderr: ERROR: For window with gravity West in state MAPPED, expected y = 43+11, but got 55
Read on stderr: ERROR: For window with gravity Center in state CREATED, expected y = 70+11, but got 82
Read on stderr: ERROR: For window with gravity Center in state RESIZED1, expected y = 70+23, but got 94
Read on stderr: ERROR: For window with gravity Center in state MAPPED, expected y = 43+11, but got 55
Read on stderr: ERROR: For window with gravity East in state CREATED, expected y = 70+11, but got 82
Read on stderr: ERROR: For window with gravity East in state RESIZED1, expected y = 70+23, but got 94
Read on stderr: ERROR: For window with gravity East in state MAPPED, expected y = 43+11, but got 55
```
However, if the titlebar size is odd, the test passes.

psychon:
Rebased on master. Still fails on Travis, but no longer fails locally. No idea what to do about this...

actionless:
mb add scrot call and upload it to fileshare, so you could see what happens in xvfb inside travis:

https://github.com/themix-project/oomox-gtk-theme/blob/master/test/test.sh#L138-L141

actionless:
are you able to reproduce the problem inside docker image with the same ubuntu version as on travis? (i guess, 12.04)

psychon:
I don't know and I am not going to install docker in the hopes that this might be specific to the ubuntu version. All the code that decides the positions of windows is in awesome, so the same between what is run on Travis and what is run here. I might stare a bit at `test-gravity.c` and see if I can come up with a possible race that would explain this difference... :-/

psychon:
Rebased on master for #2355

PR #2329, opened 2018-07-28
Fix 3rd party toolbar unminimization and the default rc.lua focusable issue

Elv13:
Try to clear the v4.3 milestone.

The unminimize commit only add the code proposed in the pull request. It isn't tested and I get it will take a C program (or can LGI do it?) to actually trigger the issue on the CI.

The second commit fix the `c.focusable = false` issue mentionned in the bug report. This way, nothing is done when well, nothing should be done.

As said in the commit message, the alternative might be to use `request::activate` which already handle this, but it might make `rc.lua` more confusing for new users?

### DIFF
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -81,7 +81,13 @@ end
 function ewmh.activate(c, context, hints) -- luacheck: no unused args
     hints = hints or  {}
 
-    if c.focusable == false and not hints.force then return end
+    if c.focusable == false and not hints.force then
+        if hints.raise then
+            c:raise()
+        end
+
+        return
+    end
 
     local found, ret = false
 
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -98,6 +98,12 @@ function ewmh.activate(c, context, hints) -- luacheck: no unused args
         if found then break end
     end
 
+    -- Minimized clients can be requested to have focus by, for example, 3rd
+    -- party toolbars and they might not try to unminimize it first.
+    if ret ~= false and hints.raise then
+        c.minimized = false
+    end
+
     if ret ~= false and c:isvisible() then
         client.focus = c
     elseif ret == false and not hints.force then
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -15,6 +15,7 @@ local aclient = require("awful.client")
 local aplace = require("awful.placement")
 local asuit = require("awful.layout.suit")
 local beautiful = require("beautiful")
+local alayout = require("awful.layout")
 
 local ewmh = {
     generic_activate_filters    = {},
@@ -400,6 +401,15 @@ function ewmh.client_geometry_requests(c, context, hints)
     end
 end
 
+-- The magnifier layout doesn't work with focus follow mouse.
+ewmh.add_activate_filter(function(c)
+    if alayout.get(c.screen) ~= alayout.suit.magnifier
+      and aclient.focus.filter(c) then
+        return nil
+    else
+        return false
+    end
+end, "mouse_enter")
 
 client.connect_signal("request::activate", ewmh.activate)
 client.connect_signal("request::tag", ewmh.tag)
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -155,16 +155,11 @@ local tasklist_buttons = gears.table.join(
                                               if c == client.focus then
                                                   c.minimized = true
                                               else
-                                                  -- Without this, the following
-                                                  -- :isvisible() makes no sense
-                                                  c.minimized = false
-                                                  if not c:isvisible() and c.first_tag then
-                                                      c.first_tag:view_only()
-                                                  end
-                                                  -- This will also un-minimize
-                                                  -- the client, if needed
-                                                  client.focus = c
-                                                  c:raise()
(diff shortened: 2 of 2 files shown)


psychon:
Okay, so I have something like a project management issue.

The "original idea" was to mark GitHub notification mails as unread to track "I should take another look at this". Now I have more than 50 unread mails in that folder and I never look at them, to be honest.

In the last few weeks, I started to use unread notifications on https://github.com/notifications to mark things that I should look at. Well... That's not sustainable either. It makes checking for new notifications more complicated and doesn't really scale either.

So, I just invented a new, more public tracking scheme: A `psychon-should-take-a-look`-label. Since it's a GitHub label, others can set the label, too. Let's see how long it takes before this new scheme fails as well, because I am too lazy to work through the backlog. (So far, I'm a bit afraid for this to be too public and leaving a "paper trail" of "Uli decided to take a look and do nothing" around.)

A time management advice that I learnt in some course is to do things that take less than five minutes immediately, instead of doing fancy time management on them. I expect to review small PRs in less than that, but I also remember spending almost an hour reviewing some larger PRs.

psychon:
@Elv13 Does it make some sense to invent some kind of priorities? As in, you want me to look at [this] first? If so, how?
Also, at which point am I enough of a bottleneck in that you just skip me since I'm only holding you back? :-/

Elv13:
A bit off topic, but:

> @Elv13 Does it make some sense to invent some kind of priorities?

It might be premature to do anything about it. Right now I am in vacation and have a lot of bus/train/station_transfer time so I am more active. It will eventually go back to "uh, I just spent all day in front of a computer doing my job, I want to do something else".

> I'm only holding you back?

Not really. Since quite some time I just assume these long review would happen. In all honesty, if the PR code is "very well polished", "good enough" or "mediocre", the number of change request and work related to them roughly the same at about 70 large changes per 400 LoC (tests/doc/empty excluded). Because of that most recent PRs (2017-2018) are coded following that model. Most of the time is spent on doc/api/tests and less on the original implementation because doing anything else is a waste of time (for me). I rather sent you earlier prototypes and get feedback early than having to undo them and rewrite them from scratch 3 time in a row like some 2016 PRs. Bypassing you would make it confusing about what should be in a very advanced state vs. what should be RFC-ish earlier draft. In a sense it sucks for you because some of what you review is crappy, but it's unavoidable to get larger feature PRs merged. Spending large amount of time getting the original implementation looking good only to have to spent a large amount of time doing large architecture changes only to have another round of lesser polishing changes make merging features prohibitive. However, having a more flexible/simpler original draft implementation, then work it into shape based on the first review has so far produced better results at getting things merged.

My end game has always been mostly transparent, I want to achieve the project mission as stated on the mainpage and provide the ultimate flexible WMs where everything can be configured. I want it to have very good documentation so new users can get on board without having to spend unreasonable amount of time on the learning curve. For this, I slowly advance a more declarative "get things done" API style and bridge the different Awesome API concepts to feel more integrated, but without breaking the API to avoid losing users. After 3 years and 40000 lines of change, I made this a lot closer to being true. All the PRs are just doing round-robin around what I think is the next "worst" part that has to be "ported" to achieve those goals.

Until 2016 I was mostly producing modules and hacking around the limitations, but that wasn't sustainable so I started this multi-year project. Your "end game" seems to be about fixing all the bugs but otherwise leaving things mostly as-is (excluding some in progress work such as the clipboard API). I think together it works better at pushing the project forward both in term of a quality and capability/modernity point of views.

So I would leave things as-is. It's not perfect but it eventually works. I am not entitled to expect the 3.5k LoC of "ready to review" pull requests to be reviewed in a single day. As long as there is some sign of life within a week or so I am ok.

> As in, you want me to look at [this] first?

This PR is bugfixes tagged with v4.3 and not a large feature addition. It's small and fixes existing issues. Those bugs have been there either for years (the minimization one) or forever (the focusable one).

lib/awful/ewmh.lua
```diff
@@ -388,6 +401,15 @@ function ewmh.client_geometry_requests(c, context, hints)
     end
 end
 
+-- The magnifier layout doesn't work with focus follow mouse.
+ewmh.add_activate_filter(function(c)
+    if alayout.get(c.screen) ~= alayout.suit.magnifier
```
Elv13: Maybe this should move to magnifier.lua?

psychon:
> Your "end game" seems to be about fixing all the bugs but otherwise leaving things mostly as-is (excluding some in progress work such as the clipboard API).

Or rather, I prioritise fixing bugs instead of coming up with new code that brings new bugs. And calling the clipboard API "in progress" is quite a joke ;-) (although, as far as I remember, the code in that PR works fine, as long as you don't end up initiating large transfers since INCR is not yet implemented (and generally ugly, I prefer wayland's FD passing solution a lot over INCR transfers...)).
If only the day had more than 24 hours...

actionless:
>So, I just invented a new, more public tracking scheme: A psychon-should-take-a-look-label.

@psychon what about using "Assignees" field?

psychon:
I don't think that assigning PRs to me has the right meaning

actionless:
github allows to `Assign up to 10 people to this pull request` so i think it's quite fine, also there is another possibility to `Request a review`

actionless:
@psychon i've just checked, you can self-request a review

PR #2333, opened 2018-08-03
Run example-tests during build phase, not configure phase

psychon:
Before this:
- `cmake .. && make` takes ~25.9 seconds
- `cmake .. && make -j10` takes ~17.9 seconds
- Re-running `cmake .` in an already configured build directory takes ~8.8 seconds

After this:
- `cmake .. && make` takes ~26.8 seconds
- `cmake .. && make -j10` takes ~12.0 seconds
- Re-running `cmake .` in an already configured build directory takes ~0.3 seconds

(Numbers were only measured once and thus are noisy)

Especially the last point, re-running CMake, is what made working on awesome more painful than necessary, because changing any source file means that the `@FOO@` expansions were considered out of date and you get to wait quite a while for the build.

### DIFF
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -24,15 +24,13 @@ execute_process(COMMAND lua -e "p = package.path:gsub(';', '\\\\;'); io.stdout:w
 if(NOT SOURCE_DIR AND ${CMAKE_CURRENT_SOURCE_DIR} MATCHES "/tests/examples")
     get_filename_component(TOP_SOURCE_DIR
                            "${CMAKE_CURRENT_SOURCE_DIR}/../.." ABSOLUTE)
-    # Used by .luacov.
-    set(ENV{SOURCE_DIRECTORY} ${TOP_SOURCE_DIR})
 else()
     set(TOP_SOURCE_DIR ${CMAKE_SOURCE_DIR})
 endif()
 
 if (DO_COVERAGE)
   execute_process(
-    COMMAND lua -e "require('luacov.runner')('${TOP_SOURCE_DIR}/.luacov')"
+    COMMAND env "SOURCE_DIRECTORY=${TOP_SOURCE_DIR}" lua -e "require('luacov.runner')('${TOP_SOURCE_DIR}/.luacov')"
     RESULT_VARIABLE TEST_RESULT
     ERROR_VARIABLE  TEST_ERROR
     ERROR_STRIP_TRAILING_WHITESPACE)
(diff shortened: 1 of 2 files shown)


psychon:
Sigh... why does this work locally, but on Travis the example tests are just not run?!?

actionless:
@psychon 
```diff
diff --git a/.travis.yml b/.travis.yml
index a0dd7f480..b0e65793d 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -190,7 +190,7 @@ script:
         # Run tests/examples explicitly.
         (mkdir -p tests/examples/build \
          && cd tests/examples/build \
-         && cmake $CMAKE_ARGS ..)
+         && cmake $CMAKE_ARGS .. && make)
         do_codecov samples

         make check-unit-coverage || exit 1
```

blueyed:
Thank you!

psychon:
@actionless Oh, right. Thanks! However, this should already have run the example tests, I think:
```
  - travis_run_in_fold "make" make
```

tests/examples/CMakeLists.txt
```diff
@@ -24,15 +24,13 @@ execute_process(COMMAND lua -e "p = package.path:gsub(';', '\\\\;'); io.stdout:w
...
-    # Used by .luacov.
-    set(ENV{SOURCE_DIRECTORY} ${TOP_SOURCE_DIR})
 else()
     set(TOP_SOURCE_DIR ${CMAKE_SOURCE_DIR})
 endif()
 
 if (DO_COVERAGE)
   execute_process(
-    COMMAND lua -e "require('luacov.runner')('${TOP_SOURCE_DIR}/.luacov')"
+    COMMAND env "SOURCE_DIRECTORY=${TOP_SOURCE_DIRECTORY}" lua -e "require('luacov.runner')('${TOP_SOURCE_DIR}/.luacov')"
```
blueyed: Currently it fails to generate coverage:

> build/luacov.stats.out does not exist.

Likely related to this change I would guess (also above maybe).

psychon: My guess is that the missing coverage happens because `make` does not run the example tests. The later run that is now there thanks to @actionless has a different build dir, so will place the `luacov.stats.out` file somewhere completely different...

actionless: so, should SOURCE_DIRECTORY here be changed for that or should example build be changed to put output to a different location?

psychon: This PR shouldn't change anything. Things should still work as before.

Which means that the example tests should be run during `make` (with which I mean our wrapper-Makefile that runs `cmake` and `make` in a build dir). For some reason this does not happen on Travis, but it does happen on the two computers that I tested it with. And debugging things on Travis is annoying. :-/

psychon:
Oh -.-
I found the obvious:
```cmake
if (GENERATE_DOC AND DO_COVERAGE)
    message(STATUS "Not generating API documentation with DO_COVERAGE")
    set(GENERATE_DOC OFF)
endif()
```
So this is not a case of "it was running the tests here previously, why did it stop?", but it really is a problem of getting the luacov output file in the right place (but the question is still: What changed?).

psychon:
I found another "obvious" thing: `$TOP_SOURCE_DIR` is not the same as `$TOP_SOURCE_DIRECTORY`.

psychon:
Okay, those two obvious problems where the cause of the build failure. Is this good to go now?

PR #2334, opened 2018-08-03
Add support for motif wm hints

psychon:
This makes motif wm hints available on clients as c.motif_wm_hints.
Actually interpreting all the values is up to Lua. The definition of the
necessary properties is taken from motif.

-----

Side note: I have no idea what I am doing here. I just stared at some header and guessed how this property could be interpreted. The note about "all" meaning that the other bits are inverted is not even from the headers, but was found somewhere else. So... I basically expect this to be broken.


Elv13:
Question: We have a Lua `xproperty` API. Could it be used instead of this to move this code out of CAPI? If it isn't then it is a general problem worth solving or is this a "one time case"? For example, if someone wants to control URXVT or Compton using their xproperties, is this all that different from this `client_set_motif_wm_hints`? If the `xproperty` API doesn't support this, then should is be extended? Note that, as always, I have little interest in that part and I am asking these question without doing proper homework. I just saw this and wondered if doing so in C is a bit counterproductive.

psychon:
> Could it be used instead of this to move this code out of CAPI?

Our xproperty API only supports properties of type string, number, and boolean (that's the types we export to Lua; in X11 the types are UTF8_STRING and CARDINAL). The motif WM hints have type, `_MOTIF_WM_HINTS` (or something like that).

The use case that gave us the xproperty API was to save things across restarts of awesome. It was never meant to read or write arbitrary properties.

> For example, if someone wants to control URXVT or Compton using their xproperties, is this all that different from this client_set_motif_wm_hints?

This PR only provides reading the motif WM hints. The WM has no business setting them. `client_set_motif_wm_hints` is called from `property.c` to set the property of a client. It's not exported to Lua.

> If the xproperty API doesn't support this, then should is be extended? 

If someone can come up with a use case for it, then I might think about it. For now I'd say "no". X11 is complicated enough already without random code messing with random X11 properties... (oh wait, that's already a reality, just that the "random code" is so far not "Lua code running in the WM" ;-) ).

The reason why I started this PR is GTK's client side decorations: Someone asked on IRC why GTK windows get titlebars on awesome while that's not the case in other WMs and how to replicate the "other WM" behaviour in awesome.
(The answer to this question would then be: `c.motif_wm_hints and c.motif_wm_hints.decorations and (((not c.motif_wm_hints.decorations.all) and (c.motif.wm_hints.decorations.border)) or (c.motif_wm_hints.decorations.all and (not c.motif_wm_hints.decorations.border))`, or something like that. I'm almost sure I forgot to close a bracket in there; also, this only looks at the "border" hint whose meaning I don't know and this ignores all the other hints; perhaps `title` is a better fit?).

Elv13:
> The use case that gave us the xproperty API was to save things across restarts of awesome. It was never meant to read or write arbitrary properties.

True, but that was the point of the question. In the event Waycooler becomes a thing and XWayland clients have these properties, should we keep adding more C code to decode them? I mean, messing with binary payloads in Lua isn't that bad. v5.3 makes it very trivial, but even older ones are ok. Anyway, that was just an opinion. I have no strong feeling about this PR.

> The reason why I started this PR is GTK's client side decorations

Can the test client be updated to toggle CSD/SSD so we can test this?

Also, can you add an higher level `provide_titlebar` client property be added to `awful.client.object` and then to `rc.lua` 

```
properties = { titlebars_enabled = true }
```

```
properties = { titlebars_enabled = function(c) not c.provide_titlebar end }
```

It doesn't need to be in this PR or for v4.3, it's just suggestion to actually make solving the problem it attempts to solve generally available.

psychon:
> Can the test client be updated to toggle CSD/SSD so we can test this?

As far as I know: no. I recently saw https://github.com/PCMan/gtk3-nocsd which is 1.5k of C code which is `LD_PRELOAD`ed (or something similar).

> Also, can you add an higher level provide_titlebar client property be added to awful.client.object...

I don't like that name, but for the idea: sure

> ...and then to rc.lua

Hm, really? Personally, I don't like what GTK is doing. They claim that CSD provides consistency, but I think it does the exact opposite.

Elv13:
> As far as I know: no. I recently saw https://github.com/PCMan/gtk3-nocsd which is 1.5k of C code which is LD_PRELOADed (or something similar).

As far as I have been told, it is now possible for the WM to ask GTK to enable or disable CSD in the newest version. (Again, as I have been told by KWin new maintainer) they added a protocol for both X11 and Wayland to negotiate CSD or SSD.

If that doesn't happen to be practical, then feel free to ignore this.

> Hm, really? Personally, I don't like what GTK is doing. They claim that CSD provides consistency, but I think it does the exact opposite.

I generally agree, but now we have 2 titlebars, it's IMHO worst than letting GTK3 do their thing and respect it. I have no ultra strong opinion on this. If you feel it brings more inconsistency, then maybe we could simply add the proper documentation and code examples. That or investigate and see if that new GTK 3.24 protocol to control decoration is really a thing and worth implementing (I know KWin now implements it for Wayland and it was supposed to also exist in X11).

> @Elv13 i would rather use some more usual name like bool c.client_side_decorations but idea itself is wonderful

That was my first naming idea, but then we have the problem where there is something called CSD in the API and something called titlebars. Isn't it a bit confusing for non-GTK_vs_KDE_flamewar_over_decoration savvy users?

actionless:
you could ask your kwin insider what does they think about naming that feature CSD :-) but actually it seems what it implemented only in gtk, so it would weird to name it differently than it named there

Elv13:
> but actually it seems what it implemented only in gtk

Far from it. This hint is literally from the 80's. It is implemented by many (most?) toolkits, all dock/bars. splash screens and more. I recall seeing CSD (before the name was coined) back to the late 90's in apps like XMMS (gtk1) and KRuler (Qt2)

actionless:
in that case let's stick with the name you proposed, i think it should go in a separate PR, right?

psychon:
> they added a protocol for both X11 and Wayland to negotiate CSD or SSD.

I bet I could find the wayland protocol with ~5 minutes of work (it will most likely be somewhere in `wayland-protocols`). In fact: `grep decoration $(rpm --query --filesbypkg wayland-protocols-devel-1.10-1.fc27.noarch | cut -f2- -d ' ')` says it is in `xdg-shell-unstable-v5.xml`.

For an X11 protocol: No idea where I should even start looking.

> i think it should go in a separate PR, right?

Right.

psychon:
Here is an (untested) patch ontop of this PR that adds a simpler property based on this:
```diff
diff --git a/lib/awful/client.lua b/lib/awful/client.lua
index 61b97681..b94f4ba5 100644
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -1040,6 +1040,38 @@ function client.dockable.set(c, value)
     client.property.set(c, "dockable", value)
 end
 
+--- If the client requests not to be decorated with a titlebar.
+--
+-- The motif wm hints allow a client to request not to be decorated by the WM in
+-- various ways. This property uses the motif MWM_DECOR_TITLE hint and
+-- interprets it as the client (not) wanting a titlebar.
+--
+-- **Signal:**
+--
+-- * *property::requests_no_titlebar*
+--
+-- @property requests_no_titlebar
+-- @param boolean Whether the client requests not to get a titlebar
+
+function client.object.get_requests_no_titlebar(c)
+    local hints = c.motif_wm_hints
+    if not hints then return false end
+
+    local decor = hints.deocrations
+    if not decor then return false end
+
+    local result = not decor.title
+    if decor.all then
+        -- The "all" bit inverts the meaning of the other bits
+        result = not result
+    end
+    return result
+end
+client.connect_signal("property::motif_wm_hints", function(c)
+    -- We cannot be sure that the property actually changes, but whatever
+    c:emit_signal("property::requests_no_titlebar")
+end)
+
 --- Get a client property.
 --
 -- This method is deprecated. It is now possible to use `c.value` directly.
```
Testing this patch should be simple: Just change `titlebars_enabled = true` to `titlebars_enabled = function(c) return not c.requests_no_titlebar end` and see if GTK applications still get titlebars.
(Also, feel free to bikeshed over the name; I don't like `c.client_side_decorations` nor `c.provide_titlebar` because nothing actually guarantees that the client does CSD / provides a titlebar; I saw enough "stuff" out there that did crazy stuff for crazy reasons; Google claims steam sets `_MOTIF_WM_HINTS` on its window...)

PR #2339, opened 2018-08-06
build: Auto-add example tests outputs.

Elv13:
Next step

This PR add the runner.sh code again. To mitigate @psychon concerns, I propose to check if the number of outputs file is the same before and after running the tests when `${STRICT_TESTS}` is set. This avoid the overhead when adding new tests locally.

Another change might be to add a commit hook that checks if the commit add files in `tests/examples` and automatically `git add foo.output.txt` if it exists.

This pull request does not implement the last 2 ideas and exists mainly to start the discussion.

### DIFF
--- a/tests/examples/runner.sh
+++ b/tests/examples/runner.sh
@@ -38,6 +38,12 @@ then
     exit 1
 fi
 
+# Automatically add the output for new examples
+if [ ! -e "${expected_output}" ]
+then
+    cp "${file_stdout}" "${expected_output}"
+fi
+
 # Check if we got the output we wanted
 if ! cmp --silent "${file_stdout}" "${expected_output}"
 then


tests/examples/runner.sh
```diff
@@ -38,6 +38,12 @@ then
     exit 1
 fi
 
+# Automatically add the output for new examples
+if [ ! -e "${expected_output}" ]
+then
+    cp "${file_stdout}" "${expected_output}"
```
psychon: How about `echo "${expected_output} did not yet exist and was created"; exit 1`? It's not much since the next run will just succeed, but it might help in making it more noticable that this file was created?
But the current version is fine with me as well. We can bikeshed this later. ;-)

Elv13: Ok, I will merge as-is given the now 3 proposed changes are probably better suited after after #2333 is merged. Doing so right now in the configure phase isn't the same as doing it during `make`. Thus this commit fixes an immediate issue ("adding new tests is a PITA") while #2333 fixes "adding more tests is slowing down the build and is a PITA".

psychon:
```
== Running /home/travis/build/awesomeWM/awesome/tests/test-leaks.lua ==
Error org.freedesktop.DBus.Error.NoReply: Message did not receive a reply (timeout by message bus)
E: dbus-send failed.
```
Hm... no idea what is going on here. How about the following?
```diff
diff --git a/tests/run.sh b/tests/run.sh
index 008bbecc..1c22ed02 100755
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -222,7 +222,7 @@ for f in $tests; do
     fi
 
     # Execute the test file in awesome.
-    DISPLAY=$D "$AWESOME_CLIENT" 2>&1 "dofile('$f')"
+    DISPLAY=$D "$AWESOME_CLIENT" 2>&1 "dofile('$f')" || true
 
     # Tail the log and quit, when awesome quits.
     # Use a single `grep`, otherwise `--line-buffered` would be required.
```
The test should still fail since there most likely won't be a "test finished successfully" in the log, but... perhaps this could actually mask real failures?

Or does this version make more sense?
```diff

```diff --git a/tests/run.sh b/tests/run.sh
index 008bbecc..2aa73eea 100755
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -222,7 +222,7 @@ for f in $tests; do
     fi
 
     # Execute the test file in awesome.
-    DISPLAY=$D "$AWESOME_CLIENT" 2>&1 "dofile('$f')"
+    DISPLAY=$D "$AWESOME_CLIENT" 2>&1 "dofile('$f')" || echo "Error: awesome-client failed" >> "$awesome_log"
 
     # Tail the log and quit, when awesome quits.
     # Use a single `grep`, otherwise `--line-buffered` would be required.
```

Elv13:
I say it's a race in the test GC calls. I had this locally since the first day you added this test suits. Given it sometime passes, I always assumed it's a false positive and ignored it.

psychon:
I'd guess some random memory corruption between LGI and Lua, but random memory corruptions are hard to debug...

PR #2340, opened 2018-08-06
tests/run.sh: Print log on awesome-client failures

psychon:
tests/run.sh uses awesome-client to actually make awesome run the test.
If awesome-client fails, then the only error message that is printed look like:
```
== Running /home/travis/build/awesomeWM/awesome/tests/test-leaks.lua ==
Error org.freedesktop.DBus.Error.NoReply: Message did not receive a reply (timeout by message bus)
E: dbus-send failed.
```
To also get the output from awesome, this commit makes failures from
awesome-client non-fatal so that the following shell code can notice the
failure and hopefully print some useful information.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -222,7 +222,7 @@ for f in $tests; do
     fi
 
     # Execute the test file in awesome.
-    DISPLAY=$D "$AWESOME_CLIENT" 2>&1 "dofile('$f')"
+    DISPLAY=$D "$AWESOME_CLIENT" 2>&1 "dofile('$f')" || echo "Error: awesome-client failed" >> "$awesome_log"
 
     # Tail the log and quit, when awesome quits.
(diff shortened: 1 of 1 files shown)


psychon:
> The faster this gets merged, the sooner we will gather enough data to track the heisenbug.

If I remember correctly, I once built Lua 5.3 with assertions enabled and tried to figure something out. But if I still remember correctly, I didn't get anywhere and the "fail" / "not fail" status only changed randomly, so I just gave up. But at least we would know if this is the same assertion failure or not. And perhaps one day we get a good reproduce for this problem that is perhaps even free of awesome and uses just Lua and LGI...

Elv13:
For the test-leaks? No, it's just the test failing/timeout. It happens on all Lua versions. For the Lua5.3 assert, Awesome "crashes". It's not the only test I have seen failing randomly. But again, in most case it's just that since you removed the 100ms delay, it sometime goes too fast.

Issue #845, opened 2016-04-21
Java application maximized window not raising up


hari-rangarajan:
I have a set of maximized java application windows on a tag. After a while, some windows do not raise to the top of the stack, but they do get focus and switching to tiling gets them to work.  Restarting awesome at that point fixes it. Can you give me some pointers on what could be going wrong to get some debugs or other useful info?

I noticed a few other bugs related to maximized windows, wonder if they are related to this.

Thanks

psychon:
How do you maximize these windows? How do you focus this window? (E.g. "through Java" or "by pressing mod+f and mod+j", "clicking on things", ....)
Do you happen to know some not-Java program that has these problem?
What's the `xprop` output for these windows? (Specifically: Do the windows have `WM_TAKE_FOCUS` in their `WM_PROTOCOLS` and do they have `Accepts input focus: true / false` in their `WM_HINTS` (or whatever this property was called...).

hari-rangarajan:
> How do you maximize these windows? 
Maximize via shortcuts as well as through rc.lua properties.

> How do you focus this window? (E.g. "through Java" or "by pressing mod+f and mod+j", "clicking on things", ....)
all of the above

xprops:

WM_HINTS(WM_HINTS):
		Client accepts input or input focus: False
		Initial state is Normal State.

WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, WM_TAKE_FOCUS

I haven't found a pattern with respect to the controllable parameters. It works for a while before getting stuck and an awesine restart fixes it (possibly indicating internal state?). Are there debugs to see why a window refuses to come up the stack?

psychon:
These Java windows are saying us "don't focus me (WM_HINTS), just send me a message instead (WM_TAKE_FOCUS)". What Java then does with this is up to Java, it's allowed to set the input focus directly.
However, who raises these windows to the top of the stack? I'm not sure... This depends on how you focus the window. E.g. for clicking on the tasklist, the code in the default config focuses and raises a client `client.focus = c; c:raise()`. I doubt that Java raises its own windows, but this might be possible. And especially I cannot think of a way to have a race condition here, because the focusing and the raising are independent...

hari-rangarajan:
It seems to me that after moving the focus between the windows a certain number of times, the stacking order is somehow going wrong. If I tile the windows, the focus is on the correct window. Just for debugging, can I play around with the stacking order?

psychon:
[...]

PR #2345, opened 2018-08-10
Handle unsetting of .transient_for

psychon:
When a window has a WM_TRANSIENT_FOR property that is later unset,
awesome would still keep c.transient_for pointing to the previous
"parent client". This commit fixes that.

First, property_update_wm_transient_for() is fixed so that it unsets
c->transient_for_window if the WM_TRANSIENT_FOR property is deleted.
Additionally, this then calls client_find_transient_for() to update the
c->transient_for pointer.

Secondly (and a bit unrelated), this changes client_find_transient_for()
so that it always sets c->transient_for. Previously, if updating this
property would introduce a cycle in the transient_for relation, it would
just leave c->transient_for with its old value. After this change, it
gets explicitly set to NULL instead.

CC @p-himik Here is a new version of the patch (without the debugging-prints). The important part is the change to `property.c`. While writing the commit message, I noticed that the change to `client.c` is actually unrelated to this issue, but I was too lazy to change the commit...

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -983,7 +983,9 @@ client_find_transient_for(client_t *c)
 {
     int counter;
     client_t *tc, *tmp;
+    lua_State *L = globalconf_get_lua_State();
 
+    /* This might return NULL, in which case we unset transient_for */
     tmp = tc = client_getbywin(c->transient_for_window);
 
     /* Verify that there are no loops in the transient_for relation after we are done */
@@ -994,14 +996,16 @@ client_find_transient_for(client_t *c)
             counter = globalconf.stack.len+1;
         tmp = tmp->transient_for;
     }
-    if (counter <= globalconf.stack.len)
-    {
-        lua_State *L = globalconf_get_lua_State();
 
-        luaA_object_push(L, c);
-        client_set_transient_for(L, -1, tc);
-        lua_pop(L, 1);
+    if (counter > globalconf.stack.len)
+    {
+        /* There was a loop, so unset .transient_for */
+        tc = NULL;
     }
+
+    luaA_object_push(L, c);
+    client_set_transient_for(L, -1, tc);
+    lua_pop(L, 1);
 }
 
 void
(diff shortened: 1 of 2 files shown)


Elv13:
(/offtopic I am AFK for 1.5 week, should be available for review again on Aug. 23.)

p-himik:
@psychon With this patch, the issue has not been reproduced for almost a whole day.

p-himik:
@psychon Yeah, I agree. Maybe even not in IntelliJ but in Java. I'll test it for a few more days with your patch and then for a few days with another solution offered at https://youtrack.jetbrains.com/issue/IDEA-155759#focus=streamItem-27-3012463-0-0

p-himik:
The solution from JetBrains didn't work at all, but this patch worked flawlessly.

p-himik:
Mentioning #845 just for a cross-link.

psychon:
@p-himik (At least for me) the `#`-part of your link does not lead anywhere, so I'm not quite sure which comment in there you are referring to.
If you have an account on that tracker, you might want to post an update there:
- The problem is that Java/IntelliJ does funny stuff with its `WM_TRANSIENT_FOR` property, temporarily declaring one of its main windows as a dialog
- Awesome always raises a window's transients over the window itself, which is why one of the main windows could no longer be raised over the other one (because the top-most one was considered a transient of the other main window)
- There is a bug in awesome here in that it does not handle deletion of the `WM_TRANSIENT_FOR` property
- Still, the stuff that happens to this property "smells funny" and should be investigated
- The weird input-related things ("Window A shows as focused, but input is sent to window B") could very well be an unrelated bug, but it could also be that when the JDK "thinks" that window B has the focus, then all input events are handled by/for window B even though they might be sent to window A. Dunno.

psychon:
> the relevant comments are just not there

Ah, so there is activity in the last two years, they just decide not to show it... Great. (It's there if you print the web page, but come on, seriously?). In other news, I am going to ignore links to YouTrack from now on.

p-himik:
@psychon Please don't ignore them. :) I figured it out - the comment I linked previously was left with a limited visibility for whatever reason. And the rest of the activity is shown, but the order is inverted - in Incognito mode, the newest comments are on top for me.

PR #2358, opened 2018-08-18
doc: Add filter section for the taglist and tasklist

Elv13:
This is cherry picked from the [rebased version of the] popup branch.

This is done for most other independent components already. It groups the elements in the doc and makes everything more readable. It will also be used by the layoutlist (from the popup PR, soon to be its own PR) and the client layout module `tablist` widget.

### DIFF
--- a/lib/awful/widget/taglist.lua
+++ b/lib/awful/widget/taglist.lua
@@ -550,6 +550,7 @@ end
 --- Filtering function to include all nonempty tags on the screen.
 -- @param t The tag.
 -- @return true if t is not empty, else false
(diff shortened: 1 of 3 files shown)


lib/awful/widget/taglist.lua
```diff
@@ -550,19 +550,22 @@ end
 --- Filtering function to include all nonempty tags on the screen.
 -- @param t The tag.
 -- @return true if t is not empty, else false
+-- @filterfunction awful.taglist.filer.noempty
```
actionless: typo filer/filter

Elv13: done

PR #2359, opened 2018-08-18
Add common source

Elv13:
This change solves 2 problems:

 * While the taglist and tasklist objects or managed internally, layoutlist allows random ones to be created and it was difficult to handle when the layouts were removed from `awful.layout.layouts` (for example, a `lain` layout not added to the default list)
 * Ordering wasn't trivial and prevented some code re-use. This is useful if you want to order your alt+tab popup with the newest at the top

### DIFF
--- a/lib/awful/widget/tasklist.lua
+++ b/lib/awful/widget/tasklist.lua
@@ -245,7 +245,7 @@ local instances
 -- @see gears.color
 
 -- Public structures
-tasklist.filter = {}
+tasklist.filter, tasklist.source = {}, {}
 
 local function tasklist_label(c, args, tb)
     if not args then args = {} end
@@ -401,7 +401,11 @@ end
 
 local function tasklist_update(s, w, buttons, filter, data, style, update_function, args)
     local clients = {}
-    for _, c in ipairs(capi.client.get()) do
+
+    local source = args and args.source or tasklist.source.all_clients or nil
+    local list   = source and source(s, args) or capi.client.get()
+
+    for _, c in ipairs(list) do
         if not (c.skip_taskbar or c.hidden
             or c.type == "splash" or c.type == "dock" or c.type == "desktop")
             and filter(c, s) then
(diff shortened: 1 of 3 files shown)


lib/awful/widget/tasklist.lua
```diff
@@ -430,6 +434,8 @@ end
 --   update. See `awful.widget.common.list_update`.
 -- @tparam[opt] table args.layout Container widget for tag widgets. Default
 --   is `wibox.layout.flex.horizontal`.
+-- @tparam[opt=awful.tasklist.source.all_clients] function args.source The
```
actionless: mb `source_function`, to be inline with `update_function`?

Elv13: filter isn't called "filter_function". So I am not sure about this...

lib/awful/widget/tasklist.lua
```diff
@@ -672,6 +678,15 @@ function tasklist.filter.focused(c, screen)
     return get_screen(c.screen) == get_screen(screen) and capi.client.focus == c
 end
 
+--- Get all the clients in an undefined order.
```
psychon: The order is not really undefined. Dunno, if that's worth mentioning.
This is "the" order of clients. By default, this is the order in which they were created, but `:swap` modifies this order. This is e.g. also the order according to which the "next" client is defined.

Elv13: /me Back to civilization, sorry for the AFK week.

> Dunno, if that's worth mentioning.

I don't think so. I used "undefined" instead of "random" on purpose. To me, the order is an implementation detail and it doesn't make much sense on its own.

lib/awful/widget/tasklist.lua
Elv13: So... What's your opinion? Is my answer ok or do you really want this change? I am waiting for this to be merged to propose the pull request that uses it.

lib/awful/widget/tasklist.lua
psychon: Well... I approved this a week ago. So I guess, a week ago I thought this was good to go. That's good enough for me so that I don't take a second look.

Off-topic: I'll be AFK next week

Elv13:
Rebased + fixed a luacheck unused function parameter warning

PR #2360, opened 2018-08-18
awful.tag: Add awful tag layouts

Elv13:
Another one split out of the bigger popup PR.

There was an issue in my previous layoutlist implementation where if `awful.layout.layouts` was "trimmed down" or the user had an external module like lain, it would cause errors.

This PR adds `t.layouts` as a tag level way to set the accepted layouts. By default, it will try to squash `awful.layout.layouts` with the non-listed ones. If the user set this property manually, it's their problem and they have to handle this themselves.

Plus, it's nice since it allow different layout order depending on the tag. I see myself using that now.

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -21,6 +21,7 @@ local client = require("awful.client")
 local ascreen = require("awful.screen")
 local timer = require("gears.timer")
 local gmath = require("gears.math")
+local gtable = require("gears.table")
 local protected_call = require("gears.protected_call")
 
 local function get_screen(s)
@@ -73,6 +74,16 @@ layout.layouts = {
 --
 -- @field layout.layouts
 
+--- Return the tag layout index (from `awful.layout.layouts`).
+--
+-- If the layout isn't part of `awful.layout.layouts`, this function returns
+-- nil.
+--
+-- @tparam tag t The tag.
+-- @treturn nil|number The layout index.
+function layout.get_tag_layout_index(t)
+    return gtable.hasitem(layout.layouts, t.layout)
+end
 
 -- This is a special lock used by the arrange function.
 -- This avoids recurring call by emitted signals.
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -14,6 +14,7 @@ local gmath = require("gears.math")
 local object = require("gears.object")
 local timer = require("gears.timer")
 local gtable = require("gears.table")
+local alayout = nil
 local pairs = pairs
 local ipairs = ipairs
 local table = table
@@ -72,6 +73,68 @@ local function raw_tags(scr)
     end
 
     return tmp_tags
+end
+
+local function custom_layouts(self)
+    local cls = tag.getproperty(self, "_custom_layouts")
+
+    if not cls then
+        cls = {}
+        tag.setproperty(self, "_custom_layouts", cls)
+    end
+
+    return cls
+end
+
+-- Update the "user visible" list of layouts. If `from` and `to` are not the
+-- same, then `from` will be replaced. This is necessary for either the layouts
+-- defined as a function (called "template" below) and object oriented, stateful
+-- layouts where the original entry is only a constructor.
+local function update_layouts(self, from, to)
+    if not to then return end
+
+    alayout = alayout or require("awful.layout")
+    local override = tag.getproperty(self, "_layouts")
+
+    local pos = from and gtable.hasitem(override or {}, from) or nil
+
+    -- There is an override and the layout template is part of it, replace by
+    -- the instance.
+    if override and pos and from ~= to then
+        assert(type(pos) == 'number')
+        override[pos] = to
+        self:emit_signal("property::layouts")
+        return
+    end
+
+    -- Only add to the custom_layouts and preserve the ability to globally
+    -- set the layouts.
+    if override and not pos then
+        table.insert(override, to)
+        self:emit_signal("property::layouts")
+        return
+    end
+
+    pos = from and gtable.hasitem(alayout.layouts, from) or nil
+
+    local cls = custom_layouts(self)
+
+    -- The new layout is part of the global layouts. Fork the list.
+    if pos and from ~= to then
+        local cloned = gtable.clone(alayout.layouts, false)
+        cloned[pos] = to
+        gtable.merge(cloned, cls)
+        self.layouts = cloned
+        return
+    end
+
+    if pos then return end
+
+    if gtable.hasitem(cls, to) then return end
+
+    -- This layout is unknown, add it to the custom list
+    table.insert(cls, to)
+    self:emit_signal("property::layouts")
 end
 
 --- The number of elements kept in the history.
@@ -718,7 +781,26 @@ end
 -- @tparam layout|function layout A layout table or a constructor function
 -- @return The layout
 
+--- The (proposed) list of available layouts for this tag.
+--
+-- This property allows to define a subset (or superset) of layouts available
+-- in the "rotation table". In the default configuration file, `Mod4+Space`
+-- and `Mod4+Shift+Space` are used to switch between tags. The
+-- `awful.widget.layoutlist` also uses this as its default layout filter.
+--
+-- By default, it will be the same as `awful.layout.layouts` unless there the
+-- a layout not present is used. If that's the case they will be added at the
+-- front of the list.
+--
+-- @property layouts
+-- @param table
+-- @see awful.layout.layouts
+-- @see layout
+
 function tag.object.set_layout(t, layout)
+
+    local template = nil
+
     -- Check if the signature match a stateful layout
     if type(layout) == "function" or (
         type(layout) == "table"
@@ -741,12 +823,39 @@ function tag.object.set_layout(t, layout)
             t.dynamic_layout_cache[layout] = instance
         end
 
+        template = layout
         layout = instance
     end
 
     tag.setproperty(t, "layout", layout)
 
+    update_layouts(t, template or layout, layout)
+
     return layout
+end
+
+function tag.object.get_layouts(self)
+    local override = tag.getproperty(self, "_layouts")
+
+    if override then
+        return override
+    end
+
+    -- Required to get the default/fallback list of layouts
+    alayout = alayout or require("awful.layout")
+
+    local cls = custom_layouts(self)
+
+    -- Without the clone, the custom_layouts would grow
+    return #cls > 0 and gtable.merge(gtable.clone(cls, false), alayout.layouts) or
+        alayout.layouts
+end
+
+function tag.object.set_layouts(self, layouts)
+    tag.setproperty(self, "_custom_layouts", {})
+    tag.setproperty(self, "_layouts", gtable.clone(layouts, false))
+    update_layouts(self, self.layout, self.layout)
+    self:emit_signal("property::layouts")
 end
 
 function tag.object.get_layout(t)
--- a/tests/test-awful-tag.lua
+++ b/tests/test-awful-tag.lua
@@ -1,4 +1,5 @@
 local awful = require("awful")
+local gtable = require("gears.table")
 local beautiful = require("beautiful")
 
 local function check_order()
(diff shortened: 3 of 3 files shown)


lib/awful/layout/init.lua
```diff
@@ -73,6 +74,16 @@ layout.layouts = {
 --
 -- @field layout.layouts
 
+--- Return the tag layout index (from `awful.layout.layouts`).
+--
+-- If the layout isn't part of `awful.layout.layouts`, this function returns
+-- null.
```
psychon: `nil`, not `null`

lib/awful/tag.lua
```diff
@@ -749,6 +766,32 @@ function tag.object.set_layout(t, layout)
...
 end
 
+function tag.object.get_layouts(self)
+    -- The override has the priority.
+    local cls = tag.getproperty(self, "custom_layouts")
+
+    -- Allow custom layouts to be listed until GC-ed.
+    if not cls then
+        cls = setmetatable({}, {__mode="v"})
+        tag.setproperty(self, "custom_layouts", cls)
```
psychon: Why this GC magic? Can't you just return the table 'normally'? Do any callers save the table? If so, why is it fine if the GC "silently messes" with the table?

lib/awful/tag.lua
```diff
@@ -718,6 +719,22 @@ end
...
 
+--- The (proposed) list of available layouts for this tag.
+--
+-- This property allows to define a subset (or superset) of layouts available
+-- in the "rotation table". In the default configuration file, `Mod4+Space`
+-- and `Mod4+Shift+Space` are used to switch between tags. The
+-- `awful.widget.layoutlist` also uses this as its default layout filter.
+--
+-- By default, it will be the same as `awful.layout.layouts` unless there the
+-- a layout not present is used. If that's the case they will be added at the
```
psychon: "there the a"?

How about:

> By default, it will be the same as `awful.layout.layouts` unless the tag currently uses a layout that is not listed there. In this case, the current layout will be prepended to the front of the list.

And why this prepending behaviour? Currently we also don't auto-magically add things to the list of layouts.

lib/awful/tag.lua
```diff
@@ -749,6 +766,32 @@ function tag.object.set_layout(t, layout)
...
+    local cls = tag.getproperty(self, "custom_layouts")
+
+    -- Allow custom layouts to be listed until GC-ed.
+    if not cls then
+        cls = setmetatable({}, {__mode="v"})
+        tag.setproperty(self, "custom_layouts", cls)
+    end
+
+    -- Required to get the default/fallback list of layouts
+    alayout = alayout or require("awful.layout")
```
psychon: More cyclic dependencies? :-(

lib/awful/tag.lua
```diff
@@ -749,6 +766,32 @@ function tag.object.set_layout(t, layout)
...
+    if (not gtable.hasitem(cls, self.layout))
+      and not alayout.get_tag_layout_index(self) then
+        table.insert(cls, self.layout)
+    end
+
+    return gtable.merge(gtable.clone(cls), alayout.layouts)
+end
+
+function tag.object.set_layouts(self, layouts)
+    tag.setproperty(self, "layouts", layouts)
```
psychon: Where is this property `get`ed? Or does setting a property directly mean that the getter is no longer used?

lib/awful/tag.lua
Elv13: Because layouts can be objects (that part of my dynamic work was merged before 4.0) and it will leak if it uses a normal table.

lib/awful/tag.lua
Elv13: > Currently we also don't auto-magically add things to the list of layouts.

You sure can:

```
awful.layout.layout = {awful.layout.suit.max}

awful.tag.add("mytag", {
    layout = awful.layout.suit.tile.left,
    layouts = {awful.layout.suit.tile.right},
})
```

With things like the lain layouts, this will totally happen. Plus, if you got a "global" `layoutlist` widget in your wibar and click on a layout not part of `mytag.layouts`, then it will trigger the error/bug too when you display the `layoutlist` for this tag. Instead of trying to fix the error in the `layoutlist`, better handle it at the source, here.

lib/awful/tag.lua
Elv13: refactoring facepalm

Elv13:
Updated

lib/awful/tag.lua
psychon: Huh? What would it leak? If it leaks a layout, then I still have to ask again: "why is it okay when the list of available layouts on a tag changes randomly" (where "randomly" depends on the behaviour of the GC)

lib/awful/tag.lua
psychon: Huh? How does the order of layouts fix bugs in a layoutlist widget? Why is the same bug not triggered if I just toggle through the layouts until I am at the end of the layout list?
Put differently: A layoutlist widget only cares about the current layout (or the list of available layouts, if it presents all of them in some way). I don't see how the order is relevant here.

Elv13: There is some cases:

 * You use `awful.layout.layouts` like the default config. Then there is no problems.
 * You use `awful.layout.layouts` but have code somehow setting layouts not part of it. Then the layoutlist and `mod4+space` based on it can't rotate anymore because the current layout is not part of the list.
 * You use `t.layouts` and your current tag layout is part of it: there is no problem.
 * You use `t.layouts` but have a different `t.layout`: see above
 * You have `t.layouts` but also have a `awful.widget.layoutlist` which uses `awful.layout.layouts`. Then the list mismatches and again you can't rotate properly unless you keep adding each layout from `awful.layout.layouts` when you rotate.
 * You use layouts objects, then you lose the state when rotating, but it wont happen because `awful.widget.layoutlist` holds a reference so it wont get deleted until the popup/widget is gone. So `t.layouts` will then be able to GC itself later when the state isn't "locked" anymore.

> Huh? What would it leak? If it leaks a layout, then I still have to ask again: "why is it okay when the list of available layouts on a tag changes randomly" (where "randomly" depends on the behaviour of the GC)

Assuming a bunch of things, including it relates closely to the `layoutlist` not GC-ing things currently displayed, then I think it's close enough to "be ok". I mean, `t.layouts` wont change while you use it because the result will "lock" the objects with non-weak references. There is the case that if you close the layoutlist then open it again very fast it will still have "too many things" compared to waiting a couple seconds. However is it really a problem in practice?

As for "what if you have a global list in the wibar and rotate it, then the temporary entries wont get removed" corner case, the `layoutlist` PR adds a `:cleanup()` function you can connect to the tag change signal.

> How does the order of layouts fix bugs in a layoutlist widget?

* Rotating the list is important (that's what the default config does), the corrent layout need to be part of that rotation even if it isn't in the list.
* It looks weird if there is no selected item, which is the case if the list and current layout are not compatible
* There can be many lists at once (`awful.layout.layouts`, `t.layouts`, `whatever_the_user_come_up_with`) and they need to be rotated individually.

Elv13:
Changes: *emit `property::layouts` when the layout changes (to update the `layoutlist`)*

lib/awful/tag.lua
```diff
@@ -746,9 +763,47 @@ function tag.object.set_layout(t, layout)
...
+        tag.setproperty(self, "custom_layouts", cls)
+    end
+
+    -- Required to get the default/fallback list of layouts
+    alayout = alayout or require("awful.layout")
+
+    -- Add the custom and unlisted layouts to te front
+    if (not gtable.hasitem(cls, self.layout))
+      and not alayout.get_tag_layout_index(self) then
+        table.insert(cls, self.layout)
```
psychon: Which piece of code is responsible for removing this item again?
(And the answer "the garbage collector" still means: Why is it okay for the list of layouts to change randomly? (E.g.: No signal is emitted))

psychon:
Oh and:

> * You use awful.layout.layouts like the default config. Then there is no problems.
> * You use awful.layout.layouts but have code somehow setting layouts not part of it. Then the layoutlist and mod4+space based on it can't rotate anymore because the current layout is not part of the list.

As I see it, when I switch away from a layout that is not in any list, well, then that layout is gone. That's what "not being in the layout list" means.

> You use layouts objects, then you lose the state when rotating, 

Then whatever created that layout should not just set it as a layout of some tag, but also add it some list of layouts. But this could again raise the question of "who is responsible for removing the layout again?" and this is still something that is not the GC's responsibility.

I would prefer if the GC's only responsibility would be "make sure awesome does not eat all the memory". Everything else should not be the GC's job (although I bet we are already using the GC for some things where we should not).

Elv13:
Updated to be more complex, but not rely on dubious GC tricks to make the `awful.widget.layoutlist` not crash. (the build failure is unrelated, it's that spawn thing I "broke" last week)

I also added tests

lib/awful/tag.lua
```diff
@@ -74,6 +75,66 @@ local function raw_tags(scr)
...
+
+    -- Only add to the custom_layouts and preserve the ability to globally
+    -- set the layouts.
+    if override and not pos then
+        table.insert(override, to)
+        self:emit_signal("property::layouts")
+        return
+    end
+
+    pos = from and gtable.hasitem(alayout.layouts, from) or nil
```
psychon: This line is redundant and can be removed, right?

lib/awful/tag.lua
```diff
@@ -74,6 +75,66 @@ local function raw_tags(scr)
...
+
+    -- The new layout is part of the global layouts. Fork the list.
+    if pos and from ~= to then
```
psychon: How about adding `if from == to then return end` at the beginning of this function and getting rid of all the `from ~= to` cases?

lib/awful/tag.lua
```diff
@@ -74,6 +75,66 @@ local function raw_tags(scr)
...
+        local cloned = gtable.clone(alayout.layouts, false)
+        table.insert(cloned, to)
```
psychon: Shouldn't this be `cloned[pos] = to`?

lib/awful/tag.lua
```diff
@@ -74,6 +75,66 @@ local function raw_tags(scr)
...
+        self.layouts = cloned
+        return
+    end
+
+    if pos then return end
```
psychon: This is only reached if `from == to`, right? If so, this can be removed if the early return is added at the beginning of the function.

lib/awful/tag.lua
```diff
@@ -74,6 +75,66 @@ local function raw_tags(scr)
...
+
+    local cls = custom_layouts(self)
+
+    if gtable.hasitem(cls, to) then return end
```
psychon: Shouldn't this still look for `from` and replace it?

lib/awful/tag.lua
```diff
@@ -741,14 +821,47 @@ function tag.object.set_layout(t, layout)
...
+        template = layout
         layout = instance
     end
 
     tag.setproperty(t, "layout", layout)
 
+    local override = tag.getproperty(t, "layouts")
+
+    if not override then
+        t:emit_signal("property::layouts")
```
psychon: Uhm, why is the signal emitted here already? Why is it emitted at all? I fail to see how the list of layouts changes here (or at least, always changes here). Shouldn't the following `update_layouts` emit this signal if needed?

lib/awful/tag.lua
psychon: Actually, what is the supposed meaning of `update_layouts(t, a, a)`? I would expect that replacing `a` with `a` just doesn't do anything, but apparently the code is supposed to add `a` instead? I am confused since the comment in front of the function and its behaviour does not match.

lib/awful/tag.lua
```diff
@@ -741,14 +821,47 @@ function tag.object.set_layout(t, layout)
...
+    -- Without the clone, the custom_layouts would grow
+    return #cls > 0 and gtable.merge(gtable.clone(cls, false), alayout.layouts) or
+        alayout.layouts
+end
+
+function tag.object.set_layouts(self, layouts)
+    tag.setproperty(self, "custom_layouts", {})
+    tag.setproperty(self, "layouts", gtable.clone(layouts, false))
+    update_layouts(self, self.layout, self.layout)
+    self:emit_signal("property::layouts")
```
psychon: Doesn't `update_layouts` emit the signal?

lib/awful/tag.lua
Elv13: I don't think there is a valid way it could happen. `custom_layouts` is internal/private. I will add an `_` in front to make that more obvious.

Elv13:
Updated and added a new tests to catch the `table.insert` vs. `t[] =` mistake from the previous version. The old test passed because there was a second bug. `_custom_layouts` was not merged into `_layouts` when it was cloned. This caused the count to be correct, but the content wasn't.

Elv13:
bump?

Issue #2361, opened 2018-08-18
Occasional XCB connection errors on receiving Discord notification


zetaPRIME:
Output of `awesome --version`:

```awesome v4.2-407-g17018118-dirty (Human after all)
 • Compiled against Lua 5.1.4 (running with LuaJIT 2.0.5)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```
Running on Arch Linux with AUR package `awesome-luajit-git`; launched via LightDM and used concurrently with Compton.

**How to reproduce the issue:**

Recieve a DM, highlight or other notification from Discord, using Naughty as your notifier

**Actual result:**

Occasionally, seemingly at complete random, the screen goes black, followed by the Discord notification sound playing as the user is dumped back to the TTY before LightDM starts back up. No error indicated in `~/.xsession-errors` or `~/.xsession-errors.old`, except that the X server has suddenly gone missing. This only occurs with notifications from Discord, with no seeming pattern as to what about the notification causes the issue.

**Expected result:**

Notification displaying as usual.

actionless:
sometimes i have the same crashes because of rocketchat notifications

is discord also using electron?

zetaPRIME:
yep! one of two electron apps I use (other being Atom); think there's some relation?

psychon:
How about `/var/log/Xorg.0.log` and `Xorg.0.log.old`? Anything interesting in there? Anything in the kernel log? Anything in `sudo journalctl --since="5 minutes ago"` or `/var/log/messages` or something similar?

actionless:
i didn't found anything except for the message in awesome stderr what connection to X server was lost (however X server is still running after that), i'll save that trace next time i'll encounter it

psychon:
How did you determine that Xorg is still running even though awesome claims it lost its connection? Could it be that Xorg was just restarted?

actionless:
because all the applications are also remained running

i'm using this startup wrapper to avoid losing running apps if awesome crashes:
https://github.com/actionless/awesome_config/blob/devel/packaging/awesome_no_argb

psychon:
> except for the message in awesome stderr what connection to X server was lost 

What was the exact error message? The message should be something like `X server connection broke (error 42)`. The following error codes are possible:
- The other end closed the socket  / writing to the socket failed / reading from the socket failed / `select` or `poll` failed (error 1)
[...]

PR #2362, opened 2018-08-19
naughty.dbus: Duplicate surface data

psychon:
When an icon is sent over dbus, we turn this into a cairo ImageSurface.
This is done by turning the actual icon data into a string and using
cairo.ImageSurface.create_for_data() to create a surface for this data.
However, this function only creates an ImageSurface that refers to this
data. It does not copy the data. Thus, when the Lua GC later frees the
string, we have a cairo surface that refers to already-freed data.

Fix this by duplicating the cairo surface, which makes cairo create a
copy of the data. Then, we finish the original surface. While doing
this, the string is kept alive in a local variable.

(Possibly) Fixes: https://github.com/awesomeWM/awesome/issues/2361

### DIFF
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -16,6 +16,7 @@ local string = string
 local capi = { awesome = awesome,
                dbus = dbus }
 local gtable = require("gears.table")
+local gsurface = require("gears.surface")
 local cairo = require("lgi").cairo
 
 local schar = string.char
@@ -102,7 +103,14 @@ local function convert_icon(w, h, rowstride, channels, data)
         offset = offset + rowstride
     end
 
-    return cairo.ImageSurface.create_for_data(tcat(rows), format, w, h, stride)
+    local pixels = tcat(rows)
+    local surf = cairo.ImageSurface.create_for_data(pixels, format, w, h, stride)
+
+    -- The surface refers to 'pixels', which can be freed by the GC. Thus,
+    -- duplicate the surface to create a copy of the data owned by cairo.
+    local res = gsurface.duplicate_surface(surf)
+    surf:finish()
+    return res
 end
 
 capi.dbus.connect_signal("org.freedesktop.Notifications",


lib/naughty/dbus.lua
```diff
@@ -102,7 +103,14 @@ local function convert_icon(w, h, rowstride, channels, data)
         offset = offset + rowstride
     end
 
-    return cairo.ImageSurface.create_for_data(tcat(rows), format, w, h, stride)
+    local data = tcat(rows)
```
actionless: i think travis would complain what `data` is previously defined

actionless: yup, it did

psychon:
Updated with
```diff
diff --git a/lib/naughty/dbus.lua b/lib/naughty/dbus.lua
index f3ad61735..5934c85d8 100644
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -103,8 +103,8 @@ local function convert_icon(w, h, rowstride, channels, data)
         offset = offset + rowstride
     end
 
-    local data = tcat(rows)
-    local surf = cairo.ImageSurface.create_for_data(data, format, w, h, stride)
+    local pixels = tcat(rows)
+    local surf = cairo.ImageSurface.create_for_data(pixels, format, w, h, stride)
 
     -- The surface refers to 'data', which can be freed by the GC. Thus,
     -- duplicate the surface to create a copy of the data owned by cairo.
```
and then again with
```diff
diff --git a/lib/naughty/dbus.lua b/lib/naughty/dbus.lua
index 5934c85d8..bb9fcfc50 100644
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -106,7 +106,7 @@ local function convert_icon(w, h, rowstride, channels, data)
     local pixels = tcat(rows)
     local surf = cairo.ImageSurface.create_for_data(pixels, format, w, h, stride)
 
-    -- The surface refers to 'data', which can be freed by the GC. Thus,
+    -- The surface refers to 'pixels', which can be freed by the GC. Thus,
     -- duplicate the surface to create a copy of the data owned by cairo.
     local res = gsurface.duplicate_surface(surf)
     surf:finish()
```

PR #2363, opened 2018-08-20
More gravity-related changes

psychon:
Still none of the bugs reported by @sigprof are fixed, but here are some more gravity-related changes that hopefully simplify the code (and an unrelated bugfix is in here as well).

### DIFF
--- a/event.c
+++ b/event.c
@@ -329,6 +329,14 @@ event_handle_configurerequest(xcb_configure_request_event_t *ev)
 
     if((c = client_getbywin(ev->window)))
     {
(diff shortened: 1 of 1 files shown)


Elv13:
@psychon Is this ready to merge now or do you plan to do the "other half" of the gravity issues you mentioned in the first comment along with merging this?

psychon:
> or do you plan to do the "other half" of the gravity issues you mentioned in the first comment along with merging this?

I tried fixing those bugs a couple of times already and never got anywhere... :-(

PR #2369, opened 2018-08-22
Fix no_offscreen test

psychon:
Commit fec8d6aa8fe257f2d1 fixes awful.placement.no_offscreen to behave
like other placement functions. This means that the margins=40 argument
that this test used and that was previously was just ignored, now
started working. Thus, there are now 40 pixels less on each side of the
client in this test.


actionless:
@psychon since it's a test blocker and it's trivial i think it's already fine to merge

psychon:
I wanted at least one :+1: for this. Thanks.

Issue #2377, opened 2018-08-29
menubar.utils: no way to show all programs in menu anymore


Perlovka:
Hi, there is a regression, possibly introduced in https://github.com/awesomeWM/awesome/commit/8d34201ec3dd76ce9d8f1ede8d9e8df5f4395684#diff-1fdb7413b48a1da49e75ea079cf4c610L237

awesome v4.2-437-gd0afcaa3-dirty (Human after all)
 • Compiled against Lua 5.1.4 (running with LuaJIT 2.0.5)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.0

**How to reproduce the issue:**

Set menubar.utils.wm_name to "" and use menubar.menu_gen.generate() to generate menu.

**Actual result:**

Programs with OnlyShowIn option are not added to menu

**Expected result:**

Programs should be added with any value of OnlyShowIn option, if menubar.utils.wm_name is set to """

PR #2379, opened 2018-08-29
menubar.utils: Allow to add programs with OnlyShowIn option being set

Perlovka:
Bug: https://github.com/awesomeWM/awesome/issues/2377

Signed-off-by: Michael Perlov <perlovka@gmail.com>

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -350,17 +350,19 @@ function utils.parse_desktop_file(file)
         -- Only check these values is NoDisplay is true (or non-existent)
 
         -- Only show the program if there is no OnlyShowIn attribute
-        -- or if it contains wm_name
-        if program.OnlyShowIn then
-            program.show = false -- Assume false until found
-            for _, wm in ipairs(program.OnlyShowIn) do
-                if wm:match(utils.wm_name) then
-                    program.show = true
-                    break
+        -- or if it contains wm_name or wm_name is empty
+        if utils.wm_name ~= "" then
+            if program.OnlyShowIn then
+                program.show = false -- Assume false until found
+                for _, wm in ipairs(program.OnlyShowIn) do
+                    if wm == utils.wm_name then
+                        program.show = true
+                        break
+                    end
                 end
+            else
+                program.show = true
             end
-        else
-            program.show = true
         end
 
         -- Only need to check NotShowIn if the program is being shown
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -354,7 +354,7 @@ function utils.parse_desktop_file(file)
         if program.OnlyShowIn then
             program.show = false -- Assume false until found
             for _, wm in ipairs(program.OnlyShowIn) do
-                if wm == utils.wm_name then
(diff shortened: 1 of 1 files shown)


psychon:
Why do you think that Lua string match is the right matching operation here?
The specification only says:

> A list of strings identifying the desktop environments that should display/not display a given desktop entry.
> 
> By default, a desktop file should be shown, unless an OnlyShowIn key is present, in which case, the default is for the file not to be shown.
> 
> If $XDG_CURRENT_DESKTOP is set then it contains a colon-separated list of strings. In order, each string is considered. If a matching entry is found in OnlyShowIn then the desktop file is shown. If an entry is found in NotShowIn then the desktop file is not shown. If none of the strings match then the default action is taken (as above).
> 
> $XDG_CURRENT_DESKTOP should have been set by the login manager, according to the value of the DesktopNames found in the session file. The entry in the session file has multiple values separated in the usual way: with a semicolon.
> 
> The same desktop name may not appear in both OnlyShowIn and NotShowIn of a group.

https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s06.html

I agree that "if a matching entry is found" is ambiguous, but I doubt that this means "Lua's regex match".
(Note that we completely ignore this `$XDG_CURRENT_DESKTOP` dance, but whatever)

At best, I would allow to interpret the spec to say "both `$XDG_CURRENT_DESKTOP` and the value in the desktop file are a list of strings; a match is found if they have some common entry".

Perlovka:
The point is to allow user to override .desktop file setting. Let's say, I have installed XFCE and KDE and I want see their applications in awesome menu. Lua will return true on string.match("XFCE", ""), allowing to override .desktop file restriction. I just adopted current code to previous version, where match was used.

Perlovka:
Well, there is one more solution: check utils.wm_name for non-empty value before processing, since program.show is by default set to true:
```lua
    if utils.wm_name ~= "" then
        if program.OnlyShowIn then
            program.show = false -- Assume false until found
            for _, wm in ipairs(program.OnlyShowIn) do
                if wm == (utils.wm_name) then
                    program.show = true
                    break
                end
            end
        else
            program.show = true
        end
    end
```

Perlovka:
Added fix to check if wm_name is empty.

Elv13:
@actionless we could add `menubar.utils.ignore_onlyshowin` later and merge this?

blueyed:
@jd 
> mergify/pr Pending — Change requests need to be dismissed

with:

> actionless approved these changes See review ￼ Dismiss review
> Elv13 approved these changes See review ￼ Dismiss review
> blueyed was requested for review

jd:
@blueyed sorry I might be late to the party, now all I see is that `codecov/patch` is missing :)

blueyed:
@jd 
It still says:
> mergify/pr Pending — Change requests need to be dismissed

jd:
@blueyed This PR is old, I suspect this is a leftover. What's wrong is listed here:
https://github.com/awesomeWM/awesome/pull/2379/checks?check_run_id=43834990

codecov is missing.

PR #2383, opened 2018-09-07
Fix removing multibyte characters in prompt, fixes #974

### LINKED ISSUE #974: Prompt doesn't work properly with non ascii symbols
For utf-8 characters needs to press "backspace" more than once. 
Navigation by "left" and "right" keys doesn't showing current position, and after navigation "backspace" and "delete" keys doesn't delete anything.

necauqua:
This change does what i've said in #974, works well for me.


actionless:
`bit32` is available only in lua 5.2

in lua 5.3 you can do rshift with `>>`, but idk how to handle that for lua 5.1

necauqua:
Oh, i see
I know about operators, but `bit32` is still in 5.3 (which i have) so i thought this is good enough to be universally compatible.

I've just realized that you can just compare if byte is in range [128;192) instead, huh

actionless:
> but bit32 is still in 5.3 (which i have)

mb you got installed it from luarocks as dep for some other app?

necauqua:
luarocks said that it is installed and `provided by VM or rocks_provided`, idk what it means, but i guess there is really no deprecated bit32 module in Lua 5.3, yeah.
Anyway, fixed it with range check

actionless:
https://travis-ci.org/awesomeWM/awesome/jobs/425915324#L3404

necauqua:
There is always that one thing that I miss (multiple times), ugh

necauqua:
I've realized that there is also an unfixed delete
Just no cursor movement allowed (for me at least) in desktop icons prompt to notice
For delete you can also remove one byte and then remove continuation bytes in a loop, but utf-8 has the prefix code, so delete could work a bit quicker, i guess:

if byte < 128 => remove 1 byte
elseif byte < 192 => continuation byte, shoulnt get there normally, remove it(with a loop so no broken utf left)?
elseif byte < 224 => remove 2 bytes
elseif byte < 240 => remove 3 bytes
elseif byte < 248 => remove 4 bytes
else => definitely invalid utf

May i commit such a fix when i get back to my PC?

actionless:
idk, mb to add some helper to `gears` for working with utf8 strings?

necauqua:
Update:
- laptop broke, had to finally buy a good new one (and except that i am busy with university and so on)
- offset calculation function works, i only have some minor problems with surrogates
- in process of redoing that ugly if-elseif 300 or so lines in awful.prompt to be more abstract (e.g. there are duplicates and i upgraded to utf-8 only ctlr+backspace but there is still ctrl+w and so on).

And yeah, awful.prompt's cursor variable counts bytes and not codepoints (as it did) to make string.sub function useful (and not to redo awful.prompt completely) and cursor moves by calculated delta of bytes.

actionless:
@necauqua any update on this?

necauqua:
Uh, completely forgot about this, I'm sorry.
I've moved to full-time job in october and since I am also a student, I did no code for joy at home in like a month (not counting small and simple labworks for the university).
I'll try to finish this on weekend maybe

actionless:
no worries, take your time; i was just checking up for stalled PRs

Issue #1092, opened 2016-09-12
Port awful.placement.no_offscreen to the new awful.placement conventions


Elv13:
Those functions are older than the others and don't fully comply to conventions that make some composition scenarios work.

Under_mouse and no_offscreen also need to be unit tested.

psychon:
(Perhaps it would also be nice to add some comments (to `wrap_client`? Dunno, I don't get all the magic) that explain what the "conventions that make some composition scenarios work" actually are, for whoever wants to add the next placement functions.)

(Also, is the pretend-mode really necessary? I tried to look into this issue and it seems that due to pretend, `under_mouse` doesn't actually do anything and then `no_offscreen` works based on the old position of the client. Without pretend-mode, things could be easier. (And there is so much more (undocumented?) modes in there that seem to make stuff more complicated...))

Elv13:
> Also, is the pretend-mode really necessary?

Yes, changing the resolution during the composition mess up the pre-conditions and the end result changes. Try to disable it with a semi-complex composition, the output will change and wont be as predictable. It has side effects like this, but I don't think there is a way to remove it without breaking more use cases.

blueyed:
Is this the reason why `properties = { placement = (awful.placement.under_mouse + awful.placement.no_offscreen) }` does not work?

I've tried to use one of them as `callback`, but it seems they cannot be combined currently?!

blueyed:
It should work according to the FAQ (https://github.com/awesomeWM/awesome/blob/8f6e80f5c6870cb47305689e31854d795664f03e/docs/90-FAQ.md#how-to-make-windows-spawn-under-the-mouse-cursor), but does not:
```
    -- Open CopyQ window at/under mouse.
    { rule = {
      instance = 'copyq',
      type = 'normal',
      name = ' - CopyQ',
      }, properties = {
        placement = awful.placement.under_mouse+awful.placement.no_overlap+awful.placement.no_offscreen }},
```

psychon:
Removed the milestone since it is easily visible that it is meaningless here

sigprof:
After `awful.placement.no_overlap` was ported to the new conventions, the default configuration (which contains `placement = awful.placement.no_overlap+awful.placement.no_offscreen`) became broken — all new windows are placed at the top left corner of the workspace area. The change from https://github.com/actionless/awesome/commit/7b1b7d415af7b7d67171958f9d3a04e9ef34dc1b fixes the problem.

[...]

PR #2033, opened 2017-09-09
fix(awful: placement: no_offsceen): use new placement infrastructure

actionless:
first time awesome was crashing during the travis, hope this time it should be fine

fixes https://github.com/awesomeWM/awesome/issues/1092

PR #2386, opened 2018-09-11
awful.placement: Fix no_offscreen when composed with other functions

sigprof:
The `awful.placement.no_offscreen` function did not work properly when composed with other placement functions; in particular, the default configuration (`awful.placement.no_overlap+awful.placement.no_offscreen`) was broken.  The compose function sets `args.pretend=true` and puts the result of the previous placement function into `args.override_geometry`
before calling the next placement function, but `no_offscreen` did not use `args.override_geometry`, therefore the result of the previous placement function was discarded.

Fixes: https://github.com/awesomeWM/awesome/issues/1092 (the previous attempt (https://github.com/awesomeWM/awesome/pull/2033) did not fix the problem completely),

### DIFF
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -851,7 +851,7 @@ function placement.no_offscreen(c, args)
 
     c = c or capi.client.focus
     args = add_context(args, "no_offscreen")
-    local geometry = area_common(c)
+    local geometry = area_common(c, nil, false, args)
     local screen = get_screen(args.screen or c.screen or a_screen.getbycoord(geometry.x, geometry.y))
     local screen_geometry = screen.workarea
 


sigprof:
Oops, looks like I ran my tests in a wrong directory, now I see that they fail… need to find out how to fix that properly.

psychon:
Hm... two tests failed. `test-gravity` has lots of random changes (some things are suddenly two pixels wider, positions are different etc).
The other test that failed is `test-awful-rules.lua`, which basically complains about the same thing:
```lua
-- Test callback and floating
test_rule {
    properties = { floating = true, tag = "does_not_exist" },
    callback   = function(c)
        assert(type(c) == "client")
        callback_called = true
    end,
    test = function(class)
        -- Test if callbacks works
        assert(callback_called)

        -- Test that the "does_not_exist"-tag caused an error
        assert(message_printed)
        gears.debug.print_error = orig_error

        -- Make sure "smart" dynamic properties are applied
        assert(get_client_by_class(class).floating)

        -- The size should not have changed
        local geo = get_client_by_class(class):geometry()
        assert(geo.width == 100 and geo.height == 100+tb_height)

        return true
    end
}
```
The assertion about the size at the end here failed?!?

sigprof:
Looks like even `no_overlap` without `no_offscreen` is broken now — while trying to make some tests, I try to create a 100×100 window and get 102×119 instead of 100×117 from `c:geometry()` (this includes the title). The default `no_overlap+no_offscreen` hides the bug, because the buggy `no_offscreen` drops the result of buggy `no_overlap`.

actionless:
and if reverting this locally fixing the tests or your usecase?
https://github.com/awesomeWM/awesome/pull/2355

sigprof:
Found the bug in `no_overlap`. Please ignore this pull request; I'll make a proper one with appropriate tests tomorrow.

PR #2388, opened 2018-09-12
Fix awful.placement.no_overlap + awful.placement.no_offscreen

sigprof:
Fix multiple problems in `awful.placement.no_overlap` and `awful.placement.no_offscreen` related to their conversion to the new placement infrastructure:

- `no_offscreen` did not work properly when composed with other placement functions — it used the real client geometry instead of the geometry returned by the previous placement function; this usually resulted in all windows being placed at the top left corner of the workarea.
- When getting the client geometry, `no_offscreen` did not adjust it for margins specified in arguments, but the reverse adjustment was performed before returning the result; because of this, after adding margins the client size was incorrectly decreased by the values of those margins.
- `no_overlap` did not perform the reverse adjustment for the border width before returning the result, therefore after applying `no_overlap` the client width and height were increased by `2*border_width`. This error was not noticeable in the default configuration (`no_overlap + no_offscreen`), because the bug in `no_overlap` was masked by the above bug in `no_offscreen` (the result of `no_overlap` was discarded).

Actual fixes are 1 added line in `no_overlap` and 1 changed line in `no_offscreen`; the rest of the pull request contains a test for `no_overlap + no_offscreen` which should hopefully catch similar problems in the future.

### DIFF
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -851,7 +851,7 @@ function placement.no_offscreen(c, args)
 
     c = c or capi.client.focus
     args = add_context(args, "no_offscreen")
-    local geometry = area_common(c)
+    local geometry = geometry_common(c, args)
     local screen = get_screen(args.screen or c.screen or a_screen.getbycoord(geometry.x, geometry.y))
     local screen_geometry = screen.workarea
 
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -941,6 +941,7 @@ function placement.no_overlap(c, args)
     new.width = geometry.width
     new.height = geometry.height
 
+    remove_border(c, args, new)
     geometry_common(c, args, new)
     return fix_new_geometry(new, args, true)
 end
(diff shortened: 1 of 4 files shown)


sigprof:
Looks like Travis CI shows failed status only due to failed tests in intermediate commits (I committed the new test before the fixes to demonstrate that the test actually catches those bugs).

psychon:
Thanks for digging through this & special thanks for writing a new test

Elv13:
Thanks a lot for this. Even as the author of the code, everytime I have to fix it I get lost in my own abstractions... This must have been frustrating, sorry about that.

Obviously, it's a +1000 from me, those tests will come handy. How do you want this merged? Do you want the commits to be squashed or do you want to reorder them? We wont merge it as-is since it would break the automatic git-bisect (well, it would not be the first time, but we still try to have each commits pass the tests).

sigprof:
I reordered commits so that the `no_overlap` fix is before the `no_offscreen` fix (fixing `no_offscreen` exposes the bug in `no_overlap`, which breaks some tests), and the new test now comes after all fixes.

Elv13:
Thanks

actionless:
thanks a lot!

PR #2391, opened 2018-09-15
utils.lua: make it compatible with more terminals

ThanosApostolou:
Terminal applications don't work with many terminals when they have many options (for example in manjaro we use manjaro architect with `Exec = sudo /usr/bin/setup`).

For xfce4-terminal, mate-terminal, gnome-terminal it is better to use the `-x` option which interprets the rest parameters as command parameters. Termite has only `-e` option but it needs it to be in quotes in order to understand many parameters (I don't know if it gets messed up if the Exec command includes quotes on its own like for example `Exec=sh -c "something"` or how to make sure it doesn't). Terminals like xterm and lxterminal work with just the `-e` option.

I use **string.find()** instead of just **==** in case someone sets terminal with the full path like `menubar.utils.terminal = /usr/bin/xfce4-terminal`.

Adding special cases for each terminal seems like a bad solution but I cannot think of anything better.

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -399,13 +399,8 @@ function utils.parse_desktop_file(file)
             cmdline = cmdline:gsub('%%i', '')
         end
         if program.Terminal == true then
-            if string.find(utils.terminal, "xfce4%-terminal") or string.find(utils.terminal, "gnome%-terminal") or string.find(utils.terminal, "mate%-terminal") then
-                cmdline = utils.terminal .. ' -x ' .. cmdline
-            elseif string.find(utils.terminal, "termite") then
-                cmdline = utils.terminal .. ' -e \"' .. cmdline .. '\"'
-            else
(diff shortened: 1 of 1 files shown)


Elv13:
I am not convinced hardcoding the terminal names in the code is such a good idea. I didn't try and have neither installed, but isn't always adding the proper quotes always enough? If it isn't then why is this not a bug in the terminal and needs to be mitigated in Awesome?

ThanosApostolou:
Yeah my first solution was just to wrap the whole cmdline in quotes (like I've done now only for termite). However, I tested it and it is problematic if the application's Exec line already contains quotes. For example playonlinux/wine create .desktop entries which contain `WINEPREFIX="PATH"`. If we have the whole command in quotes then this will break.

I didn't look at the code very carefully, but maybe there is an easy way to read the Exec line and replace `"` with `\"` and then wrap the whole command in quotes?

ThanosApostolou:
I close it in order to do a new clean PR

Issue #2392, opened 2018-09-15
Make menubar.utils.terminal compatible with different terminals


ThanosApostolou:
Output of `awesome --version`:
```
awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.
```

**How to reproduce the issue:**

Menubar cannot launch applications which have `Terminal=true` when it's used with xfce4-terminal, mate-terminal, gnome-terminal, termite. The problematic line is
https://github.com/awesomeWM/awesome/blob/68e4dd430f3c5a356afbd3d0cf4b94ac2bf64f4f/lib/menubar/utils.lua#L402
which takes for granted that terminals use the `-e` option to execute a command. However, in most terminals the command needs to be wrapped in quotes or even better it's preferable to use the `-x` option which passes all the rest arguments into the command. If I try to wrap the cmdline with quotes like this:
```
cmdline = utils.terminal .. ' -e "' .. cmdline:gsub('"', '\\"') .. '"'
```
then xfce4-terminal, mate-terminal, gnome-terminal and termite work but it breaks it for applications like lxterminal, st, urxvt.

**Actual result:**

Menubar should work with many terminal options

**Expected result:**

Menubar only works with a few terminal options

After a lot of thinking, I believe that the best solution is to add a FIeld at menubar.utils Module like **menubar.utils.terminal_execute_option** which will be set to `-e` by default but users will be able to change it in `-x` and then menubar will execute the terminal argument according to this option. What do you think?

ThanosApostolou:
@Elv13 you were right that having hardcoded terminal options was bad. Tell me what you think about my suggestion.

PR #2393, opened 2018-09-15
add utils.terminal_execute_option field

ThanosApostolou:
Solution to #2392

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -35,7 +35,8 @@ local utils = {}
 utils.terminal = 'xterm'
 
 --- Terminal argument which indicates the execution of command
---  Usually something like ' -e ', ' -x ', ' -- '
+-- Usually something like ' -e ', ' -x ', ' -- '
+-- @tfield string menubar.utils.terminal_execute_option
 utils.terminal_execute_option = ' -e '
 
 --- The default icon for applications that don't provide any icon in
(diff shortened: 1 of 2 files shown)


lib/menubar/utils.lua
```diff
@@ -34,6 +34,10 @@ local utils = {}
 --- Terminal which applications that need terminal would open in.
 utils.terminal = 'xterm'
 
+--- Terminal argument which indicates the execution of command
+--  Usually something like ' -e ', ' -x ', ' -- '
+utils.terminal_execute_option = ' -e '
```
Elv13: Please add

    @tfeild string menubar.utils.terminal_execute_option

ThanosApostolou:
Is it now ok? Btw I think I pinged someone by mistake with that commit message :laughing:

actionless:
ahah, i think we'll end up defining it as a function like `terminal_handler` :D

ThanosApostolou:
I had spaces in the **utils.terminal_execute_option** at first because some terminals has options like `--command=` without a space afterwards, but usually this is an alias for an option like `-e` so I guess spaces in the menubar.utils field is just confusing.

I don't know how to change the default rc.lua because for example
https://github.com/awesomeWM/awesome/blob/68e4dd430f3c5a356afbd3d0cf4b94ac2bf64f4f/awesomerc.lua#L57
is set before
https://github.com/awesomeWM/awesome/blob/68e4dd430f3c5a356afbd3d0cf4b94ac2bf64f4f/awesomerc.lua#L123.
I could add another variable **terminal_execute_option** just below **terminal** and then add
**menubar.utils.terminal_execute_option=terminal_execute_option** just below line **menubar.utils.terminal = terminal**. How are you suggesting doing it?

Issue #2397, opened 2018-09-23
Naughty - colors


kajzersoze:
Is there a way to have different colors for title and the body of the text ?

r3lgar:
For text in the `text`, you can use `<span color='#RRGGBB'>your text</span>`. In the `title` you can't use pango markup (FIXME).

actionless:
use `&amp` instead of `&`

psychon:
There is `gears.string.xml_escape` that handles escaping for you

kajzersoze:
Having some problems with it, returns nothing it seems.
Any help how to use it in scenario below ?

```
function naughty.config.notify_callback(args)
		    args.text = gears.string.escape(args.text)
		    args.text = lain.util.markup.fg.color("#ffdab9",args.text)
		    return args
end
```

psychon:
Try `xml_escape` instead of the non-existing `escape`.
(And note that this was `awful.util.escape` in v4.0 and only moved to `gears.string.xml_escape` in v4.1, so depending on your awesome version, you might need the other function)

psychon:
Whoops. Thanks for the pointer! #2401 deals with that.

kajzersoze:
I don't know why it was decided that the title cannot contain markup.
Escaping strings had some problems in some cases, so I decided to take a look at `/lib/naughty/core.lua` and make it so that I can use markup on the title.
It works fine, title texts are smaller and there's a very little chance to have in them chars that need escaping, compared to the text of the notification.
Anyway, with this patch I'm not using `xml_escape`, instead of args.text I'm using args.title to change the fg color.

```
--- /lib/naughty/core.lua	2018-09-27 09:59:34.134200068 +0200
+++ /lib/naughty/core.lua	2018-09-27 10:05:54.852777572 +0200
@@ -464,7 +464,7 @@
     local textbox = notification.textbox
 
     local function setMarkup(pattern, replacements)
-        return textbox:set_markup_silently(string.format('<b>%s</b>%s', title, text:gsub(pattern, replacements)))
+        return textbox:set_markup_silently(string.format('<b>%s</b>%s', title:gsub(pattern, replacements), text:gsub(pattern, replacements)))
     end
     local function setText()
         textbox:set_text(string.format('%s %s', title, text))
@@ -472,7 +472,7 @@
 
     -- Since the title cannot contain markup, it must be escaped first so that
     -- it is not interpreted by Pango later.
-    title = title:gsub(escape_pattern, escape_subs)
[...]

PR #2401, opened 2018-09-26
Fix incorrect @name annotation in gears.string

psychon:
gears.string.xml_escape was annotated with "@name escape" and
gears.string.xml_unescape was annotated with "@name unescape".

Reference: https://github.com/awesomeWM/awesome/issues/2397#issuecomment-424630100

### DIFF
--- a/lib/gears/string.lua
+++ b/lib/gears/string.lua
@@ -10,7 +10,7 @@ local xml_entity_names = { ["'"] = "&apos;", ["\""] = "&quot;", ["<"] = "&lt;", 
 --- Escape a string from XML char.
 -- Useful to set raw text in textbox.
 -- @class function
--- @name escape
+-- @name xml_escape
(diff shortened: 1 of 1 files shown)


blueyed:
FYI: I've removed "codecov/patch" and "Codacy/PR Quality Review" from required statuses in the branch protection.

blueyed:
Should we remove Codacy altogether maybe (seems like mergify is waiting for it anyway).

actionless:
@blueyed i am up for codacy removal from required

blueyed:
@actionless 
Done already.  I've meant to remove it altogether.  But it ended up reporting after all / my comment.
Let's keep an eye on it.

actionless:
@blueyed codecov/patch is still required: https://github.com/awesomeWM/awesome/milestone/1

PR #2409, opened 2018-10-01
A better run_or_raise/spawn.once/singleton API

Elv13:
Following a [stackoverflow](https://stackoverflow.com/questions/52523256/auto-start-programs-on-specific-screen-tags) question, I took a look at how to build a more reliable way to handle this use case. Previously, we had a wiki page with a bunch of equally unreliable solutions the user could choose to avoid each other issues. We also have `awful.client.run_or_raise` which is often useless because it doesn't integrate well with the `awful.rules`.

This commit implement the proposed solution from StackOverflow. I admit it has so far limited testing but seem to work as intended for `inkscape` and `urxvt`. I post this more as a request for comment than "lets merge this as-is". It has a bunch of limitations like weak hashing and unreliable use of startup notifications.

What could change, but I am not sure makes thing any better:

 * Get the startup clients before "manage" instead of using a `delayed_call`
 * Also use the PID and a 30 second timer to prevent re-use issues to mitigate the lack of startup_id.
 * Use the Lua `debug` API to get more information into the hash using the stack and environment 
 * Use a better hashing algorithm

They would add hacks and make the code more complex for negligible gain in reliability. The code could be cleaned up a bit more, but right now I am mostly interested in opinions.

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -1140,6 +1140,7 @@ end
 --   If it is a function, it will be called with the client as argument.
 -- @see awful.spawn.once
 -- @see awful.spawn.single_instance
+-- @see awful.spawn.raise_or_spawn
 --
 -- @deprecated awful.client.run_or_raise
 -- @usage -- run or raise urxvt (perhaps, with tabs) on modkey + semicolon
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -8,7 +8,7 @@
 
 -- Grab environment we need
 local gdebug = require("gears.debug")
-local spawn = require("awful.spawn")
+local spawn = nil --TODO v5 deprecate
 local set_shape = require("awful.client.shape").update.all
 local object = require("gears.object")
 local grect = require("gears.geometry").rectangle
@@ -1138,8 +1138,10 @@ end
 -- @tparam bool|function merge If true then merge tags (select the client's
 --   first tag additionally) when the client is not visible.
 --   If it is a function, it will be called with the client as argument.
+-- @see awful.spawn.once
+-- @see awful.spawn.single_instance
 --
--- @function awful.client.run_or_raise
+-- @deprecated awful.client.run_or_raise
 -- @usage -- run or raise urxvt (perhaps, with tabs) on modkey + semicolon
 -- awful.key({ modkey, }, 'semicolon', function ()
 --     local matcher = function (c)
@@ -1148,6 +1150,10 @@ end
 --     awful.client.run_or_raise('urxvt', matcher)
 -- end);
 function client.run_or_raise(cmd, matcher, merge)
+    gdebug.deprecate("Use awful.spawn.single_instance instead of"..
+        "awful.client.run_or_raise", {deprecated_in=5})
+
+    spawn = spawn or require("awful.spawn")
     local clients = capi.client.get()
     local findex  = gtable.hasitem(clients, capi.client.focus) or 1
     local start   = gmath.cycle(#clients, findex + 1)
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -165,7 +165,9 @@ local lgi = require("lgi")
 local Gio = lgi.Gio
 local GLib = lgi.GLib
 local util   = require("awful.util")
-local timer  = require("gears.timer")
+local gtable = require("gears.table")
+local gtimer = require("gears.timer")
+local aclient = require("awful.client")
 local protected_call = require("gears.protected_call")
 
 local spawn = {}
@@ -210,6 +212,39 @@ do
     end
 end
 
+local function hash_command(cmd, rules)
+    rules = rules or {}
+    cmd = type(cmd) == "string" and cmd or table.concat(cmd, ';')
+
+    -- Do its best at adding some entropy
+    local concat_rules = nil
+    concat_rules = function (r, depth)
+        if depth > 2 then return end
+
+        local keys = gtable.keys(rules)
+
+        for _, k in ipairs(keys) do
+            local v = r[k]
+            local t = type(v)
+
+            if t == "string" or t == "number" then
+                cmd = cmd..k..v
+            elseif t == "tag" then
+                cmd = cmd..k..v.name
+            elseif t == "table" and not t.connect_signal then
+                cmd = cmd .. k
+                concat_rules(v, depth + 1)
+            end
+        end
+    end
+
+    concat_rules(rules, 1)
+
+    local glibstr = GLib.String(cmd)
+
+    return string.format('%x', math.ceil(GLib.String.hash(glibstr)))
+end
+
 spawn.snid_buffer = {}
 
 function spawn.on_snid_callback(c)
@@ -220,7 +255,7 @@ function spawn.on_snid_callback(c)
         --TODO v5: Remove this signal
         c:emit_signal("spawn::completed_with_payload", props, callback)
 
-        timer.delayed_call(function()
+        gtimer.delayed_call(function()
             spawn.snid_buffer[c.startup_id] = nil
         end)
     end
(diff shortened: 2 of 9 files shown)


lib/awful/client.lua
```diff
@@ -8,7 +8,7 @@
 
 -- Grab environment we need
 local gdebug = require("gears.debug")
-local spawn = require("awful.spawn")
+local spawn = nil --TODO v5 deprecate
```
psychon: Could we pretty-please stop adding more cyclic dependencies to things? :-(

Elv13: Well, that particular one is less evil than usual. It remove the "duty" of spawning things from `awful.client` and concentrate them into `awful.spawn`. Conceptually, I am making things cleaner. But of course, for API compatibility, a temporary hack is required. Where "temporary" means a rather long time.

psychon:
So, `run_once` is "once per X11 server reset"? How about doing `local list = {} function run_once(command) if list[command] then return end list[command] = true spawn(command) [serialize list via awesome.set_xproperty to the root window?] end`?

For `single_instance`, I'd just always require a `matcher` from the user of the function and be done (which can be either a function or a string, and a string is interpreted as the class or instance).
That's more work than relying on `startup_id` as your patch does, but also more reliable since too much stuff that people apparently want to spawn does not support SNID.

Elv13:
> For single_instance, I'd just always require a matcher from the user of the function and be done 

Then again, that's what `awful.client.run_or_raise` is and nobody seem to use it. While not *incompatible* with `awful.rules`, it doesn't mix well with them. And by that logic, since `awful.spawn` relies on the `startup_id` to work integrate with rules, a solution that involves `awful.spawn` with a matcher already depends on `startup_id` working. If course the matcher could be refactored to be used in the `awful.rules` source and it would improve the situation a little, but that would not solve the part that nobody seem to want to use `awful.client.run_or_raise` and many configs on github have some version of custom `run_or_raise`.

From that point of view and as the `matcher` is supported anyway, I don't see why the `startup_id` code should be removed. However you have a point that supporting the `matcher` in the rule source (probably along with a PID?) would improve reliability. I will do that in the next version.

psychon:
> and many configs on github have some version of custom run_or_raise.

You won't solve a cargo copy cult by adding better tools to awesome. People will keep looking at other people's config or outdated wiki pages and copy the solution from there. :-(

> The "spirit" of use case of the stack overflow question is to attach a single instance of an app to a specific tag across restart.

Hm, so this really needs rules to be re-applied after a restart...?
Somehow I am a bit unclear about what is supposed to be possible. It doesn't seem to be the simple "just ever run this once" that I had in mind.

If I remember correctly, the stackoverflow question was best solved via an `awful.rules`-rule directly.

Elv13:
The way I see the potential use case is this:

```lua

-- tags 1 to 9 exists

awful.spawn.once("urxvt", {tag = screen[1].tags[2]}
awful.spawn.once("urxvt", {tag = screen[1].tags[3]}
awful.spawn.once("urxvt", {tag = screen[1].tags[4]}

```

Then do random things in those terminals that cannot be known ahead of time by a `matcher` function. This assume the terminals are used for that specific task.

Then restart awesome and hope the terminal will still be in the right tag even if an `awful.rules` move all terminals to tags[9] for whatever reason. This also assume the said terminal has no way to redefine the class or instance or whatever.

For that specific use case, the whole point is being able to run_or_raise/spawn_once across restart with "random" clients properties that are not preserved across a restart.

I think it is a valid use case. I have some tags for some projects I am working on. Awesome has a Xephyr windows tiled with 2 terminals (one for the logs and one for git/make) and the text editor. Other are for my various IoT and Qt projects and also have terminals. I use the instance name hack and `awful.rules` (generated by Tyrannical) to place them. But that's a hack, it would be better if I could just spawn terminals in their position and get them back after a restart "the easy way". So this PR add this "easy way". It is reliable for apps that support SNID or when that other hack is applied.

psychon:
The assumption is that both the pre-restart and the post-restart awesome had those lines in their config, right?

lib/awful/spawn.lua
```diff
@@ -435,6 +470,149 @@ function spawn.read_lines(input_stream, line_callback, done_callback, close)
...
+-- The `rules` are standard `awful.rules`.
+--
+-- This function depends on the startup notification protocol to be correctly
+-- implemented by the command. See `client.startup_id` for more information.
+-- Note that this also wont work with shell or terminal commands.
+--
+-- @tparam string|table cmd The command.
+-- @tparam table rules The properties that need to be applied to the client.
+-- @tparam[opt] string uid A string to identify the client so it isn't executed
+--  multiple time.
```
psychon: I'd make it more explicit that `uid` is `unique id`. At first I was thinking of `user id`.

lib/awful/spawn.lua
```diff
@@ -435,6 +470,149 @@ function spawn.read_lines(input_stream, line_callback, done_callback, close)
...
+    end
+
+    -- The clients are not yet managed during the first execution, so it will
+    -- miss existing instances.
+    if awesome.startup then
+        gtimer.delayed_call(do_it)
+    else
+        do_it()
+    end
+end
```
psychon: After taking a closer look: Both spawning functions are basically the same. The only difference is in checking `status.exec`. It should be possible to merge them so that both functions just call the same helper function with an extra boolean parameter.

lib/awful/spawn.lua
```diff
@@ -210,6 +212,39 @@ do
     end
 end
 
+local function hash_command(cmd, rules)
+    rules = rules or {}
+    cmd = type(cmd) == "string" and cmd or table.concat(cmd)
```
psychon: I'd suggest `table.concat(cmd, " ; ")`, but I guess it does not matter much...

lib/awful/spawn.lua
```diff
@@ -210,6 +212,39 @@ do
...
+            local v = r[k]
+            local t = type(v)
+
+            if t == "string" or t == "number" then
+                cmd = cmd..k..v
+            elseif t == "tag" then
+                cmd = cmd..k..v.name
+            elseif t == "table" and not t.connect_signal then
+                cmd = cmd .. k
+                concat_rules(v, max_depth + 1)
```
psychon: Uhm, this recursion does not do anything. `max_depth` (which should just be called `depth` since it is the current depth, not the maximum one) starts out at 1. In the first recursive call, it ends up as `2`, which means that `concat_rules` immediately returns and does nothing.

Elv13:
> The assumption is that both the pre-restart and the post-restart awesome had those lines in their config, right?

This assumes it is either:

 * Directly in the `rc.lua` "root' environemnt (aka: not in a function)
 * In a `mytag:connect_signal("property::selected", function(t) if t.selected then awful.spawn.once(...) end end)`
 * In a keybinding
 * In a prompt that launch things only once or raise them (sounds useless, but why not)

So this assumes mostly nothing beside the command and the rules table being consistent enough to generate the same hash.

So the use case is based on the stackoverflow title "Auto start programs on specific screen tags [or raise if it exist]". While some instance where the client properties are deterministic enough to use `awful.rules` could already be handled, this PR tries to solve the generalization. It try to solve "I have a specific client instance for which I want to attach those specific properties and I want this to work even if I restart Awesome". It can't store the properties like the tag instance in a serialized way, but assuming the `awful.spawn.once` command is the same, it's irrelevant because it will work anyway.

psychon:
Okay. Basically :+1:

Elv13:
Version 2:

 * Add `awful.spawn.raise_or_spawn` based on `single_instance`
 * Rename `awful.rules` `switchtotag` to `switch_to_tags` (with deprecation)
 * Fix a "bug" in `gears.table` where the order or `gears.table.join` wasn't defined for no valid reason
 * Add an optional maximum number of tags to select to `awful.tag.viewmore`
 * Add `switch_to_tag` and `switch_to_tags` to `request::activate`
 * Use the `matcher` in the `awful.rules` source
 * Track the PID along with the matcher to improve reliability
 * Rename `uid` to `unique_id`
 * Fix the recursive hash issue
 * Add ';' to the table concatenation

So this adds the missing feature to bring parity with `run_or_raise` while using the normal rules and `request::` pipeline. If there is no objection, v3 will add tests and it remove the `[RFC]` in the hope of getting merged.

lib/awful/spawn.lua
```diff
@@ -435,6 +470,155 @@ function spawn.read_lines(input_stream, line_callback, done_callback, close)
...
+
+    spawn.single_instance_manager.by_uid[hash] = status
+
+    return status
+end
+
+local function run_once_common(hash, cmd, keep_pid)
+    local pid, snid = spawn.spawn(cmd)
+
+    if type(pid) == "string" or not snid then return false end
```
psychon: Why not `return pid, snid`? That would "hand out" error messages

Elv13:
I fixed the issues and added some tests. I guess I could add more, but this PR now passed 0.5kLoC.

Elv13:
@blueyed Mind reviewing this one? It concerns a couple things we discussed before.

lib/awful/spawn.lua
```diff
@@ -435,6 +470,159 @@ function spawn.read_lines(input_stream, line_callback, done_callback, close)
...
+
+    local status = spawn.single_instance_manager.by_uid[hash]
+    status.exec = true
+
+    spawn.single_instance_manager.by_snid[snid] = status
+
+    if keep_pid then
+        spawn.single_instance_manager.by_pid[pid] = status
+
+        -- Prevent issues related to PID being re-used and a memory leak
```
psychon: Apropos memory leak: But the `by_snid` table is never cleared?

tests/test-spawn.lua
```diff
@@ -10,6 +10,15 @@ local exit_yay, exit_snd = nil, nil
 -- * Using spawn with array is already covered by the test client.
 -- * spawn with startup notification is covered by test-spawn-snid.lua
 
+local tiny_client = function(class)
+    return [[lua -e "Gtk = require('lgi').require('Gtk');
+Gtk.init();window = Gtk.Window { title = 'Client'};
+window:set_wmclass(']]..class.."','".. class..[['); window:show_all();
+Gtk:main{...}"]]
```
psychon: This program does not ever exit. Especially: It keeps running when the window is closed.

psychon: Also, could this use `{ "lua", "-e", "..." }` instead of just providing a string that needs to be split up again?

Elv13:
Done

Elv13:
Updated. I forgot to remove a debug `print` from the last version.

@blueyed @actionless any comment on this pull request and API?

Elv13:
Last call. I would really appreciate a second review here, but if nobody is interested, I will merge it tomorrow.

PR #2412, opened 2018-10-03
themes/xresources/assets.lua: use raw=true with deprecation

blueyed:


actionless:
notorious `awesome: lvm.c:436: luaV_equalobj: Assertion `(((t1)->value_).gc)->tt == 5' failed.`

Issue #1971, opened 2017-08-10
gears.timer: add support for glib.timeout_add_seconds


blueyed:
glib 2.14 supports second-based timers: https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#g-timeout-add-seconds

> The grouping of timers to fire at the same time results in a more power and CPU efficient behavior so if your timer is in multiples of seconds and you don't require the first timer exactly one second from now, the use of g_timeout_add_seconds() is preferred over g_timeout_add().

We could add an argument to enable this and/or use it automatically when full seconds are passed in for the interval - although it also affects the start time.

psychon:
> glib 2.14

That's from 2007 and older than our minimum requirement (2.40) anyway: https://github.com/GNOME/glib/releases/tag/GLIB_2_14_0

PR #2413, opened 2018-10-03
[RFC] gears.timer: use timeout_add_seconds for integers

blueyed:
Used for timeout=1, but not timeout=1.0

Fixes https://github.com/awesomeWM/awesome/issues/1971.


lib/gears/timer.lua
```diff
@@ -67,6 +67,7 @@ local protected_call = require("gears.protected_call")
 -- to enable garbage collection.
 -- @tfield number timeout Interval in seconds to emit the timeout signal.
 --   Can be any value, including floating point ones (e.g. 1.5 seconds).
+--   It will use a seconds-based timeout for integer values.
```
blueyed: TODO: not really clear/complete

psychon:
- This seems a bit arbitrary to me. "Make sure your number happens to be integer, or else the timer will be imprecise"
- Before Lua 5.3, `1.0` and `1` are the same number, so your `Used for timeout=1, but not timeout=1.0` does not work at all (Aren't you using luajit? That should also mean that 1 and 1.0 are the same number)
- I do not understand that `tostring() % 1.0` magic at all.

blueyed:
Yeah, open for suggestions.

>  "Make sure your number happens to be integer, or else the timer will be imprecise"

Integers make it imprecise, you need to use a float.

`tostring(1.0)` is indeed `1` in Lua 5.1 (I've only tested it with 5.3).

It does not make sense though - I've meant to distinguish between using "1" and "1.0", but it should have been something like `tostring(1):match("%.")` then - but also does not work in 5.1.

So should we maybe only check if `timeout % 1 == 0`, and otherwise have a param to control it?
Or should there only be a param?

I think often (i.e. by default) second-based precision is good if you are using integers.

psychon:
How about adding a `precise` property to timers? I guess the default should be "precise" for now, but for v5 we can make this imprecise by default and break everyones timers:
(Untested)
```diff
diff --git a/lib/gears/timer.lua b/lib/gears/timer.lua
index c01b49f4..4435c348 100644
--- a/lib/gears/timer.lua
+++ b/lib/gears/timer.lua
@@ -61,6 +61,11 @@ local glib = require("lgi").GLib
 local object = require("gears.object")
 local protected_call = require("gears.protected_call")
 
+local precise_default = false
+if awesome.version < "v5" then
+    precise_default = true
+end
+
 --- Timer objects. This type of object is useful when triggering events repeatedly.
 -- The timer will emit the "timeout" signal every N seconds, N being the timeout
 -- value. Note that a started timer will not be garbage collected. Call `:stop`
@@ -88,7 +93,19 @@ function timer:start()
         print(traceback("timer already started"))
         return
     end
-    self.data.source_id = glib.timeout_add(glib.PRIORITY_DEFAULT, self.data.timeout * 1000, function()
+    local func, timeout
+    if self.data.precise then
+        func = glib.timeout_add,
+        timeout = self.data.timeout * 1000
+    else
+        func = glib.timeout_add_second
+        timeout = self.data.timeout
+        if select(2, math.modf(timeout)) ~= 0 then
+            -- Perhaps print some kind of warning?
+        end
+        timeout = math.ceil(timeout)
+    end
+    self.data.source_id = func(glib.PRIORITY_DEFAULT, timeout, function()
         protected_call(self.emit_signal, self, "timeout")
         return true
     end)
@@ -129,6 +146,8 @@ local timer_instance_mt = {
     __index = function(self, property)
         if property == "timeout" then
             return self.data.timeout
+        elseif property == "precise" then
+            return self.data.precise
         elseif property == "started" then
             return self.data.source_id ~= nil
         end
@@ -138,8 +157,14 @@ local timer_instance_mt = {
 
     __newindex = function(self, property, value)
         if property == "timeout" then
+            local int, fract = math.modf(value)
+            if self.data.precise == nil and fract ~= nil then
+                require("gears.debug").deprecate("Set timer mode to precise fractional timers. Timers default to imprecise, second-rounding mode.", {raw=true, deprecated_in=4})
+            end
             self.data.timeout = tonumber(value)
             self:emit_signal("property::timeout")
+        elseif property == "precise" then
+            self.data.precise = value and true or false
         end
     end
 }
@@ -158,7 +183,7 @@ function timer.new(args)
     args = args or {}
     local ret = object()
 
-    ret.data = { timeout = 0 } --TODO v5 rename to ._private
+    ret.data = { timeout = 0, precise = precise_default } --TODO v5 rename to ._private
     setmetatable(ret, timer_instance_mt)
 
     for k, v in pairs(args) do
```

blueyed:
I think we should default to precise for `select(2, math.modf(timeout))` though (instead of checking this for a warning).
This would then use `precise` always for e.g. `0.1`, which makes total sense to me.
And in v5 we would only change the behavior for `timeout % 1 == 0` - better without the magic for Lua 5.3.

Elv13:
How about `math.ceil(self.data.timeout) == math.floor(self.data.timeout)` instead of string parsing?

Also, +1 for the `precise` property. Maybe we can enable it by default for `1` and disable it by default for 2+?

psychon:
Minor comment: `math.modf(12.34)` returns two values: `12` and `34`. `select(2, ...)` throws away the first value and only leaves `34`. So, `select(2, math.modf(v)) == 0` checks if a number is an integer.

lib/gears/timer.lua
actionless: mb better mention real names of glib functions

Elv13:
Can you add a documentation paragraph in the header to explain the power efficiency implications and good practices?

lib/gears/timer.lua
```diff
@@ -88,7 +88,21 @@ function timer:start()
         print(traceback("timer already started"))
         return
     end
-    self.data.source_id = glib.timeout_add(glib.PRIORITY_DEFAULT, self.data.timeout * 1000, function()
+    local func, timeout
+    if self.data.precise == nil then
+        self.data.precise = select(2, math.modf(self.data.timeout)) ~= 0
```
psychon: Sorry, but no.

First, this is nuts once you write the docs for it:

> The default value for `precise` depends on the timeout that the timer is started with for the first time. If that timeout has no fractional value (i.e. is an integer), the default is `true`. Otherwise, the default is `false`.

Also, isn't this a behaviour change that we are not supposed to do in stable releases?

Elv13:
@blueyed ping?

blueyed:
Closing for now.

Issue #1588, opened 2017-02-20
Titlebar can be shown without any widget (awesome-4.0)


unode:
Output of `awesome --version`:

Gentoo Linux
```
awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.0
```

**How to reproduce the issue:**

Using the default `rc.lua` config.

1. Disable the awful.rule targeting clients with `type=[ "normal", "dialog" ]` to disable the default behavior of showing a titlebar.
2. Restart awesome.
3. Start a new terminal.
4. Start awesome-client on this terminal.
5. Run `awful.titlebar.toggle(client.focus)`
6. Run `awful.titlebar.toggle(client.focus)`
7. Run `client.focus:emit_signal("request::titlebars", "rules", {})`

**Actual result:**

On step 5 an empty titlebar is shown. I.e. no widgets.
On step 7 the normal titlebar is visible. All widgets rendered.

**Expected result:**

On step 5 the titlebar should be automatically initialized and all widgets should be visible.

psychon:
What is it that you *actually* want to do? You want titlebars to be toggleable with a key binding and want some hidden by default? If so, how about adding a call to `awful.titlebar.hide(c)` at the end of the section that adds widgets to the titlebar in your config?

Besides that, I have no idea what should be done here. "Titlebar can be shown without any widget" is kind of by design. I can also show a wibox without any widgets.

Unode:
@psychon I currently have:

```lua
awful.key({ machine.keys.modkey,           }, "y",
        _, size = client.focus:titlebar_top(client.focus, 'top')
        if size == 0 then
            -- This ensures it gets initialized and shown
            -- NOTE awful's code already takes care of ensuring a titlebar
            -- isn't initialized more than once so we don't need to care
            client.focus:emit_signal("request::titlebars", "rules", {})
        else
            awful.titlebar.hide(client.focus)
        end
    end,
    {description = "toggle titlebar", group = "client"}),
```

this:

```lua
awful.key({ machine.keys.modkey,           }, "y",
        _, size = client.focus:titlebar_top(client.focus, 'top')
        if size == 0 then
            awful.titlebar.show(client.focus)
        else
            awful.titlebar.hide(client.focus)
        end
    end,
[...]

PR #2421, opened 2018-10-06
Fix 2419

### LINKED ISSUE #2419: "request::titlebars" isn't sent when used with `toggle`

4.2

`titlebar_enabled =  true` -> `titlebar_enabled =  false` then add a keybinding to toggle them.

**Actual result:**

Nothing because `request::titlebars` wasn't called.

**Expected result:**

Somehow call the signal the first time a show/hide/toggle function is used.

(reporting it instead of a patch because I don't have time right now, its trivial to fix)

Elv13:
Fix #2419 #1588

### DIFF
--- a/lib/awful/titlebar.lua
+++ b/lib/awful/titlebar.lua
@@ -453,6 +453,32 @@ local function get_titlebar_function(c, position)
     end
 end
 
+--- Call `request::titlebars` to allow themes or rc.lua to create them even
+-- when `titlebars_enabled` is not set in the rules.
+-- @tparam client c The client.
+-- @tparam[opt=false] boolean hide_all Hide all titlebars except `keep`
+-- @tparam string keep Keep the titlebar at this position
+-- @treturn boolean If the titlebars were loaded
+local function load_titlebars(c, hide_all, keep)
+    if c._request_titlebars_called then return false end
+
+    c:emit_signal("request::titlebars", "awful.titlebar", {})
+
+    if hide_all then
+        -- Don't bother checking if it has been created, `.hide` don't works
+        -- anyway.
+        for _, tb in ipairs {"top", "bottom", "left", "right"} do
+            if tb ~= keep then
+                titlebar.hide(c, tb)
+            end
+        end
+    end
+
+    c._request_titlebars_called = true
+
+    return true
+end
+
 --- Get a client's titlebar.
 -- @tparam client c The client for which a titlebar is wanted.
 -- @tparam[opt={}] table args A table with extra arguments for the titlebar.
@@ -530,6 +556,7 @@ end
 --   "right", "top", "bottom". Default is "top".
 function titlebar.show(c, position)
     position = position or "top"
+    if load_titlebars(c, true, position) then return end
     local bars = all_titlebars[c]
     local data = bars and bars[position]
     local args = data and data.args
@@ -551,6 +578,7 @@ end
 --   "right", "top", "bottom". Default is "top".
 function titlebar.toggle(c, position)
     position = position or "top"
+    if load_titlebars(c, true, position) then return end
     local _, size = get_titlebar_function(c, position)(c)
     if size == 0 then
         titlebar.show(c, position)
--- a/tests/test-titlebar.lua
+++ b/tests/test-titlebar.lua
@@ -1,0 +1,94 @@
+local runner = require("_runner")
+local titlebar = require("awful.titlebar")
+local rules = require("awful.rules")
+local spawn = require("awful.spawn")
+
+local tiny_client = {"lua", "-e", [[
+local Gtk, class = require('lgi').require('Gtk'), 'client'
+Gtk.init()
+window = Gtk.Window {default_width=100, default_height=100, title='title'}
+window:set_wmclass(class, class)
+local app = Gtk.Application {}
+function app:on_activate()
+    window.application = self
+    window:show_all()
+end
+app:run {''}
+]]}
+
+-- Use the test client props
+rules.rules = {}
+
+-- Too bad there's no way to disconnect the rc.lua request::titlebars function
+
+local steps = {
+    function()
+        assert(#client.get() == 0)
+        spawn(tiny_client)
+        return true
+    end,
+    function()
+        if #client.get() ~= 1 then return end
+
+        local c = client.get()[1]
+
+        -- The rules don't set any borders nor enable the titlebar
+        assert(not c._request_titlebars_called)
+        assert(c.width == 100 and c.height == 100)
+
+        -- Should create the top titlebar
+        titlebar.toggle(c, "top")
+
+        assert(c._request_titlebars_called)
+
+        local h = c.height
+        assert(h > 100)
+
+        -- Should do nothing, there is no titlebar at the bottom by default
+        titlebar.toggle(c, "bottom")
+        assert(h == c.height)
+
+        -- Should hide the titlebar
+        titlebar.toggle(c, "top")
+        assert(c.height == 100)
+
+        c:kill()
+
+        return true
+    end,
+    function()
+        if #client.get() ~= 0 then return end
+
+        spawn(tiny_client, {titlebars_enabled=true})
+
+        return true
+    end,
+    function()
+        if #client.get() ~= 1 then return end
+
+        local c = client.get()[1]
+
+        local h = c.height
+        assert(c.width == 100 and h > 100)
+        assert(c._request_titlebars_called)
+
+        titlebar.hide(c, "top")
+
+        assert(c.width == 100 and c.height == 100)
+
+        titlebar.hide(c, "top")
+
+        assert(c.width == 100 and c.height == 100)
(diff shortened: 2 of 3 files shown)


lib/awful/titlebar.lua
```diff
@@ -453,6 +453,32 @@ local function get_titlebar_function(c, position)
...
+    if c._request_titlebars_called then return false end
+
+    c:emit_signal("request::titlebars", "awful.titlebar", {})
+
+    if hide_all then
+        -- Don't bother checking if it has been created, `.hide` don't works
+        -- anyway.
+        for _, tb in ipairs {"top", "bottom", "left", "right"} do
+            if tb ~= keep then
+                titlebar.hide(c, tb)
```
psychon: This is a behaviour change.  If I do `awful.titlebar.show(c, "top")` and then `awful.titlebar.show(c, "bottom")`, the top titlebar is now hidden. Why?!?

lib/awful/titlebar.lua
```diff
@@ -551,6 +578,7 @@ end
 --   "right", "top", "bottom". Default is "top".
 function titlebar.toggle(c, position)
     position = position or "top"
+    if load_titlebars(c, true, position) then return end
```
psychon: Why this new early return? This should break any keybindings that toggle titlebar visibility. I can also come up with an example where the new early return in `titlebar.show` breaks stuff.

lib/awful/titlebar.lua
```diff
@@ -453,6 +453,32 @@ local function get_titlebar_function(c, position)
     end
 end
 
+--- Call `request::titlebars` to allow themes or rc.lua to create them even
+-- when `titlebars_enabled` is not set in the rules.
+-- @tparam client c The client.
+-- @tparam[opt=false] boolean hide_all Hide all titlebars except `keep`
+-- @tparam string keep Keep the titlebar at this position
+-- @treturn boolean If the titlebars were loaded
+local function load_titlebars(c, hide_all, keep)
```
psychon: This function is only ever called with `hide_all=true`. Why is this a parameter?

lib/awful/titlebar.lua
Elv13: It wont because this code is only executed once

        if c._request_titlebars_called then return false end

lib/awful/titlebar.lua
Elv13: Because if `load_titlebar` returns true, then it is already shown

psychon: Okay, and who hides it then?

lib/awful/titlebar.lua
psychon: Oh, right. Then... why does this hide titlebars that cannot be there?

lib/awful/titlebar.lua
Elv13: toggle?

If load_titlebar load something, then toggle will hide it. But if the user call toggle and there is no titlebar, then he/she want to show it?

Otherwise I don't understand the question

lib/awful/titlebar.lua
Elv13: because if toggle or show is called, then the user wants to show one of them. If many are loaded, then they should not be shown.

lib/awful/titlebar.lua
psychon: Okay, so: .... oh wait, I got the return value of `load_titlebars` wrong. Sorry!

lib/awful/titlebar.lua
psychon: So, this is an intentional change in behaviour then and needs deprecation and has to wait until v5?
Previously, toggling the left titlebar, did not do anything with the other titlebars at all.

psychon:
Gentle ping on "what's wrong with this patch instead?"
```diff
diff --git a/lib/awful/rules.lua b/lib/awful/rules.lua
index c652ede31..6a2ef9b76 100644
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -583,6 +583,7 @@ end
 function rules.execute(c, props, callbacks)
     -- This has to be done first, as it will impact geometry related props.
     if props.titlebars_enabled then
+        c._requested_titlebars = true
         c:emit_signal("request::titlebars", "rules", {properties=props})
     end
 
diff --git a/lib/awful/titlebar.lua b/lib/awful/titlebar.lua
index 969b4076b..2df8f496c 100644
--- a/lib/awful/titlebar.lua
+++ b/lib/awful/titlebar.lua
@@ -468,6 +468,11 @@ end
 -- @tparam[opt=top] string args.font
 -- @function awful.titlebar
 local function new(c, args)
+    if not c._requested_titlebars then
+        c._requested_titlebars = true
+        c:emit_signal("request::titlebars", "awful.titlebar", {})
+    end
+
     args = args or {}
     local position = args.position or "top"
     local size = args.size or gmath.round(beautiful.get_font_height(args.font) * 1.5)
```

lib/awful/titlebar.lua
Elv13: As far as I see, no, it does not change anything other than fixing the bug. You go from having zero titlebars. Then you want to show one of them. One of them is shown.

Because request::titlebars can potentially load 4 titlebars, the other have to be hidden since they were not requested to be shown.

Elv13:
> Gentle ping on "what's wrong with this patch instead?"

If more than 1 titlebar is loaded by request::titlebars, it will not behave as documented.

psychon:
Hm. So, basically, everything we can do is weird because there is an imbalance between "show this one titlebar" and "set up all the titlebars". :-/

How about fixing this in the default config instead?
Every client gets a titlebar set up and `request::titlebar` then only triggers visibility of that titlebar?

I don't really know...

Elv13:
> How about fixing this in the default config instead?
> Every client gets a titlebar set up and request::titlebar then only triggers visibility of that titlebar?

I am neutral against adding or not adding 4 more signals. This should be done in another pull request as it doesn't fix the issue this pull request addresses. Also, given it require the user to also modify the titlebar declaration, you now have to specify the position **twice** (one in the signal name and the other in the `awful.titlebar(c {position=...})`. So it adds complexity for solving a problem less than <0.5% of the users will ever hit. Assuming that >99.5% of people has either zero or one titlebar, it's an imperfect solution. See 3rd section of this comment for my proposed long term alternative.

My solution final result is AFAIK correct for 100% of the users without any modified `rc.lua` code and behavior change. This proposed solution only solves half of the issue even if all users config could magically be upgraded, which it can't.

> Hm. So, basically, everything we can do is weird

Yes, this is why the code was written that way instead of any other way. It wasn't pretty when I wrote it and it fixes an equally ugly "race condition" bug. Right now (and since forever, given the titlebars created in "manage" in v3.5 had the same issue), we have two broken functions in `awful.titlebar`. After this patch, they are working as expected with only a potential flickering for <0.5% of the userbase who have multiple titlebars *and* use this function to toggle them.

> because there is an imbalance between "show this one titlebar" and "set up all the titlebars". :-/

Not exactly, the issue as I see it is that the concept of a titlebar area(s) and the titlebar widget are coupled beyond common sense. My proposed solution to this impasse has long been in the form of #655. It isn't done yet and I think we both agree I have too many long-ongoing pull request series to add even more. This pull request make `show` and `toggle` work when `titlebars_enabled` is set to `false` and add regression tests. It does not attempt to solve the reason why this bug exist.

lib/awful/titlebar.lua
Elv13: A second PR will add `hide_all`/`show_all`/`toggle_all`. I didn't want to scope creep this PR.

Elv13:
@psychon thanks

@jd I think mergify should have worked here, any idea?

jd:
@Elv13 looks like a caching issue. I just refreshed.

Elv13:
thanks

blueyed:
@Elv13 
I think this change caused an issue with my config, which was not adjusted for `request::titlebars`, but created the titlebars still manually (and hides them for most clients) in manage:
```lua
        awful.titlebar(c):set_widget(layout)

        if c.type ~= 'dialog' and c.name ~= 'Thunderbird Preferences' then
          awful.titlebar.hide(c)  -- NOTE: triggers layout arranging
        end
```
The first call to `toggle` would then do nothing (also when not hiding them).
I will adjust my config to move this into a request handler also, but just wanted to note this.

tests/test-titlebar.lua
```diff
@@ -0,0 +1,94 @@
...
+    window.application = self
+    window:show_all()
+end
+app:run {''}
+]]}
+
+-- Use the test client props
+rules.rules = {}
+
+-- Too bad there's no way to disconnect the rc.lua request::titlebars function
```
blueyed: This could be either disconnected if it was set as a function (so its reference could be used), or with something that would allow to unset any signal handlers (for a given signal).
For the 2nd idea, it could provide a mechanism to restore them later, useful for tests and maybe even in general, if you want to execute something without any signals?!

PR #2422, opened 2018-10-06
keygrabber: Rename release_event to stop_event

Elv13:
In the earlier revision of the keygrabber PR, there was a `release_key`
and it was suggested to rename it `stop_key`. However its sibling
`release_event` wasn't, so it is now confusing.

The commit adds a mild deprecation codepath to avoid breaking configs
based on git-master. However it isn't a "long term" deprecation notice
and the code can probably be removed in 5.0 without further delay.

### DIFF
--- a/lib/awful/keygrabber.lua
+++ b/lib/awful/keygrabber.lua
@@ -546,10 +546,8 @@ end
 
 --TODO v5 remove this
 function keygrabber:set_release_event(event)
-    gdebug.deprecate("release_event has been renamed to stop_event",
-        -- It has never been in a release, so it can be deprecated right away
-        {deprecated_in=4}
-    )
+    -- It has never been in a release, so it can be deprecated right away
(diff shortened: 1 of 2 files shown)


lib/awful/keygrabber.lua
```diff
@@ -544,6 +544,21 @@ function keygrabber:set_allowed_keys(keys)
     end
 end
 
+--TODO v5 remove this
+function keygrabber:set_release_event(event)
+    gdebug.deprecate("release_event has been renamed to stop_event",
+        -- It has never been in a release, so it can be deprecated right away
+        {deprecated_in=4}
+    )
```
psychon: Just remove the second argument to this function. That way, `gears.debug.deprecate` does not do any version check.

Elv13:
Fixup done in the GitHub UI, to whoever merge this, please squash

Issue #2342, opened 2018-08-06
Missing mkdir when something deleted doc/images/ in the build directory


Elv13:
Another little regression

Step to reproduce:

```sh
git clone awesome
cd awesome
mkdir build && cd build
cmake ..

# The SVG are in the BUILD_DIRECTORY

make

# The doc is generated

rm -rf doc

make

# A lot of Generating build/doc/images/AUTOGEN_awful_placement_no_offscreen.svg...
# messages, but no SVG is in the BUILD_DIRECTORY
```

psychon:
The problem is that `build/images/` is not `mkdir`'d again (it only happens in `tests/examples/CMakeLists.txt`, so only during "configure"). So... either there needs to be a target that `mkdir`s this directory and all the tests have to be changed to depend on this, or the tests themselves have to be thaught to `mkdir` this target. Dunno, both options are somewhat ugly.

The problem is so silent due to missing error checking in the example tests. The following makes the tests (that use this one template) fail instead:
```diff
diff --git a/tests/examples/awful/template.lua b/tests/examples/awful/template.lua
index 9f84404f..51b277d7 100644
--- a/tests/examples/awful/template.lua
+++ b/tests/examples/awful/template.lua
@@ -230,5 +231,6 @@ for _, h in ipairs(mouse.old_histories) do
 end
 
 img:finish()
+assert(img.status == "SUCCESS", img.status)
 
 -- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
```

> Another little regression

Nope, no regression. This problem exists since commit 02f894ce0cb14f05b78ee76dacdf1ecab7ecc09a. Before that, tests only ever run during `cmake`, so the `mkdir` was done by CMake. This commit ensures that the example tests are re-run during `make` if one of their dependencies changed.

PR #2423, opened 2018-10-07
Better awful.rules doc

Elv13:
What's done:

 * Move the documentation parser into a module
     * Add C documentation support to the parser
     * Make the parser more flexible
 * Update the CSS to render the parser output in a similar way to the other components
 * Add a table of all client properties to `awful.rules` and `awful.spawn`
 * Refactor `awful.rules` documentation
 * Refactor `awful.spawn` documentation

This pull request is mostly moving code around and adding boilerplate `ldoc` packaging for the rules own properties.

~~There is still somehow a target dependency issue where the new rules table only exists the second time the doc is built. It might just be late, but I can't see it right now.(edit: I will need to move `configure_file` into a target, but it wont be easy given it uses many variables)

Also the links are often broken in the table because `awful.client` redirects to `client` and the parser is too dumb to notice. I think it's worth merging even with that bug, but the dependency issue needs to be solved.~~ 

Finally, if anyone disagree with the new documentation, just edit the pull request instead of making comments.


Elv13:
Updated/Finished

The hack to keep the client property table up-to-date isn't pretty, but changing the build system to get rid of `configure_file` and really, really, do nothing during the configuration phase is too large for this PR. So for now it just make sure to get the file in the commit.

Beside that I updated the CSS style to make the tables blend better, it follows @actionless initial work on this by adding an header style and padding that match GitHub issues tables. It also add `<code>` and `<kdb>` style to also match GitHub ones.

lib/awful/client.lua
```diff
@@ -59,6 +59,93 @@ client.property = {}
...
+--
+-- The `request::activate` context for this call is `rules`.
+--
+-- @clientruleproperty focus
+-- @param[opt=false] boolean
+
+--- Should this client have a titlebar by default.
+-- @clientruleproperty titlebars_enabled
+-- @param[opt=false] boolean
+-- @see awful.titlebar
```
psychon: Arguably, all the `@param boolean` above are wrong, since the code also supports calling functions that then return a boolean (IIRC).

awesomeConfig.cmake
```diff
@@ -354,13 +354,24 @@ add_custom_target(lgi-check-run ALL
 
 # {{{ Generate some aggregated documentation from lua script
 
-file(MAKE_DIRECTORY ${BUILD_DIR}/script_files/)
+# file(MAKE_DIRECTORY ${BUILD_DIR}/script_files/)
+# file(MAKE_DIRECTORY ${BUILD_DIR}/docs/common/)
+# file(COPY ${SOURCE_DIR}/docs/_parser.lua DESTINATION ${BUILD_DIR}/docs/)
```
psychon: Why these new comments?

awesomeConfig.cmake
```diff
@@ -354,13 +354,24 @@ add_custom_target(lgi-check-run ALL
...
+
+add_custom_target(setup_directories DEPENDS lgi-check-run)
+
+add_custom_command(TARGET setup_directories
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${BUILD_DIR}/script_files/
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${BUILD_DIR}/docs/common/
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${BUILD_DIR}/doc/images/
+        COMMAND ${CMAKE_COMMAND} -E copy ${SOURCE_DIR}/docs/_parser.lua ${BUILD_DIR}/docs/
```
psychon: Why these new statements? I cannot find anything creating these directories previously.

docs/_parser.lua
```diff
@@ -0,0 +1,196 @@
+local gio = require("lgi").Gio
```
psychon: Could you move the addition of this file and the new `COPY` for it (why is that needed?) to the next commit? That way, it doesn't look like just new code and one can actually see which part was just moved and which is actually new.

CMakeLists.txt
```diff
@@ -289,7 +289,11 @@ if(GENERATE_DOC)
     file(COPY ${SOURCE_DIR}/docs/ldoc.css DESTINATION ${BUILD_DIR}/docs)
 
     add_custom_target(ldoc ALL
-        DEPENDS ${BUILD_DIR}/doc/index.html ${BUILD_DIR}/docs/ldoc.css)
+        DEPENDS
+            ${BUILD_DIR}/doc/index.html
+            ${BUILD_DIR}/docs/ldoc.css
+            generate_awesomerc
+    )
```
psychon: I'm not sure if this change should be in this commit. It doesn't have anything to do with the index, I think. Perhaps the previous commit (which already has some unexplained dependency additions/fixes anyway).

awesomeConfig.cmake
```diff
@@ -371,7 +371,26 @@ add_custom_command(
...
+
+add_custom_command(
+        OUTPUT ${BUILD_DIR}/docs/common/rules_index.ldoc
+        COMMAND lua ${SOURCE_DIR}/docs/build_rules_index.lua
+            ${BUILD_DIR}/docs/common/rules_index.ldoc
+
+        # Cheap trick until the ldoc `configure_file` is ported to be a build
+        # step rather than part of cmake.
+        COMMAND ${CMAKE_COMMAND} -E copy ${BUILD_DIR}/docs/common/rules_index.ldoc
+            ${SOURCE_DIR}/docs/common/rules_index.ldoc
```
psychon: Uhm... Does this copy something into the *source* directory? Why?!?
Isn't the build supposed not to touch the source directory in any way? Autoconf even checks for that during `make distcheck` by testing an out-of-tree build against read-only sources.

docs/06-appearance.md.lua
```diff
@@ -229,8 +70,9 @@ f:write[[
 
 Beautiful is where Awesome theme variables are stored.
 
+<br \><br \>
 ]]
-f:write(create_table(beautiful_vars, {"link", "desc"}))
+f:write(parser.create_table(beautiful_vars, {"link", "desc"}))
```
psychon: Looks like this only changes the code to use the new `parser` (so should be done in the commit that adds `parser`, I guess)... and then, some new `<br \><br \>` appear that seem totally unrelated. Why?

docs/06-appearance.md.lua
```diff
@@ -229,8 +70,9 @@ f:write[[
 
 Beautiful is where Awesome theme variables are stored.
 
+<br \><br \>
```
psychon: Shouldn't it be `<br />` instead of `<br \>`?

docs/common/rules_index.ldoc
```diff
@@ -0,0 +1,72 @@
+--<table class='widget_list' border=1>
```
psychon: Why is this file added to the source if it is generated during the build?

docs/load_ldoc.cmake
```diff
@@ -1,23 +1,23 @@
 # To avoid copy pasting, some documentation is stored in reusable files
 set(SHAPE_FILE "${SOURCE_DIR}/docs/common/${SHAPE_NAME}.lua")
 
-set(path "${SOURCE_DIR}/docs/common/")
+foreach(path ${BUILD_DIR}/docs/common/;${SOURCE_DIR}/docs/common/)
```
psychon: This is just for `rules_index.ldoc`, right? If so, when that file is generated in the build dir, then this piece of code is not needed.

(Also, I just remember that `rules_index.ldoc` is copied around... doesn't that mean that ldoc ends up parsing both the in-source and in-build copy of the file?)

lib/awful/rules.lua
```diff
@@ -8,17 +8,12 @@
...
--- * placement
--- * honor_padding
--- * honor_workarea
--- * tag
--- * new_tag
--- * switch_to_tags (also called switchtotag)
--- * focus
--- * titlebars_enabled
--- * callback
+-- This table contains the client properties that are valid when used in rules.
```
psychon: for use in rules?

But the whole sentence strikes me as a bit odd. You can add anything to rules and it is just set as a property on the client...

lib/awful/spawn.lua
```diff
@@ -45,45 +48,6 @@
...
---
--- This can also be used from the command line:
---
---    awesome-client 'awful=require("awful");
---      awful.spawn("urxvt -e maxima -name CALCULATOR", {
---        floating  = true,
---        tag       = mouse.screen.selected_tag,
---        placement = awful.placement.bottom_right,
---      })'
---
```
psychon: This change seems unrelated to this commit...

lib/awful/spawn.lua
```diff
@@ -48,6 +48,57 @@
...
+-- command. It is strongly recommanded not to use `io.popen` is explained in the
+-- "Getting a command's output" section. Asynchronous execution is at first a
+-- bit tricky to understand if you never used that before. The example below
+-- should demonstrate how it works.
+--
+-- If we do (but *really*, don't do that):
+--
+--     -- **NEVER, EVER, DO THIS**: your computer will freeze
+--     os.execute("sleep 1; echo foo > /tmp/foo.txt")
+--     mylabel.text = io.popen("cat /tmp/foo.txt")
```
psychon: `popen` does not work like this, does it? It returns a file handle that you can then read from.

lib/awful/spawn.lua
```diff
@@ -48,6 +48,57 @@
...
+--
+-- The label will display `foo`. But If we do:
+--
+--     -- Don't do this, it wont work.
+--     -- Assumes /tmp/foo.txt does not exist
+--     awful.spawn.with_shell("sleep 1; echo foo > /tmp/foo.txt")
+--     mylabel.text = io.popen("cat /tmp/foo.txt")
```
psychon: Same

lib/awful/spawn.lua
```diff
@@ -48,6 +48,57 @@
...
+--
+-- Then the label will be **empty**. `awful.spawn` and `awful.spawn.with_shell`
+-- will **not** block and thus the `io.popen` will be executed before
+-- `sleep 1` finishes. This is why we have async functions to execute shell
+-- commands. There is many variants with difference characteristics and
```
psychon: There *are*

lib/awful/client.lua
Elv13: Yeah, but that's true for all rules. Treating them differently for client properties make little sense here. I don't think it is worth documenting this in a different way here.

awesomeConfig.cmake
Elv13: done

awesomeConfig.cmake
Elv13: One of them is created in `tests/CMakeLists.txt`. The others by `configure_file`. It fixes #2342. I will add a note in the commit message

docs/_parser.lua
Elv13: > why is that needed?

To avoid messing with LUA_PATH. I can use either solution, I don't mind.

CMakeLists.txt
Elv13: done

awesomeConfig.cmake
Elv13: Not pretty, I will agree on that. The problem is that `configure_file` depends on it and fixing that will be non-trivial. tl;dr, we will need a `configure_file` equivalent implemented in Lua. It is worth doing and generally would make sense anyway. However I chose not to do it in this pull request as I want v4.3 released and this is part 2 of 4 of the "improve spawn" serie.

docs/06-appearance.md.lua
Elv13: > appear that seem totally unrelated. Why?

It was already there https://github.com/awesomeWM/awesome/blob/master/docs/06-appearance.md.lua#L166

docs/06-appearance.md.lua
Elv13: done

docs/common/rules_index.ldoc
Elv13: dependency loop, see above

docs/load_ldoc.cmake
Elv13: see previous comment

lib/awful/rules.lua
Elv13: done

lib/awful/spawn.lua
Elv13: done

lib/awful/spawn.lua
Elv13: done

lib/awful/spawn.lua
Elv13: done

Elv13:
Updated

psychon:
Shall I take another look or shall we do this the easy way instead?

Elv13:
> Shall I take another look or shall we do this the easy way instead?

I guess the "it modifies the source directory" is still a rather controversial issue. If you are fine with my "I am lazy" explanation, press the merge button.

Elv13:
I guess that's a yes. If nobody else cares to review, I will merge that tomorrow so I can push the next PR of this serie (and last for v4.3, the last one refactor the `.desktop` support so you can use them in spawn, I am not merging that for 4.3)

PR #2429, opened 2018-10-12
[RFC/Incomplete] Add more "honor" properties like used for size_hints.

### LINKED ISSUE #2411: gvim is started on wrong screen (i.e. not current) in 3 monitor setup
 ```
awesome v4.2-446-g3931ae37 (Human after all)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

* launch `gvim`
* window appears on leftmost monitor instead of current monitor

Note that it works correctly (gvim is displayed on the current screen), when:
* the current screen is empty
* the current screen has maximized window (and I run gvim with awesome)

For example, the following does not work (vim is started on the wrong screen):
* start `termite` (or `xterm`)
* maximize it (Mod4+m)
* launch `gvim` from the terminal

*However* if I do the following, it works:
* start `termite` (or `xterm`)
* maximize it (Mod4+m)
* run `gvim` using awesome (Mod4+r)

More info:
```
$ xrandr
Screen 0: minimum 8 x 8, current 11520 x 2400, maximum 32767 x 32767
DVI-D-0 connected 3840x2400+0+0 (normal left inverted right x axis y axis) 518mm x 324mm
   1920x1200     59.95*+
   1920x1080     60.00    59.93  
   1680x1050     59.95  
   1600x1200     60.00  
   1440x900      74.98    59.89  
   1280x1024     75.02    72.00    60.02  
   1024x768      75.03    70.07    60.00  
   800x600       75.00    72.19    60.32  
   640x480       75.00    72.81    65.99    59.94  
HDMI-0 disconnected (normal left inverted right x axis y axis)
DP-0 connected primary 3840x2160+3840+0 (normal left inverted right x axis y axis) 527mm x 296mm
   3840x2160     29.98*+  30.30  
   1920x1200     59.88  
   1920x1080     60.00    60.00    59.94    50.00    23.98  
   1680x1050     59.95  
   1600x1200     60.00  
   1280x1024     75.02    60.02  
   1280x800      59.81  
   1280x720      60.00    59.94    50.00  
   1152x864      75.00  
   1024x768      75.03    60.00  
   800x600       75.00    60.32  
   720x576       50.00  
   720x480       59.94  
   640x480       75.00    59.94    59.93  
DP-1 disconnected (normal left inverted right x axis y axis)
DP-2 connected 3840x2400+7680+0 (normal left inverted right x axis y axis) 518mm x 324mm
   1920x1200     59.95*+
   1920x1080     60.00    59.94    50.00    50.00  
   1680x1050     59.95  
   1600x1200     60.00  
   1440x900      74.98    59.89  
   1280x1024     75.02    71.93    60.02  
   1280x720      60.00    59.94    50.00  
   1024x768      75.03    70.07    60.00  
   800x600       75.00    72.19    60.32  
   720x576       50.00  
   720x480       59.94  
   640x480       75.00    72.81    65.36    59.94    59.93  
DP-3 disconnected (normal left inverted right x axis y axis)
DP-4 disconnected (normal left inverted right x axis y axis)
DP-5 disconnected (normal left inverted right x axis y axis)
```

There's no special rules:
```lua
awful.rules.rules = {
    { rule = { },
      properties = { border_width = beautiful.border_width,
                     border_color = beautiful.border_normal,
                     focus = awful.client.focus.filter,
                     raise = true,
                     keys = clientkeys,
                     buttons = clientbuttons,
                     screen = awful.screen.preferred,
                     placement = awful.placement.no_overlap+awful.placement.no_offscreen
     }
    },

    -- Floating clients.
    { rule_any = {
        instance = {
          "DTA",  -- Firefox addon DownThemAll.
          "copyq",  -- Includes session name in class.
        },
        class = {
          "Arandr",
          "Gpick",
          "Kruler",
          "MessageWin",  -- kalarm.
          "Sxiv",
          "Wpa_gui",
          "pinentry",
          "veromix",
          "xtightvncviewer"},

        name = {
          "Event Tester",  -- xev.
        },
        role = {
          "AlarmWindow",  -- Thunderbird's calendar.
          "pop-up",       -- e.g. Google Chrome's (detached) Developer Tools.
        }
      }, properties = { floating = true }},

    -- Add titlebars to normal clients and dialogs
    { rule_any = {type = { "normal", "dialog" }
      }, properties = { titlebars_enabled = false }
    },

    -- Set Firefox to always map on the tag named "2" on screen 1.
    -- { rule = { class = "Firefox" },
    --   properties = { screen = 1, tag = "2" } },
}
```

Elv13:
### Purpose

This commit adds the ability to ignore clients requests to be set
maximized, fullscreen or move themselves.

The main use case is applications that try to "help users" by
remembering their state. There is two pathways for this. They can
do it during the client initialization where normal `awful.rules`
would have been enough or they can do it later in which case rules
are useless.

Adding those new properties streamline the process and they work both
in `awful.rules` and later during execution. It removes the need to
replace `awful.ewmh.merge_maximization` and
`awful.ewmh.client_geometry_requests` request handlers.

It also adds a fake fullscreen property to "help" applications which provide superior fullscreen modes than their normal modes to provide their greatness to the users of tiling window managers.

Fix #2411

### Status

This is only a draft, it is half implemented, I didn't even try to execute the code. It's missing tests and currently cause grave maximization regressions because the X properties become out of sync. I shot myself in the foot before touching this code and I remember doing so. This isn't for v4.3 and is only provided as guidance if someone want to take over and finish it.

### Todo

 * Update the tests
 * Replace `client_set_fullscreen` by `"request::geometry", "ewmh_fullscreen"` signals
 * Make sure the X properties are kept in sync if the request are denied


Elv13:
Oh, and also, `resize_requests_honor` is weird. I called it this way because `size_hints_honor` exists. Wouldn't it be better to rename these `honor_size_hints` instead of adding more weirdly names ones for consistency?

psychon:
> Wouldn't it be better to rename these honor_size_hints instead of adding more weirdly names ones for consistency?

Sure.

lib/awful/client.lua
```diff
@@ -149,6 +149,74 @@ function client.tiled(s, stacked)
...
+-- @see awful.ewmh.client_geometry_requests
+-- @see size_hints_honor
+-- @see fullscreen
+-- @see maximize_requests_honor
+
+--- Tell to clients they are fullscreen even if they are not.
+--
+-- Some application have special fullscreen modes that are either superior,
+-- less bloated or better suited for tiling than their regular mode. For
+-- examples, web videos don't have ads for more content when played fullscreen.
```
psychon: I don't think this applies to all web videos, but whatever... :-)

Elv13:
(This is still untested and unfinished)

After another question on stackoverflow regarding blocking resize/move requests, I added some more code to this. It still needs to be extended so maximization, size_hints and fullscreen create request:: style calls, but the part to grant/deny the existing "activate" (focus) and "geometry" (resize) request is there. It mutualize the filter system already used for request::activate and makes it available for "all client requests" (note: the list is hardcoded). On top of that, a couple layers of abstraction are added to make it available from `awful.rules` and `tyrannical`. The end result is tighter integration with existing concepts.

Elv13:
Parking the `request::size_hints` commit here since it is what the PR was about. I wont use it for the first iteration of the granted_requests/permissions system. I will make a new PR for that. The system proposed here is otherwise required for the screen rules / activities project.

Issue #2095, opened 2017-10-31
WM_STATE cannot be set to Iconified on restart


wakjah:
Output of `awesome --version`:

awesome v4.2 (Human after all)
 • Compiled against Lua 5.2.3 (running with Lua 5.2)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.4
 • LGI version: 0.8.0

**How to reproduce the issue:**

Set a client to minimized in the "manage" callback. 

```
client.connect_signal("manage", function (c, startup)
    c.minimized = true
end)
```

Run a client, e.g., `xclock`

The window should disappear and `xprop -name xclock WM_STATE` should say

> WM_STATE(WM_STATE):
>        window state: Iconic

Now restart awesome. 

**Actual result:**

The window disappears as expected, but  `xprop -name xclock WM_STATE` instead says 

> WM_STATE(WM_STATE):
>        window state: Normal

**Expected result:**

The window disappears and `WM_STATE` is still `Iconic`.

Note that it appears I can work around the issue by changing the callback to 

```
client.connect_signal("manage", function (c, startup)
    c.minimized = false
    c.minimized = true
end)
```

It seems that the first line (in either of the two versions of the callback) results in something like the following being printed

> [  1308.86 ] error 3 (BadWindow) request 2 minor 0 serial 8329 ("BadWindow (invalid window parameter)")

psychon:
I'm curious: *Why* do you check the value of `WM_STATE` of `xeyes`? How did you actually come across this problem?
And follow up question: Why do you use `WM_STATE` instead of `_NET_WM_STATE` for this?

> It seems that the first line (in either of the two versions of the callback) results in something like the following being printed
> 
> [ 1308.86 ] error 3 (BadWindow) request 2 minor 0 serial 8329 ("BadWindow (invalid window parameter)")

That does not happen here / cannot reproduce.

wakjah:
I realise it might seem a little strange :) 

Some background: we have an embedded system that uses awesome as the window manager. Some of the applications are java apps that use the 'iconified' property as an optimization to avoid unnecessary updates. All works pretty well (as long as we set _JAVA_AWT_WM_NONREPARENTING of course). 

It appears that awesome has started crashing for some users (or a secondary issue is causing it to - it is under investigation, reason currently unknown), and the mechanism is breaking due to the above mentioned issue. 

From reading the source, it looks like the JVM is using WM_STATE to determine whether a window is iconified.

[...]

PR #2432, opened 2018-10-14
fix(awful: rules): don't raise every focusable client on awesome startup

actionless:
re: #2095

### DIFF
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -735,7 +735,7 @@ function rules.execute(c, props, callbacks)
 
     -- Do this at last so we do not erase things done by the focus signal.
     if props.focus and (type(props.focus) ~= "function" or props.focus(c)) then
-        c:emit_signal('request::activate', "rules", {raise=true})
+        c:emit_signal('request::activate', "rules", {raise=not awesome.startup})
     end
 end
 


psychon:
I wonder if this patch makes more sense:
```diff
diff --git a/lib/awful/ewmh.lua b/lib/awful/ewmh.lua
index 139e8e1a5..3ddb9f29f 100644
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -116,7 +116,7 @@ function ewmh.activate(c, context, hints) -- luacheck: no unused args
 
     -- Minimized clients can be requested to have focus by, for example, 3rd
     -- party toolbars and they might not try to unminimize it first.
-    if ret ~= false and hints.raise then
+    if ret ~= false and hints.raise and not awesome.startup then
         c.minimized = false
     end
 
```
By the comment seen here, this is about requests sent by 3rd party toolbars, so this code should not be necessary during startup. I don't really know if this patch is better than this PR, but at least my versions feels less "magic" or "huh? why does it look at `awesome.startup`?!?" to me. What do you think?

psychon:
But this PR is also fine with me

Elv13:
I like @actionless version more, it has a smaller scope so is less prone to side effects like "hey, I *want* this to happen, why block me" for whatever weird use case someone comes up with.

psychon:
Uhm, what happened to the mergify status and mergify itself?

actionless:
@jd ping

PR #2436, opened 2018-10-15
Migrate Mergify configuration to v2

jd:
This just convert the file to the new configuration format.
The action "delete_head_branch" is added to clean up branches after they are
merged.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,11 +1,14 @@
-rules:
-  default:
-    protection:
-      required_status_checks:
-        contexts:
-          - continuous-integration/travis-ci
-          - Codacy/PR Quality Review
-          - codecov/patch
-          - coverage/coveralls
-      required_pull_request_reviews:
-        required_approving_review_count: 2
+pull_request_rules:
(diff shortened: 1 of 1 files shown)


jd:
@Elv13 You can do that indeed. :)

Elv13:
> @Elv13 You can do that indeed. :)

Great, so the next step is to have the rebase+merge_commit and skip merge commit for single commit PRs ;)? I really miss this. One of the issue I have with mergify right now is that the commit log looks like spaghetti

screenshot_20181015_150021

All those twisted branches are really not necessary. If it was rebased on HEAD before merging, it would not happen and that screenshot would look like a "vertical snake".

actionless:
we could make two extra rulesets with:

```
    conditions:
      - label=rebase
```

and 

```
    conditions:
      - label=squash
```

for choosing explicitly action, right?

Elv13:
Almost, I will open another issue there to request adding a "commit-count" or "#commit" condition.

actionless:
yeah, that would be even better :+1:

PR #2440, opened 2018-10-16
gears.table.join: Ignore nil arguments

psychon:
When calling join with e.g. arguments (nil, {"a"}), then everything past
the nil was ignored, because the code internally used ipairs() to
iterate over the arguments and this stops at the first nil it
encounters.

Fix this by using select() to iterate over the arguments.

This also adds a unit test for this problem.


Elv13:
Merged manually because coveralls is down

PR #2441, opened 2018-10-16
tests: Fix a "rare" build failure

Elv13:
I ran a `while true; do TEST_PAUSE_ON_ERRORS=1 ../tests/run.sh ../tests/test-spawn.lua; done` on my computer last night with no failures.

Apparently about 10% of the builds currently fail "randomly" on when raising a "spawn_or_raise" client with support for startup_id.

This PR will add more asserts until it goes away or a bug is found.

### DIFF
--- a/tests/test-spawn.lua
+++ b/tests/test-spawn.lua
@@ -228,19 +228,21 @@ local steps = {
         assert(client.get()[1]:tags()[1] == screen[1].tags[3])
         client.get()[1]:kill()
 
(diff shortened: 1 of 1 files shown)


Elv13:
I have restarted the build a dozen time in a row without a failure. I think the current patch fixes the issue. Even if it doesn't, at least the asserts have been re-ordered to catch it earlier instead of catching later side effects.

Issue #2449, opened 2018-10-22
Background of transparent icons in systray is not refreshed


petoju:
Output of `awesome --version`:

awesome v4.2-496-g2458f127-dirty (Human after all)
 • Compiled against Lua 5.2.4 (running with Lua 5.2)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

**How to reproduce the issue:**

1. Update to Ubuntu 18.10 with awesome built from latest head, maybe unknowingly break drivers (3D acceleration and games still work)
2. Look into systray 

**Actual result:**

In systray, background of transparent icons is not refreshed. Background of systray itself is fine, but background of transparent icon gets color of last window or window border being there. Foreground is also fine.

**Expected result:**

Background of icons will be consistent independently from what was above the icons before.

Note: this issue is similar to other systray problems, but this differs from each of them at least a bit. I have the solution, that works for me.

petoju:
Fix for this is in
https://github.com/awesomeWM/awesome/pull/2450

psychon:
Here is a screenshot of my wibar. As you can see, at the right side there is a tray icon with a transparent background. This tray icon correctly shows the background color of my wibar.

![foo](https://user-images.githubusercontent.com/89482/47339031-36f2f580-d69a-11e8-9ffb-e5dae3f0bb03.png)

So, could you perhaps provide some more information? Anything that might be different between your system and mine?

psychon:
> background of transparent icon gets color of last window or window border being there.

> this issue is similar to other systray problems, but this differs from each of them at least a bit.

How is it different from #2190? To me, this just sounds like another problem caused by Qt grabbing a screenshot and just using that for the background of its systray icon. (Although that doesn't explain why your patch then fixes this...)

That issue then links to #1720, #1738, #1856, #283, #176, and #1947. I didn't look at all of them, I just copied over the list from #2190, but how come you are sure that your issue is different than all of these?

McSinyx:
Slightly related to [this GTK+ 3 systray bug](https://old.reddit.com/r/awesomewm/comments/9jlsef/my_system_tray_displays_the_icons_somehow_broken/):

![](https://i.redd.it/l97nf2c7hyo11.png)

I'm leaving it here so that others can find it through issue search.

psychon:
Uhh, yeah, that bug... That one would actually be "fixed" by @petoju 's patch.

[...]

PR #2450, opened 2018-10-22
Fix background of transparent systray icons. (#2449)

petoju:
Fixes background, that was not refreshed before and is refreshed and stable with this patch.

### DIFF
--- a/systray.c
+++ b/systray.c
@@ -310,6 +310,13 @@ systray_update(int base_size, bool horizontal, bool reverse, int spacing, bool f
         xcb_configure_window(globalconf.connection, em->win,
                              XCB_CONFIG_WINDOW_X | XCB_CONFIG_WINDOW_Y | XCB_CONFIG_WINDOW_WIDTH | XCB_CONFIG_WINDOW_HEIGHT,
                              config_vals);
+
+        /* Systray icons with transparent background need explicit color.
+         * Provide the one of default background color */
+        uint32_t config_back[] = { globalconf.systray.background_pixel };
+        xcb_change_window_attributes(globalconf.connection,
+                                     em->win,
+                                     XCB_CW_BACK_PIXEL, config_back);
         xcb_map_window(globalconf.connection, em->win);
         if (force_redraw)
             xcb_clear_area(globalconf.connection, 1, em->win, 0, 0, 0, 0);


petoju:
Hmm, so I just noticed this works almost all the time, but it stops working when the applet shows context menu after my rightclick.
I will try to fix that tomorrow.

psychon:
What's the story behind this patch? How did you come up with it and which specs say you should do stuff like this?

According to the specs that I know, systray icons should set their background to `XCB_BACK_PIXMAP_PARENT_RELATIVE`, meaning they get the background of their parent window applied. Awesome then sets the background of this parent window (the systray container window) to whatever is needed in `luaA_systray`. Thus... I would expect that this patch does not make any difference.

Also, your patch says that this provides systray icons with transparent background with the correct background. However, it just changes the background of all systray icons. I bet I could come up with an example program that e.g. just sets a solid green background. Your patch would then break this and make the icon disappear.

psychon:
Also, what did you test this with? Some GTK- or Qt- using app? Something that uses its own systray implementation? What is it specifically?

petoju:
> What's the story behind this patch? How did you come up with it and which specs say you should do stuff like this?

Transparent background of GTK systray icons stopped working after upgrading to newer Ubuntu, so I just tried to find a fix and this one was the obvious one. I don't know XCB nor X specs.

> According to the specs that I know, systray icons should set their background to `XCB_BACK_PIXMAP_PARENT_RELATIVE`, meaning they get the background of their parent window applied. Awesome then sets the background of this parent window (the systray container window) to whatever is needed in `luaA_systray`. Thus... I would expect that this patch does not make any difference.

It makes a difference on this GTK 3.24. It's not perfect, as it stops working after I right-click the affected icons. I know it works otherwise, because once it stops working, I cannot say from icon state, that I got dropped from my wifi network.
I know nothing about how it affects different environments.

> Also, your patch says that this provides systray icons with transparent background with the correct background. However, it just changes the background of all systray icons. I bet I could come up with an example program that e.g. just sets a solid green background. Your patch would then break this and make the icon disappear.

Note again: I don't know the specs. However, if an example program sets green background, won't it happen after we did it, so behaving correctly? In any case, if it breaks anyone, then I am not interested in pushing the patch in.

> Also, what did you test this with? Some GTK- or Qt- using app? Something that uses its own systray implementation? What is it specifically?

Yes, I am using network-manager-gnome as well as xfce4-power-manager. Both depend on GTK 3, so these are GTK apps.

psychon:
Okay, sorry then, but you are just working around a known bug in GTK 3.24. GTK no longer sets its tray icon's background to `ParentRelative`, but to `None`. Thus, it gets `None` and this leads to the behaviour that you see.

Bug is in GTK (as you already figured out in #2449 as well).

n1-cz:
I'm another victim to this :)

@psychon so according to your words the solution is to wait for fixed GTK version or is there anything we can do for that (monkey patch?). My tray is so screwed up I can't recognize the icons anymore.

McSinyx:
@n1-cz you can try to downgrade GTK+ 3 to 3.22.

n1-cz:
I wouls like to but the version is not int apt anymore. Is there any hack how to do that without breaking the whole dependency system in apt?

blueyed:
btw: just restarting awesome also fixes it for the moment.

n1-cz:
Does anyone of you have link to GTK bug tracker which reports this particular issue? Or is it already fixed in newer versions?

psychon:
https://gitlab.gnome.org/GNOME/gtk/issues/1280

n1-cz:
seems like GTK guys advanced...not merged yet but getting close...

Issue #977, opened 2016-06-23
Window rules by delayed name


dopykuh:
Is there any possibilty to setup rules with a regex to match againgst a window name?

psychon:
Uhm... What have you tried already that didn't work? This is just supposed to work (however, Lua's matches aren't really regexes):
```lua
{ rule = { name = ".*e.* }, callback = function(c) print("There is an 'e' in here:", c) end }
```
Internally this uses `string.match`, see e.g. http://lua-users.org/wiki/PatternsTutorial

dopykuh:
There are some websites which are requires popups and i want to match against the windowtitle from chromium. But i am to noob i don't get it to work. Does it work for your? I tried it for OTRS 5 ticket open context.

psychon:
I don't think I am matching against window title anywhere.
How about you add a rule like e.g. `{rule = {}, callback = function(c) naughty.notify{ title="new window", text = c.name } end },` and use this to tell us what exactly is the window title that you want to match against?

actionless:
i guess the real problem here is what browser windows all initially have the same name, it got changed after name being processed by rules

i guess to workaround what you could setup a rule in callback of which you could subscribe to signal when client name changes and only there match its name against your expression

nikanar:
I have the same question, so I'll chip in ; I try
             local matcher = function(c) return awful.rules.match(c, {name = "*Youtube - *"} ) end
         awful.client.run_or_raise('google-chrome-stable --new-window https://www.youtube.com/', matcher)

and it runs a new window, rather than raise, even though there *is* a window that has : 

    ~$ xprop WM_NAME WM_CLASS
    WM_NAME(UTF8_STRING) = "The Awesome Window Manager - YouTube - Google Chrome"
    WM_CLASS(STRING) = "google-chrome", "Google-chrome"

And, same as @actionless, I'm really trying to match browser windows with specific loaded pages, but they take a little while to load and set their `WM_NAME` correctly. I worked around it, but if some clean code to handle that was available, that'd be great.

blueyed:
@nikanar 
`name = ".*Youtube.*"` does not work in Lua.
Just use `Youtube`.
See http://lua-users.org/wiki/PatternsTutorial.

blueyed:
It should also be "YouTube" (capital T)!

blueyed:
And for OP (@dopykuh): I think that it should work, but depending on your browser @actionless might be right.

psychon:
[...]

PR #2451, opened 2018-10-24
Update awesomerc.lua

legion151:
Just a comment in rc.lua - hopefully preventing future people searching for hours.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -507,7 +507,9 @@ awful.rules.rules = {
           "Wpa_gui",
           "veromix",
           "xtightvncviewer"},
-
+            
+        -- Note that the name property shown in xprop might be set slightly after creation of the client
+        -- and the name shown there might not match defined rules here.
         name = {
           "Event Tester",  -- xev.
         },
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -507,7 +507,7 @@ awful.rules.rules = {
           "Wpa_gui",
           "veromix",
           "xtightvncviewer"},
-            
+
         -- Note that the name property shown in xprop might be set slightly after creation of the client
         -- and the name shown there might not match defined rules here.
         name = {


legion151:
Tried to define a rule for "Edit Event" dialog of thunderbird/lightning with a rule like
`name = "Edit Event*"` 
This worked on restarting awesome but not on opening new dialogs.
A person called pppp in irc-chan remembered he had this prob a while a go too, so i thought a comment might help to save some peoples time.

blueyed:
@legion151 
Thanks.

@McSinyx 
What do you mean?

I was thinking that the comment might not be necessary after all, if there is another problem/bug that could be fixed instead.  (since every comment means that someone might read it, although typically not necessary).

actionless:
i'll stick to @Elv13 , it should go to docs, not as comment

Elv13:
I am not against, let's wait for @legion151 a bit more and do that if he/she has no time/will to further develop this concept.

The problem itself isn't new https://github.com/awesomeWM/awesome/issues/977 and more people will eventually hit it.

actionless:
@psychon the problem described in that comment seems to happen only with web pages/apps but idk this to became a bikeshed so whatever :D

legion151:
Sure enough - fixing the underlying problem would be a far more elegant way. Unfortunately I am not that deep into the source. Also, as far as I understand, the problem should hit all apps/applications changing or setting titles dynamically. I guess in those cases the problem cannot be fixed within awesome since the only reasonable point in the life-cycle of a window to check the rules, is on creation. Even if they can be evaluated correctly after that, no user would want the window to be changed or put away automatically after it appeared - or worse, while he is already interacting with it.

After reading the comments i would stick with @Elv13. I am one of those people who never saw that page before - sadly. However that wouldn't solve the problem that people will stumble. My research on the problem did not lead me there. 
So what about linking the documentation in the rc.lua file as a comment?

And rewrite the comment in a more general way?

PR #2454, opened 2018-10-31
Add parameters to awful.widget.prompt constructor.

Aire-One:
Hello!

Lately, I had a few time to work on my awesomewm configs and I tried to improve my prompt widget. I don't have enough time to create my own widget so I just took the existing one on the `awful.widget` library.

To do what I wanted to do, I needed to use the `done_callback` parameter of `awful.prompt` but this parameter is not implemented in `awful.widget.prompt`. So I added it. Because it's an optional parameter of `awful.prompt`, I used `nil` as a default value and marked only `param[opt]` on the ldoc header comment.

I hope this addition can help other people.

PS 1: I'm not sure how to test widgets so I didn't do any "test script" for the addition of this parameter. Only manual testing. If it's require, please point me some documentation or examples I can use to implement tests for this specific case.

PS 2: I only implemented the `done_callback`, but if you care, I can also add the other parameters from `awful.prompt`to make a complete PR on this topic.

### DIFF
--- a/lib/awful/widget/prompt.lua
+++ b/lib/awful/widget/prompt.lua
@@ -39,12 +39,21 @@ local function run(promptbox)
     return prompt.run {
         prompt              = promptbox.prompt,
         textbox             = promptbox.widget,
+        fg_cursor           = promptbox.fg_cursor,
+        bg_cursor           = promptbox.bg_cursor,
+        ul_cursor           = promptbox.ul_cursor,
+        font                = promptbox.font,
+        autoexec            = promptbox.autoexec,
+        highlighter         = promptbox.highlighter,
+        exe_callback        = promptbox.exe_callback,
         completion_callback = promptbox.completion_callback,
-        history_path        = gfs.get_cache_dir() .. "/history",
-        exe_callback        = function (...)
-            promptbox:spawn_and_handle_error(...)
-        end,
-        done_callback = promptbox.done_callback
+        history_path        = promptbox.history_path,
+        history_max         = promptbox.history_max,
+        done_callback       = promptbox.done_callback,
+        changed_callback    = promptbox.changed_callback,
+        keypressed_callback = promptbox.keypressed_callback,
+        keyreleased_callback = promptbox.keyreleased_callback,
+        hook                = promptbox.hook
     }
 end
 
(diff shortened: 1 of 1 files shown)


Aire-One:
I eventually added all the parameters from `awful.prompt`.

Please note :
- I always used `nil` as default for all parameters I added because they are all optional in `awful.prompt` ;
- I moved `history_path` and `exe_callback` assignations from the `run` function to the constructor, to keep the same logic for all parameters ;
- `history_max` is documented as a function in [awful.prompt doc](https://awesomewm.org/apidoc/libraries/awful.prompt.html#run) but is used as an integer in the code, so I documented it as integer in `awful.widget.prompt` (maybe someone should fix the `awful.prompt` doc) ;
- Also, default value of `history_max` is 50 in `awful.prompt` but in my code I used `nil`. (`awful.prompt` would receive `nil` and take its own default (so 50)) ;
- `history_path` default value was `gfs.get_cache_dir() .. "/history"` but `gears.filesystem.get_cache_dir()` return a path with the final slash ([documentation link](https://awesomewm.org/apidoc/libraries/gears.filesystem.html#get_cache_dir)). So I removed this slash from the default value ;
- I also changed the description of `awful.widget.prompt` to explicitly refer to `awful.prompt.run`.

Hope this will help :)

lib/awful/widget/prompt.lua
```diff
@@ -5,6 +5,7 @@
 --
 -- @author Julien Danjou &lt;julien@danjou.info&gt;
 -- @copyright 2009 Julien Danjou
+-- @copyright 2018 Aire-One
```
actionless: @awesomeWM/write-access 1) i was always wondering what the difference between author and copyright in this context
2) if we need that for documentation purpose, couldn't it be done while generating documentation based on file history? currently most of the files having those fields mostly out-of-date

Aire-One: I added myself as a habit but it actually doesn't mean much. I think it can be interpreted as a "I wrote some part of code here".

About the difference between author and copyright, the ldoc documentation does not provide any hint and I am probably wrong in the use.

Anyway, I doesn't care about it and the statement can be changed/removed (I thought I said it in the PR's first message but I did forget...).

Elv13: By law (at least here), the Copyright is assigned to the employer while the author is an human being. Adding `@copyright` is not, by law, necessary since unless contractually specified otherwise, each contributor to an Open Source project is a contractor. Contractors retain the copyright unless contractually specified (CLAs). So each contributor has an implicit right to the module copyright as long as the code is tinted by his or her contribution.

actionless: tl;dr: so it's optional?

Elv13: It's at the contributor discretion

Aire-One:
I'm a little late. I'm happy to see my PR was accepted and merged upstream :) 

About the copyright statement, as I mentioned it earlier, I actually don't really care and I added it as an habit before pushing my local changes to the internet. I just pushed a quick commit to remove it on my branch. Not sure if it still matters now tho...

Anyway, thanks for your time and reviews.

actionless:
no need to remove it, i was asking the guys if other contributors also required to be added or it's optional

PR #2459, opened 2018-11-01
feat(awful: titlebar): expose fallback title

actionless:


actionless:
@psychon  a lot of dialog windows have ugly for my taste `<unknown>` title

actionless:
i think it's not good idea setting client's name instead of overriding the default value of in titlebar implementation

i'll update ldoc now

actionless:
@Elv13 i've made the change you suggested but i don't think it looks better now:

before:
2018-12-17--1545064043_953x178_scrot
2018-12-17--1545064056_770x275_scrot





after:
2018-12-17--1545064031_673x195_scrot
2018-12-17--1545064020_908x386_scrot

Elv13:
The first one is correct, the second one isn't, right?

actionless:
the second is using `@field`, the last commit

Elv13:
I pushed a proposed fix

actionless:
i see, however `opt` is not parsed for some reason:

2018-12-17--1545066187_683x345_scrot

Elv13:
yeah, that would be an `ldoc` bug... I am fine with this for now given it is technically rendering the fields correctly. Squash merge if you agree.

actionless:
:+1:

PR #2460, opened 2018-11-04
Add 3 x middle-mouse-button actions

Sorky:
1) Middle-mouse-click on client in the tasklist to close it

2) Middle-mouse-click on the layout selector to revert to the first layout

3) Middle-mouse-click on client titlebar to close it

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -202,7 +202,7 @@ awful.screen.connect_for_each_screen(function(s)
     s.mylayoutbox = awful.widget.layoutbox(s)
     s.mylayoutbox:buttons(gears.table.join(
                            awful.button({ }, 1, function () awful.layout.inc( 1) end),
-                           awful.button({ }, 2, function () awful.layout.set( awful.layout.layouts[ 1 ] ) end),
+                           awful.button({ }, 2, function () awful.layout.set( awful.layout.layouts[1] ) end),
                            awful.button({ }, 3, function () awful.layout.inc(-1) end),
                            awful.button({ }, 4, function () awful.layout.inc( 1) end),
                            awful.button({ }, 5, function () awful.layout.inc(-1) end)))
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -162,7 +162,7 @@ local tasklist_buttons = gears.table.join(
                                                   )
                                               end
                                           end),
-                     awful.button({ }, 2, function( c ) c:kill() end),
+                     awful.button({ }, 2, function(c) c:kill() end),
                      awful.button({ }, 3, client_menu_toggle_fn()),
                      awful.button({ }, 4, function ()
                                               awful.client.focus.byidx(1)
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -133,25 +133,25 @@ mytextclock = wibox.widget.textclock()
 -- Create a wibox for each screen and add it
 -- @TAGLIST_BUTTON@
 local taglist_buttons = gears.table.join(
-                    awful.button({ }, 1, function(t) t:view_only() end),
+                    awful.button({}, 1, function(t) t:view_only() end),
                     awful.button({ modkey }, 1, function(t)
                                               if client.focus then
                                                   client.focus:move_to_tag(t)
                                               end
                                           end),
-                    awful.button({ }, 3, awful.tag.viewtoggle),
+                    awful.button({}, 3, awful.tag.viewtoggle),
                     awful.button({ modkey }, 3, function(t)
                                               if client.focus then
                                                   client.focus:toggle_tag(t)
                                               end
                                           end),
-                    awful.button({ }, 4, function(t) awful.tag.viewnext(t.screen) end),
-                    awful.button({ }, 5, function(t) awful.tag.viewprev(t.screen) end)
+                    awful.button({}, 4, function(t) awful.tag.viewnext(t.screen) end),
+                    awful.button({}, 5, function(t) awful.tag.viewprev(t.screen) end)
                 )
 
 -- @TASKLIST_BUTTON@
 local tasklist_buttons = gears.table.join(
-                     awful.button({ }, 1, function (c)
+                     awful.button({}, 1, function (c)
                                               if c == client.focus then
                                                   c.minimized = true
                                               else
@@ -162,12 +162,12 @@ local tasklist_buttons = gears.table.join(
                                                   )
                                               end
                                           end),
-                     awful.button({ }, 2, function(c) c:kill() end),
-                     awful.button({ }, 3, client_menu_toggle_fn()),
-                     awful.button({ }, 4, function ()
+                     awful.button({}, 2, function(c) c:kill() end),
+                     awful.button({}, 3, client_menu_toggle_fn()),
+                     awful.button({}, 4, function ()
                                               awful.client.focus.byidx(1)
                                           end),
-                     awful.button({ }, 5, function ()
+                     awful.button({}, 5, function ()
                                               awful.client.focus.byidx(-1)
                                           end))
 
@@ -201,11 +201,11 @@ awful.screen.connect_for_each_screen(function(s)
     -- We need one layoutbox per screen.
     s.mylayoutbox = awful.widget.layoutbox(s)
     s.mylayoutbox:buttons(gears.table.join(
-                           awful.button({ }, 1, function () awful.layout.inc( 1) end),
-                           awful.button({ }, 2, function () awful.layout.set( awful.layout.layouts[1] ) end),
-                           awful.button({ }, 3, function () awful.layout.inc(-1) end),
-                           awful.button({ }, 4, function () awful.layout.inc( 1) end),
-                           awful.button({ }, 5, function () awful.layout.inc(-1) end)))
+                           awful.button({}, 1, function () awful.layout.inc( 1) end),
+                           awful.button({}, 2, function () awful.layout.set( awful.layout.layouts[1] ) end),
+                           awful.button({}, 3, function () awful.layout.inc(-1) end),
+                           awful.button({}, 4, function () awful.layout.inc( 1) end),
+                           awful.button({}, 5, function () awful.layout.inc(-1) end)))
     -- Create a taglist widget
     s.mytaglist = awful.widget.taglist {
         screen  = s,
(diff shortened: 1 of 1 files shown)


McSinyx:
> 3. Middle-mouse-click on client titlebar to close it

This is odd. Is there any conversation leading to this PR?

Sorky:
> > 1. Middle-mouse-click on client titlebar to close it
> 
> This is odd. Is there any conversation leading to this PR?

I didn't start a conversation as it seemed like a no-brainer [no click currently applied and middle-click to close tab behaviour is quite common in many apps [firefox, notepad++ for example]

Elv13:
I am not sure about this. IMO, this isn't something many people might find to be an improvements. Plus on some touchpads, middle click is easy to trigger by accident because there is some "virtual buttons area" by default. This could lead to data loss.

On the plus side, I can see why I would want this. So maybe this should be moved into the documentation or a wiki recipe instead of the default config? I will let the others weight in on this, I am unconvinced, but not against either.

Also, most of the change in the commit mess with the formatting. Please don't change the formatting in unrelated commits (note: I am neutral about the formatting changes, but it should at least have a reasoning that makes sense and be in its own commit).

Sorky:
> On the plus side, I can see why I would want this. So maybe this should be moved into the documentation or a wiki recipe instead of the default config? I will let the others weight in on this, I am unconvinced, but not against either.

I figure it is worth having in the default - if it becomes an issue for tablet users they can always remove or comment out that click action 

> Also, most of the change in the commit mess with the formatting. Please don't change the formatting in unrelated commits (note: I am neutral about the formatting changes, but it should at least have a reasoning that makes sense and be in its own commit).

Fair call - I thought the same after doing it - Happy to drop that last commit

PS: I probably should have done the layout as a separate pull as well - Happy to scrap this one and re-do three separate ones

McSinyx:
> I figure it is worth having in the default - if it becomes an issue for tablet users they can always remove or comment out that click action

Please, no, do expect new comers to accidentally close windows multiple times to figure out it's not random crashing? Notice that this change does not affect current users who quite have a sense about what their config does, but only scares away newbies.

BTW in most WM I've used, mouse2 is used to either minimize or shade. Again, closing is very unexpected.

Sorky:
BTW - "kill" is not as bad as it sounds - It doesn't close applications with unsaved changes - They popup a dialog that can continue or cancel the close

actionless:
i am agree with Elv13 and McSinyx
and also i think middle click on layoutbox could be used for something more useful, like:

```lua

    local layouts_menu_items = {}
    for _, layout in ipairs(awful.layout.layouts) do
      table.insert(layouts_menu_items, {
        layout.name,
        function()
            awful.layout.set(layout, layoutbox.screen.selected_tag)
        end,
        beautiful["layout_"..layout.name]
      })
    end
    layoutbox.menu = awful.menu({
        items = layouts_menu_items,
    })

    layoutbox.widget:buttons(awful.util.table.join(
      awful.button({ }, 2, function ()
        layoutbox.menu_id = layoutbox.menu:toggle()
      end),
```

(however for new per-tag layouts this example will require some rewrite to use `t.layouts` instead of `awful.layout.layouts`)

Elv13:
Sorry about how it ended. We like contributions a lot, but anything that adds something to `rc.lua` is going to have a tough time. It's the part that's most sensitive and we are very conservative when it comes to new features there. Anything that may cause confusion to new users isn't going to get in. That's in part why the behavior of rc.lua have not changed in nearly a decade. It looks outdated and we have wonderful features not enabled by default. That being said, it is very predictable and boring, which for defaults is a good thing.

Sorky:
Not a problem - I use it and figured it was simple enough to share

If I was to make one suggestion for the future, it would be to make rc.lua a skeleton with lots of separate pieces that can easily be plugged in like lego - Much like adding widgets - The monumental blob that it is can be rather daunting as to where to start on modifications

When I finish re-factoring my entire setup I'll put it in a repo

actionless:
here's a thread for sharing users' configurations: https://github.com/awesomeWM/awesome/issues/1395

psychon:
> it would be to make rc.lua a skeleton with lots of separate pieces that can easily be plugged in like lego

That's basically #442 (without lots of activity in the last few years, but the idea is there...)

Sorky:
Ahh to dream... 

Screenshots aren't of much interest to me - The components and the code separation is though!

I've already split my rc.lua and theme.lua into a number of separate files - Makes it easier to learn even having to deal with the the odd dead session - Good backup procedures are a must!

Once separated. mistakes are less critical. eg: massaging widgets doesn't kill anything major and is easily recoverable without a sandbox

There's another reason - not having to do a sandbox - think about how easy it could be for new users

actionless:
> Good backup procedures are a must!

`cd ~/.config/awesome && git init` ;-)


regarding splitting the config: that's good idea, the only problem is how to pass objects (for example how to reference the widget in hotkeys section) between config parts

for example, in my config i solved it by having `context` table which is passed into each config part as in factory: https://github.com/actionless/awesome_config/blob/devel/rc.lua#L141-L157

other approach could be using global variables but it just feels so bad to me

Elv13:
IMO, as said a bunch of time before but to save everybody the trouble of reading everything, we should just move the wallpaper/wibar/titlebar/notification to the theme and call it a day. This in itself would solve most of the problem and be good enough. Because the wibar and titlebars are mostly visual stuff, then moving them to the theme where mostly visual stuff is handled solves many problems.

Of course it's not perfect because things like the tasklist button get sucked into the theme. However if you see this from another angle it makes sense. Buttons only make sense on visual stuff, so they are tied to the theme anyway.

Cutting `rc.lua` into too many pieces will make things worst. The problem isn't that rc.lua is too large, the problem is that Awesome does many things and they need an entry point. Making a million config file fix a non-issue while making the real issue worst.

actionless:
i still think what theme should stay as just key-value database from which user config or different third-party widgets are just taking those values

mb we need to separate the terms of `theme table` and `theme config`

Elv13:
> mb we need to separate the terms of theme table and theme config

I am not opposed to doing this. However, what would be the point of it beside adding a third file?

actionless:
but still that `theme config` approach will create some new challenges to solve, like custom panel widgets

such `theme config`s should reserve some user-customizable sections (ie in rc.lua user can define
 ```lua
awful.theme_config.panel_custom_widgets = [.....]
awful.theme_config.load('my_theme_config.lua')
```

otherwise when switching between themes user will have to copy the theme itself and still do modifications in both rc.lua and theme

Elv13:
I am talking about removing the wibar from rc.lua entirely. It would no longer include `wibox` at all. The wallpaper, wibox, titlebar (and eventual notification if my code get merged) sections would be removed from rc.lua and moved to the default and xresources themes. Then you become free to implement your custom bar for the GTK theme. The sky theme can probably be deprecated and staged for removal in 5.0.

actionless:
my point what if user would like to add some panel widget, like https://awesomewm.org/recipes/mpc/ for example or https://github.com/stefano-m/awesome-power_widget, it will require modifying the theme to add it to the wibar

actionless:
so we'll need to have kind of "systray" for awesome widgets, so custom themes could put such "systray" there user config could put custom widgets inside

actionless:
and i think we'll need at least 3 such areas, like, left, right, center

Elv13:
> it will require modifying the theme to add it to the wibar

Yes, it will. If we split the config into more files, it will also force the user to edit more files. In this case, changing how Awesome looks, including widgets, will require to edit the theme. It is a feature, not a bug.

actionless:
i think those 3 custom areas will solve a lot of pain for potentials users because it will allow to quickly switch totally different themes while preserving beloved widgets

Elv13:
Not really. If someone wants to have 2 bars like in Ubuntu flavored Gnome2, then would it need 6 special widget area? If someone wants bar on each side, then should we reserve 12 special areas? What about multi-screen? What about a macOS like dock?

I say if the user wants to refactor her/his config so he/she is able to change the theme frequently, they are free to do so. Having special widgets area is a leaky abstraction. It only scale until the user want to change the number of type of bar. So if the issue behind #442 is "all Awesome look the same", then what you propose prevents it from being solved.

actionless:
as i told:

> and will still leave user a possibility to remove those 3 "systrays" from the theme and arrange widgets manually

which covers most of your questions

the number 3 came up as just minimal reasonable amount

and also we need to have some way how those custom widgets could be referenced inside rc.lua, for example in keybindings if user want to bind some key to affect the behavior of some of custom panel widgets

Elv13:
I disagree with this reasoning. Either widgets belong to rc.lua or they don't. There is no middle ground. Any middle ground is just arbitrary and as per the example I provided above, don't make sense.

As for the widgets interacting with the keybinding, this is a valid concern, but it still couples the "logic" with how Awesome looks. So if the user want to expose some widgets in the `theme` table, then they are free do so. However I would divide everything on the "widgets no longer belong in rc.lua" line.

That way, the `beautiful` table becomes the interface between the view and the behavior. `rc.lua` controls the behavior and `theme.lua` controls the view.

Keep in mind that it's only the default and the goal of doing this is making themes more flexible in deeply changing how Awesome looks without having to monkey-patch `rc.lua`. If you or any other user want to draw the separation of concerns line elsewhere, you are free to do so. As I mentioned above, introducing a leaky abstractions with exceptions isn't improving the situation, it just moves the problem in a way that improve toward your vision of a nice looking desktop, but make other users experience worst (anyone who use a different bar configuration).

Sorky:
> but still that `theme config` approach will create some new challenges to solve, like custom panel widgets
> 
> such `theme config`s should reserve some user-customizable sections, ie
 
Also needs to allow for the fact that multiple screens have differences in their wibars

blueyed:
(just for info, only noticed afterwards that this is an older issue already)

> BTW - "kill" is not as bad as it sounds - It doesn't close applications with unsaved changes - They popup a dialog that can continue or cancel the close

While this might be true in general for GUI apps, it is not for terminals for example.

I am -1 on killing clients easily.
(I am using some xproperty for my terminals to ask for confirmation there for alt-f4 / close even to avoid data loss)

actionless:
@blueyed i am just using tmux to avoid data loss if closing terminal accidentally or if X session died (tmux is started from init, not from within x session)

Sorky:
> While this might be true in general for GUI apps, it is not for terminals for example.

Feedback noted - I assume that is relating to editors and unsaved changes

I had a bit of a play and I see what you mean...

Using "xterm -hold" blocks the kill **sometimes** [ie: doing 'sudo apt-get update' doesn't die from a kill signal but doing 'vim file.txt' . vim exits immediately from a kill signal (and a second kill signal is needed for xterm itself)

Issue #2461, opened 2018-11-04
Weird border between widgets if dpi is not dividable by 96


actionless:
Output of `awesome --version`: `awesome v4.2-498-gb6b6bc0b-dirty`

**How to reproduce the issue:**

Set dpi to not dividable by 96

and use this widget:

```lua
function common.text_progressbar(args)
  local widget = wibox.widget {
    {
      {
          id = "progressbar",
          max_value     = 1,
          forced_width  = dpi(20),
          margins      = {
            top=beautiful.panel_height - dpi(8),
            bottom=dpi(3),
          },
          color = args.progress_fg or beautiful.bg_focus,
          background_color = args.progress_bg or beautiful.bg_normal,
          widget        = wibox.widget.progressbar,
      },
      id = "p1",
      bg = args.bg or beautiful.panel_widget_bg or "#00000000",
      layout = wibox.container.background,
    },
    {
      {
        nil,
        {
            id = "textbox",
            widget = wibox.widget.textbox,
        },
        nil,
        id = "t1",
        layout = wibox.layout.align.horizontal,
      },
      id = "t1",
      left = dpi(4),
      right = dpi(4),
      bottom = dpi(5),
      layout = wibox.container.margin,
    },
    layout = wibox.layout.stack
  }
  widget.textbox = widget.t1.t1.textbox
  widget.progressbar = widget.p1.progressbar
  return widget
end
```

**Actual result:**

![2018-11-04--1541340574_1920x1080_scrot](https://user-images.githubusercontent.com/1655669/47965179-2c464200-e044-11e8-9840-26c5bf8cc3f9.png)

**Expected result:**

![2018-11-04--1541340583_1920x1080_scrot](https://user-images.githubusercontent.com/1655669/47965180-323c2300-e044-11e8-9230-257b5bed5d0d.png)

psychon:
Uhm... could you turn this into a self-contained code example? From a quick first look `dpi` and `beautiful.panel_height` cause problems.
Sorry.

actionless:
i was already checking dpi output -- they all getting rounded so it's not the issue, i'll now readjust code example as a patch to default config

actionless:
hm, i am still struggling extracting stand alone example -- unfortunately it seems it caused by some specific combination of horizontal and vertical layouts and constraint widgets

psychon:
Where does the shape come from? The code example does not seem to set any. Also, is there more than one widget with a background? Again, the code example only has one widget with a background.

[...]

PR #2463, opened 2018-11-05
wibox.layout.flex: Do not leave empty space behind

psychon:
No idea if this is a good thing to do or not. Assigning difference sizes to the individual widgets could lead to other problems, but I don't want to just add "lots and lots" of options "just because"...

Also, this patch makes non-integer spacing lead to non-integer widget positions. Is that something people actually do and care about...?

@actionless Could you test this?

----

It is not possible to distribute 100px to three widgets equally. The
current version of wibox.layout.flex tries to do that anyway, by giving
each widget 33px and leaving one pixel outside of any widget. Thus, if
the widgets e.g. have a common background, this leads to a one pixel gap
in the background.

This patch changes the flex layout so that the extra pixel is assigned
to some widget instead. It does so by basically keeping a sum of
space_per_item for the widgets that was assigned so far. This sum is
rounded and when this leads to rounding, the corresponding child widget
gets an extra pixel.

More precisely, this tracks a pos as before. Widgets get their position
still assigned based on rounding pos. However, this now also remembers
this rounded position for the next iteration of the loop. This allows to
assign the size of widgets based on the difference between the current
and last rounded position.

(Possibly) fixes: https://github.com/awesomeWM/awesome/issues/2461

### DIFF
--- a/lib/wibox/layout/flex.lua
+++ b/lib/wibox/layout/flex.lua
@@ -10,7 +10,6 @@ local base = require("wibox.widget.base")
 local fixed = require("wibox.layout.fixed")
 local table = table
 local pairs = pairs
-local floor = math.floor
 local gmath = require("gears.math")
 local gtable = require("gears.table")
 
@@ -68,7 +67,7 @@ local flex = {}
 
 function flex:layout(_, width, height)
     local result = {}
-    local pos,spacing = 0, self._private.spacing
+    local spacing = self._private.spacing
     local num = #self._private.widgets
     local total_spacing = (spacing*(num-1))
     local spacing_widget = self._private.spacing_widget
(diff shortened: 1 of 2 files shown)


actionless:
thanks! that seems to solve the testcase with flexboxes, but i still have some weird rendering issue to investigate (i'll upload screenshot to the new ticket)

actionless:
is mergify still working?

Elv13:
@jd ^

jd:
@Elv13 @actionless typo, see https://github.com/awesomeWM/awesome/pull/2466

blueyed:
@jd 
Doesn't pick up the merged fix here apparently.

jd:
It just needed a new event to do its magic, I've approved the patch :)

actionless:
@jd thanks!

PR #2464, opened 2018-11-05
Create globalkeys early

Sorky:
Allows for keys to be added in widgets created in a theme in beautiful (as per copycatz)

Consistent with line 409 (now 412)

A small step towards making rc,lua more modular

Can't see how it can hurt anyone

Personally I prefer variables to be defined/created before use (but that's just me)

Opens the possibility of splitting the huge key definition section into chunks with areas of impact

Bring on the debate ;-)

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -45,6 +45,9 @@ do
 end
 -- }}}
 
+-- Define globalkeys table for later use
+globalkeys = {}
+
 -- {{{ Variable definitions
 -- @DOC_LOAD_THEME@
 -- Themes define colours, icons, font and wallpapers.
@@ -255,7 +258,7 @@ root.buttons(gears.table.join(
 
 -- {{{ Key bindings
 -- @DOC_GLOBAL_KEYBINDINGS@
-globalkeys = gears.table.join(
+globalkeys = gears.table.join(globalkeys,
     awful.key({ modkey,           }, "s",      hotkeys_popup.show_help,
               {description="show help", group="awesome"}),
     awful.key({ modkey,           }, "Left",   awful.tag.viewprev,


actionless:
i don't like neither relying on global vars (it would be nice to get rid of them in the rc.lua in next major release btw) neither idea of defining some hotkeys from the theme

Elv13:
I am not saying the delayed call workaround is nice: it isn't. But your idea promotes relying on global variables from the theme. This isn't ok and the globals will be removed in the next version and it will break your hack. The order of init and the dependency graph of Awesome modules is the way it is on purpose. The fact that no Awesome module uses the globals set in rc.lua is a feature. You break this. Sorry but no, just no.

Sorky:
> The fact that no Awesome module uses the globals set in rc.lua is a feature.

That I understand, but while the intent is true, there are examples out there that do and while they don't follow "the rules", they exist nonetheless

The huge lack of other examples or better explanations leave inexperienced users exposed

Closing another... And just for the record, the commit itself does NOT actualy break anything it just makes it easier and safer if bending "the rules"

actionless:
@Sorky 

i think Elv13's point is not about the rules but about deprecation of such behavior in the next releases:

if some library or config will be relying on such globals it will break as soon as globals will be removed from rc.lua, and creating more ways to rely on them will break more things in the future

actionless:
there is no punishment policy for unmerged PRs' authors ;-)

regarding release, see milestones (there is no fixed release cycle so it depends only when it's ready):

https://github.com/awesomeWM/awesome/milestones/v4.3

https://github.com/awesomeWM/awesome/milestones/v5


also here's that PR which will break configs/libs which are relying on rc.lua globals:
https://github.com/awesomeWM/awesome/pull/2199

Sorky:
Got a giggle from PR #2199 - Arguing to keep the globals ;-) 

If it wasn't for your "creating more ways to rely on them will break more things in the future" I'd have thought that me trying to make globalkeys more global than it already is would have been a no-brainer

Elv13:
The problem exists, I don't think anybody here is happy with the way global keys are implemented. Our/My point is that it isn't the solution. We need a better, more atomic way to add, remove and toggle keybindings instead of a global array. We also need stronger introspection to prevent multiple keybindings on the same combination by accident.

As for the module dependency, we enforce them. https://github.com/awesomeWM/awesome/blob/master/build-utils/check_for_invalid_requires.lua It just happen that our tooling disables "forbid all globals" on rc.lua for legacy reasons and the depgraph didn't fail because either it's too dumb to catch the global or, in this case, you did not make use of the `rc.lua` content in the theme, so the tripwire wasn't activated.

Sorky:
> the depgraph didn't fail because either it's too dumb to catch the global or, in this case, you did not make use of the `rc.lua` content in the theme, so the tripwire wasn't activated

There was no theme to trip the test ;-) I'm guessing you actually only ready my comment and didn't actually look at the code change I proposed...

<straight after the error trapping/before beautiful>
    -- Define globalkeys table for later use
    globalkeys = {}
<first instance of globalkeys>
<that's all of it>

Hence my insistence that I didn't break anything ;-)

Thanks for the discussion - Learning more and discovering more changes needed in my config every day

Elv13:
I did read the code, that's the "in this case" part of the comment. As @actionless explains, this is more about not allowing users to shoot themselves in the foot with things that will eventually break or are bad practices. Using global, and having global, is one of them. That said, as in #2199, I am uneasy when it come to any change in `rc.lua` that can cause breakages. Either now (modules being broken because of removing the globals) or in the future (people config breaking because we remove the globals).

Maintaining Awesome isn't all about fixing bugs and adding new features, it's also making sure as many existing (and future) users as possible have a smooth ride when updating. Making it easier to access the globals just adds fragility and bad practices. It would be detrimental to the project.

If the "we need a better way to handle global keybinginds" problem is an issue for you, then the only proper way to fix it is by making the API better in a backward compatible way and having unit and integration tests for everything.

Sorky:
I miss-read that

Looking into changing my keys along the lines of your example https://github.com/Elv13/tyrannical/blob/master/shortcut.lua#L135

Why a delayed call and not just using the code inside register_keys mostly as is?

Oh and I've got a little code improvement to offer for menubar,utils - how open are you to a change there?

Elv13:
> Why a delayed call and not just using the code inside register_keys mostly as is?

Because if modules are `require()`ed before the keys are set they will be overwritten. It's ugly, but right now it's the only reliable way.

> Oh and I've got a little code improvement to offer for menubar,utils - how open are you to a change there?

menubar needs a lot of love. The more love it gets the better. As you saw, we are quite strict about changes that break the API or the behavior of just about everything. If it's an option or at least is idiot proof enough so there is no way anyone will ever complain about it, then it should be fine. Well tested code is preferred, but menubar is currently the least tested module, so full coverage isn't required for it yet.

Sorky:
Argh - Just pulled menubar/utils.lua and it's a lot different = Here's me...

 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.2

What version is it I'd have looked at editing?

More work to do!

psychon:
> What version is it I'd have looked at editing?

Sorry, what exactly are you asking? If you want to submit a pull request to awesome, you should change `menubar.utils` from awesome git.
Yes, there seems to have been quite some changes since v4.2 to this file:
```
$ git diff --stat v4.2.. lib/menubar/utils.lua
 lib/menubar/utils.lua | 285 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-------------------------------------------
 1 file changed, 209 insertions(+), 76 deletions(-)
```

Issue #2437, opened 2018-10-15
Add client move global by direction


darrenswhite:
Similar to how `awful.client.focus.global_bydirection` and `awful.client.swap.global_bydirection` work except for moving a client in a given direction. This would be nice to have if possible! Thanks.

Elv13:
Here you go
https://github.com/elv13/collision

actionless:
btw it seems what `awful.client.move` is not documented

actionless:
i think it could be done as `client:move_to_screen_bydirection` as a counterpart to existing `client:move_to_screen` but i am running single-screen setup now so can't play with it

Elv13:
In that case it would be `swap_by_direction`, but it has some issues and isn't such a good idea because of floating vs. tiled clients.

If you implement a trivial swap by direction and the next client is floating and the origin isn't, you only have bad choice. Either you blindly swap them and it doesn't work. Either you make the target tiled and the origin floating, in which case you probably made your calculator or popup tiled by accident or you do nothing and it's confusing.

It's better to keep separate logic for floating and tiled clients when it comes to swapping. So a global method on the client isn't gonna work.

actionless:
but move_to_screen is not swap, move means what the client just goes away from this screen to some other

Elv13:
Ah, yeah, we could have that one. I always wanted to add move iterators and navigation methods to the screens to handle geometric navigation and iteration (from left to right, right to left, top to bottom, bottom to top) but I never did.

actionless:
but such kind of code should already reside in `*_global_bydirection` helpers, so mb it could be extracted and re-used?

Elv13:
It is in `gears.geometry.rectangle.get_in_direction`. However given the rarely changing screen layouts, it may be worth it to cache the result and daisy-chain the screens (internally, in Lua). Then add `myscreen.left/right/up/down` properties. However, as I said, it's only safe to do this for screens. Client and tags have many corner case that makes such accessors useless in practice.

darrenswhite:
I think `client:move_to_screen_bydirection` which @actionless suggested would work for me.

ngoonee:
Any updates on this? Just spent half an hour to implement this behaviour in my crappy lua:-

    awful.key({ modkey, "Shift"   }, "u",
              function (c)
                  local geo = c.screen.geometry
                  if geo.x > 0 then
                      c:move_to_screen(c.screen.index-1)
                  end
              end,
[...]

PR #2468, opened 2018-11-07
lib/menubar/utils.lua - minor improvement

Sorky:
Changed icon_formats so as to make "function is_format_supported" obsolete

Expect there will be a minor performance improvement

Code line count reduced

Code change is minimal so hopefully only needs a visual review

I have tested the changes on my revision which is AOK and I'm 99.99% sure they match

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -267,16 +267,17 @@ function utils.lookup_icon_uncached(icon_file)
         return false
     end
 
-    local icon_file_ext = icon_file:gsub(".*%.","")
+    local icon_file_ext = icon_file:match(".*%.(.-)$")
     if icon_file:sub(1, 1) == '/' and supported_icon_formats[icon_file_ext] then
         -- If the path to the icon is absolute and its format is
         -- supported, do not perform a lookup.
         return gfs.file_readable(icon_file) and icon_file or nil
     else
         for _, directory in ipairs(get_icon_lookup_path()) do
+            local possible_file = directory .. "/" .. icon_file
             if supported_icon_formats[icon_file_ext] and
-                    gfs.file_readable(directory .. "/" .. icon_file) then
-                return directory .. "/" .. icon_file
+                    gfs.file_readable(possible_file) then
+                return possible_file
             else
                 -- Icon is probably specified without path and format,
                 -- like 'firefox'. Try to add supported extensions to
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -138,19 +138,7 @@ local all_icon_sizes = {
 }
 
 --- List of supported icon formats.
-local icon_formats = { "png", "xpm", "svg" }
-
---- Check whether the icon format is supported.
--- @param icon_file Filename of the icon.
--- @return true if format is supported, false otherwise.
-local function is_format_supported(icon_file)
-    for _, f in ipairs(icon_formats) do
-        if icon_file:match('%.' .. f) then
-            return true
-        end
-    end
-    return false
-end
+local supported_icon_formats = { ["png"] = 1, ["xpm"] = 2, ["svg"] = 3 }
 
 local icon_lookup_path = nil
 --- Get a list of icon lookup paths.
@@ -279,20 +267,21 @@ function utils.lookup_icon_uncached(icon_file)
         return false
     end
 
-    if icon_file:sub(1, 1) == '/' and is_format_supported(icon_file) then
+    local icon_file_ext = icon_file:gsub(".*%.","")
+    if icon_file:sub(1, 1) == '/' and supported_icon_formats[icon_file_ext] then
         -- If the path to the icon is absolute and its format is
         -- supported, do not perform a lookup.
         return gfs.file_readable(icon_file) and icon_file or nil
     else
         for _, directory in ipairs(get_icon_lookup_path()) do
-            if is_format_supported(icon_file) and
+            if supported_icon_formats[icon_file_ext] and
                     gfs.file_readable(directory .. "/" .. icon_file) then
                 return directory .. "/" .. icon_file
             else
                 -- Icon is probably specified without path and format,
                 -- like 'firefox'. Try to add supported extensions to
                 -- it and see if such file exists.
-                for _, format in ipairs(icon_formats) do
+                for format, _ in pairs(supported_icon_formats) do
                     local possible_file = directory .. "/" .. icon_file .. "." .. format
                     if gfs.file_readable(possible_file) then
                         return possible_file
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -274,16 +274,16 @@ function utils.lookup_icon_uncached(icon_file)
         return gfs.file_readable(icon_file) and icon_file or nil
     else
         for _, directory in ipairs(get_icon_lookup_path()) do
-            local possible_file = directory .. "/" .. icon_file
+            local directory_file = directory .. "/" .. icon_file
             if supported_icon_formats[icon_file_ext] and
-                    gfs.file_readable(possible_file) then
-                return possible_file
+                    gfs.file_readable(directory_file) then
+                return directory_file
             else
                 -- Icon is probably specified without path and format,
                 -- like 'firefox'. Try to add supported extensions to
                 -- it and see if such file exists.
                 for format, _ in pairs(supported_icon_formats) do
-                    local possible_file = directory .. "/" .. icon_file .. "." .. format
+                    local possible_file = directory_file .. "." .. format
                     if gfs.file_readable(possible_file) then
                         return possible_file
                     end
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -138,7 +138,7 @@ local all_icon_sizes = {
 }
 
 --- List of supported icon formats.
-local supported_icon_formats = { ["png"] = 1, ["xpm"] = 2, ["svg"] = 3 }
+local supported_icon_formats = { png = 1, xpm = 2, svg = 3 }
 
 local icon_lookup_path = nil
 --- Get a list of icon lookup paths.


lib/menubar/utils.lua
```diff
@@ -279,20 +267,21 @@ function utils.lookup_icon_uncached(icon_file)
         return false
     end
 
-    if icon_file:sub(1, 1) == '/' and is_format_supported(icon_file) then
+    local icon_file_ext = icon_file:gsub(".*%.","")
```
psychon: Uhm. This does not behave in the same way as the old code. Previously a file called "png" was rejected:
```
$ lua -e 'print(("png"):match("%.png"))'
nil
```
now it is accepted:
```
$ lua -e 's = {png=1} print(s[("png"):gsub(".*%.","")])'
1
```

Sorky: The old way used the function I removed...
```
icon_formats = { "png", "xpm", "svg" }

icon_file = "file.png"

local function is_format_supported(icon_file)
    for _, f in ipairs(icon_formats) do            -- May try all of these { "png", "xpm", "svg" }
        if icon_file:match('%.' .. f) then         -- "file.png":match(".png")
            return true
        end
    end
    return false
end

if is_format_supported("file.png") then print("Match!") end

> Match!
```

The new way simply uses the extension as an index (is true simply if it exists)...
```
supported_icon_formats = { ["png"] = 1, ["xpm"] = -1, ["svg"] = 'a' }

icon_file = "file.png"

icon_file_ext = icon_file:gsub(".*%.","")                            -- icon_file_ext = "file.png":gsub("file.","") = "png"
if supported_icon_formats[icon_file_ext] then print("Match!") end    -- supported_icon_formats["png"] = 1

> Match!
```

Tested here... https://www.lua.org/cgi-bin/demo

psychon: What I am trying to say: If you change `file.png` to `png`, then the old code will not print anything, but the new one will still print `Match!`. So, there is a change in behaviour here.

psychon:
> Expect there will be a minor performance improvement

Sigh. Okay, since you ask for it: Ad-hoc, "weird" micro-benchmark:
```lua
icon_formats = { "png", "xpm", "svg" }
icon_file = "file.png"
local function is_format_supported(icon_file)
    for _, f in ipairs(icon_formats) do
        if icon_file:match('%.' .. f) then
            return true
        end
    end
    return false
end

for i = 0, 2^22 do
	local value = is_format_supported("file.png")
end
```
```lua
supported_icon_formats = { ["png"] = 1, ["xpm"] = -1, ["svg"] = 'a' }
icon_file = "file.png"
for i = 0, 2^22 do
	local icon_file_ext = icon_file:gsub(".*%.","")
	local value = supported_icon_formats[icon_file_ext]
end
```
Benchmarks were done with the `time` tool:
```
$ time lua old.lua
lua old.lua  1,24s user 0,00s system 99% cpu 1,243 total
$ time lua new.lua
lua new.lua  1,27s user 0,00s system 99% cpu 1,270 total
```
So, the new approach is actually slightly slower than the old one.

The above was done with Lua 5.3.4. Here is the same code with LuaJIT 2.1.0-beta3:
```
$ time luajit old.lua
luajit old.lua  0,44s user 0,00s system 99% cpu 0,440 total
$ time luajit new.lua
luajit test2.lua  1,02s user 0,00s system 99% cpu 1,020 total
```
The new code takes more than two times as much time as the old code. No idea if this is due to the silly benchmark and LuaJIT manages to optimise away too much.

Sorky:
Presume the benchmark was with a png file - now with an svg?

psychon:
> Presume the benchmark was with a png file

I just ran the actual code that I provided above. So yes, this was checking `file.png`.

> now with an svg?

I just changed `file.png` to `file.svg` in both files:
```
lua old.lua  1,26s user 0,00s system 99% cpu 1,258 total
lua new.lua  1,31s user 0,00s system 99% cpu 1,318 total
luajit old.lua  0,47s user 0,00s system 99% cpu 0,470 total
luajit new.lua  1,03s user 0,00s system 99% cpu 1,032 total
```
Note that the times here a bit arbitrary. Among the couple of results that I got was also this one:
```
lua old.lua  1,31s user 0,00s system 99% cpu 1,312 total
lua new.lua  1,29s user 0,00s system 99% cpu 1,289 total
luajit old.lua  0,46s user 0,00s system 99% cpu 0,458 total
luajit new.lua  1,00s user 0,00s system 99% cpu 1,005 total
```
(But most of the time, `lua old.lua` is faster than `lua new.lua`; also, it does not matter if this code is fast or not; it needs to be run `2^22` times before it takes about one second, so a single iteration of this code is "fast enough". I was just reacting to my allergy against "this should be faster now". Sorry!)

lib/menubar/utils.lua
Sorky: Just saw your follow up... I think it is safe to say that the use-case is for a filename ("name,ext") and not just an extension - The comments note that a filename without an extension would be acceptable but I think it's safe to say that it isn't expecting someone to just to an extension

Nice boundary value test however

actionless:
i am not getting what kind of bug is this PR fixes? also i see from the benchmark above what it makes super-slow desktop-entry code even slightly more slower so basically :-1: from me, sorry

actionless:
@Sorky if you want to make a contribution to awesome but dunno from where to start i recommend looking on the recently created issues, for example this one should be quite easy to implement: https://github.com/awesomeWM/awesome/issues/2437

Sorky:
Good idea - Still working on my config and posting as I find things that I believe could be fixed/improved but I'll be more patient 

For now I'm looking into an actual bug...

Second notification doesn't show the second line of text...

local naughty	= require( "naughty" )

local presets_1	= {	font = "Monospace Regular 12",
					text = "<span foreground='#FF0000' background='#0000FF'><u>Available updates</u>\nNone                                 </span>"	}
local presets_2	= {	font = "Monospace Regular 12",
					text = "<span foreground='#FF0000' background='#0000FF'><big><u>Available updates</u></big>\nNone                                 </span>"	}

naughty.notify( { preset = presets_1, margin = 10, timeout = 0 } )
naughty.notify( { preset = presets_2, margin = 10, timeout = 0 } )

actionless:
first think i would do before continuing investigation -- try if the issue reproducible when running awesomewm built from git master (in arch it can be installed as `awesome-git` from AUR)

Sorky:
One thing I realized looking at the code was that the documentation (like much of the rest) lacks a good description of how to use the functions(s)

In this case, (and lookup_icon which calls it) could do with the following doco...

Usage 1: Full filename and path (eg: /x/y/z/file.ext) ...
Usage 2: Full filename (eg: file.ext) ... Will search icon lookup paths
Usage 3: Partial filename (eg: file) ... Will search icon lookup paths and try valid icon types 

@psychon - Still holding out hope that case 2 and particularly case 3 may offer the small performance improvement I anticipated ;-)

Sorky:
> first think i would do before continuing investigation -- try if the issue reproducible when running awesomewm built from git master (in arch it can be installed as `awesome-git` from AUR)

Good idea - I should look into installing that in parallel with my current working 4.2 base install - easy? able to swap back and forth?

actionless:
you can build it right in some directory and run from there inside Xephyr

or, on arch, you can install from AUR `awesome-git` to replace stable AwesomeWM with git master

lib/menubar/utils.lua
blueyed: Faster: `icon_file:match("^.+(%..+)$")` (1s instead of 1.4s with @psychon's benchmark).

blueyed:
I appreciate the PR, and tested the benchmark.

```
lua old.lua  1.44s user 0.00s system 99% cpu 1.443 total
lua new.lua  1.49s user 0.00s system 99% cpu 1.488 total

luajit old.lua  0.79s user 0.00s system 98% cpu 0.799 total
luajit new.lua  1.36s user 0.00s system 99% cpu 1.369 total
```

Using `local icon_file_ext = icon_file:match("^.+(%..+)$")` to get the file ext however:
```
lua new.lua  0.90s user 0.00s system 99% cpu 0.901 total
luajit new.lua  0.79s user 0.00s system 99% cpu 0.795 total
```

lua -v: 5.3.5
luajit -v: 2.0.5

I think that the code is simpler in general, and can be changed to be actually faster, re-opening.

Sorky:
Ta - If necessary, I'll fix the case where someone enters "PNG" and it is found in the first use-case where it should not [note to self: fix = make sure icon_ext does not get set if "." not icon_file]

blueyed:
luacheck:

> lib/menubar/utils.lua:233:27: shadowing definition of variable possible_file on line 224

As for testing: you could just start with adding tests to our (busted) unit tests.. :)

psychon:
> Something more to learn - unit tests!

Look for example at `spec/gears/string_spec.lua`. Unit tests are run by running `busted` (which needs to be installed first, for example via LuaRocks; alternatively you just hope for the best and have tests only run on Travis, but that idea never worked out for me).

Sorky:
Most definitely need help on the tests - I've tested externally (as indicated in earlier comment) - Could n't figure out where to start even with the suggestion

Elv13:
There is many different test suits. The `test-something.lua` files in `/tests/` are integration tests. They run in a real awesome instance. Each steps is executed 5 time or until it returns true (to test async stuff). It should be avoided for stuff that doesn't really depend on the X server bahavior because they are "heavy" tests.

The files in `/spec/` are the unit tests. They are low level and can't really use the whole API (and need to mock/shims everything manually). It's very light and very fast, but very verbose to write (on purpose).

The files in `/tests/examples` are tests whose results can be used in the documentation (as image, code and/or `print()` result). One relevant example for the menubar is https://github.com/awesomeWM/awesome/pull/2510/files#diff-2f52c06b050e2af1f349e6dbc4d79813 . While it isn't really good at low level stuff, it emulates enough of Awesome to make normal code run without a full instance. I prefer those because it's a 2-in-1 and allows to better document how Awesome works at the same time. The main downside is that it doesn't have `assert.is_equal` and other goodies the `/spec` have because it doesn't (and "can't") depend on `lua-busted` (because it is enabled by default and busted is an optional dependency).

You also have all the static analysis and linting stuff, but that's not relevant for this pull request.

Each layer/testsuite is better at a certain type of tests. For this type of tests, the `spec` is probably the better choice. But you will have to shims all the file/IO stuff or put some files on the disk. Both are not going to be really funny to make work (which is very common for testing code...). If you want to save time/effort, you can use the other 2 testsuite. It isn't ideal, but totally acceptable.

lib/menubar/utils.lua
Sorky: Nice work & thanks again

blueyed:
> I updated the spec file with more comprehensive test cases but I don't know how to merge that with this

Just commit it to your branch (patch-4), no?

> I also need to know how to add 5 new icon file symlinks

Not sure, but found this:
```
spec/awful/completion_spec.lua
31:            test_path:get_child(cmd):make_symbolic_link(target)
```

Sorky:
Looks like I merged this into Update utils_spec.lua #2542 [I hope]

https://github.com/awesomeWM/awesome/pull/2542/files

actionless:
should this one be closed then in favor of your new PR?

Sorky:
Hopefully yes ;-)

actionless:
i am asking, if the other PR includes all the changes as in this one?
if so, please close this one

Sorky:
I surely will when I know the transfer worked and met the testing requests

blueyed:
Closing this one already to avoid confusion.  It can be re-opened if needed, and commits are still available, of course.

PR #2469, opened 2018-11-07
menubar.utils: Remove some (basically) dead code

psychon:
This removes unused public functions from menubar.utils. This is not an
API break, because these functions were only added in commit
8d34201ec3dd76ce9d8, which is after v4.2. These new functions became
unused in commit e88f1e873594bf, not long after they were added.


psychon:
> codecov/project — 84.22% (-0.01%) 

Heh. By removing dead code that no one was using, but which had tests covering it, I decrease code coverage. :-)

PR #2472, opened 2018-11-09
Fish hotkeys

actionless:


lib/awful/hotkeys_popup/keys/init.lua
```diff
@@ -13,7 +13,26 @@ local keys = {
   tmux = require("awful.hotkeys_popup.keys.tmux"),
   qutebrowser = require("awful.hotkeys_popup.keys.qutebrowser"),
   termite = require("awful.hotkeys_popup.keys.termite"),
+  fish = require("awful.hotkeys_popup.keys.fish"),
```
Elv13: At which point do we switch to "plugins" for this? I mean, this doesn't scale. We load all those files when Awesome starts and this shell doesn't even have 1% market share (neither are the others 2 above it).

Maybe we could have the class in `init.lua` then `dofile` the actual modules when required?

actionless: that not happens unless somebody would do `require("awful.hotkeys_popup.keys")` explicitly


see https://github.com/awesomeWM/awesome/blob/master/lib/awful/hotkeys_popup/init.lua

Elv13: Which happens to be everybody unless manually disabled.

https://github.com/awesomeWM/awesome/blob/master/awesomerc.lua#L20

actionless: mb comment out that line or replace it to explicit import of `...keys.vim`?

Elv13: I don't think it would help at all. First of all, why vim, but really, as other people sometime contribute new schemas, we need a solution. `dofile` is simple enough.

actionless: but only explicit import of `awful.hotkeys_popup.keys` is importing all the hotkeys

it just need to be changed in rc.lua not to do so

actionless: also mb some documentation should be generated for `keys` module to explain that

Elv13: > it just need to be changed in rc.lua not to do so

That would be changing the behavior, which I promised to keep to a strict minimum. Why not using `dofile()` it really is a small change, there is not much there?

actionless: but it will also change the behavior in the same way

Elv13: no, all it changes is that those files are loaded the first time they are used instead of when Awesome starts

actionless: but it's written in the comment https://github.com/awesomeWM/awesome/blob/master/awesomerc.lua#L18-L19 what it's intended behavior

changing this will be a breaking change for users who using it for importing all the available hotkeys

while changing default rc.lua won't be a breaking change

Elv13: I think you misunderstand what I propose. I propose to switch from `require` to moving the `hotkeys_popup.add_group_rules(group_name, group_data)` parts to `keys.lua` for the existing implementations. Then those implementation load the file on demand. Other "hacks" would be to use the filename as the class, but it isn't as flexible.

This would import all hotkeys in exactly the same way, the only thing that changes is that we have some way to avoid loading "useless" files.

actionless: > load the file on demand

i can't understand what is the demand in this example, as far as i could think about this the only one could be actually `requir`ing them

Sorky: I agree with NOT adding unused files - A generic way for users to add application specific hotkey files makes more sense to me

Issue #2476, opened 2018-11-10
Question about wibox.widget.textbox:set_font(font) and other similar methods


Sorky:
@psychon

What do you (and others) think of...

    function textbox:set_font(font)
        self._private.layout:set_font_description(beautiful.get_font(font))
        self:emit_signal("widget::redraw_needed")
        self:emit_signal("widget::layout_changed")
        **return self**
    end

To allow...

	local textbox = wibox.widget.textbox( "??" ):set_font( font )

PS: This method doesn't seem to appear in the doco

Sorky:
I was able to do 
    local textbox = wibox.widget.textbox( "Hi" )
    textbox.font = font
and
    local textbox = wibox.widget.textbox( "Hi" )
    textbox:set_font( font )

Surely the method is additional in this case? Causes a refresh?

Sorky:
Just like lua string methods

Elv13:
> Just like lua string methods

No. Lua string return a new string, not a reference to the old string.

PR #2477, opened 2018-11-10
Update textbox.lua

Sorky:
As per https://github.com/awesomeWM/awesome/issues/2476

Several methods...
  set_markup
  set_text
  set_ellipsize
  set_wrap
  set_valign
  set_align
  set_font

### DIFF
--- a/lib/wibox/widget/textbox.lua
+++ b/lib/wibox/widget/textbox.lua
@@ -171,6 +171,7 @@ function textbox:set_markup(text)
     if not success then
         gdebug.print_error(message)
     end
+    return self
 end
 
 function textbox:get_markup()
@@ -191,6 +192,7 @@ function textbox:set_text(text)
     self._private.layout.attributes = nil
     self:emit_signal("widget::redraw_needed")
     self:emit_signal("widget::layout_changed")
+    return self
 end
 
 function textbox:get_text()
@@ -211,6 +213,7 @@ function textbox:set_ellipsize(mode)
         self:emit_signal("widget::redraw_needed")
         self:emit_signal("widget::layout_changed")
     end
+    return self
 end
 
 --- Set a textbox' wrap mode.
@@ -227,6 +230,7 @@ function textbox:set_wrap(mode)
         self:emit_signal("widget::redraw_needed")
         self:emit_signal("widget::layout_changed")
     end
+    return self
 end
 
 --- The textbox' vertical alignment
@@ -243,6 +247,7 @@ function textbox:set_valign(mode)
         self:emit_signal("widget::redraw_needed")
         self:emit_signal("widget::layout_changed")
     end
+    return self
 end
 
 --- Set a textbox' horizontal alignment.
@@ -259,6 +264,7 @@ function textbox:set_align(mode)
         self:emit_signal("widget::redraw_needed")
         self:emit_signal("widget::layout_changed")
     end
+    return self
 end
 
 --- Set a textbox' font
@@ -269,6 +275,7 @@ function textbox:set_font(font)
     self._private.layout:set_font_description(beautiful.get_font(font))
     self:emit_signal("widget::redraw_needed")
     self:emit_signal("widget::layout_changed")
+    return self
 end
 
 --- Create a new textbox.


Elv13:
The official way to access those fields is using the property like `my_textbox.markup = "<b>Awesome!</b>"`. I get that you are trying to allow daisy chaining setters, but that both undocumented, undocumentable and the only module that would allow such construct.

So, why is this an improvement? I mean, if everything in Awesome was patched to support this and it was enforced but some linting or meta tests, I guess it would be fine-ish, even if a little intrusive change. However this brings the corner case where a setter has to return an error code or a boolean confirmation. In that case you would have daisy-chainable setters mixed with error returning setters and it would be very confusing. In that case, it means every property would need to be documented whether or not it allows daisy-chaining. Yet the documentation barely mention the explicit setter exist at all (on purpose).

I am not rejecting this PR, but again this needs more discussion and a plan for testing and consistency across the whole API. I don't see this as an improvement, but that may just be that my coding style isn't compatible with it and all the new kids program this way due to some web framework and I am old. That's not a logical argument, so lets keep this focused on facts rather than personal preferences.

Sorky:
The reason I found this is I tried...

    local textbox = wibox.widget.textbox( "Hi" ):set_font( font )

Why did I try it...

    It is, dare I say, normal (the whole lua stirng library?) 

What would I rather have in my code...

    local textbox = wibox.widget.textbox( "Hi" ):set_font( font )

_or_

    local textbox = wibox.widget.textbox( "Hi" )
    textbox:set_font( font )

Sorky:
As to the bigger picture...

Setters that need to return a value...

    My 2c = Don't expose them _or_ don't call them "set_"

Are your right again - there definitely are more cases [two methods I didn't touch didn't start with set / both seemed impossible to find in the doco.] 

On to documentation - The more the merrier - I know it's a pain for the developer, but it helps the user and I can categorically state that it needs consistency (read the code is what I'm doing and you know what that means ;-)

Sorky:
Just found a condition I really could do with a return value...

_ = state and widget:set_fg( fg_normal ) or widget:set_fg( fg_focus )

not possible without a return value

Elv13:
> _ = state and widget:set_fg( fg_normal ) or widget:set_fg( fg_focus )

    widget.fg = state and fg_normal or fg_focus

Plus it even allows to do

    widget.fg = state and fg_focus or widget.fg

and

    widget1.fg, widget2.fg = fg_focus, fg_focus

> local textbox = wibox.widget.textbox( "Hi" ):set_font( font )

The newer widgets use an `args` Lua format so you can do

    local s = wibox.widget.separator {
        thickness = 42,
        color     = "#ff0000"
    }

Which is Lua (and to some extent Python) official way to use named parameters. It is both cleaner and more consistent. More and more widgets are ported to this syntax in every release. The `textbox` will eventually get there. Otherwise the alternative syntax is already supported for the textbox:

    local s = wibox.widget {
        text   = "foo",
        fg     = "#ff0000",
        widget = wibox.widget.textbox 
    }

Daisy chaining like you propose is a JavaScript hack for a misguided programing language.


> On to documentation - The more the merrier - I know it's a pain for the developer, but it helps the user and I can categorically state that it needs consistency (read the code is what I'm doing and you know what that means ;-)

Our documentation has an enormous time investment in it. It also have been rewritten some time ago to make things consistent. The fact that getters and setters are not documented is because they were **removed** from the documentation. Other other components such as the widget private variable and "interface" methods such as `draw` have also been removed. This was done on purpose to focus on the property driven API.

Sorky:
I could very well be doing it wrong...

a) Don't believe textboxes gave this antway
	local textbox = wibox.widget.textbox( "?" )
	textbox:bg = "#FFFFFF"

b) Seems to me the propeties are protected 
function background:set_bg(bg)
    if bg then
        self._private.background = color(bg)
    else
        self._private.background = nil
    end
    self:emit_signal("widget::redraw_needed")
    return self
end

Either way - As you noted at the start, your not for or against (back then at least) - lets see what a few days thought has on it

Gotta fly (literally) - will check in in a couple

Elv13:
textbox.bg = "#FFFFFF"

actionless:
i think if you want quick way to set multiple values i am agree with @Elv13 what using `args` as in new widgets will be more consistent for the API

Sorky:
Hey - don't be so quick to close this - Lots of advantages and too many current shortcomings - please open your mind and listen to others

Elv13:
> Lots of advantages

You will have to make your point very clear on this. So far what you said about this is mostly:

 * I used a web framework where this is supported
 * Strings allow immutable chains, so I propose adding a mutable chain (a well known anti-pattern)
 * I tried the property syntax but made a typo and it didn't work
 * It adds a third incompatible syntax to add many properties in the constructor
 * It allows expressions, even though the property syntax also does

Sorry, but it's not very convincing.

> too many current shortcomings

How would you address those shortcomings? So far they seems pretty much insurmountable to me. Keep in mind that there is a rather large maintenance cost associated with "alternate ways" to do everything. Currently me and @actionless agree that the property attribute and named property based constructor are the future we want to converge toward. Since 3.5, a large body of work ( > 10000 lines of patch) has been invested into cleaning up the API to bring consistency. All those things in the "deprecated" section of the doc come from that project. My vision of the future and the code I submitted to make it happen is getting closer to a property driven declarative programming syntax where you express the intent rather than the mechanics. This is what QML, CSS and SQL languages provide and I see it as a good fit for something very domain-specific like a window manager.

What you propose here goes backward toward the more inconsistent imperative APIs we had in the past. The second large challenge of what you propose is that it cannot exist in a void. It would have to be done with *everything* and enforced against regression introduced by new contribution in the CI system. This would be yet another maintenance heavy subsystem.

Sorky:
You started out by saying "That's not a logical argument, so lets keep this focused on facts rather than personal preferences." then soon degenerated into ""a JavaScript **hack** for a **misguided** programing language" and now you refer to it as "I used **a web framework**" - Wow - I'm not trying to be rude (and if I've comes across that way I apologize) but you seem to be getting just plain nasty. 

You must think I'm active here just to make your time miserable - Far from it - As a recovered user (from back in 2012? on the old mailing list) I finally migrating into 4.2 and I'm reliving the "new user" experience and just sharing my learning / insights - maybe not in the way you'd like, / maybe a bit critical (particularly of the documentation) - If you take that personally - again - I apologize

I'm going to answer your allegations...

- I used a web framework where this is supported
    Never said / didn't mean that - Malicious? / Unrelated? -  Ignored
- Strings allow immutable chains, so I propose adding a mutable chain (a well known anti-pattern)
    I don't even get this - immutable chain is a Java name for a common OO technique
    Strings was just the first example that leapt to mind - maths is another
    mutable chain & anti-pattern / don't get that / Malicious? / Unrelated? -  Ignored
- I tried the property syntax but made a typo and it didn't work
    I didn't keep the actual code I used as it didn't work...
    I made the typo on the web page trying to give you an example 
    Malicious? / Sarcastic? / Unrelated? -  Ignored
- It adds a third incompatible syntax to add many properties in the constructor
    Not quite what you mean here - adds? "set_ and "get_" exist!
    incompatible syntax / constructor?
- It allows expressions, even though the property syntax also does
    Property system is ok if it is there and works - I had a lot of trouble with it and fount the "set_" AOK

_insurmountable shortcomings?_ Unfortunately I read the above as mostly a malicious ramble - I'm just a bit slow at taking the hint?

_maintenance cost?_ Now this is a good discussion point - I  don't see having "set_" functions return themselves as being a major maintenance cost, - sure no change = less work than change - you may be able to explain it in more detail?

_"alternate ways" to do everything_ - The "set_" & "get_" are already there - If I was asking to add them, then sure - it would be an alternative way - Having a return value / allowing it to chain is just an extension

_Currently me and @actionless agree that the property attribute_ - No problem with that - I had trouble and found (NOT invented) something that worked and offered what I see as a minor enhancement

_named property based constructor are the future_ - I'm guessing this falls into the same old argument I here about every suggestion : "future plans"

_All those things in the "deprecated" section of the doc come from that project. My vision of the future and the code I submitted to make it happen_ - That's great - Well done - I actually appreciate it - Hope you keep up the good work - Wonder it there are others out there trying to help? 

_is getting closer to a property driven declarative programming syntax where you express the intent rather than the mechanics. This is what QML, CSS and SQL languages provide and I see it as a good fit for something very domain-specific like a window manager_ - Hmmm OO programming???

_What you propose here goes backward toward the more inconsistent imperative APIs we had in the past_ - How - "_set" already exists - Should I raise a PR removing it claiming that the property exists and the "get_" and "set_ should NOT exist at all? - Should I raise a PR removing all property access as the "args" that @actionless espouses exist? Yes those were me being meant sarcastic

_The second large challenge of what you propose is that it cannot exist in a void._ - It already does - Limits can be put in place, much like you are doing now

_It would have to be done with everything and enforced against regression introduced by new contribution in the CI system. This would be yet another maintenance heavy subsystem_. - As before limits / details

.........................

I can't paste my actual code here as I'm travelling, but the gist of it was...

My first choice...

> create_object( limited_properties_here ):set_property( here):modify_behaviour( new )

Vs 

My second choice...

> create_object( limited_properties_here, PR for additional_previously _overlooked_property )

Vs

My third choice...

> create_object( limited_properties_here )
> object:set_property( here)
> object:modify_behaviour( new )

.........................

I still have no idea about the properties - From what I saw in whichever module section I was in [base?container?], the code I see seems s to have them nested under a "_private" which to me confirms why I could do what I tried and then... oh look - a "set_" and "get_'

Sorky:
that goes against the point of sharing learnings and assumes you two are the font of all wisdom

actionless:
LOL

actionless:
instead of continuing this pointless argument please see how properties are implemented and a newer widgets: https://github.com/awesomeWM/awesome/blob/master/lib/wibox/widget/progressbar.lua#L408-L417  and `args` handling: https://github.com/awesomeWM/awesome/blob/master/lib/wibox/widget/progressbar.lua#L436-L437

PR #2478, opened 2018-11-12
Fix c.requests_no_titlebar

sigprof:
The `c.requests_no_titlebar` property was always false because of a typo. I fixed this and added some tests for it, and also fixed a wrong example in documentation.

The test for the first commit is supposed to fail, but I want to be sure it actually fails in the CI environment; after the test completes, I will push commits in a different order so that there will be no test failures.

### DIFF
--- a/docs/05-awesomerc.md.lua
+++ b/docs/05-awesomerc.md.lua
@@ -221,7 +221,7 @@ sections.DOC_TITLEBARS = [[
 sections.DOC_CSD_TITLEBARS = [[
  For client side decorations, clients might request no titlebars via
  Motif WM hints. To honor these hints, use:
- `titlebars_enabled = function(c) return c.requests_no_titlebar end`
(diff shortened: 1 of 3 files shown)


sigprof:
https://travis-ci.org/awesomeWM/awesome/jobs/453848459 failed as expected on the first commit, but also failed at `test-spawn.lua` on the second commit, which was not expected (maybe there is a race condition in that test which causes it to fail intermittently).

https://travis-ci.org/awesomeWM/awesome/jobs/453848458 failed with an assertion failure inside Lua in some completely unrelated test.

Anyway, I pushed the commits in a different order, and those tests passed on the second attempt.

actionless:
codecov stucked for some reason, reopening PR to re-trigger it

actionless:
right after i re-opened codecov status refreshed and became fine but now travis restarted the run.. :-)

PR #2483, opened 2018-11-19
doc(readme): add questions section; split issues and contributing; fix apidoc link (in ldoc)

actionless:


docs/01-readme.md
```diff
@@ -120,20 +120,30 @@ Then in gdb set any args and run it:
 
     (gdb) set arg --replace
     (gdb) run
+    
+## Asking questions
+
+#### Stack Overflow
+You can ask questions on [Stack Overflow](http://stackoverflow.com/questions/tagged/awesome-wm).
```
actionless: however i would personally leave here (and on the website) only reddit, not to spread attention between two platforms

PR #2484, opened 2018-11-21
Add signal to call `rules.apply` when a window redefine its class property.

Aire-One:
This PR is to open discussions in the issue of Electron apps and rules in awesome wm.

# Basics

Last week, in "[This week in Usability & Productivity](https://pointieststick.wordpress.com/2018/11/11/this-week-in-usability-productivity-part-44/)" post, they mentioned KWin fixed the bug causing Electron apps to not follow window rules.

Here is the bugfix message:
> KWin window rules now work for Electron apps and other apps that change their window class during operation.

And here is the link to the merge thread in the KDE phabricator server: https://phabricator.kde.org/D16670#change-JBK9w4EqumD4

This bug is also present in awesome. Applications build with the Electron framework does not follow window rules defined by the user in the table `awful.rules.rules`.
In this PR, I propose a fix based on the solution implemented by KDE.

# Fix proposal and discussions

Based on the code from KDE, I wrote a fix in this PR. The following will be some discussions about what I wrote.

The basic idea of the fix is to attach a signal to the change of the window's `wm_class` property. With awesome it basically translates to something like this:

```lua
client.connect_signal("property::class", rules.apply)
```

To prevent multiple call of `awful.rules.apply` when a client sets its `wm_class` at its creation, this signal should only be set when the client is attached to awesome. That's why I put it in the `manage` signal defined in `awful.rules`.

Also, I disconnect this signal when the client get unmanaged. I'm not realy sure this is useful given that a client get unmanaged when it's deleted/closed. (I don't see any other case, maybe I'm wrong)

A quick explanation about why I call `c:connect_signal("property::class", rules.apply)` before `rules.apply(c)` in the client's managed signal callback. It's a trick I found to make Spotify not blinking.
Basically, I have a rule for Spotify to open on a specific tag. But the window has the time to appear on the screen just before awesome handles the `wm_class` change and calls again the `rules.apply` method. It also causes the destination tag to not appears as urgent.
This trick seams to fix this behavior.

# Unit tests

I tryed to do some unit tests but I'm actually stuck. Here is what I wrote so far in `tests/test-awful-rules.lua`:

```lua
test_rule {
    properties = {
        tag = "2"
    },
    test = function(class)
        local c = get_client_by_class(class)
        -- Make sure C-API properties are applied

        local new_class = "class_changed"
        local new_tag = "3"

        -- First, check the client follow initial rules
        assert(c:tags()[1].name == "2")

        -- Insert new rules for a specified wm_class
        table.insert(awful.rules.rules, {
            rule = { class =  new_class },
            properties = { screen = 1, tag = new_tag }
        })

        -- Change the client class
        c.class = new_class

        -- Test rules application
        assert(c:tags()[1].name == new_class)

        return true
    end
}
```

The problem is: the assertion to check if the rules are applied following the client `wm_class` change is automatically always false because awesome can't handle the event and so can't call the signal in this situation.

---

I hope my work can help doing something to fix this issue.

### DIFF
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -737,7 +737,19 @@ function rules.completed_with_payload_callback(c, props, callbacks)
     rules.execute(c, props, callbacks)
 end
 
-client.connect_signal("manage", rules.apply)
+client.connect_signal("manage", function (c)
+    -- Some applications (like Spotify) does not respect ICCCM rules correctly
+    -- and redefine the window class property.
+    -- This leads to having window which does *NOT* follow the user rules
+    -- defined in the table `awful.rules.rules`.
+    c:connect_signal("property::class", rules.apply)
+
+    rules.apply(c)
+end)
+
+client.connect_signal("unmanage", function (c)
+    c:disconnect_signal("property::class", rules.apply)
+end)
 
 return rules
 


psychon:
Meh. Can't this be fixed in Electron instead? The KDE guys already say that this is basically a bug in the electron apps (`WM_CLASS` should not change while the window is managed). Also, since this is AwesomeWM, people who want to work around Electron bugs can just do the equivalent of your fix in their `rc.lua`. That's not possible with KWin.

I would claim that rules in AwesomeWM are more powerful than in KWin. Rules can have unwanted side effects when they are run multiple times (e.g. `new_tag`).

> I tryed to do some unit tests but I'm actually stuck. 

Sorry, but `c.class` is read-only. Thus, your `c.class = new_class` silently does nothing. (Or might end up in the newindex-miss-handler, but that's almost equivalent to "silently does nothing".)

> Also, I disconnect this signal when the client get unmanaged. I'm not realy sure this is useful given that a client get unmanaged when it's deleted/closed.

I'd say: Nope, this does not actually make any difference.

psychon:
Was this bug reported to the Electron devs? I cannot find anything.

psychon:
@actionless I'm still waiting for an answer. Perhaps @Aire-One agrees to report this bug to the electron devs and leave a link here, they fix it properly and everyone is happy.
(I tried investigating what is going on there, but... changing the window class and instance should not even be possible at runtime with chromium. Since electron does not seem to be in any distro and I am allergic to "run a random binary from the internet", I cannot do much more myself.

Aire-One:
While I would like this PR to make its way upstream, I do understand and agree with your opinion.

Nonetheless, I did some more tests with the Spotify application and it seems that it does not redefine the `wm_class` but rather declare it later than expected... Anyway, I guess it's still a violation of the ICCCM.
I also tried to run Atom Editor which is the only other Electron app I'm using. Atom seams to define correctly the `wm_class` property.
At this point, I'm confused. In contradiction to what the KDE dev said, not all Electron applications are problematics but maybe only Spotify is. (?)

Anyway, this PR can be closed.

psychon:
Ah, okay. This is indeed Spotify-specific. Now I managed to find some stuff!
https://github.com/i3/i3/issues/2060
https://community.spotify.com/t5/Desktop-Linux/WM-CLASS-window-property-on-Linux/m-p/1247042

Specifically:

> We will "soon" switch the desktop client to use CEF with the Views architecture instead, which means the Chromium code will create and map the window. With Views there is a way to tell Chromium to set WM_CLASS correctly, by setting some widget parameters.

So, apparently, Spotify will "soon" update to a newer Chromium version and then stuff will just work. ("Soon" was written more than two years ago).

Closing since this is Spotify specific and Spotify is "doing it wrong".

PR #2485, opened 2018-11-21
Improve docs for client class and instance

psychon:
No idea what "If the client has multiple classes" is supposed to refer
to, but this commit gets rid of that reference. Also, this commit
clarifies how to get the class and instance out of xprop.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -324,9 +324,8 @@
 /**
  * The client class.
  *
- * If the client has multiple classes, the first one is used.
- *
- * To get a client class from the command line, use the `xprop` command.
+ * To get a client class from the command line, use the command `xprop
+ * WM_CLASS`. The class will be the second string.
  *
  * **Signal:**
  *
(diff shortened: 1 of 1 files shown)


actionless:
> No idea what "If the client has multiple classes" is supposed to refer to

```
$ xprop WM_CLASS
WM_CLASS(STRING) = "urxvt", "URxvt"
```

Elv13:
technically in that example one is the instance and one is the class

actionless:
ah, i see new comment actually mentioning xprop output

psychon:
>> No idea what "If the client has multiple classes" is supposed to refer to
>
> $ xprop WM_CLASS
> WM_CLASS(STRING) = "urxvt", "URxvt"

Well... ICCCM § 4.1.2.4, "WM_CLASS Property" says that there is only one class:

> The WM_CLASS property (of type STRING without control characters) contains two consecutive null-terminated strings. These specify the Instance and Class names to be used by both the client and the window manager for looking up resources for the application or as identifying information. This property must be present when the window leaves the Withdrawn state and may be changed only while the window is in the Withdrawn state. Window managers may examine the property only when they start up and when the window leaves the Withdrawn state, but there should be no need for a client to change its state dynamically.
> [...]

(CC #2484)

PR #2489, opened 2018-12-03
add ldoc for tag icon_only property

azhi:
it is used in taglist widget ([here](https://github.com/awesomeWM/awesome/blob/8ac83cac5fe4fd9d7e8257df385517a479ba9d9b/lib/awful/widget/taglist.lua#L362)).

Note that I didn't check how rendered doc would look like, hopefully it will be ok.


actionless:
btw i just noticed a weird behavior with that option on taglist: 

before:
2018-12-04--1543903171_82x46_scrot



```sh
awesome-client 'root.tags()[1].icon_only = true'
```

2018-12-04--1543903210_66x42_scrot


```sh
awesome-client 'root.tags()[1].icon_only = false'
```

2018-12-04--1543903210_66x42_scrot


so margin disappears after setting icon_only, and not appearing back after unsetting it back

azhi:
@actionless not sure what's causing that behavior, but IMO setting `icon_only` property without setting an `icon` is asking for trouble :).

@Elv13 Updated 1 and 2. English is not my native language, so feel free to change the wording in case it still has issues.

My take on the main concern is:
I agree that having fields specific to one of the widgets inside tag API seems bad. In fact, I would suggest moving `icon` to widget props as well (since it is used only for displaying a tag on a widget), and maybe introduce some widget prop like `title` to allow showing not only tag name, but some additional info.
On the other hand, if you look at it at more general point of view, tag is just storing an information about how it should be displayed. There are other examples of this - e.g. client's `border_color` and `icon`.

Decision is of course up to you guys, feel free to close this in favor of better implementation.

PS: I'm not sure what is these 'custom delegates' you are referring too are, if you can point me in a place where I could read something about them it would be nice :).

actionless:
@azhi  but does those commands working as expected for you?

i also checked now with the icons -- and i have the same behavior as described

azhi:
Yep, it works for me.

image

actionless:
>Tried to dynamically set icon_only (by awesome-client 'root.tags()[1].icon_only = true') - it had no effect whatsoever.

it have effect after switching a tag

azhi:
Yep, reproduced it, for some reason it is reproducible only on some tags for me.
It removes margin but does not remove a name.
So yeah, it needs fixing.

azhi:
Updated, the reason for this margin behavior was this commit - https://github.com/awesomeWM/awesome/commit/f453863794d41bb0d30313c2624e75bb8cf07266.
Overall, `taglist` and `awful.widget.common` was not ready for an update without a text.

Included fixes:
* `taglist` widget is now subscribed to icon_only property updates.
* `awful.widget.common` now erases text when it receives update that resulted in empty rendered text
* `awful.widget.common` now properly handles text margin - removes it when there is no text and restores it when text appears

The most dangerous is second one - there might be a widget that is relying on `awful.widget.common` not erasing text when it temporary doesn't render any. This also means that previously this kind of widget had it's text margin removed whenever it tries to do so, so it narrows down potential misbehaving widgets to those that don't have margin in the first place (otherwise somebody would've noticed disappearing margin).

I'll try to look at failing travis build and some additional tests tomorrow.

lib/awful/widget/common.lua
```diff
@@ -117,11 +118,40 @@ function common.list_update(w, buttons, label, data, objects, args)
...
+                if m > 0 then
+                    has_any_margin = true
+                    break
+                end
+            end
+
+            if has_any_margin then
+                -- Save current margin values so they can be restored in case
+                -- text will be present in future update
+                cache.tbm.saved_margins = current_margins
```
azhi: Btw, I don't have much experience with lua, so:
Is it OK to add custom fields to existing objects like that?

Elv13:
@azhi Thanks again for putting time into fixing bugs.

I wonder how to go forward on this. On one hand there is a real bug in a feature we currently support. On the other hand, it increases complexity in an already saturated part of the code which was refactored into delegates so we could stop adding more code to it. I will let @psychon weight on this. If Uli wants to fix the bug, then I am ok merging it. If, on the other hand, it is decided to just deprecate all of those options (without breaking them for 4.x) and focus on delegates, then this will be it.

actionless:
i am up for *both* fixing and deprecating it

Elv13:
> i am up for both fixing and deprecating it

But so far it isn't officially part of the API yet, merging this would make it official, then deprecate it. To me that makes little sense to do both.

actionless:
my point is what having something documented and deprecated better than just something undocumented

psychon:
I'd prioritise getting a new release out over most other stuff, but okay...

> i am up for both fixing 

I did not really look and no one wrote anything about it yet, but: Anyone figured out what the bug is and how to fix it? If not, we can talk a lot about what happens after it is fixed, but nothing will happen. (And no, I'm not investigating, sorry.)

actionless:
@psychon i've described a bug in a message above and this PR also fixes it

psychon:
Urgh. You mean the fix which saves the margins in some random property on the textbox? I do not really like that one at all, sorry...

actionless:
mb in that case it will be better to extract hardcoded `dpi(4)` when it initializes the margin widget (`tbm`) into a variable and do set_margin with that variable instead of doing this trick with saving and restoring?

psychon:
That would break with delegates that use any other size of margins by default, I guess :-/

PR #2500, opened 2018-12-17
prompt: Add a `with_shell` property.

### LINKED ISSUE #2486: awful.prompt: specify env variables and aliases
Hello!

I have alisases for a number of programs when I use my main shell (bash).

I also have a couple of programs in ~/bin.

It would be nice if awful.prompt could be configured in such a way that I can use these aliases and programs without specifying the full path.

This is a bit related to #1609, so I'll leave the reference here.

Cheers,

Elv13:
~~(untested)~~

@aviau: This should help, please try

Fixes #2486

### DIFF
--- a/lib/awful/widget/prompt.lua
+++ b/lib/awful/widget/prompt.lua
@@ -58,11 +58,20 @@ local function run(promptbox)
 end
 
 local function spawn_and_handle_error(self, ...)
-    local result = spawn(...)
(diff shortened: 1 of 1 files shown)


aviau:
Added to my todo, thanks!

Elv13:
Rebased since it had a merge conflict with some newly merged code

Elv13:
Last call, it will be merged if no one else steps in

Issue #2504, opened 2018-12-23
[extra rule in rc.lua] Tor Browser doesn't float by default


ArenaL5:
Output of `awesome --version`:

```
awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.2
```

**How to reproduce the issue:**

1. Boot Ubuntu, using awesome as the window manager.
2. Use the `max` or `fullscreen` window layout. (Very important)
3. Download the [Tor Browser](https://www.torproject.org/download/download-easy.html.en) bundle for Ubuntu.
4. Untar it, make `start-tor-browser.desktop` (named "Tor Browser" or "Install Tor Browser" in graphic file managers) executable and run it.
5. Follow the instructions in the dialog window to connect Tor Browser, and wait for the dialog to close and the main window to appear.

**Actual result:**

The Tor Browser window fills the whole screen.

**Expected result:**

This _is_ normal behaviour for Awesome, but that's not desirable in this case. Tor Browser uses (among other measures) a fixed window size to avoid fingerprinting by screen resolution.

If you try to maximize Tor Browser _after_ it's loaded, a warning will appear under the URL bar, along the lines of

> Maximizing the Tor Browser may allow websites to determine your screen size, which could be used to track you. Our recommendation is to leave Tor Browser's windows at their default size.

but for the moment it doesn't detect window managers overriding this default size from the start, which is a bug in Tor Browser itself. Anyways, it would be convenient and useful to add an extra rule to `awful.rules.rules`: adding `Tor Browser` to `awful.rules.rules.rule_any.class` will do. (The class of the window will always be `Tor Browser` in English, regardless of system or browser locale.)

blueyed:
Thanks for the detailed report.

I think we should change the default configuration - please consider creating a PR for it.

ArenaL5:
I never even considered creating a pull request because I never published code on the web. This is my first one.

Is it acceptable?

blueyed:
Yes, thanks!
But in general I recommend using separate branches for PRs (and not your master branch).  This is important when it might take longer for a PR to get merged, and if you do multiple - but I see that you're just starting.. :)

ArenaL5:
Pull request #2505 merged, so I'll close this issue.

PR #2505, opened 2018-12-23
Making Tor Browser a floating window by default.

ArenaL5:
Tor Browser uses a default window size to make its users indistinguishable. Overriding this default window size is a privacy risk for Awesome's users. (Details at [issue #2504 at awesomeWM](https://github.com/awesomeWM/awesome/issues/2504)

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -504,10 +504,10 @@ awful.rules.rules = {
           "Kruler",
           "MessageWin",  -- kalarm.
           "Sxiv",
+          "Tor Browser", -- Needs a fixed window size to avoid fingerprinting by screen size.
           "Wpa_gui",
           "veromix",
-          "xtightvncviewer",
-          "Tor Browser"},
+          "xtightvncviewer"},
 
         -- Note that the name property shown in xprop might be set slightly after creation of the client
         -- and the name shown there might not match defined rules here.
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!
The build failure is unrelated (https://github.com/awesomeWM/awesome/issues/2163), will restart the build.

awesomerc.lua
```diff
@@ -506,7 +506,8 @@ awful.rules.rules = {
           "Sxiv",
           "Wpa_gui",
           "veromix",
-          "xtightvncviewer"},
+          "xtightvncviewer",
+          "Tor Browser"},
```
blueyed: Just noticed that the entries are sorted, so please adjust it - a good exercise maybe for your first PR.
Please amend the existing commit and force-push it then.
But just let us know if this is too much, and we can take it from here also.

actionless:
i would also add some very short comment about why it's designed like that, otherwise users which could see that behavior could think what it was made in the config just like an example of floating app rule, not because of privacy reasons

ArenaL5:
I'm sorry for taking so long, I was away from the computer because of family matters.

I've updated `awesomerc.lua` in my `main` branch (next time I'll try to learn to use different branches), via web client. Do I need to do something else on my end?

blueyed:
All good, thanks a lot @ArenaL5!

PR #2506, opened 2018-12-27
awful.layout: move arrange locks

blueyed:
This fixes a deadlock when some arrange signal handler itself changes
a property like `border_width`:

    function border1(c)
      c.border_width = 50
    end
    client.connect_signal("property::border_width", border1)

      local border_for_s = function(s_)
        local clients = awful.client.visible(s_)
        for _, c in pairs(clients) do
          c.border_width = 2
        end
      end
      screen[s]:connect_signal("arrange", border_for_s)
    end)

It might be better to have some counter there instead and emit a warning
if this happens (while allowing for e.g. 3 recursions).

This would allow handlers to break out of this themselves (by not insisting to
change the property).

squash! awful.layout: move arrange lock

Release arrange_lock before emitting signal, but not delayed_arrange.

Not really sure, but this also fixes the example, while might allow


blueyed:
Re the code example: it is part of me handling borders automatically and back then at least arrange was needed for this really, but thanks for pointing out that it should be avoided - I might revisit it.
The deadlock happens together with awesome-cyclefocus, which decorates the current client, and then this might happen.

As for the PR itself, should we merge it then?

psychon:
For your code example: Why the `for_each_screen` dance instead of just `screen.connect_signal("arrange", ...)`?

For the commit message: I assume the issue that you are seeing is some infinite recursion. On the "arrange"-signal, you set the border width to 2, which causes "property::border_width", thus setting the border width to 50, which causes another re-arrange etc.

For the change: In the above scenario, what is the expected outcome that you want to get out of this? As-is, the effect of the patch is "ignore any kinds of window changes done by code connected to `arrange`". For example, you end up with overlapping clients or gaps between clients if something changes border width in `arrange`. Is that really something that could be considered "everything is fine now"?

blueyed:
> Why the for_each_screen dance

Likely due to this being based on old code.

> I assume the issue that you are seeing is some infinite recursion.

Yes.

> Is that really something that could be considered "everything is fine now"?

Not really, that's why I suggested to have some counter/limit, and issue a warning.

It is certainly better to abort (but preferably after some rounds only) than to have to `kill -9 awesome`.

PR #2507, opened 2018-12-27
New theme API groundwork

Elv13:
This pull request extends CAPI `luaclass` to allow global connections to be introspected. This is used to allow `awful.screen.connect_for_each_screen` like signals to be created. It could eventually be extended to also filter connections, but currently it does not implement such features.

I use this to "port" the screen and wallpaper sections of `rc.lua` to use the `request::` API as used in other components such as the titlebar. With this change, we get closer to an unified "pipeline" for components initialization. The goal of the `request::` API is to allow themes and modules to cleanly integrate themselves into Awesome without too much effort from the users.

@actionless Here's a clearer view of what I propose. I know you are not sold, but I think this PR is conservative enough to avoid conflicts. The next, and more controversial given what you said in the past, would be to allow theme to deny (with a proper warning popup) `rc.lua` from adding new bars and saying they already handle that. Idem for the titlebars, notification (WIP) and wallpapers. With those capabilities in place, the themes should now finally be free to define different GUI layout than the current unmodified `rc.lua` allows.

### DIFF
--- a/common/luaclass.c
+++ b/common/luaclass.c
@@ -22,6 +22,8 @@
 #include "common/luaclass.h"
 #include "common/luaobject.h"
 
+#define CONNECTED_SUFFIX "::connected"
+
 struct lua_class_property
 {
     /** Name of the property */
@@ -300,6 +302,24 @@ luaA_class_connect_signal_from_stack(lua_State *L, lua_class_t *lua_class,
                                      const char *name, int ud)
 {
     luaA_checkfunction(L, ud);
+
+    /* Duplicate the function in the stack */
+    lua_pushvalue(L, ud);
+
+    char *buf = p_alloca(char, a_strlen(name) + a_strlen(CONNECTED_SUFFIX) + 1);
+
+    /* Create a new signal to notify there is a global connection. */
+    sprintf(buf, "%s%s", name, CONNECTED_SUFFIX);
+
+    /* Emit a signal to notify Lua of the global connection.
+     *
+     * This can useful during initialization where the signal needs to be
+     * artificially emitted for existing objects as soon as something connects
+     * to it
+     */
+    luaA_class_emit_signal(L, lua_class, buf, 1);
+
+    /* Register the signal to the CAPI list */
     signal_connect(&lua_class->signals, name, luaA_object_ref(L, ud));
 }
 
@@ -511,5 +531,7 @@ luaA_class_new(lua_State *L, lua_class_t *lua_class)
 
     return 1;
 }
+
+#undef CONNECTED_SUFFIX
 
 // vim: filetype=c:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -163,7 +163,7 @@ local tasklist_buttons = gears.table.join(
                                           end))
 
 -- @DOC_WALLPAPER@
-local function set_wallpaper(s)
+screen.connect_signal("request::wallpaper", function(s)
     -- Wallpaper
     if beautiful.wallpaper then
         local wallpaper = beautiful.wallpaper
@@ -173,16 +173,10 @@ local function set_wallpaper(s)
         end
         gears.wallpaper.maximized(wallpaper, s, true)
     end
-end
-
--- Re-set wallpaper when a screen's geometry changes (e.g. different resolution)
-screen.connect_signal("property::geometry", set_wallpaper)
+end)
 
 -- @DOC_FOR_EACH_SCREEN@
-awful.screen.connect_for_each_screen(function(s)
-    -- Wallpaper
-    set_wallpaper(s)
-
+screen.connect_signal("request::desktop_decoration", function(s)
     -- Each screen has its own tag table.
     awful.tag({ "1", "2", "3", "4", "5", "6", "7", "8", "9" }, s, awful.layout.layouts[1])
 
--- a/docs/89-NEWS.md
+++ b/docs/89-NEWS.md
@@ -14,6 +14,9 @@ This document was last updated at commit v4.3-148-g795c792d1.
 
 ## New features
 
+* `awful.screen` now has a `request::wallpaper` and a
+  `request::desktop_decoration` signal. They make some workflow implementation
+  cleaner.
 * Lua code can interact with the selection contents via the new
   `selection.acquire`, `selection.getter`, and `selection.watcher` objects
 * Pending delayed calls (`gears.timer.delayed_call`) can be dispatched via
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -602,6 +602,54 @@ function screen.object.set_dpi(s, dpi)
     s.data.dpi = dpi
 end
 
+--- Emitted when a new screen is added.
+--
+-- The handler(s) of this signal are responsible of adding elements such as
+-- bars, docks or other elements to a screen. The signal is emitted when a
+-- screen is added, including during startup.
+--
+-- The only default implementation is the one provided by `rc.lua`.
+--
+-- @signal request::desktop_decoration
+-- @tparam screen s The screen object.
+
+--- Emitted when a new screen needs a wallpaper.
+--
+-- The handler(s) of this signal are responsible to set the wallpaper. The
+-- signal is emitted when a screen is added (including at startup), when its
+-- DPI changes or when its geometry changes.
+--
+-- The only default implementation is the one provided by `rc.lua`.
+--
+-- @signal request::wallpaper
+-- @tparam screen s The screen object.
+
+-- Set the wallpaper(s) and create the bar(s) for new screens
+capi.screen.connect_signal("added", function(s)
+    s:emit_signal("request::desktop_decoration")
+    s:emit_signal("request::wallpaper")
+end)
+
+-- Resize the wallpaper(s)
+for _, prop in ipairs {"geometry", "dpi" } do
+    capi.screen.connect_signal("property::"..prop, function(s)
+        s:emit_signal("request::wallpaper")
+    end)
+end
+
+-- Create the bar for existing screens when an handler is added
+capi.screen.connect_signal("request::desktop_decoration::connected", function(new_handler)
+    for s in capi.screen do
+        new_handler(s)
+    end
+end)
+
+-- Set the wallpaper when an handler is added.
+capi.screen.connect_signal("request::wallpaper::connected", function(new_handler)
+    for s in capi.screen do
+        new_handler(s)
+    end
+end)
 
 --- When the tag history changed.
 -- @signal tag::history::update
--- a/spec/awful/screen_spec.lua
+++ b/spec/awful/screen_spec.lua
@@ -6,6 +6,7 @@
 local fake_screens = {}
 
 _G.screen = setmetatable({
+    connect_signal = function() end,
     set_index_miss_handler = function() end,
     set_newindex_miss_handler = function() end,
 }, {


common/luaclass.c
```diff
@@ -300,6 +300,27 @@ luaA_class_connect_signal_from_stack(lua_State *L, lua_class_t *lua_class,
...
 {
     luaA_checkfunction(L, ud);
+
+    /* Duplicate the function in the stack */
+    lua_pushvalue(L, ud);
+    luaA_checkfunction(L, -1);
+    luaA_checkfunction(L, ud);
+
+    /* Create a new signal to notify there is a global connection. */
+    const char buf[128];
```
actionless: should it be const?

Elv13: done

Elv13:
@psychon mind reviewing the luaclass commit?

common/luaclass.c
```diff
@@ -300,6 +300,27 @@ luaA_class_connect_signal_from_stack(lua_State *L, lua_class_t *lua_class,
                                      const char *name, int ud)
 {
     luaA_checkfunction(L, ud);
+
+    /* Duplicate the function in the stack */
+    lua_pushvalue(L, ud);
+    luaA_checkfunction(L, -1);
+    luaA_checkfunction(L, ud);
```
psychon: The two new calls to `luaA_checkfunction` check the exact same thing as the already existing `luaA_checkfunction` call. Or am I missing something?

common/luaclass.c
```diff
@@ -300,6 +300,27 @@ luaA_class_connect_signal_from_stack(lua_State *L, lua_class_t *lua_class,
...
+    a_strcpy(buf, 128, name);
+    a_strcat(buf, 128, "::connected");
+
+    /* Emit a signal to notify Lua of the global connection.
+     *
+     * This can useful during initialization where the signal needs to be
+     * artificially emitted for existing objects as soon as something connects
+     * to it
+     */
+    luaA_checkfunction(L, -1);
```
psychon: Okay, here is the fourth duplicate. This code is really making sure that it gets a function...

common/luaclass.c
```diff
@@ -300,6 +300,27 @@ luaA_class_connect_signal_from_stack(lua_State *L, lua_class_t *lua_class,
...
+    luaA_class_emit_signal(L, lua_class, buf, 1);
```
psychon: Why is this only emitted for things that connect to a class, but not when connecting to an instance?
According to `git grep`, there are four callers of `signal_connect`:

* Connecting a signal on a class (this code here)
* Connecting a signal on an object (`luaA_object_connect_signal_from_stack`)
* Connecting a signal on the `awesome` global (`luaA_awesome_connect_signal`; don't ask me why that is duplicated code)
* Connecting a signal on the `dbus` global (`luaA_dbus_connect_signal`; this code actually only allows a single function to be connected to a signal, so it might be a good idea to just ignore that)

common/luaclass.c
```diff
@@ -300,6 +300,27 @@ luaA_class_connect_signal_from_stack(lua_State *L, lua_class_t *lua_class,
...
+
+    /* Duplicate the function in the stack */
+    lua_pushvalue(L, ud);
+    luaA_checkfunction(L, -1);
+    luaA_checkfunction(L, ud);
+
+    /* Create a new signal to notify there is a global connection. */
+    char buf[128];
+    a_strcpy(buf, 128, name);
+    a_strcat(buf, 128, "::connected");
```
psychon: Can we get rid of the 128 limit?

Untested:
```C
#define CONNECTED_SUFFIX "::connected"
char *buf = p_alloca(char, a_strlen(name) + a_strlen(CONNECTED_SUFFIX) + 1);
sprintf(buf, "%s%s", name, CONNECTED_SUFFIX);
#undef CONNECTED_SUFFIX
```

common/luaclass.c
Elv13: This is out of scope for this pull request

common/luaclass.c
Elv13: done

common/luaclass.c
Elv13: I went for growing a static buffer. Might not be very clean when quitting Awesome but at least it doesn't do 1k small allocs at startup (yes, my config is big, I know).

Elv13:
> I only looked briefly at what the rest is doing and I can't say I'm convinced.

Ok, some context from the maze of comments that led to this. This is related to #442. Me and @actionless have been discussing in various places, including emails about how to make Awesome more extendable. My position is that the bar and GUI stuff should move to the theme. His position is that it should not. However, once thing we agree on, and Sorky too, is that business logic should not exist in `rc.lua`. As long as there are local functions in `rc.lua`, modules integrate cleanly without the user having to manually patch her/his config file.

### Example #1: the wallpaper

Right now if a module or theme tries to set, let say, a tiled wallpaper, that module either will have to fight `rc.lua` using timers or ask the user to change `set_wallpaper` to use the tiled mode. Other hacks such as creating a surface, manually printing a tiled pattern and using that as `beautiful.wallpaper` is equally ugly. 

(Assuming the ability to block connections is not part of this PR (and may never be submitted at all))
With this pull request, you can ask the user to delete the `request::wallpaper` and the one from the theme will be used. No need to mess with the `connect_for_each_screen` section. It is much cleaner

### Example #2: the error handling

Right now we have:

```lua
-- {{{ Error handling
-- @DOC_ERROR_HANDLING@
-- Check if awesome encountered an error during startup and fell back to
-- another config (This code will only ever execute for the fallback config)
if awesome.startup_errors then
    naughty.notify({ preset = naughty.config.presets.critical,
                     title = "Oops, there were errors during startup!",
                     text = awesome.startup_errors })
end

-- Handle runtime errors after startup
do
    local in_error = false
    awesome.connect_signal("debug::error", function (err)
        -- Make sure we don't go into an endless error loop
        if in_error then return end
        in_error = true

        naughty.notify({ preset = naughty.config.presets.critical,
                         title = "Oops, an error happened!",
                         text = tostring(err) })
        in_error = false
    end)
end
-- }}}
```

if it was done for the notification too, we end up with:

```lua
naughty.connect_signal("display::error", function(err)
    naughty.notification { 
        preset = naughty.config.presets.critical,
        title  = "Oops, an error happened!",
        text   = err
    }
end)
```

In this example, all the business logic is gone, you could cut&paste the section in a new file and it would just work.

Now if you extrapolate this to all of `rc.lua` (of which large parts already use the `request::` API), you end up with small independent sections that can be moved to other files by simply cut&paste-ing them. Once we get there, #442 can finally be closed. 

This pull request is about de-coupling `rc.lua` and the first use case of this feature happens to be related to the theme (which in my world view includes the bar).

common/luaclass.c
psychon: `p_alloca` allocates stack space, not heap space...?

Elv13: Is this like "elv13 please just use p_alloca, vla are cool" or is it a generic comment?

psychon: I'm fine with `p_alloca` and `p_alloca` does not use a variable length array. So yes, they are cool. Also, `alloca` is a lot cheaper than `malloc` (it's just subtracting a number from the stack pointer to reserve some stack space), so I didn't understand the comment about the 1k small allocs. Also, I didn't look at that static buffer patch before now and I think `p_alloca` is simpler than that.

Elv13: done

psychon:
Okay, so... some signals now become "magical" and whenever you connect to them, some random function invocations will happen ("all the already-existing objects" in what this PR does). So, this makes signals more complex and even if you know how signal work, you now have to scan all code to see how this one signal behaves.
This does indeed make some things easier (as in "less code"), but also makes them harder (as in "what the heck is going on?!?").

> Example #1: the wallpaper
> Right now if a module or theme tries to set, let say, a tiled wallpaper, that module either will have to fight rc.lua using timers or ask the user to change set_wallpaper to use the tiled mode. Other hacks such as creating a surface, manually printing a tiled pattern and using that as beautiful.wallpaper is equally ugly.

```diff
diff --git a/awesomerc.lua b/awesomerc.lua
index fa584b8a8..6cf84bc55 100644
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -165,7 +165,9 @@ local tasklist_buttons = gears.table.join(
 -- @DOC_WALLPAPER@
 local function set_wallpaper(s)
     -- Wallpaper
-    if beautiful.wallpaper then
+    if beautiful.set_wallpaper then
+        beautiful.set_wallpaper(s)
+    elseif beautiful.wallpaper then
         local wallpaper = beautiful.wallpaper
         -- If wallpaper is a function, call it with the screen
         if type(wallpaper) == "function" then
diff --git a/themes/default/theme.lua b/themes/default/theme.lua
index b0be48012..995a9e065 100644
--- a/themes/default/theme.lua
+++ b/themes/default/theme.lua
@@ -98,6 +98,9 @@ theme.titlebar_maximized_button_normal_active = themes_path.."default/titlebar/m
 theme.titlebar_maximized_button_focus_active  = themes_path.."default/titlebar/maximized_focus_active.png"
 
 theme.wallpaper = themes_path.."default/background.png"
+function theme.set_wallpaper(s)
+    gears.wallpaper.tiled(theme.wallpaper, s, true)
+end
 
 -- You can use your own layout icons like this:
 theme.layout_fairh = themes_path.."default/layouts/fairhw.png"
```
This is fairly self-explanatory for anyone who takes a look and tries to understand what is going on.

> Example #2: the error handling
> Right now we have:

So... you basically move the check for `awesome.startup_errors` elsewhere?
As untested as the previous patch, but this now means that people with the "old config" get startup errors displayed twice:
```diff
diff --git a/luaa.c b/luaa.c
index 803d44d01..65329403b 100644
--- a/luaa.c
+++ b/luaa.c
@@ -81,6 +81,8 @@ extern const struct luaL_Reg awesome_root_lib[];
 extern const struct luaL_Reg awesome_mouse_methods[];
 extern const struct luaL_Reg awesome_mouse_meta[];
 
+static gboolean emit_startup_errors(gpointer user_data);
+
 /** A call into the Lua code aborted with an error.
  *
  * This signal is used in the example configuration, @{05-awesomerc.md},
@@ -973,6 +975,8 @@ luaA_init(xdgHandle* xdg, string_array_t *searchpath)
         { NULL, NULL }
     };
 
+    g_idle_add(emit_startup_errors, NULL);
+
     L = globalconf.L.real_L_dont_use_directly = luaL_newstate();
 
     /* Set panic function */
@@ -1069,6 +1073,18 @@ luaA_init(xdgHandle* xdg, string_array_t *searchpath)
     lua_pop(L, 1); /* pop "package" */
 }
 
+static gboolean
+emit_startup_errors(gpointer user_data)
+{
+    if (globalconf.startup_errors.len > 0)
+    {
+        lua_State *L = globalconf_get_lua_State();
+        lua_pushstring(L, globalconf.startup_errors.s);
+        signal_object_emit(L, &global_signals, "debug::error", 1);
+    }
+    return G_SOURCE_REMOVE;
+}
+
 static void
 luaA_startup_error(const char *err)
 {
```

Elv13:
I disagree with your alternative suggestions. They make the problem I am trying to fix worst. The objective is to eliminate the imperative business logic from `rc.lua`, your `set_wallpaper` suggestions adds more.

To move toward fixing #442, we first need to turn every section into independent blocks that can be cut-pasted into other files easily. This requires each block to be uncoupled from each other. The most "idiomatic" way to get there considering the existing code patterns is to add more `request::` to perform an inversion of responsibilities. Instead of `rc.lua` stating "do this", then the section requests "please, can anyone specify this?" and `rc.lua` answer this request.

In this pull request, I indeed make the signals more complex. However keep in mind that it is a private API and exists to mitigate a race condition during initialization (and no, more delayed_call wont help, they will add more corner case and nested delayed calls). This is the same use case for which you created `awful.screen.connect_for_each_screen`. But rather than this extra layer of complexity, it is now integrated into the core "signals" concept. **From the point of view of the user**, this is much more natural and obvious than another concept such as `connect_for_each_screen`. Just like the other pull request removes `gears.table.join` from `rc.lua`, the reason why I want this to be merged is to simplify, unify and modularize `rc.lua`.

psychon:
>  we first need to turn every section into independent blocks that can be cut-pasted into other files easily. This requires each block to be uncoupled from each other. The most "idiomatic" way to get there considering the existing code patterns is to add more request:: to perform an inversion of responsibilities. Instead of rc.lua stating "do this", then the section requests "please, can anyone specify this?" and rc.lua answer this request.

Doesn't this contradict what you are trying to do here? So, if I move this stuff around and a theme adds another place that reacts, I end up with two places with set the wallpaper.

We need to decide who gets to decide the wallpaper. The current answer is `rc.lua`. If you want to change the answer to `theme.lua`, then that would mean removing all wallpaper-code from the `rc.lua` and then have the theme do it. And to set a wallpaper, a theme can just use `connect_for_each_screen`.

With your patch, the whole question is turned into someone else's problem. Something like: "There is now a `request::wallpaper` where you can set a wallpaper. If multiple pieces of code connect to that, it's up to something semi-random to decide who wins (I guess something like `pairs` iteration order will make the call here).

If you want to move the wallpaper code to the theme, you should do so.

Coming back to the original quote:

> To move toward fixing #442, we first need to turn every section into independent blocks that can be cut-pasted into other files easily. This requires each block to be uncoupled from each other.

Okay, so remove all current wallpaper code from `rc.lua` and add it to `wallpaper-rc.lua`. I do not see how this PR or `request::`-style API makes that problem simpler. If multiple pieces of code want to handle these signals, the result will be confusing and not easy to debug.

Elv13:
> Doesn't this contradict what you are trying to do here?

It does indeed, but as mentioned in an earlier comment, this is "on purpose". It is also why the pull request is called groundwork. The idea here is only to move the current sections into blocks that *can* be moved. My end game is more powerful themes and simpler `rc.lua`. As long as @actionless is opposed, I don't plan to move the wallpaper, wibar, menu and titlebar to the theme. This work (and its follow-up, #2508) are just cleanup steps toward the goal. They are the parts where there is an agreement, at least with him. 

I am not imagining the issue. Reddit r/awesomewm and stackoverflow have questions from existing users who managed to get past the initial learning curve. It's an echo chamber where there is at least some assumption of existing knowledge about AwesomeWM design. On the other hand, places like the Arch forum, Gentoo forum, HackerNews and Reddit r/unixporn and r/linux are better sources to follow the frustration of new users. When AwesomeWM gets discussed there, there is a clear pattern of complains about how hard it is to learn and how hard it is to cherry pick content from other people config (and all the classic "they broke my desktop with X update", **** the developers). Some of that complexity will never go away, but that doesn't mean there's nothing to do to improve the status quo.

> And to set a wallpaper, a theme can just use connect_for_each_screen.

This API exist to solve the fact that the old (3.0-3.5) `rc.lua` used a static loop for the screen initialization. This loop is replaced 1:1 with connect_for_each_screen. This isn't mapping correctly to the concepts within this loop. The most obvious argument is that `rc.lua` has a second 

    screen.connect_signal("property::geometry", set_wallpaper)

line just for this. So using only `connect_for_each_screen` clearly isn't conveying the meaning of when the wallpaper needs to be updated. It is not the case for the titlebar either.

As a second argument, the titlebars are not only created when a new client arrives, but also when an existing client get a titlebar for the first time (like with a keybinding or, in theory, if it starts fullscreen and become tiled). So creating them in `manage` could not work. This is exactly why `request::titlebars` exist. The wallpaper are just "titlebars for the root window".

The same can be said if we ever allow the theme to be replaced at runtime or if the screen DPI changes. If those 2 were added to `rc.lua`, the wallpaper function would now be called from 4 places... 

This clearly show that the current approach isn't abstract enough for `rc.lua`. The lack of abstraction forces an ever increasing amount of wallpaper logic every time we add new features (like the dynamic screen addition and dynamic resize).

`request::wallpaper` and `request::decoration` are real abstractions for the concept they carry. They also fit in the existing framework we have for other concepts such as `request::titlebars`. `connect_for_each_screen` does neither of those. Its just a wrapper to mitigate the fact that the signals are not powerful enough to handle reality. It adds an useless and leaky new concept and confusion to `rc.lua` and must die. Instead of going down that path, this pull request makes the signal good enough and fix the problem forever.

psychon:
After sleeping about this for a bit.... well, I don't know.

The original comment when this PR was posted states:

> I use this to "port" the screen and wallpaper sections of rc.lua to use the request:: API as used in other components such as the titlebar

This could be achieved with a single delayed call that emits `request::wallpaper` once for all existing screens after awesomeWM starts up.

> The goal of the request:: API is to allow themes and modules to cleanly integrate themselves into Awesome without too much effort from the users.

I don't really buy this. If three places try to set a wallpaper, then only one of them can actually do so. I fail to see the cleanness here. If you want a single place where the wallpaper can be configured, then we need some global variable that can be set. That way, multiple pieces will still fight and overwrite each other when setting this global variable, but at least this variable is the "obvious point" where this is decided. 

Later comment:

> I disagree with your alternative suggestions. They make the problem I am trying to fix worst. The objective is to eliminate the imperative business logic from rc.lua, your set_wallpaper suggestions adds more.

So, this global variable would need to be in some module, then. Or perhaps the theme, since apparently that is the place that should decide everything about the wallpaper. (Even though "My position is that the bar and GUI stuff should move to the theme. His position is that it should not. ", so apparently it should be easily possible to decide the wallpaper in multiple places...)

Elv13:
> How about an explicit request::wibar or request::wibars?

I am not sure. The reason why I didn't use `bar` in the first place is because not every kind of screen decorations are wibars. It can also be Conky like "desktop applets". It could also be docks, desktop icons or CDE/E16 style "desktop expose" applets. `decoration` might not be the best name either. Any other idea for a better name?

psychon:
Personally, "decoration" reminds me of "window decorations", which is not what this is about... What we call "wibar", others call "panel", so `request::panel`, but that has the same problem of "conky is not a panel"... For your other examples, I cannot really come up with a category containing these. Since there is not that much precedence for these, perhaps we should just ignore desktop windows when coming up with a name? People who try to use the signal for that likely know what they are doing and so are not as easily confused by the name (or rather: they don't care about the name since they know what it is doing anyway).

Elv13:
@actionless Do you have an idea about a name?

Veratil:
'desktop_element' is the best I can think of for descibing all these types of things.

psychon:
Perhaps `desktop_decoration`? That is again `decoration`, but makes it clearer that the desktop is being decorated...?

Elv13:
I renamed `request::decoration` to `request::desktop_decoration` as suggested by @psychon (which in turn is based upon @Veratil suggestion). I also added a `request::wallpaper` signal when the DPI changes to sync the code with some of the comments. I also added a new documentation sentence to each request signals to improve clarity as to when the signal are sent.

Elv13:
If no one else object, I will assume everybody is now happy and merge this soon.

PR #2508, opened 2018-12-27
New round of API standardization (for v4.4)

Elv13:
Here's a scary one.

This PR "fixes" the issue regarding `awful.button`/`awful.key` vs. `capi.button`/`capi.button` and starts to cleanup the very old luaobject read/write accessor methods (like `c:tags()` instead is `c.tags`).

The button/key work is the follow up of the keygrabber PR I did last summer where I mentioned managing keybindings sucked. This PR does not addresses this issue, but it removes a limitation that prevents the issue from being fixed. Currently `awful.key` and `awful.button` are not objects, but rather tuples of many `capi.key`/`capi.button` objects. They are expanded by `gears.table.join` and any metadata associated with the object is now lost. @actionless hotkey popup work around that by keeping an index of the keys it thinks are set, but it isn't fully reliable. Step by step, this work will allow to really use the `awful.key` and `awful.button` objects to store a state and metadata.

The other parts is related to this by refactoring the API so `gears.table.join` is no longer necessary and while at it convert with methods into properties (it is backward compatible).

Finally, there is the first part of a bigger cleanup of `rc.lua` to use the same API style for everything. As discussed with @actionless over the last couple years, we think moving all constructors to support named argument is a cleaner way forward rather than random constructor arguments. Once all widgets (and hopefully relevant objects) are ported then the declarative and "normal" constructor can be unified so they become interchangeable and less confusing. This PR does *not* address this.

### DIFF
--- a/lib/awful/widget/common.lua
+++ b/lib/awful/widget/common.lua
@@ -129,7 +129,7 @@ function common.list_update(w, buttons, label, data, objects, args)
             cache = (args and args.widget_template) and
                 custom_template(args) or default_template()
 
-            cache.primary:buttons(common.create_buttons(buttons, o))
+            cache.primary.buttons = {common.create_buttons(buttons, o)}
 
             if cache.create_callback then
                 cache.create_callback(cache.primary, o, i, objects)
--- a/lib/wibox/widget/base.lua
+++ b/lib/wibox/widget/base.lua
@@ -155,6 +155,24 @@ function base.widget:get_all_children()
     return ret
 end
 
+--- Common implementation of the `:set_widget()` method exposed by many
+-- other widgets.
+--
+-- Use this if your widget has no custom logic when setting the widget.
+--
+-- @usage
+--   rawset(my_custom_widget, "set_widget", wibox.widget.base.set_widget_common)
+function base.set_widget_common(self, widget)
+    local w = widget and base.make_widget_from_value(widget)
+
+    if w then
+        base.check_widget(w)
+    end
+
+    self._private.widget = w
+    self:emit_signal("widget::layout_changed")
+end
+
 --- Emit a signal and ensure all parent widgets in the hierarchies also
 -- forward the signal. This is useful to track signals when there is a dynamic
 -- set of containers and layouts wrapping the widget.
--- a/lib/wibox/widget/base.lua
+++ b/lib/wibox/widget/base.lua
@@ -40,6 +40,29 @@ function base.widget:set_visible(b)
     end
 end
 
+--- Add a new `awful.button` to this widget.
+-- @tparam awful.button button The button to add.
+function base.widget:add_button(button)
+    if not button then return end
+
+    -- Simple case
+    if not self._private.buttons then
+        self:set_buttons({button})
+        return
+    end
+
+    -- This could happen if something accidentally uses rawset
+    assert(self._private.buttons_formatted)
+
+    -- an `awful.button` is a tupple of `capi.button`
+    self._private.buttons_formatted = gtable.join(
+        self._private.buttons_formatted,
+        button
+    )
+
+    table.insert(self._private.buttons, button)
+end
+
 --- Is the widget visible?
 -- @treturn boolean
 -- @method get_visible
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -219,11 +219,11 @@ end)
 
 -- {{{ Mouse bindings
 -- @DOC_ROOT_BUTTONS@
-root.buttons(gears.table.join(
+root.buttons = {
     awful.button({ }, 3, function () mymainmenu:toggle() end),
     awful.button({ }, 4, awful.tag.viewnext),
-    awful.button({ }, 5, awful.tag.viewprev)
-))
+    awful.button({ }, 5, awful.tag.viewprev),
+}
 -- }}}
 
 -- {{{ Key bindings
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -228,7 +228,7 @@ root.buttons = {
 
 -- {{{ Key bindings
 -- @DOC_GLOBAL_KEYBINDINGS@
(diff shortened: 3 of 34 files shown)


root.c
```diff
@@ -526,19 +541,92 @@ luaA_root_tags(lua_State *L)
...
+{
+    /* Call the lua root property handler */
+    if (miss_newindex_handler != LUA_REFNIL)
+        return luaA_call_handler(L, miss_newindex_handler);
+
+    return luaA_default_newindex(L);
+}
+
+static int
+luaA_root_new(lua_State *L)
```
actionless:
```
/tmp/awesome-build/root.c:600:1: error: ‘luaA_root_new’ defined but not used [-Werror=unused-function]
 luaA_root_new(lua_State *L)
 ^
cc1: all warnings being treated as errors
```

Elv13: done

Elv13:
I added a `widget:add_button()` method while porting everything away from the legacy API. It removes some of the ugly `w:buttons(gear.table.join(w:buttons(), new_button))` code. My original goal was to do that later, but the PR wont get green as long as I don't remove all deprecated code...

lib/awful/widget/common.lua
```diff
@@ -21,17 +21,23 @@ local common = {}
...
-            -- Create a proxy button object: it will receive the real
-            -- press and release events, and will propagate them to the
-            -- button object the user provided, but with the object as
-            -- argument.
-            local btn = capi.button { modifiers = b.modifiers, button = b.button }
-            btn:connect_signal("press", function () b:emit_signal("press", object) end)
-            btn:connect_signal("release", function () b:emit_signal("release", object) end)
-            btns[#btns + 1] = btn
+        for _, src in ipairs(buttons) do
+            --TODO v6 Remove this legacy overhead
```
Veratil: Actual v6 or typo v5?

lib/wibox/widget/base.lua
```diff
@@ -700,6 +736,20 @@ function base.make_widget(proxy, widget_name, args)
     mt.__tostring = function()
         return string.format("%s (%s)", ret.widget_name, orig_string)
     end
+
+    -- Even when properties are disabled, buttons is required for backward
+    -- compatibility.
+    --TODO v6 Remove this
```
Veratil: Actual v6 or typo v5?

Elv13: Actual v6. This API is very hard to both modernize and keep backward compatibility. The goal of this branch is to have a real read/write API for input handling. The old one is write only. Once you set the keys/buttons, they lose their context and you have to reverse engineer everything to manipulate them later on. It will need long term code to support the old deprecated API. Both for 4.x (not officially deprecated yet) and 5.x (deprecated but supported) so we can remove it from 6.0 (deprecated and free to break if the maintenance cost reduction is worth it, which in this case it totally will be).

/offtopic

This isn't really a pull request anyway. It's more like a branch with many, many small, long term, independent, commits that can be reviewed individually (as long as the order remain the same). Once I will be done with this, there will be a real API to manipulate input events. But given that I want to create something that's easier to review than my usual multi-hundred lines commits, its going to take some iterations of changes within this branch. I add new APIs in some commit, then "slowly"  integrate them, then tests them, then replace the usage of the old one, then deprecate it. All in different commits. It will most probably have >100 commits, but commits that I hope can be reviewed/merged with less pain than my usual work.

psychon doesn't have time to review things right now (which is 100% ok, I am in the same situation most of the year), it's better to spend my AwesomeWM time on this than on the notification PR series because it is stuck as long as the open ones are not merged. That, in turn requires both of us active at the same time, which hasn't been happening a lot this year I know you and others want the notification API and don't care about much about API standardization. The problem is that stars will have to align to be able to merge that work and it isn't the case right now.

Veratil: > Actual v6. This API is very hard to both modernize and keep backward compatibility. The goal of this branch is to have a real read/write API for input handling. The old one is write only. Once you set the keys/buttons, they lose their context and you have to reverse engineer everything to manipulate them later on. It will need long term code to support the old deprecated API. Both for 4.x (not officially deprecated yet) and 5.x (deprecated but supported) so we can remove it from 6.0 (deprecated and free to break if the maintenance cost reduction is worth it, which in this case it totally will be).

Got it! Just curious when looking over it.

> psychon doesn't have time to review things right now (which is 100% ok, I am in the same situation most of the year), it's better to spend my AwesomeWM time on this than on the notification PR series because it is stuck as long as the open ones are not merged. That, in turn requires both of us active at the same time, which hasn't been happening a lot this year

Yeah, I've been focused on other things, too. Haven't touched awesome source in a while.

> I know you and others want the notification API and don't care about much about API standardization.

On the contrary! I very much am supportive for API standardization. I'm just not as versed in the core source as y'all are. I try where I can. :)

Elv13:
@psychon Mind reviewing this one? You wont like it, neither do I. It's everything wrong with "backward compatibility at all cost". However it is a step toward removing a whole class of very ugly APIs and getting closer to having no "write only" APIs like `root.keys()` and `root.buttons()`. It also makes `awful` a real dependency of `capi`, at least for the public API, which is also ugly. Again, there's a reason for that and it's rather obvious when you see the changes to `rc.lua`.

Elv13:
Rebased. It will probably take a few iterations to bring this back to life and pass the tests.

This is the "next big thing" I plan to merge. This is required for my last few screen pull requests. As I said a few time about this pull request, I don't really like how it is done. It's a hack and will be hard to maintain. However the end result is getting rid of a lot of technical debt from the pre-property era of the API. Once this pull request and its few follow up are merged (not planned for v4.4), the API will be *a lot* more consistent. Plus, some features I wanted to add for a while, such as QAction/GtkAction equivalent will become possible.

Elv13:
Last call. If nobody steps in, I will start to merge this in chunks (to make the CI pass) tomorrow.

PR #2509, opened 2018-12-27
Fix some tag.layouts corner cases

Elv13:
The previous PR missed a few case where things like the layout incrementation would not handle the per-tag layouts.

I also cleanup `awful.layout` while at it


Elv13:
Rebased to fix a merge conflict and do a styling/dead_code change proposed by @psychon on IRC. If there isn't any more comments, I will merge this soon.

Issue #667, opened 2016-02-05
Ontop client bug


flyboy14:
When I make an ontop client window fullscreen and then back to non-fullscreen state, it also appears not to be ontop anymore.

psychon:
This is done on purpose: https://github.com/awesomeWM/awesome/blob/cd63cabaddabd61bbbea44bf4dc8a4f575901e92/objects/client.c#L1112-L1119
Alternatively, we would have to define an arbitrary "stronger than" relation between these properties to decide which one is ignored when more than one is active.

flyboy14:
So, if I understood correctly, client couldn't be both fullscreen and ontop at the same time?

psychon:
Yup.

Awesome thinks in layers. At the very top are "ontop" windows. Below that come "fullscreen" windows, then "above" windows. Then "normal", "below" and "desktop" (in this order). That's what the comment that I linked to above means with "a window cannot be in multiple layers at the same time". Put differently, a fullscreen client is automatically something like ontop. (However, we do allow clients with type "desktop" to be "ontop" and ignore their desktop property as long as they are ontop, so perhaps something similar could be done here... but would that make sense? I'm not sure)

And yet I also think that this bug report has a valid point (unexpected behaviour, but "it's a feature, not a bug" :-) ).

blueyed:
@psychon 
> define an arbitrary "stronger than" relation between these properties to decide which one is ignored when more than one is active

I'd vote for that..

Elv13:
@blueyed Then we would be better to just implement z-index and emulate the current layers on top of that. Most WMs have those ontop/normal/above/below/desktop layer concepts. Also, I don't mind the current behavior, it work fine for me.

rogorido:
@flyboy14 this is already closed, and this is only a workaround but it seems to work for me:

```
client.connect_signal("property::fullscreen",
		      function(c)
			 if c.class == "mpv" and c.fullscreen == false then
			    c.ontop = true
			 end
		      end)
```

blueyed:
@rogorido 
For this to work you need to be aware that a client should be ontop after all.

@Elv13 
> Most WMs have those ontop/normal/above/below/desktop layer concepts.

Sure.  But this is about internal implementation, and a simple fix for this might really be to allow having "fullscreen = true" and "ontop = true" set at the same time, but then only use "ontop" for stacking.

> Also, I don't mind the current behavior, it work fine for me.

[...]

Issue #2047, opened 2017-09-28
True fullscreen -> disable notifications 


kajzersoze:
Output of `awesome --version`:

awesome v4.2 (Human after all)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.0

**How to reproduce the issue:**

Start Kodi , in settings for display select fullscreen.

**Actual result:**

Desktop notifications are visible.

**Expected result:**

Desktop notifications are not visible.

I'm new to AwesomeWM and I've noticed that there is no true fullscreen, for example in Kodi while watching some video or while playing some game in Steam I can still see desktop notifications (they are not visible in other DE's while in fullscreen) , also because of that I can see slight lag in Kodi, same everything (distro and hardware) I just logout from awesomewm and login to i3wm for example (same with mate, xfce, gnome, plasma) and it works fine there.
Apart from that I really like AwesomeWM, but this fullscreen issue is just not good for me, I tried to google about it but couldn't find anything, am I missing something or it's like that and that's it?

kajzersoze:
Maybe naughty needs something like[ this](https://github.com/mate-desktop/mate-notification-daemon/blob/c556e09077bb820e752d7266d2b20328db629883/src/daemon/daemon.c#L1630) , to detect if fullscreen is active and if so to disable notifications in that case (except if timeout for that notification is 0, in which case it will be shown)

blueyed:
ah.. so you only see naughty's notifications then?
I've thought it was about notifications in general.

kajzersoze:
In other DE's I don't see them in fullscreen (except if timeout is 0), in Awesome I see them, I thought that it had something to do with fullscreen code somewhere since I've seen lots of other issues regarding fullscreen and because I can notice a slight lag while watching videos in fullscreen (minor but still noticeable), same software same movie different wm and there's no lag.

Elv13:
[...]

PR #2512, opened 2018-12-27
client: keep ontop/fullscreen/above/below when setting the others

blueyed:
Fixes https://github.com/awesomeWM/awesome/issues/667.

This does not break any test.

It should have a new test though when accepted.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -2078,13 +2078,6 @@ client_set_fullscreen(lua_State *L, int cidx, bool s)
     if(c->fullscreen != s)
     {
         /* become fullscreen! */
-        if(s)
-        {
-            /* You can only be part of one of the special layers. */
-            client_set_below(L, cidx, false);
-            client_set_above(L, cidx, false);
-            client_set_ontop(L, cidx, false);
-        }
         int abs_cidx = luaA_absindex(L, cidx); \
         lua_pushstring(L, "fullscreen");
         c->fullscreen = s;
@@ -2183,13 +2176,6 @@ client_set_above(lua_State *L, int cidx, bool s)
 
     if(c->above != s)
     {
-        /* You can only be part of one of the special layers. */
-        if(s)
-        {
-            client_set_below(L, cidx, false);
-            client_set_ontop(L, cidx, false);
-            client_set_fullscreen(L, cidx, false);
-        }
         c->above = s;
         stack_windows();
         luaA_object_emit_signal(L, cidx, "property::above", 0);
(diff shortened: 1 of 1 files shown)


blueyed:
Let's wait for @psychon to approve/review it.

I think it is a good start, and awesome itself would still behave the same.

Note that hacks similar to https://github.com/awesomeWM/awesome/issues/667#issuecomment-212395679 might break due to this, e.g. if you expect "fullscreen" to be unset when "ontop" gets set etc, but I think that's fine / easy to fix then.

psychon:
Fine by me, if you want to stay around and fix the resulting problems.

@Elv13 Is this allowed in a minor release or is this possibly-breaking enough to need to wait for v5?

psychon:
Oh and what will be your answer to the incoming bug reports saying stuff like "I tried to set a client below, but nothing happened"?

blueyed:
> I tried to set a client below, but nothing happened

Good point.
Maybe this should only take out "fullscreen" then for now?  (i.e. fullscreen is kep when setting the others and does not reset the others)
All the others define the stacking only, while "fullscreen" can be more seen like "maximized" in this sense.

Elv13:
> @Elv13 Is this allowed in a minor release or is this possibly-breaking enough to need to wait for v5?

I think the current API can be implemented on top of the new one but deprecated. I can take care of that part if you submit such API changes. Adding convoluted maintenance heavy code to support our legacy features is something I have done a lot ;) (just see that ugly `c:keys()` hack in the other PR).

blueyed:
@actionless 
What do you think about https://github.com/awesomeWM/awesome/pull/2512#issuecomment-450157426?

@Elv13 
> I think the current API can be implemented on top of the new one but deprecated.

Do you mean to keep the current behavior on the Lua side, based on this PR / removal / internal change?

actionless:
@blueyed yes, that sounds more logical to me

actionless:
@Elv13 was the current behavior documented somewhere to deserve to be preserved? for me it looks more like a bug

Elv13:
@actionless looking back at the last time I modified the maximize behavior because people reported "bugs" about it, I can say that changing these things quickly becomes a rabbit hole. People will notice and wont be very happy. Right now the design of this feature is "broken" in the sense people can't make it behave like they want. The only reason few notice is because the tiled client tweak their indices in the client list to avoid the bug(s) from being visible. I think it should be fixed properly because "random" changes in behavior will just cause more open issues with more user requests and more changes to accommodate the user whose favorite behavior broke, then more bugs due to the added complexity.

Elv13:
@actionless as long as we wait for after 4.3 is released I am ok with trying. However all comments near the end of #2047 still stand. Doing this a few days before releasing 4.3 is too risky.

actionless:
i know about z-index thing, i think we were discussing that even more times longer ago

however speaking of this particular issue (especially not the actual diff but amended version proposed by @blueyed  in the comment above):
i can't imagine how reseting ontop state after making client fullscreen could be an expected behavior

however i agree what resetting below state on fullscreen could be expected by some people

Issue #63, opened 2014-11-24
AltGr shortcut combinations not work


ivanbalashov239:
i use dvorak layout(but it is not important for this question)
and on my 3 level of layout i have useful symbols like digits and arrows
so
AltGr + a = 4
AltGr + o = 5
AltGr + e = 6

AltGr + h = left
AltGr + t = Down
AltGr + n = UP
AltGr + s = Right

But, in awesome wm 3 level doesn't work, and i don't understand why.
for example
    awful.key({ modkey }, "Left",   awful.tag.viewprev       ),
    awful.key({ modkey }, "Right",  awful.tag.viewnext       ),
it's works with simple arrows on keyboard, but not works with level3 arrows.

PR #2515, opened 2018-12-27
awful.key: add mod5/ISO_Level3_Shift to pc105 conversion

blueyed:
Not sure if this is universal/correct though?!

Ref: https://github.com/awesomeWM/awesome/issues/63.


Elv13:
I added a new pull request to generate this list using the current keyboard map. @psychon wont be happy because it uses the ancient API, but it solves this permanently instead of unreliable hacks

Elv13:
Closing because #2521 supersede this and removed the code being modified here.

PR #2521, opened 2018-12-30
awesome: Add a `modifiers` field.

Elv13:
Yes, it uses XLib, I am evil. Beside that, it allows higher level APIs
such as the keygrabber to detect when a key event is a modifier state
change. It allows keygrabber to exit when "the Mod4 modifier" is
released.

Ref #2515
Ref https://github.com/awesomeWM/awesome/pull/2315/files#r205010097

### DIFF
--- a/lib/awful/key.lua
+++ b/lib/awful/key.lua
@@ -9,14 +9,11 @@
 -- Grab environment we need
 local setmetatable = setmetatable
 local ipairs = ipairs
-local capi = { key = key, root = root }
+local capi = { key = key, root = root, awesome = awesome }
 local gmath = require("gears.math")
 local gtable = require("gears.table")
 
-
-
 local key = { mt = {}, hotkeys = {} }
-
 
 --- Modifiers to ignore.
 -- By default this is initialized as { "Lock", "Mod2" }
@@ -27,23 +24,53 @@ local key = { mt = {}, hotkeys = {} }
 key.ignore_modifiers = { "Lock", "Mod2" }
 
 --- Convert the modifiers into pc105 key names
-local conversion = {
-    mod4    = "Super_L",
-    control = "Control_L",
-    shift   = "Shift_L",
-    mod1    = "Alt_L",
-}
+local conversion = nil
+
+local function generate_conversion_map()
+    if conversion then return conversion end
+
+    local mods = capi.awesome._modifiers
+    assert(mods)
+
+    conversion = {}
+
+    for mod, keysyms in pairs(mods) do
+        for _, keysym in ipairs(keysyms) do
+            assert(keysym.keysym)
+            conversion[mod] = conversion[mod] or keysym.keysym
+            conversion[keysym.keysym] = mod
+        end
+    end
+
+    return conversion
+end
+
+capi.awesome.connect_signal("xkb::map_changed"  , function() conversion = nil end)
 
 --- Execute a key combination.
 -- If an awesome keybinding is assigned to the combination, it should be
 -- executed.
+--
+-- To limit the chances of accidentally leaving a modifier key locked when
+-- calling this function from a keybinding, make sure is attached to the
+-- release event and not the press event.
+--
 -- @see root.fake_input
 -- @tparam table mod A modified table. Valid modifiers are: Any, Mod1,
 --   Mod2, Mod3, Mod4, Mod5, Shift, Lock and Control.
 -- @tparam string k The key
 function key.execute(mod, k)
+    local modmap = generate_conversion_map()
+    local active = capi.awesome._active_modifiers
+
+    -- Release all modifiers
+    for _, m in ipairs(active) do
+        assert(modmap[m])
+        root.fake_input("key_release", modmap[m])
+    end
+
     for _, v in ipairs(mod) do
-        local m = conversion[v:lower()]
+        local m = modmap[v]
         if m then
             root.fake_input("key_press", m)
         end
@@ -53,10 +80,17 @@ function key.execute(mod, k)
     root.fake_input("key_release", k)
 
     for _, v in ipairs(mod) do
-        local m = conversion[v:lower()]
+        local m = modmap[v]
         if m then
             root.fake_input("key_release", m)
         end
+    end
+
+    -- Restore the previous modifiers all modifiers. Please note that yes,
+    -- there is a race condition if the user was fast enough to release the
+    -- key during this operation.
+    for _, m in ipairs(active) do
+        root.fake_input("key_press", modmap[m])
     end
 end
 
--- a/luaa.c
+++ b/luaa.c
@@ -66,6 +66,12 @@
 #include <xcb/xcb_atom.h>
 #include <xcb/xcb_aux.h>
 
+/* XGetModifierMapping() */
+#include <X11/Xlib.h>
+
+/* XkbKeycodeToKeysym */
+#include <X11/XKBlib.h>
+
 #include <unistd.h> /* for gethostname() */
 
 #ifdef WITH_DBUS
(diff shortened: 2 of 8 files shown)


psychon:
Why the Xlib dependency?
- `XGetModifierMapping` does the same thing as `xcb_get_modifier_mapping` (well, basically): https://sources.debian.org/src/libx11/2:1.6.7-1/src/ModMap.c/?hl=34#L34
- `XkbKeycodeToKeysym` is a bit harder to follow, but should also be possible to do with `xcb-xkb`

Also, why are you hardcoding group and level to 0 for `XkbKeycodeToKeysym`? Is it just "works for me" or are you copying code from somewhere else? (Is the license of "somewhere else" compatible with awesomeWM's license?)

psychon:
Oh and: Thanks a lot for digging out some magic way to get this information (other than hardcoding some assumptions about the keyboard)

actionless:
oh, that would be really cool to have, thanks a lot for looking into this!

Elv13:
> [..] are you copying code from somewhere else? (Is the license of "somewhere else" compatible with awesomeWM's license?)

No, I looked at `key.c` and it used `XStringToKeysym`. I looked at the manpage and it points to this https://tronche.com/gui/x/xlib/input/keyboard-encoding.html#KeySym which shows exactly what I wanted.

> Why the Xlib dependency?

Because it's well documented and simple.

Anyway, here's a patch to use `xcb_get_modifier_mapping`. Do whatever you want with it. It /seems/ to work, but I never touched that before, so I have no clue. You are the expert on this. I just did this pull request because @blueyed one keeps relying on possibly wrong assumptions and with all the input cleanup work I did "recently" building layers upon layers on top of this, it became required to add this.

```patch
diff --git a/luaa.c b/luaa.c
index 64a217088..fd35cf76b 100644
--- a/luaa.c
+++ b/luaa.c
@@ -469,17 +469,32 @@ static int luaA_get_modifiers(lua_State *L)
     if (!display)
         return 0;
 
-    XModifierKeymap *mods = XGetModifierMapping(display);
+    xcb_get_modifier_mapping_cookie_t c;
+    c = xcb_get_modifier_mapping(globalconf.connection);
+
+    xcb_get_modifier_mapping_reply_t *r;
+    r = xcb_get_modifier_mapping_reply (globalconf.connection, c, NULL);
+
+    xcb_keycode_t *mods = xcb_get_modifier_mapping_keycodes(r);
+
+    int count = xcb_get_modifier_mapping_keycodes_length(r);
 
     lua_newtable(L);
 
+    /* No clue how to handle this if this isn't true */
+    if (count / r->keycodes_per_modifier != Mod5MapIndex + 1) {
+        warn("Failed to parse the X11 modifiers");
+        p_delete(&r);
+        return 1;
+    }
+
     /* This get the MAPPED modifiers, not all of them are */
     for (int i = ShiftMapIndex; i <= Mod5MapIndex; i++) {
         lua_pushstring(L, get_modifier_name(i));
         lua_newtable(L);
 
-        for (int j = 0; j < mods->max_keypermod; j++) {
-            const KeyCode key_code = mods->modifiermap[i * mods->max_keypermod + j];
+        for (int j = 0; j < r->keycodes_per_modifier; j++) {
+            const KeyCode key_code = mods[i * r->keycodes_per_modifier + j];
             const KeySym  key_sym  = XkbKeycodeToKeysym(display, key_code, 0, 0);
             if (key_sym != NoSymbol) {
                 /* The +1 because j starts at zero and Lua at 1 */
@@ -501,9 +516,9 @@ static int luaA_get_modifiers(lua_State *L)
         lua_settable(L, -3);
     }
 
-    XFreeModifiermap(mods);
+    p_delete(&r);
 
-    return 0;
+    return 1;
 }
 
```

> Is it just "works for me"

I put it there with "I will check that later" and it just worked, so I left it as-is.

psychon:
Okay. Feel free to leave this Xlib-ified and I can eventually XCB-ify this when I get the time.

Elv13:
> Okay. Feel free to leave this Xlib-ified and I can eventually XCB-ify this when I get the time.

Leave it open or merge it and fix it later? Because that blocks a @blueyed PR/bug

psychon:
Feel free to merge it. However, someone should actually take a look at this code before then. I just noticed that there is a call to `XOpenDisplay`, but no call to `XCloseDisplay`. Want me to set aside some time to actually look at this carefully?

blueyed:
Feel free to take your time - it is not blocking me.

psychon:
I pushed some commits to port this stuff away from Xlib. Thanks for allowing edits from maintainers. ;-)

luaa.c
```diff
@@ -388,6 +388,125 @@ luaA_fixups(lua_State *L)
...
+
+    }
+
+    return 0; /* \0 */
+}
+
+/**
+ * The table of keybindings modifiers.
+ *
+ * Each modifiers have zero to many entries depending on the keyboard layout.
```
blueyed:
```suggestion
 * Each modifier has zero to many entries depending on the keyboard layout.
```

luaa.c
```diff
@@ -388,6 +388,125 @@ luaA_fixups(lua_State *L)
...
+ * For example, `Shift` usually both have a left and right variants. Each
```
blueyed:
```suggestion
 * For example, `Shift` usually both has a left and right variant. Each
```

luaa.c
```diff
@@ -388,6 +388,125 @@ luaA_fixups(lua_State *L)
...
+            xcb_get_modifier_mapping(globalconf.connection), NULL);
+    if (!mods)
+        return 0;
+
+    xcb_keycode_t *mappings = xcb_get_modifier_mapping_keycodes(mods);
+    struct xkb_keymap *keymap = xkb_state_get_keymap(globalconf.xkb_state);
+
+    lua_newtable(L);
+
+    /* This get the MAPPED modifiers, not all of them are */
```
blueyed:
```suggestion
    /* This gets only the MAPPED modifiers, not all of them */
```

psychon:
Uhm, I just wondered: Do we need a special signal for when this field changes, or should people just use the "normal xkb signals" that `xkb.c` emits for that? I think this signal should be documented (and everyone trying to use this in awesome will hate me for telling them to handle the signal...)

Elv13:
`property::modifiers` would be nice to have. I doubt it will ever be used. Is there really people should move their modifiers multiple time a day? That would be sooo confusing

psychon:
> Is there really people should move their modifiers multiple time a day?

When my awesome starts up, I think my keyboard layout is `de`. Hence, `Caps_Lock` is a thing. Then I do `setxkbmap de -option caps:backspace` and `Caps_Lock` is no longer a modifier. So I think the right signal (whichever that is) should at least be mentioned in the docs. If the right signal is `property::modifiers`, well then that also needs to be implemented.

Elv13:
Good point, then `property::modifiers` should be added. Do you take care of it or do you want me to do it when I get time? Also, feel free to squash all the commits and claim ownership of the commit.

Elv13:
I looked into this and decided to rename it to `_modifiers` and hide it from the public API until I complete an non half-assed solution for the "whole problem". I also added a `_active_modifiers` to better fix #2515.

As you explained in https://github.com/awesomeWM/awesome/pull/2315, more work is needed:

 * ~~Do not hardcode the modifier keysym~~ **DONE**
 * ~~Get the current modifier and pressed keys. This is done using `xkb_state_mod_name_is_active` and its cousins.~~ **DONE**
 * Handle injecting locale aware dead keys like UTF-8 chars. This has both the problem that UTF-8 has multiple representations for many things and that it is locale dependent. xkbcommon "compose" API seems to be for that.
 * Handle compose files properly
 * Have a way to translate chars to keysym or dead key sequences
 * Unify all keyboard related into a single `awful.keyboard` module
 * Have all the relevant signals
 * Really handle keyboards mapping instead of the current semi-never-working-properly widget we have.
 * Make sure when you send key events they get to the right client

As you mentioned there, none of that is really simple and I am back to my usual "I am not in vacation so I have no time for this mode". We need this for pull request for internal use cases. No need to expose an API that will not be future proof and duplicate C signals for "nothing". An higher level modules that consume these low level features and exposes a coherent result is preferable from an end user perspective.

lib/awful/key.lua
```diff
@@ -27,23 +24,54 @@ local key = { mt = {}, hotkeys = {} }
...
 --   Mod2, Mod3, Mod4, Mod5, Shift, Lock and Control.
 -- @tparam string k The key
 function key.execute(mod, k)
+    local modmap = generate_conversion_map()
+    local active = capi.awesome._active_modifiers
+
+    -- Release all modifiers
+    for _, m in ipairs(active) do
+        assert(modmap[m])
+        root.fake_input("key_release", modmap[m])
```
psychon: This so does not work. If I press right shift and you fake a release of left shift, then right shift is still pressed.

psychon: Also, I wonder if this even works... I know that xinput-wise, the xtest extension uses its own keyboard. Thus, there would be one input device where left shift is pressed and another input device where a key that was never pressed is released. If I were to guess, I'd say that this means that the fake release just does not do anything, but that's just a guess.

tests/examples/text/awful/keygrabber/alttab.lua
```diff
@@ -16,7 +16,7 @@ local awful = {keygrabber = require("awful.keygrabber"), --DOC_HIDE
             {{"Mod1", "Shift"}, "Tab", awful.client.focus.history.select_next    },
         },
         -- Note that it is using the key name and not the modifier name.
-        stop_key           = "Alt_L",
+        stop_key           = "Mod1",
```
psychon: From a quick look at `awful.keygrabber`: This does not work. The keygrabber watches incoming key events and checks if any of them matches what is specified here. `Mod1` is not a key, so it never comes in with a release event.

luaa.c
```diff
@@ -508,6 +508,31 @@ static int luaA_get_modifiers(lua_State *L)
...
+ */
+
+static int luaA_get_active_modifiers(lua_State *L)
+{
+    int count = 1;
+    lua_newtable(L);
+
+    for (int i = XCB_MAP_INDEX_SHIFT; i <= XCB_MAP_INDEX_5; i++) {
+        const int active = xkb_state_mod_index_is_active (globalconf.xkb_state, i,
+            XKB_STATE_MODS_DEPRESSED | XKB_STATE_MODS_DEPRESSED
```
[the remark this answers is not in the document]
Elv13: Whoops, copy paste error, I did mean XKB_STATE_MODS_EFFECTIVE

tests/examples/text/awful/keygrabber/alttab.lua
Elv13: It is handled

```lua
    local converted = generate_conversion_map()[key]

    -- Stop the keygrabber with the `stop_key`
    if (key == self.stop_key or (converted and converted == self.stop_key))
      and event == self.stop_event and self.stop_key then
        self:stop(key, modifiers)
        return false
    end
```

lib/awful/key.lua
Elv13: It depends on how the modifiers were pressed. For recursive `.execute` and other few case: "it works more than it did before". It isn't the kind of answer you like, but the code is there and it is technically an improvement. Now it works 95.5% of the time instead of 95% of the time.

Elv13:
Any other comment? Otherwise I will merge this soon-ish.

PR #2522, opened 2018-12-30
fix(menubar): escape XML entities in the item name

glenvt18:
This fixes `<Invalid text>` shown instead of the item name when the item name contains XML special characters like '&' (see `awful.widget.common.list_update`). Note that `compute_text_width` called from `get_current_page` already escapes the text before calculation.


Elv13:
Thanks

PR #2523, opened 2018-12-31
doc: Merge some documentation pages

Elv13:
Open question: Should we ship 2 documentation (ldoc) instance. One for Awesome and one for the widgets? ldoc "supports" cross referencing multiple instances. Right now 50% of the documentation index are widgets. The documentation would look cleaner if both of these thing would be in their own little universe, no?

### DIFF
--- a/lib/beautiful/xresources.lua
+++ b/lib/beautiful/xresources.lua
@@ -3,7 +3,7 @@
 --
 -- @author Yauhen Kirylau &lt;yawghen@gmail.com&gt;
 -- @copyright 2015 Yauhen Kirylau
--- @module beautiful.xresources
+-- @submodule beautiful
 ----------------------------------------------------------------------------
 
 -- Grab environment
--- a/lib/awful/hotkeys_popup/keys/termite.lua
+++ b/lib/awful/hotkeys_popup/keys/termite.lua
@@ -3,7 +3,7 @@
 --
 -- @author ikselven
 -- @copyright 2017 ikselven
--- @module awful.hotkeys_popup.keys.termite
+-- @submodule awful.hotkeys_popup
 ---------------------------------------------------------------------------
 
 local hotkeys_popup = require("awful.hotkeys_popup.widget")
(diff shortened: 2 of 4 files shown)


blueyed:
Multiple ldoc instances might be good indeed - but maybe the toc/index could get cleaned up in some other way also?

lib/beautiful/xresources.lua
```diff
@@ -3,7 +3,7 @@
 --
 -- @author Yauhen Kirylau &lt;yawghen@gmail.com&gt;
 -- @copyright 2015 Yauhen Kirylau
--- @module beautiful.xresources
+-- @submodule beautiful
```
actionless: i've a bit underlooked, i suggest reverting this line, because xresources module have quite specific usage related to one particular service integration, not some common theming helpers, like theme_assets above

blueyed: Makes sense.

@Elv13 ?

Elv13: On one hand @actionless has a point. On the other, if we consolidate more pages, the end result is less or more the same. What about taking a look what prevents the `gtk` module from being consolidated and then how to better document these things. Right now the discoverability factor of the theme API is a rather big issue (plus the "wall of text effect" of out current documentation index).

IMHO, beside the (partially deprecated) DPI stuff, these APIs are only relevant to the xresources (based) theme(s). If there was XRDB write support and you could, for example, change urxvt theme dynamically and control compton using an extended version of this, then my position would be much for favorable to make this a top level again.

Elv13:
> Multiple ldoc instances might be good indeed - but maybe the toc/index could get cleaned up in some other way also?

why not both!

The most common complain of Reddit (unixporn) is how impenetrable out doc is. It has too many modules doing too much unrelated stuff. We don't provide a comprehensive overview beside the widgets. Taking the widgets out and cross-referencing both `ldoc` (`@see` and \`\`) them.

PR #2524, opened 2019-01-01
Move code from rc.lua to awful.menu

Sorky:
Allow for the removal of the following from rc.lua...

```
-- {{{ Helper functions
local function client_menu_toggle_fn()
    local instance = nil

    return function ()
        if instance and instance.wibox.visible then
            instance:hide()
            instance = nil
        else
            instance = awful.menu.clients({ theme = { width = 250 } })
        end
    end
end
-- }}}
```

Then a small change further on in rc.lua...
`
                     awful.button({ }, 3, client_menu_toggle_fn()),
`
becomes...
` 
                    awful.button({ }, 3, function() awful.menu.client_list({ theme = { width = 250 } }) end),
`

Maintains existing user functionality by not removing/modifying awful.menu.clients

If this fits with future plans I can add the changes to rc,lua to the PR

Doco change? : I'd suggest that the existing awful.menu.clients documentation replaced by a note detailing the change to rc.lua and the existing simply moved to the new awful.menu.client_list

### DIFF
--- a/lib/awful/menu.lua
+++ b/lib/awful/menu.lua
@@ -544,7 +544,7 @@ end
 local clients_menu = nil
 function menu.client_list(args, item_args, filter)
     if clients_menu and clients_menu.wibox.visible then
-	clients_menu:hide()
+        clients_menu:hide()
         clients_menu = nil
     else
         clients_menu = menu.clients(args, item_args, filter)
--- a/lib/awful/menu.lua
+++ b/lib/awful/menu.lua
@@ -541,6 +541,16 @@ function menu.clients(args, item_args, filter)
     return m
 end
 
+local clients_menu = nil
+function menu.client_list(args, item_args, filter)
+    if clients_menu and clients_menu.wibox.visible then
+	clients_menu:hide()
+        clients_menu = nil
+    else
+        clients_menu = menu.clients(args, item_args, filter)
+    end
+end
+
 --------------------------------------------------------------------------------
 
 --- Default awful.menu.entry constructor


lib/awful/menu.lua
```diff
@@ -541,6 +541,16 @@ function menu.clients(args, item_args, filter)
     return m
 end
 
+local clients_menu = nil
```
actionless: i think it should be `menu.clients_menu_instance = nil` here instead to expose it

however i am not sure if we need to move it out from rc.lua, let's wait for more opinions

Sorky: Works for me without being exposed and I can't see a need to access it elsewhere but I'm open to anything

My reasoning for moving it out of rc.lua is simply that AFAICT there should be no reason for someone to need to edit it (except the "theme" part which would remain in rc.lua anyway in the call)

Elv13:
I am ok with moving this code out of `rc.lua`. The main "issue" I have with this is my hatred for `awful.menu`, but that should not stand it the way. It will be one more function to deprecate once a `awful.widget.common`+`awful.popup` based menu is introduced. Given the size of my merge queue, it is years away, so for now maintaining `awful.menu` is fine.

Sorky:
Moved to https://github.com/awesomeWM/awesome/pull/2524 so as to complete and squash

blueyed:
> Moved to #2524 so as to complete and squash

@Sorky 
Please do not create new PRs in this case: you can push fixups or force-push to existing ones: it only makes it harder for us to review and follow.

blueyed:
Also.. this one here is #2524?!

Sorky:
Whoops - This is the replacement https://github.com/awesomeWM/awesome/pull/2598

And sorry for that but this was "from unknown repository" so I don't believe I'd have been able to edit & squash myself

I'm almost on top of the git monster ;-)

PR #2529, opened 2019-01-01
fix(awful: hotkeys_popup): don't show group label if group itself is empty

actionless:
also moved it to a separate method since in lua there is no `continue` in iterations

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ addons:
       - xvfb
       - zsh
       - x11-apps
+      - vim-nox
       # Need xorg-macros
       - xutils-dev
       # lgi.
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -310,109 +310,115 @@ function widget.new(args)
         return margin
     end
 
+    function widget_instance:_create_group_columns(column_layouts, group, keys, s, wibox_height)
+        local line_height = beautiful.get_font_height(self.font)
+        local group_label_height = line_height + self.group_margin
+        -- -1 for possible pagination:
+        local max_height_px = wibox_height - group_label_height
+
+        local joined_descriptions = ""
+        for i, key in ipairs(keys) do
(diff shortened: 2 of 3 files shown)


Elv13:
can you squash the commits? Mergify isn't very good at this yet.

Also, I merged all the pull requests to allow a documentation example to be added for this hotkeys_popup in the last few days.

So now technically, it should be possible to add examples for your module. See the tooltips files in `/tests/example/awful/tooltip/` directory. This should prevent this from regressing and better test the corner cases.

actionless:
i think more sophisticated tests will require exposing some stuff in widget to be used only for tests (not good pattern), do you have in mind any particular examples of the additional testcases?

Elv13:
We can start with "does this fail with an error" style corner case like empty groups, groups that don't fit in the screen, many time the same action in a group, too many groups, actions without groups, all API options, and some auto grouping (`{description = "view tag #"..i, group = "tag"})`) fun.

Once we get that, then we can start to thing about more sophisticated tools to add to my test framework. I don't think we have to expose new stuff. They are exposed in the hierarchy, just hard to extract. However an higher test-oriented module to manipulate and extract information from the hierarchy can be added to my framework (and I can do it).

actionless:
it wasn't failing with error before this PR, it was just displaying labels for them, so the test wouldn't cover it

actionless:
2019-10-03--1570061221_1110x99_scrot

Elv13:
@actionless You will have to rebase this one too

actionless:
@Elv13 
> Do you plan to try to add tests

i've added the tests to cover pagination and app-specific hotkeys (using vim)

actionless:
weird, coveralls status not updated here yet: https://coveralls.io/builds/26236279

actionless:
(mb reopening will re-trigger the coveralls right)

actionless:
@psychon i was mentioning that in PR comments, but indeed to make git history more introspectable i gonna amend the commit message

actionless:
@psychon done

PR #2538, opened 2019-01-03
doc: revisit gears.table

blueyed:


lib/gears/table.lua
```diff
@@ -88,9 +88,10 @@ end
 --- Check if a table has an item and return its key.
 -- @class function
 -- @name hasitem
--- @param t The table.
+-- @tparam table t The table.
 -- @param item The item to look for in values of the table.
--- @return The key were the item is found, or nil if not found.
+-- @treturn[1] string|number The key of the item.
```
actionless: according you your other PR numeric indexes shouldn't be used with `@treturn`? https://github.com/awesomeWM/awesome/pull/2537/files

lib/gears/table.lua
```diff
@@ -10,12 +10,12 @@ local rtable = table
 local gmath = require("gears.math")
 local gtable = {}
 
---- Join all tables given as parameters.
--- This will iterate all tables and insert all their keys into a new table.
+--- Join all tables given as arguments.
+-- This will iterate over all tables and insert their entries into a new table.
```
actionless: shouldn't it be some word between `all` and `tables`?

lib/gears/table.lua
```diff
@@ -10,12 +10,12 @@ local rtable = table
...
 -- @class function
 -- @name join
--- @param args A list of tables to join
--- @return A new table containing all keys from the arguments.
+-- @tparam table ... Tables to join.
+-- @treturn table A new table containing all entries from the arguments.
```
actionless: > all entries

the same comment as above and few more occurrences of the same pattern next on

lib/gears/table.lua
blueyed: When used they refer to different return values / possibilities, i.e. here it will either return the key or nil.

lib/gears/table.lua
blueyed: IMHO it is fine, but maybe "all of the tables"?
Maybe the original "iterate all tables" was fine already though?!

lib/gears/table.lua
actionless: i see, thanks for explaining this

lib/gears/table.lua
actionless: judging from the article which i referenced above it could be `... all the tables ...`

blueyed: @Veratil 
Thanks.  This is my impression also.
Are you a native speaker?

PR #2541, opened 2019-01-04
Split naughty along model/view lines and add an extensive test suite

Elv13:
This pull request shifts code around to add a `naughty.notification` class and a `naughty.action` class. These object will be used in the last 2 pull requests of this long running serie to add `awful.rules`/`awful.ewmh.add_activate_filter` style filtering and a new set of widgets. It is also the ground work to support v1.2 of the notification spec (action icons, resident notification and friends).

Note that the line count at the top is not representative of anything since there is almost no new code and just code shifting left and right. 

* 670 lines are the *new* tests, 
* ~150 renaming existing code to avoid having deprecated code, 
* 510 lines of *new* documentation 
* 378 *new* empty lines 

(total 1708, the other 650 is the old code moving).

Finally, after a couple days of work, I added a new test suite to catch regressions. It also catches 4 bugs on our existing implementation. They are fixed in their own commit. As for the reverted commits, that's lazyness. It's all the commits that were made after I did this work 2 years ago. All of them have been backported and I added tests.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -24,9 +24,11 @@ require("awful.hotkeys_popup.keys")
 -- Check if awesome encountered an error during startup and fell back to
 -- another config (This code will only ever execute for the fallback config)
 if awesome.startup_errors then
-    naughty.notify({ preset = naughty.config.presets.critical,
-                     title = "Oops, there were errors during startup!",
-                     text = awesome.startup_errors })
+    naughty.notification {
+        preset = naughty.config.presets.critical,
+        title  = "Oops, there were errors during startup!",
+        text   = awesome.startup_errors
+    }
 end
 
 -- Handle runtime errors after startup
@@ -37,9 +39,12 @@ do
         if in_error then return end
         in_error = true
 
-        naughty.notify({ preset = naughty.config.presets.critical,
-                         title = "Oops, an error happened!",
-                         text = tostring(err) })
+        naughty.notification {
+            preset = naughty.config.presets.critical,
+            title  = "Oops, an error happened!",
+            text   = tostring(err)
+        }
+
         in_error = false
     end)
 end
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -25,9 +25,9 @@ require("awful.hotkeys_popup.keys")
 -- another config (This code will only ever execute for the fallback config)
 if awesome.startup_errors then
     naughty.notification {
-        preset = naughty.config.presets.critical,
-        title  = "Oops, there were errors during startup!",
-        text   = awesome.startup_errors
+        preset  = naughty.config.presets.critical,
+        title   = "Oops, there were errors during startup!",
+        message = awesome.startup_errors
     }
 end
 
@@ -40,9 +40,9 @@ do
         in_error = true
 
         naughty.notification {
-            preset = naughty.config.presets.critical,
-            title  = "Oops, an error happened!",
-            text   = tostring(err)
+            preset  = naughty.config.presets.critical,
+            title   = "Oops, an error happened!",
+            message = tostring(err)
         }
 
         in_error = false
--- a/lib/naughty/constants.lua
+++ b/lib/naughty/constants.lua
@@ -1,0 +1,74 @@
+----------------------------------------------------------------------------
+--- This file hosts the shared constants used by the notification subsystem.
+--
+-- [[documented in core.lua]]
+--
+-- @author koniu &lt;gkusnierz@gmail.com&gt;
+-- @author Emmanuel Lepage Vallee &lt;elv1313@gmail.com&gt;
+-- @copyright 2008 koniu
+-- @copyright 2017 Emmanuel Lepage Vallee
+----------------------------------------------------------------------------
+local beautiful = require("beautiful")
+local dpi = beautiful.xresources.apply_dpi
+
+local ret = {}
+
+ret.config = {
+    padding         = dpi(4),
+    spacing         = dpi(1),
+    icon_dirs       = { "/usr/share/pixmaps/", },
+    icon_formats    = { "png", "gif" },
+    notify_callback = nil,
+}
+
+ret.config.presets = {
+    low = {
+        timeout = 5
+    },
+    normal = {},
+    critical = {
+        bg      = "#ff0000",
+        fg      = "#ffffff",
+        timeout = 0,
+    },
+    ok = {
+        bg      = "#00bb00",
+        fg      = "#ffffff",
+        timeout = 5,
+    },
+    info = {
+        bg      = "#0000ff",
+        fg      = "#ffffff",
+        timeout = 5,
+    },
+    warn = {
+        bg      = "#ffaa00",
+        fg      = "#000000",
+        timeout = 10,
+    },
+}
+
+ret.config.defaults = {
+    timeout      = 5,
+    text         = "",
+    screen       = nil,
+    ontop        = true,
+    margin       = dpi(5),
+    border_width = dpi(1),
+    position     = "top_right"
+}
+
+ret.notification_closed_reason = {
+    silent               = -1,
+    expired              = 1,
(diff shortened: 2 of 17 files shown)


Elv13:
There seems to be a little regression when removing a screen, I will fix that tomorrow

awesomerc.lua
```diff
@@ -37,9 +39,15 @@ do
         if in_error then return end
         in_error = true
 
+        naughty.notification {
+            preset  = naughty.config.presets.critical,
+            title   = "Oops, an error happened!",
+            message = tostring(err)
+        }
+
         naughty.notify({ preset = naughty.config.presets.critical,
```
actionless: two notifications for the same matter?

lib/naughty/constants.lua
```diff
@@ -0,0 +1,75 @@
+----------------------------------------------------------------------------
+--- This file host the shared constants used by the notification subsystem.
```
Veratil: hosts

lib/naughty/core.lua
```diff
@@ -238,153 +146,133 @@ capi.screen.connect_signal("removed", function(scr)
...
-function naughty.toggle()
-    if suspended then
-        naughty.resume()
-    else
-        naughty.suspend()
-    end
+--- Connect a global signal on the notification engine.
+--
+-- Functions connected to this signal source will be executed when any
+-- notification object emit the signal.
```
Veratil: emits

lib/naughty/core.lua
```diff
@@ -420,159 +318,178 @@ function naughty.getById(id)
...
+    gdebug.deprecate("Use notification:reset_timeout(new_timeout)", {deprecated_in=5})
+    notification:reset_timeout(new_timeout)
+end
 
-    local timeout = new_timeout or notification.timeout
-    set_timeout(notification, timeout)
-    notification.timeout = timeout
+--- Replace title and text of an existing notification.
+--
+-- This function is deprecated, use `notification.text = new_text` and
```
Veratil: I see above that `message` is replacing `text`. Is this `text` still used here?

lib/naughty/core.lua
```diff
@@ -420,159 +318,178 @@ function naughty.getById(id)
...
+--
+-- To add an handler, use:
+--
+--    naughty.connect_signal("request::display", function(notification, args)
+--        -- do something
+--    end)
+--
+-- @tparam table notification The `naughty.notification` object.
+-- @tparam table args Any arguments passed to the `naughty.notify` function,
+--  including, but not limited to all `naughty.notification` properties.
```
Veratil: `, but not limited to, all` <-- add comma between to and all

lib/naughty/core.lua
```diff
@@ -420,159 +318,178 @@ function naughty.getById(id)
...
-    -- crop to workarea size if too big
-    local workarea = n.screen.workarea
-    local border_width = s.border_width or 0
-    local padding = naughty.config.padding or 0
-    if width > workarea.width - 2*border_width - 2*padding then
-        width = workarea.width - 2*border_width - 2*padding
-    end
-    if height > workarea.height - 2*border_width - 2*padding then
-        height = workarea.height - 2*border_width - 2*padding
+--- Emitted when a notification need pre-display configuration.
```
Veratil: needs

lib/naughty/core.lua
```diff
@@ -420,159 +318,178 @@ function naughty.getById(id)
...
+--
+-- @tparam table notification The `naughty.notification` object.
+-- @tparam table args Any arguments passed to the `naughty.notify` function,
+--  including, but not limited to all `naughty.notification` properties.
```
Veratil: `to, all` add comma

lib/naughty/constants.lua
Elv13: Done, thanks

lib/naughty/core.lua
Elv13: Done, thanks

lib/naughty/core.lua
Elv13: Done, thanks

Elv13:
Rebased.

Oh well, once everybody is done with the release hangover, this scary stuff needs to be handled early if we want a shorter v4.4 release cycle. It is an high risk, regression prone, pull request.

On the other hand, it has more tests than the current implementation and if I squash enough commits it passes the CI (at the cost of an enormous commit). It solves a long list of long standing complains about our notification support and allows to catch up with dunst on the feature front. More importantly, it allows users to do their own stuff without further bloating naughty default popup.

lib/naughty/core.lua
```diff
@@ -238,153 +146,133 @@ capi.screen.connect_signal("removed", function(scr)
...
-            arrange(scr)
-        end
-
-        if notification.destroy_cb and reason ~= naughty.notificationClosedReason.silent then
-            notification.destroy_cb(reason or naughty.notificationClosedReason.undefined)
-        end
-        return true
-    end
+    gdebug.deprecate("Use notification:destroy(reason, keep_visible)", {deprecated_in=5})
+    return notification:destroy(reason, keep_visible)
```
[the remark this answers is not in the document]
Elv13: Good catch, thanks

awesomerc.lua
Elv13: done

lib/naughty/core.lua
Elv13: done, thanks!

Elv13:
Thanks for all the reviews so far.

Last call to review this before I squash the commits (to pass the CI) and merge  it. In a week for now, unless there's more reviews, I will press the button. I want this merged early because it is almost certain it will regress something. If we are to release v4.4 is the near future, then we need as long as possible to get feedback.

psychon:
>  As for the reverted commits, that's lazyness. It's all the commits that were made after I did this work 2 years ago. All of them have been backported and I added tests.

Sigh... I came to ask for all the weird reverts, but I already found the answer. :-(

lib/naughty/core.lua
```diff
@@ -239,6 +239,51 @@ function naughty.suspend()
...
 end
 
+local conns = {}
+
+--- Connect a global signal on the notification engine.
+--
+-- Functions connected to this signal source will be executed when any
+-- notification object emits the signal.
+--
+-- It is also used for some generic notification signals such as
```
psychon: Don't we have a common implementation of `connect_signal` and friends in `gears.object`? Why do we need another implementation of that instead of just using that one?

Elv13: In this case because it is a module and not a class. However, if you prefer, I can move the class connect signal implementation to `gears`. And yes, the only difference is `:` vs. `.`, but for consistency with the other modules, `.` is required.

Elv13:
Now that CodeCov is alive again, it helped find a new issue. If you used the deprecate "reset_timeout" after called "naughty.suspend()" (also deprecated), it would not work. This is fixed and now I am trying to find a commit log that passes the tests. Once this is done I will merge this.

Elv13:
To anyone reading this, please update to git-master and report any regressions. I put a huge amount of effort testing this, but everybody has their own corner case or apps abusing the spec in different ways.

alfunx:
Unfortunately found one already. In the default `rc.lua` append:

```lua
local _clock_notif
mytextclock:buttons(gears.table.join(
   awful.button({ }, 1, function()
      naughty.destroy(_clock_notif)
      _clock_notif = naughty.notify {
         title = "hello",
         text = "world",
      }
   end)
))
```

Click on the clock, dismiss the notification, click the clock again. Found this usage in `lain`, I guess this goes into

> abusing the spec in different ways

Elv13:
> I guess this goes into abusing the spec in different ways

Yes, destroying an already destroyed notification wasn't a feature, but it didn't create an error either. I changed this to print a warning on stderr.

https://github.com/awesomeWM/awesome/pull/2662

Issue #1275, opened 2016-12-16
apidoc: include search box


blueyed:
When building and pushing the apidocs, we should change the template to include a search box, e.g. via https://duckduckgo.com/search_box:

```
<iframe src="https://duckduckgo.com/search.html?site=https://awesomewm.org/apidoc/&prefill=Search apidoc" style="overflow:hidden;margin:0;padding:0;width:408px;height:40px;" frameborder="0"></iframe>
```

This needs to be done in https://github.com/awesomeWM/awesome/blob/master/build-utils/travis-apidoc.sh, e.g. by either changing the template, or changing the generated output afterwards.

blueyed:
The results from Google are much better though: https://www.google.de/?q=watch+site:https:%2F%2Fawesomewm.org%2Fapidoc%2F

https://duckduckgo.com/?q=watch+site%3Ahttps%3A%2F%2Fawesomewm.org%2Fapidoc%2F

Likely since the `<title>` is the same on all apidoc pages?!  (`<title>awesome API documentation</title>`).

blueyed:
Google however fails to find any results for "dockable": https://www.google.de/?q=dockable+site:https:%2F%2Fawesomewm.org%2Fapidoc%2F.
Two results currently with DDG: https://duckduckgo.com/?q=dockable+site%3Ahttps%3A%2F%2Fawesomewm.org%2Fapidoc%2F&ia=web.

PR #2544, opened 2019-01-04
ldoc: use custom ldoc.ltp for improved HTML title

blueyed:
This should help a lot with making the docs appear better in search results (https://github.com/awesomeWM/awesome/issues/1275).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,13 +286,13 @@ if(GENERATE_DOC)
 
     # Copy the aliases to the build directory
     file(COPY ${SOURCE_DIR}/docs/aliases DESTINATION ${BUILD_DIR}/docs)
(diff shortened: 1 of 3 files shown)


actionless:
could you make a screenshot of how it looks when built?

blueyed:
@actionless 
I've checked that `<title>` tags are changed with this..
A screenshot would not really help, but the idea is that https://duckduckgo.com/?q=watch+site%3Ahttps%3A%2F%2Fawesomewm.org%2Fapidoc%2F&ia=web would look better.

actionless:
ah, it's only about <title>? no visible markup changes at all?

blueyed:
Yes.. the `<title>` however is visible in the browser's titlebar / tab / etc.

This is based on the default template, with only adjustments for title (based on what is used in `<h1>`s already further down).

Elv13:
thanks

Elv13:
Build failed because lua.org is under DDOS (or we had too many awesome contributions today, choose one)

PR #2545, opened 2019-01-04
cmake: fix generate-examples/check-examples

blueyed:
Those do not generate files, but only run the tests in there, and should
not be triggered with `make ldoc`.

Move them to `check-examples` instead.

TODO:

- [ ] squash-merge
- [ ] investigate coverage diff (after https://github.com/awesomeWM/awesome/pull/2574)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -199,10 +199,7 @@ script:
         travis_fold_start "DO_COVERAGE"
 
         # Run tests/examples explicitly.
-        (mkdir -p tests/examples/build \
-         && cd tests/examples/build \
-         && cmake $CMAKE_ARGS .. \
-         && make)
+        make check-examples || exit 1
         do_codecov samples
 
         make check-unit || exit 1
--- a/.travis.yml
+++ b/.travis.yml
@@ -212,6 +212,7 @@ script:
 
         travis_fold_end
       else
+        # TODO: does not run check-examples.  Should it?
         travis_run_in_fold "make.check-unit" make check-unit
         travis_run_in_fold "make.check-integration" make check-integration
         travis_run_in_fold "make.check-themes" make check-themes
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -310,11 +310,12 @@ foreach(file ${test_files})
     endif()
 endforeach()
 
-add_custom_target(generate-examples ALL DEPENDS ${EXAMPLE_DOC_GENERATED_FILES} ${EXAMPLE_DOC_EXTRA_TARGETS})
+add_custom_target(generate-examples ALL DEPENDS ${EXAMPLE_DOC_GENERATED_FILES})
 
 add_custom_target(check-examples
     ${CMAKE_COMMAND} ${CMAKE_SOURCE_DIR}
-    WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+    DEPENDS ${EXAMPLE_DOC_EXTRA_TARGETS})
 list(APPEND CHECK_TARGETS check-examples)
 
 # vim: filetype=cmake:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80:foldmethod=marker
--- a/tests/examples/CMakeLists.txt
+++ b/tests/examples/CMakeLists.txt
@@ -272,7 +272,7 @@ function(run_test test_path namespace escaped_content)
         file(RELATIVE_PATH rel_output "${TOP_SOURCE_DIR}" "${OUTPUT_IMAGE_PATH}")
         add_custom_command(
             COMMAND "${TOP_SOURCE_DIR}/tests/examples/runner.sh" "${expected_output_path}" ${LUA_COV_RUNNER} ${template} ${test_path} ${IMAGE_PATH} ${IGNORE_ERRORS}
-            COMMENT "Running ${rel_test_path} (via ${rel_template}, generating ${rel_output})"
+            # COMMENT "Running ${rel_test_path} (via ${rel_template}, generating ${rel_output})"
             DEPENDS ${template} ${test_path}
             OUTPUT  ${OUTPUT_IMAGE_PATH}
             VERBATIM)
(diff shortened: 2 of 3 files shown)


tests/examples/CMakeLists.txt
```diff
@@ -310,11 +310,12 @@ foreach(file ${test_files})
...
 endforeach()
 
-add_custom_target(generate-examples ALL DEPENDS ${EXAMPLE_DOC_GENERATED_FILES} ${EXAMPLE_DOC_EXTRA_TARGETS})
+add_custom_target(generate-examples ALL DEPENDS ${EXAMPLE_DOC_GENERATED_FILES})
 
 add_custom_target(check-examples
     ${CMAKE_COMMAND} ${CMAKE_SOURCE_DIR}
-    WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+    DEPENDS ${EXAMPLE_DOC_EXTRA_TARGETS})
```
blueyed: I feel like this should also use `add_custom_command` above (line 272, https://github.com/awesomeWM/awesome/pull/2545/files#diff-99add2d86fe5b8809426e4385f9e6403R272), and generate a "stamp file" / target for CMake to only run this when the input file changes.

actionless:
what about renaming `EXAMPLE_DOC_EXTRA_TARGETS` to `EXAMPLE_DOC_TEST_TARGETS`?

blueyed:
Will do this after https://github.com/awesomeWM/awesome/pull/2548.

blueyed:
> what about renaming `EXAMPLE_DOC_EXTRA_TARGETS` to `EXAMPLE_DOC_TEST_TARGETS`?

Both/all of them also run tests, it is just that some produce output (images), while others do not.
With produced output CMake will not re-run them, but the others are added as explicit targets that get used with `make ldoc`.
I will rebase / fix this further after https://github.com/awesomeWM/awesome/pull/2548.

blueyed:
Reworked it:
- `make check-examples` runs the tests now always.
- `make` will generate output files only once, and does not run the examples that do not generate any output.  The output from `check-examples` is used/valid for it, i.e. after `make check-examples` output files are not regenerated.

tests/examples/CMakeLists.txt
```diff
@@ -256,26 +258,26 @@ function(run_test test_path namespace escaped_content)
...
     file(RELATIVE_PATH rel_test_path "${TOP_SOURCE_DIR}" "${test_path}")
+
+    # Add target to run this as test (via check-examples, not ignoring errors).
+    set(TARGET_NAME "run-${rel_test_path}")
+    STRING(REPLACE "/" "-" TARGET_NAME ${TARGET_NAME})
+    set(EXAMPLE_DOC_TEST_TARGETS ${EXAMPLE_DOC_TEST_TARGETS} ${TARGET_NAME}
+        PARENT_SCOPE)
+    add_custom_target(${TARGET_NAME}
+        COMMAND "${TOP_SOURCE_DIR}/tests/examples/runner.sh" "${expected_output_path}" ${LUA_COV_RUNNER} ${template} ${test_path} ${IMAGE_PATH}
+        VERBATIM)
```
blueyed: This generates a lot of internal targets (bloating the generated Makefile).
Is there a better way?
Maybe this should get moved into a wrapper script so that only a single command/target is needed?
On the other hand it can be useful to run a specific test only (e.g. for coverage debugging), but that could be left commented out then also.

psychon: > This generates a lot of internal targets (bloating the generated Makefile).

Why is that bad?

The current version allows to run these targets in parallel.

blueyed: It is not really bad, but each target adds like 10-20 lines there etc, and I got the impression that it would also slow down `make clean`, but have not investigated.
I think it is much better than the current version anyway.

.travis.yml
```diff
@@ -203,6 +200,7 @@ script:
 
         travis_fold_end
       else
+        # TODO: does not run check-examples.  Should it?
         travis_run_in_fold "make.check-unit" make check-unit
```
blueyed: Do we want to run `check-examples` also with non-coverage builds?

blueyed:
@Elv13 on IRC:
> the example do modify the doc, they need to be executed

You mean the ones not generating images still need to run when generating docs?  Makes sense.
Then we need some stamp file to not run this always.  Couldn't/shouldn't that be the changed doc file then?

Elv13:
@blueyed I think we need to add `check-examples` to the `DEPENDS` of `ldoc` if we are to make this change? I don't have time to test right now

Elv13:
There is 3 steps, I think.

 * Check what the documentation example provides (text, image, etc)
 * Execute those files and generate the artifacts
 * Integrate/inject those artifacts in the `ldoc` input

The images are simple because they are deterministic, but the generated text may change when `step2` is performed, so it must be executed before CMake `configure_file` is executed. 

This could be mitigated by moving `ldoc` and the `configure_file` to a custom_target instead of directly in the configuration step. That way most targets would use the raw `.lua` files from awful/wibox/gears instead of the post-processed output of `configure_file`. By doing that the weird circular target dependency would be gone.

blueyed:
`make ldoc` on master runs e.g. `tests/examples/text/awful/keygrabber/root_keybindings.lua` always.
But changing the file does not change any documentation.

So while it gets run always, it does not update `lib/awful/keygrabber.lua` for
ldoc (which replaces the
`@DOC_text_awful_keygrabber_root_keybindings_EXAMPLE@`).

Therefore I think this PR does not cause any regression in this regard.

blueyed:
Would like to get this in for https://github.com/awesomeWM/awesome/pull/2672, to ease handling the examples tests therein / from there.

Elv13:
> make ldoc on master runs e.g. tests/examples/text/awful/keygrabber/root_keybindings.lua always.
But changing the file does not change any documentation.

That's a bug, the `keygrabbing.lua` target should `DEPENDS` on `tests/example/text/awful/keygrabber/*` and `tests/example/awful/keygrabber/*` (if, of course, they exist). That in turn should re-run ldoc. But right, this is not a bug in *this* pull request.

PR #2548, opened 2019-01-04
cmake: remove check-examples

blueyed:
This was meant to explicitly test/check the examples, when caching for
generated output was added (02f894ce0), but is no longer useful/required
anymore, especially since it only works when run from tests/examples
anyway (what Travis is doing explicitly).


blueyed:
Hmm, I think it should be kept, but get fixed instead (basicallay making it use what Travis does, enforcing STRICT_TESTS).
And then also used by Travis.

PR #2550, opened 2019-01-05
travis: Pull the release from github.com using https://

Elv13:
(only merge is the lua.org outage end up lasting for long enough to become a problem, it's been 12h already)

www.lua.org is currently down and it may last for a while. They say on
IRC we can trust the github.com mirror as it is mostly official (but
maintained by the community and often off by a few days from the official
release)

The official mirrors are:

http://www.lua.org/ftp
http://webserver2.tecgraf.puc-rio.br/lua/mirror/ftp/

None have working https.


Elv13:
lua.org went back a few minutes ago

PR #2552, opened 2019-01-05
doc: Fix a copy/paste mistake in the screen module

Elv13:


Elv13:
Eventually we will need to do something about the unstable C coverage. What kind of events it gets is pseudo random.

Issue #2556, opened 2019-01-06
wibox.layout.manual:insert() exists in the documentation but not in the code


Aire-One:
Output of `awesome --version`: latest git master version
```
awesome v4.2-574-ged7c243c (Human after all)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

So I tried to use the method `insert` from a `wibox.layout.manual` to insert a widget at the 1st position.

I find it on the documentation I generated from sources and it can also be found at https://awesomewm.org/apidoc/classes/wibox.layout.manual.html#wibox.layout.manual:insert

However, using this method trigger a `attempt to call a nil value` exception...

Looking at the code, the comment-documentation is here, but there is no definition for this method.

The comment in question:
https://github.com/awesomeWM/awesome/blob/ed7c243c03cba3d5c1fe1aa613af43c40da900d7/lib/wibox/layout/manual.lua#L30-L35

Aire-One:
I close this issue since discussion goes on at #2557.

Issue #2181, opened 2018-02-07
Make the wibox get_children_by_id reliable


Elv13:
Output of `awesome --version`:

4.0 to git-master.

Right now it doesn't cross the `wibox.widget {}` sections that can be pasted from the documentation inside of a wibox `:setup{}` section. So copy-pasted documentation examples end up confusingly "working" visually, but are not reachable. This was reported on reddit and is indeed a prime example of confusing anti-features.

Opening this issue so it is not forgotten.

psychon:
So... basically, `get_children_by_id` and the concept of an id have to be added to `wibox.widget.base` and the implementation would then use `for _, w in self:get_children() do if w.id == id then table.insert(result, w) end end`?

actionless:
does that also implies what we need to add some recursive check to ensure what id is unique?

or if id is the same list of widgets should be returned

Elv13:
`id` are not unique. The API is like JavaScript. It returns a table with all the result for that `id`.

The problem right now is that if you create widgets with `wibox.widget {...}` you end up with a `get_children_by_id` method there and they are not in sync with the `wibar`/`wibox` top level `get_children_by_id`.

As @psychon sugested solution demonstrate, the easy way to do this is `O(N)` during lookup instead of `O(1)` when cached. That sucks. A less terrible way is to do it in `wibox.widget.base` by checking if there is an existing `id` table when building the tree. But that has the drawback of keeping track when widgets are removed or added to all subtrees. In that case the widgets created using `wibox.widget{...}` can be shared across multiple wiboxes so we can't just merge all subtrees.

So both solutions have big drawbacks and it's why it wasn't implemented yet. In theory, the lookup happens more often than building new trees, so it should be prioritized in term of performance (slower tree building for faster lookups instead of the other way around). This is why I think the solution should involve being able to track changes to the subtrees. That also make the solution more complicated.

actionless:
and does having ids worth it?

i think using variable or table to store references when it's needed is quite viable solution. especially if infrastructure for ids could not only encomplicate the code but also cause some performance affect

Elv13:
Honestly, I don't know. I made it this way so people coming from JavaScript don't have to learn anything new and make the wibox "DOM-like". This issue was reported twice and I never seen the alternate `wibox.parent.parent.widget` style syntax used anywhere. I guess few people who care about accessing widgets *do* prefer the `get_children_by_id()` API.

[...]

PR #2557, opened 2019-01-06
`wibox.layout.manual:insert()` implementation

Aire-One:
Following my issue #2556, here is my implementation try :)

Some notes:

* From my understanding, `manual_layout._private.widgets` and `manual_layout._private.pos` are arrays so elements can be "magically" shifted/organized ;
* I used `table.insert` because it shifts up the elements, so it will keep the widgets ordered (https://www.lua.org/manual/5.3/manual.html#pdf-table.insert) ;
* I wrote the code under the documentation-comment because I think it's easier to read the code / review it. All functions from this file are not written with this rule... Tell me if I need to change that ; 
* I don't know how to do that : New elements will not be accessible as children by their id property ;

### DIFF
--- a/lib/wibox/layout/manual.lua
+++ b/lib/wibox/layout/manual.lua
@@ -33,6 +33,16 @@ local manual_layout = {}
 -- @treturn boolean If the operation is successful
 -- @name insert
 -- @class function
+function manual_layout:insert(index, widget)
+    table.insert(self._private.widgets, index, widget)
+
+    -- Add the point
+    if widget.point then
+        table.insert(self._private.pos, index, widget.point)
+    end
+
+    self:emit_signal("widget::layout_changed")
+end
 
 --- Remove one or more widgets from the layout
 -- The last parameter can be a boolean, forcing a recursive seach of the


actionless:
@Elv13 i see the implementation of those 2 methods is slightly different

so i have to clarify, what is the difference in expected result of both?

Elv13:
A right nevermind, that would be `insert = function(_, ...) return add(...) end`. The current version is better. The logic was that `manual` order is irrelevant.

actionless:
sorry, i still didn't got it clearly

Elv13:
> sorry, i still didn't got it clearly [...] what is the difference in expected result of both?

In the end there is little difference. `insert` was accidentally documented and is implemented by the other layouts so it's a pseudo expected method to exist. The order exists in the API because the `.children` property is part of the base widget API and is ordered. Therefor `:insert()` technically make sense in all layouts, regardless of how they get displayed. However since the manual layout widget position is provided by a point or a callback, then it has no *visual* impact.

actionless:
i still can't understand if it makes sense to just use the same implementation instead of copying the code or those differences are intended?

Elv13:
@Aire-One version is better as it produces the correct `.children`. My hack would have looked the same, but it would not have produced a correct `.children`.

Aire-One:
Sorry, I'm a little late in the discussion, I was studying for an exam...

I'm glad you approved my code.
Thanks for yours time and reviews :)

However I have to disagree with these statements:
> The logic was that manual order is irrelevant.

> However since the manual layout widget position is provided by a point or a callback, then it has no visual impact.

While the order does *theoretically* not have an impact, I found out it has. It changes the drawing order of the widgets. When some children of a manual layout overlaps each other, the order will change the one that is drawn on top of the others.

Here is a short test code I wrote to run within `xephyr` to test widget overlapping:
```lua
require('rc')

local beautiful = require('beautiful')
local gears = require('gears')
local wibox = require('wibox')

local w_background = wibox.widget {
    resize = true,
    image = beautiful.awesome_icon,
    forced_height = 16,
    forced_width = 16,
    point = awful.placement.centered
}

local w_icon = wibox.widget {
    layout = wibox.container.background,
    wibox.widget.base.empty_widget(),
    bg = '#eeeeee',na
    shape = gears.shape.rounded_bar,
    forced_height = 16,
    point = awful.placement.maximize_horizontally + awful.placement.bottom
}

local w1 = wibox.widget {
    layout  = wibox.layout.manual,
    w_background,
    w_icon
}

local w2 = wibox.widget {
    layout  = wibox.layout.manual,
    w_icon,
    w_background
}

awful.screen.w = wibox {
    type = 'dock', ontop = true,
    visible = true, opacity = 1,
    width = 200, height = 30
}
awful.screen.w:setup {
    layout = wibox.layout.flex.horizontal,
    w1,
    w2
}
```
And here is a preview of the result:
Preview

So, order do have an impact on the visual aspect of the layout. That's why the implementation of `insert` can't be the same than `add`.

---

> @Aire-One version is better as it produces the correct .children. My hack would have looked the same, but it would not have produced a correct .children.

But it still misses something because you can't access inserted children by their ids.

Here is an example of what I mean:
```lua
local my_manual_layout = wibox.widget {
    layout  = wibox.layout.manual,
    { id = 'w_role', ... }
}
my_manual_layout:insert(1, wibox.widget { id = 'insert_role', ...})
my_manual_layout.w_role -- this works
my_manual_layout.new_role -- this fails
```

I suspect it have something to do with #2181 but I'm not sure on how it actually works...

Elv13:
> I suspect it have something to do with #2181 

It has

> but I'm not sure on how it actually works...

It does not ;).

The original "scope" of `:get_children_by_id()` was limited to DOM-like queries on the widget trees created using `:setup()`. At that point in time, the imperative and declarative styles were still very different and not quite compatible with each other. Since then many pull requests have better integrated them using concepts such as delegate template. #2508 goes further by allowing both widget declaration and definition in every containers and layouts. Some ongoing pull requests move all widgets constructor to use named parameters. The finality is to re-unify both style. For this, `:get_children_by_id()` has to be moved to the hierarchy instead of being implemented in the declarative tree parser. Once all 3 "projects" are finished, then there will be one widget system with a very flexible and permissive syntax. But right now, [4.5 years](https://github.com/elv13/retrograde) since I started upstreaming my config widget system, this is still work in progress, but large progress has been made.

Elv13:
Thanks

(the force push was to set a commit message instead of "try 1")

PR #2558, opened 2019-01-07
Factor out awful.rules.get_props_and_callbacks

blueyed:
This is useful for third-party usage to use the props/callbacks
mechanism provided by awful.rules in a different way.

### DIFF
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -326,13 +326,37 @@ function rules.remove_rule_source(name)
     return false
 end
 
--- Add the rules properties
-local function apply_awful_rules(c, props, callbacks)
-    for _, entry in ipairs(rules.matching_rules(c, rules.rules)) do
-        gtable.crush(props,entry.properties or {})
+--- Get (merged) properties and (list of) callbacks of rules matching a client.
+-- @client c The client.
+-- @tparam table _rules The rules to check. List with "rule", "rule_any",
+--   "except" and "except_any" keys.
+-- @treturn[0] table Merged properties from "properties" table of matching rules.
+-- @treturn[0] table List of callbacks from "callback" property of matching rules.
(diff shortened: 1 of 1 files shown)


Elv13:
Sorry for nitpicking, but *why* exactly is this needed? There is already a bunch of ways to integrate different rule system in the current pipeline. What are they missing (serious question given I personally added them and my use case may not fit yours)? The change you propose is incompatible with the new pipeline and when/if we remove the rule priority groups to use the newer graph dependency system as proposed the other day in some other discussions. I would prefer to avoid exposing such implementation details.

If there is a valid use case, please explain it to see if we can come up with a more future proof design.

blueyed:
@Elv13 
I had this in an older stash (before the rewrite), where it made more sense.
The use case is to being able to use awful-style rules in other libraries (https://github.com/awesomeWM/awesome-www/pull/111/files#diff-8e28efad3f40a91f34c842c2155a72bdR38).

I can see that it does not really fit anymore into awful.rules itself given the reasons you've mentioned, but wondered if it could be made fitting?

I wouldn't want to use `add_rule_source` etc with a simple external lib though.

Elv13:
> The use case is to being able to use awful-style rules in other libraries

It's a valid use case. My notification filter PR (unsubmitted, comes after the "split notification model/view PR") indeed does something like this instead of the hardcoded presets.

What about moving that code to `gears.rules` and keep `awful.rules` (or `awful.client.rules`? given we will have rules elsewhere too?) for the client related code? It would be cleaner than exposing some internals that *will* break for Awesome 5.0 due to the dependency graph integration.

Can you modify the pull request or do you prefer to go review #2541 and wait for me to do it in its follow-up?

Elv13:
Closing because #2600 hopefully allows to reuse the matching code without making the internal client properties handling part of the public API.

PR #2565, opened 2019-01-09
Travis: use Ubuntu Xenial image

blueyed:
This has lua5.3 already.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
-sudo: required
-dist: trusty
+dist: xenial
 language: c
 
 # Build only master and stable branches.  Other branches go through PRs.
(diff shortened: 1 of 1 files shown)


Elv13:
@blueyed Maybe we could keep the compile code and use it for `lua-5.4.0-work2.tar.gz`? I don't think Awesome supports it yet, but at least we will know. We could also revert the change later for it.

On the plus side with Xenials, we will also be able to enable the official Clang PPA and have the memory leak and address sanitizer enabled. As far as I know, the leak sanitizer currently fails in the multi-screen tests so it needs to be fixed first. That can wait and be done in other PRs.

psychon:
> Unfortunately lgi fails to build: https://travis-ci.org/awesomeWM/awesome/builds/477730589#L1487-L1512

That one should be easy to fix. IIRC this is documented in the Lua 5.4 docs under breaking changes. Something like 0 or `lua_gettop` should be used for this new argument to get the old behaviour.

PR #2567, opened 2019-01-09
Makefile: run CMake with CMAKE_ARGS defined

blueyed:
This ensures to re-run CMake when using something like:

> make check-unit CMAKE_ARGS="-DSTRICT_TESTS=true -DDO_COVERAGE=1"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,8 @@ cmake-build: $(BUILDDIR)/Makefile
 	$(ECHO) "Building…"
 	$(MAKE) -C $(BUILDDIR)
 
-# Run CMake always with CMAKE_ARGS defined.
-ifdef CMAKE_ARGS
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,6 +12,11 @@ all: $(TARGETS) ;
 
 $(TARGETS): cmake-build
 
+# Run CMake always with CMAKE_ARGS defined.
+ifdef CMAKE_ARGS
+.PHONY: $(BUILDDIR)/Makefile
+endif
+
```
blueyed: Stupid/inefficient with exported CMAKE_ARGS like on Travis.

But could be made to work only for `make … CMAKE_ARGS=…` (by looking at the origin of the var).

blueyed: Done.

Issue #2576, opened 2019-01-11
Travis: BUILD_APIDOC is broken on master


blueyed:
https://travis-ci.org/awesomeWM/awesome/jobs/478411532#L3952

```
$ if [ "$BUILD_APIDOC" = "true" ]; then build-utils/travis-apidoc.sh; fi
Post-processing (API) documentation.
TRAVIS_PULL_REQUEST: false
TRAVIS_BRANCH: master
Cloning into 'build/apidoc'...
Switched to a new branch 'merged-update'
patching file classes/awful.wibar.html
patching file classes/client.html
patching file classes/screen.html
patching file classes/tag.html
patching file classes/wibox.html
patching file documentation/89-NEWS.md.html
patching file images/AUTOGEN_awful_tooltip_textclock2.svg
patching file images/AUTOGEN_awful_tooltip_textclock.svg
[merged-update e6ffff5] [relevant] Update docs for master-gd2fd15c via Travis
 8 files changed, 95 insertions(+), 95 deletions(-)
[merged-update ee043fc] [boilerplate] Update docs for master-gd2fd15c via Travis
 2 files changed, 2 insertions(+), 2 deletions(-)
error: unable to unlink old 'classes/awful.wibar.html': Permission denied
error: unable to unlink old 'classes/client.html': Permission denied
error: unable to unlink old 'classes/screen.html': Permission denied
error: unable to unlink old 'classes/tag.html': Permission denied
error: unable to unlink old 'classes/wibox.container.place.html': Permission denied
error: unable to unlink old 'classes/wibox.html': Permission denied
error: unable to unlink old 'documentation/89-NEWS.md.html': Permission denied
fatal: Could not reset index file to revision 'HEAD~2'.
```

Apparently broken by 0b36da273, but not clear to me why yet.

PR #2579, opened 2019-01-11
Makefile: keep cmake-build as default target

blueyed:
Fixes https://github.com/awesomeWM/awesome/issues/2576.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,12 @@ endif
 
 BUILDDIR=build
 
-all: awesome ;
+# Run "make" in $(BUILDDIR) by default.
+# This is required to generate all files already, which should not be generated
+# with "(sudo) make install" only later.
+cmake-build: $(BUILDDIR)/Makefile
+	$(ECHO) "Building…"
+	$(MAKE) -C $(BUILDDIR)
 
 $(BUILDDIR)/Makefile:
 	$(ECHO) "Creating build directory and running cmake in it. You can also run CMake directly, if you want."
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -7,7 +7,12 @@ endif
 
 BUILDDIR=build
 
-all: awesome ;
+# Run "make" in $(BUILDDIR) by default.
+# This is required to generate all files already, which should not be generated
+# with "(sudo) make install" only later.
+cmake-build: $(BUILDDIR)/Makefile
+	$(ECHO) "Building…"
+	$(MAKE) -C $(BUILDDIR)
```
actionless: isn't it running `$(MAKE) -C $(BUILDDIR)` twice?

blueyed: In which case?

Just `make` defaults to `make cmake-build` now, which runs `make -C build`.
`make foo` will run `make -C build foo` (due to the implicit rule at the end).

The problem before 0b36da273 was that the default was `all`, which depended on `awesome`, which depended on `cmake-build`.  So cmake-build was run (`make -C build`), then `make -C build awesome` again.

blueyed:
I am merging this already, since it is meant to fix master.

Makefile
actionless: `cmake-build` build target depends on `$(BUILDDIR)/Makefile`

and both of them are running `$(MAKE) -C $(BUILDDIR)` one of them just bypassing args at the end (`$@`)

blueyed: `make` (or `make cmake-build`) does not also build the `%` target.

blueyed: `%` is only used if not other target is specified (including the default).

PR #2583, opened 2019-01-12
Fix PR 2527

Sorky:
Fix for hotkey_popup via awful menu showing on wrong screen

#2527

Next comments will show local testing and confirmation

Issue #2527, opened 2019-01-01
Question: Discussion: Trying to understand hotkeys_popup.show_help


Sorky:
Is it ok to ask someone to explain this?
```
myawesomemenu = { { "hotkeys", function() return false, hotkeys_popup.show_help end},

globalkeys = gears.table.join( awful.key({ modkey,           }, "s", hotkeys_popup.show_help,
              {description="show help", group="awesome"}),
```

Why the "return false," in the menu version? [My testing seems to show tables getting passed by the menu - is it to eliminate that? Then again it seems it may be to hide some other bug that the latest widget.lua has fixed ]

Would it be a reasonable PR to add a menu callable version in hotkeys_popup/init.lua?
```
local awful = { screen = require( "awful.screen" ) }
hotkeys_popup.show = function() hotkeys_popup.widget.show_help( _, awful.screen.focused() ) end
```

awful.hotkeys_popup.show_help doesn't seem to be documented [but awful.hotkeys_popup.widget.show_help is]- Should it be? Why does it exist - simply to shorten the call?

psychon:
> Why the "return false," in the menu version?

`awful.menu` assigns meaning to the return value of the function:
https://github.com/awesomeWM/awesome/blob/5b45442c7341a572bd224bd87ab7af14bf3c6287/lib/awful/menu.lua#L307-L318

The first return value is "should the menu stay visible?" and the section return value is the actual action to execute. Don't ask me why there is support for this double callback, but it is definitely there.

Now, if the callback function where just `hotkeys_popup.show_help`, the return value of that function would be used. If I followed the code correctly, this return value is a completely useless internal function. This useless return value would be interpreted as `true` and so the menu would stay visible. By having the extra `return false` indirection, the menu is closed in just the way one would expect.

> Would it be a reasonable PR to add a menu callable version in hotkeys_popup/init.lua?

My gut feeling would be "no".

> awful.hotkeys_popup.show_help doesn't seem to be documented [but awful.hotkeys_popup.widget.show_help is]- Should it be? Why does it exist - simply to shorten the call?

Well, `lib/awful/hotkeys_popup/init.lua` is quite a simple file:

actionless:
> awful.hotkeys_popup.show_help doesn't seem to be documented [but awful.hotkeys_popup.widget.show_help is]- Should it be? 

they both are documented (they are different functions, not just a an alias or shortcut):

[...]

PR #2585, opened 2019-01-13
Update awesomerc.lua

Sorky:
Alternative fix for hotkey_popup via awful menu showing on wrong screen

https://github.com/awesomeWM/awesome/pull/2583
#2527

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -14,7 +14,7 @@ local beautiful = require("beautiful")
 -- Notification library
 local naughty = require("naughty")
 local menubar = require("menubar")
-local hotkeys_popup = require("awful.hotkeys_popup").widget
+local hotkeys_popup = require("awful.hotkeys_popup")
 -- Enable hotkeys help widget for VIM and other apps
 -- when client with a matching name is opened:
 require("awful.hotkeys_popup.keys")
@@ -98,13 +98,15 @@ local function client_menu_toggle_fn()
         end
     end
 end
+
+local function show_hotkeys_popup = function() hotkeys_popup.widget.show_help( nil, awful.screen.focused() ) end
 -- }}}
 
 -- {{{ Menu
 -- @DOC_MENU@
 -- Create a launcher widget and a main menu
 myawesomemenu = {
-   { "hotkeys", function() return false, hotkeys_popup.show_help end},
+   { "hotkeys", show_hotkeys_popup },
    { "manual", terminal .. " -e man awesome" },
    { "edit config", editor_cmd .. " " .. awesome.conffile },
    { "restart", awesome.restart },


Elv13:
```
/home/travis/build/awesomeWM/awesome/awesomerc.lua:102: '(' expected near '='
/usr/local/etc/xdg/awesome/rc.lua:97: '(' expected near '='
```

Also, I prefer an inline function over 3 lines rather than a `local` used only once. This is what is used by most of `rc.lua` code (with some exceptions). The code should always attempt to be close to 80 column and always less than 120. The "Hard" 80 column limit isn't enforced, but usually favored.

Sorky:
BTW - Your comments `I prefer` ... `with some exceptions` ...  `isn't enforced, but usually favored`
are appreciated and I understand the old adage "The exception that proves the rule"

As I was told to learn from existing code here are some examples of 'exceptions'...

--- View next tag. This is the same as tag.viewidx(1).
-- @function awful.tag.viewnext
-- @param screen The screen.
function tag.viewnext(screen)
    return tag.viewidx(1, screen)
end

--- View previous tag. This is the same a tag.viewidx(-1).
-- @function awful.tag.viewprev
-- @param screen The screen.
function tag.viewprev(screen)
    return tag.viewidx(-1, screen)
end

function client.object.get_marked(self)
    return client.property.get(self, "marked")
end

function client.floating.delete(c)
    client.object.set_floating(c, nil)
end

Elv13:
Those are APIs going all the way back to Awesome 3.0, before the object oriented API was implemented. I was referring to `rc.lua` not using `local function` too much. And also many of these `awful.tag` and `awful.client` have been deprecated in 4.0 in favor of inline functions.

Sorky:
I figured as much and totally understand

Although they don't seem to be deprecated, unless you mean that there usage has generally been removed and they will be fully deprecated (as in `gdebug.deprecate`) in 5.0?

Elv13:
A lot of them are

https://awesomewm.org/apidoc/classes/tag.html#Deprecated_functions
https://awesomewm.org/apidoc/classes/client.html#Deprecated_functions

But not all of them are (yet). https://github.com/awesomeWM/awesome/pull/2508 is the next round. Eventually the idea is to have strict guidelines and stick to them, but until v4.0, the API was a "design by committee who forgot to consult each other" style. It was all over the place from OOP to functional to procedural. All with camel case, snake case and no case names. https://github.com/awesomeWM/awesome/pull/757 and https://github.com/awesomeWM/awesome/pull/790 was once of the first round to move everything to properties. There was a dozen more iterations after that. There was also the whole `request::` APIs introduced to remove all the hardcoded signal callbacks and move to a real pluggable "pipeline" for everything. This is also ongoing ( next round is https://github.com/awesomeWM/awesome/pull/2507 ). Along with the series that moves all constructor to use named property (you complained about that elsewhere when I blocked one of your PR because it conflicted it that series), it will bring some unity to the API.

That's a very long term work and has been going on for many years. Yes, eventually the functions you mentioned above will go, but not now, not yet. Until a better API exist, I will need them for APIs like the keygrabber (see first example https://awesomewm.org/apidoc/classes/awful.keygrabber.html). My long term plan for that is to add a `gears.history` and move all the next/previous magic there and unify the tag/client/command/screen history.

Sorky:
I figured as much and totally understand

Although they don't seem to be deprecated, unless you mean that there usage has generally been removed and they will be fully deprecated (as in `gdebug.deprecate`) in 5.0?

PR #2593, opened 2019-01-15
Update luaa.c: Add awesome.config_path (and config.file)

Sorky:
awesomeWM#2525

awesome.icon_path exists
awesome.themes_path exists

why not awesome.config_path?

conf_path would just be the path as opposed to the existing conffile while is the path .. "/" .. filename"

Note: config_path is essentially a constant, so having it readily available seems logical

Also remove the need for gears.filesystem.get_configuration_dir() [step 1 of deprecation would be for it to simply return awesome.config_path .. "/"

### DIFF
--- a/luaa.c
+++ b/luaa.c
@@ -172,6 +172,8 @@ extern const struct luaL_Reg awesome_mouse_meta[];
 
 /** Path to config file */
 static char *conffile;
+static char *config_file;
+static char *config_path;
 
 /** Check whether a composite manager is running.
  * \return True if such a manager is running.
@@ -405,6 +407,16 @@ luaA_fixups(lua_State *L)
  */
 
 /**
+ * The configuration filename which has been loaded.
+ * @tfield string config_file
+ */
+
+/**
+ * The location of the configuration file which has been loaded.
+ * @tfield string config_path
+ */
+
+/**
  * True if we are still in startup, false otherwise.
  * @tfield boolean startup
  */
@@ -451,6 +463,18 @@ luaA_awesome_index(lua_State *L)
     if(A_STREQ(buf, "conffile"))
     {
         lua_pushstring(L, conffile);
+        return 1;
+    }
+
+    if(A_STREQ(buf, "config_file"))
+    {
+        lua_pushstring(L, config_file);
+        return 1;
+    }
+
+    if(A_STREQ(buf, "config_path"))
+    {
+        lua_pushstring(L, config_path);
         return 1;
     }
 
@@ -920,6 +944,9 @@ luaA_loadrc(const char *confpath)
     /* Set the conffile right now so it can be used inside the
      * configuration file. */
     conffile = a_strdup(confpath);
+    char * eop = strrchr(confpath,'/');
+    config_file = eop ? a_strdup(eop + 1) : a_strdup(confpath);
+    config_path = eop ? a_strndup(confpath, a_strlen(confpath) - a_strlen(config_file) - 1) : p_new(char, 1);
     /* Move error handling function before function */
     lua_pushcfunction(L, luaA_dofunction_on_error);
     lua_insert(L, -2);
@@ -935,6 +962,8 @@ luaA_loadrc(const char *confpath)
     fprintf(stderr, "%s\n", err);
     /* An error happened, so reset this. */
     conffile = NULL;
+    config_file = NULL;
+    config_path = NULL;
     /* Pop luaA_dofunction_on_error() and the error message */
     lua_pop(L, 2);
 


psychon:
I still fail to see the point in this, but apparently that doesn't count for much...
Okay, so here are some benchmarks to show how much faster things become (no idea how accurate these are, the setup here is not really meant for microbenchmarks, I guess):
```diff
diff --git a/tests/test-benchmark.lua b/tests/test-benchmark.lua
index def3c01f3..7ed1d8b58 100644
--- a/tests/test-benchmark.lua
+++ b/tests/test-benchmark.lua
@@ -5,6 +5,9 @@ local runner = require("_runner")
 local awful = require("awful")
 local GLib = require("lgi").GLib
 local create_wibox = require("_wibox_helper").create_wibox
+local gears = require("gears")
+
+print(_VERSION, jit and jit.version)
 
 local BENCHMARK_EXACT = os.getenv("BENCHMARK_EXACT")
 if not BENCHMARK_EXACT then
@@ -71,11 +74,21 @@ local function e2e_tag_switch()
     do_pending_repaint()
 end
 
+local function get_configuration_dir()
+    return gears.filesystem.get_configuration_dir()
+end
+
+local function config_path()
+    return awesome.config_path
+end
+
 benchmark(create_and_draw_wibox, "create&draw wibox")
 benchmark(update_textclock, "update textclock")
 benchmark(relayout_textclock, "relayout textclock")
 benchmark(redraw_textclock, "redraw textclock")
 benchmark(e2e_tag_switch, "tag switch")
+benchmark(get_configuration_dir, "gears.filesystem.get_configuration_dir")
+benchmark(config_path, "awesome.config_path")
 
 runner.run_steps({ function() return true end })
 
```
Results with Lua 5.3 (`awesome.config_path` is ~2.5 times faster than `gears.filesystem.get_configuration_dir`, but both are so fast that it does not matter at all):
```
$ make && BENCHMARK_EXACT=1 tests/run.sh test-benchmark.lua
[...]
== Running test-benchmark.lua ==
Lua 5.3	nil
[...]
gears.filesystem.get_configuration_dir: 5.047e-07  sec/iter (1985470 iters, 4.014 sec for benchmark)
 awesome.config_path: 1.99562e-07 sec/iter (5015226 iters, 3.035 sec for benchmark)
```
Results with LuaJIT (I had some problems with LuaJIT, hence the different way of starting the benchmark; there is also a time difference of about factor 2.5):
```
$ make  && (cd .. && CMAKE_BINARY_DIR=/tmp/awesome/build-luajit BENCHMARK_EXACT=1 tests/run.sh test-benchmark.lua)
[...]
== Running test-benchmark.lua ==
Lua 5.1	LuaJIT 2.1.0-beta3
[...]
gears.filesystem.get_configuration_dir: 3.10971e-07 sec/iter (3219824 iters, 2.985 sec for benchmark)
 awesome.config_path: 1.24845e-07 sec/iter (8012502 iters, 2.029 sec for benchmark)
```
So, coming back to the original question:

> why not awesome.config_path?

I'd say "because it does not provide Lua with any extra information that it does not already have to" and "string processing is better done in Lua than in C, because that leads to less errors like segfaults and leaks". "Time" clearly is not a valid argument here since this is already blazingly fast.

Elv13:
I also agree with @psychon it doesn't provide something we don't already have. It isn't accessed often anyway so there is little value in adding more C code for it. It's yet another API that @timidger will have to implement for the Wayland version. Yet more duplicated code between way-cooler and Awesome.

actionless:
keeping in mind what PR should receive at least two approvals and i also don't like idea moving to C things are doable in Lua easier i am closing this

Sorky:
> I still fail to see the point in this, but apparently that doesn't count for much...
Of course it matters - As no one said no I figured it was worth pursuing

> "Time" clearly is not a valid argument here since this is already blazingly fast.
Time isn't a factor I was pushing - repeating calculations that only need to be done once was my point

> I'd say "because it does not provide Lua with any extra information that it does not already have to"
Somewhat true, but I'm the only use cases I'm aware of for awesome.conffile are...
```
/_testing/awesome/awesomerc.lua:111:   { "edit config", editor_cmd .. " " .. awesome.conffile },
/_testing/awesome/lib/awful/util.lua:144:    local c = util.checkfile(capi.awesome.conffile)
-- and
/_testing/awesome/lib/gears/filesystem.lua:116:    return awesome.conffile:match(".*/") or "./"
```
Not that I'm suggesting getting rid of or changing it

> because that leads to less errors like segfaults and leaks".
Yes my first attempt did that but I cannot see that happening now

> "string processing is better done in Lua than in C
In my ideal world, code that the user shouldn't have to touch shouldn't be seen at all - Sure a similar thing could be achieved in lua - But speed isn't my point even though  the gears version could be improved so it only does the generation once...
```
local config_dir = awesome.conffile:match(".*/") or "./"
function filesystem.get_configuration_dir()
    return config_dir
end
```

My focus is on the user needing to do code that they could avoid...
- in rc.lua to determine the theme dir
- in the theme to determine the icon path
- in test code that wants to know the config_path

Ie: `config_path = require( "gears.filesystem" ).get_configuration_dir()` and using it .Vs. just using `awesome.config_path` inline

IMHO: Tested, Documented, Logical, Neat

Anyway - That's my 2c

Sorky:
> Yet more duplicated code between way-cooler and Awesome.

Never knew of it

Looking it up... 'Way Cooler is currently not being developed right now'

Sorky:
> There is a features in our C object support to enable Lua implemented properties on C object. It is disabled by default. The `mouse` module enables it and some of the open PRs enable it for `root` too. It's only a few line change.

I'm guessing it'd still need a "require" anyway which defeats the purpose AFAIC

> But still, the API already exists in gears, so it's irrelevant here for this discussion.

Maybe I should do the minor improvement to gears then...
```
local config_dir = awesome.conffile:match(".*/") or "./"
function filesystem.get_configuration_dir()
    return config_dir
end
```
And for the next couple of functions as well

psychon:
> ```
> local config_dir = awesome.conffile:match(".*/") or "./"
> function filesystem.get_configuration_dir()
>     return config_dir
> end
> ```

Before anyone actually sends a PR for the above: This does not work.

Place `require("gears.filesystem") error("error!")` in `~/.config/awesome/rc.lua`. What will happen is that `gears.filesystem` is loaded, calculates `config_dir = ~/.config/awesome/rc.lua` and then an error happens. The C code catches that and falls back to loading `/etc/xdg/awesome/rc.lua`. However, `gears.filesystem` is already loaded and already calculated the path, even though the right path now suddenly is `/etc/xdg/awesome/`.

Sorky:
> However, gears.filesystem is already loaded and already calculated the path, even though the right path now suddenly is /etc/xdg/awesome/.

Excellent point! I've seen example(s) like that and even if they are ok, it is a bit of a trap to be aware of - In the case of a restart should loaded modules get reloaded (if even possible)?

psychon:
>  In the case of a restart should loaded modules get reloaded (if even possible)?

"Restart" as in `awesome.restart()`? Here, everything is reloaded.
I would call what I described above a "fallback". And no, here it is not possible to throw away all the Lua state and start fresh (for various, bad reasons). We could mess with `package.loaded` and remove all entries, but that also feels hackish to me.

This "fallback keeps some state alive"-behaviour e.g. often causes people to have two wibars: One was created by their own config before the error caused a fallback, the second one was created by the default config.
(Back before we had naughty popups for fallbacks, people were often confused as to where this second wibar came from; The error messages now help a lot with that)

Elv13:
But really, it's important to take some facts into account here:

 * It isn't called all that often, the optimization is premature
 * It isn't anywhere near the top 1000 bottleneck by any length of the imagination
 * The current code works
 * Any change will have to handle more corner cases as @psychon points out, adding more, slower, complexity

Really, it's a bit pointless to spend time on this, even discussing it. If you (@Sorky) want to make AwesomeWM much, much faster, then there is a lot of low hanging fruits out there. As mentioned before, the fact that LGI doesn't always cache the bindings lookup is by far the largest bottleneck. It is fixable. Even then, on modern hardware it is irrelevant even for configs as insane as mine. Then the second is LGI using for-loops over non trivial amount of data lookup instead of using hash maps. Then the next one (AFAIK) is the `rc.lua` use of `gears.table.join` in a , I am currently "fixing" that one elsewhere, but it isn't merged yet. Another one is useless repaints caused by the same value being set twice in a widget property. Then stuff like `awful.client.tiled` and its cousin. Those are called often-ish and are recomputed every time instead of keeping track of the list. Then there is all the `O(N)` operations in the C core which are "slower" than hashing when there is more than ~15 clients open (but faster otherwise). 

Those non-performant code paths are dwarfing by many order of magnitude micro optimizing a function called a couple dozen time at startup and never again.

Sorky:
I'm keen to extend myself into some of these other areas, but I've got to learn to walk before I run (I believe I've at least passed the crawling phase) and the easiest way to learn is to question what I'm studying

Sorry if it detracts from your other efforts, but please do accept my sincere thanks for your feedback and guidance

PR #2600, opened 2019-01-21
Extract the awful.rules logic into a gears module

Elv13:
Fixes #2558

I will also use this for the notification filter pull request that comes once #2541 is merged

~~I did *not* test this pull request beside, "uh, Awesome doesn't crash on a Lua error when I start it". It's just a find&replace the "client" with "object" and function into methods.~~

### DIFF
--- a/lib/awful/rules.lua
+++ b/lib/awful/rules.lua
@@ -119,8 +119,8 @@ local gtable = require("gears.table")
 local a_place = require("awful.placement")
 local protected_call = require("gears.protected_call")
 local aspawn = require("awful.spawn")
-local gsort = require("gears.sort")
 local gdebug = require("gears.debug")
+local gmatcher = require("gears.matcher")
 local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
 
 local rules = {}
@@ -128,26 +128,14 @@ local rules = {}
 --- This is the global rules table.
 rules.rules = {}
 
+local crules = gmatcher()
+
 --- Check if a client matches a rule.
 -- @client c The client.
 -- @tab rule The rule to check.
 -- @treturn bool True if it matches, false otherwise.
 function rules.match(c, rule)
-    if not rule then return false end
-    for field, value in pairs(rule) do
-        if c[field] then
-            if type(c[field]) == "string" then
-                if not c[field]:match(value) and c[field] ~= value then
-                    return false
-                end
-            elseif c[field] ~= value then
-                return false
-            end
-        else
-            return false
-        end
-    end
-    return true
+    return crules:_match(c, rule)
 end
 
 --- Check if a client matches any part of a rule.
@@ -155,19 +143,7 @@ end
 -- @tab rule The rule to check.
 -- @treturn bool True if at least one rule is matched, false otherwise.
 function rules.match_any(c, rule)
-    if not rule then return false end
-    for field, values in pairs(rule) do
-        if c[field] then
-            for _, value in ipairs(values) do
-                if c[field] == value then
-                    return true
-                elseif type(c[field]) == "string" and c[field]:match(value) then
-                    return true
-                end
-            end
-        end
-    end
-    return false
+    return crules:_match_any(c, rule)
 end
 
 --- Does a given rule entry match a client?
@@ -176,8 +152,7 @@ end
 --   `except_any`).
 -- @treturn bool
 function rules.matches(c, entry)
-    return (rules.match(c, entry.rule) or rules.match_any(c, entry.rule_any)) and
-        (not rules.match(c, entry.except) and not rules.match_any(c, entry.except_any))
+    return crules:matches_rule(c, entry)
 end
 
 --- Get list of matching rules for a client.
@@ -186,13 +161,7 @@ end
 --   "except_any" keys.
 -- @treturn table The list of matched rules.
 function rules.matching_rules(c, _rules)
-    local result = {}
-    for _, entry in ipairs(_rules) do
-        if (rules.matches(c, entry)) then
-            table.insert(result, entry)
-        end
-    end
-    return result
+    return crules:matching_rules(c, _rules)
 end
 
 --- Check if a client matches a given set of rules.
@@ -201,22 +170,22 @@ end
 --   `except` and `except_any` keys.
 -- @treturn bool True if at least one rule is matched, false otherwise.
 function rules.matches_list(c, _rules)
-    for _, entry in ipairs(_rules) do
-        if (rules.matches(c, entry)) then
-            return true
-        end
-    end
-    return false
+    return crules:matches_rules(c, _rules)
+end
+
+--- Remove a source.
+-- @tparam string name The source name.
+-- @treturn boolean If the source was removed,
+function rules.remove_rule_source(name)
+    return crules:remove_matching_source(name)
 end
 
 
--- Contains the sources.
--- The elements are ordered "first in, first executed". Thus, the higher the
--- index, the higher the priority. Each entry is a table with a `name` and a
--- `callback` field. This table is exposed for debugging purpose. The API
--- is private and should be modified using the public accessors.
-local rule_sources = {}
-local rule_source_sort = gsort.topological()
+--- Apply awful.rules.rules to a client.
+-- @client c The client.
+function rules.apply(c)
+    return crules:apply(c)
+end
 
 --- Add a new rule source.
 --
@@ -257,73 +226,13 @@ local rule_source_sort = gsort.topological()
 -- @tparam[opt={}] table precede A list of names of sources this source have a
 --  priority over.
 -- @treturn boolean Returns false if a dependency conflict was found.
-function rules.add_rule_source(name, callback, depends_on, precede)
-    depends_on = depends_on  or {}
-    precede    = precede or {}
-    assert(type( depends_on ) == "table")
-    assert(type( precede    ) == "table")
+-- @function awful.rules.add_rule_source
 
-    for _, v in ipairs(rule_sources) do
-        -- Names must be unique
-        assert(
-            v.name ~= name,
-            "Name must be unique, but '" .. name .. "' was already registered."
-        )
+function rules.add_rule_source(name, cb, ...)
+    local function callback(_, ...)
+        cb(...)
     end
-
-    local new_sources = rule_source_sort:clone()
-
-    new_sources:prepend(name, precede    )
-    new_sources:append (name, depends_on )
-
-    local res, err = new_sources:sort()
-
-    if err then
-        gdebug.print_warning("Failed to add the rule source: "..err)
-        return false
-    end
-
-    -- Only replace the source once the additions have been proven safe
-    rule_source_sort = new_sources
-
-    local callbacks = {}
-
-    -- Get all callbacks for *existing* sources.
-    -- It is important to remember that names can be used in the sorting even
-    -- if the source itself doesn't (yet) exists.
-    for _, v in ipairs(rule_sources) do
-        callbacks[v.name] = v.callback
-    end
-
-    rule_sources = {}
-    callbacks[name]    = callback
-
-    for _, v in ipairs(res) do
-        if callbacks[v] then
-            table.insert(rule_sources, 1, {
-                callback = callbacks[v],
-                name     = v
-            })
-        end
-    end
-
-    return true
-end
-
---- Remove a source.
--- @tparam string name The source name.
--- @treturn boolean If the source was removed
-function rules.remove_rule_source(name)
-    rule_source_sort:remove(name)
-
-    for k, v in ipairs(rule_sources) do
-        if v.name == name then
-            table.remove(rule_sources, k)
-            return true
-        end
-    end
-
-    return false
+    return crules:add_matching_function(name, callback, ...)
 end
 
 -- Add the rules properties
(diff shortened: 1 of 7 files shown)


lib/awful/rules.lua
```diff
@@ -201,22 +170,22 @@ end
...
-            return true
-        end
-    end
-    return false
+    return crules:matches_list(c, _rules)
+end
+
+--- Remove a source.
+-- @tparam string name The source name.
+-- @treturn boolean If the source was removed
```
blueyed:
```suggestion
-- @treturn boolean If the source was removed,
```

lib/awful/rules.lua
```diff
@@ -119,55 +119,31 @@ local gtable = require("gears.table")
 local a_place = require("awful.placement")
 local protected_call = require("gears.protected_call")
 local aspawn = require("awful.spawn")
-local gsort = require("gears.sort")
 local gdebug = require("gears.debug")
+local grules = require("gears.matcher")
```
blueyed: s/grules/gmatcher ?

blueyed: s/grules/gmatcher ?

lib/awful/rules.lua
```diff
@@ -619,7 +514,9 @@ end
 -- @client c The client.
 -- @tab props Properties to apply.
 -- @tab[opt] callbacks Callbacks to apply.
-function rules.execute(c, props, callbacks)
+-- @function awful.rules.execute
+
+crules.execute = function(_, c, props, callbacks)
```
blueyed: Should be moved also to the new module?!

Elv13: It is in `gears.matcher` too. This is a "virtual function override" of `gears.matcher.execute`. I kept it this way because `awful.rules` in Awesome 4.x still uses hardcoded prioritt groups instead of using the graph dependency solver. That code makes zero sense in `gears.matcher` and its features fully redundant with the add/remove_source methods.

Elv13:
I actually tested this PR and brought it closer to mergeability.

@blueyed Please review if the new API is flexible enough for you. I took the liberty to redesign it a bit compared to `awful.rules`. 

Now that the notification rewrite is merged, I want to use `gears.matcher` to handle the notification preset. Using it to filter when @psychon new clipboard API should migrate the content to Awesome would also be useful.

Veratil:
Do you think a little more explanation of use could be added? The example is a little lacking on discussing what's happening.

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
+--- A module to build a set of properties based on a graph of rules.
+--
+-- Rules come from a graph of sources. Each source depend on other and must be
```
blueyed:
```suggestion
-- Rules come from a graph of sources. Each source depends on others and must be
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+-- applied before other sources. For this module to work, the graph needs to be
```
blueyed:
```suggestion
-- applied before those. For this module to work, the graph needs to be solvable.
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+-- solvable.
```
blueyed:
```suggestion
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+--
```
blueyed: Do rules actually depend on others?  What about the root?

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+-- @DOC_text_gears_matcher_default_EXAMPLE@
+--
+-- @author Julien Danjou &lt;julien@danjou.info&gt;
+-- @copyright 2009 Julien Danjou
```
blueyed: Update maybe?!

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+function matcher:matches_rule(o, entry)
+    local match = self:_match(o, entry.rule) or self:_match_any(o, entry.rule_any)
+    return match
+        and (not self:_match(o, entry.except))
+        and (not self:_match_any(o, entry.except_any))
+end
+
+--- Get list of matching rules for an object.
+--
+-- If the second `rules` argument is not provided, the rules added with
```
blueyed:
```suggestion
-- If the `rules` argument is not provided, the rules added with
```

blueyed: But it only iterates over `rules`?!

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+end
+
+--- Check if an object matches a given set of rules.
+-- @param o The object.
+-- @tparam table rules The rules to check. List of tables with `rule`,
+--  `rule_any`, `except` and `except_any` keys.
+-- @treturn boolean True if at least one rule is matched, false otherwise.
+function matcher:matches_rules(o, rules)
+    for _, entry in ipairs(rules) do
+        if (self:matches_rule(o, entry)) then
```
blueyed:
```suggestion
        if self:matches_rule(o, entry) then
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+    end
+end
+
+--- Add a set of matching rules.
+--
+-- @tparam string name The provider name. It must be unique.
+-- @tparam table rules A set of rules (see how they work at the top of this
+--  page).
+-- @tparam[opt={}] table depends_on A list of names of sources this source
+--  depends on (sources that must be executed *before* `name`.
```
blueyed:
```suggestion
--  depends on (sources that must be executed *before* `name`).
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+-- @tparam[opt={}] table precede A list of names of sources this source have a
```
blueyed:
```suggestion
-- @tparam[opt={}] table precede A list of names of sources this source has a
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+--  priority over.
+-- @treturn boolean Returns false if a dependency conflict was found.
```
blueyed: Not true.  See suggestion below.

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+end
+
+--- Add a matching function.
+--
+-- @tparam string name The provider name. It must be unique.
+-- @tparam function callback The callback that is called to produce properties.
+-- @tparam gears.matcher callback.self The matcher object.
+-- @param callback.o The object.
+-- @tparam table callback.properties The current properties. The callback should
+--  add to and overwrite properties in this table
```
blueyed:
```suggestion
--  add to and overwrite properties in this table.
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+-- @tparam table callback.callbacks A table of all callbacks scheduled to be
+--  executed after the main properties are applied.
+-- @tparam[opt={}] table depends_on A list of names of sources this source depends on
+--  (sources that must be executed *before* `name`.
```
blueyed:
```suggestion
--  (sources that must be executed *before* `name`).
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+-- @tparam[opt={}] table precede A list of names of sources this source have a
```
blueyed:
```suggestion
-- @tparam[opt={}] table precede A list of names of sources this source has a
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+--  priority over.
+-- @treturn boolean Returns false if a dependency conflict was found.
+function matcher:add_matching_rules(name, rules, depends_on, precede)
+    local function matching_fct(_self, c, props, callbacks)
+        default_rules_callback(_self, c, props, callbacks, rules)
+    end
+
+    self._matching_rules[name] = rules
+
+    self:add_matching_function(name, matching_fct, depends_on, precede)
```
blueyed:
```suggestion
    return self:add_matching_function(name, matching_fct, depends_on, precede)
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+    new_sources:append (name, depends_on )
+
+    local res, err = new_sources:sort()
+
+    if err then
+        gdebug.print_warning("Failed to add the rule source: "..err)
+        return false
+    end
+
+    -- Only replace the source once the additions have been proven safe
```
blueyed:
```suggestion
    -- Only replace the source once the additions has been proven to be safe.
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+    self._rule_source_sort = new_sources
+
+    local callbacks = {}
+
+    -- Get all callbacks for *existing* sources.
+    -- It is important to remember that names can be used in the sorting even
+    -- if the source itself doesn't (yet) exists.
```
blueyed:
```suggestion
    -- if the source itself doesn't (yet) exist.
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+    return true
+end
+
+--- Remove a source.
+--
+-- This removes sources added with `add_matching_function` or
+-- `add_matching_rules`.
+--
+-- @tparam string name The source name.
+-- @treturn boolean If the source was removed
```
blueyed:
```suggestion
-- @treturn boolean If the source has been removed.
```

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+    self._rule_source_sort:remove(name)
+
+    for k, v in ipairs(self._matching_source) do
+        if v.name == name then
+            table.remove(self._matching_source, k)
+            return true
+        end
+    end
+
+    self._matching_rules[name] = nil
```
blueyed: Why set it to nil?

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+
+    self:_execute(o, props, callbacks)
+end
+
+-- Execute the rules for the object `o`.
+-- @param o The object.
+-- @tparam table props A list of properties to apply.
+-- @tparam table callbacks A list of callback to execute with the object `o` as
+--  sole argument. The callbacks are executed *before* applying the properties.
+-- @see gears.matcher.apply
```
blueyed: Why are callbacks applied first?  It used to be the other way around, i.e. callbacks could update/set properties that were also specified.

lib/gears/matcher.lua
```diff
@@ -0,0 +1,292 @@
...
+-- @return A new rule solver object.
+
+local function new()
+    local ret = {}
+
+    -- Contains the sources.
+    -- The elements are ordered "first in, first executed". Thus, the higher the
+    -- index, the higher the priority. Each entry is a table with a `name` and a
+    -- `callback` field. This table is exposed for debugging purpose. The API
+    -- is private and should be modified using the public accessors.
```
blueyed:
```suggestion
    -- is private and should only be modified using the public accessors.
```

lib/gears/matcher.lua
Elv13: > Do rules actually depend on others?

This is unchanged from `awful.rules`. If different *properties* have a dependency on one another, then they must be in different rule sources. Otherwise the array order is used.

lib/gears/matcher.lua
Elv13: I didn't write this code. It's cut/pasted from `awful.rules`. The only part I wrote is 3/4 of the graph source method.

lib/gears/matcher.lua
Elv13: whoops, forgot that line

It should join all `self._matching_rules`

lib/gears/matcher.lua
Elv13: If `:matches_rules()` wasn't missing a line, it would cause the removed rules to potentially still be evaluated.

lib/gears/matcher.lua
Elv13: My bad, I copy/pasted and kept the delayed loop instead of the "normal" loop.

Elv13:
@blueyed I added a commit to "fix" cmake. There's a few regressions. For one, the generated doc has some duplicated modules. Also all tests are executed all the time, which is what you wanted to avoid. Can you take a look and see what you wish to improve? Feel free to push to this branch, I wont have much time this week either.

Elv13:
@Veratil I tried to add some more ldoc shared between `awful.rules` and `gears.matcher`, but the way we build Awesome is fighting me. I created another "parking" pull request for the commit to fix this because I have no time to fix it right now.

Other: I would like this to be merged soon ish to unblock the remaining notification pull requests so Awesome v4.4 can be released.

Veratil:
> @Veratil I tried to add some more ldoc shared between `awful.rules` and `gears.matcher`, but the way we build Awesome is fighting me. I created another "parking" pull request for the commit to fix this because I have no time to fix it right now.

It should be okay for now. I think eventually having some examples of real-world use would benefit (since for some reason I just can't wrap my head around it much 🙄 ).

Elv13:
> It should be okay for now. I think eventually having some examples of real-world use would benefit (since for some reason I just can't wrap my head around it much roll_eyes ).

I am not sure if there is such real-world use cases to begin with. It will be used by `awful.rules` (to be renames `awful.client.rules`?), `naughty.rules` and `selection.rules`. I think it is mostly the only place it makes sense. I didn't design this code and I had some issues really understanding it too. This pull requests only extracts the code from `awful.rules` so it can be used by the other 2. This has a "lower friction factor" than designing a new rules/filters system for the notifications and selection APIs. Given I am the author of Tyrannical, you can understand that I don't love this API all that much. But then again, I try to align my contributions toward an unified design for all Awesome modules. It's like my work on `awful.widget.common` (tasklist/taglist). I don't like it, I wrote Radical to get rid of it, but Radical doesn't fit well into the existing API design, so I improve `awful.widget.common`...

Elv13:
Last call, I will merge this soon if there's no reviews. This is delaying v4.4 (because it is a dependency of the "last" notification rewrite pull request)

Issue #2596, opened 2019-01-19
Discussion: Potential gears.wallpaper improvements


Sorky:
Idea: Incorporate "set_wallpaper" into gears.wallpaper [remove code users don't need to touch in rc.lua]

Plus: Allow `beautiful.wallpaper = {'bg_1.png', 'bg_2.png'}`

Plus: Emphasize the use of gears.wallpaper.@  [ie: could use maximized, tiled, centered or fit

Plus: Update the doco for all the uses of beautiful.wallpaper and clarify the file path

Possible change from...
```
local function set_wallpaper(s)
    -- Wallpaper
    if beautiful.wallpaper then
        local wallpaper = beautiful.wallpaper
        -- If wallpaper is a function, call it with the screen
        if type(wallpaper) == "function" then
            wallpaper = wallpaper(s)
        end
        gears.wallpaper.maximized(wallpaper, s, true)
    end
end

-- Re-set wallpaper when a screen's geometry changes (e.g. different resolution)
screen.connect_signal("property::geometry", set_wallpaper)

awful.screen.connect_for_each_screen(function(s)
    -- Wallpaper
    set_wallpaper(s)
```
to...
```
-- Re-set wallpaper when a screen's geometry changes (e.g. different resolution)
screen.connect_signal("property::geometry", function() gears.wallpaper.maximized(wallpaper, s, true) end)

awful.screen.connect_for_each_screen(function(s)
    -- Wallpaper
    gears.wallpaper.maximized(wallpaper, s, true)
```
Plus incorporate something like this into gears.wallpaper...
```
local function resolve_wallpaper(wallpaper, path, s)
	local wp = wallpaper or beautiful.wallpaper
	local wp_type = type( wallpaper )
	local wp_path = wallpaper:sub(1, 1) == '/' and "" or (path or beautiful.wallpaper_path or beautiful.theme_path)
	local wp_screen = (s and wallpaper[s.index]) or wallpaper[1]

	if wp_type == "string" then return wp_path .. wallpaper
	elseif wp_type == "table" then return wp_screen and (wp_path .. wp_screen) or ""
	elseif wp_type == "function"	then return wp_path .. wallpaper( s ) or ""
	else return ""
	end
end
```
Finally...
`Call resolve_wallpaper from gears.wallpaper.maximized, tiled, centered & fit`

Benefits summary...
- Add table of backgrounds
- Reduce code users shouldn't need to touch in rc.lua
- Update and clarify docuymentation

[...]

PR #2605, opened 2019-01-22
Add get_random_file_from_dir to gears.filesystem

Sorky:
Get the name of a random file from a given directory

### DIFF
--- a/lib/gears/filesystem.lua
+++ b/lib/gears/filesystem.lua
@@ -153,4 +153,33 @@ function filesystem.get_dir(d)
     end
 end
 
+math.randomseed( os.clock() % 1 * 1e6 )
+
+--- Get the name of a random file from a given directory.
+-- @tparam string path The directory to search.
+-- @tparam[opt] table exts Specific extensions to limit the search to. eg:`{ "jpg", "png" }`
+--   If ommited, all files are considered.
+-- @treturn string|nil A randomly selected filename from the specified path (with
+--   a specified extension if required) or nil if no suitable file is found.
+function filesystem.get_random_file_from_dir(path, exts)
+    local files, valid_exts = {}, {}
+
+    -- Transforms { "jpg", ... } into { [jpg] = #, ... }
+    if exts then for i, j in ipairs(exts) do valid_exts[j:lower()] = i end end
+
+    -- Build a table of files from the path with the required extensions
+    local file_list = Gio.File.new_for_path(path):enumerate_children("standard::*", 0)
+    for file in function() return file_list:next_file() end do
+        if file:get_file_type() == "REGULAR" then
+            local file_name = file:get_display_name()
+            if not exts or valid_exts[file_name:lower():match(".+%.(.*)$") or ""] then
+               table.insert(files, file_name)
+            end
+        end
+    end
+
+    -- Return a randomly selected filename from the file table
+    return #files > 0 and files[math.random(#files)] or nil
+end
+
 return filesystem
--- a/spec/gears/filesystem_spec.lua
+++ b/spec/gears/filesystem_spec.lua
@@ -1,0 +1,138 @@
+---------------------------------------------------------------------------
+-- @author
+-- @copyright 2019
+---------------------------------------------------------------------------
+
+local gfs = require("gears.filesystem")
+
+describe("gears.filesystem", function()
+    local root = (os.getenv("SOURCE_DIRECTORY") or '.') .. "/spec/gears/"
+
+    -- Check filesystem.make_directories
+    it('Check filesystem.make_directories', function()
+        assert.is_true(gfs.make_directories(root .. "filesystem_tests/x/y"))
+    end)
+
+    -- Check filesystem.make_parent_directories
+    it('Check filesystem.make_parent_directories', function()
+        assert.is_true(gfs.make_directories(root .. "filesystem_tests/x/y"))
+    end)
+
+    -- Check filesystem.file_readable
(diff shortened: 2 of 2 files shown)


Sorky:
Derived from https://github.com/awesomeWM/awesome/issues/2596

Elv13:
from the CI

    lib/gears/filesystem.lua:167:22: unused loop variable i

please use `_` for the name of unused variables to prevent the linter from having this warning

lib/gears/filesystem.lua
```diff
@@ -153,4 +153,30 @@ function filesystem.get_dir(d)
     end
 end
 
+math.randomseed( os.clock() % 1 * 1e6 )
+--- Get the name of a random file from a given directory.
```
Elv13: Please add a newline before the `---`

lib/gears/filesystem.lua
```diff
@@ -153,4 +153,30 @@ function filesystem.get_dir(d)
...
+-- @tparam string path The directory to search.
+-- @tparam[opt] table exts Specific extensions [eg: { "jpg", "png" }] to limit
```
Elv13: please change to \`{ "jpg", "png" }\`

lib/gears/filesystem.lua
Sorky: Will do

lib/gears/filesystem.lua
Sorky: This suit?
```
-- @tparam[opt] table exts Specific extensions to limit the search to.
--   If ommited, all files are considered. eg:`{ "jpg", "png" }`
```

Sorky:
> Beside that a little unit or integration test would be welcome to make our coverage bot happy. You can't prove randomness, but at least our CI can hit the code and may help detect if, for example, a LGI change was to break it.

Would you like a spec/gears/filesystem_spec.lua

Elv13:
> Would you like a spec/gears/filesystem_spec.lua?

This would be appreciated yes.

lib/gears/filesystem.lua
Elv13: yes, thank you

Sorky: Updating to...
```
-- @tparam[opt] table exts Specific extensions to limit the search to. eg:`{ "jpg", "png" }`
--   If ommited, all files are considered.
```

Sorky:
Added test for my changes plus several other existing areas.

Extra tests for areas outside of the scope of this PR that could be added marked as `-- TODO:`

lib/gears/filesystem.lua
```diff
@@ -153,4 +153,33 @@ function filesystem.get_dir(d)
...
+        if file:get_file_type() == "REGULAR" then
+            local file_name = file:get_display_name()
+            if not exts or valid_exts[file_name:lower():match("%.(.*)$")] then
+               table.insert(files, file_name)
+            end
+        end
+    end
+
+    -- Return a randomly selected filename from the file table
+    return #files > 0 and files[math.random(#files)] or ""
```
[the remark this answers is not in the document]
Sorky: OK

Elv13:
The tests are failing in the latest revision. Once fixed, I will approve the pull request, the code itself is ready to merge as far as I am concerned.

Sorky:
> [#2596 (comment)](https://github.com/awesomeWM/awesome/issues/2596#issuecomment-459641931)
> 
> It was in my category of "to be handled once someone complains" because too many review cycle suck. But... Someone complained.

A problem is a problem - No need to wait - God forbid I found one in my tests which I improved as well

Veratil:
Just a side thought which would be a refactor of the code, but it could use gears.string.endswith?

Elv13:
> Just a side thought which would be a refactor of the code, but it could use gears.string.endswith?

That would make supporting multiple extensions a bit problematic, right? Given some people with have the bright idea of downloading a "The best 1 million wallpapers for 2019" style packs, looping string comparison instead of compiled pattern is a potential issue.

Veratil:
> > Just a side thought which would be a refactor of the code, but it could use gears.string.endswith?
> 
> That would make supporting multiple extensions a bit problematic, right? Given some people with have the bright idea of downloading a "The best 1 million wallpapers for 2019" style packs, looping string comparison instead of compiled pattern is a potential issue.

True true. It would probably be looping through and checking each extension. Like I said just a side thought. 😉

Sorky:
Updated regex, added all "." in filename tests and ensured { "" } works as well

Ready for the next round

Sorky:
Not sure how the timing is going but as this is new and heavily tested and reviewed it might be good for 4.3?

Elv13:
> Not sure how the timing is going but as this is new and heavily tested and reviewed it might be good for 4.3?

4.3 is already released, so this will get into v4.4. Distributions don't backport features, so even *if* I stop being against bugfixes releases, it wouldn't get integrated into existing packaged versions anyway. Right now I prefer to only release X.y versions because in the past we had X.y.z versions and git-master ended up thousands of commits ahead of the stable branch and ended up never being released.

As for this PR, well, time's up, I was waiting for the second review to auto-merge it, but this isn't happening, so I merge it.

Sorky:
Any idea when the 4.3 release will get to the ubuntu apt?

Elv13:
In the current Ubuntu? Never.

It will be in the next upgrade, so 19.10 or 20.04 depending if you use LTS or not. They don't seem interested in pulling it for 19.04, maybe nobody asked for it.

It's not a big deal, v4.2 was a solid release.

PR #2608, opened 2019-01-25
Fix terminal exit after selecting manual from menu

demcgovern:
Installed on Arch Linux. Was having issue with terminal (termite) immediately exiting after choosing "manual" from the menu.


psychon:
Related: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=489407

demcgovern:
Sorry. Thanks for the link, psychon.

Issue #2516, opened 2018-12-27
(Possibly) rewrite how widget shapes are implemented


psychon:
I just saw a question on StackOverflow about wanting a different stroke mode than "center", namely "inside" and "outside". I guess such a magic would also be interesting for widget shapes, which currently (sometimes?) do some weird things to work around some issues which would be fixed with such a stroke mode.

https://stackoverflow.com/questions/53934276/stroke-alignment-in-cairo/53943201

psychon:
Okay, so we had some problems which shapes lately and some attempts at solving them. This comment is on how about shapes are supposed to work / look like. This is not yet about implementations problems caused by anti-aliasing (thus: the below will not avoid those and some of the images have artifacts).

There are several possible ways to define a shape with a border (really, shapes without borders are so much simpler and the current implementation is fine for those). In the following, I show all of the ones that I can come up with...

All the following examples will be based on this code, only the `draw` function will be different:
```lua
local cairo = require("lgi").cairo
local shape = require("gears.shape")

local unpack = unpack or table.unpack
local fill_color = { 0, 0, 0 }
local border_color = { 1, 1, 1 }

local function draw(cr, width, height, shap, bw)
    error("implement this")
end

local function my_pie(cr, width, height)
    return shape.pie(cr, width, height, 0.5, 5, math.min(width, height) / 2)
end

local shapes = { shape.rounded_rect, shape.infobubble, shape.powerline, my_pie }
local borders = { 0, 10, 15 }

local spacing = 10
local base_size = 100
local width = (base_size + spacing) * #borders
local height = (base_size + spacing) * #shapes
local img = cairo.ImageSurface(cairo.Format.RGB24, width, height)

local cr = cairo.Context(img)
cr:set_source_rgb(0, 1, 0)
cr:paint()

cr:translate(spacing / 2, spacing / 2)

-- draw a grid
for x=0, #borders - 1 do
    cr:move_to(x * (base_size + spacing), 0)
    cr:rel_line_to(0, height - spacing)
    cr:move_to(x * (base_size + spacing) + base_size, 0)
    cr:rel_line_to(0, height - spacing)
end
for y=0, #shapes - 1 do
    cr:move_to(0, y * (base_size + spacing))
    cr:rel_line_to(width - spacing, 0)
    cr:move_to(0, y * (base_size + spacing) + base_size)
    cr:rel_line_to(width - spacing, 0)
end
cr:set_line_width(2)
[...]

PR #2609, opened 2019-01-25
Change the way the shape is done in the background container

psychon:
Previously, the background container "just" used the shape and drew a
line around it. This means that half the line will be inside of the
shape and half of it will be outside. Thus, this hides the actual shape
that is used.

The following example images are generated for this example:
```lua
    s.mywibox.height = 40
    s.mywibox:setup {
        {
            layout = wibox.layout.stack,
            wibox.widget.textbox("This text\nis behind"),
            {
                layout = wibox.container.background,
                shape = gears.shape.rounded_rect,
                shape = gears.shape.infobubble,
                bg = "#0008",
                fg = "#fff8",
                shape_border_color = "#00f8",
                shape_border_width = 2,
                wibox.widget.textbox("Test with some long text\nand a line break\nactually two")
            }
        },
    }
```

This commit changes that so that the line is added outside of the shape.
It does this via some tricks:

- In :before_draw_children(), :push_group() is used to redirect drawing
  of the child widget to a temporary surface. (This surface is initially completely transparent).
- In :after_draw_children() we thus get the content in a temporary surface.
out
- In :after_draw_children(), the border is added to this group.
  + For this, another temporary surface is created. It will be used as a
    mask.
  + The inside of the shape on this mask is cleared, everything else is
    filled. Thus, the mask now contains everything "not content".
out
  + Everything inside the mask is filled with the background color.
out
- Also in :after_draw_children(), the group is drawn to the actual
  target surface.
  + Again, this needs a mask.
  + This time, we draw the shape to the mask with twice the border width.
    Thus, half of this line will be outside of the shape.
out
  + Then, the shape itself is also filled so that the mask contains the
    shape and the border.
out
  + This mask is then used to copy the right parts of the temporary
    surface were the child widget and border was drawn to the actual
    target surface that will be visible on screen.
Before this, the target contains:
out
Afterwards:
out

This approach has some upsides. Because we no longer have "half the
border" above content, colors with some transparency work fine for the
border. Also, this should avoid issues with anti-aliasing, because e.g.
the border is not just drawn with the border width, but also further out
to everything else so that the background cannot "bleed through".

Fixes: https://github.com/awesomeWM/awesome/issues/2516


psychon:
```
lua: ../../../../src/cairo-pattern.c:1062: cairo_pattern_reference:
   Assertion `(_cairo_atomic_int_get (&(&pattern->ref_count)->ref_count) > 0)' failed.
```
Yay, another LGI bug (that somehow I did not hit locally).

Also: This commit effectively removes `shape_clip` support by always enforcing `shape_clip=true` (but in a way that does not produce the antialiasing issues that @actionless was seeing -- I finally understand were they come from!). Does anyone really want `shape_clip=false`?

actionless:
i think shape_clip=false could be imitated easily with stack widget with actual background and with transparent background+shape

actionless:
@psychon which lua and lgi versions you have locally?

actionless:
mb some of the shims doing something weird?

psychon:
It's one of the lines that I did not modify: https://github.com/awesomeWM/awesome/pull/2609/commits/028fda93438ce22e4ae5377cd4182369c985b435#diff-885682a6d2d38210025f1f3b0d402f49L45
This calls `set_source` to draw the background, but the background pattern is invalid (I guess: already freed).

So... this does not look like a new bug.

psychon:
I pushed a work-around. Let's see what Travis says.

Elv13:
> [15:12] <psychon> Elv13: should the "shape outside" PR be merged?

You decide. After all these problems, to me it looks scary to merge this as the last commit. Still, it is a very real and (to some) very annoying bug.

psychon:
I just remembered that there is still the `shape_clip` problem. How should we handle that?

* Deprecate `shape_clip` because no one wants `shape_clip=false' anyway and it is just an historical accident that this option exists.
* Implement `shape_clip=false` again.

actionless:
i would choose deprecating

Elv13:
Add a comment in the "behavior changes" section, print a gdebug warning when it is explicitly set to `false`. Make that warning mention to move the code to a `stacked` layout. Be done with it?

psychon:
Done (except for the "behaviour changes" section, there is no such section yet, is there?)

psychon:
Sounds like a good idea (except that I bet we will forget to update the `NEWS`, but that doesn't mean that this is worse than the old approach). Done in #2624.

lib/wibox/container/background.lua
```diff
@@ -268,20 +268,16 @@ function background:get_shape_border_color()
...
--- @property shape_clip
--- @tparam boolean value If the shape clip is enable
-
 function background:set_shape_clip(value)
-    if self._private.shape_clip == value then return end
-
-    self._private.shape_clip = value
-    self:emit_signal("widget::redraw_needed")
+    if value then return end
+    require("gears.debug").print_warning("shape_clip property of background container was removed."
```
Veratil: This is removing, not deprecating as the commit says, `shape_clip`?

psychon: Two commits before that (the commit that does "the big rewrite") does the actual removing. This commit just turns "silently ignore this option" into "un-silently ignore this option".
So, this is only removing the internal implementation of the setter, but its "outside observable" behaviour is basically the same (the only observable difference is that after setting the option to `false`, the getter will still return `true`; no visual differences).

Veratil: Ah! Understood. Just used to seeing a deprecation warning rather than a removed one. :)

actionless:
so, since release is out, would we merge it for testing by a wider audience?

actionless:
seems new travis is not aware of this PR

Elv13:
> seems new travis is not aware of this PR

Yes, it is documented. It's why I waited for the release to risk pressing the migrate button. The old builds are *not* imported. Travis-CI.org and .com apparently have different databases and the import tool is documented not to import this old builds.

actionless:
it broken background's `set_fg()` under some tricky usecase, i'll create a separate ticket once i'll extract it to be reproducible on default rc.lua

Issue #2604, opened 2019-01-22
arcchart doesn't display the value correctly when the value is too small and rounded_edge is set to true


dekeract10:
Output of `awesome --version`:
awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

I noticed a bug concerning wibox.container.arcchart. When the rounded edge is set to true, and the value is too small

**How to reproduce the issue:**
Add this to theme.lua and load it in the wibox,

```lua
local github =  wibox.widget {
   rounded_edge = true,
   forced_width = 20,
   thickness = 5.5,
   bg = "#FFFFFF",
   colors = {"#FFCC00"},
   value = 5, --this is to low to be
   min_value = 0,
   max_value = 100,
   widget = wibox.container.arcchart,
}
```

**Actual result:**
![image](https://user-images.githubusercontent.com/46914170/51512741-d8a7f100-1e07-11e9-823a-37b368911fe4.png)

**Expected result:**
Something like this, but with rounded corners
![image](https://user-images.githubusercontent.com/46914170/51512774-faa17380-1e07-11e9-89dd-ef0782a63d43.png)

The workaround I found is to change the value to the minimum one that is displayable if it goes below it. I also found that forced_width, and thickness influence this behaviour.

Thank you for your patience.

psychon:
I'm not quite sure what is going on, but this code causes it:
https://github.com/awesomeWM/awesome/blob/aeab2a70e908389e6ac9b130928c7b92582d40f9/lib/gears/shape.lua#L465-L477
```
before:
arc_length=	1.4922565104552	start_angle=	-0.31415926535898	end_angle=	0
after:
arc_length=	-4.0077434895448	start_angle=	0.26478810306207	end_angle=	-0.57894736842105
```
This code seems to try to make the arc a little shorter to add space for the "rounded heads" that are added later. However, the value here is so small that `start_angle` becomes larger than `end_angle` and thus we end up with an almost full circle. I guess in this case this just should not draw an actual arc at all, only the "heads"...?
(This can also be almost seen in the screenshot: There is a single almost white pixel about where the arc should be, i.e. the arc is now exactly on the parts where it should not be.)

Elv13:
[...]

PR #2610, opened 2019-01-26
Fix the arcchart rounded edges

Elv13:
Ref #2604

### DIFF
--- a/lib/gears/shape.lua
+++ b/lib/gears/shape.lua
@@ -472,8 +472,11 @@ function module.arc(cr, width, height, thickness, start_angle, end_angle, start_
     if end_rounded then
         arc_length = arc_length - thickness/2
 
-        -- And back to angles
-        end_angle = start_angle + (arc_length/(radius - thickness/2))
+        -- And back to angles. Also make sure to avoid underflowing when the
+        -- rounded edge radius is greater than the angle delta.
+        end_angle = start_angle + math.max(
+            0, arc_length/(radius - thickness/2)
+        )
     end
 
     -- The path is a curcular arc joining 4 points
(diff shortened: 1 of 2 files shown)


lib/gears/shape.lua
```diff
@@ -472,8 +472,11 @@ function module.arc(cr, width, height, thickness, start_angle, end_angle, start_
     if end_rounded then
         arc_length = arc_length - thickness/2
 
-        -- And back to angles
-        end_angle = start_angle + (arc_length/(radius - thickness/2))
+        -- And back to angles. Also make sure to avoid underflowing when the
+        -- rounded edge radius is greater than the angle delta.
+        end_angle = math.max(
+            start_angle, start_angle + (arc_length/(radius - thickness/2))
```
psychon: Can this be rewritten as `end_angle = start_angle + math.max(0, ...)`? I was going to say "but angles describe a circle, after going around it once you are back to where you started" (and other complicated things). Before someone else is also confused, I'm suggesting this reformulation to make the intention here clearer. (Assuming I understood the intention correctly...)

psychon: Also, doesn't the subtraction around `start_angle` in line 469 need the same treatment?

psychon: I seem to have forgotten to press "comment" or something:

Doesn't the subtraction with `start_angle` that happened some lines above need the same treatment?

Elv13: > Doesn't the subtraction with start_angle that happened some lines above need the same treatment?

I tested with 0.001 and it rendered correctly. Small values may be a bit off where they "should be" but I think we can neglect this fact given it renders properly.

> 'm suggesting this reformulation to make the intention here clearer. (Assuming I understood the intention correctly...)

ok

Issue #2560, opened 2019-01-08
wibox.widget.systray triggers parent widget mouse::leave when mouse is inside a systray icon


sezaru:
Output of `awesome --version`:

awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

**How to reproduce the issue:**

I'm using a modified version of https://github.com/elenapan/dotfiles awesome config.

This config have a sidebar which is shown only when I move the mouse to the right border of the monitor and will hide itself when the mouse is not inside the sidebar area (via mouse::leave signal).

I'm trying to add a systray to that sidebar, so I added this code to the `sidebar.lua` file:
```
local systray = wibox.widget.systray()
systray:set_base_size(30)

local traybox = wibox.widget{
   pad(0),
   {
      {
         pad(3),
         systray,
         pad(3),
         layout = wibox.layout.align.horizontal,
      },
      shape         = gears.shape.rounded_bar,
      bar_shape     = gears.shape.rounded_bar,
      bg     = '#ff0000',
      widget = wibox.container.background,
   },
   pad(0),
   expand = "none",
   layout = wibox.layout.align.horizontal,
}
```

And added the traybox widget to the sidebar widgets hierarchy. This is the result for now:
![screenshot from 2019-01-07 22-01-32](https://user-images.githubusercontent.com/279828/50800571-e9b11800-12c7-11e9-8525-df4d7c1dec48.png)

(Yes, I know, the background of the nm-applet is broken, but as far as I know this is because of my gtk+ version).

**Actual result:**

Now I just need to move the mouse the a systray icon (in this case the nm-applet one) and the sidebar will trigger the `mouse::leave` event and hide the entire sidebar.

**Expected result:**

Since the systray widget is a parent of my sidebar, when I move the mouse inside it, it should not trigger the sidebar `mouse::leave` event and, as consequence, not hide the entire sidebar.

Elv13:
Umm, that will be a bit tricky to handle. The systray icons *are* X11 windows and they are *not* part of the bar. So from the bar point it view, it lost focus, so this behavior is understandable. However for your use case it sucks. The easy workaround would be to check if the mouse is still within the bar geometry and not hide it. The proper fix would be to mess with some fragile protocols and risk regressions.

[...]

PR #2611, opened 2019-01-26
Ignore all Enter/Leave events with detail==Inferior

psychon:
These events are generated when the mouse pointer moves between our
window and one of its child windows. For our purposes, this never counts
as an enter or leave, so just ignore these events completely.

Fixes: https://github.com/awesomeWM/awesome/issues/2560

### DIFF
--- a/event.c
+++ b/event.c
@@ -582,13 +582,19 @@ event_handle_leavenotify(xcb_leave_notify_event_t *ev)
 
     globalconf.timestamp = ev->time;
 
-    if(ev->mode != XCB_NOTIFY_MODE_NORMAL)
+    /*
+     * Ignore events with non-normal modes. Those are because a grab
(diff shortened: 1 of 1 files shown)


psychon:
I don't quite understand your description, but: are you changing the mouse cursor in Lua somehow? (= Does this also happen with the default config?)
Normally, the mouse cursor is controlled by the X11 server and awesome (by itself) does not change it. So, awesome not reacting to some events should not influence the current cursor...

actionless:
no, i have on titlebar signal on mouse enter and mouse leave

and after this patch when i moving mouse cursor (manually, not with script), from titlebar to inside the client -- signal for titlebar mouse leave not fired

psychon:
...but...but...but... Meh :-(

PR #2617, opened 2019-01-27
Honor *some* Enter/Leave events with detail==Inferior

psychon:
Commit 19086d70fa9eb made AwesomeWM ignore all Enter/Leave events with
detail==Inferior. However, when leaving a titlebar by moving the cursor
into the actual client window, the corresponding event actually has
detail==Inferior. The same is true for the opposite direction: Leaving a
client window by entering the titlebar.

Thus, this commit fixes the code to handle theses cases again.

This was tested with the following in rc.lua (and with a systray icon
and a client with titlebars).

```
screen[1].mywibox:connect_signal("mouse::enter", function() print("enter mywibox") end)
screen[1].mywibox:connect_signal("mouse::leave", function() print("leave mywibox") end)
client.connect_signal("mouse::enter", function() print("enter client") end)
client.connect_signal("mouse::leave", function() print("leave client") end)
client.connect_signal("manage", function(c)
    local d = c:titlebar_top()
    d:connect_signal("mouse::enter", function() print("enter titlebar") end)
    d:connect_signal("mouse::leave", function() print("leave titlebar") end)
end)
```

Fixes: https://github.com/awesomeWM/awesome/pull/2611#issuecomment-457892746

@actionless Could you test this?

### DIFF
--- a/event.c
+++ b/event.c
@@ -586,24 +586,38 @@ event_handle_leavenotify(xcb_leave_notify_event_t *ev)
      * Ignore events with non-normal modes. Those are because a grab
      * activated/deactivated. Everything will be "back to normal" after the
      * grab.
-     *
-     * Also ignore events with detail "inferior". This means that the window was
-     * left for a child window, i.e. the pointer is basically still inside of
-     * our window.
      */
-    if(ev->mode != XCB_NOTIFY_MODE_NORMAL || ev->detail == XCB_NOTIFY_DETAIL_INFERIOR)
+    if(ev->mode != XCB_NOTIFY_MODE_NORMAL)
         return;
 
     if((c = client_getbyframewin(ev->event)))
     {
-        luaA_object_push(L, c);
-        luaA_object_emit_signal(L, -1, "mouse::leave", 0);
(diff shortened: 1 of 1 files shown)


Elv13:
Semi related to a bug reported on IRC yesterday about these signal not being reliable:

> [Saturday, January 26, 2019] [4:14:49 PM EST]<mahe> I'm not familiar with awesome's under-the-hood C code but my guess is that the following line is the culprit: https://github.com/awesomeWM/awesome/blob/9bfd6a22cc865fa02dad2621d2a7404e60f78a0f/event.c#L497
> [Saturday, January 26, 2019] [4:16:43 PM EST]<mahe> In my case the globalconf.drawable_under_mouse is most likely not unref'ed when the button moves away due to the maximize, thus the quoted line presumably still triggers on the next enter event, preventing the mouse::enter from being emitted.

Elv13:
> Only partly.

I know, the point was that this bug is about enter/leave events on the titlebar are not sent. That was another way to trigger the "same" bug (of not having these event sent when entering the titlebar). Note that I didn't try to confirm this user report, but look looks legit at first glance.

PR #2619, opened 2019-01-28
doc: Fix a copy+paste mistake

Elv13:
The current `awful.widget.layoutlist` is a fork of the so far uncommitted `naughty.widget.actionlist`. It was created because some of the support code for the new `naughty` implementation needed "easier to merge" usage examples. The `layoutlist` was chosen because it was both a low hanging fruit and genuinely useful.

Thanks to @actionless for spotting it


Elv13:
Umm, @awesome-robot isn't wrong, one of the recently published SVG example seems to be a little underterministic. Oh well, it's only a couple pixel. I don't think it's worth holding the release, but I will have to investigate eventually.

PR #2621, opened 2019-01-28
Use reproducible timestamps in more doc examples

psychon:
This adds require("_date") to some example tests that use the current
date. This allows reproducibility by replacing os.date() with a function
that uses a static date. See commit 9d7eaf020d75a for details.


psychon:
This does not work :-(
The textclock does not use `os.date`.

PR #2622, opened 2019-01-28
textclock: Produce reproducible output

psychon:
When $SOURCE_DIRECTORY is set, we are most likely currently running the
examples test, i.e. generating images. These images end up in the
documentation.

To make the images reproducable, i.e. independent from the current time,
this commit makes the textclock honor $SOURCE_DATE_EPOCH if
$SOURCE_DIRECTORY is set.

See commit 9d7eaf02 for some more details.

### DIFF
--- a/lib/wibox/widget/textclock.lua
+++ b/lib/wibox/widget/textclock.lua
@@ -17,6 +17,17 @@ local TimeZone = glib.TimeZone
 
 local textclock = { mt = {} }
 
+local DateTime_new_now = DateTime.new_now
+
+-- When $SOURCE_DATE_EPOCH and $SOURCE_DIRECTORY are both set, then this code is
+-- most likely being run by the test runner. Ensure reproducible dates.
+local source_date_epoch = tonumber(os.getenv("SOURCE_DATE_EPOCH"))
+if source_date_epoch and os.getenv("SOURCE_DIRECTORY") then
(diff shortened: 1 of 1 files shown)


psychon:
CC @reinerh Do you have a better idea for this? This feels a bit hacky, but might still be the cleanest option...

reinerh:
It works for the texclock, but I also see differences in the calendar widget (i.e. a different day is highlighted in AUTOGEN_awful_popup_wiboxtypes.svg when e.g. building in different timezones).
And a different date is also displayed in AUTOGEN_awful_tooltip_textclock.svg (the date+time in the tooltip) and in AUTOGEN_awful_tooltip_textclock2.svg.

psychon:
>  (i.e. a different day is highlighted in AUTOGEN_awful_popup_wiboxtypes.svg when e.g. building in different timezones).
> And a different date is also displayed in AUTOGEN_awful_tooltip_textclock.svg (the date+time in the tooltip) and in AUTOGEN_awful_tooltip_textclock2.svg.

Oh, I'm stupid. All of them do indeed use `os.date`. I will re-open #2621 (well, re-submit, I cannot re-open that one since the branch was already deleted).

psychon:
@reinerh Oh and of course: Thanks for testing!

PR #2624, opened 2019-01-28
Start writing NEWS for the next release

psychon:
This already assumes that #2609 is merged.

### DIFF
--- a/docs/89-NEWS.md
+++ b/docs/89-NEWS.md
@@ -1,9 +1,27 @@
 # NEWS
 
-<a name="v43"></a>
-# Awesome window manager framework version 4.3 changes
+<a name="v44"></a>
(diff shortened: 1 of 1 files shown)


docs/89-NEWS.md
```diff
@@ -1,10 +1,28 @@
 # NEWS
 
-<a name="v43"></a>
-# Awesome window manager framework version 4.3 changes
+<a name="v44"></a>
+# Awesome window manager framework version 4.4 changes
```
Elv13: It would be nice to add the word "draft" or avoid using 4.4. This page is published on the website and the git doc is used by many because it is always better than the stable one. Having a release note for an unreleased version might be confusing.

Beside, lets merge this.

PR #2592, opened 2019-01-14
beautiful.init: Fix return values and improve doco. fixes: #2588

Sorky:
beautiful.init return values incomplete and some invalid parameters not picked up #2588

Some behaviour that doesn't really seem right...

Doesn't fail on beautiful.init( my_config ) when my_config = a function
Doesn't fail on beautiful.init( my_config ) when my_config = a number

As far as I can tell `if theme then` only catches the case where the protected_call fails

Doesn't fail on beautiful.init( my_config ) when my_config = {}

Would also be nice if it returned true if there was no problem so as to do...
```
if not beautiful.init( path_this .. "themes/shelby/theme.lua" ) then
	beautiful.init( path_this .. "themes/shelby/.last.theme.lua" ) -- a proven config 
end
```
The above would be nice assuming a theme error would not cause a full reversion to the default rc.lua

PR #2625, opened 2019-01-28
Change the badge URL from travis-ci.org to travis-ci.com

Elv13:
I moved the CI from .org to .com to get updates and new features again.

### DIFF
--- a/docs/01-readme.md
+++ b/docs/01-readme.md
@@ -168,7 +168,7 @@ You can submit pull requests on the [github repository](https://github.com/aweso
 Please read the [contributing guide](https://github.com/awesomeWM/awesome/blob/master/docs/02-contributing.md) for any coding, documentation or patch guidelines.
 
 ## Status
-[![Build Status](https://travis-ci.org/awesomeWM/awesome.svg?branch=master)](https://travis-ci.org/awesomeWM/awesome)
+[![Build Status](https://travis-ci.com/awesomeWM/awesome.svg?branch=master)](https://travis-ci.com/awesomeWM/awesome)
 
 ## Documentation
 


Elv13:
Apparently it updated itself with the correct result after reporting -8%.

I also sent a support ticket their way about the broken links. Let see if I am more lucky than previous attempts.

Elv13:
> did you sent them a mail to migrate travis build?

Yes. That way the "Checks" tab in GitHub can be used to cancel or restart build without having to have extra privileges. At least if it worked. It also held up CodeCov from showing up there. Hopefully with some work we could get nice luacheck and busted reports directly in GitHub.

blueyed:
@Elv13 
@codecov is working (since months though) on the timeout with their links.. very annoying indeed.

I also noticed it posting a comment already while the build is still running on a private project, so that might be related to travis-ci.com (instead of .org), if it would not happen with .org now.

blueyed:
We would need to update the required status in https://github.com/awesomeWM/awesome/settings/branch_protection_rules/2336216 .

blueyed:
> Hopefully with some work we could get nice luacheck and busted reports directly in GitHub.

How so?

blueyed:
> We would need to update the required status in https://github.com/awesomeWM/awesome/settings/branch_protection_rules/2336216 .

Done.

Elv13:
> Done.

Thanks!

> How so?

One issue I see with onboarding new people is that they usually wont run the test locally and reading the "raw" output of Travis to find why the build failed is often a very confusing experience (especially when the error gets hidden from the normal log and only visible in the raw log). So I wondered if we could provide better reports. This could be done using @awesome-robot or using "github apps" and the "Checks" tab above. I migrated Travis to see how usable this feature is. 

> I also noticed it posting a comment already while the build is still running on a private project, so that might be related to travis-ci.com (instead of .org), if it would not happen with .org now.

Codecov migrated itself a few builds later. I didn't do it. It might be a feature. But as you have seen, it now displays the comment during the build instead of after it and the "you broke the coverage!" message is visible for ~5 minutes instead of a few seconds now. I guess this *is* a feature of the "Checks" system to have live results. However in this case it causes some extra confusion. Maybe we can batch the upload?

blueyed:
Re codecov: @psychon said on IRC that it was like this before already?!

Batching via Travis is not trivial (no assets etc), and AFAIK @codecov should check if the build is running still, before commenting.

Elv13:
> it was like this before already?!

Yes, ever since we started using Mergify it has been an issue. I restarted quite a few builds due to that. It eventually works, but has not been reliable, predictable or timely.

blueyed:
Hmm?  You restarted build, because Codecov posted a comment too early already?  Wasn't its status not updated later then?  (maybe only with Mergify?)

Elv13:
I was referring to codacy, sorry for the confusion.

Elv13:
While it has been a while since codecov send 2 emails in quick succession, it is new that it adds the commend so early in the build.

blueyed:
I would also say that this is new behavior, likely to the move to travis-ci.com then.
@psychon ?

blueyed:
btw: we could also disable comments altogether, at least for now maybe then?

blueyed:
Contacted codecov support: https://codecov.slack.com/archives/C04KHQ1C8/p1548763675020100

actionless:
i was asking by e-mail about codecov PR pages a bit earlier and they replied:


> Hi!
> When I open coverage details for a pull request (https://github.com/awesomeWM/awesome/pull/2592)
> the page with coverage details is not returned: https://codecov.io/gh/awesomeWM/awesome/compare/7e7733c6d18cee1d987896866a9fb57287b6780b...366be2105a44364c1e2c67897acae992e0ace840
> While it works well for specific commits:
> https://codecov.io/gh/awesomeWM/awesome/commit/7e7733c6d18cee1d987896866a9fb57287b6780b
> 
> All best, Actionless
> 
> 
> Ah, and it also not opens if I am trying to open PR details by different kind of link: https://codecov.io/gh/awesomeWM/awesome/pull/2592 (it waits for some time and returning empty response)


reply:

> Hi Actionless ,
> 
> 
> Sorry for the late reply, we had a problem in our support ticket inbox that has already been fixed and we’re currently working to solve your issue as soon as possible.
> 
> 
> Thanks for reaching out to Codecov Support. We are sorry to see that you are running into some issues. Please note that, with hundreds of thousands of developers using Codecov, we can only offer “best-effort” support to our open source / free users at this time.
> 
> 
> As possible, one of our support team members will review your support ticket and circle back with a resolution. Should you have any questions in the meantime, please do not hesitate to reach out. 
> 
> 
> In case you are on a paid support plan and this message was an error, please contact us back and someone from our team will get back to you as soon as possible.
> 
> 
> Very best,
> 
> 
> Customer Support Hero 
>

Elv13:
They sent me the same. Looks like everyone on the team sent a support
ticket in the last couple months

PR #2626, opened 2019-01-28
Use reproducible timestamps in more doc examples

psychon:
This adds require("_date") to some example tests that use the current
date via os.date. This allows reproducibility by replacing os.date()
with a function that uses a static date from $SOURCE_DATE_EPOCH. See
commit 9d7eaf0 for more details.

CC @reinerh Thanks for pointing out that this is indeed needed.

### DIFF
--- a/tests/examples/awful/popup/wiboxtypes.lua
+++ b/tests/examples/awful/popup/wiboxtypes.lua
@@ -1,6 +1,7 @@
 --DOC_GEN_IMAGE
 --DOC_HIDE_ALL
 --DOC_NO_USAGE
+require("_date")
 local awful     = require("awful")
 local gears     = require("gears")
 local wibox     = require("wibox")
--- a/tests/examples/awful/tooltip/textclock.lua
+++ b/tests/examples/awful/tooltip/textclock.lua
@@ -1,5 +1,6 @@
 --DOC_GEN_IMAGE
 --DOC_NO_USAGE
+require("_date") --DOC_HIDE
 screen[1]._resize {width = 300, height = 75} --DOC_HIDE
 local awful = {tooltip = require("awful.tooltip"), wibar = require("awful.wibar")} --DOC_HIDE
 local wibox = { widget = { textclock = require("wibox.widget.textclock") }, --DOC_HIDE
--- a/tests/examples/awful/tooltip/textclock2.lua
+++ b/tests/examples/awful/tooltip/textclock2.lua
@@ -1,5 +1,6 @@
 --DOC_GEN_IMAGE
 --DOC_NO_USAGE
+require("_date") --DOC_HIDE
 screen[1]._resize {width = 300, height = 75} --DOC_HIDE
 local awful = {tooltip = require("awful.tooltip"), wibar = require("awful.wibar")} --DOC_HIDE
 local wibox = { widget = { textclock = require("wibox.widget.textclock") }, --DOC_HIDE


tests/examples/awful/popup/wiboxtypes.lua
```diff
@@ -1,6 +1,7 @@
 --DOC_GEN_IMAGE
 --DOC_HIDE_ALL
 --DOC_NO_USAGE
+require("_date")
```
reinerh: `--DOC_HIDE` is not required in this case?

reinerh: Hm, because of `DOC_HIDE_ALL` above?

reinerh:
There was also a difference in AUTOGEN_awful_widget_layoutlist_bar.svg that didn't look time-related.
But after applying your change to tests/examples/awful/widget/layoutlist/wibar.lua from #2621 as well, this also built reproducibly.

tests/examples/awful/popup/wiboxtypes.lua
psychon: Yup.

The test examples are primarily meant to generate images for the docs. For this, `@DOC_awful_popup_wiboxtypes_EXAMPLE@` is replaced with something. By default, that's all the code used to generate the example and then the example image (or text, in some cases) itself. With `--DOC_HIDE` we can "hide some hacks" and `--DOC_HIDE_ALL` is used for cases where we only want the image (e.g. some overview of all the widgets).

psychon:
@reinerh Thanks for testing!

> There was also a difference in AUTOGEN_awful_widget_layoutlist_bar.svg that didn't look time-related.

Was it something like this? https://github.com/awesomeWM/apidoc/commit/ab926e95205c2c50cab7134f53dff1e9a989b665#diff-1c97ba2002b724e67f79d5225c015ce3 It seems that the IDs that cairo generates in the SVG have a random offset...?

In the end, the IDs that cairo uses some from here: https://gitlab.com/cairo/cairo/blob/ef8c379e0f560ced63c2a07566a3a7d608e3996c/src/cairo-surface.c#L264-285
This is just a counter that is incremented each time a cairo surface is allocated. So... something (no idea what, usual suspect is `pairs` iteration order) causes us to create cairo surfaces in different order / a different number of cairo surfaces. Yay, this will be lots of fun to debug.

psychon:
@Elv13 Mergify got broken by the Travis switch. The pull request status says `continuous-integration/travis-ci Expected — Waiting for status to be reported`, but also says `Travis CI - Pull Request Successful in 8m — Build Passed`.

blueyed:
@psychon 
Please consider creating an issue for the incrementing ID.  You can try/use the "Reference new issue" from your comment's actions (upper right).

blueyed:
Oh, it needs to be updated here: https://github.com/blueyed/awesome/blob/81a0a8cc48654f3670c377e11ce8ee07366c0de5/.mergify.yml#L6-L7 (will do, and also remove Codacy for now).

reinerh:
> > There was also a difference in AUTOGEN_awful_widget_layoutlist_bar.svg that didn't look time-related.
> 
> Was it something like this? [awesomeWM/apidoc@ab926e9#diff-1c97ba2002b724e67f79d5225c015ce3](https://github.com/awesomeWM/apidoc/commit/ab926e95205c2c50cab7134f53dff1e9a989b665#diff-1c97ba2002b724e67f79d5225c015ce3) It seems that the IDs that cairo generates in the SVG have a random offset...?

Yes, something like that. Now also visible in the [reproducible-builds CI](https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/diffoscope-results/awesome.html).

PR #2630, opened 2019-02-01
background container: Actually set the foreground color

psychon:
Commit ba75da7976e6 worked around a bug in LGI. However, it did so by
just dropping the code that set the foreground color. Instead, it should
have changed the code so that cr:set_source() is only called if the
background container has a foreground color configured instead of "just
always".

Fixes: https://github.com/awesomeWM/awesome/pull/2609#issuecomment-459580395

------

@actionless When you wrote the following, what did you actually mean? As far as I see, my "work around a bug in LGI"-commit just always broke the foreground.

> it broken background's `set_fg()` under some tricky usecase, 

------

Oh, by the way: Testing done by myself: None at all.

### DIFF
--- a/lib/wibox/container/background.lua
+++ b/lib/wibox/container/background.lua
@@ -55,6 +55,10 @@ function background:before_draw_children(context, cr, width, height)
         end
         cr:restore()
     end
+
+    if self._private.foreground then
+        cr:set_source(self._private.foreground)
+    end
 end
 
 -- Draw the border


actionless:
> what did you actually mean?

it seemed to me it was reproducing not in every of my scenarios, but looks like i had false-positive result (mb because of fg set in cairo markup in those widgets or so, i wasn't yet digging into it)

PR #2638, opened 2019-02-06
Add selection watcher objects

psychon:
This implements a tiny part of #492 by adding a new class to awesomeWM: Selection watcher.
A selection watcher watches some selection and emits a signal when the selection owner changes.

I put this into its own object, because watching for selection changes is completely independent from actually transferring selection contents (Copy&paste).

Simple example showing how to use this new API:
```lua
local watcher = selection_watcher("PRIMARY")
watcher.active = true
watcher:connect_signal("selection_changed", function(_, owned)
    if owned then
        print("PRIMARY selection changed")
    else
        print("PRIMARY selection is now unowned and thus guaranteed to be empty")
    end
end)
```
The plan is to also add separate classes for copy & paste to keep things simple in the C code.

Also, I did not add any API documentation on this new objects (yet). Mostly because I kind of expect @Elv13 to later remove them anyway and add some Lua code that simplifies the uglyness that the C API provides. Should I add LDoc documentation anyway?

Comments?

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -28,6 +28,7 @@ read_globals = {
     "mousegrabber",
     "root",
     "selection",
+    "selection_watcher",
     "tag",
     "window",
     "table.unpack",
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ install:
 
   # Install build dependencies.
   # See also `apt-cache showsrc awesome | grep -E '^(Version|Build-Depends)'`.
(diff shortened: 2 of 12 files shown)


psychon:
One well-placed `io.stdout:flush()` later: Yay, the test actually works now!

actionless:
> to later remove them anyway and add some Lua code that simplifies the uglyness that the C API provides

IMHO the current API as in example already user-friendly enough

Elv13:
> Also, I did not add any API documentation on this new objects (yet). Mostly because I kind of expect @Elv13 to later remove them anyway and add some Lua code that simplifies the uglyness that the C API provides. Should I add LDoc documentation anyway?

IMHO we should try to stop exposing raw C stuff, if it ever has to be changed, the compat code is hearth attack caliber ugly. See the PR that fixes `:keys()` and `:buttons()` for prime example of eye bleeding compat hacks.

actionless:
yeah, refactoring is a good point

and also i took a look on this together with the selection watcher -- altogether they doesn't look so userfriendly, so indeed some `clipboard` object would be useful

PR #2639, opened 2019-02-07
Add selection getter objects

psychon:
This fixes a large part of #492. Namely, this PR adds the "paste" part of "copy&paste". Also, this complements #2638 which is the PR that adds notifications for when some selection changes.

Simple usage example:
```lua
    local s = selection_getter("CLIPBOARD", "UTF8_TEXT")
    s:connect_signal("data", function(data) print("got data", data) end)
    s:connect_signal("data_end", function() print("data end") end)
```
The "data" signal is emitted for each piece of data that is received. After everything was received, "data_end" is emitted.

Simple example showing how to get the list of supported targets (in this case the argument is a table with the list of supported targets; everything else is (currently) a string):
```lua
    local s = selection_getter("CLIPBOARD", "TARGETS")
    s:connect_signal("data", function(d) gears.debug.dump(d) end)
    s:connect_signal("data_end", function() print("data end") end)
```
This has only received minimal testing so far...


Elv13:
/me is busy right now, will check this weekend, but thanks for this, great work

objects/selection_getter.c
```diff
@@ -0,0 +1,277 @@
...
+luaA_selection_getter_new(lua_State *L)
+{
+    size_t name_length, target_length;
+    const char *name, *target;
+    xcb_intern_atom_cookie_t cookies[2];
+    xcb_intern_atom_reply_t *reply;
+    selection_getter_t *selection;
+    xcb_atom_t name_atom, target_atom;
+
+    name = luaL_checklstring(L, 2, &name_length);
```
Elv13: Could this use an `args` table for consistency with the other constructor and future proofness?

psychon:
> With some way to set the selection

I thought a bit about how an API could look for setting the selection and got nowhere. This is complicated...

> and swap the `PRIMARY`, `SECONDARY` and `CLIPBOARD` buffers,

Uhm, I cannot see any special tricks that the C code could do so. As soon as Lua can set a selection, it can swap things. This is also why I ignored your request to have the selection content "stay in the server" and avoid some copying: I don't see how to do that with X11.

If you have some pointers like "$FOO can swap selection", I can look at how they do that.

(However, this is offtopic here and I guess a new issue about setting the selection should be opened.)

objects/selection_getter.c
psychon: Done. Note that this now produces quite useless error messages for type errors (e.g. talking about argument -2), but awesomeWM is quite good at that anyway, so whatever.

Elv13:
Thanks for the changes

blueyed:
@psychon 
Thanks for this.
Would it be possible to get information about the client that owns the clipboard/selection also?
(I am interested in the pid mainly, but that could be queried via `c.pid` then.)

psychon:
Uhm... nope, not really. We can query the selection owner, but that only gives us a window. I just tried with GVim and this is all the information that you can get from the selection owner info:
```
$ xprop -id 0x3800009
$
```
(Yes, there really is not a single property)

In theory, all of the things that are listed in ICCCM can be queried: https://tronche.com/gui/x/icccm/sec-2.html#s-2.6.2

For example, `NAME` and `HOST_NAME` can be queried as per ICCCM. However, GTK does not seem to automatically supply them, so they are in practice not supported. Same for Qt.

actionless:
mb lua code could iterate through all the windows and match the id?

blueyed:
IIRC this is just a special window only for the selection, isn't it?

psychon:
> IIRC this is just a special window only for the selection, isn't it?

Yup.

You could try matching for closest window ID, but that means that you end up with "just a random window" for `xclip`. Plus, when a X11 client runs out of the XID space that the X11 server allocated at connection time, the client can ask for a new space. Thus, I'm not even sure if "closest XID" would always work.

actionless:
btw, just very curious what is the usecase for that

blueyed:
@actionless 
https://github.com/awesomeWM/awesome-www/pull/111 - I guess it is quite out of date online, but currently I am basically spawning xclip for both the selection and clipboard before sending SIGSTOP, so that trying to use the clipboard/selection does not "freeze" when the process is stopped.
But that is sometimes still a bit wonky, I guess when the SIGSTOP happens before xclip really takes ownership etc, so I've thought knowing about the owner would help in that case.

PR #2643, opened 2019-02-11
Add API for setting clipboard contents

### LINKED ISSUE #2642: Extend the clipboard API part two: Setting the clipboard contents
I will pretend that #492 is about getting the selection and watching it change. Now, apparently @Elv13 also wants to set the clipboard's content. This issue is about that part.

Some complications that are required:

* Actually acquiring selection ownership can fail (see ICCCM § 2.1)
* INCR transfers (= large pieces of data have to be transferred in incremental chunks; most of this should be abstracted away by the C code)
* MULTIPLE target (= for efficiency reasons (???), it is possible to request multiple targets at once; I hope I can abstract this away in the C code, but it will be ugly and impossible to test, because even though this is required to be supported, nothing actually uses it)

Some complications that someone might ask for, some day:

* Be able to "give up" selection ownership
* Work fine with external clipboard managers? This means that `SAVE_TARGETS` might be needed when a selection is given up
* Should it be possible to abort current transfers? E.g. for timeouts?

Is anything from the list above required now?

The "real meat":

* How should transfers be done? Multiple transfers can be active at the same time etc
* Requests can be rejected

-----

First attempt at an API:
```lua
local o = selection_acquire("CLIPBOARD") -- Either returns an object, or nil if acquiring fails
o:connect_signal("request", function(o, target, t) -- t is a "selection transfer" for this specific request
  if target == "TARGETS" then
    t:send{ data = { "TARGETS", "UTF8_STRING" }, format = "atom" } -- Send an ATOM list
    return
  end
  if target ~= "UTF8_STRING" then
    -- Do nothing here to reject the request
  elseif math.random(2) == 1
    -- Do an incremental transfer
    -- continue is either true (to just indicate an incremental transfer) or a lower bound on the length of the answer
    t:send{ data = "This is the ", continue = 10 }
    t:connect_signal("continue", function(t)
      t:send{ data = "response", continue = false } -- indicate this is the end of the transfer
    end
  else
    -- Just answer directly
    t:send{ data = "This is the response" }
  end
end)
```

psychon:
This fixes #2642. (although I was too lazy to implement `MULTIPLE` and will just wait for someone to request this; even though the spec requires support for it, I still doubt it is used much. We will see...)

Example code that I used for testing:
```lua
    local o1 = selection_acquire("PRIMARY")
    o1:connect_signal("release", function() print("o1 released") end)
    o1:connect_signal("request", function(_, target, t)
        if target == "TARGETS" then
            t:send{ data = { "TARGETS", "UTF8_STRING" }, format = "atom" }
        else
            local data = {}
            for i=1, 2^23 do
                table.insert(data, "Hi")
            end
            print(#data)
            local data = table.concat(data)
            print(#data)
            print(string.format("data has length %d == 0x%x, %d bytes of space left", #data, #data, 16777212-#data))
            t:send{ data = data, continue = true }

            local count = 0
            t:connect_signal("continue", function()
                print("got continue signal");
                count = count + 1
                if count == 0 then
                    t:send{ data = ". Thanks for listening.", continue = true }
                elseif count == 1 then
                    t:send{ data = "", continue = true }
                elseif count == 2 then
                    t:send{ data = "\n" }
                end
            end)
        end
    end)
```

Fun fact: I looked at the GTK API after writing this and the API that I came up with is surprisingly close to GTK's API. Isn't GTK's API known for being hard to use...?

### DIFF
--- a/objects/selection_acquire.c
+++ b/objects/selection_acquire.c
@@ -1,0 +1,29 @@
+/*
+ * selection_acquire.c - objects for selection ownership
+ *
+ * Copyright © 2019 Uli Schlachter <psychon@znc.in>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+
+#include "objects/selection_acquire.h"
+
+void
+selection_acquire_class_setup(lua_State *L)
+{
+}
+
+// vim: filetype=c:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/objects/selection_acquire.c
+++ b/objects/selection_acquire.c
@@ -28,16 +28,84 @@
 typedef struct selection_acquire_t
 {
     LUA_OBJECT_HEADER
+    /** The selection that is being owned. */
+    xcb_atom_t selection;
     /** Window used for owning the selection. */
     xcb_window_t window;
     /** Timestamp used for acquiring the selection. */
     xcb_timestamp_t timestamp;
-    /** Reference in the special table to this object. */
-    int ref;
 } selection_acquire_t;
 
 static lua_class_t selection_acquire_class;
 LUA_OBJECT_FUNCS(selection_acquire_class, selection_acquire_t, selection_acquire)
+
+static void
+luaA_pushatom(lua_State *L, xcb_atom_t atom)
+{
+    lua_pushnumber(L, atom);
+}
+
+static int
+selection_acquire_find_by_window(lua_State *L, xcb_window_t window)
+{
+    /* Iterate over all active selection acquire objects */
+    lua_pushliteral(L, REGISTRY_ACQUIRE_TABLE_INDEX);
+    lua_rawget(L, LUA_REGISTRYINDEX);
+    lua_pushnil(L);
+    while (lua_next(L, -2) != 0) {
+        if (lua_type(L, -1) == LUA_TUSERDATA) {
+            selection_acquire_t *selection = lua_touserdata(L, -1);
+            if (selection->window == window)
+            {
+                /* Remove table and key */
+                lua_remove(L, -2);
+                lua_remove(L, -2);
+                return 1;
+            }
+        }
+        /* Remove the value, leaving only the key */
+        lua_pop(L, 1);
+    }
+    /* Remove the table */
+    lua_pop(L, 1);
+
+    return 0;
+}
+
+static void
+selection_release(lua_State *L, int ud)
+{
+    selection_acquire_t *selection = luaA_checkudata(L, ud, &selection_acquire_class);
+
+    luaA_object_emit_signal(L, ud, "release", 0);
+
+    /* Destroy the window, this also releases the selection in X11 */
+    xcb_destroy_window(globalconf.connection, selection->window);
+    selection->window = XCB_NONE;
+
+    /* Unreference the object, it's now dead */
+    lua_pushliteral(L, REGISTRY_ACQUIRE_TABLE_INDEX);
+    lua_rawget(L, LUA_REGISTRYINDEX);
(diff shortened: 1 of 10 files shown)


objects/selection_acquire.c
```diff
@@ -0,0 +1,240 @@
...
+luaA_selection_acquire_new(lua_State *L)
+{
+    size_t name_length;
+    const char *name;
+    xcb_intern_atom_reply_t *reply;
+    xcb_get_selection_owner_reply_t *selection_reply;
+    xcb_atom_t name_atom;
+    selection_acquire_t *selection;
+
+    name = luaL_checklstring(L, 2, &name_length);
```
Elv13: For consistency and futureproofness considerations, can we have an `args` table even if there's (currently) only a single argument?

Elv13:
Thanks! There's a Codacy warning, but beside :+1:

psychon:
> The scope of the variable 'data' can be reduced.

Aha... -.-
Updated to fix that and that dismissed your review. No force-push, I just added another commit to this PR.

tests/test-selection-transfer.lua
```diff
@@ -0,0 +1,235 @@
...
+                continue = true
+            end })
+        return true
+    end,
+
+    wait_a_bit,
+    wait_a_bit,
+    wait_a_bit,
+    wait_a_bit,
+    wait_a_bit,
```
blueyed: Why are those `wait_a_bit` calls necessary?  Could have a comment maybe, if necessary.

psychon: Done in a new commit.

Because transferring 100 MiB of data (inefficiently) is slow. I had a timeout without `wait_a_bit` (which means that it took longer than 0.5 seconds, I think). I just added a couple of these calls because others might have slower computers than I do.

The previous step tests large data transfer. The goal here is to have a single piece of data that is so large that the C code has to split it up into pieces, too. Thus, a single `:send` turns into multiple `INCR` steps for the C code. This also tests continuation, i.e. Lua adding more data later. If my math is correct, this almost transfers 100 MiB of data through the X11 server.

The maximum request size of my current X11 server is a bit below 16 MiB. Thus, each of the 32 MiB pieces that Lua gives to C has to split up into three increments, two almost of full size and one with just 8 bytes of data. :-)

Elv13:
Last call. If there is no other reviews I will merge it manually tomorrow. This is blocking two pull requests.

@blueyed Can you draft the public API you want for this? I think I get what kind of use cases you are after, but if you can actually document them, it would be useful. So far you mentioned:

 * Using rules like awful.rules to define when to transfer ownership
 * Having some ways to running these rules both when the seletion changes and "at any time" like before SIGSTOPping clients

And me:

 * Being able to store and keep "n" selection from the history for some payload types
 * Have an awful.widget.common widget to list them
 * Being able to restore a recent selection from history
 * Move a selection from the PRIMARY and SELECTION buffers using a keybinding

This leaves the problem of knowing which client the selection comes from (if any, ei. `xsel`) unsolved. But it isn't really important for my use case and apparently isn't possible to implement in a reliable way (like assuming the focused client is the source, traveling the winID tree until WM_NAME is set or using some LD_PRELOAD magic to set an extra xproperty on it when a client uses `awful.spawn` to be launched)

objects/selection_acquire.c
```diff
@@ -0,0 +1,243 @@
...
+    lua_rawget(L, LUA_REGISTRYINDEX);
+    lua_pushnil(L);
+    while (lua_next(L, -2) != 0) {
+        if (lua_type(L, -1) == LUA_TUSERDATA) {
+            selection_acquire_t *selection = lua_touserdata(L, -1);
+            if (selection->window == window)
+            {
+                /* Remove table and key */
+                lua_remove(L, -2);
+                lua_remove(L, -2);
```
blueyed: I am confused by `lua_remove` using `-2` here twice.. would have expected it to be -2 and -1.

blueyed:
> Last call. If there is no other reviews I will merge it manually tomorrow. This is blocking two pull requests.

Left a hopefully minor comment.
In general I cannot say to much about it, feel free to merge after confirming that the `-2` is correct.

blueyed:
> @blueyed Can you draft the public API you want for this? I think I get what kind of use cases you are after, but if you can actually document them, it would be useful. So far you mentioned:
> 
> * Using rules like awful.rules to define when to transfer ownership
> * Having some ways to running these rules both when the seletion changes and "at any time" like before SIGSTOPping clients

I thought about using it instead of `xclip` to transfer ownership before SIGSTOPping a client, and thought it might be good to see if it needs to be transfered (owner).

objects/selection_acquire.c
psychon: Well... the stack at this point (from bottom to top):

- The table that we are iterating over (Index -3)
- The current key in the table (Index -2)
- The current value in the table (Index -1)

This wants to get rid of the table and the key, but wants to preserve the value. Thus, the elements at stack index -2 and -3 should be removed. The code here first removes the key, so afterwards the table is at index -2, i.e. below the value.

Does this make it clearer?

blueyed: Yes, thanks!

PR #2646, opened 2019-02-14
Fix a CMake warning when GENERATE_DOC is off

psychon:
This fixes the following CMake warning that I saw in some NixOS build:

CMake Warning (dev) at CMakeLists.txt:457 (add_dependencies):
  Policy CMP0046 is not set: Error on non-existent dependency in
  add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.
  The dependency target "check-examples" of target "check" does not exist.
This warning is for project developers.  Use -Wno-dev to suppress it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -456,7 +456,10 @@ if(LUACHECK_EXECUTABLE)
     list(APPEND CHECK_QA_TARGETS luacheck)
 endif()
 add_custom_target(check-qa DEPENDS ${CHECK_QA_TARGETS})
-add_dependencies(check check-qa check-examples)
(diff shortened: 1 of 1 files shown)


blueyed:
Note that the include of `tests/examples/CMakeLists.txt` is unconditional with https://github.com/awesomeWM/awesome/pull/2545 (and therefore this would not be necessary / get reverted with it).

blueyed:
Since https://github.com/awesomeWM/awesome/pull/2545 has been merged this should not be necessary anymore.

PR #2648, opened 2019-02-14
ci: codecov: use env vars instead of flags

blueyed:
Using flags appears to trigger timeouts with codecov's backend.
This uses env variables instead.

PR #2649, opened 2019-02-14
ci: codecov: use env vars instead of flags

blueyed:
Using flags appears to trigger timeouts with codecov's backend.
This uses env variables instead.


blueyed:
Was a test to debug Travis checks issue (https://github.com/awesomeWM/awesome/pull/2648).

PR #2650, opened 2019-02-14
No-change to test Travis checks

blueyed:
Test/debug duplicate Travis checks (as seen in https://github.com/awesomeWM/awesome/pull/2648).


blueyed:
Happened once when force-pushing the same branch from #2648, but not again later.
Appears to be something flaky with Travis.

PR #2564, opened 2019-01-09
Code Climate integration

blueyed:

PR #2651, opened 2019-02-14
ci: codecov: improve flags/env

blueyed:
Wasn't showing up at https://codecov.io/gh/awesomeWM/awesome/commit/df9ed76418e54407d8d101b257d61fecd595b9ce/build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -116,7 +116,7 @@ install:
       if [ "$DO_COVERAGE" = "codecov" ]; then
         test -f build/luacov.stats.out || { echo 'build/luacov.stats.out does not exist.'; return 1; }
         luacov || return 1
-        travis_retry bash /tmp/codecov-bash -f build/luacov.report.out -X gcov -X coveragepy -F luacov -e LUANAME,LGIVER || return 1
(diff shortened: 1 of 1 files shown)


blueyed:
Env vars do not show up with the files though (only the job can be selected there).
So I think we should re-add the flags "lua_code" and "c_code".

psychon:
> So I think we should re-add the flags "lua_code" and "c_code".

I may be missing something but... doesn't the file name already provide this information? If it ends with `.lua`, then it is lua code. If it ends with `.c` or `.h`, then it is C code.

blueyed:
@psychon 
Yeah, "c_code" is not really good - that's why I've used "gcov" now for what was "c_code" before.
And all the others are "luacov" - so we have only two flags now, saying where the coverage is coming from.

blueyed:
https://codecov.io/gh/awesomeWM/awesome/compare/df9ed76418e54407d8d101b257d61fecd595b9ce...d1b7f8993fa88faf294acca3dd2dfe0d62e3fa4d times out though again.. :/

blueyed:
Now it worked (did not timeout).

blueyed:
btw: we only have a single job reporting to codecov after all (with regard to duplicate comments etc).

Might pick up #2564 soon again, and then we could upload more often there also, which likely would also help to reduce the flaky coverage.

PR #2656, opened 2019-02-15
Add gears.delayed_call and gears.delayed_call.run_now()

psychon:
Here is the non-broken part of #2655 and a beginning of getting rid of the `refresh` signal.

### DIFF
--- a/lib/gears/delayed_call.lua
+++ b/lib/gears/delayed_call.lua
@@ -15,12 +15,17 @@ local protected_call = require("gears.protected_call")
 local module = {}
 
 local delayed_calls = {}
-capi.awesome.connect_signal("refresh", function()
+
+--- Run all pending delayed calls now. This function should best not be used at
+-- all, because it means that less batching happens and the delayed calls run
+-- prematurely.
+-- @function gears.delayed_call.run_now
+function module.run_now()
     for _, callback in ipairs(delayed_calls) do
         protected_call(unpack(callback))
     end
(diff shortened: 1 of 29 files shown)


lib/gears/delayed_call.lua
```diff
@@ -0,0 +1,43 @@
+---------------------------------------------------------------------------
```
Elv13: I am all for smaller, more dedicated *files*, but on the other hand, small *modules* have a big problem. This problem isn't technical, it's human.

Back when I went through years or IRC, Arch forum posts, StackOverflow and Reddit posts to build #1373, one thing was clear: we had too many modules in the doc. The downside was that users could not find what they were looking for. It's not a problem for both of us because we have the whole API "in memory", but it is a major problem for new users. To mitigate that, I cut the number of non-widget documented modules in half, which is still not enough. My next move, as mentioned a couple time, will be to try to split the widget into their own ldoc instance and cross-reference them. But this is off topic.

So lets please avoid **documented modules** (the ones in the sidebar index) with a single or couple functions because people wont find them. This sucks since from the point of view of separations of concern and GRASP patterns philosophy: it should be split. Thankfully, we already have a way to have small files and few modules thanks to `ldoc` `@submodule gears.timer`. This way it will get its own little section in the timer documentation and it wont even require deprecating anything. Plus, it allows us to keep consolidating topics (such as in this case, timing related APIs) into a single page.

PR #2661, opened 2019-02-16
Change the selection global

psychon:
3180ae878cc3106ed2a15b187e20c6df9303310a is the main commit:

> This moves selection_acquire to selection.acquire, selection_getter to
> selection.getter and selection_watcher to selection.watcher. The only
> user-visible change due to this is that type(selection) is now "table"
> instead of "function". If that breaks something for you, tough luck.

Additionally, this deprecates `selection()`.

Finally, this is WIP, because it contains a squash of #2660 and #2643. So, this only depends upon these two PRs to be merged and me to rebase this one afterwards.


psychon:
Rebased this on master. This should be ready for review now.

PR #2665, opened 2019-02-17
Use xcb-errors library if it is available

psychon:
This library allows to get a human-readable string describing X11
requests, events, and errors. We now use this library to pretty-print
X11 errors if we get any.

To test this code, I added the following two lines to AwesomeWM so that
X11 errors are generated:

    xcb_set_input_focus(globalconf.connection, 42, 42, 42);
    xcb_randr_set_output_primary(globalconf.connection,
        globalconf.screen->root, 42);

Output without xcb-errors:

    X error: request=SetInputFocus (major 42, minor 0), error=BadValue (2)
    X error: request=(null) (major 140, minor 30), error=(null) (147)

Output with xcb-errors:

    X error: request=SetInputFocus (major 42, minor 0), error=Value (2)
    X error: request=RandR-SetOutputPrimary (major 140, minor 30), error=RandR-BadOutput (147)

----

One still open question

- [x] Should (one of the) travis build(s) install xcb-errors to test compilation of this?


actionless:
>     * Should (one of the) travis build(s) install xcb-errors to test compilation of this?

i think yes

awesomeConfig.cmake
```diff
@@ -218,6 +219,17 @@ if(WITH_DBUS)
...
+if(WITH_XCB_ERRORS)
+    pkg_check_modules(XCB_ERRORS xcb-errors)
+    if(XCB_ERRORS_FOUND)
+        set(AWESOME_OPTIONAL_LDFLAGS ${AWESOME_OPTIONAL_LDFLAGS} ${XCB_ERRORS_LDFLAGS})
+        set(AWESOME_OPTIONAL_INCLUDE_DIRS ${AWESOME_OPTIONAL_INCLUDE_DIRS} ${XCB_ERRORS_INCLUDE_DIRS})
+    else()
+        set(WITH_XCB_ERRORS OFF)
+        message(STATUS "xcb-errors not found. Disabled.")
+    endif()
+endif()
```
blueyed: Could this default to AUTO, and cause an error if ON was used, but it is not available?

psychon: There is no `AUTO` for `option()`. It is a `TRUE` / `FALSE` thing. I briefly looked into it (it is possible, just not with `option()`) and then decided to just copy what we do for DBus, since apparently that was good enough so far.

If you disagree, I can look into implementing `AUTO` for `WITH_DBUS`, `GENERATE_MANPAGES`, `GENERATE_DOC`, and anything else that needs this. But I'd do that in a separate PR.

blueyed: Sure, just a quick idea.
I think it would be nice in general, but not really necessary.

.travis.yml
```diff
@@ -52,6 +52,16 @@ install:
...
      make && sudo make install)
 
+  # Install xcb-errors if needed
+  - |
+    set -e
+    if [[ "$WITH_XCB_ERRORS" == "yes" ]]; then
+      git clone --depth 1 --recursive https://gitlab.freedesktop.org/xorg/lib/libxcb-errors.git /tmp/xcb-errors
+      (cd /tmp/xcb-errors
+       ./autogen.sh --prefix=/usr
+       make && sudo make install)
```
blueyed: Also needs xcb-proto it seems:
```
checking for XCBPROTO... no
configure: error: Package requirements (xcb-proto) were not met:
No package 'xcb-proto' found
```

blueyed: Maybe APT addons can be used?  https://docs.travis-ci.com/user/installing-dependencies/  (this can be configured for single jobs)

psychon: Heh, seems like we were looking at this simultaneously. :-)
I just added xcb-proto to our big list of packages to install.

blueyed: I would prefer to have it only with / nearby where it is needed.

psychon: Sorry, what do you mean? I should move the installation of `xcb-proto` to a separate invocation of `apt-get install`?

blueyed: Yes.

blueyed: Then it is easier in the future to see what is required because of what (and it is a bit less overhead , and more explicit).

psychon: Yup, you are right.

blueyed:
Let's have this (rebased) after https://github.com/awesomeWM/awesome/pull/2671.

blueyed:
Should also come after https://github.com/awesomeWM/awesome/pull/2682.

actionless:
@psychon so #2682 got merged

psychon:
Rebased and updated for those two other PRs.

common/version.c
```diff
@@ -54,6 +54,11 @@ eprint_version(void)
 #else
     const char *has_dbus = "✘";
 #endif
+#ifdef WITH_XCB_ERRORS
+    const char *has_xcb_errors = "✔";
+#else
+    const char *has_xcb_errors = "✘";
```
actionless: btw that's obviously out of scope of this PR but since i noticed this line here it reminded me a thing:

on some systems i've noticed the font didn't had those characters, so it was just a square or empty space

and sometimes they're just a bit shifted in position (if font doesn't have them and using from fallback font which have incompatible dimensions of characters)

how do you think is it worth creating an issue to discuss/address it or it could just waste time on bikeshedding?

psychon: Well, if you have a suggestion for what to do instead, try a PR. That would avoid bikeshedding. I do not have any ideas right now.

actionless: https://github.com/awesomeWM/awesome/pull/2687/files

.travis.yml
```diff
@@ -65,13 +65,15 @@ jobs:
             - liblua5.3-dev
             - lua5.3
     # Note: luarocks does not work with Lua 5.0.
-    - env: LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build
+    - env: LUA=5.1 LUANAME=lua5.1 BUILD_IN_DIR=/tmp/awesome-build WITH_XCB_ERRORS=yes
```
blueyed: Sorry for noticing earlier, but it would be better to add this to the `DO_COVERAGE=codecov` job.

blueyed: Went ahead and amended your last commit.

actionless:
@blueyed it have a conflict again

actionless:
@psychon and again :-)

psychon:
Rebased on master

PR #2668, opened 2019-02-18
Fix extra newline with legacy naughty layout

blueyed:
Fixes https://github.com/awesomeWM/awesome/commit/9df77e5c765d1b523babcdbc6a45e6cbafebcc21#commitcomment-32354198.

TODO:

- [ ] test - somewhere along https://github.com/blueyed/awesome/blob/8072d717cfb6ccd3254875603b6d1efe20789be3/tests/test-naughty-legacy.lua#L750-L751 likely


Elv13:
thanks for testing this early

Elv13:
Edit `test-naughty-legacy.lua` and check `n.textbox` where `n` is the notification. The test suite has 3 sections. The top is the "normal" stuff, the second part is optional when and the third part checks the deprecated API. This test would go in the step where it tests replacing the message and title.

PR #2671, opened 2019-02-18
CMake: Add "AUTO" state for WITH_DBUS and GENERATE_MANPAGES

psychon:
Previously, these options had two possible values: ON and OFF.

Now, they have three possible values: ON, OFF, and AUTO.

OFF still does what it always did: The feature is just disabled.

With ON and AUTO, we check for the feature. The difference is what
happens when the feature could not be enabled, e.g. because some
dependencies is missing: With AUTO, we just disable the feature (this is
what happened previously with ON). However, with ON, CMake aborts and
prints an error.

Implements: Suggestion by Daniel
    https://github.com/awesomeWM/awesome/pull/2665#pullrequestreview-204595138

### DIFF
--- a/AutoOption.cmake
+++ b/AutoOption.cmake
@@ -1,0 +1,42 @@
+# Add a new option with default value "auto":
+#   autoOption(FOOBAR "Support foobar")
+#
+# To check if FOOBAR support should be enabled, use:
+#   if(FOOBAR)
+#
+# If necessary dependencies for FOOBAR are missing, use:
+#   autoDisable(FOOBAR "Did not find baz")
+#
+# Example:
+#   autoOption(FOOBAR "Support foobar")
+#   if(FOOBAR)
+#     Check for FOOBAR here.
+#     if(NOT BAZ_FOUND)
+#       autoDisable(FOOBAR "Did not find baz")
+#     endif()
+#   endif()
+
+function(autoOption name description)
+    set(${name} AUTO CACHE STRING "${description}")
+    set_property(CACHE ${name} PROPERTY STRINGS AUTO ON OFF)
+
+    if((NOT ${name} STREQUAL ON) AND
+        (NOT ${name} STREQUAL OFF) AND
+        (NOT ${name} STREQUAL AUTO))
+        message(FATAL_ERROR "Value of ${name} must be one of ON/OFF/AUTO, but is ${${name}}")
+    endif()
+endfunction()
+
+function(autoDisable name reason)
+    message(STATUS "${reason}")
+    if(${name} STREQUAL AUTO)
+        message(STATUS "Disabled.")
+        SET(${name} OFF PARENT_SCOPE)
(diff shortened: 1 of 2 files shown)


AutoOption.cmake
```diff
@@ -0,0 +1,45 @@
...
+#  autoOption(FOOBAR "Support foobar")
+#
+# To check if FOOBAR support should be enabled, use:
+#   if(FOOBAR)
+#
+# If necessary dependencies for FOOBAR are missing, use:
+#   autoDisable(FOOBAR "Did not find baz")
+#
+# Example:
+#  autoOption(FOOBAR "Support foobar")
```
blueyed:
```suggestion
#   autoOption(FOOBAR "Support foobar")
```

AutoOption.cmake
```diff
@@ -0,0 +1,45 @@
...
+#   if(FOOBAR)
+#     Check for FOOBAR here.
+#     if(NOT BAZ_FOUND)
+#       autoDisable(FOOBAR "Did not find baz")
+#     endif()
+#   endif()
+#
+# If necessary dependencies for FOOBAR are missing, use:
```
blueyed:
```suggestion
```

AutoOption.cmake
```diff
@@ -0,0 +1,45 @@
...
+#   autoDisable(FOOBAR "Did not find baz")
```
blueyed:
```suggestion
```

AutoOption.cmake
```diff
@@ -0,0 +1,45 @@
...
+
+    if((NOT ${name} STREQUAL ON) AND
+        (NOT ${name} STREQUAL OFF) AND
+        (NOT ${name} STREQUAL AUTO))
+        message(FATAL_ERROR "Value of ${name} must be one of ON/OFF/AUTO, but is ${${name}}")
+    endif()
+endfunction()
+
+function(autoDisable name reason)
+    message(STATUS "${reason}")
```
blueyed: Should `name` be included in the status message?

AutoOption.cmake
psychon: Huh? These two lines should just be removed?

AutoOption.cmake
psychon: The two calls to this function have message `DBus not found.` and `Not generating manpages. Missing: [stuff]`. That seems quite self-explanatory to me and adding `WITH_DBUS` or `GENERATE_MANPAGES` in there won't really improve things, I think.

AutoOption.cmake
blueyed: Yes.. the same are above.

AutoOption.cmake
blueyed: Ok.

AutoOption.cmake
psychon: Oh, thanks, I wonder why I didn't notice this duplication...

PR #2673, opened 2019-02-18
beatiful.theme_assets: recolor titlebar icons for `active` and `inactive` states

alfunx:
Recolor all titlebar icons, `active` ones only, or `inactive` ones only.

### DIFF
--- a/lib/beautiful/theme_assets.lua
+++ b/lib/beautiful/theme_assets.lua
@@ -213,21 +213,22 @@ end
 -- @tparam color color Icons' color.
 -- @tparam string state `"normal"` or `"focus"`.
 -- @tparam string postfix `nil`, `"hover"` or `"press"`.
+-- @tparam string toggle_state `nil`, `"active"` or `"inactive"`.
 -- @treturn table Beautiful theme table with the images recolored.
-function theme_assets.recolor_titlebar(theme, color, state, postfix)
+function theme_assets.recolor_titlebar(theme, color, state, postfix, toggle_state)
     if postfix then postfix='_'..postfix end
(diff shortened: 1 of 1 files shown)


lib/beautiful/theme_assets.lua
```diff
@@ -228,13 +230,15 @@ function theme_assets.recolor_titlebar(theme, color, state, postfix)
         'titlebar_maximized_button_'..state..'_inactive',
         'titlebar_maximized_button_'..state..'_active',
```
actionless: this whole block looks weird now after you added active/inactive as parameters

alfunx: Yes indeed, the problem was that close/minimize doesn't have the active/inactive variants (and I didn't want to change too much). With the current patch, we just filter the non-matching entries. Alternatively that table could be generated, instead of being "hardcoded" (and filtered). What do you think?

actionless: >  Alternatively that table could be generated, instead of being "hardcoded" (and filtered). What do you think?

yup, that's what i was implying about

however, as you yourself mentioned:

>  the problem was that close/minimize doesn't have the active/inactive variants

so if you or @Elv13 or somebody else have some nicer solution in mind, i don't want to nitpicking too much and the current code would be fine

alfunx: Added a fixup commit with the table being generated. Not sure if this is overkill...

alfunx:
@actionless Should I squash this?

actionless:
yup, our mergify didn't learned that trick yet :-)

alfunx:
Okay, done :)

Issue #1093, opened 2016-09-14
Switch to some external, general DBus bindings


psychon:
Currently awesome provides some built-in DBus bindings. That's bad. There is no reason why a WM should provide bindings to DBus and this makes it hard to reuse code outside of awesome. (Plus, our bindings are *bad*).

We should use some already-existing DBus bindings instead. They should provide:
- Support for asynchronous operations
- Integrate with GLib's MainLoop so that it automatically works in awesome
- [Add your requirements here; I don't have much clue about DBus]

Existing solutions are:
- lgi's Gio / GDBus bindings (supposedly crashy, but that should be fixable)
- https://github.com/daurnimator/ldbus (Only synchronous operations are possible, doesn't seem to have GLib MainLoop integration)
- https://github.com/dodo/lua-dbus (I'm not sure what this actually is; seems to be based on either ldbus or awesome's dbus bindings)
- https://github.com/xs-embedded-llc/l2dbus (Looks interesting)
- https://github.com/Elv13/wirefu (As I understand it: High-level interface based on lgi's low-level interface; Supposedly crash, but I could never reproduce anything and those should be fixable)
- [Add more solutions here]

psychon:
Just so that I do not have to look up lots of details again, some demo program with LGI (completely Async AFAIK):
```lua
local lgi = require("lgi")
local Gio = lgi.require("Gio")
local GLib = lgi.require("GLib")

-- Workaround for https://github.com/pavouk/lgi/issues/142
local function bus_get_async(type)
	Gio.bus_get(type, nil, coroutine.running())
	local a, b = coroutine.yield()
	return Gio.bus_get_finish(b)
end

local function method_call(bus)
	local name = "org.freedesktop.systemd1"
	local object = "/org/freedesktop/systemd1"
	local interface = "org.freedesktop.DBus.Properties"
	local result, err = bus:async_call(name, object, interface, "GetAll",
		GLib.Variant("(s)", { "org.freedesktop.systemd1.Manager" }), -- Arguments
		nil, Gio.DBusCallFlags.NONE,
		-1) -- Use default timeout
	if not result then
		return nil, tostring(err)
	end
	print(result:get_type_string())

	-- Get (only?!?) entry in the array
	local res = result:get_child_value(0)
	-- Get entries from the dictionary
	for i = 0, res:n_children() - 1 do
		local value = res:get_child_value(i)
		local key, val = value:get_child_value(0), value:get_child_value(1)
[...]

PR #2679, opened 2019-02-19
WIP: Port our DBus code to Gio

psychon:
This is basically untested. Especially the signals seem to be broken in
naughty and thus were not tested at all.

This might be a start to fixing #1093.


psychon:
Of course the signal emission was broken. :-)

Anyway, this now crashes. When LGI frees a reference. That it should not be owning any more, I guess.

The "method call handler" gets an object of type `GDBusMethodInvocation`. On this object, things like `return_value` are called to answer the method invocation. This method is annotated with `[transfer full]`, i.e. it takes ownership of the `GDBusMethodInvocation` and Lua loses ownership. However, LGI still tries to free this later on and we get a crash in LGI's `object_unref()`.
(Debugged by: `printf("%p\n", obj);` in `object_unref()` and lots of `print`s in awesome's Lua code to see the pointer to the object when it is still alive. I'm not sure what is going wrong here.)

psychon:
The crash turns out to be due to human error. I was replying twice to a DBus invocation, which caused some reference count to be decremented twice. That's what happens when one replaces a `return` with a method call without thinking enough about the resulting control flow change...

psychon:
The new error is:
```
Error org.freedesktop.DBus.Error.InvalidArgs: Type of message, “(susssasa{ss}i)”, does not match expected type “(susssasa{sv}i)”
```
This seems to be a bug in the test. It sends wrong types and GDBus catches this and does not accept the method call. However, this is not fixable with `dbus-send`. From `man dbus-send` (we need a dict with string keys and variant values here):
```
D-Bus supports more types than these, but dbus-send currently does not. Also, dbus-send does not permit empty containers or nested containers (e.g. arrays of variants).
```
@Elv13 Would you be okay with me porting the new test from spawning `dbus-send` to just using `GDBusConnection` directly through LGI?

(Also: What does this mean for our switch away from our bad, homegrown DBus bindings? Do we just wait for "people out there" to report that this stricter testing broke something for them and then tell them "sorry, but fix your stuff")

Elv13:
> @Elv13 Would you be okay with me porting the new test from spawning dbus-send to just using GDBusConnection directly through LGI?

Sure, but having notify-send for the normal one is, I think, a good test in itself.

lib/awful/remote.lua
```diff
@@ -7,43 +7,90 @@
...
+    local code = parameters.value[1]
+
+    local f, e = load(code)
+    if not f then
+        invocation:return_dbus_error("org.awesomewm.awful.Remote.InvalidCode", e)
+        return
+    end
+    local results = { pcall(f) }
+    if not table.remove(results, 1) then
+        invocation:return_dbus_error("org.awesomewm.awful.Remote.ExecutionError", results[1])
```
psychon: Another case of "missing `return`".

Elv13:
What's the status of this? I would like to have this when I merge the "view" part of the notification API. With this, it is easier to get the sender information. This is the "only" reliable way (the `desktop-entry` most often doesn't have the `class` and may have more than 1 instance) to get the client. The specs talk about having the winId in the hints, but it is nowhere to be found in `libnotify` and `knotify`. The PID hint was deprecated in favor of the dbus sender info and is not implemented by most providers.

psychon:
> What's the status of this? 

Well... would this be suitable to have in a minor release? I thought this is something that would have to wait for v5.

Anyway, the status is: I managed to work around all occurring problems. So far, everything works fine. Only known issues are
- missing return that I only saw here and didn't actually fix yet
- the code is now stricter about the types it receives and stuff outside of AwesomeWM could break (which I guess we just accept)

One small todo list item is to figure out precisely the lifetime requirements.
- The `reference_count` fields are automatically initialised to 1. I have an idea for where this magic happens, but I want to make sure this is fine.
- Do structs keeping pointers to other structs keep these things alive or do we have to "anchor everything" in Lua? Based on this info, a module in gears could make this API easier to work with.

> I would like to have this [for...]

I don't really understand what you are talking about here, but okay, whatever.

psychon:
> What's the status of this?

#2722 does the naughty part of this in a nicer way. That should not "really" break anything.

The `awful.dbus` / `awful.remote` part is more of a problem (and has a higher potential of breaking user configs, I think).

Issue #2678, opened 2019-02-19
naughty.notify does not return id anymore, breaking replaces_id feature


BarbUk:
Output of `awesome --version`:

```
awesome v4.3-81-gddf422dd-dirty (Too long)
 • Compiled against Lua 5.1.4 (running with LuaJIT 2.0.5)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

**How to reproduce the issue:**
In rc.lua:
```lua
noti = {}
noti.id = nil
function noti:notify (text)
  self.id = naughty.notify({ text = text, timeout = 5, replaces_id = self.id}).id
end
```

From a terminal, execute several times
```bash
echo "noti:notify('test')" | awesome-client
```

**Actual result:**

Multiple notification

**Expected result:**

Only one notification, each new one replacing the old one.

**Notes**

It works as expected with this version:
```
awesome v4.3 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

Bisecting the code give this commit as not working: e13b9a48fffa5fdd99b6f676a2389a0b3ff15f45

The commit message say:
> The leaves the old
> `replaces_id` logic mostly intact for now to keep the full backward
> compatibility.

Is there a new way to replace a notification ?

blueyed:
As a workaround you could get the `id` through the callback I guess.
At least that's tested (see #2680).

psychon:
@blueyed This issue is about the following:
```
$ awesome-client 'return require("naughty").notify{ text = "text"}.id
$
```
Previously, there was an `id` property on notification objects. Now, there is apparently not any more.
The test in #2680 that gets the ID tests the DBus API and not the Lua API, so is (slightly) unrelated.

Elv13:
Now technically you can do `.message` or `.title` on the notification object without using the old ID based stuff. The IDs exist, but are currently restricted to notifications from dbus while before all notifications ended up with one. It's a regression and I will fix it. Not today, but soon.

BarbUk:
> Now technically you can do `.message` or `.title` on the notification object

Nice, I didn't know about that.

So to do a replacing notification, one can do something like that:
```lua
local noti = {}
noti.notification = nil
function noti:notify (message)
  if self.notification then
    self.notification.message = message
  else
[...]

PR #2680, opened 2019-02-19
test-naughty-legacy: Do not spawn external processes

psychon:
This changes the test to use `GDBusConnection` instead of spawning external processes. This was inspired by https://github.com/awesomeWM/awesome/pull/2679#issuecomment-465088754. Let's see which explosions we see on Travis due to this. :-)

### DIFF
--- a/tests/test-naughty-legacy.lua
+++ b/tests/test-naughty-legacy.lua
@@ -4,16 +4,41 @@ local spawn     = require("awful.spawn")
 local naughty   = require("naughty"    )
 local gdebug    = require("gears.debug")
 local cairo     = require("lgi"        ).cairo
+local Gio       = require("lgi"        ).Gio
+local GLib      = require("lgi"        ).GLib
 local beautiful = require("beautiful")
 
 -- This module test deprecated APIs
 require("gears.debug").deprecate = function() end
 
+local dbus_connection = assert(Gio.bus_get_sync(Gio.BusType.SESSION))
+
+local function simple_send_notify_callback(conn, result)
+    -- This gives us the id which we don't need, but it still provides it
+    local _ = conn:call_finish(result).value[1]
+end
+
+local function simple_send_notify(summary, body, timeout, urgency)
+    local hints = {}
+    if urgency then
+        local urgency_map = { low = 0, normal = 1, critical = 2 }
+        hints.urgency = GLib.Variant("y", assert(urgency_map[urgency], urgency))
+    end
+    local parameters = GLib.Variant("(susssasa{sv}i)", {
+        "notify-send", 0, "", summary, body, {}, hints, timeout, urgency
+    })
+    local reply_type = GLib.VariantType.new("(u)")
+    dbus_connection:call("org.freedesktop.Notifications",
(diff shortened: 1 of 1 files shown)


tests/test-naughty-legacy.lua
```diff
@@ -563,120 +548,86 @@ table.insert(steps, function()
...
-        'org.freedesktop.Notifications.Notify',
-        'string:"Awesome test"',
-        'uint32:0',
-        'string:""',
-        'string:"title"',
-        'string:"message body"',
-        'array:string:1,one,2,two,3,three',
-        'dict:string:string:"",""',
-        'int32:25000'
-    }
```
blueyed: Could/should this be left?

tests/test-naughty-legacy.lua
```diff
@@ -563,120 +548,86 @@ table.insert(steps, function()
...
+--TODO Test action with long names.
 
-    local nid, name_u, message_u, actions_u = nil
+local nid, name_u, message_u, actions_u = nil
 
-    -- Test updating a notification.
-    table.insert(steps, function()
-        spawn.easy_async(cmd, function(out)
-            nid = tonumber(out:match(" [0-9]+"):match("[0-9]+"))
+-- Test updating a notification.
```
blueyed: This should/could also test https://github.com/awesomeWM/awesome/issues/2678.
I assume it works here, because the id is assigned throught the callback?

blueyed:
Nice!
Also a test for the regression in https://github.com/awesomeWM/awesome/pull/2668#issuecomment-464805640 would be good to have.

Elv13:
Testing with the real `notify-send` is kind of the point. This is the reference implementation. This is what should work.

tests/test-naughty-legacy.lua
psychon: Could: Yes, of course. But why?
Should: I don't know. What would be a reason for leaving it in?

tests/test-naughty-legacy.lua
psychon: Uhm, there is already a test for that. Exactly this piece here. It gets the notification it and then replaces the notification and tests that it was really replaced. I did not look at #2678, but I would guess that it is a problem with the Lua API, not the DBus API.

psychon:
Well, okay. @Elv13 so what about the second commit that gets rid of `dbus-send`? Should I just get rid of 49ceb30205235b1e2c1a0c053272dc04502a06e8 and have the second commit?

tests/test-naughty-legacy.lua
blueyed: I am not sure, just noticed this.
@Elv13 ?

tests/test-naughty-legacy.lua
blueyed: Yes, here it appears to use the id via the callback, while with #2678 it appears to not get returned?

tests/test-naughty-legacy.lua
blueyed: Maybe this could be kept to test `notify-send` as a reference implementation?

psychon:
Superseded by #2683.

PR #2682, opened 2019-02-19
Travis: install via apt addons

blueyed:


blueyed:
Removed the extra stage again: not really that useful.
Note: we could skip it still by looking at `$TRAVIS_COMMIT_MESSAGE`, but that's also not that useful usually.

PR #2687, opened 2019-02-20
chore(common: version): replace unicode symbols to yes/no

actionless:
There are some usecases when terminal font not providing those characters at all or they're provided by fallback font with different dimensions.
Both of the issues above could affect readability/clarity of the version output.

tty

### DIFF
--- a/common/version.c
+++ b/common/version.c
@@ -50,19 +50,19 @@ eprint_version(void)
     (void) luaL_dostring(L, "return require('lgi.version')");
 
 #ifdef WITH_DBUS
-    const char *has_dbus = "✔";
+    const char *has_dbus = "yes";
 #else
-    const char *has_dbus = "✘";
+    const char *has_dbus = "no";
 #endif
 #ifdef WITH_XCB_ERRORS
-    const char *has_xcb_errors = "✔";
+    const char *has_xcb_errors = "yes";
 #else
-    const char *has_xcb_errors = "✘";
+    const char *has_xcb_errors = "no";
 #endif
 #ifdef HAS_EXECINFO
-    const char *has_execinfo = "✔";
+    const char *has_execinfo = "yes";
 #else
-    const char *has_execinfo = "✘";
+    const char *has_execinfo = "no";
 #endif
 
     printf("awesome %s (%s)\n"


Veratil:
While we're changing that, change the unicode list dots "•" to "*" ?

https://github.com/awesomeWM/awesome/blob/master/common/version.c#L64

actionless:
@Veratil  i didn't meant to start anti-unicode campaign, my only point was what yes/no thing is important for debugging

however i'd agree what unicode/emoji symbols looking usually a bit alienated in the terminal because not every font have them, and when it comes to fallback fonts -- terminals usually not playing very nice typography of such "mixed" font

blueyed:
What about using `[ ]` and `[x]`?

actionless:
all those smileys could work for interactive UIs but imho for output of such info which is needed for debugging and usually being accessed already when things are going wrong it's better to stay more explicit on yes/no, or 1/0 (though the latter could be read as version number if set to 1)

blueyed:
Good point.
What do you think about "D-Bus support" vs. "No D-Bus support"?

actionless:
no strong preference

blueyed:
Yes, me neither.. yes/no sounds also good to me.

psychon:
Neither have I. Random thought: "yes/no" is easier to find/read in a quick glance than than "DBus support / No DBus support".

blueyed:
Ok.  Let's have it after https://github.com/awesomeWM/awesome/pull/2665 then?!

actionless:
https://github.com/awesomeWM/awesome/pull/2665 is merged now

blueyed:
Rebase it then?

actionless:
done

blueyed:
btw: still used here: https://github.com/awesomeWM/awesome/blob/ab0b4dc8ac0520e4b7f52e9f4c0413c5534f4017/awesome.c#L701-L710

Would be nice to a) have this include the filename, and b) write to stdout (instead of stderr).
/cc @psychon

actionless:
there it's duplicating the text message, so i wasn't sure how to replace it (yes/no thing wasn't applying there)

blueyed:
Yes, we do not need it there really anymore, but could just having something like "checking config %s..", and then "ok" or the error after it.

actionless:
> write to stdout (instead of stderr).

also i'm strongly against this part (didn't noticed initially):

by default in most of the distributions awesome prints stderr to ~/.xsession-error and stdout going just to nowhere

PR #2689, opened 2019-02-21
Fix naughty legacy wrt to id

blueyed:
Ref: https://github.com/awesomeWM/awesome/issues/2678

### DIFF
--- a/tests/test-naughty-legacy.lua
+++ b/tests/test-naughty-legacy.lua
@@ -257,6 +257,7 @@ local function test_overlap()
 
     for _, n1 in ipairs(active) do
         assert(n1.box)
+        assert(n1.id)
 
         -- Check for overlapping the workarea
         assert(n1.box.y+default_height < wa.y+wa.height)
@@ -405,6 +406,7 @@ table.insert(steps, function()
         timeout = 25000,
     }
 
+    assert(n1.id)
     assert(n1.iconbox)
     assert(n1.iconbox._private.image:get_width () == beautiful.notification_icon_size)
     assert(n1.iconbox._private.image:get_height() == beautiful.notification_icon_size)
@@ -420,6 +422,7 @@ table.insert(steps, function()
         timeout = 25000,
     }
 
+    assert(n2.id)
     assert(n2.iconbox)
     assert(n2.iconbox._private.image:get_width () == 32)
     assert(n2.iconbox._private.image:get_height() == 32)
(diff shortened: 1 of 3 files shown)


Elv13:
That's not really related to the issue. The issue is that there should always be an id. But this is no one liner. It needs to have a private id factory in core.lua and use the result in dbus.lua. The current bug is that the `id`s are generated in dbus.lua so when you create a notification "manually" from within Awesome that code path never get hit.

blueyed:
@Elv13 
Yes, I found that it is not trivial, but wanted at least to test the current behavior.

blueyed:
@Elv13 
Do you plan to bring compatibility back?
Maybe a special id/counter could just be used with the compat wrappers?

Elv13:
> Do you plan to bring compatibility back?

Yes, but I wont have time for another couple days.

Elv13:
@blueyed I made progress based based on your work, I took the liberty to update this PR

Elv13:
> i have a feeling what i was already reviewing or even submitting a similar change about a year ago, mb it got lost while merging new stuff

It changed because the code was split in more files and the counter ended up in the wrong file. It worked fine for notifications created via dbus, but wasn't working for notification created locally.

tests/test-naughty-legacy.lua
```diff
@@ -698,10 +742,9 @@ table.insert(steps, function()
     assert(n.text    == "baz")
 
     -- Test the ID system
-    n.id = 1337
-    assert(n.id == 1337)
-    assert(naughty.getById(1337) == n)
-    assert(naughty.get_by_id(1337) == n)
+    local id = n.id
+    assert(naughty.getById(id) == n)
```
actionless: is that camelcase notation supported for all the methods as an alternative to underscore?!

Elv13: more like a snake case alternative to an API that used CamelCase. It's not the only example. There is about a hundred API elements using nocase or CamelCase that predates the standardization to snake_case. Some are still not deprecated (mostly in awful).

actionless: but i meant are those both methods are implemented explicitly or we have some general handler for that?

Elv13: > but i meant are those both methods are implemented explicitly or we have some general handler for that?

All existing CamelCase and nocase functions and method part of the public API get deprecated when I work in the module they belong to. The CamelCase and nocase end up with a `gears.deprecate` call and then call the snake_case equivalent.

tests/test-naughty-legacy.lua
```diff
@@ -698,12 +698,12 @@ table.insert(steps, function()
...
-    assert(naughty.getById(1337) == n)
-    assert(naughty.get_by_id(1337) == n)
-    assert(naughty.getById(42) ~= n)
-    assert(naughty.get_by_id(42) ~= n)
+--     n.id = 1337
+--     assert(n.id == 1337)
+--     assert(naughty.getById(1337) == n)
+--     assert(naughty.get_by_id(1337) == n)
+--     assert(naughty.getById(42) ~= n)
+--     assert(naughty.get_by_id(42) ~= n)
```
blueyed: Use `set_id` for this test maybe?

Also could test the assertion error with setting it directly here.

Elv13: > Use set_id for this test maybe?

The property <-> accessor API is tested in thousands of places, it would be overkill.

> Also could test the assertion error with setting it directly here.

Done

blueyed:
Thanks for fixing this regression!

psychon:
Should this have closed #2678?

blueyed:
Yes, closed.

PR #2691, opened 2019-02-22
Travis: allow skipping TEST_PREV_COMMITS with issue label

blueyed:


blueyed:
Related: we could have a status for if testing prev commits failed (https://developer.github.com/v3/repos/statuses/).

PR #2693, opened 2019-02-22
[RFC] cmake: GENERATE_DOC=OFF

blueyed:
It takes quite a while to build the docs (mostly due to the examples),
so this is turning it off by default.


blueyed:
This is something that would affect packagers.

PR #2696, opened 2019-02-23
naughty: Resize notification when setting text

alfunx:
The notification used to update its size when the text or title was set through `naughty.replace_text`.

Moved a function, most relevant change: https://github.com/awesomeWM/awesome/pull/2696/files#diff-7950c90cea9d7c92298f5efc977117c6R241

### DIFF
--- a/lib/naughty/layout/legacy.lua
+++ b/lib/naughty/layout/legacy.lua
@@ -126,42 +126,6 @@ local function get_offset(s, position, idx, width, height)
     return v
 end
 
-local escape_pattern = "[<>&]"
-local escape_subs    = { ['<'] = "&lt;", ['>'] = "&gt;", ['&'] = "&amp;" }
-
--- Cache the markup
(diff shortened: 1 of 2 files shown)


psychon:
I want to wait for @Elv13 to look at this, so I am not going to :+1: this, but just a quick question: What was/is the problem here? If I see correctly, this patch just moves some code down. Does that mean that this is fixed by just calling `connect_signal` later (= does this depend on signal emission to call the functions in the order they were connected?) or what? Basically, I don't understand this.

alfunx:
It's mainly about calling `update_size` at the end of the `set_escaped_text` function (the line I linked), for which I had to move the latter function down.

Elv13:
Thanks for this, sorry for the disruption. A test would be nice, but this can be merged first so less users get regression.

alfunx:
I added a test for this since it's not merged yet.

PR #2699, opened 2019-02-24
[RFC/WIP] call awesome_restart_delayed via main loop

blueyed:
Fixes `echo 'awesome.restart()' | awesome-client` to not display an
error about failed dbus-send, caused by awesome restarting before
replying through DBus.

Inspired by / via https://www.reddit.com/r/awesomewm/comments/a3xug0/how_to_restart_awesome_from_the_command_line/ebcisiw/ .

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -524,12 +524,19 @@ exit_on_signal(gpointer data)
     return TRUE;
 }
 
-void
-awesome_restart(void)
+static gboolean
+awesome_restart_delayed(gpointer unused)
 {
     awesome_atexit(true);
     execvp(awesome_argv[0], awesome_argv);
     fatal("execv() failed: %s", strerror(errno));
+    return true;
+}
+
+void
+awesome_restart(void)
+{
+    g_idle_add_full(G_PRIORITY_DEFAULT, awesome_restart_delayed, NULL, NULL);
 }
 
 /** Function to restart awesome on some signals.


awesome.c
```diff
@@ -524,12 +524,19 @@ exit_on_signal(gpointer data)
...
     awesome_atexit(true);
     execvp(awesome_argv[0], awesome_argv);
     fatal("execv() failed: %s", strerror(errno));
+    return true;
+}
+
+void
+awesome_restart(void)
+{
+    g_idle_add_full(G_PRIORITY_DEFAULT, awesome_restart_delayed, NULL, NULL);
```
blueyed: Totally not sure about signature / name.

psychon:
This breaks the Lua Panic handler.

From http://pgl.yoyo.org/luai/i/lua_atpanic:

> If an error happens outside any protected environment, Lua calls a panic function and then calls exit(EXIT_FAILURE), thus exiting the host application. Your panic function can avoid this exit by never returning (e.g., doing a long jump).

After this change, the panic handler returns and Lua will call `exit`.

------

Also, this is a slight behaviour change. If anyone out there has something like `awesome.restart() print("foo")` (I'm sure one can come up with more realistic examples of this). I'll leave it for others to decide if this change is okay. On one hand, it makes things more consistent with `awesome.exit()` (which also returns and finishes the current main loop iteration). On the other hand, I remember complains about `awesome.exit()` returning.

In case this behaviour change is okay: You could just move the call to `g_idle_add` to `luaA_restart`.

Elv13:
> (I'm sure one can come up with more realistic examples of this).

Sigh... 2 did.

https://old.reddit.com/r/awesomewm/comments/a3xug0/how_to_restart_awesome_from_the_command_line/ebcisiw/

blueyed:
@Elv13 
> Sigh... 2 did.

This is the example I've read and wanted to look into if this confusion can be fixed.

(I am having a loop around `awesome`, so for me `awesome.quit()` would work without any error/warning)

@psychon 
> After this change, the panic handler returns and Lua will call `exit`.

Ah, because `luaA_panic` calls `awesome_restart`.  We could still call the inner function (`awesome_restart_delayed`) there I guess (and rename it); but that's maybe what you've meant with moving it to `luaA_restart`?

> If anyone out there has something like `awesome.restart() print("foo")`

What do you mean exactly?  Calling something after `awesome.restart()`?

psychon:
> What do you mean exactly? Calling something after `awesome.restart()`?

Yup. Perhaps a better example would be something like `if foo then awesome.restart() end print("Now we do something here")`. Your patch wold make `awesome.restart` return and now more code runs. This could have unintended consequences somewhere.

> but that's maybe what you've meant with moving it to `luaA_restart`?

To make it more explicit what I meant:
```diff
diff --git a/luaa.c b/luaa.c
index 317f2cf0d..3572c43a8 100644
--- a/luaa.c
+++ b/luaa.c
@@ -250,7 +250,7 @@ luaA_exec(lua_State *L)
 static int
 luaA_restart(lua_State *L)
 {
-    awesome_restart();
+    g_idle_add(42, awesome_restart, 0);
     return 0;
 }
 
```

Issue #2701, opened 2019-02-24
The documentation DOC_GEN_OUTPUT is no longer working


Elv13:
Since the recent changes, `DOC_GEN_OUTPUT` based examples are still displayed, but they are no longer tested and the `.output.txt` is no longer generated.

blueyed:
Due to https://github.com/awesomeWM/awesome/pull/2545 I assume?!
Should we revert it for now?

Elv13:
Probably

Elv13:
I spent some time fixing it, but it wont be ready for now. There's a lot of corner cases and circular dependencies to move the lua `configure_file` to a target, the `load_ldoc` to a target and have all the depend on all the files. Then have `ldoc` depend on *that* target.

psychon:
> Due to #2545 I assume?!

Before #2545, all the examples where collected in `${EXAMPLE_DOC_EXTRA_TARGETS}` and then `generate-examples` depended on this. Since this target was marked `ALL`, it was built by default.

After this commit, this target is no longer `ALL` and it only depends on the generated files (`${EXAMPLE_DOC_GENERATED_FILES})`), not the extra targets (`${EXAMPLE_DOC_EXTRA_TARGETS}` which even no longer exists).

Thus... isn't this "don't run the tests" exactly what #2545 was trying to do (besides removing this one `ALL`)?

blueyed:
> Thus... isn't this "don't run the tests" exactly what #2545 was trying to do (besides removing this one `ALL`)?

Yes.
I think we could/should add the `.output.txt` files also to `EXAMPLE_DOC_GENERATED_FILES` maybe.

blueyed:
https://github.com/awesomeWM/awesome/pull/2702

PR #2702, opened 2019-02-25
tests/examples: also generate rules for output.txt files

blueyed:
Ref: https://github.com/awesomeWM/awesome/issues/2701#issuecomment-466890210


Elv13:
You might want to consider https://github.com/awesomeWM/awesome/pull/2600/commits/f2c3c3698369d282b7e316c2699a8a61cdce689e instead of this. It ain't really ready to shine, but by removing the remaining `configure_file` from the "configuration" cmake execution, you get rid of some hacks that tend to cause the regression we are dealing with. It has some other advantages, such as the shared documentation (docs/common/*) being able to embed example output themselves.

blueyed:
@Elv13 
Nice!
I'd say to get this one here in maybe before already, and then your big cleanup/refactor based on that - to keep it more stable for now.

blueyed:
Hmm, with this `make clean` currently will delete the `output.txt` files..

PR #2706, opened 2019-02-25
Mention the new selection API in NEWS

psychon:

This is a draft since it depends on #2643 (which seems to be stuck waiting for reviews, perhaps @blueyed can take a look again? You already reviewed a previous version and there were only minimal changes) and #2661 (which is also blocked by #2643).

-------

- [x] Personal todo: After this is merged, go through all commits in master since the release, add items to the NEWS and also mention the commit up to where things are included in this NEWS.

### DIFF
--- a/docs/89-NEWS.md
+++ b/docs/89-NEWS.md
@@ -10,15 +10,36 @@
 Some general text about Awesome v4.4 should be here. Hopefully someone will
 notice that it is not yet here before we release...
 
+This document was last updated at commit v4.3-148-g795c792d1.
+
 ## New features
+
+* Lua code can interact with the selection contents via the new
+  `selection.acquire`, `selection.getter`, and `selection.watcher` objects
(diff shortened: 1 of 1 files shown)


psychon:
I went through the changes since the release and added some entries to NEWS for them. I also left a comment mentioning the current commit id of master. That way, the next time the file is updated, it should be obvious which part of the git history needs to be looked at.

blueyed:
I assume the TODOs are there for a later release?
Do you mean to merge this already, or just have it edited here from time to time?
(I think it is not bad to have it updated already in master)

psychon:
The TODOs are in there because I never looked closely at the recent naughty changes and perhaps someone wants to say more about them.

Yup, I fully intend them to go away before a release, but I don't mind having them in master already. It's work-in-progress. ;-)
(Also, I think it's a bad idea to have PRs touching NEWS sitting around for long. It's just a recipe for lots of conflicts. So yes, I mean to merge this already.)

PR #2711, opened 2019-02-28
Improve the test runner: Allow "direct" tests (instead of only steps-based tests)

psychon:


tests/_runner.lua
```diff
@@ -43,7 +43,54 @@ runner.step_kill_clients = function(step)
     end
 end
 
-runner.run_steps = function(steps, options)
+--- Print a message if verbose mode is enabled
```
blueyed:
```suggestion
--- Print a message if verbose mode is enabled.
```

tests/_runner.lua
```diff
@@ -43,7 +43,54 @@ runner.step_kill_clients = function(step)
...
+-- @tparam string message The message to print
```
blueyed:
```suggestion
-- @tparam string message The message to print.
```

tests/_runner.lua
```diff
@@ -43,7 +43,54 @@ runner.step_kill_clients = function(step)
...
+function runner.verbose(message)
+    if verbose then
+        io.stderr:write(message .. "\n")
+    end
+end
+
+--- When using run_direct(), this function indicates that the test is now done.
+-- @tparam[opt=nil] string message An error message explaining test failure.
```
blueyed:
```suggestion
-- @tparam[opt=nil] string message An error message explaining the test failure, if it failed.
```
It was not clear to me if this was called always, or only for errors.

tests/_runner.lua
```diff
@@ -92,45 +131,24 @@ runner.run_steps = function(steps, options)
                 wait = 20
                 t.timeout = 0
                 t:again()
-                return
+            else
+                -- All steps finished, we are done
```
blueyed:
```suggestion
                -- All steps finished, we are done.
```

tests/_runner.lua
```diff
@@ -65,24 +111,17 @@ runner.run_steps = function(steps, options)
...
         -- Call the current step's function.
         local success, result = xpcall(function()
             return steps[step](step_count)
         end, debug.traceback)
 
         if not success then
-            io.stderr:write('Error: running function for step '
-                            ..step_as_string..': '..tostring(result)..'!\n')
+            runner.done('running function for step '
+                            ..step_as_string..': '..tostring(result)..'!')
```
blueyed: Wrap this into a single line / or align the continuation line?!

tests/_runner.lua
```diff
@@ -92,45 +131,24 @@ runner.run_steps = function(steps, options)
...
             -- No result yet, run this step again.
             wait = wait-1
             if wait > 0 then
                 t.timeout = 0.1
                 t:again()
             else
-                io.stderr:write("Error: timeout waiting for signal in step "
-                                ..step_as_string..".\n")
+                runner.done("timeout waiting for signal in step "
+                                ..step_as_string..".")
```
blueyed: Also wrap/align.

tests/_runner.lua
```diff
@@ -43,7 +43,54 @@ runner.step_kill_clients = function(step)
...
+
+    if not message then
+        io.stderr:write("Test finished successfully.\n")
+    end
+    awesome.quit()
+end
+
+--- This function is called to indicate that a test does not use the run_steps()
+-- facility, but instead runs something else directly.
+function runner.run_direct()
```
blueyed: s/run_direct/is_running/ maybe?
The current name indicated that it would do something, although with the example in test-gravity it is called only after the test - and it seems like the `runner.run_direct()` call is not necessary there after all?!
It could also just set `runner.is_running` instead maybe.

tests/_runner.lua
psychon: Both of these do not fit into the textwidth=80, so I just left the indentation the way it was before. Only now do I understand that this was meant to be aligned to the `(` in the line before...

Will handle this later.

tests/_runner.lua
psychon: Well, no. `test-gravity.lua` starts an external process and afterwards it calls `run_direct()`. However, at this time, the test is not yet done. Only when the external process finishes some time later is `done()` actually called. So, it's still asynchronous.

The idea why I call `run_direct()` at the end is because it "disarms" a timer in `_runner.lua` that would otherwise complain that the test was not started correctly and quit awesome. So, if starting the external process would fail, awesome would quit relatively quickly instead of waiting for the 180 seconds timeout.

Right now I don't see a nice way to make this more explicit... well, `run_direct` could get a function as its argument which it just runs, but that might also be weird...

Ideas?

blueyed: Thanks for the detailed explanation, as always.

It is fine by me.

PR #2714, opened 2019-03-02
awful.spawn: Check arguments in once and single_instance

psychon:
Doing awful.spawn.once("gvim") and awful.spawn.single_instance("gvim")
resulted in the following non-intuitive error:
```
2019-03-02 11:55:46 W: awesome: luaA_dofunction:78: error while running function!
stack traceback:
	[C]: in function 'pairs'
	lib/gears/table.lua:51: in function 'gears.table.crush'
	lib/awful/rules.lua:462: in field 'callback'
	lib/awful/rules.lua:491: in function 'awful.rules.apply'
error: lib/gears/table.lua:51: bad argument #1 to 'pairs' (table expected, got nil)
```
What is going on here is that thanks to startup-notification, the new
instance of GVim is found. Then, rules are applied. However, no rules
were provided.

Since this error happens only some time later, the traceback does not
say anything useful and I was at first confused at was going on.

Improve this by making sure these two functions actually get the
required arguments.

### DIFF
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -645,6 +645,8 @@ end
 -- @tparam[opt] function callback A callback function when the client is created.
 -- @see awful.rules
 function spawn.once(cmd, rules, matcher, unique_id, callback)
+    assert(type(cmd) == "string" or type(cmd) == "table")
+    assert(type(rules) == "table")
     local hash = unique_id or hash_command(cmd, rules)
     local status = register_common(hash, rules, matcher, callback)
     run_after_startup(function()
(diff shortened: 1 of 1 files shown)


Elv13:
I think we could make the rules optional instead of asserting? The part where the property is set on the client is sometime all that's needed.

(Also, as "reported" on Reddit, we should probably also flag the root window for `spawn.once` to mitigate the ones without working startup notification IDs)

psychon:
> we should probably also flag the root window for spawn.once to mitigate the ones without working startup notification IDs

After starring at the code for a bit: No idea what you mean here. All info about SNIDs will be lost when awesomeWM restarts, so I don't see how "client does not support SNID" should make a difference here...

psychon:
Superseded by #2716

Issue #2721, opened 2019-03-09
Weird issue with the new naughty code


psychon:
Patch that should not break anything:
```diff
diff --git a/lib/naughty/dbus.lua b/lib/naughty/dbus.lua
index 7218f1e35..cce2f838c 100644
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -65,6 +65,7 @@ local function sendActionInvoked(notificationId, action)
 end
 
 local function sendNotificationClosed(notificationId, reason)
+    assert(type(reason) == "number", "Expected number, but got: " .. type(reason))
     if capi.dbus then
         capi.dbus.emit_signal("session", "/org/freedesktop/Notifications",
         "org.freedesktop.Notifications", "NotificationClosed",
```
The way in which actually `tests/test-naughty-legacy.lua` breaks:
```
Error: running function for step 25/27 (@1): lib/naughty/dbus.lua:68: Expected number, but got: table
stack traceback:
	[C]: in function 'assert'
	lib/naughty/dbus.lua:68: in upvalue 'sendNotificationClosed'
	lib/naughty/dbus.lua:178: in method 'destroy'
	...e/psychon/projects/awesome/tests/test-naughty-legacy.lua:713: in function <...e/psychon/projects/awesome/tests/test-naughty-legacy.lua:707>
	(...tail calls...)
	[C]: in function 'xpcall'
	/home/psychon/projects/awesome/tests/_runner.lua:123: in function </home/psychon/projects/awesome/tests/_runner.lua:116>
	[C]: in function 'xpcall'
	lib/gears/protected_call.lua:36: in function <lib/gears/protected_call.lua:35>
	(...tail calls...)
	lib/gears/timer.lua:237: in function 'gears.timer.run_delayed_calls_now'!
```
What happens is that the test creates a notification through Lua. Then it replaces it through DBus. This causes this code to run:
https://github.com/awesomeWM/awesome/blob/f69b72ab3657d716eca9af5f5df31d7810cfd577/lib/naughty/dbus.lua#L176-L178
Next, this code runs:
https://github.com/awesomeWM/awesome/blob/f69b72ab3657d716eca9af5f5df31d7810cfd577/lib/naughty/dbus.lua#L209-L211
[...]

PR #2722, opened 2019-03-09
naughty.dbus: Switch to using Gio for DBus bindings

psychon:
This starts the switch from our own, semi-broken DBus bindings to using
the sane bindings that Gio provides.

Part-of: https://github.com/awesomeWM/awesome/issues/1093

------

This will break on Travis because of #2721, because LGI/Gio actually checks that a number is a number and not a table.

### DIFF
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -191,7 +191,17 @@ local function protected_method_call(_, sender, object_path, interface, method, 
             if icon ~= "" then
                 args.icon = icon
             elseif hints.icon_data or hints.image_data then
-                if hints.icon_data == nil then hints.icon_data = hints.image_data end
+                -- Icon data is a bit complex and hence needs special care:
+                -- .value breaks with the array of bytes (ay) that we get here.
+                -- So, bypass it and look up the needed value differently
+                local icon_data
+                for k, v in parameters:get_child_value(7 - 1):pairs() do
+                    if k == "icon_data" then
+                        icon_data = v
+                    elseif k == "image_data" and icon_data == nil then
+                        icon_data = v
+                    end
+                end
 
                 -- icon_data is an array:
                 -- 1 -> width
@@ -201,8 +211,12 @@ local function protected_method_call(_, sender, object_path, interface, method, 
                 -- 5 -> bits per sample
                 -- 6 -> channels
                 -- 7 -> data
-                local w, h, rowstride, _, _, channels, icon_data = unpack(hints.icon_data)
-                args.icon = convert_icon(w, h, rowstride, channels, icon_data)
+
+                -- Get the value as a GVariant and then use LGI's special
+                -- GVariant.data to get that as an LGI byte buffer. That one can
+                -- then by converted to a string via its __tostring metamethod.
+                local data = tostring(icon_data:get_child_value(7 - 1).data)
+                args.icon = convert_icon(icon_data[1], icon_data[2], icon_data[3], icon_data[6], data)
             end
             if replaces_id and replaces_id ~= "" and replaces_id ~= 0 then
                 args.replaces_id = replaces_id
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -7,17 +7,16 @@
 -- @module naughty.dbus
 ---------------------------------------------------------------------------
 
-assert(dbus)
-
 -- Package environment
 local pairs = pairs
 local type = type
 local string = string
-local capi = { awesome = awesome,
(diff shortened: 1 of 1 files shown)


dtzWill:
Seems to work! One fixup noticed introspecting dbus signatures: 485eee9a2c8d7441d853a54eea1447baf6489447

psychon:
Whoops. Thanks for spotting that. I cherry-picked your commit into this PR.
(And since you write "it is magical enough not to need them": The types for everything are specified in `on_bus_acquire` when the `Gio.DBusInterfaceInfo` is constructed. It says here that `GetCapabilities` returns an array of strings ("as") called "caps": ` method{ name = "GetCapabilities", out_args = { arg("caps", "as") } }`)

psychon:
Rebased on master to get the fix for #2721 and to fix the merge conflict with the fix (#2728). I expect that the Travis build will pass now.

dtzWill:
Here's a commit I put together that fixes handling of `icon_data` when using this PR's changes:

https://github.com/dtzWill/awesome/commit/eecdeb7d469b061a2499eff3a60c6847d5cee020

I'm new to lua so this very likely can/should be improved, but this does seem to fix things for me, so thought I'd share.
I believe this is a problem introduced by the changes in this PR,
but honestly not sure if it was working with master previously.

-------

What was happening is the `icon_data` hint was unpacked as a "string" that was empty for some reason-- looking at the lgi docs for variants I think it's because complex types need to be indexed using the provided methods not just `unpack`? Anyway this gets icons sent as bytes working again.  LMK if you have any questions or this seems unrelated to changes here.

psychon:
@dtzWill Thanks for testing this and fixing stuff. Also, sorry for causing so much breakage.

Per https://developer.gnome.org/notification-spec/#icons-and-images-formats, the image data is sent as signature `iiibiiay` (yay for DBus type readability). So, we have here:
- int32
- int32
- int32
- boolean
- int32
- int32
- array of bytes

Anyway... I'm not completely happy with your patch (sorry!) and will (hopefully) find some time to look more closely at this on Saturday.

psychon:
After some "fun", I figured out that LGI truncates GVariant bytestrings at the first embedded nul byte, because the C API that it is based on does so. More details can be found in https://github.com/pavouk/lgi/pull/223.
Since I doubt AwesomeWM can wait for a fix in a released version of LGI, I'll do a work-around similar to what @dtzWill came up with.

psychon:
I pushed a commit that also works around this problem. It is based heavily on the one from @dtzWill so thanks a lot for figuring out how to work around the problem!

I tested my commit with the following C program:
```C
#include <gio/gio.h>
#include <libnotify/notify.h>
int main() {
	notify_init ("Hello world!");
	NotifyNotification * Hello = notify_notification_new ("Hello world", "This is an example notification.", NULL);
	notify_notification_set_image_from_pixbuf(Hello, gdk_pixbuf_new_from_file("/tmp/gradient.png", NULL));
	notify_notification_show (Hello, NULL);
	g_object_unref(G_OBJECT(Hello));
	notify_uninit();
	return 0;
}
```

dtzWill:
On Fri, 22 Mar 2019 10:13:04 -0700, Uli Schlachter <notifications@github.com> wrote:
> I pushed a commit that also works around this problem. It is based heavily on the one from @dtzWill so thanks a lot for figuring out how to work around the problem!

Thanks for tackling this! Switched to your much nicer patch, works great!

\o/

psychon:
> One little "downside" of this PR is that the old dbus.c lost a lot of coverage. 

Well, let's just say that everyone who wants to change/fix something in `dbus.c` is "doing it wrong". That way, we do not need coverage for `dbus.c` since it does not change. Once nothing in AwesomeWM uses it anymore, we can deprecate it and remove it in something like 10 years.

Elv13:
If there is no no other reviews, I will merge this soon. It needs more widespread testing before we can release v4.4.

Elv13:
Should there be a CMakeLists.txt and README.md change to detect GDBus support for Gio?

psychon:
I do not understand the question. `CMakeLists.txt` does not check for dependencies that are only accessed through LGI. Instead, `build-utils/lgi-check.c` does. Also, GDBus is part of Gio (e.g. `require("lgi").Gio.DBusMethodInfo`) and this file already checks for Gio. Same for `README.md`: Gio is already listed as a dependency.

So... could you clarify what you mean with your comment?

Elv13:
> So... could you clarify what you mean with your comment?

Sorry, I was under the impression is was an optional part of Gio like GVFS, but it isn't.

PR #2723, opened 2019-03-11
Fixed input_passthrough property not being set

lawsdontapplytopigs:
In the table of properties supplied to the `wibox` function, you couldn't set the `input_passthrough` property. You could only set it after the wibox was created. 
So this wouldn't work: 
`my_wibox = wibox({
    x = 0,
    y = 0,
    -- etc, etc,
    input_passthrough = true 
    })
`
But this would work:
`my_shlick_wibox = ({
    x = 0,
    y = 0,
    -- etc, etc,
    })
my_shlick_wibox.input_passthrough = true` 
This commit fixes that and now you can set it in both ways.
Sorry if I did something wrong this is my first time submitting a pull request ever.

### DIFF
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -81,8 +81,7 @@ function wibox:to_widget()
         shape              = self._shape,
         forced_width       = self:geometry().width  + 2*bw,
         forced_height      = self:geometry().height + 2*bw,
-        widget             = wibox.container.background,
-        input_passthrough  = false
+        widget             = wibox.container.background
     }
 end
 
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -81,7 +81,8 @@ function wibox:to_widget()
         shape              = self._shape,
         forced_width       = self:geometry().width  + 2*bw,
         forced_height      = self:geometry().height + 2*bw,
-        widget             = wibox.container.background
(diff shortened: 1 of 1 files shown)


lib/wibox/init.lua
```diff
@@ -81,7 +81,8 @@ function wibox:to_widget()
         shape              = self._shape,
         forced_width       = self:geometry().width  + 2*bw,
         forced_height      = self:geometry().height + 2*bw,
-        widget             = wibox.container.background
+        widget             = wibox.container.background,
+        input_passthrough  = false
```
actionless: shouldn't it be `self.input_passthrough` instead of `false`?

lawsdontapplytopigs: I'm not exactly sure, I'm still trying to understand better how the entire codebase works. I thought those were default values and I just went by the documentation and set that to false. I tried changing it with what you suggested and it still works.

psychon: I'd say that this is completely unrelated. `:to_widget()` is seldom used.

Can you get rid of the change above? The other change in here looks fine. Also, thanks a lot for figuring this out and thanks for sending a PR to fix this.

lawsdontapplytopigs: Yeah, no problem! I changed it back!

psychon:
Thanks!

PR #2726, opened 2019-03-13
notification: Add the view components (not used by default until v4.5)

Elv13:
Ok, so just to clear my git stash to fix some of the other open issues, here's a second third of the new notification API.

### Why

The previous notification implementation was hardcoded and lagged both in term of the XDG notification standard and other alternatives such as dunst. The part where a new `naughty.notification` object was merged a while back and is undergoing stabilization work. The `awful.rules` style API to allow filtering and "be better than hardcoded presets" is not submitted yet.

This pull request adds flexible new widget to make custom notifications possible. Each notification can have their own layout (ei. music notifications can have a different look than chat or low power ones). 

To make things a bit more simple to use, there's also widget for each components. This makes the `widget_template` much, much shorter:

Screenshot_20190313_172726

Those new widgets wont be enabled in the default `rc.lua` until they are mature and stable, so merging this pull request is rather low risk compared to the other 2 third of the new API.

### Documentation:

https://elv13.github.io/classes/naughty.notification.html

### To use the new system:

Add this to `rc.lua`. It will disable the legacy widget and use the new ones. Note that as-is, it is pretty useless since it looks nearly the same. However with a `widget_template`, the possibilities are endless.

    naughty.connect_signal("request::display", function(n) naughty.layout.box { notification = n } end)

a more complete widget template would be:

```lua
{
    {
        {
            {
                {
                    naughty.widget.icon,
                    {
                        naughty.widget.title,
                        naughty.widget.message,
                        spacing = 4,
                        layout  = wibox.layout.fixed.vertical,
                    },
                    fill_space = true,
                    spacing    = 4,
                    layout     = wibox.layout.fixed.horizontal,
                },
                naughty.list.actions,
                spacing = 10,
                layout  = wibox.layout.fixed.vertical,
            },
            margins = beautiful.notification_margin,
            widget  = wibox.container.margins,
        },
        id     = "background_role",
        widget = naughty.container.background,
    },
    strategy = "max",
    width    = width(beautiful.notification_max_width
        or beautiful.xresources.apply_dpi(500)
    widget   = wibox.container.constraint,
}

```

### Also

There is some changes to `wibox.widget.base`, `awful.widget.common` and the tag/tasklist commit to use there improvements. This means the `tasklist` now uses `awful.widget.clienticon`. The vast majority of the changes files are either new or are tests. @psychon already reviewed many of these commits 2 years ago, but they changed quite a bit since that. Most of the effort wasn't the original implementation, but rather documentation, tests and integration (which have been done since December)


actionless:
thanks for this!

Elv13:
I don't have that much time to be honest, but I would appreciate if some people could test and review this. Again, merging this pull request wont regress anything for existing users. It is an opt-in feature (at least for v4.4) to get the ability to use custom notification widgets. Even if you read this and never ever reviewed a pull request, step in! This is a great way to start contributing to AwesomeWM.

The build would pass if a few luacheck issues were resolved and it hit a fast enough builder not to burst the time limit. This can wait until after the reviews (like to split the PR in smaller, but meaningless, pull requests).

dtzWill:
The dbus `replaces_id` code doesn't seem to work well with this, any thoughts/plans on accommodating this functionality? Currently I'm seeing errors along these lines:

```
Mar 25 04:30:35 dtznix xsession[1692]: 2019-03-25 04:30:35 E: awesome: Error during a protected call: ...e-2019-03-25/share/awesome/lib/naughty/layout/legacy.lua:231: attempt to index a nil value (upvalue 'textbox')
Mar 25 04:30:35 dtznix xsession[1692]: stack traceback:
Mar 25 04:30:35 dtznix xsession[1692]:         ...e-2019-03-25/share/awesome/lib/naughty/layout/legacy.lua:231: in local 'set_markup'
Mar 25 04:30:35 dtznix xsession[1692]:         ...e-2019-03-25/share/awesome/lib/naughty/layout/legacy.lua:242: in local 'func'
Mar 25 04:30:35 dtznix xsession[1692]:         ...ny-awesome-2019-03-25/share/awesome/lib/naughty/core.lua:214: in function 'naughty.emit_signal'
Mar 25 04:30:35 dtznix xsession[1692]:         ...ny-awesome-2019-03-25/share/awesome/lib/gears/object.lua:154: in function 'gears.object.emit_signal'
Mar 25 04:30:35 dtznix xsession[1692]:         ...me-2019-03-25/share/awesome/lib/naughty/notification.lua:605: in function 'naughty.notification.set_title'
Mar 25 04:30:35 dtznix xsession[1692]:         (...tail calls...)
Mar 25 04:30:35 dtznix xsession[1692]:         ...ny-awesome-2019-03-25/share/awesome/lib/naughty/dbus.lua:235: in function <...ny-awesome-2019-03-25/share/awesome/lib/naughty/dbus.lua:123>
Mar 25 04:30:35 dtznix xsession[1692]:         [C]: in function 'xpcall'
Mar 25 04:30:35 dtznix xsession[1692]:         ...me-2019-03-25/share/awesome/lib/gears/protected_call.lua:36: in function <...me-2019-03-25/share/awesome/lib/gears/protected_call.lua:35>
Mar 25 04:30:35 dtznix xsession[1692]:         (...tail calls...)
Mar 25 04:30:35 dtznix xsession[1692]:         ...ny-awesome-2019-03-25/share/awesome/lib/naughty/dbus.lua:267: in upvalue 'target'
Mar 25 04:30:35 dtznix xsession[1692]:         ...i-0.9.2-1/share/lua/5.3/lgi/override/GObject-Closure.lua:286: in function <...i-0.9.2-1/share/lua/5.3/lgi/override/GObject-Closure.lua:283>
```

Still looking through the rest, but I can report that other than this `replaces_id` issue it indeed seems to not have regressed.  And the icon improvement is nice! ;).

Elv13:
Umm, apparently the "legacy" system remains partially active. You can remove the lines like 

```
naughty.connect_signal("property::text" ,set_escaped_text)
naughty.connect_signal("property::title",set_escaped_text)
```

From `legacy.lua` and it will stop interfering with the new system. I will need to add an `if not notif.box then return end` in the callback to prevent to make sure no legacy code is executed on notifications created with the new API. I don't use any application that update the notifications. Unsurprisingly most of the regressions are there...

lib/wibox/widget/base.lua
```diff
@@ -420,6 +420,14 @@ function base.place_widget_at(widget, x, y, width, height)
...
 end
 
+-- Check if `obj` can be called (either using the metacall or as a function)
+local function is_callable(obj)
+    local t = type(obj)
+    return t == "function" or (
+        t == "table" and getmetatable(obj) and getmetatable(obj).__call
+    ), t
+end
+
```
psychon: @ Commit message: Well, I like that boilerplate of having to add an explicit `widget =`. That provides at least some clarity between "this already is a widget, just use it" and "this is how you can construct a widget". But this declarative fun is your playground, so... whatever.

tests/examples/shims/screen.lua
```diff
@@ -19,10 +19,12 @@ local function create_screen(args)
     }
 
     function s._resize(args2)
+        local old  = s.geometry
         geo.x      = args2.x      or geo.x
         geo.y      = args2.y      or geo.y
         geo.width  = args2.width  or geo.width
         geo.height = args2.height or geo.height
+        s:emit_signal("property::geometry", old)
```
psychon: If you want a suggestion on how to split this PR up into smaller pieces: All these "make the shims behave better"-commits can likely go to their own PR.

lib/wibox/container/margin.lua
```diff
@@ -95,20 +95,30 @@ end
...
+--  the *left*/*right*/*top*/*bottom* keys.
 
 function margin:set_margins(val)
-    if self._private.left   == val and
+    if type(val) == "number" and
+       self._private.left   == val and
        self._private.right  == val and
        self._private.top    == val and
        self._private.bottom == val then
         return
```
psychon: Can you update this to catch "nothing changed" and not emit the signal in that case? For example:
```lua
function margin:set_margins(val)
    if type(val) == "table" then
        self:set_left(assert(val.left))
        etc...
    else
       existing code here
    end
end
```
(The `assert` would catch tables without the necessary entries; that way one gets an error with a better traceback than "something went wrong while drawing").

lib/awful/widget/common.lua
```diff
@@ -46,6 +46,8 @@ local function default_template()
     local ibm = wibox.container.margin(ib, dpi(4))
     local l   = wibox.layout.fixed.horizontal()
 
+    bgb:set_border_strategy("inner")
```
psychon: Isn't that kind of a backwards-incompatible change in the default?

lib/wibox/container/background.lua
```diff
@@ -30,8 +31,11 @@ end
 
 -- Prepare drawing the children of this widget
 function background:before_draw_children(context, cr, width, height)
+    local bw    = self._private.shape_border_width or 0
```
psychon: I bet luacheck complains that `bw` is unused.

lib/wibox/container/background.lua
```diff
@@ -284,6 +327,20 @@ function background:get_shape_clip()
     return true
 end
 
+--- How the border width affects when contained widget.
```
psychon: `when`? Should this be `the`?

lib/wibox/container/background.lua
```diff
@@ -284,6 +327,20 @@ function background:get_shape_clip()
...
+-- The valid values are:
+--
+-- * *none*: Just apply the border, do not affect the content size (default).
+-- * *inner*: Squeeze the size of the content by the border width.
+--
+-- @property border_strategy
+-- @param[opt="none"] string
+
+function background:set_border_strategy(value)
+    self._private.border_strategy = value
```
psychon: You need to (possibly) emit a `layout changed` signal here:

lib/awful/widget/common.lua
psychon: Also, I just noticed: There is no such property yet. Either GitHub shows me the commits in a wrong order, or you use this strategy thing before it exists.

lib/wibox/container/background.lua
```diff
@@ -240,38 +258,63 @@ end
...
     self:emit_signal("widget::redraw_needed")
 end
 
-function background:get_shape_border_width()
+function background:get_border_width()
     return self._private.shape_border_width
 end
 
+background.get_shape_border_width = background.get_border_width
+background.set_shape_border_width = background.set_border_width
```
psychon: Doesn't this need wrappers emitting actual deprecation notices?

lib/wibox/container/background.lua
```diff
@@ -240,38 +258,63 @@ end
...
     self:emit_signal("widget::redraw_needed")
 end
 
-function background:get_shape_border_color()
+function background:get_border_color()
     return self._private.shape_border_color
 end
 
+background.get_shape_border_color = background.get_border_color
+background.set_shape_border_color = background.set_border_color
```
psychon: Same

tests/examples/naughty/helloworld.lua
```diff
@@ -0,0 +1,18 @@
+--DOC_HIDE_ALL
+-- local naughty = require("naughty")
+
+dbus.notify_send(
```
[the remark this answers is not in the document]
Elv13: It is in `tests/examples/shims/dbus.lua` and has been there for years. Just like most of the shims, it isn't compliant with the real code but is the minimal implementation to execute the code it aims to test without errors. This is a design choice and was done on purpose.

```
function dbus.notify_send(...)
    dbus.emit_signal("org.freedesktop.Notifications", ...)
end
```

tests/examples/naughty/colors.lua
```diff
@@ -0,0 +1,19 @@
...
+<i>notification</i>
+]]
+
+require("naughty").notify {
+    title        = "Hello world!",
+    text         = text,
+    icon         = beautiful.icon,
+    bg           = "#0000ff",
+    fg           = "#ff0000",
+    fond         = "verdana 14",
```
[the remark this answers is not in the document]
Elv13: Opps, the fixup was in the wrong commit.

lib/naughty/core.lua
```diff
@@ -380,6 +399,14 @@ local function cleanup(self, reason)
         n.idx = k
     end
 
+    -- Remove from the global active list.
+    for k, n in ipairs(naughty._active) do
+         if n == self then
+            table.remove(naughty._active, k)
+            naughty.emit_signal("property::active")
+         end
+    end
```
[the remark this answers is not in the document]
Elv13: The logic was that the number of active notification never gets that large. Also keeping a inverted table with only the notification as key and index as value would require to update the most of the list entries every time a notification gets destroyed anyway. This would not be fully linear since some long running notification (index) will eventually bubble up which would mean their indices would remain stable for longer. However I am not sure this is worth the extra complexity.

Keeping track of the index for each "position" as the old implementation did ended up causing half of the regressions we had so far. This could be modified if it ever becomes a problem, but right not I believe using other strategies wont be faster. If some day it becomes common to have hundreds of active notifications (smartphone style), then the rational will no longer be valid.

tests/examples/shims/screen.lua
Elv13: ok

lib/wibox/container/margin.lua
Elv13: I made the table more permissive instead (like for the `c:geometry()` and other similar APIs.

lib/awful/widget/common.lua
Elv13: > Isn't that kind of a backwards-incompatible change in the default?

Given you already proposed a change to the background `clip`, I see this has the logical follow up. Just as the `clip`, yes, it is a behavior change. Just as the clip, I think it falls into the "this is what everybody probably wants anyway" category?

> Either GitHub shows me the commits in a wrong order, or you use this strategy thing before it exists.

GitHub commits for updated PRs are not ordered.

lib/wibox/container/background.lua
Elv13: It is used in the line directly below.

Elv13:
I rebased this and fixed the luacheck issues in each individual commit. This is now green. @blueyed can you do your magic for the typos and doc stuff? Would you have some time in the near future or should we look for alternatives? After over 30 pull requests for this project in the last 2 years, this one is second to last. Once this is merged, what remains is introducing `gears.matching` filters and making this play nice with the preset.

Once done, the rewrite me (API) and @psychon (Gio.GDBus) did will be complete. It will pave the way to then add support for the version 1.2 of the spec, better integration with the clients and better keybinding support for the notifications. Also having better integration with web notifications, mobile notifications (the gnome/kde-connect notification bridge) and implementation specific extensions (like task progress report and Cannonical/KDE systray protocol) can be "fun" projects leveraging the new notification API.

Otherwise merging this isn't very scary. It will stay an experimental feature until it is mature enough to be enabled by default.

Elv13:
Can anyone review that? Otherwise it will get merged in its typo filled glory. At some point I need to push the button, because it's blocking the next one, which is blocking v4.4 from being released. While releasing in April 2019 obviously didn't work, that doesn't mean it has to take a year to get v4.4 out of the door.

docs/03-declarative-layout.md
```diff
@@ -28,6 +28,15 @@ configurable rules.
 
 @DOC_layout_WIDGET_LIST@
 
+### Other
+
+The notifications also have their own widgets.
```
Veratil:
```diff
-The notifications also have their own widgets.
+Notifications also have their own widgets.
```

lib/naughty/layout/box.lua
```diff
@@ -0,0 +1,261 @@
+----------------------------------------------------------------------------
+--- A notification popup widget.
+--
+-- By default, the box is composed of many other wisgets:
```
Veratil: *widgets

docs/03-declarative-layout.md
```diff
@@ -424,3 +436,27 @@ Code:
 
     s.mywibox : setup (three_circle)
 
+### Instantiation rules
+
+Whenever it can, Awesome tries to be asynchronous. This can take various form
+depending on the situation. For example, the `connect_signal` method allows to
+execute code when an event arrives. `awful.screen.connect_for_each_screen` also
+allow to instantiate various elements when a new screen is added. In the later
```
Veratil: *allows

docs/03-declarative-layout.md
```diff
@@ -424,3 +436,27 @@ Code:
...
+case, it is why some widgets are added as properties to other objects instead of
+being global variables like in previous versions of Awesome.
+
+However, there is a case where this isn't enough an another abstract widget has
```
Veratil: *and another

docs/03-declarative-layout.md
```diff
@@ -424,3 +436,27 @@ Code:
...
+to be used. This concept is called the `widget_template` and is an optional
+property of many widgets such as the `awful.widget.taglist`,
+`awful.widget.tasklist` and `naughty.widget.box`. These templates are **table**
```
Veratil: Maybe I'm missing it, but where is `naughty.widget.box`? I only see `naughty.layout.box`.

Also, *are a **table**

docs/03-declarative-layout.md
```diff
@@ -424,3 +436,27 @@ Code:
...
+using the exact same syntax as the declarative widgets, but without the
+`wibox.widget` prefix in front of the curly braces. These template represents
```
Veratil: *These templates represent

docs/03-declarative-layout.md
```diff
@@ -424,3 +436,27 @@ Code:
...
+future widgets that will be created by their parent widget. This is necessary
+for three reasons:
+
+ * The widget must create many instances of the template at different point in
```
Veratil: *points

lib/naughty/core.lua
```diff
@@ -122,6 +122,10 @@ gtable.crush(naughty, require("naughty.constants"))
 -- @property active
 -- @param table
 
+--- True when there is an handler connected to request::display.
```
Veratil: *a handler

lib/naughty/layout/box.lua
```diff
@@ -0,0 +1,261 @@
...
+local popup      = require("awful.popup")
+local awcommon   = require("awful.widget.common")
+local placement  = require("awful.placement")
+local abutton    = require("awful.button")
+
+local default_widget = require("naughty.widget._default")
+
+local box, by_position = {}, {}
+
+-- Init the weak tables for each positions. It is done AOT rather than when
```
Veratil: What is AOT?

lib/naughty/list/actions.lua
```diff
@@ -0,0 +1,324 @@
...
+-- process. `notify-send` doesn't support actions, but `libnotify` based
+-- applications do.
+--
+--@DOC_wibox_nwidget_actionlist_simple_EXAMPLE@
+--
+-- This example has a custom vertical widget template:
+--
+--@DOC_wibox_nwidget_actionlist_fancy_EXAMPLE@
+--
+-- This example has an horizontal widget template and icons:
```
Veratil: *a horizontal

lib/naughty/list/actions.lua
```diff
@@ -0,0 +1,324 @@
...
+
+--- The shape border width for selected actions.
+-- @beautiful beautiful.notification_action_shape_border_width_selected
+-- @param[opt=0] number
+
+--- The action icon size.
+-- @beautiful beautiful.notification_action_icon_size_normal
+-- @param[opt=0] number
+
+--- The action icon size.
```
Veratil: *selected action

lib/naughty/list/notifications.lua
```diff
@@ -0,0 +1,418 @@
...
+
+--- The shape border width for selected notifications.
+-- @beautiful beautiful.notification_shape_border_width_selected
+-- @param[opt=0] number
+
+--- The notification icon size.
+-- @beautiful beautiful.notification_icon_size_normal
+-- @param[opt=0] number
+
+--- The notification icon size.
```
Veratil: *selected notification

lib/naughty/core.lua
Veratil: Also, do we want to backtick `request::display`?

lib/naughty/layout/box.lua
Ryuno-Ki: Ahead Of Time.

lib/naughty/layout/legacy.lua
```diff
@@ -211,6 +211,7 @@ local escape_subs    = { ['<'] = "&lt;", ['>'] = "&gt;", ['&'] = "&amp;" }
 
 -- Cache the markup
 local function set_escaped_text(self)
+    -- Only handle notifications using the legacy layout.
```
psychon: A quick look at the last commit: This commit message and the patch do not fit together. The commit message talks about "if the notification has no widget" while the patch only adds a comment.

Could you re-read your commits and check for things that went "weird" in rebases?

Elv13: I updated the commit messages and dropped that "failed rebase" commit you mentioned. There is some issues with some commits (like typo fixed in the wrong one, etc), but hopefully nothing that will be an issue when using git blame or bisecting. So I am merging this now given there is no other reviews and I want to move on. This has been in the merge queue since summer 2017. Some commits are celebrating their second anniversary in the coming weeks...

Elv13:
The naughty rewrite is now fully merged. The last component is the filter system based on `awful.rules` and is a new feature rather than a replacement API.

So, everybody: Try, this, crash this, report bugs. 2.5k of new code, even with 94% coverage is doomed. It will have issues, including "how did you miss that" caliber ones. That's the way it is and investing more time in testing have a diminishing return. Crowd sourcing tests isn't optimal either and I apologize if you hit bugs.

To enable the system, see the first message. I recommend doing so since, if you are reading this, you already care a lot already. One day, it will be enabled by default and only the users with an older `rc.lua` wont be using it. **This system will not enable itself, you have to explicitly enable it**.

Good luck.

SethBarberee:
@Elv13 you mention `widget_template` for the notifications but the docs aren't updated with that. Was trying to test it and needed my pretty docs to see where to plug in the code for it :P

Elv13:
> @Elv13 you mention widget_template for the notifications but the docs aren't updated with that. Was trying to test it and needed my pretty docs to see where to plug in the code for it :P

"apidoc" is the one tracking git-master, the "api" one tracks the stable release, so it will take some time before it gets there:

https://awesomewm.org/apidoc/popups%20and%20bars/naughty.layout.box.html#widget_template

https://awesomewm.org/apidoc/widgets/naughty.list.notifications.html

PR #2728, opened 2019-03-17
don't replace 'destroy' in existing notification, set destroy_cb

### LINKED ISSUE #2692: infinite loop sometimes using keybind to clear notifications via naughty.destroy_all_notifications

```
$ awesome --version
awesome 4.3 (Too long)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

FWIW this is with git and has been happening after recent-ish changes in past week or two.
Haven't chased it down yet, likely result of naughty rework?

With the following in my rc.lua at the expected location for keybinds:
```
    awful.key({ modkey, }, "\\", naughty.destroy_all_notifications,                                                                                         
              {description = "clear notifications", group = "awesome"})
```

Sometimes using this keybind causes awesome to appear to hang,
with some basic htop/gdb inspection showing an infinite loop.

Unfortunately my lua-fu is weak but perhaps this doesn't quite work right after the change to notification destruction?

https://github.com/awesomeWM/awesome/blob/master/lib/naughty/core.lua#L289

Anyway if someone can take a look to see if there's anything obviously overlooked hopefully that's all there is to it.  Otherwise in the next few days I can try to chase down the introducing commit and perhaps more useful details about where things go wrong.

**Actual result:**

infinite loop

**Expected result:**

no hang or infinite loop

-----

May be duplicate of https://github.com/awesomeWM/awesome/issues/2138 but I'm not sure so filing new to avoid cluttering potentially unrelated ;).

### LINKED ISSUE #2721: Weird issue with the new naughty code

dtzWill:
The value passed as `args.destroy` becomes `destroy_cb`,
so use that name when updating an existing notification.

Fixes #2721.
Might also resolve #2692.

### DIFF
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -207,6 +207,7 @@ capi.dbus.connect_signal("org.freedesktop.Notifications",
 
                 if notification then
                     for k, v in pairs(args) do
+                        if k == "destroy" then k = "destroy_cb" end
                         notification[k] = v
                     end
                 else


Elv13:
Thanks again

psychon:
@Elv13 I didn't look at this in detail yet, but is there some kind of API change or deprecation needed to make this nice and not ugly? E.g. renaming `destroy` to `destroy_cb` in more places?

PR #2738, opened 2019-04-05
Split the reviewed groundwork commits from the notification widgets pull request

Elv13:
These commits are (mostly) the reviewed ones plus the taglist/tasklist ones and the one to update the NEWS.

The other pull request is so large it time out in the CI (and has some luacheck issues in some commits). This has half the commits, so it will greatly reduce the number of commits of the other pull requests once merged.

edit: I added 3 more commit to fix some CI issues. One of them seems to be in an existing test suite while the other 2 are the consequences of deprecating `wibox.container.background.shape_border_width`.

### DIFF
--- a/lib/wibox/container/background.lua
+++ b/lib/wibox/container/background.lua
@@ -15,6 +15,7 @@ local beautiful = require("beautiful")
 local cairo = require("lgi").cairo
 local gtable = require("gears.table")
 local gshape = require("gears.shape")
+local gdebug = require("gears.debug")
 local setmetatable = setmetatable
 local type = type
 local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
@@ -282,8 +283,19 @@ function background:get_border_width()
     return self._private.shape_border_width
 end
 
-background.get_shape_border_width = background.get_border_width
-background.set_shape_border_width = background.set_border_width
+function background.get_shape_border_width(...)
+    gdebug.deprecate("Use `border_width` instead of `shape_border_width`",
+        {deprecated_in=5})
+
+    return background.get_border_width(...)
+end
+
+function background.set_shape_border_width(...)
+    gdebug.deprecate("Use `border_width` instead of `shape_border_width`",
+        {deprecated_in=5})
+
+    background.set_border_width(...)
+end
 
 --- When a `shape` is set, also draw a border.
 --
@@ -312,8 +324,19 @@ function background:get_border_color()
     return self._private.shape_border_color
 end
 
-background.get_shape_border_color = background.get_border_color
-background.set_shape_border_color = background.set_border_color
+function background.get_shape_border_color(...)
+    gdebug.deprecate("Use `border_color` instead of `shape_border_color`",
+        {deprecated_in=5})
+
+    return background.get_border_color(...)
+end
+
+function background.set_shape_border_color(...)
+    gdebug.deprecate("Use `border_color` instead of `shape_border_color`",
+        {deprecated_in=5})
+
+    background.set_border_color(...)
+end
 
 function background:set_shape_clip(value)
     if value then return end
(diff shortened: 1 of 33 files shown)


Elv13:
@dtzWill thanks for testing the other pull request. The fix for the issue you reported is applied here.

@psychon I would like your opinion on the `awful.widget.common` border strategy issue you mentioned in the previous review.

To everyone else, please review. Since most of these commits are already reviewed, I will merge this next weekend if no one else steps up and find problems.

psychon:
> @psychon I would like your opinion on the awful.widget.common border strategy issue you mentioned in the previous review.

Sorry, I have no idea what you are talking about. Which other review? GitHub does not show one.
I took a quick look at https://github.com/awesomeWM/awesome/pull/2738/files and `awful.widget.common` is changed from `shape_border_width` to `border_width`. Is that a fix for some issue that I mentioned?

lib/wibox/container/background.lua
```diff
@@ -284,6 +350,21 @@ function background:get_shape_clip()
...
 end
 
+--- How the border width affects the contained widget.
+--
+-- The valid values are:
+--
+-- * *none*: Just apply the border, do not affect the content size (default).
+-- * *inner*: Squeeze the size of the content by the border width.
+--
+-- @property border_strategy
```
actionless: thanks a lot for this!

Elv13:
Umm the CI passed before merging but failed after once merged... I am investigating.

psychon:
> lua: /usr/local/share/lua/5.3/lgi/record.lua:118: bad argument #3 to 'field' (-inf is out of <-2147483648.0, 2147483647.0>)

It's GDBus complaining about broken input O.o

Elv13:
This will require monkey patching the gdbus part away. The documentation examples are supposed to be running in sandboxes. Given the old dbus API was shimmed, it "used to be fine".

psychon:
Uhm. I don't see how this is related to GDBus. According to the traceback...
```
Scanning dependencies of target run-tests-examples-naughty-actions.lua
Result: 1
Error:
lua: /usr/share/lua/5.2/lgi/record.lua:118: bad argument #3 to 'field' (-inf is out of <-2147483648, 2147483647>)
stack traceback:
	[C]: in function 'field'
	/usr/share/lua/5.2/lgi/record.lua:118: in function </usr/share/lua/5.2/lgi/record.lua:92>
	(...tail calls...)
	[C]: in function '__newindex'
	/usr/share/lua/5.2/lgi/record.lua:178: in function </usr/share/lua/5.2/lgi/record.lua:162>
	(...tail calls...)
	/tmp/awesome/lib/wibox/hierarchy.lua:189: in function 'hierarchy_update'
	/tmp/awesome/lib/wibox/hierarchy.lua:222: in function 'update'
	/tmp/awesome/lib/wibox/hierarchy.lua:208: in function 'new'
	/tmp/awesome/lib/wibox/widget/init.lua:43: in function 'draw_to_cairo_context'
	/tmp/awesome/lib/wibox/widget/init.lua:56: in function 'draw_to_svg_file'
	/tmp/awesome/tests/examples/naughty/template.lua:51: in main chunk
	[C]: in ?
make[4]: *** [CMakeFiles/run-tests-examples-naughty-actions.lua.dir/build.make:57: CMakeFiles/run-tests-examples-naughty-actions.lua] Fehler 1
```
This is failing in `wibox.hierarchy`. Specifically, the line is:
```lua
        region:union_rectangle(cairo.RectangleInt{
            x = new_x, y = new_y, width = new_width, height = new_height
        })
```
So, something wants to be redrawn in an area that is larger than representable by an `int`. I don't see how this should be related to GDBus or worked before, but I cannot investigate right now.

Elv13:
Because of the (working) codepath from the dbus shims vs. non-working "this does nothing so the drawable is now math.inf". I got a semi working trick to fix this, but it needs more work and its sunny right now.

psychon:
Ahh, sorry. Now I get it. Previously, `dbus.send_notification` was able to make `naughty` handle the notification. Now, it silently does nothing since `naughty` listens via `GDBus` instead of on the old `dbus` object. The actual error message is merely a side effect of this underlying problem.

For a work-around, one could make something like `naughty.dbus._internal_send_notification` accessible to be able to inject something.... I guess.

Issue #2729, opened 2019-03-20
Notification actions disappering when replaces_id is set


actionless:
after clicking 2-3 times on this notification example: https://web-push-book.gauntface.com/demos/notification-examples/

first time notification have actions, but clicking on them returns trace what `trigger` method undefined

next times notification replaces previous but without actions at the bottom

![2019-03-06--1551849896_3840x2160_scrot](https://user-images.githubusercontent.com/1655669/53858138-3ddf2c80-3fd9-11e9-80fd-267c362956f9.png)

![2019-03-06--1551849905_3840x2160_scrot](https://user-images.githubusercontent.com/1655669/53858142-420b4a00-3fd9-11e9-971a-6154e2197e85.png)

_Originally posted by @actionless in https://github.com/awesomeWM/awesome/issues/2692#issuecomment-469972615_

psychon:
Can you run `dbus-monitor` in a terminal and capture the output for this?

actionless:
and just to make it more clear: the problem reproducible also with some other chrome notifications, but not all of them, though the ones from this website are always triggering the behavior

here's the captured dbus-monitor:
```

method call time=1553232784.999150 sender=:1.52124 -> destination=org.freedesktop.Notifications serial=13 path=/org/freedesktop/Notifications; interface=org.freedesktop.Notifications; member=Notify
   string "Google Chrome"
   uint32 0
   string "file:///tmp/.com.google.Chrome.Leg8o4"
   string "Web Push Book"
   string "web-push-book.gauntface.com

This would be the body text of the notification.
It can hold two lines of text."
   array [
      string "0"
      string "Download Book"
      string "default"
      string "Activate"
      string "settings"
      string "Settings"
   ]
   array [
      dict entry(
         string "urgency"
         variant             uint32 1
      )
      dict entry(
         string "desktop-entry"
         variant             string "google-chrome"
      )
      dict entry(
         string "image_path"
         variant             string "/tmp/.com.google.Chrome.4g2bog"
      )
      dict entry(
         string "image-path"
         variant             string "/tmp/.com.google.Chrome.4g2bog"
      )
   ]
   int32 25000
method return time=1553232785.010725 sender=:1.44339 -> destination=:1.52124 serial=153 reply_serial=13
   uint32 183


[...]

PR #2740, opened 2019-04-06
Fix 2 more notification issues

Elv13:
(and for the `naughty.constants`, it is already added by `core.lua`. Not having a `require` is "normal". It was split into a different file to avoid some circular `require`s).

### DIFF
--- a/lib/naughty/layout/legacy.lua
+++ b/lib/naughty/layout/legacy.lua
@@ -400,8 +400,8 @@ function naughty.default_notification_handler(notification, args)
             local action_width = w + 2 * margin
 
             actionmarginbox:buttons(gtable.join(
-                button({ }, 1, function() action:trigger() end),
-                button({ }, 3, function() action:trigger() end)
+                button({ }, 1, function() action:invoke() end),
+                button({ }, 3, function() action:invoke() end)
             ))
             actionslayout:add(actionmarginbox)
 
--- a/lib/naughty/init.lua
+++ b/lib/naughty/init.lua
@@ -9,6 +9,7 @@ if dbus then
     naughty.dbus = require("naughty.dbus")
 end
 
+naughty.action = require("naughty.action")
 naughty.layout = require("naughty.layout")
 naughty.notification = require("naughty.notification")
 


lib/naughty/layout/legacy.lua
```diff
@@ -400,8 +400,8 @@ function naughty.default_notification_handler(notification, args)
             local action_width = w + 2 * margin
 
             actionmarginbox:buttons(gtable.join(
-                button({ }, 1, function() action:trigger() end),
-                button({ }, 3, function() action:trigger() end)
+                button({ }, 1, function() action:invoke() end),
+                button({ }, 3, function() action:invoke() end)
```
psychon: Uhm... Since I could never reproduce #2729... is something hiding some error here, or did I just "do it wrong" in trying to reproduce this? In the later case: :+1: and otherwise :-1:

actionless:
i'll check it later today, thanks!

actionless:
nope, traceback changed, but the problem is not fixed:

```lua
2019-04-06 18:57:08 W: awesome: luaA_dofunction:78: error while running function!
stack traceback:
        /usr/share/awesome/lib/naughty/action.lua:101: in function 'invoke'
        /usr/share/awesome/lib/naughty/layout/legacy.lua:403: in function 'press'
        /usr/share/awesome/lib/awful/button.lua:47: in function </usr/share/awesome/lib/awful/button.lua:47>
        [C]: in function 'emit_signal'
        /usr/share/awesome/lib/wibox/widget/base.lua:383: in function 'func'
        /usr/share/awesome/lib/gears/object.lua:148: in function 'emit_signal'
        /usr/share/awesome/lib/wibox/drawable.lua:431: in function </usr/share/awesome/lib/wibox/drawable.lua:426>
error: /usr/share/awesome/lib/naughty/action.lua:101: Cannot invoke an action without a notification
```

psychon:
Can someone explain to me how this is supposed to work?
The error happens when `self._private.notification` is `nil`. The only way to set that field is through the `naughty.action` constructor (argument `notification`) or `set_notification`. The only callers of the constructor are in `naughty.dbus` and `naughty.notification` and they provide `position` and `name` as parameter. Finding uses of the setter is a bit harder, but grepping for `set_notification` and `notification =` does not find anything.

actionless:
@psychon but what happens to you when yuo open in chromium that website, calling notification and press one of its actions, let's say "Download E-Book"?

psychon:
@actionless It doesn't get past Chromium asking me if I want some web page to be able to show notifications. Because it's a web page and I do not.

lib/naughty/layout/legacy.lua
Elv13: This doesn't fix #2729 since I currently cannot reproduce it either. In my case Chromium fallback to its built-in notification for the problematic example even when the register key is set to use the native notification. Firefox doesn't display any notification for that example (not even in `dbus-monitor`, but do for the other examples).

The issue this commit solves was introduced during a rebase. Since the legacy notifications are hard to test (at least without playing with the hierarchy + fake_input), this went unnoticed.

Elv13:
@actionless Please investigate, I cannot reproduce.

actionless:
@Elv13 

it seems it would require some logical change:

currently when you initializing an `action` you need instance of `notification`

and when you initializing `notification` you need `args.actions`


i've managed to fix it locally with this hack:

```diff
--- lib/naughty/dbus.lua        2019-04-06 21:36:17.191213289 +0200
+++ /usr/share/awesome/lib/naughty/dbus.lua     2019-04-06 21:38:13.551108079 +0200
@@ -214,6 +214,10 @@
                     notification = nnotif(args)
                 end

+                for _, a in ipairs(args.actions) do
+                    a._private.notification = notification
+                end
+
                 return "u", notification.id
             end
```

actionless:
but the workaround above resolves only half of https://github.com/awesomeWM/awesome/issues/2729

actions are working now

however when handling replaces_id `.actions` are still not added to the new notification

psychon:
@actionless `naughty.action` objects have a `set_notification` method, so no need to mess with `_private` of something else here. Also, I'm not sure if actions are supposed to be movable between notifications, but the API seems to imply the possibility, so:
```diff
diff --git a/lib/naughty/action.lua b/lib/naughty/action.lua
index a16a95706..b4907b410 100644
--- a/lib/naughty/action.lua
+++ b/lib/naughty/action.lua
@@ -86,6 +86,10 @@ for _, prop in ipairs { "name", "icon", "notification" } do
     end
 
     action["set_"..prop] = function(self, value)
+        local old
+        if prop == "notification" then
+            old = self._private.notification
+        end
         self._private[prop] = value
         self:emit_signal("property::"..prop, value)
 
@@ -93,6 +97,9 @@ for _, prop in ipairs { "name", "icon", "notification" } do
         if self._private.notification then
             self._private.notification:emit_signal("property::actions")
         end
+        if old then
+            old:emit_signal("property::actions")
+        end
     end
 end
 
```
Somehow I feel like this whole interaction thingie is only weekly defined. Locally, I'd say a notification owns a number of actions, but then the action should not need to know about its notification. Instead, I'd expect things to be similar to what we do in `wibox`: The notification instance connects to the necessary signals of the instance and updates itself.
(While right now, it seems that instead the actions are responsible for emitting `property::actions` on the notification.)

actionless:
@psychon i wasn't saying about moving actions from old notification to new but about adding actions of new notification

Elv13:
> Also, I'm not sure if actions are supposed to be movable between notifications, but the API seems to imply the possibility

It isn't currently defined. Do you think they should? In that can we can remove the notification property and watch for the more signals instead of having the action update its notification. I am not sure there's a valid use case for sharing actions across notifications, but it might be cleaner from a design pov.

dtzWill:
> @actionless `naughty.action` objects have a `set_notification` method, so no need to mess with `_private` of something else here. Also, I'm not sure if actions are supposed to be movable between notifications, but the API seems to imply the possibility, so:
> 
> ```diff
> diff --git a/lib/naughty/action.lua b/lib/naughty/action.lua
> index a16a95706..b4907b410 100644
> --- a/lib/naughty/action.lua
> +++ b/lib/naughty/action.lua
> @@ -86,6 +86,10 @@ for _, prop in ipairs { "name", "icon", "notification" } do
>      end
>  
>      action["set_"..prop] = function(self, value)
> +        local old
> +        if prop == "notification" then
> +            old = self._private.notification
> +        end
>          self._private[prop] = value
>          self:emit_signal("property::"..prop, value)
>  
> @@ -93,6 +97,9 @@ for _, prop in ipairs { "name", "icon", "notification" } do
>          if self._private.notification then
>              self._private.notification:emit_signal("property::actions")
>          end
> +        if old then
> +            old:emit_signal("property::actions")
> +        end
>      end
>  end
>  
> ```
> 
> Somehow I feel like this whole interaction thingie is only weekly defined. Locally, I'd say a notification owns a number of actions, but then the action should not need to know about its notification. Instead, I'd expect things to be similar to what we do in `wibox`: The notification instance connects to the necessary signals of the instance and updates itself.
> (While right now, it seems that instead the actions are responsible for emitting `property::actions` on the notification.)

I've been running with this patch applied for quite some time-- more or less since you first posted it I believe :).

I honestly don't remember what breaks if I drop it, but I remember last time it was severely painful enough that I immediately re-applied and decided this was a battle for a different day.  If it's unclear I can try again, but since it was written for a purpose I bet you know the sorts of problem(s) it is intended to fix :).

---------

Is it expected to need this change on top of latest master? If so, would a PR requesting such make sense/be controversial? I'm almost always referring to notifications created via dbus which despite being referred to around here as "legacy" for notifications seems to be the sole way everything non-awesome-wm might create/interact with notifications.  I guess maybe via `awesome-client` or something? I'm honestly a bit shocked at the state of notifications, in the sense I can't help but feel I'm the oddball using things wrong and wouldn't have these problems if I used the stack everyone else is (?).

---- 

Currently updating from 99e81c097affb570d144a1f9e0ea64872e242540, so maybe the answer lies in the commits between that and master.... ? fingers crossed ;)

psychon:
@Elv13 ^
(I have no idea about this code and thus will not turn the above patch into a PR; also, I do not have much programs sending DBus notifications)

PR #2741, opened 2019-04-06
Multi keygrabber

Elv13:


psychon:
Does this need a new test?

alfunx:
Thanks for this!

@psychon Where would a test for this go? To `tests/examples/text/awful/keygrabber/`? Or to `spec/awful/`? I could write one based on my reproducer.

Elv13:
`tests/examples/text/awful/keygrabber/` or `tests/test-keybindings.lua`. It isn't a great fit for the `examples` suite unless you can turn it into something relevant for the documentation (which would be great). The `spec` is going to fight you a bit since this calls into a bunch of `capi` and you will have to shim them in some non-trivial ways. So creating a new integration test suite is a viable option.

Issue #2712, opened 2019-02-28
keygrabber: Setup multiple keygrabbers


alfunx:
If you set up multiple keygrabbers, only the first one seems to be used. The first keygrabber has `Mod4+a` as `root_keybinding`, the second one has `Mod4+b`. The first one works as expected. However if you press `Mod4+b` for the second one (while the first keygrabber is **not** running), the first keygrabber is started and the notifications for the first one are shown. I'm not sure if I'm using it correctly though, and didn't yet look at the code of `awful.keygrabber`.

Append to the default `rc.lua`:

```lua
local a = awful.keygrabber {
    stop_key = "Escape",
    start_callback = function() naughty.notification { message = "start 1" } end,
    stop_callback  = function() naughty.notification { message = "stop 1"  } end,
    root_keybindings = {
        { { "Mod4" }, "a", function() end },
    },
    keybindings = {
        { {        }, "x", function()
            naughty.notification { message = "in grabber 1" }
        end },
    },
}

local b = awful.keygrabber {
    stop_key = "Escape",
    start_callback = function() naughty.notification { message = "start 2" } end,
    stop_callback  = function() naughty.notification { message = "stop 2"  } end,
    root_keybindings = {
        { { "Mod4" }, "b", function() end },
    },
    keybindings = {
        { {        }, "x", function()
            naughty.notification { message = "in grabber 2" }
        end },
    },
}
```

By the way, if `function() end` is omitted in the `root_keybindings` is omitted, the keygrabber is not triggered (the example for modal keybindings in the docs suggests this should work).

Elv13:
That's currently a non-feature. A keygrabber grabs the input. That includes grabbing the input away from another keygrabber.

That being said, adding support for "keygrabber groups" isn't out of the question. I don't plan to do it myself, but if you jump into the code, I am sure you can come up with a patch for this rather easily. The current default behavior has to be preserved, but this could be a new option. There's definitely valid use cases for this. That should simply be added by looping all keygrabbers in a group in `keygrabber.lua: local function runner(self, modifiers, key, event)` and adding the `group` (string) propety + black magic to internally keep the list of grabers "alive" (table `{__mode = 'v'}`) in each group.

alfunx:
> A keygrabber grabs the input. That includes grabbing the input away from another keygrabber.

[...]

PR #2744, opened 2019-04-07
keygrabber: Add test for multiple instance creation (#2712)

alfunx:
Add test for #2712 / #2741.

Putting the asserts in different step-functions didn't work, and one step-function without `count` didn't do it either. Am I missing something or is this fine?

### DIFF
--- a/tests/test-keygrabber.lua
+++ b/tests/test-keygrabber.lua
@@ -1,0 +1,73 @@
+-- Test for #2712: It should be possible to create multiple instances of
+-- keygrabber during the same loop.
+
+local runner = require("_runner")
+local awful = require("awful")
+
+local keygrabber_a_active = false
+local keygrabber_b_active = false
+
+local steps = {
+    function()
+        awful.keygrabber {
+            keybindings = {
+                {{}, "a", function() keygrabber_a_active = true end},
+            },
+            stop_key = "Escape",
+            stop_callback = function() keygrabber_a_active = false end,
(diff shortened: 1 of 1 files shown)


Elv13:
Thanks a lot!

For the next time it might be more scalable if it was done in 6 functions instead of an if/elseif on `count`, but this isn't important. Thanks again for writing tests for bugs I totally caused.

Elv13:
I merged the other PR so you can rebase this on master (for the test to pass)

alfunx:
No worries!

> For the next time it might be more scalable if it was done in 6 functions instead of an if/elseif on `count`, but this isn't important.

That's what I meant, I added a WIP commit. However the asserts fail this way, I suspect it's due to the functions being called in parallel practically (https://github.com/awesomeWM/awesome/blob/master/tests/_runner.lua#L110). Is there a nicer way of doing this?

Elv13:
you can call `awesome.sync()` or simply check with `if` instead of `assert`. If the function returns `nil`, it will be executed again until it timeout. This helps getting rid of the race conditions. Note that using `count` doesn't mean your code is immune to races. They are just less likely because there's a longer delay. It could fail from time to time, just like this one I recently found https://github.com/awesomeWM/awesome/pull/2738/commits/68f6a567d1887e3039c2f16cabfbdeec70d7aded .

alfunx:
Thanks, that makes sense - updated and squashed!

PR #2746, opened 2019-04-13
Make Guake float

SoniEx2:


psychon:
Can we please stop matching based on window names and instead use e.g. class instead? (Also: Why does `xev` not set `WM_CLASS`? :-( )

SoniEx2:
guake preferences has the same class as guake, so you can't tell the two apart unless you match on name or multiple properties or something.

psychon:
Do they also have the same instance? Also, why do you want guake to float but its preference window should not float?

SoniEx2:
because guake is a drop-down terminal. you want that to float so the drop-down works properly, otherwise it'll just follow the tiling window placement, which kinda defeats the point of using a drop-down terminal with transparent/translucent backdrop that lets you vim while you peertube.

its preferences window is just a plain old window, otoh.

Sorky:
I don't believe it needs it - All that I have for Guake...

	-- Remove Guake titlebar
	{	rule = { name = "Guake!" },
		properties = {	titlebars_enabled = false}
	},

I personally believe the default rules should be basic examples and you should add application specific ones like I did above

PR #2751, opened 2019-04-14
Fix the broken test due to the "soft" merge conflict of two notification pull requests.

Elv13:
One pull request relied on the fact that the shims for `capi.dbus` were capable of creating notifications object (using the default `naughty.dbus` code) while the other removed support for receiving notifications using `capi.dbus`. Gio is, of course, the better way to do this, but it is also a bit harder to test because the networking is done by a library we have no control upon. To mitigate this, this pull request move some previously private methods to an exported private API. It also upgrade the shims to call those methods with a faked list of function arguments. A third commit fix yet another regression caused by my rewrite.


Elv13:
(this is ready, the test pass if this is squash-merged)

psychon:
Looks good enough to me, so :+1: 

Can you teach me some Lua? What's up with `function dbus.notify_send(_--[[appdata]], ...)`? That parameter confuses me (and I do not have the time right now to scan through the Lua docs to figure out what this is).

Elv13:
> Can you teach me some Lua? What's up with `function dbus.notify_send(_--[[appdata]], ...)`? That parameter confuses me (and I do not have the time right now to scan through the Lua docs to figure out what this is).

It's the syntax for inline comments "near" the syntax for unused function arguments. This is the equivalent of the C `my_function(int /*unused_arg_name*/)` to mute `-Wunused`

snippet: 


    print("foo"--[[ddddd]])

result:

    foo

an alternative syntax would have been `--luacheck: no unused args`

PR #2753, opened 2019-04-18
Draw slider's bar with different color depending on current value

ndgnuh:
`bar_color` will be used to draw the part of the bar which indicates values those are less than current value.
`bar_bg` will be used to draw the rest of the bar.

edit: preview:
image
```lua
wibox.widget {
  bar_bg          = colors.nord5,
  bar_color       = colors.nord14,
  bar_height      = dpi(3),
  bar_shape       = gears.shape.rounded_bar,

  handle_color    = colors.nord14,
  handle_shape    = gears.shape.circle,
  
  forced_width    = dpi(160),
  widget          = wibox.widget.slider
}
```


Elv13:
Hi,

Please update your patch to be based on the original source code. The one you submitted is based on the post-processed code with extra documentation inserted.

ndgnuh:
Okay, i will send another pull request later.

ndgnuh:
By "original source code", did you mean [this one](https://github.com/awesomeWM/awesome/blob/master/lib/wibox/widget/slider.lua)?

psychon:
I can't see the changes (there are now zero commits in this PR), but to answer your question: Yes.

There is the "real source code", which is what your link points to. While building awesome, things like `@DOC_wibox_widget_slider_handle_color_EXAMPLE@` are replaced with something else. The resulting file is what is installed. Thus, I guess your PR removed things like `@DOC_wibox_widget_slider_handle_color_EXAMPLE@` and instead used the version of the file that is actually installed.

ndgnuh:
I force pushed the repo so the commit disappeared. But i have opened another PR, if you are interested then [here it is](https://github.com/awesomeWM/awesome/pull/2754).

PR #2759, opened 2019-04-27
Improve the shims

Elv13:
I was rebasing some of my branches and some "regressions" due to various merged changes were easier to debug with this little change. It catches attempts to use invalid screens. It's useful for documentation tests that add and remove screens.

### DIFF
--- a/tests/examples/shims/mouse.lua
+++ b/tests/examples/shims/mouse.lua
@@ -53,9 +53,14 @@ function mouse.push_history()
     mouse.history = {}
 end
 
+local forced_screen = nil
+
 return setmetatable(mouse, {
     __index = function(self, key)
         if key == "screen" then
+            if forced_screen and screen._deleted[forced_screen] then
+                forced_screen = nil
+            end
             return screen[1]
         end
         local h = rawget(mouse,"_i_handler")
@@ -63,10 +68,13 @@ return setmetatable(mouse, {
             return h(self, key)
         end
     end,
-    __newindex = function(...)
+    __newindex = function(_, k, v)
         local h = rawget(mouse,"_ni_handler")
-        if h then
-            h(...)
+        if k == "screen" then
+            -- This will assert if the screen is invalid
+            forced_screen = v and screen[v] or nil
+        elseif h then
+            h(_, k, v)
         end
     end,
 })
--- a/tests/examples/shims/mouse.lua
+++ b/tests/examples/shims/mouse.lua
@@ -61,7 +61,14 @@ return setmetatable(mouse, {
             if forced_screen and screen._deleted[forced_screen] then
                 forced_screen = nil
             end
-            return screen[1]
+
+            -- Using capi.mouse.screen is *not* supported when there is zero
+            -- screen. Nearly all the code uses `mouse.screen` as its ultimate
+            -- fallback. Having no screens is tolerated during early
+            -- initialization and that's it.
+            return screen.count() > 0 and screen[1] or assert(
+                false, "Calling `mouse.screen` without screens isn't supported"
+            )
         end
         local h = rawget(mouse,"_i_handler")
         if h then
--- a/tests/examples/shims/mouse.lua
+++ b/tests/examples/shims/mouse.lua
@@ -62,6 +62,13 @@ return setmetatable(mouse, {
                 forced_screen = nil
             end
 
+            for s in screen do
+                if coords.x > s.geometry.x and coords.x < s.geometry.x +s.geometry.width
+                  and coords.y > s.geometry.y and coords.y < s.geometry.y +s.geometry.height then
+                    return s
+                end
+            end
+
             -- Using capi.mouse.screen is *not* supported when there is zero
             -- screen. Nearly all the code uses `mouse.screen` as its ultimate
             -- fallback. Having no screens is tolerated during early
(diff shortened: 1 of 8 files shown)


tests/examples/shims/screen.lua
```diff
@@ -34,32 +34,27 @@ local function create_screen(args)
 
     local wa = args.workarea_sides or 10
 
-    -- This will happen if `clear()` is called
-    if mouse and not mouse.screen then
-        mouse.screen = s
-    end
```
psychon: I wouldn't expect a commit with commit message "shims: Fix indentation" to remove four lines from a file....

tests/examples/shims/mouse.lua
```diff
@@ -53,20 +53,28 @@ function mouse.push_history()
     mouse.history = {}
 end
 
+local forced_screen = nil
+
 return setmetatable(mouse, {
     __index = function(self, key)
         if key == "screen" then
+            if forced_screen and screen._deleted[forced_screen] then
```
psychon: What's the difference between `screen._deleted[s]` and `s.valid`? I would have expected that this patch assumes that no one "messes with" `s.valid` and just add `assert(s.valid)` in some places...?

tests/examples/shims/screen.lua
```diff
@@ -85,8 +87,36 @@ function screen._get_extents()
...
 
+-- The way the `screen` module is used to store both number and object is
+-- problematic since it can cause invalid screens to be "resurrected".
+local function catch_invalid(_, s)
+    -- The CAPI implementation allows `nil`
+    if s == nil or type(s) == "string" then return end
+
+    assert(screen ~= s, "Some code tried (and failed) to shadow the global `screen`")
+
+    -- Valid numbers wont get this far.
```
psychon: Why not? Only `nil` and `string` were checked so far.

tests/examples/shims/screen.lua
```diff
@@ -85,8 +87,36 @@ function screen._get_extents()
...
+    assert(s.geometry, "The object is not a screen")
+    assert(s.outputs, "The object is not a screen")
+
+    assert((not screen._deleted[s]) or (not s.valid), "The shims are broken")
+
+    assert(not screen._deleted[s], "The screen "..tostring(s).."has been deleted")
+
+    -- Other errors. If someone place an object in the `screen` module, it wont
+    -- get this far. So the remaining cases are probably bugs.
+    assert(false)
```
psychon: The last four lines of `assert` are basically dead code, aren't they? This code will end up in `assert(false)` anyway, so is the point just do produce a better error message?

tests/examples/shims/screen.lua
```diff
@@ -85,8 +87,36 @@ function screen._get_extents()
     return xmax.geometry.x+xmax.geometry.width, ymax.geometry.y+ymax.geometry.height
 end
 
+-- The way the `screen` module is used to store both number and object is
+-- problematic since it can cause invalid screens to be "resurrected".
+local function catch_invalid(_, s)
```
psychon: I do not understand this method. It is equivalent to the following (minus producing different error messages):
```lua
local function catch_invalid(_, s)
    if s == nil or type(s) == "string" then return end
    error("Well, error") -- The only other return (type(s) == "number") is after an assert(type(s) == "table") and thus is dead code.
end
```

tests/examples/shims/screen.lua
Elv13: Because if you had 42 screens, remove them all, then add 2 screens, if you `screen[42]`, your code probably cached the screen number and that's evil (but common for older code written with v3.5 in mind).

tests/examples/shims/screen.lua
Elv13: >  (minus producing different error messages):

That's the whole point

tests/examples/shims/screen.lua
Elv13: whoops, rebase conflict issue, it belongs to the other commit. That's now handled by the mouse metadable instead of the screen shim.

tests/examples/shims/screen.lua
Elv13: > so is the point just do produce a better error message?

Yes

tests/examples/shims/mouse.lua
Elv13: There's a slim way to get a client (with its own `.valid`) interpreted as a screen if someone adds an `outputs` property. The `_deleted` table is harder to corrupt.

tests/examples/shims/screen.lua
Elv13: done

Elv13:
Updated and added more shim commits from my stash.

Elv13:
This PR is now "ready" to merge. It took many iterations to get to the point where all commits individually pass all tests. This was probably a waste of time and squashing the whole thing would have been a better choice, but it's done... Some of the commits are a bit weird since they add some code to pass the tests only the for next commit to remove it in favor of a "more correct" alternative.

The goal of this PR is to be able to execute the `awful.rules` examples using the example test framework rather than having them hardcoded in the doc. This will be followed by a PR to introduce the "sequence" documentation template for multi-iteration examples. Then, this will be followed by another PR to move the current `awful.rules` to the new rules module (rather than awful). More PRs follow adding rules to both tags and screens (as mentioned quite a few time already).

These PR also use the code present in this PR and will bring the coverage to 100%. Right now there isn't anything using this code in the examples, so it is uncovered.

Elv13:
last call

PR #2770, opened 2019-05-23
Replace doubled immobilized with correct properties for each immobilized

frioux:


lib/awful/client.lua
```diff
@@ -727,11 +727,10 @@ end
 -- fullscreen, maximized, or horizontally maximized?
 --
 -- This property is read only.
--- @property immobilized
+-- @property is_immobilized_horizontal
```
Elv13: Good catch, but I think the property are called `immobilized_horizontal` and `immobilized_vertical`. The `is_` is removed when the getter becomes a property (at least it should be).

Ref: https://github.com/awesomeWM/awesome/blob/master/lib/gears/object/properties.lua#L40

frioux:
Fixed; thanks

frioux:
@Elv13 I think your comments should be resolved, but I can't figure out a way to mark them as resolved in the GH UI

PR #2773, opened 2019-05-27
Slider: Emit lmb signal. Fixes #1241

### LINKED ISSUE #1241: wibox.widget.slider "button::release" don't work with LMB

```
awesome v3.5.2-2192-g333cd649-dirty (The Fox)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.0
```

```lua
test_slider = wibox.widget {
  bar_color = "#000000",
  handle_color = "#FFFFFF",
  handle_shape = shape.circle,
  value = 0,
  minimum = 0,
  maximum = 100,
  widget = wibox.widget.slider,
}

test_slider:connect_signal("button::release", function(w)
  naughty.notify({text = tostring(w.value)})
  return true
end)
```

**Actual result:**

Don't work with LMB (button_id == 1), only with others (2, 3, 4, 5).

**Expected result:**

Run callback function on LMB (button_id == 1) on "button::release" signal.

unai-ndz:


unai-ndz:
Sorry about the first point, should had rechecked before making the PR.
I can't think of a way to get the modifiers and the more time I spent on this the more I agree with you in point three.

I'm following your suggestions for addressing this issue on #1241 but that was 2 years and a half ago. Maybe there is a better way to do it now?

psychon:
Perhaps some special signals on the slider? `drag::begin` and `drag::end`?

psychon:
Does the slider need a key grabber at all? Is it perhaps possible to do this without grabs? Although the resulting behaviour certainly would be unexpected...
(Just thinking out loud here before I have to go again)

psychon:
(For the use case from #1241, one could also implement some rate-limiting outside of the slider widget instead of spamming `awful.spawn` calls, but I guess that doesn't really count as a solution. For example, one could do `active = true` when spawning a new process and use `awful.spawn.with_line_callback`'s `exit_callback` to set that to `false` once the spawned process is done...)

Elv13:
> I can't think of a way to get the modifiers

There is the top secret `_active_modifiers` since AwesomeWM v4.3. It is how the `awful.keygrabber` gets around the same problem. The widgets under the cursor can be obtained using the `mouse` API.

As for "just implement the expected signals" vs. "add new signals". I would rather implement the documented signals to make the module less unexpected.

PR #2774, opened 2019-05-27
Fix instance existing check for application spawn

worron:
Matcher argument for `spawn.once` and others was ignored in most cases during instance existence check. Now it has higher priority against `startup_id` and should work according docs.

### DIFF
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -547,16 +547,16 @@ local function is_running(hash, matcher)
     local status = spawn.single_instance_manager.by_uid[hash]
     if not status then return false end
 
-    if #status.instances == 0 then return false end
-
-    for _, c in ipairs(status.instances) do
-        if c.valid then return true end
-    end
-
     if matcher then
         for _, c in ipairs(client.get()) do
             if matcher(c) then return true end
         end
+    end
+
+    if #status.instances == 0 then return false end
+
+    for _, c in ipairs(status.instances) do
+        if c.valid then return true end
     end
 
     return false
--- a/tests/test-spawn.lua
+++ b/tests/test-spawn.lua
@@ -37,8 +37,6 @@ app:run {''}
 ]]}
 end
 
-local matcher_called = false
-
 local steps = {
     function()
         -- Test various error conditions. There are quite a number of them...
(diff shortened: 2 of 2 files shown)


lib/awful/spawn.lua
```diff
@@ -279,7 +279,7 @@ local function hash_command(cmd, rules)
     cmd = type(cmd) == "string" and cmd or table.concat(cmd, ';')
 
     -- Do its best at adding some entropy
-    local concat_rules = nil
+    local concat_rules
```
Elv13: I would prefer if `nil` remained. We still have some C code, so the convention is usually to initialize a variable when you plan to use it as-is. Also, while there is some of both, initialized nil are use more often.

tests/test-spawn.lua
```diff
@@ -321,19 +319,47 @@ local steps = {
...
+        return true
+    end,
+    function()
+        -- clean up
+        if #client.get() ~= 1 then return end
+        client.get()[1]:kill()
+        return true
+    end,
+    -- Test rules works with matcher if client doesn't support startup id
+    -- Can this test be performed using '_client' module without external 'xterm' application?
```
Elv13: > -- Can this test be performed using '_client' module without external 'xterm' application?

`_client` support startup notification, which allows properties to be applied without a matcher. `xterm` doesn't support this. So, in practice they will always go down 2 different code paths. In the perfect world, we could use another tiny X11 client with the same-ish size_hints support and get rid of xterm as a test dependency.

What has to be tested:

 * `_client` without a matcher (using startup notifications)
 * `_client` with a matcher to make sure the matcher is used
 * `xterm` with a matcher, because it wont work without one

psychon: If the `xterm` dependency is a problem, we could always just compile our own test program and use that. `test-gravity.c` already does this.

worron: @Elv13 
The question was can `_client` (or its mod) be used  as client without startup notification. I saw [this test](https://github.com/awesomeWM/awesome/blob/master/tests/test-spawn-snid.lua#L33) and suggest that some build-in gtk3 lgi app can emulate `xterm` behavior. But possible I get it wrong.

@psychon 
Not a problem at all, but would be nice get rid of external applications for tests. I personally had no `xterm` nor `xeyes` install. There are some build-in apps done with lgi so having `xterm` extra feels like half done work.  Still not sure if it's worth it.

lib/awful/spawn.lua
worron: Ok, will take in mind.

Also tell me please how PR fixes should be usually organized. Should I amend commit and do force push or leave chain of commit as is?

Veratil: They'll squash the commit before merge.

Elv13:
> Should I amend commit and do force push or leave chain of commit as is?

We use a service called mergify.io to auto-merge when conditions are met (test, coverage, reviews, linting, etc). This service is quite dumb when it comes to squashing vs. merging. Usually, for single commit PRs, I squash them to avoid the "useless" `git log --graph` spagatti mergify used to create. They recently made it possible to properly rebase when merging (`git merge --no-ff`) to make the graph less horrible, but we still have to enable it for AwesomeWM. I think we should switch to always use the `--no-ff` even for single commits PRs to preserve the metadata. This makes non-force-pushed fixups a no-go.

For PRs with more than 1 commit, please always force-push, otherwise I have to fetch them locally and force push myself before accepting to prevent the mergify bot from merging the fixup commits into the master branch. Smaller commits with "atomic" changes are preferred to large commits that do multiple things. Small commits are easier to use with `git bisect` and `git blame`.

ref: https://github.com/Mergifyio/mergify-engine/issues/115

Elv13:
Thanks

Issue #390, opened 2015-08-04
Better SVG support


worron:
Is it possible use vector scaling for SVG icons in awesome?

worron:
@psychon
I made an experimental [widget](https://github.com/worron/redflat/blob/ebbb6433a07fc2ed1455a9d356b0e1c2ad97c674/gauge/svgbox.lua) based on your example. It works and works better than my workarounds used before. But render to temporary file still seems like overhead. Main question: is there any other way translate  pixbuf to cairo surface?

psychon:
Just do `require("lgi").Gdk`. Gdk provides a function `gdk_cairo_set_source_pixbuf` (or something like that) which lgi makes available on cairo contexts. After this you can just use `cr:set_source_pixbuf(some_pixbuf)`.

However, I would like to avoid making awesome depend on Gdk, because using Gdk from a window manager doesn't really work well (too much traps to fall into).

Elv13:
Here is an example how to load GDK/GTK icons asynchronously using the
function mentioned by Uli
https://github.com/Elv13/awesome-configs/blob/master/utils/fd_async.lua#L414

Have fun

worron:
Thank you very much. Now it works as desired.  And example is useful.
I admit it may be not safe globally, but now I'm fully satisfied with this Gdk trick. Hope it will be ok for personal using, until awesome will get foolproof vector scaling feature.

actionless:
can we move it somehow to imagebox to be able to set size automatically?

worron:
It is quite possible, but there are some reasons not to do it.
1) In post above Uli Schlachter spoke against making awesome depend on Gdk.
2) It will conflict with recoloring feature 5806048. In my imagebox modification recolor is going after scaling and it's ok. In your approach icon transformed to recolored surface long before it reach imagebox, and there are no way to proper scaling.

psychon:
Awesome should still do something better by default for SVGs. Perhaps it might even make sense to support animations (GdkPixbuf does!) :-)
Some quick API idea: Add a new object kind (image? I don't know) which has:
- `:query_size()` for some preferred size
- `:get_for_size(width, height)` for getting a surface (this could then render a SVG at the needed size)

This could also be extended to support animations. Just emit some signal when a redraw is needed. The signal is only emitted if `:get_for_size()` was called since the last time it was emitted, so animations which aren't displayed would automatically stop.

Also: https://github.com/awesomeWM/awesome/pull/369#issuecomment-127178507

[...]

PR #2779, opened 2019-06-03
Better svg support

worron:
Use gdk-pixbuf to load cairo surface at requested size.
Related #390.

I gave it some quick testings and seems ok now. Tell me if any fixes or improvements needed.

### DIFF
--- a/lib/wibox/widget/imagebox.lua
+++ b/lib/wibox/widget/imagebox.lua
@@ -6,20 +6,53 @@
 -- @widgetmod wibox.widget.imagebox
 ---------------------------------------------------------------------------
 
+local lgi = require("lgi")
+local cairo = lgi.cairo
+
 local base = require("wibox.widget.base")
 local surface = require("gears.surface")
 local gtable = require("gears.table")
+local gdebug = require("gears.debug")
 local setmetatable = setmetatable
 local type = type
-local print = print
+local math = math
+
 local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
+
+-- Safe load for optional Rsvg module
+local Rsvg = nil
+do
+    local success, err = pcall(function() Rsvg = lgi.Rsvg end)
+    if not success then
+        gdebug.print_warning(debug.traceback("Could not load Rsvg: " .. tostring(err)))
+    end
+end
 
 local imagebox = { mt = {} }
 
+local rsvg_handle_cache = setmetatable({}, { __mode = 'v' })
+
+---Load rsvg handle form image file
+---@tparam string file Path to svg file.
+---@return Rsvg handle
+local function load_rsvg_handle(file)
+    if not Rsvg then return end
+
+    local cache = rsvg_handle_cache[file]
+    if cache then
+        return cache
+    end
+
+    local handle, err = Rsvg.Handle.new_from_file(file)
+    if not err then
+        rsvg_handle_cache[file] = handle
+        return handle
+    end
+end
+
 -- Draw an imagebox with the given cairo context in the given geometry.
 function imagebox:draw(_, cr, width, height)
-    if not self._private.image then return end
-    if width == 0 or height == 0 then return end
+    if width == 0 or height == 0 or not self._private.default then return end
 
     -- Set the clip
     if self._private.clip_shape then
@@ -28,86 +61,101 @@ function imagebox:draw(_, cr, width, height)
 
     if not self._private.resize_forbidden then
         -- Let's scale the image so that it fits into (width, height)
-        local w = self._private.image:get_width()
-        local h = self._private.image:get_height()
-        local aspect = width / w
-        local aspect_h = height / h
-        if aspect > aspect_h then aspect = aspect_h end
-
+        local w, h = self._private.default.width, self._private.default.height
+        local aspect = math.min(width / w, height / h)
         cr:scale(aspect, aspect)
     end
 
-    cr:set_source_surface(self._private.image, 0, 0)
-    cr:paint()
+    if self._private.handle then
+        self._private.handle:render_cairo(cr)
+    else
+        cr:set_source_surface(self._private.image, 0, 0)
+        cr:paint()
+    end
 end
 
 -- Fit the imagebox into the given geometry
 function imagebox:fit(_, width, height)
-    if not self._private.image then
-        return 0, 0
-    end
+    if not self._private.default then return 0, 0 end
+    local w, h = self._private.default.width, self._private.default.height
 
-    local w = self._private.image:get_width()
-    local h = self._private.image:get_height()
-
-    if w > width then
-        h = h * width / w
-        w = width
-    end
-    if h > height then
-        w = w * height / h
-        h = height
-    end
-
-    if h == 0 or w == 0 then
-        return 0, 0
-    end
-
-    if not self._private.resize_forbidden then
-        local aspect = width / w
-        local aspect_h = height / h
-
-        -- Use the smaller one of the two aspect ratios.
-        if aspect > aspect_h then aspect = aspect_h end
-
-        w, h = w * aspect, h * aspect
+    if not self._private.resize_forbidden or w > width or h > height then
+        local aspect = math.min(width / w, height / h)
+        return w * aspect, h * aspect
     end
 
     return w, h
 end
 
+---Apply cairo surface for given imagebox widget
+local function set_surface(ib, surf)
+    local is_surd_valid = surf.width > 0 and surf.height > 0
+    if not is_surd_valid then return end
+
+    ib._private.default = { width = surf.width, height = surf.height }
+    ib._private.handle = nil
+    ib._private.image = surf
+    return true
+end
+
+---Apply RsvgHandle for given imagebox widget
+local function set_handle(ib, handle)
+    local dim = handle:get_dimensions()
+    local is_handle_valid = dim.width > 0 and dim.height > 0
+    if not is_handle_valid then return end
+
+    ib._private.default = { width = dim.width, height = dim.height }
+    ib._private.handle = handle
+    ib._private.image = nil
+    return true
+end
+
+---Try to load some image object from file then apply it to imagebox.
+---@tparam table ib Imagebox
+---@tparam string file Image file name
+---@tparam function image_loader Function to load image object from file
+---@tparam function image_setter Function to set image object to imagebox
+---@treturn boolean True if image was successfully applied
+local function load_and_apply(ib, file, image_loader, image_setter)
+    local image_applied
+    local object = image_loader(file)
+    if object then
+        image_applied = image_setter(ib, object)
+    end
+    return image_applied
+end
+
 --- Set an imagebox' image
 -- @property image
--- @param image Either a string or a cairo image surface. A string is
---   interpreted as the path to a png image file.
+-- @param image This can can be string, cairo image surface, rsvg handle object or nil. A string is
+--   interpreted as the path to an image file. Nil will deny previously set image.
 -- @return true on success, false if the image cannot be used
-
 function imagebox:set_image(image)
+    local setup_succeed
+
     if type(image) == "string" then
-        image = surface.load(image)
-        if not image then
-            print(debug.traceback())
-            return false
+        -- try to load rsvg handle from file
+        setup_succeed = load_and_apply(self, image, load_rsvg_handle, set_handle)
+
+        if not setup_succeed then
+            -- rsvg handle failed, try to load cairo surface with pixbuf
+            setup_succeed = load_and_apply(self, image, surface.load, set_surface)
         end
+    elseif Rsvg and Rsvg.Handle:is_type_of(image) then
+        -- try to apply given rsvg handle
+        setup_succeed = set_handle(self, image)
+    elseif cairo.Surface:is_type_of(image) then
+        -- try to apply given cairo surface
+        setup_succeed = set_surface(self, image)
+    elseif not image then
+        -- nil as argument mean full imagebox reset
+        setup_succeed = true
+        self._private.handle = nil
+        self._private.image = nil
+        self._private.default = nil
     end
 
-    image = image and surface.load(image)
-
-    if image then
-        local w = image.width
-        local h = image.height
-        if w <= 0 or h <= 0 then
-            return false
-        end
-    end
-
-    if self._private.image == image then
-        -- The image could have been modified, so better redraw
-        self:emit_signal("widget::redraw_needed")
-        return true
-    end
-
-    self._private.image = image
+    if not setup_succeed then return false end
 
     self:emit_signal("widget::redraw_needed")
     self:emit_signal("widget::layout_changed")
@@ -143,7 +191,6 @@ end
 -- @property resize
 -- @param allowed If false, the image will be clipped, else it will be resized
 --   to fit into the available space.
-
 function imagebox:set_resize(allowed)
     self._private.resize_forbidden = not allowed
     self:emit_signal("widget::redraw_needed")


Elv13:
I am very grateful you start to upstream this. It has been requested so many times.

Please note that the CI fails because:

https://travis-ci.com/awesomeWM/awesome/jobs/204956061#L3001

Elv13:
Oh, and this will also require to update the README.md to mention that librsvg is now "really" an optional dependency. SVG "loaded" before, but it was rendered incorrectly and thus wasn't advertised as a feature.

psychon:
Quick drive-by comment: I'm not sure that reloading the SVG "all the time" (yes, there is a cache; still), is much of a good idea. I'd rather have the code use librsvg directly and load the SVG just once. Afterwards it can be easily rendered at different scales.
I vaguely remember having hacked up something like this once upon a time, but I cannot find the code just now.

worron:
Please share mentioned code if you'll be able to find it somehow.

psychon:
Quick version of what I had in mind (no idea where to look for that code that I imagine I wrote once):
```lua
local lgi = require("lgi")
local Rsvg = lgi.Rsvg
local cairo = lgi.cairo

local function load_svg(file)
	local handle, err = Rsvg.Handle.new_from_file(file)
	if err then
		return nil, tostring(err)
	end

	dim = handle:get_dimensions()
	print("size is", dim.width, dim.height, dim.em, dim.ex)

	return handle
end

local function draw_handle(handle)
	local s = cairo.ImageSurface('RGB32', 100, 100)
	local cr = cairo.Context(s)

	handle:set_dpi(42) -- Could wire this up to AwesomeWM's DPI code

	-- can use cr:scale() here to resize the image
	handle:render_cairo(cr)
end

print(load_svg("does not exist"))
draw_handle(load_svg("/usr/share/inkscape/icons/feOffset-icon.svg"))
```
This loads the file once, producing an RsvgHandle for it. This can then be drawn to a cairo context at any size.
API docs can be found at https://developer.gnome.org/rsvg/stable/rsvg-RsvgHandle.html

worron:
Interesting. I'll try to integrate this approach.

worron:
@psychon Any thoughts about this implementation?

lib/wibox/widget/imagebox.lua
```diff
@@ -6,29 +6,43 @@
...
+
 local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
 
 local imagebox = { mt = {} }
 
+---Load rsvg handle form image file
+---@tparam string file Path to svg file.
+---@return Rsvg handle
+local function load_svg_and_handle_errors(file)
+    local handle, err = Rsvg.Handle.new_from_file(file)
```
Elv13: `Rsvg` can be `nil`. Either because it isn't compiled with GObject Introspection or because it isn't installed. This is the case by default in many server and KDE based distributions. I will take care of notifying the distributions if this gets merged about added librsvg as a suggested dependency.

lib/wibox/widget/imagebox.lua
```diff
@@ -79,36 +75,38 @@ end
...
-            return false
+        -- try to load rsvg handle from file if it's vector
+        local is_vector = string.match(image, "%.svg$")
+        self._private.handle = is_vector and load_svg_and_handle_errors(image) or nil
+
+        if not self._private.handle then
+            -- try to load cairo surface from raster image file
+            local _surface = surface.load(image)
+            if not _surface or _surface.width <= 0 or _surface.height <= 0 then
+                return false
```
Elv13: If this returns here, `self._private.image` might still have the vector size table.

lib/wibox/widget/imagebox.lua
```diff
@@ -79,36 +75,38 @@ end
...
+            end
+            self._private.image = _surface
+        else
+            -- set dummy surface object for default sizes reference
+            local dim = self._private.handle:get_dimensions()
+            self._private.image = { width = dim.width, height = dim.height }
```
Elv13: Given they are different types, what about using 2 different attribute variables? This would avoid some type confusion issues down the line? This will require to cleanup both, but that's less effort than adding `if` elsewhere.

lib/wibox/widget/imagebox.lua
```diff
@@ -79,36 +75,38 @@ end
...
 -- @return true on success, false if the image cannot be used
-
 function imagebox:set_image(image)
     if type(image) == "string" then
-        image = surface.load(image)
-        if not image then
-            print(debug.traceback())
-            return false
+        -- try to load rsvg handle from file if it's vector
+        local is_vector = string.match(image, "%.svg$")
```
Elv13: Please use `:match()`, this is used more often in the codebase (but not by much...).

lib/wibox/widget/imagebox.lua
```diff
@@ -6,29 +6,43 @@
...
+
 local base = require("wibox.widget.base")
 local surface = require("gears.surface")
 local gtable = require("gears.table")
+local gdebug = require("gears.debug")
 local setmetatable = setmetatable
 local type = type
-local print = print
+local math = math
+local string = string
```
Elv13: This can be removed if you use `:match()`

worron:
Some fixes done, let me know if something missed or another improvement needed.

lib/wibox/widget/imagebox.lua
psychon: Rsvg cannot be `nil`, because lgi will throw an error instead of returning `nil`:
```lua
$ lua -e 'lgi = require("lgi") ; a = lgi.does_not_exist ; print("I survived")'
lua: /usr/share/lua/5.2/lgi/namespace.lua:158: Typelib file for namespace 'does_not_exist' (any version) not found
stack traceback:
	[C]: in function 'assert'
	/usr/share/lua/5.2/lgi/namespace.lua:158: in function </usr/share/lua/5.2/lgi/namespace.lua:155>
	(...tail calls...)
	[C]: in ?
```
(Not much time, I didn't look at anything here, I just saw this comment)

psychon:
Well, okay, one quick thought: (@Elv13)
Is it worth introducing an abstraction over something that is either a cairo surface or an Rsvg handle? Something like `gears.image` with an API something like:
```
local img, err = gears.image("path/to/file") -- Can load from file
local width, height = img:get_size() -- Dunno if get_size() is the right name, I think Rsvg uses "dimensions" and cairo does not directly expose this information, so cannout use the same name
img:draw(cr) -- Draws the image to the cairo context
```
Pseudo-code of a possible implementation:
```lua
local mt_cairo, mt_rsvg = {}
local function new(file)
    if rsvg_was_loaded then
        local handle = load_svg(file)
        if handle then
            return setmetatable({handle}, mt_rsvg)
        end
    end
    local surface, err = gears.surface.load(file)
    if err then
        return err
    end
    return setmetatable({surface}, mt_cairo)
end
function mt_rsvg:draw(cr)
    return self[1]:render_to_cairo_or_whatever_the_function_was_called(cr)
end
function mt_cairo:draw(cr)
    cr:save()
    cr:set_source_surface(self[1], 0, 0)
    cr:paint()
    cr:restore()
end
[etc]
```
If only the imagebox were to ever use this, then I guess it makes no sense. If there could be lots of users, then... well, might make sense to have this.
Someone with too much time could even make this support animations. At least GdkPixbuf supports some animations.

worron:
@psychon Cairo surface and an Rsvg handle both have some  set of unique methods which can be used in future. Trying  to reach original methods within such abstraction may result a messy code, is not it?

psychon:
True, they both have lots of unique methods. But from ten feet away (well, 20 meters away), they both describe an image which can be displayed on screen. And an image has a size and that's about it.

But I guess you are right and I am just over-engineering things.
Will try to find some time to look at your commits in here.

Elv13:
> Is it worth introducing an abstraction over something that is either a cairo surface or an Rsvg handle?

As a internal/private API, "meh, maybe", but I prefer not having a new public API for image. I already have enough pain maintaining backward compatibility while trying to swap out the broken/deadend APIs. Most images can already be string, surfaces, functions (some of them) and now `gears.image`. Sure, we can add an universal "just convert them" function like `gears.color`. However it means adding this to all modules too. This is a lot of pain for a whole lot of nothing. 99% of the users use strings everywhere and 1% use surfaces. And its enough. The only good use case is having something like @Veratil `beautiful.colorful` module for pixmaps. My "Blind" module has such a class. But doesn't belong to `gears` and is a special case to auto-generate textures for more advanced themes. Not something for the average user.

Veratil:
> The only good use case is having something like @Veratil `beautiful.colorful` module for pixmaps. My "Blind" module has such a class.

I'm all for adding that. I just have no experience using pixmaps so I wouldn't know where to start. :)

Elv13:
> I'm all for adding that. I just have no experience using pixmaps so I wouldn't know where to start. :)

A bit off topic, but:

* https://github.com/Elv13/awesome-configs/blob/master/blind/common/pixmap.lua
* https://github.com/Elv13/awesome-configs/blob/master/blind/common/pattern2.lua

Of course, the API is unsalvageable. I am not even done working on the colorful branch to propose some changes... I had some time late last week and I spent it on the "rules for everything" branch and ended up spending some in the doc and history branches because I was pinged about them. The "rules for everything" is "officially" my top priority since it has to be in v4.4 to make the new notifications usable. The current preset API needs changes which cannot be done once released without breaking *released* API compatibility.

lib/wibox/widget/imagebox.lua
```diff
@@ -6,29 +6,49 @@
...
 
 local imagebox = { mt = {} }
 
+---Load rsvg handle form image file
+---@tparam string file Path to svg file.
+---@return Rsvg handle
+---@treturn number Image default width
+---@treturn number Image default height
+local function try_to_load_svg(file)
+    if not (Rsvg and file:match("%.svg$")) then return end
```
psychon: I'm not sure that matching on the file extension is a good idea and the check for `Rsvg` is dead code: This will always be true-ish (see https://github.com/awesomeWM/awesome/pull/2779/commits/baabe51eef528440da0860a4dd2feae9fb296a8f#r291985379 )

lib/wibox/widget/imagebox.lua
```diff
@@ -6,29 +6,49 @@
...
+
+    local handle, err = Rsvg.Handle.new_from_file(file)
+    if not handle then
+        gdebug.print_error(debug.traceback(
+            "Rsvg failed to load '" .. tostring(file) .. "': " .. tostring(err)))
+        return
```
psychon: Should this error be turned into "the callers problem" instead of being printed immediately? I.e. `return nil, err` and then handle this in the caller?

lib/wibox/widget/imagebox.lua
```diff
@@ -79,36 +81,48 @@ end
 --- Set an imagebox' image
 -- @property image
 -- @param image Either a string or a cairo image surface. A string is
---   interpreted as the path to a png image file.
+--   interpreted as the path to a image file.
```
psychon: `an image file`, I think

lib/wibox/widget/imagebox.lua
```diff
@@ -79,36 +81,48 @@ end
...
 -- @return true on success, false if the image cannot be used
-
 function imagebox:set_image(image)
```
psychon: I get a bit lost in this function. I feel like it could help to factor out some helper functions from this, do you agree? (But only those that would be called at least twice, but my feeling is that both of these would be called at least twice)
```
local function set_surface(self, surf)
   self._private.default_size = { width = surf:get_width(), height = surf:get_height() } -- Does it make sense to rename .default like this?
   self._private. handle = nil
   self._private.image = surf
end
local function set_handle(self, handle)
    local dim = handle:get_dimensions()
    self._private.default_size = { width = dim.width, height = dim.height }
    self._private.handle = handle
    self._private.image = nil
end
```
(Sorry that it took me so long to look at this)

lib/wibox/widget/imagebox.lua
```diff
@@ -79,36 +81,48 @@ end
...
     if type(image) == "string" then
-        image = surface.load(image)
-        if not image then
-            print(debug.traceback())
-            return false
+        -- try to load rsvg handle from file if it's vector
+        local handle, width, height = try_to_load_svg(image)
```
psychon: With my comment above, `try_to_load_svg` would only have to return the handle and not also get the dimensions. I feel like this then really does not need to be a separate function and can just be inlined here. Do you agree?

lib/wibox/widget/imagebox.lua
worron: Excuse me, I'm not sure how to properly avoid such error throwing. May I use `pcall`?

lib/wibox/widget/imagebox.lua
worron: Can't see why`set_handle` should be called more than once. I personally don't mind single use helpers btw.

lib/wibox/widget/imagebox.lua
worron: >try_to_load_svg would only have to return the handle and not also get the dimensions

Theoretically rsvg can load image with zero width or height. According current surface handling it should count as 'load failed'.  So we have to mess around `handle:get_dimensions()` in the `set_image` body. Or modify helper to return validation result, something like
```lua
local function set_handle(ib, handle)
    local dim = handle:get_dimensions()
    local is_handle_valid = dim.width > 0 and dim.height > 0
    if not is_handle_valid then return end
    
    ib._private.default = { width = dim.width, height = dim.height }
    ib._private.handle = handle
    ib._private.image = nil
    return true
end

-- somewhere inside 'imagebox:set_image'
local rsvg_succeed
local handle, err = Rsvg.Handle.new_from_file(image)
if not err then
	rsvg_succeed = set_handle(self, handle)
end

if not rsvg_succeed then
	-- rsvg handle failed, try to load cairo surface with pixbuf
end
```
Is this approach ok?

lib/wibox/widget/imagebox.lua
psychon: Yup, I guess that's the best approach. Quick attempt (with a typo in `Rsvgg` to show that it is working):
```lua
local lgi = require("lgi")
local Rsvg = nil
do
	-- Load Rsvg if it is available, else set it to nil.
	local success, err = pcall(function()
		Rsvg = lgi.Rsvgg
	end)
	if not success then
		-- I guess this should go to gears.debug.print_warning...?
		print("Could not load Rsvg: " .. tostring(err))
	end
end
print(Rsvg)
```

lib/wibox/widget/imagebox.lua
psychon: Should `:set_image` work when called with an RsvgHandle, just like it can already be called with a cairo surface? That would mean there is another call...

lib/wibox/widget/imagebox.lua
psychon: > Theoretically rsvg can load image with zero width or height. 

Well, but in that case the image has width/height zero, so any other attempt at loading it will also produce zero width/height. So... I'm not quite sure what should be done when a zero-sized image is loaded, except, well, having a zero sized image. Dunno....

> Is this approach ok?

Fine with me. Sorry if I am being annoying. :-)

lib/wibox/widget/imagebox.lua
worron: Sorry, I was not consider RsvgHandle as `set_image` argument. Was thought it will be purely imagebox inner object.

Just to confirm: you are meaning `string|surface|handle|nil`all the bunch of types should be allowed for `imagebox:set_image`?

psychon: Kind of. I am asking if it makes sense to allow that, or not. So, basically yes, that's what I mean.

worron:
Some fixes done, another review iteration please.

lib/wibox/widget/imagebox.lua
```diff
@@ -20,43 +19,45 @@ local math = math
...
         local w, h = self._private.default.width, self._private.default.height
         local aspect = math.min(width / w, height / h)
         cr:scale(aspect, aspect)
     end
 
-    -- Set the clip
-    if self._private.clip_shape then
-        cr:clip(self._private.clip_shape(cr, width, height, unpack(self._private.clip_args)))
-    end
-
```
psychon: Can you rebase on master? I think that should get rid of this change. Sorry for extra trouble I am causing here.

worron: Done.

PR #2785, opened 2019-06-08
Fix a thousand documentation issues.

Elv13:
I started working on that as a quick fix for an issue, then it kept pulling strings. Then after a while I had no choice but to finish because this PR is unrebasable.

 * Fix all missing property types
 * Fix all field names to have their "real" name (ei. `module.foo` instead of `awfu.layout.foo`)
 * Fix all methods interpreted as functions by ldoc
 * Fix all functions interpreted as methods by ldoc
 * Standardize all function names to have the full module path instead of a mix
 * Add dozens of accidentally undocumented functions
 * Fix a rendering issues where bullet points list got merged with parameter definition.
 * Replace some ldoc workaround by ldoc extensions
 * "Really" add support for typed properties to ldoc (using undocumented APIs)
 * Stop using the implicit method vs. function and explicitly declare everything to avoid errors.
 * Change how methods definition is rendered to avoid ambiguities
 * Add more sections to the left sidebar
 * Fix rendering issues all over the place
 * Convert some `@module` to classes (it should always have been this way)
 * Add a constructor stereotype section to the doc
 * Many, many more issues

As soon as this passes the CI (which I expect will take a few iterations), this has to be merged because it will have rebase conflict with everything.

@blueyed Everything we discussed in Berlin as something we want for the doc is implemented in this PR. The result is a very large improvement.


Elv13:
Before:

Screenshot_20190608_012336

After:

Screenshot_20190608_012253


Highlight:

image

Less entries in the sidbars:

Screenshot_20190608_012637

Type information for the description:

Screenshot_20190608_012734

Elv13:
Merging this now. If there is requested changes, they will be done in future pull requests. If it wasn't merged "this soon", it would have merge conflicts with nearly everything.

worron:
Seems index has broken links  in sections which contain space in name. For example `widget_layouts` instead of `widget%20layouts`.

alfunx:
Apparently `no_spaces` puts underscores at some places. At least, removing `no_spaces` call in [`docs/ldoc.ltp:308`](https://github.com/awesomeWM/awesome/blob/master/docs/ldoc.ltp#L308) fixes it for me.

Elv13:
I don't have much time today, but I will see what I can do "soon". One mitigation would be to use the custom "hook" function ldoc provides to manipulate the references. We don't use it so far. Another mitigation would be to monkey-patch the faulty function from the template. This is an ldoc bug. They should use the tag (in this case, `@widgetmod`) and not the human readable string... But ldoc is dead, so we are on our own.

PR #2786, opened 2019-06-09
Update 01-readme.md

DrEuler:
Fixed a spelling error

PR #2787, opened 2019-06-09
Fix minor stuff in README

alfunx:
Some more formatting / spelling fixes in the README. (#2786 follow-up)


alfunx:
@psychon Yes for `arg` is a valid abbreviation for `args`. Just thought `args` would better convey the fact that it can take multiple arguments, for people not familiar with gdb.

PR #2790, opened 2019-06-11
Allow the screens to be fully managed by Lua

### LINKED ISSUE #1382: Switching monitor causes clients to move to the first tag
```
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✘
 • LGI version: 0.9.0
```

- attach another monitor to your machine
- switch to it using `xrandr`

**Actual result:**

All of the tags are moved to the first client. All layouts are as configured, but rules didn't apply.

**Expected result:**

Clients stay where they're.

Elv13:
This pull request ~~publishes the draft commits to~~ allow AwesomeWM to be started without screens and let some Lua code add "fake" ones. This is hidden behind a command line option (`--screen off`) because it would otherwise break some configs. With this option enabled, `mouse.screen` can be nil for a short time and this isn't supported. With all the `request::` sections I am promoting for a more modular `rc.lua`, this is less of an issue because less code runs in the `_G` context and more within callbacks (which are unaffected).

The commits also add some new signals and method to fetch the rectangle areas used by physical screens without creating the screen objects. ~~So far, it doesn't fetch the outputs, which is a problem. There is some memory leaks and this is actually untested with multiple "real" screens. It also has an obsolete/unused data structure in the diff because I am lazy.~~

The reason I publish this now (the older parts are from 2017) is because of my ongoing effort to add an `awful.rules` like API to all core objects. Screens are core objects and the only way to integrate them with the rules is for Lua to manage them. I will also need to use multiple/changing screens again for my new job, so better work on this now.

~~So this PR is mostly for opinions and feedback about the design and initialization sequence. The code isn't so great (yet).~~

The most basic Lua implementation is 

```lua
screen.connect_signal("scanning", function()
    screen.automatic_factory = false
end)

client.connect_signal("property::viewports", function()
    for _, viewport in ipairs(screen.viewports()) do
       if not find_existing_screen_function(viewport) then
            local geo = viewport.geometry
            screen.fake_add(geo.x, geo.y, geo.width, geo.height)
        end
    end

    --TODO Check for dead screens.
end)

-- Should never happen, but just in case.
screen.connect_signal("scanned", function()
    if screen.count() == 0 then
       screen.fake_add(0, 0, 640, 480)
    end
end)
```

Fixes #1382

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -937,11 +937,12 @@ require("awful.screen.dpi")(screen, data)
 
 -- Set the wallpaper(s) and create the bar(s) for new screens
 
-capi.screen.connect_signal("added", function(s)
+capi.screen.connect_signal("_added", function(s)
     -- If it was emited from here when screens are created with fake_add,
     -- the Lua code would not have an opportunity to polutate the screen
     -- metadata. Thus, the DPI may be wrong when setting the wallpaper.
     if s._managed ~= "Lua" then
+        s:emit_signal("added")
         s:emit_signal("request::desktop_decoration")
         s:emit_signal("request::wallpaper")
     end
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -740,6 +740,26 @@ end
 -- @property preferred_dpi
 -- @param number the DPI value.
 
+--- The maximum diagonal size in millimeters.
+--
+-- @property mm_maximum_size
+-- @param number
+
+--- The minimum diagonal size in millimeters.
+--
+-- @property mm_minimum_size
+-- @param number
+
+--- The maximum diagonal size in inches.
+--
+-- @property inch_maximum_size
+-- @param number
+
+--- The minimum diagonal size in inches.
+--
+-- @property inch_minimum_size
+-- @param number
+
 --- Emitted when a new screen is added.
 --
 -- The handler(s) of this signal are responsible of adding elements such as
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -13,6 +13,7 @@ local capi =
     screen = screen,
     client = client,
     awesome = awesome,
+    root = root,
 }
 local gdebug = require("gears.debug")
 local gmath = require("gears.math")
@@ -260,6 +261,58 @@ function screen.object.set_padding(self, padding)
     end
 end
 
+--- A list of outputs for this screen with their size in mm.
+--
+-- Please note that the table content may vary. In some case, it might also be
+-- empty.
+--
+-- An easy way to check if a screen is the laptop screen is usually:
+--
+--    if s.outputs["LVDS-1"] then
+--        -- do something
+--    end
+--
+-- **Signal:**
+--
+--  * *property::outputs*
+--
+-- **Immutable:** true
+-- @property outputs
+-- @param table
+-- @tfield table table.name A table with the screen name as key (like `eDP1` on a laptop)
+-- @tfield integer table.mm_width The screen physical width.
+-- @tfield integer table.mm_height The screen physical height.
+-- @tfield integer table.name The output name.
+-- @tfield integer table.viewport_id The identifier of the viewport this output
+--  corresponds to.
+
+function screen.object.get_outputs(s)
+    local ret = {}
+
+    local outputs = s._custom_outputs
+        or (s.data.viewport and s.data.viewport.outputs or s._outputs)
+
+    -- The reason this exists is because output with name as keys is very
+    -- convenient for quick name lookup by the users, but inconvenient in
+    -- the lower layers since knowing the output count (using #) is better.
+    for k, v in ipairs(outputs) do
+        ret[v.name or k] = v
+    end
+
+    return ret
+end
+
+function screen.object.set_outputs(self, outputs)
+    self._custom_outputs = outputs
+    self:emit_signal("property::outputs", screen.object.get_outputs(self))
+end
+
+capi.screen.connect_signal("property::_outputs", function(s)
+    if not s._custom_outputs then
+        s:emit_signal("property::outputs", screen.object.get_outputs(s))
+    end
+end)
+
 --- Get the preferred screen in the context of a client.
 --
 -- This is exactly the same as `awful.screen.focused` except that it avoids
@@ -486,7 +539,7 @@ end
 function screen.object.get_tags(s, unordered)
     local tags = {}
 
-    for _, t in ipairs(root.tags()) do
+    for _, t in ipairs(capi.root.tags()) do
         if get_screen(t.screen) == s then
             table.insert(tags, t)
         end
@@ -560,75 +613,32 @@ function screen.set_auto_dpi_enabled(enabled)
     data.autodpi = enabled
 end
 
-
 --- The number of pixels per inch of the screen.
+--
+-- The default DPI comes from the X11 server. In most case, it will be 96. If
+-- `autodpi` is set to `true` on the screen, it will use the least dense dpi
+-- from the screen outputs. Most of the time, screens only have a single output,
+-- however it will have two (or more) when "clone mode" is used (eg, when a
+-- screen is duplicated on a projector).
+--
 -- @property dpi
 -- @param number the DPI value.
 
-local xft_dpi, fallback_dpi
+--- The lowest density DPI from all of the (physical) outputs.
+-- @property minimum_dpi
+-- @param number the DPI value.
 
-local function get_fallback()
-    local mm_per_inch = 25.4
+--- The highest density DPI from all of the (physical) outputs.
+-- @property maximum_dpi
+-- @param number the DPI value.
 
-    -- Following Keith Packard's whitepaper on Xft,
-    -- https://keithp.com/~keithp/talks/xtc2001/paper/xft.html#sec-editing
-    -- the proper fallback for Xft.dpi is the vertical DPI reported by
-    -- the X server. This will generally be 96 on Xorg, unless the user
-    -- has configured it differently
-    if root and not fallback_dpi then
-        local _, h = root.size()
-        local _, hmm = root.size_mm()
-        fallback_dpi = hmm ~= 0 and h * mm_per_inch / hmm
-    end
-
-    return fallback_dpi or 96
-end
-
-function screen.object.get_dpi(s)
-    local mm_per_inch = 25.4
-
-    if s.data.dpi or s.data.dpi_cache then
-        return s.data.dpi or s.data.dpi_cache
-    end
-
-    -- Xft.dpi is explicit user configuration, so honor it
-    if not xft_dpi and awesome and awesome.xrdb_get_value then
-        xft_dpi = tonumber(awesome.xrdb_get_value("", "Xft.dpi")) or false
-    end
-
-    if xft_dpi then
-        s.data.dpi_cache = xft_dpi
-        return s.data.dpi_cache
-    end
-
-    if not data.autodpi then
-        s.data.dpi_cache = get_fallback()
-        return s.data.dpi_cache
-    end
-
-    -- Try to compute DPI based on outputs (use the minimum)
-    local dpi = nil
-    local geo = s.geometry
-    for _, o in pairs(s.outputs) do
-        -- Ignore outputs with width/height 0
-        if o.mm_width ~= 0 and o.mm_height ~= 0 then
-            local dpix = geo.width * mm_per_inch / o.mm_width
-            local dpiy = geo.height * mm_per_inch / o.mm_height
-            dpi = math.min(dpix, dpiy, dpi or dpix)
-        end
-    end
-    if dpi then
-        s.data.dpi_cache = dpi
-        return dpi
-    end
-
-    s.data.dpi_cache = get_fallback()
-    return s.data.dpi_cache
-end
-
-function screen.object.set_dpi(s, dpi)
-    s.data.dpi = dpi
-end
+--- The preferred DPI from all of the (physical) outputs.
+--
+-- This is computed by normalizing all output to fill the area, then picking
+-- the lowest of the resulting virtual DPIs.
+--
+-- @property preferred_dpi
+-- @param number the DPI value.
 
 --- Emitted when a new screen is added.
 --
@@ -651,6 +661,177 @@ end
 --
 -- @signal request::wallpaper
 -- @tparam screen s The screen object.
+
+--- When a new (physical) screen area has been added.
+--
+-- Important: This only exists when Awesome is started with `--screen off`.
+-- Please also note that this doesn't mean it will appear when a screen is
+-- physically plugged. Depending on the configuration a tool like `arandr` or
+-- the `xrandr` command is needed.
+--
+-- The default handler will create a screen that fills the area.
+--
+-- To disconnect the default handler, use:
+--
+--    screen.disconnect_signal(
+--        "request::create", awful.screen.create_screen_handler
+--    )
+--
+-- @signal request::create
+-- @tparam table viewport
+-- @tparam table viewport.geometry A table with `x`, `y`, `width` and `height`
+--  keys.
+-- @tparam table viewport.outputs A table with the monitor name and possibly the
+--  `mm_width` and `mm_height` values if they are available.
+-- @tparam number viewport.id An identifier for this viewport (by pixel
+--  resolution). It
+--  will not change when outputs are modified, but will change when the
+--  resolution changes. Note that if it fully disappear, the next time an
+--  viewport with the same resolution appears, it will have a different `id`.
+-- @tparam number viewport.minimum_dpi The least dense DPI.
+-- @tparam number viewport.maximum_dpi The most dense DPI.
+-- @tparam number viewport.preferred_dpi The relative least dense DPI.
+-- @tparam table args
+-- @tparam string args.context Why was this signal sent.
+-- @see outputs
+-- @see awful.screen.create_screen_handler
+
+--- When a physical monitor viewport has been removed.
+--
+-- Important: This only exists when Awesome is started with `--screen off`.
+--
+-- If you replace the default handler, it is up to you to find the screen(s)
+-- associated with this viewport.
+--
+-- To disconnect the default handler, use:
+--
+--    screen.disconnect_signal(
+--        "request::remove", awful.screen.remove_screen_handler
+--    )
+--
+-- @signal request::remove
+-- @tparam table viewport
+-- @tparam table viewport.geometry A table with `x`, `y`, `width` and `height`
+--  keys.
+-- @tparam table viewport.outputs A table with the monitor name and possibly the
+--  `mm_width` and `mm_height` values if they are available.
+-- @tparam number viewport.id An identifier for this viewport (by pixel
+--  resolution). It will not change when outputs are modified, but will change
+--  when the resolution changes. Note that if it fully disappear, the next time
+--  an viewport with the same resolution appears, it will have a different `id`.
+-- @tparam number viewport.minimum_dpi The least dense DPI.
+-- @tparam number viewport.maximum_dpi The most dense DPI.
+-- @tparam number viewport.preferred_dpi The relative least dense DPI.
+-- @tparam table args
+-- @tparam string args.context Why was this signal sent.
+-- @see awful.screen.remove_screen_handler
+
+--- When a physical viewport resolution has changed or it has been replaced.
+--
+-- Important: This only exists when Awesome is started with `--screen off`.
+--
+-- Note that given the viewports are not the same, the `id` wont be the same.
+-- Also note that if multiple new viewports fit within a single "old" viewport,
+-- the resized screen will be the one with the largest total overlapping
+-- viewport (`intersection.width*intersection.height`), regardless of the
+-- outputs names.
+--
+-- To disconnect the default handler, use:
+--
+--    screen.disconnect_signal(
+--        "request::resize", awful.screen.resize_screen_handler
+--    )
+--
+-- @signal request::resize
+-- @tparam table old_viewport
+-- @tparam table old_viewport.geometry A table with `x`, `y`, `width` and
+--  `height` keys.
+-- @tparam table old_viewport.outputs A table with the monitor name and
+--  possibly the `mm_width` and `mm_height` values if they are available.
+-- @tparam number old_viewport.id An identifier for this viewport (by pixel
+--  resolution). It will not change when outputs are modified, but will change
+--  when the resolution changes. Note that if it fully disappear, the next
+--  time an viewport with the same resolution appears, it will have a different
+--  `id`.
+-- @tparam number old_viewport.minimum_dpi The least dense DPI.
+-- @tparam number old_viewport.maximum_dpi The most dense DPI.
+-- @tparam number old_viewport.preferred_dpi The relative least dense DPI.
+-- @tparam table new_viewport
+-- @tparam table new_viewport.geometry A table with `x`, `y`, `width` and
+--  `height` keys.
+-- @tparam table new_viewport.outputs A table with the monitor name and
+--  possibly the
+--  `mm_width` and `mm_height` values if they are available.
+-- @tparam number new_viewport.id An identifier for this viewport (by pixel
+--  resolution). It will not change when outputs are modified, but will change
+--  when the  resolution changes. Note that if it fully disappear, the next time
+--  an viewport with the same resolution appears, it will have a different `id`.
+-- @tparam number new_viewport.minimum_dpi The least dense DPI.
+-- @tparam number new_viewport.maximum_dpi The most dense DPI.
+-- @tparam number new_viewport.preferred_dpi The relative least dense DPI.
+-- @tparam table args
+-- @tparam string args.context Why was this signal sent.
+-- @see awful.screen.resize_screen_handler
+
+--- Default handler for `request::create`.
+--
+-- Important: This only exists when Awesome is started with `--screen off`.
+--
+-- A simplified implementation looks like:
+--
+--    function(viewport --[[, args]])
+--        local geo = viewport.geometry
+--        local s = screen.fake_add(geo.x, geo.y, geo.width, geo.height)
+--    end
+--
+-- If you implement this by hand, you must also implement handler for the
+-- `request::remove` and `request::resize`.
+--
+-- @signalhandler awful.screen.create_screen_handler
+-- @see request::create
+
+--- Default handler for `request::remove`.
+--
+-- Important: This only exists when Awesome is started with `--screen off`.
+--
+-- A simplified version of the logic is:
+--
+--    function (viewport --[[, args]])
+--        local geo = viewport.geometry
+--        for s in screen do
+--            if gears.geometry.rectangle.are_equal(geo, s.geometry) then
+--                s:fake_remove()
+--                return
+--            end
+--        end
+--    end
+--
+-- @signalhandler awful.screen.remove_screen_handler
+-- @see request::remove
+
+--- Default handler for `request::resize`.
+--
+-- Important: This only exists when Awesome is started with `--screen off`.
+--
+-- A simplified version of the logic is:
+--
+--    function (old_viewport, new_viewport --[[, args]])
+--        local old_geo, new_geo = old_viewport.geometry, new_viewport.geometry
+--        for s in screen do
+--            local sgeo = new_viewport.geometry
+--            if gears.geometry.rectangle.are_equal(old_geo, s.geometry) then
+--                s:fake_resize(
+--                    sgeo.x, sgeo.y, sgeo.width, sgeo.height
+--                )
+--            end
+--        end
+--    end
+--
+-- @signalhandler awful.screen.resize_screen_handler
+-- @see request::resize
+
+-- Add the DPI properties.
+require("awful.screen.dpi")(screen, data)
 
 -- Set the wallpaper(s) and create the bar(s) for new screens
 capi.screen.connect_signal("added", function(s)
@@ -676,30 +857,6 @@ end)
 capi.screen.connect_signal("request::wallpaper::connected", function(new_handler)
     for s in capi.screen do
         new_handler(s)
-    end
-end)
-
--- Create some screens when none exist. This can happen when AwesomeWM is
--- started with `--screen off` and no handler is used.
-capi.screen.connect_signal("scanned", function()
-    if capi.screen.count() == 0 then
-        -- Private API to scan for screens now.
-        if #capi.screen._viewports() == 0 then
-            capi.screen._scan_quiet()
-        end
-
-        local viewports = capi.screen._viewports()
-
-        if #viewports > 0 then
-            for _, area in ipairs(viewports) do
-                local geo = area.geometry
-                capi.screen.fake_add(geo.x, geo.y, geo.width, geo.height)
-            end
-        else
-            capi.screen.fake_add(0, 0, 640, 480)
-        end
-
-        assert(capi.screen.count() > 0, "Creating screens failed")
     end
 end)
 
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -882,6 +882,8 @@ end
 --    function(viewport --[[, args]])
 --        local geo = viewport.geometry
 --        local s = screen.fake_add(geo.x, geo.y, geo.width, geo.height)
+--        s:emit_signal("request::desktop_decoration")
+--        s:emit_signal("request::wallpaper")
 --    end
 --
 -- If you implement this by hand, you must also implement handler for the
@@ -934,9 +936,15 @@ end
 require("awful.screen.dpi")(screen, data)
 
 -- Set the wallpaper(s) and create the bar(s) for new screens
+
 capi.screen.connect_signal("added", function(s)
+    -- If it was emited from here when screens are created with fake_add,
+    -- the Lua code would not have an opportunity to polutate the screen
+    -- metadata. Thus, the DPI may be wrong when setting the wallpaper.
+    --if capi.screen.automatic_factory then
     s:emit_signal("request::desktop_decoration")
     s:emit_signal("request::wallpaper")
+    --end
 end)
 
 -- Resize the wallpaper(s)
@@ -948,15 +956,19 @@ end
 
 -- Create the bar for existing screens when an handler is added
 capi.screen.connect_signal("request::desktop_decoration::connected", function(new_handler)
-    for s in capi.screen do
-        new_handler(s)
+    if capi.screen.automatic_factory then
+        for s in capi.screen do
+            new_handler(s)
+        end
     end
 end)
 
 -- Set the wallpaper when an handler is added.
 capi.screen.connect_signal("request::wallpaper::connected", function(new_handler)
-    for s in capi.screen do
-        new_handler(s)
+    if capi.screen.automatic_factory then
+        for s in capi.screen do
+            new_handler(s)
+        end
     end
 end)
 
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -941,10 +941,10 @@ capi.screen.connect_signal("added", function(s)
     -- If it was emited from here when screens are created with fake_add,
     -- the Lua code would not have an opportunity to polutate the screen
     -- metadata. Thus, the DPI may be wrong when setting the wallpaper.
-    --if capi.screen.automatic_factory then
-    s:emit_signal("request::desktop_decoration")
-    s:emit_signal("request::wallpaper")
-    --end
+    if s._managed ~= "Lua" then
+        s:emit_signal("request::desktop_decoration")
+        s:emit_signal("request::wallpaper")
+    end
 end)
 
 -- Resize the wallpaper(s)
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -584,6 +584,106 @@ function screen.object.get_selected_tag(s)
     return screen.object.get_selected_tags(s)[1]
 end
 
+local function normalize(ratios, size)
+    local sum = 0
+
+    for _, r in ipairs(ratios) do
+        sum = sum + r
+    end
+
+    -- Avoid to mutate the input.
+    local ret = {}
+    local sum2 = 0
+
+    for k, r in ipairs(ratios) do
+        ret[k] = (r*100)/sum
+        ret[k] = math.floor(size*ret[k]*0.01)
+        sum2 = sum2 + ret[k]
+    end
+
+    -- Ratios are random float number. Pixels cannot be divided. This adds the
+    -- remaining pixels to the end. A better approach would be to redistribute
+    -- them based on the ratios. However, nobody will notice.
+    ret[#ret] = ret[#ret] + (size - sum2)
+
+    return ret
+end
+
+--- Split the screen into multiple screens.
+--
+-- This is useful to turn ultrawide monitors into something more useful without
+-- fancy client layouts:
+--
+-- @DOC_awful_screen_split1_EXAMPLE@
+--
+-- It can also be used to turn a vertical "side" screen into 2 smaller screens:
+--
+-- @DOC_awful_screen_split2_EXAMPLE@
+--
+-- @tparam[opt] table ratios The different ratios to split into. If none is
+--  provided, it is split in half.
+-- @tparam[opt] string mode Either "vertical" or "horizontal". If none is
+--  specified, it will split along the longest axis.
+-- @method split
+function screen.object.split(s, ratios, mode, _geo)
+    s = get_screen(s)
+
+    _geo = _geo or s.geometry
+    ratios = ratios or {50,50}
+
+    -- In practice, this is almost always what the user wants.
+    mode = mode or (
+        _geo.height > _geo.width and "vertical" or "horizontal"
+    )
+
+    assert(mode == "horizontal" or mode == "vertical")
+
+    assert((not s) or s.valid)
+    assert(#ratios >= 2)
+
+    local sizes, ret = normalize(
+        ratios, mode == "horizontal" and _geo.width or _geo.height
+    ), {}
+
+    assert(#sizes >=2)
+
+    if s then
+        if mode == "horizontal" then
+            s:fake_resize(_geo.x, _geo.y, sizes[1], _geo.height)
+        else
+            s:fake_resize(_geo.x, _geo.y, _geo.width, sizes[1])
+        end
+        table.insert(ret, s)
+    end
+
+    local pos = _geo[mode == "horizontal" and "x" or "y"]
+        + (s and sizes[1] or 0)
+
+    for k=2, #sizes do
+        local ns
+
+        if mode == "horizontal" then
+            ns = capi.screen.fake_add(pos, _geo.y, sizes[k], _geo.height)
+        else
+            ns = capi.screen.fake_add(_geo.x, pos, _geo.width, sizes[k])
+        end
+
+        table.insert(ret, ns)
+
+        if s then
+            ns.data.viewport = s.data.viewport
+
+            if not ns.data.viewport then
+                ns.outputs = s.outputs
+            end
+        end
+
+        pos = pos + sizes[k]
+    end
+
+    return ret
+end
+
 --- Enable the automatic calculation of the screen DPI (experimental).
 --
 -- This will cause many elements such as the font and some widgets to be scaled
--- a/lib/awful/screen/dpi.lua
+++ b/lib/awful/screen/dpi.lua
@@ -152,13 +152,14 @@ function module.create_screen_handler(viewport)
         geo.x,
         geo.y,
         geo.width,
-        geo.height
+        geo.height,
+        {_managed = true}
     )
 
     update_screen_viewport(s)
 
-    --s:emit_signal("request::desktop_decoration")
-    --s:emit_signal("request::wallpaper")
+    s:emit_signal("request::desktop_decoration")
+    s:emit_signal("request::wallpaper")
 end
 
 function module.remove_screen_handler(viewport)
--- a/lib/awful/screen/dpi.lua
+++ b/lib/awful/screen/dpi.lua
@@ -160,6 +160,9 @@ function module.create_screen_handler(viewport)
 
     s:emit_signal("request::desktop_decoration")
     s:emit_signal("request::wallpaper")
+
+    -- Will call all `connect_for_each_screen`.
+    s:emit_signal("added")
 end
 
 function module.remove_screen_handler(viewport)
--- a/lib/awful/screen/dpi.lua
+++ b/lib/awful/screen/dpi.lua
@@ -197,10 +197,44 @@ function module._get_xft_dpi()
     return xft_dpi
 end
 
+-- Some of them may be present twice or a giant viewport may exist which
+-- encompasses everything.
+local function deduplicate_viewports(vps)
+    local min_x, min_y, max_x, max_y = math.huge, math.huge, 0, 0
+
+    -- Sort the viewports by `x`.
+    for _, vp in ipairs(vps) do
+        local geo = vp.geometry
+        min_x = math.min(min_x, geo.x)
+        max_x = math.max(max_x, geo.x+geo.width)
+        min_y = math.min(min_y, geo.y)
+        max_y = math.max(max_y, geo.y+geo.height)
+    end
+
+    -- Remove the "encompass everything" viewport (if any).
+    if #vps > 1 then
+        for k, vp in ipairs(vps) do
+            local geo = vp.geometry
+            if geo.x == min_x and geo.y == min_y
+            and geo.x+geo.width == max_x and geo.y+geo.height == max_y then
+                table.remove(vps, k)
+                break
+            end
+        end
+    end
+
+    --TODO when the rules are added, mark the viewports that are embedded into
+    -- each other so the rules can decide to use the smaller or larger viewport
+    -- when creating the screen. This happens a lot then connecting to
+    -- projectors when doing presentations.
+
+    return vps
+end
+
 -- Provide a way for the tests to replace `capi.screen._viewports`.
 function module._get_viewports()
     assert(type(capi.screen._viewports()) == "table")
-    return capi.screen._viewports()
+    return deduplicate_viewports(capi.screen._viewports())
 end
 
 local function get_dpi(s)
--- a/lib/awful/screen/dpi.lua
+++ b/lib/awful/screen/dpi.lua
@@ -156,6 +156,9 @@ function module.create_screen_handler(viewport)
     )
 
     update_screen_viewport(s)
+
+    --s:emit_signal("request::desktop_decoration")
+    --s:emit_signal("request::wallpaper")
 end
 
 function module.remove_screen_handler(viewport)
--- a/lib/awful/screen/dpi.lua
+++ b/lib/awful/screen/dpi.lua
@@ -48,29 +48,65 @@ local function dpi_for_output(viewport, output)
 end
 
 local function dpis_for_outputs(viewport)
-    local max, min = 0, math.huge
+    local max_dpi, min_dpi, max_size, min_size = 0, math.huge, 0, math.huge
 
     for _, o in pairs(viewport.outputs) do
         local dpi = dpi_for_output(viewport, o)
         o.dpi = dpi
-        max = math.max(max, dpi)
-        min = math.min(min, dpi)
+
+        max_dpi = math.max(max_dpi, dpi)
+        min_dpi = math.min(min_dpi, dpi)
+
+        -- Compute the diagonal size.
+        if o.mm_width and o.mm_height then
+            o.mm_size   = math.sqrt(o.mm_width^2 + o.mm_height^2)
+            o.inch_size = o.mm_size/mm_per_inch
+            max_size    = math.max(max_size, o.mm_size)
+            min_size    = math.min(min_size, o.mm_size)
+        end
     end
 
     -- When there is no output.
-    if min == math.huge then
-        min = get_fallback_dpi()
-        max = min
+    if min_dpi == math.huge then
+        min_dpi = get_fallback_dpi()
+        max_dpi = min_dpi
     end
 
     --TODO Some output may have a lower resolution than the viewport, so their
     -- DPI is currently wrong. Once fixed, the preferred DPI can become
     -- different from the minimum one.
-    local pref = min
+    local pref_dpi = min_dpi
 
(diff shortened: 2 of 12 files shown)


objects/screen.c
```diff
@@ -94,12 +94,33 @@
  * @signal swapped
  */
 
+/** This signal is emitted when the list of physical screen area changes.
+ *
+ * Each areas in the list correspond to a **physical** screen rectangle. This
```
Veratil: TYPO: `Each areas` --> `Each area`
`correspond` --> `corresponds`

Additionally, L97, I'm not familiar with how the list is, but `list of physical screen areas` sounds correct to me.

objects/screen.c
```diff
@@ -94,12 +94,33 @@
...
+ */
+
  /**
   * The primary screen.
   *
   * @tfield screen primary
   */
 
+/**
+ * If screens objects are created automatically when new areas are detected.
```
Veratil: TYPO: `screens` --> `screen`

objects/screen.c
```diff
@@ -94,12 +94,33 @@
...
+ *
+ * Be very, very careful when setting this to false. You might end up with
+ * no screens. This is **not** supported. Always connect to the `scanned`
```
Veratil: `scanned::screen` to be specific?

objects/screen.c
```diff
@@ -94,12 +94,33 @@
...
+ * signal to make sure to create a fallback screen if none were created.
+ *
+ * @tfield[opt=true] boolean scrren.automatic_factory
```
Veratil: TYPO: `scrren` --> `screen`

objects/screen.h
```diff
@@ -30,7 +30,12 @@
 typedef struct screen_output_t screen_output_t;
 ARRAY_TYPE(screen_output_t, screen_output)
 
-struct a_screen
+/**
+ * A virtual screen as used by literally everything.
+ *
+ * Note that this structure doesn't always maps to physical screens.
```
Veratil: TYPO: `maps` --> `map`

objects/screen.h
```diff
@@ -46,6 +51,31 @@ struct a_screen
 };
 ARRAY_FUNCS(screen_t *, screen, DO_NOTHING)
 
+/**
+ * This structure tracks the area where there is physical screens.
```
Veratil: If an area can cross physical screen boundaries, then `is` --> `are`.
If an area cannot cross physical screen boundaries, then `is a physical screen`.

luaa.c
```diff
@@ -150,13 +154,56 @@ extern const struct luaL_Reg awesome_mouse_meta[];
...
+ */
+
+/** AwesomeWM is done scanning for screens.
+ *
+ * Connect to this signal to execute code after the screens have been created,
+ * but before the clients are added. This signal can also be used to split
+ * physical screens into multiple virtual screens before the clients (and their
+ * rules) are executed.
+ *
+ * @signal scanned::screen
```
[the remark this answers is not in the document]
Elv13: The difference is literally that they are different signals. There is no way to re-order signals connections. So, given modules needs to assume they are loaded in random order, having more signals allows for finer control over when the code is executed.

There's also the case that `scanned::screen` wont be emitted in `auto` mode because `rc.lua` is only initialized *after* they should have been sent. This makes it easier for module specifically designed to only work in `manual` mode to have less side effects when accidentally loaded in `auto` mode.

Also, because: consistency ;)

Elv13:
@blueyed Any idea how to correctly pass `--screen manual` for the 2 coverage tests? It also seem to sometime fail with `test-awful-client.lua`, but I can't reproduce locally, does it happen for you?

.travis.yml
```diff
@@ -50,14 +50,14 @@ addons:
...
-    - env: LUA=5.2 LUANAME=lua5.2 DO_COVERAGE=coveralls
+    - env: LUA=5.2 LUANAME=lua5.2 DO_COVERAGE=coveralls MANUAL_SCREENS=1
       addons:
         apt:
           packages:
             - *BASE_PACKAGES
             - liblua5.2-dev
             - lua5.2
-    - env: LUA=5.3 LUANAME=lua5.3 DO_COVERAGE=codecov
+    - env: LUA=5.3 LUANAME=lua5.3 DO_COVERAGE=codecov MANUAL_SCREENS=1
```
blueyed: I think it might make sense to turn "env" into a list, but not really necessary here already.

.travis.yml
```diff
@@ -229,7 +229,8 @@ install:
       return 0
     }
 script:
-  - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -DLUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -DSTRICT_TESTS=true -D DO_COVERAGE=$DO_COVERAGE -D CMAKE_C_FLAGS=-Werror"
+  - if [ "$MANUAL_SCREENS" != "1" ]; then export MANUAL_SCREENS=0; fi
+  - export CMAKE_ARGS="-DLUA_LIBRARY=${LUALIBRARY} -D LUA_INCLUDE_DIR=${LUAINCLUDE} -D OVERRIDE_VERSION=$AWESOME_VERSION -D STRICT_TESTS=true -D DO_COVERAGE=$DO_COVERAGE -D TEST_MANUAL_SCREENS=$MANUAL_SCREENS -D CMAKE_C_FLAGS=-Werror"
```
blueyed: This could also look like:
```
￼- if [  "$MANUAL_SCREENS" = "1" ]; then CMAKE_ARGS="$CMAKE_ARGS -D TEST_MANUAL_SCREENS=1"; fi
```

I.e. keep the original "export CMAKE_ARGS" (but OK to fix whitespace there maybe still), and then extend it for MANUAL_SCREENS=1.

CMakeLists.txt
```diff
@@ -404,6 +404,13 @@ target_link_libraries(test-gravity
 if(DO_COVERAGE)
     set(TESTS_RUN_ENV DO_COVERAGE=1)
 endif()
+
+# Start AwesomeWM tests with `--screen manual`
+if ("${TEST_MANUAL_SCREENS}" MATCHES "1" )
```
blueyed:
```suggestion
if ("${TEST_MANUAL_SCREENS}" MATCHES "1")
```

CMakeLists.txt
```diff
@@ -404,6 +404,13 @@ target_link_libraries(test-gravity
...
+    message("\n\nADDING TEST_MANUAL_SCREENS")
+    set(TEST_RUN_ARGS "--W --m")
```
blueyed: Don't scream.. ;)

blueyed: Two dashes?

awesome.c
```diff
@@ -559,13 +559,14 @@ exit_help(int exit_code)
...
-  -k, --check            check configuration file syntax\n\
-  -a, --no-argb          disable client transparency support\n\
-  -r, --replace          replace an existing window manager\n");
+  -h, --help                show help\n\
+  -v, --version             show version\n\
+  -c, --config FILE         configuration file to use\n\
+      --search DIR          add a directory to the library search path\n\
+  -k, --check               check configuration file syntax\n\
+  -a, --no-argb             disable client transparency support\n\
+  -m, --screen auto|manual  enable or disable automatic screen detection\n\
```
blueyed: Not clear what that means.. and what the default is.

awesome.c
```diff
@@ -634,6 +636,14 @@ main(int argc, char **argv)
             if (confpath != NULL)
                 fatal("--config may only be specified once");
             confpath = a_strdup(optarg);
+            break;
+          case 'm':
+            /* Validation */
+            if ((!optarg) || !(A_STREQ(optarg, "manual") || A_STREQ(optarg, "auto")))
+                fatal("The possible values of -m/--screen are \"auto\" and \"manual\"");
```
blueyed:
```suggestion
                fatal("The possible values of -m/--screen are \"auto\" or \"manual\"");
```

awesome.c
```diff
@@ -901,20 +911,38 @@ main(int argc, char **argv)
...
 
     xdgWipeHandle(&xdg);
 
+    screen_emit_scanned();
+
+    /* Exit if the user doesn't read the instructions properly */
+    if (globalconf.no_auto_screen && !globalconf.screens.len)
+        fatal("When -m/--screen is set to \"manual\", you **must** create a "
+              "screen object before before or inside the screen \"scanned\" "
+              " signal. Using AwesomeWM with no screen is **not supported**.");
```
blueyed: curly braces?

awesome.c
```diff
@@ -901,20 +911,38 @@ main(int argc, char **argv)
...
     p_delete(&confpath);
 
     xdgWipeHandle(&xdg);
 
+    screen_emit_scanned();
+
+    /* Exit if the user doesn't read the instructions properly */
+    if (globalconf.no_auto_screen && !globalconf.screens.len)
+        fatal("When -m/--screen is set to \"manual\", you **must** create a "
+              "screen object before before or inside the screen \"scanned\" "
```
blueyed:
```suggestion
              "screen object before or inside the screen \"scanned\" "
```

lib/naughty/init.lua
```diff
@@ -17,6 +19,65 @@ naughty.container = require("naughty.container")
 naughty.action = require("naughty.action")
 naughty.notification = require("naughty.notification")
 
+-- Attempt to handle early errors when using the manual screen mode.
+--
+-- Creating a notification popup before the screens are added wont work. To
```
blueyed:
```suggestion
-- Creating a notification popup before the screens are added won't work. To
```

lib/naughty/init.lua
```diff
@@ -17,6 +19,65 @@ naughty.container = require("naughty.container")
...
+-- work around this, the code below initialize some screens. One potential
```
blueyed:
```suggestion
-- work around this, the code below initializes some screens. One potential
```

lib/naughty/init.lua
```diff
@@ -17,6 +19,65 @@ naughty.container = require("naughty.container")
...
+-- problem is that it could emit enough signal to cause even more errors and
+-- lose the original error.
+--
+-- For example, the following error can be displayed using this fallback:
+--
+--    screen.connect_signal("scanned", function() foobar() end)
+--
+local function screen_fallback()
+    if screen.count() == 0 then
+        gdebug.print_warning("An error occurred before a scrren was added")
```
blueyed:
```suggestion
        gdebug.print_warning("An error occurred before a screen was added")
```

luaa.c
```diff
@@ -19,14 +19,18 @@
  *
  */
 
-/** awesome core API
+/** AwesomeWM lifecycle API.
  *
- * Additionally to the classes described here, one can also use X properties as
- * described in @{xproperties}.
+ * This module contains the functions and signal to manage the lifecycle of the
+ * AwesomeWM process. It allows to execute code at specific point from the early
```
blueyed:
```suggestion
 * AwesomeWM process. It allows to execute code at a specific point during early
```

luaa.c
```diff
@@ -19,14 +19,18 @@
...
+ * initialization all the way to the last events before exiting or restarting.
```
blueyed: Should be rephrased maybe.

luaa.c
```diff
@@ -19,14 +19,18 @@
...
+ *
+ * Additionally, it has additional signals for spawn and keyboard related
```
blueyed:
```suggestion
 * Additionally it handles signals for spawn and keyboard related events.
```

?

luaa.c
```diff
@@ -19,14 +19,18 @@
...
+ * events.
```
blueyed:
```suggestion
```

? (with the above)

manpages/awesome.1.txt
```diff
@@ -46,6 +46,8 @@ OPTIONS
     Add a directory to the library search path.
 *-a*, *--no-argb*::
     Don't use ARGB visuals.
+*-m*, *--screen*:: 'manual' or 'auto'::
+    Use "manual" to execute rc.lua before creating the screens.
```
blueyed: Should explain maybe also what you need to do then (the requirement of creating it manually).  Also with the error message for it then (if that's not the case already).

objects/client.c
```diff
@@ -132,6 +132,21 @@
...
+ *
+ * Connect to this signal when code needs to be executed after screens are
+ * initialized, but before clients are added.
+ *
+ * @signal scanning
+ */
+
+/** AwesomeWM is done scanning for clients.
+ *
+ * This emitted before the `startup` signal and after the `scanning::client`
```
blueyed:
```suggestion
 * This is emitted before the `startup` signal and after the `scanning::client` signal.
```

objects/screen.c
```diff
@@ -62,6 +62,37 @@
  */
 #define FAKE_SCREEN_XID ((uint32_t) 0xffffffff)
 
+/** AwesomeWM is about to scan for existing screens.
+ *
+ * Connect to this signal when code needs to be executed after the lua context
```
blueyed:
```suggestion
 * Connect to this signal when code needs to be executed after the Lua context
```

objects/screen.c
```diff
@@ -62,6 +62,37 @@
...
+ * is initialized and modules loaded, but before screens are added.
```
blueyed:
```suggestion
 * is initialized and modules are loaded, but before screens are added.
```

objects/screen.c
```diff
@@ -62,6 +62,37 @@
...
+ *
+ * To manage screens manually, set `screen.automatic_factory = false` and
+ * connect to the `property::areas` signal. It is then possible to use
+ * `screen.fake_add` to create virtual screens. Be careful when using this,
+ * when done incorrectly, no screens will be created. Using Awesome with zero
+ * screen is **not** supported.
```
blueyed:
```suggestion
 * screens is **not** supported.
```

objects/screen.c
```diff
@@ -62,6 +62,37 @@
...
+
+/** AwesomeWM is done scanning for screens.
+ *
+ * Connect to this signal to execute code after the screens have been created,
+ * but before the clients are added. This signal can also be used to split
+ * physical screens into multiple virtual screens before the clients (and their
+ * rules) are executed.
+ *
+ * Note that if no screens exist at this point, the fallback code will be
+ * triggered and the screens will be added.
```
blueyed:
```suggestion
 * triggered and the default (detected) screens will be added.
```

objects/screen.c
```diff
@@ -94,12 +125,33 @@
  * @signal swapped
  */
 
+/** This signal is emitted when the list of physical screen areas changes.
+ *
+ * Each area in the list corresponds to a **physical** screen rectangle. This
```
blueyed:
```suggestion
 * Each area in the list corresponds to a **physical** screen rectangle, which
```

objects/screen.c
```diff
@@ -94,12 +125,33 @@
...
+ * is **not** the areas of the `screen` objects.
```
blueyed:
```suggestion
 * is **not** the `areas` property of the `screen` objects.
```
?

awesome.c
Elv13: What do you propose?

blueyed: No idea now, but let's just keep it open / in mind.

manpages/awesome.1.txt
Elv13: Technically there's nothing to do. However, if you modified your config and it uses `mouse.screen` in the global context (aka, called directly from `rc.lua` rather than a signal callback), then there will be some errors.

tests/run.sh
```diff
@@ -174,6 +181,10 @@ fi
 # Start awesome.
 start_awesome() {
     cd "$build_dir"
+    echo STARTING DISPLAY="$D" SOURCE_DIRECTORY="$source_dir" \
+        AWESOME_THEMES_PATH="$AWESOME_THEMES_PATH" \
+        AWESOME_ICON_PATH="$AWESOME_ICON_PATH" \
+        timeout "$TEST_TIMEOUT" "$AWESOME" -c "$RC_FILE" "${awesome_options[@]}" > "$awesome_log"
```
blueyed: This gets overwritten/clobbered below - in case you want to keep it.

awesome.c
Elv13: I changed it to `on` and `off`

Elv13:
It might take a few more iteration to pass all tests, this is now "complete". All of what was previously done by the C code now has an equivalent Lua implementation. The implementation is fully based on request handlers, so they can be disconnected and replaced with another implementation if need be. It works with the default `rc.lua` and most configs except the ones calling `mouse.screen` from `_G`. This isn't a problem since the is currently disabled by default.

To recap, the goal of this pull request is to make screens a Lua concept rather than something tied to the X11 outputs. It is useful enough as-is to prevent bugs where all screens are removed during suspend (which already had a hack to never remove the last screen, but that's incorrect for multi-screen setup). It will get a whole lot more useful when the future `awful.screen.rules` API will be submitted. For that, the concept of volatile screens and "split by default" will become possible. This will make workflows which involve adding and removing screens, such as work laptops, much easier to "properly" implement.

@actionless Please review, I refactored a lot of the DPI code you care about.
@Oblomov I don't know if you still care about this, but this pull request opens a path toward proper HiDPI. It solves some race conditions where the DPI was set after the screen was created instead of the other way around. Along with the concept of "modeline" I plan to add soon which will allow to set the "api level" directly from a pre-processed comment in `rc.lua`. I also have a `wibox.container.scale` which in a local branch. In theory, the next step would be to add a scale factor to the drawing context so that layouts can properly align their virtual and physical pixels, then maybe that will be enough. The mentioned `awful.screen.rules` will have a `scale` (or `scale_factor`?) property.

actionless:
@Elv13 it seems what all the previous dpi-related code (like `function xresources.get_dpi(s)`) is still left, i guess you need to remove it and point to the new implementation

Elv13:
Changes:

 * rebased
 * delay the request::wallpaper signal in the chain to allow room for setting the DPI before it is called
 * Fix an issue where `s.outputs` now had indexed keys instead of named ones.
 * Rename monitors and areas to viewports in most places for consistency.

Elv13:
More changes:

 * Refactor the C "valid" boolean to be a set of lifecycle flags
 * Use the new lifecycle information only remove screens when they are auto-created in capi
 * Export the `_managed` property to Lua to avoid a race condition when the `added` signal is sent before the (Lua) constructor runs and thus `connect_for_each_screen` is called with incomplete objects (ei, without the DPI)

Elv13:
Changes:

 * Rebased
 * Fix a couple memory leaks including one present since AwesiomeWM v3.5 (randr output not always wiped when AwesomeWM quits)
 * Actually test "for real" with 5 screens (xrandr). It works fine with `--screen on` (default) and `--screen off` is usable, but has more corner case to handle before merging.

Elv13:
Changes:

 * rebased
 * add properties to get the diagonal size and maximum size. This is useful to fingerprint setups in the rules
 * add a `name` property to `capi.screen`. Getting the screen by output name is already supported, this is now extended to custom names created using some fingerprinting rules.

lib/awful/autofocus.lua
```diff
@@ -21,7 +21,7 @@ end
 --
 -- @param obj An object that should have a .screen property.
 local function check_focus(obj)
-    if not obj.screen.valid then return end
+    if (not obj.screen) or not obj.screen.valid then return end
```
psychon: Why do you have to add "screen might be `nil`"-checks to objects now? I thought (and read) that running without any screens is still not supported, so I would expect that no new "object suddenly has no screen"-cases can appear.

lib/awful/menu.lua
```diff
@@ -684,7 +684,7 @@ end
 -- -- a particular rule.
 -- -- Bound to a key, it can be used to select from dozens of terminals open on
 -- -- several tags.
--- -- When using @{rules.match_any} instead of @{rules.match},
+-- -- When using @{awful.rules.match_any} instead of @{awful.rules.match},
```
psychon: This seems totally unrelated to screens. Did this chunk somehow end up in the wrong pull request? Should it just get its own PR and be merged to master basically immediately?

lib/awful/screen.lua
```diff
@@ -234,6 +235,63 @@ function screen.object.set_padding(self, padding)
     end
 end
 
+--- A list of outputs for this screen with their size in mm.
+--
+-- This information is only available when RANDR is used to manage the screens.
+--
+-- Please note that the table content may vary.
```
psychon: What does this mean? Can this text be made more specific and less vague?

The current text could be read as "this table is not always equivalent to `{}`".

lib/awful/screen.lua
```diff
@@ -234,6 +235,63 @@ function screen.object.set_padding(self, padding)
...
+--
+--  * *property::outputs*
+--
+-- **Immutable:** true
+-- @property outputs
+-- @param table
+-- @tfield table table.name A table with the screen name as key (like `eDP1` on a laptop)
+-- @tfield integer table.mm_width The screen physical width
+-- @tfield integer table.mm_height The screen physical height
+-- @tfield integer table.name The screen physical height
```
psychon: ...what?!? Could you please rename this to something other than `name` then?

lib/awful/screen.lua
```diff
@@ -626,10 +753,187 @@ end
 -- @signal request::wallpaper
 -- @tparam screen s The screen object.
 
+--- When a new physical viewport has been added.
```
psychon: What is a viewport? This seems to be a new concept and I fail to find any documentation about it.

lib/awful/screen.lua
```diff
@@ -234,6 +235,63 @@ function screen.object.set_padding(self, padding)
...
+--
+-- **Immutable:** true
+-- @property outputs
+-- @param table
+-- @tfield table table.name A table with the screen name as key (like `eDP1` on a laptop)
+-- @tfield integer table.mm_width The screen physical width
+-- @tfield integer table.mm_height The screen physical height
+-- @tfield integer table.name The screen physical height
+-- @tfield integer table.viewport_id The identifier of the viewport this output
+--  corresponds to.
```
psychon: Why just an identifier instead of providing the viewport directly? (I guess whatever a viewport is, it has some kind of `.id` property, so nothing is lost here)

lib/awful/screen.lua
```diff
@@ -234,6 +235,63 @@ function screen.object.set_padding(self, padding)
...
+
+function screen.object.get_outputs(s)
+    local ret = {}
+
+    local outputs = s._custom_outputs
+        or (s.data.viewport and s.data.viewport.outputs or s._outputs)
```
psychon: Do I read this correctly: There are three different ways how the outputs of a screen can be saved? That seems a bit excessive to me, but I have no clue what I am looking at currently anyway...

lib/awful/screen.lua
```diff
@@ -234,6 +235,63 @@ function screen.object.set_padding(self, padding)
...
+    for k, v in ipairs(outputs) do
+        ret[v.name or k] = v
+    end
+
+    return ret
+end
+
+function screen.object.get_outputs(self)
+    return self._custom_outputs or self._outputs
+end
```
psychon: Wait... does this overwrite the function that was just defined? Both are called `screen.object.get_outputs`?!?

lib/awful/screen.lua
```diff
@@ -438,6 +496,7 @@ function screen.connect_for_each_screen(func)
     for s in capi.screen do
         func(s)
     end
+
```
psychon: This looks like a pointless change. Why adding a new empty line somewhere?

lib/awful/screen.lua
```diff
@@ -505,6 +564,100 @@ function screen.object.get_selected_tag(s)
...
+        sum = sum + r
+    end
+
+    -- Avoid to mutate the input.
+    local ret = {}
+    local sum2 = 0
+
+    for k, r in ipairs(ratios) do
+        ret[k] = (r*100)/sum
+        ret[k] = math.floor(size*ret[k]*0.01)
```
psychon: Why `*100/0.01`? That can be left out:
```
ret[k] = math.floor(size*r/sum)
```
Also, why does this round down? Doesn't that mean that everything is basically rounded to zero, because this only gets numbers between 0 and 1? Oh, wait... all of this is multiplied with `size`.

lib/awful/screen.lua
```diff
@@ -505,6 +564,100 @@ function screen.object.get_selected_tag(s)
...
+        sum2 = sum2 + ret[k]
+    end
+
+    -- Fix the sub-pixel issue.
```
psychon: No idea what "the sub-pixel issue" is, so this comment is (to me) useless. Please either clarify or just remove it.

lib/awful/screen.lua
```diff
@@ -505,6 +564,100 @@ function screen.object.get_selected_tag(s)
...
+    ret[#ret] = ret[#ret] + (size - sum2)
```
psychon: I would prefer something smarter than just giving "everything that was rounded away" to the last item, but since I do not know yet what this function is used for, I do not know whether this matters much.

lib/awful/screen.lua
```diff
@@ -505,6 +564,100 @@ function screen.object.get_selected_tag(s)
...
+-- It can also be used to turn a vertical "side" screen into 2 smaller screens:
+--
+-- @DOC_awful_screen_split2_EXAMPLE@
+--
+-- @tparam[opt] table ratios The different ratios to split into. If none is
+--  provided, it is split in half.
+-- @tparam[opt] string mode Either "vertical" or "horizontal". If none is
+--  specified, it will split along the longest axis.
+-- @method split
+function screen.object.split(s, ratios, mode, _geo)
```
psychon: `_geo` is undocumented

lib/awful/screen.lua
```diff
@@ -534,75 +687,49 @@ function screen.set_auto_dpi_enabled(enabled)
     data.autodpi = enabled
 end
 
-
 --- The number of pixels per inch of the screen.
+--
+-- Unless manually overriden, the dpi will be computed using the output side is
+-- `autodpi` is `true` or otherwise the X11 fallback (usually 96).
```
psychon: I fail to parse this sentence and I have no idea what "the output side" is supposed to refer to

lib/awful/screen/dpi.lua
```diff
@@ -0,0 +1,414 @@
...
+    -- the proper fallback for Xft.dpi is the vertical DPI reported by
+    -- the X server. This will generally be 96 on Xorg, unless the user
+    -- has configured it differently
+    if not fallback_dpi then
+        local _, h = root.size()
+        local _, hmm = root.size_mm()
+        fallback_dpi = hmm ~= 0 and h * mm_per_inch / hmm
+    end
+
+    return math.floor(fallback_dpi or 96)
```
psychon: Why are you rounding the DPI down? At least Pango uses `double` for DPI and besides that I do not know which piece of code should care.

lib/awful/screen/dpi.lua
```diff
@@ -0,0 +1,414 @@
...
+    local dpi = nil
+    local geo = viewport.geometry
+
+    -- Ignore outputs with width/height 0
+    if output.mm_width ~= 0 and output.mm_height ~= 0 then
+        local dpix = geo.width * mm_per_inch / output.mm_width
+        local dpiy = geo.height * mm_per_inch / output.mm_height
+        dpi = math.min(dpix, dpiy, dpi or dpix)
+    elseif ascreen._get_xft_dpi() then
+        dpi = ascreen._get_xft_dpi()
```
psychon: This can be simplified to `else dpi = ascreen._get_xft_dpi()`. The function is called one time less and the result is only used in an `dpi or foo`, so the result should be equivalent.

lib/awful/screen/dpi.lua
```diff
@@ -0,0 +1,414 @@
...
+    end
+
+    return math.floor(dpi or get_fallback_dpi())
```
psychon: Please remove the `floor`.

awesome.c
```diff
@@ -131,6 +131,8 @@ awesome_atexit(bool restart)
     /* Close Lua */
     lua_close(L);
 
+    screen_cleanup();
```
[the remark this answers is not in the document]
Elv13: Because that's actually the "last" real leak in during the tests if fontconfig and dbus are added to a suppression file. The leak checks come with "Address sanitizer" unless you disable it. I plan to enable ASAN in the CI eventually (it catches a lot more than just leaks). Then fixing the leaks makes sense. And, yes, I am aware there's a way to disable the memory leak component from ASAN, but then you also lose real leaks.

awesome.c
```diff
@@ -901,20 +913,44 @@ main(int argc, char **argv)
...
+    if (globalconf.no_auto_screen)
+    {
+        /* Disable automatic screen creation, awful.screen has a fallback */
+        globalconf.ignore_screens = true;
+
+        if(!luaA_parserc(&xdg, confpath))
+            fatal("couldn't find any rc file");
+    }
+
+    screen_emit_scanning();
```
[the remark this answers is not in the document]
Elv13: Because there is more PRs building on this. I can add more commits to this one, but it's already enormous ;). As you why not add the signal in those PRs: because users/modules can use it sonner.

The general idea is to have a better defined initialization for modules. With better defined steps, things like handling startup errors become a lot easier. In each steps, some "things", like having screen objects, is well defined. Given screens have 2 modes and different signals depending on the modes, it makes some sense to have both `scanning` and `scanned` for both client and screens.

lib/awful/autofocus.lua
Elv13: To allow the default rc.lua to work without screens. That's purely to make the CI pass. Later commits make up for the lack of screens by adding them at the right point in time to restore "sanity". However to have smaller commits that individually pass the tests, we have this commit...

lib/awful/menu.lua
Elv13: Ok.

I am not sure why it didn't caused errors before. Isn't the point of `@` is to make those link explicit so they fail when the target doesn't exist? In that case, it didn't exist and it didn't fail until now.

lib/awful/screen.lua
Elv13: /me copy paste facepalm

lib/awful/screen.lua
Elv13: ### First way

The user or a module sets the outputs manually for unknown reasons:

    my_screen.outputs = {foo = {mm_width =1, mm_height = 1}}

### Second way

Screens managed by CAPI have use the outputs provided by CAPI

### Third way

Screens managed by Lua provide the output stored the semi private `viewport` table. These outputs objects have previously been extended with a `dpi` and diagonal size. In the "new" screen mode, DPIs are per outputs to better handle some clone mode DPI problems previously handled, but unconfigurable.

lib/awful/screen.lua
Elv13: whoops, rebase issue

lib/awful/screen.lua
Elv13: On purpose.

lib/awful/screen.lua
Elv13: The main use case for this, as documented somewhere (I think it is in another unsubmitted PR), is that ultrawide monitors are larger than the field of view.

Lets assume that the rational from moving from the classic 4:3 ratio was to align the screen ratio to the common people field of view and that they done their homework.

Lets assume that any window that's larger than the user field of view is useless. Yes, some apps have their own tiling like blender or use canvas and tools like sidebars. Let's ignore those. Nobody wants a web browser spanning a 32:9 ratio with an hardcoded 1024 pixel large blog with aligned in the middle.

Now, screens with 21:9 and 32:9 ratios are being sold. I have one.

They are totally unusable with the current tiling system because clients get too large. That leaves either using custom layouts and/or configuration like @actionless `master_fill_policy` or split the screens into many virtual screens.

Virtual screens are superior because some common layouts like `awful.layout.suit.max` don't have `master_fill_policy`. Virtual screens also allow different taglists. This emulates the classic dual/tri-screen workflow, but without the gaps between the physical screens.

lib/awful/screen.lua
Elv13: Screens geometry are integers, ratios are floats.

lib/awful/screen/dpi.lua
Elv13: I am unsure if rounding is a good idea here. The reason I did it was because writing a rule like:

```
-- I declare I like small bars.
ruled.screen.append_rule {
    rule = {dpi = 321.325325435435 },
    properties = {dpi = 128 }
}
```

Fells a bit wrong and error prone. `==` is by default for matching (even if, as you saw, I added a magic way to work around this). So most users will find having to enter long floating point number inconvenient rather than having them rounded.

Yes sure, I added rule ranges for exactly this use case, but still...

What do you think?

actionless:
and this thing still actual btw:

> @Elv13 it seems what all the previous dpi-related code (like function xresources.get_dpi(s)) is still left, i guess you need to remove it and point to the new implementation

Elv13:
@actionless Why? It was there before, it is part of the public API and this doesn't seems to be in the pull request scope to do so? This pull request moves the concept of DPI from the screen and into its individual outputs. Then it adds an aggregate (maximum/minimum/preferred) DPIs for the screens. This is intended to be later used by the new screens rules. Once *thats* done, there is a pull request that adds the concept of scaling factor to the drawables (instead of DPIs, but based on them). Then after *that* is merged, *then* I think we have to care about cleanup up the past attempts at HiDPI. It wont be enabled by default until AwesomeWM 5 anyway since even with all the care in the world to scale everything automagically, it will still break some assumption that everything are "equals" pixels. That wont be the case since unless I come up with some great hacks, the client geometry will still be in pixels since they can move across the screens which may have different scaling factors, some "points" as an unit becomes meaningless.

lib/awful/menu.lua
psychon: Well, the LDoc doc say:

> You can also put references in backticks, like `\`stdvars\``. This is commonly used in
> Markdown to indicate code, so it comes naturally when writing documents. The difference
> is that the backticked expression does not have to be a reference and then will appear
> in code style; with @ references you will get a warning for unrecognized symbols
> and the result will be rendered as '???'.

Out current docs say:

> -- When using ??? instead of ???,

https://awesomewm.org/apidoc/popups_and_bars/awful.menu.html

And when building things, I get the following warning from LDoc:

> usage:6: module not found: rules rules.match_any
>usage:6: module not found: rules rules.match

So everything seems to work as advertised for me. :-)

(But indeed: For the stable version the links work fine, so something changed somewhere; I'm to lazy to figure that one out right now)

lib/awful/screen/dpi.lua
psychon: I think that someone somewhere somewhen will want to have a fractional DPI (I could link to the relevant XKCD if you want) and "but then rules do not work" will not make them happy. And it feels like a bad idea.

Your "I like small bars"-example should use "if DPI is larger than 200, use 128 instead".

If you really want to round DPIs for rule matching, I'd say that then DPIs should be rounded only for rule matching and not globally. I bet that will result in uglier code, but at least conceptually it feels a lot less weird.

Elv13: image

psychon:
Today I remembered that the RANDR extensions also has CRTCs, which are yet another factor in all of this....

@Elv13 Could you perhaps describe some example?

Let's say I have two outputs (physical displays) in clone mode (they show the same part of the screen; perhaps one is 10 pixels wider than the other). Additionally I have another high-DPI output that is right of the other two (according to RandR). This high-DPI output should be split into two $SOMETHING. What maps to a viewport, what to a screen, what to $SOMETHING_ELSE?

lib/awful/screen/dpi.lua
actionless: or mb just allow `gt`, `ge`, `lt`, `le` conditions in the rules :D

lib/awful/menu.lua
Elv13: done

lib/awful/screen.lua
Elv13: done

lib/awful/screen.lua
Elv13: modified to avoid using the "viewport" word in the public doc.

lib/awful/screen.lua
Elv13: To avoid using "miss_handlers" or "valid" because they are convoluted anti patterns and I prefer not to add more of them. It also makes improving the API harder because the code to handle the legacy is horrifying.

I added a comment in the code to reflect that design choice.

Elv13:
> Could you perhaps describe some example?

The viewport is an internal concept. It *exactly* maps to what the `C` screens are currently initialized to. It has the geometry and the list outputs. This is abstracted out of the screen class to allow screen rules and support for ultra-wide, which (almost) require to use screen splitting to make tiling usable.

In the case you defined, there will be 2 viewport and 3 screens. The fist viewport is the size of the larger (by 10px) screen. The second is the size of the second monitor. The first viewport will have a screen and the second will be shared by 2 screen objects.

More examples are in https://elv13.github.io/declarative_rules/ruled.screen.html

Elv13:
Changes:

 * Rebased to solve the conflicts with the notification PR.
 * Renamed `viewport_area_t` to `viewport_t` to avoid confusion.
 * Fixed most issues reported by @psychon 

Looking forward for another row of review.

lib/awful/screen/dpi.lua
Oblomov: DPI rounding isn't even that much of an issue, since the DPI value itself is rarely used as-is, and any rounding can (and arguably should) be applied at point-of-use (e.g. scaled lengths and font sizes). Most users would probably care more about the preferred scaling factor, expressable as a ratio of DPIs, and _that_ one could (should?) be rounded.

If we do provide a per-screen scaling_factor / pixel_ratio property (see e.g. Firefox' about:config and the Qt API), then _that_ should be rounded (per the user preference), otherwise it might be better to leave the DPI unrounded, and instead provide functions to round the scaled lenghs.

(The reason why I think that the scaling factor rounding should be user-controlled is that with non-perfectly integer scaling it's entirely up to the user to choose if they want rounding up, down or no rounding at all. Maybe define some kind of rounding policy property? [The Qt issue on the matter](https://bugreports.qt.io/browse/QTBUG-53022) has some potentially interesting ideas here.)

Oblomov:
> Today I remembered that the RANDR extensions also has CRTCs, which are yet another factor in all of this....

I don't think we should actually care about CRTCs. Rather, what we should look into is the difference between a RANDR Screen and a RANDR Monitor. Monitors were introduced in RANDR 1.5 (IIRC), to work around the issue of some very wide or high-resolution monitors actually having _two separate connectors_ internally, and using MST (Multi-Stream Transport): so at the hardware level they are two “monitors” (and map to two separate RANDR Screens, one for the left side and one for the right side), but one RANDR Monitor. Screens belonging to the same monitor share all the physical characteristics, and for most WM purposes should be viewed as a single screen, except for @Elv13 comment about tiling not being nice when it comes to ultrawide displays …

Oblomov:
> The viewport is an internal concept. It _exactly_ maps to what the `C` screens are currently initialized to. It has the geometry and the list outputs. This is abstracted out of the screen class to allow screen rules and support for ultra-wide, which (almost) require to use screen splitting to make tiling usable.

So, the `viewport` corresponds to the RANDR Monitor (1.5+) or Screen (RANDR 1.4 or under), while the `screen` is a (possibly arbitrary) section of a viewport?

Elv13:
I just found that I messed up a rebase in the last push and it's missing some changes, but from those missing changes:

```c
 /** Keep track of the screen viewport(s) independently from the screen objects.
  *
  * A viewport is a collection of `outputs` objects and their associated
  * metadata. This structure is copied into Lua and then further extended from
  * there. The `id` field allows to differentiate between viewports that share
  * the same position and dimensions without having to rely on userdata pointer
  * comparison.
  *
  * Screen objects are widely used by the public API and imply a very "visible"
  * concept. A viewport is a subset of what the concerns the "screen" class
  * previously handled. It is meant to be used by some low level Lua logic to
  * create screens from Lua rather than from C. This is required to increase the
  * flexibility of multi-screen setup or when screens are connected and
  * disconnected often.
  *
  * Design rationals:
  *
  * * The structure is not directly shared with Lua to avoid having to use the
  *   slow "miss_handler" and unsafe "valid" systems used by other CAPI objects.
  * * The `viewport_t` implements a linked-list because its main purpose is to
  *   offers a deduplication algorithm. Random access is never required.
  * * Everything that can be done in Lua is done in Lua.
  * * Since the legacy and "new" way to initialize screens share a lot of steps,
  *   the C code is bent to share as much code as possible. This will reduce the
  *   "dead code" and improve code coverage by the tests.
  *
  */

```

Elv13:
Rebased on top of the merged "fix the awful.rules doc reference" PR.

awesome.c
```diff
@@ -559,13 +561,14 @@ exit_help(int exit_code)
...
-  -k, --check            check configuration file syntax\n\
-  -a, --no-argb          disable client transparency support\n\
-  -r, --replace          replace an existing window manager\n");
+  -h, --help           show help\n\
+  -v, --version        show version\n\
+  -c, --config FILE    configuration file to use\n\
+      --search DIR     add a directory to the library search path\n\
+  -k, --check          check configuration file syntax\n\
+  -a, --no-argb        disable client transparency support\n\
+  -m, --screen on|off  enable or disable automatic screen creation (default: on)\n\
```
psychon: `--screen off` really is not self-explanatory. How about `--auto-screen`? Or inverting this and calling it `--manual-screen` so that `-m` makes more sense?

awesome.c
```diff
@@ -901,20 +913,44 @@ main(int argc, char **argv)
...
     p_delete(&confpath);
 
     xdgWipeHandle(&xdg);
 
+    screen_emit_scanned();
+
+    /* Exit if the user doesn't read the instructions properly */
+    if (globalconf.no_auto_screen && !globalconf.screens.len)
+        fatal("When -m/--screen is set to \"off\", you **must** create a "
+              "screen object before or inside the screen \"scanned\" "
```
psychon: Well, it does not really have to be inside of `scanned`. I could e.g. also create one when the config is being loaded. However, I guess it makes sense to be helpful-clear instead of pendantic-clear here, right?

awesome.c
```diff
@@ -901,20 +913,44 @@ main(int argc, char **argv)
...
+              " signal. Using AwesomeWM with no screen is **not supported**.");
+
+    client_emit_scanning();
+
     /* scan existing windows */
     scan(tree_c);
 
+    client_emit_scanned();
```
psychon: Should these signal emissions be moved to `scan` (and same for the screen scanning signals being moved to `screen_scan()`)? It feels to me like they belong there.

globalconf.h
```diff
@@ -111,6 +112,10 @@ typedef struct
     bool have_randr_15;
     /** Do we have a RandR screen update pending? */
     bool screen_refresh_pending;
+    /** Should screen be created before rc.lua? */
```
psychon: screenS (same thing in the next line for the comment for `ignore_screens`)

Also, append `rc.lua is loaded` (or else it sounds like the file itself is created)

luaa.c
```diff
@@ -19,14 +19,17 @@
  *
  */
 
-/** awesome core API
+/** AwesomeWM lifecycle API.
```
psychon: The changes to `luaa.c` also look quite simple and non-controversial. Could be moved to a separate PR, if you want. (After incorporating Daniel's feedback.)

luaa.h
```diff
@@ -325,6 +325,7 @@ int luaA_default_index(lua_State *);
 int luaA_default_newindex(lua_State *);
 void luaA_emit_startup(void);
 
+
```
psychon: This change can just be dropped.

objects/screen.h
```diff
@@ -30,17 +30,27 @@
...
-    /** Is this screen still valid and may be used? */
-    bool valid;
+    screen_flags_t flags;
     /** Screen geometry */
     area_t geometry;
     /** Screen workarea */
     area_t workarea;
-    /** The screen outputs informations */
-    screen_output_array_t outputs;
+    /** Access the screens by names instead of index */
```
psychon: Uhm... shouldn't this just be `The name of the screen` or something like that?

objects/screen.h
```diff
@@ -30,17 +30,27 @@
...
+    char *name;
+    /** Opaque pointer to the psysical geometry */
```
psychon: `physical` and it is a pointer to a viewport, not a physical geometry (although`struct viewport_t` was not yet defined anywhere that could be included here...)

objects/screen.h
```diff
@@ -30,17 +30,27 @@
 typedef struct screen_output_t screen_output_t;
 ARRAY_TYPE(screen_output_t, screen_output)
 
+/** Different flags useful to manage screens */
+typedef enum {
+    SCREEN_FLAGS_NONE        = 0x0 << 0, /*!< No flags                       */
+    SCREEN_FLAGS_VALID       = 0x1 << 0, /*!< Not pending garbage collection */
+    SCREEN_FLAGS_FAKE        = 0x1 << 1, /*!< Is created using fake_add      */
+    SCREEN_FLAGS_MANAGED_LUA = 0x1 << 2, /*!< Is managed internally by Lua   */
+    SCREEN_FLAGS_MANAGED_C   = 0x1 << 3, /*!< Is managed internally by Lua   */
```
psychon: `by C` instead of `by Lua`, I think...?

Could you rewrite this a bit? `NONE` can just get the value `0` and the `0x` elsewhere confuse me and can be dropped.

Also... I would split this up.
- `valid` seems to be orthogonal to the other values, so just leave that as a boolean. This also gets rid of some bit-fiddling in the implementation.
- `FAKE` is basically unused (it is set once, but never checked), so can be dropped completely.
- This leaves `MANAGES_LUA` / `MANAGED_C` / `NONE`. I am not quite sure about these. I would suggest replacing this with a `bool managed_by_c`, since i do not understand the `NONE` state. `MANAGED_LUA` is only checked for in `luaA_screen_get_managed()`, so I *think* that from the C code's perspective `NONE` and `MANAGED_BY_LUA` are the same. Lua seems to access `_managed` at one place and this could be just replaced with a table that remembers which screens are managed by `awful.screen` instead?
  - So I think that `bool managed_by_c` (or something like that) should also be an okay replacement.

objects/screen.c
```diff
@@ -94,12 +125,41 @@
...
+ * * `geometry` *(table)*: A table with an `x`, `y`, `width` and `height` keys.
+ * * `outputs` *(table)*: All outputs sharing this viewport.
+ * * `maximum_dpi` *(number)*: The DPI of the most dense output.
+ * * `minimum_dpi` *(number)*: The DPI of the least dense output.
+ * * `preferred_dpi` *(number)*: The optimal DPI.
+ *
+ * @signal property::viewports
+ * @tparam table viewports A table containing all physical viewports.
+ * @see automatic_factory
+ */
```
psychon: There is no such signal. `property::viewports` only appears four times in comments. Could you please re-read the documentation that you are adding and see if it still applies? (Or tell me where this `property::viewports` is emitted?)

Veratil:
>     * `gears.geometry.rectangle.are_equal` could be moved to its own PR in the hope of shrinking this one (sorry that I keep proposing this kind of thing; I know this causes extra work for you)
>       
>       * and I do not like `are_equal`... just "equal"? Or shall I stop bikeshedding?

I think `are_equal` fits the purpose of the function since it takes two separate parameters.

objects/screen.h
Elv13: > so I think that from the C code's perspective NONE and MANAGED_BY_LUA are the same.

Not quite. If you do `c:fake_remove()` on a screen created in C to map an output. Then you do `fake_create()` with the same resolution, it would have a different, unwanted, result.

objects/screen.c
Elv13: It is emitted from Lua. It's a proxy for `property::_viewports`. The difference is that `property::viewports` is sent after `awful` (dpi.lua) had a chance to modify the values (which, in this case, means adding the DPI). It was also added for the tests since it was easier to monkey-patch awful to set the test values. The doc is in the original commit to avoid having to add it in one commit and move it to `awful` in the next commit. That would make the diff noisier.

awesome.c
Elv13: https://github.com/awesomeWM/awesome/pull/2790/files/e69d1493f21e6d0b65d13a4c4c9b252ece486589#diff-5bb7c29334cbe6889621399f4daa8bc5

Please argue with @blueyed about the name. What you propose is what I originally did.

awesome.c
Elv13: I think this is clear enough. It has the "before or inside" part. This is a user facing message. Being  pedantic is counter intuitive.

The point of documentation and errors is to make the user *less* confused if he doesn't know the ins and outs of the code. If someone knows the code, they don't need the doc or error messages.

awesome.c
Elv13: done

globalconf.h
Elv13: done

luaa.h
Elv13: done

objects/screen.h
Elv13: done

objects/screen.h
Elv13: done

objects/screen.h
Elv13: Done-ish

```c
/** Different ways to manage screens */
typedef enum {
    SCREEN_LIFECYCLE_USER =        0, /*!< Unmanaged (ei. from fake_add) */
    SCREEN_LIFECYCLE_LUA  = 0x1 << 0, /*!< Is managed internally by Lua  */
    SCREEN_LIFECYCLE_C    = 0x1 << 1, /*!< Is managed internally by Lua  */
} screen_lifecycle_t;
```

Elv13:
I updated everything. @psychon You might want to review the final diff. First, it will prevent GitHub to send a million emails because reply to individual commit comment are not batched by default. Also, a lot of commit have been made to make sense when `git show` is used on them once this PR is merged. Individually, they make changes which only make sense in the context of commit added later. The way `property::viewport` is added is an example of that.

psychon:
I only looked at the output of `git diff f1335be21ad0cbca90c63d9a25982904b966ac65..516540156df4616089f6806eb8521d2bfb0a9301 *.[ch] objects/*.[ch]`. That's why I didn't see the signal emitted in Lua.

And yes, I only looked at the final diff and not individual commits, so I am not quite sure what you want me to do different...?

> +    SCREEN_FLAGS_MANAGED_LUA = 0x1 << 2, /*!< Is managed internally by Lua   */
> +    SCREEN_FLAGS_MANAGED_C   = 0x1 << 3, /*!< Is managed internally by Lua   */
> 
> Done-ish

This still has the "by Lua" for the `_C` define.

Hopefully I'll have some time this afternoon to look at the rest of the changes to the C code. If I then continue at this speed, I will have managed to take a look at everything by 2021...

Elv13:
Rebased without the `gears.geometry` commits (merged by the other PR) and with the copy/pasted comment "typo" fixed (well, it already was fixed, but I didn't mention it).

objects/screen.c
```diff
@@ -297,18 +334,314 @@ screen_deduplicate(lua_State *L, screen_array_t *screens)
...
+ *   slow "miss_handler" and unsafe "valid" systems used by other CAPI objects.
+ * * The `viewport_t` implements a linked-list because its main purpose is to
+ *   offers a deduplication algorithm. Random access is never required.
+ * * Everything that can be done in Lua is done in Lua.
+ * * Since the legacy and "new" way to initialize screens share a lot of steps,
+ *   the C code is bent to share as much code as possible. This will reduce the
+ *   "dead code" and improve code coverage by the tests.
+ *
+ */
+typedef struct viewport_t
```
Timidger: Shouldn't this be declared/defined in `objects/screen.h`?

Elv13: I don't think it should, it is an implementation detail. It isn't used anywhere else in the code. It is an equivalent of `private` in Java/C++, also called "pimpl".

Timidger: Yes I see now that I dig into the code more. Was confused since I usually don't use typedefs and forward declare opaque structs in my headers. Was also confused since the implementation details do seem exposed to Lua, but not to the rest of the C code (not that those are related).

Elv13:
Everybody, please review this. This is very large and will impact a lot of things coming in the near future. The less bugs it has before it gets merged, the less bugs it will have when you start to use it.

Otherwise I will merge this pull request on Saturday.

manpages/awesome.1.txt
```diff
@@ -46,6 +46,8 @@ OPTIONS
     Add a directory to the library search path.
 *-a*, *--no-argb*::
     Don't use ARGB visuals.
+*-m*, *--screen*:: 'off' or 'on'::
+    Use "off" to execute rc.lua before creating the screens.
```
Aire-One: BTW @Veratil should confirm but I think the "the" is unneeded:

```suggestion
    Use "off" to execute rc.lua before creating screens.
```

Veratil: This is sort of a mixed bag here. Both are correct grammar, but carry slightly different feel of the word "screens".

After saying it a few times to myself, I do believe removing "the" carries the proper feel for the intention of what "screens" are.

Elv13:
Well, thanks to everybody who contributed to this PR discussions. As usual, please upgrade to the latest git version and test this. "Thankfully" if there is regressions they will be catastrophic enough to show up quickly.

As to where this is going next, let me expand a bit. There is 2 major follow up to this (in no particular order). One of them is a `rules` API like `awful.rules`, but for screens. It will help quickly define how to handle adding and removing screens (DPI, what to do with the tags, splits, swaps, etc). The second one is the equivalent of this pull request, but for the tags. Just like this PR introduces the "viewport" abstraction, I will add another abstraction for the tags. I am not sure on the name yet, it could be "activities" like in KDE Plasma or "worksets" (because it means nothing). tl;dr;, it formalize the "set of tags" currently loosely represented by `my_screen.tags`, `awful.widget.taglist` and `awful.tag.viewnext/viewprevious` into a new class. This will allow to share a set of tags between multiple screens. In turn, this allows Xmonad like shared workspaces. By default, each screen will have their own, so the current behavior will be unchanged. However multiple screens can share the same "activity" and (in theory), there could be "activities" without screens to hide clients away.

The main use case for this PR and both follow up is to make AwesomeWM the best WM for dynamic screen setups. The corporate users (for meetings) and students (for presentations) are plugging/unplugging projectors or workstation monitors multiple time a day. Right now, it is very hard to do so without losing the "state" of the tags (layouts, assigned clients, etc). This is due to the fact that AwesomeWM associate the tags to the screens directly and these come and go. By adding another layer of abstraction, it becomes possible to add and remove screens without ever losing the tag(s) state.

blueyed:
Thanks, @Elv13!

PR #2794, opened 2019-06-13
ci: enable ccache

blueyed:
Ref: https://docs.travis-ci.com/user/caching/#ccache-cache

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 dist: xenial
 language: c
+cache: ccache
 
 # Build only master and stable branches.  Other branches go through PRs.
 branches:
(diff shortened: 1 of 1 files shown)


blueyed:
Looks slightly faster overall, but does not gain us too much.

Stats:
```
$ ccache -s
cache directory                     /home/travis/.ccache
primary config                      /home/travis/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
cache hit (direct)                    54
cache hit (preprocessed)               0
cache miss                             0
called for link                       12
no input file                          1
files in cache                       232
cache size                          23.4 MB
max cache size                     500.0 MB
```

blueyed:
Closing for now.. not really required / worth it, given that it could cause build issues / problems not being detected properly.

PR #2672, opened 2019-02-18
Add coverage reporting to more jobs

blueyed:
TODO:

- [x] more jobs
- [x] integrate https://github.com/awesomeWM/awesome/pull/1993
- [x] ensure there is nothing from `build` reported with coveralls

PR #2795, opened 2019-06-14
[WIP] ci: Travis: use gcovr with gcov coverage

blueyed:
Just thought giving this a try while improving coverage with Neovim, and coming across gcovr.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -190,7 +190,7 @@ install:
           # Install gcovr (once).
           (
             pyenv shell 3.7
-            pip install --quiet --user gcovr
+            pip install --quiet --user git+https://github.com/gcovr/gcovr@master#egg=gcovr
             ~/.local/bin/gcovr --version
           )
         fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -190,7 +190,7 @@ install:
           # Install gcovr (once).
           (
             pyenv shell 3.7
-            pip install --quiet --user git+https://github.com/gcovr/gcovr@master#egg=gcovr
+            pip install --quiet --user gcovr
             ~/.local/bin/gcovr --version
           )
         fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -186,28 +186,22 @@ install:
     do_codecov_gcov() {
       echo "== do_codecov_gcov in $PWD: $*: =="
       if [ "$DO_COVERAGE" = "codecov" ]; then
-        # Report coverage for each .gcno file separately.
-        # gcov will create .gcov files for the same source (e.g. for
-        # globalconf.h from awesome.c.gcno and event.c.gcno).
-        i=0
-        set -x
-        (
-          cd build
-          find -path "*/lgi-check.dir" -prune -o \( -name '*.gcno' -print \) | while read -r gcno; do
-            gcov -pb "$gcno"
+        if ! command -v gcovr; then
+          # Install gcovr (once).
+          (
+            pyenv shell 3.7
+            pip install --quiet --user gcovr
+            ~/.local/bin/gcovr --version
+          )
+        fi
+        ~/.local/bin/gcovr --branches --print-summary -j 2 --root build --delete -o coverage.xml --xml-pretty --xml
 
-            mkdir gcov.$(( ++i ))
-            mv *.gcov "gcov.$i"
+        # Debug
+        cat coverage.xml
 
-            # Delete any files for /usr.
-            # They are not relevant and might cause "Invalid path part" errors
-            # with Code Climate.
-            find "gcov.$i" -maxdepth 1 -type f -name '#usr#*.gcov' -delete
-          done
-
-          # Upload to Codecov.
-          travis_retry bash /tmp/codecov-bash -X gcov -X coveragepy -F gcov -e LUANAME,LGIVER
-        )
+        # Upload to Codecov.
+        travis_retry bash /tmp/codecov-bash -X gcov -X coveragepy -F gcov -f coverage.xml -e LUANAME,LGIVER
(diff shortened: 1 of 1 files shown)


blueyed:
Using the XML (Cobertura) format here appears to change affected branch coverage - might be legit after all, but needs to be verified.  Could also just mean thaat either gcovr is buggy with generating it, or codecov does not handle branch coverage in .gcov files properly.

Elv13:
According to https://codecov.io/gh/awesomeWM/awesome/compare/99e81c097affb570d144a1f9e0ea64872e242540...644a8e61130e14c2ed6562a3f48b31a3d16ad64d/changes

it seems that the -2% is not actually "red" loss of coverage, but rather yellow partial coverage. At first, I was under the impression that it was because some `if` with multiple logical clauses were better introspected. If `if (true || false || false || true)` is there, then the later false, false and true are not hit. However that doesn't explain why `if(keygrabber_handlekpress(L, ev))` is not yellow. I am not sure what to do about this.

psychon:
> However that doesn't explain why if(keygrabber_handlekpress(L, ev)) is not yellow. 

You mean "is yellow" instead of "is not yellow", right? Or am I looking in the wrong place?

Well, `keygrabber_handlekpress` only ever returned `true`, never `false`. Thus, of the two possible branches, only one was taken. Thus, the line is partially covered (not red, because one branch was taken; not green, because one branch was not taken).

Fun fact: `keygrabber_handlekpress` always returns `true`.

blueyed:
So, do you agree that we have better branch coverage with this after all then?
(IIRC this is also what cpp-coveralls would end up with, and therefore likely means codecov is not parsing gcov files properly the..!)
It also causes some macros to become uncovered though: https://codecov.io/gh/awesomeWM/awesome/compare/99e81c097affb570d144a1f9e0ea64872e242540...644a8e61130e14c2ed6562a3f48b31a3d16ad64d/changes#L67 (`HANDLE_TEXT_PROPERTY` in `property.c`).

blueyed:
As for cpp-coveralls see https://github.com/awesomeWM/awesome/pull/2672:

> coverage/coveralls — Coverage decreased (-3.5%) to 85.917%

blueyed:
For reference: gcovr master:
```
++pip install --quiet --user git+https://github.com/gcovr/gcovr@master#egg=gcovr
[33mYou are using pip version 10.0.1, however version 19.1.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.[0m
++/home/travis/.local/bin/gcovr --version
gcovr 4.1

Copyright 2013-2018 the gcovr authors
Copyright 2013 Sandia Corporation
Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
the U.S. Government retains certain rights in this software.
++/home/travis/.local/bin/gcovr --branches --print-summary -j 2 --root build --delete -o coverage.xml --xml-pretty --xml
lines: 75.1% (5679 out of 7565)
branches: 57.4% (2537 out of 4420)
++cat coverage.xml
<?xml version='1.0' encoding='UTF-8'?>
<!DOCTYPE coverage SYSTEM 'http://cobertura.sourceforge.net/xml/coverage-04.dtd'>
<coverage line-rate="0.7506939854593523" branch-rate="0.5739819004524886" lines-covered="5679" lines-valid="7565" branches-covered="2537" branches-valid="4420" complexity="0.0" timestamp="1560591549" version="gcovr 4.1">
```

current stable:
```
++pip install --quiet --user gcovr
[33mYou are using pip version 10.0.1, however version 19.1.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.[0m
++/home/travis/.local/bin/gcovr --version
gcovr 4.1

Copyright 2013-2018 the gcovr authors
Copyright 2013 Sandia Corporation
Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
the U.S. Government retains certain rights in this software.
++/home/travis/.local/bin/gcovr --branches --print-summary -j 2 --root build --delete -o coverage.xml --xml-pretty --xml
lines: 75.1% (5690 out of 7575)
branches: 57.4% (2538 out of 4420)
++cat coverage.xml
<?xml version="1.0" ?>
<!DOCTYPE coverage
  SYSTEM 'http://cobertura.sourceforge.net/xml/coverage-04.dtd'>
<coverage branch-rate="0.57420814479638" branches-covered="2538"
 branches-valid="4420" complexity="0.0" line-rate="0.7511551155115511"
 lines-covered="5690" lines-valid="7575" timestamp="1560560902" version="gcovr
 4.1">
```

So it looks similar - at least from the header / start.

latk:
Hi, I'm the gcovr maintainer and will be happy to help with any questions.

> At first, I was under the impression that it was because some if with multiple logical clauses were better introspected. If `if (true || false || false || true)` is there, then the later false, false and true are not hit. However that doesn't explain why `if(keygrabber_handlekpress(L, ev))` is not yellow. I am not sure what to do about this.

Gcov and gcovr use assembly level branch coverage data. If the compiler puts a branch somewhere, it will turn up in the statistics. Syntax-level control flow is *not* looked at. This also means your branch coverage results may be quite dependent on compiler optimization level. Unfortunately codecov only distinguishes between covered/partial/uncovered, and will not display more detailed info about the branches on a line.

> It also causes some macros to become uncovered though.

That seems odd, considering that gcovr is parsing the same gcov data that you used before and that the XML in the Travis build log shows those macros as covered. My hunch is that codecov gets confused by the 0% branch coverage on some of those macros. This is impossible for normal functions, but your macros generate multiple functions. Thus, executing one of those function marks the macro as line-covered, but the function with the branch might not have been invoked (`property_handle_...()`).

PR #2796, opened 2019-06-14
tests/run.sh: cat logfile with unsuccessful tests

blueyed:
It wasn't displaying the log on hangs before, and should display it
altogether already for missing success indicator.

Would have helped with
https://travis-ci.com/awesomeWM/awesome/jobs/208264756#L2577.

/cc @Elv13


blueyed:
Oh well, the log gets tail-f'ed already - so there is nothing new there then.

PR #2804, opened 2019-06-29
Fix partially rounded rect breaking with some layout

ndgnuh:
I fix the bug from [this post](https://www.reddit.com/r/awesomewm/comments/c63ia0/wibox_with_shapepartially_rounded_rect_breaks/)

### DIFF
--- a/lib/gears/shape.lua
+++ b/lib/gears/shape.lua
@@ -141,6 +141,13 @@ function module.partially_rounded_rect(cr, width, height, tl, tr, br, bl, rad)
         cr:line_to(0, height)
     end
 
+    -- Close the path
+    -- If arc was drawn at top left then the end point would be (0, rad)
+    if tl then
+        cr:line_to(0, rad)
+    else
+        cr:line_to(0, 0)
+    end
     cr:close_path()
 end
 


psychon:
Thanks a lot for this patch and sorry that I am making things complicated.

Could you provide some more information about the problem that you are fixing?
I tried reproducing based on https://i.redd.it/g736b547dv631.png and the result is:
foo
```diff
diff --git a/awesomerc.lua b/awesomerc.lua
index 5d2bd2c10..06924e574 100644
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -206,14 +207,45 @@ screen.connect_signal("request::desktop_decoration", function(s)
 
     -- @DOC_WIBAR@
     -- Create the wibox
-    s.mywibox = awful.wibar({ position = "top", screen = s })
+    s.mywibox = awful.wibar({ position = "top", screen = s, height = 30 }) -- random guess
 
     -- @DOC_SETUP_WIDGETS@
     -- Add widgets to the wibox
+    s.info = wibox.widget({
+        {
+            {
+                wibox.widget({
+                    markup = "<span color='#ffffff'>user</span>",
+                    font = 'Segoe UI bold 11',
+                    align = 'center',
+                    widget = wibox.widget.textbox,
+                }),
+                {
+                    {
+                        mytextclock,
+                        wibox.widget.textbox("VOL"),
+                        layout = wibox.layout.fixed.horizontal, -- original code had a string here 'this cause the wibox to become like that'
+                    },
+                    widget = wibox.container.margin, -- Uhm... a margin container, but no margins are set?
+                },
+                layout = wibox.layout.ratio.vertical,
+            },
+            margins = 0,
+            widget = wibox.container.margin, -- This margin container does not do anything either?
+        },
+        bg = '#f00', -- original used beautiful.shade,
+        widget = wibox.container.background,
+        border_width = 1, -- original used beautiful.taglist_shape_border_width
+        border_color = '#0f0', -- original used beautiful.taglist_shape_border_color
+        shape = gears.shape.partially_rounded_rect, -- original used beautiful.common_shape(dpi(16)),
+        forced_width = 284, -- original used dpi(284),
+        forced_height = s.mywibox.height, -- original used beautiful.wibar_height
+    })
     s.mywibox:setup {
         layout = wibox.layout.align.horizontal,
         { -- Left widgets
             layout = wibox.layout.fixed.horizontal,
+            s.info,
             mylauncher,
             s.mytaglist,
             s.mypromptbox,
```
I may not be seeing the obvious, but... your patch shouldn't do anything, I think...?
`close_path` should add a line to the initial point of the path. If `tl` is false, then the path was started with `cr:move_to(0, 0)`, so in this case `cr:close_path()` should do the same as `cr:line_to(0, 0)` that your patch adds. If `tl` is true, then the path was started with `cr:arc(rad, rad, rad, math.pi, 3*math.pi/2)`. Your patch adds a `cr:line_to(0, rad)`. I am too lazy to do the math right now, but I expect this to start the path at `(0, rad)`. Mostly because: If this were something else, then the shape would be quite broken.

Anyway, what I am trying to say is: I do not understand the problem that you are fixing and thus worry that the problem could actually be elsewhere. For example, did something else leave some path around that should not be here? Or is there a bug in cairo? (Since your screenshot clearly shows some discontinuity and `cairo_close_path()` should not generate something like that.)

Again, sorry for the many questions

psychon:
I just looked at the Travis build failure: Busted did a new release with a broken download URL: https://github.com/Olivine-Labs/busted/issues/602

psychon:
I was so close, but missed the most important part:
```
        shape = function(cr, width, height) return gears.shape.partially_rounded_rect(cr, width, height, 3, 3, 3, 3) end, -- original used beautiful.common_shape(dpi(16)),
```
This allows to reproduce the problem.

Also, the problem really is that there is already a `move_to` on the path before the shape is drawn...

ndgnuh:
Sorry for providing such bad bug reproduction code. After testing stuff, i found out that the thing that broke wasn't the shape itself, but the border of the shape. Also, the layout doesn't cause the bug. A `widget.textbox` caused it. 

Here's the widget that caused the bug:
```lua
local mywibox = wibox({
	x       = 100,
	y       = 200,
	width   = 400,
	height  = 64,
	bg = '#313131',
	ontop   = true,
	visible = true,
})

local testwidget = wibox.widget({
	{
		markup = 'asdasadas',
	},
	widget       = wibox.container.background,
	border_width = 10, -- if i set this to 0, the shape renders normally
	border_color = "#FF0000",
	bg = '#fff',
	shape        = function(cr, w, h)
		return gears.shape.partially_rounded_rect(cr, w, h, true, false, true, false)
	end,
})

mywibox:setup({
	wibox.widget.textbox("abc"),
	testwidget,
	wibox.widget.textbox("xyz"),
})
```

image

without the border:
image

without the `widget.textbox`:
image

with the patch:
image

The patch did help, the border didn't render too good though. But like you said, the problem real could be something else, i haven't test on other shape.

Also i don't know why Travis build failed, i did use `luacheck` and build everything successfully on my pc, including the doc.

Elv13:
> @Elv13 What do you think?

Number 2 looks promising. I wonder if it will break something else. Maybe we should just merge this pull request too? Technically calling close_path from a point other than the origin is asking for troubles.

psychon:
Opened #2805 and #2806 based on (1) and (2) above. I doubt a lot that any of these patches would break anything, because it would require assuming that (for (1)) one can influence the shape based on a random pre-existing path, which just asks for troubled, or that (for (2)) one can just pass a path from one widget to another, which asks for trouble and might easily break if only one of the two widgets involved is repainted.

Elv13:
The bug should not happen anymore thanks to @psychon improvements to the general robustness of the shapes. 

Thanks to both of you for reporting this and taking the time to investigate.

PR #2816, opened 2019-07-09
Fix 4 "new API" notification bugs

Elv13:
Well, updating now works "better". Better as-in, I have zero apps that use this, but tested using some Lua scripts and it now works for me. 2 of the "fixes" are more for general robustness than major issues. The signal typo is quite a facepalm.

```lua
local Gio       = require("lgi"        ).Gio
local GLib      = require("lgi"        ).GLib
local main_loop = GLib.MainLoop()

local dbus_connection = assert(Gio.bus_get_sync(Gio.BusType.SESSION))

local counter = 1

local actions = { "1", "one", "2", "two", "3", "three" }

local function send_notify(app, id, icon, summary, body, actions, hints, timeout, callback)
    local parameters = GLib.Variant("(susssasa{sv}i)", {
        app, id, icon, summary, body, actions, hints, timeout
    })
    local reply_type = GLib.VariantType.new("(u)")

    local function invoke_callback(conn, result)
        local new_id = conn:call_finish(result).value[1]
        print("RESULT", new_id)
        counter = counter + 1

        if counter >= 5 then os.exit(0) end

        os.execute("sleep 3")

        send_notify("app", new_id, "", "one two three"..counter, "body"..counter, actions, {}, 25000)
    end

    print("CALL")

    dbus_connection:call("org.freedesktop.Notifications",
        "/org/freedesktop/Notifications", "org.freedesktop.Notifications",
        "Notify", parameters, reply_type, Gio.DBusCallFlags.NO_AUTO_START, -1,
        nil, invoke_callback)
end

GLib.idle_add(GLib.PRIORITY_DEFAULT_IDLE, function()
    send_notify("app", 9999, "", "one two three", "body", actions, {}, 25000)
end)

main_loop:run()
```

### DIFF
--- a/lib/naughty/action.lua
+++ b/lib/naughty/action.lua
@@ -44,6 +44,9 @@ local action = {}
 -- @tparam gears.surface|string icon
 
 --- If the action should hide the label and only display the icon.
+--
+-- @DOC_wibox_nwidget_actionlist_icon_only_EXAMPLE@
+--
 -- @property icon_only
 -- @param[opt=false] boolean
 
@@ -125,9 +128,12 @@ local function new(_, args)
         position     = args.position,
         icon         = args.icon,
         notification = args.notification,
+        icon_only    = args.icon_only or false,
     }
 
     rawset(ret, "_private", default)
+
+    gtable.crush(ret, args)
 
     return ret
 end
--- a/lib/naughty/action.lua
+++ b/lib/naughty/action.lua
@@ -103,6 +103,16 @@ for _, prop in ipairs { "name", "icon", "notification", "icon_only" } do
     end
 end
 
+local set_notif = action.set_notification
+
+function action.set_notification(self, value)
+    local old = self._private.notification
+    set_notif(self, value)
+    if old then
+        old:emit_signal("property::actions")
+    end
+end
+
 --- Execute this action.
 --
 -- This only emits the `invoked` signal.
--- a/lib/naughty/container/background.lua
+++ b/lib/naughty/container/background.lua
@@ -48,13 +48,13 @@ function background:set_notification(notif)
     if self._private.notification == notif then return end
 
     if self._private.notification then
-        self._private.notification:disconnect_signal("poperty::bg",
+        self._private.notification:disconnect_signal("property::bg",
             self._private.background_changed_callback)
-        self._private.notification:disconnect_signal("poperty::border_width",
+        self._private.notification:disconnect_signal("property::border_width",
             self._private.background_changed_callback)
-        self._private.notification:disconnect_signal("poperty::border_color",
+        self._private.notification:disconnect_signal("property::border_color",
             self._private.background_changed_callback)
-        self._private.notification:disconnect_signal("poperty::shape",
+        self._private.notification:disconnect_signal("property::shape",
             self._private.background_changed_callback)
     end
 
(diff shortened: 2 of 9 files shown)


vn-ki:
@Elv13 This doesn't work with the new `naughty.notification`. See image

2019 07 09-22 35 38 screenshot

With the old `naughty.notify` api, it works. But the timeout for the notification uses the first notification's timer (I think). This makes the timeout different for later notifications.

<details>
<summary>The code I used, in case</summary>

```lua
  if naughty.notification ~= nil then
    notification = naughty.notification({
        title = title,
        text = artist,
        icon = notification_icon,
        timeout = 20,
        replaces_id = last_notification_id,
        actions = {
          naughty.action {
            name = 'Prev',
            icon = beautiful.playerctl_prev_icon,
            icon_only=true
          },
          naughty.action {
            name = 'Play/Pause',
            icon = beautiful.playerctl_toggle_icon,
            icon_only=true
          },
          naughty.action {
            name = 'Next',
            icon = beautiful.playerctl_next_icon,
            icon_only=true
          },
        }
    })
    last_notification_id = notification.id
  else
    notification = naughty.notify({
        title = title,
        text = artist,
        icon = notification_icon,
        timeout = 4,
        replaces_id = last_notification_id
    })
    last_notification_id = notification.id
  end

```

</details>

Elv13:
@vn-ki This is normal, the new API doesn't have a `replace_id`. You are creating a new notification object, which is a new notification. You need to update the old notification object (`my_old_notification.message = "foo"`). `naughty.notify` is a proxy that may or may not create new objects, `naughty.notification()` is directly the constructor of a new notification.

Elv13:
> Does any of this warrant the addition of a new test?

Yes. I will add them when I submit the last part. As seen in @vn-ki example, having custom notification for some apps is currently cumbersome. The last series of patches will add support for an `awful.rules` like API (sharing most of the code) to filter and modify notifications before they are displayed. There is a work done in that PR, but given I "saw big" and went on to add `awful.rules` to tags, screens, notifications, clipboard (and of course clients), this is still not polished enough.

I say that because one of the pull request will add a better test template/framework to test those details, see https://elv13.github.io/declarative%20rules/ruled.tag.html (disregard the API, it is a very early draft, but look at the "pretty pictures"). It has steps like the integration tests, but better built in tooling to compare stuff across steps.

vn-ki:
> This is normal, the new API doesn't have a replace_id.

Oh. Alright. I think I saw replaces_id in the docs.

> This makes the timeout different for later notifications.

You might have missed this @Elv13

Elv13:
> You might have missed this @Elv13

Indeed. Please report another issue (or better, a pull request). It's a bit unrelated to what this PR attempts to fix.

actionless:
> This is normal, the new API doesn't have a replace_id. You are creating a new notification object, which is a new notification. You need to update the old notification object (my_old_notification.message = "foo"). naughty.notify is a proxy that may or may not create new objects, naughty.notification() is directly the constructor of a new notification.

what about printing deprecation warning when it used?


also documentation for naughty.notification{} still have it documented: file:///usr/share/doc/awesome/doc/core%20components/naughty.notification.html#naughty.notification

Ryuno-Ki:
We can't access your machine, @actionless (i.e. file: protocol instead of a web resource with https:).

Elv13:
> also documentation for naughty.notification{} still have it documented: file:///usr/share/doc/awesome/doc/core%20components/naughty.notification.html#naughty.notification

whoops. I will remove it. It cannot work the way it's implemented. `naughty.notification()` is a constructor, if it was to try to supersede another object, it would explode elsewhere.

actionless:
@Ryuno-Ki everybody _must_ have there docs from the latest git master ;)

Elv13:
Merging this for now. I will open a new one for the next batch.

PR #2820, opened 2019-07-09
doc: Fix a rendering regression regarding backgrounds clipping changes

### LINKED ISSUE #2727: Since the new background shape implementation, the SVG no longer work
See https://awesomewm.org/apidoc/documentation/03-declarative-layout.md.html

vs.

https://awesomewm.org/doc/api/documentation/03-declarative-layout.md.html

Another problem I see is that the SVG are now full of base64 png images. As far as I recall, it was fully vector based until recently. This is a bit problematic for debugging since the old SVG + WebBrowser introspection mode was very useful to track the results.

Maybe we could revert these commit and think outside the box to solve the problem (ei, more fine grained APIs). The notification pull requests adds a commit to mitigate the border offset from the shape, just that is enough to make some instances of the problem go away (but it isn't enough, making the border mode user selectable will fix another class of issues).

Elv13:
The way background are rendered changed to accomodate issues regarding
cliping and border. However this broke the documentation examples.

This commit fixes this in the least hacky way I found. @psychon if you have a better idea to solve:

 1) the rendering bug (which you reported upstream)
 2) the fact that it renders everything as base64 raster within a SVG container

Using raster images within SVG is probably what you expect from `cr:set_source_surface`, so I guess the current behavior is expected, but unwanted. This doesn't work well with matrices and render blurry results. There is probably issues/regressions in real wibox themselves when using such matrix. Is that a problem when nobody complains (yet)? Should the clipping change be reverted or should this PR be used? @actionless would be disappointed that the background border will regress *again*, but on the other hand the older code is simpler and has less side effects like this.

Fixes #2727

### DIFF
--- a/lib/wibox/container/background.lua
+++ b/lib/wibox/container/background.lua
@@ -22,6 +22,61 @@ local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility
 
 local background = { mt = {} }
 
+-- The Cairo SVG backend doesn't support surface as patterns correctly.
+-- The result is both glitchy and blocky. It is also impossible to introspect.
+-- Calling this function replace the normal code path is a "less correct", but
+-- more widely compatible version.
+function background._use_fallback_algorithm()
+    background.before_draw_children = function(self, _, cr, width, height)
+        local bw    = self._private.shape_border_width or 0
+        local shape = self._private.shape or gshape.rectangle
+
+        if bw > 0 then
+            cr:translate(bw, bw)
+            width, height = width - 2*bw, height - 2*bw
+        end
+
+        shape(cr, width, height)
+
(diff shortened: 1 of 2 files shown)


Elv13:
> Instead of having wibox.container.background "monkey-patch itself", one could also monkey patch it in tests/. Is this option less ugly and your proposed patch? I don't know, hence this weird review.

This is what my original implementation did. However I went this route because the changes to sub-surfaces might have other side effects beyond the documentation. In that case, using `wibox.background._use_fallback_algorithm()` will exist as a magic workaround to temporarily fix the issues until we can improve it further in future releases. I am still not convinced *all* matrix scaling corner cases are property handled with the new code path.

I feel like merging this version rather than move the code back to the shims.

PR #2821, opened 2019-07-10
do reset_timeout when notification content changes

vn-ki:
Referenced [here](https://github.com/awesomeWM/awesome/pull/2816#issuecomment-509931314)

This probably needs a test. I'm not sure where to add it and how. :/

### DIFF
--- a/lib/naughty/widget/message.lua
+++ b/lib/naughty/widget/message.lua
@@ -70,6 +70,7 @@ local function new(args)
 
     function tb._private.message_changed_callback()
         tb:set_markup(markup(tb._private.notification, tb))
+        tb._private.notification:reset_timeout()
     end
 
     if args.notification then


Elv13:
Thanks for the patch

> This probably needs a test. I'm not sure where to add it and how. :/

Don't worry about it for now. Testing time is harder than degular tests. You would have to shim the timer API to provide a way to trigger timeouts on demand. Then there isn't an integration test suit for this module anyway. Only usage tests, but they are even worst at testing this kind of stuff (but I am working on a solution).

Elv13:
@vn-ki Would https://github.com/awesomeWM/awesome/pull/2823/commits/30b0dfaac2d26f2eefac68ade32407be5e5e4089 work? This is a more generic solution. It would be more friendly to the future notification `rules` API (which will semi-replace the old preset API).

vn-ki:
Yes, that's probably better. I haven't tested that though.

PR #2825, opened 2019-07-14
Support the notification spec v1.2

### LINKED ISSUE #2729: Notification actions disappering when replaces_id is set

Elv13:
Little testing ~~and no automated tests at all.~~

     * action icons (requires to implement `naughty` `request::icon` to avoid a dependency on `menubar`)
     * persistence (does nothing, but add a property to advertise support)
     * residence (may work, I didn't find a real app that implements it)
     * categories (should work)
     * add an urgency property (expose a previously hidden API)
     * animated icons (on the object side, not the widgets)
     * more ways to get icons

The most noticeable change is that it can now use the client icons as a fallback since I also added a rather convoluted way to map clients to notifications.

@actionless Chrome still refuses to let me try those web notification things, but this PR has some homework for you. If something doesn't work, please figure out why.

**_actionless_**:  closes #2729

Spec: https://developer.gnome.org/notification-spec/#icons-and-images

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -20,34 +20,15 @@ local hotkeys_popup = require("awful.hotkeys_popup")
 require("awful.hotkeys_popup.keys")
 
 -- {{{ Error handling
--- @DOC_ERROR_HANDLING@
 -- Check if awesome encountered an error during startup and fell back to
 -- another config (This code will only ever execute for the fallback config)
-if awesome.startup_errors then
+naughty.connect_signal("request::display_error", function(message, startup)
     naughty.notification {
-        preset  = naughty.config.presets.critical,
-        title   = "Oops, there were errors during startup!",
-        message = awesome.startup_errors
+        urgency = "critical",
+        title   = "Oops, an error happened"..(startup and " during startup!" or "!"),
+        message = message
     }
-end
-
--- Handle runtime errors after startup
-do
-    local in_error = false
-    awesome.connect_signal("debug::error", function (err)
-        -- Make sure we don't go into an endless error loop
-        if in_error then return end
-        in_error = true
-
-        naughty.notification {
-            preset  = naughty.config.presets.critical,
-            title   = "Oops, an error happened!",
-            message = tostring(err)
-        }
-
-        in_error = false
-    end)
-end
+end)
 -- }}}
 
 -- {{{ Variable definitions
--- a/lib/gears/object.lua
+++ b/lib/gears/object.lua
@@ -134,6 +134,42 @@ function object:emit_signal(name, ...)
     end
 end
 
+function object._setup_class_signals(t)
+    local conns = {}
+
+    function t.connect_signal(name, func)
+        assert(name)
+        conns[name] = conns[name] or {}
+        table.insert(conns[name], func)
+    end
+
+    --- Emit a notification signal.
+    -- @tparam string name The signal name.
+    -- @param ... The signal callback arguments
+    function t.emit_signal(name, ...)
+        assert(name)
+        for _, func in pairs(conns[name] or {}) do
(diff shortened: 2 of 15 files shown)


Elv13:
I added some commits to test this, fixed half a dozen bug and also added more stuff to make the rule API more usable:

 * Expose the urgency and category
 * Make sure actions can be shared by multiple notification (warning: break unreleased APIs)
 * Begin to deprecate the old preset API in favor of "real" rules (the rule part is not in this PR)
 * Import the "request::display_error" from some other PRs

actionless:
the "testcase" with chrome is finally working :tada:

actionless:
@elenapan i think here's a future fix for that: https://github.com/awesomeWM/awesome/pull/2790/files#diff-caa2908d246dcb889391ddaeda8fd13f

Elv13:
> Notifications do not appear for errors.

Fixed, thanks for noticing.

> i think here's a future fix for that: 

Spot on! Both PRs have different versions of this commit. The other one introduces a better API to handle events during startup and use it to mitigate this issue. This pull request had no equivalent so it has this bug. I added a commit I will revert if this gets merged before the screen refactoring PR (which is looking likely). It uses the "startup" signal to delay the notification long enough to give a window of opportunity for the handler to be connected first. But it's a little "late" during the startup sequence compared to the current situation. The later it is, the higher the chance there will be more errors "first". This will cause the startup errors to be below other notifications (potentially errors), which is confusing. However as a temporary fix to address 2 different pulls requests "conflicting", this is good enough.

Elv13:
Anyone tried the latest version? Any other issue? I feel like merging this and see who complain. This is a lot less regression prone than the big rewrite pull requests. The action changes will probably break a couple things from early adopters, but the way it implemented would never have worked. As actionless points out, supporting v1.2 in itself solves a bunch of issues (probably caused by clients checking the advertised supported version).

elenapan:
The latest changes work fine on my end.

actionless:
noticed weird issue

some notifications from chrome (for example from reddit) are displayed with error template (ie with red background)

but since so far it was happening only with reddit notifications, idk how easy it would be to reproduce it

Elv13:
> are displayed with error template

This is a feature. They now *can* request to be displayed as urgent. Technically they already could, but the client "1.0" codepath was probably too conservative to try.

The `rules` pull request will make it trivial to disallow. I don't think this a blocker for this pull request.

actionless:
now when i know the reasons of such behavior -- it's indeed not a blocker

and configuring that behavior will be possible with that `rules` PR, right?

Elv13:
Ok, thanks for testing this, both of you.

If someone wants to do a typo review, that would be appreciated. Otherwise I will merge this during the weekend to fix the open issues.

Elv13:
This is now merged. Please upgrade, test and crash this new code and report back. The earlier we find the flaws, the earlier they get fixed. Again, this should not be as problematic as the rewrite itself. Some application may magically decide to behave differently now that AwesomeWM advertise to support the protocol v1.2 rather than 1.0. Icons will probably work a lot more often now, so this will be visible.

> and configuring that behavior will be possible with that rules PR, right?

Absolutely, a trivial one liner. See file:///home/lepagee/dev/awesome/build/doc/declarative_rules/ruled.notifications.html

Issue #2832, opened 2019-07-21
Something left on the lua stack, lgi async


DorianGray:
Output of `awesome --version`:

awesome v4.3-314-g1116bc47 (Too long)
 • Compiled against Lua 5.1.4 (running with LuaJIT 2.0.5)
 • D-Bus support: ✔
 • xcb-errors support: ✘
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

**How to reproduce the issue:**
I hate callbacks
```
local lgi = require 'lgi'
local awful = require 'awful'
local awesome = require 'awesome'

local function read_stream(st)
  local unix_stream = lgi.Gio.UnixInputStream.new(st, true)
  local stream = lgi.Gio.DataInputStream.new(unix_stream)

  local output = {}
  while not stream:is_closed() do
    local line, len = stream:async_read_line()
    if line then
      table.insert(output, line)
    else
      stream:async_close()
    end
  end
  return table.concat(output, '\n')
end

local function run(command, shell)
  if not shell then
    shell = false
  elseif shell == true then
    shell = awful.util.shell
  end
  local pid, _, stdin, stdout, stderr = awesome.spawn(
    shell and {shell, '-c', command} or command,
    false,
    false,
    true
  )

  if type(pid) == "string" then
    return pid
  end

  return lgi.Gio.Async.call(gears.protected_call)(read_stream, stdout)
end

run('cat /proc/net/wireless', true)

```

The above code works fine with no errors, but if I run the last line inside of a gears timer...

```
gears.timer({
  timeout=10,
  autostart=true,
  callback=function()
    run('cat /proc/net/wireless', true)
  end,
})
```
**Actual result:**

```

2019-07-20 23:06:29 W: awesome: a_glib_poll:451: Something was left on the Lua stack, this is a bug!
-------- Lua stack dump ---------
2: userdata     #16     0x419ac128
1: userdata     #208    0x419b3128
------- Lua stack dump end ------
2019-07-20 23:06:29 W: awesome: a_glib_poll:451: Something was left on the Lua stack, this is a bug!
-------- Lua stack dump ---------
2: userdata     #16     0x41c8d270
1: userdata     #208    0x419b3128
------- Lua stack dump end ------
2019-07-20 23:06:29 W: awesome: a_glib_poll:451: Something was left on the Lua stack, this is a bug!
-------- Lua stack dump ---------
2: userdata     #16     0x402543a8
[...]

PR #2833, opened 2019-07-21
awful.tooltip:show() - swap timer.started and timer_function call

DorianGray:
This is needed because if async code is run inside of a tooltip timer func the started property may be true by the time we call timer:start().
The current version causes random spurious timer already started errors.
This fix should also result in more uniform timer behavior by making the timer fire on time regardless of if the timer_function has finished executing yet.

### DIFF
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -172,8 +172,8 @@ local function show(self)
     if self._private.visible then return end
     if self.timer then
         if not self.timer.started then
-            self:timer_function()
             self.timer:start()
+            self:timer_function()
         end
     end
     set_geometry(self)


Elv13:
> regardless of if the timer_function has finished executing yet.

This smells like you are calling synchronous code in the tooltip. This should really, really be avoided. `io.popen` and `os.execute` are **not** supported in AwesomeWM (see the FAQ and `awful.spawn` doc). Any bug caused by doing this is likely to be triaged as "wont fix". Using these function does not and cannot work flawlessly. This is why there is async equivalents available. If on the other hand you *are* using the async ones and this causes a race condition, then it will require a bit more logic to synchronize the 2 events (externally from the tooltip timer code).

I am not sure about the patch. If the tooltip is supposed to be visible for 5 seconds but your synchonous code takes 6 seconds, then with this patch it becomes even more inconsistent?

DorianGray:
I am not running synchronous code in the tooltip, in fact I am using lgi.Gio.Async.call and leveraging LGI's aio facilities to read `acpi`. I have confirmed that my code is not blocking the event loop at all. You can see the issue I opened for code that reproduces this problem if you'd like.

DorianGray:
https://github.com/awesomeWM/awesome/issues/2832

DorianGray:
I can make the tooltip timer behavior identical to the way it is not by adding another condition, but I feel pretty strongly that just swapping the two lines is a better approach overall.

DorianGray:
It works well with no crashes, but there seems to be a bug somewhere in lgi or awesome that ends up with trash left on the lua stack(linked issue).

It took several attempts to get working and until psychon's fixes to lgi it wasn't really possible to detect end of stream reliably. I'm planning on building a standalone library to do process management, but I need to get reading bytes working instead of lines. I'd also prefer to break from the existing api's usage of callbacks and instead use coroutines as iterators for line callbacks, for instance... but we'll see. I haven't been able to get lgi to mesh well with lua cooutines either for some reason... like I'd call resume on a lua coroutine from an lgi async context and get nil regardless of what I sent in the resume.

Elv13:
> a bug somewhere in lgi or awesome that ends up with trash left on the lua stack(linked issue).

Probably LGI, Awesome prints the error only because it has better ways to detect them.

> I'd also prefer to break from the existing api's usage of callbacks and instead use coroutines as iterators for line callbacks

In a perfect world, everybody agrees this makes much more readable code. I would love so that all signal handlers are (LGI async) coroutines. This means only the code executed directly in _G would be sync. Logic code in _G will slowly become less and less recommended since, for example, the new screen rules API create screen after _G is executed (but before the mainloop starts). This means a lot of the API like `mouse`, `screen` and `wibox` don't work in _G when the screens are set to be created by rules.

Last time I tried coroutines to write a "pretty" dbus binding, it wasn't usable. But that was more than 5 years ago and while LGI hasn't changed much since then, many of the bugs in question have been fixed. Plus there is a lot of case within the Awesome code to safeguard before allowing this. For example, the widget drawing code has to run in the main context and the mutex has to be aware of this. A lot of code already works fine because it makes heavy use of `delayed_call`, which makes it "safer" than calling code from coroutines (because, for now, it is sync). But stuff like `fake_input`, `awful.keygrabber` and friends will become a pain to debug. Also making all the `delayed_call`s coroutine would currently make a lot of things explode.

psychon:
> I am not running synchronous code in the tooltip, in fact I am using lgi.Gio.Async.call and leveraging LGI's aio facilities to read acpi.

I would consider `Async.call` to be synchronous (and `Async.start` to be asynchronous). It runs some code and waits for it to be done. Inside of that you can do async calls and at the same time other code can run (e.g. a timer can fire). However, `Async.call` only returns after the called function is done. My proposed patch for #2832 would break your code.

I am not really sure if we want to allow `Async.call`. As you see, it causes quite some problems with the existing code base that expects things to be non-reentrant.

Since this is Lua where functions have upvalues: Can't you use `Async.start` instead of `Async.call`? I would expect that this would make at least some of the issues disappear that you run into.

----

For this specific issue: Do I understand correctly? Inside of your `timer function`, you do an `Async.call` that takes quite a long time. Inside this call, the normal event loop runs and more input events are handled. This causes another call to the `show` function in `awful.tooltip` which then starts the timer. Later, the first call to `timer_function` returns and tries to start the timer again that was already started?

DorianGray:
Semantic aside, the preferred asynchronous methodology in the lua world is coroutines which behave like synchronous code to the developer but actually allow other coroutines to run while the current one is yielded. My end goal is to do away with callbacks in most of awesome, except where they make sense. My goal is to slowly make awesome tolerant of re-entrant code.

If I were to use async.start, I would end up creating a coroutine to grab the callback's result which would end up with the same architecture with an extra layer...so no, I don't think async.start would be better.

in my code, I have a function that finishes in milliseconds but yields the event loop. Because it yields and the behavior of tooltips is to call show() whenever a mouse enters it's space, show is called multiple times consecutively. This change fixes the timer to be tolerant of re-entrant code while leaving the behavior of non-re-entrant code basically unchanged.

DorianGray:
Specifically, what I'm building right now is a library that will capture pipe output in lua and read files asynchronously while appearing to be synchronous to the developer, with optional callbacks and iterators over lines, for instance.

When I say iterators, I mean something like 
```
for line in file:lines() do
  print(line)
end
```
where only 1 line is ever in memory at a time.

I'm also building a form library that would benefit from this, since I intend on it saving config eventually.

Issue #2835, opened 2019-07-23
gears.wallpaper.prepare_context - delayed call


DorianGray:
latest master of awesome and lgi

After adding some lgi async code with gears.protected_call in a totally unrelated part of the code, the gears.wallpaper.prepare_context function in the delayed call section suddenly sees the pending_wallpaper variable as nil.

I determined that the function is only being called once, but apparently the delayed call is being called before the pending_wallpaper variable is set later in the function? My wallpaper is being drawn correctly and on all screens... can someone explain the purpose of the delayed call here?

Elv13:
> can someone explain the purpose of the delayed call here?

Setting a wallpaper on a large root surface (let say 5 screens) is very slow (> 1 seconds). The last thing you want is to do it 5 time in the same event loop iteration (and freeze for > 5 seconds).

psychon:
> I determined that the function is only being called once, but apparently the delayed call is being called before the pending_wallpaper variable is set later in the function? 

I did another quick look at this function. So, if `pending_wallpaper` is `nil`, then `gears.wallaper.prepare_context()` only calls `cairo` between the calls to `gears.timer.delayed_call` and setting `pending_wallpaper`. Are you somehow messing with LGI's cairo bindings (=monkey patching functions in there)?

Could you dump a call to `print(debug.traceback())` in the delayed call in `gears.wallpaper` to get us at least the Lua part of the traceback that got us here? That should point out how you managed to start a `GMainLoop` before `gears.wallpaper.prepare_context` returned.

DorianGray:
Yes, I'll look into it and dig deeper into awesome's code. I want to learn how this occurred myself, but it sounds like since @Elv13 wants to deprecate this method of setting a wallpaper anyways it might not be that important to fix now anyways.

PR #2836, opened 2019-07-23
Fix nil pending_wallpaper error

DorianGray:
See https://github.com/awesomeWM/awesome/issues/2835

Under certain circumstances, the delayed_call  callback is run when pending_wallpaper is nil. Specifically, adding local lgi async contexts changed the call order so this could occur. 

The wallpaper.maximized function continues to work fine on 2 screens despite pending_wallpaper being nil at the time of the delayed_call.

### DIFF
--- a/lib/gears/wallpaper.lua
+++ b/lib/gears/wallpaper.lua
@@ -66,10 +66,12 @@ function wallpaper.prepare_context(s)
 
         -- Set the wallpaper (delayed)
         timer.delayed_call(function()
-            local paper = pending_wallpaper
-            pending_wallpaper = nil
-            wallpaper.set(paper.surface)
-            paper.surface:finish()
+            if pending_wallpaper then
+                local paper = pending_wallpaper
+                pending_wallpaper = nil
+                wallpaper.set(paper.surface)
+                paper.surface:finish()
+            end
         end)
     elseif root_width > pending_wallpaper.width or root_height > pending_wallpaper.height then
         -- The root window was resized while a wallpaper is pending
--- a/lib/gears/wallpaper.lua
+++ b/lib/gears/wallpaper.lua
@@ -40,6 +40,19 @@ end
 -- Information about a pending wallpaper change, see prepare_context()
 local pending_wallpaper = nil
 
+local pending_wallpaper_callback = nil
+pending_wallpaper_callback = function(calls)
+    calls = calls or 1
+    if pending_wallpaper then
+        local paper = pending_wallpaper
+        pending_wallpaper = nil
+        wallpaper.set(paper.surface)
+        paper.surface:finish()
+    elseif calls <= 3 then
+        timer.delayed_call(pending_wallpaper_callback, calls + 1)
+    end
+end
+
 local function get_screen(s)
     return s and screen[s]
 end
@@ -65,14 +78,7 @@ function wallpaper.prepare_context(s)
         target = source:create_similar(cairo.Content.COLOR, root_width, root_height)
 
         -- Set the wallpaper (delayed)
-        timer.delayed_call(function()
-            if pending_wallpaper then
-                local paper = pending_wallpaper
-                pending_wallpaper = nil
-                wallpaper.set(paper.surface)
-                paper.surface:finish()
-            end
-        end)
+        timer.delayed_call(pending_wallpaper_callback)
     elseif root_width > pending_wallpaper.width or root_height > pending_wallpaper.height then
         -- The root window was resized while a wallpaper is pending
         source = pending_wallpaper.surface


lib/gears/wallpaper.lua
```diff
@@ -66,10 +66,12 @@ function wallpaper.prepare_context(s)
...
         timer.delayed_call(function()
-            local paper = pending_wallpaper
-            pending_wallpaper = nil
-            wallpaper.set(paper.surface)
-            paper.surface:finish()
+            if pending_wallpaper then
+                local paper = pending_wallpaper
+                pending_wallpaper = nil
+                wallpaper.set(paper.surface)
+                paper.surface:finish()
```
Elv13: wont this need an "else" to start the delayed call again and a number of attempts before giving up? Right now I think this could cause the wallpaper to be lost.

DorianGray: It doesn't on my system at least, I had thought about doing that but it seems like it's not necessary? I can add it to be extra defensive, but I believe it might cause the callback to be called infinitely.

DorianGray: sorry that was stupid. I'm testing it more thoroughly now

DorianGray:
Updated to address concerns

DorianGray:
I can't reproduce the test failure locally 🤷‍♂

DorianGray:
My current config reproduces this consistently, 2 4k screens with a wallpaper being drawn maximized on both. I'll get back to you with a simple case to reproduce it.

The async code caused the error to be captured...it was getting swallowed before. The error was actually being reported in another async context because of lgi's async context error handling behavior. pending_wallpaper is seldom not nil.

psychon:
> The async code caused the error to be captured...it was getting swallowed before.

How is async code involved here?

> pending_wallpaper is seldom not nil.

Quick attempt at reproducing this (only run under Xephyr and with just a single wallpaper, but I duplicated the call to `gears.wallpaper.maximized` in the default config a couple of times):
```diff
diff --git a/lib/gears/wallpaper.lua b/lib/gears/wallpaper.lua
index 8b3f796a9..090925c68 100644
--- a/lib/gears/wallpaper.lua
+++ b/lib/gears/wallpaper.lua
@@ -40,6 +40,12 @@ end
 -- Information about a pending wallpaper change, see prepare_context()
 local pending_wallpaper = nil
 
+print("during startup, pending wallpaper is", nil)
+require("gears.timer").start_new(1, function()
+    print("timer: pending wallpaper is", pending_wallpaper)
+    return true
+end)
+
 local function get_screen(s)
     return s and screen[s]
 end
@@ -60,6 +66,7 @@ function wallpaper.prepare_context(s)
     local source, target, cr
 
     if not pending_wallpaper then
+        print("prepare_context: pending wallpaper was nil")
         -- Prepare a pending wallpaper
         source = surface(root.wallpaper())
         target = source:create_similar(cairo.Content.COLOR, root_width, root_height)
@@ -68,6 +75,7 @@ function wallpaper.prepare_context(s)
         timer.delayed_call(function()
             local paper = pending_wallpaper
             pending_wallpaper = nil
+            print("in delayed call, setting pending wallpaper to nil")
             wallpaper.set(paper.surface)
             paper.surface:finish()
         end)
@@ -97,6 +105,7 @@ function wallpaper.prepare_context(s)
         width = root_width,
         height = root_height
     }
+    print("pending wallpaper set to", pending_wallpaper)
 
     -- Only draw to the selected
```
```
during startup, pending wallpaper is	nil
2019-08-02 17:03:44 W: awesome: Could not load Rsvg: /usr/share/lua/5.3/lgi/namespace.lua:158: Typelib file for namespace 'Rsvg' (any version) not found
stack traceback: [SNIP]
prepare_context: pending wallpaper was nil
pending wallpaper set to	table: 0x555de29c6e60
pending wallpaper set to	table: 0x555de29c8080
pending wallpaper set to	table: 0x555de29c9580
pending wallpaper set to	table: 0x555de29c9170
in delayed call, setting pending wallpaper to nil
timer: pending wallpaper is	nil
timer: pending wallpaper is	nil
timer: pending wallpaper is	nil
```
So, at least I cannot reproduce this claim.

DorianGray:
I added async code in an unrelated part of my config which is what caused the is nil error to be spit out in the unrelated async function. I think this happens when an error occurs in an async context without a pcall. I haven't had time to build a concise test case yet, but it does require 2 screens as well. I'll get to it soon =)

psychon:
> I think this happens when an error occurs in an async context without a pcall. 

Well... according to my reading of the lgi source code...
```lua
   local coro = coroutine.create(function(...)
	 (function(...)
	     results = { n = select('#', ...), ... }
	     loop:quit()
	 end)(func(...))
   end)
```
You got a problem in this case. Nothing will quit this main loop if the coroutine runs into an unprotected error and just stops running. I guess the C code in `lgi/callable.c` will catch this and just print a warning basically saying "this is not allowed to happen":
```C
      if (callable->throws)
        res = lua_pcall (L, npos, LUA_MULTRET, 0);
      else if (lua_pcall (L, npos, LUA_MULTRET, 0) != 0)
        {
          callable_describe (L, callable, closure);
          g_warning ("Error raised while calling '%s': %s",
                     lua_tostring (L, -1), lua_tostring (L, -2));
          lua_pop (L, 2);
        }
```
Alternative, if this happens with a coroutine instead of a callback (I'm not quite sure which code path we end up in here), this code is executed:
```C
      else if (res == LUA_ERRRUN && !callable->throws)
	{
	  /* If closure is not allowed to return errors and coroutine
	     finished with error, rethrow the error in the context of
	     the original thread. */
	  lua_xmove (L, block->callback.L, 1);
	  lua_error (block->callback.L);
	}
```
So, Lua will do a `longjmp` to get out of this here and you end up with at least a memory leak and a possible corruption of something, because basically no C code is `longjmp`-safe.

The short version: From Awesome's POV, I would close such bugs telling you to report the problem to LGI instead. There, you end up with me again and I will tell you that you must not let errors bubble out of places where LGI cannot forward the error via a `GError` instance.

DorianGray:
There may be a related error in gears.timer.delayed_call then. I'll look into that as well.

lib/gears/wallpaper.lua
```diff
@@ -40,6 +40,19 @@ end
...
 
+local pending_wallpaper_callback = nil
+pending_wallpaper_callback = function(calls)
+    calls = calls or 1
+    if pending_wallpaper then
+        local paper = pending_wallpaper
+        pending_wallpaper = nil
+        wallpaper.set(paper.surface)
+        paper.surface:finish()
+    elseif calls <= 3 then
```
actionless: what is so magic about number `3`?

PR #2838, opened 2019-07-25
Use GLib idle sources for lots of house keeping

psychon:
This is loosely related to #2832. It does not help with that issue, but it reduces the amount of code that is "hidden" from the main loop. CC @DorianGray

These commits can easily break user configs. If you have a timer with timeout 0, it fires all the time. Timers have the GLib default priority (0) which is a higher priority than the default priority for idle callbacks (which is 200). Thus, a timer with timeout 0 will prevent all of this from running. I'd be tempted to tell anyone affected not to start a timer with a timeout of 0, even though I am guilty of writing such code myself once.
Also, any code that uses a timer with a timeout of 0 to get to a point where the main loop "did its thing" and all the lazily done stuff is, well, done, will now actively prevent such progress...

Long-term, I want to get rid of all the code in `a_glib_poll()` and `awesome_refresh()`, but this PR only handles the easy part. The hard parts (like the `refresh` signal) are still left.

Side note: At one point I had `test-gravity` break while writing this. I might have found the bug that prevents #2323 from being merged...

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -727,7 +727,7 @@ main(int argc, char **argv)
     sigaction(SIGCHLD, &sa, 0);
 
     /* We have no clue where the input focus is right now */
-    globalconf.focus.need_update = true;
+    client_focus_need_update();
 
     /* set the default preferred icon size */
     globalconf.preferred_icon_size = 0;
--- a/banning.c
+++ b/banning.c
@@ -23,34 +23,13 @@
 #include "globalconf.h"
 #include "objects/client.h"
 
-/** Reban windows following current selected tags.
- */
-void
-banning_need_update(void)
-{
-    /* We update the complete banning only once per main loop to avoid
-     * excessive updates...  */
-    globalconf.need_lazy_banning = true;
-
-    /* But if a client will be banned in our next update we unfocus it now. */
-    foreach(_c, globalconf.clients)
-    {
-        client_t *c = *_c;
-
-        if(!client_isvisible(c))
-            client_ban_unfocus(c);
-    }
-}
-
 /** Check all clients if they need to rebanned
  */
-void
-banning_refresh(void)
+static gboolean
+banning_refresh(gpointer unused)
 {
-    if (!globalconf.need_lazy_banning)
-        return;
-
-    globalconf.need_lazy_banning = false;
+    assert(globalconf.banning_update_id != 0);
+    globalconf.banning_update_id = 0;
 
     foreach(c, globalconf.clients)
         if(client_isvisible(*c))
@@ -61,6 +40,28 @@ banning_refresh(void)
     foreach(c, globalconf.clients)
         if(!client_isvisible(*c))
             client_ban(*c);
+
+    return G_SOURCE_REMOVE;
+}
+
+/** Reban windows following current selected tags.
+ */
+void
+banning_need_update(void)
+{
+    /* We update the complete banning only once per main loop to avoid
+     * excessive updates...  */
+    if (globalconf.banning_update_id == 0)
+        globalconf.banning_update_id = g_idle_add(banning_refresh, NULL);
+
+    /* But if a client will be banned in our next update we unfocus it now. */
+    foreach(_c, globalconf.clients)
+    {
+        client_t *c = *_c;
+
+        if(!client_isvisible(c))
+            client_ban_unfocus(c);
+    }
 }
 
 // vim: filetype=c:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/banning.h
+++ b/banning.h
@@ -23,7 +23,6 @@
 #define AWESOME_BANNING_H
 
 void banning_need_update(void);
-void banning_refresh(void);
 
 #endif
 // vim: filetype=c:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/event.c
+++ b/event.c
@@ -686,7 +686,7 @@ event_handle_enternotify(xcb_enter_notify_event_t *ev)
         /* When there are multiple X screens with awesome running separate
          * instances, reset focus.
          */
-        globalconf.focus.need_update = true;
+        client_focus_need_update();
     }
 }
 
--- a/event.h
+++ b/event.h
@@ -31,9 +31,6 @@
 /* luaa.c */
 void luaA_emit_refresh(void);
 
-/* objects/drawin.c */
-void drawin_refresh(void);
-
 /* objects/client.c */
 void client_refresh(void);
 void client_focus_refresh(void);
@@ -43,7 +40,6 @@ static inline int
 awesome_refresh(void)
 {
     luaA_emit_refresh();
-    drawin_refresh();
     client_refresh();
     banning_refresh();
     stack_refresh();
--- a/event.h
+++ b/event.h
@@ -34,14 +34,12 @@ void luaA_emit_refresh(void);
 /* objects/client.c */
 void client_refresh(void);
 void client_focus_refresh(void);
-void client_destroy_later(void);
 
 static inline int
 awesome_refresh(void)
 {
     luaA_emit_refresh();
     client_refresh();
-    client_destroy_later();
     return xcb_flush(globalconf.connection);
 }
 
--- a/event.h
+++ b/event.h
@@ -41,7 +41,6 @@ awesome_refresh(void)
 {
     luaA_emit_refresh();
     client_refresh();
-    banning_refresh();
     stack_refresh();
     client_destroy_later();
     return xcb_flush(globalconf.connection);
--- a/event.h
+++ b/event.h
@@ -41,7 +41,6 @@ awesome_refresh(void)
 {
     luaA_emit_refresh();
     client_refresh();
-    stack_refresh();
     client_destroy_later();
     return xcb_flush(globalconf.connection);
 }
--- a/globalconf.h
+++ b/globalconf.h
@@ -151,7 +151,7 @@ typedef struct
         /** Focused client */
         client_t *client;
         /** Is there a focus change pending? */
-        bool need_update;
+        guint source_id;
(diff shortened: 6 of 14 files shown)


DorianGray:
Is it possible to just convert timers with timeout=0 to delayed call to prevent breakage?

psychon:
It's also possible to just give timers with timeout=0 a different priority than others. However, I'd like to avoid doing such magic.

Instead, one could raise the priority of all the idle sources that I am adding here to something "above normal", e.g. -10. That might be easier. But I'm not sure if that is necessary for every one of them or not...

psychon:
Oh and: This kind of breaks the `selection()` global function (which we should perhaps formally deprecate since it has better, async replacements these days). The `selection()` global just calls `xcb_wait_for_event()` in a loop until the "right" event arrives. The events it sees are handled, but no main loop runs, so nothing else is handled. So, everything that I am moving to idle sources will no longer run while `selection()` waits. I am not sure if any of this makes a difference, since `selection()` is supposed not to block for extended periods of time.

Elv13:
> This kind of breaks the selection() global function

Can we add a magic constructor flag to make the new one blocking (not by default, and is this even needed?) and we could just drop the old one and add a Lua implementation to the `_legacy.lua` file one of my PR adds somewhere (graveyard of all broken APIs needed for compat)? (I am at the Airport on a Mac right now, I can't check).

psychon:
If you want to reimplement `selection()` in Lua to "pretend to be blocking", then @DorianGray came up with a semi-magic way on how to do that that I kind of knew about, but also never considered. Just do what `lgi.Async.call` does and start your own main loop! (Following code is completely untested)
```lua
function do_something_and_pretend_not_to_be_blocking()
  local loop = GLib.MainLoop()
  -- Setup the operation, e.g. request the current selection through the new API.
  local s = selection.getter{ selection = "PRIMARY", target="UTF8_STRING" }
  local data = {}
  s:connect_signal("data", function(_, d) table.insert(data, d) end)
  s:connect_signal("data_end", function()
    -- This is part of the magic: When we are done with blocking, quit the main loop we created
    loop:quit()
  end)
  -- Then start the main loop and wait for it to be done
  loop:run()
  return table.concat(data)
end
```
If something goes wrong during the transfer and e.g. the other end doesn't answer, then the above will never return. But the current `selection()` global does not handle that either.

The question is: Do we want to allow Lua to start another main loop instance? GLib (and GTK and....) is totally fine with this and it is how the API is supposed to work. For AwesomeWM, we will keep running into problems with this for the foreseeable future (and hence: @Elv13 Do we want to allow this or do we want to "not going to fix that" the resulting bug reports?).

Random other example of possible bugs (imagine this being a bit less obvious and hidden via some layers of abstractions and code doing... things): If you start a main loop in a signal handler, then you just blocked calls to other functions connected to the same signal that might possible run later. For example, add the following to the beginning of your `rc.lua` (before other code connects to `manage` so that this runs first) and you just blocked delivery of the `manage` signal without locking up all of the WM:
```lua
client.connect_signal("manage", function()
  local loop = require("lgi").GLib.MainLoop()
  -- If we want to be semi-nice, we could just delay this signal for 60 seconds:
  -- gears.timer.start_new(60, function() loop:quit() end)
  loop:run()
end)
```
(No idea how this could be fixed sanely. An insane fix would be to move signal emission to idle sources. We could then add a separate source for every connected callback function and the recursive main loop would still invoke the other callbacks.)

DorianGray:
if each individual handler function were scheduled on a loop as a separate coroutine they could all run concurrently. That would prevent one signal handler from blocking another, since afaict that is never the desired behavior.

Elv13:
> and hence: @Elv13 Do we want to allow this or do we want to "not going to fix that" the resulting bug reports?).

@DorianGray  Isn't there ways to use the async APIs without messing with this? What are the limitations of using them without messing with nested event loops?

>  An insane fix would be to move signal emission to idle sources. We could then add a separate source for every connected callback function and the recursive main loop would still invoke the other callbacks.)

I had a similar idea when first reading the original report. But mine was slightly less invasive. Adding a `:connect_signal_async()` method which runs the callback from an idle source as a coroutine might be an interesting alternative. Making all signal handlers running from their "own" main loop iteration would expose a flurry of bugs and slightly reduce performance. But allowing explicit `:connect_signal_async()` is at least explicit enough and new enough so wont regress existing code and users will have lower expectation.

Timers can be implemented from the "real" mainloop and just `yield()` the coroutine and magically wake up when the timer is over. No need for multiple even loops. We could inject some magical token in the coroutine environment using the Lua debug API then add methods such as `:wait()` to `gears,timer`. Those method would error if they are not called from a compliant coroutine (with the said token). Same for my object oriented `spawn` module I have in a branch somewhere. It would just wait on client factories. Then implement `:fetch_line()` or `.result` on an `awful.spawn.easy_async` equivalent, etc, etc. After a while it would be a nice little async API integrated well into existing APIs.

An alternative is a whole new module with all the async stuff. This would have the downside of duplicated APIs. On the plus side it would solve the classic CS "color label" issue. This is a problem in cooperative multithreading where each method has to externally define if it is safe or not to call from a coroutine. Integrating blocking calls that only works in coroutine into existing APIs will make it confusing to the people who don't read the doc properly. With a new module, you can assume everything is safe as long as it's namespaced with that module.

Overall, making AwesomeWM APIs "like Golang" would allow much higher quality code (less callbacks), better performance and we could "solve" `io.popen`/`os.execute` by monkey-patching them to be async. I am already making incremental changes to get rid of code executed directly in the global context. Once all code is executed from "top level" signal handlers, it becomes easier to make some handler async.

DorianGray:
The reason for using call right now is to be able to return values across the c-boundary. It quits the main loop once all work is done, sets a variable on the lua side(results) then aggregates and returns them. Without the nested main loop, I can't find a way to return values from any async work(no way to signal that a value is ready without yielding across the c-boundary).

I'm trying a few different approaches right now...

I'm trying to isolate lua coroutines on each side of the c-boundary so they don't yield to one another. I'd then be able to use a single glib mainloop and have each "type" of coroutine(one type that uses glib's async features and the other where they are unavailable) be able to execute side by side but be guaranteed not to yield to one another. There would be a single controlled place where a loop would watch a list of work in the glib async side and check to see if a function had completed, then yield the result to the appropriate place. It's basically what Gio.Async.call does right now, but minimizes the number of mainloops that have to be instantiated.

I'm also investigating whether or not a master coroutine scheduler could be built on the awesome side to schedule all coroutines in glib and handle the above problems in a more all encompassing way... I haven't fleshed out this idea yet, but I'm hoping to this week. It would probably involve overriding the way that coroutines work at a low level, like other libev bindings do.

The last way is the way I've unfortunately spent so much time on already, which is using Gio.Async.start as is to manage work. I keep running into the c-boundary brick wall here and I've tried many different methods of gathering results but either awesome core dumps or lua throws the c-boundary error at me.

I don't have a working solution outside of using Async.call yet.

I am working on process and stream management high-level apis to prove that one of the above solutions works that are very different from the basic lua ones and much more featureful. I have starting processes, capturing stream output, and reading files built and the api is pretty clean so far. For instance, 
```
local output = process.run('cat /proc/net/wireless').stdout() -- short form which expects a process to run and end shortly
local p = process('cat /proc/net/wireless') -- the process constructor also takes an args table which allows setting all the current process options and more
p:start()
local stderr = p.stderr() -- these two are equivalent, just nice sane defaults
local stderr = p.stderr:read_all()
for line in p.stdout:read_lines() do -- each line is yielded separately and does not block the event loop, allowing for potential back and forth communication once I get writing to stdin working. I plan on integrating my wifi widget to talk directly to wpa_supplicant this way
  print(line)
end
```
```
local file_content = file('/proc/net/wireless'):read_all() -- this is just passthrough to the file's read stream. read_lines, etc are there too.
```

I was thinking that the existing awesome apis could be patched to just use these objects. I always get so frustrated with the amount of boilerplate I need to write to manage processes in awesome so I thought that changing the apis with a separate library would make more sense.

DorianGray:
The code is at https://github.com/DorianGray/awesome-config/tree/master/awful/io right now

psychon:
> I've tried many different methods of gathering results but either awesome core dumps or lua throws the c-boundary error at me.

Each way to cause core dumps is interesting me. They sound like bugs to investigate (even though I do not have much time to work on them currently).

In theory Lua (or just some Lua versions?) allow to yield across the c-boundary via a special call function and a "continuation" callback. I'm not sure I want to go there, but at least in theory it should thus be possible to make things yieldable.

> The code is at https://github.com/DorianGray/awesome-config/tree/master/awful/io right now

These two functions confuse me. Either they do not work, or I do not understand how they work (`start` instead of `call`): https://github.com/DorianGray/awesome-config/blob/95824a287f5368cd1ec5da01fae97c87c4ab654c/awful/io/stream/init.lua#L70-L76

This uses `call` to start a new main loop. Seems fine: https://github.com/DorianGray/awesome-config/blob/95824a287f5368cd1ec5da01fae97c87c4ab654c/awful/io/stream/init.lua#L78-L87

I'm not sure this one needs a coroutine: https://github.com/DorianGray/awesome-config/blob/95824a287f5368cd1ec5da01fae97c87c4ab654c/awful/io/stream/init.lua#L89-L95
Something like this should be (mostly) equivalent?
```lua
function stream:read_chunks(chunk_size)
  return function()
    if not self.closed do
      return self:read(chunk_size)
    end
  end
end
```

DorianGray:
I'm really glad you picked up on that. The reason those work is that they are not using coroutines, they run a lua function in, triggered by glib, and return a value. The coroutine starter function simply returns a value across the c boundary instead of yielding.

The functions that do more complicated work, unfortunately, require call because they have to yield values. To get the right results back from yield we need a mainloop to determine when that entire stack is through executing and return the results.

Yes, that iterator doesn't require a coroutine with the current async.call based backend. I use this to test for potential of yielding across the c boundary.

PR #2847, opened 2019-08-07
layoutlist: Use weak tables to store the cache.

### LINKED ISSUE #2846: Layoutlist memory leak

awesome v4.3 (Too long)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

Add layoutlist popup to your config (straight from docs) :
`awful.popup {`
`    widget = awful.widget.layoutlist {`
`         screen      = 1,`
`         base_layout = wibox.layout.flex.vertical`
`    },`
`    maximum_height = #awful.layout.layouts * 24,`
`    minimum_height = #awful.layout.layouts * 24,`
`    placement      = awful.placement.centered,`
` }`

Then keep switching tags, rapidly if you want to see results.

**Actual result:**

Awesomes memory usage keeps going up. With default config ~2mb per switch, with more specified layoutlist i experienced 5-10mb per tag switch.

**Expected result:**

Memory should not be eaten like that.

Elv13:
Mitigate #2846

### DIFF
--- a/lib/awful/widget/layoutlist.lua
+++ b/lib/awful/widget/layoutlist.lua
@@ -327,7 +327,7 @@ function layoutlist:set_widget_template(widget_template)
     self._private.widget_template = widget_template
 
     -- Remove the existing instances
-    self._private.data = {}
+    self._private.data = setmetatable({}, { __mode = 'k' })
 
     -- Prevent a race condition when the constructor loop to initialize the
     -- arguments.
@@ -403,7 +403,7 @@ local function new(_, args)
     ret._private.style   = args.style or {}
     ret._private.buttons = args.buttons
     ret._private.source  = args.source
-    ret._private.data = {}
+    ret._private.data = setmetatable({}, { __mode = 'k' })
 
     reload_cache(ret)
 
--- a/lib/awful/widget/layoutlist.lua
+++ b/lib/awful/widget/layoutlist.lua
@@ -118,8 +118,12 @@ function module.source.default_layouts()
     return alayout.layouts
 end
 
+-- Keep the object for each layout. This avoid creating little new tables in
+-- each update, which is painful for the GC.
+local l_cache = setmetatable({}, {__mode = "k"})
+
 local function reload_cache(self)
-    self._private.cache = {}
+    self._private.cache = setmetatable({}, {__mode = "v"})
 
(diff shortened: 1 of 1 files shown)


Elv13:
I added a second commit to make the GC life easier. It has fewer strong references and creates less temporary tables that get dismissed when the content change.

Another optimization would be to detect when the selected tag uses the same layouts as the previous one and has the same selected layout. However if I do this optimization now, it will shadow the remaining leak.

Elv13:
Amended the commit to handle another missed caching opportunity.

Elv13:
>  This table is marked as "feel free to forget entries"

Unless my understanding is wrong, `__mode = "v"` is not "feel free to forget entries". It is "feel free to remove entries with no other strong references". There is an explicit strong reference a line below and the `table.insert` is below *that*. So no, it wont just free them randomly. They actually wont ever get freed unless someone uses dynamic layout. Since the layoutlist doesn't allow to set custom buttons, that cache table can be shared by all instances.

What this pull request does is reduce the number of widgets being created. This is what was leaking, your patch only fixes the consequence by letting the `data` table free the orphan widgets that were created rather than preventing orphan widgets to be created like the commit I added. Both together fix the leak.

psychon:
Oh, I mis-read the code. `l_cache` has weak keys, meaning that when the *layout* becomes unreachable, its values can be collected. Somehow I was mixing this up with the other table and thought that that it had weak values. Sorry!

So, something references the layout. Through `l_cache`, this keeps "the big table" alive (Hm... is it a problem that the value references the layout? I think not, but weak tables can be confusing...). This table is then added to `self._private.cache`, which has weak values.

So, if the layout is no longer strongly reachable, everything else can be GC'd.

Elv13:
>  (Hm... is it a problem that the value references the layout? I think not, but weak tables can be confusing...)

last time I tested this hypothesis, there was some issues, but overall Lua GC handles this across all supported versions. In *this* case it is a non-issue, because only people using dynamic layouts (and that's probably < 5 people) will ever have layout objects. The weak cache is for future-proofness.

> So, if the layout is no longer strongly reachable, everything else can be GC'd.

Yes, that's the intent of this commit and it's weird combination of weak tables.

PR #2853, opened 2019-08-18
Add the notification rules.

Elv13:
This is the final "main" PR in the notification rewrite. It replaces the old "preset" concept with a full `awful.rules` like system to control the notification. It can be used to set the color, ignore them, set custom widget templates, etc. The basic usage is:

```lua
   ruled.notification.connect_signal("request::rules", function()
       -- Add a red background for urgent notifications.
       ruled.notification.append_rule {
           rule       = { urgency = "critical" },
           properties = { bg = "#ff0000", fg = "#ffffff", timeout = 0 }
       }

       -- Or green background for normal ones.
       ruled.notification.append_rule {
           rule       = { urgency = "normal" },
           properties = { bg      = "#00ff00", fg = "#000000"}
       }
   end)
```

Please note that this is currently in a new module called "ruled". I prefer this over adding `naughty.rules` since I also plan to expand the rules to tags, screens and the clipboard. `awful.rules` itself would be moved to this new module. If someone has a better name, please come forward.

The content of this PR is less or more final. ~~The tests are lacking and a bit broken~~. The NEWS.md also needs to updated to reflect the entire rewrite effort. I don't plan to merge this in the immediate future. It will co-exist for a while with some other PRs to introduce the other components of the new `ruled` module.

### DIFF
--- a/lib/ruled/notification.lua
+++ b/lib/ruled/notification.lua
@@ -27,7 +27,6 @@ local capi = {screen = screen, client = client, awesome = awesome}
 local matcher = require("gears.matcher")
 local gtable  = require("gears.table")
 local gobject = require("gears.object")
-local naughty = require("naughty")
 
 --- The notification is attached to the focused client.
 --
@@ -206,7 +205,7 @@ capi.awesome.connect_signal("startup", function()
         module.emit_signal("request::rules")
 
         -- This will disable the legacy preset support.
-        naughty.connect_signal("request::preset", function(n)
+        require("naughty").connect_signal("request::preset", function(n)
             module.apply(n)
         end)
     end
--- a/lib/ruled/notification.lua
+++ b/lib/ruled/notification.lua
@@ -71,6 +71,10 @@ local gobject = require("gears.object")
 -- @clientruleproperty implicit_timeout
 -- @param number
 
+--- Do not let this notification timeout, even if it asks for it.
+-- @clientruleproperty never_timeout
+-- @param boolean
+
 local nrules = matcher()
 
 local function client_match_common(n, prop, value)
@@ -116,7 +120,20 @@ end)
 
 nrules:add_property_setter("implicit_timeout", function(n, value)
     -- Check if there is an explicit timeout.
-    if not n._private.timeout then
+    if (not n._private.timeout) and (not n._private.never_timeout) then
+        n.timeout = value
+    end
+end)
+
+nrules:add_property_setter("never_timeout", function(n, value)
+    if value then
+        n.timeout = 0
+    end
+end)
+
+nrules:add_property_setter("timeout", function(n, value)
+    -- `never_timeout` has an higher priority than `timeout`.
+    if not n._private.never_timeout then
         n.timeout = value
     end
 end)
--- a/lib/ruled/notification.lua
+++ b/lib/ruled/notification.lua
@@ -213,6 +213,15 @@ function module.add_rule_source(name, cb, ...)
     return nrules:add_matching_function(name, cb, ...)
 end
 
+-- Allow to clear the rules for testing purpose only.
+-- Because multiple modules might set their rules, it is a bad idea to let
+-- them be purged under their feet.
+function module._clear()
+    for k in pairs(nrules._matching_rules["ruled.notifications"]) do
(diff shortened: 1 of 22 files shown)


Elv13:
I added more tests. As for the "build failure", it is because the build directory isn't clean when iterating each commit. Because `ruled/notification.lua` is modified once, the "old" copy is left in the build tree after the builder finish building the last commit and start iterating each commit from the first. One way to do this is to add another commit to use the `@rulelib` and other new tags in the last commit. However that's a hack a "someone" will complain how ugly that it, so I leave it as-is for now.

@actionless Can you try this new API and see if it fits your needs? You wondered how easy some tasks were going to be. Time to try it!

SethBarberee:
@Elv13 I can give this a run for a bit on my machine and give feedback. Any specific test cases you want tested? I can give some feedback within the week

Elv13:
Sure, go hard at it. There is no specific things to try. This is probably
the most useful for filtering and to replace the old preset API. And of
course for custom template.

SethBarberee:
```
    ruled.notification.append_rule {
        rule       = { app_name = 'Spotify' },
        properties = { 
            bg      = '#000000', 
            fg = '#00ff00',
            append_actions = {
               naughty.action {
                   name = 'Snooze (5m)',
               },
               naughty.action {
                   name = 'Snooze (15m)',
               },
               naughty.action {
                   name = 'Snooze (1h)',
               },
            },
        }
    }
```
Good news: 
Was able to add actions to Spotify... you are a genius, Elv. Happy to see that my reddit post in regards to notifications special to spotify is possible now :smile: 

Bad news (only in `naughty.layout.box` formatted notifications):
While the said notification is green, actions are default foreground.
**This doesn't occur in the old style as actions are changed to said green**

Elv13:
> While the said notification is green, actions are default foreground.

Yeah, `beautiful` has precedence unless you set a specific `fg` for the actions. This might not have been the best design decision. I will see what I can do about it. You *can* make rules to color the actions, but they should probably follow the whole notification fg unless explicitly overridden rather than use their own fallback when a notification `fg` is set, but not an action `fg`.

actionless:
could you please rebase it so i could merge it to my local branch to test

Elv13:
> could you please rebase it so i could merge it to my local branch to test

Done. I pressed the "resolve conflicts" button in the GUI. It seems to have worked.

SethBarberee:
With #2895 merged, is this still WIP? Wanted to make sure before I deep dive and do more testing/review of it. Looking forward to this PR.

Elv13:
>  is this still WIP?

Yes, there is a bunch of problems:

 * The multi-screen support for the `box` notification isn't behaving properly
 * Some signal signature are incompatible with my API standardization work I just merged (facepalm)
 * The icon update for Spottily notification is still broken.
 * It doesn't have a permission per-client block the notification using the new permission framework

Beside, it is working and not much will change beside the signal format.

Elv13:
I consider this PR to no longer be WIP/RFC. While I didn't fix the spotify bug because I don't have a setup to reproduce it, I think otherwise, it is generally useful and stable enough to merge and enable by default. At this point, most "major" configs as seen on `/r/unixporn` and our screenshot page migrated to the new notification API. I guess it works for them, so lets make this a bit easier for newcomers to use. I started this notification project 3 years ago (almost, technically in June). It has been a long journey to get this merged. Hundreds of commits are related to the groundwork leading to this PR getting merged...

lib/ruled/notification.lua
```diff
@@ -0,0 +1,242 @@
...
+
+--- Append some actions to a notification.
+--
+-- Using `actions` directly is destructive since it will override existing
+-- actions.
+--
+-- @clientruleproperty append_actions
+-- @param table
+
+--- Set a fallback timeout the notification has no explicit timeout.
```
SethBarberee:
```suggestion
--- Set a fallback timeout when the notification doesn't have an explicit timeout.
```

Aire-One:
I never have desktop notification so I never took the time to implement your new API in my setup but I silently followed each step of the development. It was a long journey.

This is great to see it makes its way to default `rc.lua`. ea79035 is the final step of this long running project. 

Congrats! 🎊

PR #2854, opened 2019-08-19
Add some navigation helpers to the doc (in the form of UML tables)

Elv13:
I had this for a while and there's no reason not to upstream it. The general problem it tries to solve is that it requires a lot of reading to find how to acquire an object when you have another object from another class. This PR adds a little chart to easily navigate without having to read the doc for half an hour.

Screenshot_20190819_005329

Note that the composition vs. aggregation vs. association is partially implemented in the code, but not really working. I think it is good enough for now. When there will be more of them, it might be worth fixing.

@blueyed What do you think?

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -82,6 +82,8 @@
  * To get all the clients for a screen use either `screen.clients` or
  * `screen.tiled_clients`.
  *
+ * @DOC_uml_nav_tables_client_EXAMPLE@
+ *
  * @author Julien Danjou &lt;julien@danjou.info&gt;
  * @copyright 2008-2009 Julien Danjou
  * @coreclassmod client
@@ -129,7 +131,7 @@
  * it is only useful when setting or getting these properties require code to
  * executed.
  *
- * @table awful.object
+ * @table awful.client.object
  */
 
 /** When a client gains focus.
--- a/objects/screen.c
+++ b/objects/screen.c
@@ -40,6 +40,8 @@
  * Furthermore to the classes described here, one can also use signals as
  * described in @{signals}.
  *
+ * @DOC_uml_nav_tables_screen_EXAMPLE@
+ *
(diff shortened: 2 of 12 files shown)


blueyed:
Hmm.. it adds a lot of boilerplate/code/lines.
We should rather ensure that you can find it easily from the current docs (by searching the page)?

Elv13:
> Hmm.. it adds a lot of boilerplate/code/lines.

The template that generated it is 180 lines of code (minus empty lines), everything else is the auto generated HTML or the "json" like "code" that stores the information. These files have zero maintenance cost. 

The information cannot be acquired by searching the page because the information is from other pages (how to get clients from a tag is in the tag page while you are looking in the client page). So `ctrl+f` doesn't gives you the important half of the information provided by this table. I also don't think a few lines of auto generated HTML is a problem at all. Any SVG is has thousands of them. If by searching the page you mean a search engine, then I don't think it is relevant to this discussion because we don't have one and using DuckDuckGo don't gives useful results because searching an API is different from searching the web.

The API doc being hard to navigate is a common complain on reddit (from r/unixporn, people populating r/awesomewm  are users who overcame the doc limitations, so they have the survivor bias). How to get objects from "scratch" isn't covered unless you read the entire doc and piece back together the bigger picture. This PR intends on mitigating this problem.

Ideally, every API property/method/class should have examples and images. They help people grasp the concepts better. The learning curve ls quicker and less steep. Those images and example make the build slower and add tens of thousand of lines to the project. Those cost are hidden from the users and only relevant to the ~10 people who compile AwesomeWM more than once a year. I think the tradeoff is vastly more beneficial to the project.

Elv13:
Assuming the `+1` for my comment means you approve the reasoning behind this PR, I will merge it tomorrow. If you still disagree, please say so.

PR #2855, opened 2019-08-19
doc: Add some images to represent the various area of a screen.

Elv13:
Another low hanging fruit from one of my branch. This template is half-done. It is intended to help show how docking and struts work. It was part of a larger body of work to add images to `awful.wibar` to be closer to how well `awful.tooltip` is documented. I have the first half of the template working fine and some examples use it. Better merge this now than wait another year to so I find time to write the other half.

Screenshot_20190819_014005

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -174,6 +174,30 @@ function screen.focus_relative(offset)
                                    screen.focused().index + offset))
 end
 
+--- The area where clients can be tiled.
+--
+-- This property holds the (read only) area where clients can be tiled. Use
+-- the `padding` property, `wibox.struts` and `client.struts` to modify this
+-- area.
+--
+-- @DOC_screen_tiling_area_EXAMPLE@
+--
+-- @property tiling_area
+-- @tparam table tiling_area
+-- @tparam number tiling_area.x
+-- @tparam number tiling_area.y
+-- @tparam number tiling_area.width
+-- @tparam number tiling_area.height
+-- @see padding
+-- @see get_bounding_geometry
+
+function screen.object.get_tiling_area(s)
+    return s:get_bounding_geometry {
+        honor_padding  = true,
+        honor_workarea = true,
+    }
+end
+
 --- Get or set the screen padding.
 --
 -- @deprecated awful.screen.padding
@@ -198,6 +222,8 @@ end
 -- **Signal:**
 --
 -- * *property::padding*
+--
+-- @DOC_screen_padding_EXAMPLE@
 --
 -- @property padding
 -- @param table
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -746,7 +746,7 @@ end
 
 --- The tag client layout.
 --
--- This property hold the layout. A layout can be either stateless or stateful.
+-- This property holds the layout. A layout can be either stateless or stateful.
 -- Stateless layouts are used by default by Awesome. They tile clients without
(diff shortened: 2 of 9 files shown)


Elv13:
last call for comments

Aire-One:
Thank you for your work on the documentation!

I would change the `screen.padding` image tho. It would be better if we can highlight the padding itself and not the tiling_area. My knowledge of cairo are pretty bad, so here is what I drafted:

screen_padding

```
diff --git a/tests/examples/screen/padding.lua b/tests/examples/screen/padding.lua
index dba00d73..79b9e049 100644
--- a/tests/examples/screen/padding.lua
+++ b/tests/examples/screen/padding.lua
@@ -34,5 +34,5 @@ end --DOC_HIDE
 return { --DOC_HIDE
     factor                = 2    , --DOC_HIDE
     show_boxes            = false, --DOC_HIDE
-    highlight_tiling_area = true , --DOC_HIDE
+    highlight_padding_area = true , --DOC_HIDE
 } --DOC_HIDE
diff --git a/tests/examples/screen/template.lua b/tests/examples/screen/template.lua
index 75d4f701..28c602bc 100644
--- a/tests/examples/screen/template.lua
+++ b/tests/examples/screen/template.lua
@@ -79,6 +79,7 @@ local colors = {
     geometry    = "#000000",
     workarea    = "#0000ff",
     tiling_area = "#ff0000",
+    padding_area= "#ff0000",
 }
 
 local function draw_area(_, rect, name, offset, highlight)
@@ -96,6 +97,15 @@ local function draw_area(_, rect, name, offset, highlight)
     cr:stroke()
 end
 
+local function draw_bounding_area(_, rect, hole, name, offset)
+    draw_area(_, rect, name, offset, true)
+
+    local x, y = hole.x*factor+offset, hole.y*factor+offset
+    cr:rectangle(x, y, hole.width*factor, hole.height*factor)
+    cr:set_source_rgb(1, 1, 1)
+    cr:fill()
+end
+
 local function compute_ruler(_, rect, name)
     table.insert(hrulers, {
         label = name, x = rect.x, width = rect.width
@@ -314,6 +324,9 @@ for k=1, screen.count() do
     draw_area(s, s.workarea, "workarea", (k-1)*10, args.highlight_workarea)
 
     -- The padding.
+    if args.highlight_padding_area then
+        draw_bounding_area(s, s.workarea, s.tiling_area, "padding_area", (k-1)*10)
+    end
     draw_area(s, s.tiling_area, "tiling_area", (k-1)*10, args.highlight_tiling_area)
 
     -- Draw the ruler.
```
I know using a white rectangle is a shitty solution but that's the best workaroud I find to make the inher area "transparent" back...

I just noticed (while writing this) that the red border I used for the padding area and the blue border of the workarea melted together to make purple...

---

Also, a little off topic here, but I think we should have an example of what the `screen.geometry` table looks like.

Elv13:
Thanks **a lot** for caring about my little "let's make the doc better" project. Honestly, most people don't see the point. I modified your patch a little bit to call `:set_operator`. This allows the inner "hole" to be transparent instead of white. As for the purple border, this isn't really a concern. It wont confuse anybody.

> Also, a little off topic here, but I think we should have an example of what the screen.geometry table looks like.

True. I will wait until the "API standardization" PR gets some reviews and is merged (along with the new follow-up PRs currently blocked by it). Right now `c:geometry().x` vs. `s.geometry.x` is very confusing. Once all objects will have a standardized getter/setter convention, it will be easier to document "at scale".

PR #2865, opened 2019-09-08
Disable colors in luacheck to make the CI output more readable.

Elv13:
(made in GitHub UI, untested)

The Travis "raw log" is hard to read with all the ANSI escape sequences.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -49,4 +49,7 @@ globals = {
 -- Enable cache (uses .luacheckcache relative to this rc file).
 cache = true
 
+-- Do not enable colors to make the Travis CI output more readable.
+color = false
+
 -- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80


blueyed:
> The Travis "raw log" is hard to read with all the ANSI escape sequences.

But it looks good in the "normal view": https://travis-ci.com/awesomeWM/awesome/jobs/231922979#L3100

Travis used to have a "?deansi=true" mode, but this appears to not work anymore?

I'm +0 on this - there are a lot of other ANSI escape sequences anyway.

Elv13:
> Travis used to have a "?deansi=true" mode, but this appears to not work anymore?

This doesn't seem to do anything for me (as you said). Right now it prevents me from ctrl+f into the logs and since I often have tons of commits into the PRs, I waste time on this. The JavaScript-ified logs are utterly useless. The important parts are the ones hidden the deeper. If it's supposed to help make the logs more readable, then it's a failure. I would rather have no colors are a searchable log (without curl-ing it in a terminal). So I am merging this.

PR #2870, opened 2019-09-13
naughty: fix extra newline with only title/message

blueyed:
With only title or message it should not have an extra newline.

Fixes: https://github.com/awesomeWM/awesome/commit/423aeebe8#commitcomment-35062951

This has been regressed again after I've fixed it in 8072d717c already - so it would be good to have a test for it.

TODO:

- [x] test

### DIFF
--- a/lib/naughty/layout/legacy.lua
+++ b/lib/naughty/layout/legacy.lua
@@ -237,23 +237,39 @@ local function set_escaped_text(self)
     if not self.box then return end
 
     local text = self.message or ""
-    local title = self.title and self.title .. "\n" or ""
-
+    local title = self.title or ""
     local textbox = self.textbox
 
     local function set_markup(pattern, replacements)
-        return textbox:set_markup_silently(string.format('<b>%s</b>%s', title, text:gsub(pattern, replacements)))
+        local parts = {}
+        if title ~= "" then
+            table.insert(parts, "<b>" .. title .. "</b>")
+        end
+        if text ~= "" then
+            local markup = text:gsub(pattern, replacements)
+            if markup ~= "" then
+                table.insert(parts, markup)
+            end
+        end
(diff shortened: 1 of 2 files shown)


blueyed:
Now with improved fix and test.

lib/naughty/layout/legacy.lua
```diff
@@ -237,23 +237,39 @@ local function set_escaped_text(self)
...
+        end
+        textbox:set_text(table.concat(parts, "\n"))
     end
 
     -- Since the title cannot contain markup, it must be escaped first so that
     -- it is not interpreted by Pango later.
     title = title:gsub(escape_pattern, escape_subs)
     -- Try to set the text while only interpreting <br>.
-    if not set_markup("<br.->", "\n") then
+    if not set_markup("<br.*>", "\n") then
```
blueyed: Just for clarity.

lib/naughty/layout/legacy.lua
```diff
@@ -237,23 +237,39 @@ local function set_escaped_text(self)
...
         -- That failed, escape everything which might cause an error from pango
         if not set_markup(escape_pattern, escape_subs) then
             -- Ok, just ignore all pango markup. If this fails, we got some invalid utf8
```
blueyed: NOTE: "set_text" (below) does not fail with invalid utf8.
This could be achieved by checking if the text was set according to the expected (input) one in https://github.com/blueyed/awesome/blob/813ed34a8502efec85a841ba2e5037b82238e156/lib/wibox/widget/textbox.lua#L196, and then throwing an error from there - but I do not feel like covering that now additionally..

blueyed: Hmm.. the diff being covered now surprises me - but ok.. ;)

PR #2872, opened 2019-09-13
tests/run.sh: output "awesome --version"

blueyed:
Might be good to have a short version of it (single line), but extra
lines are indented, so this seems to be OK.

This helps / shows when the built version is not up-to-date (e.g. after
git-bisect), and serves as a basic check that it can be run in the first
place.


blueyed:
> Fine with me, but I would totally expect me to miss this output in the situation you describe (outdated build after bisect)

Yeah - you would not read it always, but then have it later in the terminal history at least.

Also useful in CI logs I assume when investigating in general.

PR #2877, opened 2019-09-13
tests: runner.run_steps: add fname/lnum for failed steps

blueyed:


tests/_runner.lua
```diff
@@ -136,18 +138,26 @@ function runner.run_steps(steps, options)
...
-                runner.done("timeout waiting for signal in step "
-                            ..step_as_string..".")
-                t:stop()
+                -- No result yet, run this step again.
+                wait = wait-1
+                if wait > 0 then
+                    t.timeout = 0.1
+                    t:again()
+                else
+                    require("gears.debug").dump(debug.getinfo(step_func))
```
psychon: Is this addition a left-over from development?

tests/_runner.lua
```diff
@@ -136,18 +138,26 @@ function runner.run_steps(steps, options)
...
         else
-            -- No result yet, run this step again.
-            wait = wait-1
-            if wait > 0 then
-                t.timeout = 0.1
-                t:again()
+            -- Append filename/lnum of failed step function.
+            local step_info = debug.getinfo(step_func)
+            local step_loc = string.format("%s:%d", step_info["short_src"], step_info["linedefined"])
+            step_as_string = step_as_string .. " ("..step_loc..")"
```
psychon: I know that the `debug` library is supposedly slow, but I also doubt that this is relevant here, because this code is not executed that often, so:

Does it make sense to move this up to the code that sets up `step_as_string` so that all of it is together in one place? Or does that not make sense?

blueyed: I thought it would be only useful for failed tests, no?

tests/_runner.lua
blueyed: Yeah.

tests/_runner.lua
psychon: What I meant: Move this up to next to line 114 where `step_as_string` is defined the first time.

blueyed: Got that.  Then it is used for the "running …" msg(s) though also.  I.e. more noisy.

blueyed:
Missing diff coverage is expected.

Should I squash it already (for later auto-merge), or can we squash it with a single review of @psychon then?

tests/_runner.lua
psychon: Ah, okay.

PR #2885, opened 2019-09-29
doc: Add back an accidentally removed section.

Elv13:
This was probably due to copy pasting from the preprocessed file
back to the raw file.

Fix #2882


docs/05-awesomerc.md.lua
```diff
@@ -29,6 +29,7 @@ Here are the modules that we import:
 ]]
 
 sections.DOC_ERROR_HANDLING = [[
+
```
blueyed: Is this newline required?

Elv13: Yes, it changes how the markdown is expanded.

PR #2887, opened 2019-09-30
build: Allow Awesome to be built with luarock LGI.

Elv13:
Copy/paste the same trick as `rc.lua` uses to find `luarocks` module into the build+test scripts.

Mitigates: https://github.com/pavouk/lgi/issues/236

### DIFF
--- a/build-utils/lgi-check.c
+++ b/build-utils/lgi-check.c
@@ -26,6 +26,7 @@
 #include <stdlib.h>
 
 const char commands[] =
+"pcall(require, 'luarocks.loader')\n"
 "print(string.format('Building for %s.', jit and jit.version or _VERSION))\n"
 "local lgi_version = require('lgi.version')\n"
 "print(string.format('Found lgi %s.', lgi_version))\n"
--- a/common/version.c
+++ b/common/version.c
@@ -47,7 +47,8 @@ eprint_version(void)
         lua_pop(L, 1);
 
     /* Either push version number or error message onto stack */
-    (void) luaL_dostring(L, "return require('lgi.version')");
+    (void) luaL_dostring(L, "pcall(require, 'luarocks.loader')\n"
+                            "return require('lgi.version')");
 
 #ifdef WITH_DBUS
     const char *has_dbus = "✔";
--- a/docs/_parser.lua
+++ b/docs/_parser.lua
@@ -1,3 +1,5 @@
+pcall(require, "luarocks.loader")
+
 local gio = require("lgi").Gio
 local gobject = require("lgi").GObject
 local glib = require("lgi").GLib
--- a/tests/_client.lua
+++ b/tests/_client.lua
@@ -4,6 +4,7 @@ local spawn = require("awful.spawn")
 -- It is used to test the `awful.rules`
 
 local test_client_source = [[
+pcall(require, 'luarocks.loader')
 local lgi = require 'lgi'
 local Gdk = lgi.require('Gdk')
 local Gtk = lgi.require('Gtk')
--- a/tests/examples/shims/_common_template.lua
+++ b/tests/examples/shims/_common_template.lua
@@ -1,3 +1,5 @@
+pcall(require, "luarocks.loader")
+
 -- luacheck: globals string
 function string.wlen(self)
     return #self
--- a/tests/test-selection-getter.lua
+++ b/tests/test-selection-getter.lua
@@ -9,6 +9,7 @@ local Gio = lgi.Gio
 local GdkPixbuf = lgi.GdkPixbuf
 
 local header = [[
+pcall(require, "luarocks.loader")
 local lgi = require("lgi")
 local Gdk = lgi.Gdk
 local Gtk = lgi.Gtk
--- a/tests/test-selection-transfer.lua
+++ b/tests/test-selection-transfer.lua
@@ -14,6 +14,7 @@ local large_transfer_piece_count = 3
 local large_transfer_size = #large_transfer_piece * large_transfer_piece_count
 
 local header = [[
+pcall(require, "luarocks.loader")
 local lgi = require("lgi")
 local Gdk = lgi.Gdk
 local Gtk = lgi.Gtk
--- a/tests/test-selection-watcher.lua
+++ b/tests/test-selection-watcher.lua
@@ -4,6 +4,7 @@ local runner = require("_runner")
 local spawn = require("awful.spawn")
 
 local header = [[
+pcall(require, "luarocks.loader")
 local lgi = require("lgi")
 local Gdk = lgi.Gdk
 local Gtk = lgi.Gtk
--- a/tests/test-spawn.lua
+++ b/tests/test-spawn.lua
@@ -12,6 +12,7 @@ local exit_yay, exit_snd = nil, nil
 
 local tiny_client = function(class)
     return {"lua", "-e", [[
+pcall(require, "luarocks.loader")
 local lgi = require 'lgi'
 local Gtk = lgi.require('Gtk')
 local class = ']]..class..[['
--- a/tests/test-titlebar.lua
+++ b/tests/test-titlebar.lua
@@ -4,6 +4,7 @@ local rules = require("awful.rules")
 local spawn = require("awful.spawn")
 
 local tiny_client_code_template = [[
+pcall(require, 'luarocks.loader')
 local Gtk, class = require('lgi').require('Gtk'), 'client'
 Gtk.init()
 window = Gtk.Window {default_width=100, default_height=100, title='title'}


blueyed:
This is too much copy'n'paste for my taste.
Can we do it in a more central place?
Would likely add some overhead to where it is not necessary then though?

Elv13:
> Can we do it in a more central place?

No, this is all the "entry points" for the Lua command. The other option is changing packages.path and some environment variables. However since some of those scrips are callable by random shells, this will create a state where half the place have the variables and the other half doesn't. This would make everything worst. Making mandatory to add the luarocks like in each entry points is simpler.

blueyed:
Wouldn't it also be possible via busted's loader script?

blueyed:
> Wouldn't it also be possible via busted's loader script?

For the tests then only..

In general it would be better to have a central entrypoint for all those instead - that would still be loaded/added there then now, but would allow for changing only once place later.

Elv13:
> In general it would be better to have a central entrypoint for all those instead

They are scripts. They are entry points by definition. I am not sure what you are proposing here. Even if you placed some magic in CMake, short of creating overkill stuff like `run.sh.in` to inject the proper code, it would not work. Some scripts are called using `io.popen` or `awful.spawn` by the integration tests. Some like `awesome --version` are called on a clean Lua context *on purpose* to avoid polluting it. Some are just scripts called during the build, etc, etc.

blueyed:
As for the tests I've thought it could be a central loader script (busted supports this).
As for the others, it could be put into a new separate script (preferably the one that the busted loader also uses then), so there's only a single script to load for them.

Just imagine the next time something like this would be changed: do you want to touch all those files again then?

Elv13:
>  Just imagine the next time something like this would be changed: do you want to touch all those files again then?

This is not what this pull request does. This pull request makes sure the Lua scripts entry points load the environment and paths configured in luarocks (only when it is installed). Any changes are made in luarocks and they will be automatically imported into Awesome.

blueyed:
..and what I mean is that this should be done in 1-2 places only.

Elv13:
> ..and what I mean is that this should be done in 1-2 places only.

Sorry if it sounded harsh. That wasn't my intention. I may lack imagination, but I don't see how what you propose could ever be simpler than this pull request. What you propose is halfway between impractical and impossible. You could maybe at most save 3 of the copy paste (the hardcoded ones in the integration tests). By making some parts of the build process more fragile, which I want to avoid, you may remove 2 more. The first line of `rc.lua` is already a luarocks setup `require`. Then we have hundreds of `local lgi = require("lgi")` and/or it's cairo/pango/glib/gio subcomponents. How are a dozen of luarocks require more problematic? We don't do dependency injection anywhere else. The closest thing to what you ask is how the examples set `LUA_PATH` and C PATH. This is not what this code does, so it is not relevant. Busted is also not an AwesomeWM build dependency, nor do I want it to become one. It's an optional component to run one of the 4 test suits. Plus, whatever it does has nothing to do with that this pull request does anyway.

What this pull request does is to allow having some prefixed luarocks and let it configure Lua correctly so AwesomeWM loads without the package manager provided libraries.

Unless you know some magic to make what you propose simpler and more maintainable than this limited number of extra `require` calls, I don't think overriding environment variables all over the place can be considered a better alternative to the current pull request content. Plus, in the case of `check-lgi.c` and `--version`, it will not work at all.

actionless:
LGTM, but i'm not approving it not to trigger mergify since you requested psychon's review

psychon:
Not much time right now, but I wanted to leave a long-term item from my todo list here:

During build, awesome uses `/usr/bin/lua` to execute Lua code. During runtime, it creates its own Lua context. Nothing makes sure that they actually refer to the same Lua version. This also means that our LGI-check is half-wrong, since it only checks one of these two.

Anyway, my long-term plan is to build our own Lua interpreter and use that during the build. One could make this part of awesome itself and e.g. have `awesome --run foo.lua` execute the lua script and then exit.

If this were to become a reality, there would only be a single entry point. Do you think this is realistic?

----

Alternatively: How can I tell AwesomeWM which Lua version to build against? The answer to this question is "it's complicated". Do you think it would make sense to simplify this a bit and add a `@LUA_INTERPRETER@` variable that can be set through CMake? Then, this could be set to e.g. `/usr/bin/lua5.2` if one wants to.

What does this have to do with this PR? Well, one could then set this to `lua -e 'require("luarocks.loader")` if one wants to.

----

Does any of these two seem sensible?

For this PR itself: Well, what is so special about Luarocks? What if I installed LGI by hand somewhere and want awesome to use that? Also, this also makes everything else in luarocks `require`able, despite the title saying that this only makes it find LGI...
This PR seems like a specific work-around for a problem on your system. I am curious to understand the underlying reason for those crashes, but (if I understand you correctly), it occurs only when loading GTK3 through LGI, but your patch also touches code that does not use GTK...

Another work-around for you (that does not require changes to awesome) would be to use `eval $(luarocks path --bin)` before building awesome. I am not 100% sure about this invocation (`--bin`?) but something like this should work, shouldn't it?

blueyed:
> Another work-around for you (that does not require changes to awesome) would be to use `eval $(luarocks path --bin)` before building awesome. I am not 100% sure about this invocation (`--bin`?) but something like this should work, shouldn't it?

More like `eval $(luarocks --lua-version=5.1 path --no-bin)`, but good point for a workaround probably.  Might not really be require before building also, but only during runtime.

Elv13:
> Another work-around for you (that does not require changes to awesome) would be to use eval $(luarocks path --bin) before building awesome. I am not 100% sure about this invocation (--bin?) but something like this should work, shouldn't it?

It would need to be in a few more place to work, but it might be a better alternative.

It will work for:

 * The documentation preprocessors (they use Gio)
 * The selection getter and setter programs

It can be extended to work for:

 * The integration tests (by copy/pasting this into run.sh too)
 * `check-lgi.c` (kind of work-ish as-is, but needs extra care about where to setup the extra variables to make sure they are tested/set before `check-lgi.c`)
 * The documentation examples (its CMake script messes with the environment, it needs some extra care to make sure they don't clash)

It will not work for:

 * `awesome --version`: It eval in its own context to print the LGI version
 * `rc.lua`: It depends on runtime ENV, not build-time ones. It already required luarocks.
 * The test client and notification client (they can be useful outside of the traditional test framework. They should work without extra magic).

The question is: Is doing what is described above more maintainable than this pull request? I don't think so.

> Well, what is so special about Luarocks? 

We already support it in `rc.lua` (added by you) in commit 202be62157. If we support it, then we support it. If we don't support it, then lets revert those commits and move the burden back to the users.

I tried the luarocks support for the first time last week and it has some serious issues:

 1. When LGI is installed using Luarocks, `awesome --version` explodes, but `rc.lua` loads. It should be the other way around.
 2. The tests cannot be executed with that version of awesome. While it *runs*, the tests scripts which include LGI don't because.
 3. Awesome cannot be built, but it can be executed. This is inconvenient if some power users want to salvage part of the test infrastructure to test their configs. And, obviously, it is inconvenient for me since I decided to stop using the distribution packages because they crash and are outdated.

> This PR seems like a specific work-around for a problem on your system. 

This PR was developed by first removing the system Lua (and all lua related packages), installing it in a prefix, changing `$PATH`, installing luarocks and making that work. It should be the most generic case possible, not something specific to my system.

psychon:
> It can be extended to work for:
> 
> * The integration tests (by copy/pasting this into run.sh too)
> * `check-lgi.c` (kind of work-ish as-is, but needs extra care about where to setup the extra variables to make sure they are tested/set before `check-lgi.c`)
[...]
> * The documentation examples (its CMake script messes with the environment, it needs some extra care to make sure they don't clash)
> 
> It will not work for:
> 
> * `awesome --version`: It eval in its own context to print the LGI version
> * The test client and notification client (they can be useful outside of the traditional test framework. They should work without extra magic).

Huh? Why? I thought Lua just automatically honors the `LUA_whatever` environment variables? So, as long as nothing clears those env vars (which the CMake stuff for `tests/example/` explicitly does, as you write), it should just work.

So, I would expect problems only for:

> * The documentation examples (its CMake script messes with the environment, it needs some extra care to make sure they don't clash)
[...]
> * `rc.lua`: It depends on runtime ENV, not build-time ones. It already required luarocks.
>
> The question is: Is doing what is described above more maintainable than this pull request? I don't think so.

Well, "maintainable" is easy. No code is more maintainable than some code... ;-)

This PR is basically about "support LGI installed in a non-standard location". And the official Lua way for non-standard solutions are the `LUA_PATH` env vars, which is what my suggestion uses.

> We already support it in `rc.lua` (added by you) in commit [202be62](https://github.com/awesomeWM/awesome/commit/202be62157103c34fab12283fe8d6e46df615b67). If we support it, then we support it. If we don't support it, then lets revert those commits and move the burden back to the users.

I think of that commit more of a hint to users on how to integrate luarocks, because it is quite hard to set up environment variables before starting awesome through some DMs...

> 
> I tried the luarocks support for the first time last week and it has some serious issues:
> 
> 1. When LGI is installed using Luarocks, `awesome --version` explodes, but `rc.lua` loads. It should be the other way around.

How does it explode? I just tried `LUA_PATH=/foo awesome --version` and it fails to find `lgi.version`, so I would expect this to just work in your setup.

> 2. The tests cannot be executed with that version of awesome. While it _runs_, the tests scripts which include LGI don't because.

Sorry, I fail to parse that sentence. However, I would expect environment variables to "just work", unless something clears them / overwrites them.

[...]
> This PR was developed by first removing the system Lua (and all lua related packages), installing it in a prefix, changing `$PATH`, installing luarocks and making that work. It should be the most generic case possible, not something specific to my system.

Uhm. How does CMake find `liblua.so` in this setup? Does Luarocks set the necessary magic env vars somehow?
I am wiling to try something like the above (`chmod 0` instead of uninstalling) with a self-built Lua installed to `/tmp/lua-experiment` and seeing if I can make this work (via env vars). I *expect* most of it to just work (after I convince CMake how to find lua), but I am curious about the differences that we end up with.

psychon:
This PR seems to be already merged?!? At least partially. `git grep luarocks.loader` finds six matches in current master.

Anyway, I ended up with a chroot for building awesome. Here are the instructions:
```sh
# create debian stable chroot
mkdir ~/debian-stable-chroot
sudo debootstrap stable ~/debian-stable-chroot http://deb.debian.org/debian
sudo mount proc ~/debian-stable-chroot/proc -t proc
sudo mount sysfs ~/debian-stable-chroot/sys -t sysfs
sudo mount devpts ~/debian-stable-chroot/dev/pts -t devpts

# Get the necessary sources
git clone https://github.com/awesomeWM/awesome.git  ~/debian-stable-chroot/awesome
git clone git://github.com/pavouk/lgi.git ~/debian-stable-chroot/lgi
wget -O ~/debian-stable-chroot/lua-5.3.5.tar.gz https://www.lua.org/ftp/lua-5.3.5.tar.gz
( cd ~/debian-stable-chroot && tar -xvzf lua-5.3.5.tar.gz )

# Enter the chroot
sudo chroot ~/debian-stable-chroot /bin/bash
# From now on, everything is in the chroot

# Install necessary dependencies (this does NOT install lua or lgi)
apt-get install -y build-essential libreadline-dev cmake imagemagick pkg-config libxcb-image0-dev libxcb-cursor-dev libxcb-util0-dev libxcb-keysyms1-dev libxcb-icccm4-dev libxcb-xfixes0-dev libxcb-randr0-dev libxcb-xtest0-dev libxcb-xinerama0-dev libxcb-xkb-dev libxcb-xrm-dev libxkbcommon-x11-dev libstartup-notification0-dev libpango1.0-dev libgirepository1.0-dev libxdg-basedir-dev libgdk-pixbuf2.0-dev xvfb xterm xdotool dbus-x11 x11-apps x11-xserver-utils libdbus-1-dev gir1.2-gtk-3.0 gir1.2-rsvg-2.0 libnotify-bin

mkdir /var/run/dbus
dbus-daemon --system

# Build Lua (but do not install it, just to make things more complicated)
# (Also, this builds a static library, not a dynamic one)
cd /lua-5.3.5
make linux
export PATH="$PATH:/lua-5.3.5/src"

# Build LGI
cd /lgi
make LUA_CFLAGS='-I/lua-5.3.5/src' LUA_LIB=/lua-5.3.5/src/liblua.a
export LUA_PATH='/lgi/?.lua;;'
export LUA_CPATH='/lgi/?.so;;'

# Finally, build awesome and run its tests
cd /awesome
export HEADLESS=1 # So the tests use Xvfb instead of Xephyr
make CMAKE_ARGS='-DLUA_INCLUDE_DIR=/lua-5.3.5/src "-DLUA_LIBRARIES=/lua-5.3.5/src/liblua.a;/usr/lib/x86_64-linux-gnu/libdl.so;/usr/lib/x86_64-linux-gnu/libm.so" -DOVERRIDE_VERSION=3' all check
```
No patches to awesome needed. `$PATH`, `$LUAT_PATH`, and `$LUA_CPATH` are enough for everything.

psychon:
I just tried the above also with a commit before all of those `luarocks.loader` things were added: 9c0e16e6234657e7fa3a98fbdb5433ffa6f97580 builds and tests without problems

actionless:
> Anyway, I ended up with a chroot for building awesome. Here are the instructions:

in that case i think this snippet need to be shared somewhere in the docs and/or referenced in readme

Elv13:
ok then

PR #2891, opened 2019-10-02
Fix memory leaks

Timidger:


Timidger:
Ignore, github for some reason saw fit for it to open this against upstream for whatever reason...

PR #2893, opened 2019-10-05
API standardization merge part 1

Elv13:
Splitting the other pull request into < 20 commit chunks so it can pass the CI before the 50 minute timeout.

As a word of warning from the other PR message, this has the potential to break quite a few things (please report those as bugs). Paying technical debt is never fun, nor is layers upon layers of compatibility hacks to avoid breaking people configs. The PR series has both, lots of both. The end result is making the API more predictable, less ugly, easier to learn and more future proof. The dark side is how I get there.

I will merge this when the test pass. These commits are 10 months old and unmodified, there was plenty of time to complain and I had hoped some people would have, but the time is out for that.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -26,7 +26,6 @@ read_globals = {
     "key",
     "keygrabber",
     "mousegrabber",
-    "root",
     "selection",
     "tag",
     "window",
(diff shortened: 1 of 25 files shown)


Elv13:
Sigh... I will have to do something about all those flaky test issues we have as of late. It's not "us", it's about getting over provisioned Travis builder and some clients never spawning and/or spawning in very unlikely order. Anyway, after restarting the tests enough time, it proves it isn't due to this PR (which does not affect client spawning).

PR #2894, opened 2019-10-06
API standardization part 2

Elv13:
More old commits following the ones I merged yesterday. There is 3 more of these PRs before all legacy properties (and `:setup()` vs. `.widget` for the widgets) get standardized to Lua properties.

These commits need more testing and I had a pretty hard time rebasing them. All the keygrabber fixes done in the last year might have little regressions. I will make sure to add extra tests before this goes forward.

There's some diff noise due to the keybindings being moved to their own sections (as currently shown in `mod4+s`). This will eventually be removed in favor is "actions". The client keys will also probably be moved to a `request::` signal to simplify adding and removing keys per-client class in the client rules.

### DIFF
--- a/tests/examples/shims/screen.lua
+++ b/tests/examples/shims/screen.lua
@@ -31,8 +31,11 @@ local function create_screen(args)
     local s = gears_obj()
     awesome._forward_class(s, screen)
 
-    s.data = {}
+    s._private = {}
     s.valid = true
+
+    -- Deprecated.
+    s.data = s._private
 
     -- Copy the geo in case the args are mutated
     local geo = {
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -219,25 +219,56 @@ end)
 
 -- {{{ Mouse bindings
 -- @DOC_ROOT_BUTTONS@
-root.buttons = {
+awful.mouse.append_global_mousebindings({
     awful.button({ }, 3, function () mymainmenu:toggle() end),
     awful.button({ }, 4, awful.tag.viewnext),
     awful.button({ }, 5, awful.tag.viewprev),
-}
+})
 -- }}}
 
 -- {{{ Key bindings
 -- @DOC_GLOBAL_KEYBINDINGS@
-globalkeys = {
+
+-- General Awesome keys
+awful.keyboard.append_global_keybindings({
     awful.key({ modkey,           }, "s",      hotkeys_popup.show_help,
               {description="show help", group="awesome"}),
+    awful.key({ modkey,           }, "w", function () mymainmenu:show() end,
+              {description = "show main menu", group = "awesome"}),
+    awful.key({ modkey, "Control" }, "r", awesome.restart,
+              {description = "reload awesome", group = "awesome"}),
+    awful.key({ modkey, "Shift"   }, "q", awesome.quit,
+              {description = "quit awesome", group = "awesome"}),
+    awful.key({ modkey }, "x",
+              function ()
+                  awful.prompt.run {
+                    prompt       = "Run Lua code: ",
+                    textbox      = awful.screen.focused().mypromptbox.widget,
+                    exe_callback = awful.util.eval,
+                    history_path = awful.util.get_cache_dir() .. "/history_eval"
+                  }
+              end,
+              {description = "lua execute prompt", group = "awesome"}),
+    awful.key({ modkey,           }, "Return", function () awful.spawn(terminal) end,
+              {description = "open a terminal", group = "launcher"}),
+    awful.key({ modkey },            "r",     function () awful.screen.focused().mypromptbox:run() end,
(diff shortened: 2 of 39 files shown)


Elv13:
Ok, I finally found a not-too-horrible way around the memory leak regression which eluded me for some weeks now. The problem was well known in the past, but made worst by this changeset (there is already code in `awful.titlebar` to mitigate the issue). In summary, the titlebar button callback usually have the client object in their environment. The button is then attached to a widget, which is attached to the titlebar drawin, which is part of the titlebar, which is part of the client. Lua 5.2+ can solve this loop, but v5.1 and luajit can't. To fix it, I had to use a weak table (`__mode="k"` with the button as key) to store the button properties. Then in *that*, have a `__mode ="v"` table to store the callbacks. Then *on top of that*, use connect_signal to add a reference to the callback on the `capi.button` to prevent the `__mode="v"` from freeing it. Storing the callbacks as property is needed because otherwise it's impossible to disconnect them later since they are often lambdas part of a declarative construct. This seems to work.

Otherwise, compared to the initial version of this PR, there is the following changes:

 * Bring back `awful.keyboard`. @psychon initially rejected that module on the ground that I hardcoded the name of the modifier keys. This has since been resolved and added to `capi`. The module, as of this PR, only has the ability to add/remove global keybindings. Another PR will add the previously rejected modifier key status, but with the now correct capi code to back them. It will also add keybinding collision detection and provide a real public API @actionless hotkey popup can query to get the keybindings.
 * Add the `group` and `description` keybindinds property the @actionless hotkey popup already uses and make them official object properties.
 * Try harder to convert the `capi.button`/`capi.key` back to `awful.button`/`awful.key` objects. This is needed to provide the future support for keybinding collision handling.
 * Restore the "ability" to pass `boolean`s to the `:keys()` and `:buttons()` method since apparently this worked before and is used in the wild. Technically, its both a bug and an undocumented behavior that it ever worked, but I agree it is convenient and make the user code more readable.
 * Support `args` based constructors in addition to the current constructor for `awful.key` and `awful.button`. Both constructor and official and non-deprecated. They both have their use case, so I wont deprecate the old one just because the new one is more consistent with newer code.
 * Support `awful.key` objects in `awful.keygrabber`. While the constructor looked the same, you previously could not use "real" keybindings in the keygrabber. I will do the same to the prompt in a later PR.
 * It has some tests. Maybe not a stellar coverage, but at least a decent one.
 * It is 3x larger...

To anyone who use the keygrabber, I strongly suggest you test this soon before I merge this. It may take a couple more iterations to get all commits 100% free of test failure, but this is nearly there, so I will probably merge this next weekend and move to the next PR. It wont be part of the standardization series, but rather a follow-up to this one with new features now possible with the cleaned up input API. It has the ability to have default keybindings/mousebindings for clients, moves `awful.key`/`awful.button` to `wibox` (because they already depend on them anyway, voiding the dependency restriction we have in place), expose the currently private modifier status API, add some signals, provide collision detection and allow the hotkey popup to use a less hacky source of truth as mentioned above.

actionless:
thanks for the update (and work on this of course :3), will be keeping an eye on this, to make sure to refactor hotkeys widget as it will be ready :+1:

tests/examples/shims/screen.lua
```diff
@@ -31,9 +31,12 @@ local function create_screen(args)
     local s = gears_obj()
     awesome._forward_class(s, screen)
 
-    s.data = {}
+    s._private = {}
     s.valid = true
 
+    -- Deprecated.
```
actionless: mb also add info in which version it will be removed? (or better use gears deprecate so it will print the warning?)

Elv13: done

Elv13:
I merged this. The CI problem is that `placement` test that has been flacky for a while due to races in launching clients. I don't want to spend the next 15 hours pressed the "restart" button in Jenkins, so I merged it manually.

As always, if you read this, please test. Those modernization/standardization PRs are extremely regression prone. There is no way around user testing since the way people use these APIs far exceed any single developer imagination. Some undocumented behaviors also tend to be relied upon in surprising ways. There is no way to know that beside user testing. In the end everybody benefit from both the modernization and from the bug reports since, in both case, they come with new regression tests.

alfunx:
Was this supposed to be a breaking change? `root.keys(globalkeys)` doesn't work anymore. Besides that, this is fine so far. Couldn't migrate my `keygrabber`s yet, will do soon.

psychon:
> Was this supposed to be a breaking change?

Did anything happen with this problem? (= Fixed already? Anyone investigated it?) I am confused at what is going on, but "breaking change" sounds troublesome.

actionless:
@psychon i'm using latest git master and it still works for me: https://github.com/actionless/awesome_config/blob/master/config/keys.lua#L686

PR #2895, opened 2019-10-07
Move `awful.rules` to `ruled.client`.

Elv13:
(first, ignore the line count, there is an SVG image taking 99% of the raw diff)

This PR goes with #2853. I pried it out of my experimental branch. I don't expect to pass the tests anytime soon. This wont be merged in the short term. Thanks to the recently merged PRs, I am now unblocked to submit many of these pull requests which depended either on the new screen API or on the sequence template. They are part of a larger effort to have rules for most core objects (client, notification, screen, tag, clipboard, etc). Note that some commits are shared with the other rules PRs. The first one which gets merged will get them in, then the other will get rebased.

This pull request also start to use `.append_rule` instead of setting the rules once. This allow modules to add rules. It also uses `request::rules` instead of doing it from `rc.lua` root context. This allows `rc.lua` to be fully parsed/executed before the rules gets added. Again, it allows modules to be on the same ground as `rc.lua` rules. In the long run, moving more and more of the un-undoable code from `rc.lua` into append/add/remove operations in signals will allow more modular config to be created on top of the default `rc.lua`. This will allow users to share self contained modules like never before. Until now building such modules (and I built plenty) required many `delayed_call` to undo what `rc.lua` did using wizardry.

Finally, this PR moves a lot of previously hardcoded examples in the doc to the new sequence template.

image


Elv13:
Rebased and extended. The client documentation has been extensively expanded and ported to the latest ldoc tags.

@psychon Please review and make a commit to the client documentation commit. There is some simplifications I made to make it easier for our new users to understand. However you might disagree and like to reword some of them. I would appreciate if you could take a look. Again, don't trust the line count, it's because GitHub treats SVG as text files.

Otherwise this will be my next PR. Given the large size of the client changes, it will be hard/painful to rebase and thus has to go in. The notification PR equivalent needs some work since I wrote more tests and discovered rather glaring issues with multi-screen handling (plus some other user-reported bugs). Those 2 PRs should get in before I submit the screen_rule/activity one(s) since they both contain some groundwork commits for the ruled framework.

objects/client.c
```diff
@@ -363,151 +377,219 @@
...
  */
 
 /**
  * The client PID, if available.
  *
- * **Signal:**
- *
- *  * *property::pid*
+ * Some clients might have funky PID `fork` behavior, but
+ * normally this will never change.
```
psychon: No idea what this is supposed to mean

objects/client.c
```diff
@@ -797,103 +936,168 @@
...
  */
 
 /**
  * The client's bounding shape as set by awesome as a (native) cairo surface.
  *
- * @see gears.surface.apply_shape_bounding
+ * The bounding shape is the outer shape of the client. It is outside of the
+ * border.
+ *
+ * Do not use this directly unless you want total control over the shape (such
```
psychon: Can one even do so? I think `awful.client.shape` will interfere as soon as someone starts e.g. `xeyes` and now two implementations will "fight" over the shape.

docs/common/rule.ldoc
```diff
@@ -50,6 +50,24 @@
...
+-- The comparison will be made using the lesser (`<`) operator.
+--
+-- @rulecomponent rule_lesser
+-- @param table
+-- @see rule
+-- @see except
+
+--- A table which content will be compared to the target object current properties.
+--
+-- The comparison will be made using the lesser (`>`) operator.
```
SethBarberee: Shouldn't this be `The comparison will be made using the greater (`>`) operator`?

objects/client.c
```diff
@@ -797,103 +936,168 @@
...
+ * as shape with holes). Even then, it is usually recommanded to use transparency
```
SethBarberee:
```suggestion
 * as shape with holes). Even then, it is usually recommended to use transparency
```

awesomerc.lua
```diff
@@ -439,60 +440,60 @@ end)
...
     -- All clients will match this rule.
-    { rule = { },
-      properties = { focus = awful.client.focus.filter,
-                     raise = true,
-                     screen = awful.screen.preferred,
-                     placement = awful.placement.no_overlap+awful.placement.no_offscreen
-     }
-    },
+    ruled.client.append_rule {
+        id         = "global",
```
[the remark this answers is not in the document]
SethBarberee: I'm assuming this is used with `gears.matcher` since you can check if a specific rule exists

awesomerc.lua
```diff
@@ -439,60 +440,60 @@ end)
...
     -- @DOC_GLOBAL_RULE@
     -- All clients will match this rule.
-    { rule = { },
-      properties = { focus = awful.client.focus.filter,
-                     raise = true,
-                     screen = awful.screen.preferred,
-                     placement = awful.placement.no_overlap+awful.placement.no_offscreen
-     }
-    },
+    ruled.client.append_rule {
```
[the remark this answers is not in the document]
Elv13: > That seems pointless.

I could invent some convoluted reasons, but there is actually a good one for this. First of all, doing this for the sake of consistency is good enough on its own.

However the real reason is the new screen API (and its rules, coming after this PR gets merged). I want to get rid of all user code running in the global context (code directly in the top-level indentation rather than a signal handler) and eventually fully deprecate doing so. With the screen rules, `mouse.screen` is `nil` when AwesomeWM starts, which is unsupported by most APIs. The theme API is also uninitialized, which is already causing issues if people run code ahead of `beautiful.init()`. On top of that, if users run code before `awful` is `require()`d, it also already blows up quite bad.

Updating the documentation to say what is and isn't available would to impossible to maintain and make the user experience more complicated that is has to be.

All in all, running code in the global context is already too fragile to support. Having a "managed" initialization where everything is called only after it becomes "really" available is much safer and easier to support.

> rules can be added to instead of having to call some magic global function

There is such an object (`gears.matcher`), but its hidden. Using that as a signal parameter would be interesting. It would allow to deprecate the global API. The magic `::connected` signal API could be used to make sure all new `connect_signal` gets called even if they come later.

The main problem with this would be that the full `gears.matcher` API cannot currently get exposed for all rules type. I fear that exposing such low-level object would enable some self-foot-shooting opportunities comparing to restricting the API to a known safe subset. Given that each type of rules (client, notification, screen, tags, clipboard) have a different safe subset, I am not sure how this could be documented. Given only 1 `gears.matcher`. object can exist per type of rules, I don't see why using the object oriented approach here would make the API simpler.

In any case, if you think we should still use objects in the signal callback, I will do it. I am fine either way.

> (which I think they should NOT do; what's a valid use case for that?!?)

Mostly simplicity. The default config has some built-in rules. A module could use the `id` to disable them. This is simpler for the non-programmer users than having instructions to remove code from `rc.lua`. That's about the only "good" use case.

The reason there is an `append_`/`remove_` is to allow modules to add new rules without fear that another one (or `rc.lua`) will just set `awful.rules.rules` directly and overwrite what they just did. Remove is added to all those APIs for consistency rather than having a real use case.

awesomerc.lua
```diff
@@ -22,7 +24,6 @@ require("awful.hotkeys_popup.keys")
 -- {{{ Error handling
 -- Check if awesome encountered an error during startup and fell back to
 -- another config (This code will only ever execute for the fallback config)
--- @DOC_ERROR_HANDLING@
```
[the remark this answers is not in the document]
Elv13: Rebase issue. This PR is (originally) older than the notification API (merge).

lib/gears/debug.lua
```diff
@@ -135,7 +135,15 @@ end
 -- @tparam string new_name The new class name
 -- @treturn table A proxy class.
 -- @staticfct gears.debug.deprecate_class
-function debug.deprecate_class(fallback, old_name, new_name)
+function debug.deprecate_class(fallback, old_name, new_name, args)
```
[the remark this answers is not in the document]
Elv13: done

awesomerc.lua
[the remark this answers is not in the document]
Elv13: Done

awesomerc.lua
```diff
@@ -13,6 +13,8 @@ local wibox = require("wibox")
 local beautiful = require("beautiful")
 -- Notification library
 local naughty = require("naughty")
+-- Declarative object management
+local ruled = require("ruled")
```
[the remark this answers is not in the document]
Elv13: I re-ordered the commits. The CI doesn't detect it because `awesomerc.lua` is post-processed by CMake. Another instance of this causing CI issues is when a file is added in a later commit, "check each commit" will always have it since its copy will be in the build directory. This happens because Travis run make for the last commit in the PR. This calls `configure_file`. Then the first commit is checked out. So the build directory is polluted with artifacts of the last commit.

objects/client.c
```diff
@@ -363,151 +377,219 @@
...
+ * `client.focus = c`. Setting the focus directly bypasses
+ * all the filters and emits fewer signals, which tend to
+ * cause unwanted side effects and make it harder to alter
+ * the code behavior in the future. It usually takes *more*
+ * code to use this rather than `:activate{}` because all
+ * the boilerplate code (such as `c:raise()`) needs to be
+ * added everywhere.
+ *
+ * The main use case for this field is to check *when* there
+ * is a focused client.
```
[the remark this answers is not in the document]
Elv13: I added the example. I am not sure how to generally handle this. `focus` and `active` are non-symmetrical concepts. I am not the one who started calling stuff `active` in the API, so I just deal with it. The *active* client must be the focused client, even if it is not on top. *Activating* a client is more than just focusing it. It involves potentially raising it, unminimizing it and maybe changing the current tag(s). These days (ever since v4.0), changing the focus using `client.focus` causes undesirable side effects.

Over the last years, I kept preferring the "active" word over the "focus" word across the API. Another PR, the phase 5 of the API standardization serie, adds `awful.client.focused()` to mirror `awful.screen.focused()` and formally deprecates `client.focus`.

"focus" would have probably been a better word all along, but its very late to fix that. Qt also has this non-symmetrical behavior in the API, but calls it focus either way.

objects/client.c
```diff
@@ -363,151 +377,219 @@
...
+ * the `awful.widget.tasklist` and should not be shown there.
  *
  * @property skip_taskbar
- * @param boolean
+ * @tparam[opt=false] boolean skip_taskbar
+ * @propemits false false
+ * @see sticky
  */
 
+//TODO v5: Rename to skip_tasklist?
```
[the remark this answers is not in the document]
Elv13: I am not sure what to do. I don't like multiple words to be used for the same concept even if this is due to the specification. Anyway, this isn't an issue for now.

objects/client.c
```diff
@@ -363,151 +377,219 @@
...
+
 /**
  * The window type.
  *
+ * This is useful in the `ruled.client` rules to apply different properties
+ * depending on the client types.
```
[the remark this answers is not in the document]
Elv13: I made changes to lessen this feeling. I added "among other places". In general, I am fine with referring the rules behavior in the core objects documentation. The users are more likely to interact with those properties using the rules than with imperative code. Some hand holding doesn't hurt.

objects/client.c
Elv13: I replaced the text. tl;dr; I referred to `fork()` and how it can mess things up. As far as I understand (please correct me), it is even possible kill a process and have another one later on claim ownership of a window. Even as far as transferring ownership to another computer. Do the *protocol* explicitly prevents this? I am aware most (all) GUI toolkits explicitly prevent this.

objects/client.c
Elv13: I also wondered about this when writing this. Should we deprecate this and officially call it a private API?

docs/common/rule.ldoc
Elv13: thanks, done

objects/client.c
Elv13: done

Elv13:
I fixed some commits so they pass the tests individually. The only 2 remaining CI issues is the placement one I cannot reproduce and the `@rulelibs` ldoc tags missing because `ruled/client.lua` is left over in a dirty build directory, which is a false positive.

The other changes I did is to improve upon what @psychon says that messing with the rules should not be possible. Doing `awful.rules.rules = {}` and `table.remove(awful.rules.rules, 2)` now bark a lot more (but work). I also added that to the "Behavior changes" section of the changelog. Doing `ruled.remove_rule` is still supported. All integration tests messing with the rules had to be modified avoid aborting the CI run.

There is some comments above that need to be answered. If it turns out @psychon agree with leaving things as-is, then this is ready to be merged. On the other hand, if he really want that `request::rules` matcher object to be passed in the callback, then I will have to redesign some of this.

objects/client.c
psychon: When a process dies, its X11 connection is closed. The X11 server will then destroy all resources that are owned by that client, thus the window should be closed. One can tell the X11 server "please don't destroy my resources when I disconnect", but the only use that I ever saw of that was the wallpaper (you do not want the wallpaper to disappear when the wallpaper setter program closes its X11 connection).

There is no built-in way to transfer ownership, I think, but "ownership" is just "which stuff has to be destroyed when this X11 connection closes", nothing else (AFAIK).

objects/client.c
psychon: Dunno. One could "get rid" of `awful.client.shape` via e.g. `package.loaded["awful.client.shape"] = {}` and write an own version in the config.

psychon:
> There is some comments above that need to be answered. If it turns out @psychon agree with leaving things as-is, then this is ready to be merged. On the other hand, if he really want that request::rules matcher object to be passed in the callback, then I will have to redesign some of this.

Feel free to just ignore me. You have the big picture. I am running a version of AwesomeWM that is far beyond git/master (haven't updated since September).

objects/client.c
Elv13: Ok, thanks for the clarification. That demo where someone was moving an application across computers was probably using some connections proxies or something, I can't even find it anymore. I removed that paragraph.

objects/client.c
Elv13: Ok. That seems hacky. I will think about it before we get to 5.0 to see if hiding those APIs is the best way forward. We will need to come up with some kind of plans for shape with holes in them. Maybe compositing is the way to go. I can't really think of good use cases where it would be worth it, but maybe I lack imagination.

Elv13:
> Feel free to just ignore me. You have the big picture.

This isn't really an ideal situation to be in. If you don't have time (or lost interest, I don't know), then I guess there isn't much to do. Some of these PRs I merge are way too scary to be merged with the little amount of review I get. Still, while we have different vision of what it is to maintain software, I try to keep the boat afloat the best I can. To me that's documentation, API design (hiding all the ugliness under layers of plasters), trying to fix real world issues with form factors which have never really been all that well supported (HiDPI, frequent screen count change, input/gestures, etc) and keeping up with the specs (notification v1.2, XDG menu etc). v4.4 will be gigantic with my foolish attempt to "fix" everything. The fact that I don't have much time either just drags the time it will take to release that thing well into +infinity. Even this PR is mostly from 2018. The notification one is from 2017.

Anyhow, I am merging this when the CI finishes. Due to the large amount of client doc changes, it cannot be rebased easily (or I will end up missing rebase issues as was the case many time before).

Elv13:
It is merged. The request::rules signal arguments are still to be discussed. Right now there is zero and the other global function are not deprecated (they come from `awful.rules`, they have to be there to preserve the API compatibility). This can change in later commits.

psychon:
> Some of these PRs I merge are way too scary to be merged with the little amount of review I get.

Remember back when I rewrote the widget system? The old one had widgets in C and lots of hacks around "image" widgets. The new one... well, I am not sure how much review it got. At least I was surprised (and still am) that only few bugs were found.
I guess what I want to say is: Such is open source development. ;-)

Anyway, thank you very much for your work and for actually improving things. Also, it was in my most active times that people started complaining about all the breakage. You, however, value stability and good documentation.

awesomerc.lua
```diff
@@ -439,60 +441,60 @@ end)
...
-        -- and the name shown there might not match defined rules here.
-        name = {
-          "Event Tester",  -- xev.
+    ruled.client.append_rule {
+        id       = "floating",
+        rule_any = {
+            instance = { "copyq", "pinentry" },
+            class    = {
+                "Arandr", "Blueman-manager", "Gpick", "Kruler", "Sxiv",
+                "Tor Browser", "Wpa_gui", "veromix", "xtightvncviewer"
```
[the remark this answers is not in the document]
Elv13: Some were removed. DownThemAll and KAlarm are rewrites of their original implementations. So I assume those rules no longer worked.

Also, since the rules are easier to use from modules now, it will be easier for modules dedicated to rules to pop up. We don't need to have that many in `rc.lua` beside being a good example beginners will appreciate.

awesomerc.lua
```diff
@@ -439,60 +441,60 @@ end)
...
+        id         = "titlebars",
+        rule_any   = { type = { "normal", "dialog" } },
+        properties = { titlebars_enabled = true      }
+    }
 
     -- Set Firefox to always map on the tag named "2" on screen 1.
-    -- { rule = { class = "Firefox" },
-    --   properties = { screen = 1, tag = "2" } },
-}
+    -- ruled.tag.append_rule {
```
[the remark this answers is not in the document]
Elv13: Indeed. `ruled.tag` isn't ready yet.

awesomerc.lua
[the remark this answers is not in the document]
Aire-One: Yes, right :)

I was just curious to know if there is any reason on why some applications were removed and others stay.

lib/ruled/client.lua
```diff
@@ -0,0 +1,717 @@
...
+                -- It isn't possible to just set it to {}, there is other
+                -- references to the table.
+                for k2 in pairs(crules._matching_rules["awful.rules"]) do
+                    crules._matching_rules["awful.rules"][k2] = nil
+                end
+            else
+                crules:append_rules("awful.rules", v)
+            end
+        else
+            rawset(k, v)
```
blueyed, line 697: @Elv13 
This `rawset` appears to be wrong.

I am seeing an error because of this, which probably at least could be more helpful / meaningful - if it cannot be fixed to "just keep working":

```
error while running function!
stack traceback:
        /usr/share/awesome/lib/ruled/client.lua:894: in function '__newindex'
        /home/user/.config/awesome/lib/stop_unfocused.lua:53: in main chunk
        [C]: in function 'require'
        /home/user/.config/awesome/rc.lua:170: in main chunk
error: /usr/share/awesome/lib/ruled/client.lua:894: bad argument #1 to 'rawset' (table expected, got string)
```

This is triggered through some monkeypatching / manually adding a function to `awful.rules`:

```lua
if not awful.rules.get_props_and_callbacks then
function awful.rules.get_props_and_callbacks(c, _rules)
    local matching_rules = awful.rules.matching_rules(c, _rules)
    if #matching_rules == 0 then
        return
    end

    local props = {}
    local callbacks = {}
    for _, entry in ipairs(matching_rules) do
        gears.table.crush(props, entry.properties or {})
        if entry.callback then
            table.insert(callbacks, entry.callback)
        end
    end
    return props, callbacks
end
end
```

lib/ruled/client.lua
```diff
@@ -0,0 +1,717 @@
...
+
+capi.client.connect_signal("scanning", request_rules)
+
+--@DOC_rule_COMMON@
+
+return setmetatable(module, {
+    __newindex = function(_, k, v)
+        if k == "rules" then
+            gdebug.deprecate(
+                "Use ruled.client.append_rules instead awful.rules.rules",
```
blueyed, line 681: s/instead/instead of/ ?  (but it is not clear to me currently if that would be correct then really)

Issue #2897, opened 2019-10-08
Strange errors with rules


vp1981:
Output of `awesome --version`:

awesome v4.3-486-gf294ab37 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

**How to reproduce the issue:**

Not sure because seems default `rc.lua` works fine but mine, worked well until recent changes (after commit 78c349677099e06f6059c7a6ea5adee3e6c1713b I think), produces four error messages for every new window.

**Actual result:**

No error messages.

**Expected result:**

Error messages appear after every new window is shown, see screenshot.

![2019-10-08-084923_577x167_scrot](https://user-images.githubusercontent.com/287444/66375668-84b65a00-e9e0-11e9-8685-12593870e9aa.png)

According to [`~/.xsession-errors`](https://gist.github.com/vp1981/5a53cef5d5016fb1b21d70d86923f5f5) these "errors" connected with `gears/matcher.lua` and `awful/rules.lua`. I checked my `rc.lua` several times, adapted it to some changes, tried without library `lain` but still see these messages.

My [`rc.lua`](https://gist.github.com/vp1981/9807cb2cd005244d9d9d678f35608f5c).

P.S. Besides these messages all seems work fine.
P.P.S. I had to comment the "focus" and "unfocus" stuff because otherwise I get much more error messages (but of different type).

wimstefan:
I can confirm this behaviour.

```sh
awesome v4.3-486-gf294ab37-dirty (Too long)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

I haven't changed my [rc.lua](https://github.com/wimstefan/dotfiles/blob/master/config/awesome/rc.lua) for quite a while  ....

Elv13:
I made a PR with an emergency fix. Please test. It wont fix the problem fully, but should prevent the error and "kind of work" until the proper solution and tests are ready and reviewed.

wimstefan:
> I made a PR with an emergency fix. Please test. It wont fix the problem fully, but should prevent the error and "kind of work" until the proper solution and tests are ready and reviewed. There might be more problem the tests missed.
> 
[...]

PR #2898, opened 2019-10-09
legacy: Temporary workaround for #2897.

Elv13:
Temporary workaround until next weekend. I will merge this real soon to prevent more damage when people upgrade. A more complete solution with tests will have to wait until Saturday.

The `:keys()` and `:buttons()` APIs moved from get/set single methods to properties. It works fine if you use the new or old API, but has limitations when mixing them. `awful.rules` calls properties in a loop after checking if it is a function. Thus it triggers the secondary codepath to try to handle the mixed new/old API case.

This codepath was tested with gears.objects based components (ie. widgets). It was not tested with clients and tags, and it didn't work because they use `awful.tag.getproperty` and `awful.client.property.get` instead of `._private` like all newer components. Those old functions are officially deprecated, but used by tons of configs and modules ported from v3.5 and thus still the default way to access Lua properties in our implementation.

This commit adds a `_private` to anything that doesn't have one to at least make the error stop. It will "mostly" work until a more complete solution is added. Reverting the 2 PRs that changed this
would delay getting more feedbacks. The more complete solution is to move `awful.tag.getproperty` and `awful.client.property.get` to be implemented on top of `_private`.

### DIFF
--- a/lib/gears/object/properties.lua
+++ b/lib/gears/object/properties.lua
@@ -161,9 +161,15 @@ function object._legacy_accessors(obj, name, capi_name, is_object, join_if, set_
     magic_obj["get_"..name] = function(self)
         self = is_object and self or obj
 
+        --FIXME v5 all objects should use _private instead of getproperty.
+        if not self._private then
+            self._private = {}
+        end
+
         self._private[name] = self._private[name] or copy_object(
             obj, {}, name, capi_name, is_object, join_if, set_empty
         )
+
 
         assert(self._private[name])
         return self._private[name]


blueyed:
btw: it is unfortunate that you cannot really tell where the error originates really (in the traceback).  Can this be improved (unrelated to this PR)?

Elv13:
> btw: it is unfortunate that you cannot really tell where the error originates really (in the traceback). Can this be improved (unrelated to this PR)?

In that case the traceback originates from `awful.rules` loop to apply all properties to client. There is no way to get the meaningful property it was trying to see. However in the case it is either "buttons" or "keys". The regression is caused because the rules code attempted to put the `gears.table.join` result using the new property which expects a list of `awful.key` or a list of `awful.buttons`.

blueyed:
> In that case the traceback originates from `awful.rules` loop to apply all properties to client.

Could we wrap that with `pcall`, so that we can add more info when something fails there?
I assume it is done for explicit user callbacks already, but might make sense to wrap it in total also maybe.

Elv13:
Good idea. This would give better backtraces when modules add semi-broken properties to clients, tag and screens.

PR #2900, opened 2019-10-10
Remove FAQ section "Why are new clients urgent by default?"

codebling:
The solution in "Why are new clients urgent by default?" prevents minimised windows from being restored.


blueyed:
Are clients urgent by default after all?  (sorry for being lazy/ignorant)
If so the section should get fixed / extended instead I guess.

codebling:
So...it's not even clear to me what "urgent" means. If it means that new windows steal focus from other windows, then I have a solution that works for that, and maybe we could rename the question?

actionless:
@codebling "urgent" means WM_HINT_URGENT on window

how it looks "in real life"?

1) imagine you're on tag "1"
2) some already opened application on tag "4" could receive an event (think of new message in Pidgin) or  you open new application, which by awful.rules opens on tag "4"
3) at your taglist tag "4" marks red, as it have urgent clients on it

blueyed:
Yeah, so it's a misunderstanding.
A good way to test/trigger this is `sleep 3; printf '\a'` in a terminal, then go to another tag (if your terminal is configured to trigger the urgent flag on a bell etc).

codebling:
So, after a few more tests,
> Are clients urgent by default after all? (sorry for being lazy/ignorant)

Yes, they are! 

Ok, this seems to fix it: 

```diff
client.disconnect_signal("request::activate", awful.ewmh.activate)
function awful.ewmh.activate(c)
+    if c:isvisible() or c.minimized then
-    if c:isvisible() then
        client.focus = c
        c:raise()
    end
end
client.connect_signal("request::activate", awful.ewmh.activate)
```

I prefer this code, which doesn't modify the original activate:
```
client.disconnect_signal("request::activate", awful.ewmh.activate)
function newactivate(c)
    if c:isvisible() or c.minimized then
        awful.ewmh.activate(c)
    end
end
client.connect_signal("request::activate", newactivate)
```


What do you think

Elv13:
Well, the `request::activate` handler is meant to be disconnected and replaced by something that suits your needs. I am not convinced the default should change. In the past, doing those kind of "fixes" led to other user complaining that something regressed.

codebling:
> Well, the `request::activate` handler is meant to be disconnected and replaced by something that suits your needs. I am not convinced the default should change. In the past, doing those kind of "fixes" led to other user complaining that something regressed.

I'm not clear on what you mean. It sounds like you might think I'm proposing to change the default `rc.lua`? I'm only proposing to change the documentation because the current solution in the FAQ is broken. The diff in my last comment shows the change I'm proposing, or the second one is the same thing but probably more future-proof.  

The change allows windows to be restored by clicking on them in the taskbar, which is currently broken if the user follows the instructions in the FAQ. A bit of a deal-breaker for most people, I think. If you're ok with making the change, pick one of the two solutions from the comment above and I'll update the PR with it.

blueyed:
> I prefer this code, which doesn't modify the original activate:

It makes sense to use a new/separate function here in general.  So the FAQ should be updated in this regard then anyway.

Apart from that, the default `request::activate` handler changed a lot since the FAQ was written apparently (https://github.com/blueyed/awesome/blob/c41cf95eeccf31169f96b64537d02c4e630144b2/lib/awful/ewmh.lua#L92-L142), and therefore appears to be missing handling of minimized clients etc.
Is that your main (only?) issue (i.e. that minimized clients are not handled with the code from the FAQ)?

codebling:
> Is that your main (only?) issue (i.e. that minimized clients are not handled with the code from the FAQ)?

Yes! Updated the code.

codebling:
Any blockers to merging this? Should I have just created an issue instead?

actionless:
did you meant?

```lua
    function newactivate(c)
        if not c.minimized then
            c.minimized = true
        end
        if c:isvisible() then
            awful.ewmh.activate(c)
        end
    end
```

codebling:
> 
> (it behave differently in the case when you have minimized client on another tag)

Can you explain? Differently how?

actionless:
when you have minimized client on another tag

1) it will check if it minimized
2) will unminize it
3) check if it's visible
4) it's not visible -- no ewmh.activate fired

codebling:
> (it behave differently in the case when you have minimized client on another tag)
> 

Basically, I'm not sure this is the best code, just the smallest change to the code that was there that also fixes the problem. In the end, I don't really care what happens, I'd just like to close this PR out, whether it is merged or not. 

The issue is that it doesn't the code in the FAQ prevents minimised windows from being restored. Maybe I should have just opened an issue, but I thought that an easy PR is usually nicer than a low-priority issue.

codebling:
> when you have minimized client on another tag
> 
> 1. it will check if it minimized
> 2. will unminize it
> 3. check if it's visible
> 4. it's not visible -- no ewmh.activate fired

I understand what it does, but I don't understand why you want to unminimise it instead of letting ewmh.activate do that. 

with this code 
```
    if c:isvisible() or c.minimized then
        awful.ewmh.activate(c)
```
The end result is the same, isn't it?

codebling:
> when you have minimized client on another tag
> 
> 1. it will check if it minimized
> 2. will unminize it
> 3. check if it's visible
> 4. it's not visible -- no ewmh.activate fired

I think I see what you mean, if the window is minimised on a tag that is not visible, my tag will still activate it, while yours will not? 

In that case, maybe we should re-minimise the window if it was not visible? 
```
  function newactivate(c)
        wasMinimized = false
        if c.minimized then
            c.minimized = false
            wasMinimized = true
        end
        if c:isvisible() then
            awful.ewmh.activate(c)
        else 
            c.minimized = true
        end
    end
```

But this seems rather complicated. Can newly-opened windows be minimised? Is that even possible?

actionless:
some apps are restoring their minimized state on startup (if it was minimized on exit)

imagine you have such app awful.rule-d to be started on tag `2`

when you on tag `1` and start that app:

a) in case if its last state was not minimized it will work as expected
b) but in case if its last state was minimized it will still get `urgent` by awful.ewmh.activate without checking for visibility

actionless:
> some apps are restoring their minimized state on startup 

at least firefox does that

actionless:
> wasMinimized = false
> wasMinimized = true

i see you setting the var but it's not being used further in code

codebling:
> > wasMinimized = false
> > wasMinimized = true
> 
> i see you setting the var but it's not being used further in code

Sorry, it's late. 

```lua
  function newactivate(c)
        wasMinimized = false
        if c.minimized then
            c.minimized = false
            wasMinimized = true
        end
        if c:isvisible() then
            awful.ewmh.activate(c)
        else 
            if wasMinimized then
                c.minimized = true
            end
        end
    end
```

actionless:
what about this?

```lua
    function newactivate(c)
        awful.ewmh.activate(c)
        if not c:isvisible() then
          c.urgent = false
        end
    end
```

codebling:
I definitely did not think this would be so complicated. The goal was just to make things better with a quick small contribution, not to create headaches! 

> ```lua
>     function newactivate(c)
>         awful.ewmh.activate(c)
>         if not c:isvisible() then
>           c.urgent = false
>         end
>     end
> ```

Yes, that looks great! I'll try it and update the PR tomorrow

codebling:
> ```lua
>     function newactivate(c)
>         awful.ewmh.activate(c)
>         if not c:isvisible() then
>           c.urgent = false
>         end
>     end
> ```

So this also broke restoring minimised windows. Turns out the signature for `activate` changed in v3.5.7, and needs the context and hints to work. 

```lua
client.disconnect_signal("request::activate", awful.ewmh.activate)
function newactivate(c, context, hints)
    awful.ewmh.activate(c, context, hints)
    if not c:isvisible() then
        c.urgent = false
    end
end
client.connect_signal("request::activate", newactivate)
```

Do we need `if not c:isvisible()` ?  Can we just disable urgent for all windows?

actionless:
(squash before merging)

codebling:
Travis build failed but it's not clear why.

actionless (dismissed):

Aire-One:
What's the status of this PR? Especially regarding the recent merge of #2941 with commit 5818de41ce181f36207bb0d7b090aff292737089 renaming `awful.ewmh` to `awful.permissions`.

(off topic: BTW shouldn't this change the features list from https://awesomewm.org/ where it's said `Implements many Freedesktop standards: EWMH, [...]`?)

Elv13:
> What's the status of this PR?

There will be another pull request after `ruled.client` gets merged to add permission granting/denying to the rules. This will be a much nicer way to address this issue.

PR #2901, opened 2019-10-10
feat(awesome: main: check): print config path, and do it to stdout

actionless:
(follow-up on @blueyed comment here: https://github.com/awesomeWM/awesome/pull/2687#issuecomment-540129901)

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -619,13 +619,14 @@ main(int argc, char **argv)
         bool success = true;
         /* Get the first config that will be tried */
         const char *config = luaA_find_config(&xdg, confpath, true_config_callback);
+        fprintf(stdout, "Checking config '%s'... ", config);
 
         /* Try to parse it */
         lua_State *L = luaL_newstate();
         if(luaL_loadfile(L, config))
         {
             const char *err = lua_tostring(L, -1);
-            fprintf(stderr, "%s\n", err);
+            fprintf(stdout, "\nERROR: %s\n", err);
             success = false;
         }
         p_delete(&config);
@@ -633,12 +634,11 @@ main(int argc, char **argv)
 
         if(!success)
         {
-            fprintf(stderr, "✘ Configuration file syntax error.\n");
             return EXIT_FAILURE;
         }
         else
         {
-            fprintf(stderr, "✔ Configuration file syntax OK.\n");
+            fprintf(stdout, "OK\n");
             return EXIT_SUCCESS;
         }
     }


awesome.c
```diff
@@ -700,12 +701,12 @@ main(int argc, char **argv)
 
         if(!success)
         {
-            fprintf(stderr, "✘ Configuration file syntax error.\n");
+            fprintf(stdout, "ERROR\n");
```
blueyed: I think "ERROR: " could be used as a prefix with the error above already.
Then it would only return here.

actionless: could you insert is as a "Suggestion" here?
i'm not sure i'm following

blueyed: Remove the line above, and add "ERROR: " in front of `fprintf(stderr, "%s\n", err);` (I cannot make a suggestion there).
=> `fprintf(stderr, "ERROR: %s\n", err);`

actionless: do you mean like this?

psychon:
I just tried a random non-Lua file with the current code:
```
$ awesome -c  /tmp/b -k
/tmp/b:1: syntax error near 'MIN'
✘ Configuration file syntax error.
```
So, after this PR, the output would be... uhm, I guess this:
```
$ awesome -c  /tmp/b -k
Checking config /tmp/b... ERROR: /tmp/b:1: syntax error near 'MIN'
```
Right?

And for a working file, I would get:
```
Checking config /tmp/b... OK
```
Fine with me, It's just so that we know what is going on.

blueyed:
Yes.
Maybe the "ERROR: …" could be on a new line (if you want to parse it etc)?

actionless:
updated

awesome.c
```diff
@@ -686,6 +686,7 @@ main(int argc, char **argv)
         bool success = true;
         /* Get the first config that will be tried */
         const char *config = luaA_find_config(&xdg, confpath, true_config_callback);
+        fprintf(stdout, "Checking config %s... ", config);
```
psychon:
```suggestion
        fprintf(stdout, "Checking config '%s'... ", config);
```
(No idea if this makes sense or not)

Elv13 (dismissed):

psychon (dismissed):

actionless:
this one will require manual merge as too much old commits are failing in "test prev commits" travis scenario

actionless:
bump

Aire-One (approved):

actionless:
this time, codecov stucked :facepalm:

sclu1034 (approved):

Elv13 (approved):

Issue #2903, opened 2019-10-11
Client Icons not returning value


pyrocix:
Output of `awesome --version`:
awesome v4.3-491-g3340a7dc-dirty (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

**How to reproduce the issue:**
Updating Awesome v4.3 (Too long) to awesome v4.3-491-g3340a7dc-dirty
I used this block of code to add an image to an imagebox to generate an icon of each clients inside the task list:

    if icon then
      ib.image = icon
    else
      ibm:set_margins(0)
    end

But after switching to git version of awesome v4.3-491-g3340a7dc-dirty, it does not work anymore. I tried fixing this using the documentation but I have no luck. Switching back to the stable version fixes the problem  but I want to stay maintain my setup for the incoming new versions of awesome. 

**Actual result:**

![actual](https://user-images.githubusercontent.com/56428807/66635046-f7257500-ec40-11e9-9651-b2a1c860dafd.png)

**Expected result:**

![expected](https://user-images.githubusercontent.com/56428807/66635017-ea088600-ec40-11e9-8fc4-c7e947db2118.png)

psychon:
> I used this block of code

What exactly did you do with this block of code? I have some trouble understanding this, sorry. Could you provide step by step instructions on how to reproduce your issue?

Aire-One:
I can confirm this bug.

I just stared Xephyr with my work-in-progress-custom-taglist-widget (it uses the "main" client icon as tag icon) and I had the same issue. I didn't work on this widget since a long time, so I can't say where this regression is from.

Using `gears.debug.dump` show that `client.icon` is now an "userdata". IIRC, it used to be a cairo surface before. Also, `wibox.widget.imagebox:set_image()` doesn't handle this and return `false` so obviously it failed.

---

 As a side note:

Working with `client.icon` has been a bad idea since the beginning tho. [As the documentation states](https://awesomewm.org/apidoc/core_components/client.html#icon), you (and I) shouldn't work with it.

Instead, there is the `awful.widget.clienticon` widget. Another way to work with `client.icon` is to duplicate it. I tried the code from the doc and I can attest it's working.

psychon:
[...]

PR #2905, opened 2019-10-11
imagebox:set_widget(): Handle userdata again

psychon:
This function is not documented to handle userdata, but historically it
would just pass all arguments to gears.surface.load(), which interpreted
userdata as cairo surfaces. After 3295e9f33d330678384c84031, userdata
objects are restricted.

Accept userdata objects again by explicitly passing them to
gears.surface.load().

Fixes: https://github.com/awesomeWM/awesome/issues/2903


psychon:
I did no testing at all on this. Someone should test that this really fixes #2903.

PR #2910, opened 2019-10-15
fix ewmh.merge_maximization

xinhaoyuan:
This should fix the error reported on [reddit](https://www.reddit.com/r/awesomewm/comments/dghuk5/window_csd_restore_not_functioning_properly/), or the version that I tried with wmctrl.

### DIFF
--- a/lib/awful/ewmh.lua
+++ b/lib/awful/ewmh.lua
@@ -353,14 +353,37 @@ function ewmh.merge_maximization(c, context, hints)
 
     if not c._delay_maximization then
         c._delay_maximization = function()
-            -- This ignores unlikely corner cases like mismatching toggles.
-            -- That's likely to be an accident anyway.
-            if c._delayed_max_h and c._delayed_max_v then
-                c.maximized = c._delayed_max_h or c._delayed_max_v
-            elseif c._delayed_max_h then
-                c.maximized_horizontal = c._delayed_max_h
-            elseif c._delayed_max_v then
-                c.maximized_vertical = c._delayed_max_v
+            -- Computes the actual X11 atoms before/after
+            local before_max_h = c.maximized or c.maximized_horizontal
+            local before_max_v = c.maximized or c.maximized_vertical
+            local after_max_h, after_max_v
+            if c._delayed_max_h ~= nil then
+                after_max_h = c._delayed_max_h
+            else
+                after_max_h = before_max_h
+            end
+            if c._delayed_max_v ~= nil then
+                after_max_v = c._delayed_max_v
+            else
+                after_max_v = before_max_v
+            end
+            -- Interprets the client's intention based on the client's view
+            if after_max_h and after_max_v then
+                c.maximized = true
+            elseif before_max_h and before_max_v then
+                -- At this point, c.maximized must be true, and the client is
+                -- trying to unmaximize the window, and potentially partial
+                -- maximized the window
+                c.maximized = false
+                if after_max_h ~= after_max_v then
+                    c.maximized_horizontal = after_max_h
+                    c.maximized_vertical = after_max_v
+                end
+            else
+                -- At this point, c.maximized must be false, and the client is
+                -- not trying to fully maximize the window
+                c.maximized_horizontal = after_max_h
+                c.maximized_vertical = after_max_v
             end
         end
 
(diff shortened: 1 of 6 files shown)


xinhaoyuan:
Hmm ... the test error here looks interesting. It actually implies there could be a client state where `maximized and not maximized_vertical and not maximized_horizontal` holds. I think this implies a bigger problem for the design of `ewmh.merge_maximization`. When `maximized and not maximized_vertical and not maximized_horizontal` holds, there is no way for client to toggle back to normal window (ending up not changing the client state)?

Elv13:
Thanks for this, I will review it soon. It looks alright at first sight. Don't worry about the `test-placement` failure, it's flaky since a few weeks for uninvestigated reasons. Please put the fix commit before the test commit to get the CI to pass. We do it in that order so we can use `git-bisect` and have as little commit as possible to fail the tests. Getting the test coverage higher would also be appreciated, but that may require too much work, so feel free to ignore this.

tests/test-maximize.lua
```diff
@@ -232,6 +232,30 @@ gears.table.merge(steps, {
...
     end,
+    -- Maximize then unmaximize. The window state should be restored.
+    function()
+        if #client.get() > 0 then return end
+
+        test_client(nil,nil,nil,nil,nil,{maximize_before=true,unmaximize_after=true})
+
+        return true
+    end,
+    function()
```
Elv13: This needs a `if #client.get() ~= 1 then return end` to avoid flackyness when the client takes more time to start for whatever reasons.

xinhaoyuan: I think `if not c then return end` will catch the flakiness, since #client.get() could only be 0 or 1 at this point.

xinhaoyuan:
For the coverage, I still do not get why a change like this would impact the coverage of unrelated modules, such as `lib/awful/client/shape.lua` ...

tests/test-maximize.lua
xinhaoyuan: In other words, if `#client.get() ~= 1`, c would be `nil`, thus will this step will return and be retried.

Elv13:
> For the coverage, I still do not get why a change like this would impact the coverage of unrelated modules, such as lib/awful/client/shape.lua

Ignore them if they are fully unrelated. I should have said that in the last post, but I was in a hurry. I need to look into that. tl;dr, there is some "random" [X11] events that occur during the CI run and since a month or so have gotten a lot worst. Some tests that never failed before started failing randomly and the coverage flackyness got worst. My theory is that Travis stated to overprovision their nodes more aggressively and made some previously "mostly" predictable sequences unpredictable. Since they used to be predictable, nobody coded their test with those corner case in mind and that just blew up.

tests/test-maximize.lua
Elv13: right, sorry for the noise.

lib/awful/ewmh.lua
```diff
@@ -355,11 +355,11 @@ function ewmh.merge_maximization(c, context, hints)
         c._delay_maximization = function()
             -- This ignores unlikely corner cases like mismatching toggles.
             -- That's likely to be an accident anyway.
-            if c._delayed_max_h and c._delayed_max_v then
+            if c._delayed_max_h ~= nil and c._delayed_max_v ~= nil then
```
Elv13: Umm, looking at this again, wont this set `c.maximized` to true even if one of them is false? That's not the intended behavior.

xinhaoyuan: Agreed. I don't have a perfect solution in mind. Before this patch it is not possible to set maximized to false in this if branch. Maybe changing the assignment to `c.maximized = ... and ...` would make more sense - what do you think?

Elv13: Maybe a variant of:

```
if c._delayed_max_h == true and c._delayed_max_v == true then
    c.maximized = true
elseif c._delayed_max_h == false and c._delayed_max_v == false then
    c.maximized = false
elseif ...
```

?

Because otherwise it cannot be merged. This cause a rather grave regression for every programs for the sake of fixing something that only happens with CSD apps, aka the < 100 Gnome3 apps that use it.

xinhaoyuan: I refactor the logic here to interpret the client's intention, based on the client's view of how the X11 atoms would change -- please let me know what do you think.

xinhaoyuan:
No problem -- I'm happy to contribute to it as it is my daily driver :), and I understand the intention to keep the project healthy. Please take another look at my changes to make wmctrl optional.

Thanks for maintaining this amazing project. :+1:

PR #2916, opened 2019-10-28
Fix tag deletion not emitting signals (#2914)

### LINKED ISSUE #2914: Deleting a tag does not always refresh the taglist
```
awesome v4.3-505-g0297bfff9-dirty (Too long)
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

Add keybinding to delete tags, e.g. in the default config I made `mod+#` delete the tag by index rather than view it.
Delete tag 9. It will still be displayed in the taglist. Open a client and tag 9 will disappear.
This only occurs if the last tag is deleted.

**Actual result:**
Tag 9 is displayed in the taglist after being deleted.

**Expected result:**
Tag 9 is removed from the taglist when deleted.

jcrd:
Fixes #2914.
Thanks @psychon.

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -402,7 +402,7 @@ function tag.object.delete(self, fallback_tag, force)
     end
 
     -- delete the tag
-    self.data.awful_tag_properties.screen = nil
+    self.screen = nil
     self.activated = false
 
     -- Update all indexes


psychon:
`test-awful-tag.lua` broke:
https://github.com/awesomeWM/awesome/blob/0297bfff9ad88535fd9302fdb7d9b11459d6b1b4/tests/test-awful-tag.lua#L215-L217
The `assert(t.screen == nil)` specifically fired.

Sigh, this needs investigation. Sorry!

psychon:
Found the reason for the failure:
https://github.com/awesomeWM/awesome/blob/0297bfff9ad88535fd9302fdb7d9b11459d6b1b4/lib/awful/tag.lua#L568-L570
`t.screen = nil` moves a tag to the "focused" screen. Sigh.

psychon:
This version works:
```diff
diff --git a/lib/awful/tag.lua b/lib/awful/tag.lua
index 8874d69c7..daf5d9020 100644
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -388,7 +388,7 @@ function tag.object.delete(self, fallback_tag, force)
     end
 
     -- delete the tag
-    self.data.awful_tag_properties.screen = nil
+    tag.setproperty(self, "screen", nil)
     self.activated = false
 
     -- Update all indexes
```
The old code bypasses "everything" and directly modifies the data table. The patch above only bypasses `awful.tag.object.set_screen` and uses the property setting method that also emits a signal. Better.

jcrd:
With this change, I'm experiencing the same problem as reported in the original issue. The taglist will be refreshed if any tag but the last is deleted.

psychon:
Ah, sorry. I forgot about the original issue and now... I wonder why only the last tag causes this problem.
Random idea: This PR is about not emitting `property::screen` when a tag is deleted and its screen property changes to `nil`.
A follow-up PR would make the taglist update more aggressively:
```diff
diff --git a/lib/awful/widget/taglist.lua b/lib/awful/widget/taglist.lua
index 4b39d4e9c..9a14556c9 100644
--- a/lib/awful/widget/taglist.lua
+++ b/lib/awful/widget/taglist.lua
@@ -528,6 +528,13 @@ function taglist.new(args, filter, buttons, style, update_function, base_widget)
                 for _, tlist in pairs(i) do
                     tlist._do_taglist_update()
                 end
+            else
+                -- No screen? Update all taglists
+                for _, list in pairs(instances) do
+                    for _, tlist in pairs(list) do
+                        tlist._do_taglist_update()
+                    end
+                end
             end
         end
         local uc = function (c) return u(c.screen) end
```
If someone has a better idea, please speak up.

@Elv13 does all of this need some new tests?

Elv13:
> @Elv13 does all of this need some new tests?

Thanks for investigating. It would be non-trivial to test for little benefits, so I guess it is fine.

jcrd:
This fixes the issue in my manual testing. Sorry for all the back and forth. Thanks again @psychon.
This PR has been superseded by #2920.

PR #2918, opened 2019-11-15
fix: bad argument to apply_size_hints

0x501D:
While in floating mode user can resize window to a zero width or height size.
If this happens an error occures: bad argument 1 to apply_size_hints (value in [1, 65535] expected, got 0).
After this error window size cannot be changed.

### DIFF
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -1137,7 +1137,7 @@ function placement.resize_to_mouse(d, args)
     remove_border(d, args, ngeo)
 
     -- Now, correct the geometry by the given size_hints offset
-    if d.apply_size_hints then
+    if d.apply_size_hints and ngeo.width ~= 0 and ngeo.height ~= 0 then
         local w, h = d:apply_size_hints(
             ngeo.width,
             ngeo.height


psychon:
I guess this needs a test, but... how exactly does the patch work? This just skips size hints and alignment for when a window is resized to width or height zero? Shouldn't that still cause errors/issues later on?

0x501D:
/*This just skips size hints and alignment for when a window is resized to width or height zero?*/
Exactly.
/*Shouldn't that still cause errors/issues later on?*/
Unfortunately I cannot deep dig in architecture of awesome VM but on my local machine this patch fix the problem: an error goes away and window size can be changed back after window was resized to zero length (both direction was tested)

lib/awful/placement.lua
```diff
@@ -1137,7 +1137,7 @@ function placement.resize_to_mouse(d, args)
     remove_border(d, args, ngeo)
 
     -- Now, correct the geometry by the given size_hints offset
-    if d.apply_size_hints then
+    if d.apply_size_hints and ngeo.width ~= 0 and ngeo.height ~= 0 then
         local w, h = d:apply_size_hints(
             ngeo.width,
```
Elv13: @psychon  what about:

    ngeo.width == 0 and 1 or ngeo.width,
    ngeo.height == 0 and 1 or ngeo.height,

This would still cause issues, but less than skipping the size hints? Fixing the floating code to stop at `1x1` would fix the other half of the problem. I have a PR which touches that code (moves code around to remove ~4 circular `require()`), I will add a commit for it and see if I can submit it soon.

psychon: Makes me wonder if the C code handles this correctly. Let's say I have a window with an aspect ratio of 1:2 and I call `c:apply_size_hints(1, 1)`. I would expect the C code to return `0, 1` (or `1,0`?), which would then still cause the issue / bug that is described here...

PR #2921, opened 2019-11-22
[Draft/Proposal] DOC - Fix modules summaries.

Aire-One:
Hello!

I was bored by the documentation index being broken by modules missing the final dot `.` symbol at the end of their summary.

I started this PR to fix it. I then realized summaries are sometime not clear, misleading and even missing for some modules... So I took `wibox.widget.imagebox` and tried to make its description better.

Here is a listing of the things I changed in the `wibox.widget.imagebox` documentation:

* Add a short description after the summary ;
* Show both procedural and declarative pattern to create an imagebox widget : 
    * Change `tests/examples/wibox/widget/defaults/imagebox.lua` to procedural definition of imagebox ;
    * Add a `tests/examples/wibox/widget/declarative-pattern/imagebox.lua` file to show off the declarative pattern to create an imagebox ;
* Update ldoc tags to only use typed version (`@tparam` and `@treturn`) ;
* Separate documentation for object properties and methods to not mix their summaries / definitions (looks like ldoc manage them better like this)

What I would like is to add an "object hierarchy section" to make it easier to understand `imagebox` is a derived of `wibox.widget` which derives `gears.object`. (We could also mark derived methods to better understand what comes from where)

If you're ok with what I did for `imagebox`, I'll try to port these changes to every widgets (first from the `wibox` library then `awful.widget`).


lib/wibox/widget/imagebox.lua
```diff
@@ -1,6 +1,16 @@
 ---------------------------------------------------------------------------
+-- A widget to render image(s).
+--
+-- The `wibox.widget.imagebox` is part of the Awesome WM's wiboxes system.
+-- It renders image in a box widget.
```
psychon: Should this be `images`?
I am not sure if the `box`-part of the sentence adds much meaning. I guess it is there just to explain the name `imagebox`?
How about `This widget displays an image.`?

(Feel free to just ignore this.

lib/wibox/widget/imagebox.lua
```diff
@@ -135,11 +135,24 @@ local function load_and_apply(ib, file, image_loader, image_setter)
...
 end
 
---- Set an imagebox' image
+--- The image rendered by the `imagebox`.
+-- <br />It can can be any of the following:
+--
+-- * A `string` : Interpreted as the path to an image file ;
+-- * A cairo image surface : Directly used as is ;
+-- * An rsvg handle object : Directly used as is ;
+-- * `nil` : Deny previously setted image.
```
psychon: I have basically no clue about ldoc, so: Is the `<br />` necessary? Are the `;` necessary?

psychon: `Deny previously setted image` sounds weird to me. How about `Unset image`? `Remove image`? Dunno...

lib/wibox/widget/imagebox.lua
```diff
@@ -206,21 +219,25 @@ end
...
 
---- Returns a new imagebox.
--- Any other arguments will be passed to the clip shape function
--- @param image the image to display, may be nil
--- @param resize_allowed If false, the image will be clipped, else it will be resized
---   to fit into the available space.
--- @param clip_shape A `gears.shape` compatible function
--- @treturn table A new `imagebox`
+--- Returns a new `wibox.widget.imagebox` instance.
+-- Any additional arguments will be passed to the clip shape function.
```
psychon: (I know you are not adding this, but) what the heck does `any additional arguments will be passed to the clip shape function` mean? I do not see anything like this in the code.

lib/wibox/widget/imagebox.lua
```diff
@@ -206,21 +219,25 @@ end
 
 --- Should the image be resized to fit into the available space?
 -- @property resize
--- @param allowed If false, the image will be clipped, else it will be resized
---   to fit into the available space.
+-- @tparam boolean resize
+
+--- Should the image be resized to fit into the available space?
+-- @tparam boolean allowed If `false`, the image will be clipped, else it will
+--   be resized to fit into the available space.
```
psychon: I feel like "the image will be clipped" is slightly misleading. Yes, this is what happens when the image is larger than the available space. However, `resize = false` also has an effect when the image is smaller than the available space (the image is simply displayed as-is). No idea how to make this clearer...

(This also appears in the description of the constructor.)

Aire-One: I think we can have an HTML table showing 3 use cases (image smaller than the widget, image size == widget size, and image bigger than the widget) for both values of the `resize` property.

But I'll need help on that. I'm not sure how this kind of tables are defined in other parts of the documentation.

(About the constructor, I think we need some changes on its implementation. When a property isn't explicitly given to it, the constructor will not set it at all! So `local ib = imagebox(beatiful.icon)` will result on `ib.resize == nil`. I didn't dig too much on that but I guess we can have errors depending on how it's used)

lib/wibox/widget/imagebox.lua
Aire-One: It was the words used before. You're right, tho. I'll change that

Aire-One: The `<br />` is basically used as an HTML tag to insert a carriage-return character. I used it to separate the summary from the body description of the method. Summary only is shown in the "methods table" at the beginning of the doc. Then in the detail of the method you have Summary+Body. I wanted them to be line separated here so I added this `<br />`.

It could also have be done with an empty line:

```lua
--- The image rendered by the `imagebox`.
-- 
-- It can can be any of the following:
```

The `;` is more an habit I have when doing lists. According to https://getitwriteonline.com/articles/vertical-lists/ I should rather use `,` or `.`, tho.

lib/wibox/widget/imagebox.lua
Aire-One: Yeah... Summary/Description still need to be improved. I need to come back on this later. I lake inspiration for now...

> I am not sure if the `box`-part of the sentence adds much meaning. I guess it is there just to explain the name `imagebox`?

You're right. I tried to use the work "box" because it's in the name. Obviously it was a bad idea.

> How about `This widget displays an image.`?

Simple, easy, clear. I actually like it! However, I have a preference for the "render" word because you can resize/use a cairo surface for the image. So render seems more appropriate to me.

Aire-One:
https://github.com/awesomeWM/awesome/pull/2921#discussion_r349626333

> (I know you are not adding this, but) what the heck does `any additional arguments will be passed to the clip shape function` mean? I do not see anything like this in the code.

I think we can fix that with the following patch (BTW `imagebox:set_clip_shape` work like this):

```patch
diff --git a/lib/wibox/widget/imagebox.lua b/lib/wibox/widget/imagebox.lua
index 2426f401..1f7f25f5 100644
--- a/lib/wibox/widget/imagebox.lua
+++ b/lib/wibox/widget/imagebox.lua
@@ -261,7 +261,7 @@ end
 -- @tparam[opt] function clip_shape A `gears.shape` compatible function.
 -- @treturn wibox.widget.imagebox A new `wibox.widget.imagebox` widget instance.
 -- @constructorfct wibox.widget.imagebox
-local function new(image, resize_allowed, clip_shape)
+local function new(image, resize_allowed, clip_shape, ...)
     local ret = base.make_widget(nil, nil, {enable_properties = true})
 
     gtable.crush(ret, imagebox, true)
@@ -274,7 +274,7 @@ local function new(image, resize_allowed, clip_shape)
     end
 
     ret._private.clip_shape = clip_shape
-    ret._private.clip_args = {}
+    ret._private.clip_args = {...}
 
     return ret
 end
```

Aire-One:
The last commit (a8cd496) will be a controversial one I want to work on. This was a journey for me in the ldoc configuration and documentation template and the work is not finished yet but I really ask for your feedback on this commit before going any deeper in this. :)

Basically, it changes how inherited members are rendered in the documentation. I added new tags `@inheritedproperty`, `@inheritedmethod` and `@inheritedsignal` to respectively replace `@property`, `@method` and `@signal` tags in files included in objects documentation like `docs/common/object.ldoc` and `docs/common/widget.ldoc`.

Because I target `wibox.widget` for now (and use `wibox.widget.imagebox` as my sandbox), this commit is limited to this for now. There are most probably a lot of corner cases I didn't managed here.

Some issue I have with the current commit:

* I can find an elegant way to find the inheritance provenance for signals.
* I can't find a way to link the inheritance provenance to the parent object ("Inherited from ..." are broken links).
* The `wibox.widget:index` method from `docs/common/widget.ldoc` never makes its way to the rendered documentation for some reason.
* Inherited elements can appear as duplicated, example with `wibox.layout.align` where the `children` property appear in both, object properties and inherited object properties (because `wibox.layout.align` redefines/overrides the `children` property).

Overall, I'm pretty proud of the result I have even if it's not yet fully working. Here is a screenshot of the page to help you have an idea without compiling my branch:

Screenshot_2019-11-23 Widgetmod wibox widget imagebox - awesome API documentation

Elv13:
> @Elv13 Any idea what happened to Travis?

Nope, it has been like since since late July. I tried to debug it a few time, but it's hard to reproduce. I also could not reproduce it locally. Some of my recent PRs were just restarted ~10 times until they passed clean. This along with the flacky event.c/property.c has is annoying. I don't have much time right now and the time I have, I try to invest into finishing the large screen rules endeavor, which pulls a lot of strings.

Elv13:
> I can find an elegant way to find the inheritance provenance for signals.

I think it is possible to add new `@foo` properties. LDoc intends this to be used to "invisible purpose" like `@todo` and `@warning`. However I think we can access the result from the template, so we could use them for this purpose.

I know @Veratil started to work on a second template to generate machine readable "documentation" so we can data mine the API and meta-generate more doc. When I really get time to work on Awesome again, I hope we can leverage this to generate all the missing property signal information along with their arguments and unit test them all. Right now it is nearly impossible to know if a property have a signal without testing it.

Veratil:
> I know @Veratil started to work on a second template to generate machine readable "documentation" so we can data mine the API and meta-generate more doc. When I really get time to work on Awesome again, I hope we can leverage this to generate all the missing property signal information along with their arguments and unit test them all. Right now it is nearly impossible to know if a property have a signal without testing it.

I forget if I shared the file with you or not. I had that on my work laptop which I had to give up recently, but I saved my home dir stuff which should have the updated file just in case I didn't save it in my fork.

Aire-One:
Because @Elv13 is ready to merge #2924, I'll wait it goes to master to get back on this PR. I'll remove my last commit and use the new tags.

Elv13:
@Aire-One Can you un-draft this, rebase it and many address some @psychon comments? I cannot press the merge button right now (its disabled).

lib/wibox/widget/imagebox.lua
Aire-One: I added an image example: `tests/examples/wibox/widget/imagebox/resize.lua`

imagebox_resize

Aire-One:
I addressed all the points @psychon came with.

I also removed my commit about inherited stuff in favor of what @Elv13 did lastly.

I'm not sure how to remove the draft status, tho. It looks like github already removed this status automatically... Not sure what's going on...

actionless:
please also rebase to make the travis pass

Aire-One:
> please also rebase to make the travis pass

Done.

Issue #2919, opened 2019-11-16
Improvement on generated images for the screen API documentation


Aire-One:
I wasn't in the mood of creating a PR for this. So here is an issue so we can discuss of what I did :)

You can find my changes at https://github.com/Aire-One/awesome/tree/doc_screen_images.

I were looking at the documentation to find what I can do to make it more easy to read and I found these two "issues" on the generated images for the screen API which are fairly easy to fix.

I'd like your feedback :)

ping @Elv13 :)

# Rulers order

![old-rulers-order](https://user-images.githubusercontent.com/6602958/68998096-181b5e80-08ae-11ea-90dd-a3b9d8d2d085.png)

It just don't seems natural the `geometry` rulers is in the middle. The bubble sort algorithm order rulers correctly but when they have the same size. If rulers have the same size, they keep the order they were inserted into the rulers list. So rulers insertion order impact the result of the bubble sort algorithm.

By inverting the order, we can make the `geometry` ruler be always the most outer one. The generated image looks more natural.

![rulers-order](https://user-images.githubusercontent.com/6602958/68998383-c8d72d00-08b1-11ea-9490-452e0679e992.png)

# Wibar

Generated screen images always have a little margin at the top of the `workarea` rect. This space is caused by the wibar which is at the top of the screen in the default `aweseomerc.lua` and in the code to generate images:

```lua
-- This will shift the workarea by 24px at the top.
awful.wibar {
    position = "top",
    height   = 24,
}
```

It's not a big deal once you know it. However, I think there should be an indication of this mechanism. I added a `draw_wibar` parameter to the template generator to draw it in the generated image:

![show-wibar](https://user-images.githubusercontent.com/6602958/68998249-64679e00-08b0-11ea-89cf-56b84df286cf.png)

I think this still need additional work to be drawn differently than area with the stripe pattern, tho..

Elv13:
Thanks for this. Please make a PR when its ready.

This was my vision for this (see the header comment in https://github.com/awesomeWM/awesome/pull/2855). I was the one too lazy to do it :P. Now its done! Just make sure to also take into account clients with their own `:struts()`. They are the "bars"/"docked" clients. AwesomeWM doesn't support 3rd party bars and docked client very well, but it *tries*.

[...]

PR #2922, opened 2019-11-22
Documentation `screen` - Improvement on generated images.

Aire-One:
This PR follows my issue #2919.

# Changes from the draft I wrote for the issue

I changed the wibar drawing to make it a solid surface instead of using the stripe pattern and force pushed to my branch.

Here is a preview:

generated-wibar

# Original message from the issue:

> I were looking at the documentation to find what I can do to make it more easy to read and I found these two "issues" on the generated images for the screen API which are fairly easy to fix.
>
> I'd like your feedback :)
>
> ping @Elv13 :)
>
> # Rulers order
>
> old-rulers-order
>
> It just don't seems natural the `geometry` rulers is in the middle. The bubble sort algorithm order rulers correctly but when they have the same size. If rulers have the same size, they keep the order they were inserted into the rulers list. So rulers insertion order impact the result of the bubble sort algorithm.
> 
> By inverting the order, we can make the `geometry` ruler be always the most outer one. The generated image looks more natural.
> 
> rulers-order
> 
> # Wibar
> 
> Generated screen images always have a little margin at the top of the `workarea` rect. This space is caused by the wibar which is at the top of the screen in the default `aweseomerc.lua` and in the code to generate images:
> 
> ```lua
> -- This will shift the workarea by 24px at the top.
> awful.wibar {
>     position = "top",
>     height   = 24,
> }
> ```
> 
> It's not a big deal once you know it. However, I think there should be an indication of this mechanism. I added a `draw_wibar` parameter to the template generator to draw it in the generated image:

> show-wibar
> 
> I think this still need additional work to be drawn differently than area with the stripe pattern, tho..

### DIFF
--- a/tests/examples/screen/geometry.lua
+++ b/tests/examples/screen/geometry.lua
@@ -23,7 +23,7 @@ end --DOC_HIDE
     --DOC_NEWLINE
 
     -- This will shift the workarea by 24px at the top.
-    awful.wibar {
+    local wibar = awful.wibar {
         position = "top",
         height   = 24,
     }
@@ -32,4 +32,5 @@ return { --DOC_HIDE
     factor             = 2    , --DOC_HIDE
     show_boxes         = false, --DOC_HIDE
     highlight_geometry = true , --DOC_HIDE
+    draw_wibar         = wibar, --DOC_HIDE
 } --DOC_HIDE
--- a/tests/examples/screen/padding.lua
+++ b/tests/examples/screen/padding.lua
@@ -26,7 +26,7 @@ end --DOC_HIDE
         bottom = 20,
     }
 
-    awful.wibar { --DOC_HIDE
+    local wibar = awful.wibar { --DOC_HIDE
         position = "top", --DOC_HIDE
         height   = 24, --DOC_HIDE
     } --DOC_HIDE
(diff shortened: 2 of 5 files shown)


Aire-One:
Before this got merged, I'd like to finish the job with what @Elv13 [requested in the issue](https://github.com/awesomeWM/awesome/issues/2919#issuecomment-554673576):

> Just make sure to also take into account clients with their own `:struts()`. They are the "bars"/"docked" clients. AwesomeWM doesn't support 3rd party bars and docked client very well, but it _tries_.

I'd like to have more indications on this, tho:

* Should I only consider we want to be able to draw additional clients on generated images? So basically add a new argument `dockable_clients` to manage them in the template, just like I did for wiboxes.
* How could I test that? Can you point me examples in the documentation where it will be used?

Elv13:
Thanks a lot for this. I think we can already merge this as-is.


> Should I only consider we want to be able to draw additional clients on generated images? So basically add a new argument dockable_clients to manage them in the template, just like I did for wiboxes.

Clients and wibox can set their `:struts()` property to tell how much space they want to reserve. I am not sure it 100% works with the shims, but I think most of the code is there. You can create fake clients and wibox and set their `:struts()`.

> How could I test that? Can you point me examples in the documentation where it will be used?

The client and wibox `:struts()` property documentation (which is also copy/pasted into `wibar` and a few other places).

Aire-One:
NOOOOooooo

I was coming with additional commits to try to add clients support :'(

actionless:
IMHO it would be easier to review it in smaller chunks

Aire-One:
You're right. Better to have this merged while we work on the next part :)

I opened I new draft PR with my current code to have feedback from @Elv13.

PR #2923, opened 2019-11-23
Emit property::screen in tag:delete()

psychon:
Before this commit, the code directly modified the table where the tag's
properties were saved. This commit changes the code to call
awful.tag.setproperty() instead. This function ensures that
property::screen is now also emitted.

----

This is my patch from https://github.com/awesomeWM/awesome/pull/2916#issuecomment-551129210 / for #2914. I still think this should be merged.

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -402,7 +402,7 @@ function tag.object.delete(self, fallback_tag, force)
     end
 
     -- delete the tag
-    self._private.awful_tag_properties.screen = nil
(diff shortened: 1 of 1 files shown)


Elv13:
Umm, Travis seems to be on vacation...

actionless:
travis build failing for previous commit (so should be rebased or merged manually)

Elv13:
> travis build failing for previous commit (so should be rebased or merged manually)

Indeed, that was the problem. I rebased it.

PR #2924, opened 2019-11-24
Extend the documentation with new custom tags.

Elv13:
This is to allow @Aire-One to be unblocked with the inheritance support. There is now a new `@baseclass` tag we can define in the `@DOC_COMMON@` entries. It can then be accessed using `item.inherited` and `item.baseclass` in the template.

I also add a generic parser so we can add as many custom tags as we want. It re-implement some code which cannot be accessed from the template or the config. So now parts of ldoc are re-implemented in the ldoc config... However it adds some flexibility. An extended version of this will be able to auto-generate code examples for signal connections and other goodies by adding more custom attributes. It also allows custom tags to be rendered in tables rather than trees:

Screenshot_20191124_040455

@Aire-One If you think this is the way to go, I can merge this and then I will go ahead and update the entire API to use these (that will be painful...). It will slightly conflict with your PR.

I will also add some basic JavaScirpt to hide those new sections by default behind a "show more" summary. If I was to add this to everything and leave them visible, the signal to noise ratio would worsen.

edit:

```lua
--- The padding between the outline and the progressbar.
--@DOC_wibox_container_arcchart_paddings_EXAMPLE@
-- @property paddings
-- @tparam[opt=0] table|number paddings A number or a table
-- @tparam[opt=0] number paddings.top
-- @tparam[opt=0] number paddings.bottom
-- @tparam[opt=0] number paddings.left
-- @tparam[opt=0] number paddings.right
-- @emits [opt="optional_value"] property::paddings When the `paddings` changes.
-- @emitstparam property::paddings widget self The object being modified.
-- @emitstparam property::paddings table paddings The new paddings.
-- @usebeautiful beautiful.arcchart_paddings Fallback value when the object `paddings` isn't specified.

```

### DIFF
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -186,7 +186,7 @@ local function default_format(self, callback)
     end
 end
 
-local named_tags = {}
+local named_tags, item_id = {}, 1
 
 -- Add a new @something which can be used in any types.
 -- @tparam table args
@@ -467,6 +467,9 @@ end
 -- of content, which is before any custom types is used.
 local function init_custom_types(item)
     if item.is_init then return end
+
+    -- Give each item an unique identifier so the JavaScript can locate them.
+    item.uid, item_id = item_id, item_id + 1
 
     item.delims, item.auto_usage, item.auto_params = {}, {}, {}
 
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -262,6 +262,13 @@ add_custom_tag {
     }
 }
 
+-- Define the base class where a method/property is implemented.
+-- @baseclass my_module.my_submodule.my_baseclass
+add_custom_tag {
+    name   = "baseclass",
+    hidden = true
+}
+
 -- More fitting section names
 kind_names={topic='Documentation', module='Libraries', script='Sample files'}
 
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -306,6 +306,37 @@ add_custom_tag {
     }
 }
 
+-- For all properties which have a standard `@beautiful` variable for them
+-- @propbeautiful fallback1 fallback2 fallback3 fallback4
+add_custom_tag {
+    name   = "propbeautiful",
+    title  = "Consumed theme variables",
+    params = {
+        { name = "fallback1" },
+        { name = "fallback2" },
+        { name = "fallback3" },
+        { name = "fallback4" },
+    },
+    table  = {
+        "Variable", "Usage"
+    },
+    format = function(self, p, item, md)
+        local modname = item.module.name:gmatch("[^.]+$")()
+        local last = "beautiful."..(modname.."_"..item.name):gsub("[.]", "_")
+        local ret = "<td>"..md("`"..last.."`").."</td><td>Fallback when "..md("`"..item.name.."`")..
+            " isn't set.</td>"
+
+        for _, fallback in ipairs({p.fallback1, p.fallback2, p.fallback3, p.fallback4 }) do
+            ret = ret .. "</tr><tr>"..
+                "<td>"..md("`"..fallback.value.."`").."</td><td>Fallback when "..md("`"..last.."`")..
+            " isn't set.</td>"
+            last = fallback.value
+        end
+
+        return ret
+    end
+}
+
 -- Define the base class where a method/property is implemented.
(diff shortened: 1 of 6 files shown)


docs/config.ld
```diff
@@ -93,6 +93,182 @@ new_type("filterfunction", "List filters", false)
...
+            auto_subtags = false,
+        }
+    end
+end
+
+-- When a type will request a permission.
+-- @emits class signal Message[...]
+add_custom_tag {
+    name   = "emits",
+    title  = "Emit signals:",
```
Aire-One: The `:` is already added by the template. (BTW I think they are mostly useless IMO considering they are `<h3>` so they seen as titles).

```suggestion
    title  = "Emit signals",
```

docs/config.ld
```diff
@@ -93,6 +93,182 @@ new_type("filterfunction", "List filters", false)
...
+        return ret .. "<td>"..md(params.description).."</td>"
+    else
+        if params.name then
+            ret = '<span class="parameter">'..
+                md("`"..params.name.value.."`")..
+            "</span> "
+        end
+
+        if params.type then
+            ret = ret .. '<span class="parameter">( '..
```
Aire-One:
```suggestion
            ret = ret .. '<span class="parameter">('..
```

docs/config.ld
```diff
@@ -93,6 +93,182 @@ new_type("filterfunction", "List filters", false)
...
+                params.type.value..
+            " )</span>"
```
Aire-One:
```suggestion
            ")</span>"
```

docs/config.ld
```diff
@@ -93,6 +93,182 @@ new_type("filterfunction", "List filters", false)
...
+    else
+        if params.name then
+            ret = '<span class="parameter">'..
+                md("`"..params.name.value.."`")..
+            "</span> "
+        end
+
+        if params.type then
+            ret = ret .. '<span class="parameter">( '..
+                params.type.value..
```
Aire-One:
```suggestion
                md("`"..params.type.value.."`")
```

Aire-One:
I PR-ed your PR to add the missing part to match my inheritance documentation goal on #2921. 

Thanks again for your work!

> then I will go ahead and update the entire API to use these (that will be painful...).

I'll use these new tags while updating the widgets documentation in my PR. So you can remove them from your workload.

docs/config.ld
Elv13: done

docs/config.ld
Elv13: done

docs/config.ld
Elv13: done

docs/config.ld
Elv13: done

Elv13:
Rebased with your suggested changes and some of my local commits. I added 2 new tags which do exactly as the `@emits` and `@usebeautiful`, but auto-generate the entire boilerplate for standard properties. Given I am pushing for those standards in other PRs, it makes sense to leverage them here. The code is pretty meta, so it doesn't add too much (code) weight beside making the build noticeably slower.

I didn't add the commits which add the "tag cloud" next to the item name. I am still evaluating different designs (and I suck at modern CSS, my last webdev work was when PHP4 was the cool kid favorite tech).

Thanks for working on this. I guess the best way forward is to merge this very soon so we stop diverging and iterate toward a better template. Given the inheritance stuff got auto merged (and I am ok with that, it's easier to iterate when things get merged), we are already down that path. So unless there is an outcry about how ugly these hacks are, I will merge this tomorrow.

docs/config.ld
```diff
@@ -93,6 +93,257 @@ new_type("filterfunction", "List filters", false)
...
+        local new_params = {
+            name = { name  = name, value = "property::"..item.name },
+            description = params.description
+        }
+
+        return default_format_callback(self, new_params, item, md)
+    end
+}
+
+-- List the beautiful variables used by the method or property fllbacks.
```
actionless:
```suggestion
-- List the beautiful variables used by the method or property fallbacks.
```

Elv13:
Merging since the only commit that fail are due to the flacky geometry tests.

Issue #2869, opened 2019-09-13
Missing gears.table.iterate_value in awful.widget.layoutlist example


krzygorz:
Output of `awesome --version`:

awesome v4.3 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

**How to reproduce the issue:**

I pasted the center popup example from https://awesomewm.org/doc/api/classes/awful.widget.layoutlist.html into a file and `require`d it in my config. When i press mod4+space awesome throws an error.

**Actual result:**

    ~/.config/awesome/layoutlist.lua: 54: attempt to call a nil value (field 'iterate_value')

**Expected result:**

The popups appears, no errors are thrown.

**Additional notes:**

I found a [post](https://www.reddit.com/r/unixporn/comments/akbu6k/awesomewm_wait_why_did_i_do_this_again/) by @Elv13 in /r/unixporn which includes a [config](https://gist.github.com/Elv13/b679fe3cb100cd63a7082cc808d316b8) with a definition of `gears.table.iterate_value`. If this function is not supposed to be part of the awesome library, maybe the code should be included in the example?

gunar:
Sorry, anyone has an idea why I can't find `gears.table.cycle_value` on `awesome 4.3`?

psychon:
That function was added in commit 9e3c418a03 which is after 4.3. So the answer to your question is "because it was only added after the release of v4.3, not before".

PR #2925, opened 2019-11-25
More gears.table tests

SethBarberee:
Slowly working on #2869 but ended up doing these simple tests first to help... learning to write these in order to test iterate_value correctly

* Added tests for gears.table.hasitem and gears.table.find_keys

### DIFF
--- a/spec/gears/table_spec.lua
+++ b/spec/gears/table_spec.lua
@@ -42,6 +42,27 @@ describe("gears.table", function()
         end)
     end)
 
+    describe("table.find_keys", function()
+        it("nil argument", function()
+            local t = { "a", "b", c = "c", "d" }
+            local f = gtable.find_keys(t, function(k) return k == "c" end)
+            assert.is.same(f, {"c"})
+        end)
+    end)
+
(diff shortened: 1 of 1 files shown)


Elv13:
Thanks a lot for this. 

The current build fails because of : 

    spec/gears/table_spec.lua:48:54: unused argument 'v'

SethBarberee:
> Thanks a lot for this.
> 
> The current build fails because of :
> 
> ```
> spec/gears/table_spec.lua:48:54: unused argument 'v'
> ```

Fixed that right up. Thanks for the review.

Elv13:
Thanks. Can you squash all commits? Right now there is a revert and a fixup. To get merged, all commits have to pass the test. If they didn't that would break our ability to git-bisect the code.

If that's too much work, I can do it "soon".

SethBarberee:
> Thanks. Can you squash all commits? Right now there is a revert and a fixup. To get merged, all commits have to pass the test. If they didn't that would break our ability to git-bisect the code.
> 
> If that's too much work, I can do it "soon".

Done.

PR #2927, opened 2019-11-25
Correct the name of get_children's return value

xordspar0:
get_children's return value should not be called "The". That was a
mistake because ldoc requires a name for return values and parameters.
If a name isn't provided, it thinks the first word of the description
is the name.

### DIFF
--- a/docs/common/widget.ldoc
+++ b/docs/common/widget.ldoc
@@ -10,14 +10,14 @@
 
 --- Get or set the children elements.
 -- @property children
--- @tparam table The children
+-- @tparam table children The children.
 
 --- Get all direct and indirect children widgets.
 -- This will scan all containers recursively to find widgets
 -- Warning: This method it prone to stack overflow id the widget, or any of its
 -- children, contain (directly or indirectly) itself.
 -- @property all_children
--- @tparam table The children
+-- @tparam table children The children.
 
 --- Set a declarative widget hierarchy description.
 -- See [The declarative layout system](../documentation/03-declarative-layout.md.html)
--- a/lib/wibox/container/place.lua
+++ b/lib/wibox/container/place.lua
@@ -73,7 +73,7 @@ end
 
 --- Get or set the children elements.
 -- @property children
--- @tparam table The children
+-- @tparam table children The children.
 
 function place:get_children()
     return {self._private.widget}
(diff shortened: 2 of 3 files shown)


xordspar0:
Before:
image

After:
image

xordspar0:
> It would be also nice to have the final dot . at the end of the description.

Do you mean that it should look like this?

    children: (table) The children.

Elv13:
> Do you mean that it should look like this?

LDoc is a bit buggy when sentences don't end with `.`. It uses them to split the `<p>` and it causes surprising results when they are missing, but "not always due to ?".

Aire-One:
Oops I through I have approved the changes but it seems I did a comment only...

Yeah final dots aren't mandatory for ldoc to work but it seems it's better with them...

xordspar0:
Hm, I don't quite understand. The HTML doesn't change when I add a dot to the end of the description.

Without dot:
```html
<ul>
        <li><span class="parameter">children</span>
            <span class="types"><a class="type" href="https://www.lua.org/manual/5.3/manual.html#6.6">table</a></span>
         The children
        </li>
    </ul>
```

With dot:
```html
<ul>
        <li><span class="parameter">children</span>
            <span class="types"><a class="type" href="https://www.lua.org/manual/5.3/manual.html#6.6">table</a></span>
         The children.
        </li>
    </ul>
```

Anyway, I've added dots to all the descriptions I changed.

Elv13:
> Hm, I don't quite understand. The HTML doesn't change when I add a dot to the end of the description.

It isn't *always* necessary. The problem is that the places where is it and isn't are totally counter intuitive, so better always add them. See the ones @Aire-One is trying to fix in the global index https://awesomewm.org/apidoc/ . The multi line ones are caused by missing dots. But it isn't limited to that. I managed to produce stray HTML tags due to this multiple time.

xordspar0:
Interesting. That's helpful, thanks. You guys obviously have more experience than I do with LDoc since you're using it for such a big project, and I respect that.

Aire-One:
(for the reference: @Elv13 is referring to my PR #2921 (see the first commit 814719f220adf07a1454a4026f07d0ae83917b99 which finally fix the API-DOC broken index page by adding a lot of missing final dots))

PR #2928, opened 2019-11-26
[Documentation] `screen` module - Improvement on generated image: client part.

Aire-One:
This PR is the continuation of my previous PR #2922. 

The goal of this PR is to bring support for clients into the screen template image generator used by the screen API documentation.

I'd like to have feedback from @Elv13 to see if I correctly understood what we should do on this one.
The code is a little dirty now and requires a rework of my commit before being merged anyway.

Here is a previous of the generated image for the `screen.tiled_clients` property:
tiled_clients

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -483,6 +483,8 @@ end
 -- * maximized clients
 -- * floating clients
 --
+-- @DOC_screen_tiled_clients_EXAMPLE@
+--
 -- @property tiled_clients
 -- @param table The clients list, ordered from top to bottom.
 
--- a/tests/examples/screen/template.lua
+++ b/tests/examples/screen/template.lua
@@ -81,6 +81,7 @@ local colors = {
     tiling_area = "#ff0000",
     padding_area= "#ff0000",
     wibar       = "#000000",
+    tiling_client = "#ff0000",
 }
 
 local function draw_area(_, rect, name, offset, highlight)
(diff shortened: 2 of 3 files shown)


lib/awful/screen.lua
```diff
@@ -483,6 +483,8 @@ end
 -- * maximized clients
 -- * floating clients
 --
+-- @DOC_screen_tiled_clients_EXAMPLE@
```
actionless: shouldn't this example go to documentation of `awful.layout.suit.tile.right`?

Elv13: In the short term yes, in the long term, I still plan to merge https://elv13.github.io/client_tiling/dynamite.html for v5.0. It has a different doc template because these doc examples have a lot more `assert` and tests built-in.

Aire-One:
> Thanks! We can definitely build from that.

Ok nice. Let me the time to come back on this and clean the code.

lib/awful/screen.lua
Aire-One: It definitely could. I'd like to work on the client layout documentation. We deeply need a user documentation on this API and better examples for available client layouts.

I'm not sure it would be worth the time spending in it, tho, given the fact @Elv13 will eventually merge his module... So this API will be deprecated/removed "soon".

Aire-One:
Should be ready now! :)

lib/awful/screen.lua
Elv13: > "soon"

For undefined version of "soon". It can definitely be used in the meantime. You will hit some problems in the shims where some layouts are just not rendered properly without serious effort. I think there is lower hanging fruits to spend time on. The inheritance stuff is a great example of how to add value to everything with a one-time effort.

PR #2929, opened 2019-11-28
Doc fix modules summaries

Elv13:
This is a subset (and rebased) version of @Aire-One pull request. I stripped the new ldoc tags since, for now, they didn't do anything anymore. I don't mind if they are added again for some reasons. I also fixed a bug where the method/property summary would be overly cropped on small resolution.

If we merge this fast, then I will submit a new PR with better "tags" next to the property/method title (in the description). I will also remove the `@deprecated`/`@deprecatedproperty` in favor of using the same codepath as the inherited items.

### DIFF
--- a/docs/common/fixed.ldoc
+++ b/docs/common/fixed.ldoc
@@ -1,42 +1,65 @@
 --- Set a widget at a specific index, replace the current one.
--- **Signal:** widget::replaced The argument is the new widget and the old one
--- and the index.
+--
 -- @tparam number index A widget or a widget index
--- @param widget2 The widget to take the place of the first one
+-- @tparam widget widget2 The widget to take the place of the first one
 -- @treturn boolean If the operation is successful
 -- @method set
+-- @emits widget::replaced
+-- @emitstparam widget::replaced widget self The layout.
+-- @emitstparam widget::replaced widget widget index The inserted widget.
+-- @emitstparam widget::replaced widget previous The previous widget.
+-- @emitstparam widget::replaced number index The replaced index.
+-- @interface layout
 
 --- Replace the first instance of `widget` in the layout with `widget2`.
+--
 -- **Signal:** widget::replaced The argument is the new widget and the old one
 -- and the index.
--- @param widget The widget to replace
--- @param widget2 The widget to replace `widget` with
+-- @tparam widget widget The widget to replace
+-- @tparam widget widget2 The widget to replace `widget` with
 -- @tparam[opt=false] boolean recursive Dig in all compatible layouts to find the widget.
 -- @treturn boolean If the operation is successful
 -- @method replace_widget
+-- @emits widget::replaced
+-- @emitstparam widget::replaced widget self The layout.
+-- @emitstparam widget::replaced widget widget index The inserted widget.
+-- @emitstparam widget::replaced widget previous The previous widget.
+-- @emitstparam widget::replaced number index The replaced index.
+-- @interface layout
 
 --- Swap 2 widgets in a layout.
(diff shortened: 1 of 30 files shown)


Aire-One:
I'm a little confused about this PR.

> I stripped the new ldoc tags since, for now, they didn't do anything anymore. I don't mind if they are added again for some reasons.

I'm really not sure I understand what you mean here... Why wouldn't the new ldoc tags work here?

> If we merge this fast, then I will submit a new PR with better "tags" next to the property/method title (in the description). 

I think these PR are unrelated.

This one and my #2921 PR are about improving the modules documentation. They shouldn't touch the ldoc configs/templates stuff and only use new tags provided by your #2924 PR. (Which is merged on master now, so rebase or merge these PR will automatically made use of the new stuff)

If you do a new PR to implement a "better tags next to the property/method title", this new PR will change the ldoc template (i guess) using the new ldoc tag system of your #2924 PR. (and again, it will be applied to our other PR after merge)

(I removed the ldoc inheritance stuff from my #2921 to not conflict with your #2924. BTW I rebased my PR on master on my local machine and it works just as expected and show in the screenshots from https://github.com/Elv13/awesome-1/pull/14)

Elv13:
> Why wouldn't the new ldoc tags work here?

Well, it works without them:

Screenshot_20191128_170021

Screenshot_20191128_165934

The only differences are:

 * The "Inherited from" is only displayed once instead of twice in the summary.
 * The inherited content are in the same sections as the non-inherited ones.

I think both of those differences are features. When I merge the "deprecated" sections back into the normal ones, then inherited/deprecated will just inline metadata and that's good. Since "logically" methods are methods and properties are properties, there is no need to segregate them into different sections. They were segregated because it was easier to do it that way. Plus we could add "New!" for content introduced in the current release using the same conventions.

Plus another slight problem with having `@deprecatedfoo` and `@inheritedfoo` is that it makes new contributions harder because new contributors will always get the tags wrong the first time.

If we look at this from the "other side", having these tags would allow for future improvements and more customization. However given that none of that exists *now*, I am not sure if its worth adding these tags now. Do you have something in mind that would use them to their full potential?

Elv13:
I added the commits using the new tags for the `wibox` module. Backfilling this is a lot more labor intensive than in looks...

Technically, the part to describe the deprecation is in those commits, but it isn't used in the template yet. Same goes for `@baseclass` cousin, the new `@interface` tag.

Aire-One:
Wow! It goes quicker than what I expected... Nice work!

> Technically, the part to describe the deprecation is in those commits, but it isn't used in the template yet. Same goes for @baseclass cousin, the new @interface tag.

Do you have any plan regarding how they should be used in the template? Is the way I implemented `@baseclass` the right direction, or do you plan to implement something more generic?

Elv13:
> Do you have any plan regarding how they should be used in the template? Is the way I implemented @baseclass the right direction, or do you plan to implement something more generic?

The same way. No need to be more generic yet. I like the way you did it.

Elv13:
Ok, I fixed the `luacheck` warning and got lucky and the Travis build passed. I think I am done with the doc for now. That way I will avoid creating more merge conflicts against @Aire-One great work. I think we have a stronger framework to move forward now. Given how trivial it now is to add metadata, over time it will really make a difference. If anyone wants to pick up the work on adding the metadata tags to the other modules (see this PR diff for endless examples), that would be greatly appreciated. It requires no real knowledge of the codebase. A great first contribution!

I will go bash my head at the snail pace work on the remaining API standardization work, which block my screen rules project, which I want merged before 4.4.

PR #2936, opened 2019-12-01
Untangle some easily avoidable circular dependencies

Elv13:
These commits were part of the "API standardization part 4" PR, but that's blocked on "it's really boring, annoying, hard-not-to-regress-anything and I don't have time". The following commits are cherry-picked from it and could be merged sooner.

It refactor the code to have less internal dependencies (either on modules or behavior). It was required mostly because changing `c:geometry()` to `c.geometry` and other "legacy accessors" along with adding `append/remove` for "set once" list have some weird side effects.

Anyway, having less internal dependencies is a good thing.

### DIFF
--- a/lib/awful/mouse/client.lua
+++ b/lib/awful/mouse/client.lua
@@ -1,0 +1,143 @@
+--- Client related mouse operations.
+--
+-- @author Emmanuel Lepage Vallee &lt;elv1313@gmail.com&gt;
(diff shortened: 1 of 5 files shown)


Elv13:
Ok, I added more tests. This is now more palatable to merge. I will do it soon because it blocks my other API standardization pull request.

PR #2937, opened 2019-12-02
More awful.key / awful.button improvements.

Elv13:
This is a subset of the larger `awful.keyboard` PR I don't have time to finish. It is useful enough on its own. It tries harder to use `awful.key` and `awful.button` instead of `capi.key`/`capi.button`. It is now advanced enough to add features like key "groups" for "numpad", "numrow" and "arrows" keys with a single `awful.key` object. This removed the imperative loop from `rc.lua` and will eventually make @actionless hotkey popup simpler since it no longer has to reverse-engineer this information from the way keys list. This also add the `append`/`remove` to the clients for the keys and button. That's to make modules easier to develop. 

~~It still needs tests, I don't think the CI will pass as-is. It's good enough for reviews rather than rotting in my local branches.~~ done

### DIFF
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -274,7 +274,9 @@ function widget.new(args)
                 end
                 table.sort(
                     sorted_table,
-                    function(a,b) return (a.mod or '')..a.key<(b.mod or '')..b.key end
+                    function(a,b)
+                        local k1, k2 = a.key or a.keys[1][1], b.key or b.keys[1][1]
+                        return (a.mod or '')..k1<(b.mod or '')..k2 end
                 )
                 target[group] = sorted_table
             end
@@ -287,7 +289,9 @@ function widget.new(args)
             return
         end
         for _, data in pairs(awful.key.hotkeys) do
-            self:_add_hotkey(data.key, data, self._cached_awful_keys)
+            for _, key_pair in ipairs(data.keys) do
+                self:_add_hotkey(key_pair[1], data, self._cached_awful_keys)
+            end
         end
         self:_sort_hotkeys(self._cached_awful_keys)
     end
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -378,64 +378,63 @@ clientkeys = {
 }
 
 -- @DOC_NUMBER_KEYBINDINGS@
--- Bind all key numbers to tags.
--- Be careful: we use keycodes to make it work on any keyboard layout.
--- This should map on the top row of your keyboard, usually 1 to 9.
-for i = 1, 9 do
-    -- View tag only.
-    awful.keyboard.append_global_keybinding(awful.key(
-        { modkey }, "#" .. i + 9,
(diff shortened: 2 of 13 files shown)


lib/awful/hotkeys_popup/widget.lua
```diff
@@ -274,7 +274,9 @@ function widget.new(args)
                 end
                 table.sort(
                     sorted_table,
-                    function(a,b) return (a.mod or '')..a.key<(b.mod or '')..b.key end
+                    function(a,b)
+                        local k1, k2 = a.key or a.keys[1][1], b.key or b.keys[1][1]
```
actionless: shouldn't it be `join(a.keys[1])` instead of `a.keys[1][1]`?

Elv13: Currently, it looks like that:

Screenshot_20191204_014332

Is it correct?

As for the question itself, I guess the answer is that once the whole keyboard API matures, this code will need a serious cleanup since half of what it does (the whole reverse engineering part) will no longer be needed. Also the private APIs it uses will probably be fully obsolete and removed (like `awful.key.hotkeys`)

Elv13:
I added tests. It was working better than I expected. I think this can be merged soon unless people find more issues.

SethBarberee:
> This also add the `append`/`remove` to the clients for the keys and button. That's to make modules easier to develop.

So if I understand correctly, does that mean you could add keys/buttons based on certain clients with a rule? If so, that's a big thing in my opinion. I'd imagine with the `awful.rules` revamp of yours that this is the goal, right?

Elv13:
> So if I understand correctly, does that mean you could add keys/buttons based on certain clients with a rule? If so, that's a big thing in my opinion. I'd imagine with the awful.rules revamp of yours that this is the goal, right?

Yes, that's the goal.

Aire-One:
I only quickly read through your code here. I like how elegant the `awesomerc.lua` become :)

It seems good to me and it's working with my config so far.

Elv13:
> I only quickly read through your code here. I like how elegant the awesomerc.lua become :)

Thanks. There is a lot more to come (the rules PR also adds some). The Activity/Screen_rules one will be the big one for the `rc.lua` modularity project. After that only the "default apps" and "mymainmenu" section will remain non-modularizable. In both case, this will require a giant amount of work to replace those broken things. I have a plan, but I didn't start the work. So it is likely to be done in 5.x rather than 4.x.

Elv13:
I added some signal doc along with an entry in NEWS.md for the little breaking change present in this PR. I will merge this once the test finish since there is apparently a general enthusiasm around this new feature. I also need this merged to rebase the awful.rules->ruled.client PR.

Elv13:
Merged, please report any regression.

lib/awful/hotkeys_popup/widget.lua
actionless: as i understood, the sorting with [1][1] would be an issue only in case when hotkey have multiple `keys`

PR #2941, opened 2019-12-06
API standardization, part 3

Elv13:
Here are some commits of the next API standardization step.

The third part is about `request::` signal and `awful.ewmh`. `awful.ewmh` had nothing to do with `EWMH` for quite a while. It is now renamed `awful.permission`. The `request::` signal are standardized (not all of them yet, there is more commits to come). With this standardization, it is now possible to implement a standard permission framework to grant or deny those requests.

The PR also touches the focus code. There is a new `client:activate{}` method to handle the increasing amount of boilerplate code in `rc.lua` regarding focus and raise. The `action` property is a bit weird for now, but a (far away) pull request will eventually add a new "functional" API to obsolete and supplant all the "increment" and "select" static functions we have in the API. However this isn't for the next release, so for now it will look a bit out of place.

~~It wont "work" for now since it depends on unmerged code. It also lacks good tests.~~


Elv13:
At this point it seems the only build failure is the flacky placement rule test (please, if someone manage to reproduce that thing on their hardware, fix it). Those PRs are always scary and regression prone. Anyhow, there is a lot of nice features in this work. 

* There is now official theme APIs to set the border color and width for the different client states.
* The client opacity now has some `beautiful` theme variables.
* There is better support for different client layouts per tag.
* There is now a `request::` signal for the default/fallback layouts list, so modules can easily add more.
* A lot of code that was in `rc.lua` for historical reason has been moved into the core, allowing to fix some edge case issues along with it (the code was much shorter that it should have been to keep `rc.lua` simple).
* The new client `:activate{}` method will ease changing the focus without side effects.
* The `request::manage` (and `unmanage`) signals provide a context the old `manage` lacked.
* The clients now have an `active` property, so it is no longer necessary to do `client.focus == c`, which was different than other AwesomeWM APIs. It also has an atomic signal, which the previous code lacked.
* While not yet implemented for size hints, fullscreen and maximization, the client `:grant()` and `:deny()` methods allow to control a wide range of events. It is kind of a firewall for what the client can and cannot do.
* The documentation better reflects where `request::` are emitted and which signals are class-only rather than on objects and class.

docs/89-NEWS.md
```diff
@@ -62,6 +62,14 @@ This document was last updated at commit v4.3-197-g9085ed631.
   objects. If you used these low level APIs to add keys and buttons dynamically,
   please migrate your code to the corresponding `:append_` and `:remove_`
   client methods.
+ * `beautiful.border_width` and `beautiful.border_color` are not honored even
```
actionless: i see `border_width` is still set in the updated themes above (or they're not finished yet?)

also i think it makes sense to leave smth like `beautiful.client_border_width` to be able to set default width for all the clients disregard their state (while not affecting other widgets which have border)

and the same for `client_border_color`, as i could easily imagine the scenario when somebody would like to set one color as default one, and set explicitly colors only for some of the states

Elv13: s/not/now/

Sorry for the "typo", but good catch. So half of what you asked was already there, but I did add a `beautiful.border_color` field as a generic fallback to address the other half. I also added a `fullscreen` variant to address another issue (well, kind of, the bug is still there but it will be easier to fix).

Elv13:
And in it goes.

As always, this serie is regression prone, so please test so we can stabilize it as soon as possible. This was the part 3 of 5. The other 2 parts, in no specific order is the one which convert all legacy "method properties" like `:geometry()` and `:struts()` to proper Lua property. The other is converts all widget constructor to use the same syntax as the declarative widget definition, finally unifying them.

awesomerc.lua
```diff
@@ -95,10 +74,33 @@ mylauncher = awful.widget.launcher({ image = beautiful.awesome_icon,
 menubar.utils.terminal = terminal -- Set the terminal for applications that require it
 -- }}}
 
+-- {{{ Tag
+-- @DOC_LAYOUT@
+-- Table of layouts to cover with awful.layout.inc, order matters.
+tag.connect_signal("request::default_layouts", function()
```
actionless: i can't understand what this warning means:
```
2020-01-12 02:17:37 W: awesome: `awful.layout.layouts` was set before `request::default_layouts` could be called. Please use `awful.layout.append_default_layouts` to  avoid this problem
```

when i do
```lua
tag.connect_signal("request::default_layouts", function()
  print("HI")
end)
```
nothing happens

--------------------------------

when i do
```
for _, l in ipairs(awful.layout.layouts) do
  awful.layout.remove_default_layout(l)
end
awful.layout.append_default_layouts(mylayoutslist)
```
it only adds layouts to default ones, but not replacing as `awful.layout.layouts = mylayoutlist` would do

awesomerc.lua
```diff
@@ -562,13 +534,4 @@ client.connect_signal("request::titlebars", function(c)
         layout = wibox.layout.align.horizontal
     }
 end)
-
--- Enable sloppy focus, so that focus follows mouse.
```
actionless: and how to disable it now?

Elv13: The permission system or by disconnecting the request. This is how all other `request::` are already used. The `request::` API was originally designed to move some hard coded behavior into something you could disable and replace by your own logic. In this case, it was used to move things that boilerplate for most people (like setting the border) into something invisible, but easy to disable.

Add to that the new permission API, and you can now define what you wish to enable/disable in a per-client basis. Another pull request (because the `ruled.client` one is already maxed-out) will add an easy way to do that from the client rule. The end result will be superior and easier to manipulate than having the code in `rc.lua`. See the new "Permission" section in the doc (you need to click "Show more" to see them, but they are in the summary).

awesomerc.lua
Elv13: > nothing happens

I will investigate that, the signal is definitely sent when the first tag is created.

Overall, this change is a tradeoff. The problem with `awful.layout.layouts` is that it could only be set safely from one place. Setting it from multiple place required each setter to co-operate and use `table.insert`, which would break as soon as one setter just sets it. The overall goal, as we discussed a couple years ago, is that each API should be "module friendly" and can be called by modules rather than being only safe to use in `rc.lua`.

The use case for *this* one is a module which provide new layouts. Just by doing a `require("mylayoutmodule")`, you can add your layouts to the default set. 

> it only adds layouts to default ones, but not replacing as awful.layout.layouts = mylayoutlist would do

Should I add a `clear_layouts()` function? This would cover this use case. It is also mostly covered on the "tag object level" where you can set a list of layouts for a specific tag,

awesomerc.lua
actionless: but how you going to disconnect lambda?

awesomerc.lua
actionless: that's actually weird what in default config it works, but not in mine:

when i replace both layouts and tags part from rc.lua i'm ending up with no tags at all, as neither `default_layouts` nor `desktop_decorations` signals are being called

awesomerc.lua
actionless: ping, that's still a noticeable functional regression in the current master

Elv13: Can you provide an example based on the default config where the behavior changed?

actionless: it's not possible anymore to disable a "sloppy focus", in order to disconnect a signal it should be an exported function of the module, not an anonymous or local one

Elv13: It is still possible using the permission API, but fair enough. I worked on an API level system last weekend, but it isn't ready for review yet. These API changes are "required" going forward for the next round. You will like them because they get us closer to HiDPI.

But right, the autofocus stuff needs to go back to `rc.lua` for API level 4, there is no way to detect why the user removed it (either because he/she use the new API or because they don't want it)

actionless: so to simplify the things up for now -- what about just moving the anonymous function to be a module-level method?

actionless: (in a similar manner like `client.disconnect_signal("manage", awful.rules.apply)`)

PR #2944, opened 2019-12-12
[Draft/WIP] Documentation - Client layout. 

Aire-One:
Hi!

Here is my first attempt at writing the long awaiting *client layouts documentation*. I want your help to write this documentation, any feedback is welcomed.

For now, this adds a new "plain documentation" page and project level section for client layout. I still have _a lot_ of work on this before it's ready. This still WIP stuff.

What I want for this PR:

* A general documentation page about client layouts (Based on the [widget system documentation](https://awesomewm.org/apidoc/documentation/03-declarative-layout.md.html))
    * Description of client layouts [Done, Need Improvement]
    * How to use / How to configure client layout [Need Improvement]
    * List of client layouts from `awful.layout` module
    * How to create your own layout
* A new project level section for client layouts
    * Define each client layout individually [done]
    * Good description for each layout with examples and images (cf https://github.com/awesomeWM/awesome/pull/2928#discussion_r350857600)
    * List of properties to configure each layout
* Make the `awful.layout` documentation an abstract/internal documentation

### DIFF
--- a/docs/08-client-layout-system.md
+++ b/docs/08-client-layout-system.md
@@ -1,0 +1,54 @@
+# The AwesomeWM client layout system
+
+This document explains how to use clients layouts and how awesome manage them.
+
+**Client layout**  refers to the mechanism awesome uses to place client on the screen. The layout definition can be a basic system where clients are all *floating* like in a normal *DE* (eg GNOME, KDE, ...) or tiled on the screen like in *tilled window manager* (eg i3, BSPWM, ...). It is also possible to define complex client layout mixing both concepts and even implementing a *[stateful](https://en.wikipedia.org/wiki/State_(computer_science)) placement strategy*.
+
+Awesome WM manages client layouts per tag. It means each tag has its own layout selection and uses them independently from other tags. When multiple tags are selected at once, Awesome uses only client layouts from the first selected tag and apply it to all current clients.
+
+## Layouts configuration
+
+Layout can be configured by setting properties from the tag instance.
+
+Example of creating a new tag with client layout parameters:
+```lua
+awful.tag.add("My Tag", {
+    screen = screen.primary,
+    layout = awful.layout.suit.tile,
+    master_fill_policy = "master_width_factor",
+    gap_single_client = false,
+    gap = 15
+})
+```
+
+Example of changing client layout parameters on an existing tag:
+```lua
+-- Change the gap for the tag `my_tag`:
+my_tag.useless_gap = 10
+
+-- Change the master_width_factor for the tag `my_tag`:
+my_tag.master_width_factor = 0.5
+```
+
+Here is a list of common properties used to configure tags:
+
+<table class='widget_list' border=1>
+<tr>
+<th align='center'>Property</th>
+<th align='center'>Type</th>
+<th align='center'>Description</th>
+</tr>
+<tr><td>gap</td><td>number</td><td>The gap (spacing, also called useless_gap) between clients.</td></tr>
+<tr><td>gap_single_client</td><td>boolean</td><td>Enable gaps for a single client.</td></tr>
+<tr><td>master_fill_policy</td><td>string</td><td>Set size fill policy for the master client(s).</td></tr>
+<tr><td>master_count</td><td>integer</td><td>Set the number of master windows.</td></tr>
+<tr><td>icon</td><td>path or surface</td><td>Set the tag icon.</td></tr>
+<tr><td>column_count</td><td>integer</td><td>Set the number of columns.</td></tr>
+</table>
+
+
+## Creating new client layouts
+
+* arrange function (params definition)
+* needs a name property
+* mouse_resize_handler function
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -38,6 +38,7 @@ topics={
     '05-awesomerc.md',
     '06-appearance.md',
     '07-my-first-awesome.md',
+    '08-client-layout-system.md',
     '16-using-cairo.md',
     '17-porting-tips.md',
     '90-FAQ.md',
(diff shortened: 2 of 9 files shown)


Elv13:
Hi, thanks for this. The new doc page will be a nice addition. One thing I am not a fan of is adding the individual client layouts back into the left page index. I removed them on purpose because I wanted to have one documentation page to document them all, next to each other. One page for each dilutes the information across too many places. I would prefer to keep `awful.layout` as a single doc module as-is. What do you think?

Aire-One:
> One thing I am not a fan of is adding the individual client layouts back into the left page index. I removed them on purpose because I wanted to have one documentation page to document them all, next to each other. One page for each dilutes the information across too many places. I would prefer to keep `awful.layout` as a single doc module as-is. What do you think?

I published this PR as a draft (and long before it's completed) to have this kind of opinion. I totally understand your POV.

I'd like to document every tag properties for each client layout and add generated images to show what happen when setting them. I'm afraid the `awful.layout` documentation became too hard to read if it stays as one block for every client layouts.

I'm in the process of writing an image generator template for `awful.layout.suit.*`. I'll try to build documentation with both styles (all in `awful.layout` and with individuals pages) to see the differences.

Aire-One:
Also, the current travis error report seems unrelated to my PR: 

```
Failure → spec/beautiful/init_spec.lua @ 14
beautiful init Check beautiful.get_font
spec/beautiful/init_spec.lua:15: Expected objects to be the same.
Passed in:
(userdata) 'lgi.rec 0x2458af0:Pango.FontDescription'
Expected:
(userdata) 'lgi.rec 0x1f05d20:Pango.FontDescription'

stack traceback:
	spec/beautiful/init_spec.lua:15: in function <spec/beautiful/init_spec.lua:14>
```

Elv13:
What do you think about importing some images from https://elv13.github.io/client_tiling/dynamite.html as SVG instead of spending that many hours making them with a template? It's going to take a looooong time if you want to start over.

Aire-One:
> What do you think about importing some images from https://elv13.github.io/client_tiling/dynamite.html as SVG instead of spending that many hours making them with a template? It's going to take a looooong time if you want to start over.

Sorry for the delay, I've had a busy week-end.

TBH, I didn't even think about looking at your layout documentation. Which would be better starting point than starting from scratch like I did...

Anyway, I based my template on what you did for the screen API recently. I only need additional helpers to make the image generator code easier to write and it will be ok-ish. Most features I want are already there. I only need a better system to handle multiple screen representation (to show the tiling process by popping clients one-by-one).

Currently I use multiple screens (screen[1] = 1 client ; screen[2] = 2 clients ; ...) but I don't think it's the best way to do that... I can't find back your dynamite module repository to check how you generated large screens matrices like for the "dynamite.suit.corner Client count scaling" section.

Elv13:
> I can't find back your dynamite module repository to check how you generated large screens matrices like for the "dynamite.suit.corner Client count scaling" section.

It is in a bunch of branches, but none of them are easy to find. Here it is:

https://github.com/Elv13/awesome-1/blob/doc_tests_and_notif/lib/dynamite/tests/suit/template.lua
https://github.com/Elv13/awesome-1/blob/doc_tests_and_notif/lib/dynamite/tests/suit/tile/mwfact.lua

I think the way my template does it is rather clean. I don't think you can make the `sequence` template inspired code work with all layouts. There is some serious hacks in the way they are rendered by that template and it doesn't work with all of them. That's why the examples always skip some suits. My `Collision` module is where that code comes from and it has extra hacks to make more edge case work. This is why it is probably a better idea to just use those images as-is and adding them to the repo.

Aire-One:
I'm working slowly and inconsistently on this PR, so I'm gonna open up the discussion to make my work faster when I'll get back on it.

---

> I don't think you can make the sequence template inspired code work with all layouts. There is some serious hacks in the way they are rendered by that template and it doesn't work with all of them. That's why the examples always skip some suits.

Hm... I should have say it more precisely but I don't intent to make a sequence template (Maybe a really limited one to show clients arrangement but that's all). From the quick look I took to your template (I'm maybe wrong because I only superficially read the code), it can also be used to run step-by-step tests to check tags properties, mouse movements, check titlebars, widgets and wiboxes, ... I don't target this here.

My goal is to take a client layout with some tag's properties configured and add clients to show how it renders. For now, I only worked with the `suit.tile.*` layouts, so I'm a little surprise about what corner cases we can have with other layouts. Can you elaborate on this please?

---

About my _each-client-layout-will-get-its-own-documentation-page_ project: I know @Elv13 is opposed to this, I'd like to discuss about it.

What I want here is to define additional ldoc tags to add a section for client layout parameters *. This way, we can easily describe what client layout can be configured with what properties and also show how it impacts the layout render.

I think it's cleaner this way than defining all client layouts in one single page. By experience, I know client layouts implement tags properties rather consistently but as a user, you have to try them to understand it, and third party layouts are not guaranty to follow the same guidelines. So having a clear documentation for our client layouts is better IMO.

\* This is my idea, but I'd like your feedback: It would add an ldoc tag with a very limited use case so I think it's a bad idea. BUT, I think it'll be rather useful for users to have this section and the fact we use a ldoc tag make its formatting easier and consistent across layouts. Which is better than having a long markdown description.

Aire-One:
I pushed some quick code I wrote to show what I want to do.

Here is the result I have for the `suit.tile` layout documentation:
Screenshot_2019-12-19 Clientlayoutmod awful layout suit tile - awesome API documentation

Aire-One:
I had a few time recently to work on this PR.

I moved everything to a backup branch and added back only the documentation part to complete it before going further into the other stuff...

I think this documentation is near to be ready, now.

I don't like how I wrote tables in the markdown but I'm not sure how else I can do with ldoc...

Also, I'd like help to document stateful clients layout. I only find about them in weird places in the API documentation (and I can't find it back now), so I'm not sure how it works...

PR #2954, opened 2019-12-30
Fix off by one error in keyboardlayout.lua

Babar:
Otherwise, when reaching the end and wrapping, we'd end up at 0 which will return nil and crash

### DIFF
--- a/lib/awful/widget/keyboardlayout.lua
+++ b/lib/awful/widget/keyboardlayout.lua
@@ -273,7 +273,7 @@ function keyboardlayout.new()
     end
 
     self.next_layout = function()
-        self.set_layout((self._current + 1) % (#self._layout + 1))
+        self.set_layout(((self._current + 1) % #self._layout) + 1)
     end
 
     self.set_layout = function(group_number)


Elv13:
Thanks for this. I wonder how we could test that. Do you have a small seld-contained example of how to reproduce the issue in mid? Looking at the code, that method is also undocumented. I can fix that later.

Babar:
I did test this. The way I tested it:
0. In order to recover from this easily, open a shell with something like: `setxkbmap us`
1. Use the default configuration, which should show your keyboard in the top bar, next to the clock
2. Click the two letters of your layout (for me, `us`)
3. Observe that the `"us"` letters / widget disappears. It might log an error somewhere, but I did not find anything. Also observe that the keyboard is now behaving very weirdly. For example, I remap most of my keyboard to get accentuated characters using `xmodmap`. Most key codes look like this:
`keycode  24 = q Q eacute Eacute`
In this state, I can only type accentuated characters, therefore pressing `q` will always yield `é` or `É`, never `q` nor `Q`.
4. Press enter on your shell to run `setxkbmap us`
5. Observe that the widget re-appears

Then I applied this patch to `/usr/share/awesome/lib/awful/widget/keyboardlayout.lua`, and when I pressed `us` in step 2, nothing happened (I only have one layout defined, which is the expected behavior)

Aire-One:
IMHO, this doesn't need test. Other than checking that the selected layout exists...

The bug was a mathematical error:

```next = (  (current + 1) % MAX  )```

Then add a last `+1` for the lua array indexing shift.

The only way I see to test this would be to call the `next_layout` function then check if we have something (and probably also check if it's the correct next layout).

To avoid this kind of error in the future, we could also have a circular buffer/array declaration somewhere but it's probably a bit overkill I think...

**Comment by psychon** [comment]
```lua
-        self.set_layout((self._current + 1) % (#self._layout + 1))
+        self.set_layout(((self._current + 1) % #self._layout) + 1)
```
Uhm. I'm not sure this patch is correct. Let's put in some numbers: `_current = 1` and `#self.layout = 10`. With the old code, we end up with `(1 + 1) % (10 +1) = 2`. With the new code, we end up with `((1 + 1) % 10) + 1 = 3`. So, this should produce different numbers than the old code.

I'm not sure about this, but... should the inner `+1` be dropped?

Aire-One:
Duuh... I was too focused on the loop back from the last value to the first part and totally forgot about basic numbers...

In addition, looking closer to the code, the `update_state` function handle the array shift by itself by adding `+1` to the current index. So I was wrong thinking we should handle the lua array shift...

You're right, except the old code still doesn't handle the loop back part correctly: `(9 + 1) % (10 +1) = 10` which is out of bound for an array of 10 elements (index: `[0 ; 10[`). The next of `9` should be `0` not `10`.

> I'm not sure about this, but... should the inner `+1` be dropped?

I guess this is the correct way here.

```lua
-        self.set_layout((self._current + 1) % (#self._layout + 1))
+        self.set_layout((self._current + 1) % #self._layout)
```

PR #2955, opened 2020-01-01
Very WIP but slowly convert layoutbox to accept template

### LINKED ISSUE #2950: Add callback and widget_template parameters for awful.widget.layoutlist
In widgets like awful.widget.tasklist and awful.widget.taglist one can pass a callback function and a widget template, so it's easy to change the whole widget.

In awful.widget.layoutlist it's only possible to set the screen and buttons, so it's very inconvenient to change the behaviour of this widget.

SethBarberee:
Closes #2950 

Took a crack at it... still trying to figure out `awful.widget.common` and `wibox.widget.base` for the purposes of the template. Commented with my TODOs of where I'm currently tripped up.


lib/awful/widget/layoutbox.lua
```diff
@@ -10,38 +10,83 @@
...
+}
+
+function layoutbox:layout(_, width, height)
+    if self._private.layout then
+        return { wibox.widget.base.place_widget_at(self._private.layout, 0, 0, width, height) }
+    end
+end
+
+function layoutbox:fit(context, width, height)
+    if self._private.layout then
```
Aire-One: Are you missing a `not` here?

lib/awful/widget/layoutbox.lua
```diff
@@ -67,15 +112,49 @@ function layoutbox.new(args)
...
+
+    function w._do_layoutbox_update_now()
+        if screen.valid then
+            update(w, screen, args.buttons, data, uf, args)
+        end
+        queued_update[screen] = false
+    end
+
+    function w._do_layoutbox_update()
+        -- Add a delayed callback for the first update.
```
actionless: what is the idea behind it?
as far as i understand it, in case of multiple updates, it will take the first of them, not the last one

lib/awful/widget/layoutbox.lua
SethBarberee: Yeah, need to push my commit.

lib/awful/widget/layoutbox.lua
SethBarberee: Still working with it. I've mainly been looking at `tasklist` and `taglist` as templates. My answer right now: :man_shrugging:

Elv13: > what is the idea behind it?

For stuff like the taglist, calling `awful.widget.common` "update" and "label" is very expansive. Usually, more than 1 tag is created in the same event loop iteration. Without batching those calls, it slows down startup a lot. With the new client layout `append` API, the layoutbox will have the same problem, so I am happy to see batching being integrated so early! Usually it was the result of my profiling AwesomeWM and trying to address some bottlenecks.

actionless: @Elv13 and what about second part of my question?

SethBarberee:
Rebased and took another crack. Mainly getting errors around the reset part. Reading more of the code to understand what is going on.

```
[ 77%] Generating doc/images/AUTOGEN_awful_notification_geometry.svg
Result: 1
Error:
lua: /home/seth/seth-awesome/lib/awful/widget/common.lua:124: attempt to call a nil value (method 'reset')
stack traceback:
	/home/seth/seth-awesome/lib/awful/widget/common.lua:124: in function 'awful.widget.common.list_update'
	/home/seth/seth-awesome/lib/awful/widget/layoutbox.lua:88: in upvalue 'update'
	/home/seth/seth-awesome/lib/awful/widget/layoutbox.lua:179: in function 'awful.widget.layoutbox.new'
	(...tail calls...)
	...seth/seth-awesome/tests/examples/shims/_default_look.lua:10: in main chunk
	[C]: in function 'require'
	...h-awesome/tests/examples/awful/notification/geometry.lua:6: in main chunk
	/home/seth/seth-awesome/tests/examples/awful/template.lua:12: in main chunk
	[C]: in ?
```

lib/awful/widget/layoutbox.lua
SethBarberee: Fixed

Elv13:
Hi, thanks for this effort. I created a `relax_awful_common` branch in my fork with a commit to help with the reset issue. Maybe an alternative would be to only use/expose `awful.widget.common._set_common_property` and expose the `custom_template` local function rather than try to use the whole stack. That code has only really been used with layouts and is *ancient*. It predates even the current widget system itself (we had 3 different rewrite of that system).

lib/awful/widget/layoutbox.lua
Aire-One: I have one more question : `queued_update` is a local table. It means each instance of layoutbox will have its own `queued_update`. `screen` is also local to the layoutbox instance.

So, why don't we use a basic boolean variable, instead?

OR: the declaration of `queued_update` should be moved out of the constructor. (And probably use something else than the screen to identify the layoutbox instance, IMO.)

PR #2958, opened 2020-01-03
Docify gears.cache

SethBarberee:
Messed around a bit and added a doc for `gears.cache`. Never used it until now so did a little sample case to mess with it. Might use it more for some of my modules. 

Anyway, added a test for it and added it to the doc.

### DIFF
--- a/lib/gears/cache.lua
+++ b/lib/gears/cache.lua
@@ -1,4 +1,16 @@
 ---------------------------------------------------------------------------
+--- Cache object with data that can be garbage-collected.
+--
+-- Here is an example with a basic cache:
+--
+--@DOC_text_gears_cache_cache_EXAMPLE@
+--
+-- The example below demonstrates how the garbage collector will clear the
+-- cache:
+--
+--@DOC_text_gears_cache_another_cache_EXAMPLE@
+--
+--
 -- @author Uli Schlachter
 -- @copyright 2015 Uli Schlachter
 -- @classmod gears.cache
--- a/tests/examples/text/gears/cache/another_cache.lua
+++ b/tests/examples/text/gears/cache/another_cache.lua
@@ -1,0 +1,21 @@
+--DOC_GEN_OUTPUT --DOC_HIDE
+local gears = require("gears") --DOC_HIDE
+
+local function tostring_for_cache(obj)
+    return obj[1]
+end
+local counter = 0
+local wrapper_cache = gears.cache.new(function(arg)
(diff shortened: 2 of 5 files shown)


psychon:
Basically :+1:. However, you need to squash your two commits into one to make Travis happy.

Plus, the doc example might be a bit misleading since the function does not actually compute anything. How about adding something like `return test * 42 -- Pretend this is some expensive computation` and also printing the return value in the caller?
Downside: This makes it more complicated to figure out what is going on.

What do you think?

SethBarberee:
> Plus, the doc example might be a bit misleading since the function does not actually compute anything. How about adding something like `return test * 42 -- Pretend this is some expensive computation` and also printing the return value in the caller?
> Downside: This makes it more complicated to figure out what is going on.

Would you be opposed to this becoming another example? I'm always in favor of more examples :)

actionless:
what about?

```lua

local test_cache = cache.new(function(test)
    -- let's just print about what we created
    print("new entry created with value " .. test)
    return test * 42 -- Pretend this is some expensive computation
end)

print(print(test_cache:get(0))
print(test_cache:get(1))
print(test_cache:get(2))
print(test_cache:get(1)) -- no message since it exists
```

and result:

```
new entry created with value 0
0
new entry created with value 1
42
new entry created with value 2
84
42
```

psychon:
@actionless Yup, that's what I had in mind.

For "more is always better", I can submit another contestant:
```lua
local function tostring_for_cache(obj)
    return obj[1]
end
local counter = 0
local wrapper_cache = gears.cache.new(function(arg)
    local kind = "cache object #" .. tostring(counter) .. " for " .. tostring(arg)
    counter = counter + 1
    return setmetatable({ kind }, { __tostring = tostring_for_cache })
end)
print(wrapper_cache:get("first"))
print(wrapper_cache:get("second"))
print(wrapper_cache:get("first")) -- No new object since it already exists
collectgarbage("collect") -- The GC can *always* clear the cache
print(wrapper_cache:get("first"))
print(wrapper_cache:get("second"))
```
```
cache object #0 for first
cache object #1 for second
cache object #0 for first
cache object #2 for first
cache object #3 for second
```

SethBarberee:
I can add your example, @psychon and update my current example with what @actionless provided. I feel like both of those show the functionality of the module pretty well.

SethBarberee:
Docs aren't coming out pretty though. I have a feeling that it's a formatting thing but not sure.

PR #2966, opened 2020-01-16
Fix firefox example in awful.permissions

SethBarberee:
Was trying out the example and it didn't work until I removed the context string.


SethBarberee:
Would like to figure out another example that can show the purpose of the context part but need more time to mess around with this API.

Aire-One:
@mergifyio rebase

actionless (approved):
thanks!

PR #2970, opened 2020-01-19
Fix prompt navigation

laith-m0:
1. Open any awful.prompt(mod+r or mod+x)
2. Type ()->
3. Press alt+b two times


Elv13:
And in goes commit 9999.

This took some time to be found, eww, that code is old. Thanks for the patch.

PR #2975, opened 2020-01-25
Fix awful.mouse.move/awful.mouse.resize

### LINKED ISSUE #2971: awful.mouse.wibox.move is broken

awesome v4.3-655-g67672bf5 (Too long)
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

Create a wibox with awful.mouse.wibox.move binded to button, then try to drag it:

```
wibox1=wibox{visible=true,height=100,width=200,x=500,y=500}
wibox1.buttons=gears.table.join(wibox1.buttons,
    awful.button({}, 1, function() awful.mouse.wibox.move() end))
```

**Actual result:**

Nothing happens. When callback changed to `function() awful.mouse.wibox.move(wibox1) end` on drag cursor will change, but the wibox will only be dragged corner snaps.

**Expected result:**

Wibox should be dragged like windows are dragged with modkey pressed.
The problem is `awful.mouse.resize()` that updates geometry by emitting signal "request::geometry" which works for client and does not work for wibox.

As a workaround it can be fixed by changing `client:emit_signal(req, context, geo)` to  `client:geometry(geo)` but consequences are unclear to me.

Elv13:
These methods have been broken for a while. There is tests actually calling them, but it doesn't check that they actually worked. There was recent change to this code and I assumed I regressed something, but no, it was just already broken.

Fix #2971

### DIFF
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -399,6 +399,32 @@ end
 -- @param wibox
 -- @method draw
 
+--- Connect a global signal on the wibox class.
+--
+-- Functions connected to this signal source will be executed when any
+-- wibox object emits the signal.
+--
+-- It is also used for some generic wibox signals such as
+-- `request::geometry`.
+--
+-- @tparam string name The name of the signal
+-- @tparam function func The function to attach
+-- @staticfct wibox.connect_signal
+-- @usage wibox.connect_signal("added", function(notif)
+--    -- do something
+-- end)
+
+--- Emit a wibox signal.
+-- @tparam string name The signal name.
+-- @param ... The signal callback arguments
+-- @staticfct wibox.emit_signal
+
+--- Disconnect a signal from a source.
+-- @tparam string name The name of the signal
+-- @tparam function func The attached function
+-- @staticfct wibox.disconnect_signal
+-- @treturn boolean If the disconnection was successful
+
 function wibox.mt:__call(...)
     return new(...)
 end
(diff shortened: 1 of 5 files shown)


lib/wibox/init.lua
```diff
@@ -310,8 +310,9 @@ local function new(args)
         ret._drawable:_inform_visible(w.visible)
     end)
 
+    --TODO v5 deprecate this and use `wibox.object`.
     for k, v in pairs(wibox) do
-        if type(v) == "function" then
+        if (not rawget(ret, k)) and type(v) == "function" then
```
psychon: I'm curious: Which functions are overwritten here? In my local git checkout (which is not up to date), this change makes no difference.

psychon:
I can't say that I really understand what is going on, but at least I can say: I read through this and didn't really find comments to leave. So, feel free to count that as :+1:.

Travis however says that something is broken:
```
Checking lib/awful/permissions/init.lua           1 warning
    lib/awful/permissions/init.lua:408:26: accessing undefined variable 'c'
```

lib/wibox/init.lua
Elv13: `gears.object.connect_signal` (method) was replaced by `wibox.connect_signal` (global class level function)

Elv13:
> Travis however says that something is broken:

Done

day4me:
did you try to use it? because i tried to use awful.mouse.wibox.move(), but it still doesn't work
`todo = wibox({width=400, height=400, visible = false, type = "normal", screen = s})
--awful.placement.maximize(todo)

todo:buttons(gears.table.join(awful.button({}, 1, function () awful.mouse.wibox.move() end)))`

PR #2982, opened 2020-02-03
Support shebangs (#!), modelines, API levels and add more doc.

Elv13:
I am finally mostly happy with this one. I wanted to do that for a long time.

Since I broke the behavior and @actionless noticed, I had no real choice but to add API levels now rather than after v4.4 release. I cannot really backtrack from the change I made since it would break some features down the line. So the default is restored for API level 4, but the change will take effect for 5. There is no feature regression, it is still possible to disable sloppy focus.

So now we have API levels like Android and iOS. We can change the behavior and hide it behind a flag. It also satisfy the feature request we had a while ago to add an easy to to get the pending deprecation displayed in the logs.

Since API level affects thing early on, it had to be set before `rc.lua` is loaded. To do this I parse the config file and look for options. There was multiple glib ways to implement this, but they had downside and I rolled my own state machine parser. It doesn't support escaping, quoting and UTF-8. This could be bolted on later, but this PR is large enough and the API level is an integer, no need to over-do the parser.

Since I was adding a parser, I chose to implement the shebang mode. This means you can now `chmod +x` and Lua script and use AwesomeWM to interpret it.

```lua
#! /usr/bin/env awesome --replace
print("executed with awesome, ha ha, I killed your WM")
```

The code is fully shared between the command line, modeline and shebang modes. That's a plus for maintainability, but I admit the thing is downright magic. If you don't understand state machines, this looks *really* weird. I usually uses vtable matrices to implement multi-dimensional parser, but that would have made it too magic for most readers. So it's just embedded `switch` statements.

Finally, I added documentation for the command line. We already had `--help` and a manpage, but now it's in the main doc too.

### DIFF
--- a/lib/awful/permissions/init.lua
+++ b/lib/awful/permissions/init.lua
@@ -740,9 +740,14 @@ screen.connect_signal("property::workarea", function(s)
 end)
 
 -- Enable sloppy focus, so that focus follows mouse.
-client.connect_signal("mouse::enter", function(c)
-    c:emit_signal("request::autoactivate", "mouse_enter", {raise=false})
-end)
+if awesome.api_level > 4 then
+    --TODO v5: Remove the code from `rc.lua`. It cannot be done yet because we
+    -- cannot know if the user removed it to disable sloppy focus or because
+    -- they want to use the permissions to manage it.
+    client.connect_signal("mouse::enter", function(c)
+        c:emit_signal("request::autoactivate", "mouse_enter", {raise=false})
+    end)
+end
 
 return permissions
 
--- a/options.c
+++ b/options.c
@@ -360,7 +360,7 @@ exit_help(int exit_code)
   -v, --version          show version\n\
   -c, --config FILE      configuration file to use\n\
   -f, --force            ignore modelines and apply the command line arguments\n\
-      --search DIR       add a directory to the library search path\n\
+  -s, --search DIR       add a directory to the library search path\n\
   -k, --check            check configuration file syntax\n\
   -a, --no-argb          disable client transparency support\n\
   -l  --api-level LEVEL  select a different API support level than the current version \n\
--- a/options.c
+++ b/options.c
@@ -428,9 +428,11 @@ options_check_args(int argc, char **argv, int *init_flags, string_array_t *paths
 
             break;
           case 's':
+            globalconf.have_searchpaths = true;
             string_array_append(paths, a_strdup(optarg));
             break;
           case 'a':
+            globalconf.had_overriden_depth = true;
             (*init_flags) &= ~INIT_FLAG_ARGB;
             break;
           case 'r':
--- a/options.c
+++ b/options.c
@@ -37,6 +37,7 @@ exit_help(int exit_code)
 "Usage: awesome [OPTION]\n\
   -h, --help           show help\n\
   -v, --version        show version\n\
+  -f, --force          ignore modelines and apply the command line arguments\n\
   -c, --config FILE    configuration file to use\n\
       --search DIR     add a directory to the library search path\n\
   -k, --check          check configuration file syntax\n\
@@ -58,6 +59,7 @@ options_check_args(int argc, char **argv, int *init_flags, string_array_t *paths
         { "help",    NO_ARG, NULL, 'h'  },
         { "version", NO_ARG, NULL, 'v'  },
         { "config",  ARG   , NULL, 'c'  },
+        { "force" ,  NO_ARG, NULL, 'f'  },
         { "check",   NO_ARG, NULL, 'k'  },
         { "search",  ARG   , NULL, 's'  },
         { "no-argb", NO_ARG, NULL, 'a'  },
@@ -80,6 +82,9 @@ options_check_args(int argc, char **argv, int *init_flags, string_array_t *paths
           case 'h':
             if (! ((*init_flags) & INIT_FLAG_ALLOW_FALLBACK))
                 exit_help(EXIT_SUCCESS);
+            break;
+          case 'f':
+            (*init_flags) |= INIT_FLAG_FORCE_CMD_ARGS;
             break;
           case 'k':
             (*init_flags) |= INIT_FLAG_RUN_TEST;
--- a/options.c
+++ b/options.c
@@ -24,6 +24,7 @@
 
 #include <unistd.h>
 #include <stdio.h>
+#include <sys/stat.h>
 #include <getopt.h>
 
 #define KEY_VALUE_BUF_MAX 64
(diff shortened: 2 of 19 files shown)


lib/awful/permissions/init.lua
```diff
@@ -723,9 +723,14 @@ screen.connect_signal("property::workarea", function(s)
...
 
 -- Enable sloppy focus, so that focus follows mouse.
-client.connect_signal("mouse::enter", function(c)
-    c:emit_signal("request::autoactivate", "mouse_enter", {raise=false})
-end)
+if awesome.api_level > 4 then
+    --TODO v5: Remove the code from `rc.lua`. It cannot be done yet because we
+    -- cannot know if the user removed it to disable sloppy focus or because
+    -- they want to use the permissions to manage it.
+    client.connect_signal("mouse::enter", function(c)
```
actionless: but it seems it will still require changing to smth like

```lua
function permission.sloppy_focus(c)
    c:emit_signal("request::autoactivate", "mouse_enter", {raise=false})
end

client.connect_signal("mouse::enter", permissions.sloppy_focus)
```

options.c
```diff
@@ -0,0 +1,454 @@
...
+    if (!fgets(file_buf, READ_BUF_MAX, fp)) {
+        fclose(fp);
+        return false;
+    }
+
+    unsigned char c;
+
+    /* Simple state machine to translate both modeline and shebang into argv */
+    for (int i = 0; (c = file_buf[i++]) != '\0';) {
+        /* Be very permissive, skip the unknown, UTF is not allowed */
```
actionless: mb print some warning about unknown options/characters?
could be helpful in case of typo or so

actionless:
generally cool idea with api levels :+1:

btw is there already a plan on maintaining some specific amount of api levels/strategy on minor major versions/etc?

options.c
Elv13: For unknown options, the reason I don't print them is a bit facepalm-ish. tl;rd, it is to let `#! /usr/bin/env awesome`  "just work" and not having to bother understanding that those 2 arguments should be ignored. Given that the executable can have any name and symlinks, correctly mapping it back to `argv[0]` is more code than this entire PR and I am very lazy.

lib/awful/permissions/init.lua
Elv13: I don't think this is the best way forward. Disconnecting `request::` signals has caused some bugs report in the past (ie, broke maximization). It is error prone and hard to future-proof. The current way to disable this is

    c:deny("autofocus", "mouse_enter")

Once I am done with the notification PR, the follow-up will add this to `ruled.client`:

```lua
ruled.client.append_rule {
  [...]
  properties = {
    denied_permissions = {
      autofocus = {
        "mouse_enter"
      }
    }
  }
```

The doc will be very explicit about the permissions and the contexts (they are already listed in "show more", but I have some code to auto-generate the index/table_of_content for them). The permissions framework will work on all objects (client, notification, tag, screen, clipboard).

That's my vision for this API going forward. I think it will be great and easier to understand than the current `disconnect_signal` thing.

Elv13:
> btw is there already a plan on maintaining some specific amount of api levels/strategy on minor major versions/etc?

I think it would be too painful for the average users. It's great we have a way to change the behavior, but I don't plan to abuse it. Breaking stuff should be done only when there is no way on earth I can pull off a compat mode. I went *really* far for the `.key = {}` vs. `:key(gears.table.join())` compat. I so far failed to fix `.geometry = {}` vs. `:geometry({})`, but I think it can be done with enough effort.

The HiDPI work will definitely be a major source of behavior change. All the code which assume a pixel is the base unit will break. That includes everything calling `xresources.dpi()`. The stateful client layout system will be another pain in the ass for user switching to AwesomeWM v5.0 from v4.x.

I want to avoid those breakages as much as I can. I don't plan to have API level changes outside of major versions. Nor do I plan making major releases more often. Our users have invested a great deal of effort into their config, I respect it and wont break them.

lib/awful/permissions/init.lua
actionless: > The current way to disable this is
> `c:deny("autofocus", "mouse_enter")`

did i missed smth or it isn't documented yet?

Elv13: > did i missed smth or it isn't documented yet?

It's there:
https://awesomewm.org/apidoc/core_components/client.html#deny

If you search some pages for "1 permission" and click "show more" on these entries, you will see the table of permissions. There will be much better doc/examples in the PR that adds it to the rules.

actionless: i mean it would be better to mention the sloppy focus case explicitly, mb in FAQ section, or in release_notes for 5.x

actionless: also i've just tested it with git master,

```lua
  awful.rules.rules = {
      { rule = { },
        properties = {
          focus = awful.client.focus.filter,
          raise = true,
          keys = awesome_context.clientkeys,
          buttons = awesome_context.clientbuttons,
          placement =
            awful.placement.centered +
            awful.placement.no_overlap +
            awful.placement.no_offscreen,
          size_hints_honor = false,
          screen = awful.screen.preferred,
        },
        callback = function(c)
          c:deny("autofocus", "mouse_enter")
          awful.client.setslave(c)
          apply_delayed_rule(c)
        end
      },
```

adding `c:deny("autofocus", "mouse_enter")` doesn't disable the sloppy focus

actionless: after a small investigation it seems you meant `c:deny("autoactivate", "mouse_enter")`?

Aire-One:
API levels is an interesting addition. It'll make API upgrade more reliable and will really help with all those deprecated API.

I however can't fully understand the shebang stuff.

From my first look, I guessed we could write scripts that awesome executes. Something similar to what we have with the `awesome-client` tool. Reading the code showed shebangs are used to interpret argv arguments at awesome startup. So they can only be used by the main `rc.lua` in a replacement to command line arguments.

Did I missed something or is there another chunk not yet integrated in?

Elv13:
> Did I missed something or is there another chunk not yet integrated in?

Not for now. Eventually someone could add the ability to disable reparenting and claiming to be the window manager. That way some people could use `naughty` or `wibox` in other window manager. Another use case would be to split some configs into different process. As of now, I don't have patches to do this. The last use case is people who want to do "distributions" of AwesomeWM but still rely on the system binary. With this magic, it's easier to create a session file.

Elv13:
This is ready to merge. I will do it next week unless someone wants to review the code more deeply and find issues,

lib/awful/permissions/init.lua
Elv13: >  it seems you meant c:deny("autoactivate", "mouse_enter")?

Yes, sorry about the naming mistake in my previous comment.

Elv13:
> From my first look, I guessed we could write scripts that awesome executes. Something similar to what we have with the awesome-client tool.

That's already supported (kind of)

```lua
#! /bin/bash
awesome-client << EOF

require("naughty").notification {
    message = "foo"
}

print("This will print in AwesomeWM stdout!")

EOF
```

Someone with a lot of time could improve `awesome-client` to allow local variables to work properly. Maybe tab-completion, syntax highlight and python-like instant feedback for return values would be cool too. Right now `awesome-client` isn't very user friendly.

actionless:
guess this require squashing to fix PREV_COMMITS test?

Elv13:
> guess this require squashing to fix PREV_COMMITS test?

No, its a false positive. If anyone can reproduce that error, please fix it. I tried to add some `awesome.sync()` around it a long time ago but could not make it deterministic enough to pass the placement test.

Issue #2851, opened 2019-08-11
Notification Icon isn't replaced


SethBarberee:
@Elv13, sorry I keep hitting you with naughty bugs :smile: 

Output of `awesome --version`:

```
awesome v4.3-350-ged091838 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • xcb-errors support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

**How to reproduce the issue:**

Trigger a new notification that uses icons (used spotify since it shows album art on each new song) during when the first notifcation is displayed. The text is replaced but icon is not replaced.

NOTE: When first notification is destroyed, the icon is properly set. This does not happen when the first notification is still displayed.

**Actual result:**

Album Art (Icon) from first notification is still there while text is updated.

**Expected result:**

Album art (icon) for the recent notification should be the icon and properly replace just like the text.

day4me:
+1, but in 
```
awesome v4.3 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```
everything is ok))

SethBarberee:
This has been fixed so closing this for now

actionless:
i still can reproduce it in a bit more limited test case:

if content of the image file changes
then setting `notification.icon` to the same filename doesn't update the icon

psychon:
Should https://github.com/awesomeWM/awesome/blob/87928befe2dc981bfdbca59c0474c0ff5d39544f/lib/naughty/widget/icon.lua#L93 and https://github.com/awesomeWM/awesome/blob/87928befe2dc981bfdbca59c0474c0ff5d39544f/lib/naughty/widget/icon.lua#L160 use `load_uncached_silently`?

actionless:
mb cache should check the filesize, not only filename?

actionless:
yup

Aire-One:
The initial problem reported here is resolved. @actionless, I think the newer issue discussed is not a regression compared to v4.3. So maybe we should close this issue to free up bandwidth in the v4.4 milestone, and refer to PR https://github.com/awesomeWM/awesome/pull/3883 to continue the discussion.

actionless:
>  is resolved.

but not yet merged:

[...]

PR #2988, opened 2020-02-10
Fix multiple notifications issue

Elv13:
Originally, I was planning to get the notification rules out. I know many users are awaiting this and I get multiple request/queries about it. Sorry for the delays.

The fact is, the notification rules make it too easy to push the notification system to its limit. When you do so, bugs appear. I would rather fix those before too many people run into them.

This pull request backs down on one of my original goal. I originally wanted notification objects to be decoupled from the screen. That coupling was pushed to the `naughty.layout.box` popup. From a design PoV, it made perfect sense, but not in practice. Removing screens created a reference mess and they ended up leaking. So this PR brings back harsher coupling with the screens. There is now an extra test suite just for this. It still fails and part of it is commented, but it is much closer to work than without all of these commits. The tests also uncovered a bug in `awful.placement`.

I also cherry picked some other commits from the branch like the better documentation. There is also a bunch of commits to fix previously undetected usage of deprecated APIs in the code. There is also a new `gears.geometry` function+tests. It isn't used in the code, but is used in the test and was generic enough to merge.

This will get merged next weekend if I can get the tests to pass in time.

### DIFF
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -1533,7 +1533,7 @@ function placement.next_to(d, args)
 
             geo.width, geo.height = dgeo.width, dgeo.height
 
-            fit = fit_in_bounding(pos.region.screen, geo, args)
+            fit = fit_in_bounding(pos.region, geo, args)
 
             if fit then break end
         end
--- a/lib/awful/popup.lua
+++ b/lib/awful/popup.lua
@@ -21,7 +21,7 @@
 -- @popupmod awful.popup
 ---------------------------------------------------------------------------
 local wibox     = require( "wibox"           )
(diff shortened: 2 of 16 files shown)


SethBarberee:
Is this a go for testing? Wanted to make sure it fixed #2851

Elv13:
>  Is this a go for testing? Wanted to make sure it fixed #2851

It does not, sorry. I need to install Ubuntu in a VM to fix that one. The Spotify client refuse to make the "right" kind of notifications on my Gentoo machines. Same for Google Chrome, so I guess all Electron related apps somehow detect that the older fallback notifications must be used. This PR is more about fixing the multi-screen issues, which I *can* reproduce.

PR #2994, opened 2020-02-16
Add keygroups, abbreviate cheatsheet, add ways to select a layout directly

ArenaL5:
The first commit is just a quick fix to avoid critical errors every time I'm testing `master`. It doesn't seem to break anything.

The second commit revolves around adding a new method to change layouts: besides pressing Super+Space and cycling through the layouts in the main list, one can choose directly a layout by pressing Super and a key in the numeric keypad.

There's a mnemonic method for most keys:

- 1, 7, 9 and 3 lay out the master zone on the corresponding corner (so they set the layout to `corner.sw`, `corner.nw`, `corner.ne` and `corner.se`, respectively).
- 8, 4, 2, 6 lay out the master zone on the corresponding side (so they choose `tile.bottom`, `tile`, `tile.right` and `tile.top`.
- 5 is `magnifier`, with the master zone on the middle.
- 0 is "zero gravity", i.e. `floating`.
- Return is `max.fullscreen` because it's one of the most promiment keys
- The decimal dot besides `max.fullscreen` is `max`.

The third commit adds a method to abbreviate long rows of keys with the same shortcut (F1/F2/F3/F4/F5/F6/F7 would contract to F1…F7) and adds some more aliases for keys. There's still a problem in that keycodes `#10` to `#21` will be printed as `1234567890'=` despite the current keyboard layout (like in Spain, where `#21` maps to `¡`). A function like xev's XLookupString would be needed to print the actual character.

I tested this code manually, but I expect to rewrite some code to improve some minor things. I'd appreciate any suggestion.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -372,6 +372,33 @@ awful.keyboard.append_global_keybindings({
                 end
             end
         end,
+    },
+    awful.key {
+        modifiers   = { modkey },
+        keygroup    = "numpad",
+        description = "select layout directly",
+        group       = "layout",
+        on_press    = function (index)
+            local everylayout = {
+                KP_End      = awful.layout.suit.corner.sw,
+                KP_Down     = awful.layout.suit.tile.top,
+                KP_Next     = awful.layout.suit.corner.se,
+                KP_Left     = awful.layout.suit.tile,
+                KP_Begin    = awful.layout.suit.magnifier,
+                KP_Right    = awful.layout.suit.tile.left,
+                KP_Home     = awful.layout.suit.corner.nw,
+                KP_Up       = awful.layout.suit.tile.bottom,
+                KP_Prior    = awful.layout.suit.corner.ne,
+                KP_Insert   = awful.layout.suit.floating,
+                KP_Delete   = awful.layout.suit.max,
+                KP_Divide   = awful.layout.suit.fair,
+                KP_Multiply = awful.layout.suit.spiral,
+                KP_Subtract = awful.layout.suit.fair.horizontal,
+                KP_Add      = awful.layout.suit.spiral.dwindle,
+                KP_Enter    = awful.layout.suit.max.fullscreen
+            }
+            awful.layout.set(everylayout[index])
+        end,
     }
 })
 
--- a/lib/awful/hotkeys_popup/keys/firefox.lua
+++ b/lib/awful/hotkeys_popup/keys/firefox.lua
@@ -19,7 +19,7 @@ local firefox_keys = {
     ["Firefox: tabs"] = {{
         modifiers = { "Mod1" },
         keys = {
-            ["1..9"] = "go to tab"
+            ["1…9"] = "go to tab"
         }
     }, {
         modifiers = { "Ctrl" },
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -55,7 +55,7 @@ local widget = {
 widget.hide_without_description = true
 
 --- Merge hotkey records into one if they have the same modifiers and
--- description.
+-- description. Records with five or more keys will abbreviate them.
 -- @tfield boolean widget.merge_duplicates
 -- @param boolean
 widget.merge_duplicates = true
@@ -110,7 +110,8 @@ widget.merge_duplicates = true
 -- @tparam[opt] table args Configuration options for the widget.
 -- @tparam[opt] boolean args.hide_without_description Don't show hotkeys without descriptions.
 -- @tparam[opt] boolean args.merge_duplicates Merge hotkey records into one if
--- they have the same modifiers and description.
+-- they have the same modifiers and description. Records with five keys or more
+-- will abbreviate them.
 -- @tparam[opt] int args.width Widget width.
 -- @tparam[opt] int args.height Widget height.
 -- @tparam[opt] color args.bg Widget background color.
@@ -141,45 +142,63 @@ function widget.new(args)
         ) and widget.merge_duplicates or args.merge_duplicates,
         group_rules = args.group_rules or gtable.clone(widget.group_rules),
         labels = args.labels or {
-            Mod4="Super",
-            Mod1="Alt",
-            Escape="Esc",
-            Insert="Ins",
-            Delete="Del",
-            Backspace="BackSpc",
-            Return="Enter",
-            Next="PgDn",
-            Prior="PgUp",
-            ['#108']="Alt Gr",
-            Left='←',
-            Up='↑',
-            Right='→',
-            Down='↓',
-            ['#67']="F1",
-            ['#68']="F2",
-            ['#69']="F3",
-            ['#70']="F4",
-            ['#71']="F5",
-            ['#72']="F6",
-            ['#73']="F7",
-            ['#74']="F8",
-            ['#75']="F9",
-            ['#76']="F10",
-            ['#95']="F11",
-            ['#96']="F12",
-            ['#10']="1",
-            ['#11']="2",
-            ['#12']="3",
-            ['#13']="4",
-            ['#14']="5",
-            ['#15']="6",
-            ['#16']="7",
-            ['#17']="8",
-            ['#18']="9",
-            ['#19']="0",
-            ['#20']="-",
-            ['#21']="=",
-            Control="Ctrl"
+            Mod4       ="Super",
+            Mod1       ="Alt",
+            Escape     ="Esc",
+            Insert     ="Ins",
+            Delete     ="Del",
+            Backspace  ="BackSpc",
+            Return     ="Enter",
+            Next       ="PgDn",
+            Prior      ="PgUp",
+            ['#108']   ="Alt Gr",
+            Left       ="←",
+            Up         ="↑",
+            Right      ="→",
+            Down       ="↓",
+            ['#67']    ="F1",
+            ['#68']    ="F2",
+            ['#69']    ="F3",
+            ['#70']    ="F4",
+            ['#71']    ="F5",
+            ['#72']    ="F6",
+            ['#73']    ="F7",
+            ['#74']    ="F8",
+            ['#75']    ="F9",
+            ['#76']    ="F10",
+            ['#95']    ="F11",
+            ['#96']    ="F12",
+            ['#10']    ="1",
+            ['#11']    ="2",
+            ['#12']    ="3",
+            ['#13']    ="4",
+            ['#14']    ="5",
+            ['#15']    ="6",
+            ['#16']    ="7",
+            ['#17']    ="8",
+            ['#18']    ="9",
+            ['#19']    ="0",
+            ['#20']    ="-",
+            ['#21']    ="=",
+            [' ']      ="space",
+            Control    ="Ctrl",
+            KP_End     ="Num1",
+            KP_Down    ="Num2",
+            KP_Next    ="Num3",
+            KP_Left    ="Num4",
+            KP_Begin   ="Num4",
+            KP_Right   ="Num6",
+            KP_Home    ="Num7",
+            KP_Up      ="Num8",
+            KP_Prior   ="Num9",
+            KP_Insert  ="Num0",
+            KP_Delete  ="Num.",
+            KP_Divide  ="Num⁄", -- a fraction slash, to avoid confusion with
+                                -- the solidus, which separates keys.
+            KP_Multiply="Num*",
+            KP_Subtract="Num-",
+            KP_Add     ="Num+",
+            KP_Enter   ="NumEnter",
         },
         _additional_hotkeys = {},
         _cached_wiboxes = {},
@@ -187,7 +206,15 @@ function widget.new(args)
         _colors_counter = {},
         _group_list = {},
         _widget_settings_loaded = false,
+        _keygroups = {},
     }
+    for k, v in pairs(awful.key.keygroups) do
+        widget_instance._keygroups[k] = {}
+        for k2, v2 in pairs(v) do
+            widget_instance._keygroups[k][k2] = widget_instance.labels[v2[1]] or v2[1]
+        end
+        widget_instance._keygroups[k]["text"] = table.concat(widget_instance._keygroups[k], "/")
+    end
 
 
     function widget_instance:_load_widget_settings()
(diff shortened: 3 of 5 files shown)


actionless:
> The first commit is just a quick fix to avoid critical errors every time I'm testing master. 

does that happens with default rc.lua?

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -285,6 +311,52 @@ function widget.new(args)
...
+                else
+                    return {
+                        key = value.key,
+                        mod = value.mod,
+                        description = value.description
+                    }
+                end
+            end, keys)
+        end
+        self._cached_awful_keys = clone
```
actionless: why it taking `keys=self._cached_awful_keys` as input argument if anyway further here it's explicitly saving the result to `self._cached_awful_keys`?

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -295,6 +367,9 @@ function widget.new(args)
             end
         end
         self:_sort_hotkeys(self._cached_awful_keys)
+        if self.merge_duplicates then
+            self:_abbreviate_hotkeys(self._cached_awful_keys)
```
actionless: why it's called from inside `_import_awful_keys()`?

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -285,6 +311,52 @@ function widget.new(args)
     end
 
 
+    function widget_instance:_abbreviate_hotkeys(target)
+        -- This is a very crude method and only useful for keygroups.
+        -- It compares the current key value without modifiers, as it would be
+        -- printed (e.g 2/3/4/5/6) with the string that the entire
```
actionless: what about the case with 2/3/6/7?

ArenaL5:
> > The first commit is just a quick fix to avoid critical errors every time I'm testing master.
> 
> does that happens with default rc.lua?

Yes. I've never had a custom rc.lua in my home folder on this computer, and I deleted every file from awesome (even at `/etc`) before testing master.

lib/awful/hotkeys_popup/widget.lua
actionless: or if they keys are not sorted

lib/awful/hotkeys_popup/widget.lua
ArenaL5: There's no real reason for this, I admit. I copied a subroutine of yours that uses an input argument, but couldn't find a way to get the variables to update until I saved explicitly to `self._cached_awful_keys`. Then I forgot about the input argument.

I've done a lot of tweaks in other computer with `rc.lua`, but I have very little experience in actual Lua programming.

awesomerc.lua
```diff
@@ -372,6 +372,33 @@ awful.keyboard.append_global_keybindings({
...
+                KP_Insert   = awful.layout.suit.floating,
+                KP_Delete   = awful.layout.suit.max,
+                KP_Divide   = awful.layout.suit.fair,
+                KP_Multiply = awful.layout.suit.spiral,
+                KP_Subtract = awful.layout.suit.fair.horizontal,
+                KP_Add      = awful.layout.suit.spiral.dwindle,
+                KP_Enter    = awful.layout.suit.max.fullscreen
+            }
+            awful.layout.set(everylayout[index])
+        end,
```
actionless: don't have numpad on any of my keyboards but nice idea

lib/awful/hotkeys_popup/widget.lua
ArenaL5: I put it after `_sort_hotkeys` because I supposed that's where you start processing the keys. A bad judgment call on my part, I guess.

I was looking for any method that's called before joining the texts, coloring the texts and laying out the columns is fine.

lib/awful/hotkeys_popup/widget.lua
ArenaL5: It won't be abbreviated. It will remain 2/3/6/7.

This code is really supposed to complement `awful.key` bindings with `keygroup`, which naturally creates bindings with the same description. 2/3/6/7 can only happen if the user individually creates bindings for 2, 3, 6, 7 while using the same description... which I don't think that happens very often.

It won't work with 7/6/5/4/3/2, either. Anyways, if it doesn't find a match, it won't try abbreviating them. No information is lost.

lib/awful/hotkeys_popup/widget.lua
actionless: > I put it after _sort_hotkeys
i understand the logic, but in that case `_sort_hotkeys` is called from two different places, and with different target for sorting, i suppose what abbreviating should happen in the same way

awesomerc.lua
ArenaL5: Thank you.

I thought every keyboard had one. Doesn't your keyboard have a Fn key, or something similar? My old laptop had the numeric keypad bound to the keys 789UIOJKLM.

actionless:
> Yes. I've never had a custom rc.lua in my home folder on this computer, and I deleted every file from awesome (even at /etc) before testing master.

that's interesting, could you please provide the full stack trace and `awesome -v` output, as i can't reproduce it

awesomerc.lua
actionless: such keyboards are called "tenkeyless/TKL" keayboards

no, laptop doesn't have Fn-enabled numpad

lib/awful/hotkeys_popup/widget.lua
actionless: > 2/3/6/7 can only happen if the user individually creates bindings for 2, 3, 6, 7 while using the same description... which I don't think that happens very often.

but it's not handled here as far as i can see from the function (or guide me to the exact line where it is, if i am mistaken)

ArenaL5:
...I'm sorry but I've clearly made a mistake, I can't reproduce it now. Obviously commit 93adb510f491373e09330c7665037bf84e006b68 has no reason to be anymore.

My best guess to why did I find that bug is that I started modifying the code a few days ago, and my commits were originally based on af4b4a254f268b57a0fe2e5e248c51e145c5a08c, if I'm not mistaken. I rebased my branch just before writing this pull request and didn't think of testing the latest master, as you didn't modify any of the files I was working on.

The original error was a couple of Lua errors were printed on the screen as critical notifications, one when unfocusing a window and another when focusing a new window, because it expected a string instead of a nil.

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -141,53 +141,79 @@ function widget.new(args)
...
         _group_list = {},
         _widget_settings_loaded = false,
+        _keygroups = {},
     }
+    for k, v in pairs(awful.key.keygroups) do
+        widget_instance._keygroups[k] = {}
+        for k2, v2 in pairs(v) do
+            widget_instance._keygroups[k][k2] = widget_instance.labels[v2[1]] or v2[1]
+        end
+        widget_instance._keygroups[k]["text"] = table.concat(widget_instance._keygroups[k], "/")
```
actionless: why you storing keys as concatted with '/' instead of storing them as a table?

that would avoid false-positive matching in `first = string.find(v.text, value.key, 1, true)` next in your code

lib/awful/hotkeys_popup/widget.lua
actionless: > It won't work with 7/6/5/4/3/2, either.

also i can't see from the code where it skips them

actionless:
unless you can't reproduce the problem with props -- please rebase to exclude it from the PR

lib/awful/hotkeys_popup/widget.lua
ArenaL5: It's precisely because I found a text match easier to program than finding every key in a table.

There should be no real chance of false-positives. Abbreviations only trigger with five or more keys, `/` is only ever used to separate keys(*), and it doesn't matter if two keygroups share keys because it doesn't check for the name of the keygroup.

In your previous example, if `numrowA="1/2/3/4/5/6/7/8"` and `numrowB="2/3/4/5/6/7/8/9"`, the abbreviation's still going to be `2…6` regardless of which keygroup is tested first.

There _is_ a hypothetical false-positive, now you mention it, but this would only happen if the first/last key in the cheatsheet matches part of the name in a keygroup. For example, testing `2/3/4/5/6` against `Num2/3/4/5/6/7`. It still can be worked around with a regexp.

(*) I took care of defining the `/` key in the alias list as `Num⁄`, which doesn't use an actual `/` but a different Unicode character. I'm not sure if any keyboard has a dedicated `/` outside the numeric keypad (mine doesn't), but if that's the case it can be added to the aliases list, too.

ArenaL5:
> unless you can't reproduce the problem with client props -- please rebase to exclude it from the PR

Done.

lib/awful/hotkeys_popup/widget.lua
ArenaL5: The other call of `_sort_hotkeys` is for cheatsheets for external programs (the files under `lib/awful/hotkeys_popup/keys`).

Those cheatsheets weren't created by `awful.key`, they're hand-made, and they already were abbreviated by hand, too. No need to use `_abbreviate_hotkeys` there.

lib/awful/hotkeys_popup/widget.lua
actionless: example of false positive is `KP_Enter/...` will match `n`, `t`, `e`, etc

it must be a table

lib/awful/hotkeys_popup/widget.lua
actionless: that should be then explicitly mentioned in docs for `.merge_duplicates`

lib/awful/hotkeys_popup/widget.lua
actionless: this should be fixed before it will be fine for merging

actionless:
also `luacheck` is giving 3 warnings which should be resolved: https://travis-ci.com/awesomeWM/awesome/jobs/287854778#L3354

lib/awful/hotkeys_popup/widget.lua
ArenaL5: That's the reason of the minimum. This method works `after` keys with the same description are merged, and counts the keys in the merged entries before trying to abbreviate anything.

In an entry like `Super+n/t`, `n/t` only contains two keys, so the subroutine aborts before trying to find a match. The same for `t`, `n`, `t/n/e` or even `n/t/e/r`, because the minimum is five.

And `n/t/e/r` isn't a subtring of `Enter`, so there's no match anyways.

ArenaL5:
> also `luacheck` is giving 3 warnings which should be resolved: https://travis-ci.com/awesomeWM/awesome/jobs/287854778#L3354

Okay. Two questions:

- One of the warnings is in line 331, where I wrote `function(k, v)` because `find_first_key` expects a two-argument function, even though `k` is never used. How do I fix that?
- For this kinds of edits, do you usually rebase/remake the commits, or can I just add new commits to this branch?

actionless:
1) use underscore
2) you can do separate commits and squash rebase when CI and reviews will be green

lib/awful/hotkeys_popup/widget.lua
actionless: i see now, it skips those cases by matching together with slashes

but that will also skip a valid cases like 1/3/2

ArenaL5: Yes, that's correct. Cases like 1/3/2/4/5 will be skipped because they're not in the correct order. (1/3/2 would be skipped anyways because it's only three keys.)

This only happens if users create bindings for each key separately, instead of using a keygroup. To fix it... I'd need to operate on tables.

lib/awful/hotkeys_popup/widget.lua
actionless: > It still can be worked around with a regexp.

that still enough of reason to refactor it to be a table

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -311,14 +311,14 @@ function widget.new(args)
     end
 
 
-    function widget_instance:_abbreviate_hotkeys(target)
+    function widget_instance:_abbreviate_hotkeys()
```
actionless: as its sorting only `awful` keys now i'd rename it into smth like _abbreviate_awful_keys()

lib/awful/hotkeys_popup/widget.lua
actionless: >  To fix it... I'd need to operate on tables.

yeah, that's exactly the point :+1:

lib/awful/hotkeys_popup/widget.lua
ArenaL5: I uploaded a patch that uses no regexps and no tables.

By the way, how can I use your static code check tools locally? I couldn't find at `CONTRIBUTING.md` how to replicate the tests done with Travis CI.

actionless: you can look for it in travis's config file accordingly

in this case it's `make check`

lib/awful/hotkeys_popup/widget.lua
ArenaL5: Okay, done.

ArenaL5:
I think that everything's covered now, please double-check.

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -141,53 +142,79 @@ function widget.new(args)
...
-            ['#14']="5",
-            ['#15']="6",
-            ['#16']="7",
-            ['#17']="8",
-            ['#18']="9",
-            ['#19']="0",
-            ['#20']="-",
-            ['#21']="=",
-            Control="Ctrl"
+            Mod4       ="Super",
```
Elv13: This should be ported to C, the map is incomplete and doesn't support everything. We already have such mechanism for the modkeys, see awesome._modifiers (`luaa.c` `static int luaA_get_modifiers(lua_State *L)`).

lib/awful/hotkeys_popup/keys/firefox.lua
```diff
@@ -19,7 +19,7 @@ local firefox_keys = {
     ["Firefox: tabs"] = {{
         modifiers = { "Mod1" },
         keys = {
-            ["1..9"] = "go to tab"
+            ["1…9"] = "go to tab"
```
Elv13: I think these changes should be in their own commit and own pull request. Some of what is proposed here might drag on since it is going to be rather controversial.

awesomerc.lua
```diff
@@ -372,6 +372,33 @@ awful.keyboard.append_global_keybindings({
                 end
             end
         end,
+    },
+    awful.key {
```
Elv13: The concept itself is fine, I did that for many years. They way this is implemented isn't really gonna work in practice. I mean, technically "works" for the default scenario, but many people don't use layouts that way. Also, when the new layout engine lands, without further abstraction layer(s), this will blow up.

### What are my concerns with this

* If the user doesn't have a numpad, this doesn't work. Most keyboard these days don't.
* It adds a lot of imperative code to `rc.lua` while we try very hard to remove all imperative code.
* Some international keyboard layouts have different key name.
* Numlock will mess this up.
* Individual tags can have their own layout list, this will conflict.
* The `awful.widget.layoutlist` is based either on the global set of layouts or the individual tag ones. If the user use that widget (for an alt-tab style popup or in the wibar), using this mechanism will cause a lot of fallback code to kick in. This fallback code has unwanted side effects, it should not be possible to enter these code paths using the default config.
* This bypasses the `append_default_layout`/`remove_default_layout`. The reason those exist is because they allow self-contained modules to add their own layouts. In the long term, I am removing all "write once" APIs in favor of read/write/append/clear ones. The reason for this is modularity.

Now, the better news: I agree we should add this feature, it is useful. However, if you want this merged, you will have to take the long way to get there.

### Expand the C code API to convert the keysym and keynames correctly in a way that universially works

We have to maintain this code going forward. History has proven time and again that if such hardcoding make its way into actual business logic (ie, not the hotkey popup, which is decoupled for the core), people will report bugs. This is why I added the C code for the modifiers, I did this hack and it blew up in my face.

This C code will need a minimal test suite. Don't worry too much about that, integration tests for such thing are really trivial to add (I don't mind doing it if you have no time).

Please do so in its own pull request, that's would be a rather nice and uncontroversial addition and pave the way for "what you really want".

### Use the current tag `layout` property instead of the array

As mentioned above, there is very valid reasons to not hardcode this in `rc.lua`.

* It is not compatible with self-contained modules
* It has side effects when the `layoutlist` widget is used
* It triggers some fallback code in `lib/awful/tag.lua`
* It disregard user defined tag list. Those list will soon be easier to setup with I merge `ruled.tag`.
* When the new layout module lands, people will start to use it and it will conflict with this hardcoded logic.

By using `awful.screen.focused().selected_tag.layouts[index]`, you could mitigate most of those concerns. It would also keep the size overhead to a minimum.

lib/awful/key.lua
```diff
@@ -298,20 +303,44 @@ local function new_common(mod, _keys, press, release, data)
...
+    -- Left: the keycode in a format which regular awful.key understands.
+    -- Right: the argument of the function ran upon executing the key binding.
     numrow = {},
     arrows = {
         {"Left"  , "Left"  },
         {"Right" , "Right" },
-        {"Top"   , "Top"   },
-        {"Bottom", "Bottom"},
-    }
+        {"Up"    , "Up"    },
```
Elv13: whoops, I should really have added unit tests for this. Can you make a PR just with this?

ArenaL5:
Sorry for the delay in answering, I had a couple of hectic days and couldn't find the time.

@actionless: "Strings' fortune telling" is a very cool nickname for my way of doing things. I will refactor the abbreviation method to work on a table, no problem.

@Elv13: I've no problem refactoring and splitting this PR, but, as you said, it might take a while. especially if I need to expand a C codebase I've never looked at. It will certainly solve one of the problems I pointed out (the mapping between keysyms and keycodes changes from layout to layout) Checking `awesome._modifiers` will be very helpful, thanks for the pointer.

Num Lock fortunately won't be a problem. Awesome is configured by default to ignore Num Lock altogether, and I already tested keycodes works either way.

I still need to study and try `awful.widget.layout`, as I'm not sure how do you want me to adapt my concept to that. I'll close this PR as it's clear it won't be merged like it is, but I might ask questions for here if I have any doubts. Talk to you all soon.

actionless:
@ArenaL5 thanks for understanding

in the meantime you could prepare PR for separate parts which already got positive reviews (see Elv's messages above)

lib/awful/key.lua
ArenaL5: Please check #2999.

lib/awful/hotkeys_popup/widget.lua
ArenaL5: Please check #3001.

awesomerc.lua
ArenaL5: Attempting to expand the API at #3011.

PR #3001, opened 2020-02-20
Abbreviate cheatsheet

ArenaL5:
Refactor of code in #2994 as @actionless requested. I also sent a second commit to add documentation to the newly exposed API (and tweak a couple sentences), please tell me back if I need to edit or remove it.

(By the way, the argument `keygroup` should be added to the documentation for both constructors `awful.key`, but I decided against doing it myself partly because the documentation for the whole page is a tad confusing and I'm afraid to break something.)

### DIFF
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -55,7 +55,7 @@ local widget = {
 widget.hide_without_description = true
 
 --- Merge hotkey records into one if they have the same modifiers and
--- description.
+-- description. Records with five or more keys will abbreviate them.
 -- @tfield boolean widget.merge_duplicates
 -- @param boolean
 widget.merge_duplicates = true
@@ -110,7 +110,8 @@ widget.merge_duplicates = true
 -- @tparam[opt] table args Configuration options for the widget.
 -- @tparam[opt] boolean args.hide_without_description Don't show hotkeys without descriptions.
 -- @tparam[opt] boolean args.merge_duplicates Merge hotkey records into one if
--- they have the same modifiers and description.
+-- they have the same modifiers and description. Records with five keys or more
+-- will abbreviate them.
 -- @tparam[opt] int args.width Widget width.
 -- @tparam[opt] int args.height Widget height.
 -- @tparam[opt] color args.bg Widget background color.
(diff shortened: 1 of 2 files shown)


lib/awful/hotkeys_popup/widget.lua
```diff
@@ -55,7 +55,7 @@ local widget = {
 widget.hide_without_description = true
 
 --- Merge hotkey records into one if they have the same modifiers and
--- description.
+-- description. Records with five or more keys will abbreviate them.
```
actionless: please mention what it's only for awful.keys but not manually added hotkeys for the apps

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -285,6 +295,53 @@ function widget.new(args)
...
+        -- applicable: they do not need this method.
+        for _, keys in pairs(self._cached_awful_keys) do
+            for _, params in pairs(keys) do
+                if #params.keylist > 4 then
+                    -- assuming here keygroups will never overlap;
+                    -- if they ever do, another for loop will be necessary:
+                    local keygroup = gtable.find_first_key(self._keygroups, function(_, v)
+                        return not not gtable.hasitem(v, params.keylist[1])
+                    end)
+                    local first, last, count, tally = nil, nil, 0, {}
```
actionless: i'm not sure, mb you could use `#tally` instead of `count`?

ArenaL5: It's not reliable. [The manual for 5.3 is verbose but pretty clear:](https://www.lua.org/manual/5.3/manual.html#3.4.7), `#table` may return any key `n` as long as:
* `n` ≥ 0 and integer,
* `table[n]` exists,
* and `table[n+1]` is nil.

If `tally[1]` doesn't exist (if, for example, `2/3/4/5/6` is checked against `numrow`) or there are gaps in the table, `#tally` has undefined behaviour. ([Lua 5.1](https://www.lua.org/manual/5.1/manual.html#2.5.5) explicitly says that if `tally[1]` is nil, `#tally` **may** be 0)

lib/awful/hotkeys_popup/widget.lua
ArenaL5: Oh, right, sorry.

Just pushed another commit. Is 2d3f7e1e94b60d34cadf3e5ec531eca6884f38c6 okay?

actionless: yup, thanks!

lib/awful/hotkeys_popup/widget.lua
actionless: yeah, it's indeed not good:

```lua
> t = {}
> #t
0
> t[2] = 123
> #t
0
> t[1] = 123
> #t
2
```

ArenaL5:
Thank you both.

PR #3011, opened 2020-02-28
Extending the C API to query the name of a key

ArenaL5:
This is another attempt at covering a part of closed PR #2994, addressing @Elv13's concerns about the names for the keys in `hotkeys.popup`. These commits are centered on `awesome.get_key_name()`, a function to query the keysym and the grapheme it stands for, if printable: `ñ` for n_tilde, etc. Documentation on `luaa.c`.

I had no prior experience with C and I think it shows at times. Again, please tell me if I need to change or tweak anything.

I tested this code both manually and with `tests/run.sh`, but it's inelegant to the extreme: in order to support querying symbols (for example, `awesome.get_key_name(".")`) I had to copypaste a function from `libxkbcommon` that's not been published in any standard API: https://github.com/xkbcommon/libxkbcommon/commit/0345aba082c83e9950f9dd8b7ea3bf91fe566a02. I also copied the table of keysyms because I couldn't figure out how to access the table at `/usr/lib`.

The good news is that all of this copypasted code is in its separate file, so it'll be trivial to delete when their function gets to the package repositories for every major distro.

There's still a Lua table for the more "controversial names" that `get_key_name` won't return: "space" (because `Ctrl+ ` is illegible), "Vol+", "Num8"...

No automated tests for this yet: is there any way to write tests with different locales? (Keycodes like `"#47"` will return different values for every keyboard layout.)

### DIFF
--- a/lib/awful/keyboard.lua
+++ b/lib/awful/keyboard.lua
@@ -197,6 +197,29 @@ function module.remove_client_keybinding(key)
     return false
 end
 
+--- Get X11 keysym and a one-character representation from an Awesome keycode.
+--
+-- A "one-character representation" is a single UTF-8 representing the typical
+-- output from that keysym in a text editor (e.g. " " for space, "ñ" for
+-- n_tilde, "Ā" for A_macron). It usually matches the main engraving of the key
+-- for level-0 symbols (but lowercase).
+--
+-- Keycodes may be given in a string in any valid format for `awful.key`:
+-- "#" + keycode, the symkey name and the UTF-8 representation will all work.
+--
+-- If no suitable keysym is found, or a malformed keycode is given as an
+-- argument, this function will return (nil, nil)
+--
+-- @treturn[1] string keysym The keysym name
+-- @treturn[1] nil keysym If no valid keysym is found
+-- @treturn[2] string printsymbol The xkb_keysym_to_utf8 result
+-- @treturn[2] nil printsymbol If the keysym has no printable representation.
+-- @staticfct awful.keyboard.get_key_name
+function module.get_key_name(key)
+    return capi.awesome._get_key_name(key)
+end
+
+
 capi.client.connect_signal("scanning", function()
     capi.client.emit_signal("request::default_keybindings", "context")
 end)
--- a/luaa.c
+++ b/luaa.c
@@ -72,6 +72,7 @@
 
 #include <xcb/xcb_atom.h>
 #include <xcb/xcb_aux.h>
+#include "xkb_utf32_to_keysym.c"
 
 #include <unistd.h> /* for gethostname() */
 
@@ -409,6 +410,126 @@ get_modifier_name(int map_index)
     }
 
     return 0; /* \0 */
+}
+
+// Helper function for luaA_get_key_name() below.
+// Will return the UTF-32 codepoint IF AND ONLY IF the input is exactly one
+// valid UTF-8 character. Otherwise, it will return zero.
+static uint32_t
+one_utf8_to_utf32(const char* input, const size_t length) {
+    // When casting a char to a uint32_t, the missing bits will all be set to
+    // the highest bit in the original char. Thus the exaggerated masking of
+    // every value (Unicode is defined around bytes)
+    if (length < 1 || length > 4)
+        return 0;
+    if (length == 1)
+    {
+        if ((*input & 0x80) == 0)
+            return *input;
+        else
+            return 0;
+    }
+    uint32_t mask = (0xff << (7 - length)) & 0x000000ff;
+    const uint32_t check = (mask << 1) & 0x000000ff;
+    uint32_t ucs = 0x00000000;
+    if ((*input & mask) == check)
+    {
+        mask = (~mask) & 0x000000ff;
+        ucs = (*input & mask) << (6 * (length-1));
+    }
+    else
+        return 0;
+    for (size_t i = 1; i < length; i++)
+    {
+        if ((*(input + i) & 0x000000c0) != 0x00000080)
+            return 0;
+        ucs += (*(input + i) & 0x0000003f) << (6 * (length-i-1));
+    }
+    return ucs;
+}
+
+/** Get Xlib keysym and xkb_keysym_to_uf8 from an Awesome keycode.
+ *
+ * xkb_keysym_to_utf8 returns a UTF-8 representation of the symbol a given
+ * symkey generates. It matches the engraving of the key for level-0 symbols
+ * (but lowercase)
+ *
+ * Keycodes may be given in a string in any valid format for `awful.key`:
+ * "#" + keycode, the symkey name and the UTF-8 representation will all work.
+ *
+ * If no suitable keysym is found, or a malformed keycode is given as an
+ * argument, this function will return (nil, nil)
+ *
+ * @treturn[1] string keysym The keysym name
+ * @treturn[1] nil keysym If no valid keysym is found
+ * @treturn[2] string printsymbol The xkb_keysym_to_utf8 result
+ * @treturn[2] nil printsymbol If the keysym has no printable representation.
+ * @staticfct get_key_name
+ */
+
+static int
+luaA_get_key_name(lua_State *L)
+{
+    // check if argument is valid
+    if (lua_gettop(L) > 1 || lua_type(L, 1) != LUA_TSTRING)
+    {
+        return 0;
+    }
+
+    const char* input = luaL_checkstring(L, 1);
+    const xkb_keysym_t *keysyms;
+    xkb_keysym_t keysym = XKB_KEY_NoSymbol;
+    const size_t length = strlen(input);
+    uint32_t ucs;
+
+    /* Checking for the three possible syntaxes awful.key uses:
+     * 1: #keycode (#8 to #255, any other is invalid)
+     * 2: the symbol itself (the result of xkb_keysym_to_utf8, e.g. @ for at).
+     * 3: the keysym
+     */
+    if (length > 1 && *input == '#' && length < 5) // syntax #1
+    {
+        int keycode_from_hash = 0;
+        for (size_t i = 1; i < length; i++)
+        {
+            if ('9' < *(input + i) || *(input + i) < '0')
+                return 0;
+            else
+            {
+                keycode_from_hash *= 10;
+                keycode_from_hash += input[i] - 48;
+            }
+        }
+        // A keycode cannot go lower than 8 or higher than 255
+        if ((keycode_from_hash < 8) || keycode_from_hash > 255)
+            return 0;
+        xkb_keycode_t keycode = (xkb_keycode_t) keycode_from_hash;
+        struct xkb_keymap *keymap = xkb_state_get_keymap(globalconf.xkb_state);
+        xkb_keymap_key_get_syms_by_level(keymap, keycode, 0, 0, &keysyms);
+        keysym = keysyms[0];
+    }
+    else if ((ucs = one_utf8_to_utf32(input, length)) > 0) //syntax #2
+        keysym = xkb_utf32_to_keysym(ucs);
+    else //syntax #3
+        keysym = xkb_keysym_from_name(input, XKB_KEYSYM_NO_FLAGS);
+
+    if (keysym == XKB_KEY_NoSymbol)
+    {
+        lua_pushnil(L);
+        lua_pushnil(L);
+    }
+    else
+    {
+        char *name = key_get_keysym_name(keysym);
+        lua_pushstring(L, name);
+        char *utfname;
+        utfname=malloc(8);
+        if (xkb_keysym_to_utf8(keysym, utfname, 8) > 0)
+            lua_pushstring(L, utfname);
+        else
+            lua_pushnil(L);
+    }
+    return 2;
 }
 
 /* Undocumented */
@@ -1007,6 +1128,7 @@ luaA_init(xdgHandle* xdg, string_array_t *searchpath)
         { "xrdb_get_value", luaA_xrdb_get_value},
         { "kill", luaA_kill},
         { "sync", luaA_sync},
+        { "get_key_name", luaA_get_key_name},
         { NULL, NULL }
     };
 
--- a/luaa.c
+++ b/luaa.c
@@ -417,9 +417,10 @@ get_modifier_name(int map_index)
     return 0; /* \0 */
 }
 
-// Helper function for luaA_get_key_name() below.
-// Will return the UTF-32 codepoint IF AND ONLY IF the input is exactly one
-// valid UTF-8 character. Otherwise, it will return zero.
+/* Helper function for luaA_get_key_name() below.
+ * Will return the UTF-32 codepoint IF AND ONLY IF the input is exactly one
+ * valid UTF-8 character. Otherwise, it will return zero.
+ */
 static uint32_t
 one_utf8_to_utf32(const char* input, const size_t length) {
     gunichar character = g_utf8_get_char_validated(input, length);
@@ -432,7 +433,7 @@ one_utf8_to_utf32(const char* input, const size_t length) {
     return character;
 }
 
-/** Get X11 keysym and a one-character representation from an Awesome keycode.
+/* Get X11 keysym and a one-character representation from an Awesome keycode.
  *
  * A "one-character representation" is a single UTF-8 representing the typical
  * output from that keysym in a text editor (e.g. " " for space, "ñ" for
@@ -449,7 +450,7 @@ one_utf8_to_utf32(const char* input, const size_t length) {
  * @treturn[1] nil keysym If no valid keysym is found
  * @treturn[2] string printsymbol The xkb_keysym_to_utf8 result
  * @treturn[2] nil printsymbol If the keysym has no printable representation.
- * @staticfct get_key_name
+ * @staticfct awful.keyboard.get_key_name
  */
 
 static int
@@ -1101,7 +1102,7 @@ luaA_init(xdgHandle* xdg, string_array_t *searchpath)
         { "xrdb_get_value", luaA_xrdb_get_value},
         { "kill", luaA_kill},
         { "sync", luaA_sync},
-        { "get_key_name", luaA_get_key_name},
+        { "_get_key_name", luaA_get_key_name},
         { NULL, NULL }
     };
 
--- a/luaa.c
+++ b/luaa.c
@@ -68,6 +68,11 @@
 #include <lauxlib.h>
 #include <lualib.h>
 
+/* for strings and Unicode handling */
+#include <glib.h>
+#include <stdlib.h>
+#include <string.h>
+
 #include <basedir_fs.h>
 
 #include <xcb/xcb_atom.h>
@@ -417,42 +422,22 @@ get_modifier_name(int map_index)
 // valid UTF-8 character. Otherwise, it will return zero.
 static uint32_t
 one_utf8_to_utf32(const char* input, const size_t length) {
-    // When casting a char to a uint32_t, the missing bits will all be set to
-    // the highest bit in the original char. Thus the exaggerated masking of
-    // every value (Unicode is defined around bytes)
-    if (length < 1 || length > 4)
+    gunichar character = g_utf8_get_char_validated(input, length);
+    // Return 0 if there is more than one UTF-8 character:
+    if (g_unichar_to_utf8(character, NULL) != (gint)length)
         return 0;
-    if (length == 1)
-    {
-        if ((*input & 0x80) == 0)
-            return *input;
-        else
-            return 0;
-    }
-    uint32_t mask = (0xff << (7 - length)) & 0x000000ff;
-    const uint32_t check = (mask << 1) & 0x000000ff;
-    uint32_t ucs = 0x00000000;
-    if ((*input & mask) == check)
-    {
-        mask = (~mask) & 0x000000ff;
-        ucs = (*input & mask) << (6 * (length-1));
-    }
-    else
+    // Return 0 if the character is invalid.
+    if (character == (gunichar)-1 || character == (gunichar)-2)
         return 0;
-    for (size_t i = 1; i < length; i++)
-    {
-        if ((*(input + i) & 0x000000c0) != 0x00000080)
-            return 0;
-        ucs += (*(input + i) & 0x0000003f) << (6 * (length-i-1));
-    }
-    return ucs;
+    return character;
 }
 
-/** Get Xlib keysym and xkb_keysym_to_uf8 from an Awesome keycode.
+/** Get X11 keysym and a one-character representation from an Awesome keycode.
  *
- * xkb_keysym_to_utf8 returns a UTF-8 representation of the symbol a given
- * symkey generates. It matches the engraving of the key for level-0 symbols
- * (but lowercase)
+ * A "one-character representation" is a single UTF-8 representing the typical
+ * output from that keysym in a text editor (e.g. " " for space, "ñ" for
+ * n_tilde, "Ā" for A_macron). It usually matches the main engraving of the key
+ * for level-0 symbols (but lowercase).
  *
  * Keycodes may be given in a string in any valid format for `awful.key`:
  * "#" + keycode, the symkey name and the UTF-8 representation will all work.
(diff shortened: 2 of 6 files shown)


actionless:
why you copied xkb_utf32_to_keysym.{c,h} from libxkbcommon?

luaa.c
```diff
@@ -411,6 +412,126 @@ get_modifier_name(int map_index)
...
+    for (size_t i = 1; i < length; i++)
+    {
+        if ((*(input + i) & 0x000000c0) != 0x00000080)
+            return 0;
+        ucs += (*(input + i) & 0x0000003f) << (6 * (length-i-1));
+    }
+    return ucs;
+}
+
+/** Get Xlib keysym and xkb_keysym_to_uf8 from an Awesome keycode.
```
psychon:
```suggestion
/** Get X11 keysym and xkb_keysym_to_uf8 from an Awesome keycode.
```
No idea what that second part with `xkb_keysym_to_uf8` is supposed to mean. Can we perhaps call this something understandable? `A string representation of the key`?

luaa.c
```diff
@@ -411,6 +412,126 @@ get_modifier_name(int map_index)
...
+        for (size_t i = 1; i < length; i++)
+        {
+            if ('9' < *(input + i) || *(input + i) < '0')
+                return 0;
+            else
+            {
+                keycode_from_hash *= 10;
+                keycode_from_hash += input[i] - 48;
+            }
+        }
```
psychon: `luaA_keystore()` (the code behind `k.key = whatever`) just uses `int keycode_from_hash = atoi(input + 1);`. Who needs error handling anyway? ;-)
But this saves 10 lines of code that I had to grok first.

luaa.c
```diff
@@ -411,6 +412,126 @@ get_modifier_name(int map_index)
...
+    xkb_keysym_t keysym = XKB_KEY_NoSymbol;
+    const size_t length = strlen(input);
+    uint32_t ucs;
+
+    /* Checking for the three possible syntaxes awful.key uses:
+     * 1: #keycode (#8 to #255, any other is invalid)
+     * 2: the symbol itself (the result of xkb_keysym_to_utf8, e.g. @ for at).
+     * 3: the keysym
+     */
+    if (length > 1 && *input == '#' && length < 5) // syntax #1
```
psychon: Why...? `length > 1` is unnecessary, because `length == 0` means that `*input == '\0'`, so not equal to `#`. No idea about that `length < 5`, but I would guess this wants to handle (some cases of) overflow...?

Actually... I guess you want to handle the case where input is just the symbol `#`.
```suggestion
    if (input[0] == '#' && input[1] != '\0') // syntax #1
```

luaa.c
```diff
@@ -411,6 +412,126 @@ get_modifier_name(int map_index)
...
+            if ('9' < *(input + i) || *(input + i) < '0')
+                return 0;
+            else
+            {
+                keycode_from_hash *= 10;
+                keycode_from_hash += input[i] - 48;
+            }
+        }
+        // A keycode cannot go lower than 8 or higher than 255
+        if ((keycode_from_hash < 8) || keycode_from_hash > 255)
```
psychon: Says who? I think the server can specify this.
```suggestion
        xcb_setup_t *setup = xcb_get_setup(globalconf.connection);
        if ((keycode_from_hash < setup.min_keycode) || keycode_from_hash > setup.max_keycode)
```

luaa.c
```diff
@@ -411,6 +412,126 @@ get_modifier_name(int map_index)
...
+ *
+ * xkb_keysym_to_utf8 returns a UTF-8 representation of the symbol a given
+ * symkey generates. It matches the engraving of the key for level-0 symbols
+ * (but lowercase)
+ *
+ * Keycodes may be given in a string in any valid format for `awful.key`:
+ * "#" + keycode, the symkey name and the UTF-8 representation will all work.
+ *
+ * If no suitable keysym is found, or a malformed keycode is given as an
+ * argument, this function will return (nil, nil)
```
psychon: Actually... it just does not return anything, I think. That's a difference. Is it worth documenting?

luaa.c
```diff
@@ -411,6 +412,126 @@ get_modifier_name(int map_index)
...
+                keycode_from_hash *= 10;
+                keycode_from_hash += input[i] - 48;
+            }
+        }
+        // A keycode cannot go lower than 8 or higher than 255
+        if ((keycode_from_hash < 8) || keycode_from_hash > 255)
+            return 0;
+        xkb_keycode_t keycode = (xkb_keycode_t) keycode_from_hash;
+        struct xkb_keymap *keymap = xkb_state_get_keymap(globalconf.xkb_state);
+        xkb_keymap_key_get_syms_by_level(keymap, keycode, 0, 0, &keysyms);
```
psychon: Why is it okay to hardcode `layout=0` and `level=0` here? I am not actually sure what they do, but I bet eventually someone will have a use for that.
(Note that I do not really understand XKB, so no idea what I am even suggesting here)

luaa.c
```diff
@@ -411,6 +412,126 @@ get_modifier_name(int map_index)
...
+    else if ((ucs = one_utf8_to_utf32(input, length)) > 0) //syntax #2
+        keysym = xkb_utf32_to_keysym(ucs);
+    else //syntax #3
+        keysym = xkb_keysym_from_name(input, XKB_KEYSYM_NO_FLAGS);
+
+    if (keysym == XKB_KEY_NoSymbol)
+    {
+        lua_pushnil(L);
+        lua_pushnil(L);
+    }
```
psychon: Since the other cases above also do `return 0;` for error handling, how about replacing this with `return 0;` as well?

luaa.c
```diff
@@ -411,6 +412,126 @@ get_modifier_name(int map_index)
...
+    else
+    {
+        char *name = key_get_keysym_name(keysym);
+        lua_pushstring(L, name);
+        char *utfname;
+        utfname=malloc(8);
```
psychon: That's a clear memory leak. Just do (how do I make a suggestion to replace two lines with one on GitHub?):
```suggestion
        char utfname[8];
```

xkb_utf32_to_keysym.h
```diff
@@ -0,0 +1,7 @@
+/* As of Feb 26, 2020, xkb_utf32_to_keysym.c exists only in development code
+ * for xkbcommon/libxkbcommon. This source code is to adapt
+ * https://github.com/xkbcommon/libxkbcommon/commit/0345aba082c83e9950f9dd8b7ea3bf91fe566a02
+ * which Jaroslaw Kubik authored and @bluetech committed.
+ * */
+
+static xkb_keysym_t xkb_utf32_to_keysym(uint32_t ucs);
```
psychon: This file is not included anywhere and can just be removed.

luaa.c
```diff
@@ -72,6 +72,7 @@
 
 #include <xcb/xcb_atom.h>
 #include <xcb/xcb_aux.h>
+#include "xkb_utf32_to_keysym.c"
```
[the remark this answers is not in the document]
ArenaL5: I tried, and I got this error. (`No tenemos la clave secreta` means `We don't have the secret key`.)

```
gpg: no default secret key: No tenemos la clave secreta
gpg: signing failed: No tenemos la clave secreta
make[4]: *** [CMakeFiles/dist.dir/build.make:57: CMakeFiles/dist] Error 2
make[3]: *** [CMakeFiles/Makefile2:175: CMakeFiles/dist.dir/all] Error 2
make[2]: *** [CMakeFiles/Makefile2:182: CMakeFiles/dist.dir/rule] Error 2
make[1]: *** [Makefile:225: dist] Error 2
make: *** [Makefile:45: dist] Error 2
```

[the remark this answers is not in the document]
Elv13: That error is not a C error. GPG is not used in the build process. I think you can just remove that line. If "just `make`" works, then that's good enough for the PR.

luaa.c
```diff
@@ -411,6 +412,126 @@ get_modifier_name(int map_index)
...
+    if (length == 1)
+    {
+        if ((*input & 0x80) == 0)
+            return *input;
+        else
+            return 0;
+    }
+    uint32_t mask = (0xff << (7 - length)) & 0x000000ff;
+    const uint32_t check = (mask << 1) & 0x000000ff;
+    uint32_t ucs = 0x00000000;
```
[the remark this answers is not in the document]
ArenaL5: I have programmed on other languages, done some experiments with bit shifts and bit masking, just not C.

I still haven't fully wrapped my head around the dereference operator, it's just that this way seems to work and won't get me any warning in the compiler hahaha

luaa.c
```diff
@@ -411,6 +412,126 @@ get_modifier_name(int map_index)
     return 0; /* \0 */
 }
 
+// Helper function for luaA_get_key_name() below.
+// Will return the UTF-32 codepoint IF AND ONLY IF the input is exactly one
+// valid UTF-8 character. Otherwise, it will return zero.
```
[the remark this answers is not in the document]
ArenaL5: I'll try it.

luaa.c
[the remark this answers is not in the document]
psychon: No idea how this stuff works / how to disable GPG. It also asks for a key here.
Anyway: Yes, it does include the necessary files in the tarballs.

luaa.c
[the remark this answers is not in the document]
psychon: If you want to know a bit more about the ugly details: `char` can be signed or unsigned, depending on the target system. The easiest way to get stuff to work here would be change the type of input to `uint8_t`. However, not writing a utf8 decoder would still be simpler, I think. :-)

(Feel free to ask if my proposed glib solution does not work)

luaa.c
ArenaL5: Yes, please. It's not even correctly written, it should be `xkb_keysym_to_utf8`. (I forgot the T.)

`A string representation of the key` is pretty legible, but if you can write a clearer expression, please tell me.

luaa.c
ArenaL5: Both your guesses are right: I want to avoid overflows, and I want to handle `#\0` separately.

I'll accept your suggestion when I find the time, but I'd rather keep the `length < 5`.

luaa.c
ArenaL5: Absolutely. I'll change this.

luaa.c
ArenaL5: It's not a difference for Lua.

`return 0;` passes 0 values from the stack to Lua, and Lua reports `nil` if you ask too many values from a function.

Meaning that you're right about your suggestion for lines 518 and 519: `lua_pushnil(L); lua_pushnil(L); return 2;` has the same effect as `return 0;`.

luaa.c
ArenaL5: Line 508 is pure copypaste. I copied it somewhere from `luaa.c`.

I can justify `level=0`: levels represent modifiers like Shift, AltGr, Shift+AltGr... we want the base keysym with no modifiers because that's what's printed on the key. The modifiers are going to be printed separately in the cheatsheet anyways: (e.g. `Shift+e`, not `E`).

I'm not sure about `layout=0`, though.

luaa.c
ArenaL5: Thank you. I'll ask you if I can't make it work. (The compiler was giving me some warnings about passing a pointer instead of a value or something like that...)

xkb_utf32_to_keysym.h
ArenaL5: I just added this because the compiler told me to. If you're sure, I'll delete it.

luaa.c
ArenaL5: What?? That's a thing??

...I wasn't expecting a function like `atoi()` to exist in C, at least not without some specific libraries. I'll be really glad to change it.

> Who needs error handling anyway? ;-)

What could possibly go wrong??

luaa.c
psychon: Hm... I was sure there was a way to tell apart `nil` and 'nothing', but I cannot find a way to do so from Lua, only from C.

psychon: Oh and: Sorry for making a useless suggestion then.

luaa.c
psychon: Well, okay, then I guess this is fine. I bet I already complained at whichever patch added that hardcoded `level=0`.

(So... with my `de` keyboard layout, there is a key with `#`, but no key with `'` because that's Shift+#.... Some day one might want an API that can handle "more".)

luaa.c
psychon: In the "worst case", you have to pass in `&utfname[0]` instead of `utfname` to the function in the next line.

xkb_utf32_to_keysym.h
psychon: Huh? But... the file is not referenced anywhere...? I thought.

luaa.c
psychon: There are even functions like `strtol` that actually allow handling errors (even though that is a bit complicated to use) :-)

xkb_utf32_to_keysym.h
ArenaL5: Nah, you were right. I deleted it, and the compiler threw no warning.

I found so many errors at compile-time that I had forgotten what does what.

luaa.c
ArenaL5: Okay, done. No problems with `char utfname[8]`.

By the way, I rolled all the fixes you suggested in the same commit. I can rebase later if you all want a cleaner history.

luaa.c
ArenaL5: Hey, don't be. _Every other one_ has been a good suggestion.

ArenaL5:
I don't know why Travis CI failed. The log for the last job just cuts off mid-word at line 3442.

`tests/run.sh` throws no errors.

luaa.c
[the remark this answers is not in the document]
ArenaL5: It worked. I just added one more bit of error handling, just in case,

psychon:
> I don't know why Travis CI failed. The log for the last job just cuts off mid-word at line 3442.

I clicked on "raw log" and was able to scroll down:
```
[31mThere were 1 errors:
 - /home/travis/build/awesomeWM/awesome/tests/test-awful-placement.lua: Error: running function for step 13/50 (@2): ...s/build/awesomeWM/awesome/tests/test-awful-placement.lua:37: (4, 20, 1911, 200) ~= (4, 20, 1911, 1009)
```
This error isn't from you. I think it has been randomly failing for a while now.

ArenaL5:
I didn't find any problem in my machine. Is it possible that the new test case I added is interfering `test-awful-placement.lua`?

If not, and if it really happens at random, force-pushing a clone of commit dc3f8838b4160ba75845bf36577b262ff3bd51b5 should give Travis another chance. Do you want me to try that? I could meanwhile merge my first two commits.

Elv13:
> I didn't find any problem in my machine. Is it possible that the new test case I added is interfering test-awful-placement.lua?

No. That test has been flaky since last summer. It started out rare (~1% of the time), then became more and more common (~25% right now). I am not able to reproduce it locally. There is probably an `awesome.sync()` missing *somewhere* in the code. I tried adding it in random places, but I could always reproduce the failure on Travis. I ran it 1k time on my laptop and never could reproduce.

psychon:
> There is probably an awesome.sync() missing somewhere in the code.

Uhm. I doubt that. The X11 server should not really be involved in what AwesomeWM things the current geometry of a window is (unless a window changes its own geometry, which is not done here). Sadly, the traceback is useless due to tail calls, so no idea what actually fails here. However, the error message sounds like something has height 200, but is instead expected to have height 1009 (which I guess is the height of the workarea).
Random guess where the 200 comes from: Uhm... there is no 200 anywhere here. It's also oddly round, given that some space for the titlebar should have been added.

`13/50 (@2)` does not really help me either. Unless I am mis-counting, each iteration of the outer `for` loop in the tests adds. 17 tests. 3 times that should be 51...?? Anyway, this would then be one of the following two that is failing:
https://github.com/awesomeWM/awesome/blob/382635a94a3489a827890676bd10d2e8517ce484/tests/test-awful-placement.lua#L268-L292

Hm, yeah, dunno...

Elv13:
> Hm, yeah, dunno...

The reason I blame sync is because the failure appears at random. So it's either a hash key issue or some prior steps (titlebars, borders, resize) took a different code path. Something, somewhere make this non-deterministic, but I can't reproduce, so it's painfully hard to debug (and it probably a little obvious thing, and I didn't spend that much time trying anyway).

psychon:
Let's thing a bit about what this could be. Perhaps we find something.

`add_client` spawns a client and then continue when the number of clients increased. It then assumes that... either something already identified the client correctly, or `client.get()[1]` is the one to check. Could this just be that the code somehow grabs the wrong client to test?
Anyway, "waiting for a new client when only a single client spawns" is not something that should be influenced by `awesome.sync()`. Perhaps some delayed call somewhere? But the test runner should make sure that there are no pending calls when it calls the test steps, I think.
Other steps do stuff like `c.hidden = true` or `c.minimized = true`.. This also does not need `awesome.sync()`.
And... that's basically it.

Another debugging idea could be to set the timeout to `0.1` here and see if that makes the problem go away on Travis. That would point towards some kind of race condition.
https://github.com/awesomeWM/awesome/blob/382635a94a3489a827890676bd10d2e8517ce484/tests/_runner.lua#L125

Anyway, I will stop now spamming this PR with unrelated stuff. And I do not have time to influence this flaky test much soon. Sorry.

luaa.c
```diff
@@ -1007,6 +1101,7 @@ luaA_init(xdgHandle* xdg, string_array_t *searchpath)
         { "xrdb_get_value", luaA_xrdb_get_value},
         { "kill", luaA_kill},
         { "sync", luaA_sync},
+        { "get_key_name", luaA_get_key_name},
```
Elv13: Please change this to `_get_key_name`. If you wish to expose the function, you can make a new function in `awful.keyboard` and call `_get_key_name` from there. This way people will find it faster than if it is in the `awesome` module. It also allows to break the API more easily in the future without impacting anyone.

tests/test-keycodes.lua
```diff
@@ -0,0 +1,56 @@
+local naughty = require("naughty")
+
+local prettyerror = function(key, keysym, character, expected_keysym, expected_character)
```
Elv13: `local function prettyerror(...`

luaa.c
```diff
@@ -411,6 +417,94 @@ get_modifier_name(int map_index)
     return 0; /* \0 */
 }
 
+// Helper function for luaA_get_key_name() below.
```
Elv13: Please use `/*` for the comments for consistency. There is some `//`, but 99% of the codebase (and most C code) use `/*`.

luaa.c
```diff
@@ -411,6 +417,94 @@ get_modifier_name(int map_index)
...
+    // Return 0 if there is more than one UTF-8 character:
+    if (g_unichar_to_utf8(character, NULL) != (gint)length)
+        return 0;
+    // Return 0 if the character is invalid.
+    if (character == (gunichar)-1 || character == (gunichar)-2)
+        return 0;
+    return character;
+}
+
+/** Get X11 keysym and a one-character representation from an Awesome keycode.
```
Elv13: Please use `/*` here rather than `/**`. If you want to expose this in the public API, use `awful.keyboard` (see other comment).`

tests/test-keycodes.lua
```diff
@@ -0,0 +1,56 @@
...
+        ["\27"] = "\\e",
+        ["\f"] = "\\f",
+        ["\r"] = "\\r",
+        ["\n"] = "\\n",
+        ["\t"] = "\\t",
+        ["\v"] = "\\v",
+    })
+end
+
+local keycode_test = function(key, expected_keysym, expected_character)
```
Elv13: `local function keycode_test`

tests/test-keycodes.lua
```diff
@@ -0,0 +1,56 @@
...
+    naughty.notification {
+        urgency = "low",
+        title   = keysym or "[nil]",
+        message = character or "[nil]"
+    }
+    return assert(keysym == expected_keysym and character == expected_character,
+        prettyerror(key, keysym, character, expected_keysym, expected_character))
+end
+
+local triple_test = function(keycode, expected_keysym, expected_character)
```
Elv13: `local function triple_test`

psychon:
> If @psychon and @actionless agree with the other parts,

I looked at the C parts and left my comments. No more complaints from me.

ArenaL5:
The next step would be that, and the step after would be to bind every layout to a different shortcut. You told me that few keyboards nowadays even have numeric keyboard... so maybe you could recommend another shortcut? I'm flipflopping `Super+(Fkeys)` or `Super+Alt+(number keys)`.

I pushed a new commit with your suggestions, @Elv13. Did I miss anything?

actionless:
i still think what numpad as default keybinding for them is OK, people without numpad still could use old way or rebind the hotkeys

awesomerc.lua
```diff
@@ -22,6 +22,15 @@ local hotkeys_popup = require("awful.hotkeys_popup")
 -- when client with a matching name is opened:
 require("awful.hotkeys_popup.keys")
 
+local keyname_test = function(key)
```
Elv13: I believe this is test code that got accidentally committed?

lib/awful/keyboard.lua
```diff
@@ -197,6 +197,29 @@ function module.remove_client_keybinding(key)
...
+-- "#" + keycode, the symkey name and the UTF-8 representation will all work.
+--
+-- If no suitable keysym is found, or a malformed keycode is given as an
+-- argument, this function will return (nil, nil)
+--
+-- @treturn[1] string keysym The keysym name
+-- @treturn[1] nil keysym If no valid keysym is found
+-- @treturn[2] string printsymbol The xkb_keysym_to_utf8 result
+-- @treturn[2] nil printsymbol If the keysym has no printable representation.
+-- @staticfct get_key_name
```
Elv13: Please use `awful.keyboard.get_key_name`. The reason is that it is what the users with copy paste and use. We had trouble before because some new users also confused methods with static function, so we moved to the full name everywhere.

awesomerc.lua
ArenaL5: Shit. How did I send that?

awesomerc.lua
```diff
@@ -434,6 +434,11 @@ client.connect_signal("request::default_keybindings", function()
                 c:raise()
             end ,
             {description = "(un)maximize horizontally", group = "client"}),
+        awful.key({ modkey,           }, "z",
```
Elv13: Still here ;)

awesomerc.lua
```diff
@@ -561,3 +566,4 @@ end)
 client.connect_signal("mouse::enter", function(c)
     c:activate { context = "mouse_enter", raise = false }
 end)
+
```
Elv13: and here

awesomerc.lua
ArenaL5: Oh my God hahahahahaahahahah

THANK YOU for combing this line by line, these last commits have been a trainwreck.

ArenaL5: Line 441: `description = "test keyname", group = "shit"`. Real classy, myself. _That's_ code ready for release.

awesomerc.lua
ArenaL5: A blank line. Okay, done.

The good news is that I only added test code to `awesomerc.lua`. The other files shouldn't have test code on them.

If this time everything goes okay, I assume you'll want me to squash all the commits between 392eff73f001d163ec6649fc90657eb6fefdac89 and 934052d2381398f21aff7f43dc6f65a97eedd651 into a single one?

Elv13:
I made the following changes:

```
diff --git a/tests/test-keycodes.lua b/tests/test-keycodes.lua
index 2b5c76e44..d70e810b0 100644
--- a/tests/test-keycodes.lua
+++ b/tests/test-keycodes.lua
@@ -32,13 +32,13 @@ local function keycode_test(key, expected_keysym, expected_character)
 end
 
 local function triple_test(keycode, expected_keysym, expected_character)
-    result = keycode_test(keycode, expected_keysym, expected_character)
+    local result = keycode_test(keycode, expected_keysym, expected_character)
     if expected_keysym then
-        local_result = keycode_test(expected_keysym, expected_keysym, expected_character)
+        local local_result = keycode_test(expected_keysym, expected_keysym, expected_character)
         result = result and local_result
     end
     if expected_character then
-        local_result = keycode_test(expected_character, expected_keysym, expected_character)
+        local local_result = keycode_test(expected_character, expected_keysym, expected_character)
         result = result and local_result
     end
     return result
```

Because for unknown reasons one of the test suite "just aborts", the `luacheck` was not executed. Otherwise I moved the "fixup" changes to test-keycode to the right commit and squashed the `rc.lua` commits to avoid having `description = "test keyname", group = "shit"` stuff in the git history. I will merge this now. Thanks.

ArenaL5:
Thank you for everything (and for squashing it yourself). The next PRs should be much more easier.

PR #3014, opened 2020-02-29
Fix the keyboard widget

Elv13:
The reason `buttons` was no longer working is mostly bitrot. That widget, along with the systray and textclock, are among the last remaining v3.5 era widgets which were not modernized at some point or another.


Elv13:
Thanks to @Perlovka for testing.

PR #3016, opened 2020-03-01
Added support for custom text for layouts in layoutbox

EuanFH:
If theme variable beautiful.layout_LAYOUTTEXT_text is set for a layout instead of the default name showing it will show the custom text set in the variable.
Precedents is still to the layout image meaning custom text will only be shown if there is no img already set. With no image or custom text set it will still use the layout name.

Use Case: you don't want to use images and would rather use a font like nerdfonts for your icons.

I couldn't find the documentation source for https://awesomewm.org/doc/api/libraries/awful.layout.html so I haven't updated the documentation to show the new theme variables. If anyone could tell me where its located so I could add them it would be greatly appreciated.

If you want anything changed just ask.

### DIFF
--- a/lib/awful/widget/layoutbox.lua
+++ b/lib/awful/widget/layoutbox.lua
@@ -32,8 +32,9 @@ local function update(w, screen)
     w._layoutbox_tooltip:set_text(name or "[no name]")
 
     local img = surface.load_silently(beautiful["layout_" .. name], false)
+    local text = beautiful["layout_" .. name .. "_text"]
     w.imagebox.image = img
-    w.textbox.text   = img and "" or name
+    w.textbox.text   = img and "" or text or name
 end
 
 local function update_from_tag(t)


actionless:
> I couldn't find the documentation source for https://awesomewm.org/doc/api/libraries/awful.layout.html so I haven't updated the documentation to show the new theme variables. If anyone could tell me where its located so I could add them it would be greatly appreciated.

`lib/awful/layout/init.lua`

Aire-One:
I don't know what is the point of view of the main developers on this PR but if it makes its way to master, I suggest to support this feature more in depth.I'd like to also have the possibility to change the text from code, not only beautiful configuration. With additional parameters to the constructor and new properties we could access.

Another approach would be to go the #2955 way and use widget template. We probably need to implement some [IoC](https://en.wikipedia.org/wiki/Inversion_of_control) to allow user interaction on widget update.

Thomashighbaugh:
Its really annoying that each of these that are all cleared for take off linger around waiting for 2 approving reviews. Like I get everyone has a life, and really appreciate this project in particular and thus its devs, 

but how long really do these things need to linger? 

I mean in nearly six months someone should be able to go over the thing and determine what's what in addition to actionless who is the dev that comments on all of these (thanks for your work especially).

actionless:
@ThomasAH lol bae you know how to whine but not how to read comments

Aire-One:
@Thomashighbaugh, please note this PR is not "all cleared" and ready to be merged. We don't merge new features without documentation and @actionless pointed where to write the doc. It's now @EuanFH role to either, write the necessary doc or ask for more help.

(+ I don't think it's a good feature we should merge, I would instead prefer to have a widget template mechanism)

nullbyto:
Is there any plans to add this ?

sclu1034:
> (+ I don't think it's a good feature we should merge, I would instead prefer to have a widget template mechanism)

I agree. Awesome is, first and foremost, a framework for people to write their own piece of window manager software. The use case of using `awful.widget.layoutbox` with just layout names is very simple to implement yourself (a `textbox` and a handful of signal handlers, and the code is already there to reference), and adding custom names to that amounts to a simple map lookup:

```lua
local CUSTOM_LAYOUT_NAMES = {
	tilebottom = "foo",
	tiletop = "bar",
	max = "baz"
}

-- ...
local name = awful.layout.getname(awful.layout.get(screen))
w:set_text(CUSTOM_LAYOUT_NAMES[name])
```
	
As far as I see it, this is not something that needs to be handled explicitly by Awesome. We shouldn't keep on piling more and more things into `beautiful` variables.

nullbyto:
> a `textbox` and a handful of signal handlers, and the code is already there to reference

Thanks, thats a good idea. Can you please explain how i would do that with signals, to help change the names in the widget when changing layouts? I'm relatively new to awesome and lua.

sclu1034:
The signals wouldn't be for showing custom names. The lookup table I showed in the previous comment does that.

nullbyto:
I know this is offtopic but in case someone needs it i found it out:
```lua
tag.connect_signal("property::layout", 
    function()    
        local name = awful.layout.getname(awful.layout.get(awful.screen.focused()))
        if CUSTOM_LAYOUT_NAMES[name] ~= nil then
            layout_textbox:set_text(CUSTOM_LAYOUT_NAMES[name])
        else
            layout_textbox:set_text(name)
        end

    end
)

```

PR #3023, opened 2020-03-05
convert hotkeys_popup to awful.popup instead of wibox

### LINKED ISSUE #2614: Refactor awful.hotkeys_popup to use awful.popup

SethBarberee:
Fixes #2614 

By moving to `awful.popup`, I can just use `awful.placement.centered` for geometry instead of hand calculations. I didn't make any other modifications so functionality is exactly the same API.

### DIFF
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -570,7 +570,16 @@ function widget.new(args)
         end
         table.insert(pages, columns)
 
-        local mywibox = wibox({
+        -- Function to place the widget in the center and account for the
+        -- workarea. This will be called in the placement field of the
+        -- awful.popup constructor.
+        local place_func = function(c)
+            awful.placement.centered(c, {honor_workarea = true})
+        end
+
+        -- Construct the popup with the widget
+        local mypopup = awful.popup {
+            widget = pages[1],
             ontop = true,
             bg=self.bg,
             fg=self.fg,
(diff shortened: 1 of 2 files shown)


SethBarberee:
@actionless I changed the placement call to `honor_workarea`. That should work for what you want.

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -570,47 +570,40 @@ function widget.new(args)
...
-            x = wa.x + math.floor((wa.width - wibox_width - self.border_width*2) / 2),
-            y = wa.y + math.floor((wa.height - wibox_height - self.border_width*2) / 2),
-            width = wibox_width,
-            height = wibox_height,
-        })
-        mywibox:set_widget(pages[1])
-
-        mywibox.buttons = {
-            awful.button({ }, 1, function () widget_obj:hide() end),
-            awful.button({ }, 3, function () widget_obj:hide() end)
```
actionless: why you removed the mouse bindings?

SethBarberee: Handled already in [`awful.popup`](https://github.com/awesomeWM/awesome/blob/master/lib/awful/popup.lua#L282`)

SethBarberee: Did a test drive to verify and I was wrong.. Doing a fixup commit real quick

PR #3028, opened 2020-03-07
use user specified permission.

handsome0hell:
User settings should override default behavior.

### DIFF
--- a/lib/awful/permissions/_common.lua
+++ b/lib/awful/permissions/_common.lua
@@ -20,10 +20,6 @@ local default_permissions = {
 }
 
 function module.check(object, class, request, context)
-    if not default_permissions[class] then return true end
-    if not default_permissions[class][request] then return true end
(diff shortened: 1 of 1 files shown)


handsome0hell:
@Elv13 Thanks for your work!

Some programs running with wine frequently request for changing geometry, which causes the window to move away when the pointer is hovered over it. Denying their geometry permission in ewmh context seems fixed this problem.

I'm thinking about maybe in some cases we need more detailed permission control, like separating `geometry` and `client_geometry_requests` permissions.

Elv13:
Cool!

Yes, this is probably the most useful one that currently works correctly. The other important ones are blocking size hints, blocking maximization and blocking fullscreen. However those are not merged yet because they are very hard to get right (they require a full state machine in C for each one). Anyway, thanks for this, I will merge it now.

Elv13:
Merged (after fixing the tests, this PR setting disabled the ability to update your branch)

PR #3029, opened 2020-03-08
Fix most reported notification issues.

Elv13:
I finally setuped a VM with the correct configuration to attack some of these bugs. 

@manilarome @SethBarberee @unai-ndz @madprops please test.

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -1478,8 +1478,10 @@ for _, prop in ipairs { "border_width", "border_color", "opacity" } do
         return self["_"..prop]
     end
     client.object["set_"..prop] = function(self, value)
-        self._private["_user_"..prop] = true
-        self["_"..prop] = value
+        if value ~= nil then
+            self._private["_user_"..prop] = true
+            self["_"..prop] = value
+        end
     end
 end
 
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -323,6 +323,13 @@ function notif_methods.Notify(sender, object_path, interface, method, parameters
                 notification[k] = v
             end
 
+            -- Update the icon if necessary.
+            if app_icon ~= notification._private.app_icon then
+                notification._private.app_icon = app_icon
+                notification._private.icon = nil
+                notification:emit_signal("property::icon")
+            end
+
             -- Even if no property changed, restart the timeout.
             notification:reset_timeout()
         else
--- a/lib/naughty/notification.lua
+++ b/lib/naughty/notification.lua
@@ -618,6 +618,17 @@ for _, prop in ipairs(properties) do
 
 end
 
+-- Changing the image will change the icon, make sure property::icon is emitted.
+for _, prop in ipairs {"image", "images" } do
+    local cur = notification["set_"..prop]
+
+    notification["set_"..prop] = function(self, value)
+        cur(self, value)
+        self._private.icon = nil
+        self:emit_signal("property::icon")
+    end
+end
+
 local hints_default = {
     urgency  = "normal",
     resident = false,
--- a/lib/naughty/widget/_markup.lua
+++ b/lib/naughty/widget/_markup.lua
@@ -1,0 +1,62 @@
+local beautiful = require("beautiful")
+
+local module = {}
+
+-- Since some escaping needs to be undone, we have to escape the escaped <>.
+local pre_escape = {["&lt;"] = "&zzlt;", ["&gt;"] = "&zzgt;"}
+
+local escape_pattern = "[<>&]"
+local escape_subs    = { ['<'] = "&lt;", ['>'] = "&gt;", ['&'] = "&amp;" }
+
+-- Also reverse escaping some allowed tags because people actually use them.
+local escape_undo = {['&lt;span '] = "<span "}
+
+for _, allowed in ipairs {'b', 'i', 'u', 'span'} do
+    escape_undo['&lt;' ..allowed..'&gt;'] = "<" ..allowed..">"
+    escape_undo['&lt;/'..allowed..'&gt;'] = "</"..allowed..">"
+end
+
+-- Best effort attempt to allow a subset of pango markup in the text while
+-- removing invalid content. If invalid content is present, nothing is
+-- displayed.
+local function escape_text(text)
+    -- Take care of the already escaped content.
+    for pattern, subs in pairs(pre_escape) do
+        text = text:gsub(pattern, subs)
+    end
+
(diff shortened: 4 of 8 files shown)


manilarome:
Looks good to me! Well done!

manilarome:
I'm still having the problem with `naughty.widget.icon` not updating. With this gif below I'm using `notify-send -i $ICON_PATH`. I also tried creating a `naughty.notification` using `awesome-client` but same issue.

ezgif com-video-to-gif(1)

Elv13:
@manilarome Can you add a `print("DBUS", k, v)` under `if k == "destroy" then k = "destroy_cb" end` in `lib/naughty/dbus.lua`. Then a `print("CALLED")` in function `tb._private.icon_changed_callback` of `naughty.widget.icon`. Then tell me which property DBus is trying to use to set the icon (and it is a path or a pixmap). After testing with Spotify and dbus-send, I can confirm the icon is updated for me.

lib/awful/client.lua
```diff
@@ -1478,8 +1478,10 @@ for _, prop in ipairs { "border_width", "border_color", "opacity" } do
         return self["_"..prop]
     end
     client.object["set_"..prop] = function(self, value)
-        self._private["_user_"..prop] = true
-        self["_"..prop] = value
+        if value ~= nil then
+            self._private["_user_"..prop] = true
+            self["_"..prop] = value
+        end
```
psychon: So, now it is generally impossible to "unset" something? Uhm.. client shape comes to my mind for something where this is actually used.

(Also, I never understood where this new "set to `nil`" errors come from... what changed that broke stuff that previously worked?)

Elv13: This line is only called for `border_width`, `border_color` and `opacity`. Setting it to `nil` never worked. It is an issue now because my API-backward-compatibility code manages to trigger this. By trying too hard to preserve the API compatibility, I managed to introduce an obscure breakage.

Longer version:

The code to set the border color and width for focused vs. unfocused client was in `rc.lua`. There was also some code paths trying to remove borders for maximized and fullscreen client. Those were mostly fighting each other and "worked" in some case because the signals were connected in a specific way. It was also the only `beautiful` variables which were only used in `rc.lua`.

When I made the permission system and standardized the `request::` signals, I wanted to fix this. Using the permission system allowed the rules to select different border width and color for different clients, etc (well, kind of). It also allowed to add more `beautiful` variables for other states than just `normal` and `focused`. For example `urgent` and `floating`. Finally, it allows to fix the race condition to disable borders for maximized client, which was the point.

To do that change without breaking the API and preserve the behavior when people still had that section in `rc.lua`, I added magic. My magic broke.

psychon: Thanks for the explanation

manilarome:
> @manilarome Can you add a `print("DBUS", k, v)` under `if k == "destroy" then k = "destroy_cb" end` in `lib/naughty/dbus.lua`. Then a `print("CALLED")` in function `tb._private.icon_changed_callback` of `naughty.widget.icon`. Then tell me which property DBus is trying to use to set the icon (and it is a path or a pixmap). After testing with Spotify and dbus-send, I can confirm the icon is updated for me.

Can confirm, the icon is updating in Spotify. Anyway, I'm sorry but I can't find the output for this `print("DBUS", k, v)`. I already tried `tail`ing the  `xorg-session.log` which is the output of `ls -l /proc/$(pidof awesome)/fd/2` but I can't find a relevant output when creating a notification with icon.

Elv13:
> I can't find the output

The easiest way is to use Xephyr. It will print the output in a terminal.

```sh
Xephyr :1 -screen 1280x800 &
DISPLAY=:1 awesome
```

elenapan:
For me, updating a notification icon with `naughty` works as expected.

With `notify-send` the icon is not updated.
To clarify, I am using https://github.com/vlevit/notify-send.sh which allows updating existing notifications.

Notification command:
```sh
notify-send -R /tmp/somenotif "hello" "there" -i /tmp/icon1.png
sleep 1
notify-send -R /tmp/somenotif "good" "bye" -i /tmp/icon2.png
```

Output from AwesomeWM (after applying the changes in https://github.com/awesomeWM/awesome/pull/3029#issuecomment-596166498):
```
2020-03-08 15:19:10 W: awesome: Notification hello is being updatedby a different DBus connection (:1.1713), this is suspicious. The original connection was :1.1712
DBUS	app_icon	/tmp/icon2.png
DBUS	destroy_cb	function: 0x5591ae993c70
DBUS	appname	notify-send
DBUS	title	good
DBUS	urgency	normal
DBUS	app_name	notify-send
DBUS	actions	table: 0x5591ae993c30
DBUS	freedesktop_hints	table: 0x5591ae993aa0
DBUS	message	bye
DBUS	replaces_id	5.0
```

`CALLED` is not printed.

manilarome:
```
2020-03-08 21:30:25 W: awesome: DBUS
2020-03-08 21:30:25 W: awesome: DBUS
2020-03-08 21:30:25 W: awesome: DBUS
2020-03-08 21:30:25 W: awesome: DBUS
2020-03-08 21:30:25 W: awesome: DBUS
2020-03-08 21:30:25 W: awesome: DBUS
2020-03-08 21:30:25 W: awesome: DBUS
2020-03-08 21:30:25 W: awesome: DBUS
2020-03-08 21:30:25 W: awesome: DBUS
2020-03-08 21:30:25 W: awesome: DBUS
2020-03-08 21:30:26 W: awesome: Notification Cure For The Itch is being updatedby a different DBus connection (:1.3144), this is suspicious. The original connection was :1.3142
2020-03-08 21:30:26 W: awesome: DBUS
2020-03-08 21:30:26 W: awesome: DBUS
2020-03-08 21:30:26 W: awesome: DBUS
2020-03-08 21:30:26 W: awesome: DBUS
2020-03-08 21:30:26 W: awesome: DBUS
2020-03-08 21:30:26 W: awesome: DBUS
2020-03-08 21:30:26 W: awesome: DBUS
2020-03-08 21:30:26 W: awesome: DBUS
2020-03-08 21:30:26 W: awesome: DBUS
2020-03-08 21:30:26 W: awesome: DBUS
2020-03-08 21:35:23 W: awesome: a_glib_poll:470: Last main loop iteration took 0.114418 seconds! Increasing limit for this warning to that value.
```

These are only the output. And `CALLED` is not printed.

Elv13:
Thanks to both of you. The `notify-send.sh` script did the trick. It's fixed now.

Elv13:
I will merge this soon given it has been tested by some people. Thanks, everybody!

PR #3030, opened 2020-03-08
[Ignore this] Random attempts to fix the flacky placement test

Elv13:
I did a few of these commits last years to try a few things (with no success). But this is an "official" PR to test random things until it works or I give up again. I can't reproduce the issue locally.


Elv13:
I can now confirm that changing the timeout does not make the flackyness go away.

PR #3040, opened 2020-03-16
[RFC] Redesign how notification icons are handled.

Elv13:
(WARNING: This breaks unreleased APIs)

Following some recent problems with how icons are handled in the new notification API, I redesigned it. It became clear from all the bug reports and recent Reddit thread that the old/current design wont scale. Icons can come in ~15 different format or protocols. My previous rewrite supported 4 and the old API supported 6. In either case, the problem remain: some people wont be happy.

As usual, when something is too complex to hide in the private implementation, I ported it to the `request::` API. With this PR, it becomes possible to plug new handlers to support more notification icons. All existing code has been moved to handlers.

For example, it is now possible to add an handler to support icons passed with an HTTPS URI rather than a local file. It is also possible to use the permission system to control which kind of handler can be used for each notification.

A documented example shows how to enable XDG icon lookup. The legacy API had a very broken implementation of this built-in (`awful.util.geticonpath`), but it failed more than 60% of the time. The example shows how to use the menubar implementation of the XDG icon spec. An even better alternative would be to use GTK own implementation through LGI. I don't plan to enable XDG lookup by default for the next few releases. Our current implementations (we have 3!) are all badly unsalvageable. The day will come when someone brave enough assemble all the pieces into a coherent and fast implementation. If you read this, you can be this Hero, I wont be.

@elenapan @manilarome Please review / comment on the changes. It breaks the API a little bit (`request::icon` was used for notification actions and it was moved to `request::action_icon`). You input is also generally welcome since you use this the most.

### DIFF
--- a/lib/gears/object.lua
+++ b/lib/gears/object.lua
@@ -134,7 +134,8 @@ function object:emit_signal(name, ...)
     end
 end
 
-function object._setup_class_signals(t)
+function object._setup_class_signals(t, args)
+    args = args or {}
     local conns = {}
 
     function t.connect_signal(name, func)
@@ -142,6 +143,18 @@ function object._setup_class_signals(t)
         conns[name] = conns[name] or {}
         table.insert(conns[name], func)
     end
+
+    -- A variant of emit_signal which stops once a condition is met.
+    if args.allow_chain_of_responsibility then
+        function t._emit_signal_if(name, condition, ...)
+            assert(name)
+            for _, func in pairs(conns[name] or {}) do
+                if condition(...) then return end
+                func(...)
+            end
+        end
+    end
+
 
     --- Emit a notification signal.
     -- @tparam string name The signal name.
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -18,6 +18,7 @@ local gdebug  = require("gears.debug")
 local screen  = require("awful.screen")
 local gtable  = require("gears.table")
 local gobject = require("gears.object")
+local gsurface = require("gears.surface")
 
 local naughty = {}
 
@@ -590,6 +591,7 @@ naughty.connect_signal("request::screen", naughty.default_screen_handler)
 --
 -- * app_icon
 -- * clients
+-- * path
 -- * image
 -- * images
 --
@@ -607,11 +609,18 @@ naughty.connect_signal("request::screen", naughty.default_screen_handler)
 --        end
 --    end)
 --
+-- The `images` context has no handler. It is part of the specification to
+-- handle animations. This is not supported by default.
+--
 -- @signal request::icon
 -- @tparam notification n The notification.
 -- @tparam string context The source of the icon to look for.
 -- @tparam table hints The hints.
 -- @tparam string hints.app_icon The name of the icon to look for.
+-- @tparam string hints.path The path of the icon.
+-- @tparam string hints.image The path or pixmap of the icon.
+-- @see naughty.icon_path_handler
+-- @see naughty.client_icon_handler
 
 --- Emitted when the screen is not defined or being removed.
 -- @signal request::screen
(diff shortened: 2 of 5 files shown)


manilarome:
Neat! I did not encounter any breakage in my config. This also fixes the problem with the icon not updating in my ncmpcpp script I showed the last time. It's all good to me :+1:  Thanks for this!

SethBarberee:
> It is also possible to use the permission system to control which kind of handler can be used for each notification.

I would love to see some docs on this for `awful.permissions` since it doesn't have a lot of docs in the first place.

elenapan:
Things look smooth with my config as well. 👍

The XDG icon lookup seems to be working using the documented examples. I tested it with `NetworkManager` and `transmission` notifications.

Manually setting the icon with `notify-send` does not work though.
Using the following command:
```shell
notify-send -i "network-receive" "the title" "the message"
```
- In AwesomeWM, the `network-receive` icon was not displayed.
- In other WMs, `dunst` managed to display the icon.

   The `icon_path` dunst configuration variable was hardcoded to:

   ```conf
   icon_path = /usr/share/icons/gnome/32x32/status/:/usr/share/icons/gnome/32x32/devices/
   ```

manilarome:
The icon lookup is finally working!

I just put this snippet in my config

```lua
local menubar = require("menubar")

naughty.connect_signal("request::icon", function(n, context, hints)
    if context ~= "app_icon" then return end

    local path = menubar.utils.lookup_icon(hints.app_icon) or
        menubar.utils.lookup_icon(hints.app_icon:lower())

    if path then
        n.icon = path
    end
end)
```

Elv13:
> Manually setting the icon with notify-send does not work though.

@elenapan Is your `beautiful.icon_theme` set to "gnome"? Did you add the code snippet for the `request::icon`? Also an alternative would be to use the equivalent snippet using `menubar.icon_theme`. All those parallel XDG icon implementations have to eventually be unified as I mentioned in the header. They are all incomplete and very slow.

> I would love to see some docs on this for awful.permissions since it doesn't have a lot of docs in the first place.

@SethBarberee There is another pull request which will land eventually and complete that system. Right now it's in limbo. It "works", but is hard to use. When this gets integrated in the rules, it will make much more sense. This isn't currently near the top of my PR queue (the screen rules is, but keeps being delayed with bug fixes). I have an half working implementation which move `ldoc` parsing into 2 steps. The first pass calls all the custom tag code, then the second pass does the doc. The idea is that it will allow to generate documentation "indexes" for things like "here is everywhere where this `@beautiful` variable is used" section in the theme. It also allows "here are all contexts for this permission and here are all handlers attached to it by default". Those auto-generated lists are what I will base the "good" permission documentation on. As of right now, there is no way to generate those lists the way the documentation is implemented.

elenapan:
> Is your `beautiful.icon_theme` set to "gnome"?
> Did you add the code snippet for the `request::icon`?

Yes and yes.

Turns out some icons just do not appear and `network-receive` is one of them, even though it is accessible in the following locations inside the `gnome` icon theme.
```
16x16/status/network-receive.png
22x22/status/network-receive.png
24x24/status/network-receive.png
256x256/status/network-receive.png
32x32/status/network-receive.png
48x48/status/network-receive.png
```

The `firefox` icon and many others worked.

I tested with the default config as well.

Elv13:
Right. The menubar code doesn't look in the "status" and "actions" directories. This is why it doesn't work. This would require a patch. I think it is out of scope of the PR. So if the notification icon design changes are useful to both of you, then can we merge this and iterate on the icon theme problems later? Can you think of anything the current request::icon API could not handle with the current signature/API?

elenapan:
Understood.

> Can you think of anything the current request::icon API could not handle with the current signature/API?

It already seems super flexible to me (more than I'll ever take advantage of, probably) and it works great 👍.

Elv13:
> and it works great +1.

Awesome! (pun intended). Thanks to both of you for testing, I will merge this for now and we can plan to improve the whole story about the XDG icon spec later on. Maybe I will do a recipe for how to use the GTK implementation itself. This would definitely solve your problem and is the fastest/most efficient way to load icons. However it cannot be mainlined (ever) because any GTK update could potentially render AwesomeWM non-functional. That being said, my config have been using it for 11 years without problems.

PR #3045, opened 2020-03-22
Add a method to select a layout directly, without cycling through the whole list

ArenaL5:
With this PR plus my PRs you already merged, every goal originally intended in #2994 should be covered.

Commits are very small and should be self-explanatory. Is there anything I need to change?

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -372,6 +372,18 @@ awful.keyboard.append_global_keybindings({
                 end
             end
         end,
+    },
+    awful.key {
+        modifiers   = { modkey },
+        keygroup    = "numpad",
+        description = "select layout directly",
+        group       = "layout",
+        on_press    = function (index)
+            local t = awful.screen.focused().selected_tag
+            if t then
+                t.layout = t.layouts[index] or t.layout
+            end
+        end,
     }
 })
 
--- a/lib/awful/key.lua
+++ b/lib/awful/key.lua
@@ -32,6 +32,10 @@ local gobject = require("gears.object")
 --   and its derivative. This is usually the number 1-9 followed by 0.
 -- * **arrows**: The Left/Right/Top/Bottom keys usually located right of the
 --   spacebar.
+-- * **fkeys**: The keys F1 through F12 located at the topmost row of any
+--   keyboard, plus F13 through F35 on specialist keyboards.
+-- * **numpad**: The number keys on the keypad to the right of the letters and
+--   the arrow keys. Not present in every keyboard.
 --
 -- @property keygroup
 
@@ -333,13 +337,29 @@ key.keygroups = {
         {"Right" , "Right" },
         {"Up"    , "Up"    },
         {"Down"  , "Down"  },
-    }
+    },
+    fkeys = {},
+    numpad = {
+        {"#87"   , 1},
+        {"#88"   , 2},
+        {"#89"   , 3},
+        {"#83"   , 4},
+        {"#84"   , 5},
+        {"#85"   , 6},
+        {"#79"   , 7},
+        {"#80"   , 8},
+        {"#81"   , 9},
+        {"#90"   , 10},
+    },
 }
 
 -- Technically, this isn't very popular, but we have been doing this for 12
 -- years and nobody complained too loudly.
 for i = 1, 10 do
     table.insert(key.keygroups.numrow, {"#" .. i + 9, i == 10 and 0 or i})
+end
+for i = 1, 35 do
+    table.insert(key.keygroups.fkeys, {"F" .. i, "F" .. i})
 end
 
 -- Allow key objects to provide more than 1 key.
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -218,6 +218,7 @@ function widget.new(args)
             KP_Multiply      = "Num*",
             KP_Subtract      = "Num-",
             KP_Add           = "Num+",
+            KP_Enter         = "NumEnter",
             -- Some "obvious" entries are necessary for the Escape sequence
             -- and whitespace characters:
             Escape           = "Esc",


lib/awful/key.lua
```diff
@@ -333,14 +337,36 @@ key.keygroups = {
...
+        {"#104"  , "\n"},
+    },
 }
 
 -- Technically, this isn't very popular, but we have been doing this for 12
 -- years and nobody complained too loudly.
 for i = 1, 10 do
     table.insert(key.keygroups.numrow, {"#" .. i + 9, i == 10 and 0 or i})
 end
+for i = 1, 12 do
```
Elv13: F keys goes beyond F12. I am not sure I remember, I think its F24. It needs to be double checked. And yes, I got a keyboard with both rows, they exist (inputs are sooooo fun, right?).

lib/awful/key.lua
```diff
@@ -333,14 +337,36 @@ key.keygroups = {
         {"Right" , "Right" },
         {"Up"    , "Up"    },
         {"Down"  , "Down"  },
-    }
+    },
+    fkeys = {},
+    numpad = {
```
Elv13: A possibility would be to have 2 groups, one with the real characters like this one and one fully indexed?

awesomerc.lua
```diff
@@ -372,6 +372,17 @@ awful.keyboard.append_global_keybindings({
...
             end
         end,
+    },
+    awful.key {
+        modifiers   = { modkey },
+        keygroup    = "numpad",
+        description = "select layout directly",
+        group       = "layout",
+        on_press    = function (index)
+            local keys_to_indices = "1234567890./*-+\n"
```
Elv13: I know I am being super conservative with what gets added to `rc.lua`, sorry about that.

This is *much* nicer than the previous attempt, but I think we can still hide a bit more of the imperative code away from `rc.lua` by adding a `numpad_indexed` group containing this.

awesomerc.lua
```diff
@@ -372,6 +372,17 @@ awful.keyboard.append_global_keybindings({
...
+            local i = string.find(keys_to_indices, index)
+            awful.layout.set(awful.screen.focused().selected_tag.layouts[i])
```
Elv13: This will not universally work. There can be no selected tag and no such layout.

I don't like adding more imperative code, but we have no choice until my `activity` PR gets merged (which adds a safer way to handle this particular `nil`).

```lua
local t = awful.screen.focused().selected_tag
if t then
    t.layout = t.layouts[i] or t.layout
end
```

lib/awful/key.lua
psychon:
```
$ grep -E 'XK_F[0-9]' /usr/include/X11/keysymdef.h 
#define XK_F1                            0xffbe
#define XK_F2                            0xffbf
#define XK_F3                            0xffc0
[snip]
#define XK_F34                           0xffdf
#define XK_F35                           0xffe0
```

lib/awful/key.lua
ArenaL5: I didn't understand, what are you asking for here? Are you talking about `fkeys` or `numpad`?

What do you mean by "fully indexed"? Do you want me to add the keysyms, like this?

```
        {"#87"   , "KP_End"},
        {"#88"   , "KP_Down"},
        {"#89"   , "KP_Next"},
...
```

(I'll use the same commit to both extend the fkeys to F35 and fix this)

awesomerc.lua
ArenaL5: If you want me to. But I'm not so sure about that idea.

Users may decide that "obviously" 0 should go before 1, or Enter should go after 9. Having `keys_to_indices` fully exposed will allow them to customize it.

awesomerc.lua
ArenaL5: Okay. I'll fix this in the same commit as the `keys_to_indices` issue.

awesomerc.lua
Elv13: Then maybe a category to only add 1..9+0 where 0==10, or just = 1..9? I am pretty much against adding imperative logic in `rc.lua`. It should stay/become a config file without imperative business logic. I like the feature, but we need some sort of compromise.

Adding the key groups was a compromise in the first place. The rational was the previous `rc.lua` had an imperative for-loop. I wanted to get rid of it for the same reason why I am asking for this change. However, as proven by this PR, the key groups are a leaky abstraction. Some people will inevitably want to add a limitless number of them for all weird keyboard layouts in the universe. For example, I own a Logitech G15-2nd-gen, G510 and G910. All of them have extra key groups for random tasks. Most mid-range external keyboards also have multimedia keys, but not the same ones and not the same key order. The problem with key groups is just that: there is no right way. In this case our problem is how to handle `/*-+`. Some people might argue you put them in the wrong order too. You see, this is tricky and given this is the default config, we have to please the largest subset of users. So maybe 1..9 or 1..9+10 are the most logic choice? I am looking for a way forward here.

lib/awful/key.lua
ArenaL5: I decided to make this its own commit (1a423e8eb597db5157930ecfba2ed125d39a0036) and squash them later in a rebase. Should make it easier to review this.

_Thirty-five function keys_, that's crazy.

awesomerc.lua
ArenaL5: Okay, I understand your concerns here.

This commit depended on the definition of `numpad`, so I removed it, and pushed a new commit with your 1..9+10 proposal (8aa8c6c9dd413e5e7af42524284c46879acfb191). Is that okay?

awesomerc.lua
ArenaL5: As I said above, I'll reorganize the commits later in a rebase.

Your code is now on commit 571ebdd7183cbf7adb6925f8cb65beb41a03d432, which replaces old commit 0873f70a078041a44aa65228388b35fe3da2be9e (whose summary also was "Add method to select a layout directly").

(Do Git or GitHub ever forget commits when they're not on any branch?)

ArenaL5:
That's normal... especially lately. Don't ever worry about it. This can wait. :slightly_smiling_face:

ArenaL5:
I left the first commit separate because that was a fix to a PR you already merged.

Is this okay?

PR #3050, opened 2020-03-28
New shape: star

ndgnuh:
Draw a star polygon with any number of sides (3 sides at least)

Preview from the doc:

image

Edit: update the preview

### DIFF
--- a/lib/gears/shape.lua
+++ b/lib/gears/shape.lua
@@ -67,7 +67,9 @@ function module.star(cr, width, height, n)
     n = n or 5
     local a = 2 * math.pi / n
 
+
     -- place the star at the center
+    cr:save()
     cr:translate(width/2, height/2)
     cr:rotate(-math.pi/2)
 
@@ -77,8 +79,7 @@ function module.star(cr, width, height, n)
     end
 
     -- restore the context
-    cr:rotate(math.pi/2)
-    cr:translate(-width/2, -height/2)
+    cr:restore()
 
     cr:close_path()
 end
--- a/lib/gears/shape.lua
+++ b/lib/gears/shape.lua
@@ -43,8 +43,45 @@
 local g_matrix = require( "gears.matrix" )
 local unpack   = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
 local atan2    = math.atan2 or math.atan -- lua 5.3 compat
+local min      = math.min
+local cos      = math.cos
+local sin      = math.sin
 
 local module = {}
+
+--- Add a star shape to the current path.
+-- The star size will be the minimum of the given width and weight
+--
+-- @DOC_gears_shape_star_EXAMPLE@
+--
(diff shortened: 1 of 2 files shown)


ndgnuh:
I've just noticed something weird. The example code I wrote was:
```lua
shape.transform(shape.star) : translate(70/2, 70/2)
: rotate(math.pi) : scale(0.5, 0.75)
: translate(-70/2, -70/2) (cr, 70, 70)
```

Somehow even after rebuilding the doc by `make clean` and `make`, it keeps displaying this:
```lua
shape.transform(shape.star) : translate(70/2, 70/2)
:rotate(math.pi/2) (cr, 70, 70)
```

which won't produce the same shape on the image.

lib/gears/shape.lua
```diff
@@ -43,9 +43,46 @@
...
+-- @tparam number n Number of grams (default n = 5 -> pentagram)
+-- @staticfct gears.shape.star
+function module.star(cr, width, height, n)
+    -- use the minimum as size
+    local s = min(width, height) / 2
+
+    -- draw pentagram by default
+    n = n or 5
+    local a = 2 * math.pi / n
+
```
psychon:
```suggestion
cr:save()
```

lib/gears/shape.lua
```diff
@@ -43,9 +43,46 @@
...
+    cr:rotate(-math.pi/2)
+
+    for i = 0,(n - 1) do
+        cr:line_to(s   * cos((i      ) * a), s   * sin((i      ) * a))
+        cr:line_to(s/2 * cos((i + 0.5) * a), s/2 * sin((i + 0.5) * a))
+    end
+
+    -- restore the context
+    cr:rotate(math.pi/2)
+    cr:translate(-width/2, -height/2)
```
psychon: You can replace this with `cr:restore()` (and add the `cr:save()` that I suggested above)

ndgnuh: Thanks, I didn't know I could do that! I pushed another commit based on your suggestion.

lib/gears/shape.lua
```diff
@@ -67,7 +67,9 @@ function module.star(cr, width, height, n)
     n = n or 5
     local a = 2 * math.pi / n
 
+
```
psychon: Now you have two empty lines here (both line 69 and line 70 are empty)

psychon:
Looks good to me :+1: 
(But I cannot help you with your doc question, sorry)

lib/gears/shape.lua
ndgnuh: I'm gonna remove it right away! :+1:

Elv13:
Thanks for contributing this!

> Somehow even after rebuilding the doc by make clean and make, it keeps displaying this:

Our meta generator does funny things. I have a zombie PR somewhere to rewrite parts of it to solve your issue, but it isn't going anywhere. The current one "work", just weirdly.

PR #3053, opened 2020-04-01
Apply placement margins on awful.tooltip

manilarome:
With this PR, you can set the margins/add a gap between the `awful.tooltip` and its parent if using the outside mode. This is my very first PR and contribution here, so please go easy on me. But of course feedbacks are welcome! :) Thanks!

image

### DIFF
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -553,7 +553,7 @@ end
 
 --- Set the margins between the tooltip and its parent.
 --
--- @DOC_awful_gaps_EXAMPLE@
+-- @DOC_awful_tooltip_gaps_EXAMPLE@
 --
 -- @property gaps
 -- @tparam number|table New margins value.
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -615,15 +615,15 @@ end
 --   seconds.
 -- @tparam[opt=apply_dpi(5)] integer args.margin_leftright The left/right margin for the text.
 -- @tparam[opt=apply_dpi(3)] integer args.margin_topbottom The top/bottom margin for the text.
--- @tparam[opt=nil] gears.shape args.shape The shape
--- @tparam[opt] string args.bg The background color
--- @tparam[opt] string args.fg The foreground color
--- @tparam[opt] string args.border_color The tooltip border color
--- @tparam[opt] number args.border_width The tooltip border width
--- @tparam[opt] string args.align The horizontal alignment
--- @tparam[opt] string args.font The tooltip font
--- @tparam[opt] number args.opacity The tooltip opacity
--- @tparam[opt] table|number args.gaps The tooltip margins
+-- @tparam[opt=nil] gears.shape args.shape The shape.
+-- @tparam[opt] string args.bg The background color.
+-- @tparam[opt] string args.fg The foreground color.
+-- @tparam[opt] string args.border_color The tooltip border color.
+-- @tparam[opt] number args.border_width The tooltip border width.
+-- @tparam[opt] string args.align The horizontal alignment.
+-- @tparam[opt] string args.font The tooltip font.
+-- @tparam[opt] number args.opacity The tooltip opacity.
+-- @tparam[opt] table|number args.gaps The tooltip margins.
 -- @treturn awful.tooltip The created tooltip.
 -- @see add_to_object
 -- @see timeout
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -563,7 +563,7 @@ function tooltip:set_gaps(val)
     self._private.tooltip_gaps = val
 end
 
-function tooltip:get_gaps(val)
+function tooltip:get_gaps()
     return self._private.tooltip_gaps
 end
 
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -114,7 +114,7 @@ local offset = {
 
 --- The default tooltip shape.
 -- The default shape for all tooltips is a rectangle. However, by setting
--- this variable they can default to rounded rectangle or stretched octogons.
+-- this variable they can default to rounded rectangle or stretched octagons.
 -- @beautiful beautiful.tooltip_shape
 -- @tparam[opt=gears.shape.rectangle] gears.shape shape A `gears.shape`
 --  compatible function
--- a/lib/awful/tooltip.lua
+++ b/lib/awful/tooltip.lua
@@ -551,6 +551,22 @@ function tooltip:set_margins_topbottom(val)
     self:set_margin_topbottom(val)
 end
 
+--- Set the margins between the tooltip and its parent
+--
+-- @DOC_awful_tooltip_gaps_EXAMPLE@
+--
+-- @property tooltip_gaps
+-- @tparam number|table New margins value
+-- @propemits true false
+
+function tooltip:set_gaps(val)
+    self._private.tooltip_gaps = val
+end
+
+function tooltip:get_gaps(val)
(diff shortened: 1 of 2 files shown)


lib/awful/tooltip.lua
```diff
@@ -621,6 +627,10 @@ function tooltip.new(args)
     self._private.align   = args.align or beautiful.tooltip_align  or "right"
     self._private.shape   = args.shape or beautiful.tooltip_shape
                                 or shape.rectangle
+    self._private.tooltip_gaps  = args.tooltip_gaps or beautiful.tooltip_gaps or {
+        left = args.tooltip_gaps or 0, right  = args.tooltip_gaps or 0,
+        top  = args.tooltip_gaps or 0, bottom = args.tooltip_gaps or 0
+    }
```
actionless, line 633: > **args.tooltip_gaps** or beautiful.tooltip_gaps or { left = **args.tooltip_gaps** or 0, ... }

i can't undeerstand what do you mean with that

manilarome: @actionless is there a better way to set the margins? I just copied the table from awful.placement https://github.com/awesomeWM/awesome/blob/f997009ff4c4cb73c1364126e6df7867832705d3/lib/awful/placement.lua#L317-L320

lib/awful/tooltip.lua
```diff
@@ -602,6 +607,7 @@ end
 -- @tparam[opt] string args.align The horizontal alignment
 -- @tparam[opt] string args.font The tooltip font
 -- @tparam[opt] number args.opacity The tooltip opacity
+-- @tparam[opt] table|number args.tooltip_gaps The tooltip margins
```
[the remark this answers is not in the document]
manilarome: Will do! Anyway, should I change the property name to `gaps` or should we stay with `tooltip_gaps?`

lib/awful/tooltip.lua
```diff
@@ -546,6 +551,22 @@ function tooltip:set_margins_topbottom(val)
...
+--- Set the margins between the tooltip and its parent
+--
+-- @DOC_awful_tooltip_gaps_EXAMPLE@
+--
+-- @property tooltip_gaps
+-- @tparam number|table New margins value
+-- @propemits true false
+
+function tooltip:set_gaps(val)
+    self._private.tooltip_gaps = val
```
Elv13, line 563: This will need a little more boilerplate code here:

Emit a signal:
```lua
self:emit_signal("property::gaps", val)
```

And change the geometry dynamically:

```lua
if self._private.visible then
    set_geometry(self)
end
```

This way code built on top of the tooltip API will be more reactive/dynamic.

lib/awful/tooltip.lua
```diff
@@ -546,6 +551,22 @@ function tooltip:set_margins_topbottom(val)
     self:set_margin_topbottom(val)
 end
 
+--- Set the margins between the tooltip and its parent
+--
+-- @DOC_awful_tooltip_gaps_EXAMPLE@
+--
+-- @property tooltip_gaps
+-- @tparam number|table New margins value
```
Elv13, line 559: Add a `.` at the end of the line. This is annoying, but ldoc is often annoying.

tests/examples/awful/tooltip/gaps.lua
```diff
@@ -0,0 +1,30 @@
...
+require("gears.timer").run_delayed_calls_now()
+
+for _, side in ipairs{ "left", "right", "bottom", "top" } do
+    local tt = awful.tooltip {
+        text = side,
+        objects = {wb},
+        mode = "outside",
+        preferred_positions = {side},
+        preferred_alignments = {"back"},
+        tooltip_gaps = 5
```
Elv13, line 23: It's called `gaps` now, right?

lib/awful/tooltip.lua
```diff
@@ -546,6 +551,22 @@ function tooltip:set_margins_topbottom(val)
     self:set_margin_topbottom(val)
 end
 
+--- Set the margins between the tooltip and its parent
+--
+-- @DOC_awful_tooltip_gaps_EXAMPLE@
+--
+-- @property tooltip_gaps
```
Elv13, line 558: The property is called `gaps`. `set_gaps` will make `my_tooltip.gaps = 42` work. The fact that `self._private.tooltip_gaps` exists rather than `self._private.gaps` will have some little side effects, but since this is undocumented, I don't mind much.

tests/examples/awful/tooltip/gaps.lua
manilarome: Yeah haha. Anyway, what about the theme variable, `beautiful.tooltip_gaps`? Should we stay with that name? I feel that `beautiful.gaps` is a bad name for that variable.

lib/awful/tooltip.lua
manilarome: Should I add `.` to all the lines without it?

lib/awful/tooltip.lua
manilarome: Thanks! I realized that the best thing to do after creating all those setups is to contribute to the project itself.

tests/examples/awful/tooltip/gaps.lua
Elv13: > beautiful.tooltip_gaps

Yes, that's correct. Beautiful "naming convention" is to go from left to right with module prefix. So in this case, `tooltip_` is the prefix. Some properties also have suffixes for the state.

Another example would be `bg`. You have `beautiful.tasklist_fg_focus` and `beautiful.tasklist_fg_normal`. They fallback to `beautiful.bg_focus` and `beautiful.bg_normal` respectively. This is a good example of the namespace/module prefix + state suffix is applied.

lib/awful/tooltip.lua
Elv13: Just the lines with a description and the first line with the `---`. For the later, `ldoc` uses that for the summary in the table of content. It will put everything until the first `.`, even if it is over multiple lines. For the `@tsomething`, not having the `.` usually works, but sometime triggers weird bugs. It is better to just have it everywhere.

manilarome:
I don't know if I did the right thing. I just `git rebase -i HEAD~8` then force push. Pardon me for my lack of git knowledge.

Elv13:
> I don't know if I did the right thing. I just git rebase -i HEAD~8 then force push. Pardon me for my lack of git knowledge.

Nearly. In the editor window you see when doing `git rebase origin/master` (or HEAD offsets), you have to edit the lines and set `f` or `s` to "fixup" or "squash" commits. You can also re-order the lines to re-order the commit. It is very powerful, but not super intuitive. I wrote an ncurses TUI for that some time ago https://gist.github.com/Elv13/eece84e72f815a231a746fca64fa339f

In any case, I will do it for you since I can use the GitHub UX for it (admin power!). Thanks a lot for contributing upstream!

lib/awful/tooltip.lua
actionless: @Elv13 this still looks like it should be fixed before merging it

Elv13: It was merged, but this looks fine to me

actionless: it could be safely simplified to just ` = args.tooltip_gaps or beautiful.tooltip_gaps or 0`

manilarome: Yeah, I think it should be `= args.tooltip_gaps or beautiful.tooltip_gaps or 0` because awful.placement already handles the table part. Also, I forgot to add

```lua
self:emit_signal("property::gaps", val)
if self._private.visible then
    set_geometry(self)
end
```

in the set_gaps(). Sorry for the mess. :grimacing:

Issue #1547, opened 2017-02-09
v4.1 changelog draft


Elv13:
Awesome v4.1 is the first stable release for the Awesome 4.0 API. It adds
non-breaking new features and fix bugs. The main purpose of the release is to provide additional stability while also adding features submitted by our contributors. This release contains about 350 commits by 35 contributors, including many new developers. Thanks a lot.

## New features

The shape API has been extended to both client, notifications and wibox.

The prompt now supports syntax highlight and more advanced key hooks.

The prompt widget gained many new themes variables.

There is a new 2D grid layout with rowspan and colspan support.

There is a new `only_on_screen` container to make it easier to share wiboxes across multiple screens.

Various documentation improvements. Thanks for the feedbacks.

The taglist now has `volatile` theme variables.

There is now extra enviroment variables such as `AWESOME_ICON_PATH` and
`AWESOME_THEMES_PATH` for those who prefer not installing Awesome.

Dynamic "C" Lua libraries are now detected like pure Lua ones.

`gears.timer` gained many new constructor arguments to make it easier to use.

Input shape mask are now supported. It is possible to create a wibox with
passthough inputs.

There is a new `awful.widget.client_icon` widget capable of fetching icons of different sizes.

## New theme variables

This release adds a ton of new theme variables to make Awesome prettier. We also thank all users who [submitted screenshot](https://github.com/awesomeWM/awesome/issues/1395).

```
theme.arcchart_thickness
theme.fullscreen_hide_border
theme.hotkeys_bg
theme.hotkeys_border_colo
theme.hotkeys_border_width
theme.hotkeys_description_font
theme.hotkeys_fg
theme.hotkeys_font
theme.hotkeys_group_margin
theme.hotkeys_label_bg
theme.hotkeys_label_fg
theme.hotkeys_modifiers_fg
theme.hotkeys_shape
theme.maximized_honor_padding
theme.notification_bg
theme.notification_border_color
theme.notification_border_width
theme.notification_fg
theme.notification_font
theme.notification_height
theme.notification_margin
theme.notification_opacity
theme.notification_shape
theme.notification_width
theme.prompt_bg_cursor
theme.prompt_bg
theme.prompt_fg_cursor
theme.prompt_fg
theme.prompt_font
theme.taglist_bg_volatile
theme.taglist_fg_volatile
theme.taglist_shape_border_color_volatile
theme.taglist_shape_border_width_volatile
theme.taglist_shape_volatile
theme.taglist_spacing
[...]

PR #3065, opened 2020-04-17
Move variable declarations from header to C file to fix build with GCC 10

reinerh:
GCC 10 builds with -fno-common by default, which causes linker errors when
variables are declared in header files and included in multiple places.

See also: https://gcc.gnu.org/gcc-10/porting_to.html
Reported here: https://bugs.debian.org/957027

### DIFF
--- a/common/lualib.c
+++ b/common/lualib.c
@@ -20,6 +20,8 @@
 #include "common/lualib.h"
 #include "luaa.h"
 
+lua_CFunction lualib_dofunction_on_error;
+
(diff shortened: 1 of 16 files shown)


blueyed:
This would make sense to have in a stable branch / patch release / backported probably - but there is no such thing currently.
(found some discussion in https://github.com/awesomeWM/awesome/issues/1547#issuecomment-279101099; mainly for reference / stating that it would be useful for distributions etc)

PR #3069, opened 2020-04-21
Don't force defaults in menubar

aDogCalledSpot:
Currently, all members of `menubar.prompt_args` will be ignored if there is a default for them.

I suggest instead using all prompt_args and falling back to defaults if the required argument is not set by the user.

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -121,7 +121,7 @@ menubar.left_label = "◀◀"
 local list_interspace = theme.xresources.apply_dpi(4) * 3
 
 --- Allows user to specify custom parameters for prompt.run function
--- (like colors).
+-- (like colors). These will override default parameters.
 -- @see awful.prompt
 menubar.prompt_args = {}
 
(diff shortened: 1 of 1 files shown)


actionless:
IMO it also requires updating the documentation for `prompt_args`

aDogCalledSpot:
I added a short note that the value will override defaults. I could elaborate if necessary.

lib/menubar/init.lua
```diff
@@ -121,7 +121,7 @@ menubar.left_label = "◀◀"
 local list_interspace = theme.xresources.apply_dpi(4) * 3
 
 --- Allows user to specify custom parameters for prompt.run function
--- (like colors).
+-- (like colors). These will override default parameters.
```
actionless, line 124: but with your patch instead of overriding the default, it will _merge_ with defaults

aDogCalledSpot:
```suggestion
-- (like colors). This will merge with the default parameters, overriding affected values.
```

PR #3070, opened 2020-04-21
Menubar match empty

aDogCalledSpot:
For minimalist setups, it looks good when the you can only see the suggestions when you start needing them i.e. when you enter something. This allows the user to set the property `menubar.match_empty = false` if the menubar should be completely empty before anything is typed.

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -90,6 +90,10 @@ menubar.cache_entries = true
 -- @tfield[opt=true] boolean show_categories
 menubar.show_categories = true
 
+--- When false will hide results if the current query is empty
+-- @tfield[opt=true] boolean match_empty
+menubar.match_empty = true
+
 --- Specifies the geometry of the menubar. This is a table with the keys
 -- x, y, width and height. Missing values are replaced via the screen's
 -- geometry. However, missing height is replaced by the font size.
@@ -310,7 +314,7 @@ local function menulist_update(scr)
     end
 
     -- Add the applications according to their name and cmdline
-    for _, v in ipairs(menubar.menu_entries) do
+    local add_entry = function(v)
         v.focused = false
         if not current_category or v.category == current_category then
 
@@ -341,6 +345,14 @@ local function menulist_update(scr)
             end
         end
     end
+
+    -- Add entries if required
+    if query ~= "" or menubar.match_empty then
+        for _, v in ipairs(menubar.menu_entries) do
+            add_entry(v)
+        end
+    end
+
 
     local function compare_counts(a, b)
         if a.prio == b.prio then
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -314,34 +314,34 @@ local function menulist_update(scr)
     end
 
     -- Add the applications according to their name and cmdline
-    local add_entry = function(v)
-        v.focused = false
-        if not current_category or v.category == current_category then
+    local add_entry = function(entry)
+        entry.focused = false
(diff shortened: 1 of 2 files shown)


actionless:
i see it's changing the default behavior, which doesn't seems fine

aDogCalledSpot:
The default value was set incorrectly, should be fixed now

lib/menubar/init.lua
```diff
@@ -90,6 +90,10 @@ menubar.cache_entries = true
 -- @tfield[opt=true] boolean show_categories
 menubar.show_categories = true
 
+--- When true will hide results if the current query is empty
```
actionless, line 93: this docstring doesn't match to what the code do

Elv13:
It would be nice to have a test for this one. It can be something simple in `tests/` showing the menubar when the option is set.

aDogCalledSpot:
I've added it to the menubar tests

tests/test-menubar.lua
```diff
@@ -37,22 +37,33 @@ do
...
     function(count)
-        -- Just show the menubar and hide it.
-        -- TODO: Write a proper test. But for the mean time this is better than
-        -- nothing (and tells us when errors are thrown).
-
-        if count == 1 then
-            menubar.show()
-        end
+        -- Show and hide with defaults
+        show_menubar_and_hide(count)
```
Elv13, line 60: you need to `return true` when the test passed, otherwise it retries and timeout. You can run these in your computer by running `tests/run.sh tests/test-menubar.lua`

Elv13:
Thanks for adding the change. The build pass now. However the coverage isn't seeing your change. I am not working on Awesome this month (and probably next) and I don't have a dev environment on this (new) laptop (yet). Is it possible there is some caching going on that bypasses the `menubar.show_empty = false`?

If that's the case, the way to detect the property change is usually using a metatable or `gears.object`. For the metatable way, the code looks like:

```lua

local my_real_value = true

local t = setmetatable({}, {
    __index = function(_, k)
        if k == "my_real_value" then return my_real_value end
    end,
    __newindex = function(_, k, v)
        if k == "my_real_value" then
            my_real_value = v
            -- code to dismiss cache here
        end
    end
})

-- Usage example: this will call the code to dismiss the cache
t.my_real_value = false
```

aDogCalledSpot:
Could it be that the coverage scan uses Lua 5.3?  The top of test-menubar.lua say this:

```
-- XXX We are building Lua 5.3 with -DLUA_USE_APICHECK=1 and this catches some
-- bugs in lgi. Thus, do not run this test with Lua 5.3 until lgi is fixed.
-- We run into the same bug when doing code-coverage analysis, supposedly
-- because the additional GC activity means that something which is GC-able when
-- it should not gets collected too early.
-- This test also sporadically errors on LuaJIT ("attempt to call a number
-- value"). This might be a different issue, but still, disable the test.
if _VERSION == "Lua 5.3" or debug.gethook() or jit then --luacheck: globals jit
    print("Skipping this test since it would just fail.")
    runner.run_steps { function() return true end }
    return
end
```

actionless:
yup, codecov build config uses lua 5.3

aDogCalledSpot:
So it isn't running the test. You can also see that it only runs the lines I mentioned in my comment before. There's nothing I can do about that.

actionless:
@awesomeWM/write-access should we move codecov to a different travis build env?

Elv13:
> should we move codecov to a different travis build env?

Probably.

PR #3071, opened 2020-04-21
fix(gears: matcher: default_matcher): handle sidecase of string.match-ing the empty string ('')

actionless:
without the patch the test would fail:

```
Error → ./spec/gears/matcher_spec.lua @ 37
gears.matcher not matching by empty string value
./spec/gears/matcher_spec.lua:41: Expected objects to be the same.
Passed in:
(boolean) true
Expected:
(boolean) false

stack traceback:
        ./spec/gears/matcher_spec.lua:41: in function <./spec/gears/matcher_spec.lua:37>
```

### DIFF
--- a/lib/gears/matcher.lua
+++ b/lib/gears/matcher.lua
@@ -80,10 +80,10 @@ local matcher = {}
 
 local function default_matcher(a, b)
     local result = a == b
-    if result then return result end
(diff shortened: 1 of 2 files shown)


Elv13:
thanks

PR #3073, opened 2020-04-23
feat(awful: placement): add "skip_fullscreen"

actionless:


actionless:
not sure about the naming, mb `force_fullscreen` might be better

actionless:
for wiboxes and other non-clients it will just skip always, as they'll have fullscreen always `nil` (unset)

lib/awful/placement.lua
```diff
@@ -1624,6 +1624,26 @@ function placement.restore(d, args)
...
+    args = add_context(args, "skip_fullscreen")
+    d = d or capi.client.focus
+
+    if d.fullscreen then
+        return {get_screen(d.screen).geometry, {}, true}
+    else
+        local ngeo = geometry_common(d, args)
+        remove_border(d, args, ngeo)
+        geometry_common(d, args, ngeo)
+        return fix_new_geometry(ngeo, args, true)
```
actionless, line 1643: i only not sure, if all those 4 lines are needed here

PR #3074, opened 2020-04-23
fix(naughty: widget: _markup): correctly un-escape "&lt;span property=value&gt;"

### LINKED ISSUE #3049: Coloring notification text with markup makes it not appear
The following command should create a notification with a red `message`, but instead the `message` does not appear at all.
```shell
notify-send "the title" "<span foreground='#ff0000'>the red message</span>"
```

`naughty` is also affected.

The commit that breaks this: b6ce95cd

actionless:
fixes #3049

### DIFF
--- a/lib/naughty/widget/_markup.lua
+++ b/lib/naughty/widget/_markup.lua
@@ -9,7 +9,7 @@ local escape_pattern = "[<>&]"
 local escape_subs    = { ['<'] = "&lt;", ['>'] = "&gt;", ['&'] = "&amp;" }
 
 -- Also reverse escaping some allowed tags because people actually use them.
-local escape_undo = {['&lt;span '] = "<span "}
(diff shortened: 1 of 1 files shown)


psychon:
I neither know what the current code is doing (why does `<span` need special treatment since the naughty rewrite?!) nor do I really know what the patch is doing. Sorry, I'm out.

(Random idea: Can this get some unit tests? Or would that run into `require()` problems?)

actionless:
btw it got confirmed what it fixes the original issue: https://github.com/awesomeWM/awesome/issues/3049#issuecomment-618421762

however right now i'm a bit lazy writing the tests, as i'm busy trying to understand/debug other things of new notification API right now

actionless:
yup i remember, i think i was even reviewing those changes to markup, but such kind of regex thing is very easy to miss on review

PR #3082, opened 2020-04-26
fix wibox declarative opacity, Fixes #2997

### LINKED ISSUE #2997: Wibox not setting opacity with declarative syntax

```
awesome v4.3-731-g81d652114 (Too long)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

Create a wibox, setting the opacity using the declarative syntax.

```
    local w = wibox {
        x       = 0,
        y       = 0,
        visible = true,
        ontop   = true,
        opacity = 0.8,
        height  = 200,
        width   = 200
    }
    print(w.opacity) --Prints 1.0
   w.opacity = 0.8
   print(w.opacity) --Prints 0.8
```

This worked a few commits ago, probably broke between `v4.3-731-g81d652114` and `v4.3-694-gcc4ebd80f`.

**Actual result:**

The opacity only gets set when using the imperative syntax.

**Expected result:**

The wibox should set the assigned opacity on creation.

unai-ndz:
I finally did a git bisect, the commit which introduced this issue is: https://github.com/awesomeWM/awesome/commit/04c757322c079a8b9f13a51d0dc4c37e5a5b7965

The test I'm using:
```lua

local wb = wibox {
    opacity = 0.50,
    visible   = true,
    height  = 200,
    width   = 200,
}

local steps = {
    function()
        local rounded = tonumber(string.format("%.3f", wb.opacity))
        assert(rounded == 0.5)
        return true
    end,
}

runner.run_steps(steps)
```

### DIFF
--- a/lib/wibox/init.lua
+++ b/lib/wibox/init.lua
@@ -384,6 +384,10 @@ local function new(args)
         ret.border_color = args.border_color
     end
 
+    if args.opacity then
+        ret.opacity = args.opacity
+    end
+
     if args.input_passthrough then
         ret.input_passthrough = args.input_passthrough
     end


actionless:
for some reason, travis is not picking up the PR to build

psychon:
Random drive-by comment: Any other properties that got lost / broken?

unai-ndz:
AFAIK the `wibox` is fine but I think `border_marked` should be `border_color_marked` like the rest of the client border colors which got changed by that commit. Actually only the default theme uses `border_marked` while the other themes use `border_color_marked`.

`border_marked` is also documented two times, in `docs/common/client_theme.ldoc` and in `lib/awful/client.lua`.

Should I make a PR for that?

Also what is the usual "protocol" regarding tests, which things should get tested and which ones not?
This one for example seems too specific too add to the tests but maybe worth it to avoid regressions.

Elv13:
> Should I make a PR for that?

Probably. The refactoring to "fix" the border_color/border_width beautiful variables caused side effects left and right. There is probably more lurking as @psychon mentions.

> Also what is the usual "protocol" regarding tests, which things should get tested and which ones not?
This one for example seems too specific too add to the tests but maybe worth it to avoid regressions.

It is preferred to always have tests. We are slowly approaching 90% *code* coverage from 0% ~5 years ago. However the state of the "does this feature *really* works" coverage isn't as good yet. A lot of that rely on screenshots taken using the SVG backend rather than the X11 backend. It would not catch this bug. The integration tests also run without compositing, so opacity isn't "visible" there either. I think checking that `w.opacity = 0.5; assert(w.opacity == 0.5)` would be enough in this case.

lib/wibox/init.lua
```diff
@@ -384,6 +384,10 @@ local function new(args)
         ret.border_color = args.border_color
     end
 
+    if args.opacity then
```
Elv13, line 387: I *think* adding `opacity` to the `force_forward` set will be enough. The code around this could probably be refactored to use `force_forward` too.

unai-ndz: I'm not sure I understand you correctly but just adding `opacity = true` to `force_forward` does not make a difference, the `__newindex` metatable does not run unless we set `ret.opacity = args.opacity` when using the declarative syntax. The imperative one works either way.

unai-ndz:
Any idea why this:

```lua
local wb = wibox {
    opacity = 0.5,
    visible = true,
    height  = 200,
    width   = 200,
}
print(wb.opacity)
wb.opacity = 0.5
print(wb.opacity)

local steps = {
    function()
        print(wb.opacity)
        wb.opacity = 0.5
        print(wb.opacity)
        return true
    end,
}

runner.run_steps(steps)
```

prints this?

```
0.5
0.5
0.49999999988358
0.5
```

psychon:
Would "rounding issue" as an answer to that question count?

If not: Setting a window's `opacity` property eventually ends up here:
https://github.com/awesomeWM/awesome/blob/29f6387defd18c1d44b7d2976b48847fe80686b3/xwindow.c#L209-L211
This causes the property to actually be set on the X11 window. This change causes a notify to be generated by the X11 server to tell us that the value changed. Before this notification is handled, AwesomeWM returns a value that it cached somewhere internally. When this notification is later handled (in the next main loop iteration), we end up here:
https://github.com/awesomeWM/awesome/blob/29f6387defd18c1d44b7d2976b48847fe80686b3/xwindow.c#L191

So, my answer is still "rounding issue", but this time with a bit more information: `(double)(uint32_t)(0.5 * 0xffffffff) / (double) 0xffffffff` seems to produce such a rounding error.

In GDB:
```
 p (double)(unsigned int)(0.5 * 0xffffffff) / (double) 0xffffffff
$2 = 0.49999999988358468
```

PR #3084, opened 2020-04-27
Border color consistency

unai-ndz:
Rename `border_marked` to `border_color_marked` like the rest of `border_color`s.
Also fix some issues in the client border color documentation.

### DIFF
--- a/docs/common/client_theme.ldoc
+++ b/docs/common/client_theme.ldoc
@@ -485,7 +485,7 @@
 /**
  * The marked clients border color.
  * Note that only solid colors are supported.
- * @beautiful beautiful.border_marked
+ * @beautiful beautiful.border_color_marked
  * @param color
  */
 
--- a/docs/common/client_theme.ldoc
+++ b/docs/common/client_theme.ldoc
@@ -2,6 +2,15 @@
 
 
 /**
+ * The border color when the client is marked.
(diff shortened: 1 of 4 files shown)


docs/common/client_theme.ldoc
```diff
@@ -1,6 +1,15 @@
 */
 
 
+/**
+ * The border color when the client is marked.
+ * It has priority over the rest of beautiful border color properties.
+ * Note that only solid colors are supported.
```
unai-ndz, line 7: The `border_color_`... properties don't support solid colors, not only this one.
Maybe `@param solid_color` should be used instead of adding this note to every one of them?

lib/awful/permissions/init.lua
```diff
@@ -637,8 +637,8 @@ function permissions.update_border(c, context)
         -- never have been added to the core. The documentation claims it works,
         -- even if it has been broken for 90% of AwesomeWM releases ever since
         -- it was added.
-        if c.marked and beautiful.border_marked then
-            c._border_color = beautiful.border_marked
+        if c.marked and beautiful.border_color_marked then
+            c._border_color = beautiful.border_color_marked
```
actionless, line 641: what about still leaving `beautiful.border_marked` and adding deprecation warning?

Issue #3091, opened 2020-05-04
[Documentation] Missing argument in "destroyed" signal of naughty.notification


manilarome:
Output of `awesome --version`:

```
awesome v4.3-826-g29f6387de (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

**How to reproduce the issue:**

I'm playing with `naughty` API and I want to execute something if the notification expired, so I used the `destroyed` signal for this and get the `reason` of closing. 

![image](https://user-images.githubusercontent.com/40349590/80933285-a7cdae80-8df5-11ea-836e-7692b2010f9a.png)

Here's the relevant code:

```lua
local naughty_expired = function(n)
	n:connect_signal(
		'destroyed',
		function(reason, keep_visble)
			if reason == 1 then
				-- Execute something if the notification expired
			end
		end
	)
end

-- Connect to naughty
naughty.connect_signal(
	"request::display",
	function(n)
		naughty_expired(n)
	end
)
```

**Actual result:**

The `reason` arg of `destroyed` signal should return a [number](https://developer.gnome.org/notification-spec/#signals). Instead, it returns a table.

**Expected result:**

`reason` arg should return a number(closed reason).

So I tried to debug it and it turns out the first arg should be `self` right? I'm just guessing, though. Haha

psychon:
To quote from https://github.com/awesomeWM/awesome/pull/3096#issuecomment-626279064:

> I think we need a proper plan before merging (or not merging) this. All object signals have an implicit self first argument. A minority of the @signal doc has it, but the vast majority doesn't.
> [...]
> The alternative is to modify config.ld to auto-add the implicit self when @classsignal is not present.

PR #3096, opened 2020-05-10
fix missing arg in naughty.notification detroyed signal

manilarome:
fixes https://github.com/awesomeWM/awesome/issues/3091. Just tell me if I need to change the argument's description.

### DIFF
--- a/lib/naughty/notification.lua
+++ b/lib/naughty/notification.lua
@@ -419,6 +419,7 @@ local notification = {}
 
 --- Emitted when the notification is destroyed.
 -- @signal destroyed
+-- @tparam table self The current transaction object.
 -- @tparam number reason Why it was destroyed
 -- @tparam boolean keep_visible If it was kept visible.
 -- @see naughty.notification_closed_reason


psychon:
I'm not sure about this. All signals on objects have the objects as first argument. For example, the docs for `swapped` on `screen` say there are two argument:
- The other screen
- `is_source`

This just implicitly assumes that the object is emitted on a screen which is also, well, "participating".

manilarome:
Oh, okay. Thanks for the clarification. Maybe I'm just using the `destroyed` signal in wrong way haha. I'll just close this and the issue. Thanks again!

Elv13:
Thanks for this PR.

I think we need a proper plan before merging (or not merging) this. All object signals have an implicit `self` first argument. A minority of the `@signal` doc has it, but the vast majority doesn't.

We have `@classsignal` [here](https://github.com/awesomeWM/awesome/blob/master/docs/config.ld#L352) to specify signals that are only class level and not object+class. It should be assumed that all `@signal` without this attribute have a `self`.

For the object property, we already have the magic to auto-add this everywhere. I guess we need the same magic to fix this #3091. This would go into `config.ld`. That file is very hard to read because `ldoc` doesn't expose the Lua stdlib, so everything re-invent the API wheel in raw code (because `require()` and `dofile()` are not there, it also has to be a giant file... Then there is the weird way the HTML template calls the code in `config.ld` that make that file even harder to understand.

So if we merge this, then we have to go trough the entire API and add all implicit `self`. I think this will be a real nightmare to maintain. The alternative is to modify `config.ld` to auto-add the implicit `self` when `@classsignal` is not present. This means we still have to `grep` the codebase `@signal` to make sure all of them are properly using (or not using) `@classsignal`. Then also make sure the ones with `self` get fixed.

Sorry for turning a simple fix into a large and boring multi hour project. I wont have time to do it in the near term. Patches would be appreciated. But fixing the "there is missing self in the signal doc" is really a project we have to undertake to improve our doc and help ease the learning curve. Implicit concepts are horrible documentation bugs.

manilarome:
I see. Thanks for the detailed explanation! I'll keep that in mind! I really should read and learn more about the API and the source code.

psychon:
Closing to prevent this from getting accidentally merged. I'll reopen #3091 instead.

PR #3103, opened 2020-05-17
fix(naughty: layout: box): take preset.padding and beautiful.notification_spacing into account

### LINKED ISSUE #3051: Variable to increase gap between notification and the edge of the screen
Hello,

There is no way to increase the gap between a notification and the edge of the screen.

I did read the APIDoc and except `notification_spacing`, which increase the gap between each notification, I did not find anything.

Is it possible to add a new theme variable ?

actionless:
Fixes #3051

### DIFF
--- a/lib/naughty/layout/box.lua
+++ b/lib/naughty/layout/box.lua
@@ -57,12 +57,23 @@ capi.screen.connect_signal("removed", function(scr)
 end)
 
 local function get_spacing()
-    local margin = beautiful.notification_spacing or 2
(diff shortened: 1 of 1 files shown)


actionless:
actually fixing/testing it really properly would be quite a thing, as currently offset could be applied only to the first notification, so given the case when we have multiple notifications with different `preset.padding` we'll need to store the offset of previous (upper) notification in the stack, and compute offset based on delta of offset of previous notification and preset.padding of current

PR #3125, opened 2020-07-02
feat(awful: hotkeys_popup: vim): add vimdiff

actionless:


lib/awful/hotkeys_popup/keys/vim.lua
```diff
@@ -173,6 +173,16 @@ local vim_keys = {
             zR="open all",
         }
     }},
+
+    ["VIM: diff"] = {{
```
actionless, line 177: or would be better `"VIM: vimdiff"`?

psychon (approved):

PR #3131, opened 2020-07-15
modular default config

seniorivn:
split rc.lua into a modular library «default» and small rc.lua with a bunch of imports


seniorivn:
As mentioned in #442 
And after merge of #2507
I think it's time to make default rc.lua less monolithic
This way:
- users will be encouraged to make more flexible configs
- users will be able to replace only those parts of default config that they don't need and benefit from updates without any additional work.
- it will be possible to add examples of good hardcore modifications to awesome config commented in rc.lua by default like
```lua

-- require("default.panel_with_a_lot_of_widgets")
-- require("default.icons_only_taglist")
```

lib/default/default/common.lua
```diff
@@ -0,0 +1,42 @@
...
+local hotkeys_popup = require("awful.hotkeys_popup")
+local common = {}
+
+
+function common.init(args)
+   local args = args or {}
+   -- {{{ Variable definitions
+
+   -- This is used later as the default terminal and editor to run.
+   common.terminal = args.terminal or "xterm"
```
actionless, line 12: idk what is the faith of this PR as a whole, however if the idea itself will be approved in this way, i suggest using smth like `common._data.terminal` or so instead of storing variables just on the first module level of that lib

seniorivn: noted
btw, what do you think of the idea itself?

actionless: i'm using a similar layout in my config, but with 2 modules instead of 1, as it's a bit bigger than default one

but at the same time idk how would it apply for the most of users in terms of ease of changing some small value in default config (as it will require making a copy of one of default-submodules into local dir for doing that change)

seniorivn: the fact that you and other hardcore users use their own implementations of big frameworks on top of awesome is the whole reason why some version of this should be merged. And ideally a culture of easily sharable code should be established.

as to the most users and copying, they already do that, they copy one file and change 5 lines in it, and than they add 5 more lines, and more and more, and it becomes bloated crap, all because they started with monolithic huge file.  With this approach they will have to copy and edit only files that they need, and for adding new widgets and new key bindings they don't need to edit any default configs)

actionless:
as far as i remember @Elv13 had some ideas in this direction, let's see if this effort goes along with it or not

lib/default/default/common.lua
actionless: i really don't think my config would be any smaller, as i have almost any line of default config redefined, including default hotkeys, layouts, signals, panels, whatever -- so actually on this question it might be more valuable to get the feedback from people who have their config 50+ % match the default one

psychon:
I didn't look at this at all, but I noticed that this did not work on Travis. All the builds seem to have failed with something like the following.
```
error: /tmp/tmp.MjkR4fBLYN/awesomerc.lua:21: module 'default' not found:No LuaRocks module found for default
```
Well, all but the last one. The one build that runs LDoc complains about:
```
/home/travis/build/awesomeWM/awesome/build/lib/default/default/bindings/client_keys.lua:52: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/bindings/client_mouse.lua:20: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/bindings/global_focus_keys.lua:41: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/bindings/global_general_keys.lua:16: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/bindings/global_layout_keys.lua:44: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/bindings/global_mouse.lua:15: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/bindings/global_tags_keys.lua:58: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/client_rules.lua:55: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/common.lua:43: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/error.lua:13: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/focus.lua:6: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/hotkeys_popup.lua:12: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/keyboardlayout.lua:7: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/launcher.lua:10: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/layoutbox.lua:22: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/layouts.lua:21: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/menubar.lua:14: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/notifications.lua:22: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/panel.lua:52: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/promptbox.lua:29: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/rc.lua:3: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/systray.lua:10: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/taglist.lua:33: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/tags.lua:7: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/tasklist.lua:26: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/textclock.lua:7: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/titlebars.lua:44: no module() call found; no initial doc comment
/home/travis/build/awesomeWM/awesome/build/lib/default/default/wallpaper.lua:17: no module() call found; no initial doc comment
```
The path here could provide a hint on what is going on. `default/default` sounds at least surprising.

Speaking of which: Is `default` a good name for this module? I am not sure. One possible alternative could be something like `awesome_default_config` (but that is waaaay too long).

seniorivn:
@psychon what do you have to say on the idea of modular default config(whatever a name for the lib should be) Is it needed? Is awesome ready? How it should be done?

psychon:
> @psychon what do you have to say on the idea of modular default config(whatever a name for the lib should be) Is it needed?

https://github.com/awesomeWM/awesome/issues/442#issuecomment-139482282

> Is awesome ready?

Uhm, I guess so. Lua's `require` is all that is really needed for this...?

> How it should be done?

Split the default config into multiple files and `require` the parts. :-)

lib/default/default/common.lua
ShayAgros: Hi, I'm rather new to awesome wm, and haven't changed large chunks of the code.

I think that most new comers start by changing the wibar, theme and client key bindings (shortcuts to launch programs) sections. Also,  dividing the config to multiple files can help users to maintain the same pattern when creating their own versions of the changed files. I myself struggle with deciding how to divide the monolithic config file into smaller, more easily editable files.

lib/default/default/bindings/client_keys.lua
```diff
@@ -0,0 +1,51 @@
+local awful = require("awful")
```
ShayAgros: I might be reading it wrong, but it seems like you created a 'default' directory inside lib, and then added another 'default' directory to it.

Can I ask what are the benefits of doing so over getting rid of one of these 'default' folders?

Elv13:
One of the big fear I have with this approach is that users might start forking individual files. If they start doing so, then it will become significantly harder to make changes to the default config (since technically they might have side effects on other sections). If I saw "we don't support that", then these people wont be happy.

With a single file, then it is all or nothing. Either they fork and they carry on their config to the new versions or they use the default. That allows us to keep the API surface on the public API and we don't have to worry about making changes to the default `rc.lua`. With this, there is a whole new layer of complexity.

That doesn't mean the config cannot be provided as a multi-file repository. Maybe making this it's own git repo as part of the official project and leave the single file in the main package a better option? That way users are "forced" to git-clone the modular config into `~/.config/awesome` and it wont be altered by the system AwesomeWM updates? It has some nice advantages too, it is easier to git-pull or cherry pick changes and as long of you didn't modify the affected files, it might just apply/work cleanly.

lib/default/default/bindings/client_keys.lua
seniorivn: that was my idiocy(caused by not checking how it works after copying "default" dir that i checked as a simple module in a regular installation, copying into the repo), already fixed, but haven't actually tested anything, will get back to it when more knowledgeable contributors tell how they see it merged.

seniorivn:
@Elv13 I actually think making it a part of awesome and updating parts of default config that are affected by some change in apis/libs is a better way to deliver new features to an average user.

right now  there are at least two big groups of users:
 1) people who don't update their configs until you present a breaking change. At that point they either invest into upgrading their config or ditch awesome for a less powerful but more stable alternative. And if AwesomeWM project is interested in growing it's user base(which i believe it does, all those people are likely to become contributors at some point) it should make it easier to be on the cutting edge for people who don't keep track of the development and possibly even major updates.

 2) people who use awesome as a very flexible framework to develop their own desktop environment(you and other devs are the best examples). You update your configs either before new version comes out or as soon it comes out. For you default config is just something your setup falls back to in an unlikely event of a failure.

I understand that for the second group it's very hard to think of the first one, but without it the second is unlikely to grow.

Maybe it's one of those issues that need a survey. Can we organize a survey to find out what people actually change in their configs? How people use awesome?

MathiasKandelborg:
This is a definite move in the right direction. The current configuration is not something for newcomers to mess around with.

Having a keybindings.rc, rules.rc, etc. people would have an easier time getting started.

There should also be documentation on each default custom.rc such that what is being done is explained with some reasoning and maybe even alternate ways. This way people won't have to spend hours upon hours to get awesome running with a few changes of their own.

Elv13:
> people who don't update their configs until you present a breaking change. At that point they either invest into upgrading their config or ditch awesome for a less powerful but more stable alternative. And if AwesomeWM project is interested in growing it's user base(which i believe it does, all those people are likely to become contributors at some point) it should make it easier to be on the cutting edge for people who don't keep track of the development and possibly even major updates.
> people who use awesome as a very flexible framework to develop their own desktop environment(you and other devs are the best examples). You update your configs either before new version comes out or as soon it comes out. For you default config is just something your setup falls back to in an unlikely event of a failure.

This is a valid use case, but I went a different way for this. In git-master, there is this line:

```
-- awesome_mode: api-level=4:screen=on
```

at the top. This is to control exactly that. The config will not break ever again. When there is a API breaking change or behavior change, then it must be done in the current+1 API level. Users are free to stay in older ones as long as they want. All backward compatible changes are currently available in older API level. If AwesomeWM API was ever to change so much this would not be possible, then there is some code to fork the entire `lib/` directory and silently load the older copy. However I don't plan to actually use that capability anytime soon. So far I always managed to preserve backward compatibility since 4.0 (but API level 5 does exist and it does change the bahavior somewhat to enable better laptop support, but I never finished/merged some of those feature like rules for screens).

The API level feature exist for the exact use case you described and, as Android has shown, is a much better systems to support evolving APIs. In the case of Android, it created some fragmentation because almost nobody care about the actual Android upgrades because they get support from new apps anyway (due to forward compatibility layers on top of old API levels to bridge the gap transparently).  If we were to merge this change as-is, then it would still make it more difficult to make change. So I am rather against replacing the *default* config with this work. Merging that will be shooting ourselves in the foot as mentioned above. As soon as any user start fragmenting/forking part of the default config but not all of it, then any update will start breaking they config again, which is exactly what we want to avoid happening (ever).

However I 100% support having this part of the larger AwesomeWM project and recommending new users to use that rather than the default config in the documentation (like "my first awesome"/"getting started"/FAQ pages).

MArpogaus:
Hello People,

I just stumbled over this discussion and thought that I could maybe contribute my work in this context.

Over the last couple of months, my [awesome config](https://github.com/MArpogaus/awesome/tree/dev) evolved into something like a extensible modular configuration framework (loosely inspired by what [doom](https://github.com/hlissner/doom-emacs) offers to emacs users).

What began as a hobby project to make my config more reusable over several machines, got quite mature now and offers a flexible and easy extensible configuration with reasonable defaults.

I don't think this should replace the default rc.lua either, but it might become what you had in mind @Elv13.

In any case, I am happy if this is useful for someone else, and of course any contributions are greatly appreciated.

raven2cz:
It is essential to distinguish between two perspectives.

First sight.
People who start with Lua have experience of xfce, i3, xmonad, qtile. These people can't coverage a complex modular structure in the first weeks or months and go away; saying that it's too complicated for them. While this is not true in principle, they only count the number of files they had to enter to understand the whole "default" project. But they leave before they reach the poodle's core.

The second group are people who already know what they want, have experience and know that the modular arrangement is important for them for the aggregation of other components and the possibility of confusion in the form of plugins and wibars. But this group don't go away from awesomewm!

Also keep in mind that there have been tens to hundreds of tutorials on the Internet for many years, as well as git repositories with patches + examples that take into account the structure of the current default rc.lua. In a moment of change, all this will cease to apply and those people will leave again because they will not want to waste time with it.

A beginner will become an awesomewm lover only after a while, and this must be made possible by a simple or "monolithic" path that is close to him, which unfortunately will never be a complex modular structure.

Qtile is starting to be very successful again, mainly due to the stupid configuration and the fact that everywhere they say that python is the main and trivial programming language. So again, I appeal to hasty things and you really need to think a lot about what it will be for beginners as an easy-to-use advertising ad!

sclu1034:
That's a very good point. 

I'm relatively active and constantly lurking in the various channels that offer help for Awesome, and from my experience most of those users either outright say they never had experience with Lua (or programming in general) before, or act like it. By making a complex, modular config the default, we might be able to make stuff like "enable module X, disable module Y" easier (i.e. "to `require` or not to `require`"), but any kind of logic changes will become vastly more complex for beginners to find and change.
In fact, the current version of this PR moves all of those modules to `/usr/share`, so that "to `require` or not to `require`" is all the customization that is left, until the user starts copying over files from `/usr/share` (and you first need to tell them that it is possible).

If we wanted to make things easier for those users, the better approach would instead be to "dumb down" the default `rc.lua` until it is practically a `key=value` file. There would be some module (either in Awesome's `lib/` directory or as part of what you copy into `~/.config/awesome`) that the default `rc.lua` imports. That module would implement a variety of popular options and functionality and would be controlled through a single `config` table.

Example:
For key binds, the current version of `rc.lua` assumes that if you want to
- remove a default bind, delete the `awful.key` call
- add a new bind, create a `awful.key` call and implement the callback

The "dumbed down" version would instead offer a readable, logic-free config table like this:
```lua
config.keys = {
	["Meta+R"] = helpers.keys.show_run_prompt,
	["Ctrl+Meta+R"] = helpers.keys.restart_awesome,
}
```
Where `helpers.keys` is some structure provided by that config module that implements a broad variety of functions that a beginner might be looking for.

A modularized version of the current `rc.lua` makes the most sense for experienced users. They know how things work, so they are fine with editing logic directly and can benefit from both the low-level access of the current ways and the readability and separation in a modularized structure.

----

But I'm not sure if Awesome should go the route of that "dumbed down" config. There already is a decent amount of config file based window managers out there, and Awesome does call itself a "window manager _framework_".
We might be stuck with a good chunk of users that don't want to "graduate" to an experienced user and instead expect us to continuously add new stuff that can be controlled with their `key=value` config.

raven2cz:
> Awesome does call itself a "window manager framework". 

I built large frameworks whole my live. And there is just one important rule if the framework has to survive the evolution; and it is **simplicity**. But to reach the simplicity is most complicated think, because people thinking is complicated.

From my side, good framework has to cover complex structure, but **HAS TO SHOW SIMPLE FACADE** for common users, not for advanced user. Facade is very important for any framework. The facade can be used or not. But it is first entrance for common approach and first looks - demonstration, first explanation of business logic, examples. 

So, from my side, this is not accurate, and it needs much more thinking about to design it.

raven2cz:
I didn't want this thread to be dumb.
Will it continue to find a solution for the default configuration?

actionless:
there are several open or wip PRs mostly by @Elv13 with some effort into making it possible, but as you see yourself from the comments to this PR in the current state of codebase there is no much of painless way to move there instantly

actionless:
if i understood the total discourse right - the main direction is being able to override as most of your awesomewm behavior just from the theme itself, using signals

raven2cz:
I think it is necessary to agree on **common interests** for the default configuration for beginners first.

Maybe create discussion thread for it?

Modularity is important, but a high degree of modularity is inappropriate for a start. Why, see my post here: 

https://github.com/awesomeWM/awesome/pull/3131#issuecomment-891106543

However, the current default configuration is not suitable either.

From my point of view, I see two main problems:

* First, what is a **theme** and its definition? It's definitely **not just a colorscheme**. We will all agree on that. This is wrong in the current default configuration. It is necessary to properly DETACH the GUI components (thus the whole theme) from the **backend configurations**.

For example my config: https://github.com/raven2cz/awesomewm-config

This my repo is NOT proposal, but try to explain where is the main problem for actual default config.

* Second, the current configuration using `connect_signal()` listeners is very good and has greatly aided modularization!
**For beginners, simplicity is needed, especially in the overall view, where what is with the description.**

**NOTE:** Be sure to also add where and how to run autorun scripts, which is very chaotic now.

The last recommendation is to create a "facade module" which hides the still complex components and business logic that is often used in the default system configuration. This is very easy to use via a facade call in the default configuration and thus simplify some complex settings. The default configuration complexity is often blamed in AW evaluation and there is a reason to simplify it.

This facade approach can improve the management of wallpapers, application rules, notifications and notification center facilitation, better organization of layouts (so that they do not have to be configured in multiple places), etc.

**But all of it is still just 2 files only.** One for backend configuration and second about GUI (theme) configuration. In both cases, facade can help to simplify it's content (both files). The design of these files has to ensured that the GUI theme can be easily swap with others themes! It is a requirement.

Elv13:
I am personally fine with the notion of moving the wallpaper, titlebar, notification and wibar listeners to the theme file (for v5.0). I think @actionless was opposed to it at some point, IDK if that still stands. The current theme concept is IMHO not at the right cutoff point. These themes only help with colors and not really the "look". That makes the "look" pretty much impossible to port from config to config. Since both "look" and "usage" part are very coupled, making themes 100% pluggable is impossible which major breaking changes in the API (which wont happen because they cause user exodus). But they could me "more portable" by having the wallpaper/wibox/notification/titlebar in them. I am also for making the theme a directory rather than a single lua file (removing the `theme.lua` from this `rc.lua` line: `beautiful.init(gears.filesystem.get_themes_dir() .. "default/theme.lua")`).

I am not really enthusiastic about adding another layer of abstraction on top the the current `rc.lua` business logic. The reason it has those section is because (beside the keybindings) are the entry point for each subsystems. The goal of AwesomeWM is to bubble-up all of those things into the default config to lesser the learning curve. Even if it makes the config longer, I think in the end it helps. "In theory" and empty `rc.lua` would do nothing at all. AwesomeWM doesn't have any defaults, everything starts from `rc.lua` (in practice, there is a large number of exceptions, but the users don't see them).

Compared to v4.3, the current `rc.lua` (and even more so with the wallpaper and wibar PRs which will be merged soon) removes most of the imperative logic and turns everything into a `connect_signal` listener which contains only declarative nested key/value constructs. Not only are they easier to move to other file (cut, paste and `require`), but they untangles the whole thing.

The "top" of `rc.lua` is still a mess. The menu and default apps still have no nice ways to encapsulate them. They are tied with the wibar in ways that are just plain ugly. I could make a dummy `request::something` that does nothing except being executed. But that's "cheating" and I would rather have a proper API for it. It would also not help the coupling between the default apps and the wibar. A replacement would actually need to implement the xdg-alternative (set `x-www-whatever` in a `~/.local` `$PATH` entry) and xdg-auto-start (with optionally systemd cgroup like Gnome/KDE now do) protocols *correctly*. Then find some way to decouple the menu content from the wibar definition. 

But that would/will take a lot of time. Still, I think moving more and more of these section into `request::` listeners and removing every trace of the imperative logic is the way to go. Some people understand the imperative code better than the declarative one, but I think the majority prefer to start from key/value (and then do other changes) rather than start from imperative logic and make a mess until they know better.

actionless:
> I think @actionless was opposed to it at some point, IDK if that still stands

at the moment, already some movement into that direction is done - so instead of opposing it i rather think of how it could be done in a way which will still practically suit my vision of things :D 

in particular - that would be important to have some mechanism of "singleton signals", where the last subscriber to that signal is replacing the previous one - that would help to prevent some things conflicting between rc.lua and theme. or would allow combining several themes without conflict by just loading them one after another

> I am not really enthusiastic about adding another layer of abstraction on top the the current rc.lua business logic. The reason it has those section is because (beside the keybindings) are the entry point for each subsystems. The goal of AwesomeWM is to bubble-up all of those things into the default config to lesser the learning curve. Even if it makes the config longer, I think in the end it helps. "In theory" an empty rc.lua would do nothing at all. AwesomeWM doesn't have any defaults, everything starts from rc.lua (in practice, there is a large number of exceptions, but the users don't see them).

double this

actionless:
> Maybe create discussion thread for it?

https://github.com/awesomeWM/awesome/issues/442

raven2cz:
I'm happy that you agree with separate theme (GUI) and backend. Question is when 5.0 version will be created, if 4.4 takes 2 years. I want to be realistic not live in dreams. I understand that you want to make it clean way. But I thought change default config for actual implementation state, it means AW API of 4.4+. What I see the actual API is prepared to separate UI to theme. If you look to my config, which is not perfect, but the UI is not in rc.lua now and standard rc.lua is still kept. Add your best next ideas (without api changes) and it is done to release 4.4 ;-)  or some bugfix version 4.4.1.
In my opinion, the default rc.lua config and themes are "business card" for new release 4.4. And this investment will bring great value.

PR #3134, opened 2020-07-26
Fix adjust misspelling

jcrd:


unai-ndz:
The tests should also be ~~adjusted~~ modified or will break, it's a [spread](https://github.com/awesomeWM/awesome/search?q=ajust&type=Code) typo.

Though better wait for `writeaccess` to chime in as this breaks the API. Probably can't be merged until `v5.0`(?).

lib/wibox/layout/ratio.lua
```diff
@@ -293,14 +293,14 @@ end
 --- Update all widgets to match a set of a ratio.
 -- The sum of before, itself and after must be 1 or nothing will be done.
 --
---@DOC_wibox_layout_ratio_ajust_ratio_EXAMPLE@
+--@DOC_wibox_layout_ratio_adjust_ratio_EXAMPLE@
```
psychon, line 296: If you change the "path" here, you also need to rename `tests/examples/wibox/layout/ratio/ajust_ratio.lua`.

psychon:
> Though better wait for writeaccess to chime in as this breaks the API. Probably can't be merged until v5.0(?).

How about having both the old and new function? Something like this:
```lua
function ratio:adjust_ratio(index, before, itself, after)
   [stuff here]
end
ratio.ajust_ratio = ratio.adjust_ratio -- For backwards compatibility TODO v5: Remove this
```

jcrd:
Thanks for the feedback. I didn't realize it was widespread. I fixed all instances of the typo, and followed @psychon's suggestion to maintain API compatibility.

lib/wibox/layout/ratio.lua
```diff
@@ -334,18 +334,20 @@ function ratio:ajust_ratio(index, before, itself, after)
 
     self:emit_signal("widget::layout_changed")
 end
+ratio.ajust_ratio = ratio.adjust_ratio -- For backwards compatibility TODO v5: Remove this
```
Aire-One, line 337:
```suggestion
--- Update all widgets to match a set of a ratio.
--
-- This method is kept for backwards compatibility, please use `:adjust_ratio` instead.
-- @see wibox.layout.ratio.adjust_ratio
-- @deprecated wibox.layout.ratio.ajust_ratio
-- @method ajust_ratio
-- @tparam number index The index of the widget to change
-- @tparam number index The index of the widget to change
-- @tparam number before The sum of the ratio before the widget
-- @tparam number before The sum of the ratio before the widget
-- @tparam number itself The ratio for "widget"
-- @tparam number itself The ratio for "widget"
-- @tparam number after The sum of the ratio after the widget
-- @tparam number after The sum of the ratio after the widget
function ratio:ajust_ratio(...)
    require('gears.debug').deprecate(
        "Use `:adjust_ratio` rather than `:ajust_ratio`",
        { deprecated_in = 5 }
    )
    return self:adjust_ratio(...)
end
```

lib/wibox/layout/ratio.lua
```diff
@@ -334,18 +334,20 @@ function ratio:ajust_ratio(index, before, itself, after)
...
 -- @tparam number before The sum of the ratio before the widget
 -- @tparam number itself The ratio for "widget"
 -- @tparam number after The sum of the ratio after the widget
-function ratio:ajust_widget_ratio(widget, before, itself, after)
+function ratio:adjust_widget_ratio(widget, before, itself, after)
     local index = self:index(widget)
-    self:ajust_ratio(index, before, itself, after)
+    self:adjust_ratio(index, before, itself, after)
 end
+ratio.ajust_widget_ratio = ratio.adjust_widget_ratio -- For backwards compatibility TODO v5: Remove this
```
Aire-One, line 350:
```suggestion
--- Update all widgets to match a set of a ratio.
--
-- This method is kept for backwards compatibility, please use `:adjust_widget_ratio` instead.
-- @see wibox.layout.ratio.adjust_widget_ratio
-- @deprecated wibox.layout.ratio.ajust_widget_ratio
-- @method ajust_widget_ratio
-- @tparam widget widget The widget to adjust
-- @tparam number before The sum of the ratio before the widget
-- @tparam number before The sum of the ratio before the widget
-- @tparam number itself The ratio for "widget"
-- @tparam number itself The ratio for "widget"
-- @tparam number after The sum of the ratio after the widget
-- @tparam number after The sum of the ratio after the widget
function ratio:ajust_widget_ratio(...)
    require('gears.debug').deprecate(
        "Use `:adjust_widget_ratio` rather than `:ajust_widget_ratio`",
        { deprecated_in = 5 }
    )
    return self:adjust_widget_ratio(...)
end
```

actionless:
please squash the commits into one

psychon:
Merging based on the two :+1: (no idea why codecov is sad and prevents mergify from doing its job)

PR #3160, opened 2020-09-03
menubar: correct max label width and height detection

### LINKED ISSUE #3158: Wrong font size in hotkeys_popup

awesome v4.3 (Too long)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

Add to your theme the following:
theme.hotkeys_font                              = "Monospace 12"
theme.hotkeys_description_font         = "Monospace 14"

**Actual result:**
Columns are truncated

image

**Expected result:**

Columns height calculated correctly and pagination if necessary (a larger popup would be nice) as with:

image

I normally rule out tmux keys but for this report I uses rc.lua and the default theme as distributed with no other modification. I also tried to set hotkeys_font and hotkeys_description_font via beautiful.get() with exactly (as expected) the same result. My eyes cannot read the original font size on a 4k monitor it is simply too small...

Otherwise, after playing with Awesome for a couple of days, I have to agree it is truly AWESOME...

Denis

actionless:
initial font width code could work only with fixed with fonts

font height also was ignoring custom font

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -455,7 +455,7 @@ function menubar.show(scr)
     local bg_color = theme.menubar_bg_normal or theme.menu_bg_normal or theme.bg_normal
     local border_width = theme.menubar_border_width or theme.menu_border_width or 0
     local border_color = theme.menubar_border_color or theme.menu_border_color
-    local font = theme.menubar_font or theme.font
+    local font = theme.menubar_font or theme.font or "Monospace 10"
 
     if not instance then
         -- Add to each category the name of its key in all_categories
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -70,12 +70,18 @@ end
 -- @beautiful beautiful.menubar_bg_focus
 -- @param color
 
+--- Menubar font.
+-- @beautiful beautiful.menubar_font
+-- @param[opt=beautiful.font] font
+
 
 -- menubar
 local menubar = { menu_entries = {} }
 menubar.menu_gen = require("menubar.menu_gen")
 menubar.utils = require("menubar.utils")
-local compute_text_width = menubar.utils.compute_text_width
+local compute_text_width = function(text, s)
+    return wibox.widget.textbox.get_markup_geometry(text, s, theme.menubar_font)['width']
+end
 
 -- Options section
 
@@ -243,7 +249,7 @@ local function get_current_page(all_items, query, scr)
     local current_page = {}
     for i, item in ipairs(all_items) do
         item.width = item.width or
-            compute_text_width(item.name, scr) +
+            compute_text_width(label(item), scr) +
             (item.icon and instance.geometry.height or 0) + list_interspace
         if width_sum + item.width > available_space then
             if current_item < i then
@@ -449,6 +455,7 @@ function menubar.show(scr)
     local bg_color = theme.menubar_bg_normal or theme.menu_bg_normal or theme.bg_normal
     local border_width = theme.menubar_border_width or theme.menu_border_width or 0
     local border_color = theme.menubar_border_color or theme.menu_border_color
+    local font = theme.menubar_font or theme.font
 
     if not instance then
         -- Add to each category the name of its key in all_categories
@@ -467,6 +474,7 @@ function menubar.show(scr)
                 fg = fg_color,
                 border_width = border_width,
                 border_color = border_color,
+                font = font,
             },
             widget = common_args.w,
             prompt = awful.widget.prompt(),
@@ -490,7 +498,7 @@ function menubar.show(scr)
     local geometry = menubar.geometry
     instance.geometry = {x = geometry.x or scrgeom.x,
                              y = geometry.y or scrgeom.y,
-                             height = geometry.height or gmath.round(theme.get_font_height() * 1.5),
+                             height = geometry.height or gmath.round(theme.get_font_height(font) * 1.5),
                              width = (geometry.width or scrgeom.width) - border_width * 2}
     instance.wibox:geometry(instance.geometry)
 
--- a/lib/wibox/widget/textbox.lua
+++ b/lib/wibox/widget/textbox.lua
@@ -381,6 +381,7 @@ end
 -- @tparam[opt=nil] integer|screen s The screen on which the textbox would be displayed.
 -- @tparam[opt=beautiful.font] string font The font description as string.
 -- @treturn table Geometry (width, height) hashtable.
+-- @staticfct wibox.widget.textbox.get_markup_geometry
 function textbox.get_markup_geometry(text, s, font)
     font = font or beautiful.font
     local pctx = PangoCairo.font_map_get_default():create_context()
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -19,7 +19,6 @@ local glib = lgi.GLib
 local w_textbox = require("wibox.widget.textbox")
 local gdebug = require("gears.debug")
 local protected_call = require("gears.protected_call")
-local gstring = require("gears.string")
 local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
 
 local utils = {}
@@ -417,14 +416,9 @@ function utils.compute_textbox_width(textbox, s)
     return w
 end
 
---- Compute text width.
--- @tparam str text Text.
--- @tparam number|screen s Screen
--- @treturn int Text width.
--- @staticfct menubar.utils.compute_text_width
-function utils.compute_text_width(text, s)
-    local w, _ = w_textbox(gstring.xml_escape(text)):get_preferred_size(s)
-    return w
+function utils.compute_text_width(text, s, font)
+    gdebug.deprecate("Use 'width = textbox.get_markup_geometry(text, s, font)['width']'.", {deprecated_in=4})
+    return w_textbox.get_markup_geometry(text, s, font)['width']
 end
 
 return utils


actionless:
it seems codecov status got stucked for some reasons

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -481,32 +484,28 @@ function widget.new(args)
...
                 end
                 local rendered_hotkey = markup.font(self.font,
                     modifiers .. (key.key or "") .. " "
                 ) .. markup.font(self.description_font,
                     key.description or ""
                 )
-                if length > max_label_width then
-                    max_label_width = length
-                    max_label_content = rendered_hotkey
+                local label_width, _ = wibox.widget.textbox(rendered_hotkey):get_preferred_size(s)
```
Elv13, line 500: Can you switch this to Pango code rather than make a disposable widget? There is plenty of examples in the `tests/example` directory (git-grep -i pango in there).

actionless:
@Elv13 for some reason i'm getting different result when comparing `wibox.widget.textbox(rendered_hotkey):get_preferred_size(s)` vs `get_label_geometry(rendered_hotkey, s).width`


```lua
local Pango = require("lgi").Pango
local PangoCairo = require("lgi").PangoCairo
local dpi = beautiful.xresources.apply_dpi


local function get_label_geometry(text, s, font)
    local pctx    = PangoCairo.font_map_get_default():create_context()
    local playout = Pango.Layout.new(pctx)
    local pdesc
    if font then
        local font_name = beautiful.get_font(font)
        pdesc   = Pango.FontDescription(font_name)
    else
        pdesc   = Pango.FontDescription()
    end
    playout:set_font_description(pdesc)
    local attr, parsed = Pango.parse_markup(text, -1, 0)
    playout.attributes, playout.text = attr, parsed
    local _, logical = playout:get_pixel_extents()
    logical.width = dpi(logical.width, s)
    return logical
end

```

psychon:
`:get_preferred_size(s)` also does this, where `dpi` is `screen[s].dpi`: https://github.com/awesomeWM/awesome/blob/7a759432d3100ff6870e0b2b427e3352bf17c7cc/lib/wibox/widget/textbox.lua#L27-L34
Your code instead seems to get the width with the default DPI and then DPI-scales that. Due to hinting, this can give different rules.

If this guess is wrong: Hm, dunno. Sorry.

actionless:
thanks, @psychon, it's indeed was the reason, i'll push updated version now

actionless:
also added a spec

actionless:
squashed the commits

Elv13 (approved):

lib/wibox/widget/textbox.lua
```diff
@@ -373,6 +373,26 @@ function textbox.mt.__call(_, ...)
...
+    local playout = Pango.Layout.new(pctx)
+    playout:set_font_description(beautiful.get_font(font))
+    local dpi_scale = beautiful.xresources.get_dpi(s)
+    pctx:set_resolution(dpi_scale)
+    playout:context_changed()
+    local attr, parsed = Pango.parse_markup(text, -1, 0)
+    playout.attributes, playout.text = attr, parsed
+    local _, logical = playout:get_pixel_extents()
+    return logical
+end
```
psychon, line 394: Isn't the above equivalent to the following, already existing functionality?
```lua
local t = wibox.widget.textbox(text)
t:set_font(font)
local w, h = t:get_preferred_size(s)
return { width = w, height = h }
```
Why a new function for this?

actionless: it is, but @Elv13 prefers it like that: https://github.com/awesomeWM/awesome/pull/3160#discussion_r482705037

psychon: Hm, okay... @Elv13 why? This adds new public API for something that is already possible and creating a throw-away widget should be so cheap that it doesn't make much difference that some more tables are created.

keryell: At least there should be more comments in this new function to explain the use case and how it works.

actionless: @keryell what exactly would you add there?

actionless: like what?

actionless: ping @Elv13

actionless: @Elv13 , bump

actionless:
bump

lib/wibox/widget/textbox.lua
Elv13: If psychon prefers not adding a new API, then I backtrack. Let's not add a new API and keep using this hack in 9999 places (but most of them are test code, so they don't count)

actionless: so, @psychon ?

actionless: btw, while fixing some stuff in menubar i also found what it have similar functions already:
https://github.com/awesomeWM/awesome/blob/master/lib/menubar/utils.lua#L413-L428

so mb it would be better to merge this as is, and next in a separate PR to move and deprecate the one from menubar.utils

the important difference, what the new function from this PR is taking font into consideration (while the one in menubar is actually broken, as it will show wrong value if using non-default font or even font size in menubar)

psychon: If you think this is useful, then go ahead and add a new API for this.

psychon: However... would it make sense to implement this new function via my suggestion above? That would make the code a lot shorter.
```lua
local t = wibox.widget.textbox(text)
t:set_font(font)
local w, h = t:get_preferred_size(s)
return { width = w, height = h }
```

actionless: i personally don't mind this implementation, the concern about it was what it creates the whole widget object without a reason

actionless:
@psychon , @Elv13 i've pushed the latest version, according to what agreed in the discussion above

actionless:
BUMP

spec/wibox/widget/textbox_spec.lua
```diff
@@ -113,6 +124,29 @@ describe("wibox.widget.textbox", function()
...
+        end)
+
+        it("can compute text geometry w/hardcoded font", function()
+            local text = "<span font='Monospace 16'><b><i>test</i></b></span>"
+            local s = 1
+            local pango_geometry = textbox.get_markup_geometry(text, s)
+            local actual_textbox_width, actual_textbox_height = textbox(text):get_preferred_size(s)
+            assert.is.equal(pango_geometry.width, actual_textbox_width)
+            assert.is.equal(pango_geometry.height, actual_textbox_height)
+        end)
```
sclu1034, line 146: The control value is computed in the same way that `get_markup_geometry` uses internally. This means that this test doesn't actually test whether the return value of `get_markup_geometry` is correct. It merely ensures that `get_markup_geometry` doesn't alter the result of `get_preferred_size` before returning it, while that result may still be incorrect.
It also means that this test is tied to implementation details of the function being tested, which should not be the case for a unit test.

You should get rid of the "default font" test variant completely and change the "hardcoded font" variant to
- use an actual "hardcoded font", e.g. `Terminus 8` or `Dejavu 10`. `Monospace` isn't really "hardcoded", since that still depends on how the system is configured that runs the test.
- determine what the result of the tested function should be (i.e. run it once with the test's inputs) and provide that as fixed numeric literals in the `assert`. Also, the first parameter is the `expected` one
- change the `assert`s to this schema: `assert.is.same({ width = 10, height = 10 }, pango_geometry)`. This way the test also ensures that no unexpected stuff may appear in the result table.

sclu1034 (changes requested):

spec/wibox/widget/textbox_spec.lua
actionless: the bikeshed discussion for implementing this trivial height computation might continue for 2 more years, as every next reviewer have the opposite opinion to previous one, so i'll first wait for 2 same opinions on this, before changing smth further

Elv13 (approved):

actionless:
squashed cos luacheck was failing on prev commits

actionless:
codecov is being weird -- those lines are for sure covered in `tests/test-menubar.lua`, as it was failing due to typo in one of those lines before

lib/menubar/init.lua
```diff
@@ -449,6 +455,7 @@ function menubar.show(scr)
     local bg_color = theme.menubar_bg_normal or theme.menu_bg_normal or theme.bg_normal
     local border_width = theme.menubar_border_width or theme.menu_border_width or 0
     local border_color = theme.menubar_border_color or theme.menu_border_color
+    local font = theme.menubar_font or theme.font
```
Aire-One, line 458: Do we need to have a default here?
I'm afraid the `theme.get_font_height(font)` call on line 501 would crash if the theme don't have values for neither of these two...

actionless:
```suggestion
    local font = theme.menubar_font or theme.font or "Monospace 10"
```

WDYT?

Aire-One: With a second thought, I guess this is not a good design as we have to use a magic string... But I fail to see a better alternative to fail-proof the font variable... Maybe modules should declare their default theming variables to beautiful... Anyway, this is out of scope here.

Yes, I think this is necessary to fail-proof the function call from line 501.

Aire-One (approved):

lib/menubar/init.lua
actionless: i think in a long run that might be achieved by having some test setup with empty theme (`return {}`)

sclu1034:
That test is skipped during coverage collection (when `debug.gethook() ~= nil`). I assume one of the other jobs failed, where the test does run.

actionless:
updated https://github.com/awesomeWM/awesome/pull/3226 to include these changes

Issue #3162, opened 2020-09-04
awful.titlebar.hide doesn't resize window to fill the space left by the titlebar missing


Pablo1107:
Output of `awesome --version`:

```
awesome 4.3 (Too long)
 • Compiled against Lua 5.2.4 (running with Lua 5.2)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: [string "return require('lgi.version')"]:1: module 'lgi.version' not found:
	no field package.preload['lgi.version']
	no file '/usr/share/lua/5.2/lgi/version.lua'
	no file '/usr/share/lua/5.2/lgi/version/init.lua'
	no file '/usr/lib/lua/5.2/lgi/version.lua'
	no file '/usr/lib/lua/5.2/lgi/version/init.lua'
	no file './lgi/version.lua'
	no file '/usr/lib/lua/5.2/lgi/version.so'
	no file '/usr/lib/lua/5.2/loadall.so'
	no file './lgi/version.so'
	no file '/usr/lib/lua/5.2/lgi.so'
	no file '/usr/lib/lua/5.2/loadall.so'
	no file './lgi.so'
 • Transparency enabled: yes
 • Custom search paths: yes
```

**How to reproduce the issue:**

1. Add this to your `rc.lua`:

```
client.connect_signal("property::maximized", function(c)
  if c.maximized then
      awful.titlebar.hide(c)
  else
      awful.titlebar.show(c)
  end 
end)
```

2. Restart awesome.
3. Try maximize a window.

**Actual result:**

![image](https://user-images.githubusercontent.com/17091659/92194777-dd38e600-ee41-11ea-9990-5782d11a6476.png)

**Expected result:**

When hiding the titlebar, the client should be resized to accommodate for the space left by the titlebar hiding, so that bottom margin is not created.

psychon:
Sorry, but I disagree. Yeah, for your use case it might make sense, but for a floating terminal with size hints enabled (meaning: the window height must be a multiple of the terminals font size so that it can display a whole number of rows), it might not even be possible to do what you want this to do. And since the terminal is floating, it also does not make sense, I think.

Can't you just resize the client after hiding its titlebar? Something like `c:geometry{c.screen.workarea}` after `awful.titlebar.hide()` should do the trick.

Aire-One:
[...]

PR #3163, opened 2020-09-05
Add optional client resize on titlebar show/hide.

Aire-One:
Following the discussions at #3162, here is my attempt at adding a feature to resize a client when hiding/showing its titlebars.

While the `titlebar.hide` changes were pretty obvious, I had more difficulties to manage the `titlebar.show` function. It leads me to add the `resize_client` parameter to the titlebar constructor, which is for the better I think, I would probably have done it anyway.

The `load_titlebars` utility gave me some headache, tho... From my understanding, its role is to be sure the `request::titlebar` signal is called once (and only once) before actually building titlebars. It gave me a corner case when the user creates clients with the `titlebar_enabled = false` rule and call the `titlebar.toggle` function. The solution was to also have the `resize_client` parameter here, so it can be passed to the underneath `titlebar.hide` call.

Anyway, it was a little more tricky than I expected but it's working as expected now :)

(ping @Pablo1107)


Pablo1107:
I build my awesome overlay using your feature branch and when I do:

```
client.connect_signal("property::maximized", function(c)
  if c.maximized then
      awful.titlebar.hide(c)
  else
      awful.titlebar.show(c)
  end 
end)
```

It still makes the same effect.

Aire-One:
Yes, you have to use theses functions like this actually:

```lua
awful.titlebar.hide(c, 'top', true )
awful.titlebar.show(c, 'top', true )
```

BTW, from the report Travis did, you should have an error with your current code:

```
/home/travis/build/awesomeWM/awesome/tests/test-titlebar.lua: Error: running function for step 2/11 (@2): lib/awful/titlebar.lua:612: attempt to index a nil value (local 'args')
```

I wrongly assumed args would always be a table but it can actually but `nil` if user don't give any argument. It's easy to fix but I got lazy...

Pablo1107:
Oh, you're absolutely right, I forgot we were adding this as an optional setting.

> It's easy to fix but I got lazy...

Go doesn't get lazy when coding? lol

Another thing I think is, how can we document this? I think the API docs are good overall but this kind of edge cases are not explained. Is there a repo I can make a pull request when this change is merged?

psychon:
> I think the API docs are good overall but this kind of edge cases are not explained. Is there a repo I can make a pull request when this change is merged?

The API docs are automatically build from the source code via ldoc. So, the repo to do PRs against for the API docs is this repo.

Pablo1107:
But what about the ones that are manually written. Like the declarative
layout and others.

psychon:
@Pablo1107 https://github.com/awesomeWM/awesome/blob/master/docs/03-declarative-layout.md and https://github.com/awesomeWM/awesome/blob/7a759432d3100ff6870e0b2b427e3352bf17c7cc/docs/config.ld#L36

Aire-One:
I couldn't make busted works again since my arch install now defaults to lua5.4 and awesome still needs lua5.3... Busted complains about wrong lua-lgi dependencies as he's trying to access the lua5.4 version (which is excluded from my machine because it conflicts with lua53-lgi)...

Anyway, I fixed the error based on the Travis report. Let's hope I don't have to fix more issues this ways as I can't run tests locally for now...

I also added a missing condition to check if the client is actually floating when resizing client from the titlebar constructor.

I'll squash my commits once Travis is ok.

Aire-One:
> Another thing I think is, how can we document this? I think the API docs are good overall but this kind of edge cases are not explained. Is there a repo I can make a pull request when this change is merged?

I already added the ldoc stuff so we have this new parameter described in the API references. You can consult this on your local version of the documentation after a build of awesome (if you have ldoc installed on your system) at _file:///PATH-TO-AWESOME/build/doc/popups_and_bars/awful.titlebar.html_ (replace PATH-TO-AWESOME with the actual path).

actionless:
i think adding `args` table instead of `resize_client` bool would be more inline with the latest API additions' style

also code coverage should be added in tests, there are already exist few similar tests related to `xterm` geometry, so i think it should be relatively easy to base the new tests on them

Pablo1107:
> i think adding `args` table instead of `resize_client` bool for public functions would be more inline with the latest API additions' style

Yeah, and it's more clean to call it like that, even so when you add too many args. But one thing is that in the docs, it's often missing the arguments you can provide to each function. Some examples:

image
image

actionless:
@Pablo1107 if you spot any doc infidelities please feel free to PR them, ideally `args` doc should like somehow like that:

2020-09-06--1599427881_1241x882_scrot

Pablo1107:
@actionless Perfect. I think I can help with that. I'll try to see my way around the code to add those when I see one missing.

Pablo1107:
> Yes, you have to use theses functions like this actually:
> 
> ```lua
> awful.titlebar.hide(c, 'top', true)
> awful.titlebar.show(c, 'top', true)
> ```
> 
> BTW, from the report Travis did, you should have an error with your current code:
> 
> ```
> /home/travis/build/awesomeWM/awesome/tests/test-titlebar.lua: Error: running function for step 2/11 (@2): lib/awful/titlebar.lua:612: attempt to index a nil value (local 'args')
> ```
> 
> I wrongly assumed args would always be a table but it can actually but `nil` if user don't give any argument. It's easy to fix but I got lazy...

Sorry for the late reply, I tried running now the `connect_signal` like this:

```
client.connect_signal("property::maximized", function(c)
    if c.maximized then
        awful.titlebar.hide(c, 'top', true)
    else
        awful.titlebar.show(c, 'top', true)
    end 
end)
```

And something weird happens. My terminal windows resize no problem with this, but gtk applications does not (Firefox and Element).

actionless:
@Pablo1107  might be related to `.maximized_vertical`/`_horizontal` vs `.maximized` issue?

Elv13:
>  (cc @Elv13)

(semi off topic thought)

The feature itself makes perfect sense.

The big question is "should be burn the titlebar API to the ground and make a new one?". Recently, we had this /r/unixporn entry with the super nice AA titlebar module. There is also https://github.com/awesomeWM/awesome/issues/650 which needs to be implemented at some point or another (along with `awful.wallpaper` and my dynamic client layout system) to "complete" AwesomeWM v5.

My grand design dream is to unify everything  (systray, screens, wiboxes, titlebars/serverside_decorations, wallpapers, client layouts, etc) under the same `gears.hierarchy` based backend and declarative syntax system. Then bridge the gap and re-unify the declarative and imperative APIs into 1 (currently blocked on https://github.com/awesomeWM/awesome/issues/2181). This would make AwesomeWM5 feel very unified and integrated.

I don't know if we should aim for that or admit this is too much work and iterate the current titlebar API. The volume of work I managed to do this year is... disappointing...

The current titlebar API is one of the last holdout of the 3.4->3.5 major overhaul of AwesomeWM design. It still has the very 3.4-ish procedural style API rather than the 4.0 OOP+Declarative API style. It's kind of stuck in limbo between the AwesomeWM 3.x and the present era. It's a bit inflexible for how people want to use it (2bwm multi border, anti aliasing, client shading, Fluxbox style resizable borders, toggle for fullscreen/maximized, etc). Everything *can* be done if you spend enough time writing a couple layers of abstraction on top of the current API like I did in Blind or @mut-ex did in his module. The current API will fight you and make adding new features very un-elegant.

actionless:
i've actually meant what if client have `maximixed_vertical+horizontal` instead of just `maximized` this code wouldn't work:

```
client.connect_signal("property::maximized", function(c)
    if c.maximized then
```

i've mentioned you because i thought you were making some changes recently unifying maximized states of a client

Elv13:
> i've mentioned you because i thought you were making some changes recently unifying maximized states of a client

Quite the opposite. The vert/hor maximization state was splitted and isolated from the total maximization state. The effort was to make them drift away, not unite. Having those 2 things too close to each other caused side effects and bugs.

Pablo1107:
> The big question is "should be burn the titlebar API to the ground and make a new one?". Recently, we had this /r/unixporn entry with the super nice AA titlebar module. There is also #650 which needs to be implemented at some point or another (along with `awful.wallpaper` and my dynamic client layout system) to "complete" AwesomeWM v5.

Well, actually the issue was originated because of mut-ex/awesome-wm-nice#10. What a nice titlebar.

Pablo1107:
> i've actually meant what if client have `maximixed_vertical+horizontal` instead of just `maximized` this code wouldn't work:
> 
> ```
> client.connect_signal("property::maximized", function(c)
>     if c.maximized then
> ```
> 
> i've mentioned you because i thought you were making some changes recently unifying maximized states of a client

But it actually ran the code afterwards, just sometimes with the gap, sometimes not. But it was because it was running twice because of another thing that I fix in my config. Probably some race condition or something.

Aire-One:
Hi all, sorry for long inactivity on this PR.

> The big question is "should be burn the titlebar API to the ground and make a new one?".

Well, this is a possibility but I think it would take some time, so I guess I should finish this feature and see where we go next.

With my last commits, I changed the statics functions to use the "args pattern" instead of many parameters as requested by @actionless.

I'd like to add test for this (also requested by @actionless) but I'm not sure if I should do that using busted unit tests with a new spec definition for awful.titlebar or within examples in tests/test-titlebar.lua... Actually, I'm not even sure what the coverage is based on...

actionless:
i think functional tests, the ones in `./tests` dir would be the best place

there are already existing tests for clients' geometry, so you could just ensure what geometry with and without titlebar is as expected

Aire-One:
I'm slowly working on this... I can see on the Travis report I broke the `tests/test-titlebar.lua` with my last commit. I'm unable to check if I fixed it with my local changes as I can't run `make check-integration`... It gives me error with my lua stack...

```
== Running /home/aireone/documents/prog/awesome/tests/test-awful-layout.lua ==
2020-12-28 15:33:24 W: awesome: a_glib_poll:470: Last main loop iteration took 0.115379 seconds! Increasing limit for this warning to that value.
2020-12-28 15:33:25 W: awesome: a_glib_poll:470: Last main loop iteration took 0.119123 seconds! Increasing limit for this warning to that value.
== Running /home/aireone/documents/prog/awesome/tests/test-awful-placement.lua ==
_client lua: (command line):2: module lgi not found:
_client         No LuaRocks module found for lgi
_client         no field package.preload[lgi]
_client         no file /usr/share/lua/5.4/lgi.lua
_client         no file /usr/share/lua/5.4/lgi/init.lua
_client         no file /usr/lib/lua/5.4/lgi.lua
_client         no file /usr/lib/lua/5.4/lgi/init.lua
_client         no file ./lgi.lua
_client         no file ./lgi/init.lua
_client         no file /dev/null/.luarocks/share/lua/5.4/lgi.lua
_client         no file /dev/null/.luarocks/share/lua/5.4/lgi/init.lua
_client         no file /usr/lib/lua/5.4/lgi.so
_client         no file /usr/lib/lua/5.4/loadall.so
_client         no file ./lgi.so
_client         no file /dev/null/.luarocks/lib/lua/5.4/lgi.so
_client stack traceback:
_client         [C]: in function require
_client         (command line):2: in main chunk
_client         [C]: in ?
Error: timeout waiting for signal in step 1/50 (@20) (...ne/documents/prog/awesome/tests/test-awful-placement.lua:54).
```

Based on https://github.com/archlinux/svntogit-community/blob/packages/awesome/trunk/PKGBUILD and https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=awesome-git, I already did some lua version changes to compile awesome correctly with lua5.3. I use Arch and I have lua5.3 and lua53-lgi packages installed on my system. I can correlty start awesome as main WM and from a Xephyr instance without this lua version issue.

I already checked awesome uses lua5.3 with `awesome --version` and other make checks seems to work as expected... `make check-unit` corretly uses busted I installed with luarocks for lua 5.3 for example. So I guess the issue is only from how `make check-integration` works.

From my understanding, `make check-integration` uses `awesome-client`. Can the interpreter used by awesome with awesome-client be a different lua binary (like the default one which is lua5.4 under arch) while awesome uses 5.3?

psychon:
That error comes from `/usr/bin/lua`. The test spawns a new Lua process (with `awful.spawn`) which is supposed to open a new window with LGI & GTK. Basically: `lua -e 'require("lgi")'` is failing for you.

psychon:
If you want the test to use Lua 5.3:
```diff
diff --git a/tests/_client.lua b/tests/_client.lua
index 31b9c6188..04fd67d11 100644
--- a/tests/_client.lua
+++ b/tests/_client.lua
@@ -106,7 +106,7 @@ local pipe
 local function init()
     if initialized then return end
     initialized = true
-    local cmd = { "lua", "-e", test_client_source }
+    local cmd = { "lua5.3", "-e", test_client_source }
     local _, _, stdin, stdout, stderr = awesome.spawn(cmd, false, true, true, true)
     pipe = Gio.UnixOutputStream.new(stdin, true)
     stdout = Gio.UnixInputStream.new(stdout, true)
```

Aire-One:
Thanks @psychon! It's exactly what I needed. I fixed a few more instances of this in other tests to make all integration tests work, now it's working like a charm. :fireworks:

Aire-One:
Ok so, 3c60f58 was a very bad idea as the drawable widget expected here should be something awful.layout related and these don't know about their geometry. This means `ret.widget.height` and `ret.widget.width` are `nil` hence, we always use 0. It works great to fix the use case where there is no titlebar defined, so their size should be 0. But it break everything when we actually have a titlebar defined and the test (45e681d) shows that...

I can think of 3 possible solutions here:

- We expect the user to not try to play with titlebars on position they don't have defined, it means I should remove the "position loop" from my test to only test the top titlebar (and drop 3c60f58) ;
- We improve 3c60f58 to find a better way to determine if the user have defined a titlebar for the given position ;
- Force somehow the titlebar drawable to be the size computed at the function beginning even if it's empty, would be a beaking change (if possible but I don't think we can force a size to an empty drawable).

I'm not sure which solution is better, and I don't even know how to implement the two last...

psychon:
I'm not quite sure what issue you are trying to fix. but not only `awful.titlebar` uses "font height * 1.5" as the default size. `awful.wibar` also uses that as its default height. So... perhaps I have to read up again on what the issue actually is, but that seems like a sensible default...?

If you want to determine whether e.g. the left titlebar has any widget attached to it or is just empty, shouldn't it be possible to check whether its widget is `nil`? @Elv13 added magic somewhere to go from a `drawable` to a `wibox`, so it should also be possible to go from a `drawable` to an `awful.titlebar` and then one can somehow dig into that and check whether it has a widget (= whether `wibox.drawable.get_widget()` return `nil`).

Aire-One:
> I'm not quite sure what issue you are trying to fix.

Sorry, I wasn't so descriptive about the issue itself. 

The default awesomerc.lua only define a top titlebar in the "request::titlebar" signal. User can still call awful.titlebar.show/hide/toggle with another position than "top" as parameter. It's why I used a for loop to try all 4 sides in my test. 

In the awful.titlebar.new function, I added a block of code to resize the client depending on the resize_client parameter. The issue is that this code is broken when the user calls this function for a non-defined titlebar position, it will compute new client geometries (with the default size of font*1.5) while the titlebar would actually not take space as it's empty (and will not be drawn).

It's what I tried to fix with 3c60f58. 

> If you want to determine whether e.g. the left titlebar has any widget attached to it or is just empty, shouldn't it be possible to check whether its widget is `nil`? @Elv13 added magic somewhere to go from a `drawable` to a `wibox`, so it should also be possible to go from a `drawable` to an `awful.titlebar` and then one can somehow dig into that and check whether it has a widget (= whether `wibox.drawable.get_widget()` return `nil`).

I didn't try to check the `wibox.drawable.get_widget()`. It should do the trick.  I'll check that later.

Aire-One:
I took an hour or two today to finish this PR as it's starting to rot in my todo list.

I found some interesting behaviors of the constructor method: the `position` variable is always set to 'top', whatever is the actual position of the titlebar you are playing with.

Also, I had some head-ash to make the test working as I was running my manual tests with Xephyr using an XTerm window. It has to be the size of a multiple of a character cell, and guess what, the titlebar is smaller than the height of a character, so one row was removed from the window and I had the idea to increase the client size by the removed titlebar size. It obviously wasn't a good idea...

Anyway, tests now are working, and I guess the feature itself behaves as I should.

As an end-user the result is very depressing, tho... Here is a quick video to show how it behaves with 3 different clients: A basic GTK app (tiny_client from the test), XTerm and Spotify. https://streamable.com/53cm4c

At this point, I don't even care if we drop the feature, even it's technically working. I don't see anyone using something that sketchy...

PR #3164, opened 2020-09-06
Fixed showing hotkeys_popup on the focused screen

### LINKED ISSUE #3118: hotkeys_popup shows up on the other screen

 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no

Just type Mod4-s
on vanilla awesome master https://github.com/awesomeWM/awesome/commit/e7113d7191e37c5b0faedccff6caf82d8e0bd77c configuration.

**Actual result:**

I have 2 displays connected to my Ubuntu 20.04 laptop and I work mainly on my big 34" display.
When I want to see the current key-binding it actually shows up on the other display on my laptop, far away from my eyes.

**Expected result:**

I would like to have it just in front of me, displayed on screen where my mouse is whet I type Mod4-s

Actually I understand that having the help on another screen is also useful in certain cases, so it would be nice to have this configurable and mapped on Mod4-Ctrl-s or whatever.
I do not have the big picture of `awful.hotkeys_popup`, so I do not know if it is possible or not already (without changing the source of `awful.hotkeys_popup` :-) ).
Perhaps having along `awful.hotkeys_popup.show_help` a `awful.hotkeys_popup.show_help_opt(some_options)` that would return a callable doing the right thing. Another trivial use case for such an optional API would be to change the font size of the help without changing the source of hotkeys_popup.

kdrobnyh:
I faced the following issue: when hotkeys_popup is called, it's shown on the same screen, regardless of the focus. Someone forgot to add x and y arguments when changed hotkeys_popup from wibox to popup.
Fix #3118

### DIFF
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -591,6 +591,7 @@ function widget.new(args)
             placement = place_func,
             minimum_width = wibox_width,
             minimum_height = wibox_height,
+            screen = s,
         }
 
         local widget_obj = {


actionless:
isn't it enough just adding `screen` there?

kdrobnyh:
Good point, it's enough!

actionless (approved):
thanks a lot!

SethBarberee (approved):
Hah, so I forgot that. Nice catch!

keryell:
Thanks for fixing this! I was not able to find the bug myself since I am not fluent in Lua and all the frameworks used here.

PR #3172, opened 2020-09-14
modeline: Check the first XDG and fallback paths for modelines.

### LINKED ISSUE #3166: modeline being ignored

 • Compiled against Lua 5.1.4 (running with LuaJIT 2.0.5)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no

Change (or add) the modeline on top of rc.lua like:
```lua
-- awesome_mode: api-level=6:screen=on
```

**Actual result:**

No changes;
`awesome.api_level` returns `4`.
`screen=off` does not seams to have an effect either.

**Expected result:**

`awesome.api_level` returns `6`.

Starting awesome like: `awesome --api-level 6 --force` works as expected.

Elv13:
This is not perfect. If the first config is invalid, then it wont use
the correct modeline. However there is no way to know if the config
is valid before attempting to execute it, so it's the best we can do.

Fix #3166

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -645,7 +645,7 @@ main(int argc, char **argv)
 
     /* Parse `rc.lua` to see if it has an AwesomeWM modeline */
     if (!(default_init_flags & INIT_FLAG_FORCE_CMD_ARGS))
-        options_init_config(awesome_argv[0], confpath, &default_init_flags, &searchpath);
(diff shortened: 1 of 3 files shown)


actionless (approved):

psychon (approved):
Thanks for looking into this!

PR #3173, opened 2020-09-14
screen: Correctly detect tiled_clients.

### LINKED ISSUE #3169: screen.tiled_clients does not consider c.maximized
```
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

1. Have 2 clients opens on the focused tag
1. Make one client maximized via `c.maximized = true`
1. Make a notification to print `awful.screen.focused().focused.tiled_clients`

**Actual result:**

Output is `2`

**Expected result:**

Output is `1`

**Comments**
Reviewing function `function screen.object.get_tiled_clients` in `screen.lua` it only accounts for `c.floating`, `c.fullscreen`, `c.maximized_vertical` and `c.maximized_horizontal` but not for `c.maximized`. Is there a specific reason for it?

I have noticed it when reviewing my rules for showing the title bar for clients and now I have set my maximize function to not only toggle `c.maximized` but also to set `c.maximized_horizontal` and `c.maximized_vertical` to be set to new `c.maximized` value which is a bit ugly but still functional.

Just trying to figure out if that is an oversight or intentional for some reason.

Elv13:
At some point we added an unified `maximized` property to the client class. Originally it was just setting both horizontal and vertical maximization. Later on, this was rewritten to make the state change more atomic and reversible.

Soon after, we added  `immobilized_horizontal` and `immobilized_vertical` to make most `if` simpler. `screen.tiled_clients` was missed in that refactoring.

Fix #3169

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -503,9 +503,8 @@ function screen.object.get_tiled_clients(s, stacked)
     -- Remove floating clients
     for _, c in pairs(clients) do
         if not c.floating
-            and not c.fullscreen
-            and not c.maximized_vertical
-            and not c.maximized_horizontal then
+            and not c.immobilized_horizontal
+            and not c.immobilized_vertical then
             table.insert(tclients, c)
         end
     end


lib/awful/screen.lua
```diff
@@ -503,9 +503,8 @@ function screen.object.get_tiled_clients(s, stacked)
     -- Remove floating clients
     for _, c in pairs(clients) do
         if not c.floating
-            and not c.fullscreen
```
actionless, line 506: why fullscreen was removed?
or `immobilized_*` is set when fullscreen is on?

psychon: Yup: https://github.com/awesomeWM/awesome/blob/1518b0f4c0e533b9cf076f0b1f418c99e4d20747/lib/awful/client.lua#L802-L803

psychon (approved):
Thanks for submitting this and thanks for remembering this `immobilized` properties

actionless (approved):

PR #3174, opened 2020-09-14
Fix clearing the tag clients.

Elv13:
This fixes the bug and add an official API to hide the `t:clients({})` legacy style API.
Screenshot_20200914_010335

### DIFF
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -349,6 +349,42 @@ function tag.find_fallback(screen, invalids)
     end
 end
 
+--- When all clients are removed from the tag.
+-- @signal cleared
+-- @see clear
+
+--- Remove all tagged clients.
+-- @method clear
+-- @tparam table args The arguments.
+-- @tparam tag args.fallback_tag A fallback tag.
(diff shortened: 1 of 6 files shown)


actionless (approved):

psychon (approved):
Wow, the fix can be so simple. I was wondering about how to fix this while still using `foreach`.

Also thanks for improving the docs all the time.

Elv13:
> Wow, the fix can be so simple. I was wondering about how to fix this while still using foreach.

Yeah, I did evaluate a bunch of avenues to fix this. At some point I played with adding a `FOREACH_POP()` macro. The main problem is to cleanly get the loop iterator variable. Either you are forced to add a `prop` argument to the `pop` macro or you have to move to preprocessor counter and storing the type as a string into each array to add introspection. That would have balooned complexity to save 1.5 line.  Another problem with that was that it would "leak complexity" into untag.

The second one I seriously evaluated was to convert the array into a linked list. The rational was that the array can have ~10 elements and has a lot of insert/remove ops. However there is also a lot of "read all elements sequentially and push them into the lua stack" operations. So it would have been clean-ish, but probable a tiny bit slower due to the cache faults when switching tags. It was also a lot more work, so I got scared.

PR #3175, opened 2020-09-14
notifications: Make `:reset_timer()` more intuitive.

### LINKED ISSUE #3147: naughty.notification.timeout doesn't work
I hope that this question question will be better than before.

 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no

This is (imporant part of) the code:
```lua
beautiful.notification_bg = beautiful.bg_normal
beautiful.notification_fg = "#5B8234"
beautiful.notification_border_color = "#93a1a1"

naughty.notification.timeout = 10
naughty.notification.font = "Liberation Sans Bold 13"

(...)

naughty.connect_signal("request::display", function(notification)

	-- Only if there's an icon: Add the icon-widget
	if notification.icon == nil then
		naughty.layout.box {
			notification = notification,

			border_width = dpi(5),
			screen = 1,

			widget_template = template_without_icon
		}

	else
		naughty.layout.box {
			notification = notification,

			border_width = dpi(5),
			screen = 1,

			widget_template = default_template
		}
	end
end)
```

**Actual result:**

The notification always disappears after about 4,8 seconds whatever I'm using in `naughty.notification.timeout`.
Using `notification:reset_timeout(x)` doesn't help either:
```lua
naughty.connect_signal("request::display", function(notification)

	notification:reset_timeout(10)

	-- Only if there's an icon: Add the icon-widget
	if notification.icon == nil then
		naughty.layout.box {
			notification = notification,

			border_width = dpi(5),
			screen = 1,

			widget_template = template_without_icon
		}

	else
		naughty.layout.box {
			notification = notification,

			border_width = dpi(5),
			screen = 1,

			widget_template = default_template
		}
	end
end)
```

**Expected result:**
The notification should disappear after x seconds as described in the documents.

Elv13:
Previously it would only work if there already was a timeout to avoid some infinite loops. Now it just delay the new timer to the next loop.

Fix #3147

### DIFF
--- a/lib/naughty/notification.lua
+++ b/lib/naughty/notification.lua
@@ -503,6 +503,13 @@ function notification:reset_timeout(new_timeout)
     -- timer before the constructor ends.
     if new_timeout and self.timer then
         self.timeout = new_timeout
+    elseif new_timeout and new_timeout ~= self._private.timeout then
+        local previous_timer = self._private.timeout
+        timer.delayed_call(function()
+            if self._private.timeout == previous_timer then
+                self.timeout = new_timeout
+            end
+        end)
     end
 
     if self.timer and not self.timer.started then
--- a/tests/test-naughty-legacy.lua
+++ b/tests/test-naughty-legacy.lua
@@ -1076,7 +1076,14 @@ table.insert(steps, function()
     assert(naughty.get_by_id(42) ~= n)
 
     -- The timeout
+    local real, called = n.reset_timeout, false
+    n.reset_timeout = function(...)
+        called = true
+        return real(...)
+    end
     naughty.reset_timeout(n, 1337)
+    assert(called)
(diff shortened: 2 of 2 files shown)


actionless (approved):

psychon (commented):
What if someone does `n.timeout = 42` after calling `reset_timeout`? Do they just deserve what they get?

Anyway, :+1: from me.

Elv13:
> What if someone does n.timeout = 42 after calling reset_timeout? Do they just deserve what they get?

Good point. Fixed. I also added a "test" that doesn't really test the right thing, but at least ensure the code gets called.

actionless (approved):

PR #3176, opened 2020-09-14
naughty: Improve backward compatibility to protecting the constants.

### LINKED ISSUE #3145: Notifications are stuck in my config
```
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

1. Clone [my configs](https://github.com/TornaxO7/my_configs/tree/master/awesome) (if you want, it should work out of the box)
2. Install spotify, ~~deepin-screen-recorder~~

**Actual result:**
- Spotify is kinda laggy. It doesn't react on events like switching song:
GIF

- ~~deepin-screen-recorder shortcut doesn't work~~

- notify-send is stuck
GIF

**Expected result:**

- Spotify && notify-send
should work as usual.

- deepin-screen-recorder shortcut:
I changed the shortcut from 
```lua
    awful.key({}, "#107", function() 
		awful.spawn.with_shell("deepin-screen-recorder -s ~/Screenshot.png")
	end,
```
to
```lua
    awful.key({}, "#107", function() 
		awful.spawn("deepin-screen-recorder -s ~/Screenshot.png")
	end,
```
And now it's working. I can use it multiple times now without any problems.

Elv13:
If the user copy/pasted `naughty.config.*` into their config rather than set values 1 by 1, we could no longer add new values since they would get removed. To prevent more users being affected by this, we now silently ignore the new table while still setting all the values.

It's not perfect. If the user do

```
local tab = {}
naughty.config.defaults = tab

tab.foo = bar
```

Then that currently did something and it wont work anymore. However if they do

```
naughty.config.defaults = {
    foo = bar
}
```

Then that will help.

Given that the alternative is a fully broken notification subsystem when upgrading (from any version to any version really, this isn't related to my work for 4.4), this is an improvement. 

Fix #3145

### DIFF
--- a/lib/naughty/constants.lua
+++ b/lib/naughty/constants.lua
@@ -9,9 +9,10 @@
 -- @copyright 2017 Emmanuel Lepage Vallee
 ----------------------------------------------------------------------------
 local beautiful = require("beautiful")
+local gtable = require("gears.table")
 local dpi = beautiful.xresources.apply_dpi
 
-local ret = {}
+local ret, no_clear = {}, {}
 
 ret.config = {
     padding         = dpi(4),
@@ -21,7 +22,7 @@ ret.config = {
     notify_callback = nil,
 }
 
-ret.config.presets = {
+no_clear.presets = {
     low = {
         timeout = 5
     },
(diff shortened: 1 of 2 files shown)


actionless (approved):

psychon (approved):
Whatever...

PR #3184, opened 2020-09-20
naughty: Correctly notify the client when a notification is dismissed.

### LINKED ISSUE #3182: Clicking notification does not urgent/raise client
This issue has been reported on reddit and seems to be an regression from v4.3. If this is not a regression please just close this one. I tried searching the existing issues but wasn't able to find something related.

[Reddit Post](https://www.reddit.com/r/awesomewm/comments/itsb5o/clicking_to_notifications_doesnt_raise_app/)

v4.3 according to the reddit discussion.

Click on a notification sent by a client (for example Spotify)

**Actual result:**

The notification disappears, that's all.

**Expected result:**

Clicking a notification in v4.3 seems to have raised the sending client and set it to urgent. This does not work with the latest git version anymore.

Note that I didn't have time to verify the behavior in v4.3 yet but the reddit user reports that it worked that way. See link above.

**Workaround**

Using the following signal handler seems to at least mimic the old behavior:

```lua
local naughty = require('naughty')
local cst = require("naughty.constants");

naughty.connect_signal('destroyed', function(n, reason)
    if not n.clients then return end
    if reason == cst.notification_closed_reason.dismissed_by_user then
        for _, cli in ipairs(n.clients) do
            cli.urgent = true
            cli:emit_signal("request::activate", {raise=true})
        end
    end
end)
```

Elv13:
Fix #3182

### DIFF
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -341,6 +341,8 @@ function notif_methods.Notify(sender, object_path, interface, method, parameters
             args._unique_sender = sender
 
             notification = nnotif(args)
+
+            notification:connect_signal("destroyed", function(_, r) args.destroy(r) end)
         end
 
         invocation:return_value(GLib.Variant("(u)", { notification.id }))


psychon (approved):
Feels weird to approve my own commit that I never committed...
Also, I was almost sure that my approach was not the way to go, so... this time I thought a bit more about it.

Is the whole "`args.destroy` dance" needed? Wouldn't `notification:connect_signal("destroyed", function(notif, reason) sendNotificationClosed(notif.id, reason) end)` be better? Afterwards, the `args.destroy =`-line and `if k == "destroy" then k = "destroy_cb" end`-line can be removed.

:+1: anyway since this fixes the issue.

seqizz:
Thanks for looking into this, but I couldn't approve this fixes the issue.

I've tested this patch [1] / upstream master / release (4.3) versions with Slack app.

- This patch and upstream master: Nothing happens, only notification disappears.
- Release version: Notification disappears. Also client is not only raised and set to urgent (like the function provided [here](https://github.com/awesomeWM/awesome/issues/3182#issue-704210456) accomplishes), it also switches to the user who sent me the message inside the application. So some real nice integration is going on :thinking: 

If you suspect any commit which might be causing this issue, I can test.

[1]
```
nix-repl> pkgs.myAwesome.src.meta
{ homepage = "https://github.com/Elv13/awesome-1"; }

nix-repl> pkgs.myAwesome.src.rev
"59ff7c2d3413c5212bac05ea3ca7122f790cfcc7"
```

blueyed:
@seqizz 
> If you suspect any commit which might be causing this issue, I can test.

Depending on your motivation / build setup (duration etc) of awesomeWM etc you might want to git-bisect it yourself.

seqizz:
> Depending on your motivation / build setup (duration etc) of awesomeWM etc you might want to git-bisect it yourself.

Sadly I am not a developer to understand what might cause this (to narrow the scope enough) and my setup for this is manual.

Elv13:
> Depending on your motivation / build setup (duration etc) of awesomeWM etc you might want to git-bisect it yourself.

The entire module was rewritten nearly from scratch by @psychon (dbus) and me (`naughty`) to bring feature parity with dunst. Why/when it broke is pretty well known. What exactly is preventing all the signals to get back to the notification caller needs investigation.

seqizz:
Did something change to fix the issue stated on [previous comment](https://github.com/awesomeWM/awesome/pull/3184#issuecomment-699543057)? Anyway I'll test again to check when I have time. Thanks.

PR #3185, opened 2020-09-26
Hotkeys popup: better split key labels for multiple keys per one action

actionless:


actionless:
BUMP

Elv13:
Can you provide a description of the change and some screenshots? Also, can you edit this PR and try to keep the commits first line under 72 chars? It is the soft limit where most tools auto-add the "...".

I looked at this last march, but at the time, I wasn't so sure what you were attempting here.

actionless:
1) initial font width code could work only with fixed with fonts
2) font height detection was working only if hotkey font >= description font
3) old one was crashing on hotkeys like '>'

spec/wibox/widget/textbox_spec.lua
```diff
@@ -5,6 +5,17 @@
 
 local textbox = require("wibox.widget.textbox")
 
+local test_dpi_value = 192
+_G.screen = {
```
Aire-One, line 9: This is a general note:

I think I have not seen usage of `_G` outside of spec. I think it's a good practice when accessing to global C API objects.

Are there particular reason for using `_G` only in specs?
Should we recommend to use `_G` as a general good practice to use everywhere in the future?

Aire-One (changes requested):

lib/wibox/widget/textbox.lua
```diff
@@ -375,6 +375,30 @@ function textbox.mt.__call(_, ...)
...
+    playout:context_changed()
+    local attr, parsed = Pango.parse_markup(text, -1, 0)
+    playout.attributes, playout.text = attr, parsed
+    local _, logical = playout:get_pixel_extents()
+    return logical
+end
+
+--@DOC_widget_COMMON@
+
+--@DOC_object_COMMON@
```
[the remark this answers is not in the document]
actionless: thanks, fixed

spec/wibox/widget/textbox_spec.lua
actionless: i don't think here is appropriate time and place to discuss it

Elv13 (approved):

Aire-One (approved):

actionless:
thanks for reviewing, guys! next i gonna rebase 3160 on top of master

PR #3192, opened 2020-09-29
Fix wibox.layout.stack showing the top layer at the bottom

### LINKED ISSUE #3191: `wibox.layout.stack` with offsets draws top layer at the bottom

```
awesome v4.3-886-ged6cdf87b-dirty (Too long)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

Simple demonstration that shows in the top right:

```lua

local red = wibox.widget {widget = wibox.container.background, bg = "#ff0000"}
local green = wibox.widget {widget = wibox.container.background, bg = "#00ff00"}
local blue = wibox.widget {widget = wibox.container.background, bg = "#0000ff"}
local stack = wibox.widget {
   layout = wibox.layout.stack,
   horizontal_offset = 20,
   vertical_offset = 20,
   red,
   green,
   blue
}
stack:raise_widget(red) -- Just to make sure that red is the top layer

local a_wibox = wibox {
   visible = true,
   ontop = true,
   width = 200,
   height = 200,
   bg = "#ffffff",
   widget = stack
}
```

**Actual result:**

Red is the top layer, but is drawn below the others.

image

**Expected result:**

image

dixslyf:
Fixes #3191.

**This fix also makes index 1 the top layer, and updates the docs accordingly**. Previously, the docs stated that index 1 was the bottom, but I think having index 1 as the top is more intuitive and makes more sense. If you want to use index 1 as the bottom, I can update this PR accordingly &mdash; a few other functions would have to be modified though.

### DIFF
--- a/lib/wibox/layout/stack.lua
+++ b/lib/wibox/layout/stack.lua
@@ -5,7 +5,7 @@
 -- a `wibox.widget.textbox` on top of a `awful.widget.progressbar` or manage
 -- "pages" where only one is visible at any given moment.
 --
--- The indices are going from 1 (the bottom of the stack) up to the top of
+-- The indices start from 1 (the top of the stack) down to the bottom of
 -- the stack. The order can be changed either using `:swap` or `:raise`.
 --
 --@DOC_wibox_layout_defaults_stack_EXAMPLE@
--- a/lib/wibox/layout/stack.lua
+++ b/lib/wibox/layout/stack.lua
@@ -74,7 +74,9 @@ function stack:layout(_, width, height)
 
     local h_off, v_off = spacing, spacing
 
-    for _, v in pairs(self._private.widgets) do
+    -- Iterate backwards to draw top (index 1) last so that it appears above other layers
+    for i = #self._private.widgets, 1, -1 do
+        local v = self._private.widgets[i]
         table.insert(result, base.place_widget_at(v, h_off, v_off, width, height))
         h_off, v_off = h_off + self._private.h_offset, v_off + self._private.v_offset
         if self._private.top_only then break end
--- a/lib/wibox/layout/stack.lua
+++ b/lib/wibox/layout/stack.lua
@@ -74,12 +74,16 @@ function stack:layout(_, width, height)
 
     local h_off, v_off = spacing, spacing
 
-    -- Iterate backwards to draw top (index 1) last so that it appears above other layers
-    for i = #self._private.widgets, 1, -1 do
-        local v = self._private.widgets[i]
+    if self._private.top_only then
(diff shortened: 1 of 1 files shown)


lib/wibox/layout/stack.lua
```diff
@@ -74,10 +74,16 @@ function stack:layout(_, width, height)
 
     local h_off, v_off = spacing, spacing
 
-    for _, v in pairs(self._private.widgets) do
+    if self._private.top_only then
```
PlayerNameHere, line 77: Does anyone know if there is a way to leave this check in the for loop? Having the check outside the loop doesn't look very elegant.

I had to take it out because the loop starts with the last layer, so the last layer would show when `top_only` is set instead of the top layer.

Lua doesn't allow us to set loop counter variables, so we can't do something like:
```diff
        -- Iterate backwards to draw top (index 1) last so that it appears above other layers
        for i = #self._private.widgets, 1, -1 do
+           if self._private.top_only then i = 1 end
            local v = self._private.widgets[i]
            table.insert(result, base.place_widget_at(v, h_off, v_off, width, height))
            h_off, v_off = h_off + self._private.h_offset, v_off + self._private.v_offset
        end
```

actionless:
i'm a bit concerned what changing index ordering might affect existing user configs or libraries/widgets (although i'm not using it anywhere myself)

PlayerNameHere:
Yeah, that's definitely a concern.

It seems like there had been mix-ups in the `wibox.layout.stack` code regarding the index ordering for a while though &mdash; for example, `:raise_widget()` checks that `idx ~= 1` before calling `:raise(idx)` at line 144, which means that it uses index 1 as the top. The loop in `:layout()` also uses index 1 as the top layer when `top_only` is set due to the `break` on the first iteration. These were present even before my recently-merged PR. So, if we were to keep the current index ordering where index 1 is used as the bottom, we would have to modify said functions as well, which may also break existing configs.

So it seems like no matter which ordering we decide to go with, someone's config may break.

PlayerNameHere:
This just came across my mind, and now I'm not sure what the expected result should be.

Just to refresh, **positive offsets** were used in the example in #3191, and the following is the **actual result, which is wrong** because red is the top layer but is shown below the other layers:
image

In the linked issue, I put this as the expected result:
image

But *should* that be the expected result? This could fit as the expected result just as well (though is harder to implement):
confusion

dixslyf:
Closing as this is old and I don't have plans to work on this.

PR #3195, opened 2020-10-06
`ruled.client` fixes

Elv13:
Reported by @blueyed

### DIFF
--- a/lib/ruled/client.lua
+++ b/lib/ruled/client.lua
@@ -694,7 +694,7 @@ return setmetatable(module, {
                 crules:append_rules("awful.rules", v)
             end
         else
-            rawset(k, v)
+            rawset(module, k, v)
         end
     end,
     __index = function(_, k)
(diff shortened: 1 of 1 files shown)


actionless (approved):

lib/ruled/client.lua
```diff
@@ -694,7 +694,7 @@ return setmetatable(module, {
                 crules:append_rules("awful.rules", v)
             end
         else
-            rawset(k, v)
+            rawset(module, k, v)
```
blueyed, line 697: Might be worth covering this with a test now?

Let me know and I can look into this myself.

blueyed (commented):
Thanks!

lib/ruled/client.lua
actionless: i was thinking the same from the same glance, until noticed what it's deprecated in 5.x anyway

PR #3205, opened 2020-10-31
textbox: print traceback on error

unai-ndz:
Right now if a textbox fails to parse the markup will print a pretty hard to debug message: `E: awesome: Error on line 28: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;` in the log.

This makes it print the traceback and the error.

### DIFF
--- a/lib/wibox/widget/textbox.lua
+++ b/lib/wibox/widget/textbox.lua
@@ -188,7 +188,7 @@ end
 function textbox:set_markup(text)
     local success, message = self:set_markup_silently(text)
     if not success then
-        gdebug.print_error(message)
+        gdebug.print_error(debug.traceback("Error parsing markup: "..message.."\nFailed with string: '"..text.."'"))
     end
 end
 


lib/wibox/widget/textbox.lua
```diff
@@ -188,7 +188,7 @@ end
 function textbox:set_markup(text)
     local success, message = self:set_markup_silently(text)
     if not success then
-        gdebug.print_error(message)
+        gdebug.print_error(debug.traceback('Error parsing markup: '..message))
```
actionless, line 191: mb also print `text` itself in the end of the message?

unai-ndz:
Now it prints:

> E: awesome: Error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;
Failed with string: 'This text it's not valid pango markup&'
stack traceback:
        /usr/share/awesome/lib/wibox/widget/textbox.lua:201: in function 'set_markup'
        /usr/share/awesome/lib/wibox/widget/textbox.lua:375: in function 'textbox'
        /home/xyze/.config/awesome/rc.lua:44: in main chunk

actionless (approved):
thanks!

Elv13 (approved):

PR #3208, opened 2020-11-05
widget: Added widget_template argument to layoutbox

suharelli:


actionless (approved):

lib/awful/widget/layoutbox.lua
```diff
@@ -32,8 +32,8 @@ local function update(w, screen)
     w._layoutbox_tooltip:set_text(name or "[no name]")
 
     local img = surface.load_silently(beautiful["layout_" .. name], false)
-    w.imagebox.image = img
-    w.textbox.text   = img and "" or name
+    w:get_children_by_id("imagebox")[1].image = img
+    w:get_children_by_id("textbox")[1].text = img and "" or name
```
Aire-One, line 36: What if the user specify a custom widget template with no children "imagebox" and/or "textbox"?

You should also expose the update function and keep this one as a fallback like you did for the widget_template. 

Also, please be sure to update the API documentation to explain this template behavior. (you can take inspiration of the `awful.widget.taglist` doc) 


(next stuff are some thought I have on this topic, you can dismiss it) 

Just like it's done for the `awful.widget.taglist`, widget_template should support create and update callbacks to give the user control over this.

#2955 also started to work in this area. I think we need a common "template widget" build around IoC concepts so we can use it for many pre-defined standard widgets from the awful.widget library.

awful.widget.common is too specific to the taglist. It needs to be refactored to be less specific. OR, we need a new object with more abstraction (awful.widget.template?).

Aire-One (changes requested):

actionless (changes requested):
oh really, docsting is missing for `widget_template`

actionless (approved):
thanks!

Aire-One (approved):
Thanks :)

PR #3217, opened 2020-11-29
[WIP/RFC] Either magic or insanity, be the judge

Elv13:
(first, ignore the line count, there is a large svg, the real one is ~3.5k)

After months of trying to find the will power to finish this: **behold, this is
magic!**. Depending on who you are, this is either the most useless feature I
ever proposed or a game changer. It is also genuinely the largest new feature I
added in years. Everything I have done in the last 4 years was mostly improving
existing stuff, replacing broken/ugly APIs and writing an absurd amount of doc (and
watch people on Reddit keep complaining there is no doc...).

### So, what is it?

This adds 3 new `gears` module to bring (language) feature parity to QML.

#### gears.watcher

This is a rather boring successor to `awful.widget.watch`, but with support
for all types of widgets (not just textboxes). It supports both file and
commands (and may eventually be extended for urls and background threads too).

```
wibox.widget {
    value = gears.watcher {
        files    = {
            capacity = '/sys/class/power_supply/BAT0/energy_full',
            current  = '/sys/class/power_supply/BAT0/energy_now'
        },
        filter   = function(content)
            return tonumber(content.current) / tonumber(content.capacity)
        end,
        interval = 5,
    },
    max_value = 1,
    min_value = 0,
    widget    = wibox.widget.progressbar
}
```

#### gears.connection

A declarative way to add `my_obj:connect_signal` directly in declarative trees.
This removes the need to get an object handle for many simple scenarios. Getting
object from declarative trees is a common pain point. It also expose the objects
`id`s as variables, which has been requested a few time.

```
wibox.widget {
    {
        {
            id        = 'my_graph',
            widget    = wibox.widget.graph
        },
        {
            id     = 'my_label',
        },
        layout = wibox.layout.stack
    },
    id            = 'my_progress',
    widget        = wibox.container.radialprogressbar,

    -- Set the value of all 3 widgets.
    gears.connection {
        source          = my_source_object,
        source_property = 'value',
        callback        = function(_, _, value)
            my_graph:add_value(value)
            my_label.text = value .. 'mB/s'
            my_progress.value = value
        end
    },
}
```

### gears.reactive

The real magic. This brings reactive programming to AwesomeWM. "Any" function
wrapped in a `gears.reactive` object will automatically recall itself each time
a value used within the function changes. This is like an Excel cell where the
value is updated each time the value of its component changes.

```
-- So you don't *need* a hub object, but it's safer to use one.
local my_hub = gears.object {
    enable_properties   = true,
    enable_auto_signals = true
}

-- Better set a default value to avoid weirdness.
my_hub.some_property = 42

-- This is an example, in practice do this in your
-- wibar widget declaration tree.
local w = wibox.widget {
    markup = gears.reactive(function()
        -- Each time my_hub.some_property changes, this will be
        -- re-interpreted.
        return '<i>' .. (my_hub.some_property / 100) .. '</i>'
    end),
    widget = wibox.widget.textbox
}

-- This will update the widget text to '<i>13.37</i>'
my_hub.some_property = 1337
```

#### generic stuff

The adds some generic magic to intercept the magical setter call used by
declarative properties. Any object can define a magic handler and set itself
rather than being automatically set. This is how the 3 classes introduced in
this PR *know* their left hand side property and parent object. It's a private
API for now because it has some limitations. It is rather hard to disconnect
those objects since the setter is too dumb (for now) to know it attached to
such magic object. I have a plan for this, but it's blocked on some other
feature and would make this PR way, way too large.

This adds a way to insert all tree `id`s variables into the function
environment. Eventually, it will be more generic, for now both `gears.reactive`
and `gears.connection` have their own implementation. Long term, I will add
some kind of `on_completed` callback to each widget tree (probably exposed
using `gears.connection`).

I moved a lot of `awful.spawn` code into `gears` as a semi-temporary home.
Eventually, refactoring that into a proper `gears.spawn` module (with more
modern cgroup and limit support) would make sense, but this is out of scope
of this PR. I just wanted to avoid having `gears.watcher` duplicate
`awful.spawn`.

### History / context

I first implemented this back in 2017. However I never submitted it because
@psychon would have hated every single line of the original implementation. I
rewrote it a few time since. This implementation is mostly brand new. I decided
to "degrade" Lua 5.1 support to avoid having 2 entire implementation (one for
Lua 5.1 based on `setfenv` and one for LuaJIT 2.1-beta and Lua 5.2+ based on
`debug.upvaluejoin`.

I think for the first time, the implementation is "less than totally insane" and
might actually be the way to go. It uses proxy objects rather than bloat the
entire codebase. The code takes a few reads before it makes any sense, but it's
not /that/ wierd.

Of course, Lua isn't a reactive language, so bolting that on relies heavily on
on `gears.object` and signals (QML do that too). I also made a few design
decision different from QML. For of all, reactive is **off by default**. It only
do it's thing when wrapped in a `gears.reactive` object. When QML is and isn't
reactive is super confusing and inconsistent. Second, I hardcoded limits on the
reactive scope. It could have been recursive, but that's too insane.

### Todo (in future PRs):

* move the object `id` system to gears.hierarchy
* support the :_set_declartive_handler and friend on all `gears.object`
  properties and constructors
* refactor `wibox.container.conditional` to use this
* unify all object constructor syntax against the `args` notation and make sure
  everything supports `gears.connection` and `gears.reactive` objects.
* add a `gears.connection` based `on_completed` for all declarative trees and
  object constructors.

@psychon This is a lot to ask, but can you review this? I am pretty sure you will
hate some parts. I am also pretty sure you will think this is a lot or code to "just"
"simplify" setting values on widgets. That's a somewhat valid view. However, in
the defense of this, a lot of Stackoverflow and Reddit question are some variants
of "how can I access the widget in <enter specific scenario>". This attack the
problem from 3 angles and hopefully will make those people happy. And, umm,
QML did it first, so I am not *that* insane...


psychon:
> and writing an absurd amount of doc (and watch people on Reddit keep complaining there is no doc...).

Thanks a lot for your work on docs!

> This is a rather boring successor to awful.widget.watch, but with support for all types of widgets (not just textboxes). 

```lua
local foo = awful.widget.watch("echo /path/to/image", 10, function(widget, stdout) widget.image = stdout end, wibox.widget.imagebox)
```

> This is a lot to ask, but can you review this?

Sigh... Uhm... I'm not sure. I'll try to take a look right now. Let's see what happens.

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+--   can be fixed.
+-- * Don't use rawset in the expression.
+-- * If the return value is a table, only the first 3 nesting levels are sanitized.
+--   Avoid using nested tables in the returned value if you can. `gears.object`
+--   instances *should* be fine.
+-- * This module will **NOT** try to track the change of other
+--   functions and methods called by the expression. It is **NOT** recursive
+--   and only the top level properties are tracked. This is a feature, not a
+--   bug. If it was recusrsive, this list of limitations or gotchas would be
+--   endless.
```
psychon, line 76: I think this limitation should be the first one. It actually surprised me at first and only then I thought "the PR mentioned `setfenv`, so this is actually natural...".

Speaking of which: What happens when I use the same function in several different context, some of which should not be tracked? Does this code assume there is only a single caller?

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+-- * There is currently no way to disable a reactive expression once it's
+--   been defined. This may change eventually.
+-- * Rio Lua 5.1 (not LuaJIT 2.1) is currently not support. If you need it,
+--   report a bug.
```
psychon, line 80: Might be worth adding a link at https://github.com/awesomeWM/awesome/issues/new.
At first I somehow thought "report a bug against Lua 5.1? They will just tell you not to use such an old version."

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+--
+-- @author Emmanuel Lepage-Vallee &lt;elv1313@gmail.com&gt;
+-- @copyright 2017-2020 Emmanuel Lepage-Vallee
+-- @classmod gears.reactive
+
+-- This file is provided under the BSD 2 clause license since it is better than
+-- any existing implementation (most of which only support Lua 5.1).
```
psychon, line 87: Please also license it under the same license as AwesomeWM.

I do not know how these license combinations really work (IANAL), but introducing code with a different license sounds more complicated.

psychon (commented):
aaaaand I'm out of time. Sorry, wife requests that we go for a walk and I bet afterwards it is time to set up the tree.

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+end
+
+-- If the sandbox proxies were to leave the reactive expressions, bad
+-- things would happen. Catch functions and constructors to make sure
+-- we strip away the proxy.
+local function is_callable(object)
+    local t = type(object)
+
+    if t == "function" then return true  end
+    if t ~= "table"    then return false end
```
psychon, line 129: What about userdata? Those can also be callable through a metatable.

(I can also make booleans callable with a metatable, but let's ignore that...)

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+    -- likely to happen.
+    for _, v in ipairs(rets) do
+        local ret, _ = create_proxy(v, self._reactive.origin)
+        table.insert(newrets, ret)
+    end
+
+    return unpack(newrets)
+end
+
+-- Build a tree if proxies or return the primitives.
```
psychon, line 239: typo: if -> of

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+-- NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN
+-- IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+local gobject = require("gears.object")
+local gtable = require("gears.table")
+local gtimer = require("gears.timer")
+local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
+
+-- It's only possible to know when something changes if it has signals.
+local function is_gobject(input)
```
psychon, line 118: In a world where `lgi.GObject` exists: Please do not shorten `gears.object` to `gobject`.

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+-- of the sandbox by the reactive expression "siblings" sharing the
+-- same environment.
+local function copy_upvalue_reference(loaded, idx, value)
+    -- Make sure it has a fresh upvalue reference. Something that
+    -- we are sure cannot be shared with something else.
+    local function fake_upvalue_env()
+        return value
+    end
+
+    debug.upvaluejoin(loaded, idx, fake_upvalue_env, 1) --luacheck: ignore
```
psychon, line 345: I definitely do not understand all the magic here, but... by breaking upvalues, could you not break code / cause hard to debug issues? The "function in the sandbox" changed some upvalue but that change is not visible to outside code.

Basically... all writes to upvalues must be forbidden...?

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+    return fake_upvalue_getter
+end
+
+-- Create a copy of the function and replace it's environment.
+local function sandbox(fct, env, vars, values)
+    -- Serialize to make sure it cannot affect the real
+    -- upvalues anymore. Without this setting the proxy actually
+    -- affects other function with the same uplavue environment.
+    local dump   = string.dump(fct)
+    local loaded = load(dump, nil, nil, env)
```
psychon, line 375: Ahhh... that's why you are not messing with functions outside of... things. I bet this cryptic sentence is not understandable, but something makes now a bit more sense to me.

spec/gears/reactive_spec.lua
```diff
@@ -0,0 +1,206 @@
+---------------------------------------------------------------------------
+-- @author Uli Schlachter
+-- @copyright 2014 Uli Schlachter
```
psychon, line 3: :)

lib/awful/util.lua
```diff
@@ -31,8 +31,8 @@ local util = {}
 util.table = {}
 
 --- The default shell used when spawing processes.
--- @param string
-util.shell = os.getenv("SHELL") or "/bin/sh"
+-- @tfield string awful.util.shell
+util.shell = watcher._shell
```
psychon, line 35: Uhm. This means that users cannot configure the shell used by `gears.watcher`.

Since it causes so many problems to have this in `gears`: Why isn't it in `awful` instead? That way, you do not have to move so much stuff over.

lib/gears/watcher.lua
```diff
@@ -0,0 +1,733 @@
...
+local gtimer = require("gears.timer")
+local gtable = require("gears.table")
+local lgi = require("lgi")
+local Gio = lgi.Gio
+local GLib = lgi.GLib
+
+local module = {}
+
+-- This is awful.util.shell
+module._shell = os.getenv("SHELL") or "/bin/sh"
```
psychon, line 18: ...but is useless. The point of `awful.util.shell` is being configurable, I guess the `_` in `_shell` here means that this is not supposed to be configurable here.

psychon:
Okay, my brain hurts. Too much magic going on. (and moving half of `awful.spawn` to `gears` in a non-public place feels weird.)

Kudos for getting this magic to work. I definitely do not understand it and I am not sure I can properly review this. But I definitely understand your goal here. There is just a disconnect for me between the goal and the magic code that makes the magic happen. :-)

psychon:
Shower thought: How about adding support for "a property was access" to `gears.object`? Something like the following:
```lua
local object = {}
do
	local active_watchers = {}
	local object_mt = {}

	function object_mt:__index(k)
		local getter = rawget(self, "get_" .. tostring(k))
		if not getter then error() end

		for watcher in pairs(active_watchers) do
			local tbl = watcher[self]
			if not tbl then
				tbl = {}
				watcher[self] = tbl
			end
			tbl[k] = true
		end

		return getter(self)
	end

	function object.new()
		return setmetatable({ _private = {} }, object_mt)
	end

	function object.watch_object_accesses(f)
		local tbl = {}
		active_watchers[tbl] = true

		-- this needs a pcall()
		f()

		active_watchers[tbl] = false
		return tbl
	end
end

local my_obj = object.new()
function my_obj:get_number1()
	return 1
end
function my_obj:get_number2()
	return 42
end

local accessed = object.watch_object_accesses(function()
	print("the number is:", my_obj.number2)
end)
for obj, access in pairs(accessed) do
	print(obj)
	for key in pairs(access) do
		print("", key)
	end
end
```
Upside: We avoid the `debug` library. The function may call other functions.
Downside: Lots. For example, all objects must use getters etc.

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+-- emits one of the monitored signal, then the whole function is re-evaluated.
+-- Each time the function is evaluated, the "target" properties are updated. The
+-- reactive function result or any call to external function from within goes
+-- through the firewall again and any proxies are removed.
+--
+-- That design has one big limitation. It cannot detect any changes which are
+-- not directly part of `gears.object` instance. You cannot use random tables
+-- and expect the function to be called when it's content change. To work
+-- around this, it is recommanded to make "hub" objects to store the data used
+-- within the reactive function:
```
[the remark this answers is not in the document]
Elv13: fixed

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+-- * This only works with `gears.object` and Core API objects which implement
+--   the `property::*****` signals correctly. If it is a regular Lua table
+--   or the property signals are incorrectly used, the value changes cannot
+--   be detected. If you find something that should work, but doesn't in
+--   one of the AwesomeWM API, please report a bug.
+-- * More generally, when making a custom `gears.object` with custom setters,
+--   it is the programmer responsibility to emit the signal. It is also
+--   required to only emit those signals when the property actually changes to
+--   avoid an unecessary performance loss.
+-- * Changing upvalue types after the reactive expression has been created wont
```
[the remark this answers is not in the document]
Elv13: > but then I thought "I guess you mean having local foo = 42 first and then local foo = some_object_whose_properties_are_accessed", right? 

right

> If so, then this could be made clearer.

Done

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+--   be detected and will cause errors.
```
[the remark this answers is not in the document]
Elv13: fixed

lib/gears/reactive.lua
```diff
@@ -0,0 +1,647 @@
...
+--   you encounter such an issue. But set your expectations, not all corner case
+--   can be fixed.
+-- * Don't use rawset in the expression.
+-- * If the return value is a table, only the first 3 nesting levels are sanitized.
+--   Avoid using nested tables in the returned value if you can. `gears.object`
+--   instances *should* be fine.
+-- * This module will **NOT** try to track the change of other
+--   functions and methods called by the expression. It is **NOT** recursive
+--   and only the top level properties are tracked. This is a feature, not a
+--   bug. If it was recusrsive, this list of limitations or gotchas would be
```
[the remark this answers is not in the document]
Elv13: fixed

lib/gears/reactive.lua
Elv13: > Speaking of which: What happens when I use the same function in several different context, some of which should not be tracked? Does this code assume there is only a single caller?

The function is copied with `strings.dump` and `load` to avoid this issue. I guess you found out "in the future" after this comment. That being said, there is a whole zoo of corner cases to handle when going recursive (and I *did* implement it, then conceded it was a terrible idea):

 * You have to wrap every single function/method/metacall into a custom environment, which is slow
 * It doesn't work past the userdata boundary, which can cause proxies to "leak"
 * You need a custom (and very complex) implementation of `rawset` because some code deep down will use it.
 * If a proxy object ends up leaking, the debugging will be terrible
 * You cannot have `gears.reactive` within a `gears.reactive`. It might happen indirectly if you allow recursion. None of the code above make sure the proxy is for the right sandbox, because you have to shoot yourself in the foot real hard to get there right now.

But more importantly, it would be confusing as hell for the user and might cause infinite loops and/or dead locks and/or stack overflow by calling some internal APIs and I don't want to fix the whole codebase to *always* do `if old ~= new then self:emit_signal("property::foo") end`. It's actually where I gave up the previous time I tried to implement this feature. As it stands right now, there is 2 layers of defense to prevent this system going offtrack and locking AwesomeWM. First of all, it's not recursive, so the "surface" to enter a loop is limited to the function itself. Also, by default, it uses delayed calls, so a loop has to be directly caused by the setter, not some random other code it calls. This removes a whole bunch of potential footguns like the same object signal being emitted twice in the same event loop. It doesn't remove *all* the possible way to trigger a loop, but I think it's 95%+ safe for our API and 99%+ of they use an hub object like the code examples. If the /getter/ causes `property::foo` to be emitted, then it will still enter a loop. Also, if the reactive function calls a setter and *it* doesn't check if the value changed, then it might enter a loop.

One way to debug this would be to track the number of "delayed" loop without a real idle one (by adding some code to the `refresh` signal). Just as we track "slow" even loops, we might track loop chains count and complain if there is >= 100(?) consecutive round of delayed call?

lib/gears/reactive.lua
Elv13: done

lib/gears/reactive.lua
Elv13: BSD 2 clause code can be used in GPL2 code as-is, the dual license isn't required. The GPL virality converts the resulting aggregate from MIT or BSD2 to GPL2. The number of BSD clause is important, BSD4 is incompatible with the GPL (it requires to add the extra clause to your copy of the GPL2, which requires to get every copyright holder to agree. The GPL3 forbids extra clauses unless you rename the license). (IANAL, but have been working with them for 15 years with all the FOSS stuff I did over the years).

lib/gears/reactive.lua
Elv13: > What about userdata?

"maybe fixed". I admit I didn't test this much beyond basic client properties. `gears.reactive` is pretty hard to use with native capi objects... It "works", but there is no reliable way to detect changes to let say, `client.focus` or `awful.screen.focused()` using the "standard" property magic. You have to use the "hub object" concept to mitigate that (as mentioned in the doc). If you do that, you might as well put the values directly in the hub object rather than the client object?

> (I can also make booleans callable with a metatable, but let's ignore that...)

And yet I was crazy enough to do that when replacing some getter/setter methods with properties... Oh well, in retrospect that wasn't the best idea ever...

lib/gears/reactive.lua
Elv13: Fixed

lib/gears/reactive.lua
Elv13: Fixed

lib/gears/reactive.lua
Elv13: I think you get it as per the comment below, but it's "fine" to write the upvalue. `string.dump`/`load` make a copy of the function, then this code makes a copy of the upvalue "vtable". So the lambda function above has a single upvalue (since it has almost no code and the Lua doesn't put all possible upvalues in the vtable. Then I take the *reference* to the lambda upvalue and put it in the target function. This means that when `loaded` accesses that upvalue, it actually "live" in the lambda environment. If you set something, then the value of... `value` in the lambda would change (assuming it has not been GC-ed already), which is harmless. From the point of view of `loaded`, the value was really set. From the point of view of the "real world", nothing happened.

lib/gears/reactive.lua
Elv13: > I bet this cryptic sentence is not understandable,

Rewritten

spec/gears/reactive_spec.lua
Elv13: oops

lib/awful/util.lua
Elv13: fixed

lib/gears/watcher.lua
Elv13: facepalm... fixed

Elv13:
> Okay, my brain hurts. Too much magic going on. 

Thanks a lot for the review. I know this code is... special...

> (and moving half of awful.spawn to gears in a non-public place feels weird.)

It is, but I don't have the energy to fix it the right way. Wayland, Flatpak/Snap and SystemD changed the landscape quite a lot. The next `spawn` API needs to take them into account. SNIDs are dead. CGroups are the new "right way" to know which window belong to which `spawn` call. There is also the extensions to AppStream, Flatpak/Snap/Bubblewrap/FireJail/AppImage metadata and .desktop extensions to handle. A "good" modern "Linux-centric" (but which gracefully degrade itself) API also need to handle limits and quotas... I don't want to make `awful.spawn` to `gears.spawn` and have to deprecate all of it if I ever find the energy to implement the new-cool-things.

Elv13:
Ok, so after a few iteration, I drove the coverage up to 96%. There is still some operator overload stuff to check in `gears.reactive`, but otherwise, this is pretty close to full coverage and full documentation examples.

lib/gears/reactive.lua
psychon: > And yet I was crazy enough to do that when replacing some getter/setter methods with properties...

Wait, what? Where? You cannot "specialise" booleans. Setting the metatable of one boolean sets the metatable of all 
(current and future) booleans.

actionless:
```lua
        callback        = function(_, _, value)
            my_graph:add_value(value)
            my_label.text = value .. 'mB/s'
            my_progress.value = value
        end
```

wouldn't it make luacheck like really crazy?

Elv13:
> wouldn't it make luacheck like really crazy? or that's just a pseudo-code example implying blahblahblah.get_by_id(....)?

It does, but you can just add `--luacheck ignore_stuff` or `--luacheck globals_stuff` labels. It's how I managed it make it shut-up in the tests.

actionless:
i would prefer smth like:

```lua
        callback        = function(_, _, widgets, value)
            widgets.my_graph:add_value(value)
            widgets.my_label.text = value .. 'mB/s'
            widgets.my_progress.value = value
        end
```

Elv13:
> i would prefer smth like:

I see where you are coming from, but I think the "using `id` is declaring a variable" concept is superior on the long term. I think the main point you might have is "this is unexpected". This is a valid concern, but I think it is only unexpected of you never used "this kind" of declarative framework. QML is the most well known of these declarative toolkits. Unity (both the game engine and Ubuntu Unity8 actually) is/are another. CSS "kind of" works like this in the sense you can access HTML/Javascript DOM `id`s. SQL also do it (with `HAVING` and `AS` at least). So this pattern is very common for declarative techs.

A while ago, you proposed/asked for each constructor to be standardized to the `args` format. This is where this design choice will really shine. First of all, it will go beyond widgets and into "all the things" if we are to re-unify the declarative and constructor syntax into one. Something like:

```lua

awful.tag { --yes, I know I will have to break the API for this, it's `awful.tag.add(name, args)` right now.
    id     = "my_tag_id";
    name   = "my_tag";
    screen = screen.primary;

    -- Use `gears.object` directly to "namespace" things, QML style
    some_custom_prop1 = gears.object {
        id   = "custom_prop1";
        foo  = bar;

        -- Use `gears.reactive` expressions to add "plus value"
        -- without any kind of callbacks or imperative code.
        text = gears.reactive(function()
            return "path/to/thing" .. my_tag_id.name .. ".png"
        end);
    };

    -- Declare bare-bone objects and manipulate them externally.
    custom_widget = wibox.widget.imagebox {
        id = "custom_tag_widget"
    };

    -- Do something on an event from a custom property
    gears.connection {
        source          = "custom_prop1";
        source_property = "text",
        callback        = function()
            custom_tag_widget.image = some_custom_prop1.text

             -- OR --
             -- my_tag_id.custom_widget.image = some_custom_prop1.text
             -- OR --
             -- my_tag_id.custom_tag_widget.image = some_custom_prop1.text
             -- 
             -- the difference here is that `custom_tag_widget` always point to the same object
             -- while `my_tag_id.custom_widget` point to whatever is currently attached to the
             -- `my_tag_id` `custom_widget` property.
        end;
    };

    -- Do something when the declarative tree is populated.
    gears.connection {
        signal  = "loaded",
        delayed = true, -- and in the next event loop
        [...]
    }l
}

```

Now, this example in itself is useless and convoluted, but it is to demonstrates how direct access to objects_by_id in `gears.reactive` and `gears.connection` will really simplify the declaration. I also don't like callback functions with a million parameters and you have to count how many `_` you have. Unnamed parameters are really not a nice UX.

@actionless Does it clarify my vision on this and why using `id` to define variables would, if fully implemented, make everybody life better?

actionless:
not by the price of disabling linter - it's extremely important in my coding environment

i can bet all/most of the tools you're referencing have implemented linting extensions which are able to process such variables correctly, but i don't think you are going to do the same for luacheck

i think just one positional argument `context` should be enough, to namespace everything in those callbacks

Elv13:
> not by the price of disabling linter - it's extremely important in my coding environment

It doesn't disable the linter, it makes the linter works *better*.

With what you propose:

```
function(context)
    print(context.myy_graph)
end
```

Prints `nil` with no warning.

With what I propose:

```
function(context)
    -- luacheck: globals my_graph my_label my_progress
    print(myy_graph)
end
```

prints `tests/examples/wibox/decl_doc/connection5.lua:52:17: accessing undefined variable 'myy_graph'`.

Plus, you could easily auto-generate the list and integrate it with luacheck with a simple lpeg or lua-parser module integrated in `make luacheck`. It would not be perfect and some multi-file scenarios would still require the `  -- luacheck: globals varname`, but assuming it's a single file, it would work 90%+ of the time. We could ship a luacheck wrapper as part of `awesome -k` too?

So I don't really buy your argument. If anything, what I propose improves the status quo.

Is such module a mandatory addition to this PR to make you approve it? Can it be done later? Are you still unconvinced?

actionless:
> We could ship a luacheck wrapper

that for sure resolve my concern

because:

```
function(context)
    -- luacheck: globals myy_graph my_label my_progress
    print(myy_graph)
end
```

OmegaRogue:
How well does this currently work? Is it usable? having reactive would really clean up my code, currently I'm wrapping basically everything in [awestore stores](https://github.com/K4rakara/awestore)

Elv13:
It works fine on Lua 5.2 and higher, including luaJIT. However the concerns mentioned above are valid and unresolved.

actionless:
> Prints nil with no warning.
> prints tests/examples/wibox/decl_doc/connection5.lua:52:17: accessing undefined variable 'myy_graph'.

i think in lua is more canonical achieving that same behavior by just setting up a custom setter in the table, which would be throwing that desired error instead of returning `nil`

PR #3233, opened 2020-12-24
fix(gears: colors: recolor_image): always duplicate the surface

actionless:


psychon:
Fine with me, although technically this is a breaking change since some code out there could depend on this function not creating a copy of the surface.

actionless:
but from reading documentation no one could get an impression what it's modifying existing object if it's not a string with path

in documentation is stated what it returns recolored image

so IMO it's a fix, not API change

actionless:
i was just never noticing the bug because never attempting to use the same image with different colors until yesterday :)

actionless:
BUMP

sclu1034 (approved):

Aire-One:
I think this is ok for merge.

However, I understand the concern @psychon has.
Would it be a good practice to wrap this change behind an `awesome.api_lvl > 4` condition? So the old behavior wouldn't be break and we still introduce the fix into the code base.

Maybe it's a bit overshot for such a bug fix?

actionless:
i feeling a bit stupid repeating myself, but here we are:

that behavior was always a bug, nothing  in the documentation implies what the current behavior is intentional

Elv13 (approved):
Yeah, this is a bug, I agree.

Issue #3236, opened 2020-12-29
Touchegg CHANGE_DESKTOP action support (_NET_CURRENT_DESKTOP)


F0xedb:
Output of `awesome --version`:

```
awesome v4.3-897-g66b1780d8-dirty (Too long)
 • Compiled against Lua 5.3.6 (running with Lua 5.3)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

**How to reproduce the issue:**

1. Install `touchegg` 
2. Add the following gesture in touchegg: `~/.config/touchegg/touchegg.conf`
```xml
<gesture type="SWIPE" fingers="4" direction="RIGHT">
  <action type="CHANGE_DESKTOP">
    <direction>previous</direction>
    <animate>true</animate>
    <animationPosition>left</animationPosition>
    <color>3E9FED</color>
    <borderColor>3E9FED</borderColor>
  </action>
</gesture>
```
3. start touchegg daemon: `systemctl start touchegg`
4. Run touchegg client: `touchegg`
5. Swipe with 4 fingers to the right on a touchpad

**Actual result:**

When doing so you are greater with an arrow pointing to the next tag indicating that touchegg wants to switch to that tag.
However nothing happens afterwards.
[Related to this touchegg issue](https://github.com/JoseExposito/touchegg/issues/418)

**Expected result:**

The tags don't get switched.
As far as I understand it, touchegg sends a `_NET_CURRENT_DESKTOP` client message to the root window with the updated value. However `awesomewm` doesn't update it.

The `_NET_CURRENT_DESKTOP` spec: https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#idm45760170877056

psychon:
A `_NET_CURRENT_DESKTOP` client message is handled something like by the C code (pseudo-lua-code):
```lua
local t = tag_that_the_message_is_sent_for();
t:emit_signal("request::select", "ewmh")
```
The rest is up to Lua code. `lib/awful/tag.lua` seems to handle this by calling `t:view_only()`, i.e. it always honors the request. Uhm...

Could you add this to your config and see if it works?
```lua
tag.connect_signal("request::select", function(t, context)
    naughty.notify{ text = "request to select " .. tostring(t.name) .. " with context .. tostring(context) }
[...]

Issue #3204, opened 2020-10-30
xdotool set_desktop 1 doesn't change workspace


daraul:
I just realized that `xdotool set_desktop 1` doesn't change workspace like it should.  

My awesome version:

```
awesome v4.3-862-g7a759432d-dirty (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

psychon:
I think #3236 is the duplicate where this was debugged and #3237 is the fix.

Issue #3047, opened 2020-03-23
Switching windows with wmctrl is broken


h5vx:
Output of `awesome --version`:
```
awesome v4.3-809-g5d1394b91 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

**How to reproduce the issue:**
1. Have some windows open
1. Choose window you want to switch, remember window _title_
1. Switch to some another workspace without that window
1. Run command: wmctrl -a _title_

**Actual result:**
* That window is urgent now
* Workspace is not switched to that window
* That window is not made active

**Expected result:**
* Workspace is switched to that contains that window
* That window is made active

**That works correctly on same system and config with that version**:
```
awesome v4.3 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

h5vx:
Only reason i use git build is that cool feature that add window icons near title... It would be cool if that feature come out to Arch repo build

psychon:
@Elv13 I took a quick look at `awful.permissions` and... I have no idea what is "the right way" to make this work. I only see https://github.com/awesomeWM/awesome/blob/master/lib/awful/permissions/init.lua#L196-L200 which seems loosely related, but also not really helpful for this concrete request (no idea if I am even looking in the right place).

Elv13:
Like that: https://awesomewm.org/apidoc/core_components/client.html#grant but right now only the skeleton of that is implemented, the end result when the other half will be merged will favor using that from the rules. Anyway, yes, it is probably what regressed wmctrl. I don't have that much time right now, but I will add a wmctrl test suite next time I work on it.

daraul:
I'm not sure if this is related, but `xdotool set_desktop 1` doesn't work either. My awesome version: 

```
awesome v4.3-862-g7a759432d-dirty (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
[...]

PR #3237, opened 2020-12-30
Fix _NET_CURRENT_DESKTOP ClientMessage handling

psychon:
This was broken in commit c5202a48708585cc33 by adding an extra string
to the Lua stack. Hence, the offset -1 now referred to the string and
not the tag.

Fixes: https://github.com/awesomeWM/awesome/issues/3236

### DIFF
--- a/ewmh.c
+++ b/ewmh.c
@@ -468,7 +468,7 @@ ewmh_process_client_message(xcb_client_message_event_t *ev)
             lua_State *L = globalconf_get_lua_State();
             luaA_object_push(L, globalconf.tags.tab[idx]);
             lua_pushstring(L, "ewmh");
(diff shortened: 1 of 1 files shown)


Aire-One (approved):

actionless (approved):

psychon:
I just found some duplicates. This not only fixes #3236, but also #3047 and #3204, I bet.

blueyed (approved):
Would be great to have this covered with some test(s).
@Elv13 
Do you think this can be done easily?

Aire-One:
Do we need to close/reopen this to make codecov run again? It seems it just randomly crashed...

Elv13 (approved):

PR #3238, opened 2021-01-02
Fixed wrong handling of negative spacing in layout.fixed

### LINKED ISSUE #3235: Improper layout in aweful.taglist when using neagative spacing and a fixed horizontal layout in the containing widget

awesome v4.3 (Too long)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

Copy the example at the top of the documentation for awful.taglist into the default rc.lua:

```
   s.mytaglist = awful.widget.taglist {
        screen  = s,
        filter  = awful.widget.taglist.filter.all,
        style   = {
            shape = gears.shape.powerline
        },
        layout   = {
            spacing = -12,
            spacing_widget = {
                color  = '#dddddd',
                shape  = gears.shape.powerline,
                widget = wibox.widget.separator,
            },
            layout  = wibox.layout.fixed.horizontal
        },
        widget_template = {
            {
                {
                    {
                        {
                            {
                                id     = 'index_role',
                            },
                            margins = 4,
                        },
                        bg     = '#dddddd',
                        shape  = gears.shape.circle,
                    },
                    {
                        {
                            id     = 'icon_role',
                        },
                        margins = 2,
                    },
                    {
                        id     = 'text_role',
                    },
                },
                left  = 18,
                right = 18,
                widget = wibox.container.margin
            },
            id     = 'background_role',
        },
        buttons = taglist_buttons
    }
```

I cut down version of the example with only the "text role" is sufficient to reproduce the results.

**Actual result:**

The taglist is drawn with the last two tags clipped, however it appears that enough space has been allocated on the wibar for them, they are just no being drawn correctly.  I have set the background colour to bright green to show the space that has been allocated on the wibar for the taglist.

sc2

**Expected result:**

If you remove all of the wigets from the left side of the default wibar configuration (including the layout), except the taglist, the taglist will be drawn properly: 

sc3

Adding the fixed horizontal layout back into the the left side of the default wibar is enough to cause the erroneous behaviour.  Switching the outer layout to anything other than fixed horizontal removes this layout issue.  So I think that this might be an issue with interactions between the instanced of the fixed horizontal layout.

ShayAgros:
While writing unit-testing for the layout.fixed module I noticed other issues caused by placing widgets with negative spacing. This PR fixes all these issues and refactors layout and fit functions to remove code duplication and improve readability (imho at least).

Tested the code with unit testing and ran the module locally with no problems noticed. I tried to be as verbose as possible in the commit message to describe what I'm trying to fix

Fixes #3235

### DIFF
--- a/lib/wibox/layout/fixed.lua
+++ b/lib/wibox/layout/fixed.lua
@@ -263,16 +263,17 @@ end
 function fixed:fit(context, orig_width, orig_height)
     local width_left, height_left = orig_width, orig_height
     local spacing = self._private.spacing or 0
+    local widgets_nr = #self._private.widgets
     local is_y = self._private.dir == "y"
     local used_max = 0
 
     -- when no widgets exist the function can be called with orig_width or
     -- orig_height equal to nil. Exit early in this case.
-    if #self._private.widgets == 0 then
+    if widgets_nr == 0 then
         return 0, 0
     end
 
-    for _, v in pairs(self._private.widgets) do
+    for k, v in pairs(self._private.widgets) do
         local w, h = base.fit_widget(self, context, v, width_left, height_left)
         local max
 
@@ -288,23 +289,33 @@ function fixed:fit(context, orig_width, orig_height)
             used_max = max
         end
 
-        if width_left <= 0 or height_left <= 0 then
+        if k < widgets_nr then
             if is_y then
-                height_left = 0
+                height_left = height_left - spacing
             else
-                width_left = 0
+                width_left = width_left - spacing
+            end
+        end
+
+        if width_left <= 0 or height_left <= 0 then
+            -- this complicated two lines determine whether we're out-of-space
+            -- because of spacing, or if the last widget doesn't fit in
+            if is_y then
+                 height_left = k < widgets_nr and height_left + spacing or height_left
+                 height_left = height_left < 0 and 0 or height_left
+            else
+                 width_left = k < widgets_nr and width_left + spacing or width_left
+                 width_left = width_left < 0 and 0 or width_left
             end
             break
         end
     end
 
-    local total_spacing = spacing * (#self._private.widgets-1)
-
     if is_y then
-        return used_max, orig_height - height_left + total_spacing
+        return used_max, orig_height - height_left
     end
 
-    return orig_width - width_left + total_spacing, used_max
+    return orig_width - width_left, used_max
 end
 
 function fixed:reset()
--- a/lib/wibox/layout/fixed.lua
+++ b/lib/wibox/layout/fixed.lua
@@ -20,45 +20,61 @@ local fixed = {}
 -- @param height The available height.
 function fixed:layout(context, width, height)
     local result = {}
-    local pos,spacing = 0, self._private.spacing
-    local spacing_widget = self._private.spacing_widget
+    local spacing = self._private.spacing or 0
     local is_y = self._private.dir == "y"
     local is_x = not is_y
     local abspace = math.abs(spacing)
     local spoffset = spacing < 0 and 0 or spacing
+    local widgets_nr = #self._private.widgets
+    local spacing_widget
+    local x, y = 0, 0
+
+    spacing_widget = spacing ~= 0 and self._private.spacing_widget or nil
 
     for k, v in pairs(self._private.widgets) do
-        local x, y, w, h, _
+        local w, h = width - x, height - y
+
         if is_y then
-            x, y = 0, pos
-            w, h = width, height - pos
-            if k ~= #self._private.widgets or not self._private.fill_space then
-                _, h = base.fit_widget(self, context, v, w, h);
+            if k ~= widgets_nr or not self._private.fill_space then
+                h = select(2, base.fit_widget(self, context, v, w, h))
             end
-            pos = pos + h + spacing
+
+            if y - spacing >= height then
+                -- pop the spacing widget added in previous iteration if used
+                if spacing_widget then
+                    table.remove(result)
+                end
+                break
+            end
         else
-            x, y = pos, 0
-            w, h = width - pos, height
-            if k ~= #self._private.widgets or not self._private.fill_space then
-                w, _ = base.fit_widget(self, context, v, w, h);
+            if k ~= widgets_nr or not self._private.fill_space then
+                w = select(1, base.fit_widget(self, context, v, w, h))
             end
-            pos = pos + w + spacing
+
+            if x - spacing >= width then
+                -- pop the spacing widget added in previous iteration if used
+                if spacing_widget then
+                    table.remove(result)
+                end
+                break
+            end
         end
 
-        if (is_y and pos-spacing > height) or
-            (is_x and pos-spacing > width) then
-            break
-        end
+        -- Place widget
+        table.insert(result, base.place_widget_at(v, x, y, w, h))
 
-        -- Add the spacing widget
-        if k > 1 and abspace > 0 and spacing_widget then
+        x = is_x and x + w + spacing or x
+        y = is_y and y + h + spacing or y
+
+        -- Add the spacing widget (if needed)
+        if k < widgets_nr and spacing_widget then
             table.insert(result, base.place_widget_at(
                 spacing_widget, is_x and (x - spoffset) or x, is_y and (y - spoffset) or y,
                 is_x and abspace or w, is_y and abspace or h
             ))
         end
-        table.insert(result, base.place_widget_at(v, x, y, w, h))
     end
+
     return result
 end
 
--- a/lib/wibox/layout/fixed.lua
+++ b/lib/wibox/layout/fixed.lua
@@ -261,40 +261,50 @@ end
 -- @param orig_width The available width.
 -- @param orig_height The available height.
 function fixed:fit(context, orig_width, orig_height)
-    local width, height = orig_width, orig_height
-    local used_in_dir, used_max = 0, 0
+    local width_left, height_left = orig_width, orig_height
+    local spacing = self._private.spacing or 0
+    local is_y = self._private.dir == "y"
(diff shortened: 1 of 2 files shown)


psychon (dismissed):
Let's see what happens...

ShayAgros:
@psychon Tweaked my Awesome config today, and I found a use-case where my third commit (`layout/fixed: Prevent overloading widgets with negative spacing`) might be suboptimal.

Take for example the [second example](https://awesomewm.org/apidoc/widgets/awful.widget.tasklist.html) of tasklist widget which creates the following tasklist element:

tasklist_example

Here is a reference to its widget template
```lua
widget_template = {
        {
            wibox.widget.base.make_widget(),
            forced_height = 5,
            id            = 'background_role',
            widget        = wibox.container.background,
        },
        {
            awful.widget.clienticon,
            margins = 5,
            widget  = wibox.container.margin
        },
        nil,
        layout = wibox.layout.align.vertical,
    },
```

I tried to change the order of the `client icon` and the `backgroud_role` to make the highlighted line below the client's icon. I found that this doesn't work because the `client icon` takes all the available space and the second one doesn't get layout afterwards.

To mitigate it I added negative spacing, making the widget template look like this
```lua
widget_template = {
        {
            awful.widget.clienticon,
            margins = 5,
            widget  = wibox.container.margin
        },
        {
            wibox.widget.base.make_widget(),
            forced_height = 5,
            id            = 'background_role',
            widget        = wibox.container.background,
        },
        spacing = -5,
        layout = wibox.layout.fixed.vertical,
    },
```

This code doesn't work with my third commit because the fixed layout no longer "overloads" widgets one upon another in case the first one exhausts all available space. Nevertheless, the use case I described above seems valid and might be used by other people.

It should be possible to achieve what I tried to do with `ratio` layout. Do you think the third commit should stay?

psychon:
Puh... I'm out, sorry.

ShayAgros:
Hmm okay. I don't mind dropping the last commit, it isn't related to the bug in #3235, just something I spotted and thought needs changing.
@Aire-One @Elv13 @actionless do you have a take on this ? (sorry if I missed someone, these names come frequently in issues)

sclu1034:
> I tried to change the order of the `client icon` and the `backgroud_role` to make the highlighted line below the client's icon.

Did you make sure to move the `nil` "widget" accordingly?
This works fine for me:

```lua
widget_template = {
    nil,
    {
        awful.widget.clienticon,
        margins = 5,
        widget = wibox.container.margin
    },
    {
        wibox.widget.base.make_widget(),
        forced_height = 5,
        id = 'background_role',
    },
    layout = wibox.layout.align.vertical
}
```
---

Regarding the third commit, I think the version without overloading makes more sense. Without spacing, the widget drops everything once it has filled its available space, so it should do the same with spacing.

If anything, I would expect negative spacing to act like a sort of negative margin. Taking the example from the commit message in https://github.com/awesomeWM/awesome/commit/857890d77fecfa50cd3e7d9a845ac453be150f61, it would create this:

```lua
{
    widget1: { x = 0, y = 0, width = 10, height = 10 }
    -- spacing: { x = 5, y = 0, width = -5, height = 10  }
    widget2: { x = 5, y = 0, width = 5, height = 10  }
}
```

The spacing widget gets removed, as a width of `-5` doesn't make sense. `widget2` starts at `widget1.x + widget1.width + spacing`, as usual. But since `spacing == -5`, it moves _into_ `widget1`. And `widget2.width == 5`, because there is only that much space left in the layout.

---

Another idea regarding `layout.fixed` in this situation: How about adding a `reverse` flag to `layout.fixed`?
The current version of `layout.fixed` always starts at the beginning of its widget list and places them from the top/left. With `reverse = true`, it would instead start at the end of its widget list and place them from the bottom/right.

The following template would then produce the same output as my code with `layout.align` above, as it first places the indicator and then the icon:

```lua
widget_template = {
    {
        awful.widget.clienticon,
        margins = 5,
        widget = wibox.container.margin
    },
    {
        wibox.widget.base.make_widget(),
        forced_height = 5,
        id = 'background_role',
    },
    reverse = true,
    layout = wibox.layout.fixed.vertical
}
```

ShayAgros:
@sclu1034 thank you for taking the time to review this, very (very) much appreciated.

Your snippet works for me as well, so I guess there is no real damage in applying the third commit as the work-around isn't complicated. Changed the code on my setup to your suggestion (: thanks for that

> If anything, I would expect negative spacing to act like a sort of negative margin. Taking the example from the commit message in 857890d, it would create this:
```
{
    widget1: { x = 0, y = 0, width = 10, height = 10 }
    -- spacing: { x = 5, y = 0, width = -5, height = 10  }
    widget2: { x = 5, y = 0, width = 5, height = 10  }
}
```

Hmm I'm not sure where did you take the `width = -5` part from. I don't see it in the commit message you linked to.
I'm not sure I follow your suggestion ... Are you proposing that we layout the second widget in this example ? Even though the first widget takes all the available space ?
Also not sure why you commented out the spacing widget, I cannot layout 2 adjacent widgets w/o drawing the spacing widget between them, this would make `fixed layout` inconsistent in handling the spacing.

To be consistent with the case where no spacing is applied (or when spacing is positive for that matter), the third commit should be applied as is (might be missing something here since I didn't follow your argument).

> Another idea regarding layout.fixed in this situation: How about adding a reverse flag to layout.fixed?

I don't know how practical it would be since you can "simulate" its purpose using `align` version of this layout, but for the sake of user's freedom to tweak as he/she likes, I'm ok with adding it.

If/once this patchset is merged, I'd submit a PR that proposes this change.

psychon:
> @ShayAgros ShayAgros closed this 10 hours ago

Huh...?

ShayAgros:
Sorry, I apparently don't know how to do proper `git push`, removed the commits from my local fork by mistake. Reopening

sclu1034:
> I don't know how practical it would be since you can "simulate" its purpose using `align` version of this layout

That only works nicely for up to three widgets. If you want more widgets than that, it'll become a complex nesting structure.
`flex` and `fixed`, however, work fine with an arbitrary number of children.

---

The way I see negative spacing:

**Option A:**
Consider spacing a positive offset/width and clamp every `spacing < 0` to `0`.
This is probably the easiest version to document and wrap your head around.

**Option B:**
Try to find some use for negative spacing. I don't see a use for it when there is a spacing widget provided by the user, but without a widget it could be used similar to how negative margin works in HTML+CSS.
And this wouldn't be inconsistent with how positive spacing without a widget works. In both cases it's just an offset.

Which leads to

```lua
-- spacing: { x = 5, y = 0, width = -5, height = 10  }
```

To visualize what I was trying to say with this, take a piece of paper and fold it twice in a _Z_ shape. You will get two outer layers (the user widgets) which may cover each other slightly or even fully and an inner layer (the spacing widget), which "moves into" the outer layers (spacing is negative), is flipped (width is negative) and is always fully covered (commented out).

A use case could be something like Chrome's old tab bar. When the tab shape was still a trapezoid, they would overlap at the corners.

ShayAgros:
> That only works nicely for up to three widgets. If you want more widgets than that, it'll become a complex nesting structure.
flex and fixed, however, work fine with an arbitrary number of children.

Yup you're right here, this does make it essential

> Option A:

Don't like it, I have use cases myself (such as the example I gave initially) for negative spacing, I'd rather have it there.

> Option B:

I don't see the harm in leaving the spacing widget. You could of course get the same affect with just negative spacing and an explicit spacing widget, but I don't see a reason to disable this functionality. This might break other people's setups, with no evident gain (at least from how I see it).

I agree with @sclu1034 argument that negative spacing layout should behave the same as positive one in the sense that once we used all the available width, we stop laying out more widgets.

My work plan then is to add this `reverse` flag to the third commit to allow creating the example I showed using `fixed` layout. Hopefully by the time it is done we'd have our automation back and enough people in the community with write access to this repository.

@sclu1034 agreed ?

sclu1034:
Sounds good.

blueyed (commented):
I cannot help here really, sorry.
But I think the code cleanup/refactoring could be merged already through a separate PR maybe, until questions/issues here are solved.

sclu1034:
@ShayAgros FYI, you're currently pushing your tests for GitHub Actions to this PR. You might want to move those tests to a separate branch.

ShayAgros:
> @ShayAgros FYI, you're currently pushing your tests for GitHub Actions to this PR. You might want to move those tests to a separate branch.

This was a grave mistake to make this pull request to track master in the first please \= I'll stop doing that.
Thanks @sclu1034 . Sorry for whole the noise.
I hope this PR could just be merged once the CI is done (I'm testing your PR w/ some suggestions).

ShayAgros:
Couldn't find how to make this PR to track any branch that isn't master on my side.

I'll just continue to do my tests on a different branch to not mess this PR. Sorry for all the noise. I hope this PR would be accepted soon (it really seems like everyone is ok w/ it and only a `reverse` option is missing which I don't add because of the CI that is pending).

Again, sorry for the noise

Elv13:
@ShayAgros Can you make that other PR with the refactoring part and the new tests that pass? I will try to take a deeper look at this soon-ish, but I got a literal year of backlog tasks in front, so it might take some time to find the time to deep dive into the geometry code.

ShayAgros:
@Elv13 hi, sorry for the late reply. It's been intense last weeks at work, so I was less active here as well. I can submit a new PR, but what tests don't pass ?

actionless (approved):

Elv13 (approved):

albel727:
This seems to break the awful.widget.prompt widget. E.g. mod+R and typing stuff doesn't display anything until one does something to refresh the wibar, like switching tags. I haven't investigated why that happens, but git bisect points me at the `layout/fixed: Prevent overloading widgets with negative spacing` commit.

PR #3243, opened 2021-01-12
Update 01-readme.md, add more details to debian install instructions, and give cmake arguements, so nobody has to suffer through the hell that is installing wrong

### LINKED ISSUE #3106: Cannot Upgrade to Git from Repo [debian]

awesome v4.3 (Too long)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

sudo apt install awesome
sudo apt build-dep awesome
git clone https://github.com/awesomewm/awesome
cd awesome
cd build
make package
sudo dpkg -i *.deb

**Actual result:**
Awesome breaks and doesn't show in DM, and is broken

**Expected result:**
Updated brinaries

Outcome: https://pastebin.com/mwV8a62g
Expected: https://pastebin.com/WDuSDhCG

HolySuicidalTurkey:
Edit to installation instructions, described in 
"Cannot Upgrade to Git from Repo [debian]" #3106

this is my first pull request ever :) 
sorry if i did it wrong

Fixed #3106

### DIFF
--- a/docs/01-readme.md
+++ b/docs/01-readme.md
@@ -29,7 +29,9 @@ pass arguments to `cmake`, please use the `CMAKE_ARGS` environment variable. For
 instance:
 
     CMAKE_ARGS="-DCMAKE_INSTALL_PREFIX=/opt/awesome" make
-
+    
+cmake argument for debian:
+`"-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -DSYSCONFDIR=/etc"`
 
 ### Installing current git master as a package receipts
 
@@ -49,6 +51,8 @@ sudo apt build-dep awesome
 git clone https://github.com/awesomewm/awesome
 cd awesome
 make package
+cd build
+make distclean package CMAKE_ARGS="-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -DSYSCONFDIR=/etc"
 sudo apt install *.deb
 ```
 
@@ -186,4 +190,4 @@ Online documentation is available [here](https://awesomewm.org/apidoc/).
 
 The project is licensed under GNU General Public License v2 or later.
 You can read it online at ([v2](http://www.gnu.org/licenses/gpl-2.0.html)
-or [v3](http://www.gnu.org/licenses/gpl.html)).
+or [v3](http://www.gnu.org/licenses/gpl.html)). 


docs/01-readme.md
```diff
@@ -49,6 +51,8 @@ sudo apt build-dep awesome
 git clone https://github.com/awesomewm/awesome
 cd awesome
 make package
+cd build
+make distclean package CMAKE_ARGS="-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -DSYSCONFDIR=/etc"
```
psychon, line 55: This does not work. I tried following the instructions (more or less, I didn't clone from GitHub but from my local git repo, but that shouldn't make a difference, I guess).
```
$ LC_ALL=C make distclean
make: *** No rule to make target 'distclean'.  Stop.
```
However, half of this isn't necessary, I think. Just add the `CMAKE_ARGS` directly to the first `make package` invocation. The `make distclean` isn't needed when one freshly cloned the repo.

For the CMake arguments: Well, this seems to be what Debian uses on x86-64. I'm not sure how much of this is really necessary for users. I would guess that the following would be enough, but cannot really test:
```
make package CMAKE_ARGS="-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_SYSCONFDIR=/etc -DSYSCONFDIR=/etc"
```

docs/01-readme.md
```diff
@@ -29,7 +29,9 @@ pass arguments to `cmake`, please use the `CMAKE_ARGS` environment variable. For
 instance:
 
     CMAKE_ARGS="-DCMAKE_INSTALL_PREFIX=/opt/awesome" make
-
+    
+cmake argument for debian:
+`"-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -DSYSCONFDIR=/etc"`
```
psychon, line 34: I'm not sure about this. I would prefer if this were trimmed down to the necessary options and these options are also explained. However, do we want/need that? Isn't it enough to have them in the list below (where they IMO do not need to be explained)?

psychon (changes requested):
Thanks a lot for your pull request. You did nothing wrong. I still found something to complain about. Sorry. ;-)

If you want to make further changes to the PR, you can simply push more commits to your `patch-2` branch (it seems like you created the pull request on GitHub's website: Well, it automatically created a `patch-2` branch for you and... well... I bet the docs explain it better than I can, but I am not sure where those docs would be.)

Anyway, what do others think about this? I am not really sure what the best course of actions here is, but it is definitely a good idea to improve the docs.

As a tip for your second pull request, you could add something like `Fixed #3106` to the commit message. This makes GitHub automatically close that issue when the PR is merged.

docs/01-readme.md
HolySuicidalTurkey: okay, so now i edit the pull request? or do i submit a new one with the edit that you outlined?

docs/01-readme.md
HolySuicidalTurkey: i really don't know the necessary options, i just pasted what i saw in the example on #3106

docs/01-readme.md
```diff
@@ -113,10 +114,12 @@ Additionally, the following optional dependencies exist:
...
 You can directly select Awesome from your display manager. If not, you can
-add the following line to your `.xinitrc` to start Awesome using `startx`
+add the following line to the end of your `.xinitrc` to start Awesome using `startx`
 or to `.xsession` to start Awesome using your display manager:
 
     exec awesome
+    
+make sure this .xinit file is in your home directory. Do this by typing `vim ~/.xinitrc` and add the line above
 
 In order to connect Awesome to a specific display, make sure that
```
ShayAgros, line 126: Not directly related to your commit, but how does adding `exec awesome` to `~/.xinitrc` connects the WM to a specific display ?

I'm not an Xorg expert but since `xrandr` works regardless of the WM/DE being used. Maybe this line should change as well in this patch

docs/01-readme.md
```diff
@@ -186,4 +190,4 @@ Online documentation is available [here](https://awesomewm.org/apidoc/).
 
 The project is licensed under GNU General Public License v2 or later.
 You can read it online at ([v2](http://www.gnu.org/licenses/gpl-2.0.html)
-or [v3](http://www.gnu.org/licenses/gpl.html)).
+or [v3](http://www.gnu.org/licenses/gpl.html)).
```
ShayAgros, line 194: Please remove the added trailing space

docs/01-readme.md
ShayAgros: @psychon I don't think this addition belongs here since there already a section on how to install awesome on debian.

Maybe adding this `make` command as part of the `Debian-based` section as part of the `Installing current git master as a package receipts` section ? Something like this
```
sudo apt build-dep awesome
git clone https://github.com/awesomewm/awesome
cd awesome
make package CMAKE_ARGS="-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -DSYSCONFDIR=/etc"
sudo apt install *.deb
```
(similar to what is already done in this commit)

I still think it's better to remove redundant cmake parameters. Maybe do trial and error on a debian machine to see when it works ?

ShayAgros (changes requested):
Thank you for the work you've done.

The commits should be reworked. You submitted 3 commits (where only the first contains a commit message) which can be merged into a single commit.
Also it is a little irregular to add
```
this is my first pull request ever :) 
sorry if i did it wrong
```

As part of the commit message since the message is intended to explain the code for future programmers that look at it.

I think that modifying `debian` section in the `installation` part gives enough information on how to install `debian` without duplicating the `cmake` arguments in several spots in the document.
Also maybe do some trial and error to see what arguments really *have* to be present.
The `startx` edit lgtm, but I'd read ahead and see if the rest of the description in this part makes sense (see my 'specific display' comment there).

Again, thanks for your work. I hope I didn't scare you with the amount of comments I left... This is part of the code review process (in my work at least), and it is done to create a better code at the end.

PR #3249, opened 2021-01-27
Make gears.string.*with functions more permissive

### LINKED ISSUE #3248: gears.string startswith/endswith, client.name and nil value
awesome v4.3-895-g538586c17-dirty (Too long)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no

The issue appears when there is a client with no title (like a "save as" dialog).
```
local gstring = require("gears.string")
for  _, c in ipairs(client.get()) do
    if gstring.endswith(c.name, "anything") then
        (...)
    end
end
```

**Actual result:**

```
stack traceback:
        [C]: in function 'string.sub'
        /usr/share/awesome/lib/gears/string.lua:195: in function 'gears.string.endswith'
        (...)
        error: /usr/share/awesome/lib/gears/string.lua:195: bad argument #1 to 'sub' (string expected, got nil)
```

**Expected result:**

No error, just gstring.endswith to return false for the nameless client.

My custom function containing the code above worked for weeks and suddenly, of course when I was in a hurry for a task, it stopped working. 
I did not expect to have to check for client's name validity. One should not have to know the internals of awesome for such a simple code.

Either a client with no name should have a client.name value to an empty string,
or gears.string.*with functions should return false if the input str is nil.

I would go for the gears.string "protection", because nil can never start or end with something, so it should just return false. It's logical, and can't break existing code, so it's the safest fix for this particular case.

Grumph:
Fix #3248


Aire-One (approved):
Looks good to me. You should remove the draft status to allow maintainers to merge. 

The only check falling is the flacky `test-awful-placement.lua`, so there is nothing to see here.

Thank you for the work.

psychon (approved):

Elv13 (approved):

PR #3253, opened 2021-01-30
Relocate a client window as if it is undecorated when reparenting it back

xinhaoyuan:
This helps resolve some annoying offsets in floating mode after closing and reopening some trayed applications, for example, Telegram. I'm not sure if this is breaking any protocols (I read some ICCCM but haven't found anything related yet).

### DIFF
--- a/tests/test-geometry.lua
+++ b/tests/test-geometry.lua
@@ -6,9 +6,13 @@ local wibox     = require( "wibox"     )
 local beautiful = require( "beautiful" )
 local cruled = require("ruled.client")
 local gdebug = require("gears.debug")
+local test_client = require("_client")
+local lgi = require("lgi")
+local gears = require("gears")
 
 local w = nil
 local w1_draw, w2_draw
+local windowid
 
 -- Replacing the rules is not supported anymore.
 local dep = gdebug.deprecate
@@ -257,6 +261,113 @@ local steps = {
     end
 }
 
-runner.run_steps(steps)
+-- Tests for unmanaged client geometries with different gravity settings.
+local test_data = {
+    -- gravity => expectation of unmanaged x, y, w, h.
+    ["NORTH_WEST"] = {0,  0,  100, 80},
+    ["NORTH"] =      {10, 0,  100, 80},
+    ["NORTH_EAST"] = {20, 0,  100, 80},
+    ["WEST"] =       {0,  20, 100, 80},
+    ["CENTER"] =     {10, 20, 100, 80},
+    ["EAST"] =       {20, 20, 100, 80},
+    ["SOUTH_WEST"] = {0,  40, 100, 80},
+    ["SOUTH"] =      {10, 40, 100, 80},
+    ["SOUTH_EAST"] = {20, 40, 100, 80},
+    ["STATIC"] =     {10, 30, 100, 80}
+}
+for gravity, expectation in pairs(test_data) do
+    gears.table.merge(steps, {
+        function()
+            set_rules { }
+            -- Wait for the previous cleanup to be done
+            if #client.get() == 0 then
+                return true
+            end
+        end,
+        function(count)
+            if count == 1 then
+                print("testing gravity " .. gravity)
+                test_client(nil,nil,nil,nil,nil,{gravity=lgi.Gdk.Gravity[gravity]})
+            else
+                local c = client.get()[1]
+                if c then
+                    assert(c.size_hints.win_gravity == gravity:lower())
+                    c.border_width = 10
+                    c:titlebar_top(20)
+                    c:geometry({
+                            x = 0,
+                            y = 0,
+                            width = 100,
+                            height = 100
+                    })
+                    windowid = c.window
+                    return true
+                end
+            end
+        end,
+        -- For some reason unmanage needs to happen in a separate step to pass the tests..
+        function()
+            local c = client.get()[1]
+            c:unmanage()
+            awesome.sync()
+            return true
+        end,
+        function()
+            local display = lgi.Gdk.Display.open(os.getenv("DISPLAY"))
+            local window = lgi.GdkX11.X11Window.foreign_new_for_display(display, windowid)
+            local x, y, width, height = window:get_geometry()
+            print(gravity, x, y, width, height)
+            assert(x == expectation[1])
+            assert(y == expectation[2])
+            assert(width == expectation[3])
+            assert(height == expectation[4])
+            window:destroy()
+            return true
+        end,
+        function()
+            local display = lgi.Gdk.Display.open(os.getenv("DISPLAY"))
+            local window = lgi.GdkX11.X11Window.foreign_new_for_display(display, windowid)
+            if window then return end
+            return true
+        end,
+        -- Additionally, checks our expectations are correct by adding decoration back.
+        function(count)
+            if count == 1 then
+                test_client(nil,nil,nil,nil,nil,{gravity=lgi.Gdk.Gravity[gravity]})
+            else
+                local c = client.get()[1]
+                if c then
+                    assert(c.size_hints.win_gravity == gravity:lower())
+                    c.border_width = 0
+                    c:titlebar_top(0)
+                    c:geometry({
+                            x      = expectation[1],
+                            y      = expectation[2],
+                            width  = expectation[3],
+                            height = expectation[4]
+                    })
+                    return true
+                end
+            end
+        end,
+        function()
+            local c = client.get()[1]
+            c:titlebar_top(20)
+            c.border_width = 10
+            return true
+        end,
+        function()
+            local c = client.get()[1]
+            assert(c.border_width      == 10  )
+            assert(c:geometry().x      == 0   )
+            assert(c:geometry().y      == 0   )
+            assert(c:geometry().height == 100 )
+            assert(c:geometry().width  == 100 )
+            c:kill()
+            return true
+        end,
+    })
+end
 
+runner.run_steps(steps)
 -- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/awesome.c
+++ b/awesome.c
@@ -111,8 +111,9 @@ awesome_atexit(bool restart)
     /* Move clients where we want them to be and keep the stacking order intact */
     foreach(c, globalconf.stack)
     {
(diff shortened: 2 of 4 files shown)


psychon:
From a quick look: Seems to be closely related to #2323 which I stopped working on (IIRC) since I couldn't get the tests to work on CI (they passed locally).

xinhaoyuan:
Yes, that is the same issue I was trying to fix. I think my tests somehow works. Feel free to adopt anything to your PR if you think they can be useful.

Aire-One (approved):
I'm glad you submitted this patch! 

It looks even better than the path you linked on reddit as your `client_get_undecorated_geometry` function here also takes into account of the borders.

Thanks :)

(In the test, I can see you use `xwininfo`. I guess it should be added to the README next to `wmctrl` in the dependencies list)

tests/test-geometry.lua
```diff
@@ -257,6 +261,134 @@ local steps = {
...
+                local c = client.get()[1]
+                if c then
+                    assert(c.size_hints.win_gravity == gravity:lower())
+                    assert(c.border_width      == 10  )
+                    assert(c:geometry().x      == 0   )
+                    assert(c:geometry().y      == 0   )
+                    assert(c:geometry().width  == 100 )
+                    assert(c:geometry().height == 100 )
+                    windowid = c.window
+                    c:unmanage()
```
psychon, line 318: Could you add a call to `awesome.sync()` afterwards? That should make sure that the X11 server *really*  already handled the `c:unmanage()` before we get to the next step.

tests/test-geometry.lua
```diff
@@ -257,6 +261,134 @@ local steps = {
...
+            print(gravity, x, y, w, h)
+            assert(x == expectation[1])
+            assert(y == expectation[2])
+            assert(w == expectation[3])
+            assert(h == expectation[4])
+            return true
+        end,
+        function()
+            awful.spawn("wmctrl -c " .. tostring(windowid), false)
+            -- Is there a way to make sure the window is removed?
```
psychon, line 340: What does this do? What do you mean with "make sure the window is removed"?

I *think*  that `wmctrl -c` doesn't do anything. According to https://github.com/dancor/wmctrl/blob/0181e8e4a7ed98255f6ef930875e56560f19aa95/main.c#L738-L740, this sends a `_NET_CLOSE_WINDOW` message to the root window. Such a message is handled by the WM. However, due to your `c:unmanage()` above, AwesomeWM does not know the window any more and I would expect this to do nothing.

psychon (commented):
I'm not sure I understand the new tests.

- It adds a rule that moves the client to (0, 0) with size (100, 100), a border width of 10 and a titlebar of height 20 at the top
- It opens a test client with the gravity under test and tests that the rule was applied, then does `c:unmanage()`
- Afterwards you use `xwininfo` to try to figure out where the window ends up and check that the position is as expected for the gravity.
- Then you open another window, this time with a rule that enforces your "wanted" geometry via a rule and then test that the rule was correctly applied. This seems to be that it only tests `awful.rules` (or whatever it is called now), but does not test anything with gravity handling.
- Here, you also then explicitly test how the geometry changes due to the addition of the titlebar. Is this what you had in mind here? Was this broken before or why is this part of this PR? I'm a bit confused whether the "change itself" should fix this test or whether it is unrelated.

Also: Does this fix #2308?

tests/test-geometry.lua
xinhaoyuan: TIL, added.

tests/test-geometry.lua
xinhaoyuan: That's a really good point. I didn't pay much attention to how the close request is handled. I now uses another method to detect window existence - please let me know.

xinhaoyuan:
Thanks for look into the PR! To answer your questions:

> (In the test, I can see you use xwininfo. I guess it should be added to the README next to wmctrl in the dependencies list)

I figured that I can use lgi instead of external programs, but I know sometimes it can cause issues. Currently the test works fine, but please let me know if there are actually issues with that.

> I'm not sure I understand the new tests.
> 
>     * It adds a rule that moves the client to (0, 0) with size (100, 100), a border width of 10 and a titlebar of height 20 at the top
> 
>     * It opens a test client with the gravity under test and tests that the rule was applied, then does `c:unmanage()`
> 
>     * Afterwards you use `xwininfo` to try to figure out where the window ends up and check that the position is as expected for the gravity.
> 
>     * Then you open another window, this time with a rule that enforces your "wanted" geometry via a rule and then test that the rule was correctly applied. This seems to be that it only tests `awful.rules` (or whatever it is called now), but does not test anything with gravity handling.

Yes, this is just to make sure the rule is applied. This step and the following step is to make sure that the initial expectation is "correct". So the entire steps checks that when a client is unmanaged then managed again, it is first relocated back to the expectation, then moved back to the initial geometry 100x100+0+0.

> 
>     * Here, you also then explicitly test how the geometry changes due to the addition of the titlebar. Is this what you had in mind here? Was this broken before or why is this part of this PR? I'm a bit confused whether the "change itself" should fix this test or whether it is unrelated.
> 
> 
> Also: Does this fix #2308?

I think it does - Telegram uses STATIC gravity, and closing and reopening it with the change does not change it position.

tests/test-geometry.lua
psychon: Since you write "TIL", let's actually explain my thinking here:
- `client_unmanage()` in `objects/client.c` "does some stuff" and sends requests to the X11 server.
- However, libxcb actually has an output buffer and does not send requests unless it actually has to.
- Even when the output buffer is flushed, that only means that the requests were sent, but not that they were handled by the X11 server.
- Thus, in theory it could be possible that  `wmctrl` manages to start and talk to the X11 server before the unmanage is actually done / handled by the X11 server.

The call to `awesome.sync()` sends a request to the X11 server and actually waits for the reply. X11 guarantees that the requests are handled in-order. Thus, by waiting for a reply to a new request, we can be sure that all the previous requests were handled.

psychon:
Thanks for your comment. I think someone should run the test program from #2308 before this is merged. However, I do not have time right now. Sorry.

xinhaoyuan:
I ran the code in #2308 and I think the issue is fixed.

But for some reason the CI checks in this PR seems to be not working..

psychon:
> But for some reason the CI checks in this PR seems to be not working..

Sigh. The Travis web page says:

> Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.

If I understand this correctly, we cannot run builds until next month...?

Elv13:
@xinhaoyuan Hello, I purchased some Travis credits to be able to run builds again. Mind making a random change to this to the CI sees it?

actionless:
@Elv13 re-opening the PR also helps, will do it now

tests/test-geometry.lua
```diff
@@ -257,6 +261,135 @@ local steps = {
     end
 }
 
-runner.run_steps(steps)
+print("Added tests for unmanaged client geometries.")
```
actionless, line 264: i guess this should be a comment, not a print statement

actionless:
@xinhaoyuan test prev commits setup is failing in CI: https://travis-ci.com/github/awesomeWM/awesome/builds/222130810

please squash your commiits

psychon:
I don't think this is `$TEST_PREV_COMMITS`. Luacheck also complains locally:
```
Checking tests/test-geometry.lua                  3 warnings

    tests/test-geometry.lua:327:25: shadowing upvalue 'w' on line 13
    tests/test-geometry.lua:366:27: unused variable 'geo'
    tests/test-geometry.lua:381:19: unused variable 'geo'

```

tests/test-geometry.lua
xinhaoyuan: Done

xinhaoyuan:
I revised the patch with some simplification. I also realized that some default settings in awesomerc.lua are interfering with how new clients are placed/decorated. So I tried to override them in the tests and make them more robust.

actionless:
CI is still failing

xinhaoyuan:
> CI is still failing

Hmm.. I couldn't figure out what's wrong from the CI log. Any clues?

psychon:
> Any clues?

More or less still the same `luacheck` complaint as in https://github.com/awesomeWM/awesome/pull/3253#issuecomment-813360765:
```
Checking tests/test-geometry.lua                  4 warnings

    tests/test-geometry.lua:304:27: unused variable 'geo'
    tests/test-geometry.lua:319:25: shadowing upvalue 'w' on line 13
    tests/test-geometry.lua:341:27: unused variable 'geo'
    tests/test-geometry.lua:363:19: unused variable 'geo'
```
(I have `luacheck` installed and can just run `make luacheck` to have luacheck run on the source code)

xinhaoyuan:
> > Any clues?
> 
> More or less still the same `luacheck` complaint as in [#3253 (comment)](https://github.com/awesomeWM/awesome/pull/3253#issuecomment-813360765):
> 
> ```
> Checking tests/test-geometry.lua                  4 warnings
> 
>     tests/test-geometry.lua:304:27: unused variable 'geo'
>     tests/test-geometry.lua:319:25: shadowing upvalue 'w' on line 13
>     tests/test-geometry.lua:341:27: unused variable 'geo'
>     tests/test-geometry.lua:363:19: unused variable 'geo'
> ```
> 
> (I have `luacheck` installed and can just run `make luacheck` to have luacheck run on the source code)

Ah, nice find! I didn't have luacheck installed. Fixed.

Aire-One:
Is there still any blocker for this to be merged?

For what it's worth, I just tried to use this branch and run the scenario from #2308:

- remove placement from rules
- build the C program with given gcc parameters
- start Xephir and run the C program

My observations were exactly what there is in the "expected" description:

> A 100×100 window should appear, the top left corner of the client area should be placed at (100, 100); the window should flash 5 times, but remain at the same position.

actionless (approved):

Aire-One:
So, I'd like to approve this PR and give the opportunity to mergify to auto-merge it. But @Elv13 self-assigned it. Is there more to do in the scope of this PR?/Is there any remaining blocker?

Elv13:
No, it was just so it was in my queue so I didn't forgot about it

Aire-One (approved):

PR #3254, opened 2021-01-30
Add hotkeys_popup for kitty terminal

samuelabreu:
A few shortcuts from: https://sw.kovidgoyal.net/kitty/#other-keyboard-shortcuts

### DIFF
--- a/lib/awful/hotkeys_popup/keys/init.lua
+++ b/lib/awful/hotkeys_popup/keys/init.lua
@@ -13,6 +13,7 @@ local keys = {
   tmux = require("awful.hotkeys_popup.keys.tmux"),
   qutebrowser = require("awful.hotkeys_popup.keys.qutebrowser"),
   termite = require("awful.hotkeys_popup.keys.termite"),
+  kitty = require("awful.hotkeys_popup.keys.kitty"),
 }
 return keys
 
--- a/lib/awful/hotkeys_popup/keys/kitty.lua
+++ b/lib/awful/hotkeys_popup/keys/kitty.lua
@@ -1,0 +1,49 @@
+---------------------------------------------------------------------------
+--- Kitty hotkeys for awful.hotkeys_widget
+--
+-- @author Samuel Abreu
+-- @copyright 2021 Samuel Abreu
+-- @submodule awful.hotkeys_popup
+---------------------------------------------------------------------------
+local hotkeys_popup = require("awful.hotkeys_popup.widget")
+
+local kitty_rule = { class = { "kitty", "Kitty" } }
+for group_name, group_data in pairs({
+    ["Kitty"] = { color = "#35A91E", rule_any = kitty_rule },
+    ["Kitty: Hints"] = { color = "#35A91E", rule_any = kitty_rule },
+}) do
+    hotkeys_popup.add_group_rules(group_name, group_data)
+end
+
+local kitty_keys = {
+    ["Kitty"] = {{
+        modifiers = { "Ctrl", "Shift" },
+        keys = {
+            Enter = "Split new Terminal",
+            l = "Next layout",
+            t = "New tab",
+            q = "Close tab",
+            Right = "Next tab",
+            Left = "Previous tab",
+            n = "New OS Window",
+            ["]"] = "Next Window",
+            ["["] = "Previous Window",
+            ["0..9"] = "Focus specific window"
+        }
+    }},
+    ["Kitty: Hints"] = {{
+        modifiers = { "Ctrl", "Shift" },
+        keys = {
+            e = "Choose URL to Open",
+        }
+    },
+    {
+        modifiers = { "Ctrl", "Shift", "p" },
+        keys = {
+            f = "Choose path to paste",
+            n = "Choose filename:n to open"
+        }
+    }}
+}
+
+hotkeys_popup.add_hotkeys(kitty_keys)


ShayAgros (approved):
Looks good. Tested it on my setup with kitty

Aire-One (commented):
While I'm happy to see this addition, I'm also afraid we're going to have a lot of unnecessary code (and maybe unmaintained code in the long term). 

Maybe it would be better to have this kind of patches proposed as modules in the awesome recipes page? 

What is your point of view on this? @awesomeWM (I hope it'll tag maintainers)

unai-ndz:
There was some discussion about this [here](https://github.com/awesomeWM/awesome/pull/2472).

To me the recipes page sounds appropriate for this. To your arguments I would add that keybinds are pretty personal, you may want to display only a subset of them, have different ones from default, etc.
And if you need to copy over the hotkey files into your config to make those changes anyway having them in the main repo is not a big advantage.

actionless:
i think for many people the main idea behind it is having help for unknown programs, rather than having help for programs they already know and familiar with

so moving it into recipes will make the feature very unreachable, if everyone around here hates so much the idea of making life of non-manpage-nerd users easier, then all hotkey hints should be at least under all the same repo (even if it not under the same as wm code), not under separate web pages

sclu1034:
I agree with @Aire-One that adding more of these might not be maintainable.

For one, few people will ever have a use for all (or even a majority) of these. Especially when adding different applications of the same "kind" (e.g. terminal emulators), users will only really need one of them.
And yet, as @Elv13 pointed out in the other PR, currently you can only either load all, needlessly increasing start time and memory footprint, or none.
And while new users might be willing to create the initial implementation for a new set of key definitions, many of them will likely not stick around to actively maintain them in case these applications add more hotkeys or change default settings. That job would fall onto the maintainers.

Additionally, the concept of "plugin system + recipes" also changes things in a way that users have immediate access to edit hotkey definitions for whatever recipes they add in their `$HOME/.config`. With that they would be able to customize these definitions based on how they configure their application. Whereas with the current model, the key definitions are hardcoded and inaccessible, and therefore become incorrect as soon as the user customizes their applications.

---

I do understand the idea of how this might help people try/use new applications. But I think that, as it is currently implemented, it doesn't really accomplish that.

It would need to implement at least a dozen or two of the more popular applications for it to be useful to a large amount of people (at which point the previously mentioned performance and maintainability issues increase).
And even then, the user needs to learn of its existence and functionality first, which currently is only done through (1) finding one of the few lines in the default config that reference it and (2) opening it while one of the supported applications is focused to demonstrate that it doesn't include just Awesome's keys.

actionless:
the way how you propose it not much different from "if you want to see some keys just write a text note with them"

it's the feature which make awesome really stand out in compare to all currently existing wms/des (only gtk have something similar, but only or some of the apps, not gnome shell; and unity and chrome os had it for the shell)

but ok, if you dont like my stuff i dont see any reason wasting my time on reviewing yours

unai-ndz:
>the way how you propose it not much different from "if you want to see some keys just write a text note with them"

He is not saying that at all.

With a plugin system you can still have a repo with some default keybindings and you are a `git clone` away to have them all in your config. But this way you can also select the ones you really want and modify the ones that do not fit your needs perfectly.

actionless:
@unai-ndz 
i was proposing that before:

> then all hotkey hints should be at least under all the same repo (even if it not under the same as wm code), not under separate web pages

however since awesome supports luarocks out of the box, it might make sense also to export it there

actionless:
actually, luarocks way would be easier for just default setup, and git clone in case if user wants to make some edits

Elv13:
While it isn't implemented for "all" application, there is some dbus interfaces for generic HUD/GlobalMenubar/Accessibility. I am very aware of the effort it would take to implement those APIs into AwesomeWM (because, well, [I tried and gave up ~10 years ago](https://www.reddit.com/r/awesomewm/comments/erahyo/implementing_global_menu/)). The thing is, if we had *that* implemented, we would get generic hotkey popup, macOS like global menubars and the ability to implement HUD style power user interfaces.

The problem is that that it needs to be implemented in C in a different thread and exposed using unfunny async even loop shenanigans to work properly. I already did a pure Lua implementation (which was based on my wirefu module) and it was too slow for the main thread. C and C++ AwesomeWM plugins are possible/supported, but AFAIK, none was ever written.

So we are kind of stuck here. Adding more files like that doesn't scale, it already shows up on startup time and the older ones are already outdated. Adding an (Ubuntu) Unity style HUD is beyond the amount of time I will probably ever have, but would be a really, really killer features since. There has been a few attempts like https://github.com/jamcnaughton/hud-menu , but so far nobody I am aware of nailed this feature. You need to implement like 5 more APIs than the one in that repo to get "most toolkits" working. A lot of these interfaces are poorly documented and require caching to perform properly.

actionless:
@Elv13 gtk apps already have their own native hotkey help window (Ctrl+Shift+?), i see no urgency duplicating that info in near future

same goes about global menus -- if app already have menus, you can simply see the hotkeys there (or even use smth like `plotinus` instead of hotkeys)

but for terminals, shells or apps with non-standard/minimal UI to see the hotkeys user have to resort to manpage or even developer website

regarding performance i'm not really convinced (and considering what wibox and awful have in init all their submodules and their submodules have in init all their submodules -- adding any submodule or exported field/function there increasing loading time of awesomewm -- let's then remove every new submodule, function or field which were added after api freeze if we do such microoptimizations) -- we have tests for measuring those. if it would be real performance drop instead of figure of speech you would show it there. but after all user can always comment out the line in the config which loading all available hotkeys, the same as they could comment out/remove "awful.autofocus" and any other existing awesomewm features

P.S. if you're still looking for universal appmenu implementation, try https://github.com/rilian-la-te/vala-panel-appmenu, i never tried it myself but got feedback from few of its users what it seems to work everywhere

PR #3257, opened 2021-02-02
fix client:activate action mouse_center

jpkotta:
Warning: I am new to both Lua and Awesome.  

I got errors like `error: /usr/share/awesome/lib/awful/client.lua:1631: attempt to index a nil value (field 'mouse')`.  With this change the errors go away and the behavior seems correct.

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -1531,13 +1531,13 @@ function client.object.activate(c, args)
     elseif new_args.action and new_args.action == "mouse_resize" then
         amousec.resize(c)
     elseif new_args.action and new_args.action == "mouse_center" then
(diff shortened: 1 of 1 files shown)


psychon (approved):

ShayAgros (approved):
For what it's worth, launched Awesome with this commit. Seems ok

Aire-One (approved):
Seems like typos. There is no `mouse` object in this scope. The only mouse object available is the `capi.mouse` declared at the top of the file...

Thanks for spotting the issue and coming here with an actual fix!

Elv13 (approved):

Elv13:
thanks! Sorry for the merge delay.

Issue #3220, opened 2020-12-04
How do i change the clock from 24 hour time to 12 hour time


CardboardTurtle:
I looked in the documentation. it doesn't work. ive looked online. all search results are either about awesome 3.5, or windows 10. how do i change the 24 hour clock into a 12 hour one

info: awesome v4.3 (Too long)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

psychon:
The textclock uses this function from GLib: https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format

The call to `wibox.widget.textclock()` is equivalent to `wibox.widget.textclock(" %a %b %d, %H:%M ")`.

The docs for `g_date_time_format()` say that:
- %H: the hour as a decimal number using a 24-hour clock (range 00 to 23)
- %I: the hour as a decimal number using a 12-hour clock (range 01 to 12)

So I guess you want `wibox.widget.textclock(" %a %b %d, %I:%M ")`

ShayAgros:
@psychon 
The [textclock widget site](https://awesomewm.org/doc/api/classes/wibox.widget.textclock.html) isn't the same as in this repository. A reference to the glib site in the documentation were [merged](https://github.com/awesomeWM/awesome/pull/2926) about a year ago.

How frequently is the site updated from the docs in the repo ?

psychon:
@ShayAgros See https://awesomewm.org/doc/. The link above goes to the docs for the latest stable release. https://awesomewm.org/apidoc/widgets/wibox.widget.textclock.html contains the API docs for git/master.

> How frequently is the site updated from the docs in the repo?

The "stable" docs are updated after every release. Updating that is part of the release procedure. The git/master ones are updated after every commit (automatically via Travis CI).

ShayAgros:
> Edit: I guess this issue should only be closed after the docs were improved.

ok, this is labeled as "good first issue". What is there to do here ?

psychon:
From a quick look: Uhm, nothing. The docs were already improved before this issue was opened, but that did not end up in a release. Sorry for missing that.

Could someone else double-check this, please? Thanks.

Aire-One:
> Could someone else double-check this, please? Thanks.

Yup, the new API documentation has been improved and it's waiting for a new version to be released to the stable documentation. 

[...]

PR #3263, opened 2021-02-07
[doc] Improve wibox.widget.textclock format documentation

Aire-One:
Hello,

I initially started this work in the scope of #3220.

Textclock is an interesting widget as it inherits from another widget: textbox. I think the doc should make it clear, and we can play with the new ldoc tags added by #2924 in this effect.

I'm not happy with how commit 4f64f80 turned out, tho. It involved a big copy-paste of the textbox documentation to `docs/common/wibox_widget_textbox.ldoc`. I'm afraid it will not be maintained when textbox will be updated...

Commit 99866f4 is a bit off-topic but still deeply related to this inheritance subject. I think we can have a new section to present objects hierarchy diagram to the user. It makes the inheritances relations clear, and we can benefit from clarity here.
The current implementation for the diagram relies on HTML nested list. I think we could add more here. Maybe a "known subclasses" list... Anyway, I'm open to feedback on it :) 

object-hierarchy

Funny note: `tests/examples/wibox/widget/defaults/textclock.lua` instantiates a textclock widget to generate its image, so it will display the date and time the doc is built.

### DIFF
--- a/lib/wibox/widget/textclock.lua
+++ b/lib/wibox/widget/textclock.lua
@@ -1,6 +1,42 @@
 ---------------------------------------------------------------------------
 --- Text clock widget.
 --
+-- The `wibox.widget.textclock` widget is part of the Awesome WM's widget
+-- system (see @{03-declarative-layout.md}).
+--
+-- This widget displays a text clock formatted by the
+-- [GLib Date Time format](https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format)
+-- and [GTimeZone](https://developer.gnome.org/glib/stable/glib-GTimeZone.html#g-time-zone-new).
+--
+-- The `wibox.widget.textclock` inherits from `wibox.widget.textbox`. It means
+-- that, once created, the user will receive a derivated instance of
+-- `wibox.widget.textbox` associated with a private `gears.timer` to manage
+-- timed updates of the displayed clock.
+--
+-- Use a `wibox.widget.textclock`
+-- ---
+--
+-- @DOC_wibox_widget_defaults_textclock_EXAMPLE@
+--
+-- Alternatively, you can declare the `textclock` widget using the
+-- declarative pattern (Both codes are strictly equivalent):
+--
+-- @DOC_wibox_widget_declarative-pattern_textclock_EXAMPLE@
+--
+-- The GLib DateTime format
+-- ---
+--
+-- The time displayed by the textclock widget can be formated by the GLib
+-- DateTime format.
+--
+-- Here is a short list with commonly used format specifiers (extracted from
+-- the Glib API references):
+--
+--@DOC_glib_timedate_format_COMMON@
+--
+-- You can read more on the GLib DateTime format in the
+-- [GLib documentation](https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format).
+--
 -- @author Julien Danjou &lt;julien@danjou.info&gt;
 -- @copyright 2009 Julien Danjou
 -- @widgetmod wibox.widget.textclock
--- a/.travis.yml
+++ b/.travis.yml
@@ -239,7 +239,9 @@ script:
         && grep -q 'May not access' lib/beautiful/init.lua
     fi
   - |
-    set -e
+    trap 'echo ERR CODE $? from $BASH_COMMAND : $LINENO' ERR
+    trap 'echo EXIT CODE $? from $BASH_COMMAND : $LINENO' EXIT
+    set -ex
     if [ -n "$BUILD_IN_DIR" ]; then
       # Explicitly remove the Makefile to not build from the src dir accidentally.
       rm Makefile
@@ -289,7 +291,7 @@ script:
     fi
   - |
     if [ "$TEST_PREV_COMMITS" = 1 ] && ! [ "$TRAVIS_PULL_REQUEST" = false ]; then
-      set -e
+      set -ex
       # Check each commit separately (to make git-bisect less annoying).
       # Fix Travis' commit range (https://github.com/travis-ci/travis-ci/issues/4596).
       commit_range="${TRAVIS_COMMIT_RANGE/.../..}"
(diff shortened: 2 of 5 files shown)


lib/wibox/widget/textclock.lua
```diff
@@ -1,6 +1,42 @@
...
+-- The GLib DateTime format
+-- ---
+--
+-- The time displayed by the textclock widget can be formated by the GLib
+-- DateTime format.
+--
+-- Here is a short list with commonly used format specifiers (extracted from
+-- the Glib API references):
+--
+--@DOC_glib_timedate_format_COMMON@
```
ShayAgros, line 37: Need to add space after the double dash. Otherwise the link below (g-date-time-format) isn't rendered properly

Elv13:
Sorry for the delay. Awesome has been rather low on my COVID-era priority list. Do you think this is ready to merge once @ShayAgros comments are addressed?

Also, do you think we could auto-generate this using  a `@something` ldoc tag in the header rather than using the examples framework to render the markdown? This would be easier to apply to the entire doc rather than having to make an example per class. However it makes the already gigantic `config.ld` and html template larger. Still, I think it would be cleaner and be rather easy to convert your current code into config.ld/template

ShayAgros:
When running `make` and opening `build/doc/widgets/wibox.widget.textbox.html` I see the addition of the examples there (nicely done btw), but don't see the inheritance diagram you showed at the pr header. Is it expected ? Tried to dig into it myself, but haven't pinned down the reason yet.

lib/wibox/widget/textclock.lua
Aire-One: Nope, with the additional space, the generated documentation will have the double dash drawn before the table. If you check how other files from `docs/common/*.ldoc` are included, they do the same.

It's a tricky hack to integrate the file content without making it a "comment of a comment" (output `----` vs `-- --`). ldoc will interpret `----` as a regular comment opening while `-- --` will be a comment followed by a double dash (to be rendered).

You can check that after running `make ldoc` by inspecting the `build/lib/wibox/widget/textclock.lua` file and checking the generated code that ldoc will parse to generate the API references.

lib/wibox/widget/textclock.lua
```diff
@@ -1,6 +1,42 @@
 ---------------------------------------------------------------------------
 --- Text clock widget.
 --
+-- The `wibox.widget.textclock` widget is part of the Awesome WM's wiboxes
```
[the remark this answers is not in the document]
Aire-One: The full sentence says `Awesome WM's wiboxes system` with a link to the documentation at https://awesomewm.org/apidoc/documentation/03-declarative-layout.md.html.

Still, "wiboxes system" is probably an unnecessarily complicated way to says it's a widget from Awesome library...

I'll go with `Awesome WM's widgets system` for now. The objective of this sentence was to help visualize where the class is on the internal API conception. I think this could be removed once I'll have finalized the "inheritance diagram" stuff. I guess a good diagram will be better than a strange sentence...

Aire-One:
I have split this PR to make the "improve textclock doc" a first ready to merge work. Thank you, @ShayAgros, for your review!

I'll work on a new branch to make progress on the inheritance diagram.

(The work I previously submitted is still available in a backup branch at https://github.com/Aire-One/awesome/tree/doc_textclock_backup)

> Also, do you think we could auto-generate this using a @something ldoc tag in the header rather than using the examples framework to render the markdown? This would be easier to apply to the entire doc rather than having to make an example per class. However it makes the already gigantic config.ld and html template larger. Still, I think it would be cleaner and be rather easy to convert your current code into config.ld/template

Using a ldoc tag is definitely something I'd like to explore. I was not sure if it was the right way to generate html content at the "class/module" level... Anyway, I'll try to see how far we can go with ldoc tags, maybe I see too far right now, but I hope we can use ldoc to automatically generate an inheritance tree we can use to render the inheritance diagram and even include inherited properties!

lib/wibox/widget/textclock.lua
ShayAgros: Thanks for this detail explanation, I wasn't aware of `build/lib/wibox/widget/textclock.lua` file (this simplifies my code investigation quite a lot).  Running `make ldoc` on the following snippet:
```
----<table class='widget_list' border=1>
-- </table>
--
-- You can read more on the GLib DateTime format at
-- [https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format](https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format).
```

Causes the 'html' output `build/doc/widgets/wibox.widget.textclock.html` to show the `https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format` part as simple text rather than a link. Looking at the html of this file I see:
```
 You can read more on the GLib DateTime format at
 [https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format](https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format).
```
(so `ldoc` didn't render the markdown at all)

Adding space to the `<table>` line:
```
---- <table class='widget_list' border=1>
```

make the link being rendered correctly. I don't know how to solve this problem atm (I did miss the extra dashes added by my proposed solution), but I'd look into it.

Is the link, in this commit, rendered correctly on your setup ? (maybe I have a version issue, as I often have with Ubutnu)

Aire-One:
Fixes #3220

Aire-One:
Closes #3220

Elv13 (dismissed):

actionless (dismissed):

Aire-One:
(close/reopen to trigger Travis. We already have 2 approvals, so I hope PR will be auto merged when CI will be completed)

Aire-One:
> I hope PR will be auto merged when CI will be completed

Unless Travis failed... It seems to 5th job was "canceled". I'm not sure what happened.

Elv13:
> Unless Travis failed... It seems to 5th job was "canceled". I'm not sure what happened.

I restarted it a few time, it doesn't want to finish. I will need a debug this, so I will keep this unmerged so I can reproduce. What happens is that `set -x` seems to kill the script for unknown reason. Something is returning a non-zero ret code while `luacheck` is printing stuff. The logs doesn't make it obvious about *what* is causing this. It's rather similar to the bug that prevented the API doc to be generated 95% of the time in the last couple weeks. If I restart it enough time, it will eventually pass, but I would like to know why it fails.

Elv13:
Looks like there's 2 issues. One seems to be a Travis-ci regression where the log is just cropped. I added a couple `trap` and they are not visibly triggerred. Plus the fact that the output cuts mid-line is also highly suspicious.

But beside, the problem was that one of the URL in the doc tripped the line length limit. So that's what "really" made the build fail, despite Travis not giving any indication of that being the case. I rebased the PR with a fix.

Elv13 (approved):

actionless (approved):

actionless:
im also fine leaving debug stuff in travis, as in other PR he gonna be removed anyway

PR #3272, opened 2021-02-23
Added NixOS instalation guide.

branwright1:
Hello, I just wanted to add instalation guide from master branch for nixos users, it's quite popular distrubution along arch and debian. Instalation from source has different steps as you can see, you can't just use make and make install because packages are stored in `/nix/store` and there is no `/usr/bin`.

### DIFF
--- a/docs/01-readme.md
+++ b/docs/01-readme.md
@@ -52,26 +52,30 @@ make package
 sudo apt install *.deb
 ```
 
-#### NixOS (overlay in configuration.nix)
+#### NixOS (put this in configuration.nix)
+
 ```nix
+
 # Overlay bellow will build AwesomeWM from selected commit
 # NOTE: Manually change revision (commit numbers) on every update.
 
-nixpkgs.overlays = [ 
-(self: super: {
-  awesome = super.awesome.overrideAttrs (old: {
-    src = super.fetchFromGitHub {
-      owner = "awesomewm";
-      repo = "awesome";
-      rev = "4319b161103a81403ba3516924e86e13ff33c036";
-      sha256 = "1p7c3cmdy0iyk28fgc579rkzfhnwn1i1b541rs2nw80h1xzs1g80";
+  nixpkgs.overlays = [
+    (self: super: {
+      awesome = super.awesome.overrideAttrs (old: {
+        src = super.fetchFromGitHub {
+          owner = "awesomewm";
+          repo = "awesome";
+          rev = "4319b161103a81403ba3516924e86e13ff33c036";
+          sha256 = "1p7c3cmdy0iyk28fgc579rkzfhnwn1i1b541rs2nw80h1xzs1g80";
         };
-    });
-  })
-];
+      });
+    })
+  ];
 
 services.xserver.windowManager.awesome.enable = true;
+
 ```
+
 ### Build dependencies
 
 Awesome has the following dependencies (besides a more-or-less standard POSIX
--- a/docs/01-readme.md
+++ b/docs/01-readme.md
@@ -52,7 +52,26 @@ make package
 sudo apt install *.deb
 ```
 
+#### NixOS (overlay in configuration.nix)
+```nix
+# Overlay bellow will build AwesomeWM from selected commit
+# NOTE: Manually change revision (commit numbers) on every update.
 
+nixpkgs.overlays = [ 
+(self: super: {
+  awesome = super.awesome.overrideAttrs (old: {
+    src = super.fetchFromGitHub {
+      owner = "awesomewm";
+      repo = "awesome";
+      rev = "4319b161103a81403ba3516924e86e13ff33c036";
+      sha256 = "1p7c3cmdy0iyk28fgc579rkzfhnwn1i1b541rs2nw80h1xzs1g80";
+        };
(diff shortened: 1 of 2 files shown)


docs/01-readme.md
```diff
@@ -52,7 +52,26 @@ make package
...
+# Overlay bellow will build AwesomeWM from selected commit
+# NOTE: Manually change revision (commit numbers) on every update.
+
+nixpkgs.overlays = [ 
+(self: super: {
+  awesome = super.awesome.overrideAttrs (old: {
+    src = super.fetchFromGitHub {
+      owner = "awesomewm";
+      repo = "awesome";
+      rev = "4319b161103a81403ba3516924e86e13ff33c036";
```
ShayAgros, line 66: This is interesting (quite different from what ever saw). Cannot the 'rev' attribute be something like tag name or branch name ? This seems like you're pointing to current 'master'

I'm asking because I'm guessing Nix uses `git` to download the repo. `git` does support fetching a branch

docs/01-readme.md
```diff
@@ -52,7 +52,26 @@ make package
 sudo apt install *.deb
 ```
 
-
+#### NixOS (overlay in configuration.nix)
+```nix
+# Overlay bellow will build AwesomeWM from selected commit
+# NOTE: Manually change revision (commit numbers) on every update.
+
+nixpkgs.overlays = [
```
ShayAgros, line 60: Can you please remove the trailing space at the end of line 60 ?

docs/01-readme.md
```diff
@@ -52,7 +52,26 @@ make package
 sudo apt install *.deb
 ```
 
-
+#### NixOS (overlay in configuration.nix)
+```nix
```
ShayAgros, line 56: Just out of curiosity, does your markdown formatter have css class for 'nix' typed code ?

docs/01-readme.md
```diff
@@ -52,7 +52,26 @@ make package
...
+      repo = "awesome";
+      rev = "4319b161103a81403ba3516924e86e13ff33c036";
+      sha256 = "1p7c3cmdy0iyk28fgc579rkzfhnwn1i1b541rs2nw80h1xzs1g80";
+        };
+    });
+  })
+];
+
+services.xserver.windowManager.awesome.enable = true;
+```
```
ShayAgros, line 74: nit, can you leave en empty line between lines 74 and 75. Doesn't affect the markdown compilation, but having this line would make the code more readable (imho) in raw text format

branwright1:
Stable version is packaged already, this code override source of [default.nix](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/x11/window-managers/awesome.nix) for Awesome from nixpkgs repo and changes branch to master then build latest selected commit also I'm sorry for those spaces etc. but I was writting it in vim it gives a lot of unwanted spaces...

docs/01-readme.md
branwright1: I tried without "revision" but it gives mi error about missing "rev" function and cannot build it. ``error: anonymous function at /nix/store/7579j58gq0v5jwm3d25b9wfz2nwbc3a6-nixos-21.05pre271938.c7d0dbe094c/nixos/pkgs/build-support/fetchgithub/default.nix:3:1 called without required argument 'rev', at /etc/nixos/modules/display/X11.nix:10:11``

docs/01-readme.md
branwright1: As you can see, text on github has syntax highlightning
image

docs/01-readme.md
ShayAgros: ok, seems like you can pass it a tag name according to this [site](https://rawkode.com/articles/contributing-nixpkgs/), not really bothers me

ShayAgros (approved):

Aire-One:
I'm not familiar with NixOS, so I'm probably wrong.

I don't think it feet a README. It looks like it needs to be maintained up to date at every new commit because of the `rev` and `sha` being hard coded. According to my understanding of NixOS (which is limited), it seems like NixOS want you to use a stable point (that have been tested) to build your package. So following a git master is probably not intended.

Maybe NixOS has something similar to the AUR, where this script can feet. We can then add a link to this in the readme (kinda like what we currently have for Arch git install).

branwright1:
There is NUR and Cachix (like `cachix use awesome`) cachix needs github actions or other CI which will build binary cache adn will he able to downllad it through adding something like awesome-git as packache if user has added cachix repo which is a lot easier and popular compared to NUR (nur is complicated and almost dead because official repo has over 80k pkga). Here is cachix website https://cachix.org/ . People are using it to e.g maintain neovim-nightly, missing wayland compositor and packages etc.

Elv13:
Hello,

Thanks for this. I will echo @Aire-One comment. Can you make a new `xx-installation-guide.md` file, put a NixOS section there and link it from the readme.md file? This might take a bit too much space for the main README given the Nix relative market share.

branwright1:
Hello, I'm just decided to make pull request as "awesome-git" in nixpkgs to get it in official unstable channel without bothering with overlays and fetching hashes on every update. Thanks for you all for attention, have a nice day :)

PR #3282, opened 2021-03-08
Delete useless symlink & add to readme. Update docs readme to match. Also, you should update license to v3.

HolySuicidalTurkey:
I added some content to your debug & troubleshooting section.

It's moderately shitty that I had to find the solution to a problem with your software **on somebody else's [repository](https://github.com/elenapan/dotfiles/wiki/Troubleshooting).**
btw, thank you @elenapan

It took me weeks to figure this out. Please save future users from this suffering.


README.md
```diff
@@ -1 +1,203 @@
-docs/01-readme.md
```
sclu1034, line 1: The symlink is not useless. By linking, rather than copying the file, we don't have to make identical changes in multiple files.

In fact, your second commit here is a good example for that.

sclu1034 (changes requested):

README.md
HolySuicidalTurkey: I mean, then why do we have a copy of the readme in the docs folder anyways? Doesn't seem to contribute much.

sclu1034: The file at `docs/01-readme.md` exists, so that [this page](https://awesomewm.org/apidoc/documentation/01-readme.md.html) exists in the official docs.
The link at `README.md` exists to satisfy GitHub's convention where the repo's introductory reading should be.

And since we don't want to have two copies that we would need to update individually, we link one to the other.

sclu1034 (dismissed):

HolySuicidalTurkey (commented):
i have no idea wut i did, imma just click 'submit review'

Aire-One:
I don't think this is the right place for it. This is not the responsibility of Awesome WM to tell you how to configure your display manager. What about people who use startx or xinit?

I also quickly checked the repo, awesome has a .desktop file (https://www.github.com/awesomeWM/awesome/tree/master/awesome.desktop) and the `make install` rule does install properly this file (https://www.github.com/awesomeWM/awesome/tree/master/CMakeLists.txt#L393)

Elv13:
@HolySuicidalTurkey Hi, can you move the xsession part to the FAQ as the only change? I think it is useful and we had this question before. I also agree with @aire-one that we should avoid making the readme too long.

As for changing the license, no, I don't plan on doing that.

docs/01-readme.md
```diff
@@ -141,6 +141,20 @@ If Awesome does not start or the configuration file is not producing the
 desired results the user should examine this file to gain insight into the
 problem.
 
+# Awesome not an option in your desktop selection? Try this.
+In order for your login screen or, to be more precise, your display manager, to display this button, it needs to find an awesome.desktop file inside /usr/share/xsessions/. When installing AwesomeWM from source, this file is not installed automatically.
+
+Run the following command to create the file:
+
+`echo '[Desktop Entry]
```
actionless, line 149: this is not correct markdown syntax for multiline codeblocks

README.md
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+docs/01-readme.md
```
actionless, line 1: please rebase out this change at all

actionless (changes requested):

docs/01-readme.md
actionless: (you current version will eat up the linebreaks after rendering)

README.md
HolySuicidalTurkey: deleted

docs/01-readme.md
HolySuicidalTurkey: fixed, thank you for catching this mistake.


"
### Awesome not an option in your desktop selection? Try this.

In order for your login screen or, to be more precise, your display manager, to display this button, it needs to find an awesome.desktop file inside /usr/share/xsessions/. When installing AwesomeWM from source, this file is not installed automatically.

Run the following command to create the file:

```
echo '[Desktop Entry]
Name=awesome
Comment=Highly configurable framework window manager
TryExec=awesome
Exec=awesome
Type=Application
' | sudo tee /usr/share/xsessions/awesome.desktop
```
"

README.md
actionless: i mean change rebase out the change to the symlink -- leave it without changes

HolySuicidalTurkey: it still has no effective difference, and I can't figure out how to "rebase" it.

sclu1034 (changes requested):
`README.md` is now a text file rather than a symlink. You'll have to check out your branch locally and create the link with `ln -s`.

Or, as @actionless said, [rewrite history](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) to drop all commits that changed `README.md`. See the section "Deleting a commit" for that.

And I agree with @Elv13's suggestion to move the added section to the FAQ. A situation where Awesome is not installed through a package manager or `make install` is rather uncommon. It's great to provide help for these edge cases, but it doesn't need to sit on the front page.

So, in conclusion, my suggestions are:

- Move the new troubleshooting section to "Common Issues" in `docs/90-FAQ.md`
- Drop all commits (see link above) related to `README.md` and `docs/01-readme.md`

sclu1034:
This has now been implemented in #3347.
See <https://awesomewm.org/apidoc/documentation/90-FAQ.md.html>

Aire-One:
> This has now been implemented in #3347.
> See https://awesomewm.org/apidoc/documentation/90-FAQ.md.html, section `Awesome doesn’t show up on my login screen`.

Thank you, @sclu1034, I'm closing this PR.

@HolySuicidalTurkey, feel free to reopen/comment if you think there is something still missing. As always, any contribution is welcomed :)

PR #3284, opened 2021-03-11
Fix composite widgets with top level container

### LINKED ISSUE #3213: Can't use a complex/composed custom widget in declarative style

```
awesome v4.3-895-g538586c17-dirty (Too long)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

1. Create file `my_widget.lua`

```lua
local wibox  = require('wibox')
local gtable = require('gears.table')

local my_widget = {}

function my_widget:set_image(img)
	self:get_children_by_id("img")[1]:set_image(img)
end

local function new(img)
	local ret = wibox.widget.base.make_widget_declarative{
		{
			image = img,
		},
		margins = { 1, 1, 1, 1 },
		widget = wibox.container.margin,
	}

	gtable.crush(ret, my_widget, true)

	return ret
end

return setmetatable(my_widget, { __call = function(_, ...) return new(...) end })
```

2. Create a `wibox`:

```lua
local my_widget = require('my_widget')
wibox({
	visible = true,
    x = 0,
	y = 0,
	width = 100,
	height = 100,
	bg = "#000000",
	widget = {
		image = "some/icon.png",
		widget = my_widget,
	},
})
```

_Edit: The snippets above are incomplete, here is a full test case: [Gist](https://gist.github.com/sclu1034/e6859b30b627822d60871e10fc33e129)._

**Actual result:**

An empty `wibox`. The custom widget is not drawn.

**Expected result:**

My icon filling the `wibox`. Or an error message stating that I can't use `my_widget` in this way.

It seems that it is not possible to use reusable, complex/composed widgets in declarative style.
If I change the widget definition in the `wibox` to `my_widget("some/icon.png")`, the icon will appear.
It also appears if I remove the wrapping `wibox.container.margin` from `my_widget`, making this just a thin wrapper around a single `wibox.widget.imagebox`.

Similarly, if `my_widget` is built in imperative style:

- Works: `ret = wibox.widget.imagebox(img)`
- Doesn't work: `ret = wibox.container.margin(wibox.widget.imagebox(img))`

It would at least expect to see an error that this specific use case is not supported, rather than just an empty widget. However, it would be quite useful, especially for libraries I presume, to be able to create reusable widgets (e.g. a button that combines `background`, `margin` & `image`) and users of that library can use it in whatever style they prefer.

sclu1034:
When wrapping container widgets to create reusable composite widgets, `drill` will be called twice on the same widget definition. The first call happens within the wrapping function and applies the children widgets fine. The second call happens when the composite widget is used, but since there are no children widgets defined, the call to `set_children` sets the existing child to `nil` instead.

This change will prevent the second call from overriding the children created by the first call.

Fixes #3213.

### DIFF
--- a/lib/wibox/widget/base.lua
+++ b/lib/wibox/widget/base.lua
@@ -533,32 +533,34 @@ local function drill(ids, content)
         end
     end
 
-    -- Add all widgets.
-    for k = 1, max do
-        -- ipairs cannot be used on sparse tables.
-        local v, id2, e = widgets[k], id, nil
-        if v then
-            -- It is another declarative container, parse it.
-            if (not v.is_widget) and (v.widget or v.layout) then
-                e, id2 = drill(ids, v)
-                widgets[k] = e
-            elseif (not v.is_widget) and is_callable(v) then
-                 widgets[k] = v()
-            end
-            base.check_widget(widgets[k])
+    if widgets and max > 0 then
+        -- Add all widgets.
+        for k = 1, max do
+            -- ipairs cannot be used on sparse tables.
+            local v, id2, e = widgets[k], id, nil
+            if v then
+                -- It is another declarative container, parse it.
+                if (not v.is_widget) and (v.widget or v.layout) then
+                    e, id2 = drill(ids, v)
+                    widgets[k] = e
+                elseif (not v.is_widget) and is_callable(v) then
+                     widgets[k] = v()
+                end
+                base.check_widget(widgets[k])
 
-            -- Place the widget in the access table.
-            if id2 then
-                l  [id2] = e
-                ids[id2] = ids[id2] or {}
-                table.insert(ids[id2], e)
+                -- Place the widget in the access table.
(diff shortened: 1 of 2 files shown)


Elv13:
Hello, thanks for this.

Can you make random change and update the PR? At the time we had ran out of free CI credits and your PR isn't in the list. I cannot add it and I cannot push to your branch (permission denied because you didn't check the checkbox when creating the PR). I purchased some CI credits so Travis should run next time. If it passes, then we can merge this.

sclu1034:
Touched the last commit (no code change).

> permission denied because you didn't check the checkbox when creating the PR

Weird, GitHub tells me that maintainers should have access here.

Elv13 (approved):

Aire-One (approved):

PR #3288, opened 2021-03-15
[doc] Add ldoc @supermodule tag to draw inheritance diagram and manage inherited properties

Aire-One:
This is a new implementation of the "let's draw a hierarchy diagram" thing I started with #3263. For this implementation, I used, as @Elv13 suggested, a new ldoc tag.

This adds a new ldoc tag: `@supermodule`. We can use it to indicate the module's supermodule. The logic to build the hierarchy tree and find supermodules contents is implemented in the template itself. I couldn't find a better solution and this is what was used in this other project (https://gricad-gitlab.univ-grenoble-alpes.fr/brenona/arcades/-/blob/master/assets/ldoc/ldoc.ltp#L24-56) I found while digging ldoc (https://github.com/lunarmodules/LDoc/pull/293)...

Anyway, the last commit (4fee67d) is actually a POC, so you can easily see what this PR is about (to compare with https://awesomewm.org/apidoc/widgets/wibox.widget.textclock.html):

https://imgur.com/a/EBgVpuO
https://user-images.githubusercontent.com/6602958/111206863-3d66cc80-85c9-11eb-83ac-995f006b1951.png
(links as it was really long and kinda spam the page)

This isn't perfect and still needs some adjustments. The goal would be to completely remove the use of `docs/common/object.ldoc` and `docs/common/widget.ldoc` as the content of these file should be automatically added by inheritance.

Looking at the length of the final render, I think we probably shouldn't include all the descriptions and contents from supermodules. Maybe we could try another approach similar to what the Android API references does (for example: https://developer.android.com/reference/android/widget/ImageView, see the "Inherited *" sections). I'm open to feedbacks and suggestions :smile: 

Also, I'd like some inputs on how we should show the root classes (`gears.object` vs `wibox.widget` vs `wibox.widget.base`). 
The `wibox.widget.base.make_widget()` function (used as a base for widgets) returns an instance of `gears.object` monkeypatched with the internal `wibox.widget.base.widget` properties and additional properties that we once called `wibox.widget` for `docs/common/widget.ldoc`...

### DIFF
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -456,6 +456,16 @@ add_custom_tag {
     },
 }
 
+-- Define the supermodule class.
+-- This tag should be used at the module level. All properties from the
+-- supermodule will be recursively added to the module by our ldoc template.
+-- @supermodule supermodule
+add_custom_tag {
+    name = "supermodule",
+    hiden = true,
+    auto_subtags = false
+}
+
 -- More fitting section names
 kind_names={topic='Documentation', module='Libraries', script='Sample files'}
 
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -462,7 +462,16 @@ add_custom_tag {
 -- @supermodule supermodule
 add_custom_tag {
     name = "supermodule",
-    hiden = true,
+    hidden = true,
+    auto_subtags = false
+}
+
+-- Mark the item ad hidden.
+-- This tag should be used to hide items from the documentation.
+-- @hidden
+add_custom_tag {
+    name = "hidden",
+    hidden = true,
     auto_subtags = false
 }
 
--- a/docs/ldoc.css
+++ b/docs/ldoc.css
@@ -503,3 +503,18 @@ pre .url { color: #272fc2; text-decoration: underline; }
 .index_guides div a:hover {
     background-color: #99b3ec;
 }
+
+/* Inheritance diagram */
+.inheritance .inheritance__level {
+    list-style: none;
+}
+
+.inheritance .inheritance__level__node::before {
+    content: "↳";
+}
+
+.inheritance .inheritance__level__node--root::before {
+    /* simulate the spacing of the arrow character */
+    content: "   ";
+    white-space: pre;
+}
--- a/docs/ldoc.css
+++ b/docs/ldoc.css
@@ -509,6 +509,10 @@ pre .url { color: #272fc2; text-decoration: underline; }
     list-style: none;
 }
 
+.inheritance .inheritance__level--root {
+    padding-left: 0;
+}
+
 .inheritance .inheritance__level__node::before {
     content: "↳";
 }
@@ -517,4 +521,14 @@ pre .url { color: #272fc2; text-decoration: underline; }
     /* simulate the spacing of the arrow character */
     content: "   ";
     white-space: pre;
+}
+
+.extra-header {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+}
+
+.extra-header__section {
+    flex-grow: 1;
 }
--- a/docs/ldoc.ltp
+++ b/docs/ldoc.ltp
@@ -259,6 +259,7 @@
 #  end
 #  for item in iter(k.items) do
 #   local dn = display_name(item)
+#   local inherited = item.inherited or (item.baseclass ~= module.name)
 #   if item.sanitize_type then item.sanitize_type(item, ldoc) end
     <tr>
 #   if item.display_type and not item.compact_signature then
@@ -272,8 +273,8 @@
 #     end
       </td>
 #   end
-      <td colspan="$(item.inherited and 1 or 2)" class="summary">$(M(item.summary,item))</td>
-#   if item.inherited then
+      <td colspan="$(inherited and 1 or 2)" class="summary">$(M(item.summary,item))</td>
+#   if inherited then
       <td class="baseclass" $(nowrap)>
         Inherited from $(item.baseclass)
       </td>
@@ -332,8 +333,8 @@
     <span class="proptype">$(item.display_type)</span>
 #   end
     <span class="baseclass" $(nowrap)>
-#   if item.inherited then
-    &middot;&nbsp;Inherited from $(item.baseclass)
+#   if item.inherited or (item.baseclass ~= module.name) then
+    &middot;&nbsp;Inherited from $(M(item.baseclass, item))
 #   end
 #   if item.extra_summary then
 #       for _, col in ldoc.ipairs(item.extra_summary) do
--- a/docs/ldoc.ltp
+++ b/docs/ldoc.ltp
@@ -184,17 +184,24 @@
 <h1>Module: <code>$(module.name)</code></h1>
 <p>$(M(module.summary,module))</p>
 <p>$(M(module.description,module))</p>
-
+#   if module.usage then
+#     local li,il = use_li(module.usage)
+    <h3>Usage:</h3>
+    <ul>
+#     for usage in iter(module.usage) do
+        $(li)<pre class="example">$(ldoc.escape(usage))</pre>$(il)
+#     end -- for
+    </ul>
+#   end -- if usage
+<div class="extra-header">
 # if module.tags.supermodule or module.tags.knownusage then
+  <div class="extra-header__section">
     <h3>Class Hierarchy</h3>
     <div class="inheritance">
 # local function draw_hierary_recursifly(i)
-    <ul class="inheritance__level">
-#   if i == #hierarchy then
-        <li class="inheritance__level__node inheritance__level__node inheritance__level__node--root">
-#   else
-        <li class="inheritance__level__node inheritance__level__node">
-#   end
+#   local is_root_level = (i == #hierarchy)
+    <ul class="inheritance__level $(is_root_level and 'inheritance__level--root' or '')">
+        <li class="inheritance__level__node $(is_root_level and 'inheritance__level__node--root' or '')">
 #   local mod = hierarchy[i]
 #   local name = display_name(hierarchy[i])
 #   if mod == module then
@@ -209,7 +216,7 @@
         </li>
 #   end
     </ul>
-# end
+# end -- module.tags.supermodule
 # draw_hierary_recursifly(#hierarchy)
     </div>
 # end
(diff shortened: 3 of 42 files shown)


Elv13:
Nice work!

I think we should eventually add a commit to remove the "useless" methods from `widget.widget.base`. Right now we have both the `visible` property and the `get_visible`/`set_visible` methods. The methods are redundant since the `gears.object` magic turns them into properties.

Beside, I think we can merge this as-is and fix this in other PRs? I am trying to finish my old (over 1 years old...) PR to upgrade the landing page. With both of these merged (and the fix to have fewer methods), I am tempted to get 4.4 out of the door.

Aire-One:
Oops... Sorry @Elv13, I didn't realized you was asking for this PR to be ready ASAP because of the Travis issue and the migration to GitHub Actions can virtually block releases while we're working on it. Hence, we should do a release while we still have Travis credits. 

I don't think it would be nice to have this PR on the release without the additional transition work to complete what I started with 4fee67d253532c12d2da5dc9c6c4c7728f229a07. 

I also need to fix how I handled inheritance with `wibox.widget` and `wibox.widget.base`. And my "ldoc will automatically find inherited properties/methods" method also imply we do some clean up in the documentation comments to have a good result. As you mentioned, some methods are duplicated... 

This shouldn't be too much work, and I'm slowly working on it. But I don't think I'll be able to deliver a proper update before Sunday. 

So anyway, if we need to make a release now, I think we should skip this PR. If we can wait about 1 more week, I'll do my best to deliver!

Elv13:
> So anyway, if we need to make a release now, I think we should skip this PR. If we can wait about 1 more week, I'll do my best to deliver!

It isn't super urgent, I aim for May-ish. There is a lot of open bugs right now and some of them are totally worth fixing given how often we release stuff

Aire-One:
I have reworked my code and pushed the usage of `@supermodule` in most modules.

Looking my "`gears.object` vs `wibox.widget` vs `wibox.widget.base`" issue, I took the initiative to say I was wrong in commit 589908adeff796258196bb19754fc8a52201c276 and there is no such `wibox.widget` things. Widgets that are based on constructors from `wibox.widget.base` (e.g. `wibox.widget.base.make_widget()`) inherits from `wibox.widget.base` (and `wibox.widget.base` inherits from `gears.object`). This is what's the truest if you take a closer look at the code.

`wibox.layout` was another story, tho!
I skipped to work on the `@DOC_fixed_COMMON@` and `@interface` documentation. The current render we have is good enough I think. Some items are marked as inherited because parents already implement them and some others are marked as native from the class because parents didn't implement them... Maybe a future PR will take care of this interface stuff...

> I think we should eventually add a commit to remove the "useless" methods from widget.widget.base. Right now we have both the visible property and the get_visible/set_visible methods. The methods are redundant since the gears.object magic turns them into properties.

I added an ldoc tag `@hiden`. We can use it to dismiss items from being rendered in the documentation. I wasn't happy to destroy methods documentation even if you want to "force" the usage of `gears.object` magic. Maybe these `@hiden` items could be part of an "extendable" section called "See More stuff".

Anyway, I think I reached a point where this PR gives a good/reliable render and is used enough in the documentation to make sense.
Please pull the branch and see for yourself the documentation it generates. As always, I'm open to any feedbacks :smile:

docs/config.ld
```diff
@@ -456,6 +456,25 @@ add_custom_tag {
...
+    hiden = true,
+    auto_subtags = false
+}
+
+-- Mark the item ad hiden.
+-- This tag should be used to hide items from the documentation.
+-- @hiden
+add_custom_tag {
+    name = "hiden",
+    hiden = true,
```
sclu1034, line 474: This should be "hide" or "hidden". "hiden" doesn't exist.

sclu1034 (changes requested):

docs/config.ld
Aire-One: Oops... And I did this typo everywhere!

Elv13:
I tried this and it works great! I made a commit with a few more `@supoermodule` for the wibox based modules.

I also added another commit to ask for your opinion. Feel free to delete it the next time you rebase this is you don't like it. What about laying the Hierarchy/Info/See section horizontally?

image

Individually, they are not very wide. So I think it makes the page a little less intimidating? I am not 100% happy with how it looks, so alternatives are welcome.

Otherwise, everything seems great. If you want to merge this PR as it is and iterate to add more awesomeness, I am totally open to that.

sclu1034:
> Individually, they are not very wide. So I think it makes the page a little less intimidating? I am not 100% happy with how it looks, so alternatives are welcome.

I quite like this one. Definitely less cluttered than the previous layout.
The only thing that bugs me about it, is the fact that each section has a different left margin/padding, but that should be easy enough to adjust.

---

One general issue I see is the fact that sometimes items are included as inherited when they aren't actually used.
For example, `layout.flex` and `layout.ratio` inherit `layout.fixed.fill_space`. This property is used in `layout.fixed:layout`, but since both `layout.flex` and `layout.ratio` implement their own `:layout`, they do not use that property (and set it to `nil` in their constructors).
So the fact that `fill_space` shows up in the docs for `layout.flex` and `layout.ratio` is misleading.

Aire-One:
(@Elv13 I'm not sure why I had to rebase the whole branch to correctly merge your 2 commits. Did you apply modification on others commits? I can't see differences...)

>> Individually, they are not very wide. So I think it makes the page a little less intimidating? I am not 100% happy with how it looks, so alternatives are welcome.
>
> I quite like this one. Definitely less cluttered than the previous layout.
The only thing that bugs me about it, is the fact that each section has a different left margin/padding, but that should be easy enough to adjust.

I actually like it too!
This margin/padding issue will definitely not be hard to fix, but the hierarchy diagram will most probably require something similar to `margin: -14px` and I'm a little lazy to try that now...
Also, I'm concerned about how this 3 columns design will scale on "less wide browsing". I usually have the doc tiled on the left side of my screen. Here is how it looks currently for half my screen (1920x1080, so about 960px wide)

image

> One general issue I see is the fact that sometimes items are included as inherited when they aren't actually used.
For example, layout.flex and layout.ratio inherit layout.fixed.fill_space. This property is used in layout.fixed:layout, but since both layout.flex and layout.ratio implement their own :layout, they do not use that property (and set it to nil in their constructors).
So the fact that fill_space shows up in the docs for layout.flex and layout.ratio is misleading.

I think we can fix that by overriding the property with a `@property` block and hide it with `@hidden`.

Aire-One:
>> One general issue I see is the fact that sometimes items are included as inherited when they aren't actually used.
>> For example, layout.flex and layout.ratio inherit layout.fixed.fill_space. This property is used in layout.fixed:layout, but since both layout.flex and layout.ratio implement their own :layout, they do not use that property (and set it to nil in their constructors).
>> So the fact that fill_space shows up in the docs for layout.flex and layout.ratio is misleading.
>
> I think we can fix that by overriding the property with a @property block and hide it with @hidden.

My implementation to hide a member in the documentation was incompatible with override members. I added a quick fixup commit to use another approach, more flexible (in regard to "override items to hide them"), but more ugly...

Hiding the property with an override in `wibox.layout.flex` (as I did in my second commit) was even smarter than expected as it also automatically hide it for `wibox.layout.ratio`. The property in `ratio` actually comes with the tag `@hidden` already sets, as it's inherited from `flex` (i.e. not from `fixed`).

sclu1034:
Suggestion to wrap `flex-wrap` the horizontal sections:

image

The important change is

```diff
diff --git a/docs/ldoc.css b/docs/ldoc.css
index a51fb80b..18544e54 100644
--- a/docs/ldoc.css
+++ b/docs/ldoc.css
@@ -522,8 +522,9 @@ pre .url { color: #272fc2; text-decoration: underline; }
 .extra_header_box {
     display: flex;
     flex-direction: row;
+    flex-wrap: wrap;
 }
 
 .extra_header_section {
-    flex: 1;
+    flex-grow: 1;
 }
```

I also replaced the nested `<ul>`s with a simpler `<div>` structure, so my full patch is

```diff
diff --git a/docs/ldoc.css b/docs/ldoc.css
index a51fb80b..436dcd9b 100644
--- a/docs/ldoc.css
+++ b/docs/ldoc.css
@@ -505,25 +505,21 @@ pre .url { color: #272fc2; text-decoration: underline; }
 }
 
 /* Inheritance diagram */
-.inheritance .inheritance__level {
-    list-style: none;
+.inheritance {
+    margin-bottom: 15px;
+    margin-left: 15px;
 }
 
-.inheritance .inheritance__level__node::before {
-    content: "↳";
-}
-
-.inheritance .inheritance__level__node--root::before {
-    /* simulate the spacing of the arrow character */
-    content: "   ";
-    white-space: pre;
+.inheritance .inheritance__level > .inheritance__level {
+    margin-left: 30px;
 }
 
 .extra_header_box {
     display: flex;
     flex-direction: row;
+    flex-wrap: wrap;
 }
 
 .extra_header_section {
-    flex: 1;
+    flex-grow: 1;
 }
diff --git a/docs/ldoc.ltp b/docs/ldoc.ltp
index 9ccfff00..1a0c0b7a 100644
--- a/docs/ldoc.ltp
+++ b/docs/ldoc.ltp
@@ -199,12 +199,9 @@
     <h3>Class Hierarchy</h3>
     <div class="inheritance">
 # local function draw_hierary_recursifly(i)
-    <ul class="inheritance__level">
-#   if i == #hierarchy then
-        <li class="inheritance__level__node inheritance__level__node inheritance__level__node--root">
-#   else
-        <li class="inheritance__level__node inheritance__level__node">
-#   end
+    <div class="inheritance__level">
+        <span>
+            <span>↳</span>
 #   local mod = hierarchy[i]
 #   local name = display_name(hierarchy[i])
 #   if mod == module then
@@ -212,13 +209,11 @@
 #   else
             <a href="$(ldoc.ref_to_module(mod))">$(name)</a>
 #   end
-        </li>
+        </span>
 #   if i > 1 then
-        <li>
-#           draw_hierary_recursifly(i - 1)
-        </li>
+#       draw_hierary_recursifly(i - 1)
 #   end
-    </ul>
+    </div>
 # end -- module.tags.supermodule
 # draw_hierary_recursifly(#hierarchy)
     </div>
```

Aire-One:
@sclu1034, I have kept your first diff about flex wrapping. I think nested lists are better to draw trees with a regard to data structure and accessibility, tho, so I dismissed the second part.

@Elv13 I have changed in your extra-header feature the class name to better match BEM naming convention (http://getbem.com/naming/) and I used `div` instead of `span` for the extra-header sections because `span` shouldn't be used as container.

If we don't have more changes request or places where `@supermodule` and `@hidden` tags should be added, I think this PR is ready to be merged now. Further improvements (and places where we forgot new tags) can be added later, I guess.

sclu1034 (approved):

Elv13 (approved):

Elv13:
Thanks a lot for this!

Elv13:
The doc has been regenerated using this. Overall, it seems to work. Some pages like https://awesomewm.org/apidoc/core_components/screen.html will need some fixup, it seems some tag don't close. I also managed to introduce some broken link in the landing page https://awesomewm.org/apidoc/index.html, I will fix that next week

actionless:
this class hierarchy thing is really cool :+1:

Aire-One:
> Some pages like https://awesomewm.org/apidoc/core_components/screen.html will need some fixup, it seems some tag don't close.

@Elv13 I think I closed one `<div>` tag too much when there is no extra-header for hierarchy... I'll take a look at that now.

PR #3292, opened 2021-03-22
Migrate CI to GitHub Actions

### LINKED ISSUE #3274: Travis CI credit issue
So, we have a few PR where the Travis CI check seems to be blocked. 

I did a quick Google search, and it seems Travis CI is no longer providing free CI for Open Source Software:

* https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom
* https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
* https://twitter.com/james_hilliard/status/1336081776691843072
    * https://www.reddit.com/r/linux/comments/k92tc0/travis_is_no_longer_free_for_opensource_projects/
    * https://www.reddit.com/r/programming/comments/k8v6u2/travis_ci_is_no_longer_providing_ci_minutes_for/
* https://ropensci.org/blog/2020/11/19/moving-away-travis/#a-change-of-management
* https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

What's our situation with Travis CI? Do we have a plan (/ should consider on taking one)?
Many people seem to move to Gihub Action. Should we plan such migration?
Should we consider other CI services?

sclu1034:
Fixes #3274. See that issue for rationale and discussions as well.

(TLDR: Travis changed their pricing which now handles billing via Credits. The initial 10000 Credits from the free tier ran out and builds started failing. GitHub Actions were determined as suitable replacement.)

### Workflows

This PR adds two workflows: `.github/workflows/main.yml` and `.githubs/workflows/apidoc.yml`.
They were build from `.travis.yml` and try to mirror its functionality as closely as possible.

**main.yml:**

This is the main workflow that builds the application and runs tests.

Mirroring Travis, tests are run in sequence, which is slow. If someone knows if/how it might be possible to split and later merge coverage reports (this includes C code coverage, since tests implicitly run the `awesome` binary), splitting tests into individual jobs would speed up the overall run time tremendously.

Another issue that increases run times is the fact that killing clients seems to be very slow in the GitHub Actions VM. I had to add explicit wait loops in multiple places for tests to pass.

To improve run times where possible, I used GitHub's `actions/cache`, which allows caching build artifacts and package files for dependencies. Since GitHub does not provide a way to manually clear these caches, the `key`s are very specific and affected by multiple factors to ensure caches are invalidated when needed.

**apidoc.yml:**

This builds the documentation and updates https://github.com/awesomeWM/apidoc. The steps to install dependencies and build the applications are mostly identical with `main.yml`. The rest is done in `.github/scripts/apidoc.sh`, where the previous `build-utils/apidoc-travis.sh` moved to. Other than renaming envvars where necessary, the script remains the same.

This one is not fully tested yet, since my fork lacks the required access. But given that I'm using the same shell script as the Travis build, this shouldn't bring too many issues.

### Required Settings

Two secrets must be [specified in this repository's settings](https://docs.github.com/en/actions/reference/encrypted-secrets).

1. Compared to Travis, Coveralls and `luacov-coveralls` are not able to determine access automatically. Instead, a repo token must be stored in a secret called `COVERALLS_REPO_TOKEN`. The value for that can be found in the repository settings in the Coveralls UI.
2. The `apidoc.sh` script requires write access to https://github.com/awesomeWM/apidoc. On Travis it used a token stored in `GH_APIDOC_TOKEN`. This can now be stored in a secret `APIDOC_TOKEN`. Additionally, GitHub Actions already include an internal token linked to the repo where they are run. Since these two repos belong to the same org, that default token might already have write access, so I've added that as a fallback.


sclu1034:
Logs for these workflows can be seen here: https://github.com/sclu1034/awesome/actions.

.github/workflows/main.yml
```diff
@@ -0,0 +1,405 @@
...
+defaults:
+  run:
+    # GitHub Actions adds `errexit` and `pipefail` by default, but we add `xtrace`
+    # to improve debugging some of the longer scripts.
+    shell: /bin/bash -o errexit -o pipefail -o xtrace {0}
+
+jobs:
+  main:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-20.04
```
Aire-One, line 23: same question.

```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,405 @@
...
+          - test_name: "lua5.1"
+            lua_version: "5.1"
+            lua_name: "lua5.1"
+
+          - test_name: "luajit"
+            lua_version: "5.1"
+            lua_name: "luajit"
+            lua_library: "/usr/lib/x86_64-linux-gnu/libluajit-5.1.so"
+            lua_include: "/usr/include/luajit-2.1"
+            luarocks_args: "--lua-suffix=jit-2.1.0-beta3"
```
Aire-One, line 58: Why is there differences in the used versions? I never used luajit before, so there is maybe something I'm missing...

Aire-One (commented):
This is good work!

I took an overview and I left a few comments with my "passing-by-thoughts".

I'll try to find time to pull your branch on my fork and play with it.

.github/workflows/apidoc.yml
```diff
@@ -0,0 +1,155 @@
...
+
+defaults:
+  run:
+    # GitHub Actions adds `errexit` and `pipefail` by default, but we add `xtrace`
+    # to improve debugging some of the longer scripts.
+    shell: /bin/bash -o errexit -o pipefail -o xtrace {0}
+
+jobs:
+  main:
+    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
sclu1034: About a week ago, I read "`ubuntu-latest` still points to 18.04 but will change to 20.04 soon", which means that this change happens transparently to anyone who doesn't explicitly check it.

At first, I just used it to force 20.04, but given the above, I still think it's better to pin the version. I already had issues between 16.04, 18.04 and 20.04, so it's likely that issues will also arise with whatever comes after 20.04.
If we were to use `ubuntu-latest`, the version switch would happen months after Canonical's release. And I doubt that at that point anyone would think to check the OS version for weird issues like "why does this unrelated test run into a timeout now?".

.github/workflows/main.yml
sclu1034: You mean the difference compared to Travis?

I figured that if we update the OS by two versions, the package is likely to update as well. `2.1.0-beta3` is the version reported by `apt-cache`.

.github/workflows/apidoc.yml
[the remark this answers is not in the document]
Aire-One: Yeap, I also noticed they where in a transition status for the label `ubuntu-latest` from 18.04 to 20.04. That's why I thought it would be better to use this label. We would have Ubuntu updates effortless.

You raised a good point, tho. I didn't expect Ubuntu updates to be source of issues... Well, I think it was actually expectable in a way.

Now I feel worried about how long GitHub is supposed to support a platform version and how often we'll need to rework/check workflows. Given Ubuntu 16.04 and 20.04 are LTS (and 16 is still an available platform), I guess Ubuntu 20 will stay for the while.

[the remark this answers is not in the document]
sclu1034: Ubuntu LTS lasts five years, and given that 16.04 (which runs out this April) is still available in GitHub Actions, it seems that GH goes with a similar life cycle.

Of course, GH has no obligations to mirror Cannonical's LTS schedule. But given that they already limit Linux VMs to just Ubuntu, it seems likely that they would support the full LTS period.

.github/workflows/main.yml
```diff
@@ -0,0 +1,425 @@
...
+            wmctrl \
+            x11-apps \
+            xcb-proto \
+            xdotool \
+            xorg \
+            xserver-xephyr \
+            xterm \
+            xutils-dev \
+            xvfb \
+            zsh \
```
ShayAgros, line 141: Any guesses when zsh is used in the builds ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,425 @@
...
+        with:
+          path: /tmp/apt/*.deb
+          # The trailing number serves as a version flag that can be incremented
+          # to invalidate the cache after changing the list of packages.
+          key: ${{ runner.os }}-${{ steps.get-date.outputs.date }}-apt-3
+
+      - name: Download apt packages
+        if: steps.cache-apt.outputs.cache-hit != 'true'
+        run: |
+          sudo apt-get update
```
ShayAgros, line 103: You should use [apt-fast](https://github.com/ilikenwf/apt-fast) instead. It comes pre-installed on Github's instances and makes the package download parallel.
It speeds up download time in my tests

Also it gets the same arguments as `apt` so it's a matter of  replacing `apt-get` with `apt-fast`

ShayAgros (changes requested):
It looks good to me overall. I'll continue reviewing it later.
This is a great work you did Lucas, thank you

.github/workflows/main.yml
```diff
@@ -0,0 +1,425 @@
...
+
+      - name: Configure CMake
+        run: |
+          cmake -S ${{ github.workspace }} -B "${{ github.workspace}}/build" \
+            -DAWESOME_VERSION=$AWESOME_VERSION \
+            -DLUA_LIBRARY=${{ matrix.lua_library || env.LUALIBRARY }} \
+            -DLUA_INCLUDE_DIR=${{ matrix.lua_include_dir || env.LUAINCLUDE }} \
+            -DGENERATE_DOC=${{ matrix.build_api_doc && 'ON' || 'OFF' }} \
+            -DGENERATE_MANPAGES=${{ matrix.build_api_doc && 'ON' || 'OFF' }} \
+            -DDO_COVERAGE=${{ matrix.coverage != '' && 'ON' || 'OFF' }} \
```
ShayAgros, line 266: I think it would work with `matrix.coverage` alone, works in my tests at least

.github/workflows/main.yml
sclu1034: Are you sure about `apt-fast`? In my tests it's on par at best, but worst in most cases. While on GitHub the difference is just a few seconds, when building a Docker image, `apt-fast` takes twice as long as `apt-get`.

### `apt-fast`:

#### GitHub Actions

> https://github.com/sclu1034/awesome/actions/runs/690632523
> Time according to the log, one for each job in that run

```
18, 16, 17, 13, 14
min: 13
max: 14
avg: 15.6
```

#### Docker

> Time according to `time`, three consecutive `docker build` calls. `Dockerfile` below.

```
real    0m47.414s
user    0m9.433s
sys     0m16.694s
```

```
real    0m46.554s
user    0m9.522s
sys     0m18.602s
```

```
real    0m44.212s
user    0m9.142s
sys     0m16.386s
```

### `apt-get`:

#### GitHub Actions

> https://github.com/sclu1034/awesome/actions/runs/690646513

```
11, 13, 13, 13, 12
min: 11
max: 13
avg: 12.4
```

#### Docker

```
real    0m24.696s
user    0m3.081s
sys     0m1.454s
```

```
real    0m26.177s
user    0m3.123s
sys     0m1.420s
```

```
real    0m24.770s
user    0m3.147s
sys     0m1.343s
```

---

<details>
<summary>Dockerfile</summary>

```Dockerfile
FROM ubuntu

ENV DEBIAN_FRONTEND=noninteractive

SHELL [ "/bin/bash", "-c" ]

RUN set -ex; \
    apt-get update; \
    apt-get install -y --no-install-recommends software-properties-common; \
    add-apt-repository ppa:apt-fast/stable; \
    apt-get update; \
    apt-get -y install apt-fast

RUN set -ex; \
    time apt-get install -y --no-install-recommends --download-only \
        build-essential \
        git \
        sudo \
        curl \
        cmake \
        automake \
        gettext \
        libtool \
        unzip \
        asciidoctor \
        libnotify-bin \
        libcairo2-dev \
        gir1.2-gtk-3.0 \
        libpango1.0-dev \
        libxcb-xtest0-dev \
        libxcb-icccm4-dev \
        libxcb-randr0-dev \
        libxcb-keysyms1-dev \
        libxcb-xinerama0-dev \
        libdbus-1-dev \
        libxdg-basedir-dev \
        libstartup-notification0-dev \
        imagemagick \
        libxcb1-dev \
        libxcb-shape0-dev \
        libxcb-util0-dev \
        libx11-xcb-dev \
        libxcb-cursor-dev \
        libxcb-xkb-dev \
        libxcb-xfixes0-dev \
        libxkbcommon-dev \
        libxkbcommon-x11-dev \
        xorg \
        dbus \
        librsvg2-dev \
        dbus-x11 \
        xterm \
        xdotool \
        wmctrl \
        xterm \
        xvfb \
        zsh \
        x11-apps \
        xutils-dev \
        gir1.2-pango-1.0 \
        libgirepository1.0-dev \
        libgdk-pixbuf2.0-dev \
        liblua5.3-dev \
        lua5.3 \
    ;

CMD "/bin/bash"
```

</details>

.github/workflows/main.yml
sclu1034: Not in the build, but when testing `lib/awful/completion.lua`: https://github.com/awesomeWM/awesome/blob/aeb2d85dad31073eaf979201f35dece4a51ccd91/spec/awful/completion_spec.lua

.github/workflows/main.yml
```diff
@@ -0,0 +1,425 @@
...
+      - name: Install fresh Luarocks
+        if: steps.cache-luarocks.outputs.cache-hit != 'true'
+        run: |
+          wget -O /tmp/luarocks.tar.gz https://github.com/luarocks/luarocks/archive/v3.5.0.tar.gz
+          mkdir /tmp/luarocks
+          tar -xf /tmp/luarocks.tar.gz -C /tmp/luarocks --strip-components=1
+          cd /tmp/luarocks
+          ./configure --lua-version=${{ matrix.lua_version }} --with-lua-include=${LUAINCLUDE} ${{ matrix.luarocks_args }}
+          make build
+          sudo make install
```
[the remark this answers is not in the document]
sclu1034: I quite like the simplicity of "run A xor B" over "A runs sometimes and when it does, it'll affect B as well".

.github/workflows/main.yml
ShayAgros: You did a much more thorough check than me. I tested `apt-fast` using `act` where it took 4-5 less to install packages.
I'm surprised to see that this isn't the behavior you see as well (it's counter intuitive to me actually).

Anyway, it seems like `apt-fast` performance worse (even if on some cases) than `apt-get`, so I withdraw my request to change it. Thank you for diving deep into it

.github/workflows/main.yml
Aire-One: Yes, it was sorted in the other comments thread. I'd like to mark these threads as resolved, but it seems only @sclu1034 can do. The only option I had was to hide my comment, and it has only hid my comment, not the full thread...

.github/workflows/main.yml
Aire-One: Nahh, I was just wondering myself why we use lua_library jit-5.1 with lua_include jit-2.1 and a lua-suffix jit-2.1.0-beta3. Why aren't these numbers more homogeneous?

Please dismiss this comment, tho. I'm not familiar with luajit and looking at the AUR PKGBUILD for [awesome-luajit-git](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=awesome-luajit-git), I guess it's the normal behavior. :shrug:

.github/scripts/apidoc.sh
```diff
@@ -0,0 +1,158 @@
...
+# This will re-use already existing branches (updated PR).
+if [ "$PR_NUMBER" != false ]; then
+    BRANCH="pr-$PR_NUMBER"
+elif [ "$GITHUB_HEAD_REF" != master ]; then
+    # Use merge-base of master in branch name, to keep different branches with
+    # the same name apart.
+    # shellcheck disable=SC2015
+    BRANCH="$GITHUB_HEAD_REF-$(cd "$REPO_DIR" \
+        && git fetch --unshallow origin master \
+        && git rev-parse --short "$(git merge-base HEAD FETCH_HEAD || true)" || true)"
```
ShayAgros, line 49: I might not understand this correctly.
This `if` part (`"$GITHUB_HEAD_REF" != master`) is when a commit is merged (git push), how can it have two branches with the same name ?
Even if it is possible, the merge base for all such branches would probably be `origin/master` so I'm not sure whether the issue is fixed in this code

.github/scripts/apidoc.sh
```diff
@@ -0,0 +1,158 @@
...
+# Use a temporary branch for the two commits, which allows for a better UI.
+git checkout -b merged-update
+
+# Create the README for the Git repo (https://github.com/awesomeWM/apidoc).
+cat > ../doc/README.md <<END
+# Awesome API documentation
+
+This repository contains the built API documentation for the
+[awesome](https://github.com/awesomeWM/awesome) window manager. It is
+automatically updated via Travis when the master branch changes. Hence:
```
ShayAgros: Travis -> Github Actions ?

.github/scripts/apidoc.sh
```diff
@@ -0,0 +1,158 @@
...
+    -I "<h2>API documentation for awesome, a highly configurable X window manager (version .*)\.</h2>" \
+    -x .git | patch -p1
+
+git add --all .
+if git diff --cached --exit-code --quiet; then
+    echo "Documentation has not changed."
+    exit
+fi
+
+LAST_COMMIT_MSG="$(cd "$REPO_DIR" && git log -1 --pretty=format:%s)"
```
ShayAgros, line 85: nit, `format` keyword in `pretty` argument seems redundant

.github/scripts/apidoc.sh
```diff
@@ -0,0 +1,158 @@
...
+
+# Commit the relevant changes.
+COMMIT_MSG="Update docs for $AWESOME_VERSION via Github Actions
+
+Last commit message:
+$LAST_COMMIT_MSG
+
+Commits: https://github.com/awesomeWM/awesome/compare/${GITHUB_BASE_REF}..${GITHUB_HEAD_REF}
+Build URL: $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID"
+git commit -m "[relevant] $COMMIT_MSG"
```
ShayAgros, line 96: nit, can you insert an empty line between lines 95 and 96 ?
This would help to identify lines 89-95 as a single quoted block

.github/workflows/main.yml
```diff
@@ -158,7 +158,7 @@ jobs:
         with:
           path: /tmp/luarocks
           # The build input for luarocks changes per test, so we need separate caches
-          key: ${{ runner.os }}-${{ matrix.test_name }}-luarocks-3.5.0
+          key: ${{ github.workflow }}-${{ runner.os }}-${{ matrix.test_name }}-luarocks-3.5.0
```
ShayAgros, line 152: Why can't we use the same `luarocks` cache ?

.github/workflows/main.yml
```diff
@@ -95,7 +95,7 @@ jobs:
           path: /tmp/apt/*.deb
           # The trailing number serves as a version flag that can be incremented
           # to invalidate the cache after changing the list of packages.
-          key: ${{ runner.os }}-${{ steps.get-date.outputs.date }}-apt-3
+          key: ${{ github.workflow }}-${{ runner.os }}-${{ steps.get-date.outputs.date }}-apt-3
```
ShayAgros: I guess once your PR that allows to specify lua version is merged we could use the same apt cache for both workflows right ? (didn't cross check the packages thoroughly but I'm guessing the lua5.3 package installed in this workflow is what makes them different)

.github/workflows/main.yml
```diff
@@ -184,7 +184,7 @@ jobs:
           path: /tmp/xcb-errors
           # xcb-errors doesn't have versioned releases, so we checkout and cache by commit.
           # Make sure to keep this in sync with the `git checkout` below.
-          key: ${{ runner.os }}-xcb-errors-5d660ebe872cadcdc85de9d6f9afe05de629c030
+          key: ${{ github.workflow }}-${{ runner.os }}-xcb-errors-5d660ebe872cadcdc85de9d6f9afe05de629c030
```
ShayAgros: Same question as in line 161

ShayAgros (changes requested):
Sorry for the time it took me to review it (busy weeks) but now I've reviewed it all.

First I want to say that that this is an extraordinary work you did here. This is by no means a small task, and the resulting PR is written very well

I prepended comments which are nice-to-have in my opinion with _nit_ and I'm ok if you decide not to use them.
Other than some small comments I think it can be merged (gonna need people w/ write access, but I hope my feedback would help them).

Again, amazing work

Elv13:
> First I want to say that that this is an extraordinary work you did here. This is by no means a small task, and the resulting PR is written very well

Yes, really, to everybody who contribute to this ticket, it is super appreciated. I don't have that much time these days, I just love to see the dedication here to move AwesomeWM forward!

.github/workflows/main.yml
sclu1034: The main workflow installs various packages related to running tests that aren't needed when building docs.

.github/workflows/main.yml
sclu1034: We could, assuming that it handles the case correctly where both workflows run into the same cache miss at the same time.
But do we need to? If we have at least one CI run every 7 days, this cache will never invalidate. And even if it does, builds times for LuaRocks aren't bad (don't remember the number, but should be less than 2 minutes).

.github/scripts/apidoc.sh
```diff
@@ -0,0 +1,158 @@
...
+    exit
+fi
+
+# NOTE: DO NOT USE "set -x", or anything else that would reveal APIDOC_TOKEN!
+# GitHub Actions does filter secrets, but extra caution won't hurt
+set -e
+set +x
+
+# Display exit code in case of failure (probably due to 'set -x').
+trap '[ "$?" = 0 ] || echo "EXIT CODE: $?"' EXIT
```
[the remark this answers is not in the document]
sclu1034: This file was moved from `build-utils/travis-apidoc.sh` (and looking at the comments below, I missed some references). Maybe GitHub wasn't able to display that accurately.

All code that isn't specific to Travis/GitHub Actions was left as-is, since this can't be fully tested in a fork. Some parts require write access to https://github.com/awesomeWM/apidoc, so I was only able to run it with some NOOPs.
I'll implement the formatting suggestions, but the suggestions related to functionality should be done in a PR that has the required access to properly test such changes.

.github/workflows/main.yml
```diff
@@ -0,0 +1,425 @@
...
+
+          git checkout -qf ${{ github.head_ref }}
+          if [ -n "$failed" ]; then
+            echo "Checks failed for these commits:"
+
+            for c in $failed; do
+              git show --no-patch --pretty="%h %s (%an, %ad)" "$c"
+            done
+
+            exit 1
```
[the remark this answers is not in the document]
sclu1034: Yes. If we were to exit with `0` here (i.e. no `exit` at all), the CI would consider this step successful.

.github/workflows/main.yml
```diff
@@ -0,0 +1,425 @@
...
+      # Check each commit separately (to make git-bisect less annoying).
+      - name: Test previous commits
+        if: matrix.test_prev_commits && github.event_name == 'pull_request'
+        run: |
+          # `actions/checkout` creates a shallow repo (`--depth 1`) by default,
+          # which is fine for everything up until now. But we need individual commits now.
+          # And we only want to unshallow now, to not slow down the checkout for other jobs.
+          git fetch --unshallow
+
+          rev_list="$(git rev-list --bisect-all origin/${{ github.base_ref }}..origin/${{ github.head_ref }})"
```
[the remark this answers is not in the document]
sclu1034: True. It may have only been working for me because I was opening PRs within my own repo. But the docs are quite lackluster on what these values can be.
However, they wouldn't be there if it was as simple as `HEAD`.

My best bet would be to simply drop the `origin/` and hope that `base_ref` and `head_ref` get adjusted according to the context the workflow is run in. This might be another thing that can only be tested after this has been merged.

.github/workflows/main.yml
```diff
@@ -0,0 +1,425 @@
...
+        run: cd "${{ github.workspace }}/build" && make check-qa
+
+      # Check each commit separately (to make git-bisect less annoying).
+      - name: Test previous commits
+        if: matrix.test_prev_commits && github.event_name == 'pull_request'
+        run: |
+          # `actions/checkout` creates a shallow repo (`--depth 1`) by default,
+          # which is fine for everything up until now. But we need individual commits now.
+          # And we only want to unshallow now, to not slow down the checkout for other jobs.
+          git fetch --unshallow
```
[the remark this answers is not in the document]
sclu1034: It certainly would. But other than some trickery to count the number of commits in a shell script and expose that to `actions/checkout`, I don't see a way to know the correct depth.

And I think the current solution is good enough. The "Check previous commits" only runs for one of the jobs anyways.

tests/run.sh
```diff
@@ -13,7 +13,7 @@
 set -e
 
 export SHELL=/bin/sh
-export HOME=/dev/null
+export HOME="${TMP:-/tmp}"
```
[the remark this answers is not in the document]
sclu1034: Correct.

.github/workflows/main.yml
ShayAgros: Sure, but most of them overlap I guess. Anyway, it's irrelevant for now

.github/workflows/main.yml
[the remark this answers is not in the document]
ShayAgros: Hmm I guess we better get it merged and see afterwards if things actually work. I feel that dropping `origin` part there is the correct approach, but let's see..

ShayAgros (dismissed):
I believe it would benefit the repository greatly if this PR is merged.

Great work Lucas

actionless:
is it already working?

sclu1034:
> is it already working?

Not in this repository. GitHub doesn't allow running actions added/changed by PRs, unless the source and destination repositories are identical. Otherwise, a malicious actor would be able to execute arbitrary code in the context of someone else's repository.
For this to run within the context of https://github.com/awesomeWM/awesome, it would need to be merged first.

Until then, you can check out the runs in my repository: https://github.com/sclu1034/awesome/actions.

actionless (dismissed):

.github/workflows/apidoc.yml
```diff
@@ -0,0 +1,151 @@
...
+      - name: Install cached Luarocks
+        if: steps.cache-luarocks.outputs.cache-hit == 'true'
+        run: |
+          cd /tmp/luarocks
+          sudo make install
+
+      - name: Install rocks
+        run: |
+          sudo -H luarocks install lgi
+          sudo -H luarocks install ldoc
```
Aire-One, line 107:
```suggestion
          sudo -H luarocks install lgi
          sudo -H luarocks install ldoc
          sudo -H luarocks install lua-discount
```

We should add `lua-discount` rock installation as it is what we use:

https://github.com/awesomeWM/awesome/blob/a4572b9b52d89369ce3bd462904d536ec116dc35/docs/config.ld#L11

(ldoc reference: https://stevedonovan.github.io/ldoc/manual/doc.md.html#Markdown_Support)

Not having this package installed results on the `format: using built-in markdown` message on your last "Update API doc" Actions run (https://github.com/sclu1034/awesome/runs/2371543797?check_suite_focus=true#step:14:421).

With this rock installed, it seems ldoc finishes faster on my machine.

Aire-One: I just figured there is a `lua-discount` package installed from apt-get in the "Build & test" workflow. (That explains why we don't have the "format message" in logs)

Maybe we should do the same here instead of installing the rock as Ubuntu seems to be more up-to-date than Luarocks (1.2.10.1 vs 2.1.8).

- https://packages.ubuntu.com/focal/lua-discount
- https://luarocks.org/modules/luarocks/lua-discount

Aire-One (changes requested):

actionless (dismissed):

Aire-One:
Sorry for the delay @actionless.

I think this PR is at 1:1 status compared to our Travis CI pipeline. (Maybe even a few bugs seems to be corrected, including the placement test that was randomly failing on Travis).

However, I'll not give my "validation review" as it would trigger the merge. This PR is going to alter our workflow, so I'd prefer @psychon or @Elv13 to take the merge responsibility. (Especially given Elv recently paid for Travis CI (BTW do we have a way to see how much credit left we have?))

Oh, wait! I just had a thought about Mergify. We should also update its configuration in the scope of this PR. (https://docs.mergify.io/conditions/#github-actions)

sclu1034:
> However, I'll not give my "validation review" as it would trigger the merge.
I don't think it would have triggered a merge, since the checks for Travis and coverage will never complete here.
But now that this PR changes `.mergify.yml`, Mergify won't run anyways.

Aire-One:
Thank you, @sclu1034 for these details.

With mergify updated, I don't see any other blocker now. I think we're ready to merge!

Aire-One (approved):
>> However, I'll not give my "validation review" as it would trigger the merge.
>
> I don't think it would have triggered a merge, since the checks for Travis and coverage will never complete here.
> But now that this PR changes .mergify.yml, Mergify won't run anyways.

Fair enough...

Thank you, @sclu1034  and @ShayAgros for the work here! :heart:

actionless (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,394 @@
...
+          # Lua 5.2 with fixed lgi version and screen size not divisible by 2.
+          - test_name: "fixed-lgi-lua5.2"
+            lua_version: "5.2"
+            lua_name: "lua5.2"
+            lgi_version: "0.9.2"
+            tests_screen_size: "1921x1079"
+            check_qa: true
+            empty_theme_while_loading: true
+            # SOURCE_DATE_EPOCH: used for stable dates during build.
+            source_date_epoch: "1893456000"
```
psychon, line 63: AFAIR this was added to make the apidoc stable. I guess lua 5.2 was the apidoc build before...?

Anyway, this makes more sense to add to the apidoc workflow and not here, I think (but I might misunderstand something)

.github/workflows/main.yml
```diff
@@ -0,0 +1,394 @@
...
+          cd /tmp/luarocks
+          sudo make install
+
+      - name: Cache xcb-errors
+        id: cache-xcb-errors
+        uses: actions/cache@v2
+        with:
+          path: /tmp/xcb-errors
+          # xcb-errors doesn't have versioned releases, so we checkout and cache by commit.
+          # Make sure to keep this in sync with the `git checkout` below.
```
psychon, line 177: What exactly does this mean and what is wrong with https://xcb.freedesktop.org/dist/xcb-util-errors-1.0.tar.bz2? (Found by looking at what Arch does: https://github.com/archlinux/svntogit-community/blob/55a631b407cc54dd60a85328c8fc1f60468c5dbf/trunk/PKGBUILD#L13 )

tests/run.sh
psychon, line 16: Might I ask: Why this change? Is it worth adding a comment explaining why need `$HOME` to be a directory?
(I don't really think this matters much, but could this have security implications? Someone else could create `/tmp/.bashrc` and I bet that would end up getting executed)

tests/run.sh
```diff
@@ -133,7 +134,7 @@ echo "awesome_log: $awesome_log"
 
 wait_until_success() {
     if (( verbose )); then set +x; fi
-    wait_count=60  # 60*0.05s => 3s.
+    wait_count=$(echo "${TEST_TIMEOUT} / 0.05" | bc)
```
psychon, line 137:
```suggestion
    wait_count=$((${TEST_TIMEOUT} * 20))
```

tests/test-leak-client.lua
```diff
@@ -85,12 +85,23 @@ local steps = {
...
             local success, msg = pcall(function() tostring(objs[1]) end)
-            assert(not success)
+            assert(not success, msg)
             assert(msg:find("invalid object"), msg)
 
-            -- Check that it is garbage-collectable
-            collectgarbage("collect")
-            assert(#objs == 0)
+            -- On GitHub Actions, it can take a while for clients to be killed
+            -- properly.
```
psychon, line 92: This makes no sense to me. With "clients to be killed properly" you mean that you are waiting for something to close its X11 window? If so, then a `while` loop in Lua will not work for waiting for that. The C code does not handle X11 events until you return to it, thus this code cannot wait until some window closed.

The whole test suite stuff exists to "wait for something". So, if you want to wait for something, then add another step to this.

Also, this whole code is inside a `if #client.get() == 0`, so at this point we are sure that there are no open clients.

So, what is this even waiting for? The only thing that happens in this loop is that you repeatedly run the garbage collector and run `sleep 1` in between. The `sleep 1` will only cause delays and have no other effects besides that. So, that leaves `collectgarbage("collect")`. Did something change so that we now need more iterations of the GC before something becomes collectable? What keeps clients alive now?

:-/

tests/test-leak-client.lua
```diff
@@ -85,12 +85,25 @@ local steps = {
...
             assert(msg:find("invalid object"), msg)
 
             -- Check that it is garbage-collectable
             collectgarbage("collect")
-            assert(#objs == 0)
+
+            -- On GitHub Actions, it can take a while for clients to be killed
+            -- properly.
+            local tries = 0
+            while (#objs > 0 and tries < 60) do
```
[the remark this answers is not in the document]
psychon: No. It's a weak table. The only thing removing entries is the garbage collector.

(Also, Lua does not support multithreading)

.github/workflows/main.yml
```diff
@@ -0,0 +1,394 @@
...
+      - name: Install codecov.io uploader
+        if: matrix.coverage == 'codecov'
+        run: wget -O /tmp/codecov-bash https://codecov.io/bash
+
+      # Check out repository to ${{ github.workspace }}
+      # Automatically picks the current branch/PR
+      - uses: actions/checkout@v2
+
+      - name: Create Build Environment
+        run: cmake -E make_directory -B "${{ github.workspace }}/build"
```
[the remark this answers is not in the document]
sclu1034: Nothing should be wrong with that, as far as I know.
This is simply one of the few lines from GitHub's CMake template that I never had to touch since I first started all this.

.github/workflows/main.yml
sclu1034: The Travis job simply cloned `master`. To pin the version, I checked https://gitlab.freedesktop.org/xorg/lib/libxcb-errors/-/releases, but found it to be empty. I did forget to check the tags, apparently.
So I used the one revision that Git always offers.

.github/workflows/main.yml
sclu1034: Looking at the initial issue for it (#2070), that seems correct.
And I guess while it is evaluated during tests, the fact that the other jobs run fine without it suggests that it has no effect on test outcome and is therefore not needed here.

sclu1034:
Triggered by https://github.com/awesomeWM/awesome/pull/3292#discussion_r628767745, I ran another small set of tests, where I "toggled" the loop in `tests/test-leak-client.lua`.

| Loop | Fails | Run |
|-|-|-|
| no | yes | [823852962](https://github.com/sclu1034/awesome/actions/runs/823852962) |
| yes | no | [824876995](https://github.com/sclu1034/awesome/actions/runs/824876995) |
| no | no | [824916481](https://github.com/sclu1034/awesome/actions/runs/824916481) |
| yes | yes | [824943015](https://github.com/sclu1034/awesome/actions/runs/824943015) |
| yes | no | [824951769](https://github.com/sclu1034/awesome/actions/runs/824951769) |
| no | yes | [824952563](https://github.com/sclu1034/awesome/actions/runs/824952563) |
| yes | no | [824953364](https://github.com/sclu1034/awesome/actions/runs/824953364) |

While it is a small sample size, it does mirror my experience where the "no loop" version is very likely to fail. It also shows that LuaJIT is the most likely to fail, but in previous runs I had other jobs fail as well.

However, [824943015](https://github.com/sclu1034/awesome/actions/runs/824943015) is problematic, as it shows that there is still a small chance for it to fail anyways. As outlined in the linked comment, I still don't know the actual cause for the issue, so the best I could do here is increase the number of tries.

But at this point, I'm not sure if we should do that. It's already obvious that the GitHub Actions isn't reliable for this particular test, so unless someone finds the actual issue/cause, it might be better to just disable the test here.

actionless:
@sclu1034 as i see from your table it's flakiness not depends on the loop, so just undo the arguable change for now, so we could get this merged quicker

sclu1034:
@actionless I'm not sure how that would help. The loop doesn't fix the issue, but without it, the test fails on almost every run.
We'd be better off to just disable this specific test during CI runs, so that we get reliable test runs.

.github/scripts/apidoc.sh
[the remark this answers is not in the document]
blueyed: This was added in 30eafb113.
@ShayAgros 
At least it should have said `probably due to 'set -e'`, not `set -x`.
I think it achieves what it wanted: to display the exit code if it is not zero, no?
Apart from that I think it could be fixed up (in a separate patch/PR).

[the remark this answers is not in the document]
ShayAgros: First of all sorry for not being more involved in this PR. I hope this would change soon.

@blueyed I'm not sure what the code actually tries to achieve. I'd guess that it was added to give more information about why the script failed by tapping on the exit signal. Using `set -e` would make more sense here imo, not sure why it didn't.

The problem with this line is that by the time `echo "EXIT CODE: $?` is executed the value of `?` is the one returned from the expression `[ "$?" = 0 ]` and not by the failed command. The snippet below should have returned 127 which is the error code of unknown command, but instead it returns 1:
```bash
set -e

trap '[ "$?" = 0 ] || echo "EXIT CODE: $?"' EXIT

dsajkld
```

So printing the exit code here doesn't add information. The snippet I suggested traps the error code correctly:
```bash
set -e

trap 'echo EXIT CODE: $?' ERR

dsajkld
```
returns:
```
./test.sh: line 8: dsajkld: command not found
EXIT CODE: 127
```

sclu1034:
> We'd be better off to just disable this specific test during CI runs, so that we get reliable test runs.

I guess, I'll just do that now. Feel free to comment on the change.

Elv13:
@sclu1034 Do you think this is ready to merge? With the delays in finishing 4.4 and the nice screenshot page proposed elsewhere, I think we could make the transition before 4.4. Originally, I wanted to avoid making CI changes "just before a release", but given the delays, it starts to be pointless to block this great work.

sclu1034:
In general, yes, I'd say this is ready.
I'd probably give it another test run, though, given that the last one has been quite a while ago.

Elv13 (approved):

actionless:
@sclu1034 please inform about the results of that aforementioned test run

sclu1034:
Rebased branch went through fine: https://github.com/sclu1034/awesome/actions/runs/1222501408

actionless (approved):
let's see how it goes :D

Aire-One:
🚀

Elv13:
It seems it worked very well!

The coverage run failed with:

```
--------------------
Uploading ...
Upload error: Couldn't find a repository matching this job.
Raw data:	Couldn't find a repository matching this job.
--------------------
Upload error: Couldn't find a repository matching this job.
Raw data:	Couldn't find a repository matching this job.
--------------------
Upload error: Couldn't find a repository matching this job.
Raw data:	Couldn't find a repository matching this job.
--------------------
Uploading done
```

Any idea what the issue is? But overall, for such a large change, it worked great. Thanks a ton for this work.

sclu1034:
I assume that's from coverage. Did you set the API token as described at the top?

Elv13:
> Did you set the API token as described at the top?

whoops

Elv13:
I gave it a few more tests using the wallpaper PR. I tried to add the coverall token, but I probably did it wrong because it still doesn't work. Maybe we could just remove coveralls.io? CodeCov is much better these days.

Another little papercut is that the commit by commit test compared `origin/master` with `origin/awful_wallpaper`. I think it should be `elv13/awful_wallpaper`?

PR #3293, opened 2021-03-22
fix(doc): gears.table.keys: s/integer keys/integer/

blueyed:


blueyed:
btw: is `#gears.table.keys(tab)` a good / recommended way to get the length of (any) table?

sclu1034 (approved):

Elv13 (approved):

sclu1034:
> btw: is `#gears.table.keys(tab)` a good / recommended way to get the length of (any) table?

That's `O(n^4)`. Something like this would be a better option:

```lua
function gtable.count_keys(t)
    local count = 0
    for _, _ in pairs(t) do
        count = count + 1
    end
    return count
end
```

Aire-One:
> btw: is `#gears.table.keys(tab)` a good / recommended way to get the length of (any) table?

Let's see:
`gears.table.keys` returns a table with all keys on a given table. The length of this table of keys would be the number of keys on the initial table. So, I guess, yes.

It works only because `gears.table.keys` uses `pairs` to iterate over the table keys, so it also gets keys for "non-attributes properties".

Elv13:
@sclu1034 Can you do a PR for this?

For now I will merge this PR once the CI finishes

Aire-One (approved):

PR #3294, opened 2021-03-22
Add utility to count table keys

sclu1034:
See #3293.

### DIFF
--- a/lib/gears/table.lua
+++ b/lib/gears/table.lua
@@ -176,7 +176,7 @@ end
 -- @staticfct gears.table.count_keys
 function gtable.count_keys(t)
     local count = 0
-    for _, _ in pairs(t) do
+    for _ in pairs(t) do
         count = count + 1
     end
     return count
--- a/lib/gears/table.lua
+++ b/lib/gears/table.lua
@@ -165,6 +165,21 @@ function gtable.keys(t)
     return keys
 end
 
+--- Get the number of keys in a table, both integer and string indicies.
+--
+-- This is functionally equivalent, but faster than `#gears.table.keys(t)`.
+--
+-- @tparam table t The table for which to count the keys.
+-- @treturn number The number of keys in the table.
+-- @staticfct gears.table.count_keys
+function gtable.count_keys(t)
+    local count = 0
+    for _, _ in pairs(t) do
+        count = count + 1
+    end
+    return count
+end
+
 --- Filter a table's keys for certain content type.
 --
 -- @tparam table t The table to retrieve the keys for.
--- a/lib/gears/table.lua
+++ b/lib/gears/table.lua
@@ -169,6 +169,8 @@ end
 --
 -- This is functionally equivalent, but faster than `#gears.table.keys(t)`.
 --
+-- @DOC_text_gears_table_count_keys_EXAMPLE@
+--
 -- @tparam table t The table for which to count the keys.
 -- @treturn number The number of keys in the table.
 -- @staticfct gears.table.count_keys
--- a/lib/gears/table.lua
+++ b/lib/gears/table.lua
@@ -415,3 +415,5 @@ function gtable.map(f, tbl)
 end
 
 return gtable
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/spec/gears/table_spec.lua
+++ b/spec/gears/table_spec.lua
@@ -7,6 +7,34 @@ describe("gears.table", function()
         assert.is.same(gtable.keys(t), { 1, 2, "a" })
     end)
 
+    describe("gears.table.count_keys", function()
+        it("counts keys in an empty table", function()
+            local t = {}
+            assert.is.same(gtable.count_keys(t), 0)
+        end)
+
+        it("counts keys in a sparse array", function()
+            local t = { 1, nil, 3 }
+            assert.is.same(gtable.count_keys(t), 2)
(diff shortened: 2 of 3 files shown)


Aire-One (dismissed):

Aire-One:
Thanks for the extra effort on unit testing!

lib/gears/table.lua
```diff
@@ -165,6 +165,21 @@ function gtable.keys(t)
     return keys
 end
 
+--- Get the number of keys in a table, both integer and string indicies.
+--
+-- This is functionally equivalent, but faster than `#gears.table.keys(t)`.
+--
```
Elv13, line 171: Can you add an `@DOC_..._EXAMPLE@` here? Copy paste a `tests/examples/text/gears/table/<something>.lua` file with a new name, then reference it here.

I/we are trying to enforce that all new APIs have an unit-tested code example these days. Think the of "StackOverflow syndrome".  People tend to copy paste code even if they actually know how to write it :P

Elv13 (dismissed):

blueyed (commented):
Is there a better name maybe?
Initially I've also search for `len` (via/with https://github.com/awesomeWM/awesome/pull/3293).
But in general it appears to be good / not too bad, since it is explicit about what it does exactly.

sclu1034:
I'm not sure that there is a perfect name.
Generic names like `len()`, `length()` or `size()` are conventions in other languages, but I think they would always be ambiguous in whether they operate on continuous arrays, sparse arrays or the whole table.
And since we're not chaining like `.keys().len()`, I think `count` is better at conveying the fact that it's about an amount. Something like `keys_len` could be possible, but doesn't feel right. Flipping things into `key_count`, could be an option though.

Regardless, `values` instead of `keys` could be an option as well, as it maps to how you write a table literal. I.e. in `{ 1, nil, 2, a = 3 }` you only see one key, but three values. Obviously not a big deal to most, but it might be more intuitive for people new to Lua.

lib/gears/table.lua
```diff
@@ -165,6 +165,23 @@ function gtable.keys(t)
...
+-- @DOC_text_gears_table_count_keys_EXAMPLE@
+--
+-- @tparam table t The table for which to count the keys.
+-- @treturn number The number of keys in the table.
+-- @staticfct gears.table.count_keys
+function gtable.count_keys(t)
+    local count = 0
+    for _, _ in pairs(t) do
```
adamnejm, line 179: Value assignment is unnecessary.
`for _ in pairs(t) do` will be sufficient.

adamnejm (changes requested):

ShayAgros (changes requested):
looks good. if `for _ in pairs(t) do` passes your CI for all lua versions, I'd rather use it as @adamnejm suggested

ShayAgros (approved):

Aire-One (approved):

Elv13:
> Is there a better name maybe?

I think count is fine. Length is more relevant for arrays, but this is for dictionaries. I always consider python choice to use the same name on all structures to be strange.

Elv13 (approved):

adamnejm:
It's good to think about the future and I believe `key_count` is more versatile.
Assume Awesome didn't allow `get`, `has`, `is`, etc. prefixes for functions names and you're presented with the whole ordeal where you want related functions to be called similarly, yet it would sometimes be borderline impossible to make that work.
Take this expression as an example:
```lua
next(tbl) == nil
```
Now put it in a function with name similar to `count_keys`. The first conclusion would be to call it `empty_keys` since `keys` is present at the end in other related functions, but that points to a destructive, not evaluative behavior of removing keys from a table.

If however there was a function called `key_count`, you would naturally gravitate towards naming it `keys_empty`, which isn't perfect either, but less deceptive and I'm positive it could eliminate the documentation / source lookup for some people.

That said, the function would probably be called `is_empty` / `isempty` because those prefixes are already present in the code base.
Despite that, I still think considering those is helpful and may pay off in the future.

*TL;DR:* Rename `count_keys` to `key_count`.

sclu1034:
That suggestion violates a convention even more general than `get`/`set`:

- noun -> value/property
- verb -> function

By that convention (and disregarding the contextual meaning of `gears` as a utility namespace), I would expect `gears.table.key_count` to be a property and these to be valid usage:

```lua
print(gears.table.key_count + 1)
gears.table.key_count = 5
```

With `count_keys`, however, the name itself makes it immediately clear to me that this is an operation (i.e. function) that `count`s `keys`.

And I don't see how `count_keys` would need to be the template for all future function names. As you say, `empty` as a verb has a different meaning, so I wouldn't have considered that at all.
`is_empty` would have been my first choice with your example, but even if the `keys` had to be part of the name, I would simply negate the expression and call it `has_keys`.

Elv13:
Yeah, accessors are pretty much banned these days. There is a few exceptions, but they are mostly legacy APIs. Properties and fields are preferred on object. But `count_keys` isn't on an object. It's an helper function to work with non `gears.object` tables. So I am fine with the name as it currently is.

@blueyed are you fine with the current PR? If so, please approve.

sclu1034:
I guess Mergify picked up @Aire-One becoming a member without re-approving.

Aire-One:
> I guess Mergify picked up @Aire-One becoming a member without re-approving.

Oops, I guess so... Sorry guys. 

@blueyed, @adamnejm, @sclu1034 : Just keep going on the discussion here / reopen a new PR for additional changes if needed.

PR #3295, opened 2021-03-22
Docs: Fix code example for cairo usage.

potamides:
I was playing around with cairo, when I noticed that the code example in the docs for cairo surfaces didn't work. `cr:set_source(1, 0, 0)` should actually be `cr:set_source_rgb(1, 0, 0)`.


Elv13 (approved):

Elv13:
Thanks!

PR #3297, opened 2021-03-24
fix(beautiful: gtk): don't crash when both gtk3 and gtk4 are installed

actionless:


sclu1034 (approved):

Aire-One (approved):

Aire-One:
Out of curiosity : when LGI will support GTK 4, we'll need to update this `lgi.require` to ask for GTK 4 (with fallback on version 3, maybe)? Or we'll revert this commit?

actionless:
@Aire-One taking into consideration what all the time starting from GTK 3 and onward its devs were breaking compatibility in every commit -- no, i think we'll have to got two func implementations, for gtk3 and gtk4

Elv13 (approved):

psychon (approved):

PR #3309, opened 2021-04-01
Implement overflow container

sclu1034:
This container adds overflow behavior:

- adds a scrollbar
- adds scrollbar dragging to scroll
- adds mouse scroll behavior
- allows children to render at (near) infinite size in scroll direction

## TODOs

- [x] Implement
- [ ] Write docs
- [ ] Write tests
- [ ] Write examples

## Demonstration

https://user-images.githubusercontent.com/4508454/113277319-e3f8e000-92e0-11eb-8610-f89ce8ada1f7.mp4


Aire-One (changes requested):
Here is a first review. I'll do a second pass later...

I have adopted the "nit" notation for "not important changes" (aka dismissable comments) ;) 

Here are some oter notes:

I think the `position` property name is misleading. It can lead to confusion with the `scrollbar_position` property and is not very descriptive to its usage. Maybe something like "scroll_factor" can be better?

I can see you have added inherited documentation from fixed layout. Can you add the ldoc `@baseclass wibox.widget.fixed` tag to these blocks waiting for my other PR to be merged? So that, the doc for this module is fully operational by itself. We can do a patch later to add the `@supermodule` tag and remove duplicated doc.

BTW, I can't see some methods like `fixed:add` and `fixed:remove`.  I think we are supposed to have access to these methods as well.

~Also, in regard to the "use gears magic", I suggest removing the documentation for your getters and setters.~
Unless you want to wait my other PR to use `@hidden` here...

Examples `tests/examples/wibox/layout/overflow/scrollbar_width.lua` and `tests/examples/wibox/layout/overflow/scrollbar_widget.lua` don't show scrollbar at all. I have fixed it by adding a  fourth widget.

Elv13 (changes requested):
Fantastic contribution. Many people, me included, have used many hacky tricks to get this kind of feature. At first I feared it one one of those sub-surface type container, but no, it is implemented very nicely! I really like that it's using a custom widget for the scrollbar. At first I was thinking the slider widget code might have been re-used/shared, but looking into it closer, what you did it better.

One bug I think this has, but I have not tested, is when using this within a mirror or rotate container. This is why the `geo.hierarchy:get_matrix_from_device` and `matrix:transform_point(mouse.x, mouse.y))` lines from the slider widget are important. There is some legitimate use case for mixing those containers with the overflow one. The main one if people using it to rotate their entire wibox 90 degree and having something like the taskbar in an overflow container. It's one of those annoying corner case, but I would like it to be fixed. It ain't *that* hard to fix.

lib/wibox/layout/overflow.lua
```diff
@@ -0,0 +1,408 @@
...
+            used_in_dir = used_in_dir + w
+            used_max = math.max(used_max, h)
+        end
+    end
+
+    used_in_dir = used_in_dir + self._private.spacing * (#widgets-1)
+
+    -- Save size for scrolling behavior
+    self._private.avail_in_dir = avail_in_dir
+    self._private.used_in_dir = used_in_dir
```
psychon, line 126: What if the overflow container is shown in more than one place with different sizes? Assigning size-information to "something in `self`" sounds like either a bug or a bug-waiting-to-happen to me.

This problem is why the scroll container does all that funky business with weak tables.

(Being able to show a widget in more than one place is great for users, but it brings so many complications...)

lib/wibox/layout/overflow.lua
```diff
@@ -0,0 +1,408 @@
...
+        end
+    end
+
+    return result
+end
+
+function overflow:before_draw_children(_, cr, width, height)
+    -- Clip drawing for children to the space we're allowed to draw in
+    cr:rectangle(0, 0, width, height)
+    cr:clip()
```
psychon, line 262: This clips what is drawn, but the input code does not know that you clipped something. Assuming you place a widget partially outside of the container (I have not yet entirely understood `:layout()` yet and am not 100% sure what is done here), it will still be possible to click into the not-visible parts of the widget.

(The scroll container does dark magic that results in *no* mouse clicks being possible; not really much better.)

lib/wibox/layout/overflow.lua
sclu1034: This might be a showstopper then. Knowing the exact geometry is pretty much required for (useful) scrolling to work.

I did check out the `scroll` container, but can't claim to understand its funky business yet. And even if I did, combining funky business with user input sounds like a "bug-waiting-to-happen" as well.

It would probably require changes to the core of the widget system so that when something attempts to re-use an existing instance, that widget may request to be cloned instead (maybe with custom logic attached to the clone function).

lib/wibox/layout/overflow.lua
sclu1034: Your assumption is correct. Another user reported this as well, as a general problem with clipped drawing. If I remember them correctly, clicks fire for any non-visible part of a widget inside a container with some `cr:clip()`, even if that part doesn't extend outside the container's geometry.

For the specific case here, this patch supposedly helps: https://github.com/xinhaoyuan/awesome/commit/384b9eaa597396f53e944865c61ca0edd5e4c509, but I guess there are other use cases that need things to work like they do currently.

lib/wibox/layout/overflow.lua
```diff
@@ -0,0 +1,619 @@
...
+    -- If the direction perpendicular to scrolling (e.g. width in vertical
+    -- scrolling) is not fully covered by any of the widgets, we can add our
+    -- scrollbar width to that value. Otherwise widget size will be reduced
+    -- during `layout` to make space for the scrollbar.
+    if need_scrollbar
+        and (
+            (is_y and used_max < orig_width)
+            or (not is_y and used_max < orig_height)
+        ) then
+        used_max = used_max + scrollbar_width
```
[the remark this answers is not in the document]
sclu1034: Then

https://github.com/awesomeWM/awesome/blob/fda950d186efdc43a64c4300a27a5e19a55c2bd7/lib/wibox/widget/base.lua#L542-L545

With that in mind, I don't even need the `and (...)` check.

lib/wibox/layout/overflow.lua
psychon: If clipping of child widgets is something that is needed... well, I guess it might (in theory) be possible to extend `wibox.widget.base.place_widget_*` with a `shape` parameter.

However, the result can easily end up being ugly: Doing hit tests against a `gears.shape` requires creating a temporary cairo surface and context, applying the shape, and then checking with something like `cr:in_fill(x, y)`. This sounds like a "nice" way to end up with high memory usage and/or slow code.

sclu1034: I don't think we would need a separate clip shape per child widget. A single method `widget.clip_shape` on the layout/parent should be enough.

And the special case of `widget.clip_shape == gears.rectangle` (or by checking an additional property) could then be optimized to skip the temporary cairo surface and use the patch linked above.

lib/wibox/layout/overflow.lua
actionless: could it be imposed some limitation against multiple instances? (so it will throw a warning and won't attempt to draw more than one instance)

sclu1034: That would certainly be an option and I would be fine with imposing such a limitation.
But this would certainly be the first widget to do so, although looking at the rework of `wibox.widget.graph` it's not the only one that would benefit from adding widgets with this kind of limitation.

actionless: arent wibox/wibar/popup also could have only one instance of a given geometry?

psychon: > could it be imposed some limitation against multiple instances? 

Not really. At least I wouldn't know how.

The systray widget has the same limitation. The current code basically works on hopes and dreams: Wherever the systray widget was drawn last is where it is visible.

There is some special magic in `wibox.drawable` in the (undocumented on purpose) method `_set_systray_widget`. This calls `wibox.hierarchy.count_widget` which makes the hierarchy count the number of systray widgets in hierarchies. When re-computing the layout, the count is checked before/after and if it goes to zero, a special (also undocumented) C function is called to really hide the systray.

So, a lot of work just to make sure hiding the systray widget really hides it. This is not directly related to this issue, but should just show that I already had lots of problems making the systray work.

actionless: wait, but for systray we have `container.only_on_screen()`

i guess, that shouldnt be much harder to implement `container.display_only_once` or so?

psychon: That container works via the context argument that is passed in to widgets. When `context.screen` is the expected screen, it works normally. Otherwise, `:fit` returns `0, 0` and `:layout` does nothing. Anyway, you can still show a widget multiple times on a given screen.

(And I guess this widget does not work correctly when used in a titlebar and you move the client to another screen. Or when used in a wibox and that wibox is moved to another screen. Nothing invalidates the caches or triggers a relayout in these cases.)

actionless: i think we could create the new container to accomodate this case, by just storing which parent widget first displayed it, and ignore all other fit/layout request from different parent widgets?

actionless: however, from the second thought, i feel like it should be acomodated here in overflow container - just have internal table for storing data for multiple parent widgets

actionless: like

```lua
    self._private.instances[parent_id].avail_in_dir = avail_in_dir
    self._private.instances[parent_id].used_in_dir = used_in_dir
```

HumblePresent:
I have done some testing with this layout and have encountered a couple issues/unexpected behavior.
1. A widget like the one below will not be shown in the overflow layout but is shown in a normal fixed layout
```lua
local test_layout = wibox.layout.overflow.vertical()
local test_constraint = wibox.widget {
    test_layout,
    strategy = "exact",
    height = 600,
    widget = wibox.container.constraint
}

local test_widget = wibox.widget {
    {
        {
            {
                text = "Testing overflow layout",
                widget = wibox.widget.textbox
            },
            {
                text = "Testing overflow layout",
                widget = wibox.widget.textbox
            },
            layout = wibox.layout.fixed.vertical
        },
    },
    strategy = "exact",
    width = 300,
    widget = wibox.container.contraint
}

test_layout:insert(1, test_widget)
```
Interestingly, this slighty modified widget **will** be displayed
```lua
local test_layout = wibox.layout.overflow.vertical()
local test_constraint = wibox.widget {
    test_layout,
    strategy = "exact",
    height = 600,
    widget = wibox.container.constraint
}

local test_widget = wibox.widget {
    {
        {
            text = "Testing overflow layout",
            widget = wibox.widget.textbox
        },
    },
    strategy = "exact",
    width = 300,
    widget = wibox.container.contraint
}

test_layout:insert(1, test_widget)
```
2. If the layout is constrained by a maximum size rather than exact, the scrollbar will sometimes be briefly visible when a widget is added to the layout as the layout expands with the added widgets until overflow. After overflow the scrollbar is always visible so the issue is mitigated. From my own debugging this appears to be caused by the way the `fit()` and `layout()` methods are called each time a widget is added. Here is a concrete example.
```lua
local test_layout = wibox.layout.overflow.vertical()
local test_constraint = wibox.widget {
    test_layout,
    strategy = "max",
    height = 600,
    widget = wibox.container.constraint
}

local test_widget = wibox.widget {
    {
        {
            text = "Testing overflow layout",
            widget = wibox.widget.textbox
        },
    },
    strategy = "exact",
    width = 300,
    height = 200,
    widget = wibox.container.contraint
}

-- Link to a key mapping or other event to insert multiple widgets
test_layout:insert(1, test_widget)
```
Here are some of the logs from inserting a widget into the overflow layout
```
Call to `overflow:fit()`
2021-12-16 23:09:14 W: awesome: orig_width: 9999, orig_height: 469
2021-12-16 23:09:14 W: awesome: used_max: 0, used_in_dir: 0

Call to `overflow:layout()`
2021-12-16 23:09:14 W: awesome: orig_width: 1, orig_height: 1
2021-12-16 23:09:14 W: awesome: used_in_dir: 156, avail_in_dir: 1
2021-12-16 23:09:14 W: awesome: Need scrollbar: true

Call to `overflow:layout()`
2021-12-16 23:09:14 W: awesome: orig_width: 234, orig_height: 156
2021-12-16 23:09:14 W: awesome: used_in_dir: 156, avail_in_dir: 156
2021-12-16 23:09:14 W: awesome: Need scrollbar: false
```
Due to the first call to `overflow:layout()` causing the scrollbar to be placed in the layout but then being removed right after by the second call there is an effect where the scrollbar "flashes" on the screen. The scope of this issue might be beyond the overflow layout and also have to do with the widget hierarchy but I included the info here anyway.

sclu1034:
**1.**
I'm not quite sure what's supposed to be going on here. Your layout is set to `height = 600`, the child widget to `height = 300`. `300 < 600`, so there is no need for overflow behavior, and the layout will "fall back" to being a regular `wibox.layout.fixed`.
So it sounds like this is working as intended.

**2.**
```
Call to `overflow:fit()`
2021-12-16 23:09:14 W: awesome: used_max: 0, used_in_dir: 0
```
Is this really logged _after_ the child widget has been added? And is it logged [at the end of `:fit`](https://github.com/sclu1034/awesome/blob/feature/overflow_container/lib/wibox/layout/overflow.lua#L73)?
`overflow:fit` shouldn't return `0, 0` when there is a child widget with a size > 0.

And this line looks weird as well
```
Call to `overflow:layout()`
2021-12-16 23:09:14 W: awesome: orig_width: 1, orig_height: 1
```
The overflow layout merely emits `widget::layout_changed` on `:add`/`:insert`. There's something going on in the widget system that calls `:layout` with `(1, 1)`. Possibly related to `wibox.container.constrained` and the different values for `strategy`, as you indicated.

HumblePresent:
For 1 the height of the child widget is actually unbounded, the width is set to 300. The issue is not about overflow but that the child widget is not drawn in the layout at all. If I change the layout to a `wibox.layout.fixed.vertical` it **is** drawn so I would expect `wibox.layout.overflow.vertical` to also display it.

For 2 you are right I had the logs in the wrong spot. Here is the corrected version
```
Call to `overflow:fit()`
2021-12-17 09:41:38 W: awesome: num_widgets: 1
2021-12-17 09:41:38 W: awesome: orig_width: 9999, orig_height: 469
2021-12-17 09:41:38 W: awesome: used_max: 234, used_in_dir: 156

Call to `overflow:layout()`
2021-12-17 09:41:38 W: awesome: orig_width: 1, orig_height: 1
2021-12-17 09:41:38 W: awesome: used_in_dir: 156, avail_in_dir: 1
2021-12-17 09:41:38 W: awesome: Need scrollbar: true

Call to `overflow:layout()`
2021-12-17 09:41:38 W: awesome: orig_width: 234, orig_height: 156
2021-12-17 09:41:38 W: awesome: used_in_dir: 156, avail_in_dir: 156
2021-12-17 09:41:38 W: awesome: Need scrollbar: false
```
I agree the issue for 2 seems to be related to the widget system hierarchy rather than this layout implementation.

sclu1034:
**1.**
Thanks for the elaboration. The problem seems to be that `wibox.layout.fixed:fit` subtracts widget size from `orig_height` to get the total used, rather than adding from `0`.
And in this case, in [`orig_height - height_left`](https://github.com/awesomeWM/awesome/blob/master/lib/wibox/layout/fixed.lua#L387), both values are `math.huge` (subtracting from Infinity is still Infinity), and the widget reports a height of `0`.

sclu1034:
Should now be fixed, see the added test case in `overflow_spec.lua`.

HumblePresent:
So that does seem resolve issue 1 but I think the changes to `hierarchy.lua` and `fixed.lua` have broken some of my other widgets. Parts of my notification widgets that use `wibox.layout.fixed` no longer display.

sclu1034:
The changes to `hierarchy.lua` don't affect rendering. That's about restricting click event target detection and due to the condition, the behaviour changes only for the Overflow layout.
For `fixed.lua`, I would need details. The test suite suggests no change in the layout's behaviour.

HumblePresent:
Ok I think the issue presents when one or more of the child widgets in a `wibox.layout.fixed` layout has the `visible` property set the `false`. I noticed the issue in my notification widget so that is my test case. Here is the function I used to generate a notification widget template.
```lua
local function create_template(n)
    local template = {
        {
            {
                {
                    text = n.app_name,
                    widget = wibox.widget.textbox
                },
                {
                    text = n.title,
                    widget = wibox.widget.textbox
                },
                {
                    text = n.message,
                    widget = wibox.widget.textbox
                },
                layout = wibox.layout.fixed.vertical
            },
            bg = beautiful.bg_normal,
        },
        strategy = "max",
        height = 275,
        width = 375
    }
    return template
end
```
If I then generate a notification it will be properly displayed according to the template above. If I set the `visible` property of one of the texboxes in the vertical fixed layout to `false` the notification will not be displayed.
```lua
local function create_template(n)
    local template = {
        {
            {
                {
                    text = n.app_name,
                    visible = false,
                    widget = wibox.widget.textbox
                },
                {
                    text = n.title,
                    widget = wibox.widget.textbox
                },
                {
                    text = n.message,
                    widget = wibox.widget.textbox
                },
                layout = wibox.layout.fixed.vertical
            },
            bg = beautiful.bg_normal,
        },
        strategy = "max",
        height = 275,
        width = 375
    }
    return template
end
```
I also verified that this issue is only present after your latest commit 51424b5

HumblePresent:
I've looked through the code a bit and I believe the issue is caused by the changes in the `wibox.layout.fixed:fit()` method. While iterating through the widgets in the layout, if the call to `base.fit_widget()` returns zero for the dimension in the direction of the layout, the iteration ends and `fit()` returns. The problem is that `base.fit_widget()` returns `0, 0` for any widget whose `visible` property is false, so any widgets in the layout after a widget that is not visible will be ignored.

lib/wibox/layout/fixed.lua
```diff
@@ -333,61 +333,69 @@ end
...
-                 width_left = k < widgets_nr and width_left + spacing or width_left
-                 width_left = width_left < 0 and 0 or width_left
+        local w, h = base.fit_widget(self, context, v, width_left, height_left)
+
+
+        -- Determine if the widget still fits
+        local is_enough
+
+        if is_y then
+            is_enough = h > 0 and height_left >= h
```
HumblePresent, line 367:
```suggestion
            is_enough = height_left >= h
```
There may be a good reason for the `h > 0` condition needing to be true that I can't think of right now, but this does fix the issue for a widget with `visible = false`.

lib/wibox/layout/fixed.lua
```diff
@@ -333,61 +333,69 @@ end
...
+
+            if is_enough then
+                used_max = math.max(used_max, w)
+                used_in_dir = used_in_dir + h
+                height_left = height_left - h
             end
+        else
+            is_enough = w > 0 and width_left >= w
```
HumblePresent, line 375:
```suggestion
            is_enough = width_left >= w
```

lib/wibox/layout/overflow.lua
```diff
@@ -0,0 +1,549 @@
...
+    local factor = self._private.scroll_factor + (delta * amount)
+    self:set_scroll_factor(factor)
+end
+
+
+--- The scroll factor.
+--
+-- The scroll factor represents the how for the layout's content is currently
+-- scrolled. It is represented as a fraction from `0` to `1`, where `0` is the
+-- start of the content and `0` is the end.
```
Aire-One, line 305: I'm not sure to understand the description here. Can you make it simpler? Something like "The scroll factor represents the layout's current scrolled position.".

Also, I think it's "where `0` is the start of the content and **`1`** is the end."

lib/wibox/hierarchy.lua
```diff
@@ -143,12 +143,14 @@ function hierarchy_update(self, context, widget, width, height, region, matrix_t
 
     -- Calculate the draw extents
     local x1, y1, x2, y2 = 0, 0, width, height
-    for _, h in ipairs(self._children) do
-        local px, py, pwidth, pheight = matrix.transform_rectangle(h._matrix, h:get_draw_extents())
-        x1 = math.min(x1, px)
-        y1 = math.min(y1, py)
-        x2 = math.max(x2, px + pwidth)
-        y2 = math.max(y2, py + pheight)
+    if not widget.clip_child_extends then
```
Aire-One, line 146: If this property has to be canon and used in common widget code base, I'd like it to be documented as such. Can you add it to the common widget doc?

lib/wibox/layout/overflow.lua
```diff
@@ -0,0 +1,549 @@
...
+        end
+    end
+
+    used_in_dir = used_in_dir + self._private.spacing * (#widgets-1)
+
+    -- Save size for scrolling behavior
+    self._private.avail_in_dir = avail_in_dir
+    self._private.used_in_dir = used_in_dir
+
+    local need_scrollbar = used_in_dir > avail_in_dir and scrollbar_enabled
```
Aire-One, line 123: Does that mean that if the content is smaller than the available space, there is no scrollbar even if we explicitly set `scrollbar_enabled = true`? Is there a way to force scrollbar?

lib/wibox/hierarchy.lua
sclu1034: I could, but I don't think that would make sense. The API docs document the public, user-facing API. But `clip_child_extends` is internal to the widget system.
It's more akin to `:fit` or `:draw` in that regard.

Either way, I think we should first see what the others think about the functionality/implementation before looking into if/how/where it should be documented.

lib/wibox/layout/overflow.lua
sclu1034: > there is no scrollbar even if we explicitly set `scrollbar_enabled = true`?
Yes

> Is there a way to force scrollbar?
No

sclu1034:
> The problem is that `base.fit_widget()` returns `0, 0` for any widget whose `visible` property is `false`

Oh, that's bad. I didn't realize until now, but `fit_widget` actually encodes multiple different things into `0, 0`.

When spacing is involved, it actually makes a difference whether a widget is `visible = false` and both the widget and the spacing should be skipped, or if the widget only reports `0, 0` due to some internal state but should still be shown.
Similarly, if the layout doesn't have enough space to fit all children, currently the only way a child widget can report "that's not enough space for me" is, again, returning `0, 0`. But that's the condition where `wibox.layout.fixed` should `break` the loop.

HumblePresent:
>currently the only way a child widget can report "that's not enough space for me" is, again, returning 0, 0

Can't child widgets return a size beyond what's available in the layout to indicate there isn't enough space? I thought `fit()` was supposed to return a preferred size even if that's not what they end up receiving.

Elv13:
I don't know if it's relevant, but that reminds me of that bug https://github.com/awesomeWM/awesome/commit/83c31f948b57f3dfb75d343a084ef1509d740902

The 0x0 widgets are tricky to handle...

sclu1034:
> Can't child widgets return a size beyond what's available in the layout to indicate there isn't enough space? I thought `fit()` was supposed to return a preferred size even if that's not what they end up receiving.

Sure, that might be slightly less ambiguous. But that's still using a return value of two numbers that could be valid to encode information that shouldn't be encoded with numbers. And it wouldn't help at all with the `visible = false` part.

sclu1034:
I've extracted the discussion about `base.fit_widget` into a separate issue, since this affects all layouts.

HumblePresent:
I have another instance of unexpected behavior. I know I'm a bit of a thorn in the side of this PR right now, but better to work these things out now rather than later when someone creates an issue right :grin:.

When I click on the scrollbar without moving the mouse the scrollbar moves and the scroll position changes. As soon as I move the mouse a little the scrollbar and and scroll position seem to snap back to their original location before being clicked on. I would expect the scrollbar to behave in a typical manner where it does not move until the mouse is moved.

HumblePresent:
Just tested again and that last change fixed my issue.

Aire-One (approved):
@sclu1034, do you think it's now ready? I think we are alright. 

I'd like to have the input from other maintainers about the new `clip_child_extends`, tho. I really think it should be documented. Worst case scenario : it will be marked as "inherited from widget.base" and the user will ignore it, like they already ignore "children" on most widgets.

Also, to mitigate the concern @psychon pointed that this layout can only be used at one place at a time, as opposed to how the widget system should work. Maybe we should have a mention of this in the module description. (It can be a good excuse to implement some sort of information/alert component in the doc https://getbootstrap.com/docs/4.0/components/alerts/)

sclu1034:
> do you think it's now ready?

That would be up to the reviewers to decide.
It's still affected by #3531, so it's "only" as much broken as all the other layouts.

> it will be marked as "inherited from widget.base" and the user will ignore it, like they already ignore "children" on most widgets.

That's not a good comparison.
`.children` is a _user-facing_ API (changing child widgets is the whole point of a layout, after all).
However, `.clip_child_extends` is **not** a user-facing API, just like `:fit` or `:layout`. A consumer of a widget should not change that property on the widget, only the widget's' implementation.
So, just like how `:fit` doesn't show up in the "Object methods" section, `clip_child_extends` should not show up in the "Object properties" section.

Would prefixing it with `_` help with marking it as private?

Elv13 (dismissed):

Elv13:
Sorry for the delay, I didn't work on AwesomeWM since xmas. I did some testing and it works. My earlier comments are addressed and the API looks good. Thanks a lot for this.

Now, yes, this container has some unfixeable limitations, but it provides a lot of value for people who don't use the more advanced widget system features. Thanks a lot for this contribution!

HumblePresent:
Before this gets merged I do want to point out that the changes to `fixed.lua` do still break fixed layouts with any child widgets that have `visible = false`. By break I mean widgets that come after an "invisible" widget in the layout will not be shown regardless of their `visible` property. Is it possible to merge this PR without the changes to `fixed.lua` and make any necessary changes after #3531 has been resolved?

sclu1034:
As far as I can see, we have three options:
- Get #3531 fixed, which the current issue here blocks on
- Revert changes and reintroduce the bug described as `1.` [above](https://github.com/awesomeWM/awesome/pull/3309#issuecomment-996450616) (`wibox.layout.overflow` is somewhat broken)
- Keep the current bug with stopping on the first `visible = false` (regression in `wibox.layout.fixed`)

Ideally, we would do the first. However, if you do want this PR merged _now_, the second option would be preferable.

TorchedSammy:
after using this myself i have noticed 2 problems:
- the doc entry is wrong for the vertical layout (https://safe.kashima.moe/qi0g6vvj0jq7.png)
- `:reset()` does not reset the scroll position

havent used it enough to notice any other problems, besides maybe the default step value being too low in my opinion

Crylia:
I noticed when using two overflow containers (one inside another one) that both get scrolled, when scrolling with the mouse wheel. When scrolling you should probably check which is the nearest overflow widget under the cursor and only scroll that one.

TorchedSammy:
some other things i recommend adding:
- background for the scrollbar widget
- spacing between contents and scrollbar (an easier way to do that)

sclu1034:
For now this is fully blocked on #3531 .

sclu1034:

Issue #3324, opened 2021-04-11
Reject builds against Lua 5.4


psychon:
Awesome should check at startup if it is running against Lua 5.4 and provide a good error message. According to #3319, the current error message is quite bad:
```
2021-04-06 08:41:59 W: awesome: luaA_panic:855: unprotected error in call to Lua API (bad argument #1 to 'math.randomseed' (number has no integer representation))
```
Bonus points for doing this check at build time, but since it is possible that `lua` and `liblua.so` are different Lua versions, it's not completely trivial. One needs to build a small test program and link it against the lua library found by CMake *and* run the program.

ShayAgros:
Can you elaborate on what you mean ?

I can fail the build of the C files using static asserts on the macros in the `lua.h` file.

for example in lua5.3 has these macros in its lua.h header:
```
#define LUA_VERSION_MAJOR	"5"
#define LUA_VERSION_MINOR	"3"
#define LUA_VERSION_NUM		503
#define LUA_VERSION_RELEASE	"3"
``` 
so you can add this line to `awesome.c`:
```
static_assert(LUA_VERSION_MINOR < 4, "Awesome doesn't support lua version newer than 5.3");
```

From inside the interpreter I can fish the version by running `print(_VERSION)` (at least it seems to work for 5.3 and 5.2).

You want to prevent awesome from compiling in lua5.4 ?

psychon:
Is `static_assert` part of C? Somehow that feels "not C99 nor C11" to me.

> for example in lua5.3 has these macros in its lua.h header:

Yeah, problem is: Those are strings. It's `"3"` and not `3`. I doubt that `"3" < 4` works.

But I totally missed `LUA_VERSION_NUM`. That one can actually be compared.

> You want to prevent awesome from compiling in lua5.4 ?

Yes. See #3319. Right now, AwesomeWM apparently compiles, but then fails to start. In that specific case this was with no error message and the symptom was "black screen". Failing too build is way better user experience than "does not work and brooke everything" (and results in hopefully less bug reports and those bug reports that do come in would be a lot more obvious).

ShayAgros:
Yup I see what you mean. The static_assert is part of c11 standard.

In any rate the we could uise something like
```
#if LUA_VERSION_NUM >= 540
/* fail somehow */
#endif
```

I'll play around with it and submit a PR about to fail the compilation.

PR #3279, opened 2021-03-06
Make CMake Lua executable customizable

sclu1034:
Some platforms, such as Arch Linux, already moved to Lua 5.4, while
offering Lua 5.3 as a separate executable, such as `/usr/bin/lua5.3`.
To be able to build awesomeWM on these platforms without extensive
shims, this change introduces a new CMake variable `LUA_EXECUTABLE`.

Its default is set by `find_program` to the usual `/usr/bin/lua`,
but allows running CMake like this:

```sh
cmake ../ \
    -DLUA_INCLUDE_DIR=/usr/include/lua5.3 \
    -DLUA_LIBRARY=/usr/lib/liblua.so.5.3 \
    -DLUA_EXECUTABLE=/usr/bin/lua5.3
```

PR #3325, opened 2021-04-13
Fail awesome compilation for Lua5.4 or newer (#3324)

ShayAgros:
Since Awesome doesn't support these versions, preventing the project from
compiling with it saves the users some very frustrating and hard to
debug bugs.

The cmake version check is redundant with the version check in luaa.h.
Nevertheless, adding it allows to fail the project build sooner and
provides better user experience.

Fixes: #3324

### DIFF
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -71,6 +71,13 @@ if (NOT LUA_FOUND)
         "You might want to hint it using the LUA_DIR environment variable, "
         "or set the LUA_INCLUDE_DIR / LUA_LIBRARY CMake variables.")
 endif()
+
+set(LUA_FULL_VERSION "${LUA_VERSION_MAJOR}.${LUA_VERSION_MINOR}.${LUA_VERSION_PATCH}")
+if(LUA_FULL_VERSION VERSION_EQUAL 5.4.0 OR LUA_FULL_VERSION VERSION_GREATER 5.4.0 )
+    message(FATAL_ERROR "Awesome doesn't support Lua versions newer than 5.3, please refer to"
+                        "https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
+endif()
+
 # }}}
 
 # {{{ Check if documentation can be build
--- a/awesomeConfig.cmake
+++ b/awesomeConfig.cmake
@@ -73,8 +73,9 @@ if (NOT LUA_FOUND)
 endif()
 
 set(LUA_FULL_VERSION "${LUA_VERSION_MAJOR}.${LUA_VERSION_MINOR}.${LUA_VERSION_PATCH}")
-if(LUA_FULL_VERSION VERSION_EQUAL 5.4.0 OR LUA_FULL_VERSION VERSION_GREATER 5.4.0 )
-    message(FATAL_ERROR "Awesome doesn't support Lua versions newer than 5.3, please refer to"
+# 5.1 <= LUA_VERSION < 5.4
+if(NOT ((LUA_FULL_VERSION VERSION_EQUAL 5.1.0 OR LUA_FULL_VERSION VERSION_GREATER 5.1.0) AND LUA_FULL_VERSION VERSION_LESS 5.4.0))
+    message(FATAL_ERROR "Awesome only supports Lua versions 5.1-5.3, please refer to"
                         "https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
 endif()
 
--- a/luaa.h
+++ b/luaa.h
@@ -32,7 +32,7 @@
 #include "common/luaclass.h"
 
 #if !(501 <= LUA_VERSION_NUM && LUA_VERSION_NUM < 504)
-print("Awesome only supports Lua versions 5.1-5.3, please refer to, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
+print("Awesome only supports Lua versions 5.1-5.3, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
 #endif
 
 #define luaA_deprecate(L, repl) \
--- a/luaa.h
+++ b/luaa.h
@@ -32,7 +32,7 @@
 #include "common/luaclass.h"
 
 #if !(501 <= LUA_VERSION_NUM && LUA_VERSION_NUM < 504)
-print("Awesome only supports Lua versions 5.1-5.3, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
+#error "Awesome only supports Lua versions 5.1-5.3 and LuaJIT2, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building"
 #endif
 
 #define luaA_deprecate(L, repl) \
--- a/luaa.h
+++ b/luaa.h
@@ -31,8 +31,8 @@
 #include "common/lualib.h"
 #include "common/luaclass.h"
 
-#if LUA_VERSION_NUM > 503
-print("Awesome doesn't support Lua versions newer than 5.3, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
+#if !(501 <= LUA_VERSION_NUM && LUA_VERSION_NUM < 504)
+print("Awesome only supports Lua versions 5.1-5.3, please refer to, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
 #endif
 
 #define luaA_deprecate(L, repl) \
--- a/luaa.h
+++ b/luaa.h
@@ -31,6 +31,10 @@
 #include "common/lualib.h"
 #include "common/luaclass.h"
 
+#if LUA_VERSION_NUM > 503
+print("Awesome doesn't support Lua versions newer than 5.3, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
+#endif
+
 #define luaA_deprecate(L, repl) \
     do { \
         luaA_warn(L, "%s: This function is deprecated and will be removed, see %s", \


Elv13 (dismissed):

awesomeConfig.cmake
```diff
@@ -71,6 +71,13 @@ if (NOT LUA_FOUND)
         "You might want to hint it using the LUA_DIR environment variable, "
         "or set the LUA_INCLUDE_DIR / LUA_LIBRARY CMake variables.")
 endif()
+
+set(LUA_FULL_VERSION "${LUA_VERSION_MAJOR}.${LUA_VERSION_MINOR}.${LUA_VERSION_PATCH}")
+if(LUA_FULL_VERSION VERSION_EQUAL 5.4.0 OR LUA_FULL_VERSION VERSION_GREATER 5.4.0 )
```
blueyed, line 76: Note that there's `VERSION_GREATER_EQUAL`, which requires CMake 3.7 though, and we only require 3.0 currently.  So good to have it more verbose/expanded here.

awesomeConfig.cmake
```diff
@@ -71,6 +71,13 @@ if (NOT LUA_FOUND)
...
+    message(FATAL_ERROR "Awesome doesn't support Lua versions newer than 5.3, please refer "
```
blueyed, line 77:
```suggestion
    message(FATAL_ERROR "Awesome doesn't support Lua versions newer than 5.3, please refer to "
```

luaa.h
```diff
@@ -31,6 +31,10 @@
 #include "common/lualib.h"
 #include "common/luaclass.h"
 
+#if LUA_VERSION_NUM > 503
+print("Awesome doesn't support Lua versions newer than 5.3, please set refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
```
blueyed, line 35:
```suggestion
print("Awesome doesn't support Lua versions newer than 5.3, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
```

blueyed (commented):
Not tested, but LGTM - apart from string/message changes.

Aire-One (commented):
LGTM too!

Regarding your concern of doing the check twice, I think it's even required. We can't guarantee the lua version used to build awesome is the same as the one used at the runtime. So both contexts have to be checked separately anyway. 

One question, tho: I think these checks would feel more natural the other way. By testing if the lua version is **not** something we support. 
I mean, instead of `LUA >= 5.4`, do `NOT( LUA < 5.4 )`? 

(Even better, `NOT( LUA < 5.1 AND LUA < 5.3 )`. This way, we can be sure the lua provided is something we tested awesome against in our CI.) 

BTW, I guess we also need to update the dependencies list from `docs/10-building-and-testing.md`.

sclu1034:
> `NOT( LUA < 5.1 AND LUA < 5.3 )`

That would be equal to `NOT(LUA < 5.1)` or `LUA >= 5.1` and would consider `5.4` as supported version.
I guess you meant something like `is_supported_version = LUA >= 5.1 AND LUA <= 5.3`.

Aire-One:
> > `NOT( LUA < 5.1 AND LUA < 5.3 )`
> 
> That would be equal to `NOT(LUA < 5.1)` or `LUA >= 5.1` and would consider `5.4` as supported version.
> I guess you meant something like `is_supported_version = LUA >= 5.1 AND LUA <= 5.3`.

Derp, I got it all wrong. I should have done a double check before sending the message...

Thank you for fixing it ☺

ShayAgros:
Here I do the same mistake I did before again. In a effort to test some changes in the pull request I accidentally update the PR. Please ignore. I'll update once it is ready. Sorry for noise (again ..)

awesomeConfig.cmake
ShayAgros: Yup, I thought it'd be better not to assume cmake version, especially that it doesn't make the code much longer

ShayAgros:
Ok, the new version is ready for review.

Tested that compilation and cmake fail with lua5.4 (tested on arch), and  ran the patch against Lucas' CI patchset:
https://github.com/ShayAgros/awesome/runs/2397742104

It passes for all supported Lua versions (well technically it fails, but for other reasons unrelated to this patch). Please re-review

luaa.h
```diff
@@ -31,6 +31,10 @@
 #include "common/lualib.h"
 #include "common/luaclass.h"
 
+#if !(501 <= LUA_VERSION_NUM && LUA_VERSION_NUM < 504)
+print("Awesome only supports Lua versions 5.1-5.3, please refer to, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
```
sclu1034, line 35:
```suggestion
print("Awesome only supports Lua versions 5.1-5.3, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
```

sclu1034 (changes requested):

sclu1034 (approved):

luaa.h
```diff
@@ -31,6 +31,10 @@
 #include "common/lualib.h"
 #include "common/luaclass.h"
 
+#if !(501 <= LUA_VERSION_NUM && LUA_VERSION_NUM < 504)
+print("Awesome only supports Lua versions 5.1-5.3, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building")
```
Elv13, line 35: I think this should be 

```
fprintf(stderr, "Awesome only supports Lua versions 5.1-5.3, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building");
```

Elv13 (changes requested):

luaa.h
ShayAgros: I don't think it matters really. This function is called outside a function scope. The compiler will fail compilation no matter what this function is called. I called it `print` just to signal in some way that we want the user to read it.

Searching through StackOverflow I couldn't find a way to fail compilation in build time in C w/o doing some trick like this.
c11 has `static_assert` function which looks better, but I thought it's not worth it to impose requirement for c11 for awesome just to use static_assert

Elv13: `#error`?

ShayAgros: Yup, forgot about this one. According to its [documentation](https://gcc.gnu.org/onlinedocs/cpp/Diagnostics.html) this can work.

How about something like this:
```suggestion
#error "Awesome only supports Lua versions 5.1-5.3 and LuaJIT2, please refer to https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building"
```

Aire-One (approved):

Aire-One:
> Regarding your concern of doing the check twice, I think it's even required. We can't guarantee the lua version used to build awesome is the same as the one used at the runtime. So both contexts have to be checked separately anyway. 

(As a quick note to myself, the check in `luaa.h` is not runtime related at all...)

Elv13 (approved):

sclu1034 (approved):

basaran:
I'm having trouble building on Lua 5.4 after specifying the -D args

```
CMAKE_ARGS="-DLUA_EXECUTABLE=/usr/bin/lua5.3" make
```

then it returns

```
-- lua -> /usr/bin/lua5.3
-- git -> /usr/bin/git
-- asciidoctor -> /usr/bin/asciidoctor
-- gzip -> /usr/bin/gzip
-- ldoc not found.
-- ldoc.lua not found.
-- Not generating API documentation. Missing: ldoc.
-- convert -> /usr/bin/convert
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.7.3")
-- Found Lua: /usr/lib/liblua5.4.so;/usr/lib/libm.so (found version "5.4.3")
CMake Error at awesomeConfig.cmake:78 (message):
  Awesome only supports Lua versions 5.1-5.3, please refer
  tohttps://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html#Building
  CMakeLists.txt:33 (include)


See also "/home/michael/temp/awesome/build/CMakeFiles/CMakeOutput.log".
make: *** [Makefile:27: build/Makefile] Error 1
```

It seems it still is trying to use the default lua on the system which is 5.4

```
-- Found Lua: /usr/lib/liblua5.4.so;/usr/lib/libm.so (found version "5.4.3")
```

actionless:
@basaran 

```
  export _LUA_VER=5.3
  mkdir -p build
  cd build
  cmake ../$pkgname \
    -DCMAKE_BUILD_TYPE=RELEASE \
    -DCMAKE_INSTALL_PREFIX=/usr \
    -DSYSCONFDIR=/etc \
    -DLUA_INCLUDE_DIR=/usr/include/lua${_LUA_VER} \
    -DLUA_LIBRARY=/usr/lib/liblua.so.${_LUA_VER} \
    -DLUA_EXECUTABLE=/usr/bin/lua${_LUA_VER}
  make
  # make install
```

basaran:
Thank you, appeared to work fine after symlinking:

```
sudo ln -s /usr/lib/liblua.so.5.3.6 /usr/lib/liblua.so.5.3   
```

basaran:
is there anything else I need to be doing? I copied the lib folder from the build directory onto /usr/share/awesome/lib (after backing up the one that's there) and copied the compiled binary onto /usr/local/bin and `which awesome` confirmed that's the right one.

actionless:
apparently smth went wrong during copying

basaran:
I tried to install the aur package awesome-git, it had the same issue. I will try to find what's going on.

Aire-One:
> Thank you, appeared to work fine after symlinking:
> 
> ```
> sudo ln -s /usr/lib/liblua.so.5.3.6 /usr/lib/liblua.so.5.3   
> ```

This sounds like a bad idea. You shouldn't have to do this symlink manually, your package manager should handle it. (if I get it correctly, you use arch, so I can confirm pacman does)

Since #3279 is merged, we have to explicitly set 3 cmake variables for lua :

```sh
cmake ../ \
    -DLUA_INCLUDE_DIR=/usr/include/lua5.3 \
    -DLUA_LIBRARY=/usr/lib/liblua.so.5.3 \
    -DLUA_EXECUTABLE=/usr/bin/lua5.3
```

I personally don't use the variables `-DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_INSTALL_PREFIX=/usr -DSYSCONFDIR=/etc` (as recommended by actionless) because I don't install awesome after building it... Judging by the error message you get, I'd say you forget to run `make install`? (maybe as sudo?).

> I tried to install the aur package awesome-git, it had the same issue.

This is alarming, tho.

Checking https://aur.archlinux.org/packages/awesome-git/, it seems you are the only person (or the first?) to have this issue. So either, we broke something very recently, or, there is something broken in your system (sorry).

Aire-One:
> @basaran 
> 
> ```
>   export _LUA_VER=5.3
>   mkdir -p build
>   cd build
>   cmake ../ \
>     -DCMAKE_BUILD_TYPE=RELEASE \
>     -DCMAKE_INSTALL_PREFIX=/usr \
>     -DSYSCONFDIR=/etc \
>     -DLUA_INCLUDE_DIR=/usr/include/lua${_LUA_VER} \
>     -DLUA_LIBRARY=/usr/lib/liblua.so.${_LUA_VER} \
>     -DLUA_EXECUTABLE=/usr/bin/lua${_LUA_VER}
>   make
>   # make install
> ```

Oh ! With this, you would (partially?) replace the existing installation of awesome with what's build.

I guess if @basaran didn't first uninstall the stable version of awesome, they are now in a weird situation with a broken build of awesome-git mixed-up with the stable package...

basaran:
I only used the manual symlink while trying to compile the git myself (after adding a patch to one of the c files) and using the snippet you sent over. I never ran `make install` on it, so I doubt any system files were manipulated. I simply copied stuff over manually (lib folder to lib destination, and executable to the local/bin)

Once that failed, I removed the manual symlink, removed all the stuff I moved manually and tried to install awesome-git via yay. That worked no issues, and compiled fine (I think). After restarting the lightdm, I got the same xsession-errors.

I will try again later in the day, I agree it's odd. I'm using manjaro's awesome distro as the base, but changed the kernel to ck-zen. That would be the only non sub standard thing.

Aire-One:
You don't necessary have to install awesome to start it after a build. To use the version you built, just run the binary from the build directory and specify the location of the generated awesome lua libraries with `--search pathto/build/lib`. 

(https://awesomewm.org/doc/manpages/awesome.1.html)

basaran:
Thank you, after trying the replace method you advised, I noticed something on the console. 

Obviously, at some point last year I had copied the "wibox" lib directly onto ~/.config/awesome for an unknown reason possibly to override something in there without touching the actual lib, and that was the culprit. It was overriding all this time.

yay awesome-git compiled and ran just fine once I removed the stale folder. Valuable lesson learned, sorry for taking everybody's time. I will commence with the manually compiled kernel in the evening to test psychon's patch.

Issue #3289, opened 2021-03-18
graph: fix min_value with scaling


blueyed:
In https://github.com/awesomeWM/awesome/commit/68999f4a86df5f8d23d7bf06939d45985f052411 support for `min_value` was added (to fix #277).

However, I think the default for it is confusing / not wanted when `scale` is set: it will use `math.huge` by default then, scaling/adjusting it to the lowest value of the graph (https://github.com/awesomeWM/awesome/commit/68999f4a86df5f8d23d7bf06939d45985f052411#diff-a7fbbb267f5d802c8978befe19471ea94047d665e04c210cf2daf710d4860756R139-R141).

But I think it should default to `max_value - widget_height` then, so that the graph will be filled, instead of being empty with only a single set of minimal values.

Notes:

- the doc says "Note that the min_value is not supported when used along with the stack property." (https://github.com/awesomeWM/awesome/commit/68999f4a86df5f8d23d7bf06939d45985f052411#diff-a7fbbb267f5d802c8978befe19471ea94047d665e04c210cf2daf710d4860756R47-R48), but then does something for stacked graphs (https://github.com/awesomeWM/awesome/commit/68999f4a86df5f8d23d7bf06939d45985f052411#diff-a7fbbb267f5d802c8978befe19471ea94047d665e04c210cf2daf710d4860756R108-R110)!? => should get revisited
- with `max_value` it mentions the behavior with regard to "scale". This should be revisited (s/below/above/ ?), and also added for `min_value` then

Output of `awesome --version`: awesome v4.3-900-g4319b1611-dirty (Too long)

Elv13:
This was fixed a while back

Issue #2899, opened 2019-10-09
Question about wibox.widget.graph


deb75:
Output of `awesome --version`:

awesome v4.3 (Too long)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

Hello,

I would like to use `wibox.widget.graph` to display several functions, like plotting two lines on the same 2D graph for two functions y=f(x) and y = g(x). However, the few screenshots of `wibox.widget.graph` let me think it can only draw bar graphs, instead of just a line, and so that it cannot plot several functions.

Is there a way to perform what I want with `wibox.widget.graph` or another widget ?

Regards

psychon:
I'm not sure how visible just a line would be.

Did you see the `stack` feature of `wibox.widget.graph`? But I guess this also does not do what you want.

Another option would be writing your own widget. It is not as complicated as one might think.

Where do the functions come from? Are these lines like memory usage or what do you have in mind?

Elv13:
Well, there's https://github.com/blindFS/blingbling , but it is very, very old and way, way behind on all the API changes made in the last ~10 years. Some of the code can be salvaged, but it will take some solid effort to revive the core and implement all the API changes from AwesomeWM v3.4 (2009) to today.

psychon:
> If it is not complicated to write one's widget,

Here is some code that you can hopefully start from. I think that if you know some Lua, you should be able to manage.

This code creates a widget that draws some random lines. A timer is used to add new data to the widget. For your use case, you most likely also want to remove old data from the widget.
```lua
local my_widget = wibox.widget.base.make_widget()
my_widget._private.min_value = 0
my_widget._private.max_value = 1
my_widget._private.colors = { "#f00", "#0f0" }
my_widget._private.data = {}
function my_widget:add_data(data)
    local len = #data
    if self._private.len and self._private.len ~= len then
        error()
    end
    self._private.len = len
    table.insert(self._private.data, data)
    self:emit_signal("widget::redraw_needed")
end
function my_widget:fit(_, width, height)
    -- We just want all available space
    return width, height
end
function my_widget:draw(_, cr, width, height)
    local data = self._private.data
    local colors = self._private.colors

    -- Ensure we have at least two points of data
    if #data == 0 then
        return
    end
    if #data == 1 then
        data = { data[1], data[1] }
    end

[...]

PR #3326, opened 2021-04-16
Graph widget overhaul

albel727:
I've noticed a bug in the graph widget, which prompted me to delve into the code and discover a huge number of other bugs and missing features, and I got a bit carried away. The fact that few seemed to take note of the problems (e.g. #3289) makes me suspect, that nobody really cares or uses the kind of functionality that I'm fixing here. So I want to know if there's even an interest in merging this and get some feedback on it, before I do more stuff. 

I've tried to organize commits in the order of increasing intrusiveness and visibly changed behavior (e.g. the border_color fix), and with the exception of a pair of commits, every commit is a complete fix, i.e. one can truncate this pull series almost at any point and the widget will perform as expected (but better), so you can take as much as you want (but I'll be sad if it's not gonna be the entirety of it).

So what does this fix? Here are some commit highlights.
###  Fix the "border_width affects graph bar width" bug
A really visually horrible bug. It's a wonder that nobody noticed. Setting `border_width` property > 1 made the bars increasingly large while still overlapping each other. Before and after the fix:
AUTOGEN_wibox_widget_graph_border_width buggy svg AUTOGEN_wibox_widget_graph_border_width fixed svg

All the following "before" pictures will have this bug already fixed, otherwise it would have been sometimes hard to see, what is it that I'm fixing there.
### Fix wrong horizontal offset calculations in widget.graph
Increasing `step_spacing` property caused the graph to shift left as a whole, making the leftmost bars vanish offscreen. Before and after the fix:
2_AUTOGEN_wibox_widget_graph_step_spacing2 svg buggy 2_AUTOGEN_wibox_widget_graph_step_spacing2 svg fixed

But! That was the case only for `step_shape = nil`. But drawing with shapes had its own problem: the whole graph was shifted right for `step_width` pixels and the first bar touched the second bar despite `step_spacing`. 
In the following pics the first graph is `step_shape = nil`, and the rest are `rectangle`, `rounded_bar` and `arrow` respectively. Before and after the fix:
2_AUTOGEN_wibox_widget_graph_step_shape2 svg buggy 2_AUTOGEN_wibox_widget_graph_step_shape2 svg fixed
 
### Fix graph bar scaling when `min_value` is not 0
The code used a wrong formula which led to underscaling of values, so that even a value == `max_value` wasn't drawn to full widget height. Before and after the fix:
3_AUTOGEN_wibox_widget_graph_min_value svg buggy 3_AUTOGEN_wibox_widget_graph_min_value svg fixed

### Fix unsharp edges when drawing graph bars with rectangles
You may have noticed in "Fix wrong horizontal offset calculations in widget.graph", that the `step_shape = nil` case is drawn somehow blurry compared to the `step_shape = rectangle` case, even though it is logically the same. Well yes, there was a wrong `+0.5` offset, which should've been applied only for the special case of drawing 1px-wide bars with lines. Before and after the fix:
4_AUTOGEN_wibox_widget_graph_step_spacing2 svg buggy 4_AUTOGEN_wibox_widget_graph_step_spacing2 svg fixed

### Draw shaped graph bars only up to the baseline
Prior to this commit, every step_shape bar was, for some reason, drawn with height = full widget's height, regardless of the value it represents. So for the most bars their lower end was far outside the box and therefore clipped off. So e.g. one couldn't see the lower round edges of rounded rectangles. This commit fixes that by appropriately varying the height that is passed into `step_shape()`, so that lower edges of the bars always exactly touch the lower edge of the graph. Before and after the fix:
5_AUTOGEN_wibox_widget_graph_step_shape3 svg buggy 5_AUTOGEN_wibox_widget_graph_step_shape3 svg fixed

### Unify stacked and non-stacked graph rendering
Stacked graphs hadn't supported everything the nonstacked do, like `step_width`, `step_spacing` and custom `step_shape`-s. Now they do. Before and after the fix:
6_AUTOGEN_wibox_widget_graph_stacked svg buggy 6_AUTOGEN_wibox_widget_graph_stacked2 svg fixed

### Implement multiple data series drawing for non-stacked graph
The widget can already hold and draw multiple data groups, so it makes little sense to limit drawing only to one group when `graph.stack = false`. Here's what one can do now:
7_AUTOGEN_wibox_widget_graph_stacked_not3 svg 7_AUTOGEN_wibox_widget_graph_stacked_not svg 7_AUTOGEN_wibox_widget_graph_stacked_not2 svg

With one tiny change, which is not a part of this pull request yet, the user will be able to easily paint things with e.g. lines connecting the dots by only supplying an appropriate `step_shape`:
7_AUTOGEN_wibox_widget_graph_stacked_not4 svg
I guess this guy would've liked to have it: #2899

### Fix border_color/border_width bugs in graph widget
These are numerous, so I suggest to read the respective commit message. I'll only note here, that this is one of the commits that may visibly change the behavior for existing users' configs, because now the widget border is drawn with a fallback color, where it previously didn't, if `border_color` wasn't set. One of such users is awesome's documentation itself, where an explicitly set `border_width = 2` in some examples was ignored, because `border_color` was forgotten. Before and after this commit:
8_AUTOGEN_wibox_widget_graph_color svg buggy 8_AUTOGEN_wibox_widget_graph_color svg fixed

### Negative values support
`Add baseline_y and baseline_value properties to graph widget` and `Implement the simplest negative shape handling in graph widget` are some relevant commits. This allows to do e.g. the following now:
9_AUTOGEN_wibox_widget_graph_stacked_not_negative svg 9_AUTOGEN_wibox_widget_graph_negative_shape svg

You might have noted a specific customization challenge that this poses, e.g. how the user should go about it if they want to draw arrows for negative values pointing downwards. I have some ideas, which are waiting to be implemented (or not, if you don't want it).

### More careful value storage handling 
`Add capacity property to graph widget` and `Track graph draw() usage to guess necessary array capacity` are some relevant commits. Prior to them widget relied on its `width` property to guess how many values it should keep, and it has dutifully returned the value in the `fit()` callback, but widgets are not guaranteed to get the dimensions they want. 
Even some examples in awesome's documentation fell prey to this problem. Here's a before and after:
10_AUTOGEN_wibox_widget_defaults_graph svg buggy 10_AUTOGEN_wibox_widget_defaults_graph svg fixed

And here is, incidentally, the line that causes the difference: https://github.com/awesomeWM/awesome/blob/fda950d186efdc43a64c4300a27a5e19a55c2bd7/tests/examples/wibox/template.lua#L24

Moreover a widget could be drawn in several places with different widths simultaneously, and for a graph this means a different number of visible values, and of course the invisible values shouldn't affect e.g. the auto-scaling. This is all properly handled now. I'm not sure if the graph widget even needs the width and height properties now.

### And more
I'm running out of time to describe some smaller things like proper `NaN` handling, autoscaling fix for stacked graphs, and performance improvements (despite all those new features and fixes, the graph rendering is now 4 times faster with just one simple trick). And some things to be done, don't worry if they don't make sense to you:

- [x] group_start()
- [x] group_finish()
- [ ] ~~negative stack graphs~~ Too fringe a functionality. It was brought to my attention that stack graphs were already too much to include in one component, let's not make it worse.
- [x] ~~customizable negative shapes~~ Instead of building it in, I've added examples of how the user can do that.
- [x] maybe deprecate set_width
- [x] maybe deprecate weird width/height >= 5 checks
- [ ] ~~baseline shape~~ Could get in the way of future improvements. Better give user a callback to draw whatever grid they wish and to map the coordinate system however they want, but I'm not ready for that now.
- [x] ~~nan shape~~ Can be implemented by user with `step_hook()` if needed. `nan_color` should suffice.
- [x] add_value() default to NaN
- [x] docs

### DIFF
--- a/lib/wibox/widget/graph.lua
+++ b/lib/wibox/widget/graph.lua
@@ -20,9 +20,12 @@
 local setmetatable = setmetatable
 local ipairs = ipairs
 local math = math
+local math_max = math.max
+local math_min = math.min
 local table = table
-local type = type
 local color = require("gears.color")
+local gdebug = require("gears.debug")
+local gtable = require("gears.table")
 local base = require("wibox.widget.base")
 local beautiful = require("beautiful")
 
@@ -39,8 +42,6 @@ local graph = { mt = {} }
 
 --- Set the graph border color.
 --
--- If the value is nil, no border will be drawn.
---
 --@DOC_wibox_widget_graph_border_color_EXAMPLE@
 --
 -- @property border_color
@@ -50,6 +51,8 @@ local graph = { mt = {} }
 -- @see gears.color
 
 --- Set the graph foreground color.
+--
+-- This color is used, when `group_colors` isn't set.
 --
 --@DOC_wibox_widget_graph_color_EXAMPLE@
 --
@@ -69,38 +72,90 @@ local graph = { mt = {} }
 -- @propemits true false
 -- @see gears.color
 
---- Set the maximum value the graph should handle.
+--- Set the colors for data groups.
+--
+-- Colors in this table are used to paint respective data groups.
+-- When this property is unset (default), the `color` property is used
+-- instead for all data groups.
+-- When this property is set, but there's no color for a data group in it
+-- (i.e. `group_colors`[group] is nil or false), then the respective
+-- data group is disabled, i.e. not drawn.
+--
+-- @DOC_wibox_widget_graph_stacked_group_disable_EXAMPLE@
+--
+-- @property group_colors
+-- @tparam table colors A table with colors for data groups.
+-- @see gears.color
+
+--- The maximum value the graph should handle.
+--
+-- This value corresponds to the top of the graph.
+-- If `scale` is also set, the graph never scales up below this value, but it
+-- automatically scales down to make all data fit.
+-- If `scale` and `max_value` are unset, `max_value` defaults to 1.
 --
 -- @DOC_wibox_widget_graph_max_value_EXAMPLE@
---
--- If "scale" is also set, the graph never scales up below this value, but it
--- automatically scales down to make all data fit.
 --
 -- @property max_value
 -- @tparam number max_value
 -- @propemits true false
 
---- The minimum value.
+--- The minimum value the graph should handle.
+--
+-- This value corresponds to the bottom of the graph.
+-- If `scale` is also set, the graph never scales up above this value, but it
+-- automatically scales down to make all data fit.
+-- If `scale` and `min_value` are unset, `min_value` defaults to 0.
 --
 -- @DOC_wibox_widget_graph_min_value_EXAMPLE@
 --
--- Note that the min_value is not supported when used along with the stack
--- property.
 -- @property min_value
 -- @tparam number min_value
 -- @propemits true false
 
 --- Set the graph to automatically scale its values. Default is false.
 --
---@DOC_wibox_widget_graph_scale1_EXAMPLE@
+-- If this property is set to true, the graph calculates
+-- effective `min_value` and `max_value` based on the displayed data,
+-- so that all data fits on screen. The properties themselves aren't changed,
+-- but the graph is drawn as though `min_value`(`max_value`) were equal to
+-- the minimum(maximum) value among itself and the currently drawn values.
+-- If `min_value`(`max_value`) is unset, then only the drawn values
+-- are considered in this calculation.
+--
+-- @DOC_wibox_widget_graph_scale1_EXAMPLE@
 --
 -- @property scale
 -- @tparam boolean scale
 -- @propemits true false
 
---- Set the width or the individual steps.
+--- Clamp graph bars to keep them inside the widget for out-of-range values.
 --
--- Note that it isn't supported when used along with stacked graphs.
+-- Drawing values outside the [`min_value`, `max_value`] range leads to
+-- bar shapes that exceed physical widget dimensions.
+-- Most of the time this doesn't matter, because bar shapes are rectangles
+-- and bar heights aren't large enough to trigger errors in the drawing system.
+-- However for some shapes and values it does make a difference and leads
+-- to visibly different and/or invalid result.
+--
+-- When this property is set to true (the default), the graph clamps
+-- bars' heights to keep them within the graph.
+--
+-- @DOC_wibox_widget_graph_clamp_bars_EXAMPLE@
+--
+-- @property clamp_bars
+-- @tparam boolean clamp_bars
+-- @propemits true false
+
+--- The value corresponding to the starting point of graph bars. Default is 0.
+--
+-- @DOC_wibox_widget_graph_baseline_value_EXAMPLE@
+--
+-- @property baseline_value
+-- @tparam number baseline_value
+-- @propemits true false
+
+--- Set the width or the individual steps.
 --
 --@DOC_wibox_widget_graph_step_EXAMPLE@
 --
@@ -109,8 +164,6 @@ local graph = { mt = {} }
 -- @propemits true false
 
 --- Set the spacing between the steps.
---
--- Note that it isn't supported when used along with stacked graphs.
 --
 --@DOC_wibox_widget_graph_step_spacing_EXAMPLE@
 --
@@ -129,27 +182,58 @@ local graph = { mt = {} }
 
 --- Set the graph to draw stacks. Default is false.
 --
---@DOC_wibox_widget_graph_stacked_EXAMPLE@
+-- When set to true, bars of each successive data group are drawn on top of
+-- bars of previous groups, instead of the baseline.
+-- This necessitates all data values to be non-negative.
+-- Negative values, if present, will trigger @{nan_color|NaN indication}.
+--
+-- @DOC_wibox_widget_graph_normal_vs_stacked_EXAMPLE@
+--
 -- @property stack
 -- @tparam boolean stack
 -- @propemits true false
 
---- Set the graph stacking colors. Order matters.
+--- Display NaN indication. Default is true.
 --
--- @property stack_colors
--- @tparam table stack_colors A table with stacking colors.
-
---- The graph background color.
+-- When the data contains [NaN](https://en.wikipedia.org/wiki/NaN) values,
+-- and `nan_indication` is set, the corresponding area,
+-- where the value bar should have been drawn, is filled
+-- with the `nan_color` from top to bottom.
+-- The painting is done after all other data is rendered,
+-- to make sure that it won't be overpainted and go unnoticed.
 --
--- @beautiful beautiful.graph_bg
--- @param color
+-- @DOC_wibox_widget_graph_nan_color_EXAMPLE@
+--
+-- @property nan_indication
+-- @tparam boolean nan_indication
+-- @propemits true false
 
---- The graph foreground color.
+--- The color of NaN indication.
+--
+-- The color used when `nan_indication` is set.
+-- Defaults to a yellow-black diagonal stripes pattern.
+--
+-- @DOC_wibox_widget_graph_stacked_nan_EXAMPLE@
+--
+-- @property nan_color
+-- @tparam gears.color nan_color The color of NaN indication.
+-- @propemits true false
+-- @see gears.color
+
+--- The graph foreground color
+-- Used, when the `color` property isn't set.
 --
 -- @beautiful beautiful.graph_fg
 -- @param color
 
+--- The graph background color.
+-- Used, when the `background_color` property isn't set.
+--
+-- @beautiful beautiful.graph_bg
+-- @param color
+
 --- The graph border color.
+-- Used, when the `border_color` property isn't set.
 --
 -- @beautiful beautiful.graph_border_color
 -- @param color
@@ -157,70 +241,255 @@ local graph = { mt = {} }
 local properties = { "width", "height", "border_color", "stack",
                      "stack_colors", "color", "background_color",
                      "max_value", "scale", "min_value", "step_shape",
-                     "step_spacing", "step_width", "border_width" }
+                     "step_spacing", "step_width", "border_width",
+                     "clamp_bars", "baseline_value",
+                     "capacity", "nan_color", "nan_indication",
+                     "group_colors",
+}
 
-function graph.draw(_graph, _, cr, width, height)
-    local max_value = _graph._private.max_value
-    local min_value = _graph._private.min_value or (
-        _graph._private.scale and math.huge or 0)
-    local values = _graph._private.values
+-- This is what the properties are set to on widget construction.
+local prop_defaults = {
+    baseline_value   = 0,
+    clamp_bars       = true,
+    nan_indication   = true,
+    step_width       = 1,
+    step_spacing     = 0,
 
-    local step_shape = _graph._private.step_shape
-    local step_spacing = _graph._private.step_spacing or 0
-    local step_width = _graph._private.step_width or 1
-    local bw = _graph._private.border_width or 1
+-- These aren't very useful to set, and the docs don't distinguish between
+-- "defaults to" (equals to in a fresh instance) and
+-- "falls back to" (is assumed to be equal to, when nil) anyway.
+--  scale          = false,
+--  stack          = false,
+}
 
-    cr:set_line_width(bw)
+-- This is what the properties are assumed to be in the code, when unset/falsy.
+local prop_fallbacks = {
+-- This one might become beautiful-themed in the future, so we can't set it
+-- in the constructor.
+    border_width     = 0,
 
-    -- Draw the background first
-    cr:set_source(color(_graph._private.background_color or beautiful.graph_bg or "#000000aa"))
-    cr:paint()
+-- These are better left unreplaced in code, because they're used only in one
+-- place and the intent is more clear when the numbers are directly visible.
+--  min_value        = 0,
+--  max_value        = 1,
 
-    -- Account for the border width
-    cr:save()
+-- This one is set later. It's not in `prop_defaults`, because I don't
+-- want to make it accessible through the getter, lest the user somehow mutates
+-- the Cairo pattern and breaks NaN indication for all other graphs.
+--  nan_color        = make_fallback_nan_color()
+}
 
-    if _graph._private.border_color then
-        cr:translate(bw, bw)
-        width, height = width - 2*bw, height - 2*bw
+-- All property defaults are also necessarily fallbacks.
+gtable.crush(prop_fallbacks, prop_defaults, true)
+
+-- These fallbacks are themed and can change on the fly.
+setmetatable(prop_fallbacks, {
+    __index = function(_, key)
+        -- TODO: maybe theme graph.nan_color too?
+        if key == "background_color" then
+            return beautiful.graph_bg or "#000000aa"
+        elseif key == "border_color" then
+            return beautiful.graph_border_color or "#ffffff"
+        elseif key == "color" then
+            return beautiful.graph_fg or "#ff0000"
+        end
+    end
+})
+
+-- This function sets up default implementations for property getters/setters,
+-- when none exist yet. It will be called at the end of this class module.
+local function build_properties(prototype, prop_names)
+    for _, prop in ipairs(prop_names) do
+        if not prototype["set_" .. prop] then
+            prototype["set_" .. prop] = function(self, value)
+                if self._private[prop] ~= value then
+                    self._private[prop] = value
+                    self:emit_signal("widget::redraw_needed")
+                    self:emit_signal("property::"..prop, value)
+                end
+                return self
+            end
+        end
+        if not prototype["get_" .. prop] then
+            prototype["get_" .. prop] = function(self)
+                return self._private[prop]
+            end
+        end
+    end
+end
+
+-- Creates a yellow-black danger stripe @{gears.color} for NaN indication.
+local function build_fallback_nan_color()
+    local clr = color.create_pattern_uncached({
+        ["type"] = "linear",
+        from = {0, 0}, to = {4, 4},
+        stops={
+            {0, "#000000"},
+            {0.25, "#000000"}, {0.25, "#ffff00"},
+            {0.50, "#ffff00"}, {0.50, "#000000"},
+            {0.75, "#000000"}, {0.75, "#ffff00"},
+            {1, "#ffff00"},
+        },
+    })
+    clr:set_extend("REPEAT")
+    return clr
+end
+
+-- Set up the nan_color fallback.
+prop_fallbacks.nan_color = build_fallback_nan_color()
+
+--
+-- Module and prototype methods.
+--
+
+local function graph_gather_drawn_values_num_stats(self, new_value)
+    if not (new_value >= 0) then
+        return -- is negative or NaN
     end
 
-    -- Draw a stacked graph
-    if _graph._private.stack then
-
-        if _graph._private.scale then
-            for _, v in ipairs(values) do
-                for _, sv in ipairs(v) do
-                    if sv > max_value then
-                        max_value = sv
-                    end
-                    if min_value > sv then
-                        min_value = sv
-                    end
-                end
-            end
-        end
-
-        for i = 0, width do
-            local rel_i = 0
-            local rel_x = i + 0.5
-
-            if _graph._private.stack_colors then
-                for idx, col in ipairs(_graph._private.stack_colors) do
-                    local stack_values = values[idx]
-                    if stack_values and i < #stack_values then
-                        local value = stack_values[#stack_values - i] + rel_i
-                        cr:move_to(rel_x, height * (1 - (rel_i / max_value)))
-                        cr:line_to(rel_x, height * (1 - (value / max_value)))
-                        cr:set_source(color(col or beautiful.graph_fg or "#ff0000"))
-                        cr:stroke()
-                        rel_i = value
-                    end
-                end
-            end
-        end
+    local last_value = self._private.last_drawn_values_num or 0
+    -- Grow instantly and shrink slow
+    if new_value < last_value then
+        self._private.last_drawn_values_num = last_value - 1
     else
-        if _graph._private.scale then
-            for _, v in ipairs(values) do
+        self._private.last_drawn_values_num = new_value
+    end
+end
+
+--- Determine the color to paint a data group with.
+--
+-- The graph uses this method to choose a color for a given data group.
+-- The default implementation uses a color from the `group_colors` table,
+-- if present, otherwise it falls back to `color`, then
+-- `beautiful.graph_fg` and finally to color red (#ff0000).
+--
+-- @method pick_data_group_color
+-- @tparam number group_idx The index of the data group.
+-- @treturn gears.color The color to paint the data group's values with.
+function graph:pick_data_group_color(group_idx)
+    -- Use an individual group color, if there's one
+    local data_group_colors = self._private.group_colors
+    local clr = data_group_colors and data_group_colors[group_idx]
+    -- Or fall back to some other colors
+    return clr or self._private.color or prop_fallbacks.color
+end
+
+--- Determine if a data group should be rendered.
+--
+-- The graph uses this method to decide whether the given data group
+-- should get its values rendered.
+--
+-- The default implementation says yes to all data groups, unless
+-- `group_colors` property is set, in which case only those groups are
+-- rendered, for which there are colors in the `group_colors` table,
+-- so one can e.g. disable groups by setting their colors to false.
+--
+-- @method should_draw_data_group
+-- @tparam number group_idx The index of the data group.
+-- @treturn boolean true if the group should be rendered, false otherwise.
+-- @local I'm not confident, that this is good API, so I'm making it private.
+local function graph_should_draw_data_group(self, group_idx)
+    -- This default implementation decides, whether a group should be drawn,
+    -- based on presence of colors, for reasons of backward compatibility.
+    local data_group_colors = self._private.group_colors
+    if not data_group_colors then
+        -- The colors table isn't set, all data groups are deemed enabled.
+        return true
+    end
+
+    -- A group is enabled if it has a color, i.e. nil color means "don't draw it"
+    return not not data_group_colors[group_idx]
+end
+
+local function graph_preprocess_values(self, values, drawn_values_num)
+    -- TODO: elevate to function documentation, if we decide to make it public API.
+    --- Preprocesses values before drawing them.
+    -- This function can return up to 2 values: drawn_values and scaling_values
+    -- The former will be used as values to draw in place of the original data.
+    -- The latter will be used as values to scan for min/max value for scaling.
+    -- Either can be nil, which means: use values as is.
+
+    -- This default implementation is only used to implement
+    -- presumming values for stacked graphs.
+    if not self._private.stack then
+        return
+    end
+
+    -- Prepare to draw a stacked graph
+
+    -- summed_values[i] = sum [1,#values] of values[c][i]
+    local summed_values = {}
+    -- drawn_values[c][i] = sum [1,c] of values[c][i]
+    local drawn_values = {}
+
+    local nan = 0/0
+
+    -- Add stacked values up to get values we need to render
+    for group_idx, group_values in ipairs(values) do
+        local drawn_row = {}
+        drawn_values[group_idx] = drawn_row
+
+        if graph_should_draw_data_group(self, group_idx) then
+            for idx, value in ipairs(group_values) do
+                if idx > drawn_values_num then
+                    break
+                end
+
+                -- drawn_values will have NaN values in it due to negatives/NaNs in input.
+                -- we can't simply treat them like zeros during rendering,
+                -- in case step_shape() draws visible shapes for actual zero values too.
+                local acc = summed_values[idx] or 0
+                if value >= 0 then
+                    acc = acc + value
+                    drawn_row[idx] = acc
+                else
+                    drawn_row[idx] = nan
+                end
+                summed_values[idx] = acc
+            end
+        end
+    end
+
+    -- In a stacked graph it's sufficient to examine only the last summed row
+    -- to determine the max_value, since all values are necessarily >= 0
+    -- and the min_value should be always at most 0
+    local scaling_values = { {0}, summed_values }
+
+    return drawn_values, scaling_values
+end
+
+local function graph_map_value_to_widget_coordinates(self, value, min_value, max_value, height)
+    -- Scale the value so that [min_value..max_value] maps to [0..1]
+    value = (value - min_value) / (max_value - min_value)
+
+    -- Check whether value is NaN
+    if value == value then
+        if self._private.clamp_bars then
+            -- Don't allow the bar to exceed widget's dimensions
+            value = math_min(1, math_max(0, value))
+        end
+
+        -- Drawing bars up from the lower edge of the widget
+        return height * (1 - value)
+    end
+    return value --NaN
+end
+
+local function graph_choose_coordinate_system(self, scaling_values, drawn_values_num, height)
+    local scale = self._private.scale
+    local max_value = self._private.max_value or (scale and -math.huge or 1)
+    local min_value = self._private.min_value or (scale and math.huge or 0)
+
+    if scale then
+        for _, group_values in ipairs(scaling_values) do
+            for idx, v in ipairs(group_values) do
+                -- Do not let off-screen values affect autoscaling
+                if idx > drawn_values_num then
+                    break
+                end
+
+                -- We don't use math.min/max here to be sure that
+                -- min/max_value don't accidentally get assigned a NaN
                 if v > max_value then
                     max_value = v
                 end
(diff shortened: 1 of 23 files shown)


Elv13:
Hello,

Thanks a lot for this, it looks amazing.

> The fact that few seemed to take note of the problems (e.g. #3289) makes me suspect, that nobody really cares or uses the kind of functionality that I'm fixing here. So I want to know if there's even an interest in merging this and get some feedback on it, before I do more stuff.

Yes, there is interest in getting those fixed. You are correct that the number of "graph" users has been in decline over the last decade. It used to be a super popular widget, but less and less configs use it (looking at https://reddit.com/r/unixporn). Those "style" of desktops are trendy. It might very well become popular again, so it is totally worth fixing. 

Or maybe computer got so fast people don't care about resource usage anymore? I certainly did back when I was compiling C++ all day long, but I care less right now when my CI system do it for me. This isn't limited to Awesome. Conky used to be used by literally everybody in the WM customization world until ~2012 then faded. A lot of the trendy "bars" for other WMs lack graphs widgets. IDK, I like them and my config [has some](https://raw.githubusercontent.com/Elv13/radical/master/screenshot/all.png). 

> which prompted me to delve into the code and discover a huge number of other bugs and missing features

Right, totally, I made the documentation images a few weeks ago to kind of give a "baseline" of that work and what doesn't work so we could detect regressions more easily. (and I also traded some old bugs to some new, lesser, bugs, like the border width one).

The problem with the graph widget is the number of drive-by contribution and API ports it had over the years. Awesome <= 3.4 used a different widget "framework" that was barely capable of drawing rectangles. A lot of the graph design is based on "lets draw 1px large rectangle next to each other to emulate the look of a graph". So when we got Cairo widgets, those design decisions were carried over rather than a rewrite. Then various little features like the border got added. They were then made redundant by the `wibox.container.margin` widget, but because we don't break the API, are still supported (and yes, I tried to "fix" it a few weeks ago and made broke it "differently" than it was broken before). At some point (Awesome 3.5), there was a 3rd party "blingbling" module with a saner graph widget, but it has it's share of issues and hasn't been ported to 4.x.

> Stacked graphs hadn't supported everything the nonstacked do, like step_width, step_spacing and custom step_shape-s. Now they do. Before and after the fix:

Thanks for this. This was a pretty bad situation. I wonder whats your opinion. You definitely spent more time into that module code than I did. Should we split the stacked and non stacked widgets/codebase with some minimal magic to avoid breaking the API? An issue with the current code is the large number of `if`/`else` for the stacked where both side of the conditionals support different features. I wonder how much further we can scale this until it becomes unmaintainable? Your patches fit in pretty well, so I guess it isn't an issue yet?

> maybe deprecate set_width maybe deprecate weird width/height >= 5 checks

Yes, we can deprecate them, but we need to avoid breaking the API, the behavior must be preserved for 4.x, then have a deprecation warning on stderr for 5.x, then error in 6.x (if that ever happens).

> docs

Looking at the screenshots in the top post, looks like you wrote some new examples/tests, please go ahead and add them. The more examples we have the better.

Otherwise, for the whole PR, I think we could go ahead and merge a significant (if not all) fraction of it without having to wait for all the checkbox to be filled. The only thing I want before merging this is all the doc and unit tested examples to exercise the code. We have ways to detect when the image output changes due to regressions. Plus, some of the code, like the stacked stuff, had no tests and very few users until now due to the limitations you discovered. It will be nice to document and exercise the code properly so it gets more users.

Adding proper unit tests (`/spec/wibox/widget/graph.lua`) is _possible_ (by looking at the private data structures) and would be appreciated, but I don't think its fair to ask them for this. The old code had none and that's on us. So think about it and see how comfortable you are with adding some, but if it's too much work, then I will happily still merge the PR anyway. It would help with regressions as you add more of the features you mentioned, so it _might_ help you with development? Tough call. I personally often rely on some `assert` in the `tests/examples/*` code as a middleground.

albel727:
>I made the documentation images a few weeks ago to kind of give a "baseline" of that work and what doesn't work so we could detect regressions more easily.

Hah, what a lucky happenstance, that I've only lately wondered, why my temperature graph widget with a `min_value > 0` can't seem to max up. I have used the images exactly for regression testing, and if they hadn't been there, I'd have only delivered a fix for scaling with `min_value` and given up on a more fundamental widget rework.

>there was a 3rd party "blingbling" module with a saner graph widget

Yeah, I had looked at its line_graph to see if there's something to learn or salvage, and as far as I can tell, it's only a bad copy of the awesome's graph with dropped support for data groups. It only adds a text and locks user in rounded_rectangle for widget shape. Its code is unreadable and suboptimal at times. Maybe I'm missing something, and maybe it was different before, but today I'd prefer awesome's graph on every day of the week.

>An issue with the current code is the large number of if/else for the stacked where both side of the conditionals support different features.

I wonder what you'll say after seeing the commit `Refactor graph:draw() to separate concerns`. In my mind, the majority of the code doesn't really have anything to do with stacked graphs. There's only
1. `stack_colors` property, which presently should have been more correctly named `data_group_colors`, because these colors are used in exactly the same way for all graphs, stacked or not.
2. Data preprocessing for purposes of rendering, which we, yes, by default happen to utilize for stacked graphs, but which is a more general concept, see `graph:preprocess_values()`.
3. A few lines in the drawing loop, controlled by the `stack` property, and their effect is merely using the top of the previous bar instead of the baseline during rendering. I.e stacking.

https://github.com/awesomeWM/awesome/blob/d0ed6d63112c5d9c0b83c886ddfecad9d966fb63/lib/wibox/widget/graph.lua#L391-L396

Only the latter is what really behaves differently, when user toggles the `stack` property. One could argue, the commit does as much of the codebase split as is really possible. I'm not sure what else is there that could be easily split out, without having to duplicate almost all of the rendering code.

>Yes, we can deprecate them, but we need to avoid breaking the API

Ugh, that would be even more cruft than it already is. Am I doing it right?
```patch
diff --git a/lib/wibox/widget/graph.lua b/lib/wibox/widget/graph.lua
index df29e9e5..d89b8d6e 100644
--- a/lib/wibox/widget/graph.lua
+++ b/lib/wibox/widget/graph.lua
@@ -416,8 +416,8 @@ function graph:draw_values(cr, width, height, drawn_values_num)
     end
 end

-function graph.fit(_graph)
-    return _graph._private.width, _graph._private.height
+function graph.fit(_, width, height)
+    return width, height
 end

 function graph:get_drawn_values_num(widget_width)
@@ -519,12 +519,16 @@ end
 -- @propemits true false

 function graph:set_height(height)
-    if height >= 5 then
-        self._private.height = height
-        self:emit_signal("widget::layout_changed")
-        self:emit_signal("property::height", height)
+    gdebug.deprecate("Use a `wibox.container.constraint` widget or `forced_height`", {deprecated_in=5})
+    if awesome.api_level <= 5 then
+        if height >= 5 then
+            -- this sends "layout_changed" for us
+            self:set_forced_height(height)
+            -- signal, because we did it before
+            self:emit_signal("property::height", height)
+        end
+        return self
     end
-    return self
 end

 --- Set the graph width.
@@ -534,12 +538,16 @@ end
 -- @propemits true false

 function graph:set_width(width)
-    if width >= 5 then
-        self._private.width = width
-        self:emit_signal("widget::layout_changed")
-        self:emit_signal("property::width", width)
+    gdebug.deprecate("Use a `wibox.container.constraint` widget or `forced_width`", {deprecated_in=5})
+    if awesome.api_level <= 5 then
+        if width >= 5 then
+            -- this sends "layout_changed" for us
+            self:set_forced_width(width)
+            -- signal, because we did it before
+            self:emit_signal("property::width", width)
+        end
+        return self
     end
-    return self
 end

 -- Build properties function
@@ -570,15 +578,32 @@ end
 function graph.new(args)
     args = args or {}

+    if args.width then
+        gdebug.deprecate(
+            "`args.width` is deprecated. Use a `wibox.container.constraint` widget or `forced_width`",
+            {deprecated_in=5, raw=true},
+        )
+    end
+    if args.height then
+        gdebug.deprecate(
+            "`args.height` is deprecated. Use a `wibox.container.constraint` widget or `forced_height`",
+            {deprecated_in=5, raw=true},
+        )
+    end
+
     local width = args.width or 100
     local height = args.height or 20

-    if width < 5 or height < 5 then return end
+    if awesome.api_level <= 5 then
+        if width < 5 or height < 5 then return end
+    end

     local _graph = base.make_widget(nil, nil, {enable_properties = true})

-    _graph._private.width     = width
-    _graph._private.height    = height
+    if awesome.api_level <= 5 then
+        _graph._private.width = width
+        _graph._private.height = height
+    end
     _graph._private.values    = {}
     _graph._private.clamp_values = true
```

>looks like you wrote some new examples/tests, please go ahead and add them.

Yeah, they're pretty sloppy as of now, but it was my intention anyway, if even to improve code coverage.

>I think we could go ahead and merge a significant (if not all) fraction of it without having to wait for all the checkbox to be filled. The only thing I want before merging this is all the doc and unit tested examples to exercise the code.

Heh, most of the checkbox stuff is actually trivial and mostly done, I have more dread of documenting and unit-testing stuff.
Bear with me, I do still work on it, I just need some time to think it over to not regret any decisions later and not have to sprinkle stuff with 10 pages of `gdebug.deprecate()`. I do change and git-rebase some stuff even now, I hope you don't mind.
If you really want, then I guess everything up to and including `Avoid incremental transform shifting during graph bar rendering` could be merged right now. It is unlikely to be changed, fixes the worst bugs, doesn't fundamentally change anything, and doesn't require any examples or docs, because nothing new is introduced yet.

>We have ways to detect when the image output changes due to regressions.

Oh, do tell. I had to invent my own stuff based on converting the example svgs with `inkscape -d 72`, to get a pixel-correct reproduction of what the user sees.

>proper unit tests

I'll... see what I can do. >_> There's a couple of lines in graph.lua that I might not be able to cover from tests/examples,
because I need draw() called more than once, so maybe I'll write a small spec for them at least.

Elv13:
> Oh, do tell. 

Sorry to disappoint: `git diff` ;)

albel727:
Ah, so you're diffing `build/doc/images/` SVGs as text? Yeah, that works too. But I couldn't just do that, because I had to switch between using lines and rectangles, which is bound to change what is written in SVG, but doesn't change a single pixel of the on-screen output, and I was caring more for the latter, so I went directly for comparing the rasters.

Elv13:
The main use case for the diff is weirder and unexpected: reproducible builds. While making change to the code or switching to a different version of Cairo will cause the output to change, it isn't the only way.

Things like using `pairs` or `next` in the wrong place by accident can cause subtle changes from build to build. For example, the `piechart` pies were not always rendered in the same order, some padding/margins generated in for loops are unstable, etc. Same goes with stuff that somehow end up using pointer values as IDs. Debian cares about that kind of issues a lot. And, after all, they _are_ bugs.

My idea for a long time was to use the fact that SVG are XMLs to traverse the tree and point narrow where the diff are on screen, but I never implemented the code, so it's still a dream :). Something like OpenQA could be used to do pixel level diffs I guess? As of right now we don't have _that_. So, if you want snapshots, committing the svg and diffing them is a workable solution. The Inkscape or headless Chrome/Firefox based ones are better, but I never did it myself

lib/wibox/widget/graph.lua
```diff
@@ -416,8 +417,8 @@ function graph:draw_values(cr, width, height, drawn_values_num)
     end
 end
 
-function graph.fit(_graph)
-    return _graph._private.width, _graph._private.height
+function graph:fit(_, width, height)
+    return width, height
```
Elv13, line 421: This method should still check if `self._private.width` or `self._private.height` are set and honor them. Yes, I know, compatibility is annoying ;)

albel727:
Yeah, I'd considered for a second, if I should use `pairs` on data groups table to allow the user to use human-readable names for them, but decided against it for the very same reason of order instability.

And speaking of reproducible builds. Does that mean that I'm not to use math functions in tests at all and must provide my own hard-coded arrays? Or would simply quantizing computed values, so that no reasonable hardware would produce a different result, suffice?

Elv13:
>  Does that mean that I'm not to use math functions in tests at all and must provide my own hard-coded arrays? 

It is the reason why I copy pasted the array in each tests/examples. It was generated randomly, but then hardcoded.

Using a predictable seed is an alternative, but there is some magic to initialize the RNG in the codebase, so I didn't bother attempting to override it. True random functions and non deterministic hash function should be avoided.

Elv13:
> Well I had something like math.floor(math.sin(t) * 16 + 0.5) in mind. 

I am ok with sharing the array in a separate file. `sin` does make for pretty images, but I think it would mislead the users that the graph is using bezier curves rather than individually rendered discrete datapoints. I have no super strong opinion here, so feel free to ignore me on that one.

sclu1034:
You could probably do some trickery like this in the test file:

```lua
--DOC_HIDE First few values of generated data array to show off the concept.
--DOC_HIDE Actual data is imported below.
local data = {
	1,
	2,
	3,
	-- snip
}

data = require("path/to/actual/data") --DOC_HIDE

-- widget definition here
```

This way we get pretty pictures while still showing how it's just a list of values.

psychon:
I haven't (and most likely won't) look at the individual commits, but still have to write: Wow. Just writing the PR message must have taken ages for all those images. Well done!

lib/wibox/widget/graph.lua
```diff
@@ -159,6 +159,20 @@ local properties = { "width", "height", "border_color", "stack",
...
 
+local function graph_gather_drawn_values_num_stats(self, new_value)
+    if not (new_value >= 0) then
+        return -- is negative or NaN
+    end
+
+    local last_value = self._private.last_drawn_values_num or 0
+    -- Grow instantly and shrink slow
+    if new_value < last_value then
+        self._private.last_drawn_values_num = last_value - 1
```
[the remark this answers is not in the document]
albel727: I'm not *setting* the initial value of `last_drawn_values_num` above 8000, I'm allowing up to 8k elements to accumulate until the first call to `draw()`. The first call to `draw()` will set `last_drawn_values_num` for the first time, after which the number of kept elements will immediately plunge to that + approximately 64-128 elements for safety. Shrinking by halfway steps is too fast and will defeat the purpose, it's `log_2 n` vs `n` difference, e.g. 7 steps vs 128. So the component would fail to tolerate some short temporary size shrinkages.

I would actually almost think that all of this is overengineered, and one could just keep 8k elements at all times, because memory and performance are unlikely to be the problem for 99.9999% of users, while annoyingly missing values are. But well, the current algorithm is also not that complex either and allows for the eventual optimal performance. Either way, shrinking faster is not a priority and the memory would not be reclaimed by Lua after first use anyway.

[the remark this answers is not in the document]
sclu1034: Ah, I see. I missed the connection then.
Thank you for clarifying.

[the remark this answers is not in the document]
psychon: Uhm... what if the widget is shown in multiple places at once with multiple sizes...?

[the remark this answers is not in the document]
albel727: The quoted lines are supposed to handle exactly the case, by keeping the largest `last_drawn_number_num` and only slowly decreasing it, when widget is drawn with smaller sizes. There are additional 64-128 safety elements kept, so the widget could be drawn in 64 places at once with smaller sizes and the user won't notice the lack of values in the largest instance. If they somehow want more widgets, there's the new `capacity` property, which allows to override the heuristic and keep any number of elements the user wants.

Elv13:
Wow, this is really getting there!

Can you give some insight about the bezier features you added to some tests? How is it going to be used in practice? Is it something which will be used by a subclass (or main graph widget) or is it something users will have to copy paste? There's a whole lot of business logic into these example files.

I tested your examples are it's very pretty. It's nice to see the doc well on its way too!

albel727:
> Can you give some insight about the bezier features you added to some tests?

I... er... >_> I just added them after you said that drawing sines "would mislead the users that the graph is using bezier curves". I've solved this particular problem by adding some bar spacing, in order to make it obvious that the graph is discrete (e.g. `graph_negative_sine.lua`), but then I've decided to do this proof-of-concept thing, to show that bezier curves are not really impossible either.

I've intended for it to maybe serve as examples for the `step_hook` property (name bikeshedding welcome). The three files have indeed quite a lot of logic, but they're 95% the same code, so I'd merge the 3 images into one as a first measure. (Especially since there's sadly no way to add more than one image per one doc item, due to the way example substitution is done).

I'd expect users, who would, of all things, want BEZIER CURVES in their tiny widgets, to handle the brunt of copy-paste, frankly. 
One can do a helper module out of it, then the examples would be much lighter, I'm just not sure that it's in the scope of a window manager to have something like `gears.math.bezier_interpolation`. Alternatively I can just drop the examples. Users who would want to have anything to do with `step_hook`, are probably smart enough to implement it themselves.

Now that you've got a look at it, it's your call.

Elv13:
> Alternatively I can just drop the examples. 

No! Those things are amazing. This is the main reason BlingBling was used back in the days. I am really looking forward to integrating those bits. `gears.math` is a good place. Maybe using some of Cairo built-in magic where it makes sense could also work. In any case, don't delete it.

Worst case scenario, we add a new page in `docs/` like `50-graphs-examples.lua` and expose the examples there. There is `--DOC_NO_DASH` you can add in examples to be able to insert them into top level doc pages. They are too large to be added inline in the doc, but are very worthy examples.

albel727:
>Maybe using some of Cairo built-in magic where it makes sense

The only magic Cairo has for drawing the curves is... the method that draws the curves. The real trick is calculating the necessary parameters for it.

>In any case, don't delete it.

Ok, I've added a `gears.math.bezier` module and merged the three bezier examples into one and added it to the documentation. I hope it's not too big for inline docs still.

Elv13:
I did some testing on this. It's **really** high quality work compared to the average PR.

I pushed some minor changes, I hope you don't mind. First, I rebased the whole thing to make sure each individual commits passed the tests. They _nearly_ did. However there was a few `luacheck` warnings in like 4 of them. I also removed some shadowing override while at it. Some of the shadows, like `local baseline_y = baseline_y` can be quite confusing when jumping into the middle of the code and don't read what's in between. So I renamed the "first" one `main_baseline_y`.

I added 4 commits. The first fix the table of content rendering in the doc. Mixing properties and callbacks in the same section rendered the HTML incorrectly (they have different number of columns in a `<table>`).

The next 2 commits are to simplify the `opts` and `step_hook` callback. Some of the `opts` and the `step_width` callback arguments are always the same as `opts.graph.<property>`, so the callback could be simplified a lot. From a user perspective, less callback arguments means less unused variables (but my next commit adds 1...).

The last commit exposes the wibox `context` to the callback and merge it with `opts`. It also moves it to the front along with a `self` argument for consistency with other APIs (like signals and widgets). The reason for this is that HiDPI screens in a multi-screen setup will have corner cases the previous hooks could not handle. The context contains a `dpi` field which can change is the graph is drawn in multiple places or moved.

Elv13 (dismissed):

Elv13 (approved):

Elv13 (dismissed):

albel727:
> They _nearly_ did. However there was a few `luacheck` warnings in like 4 of them.

I was supposed to have fixed all of them some couple of hours before your push. I would have said explicitly when the PR is ready to be reviewed, to spare us some duplicate work. I'm sorry for taking so long with this work, but for now it simply isn't ready yet.

>The next 2 commits are to simplify the opts and step_hook callback. Some of the opts and the step_width callback arguments are always the same as opts.graph.<property>, so the callback could be simplified a lot. 

That's a performance hit and a layering violation. Why is making the `opts` table have less keys even a priority? It's populated only once per a `draw()` call and its contents aren't really hard to remember or too long to document. 

Firstly, you're trading one lookup in the `opts` table and/or a fast local argument that almost every drawing callback is guaranteed to need and use (and that the user can name arbitrarily shortly) for a cumbersome  `self.step_width` which goes through 
1. key lookup in `self`
2. the whole property metatable lookups, then hits 

```lua
if properties_defaults[prop] ~= nil and  _graph._private[prop] == nil then
    return properties_defaults[prop]
end
return _graph._private[prop]
```

I mean, just count the dots and square braces in that. And that's for *every drawn value*. I mean it's not like it would *really* be a large hit or really matter, because nobody is really going to want to draw live audio data with the graph widget, but after having taken some (admittedly pretty irrational) care of the performance, I find it a bit sad.

Secondly, a draw callback shouldn't concern itself with graph widget properties and assume their logic. What if one swapped `graph:draw_values()` (or other customization methods like some I had in mind) out with something that determines `step_width`/`step_spacing` **NOT** via the property values? What if at some future point it could become e.g. a table to set `step_width`s per data group? Etc etc.

No, everything potentially useful and customizable should be delivered into callbacks directly, in a way that encourages the user to use it. Reading graph properties should be a last resort for a callback. "Optimizing" the opts table or callback parameters out is encouraging fragile code, limiting the design space for future improvements, and is clearly sparing at the wrong end.

> It also moves it to the front along with a self argument for consistency with other APIs (like signals and widgets).

I'm not sure why the two much more rarely used arguments `(self, context)` really need to be added in front of the callbacks, so that the typical user has to type `_, _, cr` everywhere they go. The analogy with `widget:draw()` doesn't really justify it in my opinion, like it doesn't justify it with for `step_shape` callbacks, with which the specialized drawing callbacks have more in common. And even if it does, it would've made sense to add the `context` parameter in `draw_values()` at the second instead of the last place.

lib/wibox/widget/graph.lua
```diff
@@ -621,25 +626,27 @@ function graph:draw_values(cr, width, height, drawn_values_num)
     -- pass data pertaining to the same drawing session between callbacks, but
     -- all underscore-prefixed keys are reserved for future use by the widget.
```
albel727, line 629: If you're going to drop the leading underscores convention, this doc line is probably obsolete.

lib/gears/math/bezier.lua
```diff
@@ -19,6 +19,7 @@ local bezier = {}
 -- @tparam number t The value of the t parameter to evaluate the curve at.
 -- @treturn number The value of `B(t)`.
 -- @staticfct gears.math.bezier.curve_evaluate_at
+-- @see wibox.widget.graph.step_hook
```
albel727, line 22: These make no sense IMO. They're pure math functions with no direct connection to graphs or drawing in general. Many of them aren't even used in the `step_hook` example and probably wouldn't even be useful in such a callback, but were simply added for completeness, because that's what any decent generic bezier curve math library has to have. "@see wikipedia" would make more sense at this point.

lib/wibox/widget/graph.lua
```diff
@@ -597,7 +600,7 @@ local function graph_choose_coordinate_system(self, scaling_values, drawn_values
     return min_value, max_value, main_baseline_y
 end
 
-function graph:draw_values(cr, width, height, drawn_values_num)
+function graph:draw_values(cr, width, height, drawn_values_num, context)
```
albel727, line 603:
```suggestion
function graph:draw_values(context, cr, width, height, drawn_values_num)
```

lib/wibox/widget/graph.lua
```diff
@@ -433,7 +436,7 @@ function graph:draw(_, cr, width, height)
         local values_width = width - 2*border_width
         local values_height = height - 2*border_width
 
-        self:draw_values(cr, values_width, values_height, drawn_values_num)
+        self:draw_values(cr, values_width, values_height, drawn_values_num, context)
```
albel727, line 439:
```suggestion
        self:draw_values(context, cr, values_width, values_height, drawn_values_num)
```

lib/wibox/widget/graph.lua
```diff
@@ -975,6 +982,9 @@ for _, prop in ipairs(properties) do
     end
     if not graph["get_" .. prop] then
         graph["get_" .. prop] = function(_graph)
+            if properties_defaults[prop] ~= nil and  _graph._private[prop] == nil then
```
albel727, line 987: Instead of bothering with this automagic, that violates the `set_prop(v); get_prop() == v` property, I'd just set the properties once in `graph:new()` and be done with it.

lib/wibox/widget/graph.lua
```diff
@@ -21,7 +21,6 @@ local setmetatable = setmetatable
 local ipairs = ipairs
 local math = math
 local table = table
-local type = type
```
albel727, line 24: This removal isn't there in my local version of the commit and `type` is used in the file. You must have saved an old version of my commit or something.

albel727:
No response? Okay, I can't hang in limbo here, I'll be reverting some of it then. In fact I think I've done enough, those will be my last commits. If you want to modify my work, I feel that better be happening without me, for my peace of mind.

Elv13:
> No response?

I will provide one tomorrow. I don't have time right now with my job and some other stuff.

lib/wibox/widget/graph.lua
Elv13: Yes, this would work

lib/wibox/widget/graph.lua
Elv13: I first pushed 1 test commit to the branch before making the force push. Given it worked, it means the version I had locally was still the head of the PR at that point. I removed `type` because `luacheck` said it was unused and made the build fail. Maybe we have different versions of luacheck?

albel727:
> It's a lot of code. Waiting for "when it's ready" to get a review is a pretty risky thing to do. I am guilty of making giant PRs myself, but this kind of PR are hard/long to review and making changes at the end rather than earlier tends to be very depressing to both sides.

There's a difference between a review and a force push, that somehow ends up reviving old versions of my commits and pointlessly renaming variables that I had already renamed. I mean, ok, you don't have the time to really pay attention, I can understand that, we all get overworked, but I would have then preferred, if you told me what you want first. I had fixing luacheck problems on my list, as you found out, and I would've done it sooner, had you simply told me to. Same goes for the context thing, etc. 

I mean ok, some things are faster done than explained, but the checkbox for letting maintainers push changes into the pull request is here for letting them push _discussed_ changes. At the very latest, at the moment you started getting merge conflicts you would have done better by simply abandoning the merge attempt and just pointing me to a branch in your repo and asking me what I think about it. What if I hadn't accidentally noticed that you pushed your changes and inadvertently overwritten them with my own force-push? Now that would be awkward.

I've also offered at least twice to split this PR up, offers, that you passed up without a comment, so I would have appreciated if you didn't use wording like "guilty of making giant PRs" while talking to me.

>Yes, it is a priority. Once something is exposed in the public API it is very hard to go back and requires a years and years of deprecation code paths. 

Which is a thing I obviously understand, which is why I care about these parameters.

>From a maintainer point of view, I have to make sure what we expose and document isn't redundant. In this case, there is values that should belong in their accessors, not duplicated with a default value.

This is what is depressing in this for me, you have read my comment, but you don't seem to have understood it, even though I tried to be as clear as I can, and I would've actually thought that things that I say are common sense. You seemingly haven't read the comments in my newer commits either. Let me tell you for the third time, they're **NOT** the same values, they're **NOT** duplicated, they're simply named similarly because they have somewhat of a related function, and the property having a "default nil value" whereas the parameter being always a number is just a first symptom of it.

A property is just one of the things for the graph to factor in, when determining, what to pass into the callbacks. A callback, whose purpose is to draw bars, shouldn't even care from which kind of graph it is called, let alone query its properties. It should draw bars. With the given parameters that are necessary and sufficient to draw most generic bars. Which, respectively, should be passed into it directly.

(I actually have qualms about passing `step_width` as a direct parameter too instead of keeping it only in `opts`, because if a custom `group_start` callback for whatever reason wants to trick a `step_hook` callback into drawing steps with different widths, it probably should be able to do it by simply changing the value in the `opts`).

The `step_width` _parameter_ IS "a part of the state" just like the `group_idx` parameter is.

So speaking about breaking API, imagine someone writing a library with nice drawing `step_hook`s, e.g. making `curvaceous` into a library, so that other uses could just go `step_hook = require("curvaceous")` in their graph and have a nice bezier rendering with zero effort and clutter. If I tomorrow add the possibility for `step_width` to be a table of widths per data-group, or even a function (with `self, group_idx` parameters of course, because THERE it DOES makes sense), a user will not be able to use both that and curvacious, because it will be broken with your version of callbacks API, because it had to assume things about who calls it and what this property means, and my version of callback would JUST WORK™. (My original implementation of curvacious actually didn't even assume that `step_width` is the same from one value to the next, but we had to make it shorter for the docs, so...)

You would say that it's because one broke the `step_width` property API and hence that is to blame for everything and not the callback design, but I don't think you realize that you're attempting to break its API right now by making it never return nil anymore - without any deprecation warning! - and "this property was set by user to a specific value" vs "this property is a default" is a distinction one may want to observe. 

I think you're conflating different purposes. `step_width` property is just what the user has set (or hasn't set) it to and that's how it always was. It's purpose is not to answer "what step width the graph will be using during drawing", because the question can have many answers and, let's face it, nobody will ever care about that, except the drawing callbacks, and they shouldn't go through any lengths to get that answer, like querying whatever properties.

Since the property doesn't have to answer any questions, except "what the user set it to", and only answer it to the code, that originates from the user creating the widget itself, or to the code, that knows intimately, how the graph widget works, i.e. some code in awesome itself, one is free to _extend_ its functionality, like making it accept functions too, and that wouldn't be a breaking change.

The only reason I'm even passing the graph into the callback is so that hypothetical curvacious library's authors can query THEIR OWN properties on the widget, something that THEY fully control. E.g. instead of (or in addition to) the `draw_line` parameter `curvaceous()` could read `graph.curvaceous_line` property to determine if it should draw just the curve instead a full bar, and the user can then just `widget.graph { step_hook = require("curvaceous"), curvaceous_line = true }`, instead of writing out a full function to just pass a `true` into curvaceous. Now THAT is proper usage and design.

>This would make a lot of temporary tables, but opts already do it anyway. 

`opts` is just one table per `draw` call. Two (not three like you made it initially), if you count the metatable for merging it with the context. There's a table with summed values created for every row of a stacked graph, and there are lots of "default implementation" functions that get created too (but they don't close over any external variables (at least they didn't until you changed that with one of your commits, which I reverted), so they can be made to not be allocated anew on every `draw` call). So, yeah, it's a bit late to care about temporary tables, but if you care that much about garbage collection churn, stacked graphs could probably be redesigned it in a way that doesn't require creating tables. I just assumed that it's cheap enough unless one does it, like, for every value, at least I didn't notice awesome's code having much qualms about creating new tables.

I don't see any reason here to nest tables within tables within tables anyway. Current flat options design is fine as it is.

>Since if you do that, only group_idx is left, the opts/context table could be only created once and group_idx updated before calling the callback. 

It could be just as well created once and `step_width`/`step_spacing` updated in addition to `group_idx` before calling the callback, I don't see why you're splitting hairs there.

>Also, the _ in front of variables runs against the naming convention. 

I'm becoming more confident that you haven't read my commit message (I forgive you, they do tend to be verbose, but it's because the commits, I hope, are well thought out). Since we merge `context` into `opts`, there's a danger that, when adding new fields to either, one will forget about the other one, and accidentally introduce a name clash, with one field shadowing the other and no way for the callbacks to recover from it even after people learn about the fact. Another danger I originally wanted to prevent with this is that `group_start` callback from one author inadvertently overwrites a field that is later introduced by graph and is used by a `step_hook` from another author, thus preventing from using both callbacks together in the same graph instance.

Both are admittedly unlikely, so I'm willing to yield on this matter, but let my reasons remain on the record.

>Then the API will be extended then. Making preempive APIs because something might happen eventually is harmful.

No, that's called "foresight" and "proper design". It would've been harmful if we had to go at lengths to do it, or if these changes prevented further redesigns. But it's exactly the opposite of that. Populating one table with parameters needed for "a generic bar in a generic viewport" is not hard and it's actually your way to do parameters, that would prevent further redesigns of the graph widget, like I've explained.

>This is how you end up with (_, _, _, _, i_care_bout_this_one, _, _, _). Adding more parameters to a callback doesn't encourage people do use it. Quite the opposite. 

Which is funny, considering how it's your way that introduces 2 new rarely used parameters instead of the one always used `step_width` and forces people to write `(_, _, cr)`.

>What encourages people to use API is how consistent they are and if they can learn and recall which argument it has without going back and read the API doc.

I wonder why you added the context as the last parameter to `graph:draw_values()` then, if the order is so important. (Apropos `draw_values()`, that reminds me to make it private too, since it's undocumented). 

So, people, who would be smart enough to draw their things with cairo, but somehow not smart enough to remember, that a bar needs a cairo, an x, a top, a bottom and a width, are your concern? Really? How many times and how often do you think they'll be writing a `step_hook`, of all things, so that this would be a burden? Please, when answering this question, remember that we're not talking about the ubiquitous signals. No, I'd say, sparing your average `step_hook` writer one (and only in their whole lifetime) lookup in the docs is trumped by designing the API in a way that doesn't encourage the average `step_hook` writer to write fragile code, fixing which would demand more lookups in the docs from them later.

>You are correct that context is unlikely to be used compared to cr. However context is necessary for the DPI and is called context everywhere else. However context is necessary for the DPI and is called context everywhere else. So given you need to add context here, then having both opts and context is redundant.

And that settles it, you haven't even seen my commits. I'm not even arguing with THAT, I think it's actually a good idea to merge them, and I have done it again after I reverted your changes. I would've done it myself from the start, if I had really understood what `context` really is, because, you know, for something so important it doesn't seem to be really documented anywhere, except as a mysterious "The context"(sic!). Where does it even come from really? Wibox? Hierarchy? I'm not even sure, and I'm not sure if any potential user is.

By all means do document it, so that one could add a link to it in the graph `opts` docs, as opposed to documenting its fields directly in graph docs, which is bound to become out of sync eventually.

I didn't rename `opts` into `context`, because they're different things, and it would be bad, if the user sees a "context" parameter and goes "ah that's """"The context"""", well I don't care about THAT, I just want a step_spacing" and fails to read the documentation which describes its actual contents. 

`opts` can still be renamed into `options`, if we hate acronyms that much.

>Also, having opts are the end kind of prevents adding more parameters later because it would look strange.

Well... because... you know... preventing adding more parameters is, like, the whole POINT of having an "opts" argument. Having it is the only REASON, why a callback like `curvaceous` is even ALLOWED to have an additional parameter like the `draw_line`. Without it a library author couldn't implement a `function curvaceous(blah blah, draw_line)` and not fear that a user would write `graph { step_hook = curvaceous }` which would then break as soon as you add another parameter to `step_hook`.

And you've just talked about not breaking APIs and then you speak of "adding more parameters" to `step_hook`. What? "Adding more parameters" in a way that doesn't break the API can be done by simply adding new fields to the `options` parameter, which is a common wisdom among linux kernel developers, for instance. They have to do it for somewhat different reasons, but if you want for both the caller's writers and callees' writers to be able to extend the parameters, variable positional parameters are simply not enough.

>I see the 3 draw callback as more specific draw methods.

I don't see them so. They're an internal API of the graph widget, are not useful outside graph widgets, and, if anything, are more analogous to shapes, which don't have to have and indeed don't have anything like a "widget" parameter in them at all, for apparent reasons. In fact, they *are* shapes, just with parameters that are more streamlined for the graph widget, because they don't have to be universal across widgets. 

IF the shapes API were designed properly, that is, if a shape were a `function(cr, width, height, x=0, y=0, context)`, I wouldn't even have to add the awkward `step_hook` property. I would've simply reused `step_shape` by putting a few useful things like `baseline_y`, `step_spacing` and `graph` into the context and would have called it a day.

>All signal callbacks (which are the near totality of callbacks in the APIs) do it.

Which is a total non-argument for reasons you yourself explain later. These are `n:n`, and the `step_hook` is `1:1` (actually `n:0,1`, when one considers alternative graph implementations), so it doesn't have to follow their design decisions.

>Also, having self in front is just a plain Lua thing.

It's a common OOP language thing, that isn't attributable to Lua in particular. But I'm not sure "because it's OOP" is a valid justification for putting "self" everywhere without real need.

>but the method parameters and order should match those of draw() and we should stick to that for every future implementation of this concept.

"callback" is indeed a misnomer, but that argument only works in my favor, because a non-callback usually has less reasons to receive the caller as the parameter. Just because `step_hook` is a function that belongs to an object doesn't mean that it's a method and it must have a "self" parameter. Does a sorted collection that accepts a "compare" function require it to have a self parameter? No. Classes are not "Lua thing" anyway, i.e. they're not first-class citizens (pun unintended).

albel727:
Let me pipe in again with one last defense.

Suppose I even give in and agree to let `step_width` mean "with which step width the graph will draw stuff", and it will no more return nil. Suppose we will never want to extend it with stuff like "table of widths" functionality, so it will ALWAYS be a number.

How then one goes about say, implementing another property like `bar_number` which, when set, makes the graph always show a given number of bars, no matter how it's resized? Well, it's not possible. There isn't a single meaningful number you could even return from `step_width`, because a widget could be drawn with different widths at the same time.

"Well okay," you'll say, "THIS graph is for bars with fixed bars. If you want autoresizing bars, just implement another graph widget". Okay then, one obliges and implements a flexgraph, and everything is fine for a moment. Until one user wants to use curvacious in a flexgraph, and discovers, that it's not possible, because the author of curvacious was tempted by the close-proximity `self` parameter and simply used `self.step_width` instead of `options.step_width` or even forced to do it, because there were no `options.step_width`, "because it's redundant". Now the user must beg either the author of curvacious to fix it, which would be `local step_width = options.step_width or self.step_width`, or the author of flexgraph to fix it, which would be fugly hacks like emulating a `self.step_width` property which would change its value inside `draw()`. Either author can no more be maintaining their stuff or not willing to go at the lengths, and the problem is scaling as `number of bar libraries * number of bar implementations`.

Let us squint really hard at the `group_start`/`group_finish`/`step_shape` callbacks one last time. When I do it, I see
`start(cr, options)`
`step(cr, x, y, width, height, options)`
`paint(cr, options)`
which to me doesn't look like `widget:draw()` methods at all but rather awfully like a generic API for painting things in bulk. I rest my case.

Let it be known, that I've done everything I still wanted to do. I've renamed `opts` into `options` and `cur_y` into `value_y`, in order to conform to the "no acronyms" policy. You'll also see a pair of spec files there.

I'm not going to do anything else, and I'm quickly losing the ability to care, what you're going to do with this next. You can take it or you can do another PR based on this one, but this PR stays the way it is.

lib/wibox/widget/graph.lua
albel727: Sorry, I have missed these comments of yours, because my dumb overlarge force-pushes hid these comments in "94 hidden items, Load more..." and I somehow didn't get a notification.

>I removed `type`

Well this is weird, because the removal somehow ended up in one of my commits. I saw the change in the diff between our branches, but couldn't find the change in any of your commits, which puzzled me immensely, so I had to git-blame it and found it in my own very second commit, in which I vaguely remembered doing the removal, before deciding against it, but it was a long time ago, so I suspected that you saved my old commit. 

I don't know, if the problem is in luacheck (I have 0.23.0), but I do use `type` in
```lua
    if not group_start or type(group_start) == "number" then
```
so it's weird that luacheck would have complained about it, given that we even have tests that exercise this line.

lib/wibox/widget/graph.lua
albel727: Okay, I will do that then.

lib/wibox/widget/graph.lua
albel727: Oh wait, you don't tell me... could it be that every commit of a PR must pass checks on its own and not only the tip? That would explain what you've done. I'm sorry then.

The `type` variable indeed gets unused for a while, but it should be reintroduced later. I'll do that then.

albel727:
Hmm, okay, I don't know for the love of me how to get the last travis job to succeed. I can't reproduce any errors in the integration tests locally, no matter if it's lua 5.1, 5.2 or luajit. Ideas welcome.

albel727:
I had a suspicion, thanks.

albel727:
Okay, since draw callbacks have proven to be so controversial, I've stripped this PR of them. I'll make a separate PR with them, if this one gets merged.

Elv13 (approved):
The controversial code is no longer of the PR and now it has only the parts everybody agrees with. Thanks for that.

I don't want the flame war to go on. But lets just close on a few points. My last comment was a direct reply to the previous *comment* you made. At the time, I was aware you had pushed new commits and several hundreds line of commit message a few hours prior. No, I did indeed not read them at the time. I tried to provide a timely reply given I did not have the time to read all commit message and code again, which would have taken several hours and it was 1AM already. You got to understand, I only got 2-3 hours per week for AwesomeWM, and they are spread across Reddit, Github, emails and StackOverflow. There was just too much content for me to review at the time. This isn't a critique of the PR, this is just the fact that I didn't have the bandwidth to re-review it within 1 day. It is either a long review after a couple weeks or a quick reply to the comments.

As for the force push controversy, again, I didn't remove any commit. The force-push was to fix the CI which for _that_ revision, didn't pass because the `type` variable was unused in the 3rd commit (+3 other commits that at the time has warnings). I also renamed a variable because my system luacheck ignores the `--luacheck` overrides in "strict mode". My "real" changes were in commits at the end of the tree. I did push a test commit before the force push, so I am like >99.999% sure I didn't drop any of your commit.

As for the other 3k lines that are no longer in this PR. I will probably revive them in some ways (please make that PR). The Bezier support is a very nice feature. Maybe we can take a step back and bypass the whole callback controversy. What do you think about leaving much of the callback stuff as private APIs and making a new linear graph sub-class? That way, all the naming, ordering and API guidelines you don't like don't apply. If you are fine with this that path forward, I don't mind doing it myself. 

I know contributing with other codebases you don't control and face people with different opinions and priorities is frustrating. My first few contribution to AwesomeWM  (or KDE) had to be rewritten many time to satisfy the taste and opinions of the maintainers of the time. Heck, I resigned from a Google summer of code a dozen year ago because some changes got stuck in a bikeshedding loop between multiple people and I got fed up with it. For many years I only "contributed" to my own pet projects because of that. Now I am on the other side of table. It doesn't mean I don't remember that. Also, both as a hobby and professionally, I contributed to many other projects, many more than what the GitHub summary shows, which doesn't include work email addresses. It is very rare a larger contribution gets accepted as-is first time. Projects all have guidelines are tradeoffs. In my case, I care about API consistency and not exposing the internals in the public API above all else. You care more about performance and flexibility. Those thing will clash. They are not always mutually incompatible, but in this case they were. I _do_ get many of your points, even if I don't agree with some of them. The fact that we are clashing about parameters name and order is just sad.

Anyhow, we can merge this revision. It is very high quality work and I appreciate the contribution a lot. The older graph code was in a very sad state after years and years of thing bolted on top of it. This PR really cleans that up and adds along with all the new cool stuff.

albel727:
> The controversial code is no longer of the PR and now it has only the parts everybody agrees with. Thanks for that.

You're always welcome, but I would've appreciated if you told me sooner that this was a clear way forward, instead of hanging me out to dry for a week.

> You got to understand, I only got 2-3 hours per week for AwesomeWM

I sure do understand, I've even told you so. But you'll have to understand my reaction too, after I saw that my obsessively crafted code got modified in pointless and mildly harmful ways. The better strategy would've been just communicating me your wishes and letting me implement them, if you can't dedicate to the code the advanced levels of attention that wouldn't hurt my fee fees.

> As for the force push controversy, again, I didn't remove any commit.

Nor have I accused you of that. I see your lack of time prevents you from keeping track of what's actually happened, so I'll try to keep this short, as it's not really important now anyway.

>The force-push was to fix the CI which for that revision, didn't pass because the type variable was unused in the 3rd commit (+3 other commits that at the time has warnings). 

Yes, I get it now, but it would've been better if you told me at the time that you had actually modified my commits and why. I couldn't imagine that you, for whatever reason, run CI for intermediate commits too (and now that this PR got so long it doesn't pass that job anyway, because of travis' log length limit). Having a "CONTRIBUTING" document, where that's mentioned, might or might not have helped. Would've been better still if you had noticed, that `type` does get used in the later commits (which is why I didn't remove it in the first place, i.e. to avoid the churn), and reinstated it there, but that didn't happen and it's one of the things that I meant by "mildly harmful" above.

> I also renamed a variable because my system luacheck ignores the --luacheck overrides in "strict mode". 

Assuming you're talking about the `baseline_y`, the problem is, it was one variable shadowing the other, and I've renamed the other one, which made your renaming unnecessary. I'm also not sure why you'd use a strict(er?) mode on my code, than what the rest of awesome's code, which has quite a few of luacheck overrides of its own, assumes.

> What do you think about leaving much of the callback stuff as private APIs and making a new linear graph sub-class?

I'm... not really sure what that would be? How would one *choose* a bezier rendering in such a graph without having a callback/delegate (let us finally call it by this more appropriate name) property? Or will it be more like a "beziergraph", i.e. with non-customizable bezier rendering? So one can't do anything if they happen to want something ever so slightly non-standard, e.g. to paint bars with an outline, like in `custom_draw_outline.lua` and `bezier_curve.lua`, or lines with ticks, like in `custom_draw_line.lua` and `bezier_curve.lua` again?

At the very beginning I've considered simply adding a `paint_method` property, with values like "fill" and "stroke" and let the user do their thing in `step_shape`, but I've abandoned the idea immediately, because it's too inflexible and overspecialized, even though it could be extended with e.g. "bezier" and "bezier-bar", "outline:color", etc, i.e. be """user-friendly""" of sorts. A separate linear graph class would be even less flexible than this property and they both would become redundant the very moment one figures out the proper (i.e. my) way to do the draw delegates. I'm not sure why it's so hard to recognize that deeply specialized tasks like graph drawing have to have their own specialized customization methods that don't have to be """"consistent"""" with anything but themselves. 

I can't stop you anyway, if that's the way you choose to go, but I'm convinced that something so opinionated like the "linear graph" has no place in awesome itself, but rather should be a third-party library on top of a flexible graph core. People, that would somehow get too confused by the lack of the useless `self` parameter in the draw delegates, should just use libraries written by other people or simply come here for a solution like the #2899 guy did. Imagine not having to write a graph from scratch, like you guys had to do there, but simply setting a custom delegate.

>I care about API consistency 

Repeating over and over that you care about "consistency" doesn't answer any of my questions like why it must be """"""consistent"""""" with widget draw methods and signals, and not with the shape functions. I'm sorry, but in this case this "consistency" is nothing but cargo-cultism.

>and not exposing the internals in the public API above all else. 

Yeah, you just want to force the draw callbacks assume the graph internals by forcing them to interpret the meaning of its properties. That's even worse than "exposing the internals", which my code doesn't do either, so I'm not sure why you'd mention it at all.

> I do get many of your points, even if I don't agree with some of them. The fact that we are clashing about parameters name and order is just sad.

The second sentence makes it painfully obvious, that you don't get them. It's sad that the only thing you have to reply to my technical arguments with is that "it's sad". It's sad that you reduce my arguments to "clashing about parameters name and order", i.e. bikeshedding. I'm not even particular about `options` vs `context` naming, and if """The context""" were a well-known awesome-wide API extension point to pass various options through and not an obscure undocumented black box that it is now, I wouldn't even have the slightest reason to name the parameter otherwise and all the reasons to name it `context`.

I've had enough of this, tbh. It's clear that you don't want or don't have the time to discuss things the proper way, so I'm not wasting my time anymore. Let's just somehow get this reviewed and merged at last, before I regret doing anything at all.

Aire-One (approved):
I'm happy to see the graph widget gets some love!
Thank you for the hard work you have put in this PR :heart: !

I hope the removed part will come back latter with another PR, and we'll find more time for additional discussion.

---

I'm not really fan of the "randomly placed global variable declaration" [1] and local/private functions associated to variable callback to "make the git blame easier to read" [2]. I think it just adds noise to the code base. I do understand it should help commit history reviews, but I'm not patient enough to do that on a 73 commits PR...

[1] : https://github.com/awesomeWM/awesome/pull/3326/files#diff-a7fbbb267f5d802c8978befe19471ea94047d665e04c210cf2daf710d4860756R509-R510

[2] : https://github.com/awesomeWM/awesome/pull/3326/files#diff-a7fbbb267f5d802c8978befe19471ea94047d665e04c210cf2daf710d4860756R357-R361, https://github.com/awesomeWM/awesome/pull/3326/files#diff-a7fbbb267f5d802c8978befe19471ea94047d665e04c210cf2daf710d4860756R401-R405

albel727:
@Aire-One 
[1] - you are completely right, I should've put math_min/max at the top of the file together with the other such vars. Should I do it now or in another PR? I'm hesitant about resetting the PR approvals.
[2] - yeah, I'm not very fond of it either, but the alternative is a total history rewrite from the very beginning, placing the functions at the right spots to avoid needing to introduce the variables later. Or, somewhat better perhaps, an introduction of something like graph_private table variable and declaring all hidden functions on that, so that the declaration order doesn't matter anymore. I can do either, if you want.

Aire-One:
I think the code can be merged as-is. That's why I approved it.
If you want to add new commits to address these suggestions, you are welcomed. Otherwise, this is not that important changes anyway...

Regarding the "resetting the PR approvals", it doesn't matter. The PR have to be manually merged anyway as the last Travis CI job fails (and will always fail since the issue is the PR size). We're a small team, we can "manually" keep track of approbations and what was changed since the last time someone did an approval review.

> [1] - you are completely right, I should've put math_min/max at the top of the file together with the other such vars. Should I do it now or in another PR? I'm hesitant about resetting the PR approvals.

If you can add a quick commit, it would be great.

> [2] - yeah, I'm not very fond of it either, but the alternative is a total history rewrite from the very beginning, placing the functions at the right spots to avoid needing to introduce the variables later. Or, somewhat better perhaps, an introduction of something like graph_private table variable and declaring all hidden functions on that, so that the declaration order doesn't matter anymore. I can do either, if you want.

No, please don't rewrite the history. It's not worth here, and will introduce a lot of unnecessary noises...

I don't think a `graph_private` table is the right solution either. Currently, we have "module-level-private variables" associated to callbacks. With the `graph_private`, we'll have a "module-level-private table" with its fields associated to callbacks. Sound like the same to me.
My concern was to make the code more "organized" by defining the "local functions" as _"local functions"_ and _before_ they are used.

I think a simple commit to move the functions to the right spot would be enough. But as you specified, it would break git blame on these code chunks. Is it important?  I'm honestly not sure.

albel727:
>No, please don't rewrite the history. It's not worth here, and will introduce a lot of unnecessary noises...

I've done it countless times as I've developed this stuff, not sure what would be the harm to do it again now...?

>I don't think a graph_private table is the right solution either. 

Well, if you care about "define in full before use", then yes, it wouldn't help here. But generally speaking a table with private functions lends itself well to unit testing, so I thought maybe to use this opportunity to make one for the graph widget. 

Well then, I'm going to edit me some history in order to move the min/max variables and the functions to the right spots, and then I'll make a new force push. Unless there are objections to that...?

Aire-One:
No technical argument against rebase/history rewrite/force push while PR is in a WIP status. (Even if I understand it can annoy peoples.)

Only that I considered the PR as "ready to be merged" now. It would be unfortunate to rewrite the history at this point... (and also probably time-consuming on your side)

> Well, if you care about "define in full before use", then yes, it wouldn't help here. But generally speaking a table with private functions lends itself well to unit testing, so I thought maybe to use this opportunity to make one for the graph widget.

Hm, I get what you want here, but I can see 2 issues :

* You are not supposed to unit test private methods ;
* A module level table wouldn't be accessible by unit tests (Except if you expose it? That would mean user can also monkeypatch function there... And basically, it creates a second public API for the module).

albel727:
>Only that I considered the PR as "ready to be merged" now. It would be unfortunate to rewrite the history at this point... (and also probably time-consuming on your side)

Er... I've already done it, sorry. I've run check-qa on every commit, then specs and example generation, and checked that the result is pixel-perfect equal to the previous runs, so I'm reasonably confident, that I didn't break anything. Shall I push it?

>(Except if you expose it? That would mean user can also monkeypatch function there... And basically, it creates a second public API for the module).

Well, people who abuse private API deserve to get burned if/when it gets changed, but one could technically make it so that the private table gets exported only in the test builds or something. But well, I don't insist on it and wouldn't use the table myself at the moment.

albel727:
Hmm, judging by the silence, history editing wasn't taken all that well. Oh well, here's how it would look as a separate commit then.

Aire-One:
> Hmm, judging by the silence, history editing wasn't taken all that well. 

Oops, sorry. I was out. I didn't expect you to come back that quickly!

> Er... I've already done it, sorry. I've run check-qa on every commit, then specs and example generation, and checked that the result is pixel-perfect equal to the previous runs, so I'm reasonably confident, that I didn't break anything. Shall I push it? Here's how it looks like https://github.com/albel727/awesome/tree/graph-widget-overhaul-reordered

> Oh well, here's how it would look as a separate commit then.

Honestly, both solutions works for me. If you have a preference to force push your other branch with a proper history, do it :)

A quick note regarding the "test private method" topic:
This is rather a political debate I don't want to have here. (I think) we agree there is no need to implement that, right now,  in the scope of this PR. So, let's keep the discussion to another issue/PR, if someone wants to work on the topic.
My current point of view is basically: I think we shouldn't, and IF we do anyway, it should also be done to every module the same way.

Aire-One (approved):
(If you force push your other branch I'll re-approve. Please keep us informed of your decision on the matter.)

albel727:
Sorry, I was away myself. Yeah, I think I'm obsessive-compulsive enough to force push the reordered variant. Here we go.

>My current point of view is basically: I think we shouldn't, and IF we do anyway, it should also be done to every module the same way.

Yes, complete agreement here.

Aire-One (approved):

actionless:
the amount of commits makes "test prev commits" travis setup run too long:

> The job exceeded the maximum log length, and has been terminated.

albel727:
@actionless Yeah, we've noticed. I've offered several times to split the PR in two, but nobody has taken me up on the offer, so I assumed that it's not considered to be a real problem.

actionless:
i'd rather squash it, since it was reviewed already (and you have copy of commit history just in case)

albel727:
Well, if that's the consensus, I can do it too.

actionless (approved):
i see it have test coverage and not breaking existing api (at least from the docs)

so good enough for me

of course in 2.5k diff chunk might be some potential problems but i hope you'll come back later to fix them ;)

psychon (approved):
What could possibly go wrong? :-)

Thanks for your work and sorry for the problems you encountered on the way.

PR #2867, opened 2019-09-10
Added new background images for default theme

speakk:
I made these background images in Blender 2.8, as I feel the original default theme causes a bit of eye ache.

PR #3335, opened 2021-04-26
Extend wibox.widget.imagebox to encompass the required wallpaper API features

Elv13:
This is another groundwork PR in the road toward using normal widgets to set the wallpaper. The goal of the larger endeavor is to remove some limitation of our current wallpaper API, such as the lack of SVG, without duplicating too much code.

@psychon The `scaling_quality` doesn't seem to work, any idea why? See the image. There is just no difference, while, at the very least the `nearest` one should be very visible. I did exactly what the API says to do (https://www.cairographics.org/manual/cairo-cairo-pattern-t.html#cairo-pattern-set-filter), so I am confused as to why it doesn't work. Any idea?

image

### DIFF
--- a/lib/wibox/widget/imagebox.lua
+++ b/lib/wibox/widget/imagebox.lua
@@ -72,22 +72,94 @@ end
 function imagebox:draw(_, cr, width, height)
     if width == 0 or height == 0 or not self._private.default then return end
 
-    -- Set the clip
-    if self._private.clip_shape then
-        cr:clip(self._private.clip_shape(cr, width, height, unpack(self._private.clip_args)))
-    end
+    -- For valign = "top" and halign = "left"
+    local translate = {
+        x = 0,
+        y = 0,
+    }
+
+    local w, h = self._private.default.width, self._private.default.height
 
     if not self._private.resize_forbidden then
-        -- Let's scale the image so that it fits into (width, height)
-        local w, h = self._private.default.width, self._private.default.height
-        local aspect = math.min(width / w, height / h)
-        cr:scale(aspect, aspect)
+        -- That's for the "fit" policy.
+        local aspects = {
+            w = width / w,
+            h = height / h
+        }
+
+        local policy = {
+            w = self._private.horizontal_fit_policy or "auto",
+            h = self._private.vertical_fit_policy or "auto"
+        }
+
+        for _, aspect in ipairs {"w", "h"} do
+            if policy[aspect] == "auto" then
+                aspects[aspect] = math.min(width / w, height / h)
+            elseif policy[aspect] == "none" then
+                aspects[aspect] = 1
+            end
+        end
+
+        if self._private.halign == "center" then
+            translate.x = math.floor((width - w*aspects.w)/2)
+        elseif self._private.halign == "right" then
+            translate.x = math.floor(width - (w*aspects.w))
+        end
+
+        if self._private.valign == "center" then
+            translate.y = math.floor((height - h*aspects.h)/2)
+        elseif self._private.valign == "bottom" then
+            translate.y = math.floor(height - (h*aspects.h))
+        end
+
+        cr:translate(translate.x, translate.y)
+
+        -- Before using the scale, make sure it is below the threshold.
+        local threshold, max_factor = self._private.max_scaling_factor, math.max(aspects.w, aspects.h)
+
+        if threshold and threshold > 0 and threshold < max_factor then
+            aspects.w = (aspects.w*threshold)/max_factor
+            aspects.h = (aspects.h*threshold)/max_factor
+        end
+
+        -- Set the clip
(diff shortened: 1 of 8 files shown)


actionless:
is it already fixing this one? :)) https://github.com/awesomeWM/awesome/pull/2867#issuecomment-538418756

lib/wibox/widget/imagebox.lua
```diff
@@ -254,6 +328,145 @@ function imagebox:set_resize(allowed)
...
+local defaults = {
+    halign                = "left",
+    valign                = "top",
+    horizontal_fit_policy = "auto",
+    vertical_fit_policy   = "auto",
+    max_scaling_factor    = 0,
+    scaling_quality       = "good"
+}
+
+for _, prop in pairs(defaults) do
```
sclu1034, line 454: This may be why `scaling_quality` has no effect. It's creating methods based on the value (`set_good`, `set_auto`) instead of the name.

```suggestion
for prop in pairs(defaults) do
```

psychon:
> The scaling_quality doesn't seem to work, any idea why? 

Random guess: because SVG? Does it only not work in the docs, but work in actual "production use"?

lib/wibox/widget/imagebox.lua
```diff
@@ -254,6 +328,145 @@ function imagebox:set_resize(allowed)
...
+    imagebox["set_"..prop] = function(self, value)
+        self._private[prop] = value
+        self:emit_signal("widget::redraw_needed")
+        self:emit_signal("property::"..prop, value)
```
sclu1034, line 458: This is inconsistent with `get_<prop>`. When `set_<prop>` is called with `nil`, it will emit the signal with `nil`, but a subsequent call to `get_<prop>` (even in the signal handler) will return the default.

lib/wibox/widget/imagebox.lua
```diff
@@ -254,6 +328,145 @@ function imagebox:set_resize(allowed)
...
+    valign                = "top",
+    horizontal_fit_policy = "auto",
+    vertical_fit_policy   = "auto",
+    max_scaling_factor    = 0,
+    scaling_quality       = "good"
+}
+
+for _, prop in pairs(defaults) do
+    imagebox["set_"..prop] = function(self, value)
+        self._private[prop] = value
```
sclu1034, line 456: Maybe add the usual check to avoid unnecessary redraws?

```suggestion
    imagebox["set_"..prop] = function(self, value)
        if self._private[prop] == value then
            return
        end
        
        self._private[prop] = value
```

sclu1034 (changes requested):

lib/wibox/widget/imagebox.lua
Elv13: Thanks for spotting that!

well, that sure didn't help... but unfortunately isn't the issue. In introduced this bug in a last minute refactoring.

sclu1034: Interesting. It started working for me once I had fixed that loop.

sclu1034:
> Does it only not work in the docs, but work in actual "production use"?

That seems to be correct on my end. When generating the docs locally, the example images are indistinguishable. But when I use this with sample images, there is an obvious difference between the options, both for PNG and SVG.

Although the latter is kind of weird. One of the main benefits of using vector graphics for icons is the fact that they should scale nicely. But here it becomes just as blurry or blocky as a PNG.

Elv13:
@psychon is correct, facepalm

image

Elv13:
> is it already fixing this one? :)) #2867 (comment)

I mentioned it in the first line of the PR, didn't I? I am trying to avoid making a 20 commit, 5k line PR here. The whole point of this work is to fix that comment, but it wont be in the PR, nor will it be in the next. It will come 2 PRs from now. The problem with SVG is getting the right size and, if the aspect ratio doesn't match, resize it *correctly*. If I was to fix this in the current `gears.wallpaper`, I would have to write 5 version of the same code for the different type of wallpapers. If I fix it using `awful.wallpaper`+`wibox.widget.imagebox`, I only have to fix it *once*.

The next PR adds `awful.wallpaper` itself. It wont make SVG that much better yet, but will bring parity with the current `gears.wallpaper` features (and non-features). The better SVG support will come in the last PR of this serie, which will fix the resize corner cases.

sclu1034 (approved):

actionless:
@Elv13 thanks for some clarifications on the roadplan, i think i didnt realized what its not just about fixing wallpaper api, but about writing new one (gears.wallpaper -> awful.wallpaper)

Aire-One (approved):
LGTM :)

actionless (approved):

psychon:
> @psychon is correct,

I like the sound of that. May I quote this in other contexts? ;-)

PR #3346, opened 2021-05-20
Graph draw delegates

albel727:
This is the current state of the graph widget drawing delegates as initially developed as a part of #3326. The long-winded discussions there didn't yield an agreement on their proper form, which is why this is a separate PR, but can shine light on how the delegates ended up to be the way they are, and how and why they probably still need to be changed (as I'm also not entirely happy with them as they are now). Still, here's a hopefully exhaustive summary, but before I begin it should probably also be mentioned that this PR also contains `gears.math.bezier`, an implementation of bezier curves helper library, which may or may not be better split out and merged separately.

### Summary
Graph draw delegates are graph properties that can be set by users to functions in order to allow them to customize the graph in ways beyond simply specifying the `step_shape`. There are presently three of them:

`group_start(cairo_context, group_idx, options)`
`step_hook(cairo_context, x, y, baseline_y, step_width, options)`
`group_finish(cairo_context, group_idx, options)`

These are called by the graph widget before, during, and after rendering each graph data group respectively, in order to do whatever is needed to render the data as the user wishes. The purpose and parameters of the functions are mostly self-explanatory, but a more detailed explanation can be found in the commit messages of the `Add group_start and group_finish properties to graph widget` and `Add graph.step_hook property for more flexible drawing` commits. There's also a plethora of examples of how they can be used. I'll only explain what is necessary to understand the design decisions and the points of contention.

The `options` parameter is a table of various data that could be useful for custom drawing, such as:
`_width` and `_height`: full dimensions of the graph data drawing area.
`_step_width` and `_step_spacing`: respective (non-nil) numbers.
`_group_idx`: currently drawn data group index.
`_graph` and `_self`: the calling widget itself.

This list is not final and the `options` are designed to be extendable. E.g. if some user wants to implement printing the latest value over the graph or even to label each bar with the value it represents, we could put a `_value` field into the `options`, which would be indispensable for exactly that use-case.

Not only that, but user is enabled to store and pass their own data between the delegates by putting it into the `options`, and it will be preserved for the duration of the `graph:draw()` call.

Last but not least, the `options` also contain everything that comes with the `context` parameter that is passed into `widget:draw()` methods (i.e. currently only the `dpi`).

In other words, `options` is a further development of the idea of the `context`. It's an extensible bag of data, that is potentially useful for drawing, but doesn't warrant its own separate parameter.

Now, for the contention points, which I hope to represent fairly enough.

### 1. Should the delegates have a `self` parameter?
It's argued, that this would be consistent with the `widget:draw()` methods and signal callbacks and prevent confusion among users that got used to its ubiquitous presence.

I argue, that there's barely any confusion to prevent, because this parameter isn't useful in 99% of the drawing cases and would just force everyone to write `function(_, ...)` everywhere for no benefit to anyone. The purpose and usage of the delegates is sufficiently different from `draw()` methods and signals to not demand any consistency with them. Moreover, it would encourage bad programming decisions on the part of the user, more on that later.

### 2. Should the `options` table be pruned and kept at a minimum size in favor of making the user get the respective values from the widget manually?
Looking at the contents of the `options` table, one may notice that there are many values that one seemingly could just query from the graph as needed. E.g. if one just adds the aforementioned `self` parameter, one could use `self.step_width` instead of `options.step_width`.
It's argued that we should do just that and only leave `width`, `height` and `group_idx` in the `options` table.

I argue, that such minimization isn't only pointless, but even harmful. It doesn't cost us anything to populate the table, compared to drawing hundreds of values. But more importantly, forcing users to read and interpret the previously overwhelmingly write-only properties of the graph widget will result in a lot of fragile code in the wild and forever block any useful development of the widget.

The `self.step_width`, i.e. one of the user-settable customization properties, only looks like it's the same as `options.step_width`, i.e. the width of the currently drawn step in pixels. But it actually isn't. What if we'll want to extend the  `self.step_width` to be settable to a table or a function, in order to be able to define the width for each data group separately? What if we'll want to add a `step_num` property, that overrides `self.step_width` and makes the graph always draw a given number of (automatically resized) steps, no matter with which width the graph is being drawn? What if we'll implement optional rounding of `step_width` so as to keep the drawn edges sharp? Etc, etc.

All of these obviously useful extensions will break all draw delegates that falsely believed `self.step_width` to always be the desired number of pixels. In the absence of `options.step_width`, which will indeed always be exactly that number and nothing else, the user will be forced to query an unspecified set of graph's properties and fully duplicate the graph's logic that calculates it.

Which is why I argue, that the `options` table should be populated as exhaustively as possible and the `self` parameter should be absent, in order to minimize the temptation for the user to abuse the graph properties for drawing purposes. Ideally, the draw delegate shouldn't even know or care, from what kind of graph/widget it's called from.

### 3. Should the keys of the `options` table start with an underscore?
It's argued that this is against the decided-upon conventions of the awesome's codebase and will only confuse the user and add to the mishmash of legacy coding styles that people are painfully trying to get rid of even now.

My relatively weak argument for the underscore: the `options` table is a merger of an undefined number of values originating from: the draw `context`, the graph widget itself, and whatever the user's draw delegates wish to put there.

Even if there aren't any key clashes now, It would be prudent to do something to minimize them, otherwise one risks getting an ugly problem that can't be satisfactorily solved. E.g. if the `context` accidentally happens to be extended, say, with a `width` field, it will be overwritten by the `width` coming from the graph, and the user draw delegates will be unable to recover from it, and the addition of `width` to the `context` could be no longer revertible at that point, because that would be an API breakage.

One can imagine a similar situation with one of draw delegates, originating from different authors, accidentally overwriting some option field, which the other one relies on. Thus it may make sense to dedicate underscore-prefixed variables to the graph and leave unprefixed variables for users to use however they want.

### 4. Should the `options` table just be named `context`?
It's argued, that it's merely an extension of the draw `context` and should be named accordingly.

I agree in principle, but the problem is that such extension is unprecedented in the awesome's codebase, and, as far as I can tell, the draw `context` isn't documented anywhere at all and until now contained only things of fringe usefulness (dpi), so the less sophisticated user is conditioned to simply ignore the parameter outright, which might force some users to seek some hacky direct access to the graph instead of using values from the `options` table, which is undesirable as I've explained above.

Therefore I'd say that these are the necessary preconditions for the renaming:
- documenting the `context` and its fields in some dedicated place.
- mentioning in that documentation, that `context` can be extended and filled by widgets with arbitrary custom data that is valid for the duration of the `draw()` call.
- listing in that documentation all widgets that do such extension (presently only the graph widget, afaik), so that one will remember to consider all possible name clashes when extending the `context` or the widgets custom datasets.
- hyperlinking to the documentation in all places that use the `context` parameter. the user shouldn't get trained to perceive the `context` as a black box.
- making such hyperlinking in a way that immediately makes the user aware of the custom data added by the current widget. i.e. a parameter should link to a table of custom fields first, and only then should the custom table link to the common `context` documentation, so that the user doesn't accidentally miss the existence of the custom fields.

### 5. Should the `options`/`context` table be moved from the back to the front in the parameter list?
The motivation is the same as with `self`, namely the consistency with the `widget:draw(self, context)` methods. An additional argument for the move is that the presence of `options` at the end makes it awkward to add new parameters to the delegates after it.

I argue that since the `self` parameter shouldn't be present, any consistency is moot anyway, and since `self` and `options` are much less useful than the rest, we'd just be forcing the user to write `function(_, _, ...)` everywhere.

And what concerns the awkwardness, I say, that's the whole point. Any new parameters should be added as fields of `options` instead. The user should be able to fearlessly add new parameters to their delegate functions after `options` for their own purposes, like toggling debugging, different drawing modes, etc., which means that we're not allowed to pass any additional arguments that the user themselves hasn't explicitly asked us to add, as doing that will be an API breakage.

### 6. Should `step_width` be present as a parameter of `step_hook()`, when it also exists inside the `options`?
It's argued that this is redundant.

I tend to agree, but it's not completely clear cut. My initial reason for introducing it was inertial thinking and the fact that `step_width` is 99.9%-likely to be used by the delegate, so moving it into `options` table will incur one slower table lookup instead of a fast argument lookup for every drawn value. I believe now that the performance hit isn't worth talking about in presence of other Lua/LGI/Cairo inefficiencies.

My argument for removing the parameter is that if `options.step_width` becomes the one and only source of truth, a `group_start` delegate will be able to reliably trick the `step_hook` delegate into drawing steps of different widths by modifying the value in the table, which might be useful somehow. E.g. it allows to implement the aforementioned "fixed number of resizable steps" functionality without having to wait for us to oblige.

My qualms about removing the `step_width` parameter are that it will make the `step_hook` delegate be much less like a generic shape (which, abstractly speaking, always has a width, as well as x, y, height and other relevant options), and thus the delegate will lose the hope of becoming something more general and will forever remain a one-off thing for the graph widget only.

### And more
There are many other undiscussed points to consider, like the exact set of parameters of each of the delegates, what overriding them exactly means, etc. Should `group_start`/`group_finish` have the `group_idx` parameter at all, or maybe it should be present only in the `options`, so that potential implementations could be reused in other graphs that don't have the concept of data groups? Should `step_hook` remain specialized for bar drawing like it is, or should it rather be generalized to a new shape API, that wouldn't have the problems of the existing shapes that prevented them from being used in `step_hook`? Etc, etc. As unlikely as it is to be considered, I had to mention it here too for posterity, because I doubt that I'll keep working on this.


Aire-One:
I didn't take the time to review yet (and it was a part I didn't reviewed before it was removed from the other PR), but I'm happy to see this code coming back with this PR :)

I'll try to review this weekend or beginning of next week.

Thanks for your work!

albel727:
I'm not sure it needs a review as of yet. I was under impression that Elv13 will be doing something based on it, which may or may not result in a separate PR. Meanwhile I'm just trying to fix the TEST_PREV_COMMITS build job for it.

actionless:
travis being a bit mysterios with its broken log (quite ironically what they started to ask moneyz about the same time as it got broken - prolly dont have enough workpower to fix that :D)

but seems what luacheck is failing on this one https://github.com/awesomeWM/awesome/pull/3346/commits/ff882f1d80cf96a7a5829a431dd6d3a50b35d818

albel727:
I assume travis just times out after approximately an hour. There's actually nothing in any of the commits to fail on. After fixing an actual issue that is not really my fault (using globs in cmake, I assume), lots of my force pushes were just about trying to make the history smaller by merging commits together so that it takes less time, but apparently down to 13 from 22 commits is still too much for travis.

actionless:
now lua check seems to fail on d051a340bc31780191756dde72c0deb241525783
25m doesnt looks like a possible timeout

i recommend either running CI locally (for getting non--travis-truncated log) or squashing the commits

albel727:
> running CI locally

Doesn't sound too easy with Travis, tbh.

>25m doesnt looks like a possible timeout

Yeah, it doesn't. But as you may have noticed, the previously "failing" https://github.com/awesomeWM/awesome/commit/ff882f1d80cf96a7a5829a431dd6d3a50b35d818 commit has succeeded, being a part of the second job. It's either Travis or the awesome's build system being flaky.

actionless:
for git-bisect-ing

actionless:
and also undo the travis/test-related changes, as they pointless/conflicting with https://github.com/awesomeWM/awesome/pull/3292

albel727:
That's only remotely a problem with automated bisecting. The bulk of searching and reverting will have to occur on whole PRs, i.e. on merge commits anyway. And if one wants to dig deeper, there's git bisect skip.

albel727:
>and also undo the travis/test-related changes, as they pointless/conflicting with #3292

I wish I'd known sooner that travis was a dead end. Well, this PR isn't here to be merged anyway, I was just trying to make sure that whoever starts developing off of it can do it without having to fix intermediate commits.

actionless:
> I was just trying

thats out of scope of the PR for graph widget delegates

any improvements in separate PR would be appreciated

albel727:
Trying to check that all commits of a PR actually can pass Travis is in scope of any PR. I'd have dropped the Travis changes once I got a green light from it.

PR #3348, opened 2021-05-22
test(spec: beautiful: get_font): convert user object to string before comparison

actionless:


sclu1034 (approved):

actionless:
idk how it was passing before, but my guess is what smth changed what either `luassert` instead of implicitly calling `OBJ:to_string()` of that object now is calling internal lua's `tostring(OBJ)`

 either in  `lgi` some code which was directing lua's tostring() to object's :to_string() is gone

actionless:
or third guess, due to some internal caching in lgi or gobject itself, previously FontDescription object which was returned was literally the same in memory, so it's memory address was the same so lua's tostring() was showing the same representation

Aire-One (approved):

Aire-One:
@Sorky (sorry for calling you out this way), I can see with git-blame that you wrote this test 2 years ago... Is there something we don't understand here? Are the @actionless hypothesis correct?

Elv13 (approved):

Elv13:
The test has failed a few time before, but for some reason it's like <1% of the time. Thanks for looking into it. I should have done that instead of restarting the build

actionless:
if it was from time to time before -- i guess that internal caching in gi/lgi

PR #3356, opened 2021-05-31
fixed: Try to handle zero sized widgets again.

### LINKED ISSUE #3351: Run prompt doesn't show

```
awesome v4.3-1030-gb65025ef6 (Too long)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

1. launch awesome
2. press the shortcut for opening the run prompt (Super+r)

**Actual result:**

The run prompt doesn't show even after typing something into it. Only doing an update such as clicking the corner icon that opens the menu makes the run prompt show.

**Expected result:**

The run prompt should be shown immediately.

Elv13:
The newly changed code doesn't handle this well:

    local w = wibox.widget {
        {
            --add anything here,
            {
                text = "",
            },
            {
                id = "mytextbox",
                text = "",
            },
            {
                text = "",
            },
            widget = wibox.layout.fixed.horizontal
        },
        widget = wibox.layout.fixed.horizontal,
    }

    awesome.emit_signal("refresh")

    w:get_children_by_id("mytextbox")[1].text = "something"

This will cause the "inner" fixed layout to have the minimum size
it supports. In that case, if the last widget has "no size" because
it supports up to 0x0, then it isn't added to the layout.

This was done "on purpose" because if there is a spacing, then `:fit`
would have returned a size "too small" because the last spacing area
would be (correctly) missing.

But if the zero sized widget isn't added to the layout, then it's size
isn't tracked. So if it emits a layout_changed signal, nothing catches
it.

The "fix" is rather hacky and probably a little incorrect. It rely
on the behavior of `:fit()` to avoid adding the "wrong" widgets to
the layout, which is fragile.

However, I don't have a better idea. @ShayAgros any alternative to this hack?

Fix #3351

### DIFF
--- a/lib/wibox/layout/fixed.lua
+++ b/lib/wibox/layout/fixed.lua
@@ -33,39 +33,71 @@ function fixed:layout(context, width, height)
     spacing_widget = spacing ~= 0 and self._private.spacing_widget or nil
 
     for k, v in pairs(self._private.widgets) do
-        local w, h = width - x, height - y
+        local w, h, local_spacing = width - x, height - y, spacing
+
+        -- Some widget might be zero sized either because this is their
+        -- minimum space or just because they are really empty. In this case,
+        -- they must still be added to the layout. Otherwise, if their size
+        -- change and this layout is resizable, they are lost "forever" until
+        -- a full relayout is called on this fixed layout object.
+        local zero = false
 
         if is_y then
             if k ~= widgets_nr or not self._private.fill_space then
                 h = select(2, base.fit_widget(self, context, v, w, h))
+                zero = h == 0
             end
 
             if y - spacing >= height then
                 -- pop the spacing widget added in previous iteration if used
                 if spacing_widget then
                     table.remove(result)
+
+                    -- Avoid adding zero-sized widgets at an out-of-bound
+                    -- position.
+                    y = y - spacing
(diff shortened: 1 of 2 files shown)


lib/wibox/layout/fixed.lua
```diff
@@ -33,39 +33,71 @@ function fixed:layout(context, width, height)
...
-                break
             end
         else
             if k ~= widgets_nr or not self._private.fill_space then
                 w = select(1, base.fit_widget(self, context, v, w, h))
+                zero = w == 0
             end
 
-            if x - spacing >= width then
+            if x - spacing >= width and not zero then
```
sclu1034, line 73: This check is different to the one in the `is_y`-branch. Is that intended?

Elv13: fixed

sclu1034 (approved):

actionless (approved):

Aire-One (approved):
LGTM, I don't think there is any reason to hold this PR any longer.

@ShayAgros, your comment here would still be appreciated 😉

PR #3362, opened 2021-06-10
Fix the hack for multibyte characters in prompt (fixes #3308)

### LINKED ISSUE #3308: Deleting one Cyrillic symbol in run prompt requires two Backspace keystrokes

awesome v4.3 (Too long)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

* open run prompt with hot key (M+r)
* switch keyboard layout to Cyrillic (Caps Lock)
* write any Cyrillic symbols
* hit Backspace two times

**Actual result:**

One symbol is deleted.

**Expected result:**

Two symbols are deleted.

actionless:


actionless:
i think it might fix not only #3308 but also #974

actionless:
although since lua doesnt have normal unicode support - fixing almost each prompt hotkey would require separate effort :)

actionless:
but at least people who still experience problems with other prompt hotkeys could use this PR later as a reference for theirs :)

actionless:
damn coveralls

Aire-One (approved):
I can read in the linked issues that GLib can manage UTF characters but LGI is not cooperating. Does the situation evolve since then? It's maybe better to relay on GLib API.

Anyway, this will be good enough. Thanks for this fix!

lib/awful/prompt.lua
```diff
@@ -285,10 +289,14 @@ local function prompt_text_with_cursor(args)
         text_start = gstring.xml_escape(text)
         text_end = ""
     else
-        char = gstring.xml_escape(text:sub(args.cursor_pos, args.cursor_pos))
+        local offset = 0
+        if have_multibyte_char_at(text, args.cursor_pos) then
+            offset = 1
```
psychon, line 294: can't utf8 sequences take up to four bytes (or even more)? This seems to only handle one or two bytes.

actionless: previous code was handling only one-byte

i suppose users of 3-byte languages could also make the changes later on

sclu1034: Yeah. A single code point can be up to four bytes long, and certain characters are built from two consecutive code points.

I can confirm that this breaks with 3-byte or 4-byte characters (tested with U+1000 `က` and U+10000 `𐀀`):
```
Error parsing markup: Error on line 1 char 83: Invalid UTF-8 encoded text in name — not valid “� '
stack traceback:
        /home/lucas/projects/awesome/lib/wibox/widget/textbox.lua:193: in function 'wibox.widget.textbox.set_markup'
        /home/lucas/projects/awesome/lib/awful/prompt.lua:566: in upvalue 'update'
        /home/lucas/projects/awesome/lib/awful/prompt.lua:927: in local 'keygrabber_function'
        /home/lucas/projects/awesome/lib/awful/keygrabber.lua:146: in function </home/lucas/projects/awesome/lib/awful/keygrabber.lua:143>
```

The ranges for single code points are

| First code point  | Last code point  | Length  |
|:-----------------:|:----------------:|---------|
|           U+0000  |          U+007F  | 1 Byte  |
|           U+0080  |          U+07FF  | 2 Bytes |
|           U+0800  |          U+FFFF  | 3 Bytes |
|          U+10000  |        U+10FFFF  | 4 Bytes |

I would assume that the set of characters that are built from multiple code points is defined as well.

actionless: also, as i mentioned in previous comments, this PR implement multibyte support only for some of the keybindings, so it gonna serve as an example, for people who would want to either add multibyte support for more prompt hotkeys or more bytes

btw, the pcall thingy wasnt performing well enough even as a workaround -- update() function is called from several places, and pcall-ing one of them was hovering only some of the prompt hotkeys

actionless: to make it clear - i'm not going to debug tests `Assertion failed □ is not equal to □`

Elv13:
image

Some progress on the doc examples. It doesn't pass the test and this is still pretty far from complete. The multi screen panning and more complex SVG corner cases (dpi, panning, stretching) are not handled yet (and the svg stretching wont be in the PR, this is already large enough).

@actionless Do you think you could refactor your hotkey popup to expose a real widget more cleanly? I would like a add some examples where the keybindings are part of the wallpaper itself. However, the current API doesn't make it easy to get the widget without creating the wibox.

Elv13:
(whoops, I pasted a comment in the wrong page)

Aire-One:
@Elv13 I think you meant your last message to your wallpaper API PR 😅

actionless:
@Elv13 and what you were going to write on this page? ;)

ViStefan (commented):
How about the rest keyhandlers that brake multibyte chars too, like `^D`, `delete`, `Left`, `Right`? I'm going to fix them soon, but I'm afraid it will be only for lua5.3 where `utf8` support introduced.

lib/awful/prompt.lua
```diff
@@ -711,12 +722,20 @@ function prompt.run(args, textbox, exe_callback, completion_callback,
                 end
             elseif key == "f" then
                 if cur_pos <= #command then
-                    cur_pos = cur_pos + 1
+                    if have_multibyte_char_at(command, cur_pos) then
+                        cur_pos = cur_pos + 2
```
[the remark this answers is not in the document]
actionless: > I think we should

PRs are welcome

Elv13:
> @Elv13 and what you were going to write on this page? ;)

I pasted a comment on the wrong thread. But about this, I am just not sure what's the way forward here. I don't feel we are in a very good shape to not regress things. You added a lot of test, that's super appreciated, but any changes to the input code is ultra scary to me. Last time I tried to improve this, Psychon laughed at me for all the corner cases I missed and closed the PR.

If I had infinite time and energy, the entire input stack would be rewritten from scratch using libinput and other 3rd party code rather than try to split bytes and play with GLib and Lua strings. That would also give us mouse/touchpad gestures. That that _a lot_ of work. Months worth of full time work. Even there, it would still scare me.

actionless:
the workaround which is currently merged is exactly the reason why i'm replacing it with a more structured approach, and test-covered one (and i'm not hiding the fact what it's not covering a lot of cases), the main point what it doesn't make things **worse** than the current one:

1) the main reason why - because pcall inside current workaround is not actually doing anything, as exception is never being raised (`textbox:set_markup_silent` would do that instead but we are talking about what it is, not what could be)

2)because even if it would be raised and caught - the pcall is located only at one of 3 calls of that function, so it works only from time to time

3)in this PR it's at least works for limited cases, while current workaround (if it would work despite pt2 and pt3) would just clean up the whole line

4) i could add working variation of that original pcall idea, but do it properly, but not instead of current workaround (and if somebody like vimstefan would later rewrite it properly - there is no problem to remove this newly added workaround)

P.S. i think you messing up libreadline and libinput here (or again commenting on another issue mb? :) )

psychon:
> Psychon laughed at me for all the corner cases I missed and closed the PR.

I'm sorry :-(

Elv13 (approved):

Elv13:
> P.S. i think you messing up libreadline and libinput here (or again commenting on another issue mb? :) )

No, I really mean libinput. While there is a X driver for it, there is also ways to use it directly (which is what most wayland compositor do). Right now, when you switch keyboard layouts, there are still ways Awesome fail to remap keybindings and other problems like that. With libinput, you also get some ways to do touchpad gestures (but I so far failed to make it work and/or find same APIs wrapper for it).

Anyhow, I approved the change, let give this a go and see if the world ends.

> I'm sorry :-(

But you were also right ;). The rewrite where I actually used XCB APIs to map keycode to keys was definitively and improvement over the naive byte splitting.

actionless:
@Elv13 i thought you were speaking more general about prompt logic, as readline would also handle all those emacs-style keybindings like "^W"

actionless:
> let give this a go and see if the world ends.

as i mentioned before, it shouldnt make things worse, but at the same time i expect some of the issues to be reopened with "ok Backspace is working now, but when i press Left arrow things are still the same" or so

PR #3363, opened 2021-06-11
Stop timer when stopping keygrabber

sclu1034:
When stopping a keygrabber with a timeout manually or through the `stop_key`, the timer would continue and call the stop callback again some time later.

The error message in `gears.timer:stop` is removed, since there actually is no harm in just returning immediately. And the timer implementation itself calls `:stop` in certain places without checking for `.started`, which lead to a situation where the internal call to `stop` triggered
the error message.

Alternatively, I could add the `if self.started` check where the timer calls `:stop` on itself, if you prefer to keep the error message (e.g. https://github.com/sclu1034/awesome/blob/master/lib/gears/timer.lua#L188-L190).

**Minimal test `rc.lua`**
```lua

local keypress = function(_, mod, key, event)
    print(event .. ": " .. key)
end

local grabber = awful.keygrabber{
    stop_event = "press",
    stop_key = "Escape",
    stop_callback = function () print("Escape key pressed, in stop callback ...") end,
    timeout = 5,
    timeout_callback = function () print("timed out") end,
    keypressed_callback = keypress,
    keyreleased_callback = keypress
}

grabber:connect_signal("started", function() print("keygrabber started") end)
grabber:connect_signal("stopped", function () print("keygrabber stopped") end)

awful.keyboard.append_global_keybindings({
    awful.key({ "Mod4" }, "F2",  nil, function()
        print("start")
        grabber:start()
    end),
})
```

### DIFF
--- a/lib/awful/keygrabber.lua
+++ b/lib/awful/keygrabber.lua
@@ -474,11 +474,19 @@ function keygrabber:start()
 end
 
 --- Stop the keygrabber.
+--
+-- Also stops any `timeout`.
+--
 -- @method stop
 -- @emits stopped
 -- @emits property::current_instance
 function keygrabber:stop(_stop_key, _stop_mods) -- (at)function disables ldoc params
     keygrab.stop(self.grabber)
+
+    local timer = self._private.timer
+    if timer and timer.started then
+        timer:stop()
+    end
 
     if self.stop_callback then
         self.stop_callback(
--- a/lib/gears/timer.lua
+++ b/lib/gears/timer.lua
@@ -99,11 +99,13 @@ function timer:start()
 end
 
 --- Stop the timer.
+--
+-- Does nothing if the timer isn't running.
+--
 -- @method stop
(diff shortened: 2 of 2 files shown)


Aire-One:
> The error message in gears.timer:stop is removed, since there actually is no harm in just returning immediately. And the timer implementation itself calls :stop in certain places without checking for .started, which lead to a situation where the internal call to stop triggered the error message.

I honestly have mixed opinion about that... Another approach, as you said,  would be to always test if the timer is indeed started when we call `:stop`. Hm... It's probably unnecessary as the timer has the logic to do that by itself. I also tend to think we should keep the API the easiest possible. Force the user to first check the timer status at every call of `:stop` seems to overcomplicate something that should be a simple call (apparently, we don't even do that ourselves!). Yeap, I actually think I'm favorable to remove this error message.

Aire-One (approved):
I think this is alright.

Elv13 (approved):

PR #3364, opened 2021-06-12
[doc] awful.key contructor and general improvements

Aire-One:
On my way to rewrite my config, I tried to use the new (not that new now...) declarative pattern for my keybindings definition.
This is some quick fixes to improve the documentation.

1. The documentation for the declarative constructor was wrong on some parameters type.
2. I also added the `awful.key.keygroup` table. This "property" was present in the documentation, but wasn't even defined in the code. The original objective for this property was to document the possible values available for keygroups. I defined this property as an "enum style table" the same way we have `awful.buttons.names`.
3. Finally, I have added some general module level documentation to present the different ways to instantiate an `awful.key` object.


lib/awful/key.lua
```diff
@@ -99,6 +129,32 @@ local key = { mt = {}, hotkeys = {} }
...
+-- - **NUMPAD** = `"numpad"`: The number keys on the keypad to the right of the letters and
+-- the arrow keys. Not present in every keyboard.
+--
+-- @table keygroup
+key.keygroup = {
+    NUMROW = 'numrow', -- The number row.
+    ARROWS = 'arrows', -- The directionnal arrows.
+    FKEYS = 'fkeys', -- The function keys.
+    NUMPAD = 'numpad', -- The numpad keys.
+}
```
actionless, line 156: i wonder how this was working before

lib/awful/key.lua
```diff
@@ -1,6 +1,51 @@
...
+--    awful.key {
+--      modifiers = { "Mod4", "Shift" },
+--      key = 'a',
+--      on_press = function ()
+--         print("The `Mod4` + `Shift` + `a` combo is pressed")
+--      end,
+--      on_release = function ()
+--         print("The `Mod4` + `Shift` + `a` combo is released")
+--      end
+--    }
```
actionless, line 28: would be better to add `description` field to the example

actionless (approved):

lib/awful/key.lua
Aire-One: You had to specify the literal string by yourself `"numrow" ` for example instead of `awful.key.keygroup.NUMROW`.

We still can use literal string, this is just a table that hold these strings. It makes the code feels less "random" as the user can now use this "enum".

It's exactly what was done before for the mouse buttons, we have the `awful.button.name` table to replace the use of mouse button indexes.

lib/awful/key.lua
Aire-One: I wasn't totally sure how to document this part. `awful.key` can "save" additional parameters as data-fields. Keybindings uses them for the `description` and `group` data-fields. But other than that, it's not really a part of the key API...

Hm... Actually, you're right, we should have a mention for this anyway. I'll add that!

lib/awful/key.lua
actionless: in that case, i don't see in the diff the changes which are removing the old ones

Aire-One: Hm... Sorry, it wasn't clear.

Currently, the doc have: https://awesomewm.org/apidoc/input_handling/awful.key.html#keygroup

```
keygroup N/A

    A group of keys.

    The valid keygroups are:

        numrow: The row above the letters in the US PC-105/PC-104 keyboards and its derivative. This is usually the number 1-9 followed by 0.
        arrows: The Left/Right/Top/Bottom keys usually located right of the spacebar.
        fkeys: The keys F1 through F12 located at the topmost row of any keyboard, plus F13 through F35 on specialist keyboards.
        numpad: The number keys on the keypad to the right of the letters and the arrow keys. Not present in every keyboard.


```

There is, however, no implementation in code for this property (https://github.com/awesomeWM/awesome/blob/master/lib/awful/key.lua#L27-L40). (BTW, it is documented as an object property (ie instance field) which is a lie as an instance doesn't have the field.) This property documentation is here to list the possible (string) values a user can use to refer a keygroup:

```lua
awful.key {
      modifiers = { "Mod4" },
      keygroup = "numrow", -- See here the literal string to refer to the numrow keygroup
      description = "only view tag",
      group = "tag",
      on_press = function (index) ... end
}
```

(The technical explanation to use a string here is that the table defining the keygroups uses these string as key (https://github.com/awesomeWM/awesome/blob/master/lib/awful/key.lua#L305-L363). So the string can be used latter as a key/an index to get the keygroup (https://github.com/awesomeWM/awesome/blob/master/lib/awful/key.lua#L377-L378))

The table I have defined here doesn't change any technical aspect of the current implementation. It only makes the doc less misleading and gives the user this "enum like table" I like:

```lua
awful.key {
      modifiers = { "Mod4" },
      keygroup = awful.key.keygroup.NUMROW, --  It feels less clumsy
      description = "only view tag",
      group = "tag",
      on_press = function (index) ... end
}
```

Note that you can still use the literal strings as parameter as the table basically only hold them.

actionless: ah, i see, thanks a lot for the detailed clarification :)

lib/awful/key.lua
Elv13: If you get some time, can you move that code to `tests/examples/text/`? This way (hopefully), it gets executed and make the build fail if there are syntax errors.

Aire-One: Another notable benefice is that the indentation levels are easier to make right without the leading `--` + extra spaces from ldoc code block.

lib/awful/key.lua
```diff
@@ -9,43 +9,20 @@
...
---         local tag = screen.tags[index]
---
---         if tag then
---             tag:view_only()
---         end
---      end
---    }
+-- @DOC_text_awful_key_constructor_keygroup_EXAMPLE@
 --
+-- hello
```
actionless, line 25: oh hi :3

Aire-One:
I moved the inline usage examples to example files as suggested, but something wasn't working as expected...

When introducing an error in the example code, it still builds the doc (with the introduced error). To fail the `make`, I have to explicitly call the failing example `make run-tests-examples-text-awful-key-constructor-keygroup.lua`:

```
Running make run-tests-examples-text-awful-key-constructor-keygroup.lua in build…
Result: 1
Error:
/usr/bin/lua5.3: ...e/tests/examples/text/awful/key/constructor/keygroup.lua:8: attempt to index a nil value (global akey)
stack traceback:
        ...e/tests/examples/text/awful/key/constructor/keygroup.lua:8: in main chunk
        .../documents/prog/awesome/tests/examples/text/template.lua:5: in main chunk
        [C]: in ?
make[4]: *** [CMakeFiles/run-tests-examples-text-awful-key-constructor-keygroup.lua.dir/build.make:70: CMakeFiles/run-tests-examples-text-awful-key-constructor-keygroup.lua] Error 1
make[3]: *** [CMakeFiles/Makefile2:8953: CMakeFiles/run-tests-examples-text-awful-key-constructor-keygroup.lua.dir/all] Error 2
make[2]: *** [CMakeFiles/Makefile2:8960: CMakeFiles/run-tests-examples-text-awful-key-constructor-keygroup.lua.dir/rule] Error 2
make[1]: *** [Makefile:4212: run-tests-examples-text-awful-key-constructor-keygroup.lua] Error 2
make: *** [Makefile:45: run-tests-examples-text-awful-key-constructor-keygroup.lua] Error 2
```

`make`, `make ldoc` and `make all` just all work and push the erroneous code to the doc. Is there something that I am missing? Shouldn't the example be run before pushed to the doc?

lib/awful/key.lua
Aire-One: :joy:

actionless:
i have no time to look more into this, but mb smth in `tests/examples/shims/key.lua` is messing up with it

Elv13:
This may be a regression from when `--DOC_GEN_IMAGE` and the `--DOC_GEN_OUTPUT` options were added. Before, all the examples were executed all the time and it annoyed some people. It might just be that cmake ignores them because they don't produce an artifact. Maybe we need a `--DOC_GEN_NOTHING` to create a empty file and depend on the directory in another target?

In any case, I am fine merging this. Fixing the cmake dep graph is out of scope for this PR

Elv13 (approved):

actionless (approved):

Aire-One:
@Elv13, do you think it would make sense to add fake_input to the examples and test the prints?

Aire-One:
Hm codecov complains examples don't have coverage... How do I fix that?

Elv13 (approved):
> @Elv13, do you think it would make sense to add fake_input to the examples and test the prints?

There is probably a bug in the cmake code and the text example never actually executed unless they have a generated print output. But this is probably out of scope for this PR.

PR #3366, opened 2021-06-13
[RFC,doc] Change author mention

Aire-One:
This was discussed multiple time in the past. The objective with this PR is to have a place to discuss this change.

Here is a first (very basic) proposal: Change the mention "from Author" to "Originally authored by" and add a link to the GitHub contributors list.

Other implementation can use git-blame to get the real list of contributors for a given file/class. (I have no idea of where/how to do that, tho)

Another proposed solution would be to completely remove the Author field from the info section.

Pick your poison, add more propositions (implement them?), let's discuss all of this! :)

### DIFF
--- a/docs/ldoc.ltp
+++ b/docs/ldoc.ltp
@@ -231,7 +231,18 @@
     <h3>Info:</h3>
     <ul>
 #     for tag, value in module.info:iter() do
-            <li><strong>$(tag)</strong>: $(M(value,module))</li>
+#           if tag == 'Author' then
+                <li>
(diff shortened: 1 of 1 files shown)


Elv13 (approved):

actionless (approved):
indeed, a very long-discussed topic, but everyone was too lazy to do anything, thanks a lot for looking into this finally!

actionless:
although "Copyright" field is still not really clear yet

actionless:
mb both Author and Copyright actually should be replaced with this label, like "Original Author and Copyright: ....."

Issue #3371, opened 2021-06-17
wibox.widget.base:set_opacity emits "widget::redraw" instead of "widget::redraw_needed"


ezz666:
Same in both dev and release version.

`wibox.widget.base:set_opacity` emits the `"widget::redraw"` signal instead of the `"widget::redraw_needed"` signal.
As a result when opacity changes the difference  is not seen unless something else triggers redraw.
 
Here is the [problem](https://github.com/awesomeWM/awesome/blob/master/lib/wibox/widget/base.lua#L263).

Aire-One:
I don't have the time to do it right now... 

I don't think we actually have a `"widget::redraw"` signal at all, so it might be a typo. Maybe we should have a unit test for this?

actionless:
indeed looks like a typo:

```console
$ grep widget::redraw . -R | grep -v redraw_needed
./lib/wibox/widget/base.lua:        self:emit_signal("widget::redraw")
```

PR #3375, opened 2021-06-24
[Fix/3371] wibox.widget.base set opacity emit "widget::redraw_needed" 

Aire-One:
Hello,

I have also added some unit tests to be sure all setters from `wibox.widget.base` calls their signals.
About that, I had a weird issue where signals are correctly called when the test uses the setter method but wasn't working by setting the property directly (i.e. `:set_visible()` vs `visible = `). Is there a particular reason for gears.object meta-setter not being called in unit tests? (Maybe I should have waited for a GLib event loop to passes...?)

Fix: #3371

### DIFF
--- a/lib/wibox/widget/base.lua
+++ b/lib/wibox/widget/base.lua
@@ -260,7 +260,7 @@ end
 function base.widget:set_opacity(o)
     if o ~= self._private.opacity then
         self._private.opacity = o
-        self:emit_signal("widget::redraw")
+        self:emit_signal("widget::redraw_needed")
     end
 end
 
--- a/spec/wibox/widget/base_spec.lua
+++ b/spec/wibox/widget/base_spec.lua
@@ -87,6 +87,65 @@ describe("wibox.widget.base", function()
             assert.is_true(called2)
         end)
     end)
+
+    describe("Setters emit signals", function()
+        local widget
+        local signal_called
+        local callback = function () signal_called = true end
+
+        before_each(function()
+            widget = base.make_widget()
+            signal_called = false
+        end)
+
+        it("'set_visible' calls 'widget::layout_changed'", function()
+            widget:connect_signal("widget::layout_changed", callback)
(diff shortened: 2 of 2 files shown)


actionless:
you need to squash the commits

Elv13:
> you need to squash the commits

Why? I mean, in case the test and code depend on each other, yeah (bisect issues), but otherwise, I rather have testing data and logic change in their own commit.

actionless:
because test prev commiits setup is failing on travis

Aire-One:
I don't get it... Why wouldn't the test prev commits work on this 2 commits PR, and just stops after 3 minutes, while it can run for more than 10 minutes with other PR with more commits? What's the thing that can prevent this step to complete itself?

(At this point, should we consider dropping the prev commits step on Travis CI? And put more effort in completing the migration to GitHub Actions?)

Aire-One:
lol, test prev commits still stops itself without apparent reason...

sclu1034 (approved):
A slightly more idiomatic way to implement these test could be using [Spies or Stubs](https://olivinelabs.com/busted/#spies-mocks-stubs), but this works just as well.

actionless:
it's also the only one travis setup which run `make check-qa`, so mb luacheck is not happy with smth (but we can't see it due to travis log truncation bug)

actionless (approved):
:shamrock: with some luck it passed finally

actionless:
aparently it was awful-placement test again (but we would never know due to truncated travis log)

Aire-One:
Yeay it finally passed! 🎉

> A slightly more idiomatic way to implement these tests could be using [Spies or Stubs](https://olivinelabs.com/busted/#spies-mocks-stubs), but this works just as well.

Thanks for your link, I wasn't aware busted had this API. I basically based my unit test on existing tests from gears.object.

Elv13 (approved):

PR #3387, opened 2021-07-06
Stop unsing on fixed row number in calendars

### LINKED ISSUE #3342: Calendar have hardcoded amount of rows
Which results in empty rows, if number of weeks in the months is less than maximum possible number of weeks in a month, it's even visible on the generated screenshot in the docs:

2021-04-29--1619661233_1181x678_scrot

mattplm:
This fixes #3342

### DIFF
--- a/lib/wibox/widget/calendar.lua
+++ b/lib/wibox/widget/calendar.lua
@@ -139,8 +139,23 @@ end
 -- @tparam number|nil date.day Date day
 -- @treturn widget Grid layout
 local function create_month(props, date)
-    local num_rows    = 8
-    local num_columns = props.week_numbers and 8 or 7
+    local start_row    = 3
+    local week_start   = props.start_sunday and 1 or 2
+    local last_day     = os.date("*t", os.time{year=date.year, month=date.month+1, day=0})
+    local month_days   = last_day.day
+    local column_fday  = (last_day.wday - month_days + 1 - week_start ) % 7
+
+    local num_columns  = props.week_numbers and 8 or 7
+    local start_column = num_columns - 6
+    local num_rows     = last_day.day // 7
+    if last_day.day > 28 then
+        num_rows = num_rows + 1
+        if column_fday >= 6 then
+            num_rows = num_rows + 1
+        end
+    elseif column_fday > 1 then
+        num_rows = num_rows + 1
+    end
 
     -- Create grid layout
     local layout = grid()
@@ -150,13 +165,6 @@ local function create_month(props, date)
     layout:set_spacing(props.spacing)
     layout:set_forced_num_rows(num_rows)
     layout:set_forced_num_cols(num_columns)
-
-    local start_row    = 3
-    local start_column = num_columns - 6
-    local week_start   = props.start_sunday and 1 or 2
-    local last_day     = os.date("*t", os.time{year=date.year, month=date.month+1, day=0})
-    local month_days   = last_day.day
-    local column_fday  = (last_day.wday - month_days + 1 - week_start ) % 7
 
     --local flags = {"header", "weekdays", "weeknumber", "normal", "focus"}
     local cell_date, t, i, j, w, flag, text


mattplm:
Sorry for the multiple comments. Here is a before view of the year 2021 : 

2021_before

And here is the result after applying this PR : 

2021_after

lib/wibox/widget/calendar.lua
```diff
@@ -139,8 +139,23 @@ end
...
-    local num_columns = props.week_numbers and 8 or 7
+    local start_row    = 3
+    local week_start   = props.start_sunday and 1 or 2
+    local last_day     = os.date("*t", os.time{year=date.year, month=date.month+1, day=0})
+    local month_days   = last_day.day
+    local column_fday  = (last_day.wday - month_days + 1 - week_start ) % 7
+
+    local num_columns  = props.week_numbers and 8 or 7
+    local start_column = num_columns - 6
+    local num_rows     = last_day.day // 7
```
sclu1034, line 150: Integers and the integer division operator have only been introduced in Lua 5.3. Awesome supports Lua versions as old as 5.1, where this will result in a syntax error.

lib/wibox/widget/calendar.lua
```diff
@@ -139,8 +139,23 @@ end
...
+    if last_day.day > 28 then
+        num_rows = num_rows + 1
+        if column_fday >= 6 then
+            num_rows = num_rows + 1
+        end
+    elseif column_fday > 1 then
+        num_rows = num_rows + 1
+    end
```
sclu1034, line 158: The math is neat, but requires quite a few magic numbers and a while to get into if you rarely work with calendar calculations.

A short comment explaining the algorithm would make this a lot nicer to skim.

sclu1034 (changes requested):
In the _after_ example here, it looks like the cell height for February is almost double the one for May or August.
Not sure how the others feel about that, but personally, I'd rather have the row and column size fixed, and the container size vary based on the number of rows/columns.

lib/wibox/widget/calendar.lua
mattplm: Oh yeah, didn't think about that, I'll fix it.

lib/wibox/widget/calendar.lua
mattplm: True, I'm adding comments.

mattplm: Turns out the math wasn't that neat ! Hopefully that's easier to understand what's happening now.

mattplm (commented):
About the variable cell height I'll let you guys discuss that while I take a look at how to test different setups with the grid layout

Edit : As I understand it, without variable cell height, we would still have the empty line issue on the year widget. Am I correct here ? 

Individual calendars have consistent cell height

2021_indiv

lib/wibox/widget/calendar.lua
```diff
@@ -139,8 +139,36 @@ end
...
+
+        -- On month with 30+ days add 1 week if:
+        -- - if 30 days and the first day is a sunday
+        -- - if 31 days and the first days is at least saturday
+        if column_fday >= 6 then
+            if last_day.day == 30 and column_fday == 7 or last_day.day == 31 then
+                num_rows = num_rows + 1
+            end
+        end
+    -- If the first day of february is anything but a monday
```
actionless, line 167: does it scale for the case if in system locale week starts with sunday instead of monday?

mattplm: It appears to ignore the locale settings. Please correct me if I'm wrong. Maybe the comment is not the most accurate here.

sclu1034:
> Individual calendars have consistent cell height

Great, those were my main concern regarding cell size. And I think that was the initial trigger for the linked issue as well.

For the year view, I think we should also favor fixed cell size and keep the empty lines. Searching for random images of "year calendars", that seems to be common practice.
Technically, it even visualizes the additional information of "there are fewer weeks in this month than the others".

lib/wibox/widget/calendar.lua
sclu1034: `os.date` is not locale aware, as far as I can tell.
<https://www.lua.org/pil/22.1.html> sounds like it is fixed, and a quick test confirms that:
```shell
LC_TIME=de_DE.UTF-8 lua5.3 -e 'print(os.date("*t").wday)'
4 # would be 3, if locale aware
```

 So `wday == 1` is always Sunday.

actionless:
i think `flex_height` or so should be a new option

lib/wibox/widget/calendar.lua
```diff
@@ -139,25 +148,47 @@ end
...
+    -- If the first day of february is anything but the first day of the week
+    elseif column_fday > 1 then
+        -- Span over 5 weeks
+        num_rows = num_rows + 1
+    end
 
     -- Create grid layout
     local layout = grid()
-    layout:set_expand(true)
-    layout:set_expand(true)
```
mattplm, line 148: This I'm not sure about, was there some kind of trick I don't get here?

mattplm (commented):
> i think `flex_height` or so should be a new bool option

I've given that a shot, sorry for the delay.

lib/wibox/widget/calendar.lua
```diff
@@ -331,6 +362,7 @@ local function get_calendar(type, date, font)
     ret._private.week_numbers  = beautiful.calendar_week_numbers or false
     ret._private.start_sunday  = beautiful.calendar_start_sunday or false
     ret._private.long_weekdays = beautiful.calendar_long_weekdays or false
+    ret._private.flex_height   = beautiful.calendar_flex_height or true
```
sclu1034, line 365: Shouldn't this be?

```suggestion
    ret._private.flex_height   = beautiful.calendar_flex_height or false
```

It's documented as `[opt=false]`.

sclu1034 (changes requested):

lib/wibox/widget/calendar.lua
mattplm: Hum, that's embarrassing. What I meant was: 

```suggestion
    flex_height                = beautiful.calendar_flex_height
    ret._private.flex_height   = if flex_height ~= nil then flex_height else true
```
(And change the doc line too)

If we set that to false, we change the default behavior of the calendar. Maybe this is what we want? In that case we can default to false.

sclu1034: It's true that, technically, the cells always had a "flex" height and that this PR didn't change that. But from a user's POV, that _is_ what would be changing currently.

If you look at [your second comment](https://github.com/awesomeWM/awesome/pull/3387#issuecomment-874917447), the change in code might be "remove empty lines", but the change in behaviour is "stretch cells for short months". And as I already mentioned, adding empty lines in a year view seems to be the norm.

Additionally, the initial bug about extraneous lines you referenced is only about the month view, where there really is no need for those empty lines. Because of that, I would consider a default of `true` a change in behaviour outside the scope of the issue that's being fixed.

mattplm: Good points, then false it is

sclu1034 (changes requested):
Travis is having issues with the logs again, but this is what the build is failing on:

```
Checking lib/wibox/widget/calendar.lua            1 warning

    lib/wibox/widget/calendar.lua:40:121: line is too long (124 > 120)
```

actionless:
https://app.travis-ci.com/github/awesomeWM/awesome/jobs/527248843#L11713

lib/wibox/widget/calendar.lua
```diff
@@ -116,6 +121,11 @@ local properties = { "date", "font", "spacing", "week_numbers", "start_sunday",
 -- @param function Function to embed the widget depending on its flag
 -- @property fn_embed
 
+--- Allow cells to have flexible height
+--
+--@DOC_wibox_widget_calendar_flex_height_EXAMPLE@
```
Aire-One, line 126: Does this example file even exist? Did you forget to commit it?

Aire-One (commented):
Nice work!

I had a few questions. Can you address them?

lib/wibox/widget/calendar.lua
psychon: Your guess is as good as mine. I guess it must be an error.

lib/wibox/widget/calendar.lua
```diff
@@ -139,25 +149,47 @@ end
 -- @tparam number|nil date.day Date day
 -- @treturn widget Grid layout
 local function create_month(props, date)
-    local num_rows    = 8
-    local num_columns = props.week_numbers and 8 or 7
+    local start_row    = 3
+    local week_start   = props.start_sunday and 1 or 2
+    local last_day     = os.date("*t", os.time{year=date.year, month=date.month+1, day=0})
+    local month_days   = last_day.day
+    local column_fday  = (last_day.wday - month_days + 1 - week_start ) % 7
```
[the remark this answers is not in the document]
mattplm: You're correct this looks wrong. I'm sure I didn't have issues when displaying yearly calendars which is weird, I must have missed something. I'm going to take a look at it.

Good point for the naming.

actionless:
and now please squash the changes to make test_prev_commits CI pass

actionless (dismissed):

lib/wibox/widget/calendar.lua
mattplm: This file has been added.

lib/wibox/widget/calendar.lua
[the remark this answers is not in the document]
mattplm: Upon further inspection of the code (again sorry for the delay, vacations), this code was there before my commit and I'm not sure I'm confortable modifying it. The `column_fday` var seems to be used for calculation below and seems to be relative to the first day, not the index of the first day.

The part where I used it in my conditions was wrong though, I changed that to take this into consideration.

Elv13 (approved):
@actionless What do you think of the current version?

mattplm:
--Envoyé à partir d’un Smartphone Android avec GMX Mail.

actionless (approved):
thanks a lot for this improvement!

PR #3389, opened 2021-07-07
fix: persistance for client.ontop across restarts (via X properties)

ViStefan:
I've noticed that `client.ontop` state is not persistent across awesome restarts. (It also isn't persistent after going fullscreen, but x layers somehow requires that, according to sourcecode.)

Ater a minute of research, I've found that X11 WM protocol doesn't have an atom for `ontop` state at all by convention. There are only layers like `DOCK`, `ABOVE`, etc.

I've added `_NET_WM_STATE_ONTOP` atom. This fixed that for me. Is it legal?

### DIFF
--- a/spawn.c
+++ b/spawn.c
@@ -412,9 +412,21 @@ spawn_child_exited(pid_t pid, int status)
  *
  * @tparam string|table cmd The command to launch.
  * @tparam[opt=true] boolean use_sn Use startup-notification?
- * @tparam[opt=false] boolean stdin Return a fd for stdin?
- * @tparam[opt=false] boolean stdout Return a fd for stdout?
- * @tparam[opt=false] boolean stderr Return a fd for stderr?
+ * @tparam[opt="DEV_NULL"] boolean|string stdin Pass `true` to return a fd for
+ *   stdin. Use `"DEV_NULL"` to redirect to /dev/null, or `"INHERIT"` to inherit
+ *   the parent's stdin. Implementation note: Pre-2.74 glib doesn't support
+ *   *explicit* `DEV_NULL`. When `DEV_NULL` is passed on glib <2.74, Awesome will
+ *   use glib's default behaviour.
+ * @tparam[opt="INHERIT"] boolean|string stdout Pass `true` to return a fd for
+ *   stdout. Use `"DEV_NULL"` to redirect to /dev/null, or `"INHERIT"` to
+ *   inherit the parent's stdout. Implementation note: Pre-2.74 glib doesn't
+ *   support *explicit* `INHERIT`. When `INHERIT` is passed on glib <2.74,
+ *   Awesome will use glib's default behaviour.
+ * @tparam[opt="INHERIT"] boolean|string stderr Pass `true` to return a fd for
+ *   stderr. Use `"DEV_NULL"` to redirect to /dev/null, or `"INHERIT"` to
+ *   inherit the parent's stderr. Implementation note: Pre-2.74 glib doesn't
+ *   support *explicit* `INHERIT`. When `INHERIT` is passed on glib <2.74,
+ *   Awesome will use glib's default behaviour.
  * @tparam[opt=nil] function exit_callback Function to call on process exit. The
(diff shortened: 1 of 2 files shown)


ViStefan:
It is not too much about real merging, but I just want to discuss, is it a good idea at all, ot it breaks conventions? If it's all right, I'll cover it with tests.

Anybody?

Aire-One:
Sorry for the delay. We are not very active latterly, and it's summer now, so I guess everyone is busy at this time...

AFAIK, we try to not add new atoms that are not standard and that would be only interpreted by us, in awesome. I'm not very familiar with this subject, so maybe there is a world where this ONTOP atom is used by other, and we should also support it?

@psychon, @Elv13 what's your point of view of the situation?

> I've noticed that client.ontop state is not persistent across awesome restarts.

About persistence across awesome restarts, there is not too much we can do. The recommended way to persist data would be to write them somewhere on disk and read this file at awesome start, in the end user `rc.lua` to apply back the properties to your clients. I don't remember having seen any real implementation of this before, tho...

> It also isn't persistent after going fullscreen, but x layers somehow requires that, according to sourcecode.

I don't know about the X.org requirement, but there is probably something we could do at our level, in awesome, to persist the data when a client goes fullscreen. (I'm actually surprised we lost the `ontop` property value when `fullscreen` is altered...)

ViStefan:
> Im actually surprised we lost the `ontop` property value when `fullscreen` is altered
There's no wodner. If window has both `ontop` and `fullscreen`, other windows will be unusable, because whole screen is obtained by `ontop` window. :)

https://github.com/awesomeWM/awesome/blob/906dc543e423bac439942e4c755de1e452b9e1ff/objects/client.c#L2556

Aire-One:
> > Im actually surprised we lost the `ontop` property value when `fullscreen` is altered
> 
> There's no wodner. If window has both `ontop` and `fullscreen`, other windows will be unusable, because whole screen is obtained by `ontop` window. :)

It makes a lot of sense. Thanks for this!

> Probably it would be better to restore `ontop` status after leaving fullscreen, but it is possible to be made with `rc.lua`.

I don't know how we could implement this feature correctly. Also, I'm afraid it is the kind of breaking changes we don't want to introduce in awesome...

Hm... Yeah, I think it is doable from the end user config file. It should be easy enough to create a list of ontop clients and check it when a client exists fullscreen. (Even easier: create a custom property directly attached to the client object itself)

psychon:
Well... there is no `_NET_WM_STATE_ONTOP` in https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html, so... yeah. However, I don't see any reason why this couldn't be persisted in a custom property. We already do that with `floating` in `lib/awful/client.lua`:
```lua
-- Register persistent properties
client.property.persist("floating", "boolean")
```
However, I guess `ontop` cannot be as easily persisted from Lua since the property itself is implemented in C. Quick untested attempt that doesn't touch any C code:
```lua
local xprop = "awful.client.property.ontop"
awesome.register_xproperty(xprop, "boolean")
client.connect_signal("manage", function(c)
    local value = c:get_xproperty(xprop)
    if value ~= nil then
        c.ontop = value
    end
end)
client.connect_signal("property::ontop", function(c)
    c:set_xproperty(xprop, c.ontop)
end)
```
The above is completely untested, but dumping that into `lib/awful/client.lua` could be easier than implementing the same thing in C.

ewmh.c
```diff
@@ -67,6 +67,8 @@ ewmh_client_update_hints(lua_State *L)
         state[i++] = _NET_WM_STATE_HIDDEN;
     if(c->urgent)
         state[i++] = _NET_WM_STATE_DEMANDS_ATTENTION;
+    if(c->ontop)
+        state[i++] = _NET_WM_STATE_ONTOP;
```
psychon, line 71: This is a potential buffer overflow. If all the `if`s are taken (which should never happen since some states are mutually exclusive), this might write beyond the end of the array.

ViStefan:
> The above is completely untested, but dumping that into `lib/awful/client.lua` could be easier than implementing the same thing in C.

I'll try this and update request.

PR #3397, opened 2021-07-24
Added disable_task_name_focus

rxt30:
Added the option to disable the task name of the focus client.
This option can also be used to show the task name of the focused
client, if the task name of the other clients is disabled.

If `theme.tasklist_disable_task_name_focus` or the argument `disable_task_name_focus` are set to false, it will always show the title of the focused client.      
If `theme.tasklist_disable_task_name_focus` or the argument `disable_task_name_focus` are set to true, it will always hide  the title of the focused client.      
If the options isn't set, it will refer to the value given by `theme.tasklist_disable_task_name` or the argument `disable_task_name`, to maintain backwards compatibility.

Note: The assignment of `disable_task_name_focus` is pretty ugly, but i couldn't find a better way with maintaining backwards compatibility and keeping the naming consistent. I would like to hear a better solution and general improvements, if needed before the merge.

### DIFF
--- a/lib/awful/widget/tasklist.lua
+++ b/lib/awful/widget/tasklist.lua
@@ -167,6 +167,10 @@ local instances
 -- @beautiful beautiful.tasklist_disable_task_name
 -- @tparam[opt=false] boolean tasklist_disable_task_name
 
+--- Disable the tasklist client titles for the focused client.
+-- @beautiful beautiful.tasklist_disable_task_name_focus
+-- @tparam[opt=false] boolean tasklist_disable_task_name_focus
+
 --- Disable the extra tasklist client property notification icons.
 --
 -- See the <a href="#status_icons">Status icons</a> section for more details.
@@ -332,6 +336,7 @@ local function tasklist_label(c, args, tb)
     local bg_image_minimize = args.bg_image_minimize or theme.tasklist_bg_image_minimize or theme.bg_image_minimize
     local tasklist_disable_icon = args.tasklist_disable_icon or theme.tasklist_disable_icon or false
     local disable_task_name = args.disable_task_name or theme.tasklist_disable_task_name or false
+    local disable_task_name_focus = args.disable_task_name_focus or theme.tasklist_disable_task_name_focus or (disable_task_name and args.disable_task_name_focus == nil and theme.tasklist_disable_task_name_focus == nil)
     local font = args.font or theme.tasklist_font or theme.font or ""
     local font_focus = args.font_focus or theme.tasklist_font_focus or theme.font_focus or font or ""
     local font_minimized = args.font_minimized or theme.tasklist_font_minimized or theme.font_minimized or font or ""
@@ -396,9 +401,15 @@ local function tasklist_label(c, args, tb)
 
     if focused then
         bg = bg_focus
-        text = text .. "<span color='"..fg_focus.."'>"..name.."</span>"
         bg_image = bg_image_focus
         font = font_focus
+
+        if not disable_task_name_focus and disable_task_name then
+            name = name .. (gstring.xml_escape(c.name) or gstring.xml_escape("<untitled>"))
+        elseif disable_task_name_focus then
+            name = ""
+        end
+        text = text .. "<span color='"..fg_focus.."'>"..name.."</span>"
 
         if args.shape_focus or theme.tasklist_shape_focus then
             shape = args.shape_focus or theme.tasklist_shape_focus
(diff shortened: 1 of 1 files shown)


lib/awful/widget/tasklist.lua
```diff
@@ -332,6 +336,7 @@ local function tasklist_label(c, args, tb)
     local bg_image_minimize = args.bg_image_minimize or theme.tasklist_bg_image_minimize or theme.bg_image_minimize
     local tasklist_disable_icon = args.tasklist_disable_icon or theme.tasklist_disable_icon or false
     local disable_task_name = args.disable_task_name or theme.tasklist_disable_task_name or false
+    local disable_task_name_focus = args.disable_task_name_focus or theme.tasklist_disable_task_name_focus or (disable_task_name and args.disable_task_name_focus == nil and theme.tasklist_disable_task_name_focus == nil)
```
Aire-One, line 339: It means, if the user sets `disable_task_name = true` and says nothing for `disable_task_name_focus`, it will be sets to `true` as a default?

Shouldn't it be :

```suggestion
    local disable_task_name_focus = args.disable_task_name_focus or theme.tasklist_disable_task_name_focus or false
```

Why do we need to test `disable_task_name` ?

Aire-One (changes requested):
I don't know about this one... It sounds like something that can be done by the end user with the widget_template or the filter.

Even the old `disable_task_name` sounds like a bad idea to me considering the widget_template feature... However, given we have it anyway, and we'll not remove it to keep backward compatibility, I don't see why this one couldn't make its way.

lib/awful/widget/tasklist.lua
rxt30: My intention was indeed to have i depend on `disable_task_name`.

If `disable_task_name_focus` has `false` as default value, user with `disable_task_name` set to `true` would suddenly see the task name of the focused client, as `disable_task_name_focus` is still set to `false`. 

If this case is acceptable, i. e. the user must explicitly disable the task name of the focused client, i am happy to commit your suggestion, as it would tidy up the code.

Aire-One: Sorry, I don't understand the logic...

If we want to preserve the current behavior, and given the changes you did in the test line 407, we should just use `false` as a default here.

Am I missing something? Why do we need to test `disable_task_name` here?

> If disable_task_name_focus has false as default value, user with disable_task_name set to true would suddenly see the task name of the focused client, as disable_task_name_focus is still set to false.

Isn't this the current behavior? (and the one we need to keep?)

rxt30: No problem :)

In the current behaviour, `disable_task_name` set to `true` disables the task name for all clients, regardless of their status (`focused`,`minimized`...)

I implemented the test against `disable_task_name`, that for existing configuration without the newly introduced option, `disable_task_name_focus` would be set to the value of `disable_task_name`.     
With this, the task name of the focused client will NOT been show after the changes by me have been applied.


If, as suggested by your changes, `disable_task_name_focus` would be `false` by default without testing against `disable_task_name`, users with the option `disable_task_name` set to `true` would suddenly see the task-name of the focused client after the changes have been applied.

tl;dr:
`disable_task_name_focus` tests against `disable_task_name`:
`disable_task_name` set to `true` and `disable_task_name_focus` not set => Task name of focused client not shown

`disable_task_name_focus` defaults `false` and doesn't test against `disable_task_name`:
`disable_task_name` set to `true` and `disable_task_name_focus` not set => Task name of focused client will be shown

Hope that the explanation is better understandable now :)

Aire-One: Hm... ok! I think those `== nil` got me a little confused here. Thanks for the detailed explanation! :sunglasses:

rxt30: Happy to hear that :)

sclu1034: Well, "someone got confused while reading my code" isn't something that one should be happy to hear, especially with regard to maintainability of that code.
Would you be able to summarize that explanation in a code comment?

rxt30: Wrote a short comment, didn't got much into detail, as this would produce clutter in this part of the code.

Also updated the documentation, as the default value for `disable_task_name_focus` is the value of `disable_task_name` instead of `false`.

lib/awful/widget/tasklist.lua
```diff
@@ -396,10 +402,16 @@ local function tasklist_label(c, args, tb)
 
     if focused then
         bg = bg_focus
-        text = text .. "<span color='"..fg_focus.."'>"..name.."</span>"
         bg_image = bg_image_focus
         font = font_focus
 
+        if not disable_task_name_focus and disable_task_name then
+            name = name .. (gstring.xml_escape(c.name) or gstring.xml_escape("<untitled>"))
```
actionless, line 409: what's your intention behind adding this line of code?

actionless:
but more generally i think i'm agree with @Aire-One - i don't think it plays along with @Elv13 's direction of adding widget templates everywhere over having hardcoded options

rxt30:
If that's the case, my pull request can be closed.

If the general direction is towards templates, it just would add clutter to implement the feature suggested by my pull request.

But thank you for taking the time to review my pull request and maintaining the code :)

actionless:
mb it would be useful instead adding example into existing documentation on how to do it using templates

rxt30:
will look into it, may open another pr if i get it working with a template

PR #3399, opened 2021-07-31
Various doc improvements nr. 4

sclu1034:
There's a small code refactor in there, but it merely changes a few variable names to make them more readable.

### DIFF
--- a/docs/common/cursor_c.ldoc
+++ b/docs/common/cursor_c.ldoc
@@ -1,80 +1,80 @@
- * <table class='widget_list' border=1>
- * <tr><td>num_glyphs          </td></tr>
- * <tr><td>cursor              </td></tr>
- * <tr><td>arrow               </td></tr>
- * <tr><td>based_arrow_down    </td></tr>
- * <tr><td>based_arrow_up      </td></tr>
- * <tr><td>boat                </td></tr>
- * <tr><td>bogosity            </td></tr>
- * <tr><td>bottom_left_corner  </td></tr>
- * <tr><td>bottom_right_corner </td></tr>
- * <tr><td>bottom_side         </td></tr>
- * <tr><td>bottom_tee          </td></tr>
- * <tr><td>box_spiral          </td></tr>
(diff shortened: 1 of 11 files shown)


Aire-One (approved):
This is very good work, thanks a lot for the effort! :heart: 

Only comment I have is about 05ca439ed5dfa4b2e7e41866d495c00954f7377b: If we have to replace this table with something, I would have preferred to have a real HTML list here. What are the beneficts from a flex div in this situation?

Aire-One (changes requested):
Uh, well, given you even kept the `*` to create bullets, I'm moving my concern to an actual change request. I think we better go with an HTML list here.

sclu1034:
I did not keep `*` to create markdown bullet points. Those are required for the file to be properly imported into a C comment. If this was imported into Lua, they would be replaced by `--`.
The CMake "magic" that copies these files into the actual code comments doesn't prefix the lines individually. So if I were to make that file plain HTML, it would break the doc comment.

The benefit of the `<div>` structure is better screen space utilization.

On master, 1440p is not enough to show even half the options, 1080p requires even more scrolling.
90% is wasted white space, and a list wouldn't be much better.

Also notice the HTML list dot at the start, created by the fact that CMake's substitution does add the comment prefix for the first line (or rather, keeps one that's already there), but subsequent ones don't create additional `<li>`s.
image

Now, everything is visible in half a screen. And `flex-wrap` and a `width` in `em` makes it adjust properly, down to mobile.
image

Aire-One:
> I did not keep the `*` to create markdown bullet points. Those are required for the file to be properly imported into a C comment. 

Oh yes! I should have get it from the first look... My bad. 

Thanks for the additional details and screenshots comparison about the use of flexbox over a list on this situation. Now I understand your point. I think I don't really like this grid-like list style, but I can't think of a better alternative... 

Good job 👍

Aire-One (approved):

Elv13 (approved):

PR #3400, opened 2021-07-31
06 appearance.md duplicates fix (fixes #3328)

### LINKED ISSUE #3328: Some variables on `06-appearance.md` page are mentioned multiple times.
[Link to the page](https://awesomewm.org/apidoc/documentation/06-appearance.md.html)

`theme.bg_normal` and `theme.fg_normal` appear 5 times each; `theme.bg_systray`, `theme.border_width`, `theme.font` and `theme.useless_gap` appear 2 times each.

Command used to check duplicates:
```
$ cat theme.lua
local theme = {}

-- Default variables
-- theme.bg_normal = nil
[...]
-- theme.wibar_shape = nil

return theme

$ sort theme.lua | uniq -cd
     37 
      5 -- theme.bg_normal = nil
      2 -- theme.bg_systray = nil
      2 -- theme.border_width = nil
      5 -- theme.fg_normal = nil
      2 -- theme.font = nil
      2 -- theme.useless_gap = nil
```

aflorea-2k:
The table entries that were duplicated are now above each other, in the same cell; all the links are still accessible. Formatting could be better.

There are no longer duplicates in the sample file.

### DIFF
--- a/docs/_parser.lua
+++ b/docs/_parser.lua
@@ -165,17 +165,22 @@ local function parse_files(paths, property_name, matcher, name_matcher)
                         )
                     else
                         local insert_name = name:gsub("_", "_")
+                        local link = get_link(file, var, var:match(exp3):gsub("_", "\\_"))
+                        local desc = buffer:gmatch("[-*/ \n]+([^\n.]*)")() or ""
+                        local mod = path_to_module(file)
+
                         if names[insert_name] == nil then
                             count = count + 1
                             table.insert(ret, count, {
                                 file = file,
                                 name = insert_name,
-                                link = get_link(file, var, var:match(exp3):gsub("_", "\\_")),
-                                desc = buffer:gmatch("[-*/ \n]+([^\n.]*)")() or "",
-                                mod  = path_to_module(file)
+                                link = link,
+                                desc = desc,
+                                mod  = mod
                             })
                             names[insert_name] = count
                         else
+                            link = link .. "(" .. mod .. ")"
                             if type(ret[names[insert_name]].link) ~= "table" then
                                 ret[names[insert_name]].file = {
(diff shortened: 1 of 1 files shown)


docs/_parser.lua
```diff
@@ -190,7 +225,21 @@ local function create_table(entries, columns, prefix)
...
+                local firstline = true
+                for _,v in pairs(entry[column]) do
+                    if not firstline then
+                        line = line .. "<br>"
+                    else
+                        firstline = false
+                    end
+                    line = line .. v
+                end
+                line = line .. "</td>"
```
Aire-One, line 239: I think it's better to do that with list. This is basically a list that we want to display here, so it feels better to use the right HTML tag for this content, and it makes the code easier to read:

```suggestion
            if type(entry[column]) == "table" then
                line = line .. "<td><ul>"
                for _,v in pairs(entry[column]) do
                    line = line .. "<li>" .. v .. "</li>"
                end
                line = line .. "</ul></td>"
```

If we really need to remove the list's bullets, we can easily do that by CSS (I think we already have a class for that (?)).

Aire-One (changes requested):
Hey, sorry I couldn't give my help on your question at #3328, but as I can see @sclu1034 gave good hints!

You did a good job with this, so far!

I have a few comments, can you address them, so we can merge your work :sunglasses:

docs/_parser.lua
aflorea-2k: It's much cleaner, thanks!

Aire-One (approved):
Thank you for this!

I hope others maintainers will be able to come quickly and also give their approval :wink:

sclu1034 (approved):

Elv13 (approved):

PR #3409, opened 2021-08-12
Remove Vicious as async counterexample

McSinyx:
Vicious has stopped using io.popen since October 2019.

### DIFF
--- a/docs/90-FAQ.md
+++ b/docs/90-FAQ.md
@@ -36,8 +36,7 @@ Do **not** use such functions and prefer `awful.spawn.easy_async`,
 fast enough and won't impact the main event loop iteration time, you are wrong.
 *Every* calls to `io.open` are impacted by the system `iowait` queue and can
 spend hundreds of milliseconds blocked *before* being executed. Note that
-some common widget or probe libraries such as
-[Vicious](https://github.com/Mic92/vicious) do not follow this
+some common widget or probe libraries do not follow this
 advice currently and are known to cause input lag on some systems (but not all).
 
 In both case, a warning like:


actionless (approved):
lol, finally

actionless:
still using globals though?

McSinyx:
I think we localized almost everything in the `io.popen` purge, and we have this in our [contributing guidelines](
https://vicious.readthedocs.io/en/latest/contributing.html#performance-tips) now :smile:

> **Never** use global variables.

actionless:
> almost everything

ok, please post an update here when you done. thanks for working on improving the code quality.

regarding globals, i meant, if you're still using instead of return values of callback functions?

McSinyx:
> actually, i've just double-checked on master branch of vicious and i still see 16 usages of io.popen. could you clarify that, please?

I've just check and these are the current usages:

* Those in `contrib` are not supported, they are not officially supported.  We tried working on them but many don't run or we didn't know how to setup.
* `spawn.lua` use popen as a fallback when awesome isn't present.
* `helpers.lua` calls popen exactly once for uname (it's cached) as awesome rc is loaded.

> regarding globals, i meant, if you're still using instead of return values of callback functions?

All officially supported `widgets/` use callback now.

actionless (changes requested):
nice move instead of fixing or removing them to tell what they're just unsupported, but the thing is what upset users will come here with their complains about why "awesome is slow"

McSinyx:
To be precise, the [contrib directory](https://drewdevault.com/2020/06/06/Add-a-contrib-directory.html) has never been supported.  If you feel that it's still a risk for users' angry letters, feel free to close this PR.  I only noticed this part of the documentation while browsing the docs for my own use and thought it would be a good idea to communicate the situation of Vicious, after checking Awesome issue tracker for recent related tickets and not finding any.  I do not wish to waste anybody's time discussing back and forth on a few words in the docs that does no harm to neither project if they stay as-is.

Elv13 (approved):

Elv13:
@actionless What do you think about merging this? I feel like they resolved the main reason why we got all of these "AwesomeWM is slow" over the year. Asking for removing this line from the doc seems more than fair.

actionless (approved):
ok, let's see, if anyone users of that io.popen(curl ...) widget would come back for us later ;)

Elv13:
> ok, let's see, if any of users of that io.popen(curl ...) widget would come back for us later ;)

Oh, they will ;). Most configs have a copy pasted version of vicious in their trees. They never get updated and people will still fork these configs. Mentioning vicious in the official doc "forever" just doesn't help them debug all that much.

McSinyx:
Awesome, thanks!

PR #3421, opened 2021-08-28
[RFC/WIP] Implement wibox.widget.template: An abstract widget that handles a preset of concrete widget.

Aire-One:
Hello :wave: 

I think it'll need more work to be ready, I hope to get feedbacks to improve this new widget.

This PR is (what I think to be) the response we need to provide a standard API that address issues raised by initiatives like #2955 and #3208.

The idea behind this new "template widget" is to standardize how we want/try to implement widget template across the `awful.widget` library (and third-party libraries that want to implement user customizable widget). Thanks to this widget, I hope we can build more extendable widgets that allow end-users to easily implement their own customization.

In the scope of this PR, I only want to implement the `wibox.widget.template` widget and the tools it needs. If you validate this approach, more PRs will come to implement usage of the template widget across widgets from `awful.widget`. 

For the record, here is a naive attempt at implementing the widget template for the layoutbox widget : https://github.com/Aire-One/awesome/tree/feature/widget_template/layoutbox

Short list for todos :

- [ ] Documentation
- [ ] Usage examples
- [ ] Rename `widget_template` parameter (?)
- [ ] Intermediate constructor (to prevent `w = layoutbox { widget_template = wibox.widget.template{...} }`)
- [ ] Save arguments from dismissed update call (?)

### DIFF
--- a/lib/wibox/widget/template.lua
+++ b/lib/wibox/widget/template.lua
@@ -1,0 +1,90 @@
+---------------------------------------------------------------------------
+-- An abstract widget that handles a preset of concrete widget.
+--
+-- The `wibox.widget.template` widget is an abstraction layer that contains a
+-- concrete widget definition. The template widget can be used to build widgets
+-- that the user can customize at their will, thanks to the template mechanism.
+--
+-- @author Aire-One
+-- @copyright 2021 Aire-One <aireone@aireone.xyz>
+--
+-- @classmod wibox.widget.template
+-- @supermodule wibox.widget.base
+---------------------------------------------------------------------------
+
+local wbase = require("wibox.widget.base")
+local gtable = require("gears.table")
+local gtimer = require("gears.timer")
+
+local default_update_callback = function()
+    return nil
+end
+
+local default_template_widget = wbase.empty_widget()
+
+local template = {
+    mt = {},
+    queued_updates = {}
+}
+
+function template:_do_update_now(args)
+    self:update_callback(args)
+    template.queued_updates[self] = false
+end
+
+--- Update the widget.
+-- This function will call the `update_callback` function at the end of the
+-- current GLib event loop. Updates are batched by event loop, it means that the
+-- widget can only be update once by event loop. If the `template:update` method
+-- is called multiple times during the same GLib event loop, only the first call
+-- will be run.
+-- All arguments are passed to the queued `update_callback` call.
+-- @treturn boolean Returns `true` if the update_callback have been queued to be
+--   run at the end of the event loop. Returns `false` if there is already an update
+--   in the queue (in this case, this new update request is discared).
+function template:update(args)
+    local update_args = gtable.crush(gtable.clone(self.update_args, false), args or {})
+
+    if not template.queued_updates[self] then
+        gtimer.delayed_call(
+            function()
+                self:_do_update_now(update_args)
+            end
+        )
+        template.queued_updates[self] = true
+
+        return true
+    end
+
+    return false
+end
+
+--- Create a new `wibox.widget.template` instance.
+-- @tparam[opt] table args
+-- @tparam[opt] table|widget|function args.widget_template The template of widget to use.
+-- @tparam[opt] function args.update_callback The callback function to update the widget.
+-- @treturn wibox.widget.template The new instance.
+function template.new(args)
+    args = args or {}
+
+    local widget_template =
+        type(args.widget_template) == "function" and args.widget_template() or args.widget_template or
+        default_template_widget
+
+    local widget = wbase.make_widget_from_value(widget_template)
+
+    widget.update_callback = args.update_callback or default_update_callback
+    widget.update_args = args.update_args or {}
+
+    gtable.crush(widget, template, true)
+
+    return widget
+end
+
+function template.mt:__call(...) --luacheck: ignore unused variable self
+    return template.new(...)
+end
+
+return setmetatable(template, template.mt)
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/spec/wibox/widget/template_spec.lua
+++ b/spec/wibox/widget/template_spec.lua
@@ -1,0 +1,71 @@
+---------------------------------------------------------------------------
+-- @author Aire-One
+-- @copyright 2021 Aire-One
+---------------------------------------------------------------------------
+
+_G.awesome.connect_signal = function() end
+
+local template = require("wibox.widget.template")
+local gtable = require("gears.table")
+local gtimer = require("gears.timer")
+
+local function is_same_table_struture(state, arguments) -- luacheck: ignore unused argument state
+    return function(value)
+        return table.concat(gtable.keys(arguments[1])) == table.concat(gtable.keys(value))
+    end
+end
+
+assert:register(
+    "matcher",
+    "is_same_table_struture",
+    is_same_table_struture
+)
+
+describe("wibox.widget.template", function()
+    local widget
+
+    before_each(function()
+        widget = template()
+    end)
+
+    describe("widget:update()", function()
+        it("batch calls", function()
+            local spied_update_callback = spy.new(function() end)
+
+            widget.update_callback = spied_update_callback
+
+            -- Multiple calls to update
+            widget:update()
+            widget:update()
+            widget:update()
+
+            -- update_callback shouldn't be called before the end of the event loop
+            assert.spy(spied_update_callback).was.called(0)
+
+            gtimer.run_delayed_calls_now()
+
+            -- updates are batched, so only 1 call should have been performed
+            assert.spy(spied_update_callback).was.called(1)
+        end)
+
+        it("update parameters", function()
+            local spied_update_callback = spy.new(function() end)
+            local args_structure = { foo = "string" }
+            local update_args = { foo = "bar" }
+
+            widget.update_args = args_structure
+            widget.update_callback = spied_update_callback
+
+            widget:update(update_args)
+
+            gtimer.run_delayed_calls_now()
+
+            assert.spy(spied_update_callback).was.called_with(
+                match.is_ref(widget),
+                match.is_same_table_struture(widget.update_args)
+            )
+        end)
+    end)
+end)
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/lib/wibox/widget/init.lua
+++ b/lib/wibox/widget/init.lua
@@ -21,6 +21,7 @@ local widget = {
     slider = require("wibox.widget.slider");
     calendar = require("wibox.widget.calendar");
     separator = require("wibox.widget.separator");
+    template = require("wibox.widget.template");
 }
 
 setmetatable(widget, {
--- a/spec/preload.lua
+++ b/spec/preload.lua
@@ -4,7 +4,10 @@
 require("lgi")
 
 -- Always show deprecated messages
-_G.awesome = {version = "v9999"}
+_G.awesome = {
+    version = "v9999",
+    api_level = 9999,
+}
 
 -- "fix" some intentional beautiful breakage done by .travis.yml
 require("beautiful").init{a_key="a_value"}


lib/wibox/widget/template.lua
```diff
@@ -0,0 +1,90 @@
...
+    self:update_callback(args)
+    template.queued_updates[self] = false
+end
+
+--- Update the widget.
+-- This function will call the `update_callback` function at the end of the
+-- current GLib event loop. Updates are batched by event loop, it means that the
+-- widget can only be update once by event loop. If the `template:update` method
+-- is called multiple times during the same GLib event loop, only the first call
+-- will be run.
```
sclu1034, line 40: I think it makes more sense to only run the _latest_ update request.

Use case:
It's very likely that this `:update()` method will be called from `property::` signal handlers. But those are executed synchronously.

So something like
```lua
mywidget.foo = "foo"
mywidget.bar = "bar"
```
would translate to
```lua
mywidget.foo = "foo"
  -- queue `:update()`
mywidget.bar = "bar"
  -- ignore

-- run `:update()` with `{ foo = "foo" }`
```
Whereas the desired outcome would be
```lua
mywidget.foo = "foo"
  -- queue `:update()`
mywidget.bar = "bar"
  -- discard previous, queue new `:update()`

-- run `:update()` with `{ foo = "foo", bar = "bar" }`
```

I don't think there is an easy way to discard a specific handler from delayed calls.
But another way to implement this would be using table references for `args` so that calling `:update()` always updates `<some_table>` and `:_do_update_now()` holds a reference to that same table.

lib/wibox/widget/template.lua
```diff
@@ -0,0 +1,90 @@
...
+-- @supermodule wibox.widget.base
+---------------------------------------------------------------------------
+
+local wbase = require("wibox.widget.base")
+local gtable = require("gears.table")
+local gtimer = require("gears.timer")
+
+local default_update_callback = function()
+    return nil
+end
```
sclu1034, line 21: It seems quite inefficient to execute the whole of `:update()` just to run this no-op if the user didn't provide a `update_callback`.
I think it would make more sense to allow `self.update_callback == nil` and add something like this to `:update()`:
```lua
if type(self.update_callback) ~= "function" then
	return
end
```

lib/wibox/widget/template.lua
```diff
@@ -0,0 +1,90 @@
...
+-- current GLib event loop. Updates are batched by event loop, it means that the
+-- widget can only be update once by event loop. If the `template:update` method
+-- is called multiple times during the same GLib event loop, only the first call
+-- will be run.
+-- All arguments are passed to the queued `update_callback` call.
+-- @treturn boolean Returns `true` if the update_callback have been queued to be
+--   run at the end of the event loop. Returns `false` if there is already an update
+--   in the queue (in this case, this new update request is discared).
+function template:update(args)
+    local update_args = gtable.crush(gtable.clone(self.update_args, false), args or {})
```
sclu1034, line 46: Why compute this every time and not just when `template.queued_updates[self] == false`, when it's actually used?

sclu1034 (changes requested):

Aire-One:
Sorry for the delay before working again on this PR.

Thank you, @sclu1034, for your review :smiley: I'm slowly getting back to this PR, I'll try to provide more updates in the coming days.

lib/wibox/widget/template.lua
Aire-One: I have updated the way the update callback arguments are managed. Now they are saved to an attribute to be easily passed to the user defined callback.

lib/wibox/widget/template.lua
```diff
@@ -0,0 +1,83 @@
...
+--- Update the widget.
+-- This function will call the `update_callback` function at the end of the
+-- current GLib event loop. Updates are batched by event loop, it means that the
+-- widget can only be update once by event loop. If the `template:update` method
+-- is called multiple times during the same GLib event loop, only the first call
+-- will be run.
+-- All arguments are passed to the queued `update_callback` call.
+function template:update(args)
+    if type(args) == "table" then
+        self.update_args = gtable.crush(gtable.clone(self.update_args or {}, false), args)
```
sclu1034, line 43: I think `self._private.update_args` would make more sense here, as this shouldn't be public API.

spec/wibox/widget/template_spec.lua
```diff
@@ -0,0 +1,86 @@
...
+    return function(value)
+        return table.concat(gtable.keys(arguments[1])) == table.concat(gtable.keys(value))
+    end
+end
+
+assert:register(
+    "matcher",
+    "is_same_table_struture",
+    is_same_table_struture
+)
```
sclu1034, line 22: That's now unused, as far as I can tell.

Aire-One: I forget I wrote this, ahah!

lib/wibox/widget/template.lua
```diff
@@ -5,10 +5,27 @@
...
 --
+-- Common usage examples
+-- =====================
+--
+-- A basic implementation of the template widget needs a widget definition and
+-- a callback to manage widget updates.
+--
+-- The `:update()` can be called later to update the widget template. Arguments
+-- can be provided to the `:update()` method, it will be forwarded to the
+-- `update_callback` function.
```
sclu1034, line 16: I think this should be moved to a separate section, e.g. "how to create a widget/an API that utilizes widget templates".

The common usage that most users will run into will only be about things like "what to write in `awful.taglist { widget_template = ... }`?". For those cases, the user will likely just get confused about where they are supposed to call `:update()` and with what arguments.

In the examples, the same distinction should be made clear, with comments like
```lua
-- this is what the user/consumer will write in their `rc.lua`
```
```lua
-- this is what a widget creator will use in their widget's code, e.g. `lib/awful/taglist.lua`
```

As it is right now, users will likely assume that _they_ are supposed to call `:update()` in their `rc.lua`.

tests/examples/wibox/widget/template/concrete_implementation.lua
```diff
@@ -0,0 +1,55 @@
...
+--DOC_NEWLINE
+    function concrete_widget_template_builder(args)
+        -- TODO : this should be made easier
+        local ret = args.widget_template
+                and args.widget_template.is_widget
+                and args.widget_template.widget_name == "wibox.widget.template"
+                and args.widget_template
+            or wibox.widget.template(
+                args.widget_template or default_widget
+            )
```
sclu1034, line 30: You could create a static function `wibox.widget.template.ensure_is_template(widget_template, default_widget)` that does this.

Aire-One: I have this other branch (unstable since I rebase it every time I update the template widget) where I implement the usage of this new widget in the scope of the layoutbox (https://github.com/Aire-One/awesome/tree/feature/widget_template/layoutbox).

I took this big "if test" from there. Having a static function that handle that is actually an improvement I thought of. (It was part of the idea behind the "proxy contractor" I thought of in my todo list from the first message.)

Another possible approach would be to make the widget constructor capable of determining you gave a working instance to return it directly, or clone it maybe (?) I'm not sure whether it would make sense, tho...

Another way I'd like to explore is to use the make_widget function with the template widget as a proxy (https://awesomewm.org/apidoc/classes/wibox.widget.base.html#wibox.widget.base.make_widget). I never used this function that way before, so I'm not sure how it would actually behave. I need to dig the source code on that...

lib/wibox/widget/template.lua
Aire-One: I agree this can be confusing since with this widget, you have 2 use case : as the widget author and as the end-user. I'll try to improve this documentation part.

Thank you for your input! :smile: 

(As a quick note : the taglist/tasklist widget are not targets for this widget template implementation. While it is possible to use it as a base for these list widgets, I think we should have a separate list_template widget to handle all the layouting and iterate-over-a-list stuff that these two need.)

Aire-One:
I had some issue to find the time I needed to finish working on this... I think it's now ready! :smile: 

All check are green, and I'm not sure how(/whether it's necessary?) to test the remaining lines highlighted by codecov.

Thank you, @sclu1034, for your previous reviews.

tests/examples/wibox/widget/template/concrete_implementation.lua
Aire-One: I went for the easy way : use a basic table with parameters to pass to the template, and if this table doesn't exist, fallback to the default template.

There is no easy way to enforce typing of parameters without overcomplicated checks that doesn't add much to the end user. (And the test I did there was flaky anyway, you just can't find the widget type based on the widget_name property, that's a bad idea.)

Elv13:
(just a note to say I am in vacations, I will review this when I get back. I really appreciate this work)

Elv13:
Rebased. Apology for forgetting about this for a year.

I also added a commit to bring it to feature parity with the `awful.widget.common` code (with the intention of deleting the said code in the next commit). See https://github.com/awesomeWM/awesome/blob/master/lib/awful/widget/common.lua#L107 and https://github.com/awesomeWM/awesome/blob/master/lib/awful/widget/common.lua#L53

One thing I would be inclined to do is move the file from `lib/wibox/widget/template.lua` to `lib/wibox/template.lua` and make it an "official top level concept" rather than a container/widget. The rational is that it has "roles" and "update_function arguments" which are worth force-documenting for each template users. And with the slider, layoutbox and desktop icon PRs on the way, this will be used is a lot of places. I would also remove the `args` and document is as `foo.widget_template = wibox.template { ..., widget = wibox.layout.fixed }`. This would make it a lot less confusing than the current "raw table" we document. It also allows us to print an error is someone wants to pass a widget instance to a template parameter.

If you are ok with this change, then I have a few more commits to port every `widget_template` to (appear to behave like) `wibox.template`. This should be backward compatible.

lib/wibox/widget/template.lua
```diff
@@ -134,44 +158,171 @@ function template:update(args)
...
+    local apply_property
+    apply_property = function(wdgs)
+        for _, widget in ipairs(wdgs) do
+            if widget["set_"..property] then
+                widget["set_"..property](widget, value)
+            end
+
+            if recursive and widget.get_children then
+                apply_property(widget:get_children())
+            end
```
sclu1034, line 257: I never liked this approach of just blindly assigning a property to the whole tree. And I don't see any merit in it over explicitly setting role `id`s where needed.

It's inefficient compared to the `by_id` caches, especially without any checks on how big the tree is that you're recursing into.

The fact that some properties will "magically" be available on all widgets while others need an `id` will lead to confusion (and already does with things like `icon_role` vs `awful.widget.clienticon` in the tasklist examples).

With this being a public API assigning arbitrary properties to arbitrary widgets, chances rise that not every `:set_foo()` expects the same `foo`, i.e. not every widget (especially when taking 3rd-party libraries into account) might expect the same value for the same property name.
E.g. if there was a widget template that used a `fg` or `bg` property to set a color based on a value, then you wouldn't want to apply that to an entire widget tree where there might be `wibox.container.backgrounds` being overwritten.
That wasn't so big of a deal until now, as the only internal implementations of this were `tag` for the taglist and `client` for the tasklist, where you could reasonably assume that any widget that has a `:set_client()` actually wants an `awful.client` instance. But now, there is no way to reason about what a template and its usage may be.

It's also weird that when `ids` is set, it only recurses into their children. I can't think of a use case where I'd set a role on a widget and then also need that role on all of its child widgets. Given our separation of layout/container/content widgets, widgets that have child widgets don't have content to display of their own.

So while it's just an optional parameter, giving the user a choice, I think it's a choice they should never take. And while we could just reword the method docs to heavily discourage using the option, simply removing it would prevent the confusion and make the API simpler.

lib/wibox/widget/template.lua
```diff
@@ -0,0 +1,431 @@
...
+function template.make_from_value(value)
+    if not value then return nil end
+
+    assert(
+        not rawget(value, "is_widget"),
+        "This property requires a widget template, not a widget object.\n"..
+        "Use `wibox.template` instead of `wibox.widget`"
+    )
+
+    if rawget(value, "_is_template") then return value end
```
Aire-One, line 418: It means that with `local new_template = template.make_from_value(another_template)` `new_template` will be the same reference as `another_template`.

Maybe we should return `value:clone()` instead?

lib/wibox/widget/template.lua
```diff
@@ -0,0 +1,431 @@
...
+    args = args or {}
+
+    local ret = wbase.make_widget(nil, nil, { enable_properties = true })
+
+    gtable.crush(ret, template, true)
+    ret._private.connections = {}
+
+    ret:set_template(args.template)
+    ret:set_update_callback(args.update_callback)
+    ret:set_update_now(args.update_now)
```
Aire-One, line 394: Does the `update_callback` still make sense with the new `:set_property` and `:bind_property` methods?

It seems like these 2 new methods are a more powerful and easier way to achieve what the user would have manually written in the `update_callback`.

tests/examples/wibox/widget/template/clone1.lua
```diff
@@ -0,0 +1,149 @@
+--DOC_GEN_IMAGE --DOC_NO_USAGE
```
Aire-One, line 1: This example is quite interesting.

It's actually a basic implementation of the taglist/tasklist widgets with the new template. We only need a way to automatically attach a `self.widget_template:clone()` to each item on the tag/task list and assign them predefined `:set_property` and `:bind_property` to make it a perfect replacement.

Aire-One (commented):
Thank you for the amazing work @Elv13! 

> One thing I would be inclined to do is move the file from lib/wibox/widget/template.lua to lib/wibox/template.lua and make it an "official top level concept" rather than a container/widget. 

Agreed.

> I would also remove the args and document is as foo.widget_template = wibox.template { ..., widget = wibox.layout.fixed }. This would make it a lot less confusing than the current "raw table" we document. It also allows us to print an error is someone wants to pass a widget instance to a template parameter.

This is something I tried to do, but couldn't find a proper way to implement. With the new `template.make_from_value` function, I think it's ok.

> If you are ok with this change, then I have a few more commits to port every widget_template to (appear to behave like) wibox.template. This should be backward compatible.

I can't wait!

lib/wibox/widget/template.lua
sclu1034: As far as I'm aware, those two things are not interchangeable.
`:set_property` is for the developer who creates a widget that _uses_ the template, to assign values at named roles in an unknown template.
`update_callback` is for the user who _creates_ the template, to do stuff when values change.

E.g. `awful.widget.tasklist` will call `:set_property` internally, the user that creates a tasklist will assign an `update_callback` to change stuff in the template dynamically.

Crylia:
Wouldn't it make sense to merge #3521 into this one since other widgets like taglist, tasklist etc got rewritten here too?

sclu1034:
> Wouldn't it make sense to merge #3521 into this one since other widgets like taglist, tasklist etc got rewritten here too?

As a reviewer, I'd rather not have that. There is no rush to migrate these things, and the ones Elv already did blew up the PR tremendously.

lib/wibox/widget/template.lua
```diff
@@ -52,6 +61,46 @@ function template:update(args)
     end
 end
 
+function template:set_template(widget_template)
+    local widget = type(widget_template) == "function" and widget_template()
+        or widget_template
+        or wbase.empty_widget()
+
+    self._private.template = widget
```
sclu1034: We should add an early return when the new value is the same as the old one.

lib/wibox/widget/template.lua
```diff
@@ -52,6 +61,46 @@ function template:update(args)
...
+
+    self._private.update_callback = update_callback
+end
+
+-- @hidden
+function template:set_update_now(update_now)
+    if update_now then
+        self:update()
+    end
+end
```
sclu1034: The name doesn't make sense. `set_[something]` implies a setter for a property, but this just triggers a function call.
And the `update_now` argument exists only to call `self:update()` in the constructor, so the entire method could just be replaced with an `if` in the constructor:

```diff
-    ret:set_update_now(args.update_now)
+    if args.update_now then
+        ret:update()
+    end
```

lib/wibox/template.lua
```diff
@@ -0,0 +1,439 @@
...
+-- @supermodule wibox.widget.base
+---------------------------------------------------------------------------
+
+local wbase = require("wibox.widget.base")
+local gtable = require("gears.table")
+local gtimer = require("gears.timer")
+
+local template = {
+    mt = {},
+    queued_updates = {},
```
sclu1034, line 59: Since the keys in this table will be widgets, it should probably be defined with weak references on those keys.

lib/wibox/widget/template.lua
```diff
@@ -0,0 +1,431 @@
...
+    local w = lazy_load_child(self)
+
+    if not w then
+        return 0, 0
+    end
+
+    return wbase.fit_widget(self, context, w, width, height)
+end
+
+function template:draw() end
```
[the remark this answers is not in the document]
sclu1034: It's not documented in the first place, so it won't show up either way.
But as it's empty, there is no reason to keep the method itself around.

lib/wibox/template.lua
```diff
@@ -0,0 +1,439 @@
...
+    template.queued_updates[self] = nil
+end
+
+--- Update the widget.
+--
+-- This function will call the `update_callback` function at the end of the
+-- current GLib event loop. Updates are batched by event loop, it means that the
+-- widget can only be update once by event loop. If the `template:update` method
+-- is called multiple times during the same GLib event loop, only the first call
+-- will be run.
```
sclu1034, line 138: The "end of the event loop" is different from "end of the event loop iteration". The former will only happen when `awesome.quit()` is called and the main loop is shut down.
And "update are batched" is the opposite of "only the first call will be run" when documented that way.

```suggestion
-- Updates are batched. Multiple calls to @{update} within the same event loop
-- iteration will be collected. At the end of the loop iteration, they will be combined into 
-- a single call to the function provided as @{update_callback}.
```

lib/wibox/widget/template.lua
```diff
@@ -134,44 +158,166 @@ function template:update(args)
 end
 
 --- Change the widget template.
--- @tparam table|widget widget_template The new widget to use as a
+--
+-- Note that this will discard the existing widget instance. Thus, any
+-- `set_property` will no longer be honored. `bind_property`, on the other hand,
+-- will still be honored.
```
sclu1034: The connections stay intact, but until the signal fires again, the value will stay `nil`. This should be made clear, otherwise users might assume that `bind_property` values are updated immediately on the new template.

lib/wibox/widget/template.lua
```diff
@@ -362,34 +362,28 @@ end
...
 
-    ret:set_template(args.template)
-    ret:set_update_callback(args.update_callback)
-    ret:set_update_now(args.update_now)
-
-    -- Apply the received buttons, visible, forced_width and so on
-    gtable.crush(ret, args)
+    ret:set_template(tmpl)
+    ret:set_update_callback(tmpl.update_callback)
+    ret:set_update_now(tmpl.update_now)
```
sclu1034: At this point, I think a separate `create_callback` would make more sense than using this boolean to treat `update_callback` as if it was a `create_callback`.
And this would be required for the "drop-in replacement".

lib/wibox/template.lua
```diff
@@ -232,8 +232,10 @@ function template:set_property(property, value, ids)
     widgets = widgets or {target}
 
     for _, widget in ipairs(widgets) do
-        if widget["set_"..property] then
+        if rawget(widget, "set_"..property) then
             widget["set_"..property](widget, value)
+        else
+            widget[property] = value
```
sclu1034, line 244: For the reviewers' sakes, it would be nice to keep commits separated by purpose and to not mix small, unrelated and easy-to-miss changes into a file rename.

sclu1034 (changes requested):

Aire-One:
> Wouldn't it make sense to merge #3521 into this one since other widgets like taglist, tasklist etc got rewritten here too?

My idea with this Draft PR was to share a little playground where we can preview how the new template API will be used in other widgets. 

I agree with @sclu1034 that this PR is already very big. Include all the `awful.widget` migration here would basically make this PR reasonably unmergable. I think it's better to have a base API for the template, then iteratively migrate other widgets to the new API. There is some implementation decision we have to think in a per-widget basis to do the migration, so only 1 PR for all the widgets is a bad idea anyway. (One of them is the fact that some widgets comes with a cache implementation that is incompatible with the template idea)

> the ones Elv already did blew up the PR tremendously.

Yeah, 1k new lines is a big diff.

(I haven't reviewed the new commits.) [If] the migration included in this PR only includes migrating the old `widget_template` from the tasklist and taglist (as discussed earlier), I think it's justified to have it here. I'm fine with `wibox.template` been used (and proofed working as intended/backward compatible) from the first PR.

lib/wibox/widget/template.lua
Aire-One: This method is required to make the declarative pattern work as expected.

When creating a widget with the declarative pattern, the constructor doesn't receive all the properties as parameters. Instead, the properties are updated after the widget instantiation.

```lua
local w = wibox.widget {
   update_callback = update_fct,
   update_now = true,
   widget = widget.template,
}
```

The  `update_now = true` property is "transformed" to the invocation of the `template:set_update_now` method.

Note that this is why the whole method is documented as a "Hack" and is `@hidden`.

lib/wibox/widget/template.lua
Elv13: For the drop-in part, I was just planning at leaving the old code in place. The issue is the function signature. It has the tag/client as part of it, so it's hard to mutualize into a common API. There are ways around it, but it seems simpler to omit `create_callback` given it's not very useful. Something like `gears.connection` would be a better fit. I may extract it from the `gears.reactive` PR (even if they share some code) since it isn't as affected by the linting issues as the other 2 parts.

The whole callback thing is quite unique in our API when you think about it. We nearly always use signals for this kind of stuff. As you pointed elsewhere, it *does* make sense to keep it here because it makes the users life easier (I agree). `gears.connection` would overlap, but would still have some learning curve issue and has some footguns (like connecting to signals which don't exist).

lib/wibox/template.lua
Elv13: Yeah, sorry, I messed up a rebase somewhere.

Elv13:
Mostly all done. Plus I added a new commit to fix the `forced_width`/`forced_height` behavior, which wasn't very consistent depending on how the property was set.

lib/wibox/widget/template.lua
sclu1034: Then it would make more sense to just call it `update_now()`, which `drill` [also checks for](https://github.com/awesomeWM/awesome/blob/17eee18d8bd500ac6dc4907701c5ccdb71174c0c/lib/wibox/widget/base.lua#L763-L764).

Elv13: From a clean code PoV, maybe, but at the same time adding too many `if`s to `drill()` might not be super scalable. Another of my PR has some code to make meta-extending the DOM easier using some `_` prefixed extender "I am special, call this and I will do my special stuff" functions. That might be the solution on the longer term.

However for now I think the hack proposed here is good enough. Unexpected, but not in a way which impact end users.

In the even longer term, which is dubious, there is also that issue which calls to replace how the DOM is created and move it to the hierarchy because in practice right now, you can't merge 2 DOMs and that makes `:get_children_by_id()` unreliable when some composed widgets are used... Including this very widget.... This would be an opportunity to revisit some of the bolt-on hacks I made to restore the declarative widget system. Historically, AwesomeWM 3.1-3.4 had it, but 3.5 had a fully rewritten (and much, much better) widget system. It was fully imperative, which annoyed a large number of users who could not port their config. The code was also very heavy for something like `rc.lua`, which is supposed to appear like a config file. Back then I mostly did modules rather than contribute to the core, so I wrote the declarative thing as [the retrograde module](https://github.com/Elv13/retrograde). It eventually got reworked/merged into the core, but was always a bolted on hack rather than something engineered alongside `wibox.hierarchy`. So it would benefit from a larger rework to be more flexible/integrated.

Aire-One: > From a clean code PoV, maybe, but at the same time adding too many ifs to drill() might not be super scalable.

I'm not sure to follow what you mean. This `if` branch already exists. 

I have to agree with @sclu1034. A setter is unneeded here. I have somehow missed this feature of `drill`, otherwise I would have used it instead.

PR #3436, opened 2021-09-16
Fix setting the screen in a notification preset.

Elv13:
This is used by lain. It didn't work in both legacy and `ruled.notification` mode for different reasons.

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -191,6 +191,10 @@ naughty.notifications = { suspended = { }, _expired = {{}} }
 
 naughty._active = {}
 
+local function get_screen(s)
+    return s and capi.screen[s]
+end
+
 local function init_screen(s)
     if naughty.notifications[s] then return end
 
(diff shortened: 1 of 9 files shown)


Aire-One (approved):
I think this is alright. 

I'm not sure what you want to test with the FIXME parts in the test-naughty-screen... Is this another memory leak test?

Elv13:
> I'm not sure what you want to test with the FIXME parts in the test-naughty-screen... Is this another memory leak test?

Done. It was never enabled. But apparently it just works now.

Elv13:
Nevermind, I recall now. It only fails on Lua 5.1.

Elv13:
I extended the test to avoid relying on the weak table and rather manually remove the screen using the "property::screen" signal. This prevent the conditions where Lua 5.1 fail from happening.

I also cherry-picked a commit from the other notification PR and added a random fix for a flacky test.

Elv13:
I fixed another couple of tests, but there is still flakyness related to both random tests failing with GitHub Actions *and* code related to this PR. But somehow, not *always*. The garbage collection tests which was commented now passes on Lua 5.1, but sometime fails on luajit. test-struts seem to currently fail on luajit, but was passing a few hours ago. The wallpaper PR also has random tests mysteriously failing. It will keep happening for a while given we just moved to much faster hosts.

I will continue to hunt them down, but i think it doesn't have to block merging this.

Aire-One:
Thanks for the details. Let's see how annoying these flacky tests will be waiting to a proper fix.

Aire-One (approved):

Elv13:
proof that it sometime passes those tests.

Anyway, added 3 more commit:

One of them is a quality of life to be able to test external configs with submodules without installing them.
One is a fix when the first config crash and it loads the fallback one
One of them is to allow *notification messages* to be set in the configuration presets. It makes no sense conceptually, but it worked before and `lain` abused of it as a hack to set the text after the notification was created.

Aire-One (approved):

actionless (approved):

PR #3439, opened 2021-09-17
Replace /usr/bin/which with command -v

madduck:
The `which` utility is being phased out since POSIX has mandated
`command -v` for the same task, so let's use that in the awesome-client
script.

Signed-off-by: martin f. krafft <madduck@madduck.net>

### DIFF
--- a/utils/awesome-client
+++ b/utils/awesome-client
@@ -11,7 +11,7 @@ then
     # than bash's "read" itself.
     # It can be disabled/overridden using 'AWESOME_RLWRAP= awesome-client'.
     if [ -z "${AWESOME_RLWRAP+x}" ]; then
-        AWESOME_RLWRAP="$(which rlwrap 2>/dev/null)"
+        AWESOME_RLWRAP="$(command -v rlwrap 2>/dev/null)"
     fi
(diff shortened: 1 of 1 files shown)


actionless:
have you tried to read your diff yourself before submitting?

utils/awesome-client
```diff
@@ -7,11 +7,11 @@ set -o pipefail
 
 if [ -t 0 ]  # is a tty.
 then
-    # rlwrap provides readline functionality for "read", which is more enhanced
+    # rlwrap provides readline functionality for "read", command -v is more enhanced
```
sclu1034, line 10: What @actionless is probably hinting at:
It looks like you just ran a "replace all", which also affected this comment.

sclu1034 (changes requested):

madduck:
Lol, @actionless and @sclu1034, I am sorry to have wasted your time. I had made the commit manually and properly, then made a Git mistake, and then got fed up and did a replace-all, without checking. My bad. Updated now.

actionless (approved):

Aire-One (approved):

PR #3443, opened 2021-09-20
Improve suspending the notifications

Elv13:


Elv13:
Rebased to fix a merge conflict from the preset PR

Aire-One (approved):
It seems to work from my quick test!

(Note: I still have the issue of "blocking prompt when calling multiple times notify-send with suspended = true" that I mentioned in the issue thread.
> ```
> awesome-client "require('naughty').suspended = true"
> notify-send "test1"
> awesome-client "require('naughty').suspended = false"
> notify-send "test2"
> ```
> if I run multiple times the notify-send "test1", the prompt "freeze" (I have to Ctrl+C to get it back)

)

Elv13:
> (Note: I still have the issue of "blocking prompt when calling multiple times notify-send with suspended = true" that I mentioned in the issue thread.

Whoops, missed that bug. I fixed it.

Aire-One:
> @Aire-One @actionless Do you think this is ready to merge?

Yup, your last work on this branch seems to have fixed the last issue I found. I wasn't sure this could be a bug from your implementation. Thanks for taking care of it! 

> (the next PR is the timer one to allow proper do-not-disturb support based on the time of the day)

Does this feature need to be implemented by our API? It sounds like something that a third-party module should implement our be in our recipes page.

Elv13:
> Does this feature need to be implemented by our API? It sounds like something that a third-party module should implement our be in our recipes page.

I plan to do it in a documentation example. It will use the new `gears.timer` API I am working on.

Aire-One:
Ok, let's see where it goes, then 🚀

PR #3445, opened 2021-09-21
Correction of error message in awful.keygrabber.run_with_keybindings

poisson-aerohead:
Two calls to gdebug.print_warning() in run_with_keybindings use an
argument that concatenates a fixed string with the result returned by a
call to gdebug.dump(). gdebug.dump() is the debug library function for
immediate printing. It has no return value. This causes the correct
message in print_warning to not be printed, and also causes a
naughty.notify box to appear warning of an attempt to concatenate to an
empty variable.

The call to gdebug.print_warning should have an argument made by
concatenating to gdebug.dump_return(). Incidentally, this is in fact the
function used internally by gdebug.print().

Signed off by: Brian Sobulefsky <brian.sobulefsky@protonmail.com>

### DIFF
--- a/lib/awful/keygrabber.lua
+++ b/lib/awful/keygrabber.lua
@@ -714,12 +714,13 @@ function keygrab.run_with_keybindings(args)
             else
                 gdebug.print_warning(
                     "The hook's 3rd parameter has to be a function. " ..
-                        gdebug.dump(v or {})
+                        gdebug.dump_return(v or {})
                 )
             end
         else
             gdebug.print_warning(
-                "The keybindings should be awful.key objects".. gdebug.dump(v or {})
+                "The keybindings should be awful.key objects" ..
+                        gdebug.dump_return(v or {})
             )
         end
     end


Aire-One (approved):

Aire-One:
Oh cool, we have to approve new contributor to allow GitHub Actions to be run.

poisson-aerohead:
Haha sorry, I feel like this fix is too small to even be worth the effort for you. Imagine my surprise when I got the error message and went to look what "I" did wrong and I actually found a bug that was fixable with a few minutes of reading around.

Aire-One:
Nahh, this is perfect. There is no "too small fix". Your contribution is really appreciated 😀

I'm happy people are hacking around their way in awesome. Thanks for the extra effort of submitting a PR.

Elv13:
> Do you know what failed in the lua5.2 test?

We migrated to GitHub Actions (from Travis.ci) last week. There is still some papercuts. In that case, an authentication issue with the code coverage service.

Elv13 (approved):

poisson-aerohead:
> > Do you know what failed in the lua5.2 test?
> 
> We migrated to GitHub Actions (from Travis.ci) last week. There is still some papercuts. In that case, an authentication issue with the code coverage service.

Sounds good, as long as it is something else. I could not imagine a change this unobtrusive caused so many items to fail.

PR #3448, opened 2021-09-27
Content (Screenshot) API Expansion

poisson-aerohead:
Updated API in the same interface as the client.content property to support screenshots at the root window and screen levels. A call to the root.content() method will return a screenshot as a Cairo surface of the entire root window (generally all physical screens). Getting the screen.content property will return a screenshot as a Cairo surface of the screen object (generally a physical screen) just as client.content will for a client object.

Sample usage - the traditional API supported focused client screenshot as:

  c = client.focus
  if c then
    gears.surface(c.content):write_to_png("/path/to/screenshot.png")
  end

Similarly, this API extension adds:

  s = awful.screen.focused()
  if s then
    gears.surface(s.content):write_to_png("/path/to/screenshot.png")
  end

for the screen class and:

  gears.surface(root.content()):write_to_png("/path/to/screenshot.png")

for the root window. Note that the example shows how to get a screenshot of the focused screen, but this is not a limitation. A lua script could call it on any screen object.

Signed off by Brian Sobulefsky <brian.sobulefsky@protonmail.com>

### DIFF
--- a/lib/wibox/widget/graph.lua
+++ b/lib/wibox/widget/graph.lua
@@ -20,9 +20,12 @@
 local setmetatable = setmetatable
 local ipairs = ipairs
 local math = math
+local math_max = math.max
+local math_min = math.min
 local table = table
-local type = type
 local color = require("gears.color")
+local gdebug = require("gears.debug")
+local gtable = require("gears.table")
 local base = require("wibox.widget.base")
 local beautiful = require("beautiful")
 
@@ -39,8 +42,6 @@ local graph = { mt = {} }
 
 --- Set the graph border color.
 --
--- If the value is nil, no border will be drawn.
---
 --@DOC_wibox_widget_graph_border_color_EXAMPLE@
 --
 -- @property border_color
@@ -50,6 +51,8 @@ local graph = { mt = {} }
 -- @see gears.color
 
 --- Set the graph foreground color.
+--
+-- This color is used, when `group_colors` isn't set.
 --
 --@DOC_wibox_widget_graph_color_EXAMPLE@
 --
@@ -69,38 +72,90 @@ local graph = { mt = {} }
 -- @propemits true false
 -- @see gears.color
 
---- Set the maximum value the graph should handle.
+--- Set the colors for data groups.
+--
+-- Colors in this table are used to paint respective data groups.
+-- When this property is unset (default), the `color` property is used
+-- instead for all data groups.
+-- When this property is set, but there's no color for a data group in it
+-- (i.e. `group_colors`[group] is nil or false), then the respective
+-- data group is disabled, i.e. not drawn.
+--
+-- @DOC_wibox_widget_graph_stacked_group_disable_EXAMPLE@
+--
+-- @property group_colors
+-- @tparam table colors A table with colors for data groups.
+-- @see gears.color
+
+--- The maximum value the graph should handle.
+--
+-- This value corresponds to the top of the graph.
+-- If `scale` is also set, the graph never scales up below this value, but it
+-- automatically scales down to make all data fit.
+-- If `scale` and `max_value` are unset, `max_value` defaults to 1.
 --
 -- @DOC_wibox_widget_graph_max_value_EXAMPLE@
---
--- If "scale" is also set, the graph never scales up below this value, but it
--- automatically scales down to make all data fit.
 --
 -- @property max_value
 -- @tparam number max_value
 -- @propemits true false
 
---- The minimum value.
+--- The minimum value the graph should handle.
+--
+-- This value corresponds to the bottom of the graph.
+-- If `scale` is also set, the graph never scales up above this value, but it
+-- automatically scales down to make all data fit.
+-- If `scale` and `min_value` are unset, `min_value` defaults to 0.
 --
 -- @DOC_wibox_widget_graph_min_value_EXAMPLE@
 --
--- Note that the min_value is not supported when used along with the stack
--- property.
 -- @property min_value
 -- @tparam number min_value
 -- @propemits true false
 
 --- Set the graph to automatically scale its values. Default is false.
 --
---@DOC_wibox_widget_graph_scale1_EXAMPLE@
+-- If this property is set to true, the graph calculates
+-- effective `min_value` and `max_value` based on the displayed data,
+-- so that all data fits on screen. The properties themselves aren't changed,
+-- but the graph is drawn as though `min_value`(`max_value`) were equal to
+-- the minimum(maximum) value among itself and the currently drawn values.
+-- If `min_value`(`max_value`) is unset, then only the drawn values
+-- are considered in this calculation.
+--
+-- @DOC_wibox_widget_graph_scale1_EXAMPLE@
 --
 -- @property scale
 -- @tparam boolean scale
 -- @propemits true false
 
---- Set the width or the individual steps.
+--- Clamp graph bars to keep them inside the widget for out-of-range values.
 --
--- Note that it isn't supported when used along with stacked graphs.
+-- Drawing values outside the [`min_value`, `max_value`] range leads to
+-- bar shapes that exceed physical widget dimensions.
+-- Most of the time this doesn't matter, because bar shapes are rectangles
+-- and bar heights aren't large enough to trigger errors in the drawing system.
+-- However for some shapes and values it does make a difference and leads
+-- to visibly different and/or invalid result.
+--
+-- When this property is set to true (the default), the graph clamps
+-- bars' heights to keep them within the graph.
+--
+-- @DOC_wibox_widget_graph_clamp_bars_EXAMPLE@
+--
+-- @property clamp_bars
+-- @tparam boolean clamp_bars
+-- @propemits true false
+
+--- The value corresponding to the starting point of graph bars. Default is 0.
+--
+-- @DOC_wibox_widget_graph_baseline_value_EXAMPLE@
+--
+-- @property baseline_value
+-- @tparam number baseline_value
+-- @propemits true false
+
+--- Set the width or the individual steps.
 --
 --@DOC_wibox_widget_graph_step_EXAMPLE@
 --
@@ -109,8 +164,6 @@ local graph = { mt = {} }
 -- @propemits true false
 
 --- Set the spacing between the steps.
---
--- Note that it isn't supported when used along with stacked graphs.
 --
 --@DOC_wibox_widget_graph_step_spacing_EXAMPLE@
 --
@@ -129,27 +182,58 @@ local graph = { mt = {} }
 
 --- Set the graph to draw stacks. Default is false.
 --
---@DOC_wibox_widget_graph_stacked_EXAMPLE@
+-- When set to true, bars of each successive data group are drawn on top of
+-- bars of previous groups, instead of the baseline.
+-- This necessitates all data values to be non-negative.
+-- Negative values, if present, will trigger @{nan_color|NaN indication}.
+--
+-- @DOC_wibox_widget_graph_normal_vs_stacked_EXAMPLE@
+--
 -- @property stack
 -- @tparam boolean stack
 -- @propemits true false
 
---- Set the graph stacking colors. Order matters.
+--- Display NaN indication. Default is true.
 --
--- @property stack_colors
--- @tparam table stack_colors A table with stacking colors.
-
---- The graph background color.
+-- When the data contains [NaN](https://en.wikipedia.org/wiki/NaN) values,
+-- and `nan_indication` is set, the corresponding area,
+-- where the value bar should have been drawn, is filled
+-- with the `nan_color` from top to bottom.
+-- The painting is done after all other data is rendered,
+-- to make sure that it won't be overpainted and go unnoticed.
 --
--- @beautiful beautiful.graph_bg
--- @param color
+-- @DOC_wibox_widget_graph_nan_color_EXAMPLE@
+--
+-- @property nan_indication
+-- @tparam boolean nan_indication
+-- @propemits true false
 
---- The graph foreground color.
+--- The color of NaN indication.
+--
+-- The color used when `nan_indication` is set.
+-- Defaults to a yellow-black diagonal stripes pattern.
+--
+-- @DOC_wibox_widget_graph_stacked_nan_EXAMPLE@
+--
+-- @property nan_color
+-- @tparam gears.color nan_color The color of NaN indication.
+-- @propemits true false
+-- @see gears.color
+
+--- The graph foreground color
+-- Used, when the `color` property isn't set.
 --
 -- @beautiful beautiful.graph_fg
 -- @param color
 
+--- The graph background color.
+-- Used, when the `background_color` property isn't set.
+--
+-- @beautiful beautiful.graph_bg
+-- @param color
+
 --- The graph border color.
+-- Used, when the `border_color` property isn't set.
 --
 -- @beautiful beautiful.graph_border_color
 -- @param color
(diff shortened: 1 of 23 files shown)


objects/screen.c
```diff
@@ -1655,6 +1658,60 @@ luaA_screen_get_name(lua_State *L, screen_t *s)
     return 1;
 }
 
+/** Get the content of the active screen as a cairo surface
```
Aire-One, line 1661: ldoc sometime is weird when it comes to find the limit of the description. It's better to always add a final `.` at the end of the summary:

```suggestion
/** Get the content of the active screen as a cairo surface.
```

objects/screen.c
```diff
@@ -1655,6 +1658,60 @@ luaA_screen_get_name(lua_State *L, screen_t *s)
...
+ *
+ * @return A cairo surface
+ * @staticfct content
```
Aire-One, line 1664: Since it's a property ("dynamic property"?), we shouldn't document it as a method that returns something, but as a property of a given type :

```suggestion
 * @property content
 * @tparam surface A cairo surface with the screen content.
 * @see gears.surface
```

(+ I have added the same @see as client.content property)

objects/screen.c
```diff
@@ -1655,6 +1658,60 @@ luaA_screen_get_name(lua_State *L, screen_t *s)
...
+ */
+static int
+luaA_screen_get_content(lua_State *L, screen_t *s)
+{
+
```
Aire-One, line 1669:
```suggestion
```

objects/screen.c
```diff
@@ -1655,6 +1658,60 @@ luaA_screen_get_name(lua_State *L, screen_t *s)
...
+    cairo_rectangle(scrn_cr, 0, 0, scrn_geom->width, scrn_geom->height);
+    cairo_fill(scrn_cr);
+
+    cairo_surface_destroy(root_surface);
+    cairo_destroy(scrn_cr);
+
+    lua_pushlightuserdata(L, scrn_surface);
+
+    return 1;
+
```
Aire-One, line 1709:
```suggestion
```

root.c
```diff
@@ -495,6 +495,31 @@ luaA_root_wallpaper(lua_State *L)
     return 1;
 }
 
+
+/** Get the content of the root window as a cairo surface
```
Aire-One, line 499:
```suggestion
/** Get the content of the root window as a cairo surface.
```

root.c
```diff
@@ -495,6 +495,31 @@ luaA_root_wallpaper(lua_State *L)
...
+ *
+ * @return A cairo surface
+ * @staticfct content
```
Aire-One, line 502: I think it should be documented as a module level property (aka field). But it seems other similar fields are documented as object properties, tho. So I think we should also use :

```suggestion
 * @property content
 * @tparam surface A cairo surface with the root window content (aka the whole surface from every screens).
 * @see gears.surface
```

root.c
```diff
@@ -495,6 +495,31 @@ luaA_root_wallpaper(lua_State *L)
...
+ */
+static int
+luaA_root_get_content(lua_State *L)
+{
+
```
Aire-One, line 507:
```suggestion
```

root.c
```diff
@@ -495,6 +495,31 @@ luaA_root_wallpaper(lua_State *L)
...
+    int height = globalconf.screen->height_in_pixels;
+
+    surface = cairo_xcb_surface_create(globalconf.connection,
+                                       globalconf.screen->root,
+                                       globalconf.default_visual,
+                                       width, height);
+
+    lua_pushlightuserdata(L, surface);
+    return 1;
+
```
Aire-One, line 519:
```suggestion
```

Aire-One (commented):
This is a good work!

Feel free to dismiss any comment/suggestion I made.

NIT: I think properties description from `screen.c` should be placed at the file beginning, next to all other properties' documentation.

NIT2: I think we should also include documentation similar to the `client.content` property, with the usage example because the user must copy the surface to use it : https://github.com/awesomeWM/awesome/blob/eddabebac46d94b5afd2916460f22b2057628b2b/objects/client.c#L837-L864

objects/screen.c
poisson-aerohead: As I am not a project member, I am going to believe you on all these sorts of things. I tried to basically mimic what was in the source files, but I have only been looking at them for a few days now.

objects/screen.c
poisson-aerohead: Same comment. However the project likes things documented should be done.

objects/screen.c
poisson-aerohead: Whatever the white space policy is...

root.c
poisson-aerohead: Is it a property? It appeared to me that root is a library that only has functions? I'll press accept if this is what is wanted though.
https://awesomewm.org/doc/api/libraries/root.html

poisson-aerohead:
I think I fixed the commit history. Sorry, I am new to doing distributed development. Anyway, now I know a little bit about about git rebase.

Hopefully fixing the commit history does not break this PR.

poisson-aerohead:
Hey for what its worth I just developed a rough snip tool in my rc.lua (since I am already running a version of awesome with this api extension incorporated). It is nothing fancy. I mapped it to Super-Shift-PrntScrn like I said. It starts a mousegrabber and swaps the cursor for cross hairs. You click Button 1 in two locations to define the rectangle you want, and it takes a snipped screen shot of just that location. You can Button 3 to exit without taking the image.

Elv13:
This looks Awesome!

> Lua 5.2 is brutal with these tests...

This isn't your fault. We moved from Travis CI to GitHub Actions a few weeks ago and I have yet to iron out some issues.

I think what we need first here is a new `tests/test-screenshot.lua` file to call the screen/root/client `content` properties.


Also, more generally. Can we have an higher level API design discussion? It isn't 100% related to this contribution itself, which look pretty good as it stands. However, I wonder if we should create a proper high level `awful.screenshot` module. Right now, `client.content` is very cumbersome to use. It "works", but there is a lot of confusing boilerplate code associated with the raw surface. People who follow AwesomeWM development knows I am not a big fan of APIs which require boilerplate code to become useful.

What is people opinion here. Should we have `client.screenshot`/`screen.screenshot`/`root.screenshot` properties which return real/usable Cairo surface (with a built-in full copy) or should we go the `awful.screenshot` route? `awful.screenshot` could have something like:

 * save method with png and svg support
 * `rectangle` property to set arbitrary area (users could build GUIs around it?)
 * `screen`/`screens`/`client` properties (with `root` as the default screenshot mode)
 * a `widget` property to generate an usable widget
     * some aspect ratio properties
     * some refresh rate properties
     * max size to avoid needlessly keeping large surfaces in memory
 * a `:refresh()` method
 * an `auto_refresh` (number of seconds) to automate refresh? Maybe as 2 distinct properties?
 * a `delay` property? Maybe with some nice signals each seconds to implement some GUI?
 * `:crop()` method?
 
What else is needed? Also, this is just a first API draft, it could be improved in many ways. The advantage of a self-contained modules is that we can extend it without having to extend 3 different objects. It also helps the documentation a lot since everything about the "screenshot" topic is documented in 1 place.

poisson-aerohead:
I don't want to tell you how to run your project, but IMO the current implementation (if you include my changes from this PR) are sufficient for the "built ins" to keep with the Awesome "anything that can be done in Lua, is" philosophy. That said, I could see adding an awful module to make it easier.

FWIW, at this level of level of API, I got all 4 methods working, and 3 take essentially no code (most is the awful.keys boilerplate). The snip takes a little bit more (the snip is also very basic right now). Right now, this is in my rc.lua, which I think is a bit of a clog, but it could be moved to a library file. Also, the snip routine could have a wibox show the region that will be snipped (I am working on that part next). See below:

        -- Root window screen shot
        awful.key({ modkey }, "Print",
                function()
                  screenshot_dir = os.getenv("HOME") .. "/Images/Screenshots/"
                  date_time = tostring(os.date("%Y%m%d%H%M%S"))
      	    gears.surface(root.content()):write_to_png(screenshot_dir .. "Screenshot-" .. date_time .. ".png")
                end,
                  {description = "Print Screen - Root Window", group = "awesome"}),
        -- Physical screen screen shot
        awful.key({ modkey, "Mod1" }, "Print",
                function()
                  screenshot_dir = os.getenv("HOME") .. "/Images/Screenshots/"
                  date_time = tostring(os.date("%Y%m%d%H%M%S"))
                  s = awful.screen.focused()
                  if s then
                    gears.surface(s.content):write_to_png(screenshot_dir .. "Screenshot-" .. date_time .. ".png")
                  end

                end,
                  {description = "Print Screen - Physical Screen", group = "awesome"}),
        -- Client window screen shot
        awful.key({ modkey, "Control" }, "Print",
                function()
                  screenshot_dir = os.getenv("HOME") .. "/Images/Screenshots/"
                  date_time = tostring(os.date("%Y%m%d%H%M%S"))
                  c = client.focus
                  if c then
                    gears.surface(c.content):write_to_png(screenshot_dir .. "Screenshot-" .. date_time .. ".png")
                  end
                end,
                  {description = "Print Screen - Client Window", group = "awesome"}),
        awful.key({ modkey, "Shift" }, "Print",
                function()
                  screenshot_dir = os.getenv("HOME") .. "/Images/Screenshots/"
                  local result = {}
                  mousegrabber.run(
                           function(mouse_data)
                             if mouse_data["buttons"][3] then
                               return false
                             end
      		       if result[1] then
                               if mouse_data["buttons"][1] then
                                 local min_x, max_x, min_y, max_y
                                 local source, target, cr
                                 min_x = math.min(result[1], mouse_data["x"])
                                 max_x = math.max(result[1], mouse_data["x"])
                                 min_y = math.min(result[2], mouse_data["y"])
                                 max_y = math.max(result[2], mouse_data["y"])
                                 if min_x == max_x or min_y == max_y then
                                   return false
                                 end
                                 source = gears.surface(root.content())
                                 target = source:create_similar(cairo.Content.COLOR,
                                                        max_x - min_x, max_y - min_y)
                                 cr = cairo.Context(target)
                                 cr:set_source_surface(source, -min_x, -min_y)
                                 cr:rectangle(0, 0, max_x - min_x, max_y - min_y)
                                 cr:fill()
                                 date_time = tostring(os.date("%Y%m%d%H%M%S"))
                                 gears.surface(target):write_to_png(screenshot_dir ..
                                                 "Screenshot-" .. date_time .. ".png")
                                 return false
                               end
                             else
                               if mouse_data["buttons"][1] then
                                 result[1] = mouse_data["x"]
                                 result[2] = mouse_data["y"]
                               end
                             end
                             return true

                           end, "crosshair")
                end,
                  {description = "Snip Screen", group = "awesome"})
    )

poisson-aerohead:
Hi, not to clog the thread, but I have the full Snip tool working. As @Elv13 pointed out, this belongs in the Lua libraries. Whether you want to gather everything in to awful.screenshot or whatever can be determined by the project.

Please see https://gist.github.com/poisson-aerohead/2c8a3eede9d26d9fe7306d2adb9e98df

As noted, I took the frame from mouse.snap, so much of that logic should become a utility in the library usable by mouse.snap, awful.screenshot.snip, and maybe others.

I don't want to submit these changes to the rc.lua as I don't think it belongs there.

Aire-One:
@Elv13, maybe this PR can be kept to minimal and merged quickly (with the addition of necessary tests). A low-level API can be good enough for now, I think. We can take more time to work on awful.screenshot as a new API later.

> I think I fixed the commit history. Sorry, I am new to doing distributed development. Anyway, now I know a little bit about about git rebase.

Oops, I think we had a misunderstanding... It was alright that your code was split in 2 commits. What I want you to remove from your branch are the 2 firsts commits (a173254e17c90ebf8e1d8c9adbaa5743fe479ecd and a173254e17c90ebf8e1d8c9adbaa5743fe479ecd) that (I think) are part of your other PR that was merged before.

> I don't want to submit these changes to the rc.lua as I don't think it belongs there. I can live with my bloated rc.lua until the new lua library stabilizes.

Yup, I don't think this is relevant to add special keybindings for such feature in the default awesomerc.lua. However, we have a receipt page in our website where you can submit the code for your snippet tool: https://awesomewm.org/recipes/

poisson-aerohead:
> Oops, I think we had a misunderstanding... It was alright that your code was split in 2 commits. What I want you to remove from your branch are the 2 firsts commits

I have no idea why those commits are showing here, maybe because it was the last thing done, or there were errors in the actions? They were approved in a separate PR: https://github.com/awesomeWM/awesome/pull/3445


> However, we have a receipt page in our website where you can submit the code for your snippet tool:

I'll post something there, but I agree with @Elv13 that there should be a more user centered framework built into the Lua libraries (awful.screenshot or whatever). This, of course, cannot be done until this PR is merged. Since I now have 2 Awesomes on my computer (the official build from the distro under /usr and my modified build /usr/local) I am probably going to start working on library functions so I can clean up my rc.lua. If this is merged and I have something good I can submit a PR in the future.

poisson-aerohead:
> I didn't read the discussion so far, so sorry if someone already mentioned this, but: Is `screen.content` in C necessary? This can just as well be implemented in Lua based on the `root.content` API that this PR also adds.

Yes, I started by adding them in C, and when I did the snipper in Lua it was obviously possible to use that method to do the physical screen level as well. It could not be a method of a screen class itself, as that is in C, but it could be in awful.screen or something like that. I'm not sure if you want to skip screen.content or leave it as a convenience function.

psychon:
>  if you want to skip screen.content or 

Adding it as `awful.screen.object.get_content(s)` should make it available as just `s.content` so that other Lua code cannot tell the difference. This is thanks to some magic that @Elv13 implemented.

poisson-aerohead:
> Adding it as `awful.screen.object.get_content(s)` should make it available as just `s.content` so that other Lua code cannot tell the difference

Wow I did not know that. If you want to remove sreen.content from screen.c and implement an awful.screen.content as described above, I am OK with that change.

poisson-aerohead:
I have made a rough awful.screenshot.lua that exposes methods init (to set the preferred screenshot directory, prefix for the screenshot file names, and color for the snipper outline), screenshot.root() to take a root window (takes optional arguments to use a different directory and filename prefix), screenshot.screen to take a physical screen screenshot (takes optional arguments to use a different directory, filename prefix, and screen other than awful.screen.focus), screenshot.client to take a client window screenshot (takes optional directory and filename prefix), screenshot.snipper to launch an interactive snipper tool to take cropped screenshots (with optional directory and prefix), and screenshot.snip to take a programmatically defined cropped shot (it takes a mandatory geometry to define the crop and optional directory and prefix).

I will hang onto this until the current PR is successfully merged. It sounds like whether we keep this PR as is or move screen.content to the Lua libraries will be transparent.

While my awful.screenshot.lua is functional, it needs a lot of eyes on it I think.

actionless (changes requested):
it's not only should be moved to lua, but also covered with some basic tests

(i think it would make sense to use some area without text for screenshoting (to avoid false-negative test result when font rendering changes and not to depend on specific typeface etc), for example, you could make an empty wibox with a square inside, and take a screenshot only of that area

poisson-aerohead:
What is the best way to update this PR to move the screen object routine to the Lua libraries? I don't want to roadblock the flow of the PR.

actionless:
and the same for `root ` please

poisson-aerohead:
> and the same for `root ` please

@actionless the root.content() method cannot be emulated any other way through. This is what sent me down the rabbit hole of examining the awesome source code to find how the traditional client.content was implemented.

If anything, technically client.content no longer needs to be implemented in C. It can be in awful.client (assuming that works the same as awful.screen?). I will point out that there is technically a performance impact to this change beyond the usual C/Lua penalty that was not applicable to screen.content. The screen.content routine required a root window screenshot and a cropping operation in both the C implementation and the Lua implementation. A Lua implementation of client.content will require a root window screenshot and a cropping operation as well. The C implementation of client.content did _not_  require a root window screenshot or a cropping operation, as the C file has access to the clients xcb_window_t and so could request that content directly.

Considering the performance of my snip tool that I made (which obviously requires a root window screenshot and a crop), I would expect this impact to be minimal. If the project _really_ wants to adhere to "everything that can be done in Lua, is" then I guess client.content should be moved to Lua.

actionless:
yes, you can use awful.* the same for client and root

lib/awful/screen.lua
```diff
@@ -1015,7 +1062,40 @@ object.properties(capi.screen, {
     auto_emit    = true,
 })
 
---@DOC_object_COMMON@
```
actionless, line 1018: ?

poisson-aerohead: I'm confused about that. It keeps telling me there is nothing so get from upstream.

psychon:
> If anything, technically client.content no longer needs to be implemented in C. 

Sorry, but no, not really. `c.content` gives you a xcb surface for that specific client window. Both `root.content` and `screen.content` used the root window. There are differences that... well.... don't really matter in practice, I guess.

When a client is not visible (is on an unselected tag), using `c.content` will cause lots of X11 protocol errors. When a client is partly behind another client, `c.content`... technically has undefined content for the part of a window that is covered, but in practice it will most of the time (but not always) simply also contain the contents of the window that is in front of `c`.

poisson-aerohead:
So I will leave client as it.

@actionless I figured out why I messed up the comments. I'll get it reverted.

lib/awful/screen.lua
poisson-aerohead: I figured out what happened. I followed the build flags from the Archlinux repo to build it and I think there is something that changes the documentation comments. I developed the screen.lua live in /usr/local/share/awesome/lib and copied back to my home directory when it was working. This introduces the changes to the comments.

Now for some reason the patch is showing additions to client.lua and deletions from client.c. The file is correct, though. I can't figure out why that is.

poisson-aerohead:
Hey sorry everyone, I tried really hard to send a small and properly tested patch when I first submitted and somehow this got out of control. I am new to doing this but to be fair I was also given a lot of conflicting instructions. I've been hammering you all but I got it in the state that you should want. Currently:

root.content() is in C (it has to be)
screen.content is in Lua (fits with philosophy)
client.content is in C (it _should_ be according to @psychon although there is some desire to move it).

Please let me know.

psychon:
@poisson-aerohead Thanks a lot for your contribution and just a small clarification: You are not hamering us. You wanted to contribute, which is great, and we are making this hard for you by requiring more changes and disagreeing about how things should look. You are doing nothing wrong and I am quite grateful that you are doing this work.

poisson-aerohead:
OK, so one thing I will point out is that the changes made early in this PR was putting the comments into your style. Now that the changes in screen.c are taken out and there are new changes to screen.lua, I would recommend putting whatever doc comments you need and I'll accept them.

actionless:
just scroll 10 lines up and see how it's done for very similar function in screen.lua :)

and stuff from root.c still need to be moved to lua

poisson-aerohead:
I have added what I believe to be DOC comments in the style you need. @actionless, as I said before, I do not see how we can emulate root.content() in Lua. We were able to move screen.content to Lua because it is dependent on calling root.content() and then cropping to the screen geometry. I do not believe there was a pre existing method of getting a full root window screenshot before I added content() to root.c. If there is something I can use, please let me know.

lib/awful/screen.lua
```diff
@@ -201,6 +203,28 @@ function screen.object.get_tiling_area(s)
     }
 end
 
+--- Take a screenshot of the physical screen.
+--
+-- Reading this (read only) property returns a screenshot of the physical
+-- (Xinerama) screen as a cairo surface.
+--
+-- @DOC_screen_content_EXAMPLE@
```
actionless, line 211: there is no example itself in `./tests/examples/awful/screen` as far as i see from this patch

actionless:
what i meant was to just copy the surface of default_visual in c, and do the cropping on lua side

but if it's fine by other reviewers - i'm not so strongly insisting on that one

poisson-aerohead:
@actionless I am certainly pretty new to the X API but the way I figured out how to get the full root window content was to pass the root window info from globalconf.screen to cairo_xcb_surface_create. Looking at it now, I don't really need the width and height variables, they call could just be in the function call arguments. I'm not sure how to get the root window content any other way. Also, I am not cropping anything in this routine.

I have not looked into the tests directory yet. I can take a look and see how it works.

actionless:
yup, my bad - just misunderstood what's going on there

in that case just test/example remain and we good to go

poisson-aerohead:
I updated root.c to not have any variables related to the root window. I think it clarifies the function more.

How do you want to run tests? Should we actually be verifying the output file or just make sure that no error occurs?

poisson-aerohead:
> yup, my bad - just misunderstood what's going on there

Yeah, as I said, the root screenshot is what is actually necessary (it is the root after all). Everything else can theoretically be a crop of that in Lua. @psychon has raised concerns about doing the client this way regarding clients not currently visible. If the current state of affairs feels unsatisfying to you, I would say to look at it this way. For anything that is a real X (xcb) window we get the content through proper C calls to the X API. This allows "taking screenshots of invisible windows" and other cool things. For anything that is a crop we do the cropping in Lua. A physical (Xinerama) screen is not a window. Neither the X server nor xcb have any idea what it is. It is just a crop of the main root.

actionless:
yup, to compare, that's why i suggested to screenshot area without text

poisson-aerohead:
@actionless OK in keeping with your comment about text, I am not really a graphics expert so I am not sure we can rely on the screenshot making the exact same png file _to the bit_ (assuming the test is to compare the output file to a known standard using cmp or diff). Can we maybe rely on this if we take a screenshot of all black or all white?

actionless:
i was proposing to make a let's say white `wibox` with black `container.background` and some margin:
2021-10-01--1633115227_90x91_scrot

and just compare uncompressed bitmaps

i'm not sure if cairo have any methods to help with it, or you'll need to walk 2 bitmaps pixel by pixel until one of them differs

poisson-aerohead:
Hey, I am pretty close. It looks like I need to use [cairo_image_surface_get_data](https://cairographics.org/manual/cairo-Image-Surfaces.html#cairo-image-surface-get-data). Is this an lgi function? Is there somewhere I can get a plain listing of everything in lgi? I have been using [the repo](https://github.com/pavouk/lgi) as I work to try to figure it out.

Elv13:
> Hey, I am pretty close. It looks like I need to use cairo_image_surface_get_data. Is this an lgi function? Is there somewhere I can get a plain listing of everything in lgi? I have been using the repo as I work to try to figure lgi out.

Give me a few minutes. I will make you a skeleton. I tried `get_data` and while you can "see" it in LGI, you can't get the colors. GDK, however, has a pretty good API for it. I am making a prototype right now

poisson-aerohead:
OK thanks. What is the return value of get_data?

Elv13:
> OK thanks. What is the return value of get_data?

Looks like an opaque `userdata` blob from my 5 minute test.

poisson-aerohead:
Makes sense. The real function returns a char pointer, I'm guessing the bmp array.

Elv13:
Still messy, a lot of nice TODOs, but "works for me". Call it `tests/test-screenshot.lua`

Run using:

```sh
TEST_PAUSE_ON_ERRORS=1 TESTS_SCREEN_SIZE=1280x800 ../tests/run.sh  ../tests/test-screenshot.lua
```

```lua
-- This test suite tests the various screenshot related APIs.
--
-- Credit: https://www.reddit.com/r/awesomewm/comments/i6nf7z/need_help_writing_a_color_picker_widget_using_lgi/
--
local lgi = require 'lgi'
local cairo = lgi.cairo
local gdk = lgi.Gdk
local gsurface = require("gears.surface")

--TODO make a dummy blue client see ../tests/test-titlebar.lua for
-- how to make a GTK window using LGI

-- Split in the screen into 2 distict screens.
screen[1]:split()

-- Add a green wibox on screen 1 at (100, 100)
wibox {
    bg      = "#00ff00",
    visible = true,
    width   = 100,
    height  = 100,
    x       = 100,
    y       = 100,
}

local corners = {}

-- Add red squares at 1 pixel away from each screen corners.
for s in screen do
    -- Top left
    table.insert(corners, {
        x = s.geometry.x + 1,
        y = s.geometry.y + 1,
    })

    -- Top right
    table.insert(corners, {
        x = s.geometry.x + s.geometry.width - 11,
        y = s.geometry.y + 1,
    })

    --TODO bottom left and bottom right
end

for _, corner in ipairs(corners) do
    corner.bg = "#ff0000"
    corner.visible = true
    corner.width = 10
    corner.height = 10

    wibox(corner)
end

local function copy_to_image_surface(content, w, h)
    local sur = gsurface(content)

    local img = cairo.ImageSurface(cairo.Format.RGB24, w, h)
    local cr = cairo.Context(img)

    cr:set_source_surface(sur)
    cr:paint()
    img:finish()

    return img
end

local function get_pixel(img, x, y)
    local bytes = gdk.pixbuf_get_from_surface(img, x, y, 1, 1):get_pixels()
    return "#" .. bytes:gsub('.', function(c) return ('%02x'):format(c:byte()) end)
end

local steps = {}

-- Check the whole root window.
    local img = copy_to_image_surface(root.content(), root.size())

    assert(get_pixel(img, 100, 100) == "#00ff00")
    assert(get_pixel(img, 199, 199) == "#00ff00")
    assert(get_pixel(img, 201, 201) ~= "#00ff00")

    --TODO check 1,1 is red and for each corner of the root

    return true
end)

-- Check the screen.content
    --TODO
    return true
end)

-- Check the client.content
    --TODO
    return true
end)

require("_runner").run_steps(steps)

```

poisson-aerohead:
> ```lua
> x = s.geometry.x + s.geometry.width - 11
> ```

Why is "top right " at 11 pixels from the edge? Did you mean 1? Looks good otherwise.

Elv13:
> Why is "top right " at 11 pixels from the edge? Did you mean 1? Looks good otherwise.

I think we should/need to place them at least 1px away from the "real" corner, because otherwise if your `screen.object.get_content` has an "off by one" bug, the 2 screen corners would "merge" into a since 10x20 red rectangle and the test could not detect the off-by-one

poisson-aerohead:
@actionless @Elv13 I have committed the preliminary test suite. I think I addressed everything. I had to specify Lua 5.3 for it to work for me (I have Lua 5.4 but 5.3 comes with awesome and that is the only thing that has lgi for me). I may need your help ironing out how to get that part to work reliably.

poisson-aerohead:
Should I bring my repo up to date, or will that clutter this PR with a ton of changed files. I was able to successfully merge on my home machine, but I don't want to push to github if it is going to cause problems here. The root.c conflict is because there seems to have been a PR to change root.wallpaper to root._wallpaper.

Elv13:
Looks very good, thanks!

As for the rebase, you can

```sh
git remote add real <https://real/url>
git fetch real
git rebase -i real/master
```

Then git push `--force`. This should fix the conflicts and allow the CI to run. Lua 5.2 will still fail, but if everything else passes,  I will gladly approve the PR!

poisson-aerohead:
> As for the rebase, you can

Not to be the n00b, but by real/url you mean github.com/awesomeWM/awesome, right? So fetch the upstream and call it 'real' (or whatever), and then rebase my master to it? Won't this, then, get rid or my changes in root.c and screen.lua?

Elv13:
No worries, git is a complex beast, that's also a very valuable skill. So enjoy the journey ;)

> by real/url you mean github.com/awesomeWM/awesome, right?

I edited my previous post

> Won't this, then, get rid or my changes in root.c and screen.lua?

No. If you think of it as a tree of commit, it will just change the "root" commit of the pull request and "cherry pick" each of your commits on top of it. Cherry picking commit means making a copy of a commit content on top of another commit. `-i` in `git rebase` give you a "choice" in a text file. You can move commit around, remove them and/or merge multiple commits into one. In your case, you don't need anything. Just quit your editor and you are good to go.

Also, git commits are never lost unless you run `git gc` with dangling commit. So your work is save ;). If you want to be really, really, you can make a temporary branch.

poisson-aerohead:
> git is a complex beast, that's also a very valuable skill

I know, its part of the reason I wanted to push some code to software I use. You can learn a lot 'lone hacking' at home but distributed development is a gatekeeper skill.

I noticed the -i after I replied. In the past few days I had to rebase once to get rid of superfluous commits so I am aware of the -i option and the text file. I'll give this a try.

poisson-aerohead:
I think I got it. Let me know if its OK.

Elv13:
> I think I got it. Let me know if its OK.

Not quite, the conflict is still there. Do you want me to rebase it?

poisson-aerohead:
> Not quite, the conflict is still there. Do you want me to rebase it?

Haha OK. I don't know what happend then. I did rebase it.

Elv13:
whoops. Because you used the `master` branch, it didn't like my doing the rebase and auto-closed this. Next time, you can use `git checkout -b some_name` before sending a PR to avoid this. I will make a new PR

Elv13:
New copy here https://github.com/awesomeWM/awesome/pull/3451

PR #3460, opened 2021-10-09
test(screenshots, spawn): require exact major gtk/gdk versions

actionless:


actionless:
it means major exact 3, minor at least 0

PR #3461, opened 2021-10-12
User Facing Screenshot API

poisson-aerohead:
This commit begins the development of a more appropriate user facing
screenshot API. It extends a prior commit which extended the lower level
content API, which had been a property of the client object but is now
available as a property of the screen object and a method of the root
object.

This commit creates a new screenshot module for the awful module. The
public functions include root(), screen(), client(), snipper(), and
snip(). These take root window, screen, and client window screenshots,
launch an interactive snip tool for cropped screenshots, and take a
cropped screenshot of a geometry passed by argument, respectively. The
init() function is also available for configuration. Using this library
is more appropriate for the average rc.lua.

Since the API is new, this commit does not include any changes to
rc.lua. The developers can modify rc.lua when there is sufficient
confidence in API stability and robustness.

lib/awful/init.lua is modified so that the awful module includes the new
lib/awful/screenshot.lua submodule.

Signed off: Brian Sobulefsky <brian.sobulefsky@protonmail.com>


poisson-aerohead:
Hello again everyone. I wanted to post this so the project maintainers could get eyes on it and give feedback. As noted in the commit message, I did not modify rc.lua. Please see a sample binding below. At this time, I did not add test cases because it seems they cause race conditions and I am expecting churn after people examine this. I will add new test cases to tests/test-screenshot.lua before the commit is done.

```
awful.screenshot.init('/home/user/Images/Screenshots', 'Screenshot-')

        -- Root window screen shot
        awful.key({ modkey }, "Print",
                function()
                    awful.screenshot.root()
                end,
                {description = "Print Screen - Root Window", group = "awesome"}),
        -- Physical screen screen shot
        awful.key({ modkey, "Mod1" }, "Print",
                function()
                    awful.screenshot.screen()
                end,
                {description = "Print Screen - Physical Screen", group = "awesome"}),
        -- Client window screen shot
        awful.key({ modkey, "Control" }, "Print",
                function()
                    awful.screenshot.client()
                end,
                {description = "Print Screen - Client Window", group = "awesome"}),
        -- Snip tool screen shot
        awful.key({ modkey, "Shift" }, "Print",
                function()
                    awful.screenshot.snipper()
                end,
                {description = "Snip Screen", group = "awesome"})
)
```

Looking forward to hearing from you all.

Elv13:
Hello, thanks for this.

Lets get started with some of the mundane "things to learn when contributing to AwesomeWM". That's pretty much "normal" for when contributing to projects. Your previous contribution was pretty low level, so needed less of that. None of what's below is a criticism, it's just explaining many undocumented conventions and API design guidelines we use for high level APIs. This is the first pass. Once this settles down, there will be some testing and code examples work to do, but that's for another day.

First, there is a tool called `luacheck` which the CI run. Normally, it prints tons of warnings, but you only get a couple, which is great:

> luacheck ../lib/awful/screenshot.lua
> Checking ../lib/awful/screenshot.lua              2 warnings
>
>    ../lib/awful/screenshot.lua:185:5: setting non-standard global variable 'initilialized'
>    ../lib/awful/screenshot.lua:432:9: unused variable 'date_time'
>
> Total: 2 warnings / 0 errors in 1 file

Next, the documentation needs 3 `---` at the top of the comment block to be interpreted. If I add this to your changes, I get:

image

As you can see, there is still a couple trivial fixes. The format for the parameter is `@tparam[opt=something] type_name param_name description .`. In this case, `type_name` is missing, which confuses `ldoc` and it renders the same description for everything (because every parameter ends up being called "The"). You can install `ldoc`, clear your cmake work directory and run cmake again. It should detect it and start rendering the doc. You can open it in your web browser as a local file.

So that's for the trivial "how to get the CI to pass and the doc to render" things. Now, let's move into the code a little bit.

### awful.screenshot.init

Our objects constructors are (generally) directly the module name. In Lua, you do that using meta tables. Here's how:

```lua
local gobject = require("gears.object")

local module = {}

--- A property!.
-- @property my_property
-- @tparam[opt=default] string my_property

function module:get_my_property()
    -- You don't need this accessor of the default is
    -- self._privaye.property_name
    return self._private.my_property
end

function module:set_my_property(value)
    -- This accessor can contain business logic and have side effects
    self._private.my_property = value

    self:emit_signal("property::value", value)
end

--- This is a method.
-- @method my_method
function module:my_method()
    -- do something
end

local function new(_, args)
    args = args or {}                                                                                                                                
    local ret = gobject {
        enable_auto_signals = true,
        enable_properties   = true,
    }

    ret._private = {
        my_property = "default_value"
    }

    -- This adds all the method and properties
    gtable.crush(ret, module, true)

    -- This apply all the constructor arguments.
    gtable.crush(ret, args, false)

    return ret
end

return setmetatable(module, { __call = new })

```

So, `screenshot.init` should create a "real" object and be renamed to use the convention demonstrated above.

### The API in general.

For the API, right now, you have some static functions. I would like to see this moved to a more declarative/OOP style API. This means using a main objects with some properties and method.

With static function, it's hard to scale the API on the long term. For example, creating some composition like "take a screenshot of object foo, scale it, crop it, save it". With a more object oriented API, this comes pretty naturally. It is also easier to extend with new features, since you don't have to add endless parameters to each static functions. For example, the API could look like:

```lua

local ss = awful.screenshot {
    client = client.focus, -- automatically use client.content, by default, use root.content
    rectangle = {
        x = 0,
        y = 0,
        width = 100,
        height = 100
    }
}

ss:save("/tmp/image.png")

-- Because why not!
mywibar.widget.some_layout.add(ss.widget)

```

It also has the advantages of being a memento. Right now, `c.content` changes "from under", you render the same surface twice and it looks different. This is pretty unexpected if you store the surface in a variable. `awful.screenshot` could/should hold a copy of the surface within the object and only explicitly refresh it. That way, storing screenshots as variables now works.

poisson-aerohead:
Hey, before I go too crazy, can we get people to give some input on what an "object oriented" library should look like? My initial submission centered around taking the various kinds of screenshots, saving them to files, and returning the path to the newly saved file. The various kinds included root window, physical screen, client window, interactive snip tool, and programmatically defined crop. The directory and base filename of the saved screenshot could be specified, but essentially defaulted to something that would likely work. Otherwise, it could be "initialized" once and called over and over again with no further arguments.

I can change this to an "object oriented" design, but is there anything else people think makes sense?

poisson-aerohead:
Hi everyone, sorry to have dropped off on this one. I have been busy with some other things. I corrected the documentation generation syntax. The issue was that the correct format seems to be `@tparam type name DescriptionString` and I was missing the name. I rebased this PR to the current state of the master, so a month's worth of changes should be behind us (including fixing the screenshot test script!).

Please provide a little direction as to what the OO system should look like. I still think there is a disconnect between thinking of "screenshot" as the library for taking the screenshots (how I am using the term) or as an actual image. The way I am looking at it, we need some interface for the user to request a screenshot be taken and how (the root window, the client, a particular screen, a snip tool, a defined crop of the screen). If the desire is for that action to create a new object (that can be saved, scaled, put in a widget, whatever), then I understand that. However, for the screenshot library _itself_ to return objects is a little weird to me (it can certainly be done that way). For example, beautiful uses "static" methods like beautiful.init(), and gears holds all sorts of static methods like gears.wallpaper.fit(). I personally think it is natural for there to be a similar screenshot library that can be called upon to perform actions, such as awful.screenshot.root() to take a full screenshot or awful.screenshot.client() to get the active client content, but I am open to a different direction.

If you are looking for the library itself to be an object, please discuss with me the class layout you are visualizing. Otherwise, if the desire if for the screenshot methods to return objects with properties and methods, such as `ss = awful.screenshot.screen(); ss:scale(geom); ss:save("/path/to/file.png");` then I view that as a different task and we should define what sort of object that is going to be.

Input from some of the others on here welcome, @actionless @Aire-One @psychon

actionless:
i think the next step would be to improve the coverage level to pass the CI check:

> 12.16% of diff hit (target 90.65%)

poisson-aerohead:
I'm not sure how that part works.

actionless:
i think you can start from the existing tests for low-level screenshot api as a starting point

actionless:
or mb adding `examples` would be better, but here i myself not sure how it would be better to implement for examples

sclu1034:
I think the disconnect might come from the fact that, while pure static libraries like `gears` exist, most of our modules are actually more like _classes_.
`awful.screenshot` would, first and foremost, be a table with a few functions. But through metatables, it would also be callable, and `awful.screenshot()` would return an instance/object of the class.
Many other modules work just like that, and their docs can have both a  "static functions" and an "object methods" section to reflect that.

In this case, I agree with Elv that a more OOP-like API would be the better fit long-term. I.e. `awful.screenshot` would be a class, and most functionality would be handled through methods on objects that `awful.screenshot()` creates.

But that doesn't mean that it can't also cover the use cases you describe.
Building off of Elv's proposed API:

- To make the use case of "take a screenshot and immediately save it" easier, you could add module-level defaults:

```lua
awful.screenshot.set_default_path(os.getenv("HOME") .. "/Pictures/Screenshots")
awful.screenshot.set_name_template("%Y-%m-%d.png") -- see GLib.DateTime.format

local shot = awful.screenshot()
shot:save() -- Build file path from defaults
shot:save("/tmp/just_a_test") -- Override defaults. If `just_a_test` is a directory, use name template to store inside
```

- You can still create static functions like `awful.screenshot.client()` as thin wrappers around the generic constructor
- Use the "builder" pattern, where each method on the object returns the altered instance to allow chaining

```lua
local shot = awful.screenshot()
local small_shot = shot:scale(0.5)
local fancy_shot = shot
	:draw_rectangle(...)
	:draw_circle(...)
	:text(...)
```

---

If you combine the first two, your use case above becomes

```lua
-- Once at the start
awful.screenshot.set_default_path(os.getenv("HOME") .. "/Pictures/Screenshots")
awful.screenshot.set_name_template("%Y-%m-%d.png")

-- In a hotkey
awful.screenshot.client():scale(0.5):save()
```

It's a few more characters to write in the hotkey hander, but more expressive, without hidden defaults/constants and allows for a much more powerful and extensible API, as Elv already described.

poisson-aerohead:
@actionless sorry I misunderstood. I didn't realize that the CI metric was what the tests covered. Can I fix that by just adding tests that use the new methods in the tests/*.lua files?

@sclu1034 I agree with your line of thinking and would point out that you are naturally thinking of the object library itself, awful.screenshot, as something distinct from the returned screenshot, which you are calling shot. This is what I noticed too, we really have two (albeit related) object hierarchies here.

actionless:
@poisson-aerohead yup, either tests or examples - both counted for coverage

poisson-aerohead:
Hi everyone. Sorry for the long hiatus on this one, but as you know, I did have the awful.screenshot library working as I wanted on my own machine, and trying to rewrite this in the style you requested took a bit of a backseat. Some quick notes. As we discuss on this thread, the awful.screenshot library distinguishes between the library itself and the returned screenshot object. So there is a generic constructor, `new(args)`, callable as `ss = awful.screenshot(args)` that returns a "screenshot object". There are also some wrapper constructors, callable as `ss = awful.screenshot.root()`, etc.

The returned object has methods like ss:save(), that can take optional arguments. The module should work out of the box with the following rc.lua:

`
          -- Root window screen shot
          awful.key({ modkey }, "Print",
                  function()
                      local ss = awful.screenshot.root()
                      ss:save()
                  end,
                  {description = "Print Screen - Root Window", group = "awesome"}),
          -- Physical screen screen shot
          awful.key({ modkey, "Mod1" }, "Print",
                  function()
                      local ss = awful.screenshot.screen()
                      ss:save()
                  end,
                  {description = "Print Screen - Physical Screen", group = "awesome"}),
          -- Client window screen shot
          awful.key({ modkey, "Control" }, "Print",
                  function()
                      local ss = awful.screenshot.client()
                      ss:save()
                  end,
                  {description = "Print Screen - Client Window", group = "awesome"}),
          -- Snip tool screen shot
          awful.key({ modkey, "Shift" }, "Print",
                  function()
                      awful.screenshot.snipper()
                  end,
                  {description = "Snip Screen", group = "awesome"})
  )
`

You can optionally configure it first with
`
awful.screenshot.set_defaults({directory = '/path/to/your/screenshots',
                                 prefix = 'MyScreenshotPrefix-'
                              })
`
if desired. I'm sure there will be some issues with getting the tests to pass, which I am a bit underinformed on. Also, the documentation could use a bit of expansion. but it should at least have a basic description for everything.

poisson-aerohead:
Hey everyone. I don't know what's going on with the failure. On my machine, I can just call it with no configuration. What's weird is that the default prefix is hard coded to "Screenshot-" so I can't imagine any way you could end up with it being nil. I need help with running the test suite.

actionless:
idk, i was trying to look into your code but you have single-letter var names so i gave up, but good luck fixing it :)

poisson-aerohead:
@actionless most of my var names are pretty descriptive. I think there might be a few temporary variables where I use that. I got all the tests to successfully run. I currently am doing this by hard coding a path in my test script like ``local ss = awful.screenshot.root({directory = "/path/to/Images"}). I would think that this is brittle. I had to solve one problem but now that I did I will figure out why reading the environment for HOME is failing.

poisson-aerohead:
OK, problem solved, well, at least identified. It seems that $HOME in the Xephyr environment is '/dev/null/'. I can understand why this would happen. I don't know how to fix it to run on the general machine, so I need some input from you all.

Here is the thing, my initialization code is going to try set the 'directory' property to something writable. I suppose I can try to change the initialization logic, but it seems like a good idea to do it the way I did and not weaken the class module to accommodate the test suite. So, what is a good, robust way for the test script to find a writable directory? It does not actually need to write there (the test suite is currently not saving any files). But the init logic will check that the directory is writable.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,884 @@
...
+-- screenshot API.
+--
+-- This supports using a different directory or filename prefix for a particular
+-- use by passing the optional arguments. In the case of an initalized object
+-- with  no arguments passed, the stored configuration parameters are quickly
+-- returned. This also technically allows the user to take a screenshot in an
+-- unitialized state and without passing any arguments.
+local function configure_path(directory, prefix)
+
+  local d, p
```
actionless, line 181: i was talking of these specifically

poisson-aerohead: OK, I'll make them `dir, prfx` or something.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,884 @@
...
+-- The module to be returned
+local module = { mt = {} }
+-- The screenshow object created by a call to the screenshow module
+local screenshot = {}
+-- The various methods of taking a screenshow (root window, client, etc).
+local screenshot_methods = {}
+
+-- Configuration data
+local ss_dir = nil
+local ss_prfx = nil
```
actionless, line 30: but those would be nice also to make more maintainable

if you have troubles typing long names - you can just press Ctrl+xn in vim to auto-complete it

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,884 @@
...
+
+  local args = (type(args) == "table" and args) or {}
+  local ss = gears.object({
+    enable_auto_signals = true,
+    enable_properties = true
+  })
+
+  dir, prfx = configure_path(args.directory, args.prefix)
+
+  local fclr = nil
```
actionless, line 828: frame_color

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,884 @@
...
+local module = { mt = {} }
+-- The screenshow object created by a call to the screenshow module
+local screenshot = {}
+-- The various methods of taking a screenshow (root window, client, etc).
+local screenshot_methods = {}
+
+-- Configuration data
+local ss_dir = nil
+local ss_prfx = nil
+local frame_color = gears.color("#000000")
```
actionless, line 31: module.frame_color = ....

actionless:
leave short var names for either logically reasonable (like x, y coordinates, etc) or ones conventional for awesome API (c, s, cr, etc)

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,884 @@
...
+    elseif string.find(directory, "^[^/]") then
+      directory = string.gsub(os.getenv("HOME"), "/*$", "/") .. directory
+    end
+
+    -- Assure that we return exactly one trailing slash
+    directory = string.gsub(directory, '/*$', '/')
+
+    -- If we use single quotes, we only need to deal with single quotes - (I
+    -- promise that's meaningful if you think about it from a bash perspective)
+    local dir = string.gsub(directory, "'", "'\\'\\\\\\'\\''")
```
actionless, line 82: what the difference between dir and directory??

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,884 @@
...
+end
+
+-- Internal routine to verify that a filepath is valid.
+local function check_filepath(filepath)
+
+  -- This module is forcing png for now. In the event of adding more
+  -- options, this function is basically unchanged, except for trying
+  -- to match a regex for each supported format (e.g. (png|jpe?g|gif|bmp))
+  -- NOTE: we should maybe make this case insensitive too?
+  local fname_start, fname_end = string.find(filepath,'/[^/]+%.png$')
```
actionless, line 122: is it start of file *name* or file path?

actionless:
also please double-check what none of your helper functions is already implemented in `gears` standard library (particularly gears.filesystem)

lib/awful/screenshot.lua
poisson-aerohead: Yeah its OK, I don't mind. If you look, most of the names are descriptive. It is usually just temporaries in helper functions where I don't do that, but I'll clean them up for the next push.

lib/awful/screenshot.lua
poisson-aerohead: So that one is actually for a reason but I can do a better name. That is the escaped string to be passed through bash.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,884 @@
...
+  return self._private.prefix
+end
+
+--- Set screenshot prefix property
+--
+-- @property prefix
+-- @tparam string prefix The prefix prepended to screenshot files names.
+function screenshot:set_prefix(prefix)
+  if type(prefix) == "string" then
+    local prfx = check_prefix(prefix)
```
actionless, line 650: and again, `prfx = prefix` - that's very confusing to understand without unwinding every code branch

lib/awful/screenshot.lua
poisson-aerohead: Name. The regex is finding the last '/' through a path ending in '.png', so that parses a valid file name from a full path.

lib/awful/screenshot.lua
poisson-aerohead: That is just avoiding a collision between the argument name and the local variable which will be stored as a property if it is valid.

lib/awful/screenshot.lua
poisson-aerohead: The constructor I intentionally did this with everything to avoid name collision. I thought it would be easier to follow. Maybe the constructor local variables should all be named as '_frame_color' ?

lib/awful/screenshot.lua
actionless: try to name the var more clear according to their logical roles in the code, not just randomly duplicate the same name and "avoid as collision"

lib/awful/screenshot.lua
actionless: that should be clear from var name - not from your comment

it makes no sense giving basically the same name to 2 variables which have different logical meaning

lib/awful/screenshot.lua
actionless: or mb try using russian 'a' and just name them the same? ;)

lib/awful/screenshot.lua
poisson-aerohead: I have updated these three tentatively (this is how they function if no argument provides a value)
```
-- Configuration data
local global_screenshot_dir = nil
local global_screenshot_prefix = nil
local global_frame_color = gears.color("#000000")
local initialized = nil
```

actionless:
>  It seems that $HOME in the Xephyr environment is '/dev/null/'

```console
tests/run.sh
16:export HOME=/dev/null
199:        XDG_CONFIG_HOME="$build_dir"
```

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,884 @@
...
+
+-- Internal function to get the default screenshot directory
+-- 
+-- Can be expanded to read X environment variables or configuration files.
+local function get_default_dir()
+  -- This can be expanded
+  local d = os.getenv("HOME")
+  
+  if d then
+    d = string.gsub(d, '/*$', '/') .. 'Images/'
```
actionless, line 43: read XDG spec about images' dir

poisson-aerohead: I found [this on Archlinux](https://wiki.archlinux.org/title/XDG_user_directories) on the user directories. How do you want this handled for the screenshot module? It looks like it would be in '~/.config/user-dirs.dirs'. I say in the comment we can expand it to include parsing such a file. Is there already a gears routine to get this?

actionless: see our gears.filesystem please for similar functions (or mb even better add them there, not here)

actionless: why do you trying to find XDG spec on arch linux website? here is not only explained everything, but even given example (very similar to our gears, again):

https://www.freedesktop.org/wiki/Software/xdg-user-dirs/

actionless: i believe the most practical way would be to parse that file and replace $HOME to currently detected home dir

this is still not very robust - but would avoid making code asynchronous for external call of bash

poisson-aerohead:
@actionless I solved the test failure with no hard coding with something like this:

```
local fake_screenshot_dir = os.getenv("PWD")
...
local ss = awful.screenshot.root({directory = fake_screenshot_dir})
...
local ss = awful.screenshot.screen({screen = s, directory = fake_screenshot_dir})
...
local ss = awful.screenshot.client({client = c, directory = fake_screenshot_dir})
```
This should work on anyone's home computer. My question is will this be OK on the GitHub Jobs? Does the virtual environment GitHub gives you allow you to test writing files on a virtual drive?

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,884 @@
...
+    if os.execute("bash -c \"if [ -d \\\"" .. d .. "\\\" -a -w \\\"" .. d ..
```
actionless, line 44: ..speaking of which, - this should go - again, see gears.filesystem for helper function for working with filesystem

actionless:
> GitHub gives you allow you to test writing files on a virtual drive?

better tmp dir - otherwise it would create mess when running it locally

lib/awful/screenshot.lua
poisson-aerohead: I'll look at gears.filesystem and see what is available.

poisson-aerohead:
> > GitHub gives you allow you to test writing files on a virtual drive?
> 
> better tmp dir - otherwise it would create mess when running it locally

I'm sorry I don't understand this reply. What I'm asking is how to reliably get a path to a writable directory. I don't actually need to write there. I can do this on a local machine but how would one make sure it succeeds when GitHub runs the test suite automatically?

lib/awful/screenshot.lua
actionless: using module properties would be better, but now it looks at least readable 👍

actionless:
`/tmp`

poisson-aerohead:
> `/tmp`

OK. `local fake_screenshot_dir = "/tmp"` worked locally (obviously), so I'll take you at your word that this is acceptable on GitHub too.

I will address your other points before pushing again. Are there any functions in gears.filesystem that I should really take a look at?

actionless:
if you not sure you can just push and try it yourself on GH actions - the code on merge will be squashed anyway, so extra commits wouldn't 'spam' the commit log

poisson-aerohead:
> if you not sure you can just push and try it yourself on GH actions - the code on merge will be squashed anyway, so extra commits wouldn't 'spam' the commit log

OK. I was a little concerned about spamming you all.

actionless:
np :D good luck fixing it up - will check later on it

poisson-aerohead:
Hey, is it me or are there one `dir_readable` and `is_dir` routines in `gears.filesystem`? Your file functions only include `file_readable` and `file_executable` too. It seems to me that 'writable' is a good member for these utility functions, and they should be included there and not in the `awful.screenshot` library.

poisson-aerohead:
Hi everyone. I pushed the updates just to verify that we pass the checks. It looks like we pass everything except 'fixed-lgi-lua5.2' and I am pretty sure that is unrelated to me. This commit has the updated 'tests/test-screenshot.lua' file with the changes needed to pass the tests. It also has a lot of the stylistic issues addressed from the discussion above.

I still need to swap out some filesystem functionality to the provided gears library. What this module needs is to check whether a directory is writable and to determine the user's default image directory if none is provided. The first looks like it would belong in `gears.filesystem` but the check for writability is not included there. The second looks like it might also belong there, as other XDG default directories can be queried from that module, but the Images directory is not there.

Can I get some advice on how the project would like these two issues addressed?

Again, to use this library, include the following in your `rc.lua`:
```
awful.screenshot.set_defaults({directory = '/path/to/my/screenshots',
                                 prefix = 'Screenshot-'
                              })

          -- Root window screen shot
          awful.key({ modkey }, "Print",
                  function()
                      local ss = awful.screenshot.root()
                      ss:save()
                  end,
                  {description = "Print Screen - Root Window", group = "awesome"}),
          -- Physical screen screen shot
          awful.key({ modkey, "Mod1" }, "Print",
                  function()
                      local ss = awful.screenshot.screen()
                      ss:save()
                  end,
                  {description = "Print Screen - Physical Screen", group = "awesome"}),
          -- Client window screen shot
          awful.key({ modkey, "Control" }, "Print",
                  function()
                      local ss = awful.screenshot.client()
                      ss:save()
                  end,
                  {description = "Print Screen - Client Window", group = "awesome"}),
          -- Snip tool screen shot
          awful.key({ modkey, "Shift" }, "Print",
                  function()
                      awful.screenshot.snipper()
                  end,
                  {description = "Snip Screen", group = "awesome"})
  )
```
As a side note, I have verified correct expected behavior with:
```
awful.screenshot.set_defaults({directory = '/home/username/Images/My "F@vorite" $creenshot ' .. "'Directory'!",
                                 prefix = '$creensh0tz-'
                              })
```
Look forward to hearing from you.

Elv13 (changes requested):
Hello,

Thanks for the work. I did a mini review since I don't have a lot of time for AwesomeWM right now. I mostly focused on the API rendering issues.

Beside, this will be a nice addition once it matures a bit more. 51% test coverage isn't quite enough. We are usually looking at something greater than the current threshold, which is at 91%.

The documentation will also need some `tests/example/awful/screenshot` code examples. These files are executed when AwesomeWM is built. You can add them to the doc using `@DOC_..._EXAMPLE@` macros where `...` is the path separated by underscores (like `@DOC_awful_screenshot_myexamplename_EXAMPLE@`).

For the API, it will probably need a minor round of changes after that's done. Right now, one thing that stands out of moving to the path builder code to `gears` and using normal strings path in screenshot API (the result of of the path builder). This is more consistent with other APIs and allows everything, everywhere, to use these builders rather than having to have special code built-in the modules themselves.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,881 @@
...
+  local mg_callback = mk_mg_callback(ss)
+  capi.mousegrabber.run(mg_callback, "crosshair")
+
+  return true
+
+end
+
+-- Internal function executed when a snip screenshow (a defined geometry) is
+-- taken.
+function snip_screenshot(ss)
```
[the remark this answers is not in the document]
poisson-aerohead: Yeah, I'll add any missing `local`s

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,881 @@
...
+
+  local root_w, root_h
+  local root_intrsct
+  local snip_surf
+
+  root_w, root_h = root.size()
+
+  if not(ss.geometry and
```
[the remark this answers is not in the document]
poisson-aerohead: I can look into that function.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,881 @@
...
+-- @property surface
+function screenshot:get_surface()
+  return self._private.surface
+end
+
+-- Methods for the screenshot object returned from taking a screenshot.
+
+--- Set the filepath to save a screenshot
+--
+-- @function awful.screenshot:filepath_builder
```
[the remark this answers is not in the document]
poisson-aerohead: This method sets the the directory, prefix, and filepath properties based on checking a hierarchy of: the filepath argument, the directory and prefix arguments, the self.filepath property, the self.directory and self.prefix properties, the module default directory and prefix (in that order). This seems pretty specialized, so I'm not sure how you would want that moved to gears.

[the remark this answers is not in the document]
poisson-aerohead: Regarding the `@method`, I still don't really know the available syntax for your documentation.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,881 @@
...
+
+--- Set screenshot filepath
+--
+-- @property filepath
+-- @tparam[opt] string fp The full path to the filepath
+function screenshot:set_filepath(fp)
+  self:filepath_builder({filepath = fp})
+end
+
+--- Get screenshot method name
```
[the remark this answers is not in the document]
poisson-aerohead: OK, as I said above, I don't really know the available syntax for your documentation.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,881 @@
...
+function screenshot:set_directory(directory)
+  if type(directory) == "string" then
+    local checked_dir = check_directory(directory)
+    if checked_dir then
+      self._private.directory = checked_dir
+    end
+  end
+end
+
+--- Get screenshot prefix property
```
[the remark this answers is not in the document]
poisson-aerohead: OK, same comment.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,881 @@
...
+  return module(args)
+end
+
+--- Take a cropped screenshot of a defined geometry
+--
+-- This is a wrapper constructor for a snip screenshot (defined geometry). See
+-- the main constructor, new(), for details about the arguments.
+--
+-- @constructorfct awful.screenshot.snip
+-- @tparam[opt] table args Table of arguments to pass to the constructor
```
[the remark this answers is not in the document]
poisson-aerohead: OK I do define the args free form so I will move it to that syntax.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,881 @@
...
+    self._private.prefix = prefix       -- been checked.
+    self._private.filepath = filepath
+  end
+
+end
+
+--- Save screenshot
+--
+-- @function awful.screenshot:save
+-- @tparam[opt] table args Table with the filepath parameters
```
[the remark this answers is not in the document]
poisson-aerohead: I can do it that way too. It looked like you like to avoid having to deal with positional arguments and prefer the table instead. Remember that my path_builder takes both a directory and a prefix for the filename. Since I am working with table args, I also built in an explicitly passed full file path, which will be checked for validity (points to a valid writeable directory). The way I have it now, if the filepath argument is used, it will be used as is and does not append a date_time string.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,881 @@
...
+--   directory [string]: the path to the screenshot directory
+--   prefix [string]: the prefix to prepend to screenshot file names
+--   frame_color [gears color]: the color of the frame for a snipper tool
+--   method [string]: the method of screenshot to take (i.e. root window, etc).
+--   screen [index or screen]: the screen for a physical screen screenshot
+--   on_success_cb [function]: the callback to run on the screenshot taken with
+--                  a snipper tool
+--   geometry [gears geometry]: a geometry object for a snip screenshot
+--
+-- @constructorfct awful.screenshot.new
```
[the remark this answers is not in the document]
poisson-aerohead: OK

poisson-aerohead:
@Elv13 regarding test coverage, probably about 150 lines or more (of 880 so non trivial) are for the "snipper tool", which is an interactive mouse tool. Is there a way to test that? I have used it plenty of times myself, so I am confident that the idea is at least basically sound (there may always be a hidden edge case bug somewhere of course). I do not know how the test could emulate the user using the mouse though.

actionless:
regarding the tests - see for example tests/test-resize.lua which using a lot of mouse-resize emulation

lib/awful/screenshot.lua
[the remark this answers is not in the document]
actionless: for some basic idea just `grep` existing usages in already existing code

for deeper understanding you can check out https://stevedonovan.github.io/ldoc/manual/doc.md.html and `docs/config.ld`

poisson-aerohead:
Hey sorry for the delay, but I finally sat down to research the mouse emulation thing as you requested. I basically need to emulate the user dragging the mouse to add a test of the snipper tool. I found how to emulate the "drag" portion by updating the mouse coordinates using mouse.coords [here](https://awesomewm.org/apidoc/input_handling/mouse.html). I have not found a method for emulating the actual clicks of the button, either on the mouse entry linked previously or the [button](https://awesomewm.org/apidoc/input_handling/awful.button.html), which is about creating new button objects. Can someone just refer me to how the Awesome API allows me to use Lua to send mouse button 1?

Elv13:
The function  is [fake_input](https://awesomewm.org/apidoc/core_components/root.html#fake_input)

sclu1034:
You can see an example of a test involving mouse drag in #3533.

poisson-aerohead:
Hey, when I run a test from bash, I get the following error:

```
$ TEST_PAUSE_ON_ERRORS=1 TESTS_SCREEN_SIZE=1280x800 ../tests/run.sh  ../tests/test-screenshot.lua
...
Fontconfig error: No writable cache directories
...
```
I think this is new (I haven't worked on this in a while). Is there a known work around?

poisson-aerohead:
Hey, I may have figured that one out. I got more descriptive error messages when running `test-resize`:

```
$ TEST_PAUSE_ON_ERRORS=1 TESTS_SCREEN_SIZE=1280x800 ../tests/run.sh  ../tests/test-resize.lua 
awesome v4.3-1395-g4e05fbc7-dirty (Too long)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
awesome_log: /tmp/tmp.Yj50r3GBgf/_awesome_test.log
== Running ../tests/test-resize.lua ==
Fontconfig error: No writable cache directories
2022-08-22 18:05:46 W: awesome: a_glib_poll:477: Last main loop iteration took 0.104689 seconds! Increasing limit for this warning to that value.
_client lua: (command line):2: module 'lgi' not found:
_client         no field package.preload['lgi']
_client         no file '/usr/share/lua/5.4/lgi.lua'
_client         no file '/usr/share/lua/5.4/lgi/init.lua'
_client         no file '/usr/lib/lua/5.4/lgi.lua'
_client         no file '/usr/lib/lua/5.4/lgi/init.lua'
_client         no file './lgi.lua'
_client         no file './lgi/init.lua'
_client         no file '/usr/lib/lua/5.4/lgi.so'
_client         no file '/usr/lib/lua/5.4/loadall.so'
_client         no file './lgi.so'
_client stack traceback:
_client         [C]: in function 'require'
_client         (command line):2: in main chunk
_client         [C]: in ?
Error: timeout waiting for signal in step 1/45 (@20) (/path/to/awesome/tests/../tests/test-resize.lua:9).
Keeping awesome open...
Awesome was killed due to timeout after 30 seconds
```

So I obviously have lua53-lgi installed right now for awesome, and as you can my homebrew awesome is compiled against lua 5.3. So I am not sure why the test suite is looking for lua-lgi (which is for lua 5.4). Any thoughts?

poisson-aerohead:
OK, so I think I got it. An answer to the prior question would be interesting, but I did get a successful screenshot test without getting `test-resize` to run. I just looked at the source as an example. You will see it when I push. I basically separated all the mouse movements and presses into individual tests. I put a little delay in there to, especially when the callback needs to run.

poisson-aerohead:
Hey everyone. It has been a while, but looking back through this thread, I was at 51% coverage. Adding a snipper test (mouse tool for screenshot) and a snip test (programmatically defined geometry) got the coverage to 77%. This is still not the target 91%, but that is a big jump. Since these were obvious methods that were totally untested, I added these tests. To get to 91%, I'll use the codecov report.

It is worth noting that the snip test did flush out a bug. Since this method will probably rarely be used, the bug could have gone unnoticed for a while.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,870 @@
...
+
+-- Internal function used by the snipper mousegrabber to update the frame outline
+-- of the current state of the cropped screenshot
+--
+-- This function is largely a copy of awful.mouse.snap.show_placeholder(geo),
+-- so it is probably a good idea to create a utility routine to handle both use
+-- cases. It did not seem appropriate to make that change as a part of the pull
+-- request that was creating the screenshot API, as it would clutter and
+-- confuse the change.
+function show_frame(ss, geo)
```
Elv13, line 255: Please add `local ` in front (here and in a few other places)

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,870 @@
...
+end
+
+-- Internal function to be passed as the default callback upon completion of
+-- the mousgrabber for the snipper if the user does not pass one.
+function default_on_success_cb(ss)
+  ss:save()
+end
+
+-- Internal function exected when a root window screenshot is taken.
+function screenshot_methods.root(ss)
```
Elv13, line 393: please replace `screenshot_methods.root(ss)` with `screenshot_methods:root()`, then use `self` instead of `ss`. This is mostly because it's done like this elsewhere in the codebase. Please also change it in other places too.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,870 @@
...
+
+--- Function to initialize the screenshot library.
+--
+-- Currently only sets the screenshot directory, the filename, prefix and the
+-- snipper tool outline color. More initialization can be added as the API
+-- expands.
+-- @staticfct awful.screenshot.set_defaults
+-- @tparam[opt] table args Table passed with the configuration data.
+-- @treturn boolean true or false depending on success
+function module.set_defaults(args)
```
Elv13, line 499: I think this should either use raw varibles, a metatable or even nothing at all. Having a single initialization function for a module is something which ended up causing a lot of issues in the past (`awful.rules`, `awful.utils.shell`). For example, if 2  third-party modules generate screenshot for some reasons, they might end up in a situation where either neither initialize the module, then it misbehave or *both* initialize the module, and it also misbehave.

If possible, I would rather force the defaults to be passed to the screenshot builders every time than storing them. That way it's stateless. If users want to wrap it with their own defaults, they are free to do so.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,870 @@
...
+
+  local tmp
```
Elv13, line 501: Please use `local tmp = check_directory(args.directory)` rather than 3 lines to improve readability.

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,870 @@
...
+end
+
+--- Get screenshot filepath.
+--
+-- @property filepath
+function screenshot:get_filepath()
+  return self._private.filepath
+end
+
+--- Set screenshot filepath.
```
Elv13, line 657: If you have a `@property`, you don't need to document both the setter and getter. If you generate the doc, this wont render properly

Elv13 (changes requested):
Thanks for the update!

I left a few more comments, mostly minor (this is only a partial review, I need to get back into this PR when I get time). For learning the doc, I am working [on a linting bot](https://github.com/awesomeWM/awesome/pull/3676) that will eventually print warning about our documentation standards. I will also document it better. However it's still not merged, so sorry for what is probably a rather confusing learning curve. When this is closer to being ready, I can give you a hand and make the doc final tuning myself.

For the remaining tests, you can add some entries in `tests/examples/awful/screenshot/` and add them to the doc using macros [like this](https://github.com/awesomeWM/awesome/blob/master/lib/awful/wallpaper.lua#L434). Those examples are both good for the documentation and good for testing. I am not asking to make the screenshot "really" works" in the test. It's doable, but too complicated for this PR. Just having the code being displayed will be good enough. Copy paste a `test/examples/awful/wallpaper/` file into `tests/examples/awful/screenshot/`  and modify it. That's the most simple way to get started.

lib/awful/screenshot.lua
poisson-aerohead: @Elv13, just to keep the Lua scoping rules correct, I should do this for top level module functions but not functions that are elements of tables. So I would want:

```
local function show_frame(ss, geo)
```

but

```
function screenshow_methods.root(ss)
```

Is this right?

lib/awful/screenshot.lua
poisson-aerohead: These are internal functions to the library that actually take the screenshot. We have been going around in circles for almost a year now, and I am happy to design the library to whatever organization you want, but in the absence of that I had to make my own decisions. The high level way it currently works is:

``
local ss = awful.screenshot(args)
ss:save()
``

The first line calls `module.new(_, args)` as you described to me. The various parameters of the screenshot are controlled  with `args`. There are helper routines that remove the need to pass the method as an argument:

```
local ss = awful.screenshot.root(args)
ss:save()
```

These helper routines call `module()` which means it calls `new`. If no additional arguments need to be passed beyond the method:

```
local ss = awful.screenshot.root()
ss:save()
```

Works fine (this is what I do in my `rc.lua`). Please let me know if you want something different.

lib/awful/screenshot.lua
poisson-aerohead: OK, the only defaults right now are the directory (e.g. "/home/username/Images"), the prefix to the filename (e.g. "Screenshot-"), and the color used for the rectangle defining the region captured by the snipper tool.

All screenshot functions accept arguments that override the defaults. The purpose of setting the default was just to allow the user to customize what would happen when no argument is passed.

In effect, if we remove this function, we either need to make it an error to not pass arguments (I don't recommend this) or use whatever hard coded defaults are in the file. There currently are hard coded defaults right now. In fact, the library will work without ever calling the `set_defaults` function, i.e.

```
ss = awful.screenshot.root()
ss:save()
```

should work just fine (assuming it can find a directory, we need to talk about adding the ability to find the preferred directory). So I can take this out, but this basically allows a customization that the user can set up an environment to use custom defaults. This is similar to the `beautiful` library I guess. Of course, as you point out there is only one copy of those defaults and different areas of the configuration can override one another.

It is up to you, but fwiw I would want the ability to set a system default different that whatever is hard coded.

lib/awful/screenshot.lua
poisson-aerohead: I will make this change if we keep this function, per above.

lib/awful/screenshot.lua
poisson-aerohead: OK

poisson-aerohead:
@Elv13 I added another test and got my coverage from 77% to 86% (so close). However, this broke lua 5.1 and 5.2. I am not really sure why and I do not know how to test this.

I do have visibility into exactly what failed, I have a good error message in the lua 5.1 and 5.2 logs. I am not sure at this time why it failed for these versions, though.

lib/awful/screenshot.lua
sclu1034:
```lua
local function screenshot_methods.root(ss) end
```
would throw a syntax error. Just like the equivalent would

```lua
local screenshot_methods = {
  local root = function(ss) end
}
```

lib/awful/screenshot.lua
sclu1034: The functions you're talking about here are `module.root(args)` (and similar) from further down in the file. Elv is commenting on `screenshot_methods.root(ss)`, which would map to `ss:root()` in your snippets.

`ss` here looks very much like an instance of something (methods called on it, `_private` accessed). By convention, we use Lua's `:` shorthand for functions that take an object instance as first parameter `self`.

sclu1034:
> Unfortunately, the most direct way I can think of is to push versions here with debug messages and read them after the test suite runs. Sorry for clogging you inboxes.

Create a new branch off of your feature branch and a PR from that new branch to your own master branch. That PR will trigger GitHub Actions within the scope of your own repo.
Once you're done testing, squash the test commit into one fix and push that to the actual feature branch here.

poisson-aerohead:
@sclu1034 OK I apologize, that makes sense. It looks like I already fixed it, so if it happens again I will keep that workflow in mind.

poisson-aerohead:
OK everyone. Sorry about that but I don't know how to build different versions of awesome and now I know that if I made another branch of my own and opened a PR it would automatically run the actions. Anyway, we are back to passing everything except 'fixed-lgi-lua5.2' and our coverage is now at 86.5% (a little over 24 hours ago we were at 51%).

I'm not really a lua historical expert. In fact, the reason I learned lua at all was because I use awesomewm. So I am not sure about the change that was driving that bug in the old versions. I would hazard a guess it was determined an actual lua bug that they fixed in the 5.3 release. It has to do with gsub matching at the end of a string.

sclu1034:
> So I read the '_client' source and figured out how to change that:
> 
> ```
> $ LUA=lua5.3 TEST_PAUSE_ON_ERRORS=1 TESTS_SCREEN_SIZE=1280x800 ../tests/run.sh  ../tests/test-resize.lua 
> awesome v4.3-1395-g4e05fbc7-dirty (Too long)
>  • Compiled against Lua 5.3.6 (running with Lua 5.3)
>  • API level: 4
>  • D-Bus support: yes
>  • xcb-errors support: no
>  • execinfo support: yes
>  • xcb-randr version: 1.6
>  • LGI version: 0.9.2
>  • Transparency enabled: yes
>  • Custom search paths: no
> awesome_log: /tmp/tmp.N4rJ6ajGyZ/_awesome_test.log
> == Running ../tests/test-resize.lua ==
> Fontconfig error: No writable cache directories
> _client lua5.3: /usr/share/lua/5.3/lgi/override/Gdk.lua:23: bad argument #1 to 'registerlock' (userdata expected, got nil)
> _client stack traceback:
> _client         [C]: in function 'lgi.corelgilua51.registerlock'
> _client         /usr/share/lua/5.3/lgi/override/Gdk.lua:23: in main chunk
> _client         [C]: in function 'require'
> _client         /usr/share/lua/5.3/lgi/namespace.lua:183: in function 'lgi.namespace.require'
> _client         (command line):4: in main chunk
> _client         [C]: in ?
> Error: timeout waiting for signal in step 1/45 (@20) (/path/to/src/awesome/tests/../tests/test-resize.lua:9).
> Keeping awesome open...
> ```
> 
> This error I am not too sure about.

The line is related to GDK 4 compatibility, which is only available in LGI's dev version. So either building LGI from source or GDK 3 might help.

Elv13:
Just a little update. I have a some local changes to implement all of the changes I proposed, fix some little bugs and fix the doc. I plan to push them in the next few days to this branch. Since I also fix the code indentation, any change wont merge, so please don't update this branch until I push. Thanks

Elv13:
Hi,

I added a few commits to help this progress. First, I fixed the remaining warnings. I also fixed the indentation, which was inconsistent (2 space and 4 spaces).

## Change summary

 * Fix all warnings
 * Make indentation consistent across the file
 * Simplify/unify the validation (use `error()` rather than try to silently fix problems, move to setters) and fallback code (move to the getters rather than... everywhere)
 * Write the documentation
 * Finish the tests
 * Remove unnecessary constructors
 * Remove path builder because it belongs in `gears.filesystem` and wasn't really used anyway
 * Add more properties and a beautiful variable to replace `set_defaults` and hardcoded values.
 * Replace callbacks with signals (common pattern in modern AwesomeWM APIs)
 * Moved from `os.date` to GLib and some boilerplate code to make Debian devs less angry
 * Added snipping using the keyboard (`Return`, `Escape`) because people tend to complain about mouse only things.
 * Added a delay feature (because I want one :stuck_out_tongue: )
 * Changed the way the snipping tool selection wibox works. Now it freeze the screenshot before the selection. The old way had a bunch of side effects for things like Qt/GTK comboboxes
 * Add some extra code coverage to get the PR above the line and test the features I added

The business logic itself and the tests are (mostly) intact from the previous version. It change look larger than they really are. Every line changed, but the indentation issues and linting warnings are the main reason.

There's some code to eventually add an annotation tool. That's actually rather simple since we have all the widgets except the pen already and we can paint random widget on the screenshot.

image


## Change details and rationals

Fix the CI warnings (not your fault, it was broken and I fixed it last week)

```txt
WARNING: The directory property from awful.screenshot is missing it's type.
WARNING: The prefix property from awful.screenshot is missing it's type.
WARNING: The filepath property from awful.screenshot is missing it's type.
WARNING: The method_name property from awful.screenshot is missing it's type.
WARNING: The screen property from awful.screenshot is missing it's type.
WARNING: The client property from awful.screenshot is missing it's type.
WARNING: The geometry property from awful.screenshot is missing it's type.
WARNING: The surface property from awful.screenshot is missing it's type.
WARNING:        filepath_builder        from    awful.screenshot        doesn't have a return value or @noreturn
WARNING:        save    from    awful.screenshot        doesn't have a return value or @noreturn
    lib/awful/screenshot.lua:19:7: unused variable 'naughty'
    lib/awful/screenshot.lua:36:3: trailing whitespace in a comment
    lib/awful/screenshot.lua:41:1: line contains only whitespace
    lib/awful/screenshot.lua:190:24: line contains trailing whitespace
    lib/awful/screenshot.lua:202:5: setting non-standard global variable 'initilialized'
    lib/awful/screenshot.lua:204:1: line contains only whitespace
    lib/awful/screenshot.lua:318:1: line contains only whitespace
    lib/awful/screenshot.lua:320:1: line contains only whitespace
    lib/awful/screenshot.lua:326:1: line contains only whitespace
    lib/awful/screenshot.lua:329:12: accessing undefined variable 'frame'
    lib/awful/screenshot.lua:329:22: accessing undefined variable 'frame'
    lib/awful/screenshot.lua:335:1: line contains only whitespace
    lib/awful/screenshot.lua:337:1: line contains only whitespace
    lib/awful/screenshot.lua:339:15: unused variable 'date_time'
    lib/awful/screenshot.lua:340:1: line contains only whitespace
    lib/awful/screenshot.lua:344:1: line contains only whitespace
    lib/awful/screenshot.lua:347:1: line contains only whitespace
    lib/awful/screenshot.lua:352:1: line contains only whitespace
    lib/awful/screenshot.lua:356:1: line contains only whitespace
    lib/awful/screenshot.lua:360:1: line contains only whitespace
    lib/awful/screenshot.lua:362:1: line contains only whitespace
    lib/awful/screenshot.lua:364:1: line contains only whitespace
    lib/awful/screenshot.lua:371:1: line contains only whitespace
    lib/awful/screenshot.lua:420:43: line contains trailing whitespace
    lib/awful/screenshot.lua:495:9: variable 'args' was previously defined as an argument on line 493
    lib/awful/screenshot.lua:521:7: setting non-standard global variable 'frame_color'
    lib/awful/screenshot.lua:538:9: variable 'args' was previously defined as an argument on line 537
    lib/awful/screenshot.lua:552:9: variable 'args' was previously defined as an argument on line 551
    lib/awful/screenshot.lua:569:9: variable 'args' was previously defined as an argument on line 565
    lib/awful/screenshot.lua:583:9: variable 'args' was previously defined as an argument on line 582
    lib/awful/screenshot.lua:597:9: variable 'args' was previously defined as an argument on line 596
    lib/awful/screenshot.lua:708:9: variable 'args' was previously defined as an argument on line 706
    lib/awful/screenshot.lua:736:67: line contains trailing whitespace
    lib/awful/screenshot.lua:757:67: line contains trailing whitespace
    lib/awful/screenshot.lua:799:9: variable 'args' was previously defined as an argument on line 797
    lib/awful/screenshot.lua:807:9: value assigned to variable 'method' is unused
    lib/awful/screenshot.lua:808:9: value assigned to variable 'method_name' is unused
```

Then I refactored the "default behavior" code. It was spread across multiple methods/function/variables and rather hard to follow/review. Now I moved it to the getter methods. That removed 200 lines and added 4. This also removes the `set_defaults` static function. It's not that having it isn't useful, it's that it has side effects which would be unfixeable in the future. Since this is a core module, it's often better to just be a little bit verbose, but predictable. Right now, a 3rd party module could have silently affected another module. I moved the `frame_color` default to a beautiful variable. At least those have a clear cut initialization flowchart.

I also removed the code which changed the behavior or getter based on the selected method. That seems rather unexpected. If someone want to set a screen for a client screenshot, then let them. If they set a screen and no client, then let the error happen. That will be way less confusing and easier to debug. I also removed the code to check the argument type. The problem is that it was causing errors to be silently ignored. We usually prefer the things to blow up when invalid types are passed. This will print a stacktrace in the UI and is easier to debug than ignoring problematic values. For better or worst, Lua equivalent of exceptions is just to let thing blow up and have a "protected call" higher in the stack to catch errors. 

Another little change in the same philosophy is that I consolidated the validation code. Things like `check_prefix/directory/filepath` were each called (up to) 7 time for each screenshot. I moved some checks to point of use and some other to point of definition. I understand how it ended up this way, 2 years worth of fixup in a single file will do that :), there is nothing wrong with it. I just wanted to make sure this was refactored prior to merging so simplify maintenance. It also make getting code code coverage easier, because some of those `if`s were unreachable.

I temporarily removed the filepath_builder and path validation/parsing from this PR. Feel free to submit it in its own PR for `gears.filesystem`. It's useful, it is worth merging, but it isn't related to the screenshot module.

Next, I moved some hardcoded values to properties. This include the file name, mouse cursor and date format, frame_shape, accept buttons, reject buttons, etc. Someone at some point would have complained about hardcoded values. It's fairly cheap to make them configurable, so I did.

Finally, I removed all the extra constructors and leave only one. Even in the last version, they didn't do anything beside setting the method. They also modified the `args` provided by the user, which has side effects if args is reused. The method argument can be deduced from the other `args` (with `"root"` as the fallback).

tests/test-screenshot.lua
```diff
@@ -478,4 +479,38 @@ table.insert(steps, function()
...
+
+    -- Cause some validation failures.
+    ss.prefix = "//////"
+    assert(count == 1)
+    ss.directory = "/tmp/////"
+    ss.directory = "/tmp"
+    ss.directory = "/root/"
+    assert(count == 2)
+
+    gdebug.print_error = err
```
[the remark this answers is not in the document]
Elv13: Perhaps, but the reason why busted is an optional dependency is this:

```sh
> docker run --rm -ti --net=host fedora:latest /bin/bash
[root@lepagee-thinkpad-ubuntu /]# dnf search busted
Fedora 36 - x86_64                                                                                  1.2 MB/s |  81 MB     01:09    
Fedora 36 openh264 (From Cisco) - x86_64                                                            1.8 kB/s | 2.5 kB     00:01    
Fedora Modular 36 - x86_64                                                                          732 kB/s | 2.4 MB     00:03    
Fedora 36 - x86_64 - Updates                                                                        1.3 MB/s |  27 MB     00:21    
Fedora Modular 36 - x86_64 - Updates                                                                883 kB/s | 2.8 MB     00:03    
No matches found.
[root@lepagee-thinkpad-ubuntu /]# dnf search awesome
Last metadata expiration check: 0:00:44 ago on Tue Sep  6 21:39:04 2022.
================================================== Name Exactly Matched: awesome ===================================================
awesome.x86_64 : Highly configurable, framework window manager for X. Fast, light and extensible
```

Adding new dependencies in features we want contributors to use (like, writing tests) adds friction. Then *when* we make a release, it means it gets ignored upstream because the bots cannot do the upgrades. `luarock` is not really a good option. It's forbidden for distro packages and it often use a different version of Lua than the one used to package Awesome, which means it plain doesn't work for these users. Could we get better assert messages and more elegant testing than the `assert` used in the integration and doc examples? Sure, but adding *even more* friction for contributions is IMHO way worst. An alternative might be to provide our own SDK in a docker container. But that doesn't solve the distribution upgrade side of the issue. 

Plus, in the specific case, we have some code to catch all calls to `print_error` and make the test fail, so it has to be neutered if we test the error code paths.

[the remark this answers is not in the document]
sclu1034: > Adding new dependencies in features we want contributors to use (like, writing tests) adds friction.

Doesn't that apply to docs, too? Because we just merged more than a dozen new, custom LDoc directives and a system that yells at the contributor that doesn't use them. And I'm not aware of any documentation on what directives exist or how to use them.
Whereas this scenario shows that stubs are nice-to-have, but in no way mandatory for a test to pass. Nothing would yell at a contributor for not using them. They could write their entire test suite without it, and it would be good enough.
Though, honestly, a contributor that can't copy&paste the stub example from busted's documentation will need significantly more guidance on the integration test framework as a whole. And if using stubs is friction, then that already exists in the spec tests, where we already use luassert.

But maybe I'm just missing the pattern for what's acceptable friction and what isn't.

> Then when we make a release, it means it gets ignored upstream because the bots cannot do the upgrades. 

As far as I can tell, Fedora wouldn't care. They don't run any of the test pipeline, even the parts that could be run without any dependencies.

actionless (approved):

lib/awful/screenshot.lua
```diff
@@ -0,0 +1,943 @@
...
+    directory               = screenshot_validation.directory(os.getenv("HOME")),
+    cursor                  = "crosshair",
+    date_format             = "%Y%m%d%H%M%S",
+    interactive             = false,
+    reject_buttons          = {abutton({}, 3)},
+    accept_buttons          = {abutton({}, 1)},
+    reject_keys             = {akey({}, "Escape")},
+    accept_keys             = {akey({}, "Return")},
+    minimum_size            = {width = 3, height = 3},
+    auto_dave_tick_duration = 1,
```
actionless, line 610: typo

actionless (changes requested):

poisson-aerohead:
Thanks everyone, especially @Elv13  and @actionless. I originally just tracked down the screen shot ability in the C code to try to add the functionality and then added a wrapper to my rc.lua, so I was not really intending on going this far into it, but it looks like it has worked out. I hope you have it up to project standards now and sorry for not really knowing what that is. Whatever you changed does not really bother me, as I am not really an Awesome maintainer so you all should keep the code organized however you want.

lib/awful/screenshot.lua
Elv13: done

Elv13 (approved):

actionless (approved):

PR #3462, opened 2021-10-12
Fix 3458

Aire-One:
@ezz666, @actionless, can you please try this patch and report back if the test now works with both lua53 and luajit?

I think the issue was that luajit can correctly implode the returned values from `root.size()` and map them to the function parameters... But we have to do that manually with lua5.3? :shrug: 

For me the test works with this fix.

I also fixed the root vs capi.root call. I'm not sure whether it does anything, but it's cleaner, I think.

### DIFF
--- a/tests/test-screenshot.lua
+++ b/tests/test-screenshot.lua
@@ -94,7 +94,7 @@ local function copy_to_image_surface(content, w, h)
 
     cr:set_source_surface(sur)
     cr:paint()
-    img:finish()
+    img:flush()
 
     return img
 end


Elv13:
@Aire-One You need to add `local capi = { root = root }` at the top of the screenshot file for the tests to pass.

Otherwise, if that's the problem, then that's *very* strange, but :shrug:. But thanks a lot for looking into it

Aire-One:
> You need to add local capi = { root = root } at the top of the screenshot file for the tests to pass.

Oops, I'm trying some tooling for check/format code, and it has added a lot of noise while working on this... So I did my commit with -p and forget to add this line... Anyway, it should be pushed now!

> Otherwise, if that's the problem, then that's very strange, but shrug. But thanks a lot for looking into it

I think the main issue was the root.size() used as a direct parameter for both width and height. 

The root/capi.root is something I have seen before, but right now I can't find back _where_, so I can't show you... ~~I'm honestly not sure if it has any real impact in the current situation.~~ I actually can test to remove the `capi.` in my local branch... Annnd, it also works without it!

So, the real fix is 96710a595ed5c910476d9f853e1ef9a8acd5b6ae, but let's keep also the other commit, because I think it's still cleaner to use this capi table instead of global variables.

Aire-One:
> @Aire-One did you mean lua53? Because I don't think I can get 5.5 without using a time machine.

Ahah yes! I fixed the typo, thanks. 

Reading again the lua doc (https://www.lua.org/pil/5.1.html), 
`local img = copy_to_image_surface(capi.root.content(), root.size())` should indeed work as expected and map the width and height returned values of root.size() to the parameters of copy_to_image_surface. So I'm not sure why my changes here made the test pass for me... 

It's getting late here and I'll not be able to make time to investigate this tomorrow... So I'll get back here later this week, I think.

poisson-aerohead:
I have wondered about the "grab the environment" thing with capi.

Elv13:
@psychon Do you have any idea what's going on? I can't reproduce locally. Neither can the CI. So it's hard to investigate possible solutions.

ezz666:
I use Dell workstation. It has Intel Xeon W-2155 in it. It may count as pretty fast one. I can test it on a laptop to be sure.

Aire-One:
I did this no-brainer test : `./run.sh test-screenshot.lua test-screenshot.lua [...] test-screenshot.lua` (with about 30 time the test I think :shrug:).

Before the last commit, it failed most of the time (maybe every time except the first run?).
After the last commit, it works flawlessly!

Thanks a lot @psychon, and thanks @ezz666 for the patch! :rocket:

Aire-One:
(a quick note to maintainers: yesterday I pushed this branch to awesomewm/awesome repo in addition to my fork. I have deleted this branch from the awesome repo. Sorry for the noise it (probably?) generates)

PR #3475, opened 2021-10-18
Another doc upgrade.

Elv13:
In this infinite loop which wont move the needle on some people perpetual unhappyness with the doc, here is more of it.

90% of the changes are for the `client` page. Some time ago, I wrote example for everything in the `tag` class. This is the equivalent for the `client` class.

### DIFF
--- a/lib/awful/client.lua
+++ b/lib/awful/client.lua
@@ -257,6 +257,7 @@ end
 -- @tparam[opt] client sel The client.
 -- @tparam[opt=false] boolean stacked Use stacking order? (top to bottom)
 -- @treturn[opt] client|nil A client, or nil if no client is available.
+-- @see client.get
 --
 -- @usage -- focus the next window in the index
 -- awful.client.next(1)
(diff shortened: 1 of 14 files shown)


Aire-One:
> > (I'm not sure if it's a build time error from my system since I couldn't understand what causes that) I have some image generated with rectangle from (0,0) to client (bottom,right) corner :
>
> Umm, that's unexpected. It works for me

So, I tried to investigate for a bit. Until I tried to open the SVG with Firefox to study the path and understand from what these rectangles are coming, since I couldn't remove them without removing borders in a whole... With firefox the image is perfect! The issue seems to come from the qutebrowser rendering engine.

Elv13:
Makes sense. Thanks for debugging. Can you confirm they look fine in the latest Chrome(ium)? I tried and it works for me. If it works for you too, I think we can merge this PR.

Elv13:
And now we have 90% coverage!*

 \* Because we count test code in the coverage and I copy paste too much

PR #2904, opened 2019-10-11
[WIP/RFC] Improve warning with already destroyed notification

blueyed:
Closes https://github.com/awesomeWM/awesome/issues/2737.

PR #3490, opened 2021-10-29
Less flacky tests

Elv13:
There are various flacky coverage sections in `events.c`, `xproperty.c` and `spawn.c`. This PR adds some explicit tests to make 4 flacky paths coverage stable.

Much more will be needed, like reparenting tests and startup notification tests. But lets do this over many PRs.

(this is to make Mergify more useful)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -253,7 +253,7 @@ jobs:
           grep -q 'May not access' lib/beautiful/init.lua
 
       - name: Build
-        run: cd "${{ github.workspace }}/build" && make
+        run: cd "${{ github.workspace }}/build" && make -j3
 
       # Executable needs to be run once to provide coverage results
       - name: Install and run
--- a/tests/test-client-shape.lua
+++ b/tests/test-client-shape.lua
@@ -11,6 +11,18 @@ runner.run_steps{
             assert(type(a) == "number", a)
         end
         if #client.get() == 1 then
+            local c = client.get()[1]
+
+            local geo = c:geometry()
+
+            -- Resize it to test shape change events.
+            c:geometry {
+                x      = geo.x,
+                y      = geo.y,
+                width  = geo.width * 2,
+                height = geo.height * 2,
+            }
+
             return true
         end
     end,
@@ -28,7 +40,14 @@ runner.run_steps{
         assert(not surface.load_silently(c.shape_clip, false))
 
         return true
+    end,
+
+    function()
+        client.get()[1]:kill()
+        return true
     end
+
+
 }
 
 -- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
--- a/tests/test-systray.c
+++ b/tests/test-systray.c
@@ -1,0 +1,118 @@
+/*
+ * A simple client icon that "does nothing".
+ *
+ * Copyright © 2021 Uli Schlachter <psychon@znc.in>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <xcb/xcb.h>
+#include <xcb/xcb_atom.h>
+#include <xcb/xcb_aux.h>
+
+#include "common/atoms.h"
+#include "common/util.h"
+
+// Include all the C code we might need directly into this file - I am lazy
+#include "common/atoms.c"
+#include "common/util.c"
+
+#define SYSTEM_TRAY_REQUEST_DOCK 0
+
+static xcb_atom_t systray_atom(xcb_connection_t *conn, int screen) {
+    char *atom_name;
+    xcb_intern_atom_reply_t *reply;
+    xcb_atom_t atom;
+
+    atom_name = xcb_atom_name_by_screen("_NET_SYSTEM_TRAY", screen);
+    if(!atom_name)
+        fatal("Error getting systray atom name\n");
+
+    reply = xcb_intern_atom_reply(conn, xcb_intern_atom(conn, 0, strlen(atom_name), atom_name), NULL);
+    free(atom_name);
+    if (!reply)
+        fatal("Error interning systray icon\n");
+
+    atom = reply->atom;
+    free(reply);
+    return atom;
+}
+
+static xcb_window_t find_systray(xcb_connection_t *conn, xcb_atom_t net_system_tray) {
+    xcb_get_selection_owner_reply_t* reply = xcb_get_selection_owner_reply(conn, xcb_get_selection_owner(conn, net_system_tray), NULL);
+    if (!reply)
+        fatal("Failed to request selection owner\n");
+    xcb_window_t owner = reply->owner;
+    if (owner == XCB_NONE)
+        fatal("No systray running\n");
+    free(reply);
+    return owner;
+}
+
+static uint32_t get_color(xcb_connection_t *conn, xcb_screen_t *screen, uint16_t red, uint16_t green, uint16_t blue) {
+    xcb_alloc_color_reply_t *reply = xcb_alloc_color_reply(conn, xcb_alloc_color(conn, screen->default_colormap, red, green, blue), NULL);
+    if (!reply)
+        fatal("Error allocating color");
+    uint32_t pixel = reply->pixel;
+    free(reply);
+    return pixel;
+}
+
+int main() {
+    int default_screen;
+    xcb_connection_t* conn = xcb_connect(NULL, &default_screen);
+    if (xcb_connection_has_error(conn)) {
+        fatal("Could not connect to X11 server: %d\n", xcb_connection_has_error(conn));
+    }
+    atoms_init(conn);
+    xcb_screen_t* screen = xcb_aux_get_screen(conn, default_screen);
+
+    // Create a window for the systray icon
+    xcb_window_t window = xcb_generate_id(conn);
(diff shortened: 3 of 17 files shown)


actionless (dismissed):

tests/test-titlebar.lua
```diff
@@ -183,6 +197,48 @@ local steps = {
...
+            c:connect_signal(event, function()
+                table.insert(events, event)
+            end)
+        end
+
+        next_pos = { x = c:geometry().x + 5, y = c:geometry().y + 5 }
+
+        return true
+    end,
+    click,
```
Aire-One, line 211: I can't understand what behavior from coverage you want to trigger with these clicks. Can you elaborate?

tests/test-client-shape.lua
```diff
@@ -27,8 +39,17 @@ runner.run_steps{
         assert(not surface.load_silently(c.client_shape_clip, false))
         assert(not surface.load_silently(c.shape_clip, false))
 
+        client.get()[1].fullscreen = false
```
Aire-One, line 42: What does the `fullscreen = false` changes here? Isn't the client already not fullscreen?

.github/workflows/main.yml
```diff
@@ -253,7 +253,7 @@ jobs:
           grep -q 'May not access' lib/beautiful/init.lua
 
       - name: Build
-        run: cd "${{ github.workspace }}/build" && make
+        run: cd "${{ github.workspace }}/build" && make -j8
```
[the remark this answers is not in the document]
Elv13: Noted, I updated the number

tests/test-client-shape.lua
Elv13: whoops, leftover from a previous attempt at hitting that `if`.

tests/test-titlebar.lua
Elv13: This one is one of the flacky coverage `if`:
image

.github/workflows/main.yml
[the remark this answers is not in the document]
actionless: i thought you purposely set it like that for test purpose - otherwise instead of fixed number might make sense to use `$(nproc)`

tests/test-systray.c
```diff
@@ -0,0 +1,98 @@
+/* Code  https://stackoverflow.com/questions/45392284/freedesktop-xembed-systray-client-code-in-c-xlib */
```
psychon, line 1: Somehow I doubt that this code is licensed as GPLv2...

tests/test-systray.c
```diff
@@ -0,0 +1,98 @@
...
+    ev.xclient.data.l[0] = CurrentTime;
+    ev.xclient.data.l[1] = message;
+    ev.xclient.data.l[2] = data1; // <--- your window is only here
+    ev.xclient.data.l[3] = data2;
+    ev.xclient.data.l[4] = data3;
+
+    trap_errors();
+    XSendEvent(dpy, tray, False, NoEventMask, &ev);
+    XSync(dpy, False);
+    usleep(10000);
```
psychon, line 59: Why is there a `usleep` here?

(Also: Why is this Xlib at all. AFAIR there is no Xlib code in AwesomeWM...)

tests/test-systray.c
actionless: it's `Creative Commons Attribution-ShareAlike 3.0 license`

tests/test-systray.c
```diff
@@ -0,0 +1,98 @@
...
+    root = RootWindow(dpy, screen);
+    if(!XGetWindowAttributes(dpy, root, &wa))
+        return 1;
+    width = height = MIN(wa.width, wa.height);
+
+    /* create window */
+    win = XCreateSimpleWindow(dpy, root, 0, 0, width, height, 0, 0, 0xFFFF9900);
+
+    send_systray_message(dpy, win, SYSTEM_TRAY_REQUEST_DOCK, win, 0, 0);
+    XMapWindow(dpy, win);
```
psychon, line 88: I have not much clue about systray/xembed, but according to a comment in our code, this is "kinda wrong":
https://github.com/awesomeWM/awesome/blob/50b9b1043716c9ff0e27ab862112b0ee6821fc16/event.c#L811-L814

psychon:
Here is some C code that creates a systray "icon" and has no unclear license. No idea whether I got the color correct. https://gist.github.com/psychon/49cafeb4002e06cb64bb348888edb063

Elv13:
Thanks @psychon, I replaced the stackoverflow code with yours. Overall, it covers just as many lines, but not quite the same ones. For the systray, the difference doesn't seem very important:

```
if (!xembed_info_get_reply(globalconf.connection, em_cookie, &em.info)) {
    em.info.version = XEMBED_VERSION;
    em.info.flags = XEMBED_MAPPED;
}
```

is the only code path that differs. For `xembed` and reparenting, what do you think would be a good test? Something like:

1. Spawn 2 window
2. `fork()` into another program and spawn 2 more window
3. connect the pipes so it can send window id over the pipe
4. reparent both  process second window into the other
5. xembed process 2 into process 1
6. kill process 2
7. un-parent process 1 windows?

I am not really sure, I never worked with these APIs before and whatever I will come up with will be somewhat incorrect for whatever reasons. The goal is to "hit the most lines" in the coverage so the coverage flackyness goes away. Right now, the problem is that each PR has a +/- 0.02% changes in coverage due to random timing related events when running the test suite. This PR explicitly covers 4 common ones along with adding the systray. The other major ones are the startup notification timeout code and the xproperty code. Both are pretty easy to cover, but are not in this PR. I also need to extend the test client to cover the remaining states in `ewmh.c` (sticky, fullscreen, urgent, skip-taskbar, role change, modal change, etc).

psychon:
> the difference doesn't seem very important:

Yeah, well, that is because this program does things "properly" instead of that race code that the old example did :man_shrugging: 

> For xembed and reparenting, what do you think would be a good test? 

Sorry, what? What are you trying to cover? We only use xembed for the systray code and that for that we now have an example program. If you want to cover more systray/xembed code, then that needs a more complex example program. But I doubt that this would have anything to do with flaky code coverage, because there is nothing else systray-y in the test suite (and even this example likely doesn't have anything to do with flaky coverage).

Do you have an example for what exactly you'd want covered?

Elv13:
> What are you trying to cover?

Ok, lets rephrase this. If I was to write/add some random C tests to guarantee most of the remaining uncovered and/or flacky C code would get hit reliably, what would you suggest those test do? That's beside the spawn race conditions and the extending the GTK client to have more delayed property changes trigerred by the client itself (because I am aware of those issue and they are easy to fix when I get time).

psychon:
I'm not sure. The xembed code shouldn't have any flakyness and for the rest... uhm... could a link to the diff between two coverage reports for me to look at? :shrug:

.github/workflows/main.yml
```diff
@@ -253,7 +253,7 @@ jobs:
           grep -q 'May not access' lib/beautiful/init.lua
 
       - name: Build
-        run: cd "${{ github.workspace }}/build" && make
+        run: cd "${{ github.workspace }}/build" && make -j3
```
sclu1034, line 256: The Ubuntu runners only have two virtual cores. Therefore, forcing three threads makes things _slower_ because it forces the CPUs into context switching to run three threads on two cores.

Elv13:
> The xembed code shouldn't have any flakyness and for the rest

Actually, it had on most systems, but not the CI. However, it happens very indirectly:

```lua
local app = Gtk.Application {}
app:run {''}
```

This little piece of code we have a few time in the tests is the culprit. "run" actually does more than meet the eyes. It also starts some daemons in the backgrounds. This, in turn, calls either `qt5-xdgdesktopportal-platformtheme`, `xdg-desktop-portal-gtk` or `xdg-desktop-portal-kde`. In turn *they* start some more deamons. In my case, the KDE portal started `kdeconnectd`, which has a systray.

That systray was the source of the flackyness. If awesome quitted before kde-connect started, then the next `awesome` process would start with a systray icon from the get go. If the systray won the race, it would be added "live" in the current `awesome` process. That would have a couple different lines in the coverage.

The only reason it wasn't flacky on the CI is because less optional packages are installed.

> uhm... could a link to the diff between two coverage reports for me to look at? shrug

Without actually looking beyond the summary:

https://github.com/awesomeWM/awesome/pull/2904
https://github.com/awesomeWM/awesome/pull/3456
https://github.com/awesomeWM/awesome/pull/3443
https://github.com/awesomeWM/awesome/pull/3439
https://github.com/awesomeWM/awesome/pull/3434
https://github.com/awesomeWM/awesome/pull/3400

.github/workflows/main.yml
Elv13: The old wisdom might be outdated now that SSDs exists. Back in the days, it was always recommended to use core+1 (unless the cores have hyperthreading) because hard drive I/O was a major factor. 3 thread on a dual core system was much faster than 2 when installing Gentoo (hours faster). It was slower if the cores already had hyper threading. Because that's exactly what hyperthreading aims to solve (bottlenecked cores).

https://web.archive.org/web/20051126053125/http://www.gentoo.org/doc/en/handbook/handbook-x86.xml?full=1

image


(ok... I am getting old)

sclu1034: That's a good point, though. I didn't think of the I/O potentially being slow. And with everything being virtualized hardware in The Cloud, we can't really find out easily.

But I just realized that the build step itself only takes 10 seconds or less in CI, so there shouldn't be much of a difference between the two. So `-j3` is fine.

psychon:
How can I actually look at the diffs beyond the summary? https://app.codecov.io/gh/awesomeWM/awesome/compare/3400/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=awesomeWM#diff-cHJvcGVydHkuYw== says "No tracked files changed".

sclu1034:
That PR only changes `doc/_parser.lua`, which Codecov doesn't seem to track (the entire `doc/` directory doesn't appear in the overview). But I'm actually not sure where that's defined.

If you check one of the other PRs, you should see a diff. E.g. <https://app.codecov.io/gh/awesomeWM/awesome/compare/3456/diff> is one where a trivial change in a warning message results in coverage changes across multiple unrelated files.

psychon:
> multiple unrelated files.

That diff only shows me a change in `lib/naughty/init.lua`, function `screen_fallback()` was called with `capi.screen.count() == 0` where it previously wasn't.

That function... is only called in two places:
- when there were startup errors (`capi.awesome.startup_errors` is true-ish). This cannot happen in tests, I think.
- When the `debug:error` signal is called. Uhm.... shouldn't any time that this signal is emitted cause the test to also fail? Apparently not, there are tests that emit this on purpose...

Actually, `tests/test-awful-screen.lua` explicitly removes all screens and then emits `debug::error`. Thus, this being not covered before must be a bug in the test or something. How else could this coverage change occur???

PR #3492, opened 2021-11-01
LC_CTYPE replaced with LC_ALL

JezerM:
## Changes

- `LC_CTYPE` replaced with `LC_ALL`, in `setlocale`

## Explanation

As the docs explain in [textclock widget](https://awesomewm.org/apidoc/widgets/wibox.widget.textclock.html), GLIB should format the date according to the locale.

In my case, the locale is `LANG = es_NI.UTF-8` and `LANGUAGE = es_ES`. And therefore, date format should be in spanish, but AwesomeWM does not format in that way, it's just english.

In the image below, you can see the actual behavior (wibar above), which uses `setlocale(LC_CTYPE, "")`, and the expected behavior (wibar below), that uses `setlocale(LC_ALL, "")`, with `wibox.widget.textclock("%a %B %Y - %H:%M")`

Screenshot_2021-11-01-55_675x568

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -588,7 +588,7 @@ main(int argc, char **argv)
     awesome_argv = argv;
 
     /* Text won't be printed correctly otherwise */
-    setlocale(LC_CTYPE, "");
+    setlocale(LC_ALL, "");
 
     char *confpath = options_detect_shebang(argc, argv);
 


actionless:
why?

sclu1034:
> why?

They want date and time to be shown using a Spanish locale and set the appropriate settings/environment variables.
But Awesome (actually, GLib's `DateTime`) ignores those and displays English text.

Side note: The proposed patch is actually something that GLib [recommends doing to be user-locale aware](https://docs.gtk.org/glib/running.html#locale).

actionless:
> The proposed patch is actually something that GLib recommends doing to be user-locale aware.

that kind of link was smth i was expecting to be present as PR reasoning :)
thanks for looking it up!

actionless (approved):

Aire-One (approved):

PR #3517, opened 2021-11-24
Added option to center the mouse instead of restoring position. Useful

MaciekChudek:
Option for users to center mouse instead of restoring position. Useful in general and improves functionality after adding/removing monitors with xrandr.

### DIFF
--- a/lib/awful/screen.lua
+++ b/lib/awful/screen.lua
@@ -93,27 +93,34 @@ end
 -- @tparam screen _screen Screen number (defaults / falls back to mouse.screen).
 -- @request client activate screen.focus granted The most recent focused client
 --  for this screen should be re-activated.
-function screen.focus(_screen)
+function screen.focus(_screen, ...)
     client = client or require("awful.client")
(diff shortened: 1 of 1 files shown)


actionless:
it should be documented and use `args` table for extra arguments

MaciekChudek:
Is there a developer guide for how you guys prefer things done/documented?

actionless:
yeah, it's right in readme, called "Contributing code"

actionless:
however thing for `args` is not documented (yet?) anywhere, should it be mentioned in dev guide, @Elv13 ?

Elv13:
> should it be mentioned in dev guide, @Elv13 ?

Probably. But I am in vacations right now, so I don't have the means to submit a proper PR.

But tl;dr @MaciekChudek, it looks like this:

```lua
--- Some summary.
-- @method foo
-- @tparam[opt={}] table args
-- @tparam[opt=false] boolean args.center_mouse Some description.
``` 

You can install `ldoc` and see how it renders in your browser

PR #3520, opened 2021-11-27
Multi-row support of systray.

xinhaoyuan:
Also, fix the fitting logic so that the result width and height are fitted in the available space. Previously the result dimensions can be larger than the input, but cropped in https://github.com/awesomeWM/awesome/blob/e7a21947e6785f53042338c684b9b96cc9b0f500/lib/wibox/widget/base.lua#L547. But still it can cause problems e.g. when used inside wibox.container.place.

### DIFF
--- a/lib/wibox/widget/systray.lua
+++ b/lib/wibox/widget/systray.lua
@@ -30,6 +30,13 @@ local display_on_screen = "primary"
 -- @beautiful beautiful.bg_systray
 -- @param string The color (string like "#ff0000" only)
 
+--- The maximum number of rows for systray icons. Icons will fill the
+-- current column (orthogonally to the systray direction) before
+-- filling the next column.
+--
+-- @beautiful beautiful.systray_max_rows
+-- @tparam[opt=1] integer The positive number of rows
+
 --- The systray icon spacing.
 --
 -- @beautiful beautiful.systray_icon_spacing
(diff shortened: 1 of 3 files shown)


Elv13:
Thanks for this contribution, I am AFK for a few more weeks. So it mght take some time before I review/test this. However it looks very nice. I wonder if it's possible to make the API a bit more dynamic/flexible rather than only expose a `beautiful` variable? Maybe adding an `icon_size` and `maximum_row_count`/`maximum_column_count` property to the `systray.lua` object?

xinhaoyuan:
Thanks for the suggestions!

> Maybe adding an icon_size

There is the `systray.base_size` property for the desired icon size - is this what you were thinking of?

> maximum_row_count/maximum_column_count property

I guess the two properties cannot be used at once (since I'm not sure what to do when there are more icons than `max_row * max_column`). I realized that `max_rows` + `base_size` + some wibox container can subsume the original `forced_rows` mechanism I was proposing. Thus I now changed to introduce only the `max_rows` property. `max_column` can be achieved by setting `horizontal` to false and use `max_rows` instead.

About whether to keep `max_rows` as a beautiful constant or systray property. I don't have a strong opinion, but given that I don't have the need to change the value, similarly to the `icon_spacing` constant, keeping it as a beautiful constant is enough for me. But I can be missing something.

Aire-One (commented):
I think it would be great to have some examples in the doc. Maybe that's not something we want to do in the scope of this PR, tho.

actionless (approved):
so, let's try this? :)

Aire-One (approved):

Elv13 (approved):

PR #3524, opened 2021-12-03
add default wibox layout + minor change

antonilol:
when no layout is specified no error will popup but instead it will use
the default `wibox.layout.fixed.horizontal`

and a minor change:
error happened -> error occurred (2x)

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -29,7 +29,7 @@ require("awful.hotkeys_popup.keys")
 naughty.connect_signal("request::display_error", function(message, startup)
     naughty.notification {
         urgency = "critical",
-        title   = "Oops, an error happened"..(startup and " during startup!" or "!"),
+        title   = "Oops, an error occurred"..(startup and " during startup!" or "!"),
         message = message
     }
 end)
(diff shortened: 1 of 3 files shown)


Aire-One (commented):
Hello,

Thanks for your submitting a PR! :)

I don't know about the changes you proposed, tho...
I feel like we don't want to force a default layout when building a widget. If the user forget to specify it, it's the user fault, and we shouldn't "guess" what they actually tried to do. Also, I don't know if the horizontally aligned layout is the best candidate here :shrug:

antonilol:
i thought of both (align and fixed) and found out fixed is used more (in default rc.lua) 
btw, what is [this](https://github.com/awesomeWM/awesome/blob/master/awesomerc.lua#L556)? (the `()`)

Aire-One:
> btw, what is this? (the ())
>
> is there a difference between wibox.layout.fixed.horizontal and wibox.layout.fixed.horizontal()?

Let's say : no, there is no difference for the end-user.

The long version is :

`wibox.layout.fixed.horizontal` is the "class" while `wibox.layout.fixed.horizontal()` is an instance, since calling the class will actually call the constructor (thanks to our meta-table implementation). The second version (with extra parenthesis), will build/instantiate a horizontal fixed layout object, while the first version  (without the parenthesis), basically is a reference to the class.

As you probably have seen in the `drill` function, we look for certain properties to determine what is the widgets (1) that the user want to use. We either, instantiate a new one if this is a table, or use the instance provided. Then, we assign the desired values to all the properties.

(1) Widgets, layouts and containers are all derivative of the same "object", so let's call them all "widgets" here.

PR #3526, opened 2021-12-14
Support minimized client property in tasklist icons

gzbd:
As the title says, if the client is minimized add an icon to the client name in tasklist.


Elv13 (approved):

sclu1034 (approved):

Elv13:
Thanks!

PR #3530, opened 2021-12-21
feat(spec): Create visual error for widget layout tests

sclu1034:
The assert for `widget:layout` now visualizes the expected and actual
widget structure for comparison.

The "Widget" line is a placeholder. If at a later point `utils.widget_stub` and calls to it are changed to take an optional widget name, that should be assigned to `widget._name` and it will show up instead of "Widget".

Example:
image


Elv13 (approved):

actionless (approved):

Elv13:
Thanks!

PR #3533, opened 2021-12-22
Add dedicated dragging state for slider widget

sclu1034:
When trying to hook up the slider to an external value, one would usually have two data streams:

- `property::value` -> set external value
- external value changed -> `.value = new_value`

The problem is that without manual intervention, these two streams form a loop, as setting `.value` also emits `property::value`.

Having a dedicated "dragging" state with associated signals disconnects these data streams.


Elv13 (dismissed):

Elv13:
Thanks for this. Too bad testing this would be rather non trivial. If you want to do it, you could copy/paste `test-awful-widget-button.lua` and replace the button with the slider.

sclu1034:
The test runs fine when executed solo (`./tests/run.sh test-wibox-widget-slider.lua`) and it uses the same fake input as `test-awful-widget-button.lua`, so no clue why it fails when run as part of the full set of tests.

Elv13:
Usually, you have to check for completion in each step or `return nil`. This wait another event loop iteration and try again. A lot of XCB is async, so those things are not deterministic. Each step is retried 10 time when it returns nil. Adding `sync` should not be necessary, but is done in a few places.

I would also appreciate not adding a new hard dependency here. Even `busted` is optional. Many distribution package awesome and run the tests as part of the build process. Adding a new dep will potentially delay the packaging of 4.4 "because it's harder than changing the version number of the spec/control/pkgbuild file".

Elv13:
The failure is strange. I am not so sure. Maybe adding a couple empty steps after the wibox creation might help (there is a bunch of potentially async stuff going on before the widget is displayed, so the `step` might win or lost a race with other delayed calls?). It's 2:30 am here. I will see if I can reproduce it tomorrow.

actionless (approved):

sclu1034:
Spend some more time trying to investigate why the signals don't fire in the tests.

A simple function like this works perfectly fine locally, both in Xephyr and Xvfb:
```lua
-- Same wibox+slider+signals setup as in the test's first step
	mouse.coords({ x = 12, y = 24 })
	root.fake_input("button_press", 1)
	mouse.coords({ x = 100, y = 24 })
	root.fake_input("button_release", 1)
end)
```

So clearly, async is not the issue and no waiting between those actions is necessary for the individual signals to fire.
They do only fire after the step that triggered them has finished, but that's fine, as long as triggering and checking are two separate steps.

The actual reason why no signals fired in CI is that the mouse never actually hit the widget. I don't know why, but the position at which the wibox is placed is not always consistent with what `x` and `y` are set to.

actionless (approved):

Aire-One (approved):

Elv13 (approved):

tests/test-wibox-widget-slider.lua
```diff
@@ -0,0 +1,107 @@
...
+    table.insert(steps, func)
+end
+
+-- Apparently in CI the wibox won't always be at the coordinates we specify.
+-- So we'll have to click around blindly until we find our widget,
+-- then use those coordinates as offset for the actual test.
+local offset_x = nil
+local offset_y = nil
+
+local w
```
actionless, line 21: https://github.com/awesomeWM/awesome/runs/7111106748?check_suite_focus=true#step:33:903

sclu1034: Good catch.

actionless (approved):

lib/wibox/widget/slider.lua
```diff
@@ -303,6 +325,14 @@ function slider:set_value(value)
     end
 end
 
+--- Returns `true` while the user is dragging the handle.
+--
+-- @method is_dragging
```
Elv13, line 330: sorry for missing this earlier, this should be `@property` (there is magic to make it work)

sclu1034: Do you mean "transform this into a property by adding `get_is_dragging` and `set_is_dragging`?"
Because there is currently no magic involved. Indexing `.is_dragging` returns a function, not a boolean.

sclu1034:

PR #3534, opened 2021-12-23
fix: Run rule callbacks **after** applying properties

HumblePresent:
The docs specify that any rule callbacks are run **after** the rule properties are applied, but that is not currently the case for the client and notification rules as well as the generic matcher. This PR fixes that by ensuring the callbacks are run afterward. I assumed the docs correctly described the desired behavior with properties being applied before callbacks, but if that is not the case we can leave the code as is and modify the docs instead.

### DIFF
--- a/docs/common/rule.ldoc
+++ b/docs/common/rule.ldoc
@@ -9,7 +9,7 @@
 --TODO add ^
 -- @DOC_text_gears_matcher_properties_EXAMPLE@
 
---- A callback function to call *after* the `properties` have been applied.
+--- A callback function to call *before* the `properties` have been applied.
 --
 -- The callback is called with the current client as first argument.
 --
--- a/lib/gears/matcher.lua
+++ b/lib/gears/matcher.lua
@@ -488,6 +488,13 @@ end
 --  sole argument. The callbacks are executed *before* applying the properties.
 -- @see gears.matcher.apply
 function matcher:_execute(o, props, callbacks)
+    -- Apply all callbacks.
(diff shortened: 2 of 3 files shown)


Elv13:
May I suggest going the other way around? Not because it makes more sense, just because it doesn't introduce a behavior change a couple weeks before a major release.

Elv13 (approved):

actionless (approved):

actionless:
should the new ticket be created for fixing that after the release?

Elv13:
> should the new ticket be created for fixing that after the release?

We could "replace" the current callback(s) with a more fine grained set of callbacks. Having it executed before and after have different advantages. One of them can manipulate the property table before it is applied and the other can make sure some of them are only applied at the end.

sclu1034:
That should already be possible by adding separate rules for these callbacks:

```lua
-- "before" callback
ruled.client.append_rule {
  rule = { class = "xterm" },
  callback = print,
}
-- apply properties
ruled.client.append_rule {
    rule = { class = "xterm" },
    properties = {
        maximized_vertical   = true,
        maximized_horizontal = true
    },
}
-- "after" callback
ruled.client.append_rule {
  rule = { class = "xterm" },
  callback = print,
}
```

Adding more callbacks to individual rules would just increase their complexity.

HumblePresent:
I might be wrong but from what I can tell all the properties and callbacks are collected from a given source before being applied, so even in the example above all the callbacks will still be called before the properties are set.

PR #3537, opened 2021-12-25
Fix a regression that only property::x and property::width are emitted…

### LINKED ISSUE #3535: Only one of {`property::x`, `property::y`} (and similarly for {`property::width`, `property::height`}) is emitted when resizing a client.
See the logic here https://github.com/awesomeWM/awesome/blob/1413f0c4b9da13b17b7fc11e2edd78daf695f3b4/objects/client.c#L2403-L2418. It has been there since v4.1. 

First, find a client, e.g.
```
local c = client.focus
```

Then set up the initial geometry and callbacks for property::*:

```
c:geometry({x = 100, y = 100, width = 200, height = 200})
c:connect_signal("property::x", function (client) print("property::x", client) end)
c:connect_signal("property::y", function (client) print("property::y", client) end)
c:connect_signal("property::width", function (client) print("property::width", client) end)
c:connect_signal("property::height", function (client) print("property::height", client) end)
```

Then call geometry e.g.

```
c:geometry({x = 0, y = 0, width = 100, height = 100})
```

**Actual result:**

You will find in the log that only `property::x` and `property::width` is called.

**Expected result:**

All four `property::*` callbacks are called.

xinhaoyuan:
… when all {x,y,width,height} properties are changed.

This fixes #3535.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -2405,7 +2405,7 @@ client_resize_do(client_t *c, area_t geometry)
         luaA_object_emit_signal(L, -1, "property::position", 0);
         if (old_geometry.x != geometry.x)
             luaA_object_emit_signal(L, -1, "property::x", 0);
-        else
+        if (old_geometry.y != geometry.y)
             luaA_object_emit_signal(L, -1, "property::y", 0);
     }
     if (old_geometry.width != geometry.width || old_geometry.height != geometry.height)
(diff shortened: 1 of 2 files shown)


Elv13 (dismissed):

Elv13:
Thanks!

(and don't worry about the coverage going down, we are still stabilizing it. It got quite flacky when we moved from Travis CI to GitHub Actions)

xinhaoyuan:
(fixed a typo :)

actionless (approved):

Aire-One (approved):

Elv13 (approved):

Issue #3379, opened 2021-07-01
Unable to open the right click menu in chromium after a screenshot


mattbdc:
Output of `awesome --version`:

```
➜  ~ awesome --version
awesome v4.3 (Too long)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
➜  ~ 
```

**How to reproduce the issue:**

```
Open chromium or google-chrome
Run flameshot to point where you are selecting an area to capture (i.e. overlay shows), press esc
Return focus to chrome
```

**Actual result:**

```
Right click doesn't work in the chrome window (on tabs for example)
Can be fixed by switching to another tag and back
```

Identical issue with i3, they fixed it
https://github.com/i3/i3/commit/a06e66982bc921d05b68b75f8e12323c284233d0

mattbdc:
Browsing the awesomewm issues I think this could potentially fix several reported problems - spotify etc.., awesomewm randomly hands for keys+mouse etc..

basaran:
I can report the same issue. Switching between different clients restores the right click. I didn't need to switch tags. It would be nice to have this patched.

awesome v4.3 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

mattbdc:
Yes can confirm client switch also resolves it. I just had one client in this tag so didn't notice tag.

psychon:
Completely untested port of that i3 patch to AwesomeWM. No idea if I got it right.
```diff
diff --git a/event.c b/event.c
index 030f86166..8c49ea04a 100644
--- a/event.c
+++ b/event.c
@@ -696,6 +696,12 @@ event_handle_enternotify(xcb_enter_notify_event_t *ev)
 static void
 event_handle_focusin(xcb_focus_in_event_t *ev)
 {
+    if (ev->event == globalconf.screen->root) {
+        // FIXME I doubt this is entirely correct, but I'm too lazy to think too
+        // much about it right now.
+        globalconf.focus.need_update = true;
+    }
+
     if (ev->mode == XCB_NOTIFY_MODE_GRAB
             || ev->mode == XCB_NOTIFY_MODE_UNGRAB)
[...]

PR #3543, opened 2022-01-07
Refocus focused window for focus in events on the root window.

tuwuna:
This deals with (admittedly somewhat misbehaving) clients which
use XSetInputFocus to take focus, but then don't properly restore
focus. This has been observed with TK apps, but also, e.g., Steam.

Ported from https://github.com/i3/i3/commit/a06e66982bc921d05b68b75f8e12323c284233d0 by @psychon 

Fixes https://github.com/awesomeWM/awesome/issues/3379

### DIFF
--- a/event.c
+++ b/event.c
@@ -696,6 +696,11 @@ event_handle_enternotify(xcb_enter_notify_event_t *ev)
 static void
 event_handle_focusin(xcb_focus_in_event_t *ev)
 {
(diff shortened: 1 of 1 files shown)


Aire-One (approved):
Cool, thanks!

actionless:
does it also solving that notorious issue with telegram preview? (never experienced it myself, but i saw people creating it almost every month)

Issue #3123, opened 2020-06-30
incompatibility with lua5.4


multiSnow:
[Lua 5.4 was released on 29 Jun 2020.](https://www.lua.org/versions.html#5.4)

[`math.randomseed` should be integer](https://www.lua.org/manual/5.4/manual.html#pdf-math.randomseed) in lua5.4, which is `g_random_int() + g_random_double()` in awesome.c, cause luaA_panic:
```
...
2020-06-30 10:48:28 W: awesome: luaA_panic:860: restarting awesome
2020-06-30 10:48:28 W: awesome: luaA_panic:855: unprotected error in call to Lua API (bad argument #1 to 'math.randomseed' (number has no integer representation))
2020-06-30 10:48:28 W: awesome: luaA_panic:859: dumping backtrace
...
```

bittin:
Had this problem too and fighted for 2 hours and had to downgrade lua to get my wm working

psychon:
Since #3124 is just about making AwesomeWM work with Lua 5.4: Anyone wants to do the work of adding Lua 5.4 to Travis CI?

actionless:
> do the work of adding Lua 5.4 to Travis CI?

i can do, but i dunno which testsuites (and other env vars) to choose to be running for that setup

psychon:
Well... you could start from what was done in `.travis.yml` before b67e0a123cb2c013f6b4665cb6030717a1463422 (back then, we built Lua 5.3 ourselves on Travis).
Most of that is "to get things working". `-ULUA_COMPAT_5_2` is to disable compatibility, so that more things break which would otherwise only break later. Same for `-DLUA_USE_APICHECK -Dlua_assert=assert`: Enable as many consistency checks as possible to find bugs.

Anyway, testing AwesomeWM with Lua 5.4 is blocked by LGI working on Lua 5.4. I don't know what the state there is (but I would expect things not to work out of the box due to API changes in lua...?

actionless:
you right, @psychon , while luarocks is working:

```
Configuring LuaRocks version 3.0.4...

Lua interpreter found: /usr/bin/lua

Checking if /usr/bin/lua is Lua version 5.4... yes

lua.h found: /usr/include/lua5.4/lua.h

```

lgi is failing to build:

```
[...]

PR #3124, opened 2020-07-01
only use int random number for math.randomseed (for lua5.4)

multiSnow:
This PR pass an int random number instead of double number to `math.randomseed`.

It try to fix #3123.

PR #3553, opened 2022-01-22
Supports Lua 5.4

Aire-One:
Hello,

Following recent conversations from the LGI repository and Lua 5.4 support, here is a first step forward for Awesome WM.

I'll blindly follow the patch written by @Elv13 (I'll also check some old patches that were proposed when Lua 5.4 was first released) here https://github.com/awesomeWM/awesome/issues/3123#issuecomment-1019159435. The strategy I want here is to first try the CI against Lua 5.4, see the failing result, then add the fixes to "prove" they are what we need to support Lua 5.4.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -207,9 +207,21 @@ jobs:
           cd /tmp/xcb-errors
           sudo make install
 
+      # Only LGI git master supports Lua 5.4. We need to build it from source.
+      - name: Build and Install LGI
+        if: matrix.lua_version == '5.4'
+        run: |
+          wget -O /tmp/lgi.zip https://github.com/pavouk/lgi/archive/refs/heads/master.zip
+          unzip /tmp/lgi.zip -d /tmp
+          cd /tmp/lgi-master
+          make all
+          sudo make install
+
       - name: Install rocks
         run: |
-          sudo -H luarocks install lgi ${{ matrix.lgi_version }}
+          if [ "${{ matrix.lua_name }}" != "lua5.4" ]; then
+            sudo -H luarocks install lgi ${{ matrix.lgi_version }}
+          fi
           sudo -H luarocks install ldoc
           sudo -H luarocks install busted
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,6 +31,7 @@ jobs:
           - test_name: "lua5.4"
             lua_version: "5.4"
             lua_name: "lua5.4"
+            lua_library: "/usr/local/lib/liblua.a"
             lua_include: "/usr/local/include"
             manual_screens: true
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,6 +28,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - test_name: "lua5.4"
+            lua_version: "5.4"
+            lua_name: "lua5.4"
+            lua_include: "/usr/local/include"
+            manual_screens: true
+
           - test_name: "codecov-lua5.3"
             lua_version: "5.3"
             lua_name: "lua5.3"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -215,6 +215,7 @@ jobs:
           wget -O /tmp/lgi.zip https://github.com/pavouk/lgi/archive/refs/heads/master.zip
           unzip /tmp/lgi.zip -d /tmp
           cd /tmp/lgi-master
+          sed -i 's/5.1/5.4/' lgi/Makefile
           make all
           sudo make install
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -142,11 +142,23 @@ jobs:
         run: sudo dpkg -i /var/cache/apt/archives/*.deb
 
       - name: Install Lua packages
+        if: matrix.lua_version != '5.4'
         run: |
           if [ "${{ matrix.lua_name }}" = "luajit" ]; then
             sudo apt-get install libluajit-5.1-dev luajit
           fi
           sudo apt-get install liblua${{ matrix.lua_version }}-dev lua${{ matrix.lua_version }}
+
+      # Ubuntu 20.04 hasn't a package for Lua 5.4, we need to build it from source.
+      - name: Build and Install Lua 5.4
+        if: matrix.lua_version == '5.4'
+        run: |
+          wget -O /tmp/lua-5.4.3.tar.gz http://www.lua.org/ftp/lua-5.4.3.tar.gz
+          tar -xf /tmp/lua-5.4.3.tar.gz -C /tmp
+          cd /tmp/lua-5.4.3
+          make all
+          make test
+          sudo make install
 
       - name: Cache luarocks
         id: cache-luarocks
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -166,12 +166,12 @@ jobs:
         with:
           path: /tmp/luarocks
           # The build input for luarocks changes per test, so we need separate caches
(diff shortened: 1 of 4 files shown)


Elv13 (approved):

Aire-One:
Ahah, looks like only Ubuntu 21 has a Lua 5.4 package. Github Actions has Ubuntu 20. I guess we'll have to build Lua 5.4 manually to try it here...

.github/workflows/main.yml
```diff
@@ -136,12 +141,29 @@ jobs:
...
+      - name: Build and Install Lua 5.4
+        if: matrix.lua_version == "5.4"
+        run: |
+          sudo apt-get install --download-only -y --no-install-recommends \
+            curl \
+            tar
+          sudo dpkg -i /var/cache/apt/archives/*.deb
+          curl -R -O http://www.lua.org/ftp/lua-5.4.3.tar.gz
+          tar zxf lua-5.4.3.tar.gz
+          cd lua-5.4.3
```
sclu1034, line 162: It should probably be fine, but this does "pollute" the CWD.
The other steps all do that in `/tmp`, which also makes setting the cache path easier.

Aire-One:
Thank you for your review @sclu1034.

I have moved my work to my fork of awesome to prevent spamming everyone with my tries and failures while editing the GitHub actions workflow. These concern you here have been resolved on my fork.

.github/workflows/main.yml
```diff
@@ -28,6 +28,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - test_name: "lua5.4"
+            lua_version: "5.4"
+            lua_name: "lua5.4"
+            lua_include: "/usr/local/include"
```
Aire-One, line 34: This is required because we manually build Lua 5.4.

"If" a package is available in the future (because Ubuntu add it to their repositories, or we can move the GitHub Actions to uses Ubuntu 21), this path will most probably required to be updated.

.github/workflows/main.yml
```diff
@@ -136,24 +142,36 @@ jobs:
...
 
+      # Ubuntu 20.04 hasn't a package for Lua 5.4, we need to build it from source.
+      - name: Build and Install Lua 5.4
+        if: matrix.lua_version == '5.4'
+        run: |
+          wget -O /tmp/lua-5.4.3.tar.gz http://www.lua.org/ftp/lua-5.4.3.tar.gz
+          tar -xf /tmp/lua-5.4.3.tar.gz -C /tmp
+          cd /tmp/lua-5.4.3
+          make all
+          make test
```
Aire-One, line 160: The Lua doc recommends running `make test` when building to "be sure there is no error". I don't know if we want to keep that, tho...

.github/workflows/main.yml
```diff
@@ -189,9 +207,21 @@ jobs:
           cd /tmp/xcb-errors
           sudo make install
 
+      # Only LGI git master supports Lua 5.4. We need to build it from source.
+      - name: Build and Install LGI
+        if: matrix.lua_version == '5.4'
+        run: |
+          wget -O /tmp/lgi.zip https://github.com/pavouk/lgi/archive/refs/heads/master.zip
```
Aire-One, line 214: We need to build LGI from their master because there is currently no release with the latest patch for Lua 5.4. This will need to be changed when a new release is available.

Aire-One (commented):
I think the workflow itself is now ready! :)

The current failure it generates now are caused by our Lua version safeguards. I'll move on and apply the required patches until the CI succeed.

.github/workflows/main.yml
sclu1034: Highly unlikely that we'll see Ubuntu 21 in GH Actions. They've only used LTS releases so far, and the next one will be Ubuntu 22.
And we won't see Lua 5.4 in Ubuntu 20 either, as releases don't get feature upgrades.

.github/workflows/main.yml
psychon: > when a new release is available.

I admire your optimism. Please do not click on https://github.com/pavouk?tab=overview&from=2019-12-01&to=2019-12-31  :-)

.github/workflows/main.yml
```diff
@@ -189,9 +208,22 @@ jobs:
...
+          unzip /tmp/lgi.zip -d /tmp
+          cd /tmp/lgi-master
+          sed -i 's/5.1/5.4/' lgi/Makefile
```
Aire-One, line 218: I'm not very happy with this. But without this change, the installation target is lua5.1 directories. I guess I have to understand how LGI build works to do something better...

awesome.c
```diff
@@ -84,10 +84,16 @@ init_rng(void)
     lua_getfield(L, -1, "randomseed");
 
     /* Push a seed */
-    lua_pushnumber(L, g_random_int() + g_random_double());
+    lua_pushnumber(L, g_random_int());
```
Aire-One, line 87: This is the fix that was suggested by @Elv13. Other suggestions (https://github.com/awesomeWM/awesome/pull/3124#issuecomment-667464477) had different approach. I don't know what's best.

Aire-One (commented):
Build and tests for Lua 5.4 are green now!

Aire-One:
Everything seems to work here.

As usual, any feedback is welcomed! :)

I think we should solve the issue with LGI release before merging this. As you can see, we need to build and install it manually to have a version that supports Lua 5.4. I'll try to invest some time in https://github.com/pavouk/lgi/pull/280 and https://github.com/pavouk/lgi/issues/261. But not today.

Reading @psychon comment and https://github.com/pavouk/lgi/issues/225 it sounds like it's not going to happen...

.github/workflows/main.yml
psychon: >  guess I have to understand how LGI build works to do something better...

I doubt there is something very much better. Random data points:

- https://github.com/archlinux/svntogit-community/blob/fd394d24539eab3b51ddb7baa6c36cdbc4a149e7/trunk/PKGBUILD#L33-L35
- I tried to figure out what debian does, but failed to actually find it.

Elv13: I usually do

    make PREFIX=/usr LUA_VERSION=5.4 && make install LUA_VERSION=5.4 PREFIX=/usr

That's the proper way of making it work. Obviously the Makefile could/should parse the result of `pkg-config` and figure this out on its own, but this is not a huge issue.

.github/workflows/main.yml
actionless: mb depends how much time does it take

Aire-One: On my computer, I'd says `make test` is nearly instantaneous. On the GitHub Actions, the whole step "Build and Install Lua 5.4" is about 10 seconds.

So I guess we can keep this `make test` instruction :shrug:

.github/workflows/main.yml
Elv13: Well, https://github.com/lunarmodules exists. Also, don't you have admin rights? If so, can you add @aire-one?

Elv13 (approved):

actionless (approved):
thanks!

actionless:
does it builds locally for you on Arch?

i'm getitng:

```
[  3%] Built target generated_icons
[  4%] Generating doc/images/AUTOGEN_uml_nav_tables_tag.svg, /home/lie/build/awesome-git/src/awesome-git/tests/examples/uml/nav_tables/tag.output.txt
[ 82%] Built target generate-examples
make: *** [Makefile:136: all] Error 2
==> ERROR: A failure occurred in build().
    Aborting..
```

with no extra output, lua-lgi-git from latest commit

actionless:
disregard the comment above, it was related to asciidoc

actionless:
i checked, it even starts :))

Issue #3547, opened 2022-01-13
Can't maximize wallpaper using awful.wallpaper


PureArtistry:
Output of `awesome --version`:

awesome v4.3-1336-g21b908bef (Too long)
 • Compiled against Lua 5.3.6 (running with Lua 5.3)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no

**How to reproduce the issue:**

New to awesome, just setting up my config (this could be a pebkac thing, sorry for wasting your time if it is something I'm missing / doing wrong)

I'm trying to set a maximized wallpaper using awful.wallpaper and these settings:
```
screen.connect_signal('request::wallpaper', function(s)
    awful.wallpaper {
        screen = s,
        bg = FLAVOURS.base00,
        widget = {
            {
                image                 = beautiful.wallpaper,
                resize                = true,
                -- upscale               = true,
                -- downscale             = true,
                -- scaling_quality       = 'best',
                horizontal_fit_policy = "fit",
                vertical_fit_policy   = "fit",
                widget                = wibox.widget.imagebox
            },
            valign = 'center',
            halign = 'center',
            tiled  = false,
            widget = wibox.container.tile
        }
    }
end)
```
I have tried every combination of settings and values and the wallpaper is always centered with vertical borders, it appears as if some settings have no impact whatsoever.  

Using gears.wallpaper.maximized() works as expected

**Actual result:**

![scrot_20220113-012344](https://user-images.githubusercontent.com/53883649/149250517-8c3cb94b-a926-480c-97bc-f619f3c481d2.png)

**Expected result:**

![scrot_20220113-012531](https://user-images.githubusercontent.com/53883649/149250656-c8aabf00-6b5e-4538-af45-4a28a4c178ef.png)

Elv13:
Hi, can you try (untested):

```lua
screen.connect_signal('request::wallpaper', function(s)
    awful.wallpaper {
        screen = s,
        bg = FLAVOURS.base00,
        widget = {
            image                 = beautiful.wallpaper,
            upscale               = true,
            downscale             = true,
            horizontal_fit_policy = "fit",
            vertical_fit_policy   = "fit",
            widget                = wibox.widget.imagebox
        }
    }
end)
```

PureArtistry:
[...]

PR #3554, opened 2022-01-24
imagebox: Change fit_policy to singleton

alanswanson:
Having separate "horizontal_fit_policy" and "vertical_fit_policy" properties does not allow maintaining aspect ratio when trying to
fill entirety of destination image, which is preferred for wallpaper usage. While setting "auto" on both kept the aspect ratio it used edge bars and using "fit" just stretched over whole screen warping image.

Switch to a single "fit_property" with following options;
- "auto" (still the default) scales source to fill within destination image using edge bars (if required) to maintain aspect ratio
- "zoom" scales source to fill destination image maintaining aspect ratio by cropping before scaling
- "stretch" (or any other unmatched keyword as this is a non-action in the code) to strech source over the whole destination image
- "vstretch" stretches vertical only
- "hstretch" stretches horizontal only

With "zoom" valign and halign will adjust source image cropping to desired edges - top or left by default but center is usually the better option.

Recently started using awesome so using git with the new awful.wallpaper but hated the bars when scaling images not matching monitors aspect ratio. Issue already reported in #3547. (Last fixed this very same issue for Gnome back in 2004!) Will update commit with documentation updates if this is acceptable.


sclu1034:
Those new options are definitely needed to provide the functionality one would expect from desktop wallpapers.
But I think we can improve the API a bit further:

Both the terms `auto` and `zoom` seem a bit generic to me. Neither of them really tell how the image is changed, exactly.
In my experience, the more common names for these options are `fit` and `fill`, respectively.
And the fact that a "stretch" behaviour is labelled as "fit" in one of the code comments just adds to the confusion.

Additionally, I think we should also change the fact that currently `nil` and `<some truthy, unknown value>` have different behaviour. Usually, you'd expect the default value to cover both of these. So I think this should be changed so that everything except for one of the known other options falls back to `fit`.

So my proposal is:

- `"fit"`, `nil` or unknown value: Scale the image, maintaining aspect ratio, to cover as much as possible without cropping. Will add bars on the smaller side.
- `"fill"`: Scale the image, maintaining aspect ratio, until the entire canvas is covered. Will crop.
- `"stretch"`: Scale and stretch the image to cover the canvas exactly
- `"vstretch"`: Stretch the image vertically
- `"hstretch"`: Stretch the image horizontally

alanswanson:
Made the name changes, _zoom_ to _fill_ and _auto_ to _fit_, but am leaving any unknown value as the "stretch" non-action.

actionless:
it needs also updating docs/examples, but first mb let's get some feedback from @Elv13 since he was implementing those recently AFAIR

actionless:
ping @Elv13

alanswanson:
> if it didn't happened during last ~2.5 years i'm not expecting it to happen later on neither. if that's not the case - feel free to open a new PR which would include aforementioned missing parts

@actionless I assume you're referring to the lack of any review or feedback from @Elv13 after which I would have added the "missing" documentation as originally stated (and which you acknowledged in your comment https://github.com/awesomeWM/awesome/pull/3554#issuecomment-1037753816). At least the similar https://github.com/awesomeWM/awesome/pull/3811 got some inital feedback from @Elv13 but was then followed radio silence and it seems to have been separately closed today.

sclu1034:
> it seems to have been separately closed today

That's likely an accident. That PR was opened from the fork's `master` branch. The same person opened another PR a few hours ago, and likely overwrote their `master` branch when they updated their fork for the new contribution.
And when the branch gets overwritten like that, GitHub closes the PR automatically.

Issue #3555, opened 2022-01-30
Wrong documentation for return type of function gears.color.parse_color()


JGSVb:
Output of `awesome --version`:
awesome v4.3 (Too long)
 • Compiled against Lua 5.3.6 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

**How to reproduce the issue:**
When calling `gears.color.parse_color()` it returns multiple items, not a table, as stated in [gears.color.parse_color()](https://awesomewm.org/doc/api/libraries/gears.color.html#parse_color).

```lua
local t = gears.color.parse_color("#AABBCC") -- not a table, DOES NOT work
```
![image](https://user-images.githubusercontent.com/59124659/151681760-5ae9da02-1f3f-4dd3-8d66-423ecbab0c9d.png)

**Actual result:**
Multiple items are returned instead of a table:

**Expected result:**
It is expected that the function returns a table, so it would be possible to iterate over all color channels (for example).
```lua
-- >>this DOES NOT work<<
local t = gears.color.parse_color("#AABBCC")
-- usage example
for i,v in ipairs(t) do t[i] = t[i] * 2 end
```

sclu1034:
To reduce API breakage, the current return values should not be changed. Only the docs should be aligned to explain that properly, i.e. turn the single `table` return value into four `number` values.

If a table is required, it can always be build by the consumer with `table.pack` or manually writing the literal.

Aire-One:
I think it is resolved now that #3564 is merged.

Issue #3432, opened 2021-09-09
default theme's tiletop.png layout icon is sized incorrectly.


Expletive:
tiletop.png is 56x56 pixels in size. All the other icons are 64x64. tiletop.png is missing the transparent border that the other icons have.

Output of `awesome --version`:

awesome v4.3 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
**How to reproduce the issue:**

Enable the awful.layout.suit.tile.top layout while using the tiletop.png as the layout's icon in your theme.

**Actual result:**
The icon is stretched and noticeably different from the other icons.

**Expected result:**
The icon should look the same as the other icons.

PR #3558, opened 2022-02-05
Update documentation for gears.color.parse_color

pepijnKrijnsen:
Comments now accurately reflect the return value of the parse_color(col) method - issue #3555

Changed the tiletop.png file in themes/default/layout to 64x64 px to match the other layout PNGs in the same directory - issue #3432


lib/gears/color.lua
```diff
@@ -59,7 +59,7 @@ local pattern_cache
 -- Max 4 chars per channel.
 --
 -- @param col The color to parse
--- @treturn table 4 values representing color in RGBA format (each of them in
+-- @return 4 values representing color in RGBA format (each of them in
```
Aire-One, line 62: ldoc supports to use multiple `@return`/`@treturn` tags. Please document all 4 returned value with their respective types.

Aire-One (commented):
Thanks for your contribution! 

Can you please move your second commit to a separate branch/PR? These are 2 different topics.

lib/gears/color.lua
pepijnKrijnsen: I wasn't able to revert the second commit, or move it to a separate PR. Please feel free to close this PR. I will create two separate new PRs for the edits, considering also your comment above regarding return tags in ldoc.

PR #3489, opened 2021-10-29
shape: Add a "solid shadow" shape.

Elv13:
Credit to @actionless for originally using it in some of his themes.

image

In my case, I originally wrote it for this MS-DOS homage demo https://imgur.com/a/8ikCPl0

I upstream it now because it got some little interest in the chat.

PR #3559, opened 2022-02-05
Add a new border container

Elv13:
This container have 2 main aspect. First, it implement CSS-style slices. So background images stretch properly:

image

**note**: Ignore the visible artifacts in the image. It's only in the doc because it places raster images in SVG and rescale them incorrectly when the browser scales the SVG.

The second feature is the ability to have side and corner widgets for the border.

Initially, I didn't plan to finish this for this release, but some people shown interest in the chat. There was also some questions in https://github.com/awesomeWM/awesome/pull/3489 about how complex titlebars can be handled without 400 lines of ugly hacks. With this container, you can directly take a background image and end up with this:

image

This is the first part of a 2 part series to fix https://github.com/awesomeWM/awesome/issues/650#issuecomment-312502332 . I will not do the second part before v4.4. The other half is mostly the productionalized version of @psychon code. The new `awful.decoration` module will not really add size/corner resize by default. However, it will now be trivial to add rather than stupidly hard with 2 `wibox.layout.align` and 2 `wibox.layout.flex` like it is right now.

I thought "it's almost fully implemented, how bad can it be to finish it!": very bad... This container was a corner case rabbit hole. I kind of gave up in December, but finally had the guts to finish it. The DPI property is still not exposed. It will take more rabbit hunting down in the rabbit hole.

About the implementation. I hesitated between putting this in `wibox.layout` or `wibox.container`. In the end I picked container because it only have 1 "indexed" widget. This deprecated `wibox.container.background.bgimage`. That's "weird" because it's deprecating the background support for the background container. However, the shape and bgimage were always pretty much incompatible. Extending the `background` container with all the new properties would have made the code unmaintainable. I added a couple new properties to the `background` container to compensate for some of the deprecated corner case which could not be handled by the border container.

edit: I also fix a few CSS support bugs. I added CSS support in the wallpaper PR serie, but it wasn't that well tested

### DIFF
--- a/lib/wibox/container/border.lua
+++ b/lib/wibox/container/border.lua
@@ -1,0 +1,1050 @@
+---------------------------------------------------------------------------
+-- Place widgets or images on the sides, corner and back of another widget.
+--
+-- This widget main use case is to surround another widget with an uniform
+-- border. It can also be used to create soft shadows. It support CSS style
+-- image slicing or widgets within the border. The `awful.layout.align` layout
+-- an also be used to achieve border, but make it very hard to make them
+-- symmetric on both axis.
+--
+-- Note that because of legacy reasons, `wibox.container.background` also has
+-- good support for borders. If you only need simple shaped strokes, the,
+-- the `background` container is a much better choice. This module is better
+-- suited for background images and border widgets.
+--
+-- Advanced usage
+-- ==============
+--
+-- In this first example, 2 `wibox.container.borders` are combined in a
+-- `wibox.layout.stack`. The bottom one is the "real" border and the top
+-- layer is used to place some widgets. This is useful to built titlebars.
+--
+-- @DOC_wibox_container_border_dual1_EXAMPLE@
+--
+-- This example demonstrates how to use this module to create a client
+-- border with a top titlebar and borders. It does so by embedding a
+-- `wibox.container.border` into another `wibox.container.border. The outer
+-- container acts as the border around the central area while the inner one
+-- spans the top area and contains the titlebar itself. The outer border
+-- widgets area can be used to implement border resize.
+--
+-- @DOC_wibox_container_border_titlebar1_EXAMPLE@
+--
+-- This container is like every other widgets. It is possible to override the
+-- Cairo drawing functions. This is very useful here to create custom borders
+-- using cairo while still re-using the internal geometry handling code.
+--
+-- @DOC_wibox_container_border_custom_draw1_EXAMPLE@
+--
+--@DOC_wibox_container_defaults_border_EXAMPLE@
+--
+-- @author Emmanuel Lepage-Vallee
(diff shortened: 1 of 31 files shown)


actionless:
havent went through the code yet, but description looks nice 👍

lib/wibox/container/border.lua
```diff
@@ -0,0 +1,1050 @@
...
+--
+-- This widget main use case is to surround another widget with an uniform
+-- border. It can also be used to create soft shadows. It support CSS style
+-- image slicing or widgets within the border. The `awful.layout.align` layout
+-- an also be used to achieve border, but make it very hard to make them
+-- symmetric on both axis.
+--
+-- Note that because of legacy reasons, `wibox.container.background` also has
+-- good support for borders. If you only need simple shaped strokes, the,
+-- the `background` container is a much better choice. This module is better
```
Aire-One, line 12:
```suggestion
-- good support for borders. If you only need simple shaped strokes, the
-- `background` container is a much better choice. This module is better
```

lib/wibox/container/border.lua
```diff
@@ -0,0 +1,1050 @@
...
+--
+-- In this first example, 2 `wibox.container.borders` are combined in a
+-- `wibox.layout.stack`. The bottom one is the "real" border and the top
+-- layer is used to place some widgets. This is useful to built titlebars.
+--
+-- @DOC_wibox_container_border_dual1_EXAMPLE@
+--
+-- This example demonstrates how to use this module to create a client
+-- border with a top titlebar and borders. It does so by embedding a
+-- `wibox.container.border` into another `wibox.container.border. The outer
```
Aire-One, line 26:
```suggestion
-- `wibox.container.border` into another `wibox.container.border`. The outer
```

lib/wibox/container/border.lua
```diff
@@ -0,0 +1,1050 @@
...
+local function fit_common(widget, ctx, max_w, max_h)
+    if not widget then return 0, 0 end
+
+    local w, h = widget:fit(ctx, max_w or math.huge, max_h or math.huge)
+
+    return w, h
+end
+
+local function uses_slice(self)
+    return (self._private.border_widgets or self._private.border_image_widgets) == nil
```
Aire-One, line 85:
```suggestion
    return not (self._private.border_widgets or self._private.border_image_widgets)
```

Sounds easier to read. Or am I missing something?

Aire-One (commented):
I'll try to do a second pass on the border container code latter. It's getting too late for all the math :sweat_smile: 

This is very promising regarding titlebars and borders for clients. I can't wait for the second part of the series!

actionless (dismissed):
it doesn't look potentially harmful as it's not changing much of stuff, but rather adding it, so i think it's ok to go if it's still passing the CI

Elv13:
Merging since this was approved and I only rebased/fixed_new_doc_warnings

PR #3560, opened 2022-02-06
Update documentation for gears.color.parse_color(col) method

pepijnKrijnsen:
Lua comments now accurately describe the return values of this method.
Added a more comprehensive example, too. Issue #3555

### DIFF
--- a/lib/gears/color.lua
+++ b/lib/gears/color.lua
@@ -59,11 +59,17 @@ local pattern_cache
 -- Max 4 chars per channel.
 --
 -- @param col The color to parse
--- @treturn table 4 values representing color in RGBA format (each of them in
--- [0, 1] range) or nil if input is incorrect.
+-- @treturn number between 0 and 1 for the 'red' value (1st channel)
+-- @treturn number between 0 and 1 for the 'green' value (2nd channel)
+-- @treturn number between 0 and 1 for the 'blue' value (3rd channel)
+-- @treturn number between 0 and 1 for the 'opacity' value (4th channel)
+-- if the incoming color code only has 3 values (only rgb, not opacity) the 4th
+-- return value is 1.
+-- @treturn nil if input is incorrect
 -- @staticfct gears.color.parse_color
--- @usage -- This will return 0, 1, 0, 1
--- gears.color.parse_color("#00ff00ff")
+-- @usage -- Both of the following will return 0, 0.4, 0.8, 1
+-- gears.color.parse_color("#0066ccff")
+-- gears.color.parse_color("#0066cc")
 function color.parse_color(col)
     local rgb = {}
     if string.match(col, "^#%x+$") then


actionless (approved):

lib/gears/color.lua
```diff
@@ -59,11 +59,17 @@ local pattern_cache
...
 -- @param col The color to parse
--- @treturn table 4 values representing color in RGBA format (each of them in
--- [0, 1] range) or nil if input is incorrect.
+-- @treturn number between 0 and 1 for the 'red' value (1st channel)
+-- @treturn number between 0 and 1 for the 'green' value (2nd channel)
+-- @treturn number between 0 and 1 for the 'blue' value (3rd channel)
+-- @treturn number between 0 and 1 for the 'opacity' value (4th channel)
+-- if the incoming color code only has 3 values (only rgb, not opacity) the 4th
+-- return value is 1.
+-- @treturn nil if input is incorrect
```
Aire-One, line 68: Hm... I'm not sure whether this is a good way to present the nil value returned in case of error. I don't know how to represent that better, tho. So let's keep it this way for now :shrug:

Aire-One (approved):
Thank you!

lib/gears/color.lua
pepijnKrijnsen: I know what you mean - there should be an ldoc tag to indicate what a function returns in case of unexpected input.

Would you prefer to leave “nil in case of invalid input” out of the doc entirely? If someone calls parse_color with the wrong value, the error message will indicate that they’re trying to use a nil value anyway.

Aire-One: Hm, yes, you are right. Let's remove it!

> there should be an ldoc tag to indicate what a function returns in case of unexpected input.

We could do it from our `config.ld` thanks to @Elv13 magic. But this would add a layer of complexity into account in the documentation HTML template and, as a whole, in our user facing API. I don't feel like this is necessary to implement such ldoc tag for 99% of the scenarios where a function is just supposed to return some value and if it fails because of user giving wrong values, it fails anyway.

Hm... Now that I think of it, maybe instead of returning nil, this function should simply invoke error (https://www.lua.org/manual/5.1/manual.html#pdf-error)

Anyway, I think this is out of the scope for this PR.

Aire-One (dismissed):

lib/gears/color.lua
sclu1034: > Hm... Now that I think of it, maybe instead of returning nil, this function should simply invoke error

Awesome's API never panics. So we shouldn't start doing that somewhere out of the blue.

Even the fact that this prints an error on invalid input is somewhat rare.

pepijnKrijnsen:
Not sure what happened on this PR - I tried to push a separate little documentation correction but it looks like it got mixed into this PR. Apologies! If a seasoned contributor could untangle this that’d be awesome - if I need to just cancel the PR and try again I’m happy to do that. @Aire-One @actionless

actionless:
might be easier to branch anew from upstream's master and cherry-pick the needed commits from the messed up origin's master

sclu1034:
Agreed.
Assuming the usual `origin` + `upstream` remote setup:

```sh
git fetch upstream
git branch parse_color upstream/master
git checkout parse_color
git cherry-pick d05d44a
git push origin parse_color
# Git will show an URL where you can open the new PR

git checkout master
git branch documentation-correction upstream/master
git checkout documentation-correction
git cherry-pick 0c46d4f
git push origin documentation-correction
# Git will show an URL where you can open the new PR
```

PR #3566, opened 2022-02-13
theme: remove mouse finder remnants

staticssleever668:
Was searching for `animate` in the code base, was surprised to stumble upon something I've never heard of and was eager to try it, but it appears that this functionality was removed back in 2016 but had some code left behind. :)

### DIFF
--- a/themes/default/theme.lua
+++ b/themes/default/theme.lua
@@ -38,7 +38,6 @@ theme.border_color_marked = "#91231c"
 -- tasklist_[bg|fg]_[focus|urgent]
 -- titlebar_[bg|fg]_[normal|focus]
 -- tooltip_[font|opacity|fg_color|bg_color|border_width|border_color]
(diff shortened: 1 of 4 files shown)


Aire-One:
According to https://awesomewm.org/apidoc/documentation/89-NEWS.md.html, the `awful.mouse_finder` module have been removed at the v4.0 release:

> The mouse finder module is gone
> 
> It has been broken for ages, so we concluded nobody cared.

The commit seems to be b73ebabb. 

(it's older than my time with awesome, that's why I never heard of it before 🙃)

I guess these theme variables have been forgotten, and nobody touched them since then because it is legacy code that no-one knows about...

Anyway, I think we're safe to remove that. (I'll add a little CC @psychon for double-check purposes.)

Aire-One (approved):

psychon (approved):

Issue #3557, opened 2022-02-04
Tooltip placement error 


Trial97:
Output of `awesome --version`:

```
awesome v4.3-1348-g2d6244f94-dirty (Too long)
 • Compiled against Lua 5.4.3 (running with 0.9.2)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: /usr/share/lua/5.4/lgi/version.lua
 • Transparency enabled: yes
 • Custom search paths: no
```
**How to reproduce the issue:**

Wanted to have a tooltip for a widget  with the following:
```
awful.tooltip({
        objects= {widget_button},
        mode = 'outside',
        text = 'text'
})
```
Then move the cursor over the widget to display the tooltip

**Actual result:**

I'm getting an error regarding a bad argument:
```
2022-02-04 09:50:57 W: awesome: luaA_dofunction:78: error while running function!
stack traceback:
	[C]: in field 'field'
	/usr/share/lua/5.4/lgi/record.lua:118: in function </usr/share/lua/5.4/lgi/record.lua:92>
	(...tail calls...)
	[C]: in metamethod 'newindex'
	/usr/share/lua/5.4/lgi/record.lua:178: in function </usr/share/lua/5.4/lgi/record.lua:162>
	(...tail calls...)
	/usr/share/awesome/lib/awful/placement.lua:823: in upvalue 'fit_in_bounding'
	/usr/share/awesome/lib/awful/placement.lua:1817: in function </usr/share/awesome/lib/awful/placement.lua:1740>
	(...tail calls...)
	/usr/share/awesome/lib/awful/tooltip.lua:175: in upvalue 'apply_outside_mode'
	/usr/share/awesome/lib/awful/tooltip.lua:201: in upvalue 'set_geometry'
	/usr/share/awesome/lib/awful/tooltip.lua:221: in upvalue 'show'
	/usr/share/awesome/lib/awful/tooltip.lua:772: in local 'func'
	/usr/share/awesome/lib/gears/object.lua:160: in function 'gears.object.emit_signal'
	/usr/share/awesome/lib/wibox/drawable.lua:329: in upvalue 'emit_difference'
	/usr/share/awesome/lib/wibox/drawable.lua:350: in upvalue 'handle_motion'
	/usr/share/awesome/lib/wibox/drawable.lua:443: in function </usr/share/awesome/lib/wibox/drawable.lua:443>
[...]

PR #3572, opened 2022-02-16
fix(placement) bad argument to 'field'

sarubo:
I'm not good at English. If you have any problems with my English, please ask me.

I also encountered the same issue as #3557 . It is a request to fix this.

RectangleInt must use integer.
https://github.com/awesomeWM/awesome/blob/bd776c98015767e156556e6651da4241b657d5bb/lib/awful/placement.lua#L800-L813
https://github.com/awesomeWM/awesome/blob/bd776c98015767e156556e6651da4241b657d5bb/lib/awful/placement.lua#L1553-L1562
So this issue was thought to be caused by the following division (`/`).
https://github.com/awesomeWM/awesome/blob/bd776c98015767e156556e6651da4241b657d5bb/lib/awful/placement.lua#L268-L281
Just rounding this off solved my issue. (ex: `gmath.round(r.y-h/2+r.height/2)`)

However, when I was investigating the cause, I applied `force_width=12.5` to my widget and still got the same error.

I rounded the geometry to deal with errors in a wide range. All errors when building are gone.

### DIFF
--- a/lib/awful/placement.lua
+++ b/lib/awful/placement.lua
@@ -93,6 +93,7 @@ local floating = require("awful.layout.suit.floating")
 local a_screen = require("awful.screen")
 local grect = require("gears.geometry").rectangle
 local gdebug = require("gears.debug")
+local gmath = require("gears.math")
 local gtable = require("gears.table")
 local cairo = require( "lgi" ).cairo
 local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
@@ -797,19 +798,31 @@ local function get_relative_regions(geo, mode, is_absolute)
     return regions
 end
 
+local function round_geometry(geo)
+    local geo_keys = { "x", "y", "height", "width" }
+    local rounded_geo = {}
+    for _, key in pairs(geo_keys) do
+        if geo[key] ~= nil then
+            rounded_geo[key] = gmath.round(geo[key])
+        end
+    end
+    return rounded_geo
+end
+
 -- Check if the proposed geometry fits the screen
 local function fit_in_bounding(obj, geo, args)
-    local sgeo   = get_parent_geometry(obj, args)
-    local region = cairo.Region.create_rectangle(cairo.RectangleInt(sgeo))
+    local round_sgeo = round_geometry(get_parent_geometry(obj, args))
+    local region     = cairo.Region.create_rectangle(cairo.RectangleInt(round_sgeo))
 
+    local round_geo = round_geometry(geo)
     region:intersect(cairo.Region.create_rectangle(
-        cairo.RectangleInt(geo)
+        cairo.RectangleInt(round_geo)
     ))
 
     local geo2 = region:get_rectangle(0)
 
     -- If the geometry is the same then it fits, otherwise it will be cropped.
-    return geo2.width == geo.width and geo2.height == geo.height
+    return geo2.width == round_geo.width and geo2.height == round_geo.height
 end
 
 -- Remove border from drawable geometry


sarubo:
I forgot about something other than lua 5.4. I'll fix it.

Aire-One:
ahah! I came to the same conclusion with my investigation. Nice to see my findings can be double-checked :wink: 

I'm still a little surprised by the fact it works without issue on previous Lua version. Also, from the lgi doc (https://github.com/lgi-devs/lgi/blob/master/docs/cairo.md):

> Gdk.Rectangle is just a link to cairo.RectangleInt (similar to C, where GdkRectangle is just a typedef of cairo_rectangle_int_t). gdk_rectangle_union and gdk_rectangle_intersect are wired as a methods of Gdk.Rectangle as expected.

And we can confirm:

```lua
-- https://github.com/lgi-devs/lgi/blob/9eaad42eec50baa3e24f3420c35107d6d295577e/lgi/override/cairo.lua#L108-L128

   {  'Rectangle',
      fields = {
	 { 'x', ti.double }, { 'y', ti.double },
	 { 'width', ti.double }, { 'height', ti.double },
      },
   },

-- ...

   {  'RectangleInt',
      fields = {
	 { 'x', ti.int }, { 'y', ti.int },
	 { 'width', ti.int }, { 'height', ti.int },
      },
   },
```

So maybe this issue is from lgi? :shrug:

> However, when I was investigating the cause, I applied force_width=12.5 to my widget and still got the same error.

That's a nice finding. I would have proposed to round the results of `outer_positions` functions. With this in mind, I agree it's a better solution to round the geometries before calling `cairo.RectangleInt`.

There is something else I want to check on this topic: The CI doesn't fail while I can consistently reproduce the issue with the `tests/examples/awful/tooltip/preferred_alignment2.lua` example. This has to be investigated as well.

lib/awful/placement.lua
```diff
@@ -800,8 +801,18 @@ end
...
+            sgeo[key] = gmath.round(value)
+        end
+    end
     local region = cairo.Region.create_rectangle(cairo.RectangleInt(sgeo))
 
+    for key, value in pairs(geo) do
+        if type(value) == "number" then
+            geo[key] = gmath.round(value)
+        end
+    end
```
Aire-One, line 815: This applies changes to the geo table we get from parameters. I think it's a bad practice.

lib/awful/placement.lua
```diff
@@ -800,8 +801,18 @@ end
 -- Check if the proposed geometry fits the screen
 local function fit_in_bounding(obj, geo, args)
     local sgeo   = get_parent_geometry(obj, args)
+    for key, value in pairs(sgeo) do
+        if type(value) == "number" then
+            sgeo[key] = gmath.round(value)
+        end
+    end
```
Aire-One, line 808: The data structure for `cairo.RectangleInt` parameter is `{ x, y, height, width }`. I think we better have a utility function `round_geometries(geo) -> rounded_geo`. (It would also prevent code duplication for the similar loops you added)

sarubo: Thanks for the review!
```lua
local function round_geometry(geo)
    local geo_keys = { "x", "y", "height", "width" }
    local rounded_geo = {}
    for _, key in pairs(geo_keys) do
        rounded_geo[key] = gmath.round(geo[key])
    end
    return rounded_geo
end

-- Check if the proposed geometry fits the screen
local function fit_in_bounding(obj, geo, args)
    local round_sgeo   = round_geometry(get_parent_geometry(obj, args))
    local region = cairo.Region.create_rectangle(cairo.RectangleInt(round_sgeo))

    local round_geo = round_geometry(geo)
    region:intersect(cairo.Region.create_rectangle(
        cairo.RectangleInt(round_geo)
    ))

    local geo2 = region:get_rectangle(0)

    -- If the geometry is the same then it fits, otherwise it will be cropped.
    return geo2.width == round_geo.width and geo2.height == round_geo.height
end
```
Suppose I write as above.
For `return`, is it better to use `round_geo` or `geo`?

Aire-One: I think you should use `rounded_geo` since `geo2` is obtained from `region` that was generated from `round_sgeo`.

Side note : I think we should have a test on `geo[key]` before the assignation in round_geometry.

sarubo: > For `return`, is it better to use `round_geo` or `geo`?

I checked and found that the value of `geo2` is an integer, so I used `round_geo`.
It worked well in local. I force push it.

sarubo: I missed the reply. I will add a test.

sarubo:
```lua
    for _, key in pairs(geo_keys) do
        assert(type(geo[key]) == "number")
        rounded_geo[key] = gmath.round(geo[key])
    end
```
I put in an assertion because I think the default value creates confusion.

Aire-One: Hm... I was thinking more of something like `if geo[key] ~= nil then` to allow the function to work with "partial geometries"

sarubo: Oh, I see. I'll do it. Thank you.

Elv13 (approved):

actionless (approved):
thanks for looking into this!

PR #3575, opened 2022-02-17
[CI] Run `Run examples tests` on every supported Lua Version

Aire-One:
Hello,

I don't know the exact reason why we disabled this step in the first place.

The idea is to run the examples for every supported Lua version. With some luck, we would have seen the bug from https://github.com/awesomeWM/awesome/issues/3557 before.

This PR includes a fix about the Lua integer representation to force the output of the `tests/examples/wibox/widget/textbox/font1.lua` example to produce non-floating integers, regardless of the Lua version.

~~The remaining failure in Lua5.4 Job is what PR https://github.com/awesomeWM/awesome/pull/3572 fixes.~~ Now it's merged, I have rebase

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -305,7 +305,6 @@ jobs:
         run: cd "${{ github.workspace }}/build" && make check-unit
 
       - name: Run examples tests
-        if: matrix.coverage
         run: cd "${{ github.workspace }}/build" && make check-examples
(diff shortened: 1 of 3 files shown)


actionless (dismissed):

actionless:
i guess you need to rebaase now?

Elv13 (approved):

PR #3585, opened 2022-03-10
hotkeys_popup: expose labels and remove unused backspace translation

actionless:
the actual key is BackSpace, not Backspace, so this translation was always unused

### DIFF
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -40,7 +40,7 @@
 --
 -- Example of having different types of hotkey popups:
 --
---    local globalkeys = awful.util.table.join(
+--    awful.keyboard.append_global_keybindings({
 --        awful.key({modkey}, "/", function()
 --          hotkeys_popup.show_help()
 --        end, nil, {
@@ -56,8 +56,8 @@
 --        end, nil, {
 --          description = "show help for awesome only", group="HELP"
 --        })
---        -- (more hotkeys goes here)
---    )
+--        -- (more hotkeys go here)
+--    })
 --
 --
 -- @author Yauheni Kirylau &lt;yawghen@gmail.com&gt;
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -196,7 +196,6 @@ function widget.new(args)
             Mod4             = "Super",
             Insert           = "Ins",
             Delete           = "Del",
-            Backspace        = "BackSpc",
             Next             = "PgDn",
             Prior            = "PgUp",
             Left             = "←",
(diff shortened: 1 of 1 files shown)


lib/awful/hotkeys_popup/widget.lua
```diff
@@ -37,6 +37,29 @@
...
+--        end, nil, {
+--          description = "show help for current app", group="HELP"
+--        }),
+--        awful.key({altkey, modkey}, "/", function()
+--          hotkeys_popup.show_help({}, nil, {show_awesome_keys=true})
+--        end, nil, {
+--          description = "show help for awesome only", group="HELP"
+--        })
+--        -- (more hotkeys goes here)
+--    )
```
Aire-One, line 60: Maybe you should move these code examples to the examples. So that it gets syntax checking and is run at every CI build to check for breakage.

Aire-One (dismissed):
Thank you for this change! I'm happy we have an easy way to alter only some labels now that the complete table is exposed.

lib/awful/hotkeys_popup/widget.lua
actionless: that wouldn't run any code, as those are callbacks

sclu1034: You can run those callbacks by calling `trigger` on the key object.
You'd just have to hide the calls and setup for it to keep the visible portion of the example the same.

sclu1034, line 60: `awful.util.table` has been deprecated in favour of `gears.table`.
And the pattern `root.keys(globalkeys)` has been deprecated in favour of `awful.keyboard.append_global_keybindings(keys)` (where `keys` is a simple list, not `join`ed).

```suggestion
--    awful.keyboard.append_global_keybindings({
--        awful.key({modkey}, "/", function()
--          hotkeys_popup.show_help()
--        end, nil, {
--          description = "show help (all)", group="HELP"
--        }),
--        awful.key({"Shift", modkey}, "/", function()
--          hotkeys_popup.show_help(nil, nil, {show_awesome_keys=false})
--        end, nil, {
--          description = "show help for current app", group="HELP"
--        }),
--        awful.key({altkey, modkey}, "/", function()
--          hotkeys_popup.show_help({}, nil, {show_awesome_keys=true})
--        end, nil, {
--          description = "show help for awesome only", group="HELP"
--        })
--        -- (more hotkeys goes here)
--    })
```

sclu1034 (changes requested):

lib/awful/hotkeys_popup/widget.lua
actionless: feel free to add it if you want

Aire-One (approved):

PR #3588, opened 2022-03-15
Remade alignment code and added a new expand mode for wibox.layout.align container

nenikitov:
# Changes

- Rewrote align method so it would be easier to add more expand modes in future if needed
- Added a new expand mode called `"justified"`
    - Gets the minimum required size of the widgets on the sides
    - Uses the size of the largest side widget on both sides
    - Expands the second, center, widget to fill the remaining space

# TODO

- [x] Implement
- [x] Docs
- [x] Manual testing
- [ ] Unit testing

I am not really sure how the unit tests work and how can I run them; I manually compared how the old and the new code align the widgets and it seems to be working the same.

The new feature is completely done, unless my changes break something and I need to fix it.

# Demo of new align expand mode

https://user-images.githubusercontent.com/44075969/158315404-150cd983-5565-462f-a99c-9b89b790ee0a.mp4


lib/wibox/layout/align.lua
```diff
@@ -200,7 +278,6 @@ end
 -- @property third
 -- @tparam widget third
 -- @propemits true false
-
```
sclu1034, line 203: We don't usually add empty lines between doc comments and the things they document. So if you do want to make this consistent, you should remove the empty line at `set_first` instead.

nenikitov:
Here is a quick update on the progress:
* I was only able to run `make check-unit`. For me, `make check` stops at Xephyr part with error `Error: failed to X resources are ready!`.
* I rewrote a good chunk of my code while addressing all your comments:
    * Now the functions `fit-widget` and `place_widget_at` are called in the exact same order as they were previously before my changes. All the unit tests pass now, except the ones that I wrote for the new `"justify"` mode.
    * The code for the `"inside"` expand mode is now in the `else` statement.
    * I removed the empty line in between the doc comment and the code in the file I am working on. I also found more empty lines like this in the other files ([example](https://github.com/awesomeWM/awesome/blob/master/lib/awful/popup.lua#L288)) with `^-- @.*\n$\n+(function|local)`. It matched 191 times, but it probably didn't catch everything and / or caught the empty lines that should be there.

Stuff left to do:
* Figure out why I can't run `make check` and make sure everything passes.
* Unit test my code.
* Make sure that I didn't leave any old logic in the comments.

lib/wibox/layout/align.lua
```diff
@@ -41,121 +43,297 @@ local align = {}
...
+    -- Place a child widget inside the layout
+    local function place_child(child, offset, size, update_size_remains)
+        -- Can't place a widget if it does not exist or there is no more size
+        if not self._private[child] or size_remains <= 0 then
+            return
+        end
+
+        offset = floor(offset)
+        if not size then
+            size = size and floor(size) or size_remains
```
Aire-One, line 75: `size` is falsy in this branch, so the first test line 75 is always falsy. I'm not sure what `floor(size)` (aka `floor(falsy)`) would return here...

lib/wibox/layout/align.lua
```diff
@@ -41,121 +43,297 @@ local align = {}
...
-        else
-            -- the middle widget is sized first, the outside widgets are given
-            --  the remaining space if available we will draw later
-            size_remains = floor((size_remains - size_second) / 2)
+    local is_vert = self._private.dir == "y"
+
+    -- Get the required size of a child widget
+    local function fit_child(child, max_size)
+        -- Can't fit a widget if it does not exist
+        if not self._private[child] then
```
Aire-One, line 52: I'm not a fan of these `self._private[child]` access. `child` can be anything, you have to use literal strings, and that's very error-prone.

lib/wibox/layout/align.lua
```diff
@@ -266,6 +441,9 @@ end
 --   widgets are then given the remaining space on either side.
 --   If the center widget requires all available space, the outer widgets are
 --   not drawn at all.
+-- * `"justified"`: The widgets in the slot one and three are set to the same
```
Aire-One, line 444: I don't know where to put this comment, so let's say this line is the place to open the discussion:

I don't like this much to manipulate literal strings, and I think an "enum like table" to maps the available modes would be more user-friendly. I did one for the mouse buttons, for example https://github.com/awesomeWM/awesome/blob/392dbc21ab6bae98c5bab8db17b7fa7495b1e6a5/lib/awful/button.lua#L42-L48.

lib/wibox/layout/align.lua
```diff
@@ -274,9 +452,8 @@ end
 --
 -- @property expand
 -- @tparam[opt="inside"] string mode How to use unused space.
-
 function align:set_expand(mode)
-    if mode == "none" or mode == "outside" then
+    if mode == "none" or mode == "outside" or mode == "justified" then
```
Aire-One, line 456: That makes a lot of possible modes. I think something functional like this would be more readable:

```lua
if gears.table.has_item({
    "none",
    "outside",
    "justified",
}, mode) then
```

(bonus point with the "enum like table" to handles modes: this test would be always up -to-date)

lib/wibox/layout/align.lua
```diff
@@ -41,121 +43,297 @@ local align = {}
...
-                w, _ = base.fit_widget(self, context, self._private.third, size_remains, height)
-                if self._private.expand == "inside" then
-                    size_remains = size_remains - w
-                end
-            end
+
+    -- Size of the layout before any placements
+    local size_total = size_remains
+
+    if self._private.expand == "none" then
```
Aire-One, line 104: That's not your work here, but while we refactor this code: I think I'd like to see each branch of this huge state-machine to be resolved in separates utils functions to make the code easier to read.

(The association of modes and "sub-layout" functions could even be done with some sort of polymorphism :thinking:)

Aire-One (commented):
I have only reviewed the `lib/wibox/layout/align.lua` file. It's getting late here, I'll come back latter this week to do a second review.

While I have some comments regarding the design of your code, I really like the local closures you wrote to flit and place child widgets. The code feel easier to understand than what I recall of it.

Also, please only commit your changes. It seems you have duplicated files and extra files generated by tests.

Aire-One:
In addition to unit tests, I'd like to also see examples to show the difference between all the layout modes. That's probably not something we want to ask for this PR, tho...

lib/wibox/layout/align.lua
```diff
@@ -41,121 +43,297 @@ local align = {}
...
-        size_second = self._private.dir == "y" and h or w
-        -- if all the space is taken, skip the rest, and draw just the middle
-        -- widget
-        if size_second >= size_remains then
-            return { base.place_widget_at(self._private.second, 0, 0, width, height) }
-        else
-            -- the middle widget is sized first, the outside widgets are given
-            --  the remaining space if available we will draw later
-            size_remains = floor((size_remains - size_second) / 2)
+    local is_vert = self._private.dir == "y"
```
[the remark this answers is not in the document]
nenikitov: Did it in [135369c](https://github.com/awesomeWM/awesome/pull/3588/commits/135369c8e472c084806079c2353349c27169aa85).

lib/wibox/layout/align.lua
nenikitov: It is a left over from when I was trying to make it without the `if` block, I didn't notice that should be changed. Updated it in [135369c](https://github.com/awesomeWM/awesome/pull/3588/commits/135369c8e472c084806079c2353349c27169aa85).

lib/wibox/layout/align.lua
nenikitov: I tweaked how the functions work in https://github.com/awesomeWM/awesome/pull/3588/commits/d5575da0e53f65ab7c5ad55c3fad0f9e9eda6ebf). Now they accept the reference to the child widget instead of its name.

lib/wibox/layout/align.lua
nenikitov: True, now it is like this after [b37ba4e](https://github.com/awesomeWM/awesome/pull/3588/commits/b37ba4e4a54266187492729c2d42691c54ffddc2)

lib/wibox/layout/align.lua
nenikitov: Maybe. Another possible solution would be to create a table where the keys are expand modes and the values are their corresponding callback util functions. I imagine this would work well with the enum like table made in [b37ba4e](https://github.com/awesomeWM/awesome/pull/3588/commits/b37ba4e4a54266187492729c2d42691c54ffddc2)

lib/wibox/layout/align.lua
```diff
@@ -41,121 +43,297 @@ local align = {}
...
+
+    -- Get the required size of a child widget
+    local function fit_child(child, max_size)
```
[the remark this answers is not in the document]
nenikitov: Do I understand correctly that, while moving `fit` and `place` functions outside of the scope of `align:layout`, I should still keep them as `local function` so they will not be accessible outside of the file? Or should I put them inside `align` object?

nenikitov:
Thanks for your review.

I implemented a big part of the changes requested in your comments.

I still need to:
* Move the functions in the higher scope (I had a question about it, it is [here](https://github.com/awesomeWM/awesome/pull/3588#discussion_r828739748))
* Finish with the unit tests for my expand mode
* Figure out why I can't run `make check`

> Also, please only commit your changes. It seems you have duplicated files and extra files generated by tests.
* I removed the duplicate `align_f.lua` and `align_o.lua` files in [943f447](https://github.com/awesomeWM/awesome/pull/3588/commits/943f447b9a2caf48b3f9b9a7010a1eaa6f163529) since I don't need them anymore. They were the iterations on refactoring of the previous code that I kept for reference.
* Do you mean the files inside `test/examples/` directory?
* Should I also remove the generated files in the `docs/common/`?
* I am not really sure how should I delete these files. I found that running `git checkout upstream/master -- path/to/file` would restore them how they were in the original repo. Is this what I need to do?

sclu1034:
> Do you mean the files inside `test/examples/` directory?
> Should I also remove the generated files in the `docs/common/`?

The only files that should show up in any of your commits so far should be:
```
lib/wibox/layout/align.lua
spec/wibox/layout/align_spec.lua
```

> I am not really sure how should I delete these files. I found that running `git checkout upstream/master -- path/to/file` would restore them how they were in the original repo. Is this what I need to do?

That would allow you to add a commit that reverts those files (similar to the one where you removed `align_f.lua` and `align_o.lua`) to their state before this PR. But it doesn't change the commits you already made.
You will have to `git rebase` your commits and `git commit --amend` them individually to make sure those extraneous files are not added to the commit history in the first place.

And while you're rebasing, it would probably make sense to squash your various iterations on `align.lua`, such that all the intermediary stuff (e.g. one of the commits adds a large chunk of commented code) is gone.
Once the CI pipeline has been enabled for this PR, it will check every commit individually, so every commit needs to leave the code base in a state that passes the test suite. Those commit where you leave some `elseif self._private.expand == "<something>" then` branches empty will definitely break.

lib/wibox/layout/align.lua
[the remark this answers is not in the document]
sclu1034: Yes, they should be `local` functions. They make little sense outside `align:layout()`, so there is no need to export them.

lib/wibox/layout/align.lua
sclu1034: Lines like `expand_mode_impl[self._private.expand](...)` are cute, but they increase the complexity of reading/comprehending a function quite a lot.
Aire-One's suggestion is much easier to skim:

```lua
if expand == "none" then
  layout_none(...)
elseif expand == "inside" then
  layout_inside(...)
end
```

nenikitov:
OK, I will focus on finishing with the unit tests for `"justified"` expand mode now, and then squash the commits and move the functions around.

actionless (approved):

actionless:
https://github.com/awesomeWM/awesome/runs/5613279666?check_suite_focus=true#step:33:158

nenikitov:
I see there is an issue with the `fixed-lgi-lua5.2` test about `lib/wibox/layout/align.lua:450:15: shadowing upvalue 'max' on line 33`, I think I know what it is and how to correct it. I will do it as soon as possible.

nenikitov:
It should all be fixed now.
The problem was the line `local max = math.max` on top of the file, which conflicted with [this line](https://github.com/nenikitov/awesome/blob/master/lib/wibox/layout/align.lua#L448) because it defined a variable with the same name. I fixed it as soon as I discovered it, but I didn't notice that the fix ended up being in the 2nd commit after squashing, so the 1st didn't pass the tests.
I double checked all 3 commits with `make check-unit` and `make check-qa` and it should be good now.

nenikitov:
I am not too sure about this `fixed-lgi-lua5.2` error.
```
Fetching fork
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
error: could not fetch fork
Error: Process completed with exit code 1.
```
From what I understand, it is because git fetches from `git://github.com/nenikitov/awesome.git` and GitHub stopped supporting fetching from unencrypted `git://` protocol January 11 2022; now it only supports `ssh://` or `https://`.

sclu1034:
You can ignore that one.
#3592

nenikitov:
Ah OK, I see. Thanks for letting me know.

spec/wibox/layout/align_spec.lua
```diff
@@ -223,6 +223,190 @@ describe("wibox.layout.align", function()
...
+                        assert.widget_fit(layout, { 100, 30 }, { 15, 30 })
+                    end)
+
+                    it("layout", function()
+                        assert.widget_layout(layout, { 100, 30 }, {
+                            p(first,  0,  0, 100, 20),
+                            p(third,  0, 20, 100, 10),
+                        })
+                    end)
+                end)
```
sclu1034, line 349: Since you explicitly mention priority between the outer widgets based on their size, you should also add test cases for that.

E.g. one with something like `{ 100, 25 }` where `third` can still be placed, but at less than its desired width, and one with `{ 100, 15 }` or `{ 100, 20 }` where only `first` can be placed.

lib/wibox/layout/align.lua
```diff
@@ -35,127 +35,336 @@ local base = require("wibox.widget.base")
...
+            fit_widget(widget, size)
         end
-        table.insert(result, base.place_widget_at(self._private.first, 0, 0, w, h))
+
+        local placed = layout_place_widget(
+            self, width, height,
+            widget, offset, size
+        )
+        table.insert(result, placed)
+        size_remains = size_remains - size
```
Aire-One, line 129: I think this function should return `placed` (and `size_remains`?). Then, it would be the responsibility of the caller to do the insert.

I can understand we want to keep some level of closure to wrap the function call, but I just don't like a closure to have side effect and alter the result directly.

spec/wibox/layout/align_spec.lua
```diff
@@ -223,6 +223,190 @@ describe("wibox.layout.align", function()
...
+                        assert.widget_fit(layout, { 100, 30 }, { 15, 30 })
+                    end)
+
+                    it("layout", function()
+                        assert.widget_layout(layout, { 100, 30 }, {
+                            p(third,  0, 10, 100, 20),
+                            p(first,  0, 0, 100, 10),
+                        })
+                    end)
+                end)
```
sclu1034, line 404: Same extra test cases as above. Probably even more important here, to make sure things really aren't placed "first come, first serve".

sclu1034 (changes requested):

spec/wibox/layout/align_spec.lua
```diff
@@ -223,6 +223,190 @@ describe("wibox.layout.align", function()
...
+                    third = utils.widget_stub(10, 10)
+
+                    layout:set_first(first)
+                    layout:set_second(second)
+                    layout:set_third(third)
+                end)
+
+                describe("with enough space", function()
+                    it("fit", function()
+                        assert.widget_fit(layout, { 100, 100 }, { 15, 45 })
```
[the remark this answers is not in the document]
nenikitov: Yes, it should be `55`. Since this tests was passing with `45`, I imagine there might be an error in the code as well. I will fix it as soon as I get a chance.

spec/wibox/layout/align_spec.lua
nenikitov: OK, I will add more test cases.

lib/wibox/layout/align.lua
```diff
@@ -35,127 +35,336 @@ local base = require("wibox.widget.base")
...
+        local size_second = fit_widget(second)
+
+        if size_second >= size_total then
+            -- Second widget takes all layout's space, no need to place another widgets
+            place_widget(second, 0, size_total)
+        elseif size_second == 0 then
+            -- There is no second widget
+            local size_side = floor(size_total / 2)
+
+            local size_first = fit_widget(first, size_side)
```
Aire-One, line 145: nit: Since it's used only once and the variable name doesn't provide more detail than what we can guess with the function name/parameter name, you can remove the variable and call fit_widget directly in the place_widget invocation.

lib/wibox/layout/align.lua
```diff
@@ -35,127 +35,336 @@ local base = require("wibox.widget.base")
...
+            place_widget(
+                second,
+                -- After the first
+                max_size_side,
+                -- Whatever is left - expand
+                size_remains
+            )
         end
-        table.insert(result, base.place_widget_at(self._private.second, x, y, w, h))
+    else
```
Aire-One, line 343: Can you add a comment here to explicitly say we default to INSIDE

lib/wibox/layout/align.lua
```diff
@@ -35,127 +35,336 @@ local base = require("wibox.widget.base")
...
+        -- Default values
+        if not size then
+            size = fit_widget(widget, size_remains)
         else
-            if self._private.dir == "y" then
-                h = size_remains
-            else
-                w = size_remains
-            end
+            fit_widget(widget, size)
```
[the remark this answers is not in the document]
nenikitov: Now that I look at this part again, I don't think there is a need for the `fit_widget(widget, size)` if the size is passed as an argument, so this part can be simplified to:
```lua
if not size then
    size = fit_widget(widget, size_remains)
end
```
or
```lua
size = size or fit_widget(widget, size_remains)
```
I will investigate it in more detail as soon as I get a chance.

lib/wibox/layout/align.lua
```diff
@@ -35,127 +35,336 @@ local base = require("wibox.widget.base")
...
+    local third = self._private.third
+
+    -- Wrapper around layout_fit_widget function to pass less parameters
+    local function fit_widget(widget, max_size)
+        -- Can't place a widget if it does not exist
+        if not widget then
+            return 0
         end
+        -- Default values
+        max_size = max_size and max_size or size_remains
```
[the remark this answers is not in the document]
nenikitov: OK. I will correct this in any other parts of my code as well.

lib/wibox/layout/align.lua
nenikitov: Makes sense, I will change the return of the function and the code where I call it accordingly.

lib/wibox/layout/align.lua
nenikitov: OK, I will remove this variable.

lib/wibox/layout/align.lua
nenikitov: Sure, I will add the comment.

lib/wibox/layout/align.lua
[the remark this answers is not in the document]
Aire-One: With a second thought, I think you can just do a `if not max_size then max_size = size_remains end` if case. It would be even easier to read.

actionless:
> From what I understand, it is because git fetches from git://github.com/nenikitov/awesome.git and GitHub stopped supporting fetching from unencrypted git:// protocol January 11 2022; now it only supports ssh:// or https://.

rebase on master branch - that's already fixed there

spec/wibox/layout/align_spec.lua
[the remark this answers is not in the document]
nenikitov: Yes it should have been `55`, I didn't notice that I also needed to remake the `align:fit` function, because it should report different size for the `"justify"` mode. I fixed it in [9ac156d](https://github.com/awesomeWM/awesome/pull/3588/commits/9ac156db6d3c5eef7a3c02f9cff7a4a1a01f33d2), and will squash it with previous commits so it doesn't have the bug from the start.

spec/wibox/layout/align_spec.lua
nenikitov: I replaced this 1 unit test for not enough width by 6 for different width configurations for each of the 3 size cases in [6499cc4](https://github.com/awesomeWM/awesome/pull/3588/commits/6499cc496ae26b3aac941badebcb81ec0f4d5f98).
Now, in total, for the width, we have 7 tests:
* More that required for all 3 widgets (expand the smallest side and the middle)
* Just enough to fit 3 widgets (stop expanding the smallest side and the middle, place widgets right one after another)
* Barely enough to fit 3 widgets (shrink center widget)
* Just enough to fit 2 widgets (place only side widgets right one after another)
* Barely enough to fit 2 widgets (shrink the smallest side)
* Just enough to fit 1 widget (place only the longest side)
* Barely enough to fit 1 widget (place only a part of the longest side)

nenikitov:
Quick update on what I did:
* Remade the `fit` function because I didn't realize that the `"justified"` expand mode should report different size
* Added more unit tests for the `"justified"`
* Changed the function `place_widget` so it does not insert the widgets by itself. Now each of the branches in expand code has its own return statement, so there is no need for insert

Questions:
* Should I keep the variable declarations [here](https://github.com/nenikitov/awesome/blob/master/lib/wibox/layout/align.lua#L121)? Almost every expand mode needs to initialize these variables, but in different order and with different values. I could remove these declarations from the top and just prepend `local` the first time these variables are used if it would look better.
* I noticed that in the unit tests for the `"outside"` (for example [here](https://github.com/nenikitov/awesome/blob/master/spec/wibox/layout/align_spec.lua#L119)), the spec specifies that the expected behavior would be to have a 1px gap between the second and the third widgets (`42` from second offset + `15` from second length is `57` which is not equal to `58` offset of the third). Should I correct the test and the code accordingly?

sclu1034 (changes requested):

lib/wibox/layout/align.lua
```diff
@@ -35,128 +34,391 @@ local base = require("wibox.widget.base")
...
-            end
+
+    local function filter_nil(t)
+        local filtered = {}
+        for _, v in pairs(t) do
+            filtered[#filtered + 1] = v
         end
-        table.insert(result, base.place_widget_at(self._private.first, 0, 0, w, h))
+
+        return filtered
```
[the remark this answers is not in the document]
nenikitov: I corrected it. This also fixed the bug with widgets not getting placed when not all 3 slots were populated, since `pairs` would stop at the first `nil` it encounters.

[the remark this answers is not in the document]
sclu1034: > `pairs` would stop at the first `nil` it encounters.

It doesn't. As I said, `pairs` has no notion of "order" in table keys, so it cannot know a "first `nil`". `pairs` always iterates over all present fields, both string and numeric indices.
`ipairs` is the one that is defined as "iterate over integer indices starting from `1` in ascending order, _up to the first empty one_".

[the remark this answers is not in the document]
nenikitov: This makes sense, thanks for the explanation.
This code was removed and now I use `gears.table.from_sparse` instead of this function.

nenikitov:
Hello. All unit tests pass and my new align mode is fully implemented. This pull request should be done unless there are any suggestions/corrections for my code.

Aire-One:
Looks like the CI doesn't pass, all "Build & Test" jobs fail with:

```
55 tests finished.
There were 1 errors:
 - /home/runner/work/awesome/awesome/tests/test-awful-widget-only_on_screen.lua: Error: running function for step 4/20 (@1): ...e/awesome/tests/test-awful-widget-only_on_screen.lua:87: assertion failed!
 ```

**Review by sclu1034 — CHANGES_REQUESTED** [review]

**Comment by nenikitov** [comment]
> Looks like the CI doesn't pass, all "Build & Test" jobs fail with:
> 
> ```
> 55 tests finished.
> There were 1 errors:
>  - /home/runner/work/awesome/awesome/tests/test-awful-widget-only_on_screen.lua: Error: running function for step 4/20 (@1): ...e/awesome/tests/test-awful-widget-only_on_screen.lua:87: assertion failed!
> ```

OK, I will look into it as soon as I get a chance

File: lib/wibox/layout/align.lua [hunk 18]
```diff
@@ -35,128 +34,396 @@ local base = require("wibox.widget.base")
...
+
+            size_third = fit_widget(third, size_total - size_first)
+            placed_third = prepare_place_widget(
+                third,
+                -- At the end
+                size_total - fit_widget(third),
+                size_third
+            )
+
+            result = { placed_first, placed_third }
```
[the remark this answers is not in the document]
nenikitov: "First come, first served" seems like a better option. I will try to make it work like that.

lib/wibox/layout/align.lua
```diff
@@ -35,128 +34,396 @@ local base = require("wibox.widget.base")
...
-                end
-            else
-                w, _ = base.fit_widget(self, context, self._private.third, size_remains, height)
-                if self._private.expand == "inside" then
-                    size_remains = size_remains - w
-                end
-            end
+    elseif self._private.expand == align.expand_modes.OUTSIDE then
+        if not second then
+            -- Only 1 widget, place only the first one
```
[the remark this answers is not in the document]
nenikitov: That is true, we can populate widgets separately instead of sequentially. I will modify this.

lib/wibox/layout/align.lua
```diff
@@ -35,128 +34,396 @@ local base = require("wibox.widget.base")
...
+                placed_first = prepare_place_widget(
+                    first,
+                    -- At the beginning
+                    0,
+                    -- Takes the space it needs
+                    size_first
+                )
+
+                -- Whatever second widget would leave it if the third would be placed
+                size_third = math.max(size_total - size_second - size_first, size_third)
```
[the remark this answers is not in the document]
nenikitov: > Shouldn't the first and third widget always be placed with the same size if `max_size_side * 2 >= size_total` (i.e. there is enough space for them)?
Yes it should, I will tweak how the sizes are computed.

lib/wibox/layout/align.lua
```diff
@@ -35,128 +34,396 @@ local base = require("wibox.widget.base")
...
-    -- in the beginning, or in the remaining space, if it is "inside"
-    if self._private.second and size_remains > 0 then
-        local x, y, w, h = 0, 0, width, height
-        if self._private.expand == "inside" then
-            if self._private.dir == "y" then
-                h = size_remains
-                x, y = 0, size_first
+    elseif self._private.expand == align.expand_modes.JUSTIFIED then
+        size_first = fit_widget(first, size_total)
+        size_second = fit_widget(second, size_total)
```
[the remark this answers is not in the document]
nenikitov: Good catch. I will change how it is sized.

PR #3589, opened 2022-03-16
Improve layout assert output

sclu1034:
Test cases that use very small widget sizes will still not render properly, as a minimum size is required to render the widget data. In those cases, the test code should be adjusted, but that's rather low priority for existing code, as it only becomes necessary when the test fails.

Similarly, some test cases use relatively large widgets, which should render correctly, but will be hard to read.

### DIFF
--- a/spec/wibox/test_utils.lua
+++ b/spec/wibox/test_utils.lua
@@ -39,20 +39,34 @@ end
 say:set("assertion.widget_fit.positive", "Offering (%s, %s) to widget and expected (%s, %s), but got (%s, %s)")
 assert:register("assertion", "widget_fit", widget_fit, "assertion.widget_fit.positive", "assertion.widget_fit.positive")
 
+local function times(char, len)
+    local chars = {}
+    for _ = 1, len do
+        table.insert(chars, char)
+    end
+    return table.concat(chars, "")
+end
+
+--- Pad the `str` with `len` amount of characters `char`.
+-- When `is_after == true`, the padding will be appended.
+-- This allows exceeding the limit of 99 in `string.format`.
+local function pad_string(str, char, len, is_after)
+    len = len - #str
+    local pad = times(char, len)
+
+    if is_after == true then
+        return str .. pad
+    else
+        return pad .. str
+    end
+end
+
 local function draw_layout_comparison(actual, expected)
     local lines = {
         "Widget layout does not match:\n",
     }
 
     local left, right = {}, {}
-
-    local function times(char, len)
-        local chars = {}
-        for _ = 1, len do
-            table.insert(chars, char)
-        end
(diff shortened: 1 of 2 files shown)


actionless:
i had to reopen it to trigger GH actions which were ignoring this PR

actionless (approved):

spec/wibox/test_utils.lua
```diff
@@ -67,68 +83,139 @@ local function draw_layout_comparison(actual, expected)
...
-        local height = widget._height
-        local width = widget._width
+        -- We need a minimum size to render the data in.
+        -- One line/column on all sides for the boundary.
+        -- Three lines, 12 columns for the text.
+        local min_size = { 14, 5 }
+
+        if widget._width < min_size[1] or widget._height < min_size[2] then
+            too_small = true
+        elseif widget._width > 25 or widget._height > 25 then
```
Aire-One, line 93: From where do you pull this `{ 25, 25 }` limit?

spec/wibox/test_utils.lua
```diff
@@ -39,20 +39,36 @@ end
...
+end
+
+--- Pad the `str` with `len` amount of characters `char`.
+-- When `is_after == true`, the padding will be appended.
+-- This allows exceeding the limit of 99 in `string.format`.
+local function pad_string(str, char, len, is_after)
+    len = len - #str
+    local pad = times(char, len)
+
+    if is_after == true then
```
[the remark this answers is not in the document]
sclu1034: The parameter only makes sense as a boolean. Since we generally don't do `assert`s, I think it makes more sense to consider something non-sensical like `is_after == "foo"` not the same as an explicit `true`.

spec/wibox/test_utils.lua
sclu1034: There is very little to base things on.

I guess one could argue that with an overall terminal max width of 80 columns, the width could be 35.
But they really don't need to be that big, so going lower allows for scenarios where people have the terminal in a vertical layout on a small screen. So `min_width + 10` (i.e. up to 5 columns of padding), rounded to a nice number, looks good enough to me.

For the height, I don't see any useful reference, so I just took the same number.

actionless: btw, i recommend using named constants instead of hardcoded values, which would make the code much easier to understand when reading

spec/wibox/test_utils.lua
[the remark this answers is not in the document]
Aire-One: I don't agree with that, but I can understand your POV here. So let's say we agree :+1:

Aire-One (approved):

Elv13 (approved):

sclu1034:

Issue #2960, opened 2020-01-05
Menubar doesn't respect Path field of desktop files


dalz:
Output of `awesome --version`:

awesome v4.3 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2

**How to reproduce the issue:**

- put the following into a .desktop file in ~/.local/share/applications

```
[Desktop Entry]
Name=Test Path
Path=/
Exec=sh -c 'pwd; read'
Type=Application
Terminal=true
```

- run "Test Path" with menubar

**Actual result:**

A terminal opens with `/home/ale` written to it (`ale` is my username)

**Expected result:**

The output on the terminal should be the same as the `Path` field, here `/`.

Elv13:
Thanks for reporting this.

This will go into the backlog for now unless you (or someone else) can provide a patch/testcase. The whole APIs around `.desktop`/XDG_menu handling is in dire need of a large refactor. There is multiple issues both in the API design, the spec compliance and the overall buggyness/performance.

PR #3590, opened 2022-03-19
fix(menubar): Add support to Path field in desktop file(#2960)

Ancker-0:
See <https://specifications.freedesktop.org/desktop-entry-spec/1.1/ar01s05.html>

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -345,6 +345,9 @@ function utils.parse_desktop_file(file)
         if program.Terminal == true then
             cmdline = utils.terminal .. ' -e ' .. cmdline
         end
+        if program.Path then
+            cmdline = 'sh -c "cd ' .. program.Path .. ' && ' .. cmdline .. '"'
+        end
         program.cmdline = cmdline
     end
 


lib/menubar/utils.lua
```diff
@@ -345,6 +345,9 @@ function utils.parse_desktop_file(file)
         if program.Terminal == true then
             cmdline = utils.terminal .. ' -e ' .. cmdline
         end
+        if program.Path then
+            cmdline = 'sh -c "cd ' .. program.Path .. ' && ' .. cmdline .. '"'
```
sclu1034, line 349: This wraps `cmdline` in quotes, but doesn't escape any quotes inside `cmdline`. This will very likely break for any `program.Exec` that contains quotes (single or double) itself.
You'll have to alter `cmdline` so that all quotes are escaped, including once that already had been escaped.

sclu1034 (changes requested):

lib/menubar/utils.lua
actionless: and also this works only if path is exists, while spec not define the behavior, i'd rather use `;` instead of `&&`

sclu1034: I think that if `Path` is specified, the expected behaviour would be to either run with CWD set to that directory or not at all.
Starting with any other directory, because we ignore the error from `cd` could lead to confusing, hard to debug behaviour.

Ancker-0: can I use something like:
`cmdline = 'cd ' .. program.Path .. ' && ' .. cmdline`
and modify [`lib/menubar/init.lua#L211`](https://github.com/awesomeWM/awesome/blob/master/lib/menubar/init.lua#L211) with
`awful.spawn.with_shell`
This may solve the quotes problem

actionless: add new property for menuitem "with_shell", and set it here and and for program.Terminal

and at line 211 check if menuitem.with_shell is set and only when use spawn_with_shell

actionless:
and also please add some minimal tests to cover the new logic branches

Ancker-0:
Sorry, I don't know how to write tests :(

actionless:
just look into existing one as an starting point,
IMO for this stuff the `spec` tests would fit the most

Ancker-0:
I tried ``local spawn = require("awful.spawn")`` in the test, but it failed when I ran it:
```
Error → lib/gears/timer.lua @ 283
suite spec/menubar/utils_spec.lua
lib/gears/timer.lua:283: attempt to call a nil value (field 'connect_signal')

stack traceback:
	lib/gears/timer.lua:283: in main chunk

make[4]: *** [CMakeFiles/check-unit.dir/build.make:71: CMakeFiles/check-unit] Error 1
make[3]: *** [CMakeFiles/Makefile2:13914: CMakeFiles/check-unit.dir/all] Error 2
make[2]: *** [CMakeFiles/Makefile2:13921: CMakeFiles/check-unit.dir/rule] Error 2
make[1]: *** [Makefile:6591: check-unit] Error 2
make: *** [Makefile:45: check-unit] Error 2
```

I thought ``connect_signal`` is a function in ``naughty``, so I added ``local naughty = require("naughty")
``, but new errors came:
```
Error → lib/awful/screen.lua @ 337
suite spec/menubar/utils_spec.lua
lib/awful/screen.lua:337: attempt to index a nil value (field 'screen')
```
What should I do to use ``awful.spawn`` in the test?

Ancker-0:
I saw some hacks in existing code:
```
package.loaded["awful.client"] = {}
package.loaded["awful.layout"] = {}
package.loaded["awful.screen"] = {}
package.loaded["awful.tag"] = {}
package.loaded["gears.timer"] = {}
```
I added these before ``local spawn = require("awful.spawn")``, but still not work

actionless:
spec means simple unit tests, something what using spawn sounds like an integration test - in that case check examples from tests/ directory

Aire-One:
I don't think we need an itest here. A unit test should be enough. We need to use a spy to mock and observe the `awful.spawn` invokation (http://olivinelabs.com/busted/#spies-mocks-stubs).

@Ancker-0 these "hacks" are done to mock the API. When we do a unit test, we want to test a specific function. We don't need the whole application to be running, and the test framework doesn't run Awesome anyway. These `package.loaded` assignation are here to fake the modules are loaded (sometime, we have to mock bigger data structures), so we can freely use and test the function.

Ancker-0:
Thank you. My former thought is let it run a ``.desktop`` file and check the results. Actually I only need to parse the file and check the ``cmdline`` property.

Ancker-0:
In [this line](https://github.com/Ancker-0/awesome/blob/25122ca3c3c292c4e645cbb325f259215b39abe5/lib/menubar/utils.lua#L351), if ``program.Path`` contains spaces, it won't work. A solution might be similar to [``awful.spawn.with_shell``](https://github.com/Ancker-0/awesome/blob/25122ca3c3c292c4e645cbb325f259215b39abe5/lib/awful/spawn.lua#L370) like
```
awesome.spawn({"env", "-C", program.Path, "-S", program.cmdline}, false)
```
This wraps ``program.Path`` as a single parameter to ``env``. However, I'm not sure if it is appropriate to directly call ``awesome.spawn`` instead of ``awful.spawn``.

Aire-One:
Every utility functions from `awful.spawn` is based of on calling `awesome.spawn`. Basically, `awful.spawn` is an abstraction over `awesome.spawn`. You can use a table argument to "truncate" your command with `awful.spawn`. It should behave the same as what you showed, so I think I would prefer you stick with it.

Regarding the tests, I would like to see tests that mock awful.spawn/awful.spawn.with_shell calls from `perform_action` to ensure the new if-branches are correctly called according to the new checks.

sclu1034:
> In [this line](https://github.com/Ancker-0/awesome/blob/25122ca3c3c292c4e645cbb325f259215b39abe5/lib/menubar/utils.lua#L351), if `program.Path` contains spaces, it won't work
Another way to fix this would be to wrap it in quotes.

lib/menubar/utils.lua
```diff
@@ -343,8 +343,13 @@ function utils.parse_desktop_file(file)
             cmdline = cmdline:gsub('%%i', '')
         end
         if program.Terminal == true then
+            program.with_shell = true
             cmdline = utils.terminal .. ' -e ' .. cmdline
         end
+        if program.Path then
+            program.with_shell = true
+            cmdline = 'cd ' .. program.Path .. ' && ' .. cmdline
```
akinokonomi, line 351: To avoid problems with spaces, path can just be quoted like this as @sclu1034 suggested
```suggestion
            cmdline = 'cd "' .. program.Path .. '" && ' .. cmdline
```

PR #3593, opened 2022-03-19
fix(gh-actions): s/git_url/clone_url/g (fixes #3592)

actionless:

PR #3598, opened 2022-03-23
Allow codecove to be green for small PRs

Aire-One:
Hello,

We have some PRs that needs to be manually merged because codecov is red. Most of the time, they are simple fixes where codecove yields a reduction of coverage of -0.01%. The last example in date is https://github.com/awesomeWM/awesome/pull/3593 (I guess this PR will also be affected)

With this yml config, we allow a threshold of 5% (arbitrarily selected to allow a little bandwidth for quick fix/patch). 

I have no idea if it'll work. It's all assumption based on https://docs.codecov.com/docs/common-recipe-list and https://docs.codecov.com/docs/commit-status#threshold.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -24,3 +24,6 @@ indent_size = 4
 # In a Makefile, the tab is significant and not "just" whitespace
 [Makefile]
 indent_style = tab
+
+[*.yml]
+indent_size = 2
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 5%
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,4 +3,4 @@ coverage:
     project:
       default:
         target: auto
-        threshold: 5%
+        threshold: 0.5%


actionless:
what about 0.5%?

sclu1034:
I agree that 5% is excessive.

Just going through the recent PRs, it looks like for the vast majority the diff is less than 0.1%. And even for PRs that actually add new code paths without coverage, it's usually still below 1%.

actionless:
yeah, mb even smaller than 0.5

sclu1034:
If I read https://docs.codecov.com/docs/commit-status#patch-status correctly, you can add `coverage.status.patch` to still make sure that the PR's direct changes are covered at 100%.

actionless:
so let's agree on 0.5 for now then and get it merged?

sclu1034:
That's not what I said.

_If_ we added the separate threshold on the `patch` check, it would make it more likely that the diff in the `codevoc/project` check is unrelated to the PR's changes. Therefore, giving some leeway on the `project` check in that scenario would be fine.

_If_ the separate threshold is _not_ added, the `project` threshold should be kept small. Multiple "quick fixes" could still accumulate into a larger, but hard to track, reduction in coverage.

actionless (approved):

sclu1034 (approved):

Elv13 (approved):

Issue #3582, opened 2022-03-05
gap_mult_client support 


hleft:
I saw `gap_single_client` :  Enable gaps for a single client.

please support `gap_mult_client`,   Enable gaps for mult clients.

reason: I only want to use gap when only one client,  if there are mult clients, hope disable gap for save space and don't see gap in center

actionless:
i guess you just need to set useless_gap > 0 and gap_single_client to false

hleft:
@actionless  need  ` set useless_gap > 0 and gap_single_client to true and gap_mult_client to false`

Aire-One:
Hello, that's not a feature we currently have, and it looks like there isn't many interest in it.

I guess the best option you have is to implement this behavior on your config.

Aire-One:
Oh! Guess I'm wrong, your PR just made me realize there is a possible workaround: You can change the layout skip_gap function to implement the desired behavior. This would however require you to use your own layout.

PR #3603, opened 2022-04-04
gap_mult_client support

hleft:
releated https://github.com/awesomeWM/awesome/issues/3582

hope support disable gap when mult client, support gap when single client

### DIFF
--- a/lib/awful/layout/init.lua
+++ b/lib/awful/layout/init.lua
@@ -182,12 +182,18 @@ function layout.parameters(t, screen)
         gap_single_client = t.gap_single_client
     end
 
+    local gap_mult_client = true
+
+    if(t and t.gap_mult_client ~= nil) then
+        gap_mult_client = t.gap_mult_client
+    end
+
     local useless_gap = 0
     if t then
         local skip_gap = layout.get(screen).skip_gap or function(nclients)
             return nclients < 2
         end
-        if gap_single_client or not skip_gap(#clients, t) then
+        if (gap_single_client and skip_gap(#clients, t)) or (not skip_gap(#clients, t) and gap_mult_client) then
             useless_gap = t.gap
         end
     end
--- a/lib/awful/tag.lua
+++ b/lib/awful/tag.lua
@@ -50,6 +50,7 @@ defaults.gap                 = 0
 
 -- The default gap_count.
 defaults.gap_single_client   = true
+defaults.gap_mult_client     = true
 
 -- The default master fill policy.
 defaults.master_fill_policy  = "expand"
@@ -1171,6 +1172,29 @@ function tag.object.get_gap_single_client(t)
         return val
     end
     return defaults.gap_single_client
+end
+
+--- Enable gaps for mult client.
+--
+-- @beautiful beautiful.gap_mult_client
+-- @param boolean (default: true)
+-- @see gap
+-- @see gap_single_client
+
+function tag.object.set_gap_mult_client(t, gap_mult_client)
+    tag.setproperty(t, "gap_mult_client", gap_mult_client == true)
+end
+
+function tag.object.get_gap_mult_client(t)
+    local val = tag.getproperty(t, "gap_mult_client")
+    if val ~= nil then
+        return val
+    end
+    val = beautiful.gap_mult_client
+    if val ~= nil then
+        return val
+    end
+    return defaults.gap_mult_client
 end
 
 --- Get the spacing between clients.


Aire-One:
Something I don't like when adding new global flags to enable features is that it can't scale with all the features request we can have in the long term. (And also, it makes the code base unusessary complex to manage every corner case)

I think it would be more future-proof to expose a skip_gap function at the tag level. This would allow wrapping both gap_single_client and gap_mult_client behavior (and more derivates we haven't thought of yet) with only one property.

raven2cz:
> Hello, @hleft
> 
> Thank you for the PR!
> 
> Something I don't like when adding new flags to enable features is that it can't scale with all the features request we can have in the long term. (And also, it makes the code base unnecessary complex to manage every corner case)
> 
> I think it would be more future-proof to expose a skip_gap function at the tag level. This would allow wrapping both gap_single_client and gap_mult_client behavior (and more derivates we haven't thought of yet) with only one property.

# Configuration Hell

After 30 years with configuration design, I would have a few points that have helped many companies and frameworks. In general, configurations are poorly learned in IT fields and are mainly part of cybernetics lessons and studies, which are unfortunately a problem for IT programmers.

* Configuration is an essential part of the API, and if poorly designed, there are major problems in using the resulting product, especially if it is a framework.
* Configuration must be considered in design time of component, service or package. FIRST, It is always necessary to design the features as DISABLED (completely with all deps!) and after it solve activation of features and deps! NEVER do anything without the possibility of a complete deactivation (plus logging of activity/deactivity too).
* The biggest mistake for non-cybernetics programmers is that they think the configuration belongs to the object/class (or lua table). This is a big mistake! The configuration belongs to aspect oriented programming. It is an aspect, an abstraction of a given object or service instance. Imagine that you get green glasses (like Emerald city in OZ land, there was no green, it was caused by green glasses only) which you look at the component you need to configure. These green glasses contain configuration properties for a given object, but they are defined **separately** as a configuration aspect and are injected into the object as an aspect. Lua has good resources for outsourcing features. It is also possible to be inspired in the Kotlin programming language now, where this is already at the core of the language. Configurations are thus kept separate, they can be much better described, aggregated, activated/deactivated, do documentation and most importantly make a hierarchy (see next points below). However, due to the aspect injection, they are **still present** in the object, which can use them directly, but only in the presence of green glasses.
* Aspect configuration creates the separation of groups and thus the separation of information. Configuration belongs to the so-called cross-concern problems and is the main reason why not solve configurations directly per object, or worse, as a **global configuration**!
* I often get the question, but how to deal with a global configuration, or a configuration that affects some package and not just separate service or component? There are several types of configurations. The basic division is to configure the properties of the object and the second level is to configure the behavioral group. Here again, companies and open source make the biggest mistake in describing the group's configuration by table listing, for example, 6 parameters to be set to achieve the appropriate behavior. Big mistake! The configuration is not just about the property of the object, but about the 2nd level is the behavioral group. It's just adding the second aspect (say blue glasses). You define a configuration parameter for a behavioral group and its simple string values that internally contains, according to its parametrization, a table of parameters of the properties of individual objects, as they are to be configured according to the value of the behavioral group parameter. The user thus gets **only one** simple parameter that correctly controls some part of the unit/package/complex-service. It can be also easily overload it and add its own/user-specific added value. The key/trick is that if you ask for parameterization INSIDE in any object instance, you will get the **correct setting value directly** from it due to the aspect dependence and injection. There is no need to search for anything, everything works simply and most importantly dynamically! What does dynamically mean? You can easily update the configuration **on-the-fly** and change the behavior of the services without having to restart them. Which is a nightmare for all no-cybernetics developers.
* Is it possible to do this with ready-made/older applications and frameworks? The answer, of course, is yes. There is no need to redo anything, for new components and packages it is possible to go straight to this method. After understanding the basic principles, you can start redesigning older components per partes (step by step).
* **ATTENTION**. The configuration determines **the usability of the whole project and its dependencies**. Don't take it lightly. On the contrary. There are many projects that have ended due to complexity, and this is becoming mainly the growth of applications/frameworks, as everyone puts in their configuration features there, as they like and simple ignore cross-concerns.

I would like this "article" to help you in the design of frameworks and applications. The specific implementation and design of this approach depends on the agreement of the main developers and the organization of the main libraries and documentation. It is mainly about the mind change and approach to configuration and to understand these problems, which arise especially during the lifetime of the framework and incoming cross-concern features.

actionless (approved):

PR #3611, opened 2022-04-25
Always use the latest time on focus changes

felipetavares:
This is more to initiate the conversation than a final fix.

Telegram on awesome has a [very annoying issue](https://github.com/telegramdesktop/tdesktop/issues/1041) caused by focus.

Dead keys cannot be correctly used to reply to messages without switching focus out and in again to telegram. This is caused because when telegram opens the "reply to" popup and changes focus to it, that call updates the latest timestamp for a focusin in X, but awesome still has an older timestamp.

When awesome tries to focus the main telegram window again after the popup is closed, it sends an event with that prior timestamp and X ignores it.

The only solution I could think of was making X use the latest internal time, but I am not sure if that's ideal here.

# Questions

- Is there a better way of keeping track of the latest focusin timestamp?
- Why does awesome keep track of it at all?
- Could/should we use the computer clock for the time instead?

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -1881,7 +1881,7 @@ client_focus_refresh(void)
      * new client gets the input focus.
      */
     xcb_set_input_focus(globalconf.connection, XCB_INPUT_FOCUS_PARENT,
-                        win, globalconf.timestamp);
+                        win, XCB_CURRENT_TIME);
 
     /* Do this last, because client_unban() might set it to true */
     globalconf.focus.need_update = false;


Elv13:
@psychon Mind chiming in? What do you think about this?

psychon:
> Is there a better way of keeping track of the latest focusin timestamp?

Puh.... I would guess "the xfixes extension", but a quick read through its XML did not find anything... So... I guess not?

> Why does awesome keep track of it at all?

`git blame` found ce814b4dbfbda5088955fb977555b4fcc0ff79fc but that does not really say why it does this. I also really do not want to think about the possible races here, so I am fine with "the WM is always right". Also, ICCCM does not apply, since that only specified what clients should do, but does not constraint the WM (and ICCCM in general seems to forbid using CurrentTime).

> Could/should we use the computer clock for the time instead?

I'll answer the "could" part: No.

The X11 protocol does not really specify how the X11 protocol works. The only way to get a timestamp is to "do something" that causes an event containing a timestamp and wait for said event to arrive. By the time the code does something with the timestamp, it might already be old (unless one uses `GrabServer` to make the X11 server not handle requests from anyone else, but... don't do that).

> What do you think about this?

First: That merge commit seems pointless.

Second:
It's complicated. This timestamp mess exists so that X11 can work half-sane across high latency connections. But if your WM has a high latency connection to your X11 server, you'll have a bad time. I guess nothing forbids us doing this (ICCCM does not (really) apply to WMs). So... fine with me. I guess.

I guess this could still cause us to send a too old timestamp in `xwindow_takefocus()`, but... perhaps that's fine? Hopefully no clients use `WM_TAKE_FOCUS` and create windows at "ungood" times...?

PR #3616, opened 2022-05-12
Update Debian install instructions

Ttibsi:
Because of [this line](https://github.com/awesomeWM/awesome/blob/master/Makefile#L8) in the Makefile, the .deb build goes to the `build` directory instead, so you need to `cd` into it to be able to install. The `apt install` command appears to be looking for the `.deb` in their own repositories instead of local as well any more, so I've changed it to the more successful `dpkg -I` command (See the below screenshot)

image


Aire-One (approved):
Thank you for this correction!

PR #3626, opened 2022-05-31
Fix disabling notification timeout

### LINKED ISSUE #3625: Naughty.timeout = 0 not working in request::display
```
awesome v4.3-1360-gc539e0e4 (Too long)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

Inside the signal `request::display` try to set the notification to `timeout = 0`

**Actual result:**
The timeout is not applied and the notification will run out

**Expected result:**

The timeout should make the notification stay indefinitely
It works when sending a notification and setting it there, but not when using this signal and setting it there.

sclu1034:
When creating a notification with a timeout value of `0`, no timer is
created. The same behaviour was intended when setting the timeout to `0`
later, but it was never implemented.

Fixes #3625.

### DIFF
--- a/tests/test-naughty-notification.lua
+++ b/tests/test-naughty-notification.lua
@@ -1,0 +1,30 @@
+local runner = require("_runner")
+local naughty = require("naughty")
+
+local steps = {}
+
+local n
+
+-- Regression test for #3625:
+-- Make sure that setting `timeout = 0` on an existing notification
+-- correctly stops and removes the timer.
+table.insert(steps, function()
+    n = naughty.notification({ text = "regression #3625", timeout = 1 })
+    n.timeout = 0
+    return true
+end)
+
+-- Wait for the notification's timeout to pass.
+table.insert(steps, function()
+    os.execute("sleep 2")
+    return true
+end)
+
+table.insert(steps, function()
+    assert(not n.is_expired)
+    return true
+end)
+
+runner.run_steps(steps)
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80


lib/naughty/notification.lua
```diff
@@ -523,6 +523,19 @@ function notification:set_id(new_id)
     self:emit_signal("property::id", new_id)
 end
 
+
+function notification:die(reason)
```
Elv13, line 527: maybe `_die` or make it a local function?

Elv13 (changes requested):
Thanks for this. Just a minor comment. Also adding a regression test might be relevant here. The notification module seems to have many of these papercut fixes over time, so the risk of regression is rather high. Maybe a new integration test module would make sense. The current test module is getting very large and complex.

Aire-One (approved):
Thank you for the extra test 😉

actionless (approved):

tests/test-naughty-notification.lua
```diff
@@ -0,0 +1,30 @@
...
+-- correctly stops and removes the timer.
+table.insert(steps, function()
+    n = naughty.notification({ text = "regression #3625", timeout = 1 })
+    n.timeout = 0
+    return true
+end)
+
+-- Wait for the notification's timeout to pass.
+table.insert(steps, function()
+    os.execute("sleep 2")
```
Elv13, line 19: Instead of sleeping 2 full second, can it usleep for 0.3s, check, then return `nil` if the condition isn't met? I have been bitten by hardcoded blocking `sleep` in tests before

Elv13 (changes requested):
Thanks for the update. Here's 1 little thing I would like addressed

tests/test-naughty-notification.lua
sclu1034: Not really. Technically, the test is for "will this notification exist indefinitely". And "infinity has passed" is not really a condition I can check for. The closes I have is "we're still there after the previous timeout value", but I need to know that said timeout has passed for that.

I guess I could try to add up elapsed time, but I don't have enough experience with the test framework's weird "emulate async by repeatedly calling the same step" to know if that's any better of an idea than sleeping for multiple seconds.

Elv13: Our test handle this internally. If something returns `nil` too many time, the test fails. This is transparent. Just check for the condition and return `nil` if it's not met, the framework contains the boilerplate and timeout/retry logic.

sclu1034: As I said, I cannot "check for the condition". The condition is "infinity has passed and the notification is still visible".

I could access the notification's internal values to check `timer == nil`, but that wouldn't be an integration test.

paulo-granthon (approved):

tests/test-naughty-notification.lua
Aire-One: I think what Elv asked here is basically

```
   assert(notification_is_still_here)
   if max_delay_exceed then
      return true
   end
   sleep_a_little
   return nil -- iterate the test again
end)
```

`max_delay_exceed = initial timeout + additional delay for the test`, so 2 sounds ok to me.

This way, the event loop has more chances to be run, and the timer would actually tick (or not in this case!). We'd have a stronger check against the notification being destroyed.

I may be pushing a commit to implement that in a few days if no reactions come up in between.

sclu1034: > CI won't work without recent fixes, so they need to be rebased before we can merge.

The other option is to merge those changes into the PR.

But I guess through the technicalities of GitHub's PR workflow creating public forks, I have indeed published a GPL-licensed copy of my modifications even without the PR being merged. So I guess you're free to use that code within those terms.

To keep intents clear, I'll close the PR, and you can browse the fork as needed.

PR #3629, opened 2022-06-09
rectified the installation steps for .deb package

Mehdi-YC:
the deb package is generated inside a build directory , we have to cd to it before apt install

### DIFF
--- a/docs/01-readme.md
+++ b/docs/01-readme.md
@@ -52,7 +52,8 @@ sudo apt build-dep awesome
 git clone https://github.com/awesomewm/awesome
 cd awesome
 make package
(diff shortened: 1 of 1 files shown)


Aire-One:
According to https://github.com/awesomeWM/awesome/pull/3616 `apt install` doesn't work to install local packages. Can you confirm the instructions you changed works?

SAVE-POlNT:
yes the ./ work , nd yes we can ./build/*.deb i just don't know if we have adependency inside build folder that is needed to apt install this is why i separated the instructions

Elv13 (approved):

PR #3643, opened 2022-06-27
Fix for file_readable() and file_executable()

### LINKED ISSUE #3642: gears.filesystem.file_readable and gears.filesystem.file_executable file not beeing closed.
```
awesome v4.3-1360-gc539e0e4 (Too long)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```
`gears.filesystem.file_readable` and `gears.filesystem.file_executable` don't close their files leading to huge piles of opened files and a huge memory usage.

**Actual result:**

Huge memory leak when checking many files

**Expected result:**

Not leaking memory when checking many files

Crylia:
Add :close() to file_readable() and file_executable()

Fix #3642

### DIFF
--- a/lib/gears/filesystem.lua
+++ b/lib/gears/filesystem.lua
@@ -52,8 +52,12 @@ function filesystem.file_readable(filename)
     local gfile = Gio.File.new_for_path(filename)
     local gfileinfo = gfile:query_info("standard::type,access::can-read",
                                        Gio.FileQueryInfoFlags.NONE)
-    return gfileinfo and gfileinfo:get_file_type() ~= "DIRECTORY" and
+    local ret = gfileinfo and gfileinfo:get_file_type() ~= "DIRECTORY" and
         gfileinfo:get_attribute_boolean("access::can-read")
+    if gfile then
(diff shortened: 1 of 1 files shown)


Elv13:
Apparently gio doesn't have a `:close()` on the GFile handle, it has it [on the stream](https://docs.gtk.org/gio/method.IOStream.close.html). It uses reference counting too, so that releases some other stuff without explicit code. [LGI ties this to the GC](https://github.com/lgi-devs/lgi/blob/master/docs/overview.md?plain=1#L114). But also, I don't see in the [gio doc](https://docs.gtk.org/gio/method.File.query_info.html) any mention that this opens the file. It probably use some variant of `stat()`, but I didn't check. So if there is an handle leak, that's unlikely to be here. Given your evidences, it's *somewhere*, but not here.

Crylia:
grafik
Adding for @Elv13

Issue #3531, opened 2021-12-21
`base.fit_widget` encodes too much information into its return values


sclu1034:
Extracted from #3309, as this is a more general issue with layouts

> The problem is that `base.fit_widget()` returns `0, 0` for any widget whose `visible` property is `false`

Oh, that's bad. I didn't realize until now, but `base.fit_widget` actually encodes multiple different things into `0, 0`.

When spacing is involved, it actually makes a difference whether a widget is `visible = false` and both the widget and the spacing should be skipped, or if the widget only reports `0, 0` due to some internal state but should still be shown (or rather, should still be accounted for with spacing).
Similarly, if the layout doesn't have enough space to fit all children, currently the only way a child widget can report "that's not enough space for me" is, again, returning `0, 0`. But that's the condition where `wibox.layout.fixed` should `break` the loop.

sclu1034:
While this came up with the overflow layout, the current implementation of `wibox.layout.fixed` already has issues with this. In this test, both `widget:fit` and `widget:layout` fail:

```diff
diff --git a/spec/wibox/layout/fixed_spec.lua b/spec/wibox/layout/fixed_spec.lua
index a272b7910..ebf868809 100644
--- a/spec/wibox/layout/fixed_spec.lua
+++ b/spec/wibox/layout/fixed_spec.lua
@@ -183,6 +183,29 @@ describe("wibox.layout.fixed", function()
         end) -- with spacing
     end) -- with widgets
 
+    describe("with invisible child widget", function()
+        it("skips spacing", function()
+            local first = utils.widget_stub(30, 10)
+            local second = utils.widget_stub(30, 10)
+            local third = utils.widget_stub(30, 15)
+            local spacing_widget = utils.widget_stub(30, 10)
+
+            layout.spacing = 10
+            layout.spacing_widget = spacing_widget
+
+            second.visible = false
+
+            layout:add(first, second, third)
+
+            assert.widget_layout(layout, { 30, 100 }, {
+                p(first, 0, 0, 30, 10),
+                p(spacing_widget, 0, 10, 30, 10),
+                p(third, 0, 20, 30, 15),
+            })
+            assert.widget_fit(layout, { 30, 100 }, { 30, 35 })
+        end)
+    end)
+
     describe("emitting signals", function()
         local layout_changed
         before_each(function()
```

The output of `assert.widget_fit` is
```
Failure → spec/wibox/layout/fixed_spec.lua @ 187
[...]

PR #3648, opened 2022-06-29
Implement conditional container

sclu1034:
This container allows showing or hiding a child widget based on a boolean property.

I'm not quite sure about the names here. `wibox.container.if` and a property `.condition` could work as well.

### DIFF
--- a/lib/wibox/container/conditional.lua
+++ b/lib/wibox/container/conditional.lua
@@ -1,0 +1,109 @@
+---------------------------------------------------------------------------
+-- A simple container that will hide its child widget based a
+-- boolean state variable.
+--
+--@DOC_wibox_container_conditional_comparison_EXAMPLE@
+--
+-- @author Lucas Schwiderski
+-- @copyright 2022 Lucas Schwiderski
+-- @containermod wibox.container.conditional
+-- @supermodule wibox.widget.base
+---------------------------------------------------------------------------
+
+local base = require("wibox.widget.base")
+local gtable = require("gears.table")
+local gobject = require("gears.object")
+
+local conditional = { mt = {} }
+
+
+function conditional:fit(context, width, height)
+    if not self._private.state then
+        return 0, 0
+    end
+
+    return base.fit_widget(self, context, self._private.widget, width, height)
+end
+
+
+function conditional:layout(_, width, height)
+    if not self._private.state then
+        return {}
+    end
+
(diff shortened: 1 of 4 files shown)


actionless:
what the difference between setting `.state` bool of this new container vs setting `.visible` bool on already existing containers?

actionless:
(i reopened to unstuck ci)

sclu1034:
> what the difference between setting `.state` bool of this new container vs setting `.visible` bool on already existing containers?

The same difference as setting `.visible` vs calling `:remove()`/`:insert()`. The container isn't just visibly empty, the child widget is removed from the hierarchy completely.

And as outlined in #3531, the handling of `.visible` isn't all that great currently.

Because of that, I'm also not entirely comfortable with `wibox.container.visible_when`. `wibox.container.when` would be fine, though.

> Also, is it possible to refactor `only_on_screen` to be rebased on top of this module?

After a quick skim, it seems to be doing the same thing as this one, so I don't see anything that would prevent that refactor.

Elv13:
> Names:

* `wibox.container.if`: Clash with reversed keyword and syntax highlight
* `wibox.container.switch`: Clash with some syntax highlight and potentially Lua might use it in the future.
* `wibox.container.when`: No objection.

> A decision/vote whether we want:

Both IMHO, my version is complex and hard to merge. Your version is simple and fix a real problem.

Other question: 
 * Is `state` the best property name. Given it's a boolean, maybe `enabled` would make more sense? Or can we use strings states and have more than 2 (more than 1 type of child widget).
 * Does it make sense to merge this and https://github.com/awesomeWM/awesome/pull/3421 ? They kind of serve different use case, but overlap. Lazy initialization and `widget_template` compatibility makes it more useful. It makes this a better base class and there is some overlap in "future additions". Then again, it might make sense to keep them separated.

sclu1034:
> `wibox.container.when`: No objection.
>
> Both IMHO, my version is complex and hard to merge. Your version is simple and fix a real problem.
>
> maybe `enabled` would make more sense?

Fine with me.

> Does it make sense to merge this and `wibox.widget.template`?

I don't so. I don't see any major use case where it would be worth having both functionalities in one widget.

And, as you said, currently this container is quite simple to use. If they were merged, you'd need to bring the entire templating logic, unused, when you only wanted the simple `if` functionality.

sclu1034:

PR #3653, opened 2022-07-01
Make client.icon_name / WM_ICON_NAME writable

### LINKED ISSUE #3476: Make client.icon_name writable

4.3

E.g.
```
client.connect_signal("manage", function (c, startup)
    c.icon_name = 'foobar'
    io.stderr:write(string.format('icon_name: %s\n', c.icon_name))
end)
```

**Actual result:**
Displays the `WM_ICON_NAME` property in `.xsession-errors` / the variable stays as it was before.

**Expected result:**

Changes to `foobar` at least temporarily.

**Notes:**

In contrast the `c.name` variable can be modified, which is useful to change window titles etc. by connecting to the `property::name` signal.

Unfortunately `c.icon_name` is used for minimized task list widgets [here](https://github.com/awesomeWM/awesome/blob/54a5a7dce580da39152175f2668463c7751d904c/lib/awful/widget/tasklist.lua#L380) though, which makes modifying `c.name` less useful unless one accepts to lose control of the task label when the window becomes minimized.

3hhh:
client.name / WM_NAME is writable already. So it makes sense to
make its minimized version writable as well.

Closes #3476

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -631,7 +631,6 @@ lua_class_t client_class;
  * @property icon_name
  * @tparam string icon_name
  * @propemits false false
- * @readonly
  */
 
 /**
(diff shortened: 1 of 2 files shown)


actionless:
thanks!
also unit test would be needed to pass the coverage check for the PR

sclu1034:
As mentioned in #3476:

> The window manager will not change properties written by the client. 
> - https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#client_properties

Implementing this would violate ICCCM.

PR #3655, opened 2022-07-03
Implement fixes for Luacheck 0.26.1

Elv13:
* Rename or mute parameters starting with _
 * Replace `not (>=)` with `<`

(note, this is mostly untested)

### DIFF
--- a/lib/awful/keygrabber.lua
+++ b/lib/awful/keygrabber.lua
@@ -480,7 +480,7 @@ end
 -- @method stop
 -- @emits stopped
 -- @emits property::current_instance
-function keygrabber:stop(_stop_key, _stop_mods) -- (at)function disables ldoc params
+function keygrabber:stop(_stop_key, _stop_mods) --luacheck: ignore 214
     keygrab.stop(self.grabber)
 
     local timer = self._private.timer
@@ -508,7 +508,7 @@ end
 -- @tparam awful.key key The key.
 -- @tparam string description.group The keybinding group
 
-function keygrabber:add_keybinding(key, _keycode, _callback, _description)
+function keygrabber:add_keybinding(key, _keycode, _callback, _description) --luacheck: ignore 214
     local mods = not key._is_awful_key and key or nil
 
     if mods then
--- a/lib/awful/keygrabber.lua
+++ b/lib/awful/keygrabber.lua
@@ -149,19 +149,39 @@ local function grabber(mod, key, event)
     end
 end
 
+local function stop(self, stop_key, stop_mods)
+    keygrab.stop(self.grabber)
+
+    local timer = self._private.timer
+    if timer and timer.started then
+        timer:stop()
+    end
+
+    if self.stop_callback then
+        self.stop_callback(
+            self.current_instance, stop_key, stop_mods, self.sequence
+        )
+    end
+
+    keygrab.emit_signal("property::current_instance", nil)
+
+    self.grabber = nil
+    self:emit_signal("stopped")
+end
+
 local function runner(self, modifiers, key, event)
     local converted = generate_conversion_map()[key]
 
     -- Stop the keygrabber with the `stop_key`
     if (key == self.stop_key or (converted and converted == self.stop_key))
       and event == self.stop_event and self.stop_key then
-        self:stop(key, modifiers)
+        stop(self, key, modifiers)
         return false
     end
 
     -- Stop when only a subset of keys are allowed and it isn't one of them.
     if self._private.allowed_keys and not self._private.allowed_keys[key] then
-        self:stop(key, modifiers)
+        stop(self, key, modifiers)
         return false
     end
 
@@ -169,7 +189,7 @@ local function runner(self, modifiers, key, event)
     if type(self.stop_key) == "table" and event == self.stop_event then
         for _, k in ipairs(self.stop_key) do
             if k == key then
-                self:stop(k, modifiers)
+                stop(self, k, modifiers)
                 return false
             end
         end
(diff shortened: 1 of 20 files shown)


lib/awful/layout/init.lua
```diff
@@ -145,12 +145,12 @@ function layout.inc(i, s, layouts)
 end
 
 --- Set the layout function of the current tag.
--- @param _layout Layout name.
+-- @tparam layout|function l Layout object or function.
```
Aire-One, line 148: 👍

lib/gears/surface.lua
```diff
@@ -67,60 +67,60 @@ end
 --- Try to convert the argument into an lgi cairo surface.
 -- This is usually needed for loading images by file name and uses a cache.
 -- In contrast to `load()`, errors are returned to the caller.
--- @param _surface The surface to load or nil
+-- @param surface The surface to load or nil
```
Aire-One, line 70: Looks like it's not called `surface` but `self`.

(I thought ldoc would complain for this)

lib/wibox/drawable.lua
```diff
@@ -21,7 +21,7 @@ local surface = require("gears.surface")
 local timer = require("gears.timer")
 local grect =  require("gears.geometry").rectangle
 local matrix = require("gears.matrix")
-local hierarchy = require("wibox.hierarchy")
+local ghierarchy = require("wibox.hierarchy")
```
Aire-One, line 24: What is the reasoning behind the name, "Global hierarchy"?

lib/awful/keygrabber.lua
```diff
@@ -480,7 +480,7 @@ end
 -- @method stop
 -- @emits stopped
 -- @emits property::current_instance
-function keygrabber:stop(_stop_key, _stop_mods) -- (at)function disables ldoc params
+function keygrabber:stop(_stop_key, _stop_mods) --luacheck: ignore 214
```
sclu1034, line 483: What's the reason for ignoring these? There'd be no shadowing as far as I can tell.

lib/awful/keygrabber.lua
```diff
@@ -508,7 +508,7 @@ end
 -- @tparam awful.key key The key.
 -- @tparam string description.group The keybinding group
 
-function keygrabber:add_keybinding(key, _keycode, _callback, _description)
+function keygrabber:add_keybinding(key, _keycode, _callback, _description) --luacheck: ignore 214
```
sclu1034, line 511: What's the reason for ignoring these? There'd be no shadowing as far as I can tell.

lib/awful/layout/suit/floating.lua
```diff
@@ -33,71 +33,80 @@ function floating.mouse_resize_handler(c, corner, x, y)
...
-                                      else
-                                          ng = { y = (g.y + g.height) - rg.height }
-                                      end
-                                      c:geometry({ x = ng.x, y = ng.y })
-                                      return true
-                                  end
-                              end
-                              return prev_coords.x == _mouse.x and prev_coords.y == _mouse.y
-                          end, corner .. "_corner")
+    capi.mousegrabber.run(function (position)
```
sclu1034, line 50: Since this table also contains the `.buttons` field, maybe something like `state` or `mouse_state` would make more sense.

lib/awful/screen.lua
```diff
@@ -157,7 +157,7 @@ end
 -- @staticfct awful.screen.focus_bydirection
 -- @param dir The direction, can be either "up", "down", "left" or "right".
 -- @param _screen Screen.
-function screen.focus_bydirection(dir, _screen)
+function screen.focus_bydirection(dir, _screen) --luacheck: ignore 214
```
sclu1034, line 160: What about the common `s`?

lib/ruled/client.lua
```diff
@@ -152,22 +152,22 @@ end
...
 --- Get list of matching rules for a client.
 -- @tparam client c The client.
--- @tparam table _rules The rules to check. List with "rule", "rule_any", "except" and
+-- @tparam table rules The rules to check. List with "rule", "rule_any", "except" and
 --   "except_any" keys.
 -- @treturn table The list of matched rules.
 -- @staticfct ruled.client.matching_rules
-function module.matching_rules(c, _rules)
-    return crules:matching_rules(c, _rules)
+function module.matchingrules(c, rules)
```
sclu1034, line 159: Overzealous "find & replace" or is that change in API intended?

lib/wibox/widget/graph.lua
```diff
@@ -343,7 +343,7 @@ prop_fallbacks.nan_color = build_fallback_nan_color()
 --
 
 local function graph_gather_drawn_values_num_stats(self, new_value)
-    if not (new_value >= 0) then
+    if not (new_value >= 0) then --luacheck: ignore 581
```
sclu1034, line 346: I'd add a comment that this specific expression is required to include the check for NaN.

sclu1034 (changes requested):

lib/wibox/drawable.lua
Elv13: Opps, I was thinking it was in gears, but it's indeed in wibox. Fixed

lib/awful/keygrabber.lua
Elv13: It's because they are undocumented. We overload the meaning of `_` a bit between "it's private" and "it's undocumented, thus private". Luacheck folks randomly decided to implement a naming convention.

lib/awful/screen.lua
Elv13: whoops, fixed

lib/ruled/client.lua
Elv13: Eww, can't believe we have no tests for this. Thanks for finding this.

lib/wibox/widget/graph.lua
Elv13: There's already a comment 1 line below

lib/awful/layout/suit/floating.lua
Elv13: done

Elv13:
All done

lib/awful/keygrabber.lua
sclu1034: A public function with private parameters? And overloading a naming convention with a completely orthogonal meaning?

I think the much saner solution would be to document those parameters and talk about why the end user shouldn't use them, than to jump through multiple confusing hoops to hide the documentation for something the user can easily access.
Clear documentation will also help users that get confusing error messages when they accidentally pass a value, pointing to a `_stop_key` parameter in a function that, as per documentation, shouldn't have parameters.

lib/wibox/widget/graph.lua
sclu1034: That comment mentions that NaN is intended to be caught, but not how the linting rule has to be ignored because its suggested alternative doesn't actually do the same thing.

lib/awful/keygrabber.lua
Elv13: Fair point, done.

lib/wibox/widget/graph.lua
Elv13: Added. After researching, apparently `luacheck` did that on purpose and adding `--luacheck: ignore 581` is the official solution. They don't consider this a bug.

sclu1034 (approved):

PR #3659, opened 2022-07-11
Properly take into account `with_shell` arg in prompt widget

mattplm:
The argument was not taken into account in the constructor even though it is documented and the code refers to it in the `spawn_and_handle_error` function.


Aire-One (approved):
Thank you for this!

sclu1034 (approved):

Elv13 (approved):

PR #3664, opened 2022-07-22
Systray composition support

xinhaoyuan:
This allows systray to be rendered consistently in the widget hierarchy with proper shape and transparency.

The default appearance should stay the same, while the newly introduced beautiful.systray_skip_bg allows skipping the boring systray background.


awesomeConfig.cmake
```diff
@@ -148,6 +148,8 @@ set(AWESOME_DEPENDENCIES
     xcb-icccm
     xcb-icccm>=0.3.8
     xcb-xfixes
+    xcb-composite
+    xcb-damage
```
sclu1034, line 152: Those will have to be installed in CI and listed as dependencies in the docs.

sclu1034 (changes requested):

systray.c
```diff
@@ -413,4 +456,29 @@ luaA_systray(lua_State *L)
...
+        return 1;
+    }
+
+    int width = luaL_checkinteger(L, 1);
+    int height = luaL_checkinteger(L, 2);
+    /* Lua has to make sure to free the ref or we have a leak */
+    lua_pushlightuserdata(
+        L, cairo_xcb_surface_create(
+            globalconf.connection, globalconf.systray.window, globalconf.visual,
+            width, height));
```
sclu1034, line 480: Lua does no memory management on light userdata, so we do have a memory leak.
How would it even know to call `cairo_surface_destroy` on this particular `void*`?

sclu1034 (changes requested):

awesomeConfig.cmake
xinhaoyuan: Done.

systray.c
xinhaoyuan: Similar to `pixbuf_to_surface`, it will be the caller's duty to manage/destroy the returned resource. I believe that this is done in https://github.com/awesomeWM/awesome/pull/3664/files#diff-8dc3a9fa2a4507278e3c9c437bc5010216606deca220f15f825c3fbef718f06aR109 by calling `cairo.Surface([raw userdata], true)`, but I can be missing something.

tests/test-systray.lua
```diff
@@ -81,6 +81,58 @@ table.insert(steps, function()
...
+        cr:paint()
+        -- Read the first pixel (a,r,g,b) as 4 8-bit integers.
+        local data = s:get_data()
+        local array = lgi_core.record.new(wrapped_uchar, data, false)
+        local argb = {}
+        for i = 0, 3 do
+            argb[i + 1] = lgi_core.record.fromarray(array, i).v
+        end
+        -- Check with the pixel from test-systray.c
+        return
```
psychon, line 121: Dunno if I got this right, but I think all this LGI magic is not necessary:
```
local data = s:get_data()
local function get_pixel(index) return string.byte(data, index) end
return get_pixel(1) == 0 and get_pixel(2) == 153 and ...
```
(Dunno if I got the indexes right, but it's the idea that counts here)

psychon (commented):
Basically :+1: but I do not understand where this "mess with the background pixel of tray windows" comes from. I do not think any spec allows this. Is there some program out there that required this specifically?

psychon:
Ah, right after pressing "submit", I find it: https://specifications.freedesktop.org/systemtray-spec/systemtray-spec-0.3.html#visuals

> If convenient, the tray manager may set an appropriate background pixmap on the embedder window to provide the appearance of transparency. However, the preferred method of transparency is to use a visual with an alpha channel in _NET_SYSTEM_TRAY_VISUAL.

Okay. (I do not think there is much of a difference between background pixmap and background pixel, so, whatever.)

awesome.c
```diff
@@ -793,6 +797,18 @@ main(int argc, char **argv)
...
+    globalconf.have_composite = query && query->present;
+    if (globalconf.have_composite)
+        xcb_discard_reply(globalconf.connection,
+                xcb_composite_query_version(globalconf.connection, 0, 4).sequence);
+
+    query = xcb_get_extension_data(globalconf.connection, &xcb_damage_id);
+    globalconf.have_damage = query && query->present;
+    if (globalconf.have_damage)
+        xcb_discard_reply(globalconf.connection,
+                xcb_damage_query_version(globalconf.connection, 1, 1).sequence);
```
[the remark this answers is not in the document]
xinhaoyuan: Thanks for pointing it out. Fine with me to lower the requirements.

awesome.c
```diff
@@ -793,6 +797,18 @@ main(int argc, char **argv)
         xcb_discard_reply(globalconf.connection,
                 xcb_xfixes_query_version(globalconf.connection, 1, 0).sequence);
 
+    query = xcb_get_extension_data(globalconf.connection, &xcb_composite_id);
+    globalconf.have_composite = query && query->present;
+    if (globalconf.have_composite)
+        xcb_discard_reply(globalconf.connection,
+                xcb_composite_query_version(globalconf.connection, 0, 4).sequence);
```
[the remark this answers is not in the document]
xinhaoyuan: Ditto - thanks :)

objects/drawin.c
```diff
@@ -451,6 +452,8 @@ drawin_allocator(lua_State *L)
                           globalconf.default_cmap,
                           xcursor_new(globalconf.cursor_ctx, xcursor_font_fromstr(w->cursor))
                       });
+    if (globalconf.is_compositing)
+        xcb_composite_redirect_subwindows(globalconf.connection, w->window, XCB_COMPOSITE_REDIRECT_MANUAL);
```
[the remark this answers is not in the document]
xinhaoyuan: Good question. I think the composition must happen at the top-level windows (in this case, the drawins). Otherwise the drawin containing the systray widget would still draw the content of the systray window without proper composition.

I tried to use xcb_composite_redirect_window on the systray and that left the black box on the drawin (since I selected the manual update parameter).

systray.c
```diff
@@ -149,6 +169,20 @@ systray_request_handle(xcb_window_t embed_win)
     xcb_change_window_attributes(globalconf.connection, embed_win, XCB_CW_EVENT_MASK,
                                  select_input_val);
 
+
+    geom_c = xcb_get_geometry(globalconf.connection, embed_win);
+    if(!(geom_r = xcb_get_geometry_reply(globalconf.connection, geom_c, NULL)))
+        return -1;
```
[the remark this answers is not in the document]
xinhaoyuan: Valid concern. I moved the check before the property changing requests.

tests/test-systray.lua
xinhaoyuan: Unfortunately surface.get_data returns a userdata that cannot be accessed by string.byte. That's why using the magic to "cast" it into an accessible LGI type.

tests/test-systray.c
```diff
@@ -82,15 +95,54 @@ int main() {
...
+    uint8_t tray_depth;
+    if (use_tray_visual_hint) {
+        xcb_get_property_reply_t *tray_visual_r =
+            xcb_get_property_reply(
+                conn, xcb_get_property(conn, false, systray_owner,
+                                       _NET_SYSTEM_TRAY_VISUAL,
+                                       XCB_ATOM_VISUALID, 0, 1),
+                NULL);
+        tray_visual_id = screen->root_visual;
+        if(xcb_get_property_value_length(tray_visual_r)) {
```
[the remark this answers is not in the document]
xinhaoyuan: Done

objects/drawin.c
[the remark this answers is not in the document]
psychon: Okay, thanks for trying.

systray.c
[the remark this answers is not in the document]
psychon: Dunno about "valid concern", more like "should be pretty hard to hit", but thanks for taking me serious. ;-)

tests/test-systray.lua
psychon: Hm, sorry. I thought that worked somehow. I also thought there was already some code somewhere working with `.data`, but I can't find it. Sorry!

awesome.c
```diff
@@ -801,13 +801,15 @@ main(int argc, char **argv)
     globalconf.have_composite = query && query->present;
     if (globalconf.have_composite)
         xcb_discard_reply(globalconf.connection,
-                xcb_composite_query_version(globalconf.connection, 0, 4).sequence);
+                xcb_composite_query_version(globalconf.connection, 0, 3).sequence);
```
psychon, line 804: This change ended up in the wrong commit, but whatever...

psychon (dismissed):
Thanks!

awesome.c
xinhaoyuan: oops, I forgot there were two commits :p

actionless:
oh wow, wasn't expecting someone to actually pick that, many thanks in advance 🔥

tests/test-systray.c
```diff
@@ -82,15 +95,54 @@ int main() {
...
+                          | XCB_CW_BORDER_PIXEL | XCB_CW_COLORMAP,
+                      (uint32_t[]) {
+                          use_tray_visual_hint
+                              ? get_color(conn, cm, 0xffff, 0x9999, 0x0000)
+                              : XCB_BACK_PIXMAP_PARENT_RELATIVE,
+                          screen->black_pixel, cm
+                      });
     xcb_change_property(conn, XCB_PROP_MODE_REPLACE, window, _XEMBED_INFO, _XEMBED_INFO, 32, 2, (uint32_t[]) { 0, 1 });
 
     // Make our window a systray icon
```
actionless, line 145: i guess the comment itself should move together with the corresponding line

paulo-granthon (approved):

actionless:
restarting to unstuck the CI

Aceroph:
Whats preventing this PR from going any further? is it the lack of code coverage?

actionless:
@Aceroph yes + unanswered comment: https://github.com/awesomeWM/awesome/pull/3664#discussion_r973669957

given that original submitter haven't got back - i guess you're free to pick up the code in order to continue (since the current patch is gpl)

PR #3668, opened 2022-08-05
feat: 🎸 Add 3 new shortcut popup icons

kamack38:
Signed-off-by: kamack38 <64226248+kamack38@users.noreply.github.com>


lib/awful/hotkeys_popup/widget.lua
```diff
@@ -169,9 +169,9 @@ widget.labels = {
     -- Basic multimedia keys:
     XF86MonBrightnessUp   = "🔆+",
     XF86MonBrightnessDown = "🔅-",
-    XF86AudioRaiseVolume = "Vol+",
-    XF86AudioLowerVolume = "Vol-",
-    XF86AudioMute = "Mute",
+    XF86AudioRaiseVolume = "🕩+",
+    XF86AudioLowerVolume = "🕩-",
+    XF86AudioMute = "⏹",
```
Aire-One, line 174: I don't think the "Stop Button" is appropriate for the XF86AudioMute key.

According to https://emojipedia.org/muted-speaker/, the "Muted Speaker" is part of Unicode 6.0, which is the same Unicode Standard version that introduced the brightness symbols used by XF86MonBrightnessUp and XF86MonBrightnessDown. So I guess it's ok to use it.

Aire-One (changes requested):

lib/awful/hotkeys_popup/widget.lua
kamack38: Fixed in the newest commit

Aire-One (approved):
Thank you!

sclu1034 (approved):

actionless (approved):
thanks 😸👍

PR #3687, opened 2022-09-04
Fixing warnings in the documentation build.

BeniaminK:
There are warnings like:
```
WARNING: awful.hotkeys_popup.show_help from awful.hotkeys_popup seems to have ambiguous markdown in its summary, probably underscores. Please either use '`' or HTML.
```

during `ldoc` execution. This fix correctly checks when the underscores are within the '`'.


Elv13 (approved):

Elv13:
Thanks!

(Also, there might be more of these little problems in the next few months as the new doc linting system gets battle tested and code that was written before it gets merged, thanks a lot for taking the time to address this one)

PR #3689, opened 2022-09-06
doc: Fix typos

tocic:


docs/07-my-first-awesome.md
```diff
@@ -65,7 +65,7 @@ overview now also provides a cheat sheet for controlling Vim.
 
 ## Change the theme
 
-Awesome has multiple builtin themes you can choose from:
+Awesome has multiple built-in themes you can choose from:
```
sclu1034, line 68: Not a typo. In the given adjective form, "builtin" is a valid alternative to "built-in".
And as "builtin" is also a noun, and doesn't have a `-` (which is not valid in identfiers), it tends to be the more common spelling in software engineering.

sclu1034 (changes requested):

docs/07-my-first-awesome.md
tocic: I agree that it's not technically a typo to use "builtin" as an alternative to "built-in". But don't agree that it's more common.
[Searching through the repo for "built-in"](https://github.com/awesomeWM/awesome/search?q=%22built-in%22) gives 7 results 2 of which are from the docs: [09-options.md](https://github.com/awesomeWM/awesome/blob/c8d4f878a012bc6d6d0e8e520d17a1dd9237336d/docs/09-options.md?plain=1#L38) and [docs/89-NEWS.md](https://github.com/awesomeWM/awesome/blob/c8d4f878a012bc6d6d0e8e520d17a1dd9237336d/docs/89-NEWS.md?plain=1#L186).
[Searching through the repo for "builtin"](https://github.com/awesomeWM/awesome/search?q=%22builtin%22) gives 4 results the only one of which is from the docs and is proposed for fixing here.
Also [google trends](https://trends.google.com/trends/explore?cat=5&date=all&q=built-in,builtin) and  [n-grams](https://books.google.com/ngrams/graph?content=built-in%2Cbuiltin&year_start=1800&year_end=2019&corpus=26&smoothing=3&direct_url=t1%3B%2Cbuilt%20-%20in%3B%2Cc0%3B.t1%3B%2Cbuiltin%3B%2Cc0#t1%3B%2Cbuilt%20-%20in%3B%2Cc0%3B.t1%3B%2Cbuiltin%3B%2Cc0).
If you insist on reverting this line, would you like me to update other usages of "built-in" to "builtin" in the docs for consistency?

sclu1034: I guess it's not really worth the time to turn it back in this case.
And I don't have a strong preference for a specific spelling, personally, so if you were to spend the time to make it consistent, I wouldn't care which one it is.

But generally, for single changes with no clear better or worse, I tend to prefer to keep whatever is already there to avoid back and forth on personal choices.

sclu1034 (approved):

Aire-One (approved):
Thank you!

Elv13 (approved):

PR #3693, opened 2022-09-15
Improved UTF-8 characters handling in prompt

Skeevert:
I encountered a problem in the menubar, when non-latin characters weren't properly erasing, requiring to press Backspace 2 times and decided to try to fix it myself.

I am not familiar with lua, so any comments and suggestions are welcome.

### DIFF
--- a/lib/awful/prompt.lua
+++ b/lib/awful/prompt.lua
@@ -256,9 +256,15 @@ local function history_add(id, command)
     end
 end
 
+local function charlen(text, position)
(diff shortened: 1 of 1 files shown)


Skeevert:
Found

Skeevert:
Found some bugs with other hotkeys processing in my implementation, working to fix them

Skeevert:
My mistake. Tested wrong file version.

Issue #2725, opened 2019-03-11
Rule source dependencies are not honored


jcrd:
Output of `awesome --version`:
```
awesome v4.3-150-gf69b72ab (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • xcb-errors support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```
**How to reproduce the issue:**
```
local awful = require("awful")

awful.rules.add_rule_source("source1", function () print("source1") end)
awful.rules.add_rule_source("source2", function () print("source2") end, {"source1"})
```
**Actual result:**

"source2" is printed before "source1".

**Expected result:**

source2 should be run after its dependency, source1.

psychon:
I took a quick look:
https://github.com/awesomeWM/awesome/blob/0642d9296749027d86b2276f5490dc22f6cb776a/lib/awful/rules.lua#L301-L308
`add_rule_source` reverses the result of the topological sort. It repeatedly prepends (notice the argument `1`) the next item to the list of rule sources. Thus, the effect of `depends_on` and `precede` are exactly reversed, I guess..?
(What does this mean for the AwesomeWM-internal calls to `add_rule_source`? Are the dependencies the wrong way around in here, too?)

psychon:
Wait a second. I think everything is working as intended. Rule sources are not meant to actually apply anything, but to only collect the things that should be done. For example, the rule source for `awful.rules` just collects the matching rules and saves the properties and callbacks in the provided arguments:
https://github.com/awesomeWM/awesome/blob/0642d9296749027d86b2276f5490dc22f6cb776a/lib/awful/rules.lua#L330-L338

Thus, code that runs *later* overwrites settings done by *earlier* rule sources. Thus, the later a rule source runs, the higher is its priority.

@jcrd Does this make sense to you?

@Elv13 My only question now is: Should the order of callbacks be reversed as well? So that callbacks which are appended to the `callbacks` arguments last are run first?

Code:
```lua
awful.rules.add_rule_source("source1", function (c, props, callback) table.insert(callback, function() print("in cb of source1") end) end)
awful.rules.add_rule_source("source2", function (c, props, callback) table.insert(callback, function() print("in cb of source2") end) end, {"source1"})
```
Output when a new client appears:
```
in cb of source2
in cb of source1
```
[...]

PR #3705, opened 2022-09-26
Fix rule source order

kosorin:
Default rule source order is:

1. `awful.rules`
2. `awful.spawn_once`
3. `awful.spawn`

I wanted to use `fix_startup_id` function as suggested in the documentation, but the function put the `snid` rule source at the end.

1. `awful.rules`
2. `awful.spawn_once`
3. `awful.spawn`
4. `snid`

It should be the first rule because the `precede` parameter is `{"awful.spawn", "awful.rules"}`.

There is a mismatch between the code and the comments for `depends_on` and `precede` parameters in `ruled.client.add_rule_source`/`gears.matcher.add_matching_function` functions.

- `depends_on` -  A list of names of sources this source depends on (**sources that must be executed *before* `name`**).
- `precede` -  A list of names of sources this source has a priority over.

A quick fix would be to swap the parameters (I guess) however that would break the API.

### DIFF
--- a/lib/gears/matcher.lua
+++ b/lib/gears/matcher.lua
@@ -431,7 +431,7 @@ function matcher:add_matching_function(name, callback, depends_on, precede)
 
     for _, v in ipairs(res) do
         if callbacks[v] then
-            table.insert(self._matching_source, 1, {
+            table.insert(self._matching_source, {
                 callback = callbacks[v],
                 name     = v
             })
--- a/lib/ruled/client.lua
+++ b/lib/ruled/client.lua
@@ -264,7 +264,7 @@ end
 --
 -- @rulesources awful.rules
 
-crules:add_matching_rules("awful.rules", {}, {"awful.spawn"}, {})
+crules:add_matching_rules("awful.rules", {}, {}, {"awful.spawn"})
 
 -- Add startup_id overridden properties
 local function apply_spawn_rules(c, props, callbacks)
@@ -289,7 +289,7 @@ end
 --
 -- @rulesources awful.spawn
 
-module.add_rule_source("awful.spawn", apply_spawn_rules, {}, {"awful.rules"})
+module.add_rule_source("awful.spawn", apply_spawn_rules, {"awful.rules"}, {})
 
 local function apply_singleton_rules(c, props, callbacks)
     local persis_id, info = c.single_instance_id, nil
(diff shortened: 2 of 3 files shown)


Aire-One (approved):
LGTM!

@Elv13 can you double-check this change?

Elv13:
> @Elv13 can you double-check this change?

I need to sit down and have a longer look at the problem and solution here. The tests are failing and it seems the tests are right. Even if they are wrong, this would be an API change and break *a lot* of existing rules. The meaning of "higher" and "lower" priority is a bit confusing. An higher priority rules needs to be executed *last* so it can override the values of lower priority sources. This makes the dependency order look upside down, but it really isn't. It also make writing code error prone because it is unintuitive.

Still there *is* a bug to be solved here, the premise of the PR and the snid part need work. It will need a lot of new tests too, which wont be fun to write to say the least. I suspect the most trivial way to fix this bug is to create a new source and document that. The "most useful" way to solve this bug is for me to finish the patches which add support for cgroups (on Linux) and (optionally, on systems using it) systemd. It's like 100 lines, it just need time I don't have... This way SNIDs can go rot in hell and we would support the modern way of handling this.

kosorin:
> The meaning of "higher" and "lower" priority is a bit confusing. An higher priority rules needs to be executed _last_ so it can override the values of lower priority sources. This makes the dependency order look upside down, but it really isn't. It also make writing code error prone because it is unintuitive.

Well yes, it's really confusing. I just don't see how that's correct.

- `A` (depends_on = `{}`, precede = `{}`)
- `B` (depends_on = `{ A, C }`, precede = `{}`)
- `C` (depends_on = `{}`, precede = `{ A }`)

The order I expect: `C`, `A`, `B`

From docs: "*depends_on* - sources that must be executed **before** *name*"
That means `A` and `C` should be executed before `B`.

Actual order: `B`, `A`, `C` => `B` is executed first. 

So the documentation is lying.

I see you already discussed this in #2725 and I'm not the only one who thinks that it's wrong (not just confusing).
As you said, using the words "higher" and "lower" instead of "depends on" and "precede" would make a little bit more sense I guess.

Elv13:
I still think this is more or a documentation problem than something we can solve with code. As the test shows, making the change breaks existing code (by making `add_matching_function` do nothing for code which previously worked as expected from the user point of view). We work very hard to preserve compatibility. Benign (and "correct") looking changes caused a pretty big backlash from users in the past. So in practice it is unmergeable, errata or not. 

Also, it's possible to add a source that creates the intended behavior.

PR #3709, opened 2022-09-29
Restore minimized sticky clients

kosorin:
If a sticky client is minimized on a tag not assigned to it, one has to switch to the tag to be able to restore the minimized client.
A minimized sticky client should be able to be restored with `awful.client.restore()`.


Aire-One (approved):
Thank you!

Elv13 (approved):

PR #3710, opened 2022-09-30
(BREAKING CHANGE) Fix the output of menubar.utils.parse_desktop_file().Actions to be locallized

SkyyySi:
This PR will change the behavior of `menubar.utils.parse_desktop_file()` to alter its `Actions` key. Before, it returned a list of strings that were the identifiers of each available action. Now, it will instead contain the `Name`, `Exec` and `Icon` keys for each action, [as described here](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#extra-actions) (depending on which keys were set by the `.desktop`-file).

### DIFF
--- a/spawn.c
+++ b/spawn.c
@@ -412,9 +412,21 @@ spawn_child_exited(pid_t pid, int status)
  *
  * @tparam string|table cmd The command to launch.
  * @tparam[opt=true] boolean use_sn Use startup-notification?
- * @tparam[opt=false] boolean stdin Return a fd for stdin?
- * @tparam[opt=false] boolean stdout Return a fd for stdout?
- * @tparam[opt=false] boolean stderr Return a fd for stderr?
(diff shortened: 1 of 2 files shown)


Elv13:
Hi,

Can we use a non-conflicting name and document it exists? Breaking the API for this doesn't sound like something we can merge. We work very hard to maintain config compatibility across releases, even if that means being bug-compatible. For bug level compatibility, we have something called "API levels", which allows to opt-in to behavior changes, but that's not well suited for this change. A much more labor intensive solution would be to replace those APIs with something which actually comply this the specs. This means `gears.desktop`/`gears.xdg_menu`/`gears.icon_theme` (and they could use the GDK APIs rather than hardcode some Lua). However, that's a very large project, and renaming to something like `actions_table` is a one-liner.

PR #3712, opened 2022-10-07
New ldoc themes

raven2cz:
# Draft: New ldoc themes

Add several themes according to discussion. Dynamic change. Support dark/light mode.

Open topics:
* @Elv13 svg images will need more our work. Maybe some cooperation here. Light/dark color is not integrated correctly to svg text and fill strokes. We have to find some solution.
* @actionless next topic for you about "look and feel" feelings and typography improvements. In common, there can be some css styling bugs. The documentation is huge, I don't try any page. This is just draft.

Demo is here:
https://raven2cz.github.io/awesome/

### DIFF
--- a/docs/ldoc.css
+++ b/docs/ldoc.css
@@ -1200,19 +1200,6 @@ pre .url {
     background-color: var(--summary-row-hover-background-color);
 }
 
-/*
-.svg_stroke,
-.svg_fill {
-    color: #fff;
-}
-
-svg {
-    color: #fff;
-    fill: #fff;
-    stroke: #fff;
-}
-*/
-
 img[src*='.svg'] {
   filter: var(--svg-filter);
 }
--- a/docs/ldoc.css
+++ b/docs/ldoc.css
@@ -1,10 +1,11 @@
 :root {
-    color-scheme: dark light;
-    supported-color-schemes: dark light;
+    /* color-scheme: dark light;
+    supported-color-schemes: dark light; */
 
     --sidebar-width: 14em;
     --sidebar-toggle-width: 24px;
     --sidebar-theme-select: 13em;
+    --svg-filter: invert(100%);
 
     --background-color: var(--background);
     --color: var(--foreground);
@@ -94,6 +95,8 @@
 }
 
 html[data-theme='light'] {
+    --svg-filter: invert(0%);
+
     /* basic theme */
     --background: #eff4ff;
     --foreground: #47555c;
@@ -171,7 +174,7 @@ html[data-theme='light'] {
     --color10D: #0e7c6b;
     --color10E: #bbccaa;
     --color10F: #fedc56;
-    --color110: #ffffff;  
+    --color110: #ffffff;
 }
 
 html[data-theme='dark'] {
@@ -1197,10 +1200,21 @@ pre .url {
     background-color: var(--summary-row-hover-background-color);
 }
 
-/* CSS for beautiful.fg_normal */
+/*
 .svg_stroke,
 .svg_fill {
-    color: var(--svg-stroke-fill-color);
+    color: #fff;
+}
+
+svg {
+    color: #fff;
+    fill: #fff;
+    stroke: #fff;
+}
+*/
+
+img[src*='.svg'] {
+  filter: var(--svg-filter);
 }
 
 #content>.components-relationship {
@@ -1250,6 +1264,7 @@ pre .url {
 }
 
 .img-object {
+    filter: var(--svg-filter);
     max-width: 100%;
     padding: 5px;
 }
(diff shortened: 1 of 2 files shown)


actionless (approved):
i think the current version is good for merging - in case of other potential improvements it might be better to roll a separate PR

Elv13:
Nice work!

@actionless I think we should wait to merge it until the svg issues are resolved, right now it look like this:

Screenshot_20221007_232958

With high contrast enabled, it looks like this

Screenshot_20221007_233054

So it *can* work. Since this is a little problematic regression for some users (especially on bad TFT screens, where contrast isn't good even in idea conditions), I suggest this gets fixed first. After that, yeah, we can merge it! After that I guess we can iterate on the default dark theme. 

@raven2cz what kind of display are you using? The default theme looks fine on my LCD, but a little off on my OLED (which is oversaturated, smartphone-like). In any case, those are small details, I am super grateful of all the groundwork you did here! Exited to merge this soon.

raven2cz:
* **Add Contrast and good visibility**
> what kind of display are you using? The default theme looks fine on my LCD, but a little off on my OLED (which is oversaturated, smartphone-like). In any case, those are small details, I am super grateful of all the groundwork you did here! Exited to merge this soon.

It is a reason why I used more themes, not just light and dark, because different contrast. Themes have almost **exactly defined colorschemes**, which cannot be changed. But some themes are good for lower contrast, some for middle, etc. 

For "blind" people is necessary to still use light mode. Awesome-dark is for "glowing" monitors. The one dark and monokai are for middle and low contrasts. From my side. Mainly I'm using monokai and sometimes one dark, because I have middle and low contrasts set. But I don't prefer light mode.

* **SVG Images Light/Dark themes**
The problem is that `currentcolor` in autogen svg images doesn't support dynamic change of css variables. I don't know why, but all my actual tests failed. Just a direct usage of color works, but it is not for us. In addition, my tests just change the text color, not other colors of fill/stroke. I don't find correct wat with `currentcolor` feature. If anybody here has some idea, please provide me some example of usage.

Next problem, that there are still images which are inserted by `img` element and `src=*.svg`.

But I've got a idea. Fast, simple and effective without huge changes, generation of images, etc. I add revert filter and xpath to apply to any img-object and svg files. In actual case, I revert 100% of colors from original images which are designed for light mode. I'm happy with it. I haven't better solution in common scope now. But for first merge, it is not bad, not bad.

Check it in demo 
https://raven2cz.github.io/awesome/core_components/tag.html

actionless:
> The problem is that currentcolor

as far as i know, in modern browsers you can style with CSS the paths inside SVG almost same as in HTML

raven2cz:
> > The problem is that currentcolor
> 
> as far as i know, in modern browsers you can style with CSS the paths inside SVG almost same as in HTML

yes, I was not successful with it for awesome exported images. But the described filter in prev post works relatively great. There are around 500 images in doc.

Elv13 (dismissed):

Elv13:
@raven2cz Is there anything else you want to do before we merge this? With the new filter, the dark theme is very readable. It would be nice to "really" implement all colors, but i think we can iterate for that?

raven2cz:
> @raven2cz Is there anything else you want to do before we merge this? With the new filter, the dark theme is very readable. It would be nice to "really" implement all colors, but i think we can iterate for that?

yes, this first iteration is done. Yesterday, I go through all pages, fast check all and it is ok from my side. Mapping of image colors is next iteration cycle.

raven2cz:
@actionless @Aire-One There is missing second approval. @Elv13 cannot merge this themes without it. Thanks for coop.

Elv13:
I can merge it, but I need to override the checks. If there is no comment by Friday, I will.

docs/ldoc.css
```diff
@@ -679,19 +1193,31 @@ pre .url { color: #272fc2; text-decoration: underline; }
...
 
-/* CSS for beautiful.fg_normal */
-.svg_stroke, .svg_fill {
-    color: black;
+svg {
+    color: #fff;
+    fill: #fff;
+    stroke: #fff;
 }
+*/
```
Aire-One, line 1214: Should we remove these comment blocks?

Aire-One (approved):

Aire-One:
> @actionless @Aire-One There is missing second approval. @Elv13 cannot merge these themes without it. Thanks for coop.

Sorry for the delay @raven2cz. I'm a little out of the loop here, but I think it's fine to merge these changes.

Thank you a lot for all the effort you're putting on the website/apidoc redesign. :rocket:

docs/ldoc.css
raven2cz: Done. However, it will now require a new review release from your side...

Aire-One (approved):

Elv13 (approved):

actionless:
thanks a lot @raven2cz i wasn't reviewing it during wip, but i quickly went through end result - looking really slick 🔥

PR #3719, opened 2022-10-11
fix floating snap

lilykiwi:
makes snap & aerosnap respect padding & workarea, adds snapper_gap padding to build_placement subcalls using args


Elv13 (approved):

Aire-One (approved):
Thank you, @lilykiwi

PR #3722, opened 2022-10-21
awful.widget: Add a new inputbox widget to type and get text.

Crylia:
This is a new widget to type and recieve text from that inputbox. It should follow most typing standards like highlighting etc but I'm not 100% done with every key just the major one.

I probably need some help with the documentation, I tried to mimic some other files like awful.promt but its rather confusing getting behind when to use what. I also hope that the code it more or less ok since I also tried to mimic various modules

I havent written any tests yet but I'll do that later

### Currently this widget can do:

- Type text
- Delete text
- Paste text
- Select text
- Alter the selected text
- Use ctrl + arrow keys to "jump" words
- Callbacks once started, stopped and for every key press
- Styling options

### Things TODO:

- [ ] Copy text (needs rewrite of `selection.c`)
- [ ] Shift+Ctrl actions
- [ ] More key combinations
- [ ] Font variable

### Demonstration

https://user-images.githubusercontent.com/50151522/197140652-4d1d4395-0984-4155-bb86-89465c948918.mp4

### DIFF
--- a/lib/awful/widget/inputbox.lua
+++ b/lib/awful/widget/inputbox.lua
@@ -1,0 +1,530 @@
+---------------------------------------------------------------------------
+-- A widget to write text in
+--@DOC_wibox_widget_defaults_inputtextbox_EXAMPLE@
+--
+-- @author Rene Kievits
+-- @copyright 2022, Rene Kievits
+-- @module awful.widget.inputtextbox
+---------------------------------------------------------------------------
+
+local setmetatable = setmetatable
+local awful = require("awful")
+local beautiful = require("beautiful")
+local wibox = require("wibox")
+local gtable = require("gears.table")
+local gobject = require("gears.object")
+local gstring = require("gears.string")
+local keygrabber = require("awful.keygrabber")
+
+local capi =
+{
+    selection = selection,
+    mousegrabber = mousegrabber,
+    mouse = mouse,
+}
+
+local inputbox = { mt = {} }
+inputbox._private = {}
+inputbox._private.highlight = {}
+
+--- Formats the text with a cursor and highlights if set.
+local function text_with_cursor(text, cursor_pos, self)
+    local char, spacer, text_start, text_end
+
+    local cursor_fg = self._private.cursor_fg
+    local cursor_bg = self._private.cursor_bg
+    local text_color = self._private.fg
+    local hint_text = self._private.hint_text
+    local hint_fg = self._private.hint_fg or beautiful.fg_normal
+    local highlight_bg = self._private.highlight_bg
+    local highlight_fg = self._private.highlight_fg
+
+    if text == "" then
+        return "<span foreground='" .. hint_fg .. "'>" .. hint_text .. "</span>"
+    end
+
+    if #text < cursor_pos then
+        char = " "
+        spacer = ""
+        text_start = gstring.xml_escape(text)
+        text_end = ""
+    else
+        local offset = 0
+        if #text:sub(cursor_pos, cursor_pos) == -1 then
+            offset = 1
+        end
+        char = gstring.xml_escape(text:sub(cursor_pos, cursor_pos + offset))
+        spacer = " "
+        text_start = gstring.xml_escape(text:sub(1, cursor_pos - 1))
+        text_end = gstring.xml_escape(text:sub(cursor_pos + offset + 1))
+    end
+
+    if self._private.highlight and self._private.highlight.cur_pos_start and self._private.highlight.cur_pos_end then
+        -- split the text into 3 parts based on the highlight and cursor position
+        local text_start_highlight = gstring.xml_escape(text:sub(1, self._private.highlight.cur_pos_start - 1))
+        local text_highlighted = gstring.xml_escape(text:sub(self._private.highlight.cur_pos_start,
+            self._private.highlight.cur_pos_end))
+        local text_end_highlight = gstring.xml_escape(text:sub(self._private.highlight.cur_pos_end + 1))
+
+        return "<span foreground='" .. text_color .. "'>" .. text_start_highlight .. "</span>" ..
+            "<span foreground='" .. highlight_fg .. "'  background='" .. highlight_bg .. "'>" ..
+            text_highlighted ..
+            "</span>" .. "<span foreground='" .. text_color .. "'>" .. text_end_highlight .. "</span>"
+    else
+        return "<span foreground='" .. text_color .. "'>" .. text_start .. "</span>" ..
+            "<span background='" .. cursor_bg .. "' foreground='" .. cursor_fg .. "'>" ..
+            char .. "</span>" .. "<span foreground='" .. text_color .. "'>" .. text_end .. spacer .. "</span>"
+    end
+end
+
+--- The inputbox border color
+--
+-- @DOC_awful_widget_inpubox_border_color_EXAMPLE@
+--
+-- @property border_color
+-- @tparam[opt=gears.color] string border_color
+-- @see gears.color
+-- @propemits true false
+-- @propbeautiful
+function inputbox:set_border_color(v)
+    self._private.border_color = v
+    self:emit_signal("property::border_color", v)
+end
+
+--- The inputbox border width
+--
+-- @DOC_awful_widget_inpubox_border_width_EXAMPLE@
+--
+-- @property border_width
+-- @tparam[opt=0] number|nil border_width
+-- @negativeallowed false
+-- @propertyunit pixel
+-- @propemits true false
+-- @propbeautiful
+function inputbox:set_border_width(v)
+    self._private.border_width = v
+    self:emit_signal("property::border_width", v)
+end
+
+--- The inputbox background color
+--
+-- @DOC_awful_widget_inpubox_bg_EXAMPLE@
+--
+-- @property bg
+-- @tparam[opt=gears.color] string foreground
+-- @see gears.color
+-- @propemits true false
+-- @propbeautiful
+function inputbox:set_bg(v)
+    self._private.bg = v
+    self:emit_signal("property::bg", v)
+end
+
+--- The text foreground color
+--
+-- @DOC_awful_widget_inpubox_fg_EXAMPLE@
+--
+-- @property string
+-- @tparam[opt=gears.color] string foreground
+-- @see gears.color
+-- @propemits true false
+-- @propbeautiful
+function inputbox:set_fg(v)
+    self._private.fg = v
+    self:emit_signal("property::fg", v)
+end
+
+--- The shape of the inputbox
+--
+-- @DOC_awful_widget_inpubox_shape_EXAMPLE@
+--
+-- @property shape
+-- @tparam[opt=gears.shape.rectangle] shape|nil shape
+-- @see gears.shape
+-- @propemits true false
+-- @propbeautiful
+function inputbox:set_shape(v)
+    self._private.shape = v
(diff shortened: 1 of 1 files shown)


lib/awful/widget/inputbox.lua
```diff
@@ -0,0 +1,530 @@
...
+            elseif cursor_pos > #self._private.text + 1 then
+                cursor_pos = #self._private.text + 1
+            end
+        end
+
+        -- Update cycle
+        self:update(self._private.text, cursor_pos)
+
+        if callbacks.pressed then
+            callbacks.pressed(key, mod_keys)
```
sclu1034, line 404: Users are already used to the reverse order of `mods, key` from `awful.key`. That shouldn't change here.

lib/awful/widget/inputbox.lua
```diff
@@ -0,0 +1,530 @@
...
+
+    -- Change the cursor to "xterm" on hover over
+    local old_cursor, old_wibox
+    ret.widget:connect_signal(
+        "mouse::enter",
+        function()
+            local w = capi.mouse.current_wibox
+            if w then
+                old_cursor, old_wibox = w.cursor, w
+                w.cursor = "xterm"
```
sclu1034, line 506: The "hover" cursor should be configurable.

lib/awful/widget/inputbox.lua
```diff
@@ -0,0 +1,530 @@
...
+            end),
+            awful.button({}, 3, function()
+                -- TODO: Figure out how to paste with highlighted support
+                -- Maybe with a signal?
+            end)
+        }
+    )
+
+    -- Change the cursor to "xterm" on hover over
+    local old_cursor, old_wibox
```
sclu1034, line 499: What's `old_wibox` for?
It's set and unset in the handlers below, but never used.

lib/awful/widget/inputbox.lua
```diff
@@ -0,0 +1,530 @@
...
+-- @tparam[opt=beautiful.inputbox_cursor_fg] string args.cursor_fg Cursor foreground color
+-- @tparam[opt=beautiful.inputbox_highlight_bg] string args.highlight_bg Highlight background color
+-- @tparam[opt=beautiful.inputbox_highlight_fg] string args.highlight_fg Highlight foreground color
+-- @treturn awful.widget.inputbox The inputbox widget.
+-- @constructorfct awful.widget.inputbox
+function inputbox.new(args)
+    args = args or {}
+
+    local ret = gobject { enable_properties = true }
+    gtable.crush(ret, inputbox, true)
```
sclu1034, line 426: This object is not returned from the constructor, so the methods assigned to `ret` and the signals emitted from it won't be accessible to the user.

You have two options:
- Assign your methods to the actual widget. Downside is you need to do extra stuff to hide the widget's own API, and it might get complicated with a template
- Return `ret` from the constructor and implement `.fit` and `:layout` methods to make `inputbox` itself a container that renders the inner textbox

lib/awful/widget/inputbox.lua
```diff
@@ -0,0 +1,530 @@
...
+
+        if callbacks.pressed then
+            callbacks.pressed(key, mod_keys)
+        end
+    end)
+end
+
+--- Creates a new inpubox widget
+-- @tparam table args Arguments for the inpubox widget
+-- @tparam string args.text The text to display in the inpubox
```
sclu1034, line 411: Generating the documentation locally will tell you issues like this, where the optional value is missing.

sclu1034 (changes requested):
Only a cursory look so far.

1. Your naming for the widget is inconsistent. Sometimes it's `inputbox` (or `inpubox`), sometimes `inputtextbox`.
2. You added includes for example files in some documentation comments, but those files don't exist.
3. The API exposed by the `.run` function should be reworked
   - It's inconsistent in that it accepts callbacks for some things, but emits signals for other things
   - It's heavily based on the implementation detail of the keygrabber.
   
   Instead, I'd prefer to
   - only use signals, no callbacks
   - create it based on intention, not implementation
   - use/provide well known interfaces instead of rolling your own
   
   My suggestion:
  
   - `callback.started` turns into `emit_signal("focus")`
   - `callback.stopped` turns into `emit_signal("blur")`
   - `callback.pressed` turns into `emit_signal("pressed", ...)`
   - `emit_signal("keygrabber::stop", nil)` turns into `emit_signal("cancel")`
   - `emit_signal("keygrabber::stop", text)` turns into `emit_signal("submit", text)`
4. There should probably be a `property::text` signal

lib/awful/widget/inputbox.lua
Crylia: I use it to save `mouse.current_wibox` so I can set a cursor to it. I figured it would be better than to call `mouse.current_wibox` twice (e.g. `old_wibox.cursor = "xterm"`)

I could probably rename it since "old" is not too accurate in this contect, better would be "current_wibox"

lib/awful/widget/inputbox.lua
```diff
@@ -0,0 +1,530 @@
...
+
+--- Set the widget text markup
+-- @property text
+-- @tparam[opt=""] string text
+-- @propemits true false
+function inputbox:set_text_markup(text)
+    self.widget:get_children_by_id("input_text")[1]:set_markup(text)
+    self:emit_signal("property::markup", text)
+end
+
```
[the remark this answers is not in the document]
Crylia: Yeah I saw that too and couldn't find an immediate solution. I probably will make it a private function and add a `set_text` public function

lib/awful/widget/inputbox.lua
sclu1034: Also, `w` is the variable you save to temporarily to assign the cursor.
`old_wibox` is, as I said, never used.

sclu1034: Ah, yes. Missed that line.

lib/awful/widget/inputbox.lua
Crylia: I tired to implement that with #3421 but I couldn't figure out how exactly since there is now wibox.widget.template in place of a base widget.

sclu1034 (changes requested):

lib/awful/widget/inputbox.lua
```diff
@@ -195,22 +170,26 @@ function inputbox:stop()
...
+    self:connect_signal("cancel", function()
+        self:stop()
+        self:emit_signal("stopped")
+    end)
+
+    -- Emitted when the keygrabber should submit the text
+    self:connect_signal("submit", function(text)
         self:stop()
-        if callbacks.stopped then callbacks.stopped(self:get_text()) end
+        self:emit_signal("stopped", text)
```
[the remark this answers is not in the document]
Crylia: >  I don't really see a use in knowing specifically about a keygrabber starting/stopping. The thing I would want to know is "the input is focused".

Ah that makes way more sense now, yeah I can see that `focus` and `blur` would make more sense.

lib/awful/widget/inputbox.lua
```diff
@@ -398,21 +374,62 @@ function inputbox:run(callbacks)
...
+        },
+        widget = wibox.container.background,
+        bg = "#212121",
+        fg = "#F0F0F0",
+        border_color = "#414141",
+        border_width = 2,
+        shape = gshape.rounded_rect,
+        forced_width = 300,
+        forced_height = 50,
+    },
```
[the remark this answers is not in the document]
Crylia: No, I just thought that if the user doesn't provide its own template that it would make sense to provide a widget that you can see easier. (No borders but the same background could make the user think that there is nothing)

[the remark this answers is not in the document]
sclu1034: But the other side of the coin is that it now becomes harder to embed the widget into any custom widget tree, because the user needs to undo your opinionated default styling.
E.g. we don't usually force a rounded shape, or restrict widget sizes by default.

The general usage of the widget system is "when you want to add/change something, wrap it". And in the general Unix philosophy "do one thing well", each widget only does the specific things it is designed/intended for.

Maybe another point to hear opinions for, though.

actionless:
wouldn't it be better to implement it using existing `prompt` and `popup` widgets?

Crylia:
> wouldn't it be better to implement it using existing `prompt` and `popup` widgets?

I made it so it doesn't use popup because I originally created this widget to be used in a form for example (in my case for a network widget). The preview was just the inputbox because the form wasn't even close to be done and would've distracted.

I don't really like the idea of using the existing prompt for this as its more for a "fire and forget" action than an input field where you might come back later and reedit it. This widget is more of a general approach so it can be everything where you can type text (maybe it can be used as/in promt?)

Crylia:
I'm going to clean up the code later I just wanted to push it for now, is there anything else I should change(apart from documentation, styling arguments and left over stuff)?

sclu1034 (changes requested):
With "container widget" and "widget templates", you're now mixing two different systems now that don't go well together.
Since there is no need for anything besides a `wibox.widget.textbox` to implement a textinput, I don't really see the need for a fully fledged template anymore. And the inner textbox should be fully transparent to the user of this widget, especially because if they have access, they can easily mess up your input state.

To the user, this widget should behave like a "content widget" (e.g. `wibox.widget.textbox` or `wibox.widget.imagebox`).

Therefore:
- `:get_children()`, `:set_children()`, `:set_widget()`, `:reset()` and `:set_widget_template()` removed
- `self._private.widget` is set to an instance of `wibox.widget.textbox` in the constructor, and never changed

actionless:
i still can't understand your motivation to not use prompt widget, could you give like real examples instead of floaty general words?

actionless (changes requested):
after a brief review i see huge parts of logic literally same as in prompt widget (especially related to handling input hotkeys)  - smth really need to be done about it

Aire-One:
> after a brief review i see huge parts of logic literally same as in prompt widget (especially related to handling input hotkeys) - smth really need to be done about it

I would suggest that we refactor `awful.widget.prompt` to use the more generic `inputbox` widget as its base.

In this regard, it should be cared that inputbox includes all the corner cases (string manipulation/cursor movement with characters from different languages) already managed by prompt.

Also, I would have made this new widget part of `wibox.widget`, not `awful.widget`.

actionless:
> I would suggest that we refactor awful.widget.prompt to use the more generic inputbox widget as its base.

yeah - either this way or another - just the fact is that all this input and hotkey code is same

Crylia:
Sorry for not responding for a while  I've been very busy. I would happily change it to a `wibox.widget` and apply all requested changed ofc. I'll try to quickly sum up what I have in mind for this widget apart from the current implemented stuff.
- Password mode(only start input *****) and (hopefully) secure storage of that password
- Maybe mouse support (might be possible by checking where the mouse is on that widget but I need more research on that)
- Copy text (once I finished the part in selection.c)

These are the main goals that I would like to have working.
On the password mode I'm actually unsure if it would fit into this widget or if it should be separate, there I would like to hear some other thoughts.

unai-ndz:
I would say add a simple password mode, just an argument to **** the input and leave handling/encrypting to the user or for another widget based on this one.

sclu1034:
> Password mode(only start input *****) and (hopefully) secure storage of that password
> On the password mode I'm actually unsure if it would fit into this widget or if it should be separate, there I would like to hear some other thoughts.

The storage would not be in scope for any widget. As a general encryption library, maybe, but we don't even have a concept for regular I/O.

> the part in selection.c

That should be a separate PR, then, as it is a standalone feature.

Issue #3721, opened 2022-10-20
Mousegrabber without specifying cursor


Crylia:
Currently you have to use the mousegrabber like following

```lua
mousegrabber.run(
  function() return end,
  "arrow"
)
```

You **have** to specify a cursor which can be highly infurating if you are trying to have a different cursor when entering a widget.

```lua
mousegrabber.run(
  function() return end,
  ""
)
```

Passing just an empty string would then result in no change to the cursor.

sclu1034:
`nil` would make more sense for a "no change" value.
Either way, it just has to be mapped to `XCB_NONE` in C.

PR #3723, opened 2022-10-21
mousegrabber: add nil option if no cursor is wanted #3721

Crylia:
I added the option to pass no cursor see #3721 

I hope I did not forget something since I have 0 experience in lua calls in c.


sclu1034:
1. Please squash your commits so that the change in formatting doesn't show up in the first place
2. You still changed the formatting for `luaA_mousegrabber_run`
3. Please keep the `luaA_warn` for invalid cursor names

PR #3728, opened 2022-10-27
Menu items can be disabled (+ menu separator)

kosorin:
This PR allow menu items to be disabled.

I would like to add menu separators (as regular items with custom style) but I need them to be disabled. So that the separators don't change style on hover and are skipped when navigating with a keyboard.

Usage:

```lua
awful.menu {
    { "foo" },
    { "bar", enabled = false },
    { "baz" },
}
```

### Edit:

I think a menu separator is a basic feature and should be in Awesome library. The new "enabled" property together with existing `new` function allows easy implementation of the separator.

```lua
awful.menu {
    { "foo" },
    { new = awful.menu.separator },
    { "baz" },
}

function awful.menu.separator(parent, args)
    args = args or {}
    return {
        enabled = false,
        height = args.theme.height * 0.5,
        widget = wibox.widget.separator {
            span_ratio = args.span_ratio or 1,
            thickness = args.thickness or args.theme.border_width or 1,
            color = args.color or args.theme.border,
        }
    }
end
```

Maybe it could be defined directly as an item with the `new` property:
```lua
awful.menu {
    { "foo" },
    awful.menu.separator,
    { "baz" },
}
```

### DIFF
--- a/lib/awful/menu.lua
+++ b/lib/awful/menu.lua
@@ -140,6 +140,8 @@ local function load_theme(a, b)
     ret.width = a.width or b.menu_width or b.width or
                 fallback.menu_width or dpi(100)
     ret.font = a.font or b.font or fallback.menu_font or fallback.font
+    ret.opacity_disabled = a.opacity_disabled or b.menu_opacity_disabled or b.opacity_disabled or
+                           fallback.menu_opacity_disabled or 0.5
     for _, prop in ipairs({"width", "height", "menu_width"}) do
         if type(ret[prop]) ~= "number" then ret[prop] = tonumber(ret[prop]) end
     end
@@ -239,17 +241,37 @@ local function check_access_key(self, key)
    end
 end
 
+local function select_next(self, current, direction)
+    local count = #self.items
+    if count < 1 then
+        return
+    end
+
+    local index = current or self.sel or 0
+    for _ = 1, count do
+        index = index + direction
+        if index < 1 then
+            index = count
+        elseif index > count then
+            index = 1
+        end
+
+        local item = self.items[index]
+        if item and item.enabled then
+            self:item_enter(index)
+            return
+        end
+    end
+end
 
 local function grabber(self, _, key, event)
     if event ~= "press" then return end
 
     local sel = self.sel or 0
     if gtable.hasitem(menu.menu_keys.up, key) then
-        local sel_new = sel-1 < 1 and #self.items or sel-1
-        self:item_enter(sel_new)
+        select_next(self, sel, -1)
     elseif gtable.hasitem(menu.menu_keys.down, key) then
-        local sel_new = sel+1 > #self.items and 1 or sel+1
-        self:item_enter(sel_new)
+        select_next(self, sel, 1)
     elseif sel > 0 and gtable.hasitem(menu.menu_keys.enter, key) then
         self:exec(sel)
     elseif sel > 0 and gtable.hasitem(menu.menu_keys.exec, key) then
@@ -267,7 +289,7 @@ end
 function menu:exec(num, opts)
     opts = opts or {}
     local item = self.items[num]
-    if not item then return end
+    if not item or not item.enabled then return end
     local cmd = item.cmd
     if type(cmd) == "table" then
         local action = cmd.cmd
@@ -323,7 +345,7 @@ end
 function menu:item_enter(num, opts)
     opts = opts or {}
     local item = self.items[num]
-    if num == nil or self.sel == num or not item then
+    if num == nil or self.sel == num or not item or not item.enabled then
         return
     elseif self.sel then
         self:item_leave(self.sel)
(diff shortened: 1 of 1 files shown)


lib/awful/menu.lua
```diff
@@ -239,17 +241,37 @@ local function check_access_key(self, key)
    end
 end
 
+function menu:select_next(direction, current)
```
Aire-One, line 244: It looks like an internal function we don't want to expose. 

Change the declaration to a local function (the signature will also need to be changed to receive the menu instance).

lib/awful/menu.lua
```diff
@@ -239,17 +241,37 @@ local function check_access_key(self, key)
...
+    end
+
+    local index = current or self.sel or 0
+    for _ = 1, count do
+        index = index + (direction or 1)
+        if index < 1 then
+            index = count
+        elseif index > count then
+            index = 1
+        end
```
Aire-One, line 257: I think we are trying to loop over the items table as in a circular array.

Please remove this for loop and use the modulo operator instead to get the new index.

lib/awful/menu.lua
```diff
@@ -140,6 +140,8 @@ local function load_theme(a, b)
     ret.width = a.width or b.menu_width or b.width or
                 fallback.menu_width or dpi(100)
     ret.font = a.font or b.font or fallback.menu_font or fallback.font
+    ret.opacity_disabled = a.opacity_disabled or b.menu_opacity_disabled or b.opacity_disabled or
+                           fallback.menu_opacity_disabled or 0.5
```
Aire-One, line 144: I'm not sure how this widget works. Should we also have a new `beautiful` variable for this?

lib/awful/menu.lua
```diff
@@ -729,6 +764,7 @@ function menu.new(args, parent)
     args = args or {}
     args.layout = args.layout or wibox.layout.flex.vertical
     local _menu = table_update(object(), {
+        select_next = menu.select_next,
```
Aire-One, line 767: Hm... Why do we want to expose this `select_next` function? (Am I missing something?)

lib/awful/menu.lua
```diff
@@ -669,7 +698,13 @@ function menu.entry(parent, args) -- luacheck: no unused args
         layout:set_right(submenu)
     end
 
+    local enabled = args.enabled == nil and true or args.enabled
```
Aire-One, line 701: This new `enable` argument has to be documented.

lib/awful/menu.lua
```diff
@@ -443,6 +465,7 @@ function menu:add(args, index)
         return
     end
     item.parent = self
+    item.enabled = item.enabled == nil and true or item.enabled
```
Aire-One, line 468: The `item.enable` property is supposed to be a non-null boolean, based on the `menu.entry` constructor implementation below. So I would remove this extra test and keep `item.enable` as-is.

lib/awful/menu.lua
```diff
@@ -458,6 +481,9 @@ function menu:add(args, index)
         button({}, 3, function () self:hide() end),
         button({}, 1, function ()
             local num = gtable.hasitem(self.items, item)
+            if not num or not item.enabled then
+            	return
```
Aire-One, line 485: :+1:

lib/awful/menu.lua
```diff
@@ -669,7 +698,13 @@ function menu.entry(parent, args) -- luacheck: no unused args
         layout:set_right(submenu)
     end
 
+    local enabled = args.enabled == nil and true or args.enabled
+    if not enabled then
+        layout:set_opacity(args.theme.opacity_disabled)
+    end
```
Aire-One, line 704: We should have a setter defined for this behavior. The user could alter the `enable` value latter and this wouldn't update the opacity.

Aire-One (commented):
Thank you, @kosorin, for this PR, having the ability to disable `awful.menu` entries is a nice addition! 

This new feature needs to be properly documented, and should have examples demonstrating its usage.

The fact that an `enable == false` item has its `opacity` changed can probably be seen as an issue. I'm mostly fine with it, but it wouldn't surprise me that it annoys other persons...

The `awful.menu.separator` sound like a good idea. Be sure it doesn't clash with the existing `wibox.widget.separator` widget. Actually, I think `awful.menu.separator` should use `wibox.widget.separator` in its internal.

lib/awful/menu.lua
kosorin: The loop is necessary, because we want to cycle until we find the first enabled item. I use `for` to loop `count` times instead of `while true` to make sure we don't loop forever if all items are disabled.
Using just a modulo would not solve this.

kosorin (commented):
Thanks for review. 

I'll add documentation and other needed things when code will be ready and I'll bee sure it will be merged. I created this PR with the code as it is just to see if you're even interested.

> The fact that an enable == false item has its opacity changed can probably be seen as an issue. I'm mostly fine with it, but it wouldn't surprise me that it annoys other persons...

It's configurable, in beautiful theme or when creating items.

> The awful.menu.separator sound like a good idea. Be sure it doesn't clash with the existing wibox.widget.separator widget. Actually, I think awful.menu.separator should use wibox.widget.separator in its internal.

Yes, I use `wibox.widget.separator`, see example in the first comment.

Issue #3737, opened 2022-11-14
[Feature request] Use `left_ptr_watch` for a startup notification


akinokonomi:
Output of `awesome --version`:
awesome v4.3-1551-g5077c8381-dirty (Too long)

Hello!

The reason for this is that I use [Fuchsia cursor theme](https://github.com/ful1e5/fuchsia-cursor), and here is what `watch` vs. `left_ptr_watch` look like:

![image](https://user-images.githubusercontent.com/91767492/201551185-adc8bc5f-36af-4aa6-8ead-7f7ad841f439.png)

The latter one looks much more seamless to me.

I tried to patch it myself to submit a PR:

<details><summary>Click here</summary>

```diff
diff --git a/common/xcursor.c b/common/xcursor.c
index c4c691df5..356d3c68e 100644
--- a/common/xcursor.c
+++ b/common/xcursor.c
@@ -101,6 +101,7 @@ static char const * const xcursor_font[] =
     [XC_ul_angle] = "ul_angle",
     [XC_umbrella] = "umbrella",
     [XC_ur_angle] = "ur_angle",
+    [XC_left_ptr] = "left_ptr_watch",
     [XC_watch] = "watch",
     [XC_xterm] = "xterm",
 };
diff --git a/lib/awful/startup_notification.lua b/lib/awful/startup_notification.lua
index 9bc95236c..45ad9c639 100644
--- a/lib/awful/startup_notification.lua
+++ b/lib/awful/startup_notification.lua
@@ -18,7 +18,7 @@ local beautiful = require("beautiful")
 
 local app_starting = {}
 
-local cursor_waiting = "watch"
+local cursor_waiting = "left_ptr_watch"
 
 --- Show busy mouse cursor during spawn.
 -- @beautiful beautiful.enable_spawn_cursor
```

</details>

But nothing came out of it, because `XC_left_ptr` is a duplicate entry, and my LUA knowledge is not enough to go any deeper.

Just for reference, Openbox uses `left_ptr_watch` as well.

Thanks!

Elv13:
Use "left_ptr" without the C patch? `XC_left_ptr` is already handled by `xcursor.c`. There is no such thing as a left watch in the X11 enum.

akinokonomi:
> Use "left_ptr" without the C patch?

That's what I did at first, but it does not work. It would still show just the `watch` cursor (and would never remove it even after app was  started).

It actually shows `left_ptr_watch` after I patch the C file, but messes up everything else.

[...]

PR #3745, opened 2022-12-04
Rewrite the xcursor caching code

psychon:
Previously, the cache for xcursors consisted of a static array. We had a hardcoded list of supported cursor names and each one got assigned a position in this array. This hardcoded list means that one cannot simply use whatever cursor a cursor theme happens to provide, since the name needs to be in our list.

In this commit, the code is rewritten. Instead of a hardcoded list, a sorted array is now used as the cache, as provided by the BARRAY code from common/util.h. The array is sorted by the name of the cursor.

This change implies an API change for the xcursor code. Previously, one had to first translate a cursor name into a cache index (xcursor_font_fromstr()) and could then use this to actually get the cursor (xcursor_new()). With this commit, xcursor_new() is the only function provided by the cursor code and it directly gets a string as argument.

----

This PR is related to #3737. I am 80% sure that this PR implements (half of) that feature request by making it possible to use arbitrary cursors, as long as xcb-util-cursor can load a cursor by that name.

@akinokonomi Are you in a position to test this? (=Can you build awesome from a git branch and verify whether this does what you want?)

Personally, I only started awesome via `tests/run.sh /dev/null` and started xterm from the menu. The "waiting for startup" cursor briefly appeared. Nothing more was tested.

### DIFF
--- a/common/xcursor.c
+++ b/common/xcursor.c
@@ -22,112 +22,23 @@
 #include "common/xcursor.h"
 #include "common/util.h"
 
-#include <X11/cursorfont.h>
-
-static char const * const xcursor_font[] =
-{
-    [XC_X_cursor] = "X_cursor",
-    [XC_arrow] = "arrow",
-    [XC_based_arrow_down] = "based_arrow_down",
-    [XC_based_arrow_up] = "based_arrow_up",
-    [XC_boat] = "boat",
-    [XC_bogosity] = "bogosity",
-    [XC_bottom_left_corner] = "bottom_left_corner",
-    [XC_bottom_right_corner] = "bottom_right_corner",
-    [XC_bottom_side] = "bottom_side",
-    [XC_bottom_tee] = "bottom_tee",
-    [XC_box_spiral] = "box_spiral",
-    [XC_center_ptr] = "center_ptr",
-    [XC_circle] = "circle",
-    [XC_clock] = "clock",
-    [XC_coffee_mug] = "coffee_mug",
-    [XC_cross] = "cross",
-    [XC_cross_reverse] = "cross_reverse",
-    [XC_crosshair] = "crosshair",
-    [XC_diamond_cross] = "diamond_cross",
-    [XC_dot] = "dot",
-    [XC_dotbox] = "dotbox",
-    [XC_double_arrow] = "double_arrow",
-    [XC_draft_large] = "draft_large",
-    [XC_draft_small] = "draft_small",
-    [XC_draped_box] = "draped_box",
-    [XC_exchange] = "exchange",
-    [XC_fleur] = "fleur",
-    [XC_gobbler] = "gobbler",
-    [XC_gumby] = "gumby",
-    [XC_hand1] = "hand1",
-    [XC_hand2] = "hand2",
-    [XC_heart] = "heart",
-    [XC_icon] = "icon",
-    [XC_iron_cross] = "iron_cross",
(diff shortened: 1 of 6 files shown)


psychon:
> Nothing more was tested.

I had an idea!

```diff
diff --git a/awesomerc.lua b/awesomerc.lua
index c598a3e08..a5cad2eed 100644
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -191,6 +191,7 @@ screen.connect_signal("request::desktop_decoration", function(s)
         position = "top",
         screen   = s,
         -- @DOC_SETUP_WIDGETS@
+        cursor = "foobar",
         widget   = {
             layout = wibox.layout.align.horizontal,
             { -- Left widgets
```
results in
```
2022-12-04 09:38:34 W: awesome: xerror:1058: X error: request=CreateGlyphCursor (major 94, minor 0), error=BadValue (2)
2022-12-04 09:38:34 W: awesome: xerror:1058: X error: request=ChangeWindowAttributes (major 2, minor 0), error=BadCursor (6)
2022-12-04 09:38:34 W: awesome: xerror:1058: X error: request=ChangeWindowAttributes (major 2, minor 0), error=BadCursor (6)
```
Checking `ls /usr/share/icons/*/cursors`, I found `dot_box_mask`. That one works (and IMO looks confusing, but whatever...).

Will investigate the above X11 errors and report back.

psychon:
Ahhh, debian is using an old (ancient?) version of libxcb-cursor. The bug I found above was fixed 7 years ago: https://gitlab.freedesktop.org/xorg/lib/libxcb-cursor/-/commit/cf26479ece9ab9e04616bc10ba674d88a284e5b0

akinokonomi:
>  Are you in a position to test this? (=Can you build awesome from a git branch and verify whether this does what you want?)

Certainly. 

Compiled against LuaJIT 2.1 beta3, xcb-util-cursor 0.1.4.

Everything looks good. Did not notice any issue so far.

And after changing the startup cursor like this:

```diff
diff --git a/lib/awful/startup_notification.lua b/lib/awful/startup_notification.lua
index 9bc95236c..45ad9c639 100644
--- a/lib/awful/startup_notification.lua
+++ b/lib/awful/startup_notification.lua
@@ -18,7 +18,7 @@ local beautiful = require("beautiful")
 
 local app_starting = {}
 
-local cursor_waiting = "watch"
+local cursor_waiting = "left_ptr_watch"
 
 --- Show busy mouse cursor during spawn.
 -- @beautiful beautiful.enable_spawn_cursor
```

It does exactly what I want. Thank you!

Please let me know if I have to test something else.

psychon:
Thanks a lot for testing this!

actionless (approved):

Aire-One (approved):
It was a mildly challenging but interesting review since I haven't had read common/array.h before. Thank you, @psychon, for this PR.

psychon:
> since I haven't had read common/array.h before.

Don't go there! It is dark magic that wasn't touched in ages! ;-)

But yeah, that file is actually used a lot in the C code to manage lists, but I guess one seldomly has to actually interact with this file.

(Latest commit is from 2016 by yours truly and it removes a feature. Before that, there were one commit in 2015, 2014, 2011, 2010. I guess this must mean that this is good code that doesn't need any touching...)

actionless:
this would require manual merging because of codecov

Aire-One:
@mergifyio rebase

PR #3760, opened 2023-01-11
Update 07-my-first-awesome.md

Horuyuki:
The original documentation told me to just copy the `theme.lua` file, but that would yield the error: Error
I realized that the `theme.lua` by itself would call the surrounding files like so: `themes_path.. "default/example.png"

Would this change be possible?


Aire-One (approved):
I'm ok with this change.

actionless (approved):

Elv13 (approved):

PR #3765, opened 2023-01-16
Update taglist when tag.volatile value is changed

kosorin:
The `volatile` property is supported by the taglist (theme variables `beautiful.taglist_bg_volatile` and `beautiful.taglist_fg_volatile`) but taglist is not updated when the property is changed.


Aire-One (approved):

Elv13 (approved):

Elv13:
thanks!

Issue #3767, opened 2023-01-21
Invalid position middle


madhur:
Output of `awesome --version`:

➜ awesome --version                                                                                                                                                                                                                                            
awesome v4.3-1586-gee066345 (Too long)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no

**How to reproduce the issue:**

```
local naughty = require("naughty")

naughty.notify {
    text = "test",
    position = "middle"
}
```

**Actual result:**

Error: Invalid position middle

**Expected result:**

Notification should appear at center of screen

As per the docs,
https://awesomewm.org/apidoc/core_components/naughty.notification.html#naughty.notification

`middle` is a valid position property.

Aire-One:
```lua
local naughty = require("naughty")

naughty.notify {
    text = "test",
    position = "middle"
}
```

It works fine for me.

> Actual result:
> 
> Error: Invalid position middle

What is the exact output you get?

If I try to pass an invalid string to position, I have an error pointing to https://github.com/awesomeWM/awesome/blob/master/lib/naughty/core.lua#L703 because there is no `naughty.notifications[s][MY INVALID POSITION]`.

Note that `naughty.notifications[s]` is initialized with https://github.com/awesomeWM/awesome/blob/master/lib/naughty/core.lua#L211-L219. We can see the `middle` property explicitly declared there.

madhur:
I get this notification. 

![image](https://user-images.githubusercontent.com/509431/213878847-69418a31-c32c-4a6c-a5a4-9bbc105e3474.png)

Stack trace in `~/.xsession-errors`

```
error while running function!
stack traceback:
        [C]: in function 'assert'
        /usr/local/share/awesome/lib/naughty/layout/box.lua:280: in upvalue 'init'
        /usr/local/share/awesome/lib/naughty/layout/box.lua:308: in function 'naughty.layout.box.set_notification'
        /usr/local/share/awesome/lib/naughty/layout/box.lua:383: in function </usr/local/share/awesome/lib/naughty/layout/box.lua:341>
        (...tail calls...)
        /home/madhur/.config/awesome/rc.lua:190: in local 'func'
[...]

PR #3768, opened 2023-01-22
Bugfix : Invalid position middle

madhur:
Fixes: https://github.com/awesomeWM/awesome/issues/3767


Aire-One:
Thank you for the PR.

While this "hotfix" works for you, I'd prefer to keep it as a short term solution for your use case. I don't think we want to adopt this at the project level.

madhur:
:+1:

PR #3774, opened 2023-02-04
Fix error in docs

techtycho:
Isn't that supposed to be `property::urgent` and not `request::urgent`.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -340,7 +340,7 @@ lua_class_t client_class;
  *        end
  *    end)
  *
- * @signal request::urgent
+ * @signal property::urgent
  * @tparam client c The client whose property changed.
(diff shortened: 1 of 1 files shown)


Aire-One (approved):
Good catch! Thank you

Elv13 (changes requested):
No, I think it really is `request::urgent`.

image


The `request::urgent` signal is used when the client want to become urgent. AwesomeWM does not immediately do it because some users might want to block those things.

The objects properties are documented when you press "show more" in the (git-master) doc. Each object properties are not individually documented because it would make the documentation page twice as large. You can assume that each properties has a signal.

image

techtycho:
I understand now, thanks.

PR #3776, opened 2023-02-11
added gears.surface crop function

paulhersch:
Hi!

I recently shared this function to crop images in the unofficial awesomewm Discord server, and due to the seemingly high interest i wanted to contribute it to the official libs.

Please give me feedback if anything seems wrong/questionable with this.

### DIFF
--- a/lib/wibox/widget/graph.lua
+++ b/lib/wibox/widget/graph.lua
@@ -20,9 +20,12 @@
 local setmetatable = setmetatable
 local ipairs = ipairs
 local math = math
+local math_max = math.max
+local math_min = math.min
 local table = table
-local type = type
 local color = require("gears.color")
+local gdebug = require("gears.debug")
+local gtable = require("gears.table")
 local base = require("wibox.widget.base")
 local beautiful = require("beautiful")
 
@@ -39,8 +42,6 @@ local graph = { mt = {} }
 
 --- Set the graph border color.
 --
--- If the value is nil, no border will be drawn.
---
 --@DOC_wibox_widget_graph_border_color_EXAMPLE@
 --
 -- @property border_color
@@ -50,6 +51,8 @@ local graph = { mt = {} }
 -- @see gears.color
 
 --- Set the graph foreground color.
+--
+-- This color is used, when `group_colors` isn't set.
 --
 --@DOC_wibox_widget_graph_color_EXAMPLE@
 --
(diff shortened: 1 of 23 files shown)


Elv13 (dismissed):

actionless (changes requested):
would be nice to have a basic unit test, verifying resolution before and after crop and mb few side cases

actionless:
[in `./spec/`](https://awesomewm.org/apidoc/documentation/02-contributing.md.html)

paulhersch:
In the last commit i added some nix files, if You think these should not be added to the repo i could revert this

what they offer: dev environment for users of the nix package manager

actionless:
> In the last commit i added some nix files, if You think these should not be added to the repo i could revert this

that's out of scope of this pr

lib/gears/surface.lua
```diff
@@ -283,6 +284,39 @@ function surface.widget_to_surface(widget, width, height, format)
...
 end
 
+--- Crop a surface to a given ratio
+-- This creates a copy of the surface, as surfaces seem to not be
+-- resizable after creation, so you have to use the functions return
+-- value. The resulting surface is in the center of the original surface
+-- @param surf surface the Cairo surface to crop
+-- @param ratio number the ratio the image should be cropped to (widht/height)
+-- @return the a cropped copy of the input surface
+function surface.crop_surface(surf, ratio)
```
sclu1034, line 294: This is a very generic (and thereby "valuable") function name, but the signature and implementation only supports a very small subset of what that feature could be.
My initial assumption for a "crop image" API would have been to allow arbitrary values to be cut on each side, e.g. `crop(img, { left = 15, right = 27, top = 0, bottom = 3})`.

Even if arbitrary cropping is not something that's needed/implemented right now, the signature should at least be changed to allow for that in the future without breaking changes.

sclu1034 (changes requested):

lib/gears/surface.lua
actionless: yup using `args` table instead would be better in long-term indeed

paulhersch: the function i submitted was only meant to be used with a fixed ratio.
I am gonna rename this and work on a general crop function as well, as i am already on it

actionless: i meant that you could just rename `ratio` to `args` and use `args.ratio` in the code, so it would be easier to extend the func without making the API ridiculous in the future

sclu1034: With an args table, it should be the only parameter (because people got used to the fact that functions can be called like `foo { bar = 1, baz = "baz" }`). Then you'd use that to allow for both `crop({ surface = img, ratio = 3/5 })` and `crop({ surface = img, left = 3, right = 3})`.

Separated and renamed to `crop_surface_to_ratio` would be fine, too, though.

> Wouldn't it make more sense out of a users perspective if you assigned a target width and height that the image then gets cropped to?

That would still not cover all use cases. With just a width and height, the anchor would still be fixed (like your current version also fixes it to always crop towards the center).
You could do `width` + `height` + `x` + `y`, but that would usually lead to more calculations required to get those values than the one for each edge (not corner).

paulhersch:
Moved the "old" crop to another function name and added a more general purpose crop function like sclu1034 described, also added tests for both.

lib/gears/surface.lua
```diff
@@ -283,6 +284,79 @@ function surface.widget_to_surface(widget, width, height, format)
...
+
+--- Crop a surface to a given ratio
+-- This creates a copy of the surface, as surfaces seem to not be
+-- resizable after creation, so you have to use the functions return
+-- value. The resulting surface is in the center of the original surface
+-- @param surf surface the Cairo surface to crop
+-- @param ratio number the ratio the image should be cropped to (widht/height)
+-- @return the a cropped copy of the input surface
+-- @staticfct crop_surface_by_ratio
+function surface.crop_surface_by_ratio(surf, ratio)
```
actionless, line 334: now this became weird.

just leave `surface.crop_surface` with `ratio` as an argument

paulhersch: how should something like
```
surface.crop_surface({
    surface = surface.
    ratio = 0.5,
    left = 5
})
```
be handled then? Should the image be cropped before the "ratio + zoomed" operation that the use of ratio implies is performed?

My thoughts:
The surface by ratio crop makes sense from a users perspective if the goal is to crop an image for use as wallpaper for example: You just pass in your screen ratio and you have a wallpaper that fits your screen.
The more manual approach could make more sense for someone that maybe wants to crop an image with a fixed size (buttons(?)) to fit a widget size exactly.

Mixing both crop options could confuse a lot of users and you need to explain too much about how this works in the back in the docs to prevent people explaining about unwanted or random behavior.

actionless: it's more about shared implementation - switching between pixel and ratio modes basically changes only one two line of the whole function

so answering your question, how to handle that particular case - either apply both ratio and pixel crop after computing the ratio or just raise error that args are incompatible

actionless: btw you don't really need to ratio-crop in case of wallpaper - that could be done on the level of wallpaper module itself (ie underlying `imagebox`)

paulhersch:
I have now merged both variants using an args table, allowing to use both offsets and ratio. for the tests i kind of had to "rebuild" a part of the logic to check whether image areas seem to stay the same, which i think is suboptimal because if there is a fault in the original implementation one needs to change the tests as well, but at least changed behaviour is detected that way...

lib/gears/surface.lua
```diff
@@ -283,6 +284,82 @@ function surface.widget_to_surface(widget, width, height, format)
...
+    local offset_w, offset_h =  0, 0
+
+    if (args.top or args.right or args.bottom or args.left) then
+        local left = args.left or 0
+        local right = args.right or 0
+        local top = args.top or 0
+        local bottom = args.bottom or 0
+
+        if (top < 0 or right < 0 or bottom < 0 or left < 0) then
+            gdebug.print_error("negative offsets are not supported for crop_surface")
```
actionless, line 321: i think error() would make more sense here and in most of gdebug.print_error() occurences which i see around

actionless (dismissed):

lib/gears/surface.lua
```diff
@@ -283,6 +284,82 @@ function surface.widget_to_surface(widget, width, height, format)
...
+        h = h - top - bottom
+
+        -- the offset needs to be negative
+        offset_w = - left
+        offset_h = - top
+
+        -- breaking stuff with cairo crashes awesome with no way to restart in place
+        -- so here are checks for user error
+        if w <= 0 or h <= 0 then
+            error("crop_surface target size with top, right, bottom or left too small")
```
sclu1034, line 334: As a user, I would have no idea what this error is supposed to mean.

```suggestion
            error("Area to remove cannot be larger than the image size. Reduce values for `top`, `right`, `bottom` or `left`.")
```

And something that applies to all error messages:
Mentioning the function name `crop_target` isn't really needed, as the stack trace will point to it anyways. If you do want to keep it, though, make it obviously separate from the message, e.g. `"[crop_surface] Area to remove ..."`.
This makes those messages easier to skim.

actionless (approved):

actionless:
https://github.com/awesomeWM/awesome/actions/runs/4984768949/jobs/8923537391?pr=3776#step:33:113

actionless (approved):

PR #3783, opened 2023-03-01
Connect signal and disconnect signal leaks memory

mireq:
Function signal_disconnect don't deallocate signal array after last signal removed. For example this is my rc.lua:

```lua
local a = function()
end

for s in screen do
	for i=0, 10000000 do
		s:connect_signal("property::geometry" , a)
		s:disconnect_signal("property::geometry" , a)
	end
end

awesome.quit();
```
And this is memusage:
mem

This is memusage after patch:
mem

### DIFF
--- a/common/signal.h
+++ b/common/signal.h
@@ -98,8 +98,12 @@ signal_disconnect(signal_array_t *arr, const char *name, const void *ref)
             if(ref == *func)
             {
                 cptr_array_remove(&sigfound->sigfuncs, func);
-                if(sigfound->sigfuncs.len == 0)
+                if(sigfound->sigfuncs.len == 0) {
+                    if(sigfound->sigfuncs.tab) {
+                        p_delete(&sigfound->sigfuncs.tab);
+                    }
                     signal_array_remove(arr, sigfound);
+                }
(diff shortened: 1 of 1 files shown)


Aire-One (commented):
This function documentation says

> * You are in charge of reference counting.

So I guess the initial idea here was that the caller is in charge of de-allocating resources.

I don't know if we have already some places where the memory is correctly freed. With this fix, it would double-free.

Also, the `cptr_array_t` is created with some macro that I think should handle (with `[cptr]_array_delete`) what you are trying to achieve here :thinking:

mireq:
@Aire-One thanks for feedback. I have replaced `p_delete` with `cptr_array_wipe`.

Caller cannot delete this part of memory, because array is not accessible from outside. After `signal_array_remove(arr, sigfound);`, array pointer is not reachable and should be manually deallocated.

Elv13 (approved):
I propose we merge it. If it explodes, some people will complain.

Also, we should really enable address sanitizer in the CI "eventually"

actionless (approved):

Aire-One (approved):

PR #3789, opened 2023-03-25
Modeline parsing never ends before end of file

mireq:
Next line parsing checks if next character is end of buffer, but position i is already incremented with postfix operator. Branch MODELINE_STATE_VALUE never received '\n' character and state MODELINE_STATE_COMPLETE is never used. Next line should check for character on position i instead of i+1, then new line is pssed to MODELINE_STATE_VALUE branch and in this branch, current argument should be pushed to argument list.


Aire-One (approved):
lgtm

sclu1034 (approved):

Elv13 (approved):

PR #3790, opened 2023-03-26
Layoutbox rendering is blurry on HiDPI screen

mireq:
I think, layoutbox should not render icons directly. On HiDPI screens, svg files rendered this way are always in base resoltuion, which looks blurry. Method :set_image() should be called instead.

Few possible modifications:

Silent loading (surface.load_silently) could be dropped and if "layout_" .. name is set, image can be loaded directly. If it's not set, user will see just text representation. If path to image is not valid, i think, it's ok to display error message.

```
local image_name = "layout_" .. name
local theme_image = beautiful[image_name]
local success = false
if theme_image ~= nil then
    success = w.imagebox:set_image(beautiful[image_name])
end
w.textbox.text = success and "" or name
```

Or imagebox could have interface for silent loading.

layotubox_sharp

### DIFF
--- a/lib/awful/widget/layoutbox.lua
+++ b/lib/awful/widget/layoutbox.lua
@@ -15,7 +15,6 @@ local layout = require("awful.layout")
 local tooltip = require("awful.tooltip")
 local beautiful = require("beautiful")
 local wibox = require("wibox")
-local surface = require("gears.surface")
 local gdebug = require("gears.debug")
 local gtable = require("gears.table")
 
--- a/lib/awful/widget/layoutbox.lua
+++ b/lib/awful/widget/layoutbox.lua
@@ -36,7 +36,7 @@ local function update(w, screen)
     if img then
         w.imagebox:set_image(beautiful["layout_" .. name])
     end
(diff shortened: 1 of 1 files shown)


Aire-One (approved):
Sorry for the late replies on your PRs.

I have this idea to move all the `awful.widget.*` widgets to template, so that the user can define the consumer widget with whatever they want.

Regardless, this fix is more than welcome! And I don't think it would clash with my targets, since we need to preserve a default widget that is backward compatible with the current usage.

So it's all good to me. Thank you!

Elv13 (approved):

Elv13:
Sorry for my even longer delay before taking a look at this. It looks good, but to be merged the CI warning needs to be fixed:

```txt

Checking lib/awful/widget/launcher.lua            OK
Checking lib/awful/widget/layoutbox.lua           1 warning

    lib/awful/widget/layoutbox.lua:18:7: unused variable 'surface'
```

Aire-One (approved):

actionless (approved):
thanks!

Aire-One:
@Mergifyio refresh

PR #3791, opened 2023-03-26
Draft: awful.hotkeys_popup rework

real-fht:
This pull request aims to rework `awful.hotkeys_popup`, to be more modern and more customizable with the following main changes:
- Reworked the keys layout, instead of having multiple strings joined to eachother, use a wibox.layout.fixed.vertical with multiple wibox.layout.(ratio/fixed).horizontal to hold the keys and their descriptions. This allows for options like keys_spacing.
- Added many options for styling
- Added widget.group_icons for whoever wants that

TODO:

- [ ] Check for edge cases? (weird width, etc)
- [ ] Sometimes the Next/Previous page buttons look weird, maybe investigate that?
- [x] Documentation for new options

### DIFF
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -676,6 +676,10 @@ function widget.new(args)
             -- for _, w in pairs(keys_layout.children) do
             --   w:adjust_ratio(2, keys_ratio, 1 - keys_ratio, 0)
             -- end
+
+            -- Remove the header if overlaps
+            if ik_add_new_column and overlap_leftovers ~= nil then ret:remove(1) end
+
             current_column.layout:add {
                 widget = wibox.container.background,
                 bg = self.group_bg,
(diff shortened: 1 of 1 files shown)


unai-ndz:
While I like the idea and will try this PR if I find the time the style changes mud the commits for review.

From the [contributing document](https://github.com/awesomeWM/awesome/blob/master/docs/02-contributing.md):

> do not try to fix style of code you are not writing, it's just adding noise for no gain

real-fht:
> I find the time the style changes mud the commits for review.

Could you explain what do you mean by that?

Crylia:
Could you rebase all those reverts? They don't have to show up.

Also it may be smarter to use a `widget_template` and have the current proposal as a default theme, that way we can avoid further PR's trying to improve the widget's look. (May need some more opinions)

There are also some minor naming things that should be addressed like `local matcher = require("gears.matcher")` which should be `gmatcher`.

real-fht:
> Also it may be smarter to use a widget_template and have the current proposal as a default theme, that way we can avoid further PR's trying to improve the widget's look. (May need some more opinions)

I also though of using a widget_template, but sadly have to experience in working these out :(

Maybe I could do that, idk

real-fht:
Closing this for now.

PR #3811, opened 2023-05-17
imagebox: add a "cover" fit policy

Delta-official:
This fit policy emulates the behaviour of CSS' `background-size: cover`

### DIFF
--- a/lib/awful/widget/layoutbox.lua
+++ b/lib/awful/widget/layoutbox.lua
@@ -15,7 +15,6 @@ local layout = require("awful.layout")
 local tooltip = require("awful.tooltip")
 local beautiful = require("beautiful")
 local wibox = require("wibox")
-local surface = require("gears.surface")
 local gdebug = require("gears.debug")
 local gtable = require("gears.table")
 
@@ -32,9 +31,13 @@ local function update(w, screen)
     local name = layout.getname(layout.get(screen))
     w._layoutbox_tooltip:set_text(name or "[no name]")
 
-    local img = surface.load_silently(beautiful["layout_" .. name], false)
-    w.imagebox.image = img
-    w.textbox.text   = img and "" or name
+    local image_name = "layout_" .. name
+    local theme_image = beautiful[image_name]
+    local success = false
+    if theme_image ~= nil then
+        success = w.imagebox:set_image(beautiful[image_name])
+    end
+    w.textbox.text = success and "" or name
 end
 
 local function update_from_tag(t)


Elv13:
Hello,

Can you edit the following 2 files and post a screenshot of the result? This should regenerate the 2 images you see in the doc.

https://github.com/awesomeWM/awesome/blob/master/tests/examples/wibox/widget/imagebox/horizontal_fit_policy.lua
https://github.com/awesomeWM/awesome/blob/master/tests/examples/wibox/widget/imagebox/vertical_fit_policy.lua

Delta-official:
`cover` only really works if you set it on both vertical and horizontal policies

Elv13:
> cover only really works if you set it on both vertical and horizontal policies

In that case, there's probably some interaction with `valign` and `halign`. It might be worth to add new example images to the documentation to reflect that. The easiest way to do that is to copy paste an existing example file and modify it. You need to add the `@DOC_path_to_file_EXAMPLE@` line in the property documentation.

Delta-official:
> > cover only really works if you set it on both vertical and horizontal policies
> 
> In that case, there's probably some interaction with `valign` and `halign`. It might be worth to add new example images to the documentation to reflect that. The easiest way to do that is to copy paste an existing example file and modify it. You need to add the `@DOC_path_to_file_EXAMPLE@` line in the property documentation.

it does appear to. `valign` only works if the image has excess vertically and `halign` works only if there's excess horizontally (basically the same behaviour as if you did `fit` and `none`)

sclu1034:
On a general level, I'll want to throw in that `imagebox` has already become rather complex, and that I'm not sure if we really should keep piling more things on top of that.

For the change itself, the result in the latest screenshot doesn't match what I would expect from the description and name of the option.
When I read "cover", I'd expect the image to be scaled just enough to fill out the given space. But looking at the variant where both properties are set to `"cover"`, the image is clearly cropped both horizontally _and_ vertically. Given that the original has more height than width, I'd expect the image to be scaled such that the width fits perfectly, and that only the top and bottom to be copped.

actionless:
i'll just throw in how it's described in css spec, so everyone would have the same understanding of how it expected to work:

> cover

>    The replaced content is sized to maintain its aspect ratio while filling the element's entire content box. If the object's aspect ratio does not match the aspect ratio of its box, then the object will be clipped to fit.

sclu1034:
If we do go with the CSS spec, then the _smallest rectangle_ is where the above screenshot doesn't comply with the constraint:

> A cover constraint is resolved by setting the [concrete object size](https://drafts.csswg.org/css-images/#concrete-object-size) to the **smallest rectangle** that has the [object’s](https://drafts.csswg.org/css-images/#objects) [natural aspect ratio](https://drafts.csswg.org/css-images/#natural-aspect-ratio) and additionally has neither width nor height smaller than the constraint rectangle’s width and height, respectively.

actionless:
i specifically posted the pics in case if anyone would have troubles with english

sclu1034:
I meant the image from delta that demonstrates the proposed change, not the one from MDN.

actionless:
yup - that's the intention, to make it visible side-by-side

i think the reason is that image is not being centered

actionless:
😹 2023-05-22--1684758990_3160x2160_scrot

alanswanson:
This is the same as #3554 and will probably have the same issue as https://github.com/awesomeWM/awesome/issues/3547#issuecomment-1197635722. Shame that awesomwm is essentially a dead project.

actionless:
@alanswanson your PR is missing doc examples and tests as in this one - mb you could combine effort to make one good PR from two 😺

sclu1034:
@actionless As mentioned here in https://github.com/awesomeWM/awesome/issues/3547#issuecomment-1197287719, they were waiting for the code changes to be reviewed before spending time on docs.
But that's still drifting on a "ping Elv", and will continue to do so, given that Elv doesn't seem to have much time for the project these days. You and @Aire-One might have to step up and get some reviews in without him for the project to keep moving.

PR #3815, opened 2023-05-21
Fix focus.global_bydirection not moving focus from empty screen

JaCoB1123:
When having three monitors and only one client on the first monitor, the function does not work when the middle monitor is focused, as it uses the focused clients' screen instead of the currently focused screen as the base.

### DIFF
--- a/lib/awful/client/focus.lua
+++ b/lib/awful/client/focus.lua
@@ -208,7 +208,7 @@ function focus.global_bydirection(dir, c, stacked)
 
     -- if focus not changed, we must change screen
     if sel == capi.client.focus then
-        screen.focus_bydirection(dir, scr)
+        screen.focus_bydirection(dir, screen.focused())
         if scr ~= get_screen(screen.focused()) then
             local cltbl = client.visible(screen.focused(), stacked)
             local geomtbl = {}


actionless:
https://github.com/awesomeWM/awesome/actions/runs/5039591637/jobs/9038043696?pr=3815

JaCoB1123:
The failing assertion seems to concern `awful.client.focus.bydirection("right")` which I did not modify. Note how that happened? Also seems to only be the case for lua5.1? Any hints, what I should do?

actionless:
with 5.2 it's failing as well: https://github.com/awesomeWM/awesome/actions/runs/5039591637/jobs/9038043753?pr=3815

sclu1034:
> Any hints, what I should do?
The first step would be to run the tests locally. `make check`, IIRC.

JaCoB1123:
Ok, didn't see that. That error is even weirder though, as it's in line 21 even before calling any of the focus functions.

This is my first lua 

> > Any hints, what I should do?
> 
> The first step would be to run the tests locally. `make check`, IIRC.

Thanks for the hint. I unfortunately I have never used lua aside from my awesome config. Therefore I haven't got a local lua dev environment and wasn't able to set one up either because of 'lgi' being missing. I even tried installing lgi by using the steps from the github actions file.

I can't invest more time here, so feel free to close if nobody else is willing to take over.

actionless:
indeed it weirdly failed for both of them at the same time - but restarting helped for both

PR #3819, opened 2023-06-06
fix(awful.key.keygroups): can't use 0 . key

HoNamDuong:
I can't use 0. key to view the 10th tag
```
    awful.key({
        modifiers = { keys.super },
        keygroup = awful.key.keygroup.NUMROW,
        description = "Only view tag",
        group = "tag",
        on_press = function(index)
            local screen = awful.screen.focused()
            local tag = screen.tags[index]
            if tag then
                tag:view_only()
            end
        end,
    }),
```
Because the value assigned with 0. key (`#19`) is 0 and not 10

### DIFF
--- a/lib/awful/key.lua
+++ b/lib/awful/key.lua
@@ -397,7 +397,7 @@ key.keygroups = {
 -- Technically, this isn't very popular, but we have been doing this for 12
 -- years and nobody complained too loudly.
 for i = 1, 10 do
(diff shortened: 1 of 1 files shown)


Aire-One (approved):
This code is pretty old by now, so if it was buggy, I'd hope someone would have reported it earlier. At the same time, I can see why no-one had a need for this binding 🤷

PR #3822, opened 2023-06-14
Avoid loading duplicate .desktop files in menubar

### LINKED ISSUE #1816: Menubar loads desktop files incorrectly

 • Build: Mar  7 2016 18:22:52 for x86_64 by gcc version 6.0.0 (mockbuild@)
 • Compiled against Lua 5.3.2 (running with Lua 5.3)
 • D-Bus support: ✔

 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✘
 • LGI version: 0.9.0

assuming following config:
```lua
menubar.menu_gen.all_menu_dirs = { "/usr/share/applications", "/usr/local/share/applications" }
```

- install application (e.g. google-chrome)
- open menubar  (`super+p`)
- start writing `google` - there is one entry :+1: 
- `$ cp /usr/share/applications/google-chrome.desktop /usr/local/share/applications/`
- restart awesome
- open menubar  (`super+p`)
- start writing `google` - there are two entries :-1: 

> If multiple files have the same desktop file ID, the first one in the $XDG_DATA_DIRS precedence order is used.

(https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#extra-actions-implementation-notes)

dyfrgi:
Currently, if a user copies a .desktop file into some other directory, they will get duplicate entries in menubar. This PR changes that so that only one entry is kept.

The [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html) says that entries have an id derived from their path.

>  To determine the ID of a desktop file, make its full path relative to the $XDG_DATA_DIRS component in which the desktop file is installed, remove the "applications/" prefix, and turn '/' into '-'. 

Only the first file found with a given id, following the order of directories in `$XDG_DATA_DIRS`, should be used. This PR changes the logic used by menubar to do exactly this. This allows users to e.g. override a system defined .desktop file by putting a changed one in `~/.local/share/applications/`. Changing '/' into '-' allows overriding a file stored inside a directory hierarchy of the system without creating directories in .local.

To do this, the GIO Async context was hoisted above the iteration through directories as otherwise following directory precedence would be more complex.

There is one divergence from the spec in this implementation. The spec says:
>  If the desktop file is not installed in an applications subdirectory of one of the $XDG_DATA_DIRS components, it does not have an ID. 

In this implementation, we scan through `menu_bar.all_menu_dirs`, and all these directories are treated the same, regardless of whether they are "an applications subdirectory of one of the $XDG_DATA_DIRS components". Some users set this, though it's pretty uncommon. I think this is unlikely to cause the few people who use this to add more directories any problems with accidentally-ignored .desktop files.

Fixes #1816.

### DIFF
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -406,7 +406,13 @@ function utils.parse_dir(dir_path, callback)
 
     gio.Async.start(do_protected_call)(function()
         local result = {}
-        parser(gio.File.new_for_path(dir_path), result)
+        local f = gio.File.new_for_path(dir_path)
+        parser(f, result)
+        for i, entry in ipairs(result) do
+            local target = gio.File.new_for_path(entry.file)
+            entry.desktop_file_id = f:get_relative_path(target)
+            result[i] = entry
+        end
         call_callback(callback, result)
     end)
 end
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -382,11 +382,11 @@ function utils.parse_dir(dir_path, callback)
     end
 
     local result = {}
-    local f = gio.File.new_for_path(dir_path)
-    parser(f, result)
+    local dir = gio.File.new_for_path(dir_path)
+    parser(dir, result)
     for i, entry in ipairs(result) do
-        local target = gio.File.new_for_path(entry.file)
-        local relative_path = f:get_relative_path(target)
+        local desktop_file = gio.File.new_for_path(entry.file)
+        local relative_path = dir:get_relative_path(desktop_file)
         entry.desktop_file_id = string.gsub(relative_path, "/", "-")
         result[i] = entry
     end
--- a/lib/menubar/utils.lua
+++ b/lib/menubar/utils.lua
@@ -410,7 +410,8 @@ function utils.parse_dir(dir_path, callback)
         parser(f, result)
         for i, entry in ipairs(result) do
             local target = gio.File.new_for_path(entry.file)
-            entry.desktop_file_id = f:get_relative_path(target)
+            local relative_path = f:get_relative_path(target)
+            entry.desktop_file_id = string.gsub(relative_path, "/", "-")
             result[i] = entry
         end
         call_callback(callback, result)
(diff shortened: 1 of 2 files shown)


lib/menubar/utils.lua
```diff
@@ -406,7 +406,13 @@ function utils.parse_dir(dir_path, callback)
 
     gio.Async.start(do_protected_call)(function()
         local result = {}
-        parser(gio.File.new_for_path(dir_path), result)
+        local f = gio.File.new_for_path(dir_path)
+        parser(f, result)
+        for i, entry in ipairs(result) do
+            local target = gio.File.new_for_path(entry.file)
+            entry.desktop_file_id = f:get_relative_path(target)
+            result[i] = entry
```
psychon: Isn't this line redundant? `entry` came from `result`, so is already referenced there.

psychon (commented):
Looks fine to me. I looked at this locally with `git diff -w` and that makes the diff a lot smaller. :-)

lib/menubar/utils.lua
dyfrgi: Good point. My poor Lua knowledge is showing here, about what variables are references and which are not. The Lua reference manual set me straight. I'll remove this.

dyfrgi:
I started working on some unit tests for this, but I gave up due to the lack of support for async() in busted 2.x. https://github.com/dyfrgi/awesome/compare/desktop-file-id...dyfrgi:awesome:menu_gen-unit-test?expand=1 shows the starting point, though. With busted 2.0, we'd need a custom test runner. I don't quite have it in me to dig into lgi + gio async work and busted custom test runners for this PR, so I hope that manual testing is sufficient.

It would probably be possible to refactor things so as to make this easier to test without handling async in the test runner, but it's not obvious to me.

actionless:
there are several tests which have accommodation for that (via re-running test function few times), and using `count` argument as the counter of how many times test function have been attempted to execute, ex:
https://github.com/awesomeWM/awesome/blob/master/tests/test-client-shape.lua#L8

Elv13:
@dyfrgi Do you still plan to work on the test? As @actionless said, this is better done using an integration test suite since it has a full event loop and retry function if they return `nil`.

actionless:
@YuzeHao2023 what are you trying to achieve aside of loosing access to this github repo for spamming? 🤔

Issue #3844, opened 2023-08-23
Fullscreen applications doesn't work


musjj:
<details>
<summary>Output of `awesome --version`:</summary>

```
awesome 4.3 (Too long)
 • Compiled against Lua 5.1.4 (running with LuaJIT 2.1.0-beta3)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: module 'lgi.version' not found:
	no field package.preload['lgi.version']
	no file './share/lua/5.1/lgi/version.lua'
	no file './lgi/version.lua'
	no file './lgi/version/init.lua'
	no file './lib/lua/5.1/lgi/version.so'
	no file './lgi/version.so'
	no file './lib/lua/5.1/loadall.so'
	no file './lib/lua/5.1/lgi.so'
	no file './lgi.so'
	no file './lib/lua/5.1/loadall.so'
 • Transparency enabled: yes
 • Custom search paths: yes
```
</details>

**How to reproduce the issue:**

1. Set your layout mode to `tile`
2. Open a new terminal window, make sure that it's tiled
3. Run `awesome-client <<<"client.focus.fullscreen = true"`

**Actual result:**

Fullscreen doesn't work. If you set the window to floating mode before fullscreening, it will work just fine:

```bash
awesome-client <<<"client.focus.floating = true; client.focus.fullscreen = true"
```

**Expected result:**

Fullscreen should just work.

maverick1872:
I'm also experiencing the same issue with the latest update.

Aire-One:
Hello,

Can you confirm this issue also happens with the default `awesomerc.lua` and the `Super+f` keybinding.

Also, can you try

```
awesome-client 'client.focus and client.focus.fullscreen = true or require("naughty"). notification { text = "no focused client" }'
```

musjj:
Interesting, when the user config (`~/.config/awesome`) doesn't exist, fullscreen works just fine. I'm not sure what could possible be the cause, since my config is still relatively small. Gonna dig through it a bit.

musjj:
Okay, I think my mistake was that I used the wrong default config (copy-pasted from the stable documentation) as my base config. I rebuilt my config based off the `awesomerc.lua` from the latest commit and I no longer have this problem. Not sure what was the exact cause though.

actionless:
[...]

Issue #3834, opened 2023-08-02
Clients with fullscreen hint does not open in fullscreen


akinokonomi:
Output of `awesome --version`:

```sh
awesome v4.3-1606-g0e5fc4575-dirty (Too long)
 • Compiled against Lua 5.1.4 (running with LuaJIT 2.1.0-beta3)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

Also tested with Lua 5.4.6

Default config

**How to reproduce the issue:**

The following command starts client in floating mode, instead of fullscreen:
(I also experience this with some Wine applications)

```sh
> mpv --fullscreen --player-operation-mode=pseudo-gui --pause
```

Though xprop output shows fullscreen state:

```sh
> xprop _NET_WM_STATE
_NET_WM_STATE(ATOM) = _NET_WM_STATE_FULLSCREEN
```

Even with following awesome rule, mpv does not open in fullscreen as long as I start it with `--fullscreen` argument:

``` lua
ruled.client.append_rule {
  id       = "mpv_fullscreen",
  rule_any = {
    class = { "mpv" },
  },
  properties = { fullscreen = true }
}
```

I found similar report https://github.com/awesomeWM/awesome/issues/2173, but in my case application is not stuck anywhere, it also has proper borders, etc., it just starts floating.

I guess there are other, though. Sorry if it's a duplicate.

**Actual result:**

Client starts floating.

**Expected result:**

Client open in fullscreen.

---

As a workaround, following rule fixed the problem for me:

```lua
ruled.client.append_rule {
  id         = "ewmh_fix_fullscreen_state",
  rule       = { },
  callback   = function(c)
    if c.fullscreen then
      c.fullscreen = false
      c.fullscreen = true
    end
  end
}
```

akinokonomi:
Hi!

> Do you have some configurations/rules that could be interfering? Can you try again with the default awesomerc.lua?
> Can you try again with the default awesomerc.lua?

Nope. I've mentioned it, but I've already tested it with the default config (default rules). It's still the same.

I've also tested it as cleanly as possible, with all env variables unset, etc.

> It works as expected, the video opens in fullscreen mode.

Thank you very much for testing it yourself.

I see. Maybe it's due to changes in dependencies?

Here is part from the build log
[...]

PR #3845, opened 2023-08-25
fix(permissions: geometry): allow geometry change for non-floating clients if it was requested by awesome's own fullscreen and maximize handlers (fixes: #3844, #3834)

actionless:


Pikabyte:
I figured I'd add here as well that I can confirm this fixes #3844

actionless:
thanks for checking! @Pikabyte

Aire-One (approved):
🚀

Elv13 (approved):

PR #3846, opened 2023-08-25
fix(awful: layouts: tile: mouse_resize_handler): count the size of useless_gaps around the client (fixes #424)

### LINKED ISSUE #424: Mouse resize in tile layout works weird for the bottom row when `size_hints_honor = false`

actionless:


lib/awful/layout/suit/tile.lua
```diff
@@ -49,7 +49,9 @@ tile.resize_jump_to_corner = true
 local function mouse_resize_handler(c, _, _, _, orientation)
     orientation = orientation or "tile"
     local wa = c.screen.workarea
-    local mwfact = c.screen.selected_tag.master_width_factor
+    local t = c.screen.selected_tag
+    local useless_gap = t.gap
+    local mwfact = t.master_width_factor
```
Aire-One, line 54: I'm not sure whether this is correct. What happens when the user selects multiple tags?

As a possible fix, we could add a `params` parameter like the one for `arrange()`.

lib/awful/layout/suit/tile.lua
```diff
@@ -155,7 +157,7 @@ local function mouse_resize_handler(c, _, _, _, orientation)
                                       return true
                                   end
                               end
-                              return prev_coords.x == coords.x and prev_coords.y == coords.y
+                              return (prev_coords.x == coords.x) and (prev_coords.y == coords.y)
```
Aire-One, line 160: 👍

lib/awful/layout/suit/tile.lua
actionless: that's existing code, see the next line of the diff - i've just assigned it to the variable to avoid writing c.screen..... twice

actionless: answering your question - having two tags (with different params) selected in awesome i think not documented behavior, but i think everywhere it defaults to params of the first (by idx) of selected tags - although this PR is not the best place to discuss that behavior

Aire-One (approved):

Elv13 (approved):

PR #3857, opened 2023-09-26
Release all grabs when unmanaging windows

gldrk:
When a top-level window is unmapped, awesome stops tracking it, possibly leaving behind stale passive grabs.  If the window is later reparented and mapped again, these grabs can become activated, locking the pointer.  awesome does not release grabs on windows it’s not managing, even though it technically could.  The only remedy is destroying the poisoned window.

This was discovered while using [tabbed](https://tools.suckless.org/tabbed/).

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -3010,6 +3010,8 @@ client_unmanage(client_t *c, client_unmanage_t reason)
 
     if(reason != CLIENT_UNMANAGE_DESTROYED)
     {
+        xwindow_buttons_grab(c->window, &(button_array_t){ .len = 0 });
+        xwindow_grabkeys(c->window, &(key_array_t){ .len = 0 });
         area_t geometry = client_get_undecorated_geometry(c);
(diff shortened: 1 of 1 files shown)


objects/client.c
```diff
@@ -3010,6 +3010,8 @@ client_unmanage(client_t *c, client_unmanage_t reason)
 
     if(reason != CLIENT_UNMANAGE_DESTROYED)
     {
+        xwindow_buttons_grab(c->window, &(button_array_t){ .len = 0 });
+        xwindow_grabkeys(c->window, &(key_array_t){ .len = 0 });
```
psychon, line 3014: Random drive by comment:
Both these functions do "ungrab everything, then re-grab all the requested stuff". Due to the empty argument, the re-grabbing doesn't do anything. Dunno if it would be simpler to just copy the "ungrab everything" request here.

But whatever. You still get my :+1: .

psychon (approved):

Elv13 (approved):

Issue #3136, opened 2020-07-28
Electron frameless windows, using custom titlebars can not be dragged


motorlatitude:
Output of `awesome --version`:

```
awesome v4.3-858-g49a3859c8 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
 • Transparency enabled: yes
 • Custom search paths: no
```

**Issue:**

Certain electron apps will use custom title bars over native ones, or the custom title bars fit more with the applications theming. Custom title bars use the `-webkit-app-region: drag;` CSS attribute to create [draggable regions](https://www.electronjs.org/docs/api/frameless-window#draggable-region) within the window. It is not currently possible to drag a window by using these draggable regions. They don't get recognised and as such these windows either have to use native title bars or resort to using the <kbd>super</kbd> + mouse drag combination.

**Actual result:**

The window can only be dragged using the <kbd>super</kbd> + mouse drag combination, not by the title bar, which is not ideal in floating mode. I personally like the custom title bar offered by VSCode but cannot move the window through it.

**Expected result:**

To drag a window by a native or custom title bar.

P.S. I'm not sure whether this is an issue with Awesome or Electron in this case.

actionless:
> P.S. I'm not sure whether this is an issue with Awesome or Electron in this case.

please try with Openbox, from my experience it's less buggy (but still modern) WM in the world, so i always using it for double-checking any window-management-related quirks with the apps

motorlatitude:
So I quickly spun up an OpenBox VM and I used VSCode as a test and it was possible to drag the window using the custom title bar. Also quickly tried on my own electron app which also uses custom draggable regions and it also worked flawlessly on OpenBox.

actionless:
> tried on my own electron app

if you can upload here the minimal reproductive example that would be useful

psychon:
Random guess: Missing `_NET_WM_MOVERESIZE` support: https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#idm45805408003200
And I have no idea how to "send" such a request through Lua.
(Also, the spec mentioning that it is broken and that there is a possible race condition inspires lots of trust...)

Elv13:
> And I have no idea how to "send" such a request through Lua.

[...]

Issue #3849, opened 2023-08-30
Flycky CI for test-focus-bydirection.lua


Elv13:
Seen in #3815 

```txt
ests/test-drawable-bgimage.lua ==
== Running /home/runner/work/awesome/awesome/tests/test-focus-bydirection.lua ==
Error: running function for step 1/2 (@2): ...er/work/awesome/awesome/tests/test-focus-bydirection.lua:26: assertion failed!
stack traceback:
	[C]: in function 'assert'
	...er/work/awesome/awesome/tests/test-focus-bydirection.lua:26: in function <...er/work/awesome/awesome/tests/test-focus-bydirection.lua:13>
	[C]: in function 'xpcall'
	/home/runner/work/awesome/awesome/tests/_runner.lua:110: in function </home/runner/work/awesome/awesome/tests/_runner.lua:101>
	[C]: in function 'xpcall'
	lib/gears/protected_call.lua:41: in function 'protected_call'
	lib/gears/timer.lua:280: in function <lib/gears/timer.lua:278>!
NOTE: there were 2 clients left after the test.
===> ERROR running /home/runner/work/awesome/awesome/tests/test-focus-bydirection.lua <===
Error: running function for step 1/2 (@2): ...er/work/awesome/awesome/tests/test-focus-bydirection.lua:26: assertion failed!
== Running /home/runner/work/awesome/awesome/tests/test-focus.lua ==
Error: failed to wait for X clients from previous test to close!
Last reply: .
Log:
Error: running function for step 1/2 (@2): ...er/work/awesome/awesome/tests/test-focus-bydirection.lua:26: assertion failed!
stack traceback:
	[C]: in function 'assert'
	...er/work/awesome/awesome/tests/test-focus-bydirection.lua:26: in function <...er/work/awesome/awesome/tests/test-focus-bydirection.lua:13>
	[C]: in function 'xpcall'
	/home/runner/work/awesome/awesome/tests/_runner.lua:110: in function </home/runner/work/awesome/awesome/tests/_runner.lua:101>
	[C]: in function 'xpcall'
	lib/gears/protected_call.lua:41: in function 'protected_call'
	lib/gears/timer.lua:280: in function <lib/gears/timer.lua:278>!
NOTE: there were 2 clients left after the test.
```

This need to be investigated. The old travis CI had a `awful.placement` flacky test which I suspect are caused by the same race condition.

[...]

PR #3859, opened 2023-09-30
Implement support for EWMH _NET_WM_MOVERESIZE client messages

### LINKED ISSUE #2140: Support _NET_WM_MOVERESIZE
**Output of `awesome --version`:**

 • D-Bus support: ✔
 • execinfo support: ✘
 • xcb-randr version: 1.5
 • LGI version: 0.9.2

gtk+ relies on this to detect is the WM supports move and resize. Otherwise gtk+ will use a clumsy and sometimes buggy implementation of move and resize.

phantamanta44:
This PR adds support for the EWMH client message `_NET_WM_MOVERESIZE`, which allows clients to delegate mouse-based movement and resizing to the window manager. It does this by:

1. adding a handler for the `_NET_WM_MOVERESIZE` message that emits the `request::mouse_move`, `request::mouse_resize`, and `request::mouse_cancel` signals when a client makes a corresponding request.
2. adding default handlers for those signals to `awful.permissions` that carry out the mouse movement/resizing operations using `awful.mouse.resize`.
3. adding extra configuration options to the mouse resizing and placement library functions so that they behave the way clients would expect them to (default behaviour is unchanged).

Some implementation notes:

* The "keyboard movement" and "keyboard resizing" operations defined in the `_NET_WM_MOVERESIZE` spec are ignored. It's unclear what they're supposed to mean and they are similarly ignored by i3 and bspwm.
* The `_NET_WM_MOVERESIZE` spec mentions that a "race condition" is possible in the case where the user has already released the mouse button before the mouse movement message is processed. This is already dealt with by way of `awful.mouse.resize` cancelling the operation if the mouse button is no longer pressed.
* It's a little ugly that the `request::mouse_move` and `request::mouse_resize` signals and the `awful.mouse.client` library functions `move` and `resize` apparently do the same thing independently of each other. I could not find a way to delegate one to the other without requiring breaking API or behavioural changes, and so I elected, for the time being, to just leave them independent.

I've tested this on my amd64 NixOS system on the following programs:

* Firefox — fully functional.
* Thunderbird — fully functional.
* Steam — fully functional.
* Visual Studio Code — functional when windowed, but un-maximizing a maximized window seems to put it in a broken state. This also seems to happen on awesome 4.3 without this PR, so it's probably unrelated...?

This should close #2140 and, by extension, #3136.


actionless:
it's adding some new api and changing several existing code branches - so would be nice to add some test

lib/awful/permissions/init.lua
```diff
@@ -567,6 +568,62 @@ function permissions.client_geometry_requests(c, context, hints)
     end
 end
 
+function permissions.client_mouse_move(c, context, args)
```
actionless, line 571: this and onward functions added to this module are missing documentation

actionless:
> I assume title bars were originally added to all clients because of lack of _NET_WM_MOVERESIZE support; therefore, this PR should obsolete that rule

i think it was no specific reason for doing that, it just wasn't implemented that specific case, but thanks for looking into it 👍

tests/_client.lua
```diff
@@ -24,16 +24,21 @@ local function open_window(class, title, options)
         default_height = options.default_height or 100,
         title          = title
     }
+    if options.resize_increment or options.custom_titlebar then
+        -- These require Gtk3, but may fail with an obscure message with Gtk2.
+        -- Produce a better error message instead.
+        assert(tonumber(require("lgi").Gtk._version) >= 3, "Gtk 3 required, but not found")
```
actionless, line 30: there is a proper fix for gtk 3.0 / 4.0 conflict, however i don't have time right now to finish it, but i posted 2 patches inside the issue: https://github.com/awesomeWM/awesome/issues/3849

actionless:
also you need to do this:

```console
$ rg 'unpack.*='
tests/test-awful-rules.lua
5:local unpack = unpack or table.unpack -- luacheck: globals unpack (compatibility with Lua 5.1)
```

i think we should eventually move that to gears.table but it is what it is now 🥲

tests/_client.lua
actionless: no worries regarding gtk 2 - there are other tests which already depend on gtk3

actionless:
yup, exactly 👍

actionless (dismissed):

lib/awful/permissions/init.lua
```diff
@@ -567,6 +568,101 @@ function permissions.client_geometry_requests(c, context, hints)
...
+        or c.type == "desktop"
+        or c.type == "splash"
+        or c.type == "dock" then
+        return
+    end
+
+    local corner_pos = aplace[args.corner](mouse, {parent = c, pretend = true})
+    mresize(c, "mouse.resize", {
+        corner = args.corner,
+        corner_lock = true,
```
Aire-One, line 638: These two new parameters aren't documented. I think they should be part of the Placement “Common arguments” list.

Also, I think only `corner` should be required. `corner_lock` seems unnecessary considering the test in `placement.resize_to_mouse` is `if args.corner and args.corner_lock then`.

Aire-One (commented):
Thank you for this work!

lib/awful/permissions/init.lua
Aire-One: Yes, you're right. It makes more sense to have this specific parameter in the function documentation rather than the "common parameters" section.

Aire-One (dismissed):
:rocket:

Elv13:
(finally...) have time to look at this. First of all, wow, this is very well implemented.

Is this ready to be merged?

phantamanta44:
Rebased onto `master`. I've also fixed an issue I was having with Steam duplicating a mouse movement request and thus causing `mousegrabber` to raise an "already running" error. In particular, I believe `xcb_grab_pointer` ignores any such duplicate requests if the pointer is already grabbed, a behaviour which I've replicated by guarding against `mousegrabber.isrunning()` in the `request::mouse_*` handlers. I am still running this patch on my own machine with no problems, aside from this duplicate request thing.

miranda:
What is the status of this?  I just ran into this with the Battle.net launcher via Wine.  I had the normal Awesome title bars enabled and the window was behaving crazy after resizing sometimes.  It seems like a conflict between the app's custom window controls and Awesome's titlebar.  I disabled titlebars with a window rule and discovered you can't really use the other bar, it jumps all over the place.  Is this going to be completed/merged any time soon?

actionless:
@miranda ⬇️
2026-03-19--1773878442_1143x234_scrot

miranda:
> @miranda ⬇️ 

I saw that but I didn't understand what it was saying.  But I think I understand it now.

Anyway, I pulled this PR and tried it myself.  So far it seems to work perfect for Blizzard's Battle.net launcher, but I'll keep testing it.  Hopefully it gets merged despite the robot gatekeepers.

actionless:
>  I think I understand it now.

probably not:

> despite the robot gatekeepers.


________

unless you going to contribute missing tests for this feature, i don't think this conversation is going anywhere

phantamanta44:
I somehow didn't see that codecov wasn't passing; this should do it

actionless (approved):
thanks again! 😸

patrick99e99:
@Aire-One, @Elv13, @sclu1034, any chance one of you could review this and get it merged in?  It would be fantastic to be able to use this.

PR #3863, opened 2023-10-09
Send a default action when notification is closed by user

### LINKED ISSUE #3182: Clicking notification does not urgent/raise client

BarbUk:
Really fix #3182

### DIFF
--- a/lib/naughty/dbus.lua
+++ b/lib/naughty/dbus.lua
@@ -69,6 +69,9 @@ local function sendNotificationClosed(notificationId, reason)
     if reason <= 0 then
         reason = cst.notification_closed_reason.undefined
     end
+    if reason == cst.notification_closed_reason.dismissed_by_user then
+        sendActionInvoked(notificationId, "default")
+    end
     if bus_connection then
         bus_connection:emit_signal(nil, "/org/freedesktop/Notifications",
         "org.freedesktop.Notifications", "NotificationClosed",


Elv13 (approved):

seqizz:
Thanks!
This looks like properly switching to the target channel, at least on on Slack & Telegram.

BarbUk:
> Thanks!
> 
> With this, looks like I can properly switching to the target channel, at least on on Slack & Telegram.

I tested different notifications and applications from my side:

 - chrome / gmail (with two chrome instances running): it open the correct tab/mail
 - slack: open the correct channel
 - mattermost: open the correct channel

So it seems to work.

> Yet I am not sure if we should include the feature (which is given [here](https://github.com/awesomeWM/awesome/issues/3182#issuecomment-1753211773)) directly into WM or leave it out.

It needs more work to handle edge case I think.
I could include it in another PR under a config flag like `naughty.config.follow_clicked_notification`.

In any case, it should be documented somewhere if people like to jump to the right client when clicking a notification.

actionless (approved):
thanks 👌😺

BarbUk:
FYI this PR fix https://github.com/awesomeWM/awesome/issues/3182 but it's not a perfect solution as it send a default action for all notification dismissed by a user.

So if I have a notification with two actions (default, action 1, action 2), and I click on the second action, it will send:
 - default
 - action 2
 
I'll make another PR with a proper fix.

PR #3879, opened 2023-12-30
Fix the documentation warnings

Elv13:
They were added because the PRs which added those APIs predated the warnings, but were merged later.

And also fix some rendering bugs caused by the shims not behaving correctly.

### DIFF
--- a/docs/config.ld
+++ b/docs/config.ld
@@ -1739,6 +1739,13 @@ local summarize = {
     readonly      = {index = 8, title = "read only"       , count = false},
 }
 
+local no_prefix = {
+    property           = true,
+    signal             = true,
+    clientruleproperty = true,
+    deprecatedproperty = true,
+}
+
 local delimiter_for_tag = {
     usebeautiful  = { "table class='widget_list' border=1", "table", "tr", "tr", {"Theme variable", "Usage"}},
     propbeautiful = { "table class='widget_list' border=1", "table", "tr", "tr", {"Theme variable", "Usage"}},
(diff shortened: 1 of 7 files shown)


Elv13:
I will merge this soon because it's a doc only change and getting rid of those warning makes foxing other bugs easier

actionless:
> They were added because the PRs which added those APIs predated the warnings, but were merged later.

mb then CI should fail on such warnings, to prevent it in the future?

sclu1034:
> mb then CI should fail on such warnings, to prevent it in the future?

That would still not prevent such a PR from being merged, since CI runs on commit push, not on merge. Also known as "merge skew" or "semantic merge conflicts".
We'd also have to use something like GitHub's Merge Queues or [bors](https://bors.tech).

actionless:
> runs on commit push, not on merge.

we had it before, when using Travis - but didn't moved to the current setup, idk if that was a purpose choice or just an overlook during the migration

sclu1034:
What Travis did doesn't help here. Travis does merge the newly pushed commits into the base branch, but the crucial part is, it does it _at the time of the push_. That still doesn't cover the issue here of

1. Push to PR 1
2. CI builds PR 1
3. Push to PR 2
4. CI builds PR 2
5. Merge PR 2
6. Merge PR 1 <-- here, the result from `2.` is outdated, but there is no push to PR 1, so CI doesn't trigger again

Merge queues trigger at that `6.`. _At the time of the merge_, they first merge head and base branch into a staging area, run CI on that, and then merge into the actual target branch. And they do those in sequence, even if two PRs are triggered to merge at the same time.

PR #3880, opened 2023-12-30
Fix a drawin crash, a drawin memory leak and a wallpaper error.

Elv13:
I doubt anybody ever noticed this, but it existed.


actionless (approved):
i think it was actually some reports about weird un- investigated/confirmed leaks

and myself sometimes notice some leakage on awesome proc but i'm not sure how to reproduce that - so neither can tell for sure if they're gone with this PR

but seeing the test gives a good feeling about this, even if it's not addressing aforementioned leaks

sclu1034:
> and myself sometimes notice some leakage on awesome proc 

Confirmed leakage, as in something like Valgrind confirmed that the address was dropped without a `free()`? Because merely observing a slow increase in memory usage over time via system monitoring is not a confirmation of a leak.

More likely, it would be the same issue as everything else reported as an alleged leak, including the 3-4 issues that are still open:

Lua doesn't have the facilities to notify it's GC of how big a memory region in a pointer is. I.e. if you have a userdata that contains a pointer to a big memory chunk (such as a Cairo surface pointing to image data), Lua only knows about the size of the userdata struct, which would be 64 bit for the pointer. Lua doesn't know, and cannot be told about, the kilo- or megabytes of data that the pointer actually contains, because they were not allocated by Lua.
You'd have to do all memory allocation through `lua_newuserdata`, but that's not compatible with the internal memory management of GLib and other libraries.

Example:
With the default settings for the garbage collector, the allocated memory (as known to the GC) needs to double for the next cycle to start.
If the GC thinks that the current usage after the last cycle is 50 MB, then it would take ~800000 of those 64 bit userdata values to double the usage and trigger the next cycle. But if those values each point to 512 KB of memory, you'd also have about 400 MB that the GC isn't aware of, bringing the total to 500 MB.

This is easily verifiable by adding a `GIdleSource`, or a short timer, that forces full GC cycles (`collectgarbage("collect")`) constantly. If your memory usage stops growing with that, the issue is as described, and not a leak (leaked memory, by definition, would not be reachable by the GC).

actionless:
@sclu1034 if you wanna discuss each of them - i guess better place would be there (https://github.com/awesomeWM/awesome/issues?q=is%3Aissue+is%3Aopen+memory+leak+)

actionless:
> You'd have to do all memory allocation through lua_newuserdata, but that's not compatible with the internal memory management of GLib and other libraries.

idk if that's implemented for imageboxes - if it's not it could be one of sources of such unfreed memory chunks

PR #3883, opened 2023-12-31
Fix notification image reloading

Elv13:
It might work. I ran the code, but I don't have "real" tests for this right now.

### DIFF
--- a/lib/wibox/widget/imagebox.lua
+++ b/lib/wibox/widget/imagebox.lua
@@ -26,6 +26,8 @@
 
 local lgi = require("lgi")
 local cairo = lgi.cairo
+local gio = require("lgi").Gio
+local glib = lgi.GLib
 
 local base = require("wibox.widget.base")
 local surface = require("gears.surface")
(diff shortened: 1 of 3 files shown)


lib/wibox/widget/imagebox.lua
```diff
@@ -307,16 +341,27 @@ end
 -- @method set_image
 -- @hidden
 -- @tparam image image The image to render.
+-- @tparam[opt=nil] table|nil args
+-- @tparam[opt=false] boolean args.force Reload the image even if the path has
+--  not changed.
```
sclu1034, line 346: I don't think it makes sense to implement this `force` parameter for all values of `image`. For anything but a path, there will be no change, and the resulting signals would be misleading ("emitted when the image changes").
Given that this is about being able to load from a path uncached, just make it explicit:

```suggestion
-- @tparam[opt=false] boolean args.uncached Bypass the cache when loading the image.
```

Either way, you're currently still not bypassing `rsvg_handle_cache`.

sclu1034 (changes requested):

PR #3884, opened 2023-12-31
Slider dragging signals

Elv13:
Copy of https://github.com/awesomeWM/awesome/pull/3533

Changed the `:is_dragging()` method to be a property and fix `luacheck` warnings.

@sclu1034 Are you fine with this? This address a comment I left (and then forgot about) in the other PR.

### DIFF
--- a/lib/wibox/widget/slider.lua
+++ b/lib/wibox/widget/slider.lua
@@ -274,6 +274,28 @@ local slider = {mt={}}
 -- @beautiful beautiful.slider_bar_active_color
 -- @param color
 
+--- Emitted when the user starts dragging the handle.
+--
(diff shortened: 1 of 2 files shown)


sclu1034:
I mean, you already went as far as creating an entire copy of the PR to implement your own version before I even got to answer that question, which appears to me like my answer won't really affect much.
But for posterity, no I'm not. Making this a "property" doesn't make sense to me.

Users expect to be able to set a value for a property (except for the client properties where X11 dictates otherwise) and to get a corresponding signal on change (which still applies for those X11 properties).
But neither applies here. Making this a property just turns the direct function call into a function call hidden by metatables and the user not typing `()`.

PR #3910, opened 2024-04-13
Default to only use ARGB visual for ARGB client windows

### LINKED ISSUE #2408: Use ARGB parents only for windows with ARGB

```
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.2
```

awesome by default create ARGB parent windows for all windows, unless `-a` is specified on the command line, then ARGB is not used at all.

Probably better to use ARGB windows as needed.

yshui:
Don't use the same visual for all windows.

And add a `--full-argb` option to force ARGB visual for all windows, which is the current default behavior.

Fixes #2408

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -693,21 +693,23 @@ main(int argc, char **argv)
     if(xcb_connection_has_error(globalconf.connection))
         fatal("cannot open display (error %d)", xcb_connection_has_error(globalconf.connection));
 
+    if (!(default_init_flags & INIT_FLAG_ARGB))
+        globalconf.argb_mode = ARGB_MODE_DISABLED;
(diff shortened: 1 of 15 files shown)


yshui:
Maybe it's better to instead keep the current default, and enable the new behavior with an option, like `--smart-argb` or something :thinking:

Elv13:
Sorry for the delay, I don't have a lot of time for Awesome right now. Won't think break people who add transparent titlebars or use anti-aliased round corners?

yshui:
@Elv13 yes, this will. alternatively we can also keep the default and add an option to enable this behavior, maybe `--argb-if-needed`.

actionless:
yup, i also think it's better to preserve the current behavior as a default

PR #3914, opened 2024-04-26
Add Nix flake (draft)

Octelly:
I believe Nix is a helpful tool for both development as well as a (at least for some) comfortable package manager. Having an official Nix flake in the repo is nowadays pretty common (e.g. the [Neovim](https://github.com/neovim/neovim) and [Prism Launcher](https://github.com/PrismLauncher/PrismLauncher/) repos do this) and can be beneficial to Nix users and developers alike.

The current revision builds just fine on my Arch Linux install, however it is by no means finished and I am looking for feedback. The flake was heavily inspired by [Neovim's flake](https://github.com/neovim/neovim/blob/master/contrib/flake.nix).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1710146030,
+        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1714076141,
+        "narHash": "sha256-Drmja/f5MRHZCskS6mvzFqxEaZMeciScCTFxWVLqWEY=",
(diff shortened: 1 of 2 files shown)


actionless:
i'd suggest you having it in a separate repo - because following your logic that "some other popular apps have it" - we must also add PKGBUILD for Arch and etc

sclu1034:
> Having an official Nix flake in the repo is nowadays pretty common 

**by projects that either have maintainers using this themselves, or enough contributors that do**.

Keep in mind that it's very easy to drop a _currently_ working version with a PR. But from the maintainers' perspective, that's something that needs to be maintained _for years to come_, but the initial authors rarely stick around to keep their stuff up-to-date themselves.

As far as I'm aware, none of the active maintainers use Nix(OS) enough that something "official" would be feasible.

Octelly:
@actionless Nix is a general build tool, not a packaging system for a specific system. Of course I wouldn't want a PKGBUILD to be added.

@sclu1034 You're probably right. I'd love Awesome to have a Nix flake, but I haven't even used Awesome myself in quite a while. I originally made this for @Aproxia-dev (I believe she previously contributed to this repo), who complained about building Awesome under NixOS being a hassle, so I quickly coded this (since I have more Nix experience), hoping I could help some people out.

javacafe01:
@Octelly @Aproxia-dev The work is already done for you. I've contributed to a friend's personal Nix package repo, which has a derivation for AwesomeWM with a binary cache. It's easy to use with flakes.


https://github.com/moni-dz/nixpkgs-f2k

Octelly:
@javacafe01 hey Java, nice seeing you here (not as active in the AWM community as Emi)

we already have that one in our inputs actually, I just thought that something more official so to speak would be cool, but you're right

PR #3917, opened 2024-05-07
Adding zoom-fill

Stardust-kyun:
Added a zoom-fill example to the awful.wallpaper docs and replaced awful.wallpaper default to use zoom-fill in awesomerc.

### DIFF
--- a/awesomerc.lua
+++ b/awesomerc.lua
@@ -105,16 +105,11 @@ screen.connect_signal("request::wallpaper", function(s)
     awful.wallpaper {
         screen = s,
         widget = {
-            {
-                image     = beautiful.wallpaper,
-                upscale   = true,
-                downscale = true,
-                widget    = wibox.widget.imagebox,
+            image  = gears.surface.crop_surface {
+                surface = gears.surface.load_uncached(beautiful.wallpaper),
+                ratio   = s.geometry.width/s.geometry.height,
             },
-            valign = "center",
-            halign = "center",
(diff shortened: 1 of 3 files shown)


Stardust-kyun:
I've also noticed that all of the maximized examples are written as "mazimized", with a z instead of an x -- is this intentional?

actionless:
> as "mazimized",

yeah clearly a typo

> (changes in usage of wallpaper module in rc.lua)

i'll leave up the decision to @Elv13 who as far as i remember was designing the wallpaper module - i personally think that result should be achieved by just using (or mb adding new) arguments, not manually pre-cropping the wallpaper straight ni the rc.lua

Elv13:
Hi,

Thanks for the extra doc example. Can you have just the example part as part of this PR? I am not really convinced why we would change the default. With this code, it relies on the current screen geometry (which might, sometime, change without this code being re-executed) and uses a rather low level helper function (which will confuse new users). I think those 2 things have side effects which might cause problem in some cases.

Stardust-kyun:
Thanks for the reply Elv! Reverted the change to the default. I figured that it would help new users to have it as the default since using zoom-fill for wallpapers is most common, and the default code can be misunderstood as such. However, I agree with your comments about its reliance on screen geometry and gears.surface. Pull should now be ready to go.

actionless (approved):

Elv13 (approved):

Issue #3865, opened 2023-10-20
Add some way for awesome.spawn to set G_SPAWN stdio flags.


aarondill:
The specific flags in question are:

```
G_SPAWN_STD{ERR,OUT,IN}_TO_DEV_NULL
G_SPAWN_CHILD_INHERITS_STD{ERR,OUT,IN}
```

Attempting to spawn a process (using `awesome.spawn` directly _or_ indirectly ((awful.spawn.\*))) with it's stdio pointing to `/dev/null` is currently impossible (without manually spawning using lgi bindings to g_spawn_async_with_pipes, but this means snid detection can't be used).

Use case:

My personal use case is to spawn certain nosey processes without outputting their results to the console.
In my configuration, awesome's stdout and stderr are redirected to a log file so I can check for errors later, but this file also includes all processes spawned by awesome.spawn.
I tried using the `return_std{in,out,err}` parameters to avoid this issue, but ended up running out of FD and being unable to spawn processes. (super frustrating to debug...)
Using the `return_std{in,out,err}` parameters, you could return parameters and close them, but most applications will die on `SIGPIPE` (from closing stdout/stderr) so this is not feasible. Additionally, if `return_stdin` is false, stdin is connected to /dev/null anyways, so this becomes entirely pointless.

**Actual result:**

The options given for awesome.spawn are booleans that determine whether file descriptors should be returned.
If `false` or `nil` is passed for stdout/stderr then the program will inherit the corresponding stream from awesome, and if `true` is passed, a new fd is created and returned for use in the program (presumably using LGI).
This makes it impossible to simply discard the output (redirect to `/dev/null`)

note: stdin will be redirected from /dev/null by default (see: https://docs.gtk.org/glib/flags.SpawnFlags.html#child_inherits_stdin).
This instead makes it impossible to inherit stdin from awesome (unlikely situation, but perhaps a program needs console access?)

**Expected result:**

Ideally, there would be a way to indicate to `awesome.spawn` that stdout/stderr should not be returned, but rather redirected to `/dev/null`.

Possible Implementations:

- Another boolean :(
  - Pros: backwards compatability
  - Cons: _Another_ boolean. Given that awesome.spawn takes _7_ params right now, this seems bad.
- Expand the `return_std{in,out,err}` params to allow non-boolean parameters (See below for examples).
  - Pros: Potential expansion?
  - Cons: Multiple types must now be checked and supported.
- Finally, we could drop the _7_ parameters and replace it with a table?
[...]

PR #3932, opened 2024-06-30
feat: allow passing G_SPAWN stdio flags to awesome.spawn

aarondill:
Fixes: #3865
Currently works by allowing the exact strings "DEV_NULL" or "INHERIT" to be passed to return_std*.

Documentation needs to be fixed. I don't know how to use the documentation to show that a parameter can be multiple types.


aarondill:
I have no idea why the checks are failing to compile. This compiles (and runs successfully) locally. All the "missing" symbols should be provided by `<glib.h>`.

aarondill:
It turns out the GLIB Version is too low.
[The source](https://github.com/GNOME/glib/blob/main/glib/gspawn.h#L191-L216) specifies that it's only available from version >=2.74; but this is not shown in the [docs](https://docs.gtk.org/glib/flags.SpawnFlags.html#stdin_from_dev_null).
The `gir1.2-gtk-3.0` Ubuntu package depends on `GLib >= 1.39.0` (currently `1.64.0-2`), so this isn't available on Ubuntu, while on Arch, I'm running `Glib2 2.80.3-2`.

actionless:
prolly the only way for use here is to IFDEF it

and mention that in the docs (also you didn't updated the doc aside of changing the type)

aarondill:
@actionless I added a (compile-time) version check. I tried to use ifdef, but for some reason it wouldn't detect as defined.

aarondill:
The new behavior has been added to the documentation

spawn.c
```diff
@@ -412,9 +412,15 @@ spawn_child_exited(pid_t pid, int status)
  *
  * @tparam string|table cmd The command to launch.
  * @tparam[opt=true] boolean use_sn Use startup-notification?
- * @tparam[opt=false] boolean stdin Return a fd for stdin?
- * @tparam[opt=false] boolean stdout Return a fd for stdout?
- * @tparam[opt=false] boolean stderr Return a fd for stderr?
+ * @tparam[opt=false] boolean|string stdin Pass `true` to return a fd for stdin.
+ *   Use `"DEV_NULL"` to redirect to /dev/null, or `"INHERIT"` to inherit the
+ *   parent's stdin.
```
actionless, line 417: could you also mention smth, that prior to glib version X.XX only boolean value is allowed?

aarondill: @actionless this isn't true though (I especially wanted to avoid that behavior,since it's hard to check from lua).
before glib 4.74, the branches to explicitly set the defaults are compiled out, however, _no error will occur_
This should be an implementation detail which is invisible from the outside.

actionless: that's the point of documenting it - to not confuse the user why value is set but nothing happening

aarondill: I'll put it in the docs if you'd like me to, however, what i'm trying to say is that on _any_ version of glib, the behavior is the same, since these flags just _explicitly_ set the glib default behavior (ie, it's the same as passing `false`)

actionless: so, would be good to mention it all here in the docstring 👍

aarondill: @actionless I've added this to the docstring.
Here's the updated (rendered) docs:

actionless (approved):
thanks 👍

aarondill:
Is there anything left for me to do for this to be merged, or are we just waiting for another review?

actionless:
see https://github.com/awesomeWM/awesome/issues/3934

Aire-One:
There is also the `Build & Test / fixed-lgi-lua5.2` check that is unsuccessful. So the automatic merge wouldn't work anyway here.

actionless:
yes

aarondill:
Alright, I've all the commits into the first (primary) commit.

aarondill:
From what I can tell, the CI checks failed during setup (and thusly not due to my code)

Elv13:
Apparently this is documented incorrectly, but the second argument of `awful.spawn` can be a table or "arguments". Please use that table for the new features. Adding more and more argument to the function doesn't scale. You end up with `awful.spawn("cmd", false, nil, nil, nil, nil "something")`, which is hard to read.

Also, the oldest still supported glib2 version appears to be 2.56 (for RHEL 8.8). Ubuntu 20.04 and 22.04 also have older versions. So 2.74 definitively will prevent people from building on a large number of totally supported distributions. This will require some `ifdef`, otherwise this can't be merged. You can do them in CMake by adding a new `-D`, but using normal `#ifdef` provided by glib2 is preferred. Because it breaks the most used LTS distributions, raising the glib2 version this far cannot be done for many more years. The most used Linux distro in the world right now, while the number are hard to get, is most probably Ubuntu 22.04.

aarondill:
I currently haven't touched `awful.spawn`, are you saying you'd like me to add this functionality there (as well)?

I currently _do_ have the >2.74 functionality behind an `#if GLIB_CHECK_VERSION(2,74,0)`, which is removed in versions <2.74. According to the GLib docs, all the other flags are available since 2.0.

This absolutely _should_ build in its current state, without raising the minimum required GLib version. If i'm mistaken and that's not the case, please let me know so I can fix it.

aarondill:
The CI builds and tests pass on GLib version 2.64.6; so i'm fairly certain that this change is backwards-compatible.

```
--   Found glib-2.0, version 2.64.6
```

aarondill:
> see #3934

@actionless now that #3934 is fixed, are there any blockers to this pr?

actionless (approved):

aarondill:
I've rebased this onto the current `master`.
@Aire-One, Could you please review this change?

Aire-One (approved):

PR #3943, opened 2024-08-20
Comment and documentation improvement

aidinio:
I tried to make some details in the documentation more explicit.


lib/gears/timer.lua
```diff
@@ -145,7 +145,7 @@ end
 -- @see start
 -- @see stop
 
---- The timer timeout value.
+--- The timer timeout value in seconds.
```
actionless, line 148: this i like 👍
but other changes look like just replacing one synonymous term to other 🤷

aidinio: I think specifying that the "countdown" "immediately" starts, would be more clear than stating that the timer "automatically starts".

actionless:
https://github.com/awesomeWM/awesome/actions/runs/10548999597/job/29229634679?pr=3943

lib/wibox/widget/slider.lua
```diff
@@ -89,7 +89,10 @@ local slider = {mt={}}
 -- @propemits true false
 -- @propbeautiful
 
---- The cursor icon while grabbing the handle. The available cursor names are located in "/usr/include/X11/cursorfont.h" without the XC_ prefixes.
+--[[
```
sclu1034, line 92: Please verify your changes locally before pushing, because now this property will not show up at all.

The first line in doc comments must start with `---` (three dashes). And block comments are not supported in doc comments, it must all be line comments.

sclu1034 (changes requested):

lib/wibox/widget/slider.lua
aidinio: Thanks I'll fix it. But how can I verify the documentation code output locally?

aidinio: I also added two more comments under `@tparam[opt] string args.handle_cursor` so they would describe it. Would they show up? I didn't find anything related in the LDoc documents.
```lua
-- @tparam[opt] number args.handle_border_width The handle border width.
-- @tparam[opt] string args.handle_cursor
-- The cursor icon while grabbing the handle.
-- The available cursor names are located in "/usr/include/X11/cursorfont.h" without the XC_ prefixes.
-- @tparam[opt] gears.shape args.bar_shape The bar (background) shape.
-- @tparam[opt] number args.bar_height The bar (background) height.```

Comment (actionless) [reply lib/wibox/widget/slider.lua]: the best way to find it out to build the docs locally as sclu suggested above

Comment (aidinio) [reply lib/wibox/widget/slider.lua]: I built the docs locally. They docs are rendered correctly, but unfortunately, ldoc doesn't seem to write the two lines of comment in front of `handle_cursor` and `beautiful.slider_handle_cursor` descriptions under "Object properties" and "Theme variables"

Comment (aidinio) [reply lib/wibox/widget/slider.lua]: Maybe the line length limit should be changed, so we can write longer descriptions in one line.

File: lib/wibox/widget/slider.lua [hunk 3]
```diff
@@ -89,7 +89,10 @@ local slider = {mt={}}
...
+    The cursor icon while grabbing the handle.
+    The available cursor names are located in "/usr/include/X11/cursorfont.h" without the XC_ prefixes.
```
Aire-One: Isn't https://github.com/awesomeWM/awesome/blob/master/docs/common/cursor_c.ldoc based on the same source (hence a better documentation than pointing a file on the system)?

lib/wibox/widget/slider.lua
sclu1034: > ldoc doesn't seem to write the two lines of comment in front of `handle_cursor` and `beautiful.slider_handle_cursor` descriptions under "Object properties" and "Theme variables"

That's by design. Those tables are a table of contents, and specifically only show the first sentence (and shouldn't be gamed by making one extra long sentence). The full text is displayed when you scroll/jump to actual content section.

lib/wibox/widget/slider.lua
sclu1034: I think it's too big to include at each occurrence, though.
Maybe include it once at a central place like `wibox.cursor`, then mention that here.

aidinio: So do you mean I should make a whole new page in the document, then include those cursors there? Where should I make that file?

Also, where does the `cursor_c.ldoc` is even used in the documentation? I searched the compiled documents, and no page contained those names.

sclu1034: No new file needed. `wibox.cursor` is a property like the ones you've edited so far: https://awesomewm.org/apidoc/popups_and_bars/wibox.html#cursor. Since that's also the value that `slider.handle_cursor` and other widgets change internally, it's a good fir to point to as a single place to provide more detailed information.

It's currently used at https://github.com/awesomeWM/awesome/blob/master/root.c#L440

aidinio: To clarify, do you think that a list of cursors names should be added in https://awesomewm.org/apidoc/popups_and_bars/wibox.html#cursor, then referenced wherever necessary?

sclu1034: Yes. Though not just _a_ list, but `cursor_c.ldoc` specifically.

aidinio: I searched and asked, but still not sure how to include an ldoc file inside a lua file's ldoc comment. The `root.c` file that uses the ldoc table is a .c file and I don't know how to do the same in a .lua file.

Aire-One: This is the comment `@DOC_cursor_c_COMMON@` that is replaced by the actual content of the file at build time.

aidinio: > This is the comment `@DOC_cursor_c_COMMON@` that is replaced by the actual content of the file at build time.
> 
> See https://github.com/awesomeWM/awesome/blob/master/docs%2Fload_ldoc.cmake

I used the same syntax in a lua file but it didn't work (`@DOC_cursor_c_COMMON@` is used in a C file)

aidinio: I found out why the @DOC_...@ strings were not getting replaced by their corresponding files. It was because I was building the docs directly from the `/docs` directory. Instead, I had to build awesome first, then build the docs from `/build/docs` . Now I build the docs by running `ldoc .` in `/build/docs`.

aidinio: I encountered another problem while including the cursor names in the slider widget docs. When I change

```lua
--- The cursor icon while grabbing the handle. The availabe cursor names are:
-- @property handle_cursor
-- @tparam[opt="fleur"] string|nil handle_cursor
-- @propbeautiful
-- @see mousegrabber
```
to
```lua
--- The cursor icon while grabbing the handle. The availabe cursor names are:
--
--@DOC_cursor_c_COMMON@
--
-- @property handle_cursor
-- @tparam[opt="fleur"] string|nil handle_cursor
-- @propbeautiful
-- @see mousegrabber
```
the section explaining handle_cursor completely disappears from the compiled docs, and I get this error while building the docs:
```
WARNING: 	beautiful.slider_handle_cursor	is not used by anything, add @usebeautiful or @propbeautiful
```

aidinio: The problem was because the lines in `cursor_c.ldoc` file began with '*' but for being included in another lua file, they needed to be started with '--'. I made a new file called `cursor_c_lua.ldoc` with new lines starting with `--`.

lib/wibox/widget/slider.lua
```diff
@@ -228,7 +232,10 @@ local slider = {mt={}}
 -- @tparam[opt=gears.shape.rectangle] gears.shape shape
 -- @see gears.shape
 
---- The handle grab cursor.
+--- The cursor icon while grabbing the handle.
+-- The availabe cursor names are:
```
actionless, line 236: typo

actionless: (and same word on other lines)

aidinio: Thanks for the review. The typo is now fixed.

actionless (approved):

aidinio (commented):
No issues found

docs/common/cursor_c_lua.ldoc
```diff
@@ -0,0 +1,80 @@
+-- <div class='flex-list'>
```
Aire-One, line 1: This file should be named `cursor.ldoc`. the `_c` is to indicate the comments syntax is C style instead of Lua.

Aire-One, line 1: This line shouldn't have the initial `--` because when the build tool copies the file content, there is already a comment marker in the destination file (`--@DOC_cursor_c_lua_COMMON@`). Btw this "error" should be visible in the documentation after you generate it.

lib/wibox/widget/slider.lua
```diff
@@ -565,7 +572,9 @@ end
 -- @tparam[opt] number args.handle_width The slider handle width.
 -- @tparam[opt] color args.handle_border_color The handle border_color.
 -- @tparam[opt] number args.handle_border_width The handle border width.
--- @tparam[opt] string args.handle_cursor The handle grab cursor.
+-- @tparam[opt] string args.handle_cursor
+-- The cursor icon while grabbing the handle.
+-- The available cursor names are listed under handle_cursor, in the "Object properties" section.
```
Aire-One, line 577: You should add a few (2 or 3 I don't remember) extra spaces leading the additional lines

aidinio: Fixed. Thanks for the feedback!

docs/common/cursor_c_lua.ldoc
aidinio: Fixed.

Aire-One: Fair point. Let's just ignore my comment then because making these filenames more consistent is out of the scope of this PR 🤷

Aire-One (approved):

docs/common/cursor_c_lua.ldoc
Aire-One: Ah you did the change for this file anyway. Nice 👍

aidinio: > Ah you did the change for this file anyway. Nice 👍

Thanks. I might open another PR for changing those file names.

actionless (approved):

PR #3945, opened 2024-08-25
Add mwfact resizing to spiral layout

MorellThomas:
Add mwfact resizing to spiral layout.
Works the same way as in tile layout. Considers the left window of spiral as the master window.

### DIFF
--- a/lib/awful/layout/suit/spiral.lua
+++ b/lib/awful/layout/suit/spiral.lua
@@ -11,6 +11,7 @@
 -- Grab environment we need
 local ipairs = ipairs
 local math = math
+local tag = require("awful.tag")
 
 --- The spiral layout layoutbox icon.
 -- @beautiful beautiful.layout_spiral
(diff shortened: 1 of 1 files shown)


lib/awful/layout/suit/spiral.lua
```diff
@@ -11,6 +11,7 @@
 -- Grab environment we need
 local ipairs = ipairs
 local math = math
+local tag = require("awful.tag")
```
actionless, line 14: why?

MorellThomas: Valid point. Seems I forgot to remove that when cleaning up my changes. My bad.

actionless: also you need this: https://github.com/awesomeWM/awesome/blob/master/lib/naughty/core.lua#L16

actionless (approved):
thanks!

lib/awful/layout/suit/spiral.lua
```diff
@@ -25,13 +26,22 @@ local math = math
 local spiral = {}
 
 local function do_spiral(p, is_spiral)
+    local t = p.tag or capi.screen[p.screen].selected_tag
     local wa = p.workarea
     local cls = p.clients
     local n = #cls
     local old_width, old_height = wa.width, 2 * wa.height
+    local mwfact = t.master_width_factor + 0.5
```
Aire-One, line 34: Why `+ 0.5`?

MorellThomas: This is nessecary for it the default to behave as it did before this patch. Without changing the mwfact by pressing the key, the windows are split in the center of the screen. Without adding `+ 0.5`, the default center would have the left window occupy only a small part on the left of the screen and it cannot be resized until the right side.

Aire-One: @MorellThomas, if you are still around here, can you add a comment about that in the code? I don't like to have magic-number floating in code without explanation.

PR #3947, opened 2024-08-25
Move the C coverage to the CodeCov GitHub Action

### LINKED ISSUE #3934: codecov pipeline not responding in PRs
i've noticed that lately in all of the currently opened and approved PRs (https://github.com/awesomeWM/awesome/pulls?q=is%3Apr+is%3Aopen+review%3Aapproved), ex. https://github.com/awesomeWM/awesome/pull/3927 https://github.com/awesomeWM/awesome/pull/3930  https://github.com/awesomeWM/awesome/pull/3909

Elv13:
Same as Last PR. Now I expect the Lua coverage to work and the C coverage. I did it in 2 PRs to make sure the delta works.

I also added a `-j16` in the example test step because it is I/O bound and it has enough ram. This should speed-up the CI (maybe?).

Fix #3934

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -240,10 +240,6 @@ jobs:
         if: matrix.coverage
         run: sudo -H luarocks install cluacov
 
-      - name: Install codecov.io uploader
-        if: matrix.coverage == 'codecov'
-        run: wget -O /tmp/codecov-bash https://codecov.io/bash
-
       # Check out repository to ${{ github.workspace }}
       # Automatically picks the current branch/PR
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -308,7 +304,7 @@ jobs:
         run: cd "${{ github.workspace }}/build" && make check-unit
 
       - name: Run examples tests
-        run: cd "${{ github.workspace }}/build" && make check-examples
+        run: cd "${{ github.workspace }}/build" && make check-examples -j16
```
actionless, line 307: shouldn't it be `-j$(nproc)` ?

Elv13: No, the examples tests are not CPU bound, they are I/O bound. For example, I use `-j99` on my 16 thread desktop and it's 4x faster than `-j17`. As long as it doesn't take more ram than the VM has, the larger the number is, the faster it goes.

actionless (approved):

Aire-One (approved):

.github/workflows/main.yml
Aire-One: Maybe add a comment with this explanation?

PR #3960, opened 2024-09-30
Fix code example in timer.lua

elig0n:
I believe this is what is meant

### DIFF
--- a/lib/gears/timer.lua
+++ b/lib/gears/timer.lua
@@ -38,7 +38,7 @@
 --            awful.spawn.easy_async(
 --                {"sh", "-c", "acpi | sed -n 's/^.*, \([0-9]*\)%/\1/p'"},
 --                function(out)
---                    mybatterybar.battery = out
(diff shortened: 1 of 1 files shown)


lib/gears/timer.lua
```diff
@@ -38,7 +38,7 @@
 --            awful.spawn.easy_async(
 --                {"sh", "-c", "acpi | sed -n 's/^.*, \([0-9]*\)%/\1/p'"},
 --                function(out)
---                    mybatterybar.battery = out
+--                    mybatterybar.set_battery(out)
```
actionless, line 41: shouldn't both work?

sclu1034: Yes. Like any other widget, this has the `__newindex` handler for it.

sclu1034 (changes requested):
The project's preferred style is to use properties via the `__newindex` handler of `gears.object`. This usage snippet should reflect that and should therefore stay as it was.

Issue #3963, opened 2024-10-05
show help with search or fuzzy finder (like telescope in nvim)


actionless:
### Discussed in https://github.com/awesomeWM/awesome/discussions/3962

<div type='discussions-op-text'>

<sup>Originally posted by **DeepReef11** October  4, 2024</sup>
I  think the help panel would be much more useful if it had at least a search function and even better if it could be a fuzzy finder.</div>

kosorin:
Maybe I could do that, or at least some very basic searching. 

In my config, I completely rewrote the default `awful.hotkeys_popup`. It doesn't perform fuzzy searching, but a simple search is good enough for me.

It takes a query string, splits it by whitespace, and then tries to match every word in the descriptions and keys. 

Matching keys is a bit tricky. The whole word must match exactly. For example, if the query is "sup", it wouldn't make sense to highlight every keybinding containing the "super" key. Another example is if there is a keybinding `super + ⬅️`, the user can find it by typing "super left".

So, when the `awful.hotkeys_popup` is shown, it could automatically call `awful.prompt.run` and highlight the keybinding descriptions as the user types. But I'm uncertain whether this aligns with awesomewm philosophy.

![image](https://github.com/user-attachments/assets/d2abf855-27e9-47be-b0e1-fd039ec668e8)

https://github.com/user-attachments/assets/29ccc95a-7ca8-4a73-a432-179886932c4e

actionless:
yup, looks solid 👌😺

actionless:
> So, when the awful.hotkeys_popup is shown, it could automatically call awful.prompt.run and highlight the keybinding descriptions as the user types

you still need though to handle Escape key (and mb make it configurable to set on which keys it could be closed)

Thomashighbaugh:
love the idea and kosorins hotkeys popup is truly impressive!

PR #3970, opened 2024-10-18
Find in the hotkeys popup

kosorin:
Implements #3963

- find in description only
- no fuzzy or other sophisticated searching
- uses `awful.prompt.run`
- `show_help()` does not return `awful.keygrabber` (breaking change?)

## Theme variables

| Theme | Default value | Description |
|-------|---------------|-------------|
| `highlight_bg` | `"#ffff00"` | The highlighted text background color. |
| `highlight_fg` | `"#000000"` | The highlighted text foreground color. |
| `find_fg_cursor` | - | The find prompt cursor foreground color. |
| `find_bg_cursor` | - | The find prompt cursor background color. |
| `find_ul_cursor` | - | The find prompt cursor underline style. |
| `find_prompt` | `"<b>Find: </b>"` | The find prompt text. |
| `find_font` | `font` | The find prompt text font. |
| `find_margin` | `group_margin` | Margin around the find prompt. |

hotkeys_popup_find

### DIFF
--- a/lib/awful/hotkeys_popup/widget.lua
+++ b/lib/awful/hotkeys_popup/widget.lua
@@ -682,7 +682,7 @@ function widget.new(args)
         )
         local group_label_height = line_height + self.group_margin
         -- -1 for possible pagination:
-        local max_height_px = wibox_height - group_label_height - find_data.height
+        local max_height_px = wibox_height - group_label_height - find_data.container_height
 
         local joined_descriptions = ""
         for i, key in ipairs(keys) do
@@ -818,21 +818,27 @@ function widget.new(args)
         return pages
     end
 
-    function widget_instance:_create_find_data()
-        local margin = self.find_margin
-        local textbox = wibox.widget.textbox()
-        local container = wibox.container.margin(textbox, margin, margin, margin, margin)
-        local height = beautiful.get_font_height(self.find_font) + 2 * margin
-        return {
-            textbox = textbox,
-            container = container,
-            height = height,
+    function widget_instance:_create_find_data(enable_find)
+        local data = {
+            enabled = enable_find,
+            textbox = wibox.widget.textbox(),
             groups = {},
             last_query = "",
         }
+
+        if enable_find then
+            local margin = self.find_margin
+            data.container = wibox.container.margin(data.textbox, margin, margin, margin, margin)
+            data.container_height = beautiful.get_font_height(self.find_font) + 2 * margin
+        else
+            data.container = wibox.widget.empty
+            data.container_height = 0
+        end
+
+        return data
     end
 
-    function widget_instance:_create_wibox(s, available_groups, show_awesome_keys)
+    function widget_instance:_create_wibox(s, available_groups, show_awesome_keys, enable_find)
         s = get_screen(s)
         local wa = s.workarea
         local wibox_height = (self.height < wa.height) and self.height or
@@ -840,7 +846,7 @@ function widget.new(args)
         local wibox_width = (self.width < wa.width) and self.width or
             (wa.width - self.border_width * 2)
 
-        local find_data = self:_create_find_data()
+        local find_data = self:_create_find_data(enable_find)
 
         local pages = self:_create_pages(s, available_groups, show_awesome_keys, wibox_width, wibox_height, find_data)
 
@@ -920,10 +926,10 @@ function widget.new(args)
                 keypressed_callback = function(_, key)
                     if key == "Prior" then
                         w_self:page_prev()
-                        return true
                     elseif key == "Next" then
                         w_self:page_next()
-                        return true
+                    elseif not w_self.find_data.enabled then
+                        w_self:hide()
                     end
                 end,
             }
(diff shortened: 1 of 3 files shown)


lib/awful/hotkeys_popup/widget.lua
```diff
@@ -371,6 +427,22 @@ function widget.new(args)
             beautiful.hotkeys_description_font or "Monospace 8"
         self.group_margin = args.group_margin or
             beautiful.hotkeys_group_margin or dpi(6)
+        self.highlight_bg = args.highlight_bg or
+            beautiful.hotkeys_highlight_bg or "#ffff00"
```
actionless, line 431: please default to other `beautiful.*` variables which are for sure defined in the default theme.lua instead of falling back to hardcoded colors

actionless:
thanks for submitting, i found small issue right from the description (and left a separate commit), i'll start reviewing the code itself later on in upcoming days or mb even hours

actionless:
and also the test is failing:

  - /home/runner/work/awesome/awesome/tests/test-awesomerc.lua: Error: running function for step 13/18 (@3): /home/runner/work/awesome/awesome/tests/test-awesomerc.lua:289: assertion failed!

actionless:
regarding keygrabber question, i'll need first to understand:

1) what was the initial intention of returning it there
2) why it's not possible to do it anymore

actionless:
also, mb add a bool option to args for closing hotkeys on a any keypress (the current behavior), so the users who used to that, could restore that old default behavior

actionless:
i think `find_prompt` should be moved from theme option to args

lib/awful/hotkeys_popup/widget.lua
kosorin: I like this bright yellow, because the point is to easily see highlighted text. For example firefox uses `#ff00ff`, also bright.

What about `beautiful.bg_urgent` (red)? I haven't found any other beautiful colors that fit it.

actionless: yup, just choose any placeholder.
people who care more would set the right theme key anyway

kosorin:
> regarding keygrabber question, i'll need first to understand:
> 
>     1. what was the initial intention of returning it there
> 
>     2. why it's not possible to do it anymore

Because I use `awful.prompt` I no longer have access to the underlying keygrabber. Why was it returning keygrabber? I don't know, it doesn't make much sense to me.

Actually, I need to check if the prompt is stopped gracefully. Right now I just call `awful.keygrabber.stop()`.

> also, mb add a bool option to args for closing hotkeys on a any keypress (the current behavior), so the users who used to that, could restore that old default behavior

I added `show_args.enable_find` (enabled by default). If false, the prompt is not visible and the popup is hidden on any key.

> i think `find_prompt` should be moved from theme option to args

I don't know. What are the odds that you want multiple prompts for different hotkey popups? It's just a simple text "find: ", nothing more so I thought was easier for user to set it once in the theme.

actionless:
>  What are the odds that you want multiple prompts for different hotkey popups

it wasn't a point about having multiple different prompts, but about untying translation and style - and as you can see from our existing themes, it's indeed not a used practice storing such sort of info there

> I added show_args.enable_find (enabled by default). If false, the prompt is not visible and the popup is hidden on any key.

thanks 👍

> Because I use awful.prompt I no longer have access to the underlying keygrabber. Why was it returning keygrabber? I don't know, it doesn't make much sense to me.

ok i'll try to dive into `git blame` to figure that out 😺

actionless:
style checking is failing in the CI btw

lib/awful/hotkeys_popup/widget.lua
```diff
@@ -371,6 +427,22 @@ function widget.new(args)
             beautiful.hotkeys_description_font or "Monospace 8"
         self.group_margin = args.group_margin or
             beautiful.hotkeys_group_margin or dpi(6)
+        self.highlight_bg = args.highlight_bg or
+            beautiful.hotkeys_highlight_bg or beautiful.bg_urgent
+        self.highlight_fg = args.highlight_fg or
+            beautiful.hotkeys_highlight_fg or beautiful.fg_urgent
+        self.find_prompt = args.find_prompt or
+            beautiful.hotkeys_find_prompt or "<b>Find: </b>"
```
actionless, line 435: as i mentioned in previous comment this shouldn't be defined in beautiful, as its role defining theme, not translation/labels

actionless (changes requested):
⬆️

actionless:
also, this suggestion may be out of scope of this PR, but would be nice to have an option instead of just highlighting search results - to show only the matching items (and their respective group headers)

kosorin:
> > style checking is failing in the CI btw
> 
> this is still actual

I see, I fixed it and now there is another issue. How do you setup your local dev environment so I can see these warnings/errors before pushing?



> also, this suggestion may be out of scope of this PR, but would be nice to have an option instead of just highlighting search results - to show only the matching items (and their respective group headers)

Widgets' positions are calculated only once when they are created. Then they are cached. Highlighting the text will not change their position. However, hiding the text requires recalculating all positions. Which is a non-trivial task given the code. So maybe some other time...

actionless:
on arch linux it's normally enough to have the same libraries as in makedeps of the PKGBUILD


for ubuntu see:

https://github.com/awesomeWM/awesome/blob/master/.github/workflows/main.yml

common instructions for other distros are here:

https://awesomewm.org/apidoc/documentation/10-building-and-testing.md.html

PR #3975, opened 2024-11-27
Initialize allocated memory and check that node is not NULL before using it

stderr64:
Changed malloc to calloc so memory of node is initialized and added assert to check that the pointer to viewport_t struct is not NULL before using the viewport_t struct.


actionless:
adding @Elv13 because of https://github.com/awesomeWM/awesome/commit/433898599d7eb6652c1438e7d85d9aeca27abb5a#diff-20a455ee6335fa68af6e422ae31fa38d29e650c08079a7ed5c35345f2e7cd736

Issue #3867, opened 2023-11-14
KP_ENTER is ignored in main menu


unknown321:
Output of `awesome --version`:
```
$ awesome --version
awesome v4.3 (Too long)
 • Compiled against Lua 5.3.6 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

**How to reproduce the issue:**

Press Super-W to open main menu. Highlight entry (using arrow keys/mouse) and press `numpad enter` (KP_ENTER).

**Actual result:**

Nothing happens

**Expected result:**

Menu entry starts, like on `regular enter` press.

unknown321:
Patch:

```diff
---- /usr/share/awesome/lib/awful/menu.lua	2023-11-17 21:37:00.270213981 +0300
+++ /usr/share/awesome/lib/awful/menu.lua	2023-11-17 21:37:46.518324482 +0300
@@ -107,7 +107,7 @@
               down = { "Down", "j" },
               back = { "Left", "h" },
               exec = { "Return" },
-              enter = { "Right", "l" },
+              enter = { "Right", "l", "KP_Enter" },
               close = { "Escape" } }
```

Not sure if this is a right place to patch, still works for me.

Elv13:
Can you submit a pull request? You can do this directly on the GitHub web UI by pressing edit or in a terminal

PR #3979, opened 2024-12-23
awful.menu: map KP_Enter similar to Return

sicelo:
In Awesome's menu system, the KP_Enter key is currently used the same as the Right key (to descend into a child menu). This is unexpected, since in general, users expect KP_Enter to work similar to the Return key (to run the highlighted action/command). This commit corrects that behavior.

Related: https://github.com/awesomeWM/awesome/issues/3867


actionless (approved):

keryell (approved):
That makes sense!

Aire-One (approved):

PR #3989, opened 2025-02-04
fix(ci: main: xcb-errors): install from tarball instead of git tag

actionless:
See https://github.com/awesomeWM/awesome/actions/runs/13127642292/job/36626903577?pr=3988


Aire-One (approved):

sclu1034 (approved):
I assume, you'll autosquash that `fixup!` commit?

PR #3991, opened 2025-02-19
Add a rect shape with corners having different radii

aidinio:
Added a shape that can have a different radius on each corner.

### DIFF
--- a/lib/gears/shape.lua
+++ b/lib/gears/shape.lua
@@ -317,6 +317,66 @@ function module.partially_rounded_rect(cr, width, height, tl, tr, br, bl, rad)
     cr:close_path()
 end
 
+--- A rounded rect with individually defined corner radii.
+--
+-- @DOC_gears_shape_individually_rounded_rect_EXAMPLE@
+--
+-- @param cr A cairo context
+-- @tparam number width The shape width
+-- @tparam number height The shape height
+-- @tparam boolean tl The top left corner's radius
+-- @tparam boolean tr The top right corner's radius
+-- @tparam boolean br The bottom right corner's radius
+-- @tparam boolean bl The bottom left corner's radius
+-- @noreturn
+-- @staticfct gears.shape.individually_rounded_rect
+function module.individually_rounded_rect(cr, width, height, tl, tr, br, bl)
+    local corners = {tl = tl, tr = tr, br = br, bl =  bl}
+    for key, val in pairs(corners) do
+        corners[key] = val or 10
+        if width / 2 < val then
+            corners[key] = width / 2
+        end
+        if height / 2 < val then
+            corners[key] = height / 2
+        end
+    end
+
+    -- In case there is already some other path on the cairo context:
+    -- Make sure the close_path() below goes to the right position.
+    cr:new_sub_path()
+
+    -- Top left
+    if tl then
+        cr:arc( corners.tl, corners.tl, corners.tl, math.pi, 3*(math.pi/2))
+    else
+        cr:move_to(0,0)
+    end
+
+    -- Top right
+    if tr then
+        cr:arc( width-corners.tr, corners.tr, corners.tr, 3*(math.pi/2), math.pi*2)
+    else
+        cr:line_to(width, 0)
+    end
+
+    -- Bottom right
+    if br then
+        cr:arc( width-corners.br, height-corners.br, corners.br, math.pi*2 , math.pi/2)
+    else
+        cr:line_to(width, height)
+    end
+
+    -- Bottom left
+    if bl then
+        cr:arc( corners.bl, height-corners.bl, corners.bl, math.pi/2, math.pi)
+    else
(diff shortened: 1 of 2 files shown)


lib/gears/shape.lua
```diff
@@ -317,6 +317,66 @@ function module.partially_rounded_rect(cr, width, height, tl, tr, br, bl, rad)
...
+--
+-- @DOC_gears_shape_individually_rounded_rect_EXAMPLE@
+--
+-- @param cr A cairo context
+-- @tparam number width The shape width
+-- @tparam number height The shape height
+-- @tparam boolean tl The top left corner's radius
+-- @tparam boolean tr The top right corner's radius
+-- @tparam boolean br The bottom right corner's radius
+-- @tparam boolean bl The bottom left corner's radius
```
sclu1034, line 330: `boolean` doesn't look right here.

lib/gears/shape.lua
```diff
@@ -317,6 +317,66 @@ function module.partially_rounded_rect(cr, width, height, tl, tr, br, bl, rad)
...
+-- @noreturn
+-- @staticfct gears.shape.individually_rounded_rect
+function module.individually_rounded_rect(cr, width, height, tl, tr, br, bl)
+    local corners = {tl = tl, tr = tr, br = br, bl =  bl}
+    for key, val in pairs(corners) do
+        corners[key] = val or 10
```
sclu1034, line 336: `nil` is a valid value further down, so it's a waste of resources to assign a fallback value here and do all the calculations below.

sclu1034 (changes requested):

lib/gears/shape.lua
```diff
@@ -317,6 +317,66 @@ function module.partially_rounded_rect(cr, width, height, tl, tr, br, bl, rad)
     cr:close_path()
 end
 
+--- A rounded rect with individually defined corner radii.
+--
+-- @DOC_gears_shape_individually_rounded_rect_EXAMPLE@
```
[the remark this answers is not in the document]
aidinio: Yes I have compiled the docs and there was no problem compiling.

lib/gears/shape.lua
aidinio: Thanks! I'll fix it.

lib/gears/shape.lua
aidinio: Thanks for the observation :). I removed it.

lib/gears/shape.lua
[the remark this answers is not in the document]
aidinio: I removed the line.

aidinio:
I'm not sure why the test failed.

lib/gears/shape.lua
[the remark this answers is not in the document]
sclu1034: > Yes I have compiled the docs and there was no problem compiling.

By "verifying" I mean looking at the page to see if the output matches expectations. In this case, the lack of an example should have been noticeable.

> I removed the line.

The desired approach would be to have a rendered example show up in the docs, which requires creating the file that this directive points to, not to remove the line.

lib/gears/shape.lua
sclu1034: Since `nil` is a valid value, it also needs to be flagged as optional.

sclu1034:
> I'm not sure why the test failed.

Codecov fails because none of your lines are covered by tests and you end up below the target. That's another thing documentation examples are used for: they also feed into the testing framework.

aidinio:
> > I'm not sure why the test failed.
> 
> Codecov fails because none of your lines are covered by tests and you end up below the target. That's another thing documentation examples are used for: they also feed into the testing framework.

Thanks for the help. I also added tests.

sclu1034 (approved):

Aire-One (approved):

Aire-One:
Hi @aidinio, sadly most other maintainers are AWOL for some time now... Let's try to ping @awesomeWM/write-access

actionless (approved):

actionless (changes requested):
but coverage is not enough: https://github.com/awesomeWM/awesome/pull/3991#issuecomment-2669332367

aidinio:
> but coverage is not enough: [#3991 (comment)](https://github.com/awesomeWM/awesome/pull/3991#issuecomment-2669332367)

I've already added tests which resolve the coverage issue.

actionless (approved):
yeah, codecov stuck on previous commit: https://app.codecov.io/gh/awesomeWM/awesome/pull/3991/commits

actionless:
no, as CI currently broken: https://github.com/awesomeWM/awesome/issues/3994

actionless:
now it's finally fixed, so you could rebase 👌😸

aidinio:
> now it's finally fixed, so you could rebase 👌😸

Thanks!

Aire-One:
@mergifyio rebase

PR #3995, opened 2025-03-10
fix(awesome-client): ensure there are CLI arguments before reexec'ing with rlwrap

demostanis:
in my `.zshrc`, i'm hooking `cd` with `chpwd(){...}`, to execute lua functions through awesome-client whenever i switch directories in my terminal.
i installed rlwrap lately and noticed `ls` wouldn't be executed anymore if i typed really fast a command such as
`sleep 1`
`cd`
`ls`
because `cd` (which calls rlwrap now) reads from stdin, so my shell can't read `ls` anymore.

currently, awesome-client won't execute rlwrap if the stdin is not a tty (e.g. when piping through awesome-client).
however, it should also check that awesome-client is not called with arguments (since there'd be no point in reading the stdin).

### DIFF
--- a/utils/awesome-client
+++ b/utils/awesome-client
@@ -5,7 +5,7 @@
 # when used with "|tail".
 set -o pipefail
 
-if [ -t 0 ]  # is a tty.
+if [ -t 0 -a $# = 0 ]  # is a tty.
 then
     # rlwrap provides readline functionality for "read", which is more enhanced
     # than bash's "read" itself.


Aire-One (approved):
It may need rebase for the CI to work, and mergify to pick up the PR and merge.

actionless:
@Aire-One one of the tests (always the same) is still being flaky: https://github.com/awesomeWM/awesome/actions/runs/18339824060/job/52242235689?pr=3995#step:27:304

if it's still causing problems - mb we could restart it several times. i think test framework itself doesn't have any options like that, but mb it could be just some sort of iteration in the body of the test itself

Issue #3994, opened 2025-02-28
[Action Required] Ubuntu-20.04 hosted runner image is closing down


psychon:
Hi there,

I got a mail and I think it is about AwesomeWM. It says

> The Ubuntu 20.04 runner image will be fully unsupported by April 1, 2025. To raise awareness of the upcoming removal, we will temporarily fail jobs using Ubuntu 20.04. Builds that are scheduled to run during the brownout periods will fail. The brownouts are scheduled for the following dates and times:      
*March 4 14:00 UTC – 22:00 UTC 
*March 11 13:00 UTC – 21:00 UTC
*March 18 13:00 UTC – 21:00 UTC
*March 25 13:00 UTC – 21:00 UTC
> 
> What you need to do      
> Jobs using the ubuntu-20.04 YAML workflow label should be updated to ubuntu-22.04, ubuntu-24.04 or ubuntu-latest.

I am not totally sure (because why would it be helpful if the mail mentioned such details...), but [I guess this](https://github.com/awesomeWM/awesome/blob/691e36425a645e54402cb04efc4c2b00d73051bd/.github/workflows/main.yml#L26) mail [is (at least partially) about this repo](https://github.com/awesomeWM/awesome/blob/691e36425a645e54402cb04efc4c2b00d73051bd/.github/workflows/apidoc.yml#L25).

actionless:
alternative option - using docker for tests, the time overhead won't be notable, especially if using caching of docker layers

plus that would allow to 100% reproduce ci runs locally if debugging some weird ci failures

sclu1034:
The runner used for the workflow needs to be updated one way or another. Whether you run your tests directly on the runner, or run them inside a docker container on the runner won't change that fact.

On the contrary, it doubles the things you need to keep updated, because you now added the docker image to the mix. And neither of them guarantee your code to be compatible for eternity.
Within the container, the test code may now be isolated from incompatibilities with the runner version, but is now subject to becoming incompatible to the docker image in exactly the same way. Either you need to update a runner from `ubuntu-20.04` to `ubuntu-24.04`, or you need to update a docker image from `ubuntu:20.02` to `ubuntu:24.02`.

---

And https://github.com/nektos/act exists to allow running GitHub Actions workflows on your local machine.

Aire-One:
[...]

PR #3999, opened 2025-03-22
chore(workflows): upgrade actions/cache from v2 to v4

Aire-One:
actions/cache@v2 is being deprecated, and older versions are removed: https://github.com/actions/cache?tab=readme-ov-file#%EF%B8%8F-important-changes

Only updating the version seems to be enough to unblock the CI (running workflows on my fork was a success https://github.com/Aire-One/awesome/pull/3)

### DIFF
--- a/.github/workflows/apidoc.yml
+++ b/.github/workflows/apidoc.yml
@@ -22,7 +22,7 @@ permissions:
 
 jobs:
   main:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     env:
       # Used for stable dates in documentation examples. See #2070.
@@ -56,10 +56,13 @@ jobs:
             gir1.2-gtk-3.0 \
             gir1.2-pango-1.0 \
             git \
+            imagemagick \
             libdbus-1-dev \
+            libgdk-pixbuf-2.0-dev \
             libgirepository1.0-dev \
             libnotify-bin \
             libpango1.0-dev \
+            librsvg2-dev \
             libstartup-notification0-dev \
             libx11-xcb-dev \
             libxcb-cursor-dev \
--- a/.github/workflows/apidoc.yml
+++ b/.github/workflows/apidoc.yml
@@ -37,7 +37,7 @@ jobs:
 
       - name: Cache apt packages
         id: cache-apt
-        uses: actions/cache@v2
+        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
         with:
           path: /var/cache/apt/archives
           # The trailing number serves as a version flag that can be incremented
@@ -87,7 +87,7 @@ jobs:
 
       - name: Cache luarocks
         id: cache-luarocks
-        uses: actions/cache@v2
+        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
         with:
           path: /tmp/luarocks
           key: ${{ github.workflow }}-${{ runner.os }}-luarocks-3.5.0
@@ -114,7 +114,7 @@ jobs:
           sudo -H luarocks install lgi
           sudo -H luarocks install ldoc
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       - name: Build Awesome version string
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -84,6 +84,15 @@ jobs:
         id: get-date
         run: echo "date=$(/bin/date -u "+%Y%W")" >> $GITHUB_OUTPUT
 
+      # Varibales doesn't work in matrix definition. The gh-actions-lua Actions
+      # install Lua in the workspace directory, so we refer to it with a relative
+      # path.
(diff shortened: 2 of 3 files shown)


sclu1034:
Given [recent events](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised), actions should better be pinned to SHAs now, not tags.

And #3994 will need to be fixed until April as well.

Elv13 (dismissed):

Aire-One:
```
There were 1 errors:
 - /home/runner/work/awesome/awesome/tests/test-awful-layout.lua: 2025-03-29 02:32:38 E: awesome: Error during a protected call: lib/awful/layout/init.lua:267: bad argument #-2 to 'geometry' (value in [-32768.0, 32767.0] expected, got 65576.0)
 ```

Well, this one is going to be funny...

Aire-One:
Since we have some activity today, I figured I should provide an update here. I have a slow-work-in-progress branch on my fork at https://github.com/awesomeWM/awesome/pull/4024.

- I have an (uncomited) Devcontainer config that reproduce the same conditions we have in the CI environment, so it eases the debug work
- I have tracked down with the help of GitHub Copilot the overflow error to some maths in the tile layout (https://github.com/awesomeWM/awesome/pull/4024/commits/1f587b61b1a7b6868c522d17c6b2e7781f1e7484). I'm not completely satisfied with this fix, but it's the less intrusive I could think of. 
- The apidoc generation in the CI has many changes I still need to review (because it seems not normal to me that generated images has changes)
- The LuaJIT tests fail consistently, I think because of some collectgarbage (?) (At least, that's a common line in every test that randomly fails)
- EDIT: codecov seems broken, I need to understand what happen

actionless:
i wonder whatever skipping those 2 (or their number is fluctuating?) problematic tests for luajit setup specifically be a reasonable way to unlock CI in order to start merging quite a a big accumulated queue of PRs?

actionless:
and thanks a lot for all the debugging done on this 🔥🔥🔥

sclu1034:
> * The LuaJIT tests fail consistently, I think because of some collectgarbage (?) (At least, that's a common line in every test that randomly fails)

Looks like "despite" would be the more accurate wording here.
The common case in `text-naughty-screen` and `test-screen-changes` is to drop a reference and then wait for it to be removed completely. The `collectgarbage` calls are there to speed things up by forcing the GC to run now, rather than having to rely on it to "eventually" clean up the references. Removing calls to `collectgarbage` will only make it worse, as it's then even less deterministic if/when the object gets cleaned up.

The fact that these tests fail suggests that somewhere, there is still a strong reference that prevents the GC from actually removing the underlying object.

Aire-One:
Sooooo it took me some additional researching and try-and-fails, but it actually seems like there was an issue with the memory management from LuaJIT with lightuserdata. Using an updated version of LuaJIT makes the error vanish.

Also, updating the Actions for CodeCov unblocked it! I'm making big progress :rocket:

Aire-One:
Hello, I've rebased/pushed the working commits from my private branch to here. The CI is working again!

Things that should be noted : 

- As mentioned earlier here, and described in the commit message for 73f551a26b9fc321fbba5272841536c7bb1fb16b: This is just a mitigation patch to allow the test to pass. I couldn't figure out the root cause for this error.
- The `Update API docs` workflow is green, but the script actually fails. The error is that the secret token used is only available in the main repository. When submitting a PR, the workflow run as from your fork, hence don't have access to the token. I was induced in error thinking something was wrong because lasts PRs from Elv had some comments from the "awesome bot". This is because they were branches from this repo and not his fork. So ultimately, we'll see if the script works only after merge, with the workflow triggered from master.
- I used the `luarocks/gh-actions-lua` to install LuaJIT from source in the CI. LuaJIT uses rolling release, so it sounds best suited to always (and only) test against the git-master version. If really needed, we can add (with new PRs) entries in the workflow matrix to test against legacies versions. In the long term, I think it would be benefited for us to use this Actions to install all Lua versions (and also to use the other Actions to install Luarocks).
- There is a new "Correct Lua variables paths" step in the main workflow, it is needed to use the `github.workspace` variable. I corrected both `LUAINCLUDE` and `LUALIBRARY`, but only the fix for `LUAINCLUDE` is strictly necessary. The path fix is needed to build Luarocks. I have kept the `--with-lua-include` flag, but I'm pretty sure we actually don't need it anymore with the `gh-actions-lua`.
- We may have issue with how CodeCov calculate the coverage. It was unnoticed with https://github.com/awesomeWM/awesome/pull/3947 that says `- Coverage   master:91.26%   diff:48.52%   +/-:-42.74%`. So we have somehow lost about half of the coverage.
- With the very last debug session, I had a new error with integration tests with Lua5.2 `/home/runner/work/awesome/awesome/tests/test-awful-client.lua: Error: timeout waiting for signal in step 11/138 (@20) (.../runner/work/awesome/awesome/tests/test-awful-client.lua:221).` If it doesn't happen here, I'll skip debugging it now.

actionless (approved):
👌😸🔥

Aire-One:
Thank you for the merge @actionless :rocket:

actionless:
considering that @Elv13 previously approved the prev version of the PR i think there is no need to futher block the CI, so finally we could merge some of long-waiting PRs

thanks again for your great pipeline investigation, @Aire-One 🚀

PR #4010, opened 2025-06-27
Menubar: Added theme.menubar_bg_base

SuperDuperDeou:
Adding the theme.menubar_bg_base to lib/menubar/init.lua, this allows for the base widget to be drawn with a different background color than the items. This is especially useful with semitransparent themes to avoid unintentional alphas.

### DIFF
--- a/lib/menubar/init.lua
+++ b/lib/menubar/init.lua
@@ -53,6 +53,10 @@ end
 --- Menubar normal background color.
 -- @beautiful beautiful.menubar_bg_normal
 -- @param color
+--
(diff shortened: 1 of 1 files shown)


lib/menubar/init.lua
```diff
@@ -53,6 +53,10 @@ end
 --- Menubar normal background color.
 -- @beautiful beautiful.menubar_bg_normal
 -- @param color
+--
```
Aire-One, line 56:
```suggestion

```

I think the doc generation error is because of this empty comment. The line should be blank.

PR #4020, opened 2025-08-29
Possible CI fix - set DEBIAN_FRONTEND environment variable to noninteractive

aidinio:
Since the actions are taking a lot to execute (>24h) and finally getting a timeout error, I'm guessing the are waiting for user's input. We can resolve this issue by setting the environment variable `DEBIAN_FRONTEND` to `noninteractive`.

Issue: #3994


sclu1034:
Is there a particular reason why you didn't touch the runner version, even though you linked the issue that explains its deprecation and removal?
This is never going to work if you keep trying to run on a non-existent runner version.

PR #4023, opened 2025-09-16
Compute maximized geometry after applying titlebar (fixes #3778)

### LINKED ISSUE #3778: Maximized windows with titlebar have a bad size
I have the same issue as [https://github.com/awesomeWM/awesome/issues/1128](url)

```
awesome v4.3-1585-g1239cdf4b (Too long)
 • Compiled against Lua 5.4.4 (running with 0.9.2)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • Transparency enabled: yes
 • Custom search paths: no
```

Also, the rules section of my rc.lua is:

```
ruled.client.connect_signal("request::rules", function()
    ruled.client.append_rule {
        id         = "global",
        rule       = {},
        properties = {
            focus     = awful.client.focus.filter,
            raise     = true,
            screen    = awful.screen.preferred,
            placement = awful.placement.no_overlap + awful.placement.no_offscreen
        }
    }

    -- Floating clients.
    ruled.client.append_rule {
        id         = "floating",
        rule_any   = {
            instance = { "copyq", "pinentry" },
            class    = {
                "Arandr", "Blueman-manager", "Gpick", "Kruler", "Sxiv",
                "Tor Browser", "Wpa_gui", "veromix", "xtightvncviewer"
            },
            -- Note that the name property shown in xprop might be set slightly after creation of the client
            -- and the name shown there might not match defined rules here.
            name     = {
                "Event Tester", -- xev.
            },
            role     = {
                "AlarmWindow", -- Thunderbird's calendar.
                "ConfigManager", -- Thunderbird's about:config.
                "pop-up", -- e.g. Google Chrome's (detached) Developer Tools.
            }
        },
        properties = { floating = true }
    }

    -- Add titlebars to normal clients and dialogs
    ruled.client.append_rule {
        id         = "titlebars",
        rule_any   = { type = { "normal", "dialog" } },
        properties = { titlebars_enabled = true }
    }
end)
```

Maximize a window with a titlebar from an application that will maximize the window again when restarting.
Close the application.
Start the application.

**Actual result:**

The window is too large and its bottom is out of the screen. The window probably has been sized, then the titlebar added and window content moved to the bottom.

**Expected result:**

The maximized window + titlebar should not be larger than the screen (minus "task bar").

phantamanta44:
The issue in #3778 is essentially just that EWMH client hints are processed before the rule engine runs, which means the maximized client geometry is computed without accounting for titlebars (or borders). This can be partially solved by writing a rule that applies a maximizing geometry after the fact, like so:
```lua
{
  rule = { maximized = true },
  properties = { placement = awful.placement.maximize }
}
```
However, I have found that there are situations where this is insufficient, particularly when clients have the `maximized` property set by another rule.

This PR fixes this by deferring the `request::geometry` signal normally sent by `client_set_maximized_common` until after titlebars and borders are applied. This is done by emitting the signal during the first run of rule processing in `ruled.client` and disabling the geometry signal handler in `awful.permissions` until that happens.

I note that it's not possible to defer the entire EWMH hint check, since we want the `maximized` flag to be set before the rule engine runs. Otherwise, rules matching on `maximized` won't be processed correctly for new clients.

### DIFF
--- a/lib/awful/permissions/init.lua
+++ b/lib/awful/permissions/init.lua
@@ -365,6 +365,11 @@ local context_mapper = {
 function permissions.geometry(c, context, hints)
     if not pcommon.check(c, "client", "geometry", context) then return end
 
+    -- Don't update geometry if rules haven't been applied yet (e.g. when processing
+    -- EWMH client hints during client initialization), since a titlebar may perturb
+    -- the intended geometry later
+    if not c._border_geometry_rules_applied then return end
+
     local layout = c.screen.selected_tag and c.screen.selected_tag.layout or nil
 
     context = context or ""
--- a/lib/ruled/client.lua
+++ b/lib/ruled/client.lua
@@ -535,7 +535,6 @@ end
 --  rules.
 
 crules._execute = function(_, c, props, callbacks)
-
     -- Set the default buttons and keys
     local btns = amouse._get_client_mousebindings()
     local keys = akeyboard._get_client_keybindings()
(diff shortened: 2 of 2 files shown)


actionless:
@mergifyio rebase

Aire-One:
https://github.com/Mergifyio rebase

PR #4027, opened 2025-09-24
Fix: Make spawn.with_line_callback compatible with both old and new lgi

PhoeniX5971:
### Problem
Gio.UnixInputStream was removed in newer lgi2, causing spawn output handling to break.

### Solution
Use GioUnix.InputStream when available, fallback to Gio.UnixInputStream for older versions.

### Impact
- Compatible with both old and new lgi.
- No changes to callback logic or existing behavior.

### Testing
- Verified spawn output works for stdout and stderr on Arch Linux with latest lgi.
- Also works on older lgi versions.

### Related Issues
- None reported yet

### DIFF
--- a/lib/awful/spawn.lua
+++ b/lib/awful/spawn.lua
@@ -417,13 +417,20 @@ function spawn.with_line_callback(cmd, callbacks)
             done_callback()
         end
     end
+    -- Determine InputStream compatibly
+    local InputStream
+    pcall(function()
+        InputStream = lgi.GioUnix.InputStream
+    end)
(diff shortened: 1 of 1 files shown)


Aire-One:
Duplicate of https://github.com/awesomeWM/awesome/pull/4022

Aire-One:
As discussed in https://github.com/awesomeWM/awesome/pull/4022, we don't think this is necessary. Since it's now merged, I'll close this PR. 

Thank you for the PR, nonetheless.

PR #4032, opened 2025-10-10
fix missing dep on Debian 13

Konfekt:
Added installation of libxcb-xfixes0-dev for Debian-based systems.

### DIFF
--- a/docs/01-readme.md
+++ b/docs/01-readme.md
@@ -49,6 +49,7 @@ makepkg -fsri
 
 ```sh
 sudo apt build-dep awesome
+sudo apt install libxcb-xfixes0-dev 
 git clone https://github.com/awesomewm/awesome
 cd awesome
 make package


Aire-One:
Shouldn't `build-dep` includes the package?

Konfekt:
I thought this to be the idea, but maybe 4.3 from 2019 didn't need it yet?

Aire-One:
If this is a new dependency for building from git, I agree the README should mention it. Otherwise, I think this should be fixed by the Debian/apt team with the build-dep. 

At this point, I'm not sure in which case the `libxcb-xfixes0-dev` falls to. @Konfekt did you check if 4.3 is buildable only using build-dep? (Hmmm I guess the build-dep is already up-to-date, and this dependency is new 🤔)

Konfekt:
Yes, I followed the README in Debian 13, it complained about missing header files, this addition fixed it.

eylles:
for reference the [4.3-8 build-deps](https://salsa.debian.org/reiner/awesome/-/blob/debian/4.3-8/debian/control?ref_type=tags) on debian 13 do not require `libxcb-xfixes0-dev` for awesome 4.3

Issue #4035, opened 2025-11-03
Burrito application input shapes not handled correctly


DziubanMaciej:
## The application

Burrito is an overlay application for Guild Wars 2 game. It uses `XShapeCombineRegion` (which sends `ShapeRectangles` x11 packet) to be rendered on top of the game, while not blocking the input. It creates a full screen, floating, transparent window and has two modes of operation:
1.  Small icon at the top left part of the screen. The `ShapeRectangles` region is limited to only this icon. When this icon is clicked, it switches to mode 2.
2. Active menus. In this mode the menus for controlling the overlay are displayed and the `ShapeRectangles` region spans the whole screen. A red outline is also rendered to communicate it to the user. When all menus are closed, it switches to mode 1.

## The problem 
The overlay captures all input. It's as if the `ShapeRectangles` packet was ignored. It works with Gnome.

## What I tried
I thought that `ShapeRectangles` isn't implemented in Awesome, but it looks like it is. I created a simple application sending `ShapeRectangles` and it seems to be working correctly. The input is taken only in the region I specify. Burrito must be doing something different. Maybe it's a matter of different order of API calls?

I looked into Awesome's code and it looks like `xwindow_set_shape()` should be the place that services `ShapeRectangles` requests. However, it doesn't get any requests with `kind=XCB_SHAPE_SK_INPUT` (which is `2`). This is true for both Burrito (which doesn't work) and my test app (which works), so it may not be useful.

I'm using awesome v4.3. For debugging I built v4.3-1694-gd37e3c63-dirty.

See image below for logs I gathered:
- on the left there's AwesomeWM running inside Xephyr
- on the top right there are X11 calls made by Burrito (traced by xtruss)
- on the bottom right there are my `printf`s added to `xwindow_set_shape`. Note that `kind` is never `2`.

<img width="2391" height="1366" alt="Image" src="https://github.com/user-attachments/assets/aca13d82-0a0d-4f64-ac50-f160347f2be3" />

## Reproduction
Guild Wars 2 game is **not** needed to reproduce this.
1. Download Burrito from release https://github.com/AsherGlick/Burrito/releases/tag/burrito-1.1.0
2. Unzip
3. Run `burrito.x86_64`
4. Observe the app capturing all mouse input (this is not expected)
5. Click on the little burrito icon in the top left. It will be in a weird place, because it's hardcoded to nicely fit into game's interface when it's fullscreen.
6. Observe the menus showing and app still capturing all mouse input (this is expected)

[...]

PR #4036, opened 2025-11-09
Implement xcb input shapes

DziubanMaciej:


actionless:
please include your test app (https://github.com/awesomeWM/awesome/issues/4035#issuecomment-3488255106) in a similar way how xeyes is being tested

example of test app built for test: https://github.com/search?q=repo%3AawesomeWM%2Fawesome%20test-systray.c&type=code

actionless:
also speaking of tests, i also see some of the existing ones are failing

DziubanMaciej:
> please include your test app ([#4035 (comment)](https://github.com/awesomeWM/awesome/issues/4035#issuecomment-3488255106)) in a similar way how xeyes is being tested
> 
> example of test app built for test: https://github.com/search?q=repo%3AawesomeWM%2Fawesome%20test-systray.c&type=code

Hmm, but `xeyes` is just launched, not really tested. The comment states that it exists just to stabilize code coverage. Is it even worth to add my app if we won't test it?

actionless:
unlike xeyes, your case actually testable - you could return specific exit code in your app if it being clicked in that shaped surface - and assert in test for that exit code

actionless:
(and also test that when clicked outside of area - it not registers input, e.g. not exiting with that code)

DziubanMaciej:
Okay, I guess you're right. But how would I simulate the click? There are not many tests that call external processes (only test-systray and test-gravity) and none of them do it.

actionless:
there are plenty of other tests who simulate clicks, here's the most canonical example:
https://github.com/awesomeWM/awesome/blob/bc57f0ff3678f71a09f919019677e4192533f713/tests/test-wibox-shape.lua#L51-L68

DziubanMaciej:
Yeah, but they don't spawn external processes... They create wiboxes which I assume is fully synchronous. It seems a lot simpler. 

In my case I would have to spawn the process and then wait until it's initialized and can actually receive mouseclicks, wouldn't I? Otherwise the simulated clicks could be signaled before the process even initialized X11 and they would be lost. 

Also, how would I detect that the window was not clicked? The test you sent relies on click count, but again, it looks like it's synchronous.

actionless:
there is `count` function argument which is used for async tests, see here for example:

https://github.com/awesomeWM/awesome/blob/master/tests/test-spawn.lua

actionless:
> Also, how would I detect that the window was not clicked? 

by placing wibox **below** the window and catching or not catching the click on it, and double-checking afterwards during the next 'count' that click wasn't received by the app as well (e.g. it still didn't exited with 'click exit code')

actionless:
just a small word of additional explanation - for handling async case, the sync test is being started over and over (increasing count argument), until test would return smth useful or crash

DziubanMaciej:
Yeah, the count parameter confused me greatly at first, but now I get it :D

Alright, I added the test as you proposed. Wibox is behind the test app. I'm simulating clicks and expecting them to happen either in wibox (detected with a signal) or in the app (detected in app's stdout).

When I click manually, it behaves correctly. However, fake clicks on the app don't work for some reason. Neither wibox nor app receives it. If I click on the app manually before the fake click, then it works. The first manual click isn't detected by the app. It's as if this first click "activates" the window. Why could it be? I tried to [forcefully focus and raise](https://github.com/awesomeWM/awesome/pull/4036/commits/71fcf18f9b34539ead0fb381f5d8b1f304847ba0#diff-b8a7e746f4a03f3bbd0ca4124a861ca6f0be1db64808c9a8737d6948d2b5f446R123-R135) the app window, but it doesn't help.

Could you take a look?

actionless:
> If I click on the app manually before the fake click, then it works. The first manual click isn't detected by the app

mb try to do nothing on first/few `count`, and try clicking when count is more than that

tests/test-client-shape.lua
```diff
@@ -39,15 +41,142 @@ runner.run_steps{
...
+            awesome.sync()
+
+            print("OK")
+        end
+        -- DEBUG????????????????????
+
+        mouse.coords{x=x, y=y}
+        assert(mouse.coords().x == x and mouse.coords().y == y)
+        root.fake_input("button_release", 1)
+        root.fake_input("button_press", 1)
```
actionless: i think you may blindly copy the mouse-dragging logic,
as you first releasing the left button, and next keep it (indefinitely) pressed

tests/test-client-shape.lua
```diff
@@ -39,15 +41,142 @@ runner.run_steps{
         assert(not surface.load_silently(c.client_shape_clip, false))
         assert(not surface.load_silently(c.shape_clip, false))
 
+        assert(surface.load_silently(c.client_shape_input, false))
+        assert(surface.load_silently(c.shape_input, false))
```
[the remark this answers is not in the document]
DziubanMaciej: Yeah, I think bounding and input are always there and only clip is special and can be null.

DziubanMaciej:
> > If I click on the app manually before the fake click, then it works. The first manual click isn't detected by the app
> 
> mb try to do nothing on first/few `count`, and try clicking when count is more than that

I added a sleep(2) and it's still not enough. I don't think it's a timing issue. Some awesome's state must be left not updated or something.

tests/test-client-shape.lua
DziubanMaciej: I don't follow. What mouse dragging logic? We want to click.

DziubanMaciej:
@actionless I think it's good now. The coverage report doesn't detect anything new.

actionless:
please squash into one commit (or wait til this one get merged 😸 https://github.com/awesomeWM/awesome/pull/4040 )

DziubanMaciej:
Already squashed. I got hit with the "check previous commits" check yesterday

actionless (dismissed):

actionless:
great, thanks! then that's all from my side 👍

DziubanMaciej:
@Aire-One, could you review this PR? Thanks!

DziubanMaciej:
:(

DziubanMaciej:
Cool, thanks!

Sinaaaa:
Maybe I have been overly dramatic with my issue reporting, I apologize. Anyhow would it perhaps be on the table to add a toggle switch to turn all this off in rc.lua, or perhaps make it optional via a rule based whitelist, I don't even know.

This is the first big breaking change I've seen in 4 years of using awesomeWM, though I do understand the use case, GW2 is unplayable without trails, though I'm partial to jokolay myself.

PR #4044, opened 2025-12-21
Fix nil reference in clienticon

SPFabGerman:
When a `clienticon` widget is created without specifying a client (for example when the widget is created declaratively), then the `client` attribute is set to `nil`. But the `property::icon` signal is called anyways, leading to an error because a `nil` value is referenced in the first condition of the `if`-clause.
I fixed this by swapping both conditions. Since the signal is only called for existing clients, `c` is never `nil` and so the check `obj._private.client == c` doubles to ensure that `obj._private.client` exists and that `obj._private.client.valid` can be referenced.

### DIFF
--- a/lib/awful/widget/clienticon.lua
+++ b/lib/awful/widget/clienticon.lua
@@ -119,7 +119,7 @@ end
 
 client.connect_signal("property::icon", function(c)
     for obj in pairs(instances) do
-        if obj._private.client.valid and obj._private.client == c then
+        if obj._private.client == c and obj._private.client.valid then
             obj:emit_signal("widget::layout_changed")
             obj:emit_signal("widget::redraw_needed")
         end


Aire-One:
What's your use case @SPFabGerman ?

The widget is meant to be instantiated, even in declarative structure. Just like `awful.titlebar.widget`. See examples https://awesomewm.org/apidoc/popups_and_bars/awful.titlebar.html

It doesn't make sense to me to use the `clienticon` widget without specifying the client it should source the icon from. 

Maybe an evolution could be to allow declarative usage with the named parameters pattern. But that's really just a stylistic improvement.

SPFabGerman:
My use case is as a template for both a tasklist and a taglist:
(The taglist code is pretty similar.)

```
    local function update_tasklist_callback(self, c, index, objects)
        self:get_children_by_id("client_icon_role")[1].client = c
    end

    local tasklist = awful.widget.tasklist {
        screen  = s,
		...
        widget_template = {
            id     = "background_role",
            create_callback = update_tasklist_callback,
            update_callback = update_tasklist_callback,
            {
                widget = wibox.container.place,
                {
                    {
                        margins = 2,
                        {
                            id     = "client_icon_role",
                            widget = awful.widget.clienticon,
                        },
                    },
                    {
                        id     = "text_role",
                    }
                }
            }
        }
    }
```

Because I use it in a template I can only specify the client with the two callbacks. But these seem to only be called after the widget has already been created. I know that I can use an `imagebox` widget with the `icon_role` id to solve this, but that doesn't have as high of a resolution.

The code has been adapted from https://awesomewm.org/apidoc/widgets/awful.widget.tasklist.html where pretty much the same thing is done.

Aire-One:
The `default_template` from `awful.widget.tasklist` does use the `awful.widget.clienticon` widget. So there is definitely something to dig.

https://github.com/awesomeWM/awesome/blob/9475bb98f08dc051cda4f8f8c1f047cdd52baadc/lib/awful/widget/tasklist.lua#L397

Aire-One:
Oops, forget to approve with my comment.

PR #4048, opened 2026-02-06
chore(options): remove unnecessary static modifiers

cnjhb:
These variables do not need to reside in memory.
Using the static modifier on a local variable causes the modified variable to reside permanently in memory.


Aire-One (approved):
Thank you, @cnjhb

Aire-One:
@actionless can you manually merge? Otherwise, we may need to change the codecov configuration to allow this kind of patch to be merged.

PR #4049, opened 2026-02-06
chore(awesome): use g_unix_fd_add instead of iochannel

cnjhb:
We don't need iochannel.
Since we only need to monitor the G_IO_IN event rather than actually using iochannel to receive or send anything, employing iochannel in this scenario is a waste.


Aire-One (approved):
Thank you, @cnjhb

PR #4050, opened 2026-02-06
chore: remove unnecessary static modifiers

cnjhb:
Like #4048, these variables can be destroyed after use and do not need to remain resident in memory.


actionless (approved):
thanks again! 😸

Aire-One (approved):
Thank you, @cnjhb

PR #4052, opened 2026-02-08
chore: remove unnecessary array initialization

cnjhb:
Structure initialization sets uninitialized fields to zero, just like global variables, so these `array_init` calls are unnecessary.
Reference: https://github.com/lvgl/lvgl/pull/9348

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -568,7 +568,7 @@ true_config_callback(const char *unused)
 int
 main(int argc, char **argv)
 {
-    string_array_t searchpath;
+    string_array_t searchpath = {};
     int xfd;
     xdgHandle xdg;
     xcb_query_tree_cookie_t tree_c;
@@ -589,7 +589,6 @@ main(int argc, char **argv)
     globalconf.exit_code = EXIT_SUCCESS;
     globalconf.api_level = awesome_default_api_level();
     buffer_init(&globalconf.startup_errors);
-    string_array_init(&searchpath);
 
     /* save argv */
     awesome_argv = argv;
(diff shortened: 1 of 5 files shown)


Aire-One (commented):
I'm not sure about this one.

While technically correct, structures created by the `DO_ARRAY` macro are expected to follow the `ARRAY_FUNCS` (and `ARRAY_COMMON_FUNCS`) contract, including their initializer. With this PR, we skip the `pfx##_array_init`.

I'm inclined to reject this PR.

cnjhb:
> I'm not sure about this one.
> 
> While technically correct, structures created by the `DO_ARRAY` macro are expected to follow the `ARRAY_FUNCS` (and `ARRAY_COMMON_FUNCS`) contract, including their initializer. With this PR, we skip the `pfx##_array_init`.
> 
> I'm inclined to reject this PR.

@Aire-One You're right, but we skip pfx##_array_init for the signals and properties fields of every lua_class_t instance.
From this perspective, as long as the structure is initialized to zero, pfx##_array_init can be skipped.
Since lua_class_t instances are global variables that default to zero, what I did is equivalent to what global variables do.

Aire-One:
Sorry for the delay @cnjhb. I took some time to poke around the code and think about this change.

As you mentioned, there are existing sites in the code where `pfx##_array_init` is not called, and I don't think it is worth enforcing the interface-contract-style now. I guess with this PR, we can rule that `pfx##_array_init` is not a mandatory initialization method for `ARRAY_COMMON_FUNCS` structures, but only useful as a reset, e.g., in combination/after a `pfx##_array_wipe` call.

Aire-One (approved):

Aire-One:
@actionless I think this PR will need manual merge.

PR #4055, opened 2026-02-10
fix(awesome): memory leak

cnjhb:
The variable `entry` was not freed.
Since the length wouldn't be too large, I directly used a variable-length array instead. Variable-length arrays offer better performance and eliminate the need to worry about memory release.


cnjhb:
The variable `xdgconfigdirs` is only used within the `for` loop, so I placed it inside the `for` loop.

cnjhb:
I was mistaken; there is no memory leak.

PR #4057, opened 2026-02-17
feat: add the mayused modifier

cnjhb:
Many of our macro-generated functions are never used. This causes the compiler to generate warnings. Functions decorated with the mayused attribute can eliminate these warnings.
I felt the original “attribute unused” was unintuitive, so I renamed it to mayused, meaning “maybe used.”
“attribute unused” always gives the impression that the function is completely unusable.
array.h also needs modification. If this PR gets merged, I'll make the changes then.

### DIFF
--- a/common/luaclass.h
+++ b/common/luaclass.h
@@ -23,6 +23,7 @@
 #define AWESOME_COMMON_LUACLASS
 
 #include "common/signal.h"
+#include "common/util.h"
 
 #include <lua.h>
 #include <lauxlib.h>
@@ -132,14 +133,14 @@ luaA_checkudataornil(lua_State *L, int udx, lua_class_t *class)
 }
 
 #define LUA_CLASS_FUNCS(prefix, lua_class) \
-    static inline int                                                          \
+    static inline int maybe_unused                                             \
     luaA_##prefix##_class_add_signal(lua_State *L)                             \
     {                                                                          \
         luaA_deprecate(L, "signal usage without add_signal()");                \
         return 0;                                                              \
     }                                                                          \
                                                                                \
-    static inline int                                                          \
+    static inline int maybe_unused                                             \
     luaA_##prefix##_class_connect_signal(lua_State *L)                         \
     {                                                                          \
         luaA_class_connect_signal_from_stack(L, &(lua_class),                  \
@@ -147,7 +148,7 @@ luaA_checkudataornil(lua_State *L, int udx, lua_class_t *class)
         return 0;                                                              \
     }                                                                          \
                                                                                \
-    static inline int                                                          \
+    static inline int maybe_unused                                             \
     luaA_##prefix##_class_disconnect_signal(lua_State *L)                      \
     {                                                                          \
         luaA_class_disconnect_signal_from_stack(L, &(lua_class),               \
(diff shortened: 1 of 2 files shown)


Aire-One:
> This causes the compiler to generate warnings.

How? I'm maybe missing something obvious here, but none of `make` recipes gave me this kind of output, and I don't recall seeing any in the Actions logs either.

I do agree that we probably have unused macro, tho. But I'm also unsure what are the drawback, if any, and how this new macro make things better. I may be lacking backgrounds on this (?)

cnjhb:
> > This causes the compiler to generate warnings.
> 
> How? I'm maybe missing something obvious here, but none of `make` recipes gave me this kind of output, and I don't recall seeing any in the Actions logs either.
> 
> I do agree that we probably have unused macro, tho. But I'm also unsure what are the drawback, if any, and how this new macro make things better. I may be lacking backgrounds on this (?)

When I run make, I get this output:
/data/data/com.termux/files/home/Documents/awesome/build/common/luaclass.c:138:1: warning: 
      unused function ‘lua_class_property_array_remove’ [-Wunused-function]
Are you using Clang for compilation?

Aire-One:
>  Are you using Clang for compilation?

I'm using GCC since that's what comes with the `build-essential` package and what we have in the CI. I've never really considered trying other compiler. 

I can confirm the warnings are present now I'm trying clang and the PR does a good job at removing them. I'm still unsure how useful it is to remove these warning for macro generated code. I.e., "yes", not all structures generated from macro has a usage of every function from the macro :shrug:. IMO, a more useful to test if generated function are used by at least one structure, so we can mark it as dead code that should be removed.

Quick note to open up the discussion: should we consider adding a workflow/matrix to test clang with our GitHub Actions workflow? I can see clang reveals some more additional warnings (I have not yet evaluated how useful they are).

cnjhb:
> > Are you using Clang for compilation?
> 
> I'm using GCC since that's what comes with the `build-essential` package and what we have in the CI. I've never really considered trying other compiler.
> 
> I can confirm the warnings are present now I'm trying clang and the PR does a good job at removing them. I'm still unsure how useful it is to remove these warning for macro generated code. I.e., "yes", not all structures generated from macro has a usage of every function from the macro 🤷. IMO, a more useful to test if generated function are used by at least one structure, so we can mark it as dead code that should be removed.
> 
> Quick note to open up the discussion: should we consider adding a workflow/matrix to test clang with our GitHub Actions workflow? I can see clang reveals some more additional warnings (I have not yet evaluated how useful they are).

It might not be very useful after all. Clang isn't much better than GCC. As long as it ensures no warnings under the latest GCC, that's sufficient.

Aire-One:
> It might not be very useful after all. Clang isn't much better than GCC. As long as it ensures no warnings under the latest GCC, that's sufficient.

As said in https://github.com/awesomeWM/awesome/pull/4060#discussion_r2836578872, maybe this can be part of a wider work to enforce coding rules on the project and raise the code quality.

actionless:
i like the idea but the label not very clear, mb smth like `ignore_unused_warning` or `ignore_unused` would be more clear

cnjhb:
> i like the idea but the label not very clear, mb smth like `ignore_unused_warning` or `ignore_unused` would be more clear

What do you think of the name `maybe_unused`? It’s a standard attribute introduced in C++17. I think it’s a better choice than `ignore_unused`.

actionless:
sounds reasonable 😺👍

PR #4040, opened 2025-11-17
chore(mergify): squash by default

actionless:

PR #4060, opened 2026-02-20
chore(objects): use luaA_class_add_properties

cnjhb:
Dependency #4059

Using `luaA_class_add_properties` to add properties is more memory-efficient and faster for batch property additions.
Only dynamic, individual property additions require `add_property`. But we have no such need whatsoever.
In fact, we have no requirement for dynamic property additions at all. Properties can be defined statically without needing to dynamically allocate any memory. There's simply no need for `add_properties` or `add_property`.

### DIFF
--- a/objects/client.c
+++ b/objects/client.c
@@ -4567,7 +4567,7 @@ client_class_setup(lua_State *L)
                      luaA_class_index_miss_property, luaA_class_newindex_miss_property,
                      client_methods, client_meta);
     luaA_class_set_tostring(&client_class, (lua_class_propfunc_t) client_tostring);
-    lua_class_property_t properties[] = {
+    const lua_class_property_t properties[] = {
         {
             .name = "name",
             .new = (lua_class_propfunc_t) luaA_client_set_name,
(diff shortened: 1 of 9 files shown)


objects/client.c
```diff
@@ -4568,7 +4568,7 @@ client_class_setup(lua_State *L)
                      luaA_class_index_miss_property, luaA_class_newindex_miss_property,
                      client_methods, client_meta);
     luaA_class_set_tostring(&client_class, (lua_class_propfunc_t) client_tostring);
-    lua_class_property_t properties[] = {
+    const lua_class_property_t properties[] = {
```
Aire-One: I guess this will only work once #4059 is merged.

objects/selection_watcher.c
```diff
@@ -20,7 +20,9 @@
  */
 
 #include "objects/selection_watcher.h"
+#include "common/luaclass.h"
 #include "common/luaobject.h"
+#include "glibconfig.h"
```
Aire-One: I think it's no needed

cnjhb: No, it's not needed. That was automatically inserted by clangd.

cnjhb:
> As a cleanup for dead code I think we should remove `luaA_class_add_property` now it's unused.

We should definitely remove add_property, but I'd prefer it to be the next PR.
I don't like cramming too much into a single commit, and we now squash multiple commits into one by default.

Aire-One:
> > As a cleanup for dead code I think we should remove `luaA_class_add_property` now it's unused.
> 
> We should definitely remove add_property, but I'd prefer it to be the next PR. I don't like cramming too much into a single commit, and we now squash multiple commits into one by default.

Agree. That's a good point to challenge https://github.com/awesomeWM/awesome/pull/4040.

PR #4061, opened 2026-02-20
refactor: use variable-length arrays instead of p_alloca

cnjhb:
Removing p_alloca is primarily for readability and maintainability.
Since p_alloca also performs an extra memset to zero the allocated memory, this change will also yield performance improvements.
Since variable-length arrays are already used in the scan function of the awesome.c file, my modification will not cause compatibility issues.

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -123,7 +123,7 @@ awesome_atexit(bool restart)
     }
 
     /* Save the client order.  This is useful also for "hard" restarts. */
-    xcb_window_t *wins = p_alloca(xcb_window_t, globalconf.clients.len);
+    xcb_window_t wins[globalconf.clients.len];
     int n = 0;
     foreach(client, globalconf.clients)
         wins[n++] = (*client)->window;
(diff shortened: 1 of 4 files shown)


actionless (approved):
thanks again! 👌😺

Elv13:
Question is, should we get rid of all the VLAs instead of adding new ones? It's a rather broken and deprecated C feature

actionless:
mb smth like this instead, to combine the benefits of both?

```
char stackbuf[1024];
char *buf = len <= sizeof(stackbuf) ? stackbuf : malloc(len);
...
if (buf != stackbuf)
    free(buf);
```

cnjhb:
> Question is, should we get rid of all the VLAs instead of adding new ones? It's a rather broken and deprecated C feature

VLA has indeed become an optional feature in the new standard.

However, it is also a flexible and natural way to allocate stack memory. As long as it is used properly, there will be no problem.

PR #4063, opened 2026-02-26
refactor: delete buffer_t

cnjhb:
buffer_t is a string type, and glib provides a ready-made GString implementation. We don't need to maintain our own.
Our buffer_t also handles stack allocation. When stack allocation is required, I use a variable-length array as a replacement.


actionless (dismissed):
great cleanup 🔥

actionless:
although it now conflicts with another your PR which got merged earlier

actionless (approved):

PR #4067, opened 2026-02-27
fix(naughty): add missing break in cleanup() active list removal

JimmyCozza:
Add `break` after `table.remove` in `cleanup()`'s active-list removal loop. Without it, `ipairs` skips the next entry after removal so destroying multiple notifications in quick succession can leave stale entries in `naughty.active`.

Present since commit 97417121. Found while working on SomeWM, fixed there first in trip-zip/somewm#274, contributing the fix back upstream.

Includes a regression test (`tests/test-naughty-destroy.lua`).


Aire-One (approved):
Thank you, @JimmyCozza

PR #4073, opened 2026-03-02
refactor: optmize luaA_class_setup

cnjhb:
The new `luaA_class_setup` preallocates the table, resulting in faster performance.
However, it can only accept arrays for methods and meta. Fortunately, all our `luaA_class_setup` use cases involve passing arrays.


actionless (approved):
thanks again!

PR #4077, opened 2026-03-08
fix(stack): Decouple fullscreen stacking layer from focus

### LINKED ISSUE #4075: Fullscreen Firefox intermittently stays visually on top, ignoring client focus
When Firefox is in fullscreen mode (F11), switching focus to another client on the same tag with `awful.client.focus.byidx()` (`Mod4+j/k`) or `awful.client.focus.history.previous()` (`Mod4+Tab`) (or any other method, I would imagine) intermittently leaves Firefox visually on top, even though the other client correctly receives focus and keyboard input.

It looks like Thunderbird does the same thing, so it might be something Mozilla-specific (rather than Firefox-specific). Didn't find any other client that does this, so far.

Seems like some kind of race condition? Maybe something unusual about how Firefox interacts with Awesome's stacking logic, manages `_NET_WM_STATE_FULLSCREEN` or re-requests geometry?

Not affected by config (happens with no `rc.lua` on bare defaults). Not affected by compositor (happens both with picom and without).

```
% awesome --version
awesome v4.3-1721-g39143f036-dirty (Too long)
 • Compiled against Lua 5.4.8 (running with 0.9.2)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • Transparency enabled: yes
 • Custom search paths: no
```
Built from https://aur.archlinux.org/packages/awesome-git , no modifications. I was on `4.3.630.g6eef064d` previously and this was not happening, so introduced somewhere in-between (not very helpful I guess...).

Also, just in case it ends up mattering:
```
Kernel:
local/linux 6.18.13.arch1-1

GPU:
2f:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Navi 21 [Radeon RX 6800/6800 XT / 6900 XT] (rev c0)

% firefox --version         
Mozilla Firefox 148.0b15                          
% thunderbird --version
Mozilla Thunderbird 148.0
```

1. Open Firefox (clean profile is fine).
2. Open something else on the same tag. My usual test subject is Chromium, but it also happens with urxvt, Cantata, etc.
3. Switch Firefox to fullscreen (either `modkey+F` or `F11`). For me this is always "firefox native" fullscreen (with panels hidden and coming up if you put the pointer towards the top of the screen); not sure if "awesome fullscreen" would cause the same effect (and not sure how to even make it happen without Firefox recognizing it).
4. Attempt switching focus to the "something else" from step 2 via `modkey+j/k` (or any other method, really), several times as needed.
5. Non-firefox client becomes focused-invisible.

Bonus steps for extra fun:
6. If the non-firefox client happens to be Chromium and is both fullscreen (just awesome fullscreen; this doesn't happen with Chromium native fullscreen) and focused-invisible, minimize and un-minimize it (`modkey+n`, `modkey+ctrl+n`).
7. Chromium is now invisible until its fullscreen state is cycled.

"Bonus" seems to be Chromium-specific; I haven't yet found other clients that would behave like that.

**Actual result:**

Focused non-Firefox client becomes invisible _sometimes._ Firefox stays visible. Focused-invisible window can be interacted with, but is not rendered.
Switching back and forth (via `modkey+j/k`) once more may fix this (so it looks like it's pretty much a random chance to non-render the client on every switch away from fullscreen Firefox). Switching to a different tag and back, or cycling the focused client's state (e.g. fullscreen/maximized) always fixes this.

However, if the "bonus steps" are taken, both switching back-and-forth and switching tags does nothing; Chromium stays invisible until its state is cycled.

**Expected result:**

Focused window becomes visible 100% of the time.

Watkurem:
Fixes https://github.com/awesomeWM/awesome/issues/4075

Fullscreen clients are assigned a special stacking layer via `client_layer_translator()`. Previously, this was conditioned on the client having focus. When switching focus away from a fullscreen client (e.g. via `awful.client.focus.byidx`), the fullscreen client could intermittently remain visually on top despite losing focus, because the layer reassignment depended on a restack that was not always triggered by the focus change.

This is fixed by conditioning the fullscreen layer on the client being at the top of the internal stack (i.e. the most recently raised client) rather than on focus. Since keyboard focus changes go through `raise()`, the fullscreen layer is correctly updated when switching away. Sloppy mouse-enter focus (which does not call `raise`) is unaffected, as it does not alter the internal stack order.

I'm wondering whether this has a potential of causing issues when the client is set as fullscreen _without_ raising (when could this happen? Could a client be spawned as fullscreen, but not raised? Should it even be given its own layer in that case?). One might switch to a client without raising it — but, once again, should the client be given its own layer in that case?

Admittedly, I'm not very familiar with Awesome's internals and don't _really_ know what I'm doing, but, hopefully, this is at least close to a proper fix.

I'm not sure how to properly auto-test for this... Let me know in case a test is required, and I'll try to figure it out.

### DIFF
--- a/stack.c
+++ b/stack.c
@@ -132,8 +132,12 @@ client_layer_translator(client_t *c)
     /* first deal with user set attributes */
     if(c->ontop)
         return WINDOW_LAYER_ONTOP;
-    /* Fullscreen windows only get their own layer when they have the focus */
-    else if(c->fullscreen && globalconf.focus.client == c)
+    /* Fullscreen windows only get their own layer when they are the most
+     * recently raised client.  Using the stack position (instead of focus)
+     * ensures that focus changes alone (e.g. sloppy / mouse-enter focus)
+     * do not alter the stacking order of fullscreen windows. */
+    else if(c->fullscreen && globalconf.stack.len
+            && globalconf.stack.tab[globalconf.stack.len-1] == c)
         return WINDOW_LAYER_FULLSCREEN;
     else if(c->above)
         return WINDOW_LAYER_ABOVE;


Watkurem:
I found an issue with my original attempt at this fix. Forcing a restack on every focus broke "sloppy focus" (as in, focusing a window when the mouse cursor enters it, but without raising it to visibility), which, at least for me, is a big deal. I didn't notice this problem at first, because all my fullscreen clients sit on the same tag (so it's normally impossible to do sloppy focus stuff with them) 🤡

After digging into it some more, I took a different approach, which seems to make more sense (and which I didn't find a way to break yet). Input would be appreciated, though.

actionless (approved):
thanks! 😺👍

Watkurem:
Is there anything else I need to do? Two reviews are required to merge

Aire-One (approved):
Thank you, @Watkurem

Aire-One:
@Mergifyio queue

PR #4084, opened 2026-03-13
fix(taglist): for property filter

cnjhb:
`table filter` interfere with `property filter`, so we need to remove `table filter` from the instance.

In practical terms, this means we cannot modify the filter property of a taglist instance.

### DIFF
--- a/lib/awful/widget/taglist.lua
+++ b/lib/awful/widget/taglist.lua
@@ -652,6 +652,7 @@ function taglist.new(args, filter, buttons, style, update_function, base_widget)
     })
 
     gtable.crush(w, taglist, true)
+    rawset(w, "filter", nil)
 
     gtable.crush(w._private, {
         style           = args.style or {},
--- a/tests/test-taglist.lua
+++ b/tests/test-taglist.lua
@@ -1,0 +1,20 @@
+-- Test the taglist
+
+local awful = require "awful"
+local runner = require "_runner"
+
+local taglist = awful.widget.taglist {
+    screen = screen[1],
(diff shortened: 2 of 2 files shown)


Aire-One:
The fix seems reasonable at first glance, but we do a lot of these `gears.table.crush` in our constructor-functions. There should be a lot of other properties wrongly copied this way in many modules. Applying this `rawset` to remove them afterward would just not be maintainable.

A better solution would be to basically refactor the module pattern we use and clearly separate the module level from instance level stuff. I'm currently experimenting [Teal](https://github.com/teal-language/tl). The type system it brings would help a lot in this scenario. The big migration, in addition to being a massive work would probably be not welcomed considering the current status of the project.

cnjhb:
> The fix seems reasonable at first glance, but we do a lot of these `gears.table.crush` in our constructor-functions. There should be a lot of other properties wrongly copied this way in many modules. Applying this `rawset` to remove them afterward would just not be maintainable.
> 
> A better solution would be to basically refactor the module pattern we use and clearly separate the module level from instance level stuff. I'm currently experimenting [Teal](https://github.com/teal-language/tl). The type system it brings would help a lot in this scenario. The big migration, in addition to being a massive work would probably be not welcomed considering the current status of the project.

It’s true—it would be much better if the language supported object-oriented programming.
I’ve used Penlight’s class before, but they weren’t ideal either.

cnjhb:
@Aire-One So how should we fix this?

actionless (approved):
thanks! mb additionally some note should be added to the docstring

PR #4087, opened 2026-03-16
chore(xutil): clear compile warnings

cnjhb:
warning: '__builtin_memcpy' specified bound between 18446744071562067968 and 18446744073709551615 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
   29 |   return __builtin___memcpy_chk (__dest, __src, __len,

The length parameter in `memcpy` cannot be negative.


actionless (dismissed):
thanks again!

actionless (approved):

PR #4088, opened 2026-03-16
chore(array): optmize array_splice

cnjhb:
This might reduce the overhead of some memcpy calls.


actionless (approved):

cnjhb:
This optimization is actually quite useful;
every `array_remove` and `array_take` call can benefit from it.

PR #4095, opened 2026-03-24
doc(naughty.notification): use message instead of text

cnjhb:
The text has been deprecated; we need to correct it in the document.


actionless (approved):
thanks!

Aire-One (approved):
Thank you, @cnjhb

PR #4103, opened 2026-05-19
Maintenance work for the `main` GitHub Action workflow

Aire-One:
This is a maintenance work for our main workflow. It aims at standardizing the Actions usages, which simplify a lot the workflow, and will also unblock easier support for Lua55.

The main change is the switch to use Actions `leafo/gh-actions-lua` and `leafo/gh-actions-luarocks` to respectively install Lua and Luarocks. I have already introduced the `luarocks/gh-actions-lua` a few times ago to manage LuaJIT, this work extends its usage for all the Lua versions we test. I switched from the `luarocks` fork back to the `leafo` upstream because it got updated and is now the more up-to-date version.

This PR also cleans the `env` and removes the `manual_screens` for `lua5.4` because I'm not sure what is the additional value it brings over the `codecov-lua5.3`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,8 +34,6 @@ jobs:
           - test_name: "lua5.4"
             lua_version: "5.4"
             lua_name: "lua5.4"
-            lua_library: "/usr/local/lib/liblua.a"
-            lua_include: "/usr/local/include"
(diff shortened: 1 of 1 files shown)


actionless:
thanks! 🔥

PR #4106, opened 2026-06-05
fix(naughty): honor beautiful.notification_border_width (fixes #3786)

### LINKED ISSUE #3786: Can't set notification borders with `theme.notification_border_width` and `theme.notification_border_color`

awesome v4.3-1588-gb54e50ad6 (Too long)
 • Compiled against Lua 5.4.4 (running with 0.9.2)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: no
 • execinfo support: yes
 • xcb-randr version: 1.6
 • Transparency enabled: yes
 • Custom search paths: no

Copy the default `theme.lua` in `/usr/share/awesome/defaults` to `~/.config/awesome`, and add the following lines after `-- Variables set for theming notifications:`:
```lua
theme.notification_border_width = 6
theme.notification_border_color = "#000000"
```
Modify the `beautiful.init(...)` line in `rc.lua` to `beautiful.init("~/.config/awesome/theme.lua")`.

Run `notify-send "test"` in terminal.

**Actual result:**

No border appears for the notification window.

**Expected result:**

The notification window would have a black border.

JimmyCozza:
Fixes #3786 

The theme variables are shadowed by a built-in default because the property lookup finds a built-in default first:

  - `n.border_width` resolves: `_private` -> `preset` -> `config.defaults`.
  - `config.defaults.border_width` is `dpi(1)`, so the chain always stops there.
  - `beautiful.notification_border_width` is never read. Same for margin, position, and every other themed key with a built-in default.
  - Legacy path is worse IMO: it merges `config.defaults` into the preset, then copies the preset into `_private`, so the chain stops at step 1.

  The fix puts the theme into the chain, above the built-in defaults:

  - Getter resolves: `_private` -> `preset` -> `beautiful.notification_<prop>`  -> `config.defaults`.
  - Legacy path no longer merges the defaults into the preset.

### DIFF
--- a/lib/naughty/core.lua
+++ b/lib/naughty/core.lua
@@ -92,6 +92,10 @@ gtable.crush(naughty, require("naughty.constants"))
 
 --- Defaults for `naughty.notify`.
 --
+-- These values apply only when neither the notification itself (constructor
+-- arguments or `ruled.notification` rules), its preset, nor the theme
+-- (`beautiful.notification_*`) provides a value.
+--
 -- @table naughty.config.defaults
 -- @tfield[opt=5] integer timeout
 -- @tfield[opt=""] string text
@@ -272,8 +276,11 @@ local function update_index(n)
     local s = get_screen(n.screen
         or (n.preset and n.preset.screen)
         or screen.focused())
+    local pos = n.position
+
     naughty.notifications[s] = naughty.notifications[s] or {}
-    table.insert(naughty.notifications[s][n.position], n)
+    table.insert(naughty.notifications[s][pos], n)
+    n._private.index_position = pos
 end
 
 --- Notification state.
@@ -490,11 +497,16 @@ local function cleanup(self, reason)
     end
     local scr = self.screen
 
-    assert(naughty.notifications[scr][self.position][self.idx] == self)
+    -- Use the position the notification was indexed under rather than its
+    -- current one. `position` can fall back to `beautiful.notification_position`,
+    -- and the theme changing emits no `property::position` to re-index with.
+    local pos = self._private.index_position or self.position
+
+    assert(naughty.notifications[scr][pos][self.idx] == self)
     remove_from_index(self)
 
     -- Update all indices
-    for k, n in ipairs(naughty.notifications[scr][self.position]) do
+    for k, n in ipairs(naughty.notifications[scr][pos]) do
         n.idx = k
     end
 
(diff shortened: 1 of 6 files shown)


lib/naughty/constants.lua
```diff
@@ -70,7 +70,6 @@ no_clear.defaults = {
     screen       = nil,
     ontop        = true,
     margin       = dpi(5),
```
actionless, line 72: but there are other default values still

so i don't think that your solution of removing one specific value from defaults is a correct direction - prolly the rootcause either in in get_value or in order how those are retrieved

JimmyCozza:
That's my bad.

I had originally made a much larger change in SomeWM, then I was worried about that big a change here in Awesome, so I overcorrected the wrong way and made too small a change.

I'll push up a change that I really think is more long-term solid, though I'm trying to be careful about repercussions.

JimmyCozza:
@actionless I updated this.  Let me know if this approach is something you'd be comfortable with.

One drawback to be aware of: this changes behavior for configs that set both `naughty.config.defaults.x` and  beautiful.notification_x`. 

The theme variable now takes precedence, which is how I understand the documentation to intend this to work, but still anyone who relied on that old behavior might see an issue.

actionless:
> The theme variable now takes precedence, which is how I understand the documentation to intend this to work

could you point what exactly in docs makes you think that way?

it doesn't match the overall pattern in rest of components, where instance value have priority over theme values

JimmyCozza:
Perhaps I worded that comment poorly.

Instance values still have priority over the theme.  The lookup order in this PR is:

1. Value set on notification (constructor args, setters, ruled.notification) - Instance Value
2. preset - Instance Value
3. beautiful.notification_x - theme value
4. naughty.config.defaults.x - built-in default

whereas the code currently on master just does this:

1. Value set on notification (constructor args, setters, ruled.notification) - Instance Value
2. preset - Instance Value
3. naughty.config.defaults.x - built-in default

So my assertion is that the documentation here:

https://awesomewm.org/apidoc/core_components/naughty.notification.html#border_width

specifically this line:

`beautiful.notification_border_width or 0`

indicates that the theme value will be respected at all, and if no value is set, it will default to 0.

Neither of those 2 documented cases happens.

The theme value is never respected.

It never hits the 0 or case either, the builtin value of dpi(1) always took precedence.

actionless:
> 1. Value set on notification (constructor args, setters, ruled.notification) - Instance Value
> 
>     2. preset - Instance Value
> 
>     3. beautiful.notification_x - theme value
> 
>     4. naughty.config.defaults.x - built-in default

yup, this version makes sense 😺👍

i'll review the actual PR code soon then

actionless (approved):
thanks!

PR #4107, opened 2026-06-18
fix(awful.titlebar): honor the urgent state for titlebar buttons

keryell:
## Problem

`awful.titlebar` already selects an `urgent` variant for the titlebar background and foreground (via `get_color()`), but the titlebar **button** widget (`titlebar.widget.button`) only distinguished the `normal` and `focus` states, and was not redrawn when a client's `urgent` property changed. As a result, an urgent (but unfocused) client kept showing its `normal` button images on top of the urgent-colored titlebar background.

## Fix

- Select button images by client state with an ordered, graceful fallback: `urgent → focus → normal`. Themes that ship dedicated `titlebar_*_button_urgent*` images get them; themes that don't fall back to the existing `focus`/`normal` images, so no existing theme changes behavior.
- Redraw the button on `property::urgent`.

## Compatibility

The fallback keeps the lookup identical for the `normal` and `focus` states on every existing theme. The default/xresources themes ship no `_urgent` button images, so urgent windows now show the `focus` glyphs instead of the `normal` ones — matching the already-urgent-aware titlebar background/foreground.

## Testing

Built and ran on a live session (awesome `v4.3` HEAD + this commit). Verified with a window forced `urgent` while unfocused: the titlebar buttons now follow the urgent/focus image set and update immediately when the urgent flag is set or cleared; normal and focused windows render their buttons unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/awful/titlebar.lua
+++ b/lib/awful/titlebar.lua
@@ -993,13 +993,18 @@ function titlebar.widget.button(c, name, selector, action)
                     img = "inactive"
                 end
             end
-            local prefix = "normal"
-            if c.active then
-                prefix = "focus"
+            -- Ordered list of state prefixes to try, most specific first, so a
+            -- theme that ships no images for the urgent (or focus) state falls
+            -- back gracefully: urgent -> focus -> normal.
+            local prefixes
+            if c.urgent then
+                prefixes = { "urgent", "focus", "normal" }
+            elseif c.active then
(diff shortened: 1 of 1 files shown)


Aire-One (commented):
Changes seems reasonable to me. But the new behavior needs to be documented in the API doc and beautiful variables.

PR #4109, opened 2026-07-14
refactor: use monotonic time instead of gettimeofday

cnjhb:
`gettimeofday` has been deprecated in the new POSIX standard,
and it is no longer monotonically increasing; an error will occur as soon as the user modifies the system time.

### DIFF
--- a/awesome.c
+++ b/awesome.c
@@ -64,7 +64,7 @@ awesome_t globalconf;
 static char **awesome_argv;
 
 /** time of last main loop wakeup */
-static struct timeval last_wakeup;
+static gint64 last_wakeup;
 
 /** current limit for the main loop's runtime */
 static float main_loop_iteration_limit = 0.1;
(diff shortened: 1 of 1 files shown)


cnjhb:
By the way, the new standard recommends using `clock_gettime` instead of `gettimeofday`.
`g_get_monotonic_time` is implemented using `clock_gettime` under the hood.

actionless (approved):
thanks!

awesome.c
```diff
@@ -888,8 +887,8 @@ main(int argc, char **argv)
     luaA_emit_startup();
 
     /* Setup the main context */
-    g_main_context_set_poll_func(g_main_context_default(), &a_glib_poll);
-    gettimeofday(&last_wakeup, NULL);
+    g_main_context_set_poll_func(g_main_context_default(), a_glib_poll);
```
Aire-One, line 890: Why do we lose the `&`?

cnjhb: At the time, I wanted to test whether the ‘&’ was really necessary.
Then this change accidentally got included in the commit.
However, this symbol is indeed redundant.

Aire-One (approved):
Thank you, @cnjhb

PR #4110, opened 2026-07-23
Support for Lua 5.5

Aire-One:
These commits were sleeping on my fork for some time now. The ecosystem is slowly progressing on its update to Lua 5.5. It's time we merge the support.

As for our used dependencies, we have two groups of still-non-lua-55-deps:

- Ldoc is still missing a proper release, but the Git version is ready, I used the rockspec from GitHub to install it in the CI for now.
- Luacheck and Depgraph still don't support Lua 5.5, but we only use them for `check_qa` and  `coverage` which are limited respectively to `fixed-lgi-lua5.2` and `codecov-lua5.3` matrices.

The fact Luacheck doesn't support Lua 5.5 is not a big issue for us. It means we cannot use the new language features since it would trigger linter issue. We would not even use them anyway since our code is supposed to work with Lua 5.1. Only the for-loop immutability of the control variable breaking change can bite our back. If someone introduce a new instance of this, it would be caught by the "build with Lua 5.5" job, so I guess that's ok.

I had to fully copy the updated version of `FindLua.cmake` to allow `cmake` to work properly with Lua 5.5. The alternative would have been to require a newer version of cmake to build Awesome, which is far more aggressive since most Linux distributions still doesn't ship it (including the one GitHub Actions provide, so we would have had to manually build cmake or trust an Actions that provides it).

### DIFF
--- a/cmake/FindLua.cmake
+++ b/cmake/FindLua.cmake
@@ -1,0 +1,355 @@
+# TODO: REMOVE THIS FILE WHEN WE CAN USE CMAKE 4.3.0 OR LATER
+#
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file LICENSE.rst or https://cmake.org/licensing for details.
+
+#[=======================================================================[.rst:
+FindLua
+-------
+
+Finds the Lua library:
+
+.. code-block:: cmake
+
+  find_package(Lua [<version>] [...])
+
+Lua is a embeddable scripting language.
+
+.. versionadded:: 3.18
+  Support for Lua 5.4.
+
+.. versionadded:: 4.3
+  Support for Lua 5.5.
+
+When working with Lua, its library headers are intended to be included in
+project source code as:
+
+.. code-block:: c
+
+  #include <lua.h>
+
+and not:
+
+.. code-block:: c
+
+  #include <lua/lua.h>
+
+This is because, the location of Lua headers may differ across platforms and may
+exist in locations other than ``lua/``.
+
+Result Variables
+^^^^^^^^^^^^^^^^
+
+This module defines the following variables:
+
+``Lua_FOUND``
+  .. versionadded:: 3.3
+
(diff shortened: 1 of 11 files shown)


cmake/FindLua.cmake
```diff
@@ -0,0 +1,353 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
actionless, line 3: i'd add `@TODO` comment with the text like 'remove me when lua 5.5 support in CMake became common across distros' or so

docs/ldoc.ltp
```diff
@@ -613,9 +613,9 @@
   <p>$(M(ldoc.full_description,nil))</p>
 # end
 
-# for kind, mods in ldoc.kinds() do
+# for k, mods in ldoc.kinds() do
 <h2>$(kind)</h2>
```
actionless: but on this line it's still `kind`?

lib/naughty/dbus.lua
```diff
@@ -322,8 +322,9 @@ function notif_methods.Notify(sender, object_path, interface, method, parameters
             end
 
             for k, v in pairs(args) do
-                if k == "destroy" then k = "destroy_cb" end
-                notification[k] = v
+                notification[
+                    k == "destroy" and "destroy_cb" or k
```
actionless: i'd do `(k == "destroy") and "destroy_cb" or k` for readabilty

actionless:
thanks a lot for looking into this!

lib/naughty/dbus.lua
Aire-One: fixed

docs/ldoc.ltp
Aire-One: fixed

cmake/FindLua.cmake
Aire-One: Well, I guess it's not something that will happen soon, but that's still nice to have this reminder.

PR #4112, opened 2026-07-26
Fix: CI "Run require test" is broken

### LINKED ISSUE #4111: CI "Run require test" is broken
Recent PRs are green, but the "codecov-lua5.3" logs are not good:

```
Run cd "/home/runner/work/awesome/awesome/build" && make check-requires
+ cd /home/runner/work/awesome/awesome/build
+ make check-requires
Checking use of require()
depgraph not found
module 'depgraph' not found:
	no field package.preload['depgraph']
	no file '/usr/local/share/lua/5.3/depgraph.lua'
	no file '/usr/local/share/lua/5.3/depgraph/init.lua'
	no file '/usr/local/lib/lua/5.3/depgraph.lua'
	no file '/usr/local/lib/lua/5.3/depgraph/init.lua'
	no file '/usr/share/lua/5.3/depgraph.lua'
	no file '/usr/share/lua/5.3/depgraph/init.lua'
	no file './depgraph.lua'
	no file './depgraph/init.lua'
	no file '/usr/local/lib/lua/5.3/depgraph.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/5.3/depgraph.so'
	no file '/usr/lib/lua/5.3/depgraph.so'
	no file '/usr/local/lib/lua/5.3/loadall.so'
	no file './depgraph.so'
(this error is non-fatal; we just skip its use)
Built target check-requires
```

1. This should be a fatal error
2. It needs to be fixed

Aire-One:
The matrix gate used for the `Run requires tests` step was incorrect. `depgraph` is only installed for `matrix.check_qa`. And it makes more sense to gate the actual `make check-requires` with this rather than `matrix.coverage`.

The initial implementation for `check_for_invalid_requires.lua` from https://github.com/awesomeWM/awesome/pull/1531 allowed the script to quietly skip if `depgraph` is not installed. IMO this is a bad idea since it was silently skipped in our CI since the GitHub Actions migrations (it seems), and we had no alert for that.

`depgraph` is unmaintained since 8 years, and its dependencies are explicitly constrained with `lua >= 5.1, < 5.4`. So with an up-to-date system, you can't install it. The CI will be ok since the `check_qa` matrix is only used by `fixed-lgi-lua5.2`, but it's kind of a dealbreaker for a developer, since `make check-requires` and `make check` are most likely to fail now. **This needs to be sorted out before merge, suggestions welcomed!**

Fix: #4111

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -295,7 +295,7 @@ jobs:
         run: cd "${{ github.workspace }}/build" && make check-examples
 
       - name: Run requires tests
-        if: matrix.coverage
+        if: matrix.check_qa
         run: cd "${{ github.workspace }}/build" && make check-requires
 
       - name: Run themes tests
--- a/build-utils/check_for_invalid_requires.lua
+++ b/build-utils/check_for_invalid_requires.lua
@@ -8,10 +8,10 @@
 local have_depgraph, depgraph = pcall(require, "depgraph")
 
 if not have_depgraph then
-    print("depgraph not found")
-    print(depgraph)
-    print("(this error is non-fatal; we just skip its use)")
-    os.exit(0)
+    io.stderr:write("depgraph not found\n")
+    io.stderr:write(tostring(depgraph) .. "\n")
+    io.stderr:write("depgraph is required for check-requires\n")
+    os.exit(1)
 end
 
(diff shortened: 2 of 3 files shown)


actionless (approved):
thanks for looking into this!

PR #4115, opened 2026-08-11
fix(permissions): inherit sticky state for transient clients

corpix:
Transient clients are now tagged with their parent client tags even when sticky, and inherit sticky state when the parent is sticky.

Do you see any reason why transient windows havent receive this tag in the past? Fixes #2268.

PR #4120, opened 2026-08-25
Add dedicated workflow for test-prev-commits

Aire-One:
This isolates the `Test previous commit` to its own workflow, so it makes the intention behind it more explicit and allow a better separation of concerns from the main workflow (not dependent on a Lua version, configuration, etc...) for the actual check it performs. i.e., in the recent PR #4115, we only see a red `Build & Test / fixed-lgi-lua5.2 ` and need to go to the details to see it's the `Test previous commits` step that failed.

I have also extracted the actual script to `build-utils/test-previous-commits.sh` so it can be run from a developer computer independently of the CI and works the same.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,7 +68,6 @@ jobs:
             tests_screen_size: "1921x1079"
             check_qa: true
             empty_theme_while_loading: true
-            test_prev_commits: true
 
     env:
       LUAINCLUDE: ${{ matrix.lua_include || '.lua/include' }}
@@ -240,9 +239,7 @@ jobs:
 
           AWESOME_VERSION="${AWESOME_VERSION}-g$(git rev-parse --short HEAD)"
 
-          if [ "${{ github.event_name }}" == "pull_request" &&  -n "${{ matrix.test_prev_commits }}" ]; then
-            AWESOME_VERSION="${AWESOME_VERSION}-PR${{ github.event.number }}"
-          elif [ "${{ github.event_name }}" == "pull_request" ]; then
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
             AWESOME_VERSION="v9999-PR${{ github.event.number }}"
           fi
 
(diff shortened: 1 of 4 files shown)


actionless:
in idea of moving it to a separate workflow instead of separate workflow-env i don't like that now everything (including installation steps) is duplicated 2 times and would require synchronized updating - sounds very fragile long-term

.github/workflows/main.yml
```diff
@@ -68,7 +68,6 @@ jobs:
             tests_screen_size: "1921x1079"
             check_qa: true
             empty_theme_while_loading: true
-            test_prev_commits: true
```
actionless, line 71: what about adding here ` - test_name: "test_previous_commits"...` ?

actionless:
> I have also extracted the actual script to build-utils/test-previous-commits.sh so it can be run from a developer computer independently of the CI and works the same.

this i like, in fact i think we need eventually to move everything which longer than one line from gh yml to either dedicated shell scripts or even a single makefile (prolly separate from main one)

Aire-One:
> in idea of moving it to a separate workflow instead of separate workflow-env i don't like that now everything (including installation steps) is duplicated 2 times and would require synchronized updating - sounds very fragile long-term

Yeah, agree it's not ideal. Just found https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action that may be a solution.

> what about adding here  - test_name: "test_previous_commits"... ?

Basically, that's what we had before.

My preference would be to move the test-prev-commit to its own workflow if possible. So I will try the composite-action-thing before fallbacking to this.

I want to move this check as a separate workflow to embrace more the GitHub Actions ideology in our tooling. This check is very specific, and I think it makes the more sense as a dedicated workflow instead of part of the main workflow. (and as a bonus it gets its own timeout which allows it to run longer compared to what it had after all the other steps from main)

> > I have also extracted the actual script to build-utils/test-previous-commits.sh so it can be run from a developer computer independently of the CI and works the same.
> 
> this i like, in fact i think we need eventually to move everything which longer than one line from gh yml to either dedicated shell scripts or even a single makefile (prolly separate from main one)

I partly agree with that. Specifically, I think the steps like `Install fresh xcb-errors` (and the likes) doesn't make sense outside the workflow and shouldn't be dedicated scripts.

.mergify.yml
```diff
@@ -9,6 +9,7 @@ shared:
     - check-success = luajit
     - check-success = fixed-lgi-lua5.2
     - check-success = lua5.1
+    - check-success = test-prev-commits
```
Aire-One, line 12: Thinking aloud :bulb: 

Maybe only have this new `check-success = test-prev-commits` condition in `automatic merge [MERGE]`, so mergify can ignore the test-prev-commits result when squash merge.

e.g. https://github.com/awesomeWM/awesome/pull/4115 is partly blocked because of unclean history and `Build & Test / fixed-lgi-lua5.2 (pull_request)` is red.

actionless:
but you wrote previously that the problem is:

> we only see a red Build & Test / fixed-lgi-lua5.2 and need to go to the details to see it's the Test previous commits step that failed.

so creating `- test_name: "test_previous_commits"` would display below PR as a separate check (disregard of it being or not being part of the same workflow as other tests)

__________


> I partly agree with that. Specifically, I think the steps like Install fresh xcb-errors (and the likes) doesn't make sense outside the workflow and shouldn't be dedicated scripts.

it's not only reusability (like debugging test setup locally inside a container), but just readability/maintainability - as it much easier to manage those scripts when they inside native .sh or Makefile's which are more suitable for such stuff than yaml

PR #4122, opened 2026-08-27
fix(awful.placement): round margins and offsets to whole pixels

### LINKED ISSUE #4121: awful.placement: fractional margin totals make a drawable grow on every placement until the stack overflows
### How to reproduce the issue?

Add two lines to the default `rc.lua`, after the `beautiful.init` call on line 41

```lua
beautiful.wibar_height  = 24
beautiful.wibar_margins = { top = 10.5, left = 10.5, right = 10.5 }
```

Start awesome.

### Output of `awesome --version`

awesome  (Too long)
 • Compiled against Lua 5.5.1 (running with 0.9.2)
 • API level: 4
 • D-Bus support: yes
 • xcb-errors support: yes
 • execinfo support: yes
 • xcb-randr version: 1.6
 • LGI version: /usr/local/share/lua/5.5/lgi/version.lua
 • Transparency enabled: yes
 • Custom search paths: no

### Actual Result

The wibar comes up at 411px instead of 24px, and the log fills with:

```
Oops, an error happened!
/usr/local/share/awesome/lib/gears/object/properties.lua:59: C stack overflow
```

### Expected Result

wibar has height of 24px.

### Default Config File

- [x] I can reproduce the same problem with default `rc.lua` file

### ARGB mode

- [ ] I experience visual glitches and tried running AwesomeWM in no-argb mode and still can reproduce the problem

### Relevant log output

```shell

```

### Upload screenshots

[live_overflow.png](https://github.com/user-attachments/assets/1dc2e245-0009-46b9-b0b5-92cdd355868a)

JimmyCozza:
Fixes #4121

## Cause

When awful.placement reads a drawable's geometry, it adds the margins to it, and when it writes the result back, it subtracts them again.  The C setter stored only whole pixels though.  So if the margins on an axis do not sum to a whole number, the add and subtract don't "cancel out."

With the issue's numbers (a top wibar of height 24, margins.top = 10.5):

  1. Read: d:geometry() returns 24, margins are added: 34.5
  2. align() does math.ceil(34.5) = 35
  3. Write back: 35 - 10.5 = 24.5, which the C setter rounds up to 25
  
Both roundings go up, so the drawable gains a pixel. 

### The problem is that, for an attached placement this recurses.  
- C setter emits property::height
- attach tracker re-runs the placement inside the signal
- grows by 1 pixel per pass until the C stack overflows.

## Fix

Round the margins and offsets to whole pixels in get_decoration(), and document that fractional values are rounded.

IMO, fractional margins could never be rendered exactly anyway, since drawin geometry is integer at the protocol level, so something has to round somewhere. Rounding at the input means every consumer of the margins (the placement math, the delta path, the struts) sees the same whole numbers, and the read/write round trip cancels exactly


Elv13 (approved):

Aire-One (approved):
Thank you, @JimmyCozza !

PR #4126, opened 2026-09-13
fix(awful.spawn): avoid O(n^2) complexity for output accumulation

shuber2:
## Commit message

In spawn.easy_async() the stdout and stderr lines are concatenated iteratively. Total costs grow as an arithmetic progression, i.e., are quadratic in the number of lines.

Fix this by collecting the lines in a table and performing a single concat(). This has optimal linear costs.

In a benchmark program the old and new line accumulation code was evaluated on lines comprising 100 chars. On a AMD Ryzen 7 PRO 7840U we obtain the following timings in seconds:

| lines    |  old (s)   |   new (s)  |    ratio |
| --       | --         | --         | --
|   250    |   0.0006   |    0.0000  |    17.8x |
|  1000    |   0.0061   |    0.0001  |    45.2x |
|  4000    |   0.1452   |    0.0007  |   197.5x |
| 16000    |   3.3125   |    0.0032  |  1048.9x |

## Background

For somewm I worked on a notification wibar widget, where a do-not-disturb inhibitor is auto-triggered when a screen capture is detected. For that the output of `pw-dump` is looked at every 3 seconds. This command produces 13k lines for stddout, and caused one core on my Ryzen 7 PRO 7840U saturized at 100% utilization.

The root-cause, the quadratic complexity of easy_async(), was found by Claude.

@JimmyCozza Would you then cherry-pick this over to somewm's release/1.4 and main branch?


Aire-One:
@shuber2, do you have the script used to get this benchmark numbers?

lib/awful/spawn.lua
```diff
@@ -444,16 +444,21 @@ end
...
+        table.insert(stdout_lines, str)
+        table.insert(stdout_lines, "\n")
     end
     local function parse_stderr(str)
-        stderr = stderr .. str .. "\n"
+        table.insert(stderr_lines, str)
+        table.insert(stderr_lines, "\n")
     end
     local function done_callback()
+        -- Avoid quadratic complexity by calling concat() once
```
Aire-One: This comment seems unnecessary and unhelpful without the context of the PR.

I suggest removing it

lib/awful/spawn.lua
```diff
@@ -444,16 +444,21 @@ end
...
 function spawn.easy_async(cmd, callback)
-    local stdout = ''
-    local stderr = ''
+    local stdout_lines = {}
+    local stderr_lines = {}
     local exitcode, exitreason
     local function parse_stdout(str)
-        stdout = stdout .. str .. "\n"
+        table.insert(stdout_lines, str)
+        table.insert(stdout_lines, "\n")
```
Aire-One: Why have a dedicated entry for `"\n"`?

I guess it could be part of the same entry `table.insert(stdout_lines, str .. "\n")`. Or even better, let `table.concat(std{out,err}, "\n")` do the hard work.

Aire-One (commented):
Basically, https://www.lua.org/pil/11.6.html

lib/awful/spawn.lua
shuber2: Of course, this is much cleaner, and faster too! Thanks for that. An additional `"\n"` needs to added to match the original output at the end, though.

lib/awful/spawn.lua
shuber2: Well, one might consider the temporary table an overhead that can be removed and effectively restore the old quadratic version. This comment should prevent this. But I also sympathize with your view here.

shuber2:
Thank you for your review, which improved the commit a lot.

Here is the benchmark program:
```lua
-- Benchmark: spawn.easy_async Output Accumulation

local function old_accumulate(n, line)
    local stdout = ''
    for _ = 1, n do
        stdout = stdout .. line .. "\n"
    end
    return stdout
end

local function new_accumulate(n, line)
    local stdout_lines = {}
    for _ = 1, n do
        table.insert(stdout_lines, line)
    end
    return table.concat(stdout_lines, "\n") .. "\n"
end

-- A representative line, similar in size to a `pw-dump` entry.
local line = string.rep("x", 100)
local sizes = {250, 1000, 4000, 16000}

print("=== spawn-accumulation ===")
print(string.format("%10s  %12s  %12s  %10s", "lines", "old (s)", "new (s)", "ratio"))

for _, n in ipairs(sizes) do
    collectgarbage("collect")
    local t0 = os.clock()
    local oldout = old_accumulate(n, line)
    local t1 = os.clock()
    local newout = new_accumulate(n, line)
    local t2 = os.clock()

    assert(oldout == newout, "mismatch at n=" .. n)

    local old_t, new_t = t1 - t0, t2 - t1
    print(string.format("%10d  %12.4f  %12.4f  %9.1fx",
        n, old_t, new_t, old_t / math.max(new_t, 1e-9)))
end
```

And here are the timings with lua5.4 and luajit:

```sh
❯ luajit tests/bench/bench-spawn-accumulation.lua
=== spawn-accumulation ===
     lines       old (s)       new (s)       ratio
       250        0.0005        0.0000       17.3x
      1000        0.0023        0.0000       65.8x
      4000        0.2571        0.0002     1112.8x
     16000        7.0743        0.0021     3308.8x

❯ lua5.4 tests/bench/bench-spawn-accumulation.lua
=== spawn-accumulation ===
     lines       old (s)       new (s)       ratio
       250        0.0003        0.0000       11.9x
      1000        0.0072        0.0001       53.4x
      4000        0.1869        0.0006      337.9x
     16000        3.4780        0.0024     1459.5x
```

In the timing stats in the commit message I switched to luajit. Note that the old version was particularly bad with luajit.

lib/awful/spawn.lua
Aire-One: > one might consider the temporary table an overhead that can be removed and effectively restore the old quadratic version.

I guess the best way to prevent that would be to add a comment at the table creation with a brief descition of the actual optimization. This comment is too obscure to be meaningful IMO.

I'm also ok without any comment.

Aire-One:
Thank you @shuber2 for the benchmark script!

lib/awful/spawn.lua
```diff
@@ -444,16 +444,18 @@ end
...
-        stdout = stdout .. str .. "\n"
+        table.insert(stdout_lines, str)
     end
     local function parse_stderr(str)
-        stderr = stderr .. str .. "\n"
+        table.insert(stderr_lines, str)
     end
     local function done_callback()
+        local stdout = table.concat(stdout_lines, "\n") .. "\n"
+        local stderr = table.concat(stderr_lines, "\n") .. "\n"
```
Aire-One: I'm not sure the `.. "/n"` is necessary. At least, it seems the current behavior doesn't always push it, hence the CI tests fails.

Aire-One (commented):
This implementation looks very clean! 😎

lib/awful/spawn.lua
shuber2: To restore the old behavior, it would be necessary. However, adding it unconditionally, even if stdout/stderr is empty, changes the behavior. Fixed it. Running tests at the moment.

actionless (approved):
thanks!

actionless:
> The root-cause, the quadratic complexity of easy_async(), was found by Claude.

getting back to the point of some absurd anti-ai crowd, like https://github.com/awesomeWM/awesome/discussions/4113 and few similar points in other threads

